PR #6761, opened 2017-05-09
Remove imports and headers that are unneeded now that the glue layer is in D

yebblies:


WalterBright (approved):

WalterBright:
Of course, please look at the continuous-integration failure first.

ibuclaw:
This is OK.

yebblies:
How are you meant to read that stupid jenkins log output?  @MartinNowak?

JackStouffer:
dub and vibe.d have a lot of transient errors. For Phobos we found that the only way to fix it is to rebase to restart Jenkins.

I've asked for him to remove the requirement for Jenkins to pass for merging but he didn't respond.

wilzbach:
> I've asked for Martin to remove the requirement for Jenkins to pass before merging but he didn't respond.

You need to ping @MartinNowak via mail as he unfortunately doesn't read through all the Gh notifications that he receives. Btw setting up mail filters for direct GH mentions isn't that difficult and maybe we should explain this more on the wiki, s.t. more members of the GH dlang organization configure filters if they can't deal with the GH noise anymore?

ibuclaw:
TBH, I would prefer notifications based on tags.  There used to be a bot that pinged you if you contributed recently to a certain source file also...

wilzbach:
> TBH, I would prefer notifications based on tags.

I think GH doesn't support this out-of-the-box, but it wouldn't be that hard to integrate this into [dlang-bot](https://github.com/dlang-bots/dlang-bot) in case you are interested ;-)

> There used to be a bot that pinged you if you contributed recently to a certain source file also...

This bot was disabled as too many people complained about the added noise (the configuration wasn't perfect), for more details:

https://github.com/dlang-bots/mention-bot

JackStouffer:
I don't understand why people have such a hard time with Github notifications; it's not that hard.

Either

1. Turn off the emails and just use the web interface. The email notification option sucks because you get an email for every single post and update. The web interface, on the other hand, shows which topics have new things you haven't seen, yet and doesn't bother telling you about every single new thing. Check in once per day, mute the topics you don't care about.
2. Click on the "Unwatch" button on the dlang repos. That way you'll only get notifications for when you're tagged.

ibuclaw:
Please take this to the ml, if you wish to discuss.

PR #6762, opened 2017-05-09
make -vcg-ast print all instantiated template

UplinkCoder:
This PR enables -vcg-ast to print the instancied versions of templates and mixin templates in fullDump mode (vcg-ast)

### DIFF
--- a/src/ddmd/hdrgen.d
+++ b/src/ddmd/hdrgen.d
@@ -164,7 +164,7 @@ public:
 
     override void visit(UnrolledLoopStatement s)
     {
-        buf.writestring("unrolled {");
+        buf.writestring("/*unrolled*/ {");
         buf.writenl();
         buf.level++;
         foreach (sx; *s.statements)
@@ -1461,12 +1461,7 @@ public:
         if (hgs.fullDump)
         {
             buf.writenl();
-            if (ti.aliasdecl)
-            {
-                // the ti.aliasDecl is the instantiated body
-                // if we have it, print it.
-                ti.aliasdecl.accept(this);
-            }
+            dumpTemplateInstance(ti);
         }
     }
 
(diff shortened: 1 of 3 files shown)


UplinkCoder:
@andralex I think you'll like this.

WalterBright:
No test case(s). Note that the new code is all marked red in the coverage tests.

UplinkCoder:
@WalterBright It's not something you can write tests for.
You could but that would mean you'd have maintain them on every semantic or ast change.

WalterBright:
You don't actually have to test the output. But it needs to be exercised.

UplinkCoder:
@WalterBright should be covered now

UplinkCoder:
I have no idea why this fails on  travis with gdc.
But this is primarily a developer-tool anyhow, and not for daily use.

ibuclaw:
Thou shalt not break gdc. :-)

ibuclaw:
There's probably nothing in this PR that breaks it, but as you've never added tests for `-vcg-ast` before, there may be something funny going on elsewhere that has never been caught.

ibuclaw:
Hmm, I seem to get a segfault in evalu8.c when testing locally, and I have no idea why, must be a g++-7 regression.  @WalterBright - looks like there's some wrong-code in evalu8 also.

ibuclaw:
Seems like it's a reference counting bug that is causing a double free.  First the data in `new Outbuffer` is freed by the dtor in `File`, then it is freed again by the dtor in OutBuffer.

ibuclaw (changes requested):
You must add `cgFile._ref = 1;` immediately after `cgFile.setbuffer()`.

https://github.com/dlang/dmd/blob/3698d759cbcb6ef72668db6fe01966ce3e133cd2/src/ddmd/mars.d#L1607-L1608

UplinkCoder:
@ibuclaw done. Thanks for the heads-up, I would not have found this one.

UplinkCoder:
@WalterBright, @andralex  ready for merge.

ibuclaw (approved):

PR #6763, opened 2017-05-09
fix Issue 17356 - [Reg 2.075] __simd_sto no longer executed

WalterBright:


WalterBright:
ping @MartinNowak

wilzbach:
> ping @MartinNowak

I think you need to ping Martin manually via mail..

WalterBright (commented):
approved

PR #6764, opened 2017-05-09
Add LINKsystem, remove dependency on Target from the parser

ibuclaw:
As requested by @yebblies - I don't think this should leak to the glue, if it does, then that should be plugged.

### DIFF
--- a/src/ddmd/attrib.d
+++ b/src/ddmd/attrib.d
@@ -33,6 +33,7 @@ import ddmd.mtype;
 import ddmd.parse;
 import ddmd.root.outbuffer;
 import ddmd.root.rmem;
+import ddmd.target;
 import ddmd.tokens;
 import ddmd.utf;
 import ddmd.utils;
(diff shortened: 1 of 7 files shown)


src/ddmd/attrib.d
```diff
@@ -499,7 +500,7 @@ extern (C++) final class LinkDeclaration : AttribDeclaration
     {
         super(decl);
         //printf("LinkDeclaration(linkage = %d, decl = %p)\n", p, decl);
-        linkage = p;
+        linkage = (p != LINKsystem) ? p : Target.systemLinkage();
```
WalterBright: This would flow better as:
```
(p == LINKsystem) ? Target.systemLinkage() : p;
```

ibuclaw: Yeah, I did, but then I remembered some obscure optimization detail that dmd considers the first branch as hot.  Maybe that's one micro-optimization too far though. ;-)

ibuclaw:
Rebased.

yebblies:
I don't like that we're discarding information as we build the parse tree, but it is the same with the current version.

ibuclaw:
> I don't like that we're discarding information as we build the parse tree

Mehhh... At least two errors have improved.

If I were thinking about having the frontend as a library, you would have a point though.

ibuclaw:
Is someone going to fix Jenkins, or...

PR #6766, opened 2017-05-10
fix Issue 14639 - Assigning init value to struct uses stack, causing …

WalterBright:
…segfault

### DIFF
--- a/src/dmd/aggregate.h
+++ b/src/dmd/aggregate.h
@@ -159,6 +159,7 @@ class StructDeclaration : public AggregateDeclaration
 public:
     bool zeroInit;              // !=0 if initialize with 0 fill
     bool hasIdentityAssign;     // true if has identity opAssign
+    bool hasBlitAssign;         // true if opAssign is a blit
     bool hasIdentityEquals;     // true if has identity opEquals
     bool hasNoFields;           // has no fields
     FuncDeclarations postblits; // Array of postblit functions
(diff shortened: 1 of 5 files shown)


UplinkCoder:
@WalterBright fails in runnable/sdtor.d

andralex:
I'm unclear about this. The test seems to imply that `.init` is an rvalue for all types. Is that the case?

ZombineDev:
You can't get the address of `.init` so yes, it behaves like an rvalue, even though it's stored in the binary, AFAIK.

Geod24:
You can safely ignore the C++ interop tests

WalterBright:
Howz about an approval, then? :-)

src/dmd/aggregate.h
```diff
@@ -159,6 +159,7 @@ class StructDeclaration : public AggregateDeclaration
 public:
     bool zeroInit;              // !=0 if initialize with 0 fill
     bool hasIdentityAssign;     // true if has identity opAssign
+    bool hasBlitAssign;         // true if opAssign is a blit
     bool hasIdentityEquals;     // true if has identity opEquals
     bool hasNoFields;           // has no fields
```
Geod24: Those booleans looks like they would be a good candidate for an `enum`. In a later PR, obviously.

Geod24 (commented):
Does it still work to disable `opAssign` ? Do we have a test for that ?

WalterBright:
The `hasBlitAssign` only is set for generated opAssign, which won't happen for disabled opAssign.

Geod24 (approved):

WalterBright:
Thank you. It'll be good to put this old issue to rest.

PR #6767, opened 2017-05-10
Run dmd internal unittests on CIs

wilzbach:
This will probably take a bit to configure for all CI test systems - especially for the auto-tester...

### DIFF
--- a/circleci.sh
+++ b/circleci.sh
@@ -114,6 +114,8 @@ coverage()
     make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=../_${build_path}/host_dmd clean
     make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=../_${build_path}/host_dmd ENABLE_COVERAGE=1
 
+    cp $build_path/dmd _${build_path}/host_dmd_cov
+    make -j1 -C src -f posix.mak MODEL=$MODEL HOST_DMD=../_${build_path}/host_dmd_cov ENABLE_COVERAGE=1 unittest
     make -j1 -C test MODEL=$MODEL ARGS="-O -inline -release" DMD_TEST_COVERAGE=1
 }
 
--- a/posix.mak
+++ b/posix.mak
@@ -5,7 +5,7 @@ ECTAGS_FILES = src/*.[chd] src/backend/*.[chd] src/root/*.[chd] src/tk/*.[chd]
 .PHONY: all clean test install auto-tester-build auto-tester-test toolchain-info
 
 all:
-	$(QUIET)$(MAKE) -C src -f posix.mak
+	$(QUIET)$(MAKE) -C src -f posix.mak all
 
 auto-tester-build: toolchain-info
 	$(QUIET)$(MAKE) -C src -f posix.mak auto-tester-build ENABLE_RELEASE=1
@@ -21,6 +21,7 @@ clean:
 	$(RM) tags
 
 test:
+	$(QUIET)$(MAKE) -C src -f posix.mak unittest
 	$(QUIET)$(MAKE) -C test -f Makefile
 
 html:
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -459,6 +459,11 @@ $G/dmd: $(DMD_SRCS) $(ROOT_SRCS) $G/newdelete.o $G/backend.a $G/lexer.a $(STRING
 	CC="$(HOST_CXX)" $(HOST_DMD_RUN) -of$@ $(MODEL_FLAG) -vtls -J$G -J../res -L-lstdc++ $(DFLAGS) $(filter-out $(STRING_IMPORT_FILES) $(HOST_DMD_PATH) $(LEXER_ROOT),$^)
 endif
 
+$G/dmd-unittest: $(DMD_SRCS) $(ROOT_SRCS) $G/newdelete.o $G/lexer.a $(G_GLUE_OBJS) $(G_OBJS) $(STRING_IMPORT_FILES) $(HOST_DMD_PATH)
+	CC=$(HOST_CXX) $(HOST_DMD_RUN) -of$@ $(MODEL_FLAG) -vtls -J$G -J../res -L-lstdc++ $(DFLAGS) -g -unittest -main $(filter-out $(STRING_IMPORT_FILES) $(HOST_DMD_PATH),$^)
+
+unittest: $G/dmd-unittest
+	$<
 
 clean:
 	rm -R $(GENERATED)
--- a/src/ddmd/mars.d
+++ b/src/ddmd/mars.d
@@ -1077,6 +1077,7 @@ Language changes listed by -transition=id:
  * Returns:
  *   Return code of the application
  */
+version(unittest) {} else
 int main()
 {
     import core.memory;


andralex:
works?

wilzbach:
> works?

Yes, it should work now. As the PR title states this allows to build all `unittest` blocks within the dmd codebase and runs them on our CI systems (CircleCi + auto-tester).
Also, the Makefile target can be invoked manually:

```
make -f posix.mak test
```

I will add a similar target for the Windows Makefiles once this has been approved.

src/posix.mak
```diff
@@ -459,6 +459,11 @@ $G/dmd: $(DMD_SRCS) $(ROOT_SRCS) $G/newdelete.o $G/backend.a $G/lexer.a $(STRING
 	CC="$(HOST_CXX)" $(HOST_DMD_RUN) -of$@ $(MODEL_FLAG) -vtls -J$G -J../res -L-lstdc++ $(DFLAGS) $(filter-out $(STRING_IMPORT_FILES) $(HOST_DMD_PATH) $(LEXER_ROOT),$^)
 endif
 
+$G/dmd-unittest: $(DMD_SRCS) $(ROOT_SRCS) $G/newdelete.o $G/lexer.a $(G_GLUE_OBJS) $(G_OBJS) $(STRING_IMPORT_FILES) $(HOST_DMD_PATH)
+	CC=$(HOST_CXX) $(HOST_DMD_RUN) -of$@ $(MODEL_FLAG) -vtls -J$G -J../res -L-lstdc++ $(DFLAGS) -unittest -main $(filter-out $(STRING_IMPORT_FILES) $(HOST_DMD_PATH),$^)
+
+test: $G/dmd-unittest
```
andralex: could you please call this target 'unittest'? It has that name in druntime and phobos.

andralex (commented):
awesome! now we need unittests :)

braddr:
This change is a good direction, but misses a few details about the current system.  There's a test target at the top level that invokes the deeper level directories as needed.  There's also a separation between build and test.  I don't think having the src level's auto-tester-build step invoking tests is correct.  The auto-tester-test target ought to do that.  Right now, the top level auto-tester-test target just invokes the test directories makefile.  Adding a line to have it invoke the src/posix.mak unittests would fit the current scheme better.

If you want to get fancy, the build steps could produce the dmd-unittest binary, but let the test steps invoke it, but that doesn't match how the other tests are treated.

posix.mak
```diff
@@ -5,7 +5,7 @@ ECTAGS_FILES = src/*.[chd] src/backend/*.[chd] src/root/*.[chd] src/tk/*.[chd]
 .PHONY: all clean test install auto-tester-build auto-tester-test toolchain-info
 
 all:
-	$(QUIET)$(MAKE) -C src -f posix.mak
+	$(QUIET)$(MAKE) -C src -f posix.mak all
```
wilzbach: This makes the default Makefile target explicit (`all` is the default target of `src/posix.mak`).

wilzbach:
Okay so the current status is that CircleCi works nicely again, but I am having troubles with the auto-tester as no `HOST_DC` is set on the "test dmd" step:

```
HOST_DC=/home/ec2-user/sandbox/at-client/release-build/install/linux/bin32/dmd
make -C src -f posix.mak unittest
no cpu specified, assuming X86
which: no dmd in (/usr/local/bin:/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/sbin:/opt/aws/bin)
make[1]: Entering directory `/home/ec2-user/sandbox/at-client/pull-2926855-Linux_32_64/dmd/src'
posix.mak:134: *** 'dmd' not found, get a D compiler or make AUTO_BOOTSTRAP=1.  Stop.
make[1]: Leaving directory `/home/ec2-user/sandbox/at-client/pull-2926855-Linux_32_64/dmd/src'
make: *** [test] Error 2
```

I tried to hack around this, but exposing `HOST_DC` would be very helpful.
I opened a PR at the auto-tester -> https://github.com/braddr/at-client/pull/6

wilzbach:
Hmm now that I reverted the hack to detect the already built D compiler and instead I was hoping that `HOST_DC` is now properly exported and available as environment variable, but sadly that doesn't seem to be the case:

 https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=2928984&isPull=true

Any ideas on what I could have missed? Should I go with the fallback hack which detects the already built D compiler for now?

braddr:
You hit some of the 3 build hosts that hadn't yet updated with the changes.  I took all of them out of the active fleet until they can be updated (I don't have remote access to them).  Slows the fleet down, but allows the tests to successfully build this change.

wilzbach:
> You hit some of the 3 build hosts that hadn't yet updated with the changes. I took all of them out of the active fleet until they can be updated (I don't have remote access to them). Slows the fleet down, but allows the tests to successfully build this change.

Awesome. Thanks a lot!
Then this should be ready to go :tada: 
CC @andralex @CyberShadow

posix.mak
```diff
@@ -21,6 +21,7 @@ clean:
 	$(RM) tags
 
 test:
+	$(QUIET)$(MAKE) -C src -f posix.mak unittest
```
ZombineDev: It looks like you don't have to modify `src/posix.mak`, if you simply replace here `unittest` with `ENABLE_DEBUG=1`.
(For reference, ,`ENABLE_DEBUG` does the following:
```Makefile
ifdef ENABLE_DEBUG
 CXXFLAGS += -g -g3 -DDEBUG=1 -DUNITTEST
 DFLAGS += -g -debug -unittest
````
).

I think going with this option, instead of introducing a new target there would be simpler/better. What do you think?

Comment (wilzbach) [reply posix.mak]: I prefer a separate target because then it's a lot easier to run the `unittest` target with the currently compiled `dmd` binary. Imho it's a common concept to use a different binary for the unittest executable and deviating from this might lead to confusion or annoyance down the road.

Comment (ZombineDev) [reply posix.mak]: Alright, you make a good point.

**Review by PetarKirov — APPROVED** [review]

File: circleci.sh [hunk 4]
```diff
@@ -114,6 +114,8 @@ coverage()
     make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=../_${build_path}/host_dmd clean
     make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=../_${build_path}/host_dmd ENABLE_COVERAGE=1
 
+    cp $build_path/dmd _${build_path}/host_dmd_cov
```
andralex: Would a `ln` suffice here?

posix.mak
andralex: Is there a disadvantage to assume the default target builds all? After all, if it's not it means we wrote that makefile poorly.

andralex (approved):
Couple optional suggestions for later.

posix.mak
wilzbach: Explicitly for the win - it was helpful during debugging this.

wilzbach:
>  to fix it by adding the main method back in but I'm not sure if that's the right solution since I don't quite understand why it was removed.

I'm copy/pasting my response from https://github.com/dlang/dmd/pull/7431:

The `main` method can't be part of the dmd unittest executable. This is a well-known limitation of D's unittest, but luckily [this will be fixed in 2.080](https://github.com/dlang/druntime/pull/1685). The correct fix is to remove `-unittest` from the `DDEBUG` flags in `win32.mak`:

https://github.com/dlang/dmd/blob/7c888c461d9c85addaed12c2f50489321305569e/src/win32.mak#L135

PR #6771, opened 2017-05-12
Fix issue 17392 - Add Dub file for the lexer and parser

jacob-carlborg:
The Dub file is intended to make the the compiler available as a library through Dub.

Currently the Dub package is only tested in Travis CI. I don't know if Dub is available in the autotester or not.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -1,0 +1,50 @@
+name "dmd"
+description "The DMD compiler"
+authors "Walter Bright"
+copyright "Copyright © 1999-2017, Digital Mars"
+license "BSL-1.0"
+
+targetType "none"
+dependency ":parser" version="*"
+
+subPackage {
+  name "root"
+  targetType "library"
+  sourcePaths "src/ddmd/root"
+}
+
+subPackage {
+  name "lexer"
+  targetType "library"
+  sourcePaths
+
+  sourceFiles \
+    "src/ddmd/console.d" \
+    "src/ddmd/entity.d" \
+    "src/ddmd/errors.d" \
+    "src/ddmd/globals.d" \
+    "src/ddmd/id.d" \
+    "src/ddmd/identifier.d" \
+    "src/ddmd/lexer.d" \
+    "src/ddmd/tokens.d" \
+    "src/ddmd/utf.d"
+
+  // stringImportPaths cannot be used because it will make Dub extremely slow
+  // https://github.com/dlang/dub/issues/1199
+  dflags "-J$PACKAGE_DIR"
+
+  dependency "dmd:root" version="*"
+}
+
+subPackage {
+  name "parser"
+  targetType "library"
+  sourcePaths
+
+  sourceFiles \
+    "src/ddmd/astbase.d" \
+    "src/ddmd/astbasevisitor.d" \
+    "src/ddmd/parse.d"
+
+  dependency "dmd:lexer" version="*"
+}
--- a/test/dub_package/.gitignore
+++ b/test/dub_package/.gitignore
@@ -1,0 +1,5 @@
+.dub
+docs.json
+__dummy.html
+*.o
+*.obj
--- a/test/dub_package/dub.sdl
+++ b/test/dub_package/dub.sdl
@@ -1,0 +1,5 @@
+name "dmd-dub-test"
+description "Test of the DMD Dub package"
+license "BSL 1.0"
+
+dependency "dmd" path="../../"
--- a/test/dub_package/source/app.d
+++ b/test/dub_package/source/app.d
@@ -1,0 +1,45 @@
+// The tests in this module are highlevel and mostly indented to make sure all
+// necessary modules are included in the Dub package.
+
+void main()
+{
+}
+
+// lexer
+unittest
+{
+    import ddmd.lexer;
+    import ddmd.tokens;
+
+    immutable expected = [
+        TOKvoid,
+        TOKidentifier,
+        TOKlparen,
+        TOKrparen,
+        TOKlcurly,
+        TOKrcurly
+    ];
+
+    immutable sourceCode = "void test() {} // foobar";
+    scope lexer = new Lexer("test", sourceCode.ptr, 0, sourceCode.length, 0, 0);
+    lexer.nextToken;
+
+    TOK[] result;
+
+    do
+    {
+        result ~= lexer.token.value;
+    } while (lexer.nextToken != TOKeof);
+
+    assert(result == expected);
+}
+
+// parser
+unittest
+{
+    import ddmd.astbase;
+    import ddmd.parse;
+
+    scope parser = new Parser!ASTBase(null, null, false);
+    assert(parser !is null);
+}
--- a/travis.sh
+++ b/travis.sh
@@ -54,6 +54,7 @@ rebuild() {
 
 # test druntime, phobos, dmd
 test() {
+    test_dub_package
     make -j$N -C ../druntime -f posix.mak MODEL=$MODEL unittest
     make -j$N -C ../phobos -f posix.mak MODEL=$MODEL unittest
     test_dmd
@@ -67,6 +68,13 @@ test_dmd() {
     else
         make -j$N -C test MODEL=$MODEL ARGS="-O -inline -release"
     fi
+}
+
+# test dub package
+test_dub_package() {
+    pushd test/dub_package
+    dub test
+    popd
 }
 
 for proj in druntime phobos; do


jacob-carlborg:
@andralex @RazvanN7 @MartinNowak ping.

jacob-carlborg:
Someone from the D foundation or the D core team would need to create an account on code.dlang.org to make the Dub package available.

dub.sdl
```diff
@@ -0,0 +1,63 @@
...
+license "BSL-1.0"
+
+targetType "library"
+
+// This is required because otherwise Dub will include all source files in src.
+// Just picked an existing directory which doesn't contain any D source files.
+sourcePaths "docs"
+
+// This is required because Dub needs to have at lease one source file
+sourceFiles "src/ddmd/package.d"
```
CyberShadow: Is there really no better solution here? Having to put hacks for our official package manager in our official compiler's package description is ... not great.

CyberShadow:
CC @s-ludwig @wilzbach

dub.sdl
s-ludwig: As far as I can see, changing to `targetType "none"` would allow dropping the `sourcePaths` and `sourceFiles` directives and should otherwise behave the same (except that it doesn't generate the extra dummy library).

dub.sdl
```diff
@@ -0,0 +1,63 @@
...
+  dflags "-J$PACKAGE_DIR"
+
+  preBuildCommands "dub run dmd:idgen -- $PACKAGE_DIR"
+  dependency "dmd:root" version="*"
+}
+
+subPackage {
+  name "parser"
+  targetType "library"
+  sourcePaths "docs"
```
CyberShadow: @s-ludwig Thanks! Is there a similar solution here to avoid the dummy `sourcePaths`, where only two `sourceFiles` are needed?

s-ludwig: `sourcePaths "docs"` can be replaced by an empty `sourcePaths`. Defining an empty list of source paths is enough to inhibit the auto-detection of the "src" folder.

dub.sdl
jacob-carlborg: Then it's not possible to build the main package:

```
Main package must have a binary target type, not none. Cannot build.
```

dub.sdl
jacob-carlborg: That worked, thanks.

dub.sdl
jacob-carlborg: Although it seems like it can still be used as a dependency without any problems. But the natural `dub build` doesn't work anymore.

dub.sdl
jacob-carlborg: Fixed.

dub.sdl
jacob-carlborg: Fixed, but not sure if it's correct.

s-ludwig: True, that's indeed a difference. But `dub build` would only build the dummy library. Dependencies are only built for dynamic library or executable targets. There is https://github.com/dlang/dub/issues/97, which would solve this issue.

jacob-carlborg: Aha, I see. At some point an executable or dynamic library needs to be built for a library to be useful. So I guess it kind of solves itself.

jacob-carlborg:
Anything else that needs to be taken care of before this can be merged?

andralex:
Cool beans! I'm on board with this, though it's a bit early :o). @MartinNowak could you do the honors of merging if all looks good?

jacob-carlborg:
> Cool beans! I'm on board with this, though it's a bit early :o)

Yes, but if we are very clear that we don't provide any stable API between release and use version `0.x.y`, I don't think it's any problem. Semantic versioning says:

"Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable"

We can append the DMD version as metadata, i.e. `0.0.1+2.075.0`. Hopefully this can speed up the process of making the frontend available as a library, since it will be more easily accessible. Two pull requests have already broken the separation between the parser and the rest of the frontend. The tests in this PR would catch that.

yebblies:
The problems with idgen really need to be fixed instead of hacked around.  It looks like we can replace idgen with ctfe, since we don't need `id.h` any more.  Any (LDC/GDC) glue uses could use `idPool` anyway.

ibuclaw:
This uses sdl :sadder:

jacob-carlborg:
> The problems with idgen really need to be fixed instead of hacked around. It looks like we can replace idgen with ctfe, since we don't need id.h any more. Any (LDC/GDC) glue uses could use idPool anyway.

Can we wait with that for another pull request?

yebblies:
> Can we wait with that for another pull request?

Why?  This PR should wait for that PR IMO, since nobody is likely to actually use this until we have at least a working parser exposed.  (See the existing dub package that exposes the lexer.)  I don't want the idgen hacks in master.

jacob-carlborg:
> Why?

Because I'm trying to keep the PR as small as possible. I've learned the hard way to not create too big PR's or change too much.

yebblies:
Yeah I'm not asking for that fix in this PR, I want it done in another PR first before this PR is merged.  Which would end up making this PR smaller.

jacob-carlborg:
> Yeah I'm not asking for that fix in this PR, I want it done in another PR first before this PR is merged. Which would end up making this PR smaller.

@yebblies I created #6837 for this.

UplinkCoder:
@jacob-carlborg idgen can be run as a prebuild step.

jacob-carlborg:
> idgen can be run as a prebuild step.

@UplinkCoder it already is run as a prebuild step [1], but https://github.com/dlang/dmd/pull/6771#issuecomment-302918182. I don't care which approach we use, please just come to an agreement.

[1] https://github.com/dlang/dmd/pull/6771/files#diff-406ece895a5b70a272cd76a2ef902836R41

jacob-carlborg:
@yebblies So Walter didn't like that the a header file wasn't generated anymore #6837. Would you like to try convince him or is the solution in this PR acceptable?

wilzbach:
@jacob-carlborg FYI: https://github.com/dlang/dlang.org/pull/1671 will expose the documentation pages for dmd on dlang.org ;-)

jacob-carlborg:
Now when https://github.com/dlang/dmd/pull/6837 has been merged I've update this pull request accordingly.

andralex (approved):

dub.sdl
```diff
@@ -0,0 +1,50 @@
+name "dmd"
+description "The DMD compiler"
+authors "Walter Bright"
+copyright "Copyright © 1999-2017, Digital Mars"
```
wilzbach: Should have been D Language Foundation ;-)

jacob-carlborg:
Thanks.

dub.sdl
jacob-carlborg: That's not what the rest of the source files says 😉.

wilzbach:
Btw the failure of gdc on Travis wasn't spurious (Travis isn't required as it unfortunately has too many spurious failures for the time being - there's [WIP to reduce them](https://github.com/dlang/dub/pull/1190)).

```
~/build/dlang/dmd/test/dub_package ~/build/dlang/dmd
+dub test
No source files found in configuration 'library'. Falling back to "dub -b unittest".
Performing "unittest" build using gdc for x86_64.
dmd:root ~master: building configuration "library"...
dmd:lexer ~master: building configuration "library"...
dmd:parser ~master: building configuration "library"...
cc1d: /home/build/share/cache/src/gcc-4.8.5/gcc/d/dfrontend/cppmangle.c:185: void CppMangleVisitor::source_name(Dsymbol*, bool): Assertion `t' failed.
../../src/ddmd/astbase.d: In member function 'addAlias':
../../src/ddmd/astbase.d:598:17: internal compiler error: Aborted
                 error("cannot have an import bind list");
                 ^
Please submit a full bug report,
with preprocessed source if appropriate.
See <https://bugzilla.gdcproject.org> for instructions.
gdc failed with exit code 1.
```

CC @ibuclaw

dub.sdl
wilzbach: Hehe, probably it was simply forgotten to update them

wilzbach: CC @WalterBright @andralex - can we replace/update the copyright notice from Digital Mars to D Language Foundation?

andralex: I'm okay with it. @WalterBright ?

wilzbach:
> Btw the failure of gdc on Travis wasn't spurious

See also https://github.com/dlang/dmd/issues/6999 for using the built compiler to run the test instead of the host compiler as all other test in this file do.

wilzbach:
Btw @MartinNowak @CyberShadow (see private email for credentials) - I tried to add dlang/dmd to code.dlang.org, but I got this nice error:

image

So how about pushing `v0.0.1` as long as [SemVer](https://trello.com/c/oz0dOIlY/321-semver-for-d) is still WIP?

ibuclaw:
> Btw the failure of gdc on Travis wasn't spurious.

Well you are using an older version of gdc. The failure itself is dmd frontend that doesn't have the same c++ support as more recent versions.

jacob-carlborg:
> So how about pushing v0.0.1 as long as SemVer is still WIP?

@wilzbach We could also set the version scheme that is currently used as metadata, i.e. `v0.0.1+2.075.0`.

ibuclaw:
By the way, doesn't the auto-merge label prevent broken PRs from being committed?  FYI @andralex on why you should not use merge directly.

CyberShadow:
@ibuclaw See https://github.com/dlang-bots/dlang-bot/issues/69

ibuclaw:
Hmm, and I suppose the autotester only merges if all of *its* tests pass.  So I guess I'm misremembering a time when this was not the case. ☺️

ibuclaw:
In the gdc repo, I protect master, and make it so that not even admins can force a merge if any required tests are failing.

Couldn't this be applied here too?

CyberShadow:
Yes, we do that too. Jenkins is not currently required because Dub/Vibe tests fail sporadically.

jacob-carlborg:
@CyberShadow but in this case the Jenkins tests passed but not Travis (gdc)

CyberShadow:
Oops, that's right.

Travis is not set as required here either. I don't know why, though.

jacob-carlborg:
IIRC, someone said that the macOS test on Travis were failing from time to time.

jacob-carlborg:
Not sure if it's possible to specify some of the tests from Travis to be required.

ibuclaw:
Travis has allow_failures or something akin to.

wilzbach:
> IIRC, someone said that the macOS test on Travis were failing

It's not only that, we also have random failures due to
- dlang.org being unreachable (it got a lot better since we use a second server as fallback, but it's still an issue sometimes
- code.dlang.org being unreachable (the dub binary is fetched from there even though it's nowadays bundled with DMD and LDC)
- random failures in the phobos testsuite
- reaching maximum run time
- failing to do any output on OSX
...

Just browse https://travis-ci.org/dlang/dmd/builds yourself.
So I think those were the reasons against setting it to required.

> Travis has allow_failures or something akin to.

Yes, but allowing failures for all builds isn't very useful.

jacob-carlborg:
> Just browse https://travis-ci.org/dlang/dmd/builds yourself.

Basically all of the ones I looked at was failing due to exceeding the time limit. One was failing due to `code.dlang.org` not being available.

wilzbach:
> So how about pushing v0.0.1 as long as SemVer is still WIP?

Martin was a bit worried that this might jeopardize the plans to [rectify SemVer at DMD](https://gist.github.com/mdparker/ab61a1309752905cf93acb2653c50beb) and confuse users.

@s-ludwig would it be possible to add DMD manually, s.t. only its branches can be used? I guess `~master` should be enough in the beginning to start testing this?

s-ludwig:
Easiest would be to add a dummy v0.0.1 tag, add it, and then remove the tag again. It will stay registered in that case.

wilzbach:
> Easiest would be to add a dummy v0.0.1 tag, add it, and then remove the tag again. It will stay registered in that case.

OK temporarily added v0.0.1:

https://github.com/dlang/dmd/releases/tag/v0.0.1

But I am still getting the error message - do I need to wait until it the cache is emptied?

image

s-ludwig:
Hm, just looked at the source and the registry pick up only the last 100 tags (in lexicographical order). Will open a ticket. You can try with v3.0.0 for the time being.

wilzbach:
> You can try with v3.0.0 for the time being.

Thanks - seems like this worked. Should be public soon:

image

wilzbach:
> Basically all of the ones I looked at was failing due to exceeding the time limit. One was failing due to code.dlang.org not being available.

That's good. At least our previous work has shown some fruits :)
The DUB failures should be gone once https://github.com/dlang/dub/pull/1190 and https://github.com/dlang/dub/pull/1198 have been merged and deployed.
The timeout issues don't affect PRs that much as [fewer arguments](https://github.com/dlang/dmd/blob/master/travis.sh#L67) are used. So we actually could try to require Travis.

What do you think @CyberShadow?

> I thought builds averaged 20 minutes. Is travisci that terrible now? Or is the CI tests just doing that much more.

I think it's a combination: performance of Travis is degrading over time due to more users and constantly new tests are added to the testsuite.

wilzbach:
> Should be public soon:

-> http://code.dlang.org/packages/dmd

@jacob-carlborg's unittest modified to an example:

```d
#!/usr/bin/env dub
/++ dub.sdl:
name "dmd_lexer_example"
dependency "dmd" version="~master"
+/
void main()
{
    import ddmd.lexer;
    import ddmd.tokens;
    import std.stdio;

    immutable sourceCode = "void test() {} // foobar";
    scope lexer = new Lexer("test", sourceCode.ptr, 0, sourceCode.length, 0, 0);
 
    while (lexer.nextToken != TOKeof)
        writeln(lexer.token.value);
}
```

@jacob-carlborg how about making an announcement of your great work?
@RazvanN7 how about using a DUB single file header for your examples?

dub.sdl
wilzbach: -> https://github.com/dlang/dmd/pull/7004

CyberShadow:
> What do you think @CyberShadow?

We can try it I guess. Stable too?

jacob-carlborg:
>  how about making an announcement of your great work?

Yes. That's a good idea 👍.

PR #6774, opened 2017-05-12
fix Issue 16197 - Constructors/postblits and destructors don't match …

WalterBright:
…up for array initialisation

### DIFF
--- a/test/runnable/sdtor.d
+++ b/test/runnable/sdtor.d
@@ -4154,6 +4154,25 @@ int test14815()
 static assert(test14815());
 
 /**********************************/
+// https://issues.dlang.org/show_bug.cgi?id=16197
+
+struct Elem {
+    static string r;
+    int x = -1;
+    this(this) { r ~= 'p'; printf("POSTBLIT %d\n", x++); }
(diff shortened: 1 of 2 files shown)


test/runnable/sdtor.d
```diff
@@ -4154,6 +4154,25 @@ int test14815()
 static assert(test14815());
 
 /**********************************/
+// https://issues.dlang.org/show_bug.cgi?id=16197
+
+struct Elem {
+    static string r;
+    int x = -1;
+    this(this) { r ~= 'p'; printf("POSTBLIT %d\n", x++); }
+    ~this()    { r ~= 'd'; printf("DTOR %d\n"    , x++); }
```
andralex: One unpleasant consequence of this idiom: calling `~=` in a destructor causes reallocation which may cause garbage collection, which aborts the program...

WalterBright: True, but this is a test case only, and in this case the destructors are not being called by the gc. So it's all right here.

andralex: The problem is real - it did occur in some of Lucia's code.

WalterBright: I know it is real, but not for this test case.

PR #6776, opened 2017-05-13
fix Issue 16540 - Attributes do not propagate correctly in lazy params

WalterBright:
Actually, it works now, I'm just adding a test case to ensure it continues to work.


wilzbach:
FWIW there's still [#12647](https://issues.dlang.org/show_bug.cgi?id=12647) for not-working attribute inference with lazy parameters, e.g.

```d
void assertThrown(E)(lazy E expression)
{
    expression();
}

void main() nothrow
{
    assertThrown( 0 );
}
```

PR #6777, opened 2017-05-13
fix Issue 17396 - Add colorized syntax highlighting to error messages…

WalterBright:
… emitted by dmd

This colorizes D code in between backticks in error messages.

### DIFF
--- a/src/ddmd/errors.d
+++ b/src/ddmd/errors.d
@@ -214,7 +214,15 @@ extern (C++) void verrorPrint(const ref Loc loc, COLOR headerColor, const(char)*
         fprintf(stderr, "%s ", p2);
     OutBuffer tmp;
     tmp.vprintf(format, ap);
-    fprintf(stderr, "%s\n", tmp.peekString());
+
+    if (global.params.color && strchr(tmp.peekString(), '`'))
+    {
+        colorSyntaxHighlight(&tmp);
+        writeHighlights(&tmp);
+        resetConsoleColor();
+    }
+    else
+        fprintf(stderr, "%s\n", tmp.peekString());
     fflush(stderr);
 }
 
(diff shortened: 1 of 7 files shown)


WalterBright:
Anyone have any idea why this is suddenly happening?
```
ddmd/backend/cc.d(15): Error: module dlist is in file 'tk/dlist.d' which cannot be read
```
The file is certainly there.

andralex:
cc @MartinNowak

rainers:
> Anyone have any idea why this is suddenly happening?

dlist.d is not on the command line of the library being compiled, so the module name should match the folder path, i.e. it should be module `ddmd.tk.dlist`, not `tk.dlist`.

WalterBright:
> dlist.d is not on the command line of the library being compiled, so the module name should match the folder path, i.e. it should be module ddmd.tk.dlist, not tk.dlist

Ok, but why is this suddenly happening? Why is it happening to this PR and not the others?

rainers:
> Ok, but why is this suddenly happening? Why is it happening to this PR and not the others?

You have added an import `ddmd.doc` to the lexer library, which now imports more or less the rest of the compiler.

WalterBright:
Oh, ok. That makes sense. Thanks!

WalterBright:
I'll leave it to someone with better color sense to pick better colors.

WalterBright:
It's working now! Of course, the rest of the error messages need to be marked with backticks for it to work with them.

andralex (approved):

CyberShadow:
@WalterBright Something is wrong here, looks like newlines are getting lost somewhere:

WalterBright:
yeah, fixed that in #6781

CyberShadow:
> yeah, fixed that in #6781

Confirmed, thanks.

CyberShadow:
> I'll leave it to someone with better color sense to pick better colors.

For one I think we should avoid red and leave that reserved for the "Error:" part.

ibuclaw:
Even just having them bold would be enough IMO.

ibuclaw:
> For one I think we should avoid red and leave that reserved for the "Error:" part.

You could make the same argument for Warning and Deprecation messages too.  Also supplemental and speculative messages if they get different highlights.  That doesn't really leave much else left to use.

ibuclaw:
> I'll leave it to someone with better color sense to pick better colors.

I may be biased towards cgdb here, but if you insist on it being coloured, the following groupings I find are the least surprising:

- keyword: blue (ie: __traits, mixin, foreach) 
- type: green (ie: Object, int, final)
- literal: red (ie: "string", 123, null)
- directive: cyan (ie: module, pragma, @safe)

grogancolin:
I also think a dmd.conf parameter to turn off colors globally is desirable. 

Colors mess with some folks terminals (different color schemes etc) - it should be easily disabled. (And preferrably not by adding --no-color on every command)

PR #6781, opened 2017-05-14
fix error recovery in color syntax highlighting

WalterBright:
The highlighter didn't recover well from errors in the highlighted section, sometimes leaving the console in a goofed up state. This should fix it. No test cases, because the highlighter is turned off when redirecting output.


andralex (approved):
Hrm, this is getting more convoluted than one would hope.

PR #6782, opened 2017-05-14
refactor: split off color handling into console.d

WalterBright:
This moves all the system-dependent color handling code into a separate module. System dependencies should not be in the main compiler logic. For example, main compiler logic should never be importing anything from `core.sys`.

Later on console.d should be refactored to be properly encapsulated. This PR just moves code.

### DIFF
--- a/src/ddmd/console.d
+++ b/src/ddmd/console.d
@@ -1,0 +1,123 @@
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Copyright:   Copyright (c) 1999-2017 by Digital Mars, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(DMDSRC _console.d)
+ */
+
+module ddmd.console;
+
+import core.stdc.stdio;
+import core.sys.posix.unistd;
+import core.sys.windows.windows;
+
+import core.stdc.string;
+import core.stdc.stdlib;
+
+version (Windows) extern (C) int isatty(int);
+
+enum COLOR : int
+{
+    COLOR_BLACK     = 0,
+    COLOR_RED       = 1,
+    COLOR_GREEN     = 2,
+    COLOR_BLUE      = 4,
+    COLOR_YELLOW    = COLOR_RED | COLOR_GREEN,
+    COLOR_MAGENTA   = COLOR_RED | COLOR_BLUE,
+    COLOR_CYAN      = COLOR_GREEN | COLOR_BLUE,
+    COLOR_WHITE     = COLOR_RED | COLOR_GREEN | COLOR_BLUE,
+}
+
+alias COLOR_BLACK = COLOR.COLOR_BLACK;
+alias COLOR_RED = COLOR.COLOR_RED;
+alias COLOR_GREEN = COLOR.COLOR_GREEN;
+alias COLOR_BLUE = COLOR.COLOR_BLUE;
+alias COLOR_YELLOW = COLOR.COLOR_YELLOW;
+alias COLOR_MAGENTA = COLOR.COLOR_MAGENTA;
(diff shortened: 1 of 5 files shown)


src/ddmd/console.d
```diff
@@ -0,0 +1,123 @@
...
+}
+
+alias COLOR_BLACK = COLOR.COLOR_BLACK;
+alias COLOR_RED = COLOR.COLOR_RED;
+alias COLOR_GREEN = COLOR.COLOR_GREEN;
+alias COLOR_BLUE = COLOR.COLOR_BLUE;
+alias COLOR_YELLOW = COLOR.COLOR_YELLOW;
+alias COLOR_MAGENTA = COLOR.COLOR_MAGENTA;
+alias COLOR_CYAN = COLOR.COLOR_CYAN;
+alias COLOR_WHITE = COLOR.COLOR_WHITE;
```
andralex: ditto... looking forward to get rid of all shouting

src/ddmd/console.d
```diff
@@ -0,0 +1,123 @@
...
+{
+    version (Windows)
+    {
+        HANDLE h = GetStdHandle(STD_ERROR_HANDLE);
+        WORD attr = consoleAttributes(h);
+        SetConsoleTextAttribute(h, attr | (bright ? FOREGROUND_INTENSITY : 0));
+    }
+    else
+    {
+        fprintf(stderr, "\033[%dm", bright ? 1 : 0);
```
andralex: s/bright ? 1 : 0/bright/

src/ddmd/console.d
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+extern (C++) void setConsoleColor(COLOR color, bool bright)
+{
+    version (Windows)
+    {
+        HANDLE h = GetStdHandle(STD_ERROR_HANDLE);
+        WORD attr = consoleAttributes(h);
+        attr = (attr & ~(FOREGROUND_WHITE | FOREGROUND_INTENSITY)) | ((color & COLOR_RED) ? FOREGROUND_RED : 0) | ((color & COLOR_GREEN) ? FOREGROUND_GREEN : 0) | ((color & COLOR_BLUE) ? FOREGROUND_BLUE : 0) | (bright ? FOREGROUND_INTENSITY : 0);
```
andralex: this line is way long

src/ddmd/console.d
```diff
@@ -0,0 +1,123 @@
...
+        SetConsoleTextAttribute(h, attr);
+    }
+    else
+    {
+        fprintf(stderr, "\033[%d;%dm", bright ? 1 : 0, 30 + cast(int)color);
```
andralex: comment as to where the 30 comes from? i.e. link to a reference?

src/ddmd/console.d
```diff
@@ -0,0 +1,123 @@
...
+    }
+}
+
+extern (C++) void resetConsoleColor()
```
andralex: this should return bool

src/ddmd/console.d
```diff
@@ -0,0 +1,123 @@
...
+        WORD attr = consoleAttributes(h);
+        SetConsoleTextAttribute(h, attr | (bright ? FOREGROUND_INTENSITY : 0));
+    }
+    else
+    {
+        fprintf(stderr, "\033[%dm", bright ? 1 : 0);
+    }
+}
+
+extern (C++) void setConsoleColor(COLOR color, bool bright)
```
andralex: should return bool

andralex (changes requested):
have enough nits to ask for a second pass

WalterBright:
I agree with all of your nits, and plan on re-engineering it to fix them and make it more useful. However, all this PR is supposed to do is split it off.

andralex (approved):

src/ddmd/console.d
adamdruppe: that's an ansi SGR sequence for set foreground color: https://en.wikipedia.org/wiki/ANSI_escape_code

tbh if you don't already know that, you probably shouldn't be editing this code so I feel a comment is redundant; it is plainly obvious to the most casual observer.

PR #6786, opened 2017-05-14
staticassert.d: fix error messages to use syntax highlighting

WalterBright:


test/fail_compilation/ctfe14465.d
```diff
@@ -1,22 +1,22 @@
...
+        super(msg);
+    }
+}
+
+bool foo()
+{
+    throw new E("message");
+}
+
+static assert(foo());
```
CyberShadow: What happened here? Are the line endings correct?

CyberShadow: Looks like they were CRLF and now they are LF (as they should have been).

CyberShadow (approved):

test/fail_compilation/ctfe14465.d
WalterBright: I didn't realize that change happened, but part of my submission process is to convert to LF line endings.

CyberShadow: PR to fix the rest of them: https://github.com/dlang/dmd/pull/6793

CyberShadow:
Please fix conflicts

PR #6793, opened 2017-05-15
Convert line endings from DOS to UNIX

CyberShadow:
`testreturn.d` even had mixed line endings! :nauseated_face:


ibuclaw (approved):

yebblies:
The problem with this is that it reduced diversity in the test suite.  Of course you shouldn't be using mixed line endings in actual D code, but we should make sure it's covered by the test suite.  Does this PR now mean we have zero coverage of mixed line endings?

On the other hand none of these files are really an appropriate place for this.

CyberShadow:
> but we should make sure it's covered by the test suite

No argument there, but it should be in separate tests that explicitly test for it, and probably encoded in such a way that's not affected by endline conversions and git's `core.autocrlf` setting.

CyberShadow:
> Does this PR now mean we have zero coverage of mixed line endings?

Looks like it. Doesn't help that the red X doesn't mean anything these days grrr @wilzbach @MartinNowak

yebblies:
> No argument there, but it should be in separate tests that explicitly test for it

That should have been a blocker for this PR.

> and probably encoded in such a way that's not affected by endline conversions and git's core.autocrlf setting.

If you use autocrlf you get what you deserve.

CyberShadow:
> That should have been a blocker for this PR.

Grats, you've successfully guilt-tripped me to go write a test: https://github.com/dlang/dmd/pull/6817

> If you use autocrlf you get what you deserve.

Should @WalterBright decide he needs to set up git on a new Windows PC and it ends up being enabled, we *all* will have to deal with the consequences :)

ibuclaw:
Right, so a few tests could be created similar to helloUTF16.d then (helloCRLF.d and helloCRLFmixed.d). Though while you are at it, why not helloCR.d, helloRS.d, helloLFCR.d, and helloEBCDIC.d too? 😉.  Any such files would need a large warning on top to prevent people auto converting it back to Unix.

Although we aren't a language that cares about whitespace, so I doubt we lost any coverage from this.

CyberShadow:
> Right, so a few tests could be created similar to helloUTF16.d then (helloCRLF.d and helloCRLFmixed.d).

In #6817 I escaped the newlines by generating the .d file from a bash script, so it's certain that they will not be mangled by anything.

> Though while you are at it, why not helloCR.d, helloRS.d, helloLFCR.d, and helloEBCDIC.d too? 😉. 

D doesn't support any of those. It's in the spec, right?

CyberShadow:
> Although we aren't a language that cares about whitespace, so I doubt we lost any coverage from this.

Err, I thought Daniel's comment above was specifically a complaint that we lost the last remnant of coverage for `\r`.

ibuclaw:
> D doesn't support any of those. It's in the spec, right?

My expectation is that we are using something like `while (*p.isspace) p++;` to skip non-identifier characters.

WalterBright:
You can also test CR's by generating a string with CR's in it, then mixin it.

You can tell git was developed on Linux because it just doesn't do well with CR's. I've tried the various autocrlf settings, none of them worked. What did work was write a submit script that removed the CR's.

CyberShadow:
> You can also test CR's by generating a string with CR's in it, then mixin it.

That was proposed, see above. Can't test DDoc with it.

I'm just going to drop the last commit. Not only do I think explicit escaping is the correct way to explicitly test for such things, but it's also apparently simpler. If you think it must be changed, I'm also enabling "edits from maintainers".

WalterBright:
> I'm just going to drop the last commit.

Not sure what you mean by that, as this PR was already merged.

CyberShadow:
Oops, I thought this was #6817.

yebblies:
> You can tell git was developed on Linux because it just doesn't do well with CR's. I've tried the various autocrlf settings, none of them worked. What did work was write a submit script that removed the CR's.

I don't know what exactly you tried, but `autocrlf = false` works perfectly.  Git will just leave the line endings alone.  Then as long as your editor can handle unix line endings (and you really should be using one that can) then everything will work just fine.

WalterBright:
I tried it and it did not work just fine. I had constant issues with git deciding the whole file changed because the line endings changed. As I recall, the documentation for autocrlf was a bunch of handwavy nonsense.

Stripping the CRs from checkins with a script solved the problem.

CyberShadow:
> I had constant issues with git deciding the whole file changed because the line endings changed.

That was probably either because you changed it *after* cloning the repository (you have to do that before, or hard-reset after you set it), or because your editor was not preserving line endings.

WalterBright:
I got really tired of the files being the "same only different" which would confuse the heck out of git. The git documentation was no help at all, and my script works, and the problem is solved.

CyberShadow:
> and the problem is solved

I think we need to take the time to understand the tools we rely on, how they work and how to use them correctly. Otherwise, it becomes a problem for everybody. Dismissing them away and reinventing the wheel is not progress.

WalterBright:
It's not possible to understand everything about every tool we use.

I run a checkin script that removes trailing spaces, replaces tabs with spaces, and strips CRs. There was an agreement a while back that code checked in here has no tabs, no trailing spaces, and no CRs. There's no problem that needs solving.

> Then as long as your editor can handle unix line endings (and you really should be using one that can) then everything will work just fine.

With the script I don't need to know nor care about what the editor does (I use different editors for different purposes) or what any file modification tool does, for that matter.

CyberShadow:
> It's not possible to understand everything about every tool we use.

If we're at the point where we're inventing new tools because we can't understand the tool's documentation, that warrants reconsideration.

> There was an agreement a while back that code checked in here has no tabs, no trailing spaces, and no CRs.

It seems that only yesterday the agreement was the polar opposite, that CRs should be allowed, and anything mangling them should be forbidden.

> With the script I don't need to know nor care about what the editor does (I use different editors for different purposes) or what any file modification tool does, for that matter.

That's just sweeping the problem under the rug. You have a git misconfiguration problem that you work around by writing a tool that repairs files corrupted by git, when it's easier to just avoid corrupting them in the first place.

The only editor I know of that does not support UNIX line endings is Windows Notepad.

WalterBright:
> If we're at the point where we're inventing new tools because we can't understand the tool's documentation,

Or maybe it was a bug in git at the time (CRLF bugs in Unix code ported to Windows are commonplace).

Need a script anyway for the tabs and trailing spaces.

> that CRs should be allowed, and anything mangling them should be forbidden.

I don't recall such agreement. Nor do I think it is very practical as then submitters have to vet what tools they use to edit source code. After all, if a tool is used that does all LFs from mixed CRLF and LF, then there's no way to set it back, as the information is lost. If CRs matter for some reason, there's no way to look at the file to see if it has CRs or not, short of using a file dumper.
 
> The only editor I know of that does not support UNIX line endings is Windows Notepad.

Many tools edit source files, there's >file redirections, etc. How can you be sure they all "do the right thing" with CR/LF? Notepad is just the start.

There's nothing wrong with running a script to "canonicalize" submissions to git.

The only point to CRs is to test the line handling coverage. Which #6817 does (thank you!), so that issue is now resolved.

CyberShadow:
> Or maybe it was a bug in git at the time (CRLF bugs in Unix code ported to Windows are commonplace).

`core.autocrlf` was written specifically for Windows, so that's highly doubtful.

> Nor do I think it is very practical as then submitters have to vet what tools they use to edit source code. After all, if a tool is used that does all LFs from mixed CRLF and LF, then there's no way to set it back, as the information is lost. If CRs matter for some reason, there's no way to look at the file to see if it has CRs or not, short of using a file dumper.

I'm glad we agree on this :) Daniel and Iain seem to disagree though.

> There's nothing wrong with running a script to "canonicalize" submissions to git.

OK, that's fair enough. Agreed.

WalterBright:
> core.autocrlf was written specifically for Windows, so that's highly doubtful.

At the risk of continuing to sell after the sale has been closed :-) as I recall I gave up on git's handling of the issue when one part of git said I had modified files that needed committing and the committer would insist that the files were the same and wouldn't commit them. That clearly was a bug in git.

WalterBright:
BTW, I had significant experience developing for both Unix and Windows at the time D was started, which is why D works smoothly across systems. The line endings, \ or / in paths, case sensitivity in file names, drive letters, etc., are all built in properly to the compiler and Phobos.

yebblies:
> At the risk of continuing to sell after the sale has been closed :-) as I recall I gave up on git's handling of the issue when one part of git said I had modified files that needed committing and the committer would insist that the files were the same and wouldn't commit them. That clearly was a bug in git.

Yeah I've seen that happen before - and it is absolutely solved by setting autocrlf to off.  IIRC it's also not a bug, it's just a consequence of autocrlf being terrible.  By turning autocrlf off you effectively get 'binary' mode files and git doesn't ever screw around with your line endings.

Your approach works with dmd/etc because those projects are already consistent in their use of line endings/tabs/etc.  If you ever need to work on a project that uses windows line endings your best bet is to turn autocrlf off and leave it off.

Disclaimer: autocrlf really needs to be turned off before cloning, or better yet as soon as git is installed.

PR #6794, opened 2017-05-15
refactor: improve console.d encapsulation

WalterBright:
1. encapsulated state in a `struct Console`
2. eliminated random global variables
3. removed hardcoding of `stderr`
4. added documentation
5. removed grossly excessive number of system calls (on Windows version)
6. provide reasonable fallback if console fails to initialize
7. reduce messy interleaved `version` declarations
8. remove ugly C++ conversion detritus
9. fix incorrect error checking of `GetConsoleScreenBufferInfo()`
10. use `fputs` instead of `fprintf` where possible
11. put initialization of console *after* command line processing, not before. Doing it before could mean it was turned on even if it failed.

### DIFF
--- a/src/ddmd/console.d
+++ b/src/ddmd/console.d
@@ -8,116 +8,206 @@
  * Source:      $(DMDSRC _console.d)
  */
 
+/********************************************
+ * Control the various text mode attributes, such as color, when writing text
+ * to the console.
+ */
+
 module ddmd.console;
 
 import core.stdc.stdio;
-import core.sys.posix.unistd;
-import core.sys.windows.windows;
+extern (C) int isatty(int);
 
-import core.stdc.string;
-import core.stdc.stdlib;
 
-version (Windows) extern (C) int isatty(int);
-
-enum COLOR : int
+enum Color : int
 {
-    COLOR_BLACK     = 0,
-    COLOR_RED       = 1,
-    COLOR_GREEN     = 2,
-    COLOR_BLUE      = 4,
-    COLOR_YELLOW    = COLOR_RED | COLOR_GREEN,
-    COLOR_MAGENTA   = COLOR_RED | COLOR_BLUE,
-    COLOR_CYAN      = COLOR_GREEN | COLOR_BLUE,
-    COLOR_WHITE     = COLOR_RED | COLOR_GREEN | COLOR_BLUE,
+    black     = 0,
+    red       = 1,
+    green     = 2,
+    blue      = 4,
+    yellow    = red | green,
+    magenta   = red | blue,
+    cyan      = green | blue,
+    white     = red | green | blue,
 }
 
-alias COLOR_BLACK = COLOR.COLOR_BLACK;
-alias COLOR_RED = COLOR.COLOR_RED;
-alias COLOR_GREEN = COLOR.COLOR_GREEN;
-alias COLOR_BLUE = COLOR.COLOR_BLUE;
-alias COLOR_YELLOW = COLOR.COLOR_YELLOW;
-alias COLOR_MAGENTA = COLOR.COLOR_MAGENTA;
-alias COLOR_CYAN = COLOR.COLOR_CYAN;
-alias COLOR_WHITE = COLOR.COLOR_WHITE;
-
-version (Windows)
+struct Console
 {
-    extern (C++) static WORD consoleAttributes(HANDLE h)
+    version (Windows)
     {
-        static __gshared CONSOLE_SCREEN_BUFFER_INFO sbi;
-        static __gshared bool sbi_inited = false;
-        if (!sbi_inited)
-            sbi_inited = GetConsoleScreenBufferInfo(h, &sbi) != FALSE;
-        return sbi.wAttributes;
-    }
+        import core.sys.windows.windows;
 
-    enum : int
-    {
-        FOREGROUND_WHITE = FOREGROUND_RED | FOREGROUND_GREEN | FOREGROUND_BLUE,
-    }
-}
+      private:
+        CONSOLE_SCREEN_BUFFER_INFO sbi;
+        HANDLE handle;
+        FILE* _fp;
 
-extern (C++) bool isConsoleColorSupported()
-{
-    version (CRuntime_DigitalMars)
-    {
-        return isatty(stderr._file) != 0;
-    }
-    else version (CRuntime_Microsoft)
-    {
-        return isatty(fileno(stderr)) != 0;
+      public:
+
+        @property FILE* fp() { return _fp; }
+
+        /*********************************
+         * Create an instance of Console connected to stream fp.
+         * Params:
+         *      fp = io stream
+         * Returns:
+         *      pointer to created Console
+         *      null if failed
+         */
+        static Console* create(FILE* fp)
+        {
+            /* Determine if stream fp is a console
+             */
+            version (CRuntime_DigitalMars)
+            {
+                if (!isatty(fp._file))
+                    return null;
+            }
+            else version (CRuntime_Microsoft)
+            {
+                if (!isatty(fileno(fp)))
+                    return null;
+            }
+            else
+            {
+                return null;
+            }
+
+            DWORD nStdHandle;
+            if (fp == stdout)
+                nStdHandle = STD_OUTPUT_HANDLE;
+            else if (fp == stderr)
+                nStdHandle = STD_ERROR_HANDLE;
+            else
+                return null;
+
+            auto h = GetStdHandle(nStdHandle);
+            CONSOLE_SCREEN_BUFFER_INFO sbi;
+            if (GetConsoleScreenBufferInfo(h, &sbi) == 0) // get initial state of console
+                return null;
+
+            auto c = new Console();
+            c._fp = fp;
+            c.handle = h;
+            c.sbi = sbi;
+            return c;
+        }
+
+        /*******************
+         * Turn on/off intensity.
+         * Params:
+         *      bright = turn it on
+         */
+        void setColorBright(bool bright)
+        {
+            SetConsoleTextAttribute(handle, sbi.wAttributes | (bright ? FOREGROUND_INTENSITY : 0));
+        }
+
+        /***************************
+         * Set color and intensity.
+         * Params:
+         *      color = the color
+         *      bright = turn intensity on
+         */
+        void setColor(Color color, bool bright)
+        {
+            enum FOREGROUND_WHITE = FOREGROUND_RED | FOREGROUND_GREEN | FOREGROUND_BLUE;
+            WORD attr = sbi.wAttributes;
+            attr = (attr & ~(FOREGROUND_WHITE | FOREGROUND_INTENSITY)) |
+                   ((color & Color.red)   ? FOREGROUND_RED   : 0) |
+                   ((color & Color.green) ? FOREGROUND_GREEN : 0) |
+                   ((color & Color.blue)  ? FOREGROUND_BLUE  : 0) |
+                   (bright ? FOREGROUND_INTENSITY : 0);
+            SetConsoleTextAttribute(handle, attr);
+        }
+
+        /******************
+         * Reset console attributes to what they were
+         * when create() was called.
+         */
+        void resetColor()
+        {
+            SetConsoleTextAttribute(handle, sbi.wAttributes);
+        }
     }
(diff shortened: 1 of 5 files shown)


CyberShadow:
Not that this is an important part of the code base or anything, so following due process isn't as important, though otherwise I would say:

1. You list 10 changes in the pull request description, but there is only one commit. Ideally, there would be one commit per change.

2. Also, pull request descriptions such as these should be in the commit message. If you put the full description in the commit message (and have only one commit), GitHub will automatically populate the pull request description accordingly, which saves having to enter a PR description again.

WalterBright:
> You list 10 changes in the pull request description, but there is only one commit. Ideally, there would be one commit per change.

They're interdependent, so that's a little impractical in this case.

CyberShadow:
> They're interdependent, so that's a little impractical in this case.

Only if you make 10 changes and then attempt to split them up into 10 commits. The generally-accepted wisdom is to commit-as-you-go, amending past commits as needed.

Investing in a better git UI might help. For example, magit allows selectively inserting a change into a commit made a while ago with a few keystrokes.

src/ddmd/console.d
```diff
@@ -8,116 +8,200 @@
...
-import core.sys.posix.unistd;
-import core.sys.windows.windows;
+extern (C) int isatty(int);
 
-import core.stdc.string;
-import core.stdc.stdlib;
-
-version (Windows) extern (C) int isatty(int);
 
 enum COLOR : int
```
andralex: s/COLOR/Color/g

src/ddmd/console.d
```diff
@@ -8,116 +8,200 @@
...
 {
-    COLOR_BLACK     = 0,
-    COLOR_RED       = 1,
-    COLOR_GREEN     = 2,
-    COLOR_BLUE      = 4,
-    COLOR_YELLOW    = COLOR_RED | COLOR_GREEN,
-    COLOR_MAGENTA   = COLOR_RED | COLOR_BLUE,
-    COLOR_CYAN      = COLOR_GREEN | COLOR_BLUE,
-    COLOR_WHITE     = COLOR_RED | COLOR_GREEN | COLOR_BLUE,
+    Black     = 0,
```
andralex: s/Black/black/g

etc

src/ddmd/console.d
```diff
@@ -8,116 +8,200 @@
...
+                nStdHandle = STD_ERROR_HANDLE;
+            else
+                return null;
+
+            auto h = GetStdHandle(nStdHandle);
+            CONSOLE_SCREEN_BUFFER_INFO sbi;
+            if (GetConsoleScreenBufferInfo(h, &sbi) == 0) // get initial state of console
+                return null;
+
+            Console* c = new Console();
```
andralex: auto

src/ddmd/console.d
```diff
@@ -8,116 +8,200 @@
...
-        attr = (attr & ~(FOREGROUND_WHITE | FOREGROUND_INTENSITY)) | ((color & COLOR_RED) ? FOREGROUND_RED : 0) | ((color & COLOR_GREEN) ? FOREGROUND_GREEN : 0) | ((color & COLOR_BLUE) ? FOREGROUND_BLUE : 0) | (bright ? FOREGROUND_INTENSITY : 0);
-        SetConsoleTextAttribute(h, attr);
-    }
-    else
-    {
-        fprintf(stderr, "\033[%d;%dm", bright ? 1 : 0, 30 + cast(int)color);
-    }
-}
+      public:
+        FILE* fp;
```
andralex: this seems to be shared across Windows and Posix

src/ddmd/console.d
```diff
@@ -8,116 +8,200 @@
...
-    version (Windows)
-    {
-        HANDLE h = GetStdHandle(STD_ERROR_HANDLE);
-        SetConsoleTextAttribute(h, consoleAttributes(h));
+        static Console* create(FILE* fp)
+        {
+            import core.stdc.stdlib : getenv;
+            const(char)* term = getenv("TERM");
+            import core.stdc.string : strcmp;
+            if (!(isatty(STDERR_FILENO) && term && term[0] && 0 != strcmp(term, "dumb")))
```
andralex: `&& term[0]` is superfluous, remove

src/ddmd/console.d
```diff
@@ -8,116 +8,200 @@
...
+                return null;
+
+            Console* c = new Console();
```
andralex: auto

src/ddmd/console.d
```diff
@@ -8,116 +8,200 @@
...
+            c.fp = fp;
+            return c;
+        }
+
+        void setColorBright(bool bright)
+        {
+            fprintf(fp, "\033[%dm", bright ? 1 : 0);
```
andralex: s/bright ? 1 : 0/bright

all bools are converted to long anyway

src/ddmd/console.d
```diff
@@ -8,116 +8,200 @@
...
+        }
+
+        void setColor(COLOR color, bool bright)
+        {
+            fprintf(fp, "\033[%d;%dm", bright ? 1 : 0, 30 + cast(int)color);
```
andralex: s/bright ? 1 : 0/bright

src/ddmd/errors.d
```diff
@@ -94,33 +94,36 @@ extern (C++) void deprecationSupplemental(const ref Loc loc, const(char)* format
 // Just print, doesn't care about gagging
 extern (C++) void verrorPrint(const ref Loc loc, COLOR headerColor, const(char)* header, const(char)* format, va_list ap, const(char)* p1 = null, const(char)* p2 = null)
 {
+    Console* con = cast(Console*)global.console;
     const p = loc.toChars();
-    if (global.params.color)
-        setConsoleColorBright(true);
+    if (con)
+        con.setColorBright(true);
```
andralex: The null object pattern comes to mind - you define an object that accepts the calls but does nothing. That way you don't need to test the pointer with every use.

src/ddmd/console.d
adamdruppe: I prefer being explicit here since it specifically MUST be 1 or 0 for this to work in the string.

src/ddmd/console.d
WalterBright: I don't trust getenv() to do it right.

src/ddmd/console.d
WalterBright: It's not shared across the default implementation.

src/ddmd/errors.d
WalterBright: There's more than that here. The null path leaves the backticks in.

src/ddmd/console.d
```diff
@@ -8,116 +8,206 @@
...
+            c.sbi = sbi;
+            return c;
+        }
+
+        /*******************
+         * Turn on/off intensity.
+         * Params:
+         *      bright = turn it on
+         */
+        void setColorBright(bool bright)
```
jacob-carlborg: Should this be name `colorBright` instead to allow to be used as a property, i.e. `console.colorBright = false`?

src/ddmd/console.d
```diff
@@ -8,116 +8,206 @@
...
+            SetConsoleTextAttribute(handle, sbi.wAttributes | (bright ? FOREGROUND_INTENSITY : 0));
+        }
+
+        /***************************
+         * Set color and intensity.
+         * Params:
+         *      color = the color
+         *      bright = turn intensity on
+         */
+        void setColor(Color color, bool bright)
```
jacob-carlborg: Wouldn't it be better to keep the current state of `bright` or provide a default value?

src/ddmd/console.d
```diff
@@ -8,116 +8,206 @@
...
-    version (CRuntime_DigitalMars)
-    {
-        return isatty(stderr._file) != 0;
-    }
-    else version (CRuntime_Microsoft)
-    {
-        return isatty(fileno(stderr)) != 0;
+      public:
+
+        @property FILE* fp() { return _fp; }
```
jacob-carlborg: Can we avoid abbreviations public symbols?

src/ddmd/mars.d
```diff
@@ -291,7 +291,7 @@ private int tryMain(size_t argc, const(char)** argv)
     files.reserve(arguments.dim - 1);
     // Set default values
     global.params.argv0 = arguments[0];
-    global.params.color = isConsoleColorSupported();
+    global.params.color = true;
```
ibuclaw: No way to turn this off? This seems like a dead field that should be removed.

src/ddmd/errors.d
```diff
@@ -92,35 +92,38 @@ extern (C++) void deprecationSupplemental(const ref Loc loc, const(char)* format
 }
 
 // Just print, doesn't care about gagging
-extern (C++) void verrorPrint(const ref Loc loc, COLOR headerColor, const(char)* header, const(char)* format, va_list ap, const(char)* p1 = null, const(char)* p2 = null)
+extern (C++) void verrorPrint(const ref Loc loc, Color headerColor, const(char)* header, const(char)* format, va_list ap, const(char)* p1 = null, const(char)* p2 = null)
```
ibuclaw: This function is private to errors.d -> remove extern(C++) linkage.

src/ddmd/console.d
andralex: Oh I had the condition backwards.

adamdruppe:
On Tue, May 16, 2017 at 01:13:14AM -0700, jacob-carlborg wrote:
> +        @property FILE* fp() { return _fp; }
> 
> Can we avoid abbreviations public symbols?

fp is one of those that isn't really an abbreviation in my mind
anymore because of how extremely common it is for exactly this
purpose. It is like `i` for a loop counter.

jacob-carlborg:
> fp is one of those that isn't really an abbreviation in my mind
anymore because of how extremely common it is for exactly this
purpose. It is like `i` for a loop counter.

I disagree.

src/ddmd/console.d
UplinkCoder: `0 != strcmp(term, "dumb")` should this be !strcmp(term, "dumb") ?

src/ddmd/mars.d
WalterBright: Look later on in the code, you'll find it.

src/ddmd/console.d
WalterBright: I prefer to not use properties to do things that do more than just set a value.

src/ddmd/console.d
WalterBright: The premise of these functions is they alter the default state; they do not build on the current state.

src/ddmd/console.d
WalterBright: Doesn't make any difference.

src/ddmd/console.d
jacob-carlborg: Ok.

src/ddmd/console.d
jacob-carlborg: Ok.

src/ddmd/console.d
UplinkCoder: readably ?
`literal != functioncall` is wired. And not use in the rest of the codebase.
`mars.d` uses `!strcmp` throughout.

src/ddmd/mars.d
ibuclaw: Ah, `-color=off`?

src/ddmd/console.d
UplinkCoder: I was wrong.
`mars.d` uses `strcmp(p + 1, "allinst") == 0`

andralex: Never cared for the `constant compare expression` idiom but not a deal breaker.

src/ddmd/errors.d
andralex: Null object pattern = a Console that is not null and accepts any call, doing nothing.

andralex (approved):
alrighty then

src/ddmd/errors.d
CyberShadow: I think what @WalterBright is saying is that in order to make use of the null object pattern, backtick handling would need to be moved to inside the `Console` interface, instead of being handled on the outside. Since the backticks are in the semantic domain of the compiler, not colored console output, it would require a second abstraction layer on top of `Console`, i.e. it's just not worth it.

PR #6795, opened 2017-05-16
Fix bugs when building with g++-7

ibuclaw:
First is a bug caught by `-Waggressive-loops` - it's pretty obvious that it's an out of bounds error.

Second, ~there seems to be a segfault when doing `esave = *e`, I have no idea why.  Seems like a g++ regression.  Using memcpy and things just work...~ building using g++7, I get the same bug as 6215, aligning memory to 16-bytes fixes the problem - particularly for copying `elem` which uses `movqda`. instruction.

### DIFF
--- a/src/ddmd/backend/evalu8.c
+++ b/src/ddmd/backend/evalu8.c
@@ -2209,7 +2209,7 @@ elem * evalu8(elem *e, goal_t goal)
             case TYllong4:
             case TYullong4:
                 for (int i = 0; i < 4; ++i)
-                    e->EV.Vullong2[i] = (targ_ullong)l1;
+                    e->EV.Vullong4[i] = (targ_ullong)l1;
                 break;
 
             default:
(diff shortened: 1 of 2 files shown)


ibuclaw:
The instruction where it segfaults is:
```
movdqa 0x40(%rbx),%xmm0
```
Something is very fishy about `movdqa`.

ibuclaw:
Ahh... you can't movdqa to/from unaligned memory addresses.  My guess is that it requires that the memory is aligned to 16 (both structs have an 8 byte alignment).  I'm now getting segfaults in other places.

I wonder if the misalignment is coming from gcc or dmd...

ibuclaw:
I'm getting the same as https://issues.dlang.org/show_bug.cgi?id=6215

ibuclaw:
@WalterBright - ready for review.  This should allow using gdc-7 in travisci.

WalterBright:
Auto-merge toggled on

PR #6803, opened 2017-05-17
denum.d: enable syntax highlighting of error messages

WalterBright:


test/fail_compilation/fail109.d
```diff
@@ -1,7 +1,7 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/fail109.d(12): Error: enum member fail109.Bool.Unknown initialization with (Bool.True + 1) causes overflow for type 'bool'
+fail_compilation/fail109.d(12): Error: enum member fail109.Bool.Unknown initialization with `Bool.True+1` causes overflow for type `bool`
```
adamdruppe: If D had error message codes, the test could just check for the correct code instead of an exact output string... it'd make changing these more robust since you could actually rely on the test and not be changing them so frequently.

We really should add them.

WalterBright: I had error codes in DMC++. It was a giant pain in other ways.

yebblies:
Why was it necessary to change error messages and refactor the test files?  I'd be much more comfortable merging this if it wasn't 90% unrelated changes.

WalterBright:
> Why was it necessary to change error messages

The syntax highlighting looks for text between backticks. This has become a de-facto standard in markup, and is used by Ddoc as well.

> and refactor the test files?

I merged a couple of them to reduce the number of files. The tests are the same.

yebblies:
> The syntax highlighting looks for text between backticks. This has become a de-facto standard in markup, and is used by Ddoc as well.

No I get that, but you also made unrelated edits to error messages for some reason.

> I merged a couple of them to reduce the number of files. The tests are the same.

Yeah, I get that too, but it didn't need to be in this PR and I assume is behind the codecov spam.  TBH I think the tradeoff is different with the fail_compilation tests and the would be better off in multiple files instead of combined together.  It's also nice to have the bugzilla number and filename match, instead of only having that information in the comments.

Also, jenkins is failing again.

WalterBright:
> but it didn't need to be in this PR

If I didn't do it in this PR it would never get done, because I wouldn't remember which files of the pile of them could be merged later.

> TBH I think the tradeoff is different with the fail_compilation tests and the would be better off in multiple files instead of combined together.

It would have been less work for me if they were grouped together more.

> It's also nice to have the bugzilla number and filename match, instead of only having that information in the comments.

`grep NNNN *` is more reliable and not any harder than `ls *NNNN*`

WalterBright:
> jenkins is failing again.

Working now.

test/fail_compilation/fail109.d
wilzbach: This probably belongs on the NG, but your comment made me curious. Would you mind elaborating a bit about the troubles resulting from using error codes?

WalterBright: Yes, it does belong in the n.g. not here. Please ask there.

andralex (approved):
LGTM, @yebblies let me know if there's any blocker.

PR #6769, opened 2017-05-11
refactor: convert cdxxx[] to CodeBuilder

WalterBright:
Switching the array is tedious, but repetitive. Rather than change all the functions at once, which is error-prone, I used a temporary macro CDXXX() to build wrappers. On a subsequent PR, I'll remove the wrappers and change the functions.

After that's done, we should see a performance increase in the code generator as it removes the quadratic performance of continually walking the linked list of instructions (making it linear).

PR #6804, opened 2017-05-17
Fix make clean to remove stale backend/cdxxx.c .

quickfur:
Otherwise it may cause compile errors when updating from older builds of dmd.

Followup to: #6769

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -384,6 +384,7 @@ endif
 clean:
 	rm -R $(GENERATED)
 	rm -f dmd $(idgen_output)
+	rm -f $(addprefix $D/backend/, $(optabgen_output))
 	@[ ! -d ${PGO_DIR} ] || echo You should issue manually: rm -rf ${PGO_DIR}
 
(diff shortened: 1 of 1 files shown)


src/posix.mak
```diff
@@ -384,6 +384,7 @@ endif
 clean:
 	rm -R $(GENERATED)
 	rm -f dmd $(idgen_output)
+	rm -f backend/cdxxx.c
```
WalterBright: Instead, add `cdxxx.c` to the definition of `idgen_output`

WalterBright: In fact, add:
```
elxxx.c cdxxx.c optab.c debtab.c fltables.c tytab.c
```

quickfur: Hmm, actually, there's already a `optabgen_output` defined. Just have to delete stuff found in it.

quickfur:
Updated.

andralex:
thx! these things can get really annoying

PR #5637, opened 2016-04-07
Pass index and length to range error function

adamdruppe:
This lets druntime add it to the error object to give a more useful message by default.

PR #6805, opened 2017-05-17
pass dims to range error runtime function for better exception messages

adamdruppe:
This is something that has bugged me for many, many years: the compiler knows what the out-of-bounds index was, but doesn't bother to tell us, leading to a bunch of wasted time trying to figure out what random input caused the error. At least knowing the index and length is a solid starting point in reproducing it in a debugger.

This PR is the dmd side of passing that information over to the runtime. I know it could still use a bit of polish (especially for associative arrays, I think something else entirely needs to be done for them, though passing all zeroes could be detected on the runtime side and change the message too - at least it wouldn't be any worse than we have now).

But, before spending the time on that: are you guy actually willing to accept this?

### DIFF
--- a/src/dmd/e2ir.d
+++ b/src/dmd/e2ir.d
@@ -2727,7 +2727,7 @@ elem *toElem(Expression e, IRState *irs)
                         c1 = el_bin(OPandand, TYint, c1, c2);
 
                         // Construct: (c1 || arrayBoundsError)
-                        auto ea = buildArrayBoundsError(irs, ae.loc);
+                        auto ea = buildArrayBoundsError(irs, ae.loc, el_copytree(elwr), el_copytree(eupr), el_copytree(enbytesx));
                         elem *eb = el_bin(OPoror,TYvoid,c1,ea);
                         einit = el_combine(einit, eb);
                     }
@@ -2823,7 +2823,7 @@ elem *toElem(Expression e, IRState *irs)
                             }
 
                             // Construct: (c || arrayBoundsError)
-                            echeck = el_bin(OPoror, TYvoid, c, buildArrayBoundsError(irs, ae.loc));
+                            echeck = el_bin(OPoror, TYvoid, c, buildArrayBoundsError(irs, ae.loc, el_long(TYsize_t, 0), el_copytree(eleny), el_copytree(elen)));
                         }
                         else
                         {
@@ -4818,12 +4818,12 @@ elem *toElem(Expression e, IRState *irs)
                     //  lwr <= upr
 
                     elem *c1 = null;
+                    elem *elen;
                     if (!se.upperIsInBounds)
                     {
                         eupr2 = el_same(&eupr);
                         eupr2.Ety = TYsize_t;  // make sure unsigned comparison
 
-                        elem *elen;
                         if (t1.ty == Tsarray)
                         {
                             TypeSArray tsa = cast(TypeSArray)t1;
@@ -4863,7 +4863,7 @@ elem *toElem(Expression e, IRState *irs)
                     if (c1)
                     {
                         // Construct: (c1 || arrayBoundsError)
-                        auto ea = buildArrayBoundsError(irs, se.loc);
+                        auto ea = buildArrayBoundsError(irs, se.loc, el_copytree(elwr2), el_copytree(eupr2), el_copytree(elen));
                         elem *eb = el_bin(OPoror, TYvoid, c1, ea);
 
                         elwr = el_combine(elwr, eb);
@@ -4966,7 +4966,7 @@ elem *toElem(Expression e, IRState *irs)
                     elem *n = el_same(&e);
 
                     // Construct: ((e || arrayBoundsError), n)
-                    auto ea = buildArrayBoundsError(irs, ie.loc);
+                    auto ea = buildArrayBoundsError(irs, ie.loc, null, null, null); // FIXME
                     e = el_bin(OPoror,TYvoid,e,ea);
                     e = el_bin(OPcomma, TYnptr, e, n);
                 }
@@ -5002,7 +5002,7 @@ elem *toElem(Expression e, IRState *irs)
                         n2x = el_bin(OPlt, TYint, n2x, elength);
 
                         // Construct: (n2x || arrayBoundsError)
-                        auto ea = buildArrayBoundsError(irs, ie.loc);
+                        auto ea = buildArrayBoundsError(irs, ie.loc, null, el_copytree(n2), el_copytree(elength));
                         eb = el_bin(OPoror,TYvoid,n2x,ea);
                     }
                 }
@@ -5937,10 +5937,13 @@ private elem *filelinefunction(IRState *irs, const ref Loc loc)
  * Params:
  *      irs = to get function from
  *      loc = to get file/line from
+ *      lwr = lower bound passed, if slice (array[lwr .. upr]). null otherwise.
+ *      upr = upper bound passed if slice (array[lwr .. upr]), index if not a slice (array[upr])
+ *      elength = length of array
  * Returns:
  *      elem generated
  */
-elem *buildArrayBoundsError(IRState *irs, const ref Loc loc)
+elem *buildArrayBoundsError(IRState *irs, const ref Loc loc, elem* lwr, elem* upr, elem* elength)
 {
     if (irs.params.checkAction == CHECKACTION.C)
     {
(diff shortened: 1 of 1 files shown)


src/ddmd/e2ir.d
```diff
@@ -4847,7 +4847,7 @@ elem *toElem(Expression e, IRState *irs)
                     if (c1)
                     {
                         // Construct: (c1 || arrayBoundsError)
-                        auto ea = buildArrayBoundsError(irs, se.loc);
+                        auto ea = buildArrayBoundsError(irs, se.loc, el_copytree(elwr2), el_copytree(eupr2), el_copytree(elen));
```
CyberShadow: I am no compiler expert but would `el_copytree` cause the expressions to be re-evaluated (incl. their side effects)? E.g. `int[1] a; int i = 5; a[++i] = 1;`

adamdruppe: Honestly, I have no idea. I know they copied other things in the function with that, and if I don't copy it, the compiler dies with an internal error when I try to use it... that's one of the reasons I opened the PR now, to get someone who knows what they're doing to weigh in.

WalterBright: @CyberShadow is correct.

adamdruppe: @WalterBright What's the right way to do it?

CyberShadow: I think you have to save the value to a temporary first.

WalterBright:
While I agree with the sentiment, this produces substantial runtime bloat, even if no range errors occur. Take a look at the generated code, and multiply it by all the places where arrays are accessed.

adamdruppe:
Building druntime debug on my computer with dmd before and after this patch resulted in a difference of about 8 KB in a 16 MB file, or a difference of ~0.05%. (obviously, the release build is unchanged). If it is a hot loop where you're counting your cache lines, you'll probably use the `.ptr` trick or something to selectively disable bounds checking anyway, and outside such places, having the info is worth the cost to me.

I haven't tried other programs, but you should - I implore you not to let your gut take the place of actual measurements. If you can prove to me that there actually is substantial runtime bloat, I'll close it (and keep the patch for my local dmd fork), but I insist on seeing reproducible numbers.

CyberShadow:
I could do some tests once you fix the side effects (otherwise it wouldn't be accurate of the end result).

I'm guessing range checking could theoretically be lowered to a Druntime templated function call... but that would probably really slow down compilation and require that the function is inlined.

adamdruppe:
On Thu, May 18, 2017 at 12:35:19PM -0700, Vladimir Panteleev wrote:
> I could do some tests once you fix the side effects (otherwise it wouldn't be accurate of the end result).

Yeah, I'll look at that later today... though I expect that will make the generated code a bit smaller anyway, since the bounds need to be loaded up for the comparison all we're really costing is a `push EAX; push EBX;` or whatever to pass the numbers to the function after the `dim < length` check...

WalterBright:
> I implore you not to let your gut take the place of actual measurements.

Adding *any* instructions to an inner loop is expensive - and that's where one finds array accesses.

WalterBright:
> all we're really costing

No. The array index and array length have to be 'live' after the check, this has consequences for register allocation. It's worse if the index and length have side effects and must be put into temporaries. Those cost 2 registers even if the bounds check passes.

WalterBright:
Generally, I don't want to make bounds checking expensive, as that will encourage people to turn it off (and those ignorant of compiler switches will declare D to be "worse code than C++"). Yes, I've seen the latter happen before.

wilzbach:
Thank you @adamdruppe for pushing this. Just a dumb question: this does only affect the debug builds anyways, right? And isn't the entire point of debug builds to be slower, but easier to debug (hence the name)?

adamdruppe:
On Fri, May 19, 2017 at 02:12:10AM -0700, Sebastian Wilzbach wrote:
> Just a dumb question: this does only affect the debug builds anyways, right?

Bounds checks are on unless you explicitly turn them off, which IMO you should NEVER do (with one rare exception, if you are doing runtime-less bare metal code in an extremely constrained environment).

I consider the `-release` and `-boundscheck` switches to dmd to be EXTREMELY harmful and it pains me any time I see someone recommending them for a slight edge in a benchmark... they are D's #1 memory safety feature (with #2 btw being the garbage collector!) and production/release builds are where they are MOST important - that's where they are actually protecting you against actual data!

So this change affects all builds except those horribly broken, memory-unsafe disasters waiting to happen.

But, you can selectively disable bounds checks in individual expressions when you are very careful using the `.ptr` property: http://arsdnet.net/this-week-in-d/dec-06.html

So when you are in an exceedingly tight loop and want max speed, you do it that way without throwing out the huge benefits the bounds checks give in the rest of the program. Of course, you also need to make this @trusted if you want to interact with @safe, so that's another reason to make sure you only do it where you actually need it and can be very careful in properly checking it.

You should only be doing that when you have verified that 1) there's actually a speed problem and 2) you aren't overstepping your bounds.

src/ddmd/e2ir.d
yebblies: There's a copytotemp function somewhere around.

WalterBright: Yes, but I suggest waiting on fixing that, as I am opposed to this PR on performance grounds.

wilzbach:
> Bounds checks are on unless you explicitly turn them off, which IMO you should NEVER do

My question was rather in the way: is there a chance we can pass this additional information on only for debug builds? AFAICT this would be a nice compromise and make everyone happy?

wilzbach:
> My question was rather in the way: is there a chance we can pass this additional information on only for debug builds? AFAICT this would be a nice compromise and make everyone happy?

Ping. AFAICT debugs builds can be slow, but should provide excellent _debug_ information to a developer (exactly what this PR does). In release mode, no performance penalty is paid.

(I also rebased the PR)

thewilsonator:
@WalterBright can you please reevaluate this PR in light of the merge of #7386 and #7392? @adamdruppe I take it that #5637 is redundant if this goes in?

adamdruppe:
I rebased and added ddoc to it too, hopefully it will pass the tests now and we can experiment a little.

BTW though remember, this just passes new arguments to the function. druntime will also need to be able to read those arguments and sink them to the user too. I don't recall if I did the separate druntime PR before or not.

CyberShadow:
That sounds like something that ought to be fixed before we can derive any meaningful results from tests, as duplicating the expressions could generate much more code than if they were saved to a temporary first.

adamdruppe:
BTW I am a bit surprised that it is passing all the tests if it is actually duplicating code.  I guess i should disassemble it with my own test case.

thewilsonator:
Thanks for rebasing. There are no open PRs by you in druntime (if you open one make sure to use the same brach name as this will cause it to use DMD + this PR for testing). 

If you could post an example of D + generated ASM of a loop with array access (with and without this PR) that would be great, assuming it doesn't add instructions to the loop I'm fine with merging this once we figure out whats the deal with copying stuff, (ideally any given function makes only a single call to the arrayBoundsError function, but I'll settle for not generating complete garbage) and the druntime PR is open and has tests.

thewilsonator:
Looks like we need to copy what is done for [`throw`](https://github.com/dlang/dmd/blob/593ee7b769ef6aa23746e7b4b88b3230afaf6935/src/dmd/s2ir.d#L974-L976), although I'm not sure how all the block stuff needs to be done. May need to convert the `(c || arrayBoundsError)` into a proper if/else to do that.

wilzbach:
> if you open one make sure to use the same brach name as this will cause it to use DMD + this PR for testing

Unfortunately that's only true for branches created at upstream at the moment.

adamdruppe:
well, i changed it to use el_same but the block change thing idk about yet.

adamdruppe:
lol the fails are build druntime "Error: variable _TMP1229 used before set" etc. So it is making a temporary and complaining about its own.

But with my original code all the tests passed so I'm not convinced it was wrong.

adamdruppe:
So, taking a closer look at the surrounding code, the reason my old code is working is because there are *already* calls to el_same or el_copytotemp earlier in the functions. By the time we get to the lines I am modifying, the potential side effects are already in the past.

So that explains why the tests are passing with copytree - the code is correct!

thewilsonator:
Good, so the next course of action is to closet this PR and open ones on upstream branches so that the druntime can be tested in sync with this, I'll do it later today.

thewilsonator:
#9015

PR #6807, opened 2017-05-18
fix Issue 17407 - [REG] __traits(compiles) triggers assertion failure

WalterBright:


test/compilable/interpret3.d
```diff
@@ -7719,3 +7719,15 @@ int[] f16094b()
 
 enum copy16094a = f16094a();
 enum copy16094b = f16094b();
+
+/**************************************************/
+// https://issues.dlang.org/show_bug.cgi?id=17407
+
+bool foo17404()
```
UplinkCoder: this should probably be foo17407 ?

UplinkCoder: @wa

src/ddmd/ctfeexpr.d
```diff
@@ -1804,7 +1804,7 @@ extern (C++) bool isCtfeValueValid(Expression newval)
     }
     if (newval.op == TOKnull)
     {
-        return tb.ty == Tnull || tb.ty == Tpointer || tb.ty == Tarray || tb.ty == Taarray || tb.ty == Tclass;
+        return tb.ty == Tnull || tb.ty == Tpointer || tb.ty == Tarray || tb.ty == Taarray || tb.ty == Tclass || tb.ty == Tdelegate;
```
UplinkCoder: You might want to include Tfunction here as well.

WalterBright: Tfunction is not a pointer, so that wouldn't work.

UplinkCoder: Ah right ... Tfunction is the actual function and not the function-ptr.

UplinkCoder: maybe we should do the same with delegate ?

mathias-lang-sociomantic:
Stable please

UplinkCoder:
@MartinNowak can you re-target this one to stable ?

wilzbach:
> @MartinNowak can you re-target this one to stable ?

Everyone with merge rights can do this - it's not that hard. Just do sth like `git rebase --onto upstream/stable`

UplinkCoder:
@wilzbach don't you have merge rights ?

MartinNowak:
@WalterBright, please learn to target stable when making bugfixes. Even if it's not much work, constantly having to rebase PRs is kind of tiresome.

PR #6810, opened 2017-05-19
add Classification enum

WalterBright:
Just makes it easier to see what the scheme is.

### DIFF
--- a/src/ddmd/errors.d
+++ b/src/ddmd/errors.d
@@ -19,6 +19,16 @@ import ddmd.root.outbuffer;
 import ddmd.root.rmem;
 import ddmd.console;
 
+/**********************
+ * Color highlighting to classify messages
+ */
+enum Classification
+{
+    error = Color.red,          /// for errors
+    gagged = Color.magenta,     /// for gagged errors
+    warning = Color.yellow,     /// for warnings
+    deprecation = Color.blue,  /// for deprecations
+}
 
 /**************************************
  * Print error message
@@ -144,7 +154,7 @@ extern (C++) void verror(const ref Loc loc, const(char)* format, va_list ap, con
     global.errors++;
     if (!global.gag)
     {
-        verrorPrint(loc, Color.red, header, format, ap, p1, p2);
+        verrorPrint(loc, Classification.error, header, format, ap, p1, p2);
         if (global.errorLimit && global.errors >= global.errorLimit)
             fatal(); // moderate blizzard of cascading messages
     }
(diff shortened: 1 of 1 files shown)


UplinkCoder:
You might as well put the in the color enum.
`Color.error = Color.red`,

src/ddmd/errors.d
```diff
@@ -19,6 +19,16 @@ import ddmd.root.outbuffer;
 import ddmd.root.rmem;
 import ddmd.console;
 
+/**********************
+ * Color highlighting to classify messages
+ */
+enum Classification
+{
+    error = Color.red,          // for errors
```
jacob-carlborg: Please use Ddoc comments.

src/ddmd/errors.d
```diff
@@ -19,6 +19,16 @@ import ddmd.root.outbuffer;
...
+    gagged = Color.magenta,     // for gagged errors
+    warning = Color.yellow,     // for warnings
+    _deprecation = Color.blue,  // for deprecations
```
ZombineDev: Nit: the convention (at least in Phobos) is `keyword_`, i.e. this should be `deprecation_`.

PetarKirov (approved):
LGTM, considering my and Jacob's comments are addressed.

src/ddmd/errors.d
wilzbach: @ZombineDev: how about adding this to the [D Style guide](http://dlang.org/dstyle.html)? The [Vibe.d](https://vibed.org/style-guide) already specifies this to some extent..

ZombineDev:
@wilzbach it's already there - search for "keywords" ;)

src/ddmd/errors.d
jacob-carlborg: I think the convention in DMD is the opposite, prefixing with underscore, unfortunately. Although I think that should be changed.

WalterBright:
> You might as well put the in the color enum.

Having usage details creep into the colorization module means it will no longer be encapsulated.

WalterBright:
ping @yebblies

yebblies (approved):
Why does `_deprecation` start with an `_` but the others don't?

jacob-carlborg:
> Why does _deprecation start with an _ but the others don't?

Confusion between `deprecated` (which is a keyword) and `deprecation`?

WalterBright:
> Why does _deprecation start with an _ but the others don't?

`deprecation` is a keyword and the others aren't. Look in `idgen.d` and there are several other examples of this.

jacob-carlborg:
> `deprecation` is a keyword and the others aren't. Look in `idgen.d` and there are several other examples of this.

@WalterBright `deprecation` is not a keyword [1], but `deprecated` is.

[1] https://github.com/dlang/dmd/blob/master/src/ddmd/access.d#L467

WalterBright:
Oops, yeah, you're right! Will fix.

PR #6813, opened 2017-05-19
fix Issue 16301 - CTFE execution of opApply keeps wrong "this" context in foreach's body

UplinkCoder:
Fixes issue 16301 by walking up the stack of previous this-pointers and looking for a match.
This will allow delegates to properly work at ctfe.

### DIFF
--- a/src/ddmd/dinterpret.d
+++ b/src/ddmd/dinterpret.d
@@ -2047,6 +2047,23 @@ public:
             assert(result.op == TOKstructliteral || result.op == TOKclassreference);
             return;
         }
+
+        // we probably have a nested delegate here.
+        // the thing to do is to iterate up the stack until we find
+        // a this that has a matching type.
+        // https://issues.dlang.org/showbug.cgi?id=16301
+        size_t stackOffset = ctfeStack.savedThis.dim;
+        while(stackOffset && (!result || !result.type.equals(e.type)))
+        {
+            result = ctfeStack.savedThis[--stackOffset];
+        }
+
+        if (result)
+        {
+            assert(result.op == TOKstructliteral || result.op == TOKclassreference);
+            return;
+        }
+
         e.error("value of 'this' is not known at compile time");
         result = CTFEExp.cantexp;
     }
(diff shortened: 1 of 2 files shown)


mathias-lang-sociomantic:
Please use a more descriptive title when submitting P.R., so that people watching this repo don't have to open the P.R. to know if it's of interest to them.
Also this lacks a test-case.

UplinkCoder:
@mathias-lang-sociomantic 
I added a testcase.

UplinkCoder:
@WalterBright 
I am not sure if the code in issue 16301 only works by accident.
Are delegates required to capture the nearest this pointer if there is no function-local context ?
It seems like an implementation detail ...

WalterBright:
When doing bugfix PRs, please have the subject match the bugzilla entry, prefixed with "fix". I already fixed this one.

wilzbach:
> When doing bugfix PRs, please have the subject match the bugzilla entry, prefixed with "fix". I already fixed this one.

Same for the commit messages...

WalterBright:
>  fix Issue 16301 [don't merge yet]

I added the [Needs Work] tag instead.

UplinkCoder:
@WalterBright I think the work is done.
You can remove the tag if you find the code agreeable.

yebblies:
There is almost certainly matching code in glue/e2ir/etc.  Can that be re-used somehow, or can the context be cached during semantic?  This will ensure the lookup has the same result at runtime and ctfe.

UplinkCoder:
@yebblies I had a look at e2ir and it add a negative offset to the stack-ptr, to walk up the stack-frames, we cannot do that in ctfe.
The code I have here comes closest to what e2ir does

yebblies:
But isn't that pretty much what we're doing here?  We do have a stack and we can index with a negative offset to our current frame...?

Your code here to walk up the scope chain and find the right 'this', which I assume e2ir also does, could that be moved into semantic somewhere?  Or does semantic already do this (for semantic checking) in which case can it save the result somewhere so we don't have to recompute?

UplinkCoder:
@yebblies the this we want to find is a "runtime" value.
therefore we cannot cannot cache this in semantic.
Maybe we could pre-compute how many stack frames we have to walk up to get at the variable but I am not sure if adding that complexity is warranted.

src/ddmd/dinterpret.d
```diff
@@ -2047,6 +2047,23 @@ public:
             assert(result.op == TOKstructliteral || result.op == TOKclassreference);
             return;
         }
+
+        // we probably have a nested delegate here.
+        // the thing to do is to iterate up the stack until we find
+        // a this that has a matching type.
+        // https://issues.dlang.org/showbug.cgi?id=16301
+        size_t stackOffset = ctfeStack.savedThis.dim;
+        while(stackOffset && (!result || !result.type.equals(e.type)))
```
EyalIO: Isn't this essentially implementing a dynamic scoping approach instead of a lexical scoping one?

If ctfeStack is the *call stack* - this sounds incorrect in the general case.

UplinkCoder: That's right, ctfeStack is the call stack.
However if a there is a closure it must have been created by a function call, therefore
The aggregate's this must be found and will be the same.
If you look at e2ir's closure handling, it's doing the same thing.
Lexical scope does not matter it just happens to be the same.

UplinkCoder:
@EyalIO has proven this unsound.
Back to the drawing board.

PR #6814, opened 2017-05-19
improve c++ mangle error message

UplinkCoder:
The error message now tells the user what is going wrong.

### DIFF
--- a/src/ddmd/cppmangle.d
+++ b/src/ddmd/cppmangle.d
@@ -563,7 +563,7 @@ public:
         }
         else
         {
-            t.error(Loc(), "Internal Compiler Error: unsupported type %s\n", t.toChars());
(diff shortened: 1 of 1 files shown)


UplinkCoder:
@WalterBright could you merge this ?

src/ddmd/cppmangle.d
```diff
@@ -563,7 +563,7 @@ public:
         }
         else
         {
-            t.error(Loc(), "Internal Compiler Error: unsupported type %s\n", t.toChars());
+            t.error(Loc(), "Internal Compiler Error: type %s can not be mapped to C++\n", t.toChars());
```
ZombineDev: Can you change ` %s ` to `` `%s` ``, so the error message could get syntax highlighting?

UplinkCoder: well it's an ice, but sure

PR #6815, opened 2017-05-19
fix Issue 17399 - core.checkedint.addu cannot inline function

WalterBright:
This could use a refactoring to get rid of the `goto`, but I wanted to keep the diff small. Refactorings should happen separately.

### DIFF
--- a/src/ddmd/inline.d
+++ b/src/ddmd/inline.d
@@ -281,7 +281,16 @@ public:
                   // 'return' must not leave in the expand result
             return;
         static if (asStatements)
-            result = new ReturnStatement(s.loc, exp);
+        {
+            /* Any return statement should be the last statement in the function being
+             * inlined, otherwise things shouldn't have gotten this far. Since the
+             * return value is being ignored (otherwise it wouldn't be inlined as a statement)
+             * we only need to evaluate `exp` for side effects.
+             * Already disallowed this if `exp` produces an object that needs destruction -
+             * an enhancement would be to do the destruction here.
+             */
+            result = new ExpStatement(s.loc, exp);
+        }
         else
             result = exp;
     }
@@ -1524,14 +1533,43 @@ bool canInline(FuncDeclaration fd, bool hasthis, bool hdrscan, bool statementsTo
         if (tf.varargs == 1)
             goto Lno;
 
+        /* No lazy parameters when inlining by statement, as the inliner tries to
+         * operate on the created delegate itself rather than the return value.
+         * Discussion: https://github.com/dlang/dmd/pull/6815
+         */
+        if (statementsToo && fd.parameters)
+        {
+            foreach (param; *fd.parameters)
+            {
+                if (param.storage_class & STClazy)
+                    goto Lno;
+            }
+        }
+
+        static bool hasDtor(Type t)
+        {
+            auto tv = t.baseElemOf();
+            return tv.ty == Tstruct || tv.ty == Tclass; // for now assume these might have a destructor
+        }
+
         /* Don't inline a function that returns non-void, but has
          * no return expression.
-         * No statement inlining for non-voids.
+         * When inlining as a statement:
+         * 1. don't inline array operations, because the order the arguments
+         *    get evaluated gets reversed. This is the same issue that e2ir.callfunc()
+         *    has with them
+         * 2. don't inline when the return value has a destructor, as it doesn't
+         *    get handled properly
          */
         if (tf.next && tf.next.ty != Tvoid &&
-            (!(fd.hasReturnExp & 1) || statementsToo) &&
+            (!(fd.hasReturnExp & 1) ||
+             statementsToo && (fd.isArrayOp || hasDtor(tf.next))) &&
             !hdrscan)
         {
+            static if (CANINLINE_LOG)
+            {
+                printf("\t3: no %s\n", fd.toChars());
+            }
             goto Lno;
         }
 
@@ -1560,6 +1598,10 @@ bool canInline(FuncDeclaration fd, bool hasthis, bool hdrscan, bool statementsTo
                      fd.hasNestedFrameRefs() ||
                      (fd.isVirtual() && !fd.isFinalFunc())))
     {
+        static if (CANINLINE_LOG)
+        {
+            printf("\t4: no %s\n", fd.toChars());
+        }
         goto Lno;
     }
 
--- a/test/compilable/extra-files/c6815.d
+++ b/test/compilable/extra-files/c6815.d
@@ -1,0 +1,5 @@
+void c()
+{
+    import d6815 : d;
+    d();
+}
--- a/test/compilable/extra-files/d6815.d
+++ b/test/compilable/extra-files/d6815.d
@@ -1,0 +1,5 @@
+void d()
+{
+    import e6815 : e;
+    static if (e("")) { }
+}
--- a/test/compilable/extra-files/e6815.d
+++ b/test/compilable/extra-files/e6815.d
@@ -1,0 +1,7 @@
+bool e(T)(T)
+{
+    f(true);
+    return true;
+}
+
+void f(lazy bool) {}
--- a/test/compilable/pull6815.d
+++ b/test/compilable/pull6815.d
@@ -1,0 +1,8 @@
+/* REQUIRED_ARGS: -inline -Icompilable/extra-files
+ */
+
+void b()
+{
+    import e6815 : e;
+    e("");
+}
--- a/test/compilable/test17399.d
+++ b/test/compilable/test17399.d
@@ -1,0 +1,18 @@
+/* REQUIRED_ARGS: -inline
+ */
+
+// https://issues.dlang.org/show_bug.cgi?id=17399
+
+pragma(inline, true)
+uint addu(uint x, uint y, ref bool overflow) {
+    uint r = x + y;
+    if (r < x || r < y)
+        overflow = true;
+    return r;
+}
+
+void foo() {
+    uint a, b;
+    bool over;
+    addu(a, b, over);
+}


UplinkCoder:
Oh  I was about to work on a fix for this myself :)
could you explain why have to rescan ?
what the bug was in the first place ?

WalterBright:
My fix doesn't work anyway, so back to the drawing board.

UplinkCoder:
@WalterBright seems like the reason why it did not inline was because the backend can't handle it.

UplinkCoder:
@WalterBright regardless if it worked, what was it supposed to do ?

UplinkCoder:
@WalterBright 
The reason it cannot inline the function, is because at the callside the return value is discarded.

UplinkCoder:
To do this inlining we need to remove the return expression when inling as statements.

UplinkCoder:
thanks for the comment

src/ddmd/inline.d
```diff
@@ -1524,14 +1533,39 @@ bool canInline(FuncDeclaration fd, bool hasthis, bool hdrscan, bool statementsTo
...
+         * 1. don't inline array operations, because the order the arguments
+         *    get evaluated gets reversed. This is the same issue that e2ir.callfunc()
+         *    has with them
+         * 2. don't inline when the return value has a destructor, as it doesn't
+         *    get handled properly
          */
         if (tf.next && tf.next.ty != Tvoid &&
-            (!(fd.hasReturnExp & 1) || statementsToo) &&
+            (!(fd.hasReturnExp & 1) ||
+             statementsToo && (isArrayOperationFunction(fd)|| hasDtor(tf.next))) &&
```
UplinkCoder: `isArrayOperationFunction(fd)|| hasDtor(tf.next)` => `isArrayOperationFunction(fd) || hasDtor(tf.next)`

UplinkCoder (commented):
you missed a space

UplinkCoder:
@WalterBright seems to break something in vibe.d

src/ddmd/inline.d
```diff
@@ -1524,14 +1533,39 @@ bool canInline(FuncDeclaration fd, bool hasthis, bool hdrscan, bool statementsTo
         if (tf.varargs == 1)
             goto Lno;
 
+        static bool isArrayOperationFunction(FuncDeclaration fd)
+        {
+            auto p = fd.ident.toChars();
+            if (memcmp(p, "_array".ptr, 6) == 0) // better to set a bit in the FuncDeclaration than this
```
yebblies: IIRC there is already a member in `FuncDeclaration` for this.

WalterBright: What's its name?

UplinkCoder: isArrayOp ?

WalterBright: that's it, thanks!

WalterBright:
This is failing vibe.d with:
```
[vibe.d] Running shell script
+ DC=dmd VIBED_DRIVER=libasync BUILD_EXAMPLE=0 RUN_TEST=0 ./travis-ci.sh
+ dub build --combined -b release --compiler=dmd --config=libasync
Performing "release" build using dmd for x86_64.
vibe-d 0.7.31-rc.2: building configuration "libasync"...
source/vibe/core/drivers/libasync.d(33,8): Deprecation: module std.c.stdio is deprecated - Import core.stdc.stdio instead
Enhanced memory security is enabled.
Memory debugger enabled
Using Linux EPOLL for events
source/vibe/http/server.d(286,33): Deprecation: alias diet.traits.FilterCallback is deprecated - Use SafeFilterCallback instead.
source/vibe/core/drivers/libasync.d(506,16): Deprecation: catch statement without an exception specification is deprecated; use catch(Throwable) for old behavior
/var/lib/jenkins/dlang_projects/distribution/bin/../imports/std/exception.d(554,3): Error: cannot cast __dgliteral4 to object.Throwable at compile time
/var/lib/jenkins/dlang_projects/distribution/bin/../imports/std/experimental/logger/core.d(147,13):        called from here: format("\x0a        static if (__traits(compiles, {import %1$s : logLevel;}))\x0a        {\x0a            import %1$s : logLevel;\x0a            static assert(is(typeof(logLevel) : LogLevel),\x0a                          \"Expect 'logLevel' to be of Type 'LogLevel'.\");\x0a            // don't enforce enum here\x0a            alias moduleLogLevel = logLevel;\x0a        }\x0a        else\x0a            // use logLevel of package or default\x0a            alias moduleLogLevel = moduleLogLevel!(parentOf(moduleName));\x0a    ", "libasync.threads_loggerconfig")
../../.dub/packages/libasync-0.8.3/libasync/source/libasync/internals/logging.d(18,50): Error: template instance std.experimental.logger.core.moduleLogLevel!"libasync.threads" error instantiating
../../.dub/packages/libasync-0.8.3/libasync/source/libasync/threads.d(34,11):        instantiated from here: defaultLogFunction!(34, "../../.dub/packages/libasync-0.8.3/libasync/source/libasync/threads.d", "libasync.threads.destroyAsyncThreads", "void libasync.threads.destroyAsyncThreads() nothrow", "libasync.threads", string, string)
dmd failed with exit code 1.
script returned exit code 2
```
@s-ludwig any ideas?

s-ludwig:
Appears to be triggered by some code in libasync/std.experimental.logger, of which I don't have any deeper knowledge. It seems strange that this is only triggered by libasync, though, considering that the error message indicates that the error is in [`enforce`](https://github.com/dlang/phobos/blob/90afbc03ba023a98672a6ff4f55881369996766e/std/exception.d#L554-L556)

WalterBright:
Who wrote libasync?

s-ludwig:
That's @etcimon.

etcimon:
Version 0.8.3 of libasync introduced the experimental std logger which was written by another @calrama

src/ddmd/inline.d
yebblies: That's the one!

We really need to fix the hacks surrounding array ops' evaluation ordering.  They absolutely should be inlined when it makes sense.

WalterBright:
@calrama can you have a look at this problem, please?

etcimon:
It looks like it's trying to execute the `enforce` in the comment "// don't enforce enum here"

etcimon:
I think it triggers in libasync because that's the only place where `defaultLogFunction` is covered.

Calrama:
@WalterBright Sure, but reading through this thread I don't know how to exactly reproduce the problem. For reference, what I did back then was copy the std.experimental.logger implementation into libasync and add @trusted to them so @safe function could produce logging output in debug mode.

UplinkCoder:
This issue has todo with a return being replaced by an expression.
@Calrama this will only trigger when compiled with dmd ~master -inline
@WalterBright I recommend taking the output of -vcg-ast before and after this pr and diff them

Calrama:
@UplinkCoder: I have set up `dmd ~master` with `druntime ~master` and `phobos 2.074.0`, using `vibe.d v0.8.0-beta.5` and `libasync 0.8.3 ` and used @WalterBright command line to build vibe.d that resulted in an error for him:
`dub build --combined --build=release --compiler=/path/to/install/linux/bin64/dmd --verbose --config=libasync`, which compiles fine for me, so I don't know how to produce the error.

UplinkCoder:
@Calrama this bug manifests only if you clone dmd from `https://github.com/WalterBright/dmd`
checkout `fix17399` and build with the generated dmd
I am debugging this now, since i have a good idea what is going wrong

UplinkCoder:
@WalterBright sorry for the noise ...
of course -vcg-ast won't work on errors...

Calrama:
@UplinkCoder Ok, thanks to you I was able to reproduce the error and as far as I can tell the issue occurs with any call to format with at least two arguments, e.g. even
```d
mixin(format("%d", 4));
```
triggers the same compilation error. As such, I am not sure this is something to fix on libasync's end?

UplinkCoder:
@Calrama lovely.
That gives me a proper starting point in reducing this.

UplinkCoder:
@Calrama I connot reproduce it with ` import std.format; uint a = mixin(format("%d", 4)); `

UplinkCoder:
@WalterBright the issue seems to be cause by a lazy parameter which is rewritten to a delegate, said delegate depends on being a return and cannot be changed to StatementExp,

MartinNowak:
The arduous task of inlining in the frontend...

WalterBright:
> the issue seems to be cause by a lazy parameter which is rewritten to a delegate, said delegate depends on being a return and cannot be changed to ExpStatement

Thanks! Can you express this problem as code, please?

UplinkCoder:
I am still working on finding a minimal test-case.

Calrama:
> Thanks! Can you express this problem as code, please?

After collaborating with @UplinkCoder, I tried reducing this myself, but whenever I reduce, the error disappears. The original code is this:

[std.format.format](https://github.com/dlang/phobos/blob/v2.074.0/std/format.d#L5531) calls
```d
enforce(n == args.length, new FormatException(text("Orphan format arguments: args[", n, "..", args.length, "]")));
```

which is defined as
[std.exception.enforce](https://github.com/dlang/phobos/blob/master/std/exception.d#L554)
```d
T enforce(T)(T value, lazy Throwable ex)
```

After the `new FormatException([...])` expression is turned into a delegate returning it that delegate is then optimized [as a statement here](https://github.com/dlang/dmd/pull/6815/commits/db789c995863ac6cf7669148aaeb40f7ac2eb591#diff-80b3197d11a2644245c0b409d54ebe25L284). Reading your PR it seems that this should not have happened (and is a bug) and your PR then seems to enable this bug's consequences [by throwing the delegate's return value away](https://github.com/dlang/dmd/pull/6815/commits/db789c995863ac6cf7669148aaeb40f7ac2eb591#diff-80b3197d11a2644245c0b409d54ebe25R292); and then, since the compiler then tries to convert a delegate without return value (but with an expression inside that gets evaluates but then thrown away) into a Throwable, it errors out.

CyberShadow:
> I tried reducing this myself, but whenever I reduce, the error disappears

Not sure if this applies here, but in case it does, and if you haven't yet, try [DustMite](https://github.com/CyberShadow/DustMite), a tool designed to solve this exact problem.

Calrama:
> Not sure if this applies here, but in case it does, and if you haven't yet, try DustMite, a tool designed to solve this exact problem.

Thanks, I am running that right now, but it's probably going to take a while (50% of the 426042 LOC removed after about 6 hours).

wilzbach:
>  I am running that right now, but it's probably going to take a while (50% of the 426042 LOC removed after about 6 hours).

... did you use the parallel mode?

Calrama:
> ... did you use the parallel mode?

If you are referring to dub's parallel mode, no, because dustmite's example used dmd directly and so I did, too. I am not aware of dmd itself, or dustmite, having a parallel mode, but I have uploaded the test case reduction start point [here](https://github.com/Calrama/dmd-reduce), which is started as `./reduce.sh /path/to/dmd` in case you want to run it.

wilzbach:
>  I am not aware of dmd itself, or dustmite, having a parallel mode

`dustmite` does support running in multiple processes, there's `-j` since [last May](https://github.com/CyberShadow/DustMite/commit/7e76bb9aa56fc377afbf34769c0904f9098467f5).

>   -j[N]              Use N look-ahead processes (8 by default)

Calrama:
> dustmite does support running in multiple processes, there's -j since last May.
> 
>     -j[N] Use N look-ahead processes (8 by default)

Does that truly create parallel jobs? Because that number seems to automatically match the logic CPU threads available, but `htop` shows only a single one in significant use. Regardless, I have not set that option to anything else, so the answer to your question would then be "yes", since that seems to be dustmite's default.

UplinkCoder:
@WalterBright as a workaround we can flag every function with lazy parameters as non-inlineable and abort the inline for that subtree.
Thoughts ?

UplinkCoder:
I also determined that the bug is trigged by inlineing of `return value` in a `std.exception.enforce!bool` instantiation.

UplinkCoder:
@WalterBright 
inserting the following at line `1473` in `inline.d` should solve the problem

```
    foreach(p;*fd.parameters)
    {
        if (p.storage_class & STClazy)
            return false;
    }
```

Calrama:
@WalterBright The reduced test case with 18 lines of code is available [here](https://github.com/Calrama/dmd-reduce)
@UplinkCoder I can confirm that your patch makes the reduced test case, as well as vibe.d as a whole, compile for me.

UplinkCoder:
@Calrama unfortunately you need to reduce the Phobos source of std.experimental.logger as well.

WalterBright:
For reference, here is @Calrama 's reduction:
```
--- a.d ---
import std.string : format;

--- b.d ---
void b()
{
    import std.string : format;
    format("%s", "");
}

--- c.d ---
void c()
{
    import d : d;
    d();
}

--- d.d ---
void d()
{
    import e : e;
    e();
}

--- e.d ---
import std.experimental.logger : defaultLogFunction, LogLevel;
alias e = defaultLogFunction!(LogLevel.critical);

---------

dmd -lib -release -inline a b c
std\exception.d(554): Error: cannot cast __dgliteral4 to object.Throwable at compile time
std\experimental\logger\core.d(147):        called from here: format("\x0a        static if (__traits
(compiles, {import %1$s : logLevel;}))\x0a        {\x0a            import %1$s : logLevel;\x0a            static assert(
is(typeof(logLevel) : LogLevel),\x0a                          \"Expect 'logLevel' to be of Type 'LogLevel'.\");\x0a
       // don't enforce enum here\x0a            alias moduleLogLevel = logLevel;\x0a        }\x0a        else\x0a
      // use logLevel of package or default\x0a            alias moduleLogLevel = moduleLogLevel!(parentOf(moduleName));
\x0a    ", "d_loggerconfig")
std\experimental\logger\core.d(482): Error: template instance std.experimental.logger.core.moduleLogLevel!"d" error instantiating
d.d(4):        instantiated from here: defaultLogFunction!(4, "d.d", "d.d", "void d.d()", "d")
```

WalterBright:
Some experimentation shows that `a.d` can be omitted.

UplinkCoder:
`b.d` `c.d` and `d.d` call also be omitted.

WalterBright:
> b.d c.d and d.d call also be omitted.

That leaves nothing on the command line. ?!?

WalterBright:
I added the patch to the code. Thank you all! But I'd still like a reduced test case I can add to the test suite. Can you please continue with dustmiting?

src/ddmd/inline.d
```diff
@@ -1524,14 +1533,43 @@ bool canInline(FuncDeclaration fd, bool hasthis, bool hdrscan, bool statementsTo
...
+        if (statementsToo && fd.parameters)
+        {
+            foreach (param; *fd.parameters)
+            {
+                if (param.storage_class & STClazy)
+                    goto Lno;
+            }
+        }
+
+        static bool hasDtor(Type t)
```
UplinkCoder: that's very conservative :o)

UplinkCoder (approved):
Accepted.

UplinkCoder:
@WalterBright fails in std.concurrency a restart should get it to pass.

Calrama:
@UplinkCoder You are right. [It is done](https://github.com/Calrama/dmd-reduce/tree/5ba9897b552bde4e5921da537b42b1470ea25885).
@WalterBright I have reduced it now further, but with phobos running at over 1 million LoC dustmite would have taken several days to run for me (took 24 hours for me for the initial ~500k LoC above in parallel mode), so this one I did by hand:

```d
--- b.d ---
void b()
{
    import e : e;
    e("");
}

--- c.d ---
void c()
{
    import d : d;
    d();
}

--- d.d ---
void d()
{
    import e : e;
    static if (e("")) {}
}

--- e.d ---
bool e(T)(T)
{
    f(true);
    return true;
}

void f(lazy bool) {}
```

```
$ dmd -lib -release -inline b c
e.d(7): Error: cannot cast __dgliteral2 to bool at compile time
d.d(4):        called from here: e("")
```

WalterBright:
Awesome work, @Calrama and @UplinkCoder !

WalterBright:
Added the test case.

UplinkCoder:
@WalterBright feel free to merge.

WalterBright:
ping @yebblies

rainers:
This introduced a regression: https://issues.dlang.org/show_bug.cgi?id=17676

rainers:
This PR also caused compilation of https://github.com/dlang/visuald/blob/master/c2d/idl2d.d with "-c -inline -O" to take 2 min 12 s instead of 1.9 s before. Not really the fault of better inlining, but a hint to be careful generating larger functions and passing them to the optimizer.

MoonlightSentinel:
Another one: https://issues.dlang.org/show_bug.cgi?id=20981

PR #6817, opened 2017-05-21
test: Add lexer CRLF tests

CyberShadow:
https://github.com/dlang/dmd/pull/6793#issuecomment-302917914

This should cover all non-error handling of `'\r'` in `lexer.d`.

### DIFF
--- a/test/compilable/crlf.sh
+++ b/test/compilable/crlf.sh
@@ -1,0 +1,56 @@
+#!/usr/bin/env bash
+
+# Test CRLF and mixed line ending handling in D lexer.
+
+name=`basename $0 .sh`
+dir=${RESULTS_DIR}/compilable
+fn=${dir}/${name}.d
+
+printf '%s\r\n' \
+       '#!/usr/bin/env dmd -run' \
+       '' \
+       '#line 4' \
+       '' \
+       'void main()' \
+       '{' \
+       '}' \
+       '' \
+       '// single-line comment' \
+       '' \
+       '/*' \
+       '  multi-line comment' \
+       '*/' \
+       '' \
+       '/+' \
+       '  nested comment' \
+       '+/' \
+       '' \
+       '/**' \
+       '  doc comment' \
+       '*/' \
+       'void documentee() {}' \
+       > ${fn}
+
+printf '// mixed\n// line\n// endings\n' >> ${fn}
+printf 'void fun()\n{\n}\n' >> ${fn}
+
+printf 'enum str = "\r\nfoo\r\nbar\nbaz\r\n";\n' >> ${fn}
+printf 'static assert(str == "%s");\n' '\nfoo\nbar\nbaz\n' >> ${fn}
+
+printf 'enum bstr = `\r\nfoo\r\nbar\nbaz\r\n`;\n' >> ${fn}
+printf 'static assert(bstr == "%s");\n' '\nfoo\nbar\nbaz\n' >> ${fn}
+
+printf 'enum wstr = q"EOF\r\nfoo\r\nbar\nbaz\r\nEOF";\n' >> ${fn}
+printf 'static assert(wstr == "%s");\n' 'foo\nbar\nbaz\n' >> ${fn}
+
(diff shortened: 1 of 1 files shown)


yebblies:
Please just make it a `.d` file.  Tools screwing around with the line endings can be fixed if/when it shows up in PRs.

CyberShadow:
> Tools screwing around with the line endings can be fixed if/when it shows up in PRs.

It won't show up as a test failure, you can't detect the line ending of the source file from within the test.

And if you mean show up in just the diff - that's not what I saw happen in practice. This entire thing started because of an unrelated newline conversion.

ibuclaw:
Yes, see my comment. https://github.com/dlang/dmd/pull/6793#issuecomment-302919757

(I should probably switch away from using my phone, as I seem quite slow here).

CyberShadow:
> Please just make it a .d file.

Another reason why this is a bad idea is that not all viewers/editors will handle this correctly. The editor would need to detect that the file uses mixed line endings, and explicitly display carriage return characters and allow editing of those.

rikkimax:
If this is meant to test the lexer, why not use a string mixin inside a D file?

yebblies:
Why would anyone be editing/viewing this file?

I really hate the shell tests, which admittedly is because I usually develop on windows.  This is essentially just obfuscating the test in the hope that nobody messes it up.

CyberShadow:
> Why would anyone be editing/viewing this file?

One example: one of the covered features is hex literals. Pretty sure I've heard they're on the chopping block.

> This is essentially just obfuscating the test in the hope that nobody messes it up.

That's exactly right, and IMO it's a good thing. We don't have a test suite for our test suite so we know when someone accidentally breaks a test, and relying that nothing converts line endings seems pretty fragile.

> If this is meant to test the lexer, why not use a string mixin inside a D file?

Assuming it works, that's a good idea.

CyberShadow:
> If this is meant to test the lexer, why not use a string mixin inside a D file?

You can't document a symbol inside a string mixin, so doing it that way would mean not testing the DDoc code.

yebblies:
> That's exactly right, and IMO it's a good thing. We don't have a test suite for our test suite so we know when someone accidentally breaks a test, and relying that nothing converts line endings seems pretty fragile.

We have pull request review.

CyberShadow:
> I really hate the shell tests, which admittedly is because I usually develop on windows.

BTW (shameless plug): Digger can run the DMD test suite on Windows, too. It will take care of any necessary dependencies (MSYS) as well.

CyberShadow:
> We have pull request review.

If that worked, #6793 wouldn't have been merged, would it? :)

ibuclaw:
As per comment in other PR, put a loud comment at the top of the file to warn reviewers of the file contents.  I'm pretty sure that helloUTF16.d and others bare this too.

CyberShadow:
> Please just make it a .d file.

Well, I'm still pretty sure it's a bad idea, but I'm not going to argue with *two* compiler hackers. Done.

CyberShadow:
Wait, how do I make it generate documentation? Do I have to write another test?

test/compilable/crlf.sh
```diff
@@ -0,0 +1,54 @@
+#!/usr/bin/env bash
+
```
WalterBright: An embedded comment saying what this does would be nice.

CyberShadow: Done. Didn't add one initially because none of the existing `.sh` files had one.

WalterBright: That one line makes all the difference!

WalterBright:
Auto-merge toggled on

CyberShadow:
Here is the coverage diff BTW:
https://codecov.io/gh/dlang/dmd/compare/01e29d4ba2658882fe63063e13a4095e37209a7d...a69fe6dec119c20055fcb3da0b0384276cf32d6a/changes

wilzbach:
> Here is the coverage diff

That's weird ... at some point during DConf the integration was also displayed in the GH status report. Does it maybe wait for the auto-tester to finish before reporting?

CyberShadow:
I saw it too but I don't know what causes it to appear. Maybe when the coverage drops?

PR #6818, opened 2017-05-21
replace hack with isArrayOp

WalterBright:


UplinkCoder (approved):
Approved can be merged after the tests pass on atleast one platform.

dnadlinger (approved):

WalterBright:
Auto-merge toggled on

wilzbach:
> Was my auto-merge not good enough for you? ;o)

I think @WalterBright isn't familiar with the "auto-merge" labels yet.
@WalterBright since a couple of months the "auto-merge" label has superseded the merging done by the auto-tester. The main reason for this was that the Auto-tester doesn't take the other CIs into account and sometimes blocked because of this. So toggling a PR for auto-merge is as simple as adding the label... (like @klickverbot did)

WalterBright:
Sorry, I didn't know the automerge label was more than just a label.

yebblies:
Does that mean I can use Brad's auto-merge to force a merge when jenkins is broken?

JackStouffer:
@yebblies Yes

CyberShadow:
> Does that mean I can use Brad's auto-merge to force a merge when jenkins is broken?

If it's broken (and you have permissions), just go ahead and unset it as required.

I'd even go as far as hiding it completely, but that's something we should discuss / work on.

yebblies:
> If it's broken (and you have permissions), just go ahead and unset it as required.

How/where do I do that?

CyberShadow:
https://github.com/dlang/dmd/settings/branches

yebblies:
Yeah I don't have permissions for that.

CyberShadow:
OK, I did it.

@MartinNowak One more thing for you to look at once you get back.

PR #6819, opened 2017-05-21
cgxmm.c: switch to CodeBuilder interface

WalterBright:


UplinkCoder:
I like this.
it looks more like newCTFE.

yebblies:
Auto-merge toggled on

PR #6822, opened 2017-05-22
fix Issue 17419 - add __traits(getLinkage, s) to the the linkage of s…

WalterBright:
…ymbol s

The use of name mangling in std.traits is slow, unreliable and brittle. Should be replaced with __traits(). This is a start.

### DIFF
--- a/src/ddmd/traits.d
+++ b/src/ddmd/traits.d
@@ -82,6 +82,7 @@ static this()
         "identifier",
         "getProtection",
         "parent",
+        "getLinkage",
         "getMember",
         "getOverloads",
         "getVirtualFunctions",
@@ -909,6 +910,35 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
 
         auto tup = new TupleExp(e.loc, mods);
         return tup.semantic(sc);
+    }
+    if (e.ident == Id.getLinkage)
+    {
+        // get symbol linkage as a string
+        if (dim != 1)
+            return dimError(1);
+
+        auto o = (*e.args)[0];
+        auto s = getDsymbol(o);
+        Declaration d;
+        if (!s || (d = s.isDeclaration()) is null)
+        {
+            e.error("argument to `__traits(getLinkage, %s)` is not a declaration", o.toChars());
+            return new ErrorExp();
+        }
+        string linkage;
+        switch (d.linkage)
+        {
+            case LINK.d:        linkage = "D";           break;
+            case LINK.c:        linkage = "C";           break;
+            case LINK.cpp:      linkage = "C++";         break;
+            case LINK.windows:  linkage = "Windows";     break;
+            case LINK.pascal:   linkage = "Pascal";      break;
+            case LINK.objc:     linkage = "Objective-C"; break;
+            case LINK.system:   linkage = "System";      break;
+            default: assert(0);
+        }
+        auto se = new StringExp(e.loc, cast(char*)linkage.ptr);
+        return se.semantic(sc);
     }
     if (e.ident == Id.allMembers ||
         e.ident == Id.derivedMembers)
--- a/test/compilable/test17419.d
+++ b/test/compilable/test17419.d
@@ -1,0 +1,29 @@
+// https://issues.dlang.org/show_bug.cgi?id=17419
+
+
+extern (C) int fooc();
+alias aliasc = fooc;
+
+static assert(__traits(getLinkage, fooc) == "C");
+static assert(__traits(getLinkage, aliasc) == "C");
+
+extern (D) int food();
+extern (C++) int foocpp();
+extern (Windows) int foow();
+extern (Pascal) int foop();
+extern (Objective-C) int fooobjc();
+extern (System) int foos();
+
+static assert(__traits(getLinkage, food) == "D");
+static assert(__traits(getLinkage, foocpp) == "C++");
+static assert(__traits(getLinkage, foow) == "Windows");
+static assert(__traits(getLinkage, foop) == "Pascal");
+static assert(__traits(getLinkage, fooobjc) == "Objective-C");
+version (Windows)
+    static assert(__traits(getLinkage, foos) == "Windows");
+else
+    static assert(__traits(getLinkage, foos) == "C");
+
+extern (C) int global;
+static assert(__traits(getLinkage, global) == "C");
+
--- a/src/ddmd/idgen.d
+++ b/src/ddmd/idgen.d
@@ -356,6 +356,7 @@ Msgtable[] msgtable =
     { "getAliasThis" },
     { "getAttributes" },
     { "getFunctionAttributes" },
+    { "getLinkage" },
     { "getUnitTests" },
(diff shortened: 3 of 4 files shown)


src/ddmd/traits.d
```diff
@@ -910,6 +911,35 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        auto o = (*e.args)[0];
+        auto s = getDsymbol(o);
+        Declaration d;
+        if (!s || (d = s.isDeclaration()) is null)
+        {
+            e.error("argument to `__traits(getLinkage, %s)` is not a declaration", o.toChars());
+            return new ErrorExp();
+        }
+        string linkage;
+        switch (d.linkage)
```
UplinkCoder: final switch ?

UplinkCoder (approved):
Love it.

src/ddmd/traits.d
WalterBright: .def is not represented

UplinkCoder: What is .def ?
give it a string representation, then.
Better then a halt.

UplinkCoder: ahh default linkage.
This should be D linkage, no ?

wilzbach:
Doesn't this need an accompanying PR to dlang.org for the doc change and probably a changelog entry as well?

src/ddmd/traits.d
```diff
@@ -910,6 +911,35 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+            case LINK.d:        linkage = "D";           break;
+            case LINK.c:        linkage = "C";           break;
+            case LINK.cpp:      linkage = "C++";         break;
+            case LINK.windows:  linkage = "Windows";     break;
+            case LINK.pascal:   linkage = "Pascal";      break;
+            case LINK.objc:     linkage = "Objective-C"; break;
+            case LINK.system:   linkage = "System";      break;
+            default: assert(0);
+        }
+        auto se = new StringExp(e.loc, cast(char*)linkage.ptr);
```
UplinkCoder: &linkage[0]

src/ddmd/traits.d
```diff
@@ -910,6 +911,35 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        auto s = getDsymbol(o);
+        Declaration d;
+        if (!s || (d = s.isDeclaration()) is null)
+        {
+            e.error("argument to `__traits(getLinkage, %s)` is not a declaration", o.toChars());
+            return new ErrorExp();
+        }
+        string linkage;
+        switch (d.linkage)
+        {
```
UplinkCoder: Maybe ... rather then using strings here we take the pointers to the identifiers out of the string table we already have ?
Id.{D, C, Windows ... }.string

src/ddmd/traits.d
UplinkCoder: also maybe use `new StringExp(e.loc, cast(void*) ident.string, ident.len` ?

test/compilable/test17419.d
```diff
@@ -0,0 +1,29 @@
...
+static assert(__traits(getLinkage, aliasc) == "C");
+
+extern (C) int food();
+extern (C++) int foocpp();
+extern (Windows) int foow();
+extern (Pascal) int foop();
+extern (Objective-C) int fooobjc();
+extern (System) int foos();
+
+static assert(__traits(getLinkage, fooc) == "C");
```
UplinkCoder: I suppose this was to be `food` ?

UplinkCoder (commented):
Maybe address the comments ?

src/ddmd/traits.d
WalterBright: Won't work because of `C++` and `Objective-C`

test/compilable/test17419.d
WalterBright: It's c, not d

src/ddmd/traits.d
UplinkCoder: right.

test/compilable/test17419.d
UplinkCoder: the symbol `food`, is not referenced in this test.
while `fooc` is referenced twice.

WalterBright: yes, my mistake

src/ddmd/traits.d
WalterBright: I don't see the point.

UplinkCoder: the current version will call `strlen` inside of StringExp constructor.
which is probably negligible but still we don't need to throw the length information away.

test/compilable/test17419.d
```diff
@@ -0,0 +1,29 @@
+// https://issues.dlang.org/show_bug.cgi?id=17419
+
+
+extern (C) int fooc();
+alias aliasc = fooc;
+
+static assert(__traits(getLinkage, fooc) == "C");
+static assert(__traits(getLinkage, aliasc) == "C");
+
+extern (C) int food();
```
UplinkCoder: this should be `extern(D)`

src/ddmd/traits.d
```diff
@@ -910,6 +911,35 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+            case LINK.d:        linkage = "D";           break;
+            case LINK.c:        linkage = "C";           break;
+            case LINK.cpp:      linkage = "C++";         break;
+            case LINK.windows:  linkage = "Windows";     break;
+            case LINK.pascal:   linkage = "Pascal";      break;
+            case LINK.objc:     linkage = "Objective-C"; break;
+            case LINK.system:   linkage = "System";      break;
```
UplinkCoder: This line is uncovered ...
I guess system linkage is replaced earlier
should this be an ice then ?

src/ddmd/traits.d
```diff
@@ -910,6 +911,35 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+            default: assert(0);
```
UplinkCoder: please consider making this a proper ice

src/ddmd/traits.d
WalterBright: Not worth the extra complexity.

src/ddmd/traits.d
WalterBright: no

src/ddmd/traits.d
WalterBright: Extra complexity of duplicating other logic is not worth it.

src/ddmd/traits.d
WalterBright: No, that's what asserts are for.

andralex (approved):
lgtm, thanks

src/ddmd/traits.d
ibuclaw: `extern(System)` is resolved when building the parsing tree.  If anyone wants to fix that to defer until semantic, be my guest.  However I suspect even then that is still before traits are evaluated.

PR #6825, opened 2017-05-23
fix Issue 17421 - add __traits(getFunctionVariadicStyle, f)

WalterBright:


src/ddmd/traits.d
```diff
@@ -911,6 +912,61 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+            return dimError(1);
+
+        LINK link;
+        int varargs;
+        auto o = (*e.args)[0];
+        auto t = isType(o);
+        TypeFunction tf = null;
+        if (t)
+        {
+            if (t.ty == Tfunction)
```
UplinkCoder: time to put this into a function ?

src/ddmd/traits.d
```diff
@@ -911,6 +912,61 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        if (tf)
+        {
+            link = tf.linkage;
+            varargs = tf.varargs;
+        }
+        else
+        {
+            auto s = getDsymbol(o);
+            FuncDeclaration fd;
+            if (!s || (fd = s.isFuncDeclaration()) is null)
```
UplinkCoder: I consider putting the assignment in the if a wired thing; but that's just me.

src/ddmd/traits.d
```diff
@@ -911,6 +912,61 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+            {
+                e.error("argument to `__traits(getFunctionVariadicStyle, %s)` is not a function", o.toChars());
+                return new ErrorExp();
+            }
+            link = fd.linkage;
+            fd.getParameters(&varargs);
+        }
+        string style;
+        switch (varargs)
+        {
```
UplinkCoder: could you give those switch cases symbolic names ?

UplinkCoder (commented):
just a few readability nits.

src/ddmd/traits.d
WalterBright: I try not to mix refactoring with other things.

src/ddmd/traits.d
WalterBright: I try not to mix refactoring with other things.

src/ddmd/traits.d
```diff
@@ -911,6 +912,61 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+         *  "D"         extern(D) void dstyle(...), with __argptr and __arguments
+         *  "C"         extern(C) void cstyle(int, ...), use core.stdc.stdarg
+         *  "typesafe"  void typesafe(T[] ...)
+         */
+        // get symbol linkage as a string
+        if (dim != 1)
+            return dimError(1);
+
+        LINK link;
+        int varargs;
```
UplinkCoder: aren't those are more commonly known as `vaargs` ?

MartinNowak: That's from C macros and functions, __VA_ARGS__ and va_arg, but in the dmd frontend it's called varargs everywhere.

MartinNowak (approved):

MartinNowak:
Changelog entry still missing @WalterBright (https://github.com/dlang/dmd/tree/master/changelog), also please submit a spec update.

src/ddmd/traits.d
klickverbot: They are commonly known as variadic arguments, or [varargs](https://en.wikipedia.org/wiki/Varargs) .The original C header was called `varargs.h` as well. As Martin points out, you might be thinking of `va_arg()`, which is one of the `va_*` family of functions and types to handle varargs.

WalterBright:
I changed the strings returned because "C" and "D" were always a bit confusing.

UplinkCoder:
argptr and stdarg is also not very intuitive.

WalterBright:
C style are accessed with `core.stdc.stdargs`, D style are accessed with `__argptr`.

src/ddmd/traits.d
ibuclaw: @UplinkCoder feel free to raise a bugzilla issue for every int field that should be an enum.

ibuclaw:
> C style are accessed with core.stdc.stdargs, D style are accessed with __argptr.

Actually, you access both C and D style with core.stdc.stdargs because of the vagary of valist types. And _argptr is created for both C and D linkage functions, what separates them is that D functions also include a hidden _arguments parameter.

Saying that, nothing is stopping some implementation detail of doing D varargs differently, say, create a `void*[N]` array and fill it with the reference to each vararg.

PR #6826, opened 2017-05-24
[scope] fix Issue 17422 - class reference not initialized as scope variable

MartinNowak:
- infer scope for class references to prevent escaping
- those are just normal class references with scope to prevent escaping (copying)
- only `scope o = new Object` remains special, because it uses the fact
  that the class ref cannot be escaped to allocate the instance on the stack

### DIFF
--- a/src/ddmd/declaration.d
+++ b/src/ddmd/declaration.d
@@ -2241,7 +2241,7 @@ extern (C++) class VarDeclaration : Declaration
                     // Destructors are not supported on extern(C++) classes
                     break;
                 }
-                if (mynew || onstack || cd.dtors.dim) // if any destructors
+                if (mynew || onstack) // if any destructors
                 {
                     // delete this;
                     Expression ec;
--- a/src/ddmd/declaration.d
+++ b/src/ddmd/declaration.d
@@ -2236,9 +2236,10 @@ extern (C++) class VarDeclaration : Declaration
                 //if (cd.isInterfaceDeclaration())
                 //    error("interface %s cannot be scope", cd.toChars());
 
+                // Destroying C++ scope classes crashes currently. Since C++ class dtors are not currently supported, simply do not run dtors for them.
+                // See https://issues.dlang.org/show_bug.cgi?id=13182
                 if (cd.cpp)
                 {
-                    // Destructors are not supported on extern(C++) classes
                     break;
                 }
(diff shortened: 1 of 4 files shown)


MartinNowak:
Looks like this might add another dtor call in some places, e.g. https://github.com/dlang/dmd/blob/c80baade09a9ca75a06dda6f35eb9d2a56639ad9/test/runnable/testdstress.d#L654.

WalterBright:
The only problem is an Object being set to 'scope' is not the same thing as an Object being declared as 'scope', because the latter means it gets allocated on the stack. Some checking of the code will be necessary to ensure these are not conflated.

MartinNowak:
Fixed the double dtor issue, filed as [Issue 17429 – [scope] each scope class references adds another delete call](https://issues.dlang.org/show_bug.cgi?id=17429).

> The only problem is an Object being set to 'scope' is not the same thing as an Object being declared as 'scope', because the latter means it gets allocated on the stack. Some checking of the code will be necessary to ensure these are not conflated.

That's actually fairly simple, scope now means that the class reference cannot escape. When we have that in a `scope o = new Object` (NewExp as initializer), we use the fact that it cannot be escaped to allocate the class on the stack. So stack allocated classes are always a subset of scoped class references.
One problem might arise though, when scope is inferred for a NewExp, `auto o = new Klazz(scopedVariable)`. This is currently a hole anyhow ([Issue 17370](https://issues.dlang.org/show_bug.cgi?id=17370) and [Issue 17428](https://issues.dlang.org/show_bug.cgi?id=17428)), but also shouldn't be hard to distinguish between an explicit and an inferred scope for that case if deemed necessary.

MartinNowak:
> One problem might arise though, when scope is inferred for a NewExp, auto o = new Klazz(scopedVariable). This is currently a hole anyhow (Issue 17370 and Issue 17428), but also shouldn't be hard to distinguish between an explicit and an inferred scope for that case if deemed necessary.

In fact it is necessary to deterministically destroy classes containing scoped ptrs anyhow, b/c a finalizer could still access the now dangling pointers. So it seems like a reasonable choice to allocate new class instances that are inferred to be scope on the stack.

WalterBright:
This has apparently caused a regression: https://issues.dlang.org/show_bug.cgi?id=17869

PR #6829, opened 2017-05-25
fix Issue 17425 - add __traits(getParameterStorageClasses, f, i)

WalterBright:
The point of this is so that `std.traits.ParameterStorageClassTuple()` can be implemented without needing to study the name mangling.

### DIFF
--- a/changelog/changelog17425.dd
+++ b/changelog/changelog17425.dd
@@ -1,0 +1,18 @@
+add __traits(getParameterStorageClasses, f, i)
+
+$(LINK2 https://issues.dlang.org/show_bug.cgi?id=17425, Bugzilla 17425)
+
+---
+ref int foo(return ref const int* p, scope int* a, out int b, lazy int c);
+
+pragma(msg, __traits(getParameterStorageClasses, foo, 0));
+
+static assert(__traits(getParameterStorageClasses, foo, 0)[0] == "return");
+static assert(__traits(getParameterStorageClasses, foo, 0)[1] == "ref");
+
+pragma(msg, __traits(getParameterStorageClasses, foo, 1));
+static assert(__traits(getParameterStorageClasses, foo, 1)[0] == "scope");
+static assert(__traits(getParameterStorageClasses, foo, 2)[0] == "out");
+static assert(__traits(getParameterStorageClasses, typeof(&foo), 3)[0] == "lazy");
+---
+
--- a/src/ddmd/traits.d
+++ b/src/ddmd/traits.d
@@ -97,6 +97,7 @@ static this()
         "getAttributes",
         "getFunctionAttributes",
         "getFunctionVariadicStyle",
+        "getParameterStorageClasses",
         "getUnitTests",
         "getVirtualIndex",
         "getPointerBitmap",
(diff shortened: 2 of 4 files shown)


WalterBright:
Blocking https://github.com/dlang/phobos/pull/5427

UplinkCoder:
@WalterBright there is some kind of error on osx32
I am not sure if it is related.

WalterBright:
> I am not sure if it is related.

Apparently another heisenbug in the testing scripts.

UplinkCoder (approved):
Looks fine to me.

changelog/changelog17425.dd
```diff
@@ -0,0 +1,18 @@
+add __traits(getParameterStorageClasses, f, i)
+
+$(LINK2 https://issues.dlang.org/show_bug.cgi?id=17425, Bugzilla 17425)
```
rainers: The changelog entry seems a bit terse. I'd use at least more descriptive variable names. Also, linking to the documentation instead of bugzilla might be more helpful.

src/ddmd/traits.d
```diff
@@ -968,6 +969,109 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         auto se = new StringExp(e.loc, cast(char*)style);
         return se.semantic(sc);
     }
+    if (e.ident == Id.getParameterStorageClasses)
+    {
+        /* Accept a function symbol or a type, followed by a parameter index.
```
rainers: How about putting this code into a function? This improves readable and avoids making this function even larger (dmd is rather slow optimizing large functions).

src/ddmd/traits.d
```diff
@@ -968,6 +969,109 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        // Set stc to storage class of the ith parameter
+        auto ex = isExpression((*e.args)[1]);
+        if (!ex)
+        {
+            e.error("expression expected as second argument of `__traits(getParameterStorageClasses, %s, %s)`",
+                o.toChars(), o1.toChars());
+            return new ErrorExp();
+        }
+        ex = ex.ctfeInterpret();
+        auto ii = ex.toUInteger();
```
rainers: This doesn't propagate errors if the expression does not evaluate to an integer. That is not checked in other uses of toInteger/toUInteger elsewhere, too, though. Maybe add test `pragma(msg, __traits(getParameterStorageClasses, foo, "1"));`?

rainers (approved):
Just a few nits, otherwise LGTM.

changelog/changelog17425.dd
wilzbach: FYI the auto-tester allows to preview this since some time (though I am not sure why this isn't part of the diff, CC @CyberShadow)
In any case the link is e.g. http://dtest.dlang.io/artifact/website-72ba79b73f635fb08bdfe69d6826833c1c4c2ee0-a9052bf749bf29e4ae70eb23d51613de/web/changelog/2.075.0_pre.html#changelog17425 (the _pre file in the changelog folder).
FWIW there also seems to be a parenthesis problem in another changelog file..

src/ddmd/traits.d
WalterBright: I leave refactorings to subsequent PRs, because they distract from what is being changed here.

src/ddmd/traits.d
WalterBright: That's a generic problem with all those functions, and fixing them should be in a separate PR.

changelog/changelog17425.dd
WalterBright: I didn't link to the documentation because that PR has not been pulled.

CyberShadow:
> though I am not sure why this isn't part of the diff, CC @CyberShadow

Oops, diff was eaten by an overly voracious disk cleanup script. Fixed for future cleanups.

PR #6831, opened 2017-05-25
Fix Issue 17437 - Incorrect range check when implicitly converting integer literals to float/double

kinke:
After casting the integer value to the floating-point type, make sure to backcast it to an integer for the equality check, so that the comparison is performed as integer instead of the fp type with potentially smaller
integer range. This prevents inexact implicit convertibility.

### DIFF
--- a/test/fail_compilation/impconv.d
+++ b/test/fail_compilation/impconv.d
@@ -1,0 +1,36 @@
+/*
+FIXME: Older 32-bit DMD host compilers with faulty optimization
+lead to unfortunate test failures, see
+https://github.com/dlang/dmd/pull/6831#issuecomment-304495842.
+
+DISABLED: win32 win64 linux freebsd32
+*/
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/impconv.d(24): Error: function `impconv.foo_float(float)` is not callable using argument types `(int)`
+fail_compilation/impconv.d(30): Error: function `impconv.foo_double(double)` is not callable using argument types `(long)`
+---
+*/
+
+void foo_float(float);
+void foo_double(double);
+void foo_real(real);
+
+void main()
+{
+    foo_float(1);        // implicitly convertible to float
+    foo_float(-int.max); // -(2^31 - 1)
+    // FIXME: Unsigned constants don't produce an error,
+    // see https://issues.dlang.org/show_bug.cgi?id=17436.
+    //foo_float(uint.max); // 2^32 - 1
+
+    foo_double(int.max);   // implicitly convertible to double
+    foo_double(-long.max); // -(2^63 - 1)
+    //foo_double(ulong.max); // 2^64 - 1
+
+    foo_real(0xffff_ffff_ffffL); // 2^48 - 1, implicitly convertible to real
+    static assert(__traits(compiles, foo_real(-long.max)) == (real.mant_dig >= 63));
+    static assert(__traits(compiles, foo_real(ulong.max)) == (real.mant_dig >= 64));
+}
--- a/test/fail_compilation/impconv.d
+++ b/test/fail_compilation/impconv.d
@@ -1,16 +1,18 @@
 /*
-FIXME: Older 32-bit DMD host compilers with faulty optimization
+FIXME: DMD host compilers < 2.073 with faulty optimization
 lead to unfortunate test failures, see
 https://github.com/dlang/dmd/pull/6831#issuecomment-304495842.
 
-DISABLED: win32 win64 linux freebsd32
+DISABLED: win32 win64 linux osx freebsd
 */
 
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/impconv.d(24): Error: function `impconv.foo_float(float)` is not callable using argument types `(int)`
+fail_compilation/impconv.d(26): Error: function `impconv.foo_float(float)` is not callable using argument types `(int)`
+fail_compilation/impconv.d(27): Error: function `impconv.foo_float(float)` is not callable using argument types `(uint)`
 fail_compilation/impconv.d(30): Error: function `impconv.foo_double(double)` is not callable using argument types `(long)`
+fail_compilation/impconv.d(31): Error: function `impconv.foo_double(double)` is not callable using argument types `(ulong)`
 ---
 */
 
@@ -22,13 +24,11 @@ void main()
 {
     foo_float(1);        // implicitly convertible to float
     foo_float(-int.max); // -(2^31 - 1)
-    // FIXME: Unsigned constants don't produce an error,
-    // see https://issues.dlang.org/show_bug.cgi?id=17436.
-    //foo_float(uint.max); // 2^32 - 1
+    foo_float(uint.max); // 2^32 - 1
 
     foo_double(int.max);   // implicitly convertible to double
     foo_double(-long.max); // -(2^63 - 1)
-    //foo_double(ulong.max); // 2^64 - 1
+    foo_double(ulong.max); // 2^64 - 1
 
     foo_real(0xffff_ffff_ffffL); // 2^48 - 1, implicitly convertible to real
     static assert(__traits(compiles, foo_real(-long.max)) == (real.mant_dig >= 63));
--- a/src/dmd/dcast.d
+++ b/src/dmd/dcast.d
@@ -366,7 +366,20 @@ extern (C++) MATCH implicitConvTo(Expression e, Type t)
             }
 
             // Only allow conversion if no change in value
-            dinteger_t value = e.toInteger();
+            immutable dinteger_t value = e.toInteger();
+
+            bool isLosslesslyConvertibleToFP(T)()
+            {
+                if (e.type.isunsigned())
+                {
+                    const f = cast(T) value;
+                    return cast(dinteger_t) f == value;
+                }
+
+                const f = cast(T) cast(sinteger_t) value;
+                return cast(sinteger_t) f == cast(sinteger_t) value;
+            }
+
             switch (toty)
             {
             case Tbool:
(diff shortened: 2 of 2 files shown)


WalterBright:
Please add a bugzilla issue identifying the problem this PR solves. Thanks!

kinke:
I'd rather summarize the linked LDC issue here - previously, when checking whether an `IntegerExp` can be implicitly converted to float/double, the `dinteger_t` value was cast to the FP type, and then that value was checked for equality against the original `dinteger_t` value. So the equality check was performed as FP type (with its potentially smaller range of representable integers).

Old: `const f = cast(T) integerValue; if (f != integerValue) { /* not convertible */ }`
New: `const f = cast(T) integerValue; if (cast(dinteger_t) f != integerValue) { /* not convertible */ }`

---

There must be another related issue involved here, as the current autotester results clearly show (missing failures). DMD 2.074 on Windows passes this:

```D
bool foo(int x)
{
    const f = cast(float) x;
    return f == x;
}

void main()
{
    assert(foo(1));
    assert(!foo(int.max));
}
```

but only with `-m32` and `-m32mscoff`; the 2nd assert fails with `-m64`.

WalterBright:
> I'd rather summarize the linked LDC issue here

Bug fixes are organized and tracked with Bugzilla. It puts an undue burden on people like Martin who carefully maintain the changlog, etc., if the bug reports are scattered about various forums. This issue also may already be in bugzilla.

kinke:
[Issue #17437](https://issues.dlang.org/show_bug.cgi?id=17437) created. The test failures on the 32-bit hosts are most likely due to the other related [issue #17436](https://issues.dlang.org/show_bug.cgi?id=17436). which could be something more severe.

wilzbach:
> Issue #17437 created. The test failures on the 32-bit hosts are most likely due to the other related issue #17436. which could be something more severe.

@kinke And another 10 points would go to you if you rename your commit message, s.t. it can be automatically parsed and included in the changelog. The easiest is to copy the bugzilla title and prepend "Fix ", but for more infos you can see [the Dlang-Bot's page](https://github.com/dlang-bots/dlang-bot).

MartinNowak:
Check seems to get optimized away in release 32-bit dmd builds with the current host compiler (2.068.2). Any 32-bit dmd binary under 2.073.0 fails the following test with `-O`.
```d
alias dinteger_t = ulong;
alias sinteger_t = long;

bool isLosslesslyConvertibleToFP(T)(const dinteger_t value)
{
    const f = cast(T) cast(sinteger_t) value;
    return cast(sinteger_t) f == cast(sinteger_t) value;
}

void main()
{
    immutable dinteger_t value = int.max;
    assert(!isLosslesslyConvertibleToFP!float(value));
}
```

MartinNowak:
> Check seems to get optimized away in release 32-bit dmd builds with the current host compiler (2.068.2). Any 32-bit dmd binary under 2.073.0 fails the following test with -O.

Maybe just write and read to a global variable to prevent optimization @kinke.

kinke:
I'd prefer not adding extremely ugly workarounds for buggy DMD host compilers. I tried restricting the tests to 64-bit targets (as I didn't find a quick way to determine the host bitness), but forgot that the autotester also cross-compiles from 32-bit hosts to 64-bit targets.

Should I just remove the tests?

MartinNowak:
We're about to add the `toPrec` intrinsics that could then be used for dmd.

wilzbach:
PRs pointing at stable shouldn't get stalled. It should be dead simple: either it's a regression fix, then merge or target master instead. I will look into implementing a warning at dlang-bot for all pulls to stable which are older than three days.
I changed the base here because PRs to stable have a priority in the auto-tester queue and thus take up quite a bit of resources.

WalterBright:
Looks like the failing test case needs a rebase. Otherwise, this fix looks good.

WalterBright (changes requested):
add `` to test case

kinke:
Test adapted, but the 32-bit DMD host compiler issue still persists.

JinShil:
> Test adapted, but the 32-bit DMD host compiler issue still persists.

Is the bug fixed in a later version of the compiler?  If so, I offer the following suggestion:

1.  `version`-out the test for 32-bit
2.  Create a Bugzilla issue documenting that the test needs to be un-`version`ed when we upgrade to a later version of the host compiler
3.  Add a link to the Bugzilla issue in the test case.
4.  Approve and merge this PR.

test/fail_compilation/impconv.d
```diff
@@ -0,0 +1,18 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/impconv.d(16): Error: function `impconv.foo(float)` is not callable using argument types `(int)`
+fail_compilation/impconv.d(17): Error: function `impconv.bar(double)` is not callable using argument types `(long)`
+---
+*/
+
+void foo(float);
+void bar(double);
```
WalterBright: Add test for `real`?

WalterBright: 6 tests are suggested here: [float, double, real] * [int.max, long.max]

kinke: I went for [1, int.max] for `float`, as long.max would just overflow as well as int.max already.

Wrt. `real`, that's compiler-specific as it depends on the front-end's `real_t` precision. Meaning that LDC and GDC would need to adapt that test, something I tried to avoid.

kinke:
AFAICT the test can't be versioned out (as fail_compilation test with checked error output) but disabled in the Makefile.

WalterBright:
You don't actually need to check the error message for the `real` cases, you can use `__traits(compiles, ...)` on it and version the result of that. It is important to cover all the cases, it's the only way we can be confident it is working and will stay working as expected.

test/fail_compilation/impconv.d
ibuclaw: The precision of `real_t` should be the same as the target platform (just not necessarily the same precision as the host). Having version ARM, MIPS, etc is safe.

JinShil:
> AFAICT the test can't be versioned out (as fail_compilation test with checked error output) but disabled in the Makefile.

Right.  Sorry.  I didn't really think that through.  But you get the idea, let's just disable the 32-bit tests for now and log the problem in Bugzilla so we can get this finished.  Another option is to add this to the test's comment header

```D
// DISABLED: win32 win64 osx linux32 freebsd32
```

test/fail_compilation/impconv.d
kinke: Not true for LDC in all cases. What happened to your huge common `real_t` type with ~160 bit mantissa?

ibuclaw: `real_t` is used for floats, doubles, and reals alike.  It may hold more precision, but when read, it's still rounded to match at the same precision as the target.

kinke:
I added `real` unittests and disabled the test on most platforms. It's enabled for OSX and FreeBSD64 as the autotester doesn't use 32-bit host compilers for these.

I also added unittests for unsigned constants, but they don't error out (incl. using 64-bit host compilers), so I commented them out and refer to closely related https://issues.dlang.org/show_bug.cgi?id=17436.

JinShil:
Why disabled for Linux64? I thought there was only a problem with 32-bit hosts.

> but they don't error out (incl. using 64-bit host compilers)

Then shouldn't the bug report be updated to include 64-bit host compilers.

Please forgive my ignorance, but this is getting confusing.  Precisely which host/test permutations are not producing expected results?

kinke:
> Please forgive my ignorance, but this is getting confusing. Precisely which host/test permutations are not producing expected results?

All permutations using an older 32-bit **host** DMD, i.e., Linux_32_64 too, whereas Linux_64_32 works (as would using a more recent host compiler, or LDC). These are CTFE tests, which actually test the host compiler, so things are confusing. ;)

> Then shouldn't the bug report be updated to include 64-bit host compilers.

My tests back then apparently worked for 64-bit host compilers. But those were runtime tests, testing the actual compiler. ;) Perhaps this was fixed in ~v2.073 (for 64-bit code only) and so the CTFE issue here only pops up with older DMD host compilers. As soon as this is merged, we can point to the commit (and commented-out tests) as follow-up in that bug report.

kinke:
This isn't blocked any longer.

test/fail_compilation/impconv.d
```diff
@@ -0,0 +1,36 @@
+/*
+FIXME: Older 32-bit DMD host compilers with faulty optimization
+lead to unfortunate test failures, see
+https://github.com/dlang/dmd/pull/6831#issuecomment-304495842.
```
JinShil: Please add a note in which version these disabled tests can be removed.

test/fail_compilation/impconv.d
```diff
@@ -0,0 +1,36 @@
...
+{
+    foo_float(1);        // implicitly convertible to float
+    foo_float(-int.max); // -(2^31 - 1)
+    // FIXME: Unsigned constants don't produce an error,
+    // see https://issues.dlang.org/show_bug.cgi?id=17436.
+    //foo_float(uint.max); // 2^32 - 1
+
+    foo_double(int.max);   // implicitly convertible to double
+    foo_double(-long.max); // -(2^63 - 1)
+    //foo_double(ulong.max); // 2^64 - 1
```
JinShil: Same here.  Needs a note in which compiler version these can be enabled.

JinShil (changes requested):

kinke:
> Perhaps this was fixed in ~v2.073 (for 64-bit code only) and so the CTFE issue here only pops up with older DMD host compilers. As soon as this is merged, we can point to the commit (and commented-out tests) as follow-up in issue 17436.

Nope, the unsigned case appears to have been fixed by 2.073 just like the signed one (I tested a Win32 build); issue 17436 is something else and still present in 2.078.0. So it's just that the issue in the signed case appears to be exclusive to 32-bit versions.

So is it worth keeping the test enabled on autotester targets without 32-bit hosts (by keeping on commenting out the unsigned cases), or should I simply fully disable the test for all targets (with that `DISABLED but PASSES!!!` annotation if it does pass)?

kinke:
I went for disabling the test on all autotester platforms, the other option is way too hacky IMO. So can we finally get this merged please? [I verified the test with LDC.]

JinShil (approved):
I can't justify having a fix for a bug and not merging it simply because we have bugs in our host compiler.  I'll approve this, but it needs a review by someone more senior than I.  cc @WalterBright @ibuclaw @MartinNowak  @andralex

rainers (approved):
LGTM. I hope the released dmd is not built with an old compiler.

wilzbach:
Closed & re-opened for SemaphoreCI to learn about this PR.

test/fail_compilation/impconv.d
```diff
@@ -0,0 +1,36 @@
+/*
+FIXME: DMD host compilers < 2.073 with faulty optimization
```
wilzbach: We should have a wiki page with things that can be enabled on the next host compiler bump.
I started one with this PR: https://wiki.dlang.org/Next_Compiler_Bump

PR #6835, opened 2017-05-27
deduplicate the 'isSomeFunction' equivalent

UplinkCoder:
Since the code is now used multiple times in `__traits` let's dedup it.

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -14,6 +14,7 @@ module dmd.traits;
 
 import core.stdc.stdio;
 import core.stdc.string;
+
 import dmd.aggregate;
 import dmd.arraytypes;
 import dmd.canthrow;
@@ -442,6 +443,52 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
     Expression False()
     {
         return new IntegerExp(e.loc, false, Type.tbool);
+    }
+
+    /**
+       Gets a function type from a given AST node 
+       if the node is a function of some sort.
+
+     Params:
+        o = an AST node to check for a `TypeFunction`  
+        fdp = optional pointer to a function declararion, to be set 
+          if `o` is a function declarartion.
+
+     Returns: 
+        a type node if `o` is a declaration of
+            a delegate, function, function-pointer
+          or a variable of the former.  Otherwise, `null`.
+    */
+
+    static TypeFunction traitsFuncArg(RootObject o, FuncDeclaration* fdp = null)
+    {
+        auto s = getDsymbolWithoutExpCtx(o);
+        auto t = isType(o);
+        TypeFunction tf = null;
+
+        if (s)
+        {
+            auto fd = s.isFuncDeclaration();
+            if (fd)
+            {
+                t = fd.type;
+                if (fdp)
+                    *fdp = fd;
+            }
+            else if (auto vd = s.isVarDeclaration())
+                t = vd.type;
+        }
+        if (t)
+        {
+            if (t.ty == Tfunction)
+                tf = cast(TypeFunction)t;
+            else if (t.ty == Tdelegate)
+                tf = cast(TypeFunction)t.nextOf();
+            else if (t.ty == Tpointer && t.nextOf().ty == Tfunction)
+                tf = cast(TypeFunction)t.nextOf();
+        }
+
+        return tf;
     }
 
     Expression isX(T)(bool function(T) fp)
@@ -978,26 +1025,8 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         if (dim != 1)
             return dimError(1);
 
-        auto o = (*e.args)[0];
-        auto s = getDsymbolWithoutExpCtx(o);
-        auto t = isType(o);
-        TypeFunction tf = null;
-        if (s)
-        {
-            if (auto fd = s.isFuncDeclaration())
-                t = fd.type;
-            else if (auto vd = s.isVarDeclaration())
-                t = vd.type;
-        }
-        if (t)
-        {
-            if (t.ty == Tfunction)
-                tf = cast(TypeFunction)t;
-            else if (t.ty == Tdelegate)
-                tf = cast(TypeFunction)t.nextOf();
-            else if (t.ty == Tpointer && t.nextOf().ty == Tfunction)
-                tf = cast(TypeFunction)t.nextOf();
-        }
+        TypeFunction tf = traitsFuncArg((*e.args)[0]);
+
         if (!tf)
         {
             e.error("first argument is not a function");
@@ -1028,17 +1057,10 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         LINK link;
         int varargs;
         auto o = (*e.args)[0];
-        auto t = isType(o);
-        TypeFunction tf = null;
-        if (t)
-        {
-            if (t.ty == Tfunction)
-                tf = cast(TypeFunction)t;
-            else if (t.ty == Tdelegate)
-                tf = cast(TypeFunction)t.nextOf();
-            else if (t.ty == Tpointer && t.nextOf().ty == Tfunction)
-                tf = cast(TypeFunction)t.nextOf();
-        }
+
+        FuncDeclaration fd;
+        TypeFunction tf = traitsFuncArg(o, &fd);
+
         if (tf)
         {
             link = tf.linkage;
@@ -1046,9 +1068,7 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         }
         else
         {
-            auto s = getDsymbol(o);
-            FuncDeclaration fd;
-            if (!s || (fd = s.isFuncDeclaration()) is null)
+            if (!fd)
             {
                 e.error("argument to `__traits(getFunctionVariadicStyle, %s)` is not a function", o.toChars());
                 return new ErrorExp();
@@ -1078,19 +1098,12 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         if (dim != 2)
             return dimError(2);
 
-        auto o1 = (*e.args)[1];
         auto o = (*e.args)[0];
-        auto t = isType(o);
-        TypeFunction tf = null;
-        if (t)
-        {
-            if (t.ty == Tfunction)
-                tf = cast(TypeFunction)t;
-            else if (t.ty == Tdelegate)
-                tf = cast(TypeFunction)t.nextOf();
-            else if (t.ty == Tpointer && t.nextOf().ty == Tfunction)
-                tf = cast(TypeFunction)t.nextOf();
-        }
+        auto o1 = (*e.args)[1];
+
+        FuncDeclaration fd;
+        TypeFunction tf = traitsFuncArg(o, &fd);
+
         Parameters* fparams;
         if (tf)
         {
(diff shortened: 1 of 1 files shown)


src/ddmd/traits.d
```diff
@@ -403,6 +403,33 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         return new IntegerExp(e.loc, false, Type.tbool);
     }
 
+    TypeFunction firstArgFunctionType()
```
rainers: IMO a better abstraction is `toTypeFunction(RootObject o)` and passing the validated parameter into this function. That would allow it not being a nested function and used elsewhere, too. 

Even better, I'd split it into two overloads, one for Type, and one for Dsymbol, as you now have widen the allowed parameters to traits that support types only so far.

UplinkCoder: the reason I did not do this is because I'd have to import `rootobject`

UplinkCoder: also let's generalize when/if needed and not on speculation.

UplinkCoder:
travis fails because of connection error

wilzbach:
@uplinkcoder: restarted and created another card for it https://trello.com/c/aM5r3Bio/320-use-more-download-fallbacks-in-the-installer-script

src/ddmd/traits.d
UplinkCoder: @rainers do you agree ?

rainers: > do you agree ?

Mostly. The diff didn't show that there's also support for function symbols in the other traits.

You should remove the handling for FuncDeclaration in traits getLinkage, getFunctionVariadicStyle and getParameterStorageClasses, though. It is duplicated now.

> the reason I did not do this is because I'd have to import rootobject

That's not really a good excuse. You can also use typeof(args[0]) to avoid an import... ;-)

UplinkCoder:
@rainers do you like it better now ?

UplinkCoder:
@WalterBright any objections ?

src/ddmd/traits.d
```diff
@@ -403,6 +404,37 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         return new IntegerExp(e.loc, false, Type.tbool);
     }
 
+    static TypeFunction traitsFuncArg(RootObject o, FuncDeclaration* fdp = null)
```
WalterBright: No documentation saying what function does.

WalterBright:
LGTM provided the function is documented

WalterBright:
>use typeof(args[0]) to avoid an import... ;-)

Not a bad idea.

src/ddmd/traits.d
```diff
@@ -403,6 +404,47 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+    /**
+     if traitsFuncArg is passed either a delegate. a function
+     or a function pointer or a variable of types.
+     it will return the functionType.
+     else it will return null
+
+     In case of a function declaration being passed verbatim
+     it will set the second  optional argument to the FuncDeclarartion
+     given it was provided.
+    */
```
WalterBright: Please use Ddoc style documentation. It this case it means having a `Params:` section and a `Returns:` section. This is the standard for new functions.

src/ddmd/traits.d
```diff
@@ -403,6 +404,47 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         return new IntegerExp(e.loc, false, Type.tbool);
     }
 
+    /**
+     if traitsFuncArg is passed either a delegate. a function
+     or a function pointer or a variable of types.
```
WalterBright: omit trailing period in middle of sentence

src/ddmd/traits.d
```diff
@@ -403,6 +404,47 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+     it will return the functionType.
+     else it will return null
+
+     In case of a function declaration being passed verbatim
```
WalterBright: `passed verbatim` has no meaning in regards to code.

src/ddmd/traits.d
wilzbach: @UplinkCoder don't forget that we already publish the documentation of ddmd on dlang.org:
- http://dlang.org/phobos-prerelease/ddmd_mars.html (Ddoc)
- http://dlang.org/library-prerelease/ddmd/mars.html (Ddox)
and we intend to publish it on the main page very soon (e.g. once #6771 is merged).

This is also why it can be very helpful for both the contributor and the reviewers to preview the style changes at Vladimir's excellent DAutoTest, e.g.

http://dtest.dlang.io/artifact/website-72ba79b73f635fb08bdfe69d6826833c1c4c2ee0-6f020744a26d63533038dc39e44ae752/web/library-prerelease/ddmd/traits.html

(in this case it's a function within a function and thus _not_ publicly accessible)

src/ddmd/traits.d
```diff
@@ -403,6 +404,50 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
 
+    /**
+     Params: o: the AST-node to check.
+     Params: fdp: optionally a pointer to a function declararion,
+        which will get set o is a function declarartion.
+
+     Returns: a function-type if
+        o is a declarion of type  delegate, function,
+            or function pointer
+        or a variable of those types.
```
WalterBright: When using periods, please do not put them in the middle of sentences. Replace it with a `,`

UplinkCoder:
@WalterBright addressed.

src/ddmd/traits.d
UplinkCoder: @WalterBright do you like it better now ?

src/ddmd/traits.d
```diff
@@ -403,6 +404,50 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         return new IntegerExp(e.loc, false, Type.tbool);
     }
 
+    /**
+     Params: o: the AST-node to check.
```
WalterBright: Please read the Ddoc documentation I provided a link for. Note the `o:` in the above line. This is not how it's done and Ddoc will not work with it. It's `o=`.

JinShil:
@UplinkCoder Are you going to take this across the finish line?

src/ddmd/traits.d
```diff
@@ -403,6 +404,50 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+     Params: fdp: optionally a pointer to a function declararion,
```
JinShil: declararion --> declaration

src/ddmd/traits.d
```diff
@@ -403,6 +404,50 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+
+     Returns: a function-type if
+        o is a declarion of a delegate, a function,
+            a function pointer
+            or a variable of those types,
+        it will return the functionType
+     else
+        it will return null.
+    */
+    static TypeFunction traitsFuncArg(RootObject o, FuncDeclaration* fdp = null)
```
JinShil: I recommend a different name.  Perhaps `getTypeFunction`.

src/ddmd/traits.d
```diff
@@ -403,6 +404,50 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        which will get set o is a function declarartion.
+
+     Returns: a function-type if
+        o is a declarion of a delegate, a function,
+            a function pointer
+            or a variable of those types,
+        it will return the functionType
+     else
+        it will return null.
+    */
```
JinShil: I offer you this documentation header:

```D
/**
 Gets the `TypeFunction` for `o`
 Params:
    o   = the AST node to get the `TypeFunction` for
    fdp = optional pointer to a function declararion, to be set 
          if `o` is a function declarartion
 Returns: 
    a `TypeFunction` if `o` is a declaration for a delegate, function,
    or avariable of those types.  Otherwise, `null`
*/
```

JinShil (changes requested):

src/ddmd/traits.d
stefan-koch-sociomantic: Neat Thanks!

src/ddmd/traits.d
```diff
@@ -403,6 +404,50 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         return new IntegerExp(e.loc, false, Type.tbool);
     }
 
+    /**
+     Params: o: the AST-node to check.
+     Params: fdp: optionally a pointer to a function declararion,
+        which will get set o is a function declarartion.
+
```
BBasile: `if o`

src/dmd/traits.d
```diff
@@ -444,6 +445,52 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        fdp = optional pointer to a function declararion, to be set 
+          if `o` is a function declarartion.
+
+     Returns: 
+        a type node if `o` is a declaration of
+            a delegate, function, function-pointer
+          or a variable of the former.  Otherwise, `null`.
+    */
+
+    static TypeFunction traitsFuncArg(RootObject o, FuncDeclaration* fdp = null)
```
JinShil: All other DDoc headers in DMD seem to follow this convention:

```D
/*************************************************************
 * Whatever
 * Whatever
 */
```

Please remove the extra space between the DDoc header and the function declaration.

I cannot, in good conscience, approve a function name like `traitsFuncArg`.  Please use a more descriptive name like `getFuncType` or something along those lines (Yes, I know my own names have been poor).

JinShil (changes requested):

src/dmd/traits.d
wilzbach: > Please remove the extra space between the DDoc header and the function declaration.

Hmm the above is the Phobos style:
Compare with e.g. https://github.com/dlang/phobos/blob/master/std/algorithm/comparison.d

If you really want to stick with this style at DMD, then this should be documented in a DMD-specific style guide?

JinShil: > Hmm the above is the Phobos style.

I see.

> If you really want to stick with this style at DMD, then this should be documented in a DMD-specific style guide.

Agreed.  

Aside from the name then, LGTM.

JinShil: Also, consider making this a non-member function instead of static.  Refer to [@WalterBright's comment](https://github.com/dlang/dmd/pull/7405#pullrequestreview-82322025)

jacob-carlborg: Phobos has a mixed style when it comes to comments [1] [2] [3].

[1] https://github.com/dlang/phobos/blob/master/std/string.d#L2194-L2205
[2] https://github.com/dlang/phobos/blob/master/std/string.d#L2254-L2267
[3] https://github.com/dlang/phobos/blob/master/std/string.d#L2332-L2358

andralex: Let's use https://github.com/dlang/phobos/blob/master/std/string.d#L2254-L2267 for new code, thanks.

JinShil:
This PR needs to have the reviewers' concerns addressed and rebased.

This PR is nice, and normally I would revive such a PR if the author was no longer active.  I would also revive it if it were an important bug fix or something of higher priority.  This PR, however, is just a refactoring, so I'm going to put it in the [Phantom Zone](http://forum.dlang.org/post/ouuutodvhmnghzbeoqen@forum.dlang.org) and close it for now.

JinShil:
Resubmitted at #7749

PR #6683, opened 2017-04-10
[WIP] Create ASTBase family

RazvanN7:
This PR creates ASTBaremetal struct which is an AST family with minimal functionality (semantic analysis specific methods and fields are not implemented).

PR #4917, opened 2015-08-22
test ddmd with latest gdc/ldc and dmd-2.067.1 host compilers

MartinNowak:

PR #6836, opened 2017-05-27
Create parser as a library

RazvanN7:
Since the former branch (#6683) which compiled the parser as a library was very old and had a lot of conflicts it was easier to create a fresh new branch with all the changes. 

The windows parser target still needs to be added.


RazvanN7:
Any ideas on why the autotester is failing?

ZombineDev:
From the log:

> CC=c++ /home/braddr/sandbox/at-client/release-build/install/linux/bin32/dmd -of../generated/linux/release/32/dmd_frontend -m32 -vtls -J../generated/linux/release/32 -J../res -L-lstdc++ -version=MARS -wi -O -release -inline ddmd/access.d ddmd/aggregate.d ddmd/aliasthis.d ddmd/apply.d ddmd/argtypes.d ddmd/arrayop.d ddmd/arraytypes.d ddmd/astbase.d ddmd/astbasevisitor.d ddmd/astcodegen.d ddmd/attrib.d ddmd/builtin.d ddmd/canthrow.d ddmd/clone.d ddmd/complex.d ddmd/cond.d ddmd/constfold.d ddmd/cppmangle.d ddmd/ctfeexpr.d ddmd/dcast.d ddmd/dclass.d ddmd/declaration.d ddmd/delegatize.d ddmd/denum.d ddmd/dimport.d ddmd/dinifile.d ddmd/dinterpret.d ddmd/dmacro.d ddmd/dmangle.d ddmd/dmodule.d ddmd/doc.d ddmd/dscope.d ddmd/dstruct.d ddmd/dsymbol.d ddmd/dtemplate.d ddmd/dversion.d ddmd/escape.d ddmd/expression.d ddmd/func.d ddmd/hdrgen.d ddmd/impcnvtab.d ddmd/imphint.d ddmd/impvisitor.d ddmd/init.d ddmd/inline.d ddmd/inlinecost.d ddmd/intrange.d ddmd/json.d ddmd/lib.d ddmd/link.d ddmd/mars.d ddmd/mtype.d ddmd/nogc.d ddmd/nspace.d ddmd/objc.d ddmd/opover.d ddmd/optimize.d ddmd/parse.d ddmd/permissivevisitor.d ddmd/sapply.d ddmd/sideeffect.d ddmd/statement.d ddmd/staticassert.d ddmd/target.d ddmd/traits.d ddmd/visitor.d ddmd/transitivevisitor.d ddmd/typinf.d ddmd/utils.d ddmd/statement_rewrite_walker.d ddmd/statementsem.d ddmd/staticcond.d ddmd/strictvisitor.d ddmd/safe.d ddmd/blockexit.d ddmd/asttypename.d ddmd/libelf.d ddmd/scanelf.d ddmd/gluelayer.d ddmd/root/aav.d ddmd/root/array.d ddmd/root/ctfloat.d ddmd/root/file.d ddmd/root/filename.d ddmd/root/man.d ddmd/root/outbuffer.d ddmd/root/port.d ddmd/root/response.d ddmd/root/rmem.d ddmd/root/rootobject.d ddmd/root/speller.d ddmd/root/stringtable.d ddmd/root/hash.d ../generated/linux/release/32/newdelete.o ../generated/linux/release/32/lexer.a -version=NoBackend

> dmd: cppmangle.c:185: void CppMangleVisitor::source_name(Dsymbol*, bool): Assertion `t' failed.

> posix.mak:378: recipe for target '../generated/linux/release/32/dmd_frontend' failed

> make[1]: *** [../generated/linux/release/32/dmd_frontend] Aborted (core dumped)

It looks you've hit an unsupported case in the cpp mangling in the old compiler. Probably you're not hitting this problem locally, because you're using a newer version of dmd where the bug has already been fixed.

RazvanN7:
@ZombineDev Thanks for the explanation, but how do we move further from here? I need to wait for the auto-tester to get updated?

ZombineDev:
Unfortunately​, I don't think upgrading the host compiler (DMD v2.067.0) on the autotester is an option, though other guys from the core team (especially Daniel (@yebblies),  @MartinNowak and @WalterBright) should be able give you a more reliable answer. AFAIK, 2.067 was selected as the last stable version of frontend  (written in C++) used bootstrap the current frontend, and also because it was highest common version supported by all of the compilers (LDC and GDC, in addition to DMD). Currently building​ DMD purely from source is two step process - building DMD 2.067 with a C++ compiler and then using that to build current version. Upgrading the minimal required host compiler version would entail making that a three part process.
In the meantime, I suggest you install DMD 2.067 on your machine and try to find a workaround.

CyberShadow:
> The last C++ frontend version is 2.068.2.

Even though that's the last version with a C++ frontend, it's definitely not the last version that's buildable without a D compiler. D was used in smaller parts before the DDMD conversion was merged, e.g. for `idgen.d`. AFAIK 2.067.0 indeed is the last version buildable without a D compiler.

CyberShadow:
> Upgrading the minimal required host compiler version would entail making that a three part process.

FWIW, Digger can bootstrap DMD from C++ source through an unlimited number of levels if necessary, but I agree that adding additional versions to the bootstrap chain should be avoided unless absolutely necessary.

ZombineDev:
> You misremembered that – it was GDC that wasn't able to compile DDMD until much later on. LDC was able to build DDMD since Saturday after DConf 2015. ;)

@klickverbot  Sorry, you're right. There were a couple of ldc issues that I have misremembered (https://github.com/ldc-developers/ldc/issues/868, https://github.com/ldc-developers/ldc/issues/1141, https://github.com/ldc-developers/ldc/issues/1311, https://github.com/ldc-developers/ldc/issues/1379). As https://github.com/dlang/dmd/pull/4917 shows LDC had equal or greater ability to compile DMD (compared to GDC), at least since Aug 2015.

yebblies:
> Thanks for the explanation, but how do we move further from here? I need to wait for the auto-tester to get updated?

The simplest solution is to not use any features that aren't supported by 2.067.  Very few features are worth upgrading the required compiler version for.

RazvanN7:
@CyberShadow It seems that the cybershadow tester fails on some documentation target, but it tries to build one of my source files but failes. Any idea why this happens?

jacob-carlborg:
@RazvanN7 `dirEntries` returns a range of `DirEntry`, not a range of `string`. It has an `alias this` to automatically convert to a string, might be failing somehow. If you're only interested in the name/path you can try explicitly type the foreach-variable as `string`:

```d
foreach(string f ; dirEntries(...)) {
}
```

RazvanN7:
@jacob-carlborg That doesn't seem to be the problem. I will try to use the dirEntry methods to obtain the path as a string. Anyway, it is really odd how the cybershadow tester fails, while locally I have no problems running it. It may be possible that I have an older phobos version

RazvanN7:
@CyberShadow @jacob-carlborg now it says that writeln is not accessible from my module.

RazvanN7:
Fixed by commenting the '''writeln''' lines.

dejlek:
If you want to have a "parser as a library" why is it in ddmd? Should it not be in the Phobos somewhere perhaps (std.lang or std.compiler), or as completely separate project...

RazvanN7:
@dejlek The idea is that the compiler is going to build itself from the library

ibuclaw:
> Ah, right, thanks. I forgot that 2.068.2 was never officially made available as a C++ build for DMD, just for LDC.

This was the version that GDC had supported at the time (2.066).  On my request as I didn't want a situation where GDC couldn't build the latest frontend in order to build itself.  I guess we could relax that to being 2.068 now.

src/ddmd/astbase.d
```diff
@@ -0,0 +1,6299 @@
...
+    alias Dsymbols              = Array!(Dsymbol);
+    alias Objects               = Array!(RootObject);
+    alias Expressions           = Array!(Expression);
+    alias TemplateParameters    = Array!(TemplateParameter);
+    alias BaseClasses           = Array!(BaseClass*);
+    alias Parameters            = Array!(Parameter);
+    alias Statements            = Array!(Statement);
+    alias Catches               = Array!(Catch);
+    alias Identifiers           = Array!(Identifier);
+    alias Initializers          = Array!(Initializer);
```
andralex: But I assume you simply copied the existing formatting, so no worries for now.

src/ddmd/astbasevisitor.d
```diff
@@ -0,0 +1,1071 @@
+module ddmd.astbasevisitor;
+
+import ddmd.astbase;
+
+class Visitor
+{
+    void visit(ASTBase.Dsymbol)
```
andralex: There are a bunch of repetitive functions that are one-liners. I suggest you take the license at least for the code you introduce to write them one per line without separating whitespace, e.g.:

```d
void visit(ASTBase.Dsymbol) { assert(0); }
void visit(ASTBase.Dsymbol) { visit(cast(ASTBase.Dsymbol)s); }
void visit(ASTBase.Declaration s) { visit(cast(ASTBase.Dsymbol)s); }
...
```

We're looking at literally 5x LOC compression over large numbers of lines. That helps reviewers and contributors wading through code and also make certain patterns easier to see.

src/ddmd/permissivevisitor.d
```diff
@@ -0,0 +1,18 @@
...
+    alias visit = super.visit;
+
+    override void visit(ASTBase.Dsymbol){}
+    override void visit(ASTBase.Parameter){}
+    override void visit(ASTBase.Statement){}
+    override void visit(ASTBase.Type){}
+    override void visit(ASTBase.Expression){}
+    override void visit(ASTBase.TemplateParameter){}
+    override void visit(ASTBase.Condition){}
+    override void visit(ASTBase.Initializer){}
```
andralex: Like here!

src/ddmd/strictvisitor.d
```diff
@@ -0,0 +1,1074 @@
+module ddmd.strictvisitor;
+
+import ddmd.astbase;
+import ddmd.astbasevisitor;
+
+class StrictVisitor : Visitor
+{
+    alias visit = super.visit;
+
+    override void visit(ASTBase.Dsymbol)
```
andralex: ditto - zip it to 20% of the size!

src/ddmd/strictvisitor.d
```diff
@@ -0,0 +1,1074 @@
...
+    {
+        assert(0);
+    }
+
+    override void visit(ASTBase.AliasThis s)
```
andralex: Omit the parameter names when not used.

src/examples/avg.d
```diff
@@ -0,0 +1,68 @@
+/* This file contains an example on how to use the tranzitive visitor.
```
andralex: s/tranzitive/transitive/

src/examples/avg.d
```diff
@@ -0,0 +1,68 @@
...
+
+import std.stdio;
+import std.file;
+
+class FunctionLengthVisitor : TransitiveVisitor
+{
+    alias visit = super.visit;
+    ulong[] lengths;
+
+    float getAvgLen(ASTBase.Module m)
```
andralex: use double as the default fp number

src/examples/avg.d
```diff
@@ -0,0 +1,68 @@
...
+
+        if (lengths.length == 0)
+            return 0;
+
+        ulong sum = 0;
+        foreach (len; lengths)
+            sum += len;
+
+        //writeln(lengths);
+        return sum/lengths.length;
```
andralex: No need for the looparoo.

```d
import std.algorithm;
return double(lengths.sum) / length.length;
```

src/examples/avg.d
```diff
@@ -0,0 +1,68 @@
...
+    override void visitFuncBody(ASTBase.FuncDeclaration fd)
+    {
+        lengths ~= fd.endloc.linnum - fd.loc.linnum;
+        super.visitFuncBody(fd);
+    }
+}
+
+void main()
+{
+
```
andralex: delete empty line

src/examples/avg.d
```diff
@@ -0,0 +1,68 @@
...
+    string fname = "examples/testavg.d";
+
+    Id.initialize();
+    global._init();
+    global.params.isLinux = true;
+    global.params.is64bit = (size_t.sizeof == 8);
+    global.params.useUnitTests = true;
+    ASTBase.Type._init();
```
andralex: That's quite a lot of initialization needed. We should discuss how to reduce that.

src/examples/impvisitor.d
```diff
@@ -0,0 +1,46 @@
...
+import ddmd.astbase;
+import ddmd.permissivevisitor;
+import ddmd.transitivevisitor;
+
+import ddmd.tokens;
+import ddmd.root.outbuffer;
+
+import core.stdc.stdio;
+
+class ImportVisitor2: TransitiveVisitor
```
andralex: space before `:`

src/examples/impvisitor.d
```diff
@@ -0,0 +1,46 @@
...
+{
+    alias visit = super.visit;
+
+    override void visit(ASTBase.Import imp)
+    {
+        printf("import %s %s\n", imp.toChars(), imp.id.toChars());
```
andralex: generate parsable D code, will this build?

andralex (approved):
Good to go modulo nits. Thx!

src/ddmd/astbase.d
```diff
@@ -0,0 +1,6299 @@
...
+    import ddmd.identifier;
+    import ddmd.globals;
+    import ddmd.id;
+    import ddmd.errors;
+    import ddmd.lexer;
+
+    import core.stdc.string;
+    import core.stdc.stdarg;
+
+    alias Dsymbols              = Array!(Dsymbol);
```
RazvanN7: I took inspiration from src/ddmd/arraytypes.d. I wanted to have consistency with the rest of the compiler code.

src/examples/impvisitor.d
RazvanN7: no. it was just a proof of concept to see if it works

RazvanN7: but now it does.

andralex (approved):
nice!

src/ddmd/astbase.d
```diff
@@ -0,0 +1,6303 @@
+module ddmd.astbase;
+
+import ddmd.astbasevisitor;
+
+/** The ASTBase  family defines a family of AST nodes appropriate for parsing with
```
jacob-carlborg: Ddoc will interpet the first paragraph as the summary [1]. So I recommend splitting it up to get a summary paragraph, perhaps after the first sentence.

[1] http://dlang.org/spec/ddoc.html#summary

src/ddmd/astbase.d
```diff
@@ -0,0 +1,6303 @@
...
+    alias PROTpackage       = PROTKIND.PROTpackage;
+    alias PROTprotected     = PROTKIND.PROTprotected;
+    alias PROTpublic        = PROTKIND.PROTpublic;
+    alias PROTexport        = PROTKIND.PROTexport;
+    alias PROTundefined     = PROTKIND.PROTundefined;
+    alias PROTnone          = PROTKIND.PROTnone;
+
+    enum Sizeok : int
+    {
+        SIZEOKnone,             // size of aggregate is not yet able to compute
```
jacob-carlborg: Ideally all symbols should be documented but perhaps we can start by turning the existing comments, like these, to ddoc comments.

PR #6837, opened 2017-05-28
Replace idgen with CTFE implementation

jacob-carlborg:
Since id.h is not necessary anymore it's just as easy to implement the same functionality using CTFE. This will also simplify the build process because one extra build step is removed.

`extern(C++)` is removed as well because it will only be called from within D code.

This is required for #6771.


UplinkCoder:
@jacob-carlborg I am sorry to dismiss your work but honestly, this does not add significant value.
(But introduces breakage)

wilzbach:
> id.h is still used by LDC

Do you need this for the C++ backend? Or would this just mean that you need to upgrade your code base?

>  this does not add significant value.

I think this is great work as this hugely simplifies the build process and building DMD with dub (or any other build tool for that matter)!
@UplinkCoder: don't forget that this is a requirement for #6771.

dnadlinger (dismissed):
Even better than all this: Replace `idgen` completely by an `id!"foo"` template that initialises a member variable from `IdPool` using a static constructor.

Something like:

```d
template id(string name) {
  __gshared Identifier id; // __gshared as the frontend is single-threaded anyway.
  shared static this() {
    id = Identifier.idPool(name);
  }
}
```

and replace `Id.foo` by `id!"foo"`.

I'd rather not replace a crutch by another crutch.

UplinkCoder:
I don't see why we'd have to get rid of idgen.
It works.

WalterBright:
> this is great work as this hugely simplifies the build process and building DMD with dub (or any other build tool for that matter)!

Even back in the 80s, IDEs and magic modern `makefile` replacements never could handle subprojects. This seems to remain a problem.

WalterBright:
> I suppose we could manually maintain a pair of .h/.cpp files with the constants we need somewhere.

That pretty much kills this PR. I guarantee that manually maintaining those files will be overlooked and then some poor sap will expend hours trying to figure out the breakage. The whole point of `idgen` was to eliminate those bugs, even when I was the only one who ever touched the code.

Until `id.h` is no longer needed, I'm closing this PR.

yebblies:
@klickverbot How many uses of `Id` are there in the LDC glue?  How many are used frequently enough that replacing with `idPool` would be an issue?

WalterBright:
> I don't see any issue with using our own cache where required.

It seems to still be extra work for ldc and gdc. ping @ibuclaw

jacob-carlborg:
I was asked by @yebblies to do this here https://github.com/dlang/dmd/pull/6771#issuecomment-303030061.

jacob-carlborg:
@klickverbot If that's the whole list, then that's 54. Out of those only 19 exist in the DMD source code.

WalterBright:
I still want @ibuclaw 's buyin on this before pulling it.

jacob-carlborg:
> Even better than all this: Replace idgen completely by an id!"foo" template that initialises a member variable from IdPool using a static constructor.

@klickverbot is that the best idea, introducing a lost of static constructors. Also that approach would require quite a few other places to be updated to use the template instead.

ibuclaw:
I'm OK with this in principle.  Compilers should really use their own local identifier caches for symbols specific to themselves.

jacob-carlborg:
So, is there a preference how to do this? The current implementation or what David suggested https://github.com/dlang/dmd/pull/6837#pullrequestreview-40680647.

rainers:
If a C++ header is still needed it could also be generated by some `pragma(msg,generate(...))` when compiling id.d with `-version=gen_header`. The identifiers should remain `extern(C++)` in this case, though.

+1 for updating the Visual Studio projects, too.

wilzbach:
FWIW the recent DAutotest was caused by the auto-generated `id.d` file ([details](https://github.com/dlang/dlang.org/pull/1686)). So I am strongely in favor of getting rid of this auto-generated file - it just tends to make our complex build process more error-prone.

jacob-carlborg:
Rebased.

jacob-carlborg:
Why are all tests failing? I don't see the same problem with any other PRs.

CyberShadow:
> Why are all tests failing?

Well, *maybe* because this PR, as it is now, is introducing a regression? :)

jacob-carlborg:
> Well, maybe because this PR, as it is now, is introducing a regression? :)

I was surprised that it manged to get as far up the chain as to druntime. If anything it would have expected the compiler not to compile.

joakim-noah:
I'd like this to get in, always good to remove a dependency.  I just ran into dealing with this file when cross-compiling ldc for Android/ARM and rather than pull in another dependency- I'd need a D compiler to compile to linux/x64 in addition to the ldc cross-compiler I'm using- I just generated these two files and added them to a patch, seen at termux/termux-packages#1078.

jacob-carlborg:
The tests are green now. I'm pretty sure that the vibe.d failures are unrelated since other PRs have the same issue..

wilzbach:
> I'm pretty sure that the vibe.d failures are unrelated since other PRs have the same issue..

Yep it's a regression introduced in https://github.com/dlang/phobos/pull/5427

wilzbach:
Ok, everyone seems to be ok with moving forward, the only open question is:

> So, is there a preference how to do this? The current implementation or what David suggested https://github.com/dlang/dmd/pull/6837#pullrequestreview-40680647.

jacob-carlborg:
Can someone please comment on what @klickverbot suggested https://github.com/dlang/dmd/pull/6837#pullrequestreview-40680647. I want to know if anyone else agree with what David suggested. I would like to move on with this.

yebblies:
The problem with `id!"foo"` is that it won't protect against typos - which was part of the goal of the current design.  This change seems worth pulling with the current design.

jacob-carlborg:
What's the status of this? Can we get this merged or are any changes required?

wilzbach (approved):
> This change seems worth pulling with the current design.

I agree - this change is immensely useful for build tools down the road. Thanks a lot for pushing this!
I don't feel comfortable yet with signing off non-trivial PRs at DMD, but it seems that we have a consensus for going in this direction. So maybe @andralex or @WalterBright could help this PR to reach the finish line?

andralex:
@WalterBright and I are in favor of this big time. From the comments our understanding is the gdc and ldc responsibles are on board.

I think there's good value in this suggestion by @klickverbot:

>Even better than all this: Replace idgen completely by an id!"foo" template that initialises a member variable from IdPool using a static constructor.

There's liability too, so that work would be exploratory and does not need to come with this PR.

I'll mark this as approved and leave it to @jacob-carlborg to work out the tests. Coverage can be overridden (obviously) but I want to make sure jenkins has no problem.

andralex (approved):

wilzbach:
>  dlang-bot merged commit a3f0ea7 into dlang:master 33 minutes ago

In case someone is wondering, the bot merged this because only DAutoTest and auto-tester are required at the dmd repo. For more details:

https://github.com/dlang-bots/dlang-bot/issues/69

However, the failure at the project tester is an [unrelated spurious failure](https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-6837/5/pipeline).

CyberShadow:
This broke dlang.org auto-deploy. I think I know why (Digger uses idgen.d to identify D versions that need a host D compiler), but it shouldn't have passed DAutoTest then. Looking into it :)

jacob-carlborg:
Thanks.

PR #6839, opened 2017-05-28
fix Issue 17450 - escaping delegate context pointer not detected for …

WalterBright:
…member functions


BBasile:
possible regression: https://issues.dlang.org/show_bug.cgi?id=17456

PR #6840, opened 2017-05-29
set e.f in the CallExp's for _d_critical*

thewilsonator:
e.f is null otherwise and will speed up semantic validation and simplify logic rather than having to go through `(cast(FuncDeclaration)(cast(VarExp)e.e1).var)`.

### DIFF
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -3482,7 +3482,7 @@ else
         Expression genCall(Identifier id, Expression argexp, STC stc)
         {
             FuncDeclaration fd = FuncDeclaration.genCfunc(args, Type.tvoid, id, stc);
-            CallExp e = new CallExp(ss.loc, new VarExp(ss.loc, fdenter, false), argexp);
+            CallExp e = new CallExp(ss.loc, new VarExp(ss.loc, fd, false), argexp);
             e.f = fd;
             e.type = Type.tvoid; // do not run semantic on e
             return e;
@@ -3499,7 +3499,6 @@ else
 
         // set the explicit __critsecNNN alignment after semantic()
         if (!ss.exp) tmp.alignment = Target.ptrsize;
-
     }
 
     override void visit(WithStatement ws)
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -3499,7 +3499,7 @@ else
 
         // set the explicit __critsecNNN alignment after semantic()
         if (!ss.exp) tmp.alignment = Target.ptrsize;
-        }
+
     }
 
     override void visit(WithStatement ws)
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -3479,7 +3479,7 @@ else
             if (ss.exp) return new VarExp(ss.loc, tmp);
             else        return new DotIdExp(ss.loc, new VarExp(ss.loc, tmp), Id.ptr);
         }
-        Expression genCall(Identifier id, Expression argexp, STC stc = STC.undefined_)
+        Expression genCall(Identifier id, Expression argexp, STC stc)
         {
             FuncDeclaration fd = FuncDeclaration.genCfunc(args, Type.tvoid, id, stc);
             CallExp e = new CallExp(ss.loc, new VarExp(ss.loc, fdenter, false), argexp);
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -3480,7 +3480,7 @@ else
                 return new VarExp(ss.loc, tmp);
             else
             {
-                auto e = new DotIdExp(ss.loc, new VarExp(ss.loc, tmp), Id.ptr);
+                Expression e = new DotIdExp(ss.loc, new VarExp(ss.loc, tmp), Id.ptr);
                 e = e.expressionSemantic(sc);
                 return e;
             }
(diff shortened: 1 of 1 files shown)


WalterBright:
I don't see how it will speed up `CallExp.semantic()` as that function starts by setting `f` to `null`.

thewilsonator:
Well maybe not for `dmd` but I need it in LDC to do [1]
```
 void visit(CallExp *e) override {
    // SynchronizedStatement is lowered to
    //    Critsec __critsec105; // 105 == line number
    //    _d_criticalenter(& __critsec105); <--
    //    ...                                 |
    //    _d_criticalexit( & __critsec105);   |
    // So we intercept it with the CallExp ----

    if (e->f->ident == Id::criticalenter) {
      e->error("cannot use 'synchronized' in @compute code");
      stop = true;
      return;
    }

    if (e->f->ident == Id::criticalexit) {
      stop = true;
      return;
    }
    ...
}  
```

otherwise I have to do `(cast(FuncDeclaration*)(cast(VarExp*)e->e1)->var` if I want to access `e->f` and this also ensures that I never have to check if `e->f == nullptr` which will therefore crash if I haven't handled a case where `e->f` is null, which is a good thing.

[1]: https://github.com/ldc-developers/ldc/blob/master/gen/semantic-dcompute.cpp#L186

thewilsonator:
The same also applies to the `_d_monitor_e[nter|xit]` but classes are banned in dcompute anyway so that is no problem.

thewilsonator:
Also the frontend semantic analysis is bypassed anyway by directly setting the type.

thewilsonator:
@WalterBright ping. I have already done this for LDC. It would be nice to reduce the upstream diff for something trivial like this.

thewilsonator:
Actually I think it would make more sense to just always assign `f` in the constructor for `CallExp`.

andralex:
status of this?

thewilsonator:
Merged as is in LDC. Correcting the constructors is possible but is more likely to have a performance impact. Rebase and cover the case of `synchronized(o)` above is probably the better option given only this seems to be the only AST inconsistency that I have come across.

andralex:
@thewilsonator thanks! do you plan to look at the conflict?

thewilsonator:
Done. I don't know what was causing the conflict, it merged conflict free locally. *shrug*

andralex:
cc @WalterBright @MartinNowak please decide on this while still fresh

JinShil (dismissed):

thewilsonator:
The one autoteser failure was a network failure on the BSD 32 bit. 
It'd be nice to merge this to reduce the LDC diff.

JinShil:
Please rebase and force push.  It will force the auto-tester to run again.

JinShil:
@thewilsonator I know you're going to hate me, but I couldn't go through with it.  It just seems to brittle.

> Actually I think it would make more sense to just always assign `f` in the constructor for `CallExp`.

What would it take to do that?  I think that would make me more comfortable putting my name next to the "merge" button.

thewilsonator:
Changes to multiple constructors in `UnaExp` because `CallExp` just forwards its ctors. You'd then have to update anywhere a UnaExp should be a CallExp. 

TL;DR more code churn for no gain.

JinShil:
It appears you only need to add a single constructor to `CallExp`, because `f` is a member of `CallExp`.  I just implemented it myself, and it was only an additional 6 lines.

thewilsonator:
Cool! code please ;)

The point of this pull is to reduce the diff between LDC and DMD. I don't care how's it done.
If we have to update it once that's fine, its the languishing and causing merge conflicts that's the problem. Feel free to close this and open yours, or just edit this one.

JinShil:
Also, you might need to add a corresponding `create` method and add it to `expression.h`.

thewilsonator:
That was why I decided to go with the simple approach of setting the fields directly. Adding a constructor is fine but it isn't used much except by the `create` function which are static, which requires updating the classname that is used everywhere else in the codebase i think. As is currently this is constrained to a single file, adding and appropriately using a `create` changes a large number of files.

JinShil:
I hear you, but simple, or fewer lines isn't necessarily best.  IMO, it would better to make the constructors rather than the silly casting of an expression in the immediate following line just to set a property.  

Also, I'm not the only one here, if others disagree with me, they are more than welcome to merge this PR.  But if it's going to be my name there next to the merge notice, I need to have more confidence in the implementation.

ibuclaw:
Or couldn't there be a helper function called `genCfuncCall` - just noting that every location updated is doing the exact same thing -> possibile refactoring opportunity.

thewilsonator:
Hmm? Why is that still failing?
`dmd/statementsem.d(3970): Error: unrecognized declaration`

`error(c.loc, "cannot put catch statement inside `%s`", Token.toChars(sc.os.tok));`
The offending line is nowhere near the code I touched and I balanced my braces.
(probably smth. very obvious but am jetlagged atm)

thewilsonator:
@ibuclaw @JinShil Refactored. Didn't reduce the line count by as much as I'd hoped, but its much easier to follow.

JinShil:
@thewilsonator, The refactoring is rather gratuitous.  I've submitted an alternative at #7916 for your consideration.

thewilsonator:
Closing in favour of #7916

PR #6842, opened 2017-05-29
better error message for returning escapes

WalterBright:


mathias-lang-sociomantic:
If the test suite doesn't need any change, we have a problem :)

PR #6847, opened 2017-05-31
fix Issue 17457 - Named Return Value Optimization (NRVO) not done wit…

WalterBright:
…h return of constructor call


WalterBright:
This should improve performance by eliminating an unnecessary struct move.

andralex (approved):

PR #6849, opened 2017-06-01
generate fewer __inlineretval temps

WalterBright:
Fewer temps makes the code go faster, as the results are written through to the final output rather than copied.


src/ddmd/inline.d
```diff
@@ -281,7 +281,16 @@ public:
                   // 'return' must not leave in the expand result
             return;
         static if (asStatements)
-            result = new ReturnStatement(s.loc, exp);
+        {
```
UplinkCoder: unrelated change

yebblies:
This is pretty ugly but also nothing new for the inliner.

PR #6852, opened 2017-06-03
replace construct-copy with a construct directly into target

WalterBright:
This is another case of https://issues.dlang.org/show_bug.cgi?id=17457

### DIFF
--- a/src/ddmd/e2ir.d
+++ b/src/ddmd/e2ir.d
@@ -2898,6 +2898,27 @@ elem *toElem(Expression e, IRState *irs)
                     e = toElem(ae.e2, irs);
                     goto Lret;
                 }
+
+                /* Look for:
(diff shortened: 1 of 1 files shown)


WalterBright:
Closely related:
https://github.com/dlang/dmd/pull/6849
https://github.com/dlang/dmd/pull/6847

mathias-lang-sociomantic:
No test ?

WalterBright:
It's implicitly tested by all the use cases already out there. I do have a more specific test case, but it requires all three of the PRs to be pulled first. I still want the PRs to be separate, however, in case one of them breaks something.

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=17632

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=17622

PR #6855, opened 2017-06-04
DIP1003: Remove body as a Keyword

WalterBright:
https://github.com/dlang/DIPs/blob/master/DIPs/DIP1003.md

It doesn't give a warning or deprecation yet over use of `body`.

### DIFF
--- a/src/ddmd/parse.d
+++ b/src/ddmd/parse.d
@@ -780,7 +780,10 @@ final class Parser(AST) : Lexer
 
                 /* Look for return type inference for template functions.
                  */
-                if (token.value == TOKidentifier && skipParens(peek(&token), &tk) && skipAttributes(tk, &tk) && (tk.value == TOKlparen || tk.value == TOKlcurly || tk.value == TOKin || tk.value == TOKout || tk.value == TOKbody))
+                if (token.value == TOKidentifier && skipParens(peek(&token), &tk) && skipAttributes(tk, &tk) &&
+                    (tk.value == TOKlparen || tk.value == TOKlcurly || tk.value == TOKin ||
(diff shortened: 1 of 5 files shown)


UplinkCoder (approved):
nice one

src/ddmd/parse.d
```diff
@@ -6618,13 +6629,19 @@ final class Parser(AST) : Lexer
                 }
                 return false;
 
+            case TOKidentifier:
```
andralex: these lines appear not covered

andralex (approved):
Please take a look at coverage. Thx!

src/ddmd/parse.d
WalterBright: done

WalterBright:
Auto-merge toggled on

MetaLang:
Why was there no corresponding changelist entry for this? I just looked and there is no reference in the changelog to allowing `do`.

ntrel:
@MetaLang I've added a related changelog entry in #7384, so at least the DIP will be mentioned in the next one.

test/fail_compilation/failcontracts.d
```diff
@@ -0,0 +1,24 @@
...
+fail_compilation/failcontracts.d(22): Error: unexpected ( in declarator
+fail_compilation/failcontracts.d(22): Error: found 'T' when expecting ')'
+fail_compilation/failcontracts.d(22): Error: enum declaration is invalid
+fail_compilation/failcontracts.d(22): Error: found ')' instead of statement
+---
+*/
+
+void test()
+{
+    auto f1 = function() bode;
```
ibuclaw: @WalterBright is this deliberate?  Or did you actually misspell `body`?

PR #6856, opened 2017-06-04
fix Issue 17468 - Internal Assertion fails during CTFE

UplinkCoder:
This error happened when because a slice of a pointer is given.
The code used to go into the pointer specific branch based on the type.
However the slice specific branch is the right choice.

I changed the order of the query
I doubt this will break anything

### DIFF
--- a/test/compilable/test17468.d
+++ b/test/compilable/test17468.d
@@ -1,0 +1,12 @@
+// PERMUTE_ARGS:
+struct S
+{
+        const char* path;
+        @disable this();
+        this(const(char)* path)
+        {
+                this.path = path;
+        }
+}
+const S CONST_S = S("/tmp".ptr);
+
--- a/src/ddmd/ctfeexpr.d
+++ b/src/ddmd/ctfeexpr.d
@@ -398,23 +398,6 @@ extern (C++) UnionExp copyLiteral(Expression e)
         r.type = e.type;
         return ue;
     }
-    if (isPointer(e.type))
-    {
-        // For pointers, we only do a shallow copy.
-        if (e.op == TOKaddress)
-            emplaceExp!(AddrExp)(&ue, e.loc, (cast(AddrExp)e).e1);
-        else if (e.op == TOKindex)
-            emplaceExp!(IndexExp)(&ue, e.loc, (cast(IndexExp)e).e1, (cast(IndexExp)e).e2);
-        else if (e.op == TOKdotvar)
-        {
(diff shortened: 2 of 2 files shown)


UplinkCoder:
@WalterBright @MartinNowak @andralex 
quick-merge please :)

jacob-carlborg:
Do we need a new test?

andralex:
Code is not covered, please add a test. Thanks!

dnadlinger (changes requested):
Yes, please add a test.

UplinkCoder:
@klickverbot done

UplinkCoder:
Although ... coverage should not have changed by a code move ....

dnadlinger (approved):

test/compilable/test17468.d
```diff
@@ -0,0 +1,11 @@
+struct S
```
klickverbot: Maybe add `// PERMUTE_ARGS:` as this is CTFE-only?

UplinkCoder: what does it do ?

wilzbach:
> @wilzbach @MartinNowak: Do squash-merges need to be performed manually again, or has the whole "too dangerous" argument been refuted?

@MartinNowak removed the "auto-merge-squash" label after @andralex tried to squash a PR, which wasn't supposed to be merged.
See also this discussion:

https://github.com/dlang-bots/dlang-bot/issues/64#issuecomment-284160756

FWIW
- it could be simply re-enabled by just adding the label "auto-merge-squash"
- I'm of the opinion that we all are adults and can reasonably judge when to use "auto-merge-squash" and when not to use ...

wilzbach:
Btw @UplinkCoder I thought that all bug fixes should be targeted at `stable`?

UplinkCoder:
closed for similar version targeting stable.

jacob-carlborg:
> Although ... coverage should not have changed by a code move ....

@UplinkCoder  Since a bug was discovered it means that that part of the code had not enough coverage. The coverage might not change when moving code but that doesn't mean that there was enough coverage from the beginning.

MartinNowak:
> @wilzbach @MartinNowak: Do squash-merges need to be performed manually again, or has the whole "too dangerous" argument been refuted?

The bot currently does not support (technically) squash merge commits. It will simply dismiss any commit messages at all, see https://github.com/dlang-bots/dlang-bot/issues/64#issuecomment-289068950.
Let's please try to keep such OT discussions out of PR reviews.

MartinNowak:
> closed for similar version targeting stable.

You can change the base branch of a PR @UplinkCoder, much better than reopening a new PR.

PR #6857, opened 2017-06-04
runnable/template1: Do not hardcode long type alignment

dnadlinger:
The alignment of D's long type is not what the test is supposed
to verify, so make it less brittle (relevant e.g. for AArch64).

### DIFF
--- a/test/runnable/template1.d
+++ b/test/runnable/template1.d
@@ -1452,14 +1452,7 @@ void test60()
         printf("thing1.sizeof: %u\n", thing1.sizeof);
         printf("thing2.sizeof: %u\n", thing2.sizeof);
 
-        version (Win32)
-            assert(thing1.sizeof == 16);
-        else version (X86_64)
-            assert(thing1.sizeof == 16);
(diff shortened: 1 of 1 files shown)


PetarKirov (approved):

WalterBright:
Why not just put in the condition for AArch64? After all, it's not like it's going to change for AArch64.

ZombineDev:
LGTM, but can you clarify why do you think AArch64 is different? (Context: I've recently started working on compiling LDC on an Android/AArch64 device). According to [AAPCS64](http://infocenter.arm.com/help/topic/com.arm.doc.ihi0055b/IHI0055B_aapcs64.pdf) Signed double-word (aka D's long) are 8 bytes sized/aligned.

WalterBright:
> but would need to be explicitly added to the list as it falls back to the wrong default.

Having the default not be `static assert(0);` is a mistake.

> Instead, I chose to fix it more generally (assuming that longs will always be aligned >= int.sizeof).

I can live with that.

PR #6859, opened 2017-06-04
fix Issue 17468 - Internal Assertion fails during CTFE

UplinkCoder:
This error happened when because a slice of a pointer is given.
The code used to go into the pointer specific branch based on the type.
However the slice specific branch is the right choice.

I changed the order of the query
I doubt this will break anything

@wilzbach  re-targeted #6856

### DIFF
--- a/src/ddmd/ctfeexpr.d
+++ b/src/ddmd/ctfeexpr.d
@@ -398,23 +398,6 @@ extern (C++) UnionExp copyLiteral(Expression e)
         r.type = e.type;
         return ue;
     }
-    if (isPointer(e.type))
-    {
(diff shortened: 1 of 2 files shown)


wilzbach:
> @wilzbach re-targeted #6856

Thanks!
FYI: you don't need to create a new PR, you can simply select select the target branch:

image

and then rebase the PR ;-)
For more info, see also the [blog article from GH](https://github.com/blog/2224-change-the-base-branch-of-a-pull-request).

UplinkCoder:
@wilzbach I don't try anything clever with git.
You never know what will explode

andralex (approved):

PR #6860, opened 2017-06-04
CodeBuilder: outblkexitcode()

WalterBright:
`outblkexitcode()` is quite a snarl, this straightens it out somewhat.


UplinkCoder:
@WalterBright is `outblkexitcode` conceptually the same as `scope(exit)` ?
If so, could you add a comment stating it ?

WalterBright:
> is outblkexitcode conceptually the same as scope(exit) ?

No, they have nothing to do with each other.

UplinkCoder:
@WalterBright what is it then ?

yebblies:
@UplinkCoder It generates instructions to move between basic blocks.

PR #6862, opened 2017-06-06
win32.mak: fix target checkwhitespace

rainers:
Current version fails if the tools folder already exists:
```
git clone --depth=1 https://github.com/dlang/tools ..\..\tools
fatal: destination path '..\..\tools' already exists and is not an empty directory.
```

Also removes references to non-existing files and adds more files to be checked.

### DIFF
--- a/src/win32.mak
+++ b/src/win32.mak
@@ -212,15 +212,15 @@ ROOT_SRCS=$(ROOT)/aav.d $(ROOT)/array.d $(ROOT)/ctfloat.d $(ROOT)/file.d \
 SRCS = $D/aggregate.h $D/aliasthis.h $D/arraytypes.h	\
 	$D/attrib.h $D/complex_t.h $D/cond.h $D/ctfe.h $D/ctfe.h $D/declaration.h $D/dsymbol.h	\
 	$D/enum.h $D/errors.h $D/expression.h $D/globals.h $D/hdrgen.h $D/identifier.h $D/idgen.d	\
-	$D/import.h $D/init.h $D/intrange.h $D/json.h $D/lexer.h	\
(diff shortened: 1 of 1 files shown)


WalterBright (approved):

CyberShadow:
> Current version fails if the tools folder already exists:

Why would the tools directory exist but not contain a `checkwhitespace.d`? An old checkout? In which case, allowing the `git` command to fail is not going to help move things along, will it?

rainers:
> Why would the tools directory exist but not contain a checkwhitespace.d?

The git command is executed unconditionally by DM make and fails if the directory exists and is non-empty. Ignoring the result forwards the error checking to the compilation of checkwhitespace.d: if it actually doesn't exist dmd will complain.

CyberShadow:
> The git command is executed unconditionally by DM make and fails if the directory exists and is non-empty.

Huh, is that so? Since the `$(TOOLS_DIR)\checkwhitespace.d` rule has no dependencies, wouldn't it be executed iff the file doesn't exist?

rainers:
Every other make behaves as expected, but not the Digital Mars version:
```
c:\tmp\make>type Makefile

Makefile:
        echo $@ does not exist!?

c:\tmp\make>c:\l\d\dmd2.074\dmd2\windows\bin\make.exe
echo Makefile does not exist!?
Makefile does not exist!?

c:\tmp\make>make
make: `makefile' is up to date.

c:\tmp\make>nmake

Microsoft (R) Program Maintenance Utility Version 12.00.21005.1
Copyright (C) Microsoft Corporation.  All rights reserved.

'Makefile' is up-to-date
```

CyberShadow:
Huh. Well, thanks for the clarification.

PR #6864, opened 2017-06-06
fix Issue 17432 - scope delegates change type, but not mangling

rainers:
This does not add "scope" to .mangleof or .stringof if it was inferred

### DIFF
--- a/src/ddmd/mtype.h
+++ b/src/ddmd/mtype.h
@@ -598,6 +598,7 @@ public:
     bool isref;         // true: returns a reference
     bool isreturn;      // true: 'this' is returned by ref
     bool isscope;       // true: 'this' is scope
+    bool isscopeinferred; // true: 'this' is scope from inferrence
     LINK linkage;  // calling convention
     TRUST trust;   // level of trust
     PURE purity;   // PURExxxx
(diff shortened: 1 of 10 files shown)


rainers:
My first attempt was to infer scope always, but not report any related errors. This causes much larger changes, but also runs into https://issues.dlang.org/show_bug.cgi?id=17449 as the info from the escape analysis is still applied even without -dip1000.

UplinkCoder (approved):
This looks very good to me. :+1: 
I was struggling with this bug.
Glad to see it's resolved.

UplinkCoder:
Travis only fails due to timeout.
@WalterBright I thinnk this is good to go.

src/ddmd/mtype.h
```diff
@@ -598,6 +598,7 @@ class TypeFunction : public TypeNext
     bool isref;         // true: returns a reference
     bool isreturn;      // true: 'this' is returned by ref
     bool isscope;       // true: 'this' is scope
+    bool isscopeinferred; // true: 'this' is scope from inferrence
```
WalterBright: inference

rainers:
Thanks for merging. Unfortunately a few issues still showed up, see https://github.com/dlang/dmd/pull/6876

PR #6866, opened 2017-06-07
Add some Ddoc comments to dstruct.d

WalterBright:
Given https://dlang.org/phobos/ddmd_dstruct.html , I redid a few comment blocks to be more in Ddoc style.

PR #6867, opened 2017-06-07
Unify comment header and expose module with Ddox

wilzbach:
I noticed in https://github.com/dlang/dmd/pull/6866 that a few modules aren't visible on the DDoX build.
It turns out that Ddox only shows modules with a DDoc module header (e.g. `/**` instead `/*`). I used this opportunity to unify the comment format to the Ddoc style.

### DIFF
--- a/src/ddmd/backend/dwarf.d
+++ b/src/ddmd/backend/dwarf.d
@@ -1,4 +1,4 @@
-/* Dwarf debug
+/** Dwarf debug
  *
  * Source: $(DMDSRC backend/_dwarf.d)
  */
--- a/src/ddmd/nspace.h
+++ b/src/ddmd/nspace.h
@@ -1,10 +1,11 @@
-
-// Compiler implementation of the D programming language
-// Copyright: Copyright (c) 2014 by Digital Mars, All Rights Reserved
-// Authors: Walter Bright, http://www.digitalmars.com
-// License: http://boost.org/LICENSE_1_0.txt
-// Source: https://github.com/dlang/dmd/blob/master/src/nspace.h
-
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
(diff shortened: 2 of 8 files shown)


src/ddmd/backend/dwarf.d
```diff
@@ -1,4 +1,4 @@
-/* Dwarf debug
+/** Dwarf debug
  *
  * Source: $(DMDSRC backend/_dwarf.d)
  */
```
wilzbach: I know that it's a debug file, but I wasn't sure whether it was set "private" on purpose.
It's currently visible on the Ddoc build:

https://dlang.org/phobos/ddmd_backend_dwarf.html

But I guess we should simply add it to the [blacklist of hidden modules](https://github.com/dlang/dlang.org/blob/master/posix.mak#L73).

src/ddmd/nspace.h
```diff
@@ -1,10 +1,11 @@
...
-// Source: https://github.com/dlang/dmd/blob/master/src/nspace.h
-
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Copyright:   Copyright (c) 1999-2017 by Digital Mars, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ */
```
wilzbach: I know that this is a C header file and not part of our documentation, but there were only two header files using this comment header style, so I thought unifying it makes sense. At some point, we will probably generate these headers anyways...

src/ddmd/backend/dwarf.d
ZombineDev: > I know that it's a debug file

It's for generation of [DWARF debug information](https://en.wikipedia.org/wiki/DWARF), so it's definitely not a utility file - it needs to be part of the documentation.

PetarKirov (approved):

PR #6868, opened 2017-06-07
Add some Ddoc comments to expression.d

WalterBright:


src/ddmd/expression.d
```diff
@@ -515,6 +526,8 @@ extern (C++) Expression resolveProperties(Scope* sc, Expression e)
 
 /******************************
  * Check the tail CallExp is really property function call.
+ * Bugs:
+ * This doesn't appear to do anything.
```
wilzbach: Why don't you remove it then? :scream:

wilzbach (approved):
LGTM

src/ddmd/expression.d
WalterBright: Because this PR is about adding documentation. It's good practice to avoid "mission creep" in PRs.

PR #6869, opened 2017-06-08
Fix deprecations in d_do_test

wilzbach:
The old string functions from D1 have been [deprecated](https://github.com/dlang/phobos/pull/5378), time to fix the deprecations.

### DIFF
--- a/test/d_do_test.d
+++ b/test/d_do_test.d
@@ -405,6 +405,8 @@ bool collectExtraSources (in string input_dir, in string output_dir, in string[]
 // marked by $n$ that contain compiler generated unique numbers
 bool compareOutput(string output, string refoutput)
 {
+    import std.ascii : digits;
+    import std.utf : byCodeUnit;
     for ( ; ; )
     {
         auto pos = refoutput.indexOf("$n$");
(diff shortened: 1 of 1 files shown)


test/d_do_test.d
```diff
@@ -416,7 +418,8 @@ bool compareOutput(string output, string refoutput)
             return false;
         refoutput = refoutput[pos + 3 ..$];
         output = output[pos..$];
-        munch(output, "0123456789");
+        auto p = output.byCodeUnit.countUntil!(e => digits.canFind(e));
```
wilzbach: In theory `countUntil("0", "1", "2", "3", "4", "5", "6", "7", "8", "9")` might be marginally faster, but it shouldn't matter at all here.

CyberShadow:
@wilzbach What's causing the coverage changes? https://codecov.io/gh/dlang/dmd/compare/d982fe83480e69f9fd0d39dfd36d20903f497a36...810e57ee216379f8165bcafed43afea8403c6f33/changes

wilzbach:
> @wilzbach What's causing the coverage changes? 

No idea, but when I looked at other PRs I did find some repeatedly occurring random coverage changes, so I think it's best if we go with a similar pragmatic solution as for Phobos:

https://github.com/dlang/dmd/pull/6879

PR #6870, opened 2017-06-08
Use generated DMD binary instead of src/dmd

wilzbach:
CC @RazvanN7

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -85,6 +85,7 @@ endif
 include ../src/osmodel.mak
 
 export OS
+BUILD=release
 
 ifeq (freebsd,$(OS))
     SHELL=/usr/local/bin/bash
@@ -100,7 +101,6 @@ export REQUIRED_ARGS=
 
 ifeq ($(findstring win,$(OS)),win)
 export ARGS=-inline -release -g -O
-export DMD=../src/dmd.exe
 export EXE=.exe
 export OBJ=.obj
 export DSEP=\\
@@ -110,21 +110,40 @@ DRUNTIME_PATH=..\..\druntime
 PHOBOS_PATH=..\..\phobos
 export DFLAGS=-I$(DRUNTIME_PATH)\import -I$(PHOBOS_PATH)
 export LIB=$(PHOBOS_PATH)
+
+# auto-tester might run the testsuite with a different $(MODEL) than DMD
+# has been compiled with. Hence we manually check which binary exists.
+# For windows the $(OS) during build is: `windows`
+ifeq (,$(wildcard ../generated/windows/$(BUILD)/64/dmd$(EXE)))
+DMD_MODEL=32
+else
+DMD_MODEL=64
+endif
+export DMD=../generated/windows/$(BUILD)/$(DMD_MODEL)/dmd$(EXE)
+
 else
 export ARGS=-inline -release -g -O -fPIC
-export DMD=../src/dmd
 export EXE=
 export OBJ=.o
 export DSEP=/
 export SEP=/
 
+# auto-tester might run the testsuite with a different $(MODEL) than DMD
+# has been compiled with. Hence we manually check which binary exists.
+ifeq (,$(wildcard ../generated/$(OS)/$(BUILD)/64/dmd))
+DMD_MODEL=32
+else
+DMD_MODEL=64
+endif
+export DMD=../generated/$(OS)/$(BUILD)/$(DMD_MODEL)/dmd
+
 DRUNTIME_PATH=../../druntime
 PHOBOS_PATH=../../phobos
 # link against shared libraries (defaults to true on supported platforms, can be overridden w/ make SHARED=0)
 SHARED=$(if $(findstring $(OS),linux freebsd),1,)
-DFLAGS=-I$(DRUNTIME_PATH)/import -I$(PHOBOS_PATH) -L-L$(PHOBOS_PATH)/generated/$(OS)/release/$(MODEL)
+DFLAGS=-I$(DRUNTIME_PATH)/import -I$(PHOBOS_PATH) -L-L$(PHOBOS_PATH)/generated/$(OS)/$(BUILD)/$(MODEL)
 ifeq (1,$(SHARED))
-DFLAGS+=-defaultlib=libphobos2.so -L-rpath=$(PHOBOS_PATH)/generated/$(OS)/release/$(MODEL)
+DFLAGS+=-defaultlib=libphobos2.so -L-rpath=$(PHOBOS_PATH)/generated/$(OS)/$(BUILD)/$(MODEL)
 endif
 export DFLAGS
 endif


test/Makefile
```diff
@@ -96,7 +97,7 @@ export REQUIRED_ARGS=
 
 ifeq ($(findstring win,$(OS)),win)
 export ARGS=-inline -release -g -O
-export DMD=../src/dmd.exe
+export DMD=..\generated\$(OS)\$(MODEL)\dmd.exe
```
braddr: This path should include $(BUILD) as well.  It's a mistake that the current windows make files don't put the binary in the right place and this change shouldn't perpetuate the error.

RazvanN7:
#6873 fixes the windows makefile, currently hardcoding BUILD to release, so you can go ahead and use generated\$OS\$BUILD\$MODEL after that one is merged

wilzbach:
Hmm it seems like the auto-tester picks up `MODEL=64` for this test Makefile, even though it's building `dmd` with `MODEL=32`...

```
CC=c++ /home/braddr/sandbox/at-client/release-build/install/linux/bin32/dmd -of../generated/linux/release/32/dmd -m32 -vtls -J../generated/linux/release/32 -J../res -L-lstdc++ -version=MARS -wi -O -release -inline ddmd/access.d ddmd/aggregate.d ddmd/aliasthis.d ddmd/apply.d ddmd/argtypes.d ddmd/arrayop.d ddmd/arraytypes.d ddmd/astcodegen.d ddmd/attrib.d ddmd/builtin.d ddmd/canthrow.d ddmd/clone.d ddmd/complex.d ddmd/cond.d ddmd/constfold.d ddmd/cppmangle.d ddmd/ctfeexpr.d ddmd/dcast.d ddmd/dclass.d ddmd/declaration.d ddmd/delegatize.d ddmd/denum.d ddmd/dimport.d ddmd/dinifile.d ddmd/dinterpret.d ddmd/dmacro.d ddmd/dmangle.d ddmd/dmodule.d ddmd/doc.d ddmd/dscope.d ddmd/dstruct.d ddmd/dsymbol.d ddmd/dtemplate.d ddmd/dversion.d ddmd/escape.d ddmd/expression.d ddmd/func.d ddmd/hdrgen.d ddmd/impcnvtab.d ddmd/imphint.d ddmd/init.d ddmd/inline.d ddmd/inlinecost.d ddmd/intrange.d ddmd/json.d ddmd/lib.d ddmd/link.d ddmd/mars.d ddmd/mtype.d ddmd/nogc.d ddmd/nspace.d ddmd/objc.d ddmd/opover.d ddmd/optimize.d ddmd/parse.d ddmd/sapply.d ddmd/sideeffect.d ddmd/statement.d ddmd/staticassert.d ddmd/target.d ddmd/traits.d ddmd/visitor.d ddmd/typinf.d ddmd/utils.d ddmd/statement_rewrite_walker.d ddmd/statementsem.d ddmd/staticcond.d ddmd/safe.d ddmd/blockexit.d ddmd/asttypename.d ddmd/printast.d ddmd/libelf.d ddmd/scanelf.d ddmd/irstate.d ddmd/toctype.d ddmd/glue.d ddmd/gluelayer.d ddmd/todt.d ddmd/tocsym.d ddmd/toir.d ddmd/dmsc.d ddmd/tocvdebug.d ddmd/s2ir.d ddmd/toobj.d ddmd/e2ir.d ddmd/eh.d ddmd/iasm.d ddmd/objc_glue_stubs.d ddmd/backend/bcomplex.d ddmd/backend/cc.d ddmd/backend/cdef.d ddmd/backend/cgcv.d ddmd/backend/code.d ddmd/backend/cv4.d ddmd/backend/dt.d ddmd/backend/el.d ddmd/backend/global.d ddmd/backend/obj.d ddmd/backend/oper.d ddmd/backend/outbuf.d ddmd/backend/rtlsym.d ddmd/backend/code_x86.d ddmd/backend/iasm.d ddmd/backend/ty.d ddmd/backend/type.d ddmd/backend/exh.d ddmd/backend/mach.d ddmd/backend/md5.d ddmd/backend/mscoff.d ddmd/backend/dwarf.d ddmd/backend/dwarf2.d ddmd/backend/xmm.d ddmd/tk/dlist.d ddmd/root/aav.d ddmd/root/man.d ddmd/root/response.d ddmd/root/speller.d ../generated/linux/release/32/newdelete.o ../generated/linux/release/32/backend.a ../generated/linux/release/32/lexer.a
cp ../generated/linux/release/32/dmd .
make[1]: Leaving directory '/home/braddr/sandbox/at-client/pull-2605257-Linux_32_64/dmd/src'
make -C src -f posix.mak
make[1]: Entering directory '/home/braddr/sandbox/at-client/pull-2605257-Linux_32_64/dmd/src'
no cpu specified, assuming X86

...

make -C ../src -f posix.mak ../generated/linux/release/64/dmd
Creating output directory: generated
make[2]: Entering directory '/home/braddr/sandbox/at-client/pull-2605257-Linux_32_64/dmd/test'
make[2]: warning: jobserver unavailable: using -j1.  Add '+' to parent make rule.
no cpu specified, assuming X86
posix.mak:86: *** 'dmd' not found, get a D compiler or make AUTO_BOOTSTRAP=1.  Stop.
make[2]: Leaving directory '/home/braddr/sandbox/at-client/pull-2605257-Linux_32_64/dmd/src'
Makefile:210: recipe for target '../generated/linux/release/64/dmd' failed
make[1]: *** [../generated/linux/release/64/dmd] Error 2
```

@RazvanN7 @CyberShadow any ideas how to deal with this (without ugly workarounds like checking for file existence)?

RazvanN7:
After reading @braddr 's scripts, I think that the problem is due to the mismatch in determining the model. While osmodel.mak uses "uname -m" to determine the model, the autotester uses the name of the machine (which in this case is linux_32_64). One solution might be to make the autotester use osmodel.mak, but I'm going to deffer this to Brad.

wilzbach:
@RazvanN7 I think I found a nice workaround at https://github.com/dlang/dmd/pull/6999:

```
ifeq (,$(wildcard ../generated/$(OS)/$(BUILD)/64/dmd))
REAL_MODEL=32
else
REAL_MODEL=64
endif
export DMD=../generated/$(OS)/$(BUILD)/$(REAL_MODEL)/dmd
```

test/Makefile
```diff
@@ -81,6 +81,7 @@ endif
 include ../src/osmodel.mak
 
 export OS
+BUILD=release
```
ZombineDev: Why is this necessary? `debug` is a better choice for testing in general and tests for specific optimization arguments, such as `-release`, do so via `PERMUTE_ARGS`. This variable affects only the test runner, right?

wilzbach: > Why is this necessary? 

Because you need to find the DMD binary somehow and as Makefile variable at least the user can overwrite it...

> debug is a better choice for testing in general and tests for specific optimization arguments

Hehe `debug` is simply a different folder as of now:
https://github.com/dlang/dmd/pull/6898

> This variable affects only the test runner, right?

Yes.

ZombineDev: > Because you need to find the DMD binary somehow and as Makefile variable at least the user can overwrite it...

I mean, is there s reason for preferring `release` over `debug`? If not I think it would be better to stick with `debug`.

wilzbach: > I mean, is there a reason for preferring release over debug? 

Hehe I agree, but `release` is the default and thus used by auto-tester.

wilzbach:
@RazvanN7 @ZombineDev @braddr FWIWI this is finally green, but I really don't like this hack. I think this is due to the fact that we use a different `MODEL` to build the compiler than for testing it?
(see below)

https://github.com/braddr/at-client/blob/master/src/do_build_dmd.sh
```
$makecmd MAKE=$makecmd MODEL=$COMPILER_MODEL
```

https://github.com/braddr/at-client/blob/master/src/do_test_dmd.sh
```
$makecmd MODEL=$OUTPUT_MODEL
```

FYI envs are as follows:
https://github.com/braddr/at-client/blob/master/src/setup_env.sh

```
COMPILER_MODEL=32
OUTPUT_MODEL=32
...
case "$1" in
...
    Linux_32_64)
        OUTPUT_MODEL=64
        ;;
    Linux_64_32)
        COMPILER_MODEL=64
```

So this leaves us the following options:

1) We go with an detection approach like this (sth. like `ls -1 ../dmd/generated/$(OS)/$(BUILD)/*/dmd` might work too)
2) We explicitly pass the BUILT_MODEL everywhere
3) We explicitly pass `DMD_PATH`  everywhere (see https://github.com/braddr/at-client/pull/5)

Ideas? Thoughts?

wilzbach:
> We go with an detection approach like this (sth. like ls -1 ../dmd/generated/$(OS)/$(BUILD)/*/dmd might work too)

Okay I switched to this approach. It's the same way the auto-tester currently deals with this problem & it looks a lot cleaner to.
As an additional bonus I added a second commit which will now allow building `d_do_test` on a hardened system.
So guys, any objections with this or can we move forward here? (this is needed as part of the ddmd -> dmd transition)

CC @CyberShadow maybe?

ZombineDev:
LGTM. Can you remove the fix for hardened systems and focus on in the other PR, so we can this more easily merged?

RazvanN7:
LGTM. I like the `ls -1` approach.

wilzbach:
> Can you remove the fix for hardened systems and focus on in the other PR, so we can this more easily merged?

Fair point. Done.

andralex (approved):

wilzbach:
Argh so I thought this was passing already (it did before the `unittest` target was merged). Now the auto-tester builds `generated/linux/release/<otherModel>/dmd-unittest` in the "test dmd" step.
This means that now `../generated/linux/release/64` exists and thus the `ls -1` detection doesn't work anymore as for e.g. `Linux_32_64` the 64-bit folder is existent and detected.
I guess I have to revert to [my earlier approach](https://github.com/dlang/dmd/pull/6870#issuecomment-315649502) of using `ifeq (,$(wildcard ../generated/$(OS)/$(BUILD)/64/dmd))` to detect the valid `DMD` binary ...

test/Makefile
```diff
@@ -110,21 +110,40 @@ DRUNTIME_PATH=..\..\druntime
...
+DMD_MODEL=64
+endif
+export DMD=../generated/$(OS)/$(BUILD)/$(DMD_MODEL)/dmd
+
 DRUNTIME_PATH=../../druntime
 PHOBOS_PATH=../../phobos
 # link against shared libraries (defaults to true on supported platforms, can be overridden w/ make SHARED=0)
 SHARED=$(if $(findstring $(OS),linux freebsd),1,)
-DFLAGS=-I$(DRUNTIME_PATH)/import -I$(PHOBOS_PATH) -L-L$(PHOBOS_PATH)/generated/$(OS)/release/$(MODEL)
+DFLAGS=-I$(DRUNTIME_PATH)/import -I$(PHOBOS_PATH) -L-L$(PHOBOS_PATH)/generated/$(OS)/$(BUILD)/$(MODEL)
```
wilzbach: Druntime and Phobos are built with the opposite MODEL on auto-tester
For example for 32_64: builds 32-bit DMD, but 64-but DRuntime + Phobos

PR #6871, opened 2017-06-08
Mention individual test file target in the test Makefile

wilzbach:
I guess everyone has found his favorite solution over the years, but why not adding a good "out-of-the-box" solution?

I re-open the Makefile as imho that's cleaner than duplicating the build commands (AFAIK there's no nice way in Make to force re-execution of dependencies).

### DIFF
--- a/test/Makefile
+++ b/test/Makefile
@@ -1,6 +1,7 @@
 # Execute the dmd test suite
 #
-# Targets:
+# Targets
+# -------
 #
 #    default | all:      run all unit tests that haven't been run yet
 #
@@ -14,8 +15,11 @@
 #
 #    clean:              remove all temporary or result files from prevous runs
 #
-#
-# In-test variables:
+#    test_results/compilable/json.d.out      runs an individual test
+#                                            (run log of the test is stored)
+
+# In-test variables
+# -----------------
 #
 #   COMPILE_SEPARATELY:  if present, forces each .d file to compile separately and linked
 #                        together in an extra setp.


braddr:
Already exists:  make resultsdir + testname including path +.out
ex: make test_results/runnable/arrayop.d.out

wilzbach:
> Already exists: make resultsdir + testname including path +.out
> ex: make test_results/runnable/arrayop.d.out

I know (after all this target gets called), but (1) `.test` behaves similarly to the Phobos `.test` target (i.e. it always gets executed) and (2) there's now a nice entry at the beginning of the Makefile, s.t. people don't need to search.

CyberShadow:
The comment is welcome, but I don't think we need the extra Makefile rules. I've observed that the more of those we have, the more painful everything becomes, so we should look for opportunities to shrink our makefiles, not grow them.

wilzbach:
> The comment is welcome, but I don't think we need the extra Makefile rules. I've observed that the more of those we have, the more painful everything becomes, so we should look for opportunities to shrink our makefiles, not grow them.

Hmm ok. I rebased then. As the target section started to grow, I used plain-text lines to help the reader to immediately visually distinguish between the sections.

CyberShadow:
I still don't understand why the new rules are needed.

1. The behavior of `rm`-ing the target - why? There should already be a dependency upon the built `dmd`, so it shouldn't make sense to need to force a rebuild of the target.

2. Instead of adding new rules to save typing a few characters, my suggestion was to improve the documentation and specify how to run individual rules.

wilzbach:
> I still don't understand why the new rules are needed.
> The behavior of rm-ing the target - why? There should already be a dependency upon the built dmd, so it shouldn't make sense to need to force a rebuild of the target.

Yes convenience. Shell completion works nicely and then you just need to add `.test` ;-)

> Instead of adding new rules to save typing a few characters, my suggestion was to improve the documentation and specify how to run individual rules.

Didn't you see that I already changed this PR?

wilzbach:
> Didn't you see that I already changed this PR?

Somehow my push didn't went through. Sorry about this. The `.test` targets are gone now.

CyberShadow:
> Yes convenience. Shell completion works nicely and then you just need to add .test ;-)

Shell completion should be able to partially complete a common prefix, so `test_results/` should be complete-able as well.

> Didn't you see that I already changed this PR?

Yep, looks good.

CyberShadow (approved):

PR #3531, opened 2014-05-07
Adding support for __traits(documentation, ...)

MasonMcGill:
`__traits(comment, expr)` now evaluates to the documentation comment on `expr`, if it exists and is available. Comments will only be available if DMD is invoked with the "-D" flag. If no comment is available for `expr`, `__traits(comment, expr)` evaluates to the empty string.

Unlike `__traits(identifier, expr)`, `__traits(comment, expr)` will not report an error if `expr` is not a symbol or anonymous function. It will instead evaluate to the empty string. This is in the interest of ease of use.

This feature is intended to aide automatic wrapper generation, metaprogramming (e.g. forwarding documentation from a template argument), and simplifying documentation generators. It is not intended to be involved in providing application functionality.

PR #6872, opened 2017-06-08
Revive #3531 - Adding support for __traits(documentation, ...)

wilzbach:
So I took the liberty to revive #3531, but unfortunately there hasn't been a clear consensus on how to move forward. The most important bits from the discussion:

> I object to turning this on by default [`doDocComment = true`] - it also affects memory consumption, which is currently a significant problem with dmd.

The discussion circled around adding a special flag for this as `-D` always produces an html file.

> - I think the newline should be stripped before returning it as a __traits(), otherwise user code will constantly have to strip it manually.

I share this opinion, but I wanted to wait on feedback before investing more work on this PR (same for adding more tests).

The PR to dlang.org is still [open](https://github.com/dlang/dlang.org/pull/574).

So three years later, what's your opinion on being able to query DDoc comments during CTFE?

### DIFF
--- a/src/ddmd/dmodule.d
+++ b/src/ddmd/dmodule.d
@@ -816,7 +816,9 @@ extern (C++) final class Module : Package
             return this;
         }
         {
-            scope p = new Parser!ASTCodegen(this, buf[0 .. buflen], docfile !is null);
+            // Always lex comments; __traits(documentation, ...) might require it.
+            bool doDocComment = true;
+            scope p = new Parser!ASTCodegen(this, buf[0 .. buflen], doDocComment);
             p.nextToken();
             members = p.parseModule();
             md = p.md;
--- a/src/ddmd/idgen.d
+++ b/src/ddmd/idgen.d
@@ -357,6 +357,7 @@ Msgtable[] msgtable =
     { "getUnitTests" },
     { "getVirtualIndex" },
     { "getPointerBitmap" },
+    { "documentation" },
 
     // For C++ mangling
     { "allocator" },
--- a/src/ddmd/traits.d
+++ b/src/ddmd/traits.d
@@ -80,6 +80,7 @@ static this()
         "isLazy",
         "hasMember",
         "identifier",
+        "documentation",
         "getProtection",
         "parent",
         "getLinkage",
@@ -594,6 +595,21 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         }
 
         auto se = new StringExp(e.loc, cast(char*)id.toChars());
+        return se.semantic(sc);
+    }
+    if (e.ident == Id.documentation)
+    {
+        if (dim != 1)
+            return dimError(1);
+
+        auto o = (*e.args)[0];
+        auto s = getDsymbol(o);
+        if (!s)
+        {
+            e.error("argument %s is not a symbol", o.toChars());
+            return new ErrorExp();
+        }
+        StringExp se = new StringExp(e.loc, cast(char*)s.comment);
         return se.semantic(sc);
     }
     if (e.ident == Id.getProtection)
--- a/test/compilable/extra-files/json.out
+++ b/test/compilable/extra-files/json.out
@@ -316,6 +316,7 @@
    {
     "name" : "bar",
     "kind" : "function",
+    "comment" : " Documentation test\n",
     "line" : 52,
     "char" : 16,
     "deco" : "FNeKkC4json4Bar2Zi",
@@ -397,6 +398,7 @@
     "members" : [
      {
       "kind" : "template",
+      "comment" : " Issue 9480 - Template name should be stripped of parameters\n",
       "line" : 85,
       "char" : 5,
       "name" : "this",
@@ -429,6 +431,7 @@
    {
     "kind" : "template",
     "protection" : "private",
+    "comment" : " Issue 9755 - Protection not emitted properly for Templates.\n",
     "line" : 89,
     "char" : 9,
     "name" : "S1_9755",
@@ -503,6 +506,7 @@
    {
     "name" : "c_10011",
     "kind" : "variable",
+    "comment" : " Issue 10011 - init property is wrong for object initializer.\n",
     "line" : 98,
     "char" : 14,
     "storageClass" : [
@@ -515,6 +519,7 @@
    {
     "name" : "Numbers",
     "kind" : "enum",
+    "comment" : "\n",
     "line" : 101,
     "char" : 1,
     "baseDeco" : "i",
--- a/test/runnable/traits_documentation.d
+++ b/test/runnable/traits_documentation.d
@@ -1,0 +1,38 @@
+/**
+ * Module documentation.
+ */
+module traits_documentation;
+
+/**
+ * Function documentation.
+ */
+void documentedFunc() {}
+/++
+ + Enum documentation.
+ +/
+enum documentedEnum = 1;
+auto documentedVar = 2; /// Some unicode: Σ σ Π π.
+
+auto emptyVar  = 2; ///
+
+///
+auto emptyVar2  = 2;
+
+void main()
+{
+    enum funcDoc = __traits(documentation, documentedFunc);
+    enum enumDoc = __traits(documentation, documentedEnum);
+    enum varDoc = __traits(documentation, documentedVar);
+    enum modDoc = __traits(documentation, traits_documentation);
+
+    static assert(funcDoc == " Function documentation.\n");
+    static assert(enumDoc == " Enum documentation.\n");
+    static assert(varDoc == "Some unicode: Σ σ Π π.\n");
+    static assert(modDoc == " Module documentation.\n");
+
+    static assert(!__traits(compiles, __traits(documentation, 3)));
+    static assert(!__traits(compiles, __traits(documentation, "4")));
+
+    static assert(__traits(documentation, emptyVar) == "\n");
+    static assert(__traits(documentation, emptyVar2) == "\n");
+}


CyberShadow:
> it also affects memory consumption

Since you have an implementation, why not measure the memory consumption?

andralex:
What would be some good use cases?

One simple way to estimate the impact on memory consumption is to build phobos with and without the feature enabled, and measure time and peak memory consumption. As phobos has profuse ddoc comments, that should be toward the upper bound. If there is no significant impact (as I suspect), then that particular argument goes away. @wilzbach can you to this test? Thanks!

CyberShadow:
> What would be some good use cases?

Here is my use case:

https://github.com/CyberShadow/Digger/blob/72eb4718eb2eaf5d24f5cd724a0e3273a0ee4641/digger.d#L64-L158

Here, each method represents a command-line action. I use a slightly ugly string literal UDA for the `--help` documentation, but it would be nicer to be able to use regular DDoc comments instead.

andralex:
@CyberShadow cool, though one counter-argument is that cmdline help documentation has restrictions (e.g. single-line, under 70 chars or so) that actually make it more suitable as a UDA. Documentation comments are usually not subjected to such restrictions.

Just to clarify the general idea: I'm generally in favor of more introspection capabilities, but each should be motivated by good applicability.

ZombineDev:
@andralex just replace cmdline with GUI - i.e. one function per GUI command; hovering over a button displays the documentation in a tooltip.

Another use case is named unittests. A test runner could show the name of a failing unittest with something along the lines of:
```D
foreach (test; __traits(getUnitTests, MyClass))
    try
        test();
    catch (AssertError e)
    {
        static immutable summary = __traits(documentation, test);
"
+-------------------------------------
| %s
+-------------------------------------
| Summary: %s
| File: %s
| Line: %s
+-------------------------------------"
        .writefln(e.msg, summary, e.file, e.line);
        continue;
```

ZombineDev:
One could also imagine using CTFE instead of ddoc macros for certain use cases, or even preprocessing documentation comments, mixing them again and leaving them to the standard ddoc pipeline.

andralex:
>One could also imagine using CTFE instead of ddoc macros for certain use cases, or even preprocessing documentation comments, mixing them again and leaving them to the standard ddoc pipeline.

Whoa, wait. Are you saying a module could define and run a compile-time function that preprocesses some or all of the documentation in that module? That would be solid!

ZombineDev:
Yes, exactly my idea, though I haven't tried it yet.

ZombineDev:
I tried the following function:

```D
import std.algorithm, std.range, std.format, std.meta, std.traits;
enum stringOf(T) = T.stringof;
enum sizeOf(T) = T.sizeof;
string makeDdoc(alias f)(string summary)
{
    return format(
"
/**
 * %s
 * 
 * Params:
%-(%s\n%)
 */",
        summary,
        [ ParameterIdentifierTuple!f ]
            .zip([ staticMap!(stringOf, Parameters!f) ], [ staticMap!(sizeOf, Parameters!f) ])
            .map!(x => " *     %s = A parameter of type `%s`, size %s.".format(x[0], x[1], x[2]))
    );
}
```

That when applied to a function `fun`:

```D
void fun(short x, float y, double z);

makeDdoc!fun("This is a great function.")
```

Yields:

```D
/**
 * This is a great function.
 *
 * Params:
 *     x = A parameter of type `short`, size 2.
 *     y = A parameter of type `float`, size 4.
 *     z = A parameter of type `double`, size 8.
 */
```

However, when I mixed in the documentation in front of the function like so:
```D
mixin (makeDdoc!fun("This is a great function."));
void fun(short x, float y, double z);
```

And ran the following command
```sh
dmd -D test.d
```

I got:
actual
When I was expecting:
expected

Which confirmed my expectation that ddoc parsing doesn't take into account mixins.
@andralex do you think there is value in opening an enhancement request for adding a flag to dmd which would force the evaluation of all mixins at module scope before ddoc is parsed?

andralex:
I'd say this PR would be a lot more valuable if it afforded users the ability to preprocess ddoc comments.

ZombineDev:
On second thought, I don't think we need more than more than __traits(documentation), because preprocessing can easily be accomplished with a little more work:

```sh
dmd -o- test.d 2> preprocessed.d && dmd -o- -Dftest.html -D preprocessed.d && rm preprocessed.d
```

`test.d`:
```D
import std.algorithm, std.array, std.range, std.format, std.meta, std.traits;

enum stringOf(T) = T.stringof;
enum sizeOf(T) = T.sizeof;
string makeDdoc(alias f)(string summary)
{
    return format(
"
/**
 * %s
 * 
 * Params:
%-(%s\n%)
 */
%s;",
        summary,
        [ ParameterIdentifierTuple!f ]
            .zip([ staticMap!(stringOf, Parameters!f) ], [ staticMap!(sizeOf, Parameters!f) ])
            .map!(x => " *     %s = A parameter of type `%s`, size %s.".format(x[0], x[1], x[2])),
        typeof(&f).stringof.replace("function", __traits(identifier, f))
    );
}

pragma (msg, makeDdoc!fun("This is a great function."));
void fun(short x, float y, double z);
```

test.html:

preprocessed

Replacing the `pragma (msg, ...)` with a mixin that iterates over the members of the module should be good enough for a lot of use cases.

wilzbach:
> One simple way to estimate the impact on memory consumption is to build phobos with and without the feature enabled, and measure time and peak memory consumption. As phobos has profuse ddoc comments, that should be toward the upper bound. If there is no significant impact (as I suspect), then that particular argument goes away. @wilzbach can you to this test? Thanks!

So we allocate 2.75 MB more at the peak, which imho would be neglectable considering that the peak is at 1.1 GB.
However there seems to be a noticable difference in avgtime, but better see the raw numbers for yourself.

With `__traits(documentation)`
-----------------------------------------

### Peak memory consumption: (i.e. last Valgrind `massif` snapshot):

```
--------------------------------------------------------------------------------
  n        time(i)         total(B)   useful-heap(B) extra-heap(B)    stacks(B)
--------------------------------------------------------------------------------
 70 23,355,923,847    1,253,708,328    1,159,290,142    94,418,186            0
```

image


### AvgTime

```
> ~/dlang/avgtime/avgtime -r30 -d -q $(cat test.sh)
------------------------
Total time (ms): 200455
Repetitions    : 30
Sample mode    : 6730 (2 occurrences)
Median time    : 6443.3
Avg time       : 6681.82
Std dev.       : 555.537
Minimum        : 6047.94
Maximum        : 8455.6
95% conf.int.  : [5592.98, 7770.65]  e = 1088.83
99% conf.int.  : [5250.85, 8112.79]  e = 1430.97
EstimatedAvg95%: [6483.02, 6880.61]  e = 198.793
EstimatedAvg99%: [6420.56, 6943.08]  e = 261.258
```


Without `__traits(documentation)`
---------------------------------------------

### Peak memory consumption:

```
--------------------------------------------------------------------------------
  n        time(i)         total(B)   useful-heap(B) extra-heap(B)    stacks(B)
--------------------------------------------------------------------------------
 87 23,227,888,304    1,250,958,184    1,156,655,280    94,302,904            0
```


image

### AvgTime

```
> ~/dlang/avgtime/avgtime -r30 -d -q $(cat test.sh)
Total time (ms): 197151
Repetitions    : 30
Sample mode    : 5860 (2 occurrences)
Median time    : 6398.85
Avg time       : 6571.69
Std dev.       : 716.995
Minimum        : 5788.27
Maximum        : 7886.89
95% conf.int.  : [5166.41, 7976.98]  e = 1405.28
99% conf.int.  : [4724.84, 8418.55]  e = 1846.86
EstimatedAvg95%: [6315.12, 6828.26]  e = 256.569
EstimatedAvg99%: [6234.5, 6908.88]  e = 337.188
```

andralex:
Yah, the increase in build time is measurable. My main problem is we're lacking a killer use case for this. In fact the converse may be more useful: make the body of a enum or function available to ddoc. Then you can click on a "+" to expand it inline.

wilzbach:
> Yah, the increase in build time is measurable. My main problem is we're lacking a killer use case for this. In fact the converse may be more useful: make the body of a enum or function available to ddoc. Then you can click on a "+" to expand it inline.

Please don't let us develop no features for Ddoc - this is far easier with Ddox.
In fact we already have a "View source code" button which takes you to the exact lines on the source code, e.g. https://dlang.org/library/std/algorithm/comparison/among.html

So what other options do we have?

- Only activate when used with `-D`
- Only activate with a special flag
- Dump the idea

andralex:
I'd say keep it on the back burner until we find a couple of good ideas for using it. Discussing it in the forum may help.

wilzbach:
> I'd say keep it on the back burner until we find a couple of good ideas for using it. Discussing it in the forum may help.

-> http://forum.dlang.org/post/cwjtsmnwwwwepvgvlqcy@forum.dlang.org

adamdruppe:
I like cybershadow's use case - loading the documentation comment is really nice for command line things. Column limits are a simple case of authoring and word-wrapping; trivial, no need to separate and duplicate the text.

As to preprocessing docs... no, bad idea. That'd be something the alternative doc parsers can't really duplicate and would make the docs likely less readable in source code. That's a net negative value.

tchaloupka:
One usecase I would add is compile time generation of the OpenAPI documentation of REST API (ie using vibe-d) - see https://swagger.io/

stefan-koch-sociomantic:
Although I cannot see a immediate usecase, I will say that at this introducing memory problems is no concern.
So technically this will fly like a bird.

WalterBright:
Putting this in the Phantom Zone due it needing a compelling use case(s) to justify consuming memory and compilation time.

PR #6874, opened 2017-06-09
CodeBuilder: prolog code

WalterBright:


src/ddmd/backend/cod3.c
```diff
@@ -799,7 +796,6 @@ void outblkexitcode(block *bl, code* c, int& anyspill, const char* sflsave, symb
             }
             logexp(cdb,e,jcond,FLblock,(code *) bs1);
             nextb = bs2;
-            bl->Bcode = NULL;
```
yebblies: Won't this change be visible outside this function?  Why wouldn't we care about Bcode's value?

WalterBright: The only time `bl->Bcode` would be non-NULL when entering this function is if's a BCasm block. It took me a while to figure out that this invariant was the case. Leaving that assignment in there would just be confusing. It's also pointless, as after the function returns then `bl->Bcode` is set.

WalterBright:
ping @yebblies

wilzbach:
Closing and reopening to rekick Jenkins

PR #6876, opened 2017-06-09
Cleanup PR 6864: scope delegates change type, but not mangling

rainers:
#6864 was not complete, a few more issues showed up:

- do not emit "scope" in mangling if inferred on a parameter
- do not infer scope if explicitly specified (this might omit it later in mangling)

Also fixed typo and updated C++ header.


WalterBright:
Auto-merge toggled on

PR #6877, opened 2017-06-09
Auto-generate traits from source

wilzbach:
This is a proof of concept following from https://github.com/dlang/dlang.org/pull/1683 and the idea is to generate the spec based on the ddoc comments.

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -542,13 +542,10 @@ tuple(3)
 tuple("string", 7)
 tuple((Foo))
 )
-)
 */
 enum getAttributes = "getAttributes";
 
 /**
-$(SECTION2 $(GNAME getFunctionVariadicStyle),
-
 Takes one argument which must either be a function symbol, or a type
  is a function, delegate or a function pointer.
 
@@ -582,7 +579,6 @@ static assert(__traits(getFunctionVariadicStyle, typesafe) == "typesafe");
 static assert(__traits(getFunctionVariadicStyle, (int[] a...) {}) == "typesafe");
 static assert(__traits(getFunctionVariadicStyle, typeof(cstyle)) == "stdarg");
 ---
-)
 */
 enum getFunctionVariadicStyle = "getFunctionVariadicStyle";
 
@@ -629,7 +625,6 @@ pragma(msg, __traits(getFunctionAttributes, S.test));
 // prints ("pure", "nothrow", "@nogc", "@trusted")
 pragma(msg, __traits(getFunctionAttributes, (int x) @trusted { return x * 2; }));
 ---
-)
 */
 enum getFunctionAttributes = "getFunctionAttributes";
 
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -1140,7 +1140,6 @@ shared static this()
         derivedMembers,
         isSame,
         compiles,
-        "parameters",
         getAliasThis,
         getAttributes,
         getFunctionAttributes,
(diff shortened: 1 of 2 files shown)


src/ddmd/traits.d
```diff
@@ -54,53 +54,1012 @@ struct PushAttributes
...
-        "getMember",
-        "getOverloads",
-        "getVirtualFunctions",
-        "getVirtualMethods",
-        "classInstanceSize",
-        "allMembers",
-        "derivedMembers",
-        "isSame",
-        "compiles",
-        "parameters",
```
wilzbach: AFAICT is `parameters` not documented _nor_ implemented. Is this a relict?

andralex:
This looks very nice. I think we should merge it regardless on future directions. @WalterBright ?

andralex:
@wilzbach mind the odd errors in testers

wilzbach:
> @wilzbach mind the odd errors in testers

Thanks.
This was due to unimplemented and unused `parameters` traits. I split these changes into a separate commit. If it's controversial, I am happy to drop this commit.

src/ddmd/traits.d
```diff
@@ -54,53 +54,1012 @@ struct PushAttributes
 
 extern (C++) __gshared StringTable traitsStringTable;
 
+/**
```
andralex: Made one more pass, a thought: this exposes a bunch of nicely-named symbols. Can we make them private and still generate the documentation?

wilzbach: I don't think so (without bug hacks/work), but DMD != Phobos, so having public symbols shouldn't a big problem. Currently the only users are GCC + LDC and they can't use this without C++ interfaces. Also we are stuck with these traits nearly forever, so I doubt that worrying about the deprecation of a public symbol is a worry here. Deprecating a trait from the language is a different pair of shoes though.

WebFreak001:
this would be really nice for example in DCD where there is a list of traits for autocompletion currently hardcoded in. It would make sense having this in a few places where D code generation is done at CTFE for example as you could go over `__traits(allMembers`. Except for auto completion programs it could also be used to check using `is(typeof(...))` instead of `__traits(compiles` if a trait is available, which might be a bit nicer

wilzbach:
Rebased to include the new `isDisabled` trait.

wilzbach:
(rebaseD again).

Another ping on this.
Seems like for the last nine months no one opposed this and @andralex was in favor...

andralex (approved):
Yes, apologies for the slowness. I think a good next step is to hook this in and eliminate the now redundant docs on dlang.org. Thanks!

src/dmd/traits.d
```diff
@@ -93,56 +93,1048 @@ private Dsymbol getDsymbolWithoutExpCtx(RootObject oarg)
...
+    $(P Prints:)
+
+$(CONSOLE
+true
+true
+false
+false
+)
+*/
+enum isArithmetic = "isArithmetic";
```
jacob-carlborg: I suggest using `__traits(identifier, isArithmetic)` to get the string. Instead of having to repeat all the enum names twice.

jacob-carlborg (changes requested):

WalterBright:
I've been finding it convenient to merely put a link in the code to the spec.

src/dmd/traits.d
ibuclaw: Isn't that what .stringof is for? I see no problem really with the repetition, as the intent is clear.

jacob-carlborg: `.stringof` will result in a forward reference error.

wilzbach:
It would have been nice to automatically generate the list of traits, but I have bigger fish to fry for now.

PR #6878, opened 2017-06-09
Add id.d to the html target

wilzbach:
Needed for https://github.com/dlang/dlang.org/pull/1686


WalterBright:
Auto-merge toggled on

PR #6880, opened 2017-06-09
Issue 9287 - Implement -stdin.

quickfur:
It's nice to be able to pipe source code into dmd via standard input, e.g., in code generation programs so that you don't have to create a temporary file just to be able to compile the code.

Based on the bug notes in the original bug, using `-` to mean stdin is probably a problematic idea, so I have chosen a different route, i.e., use `-stdin` to indicate that dmd should read source code from stdin. I think this is clearer, and less *nix-specific.

### DIFF
--- a/src/ddmd/mars.d
+++ b/src/ddmd/mars.d
@@ -1033,6 +1033,8 @@ Language changes listed by -transition=id:
                     goto Lnoarg;
                 }
             }
+            else if (p[1] == '\0')
+                files.push("__stdin.d");
             else
             {
             Lerror:
--- a/src/ddmd/root/file.d
+++ b/src/ddmd/root/file.d
@@ -77,12 +77,65 @@ nothrow:
     {
         if (len)
             return false; // already read the file
+
+        import core.stdc.string : strcmp;
+        const(char)* name = this.name.toChars();
+        if (strcmp(name, "__stdin.d") == 0)
+        {
+            /* Read from stdin */
+            enum bufIncrement = 128 * 1024;
+            size_t pos = 0;
+            size_t sz = bufIncrement;
+
+            if (!_ref)
+                .free(buffer);
+
+            buffer = null;
+            L1: for (;;)
+            {
+                buffer = cast(ubyte*).realloc(buffer, sz + 2); // +2 for sentinel
+                if (!buffer)
+                {
+                    printf("\tmalloc error, errno = %d\n", errno);
+                    break L1;
+                }
+
+                // Fill up buffer
+                do
+                {
+                    assert(sz > pos);
+                    size_t rlen = fread(buffer + pos, 1, sz - pos, stdin);
+                    pos += rlen;
+                    if (ferror(stdin))
+                    {
+                        printf("\tread error, errno = %d\n", errno);
+                        break L1;
+                    }
+                    if (feof(stdin))
+                    {
+                        // We're done
+                        assert(pos < sz + 2);
+                        len = pos;
+                        buffer[pos] = '\0';
+                        buffer[pos + 1] = '\0';
+                        return false;
+                    }
+                } while (pos < sz);
+
+                // Buffer full, expand
+                sz += bufIncrement;
+            }
+            .free(buffer);
+            buffer = null;
+            len = 0;
+            return true;
+        }
+
         version (Posix)
         {
             size_t size;
             stat_t buf;
             ssize_t numread;
-            const(char)* name = this.name.toChars();
             //printf("File::read('%s')\n",name);
             int fd = open(name, O_RDONLY);
             if (fd == -1)
@@ -135,7 +188,6 @@ nothrow:
         {
             DWORD size;
             DWORD numread;
-            const(char)* name = this.name.toChars();
             HANDLE h = CreateFileA(name, GENERIC_READ, FILE_SHARE_READ, null, OPEN_EXISTING, FILE_ATTRIBUTE_NORMAL | FILE_FLAG_SEQUENTIAL_SCAN, null);
             if (h == INVALID_HANDLE_VALUE)
                 goto err1;
--- a/test/runnable/test9287.sh
+++ b/test/runnable/test9287.sh
@@ -1,0 +1,12 @@
+#!/usr/bin/env bash
+
+src=runnable${SEP}extra-files
+dir=${RESULTS_DIR}${SEP}runnable
+output_file=${dir}/test9287.sh.out
+
+echo 'import std.stdio; void main() { writeln("Success"); }' | \
+	$DMD -m${MODEL} -of${dir}${SEP}test9287a${EXE} - || exit 1
+
+${RESULTS_DIR}/runnable/test9287a${EXE} > ${output_file}
+
+\rm -f ${dir}/test9287a{${OBJ},${EXE}}


quickfur:
Note that I had to hack together a stdio-based way of copying data from stdin into a buffer, because the existing `File` class in `src/ddmd/root/file.d` does not work with non-file handles: it uses `stat` to discover the total file length, which is not available with stdin.  I also opted for using stdio so that I don't have to figure out how to write a Windows-specific version which I currently have no way to test.

quickfur:
I don't understand the autotester failure.  Why is it compiling `test/runnable/bug12928.d`, which contains a deliberate array overrun, with `-boundscheck=off`, and then complaining that the test crashes?!

quickfur:
Ah, nevermind, just realized that it was dmd crashing because I changed `struct Global` but forgot to update `global.h`.

src/ddmd/mars.d
```diff
@@ -245,6 +245,52 @@ extern (C++) void genCmain(Scope* sc)
...
+ *
+ * The buffer is always null-terminated.
+ */
+ubyte[] copyFile(FILE* fp)
+{
+    enum bufIncrement = 4096;
+    size_t pos = 0;
+    size_t sz = bufIncrement;
+
+    ubyte* buf = cast(ubyte*).malloc(sz + 1); // +1 for sentinel
```
UplinkCoder: `malloc` is only meant for buffers which are freed later.
`use xmalloc` amd `xrealloc instead`
even better would be in walking the length of the input and the using allocmemory.

UplinkCoder:
@quickfur what is the specific usecase for this ?

wilzbach:
@quickfur if you use "Fix <bugzilla title>" as a commit message, the bot will recognize it ([more infos](https://github.com/dlang-bots/dlang-bot)).

@UplinkCoder most tools/compilers allow this - this is nice for a bunch of use cases starting with showing compile errors in my favorite editor (most editors read the un-saved instance in-memory and thus can nicely pipe it to `dmd`)  - [see the issue for more details](https://issues.dlang.org/show_bug.cgi?id=9287).

src/ddmd/mars.d
quickfur: Went with xmalloc / xrealloc.  Didn't really want to walk the input, since it could be large, if xrealloc already takes care of allocating a large-enough buffer for us.

quickfur:
Also, please advise if there is a better place to put `copyFile` than directly in `mars.d`. Somehow it feels a little out-of-place there.

WalterBright:
Protocol: when creating a PR to fix Issue https://issues.dlang.org/show_bug.cgi?id=9287, add a link in bugzilla to the PR. I already did so for this.

WalterBright:
As I commented in https://issues.dlang.org/show_bug.cgi?id=9287 please use `-` not `-stdin`. We shouldn't be reinventing common Linux conventions.

wilzbach:
> 12.82% of diff hit (target 85.504%)
> 85.461% (-0.044%) compared to ebd995a

Sorry to bother you with this, but adding a test in the testsuite for this will highly increase chances of acceptance ;-)

quickfur:
@WalterBright  I did put the PR link in bugzilla, did you miss it?

As for '-' instead of '-stdin', I'll get to it later. Thanks for the feedback. 

PS: where's the best place to put 'copyFile'?

quickfur:
@wilzbach How would you test something involving standard input? I don't know if the current test suite runner is capable of piping code via stdin.

wilzbach:
> @wilzbach How would you test something involving standard input? I don't know if the current test suite runner is capable of piping code via stdin.

The test suite is Turing-complete (it allows the use of Bash), see e.g. [this simple test](https://github.com/dlang/dmd/blob/master/test/runnable/test10567.sh):

```bash
#!/usr/bin/env bash

src=runnable${SEP}extra-files
dir=${RESULTS_DIR}${SEP}runnable
output_file=${dir}/test10567.sh.out

$DMD -m${MODEL} -I${src} -of${dir}${SEP}test10567a${OBJ} -c ${src}${SEP}test10567a.d || exit 1
$DMD -m${MODEL} -I${src} -of${dir}${SEP}test10567${EXE} ${src}${SEP}test10567.d ${dir}${SEP}test10567a${OBJ} || exit 1
${RESULTS_DIR}/runnable/test10567${EXE} || exit 1

rm ${dir}/{test10567a${OBJ},test10567${EXE}}

echo Success >${output_file}
```

Simply rewriting sth. like this to use `cat` should work ;-)
You can run an individual test with `make test_results/<name of test>.out`

quickfur:
Thanks for the tip!  Got a working test now.

quickfur:
Hmm, the Jenkins errors appear unrelated to the changes in this PR.  What's going on?

quickfur:
Nevermind, it seems to be affecting other PRs too.

src/ddmd/mars.d
```diff
@@ -245,6 +245,54 @@ extern (C++) void genCmain(Scope* sc)
...
+ubyte[] copyFile(FILE* fp)
+{
+    import ddmd.root.rmem : mem;
+
+    enum bufIncrement = 4096;
+    size_t pos = 0;
+    size_t sz = bufIncrement;
+
+    ubyte* buf = cast(ubyte*)mem.xmalloc(sz + 1); // +1 for sentinel
+    if (buf is null)
```
WalterBright: `xmalloc()` never returns null

src/ddmd/mars.d
```diff
@@ -245,6 +245,54 @@ extern (C++) void genCmain(Scope* sc)
...
+    {
+        error(Loc(), "out of memory");
+        fatal();
+    }
+
+    assert(sz > pos);
+    size_t rlen;
+    pos += rlen = fread(buf + pos, 1, sz - pos, fp);
+    while (!ferror(fp) && !feof(fp))
```
WalterBright: This loop can be written so there is only one `xmalloc` call instead of two, and one `fread` instead of two.

src/ddmd/mars.d
```diff
@@ -245,6 +245,54 @@ extern (C++) void genCmain(Scope* sc)
...
+    {
+        if (pos == sz)
+        {
+            sz += bufIncrement;
```
WalterBright: Reallocating every 4K bytes makes for a very slow read. `expression.d` is 500K, so that makes for 125 copies and reallocations. I'd probably make it at least 128K.

src/ddmd/mars.d
```diff
@@ -245,6 +245,54 @@ extern (C++) void genCmain(Scope* sc)
...
+            if (buf is null)
+            {
+                error(Loc(), "out of memory");
+                fatal();
+            }
+        }
+        assert(sz > pos);
+        pos += rlen = fread(buf + pos, 1, sz - pos, fp);
+    }
+    if (ferror(fp))
```
WalterBright: `ferror` is checking twice

src/ddmd/mars.d
```diff
@@ -245,6 +245,54 @@ extern (C++) void genCmain(Scope* sc)
     rootHasMain = sc._module;
 }
 
+/**
+ * Copies data from a FILE* into a newly-allocated buffer.
+ *
+ * The buffer is always null-terminated.
```
WalterBright: Use Ddoc style comments, with `Params:` and `Returns:` sections.

src/ddmd/mars.d
```diff
@@ -1210,6 +1260,10 @@ Language changes listed by -transition=id:
     {
         files.push(cast(char*)global.main_d); // a dummy name, we never actually look up this file
     }
+    if (global.params.readStdin)
+    {
+        files.push(cast(char*)global.stdin_d); // a dummy name, we don't actually look it up
```
WalterBright: maybe put this before line 1134, then the test doesn't have to be done twice

src/ddmd/mars.d
```diff
@@ -1343,6 +1397,22 @@ Language changes listed by -transition=id:
             }
         }
     }
+    if (global.params.readStdin)
```
WalterBright: Seems like this code should go into `root/file.d`, as reading from stdin should be a low level thing, not a high level one.

WalterBright:
I suggest that if `root.File.read()` simply checked for `"-"` as a filename in `File.name`, the logic could all be moved there and the the rest of the complicated logic can be removed from the main code.

andralex:
great review (and idea) @WalterBright

quickfur:
@WalterBright  Unfortunately, merely checking for "-" in `root.File.read()` is not quite as simple an approach as it may initially appear, because:

(1) the filename "-" fails the file extension checking code (since "-" has no extension), so we have to special-case it to get through that check;

(2) "-" is an invalid module name, so we need another special case to get it past the creation of the module identifier;

(3) in error messages, you probably want a more descriptive name than "-", so we need to create a fake module identifier anyway, similar to what `-main` does today.

So it's unlikely the "complicated logic" will go away either way. The current implementation bypasses (1), (2), and (3) by assigning a pseudo-identifier that passes both the file extension check and the module identifier check, and also serves as a semi-comprehensible name in error messages.

Having said that, though, moving the stdin-reading code into `root.File.read` totally makes sense; it doesn't belong in `mars.d`. So I'll do at least that.

quickfur:
Also, the module name is used for deriving the name of object files, ddoc output files, etc., and using "-" for that seems iffy at best, even if we special-case it to be allowed as a module name.

src/ddmd/mars.d
```diff
@@ -1316,6 +1318,8 @@ Language changes listed by -transition=id:
         /* At this point, name is the D source file name stripped of
          * its path and extension.
          */
+        if (strcmp(name, "-") == 0)
+            name = "__stdin";
```
quickfur: This is a hack to get around the fact that `-` is not a valid module name.  Also, this also gives us a saner basename for deriving the filenames of object files, ddoc output, etc..

src/ddmd/root/file.d
```diff
@@ -77,12 +77,69 @@ nothrow:
     {
         if (len)
             return false; // already read the file
+
+        import core.stdc.string : strcmp;
+        const(char)* name = this.name.toChars();
+        if (strcmp(name, "-.d") == 0)
```
quickfur: Yes this is very ugly, but it's the simplest solution that produces the smallest diff to do what needs to be done.  The alternative is to make a whole bunch of complicated changes to `Module`'s ctor involving if-statements with complicated conditions to detect the `-` special case.  By just letting the `Module` ctor do what it currently does today, all we really need to do to make this work is to check for `"-.d"` instead of `-`.

src/ddmd/mars.d
quickfur: As well as more readable error messages.

quickfur:
OK, I redid the implementation to minimize the changes to `mars.d` and moving the stdin-reading code to `root.File.read`.

Unfortunately, the simplistic hack to make a valid module name for `"-"` did not solve the problem of readable error messages, because it still uses "-" as the file name, neither did it solve the problem of producing derived filenames, so currently it would create files name `-.o`, `-.html`, etc..

andralex:
>Also, the module name is used for deriving the name of object files, ddoc output files, etc., and using "-" for that seems iffy at best, even if we special-case it to be allowed as a module name.

We should only accept "-" if used together with -of, and use the root of the object file specified as the module name.

src/ddmd/mars.d
```diff
@@ -1022,6 +1022,8 @@ Language changes listed by -transition=id:
                     goto Lnoarg;
                 }
             }
+            else if (strcmp(p + 1, "") == 0)
```
andralex: `p[1] == 0`

src/ddmd/root/file.d
```diff
@@ -77,12 +77,69 @@ nothrow:
...
+
+                // Buffer full, expand
+                sz += bufIncrement;
+            }
+        err3:
+            .free(buffer);
+            buffer = null;
+            len = 0;
+            return true;
+        }
```
andralex: (Not familiar with the codebase.) Why couldn't the existing file reading code be used?

src/ddmd/root/file.d
```diff
@@ -77,12 +77,69 @@ nothrow:
...
+                do
+                {
+                    size_t rlen;
+
+                    assert(sz > pos);
+                    pos += rlen = fread(buffer + pos, 1, sz - pos, stdin);
+                    if (ferror(stdin))
+                    {
+                        printf("\tread error, errno = %d\n", errno);
+                        goto err3;
```
andralex: We should avoid use of `goto` in new code without a strong motivation.

src/ddmd/root/file.d
```diff
@@ -77,12 +77,69 @@ nothrow:
...
+                if (!buffer)
+                {
+                    printf("\tmalloc error, errno = %d\n", errno);
+                    goto err3;
+                }
+
+                // Fill up buffer
+                do
+                {
+                    size_t rlen;
```
andralex: initialize upon definition

src/ddmd/root/file.d
```diff
@@ -77,12 +77,69 @@ nothrow:
...
+
+                    assert(sz > pos);
+                    pos += rlen = fread(buffer + pos, 1, sz - pos, stdin);
+                    if (ferror(stdin))
+                    {
+                        printf("\tread error, errno = %d\n", errno);
```
andralex: use https://www.tutorialspoint.com/c_standard_library/c_function_perror.htm?

src/ddmd/root/file.d
```diff
@@ -77,12 +77,69 @@ nothrow:
...
+                        goto err3;
+                    }
+                    if (feof(stdin))
+                    {
+                        // We're done
+                        assert(pos < sz + 2);
+                        len = pos;
+                        buffer[pos] = '\0';
+                        buffer[pos+1] = '\0';
```
andralex: spaces around operator

andralex (changes requested):
The main change requested is requiring -of

CyberShadow:
> We should only accept "-" if used together with -of

Then you can't use `-o-` with stdin, which has got to be one of the biggest use cases for it (syntax-check this editor buffer).

src/ddmd/root/file.d
quickfur: I was just following how the surrounding code in `read()` was written.

src/ddmd/root/file.d
quickfur: Ditto

src/ddmd/root/file.d
quickfur: Because:

(1) it requires the file size to be known in advance, but when reading stdin this is not possible; changing the existing code to handle this case would introduce a lot of extra complications that are usually not used (reading from stdin is a very specific use case, so why complicate the more usual code path and possibly penalizing it due to overhead of more complex code?).

(2) I'm not familiar enough with Windows API to be able to write code for copying standard input, but since C's stdio already does this, and works on both platforms, why not just use it.

src/ddmd/mars.d
quickfur: done

src/ddmd/root/file.d
quickfur: done

src/ddmd/root/file.d
quickfur: done

quickfur:
BTW, I have another idea for dealing with the problem of "-" not being a valid module name.  I don't have much time to work on this right now so I'll probably have to push the changes sometime tomorrow.  If this works, we may not need to require `-of` after all.

quickfur:
OK, I redid the implementation with the following points:

1) Reduced changeset: the only changes now are the parsing of "-" on the command line, and the code for reading from standard input in `root.file.read`.

2) Instead of adding special cases everywhere for dealing with a module named "-", the command-line parsing code just internally translates it into the reserved module name `__stdin` (precedent: `-main` does a similar thing). This takes care of (a) standard input having a valid module name; (b) generated filenames like ddoc output and object files will be `__stdin.html`, `__stdin.o`, `__stdin.exe`, etc., rather than the very ugly `-.html`, `-.o`, etc.; (c) nicer error messages (`__stdin.d:1: Syntax error...` rather than the less-readable `-:1: Syntax error`.

3) Restructured the read loop per @WalterBright's comments, got rid of goto's and did other cleanups per @andralex's comments.

src/ddmd/root/file.d
```diff
@@ -77,12 +77,65 @@ nothrow:
...
+            {
+                buffer = cast(ubyte*).realloc(buffer, sz + 2); // +2 for sentinel
+                if (!buffer)
+                {
+                    printf("\tmalloc error, errno = %d\n", errno);
+                    break L1;
+                }
+
+                // Fill up buffer
+                do
```
UplinkCoder: a while more expressive in this case.
Since the condition is supposed to be true on the first run.

UplinkCoder:
@quickfur what is a possible usecase for this ?
Where you could not create a temporary file as well ?
reading stdin introduces a few more points of failure on top of being slow.

src/ddmd/root/file.d
quickfur: @WalterBright wanted me to eliminate redundant function calls and conditions, so this is the solution I came up with.  What do you propose?

quickfur:
@UplinkCoder:  @CyberShadow already mentioned a few use cases, e.g., piping code from an editor buffer for syntax checking in an IDE. There's also JIT of generated code in D programs, and so on.  It's always possible to create a temporary file for this, of course. But it's just extra trouble. (And temporary files can be especially troublesome to handle correctly in security-conscious applications.)  Especially since other compilers like gcc support reading from stdin without any hassles.

quickfur:
Also, I'm not sure what you mean by a few extra points of failure.  This PR follows the existing dmd model of reading the entire file into memory before doing anything else; so the worst that could happen is that the read fails and dmd aborts.

As for speed, I doubt anyone would pipe, say, a 50MB source file via stdin to incur enough overhead to be noticeable. After the file is read, no further overhead happens.  Also, the usual code path of reading from a file is untouched, so the usual case of reading from file remains as fast as before.

CyberShadow:
> reading stdin introduces a few more points of failure on top of being slow.

Creating a single-purpose temporary file is not going to be faster.

BTW, dmd should support reading source code from pipes regardless, so that `dmd /dev/stdin` or `dmd <(some command)` both work.

quickfur:
Side note: the Jenkins failure is unrelated to this PR; it's caused by a deprecation introduced by https://github.com/dlang/phobos/pull/5378

timotheecour:
Just sent an forum email: "proposal: reading multiple files with dmd -stdin and allow specifying (fake) file names" (http://forum.dlang.org/post/mailman.3266.1497502126.31550.digitalmars-d@puremagic.com) relevant for an extension of this PR; it's based on using 
```
pragma(module, "foo/somefakefilename.d")
module foo1;
pragma(module, "foo/somefakefilename2.d")
module foo2;
```
But even the single-module version is relevant for this PR, adding an optional fakefilename instead of guessing one, useful for writing tests and bug reports for eg.

andralex:
Can we use the https://dlang.org/dmd-linux.html#switch-mv flag for asking the user to name the module instead of deciding it?

quickfur:
@andralex  Sounds like an interesting idea, but I'd suggest making that optional?  I'll take a look at the code tomorrow to see what can be done.

andralex:
I'm a fan of the notion of requiring the user to define names instead of
choosing an arbitrary __stdin for them.

timotheecour:
`pragma(module, "foo/somefakefilename.d")` would solve this file naming issue and also allow passing in multiple files via stdin, see http://forum.dlang.org/post/mailman.3266.1497502126.31550.digitalmars-d@puremagic.com

quickfur:
@andralex Fair enough, I'll look at the code tomorrow to see what can be done.

quickfur:
Finally got around to this again.

@andralex:  I took another look at the code today to see whether `-mv=...` could be used to specify the module name for standard input:

(1) I found that actually, the `-mv=...` code isn't even run for modules that are passed on the command-line.  For modules (source files) passed on the command line, dmd simply uses the filename as the basis for constructing the module name.  The `-mv=...` code only takes effect if you import something with an import statement but the named module resides in a non-standard filename.  So it doesn't seem to be relevant to this PR at all.

(2) It is actually possible to include a `module xyz;` declaration in the standard input code; in that case, the specified name will be used as the module name instead of `__stdin`.  So you could regard using `__stdin` as a convenient default when the user didn't specify a module name explicitly.  (In the non-stdin case, dmd would just take the filename as the module name.) But if the user wishes to use a specific module name, it's just a simple matter of adding a module declaration to the top of the input.

quickfur:
@wilzbach  How come dlang-bot isn't picking up the issue number?  What's the expected format for indicating bug numbers?

wilzbach:
> @wilzbach How come dlang-bot isn't picking up the issue number? What's the expected format for indicating bug numbers?

I guess he doesn't like the colon:

https://github.com/dlang-bots/dlang-bot#automated-references

In doubt copy the bugzilla title and prefix "Fix "

quickfur:
He's still not liking it. I'm confused.

CyberShadow:
@quickfur Git commit message, not GitHub PR title / description.

quickfur:
Just did that, he still doesn't like it.

quickfur:
Ah, he's expecting the word "issue". Got it, thanks!

wilzbach:
> Ah, he's expecting the word "issue". Got it, thanks!

Exact regex-> https://github.com/dlang-bots/dlang-bot#nerdy-details
Though I don't know why the double posts happened - maybe yet another lag in the GH event API?

quickfur:
ping @WalterBright  @andralex 

Is this PR satisfactory, or should a different approach be taken? Please advise, thanks!

andralex:
It seems the PR cannot be made much simpler. I'm mildly in favor of the notion of accepting piped input, without being able to think of a solid application. OK with me. @WalterBright ?

WalterBright (approved):

src/ddmd/mars.d
```diff
@@ -1033,6 +1033,8 @@ Language changes listed by -transition=id:
                     goto Lnoarg;
                 }
             }
+            else if (p[1] == '\0')
+                files.push("__stdin");
```
MartinNowak: Why not pushing `__stdin.d` to avoid ambiguities?

quickfur: Good idea, done.

quickfur:
Rebased, hopefully this fixes the travis error which seems unrelated.

@andralex  What else needs to be done to get this PR merged?

andralex:
@WalterBright 's call

quickfur:
@andralex I was referring to your change requested review.  @WalterBright already approved this PR.

andralex (approved):

wilzbach:
So why do we should we add verbose changelog entries?
(reminder for the future)

>> Mention passing files from stdin as potential use case
> thanks Seb, didn't know that.
https://github.com/lindt/docker-dmd/pull/5

Of course, it's in the changelog, but at the end and rather hard to spot:

image

https://dlang.org/changelog/2.076.0.html

ibuclaw:
Heh, I implemented this feature two months before this PR, and it seems to have the same `__stdin`.

https://github.com/D-Programming-GDC/GDC/commit/dd81b9ad2ac9ccb9a68ecc30f6c72d497a15d9e1

PR #6881, opened 2017-06-09
Fix Issue 17474 - non-property being treated as a property [in Assign…

UplinkCoder:
…ment]

Since this is very counter-intuitive behavior I'd like to deprecate it.

### DIFF
--- a/changelog/deprecateNonPropertyAssignment.dd
+++ b/changelog/deprecateNonPropertyAssignment.dd
@@ -1,0 +1,16 @@
+the property assignment to non-@property functions with default argument has been deprecated.
+
+Rationale: in an assignment, it is ambiguous whether the property should be invoked with no argument and the result be assigned to, or the property should be invoked with one argument.
+
+---
+ref int* getTheIntPtr(string str = "Hello") {
+    assert(str !is null);
+    return gInt;
+}
+
+unittest {
+    int x;
+    getTheIntPtr = &x; // => parenthesis-less call => getThisIntPtr() = &x
+    getTheIntPtr = null; // null coerces to string => getThisIntPtr(null)
+}
+---
--- a/src/ddmd/expression.d
+++ b/src/ddmd/expression.d
@@ -261,6 +261,11 @@ extern (C++) bool isNeedThisScope(Scope* sc, Declaration d)
     return true;
 }
 
+static bool isProperty(FuncDeclaration fd)
+{
+    return ((cast(TypeFunction)fd.type).isproperty || fd.storage_class & STCproperty || fd.storage_class2 & STCproperty);
+}
+
 /***************************************
  * Pull out any properties.
  */
@@ -301,6 +306,8 @@ extern (C++) Expression resolvePropertiesX(Scope* sc, Expression e1, Expression 
 
             Expressions a;
             a.push(e2);
+            TypeFunction tf = null;
+
 
             for (size_t i = 0; i < os.a.dim; i++)
             {
@@ -311,11 +318,15 @@ extern (C++) Expression resolvePropertiesX(Scope* sc, Expression e1, Expression 
                         return new ErrorExp();
                     fd = f;
                     assert(fd.type.ty == Tfunction);
-                    TypeFunction tf = cast(TypeFunction)fd.type;
+                    tf = cast(TypeFunction)fd.type;
+
                 }
             }
             if (fd)
             {
+                if (!isProperty(fd) && (*tf.parameters)[0].defaultArg)
+                    deprecation(e1.loc, "assignment to non-property function with default argument `%s` is deprecated.", fd.toFullSignature);
+
                 Expression e = new CallExp(loc, e1, e2);
                 return e.semantic(sc);
             }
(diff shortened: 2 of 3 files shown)


UplinkCoder:
@WalterBright @andralex any objections to this ?

andralex:
This seems to also disallow this:

```d
ref int* getTheIntPtr();
getTheIntPtr = null; // deprecated by this PR
```

Is that correct?

UplinkCoder:
@andralex no.
the deprecation only trigger when the property behavior would have been invoked.
since there is no argument in your `getThisIntPtr`
it won't trigger the deprecation.

changelog/deprecateNonPropertyAssignment.dd
```diff
@@ -0,0 +1,16 @@
+the property assignment to non-@property functions has been deprecated.
```
andralex: I think we should do this: `Properties with exactly one defaulted value cannot be assigned to.` Is that correct?

changelog/deprecateNonPropertyAssignment.dd
```diff
@@ -0,0 +1,16 @@
...
+
+Since it can result it can result in fairly unintutive errors
```
andralex: Replace with: `Rationale: in an assignment, it is ambiguous whether the property should be invoked with no argument and the result be assigned to, or the property should be invoked with one argument.`

src/ddmd/root/array.d
```diff
@@ -199,12 +199,12 @@ public:
 struct BitArray
 {
 nothrow:
-    size_t length() const pure
+    @property size_t length() const pure
```
andralex: Hmmmm... this code should not be broken because it is not ambiguous. If you have `size_t length();` and `void length(size_t)` it is unambiguous in all contexts what the get and set mean.

Ambiguity occurs when a property has a default value and returns a ref.

test/runnable/property.d
```diff
@@ -1,4 +1,10 @@
-
+/*
+TEST_OUTPUT:
+---
+runnable/property.d(24): Deprecation: assignment to non-property function `int bar(int value)` is deprecated.
+runnable/property.d(47): Deprecation: assignment to non-property function `void prop(int v)` is deprecated.
+---
+*/
```
andralex: The tests in this file should still pass.

andralex (changes requested):
This is good work but deprecates a bit more than needed. I think we should (at least for now) deprecate properties with defaulted values. cc @WalterBright

UplinkCoder:
Ambiguity occurs, when the function has a default value, and does return a type where assignment is possible with the same type the default argument had.
Think opAssign or alias this.

UplinkCoder:
I really want to avoid such special casing though ...
Can we agree to deprecate the setter-property behavior for functions which are not @property and do not return void ?

changelog/deprecateNonPropertyAssignment.dd
UplinkCoder: such special casing will lead us a c++-like forest of rules and well special cases

UplinkCoder:
@andralex I made the criterion for the deprecation stricter.
I cannot make it much narrower as then it will probably take considerable time to check.

andralex:
We are looking for simple and clear rules indeed. How about this: "Properties must have either exactly zero or one parameter. If present, the parameter cannot be defaulted."

UplinkCoder:
@andralex thinking about it some more I think its acceptable to deprecate property assignment (setter behavior) for non-property functions with default argument.
 
PR has been amended.

UplinkCoder:
jenkins is failing because of a regression introduced by now calling dtors transitively. and has nothing todo with this pr.

src/ddmd/root/array.d
UplinkCoder: I'd like to leave the `@property` here because it is one.
Any other blockers @andralex

andralex:
I can't follow from the conversation what works and what doesn't. 

One great thing to do for this PR is a complete passing test with things that work, and a number of tests required to fail, each with one thing that doesn't work. That would give us a good picture of what the behavior is.

We also need changes to the spec wording. I can do that after I see the tests.

cc @WalterBright

src/ddmd/root/array.d
andralex: So I understand this is a stylistic change. The problem with it being mixed in this PR is that now I worry the old code would no longer work.

Stylistic changes are best left for a separate PR where introducing regressions is not a problem. FWIW syle-wise we want to have as few property annotations as possible, not as many. So you'll have resistance from me if you tried to add property annotations without necessity.

UplinkCoder:
@andralex I added the failing case as test/fail_compilation/issue17474.d
which fails because of the function assigned to is not annotated `@property` and has a default argument.
If it were annotated property we would allow the ambiguous behavior.

tldr; the deprecation does only affect the one ambiguous case in issue 17474.
Everything else just works as expected which can be seen by the test-suite passing without modification.

JinShil:
@UplinkCoder Are you still interested in pursuing this?  If not, I'd like to adopt it.

AndrewEdwards:
@UplinkCoder - Need to know if you are still pursuing this. If you are, please update.

UplinkCoder:
I don't think this will make it

PR #6882, opened 2017-06-10
CircleCi/CodeCov: Don't use multiple threads for the DMD testsuite

wilzbach:
I am trying to remove the unrelated coverage changes from the CodeCov output and it seems like running the test suite in multiple threads might be problematic.
I don't know yet how the running time of CircleCi will be with this change, but given that we have __4__ parallel builds and we just run coverage tests there (i.e. all other CI take longer), it shouldn't matter if this fixes the problem.

CC @CyberShadow 

Original thread: https://github.com/dlang/dmd/pull/6879#issuecomment-307535894


CyberShadow:
Let's see if this helps with the random failures. If not, I guess we can revert it to speed it up later.

PR #6883, opened 2017-06-10
allow $(test_results) to specify an absolute path

rainers:
I stumbled across this when trying to run the tests and the dmd directory was mounted through a network share.


WalterBright:
Auto-merge toggled on

PR #6884, opened 2017-06-10
enable runnable/testxmm for Win64

rainers:
This test is disabled for Win64, but has been working for quite a while.


WalterBright:
Auto-merge toggled on

PR #6885, opened 2017-06-10
Add reserved version identifiers for uClibc and musl C libraries.

ibuclaw:
Both GCC and LLVM supports these libraries (maybe LLVM with third-party patches).  Both are popular C libraries amongst embedded and mobile platforms.

https://github.com/dlang/dlang.org/pull/1691

### DIFF
--- a/src/ddmd/cond.d
+++ b/src/ddmd/cond.d
@@ -326,6 +326,8 @@ extern (C++) final class VersionCondition : DVCondition
             "CRuntime_DigitalMars",
             "CRuntime_Glibc",
             "CRuntime_Microsoft",
+            "CRuntime_Musl",
+            "CRuntime_UClibc",
             "unittest",
             "assert",
             "all",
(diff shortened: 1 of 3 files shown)


PetarKirov (approved):
LGTM, the vibe.d failure is unrelated.

src/ddmd/cond.d
```diff
@@ -326,6 +326,8 @@ extern (C++) final class VersionCondition : DVCondition
             "CRuntime_DigitalMars",
             "CRuntime_Glibc",
             "CRuntime_Microsoft",
+            "CRuntime_Musl",
+            "CRuntime_uClibc",
```
klickverbot: I believe this should be `_Uclibc` or `_UClibc` for consistency with `_Glibc`.

ibuclaw: I just went with whatever the project calls itself.  I would probably go for the latter if pushed.

klickverbot: glibc is also called glibc, and musl musl, so yes, I'd go with the latter.

ibuclaw:
Is it just me, or are the codecov reports broken?

dnadlinger (approved):

ibuclaw:
Two positive reviews -> auto-merge.

PR #6886, opened 2017-06-10
Disallow empty statements made by semicolons

Burgos:
This turns the previous warning when using a semicolon
for an empty statement to a compilation error.

### DIFF
--- a/src/ddmd/parse.d
+++ b/src/ddmd/parse.d
@@ -5113,7 +5113,7 @@ final class Parser(AST) : Lexer
             if (!(flags & PSsemi_ok))
             {
                 if (flags & PSsemi)
-                    warning(loc, "use '{ }' for an empty statement, not a ';'");
+                    deprecation("use '{ }' for an empty statement, not a ';'");
                 else
                     error("use '{ }' for an empty statement, not a ';'");
             }
--- a/changelog/disallowSemiEmpty.dd
+++ b/changelog/disallowSemiEmpty.dd
@@ -1,0 +1,5 @@
+Deprecate empty statements made by semicolon
+
+Creating empty statements using semicolons, for which
+was previously issued a warning is now causing a deprecation
+hint.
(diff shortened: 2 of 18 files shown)


src/ddmd/parse.d
```diff
@@ -4664,7 +4664,7 @@ final class Parser(AST) : Lexer
         case TOKlcurly:
             if (f.frequire || f.fensure)
                 error("missing body { ... } after in or out");
-            f.fbody = parseStatement(PSsemi);
+            f.fbody = parseStatement(0);
```
Burgos: I was going to make a new enum value `PSdefault` or so, but I saw `parseStatement` called with `0` in other places as well, so I left it like this.

src/ddmd/parse.d
```diff
@@ -5112,10 +5112,7 @@ final class Parser(AST) : Lexer
         case TOKsemicolon:
             if (!(flags & PSsemi_ok))
             {
-                if (flags & PSsemi)
-                    warning(loc, "use '{ }' for an empty statement, not a ';'");
-                else
-                    error("use '{ }' for an empty statement, not a ';'");
+                error("use '{ }' for an empty statement, not a ';'");
```
BBasile: You cant pass from warning to error. There must be a  deprecation period before.

ghost (changes requested):

WalterBright:
This must go through a period of time where it is deprecated first.

Burgos:
Ah, true. I got mislead by the compiler's source, which states this is deprecated (See the `PSsemi`'s ddoc). Will fix, thanks!

Burgos:
Changed to a deprecation.

WalterBright (approved):
Nice work!

Burgos:
I'm not sure what's up with Jenkins, but he also failed in the trivial #6891 PR

wilzbach:
> I'm not sure what's up with Jenkins, but he also failed in the trivial #6891 PR

Yes, the Jenkins server was down for a week and now we are playing goose hunting for the introduced regressions in downstream projects, see e.g. https://github.com/dlang/phobos/pull/5469
tl;dr: don't worry about it, the errors should be fixed soon.

nemanja-boric-sociomantic:
All clear, thanks!

PR #6887, opened 2017-06-11
Add Target.cppTypeMangle hook for vendor types.

ibuclaw:
This patch adds `Target.cppMangleType`, which returns a pointer to statically allocated string containing the target/vendor-specific mangling for the given type.

While this could apply for any type, have only added it to the basic, vector, and struct type visitor passes for now.  DMD does nothing with this, however these specific places should catch all of the following:

- `real` may be a `__float128` type mangled as `g` (alpha, ia64, ppc, systemz, sparc, x86)
- `real` may be mangled as a vendor type `U10__float128` (ppc) or `u9__float80` (ia64)
- `va_list` may be mangled as if it is in the "std" namespace `St9__va_list` (arm, aarch64)
- `__vector()` may differ depending on the default C++ ABI `Dv` (abi >= 4)

This replaces `Target.realislongdouble`.

See https://github.com/dlang/dmd/pull/6887/files?w=1 to read diff without whitespace changes.

Fixes the frontend part of bug reported against gdc: https://bugzilla.gdcproject.org/show_bug.cgi?id=247

FYI: @redstar / @klickverbot - this would affect you.

### DIFF
--- a/src/ddmd/cppmangle.d
+++ b/src/ddmd/cppmangle.d
@@ -699,9 +699,18 @@ public:
         }
         if (t.isConst())
             buf.writeByte('K');
-        if (p)
-            buf.writeByte(p);
-        buf.writeByte(c);
+
+        // Handle any target-specific basic types.
+        if (auto tm = Target.cppTypeMangle(t))
+        {
+            buf.writestring(tm);
+        }
+        else
+        {
+            if (p)
+                buf.writeByte(p);
+            buf.writeByte(c);
+        }
     }
 
     override void visit(TypeVector t)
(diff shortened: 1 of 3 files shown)


UplinkCoder:
@ibuclaw vendor-type is a wired name.
How about platform-exclusiv-type ?

ibuclaw:
> @ibuclaw vendor-type is a wired name.

It's a perfectly reasonable name.

> How about platform-exclusiv-type ?

Sounds too german. ☺️

ibuclaw:
I forgot to mention that this also allows compilers to use internal information they have on the default C++ ABI.  So if `-fabi-version=4` or greater is the default, then you can return `Dv` instead of the current `U8__vector`.

ibuclaw:
@redstar / @klickverbot - SYN.

ibuclaw:
@klickverbot - ACK.

PR #6888, opened 2017-06-11
Remove auto-generated .cg file in test/compilable/vcg-ast

wilzbach:
Running the full test suite generated the `test/compilable/vcg-ast.d.cg` which is visible in `git status`.
This file is generated by [test/compilable/vcg-ast.d](https://github.com/dlang/dmd/blob/master/test/compilable/vcg-ast.d)

Thus I guess the simplest solution is to ignore all `*.cg` files?

CC @UplinkCoder

### DIFF
--- a/test/compilable/vcg-ast.d
+++ b/test/compilable/vcg-ast.d
@@ -1,6 +1,7 @@
 module vcg;
 // REQUIRED_ARGS: -vcg-ast -o-
 // PERMUTE_ARGS:
+// POST_SCRIPT: rm compilable/vcg-ast.d.cg && echo
 
 template Seq(A...)
 {


UplinkCoder:
this should be fine.

WalterBright:
What is a .cg file?

UplinkCoder:
I use ` .cg` for codegen-ast-dumps.
Normally the extension is used for shaders

wilzbach:
> What is a .cg file?

https://github.com/dlang/dmd/pull/6762/files
https://github.com/dlang/dmd/blob/master/src/ddmd/mars.d#L1591

braddr:
This is the wrong fix, imho.  The right fix is to control the placement of the generated files, putting them in the generated directory along side all the other compiler output.

UplinkCoder:
@braddr I am pretty sure that we never want to commit the debug dumps.
They are not generated as part of the build but as part of a test

wilzbach:
> This is the wrong fix, imho. The right fix is to control the placement of the generated files, putting them in the generated directory along side all the other compiler output.

AFAICT the location of the `.cg` can't be controlled, the only thing that comes to my mind would be sth. like this:

```
// POST_SCRIPT: rm compilable/vcg-ast.d.cg && echo
```

@UplinkCoder btw shouldn't`-vcg-ast` be documented (it doesn't appear on the CLI) and accept a file location alternatively (for cases like this one)?

UplinkCoder:
@wilzbach vcg-ast is undocumented because we cannot gurantee a stable format.
storing the cg files in a seperate directory is not something I have thought much about ... we would need to disambiguate the file-name is this case.

WalterBright:
@wilzbach the code used to generate a file does not say what that file is and what the point of it is. If it is a temporary file generated by the test suite, it should be deleted after use by the test suite, however the rest of the temporary files are cleaned up.

I agree with @braddr on this.

wilzbach:
>  the code used to generate a file does not say what that file is and what the point of it is. If it is a temporary file generated by the test suite, it should be deleted after use by the test suite, however the rest of the temporary files are cleaned up.

I am just trying to clean up the test suite ;-)
I rebased this now to add the `POST_SCRIPT` workaround.

@UplinkCoder feel free to adopt/improve this if you have better ideas.

UplinkCoder:
@wilzbach why the `&& echo` ?
also can windows do rm ?

wilzbach:
> @wilzbach why the && echo ?

Because `POST_SCRIPTS` is called with the curent hash, i.e. without it will fail as follows:

```
Executing post-test script: rm compilable/vcg-ast.d.cg test_results/dcde44e27807b9b422ee0b4059451e511222fcd1159ffda8ca87280e98dab504

Test failed: test_results/dcde44e27807b9b422ee0b4059451e511222fcd1159ffda8ca87280e98dab504: No such file or directory
```

> also can windows do rm ?

Let's see what auto-tester says, but yeah in theory it should be `del` on Windows, but if bash is supported, then maybe `rm` as well?

WalterBright:
Please change the subject title of this PR as it no longer is doing what the title says.

braddr:
This change is marginally better than changing .gitignore.  But it's still wrong.  It's papering over the compiler doing the wrong thing with its output.  Please at least file a bug report so that at some point someone will fix the underlying problem.

wilzbach:
> This change is marginally better than changing .gitignore. But it's still wrong. It's papering over the compiler doing the wrong thing with its output. 

Yeah, but I don't intend to fix the underlying problem if it's an unofficial, undocumented flag.

>  Please at least file a bug report so that at some point someone will fix the underlying problem.

https://issues.dlang.org/show_bug.cgi?id=17500

JinShil:
@wilzbach Are you OK with closing this?

wilzbach:
Thanks a lot for the ping!
Hmmm I still don't like that the test suite generates this mess which shows up in my git status and as it seems that @UplinkCoder isn't planning on adding support to programmatically set the output location, this PR here is still the "easy fix"...
What do other people think?

JinShil:
Although not ideal, I think this is a reasonable fix.  I'm going to pull this in a few days if no objections are raised.  @wilzbach Please take a look at the Jenkins failures.

wilzbach:
> Although not ideal, I think this is a reasonable fix. I'm going to pull this in a few days if no objections are raised. @wilzbach Please take a look at the Jenkins failures.

The Jenkins failure was unrelated:

```
No package ddox was found matching the dependency 0.15.18
```

This means that the DUB registry was offline. I restarted the test.

JinShil:
> I restarted the test.

@wilzbach The restart didn't seem to work.  Also, it looks like this is one of those languishing PRs that's reached GitHub's 1000 status update limit.  You may need to rebase.

wilzbach:
> Also, it looks like this is one of those languishing PRs that's reached GitHub's 1000 status update limit. You may need to rebase.

Done. Though I'm not sure whether rebasing helps when the 1000 status updates are reached.

JinShil:
@wilzbach It appears the script may need to be modified for Windows.

JinShil:
Studying the mars.d source code a little, I may be able to modify this to leverage the `-od` switch.  Would that solve this problem?

wilzbach:
Superseded by https://github.com/dlang/dmd/pull/7462

>  I may be able to modify vcg-ast to leverage the -od switch. Would that solve this problem?

I think that would be a nice improvement in any case, but not particularly important as this switch is mainly used for debugging.

PR #6875, opened 2017-06-09
fix Issue 17275 - [REG 2.072.0] AssertError@declaration.d(2121): Asse…

WalterBright:
…rtion failure

Replaces https://github.com/dlang/dmd/pull/6845

fix Issue 17275, 17289, 17337, 17338, 17339, 17344, 17489

PR #6889, opened 2017-06-11
Fix Issue 17489 - ICE in ddmd/argtypes.d

wilzbach:
This should change the ICE appearing on Jenkins to an normal regression error.

I am not sure whether this is the right way to fix the ICE or whether it should be catched higher.

### DIFF
--- a/src/ddmd/argtypes.d
+++ b/src/ddmd/argtypes.d
@@ -198,6 +198,8 @@ extern (C++) TypeTuple toArgTypes(Type t)
                 return t1;
             const sz1 = t1.size(Loc());
             const sz2 = t2.size(Loc());
+            if (t1.ty == Terror || t2.ty == Terror)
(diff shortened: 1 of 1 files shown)


WalterBright:
Note this on the stack trace:
```
??:? _Z10toArgTypesP4Type [0x4fecc9]
??:? _ZN17StructDeclaration12finalizeSizeEv [0x5704bf]
```
I corrected this in https://github.com/dlang/dmd/pull/6875 which hasn't been merged yet. I suggest waiting on this till #6875 is merged and seeing if it doesn't cure it.

wilzbach:
> I corrected this in #6875 which hasn't been merged yet. I suggest waiting on this till #6875 is merged and seeing if it doesn't cure it.

Yes, #6875 fixes this. Thanks a lot!

PR #6890, opened 2017-06-11
arrayop.d: color syntax highlighting

WalterBright:


src/ddmd/expression.d
```diff
@@ -11007,10 +11007,7 @@ extern (C++) final class NegExp : UnaExp
         if (tb.ty == Tarray || tb.ty == Tsarray)
         {
             if (!isArrayOpValid(e1))
-            {
-                error("invalid array operation %s (possible missing [])", toChars());
-                return new ErrorExp();
-            }
+                return arrayOpInvalidError(this);
```
braddr: A thought.. this feels like the wrong refactoring.  Consider moving the error() call to inside isArrayOpValid and changing this return to just "return new ErrorExp();"

WalterBright: This maintains the simple true/false result of `isArrayOpValid()` and the purity of it making it simpler to understand.

ZombineDev: I agree with @WalterBright.

PetarKirov (approved):
Looks good to me.

src/ddmd/expression.d
CyberShadow: The refactoring seems fine, but it should have been done in a separate commit or PR.

JinShil:
Not sure about the code changes, but the error messages look good to me.

andralex:
This is an entirely appropriate PR. Generally, we're experiencing a deflation of reviewers relative to PRs so I recommend favoring moving things along over minor blocking suggestions. Thanks.

andralex (approved):

CyberShadow:
@andralex No worries, @yebblies merges these regularly when he gets the chance to look at them.

WalterBright:
@cybershadow It's 18 days old. There are approx 1200 error() calls in dmd. Waiting 18 days to do a handful means they'll never get syntax highlighted.

CyberShadow:
> WalterBright requested a review from yebblies 20 hours ago

This isn't 18 days old though ;)

wilzbach:
> continuous-integration/jenkins/pr-merge — This commit cannot be built

CC @CyberShadow @MartinNowak - third request for escalation against the random build failures at the project-tester ;-)

CyberShadow:
I'll disable Jenkins tomorrow morning if nothing new happens.

yebblies:
>> WalterBright requested a review from yebblies 20 hours ago
> This isn't 18 days old though ;)

Exactly.  And @andralex knows how to contact me if there's something urgent.

PR #6892, opened 2017-06-11
Keep on optimizing until expression can't be simplified.

ibuclaw:
More interested in seeing what breaks - this is related to 17491.


ibuclaw:
Looks like I exposed a bug in code I wrote years ago.  Setting the result type after each expression building in `PowExp::optimize`.  Every other optimize visit pass does this.

ibuclaw:
Everything looks good, I'll raise a PR proper.

PR #6893, opened 2017-06-11
fix Issue 17491 - Compiles on invalid: *&s.init.var is not an lvalue

ibuclaw:
Same as #6892, but with added test case.

The idea here is that, the optimizer can simplify:
- `S(42).var` -> `42`, and
- `*&S(42).var` -> `S(42).var`

But it is not able to (in one pass) go from:
- `*&S(42).var` -> `42`

Because of this, the compiler accepts an assignment to non-lvalue as valid code because it hasn't sufficiently determined that the LHS is infact a constant literal.

The first commit is a prerequisite for this fix. `PowExp.optimize` is code that I wrote a few years back, and fixing 17491 exposed a hidden bug in them.

Fixes dmd bug: https://issues.dlang.org/show_bug.cgi?id=17491
Fixes gdc bug: https://bugzilla.gdcproject.org/show_bug.cgi?id=151

### DIFF
--- a/src/ddmd/optimize.d
+++ b/src/ddmd/optimize.d
@@ -1203,7 +1203,14 @@ extern (C++) Expression Expression_optimize(Expression e, int result, bool keepL
     }
 
     scope OptimizeVisitor v = new OptimizeVisitor(result, keepLvalue);
+    Expression ex = null;
     v.ret = e;
-    e.accept(v);
-    return v.ret;
+
+    // Optimize the expression until it can no longer be simplified.
+    while (ex != v.ret)
+    {
+        ex = v.ret;
+        ex.accept(v);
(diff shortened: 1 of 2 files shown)


WalterBright:
This seems to have two independent things conflated - changing how PowExp works, and the `*&s.init.var` issue. Could you please separate them?

ibuclaw:
I can't separate them here until the first commit is merged.

https://github.com/dlang/dmd/pull/6895

WalterBright:
> I can't separate them here until the first commit is merged. #6895

It's been merged.

ibuclaw:
> It's been merged.

Yes, and you may have noticed that I have already removed the other commit. ;-)

PR #6895, opened 2017-06-11
Always set the result type before returning from PowExp.optimize.

ibuclaw:
Bug fix from #6892, is a prerequisite for #6893. 

This allows optimizations to be recursively applied to ^^ expressions.  Many of which assume that `e.type` is not null.

Every other visit method in OptimizeVisitor does this (https://github.com/ibuclaw/dmd/blob/7e3d4e578eff50358f6ab35d7ec83f2d3bd838aa/src/ddmd/optimize.d#L352-L353)


WalterBright:
Auto-merge toggled on

PR #6897, opened 2017-06-12
fix Issue 17495 - __traits(getParameterStorageClasses) doesn't work w…

WalterBright:
…ith tuples as parameters

blocking https://github.com/dlang/phobos/pull/5474

### DIFF
--- a/src/ddmd/traits.d
+++ b/src/ddmd/traits.d
@@ -1021,9 +1021,9 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         }
         ex = ex.ctfeInterpret();
         auto ii = ex.toUInteger();
-        if (ii >= fparams.dim)
+        if (ii >= Parameter.dim(fparams))
         {
-            e.error("parameter index must be in range 0..%u not %s", cast(uint)fparams.dim, ex.toChars());
+            e.error("parameter index must be in range 0..%u not %s", cast(uint)Parameter.dim(fparams), ex.toChars());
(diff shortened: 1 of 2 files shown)


src/ddmd/traits.d
```diff
@@ -1021,7 +1021,7 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         }
         ex = ex.ctfeInterpret();
         auto ii = ex.toUInteger();
-        if (ii >= fparams.dim)
+        if (ii >= Parameter.dim(fparams))
```
rainers: I suspect the error message should show the new upper limit, too.

WalterBright: good catch

andralex:
So these are the other issue, right?

```d
../../.dub/packages/libasync-0.8.3/libasync/source/libasync/events.d(57,2): Error: destructor 'libasync.events.EventLoop.~this' is not nothrow
../../.dub/packages/libasync-0.8.3/libasync/source/libasync/events.d(57,2): Error: nothrow constructor 'libasync.events.EventLoop.this' may throw
```

CyberShadow:
> So these are the other issue, right?

The fix for the libasync error is here:

https://github.com/dlang/dmd/pull/6896

WalterBright:
> right?

It's the issue for getParameterStorageClassTuples.

It's a bug, regardless, and the fix for it.

rainers (approved):

PR #6898, opened 2017-06-12
Enable debug mode on BUILD=debug

wilzbach:
I expected `make -f posix.mak BUILD=debug` to work out of the box and generate a debug build of the compiler (with debug info!).

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -1,3 +1,31 @@
+################################################################################
+# Important variables:
+# --------------------
+#
+# HOST_CSS:				Host C++ compiler to use (g++,clang++)
+# HOST_DMD: 			Host D compiler to use for bootstrapping
+# AUTO_BOOTSTRAP:		Enable auto-boostrapping by download a stable DMD binary
+# INSTALL_DIR:			Installation folder to use
+#
+################################################################################
+# Build modes:
+# ------------
+# BUILD: release (default) | debug (enabled a build with debug symbols)
+#
+# Opt-in build features:
+#
+# ENABLE_RELEASE:		Optimized release built (set by BUILD=release)
+# ENABLE_DEBUG:			Add debug instructions and symbols (set by BUILD=debug)
+# ENABLE_WARNINGS: 		Enable C++ build warnings
+# ENABLE_PROFILING:		Build dmd with a profiling recorder (C++)
+# ENABLE_PGO_USE:		Build dmd with existing profiling information (C++)
+# 	PGO_DIR:			Directory for profile-guided optimization (PGO) logs
+# ENABLE_LTO:			Enable link-time optimizations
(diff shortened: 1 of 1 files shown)


CyberShadow:
I'm not sure these are equivalent. CC @yebblies

wilzbach:
> I'm not sure these are equivalent. CC @yebblies

`ENABLE_DEBUG` adds the debug flags:

```makefile
 ifdef ENABLE_DEBUG
 CXXFLAGS += -g -g3 -DDEBUG=1 -DUNITTEST		 CXXFLAGS += -g -g3 -DDEBUG=1 -DUNITTEST
 DFLAGS += -g -debug -unittest		 DFLAGS += -g -debug -unittest
 endif		 endif
```

whereas `BUILD=debug` just changes the folder (from what I can tell).

Burgos:
Yes, I was surprised by this behavior the other day as well.

src/posix.mak
```diff
@@ -27,6 +27,7 @@ ifneq ($(BUILD),release)
     ifneq ($(BUILD),debug)
         $(error Unrecognized BUILD=$(BUILD), must be 'debug' or 'release')
     endif
+	ENABLE_DEBUG := 1
```
Burgos: tabs vs. spaces as in the rest of the file.

WalterBright:
Please add documentation about what this does to the makefile, not just here.

src/posix.mak
wilzbach: There's no consistent style:

image

image

image


...

wilzbach:
> Please add documentation about what this does to the makefile, not just here.

This Makefile was more or less undocumented, so I started an attempt to document it.

wilzbach:
As no one objected for the last two months, can we move forward with this trivial improvement?

JinShil:
I ran into this confusion just the other day, so this PR is most welcome.  LGTM.

src/posix.mak
```diff
@@ -1,3 +1,31 @@
+################################################################################
+# Important variables:
+# --------------------
+#
```
WalterBright: The formatting is off on these comments - perhaps you're using tabs?

src/posix.mak
```diff
@@ -1,3 +1,31 @@
...
+# HOST_CSS:				Host C++ compiler to use (g++,clang++)
+# HOST_DMD: 			Host D compiler to use for bootstrapping
+# AUTO_BOOTSTRAP:		Enable auto-boostrapping by download a stable DMD binary
```
WalterBright: s/download/downloading/

src/posix.mak
```diff
@@ -1,3 +1,31 @@
...
+# INSTALL_DIR:			Installation folder to use
+#
+################################################################################
+# Build modes:
+# ------------
+# BUILD: release (default) | debug (enabled a build with debug symbols)
```
WalterBright: Probably ought to say that this is intended to be set on the command line

PR #6901, opened 2017-06-13
fix Issue 17499 - with -betterC switch, call C's assert failure function rather than druntime's

WalterBright:
The idea is that if the `-betterC` switch is thrown, the C standard library assert is used rather than the druntime one, so that (at least for this) druntime is not necessary.


andralex:
Makes sense. I have bigger plans for assert, but this doesn't stand in their ways.

wilzbach:
(closing and reopening to rekick Travis and Jenkins)

PR #6902, opened 2017-06-13
fix Issue 17489 - ICE in ddmd/argtypes.d

MartinNowak:
More info on the ICE https://github.com/dlang/dmd/pull/6894#issuecomment-308176723.

### DIFF
--- a/src/ddmd/aggregate.d
+++ b/src/ddmd/aggregate.d
@@ -266,17 +266,21 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
             if (v.storage_class & STCref)
                 return 0;
             auto tv = v.type.baseElemOf();
-            if (tv.ty != Tstruct)
-                return 0;
-            if (ad == (cast(TypeStruct)tv).sym)
+            // Propagate pending errors from fields to aggregate.
+            // https://issues.dlang.org/show_bug.cgi?id=17489
+            if (v.type.ty == Terror)
+                goto Lerror;
+            if (tv.ty == Tstruct && ad == (cast(TypeStruct)tv).sym)
             {
                 const(char)* psz = (v.type.toBasetype().ty == Tsarray) ? "static array of " : "";
                 ad.error("cannot have field `%s` with %ssame struct type", v.toChars(), psz);
-                ad.type = Type.terror;
-                ad.errors = true;
-                return 1;
+                goto Lerror;
             }
             return 0;
+        Lerror:
+            ad.type = Type.terror;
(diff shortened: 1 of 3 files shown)


MartinNowak:
Mmh, changing anything in struct resolution can easily introduce new regressions, might not be worth the risk, given that the ICE code fails to parse correctly.
Any other ideas how to resolve this problem?

wilzbach:
> Mmh, changing anything in struct resolution can easily introduce new regressions, might not be worth the risk, given that the ICE code fails to parse correctly.

In particular if the test suite already segfaults:

```
../src/dmd -conf= -m32 -Ifail_compilation -verrors=0   -odgenerated/fail_compilation -ofgenerated/fail_compilation/ice8100_0.o -c fail_compilation/ice8100.d
fail_compilation/ice8100.d(10): Error: no property 'Q' for type 'ice8100.Bar!bool'
fail_compilation/ice8100.d(11): Error: template instance ice8100.Foo!(Bar!bool) error instantiating
fail_compilation/ice8100.d(12):        instantiated from here: Bar!bool
Segmentation fault (core dumped)
```

WalterBright:
Is the failing call to `toArgTypes()` coming from `finalizeSize()` ?

MartinNowak:
> Is the failing call to toArgTypes() coming from finalizeSize() ?

See https://github.com/dlang/dmd/pull/6894#issuecomment-308176723 for an explanation @WalterBright.

WalterBright:
> The root of the problem is that immutable() m_nodes; is a parse error which bumps global.errors and returns a Type.terror already during parseType. This error is later not detected by StructDeclaration::semantic for Path when it runs semantic for all of it's members, b/c it only compares global.errors at the begin and end of StructDeclaration::semantic. Since the error type isn't propagated from the field m_nodes to the struct Path, we make it all the way to finalizeSize of DirectoryChange, where the assertion is triggered when setting the offset for field path. Doesn't happen when the module is part of the compilation, since we eagerly abort on any parse errors.

It strongly looks to me like detecting the error in `toArgTypes()` is the wrong place. The fix should be fixing "the error type isn't propagated from the field m_nodes to the struct Path".

I.e. the error scheme used in dmd is a `poisoning` one, where errors propagate such that any declaration that depends on something that errored is itself marked as erroneous. The same holds for types, statements, and expressions.

WalterBright:
My take on where this fix should be https://github.com/dlang/dmd/pull/6914

WalterBright:
Superseded by https://github.com/dlang/dmd/pull/6914

PR #6903, opened 2017-06-14
improve some comments and minor stuff in func.d

WalterBright:


PetarKirov (approved):
Looks good

quickfur:
Looks good.

rainers (approved):
I assume these function are not called by GDC as I checked that LDC doesn't have to.

PR #6906, opened 2017-06-14
Fix issue 17502: Allow no parameters in out contract for auto methods

Burgos:
Even in case of non-void methods, their out contract should
be able to accept no arguments. For the methods with auto return type,
there was oversight where one argument to `out` was always assumed.

### DIFF
--- a/test/compilable/test17502.d
+++ b/test/compilable/test17502.d
@@ -1,0 +1,30 @@
+class Foo
+{
+    auto foo()
+    out {}
+    body {}
+
+    auto bar()
+    out { assert (__result > 5); }
+    body { return 6; }
+
+    auto bar_2()
+    out (res) { assert (res > 5); }
+    body { return 6; }
+
+    int concrete()
+    out { assert(__result > 5); }
+    body { return 6; }
+
+    int concrete_2()
+    out(res) { assert (res > 5); }
+    body { return 6; }
+
+    void void_foo()
+    out {}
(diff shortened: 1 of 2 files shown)


MetaLang:
It may be worth including another test function that returns a concrete type such as int, as functions with a return type of `auto` are template functions.

Burgos:
Good point, updated with more test cases.

test/compilable/test17502.d
```diff
@@ -0,0 +1,30 @@
+class Foo
+{
+    auto foo()
+    out {}
+    body {}
+
+    auto bar()
+    out { assert (__result > 5); }
```
mathias-lang-sociomantic: I'm not sure using that private identifier is something we want to advocate / test :D

Burgos: Well, it's an internal compiler test, I guess, so I was going with the thoughts that whoever removes `__result`, should do it contently and remove this test. Reason I've added it here in the first place is that this patch touches the code responsible for the generation of this identifier.

Burgos:
Damn Jenkins, why can't he leave me alone.

test/compilable/test17502.d
WalterBright: As far as the test suite goes, it is not intended to be an example of best practices. It's sole purpose is to test the compiler - so whatever works is fair game. (In fact, it's often terrible style.)

PR #6907, opened 2017-06-15
convert divcoeff.c to D

WalterBright:
first one in the back end!

### DIFF
--- a/src/ddmd/backend/divcoeff.d
+++ b/src/ddmd/backend/divcoeff.d
@@ -5,7 +5,7 @@
  * Copyright:   Copyright (c) 2013-2017 by The D Language Foundation, All Rights Reserved
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
- * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/ddmd/backend/divcoeff.c, backend/divcoeff.c)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/ddmd/backend/divcoeff.d, backend/divcoeff.d)
  */
 
 /***************************************************
@@ -13,27 +13,36 @@
  * by Torbjoern Granlund and Peter L. Montgomery
  */
 
-#include <stdio.h>
-#include <assert.h>
+import core.stdc.stdio;
 
-// This MUST MATCH typedef targ_ullong in cdef.h.
-#if defined(__UINT64_TYPE__) && !defined(__APPLE__)
-typedef __UINT64_TYPE__ ullong;
-#elif defined(__UINTMAX_TYPE__)
-typedef __UINTMAX_TYPE__ ullong;
-#else
-typedef unsigned long long ullong;
-#endif
+extern (C++):
 
-void test_udiv_coefficients();
+alias ullong = ulong;
 
 /* unsigned 128 bit math
  */
 
-#define SIGN64(x) ((long long)(x) < 0)
-#define SHL128(dh,dl, xh,xl) ((dh = (xh << 1) | SIGN64(xl)),  (dl = xl << 1))
-#define SHR128(dh,dl, xh,xl) ((dl = (xl >> 1) | ((xh & 1) << 63)),(dh = xh >> 1))
-#define XltY128(xh,xl,yh,yl) (xh < yh || (xh == yh && xl < yl))
+bool SIGN64(ullong x)
+{
+    return cast(long)x < 0;
+}
+
+void SHL128(out ullong dh, out ullong dl, ullong xh,ullong xl)
+{
+    dh = (xh << 1) | SIGN64(xl);
+    dl = xl << 1;
+}
+
+void SHR128(out ullong dh, out ullong dl, ullong xh,ullong xl)
+{
+    dl = (xl >> 1) | ((xh & 1) << 63);
+    dh = xh >> 1;
+}
+
+bool XltY128(ullong xh, ullong xl, ullong yh, ullong yl)
+{
+    return xh < yh || (xh == yh && xl < yl);
+}
 
 void u128Div(ullong xh, ullong xl, ullong yh, ullong yl, ullong *pqh, ullong *pql)
 {
@@ -43,10 +52,10 @@ void u128Div(ullong xh, ullong xl, ullong yh, ullong yl, ullong *pqh, ullong *pq
 
     //ullong xxh = xh, xxl = xl, yyh = yh, yyl = yl;
 
-    assert(yh || yl);           // no div-by-0 bugs
+    //assert(yh || yl);           // no div-by-0 bugs
 
     // left justify y
-    unsigned shiftcount = 1;
+    uint shiftcount = 1;
     if (!yh)
     {   yh = yl;
         yl = 0;
@@ -85,31 +94,32 @@ void u128Div(ullong xh, ullong xl, ullong yh, ullong yl, ullong *pqh, ullong *pq
 
     // Remainder is xh,xl
 
-#if 0
-    printf("%016llx_%016llx / %016llx_%016llx = %016llx_%016llx\n", xxh,xxl,yyh,yyl,qh,ql);
-    if (xxh == 0 && yyh == 0)
-        printf("should be %llx\n", xxl / yyl);
-#endif
+    version (none)
+    {
+        printf("%016llx_%016llx / %016llx_%016llx = %016llx_%016llx\n", xxh,xxl,yyh,yyl,qh,ql);
+        if (xxh == 0 && yyh == 0)
+            printf("should be %llx\n", xxl / yyl);
+    }
 }
 
 /************************************
  * Implement Algorithm 6.2: Selection of multiplier and shift count
- * Input:
- *      N       32 or 64
- *      d       divisor (must not be 0 or a power of 2)
- *      prec    bits of precision desired
+ * Params:
+ *      N =     32 or 64
+ *      d =     divisor (must not be 0 or a power of 2)
+ *      prec =  bits of precision desired
  * Output:
- *      *pm     factor
- *      *pshpost post shift
+ *      *pm =      factor
+ *      *pshpost = post shift
  * Returns:
  *      true    m >= 2**N
  */
 
 bool choose_multiplier(int N, ullong d, int prec, ullong *pm, int *pshpost)
 {
-    assert(N == 32 || N == 64);
-    assert(prec <= N);
-    assert(d > 1 && (d & (d - 1)));
+    //assert(N == 32 || N == 64);
+    //assert(prec <= N);
+    //assert(d > 1 && (d & (d - 1)));
 
     // Compute b such that 2**(b-1) < d <= 2**b
     // which is the number of significant bits in d
@@ -127,10 +137,10 @@ bool choose_multiplier(int N, ullong d, int prec, ullong *pm, int *pshpost)
     if (N == 32)
     {
         // mlow = (2**(N + b)) / d
-        ullong mlow = (1ULL << (N + b)) / d;
+        ullong mlow = (1UL << (N + b)) / d;
 
         // uhigh = (2**(N + b) + 2**(N + b - prec)) / d
-        ullong mhigh = ((1ULL << (N + b)) + (1ULL << (N + b - prec))) / d;
+        ullong mhigh = ((1UL << (N + b)) + (1UL << (N + b - prec))) / d;
 
         while (mlow/2 < mhigh/2 && shpost)
         {
@@ -140,7 +150,7 @@ bool choose_multiplier(int N, ullong d, int prec, ullong *pm, int *pshpost)
         }
 
         *pm = mhigh & 0xFFFFFFFF;
-        mhighbit = mhigh >> N;
+        mhighbit = (mhigh >> N) != 0;
     }
     else if (N == 64)
     {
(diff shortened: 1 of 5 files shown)


jacob-carlborg:
There's a Visual Stdio project file as well.

WalterBright:
I'm having enough trouble getting these makefiles to work.

WalterBright:
It looks like circleci is using an older dmd to compile. It needs to be upgraded to 2.074.1

wilzbach:
> It looks like circleci is using an older dmd to compile. It needs to be upgraded to 2.074.1

Simply bump it here:

https://github.com/dlang/dmd/blob/master/circleci.sh#L5

WalterBright:
Thanks! That did the trick on circleci.

MartinNowak:
As said per mail, neither gdc nor ldc support anything like `-mv` which would restrict us to dmd to build releases. Looking at ~30% speedup for dmd using a better optimizing compiler, that doesn't seem like a smart move.

MartinNowak:
Sorry, doesn't look like this is using `-mv` at all. For what would that be needed?

MartinNowak:
Some info on why you need to update the required compiler would be helpful!
Also need to bump the auto_bootstrap version in posix.mak.

src/ddmd/backend/divcoeff.d
```diff
@@ -6,35 +6,44 @@
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     Distributed under the Boost Software License, Version 1.0.
  *              http://www.boost.org/LICENSE_1_0.txt
- * Source:      https://github.com/dlang/dmd/blob/master/src/ddmd/backend/divcoeff.c
+ * Source:      https://github.com/dlang/dmd/blob/master/src/ddmd/backend/divcoeff.d
```
wilzbach: How about?

```
$(DMDSRC backend/_divcoeff.d)
```

ibuclaw:
My guess is that some compilers (or ancient perl wrappers) don't understand `-betterC`.

MartinNowak:
> My guess is that some compilers (or ancient perl wrappers) don't understand -betterC.

Ah yes, [that](https://travis-ci.org/dlang/dmd/jobs/243205245#L588)'s another problem. Why do we need `-betterC` for the backend? Is this some `dmc` tribute?

ibuclaw:
> Ah yes, that's another problem. Why do we need -betterC for the backend? Is this some dmc tribute?

Well, it looks like `-betterC` is the reason why jenkins fails.

https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-6907/9/pipeline#step-74-log-128

You *could* workaround by building with `-release`, but I already am not a fan of the travisci tests doing that.

WalterBright:
> Some info on why you need to update the required compiler would be helpful!

The older compiler generated code that was looking for the module assert() function in `-betterC` mode, which had been removed.

WalterBright:
Essentially, it's time to move forward with converting the backend to D. To that end, gdc, ldc, and the build process needs to support:
1. -mv switch
2. removal of the generated assert helper functions
3. https://github.com/dlang/dmd/pull/6901
None of these are tricky or difficult. To that end, divcoeff.d is a trial run to see where the issues are in the toolchain for doing this. Looks like it's doing its job :-)

ibuclaw:
> #6901

Making compilers support features that have not made it to stable (or even in a release) is probably asking for a bit too much.

WalterBright:
> Making compilers support features that have not made it to stable (or even in a release) is probably asking for a bit too much.

These changes are simple and not controversial. Let's move forward with them.

src/ddmd/backend/divcoeff.d
andralex: yes please

andralex:
OK I'll pull the switch on this to get in once it passes tests.

andralex (approved):

wilzbach:
CC @CyberShadow for updating DAutoTest

CyberShadow:
@WalterBright Just so we're on the same page, please describe the procedure for building DMD master (incl. this pull request) from scratch without a host DMD compiler (C++ only).

MartinNowak:
Didn't we just agree that 30% speedup from building dmd with gdc is worthwhile for an upcoming release? This would block this option for quite a while. Also jumping ahead with using latest edge features sets quite a bad precedence for dmd/gdc/ldc compatibility.
But most important, it's still unclear why we even need `-betterC` here.

ibuclaw:
Consider that the oldest compiler building DMD IIRC is using v2.066 of the front-end.  I think it's safe to do a leap to 2.068, possibly 2.070 at latest, but I don't really want to break builds when using a release that is not strictly that old.

WalterBright:
> please describe the procedure for building DMD master (incl. this pull request) from scratch without a host DMD compiler (C++ only).

That's a great question, but since building DMD already requires a D compiler, it is off-topic for this PR. Let's keep this thread focused on this PR.

WalterBright:
> Didn't we just agree that 30% speedup from building dmd with gdc is worthwhile for an upcoming release?

Yes.

> This would block this option for quite a while.

Not really. The 3 changes needed to the D compiler to support this, as described, are easy and minimal, with little risk of disruption.

> Also jumping ahead with using latest edge features sets quite a bad precedence for dmd/gdc/ldc compatibility.

Getting DMD in 100% D also has advantages. I've been waiting since the front end was converted to D to get moving on this. How many years should this take?

> Most important though, it's still unclear why we even need -betterC here.

1. -betterC needs to move forward. It'll never happen until we start dogfooding it.

2. The back end is designed to run with asserts on. Turning them off, like is done in the front end, is unacceptable. If -release were left off, the invariant checks for struct member function calls is also unacceptable. -betterC makes the asserts behave like C asserts.

CyberShadow:
> it is off-topic for this PR

It's about as on-topic as updating CI services to support this change, which is not going to happen without discussion and acknowledgement from all sides of the full impact of changes like these.

ibuclaw:
> Just so we're on the same page, please describe the procedure for building DMD master (incl. this pull request) from scratch without a host DMD compiler (C++ only).

I wouldn't use anything other than a release compiler from the downloads page for DAutoTest.  We have bootstrap tests as part of the travisCI scripts.

ibuclaw:
> Not really. The 3 changes needed to the D compiler to support this, as described, are easy and minimal, with little risk of disruption.

It would be a major disruption, as suddenly you cannot use the current DMD release, nor any of the prior eight major releases to build master.  Effectively our entire CI infrastructure will be broken.

However, I'm all up for trimming the hedges a bit, if for @CyberShadow that means supporting a D compiler that can be built using C++, then we can set the new baseline at 2.068 - dropping 2.066 and 2.067 (ping @braddr for confirmation that auto-tester is using >= 2.068).

If dropping the DMD C++ bootstrap requirement altogether is more desired, I would be OK to stretch up until 2.071, which is approximately the feature set supported by GDC (no binaries available though).

You're not the first person to try submitting a patch that depends on a feature more recent than 2.066, and as a rule of thumb, there really isn't that much of value in more recent versions of DMD that would be a showstopper for anything that we'd want included in the compiler implementation.

CyberShadow:
> supporting a D compiler that can be built using C++, then we can set the new baseline at 2.068

IIRC, 2.068 is the first version that can no longer be (directly) built with just a C++ compiler.

> If dropping the DMD C++ bootstrap requirement altogether is more desired,

I'm guessing you actually mean adding another step to the bootstrap process?

ibuclaw:
> IIRC, 2.068 is the first version that can no longer be (directly) built with just a C++ compiler.

Ah, I'm forgetting about idgen.d - I was only thinking about the rest of the compiler.

> I'm guessing you actually mean adding another step to the bootstrap process?

I meant using a front end that is entirely implemented in D.

wilzbach:
@WalterBright regarding:

>> Didn't we just agree that 30% speedup from building dmd with gdc is worthwhile for an upcoming release? This would block this option for quite a while. Also jumping ahead with using latest edge features sets quite a bad precedence for dmd/gdc/ldc compatibility.
>> Most important though, it's still unclear why we even need -betterC here.


> Not really. The 3 changes needed to the D compiler to support this, as described, are easy and minimal, with little risk of disruption.

Hehe, it seems that reality proves that it's not minimal.

> -betterC needs to move forward. It'll never happen until we start dogfooding it.

It would also help `-betterC`, if we document it for a start: https://github.com/dlang/dlang.org/pull/1796
In any case, as far as I can see the situation you have three options:

(A) Wait and hope that LDC/GDC incorporate your changes soon (for LDC that might happen fairly soon, but idk about GDC)
(B) Accept [DIP1006](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1006.md) and build the backend without `-betterC`, but with `assert`s
(C) Roll your own runtime `assert`

> The back end is designed to run with asserts on. Turning them off, like is done in the front end, is unacceptable. If -release were left off, the invariant checks for struct member function calls is also unacceptable. -betterC makes the asserts behave like C asserts.

Then accept [DIP1006](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1006.md) ;-)

kinke:
> (A) Wait and hope that LDC/GDC incorporate your changes soon (for LDC that might happen fairly soon, but idk about GDC)

LDC 1.3, released a week ago, supports `-mv`.

WalterBright:
@kinke thanks, that's great news! add betterC too, please!

kinke:
We also support `-betterC`, but I think it only skips the emission of ModuleInfos. I hope that suffices. ;)

WalterBright:
@wilzbach I approved https://github.com/dlang/dlang.org/pull/1796 pending a couple trivial edits.

MartinNowak:
We could just add `-betterC` as noop to the gdmd wrapper script.
https://github.com/D-Programming-GDC/GDMD/blob/master/dmd-script

andralex:
OK folks, status of this? We need to plow forward. Thanks! cc @ibuclaw @MartinNowak

kinke:
I implemented `-betterC` for LDC in https://github.com/ldc-developers/ldc/pull/2365. We've just released LDC 1.5-beta1, so I'd guess that one more month would be needed for 1.6-beta1 which would include that PR.

WalterBright:
> one more month would be needed

Great! Ready when you are! @ibuclaw ?

andralex:
ping @ibuclaw

ibuclaw:
Wouldn't want to give a precise date, but gcc-8 is first this would land in.  As @MartinNowak said, you are free to make `-betterC` a noop for the dmd-script wrapper.

WalterBright:
> gcc-8 is first this would land in

When is that expected?

joakim-noah:
[Looks like early next year](https://www.phoronix.com/scan.php?page=news_item&px=GCC-8-Stage-3-Soon).

wilzbach:
Regarding the current failure: I think you need to add `G_DOBJS` to the new `unittest` target in the `posix.mak` or probably better: add it to `D_SRCS`.

JackStouffer:
Walter, the bot comment is going to keep coming back every time there's a new commit. If it doesn't see it's own comment to update, it makes a new one.

WalterBright:
> If it doesn't see it's own comment to update, it makes a new one.

That's not working, then, because I'd delete 10 of them in a row, not one at a time.

andralex:
so... pull?

wilzbach:
> so... pull?

AFAICT we are still waiting for GDC to catch up or someone to fork the `gdmd` script and make `-betterC` a noop:

> Wouldn't want to give a precise date, but gcc-8 is first this would land in. As @MartinNowak said, you are free to make -betterC a noop for the dmd-script wrapper.

WalterBright:
Rebooted as https://github.com/dlang/dmd/pull/7714 because I could not get `git rebase` to work with renamed files (it kept insisting I had not merged them).

wilzbach:
> I could not get git rebase to work with renamed files (it kept insisting I had not merged them).

Not sure what you did, but it works fine for me:

```shell
> go pr/6907
> git rebase upstream/master
First, rewinding head to replay your work on top of it...
Applying: convert divcoeff.c to D
Using index info to reconstruct a base tree...
A	circleci.sh
A	src/ddmd/backend/cod2.c
A	src/ddmd/backend/divcoeff.c
M	src/posix.mak
M	src/win32.mak
Falling back to patching base and 3-way merge...
Auto-merging src/win32.mak
Auto-merging src/posix.mak
CONFLICT (content): Merge conflict in src/posix.mak
CONFLICT (rename/rename): Rename "src/ddmd/backend/divcoeff.c"->"src/dmd/backend/divcoeff.c" in branch "HEAD" rename "src/ddmd/backend/divcoeff.c"->"src/ddmd/backend/divcoeff.d" in "convert divcoeff.c to D"
Auto-merging src/dmd/backend/cod2.c
Auto-merging .circleci/run.sh
error: Failed to merge in the changes.
Patch failed at 0001 convert divcoeff.c to D
The copy of the patch that failed is found in: .git/rebase-apply/patch

Resolve all conflicts manually, mark them as resolved with
"git add/rm <conflicted_files>", then run "git rebase --continue".
You can instead skip this commit: run "git rebase --skip".
To abort and get back to the state before "git rebase", run "git rebase --abort".

> git mergetool

Merging:
src/ddmd/backend/divcoeff.c
src/ddmd/backend/divcoeff.d
src/dmd/backend/divcoeff.c
src/posix.mak

Deleted merge conflict for 'src/ddmd/backend/divcoeff.c':
  {local}: deleted
  {remote}: deleted
Use (m)odified or (d)eleted file, or (a)bort? d

Deleted merge conflict for 'src/ddmd/backend/divcoeff.d':
  {local}: deleted
  {remote}: created file
Use (c)reated or (d)eleted file, or (a)bort? c

Deleted merge conflict for 'src/dmd/backend/divcoeff.c':
  {local}: created file
  {remote}: deleted
Use (c)reated or (d)eleted file, or (a)bort? d

Normal merge conflict for 'src/posix.mak':
  {local}: modified file
  {remote}: modified file
Hit return to start merge resolution tool (kdiff3): 
```

PR #6909, opened 2017-06-15
win64: speed up test suite by more than 15 minutes

rainers:
runnable/test17338.d added in https://github.com/dlang/dmd/pull/6718 triggers pretty bad identical comdat folding performance by the MS linker.

As this is known to have problems anyway (e.g. https://issues.dlang.org/show_bug.cgi?id=10664) and is disabled in sc.ini (https://github.com/dlang/dmd/blob/master/ini/windows/bin/sc.ini#L23), we can disable it in the tests, too.

### DIFF
--- a/test/d_do_test.d
+++ b/test/d_do_test.d
@@ -556,6 +556,11 @@ int main(string[] args)
                 (testArgs.mode == TestMode.FAIL_COMPILE ? "-verrors=0 " : null) ~
                 testArgs.requiredArgs;
 
+            // https://issues.dlang.org/show_bug.cgi?id=10664: exceptions don't work reliably with COMDAT folding
+            // it also slows down some tests drastically, e.g. runnable/test17338.d
+            if (msc)
+                reqArgs ~= " -L/OPT:NOICF";
+
             string compile_output;
             if (!testArgs.compileSeparately)
             {


rainers:
The build time for Win_32_64 on win-farm1 has gone down from 43 minutes for other PRs to less than 26 minutes.

WalterBright:
A comment in the code explaining why this switch is there would be helpful, as I can't see someone guessing its purpose, as the documentation for NOICF doesn't offer a clue.

rainers:
> A comment in the code explaining why this switch is there would be helpful

Amended a comment.

MartinNowak:
Quite a huge speedup.
firefox_screenshot_2017-06-16t18-16-31 686z

Does this only come from runnable/test17338.d?

Of course we should fix [Issue 10664 – Win64: exception handling does not work with COMDAT folding](https://issues.dlang.org/show_bug.cgi?id=10664) at some point, and maybe even explicity use `/OPT:REF,ICF` for release builds as recommended [-OPT (Optimizations)](https://msdn.microsoft.com/en-us/en-en/library/bxwfs976.aspx).

Using the equivalent for non-Windows linkers would be nice as well.
[c++ - GCC(/Clang): Merging functions with identical instructions (COMDAT folding)](https://stackoverflow.com/a/15169073/2371032)

MartinNowak:
Opened an issue for the ld.gold failure [Issue 17515 – ld.gold identical comdat folding triggers an assertion](https://issues.dlang.org/show_bug.cgi?id=17515).

rainers:
> Does this only come from runnable/test17338.d?

At least most of it. It took more than 15 minutes to link it on my computer. 

> Opened an issue for the ld.gold failure Issue 17515 – ld.gold identical comdat folding triggers an assertion.

Interesting. Hopefully that helps narrowing down the problem for windows, too.

PR #6412, opened 2017-01-08
fix Issue 17059 - incorrect circular reference with `is(Klass : Interface)`

MartinNowak:
- fixed by tweaking the fix for Issue 16980 (#6383)
- happened b/c TypeClass::implicitConv -> TypeClass::constConv checks
  for offset == 0, which triggered size finalization
- instead handle unfinalized classes in isBaseOf (new OFFSET_FWDREF)
  and only explicity finalize size before optimizing casts
- add a halt for any OFFSET_FWDREF leaking through to IR gen

PR #6911, opened 2017-06-16
fix typo in assertion

MartinNowak:


src/ddmd/optimize.d
```diff
@@ -610,7 +610,7 @@ extern (C++) Expression Expression_optimize(Expression e, int result, bool keepL
                 // https://issues.dlang.org/show_bug.cgi?id=16980
                 cdfrom.size(e.loc);
                 assert(cdfrom.sizeok == SIZEOKdone);
-                assert(cdto.sizeok == SIZEOKdone || !cdto.isBaseOf(cdto, null));
+                assert(cdto.sizeok == SIZEOKdone || !cdto.isBaseOf(cdfrom, null));
```
ibuclaw: I knew it!

ibuclaw (approved):

WalterBright:
Shouldn't there be a test case?

ibuclaw:
@WalterBright, the test case was in #6412.  I gave attention to this because the assert doesn't catch what it was intended for.

PR #6821, opened 2017-05-22
Fix issue 6227 - Deprecate comparison of different enums (revamp)

UplinkCoder:
This is a rebased version of #6444.
@andralex ready for merge

PR #6912, opened 2017-06-17
Fix #6821 changelog entry

CyberShadow:
This is a partial revert of 0d40de43e9fbe41817a9ace030260a94fd211cc1. This commit introduced a number of problems:

1. The title was two lines, which did not conform to the syntax as described in README.md, and caused the generated changelog to be misformed (see http://dlang.org/changelog/2.075.0_pre.html)

2. There was trailing whitespace on the second line, causing dlang.org's test target to fail

3. Subjectively, I'm having trouble understanding the new wording. The old wording was much more concise, and if it was factually incorrect, I don't see how.

CC @UplinkCoder @wilzbach

### DIFF
--- a/changelog/b6227.dd
+++ b/changelog/b6227.dd
@@ -1,5 +1,4 @@
-Comparison that takes into account the enumerated type improves code correctness by
-disallowing comparison of unrelated enumerated types. 
+Comparison of values belonging to different enums is deprecated.
 
 This change does not affect anonymous enumerations.
 


wilzbach (approved):
Thanks!

wilzbach:
>  This commit introduced a number of problems:
> The title was two lines, which did not conform to the syntax as described in README.md, and caused the generated changelog to be misformed (see http://dlang.org/changelog/2.075.0_pre.html)
> There was trailing whitespace on the second line, causing dlang.org's test target to fail

That's why I'm setting it directly to auto-deploy merge.
@UplinkCoder I hardly doubt that you have anything against this, but if so, we can always do this, once the test target is green again and we have enabled an enforcement check for its passing it ;-)

UplinkCoder:
It's fine.
Unfortunately I am not used to ddoc :)

PR #6916, opened 2017-06-18
Actually fix path of generated DMD

CyberShadow:
As `OS` was never defined, it was substituted as an empty string in the `G` variable.

This is a fixup for #6873.

I don't see where `OS` is being set by the autotester (if anywhere) - if it's not, then it's still using the current path hierarchy instead of the intended POSIX one. Let's see what the autotester says.

### DIFF
--- a/src/win32.mak
+++ b/src/win32.mak
@@ -61,6 +61,7 @@
 # fixed model for win32.mak, overridden by win64.mak
 MODEL=32
 BUILD=release
+OS=windows
 
 ##### Directories
 


CyberShadow:
> I don't see where `OS` is being set by the autotester (if anywhere)

OK, I see. The win32.mak makefile still creates a `dmd\src\dmd.exe` file, which the auto-tester still uses.

I've confirmed that this pull request fixes https://github.com/CyberShadow/Digger/issues/58 (thus making the path actually the same as on POSIX), and the build seems to be succeeding on the auto-tester Windows hosts.

CC @RazvanN7

PetarKirov (approved):
Looks good.

andralex (approved):

WalterBright:
In the future, please use more illuminating branch names than `pull-20170618-002833` which gives no clue what this might be about. Something like `path-of-DMD`.

CyberShadow:
Sorry, that's part of an automated part of my workflow, and is not going to change. Please look at the commits in the branch or the PR title or description.

PR #6919, opened 2017-06-19
Fix Issue 17522 - win64.mak broken

CyberShadow:
CC @RazvanN7

This is still not enough to build Druntime/Phobos - another regression is blocking that, which I will file shortly.

### DIFF
--- a/src/win64.mak
+++ b/src/win64.mak
@@ -7,6 +7,8 @@
 MAKE=make
 HOST_DC=dmd
 MODEL=64
+BUILD=release
+OS=windows
 
 ################################### Rules ####################################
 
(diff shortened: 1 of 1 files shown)


PetarKirov (approved):

WalterBright:
This adds two more variables to the makefile with zero documentation about what they are for and what their expected values should be.

We need to raise the bar on this.

CyberShadow:
> This adds two more variables to the makefile with zero documentation about what they are for and what their expected values should be.

The variables are the same as in win32.mak.

> We need to raise the bar on this.

Yes, by *getting rid* on win32.mak and *especially* win64.mak.

PR #6921, opened 2017-06-20
CodeBuilder: more prolog code

WalterBright:


CyberShadow:
@wilzbach Maybe let's use a slightly different message for, err, tenured contributors or BDFLs :) This looks a little funny from the outside.

wilzbach:
> @wilzbach Maybe let's use a slightly different message for, err, tenured contributors or BDFLs :) This looks a little funny from the outside.

Fair point -> PR: https://github.com/dlang-bots/dlang-bot/pull/87

yebblies:
How far through would you estimate you are?

WalterBright:
I would guess 90%. The goal is to remove all dependency on it being a linked list, so I can replace it with an array. The array should be considerably faster. It's already faster because the CodeBuilder reduces the constant looking for the end of the list, much like getting rid of C's problem with constantly looking for the terminating 0 in strings.

PR #6922, opened 2017-06-21
skip typeinfo for struct/enum in betterC mode

adamdruppe:
Currently, if you try to compile a -betterC thing with a struct, it gives a linker error. This skips typeinfo letting them work in the simple case.

Whereas -betterC is a "rip stuff out" hack in the first place, this little change fits right in with it. And with this, I consider -betterC to actually live up to its name - the C subset of D is actually usable now, out of the box, given the switch.

### DIFF
--- a/src/ddmd/toobj.d
+++ b/src/ddmd/toobj.d
@@ -831,7 +831,8 @@ void toObjFile(Dsymbol ds, bool multiobj)
                 if (global.params.symdebug)
                     toDebug(sd);
 
-                genTypeInfo(sd.type, null);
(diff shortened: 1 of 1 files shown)


adamdruppe:
BTW I'd also like to make it skip the opEquals generation, but I don't know where that is in the compiler and need to get back to work on my real job. Skipping that should enable slices inside structs to work too. And if someone actually does `arr == arr2`, then... well, maybe generate it but I don't mind just giving them the ugly linker error. You can't do that in C anyway, and if you really want to, just define an operator overload.

PR #6923, opened 2017-06-21
Make struct dtors work in -betterC via hack on Linux

adamdruppe:
They currently fail because they output a reference to druntime personality function. This hacks it out. Follow-up to my other PR: https://github.com/dlang/dmd/pull/6922

These together with Walter's at https://github.com/dlang/dmd/pull/6918 achieve the goal I set out in October to make -betterC actually usable.

Still a filthy hack, especially this one, but a *usable* filthy hack. We can go back and add stuff back in like EH later.

Note: I have not tested betterC on Windows yet. I *think* the struct dtor already works there though.

### DIFF
--- a/src/ddmd/backend/dwarf.c
+++ b/src/ddmd/backend/dwarf.c
@@ -1503,7 +1503,9 @@ void dwarf_func_term(Symbol *sfunc)
 {
    //printf("dwarf_func_term(sfunc = '%s')\n", sfunc->Sident);
 
-    if (config.ehmethod == EH_DWARF)
+    // Temporarily disabling EH generation for betterC so scope(exit) and
+    // struct dtors work. This is a filthy hack.
+    if (config.ehmethod == EH_DWARF && !global.params.betterC)
     {
         bool ehunwind = doUnwindEhFrame();
 
--- a/test/runnable/cdtor.d
+++ b/test/runnable/cdtor.d
@@ -1,0 +1,27 @@
+/* REQUIRED_ARGS: -betterC
+   PERMUTE_ARGS:
+   DISABLED: win32
+ */
+
+__gshared bool dtorRan;
+
+struct S
+{
+    ~this()
+    {
+        dtorRan = true;
+    }
+
+    void m() {}
+}
+
+extern(C) int main()
+{
+    dtorRan = false;
+    {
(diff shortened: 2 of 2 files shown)


WalterBright:
> I have not tested betterC on Windows yet. I think the struct dtor already works there though.

Sounds like a test case should be added to this PR!

CyberShadow:
@WalterBright Done

adamdruppe:
And the tests are all passing. It doesn't quite work on Windows. I figure we can change it to use C++ library EH there, but the betterC flag wasn't available in nteh.c and #including it led to other errors! So I labeled this as being just Linux as a small step forward, better than nothing.

BTW on using C/C++ eh libs, I'm for that in general, though part of betterC's appeal to me is being able to pull out the C library too, so I say we want to keep that doable. not necessarily easy, but not harder than it needs to be either. Of course, that's a separate issue to this PR as it sits now which just hacks out the ref so we can use it today and worry about this later when we have a bit more real world experience.

WalterBright:
Sorry, I haven't had a chance to look at this yet.

MartinNowak:
In general we shouldn't merge behavior that only works on some platforms.

adamdruppe:
betterC is itself a hack that only works on some platforms.

ttldtor:
Does it help to solve this problem?

  * This code
```d
extern (C):
int main() {
    scope (exit) auto x = 1;
    return 0;
}
```
produces ``(.data.DW.ref.__dmd_personality_v0+0x0): undefined reference to `__dmd_personality_v0'`` error.

* This code
```d
extern (C):
void main() {
    scope (exit) auto x = 1;
}
```
linked without errors

* This code
```d
extern (C):
int main() {
    { scope (exit) auto x = 1; }
    return 0;
}
```
\- no errors

* And this
```d
extern (C):
int foo() {
    scope (exit) auto x = 1;
    return 0;
}

void main() {
}
```
produces the same error:
``(.data.DW.ref.__dmd_personality_v0+0x0): undefined reference to `__dmd_personality_v0'``

WalterBright:
Superceded by https://github.com/dlang/dmd/pull/7305 and https://github.com/dlang/dmd/pull/7304

PR #6924, opened 2017-06-22
Reset pending changelog after 2.075.0

wilzbach:
We should probably include this into the release pipeline ...
CC @MartinNowak


MartinNowak:
Same as for https://github.com/dlang/druntime/pull/1849

PR #6925, opened 2017-06-22
deprecate hex string literals

ghost:
The library version was merged 2 years ago so it can be qualified as bug-free, allowing to remove the built-in version.

### DIFF
--- a/changelog/deprecateHexStrings.dd
+++ b/changelog/deprecateHexStrings.dd
@@ -1,0 +1,3 @@
+HexString literals are deprecated.
+
+$(LINK2 http://dlang.org/spec/lex.html#HexString, HexString literals) are deprecated. Instead, the $(LINK2 https://dlang.org/phobos/std_conv.html#hexString, hexString) template must be used.
--- a/test/fail_compilation/lexer2.d
+++ b/test/fail_compilation/lexer2.d
@@ -1,11 +1,13 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/lexer2.d(14): Error: odd number (3) of hex characters in hex string
-fail_compilation/lexer2.d(15): Error: non-hex character 'G' in hex string
-fail_compilation/lexer2.d(16): Error: heredoc rest of line should be blank
-fail_compilation/lexer2.d(18): Error: unterminated delimited string constant starting at fail_compilation/lexer2.d(18)
-fail_compilation/lexer2.d(20): Error: semicolon expected following auto declaration, not `EOF`
+fail_compilation/lexer2.d(16): Error: odd number (3) of hex characters in hex string
+fail_compilation/lexer2.d(16): Deprecation: built-in hex string literals are deprecated, use `std.conv.hexString` instead
+fail_compilation/lexer2.d(17): Error: non-hex character 'G' in hex string
+fail_compilation/lexer2.d(17): Deprecation: built-in hex string literals are deprecated, use `std.conv.hexString` instead
+fail_compilation/lexer2.d(18): Error: heredoc rest of line should be blank
+fail_compilation/lexer2.d(20): Error: unterminated delimited string constant starting at fail_compilation/lexer2.d(20)
+fail_compilation/lexer2.d(22): Error: semicolon expected following auto declaration, not `EOF`
 ---
 */
 
--- a/src/ddmd/lexer.d
+++ b/src/ddmd/lexer.d
@@ -322,6 +322,7 @@ class Lexer
                     goto case_ident;
                 p++;
                 t.value = hexStringConstant(t);
+                deprecation("built-in hex string literals are deprecated, use `std.conv.hexString` instead");
                 return;
             case 'q':
                 if (p[1] == '"')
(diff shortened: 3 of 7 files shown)


changelog/deprecateHexStrings.dd
```diff
@@ -0,0 +1,3 @@
+HexString literals are deprecated.
+
+HexString literals are deprecated. Instead, the $(LINK2 https://dlang.org/phobos/std_conv.html#hexString hexString) template must be used.
```
wilzbach: Nit: You could link to http://dlang.org/spec/lex.html#HexString

wilzbach (commented):
If this gets accepted, we shouldn't forget to update the [deprecation page](https://dlang.org/deprecate.html)

BBasile:
Actually i see a slight problem caused by the removal of built-in hex strings: people who compile betterC D code will be, in a way, discriminated.

test/fail_compilation/lexer2.d
```diff
@@ -1,18 +1,14 @@
...
+fail_compilation/lexer2.d(12): Error: heredoc rest of line should be blank
+fail_compilation/lexer2.d(14): Error: unterminated delimited string constant starting at fail_compilation/lexer2.d(14)
+fail_compilation/lexer2.d(16): Error: semicolon expected following auto declaration, not 'EOF'
 ---
 */
 
 // https://dlang.dawg.eu/coverage/src/lexer.c.gcov.html
 
-static s1 = x"123";
-static s2 = x"123G";
```
WalterBright: These should still be tested, even while deprecated, as they are still supported.

JinShil:
> Actually i see a slight problem caused by the removal of built-in hex strings: people who compile betterC D code will be, in a way, discriminated.

Those who compile with -betterC are already discriminated against in much more serious ways.  Don't worry about it.  It's a tradeoff that they voluntarily opt into.

wilzbach:
@WalterBright - I think this needs your approval. Could you please quickly confirm that you are ok with deprecating hex strings (and I guess in the next step octal strings) so that @BBasile can move on here and get this through the auto-tester?

BBasile:
I've detected a problem while removing the uses in phobos:

```d
void main()
{
    import std.conv: hexString;
    enum a1 = cast(ubyte[4]) x"40 41 42 43"; // compiles
    enum a2 = cast(ubyte[4]) hexString!"40 41 42 43"; // fails
}
```

both versions are not perfectly interchangeable.

MartinNowak:
```d
enum ubyte[4] a2 = hexString!"40 41 42 43"; // works
```
Don't ask me why though :/.

BBasile:
I'd like to put the focus on this PR: https://github.com/dlang/phobos/pull/5546.
You can see the consequences here. I think that it was not a good idea. Since the compiler and the standard library are decoupled the consequence is that the tests are not run **at all** when one makes a PR that proposes a deprecation.

You can also see that this affects a simple PR from a very casual committer. Is that just a random fact ? I don't think so.

wilzbach:
> You can see the consequences here. I think that it was not a good idea. Since the compiler and the standard library are decoupled

They have always been coupled (de facto). After all, that's why the auto-tester tests all three repos.

> the consequence is that the tests are not run at all when one makes a PR that proposes a deprecation.

I am not sure I get your point, the deprecation messages are shown:

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=2651217&isPull=true

image

Are you complaining about the fact that CircleCi fails and as a result doesn't report coverage? Or that most CIs are red?

> You can also see that this affects a simple PR from a very casual committer. 

It doesn't. This PR is blocked on @WalterBright's approval for changing the D language, when/if there's the approval, we can go ahead and merge your Phobos PR.

BBasile:
No. The test are failing since yesterday because you've decided (I mean you and the other member who has approved) that the deprecations are now errors instead of warnings in phobos. Cant you see the problem ? If in the future if something has to be deprecated the same will happen again.

CyberShadow:
> I am not sure I get your point,

@BBasile is complaining that now that Phobos is failing the *build*, the autotester never gets around to running the DMD test suite.

TBH I don't see much of a difference - you *do* want to have all uses of deprecated features removed from Phobos before the compiler change is merged, as it not only ensures there are no embarrassments about us ourselves still using features we don't want others to use, but also ensures that we have suitable replacements that are at least as good for Phobos. Whether to deprecate something also depends more on whether the idea is sound (for which being able to adequately fix the deprecation instances in Phobos makes a good litmus test), and secondly on the quality of the implementation (which the DMD test suite checks).

Furthermore, you wrote about "platform specific failure missed locally", but deprecations by their nature are going to be a front end change, so there is not going to be much platform-dependent about it.

I guess it would be possible in theory to check for deprecations in the Phobos tests instead of build, but I don't think the make file complications are worth it. Get your Phobos PR merged (https://github.com/dlang/phobos/pull/5507), and then you will see DMD test results again.

BBasile:
The difference is so obvious. From everything green yesterday to everything red today. Your new policy doesn't allow to test and integrate correctly new deprecations at the language level.

CyberShadow:
> From everything green yesterday to everything red today.

As it should be. We should not be using features that we have deprecated in any components of D.

Another point I forgot to mention above in that occasionally through negligence, we had shipped a version of D which caused deprecation warnings in Phobos when it is used as a library, i.e. someone imports a non-deprecated module and uses a non-deprecated symbol, and they get deprecation errors somewhere inside Phobos source code, making their program unbuildable with `-de`. This is a situation we definitely want to avoid.

That this PR was showing as green in the autotester was a problem, which has now been fixed. The correct process is to first fix uses of deprecated symbols in Phobos/Druntime/etc., then have the deprecation merged.

BBasile:
So i let the initiative to someone who'll be stasified by the way things are working here

PR #6927, opened 2017-06-23
betterC: send array overflows to C assert

WalterBright:
So that `betterC` mode can support runtime array overflow detection.

### DIFF
--- a/src/ddmd/e2ir.d
+++ b/src/ddmd/e2ir.d
@@ -1962,7 +1962,11 @@ elem *toElem(Expression e, IRState *irs)
             {
                 if (global.params.betterC)
                 {
-                    result = callCAssert(irs, ae);
+                    auto econd = toElem(ae.e1, irs);
+                    auto ea = callCAssert(irs, ae.e1.loc, ae.e1, ae.msg, null);
+                    auto eo = el_bin(OPoror, TYvoid, econd, ea);
+                    elem_setLoc(eo, ae.loc);
+                    result = eo;
                     return;
                 }
 
@@ -5936,6 +5940,10 @@ elem *filelinefunction(IRState *irs, Loc *loc)
  */
 elem *buildArrayBoundsError(IRState *irs, const ref Loc loc)
 {
+    if (global.params.betterC)
+    {
+        return callCAssert(irs, loc, null, null, "array overflow");
+    }
     auto eassert = el_var(getRtlsym(RTLSYM_DARRAYP));
     auto efile = toEfilenamePtr(cast(Module)irs.blx._module);
     auto eline = el_long(TYint, loc.linnum);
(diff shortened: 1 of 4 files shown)


wilzbach:
Hmm @CyberShadow - should I just deploy https://github.com/dlang-bots/dlang-bot/pull/87? I am pretty sure Martin will be okay and otherwise we can always revert it - it's not like there's anything depending on the bot's codebase.

CyberShadow:
Up to you. I don't think it's urgent unless someone finds it annoying.

src/ddmd/e2ir.d
```diff
@@ -5936,6 +5940,10 @@ elem *filelinefunction(IRState *irs, Loc *loc)
  */
 elem *buildArrayBoundsError(IRState *irs, const ref Loc loc)
 {
+    if (global.params.betterC)
+    {
+        return callCAssert(irs, loc, null, null, "array overflow");
```
MetaLang: Is it possible here to include the index that caused the asset?

WalterBright: Yes but right now I just want it to work.

WalterBright:
Added switch errors, too.

src/ddmd/e2ir.d
```diff
@@ -6005,33 +6013,35 @@ void toTraceGC(IRState *irs, elem *e, Loc *loc)
...
  */
-
-elem *callCAssert(IRState *irs, AssertExp ae)
+elem *callCAssert(IRState *irs, Loc loc, Expression exp, Expression emsg, const(char)* str)
 {
-    //printf("callCAssert.toElem() %s\n", ae.toChars());
-    elem *econd = toElem(ae.e1, irs);
-
+    //printf("callCAssert.toElem() %s\n", e.toChars());
     Module m = cast(Module)irs.blx._module;
```
andralex: auto would be better but ok to keep existing style

andralex (approved):
Great, so now we have complete C-style asserts everywhere with -betterC?

WalterBright:
> complete C-style asserts everywhere with -betterC?

Yes

test/compilable/betterCarray.d
```diff
@@ -0,0 +1,17 @@
...
+extern (C) int main(char** argv, int argc) {
+    printf("hello world\n");
+    int[3] a;
+    foo(a[], 3);
+    return 0;
+}
+
+int foo(int[] a, int i)
+{
+    return a[i];
```
MartinNowak: Not testing much here. We're not even linking the compilable tests.
You should at least run this and grep in the output, at least on some platforms.

MartinNowak:
Sorry, there are literally no active tests for this feature, it shouldn't have been merged.

PR #6928, opened 2017-06-23
Fix Issue 16640 - __FILE_FULL_PATH__ doesn't return relative path when used as default parameter

RazvanN7:
The resolveLoc function of the FileInitExp class, which is responsible for assigning the right path to the symbol used treats only the case of  __File__. I added an if to test if the symbol used is __FILE_FULL_PATH__ so that the right path is assigned

### DIFF
--- a/test/runnable/test16640.d
+++ b/test/runnable/test16640.d
@@ -1,0 +1,11 @@
+// PERMUTE_ARGS:
+
+void testFileFullPathAsDefaultArgument(string preBakedFileFullPath, string fileFullPath = __FILE_FULL_PATH__)
+{
+    assert(preBakedFileFullPath == fileFullPath);
+}
+
+void main()
+{
+    testFileFullPathAsDefaultArgument(__FILE_FULL_PATH__);
+}
--- a/src/ddmd/expression.d
+++ b/src/ddmd/expression.d
@@ -16065,6 +16065,8 @@ extern (C++) final class FileInitExp : DefaultInitExp
     {
         //printf("FileInitExp::resolve() %s\n", toChars());
         const(char)* s = loc.filename ? loc.filename : sc._module.ident.toChars();
+        if (subop == TOKfilefullpath)
+            s = FileName.combine(sc._module.srcfilePath, s);
         Expression e = new StringExp(loc, cast(char*)s);
         e = e.semantic(sc);
         e = e.castTo(sc, type);


RazvanN7:
@CyberShadow any ideas why jenkins is failing on some projects with the message "Memory allocation failed"?

CyberShadow:
Err, no. @wilzbach @MartinNowak might know, though.

wilzbach:
Not really, I thought that Jenkins checks out the latest tag, but maybe it's due to https://github.com/dlang/dub/pull/1166?

> Regression fixes should always target stable

Hehe, great to see the new warnings working :) :)

marler8997 (changes requested):
EDIT: Ignore this comment, I made a mistake when looking at the change.

In expression.d line 16068, might be better initialize `s` with an if/else instead of default/if
```D
const (char)* s.
if (subop == TOKfilefullpath)
    s = ...
else
   s = ...
```

test/runnable/test16640.d
```diff
@@ -0,0 +1,11 @@
+// PERMUTE_ARGS:
+
+void testFileFullPathAsDefaultArgument(string preBakedFileFullPath, string fileFullPath = __FILE_FULL_PATH__)
+{
+    assert(preBakedFileFullPath == fileFullPath);
+}
+
+void main()
+{
+    testFileFullPathAsDefaultArgument(__FILE_FULL_PATH__);
```
wilzbach: Hmm imho you should inject the "real" path from the outside call `testFileFullPathAsDefaultArgument` as well - otherwise this test might be completely worthless if __FILE_FULL_PATH__ is broken.

ZombineDev:
Closing & reopening this PR or rebasing it should fix the Jenkins failure.

test/runnable/test16640.d
RazvanN7: There already is a test in test/compilable/line.d that would fail if that would be the case.

RazvanN7:
@ZombineDev Rebasing solved the problem. Thanks!

RazvanN7:
@ZombineDev still, vibe.d seems to be failing for unknown reasons.

ZombineDev:
The errors from the [previous build](https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-6928/3/pipeline/) are [now fixed](https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-6928/4/pipeline/), but yes there's now the vibe.d failure. I've also seen it before on other PRs, so I think it is also transient.

test/runnable/test16640.d
marler8997: The test in line.d checks if `__FILE_FULL_PATH__` ends with the same expansion as `__FILE__` and that it's an absolute path, but doesn't actually check whether the full path is correct. @wilzbach 's idea of having a test pass in the actual full path and comparing it with `__FILE_FULL_PATH__` would be a nice addition, not sure how it would be implemented though.  I wouldn't block this PR without it, but this PR would be even better with a new test that does this.

P.S. I made a mistake in my previous comment about initialization, feel free to ignore it :)

andralex (approved):

wilzbach:
> Regression fixes should always target stable

I wonder what the reason for ignoring this is. The bot now even shows this as a warning..
Is it not clear that this applies for critical/blocking bugs as well?
Or is it simply to hard to rebase onto stable?

CyberShadow:
> Regression fixes should always target stable

It's not really visible, right now it's listed as an aside bullet point. Maybe it should pop out more? Like:

⚠️⚠️⚠️ **WARNING** ⚠️⚠️⚠️ - Regression fixes should always target stable. [Learn more](#)

Also, the warning is gone now.

wilzbach:
> It's not really visible, right now it's listed as an aside bullet point. Maybe it should pop out more? Like:

Excellent idea -> https://github.com/dlang-bots/dlang-bot/pull/120

> Also, the warning is gone now.

Probably due to the merge -> https://github.com/dlang-bots/dlang-bot/issues/119

andralex:
Thanks guys. @RazvanN7 could you please follow up?

PR #6932, opened 2017-06-25
optimize (b ? c1 : c2) to (c1 + b *(c2 - c1))

WalterBright:
When `b` will evaluate to 0 or 1 and (c2-c1) is a power of 2.

### DIFF
--- a/src/ddmd/backend/cgelem.c
+++ b/src/ddmd/backend/cgelem.c
@@ -1900,20 +1900,65 @@ STATIC elem * elcond(elem *e, goal_t goal)
             elem *ec2 = e->E2->E2;
 
             if (tyintegral(ty) && ec1->Eoper == OPconst && ec2->Eoper == OPconst)
-            {   targ_llong i1,i2;
-                targ_llong b;
+            {
+                targ_llong i1 = el_tolong(ec1);
+                targ_llong i2 = el_tolong(ec2);
+                tym_t ty1 = tybasic(e1->Ety);
 
-                i1 = el_tolong(ec1);
-                i2 = el_tolong(ec2);
-
-                /* If b is an integer with only 1 bit set then          */
-                /*   replace ((a & b) ? b : 0) with (a & b)             */
-                /*   replace ((a & b) ? 0 : b) with ((a & b) ^ b)       */
-                if (e1->Eoper == OPand && e1->E2->Eoper == OPconst &&
+                if ((ty1 == TYbool && !OTlogical(e1->Eoper) || e1->Eoper == OPand && e1->E2->Eoper == OPconst) &&
                     tysize(ty) == tysize(ec1->Ety))
                 {
-                    b = el_tolong(e1->E2);
-                    if (ispow2(b) != -1)        /* if only 1 bit is set */
+                    targ_llong b = ty1 == TYbool ? 1 : el_tolong(e1->E2);
+
+                    if (b == 1 && ispow2(i1 - i2) != -1)
+                    {
+                        // replace (e1 ? i1 : i2) with (i1 + (e1 ^ 1) * (i2 - i1))
+                        // replace (e1 ? i2 : i1) with (i1 + e1 * (i2 - i1))
+                        int sz = tysize(e1->Ety);
+                        while (sz < tysize(ec1->Ety))
+                        {
+                            // Increase the size of e1 until it matches the size of ec1
+                            switch (sz)
+                            {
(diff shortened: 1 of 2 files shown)


src/ddmd/backend/cgelem.c
```diff
@@ -1900,20 +1900,65 @@ STATIC elem * elcond(elem *e, goal_t goal)
...
+                                    continue;
+                                default:
+                                    assert(0);
+                            }
+                            break;
+                        }
+                        if (i1 < i2)
+                        {
+                            ec2->EV.Vllong = i2 - i1;
+                            e1 = el_bin(OPxor,e1->Ety,e1,el_long(e1->Ety,1));
```
yebblies: This isn't covered by the comment above...

src/ddmd/backend/cgelem.c
```diff
@@ -1900,20 +1900,65 @@ STATIC elem * elcond(elem *e, goal_t goal)
...
-                /*   replace ((a & b) ? 0 : b) with ((a & b) ^ b)       */
-                if (e1->Eoper == OPand && e1->E2->Eoper == OPconst &&
+                if ((ty1 == TYbool && !OTlogical(e1->Eoper) || e1->Eoper == OPand && e1->E2->Eoper == OPconst) &&
                     tysize(ty) == tysize(ec1->Ety))
                 {
-                    b = el_tolong(e1->E2);
-                    if (ispow2(b) != -1)        /* if only 1 bit is set */
+                    targ_llong b = ty1 == TYbool ? 1 : el_tolong(e1->E2);
+
+                    if (b == 1 && ispow2(i1 - i2))
```
yebblies: This requirement needs to be in the comment too

src/ddmd/backend/cgelem.c
```diff
@@ -1900,20 +1900,65 @@ STATIC elem * elcond(elem *e, goal_t goal)
...
+                                    sz = 4;
+                                    continue;
+                                case 4:
+                                    e1 = el_una(OPu32_64, TYullong, e1);
+                                    sz = 8;
+                                    continue;
+                                default:
+                                    assert(0);
+                            }
+                            break;
```
yebblies: Unreachable?

src/ddmd/backend/cgelem.c
```diff
@@ -1953,14 +1998,14 @@ STATIC elem * elcond(elem *e, goal_t goal)
                 // code for without using jumps.
 
                 // Try to replace (!e1) with (e1 < 1)
-                else if (e1->Eoper == OPnot && !OTrel(e1->E1->Eoper))
+                else if (e1->Eoper == OPnot && !OTrel(e1->E1->Eoper) && e1->E1->Eoper != OPand)
```
yebblies: What is this for?

yebblies (commented):
Also, why does this only apply when a - b is a power of two?  Isn't the transform valid if they're not?

src/ddmd/backend/cgelem.c
yebblies: Also, why is this checking `ispow2` is non-zero, while below it checks `ispow2(b) != -1`?

WalterBright:
> why does this only apply when a - b is a power of two?

Because then it will be transformed later into a shift.

> Isn't the transform valid if they're not?

Yes, but the multiply is more expensive than the ?:, but the shift is cheaper.

src/ddmd/backend/cgelem.c
WalterBright: If it's an `and` operation, the optimization is a pessimization.

yebblies:
Then why not emit a shift directly?  The description of this optimization is misleading since you're not actually transforming to a multiply.

src/ddmd/backend/cgelem.c
WalterBright: fixed

src/ddmd/backend/cgelem.c
WalterBright: made better

src/ddmd/backend/cgelem.c
WalterBright: good catch

WalterBright:
> Then why not emit a shift directly?

Because this is slightly simpler and easier to understand. Quite a bit of the transformations done rely on further transformations.

> The description of this optimization is misleading since you're not actually transforming to a multiply.

Yes, you are, at this point in the optimization.

PR #6933, opened 2017-06-25
optimize ((i / C1) / C2) with (i / (C1 * C2))

WalterBright:
Getting rid of divide instructions is always good.

### DIFF
--- a/src/ddmd/backend/cgelem.c
+++ b/src/ddmd/backend/cgelem.c
@@ -2272,13 +2272,11 @@ STATIC elem * elmod(elem *e, goal_t goal)
  */
 
 STATIC elem * eldiv(elem *e, goal_t goal)
-{   elem *e2;
-    tym_t tym;
-    int uns;
-
-    e2 = e->E2;
-    tym = e->E1->Ety;
-    uns = tyuns(tym) | tyuns(e2->Ety);
+{
+    //printf("eldiv()\n");
+    elem *e2 = e->E2;
+    tym_t tym = e->E1->Ety;
+    int uns = tyuns(tym) | tyuns(e2->Ety);
     if (cnst(e2))
     {
 #if 0 && MARS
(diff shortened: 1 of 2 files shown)


src/ddmd/backend/cgelem.c
```diff
@@ -2280,6 +2278,70 @@ STATIC elem * eldiv(elem *e, goal_t goal)
...
+                }
+            }
+        }
+
+        if (tyintegral(tym) && e->Eoper == OPdiv && e2->Eoper == OPconst &&
+            e1->Eoper == OP64_32 &&
+            e1->E1->Eoper == OPremquo && e1->E1->E2->Eoper == OPconst)
+        {
+            /* Replace:
+             *   (64_32 (e /% c1)) / c2
```
yebblies: Shouldn't `64_32(e /% c)` already be turned into `e / c`?  Is there any reason that wouldn't happen before this is called and be handled by the case above?

WalterBright: No. The `/` is transformed into the `/%`, because that way it can later combine `/` and `%` into one instruction. This bit here undoes that optimization.
A common problem in optimizations is one optimization undoing another. The rewrites are also done bottom up, hence the order the transformations are done.

UplinkCoder:
can't we do this higher up ?
in the AST ?

WalterBright:
> can't we do this higher up ?

The intermediate code is the right place to do it, it's where pretty much all optimizers do such things.
https://www.youtube.com/watch?v=UHv_Jog9Xuc

UplinkCoder:
> The intermediate code is the right place to do it, it's where pretty much all optimizers do such things.
Well yes, but usually such intermediate code is sill in the shared layer, whereas this is backend specific.
It is also usually visualize-able, whereas this IR has no such convenience. 
My concern the debug-abilty optimizers are notorious for introducing hard to catch bugs.

However since we are already doing optimization all over the backend I guess there is point changing horses now.

WalterBright:
> Well yes, but usually such intermediate code is in the shared layer, whereas this is backend specific.

It is in the shared layer - the optimizer pass. LLVM likely does the same thing in their shared layer. DMC++/DMD share the optimizer and codegen passes.

> It is also usually visualize-able, whereas this IR has no such convenience.

`elem_print(e)`

> My concern the debug-abilty optimizers are notorious for introducing hard to catch bugs.

Moving it elsewhere won't change that. Besides, the front end is far more complex, and is much harder to fix errors in. The whole point of doing transformations on the IR is it is much, much, **much** simpler.

> However since we are already doing optimization all over the backend I guess there is point changing horses now.

The backend is clearly divided between the target-independent optimization passes and the target-dependent code gen pass. This is normally how compilers are constructed, in fact, it's ubiquitous.

I recommend watching the linked video which goes into some detail on how this works.

UplinkCoder:
> elem_print(e)

Point taken.
If we can visualize the IR then I have no further complaints.
I must have overlooked this while sifting through the backend.

The backend ir did not strike me as particularly simple :)
But I agree that it is better then the full ast variety.

> I recommend watching the linked video which goes into some detail on how this works.

I did so a few years ago.

wilzbach:
> I did so a few years ago.

The linked video is from 2017 :wink:

UplinkCoder:
> The linked video is from 2017 :wink:

He must have given a very similar talk then at some other conf. maybe an llvm-dev meeting ?
I am very sure I heard this talk before.

src/ddmd/backend/cgelem.c
```diff
@@ -2280,6 +2278,70 @@ STATIC elem * eldiv(elem *e, goal_t goal)
 
     if (OPTIMIZER)
     {
+        elem *e1 = e->E1;
```
UplinkCoder: maybe name it E1 ?

WalterBright: locals should start with lower case. Besides, it is called `e1` everywhere else in the compiler.

src/ddmd/backend/cgelem.c
```diff
@@ -2280,6 +2278,70 @@ STATIC elem * eldiv(elem *e, goal_t goal)
...
+        {
+            /* Replace:
+             *   (e / c1) / c2
+             * With:
+             *   e / (c1 * c2)
+             */
+            targ_llong c1 = el_tolong(e1->E2);
+            targ_llong c2 = el_tolong(e2);
+            int uns1 = tyuns(e1->E1->Ety) || tyuns(e1->E2->Ety);
+            int uns2 = tyuns(e1->Ety) || tyuns(e2->Ety);
```
andralex: these really should be `bool`

src/ddmd/backend/cgelem.c
```diff
@@ -2280,6 +2278,70 @@ STATIC elem * eldiv(elem *e, goal_t goal)
...
+                         : (-0x1000000 < c1 && c1 < 0x1000000 && -0x1000000 < c2 && c2 < 0x1000000))
+                {
+                    e->E1 = e1->E1;
+                    e1->E1 = e1->E2;
+                    e1->E2 = e2;
+                    e->E2 = e1;
+                    e1->Eoper = OPmul;
+                    return optelem(e, GOALvalue);
+                }
+            }
```
andralex: If `c1*c2` overflows just generate result `0` without carrying any division. If `c1 * c2` would overflow, then dividing by both always results in zero.

To \check for overflow use `muls` and `mulu`.

src/ddmd/backend/cgelem.c
```diff
@@ -2280,6 +2278,70 @@ STATIC elem * eldiv(elem *e, goal_t goal)
...
+             */
+            elem *erq = e1->E1;
+            targ_llong c1 = el_tolong(erq->E2);
+            targ_llong c2 = el_tolong(e2);
+            int uns1 = tyuns(erq->E1->Ety) || tyuns(erq->E2->Ety);
+            int uns2 = tyuns(e1->Ety) || tyuns(e2->Ety);
+            if (uns1 == uns2)   // identity doesn't hold for mixed sign case
+            {
+                // The transformation will fail if c1*c2 overflows. This substitutes
+                // for a proper overflow check.
```
andralex: again it's easy to do the proper overflow check with `muls` and `mulu` in druntime.

src/ddmd/backend/cgelem.c
WalterBright: `muls` and `mulu` are not available to the back end. It's in C++.

src/ddmd/backend/cgelem.c
WalterBright: If only the backend were written in D :-)

My initiative on that is: https://github.com/dlang/dmd/pull/6907

src/ddmd/backend/cgelem.c
andralex: @WalterBright oh ok. A simple possibility is then to inline what they do. Alternatively, for unsigned the msb of the result is msb(lhs) + msb(rhs).

src/ddmd/backend/cgelem.c
WalterBright: done

src/ddmd/backend/cgelem.c
WalterBright: I figured on just deferring it until it gets converted to D. I doubt the high number cases are likely to present themselves in real code - just the low number cases.

The trouble with writing an overflow detection routine in C is that C compilers notoriously optimize them away. There isn't a straightforward way of dealing with that. I'd prefer to not deal with that problem - it's unpredictable behavior. We _can_ control it in D.

src/ddmd/backend/cgelem.c
yebblies: So you rely on CSE for merging them instead of having a special rule for merging `/` and `%`?  That's an interesting trick, I hadn't thought of that.

src/ddmd/backend/cgelem.c
```diff
@@ -2280,6 +2278,70 @@ STATIC elem * eldiv(elem *e, goal_t goal)
...
+             */
+            targ_llong c1 = el_tolong(e1->E2);
+            targ_llong c2 = el_tolong(e2);
+            bool uns1 = tyuns(e1->E1->Ety) || tyuns(e1->E2->Ety);
+            bool uns2 = tyuns(e1->Ety) || tyuns(e2->Ety);
+            if (uns1 == uns2)   // identity doesn't hold for mixed sign case
+            {
+                // The transformation will fail if c1*c2 overflows. This substitutes
+                // for a proper overflow check.
+                if (uns1 ? (c1 < 0x10000000 && c2 < 0x10000000)
```
yebblies: What am I missing with these numbers?  `((2^24)-1)^2` will overflow for 32 bits, but won't come close for 64.  What size can we assume here and why the seemingly arbitrary limits?  Should it be `2^(REGSIZE*8/2)-1`?

src/ddmd/backend/cgelem.c
WalterBright: Yes. It does the same thing for sincos. It took me a long time to come up with that trick :-)

src/ddmd/backend/cgelem.c
WalterBright: I thought I'd leave proper overflow checking for the D conversion. In the meantime, I'd stick with the values being less than sqrt(int.max).

yebblies: That's not `sqrt(int.max)` - did you mean `0x10000`?

WalterBright: Sorry, fixed.

PR #6038, opened 2016-08-08
[Trivial] Fix C++ header after PR #6001.

JohanEngelen:
This `module.h` header change should have been part of https://github.com/dlang/dmd/pull/6001.

Ping @yebblies

PR #6001, opened 2016-08-02
Make Module.semantic override the base class's virtual function

JohanEngelen:
The current Module interface defines `semantic()` which is _not_ overriding the Dsymbol baseclass's `semantic(Scope*)`. This PR fixes that so that the interface is more clear.
Compare with Module.importAll.

I chose to not name the `Scope*` function argument to indicate that it is not used.

PR #6934, opened 2017-06-25
Fix Issue 17548 - [REG2.072.0] Forward reference error with scope function parameters

JohanEngelen:
This PR
- reverts #6001 and #6038 , because #6001  caused a regression : https://issues.dlang.org/show_bug.cgi?id=17548 .
- adds the testcase from dlang issue 17548
- renames Module.semantic*, such that it is clear that it is not accidental that Dsymbol.semantic* are not overridden (this was the original intent of #6001).

### DIFF
--- a/src/ddmd/dimport.d
+++ b/src/ddmd/dimport.d
@@ -292,7 +292,7 @@ extern (C++) final class Import : Dsymbol
                 scopesym.addAccessiblePackage(mod, protection); // d
             }
 
-            mod.semantic(null);
+            mod.semantic();
             if (mod.needmoduleinfo)
             {
                 //printf("module4 %s because of %s\n", sc.module.toChars(), mod.toChars());
@@ -409,7 +409,7 @@ extern (C++) final class Import : Dsymbol
         //printf("Import::semantic2('%s')\n", toChars());
         if (mod)
         {
-            mod.semantic2(null);
+            mod.semantic2();
             if (mod.needmoduleinfo)
             {
                 //printf("module5 %s because of %s\n", sc.module.toChars(), mod.toChars());
@@ -476,7 +476,7 @@ extern (C++) final class Import : Dsymbol
         {
             load(null);
             mod.importAll(null);
-            mod.semantic(null);
+            mod.semantic();
         }
         // Forward it to the package/module
         return pkg.search(loc, ident, flags);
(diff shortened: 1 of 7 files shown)


WalterBright:
> renames Module.semantic*, such that it is clear that it is not accidental that Dsymbol.semantic* are not overridden (this was the original intent of #6001).

What's the problem with overriding it? It looks strange to have two different semantic2() functions.

JohanEngelen:
> What's the problem with overriding it?  It looks strange to have two different semantic2() functions.

Indeed. Before #6001, there were two different (but identically named) semantic* functions for a Module. #6001 tried to resolve it by making the Module.semantic* override Dsymbol's. But that apparently broke things. So I reverted that. But now with the revert, a different test (`compilable/ice10598.d`) breaks... 
The frontend code is so opaque that I cannot make any judgement on what the correct code should be.

Syniurge:
Thanks for looking into it Johan. This fixes my particular case, but `Module.doSemanticPass()` can still be hit at the wrong time with the same outcome if we add an import in S2:
```d
struct S2 {
    // void bar(int arg = .fwdref1.cnst) {}
    S1 s;
    import fwdref2;
}
``` 
Importing itself is contrived, but it could be importing a fwdref3 module that imports fwdref2 and it would be the same.

JohanEngelen:
Note: I just wanted to unbreak things that got broken by my PR. Now it looks like there is a bigger problem lurking, which does need a proper fix, rather than a revert of #6001. :(
Because I am currently not actively working on figuring out what's wrong (and probably won't be any time soon), shall I close this PR?

Syniurge:
Still helped :)

>Now it looks like there is a bigger problem lurking, which does need a proper fix 

Always that feeling with that kind of issue. 

I pushed a tentative one line fix but fearing it may miss the bigger picture I wanted to see other analysis/suggestions.

MartinNowak:
Any idea what caused the regression? Looks a bit like `Module.semantic(Scope *)` now gets used as `Dsymbol.semantic(sc)` somewhere, because of the apparent reentrant semantic calls. Before #6001 this just resolved to [`Package.semantic(Scope*)`](https://github.com/dlang/dmd/blob/4300bb98fc0ed35fb51ea42417d6723e000ad36f/src/ddmd/dmodule.d#L249) which doesn't do much.
We might want to add an early return to reentrant `Module.semantic*` calls, e.g. [here](https://github.com/dlang/dmd/blob/4300bb98fc0ed35fb51ea42417d6723e000ad36f/src/ddmd/dmodule.d#L1052).

PR #6935, opened 2017-06-25
use git describe for development versions

MartinNowak:
- VERSION will get bumped with a release, not directly after a release, e.g.
  v2.075.0-8aeca3c (somewhat before v2.075.0) =>
  v2.074.1-456-g8aeca3c (456 commits after v2.074.1)
- VERSION file will get updated in stable and gets merged back into master
- ensures v2.074.1-ddd-hhhhh contains all changes from v2.074.1

As discussed here (http://forum.dlang.org/post/dlbbqegxkblvhakfsogz@forum.dlang.org) getting rid of an error-prone (and home-brewed) development version scheme.


src/posix.mak
```diff
@@ -466,25 +466,20 @@ $G/idgen: $D/idgen.d $(HOST_DMD_PATH)
...
-VERSION := $(shell cat ../VERSION)
-ifneq (1,$(RELEASE))
-VERSION_GIT := $(shell printf "`$(GIT) rev-parse --short HEAD`"; \
-       test -n "`$(GIT) status --porcelain -uno`" && printf -- -dirty)
-VERSION := $(addsuffix -devel$(if $(VERSION_GIT),-$(VERSION_GIT)),$(VERSION))
+######## VERSION
+
+VERSION := $(shell cat ../VERSION) # default to checked-in VERSION file
+ifneq (1,$(RELEASE)) # unless building a release
+	VERSION_GIT := $(shell printf "`$(GIT) describe --dirty`") # use git describe
```
MartinNowak: The `--dirty` option is supported since 2009.
https://github.com/git/git/commit/9f67d2e8279e1885ef2b4681c19cef8534259783

src/posix.mak
```diff
@@ -466,25 +466,20 @@ $G/idgen: $D/idgen.d $(HOST_DMD_PATH)
...
-# into account, untracked files don't affect the dirty state).
-VERSION := $(shell cat ../VERSION)
-ifneq (1,$(RELEASE))
-VERSION_GIT := $(shell printf "`$(GIT) rev-parse --short HEAD`"; \
-       test -n "`$(GIT) status --porcelain -uno`" && printf -- -dirty)
-VERSION := $(addsuffix -devel$(if $(VERSION_GIT),-$(VERSION_GIT)),$(VERSION))
+######## VERSION
+
+VERSION := $(shell cat ../VERSION) # default to checked-in VERSION file
+ifneq (1,$(RELEASE)) # unless building a release
```
MikeWey: Doesn't this line do the exact opposite of what the comment says? 
In other places in the makefile `ifneq (,$(RELEASE))` is used to check if this is a release is build.

Wouldn't it make sense to always try `git describe` to get the version number? It only returns the tag if the current HEAD is tagged.

MartinNowak: It's `ifneq (1,$(RELEASE))` and I don't want to remove with the ability to override `VERSION` for releases.

MikeWey: Ok, but going on the other use of RELEASE in the makefile it is 1 when building a release.
https://github.com/MartinNowak/dmd/blob/c28e2d8cfebfcc5b2fb8ee4b35bfb661314bc042/src/posix.mak#L178

So `ifneq (1,$(RELEASE))` would indicate "Do this when not building a release", which doesn't match the comment. This would mean either the code or the comment is wrong.

src/posix.mak
```diff
@@ -466,25 +466,20 @@ $G/idgen: $D/idgen.d $(HOST_DMD_PATH)
...
+	VERSION_GIT := $(shell printf "`$(GIT) describe --dirty`") # use git describe
+	ifneq (,$(VERSION_GIT)) # check for git failures
```
wilzbach: The CircleCi failure for Phobos stable branches are due to `shell` errors leading to direct abortion of the build process.
See also: https://www.gnu.org/software/make/manual/html_node/Errors.html

> After each shell invocation returns, make looks at its exit status. If the shell completed successfully (the exit status is zero), the next line in the recipe is executed in a new shell; after the last line is finished, the rule is finished.
> If there is an error (the exit status is nonzero), make __gives up__ on the current rule, and perhaps on __all rules__.

It looks like this applies to `shell` invocations as well.

src/posix.mak
MartinNowak: The comment says `unless building a release` == `if not building a release`, no problem here.

src/posix.mak
MartinNowak: Just try it, e.g. changing describe to describes, it continues to run.
Guess this only applies to shell invocations in make rules, but not top-level `$(shell)` calls.

wilzbach: > Just try it, e.g. changing describe to describes, it continues to run.

Did you nuke `generated` before trying this?
It doesn't work for me, because generating `VERSION` fails ...

PR #6936, opened 2017-06-25
fix issue 17551 - VC build: bad mangling of real template argument

rainers:
fix ld_sprint not reflecting case of format character


rainers:
no test case as this code is not built by any test server anyway.

PR #6937, opened 2017-06-25
Alternative fix to issue 17548 - [REG2.072.0] Forward reference error with scope function parameters

Syniurge:
Previous PR: #6934

This one line fix ensures that if a `StructDeclaration.semantic()` call has to defer itself, `semanticRun` is set to `PASSsemantic`, because it may already have been set to `PASSsemanticdone` if a second `StructDeclaration.semantic()` call on the same struct happened while semantic'ing its members (see how in the Bugzilla issue) and then the deferred call would immediately return.

### DIFF
--- a/src/ddmd/dstruct.d
+++ b/src/ddmd/dstruct.d
@@ -323,10 +323,8 @@ extern (C++) class StructDeclaration : AggregateDeclaration
             userAttribDecl = sc.userAttribDecl;
         }
         else if (symtab && !scx)
-        {
-            semanticRun = PASSsemanticdone;
             return;
-        }
+
         semanticRun = PASSsemantic;
 
         if (!members) // if opaque declaration
@@ -374,6 +372,7 @@ extern (C++) class StructDeclaration : AggregateDeclaration
         {
             assert(type.ty == Terror);
             sc2.pop();
+            semanticRun = PASSsemanticdone;
             return;
         }
         /* Following special member functions creation needs semantic analysis
--- a/test/compilable/imports/fwdref2_test17548.d
+++ b/test/compilable/imports/fwdref2_test17548.d
@@ -1,0 +1,9 @@
+module fwdref2_test17548;
+
+import test17548;
+
+struct S2 {
+    void bar(int arg = .test17548.cnst) {}
+    S1 s;
+    import fwdref2_test17548;
+}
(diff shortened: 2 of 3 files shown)


CyberShadow:
(sorry about that, misclicked)

Syniurge:
I have a question about this code in `StructDeclaration.semantic`:
```d
    else if (symtab && !scx)
    {
        semanticRun = PASSsemanticdone;
        return;
    }
```
What's the rationale? Is it assuming that this may only happen if there was a previous `StructDeclaration.semantic` call which failed with:
```d
        if (!determineFields())
        {
            assert(type.ty == Terror);
            sc2.pop();
            return;
        }
```
which is why it sets `semanticRun` to `PASSsemanticdone`?

If yes the assumption's wrong, and a proper fix would probably be to get rid of that assignment?

Syniurge:
So this works too:
```d
         else if (symtab && !scx)
 -        {
 -            semanticRun = PASSsemanticdone;
              return;
 -        }

         (...)

         if (!determineFields())
          {
              assert(type.ty == Terror);
              sc2.pop();
 +            semanticRun = PASSsemanticdone;
              return;
          }
```
and I think this is the proper fix.

Why isn't the auto-tester running though?

CyberShadow:
@braddr Here's another instance of the autotester apparently ignoring a PR.

braddr:
In this case, it's easy.. new user, not approved yet for builds to be processed.  Just look here:
https://auto-tester.puremagic.com/pulls.ghtml?projectid=14

CyberShadow:
Aah, forgot about that, thanks!

UplinkCoder:
@WalterBright this looks okay at a glance.

Syniurge:
Ping?

src/ddmd/dstruct.d
```diff
@@ -323,10 +323,8 @@ extern (C++) class StructDeclaration : AggregateDeclaration
             userAttribDecl = sc.userAttribDecl;
         }
         else if (symtab && !scx)
-        {
-            semanticRun = PASSsemanticdone;
```
MartinNowak: Looks good, so the problem was that we set this to done here, even though we might be in a recursive call to StructDeclaration.semantic, with the outer semantic still ongoing on it's members, hence providing wrong semanticRun state info for any further members.

MartinNowak (approved):

PR #6941, opened 2017-06-26
Use a proper fallback when `git describe` fails

wilzbach:
https://github.com/dlang/dmd/pull/6935 broke CircleCi on Phobos, because now `git describe` failures lead to hard errors and abortion of the build process.
The fix even removes two lines :)

CC @CyberShadow @MartinNowak

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -470,10 +470,7 @@ $G/idgen: $D/idgen.d $(HOST_DMD_PATH)
 
 VERSION := $(shell cat ../VERSION) # default to checked-in VERSION file
 ifneq (1,$(RELEASE)) # unless building a release
-	VERSION_GIT := $(shell printf "`$(GIT) describe --dirty`") # use git describe
-	ifneq (,$(VERSION_GIT)) # check for git failures
-		VERSION := $(VERSION_GIT)
-	endif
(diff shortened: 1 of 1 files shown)


wilzbach:
How to reproduce:

```sh
git clone --depth 1 -b stable https://github.com/dlang/dmd 
cd dmd && make -f posix.mak all
```

CyberShadow:
How about using `--always`?

MartinNowak:
> How about using --always?

That's *only* the commit hash.

MartinNowak (approved):

CyberShadow:
> That's only the commit hash.

Err, no it's not? I mean it is if it's a shallow clone, but you could still combine it with VERSION if you wanted to. Using VERSION without anything else seems a bit more misleading as it can then be confused with a tagged release.

wilzbach:
Eww due to https://github.com/dlang/dmd/pull/6940 the version change is now also in `master` and breaks new CircleCi builds :(
-> https://github.com/dlang/dmd/pull/6942

MartinNowak:
> Using VERSION without anything else seems a bit more misleading as it can then be confused with a tagged release.

That's our fallback, e.g. for non-git repos, is it really worth having another fallback (e.g. `v2.074.1-ddd-722b169`) for shallow git clones? Also would require extra fiddling to distinguish different `git describe` failures.

PR #6942, opened 2017-06-26
Use a proper fallback when `git describe` fails

wilzbach:
Due to https://github.com/dlang/dmd/pull/6940 the fix from https://github.com/dlang/dmd/pull/6941 is also needed for `master`, see e.g. https://github.com/dlang/phobos/pull/5514

I cherry-picked the same commit to keep things simple as I would like to get CircleCi working at Phobos.

CC @MartinNowak @CyberShadow 

> Using VERSION without anything else seems a bit more misleading as it can then be confused with a tagged release.

Usually `git describe` shouldn't fail and if it's probably a source ball archive. Anyhow my main focus here is really fixing CircleCi - we can always change the behavior in a follow-up.

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -470,10 +470,7 @@ $G/idgen: $D/idgen.d $(HOST_DMD_PATH)
 
 VERSION := $(shell cat ../VERSION) # default to checked-in VERSION file
 ifneq (1,$(RELEASE)) # unless building a release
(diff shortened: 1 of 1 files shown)


MartinNowak:
#6941 already went into master, I picked 89beb113e3b29a70c5685d54bbec8af68c057739 over to stable

wilzbach:
> I picked 89beb11 over to stable

Thanks, Martin!

PR #6943, opened 2017-06-27
Improve error syntax highlighting colors

CyberShadow:
To [quote](https://github.com/dlang/dmd/pull/6777#issuecomment-301279251) @WalterBright:

> I'll leave it to someone with better color sense to pick better colors.

This is a PR for tweaking the color scheme of the new syntax highlighting in DMD error messages. However, since it's likely there's not going to be consensus on which color scheme is best, I'm going to post a few options, each in their own comment, and let GitHub votes (# of thumbs up) decide on the one to go with.

My goals for choosing colors were:
- Ensure the choice is legible in major terminal emulators for all major platforms;
- Avoid using colors already used elsewhere (e.g. for the Error/Warning/Deprecation message labels);
- Avoid colors that are so bright that they are distracting or annoying;
- It should still be visible at a glance which part of the error message represents D syntax, as in many places the highlighting replaced quote characters that indicated this previously.

In all screenshots, the terminals are (in order): macOS Sierra (10.12) Terminal.app; Ubuntu 16.04 default terminal; Windows 10 console.

If anyone has specific color suggestions, I can render previews and put them up here for voting.

### DIFF
--- a/src/ddmd/console.d
+++ b/src/ddmd/console.d
@@ -21,14 +21,23 @@ extern (C) int isatty(int);
 
 enum Color : int
 {
-    black     = 0,
-    red       = 1,
-    green     = 2,
-    blue      = 4,
-    yellow    = red | green,
-    magenta   = red | blue,
-    cyan      = green | blue,
-    white     = red | green | blue,
+    black         = 0,
+    red           = 1,
+    green         = 2,
+    blue          = 4,
+    yellow        = red | green,
+    magenta       = red | blue,
+    cyan          = green | blue,
+    lightGray     = red | green | blue,
+    bright        = 8,
+    darkGray      = bright | black,
+    brightRed     = bright | red,
+    brightGreen   = bright | green,
+    brightBlue    = bright | blue,
+    brightYellow  = bright | yellow,
+    brightMagenta = bright | magenta,
+    brightCyan    = bright | cyan,
+    white         = bright | lightGray,
 }
 
 struct Console
@@ -107,17 +116,16 @@ struct Console
          * Set color and intensity.
          * Params:
          *      color = the color
-         *      bright = turn intensity on
          */
-        void setColor(Color color, bool bright)
+        void setColor(Color color)
         {
             enum FOREGROUND_WHITE = FOREGROUND_RED | FOREGROUND_GREEN | FOREGROUND_BLUE;
             WORD attr = sbi.wAttributes;
             attr = (attr & ~(FOREGROUND_WHITE | FOREGROUND_INTENSITY)) |
-                   ((color & Color.red)   ? FOREGROUND_RED   : 0) |
-                   ((color & Color.green) ? FOREGROUND_GREEN : 0) |
-                   ((color & Color.blue)  ? FOREGROUND_BLUE  : 0) |
-                   (bright ? FOREGROUND_INTENSITY : 0);
+                   ((color & Color.red)    ? FOREGROUND_RED   : 0) |
+                   ((color & Color.green)  ? FOREGROUND_GREEN : 0) |
+                   ((color & Color.blue)   ? FOREGROUND_BLUE  : 0) |
+                   ((color & Color.bright) ? FOREGROUND_INTENSITY : 0);
             SetConsoleTextAttribute(handle, attr);
         }
 
@@ -175,9 +183,9 @@ struct Console
             fprintf(_fp, "\033[%dm", bright);
         }
 
-        void setColor(Color color, bool bright)
+        void setColor(Color color)
         {
-            fprintf(_fp, "\033[%d;%dm", bright, 30 + cast(int)color);
+            fprintf(_fp, "\033[%d;%dm", color & Color.bright ? 1 : 0, 30 + (color & ~Color.bright));
         }
 
         void resetColor()
@@ -199,7 +207,7 @@ struct Console
             assert(0);
         }
 
-        void setColor(Color color, bool bright)
+        void setColor(Color color)
         {
             assert(0);
         }
--- a/src/ddmd/errors.d
+++ b/src/ddmd/errors.d
@@ -24,10 +24,10 @@ import ddmd.console;
  */
 enum Classification
 {
-    error = Color.red,          /// for errors
-    gagged = Color.blue,        /// for gagged errors
-    warning = Color.yellow,     /// for warnings
-    deprecation = Color.magenta,/// for deprecations
+    error = Color.brightRed,          /// for errors
+    gagged = Color.brightBlue,        /// for gagged errors
+    warning = Color.brightYellow,     /// for warnings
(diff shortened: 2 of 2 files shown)


CyberShadow:
Current colors (no change):
current

CyberShadow:
Cyan:
cyan

CyberShadow:
Magenta:
magenta

CyberShadow:
Green:
green

CyberShadow:
Yellow:
yellow

schveiguy:
Why are the user-defined names colored? Typically in my syntax highlighters, the user-defined symbols are not colored like the keywords. I realize they are colored differently, but in some cases, you can't even tell (e.g. yellow on the white background).

CyberShadow:
> Why are the user-defined names colored?

The lexer distinguishes things only up to what is an identifier, keyword, etc., so all identifiers are highlighted in the same way, user-defined or not. Are you asking why identifiers are given a color at all, instead of using the default black or white color? If so, then I believe the answer is so that it's clear where the boundary lies between the error message text (in English) and the quoted code (in D). Previously, those were (inconsistently) indicated by quotes, but with syntax highlighting those become redundant.

jacob-carlborg:
Is it intentional that the color of `Deprecation` changes in each example?

CyberShadow:
Yes, as per goal 2 in the PR description. The original color was also bad (too dark on Windows).

jacob-carlborg:
Which code did you use to produce the those example, I would like to see how it looks in my terminal. For reference, iTerm with solarized dark and light [1].

[1] http://ethanschoonover.com/solarized

CyberShadow:
https://github.com/dlang/dmd/compare/stable...CyberShadow:error-colors - run `./dmd -w -verrors=spec`

schveiguy:
> Are you asking why identifiers are given a color at all, instead of using the default black or white color?

Yes. Here is how github highlights your code:

```D
string s = "foo";
void fun(int i);
```

I wouldn't expect the lexer to be *as* clever to highlight `string` since it's not a keyword, and I'm fine with `fun` having the same style as `i`. But in syntax highlighting, I'm not used to seeing ALL identifiers switched from the default color of the terminal.

All the other colorations look fine, it's just that there should be *some* text that isn't modified in the line of code. The examples are also quite short, I can imagine a wall of template instantiation being much less attractive with all the colorations.

> I believe the answer is so that it's clear where the boundary lies between the error message text (in English) and the quoted code (in D).

This is also possible by coloring the error message, probably the same color as the message type (and I would suggest actually making it bold if possible).

CyberShadow:
> The examples are also quite short, I can imagine a wall of template instantiation being much less attractive with all the colorations.

I think in the great majority of cases, the highlighted parts are going to be very short. Just look at any of the diffs in the test suite of [the PRs enabling syntax highlighting](https://github.com/dlang/dmd/pulls?utf8=%E2%9C%93&q=is%3Apr%20author%3AWalterBright%20syntax%20highlighting).

> This is also possible by coloring the error message, probably the same color as the message type (and I would suggest actually making it bold if possible).

Although I haven't actually tried it, I'm fairly certain it is going to look pretty terrible, not to mention very unlike any other language toolchain.

schveiguy:
Well, this is truly a subjective matter. And luckily this is almost literally bikeshedding :) Changing later if it doesn't look good will be trivial. I just don't have a vote, because I don't like any of the options 😄

CyberShadow:
> Although I haven't actually tried it, I'm fairly certain it is going to look pretty terrible,

I wanted to actually see how it looks. Here's how it looks on my dev terminal (urxvt):



Yep, pretty terrible :)

jacob-carlborg:
> Yes. Here is how github highlights your code:
string s = "foo";
void fun(int i);
I wouldn't expect the lexer to be as clever to highlight string since it's not a keyword,

Technically it's not identified as a keyword, it's identified as a built-in alias. It's used for aliases like `size_t`, `string` and other built-in symbols that are part of druntime or Phobos.

> and I'm fine with fun having the same style as i.

The identifier in function, class and struct declarations are usually specifically identified in TextMate as an entity name.

The TextMate bundle is used by GitHub to do syntax highlighting for D. But GitHub uses its own theme to colorize the text, so it's up to them which colors to choose. For example, in TextMate, function identifiers have a color very similar to keywords, but struct and class identifiers have the regular text color but with the underline style enabled.

schveiguy:
> Yep, pretty terrible :)

Agreed 😄

CyberShadow:
Off:

s-ludwig:
This is about the only thing that was left when I applied the most important ones of the rules that I proposed on the NG to the 16 base colors:
colors

The most important aspect is probably the only one not directly related to the syntax highlighting - having separate style for message text and code. Not reusing warning/error colors or those that are barely visible then mostly only left the dark cyan color for punctuation, which improves readability of the text parts within the code. The punctuation itself is usually of secondary concern when scanning the text, so having the readability slightly impaired is presumably not as critical as for identifiers or keywords.

jacob-carlborg:
For reference, here are pint screens for solarized dark and light on iTerm:
Cyan:
cyan_solarized_dark
cyan_solarized_light

Green:
green_solarized_dark
green_solarized_light

Magenta:
magenta_solarized_dark
magenta_solarized_light

Yellow:
yellow_solarized_dark
yellow_solarized_light

I have to admit, I like the blue color originally used for `Deprecation` but now used for `Error` on the fourth line. The comment is barely visible in the dark theme, but that might be an issue with the theme and not our choosing of colors.

CyberShadow:
> This is about the only thing that was left when I applied the most important ones of the rules that I proposed on the NG to the 16 base colors:

Implemented & rendered:

sönke

Not too bad!

ibuclaw:
> Implemented & rendered:
> sönke
> Not too bad!

I asked my parter who is a bit more on the creative side.  She initially preferred the cyan from the original set, but ultimately opted for this one.  When it came down to it, all the other examples posted here are just too noisy (visually).  This one is the least disruptive, and easiest to read.

However, she was very vocal about the use of blue for speculative errors, suggesting that we use magenta instead, so at least it is readable on all three consoles (which I assume are OSX, Ubuntu, and Windows).

CyberShadow:
> However, she was very vocal about the use of blue for speculative errors, suggesting that we use magenta instead, so at least it is readable on all three consoles (which I assume are OSX, Ubuntu, and Windows).

Yep, the blue sucks, but really it doesn't matter because speculative errors are not something 99% of D users have to worry about. I'll probably change it to dark red.

MartinNowak:
Is this ready @CyberShadow?

CyberShadow:
Whoops, give me a minute :)

CyberShadow:
Updated with Sönke's proposal - it got as many votes as any of the others since it was posted, and it was the only structured attempt at finding the best solution.

We can always tweak things later or make the colors configurable, of course.

PR #6946, opened 2017-06-28
Highlight Error Messages: cond, escape

WalterBright:


JinShil:
LGTM

JinShil (approved):

PetarKirov (approved):
LGTM too.

WalterBright:
There are about 1500 `error()` prints in dmd. This fixes about 15 of them. Taking days/weeks to get these pulled means it will take several years to fix them all.

Fixing them in parallel is problematic because so many of the test files emit multiple messages, so there'd be endless rebasing work.

CyberShadow:
@WalterBright If you find yourself wasting too much time on these, I could try to take over & automate things.

WalterBright:
The main time sink is getting it through the test suite. I do use grep to try to get them done, but there's always nitpicky things and typos to get it through. I suspect it would take more time to automate than to just do it manually.

Some judgement calls need to be done, too, because of variations in how the error messages present code.

ibuclaw:
Why not just convert an entire file at once, then tick it off the list.

WalterBright:
That's pretty much what I've been doing.

ibuclaw:
> That's pretty much what I've been doing.

So then it should total up to about 38 pull requests.  I suppose an upper bound of 38 weeks and conservative lower bound of 38 days seems more likely. :-)

PR #6947, opened 2017-06-28
fix issue 17559 - [REG2.073.0] Wrong line number in stack trace

rainers:
Do not generate debug line information for reuse of common sub expressions.

This also adds new syntax to d_do_test to enable OS specific command line arguments: if the test variable is followed by '(', a list of targets terminated by ')' can be specified to not apply the variable on other targets, e.g.:
```// REQUIRED_ARGS(linux freebsd): -L-export-dynamic```

~~This disables any optimizations in non-optimized builds. They badly effect the debug experience.~~

~~I measured the impact on build phobos unittests without optimizations (not the same system as for the demangling benchmarks):~~

|             | master | this PR
|---------|--------|--------
build time | 2min32s | 2min29s
exec time | 38.0s | 38.2s

### DIFF
--- a/src/ddmd/backend/cgcod.c
+++ b/src/ddmd/backend/cgcod.c
@@ -2621,15 +2621,15 @@ void codelem(CodeBuilder& cdb,elem *e,regm_t *pretregs,bool constflag)
   if (!constflag && *pretregs & (mES | ALLREGS | mBP | XMMREGS) & ~regcon.mvar)
         *pretregs &= ~regcon.mvar;                      /* can't use register vars */
 
-    if (configv.addlinenumbers && e->Esrcpos.Slinnum)
-        cdb.genlinnum(e->Esrcpos);
-
     unsigned op = e->Eoper;
     if (e->Ecount && e->Ecount != e->Ecomsub)     // if common subexp
     {
         comsub(cdb,e,pretregs);
         goto L1;
     }
+
+    if (configv.addlinenumbers && e->Esrcpos.Slinnum)
+        cdb.genlinnum(e->Esrcpos);
 
   switch (op)
   {
--- a/test/d_do_test.d
+++ b/test/d_do_test.d
@@ -84,7 +84,7 @@ struct EnvData
     bool coverage_build;
 }
 
-bool findTestParameter(string file, string token, ref string result)
+bool findTestParameter(const ref EnvData envData, string file, string token, ref string result)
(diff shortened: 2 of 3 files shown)


WalterBright:
I'm not too happy with this. Without comsubs and some jump optimization the generated code can be pretty awful. A better solution may be to not emit the line number for comsub'ed expressions.

rainers:
I suspected it would be controversial, that's why I also didn't target the stable branch.

Line numbers are pretty bad with dmd, and you cannot really improve that if the optimizations throws the information away later. My hope is that we get a good step forward without these optimizations, but I have not tried it in a debugger yet.

rainers:
The auto tester shows `runnable/test17559.d:30 _init [0x8048872]` in the call stack for posix systems. How do I get a `_Dmain` to show up there? This seems ok when built by circleCI.

yazd:
I suspect the problem can be solved by removing this [line](https://github.com/dlang/dmd/blob/e785e39a0d65e8371e5352590950e64e1aa611ce/src/ddmd/e2ir.d#L1276) (untested). I can't think of a reason for why we need line info for referencing a symbol (if I understood the code correctly).
Actually, removing that line also resolves [this](https://issues.dlang.org/show_bug.cgi?id=15189). I started preparing a pull request but got stuck at writing the tests.

rainers:
> I suspect the problem can be solved by removing this line (untested).

This looks quite different, I suspect the frontend attaches a bad line number to a symbol expression. This PR deals with common subexpressions being folded in the backend, thus removing the source line information of the duplicates.

rainers:
> but I have not tried it in a debugger yet.

Tried that now, didn't look much better.

> A better solution may be to not emit the line number for comsub'ed expressions.

I tried to implement that now.

rainers:
Ping! One of the few PRs that is all green ;)

@WalterBright the optimizations are back in place, the PR just removes the emission of debug info for duplicates of common sub expressions now.

This cannot easily target stable because the tests rely on testpdb.d being already there, but it's only on master. The bug is not really a regression, it was hidden in some situations before, though.

braddr:
Manually merging this pull as it sat around long enough waiting to be marked approved that it accumulated github's max 1000 status updates per commit id and won't ever see more until a new commit becomes current for it.  Let that sink in... over 1000 builds done for a single pull request before it got marked for merging.

PR #6948, opened 2017-06-28
improve compiletime floatingpoint to behave a little more like runtim…

UplinkCoder:
…e floatingpoint

This pr enforces `cast` to float or double if the result type is float or double.

which makes fp-math in the constant folder and therefore ctfe more like the runtime equivalent.
There are probably cases I have missed but those are not more wrong then before.
The following code used to fail.
```
float fmaddf(float a, float b, float c)
{
    return b + a * c;
}
static assert(fmaddf(6.7, 8.9, 1.3) == 0x1.19c28ep+4);
static assert(fmaddf(0x1.acccccp+2f, 0x1.166666p+3f, 0x1.4cccccp+0f) == 0x1.168f5cp+4f);
static assert(fmaddf(0x1.acccccp+2f, 0x1.166666p+3f, -0x1.4cccccp+0f) == -0x1.47a8p-7f);
```

with this pr it passes.

### DIFF
--- a/src/ddmd/constfold.d
+++ b/src/ddmd/constfold.d
@@ -72,7 +72,13 @@ extern (C++) UnionExp Neg(Type type, Expression e1)
     Loc loc = e1.loc;
     if (e1.type.isreal())
     {
-        emplaceExp!(RealExp)(&ue, loc, -e1.toReal(), type);
+        auto ty = e1.type.ty;
+        if (ty == Tfloat32)
+                emplaceExp!(RealExp)(&ue, loc, { float r = e1.toReal();r = -r; return r; }(), type);
+        else if (ty == Tfloat64)
+            emplaceExp!(RealExp)(&ue, loc, { double r = e1.toReal();r = -r; return r; }(), type);
+        else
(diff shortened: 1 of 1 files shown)


WalterBright:
This is an old issue.

1. The proposed solution is to add functions to druntime that round to specific precisions, as such rounding is only rarely needed. See https://github.com/dlang/druntime/pull/1621
2. Truncation is not the same thing as round to a precision. It is extremely important to be very careful about using the correct floating point terminology, and which is being selected.
3. There are many places in the compiler that do floating point constants. Doing just a couple leads to erratic behavior.
4. Such behavior changes are not bug fixes, they change the language specification, and so need a DIP.

UplinkCoder:
I think we should aleast strive to have the same behavior at ctfe/constant-folding time as we have at runtime.
Currently my floating-point tests for newCTFE do not pass on the old interpreter because it does ignore `cast(float)`.

UplinkCoder:
Closing as this can be worked around assigning the temporaries to variables.

PR #6949, opened 2017-06-29
fix Issue 17545 - [REG2.072] __traits(getAttributes, name) evaluates …

WalterBright:
…name to value prematurely

I hate adding flags like this, but it seemed the least disruptive way.

### DIFF
--- a/src/ddmd/expression.d
+++ b/src/ddmd/expression.d
@@ -8520,6 +8520,7 @@ extern (C++) final class DotIdExp : UnaExp
 {
     Identifier ident;
     bool noderef;       // true if the result of the expression will never be dereferenced
+    bool wantsym;       // do not replace Symbol with its initializer during semantic()
 
     extern (D) this(Loc loc, Expression e, Identifier ident)
     {
@@ -8772,8 +8773,12 @@ extern (C++) final class DotIdExp : UnaExp
                     if (v.type.ty == Terror)
                         return new ErrorExp();
 
-                    if ((v.storage_class & STCmanifest) && v._init)
+                    if ((v.storage_class & STCmanifest) && v._init && !wantsym)
                     {
+                        /* Normally, the replacement of a symbol with its initializer is supposed to be in semantic2().
+                         * Introduced by https://github.com/dlang/dmd/pull/5588 which should probably
+                         * be reverted. `wantsym` is the hack to work around the problem.
+                         */
                         if (v.inuse)
                         {
                             .error(loc, "circular initialization of %s '%s'", v.kind(), v.toPrettyChars());
--- a/src/ddmd/expression.h
+++ b/src/ddmd/expression.h
@@ -773,6 +773,8 @@ class DotIdExp : public UnaExp
 {
 public:
     Identifier *ident;
+    bool noderef;       // true if the result of the expression will never be dereferenced
+    bool wantsym;       // do not replace Symbol with its initializer during semantic()
 
     static DotIdExp *create(Loc loc, Expression *e, Identifier *ident);
     Expression *semantic(Scope *sc);
--- a/src/ddmd/traits.d
+++ b/src/ddmd/traits.d
@@ -734,6 +734,9 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         }
         else if (e.ident == Id.getMember)
         {
+            if (ex.op == TOKdotid)
+                // Prevent semantic() from replacing Symbol with its initializer
+                (cast(DotIdExp)ex).wantsym = true;
             ex = ex.semantic(scx);
             return ex;
         }
--- a/test/compilable/test17545.d
+++ b/test/compilable/test17545.d
@@ -1,0 +1,16 @@
+/* TEST_OUTPUT:
+---
+tuple((Attrib))
+---
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=17545
+
+module example;
+
+struct Attrib {}
+
+@Attrib enum TEST = 123;
+
+pragma(msg, __traits(getAttributes, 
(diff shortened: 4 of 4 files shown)


src/ddmd/expression.h
```diff
@@ -773,6 +773,8 @@ class DotIdExp : public UnaExp
 {
 public:
     Identifier *ident;
+    bool noderef;       // true if the result of the expression will never be dereferenced
```
ReneZwanenburg: Where is this being used?

CyberShadow: The header file needs to have the same type declarations as the .d file, for compatibility with C++. `noderef` was mistakenly missing here. Adding it here now was required when adding a new field to the .d file, as otherwise they wouldn't match up.

ReneZwanenburg: Ah I see, thanks.

WalterBright:
```
grep noderef *.d
```
will tell the tale!

andralex: Good example of a confusing name - could be "no deref" or "node ref". `noDeref`?

WalterBright: `noderef` is not the subject of this PR.

src/ddmd/expression.d
```diff
@@ -8772,7 +8773,7 @@ extern (C++) final class DotIdExp : UnaExp
                     if (v.type.ty == Terror)
                         return new ErrorExp();
 
-                    if ((v.storage_class & STCmanifest) && v._init)
+                    if ((v.storage_class & STCmanifest) && v._init && !wantsym)
```
JohanEngelen: To improve legibility of frontend code, it would help a lot if you use _much_ longer names for things. How about a method `bool mayReplaceSymbolWithInitializer()` ?

src/ddmd/expression.d
```diff
@@ -8520,6 +8520,7 @@ extern (C++) final class DotIdExp : UnaExp
 {
     Identifier ident;
     bool noderef;       // true if the result of the expression will never be dereferenced
+    bool wantsym;       // do not replace Symbol with its initializer during semantic()
```
JohanEngelen: "during semantic()" - is that exact? What I mean is: the way it's written now allows replacing Symbol with its initializer in semantic2() ?

WalterBright: Yes, it's exact. Normally, the replacement of a symbol with its initializer is supposed to be in semantic2(). But Kenji moved it to semantic() for enums. I did not revert that out of concern for it being a breaking change, hence opted for the simple flag.

src/ddmd/expression.d
WalterBright: For parameter names, I prefer the shorter names. Having lines of code longer than 80 columns is a readability problem.

JohanEngelen: >  longer than 80 columns

newline? :)

(the frontend code is impenetrable. Non-descriptive variable and function names is big problem.)

src/ddmd/expression.d
JohanEngelen: It'd be good to document that in the code.

src/ddmd/expression.d
WalterBright: ItIsIndeedImpenetrableButLongVariableNamesAreNotTheSolution. TheBiggestProblemIsPoorEncapsulationOfTheControlFlow. IveMadeSomeProgressWithThatInOtherPullRequests.
MostOfTheExistingNamesDoFollowAConsistentConventionThatOnceLearnedHelpAGreatDeal.

src/ddmd/expression.d
WalterBright: good idea

yebblies:
This seems pretty fragile.  Isn't there some kind of `Scope` flag for this?

WalterBright:
I thought about a scope flag, but rejected it because it would apply to all expressions in the argument, whereas it should only apply to the specific Expression.

WalterBright:
I tried the following to rebase it on stable:
```
mercury~/forks/dmd master> git fetch upstream stable
From github.com:D-Programming-Language/dmd
 * branch            stable     -> FETCH_HEAD
mercury~/forks/dmd master> git checkout fix17545
Switched to branch 'fix17545'
mercury~/forks/dmd fix17545> git rebase --onto upstream/stable upstream/master fix17545
First, rewinding head to replay your work on top of it...
Applying: fix Issue 17545 - [REG2.072] __traits(getAttributes, name) evaluates name to value prematurely
mercury~/forks/dmd fix17545> git push -f origin fix17545
Counting objects: 49, done.
Delta compression using up to 2 threads.
Compressing objects: 100% (6/6), done.
Writing objects: 100% (6/6), 690 bytes | 0 bytes/s, done.
Total 6 (delta 5), reused 0 (delta 0)
remote: Resolving deltas: 100% (5/5), completed with 5 local objects.
To git@github.com:WalterBright/dmd.git
 + 8668f84...2e6c7ac fix17545 -> fix17545 (forced update)
mercury~/forks/dmd fix17545>
```
but it did not work. As you can see, it is still a PR against master, and it acquired a bunch of other diffs.

CyberShadow:
You need to change it on GitHub too. Click the Edit button on the right of the PR title, and change the "master" dropdown to "stable".

WalterBright:
Ah, yes, that fixed it. Thanks!

src/ddmd/expression.d
MartinNowak: There might simply be different folks (or brains). I'm having problems working with variables that consist of more than 2 nouns as well. While I do read a lot, trying to read prose source code always leaves me dazed. Might be some interlink congestion between the wordy and the logical part of the brain ;).

PR #6951, opened 2017-06-29
highlight: constfold.d

WalterBright:


UplinkCoder:
fail270.d(12) needs updating.

src/ddmd/constfold.d
```diff
@@ -1226,7 +1226,7 @@ extern (C++) UnionExp Index(Type type, Expression e1, Expression e2)
         uinteger_t i = e2.toInteger();
         if (i >= es1.len)
         {
-            e1.error("string index %llu is out of bounds [0 .. %llu]", i, cast(ulong)es1.len);
+            e1.error("string index %llu is out of bounds `[0 .. %llu]`", i, cast(ulong)es1.len);
```
CyberShadow: Why not highlight the index too?

WalterBright: I thought about that, and decided against it:

1. It isn't really code
2. Numbers are quite distinct anyway
3. It's done this way throughout the error message strings

CyberShadow: I think it would be better to highlight anything that comes from the user code, which in most cases means all %* arguments.

andralex (approved):

PR #6954, opened 2017-06-30
Highlight Error Messages: parse.d part 2

WalterBright:


src/ddmd/parse.d
```diff
@@ -1230,7 +1230,7 @@ final class Parser(AST) : Lexer
             break;
 
         case TOKeof:
-            error("declaration expected following attribute, not EOF");
+            error("declaration expected following attribute, not end of file");
```
CyberShadow: Nitpicking but such changes should be done separately and globally

src/ddmd/parse.d
```diff
@@ -7833,7 +7833,7 @@ final class Parser(AST) : Lexer
                                 {
                                     if (peekNext() != TOKidentifier && peekNext() != TOKnew)
                                     {
-                                        error("identifier or new keyword expected following (...).");
+                                        error("identifier or new keyword expected following `(...)`.");
```
CyberShadow: Highlight `new`? It's not "a new keyword", but "the `new` keyword".

test/fail_compilation/diag13609b.d
```diff
@@ -2,8 +2,8 @@
 TEST_OUTPUT:
 ---
 fail_compilation/diag13609b.d(10): Error: base classes are not allowed for `struct`, did you mean `;`?
-fail_compilation/diag13609b.d(11): Error: basic type expected, not EOF
-fail_compilation/diag13609b.d(11): Error: { } expected following struct declaration
+fail_compilation/diag13609b.d(11): Error: basic type expected, not `EOF`
```
CyberShadow: This is inconsistent with the very first hunk (in parse.d)

src/ddmd/parse.d
```diff
@@ -5077,7 +5077,7 @@ final class Parser(AST) : Lexer
 
                 nextToken();
                 //if (token.value == TOKsemicolon)
-                //    error("use '{ }' for an empty statement, not a ';'");
+                //    error("use `{ }` for an empty statement, not `;`");
```
WalterBright: Doing it in another PR is just pointless churn.

CyberShadow: PRs can consist of multiple commits ;)

WalterBright:
There are quite a few inconsistencies and inadequacies in the error messages. To keep the size of this PR down, I didn't do much with that. What I am trying to do is make forward progress.

CyberShadow:
> There are quite a few inconsistencies and inadequacies in the error messages. To keep the size of this PR down, I didn't do much with that. What I am trying to do is make forward progress.

No argument here - but the point is that the inconsistencies are not merely in the error messages, but in your changes to them. It would be nice to avoid introducing more inconsistencies along the way.

src/ddmd/parse.d
WalterBright: As you know, I've long argued for smaller, more focused commits. But this is in response to PRs with hundreds of lines of unrelated changes mixed up together, in order to make reviewing easier, and isolating regressions easier. This is fixing a grammatical error in a line that is already being modified. Putting this in a separate commit does not make it easier to review.

WalterBright:
The EOF thing is already inconsistent in non-trivial ways.

src/ddmd/parse.d
andralex: @CyberShadow well put nice litmus test

ibuclaw (dismissed):
I think most points raised here are valid. Particularly highlighting keywords in messages.

Please also roll the EOF message back to its original. The not semicolon errors are fine in their new form.

Once done will gladly check it off the list.

WalterBright:
> EOF consistency

1. I intend to replace all the `EOF` messages with `end of file`, as it occurred to me that not everyone knows what `EOF` means. Rolling it back here and rolling it forward later is makework.

2. Consistency is not an issue here, as by its nature you won't see two end of file messages in one compile. No user has yet remarked on an inconsistency in multiple error messages.

3. I tend to slightly vary equivalent messages, as it makes the source of an error more conveniently greppable. It also makes it easier to help someone with an error message as to the particular circumstances that caused it.

andralex:
status of this?

CyberShadow:
> status of this?

IMHO addressing the pointed-out nitpicks would have been less effort than defending them.

Regardless, I would like to take over this overall task (adding highlighting to error messages) with some automation. Right after I clear the server maintenance off my plate.

andralex:
>IMHO addressing the pointed-out nitpicks would have been less effort than defending them.

Per my discussion with Walter that's not the case. @ibuclaw could you please approve so we move this along. Thanks!

andralex:
ping @ibuclaw

CyberShadow:
I plan to go through the remaining error messages and update them and the test suite automatically. Let's close this for now so it doesn't waste auto-tester resources.

In the future, please heed the requests here and separate different types of changes into separate PRs.

WalterBright:
This is a good PR. Please do not close it. It should be merged.

CyberShadow:
Alright, let's merge this and fix it later so we don't waste any more time debating trivia.

CyberShadow (approved):

WalterBright:
Thank you!

PR #6956, opened 2017-07-01
betterC on Win32: execute shared static ctors+dtors

WalterBright:
This uses the mechanism Digital Mars C++ uses to call C++ static constructors and destructors to call D shared static constructors and destructors in `-betterC` mode.

Each C compiler does it differently, so some research is necessary to figure it out for other platforms.

### DIFF
--- a/changelog/betterC-STI-STD.dd
+++ b/changelog/betterC-STI-STD.dd
@@ -1,0 +1,36 @@
+-betterC switch now redirects shared static constructors and shared static destructors
+to the C library implementation of how this is done for C and C++ code. This is not
+guaranteed to work on all platforms, as this capability is not required by the C Standard.
+This implementation is for Win32 only. Other platforms will be supported in due course.
+
+Although D guarantees that static construction for imports occurs first,
+no such ordering guarantee is available with -betterC. Typically, however, the order
+matches the order in which object files are presente to the linker.
+
+Example:
+
+---
+import core.stdc.stdio;
+
+shared static this() {
+    printf("hello\n");
+}
+
+shared static ~this() {
+    printf("goodbye\n");
+}
+
+extern (C) int main() {
+    printf("main\n");
+    return 0;
+}
+---
+
+dmd test -betterC
+./test
+
+prints:
+
+hello
+main
+goodbye
--- a/src/ddmd/glue.d
+++ b/src/ddmd/glue.d
@@ -177,9 +177,16 @@ void obj_write_deferred(Library library)
 
 /***********************************************
  * Generate function that calls array of functions and gates.
+ * Params:
+ *      m = module symbol (for name mangling purposes)
+ *      sctors = array of functions
+ *      ectorgates = array of gates
+ *      id = identifier string for generator function
+ * Returns:
+ *      function Symbol generated
  */
 
-Symbol *callFuncsAndGates(Module m, symbols *sctors, StaticDtorDeclarations *ectorgates,
+private Symbol *callFuncsAndGates(Module m, symbols *sctors, StaticDtorDeclarations *ectorgates,
         const(char)* id)
 {
     Symbol *sctor = null;
@@ -198,6 +205,7 @@ Symbol *callFuncsAndGates(Module m, symbols *sctors, StaticDtorDeclarations *ect
         }
 
         localgot = null;
+
         sctor = toSymbolX(m, id, SCglobal, t, "FZv");
         cstate.CSpsymtab = &sctor.Sfunc.Flocsym;
         elem *ector = null;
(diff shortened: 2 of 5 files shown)


test/runnable/betterCsharedctor.d
```diff
@@ -0,0 +1,21 @@
+/* REQUIRED_ARGS: -betterC
+ * PERMUTE_ARGS:
+ */
```
CyberShadow: Can `TEST_OUTPUT` be used here? Currently nothing tests that the ctors/dtors are actually executed.

CyberShadow (dismissed):

jpf91:
For linux it's usually just placing a function pointer into a specific section.

👍 to the idea and more betterC work. One thing I was wondering though is that the  betterC constructors then have slightly different runtime semantics (construction order is not guaranteed and cyclic module contructor dependencies therefore will be supported in betterC mode). I'm not sure if it is a good idea to change semantics with a compiler switch.

An alternative is to use an `@CContructor` UDA and normal module level functions, I think LDC already does that. Another benefit of the UDA approach is that you now can easily implement multiple named betterC constructors per module.

Ping @andralex as this is some kind of language change / discussion.

WalterBright:
Don't pull this yet, I have a better idea.

test/runnable/betterCsharedctor.d
WalterBright: Testing the output will come when the other platforms are supported.

WalterBright:
Ok, better idea done.

src/ddmd/glue.d
```diff
@@ -484,6 +492,14 @@ void genObjFile(Module m, bool multiobj)
         m.sshareddtor = callFuncsAndGates(m, &sshareddtors, null, "__modshareddtor");
         m.stest = callFuncsAndGates(m, &stests, null, "__modtest");
 
+        if (global.params.betterC)
+        {
+            if (m.ssharedctor)
+                objmod.setModuleCtorDtor(m.ssharedctor, true);
+            if (m.ssharedctor)
```
andralex: BUG should be `if (m.sshareddtor)`

andralex (changes requested):
Cool beans. This should come with a changelog entry and documentation. Please fix the bug and perhaps add more test (only ctor/dtor).

src/ddmd/glue.d
WalterBright: danged copy/pasta errors!

WalterBright:
@andralex done

WalterBright:
While this is waiting, there's https://github.com/dlang/dmd/pull/6957 for the non-controversial bits.

ibuclaw:
Perhaps `-betterC` should become a meta-flag for other features.  Same as what is done for `-release`.

jacob-carlborg:
Is this the same as `__attribute__((constructor))` on GCC/Clang? In that case I agree with @jpf91 that it should be a separate functionality, like a UDA or `pragma` on a regular function. This is a useful feature to have even when `betterC` is not used.

We also already have this for ELF targets (Linux and FreeBSD). It was added for the implementation of dynamic libraries.

changelog/betterC-STI-STD.dd
```diff
@@ -0,0 +1,36 @@
+-betterC switch now redirects shared static constructors and shared static destructors
+to the C library implementation of how this is done for C and C++ code. This is not
+guaranteed to work on all platforms, as this capability is not required by the C Standard.
+This implementation is for Win32 only. Other platforms will be supported in due course.
+
+Although D guarantees that static construction for imports occurs first,
+no such ordering guarantee is available with -betterC. Typically, however, the order
+matches the order in which object files are presente to the linker.
```
schveiguy: presented?

schveiguy (commented):
I'm against such a change, since D users expect to have proper ordering of D static ctor/dtors regardless of link order. I would recommend instead to simply use the C compiler's feature with that compiler. I can't imagine chasing a bug due to linker order is going to be fun when you switch to `-betterC` mode.

Just giving my opinion for what it's worth, not expecting a response or acceptance.

CyberShadow:
> I'm against such a change, since D users expect to have proper ordering of D static ctor/dtors regardless of link order.

Not having switches that subtly change language semantics is probably a good idea.

How about allowing the use of this feature through `extern(C) static this() { ... }` regardless of `-betterC`?

schveiguy:
> How about allowing the use of this feature through extern(C) static this() { ... } regardless of -betterC?

I don't like that either, because I guarantee there is some poor soul who put `extern(C):` at the top of their module, without putting `extern(D)` in front of their static ctor/dtor, and now you will drastically change the behavior.

jacob-carlborg:
It can be implemented using a UDA or pragma, like for GDC and LDC.

MartinNowak:
> Each C compiler does it differently, so some research is necessary to figure it out for other platforms.

See #7182

MartinNowak:
We might close this in favor of https://github.com/dlang/dmd/pull/7182

For the unlikely need of a codebase to support both, `D` and `D -betterC`, `version (D_betterC)` could be used to switch between `shared static this()` and `pragma(crt_constructor)`.

andralex:
Let's go with https://github.com/dlang/dmd/pull/7182. Further improvements will come on top of it.

changloong:
Hi, @andralex @WalterBright 

please check on https://issues.dlang.org/show_bug.cgi?id=18100 

crt_constructor need to allow set immutable(which mean it only should be call once, and can not been call from main function).   If we simple use extern(C) shared static this will fix all problem.

PR #6957, opened 2017-07-02
refactor: convert Obj.funcptr() to Obj.setModuleCtorDtor()

WalterBright:
Because https://github.com/dlang/dmd/pull/6956 has some controversy, I factored out the uncontroversial bits into this one.

I.e. redo Obj.funcptr() to make its name reflective of what it does, and remove its dependency on the specifics of the identifier so it can be ported to other platforms.

### DIFF
--- a/src/ddmd/backend/cgobj.c
+++ b/src/ddmd/backend/cgobj.c
@@ -1724,11 +1724,14 @@ void Obj::staticdtor(Symbol *s)
 
 
 /***************************************
- * Stuff pointer to function in its own segment.
- * Used for static ctor and dtor lists.
+ * Set up function to be called as static constructor on program
+ * startup or static destructor on program shutdown.
+ * Params:
+ *      s = function symbol
+ *      isCtor = true if constructor, false if destructor
  */
 
-void Obj::funcptr(Symbol *s)
+void Obj::setModuleCtorDtor(Symbol *s, bool isCtor)
 {
     // We need to always put out the segments in triples, so that the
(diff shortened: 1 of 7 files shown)


src/ddmd/backend/cgobj.c
```diff
@@ -1724,11 +1724,14 @@ void Obj::staticdtor(Symbol *s)
 
 
 /***************************************
- * Stuff pointer to function in its own segment.
- * Used for static ctor and dtor lists.
+ * Set up function to be called as static constructor on program
```
UplinkCoder: > Stuff pointer to function in its own segment. 

This is useful information.
please keep it.

andralex (approved):

MartinNowak:
See https://www.airs.com/blog/archives/51 @WalterBright for Info on ELF ctors/dtors.
Basically this
```c
void foo() __attribute__((constructor, destructor));
void foo() {}
```
will generate two entries, one in the `.init_array` section (using DT_INIT_ARRAY), one in `.fini_array` (using DT_FINI_ARRAY).
```
  [ 4] .init_array       INIT_ARRAY       0000000000000000  00000048
       0000000000000008  0000000000000008  WA       0     0     8
  [ 6] .fini_array       FINI_ARRAY       0000000000000000  00000050
       0000000000000008  0000000000000008  WA       0     0     8
```

BTW, I find the function name `setModuleCtorDtor` misleading, as this is NOT using the module ctor/dtor mechanism, but the earlier C ctors/dtors one.

PR #6958, opened 2017-07-02
fix Issue 17582 - Applying const to struct declaration should make th…

WalterBright:
…e struct type const

### DIFF
--- a/test/runnable/testconst.d
+++ b/test/runnable/testconst.d
@@ -2806,9 +2806,15 @@ void test6982()
 /************************************/
 // 7038
 
-static assert(!is(S7038 == const));
+static assert(is(S7038 == const));
 const struct S7038{ int x; }
-static assert(!is(S7038 == const));
+static assert(is(S7038 == const));
+
+shared struct S7038b{ int x; }
+static assert(is(S7038b == shared));
+
+immutable struct S7038c{ int x; }
+static assert(is(S7038c == immutable));
 
 static assert(!is(C7038 == const));
 const class C7038{ int x; }
@@ -2817,7 +2823,7 @@ static assert(!is(C7038 == const));
 void test7038()
 {
     S7038 s;
-    static assert(!is(typeof(s) == const));
+    static assert(is(typeof(s) == const));
     static assert(is(typeof(s.x) == const int));
 
     C7038 c;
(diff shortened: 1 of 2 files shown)


UplinkCoder:
do we really need this ?
I supposed it does no harm though.

WalterBright:
See also https://github.com/dlang/dmd/pull/783 where this bug was introduced.

UplinkCoder:
@WalterBright I think you can merge this.
There is no harm here.

test/runnable/testconst.d
```diff
@@ -2806,9 +2806,9 @@ void test6982()
 /************************************/
 // 7038
```
andralex: Just to really nail it down, could you please write this:
```d
static assert(is(S7038 == const S7038)));
```

There should be tests for the other qualifiers as well, notably shared.

andralex (changes requested):

andralex:
This is a change in language semantics so it needs a changelog.

I also realized the following problem. Consider:

```d
immutable struct S
{
    void fun() {}
    void fun() immutable {}
}
```

This should not compile because fun is the same function defined twice. Similarly:

```d
immutable struct S {}
void fun(S) {}
void fun(immutable S) {}
```

should not work because fun is defined twice.

WalterBright:
> This should not compile

It's actually an old issue.
```
void foo() { }
void foo() { }
```
compiles but does not link, as does your example. I.e. it is a separate issue from this PR.

test/runnable/testconst.d
WalterBright: Since the code to implement this is generic for all qualifiers, that is unnecessary. But I'll add one for shared.

andralex (approved):
Thx! Can you please add to the changelog before merging?

andralex:
The next big step is classes and interfaces.

UplinkCoder:
is `inout struct` something that even makes sense ?

WalterBright:
> Can you please add to the changelog before merging?

Since it's a regression fix it will anyway.

wilzbach:
> Since it's a regression fix it will anyway.

Yup, but as you have already recognized this as regression - why didn't you base it onto stable?
Imho one needs to try really hard to overlook this warning:

image

CC @CyberShadow - I am running out of ideas here:
- should we call it "Errors"?
- should we block auto-merge?

CyberShadow:
The PR was created before the new warning notice was added, so let's give it a bit more time.

ZombineDev:
@wilzbach the regression is from 2.059. It would be too risky to include this fix in stable, IMO.

PR #6748, opened 2017-05-07
Fix Issue 7016 - local import does not create -deps dependency

RazvanN7:

PR #6961, opened 2017-07-03
Fix issue 16746 - Please output Makefile-style depfiles for ninja and make

eigenraven:
This adds `-M`, `-Mffile`, `-Mddirectory`, `-Mttarget` commandline options akin to C(++) compilers, and implements generation of make depfiles which can be used by buildsystems such as Ninja, Make or Meson.

Link to issue: https://issues.dlang.org/show_bug.cgi?id=16746

### DIFF
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -377,6 +377,18 @@ dmd -cov -unittest myprog.d
             source module to be compiled. This name can be overridden with
             the $(SWLINK -of) switch.`,
         ),
+        Option("M",
+            "generate make-style dependencies"
+        ),
+        Option("Mf<filename>",
+            "write make-style dependencies to filename"
+        ),
+        Option("Md<directory>",
+            "write make-style dependencies to directory"
+        ),
+        Option("Mt<name>",
+            "specifies the name of the target rule in the make depfile"
+        ),
         Option("m32",
             "generate 32 bit code",
             `$(UNIX Compile a 32 bit executable. This is the default for the 32 bit dmd.)
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -382,6 +382,7 @@ extern (C++) final class Module : Package
     Dsymbols* decldefs;         // top level declarations for this Module
 
     Modules aimports;           // all imported modules
+    Strings astringImports;     // all string-imported files
 
     uint debuglevel;            // debug level
     Identifiers* debugids;      // debug identifiers
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -208,6 +208,11 @@ struct Param
     const(char)* moduleDepsFile;        // filename for deps output
     OutBuffer* moduleDeps;              // contents to be written to deps file
 
+    bool doMakeDeps;                    /// generate make-style dependencies file
+    const(char)* makeDepsFile;          /// filename for make-style dependencies output
+    const(char)* makeDepsDir;           /// directory for make-style dependencies output
+    const(char)* makeDepsTarget;        /// the make target name
+
     // Hidden debug switches
     bool debugb;
     bool debugc;
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -848,21 +848,106 @@ private int tryMain(size_t argc, const(char)** argv)
         fatal();
 
     // inlineScan incrementally run semantic3 of each expanded functions.
-    // So deps file generation should be moved after the inlinig stage.
-    if (global.params.moduleDeps)
+    // So deps file generation should be moved after the inlining stage.
+    if (global.params.moduleDeps || global.params.doMakeDeps)
     {
-        foreach (i; 1 .. modules[0].aimports.dim)
-            semantic3OnDependencies(modules[0].aimports[i]);
-
-        OutBuffer* ob = global.params.moduleDeps;
-        if (global.params.moduleDepsFile)
+        if (global.params.moduleDeps)
         {
-            auto deps = File(global.params.moduleDepsFile);
-            deps.setbuffer(cast(void*)ob.data, ob.offset);
-            writeFile(Loc.initial, &deps);
+            foreach (i; 1 .. modules[0].aimports.dim)
+                semantic3OnDependencies(modules[0].aimports[i]);
+            OutBuffer* ob = global.params.moduleDeps;
+            if (global.params.moduleDepsFile)
+            {
+                auto deps = File(global.params.moduleDepsFile);
+                deps.setbuffer(cast(void*)ob.data, ob.offset);
+                writeFile(Loc.initial, &deps);
+            }
+            else
+                printf("%.*s", cast(int)ob.offset, ob.data);
         }
-        else
-            printf("%.*s", cast(int)ob.offset, ob.data);
+        if (global.params.doMakeDeps)
+        {
+            // Recursive dependencies scanner, fills string tables with module source paths and string-imported files paths
+            void recursivelyAddModuleDependencies(Module m, StringTable* moduleImportedPaths, StringTable* stringImportedPaths)
+            {
+                auto mpath = m.srcfile.toString();
+                // Skip modules already in the table, add self
+                if (moduleImportedPaths.insert(mpath, null) is null)
+                    return;
+                // Add string imports
+                foreach (i, const(char)* imp; m.astringImports)
+                {
+                    stringImportedPaths.insert(imp, strlen(imp), null);
+                }
+                // Process module imports
+                foreach (i, impmod; m.aimports)
+                {
+                    // Skip self-imports
+                    if (impmod is m || impmod is null)
+                        continue;
+                    // Recurse
+                    recursivelyAddModuleDependencies(impmod, moduleImportedPaths, stringImportedPaths);
+                }
+            }
+
+            __gshared static OutBuffer mdout;
+            bool mdSingleFile = global.params.makeDepsFile !is null;
+            File* mdOutput;
+            if (mdSingleFile)
+            {
+                mdOutput = modules[0].setOutfile(global.params.makeDepsFile, global.params.makeDepsDir, modules[0].arg, "dep");
+                mdOutput._ref = 1;
+            }
+
+            extern(C++) int modulePathOutputSink(const(StringValue)* val)
+            {
+                mdout.writestring(" \\\n ");
+                mdout.writestring(val.toDchars()[0..val.length]);
+
+                return 0;
+            }
+
+            foreach (i, m; modules)
+            {
+                // Skip the modules that were not on the commandline explicitly
+                if (m.srcfile is null || m.arg is null || m.ident == Id.entrypoint || m.ident.toString() == "__main")
+                    continue;
+                // Each module needs to walk its entire dependency tree for the output to be correct
+                StringTable moduleImportedPaths, stringImportedPaths;
+                moduleImportedPaths._init();
+                stringImportedPaths._init();
+                recursivelyAddModuleDependencies(m, &moduleImportedPaths, &stringImportedPaths);
+                if (global.params.makeDepsTarget is null)
+                {
+                    // no actual file is created, uses just the path
+                    File* objfile = m.setOutfile(global.params.objname, global.params.objdir, m.arg, global.obj_ext);
+                    objfile.name.escapeSpaces(&mdout);
+                }
+                else
+                {
+                    mdout.writestring(global.params.makeDepsTarget);
+                }
+                mdout.writeByte(':');
+                moduleImportedPaths.apply(&modulePathOutputSink);
+                stringImportedPaths.apply(&modulePathOutputSink);
+                mdout.writeByte('\n');
+                mdout.writeByte('\n');
+                if (!mdSingleFile)
+                {
+                    File* sout = m.setOutfile(global.params.makeDepsFile, global.params.makeDepsDir, m.arg, "dep");
+                    sout._ref = 1;
+                    sout.setbuffer(mdout.data, mdout.offset);
+                    sout.write();
+                    mdout.reset();
+                }
+            }
+
+            if (mdSingleFile)
+            {
+                mdOutput.setbuffer(mdout.data, mdout.offset);
+                mdOutput.write();
+            }
+        }
     }
 
     printCtfePerformanceStats();
@@ -2188,6 +2273,32 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                     goto Lerror;
                 }
                 params.moduleDeps = new OutBuffer();
+            }
+            else if (p[1] == 'M')
+            {
+                params.doMakeDeps = true;
+                switch (p[2])
+                {
+                case '\0':
+                    break;
+                case 'f':
+                    if (p[3] == '\0')
+                        goto Lnoarg;
+                    params.makeDepsFile = p + 3 + (p[3] == '=');
+                    break;
+                case 'd':
+                    if (p[3] == '\0')
+                        goto Lnoarg;
+                    params.makeDepsDir = p + 3 + (p[3] == '=');
+                    break;
+                case 't':
+                    if (p[3] == '\0')
+                        goto Lnoarg;
+                    params.makeDepsTarget = p + 3 + (p[3] == '=');
+                    break;
+                default:
+                    goto Lerror;
+                }
             }
             else if (arg == "-main")             // https://dlang.org/dmd.html#switch-main
             {
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -771,6 +771,57 @@ nothrow:
     {
         return str ? str[0 .. strlen(str)] : null;
     }
+
+    /**
+     * Escapes '(', ')', backslashes and whitespace in the path.
+     *
+     * Params:
+     *   buf = Buffer to write the escaped path to
+     */
+    extern(D) void escapeSpaces(OutBuffer* buf) const nothrow
+    {
+        const(char)* it = this.str;
+        while (1)
+        {
+            switch (*it)
+            {
+            case 0:
+                return;
+            case '(':
+            case ')':
+            case '\\':
+            case ' ':
+            case '\t':
+            case '\f':
+            case '\v':
+                buf.writeByte('\\');
+                goto default;
+            default:
+                buf.writeByte(*it);
+                break;
+            }
+            it++;
+        }
+    }
+    ///
+    unittest
+    {
+        OutBuffer buf;
+        FileName("simple_path.d").escapeSpaces(&buf);
+        assert(buf.peekSlice() == "simple_path.d");
+        buf.reset();
+
+        FileName("spaced path.d").escapeSpaces(&buf);
+        assert(buf.peekSlice() == "spaced\\ path.d");
+        buf.reset();
+
+        FileName("windows\\folder with\ttab/and(parens).d").escapeSpaces(&buf);
+        assert(buf.peekSlice() == "windows\\\\folder\\ with\\\ttab/and\\(parens\\).d");
+        buf.reset();
+
+        FileName("specials()\\ \t\f\v.d").escapeSpaces(&buf);
+        assert(buf.peekSlice() == "specials\\(\\)\\\\\\ \\\t\\\f\\\v.d");
+    }
 }
 
 version(Windows)
--- a/test/compilable/extra-files/makedeps.sh
+++ b/test/compilable/extra-files/makedeps.sh
@@ -1,0 +1,15 @@
+#!/usr/bin/env bash
+set -e
+GDEPFILE=${RESULTS_DIR}/compilable/makedeps.dep
+GDEPFILE_a=${RESULTS_DIR}/compilable/makedeps_a.dep
+grep  'makedeps_[0-9]*.o:' ${GDEPFILE} || # some platforms use .obj instead of .o for object files.
+grep  'makedeps_[0-9]*.obj:' ${GDEPFILE}
+# The test runner will generate a single object file from both source files, hence the same target name
+grep  'makedeps_[0-9]*.o:'  ${GDEPFILE_a} ||
+grep  'makedeps_[0-9]*.obj:'  ${GDEPFILE_a}
+grep  'makedeps.d'  ${GDEPFILE}
+grep  'makedeps.sh'  ${GDEPFILE}
+grep  'makedeps_a.d'  ${GDEPFILE}
+grep  'object.d'  ${GDEPFILE}
+! grep  '__entrypoint'  ${GDEPFILE}
+rm -f ${GDEPFILE} ${GDEPFILE_a}
--- a/test/compilable/extra-files/makedeps2.sh
+++ b/test/compilable/extra-files/makedeps2.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+set -e
+grep  "make\\ deps:"  ${RESULTS_DIR}/compilable/makedeps2.dep
+grep  'makedeps2.d'  ${RESULTS_DIR}/compilable/makedeps2.dep
+grep  'makedeps.sh'  ${RESULTS_DIR}/compilable/makedeps2.dep
+grep  'makedeps_a.d'  ${RESULTS_DIR}/compilable/makedeps2.dep
+grep  'object.d'  ${RESULTS_DIR}/compilable/makedeps2.dep
+! grep  '__entrypoint'  ${RESULTS_DIR}/compilable/makedeps2.dep
+rm -f ${RESULTS_DIR}/compilable/makedeps2.dep
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -5302,6 +5302,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             ob.writestring(")");
             ob.writenl();
         }
+        sc.instantiatingModule().astringImports.push(name);
+
 
         {
             auto f = File(name);
--- a/src/dmd/globals.h
+++ b/src/dmd/globals.h
@@ -180,6 +180,11 @@ struct Param
     const char *moduleDepsFile; // filename for deps output
     OutBuffer *moduleDeps;      // contents to be written to deps file
 
+    bool doMakeDeps;                  // generate make-style dependencies file
+    const char *makeDepsFile;         // filename for make-style dependencies output
+    const char *makeDepsDir;          // directory for make-style dependencies output
+    const char *makeDepsTarget;       // the generated make target name
+
     // Hidden debug switches
     bool debugb;
     bool debugc;
--- a/src/dmd/module.h
+++ b/src/dmd/module.h
@@ -105,6 +105,7 @@ public:
     Dsymbols *decldefs;         // top level declarations for this Module
 
     Modules aimports;             // all imported modules
+    Strings astringImports;     // all string-imported files
 
     unsigned debuglevel;        // debug level
     Strings *debugids;      // debug identifiers
--- a/test/compilable/extra-files/makedeps_a.d
+++ b/test/compilable/extra-files/makedeps_a.d
@@ -1,0 +1,5 @@
+module makedeps_a;
+
+void a_func()
+{
+}
--- a/test/compilable/makedeps.d
+++ b/test/compilable/makedeps.d
@@ -1,0 +1,17 @@
+// PERMUTE_ARGS:
+// REQUIRED_ARGS: -Md${RESULTS_DIR}/compilable/ -Jcompilable/extra-files
+// POST_SCRIPT: compilable/extra-files/makedeps.sh
+// EXTRA_SOURCES: /extra-files/makedeps_a.d
+
+module makedeps;
+
+// Test import statement
+import makedeps_a;
+
+// Test import expression
+enum SH_TEXT = import("makedeps.sh");
+
+void main()
+{
+    a_func();
+}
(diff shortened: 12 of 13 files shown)


wilzbach:
Seems like `gdc` doesn't like this change:

```
.. compilable/makedeps.d          -M -MF=test_results/compilable/makedeps.dep ()
*** glibc detected *** ../src/dmd: double free or corruption (!prev): 0x0000000001fc7280 ***
======= Backtrace: =========
/lib/x86_64-linux-gnu/libc.so.6(+0x7da66)[0x2b5c8318ca66]
../src/dmd[0x63a5f1]
../src/dmd[0x65c9dd]
../src/dmd[0x660166]
../src/dmd[0x66046e]
../src/dmd[0x64060e]
../src/dmd[0x6376a5]
../src/dmd[0x637719]
../src/dmd[0x63730f]
../src/dmd[0x6374f2]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0xed)[0x2b5c831307ed]
../src/dmd[0x403c09]
======= Memory map: ========
00400000-00782000 r-xp 00000000 fc:06 17744159                           /home/travis/build/dlang/dmd/src/dmd
00782000-007d2000 rw-p 00382000 fc:06 17744159                           /home/travis/build/dlang/dmd/src/dmd
007d2000-007df000 rw-p 00000000 00:00 0 
01f41000-01fd6000 rw-p 00000000 00:00 0                                  [heap]
2b5c829d4000-2b5c829f6000 r-xp 00000000 fc:06 109174333                  /lib/x86_64-linux-gnu/ld-2.15.so
```

src/ddmd/expression.d
```diff
@@ -8413,6 +8413,22 @@ extern (C++) final class ImportExp : UnaExp
             ob.writestring(")");
             ob.writenl();
         }
+        if (global.params.makeDeps !is null)
```
wilzbach: Hmm CodeCov says this is tested:

image

ibuclaw:
> Seems like gdc doesn't like this change:

FYI, gdc compiled test uses GC allocation, not bump-the-pointer. So this means that there is a genuine logic bug in the implementation.

src/ddmd/mars.d
```diff
@@ -1519,20 +1555,56 @@ Language changes listed by -transition=id:
...
+            foreach (dep; (*global.params.makeDeps)[])
+            {
+                depBuffer.writeByte('\\');
+                depBuffer.writenl();
+                depBuffer.writeByte(' ');
+                escapePathSpaces(depBuffer, dep);
+                depBuffer.writeByte(' ');
+            }
+            auto depf = File(outpath);
+            depf.setbuffer(cast(void*)depBuffer.data, depBuffer.offset);
```
ibuclaw: This is normally why there's a double free.  File is freeing the buffer owned by OutBuffer.  You need to tell `depf._ref = 1`.

ibuclaw (changes requested):

src/ddmd/globals.h
```diff
@@ -199,6 +203,7 @@ struct Global
     const char *hdr_ext;        // for D 'header' import files
     const char *json_ext;       // for JSON files
     const char *map_ext;        // for .map files
+    const char *makedep_ext;    // for make-style dependency files
```
ibuclaw: All `_ext` fields should just go. There is no need for them.

src/ddmd/mars.d
```diff
@@ -975,6 +979,38 @@ Language changes listed by -transition=id:
...
+                        error(Loc(), "-M/-MD can only be provided once!");
+                    global.params.makeDeps = new Strings();
+                    break;
+                case 'D':
+                    if(p[3] != '\0')
+                        goto Lerror;
+                    if(global.params.makeDeps !is null)
+                        error(Loc(), "-M/-MD can only be provided once!");
+                    global.params.makeDeps = new Strings();
+                    break;
```
ibuclaw: -MD writes deps to a file implicitly.

src/ddmd/globals.d
```diff
@@ -172,6 +172,10 @@ struct Param
     const(char)* moduleDepsFile;        // filename for deps output
     OutBuffer* moduleDeps;              // contents to be written to deps file
 
+    const(char)* makeDepsFile;          // filename for make-style dependencies output
+    const(char)* makeDepsTarget;        // the generated make target name
+    Array!(const(char)*)* makeDeps;     // array of paths that the compiled module(s) depend on
```
ibuclaw: I wonder if these are strictly needed, or we should start requesting that new driver-only fields be added privately in mars.d so as to not affect other backends.

ibuclaw (changes requested):

src/ddmd/expression.d
kubasz: I've added an import expression to the test file, this should cover these lines.

src/ddmd/globals.d
kubasz: The `makeDeps` field could be private, but the `File` and `Target` can be useful for other drivers to input their commandline options to dmdfe. I've basically followed what changes the `-deps` option that was already built into dmd introduced, so that's why I added these fields here.

src/ddmd/mars.d
kubasz: Fixed. This is my first encounter with dmd's code, so sorry that I didn't notice the `_ref` field in File.

MartinNowak (changes requested):
Thanks, interesting PR :).
While dmd's command line is weird, it doesn't become better by adopting switches with different conventions. Please make your additions consistent with existing switches.
Furthermore this apparently cannot generate multiple output deps files, we need sth. that also works with multiple source and output files, while also working with parallel compilation (can't write everything into one file).

kubasz:
Sorry for the long delay in fixing my PR, I had a surgery and needed time to recover. I've pushed the fixes now:
- Changed the cmdline args to match -D*: -M, -Mf, -Md; -Mttarget follows the convention, they also respect -op by generating paths using the same logic as the rest of the code (Module's setOutfile)
- Added a test for the target itself with a semicolon in the shell script
- The code now fully handles multiple input modules:
  - if Mf is specified, outputs all dependencies to a single file
  - if Mf is not specified, generates a new ".dep" file for each input file basing on -Md/-op/cwd
  - if MtTARGET is specified, all the targets are replaced with `TARGET:` (it's also clang's behaviour)
- Dependency searching now uses a string hash table instead of the previous O(n^2) search for identical matches

kubasz:
I have rebased the PR against the current master branch.

ximion:
Any news on this? It's kind of important to make splitbuilds work out of the box in Automake, Meson & Co.

kubasz:
Would it be possible for you to re-review this PR? @MartinNowak @ibuclaw

kubasz:
Rebased against master.

wilzbach:
(closed/reopened to restart CIs)

test/compilable/extra-files/makedeps2.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+grep  "make\\ deps:"  ${RESULTS_DIR}/compilable/makedeps2.dep || exit 1
+grep  'makedeps2.d'  ${RESULTS_DIR}/compilable/makedeps2.dep || exit 1
+grep  'makedeps.sh'  ${RESULTS_DIR}/compilable/makedeps2.dep || exit 1
+grep  'makedeps_a.d'  ${RESULTS_DIR}/compilable/makedeps2.dep || exit 1
+grep  'object.d'  ${RESULTS_DIR}/compilable/makedeps2.dep || exit 1
```
wilzbach: Better use `set -e` instead of the constant `|| exit 1`. 
See also: https://github.com/dlang/dmd/pull/8083

test/compilable/extra-files/makedeps2.sh
```diff
@@ -0,0 +1,9 @@
...
+grep  '__entrypoint'  ${RESULTS_DIR}/compilable/makedeps2.dep && exit 1
+rm -f ${RESULTS_DIR}/compilable/makedeps2.dep
+exit 0
```
wilzbach: Superfluous

src/dmd/utils.d
```diff
@@ -132,3 +132,36 @@ extern (C++) void escapePath(OutBuffer* buf, const(char)* fname)
...
+            return;
+        case '(':
+        case ')':
+        case '\\':
+        case ' ':
+        case '\t':
+        case '\f':
+        case '\v':
+            buf.writeByte('\\');
+            goto default;
```
wilzbach: This is uncovered:

image

How about adding an inline unittest to cover it?

src/dmd/cli.d
```diff
@@ -352,6 +352,18 @@ dmd -cov -unittest myprog.d
...
+        ),
+        Option("Mf<filename>",
+            "write make-style dependencies to filename"
+        ),
+        Option("Md<directory>",
+            "write make-style dependencies to directory"
+        ),
+        Option("Mt<name>",
+            "specifies the name of the target rule in the make depfile"
+        ),
```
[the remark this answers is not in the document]
kubasz: I think they're necessary, and mostly follow the format of the ones in other compilers, like gcc or clang which have M, MM, MF, MT, MQ and more.

src/dmd/mars.d
```diff
@@ -826,20 +826,109 @@ private int tryMain(size_t argc, const(char)** argv)
...
+                printf("%.*s", cast(int)ob.offset, ob.data);
+        }
+        if (global.params.doMakeDeps)
+        {
+            // Recursive dependencies scanner, fills string tables with module source paths and string-imported files paths
+            void recAddMDeps(Module m, StringTable* mpaths, StringTable* spaths)
+            {
+                const(char)* mpath = m.srcfile.toChars();
+                // Skip modules already in the table, add self
+                if(mpaths.insert(mpath, strlen(mpath), null) is null)
```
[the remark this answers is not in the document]
kubasz: The File structure has no toString method and other code that interacts with file paths also uses C strings, so until that is changed I can't really do anything else in this PR without adding that functionality to File or FilePath.

src/dmd/utils.d
```diff
@@ -132,3 +132,36 @@ extern (C++) void escapePath(OutBuffer* buf, const(char)* fname)
...
+
+
+/**
+ * Takes a path, and escapes '(', ')', backslashes and whitespace
+ *
+ * Params:
+ *   buf = Buffer to write the escaped path to
+ *   fname = Path to escape
+ */
+extern (C++) void escapePathSpaces(OutBuffer* buf, const(char)* fname)
```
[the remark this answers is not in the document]
kubasz: I moved it to the D side, but because it manipulates on paths and paths in the dmd codebase are all C strings with no facilities for conversion to/from D strings, I'm following that in here and using a C string.

test/compilable/extra-files/makedeps.sh
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env bash
+GDEPFILE=${RESULTS_DIR}/compilable/makedeps.dep
+GDEPFILE_a=${RESULTS_DIR}/compilable/makedeps_a.dep
+grep  'makedeps_[0-9]*.o:' ${GDEPFILE} ||
```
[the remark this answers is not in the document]
kubasz: I've added a comment explaining this case - on Windows (and maybe other platforms) the object file extension is .obj instead of .o, so this just allows to grep over the two cases clearly without resorting to extended regular expressions.

src/dmd/utils.d
kubasz: I've added an unittest after this function that tests it for these special characters and strings without them.

kubasz:
I've updated the PR, fixing most of wilzbach's comments (many thanks for reviewing my code). I also commented on a few of them with my rationale for the changes as they are now. I hope this PR can be made ready to merge as soon as possible, because of how important it is for implementing proper build system support for D.

ZombineDev:
Wouldn't it be better if all the necessary information is put in the json produced by `-Xf` and then any tool can excract any subset of the information and the reformat it appropriately? This would solve the issue with adding yet another flag.

ximion:
> Wouldn't it be better if all the necessary information is put in the json produced by -Xf and then any tool can excract any subset of the information and the reformat it appropriately? This would solve the issue with adding yet another flag.

The Makefile syntax is standard and read by Make and Ninja for ages. There is no way they will switch to any new format. This patch is to help dmd to integrate better with existing tools, inventing a new format and expecting everybody else to adapt is unlikely (convincing others to add support for D gets *much* easier if the amount of work that needs to be done for that is also less).
Plus, I think dmd already generates a depfile in it's own custom format with the right flags.

jacob-carlborg:
> There is no way they will switch to any new format

No, but there could be a tool in between DMD and Make that extracts the necessary information from the JSON file and outputs a Make style dependency file.

ZombineDev:
Yes that's what I've meant - instead putting support for N different tools in dmd, support only json and have a tool convert from json to make, ninja, or any of the other hundreds of build tools.

ZombineDev:
That's not to say I don't want upstream support for something that will make life easier for many people using such build systems. On the contrary - I'm just more concerned with managing technical dept.
IMO it would be great to have a lean wrapper around dmd (ala rdmd) to https://github.com/dlang/tools that reformats information from the json output. Then adding support for yet another build system won't be a point of contention, as this would be the sole purpose of that tool.

ZombineDev:
There too much platform specific build logic in dmd already. IMO the compiler should simply transform source to binary and leave details like linking to other build system oriented tools.

JinShil:
> There too much platform specific build logic in dmd already. IMO the compiler should simply transform source to binary and leave details like linking to other build system oriented tools.

Exactly what I was thinking.  I've been meaning to move [all that hard-coded linking logic in DMD](https://github.com/dlang/dmd/blob/master/src/dmd/link.d) into the conf file for at least a year now.

> IMO it would be great to add a lean wrapper around dmd (ala rdmd) to https://github.com/dlang/tools that reformats information from the json output.

Yes, We already have several examples of that in the D ecosystem:
* [msvc-dmd](https://github.com/dlang/dmd/blob/master/src/vcbuild/msvc-dmc.d) - DMC-like wrapper around MSVC
* [msvc-lib](https://github.com/dlang/dmd/blob/master/src/vcbuild/msvc-lib.d) - DM Link-like wrapper around Microsoft's link.exe
* [rdmd](https://github.com/dlang/tools/blob/master/rdmd.d) - Needs no explanation.  Killer tool!
* [gdmd](https://github.com/D-Programming-GDC/GDMD) - DMD-like wrapper around GDC
* [ldmd](https://github.com/ldc-developers/ldc/blob/0c8c0e6a1821cdf3d1cf7ea0a378b273760aeefe/driver/ldmd.cpp) - DMD-like wrapper around LDC.

Such a driver could be added to https://github.com/dlang/tools or even added to dub (as an aside to emphasize my philosophy, I think Phobos should be moved to dub, too).

I suggest we always consider the [Unix Philosophy](https://en.wikipedia.org/wiki/Unix_philosophy) when designing our command-line interface and its tools.  I don't think DMD currently adheres well to that philosophy, unfortunately.

WalterBright:
> I don't think DMD currently adheres well to that philosophy, unfortunately.

Indeed it doesn't, which is one reason why it is so much faster. Reducing all its data structures to a stream and them reconstructing them in the next tool is very slow. I've many times considered building a linker into dmd so it could transform a D source file directly into an executable.

src/dmd/mars.d
```diff
@@ -2160,6 +2249,32 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 }
                 params.moduleDeps = new OutBuffer();
             }
+            else if (p[1] == 'M')
+            {
+                global.params.doMakeDeps = true;
```
WalterBright: note use of `global.` here, which is absent on line 2250.

jacob-carlborg:
> I've many times considered building a linker into dmd so it could transform a D source file directly into an executable.

That would be very nice, especially if it's a cross-linker, this would make cross-compiling that much easier. Also support for multiple formats on Windows would be nice, that is, linking both COFF and OMF in the same link step.

src/dmd/utils.d
```diff
@@ -132,3 +132,55 @@ extern (C++) void escapePath(OutBuffer* buf, const(char)* fname)
...
+
+
+/**
+ * Takes a path, and escapes '(', ')', backslashes and whitespace
+ *
+ * Params:
+ *   buf = Buffer to write the escaped path to
+ *   fname = Path to escape
+ */
+void escapePathSpaces(OutBuffer* buf, const(char)* fname)
```
WalterBright: Why is this not used for any other filename arguments?

jacob-carlborg:
> rdmd - Needs no explanation. Killer tool!

Less so now with the recent introduction of the `-i` flag and the already existing `-run` flag.

src/dmd/mars.d
```diff
@@ -869,20 +869,109 @@ private int tryMain(size_t argc, const(char)** argv)
...
+                printf("%.*s", cast(int)ob.offset, ob.data);
+        }
+        if (global.params.doMakeDeps)
+        {
+            // Recursive dependencies scanner, fills string tables with module source paths and string-imported files paths
+            void recursivelyAddModuleDependencies(Module m, StringTable* moduleImportedPaths, StringTable* stringImportedPaths)
+            {
+                const(char)* mpath = m.srcfile.toChars();
+                // Skip modules already in the table, add self
+                if(moduleImportedPaths.insert(mpath, strlen(mpath), null) is null)
```
WalterBright: strlen must die.

JinShil:
> Less so now with the recent introduction of the -i flag..

... which is a feature I (naive as I was) very much regret now that I've grown a little in wisdom.

jacob-carlborg:
> ... which is a feature I (naive as I was) very much regret now that I've grown a little in wisdom.

It goes agains the Unix philosophy but it will build faster compared to RDMD because the compiler doesn't need to be invoked twice.

src/dmd/mars.d
```diff
@@ -869,20 +869,109 @@ private int tryMain(size_t argc, const(char)** argv)
...
+                        continue;
+                    // Recurse
+                    recursivelyAddModuleDependencies(impmod, moduleImportedPaths, stringImportedPaths);
+                }
+            }
+
+            __gshared static OutBuffer mdout;
+            bool mdSingleFile = global.params.makeDepsFile !is null;
+            File* mdOutput;
+            if(mdSingleFile)
```
WalterBright: Please consistently use a space between `if` and `(`.

JinShil:
> ...it will build faster compared to RDMD because the compiler doesn't need to be invoked twice

That can be dealt with in other ways.

src/dmd/mars.d
JinShil: Agreed, but what's the alternative when you have a `char*`?

ibuclaw:
> Indeed, but the performance issues could be dealt with in other ways.

I feel this conversation has derailed somewhat from the actual PR merits/cons, shall we take this discussion to forum? :-)

src/dmd/mars.d
ibuclaw: I thought we already have a property that returns a D string in Identifier?

jacob-carlborg: No, not from `dmd.root.filename.FileName` (which is where `mpath` originates), as far as I can see.

Ideally `dmd.root.filename.FileName` should be updated to store the length as well and not only the pointer.

ximion:
@ZombineDev 
> [...]
> Yes that's what I've meant - instead putting support for N different tools in dmd, support only json and have a tool convert from json to make, ninja, or any of the other hundreds of build tools.
> [...]

The thing is though, since clang, gcc, gdc etc. all produce Makefile-style depfiles, there is no need to adding yet another format whose only purpose is to be transformed into the format that everybode else is actually using. Originally, the Make-style depfiles were - of course - to integrate with `make` (and any buildsystem generating makefiles), but then Ninja appeared as an alternative to make and it just adopted the same file format because it was just so ubiquitous. Due to that, even more compilers use it now.

Having dmd, and by extension ldc, support this out-of-the-box would mean we wouldn't have to write yet another wrapper for dmd just for this single purpose and wouldn't have to invent yet another new fileformat, but could integrate dmd with existing infrastructure immediately.

(Fun fact, I was actually transforming dmd's existing output before, but that actually adds a quite significant performance hit to any build, especially if you compile a lot of units. So, I actually do agree with Walter on the "dmd should just do the thing directly" statement. However, a libdmd would be awesome to have as well ^^)

src/dmd/mars.d
jacob-carlborg: If  you don't what to completely reimplement `FileName`, perhaps it's acceptable to add a `toString` method to `FileName` that encapsulates `strlen`. This way we at least doesn't increase the surface of where `char*` is used.

src/dmd/utils.d
kubasz: All the other places where `escapePath` was used are in either human-readable or paren-delimited custom format contexts, so there was no need to use this before (as far as I know).

src/dmd/mars.d
kubasz: I followed Jacob's suggestion with the `toString` method, unfortunately there still remains a use of `strlen` a couple of lines below, because it indexes into a `Strings` member of the `Module` class, which uses `const(char)*` as it's exposed in the C header file, please correct me if I'm wrong and I could change that into an array of D strings.

src/dmd/globals.d
```diff
@@ -208,6 +208,11 @@ struct Param
     const(char)* moduleDepsFile;        // filename for deps output
     OutBuffer* moduleDeps;              // contents to be written to deps file
 
+    bool doMakeDeps;                    // generate make-style dependencies file
+    const(char)* makeDepsFile;          // filename for make-style dependencies output
+    const(char)* makeDepsDir;           // directory for make-style dependencies output
+    const(char)* makeDepsTarget;        // the make target name
```
jacob-carlborg: All new public symbols should have Ddoc comments.

jacob-carlborg: Can these `const(char)*` fields be made into D strings, i.e. `const(char)[]`? On the C++ side an ABI compatible struct can be used.

jacob-carlborg (changes requested):

src/dmd/dmodule.d
```diff
@@ -381,6 +381,7 @@ extern (C++) final class Module : Package
     Dsymbols* decldefs;         // top level declarations for this Module
 
     Modules aimports;           // all imported modules
+    Strings astringImports;     // all string-imported files
```
jacob-carlborg: Perhaps this can be an array of `const(char)[]` and an ABI compatible struct on the C++ side.

src/dmd/mars.d
```diff
@@ -848,21 +848,107 @@ private int tryMain(size_t argc, const(char)** argv)
...
+                printf("%.*s", cast(int)ob.offset, ob.data);
+        }
+        if (global.params.doMakeDeps)
+        {
+            // Recursive dependencies scanner, fills string tables with module source paths and string-imported files paths
+            void recursivelyAddModuleDependencies(Module m, StringTable* moduleImportedPaths, StringTable* stringImportedPaths)
+            {
+                auto mpath = m.srcfile.toString();
+                // Skip modules already in the table, add self
+                if (moduleImportedPaths.insert(mpath.ptr, mpath.length, null) is null)
```
jacob-carlborg: Please add an overload to `insert` that takes `const(char)[]` instead of a pointer and length.

src/dmd/root/filename.d
```diff
@@ -765,6 +765,11 @@ nothrow:
     {
         return str;
     }
+
+    extern (D) const(char)[] toString() const pure
```
jacob-carlborg: Can be made `nothrow`.

src/dmd/root/file.d
```diff
@@ -66,11 +66,16 @@ nothrow:
...
-    extern (C++) const(char)* toChars() pure
+    extern (C++) const(char)* toChars() pure const
     {
         return name.toChars();
     }
 
+    extern (D) const(char)[] toString() pure const
+    {
+        return name.toString();
+    }
```
jacob-carlborg: Can be made `nothrow`.

src/dmd/root/file.d
```diff
@@ -66,11 +66,16 @@ nothrow:
         }
     }
 
-    extern (C++) const(char)* toChars() pure
+    extern (C++) const(char)* toChars() pure const
     {
         return name.toChars();
```
jacob-carlborg: Can be made `nothrow`, I assume.

src/dmd/mars.d
```diff
@@ -848,21 +848,107 @@ private int tryMain(size_t argc, const(char)** argv)
...
+                // Each module needs to walk its entire dependency tree for the output to be correct
+                StringTable moduleImportedPaths, stringImportedPaths;
+                moduleImportedPaths._init();
+                stringImportedPaths._init();
+                recursivelyAddModuleDependencies(m, &moduleImportedPaths, &stringImportedPaths);
+                if (global.params.makeDepsTarget is null)
+                {
+                    // no actual file is created, uses just the path
+                    File* objfile = m.setOutfile(global.params.objname, global.params.objdir, m.arg, global.obj_ext);
+                    escapePathSpaces(&mdout, objfile.toChars());
```
jacob-carlborg: Please used the newly added `toString` method instead.

src/dmd/globals.d
Geod24: As much as I'd like this to happen, I don't think it's fair to ask a contributor to break the current status quo.
Such a change should be done on its own.

src/dmd/dmodule.d
Geod24: Same comment about the status quo. I'd like to replace the `Array` aliases as much as the next guy, but I don't think we should put that kind of weight on this PR.

atilaneves:
This PR would also close the [issue I filed](https://issues.dlang.org/show_bug.cgi?id=14188).

On the whole "why makefile dependencies?" front, I'd like to add this: this is _required_ for ninja to work. Reggae only supports ninja by wrapping dmd and offering the necessary functionality.

src/dmd/dmodule.d
kubasz: I also think that this is not really in the scope of this PR, and I think a "standard" way to do it (perhaps with predefined structs for most common array typs) should be defined first and a refactor PR could be done with that in place.

kubasz:
I think it is too late now to change the behavior of the -deps switch, quite a few tools seem to rely on its format :(

src/dmd/root/filename.d
```diff
@@ -761,10 +761,66 @@ nothrow:
...
 
-    extern (C++) const(char)* toChars() const pure
+    extern (C++) const(char)* toChars() const pure nothrow
     {
         return str;
     }
+
+    extern (D) const(char)[] toString() const pure nothrow
+    {
+        return str[0 .. strlen(str)];
```
WalterBright: str can be null

WalterBright: see https://github.com/dlang/dmd/pull/8550, which should be a separate PR anyway.

JinShil:
So if this is merged, we'll have `-deps`, `-X`/`-Xf`, and then `-M`/`-Mffile`/`-Mddirectory`/`-Mttarget` all doing similar things in a different way?  Can this herd of cats be reigned in somehow?

src/dmd/dmodule.d
jacob-carlborg: Then it will never happen and at the same time we’re adding more code that needs to change later 😞

andralex:
The right design is to output the information in a simple format and leave it to tooling to translate to others. Sadly, one little uninspired idea after another, we're here.

WalterBright:
And so we arrive at the nightmarish blizzard of switches I so wanted to avoid :-(

ximion:
Well, at some well-announced point in the future, you could in theory clean up the compiler flags and adjust them to something that you like.
Personally, I always prefer when a compiler copies the gcc CLI - not because it's particularly great, but because it is very complete and familiar, so achieving the same result with the same flags on different compilers is easy.
Anything goes, though.
The one thing I really want to avoid is having some extra binary to convert depfiles into multiple formats - that's not performant at all and will slow down compilation a lot due to the sheer number of process spawns.

src/dmd/mars.d
```diff
@@ -848,21 +848,107 @@ private int tryMain(size_t argc, const(char)** argv)
         fatal();
 
     // inlineScan incrementally run semantic3 of each expanded functions.
-    // So deps file generation should be moved after the inlinig stage.
-    if (global.params.moduleDeps)
+    // So deps file generation should be moved after the inlining stage.
+    if (global.params.moduleDeps || global.params.doMakeDeps)
```
rainers: You probably don't want to reuse `-deps` functionality as it is no longer useful for `make`. Analysing code via `semantic3OnDependencies` also reports local imports that are not used when *compiling* the code without the `-deps` option. It can also slow down compilation considerably. See discussion in https://github.com/dlang/dmd/pull/6748.

kubasz: That was insightful, thank you. I've moved the `semantic3OnDependencies` call to be done only for the old `-deps` switch.

GoaLitiuM:
What happened to this PR? Can we get this rebased and back to track again? I really wanted to see this go through to improve the building process with ninja.

MartinNowak:
It's not in a particularly good shape, adds lots of extra switches with partly redundant functionality (https://github.com/dlang/dmd/pull/6961#issuecomment-411655339).
More importantly recursive import dependencies are **not** what you want for make et.al.
Those tools know well how to stich together shallow dependencies themselves. Also recursive dependencies can only be reliably generated by the compiler when actually compiling a dependency (e.g. local imports in templates).

Geod24:
Another implementation, #12011, has been merged into master.

ximion:
Excellent, many thanks to everyone involved! This helps a lot for project with mixed languages and D with buildsystems using ninja and make :-)

PR #6963, opened 2017-07-03
Fix Issue 17590 - Unnecessary GC alloc on returning static local struct

JohanEngelen:
(this is a little in-between enhancement and bug-fix, I feel)

### DIFF
--- a/src/ddmd/func.d
+++ b/src/ddmd/func.d
@@ -3376,7 +3376,7 @@ extern (C++) class FuncDeclaration : Declaration
          * 3) has a parent that escapes
          * 4) calls another nested function that needs a closure
          * -or-
-         * 5) this function returns a local struct/class
+         * 5) this function returns a local non-static struct/class
          *
          * Note that since a non-virtual function can be called by
          * a virtual one, if that non-virtual function accesses a closure
@@ -3451,13 +3451,20 @@ extern (C++) class FuncDeclaration : Declaration
             {
                 Dsymbol st = tret.toDsymbol(null);
                 //printf("\t\treturning class/struct %s\n", tret.toChars());
-                for (Dsymbol s = st.parent; s; s = s.parent)
+
+                auto ad = st.isAggregateDeclaration();
+                assert(ad !is null);
+                // A closure is only needed when returning a _non-static_ struct/class.
+                if ((ad.storage_class & STCstatic) == 0)
                 {
-                    //printf("\t\t\tparent = %s %s\n", s.kind(), s.toChars());
-                    if (s == this)
+                    for (Dsymbol s = st.parent; s; s = s.parent)
                     {
-                        //printf("\t\treturning local %s\n", st.toChars());
-                        goto Lyes;
+                        //printf("\t\t\tparent = %s %s\n", s.kind(), s.toChars());
+                        if (s == this)
+                        {
+                            //printf("\t\treturning local %s\n", st.toChars());
+                            goto Lyes;
+                        }
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


src/ddmd/func.d
```diff
@@ -3451,13 +3451,20 @@ extern (C++) class FuncDeclaration : Declaration
...
-                    if (s == this)
+                    for (Dsymbol s = st.parent; s; s = s.parent)
                     {
-                        //printf("\t\treturning local %s\n", st.toChars());
-                        goto Lyes;
+                        //printf("\t\t\tparent = %s %s\n", s.kind(), s.toChars());
+                        if (s == this)
+                        {
+                            //printf("\t\treturning local %s\n", st.toChars());
+                            goto Lyes;
```
WalterBright: Looks like this is crucial to be covered by a test case.

JohanEngelen: What exactly do you want tested?
The codepath is already taken e.g. by runable/closure.d test 9685.

WalterBright: The coverage tester shows it as red, meaning it is untested.

JohanEngelen: It turns out it is hard to find a test case where this case is really needed. E.g.
```d
auto goo(int a, bool b) {
    int x = 2;

    struct SInside {
        int getX() { return x; }
    }

    SInside res;
    lazyfun(a);
    return res;
}
```
this will not trigger the codepath, because `requiresClosure` is already set at the start of `needsClosure()` resulting in an early return. Removing `getX` does trigger the codepath, but in that case we still actually don't need a closure (or is it strictly required by the spec?).

JohanEngelen: (extra test added)

WalterBright: It isn't required by the spec. It's possible that whole construct isn't necessary. We can find out by disabling it and seeing what breaks!

JohanEngelen: To be frank, I do not have enough trust in DMD's testing to go that route. I am dealing with too many regressions already to risk adding one more. And I guess from the coverage tester we can already deduct that removing case 5 will not result in any changes for the code being tested.
Let's just fix the case that we are certain about (static structs/classes). I can remove the extra test (because it is testing that we need a closure for something that doesn't need a closure), add a TODO comment, add an enhancement bug issue, and perhaps a forum post. So that people can review things later and can come up with a way to break things before it actually breaks things. How about that?

JohanEngelen: I'm going to try removing that whole case and test it on Weka's codebase. It could be a good optimization to get rid of the closure alloc, and would allow more `@nogc`. On the other hand, it is perhaps better when people are pointed (by `@nogc` compile error) towards adding `static` to a struct that can be static...

JohanEngelen:
@WalterBright I tested this (with the whole case 5 construct removed) on Weka's codebase, and nothing bad popped up in their extensive testing.

JohanEngelen:
This is now live @ weka.

JohanEngelen:
@WalterBright ping

WalterBright (approved):

JohanEngelen:
Is this stuck on the coverage issue?
If so, I'd like someone else to take a look at that because it would take me a very long time to figure out how to trigger that codepath.

MetaLang:
Looks like @WalterBright just forgot to add the automerge tag.

JohanEngelen:
Explanation about the coverage decrease: it is in some distant code path. Because of this change, somewhere way down the line things are ever slightly so different that the distant code path is no longer taken.

wilzbach:
> Looks like @WalterBright just forgot to add the automerge tag.

I think so as well :)
@JohanEngelen Don't worry about the coverage - it's just a help for reviewers, not a requirement.

JohanEngelen:
Thanks

PR #6964, opened 2017-07-03
fix Issue 17492 - [REG 2.066] [ICE] AssertError@ddmd/dclass.d(1007): …

WalterBright:
…Assertion failure

### DIFF
--- a/src/ddmd/dclass.d
+++ b/src/ddmd/dclass.d
@@ -967,6 +967,18 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
 
         sc2.pop();
 
+        if (type.ty == Tclass && (cast(TypeClass)type).sym != this)
+        {
+            // https://issues.dlang.org/show_bug.cgi?id=17492
+            ClassDeclaration cd = (cast(TypeClass)type).sym;
+            version (none)
+            {
+                printf("this = %p %s\n", this, this.toPrettyChars());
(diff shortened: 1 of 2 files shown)


src/ddmd/dclass.d
```diff
@@ -967,6 +967,18 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
...
+        if (type.ty == Tclass && (cast(TypeClass)type).sym != this)
+        {
+            // https://issues.dlang.org/show_bug.cgi?id=17492
+            ClassDeclaration cd = (cast(TypeClass)type).sym;
+            version (none)
+            {
+                printf("this = %p %s\n", this, this.toPrettyChars());
+                printf("type = %d sym = %p, %s\n", type.ty, cd, cd.toPrettyChars());
+            }
+            error("already exists at %s. Perhaps in another function with the same name?", cd.loc.toChars());
```
ibuclaw: Maybe split this into an error an errorSupplemental?
```
error("already exists, perhaps in another function with the same name.");
errorSupplemental(cd.loc, "previous declaration found here.");
```
Rather than formating the location into the error string.

Temtaime:
Can one create two functions with the same prototype without an error ? Which purpose ?

WalterBright:
> Can one create two functions with the same prototype without an error ?

Yes. This is an old issue.

MartinNowak:
> Yes. This is an old issue.

Also responsible for duplicate manglings https://issues.dlang.org/show_bug.cgi?id=17352.

PR #6966, opened 2017-07-04
[WIP] Create Semantic Library

RazvanN7:
This PR will ultimately build the semantic library. In this commit I've done some refactorings in ASTBase so that the code used there will also be used in ASTSemantic without duplication. This commit implements the following:

- move the code from the constructors in ASTBase in mixin templates so that ASTSemantic uses the same code. The nodes in ASTBase, with few exceptions, have the same constructors as nodes in ASTSemantic. Inheriting the nodes in ASTBase is not a solution since that would make impossible the inheritance between nodes in ASTSemantic.

- move common global members like STCxxx, MODxxx etc. in a separate file called astattributes (this might not be the best name, suggestions are welcome).

- create basic ASTSemantic which, at the moment, does exactly what ASTBase does

- template the visitors with an ASTFamily in order to be able to use them with both ASTBase and ASTSemantic

Following steps:

- Move semantic functionality into ASTSemantic

- Implement semantic visitors : most likely the semantic visitors will inherit permissive, strict, transitive visitor and add visiting methods for AST nodes add in the semantic phase

- design a proper interface and AST matchers like clang does (https://clang.llvm.org/docs/LibASTMatchers.html)

==========================================================================

Later edit: 

Given that creating ASTSemantic in parallel with the compiler code will lead to massive duplication and a drastic change of the compiler code, I suggest the following plan:

The parser has already been separated from the rest of the compiler, but at the moment we have (1) ASTCodegen which basically represents the old compiler organization and (2) ASTBase which at the moment represents duplicated code from the point of view of ASTCodegen. In order to have the same code for the compiler and the library there are 2 options : either do what the initial description of this PR suggested (which will lead to massive duplication) or integrate the code that does parsing in the existing ASTCodegen classes. The latter is, in my opinion the way to go.

Implementation:

In order to integrate the parsing code in the existing classes there are 2 prerequisites: 
-> avoid code duplication
-> preserve parser/semantic independence

In order to avoid code duplication, the code from each class in ASTBase is put in a mixin template which will be instantiated for both ASTBase and the classes in ASTCodegen. That way the same code will be instantiated for both classes. This avoids code duplication and preserves independence between various AST families. All there needs to be done is to replace the code that does parsing in all ASTCodegen classes with calls to the mixin template associated with it. The second commit of this PR is a proof of concept which does exactly the modifications mentioned for the Dsymbol class. I didn't replace the visit method since that will create problems for the methods that currently override it, but once the other classes are modified, that shouldn't be a problem anymore.

I know that this might seem to complicate the code, but it actually simplifies it by making a clear separation between the code used in each phase. What is in the mixin template is information which is common for both families while the other information is completely useless for parsing.


jacob-carlborg:
@RazvanN7 I've been thinking a bit lately on the structure of the DMD source code. Instead of having different AST families, would it make sense to have a basic AST structure that is only data and no functionality. Then all the functionality could be moved to either visitors or subclasses. Does this make sense?

RazvanN7:
@jacob-carlborg Well, that is what ASTBase does. It has data and constructors (which are almost identical to those in ASTCodegen).

jacob-carlborg:
What I mean is that, currently, all AST classes are duplicated. For example `ClassDeclaration` exists both in `ddmd.astbase` and on in `ddmd.dclass`. My idea was that either:

A. `ddmd.dclass.ClassDeclaration` inherits from `dmd.astbase.ClassDeclaration` and adds its own fields and functionality

Or

B. Move all static and instance variables from `ddmd.dclass.ClassDeclaration`  to `dmd.astbase.ClassDeclaration` and move all functionality from `ddmd.dclass.ClassDeclaration` to visitors (or similar)

I guess it's also possible to do a combination of both A and B.

Then it's not necessary to duplicate the AST.

jacob-carlborg:
> Version A was my initial plan, but that is not possible since D does not support multiple inheritance.

Ah, yes. I didn't think of that.

> On the other hand, I would like to know what you think about my initial proposal in which the code in ASTBase is moved to a mixin template which can be instantiated in both ASTBase and ASTCodegen (since the code is the same). That way we avoid a lot of code duplication

It's true that a lot of code duplication is avoided. But I think it complicates the design a bit and there would still be two AST trees, as far as I understand.

Are the current implementation (what's in master) and the template mixin proposal implement like they are because it's simpler than doing B?

DmitryOlshansky:
> Version A was my initial plan, but that is not possible since D does not support multiple inheritance. Let me detail this a bit: The AST nodes have inheritance relationships between them, for example a ScopeDsymbol inherits a Dsymbol; if you want to implement version A, ASTCodegen.ScopeDsymbol has to inherit ASTCodegenDsymbol and ASTBase.ScopeDsymbol from ASTBase.

How about including ASTBase node inside of ASTCodegen and forwarding things with alias this ?

jacob-carlborg:
> How about including ASTBase node inside of ASTCodegen and forwarding things with alias this ?

That would still require two AST trees. Which I'm trying to figure out if we actually need.

RazvanN7:
@jacob-carlborg  The initial thought was to separate parsing from semantic and be able to build a parser independently. In order to achieve that ASTBase was created. When I got to separating semantic I noticed that a natural (inheritance) relationship cannot be established between the 2 AST families so I tried to reduce the duplicated code as much as possible.

Now to respond to your question : no, what you proposed is actually on my agenda. the first steps are going to be reduce the codebase complexity by moving code outside the ASTCodegen nodes (static methods, member fields or functions that are not supposed to be there) so that the nodes contain only what they absolutely have to contain. After that, I will probably implement visitors for all the semantic routines.

Hope this answers your question.

jacob-carlborg:
> Hope this answers your question.

Yes, thanks.

PR #6968, opened 2017-07-05
enable eqeqranges()

WalterBright:
Not sure why this was commented out. I guess I'll find out!

### DIFF
--- a/src/ddmd/backend/gother.c
+++ b/src/ddmd/backend/gother.c
@@ -103,7 +103,7 @@ void constprop()
 {
     rd_compute();
     intranges();                // compute integer ranges
-    //eqeqranges();               // see if we can eliminate some relationals
+    eqeqranges();               // see if we can eliminate some relationals
     elemdatafree(&eqeqlist);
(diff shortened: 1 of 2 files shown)


MartinNowak:
> Not sure why this was commented out. I guess I'll find out!

Maybe it works, maybe we'll find the same bug that lead to disabling this ~10 years ago. In the latter case it would have been better to leave it dormant. The backend is so far behind with performance that such minor optimizations hardly make any difference and we'd better use our time for more useful stuff :).

WalterBright:
With this enabled, it pass 100% of the tests both here and in my DMC++ suite. It likely got commented out sometime in the distant past when I was looking for another bug, and forgot to uncomment it.

Being afraid to do more optimizations despite passing everything out of fear something somewhere might break is to simply halt progress. We have to have some faith in our tests, otherwise there is no point in having them.

If it is found to break something, at least then we have a test case and a bug that can be fixed, and if it can't be fixed quickly at least it's a one line temporary correction (comment the line out) and there'll be a known reason.

PR #6969, opened 2017-07-05
fix Issue 17380 - Compiler segfaults on undefined symbol

WalterBright:
Another infinite recursion problem.

### DIFF
--- a/src/ddmd/mtype.d
+++ b/src/ddmd/mtype.d
@@ -2717,6 +2717,23 @@ extern (C++) abstract class Type : RootObject
      */
     final Expression noMember(Scope* sc, Expression e, Identifier ident, int flag)
     {
+        //printf("Type.noMember(e: %s ident: %s flag: %d)\n", e.toChars(), ident.toChars(), flag);
+
+        static __gshared int nest;      // https://issues.dlang.org/show_bug.cgi?id=17380
+
+        static Expression returnExp(Expression e)
+        {
+            --nest;
+            return e;
+        }
+
+        if (++nest > 500)
+        {
+            .error(e.loc, "cannot resolve identifier `%`", ident.toChars());
+            return returnExp(flag & 1 ? null : new ErrorExp());
+        }
+
+
         assert(ty == Tstruct || ty == Tclass);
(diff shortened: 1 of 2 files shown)


src/ddmd/mtype.d
```diff
@@ -2717,6 +2717,23 @@ extern (C++) abstract class Type : RootObject
      */
     final Expression noMember(Scope* sc, Expression e, Identifier ident, int flag)
     {
+        //printf("Type.noMember(e: %s ident: %s flag: %d)\n", e.toChars(), ident.toChars(), flag);
+
+        static __gshared int nest;
+
+        static Expression returnExp(Expression e)
```
UplinkCoder: any reason not to use `scope(exit)` ?

WalterBright: exception handlers are expensive

ibuclaw (approved):

src/ddmd/mtype.d
```diff
@@ -2717,6 +2717,23 @@ extern (C++) abstract class Type : RootObject
      */
     final Expression noMember(Scope* sc, Expression e, Identifier ident, int flag)
     {
+        //printf("Type.noMember(e: %s ident: %s flag: %d)\n", e.toChars(), ident.toChars(), flag);
+
+        static __gshared int nest;      // https://issues.dlang.org/show_bug.cgi?id=17380
+
+        static Expression returnExp(Expression e)
+        {
+            --nest;
```
MartinNowak: What better use-case could there be for `scope (exit)`? This function should also be outside any hot code path, if performance was a criteria to not use `scope (exit)` (though we should really fix any such issues).

src/ddmd/mtype.d
MartinNowak: > exception handlers are expensive

No, dmd's backend is bad at emitting EH code.

WalterBright: I suggest comparing the EH output with gdc/ldc. (Chandler Carruth told me that EH essentially disables the llvm optimize.)

The problem is in the front end. If the code is nothrow, EH should not be used at all. But the front end hasn't evolved that far, and even if it had, none of the dmd front end code is marked nothrow.

PR #6970, opened 2017-07-05
ddoc: avoid generating hardcoded characters

andralex:
ddoc generates a few spurious newlines and such that are annoying to deal with if precise output is to be generated. This PR moves the hardcoded characters into the macro definitions, thus improving flexibility.

This will fail at building documentation because of small differences. A follow-up PR will add the extra characters to the macros in dlang.org.


andralex:
Superseded by https://github.com/dlang/dmd/pull/6972/

PR #6972, opened 2017-07-05
Ddoc no hardcoded chars redux

andralex:
(Redux of https://github.com/dlang/dmd/pull/6970)

ddoc generates a few spurious newlines and such that are annoying to deal with if precise output is to be generated. This PR moves the hardcoded characters into the macro definitions, thus improving flexibility.

This will fail at building documentation because of small differences. A follow-up PR will add the extra characters to the macros in dlang.org.

### DIFF
--- a/res/default_ddoc_theme.ddoc
+++ b/res/default_ddoc_theme.ddoc
@@ -2,6 +2,8 @@ LPAREN = (
 RPAREN = )
 BACKTICK = `
 DOLLAR = $
+LF =
+$(LF)
 
 ESCAPES =
   /</&lt;/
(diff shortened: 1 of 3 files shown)


andralex:
Had to change the test reference output because DDOC_EXAMPLES is emitted from two distinct sites, once with the LF and once without.

res/default_ddoc_theme.ddoc
```diff
@@ -2,6 +2,8 @@ LPAREN = (
 RPAREN = )
 BACKTICK = `
 DOLLAR = $
+LF =
+$(LF)
```
MartinNowak: What's that supposed to do?

WalterBright:
> ddoc generates a few spurious newlines and such that are annoying to deal with if precise output is to be generated.

When would precise newlines be needed?

res/default_ddoc_theme.ddoc
andralex: Glad you asked. A macro present with no arguments in its own definition expands to nothing, so this makes sure there's an empty line inside the macro. Without that, newlines are removed so the macro would not contain a newline.

Similarly, one space is surprisingly subtle to define: `SPACE = $(SPACE) $(SPACE)`.

andralex:
>When would precise newlines be needed?

I generated the bigo paper (http://erdani.com/d/bigo.html) in both html and pdf using ddoc. Also, the median paper http://erdani.com/d/bigo.html also uses ddoc. I noticed certain issues (extra space, errors in LaTeX macro expansions) caused by artifacts introduced by ddoc. Not sure whether they're exactly the ones addressed by this PR, but the immutable rule is clear: ddoc should NEVER produce characters. It should only produce macros.

andralex:
Forgot to mention - I am considering writing my next book with ddo(c|x). That should provide a good deal of experience and reveal what's needed for a large project.

CyberShadow:
I don't see any visual difference in the few rendered HTML pages that I clicked around on, so this seems good to me.

Getting those newlines back is still a little desirable, to make the HTML (and especially HTML diffs) more readable.

andralex:
@CyberShadow https://github.com/dlang/dlang.org/pull/1809

andralex:
Is the jenkins fail related? I find it difficult to read through the dub build log.

CyberShadow:
It's not.

andralex:
@CyberShadow then approval is just a click away :)

CyberShadow (approved):
Alright. Generally I leave dmd reviewers to someone else as I have little experience with its codebase, but in this case the code changes are trivial.

PR #6973, opened 2017-07-05
Fixing b12685 - foreach index type to span the whole range of the type

gchatelet:


gchatelet:
Original bug for the record
https://issues.dlang.org/show_bug.cgi?id=12685

wilzbach:
> Original bug for the record
> https://issues.dlang.org/show_bug.cgi?id=12685

@gchatelet As the Dlang-Bot mentioned, it would be great if you could reference the issue in your commit message. For details see:

https://github.com/dlang-bots/dlang-bot#automated-references

Please let us know if you need help.

gchatelet:
Please do not merge this PR.
It occurred to me that the current code is indeed correct.
```
ubyte[256] data;
foreach (ubyte i, x; data) {}
```
is lowered as
```
ubyte[256] data;
for (ubyte i = 0 ; i < 256 ; ++i) {
  ubyte x = data[i]
}
```
This loop would always be true because of `ubyte(255) + 1 = 0`.
So the code is correct but the error message is wrong.

Temtaime:
The code doesn't seem to be correct not the error message.
0..256 is a valid interval for foreach and means [0, 255]. With this fix it displays 0..255 which means [0, 254].

gchatelet:
@Temtaime please have a look at [this post](http://forum.dlang.org/post/tdptnbscskpspugaopsn@forum.dlang.org) for context.

I'm actually closing this PR - it's not the right fix.

PR #6974, opened 2017-07-06
account for noreturn expressions in Reaching Definitions

WalterBright:
Other flow analyses account for expressions that do not return, so Reaching Definitions should do it, too.

### DIFF
--- a/src/ddmd/backend/gflow.c
+++ b/src/ddmd/backend/gflow.c
@@ -324,18 +324,44 @@ STATIC void accumrd(vec_t GEN,vec_t KILL,elem *n)
                 rdelem(&Gl,&Kl,n->E1);
                 rdelem(&Gr,&Kr,n->E2);
 
-                /* GEN = (GEN - Kl) | Gl |      */
-                /*      (GEN - Kr) | Gr         */
-                /* KILL |= Kl & Kr              */
+                switch (el_noreturn(n->E1) * 2 | el_noreturn(n->E2))
+                {
+                    case 0: // E1 and E2 return
+                        /* GEN = (GEN - Kl) | Gl |
+                         *       (GEN - Kr) | Gr
+                         * KILL |= Kl & Kr
+                         */
+                        vec_orass(Gl,Gr);
+                        vec_sub(Gr,GEN,Kl);
+                        vec_orass(Gl,Gr);
+                        vec_sub(Gr,GEN,Kr);
+                        vec_or(GEN,Gl,Gr);
 
-                vec_orass(Gl,Gr);
-                vec_sub(Gr,GEN,Kl);
-                vec_orass(Gl,Gr);
-                vec_sub(Gr,GEN,Kr);
-                vec_or(GEN,Gl,Gr);
+                        vec_andass(Kl,Kr);
+                        vec_orass(KILL,Kl);
+                        break;
 
-                vec_andass(Kl,Kr);
-                vec_orass(KILL,Kl);
+                    case 1: // E1 returns
+                        /* GEN = (GEN - Kl) | Gl
+                         * KILL |= Kl
+                         */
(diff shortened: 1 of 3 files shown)


WalterBright:
Blocked by https://github.com/dlang/druntime/pull/1858

src/ddmd/backend/gflow.c
```diff
@@ -324,18 +324,44 @@ STATIC void accumrd(vec_t GEN,vec_t KILL,elem *n)
...
+                         * KILL |= Kl & Kr
+                         */
+                        vec_orass(Gl,Gr);
+                        vec_sub(Gr,GEN,Kl);
+                        vec_orass(Gl,Gr);
+                        vec_sub(Gr,GEN,Kr);
+                        vec_or(GEN,Gl,Gr);
+
+                        vec_andass(Kl,Kr);
+                        vec_orass(KILL,Kl);
```
MartinNowak: If you want to go for it (using `& ~` instead of `-` for clarity).
```d
(GEN & ~Kl) | (GEN & ~Kr) == (GEN & ~(Kl & Kr))
```
So you could do the `vec_andass(Kl,Kr);` that's needed anhow first, and then simplify the rest.
**Untested**
```d
vec_andass(Kl,Kr);
vec_orass(KILL,Kl);

vec_orass(Gl,Gr);
vec_sub(Gr,GEN,Kl); // (GEN - (Kl & Kr))
vec_or(GEN,Gl,Gr);
```

src/ddmd/backend/gloop.c
```diff
@@ -1005,15 +1005,34 @@ STATIC void markinvar(elem *n,vec_t rd)
...
                 break;
         case OPcolon:
         case OPcolon2:
                 tmp = vec_clone(rd);
-                markinvar(n->E1,rd);
-                markinvar(n->E2,tmp);
-                vec_orass(rd,tmp);              /* rd |= tmp            */
+                switch (el_noreturn(n->E1) * 2 | el_noreturn(n->E2))
+                {
+                    case 0:
```
MartinNowak: The ` // E1 and E2 return` comments are quite useful for readability of the switch, please use them here as well.

src/ddmd/backend/gother.c
```diff
@@ -199,17 +199,37 @@ STATIC void conpropwalk(elem *n,vec_t IN)
...
+                        conpropwalk(n->E1,L);
+                        conpropwalk(n->E2,IN);
+                        break;
+                    case 3:
+                        conpropwalk(n->E1,L);
+                        vec_copy(L,IN);
+                        conpropwalk(n->E2,L);
+                        break;
+                }
                 vec_free(L);
```
MartinNowak: Guess it's not possible to avoid touching `L` if it's not needed.

MartinNowak (approved):
Looks OKish, I don't understand much of what this does though. What does `accumrd` stand for anyhow?

WalterBright:
> I don't understand much of what this does though.

Modifies the data flow equations to take advantage of noreturn edges.

> What does accumrd stand for anyhow?

Accumulate Reaching Definition vectors

src/ddmd/backend/gflow.c
WalterBright: `-` is faster than `&~` to compute.

src/ddmd/backend/gloop.c
WalterBright: Actually, after this was pulled, I was going to refactor to use `el_returns()` instead of `el_noreturns()`, as the extra nots drove me batty.

src/ddmd/backend/gflow.c
MartinNowak: So you no longer trust Algebra ;)? Just adding a separate commit has the same effect btw, only that it avoids yet another TODO in master.

PR #6975, opened 2017-07-06
fix Issue 17612 - [REG2.063] Segmentation fault with bad object.d

WalterBright:


WalterBright:
Due to the age of this regression, I suggest it go into `master` instead.

PetarKirov (approved):
LGTM

PR #6977, opened 2017-07-08
properly configure backend for avx2

MartinNowak:
- add -mcpu=avx2
- detect avx2 for -mcpu=native
- init backend avx with 2, 1, or 0

spec https://github.com/dlang/dlang.org/pull/1811

### DIFF
--- a/src/ddmd/backend/backconfig.c
+++ b/src/ddmd/backend/backconfig.c
@@ -51,7 +51,7 @@ void out_config_init(
                         // 2: fake it with C symbolic debug info
         bool alwaysframe,       // always create standard function frame
         bool stackstomp,        // add stack stomping code
-        bool avx,               // use AVX instruction set
+        unsigned char avx,      // use AVX instruction set (0, 1, 2)
         bool betterC            // implement "Better C"
         )
 {
@@ -478,4 +478,3 @@ unsigned eve::sizeCheck() { return sizeof(eve); }
 
 // type.d
 unsigned TYPE::sizeCheck() { return sizeof(type); }
-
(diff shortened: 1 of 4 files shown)


src/ddmd/backend/backconfig.c
```diff
@@ -51,7 +51,7 @@ void out_config_init(
                         // 2: fake it with C symbolic debug info
         bool alwaysframe,       // always create standard function frame
         bool stackstomp,        // add stack stomping code
-        bool avx,               // use AVX instruction set
+        unsigned char avx,      // use AVX instruction set (0, 1, 2)
```
ZombineDev: enum?

PetarKirov (approved):

src/ddmd/backend/backconfig.c
MartinNowak: It's currently not an enum in the backend, but would be a nice follow-up.
https://github.com/dlang/dmd/blob/a336b69ca42ac535ea365fc4e70f84a7378d7e67/src/ddmd/backend/cdef.h#L908

WalterBright: enums in C tend to wind up as ints.

ZombineDev: I know, but can't we C++11's syntax  (same as in D): `enum name : type { enumerator = constexpr , enumerator = constexpr , ... }` 
(http://en.cppreference.com/w/cpp/language/enum)

WalterBright: No, because DMC++ is not C++11, and this is such a minor thing as to not justify requiring C++11.

ZombineDev: Alright, thanks for the clarification.

PR #6978, opened 2017-07-08
deprecate mixing AVX-256 usage and legacy SSE instructions

MartinNowak:
- Don't configure the backend with avx=0, while still using AVX-256
  vectors in order to avoid mixing of vex and legacy SSE instructions.
- See https://software.intel.com/en-us/articles/intel-avx-state-transitions-migrating-sse-code-to-avx
  for information on possible penalties. While the backend could carefully
  avoid those it's cleaner to always use vex instructions when AVX-256
  registers are in use.
- check for base type first in isVectorTypeSupported to not produce 
  false hints (e.g. __vector(void*[4]) is only supported with -mcpu=avx)

### DIFF
--- a/src/ddmd/mtype.d
+++ b/src/ddmd/mtype.d
@@ -4432,12 +4432,18 @@ extern (C++) final class TypeVector : Type
             error(loc, "SIMD vector types not supported on this platform");
             return terror;
         case 2:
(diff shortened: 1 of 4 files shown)


WalterBright:
Auto-merge toggled on

PR #6979, opened 2017-07-09
fix issue 15432: Win64: bad code offset in debug line number info

rainers:
pinhole optimization should not yield different results if run twice. 

The trouble was:
- the first pinholeopt() converted the epilog code "lea rsp,000000000[rbp]" to the byte version "lea rsp,0[rbp]".
- the size of the epilog was calculated to retsize=6
- a second pinholeopt() over the full function converted the code to "mov rsp,rbp", shrinking the size of the epilog to 5
- line number was emitted for offset "function end - retsize".

### DIFF
--- a/src/ddmd/backend/cod3.c
+++ b/src/ddmd/backend/cod3.c
@@ -5386,7 +5386,7 @@ void pinholeopt(code *c,block *b)
             }
 
             /* Look for LEA reg,[ireg], replace with MOV reg,ireg       */
-            else if (op == LEA)
+            if (op == LEA)
             {   rm = c->Irm & 7;
                 mod = c->Irm & modregrm(3,0,0);
                 if (mod == 0)
--- a/test/runnable/testpdb.d
+++ b/test/runnable/testpdb.d
@@ -1,4 +1,6 @@
 // REQUIRED_ARGS: -gf
+// PERMUTE_ARGS:
+
 import core.time;
 
 void main(string[] args)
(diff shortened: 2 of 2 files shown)


WalterBright (approved):

MartinNowak:
testpdb.d [fails on Win32_64](https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=2658168&isPull=true).
```
runnable\testpdb.d(77): Error: function testpdb.findSymbolLineNumbers (IDiaSession session, IDiaSymbol sym, ubyte[]* funcRange) is not callable using argument types (IDiaSession, IDiaSymbol, void[]*)
runnable\testpdb.d(819): Error: function testpdb.dumpLineNumbers.dumpLine (int lno, ubyte* beg, ubyte* end) is not callable using argument types (int, void*, ubyte*)
runnable\testpdb.d(822): Error: function testpdb.dumpLineNumbers.dumpLine (int lno, ubyte* beg, ubyte* end) is not callable using argument types (uint, ubyte*, void*)
```

rainers:
> testpdb.d fails on Win32_64

Oops. Fix amended.

rainers:
Label auto-merge has been removed by dlang-bot due to additional commits to fix the tests. Is it ok for the author to reenable it if the PR has been approved before? I guess noone looks at these PRs again due to the red crosses in the overview...

CyberShadow:
> Is it ok for the author to reenable it if the PR has been approved before?

Unless the new commits change the PR enough to warrant additional review, yes!

rainers:
> Unless the new commits change the PR enough to warrant additional review, yes!

I guess fixing the test doesn't qualifiy for that, so let's go ahead. Thanks.

PR #6982, opened 2017-07-10
Remove prelude assert from ctors & dtors

thewilsonator:
Removes prelude `assert(this !is null, "null this");` from constructors & destructors.
Trying to assign or access any members would crash anyway.

### DIFF
--- a/changelog/removePreludeAssert.dd
+++ b/changelog/removePreludeAssert.dd
@@ -1,0 +1,5 @@
+Removed prelude assert for constructors & destructors
+
+The compiler used to insert an `assert(this, "null this");` at the start of constructors & destructors.
+To trigger these asserts one needed to construct or destruct an aggregate at the null memory location.
+This would crash upon any data member access, which is required for a constructor or destructor to do anything useful.
--- a/src/ddmd/func.d
+++ b/src/ddmd/func.d
@@ -1399,7 +1399,7 @@ extern (C++) class FuncDeclaration : Declaration
             Statement fpreinv = null;
             if (addPreInvariant())
             {
-                Expression e = addInvariant(loc, sc, ad, vthis, isDtorDeclaration() !is null);
+                Expression e = addInvariant(loc, sc, ad, vthis);
                 if (e)
                     fpreinv = new ExpStatement(Loc(), e);
             }
@@ -1408,7 +1408,7 @@ extern (C++) class FuncDeclaration : Declaration
             Statement fpostinv = null;
             if (addPostInvariant())
             {
-                Expression e = addInvariant(loc, sc, ad, vthis, isCtorDeclaration() !is null);
+                Expression e = addInvariant(loc, sc, ad, vthis);
                 if (e)
                     fpostinv = new ExpStatement(Loc(), e);
             }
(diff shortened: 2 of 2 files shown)


andralex:
To reduce the size of each function, shouldn't the assert be massaged within the invariant if it exists?

thewilsonator:
Good idea! Not sure how to do this though, too tired. Will look into it tomorrow morning.

thewilsonator:
(The diff looks weird but thats just deindeting.)

After looking at the code more closely I realised that in oder to trigger this you had to construct or destruct an aggregate at the null memory location, which would crash if you were to access any members, which is required for a constructor or destructor to do anything useful other than logging (which would likely pick up the null pointer anyway).

This removes that assert so there is no need to move it to the invariant.

thewilsonator:
@WalterBright @yebblies This works now.

src/ddmd/func.d
```diff
@@ -1388,18 +1388,18 @@ extern (C++) class FuncDeclaration : Declaration
 
             // Precondition invariant
             Statement fpreinv = null;
-            if (addPreInvariant())
+            if (addPreInvariant() && isCtorDeclaration() is null)
```
WalterBright: `addPreInvariant()` is a virtual function. Add a definition for it in `CtorDeclaration` for it instead that returns `false`.

src/ddmd/func.d
```diff
@@ -1388,18 +1388,18 @@ extern (C++) class FuncDeclaration : Declaration
...
-                Expression e = addInvariant(loc, sc, ad, vthis, isDtorDeclaration() !is null);
+                Expression e = addInvariant(loc, sc, ad, vthis);
                 if (e)
                     fpreinv = new ExpStatement(Loc(), e);
             }
 
             // Postcondition invariant
             Statement fpostinv = null;
-            if (addPostInvariant())
+            if (addPostInvariant() && isDtorDeclaration() is null)
```
WalterBright: ditto

src/ddmd/func.d
thewilsonator: Actually its already there!

changelog/removePreludeAssert.dd
```diff
@@ -0,0 +1,3 @@
+Removed prelude assert for constructors & destructors
+
+The compiler used to insert an `assert(this,”null this”);` at the start of constructors & destructors. To trigger these asserts you had to construct or destruct an aggregate at the null memory location. This would crash if you were to access any members, which is required for a constructor or destructor to do anything useful other than logging (which would likely pick up the null pointer anyway).
```
andralex: s/this,/this, /
s/you had to construct/one needed to construct/
s/if you were to access any members/upon any data member access/
s/ other than logging (which would likely pick up the null pointer anyway)//

src/ddmd/func.d
```diff
@@ -3864,71 +3864,48 @@ extern (C++) class FuncDeclaration : Declaration
...
-            v = v.addressOf();
-        e = new StringExp(Loc(), cast(char*)"null this");
-        e = new AssertExp(loc, v, e);
-        e = e.semantic(sc);
+            inv.functionSemantic();
+        }
+
+        //e = new DsymbolExp(Loc(), inv);
+        //e = new CallExp(Loc(), e);
+        //e = e.semantic(sc2);
```
andralex: Remove commented-out code, or precede with a comment describing its usefulness.

andralex (commented):
lgtm

src/ddmd/func.d
thewilsonator: That was there before (it only shows up in the diff because it is de-indented). I can do that, but I'm not sure why it was there in the first place or what purpose it originally served (illustrative?).

andralex: no matter then

WalterBright (approved):

changelog/removePreludeAssert.dd
```diff
@@ -0,0 +1,5 @@
+Removed prelude assert for constructors & destructors
+
+The compiler used to insert an `assert(this, ”null this”);` at the start of constructors & destructors.
```
yebblies: Those don't appear to be ascii double quotes.

changelog/removePreludeAssert.dd
```diff
@@ -0,0 +1,5 @@
+Removed prelude assert for constructors & destructors
+
+The compiler used to insert an `assert(this, "null this");` at the start of constructors & destructors.
+To trigger these asserts one needed to construct or destruct an aggregate at the null memory location.
```
rainers: DAutotest complains about  trailing white space in this line.

ibuclaw:
I'm fine with this. @WalterBright has already approved.

dnadlinger (approved):

BBasile:
You should have added opCast too. [There's an interesting case showing why here](https://forum.dlang.org/post/ot20f2$2gub$1@digitalmars.com).

PR #6984, opened 2017-07-11
add loop unrolling optimization

WalterBright:
This does some modest unrolling - only once, and only if the loop bounds are fixed. The flow analysis was there, I just had to put it together.

### DIFF
--- a/src/ddmd/backend/gloop.c
+++ b/src/ddmd/backend/gloop.c
@@ -113,6 +113,7 @@ STATIC void countrefs(elem **pn , bool flag);
 STATIC int countrefs2(elem *e);
 STATIC void elimspec(loop *l);
 STATIC void elimspecwalk(elem **pn);
+static bool loopunroll(loop *l);
 
 static  bool addblk;                    /* if TRUE, then we added a block */
 
@@ -629,6 +630,7 @@ STATIC int looprotate(loop *l)
             list_append(&(head2->Bsucc),list_block(bl));
             list_append(&(list_block(bl)->Bpred),head2);
         }
+        cmes2("1Rotated loop %p\n", l);
         go.changes++;
         return TRUE;
     }
@@ -661,7 +663,7 @@ STATIC int looprotate(loop *l)
         b->Bnext = head->Bnext;
         head->Bnext = tail->Bnext;
         tail->Bnext = head;
-        cmes2( "Rotated loop %p\n", l);
+        cmes2("2Rotated loop %p\n", l);
         go.changes++;
     }
 Lret:
@@ -683,7 +685,6 @@ static bool doflow;             // TRUE if flow analysis has to be redone
 void loopopt()
 {
     list_t bl;
-    loop *l;
     loop *ln;
     vec_t rd;
     loop *startloop;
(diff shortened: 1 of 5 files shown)


WalterBright:
The code:
```
int foo() {
    int result = 0;
    for (int i = 0; i < 10; i++)
        result += i;
    return result;
}
```
formerly generated:
```
                xor     EAX,EAX
                xor     ECX,ECX
L4:             add     EAX,ECX
                inc     ECX
                cmp     ECX,0Ah
                jb      L4
                ret
```
and now:
```
                xor     EDX,EDX
                xor     ECX,ECX
L4:             lea     EAX,[01h][ECX*2]
                add     EDX,EAX
                add     ECX,2
                cmp     ECX,0Ah
                jb      L4
                mov     EAX,EDX
                ret
```

src/ddmd/backend/gloop.c
```diff
@@ -794,9 +795,31 @@ void loopopt()
...
+            ln = l->Lnext;
+            if (loopunroll(l))
+            {
+                compdfo();                      // compute depth-first order
+                blockinit();
+                compdom();
+                findloops(&startloop);          // recompute block info
+                doflow = TRUE;
+                if (ln)
+                {   ln = startloop;         // start over
```
andralex: since you have no code following braces in the code introduced elsewhere, this is jarring

andralex:
Well the code that does repeated additions should be ideally optimized to `result = 45;` i.e. merge repeated additions of constants.

WalterBright:
> Well the code that does repeated additions should be ideally optimized to result = 45; i.e. merge repeated additions of constants.

True, but that would be a different optimization.

src/ddmd/backend/gloop.c
WalterBright: It's copy-pasted from line 715.

UplinkCoder:
@WalterBright this optimization needs to controllable somehow.
because it can increase i-cache pressure.

src/ddmd/backend/gloop.c
```diff
@@ -794,9 +795,31 @@ void loopopt()
     /* one first, thus moving LIs out as far as possible.       */
 
     doflow = TRUE;                      /* do flow analysis             */
+
+    if (go.mfoptim & MFtime)
+    {
+        cmes("Starting loop unrolling\n");
+        for (loop *l = startloop; l; l = ln)
```
andralex: There's hardly a more awful symbol than `l`. Just look at the rendering in this github PR - is that condition `l` or `1`?

src/ddmd/backend/gloop.c
andralex: That doesn't make it any better!

src/ddmd/backend/gloop.c
WalterBright: In general you're quite right, and I considered it. But it's used consistently throughout `gloop.c`, its meaning is clear from the context, and being typed as a pointer means the compiler won't let you conflate it with `1`.

WalterBright:
> controllable somehow

Yes, which is why it only unrolls once and only for loops with less than 100 nodes.

src/ddmd/backend/gloop.c
MartinNowak: If we're at naming `nl` for next loop is more obvious than `ln` which typically abbreviates line.

MartinNowak:
Looking forward at the regressions ;), e.g. more usage of the `O(N^2)` optimization due to big unrolled loops.

WalterBright:
> big unrolled loops

Only small loops are unrolled - there's a size check.

PR #6989, opened 2017-07-12
Only consider an identifier followed by a colon a tag if a newline follows the colon

andralex:
This solves a number of annoying problems when ddoc confuses the use of ":" in running text with the intent to insert a tag. The matter can happen after text is reflown with no intent or knowledge that a tag will be generated.

Among the most annoying: if a link `http://...` happens to fall at the start of a line the `http` is converted to a tag. 

I suspect this will expose a few cases when people do want to insert tags but don't insert a newline.

### DIFF
--- a/changelog/ddoc-tag-deduction.dd
+++ b/changelog/ddoc-tag-deduction.dd
@@ -1,0 +1,38 @@
+Ddoc: tightening the conditions under which the sequence "identifier:" is deduced as a tag
+
+Previously, ddoc would deduce that an identifier occurring at the beginning of a
+line and followed by a colon (":") is a tag. The predefined tags (introducing
+the so-called $(HTTPS dlang.org/spec/ddoc.html#standard_srction, "Standard
+Sections")) are formatted with the `DDOC_SECTION` macro and have each one
+dedicated macro. All other detected tags are formatted with the `DDOC_SECTION_H`
+macro.
+
+A known problem with the previous tag detection mechanism was that it would be
+overactive, detecting tags where they were simply meant as regular text using
+":" as punctuation. Consider:
+
+----
+/**
+Lorem ipsum dolor sit amet, nam risus, sodales orci tempus sociis vel  lacinia,
+adipiscing vestibulum viverra ipsum, consequat vestibulum dolor quam in, ante
+accumsan amet tristique est sit. Depending on the input, the function frobnicate
+returns: (a) junk, (b) meh, or (c) something awesome.
+
+Returns:
+The result of frobnication.
+*/
+----
+
+The first occurence of "returns:" is in lowercase and part of regular text. It
+is sheer happenstance (e.g. a combination of text content and editor
+settings/author preference) whether the literal use of "returns:" occurs at the
+beginning of a line. The second occurrence of "Returns:", however, is a tag
+meant to highlight the homonym section of the documentation.
+
+Other examples of undesired tag detection include the use of hyperlinks
+("http:", "mailto:" etc) in text.
+
+To distinguish between textual use and the use as a tag, the new rule is: a tag
+is detected if (a) it is an identifier at the beginning of a line; (b) it is
+followed by a colon and a whitespace; and (c) it starts with an uppercase
+letter.
--- a/test/compilable/ddoc17697.d
+++ b/test/compilable/ddoc17697.d
@@ -11,6 +11,19 @@
  *    $(LINK2 http://www.food.com/test1, test1)
  */
 
+/**
+Also piggyback a few tests for https://github.com/dlang/dmd/pull/6989
+
+not_a_tag_because_it_does_not_start_with_uppercase:
+
+not_a_tag_because_no_whitespace_after_colon:x
+
+TagGalore: yes this is a tag
+
+MoreTag:
+yes the above is also a tag
+*/
+
 module test1;
 
 int a;
--- a/test/compilable/ddoc4899.d
+++ b/test/compilable/ddoc4899.d
@@ -4,8 +4,8 @@
 /*
 TEST_OUTPUT:
 ---
-compilable/ddoc4899.d(16): Warning: Ddoc: Stray '('. This may cause incorrect Ddoc output. Use $(LPAREN) instead for unpaired left parentheses.
-compilable/ddoc4899.d(16): Warning: Ddoc: Stray ')'. This may cause incorrect Ddoc output. Use $(RPAREN) instead for unpaired right parentheses.
+compilable/ddoc4899.d(18): Warning: Ddoc: Stray '('. This may cause incorrect Ddoc output. Use $(LPAREN) instead for unpaired left parentheses.
+compilable/ddoc4899.d(19): Warning: Ddoc: Stray ')'. This may cause incorrect Ddoc output. Use $(RPAREN) instead for unpaired right parentheses.
 ---
 */
 
@@ -14,6 +14,9 @@ compilable/ddoc4899.d(16): Warning: Ddoc: Stray ')'. This may cause incorrect Dd
         foo:)
 +/
 module d;
+
+/** ( */ int a;
+/** ) */ int b;
 
 void main()
 {
--- a/test/compilable/extra-files/ddoc17697.html
+++ b/test/compilable/extra-files/ddoc17697.html
@@ -473,6 +473,28 @@
     <a href="https://www.foob.com/test1">https://www.foob.com/test1</a>
     <a href="http://www.fooc.com/test1">http://www.fooc.com/test1</a>
     <a href="http://www.food.com/test1"><code class="code">test1</code></a>
+<br><br>
+Also piggyback a few tests for <a href="https://github.com/dlang/dmd/pull/6989">https://github.com/dlang/dmd/pull/6989</a>
+<br><br>
+not_a_tag_because_it_does_not_start_with_uppercase:
+<br><br>
+not_a_tag_because_no_whitespace_after_colon:x
+
+
+  </p>
+</div>
+<div class="ddoc_section">
+  <p class="para">
+    <span class="ddoc_section_h">TagGalore:</span>
+yes this is a tag
+
+
+  </p>
+</div>
+<div class="ddoc_section">
+  <p class="para">
+    <span class="ddoc_section_h">MoreTag:</span>
+yes the above is also a tag
   </p>
 </div>
 
--- a/src/ddmd/doc.d
+++ b/src/ddmd/doc.d
@@ -1643,13 +1643,10 @@ struct DocComment
                     const(char)* q = p + utfStride(p);
                     while (isIdTail(q))
                         q += utfStride(q);
-                    if (*q == ':' &&            // 'identifier:' ends it
-                                                // but not 'http://' or 'https://'
-                        !(q[1] == '/' && q[2] == '/' &&
-                          (q - p == 4 && Port.memicmp(p, "http".ptr, 4) == 0 ||
-                           q - p == 5 && Port.memicmp(p, "https".ptr, 5) == 0)
-                         )
-                       )
+
+                    // Detected tag ends it
+                    if (*q == ':' && isupper(*p)
+                            && (isspace(q[1]) || q[1] == 0))
                     {
                         idlen = q - p;
                         idstart = p;


UplinkCoder:
LGTM

andralex:
As I suspected that caused some problems so now the heuristic is to require identifier to start with an ASCII uppercase letter and have whitespace after the colon.

andralex:
There's a bug in std.datetime.systime's documentation exposed by this. I changed the lowercase labels in https://github.com/dlang/dmd/pull/6989.

CyberShadow:
I think requiring whitespace after the colon is completely reasonable.

I'm not entirely sure about being picky about case. It precludes starting section names with things like keywords or identifiers, or digits.

In any case, I don't feel strongly about it, but it seems like this change is non-trivial enough that it warrants a changelog entry or linked issue number.

andralex:
Probably a better heuristic could be found. The egregious problem (which did hurt us, see the fixed tests) is that links starting with `http:` were transformed into tags. The space fixes that.

The second issue, which may occur in flowing text, is words followed by colons that are not intended at all to be tags. Worse, reformatting paragraphs (e.g. with emacs' M-q (fill-paragraph) makes these unintended tags spring into existence. Consider:

```
Lorem ipsum dolor sit amet, nam risus, sodales orci tempus sociis vel 
lacinia, adipiscing vestibulum viverra ipsum, consequat vestibulum dolor
quam in, ante accumsan amet tristique est sit. Depending on the input,
the function frobnicate returns: (a) junk, (b) meh, or (c) something awesome.
```

Now the author edits the paragraph here and there and at some point the word "returns:" is liable to fall at the beginning of a line. Bam, you get it formatted as a "Returns:" heading. The uppercase requirement takes care of that.

Of course there will be situations when you do have a titlecased word followed by a colon that you really want to be part of flowing text, e.g.:

```
Depending on the input, the function frobnicate returns: (a) junk,
(b) meh, or (c) something awesome. Unconvinced? Consider: some
people still edit code with Notepad.
```

Again, "Consider:" is liable to be reflown at the start of a line and unwittingly transformed into a section heading.

This kind of surprising formatting is the stuff of nightmares for authors. Any ideas for a better heuristic that doesn't break too much with the past? I'm thinking e.g. requiring a non-keyword label followed by a colon to be followed immediately by a newline. (Keywords being the predefined labels `Returns:`, `Authors:`, `Copyright:` etc). Thoughts?

andralex:
Also, it shouldn't be forgotten that an author can _always_ force a heading label by using `DDOC_SECTION` or `DDOC_SECTION_H` manually. But there's no obvious escape from automated formatting of words followed by colons.

CyberShadow:
> Any ideas for a better heuristic that doesn't break too much with the past?

How about requiring an empty line (two consecutive newlines) before a section heading, unless the previous line is also a section heading?

wilzbach:
> I'm not entirely sure about being picky about case. It precludes starting section names with things like keywords or identifiers, or digits.

I don't think being picky about the case will be a huge issue as long as it has a changelog entry and the spec is updated as there are many libraries out there which try to emulate the ddoc spec, e.g. this is how libddoc deals with sections (it follows the current "spec"). The nice thing about this change is that its more restrictive and existing parsers will continue to work. For example: "libddoc":

1) In its custom Lexer `header` is a special token - [parsed as follows](https://github.com/economicmodeling/libddoc/blob/master/src/ddoc/lexer.d#L221):

```d
current.type = Type.word; // ascii, digit, _
current.text = text[oldOffset .. offset];
if (parseHeader && prevIsNewline(oldOffset, text) && offset < text.length
&& text[offset] == ':')
{
  current.type = Type.header;
```

2) With the token stream, libbdoc [parses everything between two `header` tags as one section:](https://github.com/economicmodeling/libddoc/blob/master/src/ddoc/sections.d#L142):

```d
while (!lex.empty) switch (lex.front.type)
{
case Type.header:
	finishSection(); // <-- a previous section is closed on a new header tag
	name = lex.front.text;
	lex.popFront();
	lex.stripWhitespace(); // <- libbdoc ignores whitespace - it doesn't require it
	break;
case Type.newline:
	lex.popFront();
	if (name is null && !lex.empty && lex.front.type == Type.newline)
		finishSection();
	break;
default:
	lex.popFront();
	sliceEnd = lex.offset;
	break;
}
finishSection();
```
(as I saw `doc.d` I had to show that this parsing can be done much more legible)

>  I'm thinking e.g. requiring a non-keyword label followed by a colon to be followed immediately by a newline. (Keywords being the predefined labels Returns:, Authors:, Copyright: etc). 

Sounds reasonable, I prefer `<keyword>: <newline>` style anyways.
Just be sure to include all crazy keywords that are used in Phobos, see e.g. https://github.com/economicmodeling/libddoc/blob/master/src/ddoc/sections.d#L15 for so called "standard sections".
Btw it wouldn't be too hard to add to the `properly_documented_public_functions` Dscanner check a condition that enforces "standard sections" in Phobos - in other words a simple `comments.sections.all!(s => s.isStandard)` [here](https://github.com/dlang-community/D-Scanner/blob/master/src/analysis/properly_documented_public_functions.d#L175) would do.

> Any ideas for a better heuristic that doesn't break too much with the past?

FWIW I would only worry about breaking Phobos - most other well-documented projects have switched to Ddox or other documentation engines ;-)

wilzbach:
> There's a bug in std.datetime.systime's documentation exposed by this. I changed the lowercase labels in dlang/phobos#5606.

Nice! For quick reference:

https://dlang.org/phobos-prerelease/std_datetime_systime.html#.SysTime.fromISOExtString

image

wilzbach (approved):
As mentioned before, I like this change.

>  but it seems like this change is non-trivial enough that it warrants a changelog entry or linked issue number.

+1 (and [the spec](http://dlang.org/spec/ddoc.html#sections)).

andralex:
>How about requiring an empty line (two consecutive newlines) before a section heading, unless the previous line is also a section heading?

That's nice but a bit involved. I suggest we go with this simpler rule until we accumulate more experience.

CyberShadow:
> Among the most annoying: if a link http://... happens to fall at the start of a line the http is converted to a tag.

See https://github.com/dlang/dmd/pull/7043, which changed how URLs are treated outside of a macro. Not sure if URLs inside macros will still break the macro and be treated as a section heading.

In either case, this needs a rebase and tests.

andralex:
Found one more bug exposed and fixed by this change. In std.datetime.systime:

```
Otherwise, the resulting <span class="def-anchor" id="chunks"></span><code class="ddoc_psymbol">chunks</code> will be input ranges consuming the same
input: iterating over ...
```

The old tag detector would think "input:" is a section heading.

andralex:
@CyberShadow I experimented with the "must be preceded by an empty line and found it broke a bunch of code. See e.g. core.atomic:

```
 * Copyright: Copyright Sean Kelly 2005 - 2016.
 * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
 * Authors:   Sean Kelly, Alex Rønne Petersen
 * Source:    $(DRUNTIMESRC core/_atomic.d)
```

We'd need to change it to this to make it work:

```
 * Copyright: Copyright Sean Kelly 2005 - 2016.
 *
 * License:   $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
 *
 * Authors:   Sean Kelly, Alex Rønne Petersen
 *
 * Source:    $(DRUNTIMESRC core/_atomic.d)
```

andralex:
I've piggybacked some tests to the tests afferent to https://github.com/dlang/dmd/pull/7043, please don't rake me over coals for this.

CyberShadow:
> I experimented with the "must be preceded by an empty line and found it broke a bunch of code.

Yes, that's completely expected. This is why my suggestion also has the condition that the previous line may also be another one-line section.

andralex:
@CyberShadow that may work indeed from what I see in the diffs; let's leave it for another day.

CyberShadow:
Please remove id.h.

test/compilable/ddoc4899.d
```diff
@@ -4,8 +4,6 @@
 /*
 TEST_OUTPUT:
 ---
-compilable/ddoc4899.d(16): Warning: Ddoc: Stray '('. This may cause incorrect Ddoc output. Use $(LPAREN) instead for unpaired left parentheses.
-compilable/ddoc4899.d(16): Warning: Ddoc: Stray ')'. This may cause incorrect Ddoc output. Use $(RPAREN) instead for unpaired right parentheses.
```
CyberShadow: Same problem as https://github.com/dlang/dmd/pull/7043#discussion_r130244754.

I'm surprised this change isn't causing a merge conflict!

test/compilable/extra-files/ddoc17697.html
```diff
@@ -1,484 +1,506 @@
-
```
CyberShadow: Please either do the newline change in its own commit, or simply commit this file with the same line endings it uses currently.

CyberShadow (changes requested):

test/compilable/extra-files/ddoc17697.html
andralex: Wait I just copied the produced file over the other, so technically if I put them in different commits I create artificial steps that don't exist. So what difference does this line make if it's in a separate commit?

test/compilable/ddoc4899.d
andralex: So what's the action item here?

test/compilable/extra-files/ddoc17697.html
CyberShadow: Err, sorry, this is not about any particular new line, but about the newline characters in the file. If you look at the diff, you'll see that apparently every single line in the file changed - a symptom of the line endings changing (from CR/LF to LF or the other way).

CyberShadow: > So what difference does this line make if it's in a separate commit?

The question still makes sense despite the above, so to answer that - because it makes it actually possible to see only the lines that actually changed.

test/compilable/ddoc4899.d
CyberShadow: I provided some suggestions at the link above.

test/compilable/extra-files/ddoc17697.html
andralex: Oh I understand, thanks. Yeah that's a big and surprising change, will try to keep endlines as they were.

test/compilable/ddoc17697.d
```diff
@@ -11,6 +11,19 @@
...
+not_a_tag_because_it_does_not_start_with_uppercase:
+
+not_a_tag_because_no_whitespace_after_colon:x
+
+TagGalore: yes this is a tag
+
+MoreTag:
+yes the above is also a tag
+*/
+
```
andralex: Ew took me a bit of figuring out. Looks good now?

CyberShadow:
> Ew took me a bit of figuring out. Looks good now?

Not quite. I'm not sure what the change in https://github.com/dlang/dmd/pull/6989/commits/640a74d0bf7520feecca87d2c5290e80a284664e is about, but it doesn't look like it's related to newlines.

ddoc4899 is still useless.

test/compilable/ddoc4899.d
andralex: The change here is that previously `foo:)` was interpreted as a tag. But now, given that there's no whitespace between the colon and the close paren, the text is not interpreted as a tag anymore, the closing paren becomes legit, and the warning text disappears altogether. What am I missing?

CyberShadow: Not to repeat myself again, but, see the link above - specifically, you may want to read https://issues.dlang.org/show_bug.cgi?id=4899.

andralex:
>Not quite. I'm not sure what the change in 640a74d is about, but it doesn't look like it's related to newlines.

Oy Atom up again. Guess I'll get back to my trusty emacs!

andralex:
Oy that was unpleasant. Still have a difficult time figuring where those Windows line endings came about - I'm on Linux! @CyberShadow if you're around slack let's huddle there.

andralex:
About ddoc4899: it now verifies that a `foo:)` is not detected as a tag and the parens are closed properly. No?

changelog/ddoc-tag-deduction.dd
```diff
@@ -0,0 +1,38 @@
...
+settings/author preference) whether the literal use of "returns:" occurs at the
+beginning of a line. The second occurrence of "Returns:", however, is a tag
+meant to highlight the homonym section of the documentation.
+
+Other examples of undesired tag detection include the use of hyperlinks
+("http:", "mailto:" etc) in text.
+
+To distinguish between textual use and the use as a tag, the new rule is: a tag
+is detected if (a) it is an identifier at the beginning of a line; (b) it is
+followed by a colon and a whitespace; and (b) it starts with an uppercase
```
CyberShadow: Should be `(c)` here :)

andralex: done

CyberShadow (approved):

PR #6991, opened 2017-07-13
Template Parser Library Visitors

RazvanN7:
This PR makes the base, strict, permissive, transitive visitors templated so they can receive any AST family. Next step is to get rid of the old visitor and make ASTCodegen use the generic visitor


WalterBright:
@RazvanN7 needs a rebase

PR #6992, opened 2017-07-13
Fix Issue 17646 dmd segfaults on missing [tuple] foreach body in import

UplinkCoder:
As per title the reason this happens is because an `UnrolledLoopStatement` (which is what Tuple-foreach becomes) is not technically required to have statements in it's body.
However it seems so far  this was not trigged.

### DIFF
--- a/src/ddmd/statementsem.d
+++ b/src/ddmd/statementsem.d
@@ -739,7 +739,8 @@ private extern (C++) final class StatementSemanticVisitor : Visitor
                 }
                 st.push(new ExpStatement(loc, var));
 
-                st.push(fs._body.syntaxCopy());
+                if (fs._body)
+                    st.push(fs._body.syntaxCopy());
                 s = new CompoundStatement(loc, st);
                 s = new ScopeStatement(loc, s, fs.endloc);
                 statements.push(s);


src/ddmd/statementsem.d
```diff
@@ -739,7 +739,8 @@ private extern (C++) final class StatementSemanticVisitor : Visitor
                 }
                 st.push(new ExpStatement(loc, var));
 
-                st.push(fs._body.syntaxCopy());
+                 if (fs._body)
```
John-Colvin: `s/ {5}/ {4}/` :)

UplinkCoder: I should use an editor with auto-indent :)

UplinkCoder:
@andralex could you slap an auto-merge on this one ?
Thanks.

wilzbach:
> could you slap an auto-merge on this one ?

Any reason why this isn't based on stable?

UplinkCoder:
waahh. what did dlang bot do.
There you have the answer.

UplinkCoder:
closing in favor of version against stable

mathias-lang-sociomantic:
You know you can just change the target of this P.R.

wilzbach:
> waahh. what did dlang bot do.
There you have the answer.

Not sure what exactly happened, but in general it sends a friendly reminder when a regression is targeted at master and not stable (like it did here in the beginning), because sadly this still happens way too often.

I am only on my phone atm, but it could be that you changed the priority of the issue and the bot updated his message consequently?

UplinkCoder:
@mathias-lang-sociomantic I did just that,  changed the target.
It resulted in a PR with 201 commits.

John-Colvin:
changing the target isn't actually that useful in these cases seeing as it doesn't change the source branch at all. It's only really useful when you accidentally did the merge request to the wrong branch, not if you started your commits from the wrong branch.

wilzbach:
> It resulted in a PR with 201 commits.

Yes you need to rebase yourself. E.g.


```
git rebase - --onto=upstream/stable
```

We really need to teach the core contributors better git skills. Maybe at DConf18 there should be a talk/workshop about git?😉

mathias-lang-sociomantic:
> @mathias-lang-sociomantic I did just that, changed the target.
> It resulted in a PR with 201 commits.

Yes, then you have to rebase your P.R. onto stable

PR #6993, opened 2017-07-13
Fix Issue 17646 - dmd segfaults on missing [tuple] foreach body in import

UplinkCoder:
see  #6992

### DIFF
--- a/src/ddmd/statementsem.d
+++ b/src/ddmd/statementsem.d
@@ -739,7 +739,8 @@ private extern (C++) final class StatementSemanticVisitor : Visitor
                 }
                 st.push(new ExpStatement(loc, var));
 
-                st.push(fs._body.syntaxCopy());
+                if (fs._body)
+                    st.push(fs._body.syntaxCopy());
                 s = new CompoundStatement(loc, st);
                 s = new ScopeStatement(loc, s, fs.endloc);
                 statements.push(s);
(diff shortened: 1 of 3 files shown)


UplinkCoder:
@wilzbach can you set the auto-merge on this ?

UplinkCoder:
@andralex merge please :)

braddr:
Lacks a unit test, so not ready for a merge.  Also, not sure that this fix is correct.  It should have almost certainly failed earlier than getting down to this block of code.  Isn't it papering over an ice on invalid by just turning it into accepts invalid?

UplinkCoder:
It does not accept it.
Since in this situation we will generate an error regardless.

ZombineDev:
@UplinkCoder in any case, each bug fix needs a test case to prevent regressions in the future. In this case we need an entry in `test/fail_compilation`.

wilzbach:
> @wilzbach can you set the auto-merge on this ?

Thanks a lot for pointing out the leak of a test @braddr @ZombineDev - in addition I am sadly not very familiar with the DMD codebase, so I wouldn't merge this.
Besides at least at Phobos we try to keep non-trivial PRs open for at least a day 😉

yebblies:
@UplinkCoder Why can't you test this with the test case from the bug report?  This patch appears to be fine but unless there's a very good reason it should come with a test case, since this path is obviously not covered by the existing tests.

MartinNowak:
5 ppl. and noone adding the test from Bugzilla?

MartinNowak (approved):

PR #6994, opened 2017-07-14
Add MSP430, D_P16 version identifiers

luismarques:
This pull request adds version identifiers for the MSP430 architecture and 16-bit pointers.

### DIFF
--- a/src/ddmd/cond.d
+++ b/src/ddmd/cond.d
@@ -300,6 +300,7 @@ extern (C++) final class VersionCondition : DVCondition
             "MIPS_EABI",
             "MIPS_SoftFloat",
             "MIPS_HardFloat",
+            "MSP430",
             "NVPTX",
             "NVPTX64",
             "RISCV32",
--- a/test/fail_compilation/reserved_version.d
+++ b/test/fail_compilation/reserved_version.d
@@ -2,6 +2,8 @@
 /*
 TEST_OUTPUT:
 ---
+fail_compilation/reserved_version.d(105): Error: version identifier `MSP430` is reserved and cannot be set
+fail_compilation/reserved_version.d(106): Error: version identifier `D_P16` is reserved and cannot be set
 fail_compilation/reserved_version.d(107): Error: version identifier `DigitalMars` is reserved and cannot be set
 fail_compilation/reserved_version.d(108): Error: version identifier `GNU` is reserved and cannot be set
 fail_compilation/reserved_version.d(109): Error: version identifier `LDC` is reserved and cannot be set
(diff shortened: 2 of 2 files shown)


UplinkCoder:
please add the to the dlang site as well.

luismarques:
I did. <https://github.com/dlang/dlang.org/pull/1822>

MartinNowak (changes requested):
Same as for https://github.com/dlang/dlang.org/pull/1822#pullrequestreview-50186646.
So far we don't have any identifiers for pointer sizes (LP64 is an ABI) as you can simply use `(void*).sizeof`.

luismarques:
@MartinNowak `D_LP64` is a 64-bit version identifier flag. It does not indicate the LP64 ABI, despite the misleading name. There are several uses of `D_LP64` in druntime as 64/32 bit toggle. For instance, one of the first things in object.d is a `version(D_LP64) ... else ...` to set the types of `size_t` and `ptrdiff_t`. So there is precedent for using a `version` statement instead of a `static if`. Furthermore, in cases like that one, where we need to introduce an `else version(D_P16)` to cover the new 16-bit case, having to change the whole block to a static if would introduce further disruption.

jpf91:
> D_LP64 is a 64-bit version identifier flag. It does not indicate the LP64 ABI, despite the misleading name. 

That's also how we treat it in GDC. I think we had this discussion before so we explicitly documented this: http://dlang.org/spec/version.html `Pointers are 64 bits 		(command line switch -m64). (Do not confuse this with C's LP64 model`

```c++
  /* Define what type to use for size_t, ptrdiff_t.  */
  if (POINTER_SIZE == 64)
    {
      global.params.isLP64 = true;
      Tsize_t = Tuns64;
      Tptrdiff_t = Tint64;
    }
  else
    {
      Tsize_t = Tuns32;
      Tptrdiff_t = Tint32;
    }

// [...]

  /* LP64 only means 64bit pointers in D. */
  if (global.params.isLP64)
    VersionCondition::addPredefinedGlobalIdent ("D_LP64");
```

Geod24:
I thought 16 bits would never be supported (mostly because @WalterBright said so) ?
I was trying to find a ref when I first saw this P.R. but couldn't, but luckily for me [it came up again](http://forum.dlang.org/post/ol91im$2o0j$1@digitalmars.com).

CyberShadow:
> I thought 16 bits would never be supported (mostly because @WalterBright said so) ?

I think just x86-like 16-bit, which requires near and far pointers in the language itself.

JinShil:
> > I thought 16 bits would never be supported (mostly because @WalterBright said so) ?
>
> I think just x86-like 16-bit, which requires near and far pointers in the language itself.

Yeah, you'd have to get clarification from @WalterBright about what he means by this.  He may be against supporting 16-bit platforms due to some of the weird, legacy things that come with them.  Simply supporting 16-bit pointers doesn't seem that controversial.

MartinNowak (approved):

ibuclaw:
> He may be against supporting 16-bit platforms due to some of the weird, legacy things that come with them. Simply supporting 16-bit pointers doesn't seem that controversial.

The compiler is free to support them. The standard runtime is not expected to. But this is where projects like minlibd come in to provide the embedded support.

PR #6995, opened 2017-07-15
fix Issue 17541 - Function attribute deduction depends on compile inv…

WalterBright:
…ocation

We'll see how this goes before spreading it to other inferred attributes.

### DIFF
--- a/src/ddmd/func.d
+++ b/src/ddmd/func.d
@@ -1288,6 +1288,7 @@ extern (C++) class FuncDeclaration : Declaration
 
     final bool isNogc()
     {
+        //printf("isNogc() %s, inprocess: %d\n", toChars(), !!(flags & FUNCFLAGnogcInprocess));
         if (flags & FUNCFLAGnogcInprocess)
             setGC();
         return type.toTypeFunction().isnogc;
@@ -1306,6 +1307,13 @@ extern (C++) class FuncDeclaration : Declaration
      */
     final bool setGC()
     {
+        //printf("setGC() %s\n", toChars());
+        if (flags & FUNCFLAGnogcInprocess && semanticRun < PASSsemantic3 && _scope)
+        {
+            this.semantic2(_scope);
+            this.semantic3(_scope);
+        }
+
         if (flags & FUNCFLAGnogcInprocess)
         {
             flags &= ~FUNCFLAGnogcInprocess;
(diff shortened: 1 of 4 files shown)


JohanEngelen:
Thanks for working on this.
Yesterday I started dustmiting another testcase (nothrow deduction of postblit) but it's complex and will probably take a few more days.

WalterBright:
Please test this on your code. If it works I'll do it for `nothrow` as well, and maybe that will save you the trouble of dustmiting it.

JohanEngelen:
Just tested: it results in an ICE. I don't have debug info, other than
```
... #1 #2 #3 = crash handler stuff ...
#4 0x0000000000574a39 FuncExp::matchType(Type*, Scope*, FuncExp**, int) (/opt/ldc2/bin/ldc2+0x574a39)
#5 0x00000000005ebff0 implicitConvTo::ImplicitConvTo::visit(FuncExp*) (/opt/ldc2/bin/ldc2+0x5ebff0)
#6 0x000000000056a1a3 implicitConvTo(Expression*, Type*) (/opt/ldc2/bin/ldc2+0x56a1a3)
#7 0x00000000005127e0 TypeFunction::callMatch(Type*, Array<Expression*>*, int) (/opt/ldc2/bin/ldc2+0x5127e0)
```

The ICE is caused by a side-effect: it now detects `@safe` code calling non-`@safe` code, previously undetected, so that's nice. When I fix that, this PR indeed correctly deduces `@nogc` in the complex wekapp case too. (but things still don't work because `nothrow` is not deduced)

Getting to the cause of the ICE will be another long dustming session...
Still dustmiting the `nothrow` case, because we need more of these complex testcases.

JohanEngelen:
@WalterBright The `@safe` ICE is gone, so this seems to work now.
Note that the test you added does not actually test whether the problem is fixed.

JohanEngelen:
@WalterBright What can I do to help this go forward?

ZombineDev:
@JohanEngelen the answer in general is to ping someone else (who hopefully is available) with write access to the repo, as the policy is that one should not merge his/her own PRs.

Ping @ibuclaw  @MartinNowak  @klickverbot can you take a look?

JohanEngelen:
@ZombineDev The PR needs some more work. Right now it only fixes things for `gc`, but not yet for `nothrow`, `pure`, and `safe`/`trusted`.

WalterBright:
I suggest pulling this and doing other attributes in a follow-on.

JinShil:
According to digger, this PR introduced a regression:  https://issues.dlang.org/show_bug.cgi?id=19000

PR #6996, opened 2017-07-15
posix.mak: Fix missing dependencies on HOST_DMD_PATH

CyberShadow:
Urgent hotfix to unbreak the documentation tester.

Every use of HOST_DMD_RUN has to be accompanied with a HOST_DMD_PATH dependency, otherwise we would be trying to use a compiler which has not yet been downloaded.

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -366,16 +366,16 @@ $G/glue.a: $(G_GLUE_OBJS)
 $G/backend.a: $(G_OBJS)
 	$(AR) rcs $@ $(G_OBJS)
 
-$G/lexer.a: $(LEXER_SRCS) $(LEXER_ROOT)
+$G/lexer.a: $(LEXER_SRCS) $(LEXER_ROOT) $(HOST_DMD_PATH)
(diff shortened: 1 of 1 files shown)


wilzbach (approved):
Thanks!

wilzbach:
> Urgent hotfix to unbreak the documentation tester.

Did you figure out how the PR managed to sneak-in?

wilzbach:
> Did you already figure out how the PR managed to sneak-in?

Could be due to the fact that we build dmd three times at DAutotTest?
(once with Digger and twice at dlang.org: stable + master)

CyberShadow:
Yes, I did!

Before I had set up auto-deploy for dlang.org, to get the website building on the Ubuntu server, I had to do some gnarly work-arounds to work around the linker errors at the time. This involved adding a helper script to /usr/local/bin, and adding that directory (/usr/local/bin) to the PATH used for building the website.

Coincidentally, /usr/local/bin is also where the system-wide DMD installation lives on that machine. So, lack of a host DMD never bothered the makefiles running on DAutoTest because they would just use the global one in /usr/local/bin.

However, /usr/local/bin was not present in the auto-deploy's configuration, so the auto-deploy build failed to find a host DMD anywhere. Thus, the build failed.

PR #6999, opened 2017-07-16
Convert DUB package test into a Makefile test and run it with the proper compiler flags

wilzbach:
Fix-up to https://github.com/dlang/dmd/pull/6771#issuecomment-315622461

I think it's reasonable to test the new DUB package with the newly build D compiler and not the host compiler.

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -4,8 +4,9 @@ authors "Walter Bright"
 copyright "Copyright © 1999-2017, Digital Mars"
 license "BSL-1.0"
 
-targetType "none"
+targetType "library"
 dependency ":parser" version="*"
+sourcePaths
 
 subPackage {
   name "root"
--- a/test/Makefile
+++ b/test/Makefile
@@ -97,6 +97,7 @@ QUIET=@
 export RESULTS_DIR=test_results
 export MODEL
 export REQUIRED_ARGS=
+BUILD=release
 
 ifeq ($(findstring win,$(OS)),win)
 export ARGS=-inline -release -g -O
@@ -112,7 +113,13 @@ export DFLAGS=-I$(DRUNTIME_PATH)\import -I$(PHOBOS_PATH)
 export LIB=$(PHOBOS_PATH)
 else
 export ARGS=-inline -release -g -O -fPIC
-export DMD=../src/dmd
+# hack around the fact that on auto-tester MODEL isn't set for the testsuite
+ifeq (,$(wildcard ../generated/$(OS)/$(BUILD)/64/dmd))
+REAL_MODEL=32
+else
+REAL_MODEL=64
+endif
+export DMD=../generated/$(OS)/$(BUILD)/$(REAL_MODEL)/dmd
 export EXE=
 export OBJ=.o
 export DSEP=/
@@ -120,11 +127,14 @@ export SEP=/
 
 DRUNTIME_PATH=../../druntime
 PHOBOS_PATH=../../phobos
+DUB=dub
+# Default DUB flags (DUB uses a different architecture format)
+DUBFLAGS=--compiler=$(abspath $(DMD)) --arch=$(subst 32,x86,$(subst 64,x86_64,$(REAL_MODEL)))
 # link against shared libraries (defaults to true on supported platforms, can be overridden w/ make SHARED=0)
 SHARED=$(if $(findstring $(OS),linux freebsd),1,)
-DFLAGS=-I$(DRUNTIME_PATH)/import -I$(PHOBOS_PATH) -L-L$(PHOBOS_PATH)/generated/$(OS)/release/$(MODEL)
+DFLAGS=-I$(abspath $(DRUNTIME_PATH))/import -I$(abspath $(PHOBOS_PATH)) -L-L$(abspath $(PHOBOS_PATH))/generated/$(OS)/release/$(MODEL) -fPIC
 ifeq (1,$(SHARED))
-DFLAGS+=-defaultlib=libphobos2.so -L-rpath=$(PHOBOS_PATH)/generated/$(OS)/release/$(MODEL)
+DFLAGS+=-defaultlib=libphobos2.so -L-rpath=$(abspath $(PHOBOS_PATH))/generated/$(OS)/release/$(MODEL)
 endif
 export DFLAGS
 endif
@@ -209,3 +219,5 @@ $(RESULTS_DIR)/d_do_test$(EXE): d_do_test.d $(RESULTS_DIR)/.created
 	$(QUIET)$(DMD) -conf= $(MODEL_FLAG) -unittest -run d_do_test.d -unittest
 	$(QUIET)$(DMD) -conf= $(MODEL_FLAG) -od$(RESULTS_DIR) -of$(RESULTS_DIR)$(DSEP)d_do_test$(EXE) d_do_test.d
 
+test_dub:
+	cd ./dub_package && $(DUB) test $(DUBFLAGS) --force -v
--- a/travis.sh
+++ b/travis.sh
@@ -54,9 +54,6 @@ rebuild() {
 
 # test druntime, phobos, dmd
 test() {
-    # Temporarily skip testing the DUB package
-    #See also: https://github.com/dlang/dmd/pull/6999
-    #test_dub_package
     make -j$N -C ../druntime -f posix.mak MODEL=$MODEL unittest
     make -j$N -C ../phobos -f posix.mak MODEL=$MODEL unittest
     test_dmd
@@ -64,19 +61,13 @@ test() {
 
 # test dmd
 test_dmd() {
+    make -j$N -C test test_dub MODEL=$MODEL
     # test fewer compiler argument permutations for PRs to reduce CI load
     if [ "$TRAVIS_PULL_REQUEST" == "false" ] && [ "$TRAVIS_OS_NAME" == "linux"  ]; then
         make -j$N -C test MODEL=$MODEL # all ARGS by default
     else
         make -j$N -C test MODEL=$MODEL ARGS="-O -inline -release"
     fi
-}
-
-# test dub package
-test_dub_package() {
-    pushd test/dub_package
-    dub test
-    popd
 }
 
 for proj in druntime phobos; do


test/Makefile
```diff
@@ -112,14 +113,15 @@ export DFLAGS=-I$(DRUNTIME_PATH)\import -I$(PHOBOS_PATH)
 export LIB=$(PHOBOS_PATH)
 else
 export ARGS=-inline -release -g -O -fPIC
-export DMD=../src/dmd
+export DMD=../generated/$(OS)/$(BUILD)/$(MODEL)/dmd
```
wilzbach: CC @RazvanN7 - we really need to get this migration finished...

test/Makefile
```diff
@@ -209,3 +211,5 @@ $(RESULTS_DIR)/d_do_test$(EXE): d_do_test.d $(RESULTS_DIR)/.created
 	$(QUIET)$(DMD) -conf= $(MODEL_FLAG) -unittest -run d_do_test.d -unittest
 	$(QUIET)$(DMD) -conf= $(MODEL_FLAG) -od$(RESULTS_DIR) -of$(RESULTS_DIR)$(DSEP)d_do_test$(EXE) d_do_test.d
 
+test_dub: $(DMD)
+	cd ./dub_package && $(DUB) test --compiler=$(abspath $(DMD)) --force
```
wilzbach: `--root` didn't work & we should get Travis green ASAP:

```
> DFLAGS="-m64" dub test --compiler=/home/seb/dlang/dmd/generated/linux/release/64/dmd --force --root=/home/seb/dlang/dmd/test/dub_package
DFLAGS="-m64" dub test --compiler=/home/seb/dlang/dmd/generated/linux/release/64/dmd --force --root=/home/seb/dlang/dmd/test/dub_package
Failed to invoke the compiler /home/seb/dlang/dmd/generated/linux/release/64/dmd to determine the build platform: {
/tmp/dub_platform_probe-b1b52c05-da2e-4097-892b-89a1507f1d3a.d(86): Error: undefined identifier `string`
/tmp/dub_platform_probe-b1b52c05-da2e-4097-892b-89a1507f1d3a.d(7):        while evaluating pragma(msg, "  \"compiler\": \"" ~ determineCompiler() ~ "\",")
  "frontendVersion": 2075,
  "compilerVendor": "Digital Mars D",
  "platform": [
/tmp/dub_platform_probe-b1b52c05-da2e-4097-892b-89a1507f1d3a.d(18): Error: undefined identifier `string`
/tmp/dub_platform_probe-b1b52c05-da2e-4097-892b-89a1507f1d3a.d(11):        while evaluating pragma(msg, "    " ~ determinePlatform())
  ],
  "architecture": [
/tmp/dub_platform_probe-b1b52c05-da2e-4097-892b-89a1507f1d3a.d(43): Error: undefined identifier `string`
/tmp/dub_platform_probe-b1b52c05-da2e-4097-892b-89a1507f1d3a.d(14):        while evaluating pragma(msg, "    " ~ determineArchitecture())
   ],
}
```

wilzbach: Also I decided to avoid overwriting the DFLAGS as the `-fPIC` by default is lost then.

travis.sh
```diff
@@ -54,29 +54,22 @@ rebuild() {
...
 test() {
-    test_dub_package
     make -j$N -C ../druntime -f posix.mak MODEL=$MODEL unittest
     make -j$N -C ../phobos -f posix.mak MODEL=$MODEL unittest
     test_dmd
 }
 
 # test dmd
 test_dmd() {
+    local make_targets="run_tests test_dub"
```
wilzbach: I didn't add `test_dub` to the default `all` target as it would then by run on the auto-tester as well. While we do want to test this on the long run, I wasn't sure whether `dub` is available on the auto-tester, so I left testing this open for another PR.

wilzbach:
CC @CyberShadow @andralex `master` is currently broken on Travis due to https://github.com/dlang/dmd/pull/6771#issuecomment-315622461, so I am raising the awareness for this one ;-)

travis.sh
jacob-carlborg: Yeah, that's why I only added it to the Travis CI tests from the beginning.

test/Makefile
```diff
@@ -97,6 +97,7 @@ QUIET=@
 export RESULTS_DIR=test_results
 export MODEL
 export REQUIRED_ARGS=
+BUILD=release
```
jacob-carlborg: Is this used?

wilzbach: Yep, I had to do quite a few try to figure out a working setup and in the one you reviewed I temporarily removed the new changes with `$(BUILD)`

test/Makefile
```diff
@@ -112,19 +113,28 @@ export DFLAGS=-I$(DRUNTIME_PATH)\import -I$(PHOBOS_PATH)
...
 else
 export ARGS=-inline -release -g -O -fPIC
-export DMD=../src/dmd
+# hack around the fact that on auto-tester MODEL isn't set for the testsuite
+ifeq (,$(wildcard ../generated/$(OS)/$(BUILD)/64/dmd))
+REAL_MODEL=32
+else
+REAL_MODEL=64
+endif
+export DMD=../generated/$(OS)/$(BUILD)/$(REAL_MODEL)/dmd
```
wilzbach: I already bumped into this before: https://github.com/dlang/dmd/pull/6870
This is required as `dub` doesn't forward the `-L-L` DFLAGS to the linker (in this case `dmd` again) and thus we need the `dmd.conf`, which isn't generated for `src/dmd`  by default anymore.

```
Linking...
/home/seb/dlang/dmd/generated/linux/release/64/dmd -of.dub/build/application-$DFLAGS-linux.posix-x86_64-dmd_2075-1833BAD8BC30CEBFE35D7CC6DA0FFC50/dmd-dub-test .dub/build/application-$DFLAGS-linux.posix-x86_64-dmd_2075-1833BAD8BC30CEBFE35D7CC6DA0FFC50/dmd-dub-test.o ../../.dub/build/library-$DFLAGS-linux.posix-x86_64-dmd_2075-3768A66D24D1220DBE5011CA9AC1DEA7/libdmd_parser.a ../../.dub/build/library-$DFLAGS-linux.posix-x86_64-dmd_2075-00BFCF605E8D9C274B8361919B5DB43D/libdmd_lexer.a ../../.dub/build/library-$DFLAGS-linux.posix-x86_64-dmd_2075-831C9FFAA175BCEDD4C554013935344A/libdmd_root.a -L--no-as-needed -m64 -defaultlib=libphobos2.so
Copying target from /home/seb/dlang/dmd/test/dub_package/.dub/build/application-$DFLAGS-linux.posix-x86_64-dmd_2075-1833BAD8BC30CEBFE35D7CC6DA0FFC50/dmd-dub-test to /home/seb/dlang/dmd/test/dub_package
Running ./dmd-dub-test 
./dmd-dub-test: error while loading shared libraries: libphobos2.so.0.75: cannot open shared object file: No such file or directory
```

wilzbach:
@CyberShadow instead of discussing in length the hypothetical problem of bootstrapping with an unsecure  host, could we maybe move our focus to the fact that __Travis is broken on master__?
(for some reason only I seem to care about this)

I already had quite some "fun" with the Makefiles and auto-tester here, but I think I finally found a setup that's working and testing the DUB package properly.

CyberShadow:
Sorry Sebastian, I'm currently dealing with a broken dev box at home and replying to emails is all I can do, and this looks like I'd need to sit down and wrap my head around this change to review it properly. If you'd like to submit a PR that simply disables the broken test for now I could merge that now, or someone else could look at this instead.

CyberShadow:
You could of course also revert the bad PR instead of trying to rush in a fix.

wilzbach:
> Sorry Sebastian, I'm currently dealing with a broken dev box at home and replying to emails is all I can do, 

Didn't know that. If it's related to `PIC`, then I am pretty annoyed by this as well and would love to update our Makefile scripts to use `-fPIC` by default everywhere on Posix...

>  you'd like to submit a PR that simply disables the broken test for now I could merge that now, or someone else could look at this instead.

Sounds like a good idea: https://github.com/dlang/dmd/pull/7001

> You could of course also revert the bad PR instead of trying to rush in a fix.

Nay, I like DUB support too much for this ;-)

CyberShadow:
> If it's related to PIC, then I am pretty annoyed by this as well and would love to update our Makefile scripts to use -fPIC by default everywhere on Posix...

No, actually, @earnestly on #archlinux let me know of a simple workaround. Just create a `cc`/`gcc` wrapper which adds `-no-pie`, e.g.:

```bash
#!/bin/bash
set -eu

next=/usr/sbin/$(basename "$0")

for arg in "$@"
do
	if [[ "$arg" == -L$HOME/data/software/dmd/* || "$arg" == -L$HOME/opt/dmd/* ]]
	then
		exec "$next" -no-pie "$@"
	fi
done

exec "$next" "$@"
```

Add/symlink this to `cc` and `gcc` in your `PATH`, and old releases up to 2.060 will work again.

test/Makefile
```diff
@@ -112,19 +113,28 @@ export DFLAGS=-I$(DRUNTIME_PATH)\import -I$(PHOBOS_PATH)
 export LIB=$(PHOBOS_PATH)
 else
 export ARGS=-inline -release -g -O -fPIC
-export DMD=../src/dmd
+# hack around the fact that on auto-tester MODEL isn't set for the testsuite
```
CyberShadow: That doesn't sound right.

https://github.com/braddr/at-client/blob/master/src/do_test_dmd.sh#L33-L37

There's certainly a MODEL parameter there.

Generally, if things can be fixed in the AT scripts, they should be. I think Brad has generally been responsive in merging AT pull requests as long as they don't require more involved things like merging some pull requests simultaneously or manually updating test machines.

test/Makefile
CyberShadow: Wait, so you want to build the Dub library with the built DMD, not the host DMD?

I guess testing that the compiler can auto-bootstrap is interesting, but wouldn't it be easier and more consistent to use the host DMD? Or is that difficult because only dmd's posix.mak knows what it is?

dub.sdl
```diff
@@ -4,8 +4,9 @@ authors "Walter Bright"
 copyright "Copyright © 1999-2017, Digital Mars"
 license "BSL-1.0"
 
-targetType "none"
+targetType "library"
```
jacob-carlborg: I was told by Sönke to not do this https://github.com/dlang/dmd/pull/6771#discussion_r116365515.

wilzbach: It doesn't copy all DFLAGS otherwise.

test/Makefile
wilzbach: As mentioned I have bumped into this before: https://github.com/dlang/dmd/pull/6870
And no I don't think that we want to wait for another one or two months here. If this gets fixed at the auto-tester's side, the workaround can easily be removed.

test/Makefile
wilzbach: >  to use the host DMD

We should test both, but this entire issue exits because gdc's fronted is too old and has bugs.

dub.sdl
jacob-carlborg: Hmm, ok.

test/Makefile
MartinNowak: As the goal is to check whether the dub package recipe works we should just use the host compiler and avoid additional complications.

test/Makefile
MartinNowak: Yes that looks wrong, osmodel.mak will already determine a default MODEL, unless one is explicity specified via `make MODEL=64`. Not sure what fails for you, but you can use `$(error ${MODEL})` to inspect variables.

dub.sdl
MartinNowak: Who doesn't copy what DFLAGS?

MartinNowak (changes requested):
I'd say let's move this out of the `test/Makefile` mess and leave it in `src/posix.mak` as a test_dub target.
Furthermore there is no need to deal with the `-conf=` complications or raw DFLAGS, just testing that the dub recipe works with the HOST_DMD compiler is sufficient (we shouldn't generate a dmd.conf in the src folder during testing).

test/Makefile
```diff
@@ -97,6 +97,7 @@ QUIET=@
 export RESULTS_DIR=test_results
 export MODEL
 export REQUIRED_ARGS=
```
MartinNowak: BTW, working with exported ENV parameters is really annoying when reading a log failure and not being able to rerun a command.

wilzbach:
Superseded by https://github.com/dlang/dmd/pull/7415 which simply uses the host dmd as suggested.

PR #7000, opened 2017-07-16
Use SSL for AUTO_BOOSTRAP

wilzbach:
I believe this doesn't matter, because on a modern system a D compiler should be present and package maintainers use source tarballs (or digger) to do the bootstrapping.

Unfortunately, the year needs to be set as well as the release layout is a bit weird.

See also:
- https://github.com/dlang/installer/pull/211
- https://trello.com/c/mWHZxwo5/89-https-for-s3-downloadsdlangorg

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -90,11 +90,12 @@ else
   # Auto-bootstrapping, will download dmd automatically
   # Keep var below in sync with other occurrences of that variable, e.g. in circleci.sh
   HOST_DMD_VER=2.072.2
+  HOST_DMD_YEAR=2016
   HOST_DMD_ROOT=$(TMP)/.host_dmd-$(HOST_DMD_VER)
   # dmd.2.072.2.osx.zip or dmd.2.072.2.linux.tar.xz
   HOST_DMD_BASENAME=dmd.$(HOST_DMD_VER).$(OS)$(if $(filter $(OS),freebsd),-$(MODEL),)
   # http://downloads.dlang.org/releases/2.x/2.072.2/dmd.2.072.2.linux.tar.xz
-  HOST_DMD_URL=http://downloads.dlang.org/releases/2.x/$(HOST_DMD_VER)/$(HOST_DMD_BASENAME)
+  HOST_DMD_URL=https://s3-us-west-2.amazonaws.com/downloads.dlang.org/releases/$(HOST_DMD_YEAR)/$(HOST_DMD_BASENAME)
   HOST_DMD=$(HOST_DMD_ROOT)/dmd2/$(OS)/$(if $(filter $(OS),osx),bin,bin$(MODEL))/dmd
   HOST_DMD_PATH=$(HOST_DMD)
   HOST_DMD_RUN=$(HOST_DMD) -conf=$(dir $(HOST_DMD))dmd.conf


src/posix.mak
```diff
@@ -90,11 +90,12 @@ else
   # Auto-bootstrapping, will download dmd automatically
   # Keep var below in sync with other occurrences of that variable, e.g. in circleci.sh
   HOST_DMD_VER=2.072.2
+  HOST_DMD_YEAR=2016
   HOST_DMD_ROOT=$(TMP)/.host_dmd-$(HOST_DMD_VER)
   # dmd.2.072.2.osx.zip or dmd.2.072.2.linux.tar.xz
   HOST_DMD_BASENAME=dmd.$(HOST_DMD_VER).$(OS)$(if $(filter $(OS),freebsd),-$(MODEL),)
   # http://downloads.dlang.org/releases/2.x/2.072.2/dmd.2.072.2.linux.tar.xz
-  HOST_DMD_URL=http://downloads.dlang.org/releases/2.x/$(HOST_DMD_VER)/$(HOST_DMD_BASENAME)
+  HOST_DMD_URL=https://s3-us-west-2.amazonaws.com/downloads.dlang.org/releases/$(HOST_DMD_YEAR)/$(HOST_DMD_BASENAME)
```
CyberShadow: How long is that link going to keep working? It looks awfully fragile, with apparently having the datacenter location in it.

The solution here is rather to either enable proper SSL for downloads.dlang.org, or just add a SHA256 check to the downloads.

wilzbach: > How long is that link going to keep working? It looks awfully fragile, with apparently having the datacenter location in it.

For as long as the S3 bucket is alive (S3 buckets are bound to a datacenter and can't be moved to a different location or renamed).

> The solution here is rather to either enable proper SSL for downloads.dlang.org,

See also: https://trello.com/c/mWHZxwo5/89-https-for-s3-downloadsdlangorg

>  or just add a SHA256 check to the downloads.

This is going to be annoying for updates and ensuring that `sha256sum` is installed on all machines. Also we can't do the streaming extraction anymore.

CyberShadow: We may decide to move hosting the downloads away from Amazon to some other service at some point, which would make building older D versions overly onerous. It would be much better if we could avoid exposing "implementation details" like the real download server, and use dlang.org links.

If this is too complicated for downloads.dlang.org for some reason, even some redirects on https://dlang.org would be better. In fact, why not move the HTML pages of downloads.dlang.org to the dlang.org server?

CyberShadow: > Also we can't do the streaming extraction anymore.

If you really wanted to, you could still do it with `tee` and process substitution.

wilzbach: > It would be much better if we could avoid exposing "implementation details" like the real download server, and use dlang.org links.

Well, we could have downloads.dlang.io where the year mapping is hard-coded, but then all traffic would flow to a single point of failure and the entire point of using S3 is gone.

>  In fact, why not move the HTML pages of downloads.dlang.org to the dlang.org server?

Because they are automatically generated and I bet @MartinNowak wouldn't like sending yet another PR on every update.

CyberShadow: > Well, we could have downloads.dlang.io where the year mapping is hard-coded, but then all traffic would flow to a single point of failure and the entire point of using S3 is gone.

??? The website would just do a redirect. It would be no more of a single point of failure than the dlang.org website.

Not sure if W/A would like the downloads going though a server/domain not owned by the foundation, though.

> Because they are automatically generated and I bet @MartinNowak wouldn't like sending yet another PR on every update.

I did not say "part of dlang.org github repository", I just said same server, because apparently Jan has no problem setting SSL up.

wilzbach: > The website would just do a redirect. It would be no more of a single point of failure than the dlang.org website.

I think the best solution would be to either
- upload releases to GH releases too
- restructure the download,s.t. they are year independent
- use the releases to a provider that supports SSL out-of-the-box

Note that we could use Cloudfront on AWS to achieve SSL, but sadly that [would/might be really expensive](https://aws.amazon.com/cloudfront/pricing/) (not sure how traffic is currently generated).

> because apparently Jan has no problem setting SSL up.

https://dconf.org/

MartinNowak:
We have a story for TLS on S3, using their raw bucket links doesn't seem to be the way to go.
They are A ugly as hell, B might be subject to change, and C don't support the version redirects.
https://trello.com/c/mWHZxwo5/89-https-for-s3-downloadsdlangorg
Also how much trust does Amazon's TLS certificate provide? Or are we concerned about privacy?
For security concerns we might use checksums or the signatures instead.
You could also use the install script (with a separate `--path /tmp/stable_compilers` base folder) which already does check signatures if gpg is available. That could be extended with some https served sha256 checksum fallback.

MartinNowak:
> If it is too complicated to add SSL to downloads.dlang.org for some reason, even some redirects on https://dlang.org would be better. In fact, why not move the HTML pages of downloads.dlang.org to the dlang.org server?
> Because they are automatically generated and I bet @MartinNowak wouldn't like sending yet another PR on every update.

The redirects are S3 internal (downloads/2.x/2.075 -> downloads/2017/2.075), so we'd need to hardcode those on some dlang.org hosted download pages. For https://dlang.org -> https://s3 we'd furthermore need to hardcode the bucket links.

https://github.com/dlang/downloads.dlang.org/blob/master/make-links

Since I generally don't have the best connection to dlang.org, I'd be cautious to move all our downloads (including all the CI systems) through a bottleneck redirect at dlang.org.
In particular when it's unclear what we expect from this.
Caching and HTTPS doesn't go well together and a HTTPS keyring/checksum scheme to validate downloads might be the better choice.

PR #6394, opened 2017-01-01
fix errors with AVX vector code generation

WalterBright:
I was initially mystified how dmd could pass the test suite with these bugs, since the test suite was failing when I tried them. But then I remembered Brad had said that many (most?) of the test machines do not have AVX-capable CPUs, and hence the AVX tests did not run.

PR #6315, opened 2016-12-11
add VEX generation to cdvecfill()

WalterBright:

PR #7003, opened 2017-07-17
rewrite cdvecfill

MartinNowak:
- fix penalties for AVX-256 insns with XMM registers (see https://software.intel.com/en-us/articles/intel-avx-state-transitions-migrating-sse-code-to-avx)
- stop using AVX2 insns in AVX1 mode (to init 32-byte registers)
- avoid using temporary stack values
- improve instruction sequences for int loading
- use dedicated AVX2 int broadcast insns
- optimize fixresult to avoid temporary stack values and superfluous movs
- remove unnecessary REX.W prefixes for SIMD movs
- add asm test to test generated code against regressions

Review commit by commit!
Also see #6315 and #6394 which added the problematic AVX code paths.


WalterBright (approved):
Nice work!

MartinNowak:
I've disabled the OSX and Windows asm comparison tests completely. They use the same backend code, but due to ABI differences the assembly is different. Seems more pragmatic to just disable those tests instead of inventing some fuzzy matching.

PR #7006, opened 2017-07-18
better error messages for better C

WalterBright:
In converting C code to D, these identifiers crop up constantly, justifying a better error message.

### DIFF
--- a/src/ddmd/dscope.d
+++ b/src/ddmd/dscope.d
@@ -29,6 +29,7 @@ import ddmd.root.outbuffer;
 import ddmd.root.rmem;
 import ddmd.root.speller;
 import ddmd.statement;
+import ddmd.tokens;
 
 //version=LOGSEARCH;
 
@@ -644,6 +645,30 @@ struct Scope
         }
 
         return cast(Dsymbol)speller(ident.toChars(), &scope_search_fp, idchars);
+    }
+
+    /************************************
+     * Maybe `ident` was a C or C++ name. Check for that,
+     * and suggest the D equivalent.
+     * Params:
+     *  ident = unknown identifier
+     * Returns:
+     *  D identifier string if found, null if not
+     */
+    extern (C++) static const(char)* search_correct_C(Identifier ident)
+    {
+        TOK tok;
+        if (ident == Id.NULL)
+            tok = TOKnull;
+        else if (ident == Id.TRUE)
+            tok = TOKtrue;
+        else if (ident == Id.FALSE)
+            tok = TOKfalse;
+        else if (ident == Id.unsigned)
+            tok = TOKuns32;
+        else
+            return null;
+        return Token.toChars(tok);
     }
 
     extern (C++) Dsymbol insert(Dsymbol s)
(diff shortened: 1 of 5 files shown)


src/ddmd/mtype.d
```diff
@@ -7688,10 +7688,13 @@ extern (C++) abstract class TypeQualified : Type
                 error(loc, "'%s' is not defined, perhaps you need to import %s; ?", p, n);
             else
             {
-                auto id = new Identifier(p);
+                auto id = Identifier.idPool(p, strlen(p));
+                const(char)* q;
                 s = sc.search_correct(id);
                 if (s)
                     error(loc, "undefined identifier `%s`, did you mean %s `%s`?", p, s.kind(), s.toChars());
+                else if ((q = Scope.search_correct_C(id)) !is null)
```
ZombineDev: I think you can also use the following to narrow the scope `q` is visible in:
`else if (const(char)* q = Scope.search_correct_C(id))`

src/ddmd/expression.d
```diff
@@ -3967,9 +3967,12 @@ extern (C++) class IdentifierExp : Expression
             error("`%s` is not defined, perhaps `import %s;` is needed?", ident.toChars(), n);
         else
         {
+            const(char)* p;
             s = sc.search_correct(ident);
             if (s)
                 error("undefined identifier `%s`, did you mean %s `%s`?", ident.toChars(), s.kind(), s.toChars());
+            else if ((p = Scope.search_correct_C(ident)) !is null)
```
ZombineDev: ditto

PetarKirov (approved):
LGTM

src/ddmd/mtype.d
UplinkCoder: That would prevent inling.

src/ddmd/dscope.d
```diff
@@ -646,6 +647,30 @@ struct Scope
...
 
+    /************************************
+     * Maybe `ident` was a C or C++ name. Check for that,
+     * and suggest the D equivalent.
+     * Params:
+     *  ident = unknown identifier
+     * Returns:
+     *  D identifier string if found, null if not
+     */
+    extern (C++) static const(char)* search_correct_C(Identifier ident)
```
jacob-carlborg: Can we try to use D naming conventions? In this case, camel case.

wilzbach:
@WalterBright if you are focusing on `betterC`, it would be great if we could document it: https://github.com/dlang/dlang.org/pull/1796

WalterBright:
@wilzbach I already reviewed https://github.com/dlang/dlang.org/pull/1796

src/ddmd/dscope.d
WalterBright: I used the same convention as the existing `search_correct` to tie the two together.

jacob-carlborg: :(

atilaneves:
LGTM

MartinNowak:
This is conflating betterC (a.k.a. tinyD) with porting C code to D, let's hope that won't bite anyone.

PR #5730, opened 2016-05-06
[refactor] Move Statement semantic routines to a new file and reduce imports

yebblies:
It's not especially clear from the diff, but this removes 14 imports from `statement.d`, and the new `dsemantic.d` is only imported by `func.d`.  When complete the semantic analysis will be isolated from the rest of the compiler.

PR #7007, opened 2017-07-19
Separate Initializer semantic

RazvanN7:
This PR separates the semantic routines from the AST classes of the  Initializer nodes. It is a follow-up to @yebblies 's work #5730

### DIFF
--- a/src/ddmd/declaration.d
+++ b/src/ddmd/declaration.d
@@ -28,6 +28,7 @@ import ddmd.hdrgen;
 import ddmd.id;
 import ddmd.identifier;
 import ddmd.init;
+import ddmd.initsem;
 import ddmd.intrange;
 import ddmd.mtype;
 import ddmd.root.outbuffer;
--- a/src/ddmd/dinterpret.d
+++ b/src/ddmd/dinterpret.d
@@ -30,6 +30,7 @@ import ddmd.globals;
 import ddmd.id;
 import ddmd.identifier;
 import ddmd.init;
+import ddmd.initsem;
 import ddmd.mtype;
 import ddmd.root.array;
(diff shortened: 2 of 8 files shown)


RazvanN7:
@UplinkCoder any ideas why compiling druntime leads to CTFE errors?

UplinkCoder:
@RazvanN7 Nothing that jumps in the eye.
I suggest to leave the code as is, for now.

RazvanN7:
@UplinkCoder you mean drop the PR? why?

UplinkCoder:
The code is not likely to change.
So there is little benefit it splitting it out.

RazvanN7:
@UplinkCoder 

> Are you sure the constructors are getting called correctly ?
Yep. I just copy-pasted the code and added `i.` in front of (1) all the variables that would normally have been parameters and (2) calls to member functions.

> So there is little benefit it splitting it out.

I disagree. Refactoring the code this way makes the code easier to read and makes packaging the compiler as a library easier

wilzbach:
> I disagree. Refactoring the code this way makes the code easier to read and makes packaging the compiler as a library easier

I can't agree more. Code is read a lot more than its written and __every__ change that helps to improve the readability of the dmd codebase is imho very useful.
@UplinkCoder don't forgot that there are and will be people who want to understand the dmd codebase in the future.

@RazvanN7 Please don't stop cleaning up and refactoring the frontend. You are doing great and important work!

RazvanN7:
I plan on doing Expression next.

ibuclaw:
And C++ headers?

RazvanN7:
@ibuclaw forgot about those... done.

PR #7009, opened 2017-07-19
Require break and continue labels inside static foreach.

tgehr:
According to http://forum.dlang.org/post/okjekv$2dqs$1@digitalmars.com .

### DIFF
--- a/test/compilable/staticforeach.d
+++ b/test/compilable/staticforeach.d
@@ -558,11 +558,11 @@ struct TestStaticForeach{
 static:
     int test(int x){
         int r=0;
-        switch(x){
+    label: switch(x){
             static foreach(i;0..10){
-                case i: r=i; break;
+                case i: r=i; break label; // TODO: remove label when restriction is lifted
             }
-            default: r=-1; break;
+            default: r=-1; break label;
         }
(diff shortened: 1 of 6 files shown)


test/compilable/staticforeach.d
```diff
@@ -638,3 +638,36 @@ auto bug17660(){
...
+            }
+        }
+    })));
+    static assert(!is(typeof({
+        for(;;){
+            static foreach(i;0..1){
+                continue;
+            }
+        }
+    })));
```
jacob-carlborg: Tests that should fail to compile are usually placed inside `fail_compilation`, unless new conventions are used now.

tgehr: My test does not fail to compile. :P

WalterBright: Putting them in `fail_compilation/` allows for the exact error message generated to be tested, and so makes for a better test. There are plenty of examples in `fail_compilation/` to show how this is done - it's pretty easy.

tgehr: I have added fail_compilation tests.

andralex (approved):

PR #7013, opened 2017-07-22
fix Issue 16997 - Integral promotion rules not being followed for neg…

WalterBright:
…ation expressions

Not following integral promotion rules is a serious bug. Don't pull this yet, I want to see what impact fixing this has first.

### DIFF
--- a/changelog/fix16997.dd
+++ b/changelog/fix16997.dd
@@ -1,0 +1,34 @@
+fix Issue 16997 - Integral promotion rules not being followed for unary + - ~ operators
+
+To follow the C integral promotion rules, types are promoted to `int` before the unary
++ - or ~ operators are applied. Existing D does not do this.
+
+This is corrected when one of the following command line switches are used:
+
+    -transition=intpromote
+    -transition=16997
+
+It affects operands of type `byte`, `ubyte`, `short`, `ushort`, `char`, and `wchar`.
+The operands are promoted to `int` before the operator is applied. The result
+type will now be `int`.
+
+The values computed will be different for some values of the operands:
+
+* All values of - and ~ applied to ubyte, ushort, char, and dchar operands will change.
+
+* The values of `-byte(-128)` and `-short(-32768)` will change.
+
+If one of the '-transition=' switches is not specified, and warnings are turned on,
+a warning will be emitted for these operations. To turn off the warning and work
+as desired with or without the '-transition=' switch:
+
+Option 1:
+
+Rewrite `op b` as `op int(b)` to use correct, i.e. C-like, behavior.
+
+Option 2:
+
+Rewrite `op b` as `typeof(b)(op int(b))` to preserve the old behavior.
+
+Eventually, the warning will become a deprecation, then an error, and then the C-like behavior
+will become the default.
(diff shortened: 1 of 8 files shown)


WalterBright:
Blocked by https://github.com/dlang/phobos/pull/5646

changelog/fix16997.dd
```diff
@@ -0,0 +1,34 @@
+fix Issue 16997 - Integral promotion rules not being followed for unary + - ~ operators
+
+To follow the C integral promotion rules, types are promoted to `int` before the unary
++ - or ~ operators are applied. Existing D does not do this.
+
+This is corrected when one of the following command line switches are used:
+
+    -transition=promote
```
klickverbot: Perhaps `-transition=intpromote` for a little bit more of a hint as to the meaning when users encounter the flag in makefiles/…?

WalterBright: ok

MartinNowak:
This breaks lots of packages because this PR is using warnings instead of deprecations. This is an outdated practice from before the `-dw` and `-de` switches were added, and deprecations were changed to non-fatal by default.
Most people (and currently dub by default) treat warnings as fatal and deprecations as non-fatal, following dmd's no-noise warnings approach.
Also, fairly simple, we should handle deprecations in the language like those in libraries.

Also see https://github.com/dlang/dub/issues/493

WalterBright:
The new arrayop stuff you wrote breaks with this:
```
src/core/internal/arrayop.d-mixin-56(56): Warning: integral promotion not done for `-_param_2[pos]`, use '-transition=promote' switch or `-cast(int)(_param_2[pos])`
```
It's a bit hard to see where that is coming from in arrayop.d - can you have a look, please? In general, the idea is to replace `-x` with `0-x` and `~x` with `-1-x`

MartinNowak:
> It's a bit hard to see where that is coming from in arrayop.d - can you have a look, please? In general, the idea is to replace -x with 0-x and ~x with -1-x

Well the existing arrayop as well as the new implementation transform
`byte[] = -byte[] + byte[];`
into
```d
foreach (i; 0 .. result.length)
  result[i] = cast(byte)(-a[i] + b[i]);
```
.
There is one final cast before assigning that deals with intermediate integer promotions.
So it's somewhat of a false error of your deprecation. Could you check whether the expressions gets promoted or cast anyhow?
It's fairly somewhat laborious to add a workaround for this, as arrayops support arbitrary mixed types. I'd need to change the architecture of the codegen to run type inference of sub-expressions alongside the codegeneration in order to be able to add special code for some type+operation combinations.

andralex:
So what's the outlook on this? Deprecate or warn? I'm okay with each seeing that each approach has been massively used by various languages. Since we're not making this language change as part of a major general language update, deprecation seems a lot more sensible. @WalterBright can you please use deprecation?

WalterBright:
Blocked by https://github.com/dlang/phobos/pull/5793

WalterBright:
@andralex it's using deprecation now, and is passing.

schveiguy:
Note, the changelog here didn't reflect the update to doing deprecations instead of warnings. See https://issues.dlang.org/show_bug.cgi?id=18148

Geod24:
This deprecation triggers on valid code and is fairly annoying. See  https://issues.dlang.org/show_bug.cgi?id=20678

PR #7011, opened 2017-07-19
Rebase last-cdmd onto dmd-cxx (2.068.2)

ibuclaw:
Trying again as I seem to have invoked the wrath of dlang-bot in #7010

PR #7014, opened 2017-07-22
Merge tag 'last-cdmd' into dmd-cxx

ibuclaw:
Same as #7011, just using merge instead to preserve more of the history.


ibuclaw:
@wilzbach - FYI. :-)

ibuclaw:
Signing this off as is blocking other work.

PR #6418, opened 2017-01-09
fix Issue 16513 - slow findExistingInstance

MartinNowak:
- construct a proper hash to reduce collisions in template instance cache
- in particularly hash all sorts of expressions and combine hashes non-associatively
- the current hash degrades to linear search for many ti arguments, with this
  patch collisions while compiling phobos were reduced from ~36K to 376
- for the reported case (few 1000s alias symbols) compilation is now ~50% faster

PR #5082, opened 2015-09-16
[WIP] Generate C++ header files from public extern (C++) declarations

yebblies:
If I can get this to work, it means no more maintaining header files for the frontend.

This generates one mega-header, since it needs to reorder declarations to avoid forward references.  I need to work out what to do about declarations that can't be automatically converted but are referenced from the converted ones, like `Array` and `RootObject`.

PR #7016, opened 2017-07-23
Add missing CTFloat::hash function to ctfloat.h

ibuclaw:
Added to D implementation in #6418

### DIFF
--- a/src/ddmd/root/ctfloat.h
+++ b/src/ddmd/root/ctfloat.h
@@ -38,6 +38,8 @@ struct CTFloat
     static real_t parse(const char *literal, bool *isOutOfRange = NULL);
     static int sprint(char *str, char fmt, real_t x);
 
+    static size_t hash(real_t a);
+
     // Constant real values 0, 1, -1 and 0.5.
     static real_t zero;
     static real_t one;


wilzbach:
It's more than time that we generate these files.
@UplinkCoder @RazvanN7 do you remember https://github.com/dlang/dmd/pull/5082?
If using DMD as a library is too complicated, libdparse would do a fine job as well.
Is this generator on your roadmap?

wilzbach (approved):

UplinkCoder:
@wilzbach 
Currently using the compiler as a library is infeasible. 
However, I could do a c++ hdrgen switch for dmd akin to `-H`. Maybe `-Hc `?

wilzbach:
> Currently using the compiler as a library is infeasible.

What's missing?

> However, I could do a c++ hdrgen switch for dmd akin to -H. Maybe -Hc?

Oh I am not sure that's a good idea (except you can revive @yebblies's PR). The Unix philosophy is to do one step and do it will. DMD already does a lot of things ... with a full Bugzilla list, so imho doing this in a separate tool would be better.
As mentioned, if DMD as a library isn't ready yet, it shouldn't be a big deal to do this with libdparse.
What do other people think about this @yebblies, @andralex, @ibuclaw, @WalterBright?

PR #2194, opened 2013-06-17
Unresolved differences between gdc/dmd front ends

ibuclaw:
I recall mentioning this in one of the DDMD pulls.  FYI @yebblies

This covers all current unresolved differences between gdc and dmd front ends  (and acts as a TODO list).  An internal goal is to have them 100% shared code bases by release 2.065.  This makes a transition from the D front-end implement from C++ to D more of a 'drop in' (and is arguably a blocker and hindrance for adoption of the new implementation),  and also relieves a _HUGE_ maintenance burden when it comes round to merge-time.  However such a goal requires a bit of tug and pull from both sides of the fence so I'm raising this here to allow discussion with any core devs to iron these out in a way that we can say good riddance to the IN_GCC macro.

The diff can be summarised into the following points:
- Internal backend types differ
- Differing `_argptr` and `_arguments` front end implementation.
- Always use Itanium ABI for C++ mangling
- Support for GDC Extended Assembly in parser.
- Rejected/not yet pulled fixes for things done in the front end that send invalid/garbage code to the glue part of the front-end.

@yebblies - or anyone else - if you have any questions or suggestions, anything is welcome.  Also let me know if you'd rather this broken down... :o)

Will be sending separate pulls for things that can be folded in immediately...

Thanks
Iain.

PR #7017, opened 2017-07-23
[dmd-cxx] Apply all backports in GDC to DMD sources

ibuclaw:
All commits were taken from GDC, names of the commits are mostly specific to GDC.  This synchronises entirely GDC's copy of the frontend into this branch, with the notable exception of `ExtAsmStatement` (as documented in #2194)

On the CI side, bumped the library version to v2.071, and enabled building druntime, phobos doesn't build, but IIRC, there are a few cherry pickings from latter releases in GDC in order to be compatible.

Also blindly merged in the 2.071.2 backend release of dmd, so that is at least in sync with druntime.

One thing notably missing are tests from the testsuite.  These are present in gdc, but not here as that would require another long sitting in order to format and apply them.


ibuclaw:
Turned on the druntime unittests also.

ibuclaw:
@wilzbach - OK to sign off on this?

I promise that future PRs to this branch will be somewhat more reviewable. :-)

wilzbach:
> @wilzbach - OK to sign off on this?

Hehe not really, but considering that you are the only user of the cxx branch, the CI is passing and that no one else looked at this so far, I think the best way forward is to merge this somewhat blindly ;-)

ibuclaw:
Well I aim to keep them to a size like this:
https://github.com/D-Programming-GDC/GDC/pull/537

ibuclaw:
However, 2.072 won't be so small, and I can't comment on backend changes. Which I am doing blindly, and have no intention of keeping them up to date after they have been switched to D.

ibuclaw:
Given the way things are going, I think my last commit here would definitely be before the end of the year anyhow. . Once I have gdc prepped so I can swap the old frontend directly with the new, it shouldn't be long until the jump is made.

PR #7018, opened 2017-07-24
[WIP] Solve most spurious forward/circular referencing errors

Syniurge:
This is the implementation of the solution discussed [here](http://forum.dlang.org/post/auvbfyzmqjxtfhkvmkxu@forum.dlang.org). This PR:
- splits the `semantic()` functions into the part where the `symtab` of a `ScopeDsymbol` gets determined, and the rest. Although instead of adding another function that part is added to `importAll()` since its reason of being has been if I'm not mistaken to get the ScopeDsymbol ready for `search()`.
- makes overrides of `Dsymbol.importAll` callable multiple times recursively each time resuming from where the previous ongoing call left off.
- (todo) renames `importAll` to `determineMembers` to clarify its new role.
- makes `TemplateInstance.semantic3` call `semantic2`, and `TemplateInstance.semantic2` call `semantic` if needed. Every split `semantic` function also calls `importAll` so no additional `importAll` call is needed alongside existing `semantic` calls.

However the re-ordering of `determineMembers/semantic()` calls broke existing code. The reason is that [in some cases](https://github.com/dlang/dmd/pull/7018#issuecomment-318847785), deferring lookups cannot be avoided, which I didn't realize initially. So a large part of this PR is also to implement deferring (instead of error gagging). (will write more details soon). 

This addresses the two issues from the forum thread:

> https://issues.dlang.org/show_bug.cgi?id=17656
> https://issues.dlang.org/show_bug.cgi?id=17194

and probably a lot more cases.

Even @tgehr's code now compiles!:
```d
enum x = "enum xx = q{int y = 123;};";

struct SS{
    mixin(xx);
    mixin(x);
}

void main()
{
    import std.stdio : writeln;
    SS s; writeln(s.y); // prints 123
}
```

**To do:**
- fix the remaining breakages
- special aggregate methods currently generated by clone.d should be added to `.members` during `importAll()`, and their body should be generated not during `Struct/ClassDeclaration.semantic()` but during `FuncDeclaration.semantic()`

Along the way I've spotted possible further improvements.

**Further enhancements:**
- Merge `addMember` into `determineMembers`?
- Is passing `sc` to `semantic1.2.3` really useful once `setScope` has been called? (Was `setScope` historically an afterthought?) Passing sc only seems required if the symbol:
    * is declared inside a function through a DeclarationExp
    * is inside a `typeof()`
    * was added to a `.members` without a setScope call. Only possible case as far as I know: TemplateInstance appended to the members of the instantiating or root module.

    &nbsp;
    The scope has no reason to change and should stay the same between `importAll` and `semantic1,2,3`. Shouldn't `setScope` get called in these few special cases? Ensuring that every symbol about to be semantic'd has its scope set would also avoid having to re-create one or more Scope objects during `AggregateDeclaration/TemplateInstance/AttribDeclaration... .semantic2,3`. (However currently `_scope` is also used in several semantic() functions as a safeguard against recursive semantic calls, but I don't think it should be used that way, checking `semanticRun` would do too.)
- Symbols could retrieve basic info from `sc` during `setScope()`, so `__traits(getProtection/Linkage/...)` may work before `determineMembers`.

### DIFF
--- a/src/ddmd/aggregate.d
+++ b/src/ddmd/aggregate.d
@@ -110,9 +110,9 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
      * Create a new scope from sc.
      * semantic, semantic2 and semantic3 will use this for aggregate members.
      */
-    override Scope* newScope(Scope* sc)
+    override Scope* newScope()
     {
-        auto sc2 = sc.push(this);
+        auto sc2 = _scope.push(this);
         sc2.stc &= STCsafe | STCtrusted | STCsystem;
         sc2.parent = this;
         if (isUnionDeclaration())
--- a/src/ddmd/aggregate.d
+++ b/src/ddmd/aggregate.d
@@ -246,8 +246,9 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
 
             auto ad = cast(AggregateDeclaration)param;
 
-            if (v.semanticRun < PASSmembersdone)
-                v.importAll(null);
+            assert(v.semanticRun >= PASSmembersdone);
+//             if (v.semanticRun < PASSmembersdone)
+//                 v.importAll(null);
             // Return in case a recursive determineFields triggered by v.semantic already finished
             if (ad.sizeok != SIZEOKnone)
                 return 1;
@@ -320,7 +321,10 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
         }
 
         if (_scope)
-            importAll(_scope);
+            importAll(null);
+
+        if (semanticRun < PASSmembersdone)
+            return false; // FWDREF FIXME: could be a real error
 
         // Determine the instance size of base class first.
         if (auto cd = isClassDeclaration())
@@ -361,9 +365,10 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
     override final d_uns64 size(Loc loc)
     {
         //printf("+AggregateDeclaration::size() %s, scope = %p, sizeok = %d\n", toChars(), _scope, sizeok);
-        bool ok = determineSize(loc);
+        if (determineSize(loc))
+            return structsize;
         //printf("-AggregateDeclaration::size() %s, scope = %p, sizeok = %d\n", toChars(), _scope, sizeok);
-        return ok ? structsize : SIZE_INVALID;
+        return (semanticRun < PASSmembersdone) ? SIZE_DEFER : SIZE_INVALID;
     }
 
     /***************************************
--- a/src/ddmd/aggregate.d
+++ b/src/ddmd/aggregate.d
@@ -124,56 +124,8 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
         return sc2;
     }
 
-    override final void semantic2(Scope* sc)
+    final void genRTInfo()
     {
-        //printf("AggregateDeclaration::semantic2(%s) type = %s, errors = %d\n", toChars(), type.toChars(), errors);
-        if (!members)
-            return;
-
-        if (_scope)
-        {
-            error("has forward references");
-            return;
-        }
-
-        auto sc2 = newScope(sc);
-
-        determineSize(loc);
-
-        for (size_t i = 0; i < members.dim; i++)
-        {
-            Dsymbol s = (*members)[i];
-            //printf("\t[%d] %s\n", i, s.toChars());
-            s.semantic2(sc2);
-        }
-
-        sc2.pop();
-    }
-
-    override final void semantic3(Scope* sc)
-    {
-        //printf("AggregateDeclaration::semantic3(sc=%p, %s) type = %s, errors = %d\n", sc, toChars(), type.toChars(), errors);
-        if (!members)
-            return;
-
-        StructDeclaration sd = isStructDeclaration();
-        if (!sc) // from runDeferredSemantic3 for TypeInfo generation
-        {
-            assert(sd);
-            sd.semanticTypeInfoMembers();
-            return;
-        }
-
-        auto sc2 = newScope(sc);
-
-        for (size_t i = 0; i < members.dim; i++)
-        {
-            Dsymbol s = (*members)[i];
-            s.semantic3(sc2);
-        }
-
-        sc2.pop();
-
         // don't do it for unused deprecated types
         // or error types
         if (!getRTInfo && Type.rtinfo && (!isDeprecated() || global.params.useDeprecated) && (type && type.ty != Terror))
@@ -184,8 +136,8 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
             auto ti = new TemplateInstance(loc, Type.rtinfo, tiargs);
 
             Scope* sc3 = ti.tempdecl._scope.startCTFE();
-            sc3.tinst = sc.tinst;
-            sc3.minst = sc.minst;
+            sc3.tinst = _scope.tinst;
+            sc3.minst = _scope.minst;
             if (isDeprecated())
                 sc3.stc |= STCdeprecated;
 
@@ -199,9 +151,6 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
             e = e.ctfeInterpret();
             getRTInfo = e;
         }
-        if (sd)
-            sd.semanticTypeInfoMembers();
-        semanticRun = PASSsemantic3done;
     }
 
     /***************************************
--- a/src/ddmd/aggregate.d
+++ b/src/ddmd/aggregate.d
@@ -255,7 +255,7 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
             if (v.aliassym)
                 return 0;   // If this variable was really a tuple, skip it.
 
-            if (v.storage_class & (STCstatic | STCextern | STCtls | STCgshared | STCmanifest | STCctfe | STCtemplateparameter))
+            if (v.storage_class & (STCstatic | STCextern | STCtls | STCgshared | STCmanifest | STCctfe | STCtemplateparameter)) // FWDREF FIXME: we should skip static variables, but part of semantic should have run to know that
                 return 0;
             if (!v.isField() || v.semanticRun < PASSsemanticdone)
                 return 1;   // unresolvable forward reference
@@ -320,7 +320,7 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
         }
 
         if (_scope)
-            semantic(null);
+            importAll(_scope);
 
         // Determine the instance size of base class first.
         if (auto cd = isClassDeclaration())
--- a/src/ddmd/aggregate.d
+++ b/src/ddmd/aggregate.d
@@ -246,8 +246,8 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
 
             auto ad = cast(AggregateDeclaration)param;
 
-            if (v.semanticRun < PASSsemanticdone)
-                v.semantic(null);
+            if (v.semanticRun < PASSmembersdone)
+                v.importAll(null);
             // Return in case a recursive determineFields triggered by v.semantic already finished
             if (ad.sizeok != SIZEOKnone)
                 return 1;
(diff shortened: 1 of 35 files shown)


rainers:
Not sure how it compares to your changes, but I could compile Timon code with these changes: https://github.com/rainers/dmd/commit/acbe99d1939bba26bf10eaaeb8fb2310c86870df, though that failed a couple of other tests.

One of the major issues was that an aggregate is marked as semantically analyzed, even if a (template-)mixin had forward reference errors. Instead the analysis needs to be deferred to be tried agin later.

Syniurge:
@rainers are you sure you don't have more code beyond this commit that for example gags errors during semantic? I built your `tg_frontend` branch and the error is still here:
```
Error: undefined identifier 'xx', did you mean variable 'x'?
```
which was expected because it's still going through the following: `SS.semantic` -> `mixin(xx).semantic` -> `sc.search('xx')` -> `undefined identifier 'xx'`

rainers:
> oh nevermind sorry, I didn't realize you were talking about tgehr/d-compiler.

Sorry, it wasn't very clear in my message. It probably won't fix every error but the approach might work for the example above, too, but should indeed gag symbol lookup errors (and others) if a scope with unresolved mixins is searched.

Syniurge:
> but should indeed gag symbol lookup errors (and others) if a scope with unresolved mixins is searched

That would definitely help in many cases. but at the cost of having to defer and re-run `semantic()` a lot.

The fragility of the current approach can be attributed to unnecessary dependencies created by `semantic()` calls between symbols. By trimming down these dependencies this PR should solve all referencing errors for not too contrived code without having to defer anything. However it still can't handle a `symtab` entry popping up like this:

```d
struct S {
    mixin("enum E1 = \"882255\"; enum A = \"" ~ B ~ "\";");
    mixin("enum E2 = " ~ E1 ~ ";");
    mixin("enum B = \"yyiioo\";");
}

the_unsolved.d(3): Error: undefined identifier E1
```

For this I don't think there's any way other than gagging lookup errors, and deferring as long as there is more than one symbol being expanded in the same ScopeDsymbol. I plan on adding an intermediate semantic pass value between `PASSmembers` and `PASSmembersdone` to tackle this, but this should probably go into another PR.

Syniurge:
So I've hit a wall. The following code (reduced from `druntime/import/core/sys/posix/sys/types.d`):

```d
static if (__WORDSIZE == 64)
    enum __SIZEOF_PTHREAD_BARRIER_T = 32;
else
    enum __SIZEOF_PTHREAD_BARRIER_T = 20;

enum __WORDSIZE = 64;

struct pthread_barrier_t
{
    byte[__SIZEOF_PTHREAD_BARRIER_T] __size;
    // Error: undefined identifier __SIZEOF_PTHREAD_BARRIER_T
}
```

fails with this PR for a similar reason the previous mixin example fails => because `pthread_barrier_t.importAll()` gets called while evaluating the `static if` condition, so before `__SIZEOF_PTHREAD_BARRIER_T` becomes available.

And it worked before in vanilla because `StaticIfDeclaration.semantic` relied on `addMember` and doesn't invoke `importAll`. But this for example cannot be built with current DMD:

```d
static if (__WORDSIZE == 64) // Error: undefined identifier __WORDSIZE
    enum __SIZEOF_PTHREAD_BARRIER_T = 32;
else
    enum __SIZEOF_PTHREAD_BARRIER_T = 20;

static if (true) {
    enum __WORDSIZE = 64;
}
```

Whereas it compiles if we re-order the `static if`. Hence this isn't ideal either.

For this PR I initially thought that an option would be to temporarily stick to the current behavior and make `search()` try a first lookup without invoking `importAll()`. Meaning that this would remain fragile, and makes the order of declarations matter. If for example the condition relies on a template instantiation, later `importAll()` (or `semantic()` in current DMD) calls may introduce more specialized templates, so the evaluation of the condition will be silently incorrect.

But in any case, because the order of `importAll()/semantic()` calls has changed with the rest of the PR, the test case importing `core.sys.posix.sys.types` still fails.

So I guess the only option is to go for lookup error gagging and deferring, and do it straightaway in this PR. It'll take more `importAll/semantic` code reorganization but we'll have a robust solution instead of a half one.

Syniurge:
Slight change of plan, the current semantic methods are draining my sanity very quickly and that's because the current split into `importAll/semantic1/2/3` and the corresponding `semanticRun` states aren't adequate. Currently there is a number of tricks to only do or re-do parts of these functions, and sometimes you even have a `XXXXXExp.semantic` partially replicating by itself a `YYYYYDeclaration.semantic` call. This the consequence of having stuffed too many not necessarily dependent parts into a single function.

I remember @WalterBright saying at one point on the NG that he wanted to make the compilation lazy, and I think a small overhaul to make this happen would be very beneficial, it would make the semantic phases more robust, faster, easier to follow, and additionally fix forward referencing errors. For example the following:

```d
extern (C++) final class Import : Dsymbol
{
    override void semantic(Scope* sc)
    {
        mod.semantic(null);
    }

    override void semantic2(Scope* sc)
    {
        mod.semantic2(null);
    }
}
```

and

```d
    override void visit(ImportStatement imps)
    {
        s.semantic(sc); // s is an Import
        Module.addDeferredSemantic2(s);   
    }
```

have been doing a lot of useless work, when they're not triggering forward references errors. And there is no reason to recursively call `importAll/semanticN` on non-root modules' members, or even call `Module.parse` and `Module.importAll` at all, as long as an import doesn't get `search()`-ed. 

Making the semantic phases fully lazy doesn't mean a big rewrite, it's mainly about finer splitting. I've already split the « determine members » parts of `Dsymbol.semantic` functions, what's left to do is:

- further natural split into smaller "semantic blocks" that may be called **`at any time`**.
   `__traits()`, `typeof()`, ... may need to know a type, a protection, a base class, an aggregate size, an attribute, a manifest constant, etc. very early, during the `importAll()/addMember()` calls if they are used within a ``static if``, a `mixin`. But currently since the only entry point for any of these queries is to run `semantic()` on the symbol, the expression usually gets far more than what it's asking for. 
- there shouldn't be a preferred order for the `semanticXXX` calls. 
   The one rule should be: **`we want full semantic analysis on root modules, and the rest on demand, at any point`**.
   I.e there would be a single Module.semantic() loop in DMD's main(), gone would be semantic2, semantic3. importAll/determineMembers will be called automatically by the Dsymbol.semantic overrides, and so would the rest of semantic blocks.

Examples of smaller semantic blocks:

VarDeclaration:
 - semanticType
 - semanticInit

FuncDeclaration:
 - semanticType
 - determineAttributes
 - ...

These methods would all rely on `_scope` being set beforehand. AFAIK there is no reason for `_scope` to ever change once set. There are currently multiple paths that still don't call `setScope` and have to pass a `Scope*` argument. The right thing to do is to call `setScope` in these paths (even within `typeof()`, `is()`, `__traits(compile)`), and then never pass a `Scope*` again. Basic infos like protection, linkage can also be set during the `setScope()` call.

Each of these semantic blocks should have an independent state flag. `Dsymbol.semanticRun` shouldn't be linear anymore, but an array of 2 bits values accessed through @property methods (generated by a mixin): 

```d
enum SemState : uint
{
    Init,
    In,
    Defer, // in case of symbol lookup errors that may be due to an ongoing addMember() or determineMembers() call
    Done,
}

extern (C++) class Dsymbol : RootObject
{
    @property SemState membersState() { return semanticRun & 0x3; }
    @property SemState membersState(SemState value) { return semanticRun = (semanticRun & ~0x3) | value; }

    @property SemState addMemberState() { return semanticRun & (0x3<<2); }
    @property SemState addMemberState(SemState value) { return semanticRun = (semanticRun & ~(0x3<<2)) | (value << 2); }

    @property SemState initializerState() { return semanticRun & (0x3<<4); }
    @property SemState initializerState(SemState value) { return semanticRun = (semanticRun & ~(0x3<<4)) | (value << 4); }
    alias bodyState = initializerState;

    ...
}
```

They would replace `.sizeok`, `.baseok`, `.semantictiargsdone`, ...

I've started working on this, this expands the initial intent of this PR but reducing the redundancies and labyrinths of the semantic phases will also help prevent forward referencing regressions in the future.

src/ddmd/aggregate.d
```diff
@@ -64,8 +64,6 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
     uint structsize;        // size of struct
     uint alignsize;         // size of struct for alignment purposes
     VarDeclarations fields; // VarDeclaration fields
-    Sizeok sizeok;          // set when structsize contains valid data
```
ibuclaw: Don't forget to update c++ headers for all field and function signature changes once you are done.

src/ddmd/dsymbol.d
```diff
@@ -143,31 +143,14 @@ struct Prot
     }
 }
 
-enum PASS : int
+enum SemState : uint
```
ibuclaw: This breaks downstream compilers written in c++, can't we just keep the same PASS enum, just simplify it?

ibuclaw:
You are probably also in contention with #7119 and other semantic refactorings.

Syniurge:
> Odd looking comments will be removed and line indentation cleaned up before you're done, right?

Yes this PR is not ready at all. During the past three weeks I got distracted by other stuff but I'm getting back to it.

> You are probably also in contention with #7119 and other semantic refactorings.
> However it looks like there will be some rebase/squashing done.

About that, @RazvanN7's refactoring of Dsymbol.semantic() having been merged (#7060), I'm not really sure how my changes should be adapted to the visitor pattern once I'll rebase to current master. Should I add one visitor per new semantic function i.e semanticInitializer, semanticBody, semanticType, determineMembers, etc.?

> Just as a preference that no individual commit in a PR should leave the compiler broken. 

I'll try avoiding that when rebasing but this will probably result in huge commits since it's a redesign of semantic()?

RazvanN7:
>About that, @RazvanN7's refactoring of Dsymbol.semantic() having been merged (#7060), I'm not really sure how my changes should be adapted to the visitor pattern once I'll rebase to current master. Should I add one visitor per new semantic function i.e semanticInitializer, semanticBody, semanticType, determineMembers, etc.?

Ideally, if you want to split the semantic methods, you don't have to create visitors for them, you could just add them as private methods inside the visitor. We can talk on specific examples if you want to.

src/ddmd/dsymbol.d
Syniurge: Does this break of a lot of GDC code?

semantic1/2/3() are getting split into smaller functions, that may be called separately. For each of these functions I need to store an independent SemState value. SemState could be renamed to PASS, but won't the rest of the changes be heavy for GDC anyway?

ibuclaw: PASSobj is used to determine if code has been generated for the declaration.  And semantic may be requested in the declaration symbol pass.

Keeping the naming convention would be a noop for downstream, just adjusting the enum as necessary to reflect D definitions.

You mention that each semantic has an independent SemState, how can you determine from the outside which stage a given symbol is currently at?

JinShil:
May also address https://issues.dlang.org/show_bug.cgi?id=17267

@Syniurge Do you still plan on moving this forward?

ibuclaw:
The best way to take this forward is probably one commit at a time, split across many PRs so that each iteration has been checked and doesn't break anything.

Having this mega patch series kept open is fine in the meantime if the finished result works.

Has there been any discussion with @WalterBright about the intended design? There's been some recent refactorings being done in the semantic passes already, notably to remove the reliance on the scope variable to determine the semantic state.

ibuclaw:
I think I might grit my teeth here and say this is absolutely ridiculous and should be closed.

From a brief look at the individual commits, I don't think there's anything that could be salvaged from the scuppered wrecks.

Changes in this regard should be kept to a minimal.

Each commit should be buildable and pass the testsuite, even if that means setting up extra scaffolding with the intent to eventually take it down.

wilzbach:
> Changes in this regard should be kept to a minimal.

> Each commit should be buildable and pass the testsuite, even if that means setting up extra scaffolding with the intent to eventually take it down.

@Syniurge what Ian tries to say is that you should submit __small__ PRs that can be reviewed, checked and merged individually. `+3,430 −4,352` is just too much for a single PR. Tiny steps are the road to victory, not "Large diffs are not rendered by default."

I will close this now, but please don't see this as a rejection in general. You can always resubmit this, but in small reviewable PRs with as little churn (unrelated changes) as possible.

Syniurge:
No problem, I thought I'd have the time to work on this, rebase everything to clean the commits and catch up with master, but that fell short.

I'll be back in a few days to restart working on this, from scratch (current master). This time I'll try to split it into smaller PRs, by coming up with something that can fallback to semantic/semantic2/semantic3 so the changes are gradual instead of one big all-or-nothing reorganization.

ibuclaw:
Thanks :+1:

Much appreciated

ibuclaw:
Also bear in mind that there are a couple other patches that seek to reduce complexity made by Kenji that I will give a look over later.

They very likely conflict with your changes to design also.  Or they might just make some parts simpler.

PR #7019, opened 2017-07-25
optimizer -1-x to ~x

WalterBright:


UplinkCoder:
is that another optimization that is not predicated on `-O` ?

andralex (approved):

MartinNowak:
This introduced a regression.
[Issue 17695 – [Reg 2.076] ICE with vector negation](https://issues.dlang.org/show_bug.cgi?id=17695)

MartinNowak:
Reversion is quick and pain-free (https://github.com/dlang/dmd/pull/7034), feel free to follow up with a working implementation.

PR #7021, opened 2017-07-25
Fix Issue 17676 - bad inlining of functions with multiple return statements

UplinkCoder:
This is fixed by making sure that we only inline the return statement if it is the last statement in the function.
and leave the return there if this is not the last statement.

@andralex @WalterBright ready to pull!

### DIFF
--- a/src/ddmd/inlinecost.d
+++ b/src/ddmd/inlinecost.d
@@ -240,7 +240,8 @@ public:
     override void visit(ReturnStatement s)
     {
         // Can't handle return statements nested in if's
-        if (nested)
+        // Or is they are not the last statement in the function
+        if (fd.fbody.last() !is s || nested)
         {
             cost = COST_MAX;
         }
--- a/src/ddmd/inlinecost.d
+++ b/src/ddmd/inlinecost.d
@@ -240,7 +240,7 @@ public:
     override void visit(ReturnStatement s)
     {
         // Can't handle return statements nested in if's
-        // Or is they are not the last statement in the function
+        // Or if it's not the last statement in the function
         if (fd.fbody.last() !is s || nested)
         {
             cost = COST_MAX;
--- a/test/runnable/issue17676.d
+++ b/test/runnable/issue17676.d
@@ -1,0 +1,35 @@
+import core.stdc.stdio;
+
+__gshared bool bgEnable = 1;
+
+void smallAlloc() nothrow
+{
+    fullcollect();
+}
+
+size_t fullcollect() nothrow
+{
+    if(bgEnable)
+       return fullcollectTrigger();
+
+    return fullcollectNow();
+}
+
+size_t fullcollectNow() nothrow
+{
+    if (bgEnable)
+        assert(0);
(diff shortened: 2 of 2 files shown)


UplinkCoder:
test failures seem unrelated.

src/ddmd/inlinecost.d
```diff
@@ -240,7 +240,8 @@ public:
     override void visit(ReturnStatement s)
     {
         // Can't handle return statements nested in if's
-        if (nested)
+        // Or is they are not the last statement in the function
```
rainers: Please check the grammar of the sentence...

rainers:
I don't think this is the correct fix, because it disables too many valid optimizations. For example, the code in the bug report is compiled correctly if I assign the return value to a variable.

UplinkCoder:
@rainers 
It's an invalid optimization!
The exact kind of annoying bug that manifests only in certain circumstances and which is obscure and hard to find

UplinkCoder:
@rainers 
Look at the comment in there:
https://github.com/dlang/dmd/pull/6815/files#diff-80b3197d11a2644245c0b409d54ebe25R284

We do assume that we never want to inline a return-statement if it's not the last thing the function does

rainers:
> We do assume that we never want to inline a return-statement if it's not the last thing the function does

That's a limitation that has been (partially) removed in the past, see https://issues.dlang.org/show_bug.cgi?id=7625

UplinkCoder:
@rainers before the PR which caused the regression ReturnStatements could not be inlined as statements at all.
The correctness of the rewrite the inliner does when inling returnstatemens as statements curcially depends on the return statement being the very last thing that get's executed.

rainers:
dmd 2.074 inlines the code correctly if the return value is not discarded. It seems this limitation was ignored (see https://issues.dlang.org/show_bug.cgi?id=15878).

UplinkCoder:
@rainers  The pr in https://issues.dlang.org/show_bug.cgi?id=15878
Did pretty much exactly what the regression-inducing one did.
We cannot just write out a return-statement b/c that would terminate the calling function prematurely.
`auto caller () { ... callee .... /*something important*/ }`
If we were to inline the returnstatement as a statement
it would look like 
`auto  caller() { ... return /*whatever callee returned*/ /*something important now skipped over*/ }`

rainers:
> We cannot just write out a return-statement b/c that would terminate the calling function prematurely.

It works when inlining as expression, so it could be disabled for inlining as statements only (or the return statements could be replaced with gotos).

UplinkCoder:
> It works when inlining as expression, so it could be disabled for inlining as statements only (or the return statements could be replaced with gotos).

Yes! And I will do that in the long run.
However I experimented with those options a bunch and they did not work.
So let's first fix this very serious codegen bug.
And improve the inliner later.

rainers:
If unfixable I think it would be better to revert #6815 than to cause a performance rgression on existing code.

UplinkCoder:
true. let's revert  #6815 for now.

PR #7022, opened 2017-07-25
fix Issue 17568 - [scope] addresses to fields can be escaped from sco…

WalterBright:
…pe method

### DIFF
--- a/src/ddmd/escape.d
+++ b/src/ddmd/escape.d
@@ -175,7 +175,8 @@ bool checkParamArgumentEscape(Scope* sc, FuncDeclaration fdc, Identifier par, Ex
  */
 bool checkAssignEscape(Scope* sc, Expression e, bool gag)
 {
-    //printf("checkAssignEscape(e: %s)\n", e.toChars());
+    enum log = false;
+    if (log) printf("checkAssignEscape(e: %s)\n", e.toChars());
     if (e.op != TOKassign && e.op != TOKblit && e.op != TOKconstruct)
         return false;
     auto ae = cast(AssignExp)e;
@@ -219,7 +220,7 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
     bool result = false;
     foreach (VarDeclaration v; er.byvalue)
     {
-        //printf("byvalue: %s\n", v.toChars());
+        if (log) printf("byvalue: %s\n", v.toChars());
         if (v.isDataseg())
             continue;
 
@@ -304,7 +305,7 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
 
     foreach (VarDeclaration v; er.byref)
     {
-        //printf("byref: %s\n", v.toChars());
+        if (log) printf("byref: %s\n", v.toChars());
         if (v.isDataseg())
             continue;
 
@@ -312,7 +313,9 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
 
         // If va's lifetime encloses v's, then error
         if (va &&
-            (va.enclosesLifetimeOf(v) && !(v.storage_class & STCparameter) || va.storage_class & STCref) &&
+            (va.enclosesLifetimeOf(v) && !(v.storage_class & STCparameter) ||
+             va.storage_class & STCref ||
+             va.isDataseg()) &&
             sc.func.setUnsafe())
         {
             if (!gag)
@@ -346,7 +349,7 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
 
     foreach (FuncDeclaration fd; er.byfunc)
     {
-        //printf("fd = %s, %d\n", fd.toChars(), fd.tookAddressOf);
+        if (log) printf("byfunc: %s, %d\n", fd.toChars(), fd.tookAddressOf);
         VarDeclarations vars;
         findAllOuterAccessedVariables(fd, &vars);
 
@@ -384,6 +387,7 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
 
     foreach (Expression ee; er.byexp)
     {
+        if (log) printf("byexp: %s\n", ee.toChars());
         if (va && !va.isDataseg() && !va.doNotInferScope)
         {
             if (!va.isScope() && inferScope)
--- a/test/fail_compilation/retscope2.d
+++ b/test/fail_compilation/retscope2.d
@@ -164,12 +164,12 @@ void foo800()
 }
 
 /*************************************************/
-/+
+
 /*
-XEST_OUTPUT:
-
+TEST_OUTPUT:
+---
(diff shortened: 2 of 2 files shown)


UplinkCoder (approved):
lgtm

test/fail_compilation/retscope2.d
```diff
@@ -162,3 +162,25 @@ void foo800()
...
+
+int* p17568;
+struct T17568
+{
+    int a;
+    void escape() @safe scope
+    {
+        p17568 = &a;
+    }
+}
```
mathias-lang-sociomantic: Could you add a test case with a struct nested inside a function, and the pointer being at function scope ?

src/ddmd/escape.d
```diff
@@ -173,7 +173,8 @@ bool checkParamArgumentEscape(Scope* sc, FuncDeclaration fdc, Identifier par, Ex
  */
 bool checkAssignEscape(Scope* sc, Expression e, bool gag)
 {
-    //printf("checkAssignEscape(e: %s)\n", e.toChars());
+    enum log = false;
+    if (log) printf("checkAssignEscape(e: %s)\n", e.toChars());
```
rainers: Better use `static if(log)` here?

test/fail_compilation/retscope2.d
rainers: Here's a test as suggested by Mathias that still fails to fail to compile:
```
void test17568() @safe
{
	int* p;
	struct T
	{
		int a;
	}
	void escape(ref T t) @safe
	{
		p = &t.a;
	}
    {
        T t;
        escape(t);
    }
    auto bug = *p;
}
```

src/ddmd/escape.d
```diff
@@ -299,15 +300,17 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
...
             continue;
 
         Dsymbol p = v.toParent2();
 
         // If va's lifetime encloses v's, then error
         if (va &&
-            (va.enclosesLifetimeOf(v) && !(v.storage_class & STCparameter) || va.storage_class & STCref) &&
+            (va.enclosesLifetimeOf(v) && !(v.storage_class & STCparameter) ||
+             va.storage_class & STCref ||
+             va.isDataseg()) &&
```
rainers: This looks like a special case very much tuned towards the test case. Shouldn't the check for global data be part of enclosesLifetimeOf?

src/ddmd/escape.d
WalterBright: It doesn't make any difference there.

test/fail_compilation/retscope2.d
WalterBright: It does compile, but should not. Filed as https://issues.dlang.org/show_bug.cgi?id=17725

src/ddmd/escape.d
rainers: BTW: should all the changes to this condition also apply to the similar one at line 245?

MartinNowak: Indeed global data has infinite lifetime, thus an `if (isDataseg()) return true;` in `enclosesLifetimeOf` seems more appropriate.

MartinNowak (changes requested):

src/ddmd/escape.d
WalterBright: enclosesLifetimeOf presupposes an ordering that does not exist for statics. I.e. all statics have infinite lifetime, so there is no ordering. Locals, on the other hand, have an ordering.

WalterBright: Aside from that, this PR is meant to solve a specific issue. Redoing enclosesLifetimeOf would involve redoing all the rest of the logic. Something of that scope should be done as a separate refactoring.

WalterBright: > should all the changes to this condition also apply to the similar one at line 245?

The `va.isDataseg()` check is done a few lines down from that.

mathias-lang-sociomantic:
Ping @WalterBright (Martin's feedback)

BTW I created a `#scope` channel on slack if you ever want to join us :)

WalterBright:
I replied to @MartinNowak 's feedback. Please take a look at my comments.

ZombineDev:
Ping @MartinNowak @rainers can we move forward with this?

rainers:
> enclosesLifetimeOf presupposes an ordering that does not exist for statics

enclosesLifetimeOf pretends to do more than it does (by name and documentation). Maybe it should just be a local function inside checkAssignExp.

Currently, you shouldn't be calling enclosesLifetimeOf on statics at all, so check isDataSeg() before would be cleaner. Otherwise it returns random results.

I'm ok with merging, but maybe the bugzilla title should be narrowed down to this case (for the changelog), https://issues.dlang.org/show_bug.cgi?id=17725 is another example of the same issue. Not sure, if there are more.

WalterBright:
It works correctly regardless of the order, because those functions are pure and they're OR'd together.

You can change the title if you like. But I want to get this pulled, so I can work on the next [scope] issue. I tried working on several in parallel, but it wound up being a lot of confusing rebase extra work.

Because there are more scope issues in bugzilla, I don't care if any of these fixes are not models of clarity. I want to get it working, and have a test suite to cover the cases. Afterwards is the right time to take a step back and refactor it. Constantly refactoring before knowing what the end result needs to look like is a lot of churn.

Taking a month to pull one fix that is **one line** of code, when there are many in the queue, is a disaster. I have other PRs that consist merely of adding comments that sit around for weeks at a time, too.

MartinNowak:
> You can change the title if you like. But I want to get this pulled, so I can work on the next [scope] issue. I tried working on several in parallel, but it wound up being a lot of confusing rebase extra work.
> 
> Because there are more scope issues in bugzilla, I don't care if any of these fixes are not models of clarity. I want to get it working, and have a test suite to cover the cases. Afterwards is the right time to take a step back and refactor it. Constantly refactoring before knowing what the end result needs to look like is a lot of churn.
> 
> Taking a month to pull one fix that is one line of code, when there are many in the queue, is a disaster. I have other PRs that consist merely of adding comments that sit around for weeks at a time, too.

IMHO an indication that you shouldn't submit each commit as individual PR ;), but I don't want to open that discussion again.

We've just asked for the `isDataseg()` check to be moved into `enclosesLifetimeOf`, as any static variable has infinite lifetime and thus outlives anything else, it's the infimum of the partially ordered sequenceNumbers.
It smells like there could be more manifestation of the missing `isDataseg()` check, hence the request, which seems both trivial and reasonable.
https://github.com/dlang/dmd/blob/4d86fcba2fd2ef86cc85738cd2ac2b059dbb5800/src/ddmd/declaration.d#L2450
If you have a more strict understanding of the term `encloses`, that's fine, and indeed rewriting all the exisiting code that checks `enclosesLifetimeOf` and `isDataSeg` separately is out of scope here.

MartinNowak (approved):

PR #7024, opened 2017-07-26
fix Issue 17689 - finally clause leaks declarations

tgehr:


UplinkCoder:
@tgehr looks like you need to fix the `header1.d` test

tgehr:
That test fails on master, too. It contains this line:
```d
    writeln((i,1),2);
```

rainers:
> That test fails on master, too. It contains this line: writeln((i,1),2);

??? It fails because there is now an additional pair of braces in the finally clause. It has to be added to the header1i.di or eliminated during the header generation phase.

rainers:
The grammar explicitly doesn't add a scope to finally:
```
FinallyStatement:
    finally NoScopeNonEmptyStatement
```
Would this be a language change? Is the declared variable finally clause currently usable? When is it's destructor called?

tgehr:
Oh, I see. Well, I don't want it to add an additional pair of braces, I just want it to analyze the body in a new scope. I was going with what IfStatement/ForStatement/WhileStatement do in the parser.

I think it is a very silly design decision to make PSscope add an additional pair of braces that has to be removed during header generation, but I guess I'll bite.

This is not a language change, it is a bugfix. It does not make any sense to access a variable before its lifetime begins.

I don't think the grammar rule means there will not be a new scope. E.g.:

```
ForeachStatement:
    Foreach ( ForeachTypeList ; ForeachAggregate ) NoScopeNonEmptyStatement
```

rainers:
> I don't think the grammar rule means there will not be a new scope.

I agree it is unlikely to ever have meant not to introduce a new scope (I just verified current dmd agrees, too), but the documentation introduces different entities (No)Scope(NonEmpty)Statement and then uses them rather erratically.

The spec should be revisited, but not a blocker for this PR.

src/ddmd/hdrgen.d
```diff
@@ -574,13 +574,11 @@ public:
...
-        buf.writenl();
-        buf.level++;
+        if (!s.finalbody.isScopeStatement())
+            buf.level++;
         s.finalbody.accept(this);
-        buf.level--;
-        buf.writeByte('}');
-        buf.writenl();
+        if (!s.finalbody.isScopeStatement())
+            buf.level--;
```
WalterBright: This looks better, reduces the risk of copy/pasta errors, and does not use negation:
```
if (s.finalbody.isScopeStatement())
    s.finalbody.accept(this);
else
{
    ++buf.level;
    s.finalbody.accept(this);
    --buf.level;
}
```
(It's good practice to prefer code arrangements that avoid logical negation, as humans have cognitive problems with negation.)

Also, this needs a test case for the red statements. If it is not possible to execute those statements, then the code should be removed and replaced with an `assert`.

src/ddmd/parse.d
```diff
@@ -5845,7 +5845,7 @@ final class Parser(AST) : Lexer
                 if (token.value == TOKfinally)
                 {
                     nextToken();
-                    finalbody = parseStatement(0);
+                    finalbody = parseStatement(PSscope);
```
WalterBright: Thank you for tracking this down! It looks very similar to https://issues.dlang.org/show_bug.cgi?id=17690, can you do that one as well?

src/ddmd/hdrgen.d
tgehr: I fully agree, but I was following existing conventions. Should I also update the other locations in the file that use this pattern?

I'll add a test case.

src/ddmd/parse.d
tgehr: I had introduced a PSscope-related bug myself (https://github.com/dlang/dmd/pull/7023), so I thought I'd go hunting for more. :)

The fix for issue 17690 is here: https://github.com/dlang/dmd/pull/7026

WalterBright: > The fix for issue 17690 is here: #7026

Thanks! It helps to put a link to this PR in bugzilla. That way, nobody is going to waste time developing a fix when one already exists.

src/ddmd/hdrgen.d
WalterBright: > Should I also update the other locations in the file that use this pattern?

Not really necessary unless you feel like it.

src/ddmd/hdrgen.d
```diff
@@ -553,8 +563,18 @@ public:
     {
         buf.writestring("try");
         buf.writenl();
-        if (s._body)
-            s._body.accept(this);
+        if (s._body){
```
rainers: Apart from this tiny formatting issue, LGTM

rainers (approved):

PR #7025, opened 2017-07-26
fix Issue 17428 - [scope] class field assignment allows to escape sco…

WalterBright:
…pe parameters

### DIFF
--- a/test/fail_compilation/retscope2.d
+++ b/test/fail_compilation/retscope2.d
@@ -163,6 +163,7 @@ void foo800()
     throw e;
 }
 
+/*************************************************/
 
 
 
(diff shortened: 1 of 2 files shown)


test/fail_compilation/retscope2.d
```diff
@@ -162,3 +162,46 @@ void foo800()
...
+
+
+
+
+
+
+
+
+
+/*************************************************/
```
mathias-lang-sociomantic: No need for this since you use `#line`.

test/fail_compilation/retscope2.d
```diff
@@ -162,3 +162,46 @@ void foo800()
...
+/*
+TEST_OUTPUT:
+---
+fail_compilation/retscope2.d(1005): Error: scope variable `p` assigned to `this` with longer lifetime
+---
+*/
+
+#line 1000
+
+class C17428
```
mathias-lang-sociomantic: Could you add a test with a `scope class` ?

WalterBright: That's probably already covered in the many tests in `retscope.d` and `retscope2.d`, but in any case it is off topic to this particular bug fix.

rainers: This compiles without complaint before this PR (which seems correct):
```
class C17428
{
    int* _p;
}

void main() @safe
{
	int x;
	int* p = &x;
	scope C17428 c;
	c._p = p;
}
```
but does no longer afterwards.

WalterBright: fixed

rainers: It now allows the assignment to `c._p` even without `scope`.

WalterBright: fixed

rainers (approved):

PR #7026, opened 2017-07-26
fix Issue 17690 - scope guards leak declarations

tgehr:


rainers (approved):
I guess we should update the language spec, too, because it says "A scope block statement introduces a new scope" for ScopeBlockStatement, but that should be valid for NonEmptyOrScopeBlockStatement, too.

PR #7027, opened 2017-07-26
fix Issue 17430 - [scope] delegate can escape context ptr

WalterBright:
This works for me, so I added a test case for it.

### DIFF
--- a/test/fail_compilation/retscope2.d
+++ b/test/fail_compilation/retscope2.d
@@ -162,3 +162,67 @@ void foo800()
     scope Exception e;
     throw e;
 }
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+
+/*************************************************/
+
+/*
+TEST_OUTPUT:
+---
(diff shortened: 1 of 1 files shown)


test/fail_compilation/retscope2.d
```diff
@@ -162,3 +162,67 @@ void foo800()
...
+
+
+
+
+
+
+
+
+
+
```
mathias-lang-sociomantic: There's no point in using this hack if you use `#line`.

WalterBright: The point of this is that there are 3 PRs, and I'll have to do a rebase for two of them. These blank lines will make that easier. The test cases already use `#line`.

rainers: I guess it won't help avoid merge conflicts (you could insert the different test cases in separate places instead), but if it makes life easier for you, let's go ahead.

rainers (approved):

WalterBright:
The merge conflicts will still be there, they'll just be easier to resolve. Just delete the blank lines.

test/fail_compilation/retscope2.d
MartinNowak: Git's text-diff does merge changes to the same file without conflict if there is at least one common line between the changes, so adding newlines helps, but adding more than a couple of newlines seems excessive.

PR #7028, opened 2017-07-26
Revert "fix Issue 17399 - core.checkedint.addu cannot inline function"

UplinkCoder:
This reverts commit 8ff0a0a185e0a7d0c04812cf916fbbe6615f792c.

Because it introduced a regression.


rainers:
See https://github.com/dlang/dmd/pull/7029 for a possible fix instead of the revert.

WalterBright:
Superseded by #7029

PR #7029, opened 2017-07-26
fix issue 17676:  [REG 2.075] bad inlining of functions with multiple return statements

rainers:
do not inline function with multiple return statements as statements


src/ddmd/declaration.h
```diff
@@ -543,6 +543,7 @@ class FuncDeclaration : public Declaration
     // 2 if there's a throw statement
     // 4 if there's an assert(0)
     // 8 if there's inline asm
+    // 16 if there are multiple return statements
     int hasReturnExp;
```
ibuclaw: This should probably be enumerized.

PR #4218, opened 2014-12-19
Issue 12780 - Multiplying integer array by scalar double fails

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=12780

Support upcasting slice elements in array operations.
1. The tweak in the buildArrayIdent() retains backward compatibility.
   
   a) If no element upcasting exists in array operation, the generated function name is not changed.
   
   ``` d
   double_res[] = double_val * double_arr[];
   // _arrayExpSliceMulSliceAssign_d
   ```
   
   b) If upcasting slice elements is necessary, it will be encoded by the postfix "Of" + element-type-deco
   
   ``` d
   double_res[] = double_val * int_arr[];
   // _arrayExpSlice'Ofi'MulSliceAssign_d
   ```
2. Even if the whole array operation requires double element, the sub array operations can be processed by int.
   
   ``` d
   double_res[] = (int_val ^ int_arr[]) * double_val;
   // typeof(int_val ^ int_arr[]) == int[]
   ```

PR #7032, opened 2017-07-26
convert array ops to library calls

MartinNowak:
- capture expression tree as reverse polish notation
- pass scalar subtrees as single argument

Requires <s>https://github.com/dlang/druntime/pull/1891</s>, <s>https://github.com/dlang/druntime/pull/1899</s>, https://github.com/dlang/druntime/pull/1903
Supplemental <s>https://github.com/dlang/dlang.org/pull/1846</s>

### DIFF
--- a/test/fail_compilation/fail_arrayop3b.d
+++ b/test/fail_compilation/fail_arrayop3b.d
@@ -1,0 +1,17 @@
+// REQUIRED_ARGS: -o-
+
+/*
+example output from druntime
+----
+../../druntime/import/core/internal/arrayop.d(160): Error: static assert  "Binary op `+=` not supported for element type string."
+../../druntime/import/core/internal/arrayop.d(20):        instantiated from here: opsSupported!(true, string, "+=")
+../../druntime/import/object.d(3640):        instantiated from here: arrayOp!(string[], string[], "+=")
+fail_compilation/fail_arrayop3b.d(16):        instantiated from here: _arrayOp!(string[], string[], "+=")
+---
+*/
+void test11376()
+{
+    string[] s1;
+    string[] s2;
+    s2[] += s1[];
+}
--- a/src/ddmd/arrayop.d
+++ b/src/ddmd/arrayop.d
@@ -29,74 +29,6 @@ import ddmd.statement;
 import ddmd.tokens;
 import ddmd.visitor;
 
-/**************************************
- * Hash table of array op functions already generated or known about.
- */
-private __gshared FuncDeclaration[void*] arrayfuncs;
-
-
-/**************************************
- * Structure to contain information needed to insert an array op call
- */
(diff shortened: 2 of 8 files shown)


test/fail_compilation/fail_arrayop3b.d
```diff
@@ -0,0 +1,17 @@
+// REQUIRED_ARGS: -o-
+
+/*
+example output from druntime
+----
+../../druntime/import/core/internal/arrayop.d(160): Error: static assert  "Binary op '+=' not supported for element type string."
```
WalterBright: Using backticks instead of ' will enable color syntax highlighting!

WalterBright:
Very nice work. Will approve as soon as the changes are incorporated.

test/fail_compilation/fail_arrayop3b.d
MartinNowak: Surprising that it also works for static asserts, will change.

WalterBright:
@MartinNowak needs a rebase

MartinNowak:
Done @WalterBright

ZombineDev:
I really like the approach taken here. I wonder if this could be used for third-party linear algebra libraries like mir/ndslice. I.e. use the standard function template overloading mechanism - libraries would provide `arrayOp` overloads and the compiler would choose the most specialized one. Of course for array operands, druntime's overload would be the most specialized one.

MartinNowak:
> I really like the approach taken here, but I wonder if this could be useful for third-party linear algebra libraries like mir/ndslice. I.e. use the standard function template overloading mechanism - libraries would provide arrayOp overloads and the compiler would choose the most specialized one. Of course for array operands, druntime's overload would be the most specialized one.

Extensibility was not really a requirement for the story.
A simple extension point would be to change the qualified `object._arrayOp` to an unqualified `_arrayOp`.
https://github.com/dlang/dmd/pull/7032/files#diff-01366d4347de4bb6861fc2c52694b84aR133
Extensibility would also make sense for the normal array operands, e.g. for huge arrays and runtime-CPU/GPU/FPGA specializations.
We don't have much experience with such extension points yet, but if there is interest, it should be trivial for anyone to change the compiler to experiment with the idea.

MartinNowak:
Need https://github.com/dlang/druntime/pull/1899 to fix runnable/arrayop.d.

MartinNowak:
Yikes, someone bluntly added mixed type vector operations (int -> float, even worse int -> double) conversions (https://github.com/dlang/dmd/pull/4218).
I will only support scalar loops for those, as using complex core.simd chains is completely out of scope.
Fortunately gdc/ldc will still be able to auto-vectorize such loops.

MartinNowak:
Done https://github.com/dlang/druntime/pull/1903.

MartinNowak:
Rebased to trigger a retest now that the druntime pulls are merged.

PR #7033, opened 2017-07-26
Fix: [REG2.075.0] Covariant return types doesn't work with override i…

Kozzi11:
fix: https://issues.dlang.org/show_bug.cgi?id=17686

### DIFF
--- a/src/ddmd/mtype.d
+++ b/src/ddmd/mtype.d
@@ -775,7 +775,7 @@ extern (C++) abstract class Type : RootObject
 
                 // If t1n is forward referenced:
                 ClassDeclaration cd = (cast(TypeClass)t1n).sym;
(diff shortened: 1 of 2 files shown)


src/ddmd/mtype.d
```diff
@@ -775,7 +775,7 @@ extern (C++) abstract class Type : RootObject
 
                 // If t1n is forward referenced:
                 ClassDeclaration cd = (cast(TypeClass)t1n).sym;
-                if (cd.semanticRun < PASSsemanticdone)
+                if (cd.semanticRun < PASSsemanticdone && cd.baseok < BASEOKdone)
```
MartinNowak: Better use `!cd.isBaseInfoComplete` as in other occurrences of that check.

MartinNowak (approved):

PR #7034, opened 2017-07-26
fix Issue 17695 - [Reg 2.076] ICE with vector negation

MartinNowak:
- revert #7019 and add regression test


WalterBright:
Replaced with https://github.com/dlang/dmd/pull/7035

PR #7035, opened 2017-07-26
fix Issue 17695 - [Reg 2.076] ICE with vector negation

WalterBright:


andralex:
Thx. @MartinNowak, I'll leave the review to you!

andralex:
... and others!

MartinNowak (approved):

PR #7036, opened 2017-07-26
factor out config.sh script to update config files (VERSION, SYSCONFDIR)

MartinNowak:
- use generated/dub as string import path for copied/generated config files

### DIFF
--- a/config.sh
+++ b/config.sh
@@ -1,0 +1,17 @@
+#!/bin/sh
+
+set -ue
+
+OUTDIR="$1"
+VERSIONFILE="$2"
+SYSCONFDIR="$3"
+VERSION=$(git describe --dirty 2>/dev/null || cat "$VERSIONFILE") # prefer git describe
+
+mkdir -p "$OUTDIR"
+# only update config files when they actually differ to avoid unnecessary rebuilds
+if [ "$VERSION" != "$(cat "$OUTDIR/VERSION" 2>/dev/null)" ]; then
+    printf "$VERSION" > "$OUTDIR/VERSION"
+fi
+if [ "$SYSCONFDIR" != "$(cat "$OUTDIR/SYSCONFDIR.imp" 2>/dev/null)" ]; then
+    printf "$SYSCONFDIR" > "$OUTDIR/SYSCONFDIR.imp"
+fi
--- a/dub.sdl
+++ b/dub.sdl
@@ -29,9 +29,8 @@ subPackage {
     "src/ddmd/tokens.d" \
     "src/ddmd/utf.d"
 
-  // stringImportPaths cannot be used because it will make Dub extremely slow
-  // https://github.com/dlang/dub/issues/1199
-  dflags "-J$PACKAGE_DIR"
+  preGenerateCommands `cd "$${DUB_PACKAGE_DIR}" && ./config.sh generated/dub VERSION /etc`
+  stringImportPaths "generated/dub"
 
   dependency "dmd:root" version="*"
 }
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -459,17 +459,7 @@ optabgen.out : $G/optabgen
 
 ######## VERSION
 
-VERSION := $(shell cat ../VERSION) # default to checked-in VERSION file
-ifneq (1,$(RELEASE)) # unless building a release
-	VERSION := $(shell printf "`$(GIT) describe --dirty || cat ../VERSION`") # use git describe
-endif
-
-# only update $G/VERSION when it differs to avoid unnecessary rebuilds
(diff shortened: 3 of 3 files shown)


MartinNowak:
FYI @jacob-carlborg

wilzbach:
Could we first focus on re-enabling the Travis test?
WIP: https://github.com/dlang/dmd/pull/6999

dub.sdl
```diff
@@ -29,9 +29,8 @@ subPackage {
     "src/ddmd/tokens.d" \
     "src/ddmd/utf.d"
 
-  // stringImportPaths cannot be used because it will make Dub extremely slow
-  // https://github.com/dlang/dub/issues/1199
-  dflags "-J$PACKAGE_DIR"
+  preGenerateCommands "\"$${DUB_PACKAGE_DIR}config.sh\" \"$${DUB_PACKAGE_DIR}generated/dub\" VERSION /etc"
+  stringImportPaths "generated/dub"
```
jacob-carlborg: Will this work when building a package that has DMD as a dependency? I had problems with this before, that's why I switch from `-J.` to `-J$PACKAGE_DIR`.

MartinNowak: Well there is a difference between passing raw paths directly to dmd and telling dub which subfolder to use in a package. And yes it will work since dub knows how to deal with relative paths for the latter.

dub.sdl
```diff
@@ -29,9 +29,8 @@ subPackage {
     "src/ddmd/tokens.d" \
     "src/ddmd/utf.d"
 
-  // stringImportPaths cannot be used because it will make Dub extremely slow
-  // https://github.com/dlang/dub/issues/1199
-  dflags "-J$PACKAGE_DIR"
+  preGenerateCommands "\"$${DUB_PACKAGE_DIR}config.sh\" \"$${DUB_PACKAGE_DIR}generated/dub\" VERSION /etc"
```
jacob-carlborg: You can use backticks instead of quotes to avoid the need to escape the quotes.

dub.sdl
jacob-carlborg: Aha, ok. Using `.` does not work (at least not in this project). But other paths seem work.

dub.sdl
jacob-carlborg: Interesting, you're using one of the few undocumented environment variables 😃.

MartinNowak: It's not documented, but quite accessible to anyone invoking env and using git grep (or just searching the project).
https://github.com/dlang/dub/blob/dafdc47d690c63d074e1b9ed57349fe2eaaabd2b/source/dub/generators/generator.d#L480-L519

MartinNowak: How about you update the documentation https://github.com/dlang/dub-registry/blob/7f4f2300b93d7ebfce06d1989ca4c1a4a1758c79/views/inc.package_format.envvars.dt ;)?

dub.sdl
MartinNowak: No, `.` does exactly what you want.
I'm seeing a small issues with non-absolute arguments to `dub --root=dmd build :lexer` where dub tries to relate two relative paths.

MartinNowak: https://github.com/dlang/dub/pull/1207

MartinNowak:
> Could we first focus on re-enabling the Travis test?
> WIP: #6999

It's somewhat unrelated and #6999 isn't quite ready yet.

dub.sdl
jacob-carlborg: > No, . does exactly what you want.

I don't know how, but I got this error `Invalid variable: DFLAGS` when using a dot `.`. As you can see in https://github.com/dlang/dub/issues/1199. For some reason I don't get this error anymore today 😕.

config.sh
```diff
@@ -0,0 +1,15 @@
+#/bin/sh
```
jacob-carlborg: Missing the bang `!`.

config.sh
```diff
@@ -0,0 +1,15 @@
...
+
```
jacob-carlborg: Add some error handling, like `set -e`.

jacob-carlborg (changes requested):

config.sh
MartinNowak: Interesting that it still worked when being invoked from make and dub.

config.sh
MartinNowak: Done @jacob-carlborg

config.sh
jacob-carlborg: Yeah, but not from the shell. I got some really cryptic error message.

jacob-carlborg:
LGTM 👌.

PetarKirov (approved):

wilzbach (approved):

PR #7039, opened 2017-07-28
optimize StorageClassDeclaration.addMember()

WalterBright:
Don't execute expensive loop if not necessary.


WalterBright:
What was I thinking.

PR #6760, opened 2017-05-09
Static foreach

tgehr:
Implemented features:
 - adds 'enum' and 'alias' on foreach loop variables to force one of the two. (Also works with existing foreach over AliasSeq.)
 - static foreach and static foreach_reverse for all supported aggregate types of foreach. (Explicit ranges, Phobos ranges, opApply, ...)
 - behaves "the same" as static if and string mixins. It can be used both at function and declaration scope and any declarations in the loop body will be accessible from outside the loop.
 - __local and __previous can be used to keep local state across static foreach loop iterations. (This functionality has now been removed, still available in commit bd382a7730fe5b45ff22605a8977f0bdb10f00b2 .)

See test/compilable/staticforeach.d for a few examples of what is possible.

PR #7041, opened 2017-07-28
Document static foreach

tgehr:
This implements the changes suggested by @WalterBright on [pull request 6760](https://github.com/dlang/dmd/pull/6760#pullrequestreview-53001763).

### DIFF
--- a/src/ddmd/astbase.d
+++ b/src/ddmd/astbase.d
@@ -5930,7 +5930,7 @@ struct ASTBase
         final extern (D) this(Loc loc, ForeachStatement aggrfe, ForeachRangeStatement rangefe)
         in
         {
-            assert(!!aggrfe^!!rangefe);
+            assert(!!aggrfe ^ !!rangefe);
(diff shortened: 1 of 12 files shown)


src/ddmd/astbase.d
```diff
@@ -5930,7 +5930,7 @@ struct ASTBase
         final extern (D) this(Loc loc, ForeachStatement aggrfe, ForeachRangeStatement rangefe)
         in
         {
-            assert(!!aggrfe^!!rangefe);
+            assert(!!aggrfe ^ !!rangefe);
```
UplinkCoder: Thats a wired way of writing !aggfe != !rangefe

klickverbot: It's also the same as `!aggrfe ^ !rangefe`.

WalterBright:
Nice! A huge improvement. Thank you! Next up, getting it to pass the autotester :+1:

PR #7043, opened 2017-07-29
fix Issue 17697 - Ddoc: automatically highlight URLs outside of macro…

WalterBright:
… arguments

A nice convenience feature. Ddoc code like:
```
$(LINK http://www.foo.com)
```
will not be affected, because only URLs outside of macros are affected.

### DIFF
--- a/src/ddmd/doc.d
+++ b/src/ddmd/doc.d
@@ -1643,7 +1643,13 @@ struct DocComment
                     const(char)* q = p + utfStride(p);
                     while (isIdTail(q))
                         q += utfStride(q);
-                    if (*q == ':') // identifier: ends it
+                    if (*q == ':' &&            // 'identifier:' ends it
+                                                // but not 'http://' or 'https://'
+                        !(q[1] == '/' && q[2] == '/' &&
+                          (q - p == 4 && Port.memicmp(p, "http".ptr, 4) == 0 ||
+                           q - p == 5 && Port.memicmp(p, "https".ptr, 5) == 0)
+                         )
+                       )
                     {
                         idlen = q - p;
                         idstart = p;
@@ -2151,6 +2157,7 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
     int inCode = 0;
     int inBacktick = 0;
     //int inComment = 0;                  // in <!-- ... --> comment
+    int inMacro = 0;
     size_t iCodeStart = 0; // start of code section
     size_t codeIndent = 0;
     size_t iLineStart = offset;
@@ -2405,6 +2412,32 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
                 }
             }
             break;
+
+        case '$':
+        {
+            /* Look for the start of a macro, '$(Identifier'
+             */
+            leadingBlank = 0;
+            if (inCode || inBacktick)
+                break;
+            const slice = buf.peekSlice();
+            auto p = &slice[i];
+            if (p[1] == '(' && isIdStart(&p[2]))
+                ++inMacro;
+            break;
+        }
+
+        case ')':
+        {   /* End of macro
+             */
+            leadingBlank = 0;
+            if (inCode || inBacktick)
+                break;
+            if (inMacro)
+                --inMacro;
+            break;
+        }
+
         default:
             leadingBlank = 0;
             if (sc._module.isDocFile || inCode)
@@ -2418,7 +2451,18 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
                     size_t k = skippastURL(buf, i);
                     if (i < k)
                     {
-                        i = k - 1;
+                        /* The URL is buf[i..k]
+                         */
+                        if (inMacro)
+                            /* Leave alone if already in a macro
+                             */
+                            i = k - 1;
+                        else
+                        {
+                            /* Replace URL with '$(LINK URL)'
+                             */
+                            i = buf.bracket(i, "$(LINK ", k, ")") - 1;
+                        }
                         break;
                     }
                 }
--- a/test/compilable/ddoc17697.d
+++ b/test/compilable/ddoc17697.d
@@ -1,0 +1,16 @@
+// PERMUTE_ARGS:
+// REQUIRED_ARGS: -D -Dd${RESULTS_DIR}/compilable -o-
+// POST_SCRIPT: compilable/extra-files/ddocAny-postscript.sh 17697
+
+/***
+ * See:
+ *    http://www.fooa.com/test1
+ *    http://www.fooa.com/_test1
+ *    https://www.foob.com/test1
+ *    $(LINK http://www.fooc.com/test1)
+ *    $(LINK2 http://www.food.com/test1, test1)
+ */
+
+module test1;
+
+int a;
--- a/test/compilable/ddoc4899.d
+++ b/test/compilable/ddoc4899.d
@@ -11,7 +11,7 @@ compilable/ddoc4899.d(16): Warning: Ddoc: Stray ')'. This may cause incorrect Dd
 
 /++
        (See accompanying file LICENSE_1_0.txt or copy at
-        http://www.boost.org/LICENSE_1_0.txt)
+        foo:)
 +/
 module d;
 
--- a/test/compilable/extra-files/ddoc17697.html
+++ b/test/compilable/extra-files/ddoc17697.html
@@ -1,0 +1,484 @@
+
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
+    <title>test1</title>
+    <style type="text/css" media="screen">
+      html, body, div, span, object, iframe, h1, h2, h3, h4, h5, h6, p,
+      blockquote, pre, a, abbr, address, cite, code, del, dfn, em, figure,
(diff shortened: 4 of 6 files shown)


WalterBright:
The long diff is because `ddoc10.html` was using CRLF for line endings, and was converted to LF. The default ddoc file is 700+ lines of boilerplate.

CyberShadow:
> The long diff is because ddoc10.html was using CRLF for line endings, and was converted to LF. The default ddoc file is 700+ lines of boilerplate.

Could you please do that conversion in a separate commit?

WalterBright:
> do that conversion in a separate commit?

https://github.com/dlang/dmd/pull/7044

test/compilable/extra-files/ddoc17697.html
```diff
@@ -0,0 +1,483 @@
...
+        color: rgba(0, 0, 0, 1);
+      }
+
+      .psymbol {
+        color: rgba(0, 0, 0, 1);
+      }
+
+      .ddoc_member_header .ddoc_header_anchor .code {
+        font-size: 1em;
+      }
```
adamdruppe: so it kinda drives me nuts that all the tests cover stuff outside what they actually care about. Maybe we should defined `DDOC=$(BODY)` in the test macros to cut out all the surrounding stuff so it just tests the specific thing.

test/compilable/ddoc17697.d
```diff
@@ -0,0 +1,15 @@
+// PERMUTE_ARGS:
+// REQUIRED_ARGS: -D -Dd${RESULTS_DIR}/compilable -o-
+// POST_SCRIPT: compilable/extra-files/ddocAny-postscript.sh 17697
+
+/***
+ * See:
+ *    http://www.fooa.com/test1
```
adamdruppe: What happens if you use ` http://www.fooa.com/_test1`? Will the underscore from the URL be preserved?

WalterBright: Yes

adamdruppe: should prolly add an explicit test case for that, since that's what the bug report is really about.

WalterBright: That is a test case for that. Without the change, `test1` would get highlighted in the midst of the URL.

adamdruppe:
On Sat, Jul 29, 2017 at 01:29:17PM -0700, Walter Bright wrote:
> That is a test case for that. Without the change, `test1` would get highlighted in the midst of the URL.

That's related, but different than what I'm talking about. (And
I agree it is very important to fix that.)

But consider that an existing link in Phobos might be

---
/// calls http://www.cplusplus.com/reference/cstdio/_fclose/
void fclose(void*);
---

The underscore there is NOT the actual link - it is there to suppress
the current highlighting behaver. What happens to that now?

Will it link to literally `http://www.cplusplus.com/reference/cstdio/_fclose/` (which is a 404), or will it still detect the underscore and just automatically remove it?

If it is the former, we need to document that as a breaking change (and get everyone to fix it - which they should! This would be a positive change), and I want the test to assert that

ddoc("http://www.cplusplus.com/reference/cstdio/_fclose/") == `http://www.cplusplus.com/reference/cstdio/fclose/`;



But then, a related question is what if the link ACTUALLY IS _fclose? How do you write that? Do you have to do `$(UNDERSCORE)fclose`? Again, that should be documented and asserted whatever the answer is.

WalterBright:
It doesn't remove the _. It does nothing with it. In fact, it does nothing with it before this PR - the _ processing did not happen for full URLs.

adamdruppe:
On Sat, Jul 29, 2017 at 10:22:19PM -0700, Walter Bright wrote:
> It doesn't remove the _. It does nothing with it. In fact, it does nothing with it before this PR - the _ processing did not happen for full URLs.

Then just add the test case to prove it. It is two lines of code and would put my mind at ease.


Here's an example from Phobos that I'm concerned about

https://github.com/dlang/phobos/blob/master/std/stdio.d#L1105


and indeed, it isn't a full url... it uses the HTTP macro (LOL wtf)
instead of the http:// proper form.

src/ddmd/doc.d
```diff
@@ -2405,6 +2412,32 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
                 }
             }
             break;
+
+        case '$':
```
CyberShadow: Would it be possible to reuse the `\xFF{` / `}` painting from dmacro.d here? This seems like it's duplicating the logic from there.

test/compilable/ddoc4899.d
```diff
@@ -4,8 +4,6 @@
 /*
 TEST_OUTPUT:
 ---
-compilable/ddoc4899.d(16): Warning: Ddoc: Stray '('. This may cause incorrect Ddoc output. Use $(LPAREN) instead for unpaired left parentheses.
-compilable/ddoc4899.d(16): Warning: Ddoc: Stray ')'. This may cause incorrect Ddoc output. Use $(RPAREN) instead for unpaired right parentheses.
```
CyberShadow: This change makes this test useless. See https://issues.dlang.org/show_bug.cgi?id=4899

This test can be replaced with:

```d
/**
(
foo:
)
*/
```

or

```d
/** ( */ int a;
/** ) */ int b;
```

test/compilable/extra-files/ddoc17697.html
CyberShadow: That's a good idea. Perhaps create a `.ddoc` file that's added to DMD's command line by the autotester, so that you don't have to add a `Macros:` section for every DDoc test.

src/ddmd/doc.d
```diff
@@ -1643,7 +1643,13 @@ struct DocComment
...
                     while (isIdTail(q))
                         q += utfStride(q);
-                    if (*q == ':') // identifier: ends it
+                    if (*q == ':' &&            // 'identifier:' ends it
+                                                // but not 'http://' or 'https://'
+                        !(q[1] == '/' && q[2] == '/' &&
+                          (q - p == 4 && Port.memicmp(p, "http".ptr, 4) == 0 ||
+                           q - p == 5 && Port.memicmp(p, "https".ptr, 5) == 0)
+                         )
+                       )
```
CyberShadow: Seems like this is partially redundant with `skippastURL`?

changelog/fix17697.ddoc
```diff
@@ -0,0 +1,23 @@
+Fix Issue 17697 - Ddoc: automatically highlight URLs outside of macro arguments
```
CyberShadow: Changelog entries must have a `.dd` file extension, not `.ddoc`.

changelog/fix17697.ddoc
```diff
@@ -0,0 +1,23 @@
...
+
+and a clickable hyperlink will be emitted to the resulting .html file.
+
+This detection does not happen if the URL is part of a macro argument:
+
+/****
+ * $(ABC http://www.fooa.com/test1)
+ */
+
+The behavior of that is unchanged. URLs can start with http:// or https://
```
CyberShadow: Perhaps wrap `http://` and `https://` in backticks or such, so that they aren't linkified :)

test/compilable/ddoc17697.d
CyberShadow: I think Adam's request is valid. There is no test that ensures that any underscores prefixing the documented symbol's identifier will be preserved in the URL.

CyberShadow (commented):
Looks good otherwise.

test/compilable/ddoc17697.d
WalterBright: Except that this PR **does not change** any of the `_` logic.

CyberShadow: Sorry, is there an existing test for that, then? (If there isn't, then this would be a good opportunity to add one.)

CyberShadow: And even if there is, I think it would still be a good idea to test the combination of these two features.

src/ddmd/doc.d
WalterBright: It is, but setting up the arguments to `skippastURL` means conversion to OutBuffer, and this string isn't in an OutBuffer.

src/ddmd/doc.d
WalterBright: The `highlightText` code is run before any macros are expanded.

WalterBright:
@CyberShadow what's going on with `CyberShadow/DAutoTest` ? It's pretty hard to tell from the log.

CyberShadow:
Here are the relevant lines:

```
Searching for undefined macros
/dev/shm/dtest/work/repo/dlang.org/web/changelog/2.076.0.html:263: * <!--UNDEFINED MACRO: "ABC"-->
/dev/shm/dtest/work/repo/dlang.org/web/changelog/2.076.0_pre.html:263: * <!--UNDEFINED MACRO: "ABC"-->
posix.mak:646: recipe for target 'test' failed
```

It looks like you need to escape the `$(ABC ...)` parts of the changelog, so that the macro isn't expanded.

Did you mean to enclose the examples in code blocks (`---` `\n` ... `\n` `---`) ?

WalterBright:
Didn't know that those files would be run through Ddoc. Fixed.

CyberShadow (approved):

andralex:
What about ftp:, mailto: and others?

CyberShadow:
Those probably don't occur frequently enough that using an explicit macro for those would be too onerous.

Expanding this to any protocol (i.e. look for `://`) wouldn't be unreasonable, however, that won't apply to `mailto:`.

PR #7044, opened 2017-07-29
convert ddoc10.html to LF line endings

WalterBright:
Don't check in code with CRLF line endings.


CyberShadow (approved):

CyberShadow:
Thanks! A separate PR wasn't necessary, but is fine too of course.

PR #7046, opened 2017-07-30
fix issue 17704 - Dont parse protocols as ddoc sections

ghost:
Effect should be visible on the newly rendered https://dlang.org/phobos/ddmd_backend_code_x86.html by the DAutoTest.


WalterBright:
Already covered by https://github.com/dlang/dmd/pull/7043 except for `ftp:`.

PR #7050, opened 2017-07-31
Fix Lexer memory access beyond the given buffer when malformed input is passed.

JohanEngelen:
Let me know what token value you want to return upon errors.

This change also enforces that the buffer is really null-terminated (or EOF), and that the terminating null character is included in the buffer interval that the Lexer should lex.
When a string is passed, the implicit null-terminator is necessary and the code must be modified:
```diff
    string text = "int"; // the test relies on this string begin null-terminated
-    scope Lexer lex1 = new Lexer(null, text.ptr, 0, text.length, 0, 0);
+    scope Lexer lex1 = new Lexer(null, text.ptr, 0, text.length+1, 0, 0);
```
Without including the null terminator in the length (the `+1`), `nextToken` would return `TOKerror` instead of `TOKeof`.

### DIFF
--- a/src/ddmd/lexer.d
+++ b/src/ddmd/lexer.d
@@ -129,8 +129,8 @@ unittest
     //printf("lexer.unittest\n");
     /* Not much here, just trying things out.
      */
-    string text = "int";
-    scope Lexer lex1 = new Lexer(null, text.ptr, 0, text.length, 0, 0);
+    string text = "int"; // We rely on the implicit null-terminator
+    scope Lexer lex1 = new Lexer(null, text.ptr, 0, text.length+1, 0, 0);
     TOK tok;
     tok = lex1.nextToken();
     //printf("tok == %s, %d, %d\n", Token::toChars(tok), tok, TOKint32);
@@ -178,10 +178,11 @@ class Lexer
 
     /*********************
      * Creates a Lexer for the source code base[begoffset..endoffset].
+     * The last character, base[endoffset-1], must be null (0) or EOF (0x1A).
      *
      * Params:
      *  filename = used for error messages
-     *  base = source code
+     *  base = source code, must be terminated by a null (0) or EOF (0x1A) character
      *  begoffset = starting offset into base[]
      *  endoffset = one past the last offset to read into base[]
      *  doDocComment = handle documentation comments
@@ -268,14 +269,16 @@ class Lexer
         Loc startLoc;
         t.blockComment = null;
         t.lineComment = null;
+
+        // Return EOF token when end of buffer is already reached
+        if (p >= end) {
+            t.value = TOKeof;
+            t.loc = loc();
+            return;
+        }
+
         while (1)
         {
-            // Return error token when end of buffer is reached unexpectedly.
-            if (p >= end) {
-                t.value = TOKeof;
-                return;
-            }
-
             t.ptr = p;
             //printf("p = %p, *p = '%c'\n",p,*p);
             t.loc = loc();
--- a/src/ddmd/lexer.d
+++ b/src/ddmd/lexer.d
@@ -145,16 +145,33 @@ unittest
 
 unittest
 {
-    // Test malformed input
-    char[2] text = ['\'', 0];
-    scope Lexer lex2 = new Lexer(null, text.ptr, 0, text.length-1, 0, 0);
-    TOK tok;
-    tok = lex2.nextToken();
-    assert(tok == TOKcharv);
-    tok = lex2.nextToken();
-    assert(tok == TOKeof);
-    tok = lex2.nextToken();
-    assert(tok == TOKeof);
+    // Test malformed input: even malformed input should end in a TOKeof.
+    static immutable char[][] testcases =
+    [   // Testcase must end with 0 or 0x1A.
+        [0], // not malformed, but pathological
+        ['\'', 0],
+        ['\'', 0x1A],
+        ['{', '{', 'q', '{', 0],
+        [0xFF, 0],
+        [0xFF, 0x80, 0],
+        [0xFF, 0xFF, 0],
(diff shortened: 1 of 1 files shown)


JohanEngelen:
Actually, I think the required null-terminator is a misfeature. Why specify buffer length at all in that case? Without the null-terminator requirement, you can nicely lex just a part of a longer code string.
So instead of returning TOKerror upon buffer end, perhaps we should return TOKeof ?

JohanEngelen:
I see that `TOKeof` is returned in other locations where an error occurs, but also end of buffer is similar as EOF.
I've removed the null-terminated requirement, as that is taken care of by the endoffset.

src/ddmd/lexer.d
```diff
@@ -253,6 +270,12 @@ class Lexer
         t.lineComment = null;
         while (1)
         {
+            // Return error token when end of buffer is reached unexpectedly.
+            if (p >= end) {
```
WalterBright: Lexing performance is critical to compiler speed, and this adds a compare/branch for every character processed. This is why a comparison to the buffer end is not done, and a sentinal (0 termination) is done. The sentinel does not require any extra code in the main path.

JohanEngelen: Some extra checking is needed, perhaps elsewhere, but currently the code reads beyond the buffer whenever one of the sub-functions (e.g. `hexStringConstant`) stumbles upon the null terminator. For example with input `"\'\0"` and with `"{{q{{\0"`. Null-termination is not preventing the buffer overflow read with the current implementation.

(the performance hit of this comparison+branch is probably very low, because it's almost always false and thus easily predicted correctly)

UplinkCoder: not every processor has nice branch prediction.
add a sentinel if you must.

WalterBright: > Null-termination is not preventing the buffer overflow read with the current implementation.

A test case demonstrating buffer overflow would be apropos. In any case, perhaps the error check should go into `hexStringConstant` instead of the main loop.

> probably very low

My testing showed that every instruction per character matters.

WalterBright: BTW, dmd's compile speed has slowed down substantially in the recent past, due to an accumulation of barnacles. It's now slower than dmc++.

JohanEngelen: Of course the PR adds a testcase.... from commit 1.....

JohanEngelen:
The problem is caused by the user not knowing when the file/buffer ended, when it ends on an incomplete string literal/char literal/comment/...  The last token returned by `nextToken` would be for example TOKstring (for incomplete string literal), and then when the user calls `nextToken` again, a buffer overread occurs.
I have now moved the condition out of the loop, so it is performed once for each call to `scan`.
Fuzzing has thusfar not revealed new bugs.

The requirement of the `endoffset` parameter to be one beyond the last valid index is up for debate. That's what the original intent of `endoffset` was, judging from the code, but it does mean that the example of how to use Lexer needs to be modified with a `+1` to include the implicit string literal null-terminator (couldn't find it in the spec that the null terminator is mandatory...)

JohanEngelen:
An other more fragile solution is to move the pointer `p` back by one upon encountering e.g. an unterminated string constant (need to fix all locations where that happens).

WalterBright:
> a buffer overread occurs

Please post an example. I'm not seeing it.

JohanEngelen:
>  For example with input "\'\0" and with "{{q{{\0"

JohanEngelen:
Note: `mars.d` line 1351 sets a buffer without including the null terminator in the length of that buffer.

JohanEngelen:
Bad stuff happens here for example: https://github.com/dlang/dmd/blob/28daf534a405e9a07d42afb8f978d8746be7b89a/src/ddmd/lexer.d#L1239-L1255

A character is read from `*p` and `p` is immediately increased. Then, if the char read is the null terminator, we return but after that, `p` is pointing at an invalid memory location. The next time we call `scan`, a buffer overread occurs. The user cannot know about this, because the returned token is `TOKstring`, while the user is looping until he sees `TOKeof`.

WalterBright:
> Note: mars.d line 1351 sets a buffer without including the null terminator in the length of that buffer.

The Lexer relies on the 0 being past the end of the length, i.e. that `buf[length]` is `0`. But there was another bug there, corrected here:

https://github.com/dlang/dmd/pull/7057

JohanEngelen:
> The Lexer relies on the 0 being past the end of the length

The current code is ambiguous to what the length means. In one spot, it code says that the endoffset (i.e. the length passed in #7057) is the last valid index in the buffer, but at the same time it is assigned to the `end` field that points one past the buffer. 
I'll clarify and update the PR according to that the terminator is supposed to be at `base[endoffset]`, and thus that that's a valid memory access.
In the meanwhile, I've found another buffer over-read related to UTF8 chars.
I've cooked up a different fix, focussing on pre/post conditions that would remove the need for the `p > end` boundscheck. Work in progress.

JohanEngelen:
OK, I completely reworked the patch. The boundscheck is gone, and instead I fixed the code paths that advanced `p` beyond the terminator / end of the buffer.
This fixes the other access violation not caught by the earlier patches.

UplinkCoder:
that looks better.

JohanEngelen:
> that looks better.

I disagree, much more prone to bugs in the future. But you can have it your way.

JohanEngelen:
ping

WalterBright:
> This fixes the other access violation not caught by the earlier patches.

Please include a test case for both access violations.

JohanEngelen:
@WalterBright I've added a few more testcases.

JohanEngelen:
Thanks. Further fuzzing has not turned up anything (yet? ;-).

PR #7053, opened 2017-08-02
Fix Issue 17684 - [REG 2.062] static alias this

JinShil:


JinShil:
Well, that didn't work.

PR #7055, opened 2017-08-02
Fix Issue 17684 - [REG 2.062] static alias this

JinShil:


ZombineDev:
It looks like the Jenkins failures was due to network issues, so I've restarted the test pipeline there.

test/runnable/test17684.d
```diff
@@ -0,0 +1,12 @@
...
+    static int value;
+    static alias value this;
+}
+
+void main()
+{
+    StaticStruct = 42;
+    immutable int a = StaticStruct;
+    assert(a == 42);
+}
```
ZombineDev: Missing final end of line. Consider setting up [EditorConfig](http://editorconfig.org) on your text editor, so that repo text editor settings - https://github.com/dlang/dmd/blob/master/.editorconfig - are automatically applied.

PetarKirov (approved):
Looks good to me on a first glance, though expect more thorough review from the dmd devs.

test/runnable/test17684.d
JinShil: Done.  Thanks for the review.

MartinNowak:
Weird language feature, but rather have it fully-working that half-working. Any specific use-case in mind?

JinShil:
>  Any specific use-case in mind?

Yes, it's for my [memory-mapped-io library](https://github.com/JinShil/memory_mapped_io) where all registers and bitfields are modeled as static entities.  I currently have to put a `.value` any time I use a register or bitfield, which is verbose, redundant, and monotonous especially given the number of times registers must be used in a typical embedded board board support package (BSP).  `static alias this` allows me to write `int baudRate = UART.BR;` instead of `int baudRate = UART.BR.value;`, making the code more pleasant to use and read.  

You'd probably question the pattern I've used in that library, but it was chosen due to other limitations in the D language (e.g. modules-must-be-files and a few others).  It would take a detailed blog post to explain it, so if you really want to know, contact me offline.  Long story short...it wasn't arrived at thoughtlessly.

PR #7057, opened 2017-08-02
fix using sizeof pointer instead of sizeof buf

WalterBright:
This problem was apparently introduced by the conversion from C++ to D. The `.sizeof` should have been the length of the string, but instead it was the size of a pointer. The code didn't depend on the length being correct, but this should be fixed anyway.


PetarKirov (approved):
LGTM

wilzbach (approved):

PR #7059, opened 2017-08-03
fix travis-ci linux 64-bit tests

MartinNowak:
- the dmd/tests are using a shared phobos library but did link against
  the one that comes with the host compiler (installed by
  travis/install.sh) due to LD_LIBRARY_PATH being set
- wasn't noticed before cause dmd tests hardly rely on phobos and the
  parts used where ABI stable enough with the last release
- didn't happen for gdc/ldc with their gphobos/phobos-ldc libs

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,9 @@ d:
   - gdc
   - ldc
 
-script: ./travis.sh
+script:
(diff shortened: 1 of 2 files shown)


PetarKirov (approved):
Looks good, thanks for tracking this down.

PR #7060, opened 2017-08-04
Implement visitor for dsymbol semantic

RazvanN7:
In order to avoid conflicts on the semantic function, I created a file named semantic.d where the semantic entrypoints for the visitor classes are located. That way, if you want to use any kind of semantic you will just import semantic.d, thus eliminating the need for more imports (e.g. : import initsem; import typesem; etc. => import semantic.d)


WalterBright:
This is a pretty big gulp to swallow, while the other big changes are still being digested (much rebasing remains to be done).

WalterBright:
Upon reflection, I don't see the value in separating Dsymbol's semantic() routines out. Dsymbols are not ASTs, and they don't even exist until the semantic() routines are run from the ASTs.

RazvanN7:
@WalterBright Separating all the semantic routines enables us to have "physical" separation between parsing and semantic.

>Dsymbols are not ASTs, and they don't even exist until the semantic() routines are run from the ASTs.

That is not entirely true. VarDeclarations are Dsymbols and they do exist before semantic.

RazvanN7:
@WalterBright The classes defined in dsymbol.d implement semantic routines, but those are empty. Instead, there are a lot of AST nodes that inherit from Dsymbol and are present during parsing (Declarations, FuncDeclarations, Imports etc.). This PR eliminates all the semantic routines for all the classes that inherit Dsymbol (directly or indirectly). This is necessary if we want to eliminate all semantic related methods from the AST nodes.

It is true that eliminating the semantic routines from dsymbol.d doesn't offer too many benefits, but having empty methods which will never be overwritten is useless.

WalterBright:
You make a good point, but I still want to wait a bit, as the last upheavals are still working their way through the PRs.

wilzbach:
> but I still want to wait a bit, as the last upheavals are still working their way through the PRs.

Isn't this an argument for doing the change now, so that the PRs only need to be reworked once and not twice or even more often?

src/ddmd/semantic.d
```diff
@@ -0,0 +1,50 @@
...
+import ddmd.init;
+import ddmd.initsem;
+import ddmd.mtype;
+import ddmd.statement;
+import ddmd.statementsem;
+
+/*************************************
+ * Does semantic analysis on the public face of declarations.
+ */
+extern(C++) void semantic(Dsymbol dsym, Scope* sc)
```
jacob-carlborg: This looks like a new function. Please use common Ddoc sections like summary, description and parameters.

RazvanN7:
@WalterBright I think it's high time we move on with this. Since it will eventually be merged I think the sooner the better.

RazvanN7:
Since the merging of this was stalled, I made another commit with the semantic routines from mtype.d. When this PR is going to be merged everyone will have to rebase, so I figured that it's better to merge them altogether rather than cause multiple rebasing procedures.

PR #7061, opened 2017-08-05
add Spec links to expression.d

WalterBright:
Add some direct links to the Spec corresponding to `Expression` nodes.

### DIFF
--- a/src/ddmd/expression.d
+++ b/src/ddmd/expression.d
@@ -2683,6 +2683,7 @@ enum WANTvalue  = 0;    // default
 enum WANTexpand = 1;    // expand const/immutable variables if possible
 
 /***********************************************************
+ * http://dlang.org/spec/expression.html#expression
  */
 extern (C++) abstract class Expression : RootObject
 {
@@ -3997,6 +3998,7 @@ extern (C++) final class DsymbolExp : Expression
 }
 
 /***********************************************************
+ * http://dlang.org/spec/expression.html#this
  */
 extern (C++) class ThisExp : Expression
 {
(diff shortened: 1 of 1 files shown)


rainers:
As these are Ddoc comments they show up in the generated documentation, but pretty badly formatted, e.g. http://dtest.dlang.io/artifact/website-cd3a65fbdfa0ddac41d309455d4d0930aae7f507-21a8c51d507707f7cc0c3ed83ac6ef26/web/library-prerelease/ddmd/expression/add_exp.html

WalterBright:
> badly formatted

That was fixed by https://github.com/dlang/dmd/pull/7043

wilzbach:
> That was fixed by #7043

Only for Ddoc, not Ddox.
It might help adoption of a Ddoc change to all the other documentation engines if it's accompanied by a spec change...

rainers:
> That was fixed by #7043

It doesn't show up in the documentation, e.g. http://dtest.dlang.io/artifact/website-3bb580a55554d33712e17e39578e8b7f7164c90d-8057772f27d802cf0c39fec5774f95d6/web/phobos/ddmd_expression.html
Is that the fix or is the build outdated?

WalterBright:
> It doesn't show up in the documentation

If the documentation isn't built by master, than it wouldn't show up.

WalterBright:
> a spec change

https://github.com/dlang/dlang.org/pull/1858

rainers:
> If the documentation isn't built by master, than it wouldn't show up.

Unfortunately, the logs are not accessible if the build succeeds, so can't check that. I've verified locally that it works, though.

rainers:
It's a bit ugly though, as the comments end up on a single line, e.g.
```
econd ? e1 : e2 http://dlang.org/spec/expression.html#conditional_expressions TOKquestion 
```

WalterBright:
> It's a bit ugly though, as the comments end up on a single line, e.g.

Fixed

PR #7063, opened 2017-08-05
fix building via VS projects

rainers:
adds new files to VS project

### DIFF
--- a/src/vcbuild/dmd.visualdproj
+++ b/src/vcbuild/dmd.visualdproj
@@ -498,6 +498,7 @@
    <File path="..\ddmd\errors.d" />
    <File path="..\ddmd\escape.d" />
    <File path="..\ddmd\expression.d" />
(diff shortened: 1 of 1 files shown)


wilzbach:
Maybe it's time to auto-generate this file or does it need to be checked in into git, s.t. people on Windows only need to git clone DMD and can then import it into visual studio?

wilzbach (approved):

rainers:
> Maybe it's time to auto-generate this file or does it need to be checked in into git, s.t. people on Windows only need to git clone DMD and can then import it into visual studio?

I guess it would slightly defy its purpose if you have to run make or a batch before loading the VS solution. I see two other options:
- include a build step as part of the solution that checks the current project and regenerates it if needed.
- use the VC project/MSBuild integration of Visual D 0.44 that also allows using wild cards for filenames.

PR #7064, opened 2017-08-05
Make more Phobos math primitives CTFE-able

kinke:
These, in combination with a [CTFE-able std.math.exp2](https://github.com/ldc-developers/phobos/commit/823fc05bffbe614d9373b966b91b48d3189b5649) implementation, make the [most common math functions CTFE-able for LDC](https://github.com/ldc-developers/ldc/pull/2259#issuecomment-320461243), incl. `pow()` and the `^^` operator, [DMD issue #16474](https://issues.dlang.org/show_bug.cgi?id=16474).

[Some more builtins or CTFE-friendly Phobos implementations may still be needed for DMD. LDC's Phobos diverges where it can use LLVM intrinsics. These intrinsics are detected as CTFE builtins and have appropriate implementations in `CTFloat`.]

### DIFF
--- a/src/ddmd/builtin.d
+++ b/src/ddmd/builtin.d
@@ -101,6 +101,28 @@ extern (C++) Expression eval_ldexp(Loc loc, FuncDeclaration fd, Expressions* arg
     return new RealExp(loc, CTFloat.ldexp(arg0.toReal(), cast(int) arg1.toInteger()), arg0.type);
 }
 
+extern (C++) Expression eval_isnan(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOKfloat64);
+    return new IntegerExp(loc, CTFloat.isNaN(arg0.toReal()), Type.tbool);
+}
+
+extern (C++) Expression eval_isinfinity(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOKfloat64);
+    return new IntegerExp(loc, CTFloat.isInfinity(arg0.toReal()), Type.tbool);
+}
+
+extern (C++) Expression eval_isfinite(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOKfloat64);
+    const value = !CTFloat.isNaN(arg0.toReal()) && !CTFloat.isInfinity(arg0.toReal());
+    return new IntegerExp(loc, value, Type.tbool);
+}
+
 extern (C++) Expression eval_bsf(Loc loc, FuncDeclaration fd, Expressions* arguments)
 {
     Expression arg0 = (*arguments)[0];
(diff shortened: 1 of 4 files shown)


kinke:
CTFE-able `std.math.exp2()`: https://github.com/dlang/phobos/pull/5672

src/ddmd/builtin.d
```diff
@@ -266,6 +297,24 @@ public extern (C++) void builtin_init()
...
+    // @trusted @nogc pure nothrow bool function(T)
+    add_builtin("_D3std4math12__T5isNaNTeZ5isNaNFNaNbNiNeeZb", &eval_isnan);
+    add_builtin("_D3std4math12__T5isNaNTdZ5isNaNFNaNbNiNedZb", &eval_isnan);
+    add_builtin("_D3std4math12__T5isNaNTfZ5isNaNFNaNbNiNefZb", &eval_isnan);
+    add_builtin("_D3std4math18__T10isInfinityTeZ10isInfinityFNaNbNiNeeZb", &eval_isinfinity);
+    add_builtin("_D3std4math18__T10isInfinityTdZ10isInfinityFNaNbNiNedZb", &eval_isinfinity);
+    add_builtin("_D3std4math18__T10isInfinityTfZ10isInfinityFNaNbNiNefZb", &eval_isinfinity);
+    add_builtin("_D3std4math15__T8isFiniteTeZ8isFiniteFNaNbNiNeeZb", &eval_isfinite);
+    add_builtin("_D3std4math15__T8isFiniteTdZ8isFiniteFNaNbNiNedZb", &eval_isfinite);
+    add_builtin("_D3std4math15__T8isFiniteTfZ8isFiniteFNaNbNiNefZb", &eval_isfinite);
```
ibuclaw: Could we add a callback like `add_ctfefunction`?  These are *not* compiler built-ins, and they should be distinguished.

ibuclaw: The same also applies for the likes of `tan` and `y2lx`.

kinke: Not just those, none of the various Phobos functions here are compiler built-ins, and some don't even exist anymore (e.g., `std.math.yl2x[p1]`). So I consider this out of the scope of this PR. [I've been associating `ddmd.builtin` with *CTFE* built-ins anyway.]

ibuclaw: From what I recall, they still are intrinsics, they were just given function bodies so you can take the address of them.

kinke: I just grepped the DMD src, and `ddmd.builtin` is exclusively used by the CTFE interpreter and initialized by `ddmd.mars`. Comment for `isBuiltin()`: `Determine if function is a builtin one that we can evaluate at compile time`.

WalterBright:
Looks good, but needs test cases to turn the reds to greens.

kinke:
Test amended to 2nd commit.

WalterBright:
I hate to nitpick, but the test cases do not do a smoke test that the value computed is correct. It just tests to see that a value was computed.

kinke:
As there's a lot more coming, I don't really want to smoke-test them all, especially after having added 4 builtins only. Final smoke tests should be done in Phobos IMO, having a bunch of meaningful tests already which can probably be easily extended to cover CTFE as well.

WalterBright:
The smoke test ensures things like `cosine` is not called instead of `sine`. (Those sorts of things happen when there are multiple nearly identical functions, which is the case here.) Relying on tests in Phobos does not cover that.

I can't tell you how many times I have been tempted to omit unit tests because the code was obviously correct, only to find out that indeed the code called `cosine` instead of `sine`, or added when it should have subtracted, etc.

kinke:
> Relying on tests in Phobos does not cover that.

Not at the moment, that's correct, I meant adding new CTFE tests for the newly CTFE-able functions to Phobos after these and potentially other CTFE-additions make it into the compiler first.
I quickly grepped through std.math and could only find 2 CTFE tests, one [testing the sqrt() CTFE availability](https://github.com/dlang/phobos/blob/master/std/math.d#L1601-L1604) and a [poly smoketest](https://github.com/dlang/phobos/blob/v2.075.0/std/math.d#L7122), I was expecting (a lot) more, but oh well.

I'll add a bunch of smoketests for the 4 new builtins here.

kinke:
Test updated, now smoketesting the compiler's CTFE builtins for `std.math`, including the 4 new ones.

WalterBright:
> I quickly grepped through std.math and could only find 2 CTFE tests, one testing the sqrt() CTFE availability and a poly smoketest, I was expecting (a lot) more, but oh well.

std.math unittests should not be testing compiler CTFE.

std.math should test the correctness of sqrt's implementation. The compiler tests should test that the arguments get correctly packaged up and sent to sqrt and that the expected result comes back from it. Only one test is needed for that, i.e. a smoke test.

> I'll add a bunch of smoketests for the 4 new builtins here.

Thanks!

WalterBright:
Sorry, pushed the wrong button!

kinke:
Hmm the `static assert(isInfinity(real.infinity))` is failing. The CTFE implementation is [here](https://github.com/dlang/dmd/blob/master/src/ddmd/root/ctfloat.d#L110). So my best guess is that the CTFE `is` compares the padding bytes too.

ibuclaw:
It shouldn't do, many other tests would fail too if that were the case.

Try inspecting in a debugger first.

kinke:
You had me built dmd the first time ;). I simply dumped the values in `CTFloat::isInfinity()`; the problem seems to be the padding bytes indeed, the 10 lower bytes are identical. Switching from `is` to `CTFloat::isIdentical()` with its explicitly limited memcmp() gets it to work.

PR #7065, opened 2017-08-06
fix Issue 17720 - Wrong code using vector extensions with different types

ibuclaw:


PetarKirov (approved):
Looks good to me.

ibuclaw:
Hmm, I did not think this one entirely through.  Fixed it up with more extensive test, however it looks like DMD's `__simd` depends on implicitly being able to convert `void16` any other vector type of the same size.

Not sure if I should bite my lip and allow this.  Maybe @WalterBright might want to chime in as its his intrinsic.

WalterBright:
It's failing the autotester because of `checkwhitespace`.

I'm not sure I totally understand your comment, but you can convert an array of ints to an array of void and then to an array of float, and you'll get ints painted to floats. This is allowed behavior.

WalterBright:
When fixing a bugzilla issue, please put the PR URL in the bugzilla issue, so somebody else doesn't waste time trying to fix it because he's ignorant there is an extant fix already. I already did it for this one.

ibuclaw:
I'm not sure I totally understand your comment, but you can convert an array of ints to an array of void and then to an array of float, and you'll get ints painted to floats. This is allowed behavior.

ibuclaw:
You can see the error I'm talking about in the autotester now.

ibuclaw:
@TurkeyMan - maybe you could give a short sentence on whether allowing implicit conversion from void vector to any other vector type seems reasonable.

TurkeyMan:
How would you argue in favour?
Practically it can be convenient in a lot of typical simd code, but I kinda feel like it's antithetical to D in general. It's certainly not `@safe` ;)
I reckon it'd be better to put a bunch of overloads/wrappers in `core.simd` to match return types to the argument types like most C++ simd headers do?

ibuclaw:
Added overrides in  dlang/druntime#1900 - however using one of them exposes an ICE.

ibuclaw:
Blocked by druntime and dmd bug for the moment (this patch is already in gdc, so I have a vested interest to see that someone fixes it).

ibuclaw:
Ping, GDC has had this fix for over a year now and DMD is still blocked by their own simd intrinsics being in violation of this constraint.

ibuclaw:
What's the way forward here?  Turn dmd simd intrinsics into templates?

WalterBright (approved):

wilzbach:
Rebased as the respective druntime PR has been merged.

wilzbach (approved):

wilzbach:
Looks like druntime/phobos requires more work:

```
../druntime/import/core/internal/arrayop.d(102): Error: cannot implicitly convert expression __simd(cast(XMM)15929199, cast(__vector(void[16]))*cast(const(__vector(ulong[2])*))p) of type __vector(void[16]) to __vector(ulong[2])
../druntime/import/core/internal/arrayop.d(37): Error: template instance `core.internal.arrayop.load!(ulong, 2LU)` error instantiating
../druntime/import/object.d(4321):        instantiated from here: arrayOp!(ulong[], const(ulong)[], "u~", "=")
std/bitmanip.d(1816):        instantiated from here: _arrayOp!(ulong[], const(ulong)[], "u~", "=")
../druntime/import/core/internal/arrayop.d(102): Error: cannot implicitly convert expression __simd(cast(XMM)15929199, cast(__vector(void[16]))*cast(const(__vector(ulong[2])*))p) of type __vector(void[16]) to __vector(ulong[2])
../druntime/import/core/internal/arrayop.d(37): Error: template instance `core.internal.arrayop.load!(ulong, 2LU)` error instantiating
../druntime/import/object.d(4321):        instantiated from here: arrayOp!(ulong[], const(ulong)[], "u~", "=")
std/bitmanip.d(1816):        instantiated from here: _arrayOp!(ulong[], const(ulong)[], "u~", "=")
```

ZombineDev:
All the issue seem to come from `core.internal.arrayop`.

ibuclaw:
This may also fix https://issues.dlang.org/show_bug.cgi?id=19629

ibuclaw:
All checks pass, I'll leave it up to the RM @thewilsonator to decide if any action is required for forcing casts when using __simd().

src/dmd/mtype.d
```diff
@@ -3521,8 +3521,22 @@ extern (C++) final class TypeVector : Type
         //printf("TypeVector::implicitConvTo(%s) from %s\n", to.toChars(), toChars());
         if (this == to)
             return MATCH.exact;
-        if (ty == to.ty)
-            return MATCH.convert;
+        if (to.ty == Tvector)
```
thewilsonator, line 3524: this could be an early return.

ibuclaw:
Added changelog entry.

ibuclaw:
Github UI doesn't seem to be updating to reveal the new commit...

NilsLankila:
Why dont you choose to start a deprecation ?

ibuclaw:
> Why dont you choose to start a deprecation ?

GDC has been running with this for as long as the bug report has been open.

Actually I'd like to make this even more restricted. Implicit casting between shared and non-shared is still allowed, but shouldn't be. However that will really begin to break code, so will need a deprecation.

This just makes an error code that is broken anyway.

NilsLankila:
So the problem is just that the PR has been forgotten but it's good to merge, right ?

ibuclaw:
> So the problem is just that the PR has been forgotten but it's good to merge, right ?

Pretty much. The only blocking part was the dmd intrinsic `__simd`, which has the signature `void16(void16, void16)`.  This is a dumb intrinsic, and new overrides should be added to deal with both allowing AVX instructions and returning a result other than void (eg: float4)

NilsLankila:
@ibuclaw,  rebasing required to pass buildkite.

PR #7066, opened 2017-08-06
fix Issue 17721 - Wrong expression type using vector extensions with shift operands

ibuclaw:
GCC supports many kinds of [vector extensions](https://gcc.gnu.org/onlinedocs/gcc/Vector-Extensions.html) that could also be supported, but aren't handled by the semantic in D.

i.e:
```
int4 f1 = [1,2,3,4];
int4 f2 = [5,6,7,8];
int4 f3 = f1 << f2;      // [32, 128, 384, 1024]
```

This covers the following operations to return a type vector:
- vector << vector  (integral only)
- vector >> vector  (integral only)
- vector >>> vector (integral only)

DMD doesn't support these, but is this OK from a language POV? -> Needs discussion.

~Could add support for vector `!`, `&&` and `||` also.~

--------
Update:  After some discussion, it seems best that all logical and comparison operators should not be included in the language.

I've removed semantic support for all except shift operators (on x86_64 these can be compiled down to `vpsravd` and `vpsllvd` respectively).

### DIFF
--- a/src/ddmd/expressionsem.d
+++ b/src/ddmd/expressionsem.d
@@ -3329,7 +3329,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         }
         if (shift)
         {
-            exp.e2 = exp.e2.castTo(sc, Type.tshiftcnt);
+            if (exp.e2.type.toBasetype().ty != Tvector)
+                exp.e2 = exp.e2.castTo(sc, Type.tshiftcnt);
         }
 
         if (!Target.isVectorOpSupported(exp.type.toBasetype(), exp.op, exp.e2.type.toBasetype()))
@@ -7659,7 +7660,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             return;
         }
         exp.e1 = integralPromotions(exp.e1, sc);
-        exp.e2 = exp.e2.castTo(sc, Type.tshiftcnt);
+        if (exp.e2.type.toBasetype().ty != Tvector)
+            exp.e2 = exp.e2.castTo(sc, Type.tshiftcnt);
 
         exp.type = exp.e1.type;
         result = exp;
@@ -7696,7 +7698,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             return;
         }
         exp.e1 = integralPromotions(exp.e1, sc);
-        exp.e2 = exp.e2.castTo(sc, Type.tshiftcnt);
+        if (exp.e2.type.toBasetype().ty != Tvector)
+            exp.e2 = exp.e2.castTo(sc, Type.tshiftcnt);
 
         exp.type = exp.e1.type;
         result = exp;
@@ -7733,7 +7736,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             return;
         }
         exp.e1 = integralPromotions(exp.e1, sc);
-        exp.e2 = exp.e2.castTo(sc, Type.tshiftcnt);
+        if (exp.e2.type.toBasetype().ty != Tvector)
+            exp.e2 = exp.e2.castTo(sc, Type.tshiftcnt);
 
         exp.type = exp.e1.type;
         result = exp;


WalterBright:
When fixing bugzilla issues, please put a link to the PR in the bugzilla issue. I did it for you for this one.

WalterBright:
This needs a spec change. The spec must say what the type of the result of an expression is.

ibuclaw:
> When fixing bugzilla issues, please put a link to the PR in the bugzilla issue. I did it for you for this one.

Ah, thanks. I thought that dlang bot was able to manage this.  Perhaps I expected too much from it.

ibuclaw:
> This needs a spec change. The spec must say what the type of the result of an expression is.

Yes, I think its best to write up the spec once the correct behaviour has been agreed.

Is this an appropriate extension to add? Or should it be instead handled by a new intrinsic in `core.simd` to keep it away from the current defined semantic of "all conditions must be convertible to bool".

TurkeyMan:
Yeah... see I *really* want to think this is a good idea. It's totally cute, and it makes me feel warm and fuzzy as a programmer... but I can't help but find that defining logical operators for vectors is just kinda wrong if I'm honest with myself.

Certainly, as with Iain's implementation, the only useful result would be defined something like "a vector of integer elements with same bit width as the source vectors, set to 0 or -1 for true/false element-wise"... which is a pretty terrible definition!! I mean, I think it's as useful as it can be, but it feels very low-level, certainly not at the proper level for high-level expression definitions. A more reasonable expectation might be a result of `__vector(bool[N])`, but then the element width is unclear, although the hardware implies something very specific (and different depends on source argument type).

As nice and succinct as it might make some vector code, I've been struggling to convince myself (for 10 years or so) that benefit balances the fact that we would now be defining logical operations where the result is not bool, or even really bool-like. Do people expect that such logical operations should be usable in if-statements, or are we happy that this code should probably error:
```d
if (vec1 < vec2) { ... } // error: __vector(int[4]) is not a valid conditional type
```

??

ibuclaw:
> As nice and succinct as it might make some vector code, I've been struggling to convince myself (for 10 years or so) that benefit balances the fact that we would now be defining logical operations where the result is not bool, or even really bool-like. Do people expect that such logical operations should be usable in if-statements, or are we happy that this code should probably error:
>
> if (vec1 < vec2) { ... } // error: __vector(int[4]) is not a valid conditional type

You can already do this for static arrays.

```
float[4] arr1 = [1,2,3,4];
float[4] arr2 = [5,6,7,8];

if (arr1 < arr2) { ... } // if (_adCmp (arr1, arr2, typeid(float[])) < 0)
```

TurkeyMan:
O_o
Wtf? That looks like a strcmp... that specification is _definitely_ not simd friendly!
It's also very-not anything like the semantics we're proposing here... note that _adCmp() returns a single bool ;)

ibuclaw:
So, shall we move this to druntime then?

TurkeyMan:
You mean `vmask_lt`/`gt`/`eq`/etc... functions in druntime? I think that should definitely exist, along with a bunch of other totally useful vector primitives (select, permute, etc).  
Whether we move forward with language/operator support... I mean, a big part of me wants to agree it's really cool, but there's the part of me that just doesn't feel it's right. The definition doesn't quite match established language/type semantics. It's a bit of a shoe-horn, and that usually just means edge-cases.
At very least, I feel like the same semantics have to equally apply to non-simd arrays, and when you approach the problem from that perspective, it starts to feel problematic.

TurkeyMan:
So, here's another thought...
We have operators `<`, `>`, `==`, etc...
The trouble with comparisons in SIMD, is that there's 3 distinct, equally common *kinds* of comparison: 'all', 'any', and 'mask'.
They're all useful, and it's not clear which should have the privileged status of being assigned to the operators.
all/any comparisons result in a single bool, which is compatible with the existing language semantics that comparisons produce a bool result (useful in conditional statements), and the mask one is kinda the simd special-case.
I feel like the intuitive result of `v1 == v2` would be the same as `arr1 == arr2`, that is, 'all equal', seems simple enough, but considering operator `<`, it gets weird.
Array comparisons do an effective strcmp, that is, it compares each element in sequence for sorting purposes... that's nothing like any of our 3 useful simd definitions. So where does that take us? Does that lead to a conclusion that vectors and arrays are fundamentally different and should have different semantics with respect to the operators? Or does it just imply that all 3 simd comparison styles should be functions rather than operators, since the operator semantics are (already) defined differently?

ibuclaw:
For now, I've removed logical operators from this, keeping only shift left/right.  Both of which should be less controversial.  The spec says: `The result type is the type of the left operand after the promotions.`

Just giving a quick test, these can be compiled down to one instruction on x86_64 also (`vpsravd` and `vpsllvd` for `>>` and `<<` respectively).

TurkeyMan:
/agree, shifts are uncontroversial.
Does it correctly distinguish arithmetic vs logical shift right for signed/unsigned elements? Does `>>>` work?

ibuclaw:
The distinction is handled by the backend.  GDC handles this fine, but of course I can always check again.

PR #7067, opened 2017-08-06
fix Issue 17722 - Wrong diagnostic using __traits(compiles, e1 && e2) expressions

ibuclaw:
Manipulates `e1` and `e2` in a local `e1x` and `e2x` var respectively, and assigns them back if semantic was successful.

This is the same as what other binary expressions do via `binSemantic`/`binSemanticProp`.

### DIFF
--- a/src/ddmd/expressionsem.d
+++ b/src/ddmd/expressionsem.d
@@ -7924,59 +7924,64 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         exp.setNoderefOperands();
 
         // same as for AndAnd
-        exp.e1 = exp.e1.semantic(sc);
-        exp.e1 = resolveProperties(sc, exp.e1);
-        exp.e1 = exp.e1.toBoolean(sc);
+        Expression e1x = exp.e1.semantic(sc);
+        e1x = resolveProperties(sc, e1x);
+        e1x = e1x.toBoolean(sc);
         uint cs1 = sc.callSuper;
 
         if (sc.flags & SCOPEcondition)
         {
             /* If in static if, don't evaluate e2 if we don't have to.
              */
-            exp.e1 = exp.e1.optimize(WANTvalue);
-            if (exp.e1.isBool(true))
+            e1x = e1x.optimize(WANTvalue);
+            if (e1x.isBool(true))
             {
                 result = new IntegerExp(exp.loc, 1, Type.tbool);
                 return;
             }
         }
 
-        exp.e2 = exp.e2.semantic(sc);
+        Expression e2x = exp.e2.semantic(sc);
         sc.mergeCallSuper(exp.loc, cs1);
-        exp.e2 = resolveProperties(sc, exp.e2);
+        e2x = resolveProperties(sc, e2x);
 
-        auto f1 = checkNonAssignmentArrayOp(exp.e1);
-        auto f2 = checkNonAssignmentArrayOp(exp.e2);
+        auto f1 = checkNonAssignmentArrayOp(e1x);
+        auto f2 = checkNonAssignmentArrayOp(e2x);
         if (f1 || f2)
         {
             result = new ErrorExp();
             return;
         }
 
-        if (exp.e2.type.ty == Tvoid)
-            exp.type = Type.tvoid;
-        else
-        {
-            exp.e2 = exp.e2.toBoolean(sc);
-            exp.type = Type.tbool;
-        }
-        if (exp.e2.op == TOKtype || exp.e2.op == TOKscope)
+        // Unless the right operand is 'void', the expression is converted to 'bool'.
+        if (e2x.type.ty != Tvoid)
+            e2x = e2x.toBoolean(sc);
+
+        if (e2x.op == TOKtype || e2x.op == TOKscope)
         {
             exp.error("%s is not an expression", exp.e2.toChars());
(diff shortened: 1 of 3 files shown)


src/ddmd/expressionsem.d
```diff
@@ -7924,59 +7924,61 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+        auto f2 = checkNonAssignmentArrayOp(e2x);
         if (f1 || f2)
         {
             result = new ErrorExp();
             return;
         }
 
-        if (exp.e2.type.ty == Tvoid)
+        if (e2x.type.ty == Tvoid)
             exp.type = Type.tvoid;
```
WalterBright: By assigning to `exp.type` here, the type gets disconnected from the operands, putting `exp` into an invalid state. A better approach is to assign to a `t` here, and assign `exp.type = t;` when the operands are assigned.

src/ddmd/expressionsem.d
```diff
@@ -7924,59 +7924,61 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+        auto f1 = checkNonAssignmentArrayOp(e1x);
+        auto f2 = checkNonAssignmentArrayOp(e2x);
         if (f1 || f2)
         {
             result = new ErrorExp();
             return;
         }
 
-        if (exp.e2.type.ty == Tvoid)
+        if (e2x.type.ty == Tvoid)
```
WalterBright: By assigning to `exp.type` here, the type gets disconnected from the operands, putting `exp` into an invalid state. A better approach is to assign to a `t` here, and assign `exp.type = t;` when the operands are assigned.

src/ddmd/expressionsem.d
```diff
@@ -7924,59 +7924,61 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
             exp.type = Type.tvoid;
         else
         {
-            exp.e2 = exp.e2.toBoolean(sc);
+            e2x = e2x.toBoolean(sc);
```
WalterBright: here too

src/ddmd/expressionsem.d
```diff
@@ -7924,59 +7924,61 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
-        if (exp.e2.op == TOKerror)
+        if (e2x.op == TOKerror)
         {
-            result = exp.e2;
+            result = e2x;
             return;
         }
 
+        exp.e1 = e1x;
+        exp.e2 = e2x;
```
WalterBright: Put `exp.type = t;` here.

src/ddmd/expressionsem.d
```diff
@@ -7991,59 +7993,61 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+        auto f1 = checkNonAssignmentArrayOp(e1x);
+        auto f2 = checkNonAssignmentArrayOp(e2x);
         if (f1 || f2)
         {
             result = new ErrorExp();
             return;
         }
 
-        if (exp.e2.type.ty == Tvoid)
+        if (e2x.type.ty == Tvoid)
```
WalterBright: same problem as above.

src/ddmd/expressionsem.d
ibuclaw: Hmm, you mean that if the first time round the caller ignores the propagated error and runs semantic on this expression again?

ibuclaw: That is to say, setting type means that you are saying this expression has completed its semantic, when infact it could instead return error (in which case, semantic did not complete).

WalterBright: That's right. Even if it doesn't, it looks suspicious, which will cause extra work for anyone trying to understand the code.

ibuclaw:
Actually in this case, as the logic in-between setting `exp.type`, and returning the result is just error propagation, I can just move the setting of `exp.type` to the end, so that a temporary is not needed.

WalterBright:
When fixing a bugzilla issue, please put the PR URL in the bugzilla issue, so somebody else doesn't waste time trying to fix it because he's ignorant there is an extant fix already.

ibuclaw:
I still think that the cross reference should be handled by the dlang bot/whatever github hook posts messages to bugzilla on commits.

ibuclaw:
Tests pass.

WalterBright:
> I still think that the cross reference should be handled by the dlang bot/whatever github hook posts messages to bugzilla on commits.

I agree, but in the meantime, please add the links!

test/fail_compilation/fail17722b.d
```diff
@@ -0,0 +1,13 @@
+/* TEST_OUTPUT:
```
WalterBright: Please merge these two test files into one. It'll consume half the resources of the autotester that way.

WalterBright:
You'd have had half the work to do if you'd pulled https://github.com/dlang/dmd/pull/7069

ibuclaw:
> You'd have had half the work to do if you'd pulled #7069 :-)

I'd still end up doing double work, because then I'd have to rebase this PR.

test/fail_compilation/fail17722b.d
ibuclaw: It's kept as two separate because the compiler doesn't issue an error for both when together in the same file.

PR #7069, opened 2017-08-07
merge AndAndExp and OrOrExp into LogLogExp

WalterBright:
The two are very nearly identical, so merge them and eliminate code.

### DIFF
--- a/src/ddmd/astbase.d
+++ b/src/ddmd/astbase.d
@@ -5562,24 +5562,11 @@ struct ASTBase
         }
     }
 
-    extern (C++) final class AndAndExp : BinExp
+    extern (C++) final class LogicalExp : BinExp
     {
-        extern (D) this(Loc loc, Expression e1, Expression e2)
+        extern (D) this(Loc loc, TOK op, Expression e1, Expression e2)
         {
-            super(loc, TOKandand, __traits(classInstanceSize, AndAndExp), e1, e2);
-        }
-
-        override void accept(Visitor v)
-        {
-            v.visit(this);
-        }
-    }
-
-    extern (C++) final class OrOrExp : BinExp
-    {
-        extern (D) this(Loc loc, Expression e1, Expression e2)
-        {
-            super(loc, TOKoror, __traits(classInstanceSize, OrOrExp), e1, e2);
+            super(loc, op, __traits(classInstanceSize, LogicalExp), e1, e2);
         }
 
         override void accept(Visitor v)
--- a/src/ddmd/e2ir.d
+++ b/src/ddmd/e2ir.d
@@ -3373,36 +3373,18 @@ elem *toElem(Expression e, IRState *irs)
         /***************************************
          */
 
-        override void visit(AndAndExp aae)
+        override void visit(LogicalExp aae)
         {
             tym_t tym = totym(aae.type);
 
             elem *el = toElem(aae.e1, irs);
             elem *er = toElemDtor(aae.e2, irs);
-            elem *e = el_bin(OPandand,tym,el,er);
+            elem *e = el_bin(aae.op == TOKandand ? OPandand : OPoror,tym,el,er);
 
             elem_setLoc(e, aae.loc);
 
             if (global.params.cov && aae.e2.loc.linnum)
                 e.EV.E2 = el_combine(incUsageElem(irs, aae.e2.loc), e.EV.E2);
-            result = e;
-        }
-
-        /***************************************
-         */
-
-        override void visit(OrOrExp ooe)
-        {
-            tym_t tym = totym(ooe.type);
-
-            elem *el = toElem(ooe.e1, irs);
-            elem *er = toElemDtor(ooe.e2, irs);
-            elem *e = el_bin(OPoror,tym,el,er);
-
-            elem_setLoc(e, ooe.loc);
-
-            if (global.params.cov && ooe.e2.loc.linnum)
-                e.EV.E2 = el_combine(incUsageElem(irs, ooe.e2.loc), e.EV.E2);
             result = e;
         }
 
--- a/src/ddmd/sideeffect.d
+++ b/src/ddmd/sideeffect.d
@@ -287,14 +287,10 @@ extern (C++) bool discardValue(Expression e)
         e.error("`%s` has no effect", e.toChars());
         return true;
     case TOKandand:
-        {
-            AndAndExp aae = cast(AndAndExp)e;
-            return discardValue(aae.e2);
-        }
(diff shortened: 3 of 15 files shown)


ibuclaw:
What is a LogLog expression?

WalterBright:
Logical Logical operators

ibuclaw:
Maybe it should just be called either one of LogExp (logical expression), LogicalExp, or BinLogExp (binary logical expression).

Reading out aloud the name logical-logical operator gives me the same feeling of incorrectness that that-that or had-had would. You have to read it twice to make sure you understand it.

src/ddmd/astbase.d
```diff
@@ -5560,24 +5560,11 @@ struct ASTBase
         }
     }
 
-    extern (C++) final class AndAndExp : BinExp
+    extern (C++) final class LogLogExp : BinExp
```
JinShil: Could use a comment explaining that `LogLogExp` (or whatever name is ultimately decided on) is for `&&` _and_ `||` expressions so contributors aren't left guessing.

JinShil:
As a new spelunker of DMD source code, I have to say I'm not crazy about this change; it hides something that was previously quite obvious behind something more obscure.  

If the `AndAndExp` and `OrOrExp` share common functionality, then to whatever extent it is possible, create an `abstract LogExp` class and concrete `AndAndExp` and `OrOrExp` classes, and try to refactor the common code to the base `LogExp`.  `visit(LogExp)` can contain the common functionality with `visit(AndAndExp)` and `visit(OrOrExp)` forwarding to it.

JinShil:
> As a new spelunker of DMD source code, I have to say I'm not crazy about this change

...but the elimination of redundant code is nice.

WalterBright:
I've used `loglog` for the andand/oror expressions for a very long time:
https://github.com/dlang/dmd/blob/master/src/ddmd/backend/cod2.c?utf8=%E2%9C%93#L2223

ibuclaw:
Well this being shared code across potentially many apps, there is a need for it to be understood by anyone reading the api docs (if/when they get published).

src/ddmd/astbase.d
WalterBright: That's https://github.com/dlang/dmd/pull/7061

UplinkCoder:
@WalterBright I agree with @JinShil here.
LogLogExp is a headscratcher.
couldln't you introduce it as a superclass of AndAndExp and OrOrExp ?

ibuclaw:
> couldln't you introduce it as a superclass of AndAndExp and OrOrExp ?

Hmm, right, I've just noticed that this removes AndAnd and OrOrExp declarations, this breaks downstream projects that use these.

At the very least keep the original classes around, update c++ headers, then this is OK for inclusion.

WalterBright:
> couldln't you introduce it as a superclass of AndAndExp and OrOrExp ?

That wouldn't eliminated near as much code. What I did here is hardly different from IdentityExp and CmpExp, to name just two.

> this breaks downstream projects that use these.

Indeed it does. The fix is generally to delete code (most of this PR is just deleting code).

> At the very least keep the original classes around

Why?

> update c++ headers

expression.h is already updated in this PR

jacob-carlborg:
> Maybe it should just be called either one of LogExp (logical expression), LogicalExp, or BinLogExp (binary logical expression).

Yes, please. Ideally it should be called `LogicalExpression`, but to be consistent with the rest of the code, `LogicalExp` would be acceptable as well.

Keep in mind that we want to expose DMD as a library, then we need to think about the names we pick and avoid abbreviations.

ibuclaw:
> Why?

There's no cost keeping them around, and downstream can choose to update independently in their own time, if they even choose to.

WalterBright:
renamed LogLogExp to LogicalExp

wilzbach (approved):
> +57 / -300

This is amazing!

wilzbach:
> There's no cost keeping them around

There is always a hidden cost - especially when it's about maintaining deprecations.

> and downstream can choose to update independently in their own time, if they even choose to.

AFAICT DMD has never issued deprecations before and as DaaL (DMD as a library) isn't really used yet or rarely used, I don't see why this should start with this PR.
Besides the change itself here is really trivial from two classes into one.

JinShil:
This change is fine.  The reduction in code is nice, and as I spend more time with the DMD code, I realize it is idiomatic DMD.  However, it may be worth remembering that DMD was translated from C++ (whose idioms were more C-ish than C++-ish), and this change seems to favor the C-ish idioms over the modeling power of D.

That being said, the `LogicalExp` name is better and the reduction in code is probably worth it, but those benefits are being traded for a less intuitive API, IMO.

src/ddmd/sideeffect.d
```diff
@@ -287,14 +287,10 @@ extern (C++) bool discardValue(Expression e)
...
     case TOKandand:
-        {
-            AndAndExp aae = cast(AndAndExp)e;
-            return discardValue(aae.e2);
-        }
     case TOKoror:
         {
-            OrOrExp ooe = cast(OrOrExp)e;
-            return discardValue(ooe.e2);
+            LogicalExp aae = cast(LogicalExp)e;
```
JinShil: Nit: `aae` probably isn't such a good name anymore.

src/ddmd/e2ir.d
```diff
@@ -3373,13 +3373,13 @@ elem *toElem(Expression e, IRState *irs)
         /***************************************
          */
 
-        override void visit(AndAndExp aae)
+        override void visit(LogicalExp aae)
```
JinShil: Nit: `aae` probably isn't such a good name anymore.

src/ddmd/staticcond.d
```diff
@@ -40,25 +40,23 @@ import ddmd.utils;
 
 bool evalStaticCondition(Scope* sc, Expression exp, Expression e, ref bool errors)
 {
-    if (e.op == TOKandand)
+    if (e.op == TOKandand || e.op == TOKoror)
     {
-        AndAndExp aae = cast(AndAndExp)e;
+        LogicalExp aae = cast(LogicalExp)e;
```
JinShil: Nit: `aae` probably isn't such a good name anymore.

JinShil (changes requested):

jacob-carlborg:
> renamed LogLogExp to LogicalExp

Thanks.

andralex:
@WalterBright thanks very much for compromising on the naming, it certainly threw me off as much as some of the folks who chimed in in here.

andralex (approved):

PR #7076, opened 2017-08-10
fix Issue 17388 - [scope] e[] should be treated like &e as far as sco…

WalterBright:
…pe goes

`&e` was working correctly, so the fix was to make slicing a static array follow the same logic.

Had to make checkAddressVar() a global function so it can be called by both AddrExp and SliceExp.

### DIFF
--- a/src/ddmd/expressionsem.d
+++ b/src/ddmd/expressionsem.d
@@ -3267,6 +3267,9 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         }
         if (exp.e1.op == TOKslice || exp.e1.type.ty == Tarray || exp.e1.type.ty == Tsarray)
         {
+            if (exp.e1.op == TOKslice)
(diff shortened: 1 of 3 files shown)


src/ddmd/expressionsem.d
```diff
@@ -3267,6 +3267,9 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         }
         if (exp.e1.op == TOKslice || exp.e1.type.ty == Tarray || exp.e1.type.ty == Tsarray)
         {
+            if (exp.e1.op == TOKslice)
+                (cast(SliceExp)exp.e1).arrayop = true;
```
MartinNowak: I hope we can remove that special case once the library arrayOps story is merged. Still need to make it scope safe though.

MartinNowak:
Looks like you forgot PowAssignExp @WalterBright.

PR #7077, opened 2017-08-10
Fix test11559

UplinkCoder:
Replace ~= with slice-assignment into a pre-allocated array.

This way newCTFE will not have a heap-overflow when executing the test.

### DIFF
--- a/test/compilable/test11559upgradeoptlink.d
+++ b/test/compilable/test11559upgradeoptlink.d
@@ -2,12 +2,28 @@
 
 // If this is failing, you need optlink 8.00.14 or higher
 
+string repeatString(string s, uint repeatCount)
+{
+    char[] result;
+    uint sLength = cast(uint) s.length;
+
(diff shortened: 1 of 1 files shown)


UplinkCoder:
@andralex, @MartinNowak either one; please merge.

MartinNowak:
Well you gotta deal with such allocations patterns at some point. How complex is it to support?
I hate rewritting tests just to make immediate issues go away, they will obviously pop up again in someone else's codebase.

MartinNowak:
It's OK to fix this temporarily on the newCTFE branch, but eventually we want to merge the whole story without touching existing tests (except for really broken ones).

UplinkCoder:
@MartinNowak the only way I can deal with this is by allowing more memory to be allocated.

UplinkCoder:
@klickverbot newCTFE and the old interpreter have the same memory-allocation for ~=
it's essentailly O(n^2)

PR #6024, opened 2016-08-06
purge deferredSemantic arrays after processing them

MartinNowak:
- fixes an O(N^2) loop introduced in 201b124e6c that consumes
  20% of the runtime for our dmd, druntime, and phobos test suite

PR #7078, opened 2017-08-10
Optimize ecom

UplinkCoder:
this cuts down a little on the loads and therefore data-cache misses.
Let's remove more barnacles in the future. :)
@WalterBright

### DIFF
--- a/src/ddmd/backend/cgcs.c
+++ b/src/ddmd/backend/cgcs.c
@@ -415,7 +415,8 @@ STATIC void ecom(elem **pe)
             dbg_printf("i: %2d Hhash: %6d Helem: %p\n",
                 i,hcstab[i].Hhash,hcstab[i].Helem);
 #endif
-        if (hash == hcstab[i].Hhash && (ehash = hcstab[i].Helem) != NULL)
+        HCS entry = hcstab[i];
+        if (hash == entry.Hhash && (ehash = entry.Helem) != NULL)
         {
             /* if elems are the same and we still have room for more    */
             if (el_match(e,ehash) && ehash->Ecount < 0xFF)


src/ddmd/backend/cgcs.c
```diff
@@ -415,7 +415,8 @@ STATIC void ecom(elem **pe)
             dbg_printf("i: %2d Hhash: %6d Helem: %p\n",
                 i,hcstab[i].Hhash,hcstab[i].Helem);
 #endif
-        if (hash == hcstab[i].Hhash && (ehash = hcstab[i].Helem) != NULL)
+        typeof(hcstab[0]) entry = hcstab[i];
```
WalterBright: use `auto` instead of `typeof`

src/ddmd/backend/cgcs.c
```diff
@@ -415,7 +415,8 @@ STATIC void ecom(elem **pe)
...
+        if (hash == entry.Hhash && (ehash = entry.Helem) != NULL)
```
WalterBright: I'd check the generated code and see if it is an actual improvement. I doubt it, as compilers do these sorts of optimizations automatically.

andralex:
According to an approximate repro at https://godbolt.org/g/AofXzH there is no double evaluation to be eliminated.

The optimization at work is well established: https://sunilkumarn.wordpress.com/2010/10/19/common-subexpression-elimination-cse/

@UplinkCoder can you measure an improvement that this diff brings?

UplinkCoder:
@andralex 
https://godbolt.org/g/oMD9Ah

If you use size optimizations you will see that using the variable makes the flow clearer and allows the compiler to do the move before the branch. has positive effects on systems that support prefetching and branch-predication.

it's hard to measure these little things. :)

also this will still be better when you are compiling in debug more which I often do :)

src/ddmd/backend/cgcs.c
UplinkCoder: I think auto is not type-inferred in the c++ dialect used by this code
`error: 'entry' does not name a type`

WalterBright: Indeed, you are correct.

WalterBright:
> CSE expensive optimization to do

It does it anyway, even for debug builds, because it is a huge win.

> If you use size optimizations you will see that using the variable makes the flow clearer

The flow in the assembler is not intended to be clear, zero consideration is given for that.

> and allows the compiler to do the move before the branch. has positive effects on systems that support prefetching and branch-predication.

This is speculative. x86 processors change constantly, and one instruction sequence may be faster than other on one processor, then the other way around on the next one, then back again. Attempting such micro-optimizations in the source code is fruitless. Such things should be addressed by the code generator.

> it's hard to measure these little things. :)

It's a waste of time micro optimizing unless you've got profile measurements to back it up. I know this from a great deal of experience.

Profiling dmd is the first step in figuring out how to make it faster. Otherwise, you're practically guaranteed to optimize things that have no effect on execution speed, or even make things worse.

ibuclaw:
I agree, this seems like a bit of beating a dead horse to me.  (I'm of the opinion that optimisation starts with our use of data structures, #6024 is a good example).

UplinkCoder:
if nothing else atleast the expression ` hcstab[i]` in not repeated.
so we could merge just for a little DRY.
But feel free to close if that is not reason enough.

PR #5636, opened 2016-04-07
Refactor: Add helper functions to create temporaries

9rnsr:

PR #7079, opened 2017-08-11
Fix Issue 8006: Property binary assignment operators

JinShil:
FYI: This does not address unary assignment operators (e.g. `s.x++`, `++s.x`).  I intend to implement that in a future pull request.

DIP PR: https://github.com/dlang/DIPs/pull/97
Spec change: https://github.com/dlang/dlang.org/pull/1865

### DIFF
--- a/test/fail_compilation/test8006.d
+++ b/test/fail_compilation/test8006.d
@@ -1,0 +1,192 @@
+/*
+ * TEST_OUTPUT:
+---
+fail_compilation/test8006.d(79): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(80): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(81): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(82): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(83): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(84): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(85): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(86): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(87): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(88): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(89): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(90): Error: function `test8006.TInt.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(94): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(95): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(96): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(97): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(98): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(99): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(100): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(101): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(102): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(103): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(104): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(105): Error: `ti.y()` is not an lvalue
+fail_compilation/test8006.d(124): Error: function `test8006.TString.x()` is not callable using argument types `(string)`
+fail_compilation/test8006.d(128): Error: `ts.y()` is not an lvalue
+fail_compilation/test8006.d(154): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(155): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(156): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(157): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(158): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(159): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(160): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(161): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(162): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(163): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(164): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(165): Error: function `test8006.x()` is not callable using argument types `(int)`
+fail_compilation/test8006.d(167): Error: function `test8006.xs()` is not callable using argument types `(string)`
+fail_compilation/test8006.d(171): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(172): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(173): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(174): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(175): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(176): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(177): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(178): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(179): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(180): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(181): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(182): Error: `y()` is not an lvalue
+fail_compilation/test8006.d(184): Error: `ys()` is not an lvalue
+---
+ */
+
+// https://issues.dlang.org/show_bug.cgi?id=8006
+
+// modeled after code from runnable/testassign.d
+
+struct TInt
+{
+    int mX;
+
+    @property int x() { return mX; }
+
+    int mY;
+    int y() { return mY; }
+    int y(int v) { return mY = v; }
+}
+
+void testTInt()
+{
+    // all of these should fail to compile because there is
+    // no setter property
+    TInt ti;
+    ti.x += 4;
+    ti.x -= 2;
+    ti.x *= 4;
+    ti.x /= 2;
+    ti.x %= 3;
+    ti.x <<= 3;
+    ti.x >>= 1;
+    ti.x >>>= 1;
+    ti.x &= 0xF;
+    ti.x |= 0x8;
+    ti.x ^= 0xF;
+    ti.x ^^= 2;
+
+    // all of these should fail to compile because y is not a
+    // @property function
+    ti.y += 4;
+    ti.y -= 2;
+    ti.y *= 4;
+    ti.y /= 2;
+    ti.y %= 3;
+    ti.y <<= 3;
+    ti.y >>= 1;
+    ti.y >>>= 1;
+    ti.y &= 0xF;
+    ti.y |= 0x8;
+    ti.y ^= 0xF;
+    ti.y ^^= 2;
+}
+
+struct TString
+{
+    string mX;
+
+    @property string x() { return mX; }
+
+    string mY;
+    string y() { return mY; }
+    string y(string v) { return mY = v; }
+}
+
+void testTString()
+{
+    // this should fail to compile because there is
+    // no setter property
+    TString ts;
+    ts.x ~= "def";
+
+    // this should fail to compile because y is not a
+    // @property function
+    ts.y ~= "def";
+}
+
+
+// int @property function without a setter
+int mX;
+@property int x() { return mX; }
+
+// int non-@property functions
+int mY;
+int y() { return mY; }
+int y(int v) { return mY = v; }
+
+// string @property function without a setter
+string mXs;
+@property string xs() { return mXs; }
+
+// string non-@property functions
+string mYs;
+string ys() { return mYs; }
+string ys(string v) { return mYs = v; }
+
+void testFreeFunctions()
+{
+    // all of these should fail to compile because there is
+    // no setter property
+    x += 4;
+    x -= 2;
+    x *= 4;
+    x /= 2;
+    x %= 3;
+    x <<= 3;
+    x >>= 1;
+    x >>>= 1;
+    x &= 0xF;
+    x |= 0x8;
+    x ^= 0xF;
+    x ^^= 2;
+
+    xs ~= "def";
+
+    // all of these should fail to compile because y and ys are not
+    // @property function
+    y += 4;
+    y -= 2;
+    y *= 4;
+    y /= 2;
+    y %= 3;
+    y <<= 3;
+    y >>= 1;
+    y >>>= 1;
+    y &= 0xF;
+    y |= 0x8;
+    y ^= 0xF;
+    y ^^= 2;
+
+    ys ~= "def";
+}
+
+void main()
+{
+    testTInt();
+    testTString();
+    testFreeFunctions();
+}
--- a/test/runnable/test8006.d
+++ b/test/runnable/test8006.d
@@ -1,0 +1,679 @@
+// https://issues.dlang.org/show_bug.cgi?id=8006
+
+/**************************************************************
+* int tests
+**************************************************************/
+struct TInt
+{
+    int mX;
+
+    @property int x() { return mX; }
+    @property void x(int v) { mX = v; }
+
+    alias x this;
+}
+
+// It was found, during the implementation of binary assignment operators
+// for @property functions that if the setter was declared before the getter
+// the binary assignment operator call would not compile.  This was due
+// to the fact that if e.e1.copy() was called after resolveProperties(e.e1)
+// that the copy() call would return the wrong overload for the @property
+// function.  This is a test to guard against that.
+struct TIntRev
+{
+    int mX;
+
+    @property void x(int v) { mX = v; }
+    @property int x() { return mX; }
+
+    alias x this;
+}
+
+// Ensure that `ref @property` functions also *still* work
+struct TIntRef
+{
+    int mX;
+
+    @property ref int x() { return mX; }
+    @property ref int x(int v) { return mX = v; }
+
+    alias x this;
+}
+
+// same as above with no setter
+struct TIntRefNoSetter
+{
+    int mX;
+
+    @property ref int x() { return mX; }
+
+    alias x this;
+}
+
+// Same as TInt, but setter @property function returns a value
+struct TIntRet
+{
+    int mX;
+
+    @property int x() { return mX; }
+    @property int x(int v) { return mX = v; }
+
+    alias x this;
+}
+
+// same as TInt, but with static @property functions
+struct TIntStatic
+{
+    static int mX;
+
+    static @property int x() { return mX; }
+    static @property void x(int v) { mX = v; }
+
+    alias x this;
+}
+
+// same as TIntStatic, but setter @property function returns a value
+struct TIntRetStatic
+{
+    static int mX;
+
+    static @property int x() { return mX; }
+    static @property int x(int v) { return mX = v; }
+
+    alias x this;
+}
+
+// This test verifies typical arithmetic and logical operators
+void testTInt(T)()
+{
+    // modeled after code from runnable/testassign.d
+
+    static if (typeid(T) is typeid(TInt))
+    {
+        TInt t;
+    }
+    else static if (typeid(T) is typeid(TIntRev))
+    {
+        TIntRev t;
+    }
+    else static if (typeid(T) is typeid(TIntRef))
+    {
+        TIntRef t;
+    }
+    else static if (typeid(T) is typeid(TIntRefNoSetter))
+    {
+        TIntRefNoSetter t;
+    }
+    else static if (typeid(T) is typeid(TIntStatic))
+    {
+        alias t = TIntStatic;
+    }
+    else
+    {
+        static assert(false, "Type is not supported");
+    }
+
+    t.x += 4;
+    assert(t.mX == 4);
+    t.x -= 2;
+    assert(t.mX == 2);
+    t.x *= 4;
+    assert(t.mX == 8);
+    t.x /= 2;
+    assert(t.mX == 4);
+    t.x %= 3;
+    assert(t.mX == 1);
+    t.x <<= 3;
+    assert(t.mX == 8);
+    t.x >>= 1;
+    assert(t.mX == 4);
+    t.x >>>= 1;
+    assert(t.mX == 2);
+    t.x &= 0xF;
+    assert(t.mX == 0x2);
+    t.x |= 0x8;
+    assert(t.mX == 0xA);
+    t.x ^= 0xF;
+    assert(t.mX == 0x5);
+    t.x ^^= 2;
+    assert(t.mX == 25);
+
+    // same as test above, but through the `alias this`
+    t = 0;
+    t += 4;
+    assert(t.mX == 4);
+    t -= 2;
+    assert(t.mX == 2);
+    t *= 4;
+    assert(t.mX == 8);
+    t /= 2;
+    assert(t.mX == 4);
+    t %= 3;
+    assert(t.mX == 1);
+    t <<= 3;
+    assert(t.mX == 8);
+    t >>= 1;
+    assert(t.mX == 4);
+    t >>>= 1;
+    assert(t.mX == 2);
+    t &= 0xF;
+    assert(t.mX == 0x2);
+    t |= 0x8;
+    assert(t.mX == 0xA);
+    t ^= 0xF;
+    assert(t.mX == 0x5);
+    t ^^= 2;
+    assert(t.mX == 25);
+}
+
+// This test is to verify that the setter @property function
+// returns a value if it is explicitly coded to do so
+void testTIntRet(T)()
+{
+    static if (typeid(T) is typeid(TIntRet))
+    {
+        TIntRet t;
+    }
+    else static if (typeid(T) is typeid(TIntRetStatic))
+    {
+        alias t = TIntRetStatic;
+    }
+    else
+    {
+        static assert(false, "Type is not supported");
+    }
+
+    int r;
+    r = t.x += 4;
+    assert(r == 4);
+    r = t.x -= 2;
+    assert(r == 2);
+    r = t.x *= 4;
+    assert(r == 8);
+    r = t.x /= 2;
+    assert(r == 4);
+    r = t.x %= 3;
+    assert(r == 1);
+    r = t.x <<= 3;
+    assert(r == 8);
+    r = t.x >>= 1;
+    assert(r == 4);
+    r = t.x >>>= 1;
+    assert(r == 2);
+    r = t.x &= 0xF;
+    assert(r == 0x2);
+    r = t.x |= 0x8;
+    assert(r == 0xA);
+    r = t.x ^= 0xF;
+    assert(r == 0x5);
+    r = t.x ^^= 2;
+    assert(r == 25);
+
+    // same as test above, but through the `alias this`
+    t = 0;
+    r = t += 4;
+    assert(r == 4);
+    r = t -= 2;
+    assert(r == 2);
+    r = t *= 4;
+    assert(r == 8);
+    r = t /= 2;
+    assert(r == 4);
+    r = t %= 3;
+    assert(r == 1);
+    r = t <<= 3;
+    assert(r == 8);
+    r = t >>= 1;
+    assert(r == 4);
+    r = t >>>= 1;
+    assert(r == 2);
+    r = t &= 0xF;
+    assert(r == 0x2);
+    r = t |= 0x8;
+    assert(r == 0xA);
+    r = t ^= 0xF;
+    assert(r == 0x5);
+    r = t ^^= 2;
+    assert(r == 25);
+}
+
+/**************************************************************
+* string/array tests
+**************************************************************/
+struct TString
+{
+    string mX;
+
+    @property string x() { return mX; }
+    @property void x(string v) { mX = v; }
+
+    alias x this;
+}
+
+// same as TString, but setter @property function returns a value
+struct TStringRet
+{
+    string mX;
+
+    @property string x() { return mX; }
+    @property string x(string v) { return mX = v; }
+
+    alias x this;
+}
+
+struct TStringOp
+{
+    string mX;
+
+    @property string x() { return mX; }
+    @property void x(string v) { mX = v; }
+
+    string mB;
+    string opOpAssign(string op)(string rhs)
+    {
+        return mixin("mB "~op~"= rhs");
+    }
+
+    alias x this;
+}
+
+// same as TString, but for static @property functions
+struct TStringStatic
+{
+    static string mX;
+
+    static @property string x() { return mX; }
+    static @property void x(string v) { mX = v; }
+
+    static alias x this;
+}
+
+// same as TStringRet, but for static @property functions
+struct TStringRetStatic
+{
+    static string mX;
+
+    static @property string x() { return mX; }
+    static @property string x(string v) { return mX = v; }
+
+    static alias x this;
+}
+
+// Test string (i.e. array) operators
+void testTString(T)()
+{
+    static if (typeid(T) is typeid(TString))
+    {
+        TString t;
+    }
+    else static if (typeid(T) is typeid(TStringStatic))
+    {
+        alias t = TStringStatic;
+    }
+    else
+    {
+        static assert(false, "Type is not supported");
+    }
+
+    t.x = "abc";
+    t.x ~= "def";
+    assert(t.mX == "abcdef");
+
+    // same as test above, but through the `alias this`
+    t = "abc";
+    t ~= "def";
+    assert(t.mX == "abcdef");
+}
+
+// This test is to verify that the setter @property function
+// returns a value if it is explicitly coded to do so
+void testTStringRet(T)()
+{
+    static if (typeid(T) is typeid(TStringRet))
+    {
+        TStringRet t;
+    }
+    else static if (typeid(T) is typeid(TStringRetStatic))
+    {
+        alias t = TStringRetStatic;
+    }
+    else
+    {
+        static assert(false, "Type is not supported");
+    }
+
+    string s;
+    t.x = "abc";
+    s = t.x ~= "def";
+    assert(s == "abcdef");
+
+    // same as test above, but through the `alias this`
+    t = "abc";
+    s = t ~= "def";
+    assert(s == "abcdef");
+}
+
+/**************************************************************
+* Free @property function test
+**************************************************************/
+int mX;
+@property int x() { return mX; }
+@property void x(int v) { mX = v; }
+
+// Test that free @property functions work
+void testFreeFunctionsInt()
+{
+    x += 4;
+    assert(mX == 4);
+    x -= 2;
+    assert(mX == 2);
+    x *= 4;
+    assert(mX == 8);
+    x /= 2;
+    assert(mX == 4);
+    x %= 3;
+    assert(mX == 1);
+    x <<= 3;
+    assert(mX == 8);
+    x >>= 1;
+    assert(mX == 4);
+    x >>>= 1;
+    assert(mX == 2);
+    x &= 0xF;
+    assert(mX == 0x2);
+    x |= 0x8;
+    assert(mX == 0xA);
+    x ^= 0xF;
+    assert(mX == 0x5);
+    x ^^= 2;
+    assert(mX == 25);
+}
+
+int mXret;
+@property int xret() { return mXret; }
+@property int xret(int v) { return mXret = v; }
+
+// Same as testFreeFunctions except that we want to
+// ensure the binary assignment returns a value
+void testFreeFunctionsIntRet()
+{
+    int r;
+    r = xret += 4;
+    assert(r == 4);
+    r = xret -= 2;
+    assert(r == 2);
+    r = xret *= 4;
+    assert(r == 8);
+    r = xret /= 2;
+    assert(r == 4);
+    r = xret %= 3;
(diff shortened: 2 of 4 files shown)


JinShil:
Appears to also fix [Issue 16187](https://issues.dlang.org/show_bug.cgi?id=16187)

src/ddmd/expressionsem.d
```diff
@@ -3238,12 +3238,77 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+                    e2 = new ShrExp(exp.loc, exp.e1, exp.e2);
+                    break;
+                case TOKushrass:
+                    e2 = new UshrExp(exp.loc, exp.e1, exp.e2);
+                    break;
+                default:
+                    assert(false, "Binary assignment operator was not handled.");
+            }
+
+            if (Expression ex = resolvePropertiesX(sc, e1, e2))
```
ibuclaw: Hmm, the backend should already handle lowering `a += b` into `a = a + b`.

Maybe all you need is the call to `resolvePropertiesX()` ?

JinShil: `resolvePropertiesX` lowers `e1 = e2` into `e1(e2)` if `e1` is a property, so I need `e2` to be `e1 + e2` for that to work.  

Currently DMD errors with `e1() is not an lvalue`, so, if `e1 += e2` is lowered to `e1 = e1 + e2` in the backend, maybe I just need to figure out how to get past that error so the expression makes it to the backend.  But are you sure the backed is doing the lowering?

ibuclaw: Actually, yes, I see the reasoning behind this.  `e1 = e1 + e2` won't cut the muster, as it will assign the result to a temporary.  You need it to be `e1(e1 + e2)`

ibuclaw: Giving this a quick test with gdc, it looks like `resolvePropertiesX` is changing the return type of `ti.x()` from int to void.
You may need to store the rhs in a temporary here.

JinShil: Thanks @ibuclaw.  I've copied the rhs to a temporary and refactored in an attempt to consolidate code.

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8808,100 @@ Expression binSemanticProp(BinExp e, Scope* sc)
...
+ */
+Expression binAssignRewriteProp(BinAssignExp e, Scope* sc)
+{
+    // https://issues.dlang.org/show_bug.cgi?id=8006
+
+    if (e.e1.op == TOKdotid)
+    {
+        // Consider s.x += 1;
+        // We want to rewrite that expression to s.x(s.x() + 1);
+        // However, resolvePropertiesX will change the type of s.x() to void,
```
ibuclaw: Thanks.  Looking at this after sleeping on it, I sense that what I am see is infact memory that is being shared between both `s.x()` and `s.x(1)` calls.  When the latter gets rewritten, it changes the type of the former also (oops!).

I'm just having a look now.  I think this is close to being where it should be anyway, just a few nits.

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8808,100 @@ Expression binSemanticProp(BinExp e, Scope* sc)
...
+ * of the form, for example, `s.x += 1` to `s.x(s.x() += 1)` if s.x is
+ * a getter and setter `@property`.
+ * Params:
+ *      e = the expression to rewrite
+ *      sc = the semantic scope
+ * Returns:
+ *      the rewritten expression if the procedure succeeds, an `ErrorExp` if the
+ *      and error is encountered, or `null` if `e.e1` is not a `DotIdExp`.
+ */
+Expression binAssignRewriteProp(BinAssignExp e, Scope* sc)
```
ibuclaw: This could be renamed to `binSemanticProp`, then you don't have to call this explicitly.

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8808,100 @@ Expression binSemanticProp(BinExp e, Scope* sc)
...
+{
+    // https://issues.dlang.org/show_bug.cgi?id=8006
+
+    if (e.e1.op == TOKdotid)
```
ibuclaw: Maybe its better to complete semantic on `e1` first, then check that `e1.op == TOKcall` and `ce.e1.type.isproperty`.

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8808,100 @@ Expression binSemanticProp(BinExp e, Scope* sc)
...
+        if (e1.op == TOKerror)
+        {
+            return e1;
+        }
+
+        // int __tmp = s.x();
+        auto v = copyToTemp(0, "__tmp", e.e1);
+        auto de = new DeclarationExp(e.e1.loc, v);
+        auto ve = new VarExp(e.e1.loc, v);
+        Expression ea = Expression.combine(de, ve);
```
ibuclaw: If you pass a copy of e1 to `resolvePropertiesX`, then this shouldn't be needed.  Codegen looks better at least without this.

JinShil: What do you mean by "this"?  Do you mean I don't need to copy to a temporary?

src/ddmd/expressionsem.d
JinShil: I'm not sure what you mean by `ce`.  `isproperty` is only a field in `TypeFunction`, and I'm not sure how to get that.

src/ddmd/expressionsem.d
JinShil: Nice call!  That made the code just that much more concise.

src/ddmd/expressionsem.d
ibuclaw: After playing around, this is what I ended up with.
https://pastebin.com/iUXCpeju

Whether it is more correct than what you are doing is, I may have to leave someone else to be the judge of.

JinShil:
It appears a recent merge may have broken this :(

ibuclaw:
> It appears a recent merge may have broken this :(
> Update: Nope I broke something somewhere.

I was about to say, I couldn't see anything wrong with the diff, and running it locally passes for me.

Then again, maybe you updated it just before I looked. :-)

ibuclaw:
This might have broken `dub` (see the jenkins ci test).  However I suspect that it may be a legitimate failure.

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8784,94 @@ Expression binSemanticProp(BinExp e, Scope* sc)
...
+    //********************************************************************
+    // Check for property assignment.
+    // https://issues.dlang.org/show_bug.cgi?id=8006
+    if (e1x.op == TOKcall)
+    {
+        CallExp ce = cast(CallExp)e1x;
+
+        // Consider s.x() += 1;
+        // We want to rewrite that expression to s.x(s.x() + 1);
+        if (ce.e1.type.ty == Tfunction && (cast(TypeFunction)ce.e1.type).isproperty)
```
ibuclaw: Actually, I think it's best to have this check as:
```
auto tf = (ce.e1.type.ty) == Tfunction ? cast(TypeFunction)ce.e1.type : null;
if (tf && tf.isproperty && !tf.isref)
```
So that `ref @property` functions are *not* rewritten.

JinShil:
I think I need the return type of `s.x(s.x() + 1)` to be the same as `s.x()`.  Consider `writeln(s.x += 1)`.  If that gets rewritten as `writeln(s.x(s.x() + 1))`, it should print a number.  At the moment `s.x(s.x() + 1)` appears to have a return type of `void`.  Go back to copying rhs to a temporary?  It seems to work.

src/ddmd/expressionsem.d
JinShil: Ok, that seems to work.  I'll try to formulate a test case for that as well.  Thanks!

JinShil:
> At the moment `s.x(s.x() + 1)` appears to have a return type of `void`.

Though that is consistent with `@property void x(int v) { mX = v; }`.  Perhaps the compiler should require the user to be explicit with `@property int x(int v) { mX = v; return mX }` if they want `s.x += 1` to return a value.

src/ddmd/expressionsem.d
JinShil: Fixed.

JinShil:
> Perhaps the compiler should require the user to be explicit with @property int x(int v) { mX = v; return mX } if they want s.x += 1 to return a value.

Added tests to ensure that if a setter `@property` returns a value the resulting `s.x(s.x() + 1)` expression also returns a value.  User will have to explicitly code this if they want that behavior.

JinShil:
Added tests for `alias this`ed properties.  I suspect making the function an overload of `binSemanticProp` allowed this logic to insert itself into all existing pipelines, so `alias this`ed properties work out of the box.  😎

JinShil:
Thanks to @ibuclaw, this turned out _much_ better than I could have hoped.  I thought I was going to have to split this into a couple of different pull requests, and handle a lot of special casing, but this turned out highly cohesive, comprehensive, and surprisingly simple. 

I think this is good to go.

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8784,97 @@ Expression binSemanticProp(BinExp e, Scope* sc)
     return null;
 }
 
+/********************************************************************************
+ * Helper function to consolidate common code for `BinAssingExp`s.
```
ibuclaw: Maybe:

> Helper function to resolve `@property` functions in a `BinAssignExp`.

Also, spelling mistake in `BinAssignExp`

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8784,97 @@ Expression binSemanticProp(BinExp e, Scope* sc)
...
+                    break;
+                case TOKushrass:
+                    e2x = new UshrExp(e.loc, e1x, e2x);
+                    break;
+                default:
+                    assert(false, "Binary assignment operator was not handled.");
+            }
+
+            // s.x(e2x)
+            return resolvePropertiesX(sc, e.e1.copy(), e2x);
```
ibuclaw: Perhaps the result should be checked here, and only return if not null.

test/fail_compilation/test8006.d
```diff
@@ -0,0 +1,69 @@
+/*
+ * TEST_OUTPUT:
+---
+fail_compilation/test8006.d(48): Error: function test8006.TInt.x () is not callable using argument types (int)
```
ibuclaw: It might be worth investigating whether we could improve the fail message here.

ibuclaw:
I'd like to have the last say made by @WalterBright, incase there's something I might have missed.

src/ddmd/expressionsem.d
JinShil: Done.

src/ddmd/expressionsem.d
JinShil: Indeed.  Done.

test/fail_compilation/test8006.d
JinShil: Agreed.  I looked into this and found that the error messages really need to be improved for all usages of properties, not just for the binary assignments in this pull request.

The error message situation is broader than the scope of this pull request.  Simple property assignments are currently implemented and produce the same error message as above.  So this pull request maintains the status quo.

The following scenarios all produce different error messages, none of which read specifically to syntax of property usage.

- Attempting to set without a setter
- Attempting to get without a getter
- Attempting to set without a setter through `alias this`
- Attempting to get without a getter through `alias this`
- Attempting to set without a setter through `static alias this`
- Attempting to get without a getter through `static alias this`
- Attempting to set a property to the wrong type
- Attempting to set a property to the wrong type through `alias this`
- Attempting to set a property to the wrong type through `static alias this`
- etc...

So, instead of improving the error messages in this pull request, I suggest accepting this pull request (maintaining the status quo) and improve the error messages for properties more comprehensively in a separate issue and pull request.

JinShil: > ... improve the error messages for properties more comprehensively in a separate issue and pull request.

[Issue 5010](https://issues.dlang.org/show_bug.cgi?id=5010)

WalterBright:
Are we really sure we want to do this? This is a significant language change with essentially no discussion about its merits.

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8784,101 @@ Expression binSemanticProp(BinExp e, Scope* sc)
     return null;
 }
 
+/********************************************************************************
+ * Helper function to resolve `@property` functions in a `BinAssignExp`.
+ * It rewrites expressions of the form, for example, `s.x += 1` to `s.x(s.x() += 1)`
```
WalterBright: I think you mean `s.x(s.x() + 1)`

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8784,101 @@ Expression binSemanticProp(BinExp e, Scope* sc)
...
+    // Check for property assignment.
+    // https://issues.dlang.org/show_bug.cgi?id=8006
+    if (e1x.op == TOKcall)
+    {
+        CallExp ce = cast(CallExp)e1x;
+
+        // Consider s.x() += 1;
+        // We want to rewrite that expression to s.x(s.x() + 1);
+
+        // Only rewrite @property functions.  Don't rewrite @ref @property functions.
```
WalterBright: why not ref?

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8784,101 @@ Expression binSemanticProp(BinExp e, Scope* sc)
...
+                    e2x = new ShlExp(e.loc, e1x, e2x);
+                    break;
+                case TOKshrass:
+                    e2x = new ShrExp(e.loc, e1x, e2x);
+                    break;
+                case TOKushrass:
+                    e2x = new UshrExp(e.loc, e1x, e2x);
+                    break;
+                default:
+                    assert(false, "Binary assignment operator was not handled.");
```
WalterBright: Some people like assert messages, I don't. They bloat up the compiler. Leaving the message as a comment in the source code is plenty good enough.

WalterBright: In fact, you can just make the switch `final` and do away with the assert entirely!

WalterBright:
The implementation itself does look nice.

src/ddmd/expressionsem.d
JinShil: First of all, this already works today:
```
struct TInt
{
    int mX;

    ref @property int x()
    {
        return mX;
    }

    alias x this;
}

void main()
{
    TInt t;

    t.x += 4;
    assert(t.mX == 4);
    t.x -= 2;
    assert(t.mX == 2);
    t.x *= 4;
    assert(t.mX == 8);
    t.x /= 2;
    assert(t.mX == 4);
    t.x %= 3;
    assert(t.mX == 1);
    t.x <<= 3;
    assert(t.mX == 8);
    t.x >>= 1;
    assert(t.mX == 4);
    t.x >>>= 1;
    assert(t.mX == 2);
    t.x &= 0xF;
    assert(t.mX == 0x2);
    t.x |= 0x8;
    assert(t.mX == 0xA);
    t.x ^= 0xF;
    assert(t.mX == 0x5);
    t.x ^^= 2;
    assert(t.mX == 25);

    // same as test above, but through the `alias this`
    t = 0;
    t += 4;
    assert(t.mX == 4);
    t -= 2;
    assert(t.mX == 2);
    t *= 4;
    assert(t.mX == 8);
    t /= 2;
    assert(t.mX == 4);
    t %= 3;
    assert(t.mX == 1);
    t <<= 3;
    assert(t.mX == 8);
    t >>= 1;
    assert(t.mX == 4);
    t >>>= 1;
    assert(t.mX == 2);
    t &= 0xF;
    assert(t.mX == 0x2);
    t |= 0x8;
    assert(t.mX == 0xA);
    t ^= 0xF;
    assert(t.mX == 0x5);
    t ^^= 2;
    assert(t.mX == 25);
}
```
But more importantly, we found that we broke code in DUB when not excluding `ref @property` functions.  The specific idiom that broke was added to the test suite [here](https://github.com/JinShil/dmd/blob/ae7ddcadfb0c25f2a1afcec041c6bb9a7aef2522/test/runnable/test8006.d#L354-L372).

So it `ref @property` functions were excluded in an effort to be conservative with this change, and avoid code breakage.

src/ddmd/expressionsem.d
```diff
@@ -8771,6 +8784,101 @@ Expression binSemanticProp(BinExp e, Scope* sc)
...
+                    e2x = new XorExp(e.loc, e1x, e2x);
+                    break;
+                case TOKshlass:
+                    e2x = new ShlExp(e.loc, e1x, e2x);
+                    break;
+                case TOKshrass:
+                    e2x = new ShrExp(e.loc, e1x, e2x);
+                    break;
+                case TOKushrass:
+                    e2x = new UshrExp(e.loc, e1x, e2x);
```
ibuclaw: Another thing that sprung to mind, you'll have to assign the rewrite to another variable. If resolvePropertiesX doesn't return a value for whatever reason, you don't want to assign it to the original binassignexp.

Think `s.x += 1` being accidentally turned into `s.x += s.x + 1`

src/ddmd/expressionsem.d
JinShil: Fixed.  Thanks!

src/ddmd/expressionsem.d
JinShil: > In fact, you can just make the switch `final` and do away with the assert entirely!

I can't do that because the switch doesn't cover all possible `Token`s.

> Leaving the message as a comment in the source code is plenty good enough.

Done.

JinShil:
> Are we really sure we want to do this? This is a significant language change with essentially no discussion about its merits.

Honestly, it never occured to me that this would not be desired.  Like the other users who commented in Issue 8006 and Issue 16187, I was a bit dismayed to find out this wasn't implemented.  I was operating under the assumption that the only reason binary assignments didn't work with `@property` functions was because noone had gotten around to implementing them yet.

I'm willing to follow whatever process you'd like to reach a resolution.  Would you like me to write a justification in this pull request, make a forum post, write a DIP?

src/ddmd/expressionsem.d
JinShil: It took me a little while to see what you meant. but I got it.  Nice catch! Thanks.

src/ddmd/expressionsem.d
WalterBright: ok. Please add a summary of that explanation as a comment in the code.

WalterBright:
What would be best is a forum post with links to this PR and the bugzilla issues. Also, a spec PR would be needed.

BTW, I do think this is looking good.

src/ddmd/expressionsem.d
JinShil: > Please add a summary of that explanation as a comment in the code.

Done.

JinShil:
> Also, a spec PR would be needed.

Spec change: https://github.com/dlang/dlang.org/pull/1865

FYI: This does not address unary assignment operators (e.g. `s.x++`, `++s.x`).  I intend to implement that in a future pull request.

JinShil:
> What would be best is a forum post with links to this PR and the bugzilla issues.

http://forum.dlang.org/post/yumhltnuczivbdcmyhvw@forum.dlang.org

jpf91:
>Are we really sure we want to do this? This is a significant language change with essentially no discussion about its merits.

This has been discussed extensively. It's just been a few years since then and I don't even remember the conclusion anymore ;-)

http://www.prowiki.org/wiki4d/wiki.cgi?DocComments/Property#Semanticrewritingofproperties
https://wiki.dlang.org/DIP23#Applying_operators
https://wiki.dlang.org/Property_Discussion_Wrap-up#Proposals
And in some of the other old property DIPs as well.

JinShil:
I've discovered a problem with this pull request, and modified the test suite to deliberately cause it to fail, preventing premature merge.

The problem is, if the setter `@property` function is declared before the getter `@property` function, the binary assignment operators fail to compile:
```D
// This compiles for binary assignment operators
struct TGetterFirst
{
    int mX;
    @property int x() { return mX;}      // getter declared first
    @property void x(int v) { mX = v; }
}

// This does not compile for binary assignment operators
struct TSetterFirst
{
    int mX;
    @property void x(int v) { mX = v; }  // setter declared first
    @property int x() { return mX;}
}
```
I'll post an update when I find a fix.

ibuclaw:
> I'll post an update when I find a fix.

Looking at the error, it seems that resolveProperties just returns the first matched name, and doesn't match the parameters.

src/ddmd/expressionsem.d
jacob-carlborg: If there's a getter annotated with `ref` and a setter. If there's no rewrite for `ref` functions the setter will be bypassed.

jacob-carlborg:
I'm not sure how this is supposed to work but there should be tests verifying that:

* Free functions work/don't work
* Functions not annotated with `@property` work/don't work

src/ddmd/expressionsem.d
JohanEngelen: You should add tests that specifically test which functions are and are not called (i.e. besides testing that the calculation is performed correctly, also test that the getter/setter is/isn't called), both for the non-ref and ref case.

test/runnable/test8006.d
```diff
@@ -0,0 +1,402 @@
...
+    // Ensure a ref @property is not rewritten
+    // This was an odd corner case found in dub's source code. We
+    // test specifically for it to avoid breakage
+    TRefInt tri1;
+    TRefInt tri2;
+    tri1.mX ~= tri2;
+    tri2.mX ~= tri1;
+    tri1.x ~= tri2; // tri2.opBinaryRight!("~=")(tri1.x)
+    assert(tri1.mX.length == 2);
+}
```
JohanEngelen: Shouldn't this also not rewrite things for non-ref property? (with setter defined)
Better to make opBinaryRight do something different, so you can verify that it was called.

JinShil:
> If the setter `@property` function is declared before the getter `@property` function, the binary assignment operators fail to compile:

The problem turned out to be the sequence in which things were called.  If we call `e.e1.copy()` after `resolveProperties(sc, e.e1)`, the `copy()` call will return the wrong overload.  Why? I don't know.  For now I just change the order of the calls.

A recent merge (made my me, coincidentally) broke this, so I had to make a few additional changes to support `static alias this`.  Now the commit history's all fubar, but I'll try to fix that later after I address the other comments.

jacob-carlborg:
Do we need to test nested functions as well?

JinShil:
@WalterBright and @andralex: I've recently discovered [DIP23](https://wiki.dlang.org/DIP23#Applying_operators) which was authored by the two of you.  It proposes the following rewrite:

> If `a.prop` is a member variable, the expression `a.prop op= x` has the usual meaning. Otherwise, `a.prop op= x` gets rewritten twice. First rewrite is `(a.prop) op= x`, i.e. apply `op=` to the result of the property. Second rewrite is `a.prop = a.prop op x`. If only one of the two rewrite compiles, use it. If both compile, fail with ambiguity error.

This PR current rewrites as `a.prop(a.prop op x)`.  Would DIP23's proposal be better, and can you justify why?

WalterBright:
> This PR currently rewrites as a.prop(a.prop op x). Would DIP23's proposal be better, and can you justify why?

I'd forgotten about that. I don't remember the rationale, but it's possibly about if the getter returns a `ref` or other lvalue, and the case where there is no setter.

JinShil:
>  I don't remember the rationale, but it's possibly about if the getter returns a ref or other lvalue, and the case where there is no setter.

Well, I implemented it and tested it, but it resolves the ambiguity condition with `ref @property` functions.  I think the DIP23 algorithm requires an additional rule specified in DIP23.

> In brief, the "()" operator may NEVER be applied EXPLICITLY to a function annotated with @property. THIS IS A CHANGE OF SEMANTICS. 

I don't want to change existing behavior with this PR.  This PR excludes `@ref properties`, and the DIP23 algorithm would also require excluding them if I don't want to change existing behavior.  So I can still use either one, and I'm not sure which is best.  The current implementation is simpler, though.

@CyberShadow I accidentally closed this PR and reopened it and how DAutoTest fails with "Pull request SHA mismatch".  Is there anything I can do about that?

ZombineDev:
> DAutoTest fails with "Pull request SHA mismatch"

That's unfortunately a bug in GitHub. As far as I know, we have contacted GitHub's support, but since then there has been no resolution to the problem.
The way to solve this is by changing either rebasing your PR or opening a new one *from a different branch*.

JinShil:
> Do we need to test nested functions as well?

Well, this won't even compile today...

```D
void main()
{
    int mP;
    @property int p() { return mP; }
    @property void p(int v) { mP = v; }
}
```
```
Error: declaration p is already defined
```
... because, apparently, you can't overload nested functions (at least that's what the compiler error implies), so I can't create getter/setter pairs for the binary assignment operators to work.  But I added tests for nested `ref @property` functions that have a single getter.

JinShil:
I played around with the implementation a little more.  I tried implementing [DIP23's algorithm](https://wiki.dlang.org/DIP23#Applying_operators), but that just resulted in too much breakage.  I think I would need to implement other parts of DIP23 to make it work, but that would introduce new semantics.  So, I'm back to the original implementation (i.e. `e1 += e2` is rewritten to `e1(e1 + e2)` and explicitly excluding `ref @property` functions). 

In the process of exploring alternate implementations, I learned a lot more about how this implementation interacts with other D constructs like free functions, nested functions, `opBinary` overloads, and `opOpBinary` overloads.  I've added tests for those to ensure proper behavior and guard against breakage.  This pull requests attempts to only add new behavior, not change existing behavior.

I think this is ready for at least more scrutiny and potentially a merge.

jacob-carlborg:
> Well, this won't even compile today...

Ah, of course. I didn't really think that through.

sysint64:
@JinShil Great work! I really hope that this PR will be accepted. @WalterBright, @ibuclaw when will fate of this PR be known?

ibuclaw:
> @WalterBright, @ibuclaw when will fate of this PR be known?

From a technical review, I'd say that this is sound. I'll defer it to @WalterBright to decide if this passes policy review.

JinShil:
Added the following:

If the value returned from a `@property` function is an aggregate that overloads the binary assignment operators (i.e. `opOpAssign`) there can be an ambiguity between the rewrite `prop(prop() op x)` and `prop() op= x`.  The latest implementation in this pull request excludes aggregates with an `opOpAssign` overload from the rewrite to avoid breaking existing code.  One such breakage that was found by the jenkins test suit was [in botan](https://github.com/etcimon/botan/blob/4da3eacfb821ed40d433bebf4d8ce751f6541b1b/source/botan/algo_base/scan_token.d#L75).

That condition was likely the motivation behind [DIP23](https://wiki.dlang.org/DIP23#Applying_operators)'s...

>  If `a.prop` is a member variable, the expression `a.prop op= x` has the usual meaning. Otherwise, `a.prop op= x` gets rewritten twice. First rewrite is `(a.prop) op= x`, i.e. apply `op=` to the result of the property. Second rewrite is `a.prop = a.prop op x`. If only one of the two rewrite compiles, use it. If both compile, fail with ambiguity error. 

However, if DIP23's algorithm were implemented, existing code would break with the ambiguity error, so the implementation in this pull request gives precedence to the operator overload to avoid said breakage.

The [spec change pull request](https://github.com/dlang/dlang.org/pull/1865) was also updated to reflect this.

From my perspective this pull request is rebased in ready to go.

ZombineDev:
IIUC, the current behavior is more consistent with the rest of the language because of the rule "the most specific overload wins" (opOpAssign is considered more specific than obj.assign(obj.op(arg)).

JinShil:
> codecov:  Missing base report.  Unable to compare commits because the base of the compare did not upload a coverage report.

I don't know what to do about the codecov checks.

JinShil:
Tests are now passing.  This is ready to go.

andralex:
There are enough subtleties to warrant going through a DIP. As an example of a subtlety, the spec change does not mention whether `e1` is evaluated once or twice in the expression `e1.prop @= e2`.

I think the right way to go is to define the expression as a lowering like this:

`e1.prop @= e2`

==>

`((auto ref x) => x.prop = x.prop @ e2)(e1)`

Ironically I think the lowering does not compile because of an unrelated limitation - auto ref lambdas don't work. Nevertheless the intent is to evaluate `e1` first (without copying it if it's an lvalue), then run the getter against it, compute the operator, then run the setter.

JinShil:
> the spec change does not mention whether `e1` is evaluated once or twice in the expression `e1.prop @= e2`.

The spec change says `e1.prop @= e2`  is lowered to `e1(e1() @ e2)`, though it used the notation `setter(getter() @ e2)` to be more explicit about the following...

The lowering requires two `@property` functions:  `T e1() @property` and `T e1(T v) @property`.  That is, it is ambiguous to ask whether "`e1` is evaluated once or twice" -- which `e1` are you referring to; the getter or the setter?  Regardless, the lowering illustrated above shows that each of those functions is only ever evaluated once.  If one of them doesn't exist, the compiler will produce an error.

Properties (the concept as implemented in the D language) are nothing more than functions, so it is logical to define the lowering in terms of those functions.  When you use the notation `x.prop = x.prop @ e2`, it is ambiguous what `x.prop` is:  

- Is the first occurrence a setter `@property` function and the second occurrence a getter `@property` function?
- Or, did you imply that there is only one `@property` function that returns a value by reference?  
- Or, perhaps, you intended it to work the same for both?

The spec change already clarifies that:

> **If the getter does not return by `ref`**, the return value is not an aggregate that overloads the binary assignment operator, and a setter is defined, binary assignment is equivalent to `setter(getter() @ value)`, where `@` is the operation.

In other words, `@property` functions that return by `ref` are excluded from the lowering, and evaluated as they currently exist today to avoid breakage and keep the scope of this feature contained.

> There are enough subtleties to warrant going through a DIP

I[ specifically asked if a DIP was required](https://github.com/dlang/dmd/pull/7079#issuecomment-322156400), and [@WalterBright said a forum post would be sufficient](https://github.com/dlang/dmd/pull/7079#issuecomment-322297542).  Please discuss with @WalterBright and come back with a unified response.  Although, I'm very much _not_ looking forward to the DIP process, I'll do whatever needs to be done.

jpf91:
> That is, it is ambiguous to ask whether "e1 is evaluated once or twice" (assuming you actually mean e1.prop)

I think Andrei is referring to side-effects of `e1`. Consider this full example:
```d
struct Foo 
{
    int _x;
    @property int X() {return _x;}
    @property void X(int value) {_x = value;}
}

Foo _foo;
int _global = 0;

Foo* getFoo()
{
    _global++;    
    return _foo;
}

void main()
{
     getFoo().X += 1;
     // _global == 2 or _global == 1?
}
```

And Andrei's cryptic lambda rewrite simply means to cache the result of `e1`, i.e. only evaluate once. I guess the lambda is necessary cause if you assign to a temporary variable in frontend-terms you'll need to call copy constructors etc. In GDC backend terms, we'd simply evaluate `e1` using GCCs `SAVE_EXPR` which just generates temporary variables without all the implied frontend semantics (destructors, copy ctors, ....)

JinShil:
> I think Andrei is referring to side-effects of `e1`. 

Thanks @jpf91, I understand now... though I was happier in my ignorance.

JinShil:
So I suppose the rewrite should actually be...

`e1.prop @= e2` ==> `((auto ref x) => x.prop(x.prop() @ e2))(e1)`

Yes?

jpf91:
> e1.prop @= e2 ==> ((auto ref x) => x.prop(x.prop() @ e2))(e1)

I know Andrei likes the lowering idiom, but I think in this case there may be a better solution. Lowering to a lambda now requires the backend to inline & optimize this lambda or explicitly detect that the lambda is used as a property binop and undo the frontend processing. This only adds useless work for the backend.

We usually handle such one-time evaluation of side effects in the backend, where we have simple & efficient functions which do not depend on optimizations and do not produce useless work for the backend. For example, see https://github.com/D-Programming-GDC/GDC/blob/master/gcc/d/expr.cc and search for `stabilize_reference` and `d_save_expr`. Maybe it's better to introduce a `BinPropAssignExp` class which contains the getter & setter to call and leave saving of `e1` to the backend?

Also ping @ibuclaw

andralex:
Liking is not the term - lowering has been very successful every time we applied it because it made it easy to understand the behavior of complex language constructs in terms of simpler ones.

The lowering is in the semantics definition. We don't need to actually lower to lambdas in the implementation, just make sure the behavior is exactly the same.

andralex:
>I guess the lambda is necessary cause if you assign to a temporary variable in frontend-terms you'll need to call copy constructors etc.

It's a matter even in the case of lvalue - consider e.g. `arr[i++].prop += 1`. In such cases we need to make sure `i` is only incremented once.

test/runnable/test8006.d
```diff
@@ -544,7 +544,7 @@ int mXret;
 @property int xret() { return mXret; }
 @property int xret(int v) { return mXret = v; }
 
-// Same as testFreeFunctions except we want to
+// Same as testFreeFunctions except the we want to
```
ibuclaw: Except that?

test/runnable/test8006.d
```diff
@@ -0,0 +1,869 @@
...
+    assert(mX == 0x5);
+    x ^^= 2;
+    assert(mX == 25);
+}
+
+int mXret;
+@property int xret() { return mXret; }
+@property int xret(int v) { return mXret = v; }
+
+// Same as testFreeFunctions except the we want to
```
ibuclaw: Except that?

test/runnable/test8006.d
```diff
@@ -0,0 +1,869 @@
...
+    assert(mP == 0x2);
+    p |= 0x8;
+    assert(mP == 0xA);
+    p ^= 0xF;
+    assert(mP == 0x5);
+    p ^^= 2;
+    assert(mP == 25);
+}
+
+// Same as testNestedFunctionsIntRef except the we want to
```
ibuclaw: Except that

test/runnable/test8006.d
```diff
@@ -0,0 +1,869 @@
...
+{
+    string mP;
+    @property ref string p() { return mP; }
+
+    mP = "abc";
+    p ~= "def";
+    assert(mP == "abcdef");
+}
+
+// Same as testNestedFunctionsStringRef except the we want to
```
ibuclaw: Except that

JinShil:
Implemented the following for `e1.prop @= e2`:
if `e1` is a type (e.g. `static @property` functions) then lower to `e1.x = e1.x @ e2`, otherwise lower to `((auto ref _e1) => _e1.prop = _e1.prop @ e2)(e1)`

JinShil:
From my perspective, this is ready to go.

@WalterBright: If a DIP is required to get this pulled, please explicitly say so.  If there's anything more I can do to move this forward, please say so.  If this is just not a priority to you right now, that would also be helpful to know.  Thanks!

MartinNowak:
We obviously have temp variables in dmd as well, e.g. used in foreach range loops. Also the backend avoids adding more side-effects.
Seems like a cleaner solution than to intransparently convert this to a lambda (which needs some mangling and unique name https://issues.dlang.org/show_bug.cgi?id=14894).

MartinNowak:
> @WalterBright: If a DIP is required to get this pulled, please explicitly say so. If there's anything more I can do to move this forward, please say so. If this is just not a priority to you right now, that would also be helpful to know. Thanks!

I think it's small and simple enough to not warrant a DIP, but it definitely need a changelog entry and spec update.

JinShil:
> ... it definitely need a changelog entry and spec update.

Spec change: https://github.com/dlang/dlang.org/pull/1865
Changelog entry will come when this earns support from @WalterBright and a merge seems likely.

> We obviously have temp variables in dmd as well, e.g. used in foreach range loops.  Seems like a cleaner solution than to intransparently convert this to a lambda (which needs some mangling and unique name.

I'd be happy to change the implementation, but I need someone to elaborate (even spell out) how such an implementation would achieve the same outcome; it's only because of @ibuclaw's and @jpf91's help that I've been able to understand and implement what I've done thus far.  It would also be helpful to know if such an implementation would be generally preferred by the DMD core developers, so I have more confidence that the implementation would be approved.

ibuclaw:
You can see the change that introduced it in #5636.  This PR has changed somewhat since I last sat down and reviewed it, so you'd have to wait a short while I accustom myself to what its doing again.

WalterBright:
> This has been discussed extensively. It's just been a few years since then and I don't even remember the conclusion anymore ;-)

That's kind of a problem, and I think the "best of" the disparate discussions should be distilled into a DIP so we all know what we're talking about.

JinShil:
> You can see the change that introduced it in #5636.

Thanks! I see the utility of those functions, but what is @MartinNowak suggesting the lowering be changed to?

JinShil:
> > This has been discussed extensively. It's just been a few years since then and I don't even remember the conclusion anymore ;-)

>That's kind of a problem, and I think the "best of" the disparate discussions should be distilled into a DIP so we all know what we're talking about.

After reviewing the DIPs and their discussions, it appears that property debacle with UFCS, optional parens, etc... has indeed been discussed extensively, but this specific issue of binary assignment operators for properties has not.  The only serious mentions seem to be the [prowiki article](http://www.prowiki.org/wiki4d/wiki.cgi?DocComments/Property#Semanticrewritingofproperties), and [DIP23](https://wiki.dlang.org/DIP23#Applying_operators).

Nevertheless, a DIP PR has been submitted at dlang/DIPs#97.  At present it's still a work-in-progress, but it's coming together.

ibuclaw:
Thanks @JinShil. This is great stuff.

JinShil:
https://github.com/dlang/phobos/blob/225ea95397f943944b2081a34d11054884085d27/std/typecons.d#L2570-L2573

Should that really be a `@property`?

JinShil:
I need some help.  I want to ensure that the getter and setter both contain an `@property` attribute.  If I only have the getter, how can I get the `TypeFunction` of the setter?

ibuclaw:
If there's only a getter, wouldn't that be a readonly property?

JinShil:
> I need some help. I want to ensure that the getter and setter both contain an @property attribute. If I only have the getter, how can I get the TypeFunction of the setter?

I came up with [this](https://github.com/JinShil/dmd/blob/aac91e060bd51f2dee24c7d2dc182b57e1f3054b/src/ddmd/expressionsem.d#L9508-L9530).  I doubt it'll pass review, but it works.  The meat and potatoes of the DIP is now implemented and tests are passing.  I just need to redo the tests and finalize the DIP.

JinShil:
> I came up with this. I doubt it'll pass review, but it works. The meat and potatoes of the DIP is now implemented and tests are passing. I just need to redo the tests and finalize the DIP.

Turned out to be moot anyway as `@property` function and ordinary function overloads can't coexist.

JinShil:
The implementation now matches that which is proposed in the DIP.

ibuclaw:
@WalterBright @andralex - are we progressing this? Is there a card for this in trello?

JinShil:
> are we progressing this?

According to [a comment from @mdparker on Nov 14th, 2017](https://github.com/dlang/DIPs/pull/99#issuecomment-344201432) we are waiting on changes to the DIP process.

@WalterBright and @andralex Can you please give @mdparker whatever support he needs to unplug the DIP queue?

> Is there a card for this in trello?

I tried to add one, but I guess I don't have permission on the Trello board.  I added a card [here](https://github.com/dlang/dmd/projects/1#card-6457943)  though that probably won't do much if we don't all agree to use that GitHub feature.

wilzbach:
> I tried to add one, but I guess I don't have permission on the Trello board. I added a card here, though that probably won't do much if we don't all agree to use that GitHub feature.

Sent out an invitation to you, but the Trello board isn't very active and only used by Martin, so it's probably better to give the GitHub board a try.

mdparker:
@JinShil The hold up right now is me. I haven't finished revising the documentation yet. I'm almost there, though.

JinShil:
FYI. I am abandoning this implementation and the DIP. I'm afraid my future with D is uncertain at best. If anyone wants to pick of the torch, it's yours. I'll keep the PRs open for now unless someone believes they should be closed.

12345swordy:
cc @thewilsonator 
I am interested in picking up the dip where the JinShill have left off. However I am not a compile writer so I can't guarantee to managed this pull request.

PR #7080, opened 2017-08-13
Add overrides for __simd intrinsics.

ibuclaw:
This is blocking #7065.  Unless it's considered OK for `__simd` to require an explicit cast on the return value, in which case, I'll close this and fix up all uses instead of `__simd` instead.

This change is very mechanical.


ibuclaw:
Can do the same with templates.  However there's a backend ICE blocking.  This is no longer needed however.

PR #7081, opened 2017-08-13
Turn informational messages of transition=complex into deprecated

ibuclaw:
I think it's high time to push forward with this.

Also ironed out a small problem I found when building phobos with this turned on.

https://issues.dlang.org/show_bug.cgi?id=14488


wilzbach:
I don't think people will use this flag, so what stops us from turning it into a deprecation message for everyone?

ibuclaw:
> I don't think people will use this flag, so what stops us from turning it into a deprecation message for everyone?

And identifying and removing (or fixing) all uses of complex or imaginary types in druntime and phobos.

src/ddmd/mtype.d
```diff
@@ -3058,16 +3058,23 @@ extern (C++) abstract class Type : RootObject
...
     final void checkComplexTransition(Loc loc)
     {
         Type t = baseElemOf();
         while (t.ty == Tpointer || t.ty == Tarray)
             t = t.nextOf().baseElemOf();
+        if (t.ty == Tenum)
+            t = (cast(TypeEnum)t).sym.memtype;
+
+        // Type is forward referenced or not a scalar type.
+        if (t is null || !t.isTypeBasic())
```
WalterBright: Is the check for typebasic really necessary? This should be covered by line 3076. Also, the null check seems unnecessary. Is there a test case for this? According to the previous logic, it could only be null if `memtype` is null. Can this ever happen?

src/ddmd/mtype.d
```diff
@@ -3091,11 +3098,13 @@ extern (C++) abstract class Type : RootObject
             }
             if (t.iscomplex())
             {
-                fprintf(global.stdmsg, "%s: use of complex type '%s' is scheduled for deprecation, use 'std.complex.Complex!(%s)' instead\n", p ? p : "", toChars(), rt.toChars());
+                deprecation(loc, "use of complex type `%s` is scheduled for deprecation, use `std.complex.Complex!(%s)` instead\n",
```
WalterBright: It's clearly not `scheduled` for deprecated, it is deprecated, as it calls `deprecation()`

test/compilable/sw_transition_complex.d
```diff
@@ -134,3 +134,10 @@ creal test14488b()
     return 1 + 0i;
 }
 
+// Forward referenced types shouldn't cause spurious errors during test for complex or imaginary.
+enum E;
```
WalterBright: Is this the test case for the `null` t?

test/compilable/sw_transition_complex.d
```diff
@@ -134,3 +134,10 @@ creal test14488b()
...
+struct S;
```
WalterBright: I don't think this case is strictly necessary, but no problem leaving it in.

test/compilable/sw_transition_complex.d
ibuclaw: Yes, the error is in getMemType if I recall correctly.

ibuclaw:
Should rebase.

mathias-lang-sociomantic:
Why not make it a deprecation instead (without the switch) ?
This switch has been there for a while, people interested in it have plenty of DMD releases to use to fix it.

ibuclaw:
@mathias-lang-sociomantic I found a small bug when building phobos with it turned on.  So I'm not sure if people were using it.

I think it would be better to first push deprecations in phobos, and the compiler/druntime interface updated.  Then turn on deprecation by default.

wilzbach (approved):
LGTM. What's blocking this?
FWIW I agree that the deprecation messages should be turned on by default as no one will be using this switch, but this can always be done in a separate PR ...

JinShil:
@ibuclaw I'm itching to pull this.  Is it ready to go?  Is there anything outstanding?

JinShil:
We also need an update to the [deprecations table](https://dlang.org/deprecate.html), but I don't want it to hold up this PR.

JinShil (approved):

src/ddmd/mtype.d
```diff
@@ -3002,16 +3002,23 @@ extern (C++) abstract class Type : RootObject
...
     final void checkComplexTransition(Loc loc)
     {
         Type t = baseElemOf();
         while (t.ty == Tpointer || t.ty == Tarray)
             t = t.nextOf().baseElemOf();
+        if (t.ty == Tenum)
+            t = (cast(TypeEnum)t).sym.memtype;
+
+        // Type is forward referenced.
+        if (t is null)
```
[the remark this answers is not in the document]
ibuclaw: The comment should probably be anonymous enum.  The ICE here was because an enum was declared but never defined. See test.

JinShil:
@ibuclaw If this is ready to go, please rebase and I'll merge.

ibuclaw:
There's no merge conflicts, why rebase?

I'll update the comment to clarify why there is a null check for enums.

@WalterBright are you blocking this?

ibuclaw:
Ah, now the page is telling me that there are conflicts, and dlang bot has added a tag.

JinShil:
> There's no merge conflicts, why rebase?

The GitHub web interface seems to think there is a conflict for `src/ddmd/mars.d` and `src/ddmd/mtype.d`  I think because of the transition from `ddmd` to `dmd`. 

Either that or this is one of those old PRs that has reached GitHubs 1000 status limit and is no longer updated.  In that case, a force push should reset it.

JinShil:
@ibuclaw Sorry, but can you please rebase again; this is why we need to act swiftly on PRs that are ready to go.

 If you want to wait for @WalterBright to sign off on this first, please explicitly say so, otherwise, I'll merge right after you rebase.

ibuclaw:
@JinShil - rebased.  If you think that all concerns have been addressed sufficiently, then go ahead and merge.

PR #6101, opened 2016-09-01
introduce __ctfeWrite for strings.

UplinkCoder:
This enables __ctfeWrite for strings.
Which already is a useful debugging tool.
Using it with anything but strings is a compile error.
Since it writes to stdout, it is prone to race conditions.
However as long as ctfe is not multi-threaded  it should not be too problematic.

PR #7082, opened 2017-08-15
Implement __ctfeWrite builtin

MoritzMaxeiner:
Based closely on @UplinkCoder's work in closed PR #6101.

### DIFF
--- a/src/dmd/builtin.d
+++ b/src/dmd/builtin.d
@@ -216,6 +216,37 @@ extern (C++) Expression eval_yl2xp1(Loc loc, FuncDeclaration fd, Expressions* ar
     return new RealExp(loc, result, arg0.type);
 }
 
+extern (C++) Expression builtin_ctfeWrite(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    import core.stdc.stdio : fprintf;
+    import ddmd.ctfeexpr : resolveSlice;
+
+    Expression arg0 = (*arguments)[0];
+    StringExp se;
+
+    switch (arg0.op)
+    {
+        case TOKstring:
+            se = cast(StringExp) arg0;
+            break;
+        case TOKslice:
+            se = resolveSlice(arg0).toStringExp();
+            break;
+        case TOKarrayliteral:
+            se = arg0.toStringExp();
+            break;
+        default:
+            break;
+    }
+
+    if (se)
+    {
+        fprintf(global.stdmsg, "%.*s".ptr, se.len, se.string);
+    }
+
+    return arg0;
+}
+
 public extern (C++) void builtin_init()
 {
     builtins._init(47);
@@ -332,6 +363,8 @@ public extern (C++) void builtin_init()
     // @safe @nogc pure nothrow int function(ulong)
     if (global.params.is64bit)
         add_builtin("_D4core5bitop7_popcntFNaNbNiNfmZi", &eval_popcnt);
+    // @safe @nogc pure nothrow void function(const string)
+    add_builtin("_D6object11__ctfeWriteFNaNbNiNfxAyaZv", &builtin_ctfeWrite);
 }
 
 /**********************************
--- a/src/dmd/builtin.d
+++ b/src/dmd/builtin.d
@@ -347,7 +347,7 @@ public extern (C++) void builtin_init()
     if (global.params.is64bit)
         add_builtin("_D4core5bitop7_popcntFNaNbNiNfmZi", &eval_popcnt);
     // @safe @nogc pure nothrow void function(const string)
-    add_builtin("_D6object11__ctfeWriteFNaNbNiNfxAyaZv", &builtin_ctfeWrite);
+    add_builtin("_D6object11__ctfeWriteFNaNbNiNfMAxaZv", &builtin_ctfeWrite);
 }
 
 /**********************************
--- a/src/dmd/builtin.d
+++ b/src/dmd/builtin.d
@@ -224,7 +224,7 @@ extern (C++) Expression builtin_ctfeWrite(Loc loc, FuncDeclaration fd, Expressio
 
     if (StringExp se = arg0.toStringExp)
     {
-        fprintf(global.stdmsg, "%.*s".ptr, se.len, se.string);
+        fprintf(global.stdmsg, "%.*s".ptr, cast(int) se.len, se.string);
     }
 
     return arg0;
--- a/src/dmd/builtin.d
+++ b/src/dmd/builtin.d
@@ -219,27 +219,10 @@ extern (C++) Expression eval_yl2xp1(Loc loc, FuncDeclaration fd, Expressions* ar
 extern (C++) Expression builtin_ctfeWrite(Loc loc, FuncDeclaration fd, Expressions* arguments)
 {
     import core.stdc.stdio : fprintf;
-    import ddmd.ctfeexpr : resolveSlice;
 
     Expression arg0 = (*arguments)[0];
-    StringExp se;
 
-    switch (arg0.op)
-    {
-        case TOKstring:
-            se = cast(StringExp) arg0;
-            break;
-        case TOKslice:
-            se = resolveSlice(arg0).toStringExp();
-            break;
-        case TOKarrayliteral:
-            se = arg0.toStringExp();
-            break;
-        default:
-            break;
-    }
-
-    if (se)
+    if (StringExp se = arg0.toStringExp)
     {
         fprintf(global.stdmsg, "%.*s".ptr, se.len, se.string);
     }
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -5938,11 +5938,11 @@ extern (C++) final class SliceExp : UnaExp
 
     override StringExp toStringExp()
     {
-    	if (e1.op == TOKstring)
-    	{
+        if (e1.op == TOKstring)
+        {
             return e1.toStringExp;
-    	}
-    	return null;
+        }
+        return null;
     }
 
     override void accept(Visitor v)
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -5936,6 +5936,15 @@ extern (C++) final class SliceExp : UnaExp
         return e1.isBool(result);
     }
 
+    override StringExp toStringExp()
+    {
+    	if (e1.op == TOKstring)
+    	{
+            return e1.toStringExp;
+    	}
+    	return null;
+    }
+
     override void accept(Visitor v)
     {
         v.visit(this);
--- a/test/compilable/testCTFEWrite.d
+++ b/test/compilable/testCTFEWrite.d
@@ -1,0 +1,36 @@
+/*
+TEST OUTPUT:
+0^^2 == 1
+1^^2 == 1
+2^^2 == 4
+3^^2 == 9
+4^^2 == 16
+5^^2 == 25
+6^^2 == 36
+result == 91
+*/
+int sum_of_sq(int x) pure nothrow @safe
+{
+    import std.conv;
+
+    string newline;
+    newline = "\n";
+
+    int result = 0;
+    foreach (i; 0 .. x)
+    {
+        __ctfeWrite(to!string(i));
+        __ctfeWrite("^^2 == ");
+        int power = i ^^ 2;
+        __ctfeWrite(power.to!string);
+        __ctfeWrite(newline);
+        result += power;
+    }
+    __ctfeWrite("result == ");
+    __ctfeWrite(to!string(result));
+    __ctfeWrite(newline);
+    return result;
+
+}
+
+static assert(sum_of_sq(7) == 91);


UplinkCoder:
@Calrama Nice Work!

src/ddmd/builtin.d
```diff
@@ -216,6 +216,37 @@ extern (C++) Expression eval_yl2xp1(Loc loc, FuncDeclaration fd, Expressions* ar
...
+    Expression arg0 = (*arguments)[0];
+    StringExp se;
+
+    switch (arg0.op)
+    {
+        case TOKstring:
+            se = cast(StringExp) arg0;
+            break;
+        case TOKslice:
+            se = resolveSlice(arg0).toStringExp();
```
UplinkCoder: you should really check if this succeeds.

src/ddmd/builtin.d
```diff
@@ -216,6 +216,37 @@ extern (C++) Expression eval_yl2xp1(Loc loc, FuncDeclaration fd, Expressions* ar
...
+            break;
+        case TOKarrayliteral:
+            se = arg0.toStringExp();
```
UplinkCoder: ditto

UplinkCoder (changes requested):
toStringExp can return null.
please check for that otherwise you crash the compiler

src/ddmd/builtin.d
Calrama: AFAIU, if it doesn't succeed, then `toStringExp()` return `null`, i.e. `se` will remain `null` and the check in line 242 won't succeed. If you are referring to `resolveSlice(arg0)`, AFAIU from its body, it can't return `null`, so no `null` deref at the dot.

src/ddmd/builtin.d
Calrama: The success check is in line 242 (same as with the above case), as failure means `se` remains `null`.

Calrama:
> toStringExp can return null.
> please check for that otherwise you crash the compiler

Which is why [line 242](https://github.com/dlang/dmd/pull/7082/files/4255a8c3ae6824aad75a1aec7889853e5bd253cc#diff-743e04b36f9679067c8989e77f0ef9b5R242) checks for that.

UplinkCoder (approved):

UplinkCoder:
@klickverbot you might want to have a look

ZombineDev:
@Calrama Thanks for resurrecting this important work!

I think it would be reasonable to expect that `__ctfeWrite(...)` would support the same featureset that `pragma(msg, ...)` does currently.
Have you looked into reusing parts of the implementation?

Calrama:
@ZombineDev 
> I think it would be reasonable to expect that __ctfeWrite(...) would support the same featureset that pragma(msg, ...) does currently.

I'll take a look, but I'm not sure what the difference in features is right now (could you list them for posterity?).

> Have you looked into reusing parts of the implementation?

I haven't, because the happen at different phases at compile time and I'm not familier enough with dmd's source to be confident that the same code would work in both phases.

MetaLang:
This is one case where __ctfeWrite can't do the same thing as pragma(msg):
```
int n;
pragma(msg, typeof(n)); //prints 'int'

__ctfeWrite(typeof(n)); //Should be an error
```

For __ctfeWrite you'd have to add `.stringof` to the type.

Calrama:
> This is one case where __ctfeWrite can't do the same thing as pragma(msg):

Right, I don't think we can change that, though (and I don't think I want to).

The issue with multiple arguments (which can be of any type) remains that you'd have to call some form of to!string, then. It would also require changing the function signature in druntime to a variadic function (I don't think we can use templates here because we need to bind the builtin to the mangled name when dmd is compiled)

ZombineDev:
@Calrama Thinking over about `pragma (msg, ...)` again, apart from the automatic `type.stringof` conversion that @MetaLang mentioned, the only significant thing left is variadic arguments. However, we can easily add a variadic `ctfeWriteln` function in druntime, without much additional effort, so I don't think you need to worry about a more complicated implementation:

```D
void ctfeWriteln(Args...)(Args args)
if (allSatisfy!(isConvertibleToString, Args))
{
    foreach (arg; args)
        __ctfeWrite(arg);

    __ctfeWrite("\n");
}
```

What's missing and what may be a big performance problem (especially with the current CTFE engine) is formatting non-string arguments.
`to!string` doesn't support all that much at CTFE and you'll have a very hard time if you try to add floating-point support there.

So, for performance and ease of implementation I suggest we add the following string formatting functionality:

```D
@safe pure nothrow @nogc:

// I'm not sure if only a single character overload (dchar) will do:
string __ctfeFormat(char c);
string __ctfeFormat(wchar w);
string __ctfeFormat(dchar d);

string __ctfeFormat(bool b);

// Just need to support the biggest signed and unsigned integer type
static if (is(cent))
{
    string __ctfeFormat(cent c);
    string __ctfeFormat(ucent c);
}
else
{
    string __ctfeFormat(long l);
    string __ctfeFormat(ulong u);
}

// Need to have overloads for all distinct floating-point types
// to avoid loss of precision, because of rounding.

string __ctfeFormat(float f);
string __ctfeFormat(double d);
string __ctfeFormat(real r);

/* Built-in complex types are deprecated, so I think we don't need to support them. */
```
And of course leave derived types (arrays and AAs) and user defined types to `std.conv` / `std.format`.

With that we need only a single `__ctfeWrite` built-in:
```D
@safe pure nothrow @nogc
void __ctfeWrite(scope const(char)[] str); 

// (the current __ctfeWrite(string) signature is wrong and needs to go)
```

And we can define `ctfeWriteln` like this:
```D
void ctfeWriteln(Args...)(Args args)
{
    import std.traits : isScalarType;
    foreach (arg; args)
        static if (isScalarType!(typeof(arg)))
            arg.__ctfeFormat.__ctfeWrite();
        else
        {
            import std.conv : to:
            arg.to!string.__ctfeWrite();
            // The above will work after we use `__ctfeFormat` for __ctfe in to!string
        }
        

    __ctfeWrite("\n");
}
```

(Not important, just thinking out load) and perhaps make this change to `std.stdio`:
```D
void writeln(T...)(T args)
{
    if (__ctfe)
        ctfeWriteln(args);
    else
    {
        // Business as usual...
    }
}
```

(The above shouldn't be a change in behavior, because the current `std.stdio.writeln` doesn't work at CT.)

Calrama:
@ZombineDev 

I was thinking more along the lines of making `__ctfeWrite`'s signature be
```d
void __ctfeWrite(...); 
``

so the dmd builtin implementation can take care of everything.
But your idea makes sense to me, as well, except that your `ctfeWriteln` cannot be in druntime, since it depends on phobos (the std.conv). I would say add ctfeWrite/ctfeWriteln to std.stdio, and update std.stdio.write/writeln as you suggested.

**Comment by ZombineDev** [comment]
> I was thinking more along the lines of making __ctfeWrite's signature be
void __ctfeWrite(...) [attributes];
so the dmd builtin implementation can take care of everything.

I don't think that would be easily with `add_builtin`, since it takes the mangled name a of the function, you would need go with a template, but I don't see any need do so, since there would be nothing to put in the function body.
 
> ctfeWriteln cannot be in druntime, since it depends on phobos

The suggestion was valid for the first implementation, but yes the general idea is to put built-ins and low-level implementation details in druntime and to have high-level wrappers (such as my second version of the function) in phobos.

**Comment by Calrama** [comment]
> I don't think that would be easily with add_builtin, since it takes the mangled name a of the function, you would need go with a template, but I don't see any need do so, since there would be nothing to put in the function body.

AFAIK variadic functions have fixed mangling, regardless of what you pass in, so using them is fine.

> the general idea is to put built-ins and low-level implementation details in druntime and to have high-level wrappers (such as my second version of the function) in phobos.

Ok.

**Comment by Calrama** [comment]
@UplinkCoder Things became much simpler after overriding `toStringExp` for `SliceExp`.

@ZombineDev Implementation based on your idea: https://github.com/Calrama/druntime/commit/e9d2d22d88fb5c4e98dfeb5a27f719cf08ffcbfd https://github.com/Calrama/dmd/commit/3ee8cd9c79cc1e55826ddf271bc926f4def33083 The only thing I'm unhappy about is that I wasn't able to discern a way to inject a FormatException caught in __ctfeFormat as a return value so it can be caught in CTFE. As a result, you can't catch FormatExceptions thrown in __ctfeFormat, they crash the compiler :/

**Comment by ZombineDev** [comment]
About the issue with exceptions, you could emit a compiler error upon catching the invalid format exception. Not as flexible or elegant as actually throwing an exception in CTFE (or returning via tuple or ref parameter), but it makes the function usable in `nothrow` code, which is very important IMO.

Using `std.format` is a clever way of implementing this, but it's not workable because:
* Using Phobos is not allowed in ddmd.
* Your tying the language specification (the interpretation of format argument) to exact version of Phobos that the host compiler used to build ddmd had.
* While I find `std.format` really useful in general, I think its feature-set / interface surface is too complex to bake into the language, even we agree on a fixed subset for the point above

I think it would be much simpler to go with a `__ctfeToString(<ScalarType> s)` built-in.

> The reason why I added the format specifier is that people will want to at the very least be able to customize float formatting and then it's easier to enable it for all of them.

I agree that this would be quite useful, but we would need to really limit and specify the exact feature set.

> Calrama/druntime@e9d2d22

Can you please fix `__ctfeWrite`'s signature from taking `string` to `scope const(char)[] str` ? `string` is quite limiting.

**Comment by UplinkCoder** [comment]
@ZombineDev __ctfeWrite shall only accept strings.
Everything else is going to be a nightmare to support.

tbh. ctfeWrite in it's current form is already hard to support properly.
since it requires us to have a mechnism to gather and transport this side-channel data,
without impacting the other execution of ctfe.

Wether the particular string is of type `scope const(char)[]` is of no concern to me.
Change it if you wish.

**Comment by ZombineDev** [comment]
@UplinkCoder I don't propose a making `__ctfeWrite` taking anything else than `char` arrays. I propose only  adding `__ctfeToString` for built-in scalar types, because otherwise outputting  floating-point values would be next to impossible.

**Comment by Calrama** [comment]
> it makes the function usable in nothrow code, which is very important IMO.

Using theoretically throwing functions (you know aren't going to throw in practice) in nothrow is easy: `throwingFunction.assumeWontThrow`. Otherwise, you can just catch exceptions in nothrow code, so I don't consider this a good argument against format specifiers.

> Using Phobos is not allowed in ddmd.

Citation needed (please).

> Your tying the language specification (the interpretation of format argument) to exact version of Phobos that the host compiler used to build ddmd had.

I would contend that ddmd isn't part of the language spec, but an implementation detail, but I see your point nonetheless.

> While I find std.format really useful in general, I think its feature-set / interface surface is too complex to bake into the language, even we agree on a fixed subset for the point above

This is a good point and the one which convinced me to remove usage of format and format specifiers.

> I think it would be much simpler to go with a __ctfeToString(<ScalarType> s) built-in.

See https://github.com/Calrama/druntime/commit/d968dfb84b60ba65401fcecf67f3840817efca63, https://github.com/Calrama/dmd/commit/d8d904d307c0ace7521c9ae941436cb18cf2e65e, and https://github.com/Calrama/dmd/commit/4a5c21a7e023aaca5b01916d8a619f8b9ce1b36d

> I agree that this would be quite useful, but we would need to really limit and specify the exact feature set.

My chosen route is now that the only configurability is for floating points, and only scientific (mantissa+exponent) vs default representation and number of precision digits.

> Can you please fix __ctfeWrite's signature from taking string to scope const(char)[] str ? string is quite limiting.

You can just cast to string. Nevertheless: https://github.com/Calrama/druntime/commit/970338afcddebf3b558ae4e1e8fbc3081acde105
The corresponding dmd change is not part of this PR right now, because that would then require a druntime pull that would block this one (mangling change).

**Comment by Calrama** [comment]
druntime PR for signature change of __ctfeWrite is now open: https://github.com/dlang/druntime/pull/1904
This PR now requires it.

**Comment by JinShil** [comment]
@MoritzMaxeiner Can you please rebase this?  It appears that it has reached Github's max 1000 status updates, so the status is no longer refreshing.

Also, what's the current state of this PR?  Are you satisfied with the implementation and just waiting for approval?

**Comment by MoritzMaxeiner** [comment]
@JinShil Done. W.r.t. the status: Until feedback to the contrary comes in this is just waiting for approval as far as I am concerned.

**Comment by JinShil** [comment]
@WalterBright and @andralex:  What are your opinions on this?

**Comment by andralex** [comment]
What's the difference between this and pragma msg? Is that just the newline?

**Comment by stefan-koch-sociomantic** [comment]
@andralex pragma msg can only be used on literals and literal-like expression (such as ctfe-evaluations),
whereas __ctfeWriteln can work during the ctfe-evaluation phase.
and be used for debugging.
Until we have a better approach for ctfe debugging this is the best we have.

**Comment by andralex** [comment]
Can't we expand the working of pragma msg? It seems odd to have two syntaxes for subtly different purposes. Just make pragma msg work in more cases.

**Comment by stefan-koch-sociomantic** [comment]
@andralex that is not possible, which has to do with the nature of ctfe.
Values within a ctfe evaluation are not necessarily accessible.

**Comment by andralex** [comment]
I'm talking about syntax...

**Comment by MetaLang** [comment]
As I understand it, the issue is that `pragma(msg)` is part of a different compilation phase than CTFE so you can only use it for the most rudimentary debugging, i.e., printing out the final return value of a function.

**Comment by JinShil** [comment]
Here's what my research on the subject has found:

### Forum Discussions:

- [Jul 12, 2011 - CTFE writeln](http://forum.dlang.org/post/ivh0eu$2gi5$1@digitalmars.com)
- [Aug 7, 2011 - CTFE writeln again (__ctfeWriteln)](http://forum.dlang.org/post/j1n0l7$235r$1@digitalmars.com)
- [May 13, 2012 - How do I do printf/writefln debugging in CTFE?](http://forum.dlang.org/post/jon52b$178n$1@digitalmars.com)
- [Jul 09, 2013 - access CTFE variables at compile time](http://forum.dlang.org/post/wirvivzuzjesoxcedxbj@forum.dlang.org)
- [Sep 1, 2016](http://forum.dlang.org/post/ncvamrxgunhhioqaatgu@forum.dlang.org)
- [Sep 6, 2016 - Support stated from @MartinNowak](http://forum.dlang.org/post/pxnleritpfkmvhhpdqza@forum.dlang.org)
- [Aug 14, 2017 - Probably the impetus for the latest PR](http://forum.dlang.org/post/pjqxukmmvymjxgbgbinp@forum.dlang.org)

### Bugzilla Issues
- [Issue 3952 - pragma(msg,...) has bugs + alternative idea](https://issues.dlang.org/show_bug.cgi?id=3952)

### Prior pull requests
- [Sep 2, 2016 - Closed](https://github.com/dlang/dmd/pull/6101)
- [Feb 4, 2012 - Closed](https://github.com/dlang/dmd/pull/692)
- [Aug 8, 2011 - Closed](https://github.com/dlang/dmd/pull/296)

This looks like a compelling feature, and, in my own personal opinion, @andralex and @WalterBright would likely support it if a formal request for it were made.  I think this will needs a DIP to bring @andralex  and @WalterBright  into the conversation and thoroughly justify it's need, design, and interplay/supersession of existing language features.

@WalterBright and @andralex, could you please state a firm decision on whether a DIP is required for this feature?  Thanks!

**Comment by andralex** [comment]
A good way to ease into this is to allow `std.stdio.write` and `std.stdio.writeln` work with a subset of inputs during compilation. I don't think anyone would object to that.

**Comment by MetaLang** [comment]
What if there was a special __ctfeStdout/__ctfeStderr provided by the compiler? Then we could add a special CTFE-only path and output to those.

**Comment by MoritzMaxeiner** [comment]
@andralex If the subset contains floating points this will yield surprising (i.e. inconsistent behaviour between runtime and compile time) results on x86, because the last time I checked the CTFE interpreter treats both `float` and `double` as `real` (and it's not trivial to change that). That issue (compile time formatting) is fairly complex (compared to just printing a preformatted string), which is why this PR does not attempt to address it (I do locally have a __ctfeFormat extension, but because of the floating point issue it's not really for general usage).

**Comment by andralex** [comment]
Upon further thinking, it seems we still need this primitive even if folded within std.stdio. I think we need a way to output to either stdout and stderr.

**Comment by MoritzMaxeiner** [comment]
If stderr support is a requirement, one of the following should work:
- separate builtins (e.g. `__ctfeWrite`and `__ctfeErrorWrite`) for targeting stdout / stderr
- select target stream via argument
  Since IIRC arbitrary IO isn't wanted, it should only be possible to select between predefined targets, so either a simple boolean switch, or an enumaration (e.g. `CTFEOutputStream`)

**Comment by timotheecour** [comment]
> Since IIRC arbitrary IO isn't wanted

why not? file output would definitely be useful, if security is a concern a mechanism analog to `-J` could be used to control where CTFE can write to (eg syntax: `-Jw=/foo/`)

**Comment by MoritzMaxeiner** [comment]
> why not?

If memory serves it was a design decision to not have D be able to to arbitrary IO at compile time (and if so no arbitrary targets for `__ctfeWrite` would be a direct consequence of that), though I can't seem to find the (dlang) forum post where I read that.

**Comment by timotheecour** [comment]
IIRC the concern was security, but with what I suggested above https://github.com/dlang/dmd/pull/7082#issuecomment-371694820 that concern goes away.

**Comment by MoritzMaxeiner** [comment]
> IIRC the concern was security, but with what I suggested above #7082 (comment) that concern goes away.

Whether that's the case or not, however, this PR's purpose was merely to implement the existing API for `__ctfeWrite`, not invent something new. If the API is to be redesigned so severely, I feel that should be done in a separate PR (and this one should be rejected then)

File: src/ddmd/builtin.d [hunk 3]
```diff
@@ -216,6 +216,20 @@ extern (C++) Expression eval_yl2xp1(Loc loc, FuncDeclaration fd, Expressions* ar
...
 
+extern (C++) Expression builtin_ctfeWrite(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    import core.stdc.stdio : fprintf;
+
+    Expression arg0 = (*arguments)[0];
+
+    if (StringExp se = arg0.toStringExp)
+    {
+        fprintf(global.stdmsg, "%.*s".ptr, se.len, se.string);
```
WalterBright: the se.len must be cast to int

WalterBright:
When I've needed to debug ctfe functions, I insert a printf or writefln and make them non ctfe functions, debug them, then take out the print statements.

jacob-carlborg:
> When I've needed to debug ctfe functions, I insert a printf or writefln and make them non ctfe functions, debug them, then take out the print statements.

There are other use cases. With this feature it would be possible to implement a framework that runs unit tests at CTFE.

stefan-koch-sociomantic:
what you can do is adding an additional string to your parametes and returnvalue and use that as a stream to write logs to.

wilzbach:
In the State of D survey this was among the top10 missed language features

image

https://rawgit.com/wilzbach/state-of-d/master/report.html

baryluk:
Hi. I only noticed that `__ctfeWrite` exists today by accidently looking at `object.di` in phobos distributed with LDC2.

Could we please have documentation for this at https://dlang.org/spec/function.html#interpretation maybe?

Also does it replace `pragma(msg, ...);` ? Should be mentioned in documentation what it is really for.

UplinkCoder:
There was an idea that this might be useful.
it turns out that it's not that useful.
Therefore `__ctfeWrite` doesn't do anything.

MoritzMaxeiner:
> it turns out that it's not that useful.

What has changed? I haven't used D in a while, but when I did so last, this was useful to me.

RazvanN7:
@MoritzMaxeiner It seems that @MoonlightSentinel has taken this over. Is it ok to close this PR?

MoritzMaxeiner:
@RazvanN7 Sure. I don't personally need this anymore.

PR #7083, opened 2017-08-16
bump VERSION to v2.076.0-b1

MartinNowak:


MetaLang:
@MartinNowak Is that merge commit supposed to be there?

PR #7084, opened 2017-08-16
refactor: add parseCommandLine()

WalterBright:
This just pulls the code that loops over `arguments[]` into `parseCommandLine()`. You can't really tell that from the github diff, as github gets thoroughly confused.

This makes the inputs and outputs for parsing the command line clear, and removes the global variable references from it. All instances of `global.params` in `parseCommandLine()` are replaced with just `params`. It's much better encapsulated.

More can be done, but this is a good first approximation.

### DIFF
--- a/src/ddmd/mars.d
+++ b/src/ddmd/mars.d
@@ -370,702 +370,8 @@ private int tryMain(size_t argc, const(char)** argv)
     getenv_setargv(readFromEnv(&environment, "DFLAGS"), &arguments);
     updateRealEnvironment(&environment);
     environment.reset(1); // don't need environment cache any more
-    version (none)
-    {
-        for (size_t i = 0; i < arguments.dim; i++)
-        {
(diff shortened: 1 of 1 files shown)


JinShil:
Since you're going through the work of refactoring this entire file, is it possible to start using D's string facilities instead of C's `char *` , `strcmp`, and friends?

WalterBright:
Yes, but my main concern is better encapsulation, rather than a tangle of dependencies.

JinShil:
If you main concern is a dependency on Phobos, I understand.  But what about just druntime where `string` is defined.  You can always add a string utilities module to DMD to avoid the dependency on Phobos.

WalterBright:
By encapsulation I mean having clear inputs and outputs. Dependency on Phobos is a separate issue.

MartinNowak (approved):
LGTM

PR #7088, opened 2017-08-20
refactor: better encapsulate parseCommandLine()

WalterBright:
This moves towards making `parseCommandLine()` into a more pure, encapsulated function by:

1. pushing decisions about exiting the compiler up to the caller

2. pushing some requested actions up to the caller

3. detecting the existence of errors with return value rather than a global

Error messages are consolidated into a single nested function.

Error messages about command line are followed by a supplemental message about how to open a browser on the manual.

There are remaining problems with setting `debuglevel` and `versionlevel` via globals that should be addressed separately.

### DIFF
--- a/src/ddmd/mars.d
+++ b/src/ddmd/mars.d
@@ -371,7 +371,76 @@ private int tryMain(size_t argc, const(char)** argv)
     updateRealEnvironment(&environment);
     environment.reset(1); // don't need environment cache any more
 
-    parseCommandLine(arguments, argc, global.params, files);
+    if (parseCommandLine(arguments, argc, global.params, files))
+    {
+        Loc loc;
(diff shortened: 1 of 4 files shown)


src/ddmd/mars.d
```diff
@@ -371,7 +371,76 @@ private int tryMain(size_t argc, const(char)** argv)
...
+  =tls           list all variables going into thread local storage
+");
+        return EXIT_SUCCESS;
+    }
+
+    if (global.params.manual)
+    {
+        version (Windows)
+        {
+            browse("http://dlang.org/dmd-windows.html");
```
wilzbach: FYI: https://dlang.org/dmd.html redirects to the corresponding OS automatically, but there is no harm in being explicit here

PetarKirov (approved):
Looks good.

wilzbach (approved):

PR #7093, opened 2017-08-21
fix Issue 17761 - don't leave behind empty section groups

MartinNowak:
- check for existing .text.<mangling> section before opening new
  group section, reuse existing sections in case of colliding mangling

Until now it wasn't easily possible to lookup an existing section, hence the refactoring was necessary for the fix. At best review commits separately. I've also added more detailed documentation to those functions.

### DIFF
--- a/src/ddmd/backend/elfobj.c
+++ b/src/ddmd/backend/elfobj.c
@@ -339,8 +339,9 @@ int seg_max;
 int seg_tlsseg = UNKNOWN;
 int seg_tlsseg_bss = UNKNOWN;
 
-int elf_getsegment2(IDXSEC shtidx, IDXSYM symidx, IDXSEC relidx);
-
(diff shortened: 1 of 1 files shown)


src/ddmd/backend/elfobj.c
```diff
@@ -609,12 +610,18 @@ static IDXSEC elf_newsection2(
...
+Add a new section name or get the string table index of an existing entry.
 
+Params:
+    name = name of section
+    suffix = append to name
+    padded = set to true when entry was newly added
+Returns:
+    String index of new or existing section name.
+ */
+static IDXSEC elf_addsectionname(const char *name, const char *suffix = NULL, bool *padded = NULL)
```
WalterBright: Should return `IDXSTR`

WalterBright (changes requested):
Replace IDXSEC on line 623 with IDXSTR.

MartinNowak:
Done @WalterBright

WalterBright (approved):

PR #7094, opened 2017-08-22
Remove issue stats badge from README

JinShil:
The issue stats badge doesn't work.  The data feeding this badge hasn't been updated in, what appears to be, almost a year, and it appears the repository has been abandoned.  It's conveying incorrect information, so I think it should be removed.


wilzbach (approved):

wilzbach:
I even tried to contact the owner a couple of months ago and sadly without response.

PR #7095, opened 2017-08-22
Fix Issue 11259:  __traits(isSame) fails on the result of __traits(parent) if parent is a package

JinShil:


test/compilable/test11259.d
```diff
@@ -0,0 +1,6 @@
+// https://issues.dlang.org/show_bug.cgi?id=11259
+
+import imports.imp11259;
+
+static assert(__traits(isSame, __traits(parent, imports.imp11259), imports));
+static assert(__traits(isSame, imports, __traits(parent, imports.imp11259)));
```
ZombineDev: Can you add a test for nested packages, e.g:
```D
version (Posix)
{
    // smallest druntime module without imports on posix
    import core.sys.posix.libgen;
    static assert(__traits(isSame, __traits(parent, core.sys.posix.libgen), core.sys.posix));
    static assert(__traits(isSame, __traits(parent, core.sys.posix), core.sys));    
}
else
{
    // smallest module without imports for windows
    import core.sys.windows.lmuseflg;
    static assert(__traits(isSame, __traits(parent, core.sys.windows.lmuseflg), core.sys.windows));
    static assert(__traits(isSame, __traits(parent, core.sys.windows), core.sys));
}

static assert(__traits(isSame, __traits(parent, core.sys), core));
```

(Using phobos in the test-suite is discouraged, but I think druntime is ok.)

src/ddmd/traits.d
```diff
@@ -1353,6 +1353,22 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        for(int i = 0; i < 2; i++)
+        {
+            // if i == 0, compare s1 (import) to s2 (package)
+            // else compare s1(package) to s2 (import)
+            if (auto imp = i == 0 ? s1.isImport() : s2.isImport())
+            {
+                if (auto pkg = i == 0 ? s2.isPackage() : s1.isPackage())
+                {
+                    if (imp.pkg == pkg)
+                        return True();
```
WalterBright: This would seem to be much simpler to implement as:
```
if (s1.isImport().pkg == s2.isPackage() ||
    s2.isImport().pkg == s1.isPackage)
    return True();

Comment (JinShil) [inline src/ddmd/traits.d]: But that doesn't check that `isImport()` and `isPackage()` actually return something non-`null`.

Comment (WalterBright) [inline src/ddmd/traits.d]: Right you are. My mistake. How about this:
```
static bool cmp(Dsymbol s1, Dsymbol s2)
{
    auto imp = s1.isImport();
    return imp && imp.pkg && imp.pkg == s2.isPackage();
}

if (cmp(s1,s2) || cmp(s2,s1))
    return True();
```
I've used the loop method in C/C++ when nested functions weren't available. I think nested functions are nicer.

JinShil: > I think nested functions are nicer.

Indeed.  Updated and ready to go.

PR #7097, opened 2017-08-24
Fix Issue 12496: __traits(parent, x) returns incorrect symbol

JinShil:
~This is the 2nd of 3 pull requests to fix issue 12496.  They must be done in order to ensure the test suite is green with each step.~

~Step 1: https://github.com/dlang/phobos/pull/5709 - Deletes an incorrect Phobos unittest that is preventing Step 2 (this PR) from passing its test suite.~
~Step 2: (This PR) the actual fix for issue 11246.  This should pass its test suite after Step 1 is pulled.~
~Step 3: https://github.com/dlang/phobos/pull/5704 - Repairs the unittest that was deleted in Step 1 so it properly tests for correct behavior.~

This pull request is on its own now; it does not depend on any other pull requests.

### DIFF
--- a/src/ddmd/traits.d
+++ b/src/ddmd/traits.d
@@ -637,6 +637,25 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         auto s = getDsymbol(o);
         if (s)
         {
+            // https://issues.dlang.org/show_bug.cgi?id=12496
+            // Consider:
+            // class T1
+            // {
+            //     class C(uint value) { }
+            // }
+            // __traits(parent, T1.C!2)
+            if (auto ad = s.isAggregateDeclaration())  // `s` is `C`
+            {
+                if (ad.isNested())                     // `C` is nested
+                {
+                    if (auto p = s.toParent())         // `C`'s parent is `C!2`, believe it or not
+                    {
+                        if (p.isTemplateInstance())    // `C!2` is a template instance
+                            s = p;                     // `C!2`'s parent is `T1`
+                    }
+                }
+            }
+
             if (auto fd = s.isFuncDeclaration()) // https://issues.dlang.org/show_bug.cgi?id=8943
                 s = fd.toAliasFunc();
             if (!s.isImport()) // https://issues.dlang.org/show_bug.cgi?id=8922
--- a/test/compilable/test12496.d
+++ b/test/compilable/test12496.d
@@ -1,0 +1,13 @@
+// https://issues.dlang.org/show_bug.cgi?id=12496
+
+final abstract class T1
+{
+    final abstract class C(uint value) { }
+
+    alias Child = C!2;
+}
+
+void main()
+{
+    static assert(__traits(isSame, __traits(parent, T1.Child), T1));
+}


JinShil:
> What happens if `C` in your test case is also instantiated from a different context – does the "parent" then depend on compiler internals (semantic analysis order)?

I expect the "parent" to always be relative to the site of template instantiation.  But alas, the following test fails on the second assertion.

```D
final abstract class T1 
{
    final abstract class C(uint value)
    {
        static uint getValue() 
        {
            return value;
        }
    }

    alias Child = C!2;
}

final abstract class T2
{
    alias Child = T1.C!2;
}

void main()
{
    static assert(__traits(isSame, __traits(parent, T1.Child), T1));
    static assert(__traits(isSame, __traits(parent, T2.Child), T2));  // Error: static assert  __traits(isSame, T1, T2) is false
}
```

This needs more work.

ZombineDev:
> I expect the "parent" to always be relative to the site of template instantiation

That would be correct only if were talking about *mixin* templates. According to the spec [[1]](https://dlang.org/spec/template.html#instantiation_scope) [[2]](https://dlang.org/spec/template-mixin.html):

> TemplateInstances are always instantiated in the scope of where the TemplateDeclaration is declared [..]

> Unlike a template instantiation, a template mixin's body is evaluated within the scope where the mixin appears, not where the template declaration is defined [..]

So in your last example above, I would expect the following:
```D
static assert(__traits(isSame, __traits(parent, T1.Child), T1));
static assert(__traits(isSame, __traits(parent, T2.Child), T1));
```

However, if `C` is defined as `mixin template`, the the result would be:
```D
static assert(__traits(isSame, __traits(parent, T1.Child), T1));
static assert(__traits(isSame, __traits(parent, T2.Child), T2));
```

JinShil:
Ok, so my understanding of how this should work was wrong.  However, as far as I can tell, the implementation in this pull request works as it should according to the spec.  That is, the following compiles without any assertion failures.

```D
final abstract class T1 
{
    final abstract class C(uint value)
    {
        static uint getValue() 
        {
            return value;
        }
    }

    alias Child = C!2;
}

final abstract class T2
{
    alias Child = T1.C!2;
}

void main()
{
    static assert(__traits(isSame, __traits(parent, T1.Child), T1));
    static assert(__traits(isSame, __traits(parent, T2.Child), T1));
}
```

It does, however, cause the test in Step 1 to fail, and the test in Step 3 to read `std.traits.TemplatedStruct.foo` instead of `std.traits.TemplatedStruct!().foo`.

What, specifically, is not right with the implementation in this pull request?

JinShil:
This is ready for additional scrutiny and review.

I'm not sure, however, if it would be better to integrate the logic in this PR into https://github.com/dlang/dmd/blob/583df4c93de7ef54cc63c2ef84815378180135b0/src/ddmd/dsymbol.d#L416-L424

JinShil:
@ZombineDev & @klickverbot Could you please give this another look?  Thanks!

src/ddmd/traits.d
```diff
@@ -637,6 +637,25 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+            // }
+            // __traits(parent, T1.C!2)
+            if (auto ad = s.isAggregateDeclaration())  // `s` is `C`
+            {
+                if (ad.isNested())                     // `C` is nested
+                {
+                    if (auto p = s.toParent())         // `C`'s parent is `C!2`, believe it or not
+                    {
+                        if (p.isTemplateInstance())    // `C!2` is a template instance
+                            s = p;                     // `C!2`'s parent is `T1`
```
ZombineDev: IIUC, for the test case you'll end up doing `resolve(e.loc, sc, s, false)`, correct? Why not simply return that here, instead of relying on fall through?

PetarKirov (commented):
Thanks for the ping, I have one question:

ZombineDev:
> I'm not sure, however, if it would be better to integrate the logic in this PR into the toParent or pastMixin function

I would prefer to move the logic there (i.e. make those functions more generic), but I'm not sure if this is doable, since changing how those functions work might break something. The only way to know is to try and see what happens with.

andralex (approved):

JinShil:
> IIUC, for the test case you'll end up doing resolve(e.loc, sc, s, false), correct? Why not simply return that here, instead of relying on fall through?

Yes, however, the other code falls through, and I'm just going with the flow.  Also, the duplication of `resolve(e.loc, sc, s, false)` would make me feel a little guilty.

> I would prefer to move the logic there (i.e. make those functions more generic), but I'm not sure if this is doable, since changing how those functions work might break something. The only way to know is to try and see what happens.

I tried it last night, but it turned out to be non-trivial (at least for me).  Also, since @andralex has approved this as-is, it would be disingenuous to refactor it now.  I have it on my todo list to revisit it.  

Thank you for the review!

PR #7101, opened 2017-08-25
fix Issue 17370 - [scope] Escaping scope pointers possible via struct…

WalterBright:
… GC allocation

### DIFF
--- a/test/fail_compilation/retscope2.d
+++ b/test/fail_compilation/retscope2.d
@@ -291,5 +291,28 @@ struct T17388
     return t.foo();
 }
 
+/****************************************************/
 
+/*
+TEST_OUTPUT:
+---
+fail_compilation/retscope2.d(1306): Error: returning `& i` escapes a reference to local variable `i`
+---
+*/
+
+#line 1300
+
+// https://issues.dlang.org/show_bug.cgi?id=17370
+
+void test1300() @safe
+{
+    int i;
+    auto p = new S1300(&i).oops;
+}
+
+struct S1300
+{
+    int* oops;
+//    this(int* p) @safe { oops = p; }
+}
 
--- a/src/ddmd/escape.d
+++ b/src/ddmd/escape.d
@@ -37,6 +37,7 @@ import ddmd.arraytypes;
  * Print error messages when these are detected.
  * Params:
  *      sc = used to determine current function and module
+ *      fdc = function being called, `null` if called indirectly
  *      par = identifier of function parameter
  *      arg = initializer for param
  *      gag = do not print error messages
@@ -211,6 +212,8 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
         if (ie.e1.op == TOKvar && ie.e1.type.toBasetype().ty == Tsarray)
             va = (cast(VarExp)ie.e1).var.isVarDeclaration();
     }
+
+    if (log && va) printf("va: %s\n", va.toChars());
 
     // Try to infer 'scope' for va if in a function not marked @system
     bool inferScope = false;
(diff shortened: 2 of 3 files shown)


WalterBright:
Blocked by https://github.com/dlang/druntime/pull/1909

WalterBright:
The continuous-integration is failing for the same error that was fixed in dlang/druntime#1909

test/fail_compilation/retscope2.d
```diff
@@ -291,5 +291,28 @@ struct T17388
     return t.foo();
 }
 
+/****************************************************/
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/retscope2.d(1306): Error: returning `& i` escapes a reference to local variable `i`
```
ZombineDev: I would strongly prefer if the error message didn't have `return` in it when there are no `return` statements to be found.
For reference, if you uncomment the constructor, it looks like this:
```
Error: reference to local variable `i` assigned to non-scope parameter `p` calling `scope_bug17370.Struct.this`
```

Ideally the error message should look like this:
```
Error: reference to local variable `i` assigned to heap-allocated `S1300.oops` with possibly longer lifetime
```

PetarKirov (changes requested):

ZombineDev:
> The continuous-integration is failing for the same error that was fixed in dlang/druntime#1909

@WalterBright you need to manually restart the build at Travis CI by clicking on the [Details](https://travis-ci.org/dlang/dmd/builds/268528443?utm_source=github_status&utm_medium=notification) button, logging in with your GitHub profile (if you aren't already logged in there) and clicking the "Restart build" button that should be on the right of your screen. 
Unfortunately, Travis CI doesn't automatically restart after changes to druntime or phobos are merged, like the other CI do. (I can restart the other CIs, but I don't have rights to do so for the dmd builds on Travis CI).

test/fail_compilation/retscope2.d
WalterBright: While I agree with you, improving the message is a minor issue relative to fixing this bug. It requires some refactoring of the code to make it work. I suggest adding it as an issue on bugzilla so it can be addressed on its own.

WalterBright:
@ZombineDev thanks that worked

test/fail_compilation/retscope2.d
ZombineDev: Ok, works for me: https://issues.dlang.org/show_bug.cgi?id=17784.

test/fail_compilation/retscope2.d
```diff
@@ -291,5 +291,28 @@ struct T17388
...
+*/
 
+#line 1300
+
+// https://issues.dlang.org/show_bug.cgi?id=17370
+
+void test1300() @safe
+{
+    int i;
+    auto p = new S1300(&i).oops;
```
mathias-lang-sociomantic: This test case is quite dependent on the compiler not being able to see that this construct is in fact safe, because the reference to the newed pointer is never reachable anymore.
I would suggest a `return` instead of `auto` assignment.

mathias-lang-sociomantic (approved):
Minor comment regarding the test, otherwise LGTM

test/fail_compilation/retscope2.d
WalterBright: The compiler isn't likely to do forward flow analysis anytime soon. If it does, and decides that the assignment is not an error, then this test will fail because the error is expected. So I'm not concerned about it. It's not going to product silent breakage.

PetarKirov (approved):

test/fail_compilation/retscope2.d
mathias-lang-sociomantic: Fair enough :+1:

MartinNowak: > The compiler isn't likely to do forward flow analysis anytime soon.

It seems that we'll need data flow analysis for RC, so I'm planning to work on a DIP & POC implementation in late fall, i.e. fairly soon ;).

PR #7104, opened 2017-08-27
Fix Issue 17684 - [REG 2.062] static alias this (Part 3)

JinShil:
Followup to https://github.com/dlang/dmd/pull/7055 and https://github.com/dlang/dmd/pull/7070

The current implementation does not work for the following:
- `&&` and `||` expressions
- `!` expressions
- `return` statements

This PR fixes those.

Tests were refactored to reduce redundancy and test for the features above.


JinShil:
@MartinNowak (Since you reviewed the last two PRs)

This is ready to go.

ZombineDev:
The other PRs were targeted at the stable branch. Should this too? (Otherwise the changes in this PR will be released with 2.077.0, while the rest (part 1 and part 2) will be out with 2.076.0).

BTW Mike, considering how much effort you put into this feature, we may as well advertise it as new one, by adding a changelog entry ;)

JinShil:
Well, I tried to retarget stable through GitHub's interface, but it looks really bad.  I'll opening a new PR.

ZombineDev:
@JinShil after you change the target branch from GitHub's UI, you also need to rebase your local branch on top of dmd stable and then force push your local branch to GitHub.

JinShil:
Sorry @ZombineDev, I tried but Github wouldn't let me re-open because I force-pushed.  My git-foo is still pretty poor.

About the changelog: I never really intended this to be a new feature.  I though it was just supposed to work that way, and if you look at the original issue, I just wanted to make the assignment symmetric.  I just happened to stumble on these other  issues as I started using the feature.  

The issue will show up in the changelog anyway.   But, I'll see about adding a special entry in the changelog if that's what everyone wants.

JinShil:
Re-opened at https://github.com/dlang/dmd/pull/7106

PR #7105, opened 2017-08-27
fix issue 17782 - The identifier delimiter of a delimited string can not begin with '_'

nmtigor:
https://issues.dlang.org/show_bug.cgi?id=17782


JinShil:
I recommend adding a test at `test/compilable/test17782.d`.

PR #7106, opened 2017-08-28
Fix Issue 17684 - [REG 2.062] static alias this (Part 3)

JinShil:
Followup to https://github.com/dlang/dmd/pull/7055 and https://github.com/dlang/dmd/pull/7070

The current implementation does not work for the following:
- `&&` and `||` expressions
- `!` expressions
- `return` statements

This PR fixes those.

Tests were refactored to reduce redundancy and test for the features above.


ibuclaw:
Seems reasonable to me.

PR #7107, opened 2017-08-29

MartinNowak:


ZombineDev:
Are you aware of https://github.com/dlang/dmd/pull/7106?

I think we need to include it in the release, in order to have a well-working `static alias` support for the release. Though I guess delaying this to the point release is also an option, if you want to proceed with the release.

PR #7108, opened 2017-08-30
fix Issue 17790 - [scope] Escaping pointer possible through array of …

WalterBright:
…aggregates

### DIFF
--- a/src/ddmd/escape.d
+++ b/src/ddmd/escape.d
@@ -37,6 +37,7 @@ import ddmd.arraytypes;
  * Print error messages when these are detected.
  * Params:
  *      sc = used to determine current function and module
+ *      fdc = function being called, `null` if called indirectly
  *      par = identifier of function parameter
  *      arg = initializer for param
  *      gag = do not print error messages
@@ -212,6 +213,8 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
             va = (cast(VarExp)ie.e1).var.isVarDeclaration();
     }
 
+    if (log && va) printf("va: %s\n", va.toChars());
+
     // Try to infer 'scope' for va if in a function not marked @system
     bool inferScope = false;
     if (va && sc.func && sc.func.type && sc.func.type.ty == Tfunction)
@@ -245,10 +248,10 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
 
             // If va's lifetime encloses v's, then error
             if (va &&
-                (va.enclosesLifetimeOf(v) && !(v.storage_class & STCparameter) ||
+                (va.enclosesLifetimeOf(v) && !(v.storage_class & (STCparameter | STCtemp)) ||
                  // va is class reference
                  ae.e1.op == TOKdotvar && va.type.toBasetype().ty == Tclass && (va.enclosesLifetimeOf(v) || !va.isScope) ||
-                 va.storage_class & STCref) &&
+                 va.storage_class & STCref && !(v.storage_class & STCtemp)) &&
                 sc.func.setUnsafe())
             {
                 if (!gag)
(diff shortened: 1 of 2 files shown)


PetarKirov (approved):
LGTM

ZombineDev:
> std/datetime/date.d(9737): Error: scope variable `__r6551` assigned to `str` with longer lifetime
std/datetime/date.d(9998):        called from here: validTimeUnits(["days"][])
Looks like we need fix phobos first.

WalterBright:
No, it's a compiler problem. Working on it.

src/ddmd/escape.d
```diff
@@ -37,6 +37,7 @@ import ddmd.arraytypes;
  * Print error messages when these are detected.
  * Params:
  *      sc = used to determine current function and module
+ *      fdc = function being called, `null` if called indirectly
```
mathias-lang-sociomantic: This kind of unrelated fixes, while welcome, should go in their own commit. If there is a bug showing up with the fix, and we end up reverting it, we'll loose this fix. It also make it easier to spot for reviewer, and is trivial to do using `git`.

src/ddmd/escape.d
```diff
@@ -665,7 +669,7 @@ private bool checkReturnEscapeImpl(Scope* sc, Expression e, bool refs, bool gag)
 
     foreach (Expression ee; er.byexp)
     {
-        //printf("byexp %s\n", ee.toChars());
+        if (log) printf("byexp %s\n", ee.toChars());
```
mathias-lang-sociomantic: Same here. More than half of the diff in this file is logging / debugging change. In my opinion it shouldn't be there to begin with, but if you want it here, it should at least be in its own commit.

mathias-lang-sociomantic (changes requested):
You know what they say:
> All problems in computer science can be solved by another level of indirection.

So I checked out your branch and gave it a shot (with `-dip1000 -run test.d`
```D
void main () @safe
{
    int* ptr = bar();
    assert(ptr !is null);
}

int* bar () @safe
{
    int i;
    int*[][] arr = new int*[][](1);
    arr[0] ~= &i;
    return arr[0][0];
}
```

Sadly it did compile.

src/ddmd/escape.d
WalterBright: I understand your point, but at some level this just gets to be more pedantic than practical. It's just a comment.

src/ddmd/escape.d
WalterBright: It's trivial, and does not obfuscate this PR. I often leave in debugging aids when I develop PRs, they are not completely unrelated.

WalterBright:
> Sadly it did compile.

Glad you found it. But it should be a separate bugzilla issue, as it is unrelated to this one.

src/ddmd/escape.d
WalterBright: Part of the issue (for me) is the looooong delays in getting even the most trivial PRs pulled. These debug aids are very helpful to me.

src/ddmd/escape.d
mathias-lang-sociomantic: > but at some level this just gets to be more pedantic than practical. 

I think we are far away from this level. Proper commit messages, and proper commit separation help *a lot* on the long run. That's what we have here in Sociomantic, and while it sometimes gets annoying, I can't understate how useful it is when you have to track down information which is months or years old.

I don't think it should be a blocker for your PRs, but I just wanted to raise the point here.

src/ddmd/escape.d
mathias-lang-sociomantic: > Part of the issue (for me) is the looooong delays in getting even the most trivial PRs pulled. 

I would happily merge the `scope` related P.R. if I had access right. This is an area where I can pretend to know what I'm doing ;)

mathias-lang-sociomantic:
> But it should be a separate bugzilla issue, as it is unrelated to this one.

Sounded very similar to me, but I'm fine if you want it addressed in another P.R...

mathias-lang-sociomantic (approved):

src/ddmd/escape.d
WalterBright: I added pull privileges to you, but keep in mind that Leandro and Dicebot both have pull privileges, too.

mathias-lang-sociomantic: > I added pull privileges to you, but keep in mind that Leandro and Dicebot both have pull privileges, too.

I know they have pull access, but AFAIK they only care about sociomantic-related stuff nowadays.
`scope` is of no interest for us ATM (we're using 2.070 / 2.071), so the time used to work on it is part of our [open source initiative](https://github.com/sociomantic-tsunami).

I am still not seeing the option to pull tho :confused:

mathias-lang-sociomantic:
Travis failure is only on one job of the matrix, with :
> The job exceeded the maximum time limit for jobs, and has been terminated.

Everything else is green, so merging

PR #7109, opened 2017-08-30
[Trivial] More precise error message for TypeInfo mismatch

JinShil:
Current error message emitted by the compiler reads:  "mismatch between compiler and object.d or object.di found. Check installation and import paths with -v compiler switch."

This pull request changes it to (for example):  "TypeInfo_Struct: mismatch between compiler (136 bytes) and object.d or object.di (16 bytes) found. Check installation and import paths with -v compiler switch."


adamdruppe:
i like it

PetarKirov (approved):
Looks good to me too.

mathias-lang-sociomantic (approved):

wilzbach (approved):

PR #7110, opened 2017-08-31
fix Issue 12625 - [scope] [DIP1000] implicit slicing of RValue static…

WalterBright:
… array should be illegal

(I made it deprecated so as not to break code.)

### DIFF
--- a/test/runnable/test42.d
+++ b/test/runnable/test42.d
@@ -5824,10 +5824,15 @@ void test4414() {
     assert(x == 7);
   }
   {
-    auto x = bytes4414()[0..4];
+    auto u = bytes4414();
+    auto x = u[0..4];
     if (x[0] != 7 || x[1] != 8 || x[2] != 9 || x[3] != 10)
         assert(0);
   }
+  assert(bytes4414()[0] == 7);
+  assert(bytes4414()[1] == 8);
+  assert(bytes4414()[2] == 9);
+  assert(bytes4414()[3] == 10);
 }
 
 /***************************************************/
--- a/src/ddmd/escape.d
+++ b/src/ddmd/escape.d
@@ -413,6 +413,19 @@ ByRef:
     foreach (Expression ee; er.byexp)
     {
         if (log) printf("byexp: %s\n", ee.toChars());
+
+        /* Do not allow slicing of a static array returned by a function
+         */
+        if (va && ee.op == TOKcall && ee.type.toBasetype().ty == Tsarray && va.type.toBasetype().ty == Tarray &&
+            !(va.storage_class & STCtemp))
+        {
+            if (!gag)
+                deprecation(ee.loc, "slice of static array temporary returned by `%s` assigned to longer lived variable `%s`",
+                    ee.toChars(), va.toChars());
+            //result = true;
+            continue;
+        }
+
         if (va && !va.isDataseg() && !va.doNotInferScope)
         {
             if (!va.isScope() && inferScope)
@@ -421,6 +434,7 @@ ByRef:
             }
             continue;
         }
+
         if (sc.func.setUnsafe())
         {
             if (!gag)
--- a/test/fail_compilation/retscope4.d
+++ b/test/fail_compilation/retscope4.d
@@ -1,0 +1,22 @@
+/*
+REQUIRED_ARGS: -de
+PERMUTE_ARGS:
+*/
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/retscope4.d(3007): Deprecation: slice of static array temporary returned by `func()` assigned to longer lived variable `a`
+---
+*/
+
+#line 3000
+
+// https://issues.dlang.org/show_bug.cgi?id=12625
+
+int[16] func();
+
+void foo()
+{
+    int[] a = func();
+}


test/runnable/test42.d
```diff
@@ -5824,7 +5824,8 @@ void test4414() {
     assert(x == 7);
   }
   {
-    auto x = bytes4414()[0..4];
+    auto u = bytes4414();
+    auto x = u[0..4];
```
mathias-lang-sociomantic: Shouldn't `ubyte[4]` be infered here, instead of `ubyte[]` ?

schveiguy: Not sure if you get the point of the update? The original code slices a temporary which is immediately deallocated. The new code stores the temporary so it can live alongside the slice.

Or maybe I misunderstand your point?

mathias-lang-sociomantic: Oh I get why the bug fix is needed. However, looking at this change, it seems that the compiler infers `ubyte[]` for the type of (previously) `x`, while the return of `bytes4414` is `ubyte[4]`.

My question is, should that happen ? I think not, and the fact it happens will make this deprecation more visible. In here, instead of the diff, I would have turned `auto` into `ubyte[4]`, but I don't think we should have to do that in the first place.

I get that this question is a separate concern from the bug being fixed, but as shown in this diff, it will trigger false positive in this case.

schveiguy: But the return value was sliced. it should be typed as `ubyte[]`, no?

mathias-lang-sociomantic: Oh damn. Yeah you're right, and I got confused.

However, looking at the [original issue](https://issues.dlang.org/show_bug.cgi?id=4414) all the reports were slicing directly the return of the function. Walter seems to have picked up the last test case [when he fixed it](https://issues.dlang.org/show_bug.cgi?id=4414). As the fix seems to apply directly to taking an item from a static array returned by a function, this change doesn't test that anymore.

So maybe we could just test for one element ? `auto x = bytes4414()[2];` for example ?

schveiguy: Yes, I see that now! I didn't see the entire context, @WalterBright?

WalterBright: I don't understand what you're asking.

ZombineDev:
In the Phobos [pull request](https://github.com/dlang/phobos/pull/5714) I proposed of making this code:

```D
@safe:

char[16] helper()
{
    char[16] data;
    return data;
}

string s = helper();
```

correct, like so:

```D
@safe:

char[16] helper();

string foo()
{
    // step 1: value types without indirections implicitly convert to `const` or `immutable`
    // already implemented!
    immutable(char)[16] data = helper();

    // step 2: slicing local static array variables is allowed and the result is inferred as `scope`.
    // already implemented!
    string s = data;

    // Error: scope variable s may not be returned
    // just as expected!
    return s;
}
```

As @jpf91 summarized:
* Keeps old, correct code working (code which already keeps the slice variable scope-local will continue to work)
* Disallows incorrect code (the slice variable is marked scope, so it can't escape)
* Is consistent with existing language features, for example:
```D
@safe:

struct S { char[16] data; }
S get();

string foo()
{
    // Ok, currently allowed
    immutable S s = get();

    // Ok, currently allowed
    string str = s.data[];

    // scope_test4.d(14): Error: scope variable str may not be returned
    return str;
}
```

What do you guys think?

MartinNowak:
That would require to assign the returned static array to a temporary variable **iff** it's referenced/sliced, the rest should already be in place.
`string s = func()` -> `string s = (auto tmp = func(); tmp)`

test/runnable/test42.d
mathias-lang-sociomantic:
```diff
-    auto x = bytes4414()[0..4];
+    auto x0 = bytes4414()[0];
+    auto x1 = bytes4414()[1];
+    auto x2 = bytes4414()[2];
+    auto x3 = bytes4414()[3];
```

And then test the values.

d-random-contributor:
```d
int[][2] f();
int[] a=f()[0];
int[2][2] g();
int[] b=f()[0];
```
will this work?

test/runnable/test42.d
schveiguy: > I don't understand what you're asking.

The original behavior that the `test4414` function was trying to ensure was that you could index a temporary static array, which is safe and valid. By your edits, you have fixed obviously bad behavior, but you have changed what the test is doing to something that doesn't ensure that. Arguably, it should have been written to index a single element to begin with.

schveiguy:
> will this work?

I'm assuming you meant to call `g()` in your second test?

If you are asking in the context of what this PR is changing, no, it will fail, because you are still slicing a temporary, and therefore referring to invalid memory immediately.

If you are asking in response to @ZombineDev's proposal, yes it would work. What would happen is that the resulting static array of static arrays would be indexed, returning a single static array. That static array is sliced, so the lowering would occur for that one element.

In other words, it should lower to this:
```D
int[][2] f();
auto _tmp1 = f()[0];
int[] a = _tmp1[];
int[2][2] g();
auto _tmp2 = g()[0];
int[] b = _tmp2[];
```

schveiguy:
> That would require to assign the returned static array to a temporary variable iff it's referenced/sliced

Yes, definitely. It should always be done in the same statement, so not any extra context is needed I would think.

test/runnable/test42.d
WalterBright: ok

WalterBright:
> What do you guys think?

The problem is the value returned by the function call is defined to only exist until the end of the expression. Assigning it to a compiler generated temporary with a longer lifetime changes those semantics.

This matters for things like returned values with destructor or scope semantics. Now, I could make a special case in the semantics that if the returned value does not have destructors or scoped pointers, then allow it, but then there will be the inevitable bug report "I added a destructor and it broke", and "I returned an array of pointers and it broke".

schveiguy:
Quite right, Walter. Also, the same logic could be applied to anything, not just static arrays. So you could argue that the same behavior should apply to:

```D
int foo();
auto ptr = &(foo()); // creates a temporary
```

Which I don't think we want to get into. The whole implicit cast to a slice is really the sticky point, and I think the fix addresses it without too much breakage (of course, it's going to break some things, but those are probably not sound to begin with).

test/runnable/test42.d
andralex: @WalterBright course of action?

MartinNowak:
> Quite right, Walter. Also, the same logic could be applied to anything, not just static arrays. So you could argue that the same behavior should apply to:
> 
> int foo();
> auto ptr = &(foo()); // creates a temporary

That's not a useful comparison as it's explicitly taking an address rather than implicitly converting the return value. The proposal was only for implicit conversions.
Implicit conversions (e.g. from `int[2]` to `int[]`) remain valid everywhere else.

> Which I don't think we want to get into. The whole implicit cast to a slice is really the sticky point, and I think the fix addresses it without too much breakage (of course, it's going to break some things, but those are probably not sound to begin with).

The problem is that compiler has been way to lax with this memory corruption for a very long time, and that we even have bad APIs encouraging this bug.

> The problem is the value returned by the function call is defined to only exist until the end of the expression. Assigning it to a compiler generated temporary with a longer lifetime changes those semantics.

Obviously expected when you prolong the lifetime of a return value.

> This matters for things like returned values with destructor or scope semantics. Now, I could make a special case in the semantics that if the returned value does not have destructors or scoped pointers, then allow it, but then there will be the inevitable bug report "I added a destructor and it broke", and "I returned an array of pointers and it broke".

It's unlikely that anyone used the implicit conversion for values with destructors, as the early destruction should have been found during testing.
Not sure what role scoped pointers in the return value play? As @ZombineDev proposed, the slice would be scoped itself, it's referring to a local, and cannot be escaped anyhow.

Not judging the idea, but returning a temporary string or binary buffer to avoid heap allocations is a common use-case. Prolonging the lifetime for implicitly converted return values would also allow to make [tempCString](https://github.com/dlang/phobos/blob/8e47bfc54c106b222835db3c3a37e81b52ab2f04/std/internal/cstring.d#L89) `@safe`.

The breakage is likely not too big, `useString(getTemp()[])` remains valid. So I'm in favor of merging this and consider the lifetime idea as a further improvement.

nordlow:
If and when this gets merged, isn't time to finally add

```D
T[n] s(T, size_t n)(auto ref T[n] array) pure nothrow @nogc @safe
{
    return array;
}
```

to either druntime or perhaps std.array?

andralex (approved):
per discussion with @MartinNowak, please fix phobos deprecations

andralex:
ping @WalterBright

JinShil:
> per discussion with @MartinNowak, please fix phobos deprecations

@andralex What phobos deprecations?  Anything I can do?

@WalterBright Why is this marked blocked?  What is blocking it?

schveiguy:
@JinShil looks like this is the problem: https://github.com/dlang/phobos/blob/master/std/digest/sha.d#L1193

I think @WalterBright already fixed something similar, see here: https://github.com/dlang/phobos/pull/5714

would be an easy analogous update I think! If you submit, I'll approve.

schveiguy:
Failing test fixed, let's see if it passes now, thanks @JinShil

JinShil (approved):

MartinNowak:
https://github.com/thaven/oauth/blob/cac7ec67af5726f89859b14682f9c798513ee403/source/oauth/session.d#L346-L348

PR #7111, opened 2017-08-31
[Trivial] Enable linker demangling for Posix

JinShil:


JinShil:
> /home/travis/build/dlang/dmd/linker/ld: fatal error: unknown demangling style 'dlang'

Are the travis-ci machines using a really old version of binutils?

mathias-lang-sociomantic:
Will that be distributed with the release ?

jpf91:
I don't think Travis-CI provides newer environments: https://github.com/travis-ci/travis-ci/issues/5821

In GDC we'd probably check ld features when `./configuring` the compiler, but of course this is not possible with DMD's binary distribution approach. A possible solution is checking the linker version at runtime (ld -v), but this is a non-trivial change.

JinShil:
> Will that be distributed with the release?

If what you mean by "that" is "dmd.conf file with -L--demangle=dlang", then yes, my plan was to have that distributed with the release so demangling worked out of the box.

But I can't get it past the travis-ci Ubuntu Trusty tests, so it doesn't look very promising.  I think Ubuntu Trusty has a much older binutils package that doesn't have the dlang demangling support.

JinShil:
> A possible solution is checking the linker version at runtime (ld -v), but this is a non-trivial change.

Yeah, that doesn't sound like fun.  We could use docker packages for travis-ci, but I'm not sure travis-ci is testing for that the auto-tester is not.

mathias-lang-sociomantic:
> If what you mean by "that" is "dmd.conf file with -L--demangle=dlang", then yes, my plan was to have that distributed with the release so demangling worked out of the box.

While trusty is still widespread, I think it's too early. It goes EOL by April 2019, and as you can see many services still use it. I was myself still using it just 4 months ago.

>  I think Ubuntu Trusty has a much older binutils package that doesn't have the dlang demangling support.

Yep, that's the case :(

If the issue is with the distributed package, you could contact their author and ask them to distribute it with the correct demangling. Also what's the impact on this for `extern(C++)` symbol ?

JinShil:
> Also what's the impact on this for extern(C++) symbol ?

Don't know.  Haven't tested that.

Anyway, I don't think this is going to go anywhere.  Closing.

PR #7116, opened 2017-09-02
Trying to make the code more readable

ghost:
Hi all,

I'm running some C code analysis tool that pointed me to this small code fragment. I'm suggesting this patch to try to make the source code more understandable.

Would you guys agree with this type of changes? The tools pointed me to almost many cases like this one in the source code of the project.


CyberShadow:
I don't see an issue with the current formulation: chaining ternary expressions is a common pattern in some codebases. The whitespace could be improved, though, e.g. I think it would be more readable in the form:

```d
	cs.Iop =
		config.avx >= 2  ? VPBROADCASTQ :
		tysize(ty) == 32 ? VBROADCASTSD :
		PUNPCKLQDQ;
```

schlupa:
Sorry, the change doesn't make it more readable. It bloats the source unnecessarily. The ternary (i.e. functional approach) has the advantage of limiting the number of things that can happen within each branch of the test. 
It is true that beginners struggle sometimes with the nested ternaries but the dmd code-base isn't addressed at beginners.

WalterBright:
There's no consensus that this is an improvement, so I think we should pass on it.

PR #7118, opened 2017-09-04
fix Issue 17800 - [2.076] "static foreach" allocates closures in GC without reason

tgehr:
The foreach ranges should be semantically analyzed and interpreted before lowering.

### DIFF
--- a/src/ddmd/declaration.d
+++ b/src/ddmd/declaration.d
@@ -1067,7 +1067,7 @@ extern (C++) class VarDeclaration : Declaration
         {
             if (!type && !_init)
             {
-                printf("VarDeclaration('%s')\n", id.toChars());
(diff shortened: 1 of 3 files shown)


src/ddmd/declaration.d
```diff
@@ -907,7 +907,7 @@ extern (C++) class VarDeclaration : Declaration
         {
             if (!type && !_init)
             {
-                printf("VarDeclaration('%s')\n", id.toChars());
+                //printf("VarDeclaration('%s')\n", id.toChars());
```
ibuclaw: Did you grep for any others?

Shouldn't there be an ICE here instead of a noop?

tgehr: I commented it out because the compiler did not build. I assume this is an accidental leftover debug output.

9il:
Thanks!

PR #7120, opened 2017-09-04
fix Issue 17807 - Spurious dead code warnings on enum and static variables.

tgehr:
Local declarations (except variable declarations that are neither `enum` nor `static`) have no associated code, and therefore should not give dead code warnings.

### DIFF
--- a/src/ddmd/expression.d
+++ b/src/ddmd/expression.d
@@ -3594,6 +3594,11 @@ extern (C++) abstract class Expression : RootObject
         return .op_overload(this, sc);
     }
 
+    bool hasCode()
+    {
+        return true;
+    }
+
     void accept(Visitor v)
     {
         v.visit(this);
@@ -5711,6 +5716,15 @@ extern (C++) final class DeclarationExp : Expression
     override Expression syntaxCopy()
     {
         return new DeclarationExp(loc, declaration.syntaxCopy(null));
+    }
+
+    override bool hasCode()
+    {
+        if (auto vd = declaration.isVarDeclaration())
+        {
+            return !(vd.storage_class & (STCmanifest | STCstatic));
+        }
+        return false;
     }
 
     override void accept(Visitor v)
--- a/src/ddmd/expression.h
+++ b/src/ddmd/expression.h
@@ -214,6 +214,11 @@ public:
         return ::op_overload(this, sc);
     }
 
+    virtual bool hasCode()
+    {
+        return true;
+    }
+
     virtual void accept(Visitor *v) { v->visit(this); }
 };
 
@@ -611,6 +616,8 @@ public:
     Dsymbol *declaration;
 
     Expression *syntaxCopy();
+
+    bool hasCode();
 
     void accept(Visitor *v) { v->visit(this); }
 };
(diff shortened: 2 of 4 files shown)


src/ddmd/expression.h
```diff
@@ -612,6 +617,8 @@ class DeclarationExp : public Expression
 
     Expression *syntaxCopy();
 
+	virtual bool hasCode();
```
ibuclaw: Indentation

ibuclaw: And virtual shouldn't be needed.

src/ddmd/expression.d
```diff
@@ -5713,6 +5718,15 @@ extern (C++) final class DeclarationExp : Expression
         return new DeclarationExp(loc, declaration.syntaxCopy(null));
     }
 
+    override bool hasCode()
+    {
+        if (auto vd = declaration.isVarDeclaration())
+        {
+            return !(vd.storage_class & (STCmanifest | STCstatic));
```
jpf91: Can you elaborate why the compiler should not complain about static variables? They do not generate code but take up space in the data segment, so a warning seems reasonable nevertheless?

tgehr: Control flow passing by the static variable declaration does not do anything. Your argument justifies warnings for unused static variables, but not static variables that control flow does not reach.

WalterBright: @tgehr I agree.

WalterBright:
Trailing whitespace in expression.h is what is preventing this from passing the autotester. Please fix. I use the tool `detab` to do that.

src/ddmd/expression.d
jpf91: OK, but is there actually a case where a _static variable declaration_ is not reached by control flow __and__ the variable is still _accessible from some place reached by code flow_?

```d
void main()
{
    return;

    {
        static int x;
    }
}
```
My line of thought goes like this: If the declaration is unreachable, all code in the scope of the declaration following the declaration is unreachable as well. The variable is only accessible in the same scope (or a nested scope) defined after the variable declaration. So, such a variable seems to be always unused/used only by dead code.

OTOH `int x = void;` also does not generate any code, so I don't really understand the `passing by the static variable declaration does not do anything` point. Or are you just arguing that these cases are still invalid, but should produce a `variable unused` message instead of the `statement not reachable` message?

test/compilable/test17807.d
```diff
@@ -0,0 +1,17 @@
...
+    int y=0;
+    Lswitch: switch(2){
+        { case 0: break; }
+        enum x=0;
+        struct S{ enum x=0; }
+        int foo(){
+            return 0;
+        }
+        default: y=x+S.x+foo();
+        static foreach(i;1..5)
```
ibuclaw: Interestingly, if you remove `static` here, there's a ICE in the backend.  Has a bug report been raised for this?

tgehr: Nice catch. I couldn't find an existing issue. I have opened a new issue here: https://issues.dlang.org/show_bug.cgi?id=17834
(However, this bug will disappear once the deprecation messages are turned into compilation errors.)

PR #7126, opened 2017-09-07
fix Issue 17814 - bad output of "static foreach" with -vcg-ast

tgehr:
This implements PrettyPrintVisitor for StaticForeachDeclaration.

I'm not sure about the -vgc-ast test (I've just added new code). Is the output checked?

### DIFF
--- a/src/ddmd/hdrgen.d
+++ b/src/ddmd/hdrgen.d
@@ -1792,6 +1792,8 @@ public:
 
     override void visit(AliasDeclaration d)
     {
+        if (d.storage_class & STClocal)
+            return;
         buf.writestring("alias ");
         if (d.aliassym)
         {
@@ -1823,6 +1825,8 @@ public:
 
     override void visit(VarDeclaration d)
     {
+        if (d.storage_class & STClocal)
+            return;
         visitVarDecl(d, false);
         buf.writeByte(';');
         buf.writenl();
--- a/src/ddmd/hdrgen.d
+++ b/src/ddmd/hdrgen.d
@@ -247,7 +247,7 @@ public:
         buf.writenl();
     }
 
-    override void visit(ForeachStatement s)
+    private void visitWithoutBody(ForeachStatement s)
     {
         buf.writestring(Token.toString(s.op));
         buf.writestring(" (");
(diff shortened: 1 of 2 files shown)


rainers (approved):
LGTM. Can the travis failure be ignored? It seems it tried to build against master instead of stable.

stefan-koch-sociomantic:
@tgehr the output is not checked since `-vcg-ast` cannot really guarantee a stable output.

stefan-koch-sociomantic:
@tgehr I do not agree with this fix though.
`-vcg-ast` is supposed to show the expanded versions of everything.
the static foreach should not even be in the output.

The fact that it is worries me.

tgehr:
@stefan-koch-sociomantic That's not related to `static foreach`. `static if` is not expanded at module scope either.

stefan-koch-sociomantic:
ah, that's pretty bad. 
I'll have a look later.

WalterBright:
There's lots of red in the code - meaning it isn't tested. Are you using the browser plugin that shows which lines are executed by the test suite?

tgehr:
@WalterBright: I usually just add tests that exercise all new functionality. My commit actually increases the total code coverage slightly (because PrettyPrintVisitor.visit(ForeachStatement) had not yet been covered). However, it is not possible to cover some of the code using external code with the -vcg-ast switch: `static foreach` over a range is lowered to `static foreach` over AliasSeq and `static foreach` statements are unrolled before -vcg-ast prints them.

wilzbach (approved):
@WalterBright FWIW this even increases the overall project coverage...

WalterBright:
@tgehr is it possible to execute the code without the `-vcg-ast` switch being thrown?

tgehr:
@WalterBright: The bug report was about `-vcg-ast`, `-vcg-ast` uses the `PrettyPrintVisitor` to format source code. (I.e. `-vcg-ast` should not decrease `PrettyPrintVisitor` coverage.) I don't know DMD well enough to judge whether there is any other way to cover the parts of the `PrettyPrintVisitor` relevant for this pull request (but most code in it is currently uncovered). My best guess is that there is no such way (i.e., some of the code in the `PrettyPrintVisitor` is probably dead code unless someone is printing some AST information during debugging).

PR #7127, opened 2017-09-07
fix Issue 17812 - crash when building JSON info with static foreach

tgehr:
This adds the string representation "__local" for STClocal.

### DIFF
--- a/src/ddmd/hdrgen.d
+++ b/src/ddmd/hdrgen.d
@@ -3192,6 +3192,7 @@ extern (C++) const(char)* stcToChars(ref StorageClass stc)
         SCstring(STCsystem, TOKat, "@system"),
         SCstring(STCdisable, TOKat, "@disable"),
         SCstring(STCfuture, TOKat, "@__future"),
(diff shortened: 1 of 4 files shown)


rainers:
Isn't `__local` something that should be invisible to the user?

Anyway, the reference JSON file needs to be updated with the new declarations.

tgehr:
@rainers: I don't exactly know the purpose of the json/vcg-ast output. (I was not really aware of -X and -vcg-ast, that's why I hadn't tested them.) I have removed __local declarations from the output of both in the respective pull requests.

tgehr:
Any ideas why the auto-tester fails `cpp_abi_tests.d`? (I have not even come close to code dealing with the C++ ABI.)

rainers:
> Any ideas why the auto-tester fails cpp_abi_tests.d?

It looks like it is still failing on the json output. The shell script for comparing with json.out filters out the file names.

tgehr:
@rainers Thanks! The auto-tester is green now.

PR #7128, opened 2017-09-08
optimize unsigned / to >=

WalterBright:
Doing this also uncovered a bug in OPbtst conversion, which is also fixed here.


PetarKirov (approved):

andralex (approved):
noice

PR #7132, opened 2017-09-10
fix Issue 17787 - Add a BetterC predefined version so libraries can adapt

WalterBright:


test/compilable/betterCarray.d
```diff
@@ -15,3 +15,12 @@ int foo(int[] a, int i)
 {
     return a[i];
 }
+
+// https://issues.dlang.org/show_bug.cgi?id=17787
+version (D_betterC)
```
ZombineDev: Since all other predefined versions prefixed by `D_` are using the PascalCase naming convention,  perhaps we use `D_BetterC` for consistency.

PetarKirov (commented):
Thanks for doing this! One minor nitpick:

test/compilable/betterCarray.d
WalterBright: I decided that consistently calling it `betterC` was a better idea :-)

WalterBright:
Doc pull: https://github.com/dlang/dlang.org/pull/1892

test/compilable/betterCarray.d
ZombineDev: Fair enough

PetarKirov (approved):

JinShil (approved):

wilzbach (approved):

WalterBright:
> Please let's make this D_BetterC instead. We also have D_Coverage (rather than D_cov) and D_NoBoundsChecks (rather than D_noboundscheck).

@klickverbot I agree. How about making a PR to do it?

WalterBright:
> per-feature versions

I'd rather not, as one could as well argue that every semantic change to the language should have a predefined version. Things could turn into a hopeless swamp very easily. Besides, there are very few current users of betterC and I expect it to settle down fairly quickly.

ZombineDev:
> How about making a PR to do it?

@WalterBright Here it is: https://github.com/dlang/dmd/pull/7248

PR #7134, opened 2017-09-12
refactor: make TemplateParameterSemanticVisitor private

WalterBright:
Makes most of the code private, hence better encapsulated. Also renamed `semantic` to `tpsemantic` because of excessive and pointless overloading of `semantic`, making things impractical to grep.

### DIFF
--- a/src/ddmd/dsymbolsem.d
+++ b/src/ddmd/dsymbolsem.d
@@ -62,6 +62,7 @@ import ddmd.utils;
 import ddmd.semantic;
 import ddmd.statement;
 import ddmd.target;
+import ddmd.templateparamsem;
 import ddmd.visitor;
 
 enum LOG = false;
@@ -3633,7 +3634,7 @@ extern(C++) final class DsymbolSemanticVisitor : Visitor
                 error(tp.loc, "parameter '%s' multiply defined", tp.ident.toChars());
                 tempdecl.errors = true;
             }
-            if (!tp.semantic(paramscope, tempdecl.parameters))
+            if (!tp.tpsemantic(paramscope, tempdecl.parameters))
             {
                 tempdecl.errors = true;
             }
--- a/src/ddmd/templateparamsem.d
+++ b/src/ddmd/templateparamsem.d
@@ -21,11 +21,29 @@ import ddmd.dtemplate;
 import ddmd.globals;
 import ddmd.expression;
 import ddmd.root.rootobject;
-import ddmd.mtype;
 import ddmd.semantic;
+import ddmd.mtype;
 import ddmd.visitor;
 
-extern (C++) final class TemplateParameterSemanticVisitor : Visitor
+/************************************************
+ * Performs semantic on TemplateParameter AST nodes.
+ *
+ * Params:
+ *      tp = element of `parameters` to be semantically analyzed
+ *      sc = context
+ *      parameters = array of `TemplateParameters` supplied to the `TemplateDeclaration`
+ * Returns:
+ *      `true` if no errors
+ */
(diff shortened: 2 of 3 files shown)


src/ddmd/templateparamsem.d
```diff
@@ -21,11 +21,29 @@ import ddmd.dtemplate;
...
+ *      `true` if no errors
+ */
+extern(C++) bool tpsemantic(TemplateParameter tp, Scope* sc, TemplateParameters* parameters)
+{
+    scope v = new TemplateParameterSemanticVisitor(sc, parameters);
+    tp.accept(v);
+    return v.result;
+}
+
+
```
wilzbach: Isn't the standard one empty line between functions?

src/ddmd/dsymbolsem.d
```diff
@@ -62,6 +62,7 @@ import ddmd.utils;
 import ddmd.semantic;
 import ddmd.statement;
 import ddmd.target;
+import ddmd.templateparamsem;
 import ddmd.visitor;
```
wilzbach: How about starting to use local, selective imports?
This would also provide the additional benefit that the code is easier readable ;-)

wilzbach (approved):

wilzbach:
@CyberShadow we seem to get a lot of failed builds due to the dub registry lately. While the upcoming DUB 1.5 will add support for fallbacks, we won't be able to use it until November. Do you by chance have a good/simple idea until then?

CyberShadow:
I have an idea on how to implement restarting builds on DAutoTest. Not sure how to tie in GitHub authentication / integration yet though, haven't worked with OAuth much before.

That the dub registry is down a lot is a problem that affects not just our CIs, so it may be more effective to attack the problem at its source and migrate the registry to a server with good uptime and higher bus factor.

JinShil:
> I have an idea on how to implement restarting builds on DAutoTest

I'm not sure who manages Travis CI, but it would be nice if we could restart those tests also, as they seem to be sensitive to DUB failures as well.

src/ddmd/dsymbolsem.d
```diff
@@ -3633,7 +3634,7 @@ extern(C++) final class DsymbolSemanticVisitor : Visitor
                 error(tp.loc, "parameter '%s' multiply defined", tp.ident.toChars());
                 tempdecl.errors = true;
             }
-            if (!tp.semantic(paramscope, tempdecl.parameters))
+            if (!tp.tpsemantic(paramscope, tempdecl.parameters))
```
JinShil: Why the redundant `tp` in `tpsemantic`?

JinShil (approved):

wilzbach:
> I'm not sure who manages Travis CI, but it would be nice if we could restart those tests also, as they seem to be sensitive to DUB failures as well.

Travis is a company. Everyone with write access can restart these tests (sadly not the author) through their Web UI. As an other closing and reopening is a simple way to restart too.
However, of course, the idea is to avoid spurious failures in the first place.

src/ddmd/dsymbolsem.d
WalterBright: Because `semantic` has too many pointless overloads making it hard to search for.

src/ddmd/templateparamsem.d
WalterBright: There's no convention on that, and it is not worth the time to establish one.

src/ddmd/dsymbolsem.d
WalterBright: Let's stick to this PR in this thread.

PR #7139, opened 2017-09-16
fix Issue 17831 - [ICE] Internal error: backend/symbol.c 1039

ibuclaw:
https://issues.dlang.org/show_bug.cgi?id=17831

ICE was found by compiling a variant of test for #7120.  This causes an ICE in gdc also.  Both code and lowering is just nonsense, e.g:
```
if (i == 0)  // ICE: 'i' is not declared in this scope.
{
    int i = 0;
    ...
}
```
So make it an error.

### DIFF
--- a/src/ddmd/statementsem.d
+++ b/src/ddmd/statementsem.d
@@ -2478,7 +2478,6 @@ else
                     sw.hasVars = 1;
 
                     /* TODO check if v can be uninitialized at that point.
-                     * Also check if the VarExp is declared in a scope outside of this one
                      */
                     if (!v.isConst() && !v.isImmutable())
                     {
@@ -2489,6 +2488,24 @@ else
                     {
                         cs.error("case variables not allowed in final switch statements");
                         errors = true;
+                    }
+
+                    /* Also check if the VarExp is declared in a scope outside of this one
+                     * 'scx' is set to the scope of the switch statement.
+                     */
+                    for (Scope* scx = sc; scx; scx = scx.enclosing)
+                    {
+                        if (scx.enclosing && scx.enclosing.sw == sw)
+                            continue;
+                        assert(scx.sw == sw);
+
+                        if (!scx.search(cs.exp.loc, v.ident, null))
+                        {
+                            cs.error("case variable `%s` declared at %s cannot be declared in switch body",
+                                v.toChars(), v.loc.toChars());
+                            errors = true;
+                        }
+                        break;
                     }
                     goto L1;
                 }
--- a/src/ddmd/statementsem.d
+++ b/src/ddmd/statementsem.d
@@ -2490,8 +2490,8 @@ else
                         errors = true;
                     }
 
-                    /* Also check if the VarExp is declared in a scope outside of this one
-                     * 'scx' is set to the scope of the switch statement.
+                    /* Find the outermost scope `scx` that set `sw`.
+                     * Then search scope `scx` for a declaration of `v`.
                      */
                     for (Scope* scx = sc; scx; scx = scx.enclosing)
                     {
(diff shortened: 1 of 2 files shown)


ibuclaw:
Suggestions on diagnostic improvements welcome.

src/ddmd/statementsem.d
```diff
@@ -2490,6 +2489,19 @@ else
                         cs.error("case variables not allowed in final switch statements");
                         errors = true;
                     }
+
+                    /* Also check if the VarExp is declared in a scope outside of this one
+                     * 'sc2' is set to the scope of the switch statement.
+                     */
+                    Scope* sc2 = sc;
+                    while (sc2.enclosing && sc2.enclosing.sw == sw)
```
WalterBright: I'd rather this was turned into a `for` loop where the scope of `sc2` does not extend beyond the `for` statement's scope.

src/ddmd/statementsem.d
```diff
@@ -2490,6 +2489,19 @@ else
...
+                        sc2 = sc2.enclosing;
+                    if (!sc2.search(cs.exp.loc, v.ident, null))
+                    {
+                        cs.error("case variables cannot be declared in switch bodies");
```
WalterBright: Please print the variable being declared, something like:
```
"case variable `%s` declared at %s cannot be declared in switch body", v.toChars(), v.loc.toChars());
```

ibuclaw: I can't say I'm a fan of putting line numbers in error messages, but ok.

Just as a note, having a separate diagnostic per line is more natural for GDC (LDC too?) as the entire source line is printed and a `^` is inserted at the character position. When at the GNU Cauldron this month, they are wanting to add location ranges also to underline keywords, constants, and expressions (`^~~~`).

Anyway, that's just food for thought right now.

test/fail_compilation/ice17831.d
```diff
@@ -0,0 +1,69 @@
...
+}
+
+void test17831c()
+{
+    switch (0)
+    {
+        int i = 0;
+        while (i++ < 5)
+        {
+            case i:
```
ibuclaw: @WalterBright - thinking a bit more about this. Maybe it would be better to explicitly remove the switch statement reference from `Scope` when pushing a new for loop bodies (`sc = sc.push(); sc.sw = null;`), thus disallowing cases outright from appearing inside loop bodies.

This would be a bit more breaking, but I can't think of any justification for allowing this.

src/ddmd/statementsem.d
```diff
@@ -2490,6 +2489,24 @@ else
...
                     }
+
+                    /* Also check if the VarExp is declared in a scope outside of this one
+                     * 'scx' is set to the scope of the switch statement.
+                     */
+                    for (Scope* scx = sc; scx; scx = scx.enclosing)
+                    {
+                        if (scx.enclosing && scx.enclosing.sw == sw)
+                            continue;
+                        assert(scx.sw == sw);
```
ibuclaw: Give me a moment while I work out this assert.

WalterBright: It looks like the above 3 lines could be replaced with simply:
```
if (scx.sw != sw) continue;
```

ibuclaw: That is not the intent, and that won't work.

When entering the for loop, it is already guaranteed that `scx.sw == sw`. What we want however is the first scope that set `sw`. This means traversing all enclosing scopes until `scx.enclosing.sw != sw`.

WalterBright: Ah, I see. So, can you change the comment on the loop to reflect this? Your new explanation is much, much better than the code comment!

WalterBright: @ibuclaw waiting for your fix to the comment!

WalterBright: eh, fixed it myself

ibuclaw: Oops, I was going to get back to that this morning, now that I have a fully working compiler for gdc/ddmd.  Thanks!

PR #7141, opened 2017-09-16
Add convenience function for Identifier::idPool

ibuclaw:
GDC uses `Identifier::idPool("some const string");` in many places.  Though this convenient function was removed on the D side in favour of using `string`.  It's still useful on the C++ side to have this.

If this is ok, then I'll add a few more methods like this which are lacking for GDC.

### DIFF
--- a/src/ddmd/identifier.h
+++ b/src/ddmd/identifier.h
@@ -40,6 +40,12 @@ public:
     static Identifier *generateId(const char *prefix);
     static Identifier *generateId(const char *prefix, size_t i);
     static Identifier *idPool(const char *s, size_t len);
+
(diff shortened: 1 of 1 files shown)


WalterBright:
If you put them in the `.h` file only, it's ok with me. I'd rather be moving the rest of the D codebase towards using D `strings` and removing legacy 0 terminated strings.

ibuclaw:
Yeah, there's not too much that needs to be handled across the C++/D boundary.

The other two things that I'm thinking about are:
* static ::create methods that wrap `new ()`
* An interface for compiler specific actions - initialize, genCmain, parse files, compile, cleanups. Maybe can extend `Compiler` here.

PR #7143, opened 2017-09-17
Add static create methods for frequently used types and declarations.

ibuclaw:
In GDC, these are used in the C++ parts of the compiler for:
 - Building types and function declarations for the gcc.builtins module.
 - Implement CTFE-able builtins.

Though I guess ideally there should be a create for all front-end AST types.  Or some generic way to request a new class from C++ to D.


ibuclaw:
This PR is in conflict with #7144, so don't merge both at the same time.

PR #6318, opened 2016-12-11
Don't set ModuleInfo unless currently parsing object.d

ibuclaw:
Just gives the same guarantee that `Module.moduleinfo` is not a `struct ModuleInfo` from another module.

This same check is done for `TypeInfo`, `Error`, and many others that are handled in `ClassDeclaration.this`.

PR #7144, opened 2017-09-17
Don't set ModuleInfo unless currently parsing object.d

ibuclaw:
This is taken from gdc's implementation.  The type declaration is used directly when building module information (e.g: for debug), so shouldn't allow some other module to override its implementation.


ibuclaw:
Apparently I've already raised this as a PR once before in #6318.

MartinNowak (approved):

PR #5210, opened 2015-10-20
lexer.d: make unittest work

WalterBright:
The unittest translation was incorrect, and wasn't proper unittests anyway. Fixed that. Should open the door to proper unit testing of the lexer.

PR #7145, opened 2017-09-17
Remove initTokens from tokens.h

ibuclaw:
This static method was removed in #5210


wilzbach (approved):

wilzbach:
I can only repeat myself: we should really generate these headers automatically

ibuclaw:
> I can only repeat myself: we should really generate these headers automatically

Yes, in the meantime, the work I'm doing in attempting to use the current D sources with GDC is the best tool we have at identifying all holes on both sides of the implementation.

PR #5093, opened 2015-09-17
[refactor] Statically initialize precedence table

yebblies:
100% known at compile time.

PR #7146, opened 2017-09-17
Remove initPrecedence from expression.h

ibuclaw:
Removed in #5093


ibuclaw:
FYI, these things I'm removing were found because gdc was still calling them from C++.  So ended up as linker errors when switching the front-end to D.

WalterBright:
I'm glad there's progress in switching to the D front end!

ibuclaw:
@WalterBright of the linker errors, only long double vs. real_t remain. But I guess that is just the end of the beginning. ☺️

PR #7149, opened 2017-09-17
Fix signature of Port::memicmp to match D declaration

ibuclaw:
A case of `int` vs. `size_t`.

### DIFF
--- a/src/ddmd/root/port.d
+++ b/src/ddmd/root/port.d
@@ -38,7 +38,7 @@ extern (C++) struct Port
     {
         int result = 0;
 
-        for (int i = 0; i < n; i++)
+        foreach (i; 0 .. n)
         {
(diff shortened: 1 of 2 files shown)


src/ddmd/root/port.d
```diff
@@ -38,7 +38,7 @@ extern (C++) struct Port
     {
         int result = 0;
 
-        for (int i = 0; i < n; i++)
+        for (size_t i = 0; i < n; i++)
```
WalterBright: Might as well go whole hog with:
```
foreach (i; 0 .. n)
```

ibuclaw: Right, I'm still stuck in doing C++-isms.

ibuclaw:
Fixed up to use foreach.

PR #7150, opened 2017-09-18
reproducible builds

MartinNowak:
- fixed by appending all fixups to an array and emitting
  them when finishing an object file
- also faster than emitting fixups in-between using a hash
  table and singly-linked fixup list

[reproducible-builds.org](https://reproducible-builds.org/)

### DIFF
--- a/src/ddmd/backend/cgen.c
+++ b/src/ddmd/backend/cgen.c
@@ -713,181 +713,76 @@ void regwithvalue(CodeBuilder& cdb,regm_t regm,targ_size_t value,unsigned *preg,
 
 /************************
  * When we don't know whether a function symbol is defined or not
- * within this module, we stuff it in this linked list of references
- * to be fixed up later.
+ * within this module, we stuff it in an array of references to be
+ * fixed up later.
  */
-
-struct fixlist
+struct Fixup
 {
-    int         Lseg;           // where the fixup is going (CODE or DATA, never UDATA)
-    int         Lflags;         // CFxxxx
-    targ_size_t Loffset;        // addr of reference to symbol
-    targ_size_t Lval;           // value to add into location
+    symbol      *sym;       // the referenced symbol
+    int         seg;        // where the fixup is going (CODE or DATA, never UDATA)
+    int         flags;      // CFxxxx
+    targ_size_t offset;     // addr of reference to symbol
+    targ_size_t val;        // value to add into location
 #if TARGET_OSX
-    symbol      *Lfuncsym;      // function the symbol goes in
+    symbol      *funcsym;   // function the symbol goes in
 #endif
-    fixlist *Lnext;             // next in threaded list
-
-    static int nodel;           // don't delete from within searchfixlist
 };
 
-int fixlist::nodel = 0;
-
-/* The AArray, being hashed on the pointer value of the symbol s, is in a different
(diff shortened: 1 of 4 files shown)


src/ddmd/backend/cgen.c
```diff
@@ -913,150 +808,79 @@ size_t addtofixlist(symbol *s,targ_size_t soffset,int seg,targ_size_t val,int fl
...
-            objmod->reftoident(ln->Lseg,ln->Loffset,s,ln->Lval,ln->Lflags);
+    Obj::reftoident(f.seg, f.offset, f.sym, f.val, f.flags);
 #endif
-            *plnext = ln->Lnext;
-#if TERMCODE
-            mem_free(ln);
-#endif
-        }
-    }
-    s->Sxtrnnum = 0;
```
MartinNowak: Already handled by symbol_reset.
https://github.com/dlang/dmd/blob/4d86fcba2fd2ef86cc85738cd2ac2b059dbb5800/src/ddmd/backend/symbol.c#L2290

src/ddmd/backend/code.h
```diff
@@ -426,7 +426,7 @@ unsigned calcblksize (code *c);
 unsigned calccodsize(code *c);
 unsigned codout(int seg, code *c);
 size_t addtofixlist (symbol *s , targ_size_t soffset , int seg , targ_size_t val , int flags );
-void searchfixlist (symbol *s );
+inline void searchfixlist (symbol *s ) {}
```
MartinNowak: Kept around as noop, as it would be quite some effort to find all the locations in case we ever want to change the mechanism back to emit fixups right after a symbol gets defined.

ibuclaw:
Wasn't there an issue raised about this?  I certainly remember hitting it a couple years ago and asking about it.

MartinNowak:
> Wasn't there an issue raised about this? I certainly remember hitting it a couple years ago and asking about it.

None found, but it was certainly discussed a couple of times.

WalterBright:
Thanks, @MartinNowak 
The reason for the original design was to minimize the memory consumption of the compiler, which was critical on DOS systems. This is no longer an issue.

src/ddmd/backend/cgen.c
```diff
@@ -713,181 +713,76 @@ void regwithvalue(CodeBuilder& cdb,regm_t regm,targ_size_t value,unsigned *preg,
...
-        for (size_t i = 0; i < flarray_dim; i++)
+        if (dim == cap)
         {
-            if (flarray[i].s == s)
-            {
-                lp = &flarray[i].fl;
-                break;
-            }
+            cap = cap ? 2 * cap : 16;
+            ptr = (Fixup *)::mem_realloc(ptr, cap * sizeof(Fixup));
```
WalterBright: I bet if you put a print statement here, there will be a bunch of reallocations and copying. I suggest putting in a print, and then adjusting the initial size to be big enough for most compiles. 16 is way, way too small.

WalterBright:
I as well like the large reduction in lines of code and the simplification.

MartinNowak:
Why are you merging PRs with failing tests?

WalterBright:
This PR caused a regression: https://issues.dlang.org/show_bug.cgi?id=18412

WalterBright:
Although this PR exposed the bug 18412, it did not cause it.

PR #7151, opened 2017-09-20
fix Issue 17843 - -betterC struct with  field generates references to…

WalterBright:
… TypeInfo

### DIFF
--- a/src/ddmd/dsymbolsem.d
+++ b/src/ddmd/dsymbolsem.d
@@ -5473,9 +5473,12 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         buildOpAssign(sd, sc2);
         buildOpEquals(sd, sc2);
 
-        sd.xeq = buildXopEquals(sd, sc2);
-        sd.xcmp = buildXopCmp(sd, sc2);
-        sd.xhash = buildXtoHash(sd, sc2);
+        if (global.params.useTypeInfo)  // these functions are used for TypeInfo
+        {
+            sd.xeq = buildXopEquals(sd, sc2);
+            sd.xcmp = buildXopCmp(sd, sc2);
+            sd.xhash = buildXtoHash(sd, sc2);
+        }
 
         sd.inv = buildInv(sd, sc2);
 
--- a/src/ddmd/clone.d
+++ b/src/ddmd/clone.d
@@ -695,8 +695,9 @@ private bool needToHash(StructDeclaration sd)
         }
         if (tv.isfloating())
         {
-            // This is necessray for:
-            //  1. comparison of +0.0 and -0.0 should be true.
+            /* This is necessary because comparison of +0.0 and -0.0 should be true,
+             * i.e. not a bit compare.
+             */
             goto Lneed;
         }
         if (tv.ty == Tarray)
--- a/test/runnable/betterc.d
+++ b/test/runnable/betterc.d
@@ -10,6 +10,17 @@ void test(int ij)
     assert(ij,"it is not zero");
 }
 
+/*******************************************/
+// https://issues.dlang.org/show_bug.cgi?id=17843
+
+struct S
+{
+    double d = 0.0;
+    int[] x;
+}
+
+/*******************************************/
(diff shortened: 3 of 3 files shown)


test/runnable/cassert.d
```diff
@@ -10,6 +10,13 @@ void test(int ij)
     assert(ij,"it is not zero");
```
MartinNowak: That test file better be named `betterc.d` ;).

MartinNowak:
> If any of the fields has an opEquals, then we need it too.

Sounds similarly problematic, wouldn't it be better to just disable the toHash altogether? The runtime based AA is out of reach for -betterC anyhow.

WalterBright:
I went for disabling as little as possible.

test/runnable/cassert.d
WalterBright: I use grep to search for relevant test cases anyway. The filenames of the test cases aren't particularly helpful.

WalterBright:
`toHash` is only really of use for associative arrays, and those won't work with `-betterC` anyway (i.e. they won't link).

WalterBright:
> A large part of the appeal of -betterC is to be able to draw on existing D code when targeting constrained environments.

I think that is a bit of a misunderstanding. `betterC` code does not link to Phobos or druntime at all, and very little existing D code will work that way unless it is specifically crafted that way.

`betterC` is a tool to enable D code to be written with zero overhead, so that there's no longer any reason to use C. It can also be used to gradually convert existing C code to D. I'm doing that now with the DMC++ compiler, which is how I ran into this PR issue.

WalterBright:
> array members?

Yes. Fixed. Good catch!

test/runnable/cassert.d
WalterBright: renamed

MartinNowak (changes requested):
Sorry, I still don't follow. You're modifying the function `needToHash` which is used to decide whether or not TypeInfo gets an auto-generated xtoHash function.
Without TypeInfo those xtoHash/xopEquals/xopCmp functions are not accessible anyhow.
Seems much smarter to just disable the auto-generation instead.
https://github.com/dlang/dmd/blob/241236394b303a1ea396fe8e20a0c25b5c09a90e/src/ddmd/dsymbolsem.d#L5425-L5427

Furthermore you want to split the monolithic betterC switch into disableTypeInfo, disableEH, disableX right away, so we don't have to unravel that later on.

WalterBright:
@MartinNowak I disabled the 3 functions from being generated with -betterC, but splitting the flag into multiple ones seems a bit excessive at the moment. Catching what it the flag does is pretty simple, grepping it is just a few hits.

andralex (approved):
I'm good with this. @MartinNowak ?

andralex:
cc @MartinNowak

andralex (approved):

src/ddmd/dsymbolsem.d
```diff
@@ -5403,9 +5403,12 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         buildOpAssign(sd, sc2);
         buildOpEquals(sd, sc2);
 
-        sd.xeq = buildXopEquals(sd, sc2);
-        sd.xcmp = buildXopCmp(sd, sc2);
-        sd.xhash = buildXtoHash(sd, sc2);
+        if (!global.params.betterC)  // these functions are used for TypeInfo
```
JinShil: Due to #7288, I think this should be checking the `useTypeInfo` flag.

JinShil (changes requested):

src/ddmd/dsymbolsem.d
WalterBright: done

MartinNowak: The splitting of the betterC options was obviously coming ;), glad it's in now.

ibuclaw: Yes, I introduced that.  Please continue to enforce. :-)

PR #7153, opened 2017-09-22
fix PR 7150 - reproducible builds

WalterBright:
Fixes for https://github.com/dlang/dmd/pull/7150

1. initial size of fixup array is too small
2. fixes to work with DMC++ front end

Depends on #7165

### DIFF
--- a/src/ddmd/backend/cgen.c
+++ b/src/ddmd/backend/cgen.c
@@ -709,7 +709,10 @@ struct FixupArray
     {
         if (dim == cap)
         {
-            cap = cap ? 2 * cap : 16;
+            // 0x800 determined experimentally to minimize reallocations
+            cap = cap
+                ? (3 * cap) / 2 // use 'Tau' of 1.5
+                : 0x800;
             ptr = (Fixup *)::mem_realloc(ptr, cap * sizeof(Fixup));
         }
         ptr[dim++] = e;
@@ -778,6 +781,13 @@ size_t addtofixlist(symbol *s,targ_size_t offset,int seg,targ_size_t val,int fla
         return numbytes;
 }
 
+#if 0
+void searchfixlist (symbol *s )
+{
+    //printf("searchfixlist(%s)\n", s->Sident);
+}
+#endif
+
 /****************************
  * Output fixups as references to external or static symbol.
  * First emit data for still undefined static symbols or mark non-static symbols as SCextern.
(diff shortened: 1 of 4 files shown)


src/ddmd/backend/cgen.c
```diff
@@ -812,6 +813,13 @@ size_t addtofixlist(symbol *s,targ_size_t offset,int seg,targ_size_t val,int fla
         return numbytes;
 }
 
+#if 0
+void searchfixlist (symbol *s )
+{
+    //printf("searchfixlist(%s)\n", s->Sident);
+}
+#endif
```
ZombineDev: Looks like this can be removed.

src/ddmd/backend/code.d
```diff
@@ -217,6 +217,7 @@ extern __gshared int refparam;
 targ_size_t cod3_spoff();
 uint calccodsize(code *c);
 targ_size_t cod3_bpoffset(Symbol *s);
+void searchfixlist(Symbol *s) { }
```
ZombineDev: This too.

ibuclaw:
Looking at travisci, it seems like something is still not deterministic.

src/ddmd/backend/cgen.c
```diff
@@ -743,7 +743,8 @@ struct FixupArray
     {
         if (dim == cap)
         {
-            cap = cap ? 2 * cap : 16;
+            // 0x800 determined experimentally to minimize reallocations
+            cap = cap ? 2 * cap : 0x800;
```
stefan-koch-sociomantic: 2048 ?

src/ddmd/backend/cgen.c
WalterBright: Left as a convenient debugging feature.

src/ddmd/backend/code.d
WalterBright: #7150 explains why that is there.

ZombineDev: Thanks, I missed Martin's comment there.

MartinNowak:
Why didn't you wait on me or at least check the test results?

MartinNowak (approved):

src/ddmd/backend/cgen.c
```diff
@@ -709,7 +709,10 @@ struct FixupArray
     {
         if (dim == cap)
         {
-            cap = cap ? 2 * cap : 16;
+            // 0x800 determined experimentally to minimize reallocations
+            cap = cap
+                ? (3 * cap) / 2 // use 'Tau' of 1.5
```
andralex: I assume cap never decreases, so it's either 0 or something no less than 0x800. Otherwise the code has a bug if cap = 1.

andralex (approved):

andralex:
please take a look at the test failures

src/ddmd/backend/cgen.c
klickverbot: Nitpick: `s/'Tau'/geometric growth factor/g`

WalterBright: `cap` starts at 0, and is set to 0x800 upon first use. It is never `0<cap<0x800`.

PR #7154, opened 2017-09-22
fixed a small typo in cod3.c

dunkyp:
currently can't build with BUILD=debug

### DIFF
--- a/src/ddmd/backend/cod3.c
+++ b/src/ddmd/backend/cod3.c
@@ -2479,7 +2479,7 @@ void genmovreg(CodeBuilder& cdb,unsigned to,unsigned from)
 {
 #if DEBUG
         if (to > ES || from > ES)
-                printf("genmovreg(c = %p, to = %d, from = %d)\n",c,to,from);
+                printf("genmovreg(c = %p, to = %d, from = %d)\n",cdb,to,from);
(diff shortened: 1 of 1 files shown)


src/ddmd/backend/cod3.c
```diff
@@ -2479,7 +2479,7 @@ void genmovreg(CodeBuilder& cdb,unsigned to,unsigned from)
 {
 #if DEBUG
         if (to > ES || from > ES)
-                printf("genmovreg(c = %p, to = %d, from = %d)\n",c,to,from);
+                printf("genmovreg(c = %p, to = %d, from = %d)\n",cdb,to,from);
```
rainers: I guess it should be `&cdb`. Please change the format string, too.

wilzbach:
> currently can't build with BUILD=debug

We should definitely add this to the CI setup. Circle CI maybe?
CC @CyberShadow

src/ddmd/backend/cod3.c
WalterBright: I just changed it to:
```
                printf("genmovreg(to = %d, from = %d)\n",to,from);
```

rainers (approved):
edited online.

PR #7155, opened 2017-09-23
do not use thread local static ctors to initialize __gshared variables

rainers:


PetarKirov (approved):

src/ddmd/lexer.d
```diff
@@ -75,7 +75,7 @@ bool issinglechar(char c)
     return (cmtable[c] & CMsinglechar) != 0;
 }
 
-static this()
+shared static this()
```
wilzbach: I am just looking at this from my phone, but isn't `cmtable` `immutable ubyte[256]`?

rainers: Yes. I also wondered why it's not forbidden to change it in a thread local ctor.

PR #7156, opened 2017-09-23
fix building with version=GC

rainers:
When adding GC.disable before the code generation, building phobos unittests with the split build in win64.mak works with the GC version, but is about 20% slower.

As a touchstone compiling all phobos unittests with a single compiler invocation still doesn't work completely. It finishes semantic analysis within about 8GB of memory, but later fails after using a lot more memory during code generation.

It seems using the front end as a library with the GC enabled is feasable.

### DIFF
--- a/src/ddmd/root/rmem.d
+++ b/src/ddmd/root/rmem.d
@@ -27,6 +27,7 @@ version (GC)
 
         static void xfree(void* p) nothrow
         {
+            return GC.free(p);
         }
 
         static void* xmalloc(size_t n) nothrow
@@ -42,6 +43,14 @@ version (GC)
         static void* xrealloc(void* p, size_t size) nothrow
         {
             return GC.realloc(p, size);
+        }
+        static void error() nothrow
+        {
+            import core.stdc.stdlib : exit, EXIT_FAILURE;
+            import core.stdc.stdio : printf;
+
+            printf("Error: out of memory\n");
+            exit(EXIT_FAILURE);
         }
     }
 
--- a/src/ddmd/root/rmem.d
+++ b/src/ddmd/root/rmem.d
@@ -44,6 +44,7 @@ version (GC)
         {
             return GC.realloc(p, size);
         }
+
         static void error() nothrow
         {
             import core.stdc.stdlib : exit, EXIT_FAILURE;


src/ddmd/root/rmem.d
```diff
@@ -13,6 +13,8 @@ module ddmd.root.rmem;
 // Online documentation: https://dlang.org/phobos/ddmd_root_rmem.html
 
 import core.stdc.string;
+import core.stdc.stdlib;
+import core.stdc.stdio;
```
wilzbach: While most files at DMD still use full "global" imports, there's no reason not to start using either local imports, selective ones or even both ;-)

wilzbach (approved):

src/ddmd/root/rmem.d
rainers: Ok, amended. Also added calling GC.free, no reason to omit it.

PetarKirov (approved):

src/ddmd/root/rmem.d
```diff
@@ -43,6 +44,14 @@ version (GC)
         {
             return GC.realloc(p, size);
         }
+        static void error() nothrow
```
ibuclaw: Just a nit, add a blank line between prior and next function declaration.

ibuclaw:
> When adding GC.disable before the code generation, building phobos unittests with the split build in win64.mak works with the GC version, but is about 20% slower.

But the GC isn't disabled if `version (GC)`?

ibuclaw:
> It seems using the front end as a library with the GC enabled is feasable.

I thought that gdc on travisci was using `-version=GC` for building dmd.  So it should be somewhat tested.

rainers:
> But the GC isn't disabled if version (GC)?

No, it's not. Should I add this with this PR?

> I thought that gdc on travisci was using -version=GC for building dmd. So it should be somewhat tested.

It would have failed due to the missing symbol, so I guess it's not built with -version=GC on any CI.

src/ddmd/root/rmem.d
rainers: added newline.

ibuclaw:
> It would have failed due to the missing symbol, so I guess it's not built with -version=GC on any CI.

Ah, its using mixed GC and non-GC allocations! (`OVERRIDE_MEMALLOC` is false)  I'm surprised that the gdc builds of dmd have not been hit by memory bugs yet.

ibuclaw:
>  I'm surprised that the gdc builds of dmd have not been hit by memory bugs yet.

Actually, that's what `GC.disable()` is doing.  **Now** I understand what is going on here... :smile:

ibuclaw:
> No, it's not. Should I add this with this PR?

No, it should remain on.  This PR is good to go.

ibuclaw (approved):

PR #7157, opened 2017-09-23
Fix 17502 (again): Generate contracts after inferring return type.

ibuclaw:
https://issues.dlang.org/show_bug.cgi?id=17502
 Wait until after the return type has been inferred before generating the contracts for this function, and merging contracts from overrides.

 This was originally at the end of the first semantic pass, but required a fix-up to be done here for the `__result` variable type of __ensure() inside auto functions, but this didn't work if the out parameter was implicit.

---

The test-case for this issue was still causing an ICE in gdc because for the following test:
```
class Foo
{
    auto bar()
    out { assert (__result > 5); }
    body { return 6; }
}
```
 it was still generating the following contract.
```
void __ensure ()
{
  assert (__result > 5);
}
__ensure ();
```
With this PR, it now (correctly) generates.
```
void __ensure (const ref __result)
{
  assert (__result > 5)
}
__ensure (&__result);
```

FYI @Burgos as original author. ;-)

### DIFF
--- a/src/ddmd/func.d
+++ b/src/ddmd/func.d
@@ -156,6 +156,28 @@ enum FUNCFLAGinlineScanned    = 0x20;   /// function has been scanned for inline
 enum FUNCFLAGinferScope       = 0x40;   /// infer 'scope' for parameters
 
 
+/****************************************************
+ * Determine whether an 'out' contract is declared inside
+ * the given function or any of its overrides.
+ * Params:
+ *      fd = the function to search
+ * Returns:
+ *      true    found an 'out' contract
+ */
+private bool needsFensure(FuncDeclaration fd)
+{
+    if (fd.fensure)
+        return true;
+
+    foreach (fdv; fd.foverrides)
+    {
+        if (needsFensure(fdv))
+            return true;
+    }
+    return false;
+}
+
+
 /***********************************************************
  */
 extern (C++) class FuncDeclaration : Declaration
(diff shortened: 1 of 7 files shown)


src/ddmd/func.d
```diff
@@ -156,6 +156,61 @@ enum FUNCFLAGinlineScanned    = 0x20;   /// function has been scanned for inline
...
+
+    for (size_t i = 0; i < fd.foverrides.dim; i++)
+    {
+        FuncDeclaration fdv = fd.foverrides[i];
+
+        if (fdv.frequire)
+            return true;
+
+        if (needsFrequire(fdv))
+            return true;
```
WalterBright: test cases for the two red lines 181 and 208?

src/ddmd/func.d
```diff
@@ -156,6 +156,61 @@ enum FUNCFLAGinlineScanned    = 0x20;   /// function has been scanned for inline
...
+{
+    if (fd.frequire)
+        return true;
+
+    for (size_t i = 0; i < fd.foverrides.dim; i++)
+    {
+        FuncDeclaration fdv = fd.foverrides[i];
+
+        if (fdv.frequire)
+            return true;
```
WalterBright: These two lines are not strictly necessary, as 170+171 will test it.

src/ddmd/func.d
```diff
@@ -156,6 +156,61 @@ enum FUNCFLAGinlineScanned    = 0x20;   /// function has been scanned for inline
...
+ *      false   didn't find one
+ */
+private bool needsFrequire(FuncDeclaration fd)
+{
+    if (fd.frequire)
+        return true;
+
+    for (size_t i = 0; i < fd.foverrides.dim; i++)
+    {
+        FuncDeclaration fdv = fd.foverrides[i];
```
WalterBright:
```
foreach (fdv; fd.overrides)
```
should do it.

src/ddmd/func.d
```diff
@@ -156,6 +156,61 @@ enum FUNCFLAGinlineScanned    = 0x20;   /// function has been scanned for inline
...
+
+        if (needsFrequire(fdv))
+            return true;
+    }
+    return false;
+}
+
+/****************************************************
+ * Determine whether an 'out' contract is declared inside
+ * the given function or any of its overrides.
```
WalterBright: Why is `or any of its overrides` required? It wasn't before.

WalterBright:
The code example in the explanation above needs to be added as a test case.

src/ddmd/func.d
ibuclaw: I might have misread the implementation of `mergeFensure` I was under the impression that these merge routines *could* create a new pointer where `fensure` was previously `null`. The loop and recursion logic was copied from those routines.

src/ddmd/func.d
ibuclaw: Yep. I think it will take me a while longer to get back to using foreach.

src/ddmd/func.d
ibuclaw: This is called prior to merging `fensure`, so I thought that it would be needed.

src/ddmd/func.d
ibuclaw: OK, I have test cases for them - whether this is hit depends on order of declarations.
```
class C : B { override int foo() { return 1; } }
class B : A { override int foo() { return 2; } }
class A { int foo() out(res) { assert(res != 0); } body { return 0; } }
```
From the above, functions will be analysed in order of `C.foo`, `B.foo`, `A.foo`, in which case, recursive search is required (`B.foo` hasn't started semantic, and so does not yet have `fensure` set).

src/ddmd/func.d
ibuclaw: Actually, you are right about `needsFrequire`.  If the function being analysed does not have an `in` contract, then nothing will be generated anyway.  So undoing that from this PR.

ibuclaw:
Added tests and for 3602 and 5230, and fixed up `mergeFensure` and `mergeFrequire` to make sure we don't regress on them.

ibuclaw:
Added tests to get a little extra coverage.

src/ddmd/func.d
```diff
@@ -156,6 +156,32 @@ enum FUNCFLAGinlineScanned    = 0x20;   /// function has been scanned for inline
...
+ */
+private bool needsFensure(FuncDeclaration fd)
+{
+    if (fd.fensure)
+        return true;
+
+    foreach (fdv; fd.foverrides)
+    {
+        if (fdv.fensure)
+            return true;
```
WalterBright: The above two lines are redundant.

src/ddmd/func.d
```diff
@@ -156,6 +156,32 @@ enum FUNCFLAGinlineScanned    = 0x20;   /// function has been scanned for inline
...
 
 
+/****************************************************
+ * Determine whether an 'out' contract is declared inside
+ * the given function or any of its overrides.
+ * Params:
+ *      fd = the function to search
+ * Returns:
+ *      true    found an 'out' contract
+ *      false   didn't find one
```
WalterBright: It's redundant to add the line for `false`, since it must implicitly be the negation of the `true` condition.

WalterBright:
Thanks for adding the test cases! But it would be nice to concatenate them and put them into one file, this helps the autotester run much faster, as file lookups utterly dominate the compile time of those tiny files.

ibuclaw:
Put tests into per-issue files.

ibuclaw:
Rebased.

WalterBright:
> Put tests into per-issue files.

Thank you. Many test files have 50 or more issues tested for in one file. If this was all split up, the autotester would take a *lot* longer. I'll pull this one, but in the future please consider combining them all, or even better, appending them to an existing test file.

BBasile:
This fix probably causes another regression, see https://issues.dlang.org/show_bug.cgi?id=17893.

PR #7162, opened 2017-09-24
Add prototype for 'error(Loc, const char*, ...)' to headers.

ibuclaw:
I seem to have missed this from the list of problems encountered when linking gdc.

This function was added during the transition from C++ to D if I understand correctly - so that passing a const literal worked.


ibuclaw:
Gah, actually can't do this.  C++ sees both declarations as being ambiguous.

PR #7163, opened 2017-09-24
Change error(Loc) override to extern (D)

ibuclaw:
I don't like this, but given the choice between linker errors or C++ compiler errors.  Neither is any good.

### DIFF
--- a/src/ddmd/errors.d
+++ b/src/ddmd/errors.d
@@ -43,7 +43,7 @@ extern (C++) void error(const ref Loc loc, const(char)* format, ...)
     va_end(ap);
 }
 
-extern (C++) void error(Loc loc, const(char)* format, ...)
+extern (D) void error(Loc loc, const(char)* format, ...)
 {
     va_list ap;
     va_start(ap, format);


ibuclaw:
Alternatively, could just use the other override `error(null, 0, 0, "message");`.

ibuclaw:
> Alternatively, could just use the other override error(null, 0, 0, "message");.

Infact, that is exactly what can do for the bits that matter (most of this is non-gdc stuff).

ibuclaw:
Changed signature of error to `extern (D) void error(Loc loc, const(char)* format, ...)`.  Only dmd-specific modules use it.  Adjusted shared front-end modules to call C++ version.

WalterBright:
I'm not sure what problem this is solving?

ibuclaw:
> I'm not sure what problem this is solving?

@WalterBright - this was the last linker error I was getting when building gdc against ddmd frontend.

At first I tried adding the missing declaration and implementing it.  But that resulted in C++ compiler errors, the following function signatures are considered ambiguous.
```
void error(const Loc& loc, const char *format, ...);
void error(Loc loc, const char *format, ...);
```
So the "fix" would be to make the D-specific override internal.  But then modules that are part of the common front-end should call something else, as error routines are implemented in C++ for gdc.

This could perhaps be simplified if I start moving function implementations from C++ to D also.

ibuclaw:
Moved the implementation to D on my side.  This function should still be marked as `extern(D)` though.  If there will be some automatic translation of D -> C++, this should not be picked up.

ibuclaw:
Starting this from the top as it was a bit of a rush to get over the last hurdle.

PR #7164, opened 2017-09-25
fix Issue 17857 - T.alignof ignores explicit align(N) type alignment

kinke:
It was previously the natural type alignment, defined as the maximum of the field alignments for an aggregate. Make sure an explicit `align(N)` overrides it.

### DIFF
--- a/src/ddmd/mtype.d
+++ b/src/ddmd/mtype.d
@@ -2510,7 +2510,10 @@ extern (C++) abstract class Type : RootObject
         }
         else if (ident == Id.__xalignof)
         {
-            e = new IntegerExp(loc, alignsize(), Type.tsize_t);
+            const explicitAlignment = alignment();
+            const naturalAlignment = alignsize();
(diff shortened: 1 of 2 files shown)


src/ddmd/mtype.d
```diff
@@ -2510,7 +2510,9 @@ extern (C++) abstract class Type : RootObject
         }
         else if (ident == Id.__xalignof)
         {
-            e = new IntegerExp(loc, alignsize(), Type.tsize_t);
+            const alignment = alignment();
```
WalterBright: Using the same name for a variable as its initializer looks a little wonky. I have to look at it carefully to see if there's something going wrong here - shouldn't have to do that.

WalterBright:
Will need some documentation for this new behavior.

kinke:
You mean filing an issue about it?

WalterBright:
An issue would be sufficient.

kinke:
Done.

WalterBright:
Fixed the title to match :-)

PR #6198, opened 2016-10-16
use OPpair for complex XMM values

WalterBright:
This simplifies dealing with XMM register pairs by treating them more independently. Generates better code, too.

PR #7165, opened 2017-09-25
fix uninitialized memory access

MartinNowak:
- types `ty0` and `ty1` are only lazily initialized with the
  first expression accessing a sliced field
- fixed by checking the workaround condition after all gathering

The Travis-CI test script for OSX is fixed in #7153, so those will still fail here, but the GDC tests should now succeed consistently.

### DIFF
--- a/src/ddmd/backend/gsroa.c
+++ b/src/ddmd/backend/gsroa.c
@@ -36,7 +36,7 @@ struct SymInfo
 {
     bool canSlice;
     bool accessSlice;   // if Symbol was accessed as a slice
-    bool usePair;       // will use OPpair
+    bool usePair;       // will need OPpair to recombine sliced fields
     tym_t ty0;          // type of first slice
     tym_t ty1;          // type of second slice
     SYMIDX si0;
@@ -59,11 +59,6 @@ static void sliceStructs_Gather(SymInfo *sia, elem *e)
                     {
                         // Rewrite as OPpair later
                         sia[si].usePair = true;
-
-                        /* OPpair cannot handle XMM registers, cdpair() and fixresult()
-                         */
-                        if (tyfloating(sia[si].ty0) || tyfloating(sia[si].ty1))
-                            sia[si].canSlice = false;
                     }
                     else if (sz == REGSIZE &&
                         (e->Eoffset == 0 || e->Eoffset == REGSIZE))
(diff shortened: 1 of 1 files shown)


MartinNowak:
We might be able to remove that OPpair workaround by now, https://github.com/dlang/dmd/pull/6198#issuecomment-331747797, but testing that is out of scope as I merely wanted to fix #7150.

WalterBright:
Any way to get a test case on the uninitialized access?

BTW, maybe converting the back end to D would not have helped this one, but I would like to move forward with that as there may be other bugs it can head off.

MartinNowak:
Not really, a reproducible test would rely on a targeted poisoning of the referenced malloc memory (I had to disable ASLR to reproduce this).
I don't have even more days to spend on developing such an exploit, but we're already indirectly testing this, as the gdmd build no longer fail ;).

MartinNowak:
> BTW, maybe converting the back end to D would not have helped this one, but I would like to move forward with that as there may be other bugs it can head off.

Use `new` instead of running on an uninitialized `malloc` array would have worked.
Regarding the assignment
https://github.com/dlang/dmd/blob/403d8ccf8eea3419d8f80f46fd1bfbd96ef30633/src/ddmd/backend/gsroa.c#L235-L247
once memory is in L1 cache, you're not saving that much by not initializing a few small fields.

But this is definitely a good reminder to start using `-fsanitizer=address,undefined` for testing. Not sure how feasible [`MemorySanitizer`](https://github.com/google/sanitizers/wiki/MemorySanitizer) is in terms of resources.

WalterBright:
Sanitizers, valgrind, etc., only work if there are test cases. D tries to do it via correctness checking, which is more reliable. valgrind is incredibly slow.

I would also dearly like coverage testing on the back end.

MartinNowak:
Yes valgrind is slow by design, the newer sanitizers are fairly decent though.
https://github.com/google/sanitizers/wiki
Not sure about MemorySanitizer, but I guess it has similar problems as valgrind (though they claim better performance https://github.com/google/sanitizers/wiki/MemorySanitizer).

WalterBright:
BTW, thanks for having the persistence to doggedly find (and fix) this particular bug. Uninitialized data can be some of the hardest problems to find. I'm usually pretty careful with that, but obviously I slipped up here.

PR #7166, opened 2017-09-25
outbuf: modernize buffer reallocation strategy

WalterBright:
This changes the reallocation strategy to using a 1.5x growth factor rather than a constant increment.

### DIFF
--- a/src/ddmd/backend/outbuf.d
+++ b/src/ddmd/backend/outbuf.d
@@ -26,16 +26,16 @@ struct Outbuffer
     ubyte *buf;         // the buffer itself
     ubyte *pend;        // pointer past the end of the buffer
     ubyte *p;           // current position in buffer
-    uint len;           // size of buffer
-    uint inc;           // default increment size
     ubyte *origbuf;     // external buffer
 
     //this();
 
-    this(size_t incx); // : buf(null), pend(null), p(null), len(0), inc(incx), origbuf(null) { }
+    this(size_t initialSize); // : buf(null), pend(null), p(null), origbuf(null) { }
 
-    this(ubyte *bufx, size_t bufxlen, uint incx);
-        //: buf(bufx), pend(bufx + bufxlen), p(bufx), len(bufxlen), inc(incx), origbuf(bufx) { }
+    this(ubyte *bufx, size_t bufxlen, uint incx)
+    {
+        buf = bufx; pend = bufx + bufxlen; p = bufx; origbuf = bufx;
+    }
 
     //~this();
 
@@ -169,15 +169,15 @@ struct Outbuffer
      */
     void writeDouble(double v);
 
-    void write(const char *s);
+    void write(const(char)* s);
 
-    void write(const ubyte *s);
+    void write(const(ubyte)* s);
 
-    void writeString(const char *s);
+    void writeString(const(char)* s);
 
-    void prependBytes(const char *s);
+    void prependBytes(const(char)* s);
 
-    void prepend(const void *b, size_t len);
+    void prepend(const(void)* b, size_t len);
 
     void bracket(char c1,char c2);
 
--- a/src/ddmd/backend/outbuf.c
+++ b/src/ddmd/backend/outbuf.c
@@ -33,9 +33,17 @@ Outbuffer::Outbuffer()
     buf = NULL;
     pend = NULL;
     p = NULL;
-    len = 0;
-    inc = 0;
     origbuf = NULL;
+}
+
+Outbuffer::Outbuffer(d_size_t initialSize)
+{
+    buf = NULL;
+    pend = NULL;
+    p = NULL;
+    origbuf = NULL;
+
+    enlarge(initialSize);
 }
 
 Outbuffer::~Outbuffer()
@@ -59,22 +67,18 @@ void Outbuffer::reset()
 // Enlarge buffer size so there's at least nbytes available
 void Outbuffer::enlarge(d_size_t nbytes)
 {
-    d_size_t oldlen = len;
-    d_size_t used = p - buf;
+    const d_size_t oldlen = pend - buf;
+    const d_size_t used = p - buf;
 
-    if (inc > nbytes)
-    {
-        len = used + inc;
-    }
-    else
-    {
-        len = used + nbytes;
-        if (len < 2 * oldlen)
-        {   len = 2 * oldlen;
-            if (len < 8)
-                len = 8;
-        }
-    }
+    d_size_t len = used + nbytes;
+    if (len <= oldlen)
+        return;
+
+    const d_size_t newlen = oldlen + (oldlen >> 1);   // oldlen * 1.5
+    if (len < newlen)
+        len = newlen;
+    len = (len + 15) & ~15;
+
 #if MEM_DEBUG
     if (buf == origbuf)
     {
@@ -85,19 +89,14 @@ void Outbuffer::enlarge(d_size_t nbytes)
     else
         buf = (unsigned char *)mem_realloc(buf, len);
 #else
-    if (buf)
-    {
-        if (buf == origbuf)
-        {
-            buf = (unsigned char *) malloc(len);
-            if (buf)
-                memcpy(buf, origbuf, oldlen);
-        }
-        else
-            buf = (unsigned char *) realloc(buf,len);
-    }
-    else
-        buf = (unsigned char *) malloc(len);
+     if (buf == origbuf && origbuf)
+     {
+         buf = (unsigned char *) malloc(len);
+         if (buf)
+             memcpy(buf, origbuf, used);
+     }
+     else
+         buf = (unsigned char *) realloc(buf,len);
 #endif
     if (!buf)
     {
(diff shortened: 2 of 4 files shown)


src/ddmd/backend/outbuf.d
```diff
@@ -26,16 +26,16 @@ struct Outbuffer
...
     ubyte *pend;        // pointer past the end of the buffer
     ubyte *p;           // current position in buffer
-    uint len;           // size of buffer
-    uint inc;           // default increment size
     ubyte *origbuf;     // external buffer
 
     //this();
 
-    this(size_t incx); // : buf(null), pend(null), p(null), len(0), inc(incx), origbuf(null) { }
+    this(size_t incx); // : buf(null), pend(null), p(null), origbuf(null) { }
```
MartinNowak: Please rename incx to initsz or initialSize.

MartinNowak (approved):
Anything magic about 1.5?
For one it should nicely map onto malloc bucket sizes (though those tend to be more fine-grained).

ibuclaw:
> Anything magic about 1.5?

I think the right answer is "profile it!"

The couple general points are that it's a little less wasteful, and increases chances of memory being reusable in subsequent allocations. 

https://blog.mozilla.org/nnethercote/2014/11/04/please-grow-your-buffers-exponentially/#comment-23533

WalterBright:
> Anything magic about 1.5?

The idea is that if it is less than 1.6, then the next allocation can be filled by the concatenation of the previous three, leading to less memory commitment and cache friendliness. A factor of 2 means that a reallocation can never be satisfied by a combination of the previous allocations, meaning ever more cold memory is committed.

```
Available    Wanted
0          x
x           1.6x
x+1.6x   1.6*1.6x
```

1.5 is, of course, easier computed than 1.6.

I had encountered this before, but had forgotten it. Andrei reminded me.

All the realloc's in dmd and Phobos should be corrected.

src/ddmd/backend/outbuf.d
WalterBright: ok

MartinNowak:
Not sure I understand correctly what you're saying.
Is it that malloc is able to reuse the previous 3 now freed buffer sizes to satisfy the 4th allocation size?
Then you'd actually want to solve [`x⁴ = x³ + x² + x`](http://www.wolframalpha.com/input/?i=x%E2%81%B4+%3D+x%C2%B3+%2B+x%C2%B2+%2B+x) which is `0` or `~1.83929`.

But that seems weird, any allocator I've seen in recent years seems to use separate bins for different size classes, as that saves metadata. Otherwise you'd need to store the allocation size with each allocation.

It seems that non-standard extensions such as [malloc_usable_size(3)](http://man7.org/linux/man-pages/man3/malloc_usable_size.3.html) are more fruitful for optimizations.

Should also be fairly trivial to run a small realloc test at startup, comparing the returned pointers, to dynamically figure out what reallocation scheme is used.

WalterWaldron:
Minor comments on existing code found while reviewing:
```
     if (buf)  // <--- Redundant check, realloc(NULL, len) is equivalent to malloc(len)
     {
         if (buf == origbuf)
         {
             buf = (unsigned char *) malloc(len);
             if (buf)
                 memcpy(buf, origbuf, oldlen);  // <--- Why not only copy the used bytes?
         }
         else
             buf = (unsigned char *) realloc(buf,len);
     }
     else
         buf = (unsigned char *) malloc(len);
```

An example where only copying the ```used``` bytes might make a difference:
```
auto buf = Outbuffer(originalBuffer, originalLength);
// [...] do stuff with buf
buf.setSize(0); // Reset buf for reuse
// [...]
buf.reserve(newLength); // reuse buf, reserve for new use
```

> Is it that malloc is able to reuse the previous 3 now freed buffer sizes to satisfy the 4th allocation size?
Then you'd actually want to solve x⁴ = x³ + x² + x which is 0 or ~1.83929.

Typically the current buffer is excluded from the reuse analysis, leading to  x⁴ = x² + x + 1, which solves to ~1.47. Furthermore, solving for n block reuse leads to the [golden ratio](https://crntaylor.wordpress.com/2011/07/15/optimal-memory-reallocation-and-the-golden-ratio/).
Anyways, [most implementations](https://en.wikipedia.org/wiki/Dynamic_array#Growth_factor) simply use either 1.5 or 2 for reallocation factor.

MartinNowak:
[Benchmark realloc growth strategies](https://gist.github.com/MartinNowak/6f0bb8fe04410d12fdfa7ac8f7646563)

capacity_steps
grow_timings

Somewhat indifferent results (this is a non-zeroed plot).
Maybe it's an old story from the 90s?

If you compare those ldc values to gdc (~2.5x slower)
grow_timings

and dmd (~15x slower)
grow_timings
it seems that we have very different performance issues.

Might be worth a thought before churning phobos.

glibc-2.25
Core i7-6700
16GB DDR4 RAM

MartinNowak:
Increased the sample size for ldc to reduce the error interval.
Seems that at least for glibc's (2.25) malloc, 1.9 is the winner.
grow_timings

Not much useful to be gained by [malloc_usable_size(3)](http://man7.org/linux/man-pages/man3/malloc_usable_size.3.html).

WalterBright:
@WalterWaldron thanks for the link that explains it far better than I did.

@MartinNowak thanks for the benchmarks. They are helpful, but do not cover memory consumption. It's true that the results depend on the allocation strategy of the underlying `malloc`, which is not something we can rely on. But we do know that factors >1.6 make it not possible for the `malloc` to reuse the memory. A factor of 1.5 does seem to be moving towards being standard industry practice, which will also produce a tendency for `malloc` implementations to work better with it.

@WalterWaldron the code dates from a time when `realloc(NULL,ptr)` was not reliable with `NULL` arguments. I agree it is safe to rely on it these days.

ibuclaw:
@MartinNowak - I don't suppose you have a system with glibc 2.26 available? They've made a change in malloc that might alter the results.

WalterWaldron:
I've [modified Martin's benchmark](https://gist.github.com/WalterWaldron/da523795380000638e8cc1babea85308) to attempt to quantify realloc's data movement effect on the running time.

| Growth Ratio | Relocation Ratio | Total Relocation |
| --- | --- | --- |
| 1.1 | 1.17593 | 40841856 |
| 1.2 | 1.21249 | 42429152 |
| 1.3 | 1.22072 | 43885488 |
| 1.4 | 1.25937 | 47939376 |
| 1.5 | 0.960657 | 38311808 |
| 1.6 | **0.796346** | 30637184 |
| 1.7 | 0.826096 | 29804992 |
| 1.8 | 0.811198 | **29656784** |
| 1.9 | 1.21206 | 39149008 |
| 2 | 1.17058 | 38797440 |

With modified benchmark parameters, the timing looks  like this:
grow_timings
Major caveat of the parameters is that since the number of elements being appended increased, I had to significantly reduce the number of iterations. So the numbers should be taken lightly since I ran of time in this day.

LDC 1.4.0
glibc-2.26
Core i5-6500
16GB DDR4 RAM

WalterWaldron:
> I don't suppose you have a system with glibc 2.26 available? They've made a change in malloc that might alter the results.

grow_timings
LDC 1.4.0
glibc-2.26
Core i5-6500
16GB DDR4 RAM

MartinNowak:
> They are helpful, but do not cover memory consumption.

The first graph includes size growth (https://user-images.githubusercontent.com/288976/30816506-7023804a-a216-11e7-802a-389b84188301.png), though actual memory usage again depends on bin sizes of the allocator (fragmentation), and unless that mem is commited it's only cheap address space. The microbenchmark doesn't test cache performance.

> But that seems weird, any allocator I've seen in recent years seems to use separate bins for different size classes, as that saves metadata.

So my missing point here was that this is an optimization for large buffers which are usually not handled by separate size class bins, but are placed directly in page extends. Makes a lot more sense with that small hint :).

> @MartinNowak - I don't suppose you have a system with glibc 2.26 available? They've made a change in malloc that might alter the results.

Easy, indeed quite a difference and speedup.
2.2.5
grow_timings_2 2 5
2.2.6
grow_timings_2 2 6

Anyhow, let's get this merged, bechmarking dmd (compiling a debug libphobos2.a) yields a tiny 0.1-0.2% improvement.

ibuclaw:
> Easy, indeed quite a difference and speedup (maybe at the cost of memory usage?).

They added a per-thread cache.  https://sourceware.org/git/?p=glibc.git;a=blob;f=NEWS;h=1266276fcd50acd95175a7355a07a7252e150265;hb=HEAD#l10

ibuclaw:
> Anyhow, let's get this merged, bechmarking dmd (compiling a debug libphobos2.a) yields a tiny 0.1-0.2% improvement.

:+1:

MartinNowak:
> They added a per-thread cache. https://sourceware.org/git/?p=glibc.git;a=blob;f=NEWS;h=1266276fcd50acd95175a7355a07a7252e150265;hb=HEAD#l10

It's *2017*, one would have thought that had been added 5 years ago after TCMalloc came along.

PR #7168, opened 2017-09-25
allow to build with address and undefined sanitizers

MartinNowak:
- overhead to catch hard-to-debug issues should be fairly low
- applicable to g++/clang++ for C++ sources and also to gdc/ldc for D sources


MartinNowak:
Apparently sth. to be left for the Jenkins agents with newer c++ compilers.

ZombineDev:
LDC also has support for sanitizers (dk about GDC), so probably makes sense to enable them if we're building the front-end with such compiler.

PR #7169, opened 2017-09-26
Call vdeprecationSupplemental from deprecationSupplemental

ibuclaw:
Seems like it should do this to be consistent with `errorSupplemental` and `warningSupplemental`.

### DIFF
--- a/src/ddmd/errors.d
+++ b/src/ddmd/errors.d
@@ -99,7 +99,7 @@ extern (C++) void deprecationSupplemental(const ref Loc loc, const(char)* format
 {
     va_list ap;
     va_start(ap, format);
-    vdeprecation(loc, format, ap);
+    vdeprecationSupplemental(loc, format, ap);
     va_end(ap);
 }
 
--- a/test/fail_compilation/diag3438.d
+++ b/test/fail_compilation/diag3438.d
@@ -5,9 +5,9 @@ TEST_OUTPUT:
 fail_compilation/diag3438.d(16): Deprecation: constructor diag3438.F1.this all parameters have default arguments, but structs cannot have default constructors.
 fail_compilation/diag3438.d(17): Deprecation: constructor diag3438.F2.this all parameters have default arguments, but structs cannot have default constructors.
 fail_compilation/diag3438.d(20): Deprecation: constructor diag3438.F5.this @disable'd constructor cannot have default arguments for all parameters.
-fail_compilation/diag3438.d(20): Deprecation: Use @disable this(); if you want to disable default initialization.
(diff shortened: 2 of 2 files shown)


mathias-lang-sociomantic:
You have to change some tests it seems

ibuclaw:
I thought that might be the case.

ibuclaw:
@mathias-lang-sociomantic - you seem to have merged the commit in a weird way.

mathias-lang-sociomantic:
How weird ? It was a Rebase & Merge

ibuclaw:
There is no merge commit.

ibuclaw:
Next time, try using the auto-merge label.

ZombineDev:
@ibuclaw While in general I agree that using the `auto-merge` is a good practice, I don't see the problem with using ["Rebase and merge"](https://github.com/blog/2243-rebase-and-merge-pull-requests) as long as it keeps a reference to the pull request, which it does by default. We were using this, in addition to "Squash and merge" successfully for dlang-tour before we deployed dlang-bot there too.

wilzbach:
> before we deployed dlang-bot there too.

FWIW the bot supports the labels "auto-merge-squash" and "auto-merge-rebase"  too, they aren't available here, because unfortunately there has been a precedent in which the squashing was "abused" :/

>  I don't see the problem with using "Rebase and merge" as long as it keeps a reference to the pull request, which it does by default. We were using this, in addition to "Squash and merge" successfully for dlang-tour

Yup. Many projects I worked on even had this as the only allowed merging strategy.

mathias-lang-sociomantic:
> There is no merge commit.

Yep, that's the point of it actually. You should know that, because all D developpers around you are using this approach ;)

The point of not having a merge commit is to make the history more readable. A linear history is much easier to read than one polluted by merge commits. Compare ocean's history which use rebase approach:
2017-10-01-173040_1366x768_scrot
With dmd's:
2017-10-01-173346_1366x768_scrot

(Generated using `log --color --graph --pretty=format:'%Cred%h%Creset -%C(yellow)%d%Creset %s %Cgreen(%cr) %C(bold blue)<%an>%Creset' --abbrev-commit`)

Now if there's a technical issue or policy or strong preference from someone not to use rebase & merge I'll switch to merge approach of course, I was just surprised this could be an issue :)

wilzbach:
> Now if there's a technical issue or policy or strong preference from someone not to use rebase & merge I'll switch to merge approach of course

There such a policy (yet). In fact, e.g. we wanted to switch to squashing for a long time:

https://github.com/dlang-bots/dlang-bot/issues/64

PR #7171, opened 2017-09-27
Split TOKcatass into TOKcatass, TOKcatelemass, TOKcatdcharass

WalterBright:
This is a refactor, it makes the code significantly easier to follow, and exposed a couple harmless bugs in e2ir.d


ibuclaw:
Seems reasonable, just rolling the names around though (`TOKcatasselem` vs `TOKcatelemass`), it looks awkward, however I can't think of any better way of naming it.

mathias-lang-sociomantic:
Restart the failing Ci. Agree that rolling over the names as @ibuclaw suggested sounds clearer.
I assume that was prompted by the scope fixes ?

WalterBright:
> I assume that was prompted by the scope fixes ?

Yes. I can't work on the other cat bugzilla issue until this is pulled.

MartinNowak (approved):

PR #7172, opened 2017-09-27
disable binary diff check on osx-32 for now

MartinNowak:
- cannot reproduce the Travis-CI test failure on my local OSX installation 10.8
- disable the test until we can reproduce and fix the problem (or drop osx-32)


mathias-lang-sociomantic (approved):

wilzbach:
> or drop osx

Hasn't this been officially deprecated by Apple since five or so years?
Also note that Travis is reducing their free OSX offering: https://blog.travis-ci.com/2017-09-22-macos-update

jacob-carlborg:
"macOS High Sierra will be the last macOS release to support 32-bit apps without compromise"

"Mac app updates and existing apps must support 64-bit starting June 2018". This is for the Mac App Store.

Any new features that are not exactly the same on 32bit and 64bit will only be implemented for 64bit from my side (TLS, Objective-C integration and similar).

I kill it with 🔥.

[1] https://developer.apple.com/news/?id=06282017a

PR #5293, opened 2015-12-02
Issue 15384 - assignment is sometimes still accepted as a condition

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=15384
1. Add CommaExp.toBoolean to show error message.
2. Add checkAssignmentAsCondition and call it before semantic analysis to provide informative diagnostic.

PR #7174, opened 2017-09-27
Fix regression in condition statements caused by #5293

ibuclaw:
And add `CommaExp::toBoolean` to C++ headers.

If a comma expression was converted to boolean, the expression type was not updated.  This was a cause of an ICE in gdc backend when using opCast.

### DIFF
--- a/src/ddmd/expression.d
+++ b/src/ddmd/expression.d
@@ -7155,6 +7155,7 @@ extern (C++) final class CommaExp : BinExp
         if (ex2.op == TOKerror)
             return ex2;
         e2 = ex2;
+        type = e2.type;
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


WalterBright:
Is it possible to have a test case for this?

ibuclaw:
They are the same tests as #5293.

ibuclaw:
Other test files that were causing the compiler to ice were opover2, sdtor, test42, and xtest46.

Taking the first example:
https://github.com/dlang/dmd/blob/master/test/runnable/opover2.d#L464

The addition of CommaExp::toBoolean meant that the result type of the if condition now became  `struct S4913`instead of `bool`.  For some reason dmd backend allows non-scalar types to be used in a boolean context. Maybe this would have been caught if there was stronger type checking in dmd's backend.

I don't think other kinds of expression are affected as they explicitly set the result type as bool in the semantic pass (`&&`/`||`).

ibuclaw:
@klickverbot - did ldc run into a similar problem? As you've been using the ddmd frontend for a while iirc.

ibuclaw:
Don't give up! Just cc me in on the pr and I will take care of it.

PR #7178, opened 2017-10-03
fix issue 17791 - add __traits(deprecated, symbol) 

dunkyp:


MartinNowak:
Would it make sense to somehow make the deprecation message introspectable too?

ZombineDev:
We also need a way to obtain any compiler messages/errors, so I would prefer to leave this PR as it is and focus on a more general solution, like something along the lines of:
```D
struct S
{
    auto opDispatch(string member, T...)(T args)
    {
        static if (args.length % 42 < 13)
            static assert (0, "Solve the mystery!");

        //...
    }
}

enum string[] messages = __traits(captureStdErrOutput,
{
    S s;
    s.theAnswer(1, 2);
});

static assert (messages.canFind("Solve the mystery!"));
```

PetarKirov (approved):

wilzbach:
This should have been accompanied
- by a PR to dlang.org spec (https://dlang.org/spec/traits.html) (at least as long until we don't automatically generate it, see e.g. https://github.com/dlang/dlang.org/pull/1688)
- by a changelog entry (https://github.com/dlang/dmd/tree/master/changelog)

@dunkyp would you be so kind to submit the spec & changelog PR? Thanks!

wilzbach:
Btw has anyone considered the possibility to add this to the existing [`getFunctionAttributes` trait](https://dlang.org/spec/traits.html#getFunctionAttributes)?
After all, there's also no way to check for `@future` atm.

ZombineDev:
I also considered `getFunctionAttributes`, but the deprecated attribute can be applied to more kinds of symbols than just functions. If the trait was called `getAttributes` or something like that, then yes it would have been a better place add this functionality.

PR #3085, opened 2014-01-12
pure -> immutable fixes

dnadlinger:
https://d.puremagic.com/issues/show_bug.cgi?id=11503
https://d.puremagic.com/issues/show_bug.cgi?id=11909

The actual changes are in 3c0342e and ac4aed5, the rest are just cleanup commits.

PR #7179, opened 2017-10-03
fix Issue 17635 - [REG 2.066.0] cannot convert unique immutable(int)*…

WalterBright:
…* to immutable

Adds back in a condition deleted by https://github.com/dlang/dmd/pull/3085


WalterBright:
I'll leave this up for now, but it is the wrong fix, now that I have identified the source of the trouble https://issues.dlang.org/show_bug.cgi?id=17635

WalterBright:
Hey, it's working!

andralex (approved):

andralex:
noice

PR #7180, opened 2017-10-03
func: Clarify traverseIndirections()

dnadlinger:
This is the type-based alias analysis implementation used to
allow mutable -> immutable for pure factory functions.

No functional change intended.

(@WalterBright: As per discussion in #3085 – hope this helps.)

### DIFF
--- a/src/ddmd/func.d
+++ b/src/ddmd/func.d
@@ -2471,37 +2471,45 @@ extern (C++) Type getIndirection(Type t)
 }
 
 /**************************************
- * Returns true if memory reachable through a reference B to a value of type tb,
- * which has been constructed with a reference A to a value of type ta
- * available, can alias memory reachable from A based on the types involved
- * (either directly or via any number of indirections).
+ * Performs type-based alias analysis between a newly created value and a pre-
+ * existing memory reference:
  *
- * This relation is not symmetric in the two arguments. For example,
- * a const(int) reference can point to a pre-existing int, but not the other
+ * Assuming that a reference A to a value of type `ta` was available to the code
+ * that created a reference B to a value of type `tb`, it returns whether B
+ * might alias memory reachable from A based on the types involved (either
+ * directly or via any number of indirections in either A or B).
+ *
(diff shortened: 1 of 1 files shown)


src/ddmd/func.d
```diff
@@ -2471,37 +2471,45 @@ extern (C++) Type getIndirection(Type t)
...
- * This relation is not symmetric in the two arguments. For example,
- * a const(int) reference can point to a pre-existing int, but not the other
+ * Assuming that a reference A to a value of type `ta` was available to the code
+ * that created a reference B to a value of type `tb`, it returns whether B
+ * might alias memory reachable from A based on the types involved (either
+ * directly or via any number of indirections in either A or B).
+ *
+ * This relation is not symmetric in the two arguments. For example, a
+ * a `const(int)` reference can point to a pre-existing `int`, but not the other
  * way round.
```
WalterBright: Needs `params:` section.

src/ddmd/func.d
```diff
@@ -2471,37 +2471,45 @@ extern (C++) Type getIndirection(Type t)
...
  */
-private bool traverseIndirections(Type ta, Type tb, void* p = null, bool reversePass = false)
+private bool traverseIndirections(Type ta, Type tb)
+{
+    // To handle arbitrary levels of indirections in both parameters, we
+    // recursively descend into aggregate members/levels of indirection in both
+    // `ta` and `tb` while avoiding cycles. Start with the original types.
+    return traverseIndirectionsImpl(ta, tb, null, false);
+}
+private bool traverseIndirectionsImpl(Type ta, Type tb, void* p, bool reversePass)
```
WalterBright: This should be a function nested within `traverseIndirections`

WalterBright: BTW, making it a nested function also eliminates the casting of the `void*` parameter, which makes the function safe.

WalterBright:
@klickverbot I found the source of the trouble, though I am not sure how to fix it. Detailed in https://issues.dlang.org/show_bug.cgi?id=17635

PR #7181, opened 2017-10-04
Removed unexisting idgen.d from win32 makefile

jercaianu:


andralex:
@RazvanN7 @WalterBright what's the deal with this? I'll pull because all tests pass.

andralex (approved):

ZombineDev:
@andralex Since https://github.com/dlang/dmd/pull/6837 dmd identifier generation was transitioned from a build step to CTFE, so this PR is correct in removing `idgen.d`. I wonder however why the auto-tester didn't catch this in #6837...

PR #7182, opened 2017-10-04
fix 17868 - add pragma(crt_con-/destructor)

MartinNowak:
- allows to run con-/destructors before/after CRT startup/shutdown
- primary use-case is implementing modular startup in druntime itself

required for https://github.com/dlang/dmd/pull/6956

### DIFF
--- a/changelog/crt-constructor.dd
+++ b/changelog/crt-constructor.dd
@@ -1,0 +1,33 @@
+pragma(crt_constructor) and pragma(crt_destructor) were added
+
+This allows programs to run initialization code before or cleanup code after C
+main.
+
+In particular those pragmas can be used in $(TT -betterC) code as substitutes
+for $(LINK2 $(ROOT_DIR)spec/class.html#shared_static_constructors, `shared static this()`) and
+$(LINK2 $(ROOT_DIR)spec/class.html#shared_static_destructors, `shared static ~this()`).
+
+$(RED Note:) At the time of execution druntime is not initialized.
+
+$(RED Note:) The order in which constructors are executed is unspecified.
+
+---
+import core.stdc.stdio;
+
+pragma(crt_constructor)
+void init()
+{
+    puts("init");
+}
+
+pragma(crt_destructor)
+void fini()
+{
+    puts("fini");
+}
+
+extern(C) int main()
+{
+    puts("main");
+}
+---
--- a/src/ddmd/dsymbolsem.d
+++ b/src/ddmd/dsymbolsem.d
@@ -3048,6 +3048,12 @@ extern(C++) final class DsymbolSemanticVisitor : Visitor
                 }
             }
         }
+        else if (pd.ident == Id.crt_constructor || pd.ident == Id.crt_destructor)
+        {
+            if (pd.args && pd.args.dim != 0)
+                pd.error("takes no argument");
+            goto Ldecl;
+        }
         else if (global.params.ignoreUnsupportedPragmas)
         {
             if (global.params.verbose)
--- a/src/ddmd/backend/elfobj.c
+++ b/src/ddmd/backend/elfobj.c
@@ -1625,16 +1625,9 @@ void Obj::compiler()
  *              3:      compiler
  */
 
-void Obj::staticctor(Symbol *s,int dtor,int none)
+void Obj::staticctor(Symbol *s, int, int)
 {
-    // Static constructors and destructors
-    //dbg_printf("Obj::staticctor(%s) offset %x\n",s->Sident,s->Soffset);
-    //symbol_print(s);
-    const IDXSEC seg = s->Sseg =
-        ElfObj::getsegment(".ctors", NULL, SHT_PROGBITS, SHF_ALLOC|SHF_WRITE, 4);
-    const unsigned relinfo = I64 ? R_X86_64_64 : R_386_32;
-    const size_t sz = ElfObj::writerel(seg, SegData[seg]->SDoffset, relinfo, STI_TEXT, s->Soffset);
-    SegData[seg]->SDoffset += sz;
+    setModuleCtorDtor(s, true);
 }
 
 /**************************************
@@ -1647,14 +1640,7 @@ void Obj::staticctor(Symbol *s,int dtor,int none)
 
 void Obj::staticdtor(Symbol *s)
 {
-    //dbg_printf("Obj::staticdtor(%s) offset %x\n",s->Sident,s->Soffset);
-    //symbol_print(s);
-    // Why does this sequence differ from staticctor, looks like a bug?
-    const IDXSEC seg =
-        ElfObj::getsegment(".dtors", NULL, SHT_PROGBITS, SHF_ALLOC|SHF_WRITE, 4);
-    const unsigned relinfo = I64 ? R_X86_64_64 : R_386_32;
-    const size_t sz = ElfObj::writerel(seg, SegData[seg]->SDoffset, relinfo, s->Sxtrnnum, s->Soffset);
-    SegData[seg]->SDoffset += sz;
+    setModuleCtorDtor(s, false);
 }
 
 /***************************************
@@ -1662,9 +1648,12 @@ void Obj::staticdtor(Symbol *s)
  * Used for static ctor and dtor lists.
  */
 
-void Obj::setModuleCtorDtor(Symbol *s, bool isCtor)
+void Obj::setModuleCtorDtor(Symbol *sfunc, bool isCtor)
 {
-    //dbg_printf("Obj::setModuleCtorDtor(%s) \n",s->Sident);
+    IDXSEC seg = ElfObj::getsegment(isCtor ? ".ctors" : ".dtors", NULL, SHT_PROGBITS, SHF_ALLOC|SHF_WRITE, NPTRSIZE);
+    const unsigned reltype = I64 ? R_X86_64_64 : R_386_32;
+    const size_t sz = ElfObj::writerel(seg, SegData[seg]->SDoffset, reltype, sfunc->Sxtrnnum, 0);
+    SegData[seg]->SDoffset += sz;
 }
 
 
--- a/src/ddmd/backend/machobj.c
+++ b/src/ddmd/backend/machobj.c
@@ -1662,24 +1662,9 @@ void Obj::compiler()
  *              3:      compiler
  */
 
-void Obj::staticctor(Symbol *s,int dtor,int none)
+void Obj::staticctor(Symbol *s, int, int)
 {
-#if 0
-    IDXSEC seg;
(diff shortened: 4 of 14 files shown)


andralex:
I'm okay with this approach as opposed to clever games with `static shared this()`. @WalterBright ?

MartinNowak:
This tool is useful despite `shared static this()` and `-betterC`, but indeed, once it's there, we don't necessarily need to alter `shared static this()`.

changelog/crt-constructor.dd
```diff
@@ -0,0 +1,33 @@
...
+
+In particular those pragmas can be used in $(TT -betterC) code as substitutes
+for $(LINK2 $(ROOT_DIR)spec/class.html#shared_static_constructors, `shared static this()`) and
+$(LINK2 $(ROOT_DIR)spec/class.html#shared_static_destructors, `shared static ~this()`).
+
+$(RED Note:) At the time of execution druntime is not initialized.
+
+$(RED Note:) The order in which constructors are executed is unspecified.
+
+----
```
jacob-carlborg: Shouldn't this be three dashes? Or is it three or more?

jacob-carlborg:
Perhaps a test that attaches the pragam to something else than a function.

changelog/crt-constructor.dd
MartinNowak: 3 or more works

MartinNowak:
> Perhaps a test that attaches the pragam to something else than a function.

Done

jacob-carlborg:
What about static methods in a class or struct?

MartinNowak:
> What about static methods in a class or struct?

Works, this is a low-level tool (just as `pragma(startaddress)`), it will just stuff any pointer into the init/fini section, no matter what ABI or params.

jacob-carlborg:
It would be nice with a test for that.

MartinNowak:
> It would be nice with a test for that.

It's too low-level, I wouldn't want to promise that for all ABIs, use at your own risk.

andralex:
>This tool is useful despite shared static this() and -betterC, but indeed, once it's there, we don't necessarily need to alter shared static this().

A short list of pros and cons of the approaches woudl be very helpful.

MartinNowak:
*This* PR is *not* about `-betterC`, let's keep that discussion in #6956.

MartinNowak:
Ping @WalterBright, @andralex. Anything preventing us from adding that feature for low-level runtime hackers and C++ interop?

andralex:
@WalterBright I'm good to go with this, will approve.

andralex (approved):

src/ddmd/dsymbolsem.d
```diff
@@ -3048,6 +3048,12 @@ extern(C++) final class DsymbolSemanticVisitor : Visitor
                 }
             }
         }
+        else if (pd.ident == Id.crt_constructor || pd.ident == Id.crt_destructor)
+        {
+            if (pd.args && pd.args.dim != 0)
+                pd.error("too many arguments");
```
WalterBright: `too many` is misleading. Should be `arguments are not allowed for crt_constructor and crt_destructor functions`.

WalterBright: These functions should also be forced to use C linkage.

WalterBright:
Thank you @MartinNowak for doing this work, especially the fixing of the `*obj.c` files.

My complaint centers around using an awkward pragma to do the dirty work. It has the effect of putting every function in its scope into the initializer list, which is a bit bizarre as I can't imagine code needing that. A better placement for the pragma would be inside the function body, which would avoid the need for the `recurse()` function. The `recurse()` function provokes questions we shouldn't need to answer (like why not affect the static member functions of an aggregate?).

The complaints around `shared static this()` are that it changes the semantics. But -betterC is expected to change the semantics to not use druntime, and `shared static this()` requires druntime. So it is ok to use it for `betterC` to do the construction/destruction. -betterC changes semantics on other bits of code, too. It's not going to break code.

Another issue is if one is not using -betterC, but wishes to install a function into the C ctor/dtor list. One way is for the user to put such code into a separate file, compile that file with -betterC, and link it in. Another way is to do this:
```
extern (C) void init() { ... }
pragma(crt_constructor, init);

extern (C) void fini() { ... }
pragma(crt_destructor, fini);
```
This also avoids the need for the `recurse()` function.

MartinNowak:
That's mostly because it's a function attribute (pragma) everywhere else, e.g. gdc/ldc/GCC/clang.
https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html
https://wiki.dlang.org/LDC-specific_language_changes#LDC_global_crt_ctor_and_LDC_global_crt_dtor

The recursion was added to deal with things like the following and maybe template functions.
```d
pragma(crt_constructor)
version (linux) void init() {}
```
I could instead change it to check that it directly applies to a function.

andralex:
Do we have a precedent of a pragma only applicable to one declaration? If not, would an attribute a la `@__crt_constructor` be more appropriate?

WalterBright:
Is it really a function attribute? It doesn't affect the function signature or ABI. Given the salad of attributes we have, adding more kinda scares me, especially for one so rarely used as this one. Recall how C++'s `volatile`, rarely used, wound up with an enormously outsized impact on the spec. Whatever we pick should have an impact on the spec in proportion to its use cases.

> The recursion was added to deal with things like the following and maybe template functions.

I don't find the recursive example compelling enough to have to explain why it doesn't affect aggregate member functions. Let's pick something that will be easy to document.

jpf91:
> Is it really a function attribute? It doesn't affect the function signature or ABI.

Why do you think a function attribute does have to affect the signature or API? We also have UDAs and these are simply defined as _compile time only meta-data for certain declarations_. Recognize some special cases in the compiler and they are a perfect fit for such situations. LDC and GDC already do this for years:
https://wiki.dlang.org/LDC-specific_language_changes#Attributes
https://wiki.dlang.org/Using_GDC#Attributes
There's also precedent in GCC using function attributes which do not affect ABI or signature instead of introducing pragmas for these cases: https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#Common-Function-Attributes

The mangle pragma is also a perfect fit for such an UDA. I really don't understand why DMD doesn't just finally follow the other compilers lead here to introduce a `dmd.attribute` module:

```d
// If you don't want the import in user code, add a public import to object.d instead
import dmd.attribute;

@crt_constructor void foo() {}
@crt_destructor @mangle("some_mangled_name") void bar() {}
```

You even get compile-time reflection to detect whether any such attribute is applied to a function for free, there's no global namespace pollution, and there are many more benefits. We have this nice UDA feature, why not use it?

jacob-carlborg:
> I really don't understand why DMD doesn't just finally follow the other compilers lead here to introduce a dmd.attribute module:

We already have `core.attribute` [1]. But of course, not that many knows about it because currently there's only one attribute in that module, `@selector`, which is used for Objective-C bindings.

[1] https://github.com/dlang/druntime/blob/master/src/core/attribute.d

MartinNowak:
Can we please keep the attribute discussion off here, it's a necessary, but separate discussion.

So far similar functionality was implemented as pragmas, and there isn't even functionality to recognize special attributes in dmd yet. It's also a pragma declaration in LDC.
Reflection for crt_con-/destructor, weak, or section doesn't seem necessary either.

BTW, that's the second time this review takes a detour into an slightly OT discussion.

> Another issue is if one is not using -betterC, but wishes to install a function into the C ctor/dtor list. One way is for the user to put such code into a separate file, compile that file with -betterC, and link it in.

That's hardly feasible as it required changing build tools to compile one module with different flags. It also changes the other semantics of that module.

MartinNowak:
> I don't find the recursive example compelling enough to have to explain why it doesn't affect aggregate member functions. Let's pick something that will be easy to document.

This PR has already been updated to handle this in the same way as `pragma(mangle)`.

thewilsonator:
>My complaint centers around using an awkward pragma to do the dirty work

FWIW in LDC we have 
`pragma(LDC_global_crt_[c|d]tor [, priority]) { funcdecl(s) }` where `priority` is a `ushort`.

MartinNowak:
> FWIW in LDC we have
> pragma(LDC_global_crt_[c|d]tor [, priority]) { funcdecl(s) } where priority is a ushort.

Yes, as mentioned in https://github.com/dlang/dmd/pull/7182#issuecomment-338410538.
After some initial attempts, priority was left out, as it cannot be properly supported on all platforms.

Can we please settle the bike-shedding on pragma vs. attributes vs. pragma with function name. This PR is implemented just like pragma mangle, and now doesn't allow more than one function.

thewilsonator:
> After some initial attempts, priority was left out, as it cannot be properly supported on all platforms.

Does it need to be? it would be nice to drop LDC's specific ones and use the dmd ones (LDCs is also a block pragma but that is less of an issue). Having said that I'm not sure how widely they are used. CRT constructors are rather rare and specifying a priority is even more rare.

Perhaps you could make the priority optional and implementation defined?

MartinNowak:
No, we went through this, LDC/GDC can keep their pragmas with priority support and map the one added in *this* PR to the default priority. As you said there is hardly any use-case for priorities to begin with, and even with ldc/gdc priorities were not global, but local (translation unit).

thewilsonator:
Fair enough.

andralex (approved):

jacob-carlborg:
@MartinNowak now when this is merged, how does this related to the `obj_rtinit` [1] function in the compiler? Can that function be completely removed, or parts of it?

[1] https://github.com/dlang/dmd/blob/901fde25bdf5c31a83f2d490b81bd2365cd1736f/src/ddmd/backend/elfobj.c#L3372

changloong:
crt_constructor is not allow to set static immutable.

```d
static immutable int my_i ;
int test(){
	return 3;
}

shared static this(){
	my_i = test();  // work 
}
pragma(crt_constructor) extern(C) void init1(){
	my_i = test(); //   Error: cannot modify immutable expression my_i
}
```

ibuclaw:
You could raise a bug on that instead of commenting here ( issues.dlang.org )

ibuclaw:
@jacob-carlborg - I will assume no, because it needs to be emitted into every DSO.

changloong:
@ibuclaw thanks for tip
add bug: https://issues.dlang.org/show_bug.cgi?id=18100

9il:
Thanks for this is killer feature for betteC programs.

Tested with mir-cpuid, DMD segfaults if `-lib` flag is passed
https://issues.dlang.org/show_bug.cgi?id=18456

Geod24:
Bit late, but specs PR is here: https://github.com/dlang/dlang.org/pull/2757

PR #7183, opened 2017-10-05
also get coverage from C++ sources

MartinNowak:


wilzbach:
Seems like gcov doesn't like this yet:

```
==> Running gcov in . (disable via -X gcov)
optabgen.gcno:cannot open graph file
newdelete.gcno:cannot open graph file
go.gcno:cannot open graph file
gdag.gcno:cannot open graph file
gother.gcno:cannot open graph file
gflow.gcno:cannot open graph file
gloop.gcno:cannot open graph file
gsroa.gcno:cannot open graph file
```

WalterBright:
I'm glad this is possible! Thanks, @MartinNowak

MartinNowak:
Works now, coverage is dropping quite a bit.

PR #7185, opened 2017-10-06
fix invalid stack reads

MartinNowak:
- e.g. with IDXSTR (4 bytes) and 8 byte aligned keysize

https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fci/detail/PR-78/8/pipeline#step-88-log-380


WalterBright:
Address sanitizer pays off!

WalterBright (approved):

PR #7186, opened 2017-10-06
refactor isReturnIsolated()

WalterBright:
Do a refactoring of `isolatedReturn()` so the names and logic makes more sense. Inverted the boolean logic of `traverseIndirections()` so it was consistent with `isolatedReturn()`.


andralex (approved):
Nice. I'd prefer this taken a step further a la `isTypeIsolatedIndirect(t)` -> `t.isTypeIsolatedIndirect` (UFCS, no straggling parens).

PR #7188, opened 2017-10-07
fix Issue 15538 - [REG 2.064] wrong code with switch

WalterBright:


src/ddmd/backend/cod3.c
```diff
@@ -1463,15 +1464,16 @@ void doswitch(CodeBuilder& cdb, block *b)
...
+            if (I64 && sz == 8)
                 code_orrex(cdb.last(), REX_W);
             genjmp(cdb,JA,FLblock,b->nthSucc(0));  // JA default
         }
         if (I64)
         {
             if (!vmin)
             {   // Need to clear out high 32 bits of reg
-                genmovreg(cdb,reg,reg);                       // MOV reg,reg
+                // Use 8B instead of 89, as 89 will be optimized away as a NOP
```
rainers: Doesn't this expose a bug in the peephole optimizations?

WalterBright: No, it's designed to work that way. The peepholer is for what the compiler generates. And if it inadvertently changes, that's why there's a test suite for this case :-)

rainers: Ok. But then this (https://github.com/dlang/dmd/blob/master/src/ddmd/backend/cod4.c#L2887) becomes a nop, too?

rainers:
Better applied to stable?

rainers (approved):

src/ddmd/backend/cod3.c
WalterBright: https://github.com/dlang/dmd/pull/7192

PetarKirov (approved):

PR #7189, opened 2017-10-07
 fix issue 17619:  [REG2.072] Wrong debug line information with single line loops

rainers:
for statements without curly braces, default endloc to location of last token, not next token.


ibuclaw (commented):
Looks reasonable, I don't suppose you could create a test case?

rainers:
> I don't suppose you could create a test case?

I've added a test mimicking what I did while bisecting: use `objdump -Wl test.o` to verify that there are not two line numbers attached to a single offset.

PR #7190, opened 2017-10-07
fix -fsanitize linking with dmd

MartinNowak:
- by supporting link driver arguments in HOST_CXX variable
- The inability to pass gcc/clang arguments for linking has popped up
  a couple of times. This is a simple change that simplifies using
  features such as -static, -flto, or -fsanitize=address.


WalterBright:
I guarantee that in the future there will be a new bug report that quoted strings with spaces in them won't work. strtok is too simplistic. But this is good enough for now.

PR #7191, opened 2017-10-07
Fix up C++ headers

kinke:
Resulting from upgrading LDC to v2.076.0. I only checked the diffs from 2.075.1 which are relevant for LDC (i.e., only a subset of the .d files); that was very tedious already, and I cannot guarantee that I caught all
spots where the D implementation and the C++ headers diverged.

Pinging @ibuclaw.

### DIFF
--- a/src/ddmd/attrib.h
+++ b/src/ddmd/attrib.h
@@ -24,6 +24,7 @@ class LabelDsymbol;
 class Initializer;
 class Module;
 class Condition;
+class StaticForeach;
 
 /**************************************************************/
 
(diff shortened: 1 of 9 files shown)


WalterBright (approved):

ibuclaw:
Let me test this.

kinke:
[I double-checked it earlier against current `stable`.]

ibuclaw (approved):
Looks fine compiling against the D implementation. Compiling against the C++ implementation there's a few undefined references, but that's expected.

PR #7192, opened 2017-10-07
use 0x8B instead of 0x89 so it isn't optimized away

WalterBright:
@rainers discovered this. No known test case. The peephole optimizer removes `MOV reg,reg` when the 0x89 op is used.


PetarKirov (approved):

rainers:
Grepping the source I found another one: https://github.com/dlang/dmd/blob/master/src/ddmd/backend/cgxmm.c#L407

WalterBright:
> another one

Thanks, taken care of.

andralex (approved):
subtle

PR #7196, opened 2017-10-08
fix Issue 16011 - [REG2.068] recursive RefCounted used to work

WalterBright:


WalterBright:
Closing because it passes the test case without needing this fix.

PR #5695, opened 2016-04-23
Issue 11169 - __traits(isAbstractClass) prematurely sets a class to be abstract

9rnsr:
`__traits(isAbstractClass)` should resolve forward references of all class member functions, in order to return stable result.

PR #6659, opened 2017-03-29
Fix Issue 16273 - [REG2.072a] dmd segfault with inheritance, templates...

tramker:
It segfaults, because vtbl is empty at that stage. The two test cases (func1 and func2) are slightly different. FuncDeclaration.semantic() is called from ClassDeclaration.isAbstract() and from MixinTemplate.semantic().

The segfault with mixin was there even before this regression, this fixes both.

PR #7198, opened 2017-10-08
fix Issue 16273 - [REG 2.072a] dmd segfault with inheritance, templat…

WalterBright:
…es, override

Fixes https://github.com/dlang/dmd/pull/5695 and replaces https://github.com/dlang/dmd/pull/6659

The fix here is to move just enough of `semantic()` into `isAbstract()` to answer the question.

### DIFF
--- a/src/ddmd/dclass.d
+++ b/src/ddmd/dclass.d
@@ -12,6 +12,7 @@ module ddmd.dclass;
 
 // Online documentation: https://dlang.org/phobos/ddmd_dclass.html
 
+import core.stdc.stdio;
 import core.stdc.string;
 
 import ddmd.aggregate;
@@ -827,8 +828,20 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
      */
     final bool isAbstract()
     {
+        enum log = false;
         if (isabstract != ABSfwdref)
             return isabstract == ABSyes;
+
+        if (log) printf("isAbstract(%s)\n", toChars());
+
+        bool no()  { if (log) printf("no\n");  isabstract = ABSno;  return false; }
+        bool yes() { if (log) printf("yes\n"); isabstract = ABSyes; return true;  }
+
+        if (storage_class & STCabstract || _scope && _scope.stc & STCabstract)
+            return yes();
+
+        if (errors)
+            return no();
 
         /* https://issues.dlang.org/show_bug.cgi?id=11169
          * Resolve forward references to all class member functions,
@@ -842,9 +855,6 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
             if (fd.storage_class & STCstatic)
                 return 0;
 
-            if (fd._scope)
-                fd.semantic(null);
-
             if (fd.isAbstract())
                 return 1;
             return 0;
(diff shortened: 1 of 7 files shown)


src/ddmd/declaration.d
```diff
@@ -265,7 +265,7 @@ extern (C++) abstract class Declaration : Dsymbol
         return (storage_class & STCfinal) != 0;
     }
 
-    final bool isAbstract()
+    bool isAbstract()
```
ibuclaw: Need to mark this as virtual in C++ headers.

ibuclaw (changes requested):
Apart from one additional change, LGTM.

src/ddmd/func.d
```diff
@@ -1108,6 +1108,29 @@ extern (C++) class FuncDeclaration : Declaration
         return true; // functions can be overloaded
     }
 
+    /***********************************
+     * Override so it can work even if semantic() hasn't yet
+     * been run.
+     */
+    override final bool isAbstract()
```
ibuclaw: And add this also to the headers.

src/ddmd/declaration.d
WalterBright: done

src/ddmd/func.d
WalterBright: done

PetarKirov (approved):
Looks reasonable to me.

ibuclaw (approved):

ibuclaw:
However looks like a test has regressed.
```
... compilable/test11169.d         -o- ()
../src/dmd -conf= -m32 -Icompilable -o-  -odgenerated/compilable -ofgenerated/compilable/test11169_0.o -c compilable/test11169.d
1: true
2: true
3: true
compilable/test11169.d(42): Error: static assert  `__traits(compiles, ()
{
_error_ b = __error__;
}
)` is false
```

WalterBright:
@ibuclaw working now

src/ddmd/dclass.d
```diff
@@ -827,9 +828,21 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
      */
     final bool isAbstract()
     {
+        enum log = false;
         if (isabstract != ABSfwdref)
             return isabstract == ABSyes;
 
+        if (log) printf("isAbstract(%s)\n", toChars());
```
ibuclaw: Shouldn't this be in `version(none)` or part of some global level `LOG` mechanism?

Perhaps a new bugzilla issue is better place to discuss thoughts though. As debug tracing the compiler internals in a consistent manner would probably require a new logging interface.

WalterBright: I've used various logging schemes,  and I like this one the best. Global schemes produce a blizzard of noise and the signal gets lost. This one is very local, and since `log` is an enum it gets optimized away to nothing when `false`.

ibuclaw: Each module could define its own subsystem, then you turn them on via the command line.

John-Colvin:
This caused a regression: https://issues.dlang.org/show_bug.cgi?id=17955

src/ddmd/dclass.d
```diff
@@ -855,26 +865,59 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
...
-                return true;
+                return yes();
+            }
+        }
+
+        /* If the base class is not abstract, then this class cannot
+         * be abstract.
+         */
+        if (!baseClass || !baseClass.isAbstract())
+            return no();
```
ibuclaw: This causes another regression.  This routine now returns `false` if called with an interface.

e.g:
```
interface A { void foo(); }
interface AB : A {}
```
GDC now fails to compile this simply because it calls `fillVtbl` during the codegen stages, which results in the error `interface function is not implemented`.

MartinNowak:
REGRESSION FIXES SHOULD BE MERGED INTO STABLE!!!

This slipped into 2.077.0 but the fixup still isn't in stable.

PR #7199, opened 2017-10-08
Fix Issue 17307 - [REG2.072.0][ICE] TypeBasic::implicitConvTo

JinShil:
This fixes the ICE, but there may still be an underlying problem with the semantics.

Output for...
```
struct { enum bitsPerWord = size_t; }

void main()
{ }
```
... is now...
```
main.d(1): Error: anonymous struct can only be a part of an aggregate, not module main
main.d(1): Error: cannot interpret ulong at compile time
```
I'd be happy to add a test case, but I first need confirmation that that is the expected output.

### DIFF
--- a/src/ddmd/attrib.d
+++ b/src/ddmd/attrib.d
@@ -55,6 +55,9 @@ extern (C++) abstract class AttribDeclaration : Dsymbol
 
     Dsymbols* include(Scope* sc, ScopeDsymbol sds)
     {
+        if (errors)
+            return null;
+
         return decl;
     }
 
@@ -802,6 +805,7 @@ extern (C++) final class AnonDeclaration : AttribDeclaration
         if (!ad)
         {
             .error(loc, "%s can only be a part of an aggregate, not %s `%s`", kind(), p.kind(), p.toChars());
+            errors = true;
             return;
         }
 
@@ -1241,6 +1245,10 @@ extern (C++) class ConditionalDeclaration : AttribDeclaration
     override Dsymbols* include(Scope* sc, ScopeDsymbol sds)
     {
         //printf("ConditionalDeclaration::include(sc = %p) scope = %p\n", sc, scope);
+
+        if (errors)
+            return null;
+
         assert(condition);
         return condition.include(_scope ? _scope : sc, sds) ? decl : elsedecl;
     }
@@ -1317,6 +1325,10 @@ extern (C++) final class StaticIfDeclaration : ConditionalDeclaration
     override Dsymbols* include(Scope* sc, ScopeDsymbol sds)
     {
         //printf("StaticIfDeclaration::include(sc = %p) scope = %p\n", sc, scope);
+
+        if (errors)
+            return null;
+
         if (condition.inc == 0)
         {
             assert(scopesym); // addMember is already done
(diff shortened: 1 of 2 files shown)


ZombineDev:
The error message sounds correct to me, as D's grammar doesn't support neither inline typedefs, nor inline struct variables.

rainers:
I think this is fighting the symptoms, but a type pointer `null` will cause a lot of trouble elsewhere, too. I think a better fix might be to not stop semantic analysis if "anonymous struct can only be a part of an aggregate" has been detected, i.e. remove the `return` here: https://github.com/dlang/dmd/blob/stable/src/ddmd/attrib.d#L805

rainers:
You should also add the test somwhere to the fail_compilation test suite.

JinShil:
> I think a better fix might be to not stop semantic analysis if "anonymous struct can only be a part of an aggregate" has been detected, i.e. remove the `return` here: https://github.com/dlang/dmd/blob/stable/src/ddmd/attrib.d#L805

Done

> You should also add the test somwhere to the fail_compilation test suite.

Done

rainers (approved):
Thanks.

It's a bit annoying that code coverage marks this PR as bad as it just removes a line. The percentage drops indeed, but the number of uncovered lines stays the same.

WalterBright:
The theory of dealing with compilation errors is not to attempt to recover from them, but to 'poison' anything that depends on an AST node that has an error in it. This is the rationale of `TypeError` and similar in the compiler. Recovery attempts (i.e. guess what the user intended and attempt to repair the error) rarely work out well, and result in cascaded misleading error messages.

This PR is trying to recover from the error by putting the erroneous `AnonDeclaration` back into a known state, which later passes depend on.

A fix more in keeping with the error design in the compiler is to 'poison' the `AnonDeclaration` node by setting `AnonDeclaration.errors`. Then, when the node is used, `errors` is checked and the error is propagated instead. Making this work in the most effective manner is to put the `errors` check inside the various `include()` member functions, having them return `null` if `errors` is set.

WalterBright (changes requested):
Set `errors` instead, per my other comment.

WalterBright:
@JinShil when submitting a bug fix PR, please note in bugzilla the PR url. This helps prevent someone else from wasting their time preparing a fix for an issue if there already is an outstanding PR. Thanks!

(I already took care of it for this one.)

JinShil:
> A fix more in keeping with the error design in the compiler is to 'poison' the `AnonDeclaration` node by setting `AnonDeclaration.errors`. Then, when the node is used, `errors` is checked and the error is propagated instead. Making this work in the most effective manner is to put the `errors` check inside the various `include()` member functions, having them return `null` if errors is set.

Done.  I hope my changes are what you intended.

src/ddmd/attrib.d
```diff
@@ -55,6 +55,9 @@ extern (C++) abstract class AttribDeclaration : Dsymbol
 
     Dsymbols* include(Scope* sc, ScopeDsymbol sds)
```
WalterBright: Also need to do the overrides of `include()`.

JinShil: Done

WalterBright:
Good! Well done.

PR #7200, opened 2017-10-08
refactor: Add setError to ExpressionSemanticVisitor

ibuclaw:
Easy win on the line count.  A helper of the same name is present in `StatementSemanticVisitor` also.

### DIFF
--- a/src/ddmd/expressionsem.d
+++ b/src/ddmd/expressionsem.d
@@ -78,6 +78,11 @@ extern (C++) final class ExpressionSemanticVisitor : Visitor
         this.sc = sc;
     }
 
+    private void setError()
+    {
+        result = new ErrorExp();
+    }
+
     /**************************
      * Semantically analyze Expression.
      * Determine types, fold constants, etc.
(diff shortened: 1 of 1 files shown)


rainers (approved):

src/ddmd/expressionsem.d
```diff
@@ -4769,30 +4417,25 @@ extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+            return setError();
 
         Type t1b = exp.e1.type.toBasetype();
         if (t1b.ty == Tpointer)
         {
             if ((cast(TypePointer)t1b).next.ty == Tfunction)
             {
                 exp.error("cannot slice function pointer %s", exp.e1.toChars());
-                result = new ErrorExp();
+                return setError();
```
rainers: There was no `return` here before, but probably a bug.
(not sure my first comment got trough, I received an error from github.)

ibuclaw: Just checked the commit, it is indeed a bug introduced by the conversion to visitor.

PR #7202, opened 2017-10-09
Add Add AsmJS, Emscripten and WebAssembly version identifiers

PetarKirov:
WebAssembly (http://webassembly.org/) is a is a low-level binary format
for in-browser client-side scripting (though support for non-browser
execution environments is also planned) and also a JavaScript API. See
https://github.com/WebAssembly/design/blob/master/Portability.md
for more information on the characteristics of this virtual ISA.
All major browser vendors have reached consensus on the WebAssembly API
and binary format, so growth opportunity for D in this area is huge.

The current feature set and toolchain is based on asm.js
(http://asmjs.org/spec/latest/) and Emscripten
(http://kripken.github.io/emscripten-site/).
Emscripten is a toolchain for converting  LLVM IR to JavaScript/asm.js.
Emscripten uses the musl (http://www.musl-libc.org/) libc library,
which potentially requires improving the `version (CRuntime_Musl)`
support in druntime and phobos.

In a couple of years asm.js may be phased-out in favour of WebAssembly,
but for now it is necessary as WebAssembly is not mature enough.

CC @Ace17 @cosinus2 @dhasenan @TurkeyMan I saw you also have interest in this area.


ZombineDev:
For now I'm only adding `WASM32`, as while 64-bit WebAssembly support is planned, only 32-bit exista at the moment.

mathias-lang-sociomantic:
It's missing a test, a changelog entry and the companion dlang.org P.R.

ZombineDev:
Yes, I am aware, will amend the PR and open one for the spec later today.

ZombineDev:
About the two added versions - `Emscripten` and `WASM32` - presently the later implies the former, but if you look at documentation of the respective projects they don't mention the other almost nowhere. That's because Emscripten was built to support primary asm.js and now there's a converter tool between asm.js and WebAssembly (which the Emscripten SDK now integrates). And on the other hand WebAssembly byte-code or AST could be generated directly, which is why I added both as they're conceptually orthogonal.

ZombineDev:
@mathias-lang-sociomantic done - I added tests and opened a PR against dlang.org: https://github.com/dlang/dlang.org/pull/1905.

test/fail_compilation/reserved_version.d
```diff
@@ -2,101 +2,103 @@
...
+fail_compilation/reserved_version.d(194): Error: version identifier `D_LP64` is reserved and cannot be set
+fail_compilation/reserved_version.d(195): Error: version identifier `D_X32` is reserved and cannot be set
+fail_compilation/reserved_version.d(196): Error: version identifier `D_HardFloat` is reserved and cannot be set
+fail_compilation/reserved_version.d(197): Error: version identifier `D_SoftFloat` is reserved and cannot be set
+fail_compilation/reserved_version.d(198): Error: version identifier `D_PIC` is reserved and cannot be set
+fail_compilation/reserved_version.d(199): Error: version identifier `D_SIMD` is reserved and cannot be set
+fail_compilation/reserved_version.d(200): Error: version identifier `D_Version2` is reserved and cannot be set
+fail_compilation/reserved_version.d(201): Error: version identifier `D_NoBoundsChecks` is reserved and cannot be set
+fail_compilation/reserved_version.d(204): Error: version identifier `all` is reserved and cannot be set
+fail_compilation/reserved_version.d(205): Error: version identifier `none` is reserved and cannot be set
```
mathias-lang-sociomantic: That's a large diff for little value. Just put a `#line` directive down here.

ZombineDev: I would still need to change line-numbers after Emscripten and WASM32. Perhaps it would be better to add them at the end?

mathias-lang-sociomantic: Yep at the end sounds good

wilzbach (approved):

ZombineDev:
@mathias-lang-sociomantic done

mathias-lang-sociomantic:
Red tho

ZombineDev:
@mathias-lang-sociomantic oops, forgot to commit the `#line` thing, should be fixed now.

mathias-lang-sociomantic (approved):

WalterBright:
I did some googling and it seems `wasm32` is in common use, and I can't find mention of `WASM32`. The all caps thing seems unusual.

http://webassembly.org/

 How about `Wasm32`, like `Win32` ? Or even go further and `WebAssembly32` ? That would certainly make it google-able.

Also, we don't have `linux32` and `linux64`, instead it's `linux` and `X86_64`. I expect wasm 64 to be extremely similar to wasm 32. Perhaps `WebAssembly` and something else to give the memory model?

I'm also not sure what the point of adding `Emscripten` is, why does it need to be a version?

And as you say there's `AsmJS`. 

gcc has 239 predefined macros just on linux, including 4 variants of `linux`. I'm wondering if we should develop some code for WebAssembly to be sure we're predefining the right identifiers?

ZombineDev:
I first used `WebAssembly32` as it seemed the most logical to me at first, but then I changed it to `WASM32`, since `wasm32` is in common use, but in all caps to mimic the rest of the architecture version identifiers.

ZombineDev:
> gcc has 239 predefined macros just on linux, including 4 variants of linux. 

Yes, you've mention this before. I also don't us end-up in such situation (where no one know how versions are there, what they mean and which will be predefined), but at the same time pragmatism dictates that we should be able to write portable code and should be able to introspect the features of platform we're targeting.

> I'm wondering if we should develop some code for WebAssembly to be sure we're predefining the right identifiers?

I'm not sure I understand.

As noted earlier there are WebAssembly, Emscripten and AsmJS. In principle, they're all orthogonal (you **can** have one without the others), but in the short-term if you want to produce code for AsmJS or WebAssembly, you need to use the Emscripten toolchain and by extension, its execution environment (JavaScript port of libc, libSDL2, etc.). In order to write portable web client-side libraries, we need to be able distinguish at least between `Emscripten + AsmJS`, `Emscripten + WebAssembly` and pure `WebAssembly`. Making WebAssembly, Emscripten and AsmJS distinct version identifiers on the D side makes the most sense, IMO.

ZombineDev:
If I rename `WASM32` to `WebAssembly` (implying that users will use `D_LP64` to differentiate between 32-bit and 64-bit) and leave `Emscripten` would you be ok with this pull request?

WalterBright:
> If I rename WASM32 to WebAssembly (implying that users will use D_LP64 to differentiate between 32-bit and 64-bit) and leave Emscripten would you be ok with this pull request?

Yes, that sounds like a good plan.

ZombineDev:
Further research on my part shows that we can't get away without `version (AsmJS)`, so I added it. @WalterBright I hope your ok with this. Thankfully the other competing technology - PNaCl is being phased out, so we can avoid adding yet another version identifier.

ZombineDev:
Thanks!

polaric:
Is this project still active (WASM)? 
I am very interested in testing it out.

cosinus2:
@polaric https://github.com/cosinus2/dlang-emscripten-demo

PR #7203, opened 2017-10-09
fix issue 17853 - Switch statement without braces only works with one case

somzzz:
Passing ```PScurlyscope``` made the parser think there was a ```{ }``` block there.

### DIFF
--- a/src/ddmd/lexer.d
+++ b/src/ddmd/lexer.d
@@ -262,7 +262,7 @@ class Lexer
         {
             scan(&token);
         }
-        //token.print();
+        //printf(token.toChars());
         return token.value;
     }
 
--- a/src/ddmd/parse.d
+++ b/src/ddmd/parse.d
@@ -5672,7 +5672,9 @@ final class Parser(AST) : Lexer
                     s = new AST.CompoundStatement(loc, statements);
                 }
                 else
-                    s = parseStatement(PSsemi | PScurlyscope);
+                {
+                    s = parseStatement(PSsemi);
+                }
                 s = new AST.ScopeStatement(loc, s, token.loc);
 
                 if (last)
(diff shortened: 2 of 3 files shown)


mathias-lang-sociomantic:
Missing test case

WalterBright:
I'm unsure about this. It seems odd that adding the flag will make it compile, and I'm not sure it does work.

mathias-lang-sociomantic:
Also switch statement as presented in the test case doesn't seem much more useful than an `if`. Maybe changing the grammar would be a more sensible approach ?

somzzz:
The reason why adding the flag produced the wrong behavior is this ```if``` when parsing ```case``` tokens:
https://github.com/somzzz/dmd/blob/a629ebf9973c0860b20ecf9ba8791d74730f071b/src/ddmd/parse.d#L5665

When we parse the first case, we enter the else branch of
```if (flags & PScurlyscope)```
parse the next token, which is also a case.

If we pass ```PScurlyscope``` in ```parseStatement(PSsemi | PScurlyscope)```, the 2nd case will enter the ```then`` branch of that ```if```` which is incorrect since there is no actual curlyscope token. This will just go in a ```while``` loop grouping all the statements it finds until specific tokens, hence the bug:
```while (token.value != TOKcase && token.value != TOKdefault && token.value != TOKeof && token.value != TOKrcurly) ```

**Comment by ibuclaw** [comment]
@mathias-lang-sociomantic this certainly does  falls into the "syntax inherited from C" category, but it is valid grammar as per spec.

You should be able to have:
```
switch (s.omething().withsideeffects())
  case 1: ... case blah:
    { blockcode; }
```
I don't think it's any more stranger than braceless if statements.

andralex (approved):
This PR fixes a codegen error on source code that is currently valid. Changing the language to reject etc. can be always done later.

PR #7204, opened 2017-10-09

MartinNowak:


MartinNowak:
Could you easily add a rebuild trigger to your ci tester @CyberShadow? The next version of dub (1.6.0 Nov 1st) will have the registry mirror as fallback and should less often fail, still might be useful.

PR #7207, opened 2017-10-11
propagate errors for AttribDeclaration.semantic()

WalterBright:
This helps prevent cascaded error messages.


PetarKirov (approved):

ZombineDev:
If we restart https://travis-ci.org/dlang/dmd/jobs/286746709 it should be all green on all CIs.

JinShil (approved):

PR #7208, opened 2017-10-11
Fix 17893: [REG 2.076.1] contract in templatized final class require nothrow

ibuclaw:
https://issues.dlang.org/show_bug.cgi?id=17893.

So now it is being generated too late.

Moving the fix so that it is between _after_ return type has been inferred, but _before_ inferred function attributes have been applied.

Also merging various contract tests into one monolithic testcontracts source file, rather than create yet another test file.


WalterBright:
Perhaps instead of moving around this large block of code, factor it out into a separate function and move the call of it around?

ibuclaw:
Had a good reason to move it to a function anyway, there are two cases to handle.  Generating pre/postcondition condition contracts for a function with - and without a function body.

src/ddmd/func.d
```diff
@@ -177,6 +177,76 @@ private bool needsFensure(FuncDeclaration fd)
     return false;
 }
 
+/****************************************************
+ * Rewrite contracts as nested functions, then call them. Doing it as nested
+ * functions means that overriding functions can call them.
+ * Params:
+ *      fd = the function to search
```
WalterBright: It's `fdx`

WalterBright: `the function to search` ?? What is it searching for?

ibuclaw: Oops, I didn't update that after copy/paste.

MartinNowak:
> Perhaps instead of moving around this large block of code, factor it out into a separate function and move the call of it around?

Saves a lot of time on my side for merging as well.

schveiguy:
@ibuclaw FYI if you don't put the colon after the "Fix 12345", or at least include a space before the colon, then the bot recognizes it.

ibuclaw:
The things that you need to do these days to support tooling.

PR #7209, opened 2017-10-12
refactor: make ExpressionSemanticVisitor private

WalterBright:

Also, now that `semantic` is no longer a virtual function, there is entirely too much overloading of the name going on. This is a start at fixing that.

### DIFF
--- a/src/ddmd/semantic.d
+++ b/src/ddmd/semantic.d
@@ -24,7 +24,6 @@ import ddmd.statement;
 
 import ddmd.initsem;
 import ddmd.dsymbolsem;
-import ddmd.expressionsem;
 import ddmd.statementsem;
 import ddmd.templateparamsem;
 import ddmd.typesem;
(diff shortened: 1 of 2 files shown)


PetarKirov (approved):

src/ddmd/semantic.d
```diff
@@ -38,12 +37,10 @@ extern(C++) void semantic(Dsymbol dsym, Scope* sc)
     dsym.accept(v);
 }
 
-// entrypoint for semantic ExpressionSemanticVisitor
 extern(C++) Expression semantic(Expression e, Scope* sc)
 {
-    scope v = new ExpressionSemanticVisitor(sc);
-    e.accept(v);
-    return v.result;
+    import ddmd.expressionsem;
```
wilzbach: You could have used a selective import here.

wilzbach (approved):

src/ddmd/semantic.d
WalterBright: I'm going to delete the function shortly anyway.

PR #7210, opened 2017-10-12
refactor: make InitializerSemanticVisitor private

WalterBright:

Also, now that `semantic` is no longer a virtual function, there is entirely too much overloading of the name going on. This is a start at fixing that.

### DIFF
--- a/src/ddmd/semantic.d
+++ b/src/ddmd/semantic.d
@@ -22,7 +22,6 @@ import ddmd.init;
 import ddmd.mtype;
 import ddmd.statement;
 
-import ddmd.initsem;
 import ddmd.dsymbolsem;
 import ddmd.statementsem;
 import ddmd.templateparamsem;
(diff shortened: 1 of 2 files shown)


src/ddmd/semantic.d
```diff
@@ -60,9 +59,8 @@ extern(C++) Expression semantic(Expression e, Scope* sc)
  */
 extern(C++) Initializer semantic(Initializer init, Scope* sc, Type t, NeedInterpret needInterpret)
 {
-    scope v = new InitializerSemanticVisitor(sc, t, needInterpret);
-    init.accept(v);
-    return v.result;
+    import ddmd.initsem;
```
ibuclaw: Should probably explicitly import only `initializerSemantic` here.

WalterBright: I plan on deleting that function when this initiative is complete anyway.

PetarKirov (approved):

JinShil (approved):

PR #7211, opened 2017-10-12
refactor: make DsymbolSemanticVisitor private

WalterBright:

Also, now that semantic is no longer a virtual function, there is entirely too much overloading of the name going on. This is a start at fixing that.

### DIFF
--- a/src/ddmd/dsymbolsem.d
+++ b/src/ddmd/dsymbolsem.d
@@ -69,6 +69,15 @@ import ddmd.visitor;
 
 enum LOG = false;
 
+/*************************************
+ * Does semantic analysis on the public face of declarations.
+ */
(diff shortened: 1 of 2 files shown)


PetarKirov (approved):

src/ddmd/dsymbolsem.d
```diff
@@ -69,6 +69,15 @@ import ddmd.visitor;
 
 enum LOG = false;
 
+/*************************************
+ * Does semantic analysis on the public face of declarations.
```
wilzbach: [Shouldn't every public D function have a `Params` list](http://www.digitalmars.com/d/archives/digitalmars/D/Phobos_Documentation_-_call_to_action_258777.html)?

wilzbach (approved):

PR #7214, opened 2017-10-12
Add win64 build and test or something

dmdw64:


appveyor.yml
```diff
@@ -0,0 +1,35 @@
...
+    name: phobos
+
+init:
+  - git config --global core.autocrlf input
+
+build_script:
+  - cd c:/projects/
+  - call "c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
+  - bash --version
+  - appveyor DownloadFile "https://ftp.gnu.org/gnu/make/make-4.2.tar.gz" -FileName make.tar.gz
```
wilzbach: While I would love to use GNUmake on Windows, the sad reality is that DMD ships with DM make and thus that's the only make binary we can expect people to have on Windows. @CyberShadow and I have tried to push for changing the status quo in the past, but so far without any real progress :/

dmdw64: @wilzbach 

As mentioned, it's required to run the tests. I'd be fine with just using GNU make for all the make scripts. It's not fun using two different make systems who's executables are both named "make". I'd much rather have one make system, then GNU make can be provided with the releases instead of DM's make.

Fixing the make situation would also be nice. Windows has the most amount of make configurations but the least amount of test servers. It's not ideal at all, this pull request was to at the very least get win64.mak tested as it isn't being run as part of any tests at all.

wilzbach:
@dmdw64 I just enabled the AppVeyor hook again and closed/reopened your PR to trigger a build.

appveyor.sh
```diff
@@ -0,0 +1,46 @@
...
+
+if [ $D_COMPILER == "dmd" ]; then
+    appveyor DownloadFile "http://downloads.dlang.org/releases/2.x/${D_VERSION}/dmd.${D_VERSION}.windows.7z" -FileName dmd2.7z
+    7z x dmd2.7z > /dev/null
+    export PATH=$PWD/dmd2/windows/bin/:$PATH
+    export DMD=/c/projects/dmd2/windows/bin/dmd.exe
+    dmd --version
+fi
+
+if [ $APPVEYOR_REPO_TAG == true ]; then
```
wilzbach: We need to check whether the upstream branch (e.g. `stable`) exists on these repos too.
See e.g.:

https://github.com/dlang/dmd/blob/d28274e1c666ce8c8905587b8d86945927302315/circleci.sh#L84-L92

and 

https://github.com/dlang/dmd/blob/d28274e1c666ce8c8905587b8d86945927302315/travis.sh#L94-L102

appveyor.sh
```diff
@@ -0,0 +1,46 @@
...
+    git clone --branch $APPVEYOR_REPO_TAG_NAME https://github.com/dlang/druntime
+    git clone --branch $APPVEYOR_REPO_TAG_NAME https://github.com/dlang/phobos
+else
+    git clone --branch master https://github.com/dlang/druntime
+    git clone --branch master https://github.com/dlang/phobos
+fi
+
```
wilzbach: Not sure whether CircleCi supports it, but otherwise we need to manually merge the PR into master, see e.g.:

https://github.com/dlang/dmd/blob/d28274e1c666ce8c8905587b8d86945927302315/circleci.sh#L73-L82

wilzbach (approved):
> Fixing the make situation would also be nice. Windows has the most amount of make configurations but the least amount of test servers. It's not ideal at all,

Agreed.

> this pull request was to at the very least get win64.mak tested as it isn't being run as part of any tests at all. This PR was sitting here for ~2 months before it got its first comment..

I'm really sorry about this. Imho we should move on with this as an initial version as soon as the tests on AppVeyor pass. What do other people think?
CC @CyberShadow @ZombineDev @MartinNowak

appveyor.sh
dmdw64: I think it does do the merge first, at least by the how it grabs the commit it looks like it does.

dmdw64:
Think you can enable appveyor again? It doesn't seem to be detecting the changes. 

Might have to use an older VS, I think I remember not being able to get it work with VS2015 for whatever reason. DMD would always crash when trying to build but would work with the older VS2010.

wilzbach:
> Think you can enable appveyor again? It doesn't seem to be detecting the changes.

Sorry. I won't disable it again.

> Might have to use an older VS, I think I remember not being able to get it work with VS2015 for whatever reason. DMD would always crash when trying to build but would work with the older VS2010.

That sounds really terribly. I presume @rainers could help here to get this PR to the finish line?

wilzbach:
> /c/projects/dmd/appveyor.sh: line 25: clone: command not found

`clone` is defined in the CI scripts I referenced:

https://github.com/dlang/dmd/blob/d28274e1c666ce8c8905587b8d86945927302315/circleci.sh#L17-L31

rainers:
> That sounds really terribly. I presume @rainers could help here to get this PR to the finish line?

dmd is currently broken when building COFF due to https://github.com/dlang/dmd/pull/7455.

ZombineDev:
Let's try restarting the appveyor build.

wilzbach:
> Let's try restarting the appveyor build.

Rebased to master to incorporate the news changes (that's merging with master should be done as part of the script).

ZombineDev:
Finally! Thanks to @dmdw64 and @wilzbach's efforts the build is looking like it's going to pass for the first time. At this rate the log output in AppVeyor behaves like a browser benchmark 😄

wilzbach:
Well it was @rainers who fixed DMD, which clearly shows why there is a need for such an CI. Luckily the respective PR was merged yesterday and is available via dmd-nightly. :)

ZombineDev:
Of course, without @rainers work there would be no windows build to talk about ;)

appveyor.yml
```diff
@@ -0,0 +1,35 @@
...
+
+environment:
+  matrix:
+    - ARCH:                x64
+      D_COMPILER:          dmd
+      D_VERSION:           2.077.1
+      C_COMPILER:          MSVC
+
+artifacts:
+  - path: src/dmd.exe
```
dmdw64: Not sure if this should be removed or not, it just allows for the win64 executable to be easily accessible. Can just replace the 32-bit one, after it is installed with this.

dmdw64:
@wilzbach 

The merging with master should be done as part of the appveyor script:

```
git clone -q https://github.com/dlang/dmd.git C:\projects\dmd
git fetch -q origin +refs/pull/7214/merge:
git checkout -qf FETCH_HEAD
```

And yah that -v parameter should definitely be removed.

src/vcbuild/warnings.h
```diff
@@ -33,6 +33,11 @@
 #pragma warning(disable:4366) // The result of the unary '&' operator may be unaligned
 #pragma warning(disable:4267) // conversion from 'size_t' to 'unsigned int', possible loss of data
 #pragma warning(disable:4310) // cast truncates constant value
+#pragma warning(disable:4838) // requires a narrowing conversion
+#pragma warning(disable:4577) // 'noexcept' used with no exception handling mode specified
+#pragma warning(disable:4805) // '|': unsafe mix of types in operation
+#pragma warning(disable:4311) // type case: pointer truncation
+#pragma warning(disable:4312) // type case: conversion from 'unsigned int' to 'void *' of greater size
```
rainers: The latter 2 warnings should not be muted, but the unused code in os.c disabled for _WIN64.
The other warnings are not specific to the 64-bit build, so should be moved up.
When compiling the VS project (set to /W4) in VS2015, I also get a lot of these:
```
#pragma warning(disable:4456) // declaration of 'ty' hides previous local declaration
#pragma warning(disable:4458) // declaration of 'var' hides class member
#pragma warning(disable:4459) // declaration of 'var' hides global declaration
```

rainers:
- Running the druntime/phobos unittests too should only add a couple of minutes to the build.

- As we don't test building with -m32mscoff yet, maybe target that with the 64-bit compiler?

- Should this also try building through the VS solution? Maybe the 32-bit MS-COFF build with just a sanity check (e.g. druntime unittests)?

src/vcbuild/warnings.h
rainers: > The latter 2 warnings should not be muted, but the unused code in os.c disabled for _WIN64.

https://github.com/dlang/dmd/pull/7514

rainers:
adjusted warnings as commented, removed -v.

rainers:
Building the tests with -j3 seems to stress Appveyor too much. It takes about 20 minute, while LDC builds the tests in about 8 minutes. I'll try removing that option.

appveyor.yml
MartinNowak: We already have a go for GNU make on Windows @wilzbach, just needs someone to implement the transition.

appveyor.yml
MartinNowak: Why would you want to provide artifacts?
We already have http://nightlies.dlang.org/ for nightlies.

appveyor.yml
MartinNowak: How about caching that installation?
https://www.appveyor.com/docs/build-cache/

appveyor.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/sh
```
MartinNowak: Why should this be a bash/sh file, makes it more complex to reproduce, and there are already batch and powershell. Also this script does very little, so it hardly should a problem to reproduce it in PowerShell.

MartinNowak (commented):
Looks mostly good.

appveyor.yml
rainers: > We already have http://nightlies.dlang.org/ for nightlies.

It doesn't provide a 64-bit Windows executable of dmd, does it?

rainers:
> It takes about 20 minute, while LDC builds the tests in about 8 minutes.

LDC builds are faster because not all combinations of command line flags are tested, just the quick tests. With -j1 needed 40 minutes, we are back to 20 minutes with -j2.

appveyor.sh
dmdw64: Need to use bash in order to run the tests for dmd, so there's no avoiding it completely. Can make it all a powershell script except for the part that's needed to run the tests.

dmdw64:
Well this is why I used bash as well, it's much more CI friendly and the logs it produces are much nicer. I'll tinker around with it some more later maybe but I think it would just be better to use bash. You also can't really run the scripts locally anyways as it uses appveyor specific commands/variables.

dmdw64:
@rainers 
I don't know if we should test 32mscoff? Having 64-bit binary is of benefit, but if dmd wants to use optlink for their 32-bit builds then I don't see sense in having 2 of the same.

Should definitely make sure the VisualD project file build though. Don't know how you should go about installing it into appveyor though.

I'll add the unittests for phobos/druntime when I get a chance tomorrow.

rainers:
> I don't know if we should test 32mscoff? Having 64-bit binary is of benefit, but if dmd wants to use optlink for their 32-bit builds then I don't see sense in having 2 of the same.

The auto tester is testing the generation of 64-bit COFF and 32-bit OMF. 32-bit COFF is not covoered yet.

> Should definitely make sure the VisualD project file builds. Don't know how you would go about installing VisualD into appveyor.

I think this can go into a new PR. I can add that integration, I've done that for Visual D, too.

> I'll add the unittests for phobos/druntime when I get a chance tomorrow.

I think this can also wait for another PR.

appveyor.sh
rainers: I agree, it's not a big deal to use bash here. It is not an additional dependency as the tests require it anyway.

appveyor.yml
```diff
@@ -0,0 +1,29 @@
...
+    name: dmd
+  - path: phobos64.lib
+    name: phobos
+
+init:
+  - git config --global core.autocrlf input
+
+build_script:
+  - cd c:/projects/
+  - call "c:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" amd64
```
wilzbach: If I understand @rainers work correctly, this won't be necessary anymore in a foreseeable future

appveyor.yml
```diff
@@ -0,0 +1,29 @@
+
+os: Visual Studio 2015
+
+environment:
+  matrix:
+    - ARCH:                x64
+      D_COMPILER:          dmd
+      D_VERSION:           2.077.1
```
wilzbach: We shouldn't forget to change this to 2.078.0 once released and remove the hard-coded dmd-nightly.

wilzbach (approved):
> I think this can go into a new PR.

Yes, move in small, incremental steps, but succeed :)
Imho this is more than enough for an initial version. I prefer Bash vastly over Powershell too - we need bash anyways and Bash is something we all are familiar with and can run locally as well.
Also this would allow use to share code between Travis, CircleCi and AppVeyor.

Thanks again to everyone for pushing and working on this!

wilzbach:
> The merging with master 

This is still missing too, but it's not really blocking. I am more than happy to quickly add it afterwards.

Anyone against squashing our combined work into one commit?

rainers:
> This is still missing too

Isn't this done automatically, see for example https://ci.appveyor.com/project/greenify/dmd/build/1.0.373#L4

> Anyone against squashing our combined work into one commit?

Fine with me.

wilzbach:
> Isn't this done automatically, see for example 

Oh yes. I should take more care when looking at stuff from my phone. Sorry.

>> Anyone against squashing our combined work into one commit?
> Fine with me.

OK done.

appveyor.yml
MartinNowak: Yes

appveyor.yml
```diff
@@ -0,0 +1,29 @@
+
+os: Visual Studio 2015
+
+environment:
+  matrix:
+    - ARCH:                x64
```
MartinNowak: ARCH isn't used below.

PR #7215, opened 2017-10-13
switch to expressionSemantic() from semantic()

WalterBright:
I hadn't realized there were so many calls. Anyhow, this makes things much more greppable.

### DIFF
--- a/src/ddmd/aliasthis.d
+++ b/src/ddmd/aliasthis.d
@@ -17,6 +17,7 @@ import ddmd.aggregate;
 import ddmd.dscope;
 import ddmd.dsymbol;
 import ddmd.expression;
+import ddmd.expressionsem;
 import ddmd.globals;
 import ddmd.identifier;
 import ddmd.mtype;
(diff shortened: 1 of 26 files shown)


WalterBright:
While a lot of lines, it is just a name change. Trivial.

JinShil (approved):
Nit:  I don't really like the name, as "expressionSemantic" being a member of type "Expression" seems redundant.  It would really only make sense if you were trying to disambiguate it from other "...Semantic" members.

That being said, I don't see any problem with this PR.

ibuclaw:
> I don't really like the name, as "expressionSemantic" being a member of type "Expression" seems redundant

They are all free functions now. If the invocation still looks like a member function call, that's because of ufcs.

JinShil:
> They are all free functions now.

Still, can't D disambiguate based on the parameter types (`semantic(Express e)` vs `semantic(DSymbol s)`)?

ibuclaw:
It can, the reasoning was mentioned in #7209 (and a few other PRs that renamed other semantic functions).

WalterBright:
Please pull this one so I can do the next one.

PR #7219, opened 2017-10-15
evaluate core.math.toPrec intrinsic

MartinNowak:
requires: https://github.com/dlang/druntime/pull/1941


PetarKirov (approved):
Looks good. Glad to see that the template version looks just as simple as the non-template one.

MartinNowak:
Unfortunately it seems like we also need to lower the intrinsic's IR/codegen.
https://github.com/dlang/druntime/pull/1941#issuecomment-336741097

ibuclaw:
Rebased and retesting in #10654.

PR #7221, opened 2017-10-16
switch to statementSemantic()

WalterBright:


RazvanN7:
Are we doing this for all semantic AST node classes? Otherwise, what is the point in doing this just for statement nodes?

WalterBright:
All of them.

JinShil (approved):

src/ddmd/semantic.d
```diff
@@ -54,7 +54,7 @@ extern(C++) Initializer semantic(Initializer init, Scope* sc, Type t, NeedInterp
 }
 
 // Performs semantic analysis in Statement AST nodes
-extern(C++) Statement semantic(Statement s, Scope* sc)
+extern(C++) Statement xxsemantic(Statement s, Scope* sc)
```
ZombineDev: I guess now that you've found and replaced all of the call-sites this function can now be removed.

PetarKirov (approved):

src/ddmd/semantic.d
WalterBright: If I remove it now, then the other PRs will have merge conflicts, and I gotta do a rebase. Simpler to just do it later.

PR #7224, opened 2017-10-16
Remove outdated math headers

joakim-noah:
These glibc headers say not to use them for something like 20 years, so I'm not sure when they were actually used.  [As I noted in the forum earlier, this breaks the compile for me](http://forum.dlang.org/thread/xgtbpcvbikxlilanrzqa@forum.dlang.org) on Arch linux, which is why [I took it out of ldc last year](https://github.com/ldc-developers/ldc/pull/1447/files#diff-2268700876655bdfa9d178b9cc466277).

With all @ibuclaw's other recent commits, dmd from dmd-cxx builds for me again with this commit, though I haven't run the tests.  One issue I noticed is that this dmd says the version is 2.069-devel-*, but Iain said it's up to the 2.071 frontend in his commits.  @braddr, do you think we can get the auto-tester to build dmd master using this dmd-cxx branch, once we're sure it's working well?

Also, I'm not sure what Iain's plan is with this branch, I thought it was just to get the last mostly C++ frontend, 2.068.2, building again, to always be able to bootstrap with it, wheareas he seems to be updating it to newer backported frontends?  When is that going to stop, whenever gdc finally switches over to the ddmd frontend fully?

### DIFF
--- a/src/root/ctfloat.c
+++ b/src/root/ctfloat.c
@@ -387,10 +387,6 @@ size_t CTFloat::hash(real_t a)
 #if __linux__ || __APPLE__ || __FreeBSD__ || __OpenBSD__
 
 #include <math.h>
-#if __linux__
-#include <bits/nan.h>
-#include <bits/mathdef.h>
-#endif
 #if __FreeBSD__ && __i386__
 #include <ieeefp.h>
 #endif


ibuclaw (approved):

ibuclaw:
> With all @ibuclaw's other recent commits, dmd from dmd-cxx builds for me again with this commit, though I haven't run the tests. One issue I noticed is that this dmd says the version is 2.069-devel-*, but Iain said it's up to the 2.071 frontend in his commits. @braddr, do you think we can get the auto-tester to build dmd master using this dmd-cxx branch, once we're sure it's working well?

I'm not so sure about getting all parts of the testsuite passing is a priority for me.  The front-end as you see it passes both the library unittests and D2 testsuite when backed by gdc.

Up to 2.071 is a little stretch of the imagination.  It's up to 2.071 in the sense of bugs and regression fixes have been backported and alignment of headers done.  Not all features are present.

For instance, gdc is up to 2.076 (but explicitly does not include `static foreach`).

> Also, I'm not sure what Iain's plan is with this branch, I thought it was just to get the last mostly C++ frontend, 2.068.2, building again, to always be able to bootstrap with it, wheareas he seems to be updating it to newer backported frontends? When is that going to stop, whenever gdc finally switches over to the ddmd frontend fully?

Its partly also to give some legitimacy for gdc in my fork of the C++ front-end.  The intent was that the switch to ddmd to be as minimal as possible.  And indeed, I think I've succeeded in that (https://github.com/D-Programming-GDC/GDC/pull/550) - the changes to the glue layer are just removing code, and adding a call to `rt_init` / `gc_disable`.  That's it!

Once we've switched to ddmd fully, I think this will be archived off.  I don't intend to do that though until all primary ports are up and working.  Also having the C++ front-end included in gcc-8 means that there is at least a starting based for bootstrapping latter releases.

joakim-noah:
I looked into the phobos build failure, got it to build after cherry-picking dlang/phobos@8e6faa5, dlang/phobos@a33a6d8, and not building `std.stream`, which was removed a couple releases later.  However, druntime's tests wouldn't build, first because `test_runner.d` needed to be compiled with `-fPIC`, next because of some missing functions:
```
../dmd/src/dmd -m64 -conf= -O -release -dip25 -w -Isrc -Iimport  -fPIC -ofobj/64/test_runner src/test_runner.d -Lobj/64/libdruntime-linux64-ut.so -debuglib= -defaultlib=       
obj/64/test_runner.o:(.data.DW.ref.__dmd_personality_v0+0x0): undefined reference to `__dmd_personality_v0'     
obj/64/test_runner.o: In function `_D11test_runner6testerFZb':
src/test_runner.d:(.text._D11test_runner6testerFZb+0x181): undefined reference to `__dmd_begin_catch'                  
obj/64/libdruntime-linux64-ut.so: undefined reference to `_d_throwdwarf'
```
It is unclear where the frontend is at with this branch now, maybe Iain can update the dmd-cxx branches of druntime and phobos to the appropriate commits.

ibuclaw:
> It is unclear where the frontend is at with this branch now, maybe Iain can update the dmd-cxx branches of druntime and phobos to the appropriate commits.

Actually I should resync this branch with gdc.

PR #7225, opened 2017-10-16
Lower dynamic array comparison to __equals template

somzzz:


src/ddmd/expressionsem.d
```diff
@@ -7859,16 +7859,47 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             }
         }
 
+        Type t1 = exp.e1.type.toBasetype();
+        Type t2 = exp.e2.type.toBasetype();
+
+
+        bool needsDirectEq()
```
stefan-koch-sociomantic: could you refactor this to take parameters and not depend on internal state ?
such a refactoring would increase the reusablity.

andralex (approved):

src/ddmd/expressionsem.d
```diff
@@ -7859,16 +7859,47 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
         }
 
-        if (auto e = typeCombine(exp, sc))
+
+        if (!(t1.ty == Tarray && t2.ty == Tarray && needsDirectEq(t1, t2)))
         {
-            result = e;
-            return;
+            if (auto e = typeCombine(exp, sc))
+                {
```
stefan-koch-sociomantic: indentation seems messed up

somzzz: fixed

andralex (approved):
cc @WalterBright

JinShil:
This PR may have introduced the following regression: https://issues.dlang.org/show_bug.cgi?id=18115

MartinNowak:
Looks like it also introduced [18252 – [Reg 2.078] comparison of arrays of associative arrays no longer compiles](https://issues.dlang.org/show_bug.cgi?id=18252).

John-Colvin:
also, this: https://issues.dlang.org/show_bug.cgi?id=18267

PR #7227, opened 2017-10-16
[dmd-cxx] Merge regression tests and fixes for 2.072 

ibuclaw:
And set the library version to v2.072.2

FYI @joakim-noah - there's plenty more where this came from. :-)

@wilzbach - I'll do this one release at a time (depending on how big the change is).

### DIFF
--- a/src/attrib.c
+++ b/src/attrib.c
@@ -32,6 +32,7 @@
 #include "utf.h"
 #include "mtype.h"
 
+bool definitelyValueParameter(Expression *e);
 
 /********************************* AttribDeclaration ****************************/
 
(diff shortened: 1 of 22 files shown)


ibuclaw:
I'm not sure if it's worth updating the backend, but at some point things probably will start breaking.

joakim-noah:
The dmd backend doesn't change much, we could probably tweak that as needed.  Once you're done updating the frontend in this `dmd-cxx` branch, you should update druntime and phobos also for their `dmd-cxx` branch, so that the auto-tester can be run again.  Just get them close to where they should be and I'll cherry-pick the remaining commits to get them building and passing the tests again.

ibuclaw:
There's a dmd-cxx branch for phobos? Why?

There's was a little bit of forward cherry picking in some of the earlier backports for me, but the finished 2.076 frontend should be able to build 2.076 without modifications.

joakim-noah:
So that we could actually provide a dmd-cxx compiler that could be used for ports to new platforms, [like the guy who wanted to try dmd on OpenBSD again](http://forum.dlang.org/thread/hpauduyizitshugjjbfa@forum.dlang.org).

I don't care if these intermediate pulls (why separate your pulls?  I don't see why you wouldn't put all your C++ commits up to 2.076 into your next pull, though with a lot of commits, like you did last time) have a working stdlib in their dmd-cxx branches, but whenever you get dmd-cxx caught up to gdc, just update the dmd-cxx branch of druntime and phobos to what works.  The idea is to use this dmd-cxx branch to bootstrap new OS platforms that run on x86/x64 chips, admittedly a minority these days.

WalterBright:
I'd merge it, but it fails the auto-tester.

ibuclaw:
It's never passed the auto tester just yet.  Getting that fixed is on my maybe later pile.

joakim-noah:
@WalterBright, it does pass the stdlib tests on Travis though.  This is a new experimental branch that doesn't pass the auto-tester yet, but I will try to make it pass at the end, once Iain makes all the dmd frontend changes he wants.

WalterBright:
Ok, I'll pull it on faith :-)

wilzbach:
> There's a dmd-cxx branch for phobos? Why?

Yes there's one for druntime too as chances are that sooner or later one needs to do small fixes there. Also all/most build scripts used here at the dlang repos have the convention to checkout the respective branch if existent for the other two repositories.

ibuclaw:
Well, maybe for any ports that are added post 2.076.

PR #7228, opened 2017-10-17
refactor: replace semantic() with dsymbolSemantic()

WalterBright:


JinShil (approved):

src/ddmd/aggregate.d
```diff
@@ -21,6 +21,7 @@ import ddmd.declaration;
 import ddmd.dscope;
 import ddmd.dstruct;
 import ddmd.dsymbol;
+import ddmd.dsymbolsem;
```
wilzbach: You could have used a selective import here.

wilzbach (approved):

src/ddmd/aggregate.d
WalterBright: 1. there'd not be much value in that, as dsymbolsem has only a handful of publics in it.

2. adding them now would bloat up these PRs by hundreds of lines.

3. I'm not real sure it is the right approach for dmd

ibuclaw: I think that while we should aim to reduce toplevel imports. Selective imports should probably be kept function local.

Eg: genCmain is the only user of the module ddmd.mars. Therefore it makes sense to remove the toplevel import of mars and encapsulate it into genCmain.

ibuclaw:
It irks me that there's a lot of network related failures in our ci as of recent. @MartinNowak we have no logic for attempting actions in our scripts right? If I have time I'll look into that (ie: try 3 times to download file before failing).

WalterBright:
@ibuclaw can we get this merged sooner? Thanks!

ibuclaw (approved):

WalterBright:
Thanks!

MartinNowak:
> It irks me that there's a lot of network related failures in our ci as of recent. @MartinNowak we have no logic for attempting actions in our scripts right? If I have time I'll look into that (ie: try 3 times to download file before failing).

We do have retries and fallback URLs in most places @ibuclaw. One problem is that there are too many CI systems which increases failure rates and makes debugging harder.

PR #7229, opened 2017-10-17
Fix Issue 17748 - extern(C) do nothing on struct methods

RazvanN7:
This is just a starting point implementation. The problem is that classes defined with extern (C) can be identified at parsing time, while nested functions can be excluded (from what I've seen) only during semantic.

### DIFF
--- a/src/ddmd/attrib.d
+++ b/src/ddmd/attrib.d
@@ -439,7 +439,20 @@ extern (C++) final class LinkDeclaration : AttribDeclaration
     {
         super(decl);
         //printf("LinkDeclaration(linkage = %d, decl = %p)\n", p, decl);
+
         linkage = (p == LINKsystem) ? Target.systemLinkage() : p;
+
+        if (linkage == LINKc)
+        {
+            if (decl)
+            {
+                foreach (d; *decl)
+                {
+                    if (d.isAggregateDeclaration)
+                        d.deprecation("cannot be marked as \"extern (C)\".");
+                }
+            }
+        }
     }
 
     static LinkDeclaration create(LINK p, Dsymbols* decl)
--- a/src/ddmd/dmangle.d
+++ b/src/ddmd/dmangle.d
@@ -537,6 +537,12 @@ public:
 
     static const(char)* externallyMangledIdentifier(Declaration d)
     {
+        if (d.isFuncDeclaration && d.linkage == LINKc &&
+           (d.parent !is null && !d.parent.isModule && !d.isStatic))
+        {
+            d.deprecation("is declared \"extern (C)\" and it is neither static nor defined at top level.");
+        }
+
         if (!d.parent || d.parent.isModule() || d.linkage == LINKcpp) // if at global scope
         {
             switch (d.linkage)
@@ -565,6 +571,7 @@ public:
     {
         //printf("Declaration.mangle(this = %p, '%s', parent = '%s', linkage = %d)\n",
         //        d, d.toChars(), d.parent ? d.parent.toChars() : "null", d.linkage);
+
         if (auto id = externallyMangledIdentifier(d))
         {
             buf.writestring(id);
@@ -854,7 +861,7 @@ public:
             mangleIdentifier(s.ident, s);
         else
             toBuffer(s.toChars(), s);
-        //printf("Dsymbol.mangle() %s = %s\n", s.toChars(), id);
+        //printf("Dsymbol.mangle() %s = %s\n", s.toChars(), s.ident);
     }
 
     ////////////////////////////////////////////////////////////////////////////
--- a/test/fail_compilation/fail17748a.d
+++ b/test/fail_compilation/fail17748a.d
@@ -1,0 +1,20 @@
+/*
+REQUIRED_ARGS: -de
+TEST_OUTPUT:
+---
+fail_compilation/fail17748a.d(11): Deprecation: function fail17748a.S.foo is declared "extern (C)" and it is neither static nor defined at top level.
+---
+*/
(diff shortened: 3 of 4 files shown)


RazvanN7:
It seems that druntime has some extern (system) which evaluate to extern (C). I wonder if cleaning druntime will lead to regressions

MartinNowak:
This will break way too much code. It's also inconsistent with other attributes, all of them only apply to some declarations, e.g. a top-level `pure` doesn't have an effect for enums.
`extern(X)` only has an effect on functions, function pointers, and variable mangling, but is ignored for enums, and aggregates.
Also for aggregates there is a difference between attributes on the inside and on the outside, the referenced Issue is concerned with `extern(X)` on member methods, not on the aggregate itself.

RazvanN7:
@MartinNowak So what is your take on this? Should we just drop the patch? IMHO we should update the druntime code (as it is not too much work) so that extern (C) is not applied globally, but only for functions, function pointers and variables. Having extern (C) for all declarations makes it impossible to fix the existing bug.

andralex:
cc @WalterBright as well

stefan-koch-sociomantic:
I agree with @MartinNowak this may very well cause a massive breakage;
which and when we have no functional automatic solution to fix it, it's probably not viable.

RazvanN7:
Anyone who is using extern (C) struct out there will only get a deprecation message (we could replace it with a warning) not an error. If we fix druntime and phobos to use extern only where it actually makes sense we will have a clean code base and also it will stimulate folks to write code free of prolixity.

I volunteer to clean the code and add extern only where it makes sense. I already made a PR for druntime which passes all the tests. Next step is to clean phobos, but I will postpone this until we have a high-level decision.

jacob-carlborg:
> Anyone who is using extern (C) struct out there will only get a deprecation message (we could replace it with a warning) not an error.

And what about those that use the `-de` flag (tread deprecations as errors)?

Let's drop this and move on. There are more important things to work on.

wilzbach:
> And what about those that use the -de flag (treat deprecations as errors)?

FWIW the point about deprecations is that one can gradually upgrade its codebase. If someone is willing to circumvent this, he can't complain that there wasn't enough time or tooling. So imho `-de` isn't a good argument, but that the migration is difficult is or not worth the effort is.
(I know that `-de` is used for Phobos and druntime, but that's because we wanted to ensure that we don't deprecate something that's still being used somewhere.)

Anyhow the discussion at Bugzilla was that the current status quo isn't ideal and at least warnings for  explicit `extern(C)`  on non-static struct member functions could be triggered...

MartinNowak:
This is not so much about the work that needs to be done, and it's a lot more than what you already [find on dub](http://code.dlang.org/?sort=updated&category=library.binding.deimos&limit=100), but it about the fact that this work is necessary.
It's already a dreadful amount of noise necessary to deal with attributes atm.
https://github.com/MartinNowak/io/blob/d0ae747267445ac3033b2da0dfeecdcd0022bc4e/src/std/io/net/addr.d, and translating C headers is tedious and error prone enough.
Taking away `extern(C):` at the top, or requiring people to reorder the module (which is no longer diffable against the C header) is quite an ergonomic problem. Also the fact that it's still not possible to take-away/subtract attributes is already annoying enough, as it sometimes requires per-entity attributes instead.
Now what's on the plus side? More pedantic `extern(C)` attributes, b/c it can no longer be used when it has no effect. But as said that's neither very consistent w/ the selective effect of any attribute (`extern(C)` on struct fields is a no-op for example), nor is it very surprising that `extern(C)` has no effect on member methods. There are no such methods in C, so why should we invent a calling convention for them?

One things that's not too lucky, `extern(C)` conflates 2 concepts, mangling and calling convention.
On free functions it affects both, on static member methods it only affects the calling convention. The latter again is b/c there are no static member methods in C, so it's unclear what mangling we should use, and as structs have separate scopes, a plain mangling could easily lead to link-time collisions.

While I wished the whole design was cleaner, and I understand that many programmers have the tendency towards explicit and precise language, I don't actually think anyone wants a super pedantic compiler that constantly gets in the way.

Also I never found this to be much of an issue, but indeed it's yet another small behavior of the language that needs to be learned.

WalterBright:
I agree with @MartinNowak 's rationale and conclusion.

BTW, in C++ on Windows one could control the name mangling and the calling convention independently. I found this confusing and ultimately pointless, which is why D does not have them separated.

While the current behavior of `extern(C)` seems odd, it works and nobody seems to have an actual problem with it. But changing it will cause work for people, to little discernible purpose. 

So I recommend closing the issue as `INVALID` because it is intended behavior (though it should be documented).

RazvanN7:
Thank you all for your patience and reviews. Closing PR and the bug report.

MartinNowak:
> Thank you all for your patience and reviews. Closing PR and the bug report.

Thanks @RazvanN7 for clarifying this, another unclear point put ad acta.
Would you volunteer to add some words to https://dlang.org/spec/attribute.html#linkage and possibly https://dlang.org/spec/interfaceToC.html?

andralex:
OK. Thanks @RazvanN7 for the work - it wasn't for naught, we got to close the issue. @WalterBright and @MartinNowak for chiming in. Indeed I was about to ask @RazvanN7 to add verbiage to the spec about the behavior of `extern`. Could you please? Thanks!!

RazvanN7:
@andralex @MartinNowak 

> Would you volunteer to add some words to https://dlang.org/spec/attribute.html#linkage and possibly https://dlang.org/spec/interfaceToC.html?

Sure thing

PR #7230, opened 2017-10-18
fix Issue 17900 - FreeBSD 10.3 runnable/cpp_abi_tests.d(94): Assertio…

WalterBright:
…n failure (test suite)

Looks like the 32 bit ABI changed for FreeBSD 10.3. This puts us in the awkward position of the old autotester failing until it is upgraded to 10.3.

### DIFF
--- a/src/ddmd/argtypes.d
+++ b/src/ddmd/argtypes.d
@@ -434,6 +434,25 @@ extern (C++) TypeTuple toArgTypes(Type t)
                     }
                 }
             }
+            else if (!global.params.is64bit && global.params.isFreeBSD && t.sym.fields.dim == 1 &&
+                (sz == 4 || sz == 8))
+            {
+                /* FreeBSD changed their 32 bit ABI at some point before 10.3 for the following:
+                 *  struct { float f;  } => arg1type is float
+                 *  struct { double d; } => arg1type is double
+                 * Cannot find any documentation on it.
+                 */
+
+                VarDeclaration f = t.sym.fields[0];
+                TypeTuple tup = toArgTypes(f.type);
+                if (tup && tup.arguments.dim == 1)
+                {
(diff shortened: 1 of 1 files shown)


jmdavis:
Well, they only guarantee ABI compatibility across minor releases. Going from 9 to 10 or 10 to 11 or whatnot always risks breaking things if you're not recompiling (and worse for us with D given that we have to emulate the C side of things rather than simply #including it). I'd say that we've been pretty lucky that our dmd releases have done as well as they have across major FreeBSD versions. And having FreeBSD 12 work (especially with the same binary that works on 11) is going to be a barrel of fun thanks to them changing the size of inodes: https://issues.dlang.org/show_bug.cgi?id=17596

These are the sort of situations where it's really too bad that we can't simply #include the C headers to use from D somehow, though I have no clue how we ever could have done that.

WalterBright:
This is one of the fixes blocking upgrading the autotester to FreeBSD 10.3.

andralex (approved):

braddr:
Hrm.. did the abi really change?  This _didnt_ break on freebsd 8.x as was expected.  So, what gives?

WalterBright:
Good question. I figure the value in EAX happened to be correct.

WalterBright:
It's also entirely possible that the FreeBSD C ABI was *always* this way, and the test case just happened to pass for reasons mentioned above. The C++ compiler changed for 10.3, and EAX changed, so then it started failing.

This could explain why I have been unable to find any reference to a FreeBSD ABI change.

MartinNowak:
Sounds more likely than an ABI change.

ibuclaw:
I'm pretty sure that this applies to all x86 ABI that isn't Windows.

It's perfectly valid to pass floats and doubles in SSE registers if that is the baseline target CPU the compiler is targeting.

PR #7236, opened 2017-10-23
fix Issue 17842 - [scope] array append allows for escaping references

WalterBright:


WalterBright:
This is fix is all wrong, I need to redo it.

andralex (approved):

PR #7238, opened 2017-10-24
fix Issue 17246 - [REG2.053] Extra destructor call.

WalterBright:
This is a critical bug fix, as it can corrupt memory.

### DIFF
--- a/src/ddmd/expression.d
+++ b/src/ddmd/expression.d
@@ -1483,16 +1483,20 @@ extern (C++) bool preFunctionParameters(Loc loc, Scope* sc, Expressions* exps)
  */
 extern (C++) Expression valueNoDtor(Expression e)
 {
-    if (e.op == TOKcall)
+    auto ex = e;
+    while (ex.op == TOKcomma)
+        ex = (cast(CommaExp)ex).e2;
+
+    if (ex.op == TOKcall)
     {
         /* The struct value returned from the function is transferred
          * so do not call the destructor on it.
          * Recognize:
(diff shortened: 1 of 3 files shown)


mathias-lang-sociomantic:
> This is a critical bug fix, as it can corrupt memory.

Then shouldn't it target stable ?

andralex (approved):
cc @MartinNowak

mathias-lang-sociomantic:
@andralex : Why did you merge it in master despite the review ?

andralex:
@mathias-lang-sociomantic leaving the decision to @MartinNowak the build czar

mathias-lang-sociomantic:
And if it goes into stable, Martin has to cherry-pick it. We never merge master into stable, always the other way around. So if there's a reasonable doubt about a P.R. needing to go to stable, it shouldn't be merged right away like it just happened.

andralex:
Cool, @MartinNowak please advise on how you prefer to proceed.

src/ddmd/expression.d
```diff
@@ -1483,16 +1483,20 @@ extern (C++) bool preFunctionParameters(Loc loc, Scope* sc, Expressions* exps)
  */
 extern (C++) Expression valueNoDtor(Expression e)
 {
-    if (e.op == TOKcall)
+    auto ex = e;
+    while (ex.op == TOKcomma)
```
stefan-koch-sociomantic: This is such a common thing it's time we factor it into a function isn't it ?

PR #7239, opened 2017-10-24
CTFECache - caching D compiler

DmitryOlshansky:
A preliminary patch to cache CTFE executions in a persistent store.

This allows to reuse expensive CTFE computations across builds.
Ironically processing huge struct literal is still quite costly but is a fight for another day.

In action:
```bash
▶ cat r*.d
import std.regex;

enum r = regex(`\w`);

void main(){}import std.regex;

enum r = regex(`\w\w`);

void main(){}import std.regex;

enum r = regex(`\w\w\w`);

void main(){}import std.regex;

enum r = regex(`\w\w\w\w`);

void main(){}

▶ cat bench.sh
#!/bin/bash
for f in r*.d ; do
        echo "Compiling $f with dmd $@"
        time dmd $1 $f
done
```
## Recent master version
▶ dmd --version
DMD64 D Compiler v2.077.0-144-gbbca650
Copyright (c) 1999-2017 by Digital Mars written by Walter Bright

▶ ./bench.sh
Compiling r1.d with dmd

real    0m3.047s
user    0m2.808s
sys     0m0.232s
Compiling r2.d with dmd

real    0m4.105s
user    0m3.832s
sys     0m0.264s
Compiling r3.d with dmd

real    0m4.682s
user    0m4.372s
sys     0m0.304s
Compiling r4.d with dmd

real    0m5.457s
user    0m5.128s
sys     0m0.320s
./bench.sh  16.14s user 1.12s system 99% cpu 17.294 total

## DCache version

▶ dmd --version
DMD64 D Compiler v2.077.0-122-g7a50140
Copyright (c) 1999-2017 by Digital Mars written by Walter Bright

▶ ./bench.sh
Compiling r1.d with dmd

real    0m3.358s
user    0m3.004s
sys     0m0.348s
Compiling r2.d with dmd

real    0m3.909s
user    0m3.500s
sys     0m0.404s
Compiling r3.d with dmd

real    0m4.579s
user    0m4.152s
sys     0m0.420s
Compiling r4.d with dmd

real    0m5.332s
user    0m4.836s
sys     0m0.488s
./bench.sh  15.49s user 1.66s system 99% cpu 17.182 total

## Now with caching + cache usage traces
▶ ./bench.sh -cache=mmap
Compiling r1.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Recomputed InversionList!(GcPolicy) memoizeExpr() @trusted in 0.690174
Recomputed InversionList!(GcPolicy) wordCharacter() @property @safe in 0.690408
Recomputed Trie!(BitPacked!(bool, 1LU), dchar, 1114112LU, sliceBits!(8LU, 21LU), sliceBits!(0LU, 8LU)) codepointSetTrie(InversionList!(GcPolicy) set) pure @safe in 0.381241
Recomputed CharMatcher this(InversionList!(GcPolicy) set) ref in 0.418019
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 1.111543
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 1.112261

real    0m3.393s
user    0m3.040s
sys     0m0.348s
Compiling r2.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000605
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001930
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 1.750151
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 1.750862

real    0m2.882s
user    0m2.644s
sys     0m0.232s
Compiling r3.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000705
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001632
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 2.486691
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 2.487412

real    0m3.611s
user    0m3.212s
sys     0m0.392s
Compiling r4.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000636
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001907
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 3.138693
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 3.139398

real    0m4.285s
user    0m3.816s
sys     0m0.464s
./bench.sh -cache=mmap  12.72s user 1.44s system 99% cpu 14.176 total

## 2nd run DCache
▶ ./bench.sh -cache=mmap
Compiling r1.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000619
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001780
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002774

real    0m1.202s
user    0m1.064s
sys     0m0.132s
Compiling r2.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000714
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001862
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002727

real    0m1.247s
user    0m1.112s
sys     0m0.132s
Compiling r3.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000631
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001766
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002734

real    0m1.181s
user    0m1.016s
sys     0m0.156s
Compiling r4.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000617
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001752
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002806

real    0m1.169s
user    0m1.076s
sys     0m0.088s
./bench.sh -cache=mmap  4.27s user 0.51s system 99% cpu 4.804 total

### DIFF
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -101,6 +101,9 @@ struct Usage
         Option("c",
             "do not link"
         ),
+        Option("cache=mmap",
+            "use persistent cache for CTFE evaluations"
+        ),
         Option("color",
             "turn colored console output on"
         ),
--- a/src/dmd/dinterpret.d
+++ b/src/dmd/dinterpret.d
@@ -12,16 +12,23 @@
 
 module dmd.dinterpret;
 
+import core.stdc.stdlib;
 import core.stdc.stdio;
 import core.stdc.string;
+import core.stdc.time;
 import dmd.apply;
 import dmd.arraytypes;
+import dmd.astcodegen;
 import dmd.attrib;
 import dmd.builtin;
 import dmd.constfold;
 import dmd.ctfeexpr;
+import dmd.dcache;
 import dmd.dclass;
 import dmd.declaration;
+import dmd.dimport;
+import dmd.dmodule;
+import dmd.dscope;
 import dmd.dstruct;
 import dmd.dsymbol;
 import dmd.dsymbolsem;
@@ -30,14 +37,18 @@ import dmd.errors;
 import dmd.expression;
 import dmd.expressionsem;
 import dmd.func;
+import dmd.hdrgen;
 import dmd.globals;
 import dmd.id;
 import dmd.identifier;
 import dmd.init;
 import dmd.initsem;
 import dmd.mtype;
+import dmd.parse;
 import dmd.root.array;
+import dmd.root.outbuffer;
 import dmd.root.rootobject;
+import dmd.root.sha;
 import dmd.statement;
 import dmd.tokens;
 import dmd.utf;
@@ -720,6 +731,161 @@ extern (C++) Expression ctfeInterpretForPragmaMsg(Expression e)
     return e;
 }
 
+ubyte[20] ctfeHashingSeed()
+{
+    static ubyte[20] hash;
+    static bool inited;
+
+    if (inited) return hash;
+    else
+    {
+        SHA1 hasher;
+        hasher.start();
+        // predefined versions indicate target, bitness and asserts
+        if (global.versionids)
+        {
+            foreach (id; *global.versionids)
+            {
+                const(char)* s = id.toChars();
+                hasher.put(cast(ubyte[])s[0..strlen(s)]);
+            }
+        }
+        if (global.debugids)
+        {
+            foreach (id; *global.debugids)
+            {
+                const(char)* s = id.toChars();
+                hasher.put(cast(ubyte[])s[0 .. strlen(s)]);
+            }
+        }
+        hash = hasher.finish();
+        inited = true;
+        return hash;
+    }
+}
+
+ubyte[20] generateCtfeCacheKey(FuncDeclaration fd, ref Expressions eargs)
+{
+    SHA1 hasher;
+    hasher.start();
+    auto signature = fd.toFullSignature();
+    // Add in a hash of compilation parameters
+    hasher.put(ctfeHashingSeed);
+    debug(DCACHE) printf("Func's parent %s \n", fd.parent.toChars());
+    // Include modules of type arguments
+    if (TemplateInstance ti = fd.parent.isTemplateInstance())
+    {
+        if (ti.tiargs)
+        {
+            foreach (arg; *ti.tiargs)
+            {
+                Dsymbol sym = getDsymbol(arg);
+                if (sym)
+                {
+                    /*if (ScopeDsymbol sd = sym.isScopeDsymbol())
+                    {
+                        foreach(sc; *sd.importedScopes)
+                        {
+                            if (Module md = sc.isModule())
+                            {
+                                debug(DCACHE) auto str = toHexString(md.signature[]);
+                                debug(DCACHE) printf("\tmod %s sig: %.*s\n", 
+                                    md.toChars(), str.length, str.ptr);
+                                hasher.put(md.signature[]);
+                            }
+                        }
+                    }*/
+                    Dsymbol last;
+                    while (sym)
+                    {
+                        last = sym;
+                        sym = sym.parent;
+                    }
+                    if (Module md = last.isModule())
+                    {
+                        foreach (imp; fd._scope._module.aimports)
+                        {
+                            hasher.put(imp.signature[]);
+                        }
+                        debug(DCACHE) auto str = toHexString(md.signature[]);
+                        debug(DCACHE) printf("\tmod %s sig: %.*s\n",
+                            md.toChars(), str.length, str.ptr);
+                        hasher.put(md.signature[]);
+                    }
+                }
+            }
+        }
+    }
+    debug(DCACHE) printf("Imports from function's module:\n");
+    foreach (imp; fd._scope._module.aimports)
+    {
+        hasher.put(imp.signature[]);
+        auto str = toHexString(imp.signature[]);
+        debug(DCACHE) printf("\tmod %s sig: %.*s\n", imp.toChars(), str.length, str.ptr);
+    }
+    hasher.put(fd._scope._module.signature);
+    hasher.put(0x0); // separator
+    hasher.put(cast(ubyte[])signature[0 .. strlen(signature)]);
+    foreach(i, earg; eargs)
+    {
+        hasher.put(0x0); // separator
+        auto argument = earg.toChars();
+        hasher.put(cast(ubyte[])argument[0 .. strlen(argument)]);
+    }
+    return hasher.finish();
+}
+
+extern(C++) class ExtractSymbols : StoppableVisitor {
+    Scope* _sc;
+    Dsymbol[] symbols;
+
+    this(Scope* sc)
+    {
+        _sc = sc;
+    }
+
+    override void visit(Expression e)
+    {
+        if (!e.type.isscalar())
+        {
+            Dsymbol sym = e.type.toDsymbol(_sc);
+            if (sym)
+                symbols ~= sym;
+        }
+    }
+    alias visit = Visitor.visit;
+}
+
+void writeModuleName(Dsymbol sym, ref OutBuffer dest)
+{
+    const(char)*[] parts;
+    sym = sym.getModule();
+    while (sym)
+    {
+        parts ~= sym.toChars();
+        sym = sym.parent;
+    }
+    for (size_t i = 0; i < parts.length; i++)
+    {
+        if (i != 0)
+            dest.writestring(cast(char*)".");
+        dest.writestring(parts[$ - i - 1]);
+    }
+}
+
+void uniqueInPlace(T)(ref T[] args)
+{
+    size_t i = 0;
+    for (size_t j = 0; j < args.length; j++)
+    {
+        if (args[i] != args[j])
+        {
+            args[i++] = args[j];
+        }
+    }
+    args.length = i;
+}
+
 /*************************************
  * Attempt to interpret a function given the arguments.
  * Input:
@@ -835,6 +1001,51 @@ private Expression interpret(FuncDeclaration fd, InterState* istate, Expressions
             return CTFEExp.cantexp;
         }
         eargs[i] = earg;
+    }
+ 
+    // Skip generated symbols and only permit caching of top-level CTFE calls
+    bool cacheable = (istate == null || istate.caller == null)
+        && strncmp(fd.ident.toChars(), "__", 2) != 0;
+    clock_t ctfeStart = clock();
+    ubyte[20] cacheKey = void;
+    if (global.params.cache && cacheable)
+    {
+        cacheKey = generateCtfeCacheKey(fd, eargs);
+        auto value = dcache.get(cacheKey[]);
+        if (value != null)
+        {
+            clock_t startMixin = clock();
+            debug (DCACHE)
+            {
+                printf("interpreting global %s: %s\n", fd.loc.toChars(), fd.toFullSignature());
+                foreach (i, earg; eargs[])
+                {
+                    printf("\targ %d : %s\n", i, earg.toChars());
+                }
+                printf("\tFound in cache\n");
+                printf("BODY %*s\n", value.length, value.ptr);
+            }
+            scope p = new Parser!ASTCodegen(fd.loc, fd._scope._module, cast(string)value ~ "\0", false);
+            p.nextToken();
+            Expression e = p.parseExpression();
+            if (p.errors)
+            {
+                return new ErrorExp();
+            }
+            if (p.token.value != TOK.endOfFile)
+            {
+                e.error("incomplete mixin expression (%s)", e.toChars());
+                return new ErrorExp();
+            }
+            cachedSemantics = true;
+            e = expressionSemantic(e, fd._scope);
+            cachedSemantics = false;
+            clock_t endMixin = clock();
+            double delta = 1.0*(endMixin - startMixin) / CLOCKS_PER_SEC;
+            debug(DCACHE) printf("\tSuccessfully mixed in!\n\n");
+            version(DCACHE_STATS) printf("Mixed in %s from cache in %lf\n", fd.toFullSignature(), delta);
+            return e;
+        }
     }
 
     // Now that we've evaluated all the arguments, we can start the frame
@@ -978,7 +1189,48 @@ private Expression interpret(FuncDeclaration fd, InterState* istate, Expressions
         (cast(ThrownExceptionExp)e).generateUncaughtError();
         e = CTFEExp.cantexp;
     }
-
+    if (global.params.cache && cacheable)
+    {
+        clock_t ctfeEnd = clock();
+        double delta = 1.0*(ctfeEnd - ctfeStart) / CLOCKS_PER_SEC;
+        if (delta > 0.0)
+        {
+            debug(DCACHE)
+            {
+                printf("interpreting global %s: %s\n", fd.loc.toChars(), fd.toFullSignature());
+                foreach (i, earg; eargs[])
+                {
+                    printf("\targ %d : %s\n", i, earg.toChars());
+                }
+            }
+            OutBuffer buf;
+            scope extractor = new ExtractSymbols(fd._scope);
+            walkPostorder(e, extractor);
+            extern(C) static int comp(scope const(void*) *a , scope const(void*)* b)
+            {
+                return a > b ? 1 : (a == b ? 0 : -1);
+            }
+            qsort(extractor.symbols.ptr, extractor.symbols.length, Dsymbol.sizeof, cast(_compare_fp_t)&comp);
+            uniqueInPlace(extractor.symbols);
+            buf.writestring("(){\n");
+            foreach (t; extractor.symbols)
+            {
+                buf.writestring("import ");
+                writeModuleName(t, buf);
+                buf.writestring(" : ");
+                buf.writestring(t.toChars());
+                buf.writestring(";\n");
+            }
+            buf.writestring("return ");
+            buf.writestring(e.toChars());
+            buf.writestring(";\n");
+            buf.writestring("}()\n");
+            auto value = buf.extractString();
+            dcache.put(cacheKey[], cast(ubyte[])value[0 .. strlen(value)]);
+            debug(DCACHE) printf("\treturns (%lf) CACHED %s\n", delta, value);
+            version(DCACHE_STATS) printf("Recomputed %s in %lf\n", fd.toFullSignature(), delta);
+        }
+    }
     return e;
 }
 
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -407,6 +407,8 @@ extern (C++) final class Module : Package
 
     size_t nameoffset;          // offset of module name from start of ModuleInfo
     size_t namelen;             // length of module name in characters
+   
+    ubyte[20] signature;        // A signature of contents, currently SHA1
 
     extern (D) this(const(char)* filename, Identifier ident, int doDocComment, int doHdrGen)
     {
--- a/src/dmd/dscope.d
+++ b/src/dmd/dscope.d
@@ -16,10 +16,12 @@ import core.stdc.stdio;
 import core.stdc.string;
 import dmd.aggregate;
 import dmd.attrib;
+import dmd.dcache;
 import dmd.dclass;
 import dmd.declaration;
 import dmd.dmodule;
 import dmd.doc;
+import dmd.dimport;
 import dmd.dsymbol;
 import dmd.dsymbolsem;
 import dmd.dtemplate;
@@ -554,7 +556,7 @@ struct Scope
              * checked by the compiler remain usable.  Once the deprecation is over,
              * this should be moved to search_correct instead.
              */
-            if (!s && !(flags & IgnoreSymbolVisibility))
+            if (!cachedSemantics && !s && !(flags & IgnoreSymbolVisibility))
             {
                 s = searchScopes(flags | SearchLocalsOnly | IgnoreSymbolVisibility);
                 if (!s)
--- a/src/dmd/dsymbol.d
+++ b/src/dmd/dsymbol.d
@@ -22,6 +22,7 @@ import dmd.aliasthis;
 import dmd.arraytypes;
 import dmd.attrib;
 import dmd.gluelayer;
+import dmd.dcache;
 import dmd.dclass;
 import dmd.declaration;
 import dmd.denum;
@@ -1380,7 +1381,7 @@ public:
                     AliasDeclaration ad = void;
                     // accessing private selective and renamed imports is
                     // deprecated by restricting the symbol visibility
-                    if (s.isImport() || (ad = s.isAliasDeclaration()) !is null && ad._import !is null)
+                    if (cachedSemantics || s.isImport() || (ad = s.isAliasDeclaration()) !is null && ad._import !is null)
                     {}
                     else
                         error(loc, "%s `%s` is `private`", s.kind(), s.toPrettyChars());
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -23,6 +23,7 @@ import dmd.arraytypes;
 import dmd.attrib;
 import dmd.astcodegen;
 import dmd.canthrow;
+import dmd.dcache;
 import dmd.dscope;
 import dmd.dsymbol;
 import dmd.declaration;
@@ -2962,6 +2963,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 if (!sd.ctor)
                     sd.ctor = sd.searchCtor();
 
+                // A cached value is a literal
+                if (cachedSemantics) goto Lx;
                 // First look for constructor
                 if (exp.e1.op == TOK.type && sd.ctor)
                 {
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -146,6 +146,7 @@ struct Param
      * used to hide a feature that will have to go through deprecate-then-error
      * before becoming default.
      */
+    bool cache;             // use persistent caching of CTFE (DCache)
 
     bool showGaggedErrors;  // print gagged errors anyway
     bool manual;            // open browser on compiler manual
--- a/src/dmd/hdrgen.d
+++ b/src/dmd/hdrgen.d
@@ -1530,7 +1530,10 @@ public:
 
     override void visit(TemplateInstance ti)
     {
-        buf.writestring(ti.name.toChars());
+        if (ti.aliasdecl)
+            buf.writestring(ti.aliasdecl.toChars());
+        else
+            buf.writestring(ti.name.toChars());
         tiargsToBuffer(ti);
 
         if (hgs.fullDump)
@@ -2497,7 +2500,7 @@ public:
 
     override void visit(StructLiteralExp e)
     {
-        buf.writestring(e.sd.toChars());
+        buf.writestring(e.sd.type.toChars());
         buf.writeByte('(');
         // CTFE can generate struct literals that contain an AddrExp pointing
         // to themselves, need to avoid infinite recursion:
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -32,6 +32,7 @@ import dmd.dinifile;
 import dmd.dinterpret;
 import dmd.dmodule;
 import dmd.doc;
+import dmd.dcache;
 import dmd.dscope;
 import dmd.dsymbol;
 import dmd.dsymbolsem;
@@ -478,6 +479,8 @@ private int tryMain(size_t argc, const(char)** argv)
     Expression._init();
     Objc._init();
     builtin_init();
+    if (global.params.cache)
+        dcache.initialize(global._version[0..strlen(global._version)]);
 
     if (global.params.verbose)
     {
@@ -1537,6 +1540,10 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 }
                 else if (p[4])
                     goto Lerror;
+            }
+            else if (strcmp(p + 1, cast(char*)"cache=mmap") == 0)
+            {
+                params.cache = true;
             }
             else if (arg == "-shared")
                 params.dll = true;
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -237,7 +237,12 @@ final class Parser(AST) : Lexer
     Loc endloc; // set to location of last right curly
     int inBrackets; // inside [] of array index or slice
     Loc lookingForElse; // location of lonely if looking for an else
-
+    enum withSignature = __traits(hasMember, typeof(mod), "signature");
+    static if (withSignature)
+    {
+        import dmd.root.sha;
+        SHA1 hasher;
+    }
     /*********************
      * Use this constructor for string mixins.
      * Input:
@@ -246,8 +251,14 @@ final class Parser(AST) : Lexer
     extern (D) this(Loc loc, AST.Module _module, const(char)[] input, bool doDocComment)
     {
         super(_module ? _module.srcfile.toChars() : null, input.ptr, 0, input.length, doDocComment, false);
-
         //printf("Parser::Parser()\n");
+        static if (withSignature)
+        {
+            hasher.start();
+            hasher.put(_module.signature[]);
+            hasher.put(cast(ubyte[])input);
+            _module.signature = hasher.finish();
+        }
         scanloc = loc;
 
         if (loc.filename)
@@ -268,7 +279,12 @@ final class Parser(AST) : Lexer
     extern (D) this(AST.Module _module, const(char)[] input, bool doDocComment)
     {
         super(_module ? _module.srcfile.toChars() : null, input.ptr, 0, input.length, doDocComment, false);
-
+        static if (withSignature)
+        {
+            hasher.start();
+            hasher.put(cast(ubyte[])input);
+            _module.signature = hasher.finish();
+        }
         //printf("Parser::Parser()\n");
         mod = _module;
         linkage = LINK.d;
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -287,12 +287,16 @@ D_OBJC := 1
 endif
 endif
 
+# Enable stats for DCache
+#DFLAGS += -version=DCACHE_STATS
+# Enable debug for DCache
+#DFLAGS += -debug=DCACHE
 
 ######## DMD frontend source files
 
 FRONT_SRCS=$(addsuffix .d, $(addprefix $D/,access aggregate aliasthis apply argtypes arrayop	\
 	arraytypes astcodegen attrib builtin canthrow cli clone compiler complex cond constfold	\
-	cppmangle cppmanglewin ctfeexpr dcast dclass declaration delegatize denum dimport	\
+	cppmangle cppmanglewin ctfeexpr dcache dcast dclass declaration delegatize denum dimport	\
 	dinifile dinterpret dmacro dmangle dmodule doc dscope dstruct dsymbol dsymbolsem	\
 	dtemplate dversion escape expression expressionsem func			\
 	hdrgen id impcnvtab imphint init initsem inline inlinecost intrange	\
@@ -304,7 +308,7 @@ FRONT_SRCS=$(addsuffix .d, $(addprefix $D/,access aggregate aliasthis apply argt
 LEXER_SRCS=$(addsuffix .d, $(addprefix $D/, console entity errors globals id identifier lexer tokens utf))
 
 LEXER_ROOT=$(addsuffix .d, $(addprefix $(ROOT)/, array ctfloat file filename outbuffer port rmem \
-	rootobject stringtable hash))
+	rootobject stringtable hash sha sha_SSSE3))
 
 ROOT_SRCS = $(addsuffix .d,$(addprefix $(ROOT)/,aav array ctfloat file \
 	filename man outbuffer port response rmem rootobject speller \
--- a/src/win32.mak
+++ b/src/win32.mak
@@ -150,7 +150,7 @@ DMDMAKE=$(MAKE) -fwin32.mak C=$C TK=$(TK) ROOT=$(ROOT) MAKE="$(MAKE)" HOST_DC="$
 # D front end
 FRONT_SRCS=$D/access.d $D/aggregate.d $D/aliasthis.d $D/apply.d $D/argtypes.d $D/arrayop.d	\
 	$D/arraytypes.d $D/astcodegen.d $D/attrib.d $D/builtin.d $D/canthrow.d $D/cli.d $D/clone.d $D/compiler.d $D/complex.d	\
-	$D/cond.d $D/constfold.d $D/cppmangle.d $D/cppmanglewin.d $D/ctfeexpr.d $D/dcast.d $D/dclass.d		\
+	$D/cond.d $D/constfold.d $D/cppmangle.d $D/cppmanglewin.d $D/ctfeexpr.d $D/cache.d $D/dcast.d $D/dclass.d		\
 	$D/declaration.d $D/delegatize.d $D/denum.d $D/dimport.d $D/dinifile.d $D/dinterpret.d	\
 	$D/dmacro.d $D/dmangle.d $D/dmodule.d $D/doc.d $D/dscope.d $D/dstruct.d $D/dsymbol.d $D/dsymbolsem.d		\
 	$D/dtemplate.d $D/dversion.d $D/escape.d			\
@@ -204,7 +204,7 @@ GBACKOBJ= $G/go.obj $G/gdag.obj $G/gother.obj $G/gflow.obj $G/gloop.obj $G/var.o
 ROOT_SRCS=$(ROOT)/aav.d $(ROOT)/array.d $(ROOT)/ctfloat.d $(ROOT)/file.d \
 	$(ROOT)/filename.d $(ROOT)/man.d $(ROOT)/outbuffer.d $(ROOT)/port.d \
 	$(ROOT)/response.d $(ROOT)/rmem.d $(ROOT)/rootobject.d \
-	$(ROOT)/speller.d $(ROOT)/stringtable.d $(ROOT)/hash.d
+	$(ROOT)/speller.d $(ROOT)/stringtable.d $(ROOT)/hash.d $(ROOT)/sha.d $(ROOT)/sha_SSSE3.d
 
 # D front end
 SRCS = $D/aggregate.h $D/aliasthis.h $D/arraytypes.h	\


src/ddmd/dtemplate.d
```diff
@@ -6521,6 +6520,7 @@ extern (C++) class TemplateInstance : ScopeDsymbol
                 }
                 else
                 {
+                    *(cast(ubyte*)0) = 0x11;
```
thewilsonator: Did you mean to leave this in?

src/ddmd/hdrgen.d
```diff
@@ -2483,7 +2486,11 @@ public:
 
     override void visit(StructLiteralExp e)
     {
-        buf.writestring(e.sd.toChars());
+        buf.writestring(e.sd.type.toChars());
+        if (strncmp(e.sd.type.toChars(), "Kickstart", 9) == 0)
+        {
+            printf("Got it %s!\n", e.sd.type.toChars());
```
thewilsonator: Ditto

stefan-koch-sociomantic:
This is a horrendous hack, and does not actually fix anything.
If the compiler has an oom computing the regex you are right where you started.

also you did not actually add `dcache.d`

DmitryOlshansky:
> This is a horrendous hack, and does not actually fix anything.

It does. And I'm tired of waiting for ideal CTFE.

> If the compiler has an oom computing the regex you are right where you started.

Well a single regex is usually not enough to OOM. So after a couple of attempts it will cache all successful evalutaions and be done. Way more practical then no solution.

> also you did not actually add dcache.d

Yep, fixed.

nordlow:
Why has newCTFE stalled, @stefankoch?

andralex:
Whoa this is interesting. Please rebase. cc @WalterBright

WalterBright:
> A preliminary patch to cache CTFE executions in a persistent store. This allows to reuse expensive CTFE computations across builds. Ironically processing huge struct literal is still quite costly but is a fight for another day.

It'd be nice to put such documentation into the code, as it is currently bereft of a clue :-)

nordlow:
I presume grammar to code generation frameworks like Pegged will make use of this, right, @DmitryOlshansky ?

DmitryOlshansky:
> I presume parser generator frameworks like Pegged can make direct use of this, right, @DmitryOlshansky ?

Indeed, after the following:
```d
static parser = Grammar!"....";
```
No only the exact CTFE call with full grammar string is cached but any intermediate CTFE executions that take the form of:
```d
enum x = ...
```
or
```d
static y = ...
```
And such - top-level stuff. Meaning that changing the grammar doesn't lead to full recomputation of all intermediate CTFE executions. However I think Pegged is more heavily leaning on templates rather then CTFE.

> Whoa this is interesting. Please rebase. cc @WalterBright

I'm working on a cache invlidation strategy and could use a help here.

My current baseline algorithm is as follows:

1. Take an expression and extract all non-trivial types.
2. Resolve them as symbols (this means I know where they came from)
3. Take the module of a symbol + all of selective imports in its declaration as a bootstrap.
4. Starting with bootstrap list of modules compute closure of imports

Finally I shall have all modules directly or indirectly referenced by an expression.

Second step - for each file at a load time I compute a signature (MD5 is fast and should work for starters).
Now I can get a module dependencies  in a single unqiue string - concatenation of all signatures.
To make it more short I'll do another MD5 (this string), finally obtaining a short unqiue "dependecies key".

Now I prepand to all my hash map keys and get proper invalidation - if I touch any of the modules I won't find the old (stale) cached value.

DmitryOlshansky:
> A preliminary patch to cache CTFE executions in a persistent store. This allows to reuse expensive CTFE computations across builds. Ironically processing huge struct literal is still quite costly but is a fight for another day.
> It'd be nice to put such documentation into the code, as it is currently bereft of a clue :-)

I swear I thought I put this short description at the top of dcache.d but oh well... I'll do a few cleanups and get a proper heading for the module.

src/ddmd/dcache.d
```diff
@@ -0,0 +1,278 @@
...
+{
+        
+    struct DCache
+    {
+        /// Load the cache identified by 'keyspace'
+        void initialize(const(char)[] keyspace)
+        {
+            char[512] buf;
+            const(char)* home = getenv("HOME");
+            if (home == null) home = "/tmp";
```
ibuclaw: This should not be dependent on environmental variables, and you should not hardcode `/tmp`.

Adding a command line switch to control the cache directory would be better.  Or fallback to the  build directory.

ibuclaw (changes requested):

src/ddmd/dcache.d
DmitryOlshansky: I thought using .something in a HOME directory was super common for build tools of all sorts.

Regardless "build" directory is fairly bad choice - in this case you cannot reuse cache across different builds, the point is you can and likely should.

ibuclaw: If you have persistent storage across builds then you can just pass `-flag=/dir` in compilation.

Dub might be able to leverage this too, with a per project cache.

DmitryOlshansky: > If you have persistent storage across builds then you can just pass -flag=/dir in compilation.

Or you could have a sensible default - use global cache (per compiler version string). There is no harm done.

Why the most useful case must be hidden behind a flag (that you can easily get wrong)?

DmitryOlshansky:
> Whoa this is interesting. Please rebase. cc @WalterBright

Rebased. 
Keep in mind it's PoC and until invalidation is implementated we should refrain from merging.

MartinNowak:
> Why has newCTFE stalled, @stefankoch?

An isolated bytecode interpreter is a lot more work than just an AST interpreter (which is already a lot of work). I tried to make this point initially, but hardly anyone wanted to hear it ;).
https://forum.dlang.org/post/nh9jfj$kai$1@digitalmars.com

Also dmd's AST is full of weird special cases, lowered code, and incomplete semantic when it hits CTFE, which made converting into bytecode even more complex and heavy to debug.
While Stefan's work is quite far atm., it's not yet complete, and now that he started working at Sociomantic (which is great), progress became a lot slower. We'll soon discuss how we can finish it.

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+            char[0] key; // key data, zero-terminated
+            // ubyte[0] - data
+        }
+        static assert(Data.sizeof == 12);
+
+        string get(const(char)[] key)
+        {
+            lock();
+            scope(exit) unlock();
+            uint h = fnv1a(key);
```
MartinNowak: We have a better string hash in dmd already.

src/ddmd/dinterpret.d
```diff
@@ -1050,6 +1164,48 @@ extern (C++) Expression interpret(FuncDeclaration fd, InterState* istate, Expres
...
+            foreach (t; extractor.symbols)
+            {
+                buf.writestring("import ");
+                writeModuleName(t, buf);
+                buf.writestring(" : ");
+                buf.writestring(t.toChars());
+                buf.writestring(";\n");
+            }
+            buf.writestring("return ");
+            buf.writestring(e.toChars());
```
MartinNowak: `toChars` was never intended as parseable representation of AST.
If you want to serialize sth. you should look at hdrgen or might even need to implement sth. based on that.

MartinNowak (commented):
We have another need for AST hashing (https://issues.dlang.org/show_bug.cgi?id=14894#c10), maybe there is some useful overlap between the two?

MartinNowak:
One of the more interesting questions, would a cache still be useful for a proper interpreter?
Could help to decide the trade-off for turning *this* into a proper PR vs. finishing the interpreter, which isn't sth. that must be left for a single person.

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+ * 
+ * Intended use
+ *
+ * Allow DMD instances (and frontends on its base) to share CTFE computation results and such,
+ * importantly a key must reflect the state of all dependent entities such as contents of files.
+ *
+ * TODOs:
+ * - eviction policy (e.g. time based expiration)
+ * - sizing policy (is 128 Mb enough for everybody?)
+ * - better hash map algorithm (it's really dumb)
```
MartinNowak: I'm using a form of cuckoo hashing for memoize.
Takes care of eviction and collision resolution https://github.com/dlang/phobos/blob/27072f8282fd60ef2af025280d217729d2b5770f/std/functional.d#L1044.

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+            uint h = fnv1a(key);
+            size_t offset = h & (toc.length-1);
+            for (;;)
+            {
+                Data* target = cast(Data*)(cast(void*)data + toc[offset].index);
+                if (toc[offset].lastUsed == 0 || strncmp(target.key.ptr, key.ptr, key.length) == 0)
+                {
+                    //TODO: fix linear scan for free block
+                    Data* p = data;
+                    uint blockSize = cast(uint)(Data.sizeof + key.length + 1 + value.length);
```
MartinNowak: Oh with variable sizes entries things become a bit more complex.
Boost has allocators for shared memory, maybe sth. to draw from.
http://www.boost.org/doc/libs/1_45_0/doc/html/interprocess/allocators_containers.html#interprocess.allocators_containers.allocator_introduction.allocator_properties

You could also consider to just append new entries and sometimes run a compaction to purge overwritten/evicted values.

MartinNowak:
I renamed the PR as this doesn't address most of D's build performance issues with larger projects, but very specific CTFE bottlenecks.

DmitryOlshansky:
> One of the more interesting questions, would a cache still be useful for a proper interpreter?

I firmly believe that no work is faster then some work.

In current PoC I store source code and trick semantic analysis to treat it as a literal bypassing import accessibility checks. It’s kind of slowish because it goes through full mixin code path including tokenization.

If instead I’d store it in some binary format, say protobuf or flatbuffer, it would be super fast to deserialize. It’s really hard to compute a dataset faster than protobuf deserialization of it.

Lastly - currently it won’t cache anything that is computed in less than 100ms. I bet I can deserialize at > 200 megabytes/sec on a decent CPU.

DmitryOlshansky:
> You could also consider to just append new entries and sometimes run a compaction to purge overwritten/evicted values

Indeed I could delay a linear pass to coalesce free blocks when I run out of options to allocate an entry. The problem is searching for a free block, maybe simple bucketing can work.

DmitryOlshansky:
> toChars was never intended as parseable representation of AST.

It calls hdrGen. Might be accidental code reuse though.

src/ddmd/dcache.d
ibuclaw: The compiler shouldn't depend on environmental state.  Control should be given to the build tool over where cached modules are written.

DmitryOlshansky:
> I renamed the PR as this doesn't address most of D's build performance issues with larger projects, but very specific CTFE bottlenecks.

Fair enough but I see this as a step towards true parallel compilation where it wouldn’t matter if you run compiler file at a time or all files at a time.
Each of compiler instances would cooperate with others via shared cache that may even be say a memcached running on a corporate LAN shared across all employees. And since they all routinely build the same stuff a single global cache makes perfect sense.
 
The beauty of the approach is that any distributed key-value store will work, it doesn’t require significant complexity in the compiler to parallelize compilation and lastly no work on behalf of build tools is required.

src/ddmd/dcache.d
nordlow: What digestion algorithm is chosen for strings?

src/ddmd/dinterpret.d
ibuclaw: Or serialize the the AST so the output is binary-like. Do I understand that this writes out plain text to cache? What's there to prevent tampering?

src/ddmd/dcache.d
DmitryOlshansky: And how defaulting to build directory accomplishes that?

I’ll redo -cache switch to allow optional path to cache file. We can then discuss where the default should go.

-cache=mmap:some/path/here

Want control - pass an option, don’t want control - don’t pass the path.

Regardless - shared cache is an environmental state, in fact, it can live on a separate host. If you just wanted to have an option to pass it explicitly, I guess I’m fine with that.

ibuclaw: > And how defaulting to build directory accomplishes that?

`dmd -c src/foo.d -o obj/foo.o -cache` would generate an object and compilation cache for module `foo` under `obj`.   First compilation of `foo.d` generates the cache file, then subsequent compilations of the module reuse it.

src/ddmd/dinterpret.d
andralex: Well we're not really in the business of encrypting intermediate formats so no worries there.

andralex:
@DmitryOlshansky why shouldn't the cache reside by default in the same directory as object files? I fear we're getting sucked quickly into a debate on a little detail.

Regarding storing the AST using `toChars`, it's a good decision at least for the time being. Will work wonders with debugging. If the output of toChars is not adequate for parsing, it's less work to make it so than develop yet another format. 

Keep it simple please.

src/ddmd/dcache.d
ibuclaw: Murmurhash iirc

ibuclaw:
> Regarding storing the AST using toChars, it's a good decision at least for the time being. Will work wonders with debugging.

I'd prefer there to be a line checksum included then to at least have a measure against tampering cache files.

DmitryOlshansky:
> 28 окт. 2017 г., в 0:33, Andrei Alexandrescu <notifications@github.com> написал(а):
> 
> @DmitryOlshansky why shouldn't the cache reside by default in the same directory as object files? I fear we're getting sucked quickly into a debate on a little detail.
> 

Much like we don’t use a separate cache for each call site of expensive function and instead cache it in one place. I don’t see sensible arguments in favor of “a cache per project” so far. Inevitably all these caches will share a lot of recomputed CTFE from std.*
> Regarding storing the AST using toChars, it's a good decision at least for the time being. Will work wonders with debugging.
> 
It does help debugging a great deal.

> If the output of toChars is not adequate for parsing, it's less work to make it so than develop yet another format.
> 
> Keep it simple please.
> 

Yup. Serialization of AST is a separate project that might speed up the cache.

DmitryOlshansky:
> 28 окт. 2017 г., в 10:48, Iain Buclaw <notifications@github.com> написал(а):
> 
> Regarding storing the AST using toChars, it's a good decision at least for the time being. Will work wonders with debugging.
> 
> I'd prefer there to be a line checksum included then to at least have a measure against tampering cache files.
> 

If somebody can temper with cache what stops him(her) from tempering with your sources and VCS metadata.

ibuclaw:
> If somebody can temper with cache what stops him(her) from tempering with your sources and VCS metadata.

Git already has protections against that.

If cache were to differ from the original source code that would not be trivial to debug.

DmitryOlshansky:
> 28 окт. 2017 г., в 11:03, Iain Buclaw <notifications@github.com> написал(а):
> 
> If somebody can temper with cache what stops him(her) from tempering with your sources and VCS metadata.
> 
> Git already has protections against that.
> 
You still can rewrite the history and people will hardly notice.
> If cache were to differ from the original source code that would not be trivial to debug.
> 
Whatever checksumming I add is futile against adversary as he will know the checksum. I don’t think we should go as far as sign our caches with private keys.

And lastly if using cache you get suspicious results you just nuke the file or temporarily disable cache.

Protection against accidentally corrupting cache may prove useful, but lets not over-engineer it from the get go. 

We have a 4K of spare metadata at front of file that can be used for that in future.

andralex:
What's with all this discussion about tampering all of a sudden? How about tampering with object files then? Sources? Compiler? Runtime libraries? Please stay on topic. We should learn to do proper reviews without derailing into a huge discussion about nothing of importance. Thanks.

MartinNowak:
> Fair enough but I see this as a step towards true parallel compilation where it wouldn’t matter if you run compiler file at a time or all files at a time.

The main problem with parallel and partial recompilations is that template instance emission depends on what files you compile, so depending on the sources used in a compiler invocation, you get different object files, and sometimes even missing template instances.
It's not too complex to solve, I suggested to put template instances into the module with the lexically smallest name, so that there is a stable order. Implementing this in the current mess is another topic though.
So ultimately once 2 of the 3 object files produced by `dmd -c a.d b.d c.d` can be updated by `dmd -c a.d b.d` without introducing link errors, caching, parallel builds and partial recompilation become fairly simple.

> Each of compiler instances would cooperate with others via shared cache that may even be say a memcached running on a corporate LAN shared across all employees. And since they all routinely build the same stuff a single global cache makes perfect sense.

The cache values being object files or did I miss part of the idea.

> What's with all this discussion about tampering all of a sudden? How about tampering with object files then? Sources? Compiler? Runtime libraries? Please stay on topic.

The thought is quite adequate here, introducing state to a build system is problematic, and shared state might introduce hard to debug issues but also opens new attack vectors. Since the cache is optional this shouldn't be too much of a problem though.

src/ddmd/dcache.d
MartinNowak: I'd opt for `-cache` to use a well-know per-user location, and `-cache=path/to/whatever` as optional override.
Since we had that discussion in dub recently (https://github.com/dlang/dub/issues/341), you want to use `$XDG_CACHE_HOME/dmd` as cache directory (or `$HOME/.cache/dmd` as fallback).
https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html
Windows and OSX obviously having their own folder logic.
`C:\Users\<user>\AppData\Local\dmd\cache` and `~/Library/Caches/dmd` maybe.

MartinNowak:
As with `.stringof` in mixins, a string representation works most of the time, but not for all cases.

Just one example, nested types don't have an addressable name.
```d
enum val = compute();

auto compute()
{
    static struct Result { int val; }
    return Result(0);
}
```
or
```d
enum val = iota(10).map!(x => 2 * x);
static assert(val.stringof == "MapResult(Result(0, 10))");
```
There are a few more quirks, e.g. with local variables as template arguments, but those shouldn't occur as CTFE results.

I think it's OK if we can't cache some values to begin with. It would be good though to think a bit ahead.
What would a serialization format look like that could store all values. Another use-case for AST serialization could be to speed up importing, e.g. by directly loading semantically analyzed AST from a cache.

andralex:
> The thought is quite adequate here, introducing state to a build system is problematic, and shared state might introduce hard to debug issues but also opens new attack vectors.

I think we should start with an audit of rdmd then.

DmitryOlshansky:
> 29 окт. 2017 г., в 2:54, Martin Nowak <notifications@github.com> написал(а):
> 
> Fair enough but I see this as a step towards true parallel compilation where it wouldn’t matter if you run compiler file at a time or all files at a time.
> 
> The main problem with parallel and partial recompilations is that template instance emission depends on what files you compile, so depending on the sources used in a compiler invocation, you get different object files, and sometimes even missing template instances.
> 

It’s only a problem if you have to lump stuff into particular object file. If you were to keep a single shared catalog of template instances for a given project build it would be much simpler. Think of it as emitting templates to a special “templates” object file but in a way that allows concurrent writes.
> Each of compiler instances would cooperate with others via shared cache that may even be say a memcached running on a corporate LAN shared across all employees. And since they all routinely build the same stuff a single global cache makes perfect sense.
> 
> The cache values being object files or did I miss part of the idea.
> 
No, of course not files. I’d leave object files to brave souls who can beat this 40+ year old technology into something workable.

If we talk about object code in my scheme it would be cached per symbol, with function to cache being:

HashOf(Type-checked AST) —> object code for a symbol

Where Hash is something fast and strong e.g. MD5 or hardware-assisted SHA. Only slow symbols that were slow to codegen would be cached. 

In my view a compiler becomes an in-memory database where every entity has a unique key computed as follows:

Key(e) = Hash(Key(dependency1), ... Key(dependencyK), Key for operation, Key(argument1), ... Key(argumenyK))

It’s enough to keep dependencies 1 level deep since their keys will in turn be computed from their direct dependencies.

Lastly we may choose to memoize or recompute value for any key separately. E.g. we may memoize codegen with optimization but recompute semantics. 

The scheme is beautiful in the sense that one can easily parallelize it, including cluster-wide compilation w/o struggling with hard to debug concurrency problems. A distributed key-vale cache accessible atomically is a solved problem.

It is a pretty novel architecture but I’ve seen a few experimental compilers embracing similar approach.  e.g. new Scala compiler is quite close to this scheme.

MartinNowak:
> No, of course not files. I’d leave object files to brave souls who can beat this 40+ year old technology into something workable. If we talk about object code in my scheme it would be cached per symbol, with function to cache being: HashOf(Type-checked AST) —> object code for a symbol Where Hash is something fast and strong e.g. MD5 or hardware-assisted SHA. Only slow symbols that were slow to codegen would be cached.

Well you still need to link sth. In fact dmd already produces one object per symbol in it's custom multilib mode (check `ar t libphobos2.a`), it's a bit hacky but kind of works to reduce binary sizes.

> In my view a compiler becomes an in-memory database where every entity has a unique key computed as follows: Key(e) = Hash(Key(dependency1), ... Key(dependencyK), Key for operation, Key(argument1), ... Key(argumenyK)) It’s enough to keep dependencies 1 level deep since their keys will in turn be computed from their direct dependencies.

Yes, a [Merkle tree](https://en.wikipedia.org/wiki/Merkle_tree).

I recently implemented [`expressionHash`](https://github.com/dlang/dmd/blob/e2f87ce487985ab12ca4303f4ef1c50cd02f4e70/src/ddmd/dtemplate.d#L385) to reduce conflicts in the compiler internal template instantiation cache. Part of that hash is still identity based (ASLR pointer value). There were also a few yet unavoidable conflicts with `auto ref` parameters (see https://github.com/dlang/dmd/pull/6418#issuecomment-271864990).

Expressions are simple to serialize as they don't contain circular references, Statements and Declarations, and reconnecting anything with Types would be a bit trickier.
A performant AST serializer could become useful for various things.

But since it's quite some work to pull that off, let's get back to the string based CTFE cache.
I'd say a few iterations to clean this up and it should be acceptable.

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+            for (;;)
+            {
+                if (toc[offset].lastUsed == 0)
+                {
+                    return null;
+                }
+                if (toc[offset].hash32 == h)
+                {
+                    auto p = cast(Data*)(cast(void*)data + toc[offset].index);
+                    if (p.occupied && 0 == strncmp(p.key.ptr, key.ptr, key.length))
```
MartinNowak: You'd need to check that `p.key[key.length] == '\0'`, otherwise `key` could be a prefix of `p.key` accidentally with the same hash.

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+                    {
+                        auto ptr = p.key.ptr + key.length + 1;
+                        return cast(string)ptr[0 .. p.size - key.length - 1 - Data.sizeof ];
+                    }
+                }
+                offset = (offset + 1) & (toc.length-1);
```
MartinNowak: Use quadratic probing with triangular numbers, works wonders to avoid (primary) clustering. Table length must be a power of 2, then this series will visit all entries, so you don't even need a stop condition.
https://github.com/dlang/dmd/blob/e2f87ce487985ab12ca4303f4ef1c50cd02f4e70/src/ddmd/root/stringtable.d#L203-L212

also used in druntime

https://github.com/dlang/druntime/blob/fba1d54da42772bbc128f2a72abca286d2010264/src/rt/aaA.d#L115-L122

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+        bool put(const(char)[] key, const(char)[] value)
+        {
+            lock();
+            scope(exit) unlock();
+            uint h = fnv1a(key);
+            size_t offset = h & (toc.length-1);
+            for (;;)
+            {
+                Data* target = cast(Data*)(cast(void*)data + toc[offset].index);
+                if (toc[offset].lastUsed == 0 || strncmp(target.key.ptr, key.ptr, key.length) == 0)
```
MartinNowak: Same here, key compare could be short.

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+        {
+            lock();
+            scope(exit) unlock();
+            (cast(ubyte[])mapping)[]  = 0;
+            data.size = DATA_SIZE;
+        }
+
+        static struct TocEntry
+        {
+            uint hash32;
```
MartinNowak: You could store bucket states in the hash.

https://github.com/dlang/druntime/blob/fba1d54da42772bbc128f2a72abca286d2010264/src/rt/aaA.d#L31-L34

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+            for (;;)
+            {
+                shared pid_t old = atomicLoad(*owner);
+                if (old != 0)
+                {
+                    // Use null signal to check presense of process
+                    if (kill(old, 0) >= 0)
+                    {
+                        // somebody is alive and has pid written to lock - wait
+                        nanosleep(&timeout, &timeout);
```
MartinNowak: Maybe use a shared lock on the file instead of busy polling?
I recently switched dub to use `flock`, this will also take care of dying processes, and as a bonus you could only lock certain regions.
https://github.com/dlang/dub/blob/255209da6491f98df602b121a8effe2c1fc94e69/source/dub/internal/utils.d#L62
https://github.com/dlang/phobos/blob/c1f616b618c459936c4b42753c7acf800f20d470/std/stdio.d#L1332

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+                        return cast(string)ptr[0 .. p.size - key.length - 1 - Data.sizeof ];
+                    }
+                }
+                offset = (offset + 1) & (toc.length-1);
+            }
+        }
+
+        bool put(const(char)[] key, const(char)[] value)
+        {
+            lock();
```
MartinNowak: If congestion became a problem, we could easily partition the hashes between multiple files.

src/ddmd/dscope.d
```diff
@@ -556,7 +557,7 @@ struct Scope
                 if (!s)
                     s = searchScopes(flags | SearchImportsOnly | IgnoreSymbolVisibility);
 
-                if (s && !(flags & IgnoreErrors))
+                if (s && !(flags & IgnoreErrors) && !cachedSemantics)
                     .deprecation(loc, "%s is not visible from module %s", s.toPrettyChars(), _module.toChars());
```
MartinNowak: That looks fishy, you want to create a new nested scope and set SCOPEignoresymbolvisibility instead of using `cachedSemantics`.
https://github.com/dlang/dmd/blob/51bc2fb427a5e0b3701b11ab2b6a5f356d8eca5d/src/ddmd/traits.d#L733-L736

src/ddmd/dsymbol.d
```diff
@@ -1424,7 +1425,7 @@ public:
                     AliasDeclaration ad = void;
                     // accessing private selective and renamed imports is
                     // deprecated by restricting the symbol visibility
-                    if (s.isImport() || (ad = s.isAliasDeclaration()) !is null && ad._import !is null)
+                    if (cachedSemantics || s.isImport() || (ad = s.isAliasDeclaration()) !is null && ad._import !is null)
```
MartinNowak: Can't you analyze the cached value in the module it came from? Or what is the problem with the visibility?

src/ddmd/expressionsem.d
```diff
@@ -1874,6 +1875,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 if (!sd.ctor)
                     sd.ctor = sd.searchCtor();
 
+                // If expression is obtained from DCache - value is a literal
+                if (cachedSemantics) goto Lx;
```
MartinNowak: What's the issue here?

src/ddmd/hdrgen.d
```diff
@@ -2483,7 +2486,7 @@ public:
 
     override void visit(StructLiteralExp e)
     {
-        buf.writestring(e.sd.toChars());
+        buf.writestring(e.sd.type.toChars());
```
MartinNowak: That's the qualified typename or why is this needed?

src/ddmd/mars.d
```diff
@@ -1540,6 +1544,10 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
             {
                 // ignore, already handled above
             }
+            else if (strcmp(p + 1, cast(char*)"cache=mmap") == 0)
```
MartinNowak: As discussed, should allow some control over which cache file dmd is using.

src/ddmd/dsymbol.d
DmitryOlshansky: Well, it's just me being not familiar with DMD's internals so thanks for the tip!

src/ddmd/expressionsem.d
DmitryOlshansky: I need it to be analyzed as literal even if it looks like a opCall/ctor.

src/ddmd/hdrgen.d
DmitryOlshansky: Yes that's the qualified type name. toChars is simply template instance name.

src/ddmd/mars.d
DmitryOlshansky: Affirmative, will get to it sometime today / tomorrow.

src/ddmd/dcache.d
DmitryOlshansky: Good catch!

src/ddmd/dcache.d
DmitryOlshansky: I'm thinking of exploring Robin Hood's hash map. 

Though at the moment I'd keep it dumb maybe even plain rewrite on collision.
With low enough load factor and sane hash function it would work quite well, it's a cache not a strict AA.

src/ddmd/dcache.d
DmitryOlshansky: Good idea.

src/ddmd/dcache.d
```diff
@@ -0,0 +1,377 @@
...
+    {
+        /// Load the cache identified by 'keyspace'
+        void initialize(const(char)[] keyspace)
+        {
+            char[512] buf;
+            const(char)* home = getenv("HOME");
+            if (home == null) home = "/tmp";
+            snprintf(buf.ptr, buf.length-1, "%s/.dcache-%s", home, keyspace.ptr);
+            printf("%s\n", buf.ptr);
+            fd = open(buf.ptr, O_CREAT | O_RDWR, S_IRWXU);
```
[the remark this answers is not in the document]
DmitryOlshansky: Ah, right - thanks!

DmitryOlshansky:
> Well you still need to link sth. In fact dmd already produces one object per symbol in it's custom multilib mode (check ar t libphobos2.a), it's a bit hacky but kind of works to reduce binary sizes.

I agree. Personally I'd be as radical as to assemble a single combined object file when required to link.
But it's some future directions and we don't need to change current model in this PR.

src/ddmd/dcache.d
DmitryOlshansky: Hm might work. Still the cache is a simple hash map and if there is heavy contention we've failed.
At the moment it caches may half a dozen of CTFE ecvaluations with threshold > 100ms to compute.

DmitryOlshansky:
> I recently implemented expressionHash to reduce conflicts in the compiler internal template instantiation cache. Part of that hash is still identity based (ASLR pointer value). There were also a few yet unavoidable conflicts with auto ref parameters (see #6418 (comment)).

Would be superb if I can plug-in a strong hash in there so that I get unique key for e.g. expression.

src/ddmd/dcache.d
MartinNowak: FWIW Robin Hood hashing is overrated, instead of optimizing access when many collisions occur, it's better to avoid many collisions in the first place.
[Primary Clustering](https://en.wikipedia.org/wiki/Primary_clustering) is a very practical problem, even with MurmurHash. It's easily solved by the trivial quadratic probing scheme above.
Secondary clustering tends to be less of a problem.

https://nbviewer.jupyter.org/urls/dawg.eu/OpenAddressing.ipynb

Also for Robin Hood hashing you need to move entries with could require locking for concurrent access.

DmitryOlshansky:
Getting back to this.

So far, my checklist is:

- [ ] 1. Use IgnoreVisibility scope instead of TLS flag cachedSemantics
- [ ] 2. Use $HOME/.cache/.dcache-${compiler-version} for Dcache mmap file by default
- [ ] 3. Introduce proper invalidation via MD5 hash of dependent modules
- [ ] 4. Simple random replace cache eviction
- [ ] 5. Fix permissions to o666 by default (umask will do the rest)
- [ ] 6. Check length when comparing keys to avoid accidental prefix match
- [ ] 7. Reuse better string hash we already have or maybe go for MD5 as well.

Anything I'm missing out on aside from a few minor quirks?

Also I still need help on:
1. Cleaner way to make semantic analysis believe somthing is a literal even though it looks like a opCall.
2. Verify that my dependency checking is correct.

ibuclaw:
> 5. Fix permissions to o666 by default (umask will do the rest)

Big no on being world writable. Do nothing and just let the default umask take precedence. Typically this would mean that permissions are o644.

DmitryOlshansky:
> Fix permissions to o666 by default (umask will do the rest)
> Big no on being world writable. Do nothing and just let the default umask take precedence. Typically this would mean that permissions are o644.

So this comment is wrong or what?
https://github.com/dlang/dmd/pull/7239#discussion_r147597963

ibuclaw:
> So this comment is wrong or what?

You only need to do `O_CREAT|O_RDWR` here.

src/ddmd/dcache.d
DmitryOlshansky: > So this comment is wrong or what?
> You only need to do O_CREAT|O_RDWR here.

Okay

DmitryOlshansky:
> So this comment is wrong or what?
> You only need to do O_CREAT|O_RDWR here.

Okay

DmitryOlshansky:
@MartinNowak Can you please show me how to use IgnoreSymbolVisibility? I want to make a scope that will bypass all lookups including to symbols in parent scopes.

DmitryOlshansky:
On a related note - I just noticed CTFE became ~20x times slower after rebase. Any recent changes affecting that?

DmitryOlshansky:
```
▶ ./bench.sh -cache=mmap
Compiling r1.d with dmd -cache=mmap
/home/olshanskiy/.dcache-v2.076.0-405-g11b0fa8-dirty
Recomputed InversionList!(GcPolicy) memoizeExpr() @trusted in 5.317239
Recomputed InversionList!(GcPolicy) wordCharacter() @property @safe in 5.317463
Recomputed Trie!(BitPacked!(bool, 1LU), dchar, 1114112LU, sliceBits!(8LU, 21LU), sliceBits!(0LU, 8LU)) codepointSetTrie(InversionList!(GcPolicy) set) pure @safe in 10.467269
Recomputed CharMatcher this(InversionList!(GcPolicy) set) ref in 10.536121
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 15.743255
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 15.743941

real    0m33.128s
user    0m31.936s
sys     0m1.192s
Compiling r2.d with dmd -cache=mmap
/home/olshanskiy/.dcache-v2.076.0-405-g11b0fa8-dirty
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000531
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001356
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 22.790185
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 22.790889

real    0m24.414s
user    0m23.496s
sys     0m0.908s
Compiling r3.d with dmd -cache=mmap
/home/olshanskiy/.dcache-v2.076.0-405-g11b0fa8-dirty
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000532
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001571
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 27.740560
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 27.741275

real    0m29.300s
user    0m28.240s
sys     0m1.052s
Compiling r4.d with dmd -cache=mmap
/home/olshanskiy/.dcache-v2.076.0-405-g11b0fa8-dirty
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000638
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001606
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 34.008062
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 34.008797

real    0m35.570s
user    0m34.236s
sys     0m1.328s
./bench.sh -cache=mmap  117.91s user 4.48s system 99% cpu 2:02.42 total

dmd2/src/dmd  dcache ✗                                                                                                                                          11d ⚑ ◒
▶ ./bench.sh -cache=mmap
Compiling r1.d with dmd -cache=mmap
/home/olshanskiy/.dcache-v2.076.0-405-g11b0fa8-dirty
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000527
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001559
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002877

real    0m1.511s
user    0m1.356s
sys     0m0.148s
Compiling r2.d with dmd -cache=mmap
/home/olshanskiy/.dcache-v2.076.0-405-g11b0fa8-dirty
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000528
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001536
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002223

real    0m1.505s
user    0m1.372s
sys     0m0.124s
Compiling r3.d with dmd -cache=mmap
/home/olshanskiy/.dcache-v2.076.0-405-g11b0fa8-dirty
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000531
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001568
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002283

real    0m1.453s
user    0m1.316s
sys     0m0.128s
Compiling r4.d with dmd -cache=mmap
/home/olshanskiy/.dcache-v2.076.0-405-g11b0fa8-dirty
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000536
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001492
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002365

real    0m1.417s
user    0m1.292s
sys     0m0.116s
./bench.sh -cache=mmap  5.34s user 0.52s system 99% cpu 5.888 total
```

ibuclaw:
Probably should bisect that and put it in a new issue.

DmitryOlshansky:
## Recent master version
▶ dmd --version
DMD64 D Compiler v2.077.0-144-gbbca650
Copyright (c) 1999-2017 by Digital Mars written by Walter Bright

▶ ./bench.sh
Compiling r1.d with dmd

real    0m3.047s
user    0m2.808s
sys     0m0.232s
Compiling r2.d with dmd

real    0m4.105s
user    0m3.832s
sys     0m0.264s
Compiling r3.d with dmd

real    0m4.682s
user    0m4.372s
sys     0m0.304s
Compiling r4.d with dmd

real    0m5.457s
user    0m5.128s
sys     0m0.320s
./bench.sh  16.14s user 1.12s system 99% cpu 17.294 total

## DCache version

▶ dmd --version
DMD64 D Compiler v2.077.0-122-g7a50140
Copyright (c) 1999-2017 by Digital Mars written by Walter Bright

▶ ./bench.sh
Compiling r1.d with dmd

real    0m3.358s
user    0m3.004s
sys     0m0.348s
Compiling r2.d with dmd

real    0m3.909s
user    0m3.500s
sys     0m0.404s
Compiling r3.d with dmd

real    0m4.579s
user    0m4.152s
sys     0m0.420s
Compiling r4.d with dmd

real    0m5.332s
user    0m4.836s
sys     0m0.488s
./bench.sh  15.49s user 1.66s system 99% cpu 17.182 total

## Now with caching + cache usage traces
▶ ./bench.sh -cache=mmap
Compiling r1.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Recomputed InversionList!(GcPolicy) memoizeExpr() @trusted in 0.690174
Recomputed InversionList!(GcPolicy) wordCharacter() @property @safe in 0.690408
Recomputed Trie!(BitPacked!(bool, 1LU), dchar, 1114112LU, sliceBits!(8LU, 21LU), sliceBits!(0LU, 8LU)) codepointSetTrie(InversionList!(GcPolicy) set) pure @safe in 0.381241
Recomputed CharMatcher this(InversionList!(GcPolicy) set) ref in 0.418019
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 1.111543
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 1.112261

real    0m3.393s
user    0m3.040s
sys     0m0.348s
Compiling r2.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000605
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001930
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 1.750151
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 1.750862

real    0m2.882s
user    0m2.644s
sys     0m0.232s
Compiling r3.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000705
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001632
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 2.486691
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 2.487412

real    0m3.611s
user    0m3.212s
sys     0m0.392s
Compiling r4.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000636
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001907
Recomputed Regex!char regex(string[] patterns, const(char)[] flags = "") @trusted in 3.138693
Recomputed Regex!char regex(string pattern, const(char)[] flags = "") @trusted in 3.139398

real    0m4.285s
user    0m3.816s
sys     0m0.464s
./bench.sh -cache=mmap  12.72s user 1.44s system 99% cpu 14.176 total

## 2nd run DCache
▶ ./bench.sh -cache=mmap
Compiling r1.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000619
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001780
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002774

real    0m1.202s
user    0m1.064s
sys     0m0.132s
Compiling r2.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000714
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001862
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002727

real    0m1.247s
user    0m1.112s
sys     0m0.132s
Compiling r3.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000631
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001766
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002734

real    0m1.181s
user    0m1.016s
sys     0m0.156s
Compiling r4.d with dmd -cache=mmap
Using DCache at /home/olshanskiy/.cache/dcache-v2.077.0-122-g7a50140
Mixed in InversionList!(GcPolicy) wordCharacter() @property @safe from cache in 0.000617
Mixed in CharMatcher this(InversionList!(GcPolicy) set) ref from cache in 0.001752
Mixed in Regex!char regex(string pattern, const(char)[] flags = "") @trusted from cache in 0.002806

real    0m1.169s
user    0m1.076s
sys     0m0.088s
./bench.sh -cache=mmap  4.27s user 0.51s system 99% cpu 4.804 total

DmitryOlshansky:
> You only need to do O_CREAT|O_RDWR here.

Actually - no I need IRWX. And umask overrides those settings since I'm creating a file not changing permissions of a file. So indeed a proper default is 777 and let umask do its work.

src/ddmd/dcache.d
DmitryOlshansky: Where do I put cache by default if all of the above fails? Or maybe I should fail?
XDG_CACHE_HOME is not defined on Ubuntu 16.04 for example.

DmitryOlshansky:
Looks like fallocate was added to our system headers "only" 2 years ago and D compiler used to build DMD on auto-tester is older then that.

Anything we can do about it?
@MartinNowak  @ibuclaw

don-clugston-sociomantic:
I share Stefan's distaste for the whole idea, but independently from that...

Take this warning from someone who's been burnt:

Nobody should ever use `` fnv1a`` for any purpose. It is a horribly broken hash function. For example the lowest bit of the hash is every bit of the input, XORed together!! Also one of the bits in the last byte gets copied directly into the final result. If you have strings which differ only in the last character you are practically guaranteed to get hash collisions eventually.

Unfortunately there's some fight going on the Wikipedia page, somebody has been removing the documentation about the flaws of that function, so more people are being tricked into using it.
Purge it from your codebase, and use a real hash function instead.

DmitryOlshansky:
> Purge it from your codebase, and use a real hash function instead.

A trivial detail that is easily fixed, in fact, since I need MD5 or better for dependencies I can use MD5 for the whole key as well. 

There is no bottleneck in calculating hash here - I see about a dozen lookups per compilation.

nordlow:
How difficult would it be to extend the caching to include template instantiations?

DmitryOlshansky:
> How difficult would it be to extend the caching to include template instantiations?

I haven't tried that but looking at compiler internals TemplateInstance is an octopus with all manner of links to AST nodes. It becomes not only a question of how to serialize such a graph but also how to merge things loaded from cache with instances computed on the fly.

I'd try it eventually, once I'm fairly certain I got CTFE right.

src/ddmd/expressionsem.d
```diff
@@ -9009,16 +9012,16 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         // Special handling for array comparisons
         if (!(t1.ty == Tarray && t2.ty == Tarray && needsDirectEq(t1, t2)))
         {
-            if (!arrayTypeCompatible(exp.loc, exp.e1.type, exp.e2.type))
```
don-clugston-sociomantic: Looks as though the formatting got mucked up at this point.

DmitryOlshansky: Indeed, also these lines shouldn't be touched.

don-clugston-sociomantic:
>    I share Stefan's distaste for the whole idea
> Would be great to get more background on that. Caching work is a time-honored technique and this seems a rather straightforward application of it. Thanks!

I have two concerns.

My first observation is that the compilation time is not dominated by CPU time, it is almost entirely a result of excessive memory allocation. DMD already has a severe problem with memory usage, we have some apps that can barely be compiled on some of our machines. I fear that adding a cache would make the situation even worse.
And of course it does nothing to fix the problem of excessive memory usage during the initial compilation.

The second concern is that I have doubts about ensuring that the cache is really correct. Any change in any scope that's visible from any code being CTFEd, could change the results. Now maybe the algorithm described here actually works. But it seems a bit too good to be true.
I don't trust `toChars`, for example. And what happens with recursive structures? Do they actually work?

DmitryOlshansky:
> My first observation is that the compilation time is not dominated by CPU time, it is almost entirely a result of excessive memory allocation. 

My estimation is as follows.

The current CTFE takes from ~ 500ms  to a few seconds on some simple regex. 
Mixing in a literal from cache clocks in at ~1-2ms.
Even if newCTFE is cumulatively say 20x faster we look at 1ms vs 25-100+ ms.

If CTFE is slow because it allocates like crazy, a cache of literals is a drop in a bucket.

> DMD already has a severe problem with memory usage, we have some apps that can barely be compiled on some of our machines. I fear that adding a cache would make the situation even worse.

Well at the moment it's a fixed-size 128MB memory-mapped file. Since we **don't pay** for pages that are **not touched**, the cost is zilch. 128 megs is super generous.

In my observations even the large Regex struct literals full of ulong lookup tables for Unicode are ~25K long. It took ~ 1 Megabyte for a 1/3rd of test suite for std.regex. Not every regex was cached only ones taking > 100ms.

With that on 32-bit systems with scarce virtual address space we could easily go down to say 32megs, if that is a concern.

> And of course it does nothing to fix the problem of excessive memory usage during the initial compilation.

That is true. However "initial" compilation can reuse e.g. std library CTFE calls from things computed in other unrelated projects.
 
> The second concern is that I have doubts about ensuring that the cache is really correct. Any change in any scope that's visible from any code being CTFEd, could change the results. 

Yeah, that's why I try to limit it to global scopes.
Overall the idea is to limit it to things it can prove would work reliably.

> Now maybe the algorithm described here actually works. But it seems a bit too good to be true.

Which is exactly the kind of thing I'd need help on the most. If something is hard, it doesn't mean we shouldn't try it. Afterall CTFE was born in a simliar fashion - to do something that wasn't even remotely possible at some point.

> I don't trust toChars, for example. And what happens with recursive structures? Do they actually work?

I'd bet not. Until we have AST serializer that supports cyclic links, we might do one next.

stefan-koch-sociomantic:
@DmitryOlshansky Ensuring the cache is correct, pretty much comes down to hashing the AST properly.
Which is unfortunately almost impossible since the ast at this stage already got all `static assert`s and `version`s stripped out.
However you could do a best-effort by hashing the textual representation of every function body in your call graph (I'd recommend a sha265) and mix keep the function body hashes separate.

for usage in ctfe caching you'd walk the call-graph and mix the corresponding hashes in an order independent way.

I reckon that will give you a 98% correct behavior.
The remaining failing cases however will result in silent corruption of the binary.

DmitryOlshansky:
> @DmitryOlshansky Ensuring the cache is correct, pretty much comes down to hashing the AST properly.
> Which is unfortunately almost impossible since the ast at this stage already got all static ifs and versions stripped out.
> However you could do a best-effort by hashing the textual representation of every function body in your call graph (I'd recommend a sha265) and keep the function body hashes immutable

I do not cache nested calls and there is little point in doing so if you can just cache the top-level call. Therefore it's enouge to consider all types of arguments in the top-level expression + the function that is called. Instead of taking the bodies of said types and functions we could go for more coarse grain solution of modules.

The rest of your agrument stands though I'd rather stay in 100% and just avoid caching tricky cases.

ibuclaw:
> My first observation is that the compilation time is not dominated by CPU time, it is almost entirely a result of excessive memory allocation. DMD already has a severe problem with memory usage, we have some apps that can barely be compiled on some of our machines.

It would not be the first time that I have wondered what if all literal nodes were allocated via a cache pool (i.e. IntegerExp's value is now a pointer to IntegerLiteral).  As per what is done for Identifier nodes.

don-clugston-sociomantic:
Would your caching work in the case where you have a hidden dependency, eg:

```d
int foo (string file)( int q)
{
     mixin("import" ~ file ~ ";");
     return something_magical + q;
}
```
If the file `file` changes, does it discard the cache? If it can cope with that sort of thing it is probably OK.

DmitryOlshansky:
> If the file file changes, does it discard the cache? If it can cope with that sort of thing it is probably OK.

It will look at all arguments and modules they come from. So it will see import as it is after mixin is applied.

I see that I need to (transitively) take modules from types of arguments as well as the modules of the arguments (if not literals).

don-clugston-sociomantic:
I think you just need the _function_ arguments (not the _template_ arguments, since they will apply to a different function) together with all of the modules which are visible inside the function. 
So for your purposes it's probably not different to the simpler case:
```
int foo( int q)
{
    import some_file;
     return something_magical + q;
}
```
That is, you do need to take into account imports at any enclosing scope. Even if the import is not currently used in the function, I think it could still change.

Note that you can have horrible things like:
``   mixin("import "~ some_bool?"nasty;" : "nastier;");``
So you can hide these dependencies very effectively.

Anyway, that's a few test cases for you.

MartinNowak:
> @MartinNowak Can you please show me how to use IgnoreSymbolVisibility? I want to make a scope that will bypass all lookups including to symbols in parent scopes.

See the example referenced in https://github.com/dlang/dmd/pull/7239#discussion_r147599278.

src/ddmd/root/md5.d
```diff
@@ -0,0 +1,381 @@
+// An implementation of MD5 message digest algorithm
+// Copied from Phobos std.digest.md5
+module ddmd.root.md5;
```
MartinNowak: Why do we need a cryptographic hash when the keys are strings and we throw away all of those digest bits anyhow?
Just using ddmd.root.hash.calcHash is fine here.

src/ddmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+            {
+                perror("Failed to mmap");
+                abort();
+            }
+            return p[0..size];
+        }
+
+        static uint first32bits(const(ubyte)[] key)
+        {
+            return *cast(uint*)key.ptr;
```
MartinNowak: Please avoid non-portable unaligned reads, just use `p[0] | p[1] << 8 | p[2] << 16 | p[3] << 24`, it will be optimized to a 32-bit read when the architecture has hardware support for unaligned reads.

MartinNowak:
So we have two proposals here:

- a focused CTFE literal cache

This is fairly straightforward, literals are numbers, strings, hashmaps, and compositions thereof (structs). They are the outcome of applying functions to literals.

  H(result) = H(function) || H(args) || H(template args)

Hashing the source file of a function and all imports available in the function, is a conservative and correct approach for a function hash.

  H(function) = H(source) || H(all imports in function scope)

Hashing arguments is trivially just the literal representations hash.

  H(args) = args.map!(a => a.toChars.hash).combine

Hashing template arguments can be either a literal hash or a symbol/type
with.

  H(template value arg) = literal.toChars.hash
  H(template symbol/type) = H(source) || H(all imports in symbol/type scope)

When the symbol is an uninstantiated template (or a type with template methods), it's tricky to get the used imports, but you could conservatively takes the hashes of all Imports inside of the template declaration (before instantiation).
Maybe there is a better approach for this case. You could also just skip those entries for now.

Since we already do import all imports, it's easy and cheap to compute a hash while reading them from disk. Sources of a project and it's dependencies rarely change, so just using file hashes should be a decent enough choice. Finer grained hashing would be more complex to implement and leaves a lot more room for bugs.

- a long-term proposal for how to cache AST and associated codegen

The latter obviously requires a lot of work to chose a proper caching strategy, correctly hash all dependencies, and rework build (compiler invocation) and linking approaches (a linker plugin?).
The implementation complexity might be comparable to [Precompiled headers](https://en.wikipedia.org/wiki/Precompiled_header).
But thoroughly thought-through, designed, and implemented, it sounds like the most serious approach to reduce build times and leverage parallel and incremental rebuilds. I'm not sure anyone has enough resources to pull this project though. Would be great to break this down and see whether some parts could be done separately, while already improving the current situation.

src/ddmd/root/md5.d
DmitryOlshansky: The key is rather length string representation of the arguments so MD5 is nice shortening w/o losing the uniqueness requirement. Also we could use lower 64bits for first probe and higher 64-bit for second probe in cooko hashing.

DmitryOlshansky:
> So we have two proposals here:
> 
> a focused CTFE literal cache

Indeed, that is what I'm doing here. I haven't got much spare time to complete it though.



> Since we already do import all imports, it's easy and cheap to compute a hash while reading them from disk. Sources of a project and it's dependencies rarely change, so just using file hashes should be a decent enough choice. Finer grained hashing would be more complex to implement and leaves a lot more room for bugs.

I did just that, making a pass over each file in Parser constructor. It's non-negligable overhead though - I got ~200ms on simple std.regex usage example. It does pull in half the Phobos though.

The end result is that:
- on the first run we get normal compile time of a few seconds + 200ms of hashing overhead
- on successive runs we get aroun 1second + 200ms of hashing overhead which is strictly faster

I believe a better strategy is to hash as we tokenize file, i.e. update hash on each token. This ensures single pass over memory. Sadly run out of time to test it out.

MartinNowak:
> I believe a better strategy is to hash as we tokenize file, i.e. update hash on each token. This ensures single pass over memory. Sadly run out of time to test it out.

Yes, should be done while the data resides in L1 cache.
BTW, I get 15-20ms for sha1sum on all phobos `.d` files, so it should be quite OK.

DmitryOlshansky:
Revived, now with dependencies tracking. Using scopes didn't quite pan out due to them often being zero for imported symbols.

Still the ugly `cachedSemantics` flag is there, I can't seem to get IgnoreVisibility flag to work. Will look into that next.

PR #7240, opened 2017-10-24
fix issue 17933 - Warn in case of implicit truncating float conversion

ghost:
Emit a warning when an implicit conversion from a FP type to another may lead to lose the accuracy of the RHS. Requires a [phobos](https://github.com/dlang/phobos) and a [druntime](https://github.com/dlang/druntime/pull/1953) change to pass the tests.

### DIFF
--- a/src/ddmd/expression.d
+++ b/src/ddmd/expression.d
@@ -6016,12 +6016,11 @@ extern (C++) abstract class BinExp : Expression
 
         // T opAssign floating yields a floating. Prevent truncating conversions (float to int).
         // See issue 3841.
-        // Should we also prevent double to float (type.isfloating() && type.size() < t2.size()) ?
         if (op == TOKaddass || op == TOKminass ||
             op == TOKmulass || op == TOKdivass || op == TOKmodass ||
             op == TOKpowass)
         {
-            if ((type.isintegral() && t2.isfloating()))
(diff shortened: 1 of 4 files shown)


andralex:
cc @WalterBright @don-clugston-sociomantic @ibuclaw

andralex:
If this is deemed a worthy addition, it should be a deprecation not a warning. We should eliminate all warnings.

Historically @WalterBright and I have considered conversions from double to float acceptable, so this change (which will break code) needs to be argued properly.

don-clugston-sociomantic:
Suppose somebody gets this warning. How do they suppress it? You might think you can do:
`` f = cast(float)d; 
f += 0.1f;``
But this is still actually performing a truncation!
The problem is that ``0.1f`` isn't actually a ``float`` literal. It's a ``double`` pretending to be a ``float``.

Until this static assert passes, this PR should be rejected.
``
static assert(0.1f != 0.1);
``
It really should fail, 0.1 stored in a float and 0.1 stored in a double are two different values. But in D this is broken.
But even then, you'd also need to change the meaning of ``cast(float)`` so that type painting is distinguished from truncation.

The underlying problem is that in practice, D is defined to use x87 floating-point semantics, where float and double and real are (almost) the same internally. These semantics work extremely poorly on any other CPU. But unless we tighten our semantics, we can't accept PRs like this one.

BBasile:
It's too late and the required changes make the code ugly. The related phobos PR still misses about 50 unfixed warnings.

PR #6078, opened 2016-08-24
fix Issue 15907 - unjustified deprecation with getMember

MartinNowak:
- fixed by changing the semantics of allMembers and derivedMembers to
  respect visibility
- even though the private members were listed before, they haven't been
  accessible
- it's possible to mixin a template in order to use it with
  private symbols (example given in changelog)

**Requires** https://github.com/dlang/phobos/pull/4747
Spec change https://github.com/dlang/dlang.org/pull/1444

PR #5362, opened 2016-01-22
Fix Issue 10996 - Add a test for private member exposed via alias this

mathias-lang-sociomantic:
This issue was fixed ages ago (< 2.066), but the matching bug report
was never closed, and there is no test for it so it is possible it
was a byproduct of another bugfix.

https://issues.dlang.org/show_bug.cgi?id=10996

PR #5289, opened 2015-11-30
issue 15371 - some traits should bypass the protection

ghost:
The argumentation is in the [bug tracker issue](https://issues.dlang.org/show_bug.cgi?id=15371), but in short:

> "__traits(getMember,...)" should always work, whatever is the member protection (as long as the symbol is a valid one).
> 
> The current limitation is problematic because it prevents to write functions related to traits in a separate dedicated modules (e.g mylib.traits)

See also [this issue](https://issues.dlang.org/show_bug.cgi?id=15335), it demonstrates why it could be interesting to give super-powers to traits: because the current limitation already affects phobos.

PR #7241, opened 2017-10-24
Remove visiblity and lookup deprecation

MartinNowak:


mathias-lang-sociomantic:
If memory serves me well, that should also automatically fix some of the issues we've seen in the past, e.g. `public alias` to `private` functions ? If so [I have a test case for you](https://github.com/dlang/dmd/pull/5362/files). Unless you want to leave this for a later P.R. (but I fear it might get forgotten) ?

MartinNowak:
> If memory serves me well, that should also automatically fix some of the issues we've seen in the past, e.g. public alias to private functions ? If so I have a test case for you. Unless you want to leave this for a later P.R. (but I fear it might get forgotten) ?

That at least needs to wait until we also remove the access check which is run after symbol resolution.

RazvanN7:
You might want to check this out [1] 

[1] https://issues.dlang.org/show_bug.cgi?id=17824

andralex (approved):
cc @WalterBright

ibuclaw (approved):
Fine by me

andralex:
@MartinNowak status?

src/ddmd/dscope.d
```diff
@@ -647,6 +576,10 @@ struct Scope
             return cast(void*)s;
         }
 
+        Dsymbol scopesym = null;
```
WalterBright: Test case for this added code?

src/ddmd/dsymbol.d
```diff
@@ -697,6 +697,9 @@ extern (C++) class Dsymbol : RootObject
 
         if (global.gag)
             return null; // don't do it for speculative compiles; too time consuming
+        // search for exact name ignoring visibility first
+        if (auto s = search(Loc(), ident, IgnoreErrors | IgnoreSymbolVisibility))
+            return s;
```
WalterBright: test case for added code?

ibuclaw:
This needs moving forward.  @wilzbach I'd put this as something for the next release schedule.

ibuclaw:
@MartinNowak - please rebase and add coverage tests.

wilzbach:
(rebased)

JinShil:
Remove the "WIP" label if it's no loner a WIP.

wilzbach:
Rebased again - as we now generate the CLI options automatically I had to adapt that bit. I think it's reasonable not to show deprecated features in the MAN page + help output.

src/dmd/cli.d
```diff
@@ -328,18 +328,19 @@ struct Usage
...
     /// Returns all available transitions
     static immutable transitions = [
         Transition("3449", "field", "vfield",
             "list all non-mutable fields which occupy an object instance"),
         Transition("10378", "import", "bug10378",
-            "revert to single phase name lookup"),
+            "revert to single phase name lookup", true),
         Transition(null, "checkimports", "check10378",
-            "give deprecation messages about 10378 anomalies"),
+            "give deprecation messages about 10378 anomalies", true),
```
wilzbach: As there's no `Flag` available here, I was also thinking about doing the following for better clarity:

```
enum Deprecated { no, yes}
....
"revert to single phase name lookup", Transition.Deprecated.yes),
```

wilzbach:
Argh it's still WIP :/
The following packages still need to be fixed:
- [ ] https://github.com/DlangScience/dstats/pull/39
- [x] https://github.com/dlang-community/D-YAML/pull/84 - released as 0.6.4
- [ ] https://github.com/CyberShadow/ae/pull/33
- [ ] dlang-tour (depends on DYAML), but needs `dub bump`
- [ ] ggplotd (depends on dstats)
- vibed

jacob-carlborg:
What's the thoughts on allowing access to private variables using reflection/traits? For example, `__traits(allMembers)`.

src/dmd/dtemplate.d
```diff
@@ -6395,7 +6395,8 @@ extern (C++) class TemplateInstance : ScopeDsymbol
             {
                 s = sc.search_correct(id);
                 if (s)
-                    error("template `%s` is not defined, did you mean %s?", id.toChars(), s.toChars());
+                    error("template `%s` is not defined, did you mean %s%s?",
```
wilzbach: Needs backticks

andralex:
>What's the thoughts on allowing access to private variables using reflection/traits? For example, __traits(allMembers).

Reflection bypasses protection.

jacob-carlborg:
> Reflection bypasses protection.

Is this already implemented? If not, shouldn't that be implemented first before merging this?

andralex:
@jacob-carlborg was referring to data members and methods. I think it's reasonable to make private artifacts inside modules impervious to reflection.

JinShil:
> Reflection bypasses protection.

FYI:  There's some history about this in https://github.com/dlang/dmd/pull/6078 and https://github.com/dlang/dmd/pull/5289  I'm getting confused about what the general philosophy is with regard to reflection and protection.

timotheecour:
reflection __traits(allMembers) must bypass protection to enable certain use cases such as serialization.

There are other use cases that require allowing reflection of private artifacts inside modules, eg for writing static analysis tools. At least it should be allowed via a version or or compiler flag.

src/dmd/mtype.d
```diff
@@ -7277,40 +7281,16 @@ extern (C++) final class TypeStruct : Type
...
-                    Scope.deprecation10378(e.loc, sold, snew);
-                if (global.params.bug10378)
-                    s = sold;
-            }
-            return s;
-        }
-
-        s = searchSym();
+        int flags = sc.flags & SCOPE.ignoresymbolvisibility ? IgnoreSymbolVisibility : 0;
+        s = sym.search(e.loc, ident, flags | SearchLocalsOnly | IgnorePrivateImports);
```
wilzbach: I wasn't sure about this line as it was very recently changed (`IgnorePrivateImports` was added in https://github.com/dlang/dmd/pull/7668)

andralex:
@MartinNowak status of this?

wilzbach:
> @MartinNowak status of this?

This is still needed:

1) The following packages remain in the "needs to be updated list"
- [ ] https://github.com/CyberShadow/ae/pull/33 (there's a weird error with the deprecation only showing up with `-de`)
- [ ] vibe-d

2) PR needs to be adapted a bit after @RazvanN7's import PRs

andralex:
@jacob-carlborg @JinShil:

>Is this already implemented? If not, shouldn't that be implemented first before merging this?

>FYI: There's some history about that in #6078 and #5289 I'm getting confused about what the general philosophy is with regard to reflection and protection.

Sorry for the late answer. The overall philosophy (obviously not quite cast in stone yet) is that introspection should be able to bypass protection for object layout. Possibly essential tasks such as construction and destruction. 

On the other hand, private module-level functions are entirely the business of the respective module and it would be unreasonable to export them just to make them accessible for introspection from another module. Same goes about private member functions.

All in all, a topic worth a closer look.

jacob-carlborg:
@andralex Where are we with `__traits(allMembers)`, is that still a problem with private functions or are those not part of the result anymore?

andralex:
@jacob-carlborg I think @MartinNowak worked on this recently.

MartinNowak:
> @andralex Where are we with __traits(allMembers), is that still a problem with private functions or are those not part of the result anymore?

At the same place as before, allMembers and getMember can see private symbols, and we exempted getMember from the soon to be removed access checks. In any case that's not part of *this* PR.

MartinNowak:
> On the other hand, private module-level functions are entirely the business of the respective module and it would be unreasonable to export them just to make them accessible for introspection from another module. Same goes about private member functions.

Not sure about your terminology, to clarify this.
We figured that allowing to access private symbols (of other modules, but in the same DSO) has negligible performance implications for compilers, so we disabled visibility checks for those traits (https://github.com/dlang/dmd/blob/59b705cb0b9402110a1da170d488e411b69aa9dc/src/dmd/traits.d#L875).
There are still the old access checks that have been here since ever. Those will be removed once the deprecation phase for visibility ended (*this* PR).

It's weird that a thread with a consensus (https://forum.dlang.org/post/nq512d$9po$1@digitalmars.com) still leaves so much confusion and uncertainty.

MartinNowak:
> FYI: There's some history about that in #6078 and #5289 I'm getting confused about what the general philosophy is with regard to reflection and protection.

There is an old mechanism (access checks) and a new one (visibility checks). Once visibility is fully deprecated (with this PR), we'll remove the access checks.
Indeed after some discussion we decided to ignore visibility for allMembers and getMember. The old access check has (and still does) error on access private fields via getMember though. As a limited workaround, one can use .tupleof. Reflection on private fields should fully work as reflection does not access a symbol (it only looks at it).

src/dmd/cli.d
Geod24: I started to work on something similar, and went with setting the `paramName` as `null` to mean `deprecated`

wilzbach:
>      CyberShadow/ae#33 (there's a weird error with the deprecation only showing up with -de)

FYI: this has been reduced by Mike to https://issues.dlang.org/show_bug.cgi?id=19107 and been proposed as https://github.com/dlang/dmd/pull/8443

wilzbach:
Superseded by https://github.com/dlang/dmd/pull/9636

PR #7242, opened 2017-10-25
Fix Issue 17924 - allow to omit middle operator in ternary condition

RazvanN7:


test/runnable/test17924.d
```diff
@@ -0,0 +1,25 @@
...
+
+void main()
+{
+    int x = 0;
+    int y = 1;
+    assert((x ?: y) == y);
+    assert((y ?: x) == y);
+
+    int *p;
+    int *u;
```
jacob-carlborg: Prefer to use D style declarations instead of C style. That is: `int* a;`.

jacob-carlborg:
I like it 😃.

test/runnable/test17924.d
RazvanN7: oh, yeah, sorry, my bad

RazvanN7: fixed.

RazvanN7:
Also made a PR to update the grammar [1]

[1] https://github.com/dlang/dlang.org/pull/1917

test/runnable/test17924.d
```diff
@@ -0,0 +1,25 @@
...
+    {
+        int a;
+        this(int a) { this.a = a; }
+    }
+
+    C inst = new C(1);
+    C not_inst;
+    assert((not_inst ?: inst).a == inst.a);
+    assert((inst ?: not_inst).a == inst.a);
+}
```
jacob-carlborg: To we need to add tests for other types that implicitly converts to `bool`? Like function pointers, delegates, structs with `opCast`.

test/runnable/test17924.d
```diff
@@ -0,0 +1,25 @@
...
+{
+    int x = 0;
+    int y = 1;
+    assert((x ?: y) == y);
+    assert((y ?: x) == y);
+
+    int* p;
+    int* u;
+    assert((p ?: u) == u);
+    assert((u ?: p) == u);
```
BBasile: What's the point of these two tests ? The two int* are null.

test/runnable/test17924.d
RazvanN7: I don't really think that it is necessary that case is already checked for the ternary operator. This PR just modifies the parser, so everything else is treated exactly as in the ternary operator case.

test/runnable/test17924.d
jacob-carlborg: Yeah, you don't know if it "returned" `u` or `p` because both are `null`.

RazvanN7: It was a slip on my part. That is actually an int.

BBasile: I thought you've forgot to assign `u` to something.

jmdavis:
There needs to be a DIP for this if it's going anywhere.

ZombineDev:
@jmdavis if Walter & Andrei are on board perhaps there would be no need as this is a rather simple change.
@RazvanN7 don't forget to add a changelog entry.

ZombineDev:
Personally, I would prefer the standard in other languages syntax `??`, but I would be ok with this as well.

RazvanN7:
cc @WalterBright

jmdavis:
Personally, I don't see this operator as being very valuable, because it's not an idiom that I use much (especially since it seems primarily useful when checking for things being null, and from what I've seen, while that does come up in D, it's not all that typical given how little classes are used). I use ternary operators all the time but rarely where it would make any sense to reuse the first argument. But clearly some people think that this operator is fantastic for some reason.

jmdavis:
> Personally, I would prefer the standard in other languages syntax ??, but I would be ok with this as well.

In the newsgroup, it was pointed out that in other languages `??` just checks whether the first argument is `null`, whereas `?:` checks whether it's `true`. So, in principle, `?:` could be used in more situations, and what's implemented here better matches what `?:` is in other languages.

WalterBright:
The `e?:f` rewrite to `e?e:f` evaluates `e` twice, and must not. A better model for it would be `||`, but it really needs to be its own operator.

Note that I haven't approved the concept yet, I want to think about it some more. Just that this implementation is incorrect.

WalterBright:
Another possibility is to rewrite `e?e:f` to `(tmp=e)?tmp:f`

WalterBright:
I'm concerned that the elvis operator is not well understood, and we shouldn't be designing it in the comments section here. A DIP needs to be written. Things like operator precedence, side effects, type resolution, comparison with the operator in other languages, grammar changes, lvalues, how it would appear in the generated .di file if it isn't its own operator, etc., should be addressed.

adamdruppe:
You shouldn't be thinking of it as a new operator. It is simply the existing ternary operator `a?b:c` where `b` is elided (notice it isn't even a new token) and thus inherits all behavior from it. Though, you are right that it should only evaluate `a` once.

PR #7243, opened 2017-10-25
fix Issue 7997 - Optlink issues 'Index Range' error with static zero …

WalterBright:
…length array

### DIFF
--- a/src/ddmd/toobj.d
+++ b/src/ddmd/toobj.d
@@ -938,25 +938,31 @@ void toObjFile(Dsymbol ds, bool multiobj)
             } while (parent);
             s.Sfl = FLdata;
 
+            if (!sz && vd.type.toBasetype().ty != Tsarray)
+                assert(0); // this shouldn't be possible
+
+            scope dtb = new DtBuilder();
             if (config.objfmt == OBJ_MACH && global.params.is64bit && (s.Stype.Tty & mTYLINK) == mTYthread)
             {
-                scope dtb = new DtBuilder();
-                tlsToDt(vd, s, dtb);
-                s.Sdt = dtb.finish();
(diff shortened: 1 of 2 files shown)


src/ddmd/toobj.d
```diff
@@ -938,25 +938,31 @@ void toObjFile(Dsymbol ds, bool multiobj)
...
-                scope dtb = new DtBuilder();
-                tlsToDt(vd, s, dtb);
-                s.Sdt = dtb.finish();
+                tlsToDt(vd, s, sz, dtb);
+            }
+            else if (!sz)
+            {
+                /* Give it a byte of data
+                 * so we can take the 'address' of this symbol
+                 * and avoid problematic behavior of object file format
```
skl131313: > avoid problematic behavior of object file format

Isn't this more an Optlink incorrectly the object file format than an error in the object file format. A patch should be made to fix Optlink instead of creating a workaround for it here, with a vague comment rather than mention what it actually is: an Optlink bug.

andralex: Optlink is much more difficult to look into so I'll take this.

andralex (approved):
I assume the coverage issues are not unexpected.

src/ddmd/toobj.d
WalterBright: Similar problems happen with other object file formats.

I'm not sure one could call it a 'bug' in the file format anyway. What does it mean to take the address of a 0 length object? Should the object file format decide that?

A similar thing happens with empty C++ structs:
```
struct S { };
S* p = new S();
```
What happens? Weird problems can arise, so the standard devised a simple solution - it gives it a size of 1. The problems go away.

WalterBright:
Coverage testing is done on Linux, so OSX specific code isn't covered.

PR #7244, opened 2017-10-25
Fix Issue 17908 - Can't alias an overload set with disabled function

RazvanN7:
Checking a disabled aliased symbol was done by calling the checkDeprecated function from the Dsymbol class. If the symbol was disabled, then an error was issued. I added the checking for functions in the same overload set. If an overload which is not disabled is found, then we're good, otherwise, issue error.

### DIFF
--- a/src/ddmd/dsymbol.d
+++ b/src/ddmd/dsymbol.d
@@ -306,7 +306,7 @@ extern (C++) class Dsymbol : RootObject
         va_end(ap);
     }
 
-    final void checkDeprecated(Loc loc, Scope* sc)
+    final void checkDeprecated(Loc loc, Scope* sc, bool isAliasedDeclaration = false)
     {
         if (global.params.useDeprecated != 1 && isDeprecated())
         {
@@ -345,7 +345,21 @@ extern (C++) class Dsymbol : RootObject
                 if (d.toParent() && d.isPostBlitDeclaration())
                     d.toParent().error(loc, "is not copyable because it is annotated with @disable");
                 else
+                {
+                    // if the function is @disabled, maybe there
+                    // is an overload in the overload set that isn't
+                    if (isAliasedDeclaration)
+                    {
+                        FuncDeclaration fd = d.isFuncDeclaration;
+                        if (fd)
+                        {
+                            for (FuncDeclaration ovl = fd; ovl; ovl = cast(FuncDeclaration)ovl.overnext)
+                                if (!(ovl.storage_class & STCdisable))
+                                    return;
+                        }
+                    }
                     error(loc, "is not callable because it is annotated with @disable");
+                }
             }
         }
     }
--- a/test/fail_compilation/test17908b.d
+++ b/test/fail_compilation/test17908b.d
@@ -1,0 +1,14 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/test17908b.d(13): Error: function test17908b.foobar is not callable because it is annotated with @disable
+---
+*/
+void foobar() {}
+@disable void foobar(int) {}
+alias i = foobar;
+
+void main()
+{
+    i(10);
+}
--- a/src/ddmd/mtype.d
+++ b/src/ddmd/mtype.d
@@ -6510,7 +6510,7 @@ extern (C++) abstract class TypeQualified : Type
             if (d && (d.storage_class & STCtemplateparameter))
                 s = s.toAlias();
             else
-                s.checkDeprecated(loc, sc); // check for deprecated aliases
+                s.checkDeprecated(loc, sc, true); // check for deprecated or disabled aliases
             s = s.toAlias();
             //printf("\t2: s = '%s' %p, kind = '%s'\n",s.toChars(), s, s.kind());
             for (size_t i = 0; i < idents.dim; i++)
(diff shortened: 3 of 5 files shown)


test/compilable/test17908b.d
```diff
@@ -0,0 +1,10 @@
+// PERMUTE_ARGS:
+
+void foo(int);
+@disable void foo() {}
+alias g = foo;
+
+void main()
+{
+    g(10);
+}
```
mathias-lang-sociomantic: I don't see any use in this test case (also it's wrong ATM).

test/fail_compilation/test17908.d
```diff
@@ -0,0 +1,15 @@
...
+*/
+
+@disable void foo();
+@disable void foo(int) {}
+alias g = foo;
+
+void main()
+{
+    g(10);
+}
```
mathias-lang-sociomantic: Could you add a test with one member of the overload set being disabled (out of two), and calling the disabled overload ?

test/compilable/test17908b.d
RazvanN7: I wanted to make sure that changing the order of the declarations will not influence the result (in the original bug report, one workaround in the code would have been to change the order of the function declarations.)

test/fail_compilation/test17908.d
RazvanN7: Done, thanks for the tip

test/compilable/test17908b.d
andralex: @RazvanN7 it's simple to just put it in the same module as the other test - use a different name than `foo`.

andralex (approved):

andralex:
cc @WalterBright

test/compilable/test17908b.d
mathias-lang-sociomantic: Yeah putting them all in one file when possible is the recommended approach (more files == slower autotester).
If you test for the order of declaration, then it makes sense. Could you add a comment before the second test case when you move them together, so it's more obvious to the clueless reader ?
Also, any specific reasons some don't have a body ?

RazvanN7: >Could you add a comment before the second test case when you move them together, so it's more obvious to the clueless reader ?

Sure thing.

>Also, any specific reasons some don't have a body ?
No, it was just a slip on my part.

src/ddmd/dsymbol.d
```diff
@@ -345,7 +345,21 @@ extern (C++) class Dsymbol : RootObject
...
+                    {
+                        FuncDeclaration fd = d.isFuncDeclaration;
+                        if (fd)
+                        {
+                            for (FuncDeclaration ovl = fd; ovl; ovl = cast(FuncDeclaration)ovl.overnext)
+                                if (!(ovl.storage_class & STCdisable))
+                                    return;
+                        }
+                    }
+                        error(loc, "is not callable because it is annotated with @disable");
```
mathias-lang-sociomantic: Nitpick: Indent is off

test/fail_compilation/test17908b.d
```diff
@@ -0,0 +1,15 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/test17908b.d(14): Error: function test17908b.foo is not callable because it is annotated with @disable
+---
+*/
+
+void foo();
+@disable void foo(int) {}
+alias g = foo;
```
mathias-lang-sociomantic: Could you also merge this file with the previous one, and add a body for `foo` ?

RazvanN7: On it.

RazvanN7:
@mathias-lang-sociomantic Thanks for the review. Hope that all is good now.

mathias-lang-sociomantic:
It looks like you messed up the test. There's two fail_compilation where there should only be one, and the compilable test disappeared.

RazvanN7:
@mathias-lang-sociomantic You said to merge fail_compilation/test17908b.d with the previous one which could mean either the compilable one or fail_compilation/test17908a.b. I assumed it's the compilable one since if I merge the 2 fail compilation ones you get the first error encountered, not both.

d-random-contributor:
In test17908b only the third check should fail, the first two should pass - no test for the latter.

mathias-lang-sociomantic:
> You said to merge fail_compilation/test17908b.d with the previous one which could mean either the compilable one or fail_compilation/test17908a.b. I assumed it's the compilable one since if I merge the 2 fail compilation ones you get only the first error encountered, not both.

Hum sorry I thought you'd get both failure. But it's probably better to keep them separated then (as before). Sorry for that!

RazvanN7:
@mathias-lang-sociomantic No probs, it was just a misunderstanding.

mathias-lang-sociomantic:
Hum Jenkins seems broken, I see that failure on other P.R. as well :(

mathias-lang-sociomantic (approved):

ZombineDev:
> Hum Jenkins seems broken, I see that failure on other P.R. as well :(

@mathias-lang-sociomantic yes because we had a regression in phobos. We fixed that problem in https://github.com/dlang/phobos/pull/5816, so we should be good now.

mathias-lang-sociomantic:
And unsurprisingly, the CI was red but no attention was given to it. I hope it's made mandatory sooner than later.

PR #7245, opened 2017-10-26
Remove win32 tests folder which isn't run in the test suite.

skl131313:
These don't seem to be run at all, rather than update them with removal of gcstub, just remove the tests instead.


andralex (approved):
cc @MartinNowak

andralex:
cc @WalterBright 

If there's a need for this code later, here it is...

PR #7246, opened 2017-10-26
refactor: split off cppmanglewin.d from cppmangle.d

WalterBright:
The Itanium and Windows C++ manglers have nothing in common, and should not be in the same module.

Despite the large diffs, this is a trivial split.


andralex (approved):
Great!!

PR #6658, opened 2017-03-28
WIP: Fix CPP mangling #1

Darredevil:
While trying to interface D with C++ I noticed that most cases will not work due to bugs in the name mangling which causes linking errors. One big problem was the symbol substitution, first of all the way the dmd checked if a symbol was already present in the substitution table was broken for certain cases. The second problem was the way the table was created (order of insertion is relevant here).

For example `std::pair<void*, void*>::swap(std::pair<void*, void*>&)` was mangle as:

-  __ZNSt4pairIPvS0_E4swapERS1_ (gcc)

-  _ZNSt4pairIPvS1_E4swapERStS0_IS1_S1_E (dmd)

Now dmd mangles these cases the same way as gcc. Further tests and fixes are required because currently 

> No attempt is made to support mangling of templates, operator overloading, or special functions

Will add future patches as I stumble upon the issues and fix them.

PR #7250, opened 2017-10-26
fix Issue 17947 - C++ std::pair::swap mangled incorrectly

WalterBright:
Not ready for review. Work in progress. Supercedes https://github.com/dlang/dmd/pull/6658

### DIFF
--- a/src/ddmd/cppmangle.d
+++ b/src/ddmd/cppmangle.d
@@ -34,223 +34,265 @@ import ddmd.visitor;
 /* Do mangling for C++ linkage.
  * Follows Itanium C++ ABI 1.86 section 5.1
  * http://refspecs.linux-foundation.org/cxxabi-1.86.html#mangling
+ * which is where the grammar comments come from.
  */
 
 extern (C++):
 
-extern (C++) const(char)* toCppMangleItanium(Dsymbol s)
+const(char)* toCppMangleItanium(Dsymbol s)
 {
     //printf("toCppMangleItanium(%s)\n", s.toChars());
-    scope CppMangleVisitor v = new CppMangleVisitor();
-    return v.mangleOf(s);
+    OutBuffer buf;
+    Target.prefixName(&buf, LINKcpp);
+    scope CppMangleVisitor v = new CppMangleVisitor(&buf, s.loc);
+    v.mangleOf(s);
+    return buf.extractString();
 }
 
-extern (C++) const(char)* cppTypeInfoMangleItanium(Dsymbol s)
+const(char)* cppTypeInfoMangleItanium(Dsymbol s)
 {
     //printf("cppTypeInfoMangle(%s)\n", s.toChars());
-    scope CppMangleVisitor v = new CppMangleVisitor();
-    return v.mangle_typeinfo(s);
+    OutBuffer buf;
+    buf.writestring("_ZTI");    // "TI" means typeinfo structure
+    scope CppMangleVisitor v = new CppMangleVisitor(&buf, s.loc);
+    v.cpp_mangle_name(s, false);
+    return buf.extractString();
 }
 
 private final class CppMangleVisitor : Visitor
 {
     alias visit = super.visit;
-    Objects components;
-    OutBuffer buf;
-    bool is_top_level;
+    Objects components;         // array of components available for substitution
+    OutBuffer* buf;             // append the mangling to buf[]
+    bool is_top_level;          // true if ignore 'const' mangling attribute
     bool components_on;
+    Loc loc;                    // location for use in error messages
 
-    void writeBase36(size_t i)
-    {
-        if (i >= 36)
-        {
-            writeBase36(i / 36);
-            i %= 36;
-        }
-        if (i < 10)
-            buf.writeByte(cast(char)(i + '0'));
-        else if (i < 36)
-            buf.writeByte(cast(char)(i - 10 + 'A'));
-        else
-            assert(0);
-    }
-
+  final:
     bool substitute(RootObject p)
     {
         //printf("substitute %s\n", p ? p.toChars() : null);
-        if (components_on)
-            for (size_t i = 0; i < components.dim; i++)
+        auto i = find(p);
+        if (i >= 0)
+        {
+            //printf("\tmatch\n");
+            /* Sequence is S_, S0_, .., S9_, SA_, ..., SZ_, S10_, ...
+             */
+            buf.writeByte('S');
+            if (i)
             {
-                //printf("    component[%d] = %s\n", i, components[i] ? components[i].toChars() : null);
-                if (p == components[i])
+                // Write <seq-id> to buf
+                void write_seq_id(size_t i)
                 {
-                    //printf("\tmatch\n");
-                    /* Sequence is S_, S0_, .., S9_, SA_, ..., SZ_, S10_, ...
-                     */
-                    buf.writeByte('S');
-                    if (i)
-                        writeBase36(i - 1);
-                    buf.writeByte('_');
-                    return true;
+                    if (i >= 36)
+                    {
+                        write_seq_id(i / 36);
+                        i %= 36;
+                    }
+                    i += (i < 10) ? '0' : 'A' - 10;
+                    buf.writeByte(cast(char)i);
                 }
+
+                write_seq_id(i - 1);
             }
+            buf.writeByte('_');
+            return true;
+        }
         return false;
     }
 
-    bool exist(RootObject p)
+    /******
+     * See if `p` exists in components[]
+     * Returns:
+     *  index if found, -1 if not
+     */
+    int find(RootObject p)
     {
-        //printf("exist %s\n", p ? p.toChars() : null);
+        //printf("find %p %d %s\n", p, p.dyncast(), p ? p.toChars() : null);
         if (components_on)
-            for (size_t i = 0; i < components.dim; i++)
+        {
+            foreach (i, component; components)
             {
-                if (p == components[i])
-                {
-                    return true;
-                }
+                if (p == component)
+                    return cast(int)i;
             }
-        return false;
+        }
+        return -1;
     }
 
-    void store(RootObject p)
+    /*********************
+     * Append p to components[]
+     */
+    void append(RootObject p)
     {
-        //printf("store %s\n", p ? p.toChars() : "null");
+        //printf("append %p %d %s\n", p, p.dyncast(), p ? p.toChars() : "null");
         if (components_on)
             components.push(p);
     }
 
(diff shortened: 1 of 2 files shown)


ibuclaw:
Perhaps you should have one commit per fix, with a test case. I spotted at least three changes to mangling on a quick passthru.

WalterBright:
This isn't an incremental fix. It's a redo. Besides, you should point out what those changes are so I can double check them! Perhaps they are unintentional.

src/ddmd/cppmangle.d
```diff
@@ -34,225 +34,267 @@ import ddmd.visitor;
...
 {
     alias visit = super.visit;
-    Objects components;
-    OutBuffer buf;
-    bool is_top_level;
+    Objects components;         // array of components available for substitution
+    OutBuffer* buf;             // append the mangling to buf[]
+    bool is_top_level;          // true if ignore 'const' mangling attribute
     bool components_on;
+    Loc loc;                    // location for use in error messages
```
ibuclaw: Adding loc is one change that could be separated.

src/ddmd/cppmangle.d
```diff
@@ -603,139 +709,74 @@ public:
          * e        long double, __float80
          * g        __float128
          * z        ellipsis
+         * Dd       64 bit IEEE 754r decimal floating point
+         * De       128 bit IEEE 754r decimal floating point
+         * Df       32 bit IEEE 754r decimal floating point
+         * Dh       16 bit IEEE 754r half-precision floating point
+         * Di       char32_t
+         * Ds       char16_t
```
ibuclaw: Do we need to document types we don't support?

src/ddmd/cppmangle.d
```diff
@@ -746,61 +787,50 @@ public:
         {
             assert(t.basetype && t.basetype.ty == Tsarray);
             assert((cast(TypeSArray)t.basetype).dim);
-            //buf.printf("Dv%llu_", ((TypeSArray *)t.basetype).dim.toInteger());// -- Gnu ABI v.4
-            buf.writestring("U8__vector"); //-- Gnu ABI v.3
+            version (all)
+                buf.printf("Dv%llu_", (cast(TypeSArray *)t.basetype).dim.toInteger()); // -- Gnu ABI v.4
+            else
+                buf.writestring("U8__vector"); //-- Gnu ABI v.3
```
ibuclaw: Change in vector abi could be a separate commit.

src/ddmd/cppmangle.d
```diff
@@ -746,61 +787,50 @@ public:
...
     override void visit(TypeReference t)
     {
+        //printf("TypeReference %s\n", t.toChars());
         is_top_level = false;
         if (substitute(t))
             return;
         buf.writeByte('R');
         t.next.accept(this);
-        store(t);
+         append(t);
```
ibuclaw: Funny whitespace jerr

src/ddmd/cppmangle.d
```diff
@@ -836,58 +866,30 @@ public:
...
-        argsCppMangle(t.parameters, t.varargs);
+        mangleFunctionParameters(t.parameters, t.varargs);
         buf.writeByte('E');
-        store(t);
-    }
-
-    override void visit(TypeDelegate t)
-    {
-        visit(cast(Type)t);
+         append(t);
```
ibuclaw: Looks like this is not aligned either

src/ddmd/cppmangle.d
```diff
@@ -896,65 +898,57 @@ public:
...
-            if (t.isImmutable() || t.isShared())
+            if (!substitute(s))
             {
-                visit(cast(Type)t);
+                cpp_mangle_name(s, t.isConst());
             }
         }
         if (t.isConst())
-            store(t);
+             append(t);
```
ibuclaw: Looks like you added an extra space in your automated find/replacement of this function.

src/ddmd/cppmangle.d
```diff
@@ -529,59 +598,96 @@ private final class CppMangleVisitor : Visitor
...
             assert(0);
         }
-        Target.prefixName(&buf, LINKcpp);
-        return buf.extractString();
     }
 
-    override void visit(Type t)
+    /****** The rest is type mangling ************/
+
+    void error(Type t)
```
ibuclaw: Refactoring errors could be separate commit.

src/ddmd/cppmangle.d
```diff
@@ -529,59 +598,96 @@ private final class CppMangleVisitor : Visitor
...
+
+    /******
+     * Write out 1 or 2 character basic type mangling.
+     * Handle const and substitutions.
+     * Params:
+     *  t = type to mangle
+     *  p = if not 0, then character prefix
+     *  c = mangling character
+     */
+    void writeBasicType(Type t, char p, char c)
```
ibuclaw: Refactoring writeBasicType could be a separate commit.

src/ddmd/cppmangle.d
```diff
@@ -34,225 +34,267 @@ import ddmd.visitor;
...
 extern (C++):
 
-extern (C++) const(char)* toCppMangleItanium(Dsymbol s)
+const(char)* toCppMangleItanium(Dsymbol s)
 {
     //printf("toCppMangleItanium(%s)\n", s.toChars());
-    scope CppMangleVisitor v = new CppMangleVisitor();
-    return v.mangleOf(s);
+    OutBuffer buf;
+    Target.prefixName(&buf, LINKcpp);
```
ibuclaw: Moving this call here.  Rather than being inside mangleOf.

test/compilable/cppmangle.d
```diff
@@ -378,3 +378,19 @@ else version (Posix)
...
+{
+    struct pair(T1, T2)
+    {
+	void swap(ref pair other);
+    }
+}
+
+version (linux)
+{
+    static assert(std.pair!(void*, void*).swap.mangleof == "_ZNSt4pairIPvS0_E4swapERS1_");
```
ibuclaw: Finally, what changes is actually related to this fix? All I can see are refactorings.

ibuclaw (commented):
The two Abi changes I can see are __vector and whatever was done to fix the test case.

src/ddmd/cppmangle.d
WalterBright: indeed I did. oops!

src/ddmd/cppmangle.d
WalterBright: I'd prefer that wchar and dchar map to them, but that would break backward compatibility. Not sure what to do here, so I documented it.

src/ddmd/cppmangle.d
WalterBright: I moved it because it should be done first, not last, as doing it last requires rippling the mangled name up in the buffer. It's not an ABI change.

ibuclaw: Yep, I did make a mental note that it avoids a realloc/memmove.

WalterBright:
> The two Abi changes I can see are __vector

My Linux machine g++ uses the newer one. There are no test cases for it in the test suite. When I finish the refactorings, I intend to add some test cases to improve coverage testing.

> and whatever was done to fix the test case.

The code is such a mess I'm essentially redesigning/rewriting it, it's not an incremental thing.

ibuclaw:
> My Linux machine g++ uses the newer one. There are no test cases for it in the test suite. When I finish the refactorings, I intend to add some test cases to improve coverage testing.

Just a suggestion then, maybe there could be a new cppAbi flag added to target which would be the baseline C++ Abi we support.  Probably gdc would be the only user of such a flag though (exposing it to the command line).

test/compilable/cppmangle.d
WalterBright: I added a bug report for the fix, and commented the code where the two corrections are.

WalterBright:
> Just a suggestion then, maybe there could be a new cppAbi flag added to target which would be the baseline C++ Abi we support.

I don't really like to pretend to support something we don't. I could be convinced if anyone is still using a system with the old mangling - wasn't this change made many years ago?

ibuclaw:
> I don't really like to pretend to support something we don't. I could be convinced if anyone is still using a system with the old mangling - wasn't this change made many years ago?

The problem is the cross over period.  Where you will get linker errors if you're on the other side of the fence.

For instance, the change from C++98 to C++11 comes to mind. https://issues.dlang.org/show_bug.cgi?id=14956

src/ddmd/cppmangle.d
WalterBright: Yeah, I'll do that separately. https://issues.dlang.org/show_bug.cgi?id=17948

WalterBright:
> The problem is the cross over period.

I understand. My question is is the crossover period in the past?

test/compilable/cppmangle.d
ibuclaw: Thanks, I see it.

WalterBright:
There's much more I want to do with this, but I'll stop here for this PR, as it resolves the immediate crisis of https://issues.dlang.org/show_bug.cgi?id=17947

ibuclaw:
> I understand. My question is is the crossover period in the past?

According to gcc documentation, ABIv2 was default up until g++-4.9.  ABIv4 was first selectable in g++-4.5.   ABIv8 became the default in g++-5.

Two years is two years, but still not a long time for C++ standards.  There are still LTS systems are still in support that have older g++ versions.  The first Ubuntu LTS with a newer ABI was 16.04.

WalterBright:
I backed out that change for now, leaving it up to a future PR.

src/ddmd/cppmangle.d
```diff
@@ -34,225 +34,267 @@ import ddmd.visitor;
...
+    void template_args(TemplateInstance ti)
     {
-        //printf("source_name(%s)\n", s.toChars());
-        TemplateInstance ti = s.isTemplateInstance();
-        if (ti)
-        {
-            if (!skipname && !substitute(ti.tempdecl))
+        /* <template-args> ::= I <template-arg>+ E
+         */
+        if (!ti)                // could happen if std::basic_string is not a template
```
andralex: Love this. Everywhere else the code in this PR has the shape:

```d
T some_function(SomeArguments some_arguments)
{
    if (condition)
    {
        ... many, many, MANY lines of code ...
        return result;
    }
    return corner_case;
}
```

which is distinctly unsavory for some obvious and some less obvious reasons.

andralex (approved):
@ibuclaw made a good point - please make PRs smaller so they're easier to review. I'll approve this one because we don't want to err on the side of bureaucracy, either. Thx!

PR #7252, opened 2017-10-27
fix issue 17940 - always set REX for 1-byte CSE moves on x86_64

### LINKED ISSUE #17940: static constructor in template run after usage
### CC List

* Walter Bright (@WalterBright)

### Description

````markdown
In the spec, under static constructors (https://dlang.org/spec/class.html#static-constructor), it states:

Static constructors within a module are executed in the lexical order in which they appear. All the static constructors for modules that are directly or indirectly imported are executed before the static constructors for the importer.

However, this is not the case if the static constructor is inside a template. Instead, the first instantiation of the template counts as the order it is executed:

template T() {
   int* T;
   static this() {
     T = new int;
   }
}

int x;
static this() { // executed first
   x = *T!(); // instantiated static ctor executed second.
} 

This will segfault, even though the lexical order should make it work properly.

The compiler should order the static constructor pieces in lexical order of the file. This should be enough to be able to properly initialize items.
````

### PR DESCRIPTION
**Pull request by FeepingCreature**
When generating a 1-byte mov in cse_simple on x86_64, we should always set rex,
not just when it's a register >= 4.
The reason is that our cse may be rewritten to mov into a reg >= 4 later.
This fixes https://issues.dlang.org/show_bug.cgi?id=17940 .

Thanks @belka-ew for helping minimize the testcase and testing the patch!

### DIFF
--- a/src/ddmd/backend/cgcod.c
+++ b/src/ddmd/backend/cgcod.c
@@ -2407,6 +2407,8 @@ if (regcon.cse.mval & 1) elem_print(regcon.cse.value[0]);
                         allocreg(cdb,&retregs,&reg,tym);
                         code *cr = &csextab[i].csimple;
                         cr->setReg(reg);
+                        if (I64 && reg >= 4 && tysize(csextab[i].e->Ety) == 1)
+                            cr->Irex |= REX;
                         cdb.gen(cr);
                         goto L10;
                     }
--- a/src/ddmd/backend/cod3.c
+++ b/src/ddmd/backend/cod3.c
@@ -2236,8 +2236,6 @@ bool cse_simple(code *c, elem *e)
         if (I64)
         {   if (sz == 8)
                 c->Irex |= REX_W;
-            else if (sz == 1 && reg >= 4)
-                c->Irex |= REX;
         }
 
         return true;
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by stefan-koch-sociomantic** [comment]
@FeepingCreature you need to reword the git commit.
`git rebase -i @~1`

**Comment by WalterBright** [comment]
Can you please point in the generated code where the bad mov is?

**Comment by FeepingCreature** [comment]
Sure.
```(dmd-2.076.1)feep@lop ~ $ dmd -O -betterC bug17940.d && obj2asm bug17940.o |tail -40 |head -16
                test    RBX,RBX
                je      L50
                test    RBX,RBX
                je      L2F
                movzx   EDI,byte ptr [RBX]
                call      _D8bug179404fun1FiZv@PLT32
L2F:            movsx   EDX,byte ptr [RBX]
                test    DL,DL
                je      L50
                mov     RCX,-8[RBP]
                mov     RAX,-010h[RBP]
                mov     RDX,RAX
                mov     DH,[RBX] <--- Here it is.
                and     ESI,0FFh <--- Note register.
                xor     EDI,EDI
                call      _D8bug179404fun2FS8bug179405ArrayiiZv@PLT32
```

WalterBright:
Thanks for tracking this down! I'd rather fix this at the register rewrite place. Delete lines 2263, 2264, and the same pair 24 lines up, in cod3.c.
Add these lines in cgcod.c after `cr->setReg(reg)`:
```
if (I64 && reg >= 4 && tysize(csextab[i].e->Ety) == 1)
    cr->Irex |= REX;
```
and see if that works!

FeepingCreature:
So done.

WalterBright:
Thanks!

PR #7253, opened 2017-10-28
fix issue 17944: MSCOFF: do not write empty debug$S section

rainers:
This allows linking with LLVM's LLD with debug information.


src/ddmd/backend/mscoffobj.c
```diff
@@ -1639,6 +1633,13 @@ segidx_t MsCoffObj::seg_xdata_comdat(symbol *sfunc)
 segidx_t MsCoffObj::seg_debugS()
 {
     // Probably should generate this lazilly, too.
+    if (segidx_debugS == UNKNOWN)
```
kinke: The above comment seems obsolete now (and no need to correct the `lazily` typo ;)).

rainers: Oops, somehow forgot to remove. Done now.

andralex:
@rainers rebase

PR #7254, opened 2017-10-28
COFF: do not create empty .drectve sections

rainers:
As https://github.com/dlang/dmd/pull/7253 this reduces the size of the phobos64.lib by more than 5% (about 2 of 36 MB).

More sections always written but less likely empty are .text, .rdata, .data and .bss


WalterBright:
Nice work!

PR #7256, opened 2017-10-28
Issue a nicer error message if critical envvars are missing

andralex:
Currently the behavior if someone forgets to define HOST_DMD is quite bizarre - the makefile creates a file called `src/-` and make issues an incomprehensible error message pointing to lexer.lib. This fixes that.


src/win32.mak
```diff
@@ -65,6 +65,9 @@ OS=windows
 
 ##### Directories
 
+# DM_HOME should be defined in the user's environment
+DM_HOME=Please_define_the__DM_HOME__environment_variable
```
rainers: I never set DM_HOME (and the required naming does not fit my installation), the C compiler already knows its standard include folders.

src/win32.mak
```diff
@@ -93,6 +96,8 @@ TOOLS_DIR=..\..\tools
 CC=dmc
 # D compiler (set with env variable)
 #HOST_DC=dmd
+# For error detection:
+HOST_DC=Please_define_the__HOST_DC__environment_variable
```
rainers: This works to display a nicer error, but it no longer allows setting HOST_DC in the environment. It is only overwritten on the command line.

andralex: On wine at least it works with HOST_DC set in the environment.

andralex: Could you try it on native Windows? Thx!

src/win32.mak
andralex: @rainers what if it's in a non-default location?

rainers: dmc is not in the default location on my system, but in c:\l\dmc instead. It uses sc.ini, too, containing `INCLUDE="%@P%\..\include";"%@P%\..\mfc\include";%INCLUDE%` which points to the standard include folders.

src/win32.mak
rainers: That's what I did. Are you sure to invoke DigitalMars make?

An explicit test of the variables works, too, but has to be added as a dependency to the defaulttarget and maybe others:
```
checkvars:
	cmd /c if "$(HOST_DC)" == "" (echo Please set HOST_DC && exit 1)
```
(DM make does not seem to interpret the result of executing an implicit shell command, so I had to make it explicit).

src/win32.mak
andralex: Cool, then can you please submit a small PR to eliminate it entirely? Also it would be great if you also updated the wiki instructions. Thx!

src/win32.mak
andralex: Interesting, I'll look into this. For what it's worth I use `wine make -f win32.mak` that I assume uses dmd make, but I'll check when I get back to that machine.

src/win32.mak
rainers: See https://github.com/dlang/dmd/pull/7260 
The wiki uses DM_HOME to refer to the DMD installation directory, that seems ok.

src/win32.mak
wilzbach: BTW we could also set the `DC_HOST` variable to `dmd` which is the default that's used for `posix.mak` and then simply add a nice error message as @rainers suggested.

(if DM make supports this)

JinShil:
@andralex Does this have a future?

JinShil:
See also #7467

src/win32.mak
marler8997: I tried to get this "default" functionality working with DM make and failed...

marler8997: Nevermind, got it it to work (https://github.com/dlang/dmd/pull/7468).

andralex:
@JinShil I don't have a Windows machine handy so it's difficult for me to test things properly. Do you think you can adopt this? It's fairly simple.

JinShil:
> Do you think you can adopt this?

Is it not already being adopted with #7467 and #7468?

JinShil:
Going with #7467

PR #7257, opened 2017-10-28
wine insists on invoking git even when tools/checkwhitespace.d is alr…

andralex:
…eady there


andralex:
This fix allows wine users to get by without installing git on windows (in addition to linux).

rainers (approved):
LGTM. I didn't know the `-` doesn't work with DM make if the executable is not found.

PR #7261, opened 2017-10-30
MSCOFF32: remove dependency on druntime for long divisions

rainers:
by calling div/mod functions from the VC runtime directly.

The _ms_* functions in rt.llmath can be removed if this is merged.

As we don't have an autotester for -m32mscoff: tests pass locally, but the dmd testsuite doesn't seem to exercise it that much. A bad version failed the druntime unittests, though.

### DIFF
--- a/src/ddmd/backend/cod1.c
+++ b/src/ddmd/backend/cod1.c
@@ -1840,6 +1840,7 @@ struct ClibInfo
         #define INFwkdone       4       // if weak extern is already done
         #define INF64           8       // if 64 bit only
         #define INFpushebx      0x10    // push EBX before load_localgot()
+        #define INFpusheabcdx   0x20    // pass EAX/EBX/ECX/EDX on stack, callee does ret 16
     char push87;                        // # of pushes onto the 8087 stack
     char pop87;                         // # of pops off of the 8087 stack
 };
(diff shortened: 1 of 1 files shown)


src/ddmd/backend/cod1.c
```diff
@@ -1998,8 +2001,9 @@ void getClibInfo(unsigned clib, symbol **ps, ClibInfo **pinfo)
                 }
                 else if (I32 && config.objfmt == OBJ_MSCOFF)
                 {
-                    s = symboly("_ms_aullrem", ALLREGS);
-                    cinfo->retregs32 = mDX|mAX;
+                    // use implementation in rt.llmath
+                    s = symboly("__ULDIV__", mAX|mBX|mCX|mDX);
```
rainers: @WalterBright : the OpenBSD/Solaris version also calls signed version instead, seems like a typo. see https://github.com/dlang/dmd/blob/master/src/ddmd/backend/cod1.c#L1995

WalterBright:
Unfortunately, this will fail with -betterC, which doesn't link with druntime. Perhaps add a check for -betterC to determine which to call?

src/ddmd/backend/cod1.c
WalterBright: I think you're right.

WalterBright:
Having the same object file format is only part of the battle. Each C compiler is different in how they connect to their respective runtime libraries. The only viable solution is to add switches.

rainers:
> Unfortunately, this will fail with -betterC, which doesn't link with druntime.

It will fail right now, too, as the `_ms_...` function are in the same module and call functions from the MS runtime. It's also just the same for other platforms calling `__ULDIV__` et al.

> Having the same object file format is only part of the battle. Each C compiler is different in how they connect to their respective runtime libraries. The only viable solution is to add switches.

This is about symbols in the VC runtime that is not available via dynamic linkage. An alternative would be to use `lldiv` instead, but that is only available since VS2013 according to MSDN (though MinGW declares it exists since VS2010).

rainers:
Changed the calls to directly invoke the VC runtime functions. This now actually removed the dependency on druntime for long divisions for better-C. 

It doesn't help replacing the C runtime, though. I've added the division functions from rt.llmath to my stub instead.

PR #7262, opened 2017-10-30
Move functions from statement.d to statementsem.d and make them private

RazvanN7:
After moving the semantic methods outside the statement AST nodes and in different files, a lot of helper methods were left in the statement.d file which are not called there at all. This PR moves those functions to statementsem and makes them private.


PetarKirov (approved):

andralex (approved):
cc @WalterBright

PR #7264, opened 2017-10-30
Make PostorderExpressionVisitor class private

RazvanN7:


src/ddmd/apply.d
```diff
@@ -27,7 +27,7 @@ import ddmd.visitor;
  * It's a bit slower than using virtual functions, but more encapsulated and less brittle.
  * Creating an iterator for this would be much more complex.
  */
-extern (C++) final class PostorderExpressionVisitor : StoppableVisitor
+private extern (C++) final class PostorderExpressionVisitor : StoppableVisitor
```
andralex: is `extern(C++)` still necessary?

RazvanN7: yes, otherwise the visiting methods do not override anything.

andralex: got it thx

andralex (approved):

PR #7265, opened 2017-10-30
Move setMangleOverride to dsymbolsem.d and make it private

RazvanN7:
There is only one call to setMangleOverride and it is in dsymbolsem.d


andralex (approved):
cc @WalterBright 

I will pull this as it seems reasonable.

PR #7268, opened 2017-10-30
Make constfold.expType private

RazvanN7:


src/ddmd/constfold.d
```diff
@@ -107,7 +107,7 @@ extern (C++) UnionExp Not(Type type, Expression e1)
     return ue;
 }
 
-extern (C++) UnionExp Bool(Type type, Expression e1)
+private extern (C++) UnionExp Bool(Type type, Expression e1)
```
andralex: remove extern(C++)?

RazvanN7: yes, my bad.

andralex (approved):

PR #7269, opened 2017-10-31
Fix issue 17943 - Wrong pointer arithmetic optimization

yshui:
store result as llong.


andralex:
So the test in the bug report is not reproducible? Can you write a test that would be broken without this change? Thx!

yshui:
@andralex It's reproducible for me. I will include the test.

MartinNowak:
thx

MartinNowak (approved):

PR #7270, opened 2017-10-31
Fix Issue 16649 - Header gen skips parens

RazvanN7:


andralex (approved):

RazvanN7:
@andralex I think the merge was a bit hasty. The fix should have targeted stable... but I guess folks can wait until the next release

andralex:
cc @MartinNowak @WalterBright - should we cherry pick that into stable as well?

PR #7271, opened 2017-10-31
Enable 64-bit build with VC2015

atilaneves:
There is currently no top-level win64 makefile nor does dmd currently build with VS2015. This patch addresses both these issues.

### DIFF
--- a/src/ddmd/backend/gflow.c
+++ b/src/ddmd/backend/gflow.c
@@ -323,7 +323,7 @@ STATIC void accumrd(vec_t GEN,vec_t KILL,elem *n)
                 rdelem(&Gl,&Kl,n->E1);
                 rdelem(&Gr,&Kr,n->E2);
 
-                switch (el_returns(n->E1) * 2 | el_returns(n->E2))
+                switch (el_returns(n->E1) * 2 | int(el_returns(n->E2)))
                 {
                     case 3: // E1 and E2 return
                         /* GEN = (GEN - Kl) | Gl |
--- a/src/ddmd/backend/gloop.c
+++ b/src/ddmd/backend/gloop.c
@@ -1034,7 +1034,7 @@ STATIC void markinvar(elem *n,vec_t rd)
         case OPcolon:
         case OPcolon2:
                 tmp = vec_clone(rd);
-                switch (int(el_returns(n->E1) * 2) | int(el_returns(n->E2)))
+                switch (el_returns(n->E1) * 2 | int(el_returns(n->E2)))
                 {
                     case 3: // E1 and E2 return
                         markinvar(n->E1,rd);
--- a/src/ddmd/backend/gloop.c
+++ b/src/ddmd/backend/gloop.c
@@ -1034,7 +1034,7 @@ STATIC void markinvar(elem *n,vec_t rd)
         case OPcolon:
         case OPcolon2:
                 tmp = vec_clone(rd);
-                switch (el_returns(n->E1) * 2 | el_returns(n->E2))
+                switch (int(el_returns(n->E1) * 2) | int(el_returns(n->E2)))
                 {
                     case 3: // E1 and E2 return
                         markinvar(n->E1,rd);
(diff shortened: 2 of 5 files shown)


src/ddmd/backend/gloop.c
```diff
@@ -1034,7 +1034,7 @@ STATIC void markinvar(elem *n,vec_t rd)
         case OPcolon:
         case OPcolon2:
                 tmp = vec_clone(rd);
-                switch (el_returns(n->E1) * 2 | el_returns(n->E2))
+                switch (int(el_returns(n->E1) * 2) | int(el_returns(n->E2)))
```
John-Colvin: only need second cast

src/ddmd/backend/gother.c
```diff
@@ -219,7 +219,7 @@ STATIC void conpropwalk(elem *n,vec_t IN)
         if (op == OPcolon || op == OPcolon2)
         {
                 L = vec_clone(IN);
-                switch (el_returns(n->E1) * 2 | el_returns(n->E2))
+                switch (int(el_returns(n->E1) * 2) | int(el_returns(n->E2)))
```
John-Colvin: only need second cast

WalterBright:
I thought this was already building with VS? @rainers ?

src/ddmd/backend/gflow.c
```diff
@@ -323,7 +323,7 @@ STATIC void accumrd(vec_t GEN,vec_t KILL,elem *n)
                 rdelem(&Gl,&Kl,n->E1);
                 rdelem(&Gr,&Kr,n->E2);
 
-                switch (el_returns(n->E1) * 2 | el_returns(n->E2))
+                switch (el_returns(n->E1) * 2 | int(el_returns(n->E2)))
```
WalterBright: This looks strange - why is the cast needed?

rainers:
> I thought this was already building with VS? @rainers ?

It does, but building through the makefile enables "warnings as errors".

src/ddmd/backend/gflow.c
rainers: el_returns returns a boolean which causes:
warning C4805: '|' : unsafe mix of type 'int' and type 'bool' in operation

src/vcbuild/msvc-dmc.d
```diff
@@ -19,6 +21,15 @@ int main(string[] args)
     newArgs ~= `/Ivcbuild`;
     newArgs ~= `/Iddmd\root`;
     newArgs ~= `/FIwarnings.h`;
+
+    enum vs2015 = "Microsoft Visual Studio 14.0";
+    if (environment.get("VCINSTALLDIR", "").canFind(vs2015) || environment.get("MSVC_CC", "").canFind(vs2015))
```
rainers: The check can fail if the installation path is not the default. Checking the version directly using VisualStudioVersion (also set via vcvarsall.bat) might be less brittle.

atilaneves: Just pushed changes to do that. Thanks for the comment, I didn't know about `%VisualStudioVersion%`.

src/vcbuild/msvc-dmc.d
```diff
@@ -19,6 +21,14 @@ int main(string[] args)
     newArgs ~= `/Ivcbuild`;
     newArgs ~= `/Iddmd\root`;
     newArgs ~= `/FIwarnings.h`;
+
+    if (environment.get("VisualStudioVersion", "").to!double >= 14.0)
```
rainers: I'm not sure digger really uses the vcvarsall batches, so defaulting to version 10.0 might be better for backward compatibility. AFAIK to!double will cause an exception on empty input.

atilaneves: The autotester must, I was relying on that to catch any problems. I'll change as per your comment, it's much better that way anyway.

rainers: Thanks.

The autotester doesn't build dmd with VS, but with dmc. So no errors here will be caught.

src/ddmd/backend/gflow.c
WalterBright: ok

PR #7272, opened 2017-11-01
fix Itanium C++ mangling 2

WalterBright:
Got rid of `is_top_level` flag and `is_var_arg` flag. Untangled mess in `template_args()`. Added module level comments and links. Slightly better error messages.

### DIFF
--- a/src/ddmd/cppmangle.d
+++ b/src/ddmd/cppmangle.d
@@ -9,7 +9,25 @@
 
 module ddmd.cppmangle;
 
-// Online documentation: https://dlang.org/phobos/ddmd_cppmangle.html
+
+/**
+ * Do mangling for C++ linkage.
+ *
+ * References:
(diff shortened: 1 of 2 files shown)


src/ddmd/cppmangle.d
```diff
@@ -188,46 +207,30 @@ private final class CppMangleVisitor : Visitor
...
+                        val = -val;
+                        buf.writeByte('n');
                     }
+                    buf.printf("%llu", val);
                     buf.writeByte('E');
                 }
                 else
                 {
-                    ti.error("Internal Compiler Error: C++ %s template value parameter is not supported", tv.valType.toChars());
+                    ti.error("Internal Compiler Error: C++ `%s` template value parameter is not supported", tv.valType.toChars());
```
andralex: not covered which is great :)

andralex (approved):
Thanks!

PR #7273, opened 2017-11-01
String Switch Lowering

somzzz:
Paired with https://github.com/dlang/druntime/pull/1952

We replace a switch with string labels with a switch with integer case labels. The new label for each string is the index of the string in the sorted array of labels.

The switch condition is determined through a call to the druntime template object.__switch. Previous implementation generated a call to the druntime functions from rt/switch__.d .

The new integer switch goes to s2ir and gets generated based on the same rules as before.

This PR only eliminates special processing for string switches in IR + calls to druntime functions.

### DIFF
--- a/src/ddmd/s2ir.d
+++ b/src/ddmd/s2ir.d
@@ -557,77 +557,6 @@ extern (C++) class S2irVisitor : Visitor
             return;
         }
 
-        if (s.condition.type.isString())
-        {
-            // Number the cases so we can unscramble things after the sort()
-            for (size_t i = 0; i < numcases; i++)
-            {   CaseStatement cs = (*s.cases)[i];
-                cs.index = cast(int)i;
-            }
-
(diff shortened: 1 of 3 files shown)


ibuclaw:
Looks like something went wrong, there is no commit here.

ibuclaw:
Don't use `git merge`, instead `git rebase master` to reseat your local branches.

somzzz:
Something was not right with master on my fork. When I tried to rebase the switch_lowering branch on top of it - hell broke loose.

src/ddmd/s2ir.d
```diff
@@ -557,77 +557,6 @@ extern (C++) class S2irVisitor : Visitor
             return;
         }
 
-        if (s.condition.type.isString())
```
ibuclaw: Shouldn't this be replaced with an assert? This function no longer handles string conditions, and it would be a compiler error if it were given one.

PR #7274, opened 2017-11-01
Upgrade dub.sdl file

RazvanN7:
The parser library contains 3 visitors which were not there when the dub.sdl file was created. I added the dependencies.


wilzbach (approved):
LGTM - though it wouldn't hurt to add a test for the DUB package like https://github.com/dlang/dmd/pull/6999 attempted

PR #7275, opened 2017-11-01
WIP (do not merge/review): Trigger autotester for issue 17167

atilaneves:


rainers:
>WIP (do not merge/review):

I couldn't resist though: please note that dmd doesn't use UTF8 strings for filenames, but the raw strings passed to C's main. To convert these to wide strings it's probably best to use MultiByteToWideChar(CP_ACP) as this is also what the ...A function do.

atilaneves:
@rainers Thanks for your comment. I erroneously assumed that since the frontend is in D now that `char` meant UTF8.

atilaneves:
@rainers What about `canonicalName`? It retuns a `const(char)*` - I guess it'd have to use the opposite function `WideCharToMultiByte`?

rainers:
> What about canonicalName? It retuns a const(char)*

It just calls GetFullPathNameA, so it should be consistent with ANSI encoding. Or do you want to use it with wide characters, too? Maybe just use GetFullPathNameW directly?

Sure, UTF8 is a lot easier to deal with, but unfortunately it's rather tricky to output it to the console correctly. Also other tools (i.e. the linker) invoked by dmd or interpreting dmd's output usually assume ANSI encoding, too.

atilaneves:
>  Or do you want to use it with wide characters, too? Maybe just use GetFullPathNameW directly?

I don't actually want to use wide characters, it's just that the W versions of the Win32 API don't have the file path length limitation. If I remove the bug from part of dmd but leave it whenever it needs to call `canonicalName`...

PR #7277, opened 2017-11-02
fix Issue 17955 - compiler segfault in DsymbolSemanticVisitor::visit(…

WalterBright:
…UnittestDeclaration*)


MartinNowak:
REGRESSION FIXES SHOULD BE MERGED INTO STABLE!!!

This regression slipped into 2.077.0 while this fix still isn't in stable.

PR #7280, opened 2017-11-02
Fix Issue 4946 - Not good error message with wrongly positioned 'const'

RazvanN7:


src/ddmd/parse.d
```diff
@@ -4767,6 +4767,13 @@ final class Parser(AST) : Lexer
             else if (!f.frequire && !f.fensure) // allow these even with no body
             {
                 error("semicolon expected following function declaration");
+                TOK t = token.value;
+                import core.stdc.stdio;
```
wilzbach: Remove

src/ddmd/parse.d
```diff
@@ -4767,6 +4767,13 @@ final class Parser(AST) : Lexer
...
+                if (t == TOKconst || t == TOKimmutable || t == TOKinout || t == TOKreturn ||
+                    t == TOKshared || t == TOKnothrow || t == TOKpure)
+                    error("'%s' cannot be placed after template constraint", token.toChars);
```
wilzbach: after __a__ template constraint?

src/ddmd/parse.d
```diff
@@ -4767,6 +4767,13 @@ final class Parser(AST) : Lexer
             else if (!f.frequire && !f.fensure) // allow these even with no body
             {
                 error("semicolon expected following function declaration");
```
wilzbach: How about moving this to the `else` - i.e. as default message?

wilzbach (approved):
Looks reasonable to me.

src/ddmd/parse.d
RazvanN7: oh yeah, thanks, forgot about it.

src/ddmd/parse.d
RazvanN7: fair enough

src/ddmd/parse.d
```diff
@@ -4766,7 +4766,14 @@ final class Parser(AST) : Lexer
             }
             else if (!f.frequire && !f.fensure) // allow these even with no body
             {
-                error("semicolon expected following function declaration");
+                TOK t = token.value;
+                if (t == TOKconst || t == TOKimmutable || t == TOKinout || t == TOKreturn ||
+                    t == TOKshared || t == TOKnothrow || t == TOKpure)
+                    error("'%s' cannot be placed after a template constraint", token.toChars);
+                else if (t == TOKat)
+                    error("annotations cannot be placed after a template constraint");
```
BBasile: rather _attributes_ no ? _annotation_  isn't an official name.

RazvanN7: yeah, it does sound better. thanks!

andralex (approved):

PR #7281, opened 2017-11-03
Fix issue 7184 - parse error on *(x)++

Biotronic:
The test for C style casts was too aggressive and didn't take into account postfix ++ and --.
It also triggers when the item inside parentheses is a callable. This could potentially be reported at a later point, when the compiler knows whether the contents of the parentheses is a type. This PR does not fix that problem in any way.

### DIFF
--- a/src/ddmd/parse.d
+++ b/src/ddmd/parse.d
@@ -7966,6 +7966,13 @@ final class Parser(AST) : Lexer
                                     e = new AST.TypeExp(loc, t);
                                     e = parsePostExp(e);
                                 }
+                                else if (token.value == TOKplusplus || token.value == TOKminusminus)
+                                {
+                                    // if ()++
+                                    // or ()--
+                                    e = new AST.TypeExp(loc, t);
+                                    e = parsePostExp(e);
+                                }
                                 else
                                 {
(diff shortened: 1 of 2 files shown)


test/runnable/test7184.d
```diff
@@ -0,0 +1,8 @@
+// Tests for 7184
+
+void main() {
+    auto a = 0;
+    auto b = (a)++;
```
WalterBright: Should test the `--` case, too, for completeness.

ibuclaw:
What happens with `(int)++x` after this change?

src/ddmd/parse.d
```diff
@@ -7966,6 +7966,13 @@ final class Parser(AST) : Lexer
                                     e = new AST.TypeExp(loc, t);
                                     e = parsePostExp(e);
                                 }
+                                else if (token.value == TOKplusplus || token.value == TOKminusminus)
```
UplinkCoder: Seems like this if can be merged into the one above.

JinShil:
@Biotronic The reviewers suggestion minor changes.  Can we count on you to move this forward?

JinShil:
Revived at #7442

PR #7284, opened 2017-11-04
fix Issue 17934 - [scope] scopeness entrypoint for unique/ref-counted…

WalterBright:
… missing

This infers `scope` for local variables that have destructors.

### DIFF
--- a/test/fail_compilation/retscope3.d
+++ b/test/fail_compilation/retscope3.d
@@ -48,3 +48,45 @@ void bar4()
     {
     }
 }
+
+/**********************************************/
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/retscope3.d(3027): Error: scope variable `l` assigned to `elem` with longer lifetime
+---
+*/
+
+#line 3000
+
+struct List
+{
+    Elem front() @safe return scope;
+
+    ~this() @trusted scope;
+
+    @disable this(this);
+
+    void* data;
+}
+
+struct Elem
+{
+    void* data;
+}
+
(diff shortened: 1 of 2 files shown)


test/fail_compilation/retscope3.d
```diff
@@ -48,3 +48,45 @@ void bar4()
...
+{
+    return List();
+}
+
+void test3000() @safe
+{
+    Elem elem;
+    {
+        auto l = list(); // inferred as scope
+        elem = l.front; // escapes, b/c l isn't scoped
```
mathias-lang-sociomantic: Isn't that defeated by just using `list().front` though ?

WalterBright: The lifetime of the return value of list().front is the lifetime of list().

mathias-lang-sociomantic: Yeah, and if you change your test case from `auto l = list(); elem = l.front;` to `elem = list().front` it no longer fails. I suppose that's a case similar to [static arrays](https://github.com/dlang/dmd/pull/7110) but it doesn't look like this case would be caught either.

WalterBright: As always, previously unknown issues should be filed on bugzilla.

WalterBright: I didn't see it, so I added it: https://issues.dlang.org/show_bug.cgi?id=17977

Geod24: Thanks ! I indeed forgot :)

PR #7285, opened 2017-11-05
refactor: use Identifiers for findCondition() instead of strings

WalterBright:
The idea is that `Identifier`s should be used instead of strings where ever possible, and `parseCommandLine()` should not be setting global variables.


WalterBright:
```
ddmd/doc.d(356): Error: file "default_ddoc_theme.ddoc" cannot be found or not in a path specified with -J
```
That error seems entirely unrelated?

andralex (approved):

PR #7287, opened 2017-11-05
refactor isReserved(): string table to string switch

WalterBright:
Rather than an array with a linear search, use a string switch with a binary search!

### DIFF
--- a/src/ddmd/cond.d
+++ b/src/ddmd/cond.d
@@ -590,9 +590,6 @@ extern (C++) final class VersionCondition : DVCondition
     /**
      * Check if a given version identifier is reserved.
      *
-     * Reserved identifier are the one documented below or
-     * those starting with 'D_'.
-     *
      * Params:
      *   ident = identifier being checked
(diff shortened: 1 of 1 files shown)


wilzbach (approved):

src/ddmd/cond.d
```diff
@@ -602,105 +599,104 @@ extern (C++) final class VersionCondition : DVCondition
...
+            case "CRuntime_Bionic":
+            case "CRuntime_DigitalMars":
+            case "CRuntime_Glibc":
+            case "CRuntime_Microsoft":
+            case "CRuntime_Musl":
+            case "CRuntime_UClibc":
+            case "unittest":
+            case "assert":
+            case "all":
+            case "none":
```
andralex: You could have used the syntax `case "DigitalMars", "GNU", ...` no?

andralex (commented):
Cool! cc @somzzz please make a note to add this to your tests.

PR #7288, opened 2017-11-05
Separate -betterC into different global parameters

ibuclaw:
So its the same as what we do for the `-release` switch.

### DIFF
--- a/src/ddmd/globals.h
+++ b/src/ddmd/globals.h
@@ -110,6 +110,9 @@ struct Param
     bool nofloat;       // code should not pull in floating point support
     bool ignoreUnsupportedPragmas;      // rather than error on them
     bool enforcePropertySyntax;
+    bool useCAsserts;   // use C assert() on bounds and contract failures
+    bool useModuleInfo; // generate runtime module information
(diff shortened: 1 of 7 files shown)


src/ddmd/globals.h
```diff
@@ -110,6 +110,9 @@ struct Param
     bool nofloat;       // code should not pull in floating point support
     bool ignoreUnsupportedPragmas;      // rather than error on them
     bool enforcePropertySyntax;
+    bool useCAsserts;   // use C assert() on bounds and contract failures
+    bool useModuleInfo; // generate runtime module information
+    bool useTypeInfo;   // generate runtime type information
     bool betterC;       // be a "better C" compiler; no dependency on D runtime
```
klickverbot: This should probably be renamed to `defineBetterCVersion` or something like that to make clear it shouldn't be used directly as a feature gate.

dnadlinger (approved):

wilzbach:
Shouldn't we expose these params via the CLI?

ibuclaw:
For dmd?  Probably no for the same reason that flags controlled by `-release` are not exposed either.

Others are free to what they want.

PR #7290, opened 2017-11-05
refactor: use parseDigits() to replace strtol()

WalterBright:
Step toward making `parseCommandLine()` pure by encapsulating reliance on C's `errno`.

Blocked by https://github.com/dlang/dmd/pull/7289 and https://github.com/dlang/dmd/pull/7285

### DIFF
--- a/src/ddmd/mars.d
+++ b/src/ddmd/mars.d
@@ -11,14 +11,13 @@
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
  * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/ddmd/mars.d, _mars.d)
+ * Documentation:  https://dlang.org/phobos/ddmd_mars.html
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/ddmd/mars.d
  */
 
 module ddmd.mars;
 
-// Online documentation: https://dlang.org/phobos/ddmd_mars.html
-
 import core.stdc.ctype;
-import core.stdc.errno;
 import core.stdc.limits;
 import core.stdc.stdio;
 import core.stdc.stdlib;
@@ -60,10 +59,6 @@ import ddmd.semantic;
 import ddmd.target;
 import ddmd.tokens;
 import ddmd.utils;
-
-// strtol redeclared here because its signature changed from 2.073 to 2.074
-import core.stdc.config;
-extern(C) c_long strtol(inout(char)* nptr, inout(char)** endptr, int base);
 
 /**
  * Print DMD's logo on stdout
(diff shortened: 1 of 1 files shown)


src/ddmd/mars.d
```diff
@@ -2002,11 +2010,11 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 {
                     if (isdigit(cast(char)p[7]))
                     {
-                        errno = 0;
-                        const level = strtol(p + 7, &p, 10);
-                        if (*p || errno || level > INT_MAX)
+                        const level = parseDigits(p + 7, int.max);
+                        if (level == uint.max)
                             goto Lerror;
-                        DebugCondition.setGlobalLevel(cast(int)level);
```
ibuclaw: Who else is using this method?

src/ddmd/mars.d
```diff
@@ -1682,11 +1691,10 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
...
                 if (p[8] == '=' && isdigit(cast(char)p[9]))
                 {
-                    errno = 0;
-                    const num = strtol(p + 9, &p, 10);
-                    if (*p || errno || num > INT_MAX)
+                    const num = parseDigits(p + 9, int.max);
+                    if (num == uint.max)
                         goto Lerror;
-                    global.errorLimit = cast(uint)num;
+                    params.errorLimit = num;
```
ibuclaw: This seems to assume that some other change has been merged.

WalterBright: That's why there's the "blocked by" note above :-)

src/ddmd/mars.d
WalterBright: See https://github.com/dlang/dmd/pull/7285

src/ddmd/mars.d
```diff
@@ -1498,6 +1493,33 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
...
+     * Returns:
+     *  uint.max on error, otherwise converted integer
+     */
+    static pure uint parseDigits(const(char)*p, const uint max)
+    {
+        uint value;
+        bool overflow;
+        while (1)
+        {
+            const c = *p;
```
andralex: Save a comparison per pass, simplify flow:

```d
for (uint d; (d = uint(*p) - uint('0')) <= 9; ++p)
{
    import core.checkedint : mulu, addu;
    value = mulu(value, 10, overflow);
    value = addu(value, d, overflow);
}
return (overflow || value > max || c) ? uint.max : value;
```

andralex (approved):
Guess speed is not important but made a suggestion just in case.

src/ddmd/mars.d
WalterBright: It's beautiful, but that `c` in the return expression is a problem. `d` has gone out of scope, so I'd have to use `*p`, meaning an extra indirection.

WalterBright: Ok, I made the change.

andralex: That's only one indirection, not one per loop. You can of course change the code like this at the cost of some minor obfuscation:

```d
uint d;
for (; (d = uint(*p) - uint('0')) <= 9; ++p)
{
    import core.checkedint : mulu, addu;
    value = mulu(value, 10, overflow);
    value = addu(value, d, overflow);
}
return (overflow || value > max || d + '0' != 0) ? uint.max : value;
```

andralex: Eh, I realized there's indeed a better version with an infinite loop:

```d
while (1)
{
    const d = uint(*p) - uint('0');
    if (d > 9)
        return uint.max;
    import core.checkedint : mulu, addu;
    value = mulu(value, 10, overflow);
    value = addu(value, d, overflow);
    ++p;
}
return (overflow || value > max) ? uint.max : value;
```

Looks like this does minimal work.

andralex (approved):

andralex:
feel free to pull

stefan-koch-sociomantic:
You use a possibly valid value as error signal ?

PR #7296, opened 2017-11-08
refactor: consolidate handling of various assert errors

WalterBright:
There are 3 kinds of assert failures in DMD: assert failures, array boundscheck failures, and switch default failures. This refactor doesn't change any behavior, but brings a consistent terminology and pattern to how they are handled, making it practical to add more behavior options (such as halt on assert failures).

### DIFF
--- a/src/ddmd/globals.h
+++ b/src/ddmd/globals.h
@@ -24,12 +24,20 @@
 template <typename TYPE> struct Array;
 
 // The state of array bounds checking
-enum BOUNDSCHECK
+typedef unsigned char CHECKENABLE;
+enum
 {
-    BOUNDSCHECKdefault, // initial value
-    BOUNDSCHECKoff,     // never do bounds checking
-    BOUNDSCHECKon,      // always do bounds checking
-    BOUNDSCHECKsafeonly // do bounds checking only in @safe functions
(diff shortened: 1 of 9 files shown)


src/ddmd/globals.h
```diff
@@ -24,12 +24,20 @@
...
+    CHECKENABLEoff,     // never do bounds checking
+    CHECKENABLEon,      // always do bounds checking
+    CHECKENABLEsafeonly // do bounds checking only in @safe functions
+};
+
+typedef unsigned char CHECKACTION;
+enum
+{
+    CHECKACTION_D,        // call D assert on failure
+    CHECKACTION_C,        // call C assert on failure
```
mathias-lang-sociomantic: Why is it a separate enum BTW ? Do you envision one can call C assert from non-`betterC` code via a compiler switch ?

WalterBright: > separate

Because `safeonly` and `C` are independent attributes. Trying to mush together when an assert is triggered with what the assert should do when triggered is ultimately confusing.

> Do you envision one can call C assert from non-betterC code via a compiler switch ?

Every combination should work, even if we don't expose it to the user. I have thought about making a `pragma` for them, which would give local control.

PR #7297, opened 2017-11-10
Fix Issue 17976 - core.exception.AssertError@ddmd/dsymbolsem.d(1624)

RazvanN7:
When a parameter with the same name was encountered the parameter was no longer pushed in the parameters array resulting in the assertion failure. I fixed it by getting rid of the else branch and there doesn't seem to be any side effects.

### DIFF
--- a/src/ddmd/dsymbolsem.d
+++ b/src/ddmd/dsymbolsem.d
@@ -1621,7 +1621,8 @@ extern(C++) final class Semantic3Visitor : Visitor
         if (funcdecl.parameters)
         {
             size_t nfparams = Parameter.dim(f.parameters);
-            assert(nfparams == funcdecl.parameters.dim);
+            if (funcdecl.errors)
+                assert(nfparams == funcdecl.parameters.dim);
             foreach (u, v; *funcdecl.parameters)
             {
                 if (v.storage_class & STCmaybescope)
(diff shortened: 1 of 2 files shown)


PetarKirov (approved):
Looks good.

mathias-lang-sociomantic (approved):

WalterBright:
I'm not sure this is the right fix. Error recovery should not be about attempting to repair the AST. The idea is to poison the AST.

RazvanN7:
@WalterBright Got it. Thanks! Is this ok?

src/ddmd/dsymbolsem.d
```diff
@@ -1621,7 +1621,8 @@ extern(C++) final class Semantic3Visitor : Visitor
         if (funcdecl.parameters)
```
WalterBright: I'd say change this line to:
```
if (funcdecl.parameters && !funcdecl.errors)
```
because inferring scope is not necessary after the function has errors in it.

src/ddmd/dsymbolsem.d
```diff
@@ -1621,7 +1621,8 @@ extern(C++) final class Semantic3Visitor : Visitor
...
         {
             size_t nfparams = Parameter.dim(f.parameters);
-            assert(nfparams == funcdecl.parameters.dim);
+            if (funcdecl.errors)
```
WalterBright: The logic is inverted here, should have a `!` prepended. But in any case, delete this line and use the one mentioned above.

RazvanN7: **funcdecl.errors** wasn't updated in that case. I updated the code

PR #7298, opened 2017-11-10
Fix Issue 16183 - [REG2.068] compile-time string concatenation fails …

JinShil:
…with CTFE and char[] literal involved


UplinkCoder:
Looks good afaict

PR #7299, opened 2017-11-10
Fix issue 17167 - handle long filepaths on Windows

atilaneves:
This issue is made more grave by the fact that dub uses relative file paths for any package dependencies. This means that if your project is in `C:\foo\bar\baz\quux`, it'll try writing to `C:\foo\bar\baz\quux\..\..\..\..\Users\%USERNAME%\AppData\Roaming\dub\packages\%PACKAGE_VERSION%\%PACKAGE%\.dub\%HASH%`. This gets big quite quickly and goes over Windows's 260 character file path limit, meaning that dub build fails with a message saying that it can't write to a certain directory.

This PR fixes the issue by using the newer unicode-aware Win32 APIs, which don't have the limitation.

### DIFF
--- a/src/ddmd/root/filename.d
+++ b/src/ddmd/root/filename.d
@@ -592,15 +592,16 @@ nothrow:
         }
         else version (Windows)
         {
-            wchar[1024] wnameBuf;
-            const wname = name.toWStringz(wnameBuf);
-            const dw = GetFileAttributesW(&wname[0]);
-            if (dw == -1)
-                return 0;
-            else if (dw & FILE_ATTRIBUTE_DIRECTORY)
-                return 2;
-            else
-                return 1;
+            return name.toWStringzThen!((wname)
+            {
+                const dw = GetFileAttributesW(&wname[0]);
+                if (dw == -1)
+                    return 0;
+                else if (dw & FILE_ATTRIBUTE_DIRECTORY)
+                    return 2;
+                else
+                    return 1;
+            });
         }
         else
         {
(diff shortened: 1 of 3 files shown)


src/ddmd/root/filename.d
```diff
@@ -721,3 +731,132 @@ nothrow:
...
+      limitation of 248 characters, `mkdir` fails with less and might
+      fail due to the number of consecutive `..`s in the
+      path. `CreateDirectoryW` also normally has a 248 character
+      limit, unless the path is absolute and starts with `\\?\`. Note
+      that this is different from starting with the almost identical
+      `\\?`.
+
+      Please consult
+      https://msdn.microsoft.com/en-us/library/windows/desktop/aa363855(v=vs.85).aspx
+    */
```
WalterBright: Please use Ddoc style comments. In particular, this needs a `Params:` section, a `Returns:` section, and the `Please consult` should be `References:`. Ditto for the rest of the added functions.

src/ddmd/root/filename.d
```diff
@@ -721,3 +731,132 @@ nothrow:
...
+        enum ERROR_PATH_NOT_FOUND = 3;
+        enum ERROR_ALREADY_EXISTS = 183;
+
+        SetLastError(NO_ERROR);
+        const createRet = path.extendedPathThen!(p => CreateDirectoryW(&p[0],
+                                                                       null /*securityAttributes*/));
+        const lastError = GetLastError();
+
+        // Preserve compatibility with mkdir since the calling code expects
+        // errno to be set.
```
WalterBright: Don't need compatibility since there is only one caller of this private function. Please don't use `errno` on Windows unless forced to.

src/ddmd/root/filename.d
```diff
@@ -721,3 +731,132 @@ nothrow:
...
+    {
+        import core.stdc.string: strlen;
+
+        // cache this for efficiency
+        const strLength = strlen(str) + 1;
+        // first find out how long the buffer must be to store the result
+        const length = MultiByteToWideChar(0 /*codepage*/, 0 /*flags*/, str, strLength, null, 0);
+        if (!length) return null;
+
+        auto ret = length > buf.length ? new wchar[length] : buf;
```
WalterBright: The `new` is a memory leak in dmd. Take a look at https://github.com/dlang/phobos/blob/master/std/internal/cstring.d#L122 where malloc is used for the temporary buffer, cleaned up with free().

WalterBright:
This problem was fixed in Phobos in std.file long ago :-)

src/ddmd/root/filename.d
atilaneves: Fixed.

src/ddmd/root/filename.d
atilaneves: Fixed.

src/ddmd/root/filename.d
atilaneves: I thought that leaking was the memory strategy for dmd in order for it to be fast and that at some point the GC would be turned on. The code you linked is from Phobos, so do you mean I should create a similar struct for dmd or just to use `free` to clean up?

PR #7302, opened 2017-11-11
Fix Issue 16284 - [REG2.067] CTFE internal error: bad compare

JinShil:


UplinkCoder:
@JinShil please provide a description.
My intuition: this fix seems to allow more then intended.

WalterBright:
What this solution does is decide that void initialized fields (which is done for unions) should compare equal to other void initialized fields. This makes me uneasy. I'd rather they compared unequal. Or even issue an error.

In any case, this behavior should be documented in the spec.

JinShil:
What appears to happen is whenever a single field of a union is assigned to, the CTFE interpreter sets all other fields to `void` as a way of invalidating them, or so I assume.  

See https://dlang.org/changelog/2.065.0.html#ctfe-overlapped-field
See https://github.com/dlang/dmd/blob/03e806abaae49fe84de8fc0bdbe2ead80f58d038/src/ddmd/dinterpret.d#L3959-L3960

If that's by design, then any aggregate containing a union and an assigned-to field will never compare to equality.  I need some clarification on how this should behave in order to continue.

JinShil:
According to GitHub's blame feature @9rnsr was the author that introduced the `dinterpret.stompOverlappedFields()` feature for `union`s.  @UplinkCoder are you the resident expert on this.  If so, what are your thoughts?

stefan-koch-sociomantic:
changing the old interpreter to not relay on the void expr,
is going to be tricky;
newCTFE will not relay on the void expression in order to check the current valid fields.

JinShil:
We need a design for this.  Closing until we have one.

WalterBright:
I changed my mind, this is the right fix.

WalterBright:
Eek! Too many diffs. Will reboot.

thewilsonator:
I accident rebased again master instead of stable.

thewilsonator:
Its 15 lines.

PR #7303, opened 2017-11-11
Lower final switch default error to object.d template

somzzz:
Druntime changes: https://github.com/dlang/druntime/pull/1971

I still had to use ```SwitchErrorStatement``` because of the checks and actions the compiler performs for ```ErrorStatemenets``` vs regular ```Statements```.


src/ddmd/backend/rtlsym.d
```diff
@@ -53,9 +53,9 @@ enum
     RTLSYM_MONITOREXIT,
     RTLSYM_CRITICALENTER,
     RTLSYM_CRITICALEXIT,
-    RTLSYM_SWITCH_STRING,
-    RTLSYM_SWITCH_USTRING,
-    RTLSYM_SWITCH_DSTRING,
+    RTLSYM_SWITCH_STRING,       // unused
+    RTLSYM_SWITCH_USTRING,      // unused
+    RTLSYM_SWITCH_DSTRING,      // unused
```
wilzbach: https://issues.dlang.org/show_bug.cgi?id=9395 would be very useful here ;-)

src/ddmd/s2ir.d
```diff
@@ -691,19 +691,16 @@ extern (C++) class S2irVisitor : Visitor
...
+            //printf("SwitchErrorStatement.toIR(), exp = %s\n", s.exp ? s.exp.toChars() : "");
+            incUsage(irs, s.loc);
+            block_appendexp(blx.curblock, toElemDtor(s.exp, irs));
         }
         else
         {
-            auto efilename = el_ptr(toSymbol(cast(Dsymbol)blx._module));
-            auto elinnum = el_long(TYint, s.loc.linnum);
-            e = el_bin(OPcall, TYvoid, el_var(getRtlsym(RTLSYM_DSWITCHERR)), el_param(elinnum, efilename));
+            assert(0);
```
wilzbach: It's a good style to provide an informative error message or do sth. like 

```d
assert(s.exp !is null, "SwitchErrorStatement needs to have a valid Expression.") 
```

and skip the `if` entirely.

wilzbach (commented):
Nits

src/ddmd/s2ir.d
JinShil: > It's a good style to provide an informative error message or do sth. like...

@WalterBright disagrees: https://github.com/dlang/dmd/pull/7079#discussion_r132904181

> Some people like assert messages, I don't. They bloat up the compiler. Leaving the message as a comment in the source code is plenty good enough.

I really don't mind either way, just an FYI.

andralex (approved):

PR #7304, opened 2017-11-12
betterC: add nothrow RAII for destructors in expressions

WalterBright:
This is a start on adding RAII support for -betterC by doing it the old-fashioned way - assume a nothrow environment. It's the way C++ worked before exceptions were added to the language.

Adding this support for statement level RAII comes later.

This also suggests a significant optimization for regular D if nothrow is tracked at the statement level rather than at the function level.


src/ddmd/irstate.d
```diff
@@ -263,4 +263,18 @@ struct IRState
...
+     *  true if in a nothrow section of code
+     */
+    bool isNothrow()
+    {
+        /* Note that even nothrow functions can have throwing parts,
+         * so until we do a better job tracking that, this is
+         * the best we can do.
+         * Nothrow needs to be tracked at the Statement level.
+         */
+        return global.params.betterC;
```
andralex: dis kinda sux

andralex (approved):
let's get back to improving this real soon thx

PR #7307, opened 2017-11-12
BetterC: disallow throw and try-catch

WalterBright:
They won't work anyway, might as well give a decent error message.

### DIFF
--- a/src/ddmd/statementsem.d
+++ b/src/ddmd/statementsem.d
@@ -3607,6 +3607,14 @@ else
 
     override void visit(TryCatchStatement tcs)
     {
+        //printf("TryCatchStatement.semantic()\n");
+
+        if (global.params.betterC)
+        {
+            tcs.error("Cannot use try-catch statements with -betterC");
+            return setError();
+        }
+
         uint flags;
         enum FLAGcpp = 1;
         enum FLAGd = 2;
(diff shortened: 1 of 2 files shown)


src/ddmd/statementsem.d
```diff
@@ -3607,6 +3607,14 @@ else
 
     override void visit(TryCatchStatement tcs)
     {
+        //printf("TryCatchStatement.semantic()\n");
+
+        if (global.params.betterC)
```
ibuclaw: Could you use a gate other than betterC?

WalterBright: Another layer of indirection of meaning doesn't seem to add much value here.

jpf91: Try/catch/exceptions can probably work without GC in a custom runtime implementation and are very useful for higher level code. So 👍  for an extra gate variable.

WalterBright: An extra gate variable can always be added if and when it adds value. I'm generally not in favor of layers of semantic indirection otherwise, and I'm known to adding such when the value becomes clear, like in: https://github.com/dlang/dmd/pull/7296/

ibuclaw: This is to make it consistent with #7288. I probably should have taken up on @klickverbot's advice and rename this field to discourage use.

andralex: This matter can be easily revisited later. "The simplest design that works" applies here.

andralex (approved):

src/ddmd/statementsem.d
ibuclaw: > This matter can be easily revisited later.

It was revisited 11 hours later.  Which calls into question why the change wasn't done here instead.

PR #7308, opened 2017-11-12
add changelog entry for -betterC changes to RAII and try-finally

WalterBright:


changelog/raii.dd
```diff
@@ -0,0 +1,12 @@
+Added RAII and try-finally support for -betterC mode.
+
+This also means that scope(exit) statements will also work, because
+they are internally lowered to try-finally statements.
+
+This does not mean that exceptions are supported. Throwing,
+catching, and stack frame unwinding is not supported, as that
+requires support from Phobos.
```
klickverbot: s/Phobos/druntime/

WalterBright: Druntime is part of Phobos, and since the changelog is for users (not core D insiders) who likely are unaware of druntime, I listed it as Phobos.

Geod24: But it will be confusing to people knowing the difference, and that's a fair share of people. Could we say "the runtime" ? That shouldn't be ambiguous to anyone.

WalterBright: I don't think people who know the difference would be confused, as they surely know that druntime is where eh support is.

andralex: @WalterBright 's prerogative

andralex (approved):

PR #6404, opened 2017-01-06
Fix issue 9999,10560 - Deprecate the implicit int to bool conversion for integer literals

LemonBoy:
As stated by @MartinNowak in #1942 the implicit conversion from `int` to `bool` should be removed and today, 4 years later.
A single PR to phobos [1] and one to DRuntime [2] are needed.

[1] https://github.com/dlang/phobos/pull/5019
[2] https://github.com/dlang/druntime/pull/1732

PR #7310, opened 2017-11-12
Fix Issues 9999 & 10560 - [DIP1015] Deprecate & Remove implicit integer & character conversion to bool

JinShil:
Revival of #6404

This is pending approval of this DIP: https://github.com/dlang/DIPs/blob/master/DIPs/DIP1015.md

### DIFF
--- a/src/dmd/backend/cgcod.d
+++ b/src/dmd/backend/cgcod.d
@@ -173,7 +173,7 @@ version (SCPP)
     allregs = ALLREGS;
     pass = PASSinitial;
     Alloca.init();
-    anyiasm = 0;
+    anyiasm = false;
 
     if (config.ehmethod == EHmethod.EH_DWARF)
     {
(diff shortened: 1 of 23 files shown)


wilzbach:
FYI this definitely needs to be a deprecation - building druntime already fails:

```
src/core/thread.d(4223): Error: cannot implicitly convert expression `0` of type `int` to `bool`
src/core/thread.d(4223): Error: cannot implicitly convert expression `0` of type `int` to `bool`
```

wilzbach:
@JinShil if you change the PR back to a deprecation, Jenkins should run through all projects. You can go to "Classic" (the door button on the right) and select "Console output". It's the entire output which you can use to grep for the deprecation message. This is probably the easiest way to assess the impact of this change.

Oh and Phobos won't build with deprecation warnings.

JinShil:
I've been informed that [DIP1015](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1015.md) has been rejected, so I'm closing this.

I urge @WalterBright and @andralex to either close issues [9999](https://issues.dlang.org/show_bug.cgi?id=9999) and [10560](https://issues.dlang.org/show_bug.cgi?id=10560) as _invalid_ or offer an alternate solution.

cc @mdparker

PR #7311, opened 2017-11-13
Fix Issue 17586 - Overriding a deprecated base class function gives no warning

RazvanN7:


PetarKirov (approved):

src/ddmd/dsymbolsem.d
```diff
@@ -4567,6 +4567,10 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                             goto Lintro;
                     }
 
+                    if (fdv.isDeprecated)
+                        deprecation(funcdecl.loc, "%s is overriding the deprecated method %s",
```
JinShil: Please add tick marks (`) around the %s

JinShil (changes requested):

src/ddmd/dsymbolsem.d
RazvanN7: done, thanks!

wilzbach (approved):
Thanks!

wilzbach:
The linked issue will automatically appear in the changelog (with title). Do you think this should be explained in a more detailed message?

mathias-lang-sociomantic:
I think it wouldn't hurt to add a small explanation on every language change. I thought it was more common, but apparently we're not doing it, but we have a large audience, and communicating those changes properly (and their motivation) should be done upfront IMO. But we can just go with the status quo for now :)

PR #7312, opened 2017-11-13
Fix Issue 17585 -  Wrong error message for deprecated overrides

RazvanN7:


ghost (approved):

src/ddmd/dsymbolsem.d
```diff
@@ -4744,9 +4744,9 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             if (funcdecl.isOverride)
             {
                 if (funcdecl.storage_class & STCdisable)
-                    funcdecl.deprecation("overridden functions cannot be annotated @disable");
+                    deprecation(funcdecl.loc, "overriding function %s cannot be annotated @disable", funcdecl.toPrettyChars);
```
JinShil: I think these messages should use a different tense:  "overridden function..." instead of "overriding function...".

JinShil (changes requested):

src/ddmd/dsymbolsem.d
JinShil: Also, surround %s with tick marks (`).

JinShil (changes requested):

src/ddmd/dsymbolsem.d
JinShil: Furthermore, the problem described in the issue is not about the `@disable` attribute, but deprecation. I think the error message should instead be "overridden function ``%s`` cannot be deprecated", but a more general "overridden functions cannot be deprecated" may be even better.

RazvanN7: I am not a native English speaker but isn't overridden used for the function that is being overridden and overriding for the one that overrides?

> Furthermore, the problem described in the issue is not about the @disable attribute, but deprecation

Yes, it solves that one too. But the error regarding disabled was also wrong, so I tried to solve that one too.

> I think the error message should instead be "overridden function %s cannot be deprecated"

That's what it says except that I used overriding.

JinShil: > ...isn't overridden used for the function that is being overridden and overriding for the one that overrides?

Ah, I see what you meant. In that case it should say "The overriding function ``%s`` cannot be ...".  Without the "The", you're telling the user **that the action of** overriding function xx cannot be annotated with @disable, which is a weird thing to say.  Actually, it can be interpreted either way, so at a minimum, prepend the article to disambiguate.

> Yes, it solves that one too. But the error regarding disabled was also wrong, so I tried to solve that one too.

Ok, I understand.  

These would be correct:
"the overriding function ``%s`` cannot be annotated with ``@disable``"
"the overriding function ``%s`` cannot be marked as deprecated"
It still sounds a bit weird due to the overloaded uses of "override" in the English language.

I would prefer something like:
"``%s`` cannot be annotated with ``@disable`` because it is overriding a function in the base class"
"``%s`` cannot be marked as deprecated because it is overriding a function in the base class"

I am a native English speaker, but that doesn't make me an authority on the subject.

RazvanN7: I see. Thanks for the explanation.

> I would prefer something like:
"%s cannot be annotated with @disable because it is overriding a function in the base class"
"%s cannot be marked as deprecated because it is overriding a function in the base class"

Indeed it does sound better that way.

RazvanN7: Done. Thanks!

JinShil (approved):

src/ddmd/dsymbolsem.d
```diff
@@ -4744,9 +4744,13 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             if (funcdecl.isOverride)
             {
                 if (funcdecl.storage_class & STCdisable)
-                    funcdecl.deprecation("overridden functions cannot be annotated @disable");
+                    deprecation(funcdecl.loc,
+                                "`%s` cannot be annotated with @disable because it is overriding a function in the base class",
```
ZombineDev: In general, all code snippets in error messages must be surrounded by back-ticks, in order to get syntax highlighting.
Please also add back-ticks around `@disable` and `deprecated`.

PetarKirov (approved):
Looks good, apart from one minor comment.

src/ddmd/dsymbolsem.d
RazvanN7: Indeed, thanks!

RazvanN7:
That's odd. I just added backticks in the error message and now I have 2 tests failing when phobos is tested.

ZombineDev:
> That's odd. I just added backticks in the error message and now I have 2 tests failing when phobos is tested.

Indeed. I've restarted the failing tests to see if the issue is persistent, as I also see no reason for the failures.

RazvanN7:
@ZombineDev It seems that the restart fixed it.

ZombineDev:
Interestingly enough, the issue was reproduced on TravisCI: https://travis-ci.org/dlang/dmd/jobs/302878738#L3544 so it doesn't seem like a coincidence.

Perhaps @jercaianu could look into it as he's probably the most familiar with the test.

BBasile:
It may be worth targeting stable (for 2077.1). There's been another report today: https://issues.dlang.org/show_bug.cgi?id=17585

ZombineDev:
@RazvanN7 The phobos fix is now merged, so please rebase so we can get TravisCI green.

RazvanN7:
@ZombineDev Done. Thanks!

ZombineDev:
@JinShil you can now add the `auto-merge` tag 🎉

JinShil:
Thanks @ZombineDev for affording me the honors.  My first merge.  It's a proud day.   :smile:

ZombineDev:
@JinShil congratulations!

PR #7313, opened 2017-11-14
remove blockexit dependency on global.errors

WalterBright:
Also start modernizing the `BE` enum.

### DIFF
--- a/src/ddmd/blockexit.d
+++ b/src/ddmd/blockexit.d
@@ -36,34 +36,38 @@ import ddmd.visitor;
  */
 enum BE : int
 {
-    BEnone = 0,
-    BEfallthru = 1,
-    BEthrow    = 2,
-    BEreturn   = 4,
-    BEgoto     = 8,
-    BEhalt     = 0x10,
-    BEbreak    = 0x20,
-    BEcontinue = 0x40,
-    BEerrthrow = 0x80,
-    BEany      = (BEfallthru | BEthrow | BEreturn | BEgoto | BEhalt),
+    none      = 0,
+    fallthru  = 1,
+    throw_    = 2,
+    return_   = 4,
+    goto_     = 8,
+    halt      = 0x10,
+    break_    = 0x20,
+    continue_ = 0x40,
+    errthrow  = 0x80,
+    any       = (fallthru | throw_ | return_ | goto_ | halt),
 }
 
-alias BEnone = BE.BEnone;
-alias BEfallthru = BE.BEfallthru;
-alias BEthrow = BE.BEthrow;
-alias BEreturn = BE.BEreturn;
-alias BEgoto = BE.BEgoto;
-alias BEhalt = BE.BEhalt;
-alias BEbreak = BE.BEbreak;
-alias BEcontinue = BE.BEcontinue;
-alias BEerrthrow = BE.BEerrthrow;
-alias BEany = BE.BEany;
+alias BEnone     = BE.none;
+alias BEfallthru = BE.fallthru;
+alias BEthrow    = BE.throw_;
+alias BEreturn   = BE.return_;
+alias BEgoto     = BE.goto_;
+alias BEhalt     = BE.halt;
+alias BEbreak    = BE.break_;
+alias BEcontinue = BE.continue_;
+alias BEerrthrow = BE.errthrow;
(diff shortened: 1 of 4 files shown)


src/ddmd/blockexit.d
```diff
@@ -36,34 +36,38 @@ import ddmd.visitor;
...
-    BEerrthrow = 0x80,
-    BEany      = (BEfallthru | BEthrow | BEreturn | BEgoto | BEhalt),
+    none      = 0,
+    fallthru  = 1,
+    _throw    = 2,
+    _return   = 4,
+    _goto     = 8,
+    halt      = 0x10,
+    _break    = 0x20,
+    _continue = 0x40,
```
mathias-lang-sociomantic: https://dlang.org/dstyle.html#naming_keywords (underscore should be appended, not prepended)

src/ddmd/blockexit.d
```diff
@@ -36,34 +36,38 @@ import ddmd.visitor;
...
+alias BEnone     = BE.none;
+alias BEfallthru = BE.fallthru;
+alias BEthrow    = BE._throw;
+alias BEreturn   = BE._return;
+alias BEgoto     = BE._goto;
+alias BEhalt     = BE.halt;
+alias BEbreak    = BE._break;
+alias BEcontinue = BE._continue;
+alias BEerrthrow = BE.errthrow;
+alias BEany      = BE.any;
```
mathias-lang-sociomantic: Is there any plan to get rid of those eventually ?

WalterBright: yes

src/ddmd/blockexit.d
```diff
@@ -36,34 +36,38 @@ import ddmd.visitor;
...
- * Only valid after semantic analysis
+ * Determine mask of ways that a statement can exit.
+ *
+ * Only valid after semantic analysis.
  * Params:
+ *   s = statement to check for block exit status
+ *   func = function that statement s is in
  *   mustNotThrow = generate an error if it throws
  * Returns:
  *   BExxxx
```
mathias-lang-sociomantic: Maybe that part of the doc can be updated as well since we have a proper enum now

src/ddmd/blockexit.d
```diff
@@ -94,7 +98,7 @@ int blockExit(Statement s, FuncDeclaration func, bool mustNotThrow)
 
         override void visit(ErrorStatement s)
         {
-            result = BEany;
+            result = BE.none;
```
mathias-lang-sociomantic: That doesn't look like a refactor...

WalterBright: This is completely trivial. My complaint with mixing refactoring with semantic changes is when there'd be a hundred lines of code moved about that completely obfuscated what the actual fix was. That is not the case here.

src/ddmd/blockexit.d
```diff
@@ -36,34 +36,38 @@ import ddmd.visitor;
...
+    none      = 0,
+    fallthru  = 1,
+    throw_    = 2,
+    return_   = 4,
+    goto_     = 8,
+    halt      = 0x10,
+    break_    = 0x20,
+    continue_ = 0x40,
+    errthrow  = 0x80,
+    any       = (fallthru | throw_ | return_ | goto_ | halt),
```
andralex: so nice

src/ddmd/blockexit.d
```diff
@@ -36,34 +36,38 @@ import ddmd.visitor;
...
+alias BEnone     = BE.none;
+alias BEfallthru = BE.fallthru;
+alias BEthrow    = BE.throw_;
+alias BEreturn   = BE.return_;
+alias BEgoto     = BE.goto_;
+alias BEhalt     = BE.halt;
+alias BEbreak    = BE.break_;
+alias BEcontinue = BE.continue_;
+alias BEerrthrow = BE.errthrow;
+alias BEany      = BE.any;
```
andralex: meh vertical alignment sux

andralex: I assume these are for backward compat and will be phased out?

andralex (approved):

src/ddmd/blockexit.d
WalterBright: yes

PR #5565, opened 2016-03-23
Issue 15816 - ICE void ddmd.dclass.__assert(int) with error: anonymous classes not allowed

9rnsr:
An invalid anonymous class declaration has no chance to get internal `ident` (= "__anonclassXX"), because its `parent` is set by `Dsymbol.addMember` before the `semantic` is invoked.

Name all `ClassDeclaration` instances during the constructions, then the problematic assertion can be removed.

PR #7315, opened 2017-11-14
Fix Issue 17371 -  [REG 2.074.0] di generation broken for anonymous classes

RazvanN7:
The header generation takes into account if a class is anonymous by invoking the final method Dsymbol.isAnonymous. The specified method verifies if a symbol is anonymous by checking if the id is null, but when a class is instantiated with a null id, it generates one [1] making isAnonymous to report a wrong answer. The easiest way to fix this is to declare a new bool field (isanon) which is set to true when a null id is passed to the instantiation of a class declaration and the header generation can check for that field instead of the wrong isAnonymous method.

[1] https://github.com/dlang/dmd/pull/5565/files#diff-ddbaa5e9ca3d5c90a425a9dfafaf1734R225

### DIFF
--- a/src/ddmd/aggregate.h
+++ b/src/ddmd/aggregate.h
@@ -247,6 +247,18 @@ struct ClassFlags
     };
 };
 
+enum ClassKind
+{
+    /// the class is a d(efault) class
+    d,
+    /// the class is a C++ interface
+    cpp,
+    /// the class is an Objective-C class/interface
+    objc,
+    /// the class is anonymous
+    anonymous,
+};
+
 class ClassDeclaration : public AggregateDeclaration
 {
 public:
@@ -273,11 +285,10 @@ public:
 
     TypeInfoClassDeclaration *vclassinfo;       // the ClassInfo object for this ClassDeclaration
     bool com;                           // true if this is a COM class (meaning it derives from IUnknown)
-    bool cpp;                           // true if this is a C++ interface
-    bool isobjc;                        // true if this is an Objective-C class/interface
-    bool isscope;                       // true if this is a scope class
+    bool stack;                       // true if this is a scope class
+    ClassKind classKind;               // specifies the linkage type
+
     Abstract isabstract;                // 0: fwdref, 1: is abstract class, 2: not abstract
-    int inuse;                          // to prevent recursive attempts
     Baseok baseok;                      // set the progress of base classes resolving
     Symbol *cpp_type_info_ptr_sym;      // cached instance of class Id.cpp_type_info_ptr
 
--- a/src/ddmd/dclass.d
+++ b/src/ddmd/dclass.d
@@ -179,6 +179,24 @@ struct ClassFlags
     alias hasDtor = Enum.hasDtor;
 }
 
+/**
+ * The ClassKind enum is used in ClassDeclaration AST nodes
+ * to specify the linkage type of the class or if it is an
+ * anonymous class. If the class is anonymous it is also
+ * considered to be a D class.
+ */
+enum ClassKind : int
+{
+    /// the class is a d(efault) class
+    d,
+    /// the class is a C++ interface
+    cpp,
+    /// the class is an Objective-C class/interface
+    objc,
+    /// the class is anonymous
+    anonymous,
+}
+
 /***********************************************************
  */
 extern (C++) class ClassDeclaration : AggregateDeclaration
@@ -212,20 +230,32 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
     // the ClassInfo object for this ClassDeclaration
     TypeInfoClassDeclaration vclassinfo;
 
-    bool com;           // true if this is a COM class (meaning it derives from IUnknown)
-    bool cpp;           // true if this is a C++ interface
-    bool isobjc;        // true if this is an Objective-C class/interface
-    bool isscope;       // true if this is a scope class
-    private bool inuse;          // to prevent recursive attempts
+    // true if this is a COM class
+    bool com;
+
+    /// true if this is a scope class
+    bool stack;
+
+    /// specifies whether this is a D, C++, Objective-C or anonymous class/interface
+    ClassKind classKind;
+
+    /// to prevent recursive attempts
+    private bool inuse;
+
     Abstract isabstract;
-    Baseok baseok;      // set the progress of base classes resolving
+
+    /// set the progress of base classes resolving
+    Baseok baseok;
 
     Symbol* cpp_type_info_ptr_sym;      // cached instance of class Id.cpp_type_info_ptr
 
     final extern (D) this(Loc loc, Identifier id, BaseClasses* baseclasses, Dsymbols* members, bool inObject)
     {
         if (!id)
+        {
             id = Identifier.generateId("__anonclass");
+            classKind = ClassKind.anonymous;
+        }
         assert(id);
 
         super(loc, id);
@@ -584,7 +614,7 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
 
             alignsize = baseClass.alignsize;
             structsize = baseClass.structsize;
-            if (cpp && global.params.isWindows)
+            if (classKind == ClassKind.cpp && global.params.isWindows)
                 structsize = (structsize + alignsize - 1) & ~(alignsize - 1);
         }
         else if (isInterfaceDeclaration())
@@ -599,7 +629,7 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
         {
             alignsize = Target.ptrsize;
             structsize = Target.ptrsize;      // allow room for __vptr
-            if (!cpp)
+            if (classKind != ClassKind.cpp)
                 structsize += Target.ptrsize; // allow room for __monitor
         }
 
@@ -817,7 +847,7 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
 
     final bool isCPPclass() const
     {
-        return cpp;
+        return classKind == ClassKind.cpp;
     }
 
     bool isCPPinterface() const
@@ -932,7 +962,7 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
      */
     int vtblOffset() const
     {
-        return cpp ? 0 : 1;
+        return classKind == ClassKind.cpp ? 0 : 1;
     }
 
     /****************************************
@@ -973,7 +1003,7 @@ extern (C++) final class InterfaceDeclaration : ClassDeclaration
         if (id == Id.IUnknown) // IUnknown is the root of all COM interfaces
         {
             com = true;
-            cpp = true; // IUnknown is also a C++ interface
+            classKind = ClassKind.cpp; // IUnknown is also a C++ interface
         }
     }
 
@@ -991,9 +1021,9 @@ extern (C++) final class InterfaceDeclaration : ClassDeclaration
         auto sc2 = super.newScope(sc);
         if (com)
             sc2.linkage = LINKwindows;
-        else if (cpp)
+        else if (classKind == ClassKind.cpp)
             sc2.linkage = LINKcpp;
-        else if (isobjc)
+        else if (classKind == ClassKind.objc)
             sc2.linkage = LINKobjc;
         return sc2;
     }
@@ -1087,7 +1117,7 @@ extern (C++) final class InterfaceDeclaration : ClassDeclaration
 
     override bool isCPPinterface() const
     {
-        return cpp;
+        return classKind == ClassKind.cpp;
     }
 
     override bool isCOMinterface() const
--- a/src/ddmd/hdrgen.d
+++ b/src/ddmd/hdrgen.d
@@ -1754,7 +1754,7 @@ public:
 
     override void visit(ClassDeclaration d)
     {
-        if (!d.isAnonymous())
+        if (d.classKind != ClassKind.anonymous)
         {
             buf.writestring(d.kind());
             buf.writeByte(' ');
@@ -1781,7 +1781,8 @@ public:
     {
         if (!d || !d.baseclasses.dim)
             return;
-        buf.writestring(" : ");
+        if (d.classKind != ClassKind.anonymous)
+            buf.writestring(" : ");
         foreach (i, b; *d.baseclasses)
         {
             if (i)
--- a/src/ddmd/todt.d
+++ b/src/ddmd/todt.d
@@ -719,7 +719,7 @@ private void membersToDt(AggregateDeclaration ad, DtBuilder dtb,
         {
             dtb.xoff(toVtblSymbol(concreteType), 0);  // __vptr
             offset = Target.ptrsize;
-            if (!cd.cpp)
+            if (!cd.classKind == ClassKind.cpp)
             {
                 dtb.size(0);              // __monitor
                 offset += Target.ptrsize;
--- a/src/ddmd/declaration.d
+++ b/src/ddmd/declaration.d
@@ -1343,7 +1343,7 @@ extern (C++) class VarDeclaration : Declaration
 
                 // Destroying C++ scope classes crashes currently. Since C++ class dtors are not currently supported, simply do not run dtors for them.
                 // See https://issues.dlang.org/show_bug.cgi?id=13182
-                if (cd.cpp)
+                if (cd.classKind == ClassKind.cpp)
                 {
                     break;
                 }
--- a/src/ddmd/dsymbolsem.d
+++ b/src/ddmd/dsymbolsem.d
@@ -4497,7 +4497,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
 
                 /* These quirky conditions mimic what VC++ appears to do
                  */
-                if (global.params.mscoff && cd.cpp &&
+                if (global.params.mscoff && cd.classKind == ClassKind.cpp &&
                     cd.baseClass && cd.baseClass.vtbl.dim)
                 {
                     /* if overriding an interface function, then this is not
@@ -4523,7 +4523,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                 {
                     //printf("\tintroducing function %s\n", toChars());
                     funcdecl.introducing = 1;
-                    if (cd.cpp && Target.reverseCppOverloads)
+                    if (cd.classKind == ClassKind.cpp && Target.reverseCppOverloads)
                     {
                         // with dmc, overloaded functions are grouped and in reverse order
                         funcdecl.vtblIndex = cast(int)cd.vtbl.dim;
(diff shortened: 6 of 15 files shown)


RazvanN7:
I find it awkward to do extra processing for some information which is known at some point but gets lost.

stefan-koch-sociomantic:
I find it awkward to add a freaking 8 bit, but with padding most likely 32bit, to store a condition which is almost never true.

ZombineDev:
That's easily addressable by merging the various `is*` fields into a bit-field.

RazvanN7:
@ZombineDev thanks for the idea.

@stefan-koch-sociomantic is this ok?

src/ddmd/dclass.d
```diff
@@ -179,6 +179,12 @@ struct ClassFlags
     alias hasDtor = Enum.hasDtor;
 }
 
+enum ubyte IScom = 1;     // mask for COM class (meaning it derives from IUnknown)
+enum ubyte IScpp = 2;     // mask for a C++ interface
+enum ubyte ISobjc = 4;    // mask for an Objective-C class/interface
+enum ubyte ISscope = 8;   // mask for a scope class
+enum ubyte ISanon = 16;   // mask for an anonymous class
```
ZombineDev: Can you make this an enum, like https://github.com/dlang/dmd/blob/db221adb6a865ca0f3b1c24d922d9c3d5742a097/src/ddmd/blockexit.d#L37 ?

RazvanN7: is it going to use a single ubyte that way?

RazvanN7: if yes, of course I can

RazvanN7: Done

src/ddmd/dclass.d
```diff
@@ -179,6 +179,15 @@ struct ClassFlags
     alias hasDtor = Enum.hasDtor;
 }
 
+enum IS : ubyte
+{
+    com = 1,     // mask for COM class (meaning it derives from IUnknown)
+    cpp = 2,     // mask for a C++ interface
+    objc = 4,    // mask for an Objective-C class/interface
+    _scope = 8,   // mask for a scope class
```
ZombineDev: The convention is: `keyword_`. Was recently discussed here https://github.com/dlang/dmd/pull/7313#discussion_r150784331

stefan-koch-sociomantic:
Other then the very unclear meaning of `IS` it's fine.
Maybe call it `classflags` or something.

I approve of this change.

src/ddmd/dclass.d
```diff
@@ -179,6 +179,15 @@ struct ClassFlags
...
+    anon = 16,   // mask for an anonymous class
+}
```
jacob-carlborg: Please make these Ddoc comments.

src/ddmd/dclass.d
```diff
@@ -212,10 +221,7 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
     // the ClassInfo object for this ClassDeclaration
     TypeInfoClassDeclaration vclassinfo;
 
-    bool com;           // true if this is a COM class (meaning it derives from IUnknown)
-    bool cpp;           // true if this is a C++ interface
-    bool isobjc;        // true if this is an Objective-C class/interface
-    bool isscope;       // true if this is a scope class
+    ubyte classProperties; // bitfield that keeps track if the class is com, cpp, objc, scope or anonymous
```
jacob-carlborg: Same here, Ddoc comment.

jacob-carlborg:
Will this fix the original issue I had, generating a `.di` file for the following code:

```d

interface Foo {}

class Bar
{
    this()
    {
        auto foo = new class () Foo {};
    }
}
```

If that's the case, perhaps include the above as a test as well.

RazvanN7:
@jacob-carlborg Yes, it does fix that case as well. It doesn't matter where the anonymous class is defined, the code generated in the .di file is the same.

RazvanN7:
@ZombineDev I updated the code to used **scope_** instead of **_scope**.

@stefan-koch-sociomantic I changed it to CFM which Class Flags Masks.

@jacob-carlborg I updated the documentation and added the required test.

Thank you all.

src/ddmd/dclass.d
```diff
@@ -179,6 +179,26 @@ struct ClassFlags
     alias hasDtor = Enum.hasDtor;
 }
 
+/**
+ * CFM stands for Class Flag Masks. Each ClassDeclaration AST node
+ * allocates a bitfield which stores various class properties.
+ * These properties can be retrieved or updated by using
+ * these masks.
+ */
+enum CFM : ubyte
```
jacob-carlborg: I highly recommend calling it `ClassFlagMasks` instead of `CFM`. That will make everything more clear and avoid the extra explanation in the documentation.

src/ddmd/dclass.d
```diff
@@ -212,10 +232,10 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
     // the ClassInfo object for this ClassDeclaration
     TypeInfoClassDeclaration vclassinfo;
 
-    bool com;           // true if this is a COM class (meaning it derives from IUnknown)
-    bool cpp;           // true if this is a C++ interface
-    bool isobjc;        // true if this is an Objective-C class/interface
-    bool isscope;       // true if this is a scope class
+    /* bitfield that keeps track if the class is com, cpp, objc, scope or anonymous
+     */
+    ubyte classProperties;
```
jacob-carlborg: I recommend using Ddoc comment for this as well. Keep in mind that the compiler is exposed as a library. In my opinion all public/protected symbols should have Ddoc documentation. I don't expect that the documentation will be added to all symbols at once, but we should avoid making it worse 😃.

src/ddmd/dclass.d
```diff
@@ -179,6 +179,26 @@ struct ClassFlags
...
+    com = 1,
+    /// mask for a C++ interface
+    cpp = 2,
+    /// mask for an Objective-C class/interface
+    objc = 4,
+    /// mask for a scope class
+    scope_ = 8,
+    /// mask for an anonymous class
+    anon = 16,
+}
```
jacob-carlborg: At least some of these are mutual exclusive, if not all? `com`, `cpp`, `objc` and `anon` are at at least mutual exclusive. Then I recommend using `CFM` as the type for `classProperties` and not a mask. Then keep `scope_` as a separate bool instance variable.

andralex:
>I find it awkward to add a freaking 8 bit, but with padding most likely 32bit, to store a condition which is almost never true.

This leads to a bit of overcooking (and please let's control the language). The waste in layout needs to be balanced with a decrease in code maintainability (masking needed all over) and extra work for packing and unpacking. Also, overall the memory wasted needs to be assessed and it seems the number of `ClassDeclaration` instances scales with handwritten code, i.e. it's slow growing. We're looking at hundreds/thousands. It would be tenuous to measure any performance difference one way or another. Speculating that there would, and making the code more complicated based on the speculation, would be one of those premature optimization cases.

Besides, there are simple ways to harvest a better layout for "free", without impacting maintainability. The current layout goes (on 64 bit):

```d
    ...
    TypeInfoClassDeclaration vclassinfo; // reference (8 bytes)
    ubyte classProperties;
    Abstract isabstract; // int - could be ubyte
    int inuse;      // int - could be bool
    Baseok baseok;      // int - could be ubyte
```
Let's say we keep Abstract and Baseok int, although they are enumerated values with only a handful of cases and could be trivially packed in a ubyte just by changing their declaration. The one thing we can improve (perhaps with this pr) is `inuse`. Check it out:

```d
            if (!inuse)
            {
                // must semantic on base class/interfaces
                ++inuse;
                dsymbolSemantic(this, null);
                --inuse;
            }
```d

That's the only use in this module (!) and making the variable private reveals there's no other use of it outside the module. It follows that inuse should be a bool, which improves not only layout but also understandability of the code:

```d
            if (!inuse)
            {
                // must semantic on base class/interfaces
                inuse = true;
                dsymbolSemantic(this, null);
                inuse = false;
            }
```d

So, @RazvanN7 with this pr:

* Please return to the use of `bool`
* Please make the introduced field private if possible
* Please change `inuse` to bool and place it together with the other bools in the layout.
* Please make `inuse` private since you changed it

Thanks!

File: src/ddmd/dclass.d [hunk 8]
```diff
@@ -179,6 +179,26 @@ struct ClassFlags
...
+    /// mask for COM class (meaning it derives from IUnknown)
+    com = 1,
+    /// mask for a C++ interface
+    cpp = 2,
+    /// mask for an Objective-C class/interface
+    objc = 4,
+    /// mask for a scope class
+    scope_ = 8,
+    /// mask for an anonymous class
+    anon = 16,
```
jacob-carlborg: Please don't use abbreviations.

jacob-carlborg:
@andralex I don't think we need to worry about the layout. I also don't think we should touch `inuse` in this PR, it's unrelated to what's actually changed in this PR.

In general the DMD code base needs some cleanup. These integers that are used as bools are remains from the C++ days and even written in an almost C like style. Traditional C doesn't have bools.

jacob-carlborg:
@RazvanN7 is there any verification that the `.di` files you added the tests to actually compiles?

src/ddmd/dclass.d
WalterBright: Or just `ClassFlags`

src/ddmd/dclass.d
WalterBright: 'Mask' implies they are bits being OR'd together. But exclusive values means that isn't so. I recommend using one field for the exclusive types, and another field for modifiers, instead of mushing them together. I'm guilty of doing the mushing for a long time, and it isn't worth it unless one is desperate to minimize memory usage.
So, I recommend `ClassKind` with members `D`, `com`, `C++`, `anon`. Then, `ClassFlags` with members `stack`. (`scope_` is a bit overloaded in D, and this really is a stack class)

src/ddmd/dclass.d
WalterBright: `classProperties` has no obvious relationship with the `ClassFlagMask` values. Call the fields `classKind` and `classFlags`, to match up with `ClassKind` values and `ClassFlags` flag bits.

WalterBright:
Also, when changing the ClassDeclaration fields, the corresponding .h file has to change, too. Otherwise,  @ibuclaw 's code gets messed up.

WalterBright:
@andralex makes good points (I read them after I'd made my comments). Clarity could be improved with the `ClassKind` enum, since having `bool com, cpp;` gives no implication that they are exclusive, whereas an enumeration type does. The latter is just better design.

src/ddmd/dclass.d
RazvanN7: **com** and **cpp** aren't exclusive in this case [1] 

[1] https://github.com/dlang/dmd/blob/master/src/ddmd/dclass.d#L976

src/ddmd/dclass.d
RazvanN7: ClassFlags is already taken and is a struct [1], but it's not used in dclass at all; toobj.d is the only file that uses it, so I guess we can move it there and use ClassFlags in dclass. I thought about modifying the ClassFlags struct so that it accommodates the anonymous, stack and objc flags as well, but it seems, if I am not mistaking, that the backend expects a certain layout of that class.

[1] https://github.com/dlang/dmd/blob/master/src/ddmd/dclass.d#L154

RazvanN7:
@andralex @WalterBright what do you think about this : 

1. We create an enum : ClassKind with `bool` members  `cpp`, `objc`, `anonymous`,  and keep `stack` , `com` and inuse as bool fields since these can overlap with the former mentioned attributes; I find this to be the best compromise between layout and clarity. 

2. In a different PR we move `struct ClassFlags` [1] in toobj.d since that is the only file the struct is used.

[1] https://github.com/dlang/dmd/blob/master/src/ddmd/dclass.d#L154

RazvanN7:
@jacob-carlborg 

>@RazvanN7 is there any verification that the .di files you added the tests to actually compiles?

Sorry, but I don't understand the question. 
If you are asking how do I know that the test provided compiles, then the answer is : the header*.d files are compiled and the di files are generated. Ig any errors would occur, the autotester would report them or the generation would fail (I assume),

andralex:
Putting mutually exclusive flags in an enum sounds like a good thing to do. 

> I also don't think we should touch inuse in this PR, it's unrelated to what's actually changed in this PR.

We have been erring on the side of bureaucracy lately. Clearly doing swaths of unrelated changes in one PR is unacceptable but I'll take a good small unrelated improvement with a PR. I don't want to police people into ultimately doing less work.

In this particular case, creating a new PR takes actually just a bit longer than asking said PR should be created. https://github.com/dlang/dmd/pull/7349

src/ddmd/dclass.d
jacob-carlborg: If `ClassFlags` is already taken, what about `ClassKind`?

jacob-carlborg:
> Sorry, but I don't understand the question.
If you are asking how do I know that the test provided compiles, then the answer is : the header*.d files are compiled and the di files are generated. If any errors would occur, the autotester would report them or the generation would fail (I assume).

@RazvanN7 yes, but are the generated `.di` files compiled as well? That's how I originally discovered this issue. The compiler generated `.di` files with invalid code.

jacob-carlborg:
@RazvanN7 https://github.com/dlang/dmd/pull/7315#issuecomment-345965550, I like this approach.

RazvanN7:
@andralex @WalterBright how does this look like?

>  but are the generated .di files compiled as well?

@jacob-carlborg No,  the postscript is just doing a diff between what the output is and what is expected. I assume that folks compile their .di files before pasting the expected output. At least, that's what I am doing, so for this particular case it is a certainty that the .di file compiles.

src/ddmd/dclass.d
```diff
@@ -212,20 +234,32 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
...
-    bool isscope;       // true if this is a scope class
-    private bool inuse;          // to prevent recursive attempts
+    // true if this is a COM class
+    bool com;
+
+    /// true if this is a scope class
+    bool stack;
+
+    /// specifies whether this is a d, cpp, objc or anonymous class
+    ClassKind classkind;
```
jacob-carlborg: Please use camel casing, i.e. `classKind`.

src/ddmd/dclass.d
```diff
@@ -584,7 +618,7 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
 
             alignsize = baseClass.alignsize;
             structsize = baseClass.structsize;
-            if (cpp && global.params.isWindows)
+            if (classkind == CKcpp && global.params.isWindows)
```
jacob-carlborg: I suggest using the full name, i.e. `ClassKind.cpp`, it's more D.

src/ddmd/dclass.d
```diff
@@ -179,6 +179,28 @@ struct ClassFlags
...
+    cpp,
+    /// the class is an Objective-C class/interface
+    objc,
+    /// the class is anonymous
+    anonymous,
+}
+
+alias CKcpp = ClassKind.cpp;
+alias CKobjc = ClassKind.objc;
+alias CKanonymous = ClassKind.anonymous;
```
jacob-carlborg: I suggest to remove these aliases and always use the fully name of the enum members, it's more D like.

src/ddmd/hdrgen.d
```diff
@@ -1754,7 +1754,7 @@ public:
 
     override void visit(ClassDeclaration d)
     {
-        if (!d.isAnonymous())
```
jacob-carlborg: I count to 28 usages of the `isAnonymous` method. Why is only this one changed? Perhaps the implementation of the method should be updated to use the new enum.

src/ddmd/dclass.d
```diff
@@ -212,20 +234,32 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
...
-    bool isobjc;        // true if this is an Objective-C class/interface
-    bool isscope;       // true if this is a scope class
-    private bool inuse;          // to prevent recursive attempts
+    // true if this is a COM class
+    bool com;
+
+    /// true if this is a scope class
+    bool stack;
+
+    /// specifies whether this is a d, cpp, objc or anonymous class
```
jacob-carlborg: Nitpick:

* I think it looks better to use the official names of the languages in the documentation, i.e. `D`, `C++` and `Objective-C`
* `class` -> `class/interface`

src/ddmd/hdrgen.d
RazvanN7: `isAnonymous` is a final method in the Dsymbol AST node. A ClassDeclaration inherits a Dsymbol so changing isAnonymous everywhere will break code. What we can do is change the check for class declarations or make `isAnonymous` not final and override it in the ClassDeclaration.

RazvanN7: of the 28 usages, only 1 is actually from a class declaration and I updated the code.

src/ddmd/dclass.d
RazvanN7: Maybe we should do some refactoring and get rid of more aliases?

src/ddmd/hdrgen.d
jacob-carlborg: Ah, ok, I actually didn't check how it was used, I only counted the occurrences. But I feel there's a risk of using the wrong way to identify if a class is an anonymous class when there are both the method `isAnonymous` and the instance variable `classKind`. Perhaps it's best to override it in `ClassDeclaration`, just to be on the safe side.

src/ddmd/dclass.d
jacob-carlborg: I agree, but let's start by not making the adding more 😃.

jacob-carlborg:
I'm pretty satisfied with this, there are only two things missing:

* Update the necessary header files, at least `aggregate.h`
* Squash all commits as the last step

RazvanN7:
@jacob-carlborg Done. Thanks!

src/ddmd/aggregate.h
```diff
@@ -247,6 +247,18 @@ struct ClassFlags
...
+{
+    /// the class is a d(efault) class
+    d,
+    /// the class is a C++ interface
+    cpp,
+    /// the class is an Objective-C class/interface
+    objc,
+    /// the class is anonymous
+    anonymous,
+};
```
jacob-carlborg: Won't this be an `int`, which will not match up with the size of `ubyte` on the D side?

RazvanN7: enums in C++ are ints. Not sure what the fix should be here. Maybe declare an ClassKind as an unsigned char?

andralex (approved):

MartinNowak:
This introduced a regression as signified by the failing test-suite.
Since ci.dlang.org runs stably for quite a while and nobody checks it (not even dlang-bot), I'll make it required for dmd again once this regression is fixed.

[18020 – [Reg 2.078] no property opCmp for anon class](https://issues.dlang.org/show_bug.cgi?id=18020)

MartinNowak:
@RazvanN7 could you please have a look?

src/ddmd/todt.d
```diff
@@ -719,7 +719,7 @@ private void membersToDt(AggregateDeclaration ad, DtBuilder dtb,
         {
             dtb.xoff(toVtblSymbol(concreteType), 0);  // __vptr
             offset = Target.ptrsize;
-            if (!cd.cpp)
+            if (!cd.classKind == ClassKind.cpp)
```
JinShil: Should this have been `cd.classKind != ClassKind.cpp`?

JinShil:
> This introduced a regression as signified by the failing test-suite.

Potential fix submitted at https://github.com/dlang/dmd/pull/7315

src/ddmd/todt.d
RazvanN7: Yes, indeed. please fix in #7378

JinShil: Done.  Thanks

PR #7316, opened 2017-11-14
Fix Issue 17382 - void main(){}pragma(msg,main()); crashes DMD

RazvanN7:


stefan-koch-sociomantic:
NO!
That'll generate an uncompilable version.
As well as silencing real errors! 
I suggest you look for a different fix.

RazvanN7:
>NO!
That'll generate an uncompilable version.

Can you be more specific?

stefan-koch-sociomantic:
the generated header will have `<void>` in it which has no place in our langauge.
Also If the case happens which would have asserted before it will now get swallowed.

RazvanN7:
You might be right, however asserts are not errors; assert(0) is used on code paths that should never be taken and the user should never be faced with a failed assert so from this point of view no error is swollen.

On the  other hand, if you use pragma(msg, foo()) where foo is declared as void foo() {} the output is going to be **<void>** . Now, in the case of this PR, I agree that there might be cases where **<void>** is outputted and it shouldn't but I don't know any and it seems that the tests all pass. So, when I was asking you to be specific I was expecting an example. Thank you.

stefan-koch-sociomantic:
I cannot think of an example as of now. 
As the assert would have triggered in such a case.
However that is no excuse to open us up to more subtle bugs.

Besides having the hdrgen write-out invalid D is hardly an acceptable solution.

I'd rather fix this at the point where the main-symbol is rewritten.

MetaLang:
Excuse the drive-by comment given that I know nothing about the compiler, but why does the compiler even try to const-fold `main`?

andralex:
It doesn't matter that it's called main, does it? @RazvanN7 please arrange that a compile-time error with a good message is issued if someone attempts to `pragma msg` a `void` expression. If anyone does want to, there are many means such as `pragma(msg, (main(), ""));` etc

ZombineDev:
Yes, it should just work. For example the following snippet:  `void foo() {} pragma(msg, foo());` makes the compiler print `<void>`.

RazvanN7:
@andralex Is this ok?

src/ddmd/dsymbolsem.d
```diff
@@ -2917,6 +2917,11 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                     e = resolveProperties(sc, e);
                     sc = sc.endCTFE();
                     // pragma(msg) is allowed to contain types as well as expressions
+                    if (e.type && e.type.ty == Tvoid)
+                    {
+                        error(pd.loc, "Cannot pass argument `%s` to `pragma msg` because it is `void`", e.toChars);
```
nemanja-boric-sociomantic: Missing `()` at `toChars` call (style doesn't match anywhere in this module).

andralex (approved):
"has type void" is more pedantically correct, but this is fine as is - thanks

PR #7317, opened 2017-11-14
VRP for and,or,xor + Refactoring intrange.d + Fix Issue 10310

somzzz:
As per discussions with @andralex @WalterBright and @tgehr , this PR is the first step to improving VRP in dmd. Reimplemented ```and```, ```or```, ```xor``` over ```IntRange``` using the algorithms designed by @tgehr : 
 https://github.com/tgehr/d-compiler/blob/master/vrange.d

Refactored ```intrange.d``` to use DMD2 style op overloads. Moved implementation for ```and```, ```or```, ```xor``` from ```dcast.d```. This is a first step, the other operators will follow - this PR is quite big already.

Also fixing bug related to imprecise ```xor``` bounds:
https://issues.dlang.org/show_bug.cgi?id=10310

### DIFF
--- a/src/ddmd/intrange.d
+++ b/src/ddmd/intrange.d
@@ -30,6 +30,7 @@ struct SignExtendedNumber
 {
     uinteger_t value;
     bool negative;
+
     static SignExtendedNumber fromInteger(uinteger_t value_)
     {
         return SignExtendedNumber(value_, value_ >> 63);
@@ -77,7 +78,26 @@ struct SignExtendedNumber
             return 0;
     }
 
-    SignExtendedNumber opNeg() const
+    private alias SignExtendedNumber R;
+
+    R opUnary(string op : "++")()
+    {
+        if (value != UINT64_MAX)
+            ++value;
+        else if (negative)
+        {
+            value = 0;
+            negative = false;
+        }
+        return this;
+    }
+
+    R opUnary(string op : "~")() const
+    {
+        return SignExtendedNumber(~value);
+    }
+
+    R opUnary(string op : "-")() const
     {
         if (value == 0)
             return SignExtendedNumber(-cast(ulong)negative);
@@ -85,11 +105,26 @@ struct SignExtendedNumber
             return SignExtendedNumber(-value, !negative);
     }
 
-    SignExtendedNumber opAdd(const SignExtendedNumber a) const
+    R opBinary(string op : "&")(R rhs) const
     {
-        uinteger_t sum = value + a.value;
-        bool carry = sum < value && sum < a.value;
-        if (negative != a.negative)
+        return SignExtendedNumber(value & rhs.value);
+    }
+
+    R opBinary(string op : "|")(R rhs)
+    {
+        return SignExtendedNumber(value | rhs.value);
+    }
+
+    R opBinary(string op : "^")(R rhs)
+    {
+        return SignExtendedNumber(value ^ rhs.value);
+    }
+
+    R opBinary(string op : "+")(R rhs)
+    {
+        uinteger_t sum = value + rhs.value;
+        bool carry = sum < value && sum < rhs.value;
+        if (negative != rhs.negative)
             return SignExtendedNumber(sum, !carry);
         else if (negative)
             return SignExtendedNumber(carry ? sum : 0, true);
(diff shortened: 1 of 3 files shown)


src/ddmd/intrange.d
```diff
@@ -494,4 +524,233 @@ struct IntRange
...
+
+    R opBinary(string op:"^")(R rhs) const
+    {
+        return this & ~rhs | ~this & rhs;
+    }
+
+    R opBinary(string op:"+")(R rhs)
+    {
+        // Not yet implemented
+        assert(0);
```
andralex: Reiterating suggestion: on overflow, conservatively return the maximum range.

src/ddmd/intrange.d
```diff
@@ -494,4 +524,233 @@ struct IntRange
...
+    R opBinary(string op:"^^")(R rhs)
+    {
+        // Not yet implemented
+        assert(0);
+    }
+
+private:
+    // Credits to Timon Gehr maxOr, minOr, maxAnd, minAnd
+    // https://github.com/tgehr/d-compiler/blob/master/vrange.d
+    static SignExtendedNumber maxOr(R lhs, R rhs){
```
andralex: Formatting issues - brace should be on its own line and the function body should be indented.

src/ddmd/intrange.d
```diff
@@ -494,4 +524,233 @@ struct IntRange
...
+    uinteger_t x = 0;
+    auto xor = lhs.imax.value ^ rhs.imax.value;
+    auto and = lhs.imax.value & rhs.imax.value;
+
+    for(uinteger_t d = 1LU << (8 * uinteger_t.sizeof - 1); d; d >>= 1)
```
andralex: space after if etc (adjust from Timon's convention to ours)

andralex (changes requested):
cc @tgehr @WalterBright

src/ddmd/intrange.d
somzzz: This PR is only for the bitops, the old code doesn't use operator overloads for ```IntRange```, so currently there is no code that calls into ```assert(0)```. Most of the implementation for these operators is split between ```dcast.d``` eg
https://github.com/dlang/dmd/blob/master/src/ddmd/dcast.d#L3613
and ```SignExtendedNumber``` and should be refactored. The purpose of adding those stubs is to suggest this.

I wanted to do this in different PRs. Similarly, at the end, if we decide we don't need ```SignExtendedNumber``` anymore, we can remove it in its own PR.

If you think it's better to handle all VRP in one go, we can do that.

src/ddmd/intrange.d
```diff
@@ -494,4 +524,233 @@ struct IntRange
             nonNegRange.imax = imax;
         }
     }
+
+    private alias IntRange R;
+
+    R copy() const
```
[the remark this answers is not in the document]
somzzz: There isn't any ```IntRange``` constructor which takes another ```IntRange``` as param.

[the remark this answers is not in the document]
andralex: That would be implicitly defined, or you can define it if needed.

src/ddmd/intrange.d
```diff
@@ -77,35 +78,70 @@ struct SignExtendedNumber
             return 0;
     }
 
-    SignExtendedNumber opNeg() const
+    private alias SignExtendedNumber R;
```
andralex: This is a distraction, a future PR should eliminate it and rename `SignExtendedNumber` to `SignExtended`.

src/ddmd/intrange.d
```diff
@@ -77,35 +78,70 @@ struct SignExtendedNumber
...
+        {
+            value = 0;
+            negative = false;
+        }
+        return this;
+    }
+
+    R opUnary(string op : "~")() const
+    {
+        return SignExtendedNumber(~value);
```
andralex: I'm not sure this (and other similar operations) is correct. It converts a negative number into a positive one. Shouldn't it be `return SignExtendedNumber(~value, negative);`? cc @WalterBright @tgehr

src/ddmd/intrange.d
```diff
@@ -77,35 +78,70 @@ struct SignExtendedNumber
...
             return SignExtendedNumber(-value, !negative);
     }
 
-    SignExtendedNumber opAdd(const SignExtendedNumber a) const
+    R opBinary(string op : "&")(R rhs) const
     {
-        uinteger_t sum = value + a.value;
-        bool carry = sum < value && sum < a.value;
-        if (negative != a.negative)
+        return SignExtendedNumber(value & rhs.value);
```
andralex: same question here about signedness of result - looks like any two inputs produce a positive number

src/ddmd/intrange.d
```diff
@@ -77,35 +78,70 @@ struct SignExtendedNumber
...
+
+    R opBinary(string op : "^")(R rhs)
+    {
+        return SignExtendedNumber(value ^ rhs.value);
+    }
+
+    R opBinary(string op : "+")(R rhs)
+    {
+        uinteger_t sum = value + rhs.value;
+        bool carry = sum < value && sum < rhs.value;
```
andralex: Clearer and more efficient:

```d
bool carry = value > uinteger_t.max - rhs.value;
```

andralex (approved):
no blockers

src/ddmd/intrange.d
tgehr: ``~`` flips the sign bit, so if might need to be ``SignExtendedNumber(~value, !negative)`` if the underlying datatype is signed.

andralex: @tgehr I actually tried that, then druntime failed to compile

PR #7320, opened 2017-11-15
refactor: add EH_SEH

WalterBright:
This enables removing an ugly tangle of conditional compilation.


wilzbach (approved):
Looks a lot better!

PR #7322, opened 2017-11-15
do not account for exceptions in noexceptions code

WalterBright:
Don't worry about halt paths, either, as no need to unwind them.


WalterBright:
Blocked by https://github.com/dlang/dmd/pull/7321

andralex (approved):

PR #7326, opened 2017-11-16
refactor: make dcast.isVoidArrayLiteral private + move rawTypeMerge to dtemplate.d

RazvanN7:
...and make it private

### DIFF
--- a/src/ddmd/dcast.d
+++ b/src/ddmd/dcast.d
@@ -2649,7 +2649,7 @@ extern (C++) Expression scaleFactor(BinExp be, Scope* sc)
  * I.e., make sure that [1,2] is compatible with [],
  * [[1,2]] is compatible with [[]], etc.
  */
-extern (C++) bool isVoidArrayLiteral(Expression e, Type other)
+private bool isVoidArrayLiteral(Expression e, Type other)
 {
     while (e.op == TOKarrayliteral && e.type.ty == Tarray && ((cast(ArrayLiteralExp)e).elements.dim == 1))
     {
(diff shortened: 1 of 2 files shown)


src/ddmd/dcast.d
```diff
@@ -2666,28 +2666,6 @@ extern (C++) bool isVoidArrayLiteral(Expression e, Type other)
     return (e.op == TOKarrayliteral && t.ty == Tarray && t.nextOf().ty == Tvoid && (cast(ArrayLiteralExp)e).elements.dim == 0);
 }
 
-// used by deduceType()
```
JinShil: Seems like a useful comment.  I don't recommend deleting it.

JinShil (changes requested):

src/ddmd/dcast.d
RazvanN7: it is right above the deduceType helper which is right above deduceType. A rgrep after rawTypeMerge will show that deduceType is the only function using it. The only way to get to that function is by reading the deduceType function. It made sense to have that comment when it was in a different file, but now it seems kind of useless.

JinShil: Ok, I get it.  Thanks.

JinShil (approved):

PetarKirov (approved):

src/ddmd/dtemplate.d
```diff
@@ -2918,6 +2919,27 @@ private ubyte deduceWildHelper(Type t, Type* at, Type tparam)
     }
 }
 
+private Type rawTypeMerge(Type t1, Type t2)
```
mathias-lang-sociomantic: Documentation would be more than welcome

RazvanN7: Done

PR #7331, opened 2017-11-17
refactor: irstate.d

WalterBright:
1. remove dead constructor
2. `.symbol` is always a `FuncDeclaration`, so make it that type and remove unsafe cast
3. C++ never accesses this code, so no need for C++ linkage
4. make `.symbol` private so nobody else thinks it is not a `FuncDeclaration`

### DIFF
--- a/src/ddmd/irstate.d
+++ b/src/ddmd/irstate.d
@@ -31,7 +31,7 @@ import ddmd.backend.el;
  * Our label symbol, with vector to keep track of forward references.
  */
 
-extern (C++) struct Label
+struct Label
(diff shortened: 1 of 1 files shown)


stefan-koch-sociomantic:
cc @ibuclaw

WalterBright:
Since the `.h` files do not cover `IRState`, they could not be used by the C++ code.

PetarKirov (approved):
Looks good.

JinShil (approved):

andralex (approved):

PR #6440, opened 2017-01-13
Fix issue 11006 - Make sure the same type is used in pointer ops

LemonBoy:
> If both operands are pointers, and the operator is -, the pointers are subtracted and the result is divided by the size of the type pointed to by the operands. **It is an error if the pointers point to different types**. The type of the result is ptrdiff_t.

PR #7332, opened 2017-11-18
Fix Issue 11006 - Subtraction of pointers for void and non-void type compiles

JinShil:
Revival of #6440

### DIFF
--- a/src/ddmd/expressionsem.d
+++ b/src/ddmd/expressionsem.d
@@ -7628,6 +7628,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                     // Types don't match unless the have the same size
                     if (p1.size() != p2.size())
                     {
+                        // Deprecation to remain for at least a year
+                        // See https://github.com/dlang/dmd/pull/7332
                         deprecation(exp.loc,
                             "Operands point to types of different size; `%s` (%llu bytes), `%s` (%llu bytes).",
                             p1.toChars(), p1.size(),
--- a/src/ddmd/expressionsem.d
+++ b/src/ddmd/expressionsem.d
@@ -7625,16 +7625,11 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
 
                 if (!p1.equivalent(p2))
                 {
-                    // Types don't match unless the have the same size
-                    if (p1.size() != p2.size())
-                    {
-                        // Deprecation to remain for at least a year
-                        // See https://github.com/dlang/dmd/pull/7332
-                        deprecation(exp.loc,
-                            "Operands point to types of different size; `%s` (%llu bytes), `%s` (%llu bytes).",
-                            p1.toChars(), p1.size(),
-                            p2.toChars(), p2.size());
-                    }
+                    // Deprecation to remain for at least a year
+                    // See https://github.com/dlang/dmd/pull/7332
+                    deprecation(exp.loc,
+                        "cannot subtract pointers to different types: `%s` and `%s`.",
+                        t1.toChars(), t2.toChars());
                 }
 
                 // Need to divide the result by the stride
--- a/src/ddmd/expressionsem.d
+++ b/src/ddmd/expressionsem.d
@@ -7619,7 +7619,10 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         {
             if (t2.ty == Tpointer)
             {
-                // https://issues.dlang.org/show_bug.cgi?id=11006
+                // https://dlang.org/spec/expression.html#add_expressions
+                // "If both operands are pointers, and the operator is -, the pointers are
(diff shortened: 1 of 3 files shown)


JinShil:
This PR may have revealed the following 2 problems:

### Problem 1
On 32-bit platforms this code...
```D
deprecation(exp.loc,
    "Operands point to types of different size; `%s` (%zu bytes), `%s` (%zu bytes).",
    p1.toChars(), p1.size(),
    p2.toChars(), p2.size());
```
... results in error messages like `Deprecation: Operands point to types of different size; void (1 bytes), (null) (158464776 bytes).`

That is `p2.toChars()` returns `null`, and `p2.size()` returns nonsense.  This occurs even when the 32-bit host is generating a 64-bit binary.  I suspect this is a bug in DMD.

64-bit platforms do not exhibit this error even when compiling with `-m32`.

### Problem 2
Deprecation messages occur in `std.experimental.allocator`.  For example: `std/experimental/allocator/building_blocks/bitmapped_block.d(232): Deprecation: Operands point to types of different size; void (1 bytes), ulong (8 bytes).`

Perhaps `std.experimental.allocator`'s code should be audited and where there is no problem, intent expressed with explicit casts.

See output of auto-tester for more details.

JinShil:
@andralex @wilzbach @ZombineDev As code owners of `std.experimental.allocator`, your thoughts on the error described above?

andralex:
@JinShil thanks for pointing that out. The culprit seems to be the snippet:

```d
        auto start = _control.rep.ptr, end = _payload.ptr + _payload.length;
        parent.deallocate(start[0 .. end - start]);
```

The `end - start` operation subtracts pointers to different types. The solution is:

```d
        void* start = _control.rep.ptr, end = _payload.ptr + _payload.length;
        parent.deallocate(start[0 .. end - start]);
```

To my own surprise I managed to create https://github.com/dlang/phobos/pull/5864 just by using the web interface - nice job github.

andralex:
As an aside, the code was actually incorrect. It was equivalent to:

```d
        ulong* start = _control.rep.ptr;
        void* end = _payload.ptr + _payload.length;
        parent.deallocate(start[0 .. end - start]);
```

The computation `end - start` yields the number of bytes, but the slice passed to `deallocate` will be 8 times larger. Probably this only worked because the parent allocator tolerated the bug (e.g. Mallocator and GCAllocator don't use the length of the chunk).

JinShil:
From my perspective, this is ready to go.

src/ddmd/expressionsem.d
```diff
@@ -7619,6 +7619,24 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
             if (t2.ty == Tpointer)
             {
+                // https://issues.dlang.org/show_bug.cgi?id=11006
+                Type p1 = t1.nextOf();
+                Type p2 = t2.nextOf();
+
+                if (!p1.equivalent(p2))
+                {
+                    // Types don't match unless the have the same size
+                    if (p1.size() != p2.size())
```
ZombineDev: According to spec, subtracting pointers pointing to objects of different types (regardless of size) is an error. The result obviously would be the same if you cast one of the pointers to the same type as the other one, but may be actually be a mistake. What are your opinions on being more strict, as the spec dictates?

PetarKirov (approved):
LGTM, modulo one question about how strict we want to be.

src/ddmd/expressionsem.d
JinShil: Yeah, I'm for following the spec.  I'll update the PR.

JinShil: My understanding is the `Type.equivalent()` function is all that's needed.  I think the size comparison in the previous commit was unnecessary. 

Again, I think this is ready to go unless I'm missing something.

JinShil:
Is jenkins just finicky or what?

WalterBright:
@andralex it is nice that this PR found a bug.

src/ddmd/expressionsem.d
```diff
@@ -7619,6 +7619,19 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         {
             if (t2.ty == Tpointer)
             {
+                // https://issues.dlang.org/show_bug.cgi?id=11006
+                Type p1 = t1.nextOf();
```
WalterBright: I don't much point here to linking to the bug report. Instead link to the relevant spec
https://dlang.org/spec/expression.html#add_expressions and add the quote: "If both operands are pointers, and the operator is -, the pointers are subtracted and the result is divided by the size of the type pointed to by the operands. It is an error if the pointers point to different types."

Linking to and quoting from the spec should be a good way to clarify what's going on in the implementation.

JinShil: Done.  Thanks!

PR #7333, opened 2017-11-18
emit better code for try-finally when function does not throw

WalterBright:
Even if a function is marked `nothrow`, the exception unwind tables must still be there because a try-catch could still be there that absorbs the exception. This PR marks a function internally if it does not throw and does not have a try-catch. This knowledge is then used to omit building exception unwind tables, and the code gen does not have to assume that an external unwinder is calling the `finally` blocks. The code emitted (at least for Windows) is substantially less.

### DIFF
--- a/src/ddmd/declaration.h
+++ b/src/ddmd/declaration.h
@@ -481,6 +481,7 @@ void builtin_init();
 #define FUNCFLAGreturnInprocess 0x10    // working on inferring 'return' for parameters
 #define FUNCFLAGinlineScanned   0x20    // function has been scanned for inline possibilities
 #define FUNCFLAGinferScope      0x40    // infer 'scope' for parameters
+#define FUNCFLAGhasCatches      0x80    // function has try-catch statements
 
 class FuncDeclaration : public Declaration
 {
@@ -523,6 +524,8 @@ public:
     CompiledCtfeFunction *ctfeCode;     // Compiled code for interpreter
     int inlineNest;                     // !=0 if nested inline
     bool isArrayOp;                     // true if array operation
+    bool eh_none;                       /// true if no exception unwinding is needed
+
     // true if errors in semantic3 this function's frame ptr
     bool semantic3Errors;
     ForeachStatement *fes;              // if foreach body, this is the foreach
--- a/src/ddmd/backend/cc.d
+++ b/src/ddmd/backend/cc.d
@@ -408,8 +408,8 @@ enum
     BFLnostackopt    = 8,       // set when stack elimination should not
                                 // be done
     // NTEXCEPTIONS
-    BFLehcode        = 0x10,    // set when we need to load exception code
-    BFLunwind        = 0x1000,  // do local_unwind following block
+    BFLehcode        = 0x10,    // BC_filter: need to load exception code
+    BFLunwind        = 0x1000,  // do local_unwind following block (unused)
 
     BFLnomerg        = 0x20,    // do not merge with other blocks
     BFLprolog        = 0x80,    // generate function prolog
@@ -668,6 +668,7 @@ enum
     Fnotailrecursion = 0x4000,  // no tail recursion optimizations
     Ffakeeh          = 0x8000,  // allocate space for NT EH context sym anyway
     Fnothrow         = 0x10000, // function does not throw (even if not marked 'nothrow')
+    Feh_none         = 0x20000, // ehmethod==EH_NONE for this function only
 }
 
 struct func_t
--- a/src/ddmd/backend/cc.h
+++ b/src/ddmd/backend/cc.h
@@ -418,8 +418,8 @@ enum
     BFLnostackopt    = 8,       // set when stack elimination should not
                                 // be done
 #if NTEXCEPTIONS
-    BFLehcode        = 0x10,    // set when we need to load exception code
-    BFLunwind        = 0x1000,  // do local_unwind following block
+    BFLehcode        = 0x10,    // BC_filter: need to load exception code
+    BFLunwind        = 0x1000,  // do local_unwind following block (unused)
 #endif
     BFLnomerg        = 0x20,    // do not merge with other blocks
     BFLprolog        = 0x80,    // generate function prolog
@@ -709,6 +709,7 @@ enum
     Fnotailrecursion = 0x4000,  // no tail recursion optimizations
     Ffakeeh          = 0x8000,  // allocate space for NT EH context sym anyway
     Fnothrow         = 0x10000, // function does not throw (even if not marked 'nothrow')
+    Feh_none         = 0x20000, // ehmethod==EH_NONE for this function only
 };
 
 struct func_t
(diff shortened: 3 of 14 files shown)


andralex:
awesome possum!

andralex (approved):

andralex:
@WalterBright check the autotester

WalterBright:
Yeah, I know. It's not ready yet. Turned off automerge.

WalterBright:
This has an odd failure on Win32. I get two failures:
```
Makefile:163: recipe for target 'test_results/compilable/test16031.d.out' failed
gmake[1]: *** [test_results/compilable/test16031.d.out] Error 127
Makefile:201: recipe for target 'start_compilable_tests' failed
gmake: *** [start_compilable_tests] Error 2
```
and:
```
Makefile:170: recipe for target 'test_results/fail_compilation/cppeh2.d.out' failed
gmake[1]: *** [test_results/fail_compilation/cppeh2.d.out] Error 127
Makefile:207: recipe for target 'start_fail_compilation_tests' failed
gmake: *** [start_fail_compilation_tests] Error 2
gmake: *** Waiting for unfinished jobs....
```
In test16031.d it says:
```
// REQUIRED_ARGS: -fPIC -lib
// PERMUTE_ARGS:
// DISABLED: win32 win64
```
and in cppeh2.d it says:
```
// DISABLED: win32 win64
```
How can the tests fail on Win32 when they are disabled on Win32?

rainers:
> How can the tests fail on Win32 when they are disabled on Win32?

The tests are still run even if disabled, but the result is ignored (and shows `DISABLED but PASSES` if successful). As there is no output for running those two tests (supposed to show up with [DISABLED]), I suspect d_do_test is crashing itself. The unusual return code 127 also hints at an abnormal termination.

WalterBright:
Thanks for the tip. I retrieved and tried running d_do_test:
```
C:\cbx\mars\test>d_do_test fail_compilation cppeh2 d

std.file.FileException@std\file.d(430): fail_compilationcppeh2.d: The system cannot find the file specified.
----------------
0x0043DBA8
0x0040A9BC
0x00402CAA
0x00404A7A
0x00438B1F
0x00438AE3
0x004389A8
0x0042D2CB
0x750A336A in BaseThreadInitThunk
0x776098F2 in RtlInitializeExceptionChain
0x776098C5 in RtlInitializeExceptionChain
```
Looks like that program needs work. (The file definitely exists.)

WalterBright:
I wound up disabling the optimization for Win32 and filing a bug report for it:
https://issues.dlang.org/show_bug.cgi?id=17997

src/ddmd/declaration.h
```diff
@@ -523,6 +524,8 @@ class FuncDeclaration : public Declaration
     CompiledCtfeFunction *ctfeCode;     // Compiled code for interpreter
     int inlineNest;                     // !=0 if nested inline
     bool isArrayOp;                     // true if array operation
+    bool eh_none;                       /// true if no exception unwinding is needed
```
MartinNowak: Mmh, that doesn't belong in the frontend, any other place you could store or retrieve this flag?

WalterBright:
@CyberShadow I can't tell if the DAutoTest failure is the fault of this PR or not?

MartinNowak (commented):
Otherwise looks good.
Would be great if we could generally improve the non-exception code-path, currently dmd uses a call to jump to the finally handler, which trashes the return prediction.

ZombineDev:
@WalterBright 

From the [build log](http://dtest.dlang.io/results/a2f903ee05970eed9a43a60db10e6f82001444c0/c02fa7b6a43be2ed6cdbe0624cd1fd4afb7f6717/), it looks like an intermittent network failure:
```
.generated/stable_dmd-2.072.2/dmd2/linux/bin64/dub upgrade --missing-only --root=dpl-docs
Upgrading project in /dev/shm/dtest/work/repo/dlang.org/dpl-docs
Selected package ddox 0.15.18 doesn't exist. Using latest matching version instead.
Selected package libev 5.0.0+4.04 doesn't exist. Using latest matching version instead.
Selected package libevent 2.0.2+2.0.16 doesn't exist. Using latest matching version instead.
Selected package experimental_allocator 2.70.0-b1 doesn't exist. Using latest matching version instead.
Selected package libdparse 0.7.0 doesn't exist. Using latest matching version instead.
Selected package memutils 0.4.9 doesn't exist. Using latest matching version instead.
Selected package vibe-d 0.7.31 doesn't exist. Using latest matching version instead.
Selected package hyphenate 1.1.1 doesn't exist. Using latest matching version instead.
Selected package libasync 0.8.3 doesn't exist. Using latest matching version instead.
Root package dpl-docs references unknown package ddox
posix.mak:583: recipe for target 'dpl-docs' failed
make: *** [dpl-docs] Error 2
```

As clearly ddox is a know package: http://code.dlang.org/packages/ddox. The error message is trying to say that dub tried to contact the registry in order to find `ddox`, but it failed and fell back to searching the package on the local computer and after it failed it assumed that such package is not known to exist.

The error message is quite obscure, but has since been improved (along with the addition of registry mirrors to remedy the network failures) in later versions of Dub, but dlang.org's build uses DMD 2.072.2, and hence why the the dub version used is probably quite old.

To fix this, you need to restart the build and at the moment the only way is to rebase the branch of your pull request and force push it.

src/ddmd/declaration.h
WalterBright: The flag is per-function, and is detected by the font end. It is later transferred to the backend's symbol for that same function.

WalterBright:
> Would be great if we could generally improve the non-exception code-path, currently dmd uses a call to jump to the finally handler, which trashes the return prediction.

I know, but this is just one more incremental step to get there. I spent 3 days trying to get just this to work, and it still fails on Win32 for mysterious reasons. Debugging eh code is stupidly hard. Having `d_do_test` itself fail makes for even worse problems. Most everything has to work in order for that program to work, which is backwards.

I lost at least one day because asserts are disabled in the front end. I need to fix that, too. The asserts had found the problem, but being disabled, things just went merrily onwards and died randomly somewhere else.

WalterBright:
Ok, changed a comment to get the DAutoTest to run again.

WalterBright:
`d_do_test.d` should be written in Python or something, so the test runner itself does not depend on the compiler ecosystem working. Or maybe just have it built with the last release of the compiler, not the current one under test.

MartinNowak:
> I lost at least one day because asserts are disabled in the front end. I need to fix that, too. The asserts had found the problem, but being disabled, things just went merrily onwards and died randomly somewhere else.

Let's give it a try then https://github.com/dlang/dmd/pull/7344.

MartinNowak:
> Or maybe just have it built with the last release of the compiler, not the current one under test.

That's an idiom that we use everywhere else.

WalterBright:
BTW, getting all these EH optimizations in should substantially reduce the cost of "zero cost" RAII exceptions (haha). I had forgotten how much better code my pre-exceptions C++ compiler generated for RAII.

MartinNowak:
> Or maybe just have it built with the last release of the compiler, not the current one under test.

See https://github.com/dlang/dmd/pull/7345

PR #7334, opened 2017-11-18
Fix C++ interface to DDMD

ibuclaw:
Still there are far too many things being forgotten when changes are made to the D API.

### DIFF
--- a/src/dmd/aggregate.h
+++ b/src/dmd/aggregate.h
@@ -187,7 +187,6 @@ public:
 
     static StructDeclaration *create(Loc loc, Identifier *id);
     Dsymbol *syntaxCopy(Dsymbol *s);
-    void semantic(Scope *sc);
     void semanticTypeInfoMembers();
     Dsymbol *search(Loc, Identifier *ident, int flags = SearchLocalsOnly);
     const char *kind() const;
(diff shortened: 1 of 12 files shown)


ibuclaw:
And found a couple more in nspace.h

src/ddmd/attrib.d
```diff
@@ -626,6 +648,35 @@ extern (C++) final class AlignDeclaration : AttribDeclaration
         return createNewScope(sc, sc.stc, sc.linkage, sc.cppmangle, sc.protection, sc.explicitProtection, this, sc.inlining);
     }
 
+    structalign_t getAlignment(Scope* sc)
```
WalterBright: I believe these were deliberately turned into free functions in order to make it easier to use the compiler as a library. Please don't revert this.

andralex: ping @ibuclaw cc @RazvanN7

RazvanN7: Indeed, Walter is correct, please do not revert this.

wilzbach:
Ping @ibuclaw 
Do you need this to be part of 2.078?

ibuclaw:
I gave up trying to reseat gdc on 2.077 for the moment, there was just too much broken in the internals, and it seems like 2.078 has continued the trend with the recent rewrite of visitor.h that I had to do.

ibuclaw:
Changed base from stable to master.  Added test for at least calling `semantic(Dsymbol*)` from C++.

ibuclaw:
@wilzbach - This can be merged.

PR #7340, opened 2017-11-20
Replace `body` keyword with its replacement: `do`

wilzbach:
Automatic replacement with

    sed -i "s/^\([ ]*\)body/\1do/" -i **/*.d

See also: https://github.com/dlang/phobos/pull/5869, https://github.com/dlang/druntime/pull/1975
I left the `test` files untouched for now as they should probably be changed once the use of `body` triggers a deprecation warning.

### DIFF
--- a/src/ddmd/astbase.d
+++ b/src/ddmd/astbase.d
@@ -5924,7 +5924,7 @@ struct ASTBase
         {
             assert(!!aggrfe ^ !!rangefe);
         }
-        body
+        do
         {
             this.loc = loc;
             this.aggrfe = aggrfe;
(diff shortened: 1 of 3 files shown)


ZombineDev:
@wilzbach we can't go forward with this, because the `do` keyword as alternative to `body` was introduced in dmd 2.075.0 (https://github.com/dlang/dmd/commit/175638fb3823b26224ac7698e940972815281347), however we must support building dmd with older frontends (2.066.0 ~ 2.068 IIRC).

wilzbach:
Argh I thought we already upgraded them for https://github.com/dlang/dmd/pull/6907, but this PR is still WIP.
Well if we ever upgrade the host compiler, once could reopen this, but as it's a trivial PR I think the best course of action is to close. Sorry for the spam.

PR #5215, opened 2015-10-21
Issue 15094 - __traits(getMember) fails when the source is a struct/class field

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=15094

If `TemplateInstance.semanticTiargs` is called from `TraitsExp`, do not strip off the contexts of argument expressions:
1. give `true` to the `intypeid` parameter of `Type.resolve` call.
2. leave `TOKdotvar` and `TOKdottd` as-is.

PR #7341, opened 2017-11-20
Fix Issue 15094 - __traits(getMember) fails when the source is a struct/class field

JinShil:
Revival of #5215
Special thanks to @9rnsr

### DIFF
--- a/src/ddmd/traits.d
+++ b/src/ddmd/traits.d
@@ -40,6 +40,7 @@ import ddmd.semantic;
 import ddmd.tokens;
 import ddmd.typesem;
 import ddmd.visitor;
+import ddmd.root.rootobject;
 
 enum LOGSEMANTIC = false;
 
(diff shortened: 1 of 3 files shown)


src/ddmd/traits.d
```diff
@@ -58,6 +59,31 @@ struct PushAttributes
...
+ * additionally strip off expression contexts.
+ *
+ * Some symbol related `__traits` ignore arguments expression contexts.
+ * For example:
+ *  struct S { void f() {} }
+ *  S s;
+ *  pragma(msg, __traits(isNested, s.f));
+ *  // s.f is DotVarExp, but __traits(isNested) needs a FuncDeclaration.
+ *
+ * This is used for that common `__traits` behavior.
```
WalterBright: Function comment is not following Ddoc style.

JinShil: Done.  Thanks!

andralex (approved):
awesome possum

andralex:
I'll pull this, looks soild. cc @WalterBright

PR #7342, opened 2017-11-21
Fix Issue 9433 - Deprecate delete

JinShil:
See:
https://dlang.org/deprecate.html
https://dlang.org/deprecate.html#delete
http://forum.dlang.org/post/jzldqlkxgewtxjyffekb@forum.dlang.org
http://forum.dlang.org/post/gqgnquwhmykvpafwpemr@forum.dlang.org

Spec and Documentation Update: https://github.com/dlang/dlang.org/pull/1941.

There are only 5 lines of code change in this PR, and 4 of them are comments. The rest are mostly line number changes for the test suite, as many of the tests now emit a deprecation warning in addition to their other output.

### DIFF
--- a/changelog/deprecate_delete.dd
+++ b/changelog/deprecate_delete.dd
@@ -1,0 +1,8 @@
+The `delete` keyword has been deprecated.
+
+See the $(LINK2 $(ROOT_DIR)deprecate.html#delete, Deprecated Features) for more information.
+
+Starting with this release, using the `delete` keyword will result in a deprecation warning.
+
+As a replacement, users are encouraged to use $(REF1 destroy, object)
+if feasible, or $(REF __delete, core,memory) as a last resort.
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -7768,6 +7768,11 @@ final class Parser(AST) : Lexer
         case TOK.delete_:
             nextToken();
             e = parseUnaryExp();
+            // @@@DEPRECATED_2019-02@@@
+            // 1. Deprecation for 1 year
+            // 2. Error for 1 year
+            // 3. Removal, "delete" can be used for other identities
+            deprecation("The `delete` keyword has been deprecated.  Use `object.destroy()` instead.");
             e = new AST.DeleteExp(loc, e, false);
             break;
 
--- a/test/compilable/vgc1.d
+++ b/test/compilable/vgc1.d
@@ -1,6 +1,15 @@
 // REQUIRED_ARGS: -vgc -o-
 // PERMUTE_ARGS:
 
+/*
+TEST_OUTPUT:
+---
+compilable/vgc1.d(93): Deprecation: The `delete` keyword has been deprecated.  Use `object.destroy()` instead.
+compilable/vgc1.d(94): Deprecation: The `delete` keyword has been deprecated.  Use `object.destroy()` instead.
+compilable/vgc1.d(95): Deprecation: The `delete` keyword has been deprecated.  Use `object.destroy()` instead.
+---
+*/
+
 /***************** NewExp *******************/
 
 struct S1 { }
@@ -12,13 +21,13 @@ struct S5 { @nogc new(size_t); }
 /*
 TEST_OUTPUT:
 ---
-compilable/vgc1.d(27): vgc: `new` causes a GC allocation
-compilable/vgc1.d(29): vgc: `new` causes a GC allocation
-compilable/vgc1.d(30): vgc: `new` causes a GC allocation
-compilable/vgc1.d(32): vgc: `new` causes a GC allocation
-compilable/vgc1.d(33): vgc: `new` causes a GC allocation
-compilable/vgc1.d(34): vgc: `new` causes a GC allocation
+compilable/vgc1.d(36): vgc: `new` causes a GC allocation
 compilable/vgc1.d(38): vgc: `new` causes a GC allocation
+compilable/vgc1.d(39): vgc: `new` causes a GC allocation
+compilable/vgc1.d(41): vgc: `new` causes a GC allocation
+compilable/vgc1.d(42): vgc: `new` causes a GC allocation
+compilable/vgc1.d(43): vgc: `new` causes a GC allocation
+compilable/vgc1.d(47): vgc: `new` causes a GC allocation
 ---
 */
 
@@ -41,12 +50,12 @@ void testNew()
 /*
 TEST_OUTPUT:
 ---
-compilable/vgc1.d(55): vgc: `new` causes a GC allocation
-compilable/vgc1.d(57): vgc: `new` causes a GC allocation
-compilable/vgc1.d(58): vgc: `new` causes a GC allocation
-compilable/vgc1.d(60): vgc: `new` causes a GC allocation
-compilable/vgc1.d(61): vgc: `new` causes a GC allocation
-compilable/vgc1.d(62): vgc: `new` causes a GC allocation
+compilable/vgc1.d(64): vgc: `new` causes a GC allocation
+compilable/vgc1.d(66): vgc: `new` causes a GC allocation
+compilable/vgc1.d(67): vgc: `new` causes a GC allocation
+compilable/vgc1.d(69): vgc: `new` causes a GC allocation
+compilable/vgc1.d(70): vgc: `new` causes a GC allocation
+compilable/vgc1.d(71): vgc: `new` causes a GC allocation
 ---
 */
 
@@ -74,9 +83,9 @@ void testNewScope()
 /*
 TEST_OUTPUT:
 ---
-compilable/vgc1.d(84): vgc: `delete` requires the GC
-compilable/vgc1.d(85): vgc: `delete` requires the GC
-compilable/vgc1.d(86): vgc: `delete` requires the GC
+compilable/vgc1.d(93): vgc: `delete` requires the GC
+compilable/vgc1.d(94): vgc: `delete` requires the GC
+compilable/vgc1.d(95): vgc: `delete` requires the GC
 ---
 */
 void testDelete(int* p, Object o, S1* s)
--- a/test/fail_compilation/fail14486.d
+++ b/test/fail_compilation/fail14486.d
@@ -1,5 +1,41 @@
 // REQUIRED_ARGS: -o-
 
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail14486.d(81): Deprecation: The `delete` keyword has been deprecated.  Use `object.destroy()` instead.
+fail_compilation/fail14486.d(82): Deprecation: The `delete` keyword has been deprecated.  Use `object.destroy()` instead.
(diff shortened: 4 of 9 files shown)


rracariu:
delete is still required for cases like https://issues.dlang.org/show_bug.cgi?id=17230

JinShil:
There's only 4 lines of code change in this PR, and 3 of them are comments.  The rest are mostly line number changes for the test suite, as many of the tests now emit a deprecation warning in addition to their other output.

JinShil:
> delete is still required for cases like https://issues.dlang.org/show_bug.cgi?id=17230

That is due to a separate bug:  https://issues.dlang.org/show_bug.cgi?id=17267.  Once that bug is fixed, you should be able to use `destroy`.

src/ddmd/parse.d
```diff
@@ -7770,6 +7770,10 @@ final class Parser(AST) : Lexer
         case TOKdelete:
             nextToken();
             e = parseUnaryExp();
+            // 1. Deprecation for 1 year
+            // 2. Error for 1 year
+            // 3. Removal, "delete" can be used for other identities
```
wilzbach: How about using a similar deprecated pattern as in Phobos?
```
// @@@DEPRECATED_2017-12@@@
```

In any case, it would be good to start using easily greppable years into the source code, s.t. these deprecations don't get forgotten.

wilzbach (approved):
While it was planned for 2.072, there's a real chance we might see this deprecation for 2.078 :tada:

wilzbach:
Btw could you also send a respective PR to dlang.org, s.t. the update of the deprecation page doesn't get lost? (Unfortunately this tends to happen too often).

JinShil:
> Btw could you also send a respective PR to dlang.org, s.t. the update of the deprecation page doesn't get lost? 

I would be happy to, but I need to know what version I can expect this PR to land in.

src/ddmd/parse.d
JinShil: > // @@@DEPRECATED_2017-12@@@

 How can I be sure that this will be merged this month?

wilzbach: The trick is to annoy @andralex and @WalterBright daily ;-)
Okay all jokes aside, at Phobos these deprecation strings are put in a PR and this has never been a huge problem as at least for Phobos deprecation PRs are merged rather quickly. Yep, there have been cases in which a PR has been merged later, but it hasn't been a big deal whether a deprecation lasted for 23 or 24 months. On the contrary (1) with Martin's polished release two months release cycle, it's still 9 or 10 releases (compared to 3-4 releases two or three years ago) and people are now complaining that the deprecation periods are too long. And secondly (2) the obvious advantage is that someone will actually switch the deprecation to an error.
Anyhow, the deprecation cycle is rather ill-defined as of now.

JinShil: OK, I added `@@@DEPRECATED_2017-12@@@`.

JinShil:
@WalterBright and @andralex Can I please get a ruling on this?  I think time is running out for the 2.078.0 release.

wilzbach:
> @WalterBright and @andralex Can I please get a ruling on this? I think time is running out for the 2.078.0 release.

I had a short chat with @andralex about this last week and the only thing he was worried about was that no drop-in replacement for `delete` exists.

However, after looking at it for a bit, I don't see any reason why people if `destroy` and `GC.free` doesn't work for them, couldn't use [`_d_delclass`](https://github.com/dlang/druntime/blob/master/src/rt/lifetime.d#L145) directly:

```
extern (C) void _d_delclass(Object* p);

void main() {
    class B { int a; }
    B b = new B();
    b.a = 10;

    import std.stdio;
    b.writeln; // test.main.B
    _d_delclass(cast(Object*) &b);
    b.writeln; // null
}
```

JinShil:
>  I had a short chat with @andralex about this last week and the only thing he was worried about was that no drop-in replacement for `delete` exists.

Why does a drop-in replacement need to exist if it's deprecated?

> However, after looking at it for a bit, I don't see any reason why people if destroy and GC.free doesn't work for them, couldn't use _d_delclass directly:

Clever!   :thumbsup:

andralex:
>Why does a drop-in replacement need to exist if it's deprecated?

The idea here is purely operational - someone who has a codebase using `delete` on their hands needs a quick way to upgrade the compiler whilst keeping the code unchanged. So we need to have a simple recipe "for legacy code using `delete e;` please replace with `xxx(e);`.

src/ddmd/parse.d
```diff
@@ -7770,6 +7770,10 @@ final class Parser(AST) : Lexer
...
+            deprecation("The `delete` keyword has been deprecated.  Use `object.destroy()` instead.");
```
andralex: `destroy` is not a pound-for-pound replacement for `delete`.

andralex (dismissed):

changelog/deprecate_delete.dd
```diff
@@ -0,0 +1,5 @@
+The `delete` keyword has been deprecated.
+
+See the $(LINK2 $(ROOT_DIR)deprecate.html#delete, Deprecated Features) for more information.
+
+Starting with this release, using the `delete` keyword will result in a deprecation warning.  The warning is scheduled to remain for 1 year, after which it will be changed to an error.  The error is then to remain for 1 year, after which `delete` will be removed as a keyword.
```
andralex: The changelog should specify the replacement function.

andralex:
@wilzbach _d_delclass would be suitable. It's not an easy replacement, but probably we don't want to make it too easy.

JinShil:
> So we need to have a simple recipe "for legacy code using `delete e`; please replace with `xxx(e);`.

How long would such a recipe need to be maintained?  I understand the value of having such a recipe to make for smoother transition, but don't we eventually want to usher users away from such things and encourage them to refactor their code so they can make use of `destroy`?

andralex:
>How long would such a recipe need to be maintained? I understand the value of having such a recipe to make for smoother transition, but don't we eventually want to usher users away from such things and encourage them to refactor their code so they can make use of destroy?

There's no significant burden to having a function in the standard library that does what `delete` does. It will be marked as unsafe and documented as such.

andralex:
@JinShil just spoke to @wilzbach, he'll write the function. Thanks to both!

JinShil:
> just spoke to @wilzbach, he'll write the function.

Ok. Thanks.  What's our recommendation to users, then?  "Prefer `object.destroy()`, but if that's not suitable, resort to the <insert function name> library function"?

andralex:
@JinShil yes, something along these lines.

JinShil:
@wilzbach I'll make changes as soon as I know what function to direct users to.

JinShil:
> just spoke to @wilzbach, he'll write the function.

@wilzbach, is this statement true?

wilzbach:
> @wilzbach, is this statement true?

Yes. I'm sorry that you missed 2.078 :/
I will get to it ASAP (i.e. this year).

wilzbach:
> Yes. I'm sorry that you missed 2.078 :/
> I will get to it ASAP.

Finally managed to get to it:

https://github.com/dlang/druntime/pull/2037

Sorry for the long delay

JinShil:
This is ready for additional scrutiny and potential merge, now that https://github.com/dlang/druntime/pull/2037 is in.

changelog/deprecate_delete.dd
```diff
@@ -0,0 +1,10 @@
+The `delete` keyword has been deprecated.
+
+See the $(LINK2 $(ROOT_DIR)deprecate.html#delete, Deprecated Features) for more information.
+
+Starting with this release, using the `delete` keyword will result in a deprecation warning.  The
+warning is scheduled to remain for 1 year, after which it will be changed to an error.  The
+error is then to remain for 1 year, after which `delete` will be removed as a keyword.
+
+As a replacement, users are encouraged to use $(LINK2 $(ROOT_DIR)phobos/object.html#.destroy),
```
jacob-carlborg: Should it be `phobos` in the link?

JinShil:
The links to `object.destroy` and `core.memory.__delete` don't resolve.  `__delete` [doesn't even seem to be in the docs](http://dtest.dlang.io/artifact/website-f9fb5f80bc0b4127acd5d08058b245ed9a9cabb3-677d592a5d032c630664b91a89c8a0af/web/phobos/core_memory.html#.__delete)  I could use some help on that.

changelog/deprecate_delete.dd
```diff
@@ -0,0 +1,10 @@
+The `delete` keyword has been deprecated.
+
+See the $(LINK2 $(ROOT_DIR)deprecate.html#delete, Deprecated Features) for more information.
+
+Starting with this release, using the `delete` keyword will result in a deprecation warning.  The
+warning is scheduled to remain for 1 year, after which it will be changed to an error.  The
+error is then to remain for 1 year, after which `delete` will be removed as a keyword.
```
wilzbach: Are you aware of Jack's new deprecation DIP?
https://github.com/dlang/DIPs/pull/108
One of the things that are on the list is instead of having a fixed date, a fixed version is better for users because it's important which version will stop supporting it (there might be a few delay of publishing a version, e.g. LDC or distros in general)

wilzbach (commented):
FYI: there are plans to use scheduled versions instead of scheduled date. The upcoming DIP also mentions a few other things - I am not saying that you need to apply them hear, but I think it doesn't hurt to have a look at what could be the mandatory soon.

wilzbach:
Nothing to worry about - __delete hasn't been released yet and is only visible on the preview pages: http://dtest.dlang.io/artifact/website-f9fb5f80bc0b4127acd5d08058b245ed9a9cabb3-677d592a5d032c630664b91a89c8a0af/web/phobos-prerelease/core_memory.html#.__delete

changelog/deprecate_delete.dd
JinShil: Yes, I've seen the DIP, but it needs to go through a review and approval process which I think will take time.  Given the the fact that we don't have a well-defined deprecation process, I made something up, and that's what you see here.  Feel free to suggest an alternative wording, or maybe I should just omit the schedule altogether so we don't have to commit to anything.  I don't have an opinion either way.

JinShil:
> Nothing to worry about - __delete hasn't been released yet and is only visible on the preview pages:

What about `object.destroy`?  That gives me a 404.  Will that be resolved at publication time too?

wilzbach:
From https://github.com/dlang/dlang.org/blob/master/dlang.org.ddoc#L353

> The REF1 variants work around DDoc limitations and allow linking to
    modules not inside packages (i.e. object.)

Argh...

-> `$(REF1 object, destroy)` works. I will update DRuntime and Phobos too. Good catch!

changelog/deprecate_delete.dd
wilzbach: Yes, please omit the schedule.
In my experience, we most likely won't stick to it anyways.

wilzbach (approved):
Looks all good to me. Thanks for pushing this!

changelog/deprecate_delete.dd
JinShil: Done.

PR #6746, opened 2017-05-07
Revert #6026 - build compiler w/ optimizations to speed up test suite

ibuclaw:
I expect the testsuite to fail - and this is why we should not be doing this.

PR #7344, opened 2017-11-21
enable assertions in CI tests

MartinNowak:
I guess we'll still need to fix a couple of things and might have to disable class invariants to reduce the overhead.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -45,7 +45,7 @@ clone() {
 # build dmd, druntime, phobos
 build() {
     source ~/dlang/*/activate # activate host compiler
-    make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=$DMD ENABLE_RELEASE=1 ENABLE_WARNINGS=1 all
+    make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=$DMD ENABLE_WARNINGS=1 all
     make -j$N -C ../druntime -f posix.mak MODEL=$MODEL
     make -j$N -C ../phobos -f posix.mak MODEL=$MODEL
     deactivate # deactivate host compiler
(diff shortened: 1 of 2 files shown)


WalterBright:
As I recall, back when (Kenji?) tried it it substantially slowed down the compiler. I want to review the best way forward on this first. I think the non-release mode inserts too many checks.

MartinNowak:
> As I recall, back when (Kenji?) tried it it substantially slowed down the compiler. I want to review the best way forward on this first. I think the non-release mode inserts too many checks.

One way forward is `-release=assert,invariant` so that we can enable asserts but not invariants.

mathias-lang-sociomantic:
>so that we can enable asserts but not invariants

Shameless self-plug: https://github.com/dlang/DIPs/blob/master/DIPs/DIP1006.md

wilzbach:
I like the idea of finally moving forward with DIP1006! BTW one minor thing: if we start to use a different compiler build (non-release) than the one we ship on the auto-tester, how do we make sure that the version we ship doesn't get accidentally broken?

ibuclaw:
Guess you can close #6746 then.

wilzbach:
>  for the time being we might be better off with the slow-down.

Yes, we could simply do this on a faster CI (e.g. [SemaphoreCI](https://semaphoreci.com/product) is twice as fast as Travis - I checked this recently. Compare [SemaphoreCi build](https://semaphoreci.com/wilzbach/dmd/branches/master/builds/6), [Travis build](https://travis-ci.org/dlang/dmd/jobs/315475937)).

> I guess we'll still need to fix a couple of things and might have to disable class invariants to reduce the overhead.

FYI: our CircleCi build already runs without `ENABLE_RELEASE` - though
- the flag default is `BUILD=release, but [doesn't set `ENABLE_RELEASE`](https://github.com/dlang/dmd/pull/7473)
-  [I would like to use `ENABLE_DEBUG` on CircleCi too](https://github.com/dlang/dmd/pull/7528)

ibuclaw:
> Yes, we could simply do this on a faster CI (e.g. SemaphoreCI is twice as fast as Travis

I would say more. On a good day, gdc is built with testsuite and unittests ran in 14 minutes flat on semaphore.

Travis barely could do the same in 50 minutes.

MartinNowak:
What I found implementing https://github.com/dlang/dmd/pull/7980, we could hack this already today.
`-release -unittest` disables all run-time checks but assertions. If we versioned out dmd's few unittests, we'd get what we want for CI.

ibuclaw:
> What I found implementing #7980, we could hack this already today.
-release -unittest disables all run-time checks but assertions.

Heh, wouldn't that be a bug? Or maybe not. It does seem surprising though.

MoonlightSentinel:
FWIW CircleCI and Azure already run the test suite with debug builds. Is this still required?

Geod24:
Don't think so. Also we should have it in the release, but that's another topic.

PR #7346, opened 2017-11-21

MartinNowak:


BBasile:
possible cause of this regression https://issues.dlang.org/show_bug.cgi?id=18115

PR #7347, opened 2017-11-21
Fix compiler preferring `body` to `do`

ntrel:


MartinNowak:
It would be kind to any reviewers to include a reference to discussions, instead of letting people dig out anything themselves.
https://github.com/dlang/dlang.org/commit/4c95ab8a206e8679176e8c9dd7ef4a5b4b6b82f6#diff-463ede2bb5b4c2bf9c49ec4fd42e0540
https://github.com/dlang/DIPs/pull/48

MartinNowak (approved):

test/compilable/extra-files/header1i.di
```diff
@@ -14,7 +14,7 @@ out(result)
 {
 	assert(result == 0);
 }
-body
+do
```
JinShil: I'm OK with the text messages, but I'm not sure changing code (even if it's a header file).  See https://github.com/dlang/dmd/pull/7340  

I defer to wiser reviewers.

PetarKirov (approved):

PR #5200, opened 2015-10-18
Issue 12385 - Enum members typed as immutable should not be modifiable

AndrejMitrovic:
Fixes https://issues.dlang.org/show_bug.cgi?id=12385

Recreated from: https://github.com/D-Programming-Language/dmd/pull/3484

In reality this change will likely fix other edge-cases where the type of the member wasn't taken fully into account (shared/const/immutable), perhaps even inout.

Suggested fix was by @9rnsr

PR #7348, opened 2017-11-21
Fix Issue 12385 - Enum member should not be modifiable when the member is immutable

JinShil:
Revival of #5200
Thanks to @AndrejMitrovic and @9rnsr


MartinNowak (approved):

JinShil:
I had to modify the test for [Issue 16570](https://issues.dlang.org/show_bug.cgi?id=16570) to accommodate this fix.  @MartinNowak that issue was fixed by you, so could you please give this a second look?

mihails-strasuns-sociomantic:
Introduce regression - https://issues.dlang.org/show_bug.cgi?id=18645

PR #7349, opened 2017-11-21
Make ClassDeclaration.inuse bool and private

andralex:
See also https://github.com/dlang/dmd/pull/7315


src/ddmd/dclass.d
```diff
@@ -216,8 +216,8 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
     bool cpp;           // true if this is a C++ interface
     bool isobjc;        // true if this is an Objective-C class/interface
     bool isscope;       // true if this is a scope class
+    private bool inuse;          // to prevent recursive attempts
```
jacob-carlborg: Do we want to rename it to `inUse` as well to follow standard D style guides?

andralex: We're good here.

PetarKirov (approved):

PR #7353, opened 2017-11-22
refactor: explicitly list flags

WalterBright:
These ad-hoc flags are wretched enough without the confusing bit inversions.

### DIFF
--- a/src/ddmd/backend/cgcod.c
+++ b/src/ddmd/backend/cgcod.c
@@ -693,7 +693,7 @@ Lagain:
     spoff = 0;
     char guessneedframe = needframe;
     int cfa_offset = 0;
-//    if (needframe && config.exe & (EX_LINUX | EX_FREEBSD | EX_SOLARIS) && !(usednteh & ~NTEHjmonitor))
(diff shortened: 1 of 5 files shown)


src/ddmd/backend/cgcod.c
```diff
@@ -882,7 +882,7 @@ void prolog(CodeBuilder& cdb)
                 farfunc ||
                 config.flags & CFGstack ||
                 xlocalsize >= 0x1000 ||
-                (usednteh & ~NTEHjmonitor) ||
+                (usednteh & (NTEH_try | NTEH_except | NTEHcpp | EHcleanup | EHtry | NTEHpassthru)) ||
```
ZombineDev: Looks like a good candidate for refactoring into a new flag, just like `NTEHcpp` is combination of several ones.

WalterBright: There's a lot of refactoring that needs to be done. But one step at a time.

PetarKirov (approved):

JinShil (approved):

PR #7355, opened 2017-11-22
Refactor VRP + fix issue 15289

somzzz:
Moved implementation from ```dcast.d``` to ```intRange.d```.

@andralex Addition and substraction calculate the correct range on overflow, but since there is no type information in ```IntRange``` and ```SignedExtendedNumber```, a ```cast``` is necessary after doing ```ir1 op ir2``` (eg. in ```dcast.d```) to see the correct result based on the data type. Same for all ops.

### DIFF
--- a/src/ddmd/intrange.d
+++ b/src/ddmd/intrange.d
@@ -543,7 +543,7 @@ struct IntRange
     IntRange opBinary(string op : "&")(IntRange rhs) const
     {
         // unsigned or identical sign bits
-        if ((imin.value ^ imax.value) >= 0 && (rhs.imin.value ^ rhs.imax.value) >= 0)
+        if ((imin.negative ^ imax.negative) != 1 && (rhs.imin.negative ^ rhs.imax.negative) != 1)
         {
             return IntRange(minAnd(this, rhs), maxAnd(this, rhs));
         }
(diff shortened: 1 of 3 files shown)


andralex (approved):

andralex:
add coverage

src/ddmd/intrange.d
```diff
@@ -608,73 +603,153 @@ struct IntRange
...
+
+        if (!irNum.imin.negative)
+            irNum.imin.value = 0;
+        else if (irNum.imin < irDen.imin)
+            irNum.imin = irDen.imin;
+
+        if (irNum.imax.negative)
+        {
+            irNum.imax.negative = false;
+            irNum.imax.value = 0;
```
WalterBright: Since this is code added by this PR, it really could use a test case. VRP is so easy to get wrong, having a thorough test suite for code changes is important.

WalterBright (changes requested):
See coverage request.

andralex:
@somzzz unittests are easy to add now, @wilzbach created a unittest target for dmd!

andralex:
pinc @somzzz - let's move fast on this

andralex:
ping @somzzz

andralex:
Per chat with @somzzz I'll pull this first to have her add unittests at a later point.

PR #7358, opened 2017-11-23
error on C++ compiler warnings

WalterBright:
Really should scrub these out of the source code.


PetarKirov (approved):
Not sure how renaming `.h` to `.hpp` helps (is it to force the C compiler in C++ mode?), but it doesn't hurt either.

ZombineDev:
```
ddmd/globals.h:19:21: fatal error: ctfloat.h: No such file or directory
 #include "ctfloat.h"
```

Seems like you've got a bit more work to do.

WalterBright:
> Not sure how renaming .h to .hpp helps

It helps with the clang warning message:
```
clang: warning: treating 'c-header' input as 'c++-header' when in C++ mode, this behavior is deprecated [-Wdeprecated]
```

WalterBright:
> Seems like you've got a bit more work to do.

Oh, there's a lot to do. Each platform generates different warnings, which have been ignored for too long.

ZombineDev:
> It helps with the clang warning message

Thanks for the explanation.

> Oh, there's a lot to do. Each platform generates different warnings, which have been ignored for too long.

Sure, ping when you feel ready and I'll take another look.

jacob-carlborg:
> It helps with the clang warning message:
> ```
> clang: warning: treating 'c-header' input as 'c++-header' when in C++ mode, this behavior is > > deprecated [-Wdeprecated]
> ```

@WalterBright Stackoverflow [1] suggests that that warning occurs when a header file is being compiled, i.e. directly passed to the compiler. Are any header files being compiled? If that's the case perhaps we should look into if it's intentional.

[1] https://stackoverflow.com/questions/23534362/warning-treating-c-header-input-as-c-header-when-in-c-mode-this-behavi

ibuclaw:
Hmm, renaming .h to .hpp is not the right thing to do here.  The compiler is being called with `-fsyntax-only`, it's there to just check that the sources are valid (contributors kept breaking the build).

WalterBright:
> The compiler is being called with -fsyntax-only, it's there to just check that the sources are valid (contributors kept breaking the build).

That's right. As for it being the right thing to do, what's the point of running syntax-only if we just ignore the messages?

ibuclaw:
> what's the point of running syntax-only if we just ignore the messages?

The point of `syntax-only` is so that no one breaks compiler builds.  Too often commas and semi-colons were missed, invalid code was being put in the C++ headers.

That clang is warning is a clang problem, not a header problem.  The headers aren't being compiled, they are just being checked, clang should know better than to emit a warning.

jacob-carlborg:
@WalterBright @ibuclaw `-x c++` can be added as a workaround. The warning does not occur when that flag is specified. It tells the language of the input files.

ibuclaw:
> -x c++ can be added as a workaround. 

That workaround makes no sense, as the host compiler is `clang++` (shouldn't C++ be the assumed language already?).

But fine with that anyway.

jacob-carlborg:
> shouldn't C++ be the assumed language already?

Yes, and it is. Otherwise it would give errors instead of a warning. Using `clang` instead of `clang++` result in syntax errors because it assume the language is C, not C++. So yes, I don't understand the warning.

WalterBright:
If any of you want to post a PR with the makefile fix so the warnings about .h files goes away, I'd be happy to fix this to remove my solution.

ibuclaw:
`-xc++` works with gcc too, so just add it to the check header recipe.

WalterBright:
> -xc++ works with gcc too, so just add it to the check header recipe.

Please submit a PR to do that!

jacob-carlborg:
> Please submit a PR to do that!

Adding `-x c++` is a less disruptive change then renaming all `.h` files.

JinShil:
Attempt at shepherding this across the finish line at #7387

ibuclaw:
Let's close this in favour of then.

PR #2146, opened 2013-06-07
use literal instead of symbol for init

WalterBright:
which enables better code to be generated for small wrapper structs, i.e.:

```
mov EAX,3
```

instead of:

```
mov EAX,S.init
```

for:

```
struct S { int a = 3; this(int I) { } }
S s = S(1);
```

PR #7359, opened 2017-11-23
Use literal instead of symbol for init

JinShil:
Revival of #2146 

I have no idea if this is good idea or not.  I'm just trying to sort out all of the languishing PRs and see what's worth keeping and what isn't.  Also, based on @ibuclaw 's[ comment](https://github.com/dlang/dmd/pull/2146#issuecomment-72274296), I was curious to see what would happen.

### DIFF
--- a/src/ddmd/dsymbolsem.d
+++ b/src/ddmd/dsymbolsem.d
@@ -2443,6 +2443,13 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             }
             else if (auto e = dsym.type.defaultInit(dsym.loc))
             {
+                // For smaller types, prefer the literal as that has better code gen,
+                // using immediate instructions.
+                // Still call the previous defaultInit, as that does checking for void[n]
+                // initializations.
+                if (dsym.type.size() <= 16)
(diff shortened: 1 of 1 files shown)


andralex (approved):

JinShil:
Weird!.  This change only causes a failure on Win32_64.  Here's the assertion that's failing:

```D
// https://issues.dlang.org/show_bug.cgi?id=11672
void test11672()
{
    struct V { float f; }
    struct S
    {
        V[3] v = V(1);
    }

    S s;
    assert(s.v == [V(1), V(1), V(1)]);
}
```

@rainers You're our resident Windows expert, right?  Any ideas?

ibuclaw:
It's still a pointless change that should be fixed in dmd backend.

JinShil:
For some unknown reason this change causes Win32_64 tests to fail as mentioned above.  Also, it may be better if such an optimization were implemented in the backed;  as [the argument for implementing this in the frontend](https://github.com/dlang/dmd/pull/2146#issuecomment-20168949) seemed to be one of expediency over due diligence.  Therefore, closing.

An issue was created at https://issues.dlang.org/show_bug.cgi?id=18009 to document the potential optimization that is the subject of this PR.

PR #7360, opened 2017-11-25
improve back end debug printing of intermediate code

WalterBright:
It's a lot easier to read when blocks are numbered rather than random pointer values.


stefan-koch-sociomantic (approved):
This is a step into the right direction.

src/ddmd/backend/debug.c
```diff
@@ -318,25 +319,34 @@ void WRblock(block *b)
...
+            printf(" b_ret=B%d", b->BS.BI_FINALLY.b_ret ? b->BS.BI_FINALLY.b_ret->Bnumber : 0);
 #if MARS
         if (b->Bsrcpos.Sfilename)
             printf(" %s(%u)", b->Bsrcpos.Sfilename, b->Bsrcpos.Slinnum);
 #endif
         printf("\n");
-        if (b->Belem) elem_print(b->Belem);
+        if (b->Belem)
+        {
+            if (0)
```
JinShil: Should this be `false` instead of `0`?

JinShil (changes requested):
~I'm not able to comment on the functionality of this PR, as I don't quite understand what it does.  But I've reviewed the code itself, and have found no issues.~

Sorry a didn't mean to send that, and apparently Github won't allow me to delete it.

src/ddmd/backend/debug.c
JinShil: The reason I ask is because of #7310

JinShil (approved):

src/ddmd/backend/debug.c
WalterBright: `0` is never going to be an incorrect argument to `if`.

PR #7363, opened 2017-11-25
rootobject.d: add nothrow

WalterBright:


PetarKirov (approved):

ZombineDev:
(The TravisCI builds that are failing should go green when restarted as its a network related error.)

PR #7366, opened 2017-11-26
add OutBuffer.print()

WalterBright:
This adds a special formatter to OutBuffer to handle formatted printing of integers, a common case. The replacement is faster and safer. In dmangle.d, it also gets rid of a number of problematic casts.

### DIFF
--- a/src/ddmd/root/outbuffer.d
+++ b/src/ddmd/root/outbuffer.d
@@ -342,6 +342,18 @@ struct OutBuffer
         va_end(ap);
     }
 
+    /**************************************
+     * Convert `u` to a string and append it to the buffer.
+     * Params:
+     *  u = integral value to append
+     */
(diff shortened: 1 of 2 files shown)


src/ddmd/root/outbuffer.d
```diff
@@ -417,3 +429,34 @@ struct OutBuffer
         return extractData();
     }
 }
+
+/****** copied from core.internal.string *************/
```
JinShil: Would love to find a way to centralize this code somewhere.  I've found it useful in bare-metal/no-runtime programming also.

JinShil (approved):

JinShil:
Looks like this may be blocking https://github.com/dlang/dmd/pull/7364....
....Ah, what the hell.  Automerge toggling on.

PR #7368, opened 2017-11-26
more print() conversion

WalterBright:
Threw in a few nothrow's, too.


src/ddmd/globals.d
```diff
@@ -402,13 +405,16 @@ struct Loc
         OutBuffer buf;
         if (filename)
         {
-            buf.printf("%s", filename);
+            buf.writestring(filename);
         }
         if (linnum)
         {
-            buf.printf("(%d", linnum);
+            buf.print(linnum);
```
JinShil: It looks like you're missing the open parens here.

JinShil (changes requested):

JinShil (approved):

PR #7372, opened 2017-11-28
refactor: Move struct ClassFlags to toobj.d and make it private

RazvanN7:


src/ddmd/aggregate.h
```diff
@@ -230,23 +230,6 @@ struct BaseClass
     void copyBaseInterfaces(BaseClasses *);
 };
 
-struct ClassFlags
-{
-    typedef unsigned Type;
```
JinShil: Will this negatively affect the C++ interface?  cc @ibuclaw

src/ddmd/toobj.d
```diff
@@ -75,6 +75,33 @@ extern (C++):
 alias toSymbol = ddmd.tocsym.toSymbol;
 alias toSymbol = ddmd.glue.toSymbol;
 
+private struct ClassFlags
```
JinShil: While you're at it, it wouldn't hurt to add a DDOC comment block.  I think @WalterBright  is [encouraging us to do so](https://github.com/dlang/dmd/pull/7341#discussion_r152198939) even for private code.

src/ddmd/aggregate.h
ZombineDev: @kinke @JohanEngelen would this be a problem for LDC?

kinke: Yes it would indeed: https://github.com/ldc-developers/ldc/blob/d1b30d627c5bbb58ff306aa924805871dbda806d/gen/classes.cpp#L491

ibuclaw: Ditto gdc.

To keep everyone synchronised, it should remain in the frontend.

ibuclaw:
???

JinShil:
> ???

Yeah, me too.

PR #7375, opened 2017-11-28
Add -requireinit compiler switch to emit errors for uninitialized variables

JinShil:
[The D programming language specification](https://dlang.org/spec/function.html#local-variables) clearly states:

>It is an error to use a local variable without first assigning it a value. The implementation may not always be able to detect these cases. Other language compilers sometimes issue a warning for this, but since it is always a bug, it should be an error. 

Surprisingly, this requirement has never been implemented in the D programming language.  Perhaps the reason it was never implemented was due to the high bar required to write sophisticated flow control analysis for catching false positives, as not all uninitialized variables are used. Despite that, it would have been better to require initialization at first, knowing that it would create false positives, and as proper flow control analysis was implemented, relax the enforcement to eliminate the false positives and prevent code breakage along the way.

Unfortunately, here we are years later, with an entire ecosystem of D code that violates this requirement.  It is not feasible to implement strict enforcement out-of-the-box now, this late in the game, but we can provide users with a way to opt into it, so they can have more confidence in the robustness of their D code.

This PR adds a `-requireinit` compiler switch to enable strict enforcement of variable initialization.  When the compiler is executed with this switch, it will emit an error for any variable that is not explicitly initialized.  This will produce false positives, as not all uninitialized variables will be used, but in time, if/when proper flow control analysis is implemented, the enforcement can be relaxed to eliminate false positives, while preventing code breakage along the way.

Changelog entry and documentation will be added after initial reaction to this proposal is received.

### DIFF
--- a/src/ddmd/parse.d
+++ b/src/ddmd/parse.d
@@ -4495,7 +4495,7 @@ final class Parser(AST) : Lexer
                 else
                 {
                     if (global.params.requireinit)
-                        error("`%s` was not iniitialized", ident.toChars());
+                        error("`%s` was not initialized", ident.toChars());
                 }
 
 
--- a/src/ddmd/parse.d
+++ b/src/ddmd/parse.d
@@ -4492,6 +4492,12 @@ final class Parser(AST) : Lexer
                     nextToken();
                     _init = parseInitializer();
                 }
+                else
+                {
+                    if (global.params.requireinit)
+                        error("`%s` was not iniitialized", ident.toChars());
+                }
+
 
                 auto v = new AST.VarDeclaration(loc, t, ident, _init);
                 v.storage_class = storage_class;
--- a/test/fail_compilation/requireinit.d
+++ b/test/fail_compilation/requireinit.d
@@ -1,0 +1,34 @@
+/*
+REQUIRED_ARGS: -requireinit
+TEST_OUTPUT:
+---
+fail_compilation/requireinit.d(21): Error: `tc` was not iniitialized
+fail_compilation/requireinit.d(22): Error: `ts` was not iniitialized
+fail_compilation/requireinit.d(23): Error: `s` was not iniitialized
+fail_compilation/requireinit.d(24): Error: `i` was not iniitialized
+fail_compilation/requireinit.d(25): Error: `a` was not iniitialized
+fail_compilation/requireinit.d(29): Error: `ltc` was not iniitialized
+fail_compilation/requireinit.d(30): Error: `lts` was not iniitialized
+fail_compilation/requireinit.d(31): Error: `ls` was not iniitialized
+fail_compilation/requireinit.d(32): Error: `li` was not iniitialized
+fail_compilation/requireinit.d(33): Error: `la` was not iniitialized
+---
+*/
+
+class TestClass {}
+struct TestStruct {}
+
+TestClass tc;
+TestStruct ts;
+string s;
+int i;
+int[] a;
+
+void main()
+{
+    TestClass ltc;
+    TestStruct lts;
+    string ls;
+    int li;
+    int[] la;
+}
--- a/test/fail_compilation/requireinit.d
+++ b/test/fail_compilation/requireinit.d
@@ -2,21 +2,46 @@
 REQUIRED_ARGS: -requireinit
 TEST_OUTPUT:
 ---
-fail_compilation/requireinit.d(21): Error: `tc` was not iniitialized
-fail_compilation/requireinit.d(22): Error: `ts` was not iniitialized
-fail_compilation/requireinit.d(23): Error: `s` was not iniitialized
-fail_compilation/requireinit.d(24): Error: `i` was not iniitialized
-fail_compilation/requireinit.d(25): Error: `a` was not iniitialized
-fail_compilation/requireinit.d(29): Error: `ltc` was not iniitialized
-fail_compilation/requireinit.d(30): Error: `lts` was not iniitialized
-fail_compilation/requireinit.d(31): Error: `ls` was not iniitialized
-fail_compilation/requireinit.d(32): Error: `li` was not iniitialized
-fail_compilation/requireinit.d(33): Error: `la` was not iniitialized
+fail_compilation/requireinit.d(30): Error: `tc` was not initialized
+fail_compilation/requireinit.d(31): Error: `ts` was not initialized
+fail_compilation/requireinit.d(32): Error: `s` was not initialized
+fail_compilation/requireinit.d(33): Error: `i` was not initialized
+fail_compilation/requireinit.d(34): Error: `a` was not initialized
+fail_compilation/requireinit.d(39): Error: `tc` was not initialized
+fail_compilation/requireinit.d(40): Error: `ts` was not initialized
+fail_compilation/requireinit.d(41): Error: `s` was not initialized
+fail_compilation/requireinit.d(42): Error: `i` was not initialized
+fail_compilation/requireinit.d(43): Error: `a` was not initialized
+fail_compilation/requireinit.d(46): Error: `tc` was not initialized
+fail_compilation/requireinit.d(47): Error: `ts` was not initialized
+fail_compilation/requireinit.d(48): Error: `s` was not initialized
+fail_compilation/requireinit.d(49): Error: `i` was not initialized
+fail_compilation/requireinit.d(50): Error: `a` was not initialized
+fail_compilation/requireinit.d(54): Error: `ltc` was not initialized
+fail_compilation/requireinit.d(55): Error: `lts` was not initialized
+fail_compilation/requireinit.d(56): Error: `ls` was not initialized
+fail_compilation/requireinit.d(57): Error: `li` was not initialized
+fail_compilation/requireinit.d(58): Error: `la` was not initialized
 ---
 */
 
-class TestClass {}
-struct TestStruct {}
+class TestClass
+{
+    TestClass tc;
+    TestStruct ts;
+    string s;
+    int i;
+    int[] a;
+}
+
+struct TestStruct
+{
+    TestClass tc;
+    TestStruct ts;
+    string s;
+    int i;
+    int[] a;
+}
 
 TestClass tc;
 TestStruct ts;
--- a/src/ddmd/globals.d
+++ b/src/ddmd/globals.d
@@ -142,6 +142,8 @@ struct Param
      * before becoming default.
      */
 
+    bool requireinit;      // emit error for any uninitialized variables
+
     bool showGaggedErrors;  // print gagged errors anyway
     bool manual;            // open browser on compiler manual
     bool usage;             // print usage and exit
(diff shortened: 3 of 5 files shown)


nemanja-boric-sociomantic:
Aren't we with this going back to the C's land where the local variables are not auto-initialised to the `type.init` (which is the case in D)? When I read this PR I thought - "oh, ok, this is cool  - if the compiler can warn you if you're reading the variable that was _explicitly_ void-initialized", but looking into the test cases, I don't think this is a good addition. Perhaps updating the specs is better?

mathias-lang-sociomantic:
> knowing that it would create false positives, and as proper flow control analysis was implemented, relax the enforcement to eliminate the false positives and prevent code breakage along the way.

False positive in a compiler are never okay.
The second you have false positive, the compiler becomes your enemy instead of being here to help.
The second you tolerate false positive, you have to use warnings instead of error, and make those warnings optional because some warnings are just wrong, and if they trigger in code outside of the user's control, tough luck. Don't ever *accept* false positive.

Now regarding the pull-request, I think you might have misunderstood that passage of the spec. In particular:
> It is an error to use a local variable without first assigning it a value.
Does not say "without first *explicitly* assigning it a value".
Since we have default initialization, doing `int i; i += 5;` is perfectly okay. However, `int i = void; i += 5` is *not* okay.

JinShil:
> Since we have default initialization, doing `int i; i += 5;` is perfectly okay

Yes, but `float f; f += 5.0f` is not okay.  `MyClass c; c.myFunction()` is also not okay.  So there is an inconsistency in D's default initialization.

It was [this forum post](http://forum.dlang.org/post/pptoytctpvrxplckiawy@forum.dlang.org)...

> You ARE supposed to explicitly initialize variables to your own values in D. The automatic init is to make errors stand out more consistently if you don't do this as opposed to being random.

> So pointers are initialized to null - an invalid value that stands out if you try to use it. chars get \xFF - again, invalid that will throw if you try to utf decode it. Floats get NaN which is as close to invalid as they get.

> ints happen to get 0 not to be convenient, but because there is no clearly-invalid int value so something had to be chosen, and 0 was just easy to implement....

... and [this forum post](http://forum.dlang.org/post/tcldaatzzbhjoamnvniu@forum.dlang.org)...

> "It is an error to use a local variable without first assigning it a value. The implementation may not always be able to detect these cases. Other language compilers sometimes issue a warning for this, but since it is always a bug, it should be an error. "

> But that paragraph was never implemented (it is a bit easier said than done to actually detect given if conditions and stuff too, especially in the early days of D when the compiler didn't even make any effort to track such things, though it does now...). The compiler author took the easy way out of initializing them to invalid values instead.

> While it is more realistic to implement technically now than it was years ago when the current behavior got in, I think we're at the point now that so many people use it as a convenience thing on ints and nulls that there'd be hell to pay if the compiler actually started calling it an error :(

... that prompted this PR.  cc @adamdruppe 

> Perhaps updating the specs is better?

The spec is right "...since it is always a bug, it should be an error".  This PR is obviously not the right solution, but I'm sure there's a way to make something like this work.

ZombineDev:
@JinShil I would definitely welcome [definite assignment analysis](https://en.wikipedia.org/wiki/Definite_assignment_analysis) for `void` initialized variables, and [non-null reference types](https://github.com/dotnet/csharplang/blob/master/proposals/nullable-reference-types.md) and pointers (basically a better version, of C++'s `T&` for D class references and pointers), but I am against making default-initialization an error.  
There's a ton of code out there that relies on it. There's even a dscanner check that verifies that you're doing "useless initialization": https://github.com/dlang-community/D-Scanner/blob/master/src/analysis/useless_initializer.d, as many regard as idiomatic D code that relies on default init.

However I see value in allowing code bases to follow their own coding style, so this flag may be a good candidate for my "language configuration syntax" DIP that I'm writing.

JinShil:
> I am against making default-initialization an error. 

`-requireinit` is opt-in and not enabled by default.  It's just a tool users can choose to employ or ignore.  `-requireinit` could be easily modified to emit warnings instead of errors, if doing so would be a better solution, but I don't see at this time why one would want a warning instead of an error.

> There's a ton of code out there that relies on it.

I know.  The absence of proper enforcement has caused proliferation of this anti-pattern.  The fact that it is so wide-spread is why `-requireinit` is an optional tool and not enabled by default.

> There's even a dscanner check that verifies that you're doing "useless initialization":

I believe the dscanner rule is wrong. 

@adamdruppe 's [forum post](http://forum.dlang.org/post/pptoytctpvrxplckiawy@forum.dlang.org) was an epiphany for me.  It now makes sense why `float`s are initialized to `NaN` instead of `0.0f`, `class` instances are initialized to `null`, and even why `int`s are initialized to `0` by default.  It wasn't done for the programmer's convenience; it was done to increase the likelihood of program failure, so the error would be caught early and quickly.  And, with that realization, one can see that the design is consistent.

The false positives in this case are not a problem, as one can easily correct the error with e.g. `int myInt = int.init`.  With `-requireinit` the compiler is simply requiring the programmer to state their intent, as uninitialized variables are an ambiguity.  It is no different than the compiler requiring an explicit cast.

schveiguy:
The problem I see with this is, Phobos and druntime need to support this possibility. Otherwise, if you pass -requireinit for your program, and try to use phobos, you will get errors that aren't your fault.

This would mean that all Phobos/druntime code would need to be updated to explicit initialization. I just don't see that happening.

src/ddmd/parse.d
```diff
@@ -4492,6 +4492,12 @@ final class Parser(AST) : Lexer
                     nextToken();
                     _init = parseInitializer();
                 }
+                else
+                {
+                    if (global.params.requireinit)
+                        error("`%s` was not iniitialized", ident.toChars());
```
carun: s/iniitialized/initialized/

test/fail_compilation/requireinit.d
```diff
@@ -0,0 +1,34 @@
+/*
+REQUIRED_ARGS: -requireinit
+TEST_OUTPUT:
+---
+fail_compilation/requireinit.d(21): Error: `tc` was not iniitialized
```
carun: s/iniitialized/initialized/

yyp error probably? :

jmdavis:
IMHO, _way_ too much code just default-initializes variables for it to make it an error to rely on default initialization. And honestly, while I do think that it's generally better to explicitly initialize stuff take make intent clear, I don't really think that it's a problem for code to rely on default initialization. The fact that we have default initialization prevents all of the problems that you get in C/C++ with forgetting to initialize stuff, and having default initialization has proven to be a big boon in a variety of areas. Also, I am _completely_ against anything in the compiler that is ever going to generate false positives. Walter posted just recently in the newsgroup about how problematic it was to have the compiler try and detect when a variable is null, and the same goes for any other kind of initialization detection.

It sounds like the spec needs to be updated, since it doesn't make sense to make it an error even with void initialization, because you have all of the same problems with that as any kind of initialization detection. I suspect that that entry in the spec stems from the idea that ideally, `init` would be an invalid value, allowing you to catch when you failed to initialize a variable, and you would explicitly initialize everything, but making it an error to rely on default initialization is just going to result in false positives, and in practice, there are a number of places where it just works way better to take advantage of default initialization (e.g. when dealing with arrays). Honestly, the very fact that we _have_ default initialization flies in the face of making it an error to not explicitly initialize something. `.init` primary job is to solve what this PR is trying to solve, and `init` does not result in the compiler giving errors when it shouldn't.

WalterBright:
The spec, while pedantically correct, is unclear in that it does not mention the role of default initialization. Default initialization ensures that nobody gets an uninitialized value, unless `=void` is used.

> Yes, but float f; f += 5.0f is not okay. MyClass c; c.myFunction() is also not okay. So there is an inconsistency in D's default initialization.

Depends on what you mean by "not okay". The first is well defined, `f` is set to NaN, which has defined behavior. The second is more or less defined behavior, you'll get a seg fault. This is very different from uninitialized behavior, where you get *random* behavior.

D's semantics are very dependent on default initialization, and I would be very reluctant to add switches implying that this is erroneous and that people should "fix" their code. I oppose this change.

(BTW, default initialization is rarely a performance issue, because the optimizer can detect and eliminate most "dead" assignments. The only time it is much of an issue is when having large arrays on the stack used as buffers that are only partially used. `=void` comes in handy there.)

JinShil:
> The first is well defined, `f` is set to `NaN`, which has defined behavior. 

Setting `f` to `0.0f` would also be defined behavior.  But, `NaN` was chosen instead;  why?  Is there merit to @adamdruppe's analysis below?

> You ARE supposed to explicitly initialize variables to your own values in D. The automatic init is to make errors stand out more consistently if you don't do this as opposed to being random.

> So pointers are initialized to null - an invalid value that stands out if you try to use it. chars get \xFF - again, invalid that will throw if you try to utf decode it. Floats get NaN which is as close to invalid as they get.

> ints happen to get 0 not to be convenient, but because there is no clearly-invalid int value so something had to be chosen, and 0 was just easy to implement....

WalterBright:
> But, NaN was chosen instead; why?

As Adam said, so errors will be clear. It's exactly what NaNs are for.

Requiring a user supplied initializer has its own problems. For example,
```
int i;
if (condition) i = 3; else i = 4;
```
Now, the compiler complains about `i` not being initialized, even though it is. (I can make the second line complex enough that the data flow analysis won't be able to prove it is initialized, even if it is.) So what does the coder do? Most likely, he'll just pick a random value just to get the compiler to shut up.

The next maintainer comes along, and is now baffled by the purpose of the assignment with a value that makes no sense. I've seen it happen often enough to convince me that the cure is worse than the problem. Note that the C problem with uninitialized variables (random results) does not happen with D.

schlupa:
I'm with JinShil and Adam Ruppe on this one. Default initialisation is not a suffisant condition to avoid bugs. It can even hides some class of bugs, the ones where the default init value is the wrong value. Choosing an arbitrary default init value is just that, an arbitrary value. While 0 for ints is very often the right choice, it is not necessarly so.

schveiguy:
As Walter said, if the compiler complains about default initialization, the coder is just going to put in a default value that likely still hides those bugs. Now, sure, we can say it's their fault, not the compiler's, but really, it's still the same problem, we haven't fixed it.

Contrast that to someone who KNOWS the default value is 0 or NaN, and uses it as a convenience feature (that is what they would normally initialize with). I do this all the time. All that this change is going to do is force me to be explicit, for zero gain. When we make language changes like this, it's always a tradeoff -- how much pain do we save vs. how much do we cause. I'd argue the downside far outweighs the upside.

JinShil:
Ok, that was a fun experiment.  Time to move on.  Closing.

DmitryOlshansky:
> Ok, that was a fun experiment. Time to move on. Closing.

I think quite a few people would appreciate a lint tool like that, depending on their style of course.

wilzbach:
> I think quite a few people would appreciate a lint tool like that, depending on their style of course.

+1 - this could be a good initial use case for the compiler as a library project. CC @RazvanN7

PR #7376, opened 2017-11-29
add noeh RAII optimizations at expression level

WalterBright:
When temporaries are generated while evaluating an expression, the destructors for them must be called at the end of that expression. This PR checks the expression to see if it is nothrow, and if so, generates much more efficient code for RAII destruction.

It also resolves a memory leak in `varsInScope` by allocating it on the stack.

### DIFF
--- a/src/ddmd/e2ir.d
+++ b/src/ddmd/e2ir.d
@@ -26,6 +26,7 @@ import ddmd.root.stringtable;
 import ddmd.aggregate;
 import ddmd.arraytypes;
 import ddmd.attrib;
+import ddmd.canthrow;
 import ddmd.ctfeexpr;
 import ddmd.dclass;
 import ddmd.declaration;
@@ -5174,8 +5175,6 @@ elem *toElem(Expression e, IRState *irs)
                         }
 
                         // ed needs to be inserted into the code later
-                        if (!irs.varsInScope)
-                            irs.varsInScope = new Array!(elem*)();
                         irs.varsInScope.push(ed);
                     }
                 }
@@ -5794,9 +5793,25 @@ private elem *appendDtors(IRState *irs, elem *er, size_t starti, size_t endi)
 elem *toElemDtor(Expression e, IRState *irs)
 {
     //printf("Expression.toElemDtor() %s\n", e.toChars());
-    const starti = irs.varsInScope ? irs.varsInScope.dim : 0;
+
+    /* "may" throw may actually be false if we look at a subset of
+     * the function. Here, the subset is `e`. If that subset is nothrow,
+     * we can generate much better code for the destructors for that subset,
+     * even if the rest of the function throws.
+     * If mayThrow is false, it cannot be true for some subset of the function,
+     * so no need to check.
+     * If calling canThrow() here turns out to be too expensive,
+     * it can be enabled only for optimized builds.
+     */
+    const mayThrowSave = irs.mayThrow;
+    if (irs.mayThrow && !canThrow(e, irs.getFunc(), false))
+        irs.mayThrow = false;
+
+    const starti = irs.varsInScope.dim;
     elem* er = toElem(e, irs);
-    const endi = irs.varsInScope ? irs.varsInScope.dim : 0;
+    const endi = irs.varsInScope.dim;
+
+    irs.mayThrow = mayThrowSave;
 
     // Add destructors
     elem* ex = appendDtors(irs, er, starti, endi);
--- a/src/ddmd/glue.d
+++ b/src/ddmd/glue.d
@@ -977,11 +977,10 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
             break;
     }
 
-    IRState irs = IRState(m, fd);
     Dsymbols deferToObj;                   // write these to OBJ file later
-    irs.deferToObj = &deferToObj;
+    Array!(elem*) varsInScope;
     Label*[void*] labels = null;
-    irs.labels = &labels;
+    IRState irs = IRState(m, fd, &varsInScope, &deferToObj, &labels);
 
     Symbol *shidden = null;
     Symbol *sthis = null;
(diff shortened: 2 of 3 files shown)


src/ddmd/irstate.d
```diff
@@ -76,13 +77,19 @@ struct IRState
             deferToObj = irs.deferToObj;
             varsInScope = irs.varsInScope;
             labels = irs.labels;
+            mayThrow = irs.mayThrow;
         }
     }
 
-    this(Module m, FuncDeclaration s)
+    this(Module m, FuncDeclaration fd, Array!(elem*)* varsInScope, Dsymbols* deferToObj, Label*[void*]* labels)
```
ZombineDev: As far as I can see, you converted the only place that sets the `labels`  and `deferToObj` fields to use this constructor. If you want to improve the encapsulation, you can now mark those fields as `private`, rename them by prepending an underscore and expose them via a getter properties with the original name.

src/ddmd/irstate.d
```diff
@@ -53,8 +53,9 @@ struct IRState
     Dsymbols* deferToObj;           // array of Dsymbol's to run toObjFile(bool multiobj) on later
     elem* ehidden;                  // transmit hidden pointer to CallExp::toElem()
     Symbol* startaddress;
-    Array!(elem*)* varsInScope;      // variables that are in scope that will need destruction later
+    Array!(elem*)* varsInScope;     // variables that are in scope that will need destruction later
     Label*[void*]* labels;          // table of labels used/declared in function
```
ZombineDev: I wonder if we can change this to `Label*[void*]`, since the AA is already a reference type. This should allow to remove the `Label*[void*] labels = null;` variable declaration in https://github.com/dlang/dmd/pull/7376/files#diff-db198d44f47523b339ccc64c78f8cb53R982 and pass `null` to the constructor.

src/ddmd/glue.d
```diff
@@ -977,11 +977,10 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
             break;
     }
 
-    IRState irs = IRState(m, fd);
     Dsymbols deferToObj;                   // write these to OBJ file later
-    irs.deferToObj = &deferToObj;
+    Array!(elem*) varsInScope;
     Label*[void*] labels = null;
```
DmitryOlshansky: Wow, we are actually using D's AAs in dmd. Nice :)

src/ddmd/irstate.d
```diff
@@ -76,13 +77,19 @@ struct IRState
...
         }
     }
 
-    this(Module m, FuncDeclaration s)
+    this(Module m, FuncDeclaration fd, Array!(elem*)* varsInScope, Dsymbols* deferToObj, Label*[void*]* labels)
     {
         this.m = m;
-        symbol = s;
+        this.symbol = fd;
+        this.varsInScope = varsInScope;
```
ZombineDev: I see that the assumption in the rest of the code is that `irs.varsInScope` can't be `null`. How about changing the constructor parameter to `ref Array!(elem*) varsInScope` to solidify this assumption?

src/ddmd/e2ir.d
```diff
@@ -5794,9 +5793,25 @@ private elem *appendDtors(IRState *irs, elem *er, size_t starti, size_t endi)
...
+    const mayThrowSave = irs.mayThrow;
+    if (irs.mayThrow && !canThrow(e, irs.getFunc(), false))
+        irs.mayThrow = false;
+
+    const starti = irs.varsInScope.dim;
     elem* er = toElem(e, irs);
-    const endi = irs.varsInScope ? irs.varsInScope.dim : 0;
+    const endi = irs.varsInScope.dim;
+
+    irs.mayThrow = mayThrowSave;
```
ZombineDev: Can you move the declarations for `starti` and `endi` below this line, as they're not related to determining if the expressing may throw, as far as I can see.

ZombineDev:
Looks good, apart from a couple of coding style suggestions.

src/ddmd/irstate.d
WalterBright: Because then I have to take the address of a `ref` parameter which is not allowed in `@safe` code.

src/ddmd/e2ir.d
WalterBright: They have to bracket the call to `toElem()`.

src/ddmd/irstate.d
WalterBright: Besides, if it is null, it'll seg fault on first use. So I'm not worried about it :-)

src/ddmd/e2ir.d
ZombineDev: Yes, you're right - I see that I overlooked the code. Sorry for the noise.

src/ddmd/glue.d
WalterBright: :+1:

src/ddmd/irstate.d
WalterBright: I wanted the AA handle to be outside IRState for the same reason as varsInScope - so the destructor can be called on it there. I know there's no destructor for AAs, but I felt it should follow the same pattern as varsInScope anyway. And who knows, maybe we can replace it with a destructing AA and plug another leak.

src/ddmd/irstate.d
ZombineDev: > Besides, if it is null, it'll seg fault on first use. So I'm not worried about it getting through the test suite with such a bug :-)

I'm not that worried about bugs, as much I would like to make the interface more clear for the callers.

> Because then I have to take the address of a ref parameter which is not allowed in @safe code.

Taking the address of a local variable and passing it to a non-`scope` parameter is not `@safe` either. Ideally (in the distant future) the ownership could be clarified, so even the the compiler could see it is `@safe`:
```D
// in some function
{
    Array!(elem*) varsInScope;
    {
        // it's safe to pass `varsInScope` as `lifetime(varsInScope) > lifetime(irs)`
        IRState irs = IRState(.., /* scope ref parameter */ varsInScope, ..);
    }
}
```


Anyway, it's just a suggestion - we can revisit the discussion some other day if you want to move forward with the code as it is.

PetarKirov (approved):
I really like that we're now doing more fine-grained throw analysis and I'm interested on the possibilities that it can unlock. Overall the changes are clear, so I'm in favor of going forward.
I'm looking into doing code refactoring myself, so I can revisit some of my suggestion in the future.

PR #7377, opened 2017-11-29
toir.d: minor refactor

WalterBright:
Also added an overflow check to the size calculation.

### DIFF
--- a/src/ddmd/toir.d
+++ b/src/ddmd/toir.d
@@ -12,6 +12,7 @@
 
 module ddmd.toir;
 
+import core.checkedint;
 import core.stdc.stdio;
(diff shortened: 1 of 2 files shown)


src/ddmd/toir.d
```diff
@@ -753,7 +771,7 @@ void setClosureVarOffset(FuncDeclaration fd)
              */
             if (fd.nrvo_can && fd.nrvo_var == v)
             {
-                fd.nrvo_can = 0;
+                fd.nrvo_can = false;
```
JinShil: Good!

JinShil (approved):

PR #7378, opened 2017-11-30
Fix Issue 18020 - [Reg 2.078] no property opCmp for anon class

JinShil:
`ClassKind.anonymous` is not a linkage type.

In this case, this should be targeting "master", right?

cc @RazvanN7

### DIFF
--- a/src/ddmd/dclass.d
+++ b/src/ddmd/dclass.d
@@ -193,8 +193,6 @@ enum ClassKind : int
     cpp,
     /// the class is an Objective-C class/interface
     objc,
-    /// the class is anonymous
-    anonymous,
 }
 
 /***********************************************************
@@ -242,6 +240,10 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
     /// to prevent recursive attempts
     private bool inuse;
 
+    /// true if this class has an identifier, but was originally declared anonymous
+    /// used in support of https://issues.dlang.org/show_bug.cgi?id=17371
+    private bool isActuallyAnonymous;
+
     Abstract isabstract;
 
     /// set the progress of base classes resolving
@@ -254,7 +256,7 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
         if (!id)
         {
             id = Identifier.generateId("__anonclass");
-            classKind = ClassKind.anonymous;
+            isActuallyAnonymous = true;
         }
         assert(id);
 
@@ -707,6 +709,11 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
 
         // Calculate fields[i].overlapped
         checkOverlappedFields();
+    }
+
+    override bool isAnonymous()
+    {
+        return isActuallyAnonymous;
     }
 
     final bool isFuncHidden(FuncDeclaration fd)
--- a/src/ddmd/dsymbol.d
+++ b/src/ddmd/dsymbol.d
@@ -269,7 +269,7 @@ extern (C++) class Dsymbol : RootObject
         return false;
     }
 
-    final bool isAnonymous()
+    bool isAnonymous()
     {
         return ident is null;
     }
--- a/src/ddmd/todt.d
+++ b/src/ddmd/todt.d
@@ -719,7 +719,7 @@ private void membersToDt(AggregateDeclaration ad, DtBuilder dtb,
         {
             dtb.xoff(toVtblSymbol(concreteType), 0);  // __vptr
             offset = Target.ptrsize;
-            if (!cd.classKind == ClassKind.cpp)
+            if (cd.classKind != ClassKind.cpp)
             {
                 dtb.size(0);              // __monitor
                 offset += Target.ptrsize;
--- a/src/ddmd/aggregate.h
+++ b/src/ddmd/aggregate.h
@@ -254,9 +254,7 @@ enum ClassKind
     /// the class is a C++ interface
     cpp,
     /// the class is an Objective-C class/interface
-    objc,
-    /// the class is anonymous
-    anonymous,
+    objc
 };
 
 class ClassDeclaration : public AggregateDeclaration
@@ -285,8 +283,11 @@ public:
 
     TypeInfoClassDeclaration *vclassinfo;       // the ClassInfo object for this ClassDeclaration
     bool com;                           // true if this is a COM class (meaning it derives from IUnknown)
-    bool stack;                       // true if this is a scope class
-    ClassKind classKind;               // specifies the linkage type
+    bool stack;                         // true if this is a scope class
+    ClassKind classKind;                // specifies the linkage type
+    bool inuse;                         // to prevent recursive attempts
+    bool isActuallyAnonymous;           // true if this class has an identifier, but was originally declared anonymous
+                                        // used in support of https://issues.dlang.org/show_bug.cgi?id=17371
 
     Abstract isabstract;                // 0: fwdref, 1: is abstract class, 2: not abstract
     Baseok baseok;                      // set the progress of base classes resolving
--- a/src/ddmd/hdrgen.d
+++ b/src/ddmd/hdrgen.d
@@ -1756,7 +1756,7 @@ public:
 
     override void visit(ClassDeclaration d)
     {
-        if (d.classKind != ClassKind.anonymous)
+        if (!d.isAnonymous())
         {
             buf.writestring(d.kind());
             buf.writeByte(' ');
@@ -1783,7 +1783,7 @@ public:
     {
         if (!d || !d.baseclasses.dim)
             return;
-        if (d.classKind != ClassKind.anonymous)
+        if (!d.isAnonymous())
             buf.writestring(" : ");
         foreach (i, b; *d.baseclasses)
         {
--- a/src/ddmd/tocvdebug.d
+++ b/src/ddmd/tocvdebug.d
@@ -538,7 +538,7 @@ void toDebug(ClassDeclaration cd)
     //printf("ClassDeclaration::toDebug('%s')\n", cd.toChars());
 
     assert(config.fulltypes >= CV4);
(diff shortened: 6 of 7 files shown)


WalterBright (approved):
Mind the Jenkins failures

jacob-carlborg:
> ClassKind.anonymous is not a linkage type.

Technically classes don't have linkage a type, only functions do. In the previous PR #7315, a couple of mutually exclusive booleans were converted to a single enum to actually enforce the exclusiveness. For example, a class cannot be anonymous and C++ or Objective-C at the same time.

JinShil:
> a class cannot be anonymous and C++ or Objective-C at the same time.

... but it can be `ClassKind.d` and anonymous at the same time, right?  Hence the regression.  I think the previous PR had too much refactoring in it and navigated away from the issue at hand.

RazvanN7 (approved):
This seems to be the right way, but please also fix the issue stated in #7315

jacob-carlborg:
> but it can be ClassKind.d and anonymous at the same time, right?

Yes.

JinShil:
> but please also fix the issue stated in #7315

Done.

src/ddmd/todt.d
```diff
@@ -719,7 +719,7 @@ private void membersToDt(AggregateDeclaration ad, DtBuilder dtb,
         {
             dtb.xoff(toVtblSymbol(concreteType), 0);  // __vptr
             offset = Target.ptrsize;
-            if (!cd.classKind == ClassKind.cpp)
+            if (cd.classKind != ClassKind.cpp)
```
jacob-carlborg: Should we look for more of these?

src/ddmd/dclass.d
```diff
@@ -242,6 +240,10 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
     /// to prevent recursive attempts
     private bool inuse;
 
+    /// true if this class has an identifier, but was originally declared anonymous
```
jacob-carlborg: Do we really need to mention the second part, after the comma?

src/ddmd/todt.d
JinShil: I did. I didn't find any.  Feel free to check yourself.

src/ddmd/dclass.d
```diff
@@ -242,6 +240,10 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
...
+    /// used in support of https://issues.dlang.org/show_bug.cgi?id=17371
+    private bool isActuallyAnonymous;
```
jacob-carlborg: the D convention is, as far as I know, to use the same name as the getter but prefix it with an underscore.

jacob-carlborg:
The C++ header need to be updated as well.

src/ddmd/dclass.d
JinShil: IMO, yes.

src/ddmd/dclass.d
JinShil: I haven't seen that.  Show me where the precedent is established.

JinShil:
> The C++ header need to be updated as well.

aggregate.h was updated.  Are you referring to something else?

jacob-carlborg:
>  aggregate.h was updated. Are you referring to something else?

No, but the new field was not added.

JinShil:
> No, but the new field was not added.

Which new field?

RazvanN7:
@jacob-carlborg It shouldn't, since it is private.

jacob-carlborg:
`isActuallyAnonymous`.

jacob-carlborg:
The sizes of the D and the C++ classes need to match, as far as I know.

src/ddmd/dclass.d
jacob-carlborg: I don't know where/if it's written down somewhere. But I think the name is a bit strange anyway, having `Actually` as part of the name doesn't add anything.

JinShil:
> The sizes of the D and the C++ classes need to match, as far as I know.

None of the private fields are declared in `aggregate.h::ClassDeclaration`.  Please be more specific, I'm getting confused.  What precisely needs to be done?

jacob-carlborg:
> None of the private fields are declared in aggregate.h::ClassDeclaration

It was only one before, `inuse`, this new field, which was recently added. Which I think needs to be added as well. Currently, if the C++ code tries to access the `isabstract` field it will actually get the `inuse` field.

> Please be more specific, I'm getting confused. What precisely needs to be done?

Please add the `inuse` and the `isActuallyAnonymous` fields to the C++ header `aggregate.h::ClassDeclaration`.

JinShil:
> Please add the inuse and the isActuallyAnonymous fields to the C++ header aggregate.h::ClassDeclaration.

Done. Thank you!

jacob-carlborg:
Example:

```d
$ cat foo.d
module foo;

struct Foo
{
    private int a;
    int b;
}

extern (C) Foo* getFoo()
{
    import core.stdc.stdlib;

    auto foo = cast(Foo*) malloc(Foo.sizeof);
    foo.a = 3;
    foo.b = 4;

    return foo;
}
```
```
$ dmd -c -betterC foo.d
```
```c
$ cat main.c
#include <stdio.h>

struct Foo
{
    int b;
};

struct Foo *getFoo();

int main (int i, char** c)
{
    struct Foo *foo = getFoo();
    printf("%d\n", foo->b);
}
```
```
$ clang main.c -o main foo.o
$ ./main
3
```

The above prints 3 and not 4 because the sizes of the structs on the D side and the C side doesn't match. It's the same idea with the D class and the C++ class.

jacob-carlborg:
> Done. Thank you!

The C++ header file looks good now.

jacob-carlborg:
BTW, if you want to, the fields can be `private` on the C++ side as well.

src/ddmd/dclass.d
JinShil: The "Actually" documents the fact that we are lying, because the class has a `__anonclass` identifier.  And remember I had to overload a method called `isAnonymous`, so that name is off limits.

jacob-carlborg: > And remember I had to overload a method called isAnonymous, so that name is off limits.

That's why it's common to name the field either `_isAnonymous` or `isAnonymous_`. I prefer the latter but I think the former is more common in Phobos.

JinShil:
> BTW, if you want to, the fields can be private on the C++ side as well.

I'm don't want to refactor anything; I just want to fix the regression.  The others fields are `public` so the new fields are `public`.  If we need to be specific with access modifiers for the C++ interface, I think it should be done in a different PR.

jacob-carlborg:
> I'm don't want to refactor anything; I just want to fix the regression

Fair enough.

jacob-carlborg:
LGTM 👌.

src/ddmd/dclass.d
JinShil: >  That's why it's common to name the field either _isAnonymous or isAnonymous_

I would prefer something like that as well, but that doesn't seem to be the convention in DMD.

jacob-carlborg: No, but in the rest of the D community, at least in the core projects, the ones in the `dlang` GitHub organization. Keep in mind that DMD is quite an old code base written in a very old style, C with classes. We're trying to improve that and in my opinion any new code should follow the standard D conventions.

JinShil:
The only thing preventing me from pulling this is I'm not sure if it should target master or stable.  The PR that introduced this regression was only merged yesterday, so I'm guessing it needs to go to master.  Please advise.

jacob-carlborg:
> The only thing preventing me from pulling this is I'm not sure if it should target master or stable. The PR that introduced this regression was only merged yesterday, so I'm guessing it needs to go to master. Please advise.

@MartinNowak ^

JinShil:
It must be master.  I don't see the changes in stable yet.  auto-merge toggling on.

MartinNowak:
> ... but it can be ClassKind.d and anonymous at the same time, right?

In fact anonymous classes always seem to be ClassKind.d, at least I didn't manage to produce a non-D anon class. But it's better to make this explicit, instead of relying on the fact that sth. cannot be expressed in the language's grammar.

MartinNowak:
> The PR that introduced this regression was only merged yesterday, so I'm guessing it needs to go to master.

Hence `[Reg 2.078]` which is the next upcoming release, it means the fix needs to go into 2.078.0, which will be branched off master on Dec 15 2017, and gets released on Jan 1 2018.
Also see https://calendar.google.com/calendar/embed?src=ut29n9vq9vu3ad4kbaabeed1d8%40group.calendar.google.com.

jacob-carlborg:
> In fact anonymous classes always seem to be ClassKind.d, at least I didn't manage to produce a non-D anon class.

Yes, that's correct. It was originally put in the enum to enforce that a class in the AST should not be anonymous and Objective-C/C++ at the same time. Which was possible before and is now possible again.

MartinNowak:
> Which was possible before and is now possible again.

Well it's still not possible in the language and the code the more self-explaining now.

jacob-carlborg:
> the code the more self-explaining now

I disagree. Ideally the API should force you do to the correct things as much as possible. In this case none of the solutions are prefect.

MartinNowak:
Could easily add an assertion that anonymous implies extern(D) atm.

src/ddmd/dsymbol.d
```diff
@@ -269,7 +269,7 @@ extern (C++) class Dsymbol : RootObject
         return false;
     }
 
-    final bool isAnonymous()
+    bool isAnonymous()
```
ibuclaw: So you made this function virtual, but didn't reflect it in c++ headers?

jacob-carlborg: I thought I had the C++ headers covered but apparently not.

JinShil: See https://github.com/dlang/dmd/pull/7385

PR #7379, opened 2017-11-30
disable OSX-32 self-hosting test on Travis-CI

MartinNowak:
- stop testing OSX-32 self-hosting on Travis-CI
- target isn't too well maintained
- osx-32 binaries haven't been reproducible and we only ship
  64-bit dmd binaries, so testing self-hosting on OSX-32 is not
  necessary
- there is a new self-hosted OSX-32-only regression that produces
  incorrect binaries, fixing this would be laborious and fruitless
  (see e.g. https://travis-ci.org/dlang/dmd/jobs/309247634)
- we still have the auto-tester to test OSX-32 cross-compilation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,9 +24,6 @@ matrix:
   include:
     - os: osx
       d: dmd
-      env: MODEL=32
-    - os: osx
-      d: dmd
       env: MODEL=64
   exclude:
     # gdc doesn't neither comes w/ multilib support nor picks up system-wide -lstdc++ lgcc_s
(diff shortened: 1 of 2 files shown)


PetarKirov (approved):

wilzbach (approved):
I absolutely agree. Thanks for pushing this!
There is also the aspect of Travis running out of boxes for macOS testing:

https://blog.travis-ci.com/2017-09-22-macos-update

> we still have the auto-tester to test OSX-32 cross-compilation

FWIW testing something that won't be used, is a bit of a waste of resources / energy. Is there any good reason for keeping this test setup? AFAICT Apple is working on phasing out 32-bit entirely:

> Apple adds that High Sierra, launching this fall, will be the last version of macOS to support 32-bit apps “without compromise.” [1]

CC @braddr

[1] https://www.google.de/amp/s/www.cultofmac.com/490244/high-sierra-will-last-macos-release-support-32-bit-apps/amp/

MartinNowak:
From what I hear there are still a few people writing 32-bit AU/VST plugins for older 32-bit hosts. Since we have the codegen and MACH-O support, it's easy enough to keep that support for the while being, without commit to much effort in the long-run.
*This* PR was mainly about stopping to test dmd as 32-bit app on OSX.

WalterBright:
> Apple adds that High Sierra, launching this fall, will be the last version of macOS to support 32-bit apps “without compromise.” [1]

I interpret that as we should still support generating 32 bit apps for the time being.

PR #7381, opened 2017-11-30
fix Travis-CI tests

MartinNowak:
- run gdb tests with ASLR
- turning off ASLR is not permitted is some test contexts (e.g. Travis-CI)


MartinNowak:
Mmh, seems like we randomly loose PTRACE ability on the test runners.

> Error in sourced command file: During startup program exited normally.

Everything else (build-id, worker-type) seems to be identical.

PR #7382, opened 2017-12-01
Add a cross-platform Makefile

wilzbach:
This will include the respective Makefile depending on the users's OS.

This is for all the people out there who are too lazy to type `make -f posix.mak` all the time and of course all the people who are new to the D build chain and who can now simply type the familiar `make` without needing to know anything about `posix.mak`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,32 @@
+# Includes the respective Makefile depending on the OS
+# Have a look at either posix.mak, win32.mak or win64.mak for a list of available commands.
+
+include ./src/osmodel.mak
+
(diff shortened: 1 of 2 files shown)


WalterBright:
Some issues with this:

1. Two identical files in different directories looks bad. Any time saved is lost in "which makefile should I use?" when confronted with 5 makefiles.

2. Won't work on Windows, because the Digital Mars make.exe is used which does not support the features used by the Makefile.

3. A readme with "type: make -f posix.mak" should do just fine. People who won't read the readme will not be successful building dmd anyway, and would be best off downloading prebuilt binaries.

PR #7384, opened 2017-12-01
Fix .di function literal using `body` instead of `do`

ntrel:
Follow up from #7347.
Use do instead of body in generated header files.
Adds changelog item.

### DIFF
--- a/changelog/body.dd
+++ b/changelog/body.dd
@@ -1,0 +1,6 @@
+Generate header files using `do` instead of `body` as per DIP1003
+
+Support for
+[DIP1003](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1003.md)
+was added in release 2.075.0. Use of `body` in an error message and
+header file generation has been fixed in this release.
--- a/test/compilable/extra-files/header1.d
+++ b/test/compilable/extra-files/header1.d
@@ -10,6 +10,8 @@ static assert(true, "message");
 
 alias double mydbl;
 
+alias fl = function () in {} body {};
+
 int testmain()
 in
 {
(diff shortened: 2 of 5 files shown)


PetarKirov (approved):

test/compilable/extra-files/header1.d
```diff
@@ -10,6 +10,8 @@ static assert(true, "message");
 
 alias double mydbl;
 
+alias fl = function () in {} body {};
```
wilzbach: How about adding a function that actually uses `do` to the testsuite as well? (Consistency ...)

wilzbach (approved):
LGTM too

changelog/body.dd
```diff
@@ -0,0 +1,6 @@
+Generate header files using `do` instead of `body` as per DIP1003
+
+Support for
```
wilzbach: Mind the errors about trailing whitespace:

```
Searching for trailing whitespace
./changelog/2.078.0_pre.dd:33:Support for 
./changelog/2.078.0_pre.dd:34:[DIP1003](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1003.md) 
./changelog/2.078.0_pre.dd:35:was added in release 2.075.0. Use of `body` in an error message and 
```

test/compilable/extra-files/header1.d
ntrel: Now fixed.

changelog/body.dd
ntrel: Now fixed.

MetaLang:
Thanks @ntrel

PR #7385, opened 2017-12-02
Fix C++ Interface after PR #7378

JinShil:
See @ibuclaw's [comment](https://github.com/dlang/dmd/pull/7378#discussion_r154415087)


src/ddmd/aggregate.h
```diff
@@ -302,6 +302,7 @@ class ClassDeclaration : public AggregateDeclaration
     #define OFFSET_RUNTIME 0x76543210
     #define OFFSET_FWDREF 0x76543211
     virtual bool isBaseOf(ClassDeclaration *cd, int *poffset);
+    virtual bool isAnonymous();
```
WalterBright: Don't need to mark it `virtual` if it overrides a virtual function.

PR #7386, opened 2017-12-02
move assert() code past end of function

WalterBright:
Per Nicholas Wilson's suggestion:

http://www.digitalmars.com/d/archives/digitalmars/D/Thoughts_about_D_308441.html#N308716

### DIFF
--- a/src/ddmd/backend/blockopt.c
+++ b/src/ddmd/backend/blockopt.c
@@ -50,7 +50,8 @@ STATIC void block_check();
 STATIC void brtailrecursion();
 STATIC elem * assignparams(elem **pe,int *psi,elem **pe2);
 STATIC void emptyloops();
-int el_anyframeptr(elem *e);
+STATIC int el_anyframeptr(elem *e);
+STATIC void blassertsplit();
 
 unsigned numblks;       // number of basic blocks in current function
 block *startblock;      /* beginning block of function                  */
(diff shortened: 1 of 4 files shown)


WalterBright:
For code:
```
void test(int i) {
    assert(i, "message");
}
```
the following is now generated:
```
_D4test4testFiZv:
                sub     ESP,01Ch
                mov     018h[ESP],EAX
                cmp     dword ptr 018h[ESP],0
                je      L12
                add     ESP,01Ch
                ret
L12:            mov     ECX,offset FLAT:___a6_746573742e64
                mov     EAX,6
                mov     EDX,offset FLAT:___a7_6d657373616765
                mov     dword ptr 010h[ESP],5
                mov     EBX,7
                mov     0Ch[ESP],ECX
                mov     8[ESP],EAX
                mov     4[ESP],EDX
                mov     [ESP],EBX
                call    near ptr __d_assert_msg
```

thewilsonator:
I can't comment on the code, but I will note that this doesn't apply only to asserts, but to all slow path noreturn branches e.g. conditional branches that terminate in a `throw` or calls to functions that are noreturn (e.g. enforce) and asserts.

WalterBright:
I know, but one thing at a time!

thewilsonator:
Fair enough.

ibuclaw:
This caused a regression: https://issues.dlang.org/show_bug.cgi?id=19550

thewilsonator:
I haven't done much much complexity analysis, but that looks `O(pretty bad)` to me.

PR #7135, opened 2017-09-14
ddmd -> dmd transition

wilzbach:
Let's see what the missing bits on Windows are.
CC @RazvanN7 

Required steps

- [x] https://github.com/dlang/dmd/pull/6562
- [x] https://github.com/CyberShadow/ae/commit/13bc434ea12c90883f524818510f35970230becb
- [x] https://github.com/dlang/dmd/pull/6585
- [x] https://github.com/dlang/dmd/pull/6586
- [x] https://github.com/dlang/dmd/pull/6593
- [x] https://github.com/dlang/dmd/pull/6620
- [x] https://github.com/dlang/ci/pull/68
- [x] https://github.com/dlang/dlang.org/pull/1903
- [x] https://github.com/braddr/at-client/pull/4
- [x] https://github.com/dlang/druntime/pull/1978
- [x] https://github.com/dlang/phobos/pull/5892
- [x] https://github.com/braddr/at-client/pull/5
- [x] https://github.com/dlang/druntime/pull/1979
- [x] https://github.com/dlang/phobos/pull/5895
- [x] https://github.com/dlang/dlang.org/pull/1971

PR #6562, opened 2017-02-23
Generate all .o .deps and binaries to generated/

RazvanN7:
I created a directory src/ddmd where I moved all the files in src/, except makefiles. Other than that I modified optabgen.d to generate sources to src/ddmd/backend and I also moved tk.c to backend since it was the only .c file in src/ddmd.

I also modified the makefile to have all generated files (except .c .d and .h files - source code generated files) in a directory called generated/ which is on the root level (same level as src). After creating dmd, the executable is copied to src/dmd so that other tools aren't affected by this PR. 

Now all the other tools can be updated to use the dmd from the generated/ folder whithout causing a major stall. After all the scripts are updated to use generated/dmd/ I will issue a new PR which renames ddmd to dmd and no longer generates dmd in the src folder. This way things will get smoothly in the right direction.

This is a WIP as I only applied the modifications to the posix makefile. 

This is a duplicate of #6489 since it was more time consuming to apply these changes on that branch.

PR #7387, opened 2017-12-02
Make C++ warnings errors (i.e. -Werror) in posix.mak

JinShil:
Attempt at shepherding #7358 across the finish line.

### DIFF
--- a/src/ddmd/backend/cod1.c
+++ b/src/ddmd/backend/cod1.c
@@ -4244,17 +4244,17 @@ void pushParams(CodeBuilder& cdb,elem *e,unsigned stackalign)
         if (I32 && szb == 10)           // special case for long double constants
         {
             assert(sz == 12);
-            targ_int value = ((unsigned short *)&e->EV.Vldouble)[4];
+            targ_int value = e->EV.Vushort8[4]; // pick upper 2 bytes of Vldouble
             stackpush += sz;
             cdb.genadjesp(sz);
-            for (int i = 2; i >= 0; i--)
+            for (int i = 0; i < 3; ++i)
             {
                 unsigned reg;
                 if (reghasvalue(allregs, value, &reg))
                     cdb.gen1(0x50 + reg);           // PUSH reg
                 else
                     cdb.genc2(0x68,0,value);        // PUSH value
-                value = ((unsigned *)&e->EV.Vldouble)[i - 1];
+                value = e->EV.Vulong4[i ^ 1];       // treat Vldouble as 2 element array of 32 bit unsigned
             }
             freenode(e);
             return;
@@ -4278,7 +4278,7 @@ void pushParams(CodeBuilder& cdb,elem *e,unsigned stackalign)
 
         stackpush += sz;
         cdb.genadjesp(sz);
-        targ_uns *pi = (targ_uns *) &e->EV.Vdouble;
+        targ_uns *pi = &e->EV.Vuns;     // point to start of Vdouble
         targ_ushort *ps = (targ_ushort *) pi;
         targ_ullong *pl = (targ_ullong *)pi;
         i /= regsize;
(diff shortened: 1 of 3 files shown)


jacob-carlborg:
Aren't there any complains about the C++ files in the backend are using the `.c` extension?

ibuclaw:
> Aren't there any complains about the C++ files in the backend are using the .c extension?

That's what the testsuite will tell us.

ibuclaw:
@JinShil - Forgetting some patches?

https://github.com/dlang/dmd/pull/7358/files#diff-9b5e2f2ac206ebe201f44f8e316514bcR4247

JinShil:
> @JinShil - Forgetting some patches?

Thanks!

ibuclaw (approved):
ok, but ping @WalterBright on backend stuff.

MartinNowak:
This has broken the release builds with an older clang on OSX.

MartinNowak:
See https://github.com/dlang/dmd/pull/7430

JinShil:
> This has broken the release builds with an older clang on OSX.

Why wasn't it caught by the auto-tester?

MartinNowak:
Because that is running a newer clang version, apparently the compiler was fixed to not be so picky.

jacob-carlborg:
> Apple LLVM version 5.1 (clang-503.0.40) (based on LLVM 3.4svn)
Target: x86_64-apple-darwin12.2.0
Thread model: posix

I suggest that the compiler is updated. Clang 5 and LLVM 3.4 is pretty old. Latest version of Xcode is 9.2.

jacob-carlborg:
> This has broken the release builds with an older clang on OSX.

The same compiler that is running the tests should be building the releases. Ideally on the same computer.

wilzbach:
> I suggest that the compiler is updated. Clang 5 and LLVM 3.4 is pretty old. Latest version of Xcode is 9.2.

Good luck!
We are working since February to upgrade the auto-tester, s.t. we can finally transition from ddmd -> dmd (see e.g. https://github.com/dlang/dmd/pull/7135 or https://github.com/dlang/dmd/pull/6562).

jacob-carlborg:
How is "ddmd -> dmd" related to upgrading Clang?

wilzbach:
> How is "ddmd -> dmd" related to upgrading Clang?

It's just an example on how difficult it is to upgrade things on auto-tester.

jacob-carlborg:
Ok. Is Brad the only one with access?

braddr:
I'm the only one with access to most of the machines.  They're typically behind firewalls and on networks I don't have liberty to give direct access to.

The Feb to Now transition time for ddmd -> dmd can't be reasonably placed on me like that.  *Maybe* a month of the total time can, but the rest, sorry, no.  The rest was largely waiting for the build system and branches to be in sync enough that the changes to the build system wouldn't break every build.  I won't pull change just because they've been submitted.  They kinda have to work too.

As to the osx versions, there's three of them in the fleet, each with a different version.  I'd be happy to upgrade the oldest ones if we're deprecating support for those altogether.  The versions in place:

Darwin 13.4.0, Apple LLVM version 6.0 (clang-600.0.57) (based on LLVM 3.5svn)
Darwin 15.6.0, Apple LLVM version 8.0.0 (clang-800.0.42.1)
Darwin 16.7.0, Apple LLVM version 9.0.0 (clang-900.0.39.2)

The difficulty is that we've got a large set of platforms, several branches, and lots of project in flight and aligning them all correctly doesn't happen as fast as anyone would like.

Another great example is upgrading the freebsd supported platform from 8.x to 10 and/or 11.  I'd love to, except that the code base doesn't pass testing on them.  I even setup the auto-tester to build and test master against those platforms.  No one touched that project beyond a tiny bit of initial interest.

jacob-carlborg:
> I'm the only one with access to most of the machines. They're typically behind firewalls and on networks I don't have liberty to give direct access to.

Fair enough. Then it's also understandable why it might take longer, if only one person is responsible.

> Another great example is upgrading the freebsd supported platform from 8.x to 10 and/or 11. I'd love to, except that the code base doesn't pass testing on them. I even setup the auto-tester to build and test master against those platforms. No one touched that project beyond a tiny bit of initial interest.

I see. I would hope that upgrading just the compiler is simpler than upgrading the whole operating system.

PR #6464, opened 2017-01-17
Output DWARF informations for the enum declarations

LemonBoy:
As noted by @mathias-lang-sociomantic in [this ticket](https://issues.dlang.org/show_bug.cgi?id=15630) `gdb` wouldn't show the value of the enum-typed parameter.
Further investigation showed that this was never implemented or has been thrown away at some point so I took the chance and implemented it and also extended the functionality a bit: it now works with every enum with an integral base type, both signed and unsigned, ranging from {u,}byte to {u,}long.

It hasn't been tested on x86, only x64.

Some technical infos: the `SENforward` flag is now ignored since it was only set once in the whole codebase (in `type_enum` and never cleared) and is not needed for our purpose (I can remove the whole `#if 0` block and/or the flag itself if needed). We now emit a `DW_AT_type` for the enumeration type like other cooler compilers do (I've cross-checked this with `LDC`, we're pretty cool).

PR #7388, opened 2017-12-02
Fix Issue 15630 - gdb: <incomplete type> for enums values

JinShil:
Attempted revival of #6464

Special thanks to @LemonBoy 

cc @ibuclaw

### DIFF
--- a/src/dmd/backend/dwarf.c
+++ b/src/dmd/backend/dwarf.c
@@ -2731,6 +2731,8 @@ unsigned dwarf_typidx(type *t)
                 1,                      // child (the subrange type)
                 DW_AT_name,             DW_FORM_string,
                 DW_AT_byte_size,        DW_FORM_data1,
+                DW_AT_type,             DW_FORM_ref4,
+                DW_AT_enum_class,       DW_FORM_flag,
                 0,                      0,
             };
             static unsigned char abbrevTypeEnumMember[] =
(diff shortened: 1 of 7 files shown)


src/ddmd/toctype.d
```diff
@@ -170,14 +170,24 @@ public:
         //printf("TypeEnum::toCtype() '%s'\n", t.sym.toChars());
         if (t.mod == 0)
         {
+            auto baseType = t.sym.memtype.toBasetype();
             if (!t.sym.memtype)
```
ibuclaw: Not a good place to dereference memtype, considering this condition.

ibuclaw (dismissed):
And as per @MartinNowak comments, needs test in testsuite.

src/ddmd/toctype.d
JinShil: Ok, how about now?

JinShil: > And as per @MartinNowak comments, needs test in testsuite.

Added.

JinShil:
Fails for `enum E : long` on 32-bit platforms.  I don't know why yet.

JinShil:
I'm not willing to pursue this anymore.  Closing.

JinShil:
Filed an issue for the 32-bit backend bug at https://issues.dlang.org/show_bug.cgi?id=18191

quickfur:
ping

Any progress on this?  This is making dmd-generated executable rather debugger-unfriendly, as enums cannot be inspected (gdb just prints `<incomplete type>` for the enum value.

PR #7389, opened 2017-12-02
Make deprecation warnings compile errors

wilzbach:
This is already the way we build Phobos + Tools.
See also: https://github.com/dlang/phobos/pull/5546
The idea is that whenever sth. gets deprecated in Druntime or Phobos a replacement needs to exists. By always building with `-de`, we can enforce that the replacement really exists and force the person who submits the deprecation to look after the fallout.

CC @ZombineDev @CyberShadow


CyberShadow:
Looks like this needs some code fixes first as well.

wilzbach:
> Looks like this needs some code fixes first as well.

Fixed.

PR #7391, opened 2017-12-03
use UD2 instead of HLT instruction

WalterBright:
Adam Ruppe suggested it:

http://www.digitalmars.com/d/archives/digitalmars/D/Thoughts_about_D_308441.html#N308711

```
hlt kinda bothers me because it actually has a meaning. You're 
just abusing the fact that it is privileged so it traps on 
operating systems, but on bare metal, it just pauses until the 
next interrupt.
int 3, on the other hand, is explicitly for debugging - which is 
what we want asserts to be.
```

### DIFF
--- a/src/ddmd/backend/cod2.c
+++ b/src/ddmd/backend/cod2.c
@@ -7,6 +7,7 @@
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
  * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/ddmd/backend/cod2.c, backend/cod2.c)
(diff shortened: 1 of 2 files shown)


rainers:
The downside is that int3 also has a meaning ;-) Especially you can step over it in a debugger, which is not what you want.
On Win32, `hlt` can be handled by `catch(Throwable)` without the debugger popping up, while `int3` always stops the process in the debugger.

ZombineDev:
What about int 3, followed by [UD2](http://x86.renejeschke.de/html/file_module_x86_id_318.html)?

rainers:
> How about int 3, followed by UD2?

Just `UD2` should be fine, too. It is likely to be treated similar to `hlt` in a debugger.

ibuclaw:
More of an FYI, gdc emits ud2.  Not sure if that would influence any decision here.

WalterBright:
UD2 does sound like a better choice.

rainers (approved):

PR #7392, opened 2017-12-03
'throw' makes a block BCexit

WalterBright:
As suggested by Nicholas Wilson.

### DIFF
--- a/src/ddmd/backend/cgcod.c
+++ b/src/ddmd/backend/cgcod.c
@@ -149,9 +149,9 @@ void codgen(Symbol *sfunc)
     Alloca.init();
     anyiasm = 0;
 
-    if (config.exe & (EX_LINUX | EX_OSX | EX_OSX64 | EX_FREEBSD | EX_OPENBSD | EX_SOLARIS))
+    if (config.ehmethod == EH_DWARF)
     {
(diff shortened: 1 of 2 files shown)


WalterBright:
Later on, move the BCexit blocks past the end of the function.

WalterBright:
This is a mystery. test/runnable/testaa2.d is segfaulting under 32 bit Linux. But it runs fine on my machine (Ubuntu). Diffing testaa2.o with an without this change yields just what I expect - the generated code is exactly the same, except for:
```
  pop EBX
  pop EBP
  ret
  add [EAX],AL
```
appears after the call to `_d_throwdwarf`, which is just what I expected.

The only thing I can think of is that `_d_throwdwarf` is returning, but the source code to it puts `assert(0)` at all return points. Anyone have any ideas?

WalterBright:
Apparently some dwarf unwinders need the epilog to be present after a throw.

WalterBright:
@MartinNowak what does this error mean?
```
timelimit -t 90  generated/linux/release/64/unittest/test_runner std.xml
0.005s PASS release64 std.xml
posix.mak:379: recipe for target 'unittest/std/xml.run' failed
```

andralex (approved):
noice

PR #7393, opened 2017-12-03
autodetect Visual Studio and linker options

rainers:
This combines detection of latest VS and Windows SDK from Visual D and the DMD installer. All that is needed specific for -m64/-m32mscoff in sc.ini is

```
[Environment64]
LIB=%@P%\..\lib64;%LIB%
DFLAGS=%DFLAGS% -L/OPT:NOICF

[Environment32mscoff]
LIB=%@P%\..\lib32mscoff;%LIB%
```

Releated bugzilla issues:

https://issues.dlang.org/show_bug.cgi?id=16688
https://issues.dlang.org/show_bug.cgi?id=17999
https://issues.dlang.org/show_bug.cgi?id=14849

### DIFF
--- a/src/ddmd/link.d
+++ b/src/ddmd/link.d
@@ -248,54 +248,10 @@ public int runLINK()
                 cmdbuf.writeByte(' ');
                 cmdbuf.writestring(global.params.linkswitches[i]);
             }
-            /* Append the path to the VC lib files, and then the SDK lib files
-             */
-            const(char)* vcinstalldir = getenv("VCINSTALLDIR");
-            if (vcinstalldir)
-            {
-                cmdbuf.writestring(" /LIBPATH:\"");
-                cmdbuf.writestring(vcinstalldir);
-                if (global.params.is64bit)
-                    cmdbuf.writestring("\\lib\\amd64\"");
-                else
-                    cmdbuf.writestring("\\lib\"");
-            }
-            const(char)* windowssdkdir = getenv("WindowsSdkDir");
-            if (windowssdkdir)
-            {
-                cmdbuf.writestring(" /LIBPATH:\"");
-                cmdbuf.writestring(windowssdkdir);
-                if (global.params.is64bit)
-                    cmdbuf.writestring("\\lib\\x64\"");
-                else
-                    cmdbuf.writestring("\\lib\"");
-            }
-            cmdbuf.writeByte(' ');
-            const(char)* lflags;
-            if (detectVS14(cmdbuf.peekString()))
-            {
-                lflags = getenv("LFLAGS_VS14");
-                if (!lflags)
-                    lflags = "legacy_stdio_definitions.lib";
-                // environment variables UniversalCRTSdkDir and UCRTVersion set
-                // when running vcvarsall.bat x64
-                if (const(char)* UniversalCRTSdkDir = getenv("UniversalCRTSdkDir"))
-                    if (const(char)* UCRTVersion = getenv("UCRTVersion"))
-                    {
-                        cmdbuf.writestring(" /LIBPATH:\"");
-                        cmdbuf.writestring(UniversalCRTSdkDir);
-                        cmdbuf.writestring("\\lib\\");
-                        cmdbuf.writestring(UCRTVersion);
-                        if (global.params.is64bit)
-                            cmdbuf.writestring("\\ucrt\\x64\"");
-                        else
-                            cmdbuf.writestring("\\ucrt\\x86\"");
-                    }
-            }
-            else
-            {
-                lflags = getenv("LFLAGS_VS12");
-            }
+
+            VSOptions vsopt;
+            vsopt.initialize();
+            const(char)* lflags = vsopt.linkOptions(global.params.is64bit);
             if (lflags)
             {
                 cmdbuf.writeByte(' ');
@@ -319,20 +275,8 @@ public int runLINK()
             if (!linkcmd)
                 linkcmd = getenv("LINKCMD"); // backward compatible
             if (!linkcmd)
-            {
-                if (vcinstalldir)
-                {
-                    OutBuffer linkcmdbuf;
-                    linkcmdbuf.writestring(vcinstalldir);
-                    if (global.params.is64bit)
-                        linkcmdbuf.writestring("\\bin\\amd64\\link");
-                    else
-                        linkcmdbuf.writestring("\\bin\\link");
-                    linkcmd = linkcmdbuf.extractString();
-                }
-                else
-                    linkcmd = "optlink";
-            }
+                linkcmd = vsopt.linkerPath(global.params.is64bit);
+
             int status = executecmd(linkcmd, p);
             if (lnkfilename)
             {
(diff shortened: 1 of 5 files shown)


src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+        {
+            HRESULT hr;
+        }
+
+        const(char)* GetString(const(char)* szName, const(char)* def = null)
+        {
+            if(!key)
+                return def;
+
+            char[260] buf;
```
WalterBright: append `= void`

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+        void Close()
+        {
+            if(key)
+            {
+                RegCloseKey(key);
+                key = null;
+            }
+        }
+
+        void Create(HKEY root, wstring keyname, bool x64hive = false)
```
WalterBright: I generally dislike using default arguments unless doing it for backwards compatibility. Since this is new code, please use explicit arguments.

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+        {
+            HRESULT hr;
+        }
+
+        const(char)* GetString(const(char)* szName, const(char)* def = null)
```
WalterBright: ditto

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+        {
+            if(!key)
+                return def;
+
+            char[260] buf;
+            DWORD cnt = 260 * char.sizeof;
```
WalterBright: replace `260` with `buf.length`

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+            if (h == INVALID_HANDLE_VALUE)
+                return null;
+
+            const(char)* res = null;
+            do
+            {
+                if (fileinfo.cFileName[0] >= '1' && fileinfo.cFileName[0] <= '9')
+                    if (res is null || strcmp(res, fileinfo.cFileName.ptr) < 0)
+                        if (FileName.exists(FileName.combine(FileName.combine(baseDir, fileinfo.cFileName.ptr), testfile)))
+                            res = strdup(fileinfo.cFileName.ptr);
```
WalterBright: being in a loop, the `res` seems to be leaking memory upon reassignment

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+            }
+            while(FindNextFileA(h, &fileinfo));
+            FindClose(h);
+
+            return res;
```
WalterBright: returning `malloc`ed memory should be documented

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+            const(char)* res = null;
+            do
+            {
+                if (fileinfo.cFileName[0] >= '1' && fileinfo.cFileName[0] <= '9')
+                    if (res is null || strcmp(res, fileinfo.cFileName.ptr) < 0)
+                        if (FileName.exists(FileName.combine(FileName.combine(baseDir, fileinfo.cFileName.ptr), testfile)))
+                            res = strdup(fileinfo.cFileName.ptr);
+            }
+            while(FindNextFileA(h, &fileinfo));
+            FindClose(h);
```
WalterBright: Return value is ignored - should check it for error

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+        }
+
+        void Create(HKEY root, wstring keyname, bool x64hive = false)
+        {
+            HRESULT hr;
+        }
+
+        const(char)* GetString(const(char)* szName, const(char)* def = null)
+        {
+            if(!key)
```
WalterBright: Sometimes this PR uses `if(` and sometimes `if (`. Please use the latter style.

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+    {
+        pragma(lib, "advapi32.lib");
+
+        this(HKEY root, const(char)* szRegPath, bool x64hive = false)
+        {
+            version(Win64)
+            {
+                if (!x64hive)
+                {
+                    size_t len = strlen(szRegPath);
```
WalterBright: declare as `const`

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+                    assert(len > 9 && szRegPath[0..9] == r"SOFTWARE\");
+                    char* npath = cast(char*)mem.xmalloc(len + 13);
+                    strcpy(npath, r"SOFTWARE\WOW6432Node\");
+                    strcpy(npath + 21, szRegPath + 9);
```
WalterBright: Since in both calls to `strcpy` the lengths are known, use `memcpy` instead.

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+                return this;
+            return PathBuilder(FileName.combine(p, name));
+        }
+    }
+
+    struct RegKey
+    {
+        pragma(lib, "advapi32.lib");
+
+        this(HKEY root, const(char)* szRegPath, bool x64hive = false)
```
WalterBright: no default arguments

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+        ~this()
+        {
+            Close();
+        }
+
+        void Close()
+        {
+            if(key)
+            {
+                RegCloseKey(key);
```
WalterBright: Check for error return

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+                key = null;
+            }
+        }
+
+        void Create(HKEY root, wstring keyname, bool x64hive = false)
+        {
+            HRESULT hr;
```
WalterBright: Declaration that is unused?

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+        }
+
+    }
+
+    struct PathBuilder
+    {
+        const(char)* p;
+        alias p this;
+
+        PathBuilder opDiv(const(char)* name)
```
WalterBright: Please don't do non-arithmetic operator overloading.

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+        {
+            if (!name)
```
WalterBright: The uses I see of PathBuilder suggest that `name` is never null, so this struct is kinda pointless. In fact, `Filename.combine` should do the right thing with a `null` second argument anyway. (I checked and it does not, so that should be fixed there, not adding a wrapper here.)

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+
+        const(char)* getSDKLibPath(bool x64)
+        {
+            if(WindowsSdkDir)
+            {
+                const(char)* arch = x64 ? "x64" : "x86";
+                auto sdk = PathBuilder(WindowsSdkDir) / "lib";
+                if(FileName.exists(sdk / WindowsSdkVersion / "um" / arch / "kernel32.lib")) // SDK 10.0
+                    return sdk / WindowsSdkVersion / "um" / arch;
+                else if(FileName.exists(sdk / r"win8\um" / arch / "kernel32.lib")) // SDK 8.0
```
WalterBright: I'd rather see `a / b / c` added as a function to `root.filename`, so it can be used elsewhere. It should use variadic parameters to list them, or an array literal, or even 2,3,4 parameter overloads.

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+            if (auto p = getVCDir(VCDir.Bin, false)) // prefer 32-bit linker in case of cross-compilation
+            {
+                OutBuffer cmdbuf;
+                cmdbuf.writestring(p);
+                cmdbuf.writestring(r"\link.exe");
+                if (VSInstallDir)
+                {
+                    // debug info needs DLLs Common7\IDE for most linker versions
+                    const char* path = getenv("PATH");
+                    const char* npath = FileName.combine(VSInstallDir, r"Common7\IDE");
```
WalterBright: `path` and `npath` appear to never be used?

src/ddmd/link.d
```diff
@@ -1039,4 +983,384 @@ version (Windows)
...
+        const(char)* WindowsSdkDir;
+        const(char)* WindowsSdkVersion;
+        const(char)* UCRTSdkDir;
+        const(char)* UCRTVersion;
+        const(char)* VSInstallDir;
+        const(char)* VisualStudioVersion;
+        const(char)* VCInstallDir;
+        const(char)* VCToolsInstallDir; // used by VS 2017
+
+        void initialize()
```
WalterBright: In general, all functions need Ddoc comments.

rainers:
@WalterBright Thanks for review, I've addressed most comments.

src/ddmd/link.d
rainers: Pretty unlikely to fail, and even if it does, should it abort a build?

src/ddmd/link.d
rainers: I introduced the PathBuilder struct because FileName.combine cannot be used with UFCS, but creates unreadable code if nesting multiple calls.
Using variadic parameters instead is almost as good, though, so I replaced PathBuilder with FileName.buildPath now.

src/ddmd/link.d
rainers: Ooops, missing PATH modification.

src/ddmd/link.d
```diff
@@ -981,62 +925,409 @@ public int runProgram()
...
+                if (fileinfo.cFileName[0] >= '1' && fileinfo.cFileName[0] <= '9')
+                    if (res is null || strcmp(res, fileinfo.cFileName.ptr) < 0)
+                        if (FileName.exists(FileName.buildPath(baseDir, fileinfo.cFileName.ptr, testfile)))
+                        {
+                            const len = strlen(fileinfo.cFileName.ptr) + 1;
+                            res = cast(char*) memcpy(mem.xrealloc(res, len), fileinfo.cFileName.ptr, len);
+                        }
+            }
+            while(FindNextFileA(h, &fileinfo));
+            FindClose(h);
```
WalterBright: `FindClose()` returns 0 on failure. Please check it. https://msdn.microsoft.com/en-us/library/windows/desktop/aa364413(v=vs.85).aspx

src/ddmd/link.d
```diff
@@ -981,62 +925,409 @@ public int runProgram()
...
+         *  softwareKeyPath = path below HKLM\SOFTWARE
+         *  valueName       = name of the value to read
+         * Returns:
+         *  the registry value if it exists and has string type
+         */
+        const(char)* GetRegistryString(const(char)* softwareKeyPath, const(char)* valueName)
+        {
+            enum x64hive = false; // VS registry entries always in 32-bit hive
+
+            version(Win64)
```
WalterBright: Wait a minute! Why would the 64 bit build of DMD behave differently here? Shouldn't this be if the memory model of the target is set to 64?

src/ddmd/link.d
```diff
@@ -981,62 +925,409 @@ public int runProgram()
...
+        {
+            detectWindowsSDK();
+            detectUCRT();
+            detectVSInstallDir();
+            detectVCInstallDir();
+            detectVCToolsInstallDir();
+        }
+
+        /**
+         * retreive options to be passed to the Micrsoft linker
```
WalterBright: "retrieve", and "Microsoft"

src/ddmd/link.d
```diff
@@ -981,62 +925,409 @@ public int runProgram()
...
+                // support for old DX SDK installations
+                cmdbuf.writestring(" /LIBPATH:\"");
+                cmdbuf.writestring(p);
+                cmdbuf.writestring(x64 ? `\Lib\x64"` : `\Lib\x86"`);
+            }
+            return cmdbuf.extractString();
+        }
+
+        /**
+         * retreive path to the Micrsoft linker executable
```
WalterBright: "retrieve", and "Microsoft"

WalterBright:
Much better. A couple more issues I commented on. I'm still not happy with the avoidable memory leaks, but it's not a deal breaker here. I do like how you fixed one leak by using realloc, I hadn't thought of that!

src/ddmd/link.d
rainers: I know, but it feels a bit harsh to fail while a sensible result is already obtained. I've changed it nevertheless.

src/ddmd/link.d
rainers: No, the target architecture is not relevant here. The VS registry/Windows SDK keys are only found in the 32-bit registry hive. KEY_WOW64_32KEY doesn't seem good enough, I had to use WOW6432Node explicitly, while a 32-bit process is automatically mapped to that hive.

rainers:
> I'm still not happy with the avoidable memory leaks, but it's not a deal breaker here.

FileName.combine sometimes returns newly allocated memory, sometimes one of the arguments. This makes it necessary to predict what happens if you want to free the memory later. As this is done nowhere else, I opted not to do it here for a few more hundred bytes of leakage.

MartinNowak:
Could you please add a changelog entry mentioning that VC is now detected at link-time instead of at install-time. Should we remove the detection from the installer?

MartinNowak:
Factoring this out into a separate dmd/windows_detect_vs module would make sense IMO.

rainers:
> Could you please add a changelog entry mentioning that VC is now detected at link-time instead of at install-time. 

Will do.

> Should we remove the detection from the installer?

Part of it. We might still want to allow the option of downloading VS if not found. Please note that the "light" runtime still needs the VC 2010 distributables.

PR #7394, opened 2017-12-04
Implement -nomoduleinfo compiler switch

JinShil:


JinShil:
A different idea occurred to me.  Going to try that first.

PR #7398, opened 2017-12-05
Use generated DMD executable for the testsuite

wilzbach:
Trying a better solution instead of https://github.com/dlang/dmd/pull/6870.
Needed for the ddmd -> dmd transition (https://github.com/dlang/dmd/pull/7135)

### DIFF
--- a/posix.mak
+++ b/posix.mak
@@ -21,6 +21,7 @@ clean:
 	$(RM) tags
 
 test:
+	echo $(MODEL)
 	$(QUIET)$(MAKE) -C test -f Makefile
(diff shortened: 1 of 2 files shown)


wilzbach:
Alrighty, I think it's clear that the simple approach won't work here.
I have opened a PR at the auto-tester to explicitly pass the generated `DMD` executable to the Makefile (-> https://github.com/braddr/at-client/pull/7)

For further discussions, I think it's best to continue them here: https://github.com/dlang/dmd/pull/6870

PR #7399, opened 2017-12-05
[Do not merge] Use generated DMD executable for the testsuite [Windows]

wilzbach:
Trying a better solution instead of https://github.com/dlang/dmd/pull/6870.
Needed for the ddmd -> dmd transition (https://github.com/dlang/dmd/pull/7135).


wilzbach:
Ah well, doesn't seem to work either. I will continue at https://github.com/dlang/dmd/pull/6870
Sorry for the noise.

PR #7400, opened 2017-12-06
bug 8858 - Switch to analyze dependencies without printing them

marler8997:
…or to a file.

This is in response to (https://issues.dlang.org/show_bug.cgi?id=8858).  It allows the option `-deps-` which causes the compiler to analyze dependencies without printing them to stderr or to a file.  The typical use case for this would be when a tool like rdmd needs to know all the dependencies but is using the verbose output instead of the dependency specific output.

### DIFF
--- a/src/ddmd/cond.d
+++ b/src/ddmd/cond.d
@@ -923,9 +923,9 @@ extern (C++) bool findCondition(Identifiers* ids, Identifier ident)
 // Helper for printing dependency information
 private void printDepsConditional(Scope* sc, DVCondition condition, const(char)[] depType)
 {
-    if (!global.params.moduleDeps || global.params.moduleDepsFile)
+    if (!global.params.moduleDepsOut || global.params.moduleDepsFile)
         return;
-    OutBuffer* ob = global.params.moduleDeps;
+    OutBuffer* ob = global.params.moduleDepsOut;
     Module imod = sc ? sc.instantiatingModule() : condition.mod;
     if (!imod)
         return;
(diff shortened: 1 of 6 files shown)


andralex:
This seems like an inflation of options. What's wrong with directing deps to `/dev/null`? Is outputting deps a bottleneck?

We really need to stop defining new artifacts that are subtly different versions of existing artifacts. The right way to go is improve the existing artifacts.

marler8997:
Even if output goes to /dev/null (which I don't how you would do that on windows) dmd is still spending cycles generating the output.  I estimate this is a very minor performance savings but I thought I would do what I could to try to improve the performance of rdmd in any way I can.  Weighing the simplicity of this change with a minor performance savings in rdmd seemed like it could be worth it, but I could go either way on this one.  Based on your comments about how this complicates dmd's interface I'll go ahead and close this one.

andralex:
>I estimate this is a very minor performance savings but I thought I would do what I could to try to improve the performance of rdmd in any way I can. 

The one way to go about this is to measure.

marler8997:
For the digger project, on windows (piping output to a file, which is faster than going to the console) it saves about 100 ms out of about 8 seconds.  Around a 1 to 2% savings.

DmitryOlshansky:
This is mostly a problem of Windows console. Piping ti equivalent of
/dev/null should nullify this difference.
пт, 8 дек. 2017 г. в 5:15, Jonathan Marler <notifications@github.com>:

rainers:
> Even if output goes to /dev/null (which I don't how you would do that on windows)

FYI: the corresponding file on Windows is `NUL:`.

marler8997:
Didn't know about `NUL:`.  Piping commands to `NUL:` seems to work but it doesn't seem to work as a filename passed to `-deps`.  I tried `-deps=NUL` and `-deps=NUL:` but I just get `Error writing to file 'NUL:'`.

rainers:
> I tried -deps=NUL and -deps=NUL: but I just get Error writing to file 'NUL:'.

Both work for me with the released dmd, but not if dmd is built with VC.

rainers:
I suspect it's not VC related, but caused by https://github.com/dlang/dmd/pull/7299

PR #7401, opened 2017-12-06
[Trivial] fix C++ header for StaticForeachDeclaration

JohanEngelen:
This fixes the types of functions overridden by `StaticForeachDeclaration`, such that they actually override the base functions.

Can we use C++11 in ddmd's C++ headers? (LDC already requires C++11, @ibuclaw how about GDC?)  Marking such overrides with `override` would help in avoiding these unintended errors in overriding.

### DIFF
--- a/src/ddmd/attrib.h
+++ b/src/ddmd/attrib.h
@@ -199,10 +199,10 @@ public:
     Dsymbols *cache;
 
     Dsymbol *syntaxCopy(Dsymbol *s);
-    bool oneMember(Dsymbol *ps, Identifier *ident);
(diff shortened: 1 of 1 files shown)


WalterBright:
> Can we use C++11 in ddmd's C++ headers? (LDC already requires C++11, @ibuclaw how about GDC?) 

Dmd does not use the C++ headers any more, so I defer to the ldc and gdc teams.

ibuclaw:
Last time I checked gcc codebase was sticking to c++98 only.  Then again frontends may have a little more leeway.

PR #5794, opened 2016-05-18
Issue 16042 - Identifier on template arguments should consider eponymous member lookup

9rnsr:

PR #7404, opened 2017-12-07
Fix Issue 16042 - Identifier on template arguments should consider eponymous member lookup

JinShil:
Revival of #5794

Special thanks to @9rnsr

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -6802,6 +6802,25 @@ extern (C++) final class TypeIdentifier : TypeQualified
 
         Dsymbol scopesym;
         Dsymbol s = sc.search(loc, ident, &scopesym);
+
+        if (s)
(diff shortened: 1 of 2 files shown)


wilzbach:
> Merge upstream

FWIW it's better for the git history if you use `git rebase` instead of `git merge`. There are a couple of articles in the wiki dealing with this (e.g. https://wiki.dlang.org/Starting_as_a_Contributor#Rebasing) and you can always ping us on Slack if you need help.

JinShil:
I just recreated everything, so the git history should now be clean.

JinShil:
FYI: I will be closing this PR in 14 days if noone shows any interest in it.

WalterBright:
I'm a little nervous about what this may break, but I'll take a chance on it.

PR #7405, opened 2017-12-07
Issue 9631 - Error message not using fully qualified name

ntrel:
Fix *cannot implicitly convert expression of type T to T* ambiguous errors.
Add `Type.toAutoQualChars()`.

There are other related errors that could use this fix, this only addresses the first test case in bugzilla.

### DIFF
--- a/src/ddmd/mtype.d
+++ b/src/ddmd/mtype.d
@@ -899,6 +899,7 @@ extern (C++) abstract class Type : RootObject
         return buf.extractString();
     }
 
+    /// ditto
     final char* toPrettyChars(bool QualifyTypes = false)
     {
         OutBuffer buf;
@@ -9172,9 +9173,14 @@ extern (C++) final class Parameter : RootObject
     extern (D) static immutable bool[SR.max + 1][SR.max + 1] covariant = covariantInit();
 }
 
-/***********************************************************
+/**
  * For printing two types with qualification when necessary.
- * Disambiguates the case where toChars() on each type match.
+ * Params:
+ *    t1 = The first type to receive the type name for
+ *    t2 = The second type to receive the type name for
+ * Returns:
+ *    The fully-qualified names of both types if the two type names are not the same,
+ *    or the unqualified names of both types if the two type names are the same.
  */
 const(char*)[2] toAutoQualChars(Type t1, Type t2)
 {
--- a/src/ddmd/mtype.d
+++ b/src/ddmd/mtype.d
@@ -910,18 +910,6 @@ extern (C++) abstract class Type : RootObject
         return buf.extractString();
     }
 
-    static const(char*)[2] toAutoQualChars(Type t1, Type t2)
-    {
-        auto s1 = t1.toChars();
-        auto s2 = t2.toChars();
-        if (strcmp(s1, s2) == 0)
-        {
-            s1 = t1.toPrettyChars(true);
-            s2 = t2.toPrettyChars(true);
-        }
-        return [s1, s2];
-    }
-
     static void _init()
     {
         stringtable._init(14000);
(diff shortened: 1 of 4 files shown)


src/ddmd/mtype.d
```diff
@@ -910,6 +910,18 @@ extern (C++) abstract class Type : RootObject
         return buf.extractString();
     }
 
+    static const(char*)[2] toAutoQualChars(Type t1, Type t2)
```
JinShil: Please add DDoc comment header.

JinShil: Also, why the word "Auto"?  I would have expected something like `toFullyQualChars`.

JinShil (changes requested):

src/ddmd/mtype.d
ntrel: Will do. Auto is because it only qualifies the types if they would otherwise collide. Suggestions welcome.

JinShil: Ok, I see.  No need to change it in my opinion.  I was just curios.

WalterBright (changes requested):
Needs Ddoc header for toAutoQualChars() function. Also, please make it a non-member function instead of `static`.

src/ddmd/mtype.d
```diff
@@ -9171,3 +9171,19 @@ extern (C++) final class Parameter : RootObject
 
     extern (D) static immutable bool[SR.max + 1][SR.max + 1] covariant = covariantInit();
 }
+
+/***********************************************************
+ * For printing two types with qualification when necessary.
+ * Disambiguates the case where toChars() on each type match.
+ */
+const(char*)[2] toAutoQualChars(Type t1, Type t2)
```
JinShil: Still needs "Returns" and "Params".  See https://dlang.org/spec/ddoc.html

JinShil (changes requested):

src/ddmd/mtype.d
ntrel: Added `Returns`. I don't know anything meaningful to put for `Params`.

src/ddmd/mtype.d
```diff
@@ -9171,3 +9172,20 @@ extern (C++) final class Parameter : RootObject
 
     extern (D) static immutable bool[SR.max + 1][SR.max + 1] covariant = covariantInit();
 }
+
+/***********************************************************
+ * For printing two types with qualification when necessary.
+ * Disambiguates the case where toChars() on each type match.
+ * Returns: Names of types.
+ */
+const(char*)[2] toAutoQualChars(Type t1, Type t2)
```
JinShil: How about this?
```D
/**
* For printing two types with qualification when necessary.
* Params:
*    t1 = The first type to receive the type name for
*    t2 = The second type to receive the type name for
* Returns:  
*    The fully-qualified names of both types if the two type names are not the same, 
*    or the unqualified names of both types if the two type names are the same.
*/

Comment (ntrel) [reply src/ddmd/mtype.d]: Added.

**Review by JinShil — APPROVED** [review]

**Comment by ntrel** [comment]
BTW I have follow up PRs for 3 other kinds of error message where the same fix applies.

File: test/fail_compilation/bug9631.d [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+}
+
+template T2()
+{
+    struct F { }
+}
+
+void main()
+{
+    T2!().F x = T1!().F();
```
JinShil: Sorry @ntrel but this isn't passing coverage.  You'll have to add a test that tests the case when both `t1` and `t2` are the same.

JinShil (changes requested):

test/fail_compilation/bug9631.d
JinShil: Nevermind, The coverage report looks weird, and I project coverage looks good, so let's merge this.

PR #7408, opened 2017-12-08
Windows: fix building with Microsoft compiler

rainers:


JinShil (approved):

wilzbach:
Should we setup AppVeyor to constantly check this, s.t. stuff like this can't happen in the future?

PR #7409, opened 2017-12-08
Create semantic time TransitiveVisitor

RazvanN7:
This PR implements a visitor which automatically traverses the AST. Since traversal is a common operation, this visitor will replace some existing visitors in later PRs/commits, saving lines of code.

### DIFF
--- a/src/ddmd/astcodegen.d
+++ b/src/ddmd/astcodegen.d
@@ -44,6 +44,7 @@ struct ASTCodegen
     alias Parameter                 = ddmd.mtype.Parameter;
     alias Taarray                   = ddmd.mtype.Taarray;
     alias Tsarray                   = ddmd.mtype.Tsarray;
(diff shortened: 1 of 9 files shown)


wilzbach:
Huh? What's the issue with force-pushing? There shouldn't be a need to open a new PR every time, otherwise you need to start getting more familiar with `git` ;-)

RazvanN7:
I wanted to change the branch name and thought it would automatically update the branch on github, but that wasn't the case.

wilzbach:
> I wanted to change the branch name and thought it would automatically update the branch on github, but that wasn't the case.

Yeah you can't rename the branch name of a PR, but FYI you can rename it locally:

```
git push origin Trasitive_compiler_visitor
```

And push from whatever branch you have locally.
Or even change the remote tracking permanently:

```
git branch --set-upstream=origin/Trasitive_compiler_visitor
```

PR #7411, opened 2017-12-08
Create semantic time TransitiveVisitor

RazvanN7:
This PR implements a visitor which automatically traverses the AST. Since traversal is a common operation, this visitor will replace some existing visitors in later PRs/commits, saving lines of code.

### DIFF
--- a/src/ddmd/transitivevisitor.d
+++ b/src/ddmd/transitivevisitor.d
@@ -4,16 +4,24 @@ module ddmd.transitivevisitor;
 
 import ddmd.permissivevisitor;
 import ddmd.tokens;
-
 import ddmd.root.rootobject;
 
 import core.stdc.stdio;
 
 /** Visitor that implements the AST traversal logic. The nodes just accept their children.
   */
-extern(C++) class TransitiveVisitor(AST) : PermissiveVisitor!AST
+extern(C++) class ParseTimeTransitiveVisitor(AST) : PermissiveVisitor!AST
 {
     alias visit = PermissiveVisitor!AST.visit;
+    mixin MParseVisitMethods!AST;
+}
+
+/* This mixin implements the AST traversal logic for parse time AST nodes. The same code
+ * is used for semantic time AST node traversal, so in order to not duplicate the code,
+ * the template mixin is used.
+ */
+mixin template MParseVisitMethods(AST)
+{
 
 //   Statement Nodes
 //===========================================================
@@ -1135,4 +1143,3 @@ extern(C++) class TransitiveVisitor(AST) : PermissiveVisitor!AST
         }
     }
 }
-
--- a/src/ddmd/transitivevisitor.d
+++ b/src/ddmd/transitivevisitor.d
@@ -13,14 +13,14 @@ import core.stdc.stdio;
 extern(C++) class ParseTimeTransitiveVisitor(AST) : PermissiveVisitor!AST
 {
     alias visit = PermissiveVisitor!AST.visit;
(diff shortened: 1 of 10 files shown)


src/ddmd/visitor.d
```diff
@@ -83,6 +86,122 @@ public:
...
+{
+    alias visit = GenericPermissiveVisitor!AST.visit;
+
+    mixin MParseVisitMethods!AST;
+
+    override void visit(AST.PeelStatement s)
+    {
+        if (s.s)
+            s.s.accept(this);
+    }
```
jacob-carlborg: Please leave an empty newline between each method declaration.

src/ddmd/transitivevisitor.d
```diff
@@ -4,16 +4,24 @@ module ddmd.transitivevisitor;
...
 {
     alias visit = PermissiveVisitor!AST.visit;
+    mixin MParseVisitMethods!AST;
+}
+
+/* This mixin implements the AST traversal logic for parse time AST nodes. The same code
+ * is used for semantic time AST node traversal, so in order to not duplicate the code,
+ * the template mixin is used.
+ */
+mixin template MParseVisitMethods(AST)
```
jacob-carlborg: Is this part of the public API? If not, can we mark it as `package`?

jacob-carlborg (changes requested):
I would like that any new public symbols have proper documentation, i.e. Ddoc comments. If they should not be part of the public API we should try to use the compiler to enforce this. That is, set the appropriate viability attribute, i.e. `package`, `private` and so on.

src/ddmd/transitivevisitor.d
RazvanN7: Good point. I will make it package.

RazvanN7:
@jacob-carlborg Is this ok?

src/ddmd/visitor.d
```diff
@@ -83,6 +91,146 @@ public:
...
+    override void visit(AST.Condition){}
+    override void visit(AST.Initializer){}
+}
+
+/* Transitive visitor instantiated with the code generation AST family
+ */
+alias SemanticTimeTransitiveVisitor = GenericTransitiveVisitor!ASTCodegen;
+
+// Generic transitive visitor which implements the AST traversal logic for all AST nodes
+private extern (C++) class GenericTransitiveVisitor(AST) : GenericPermissiveVisitor!AST
```
jacob-carlborg: Perhaps it's just me, but I don't think it's very clear by reading the comment what the difference is compared to `GenericVisitor`.

src/ddmd/visitor.d
```diff
@@ -83,6 +91,146 @@ public:
...
+
+    override void visit(AST.PeelStatement s)
+    {
+        if (s.s)
+            s.s.accept(this);
+    }
+
+    override void visit(AST.UnrolledLoopStatement s)
+    {
+        foreach(sx; *s.statements)
```
jacob-carlborg: I would recommend adding curly braces to the body of the foreach loop, to avoid any confusion of which code belongs to which body. That's a general recommendation when there are nested statements like this.

jacob-carlborg (changes requested):
I found couple of more things.

andralex (approved):
thx!

RazvanN7:
@jacob-carlborg How about now? (I forgot that ddoc comments begin with /** )

jacob-carlborg:
A bit too late but I think it looks good 👍.

PR #7412, opened 2017-12-08
[WIP] Add AppVeyor config to test Visual Studio builds

wilzbach:
To avoid issues like https://github.com/dlang/dmd/pull/7408 in the future

I added AppVeyor to the list of CIs to be able to test this PR. Of course, I will remove it if we decide to not go with this.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,40 @@
+platform: x64
+environment:
+ matrix:
+  #- DC: dmd
+    #DVersion: nightly
+    #arch: x64
+  - DC: dmd
+    DVersion: nightly
+    arch: x86
+skip_tags: false
+build_script:
+ - ps: wget https://raw.githubusercontent.com/wilzbach/installer/appveyor-script/script/install.ps1 -OutFile install.ps1
+ - ps: .\install.ps1
+ - call "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall.bat" %compilersetupargs%
+ - ps: function InstallDMC
+        {
+            pushd c:\\;
+            Invoke-WebRequest "https://github.com/wilzbach/dmc-mirror/raw/master/dm857c.zip" -OutFile "dmc.zip";
+            7z x dmc.zip > $null;
+            $env:PATH = "C:\dm\bin;$($env:PATH)";
+            popd;
+        }
+ - ps: InstallDMC
+
+test_script:
+ - git clone --branch %APPVEYOR_REPO_BRANCH% https://github.com/dlang/druntime ..\druntime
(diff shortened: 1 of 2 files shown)


rainers:
There is also this attempt: https://github.com/dlang/dmd/pull/7214 which builds through the dmc wrapper.
If you want to build through VS and Visual D, see https://github.com/dlang/visuald/blob/master/appveyor.yml#L67 how to install Visual D silently.

I wonder if we want to add another CI though, when the existing ones keep producing spurious failures.

JinShil:
IMO, if it's supported it should be tested.

> There is also this attempt: #7214 which builds through the dmc wrapper.

What are the advantages/disadvantages of this PR vs that PR?

> I wonder if we want to add another CI though, when the existing ones keep producing spurious failures.

A fair point, but if building with Visual Studio is supported, what are our alternatives for testing it?

rainers:
> IMO, if it's supported it should be tested.

AFAIU it's still not officially supported, we are distributing dmc built dmd (though slower than the VC build - I just benchmarked building the phobos unittests again: dmc: 6min46, VC/32bit: 5min02, VC/64bit: 5min28).

> What are the advantages/disadvantages of this PR vs that PR?

They are testing different build systems (make vs. VC/VD projects), but use the same compiler.

> A fair point, but if building with Visual Studio is supported, what are our alternatives for testing it?

Last time I checked, the auto tester have VS2013 installed, too. They are already the bottleneck for exisiting builds, though.

wilzbach:
>>> I wonder if we want to add another CI though, when the existing ones keep producing spurious failures.
>> A fair point, but if building with Visual Studio is supported, what are our alternatives for testing it?
> Last time I checked, the auto tester have VS2013 installed, too. They are already the bottleneck for exisiting builds, though.

Well yeah there are three reasons:
- updating auto-tester is really hard and troublesome (involves a couple of pings to Brad - usually takes weeks or months), whereas an AppVeyor script can be updated more or less instantly
- I have made good experiences with AppVeyor in the past two years
- As Rainers mentioned, we don't have many Windows machines in the auto-tester building cloud

-> I would give AppVeyor a try.

> There is also this attempt: #7214 which builds through the dmc wrapper.

Wow I didn't know that PR existed.

> They are testing different build systems (make vs. VC/VD projects), but use the same compiler.

Well #7214 looks more complete as it even runs the testsuite (-> will close this PR).
@rainers you seem to have by far more experience than I do in regards of Windows and Visual D. I just hacked this together, because I was frustrated about the lack of automatic testing. However, I would feel much more confident if an actual user of Windows, Visual Studio and VisualD would add the build script.
I am more than to re-enable the AppVeyor hook and try to remove other barriers, though.

PR #7414, opened 2017-12-09
Experiment with testing `unittest` on Windows

wilzbach:
This will probably take a couple of hops again. Let's hope for the best...

### DIFF
--- a/src/win32.mak
+++ b/src/win32.mak
@@ -307,6 +307,7 @@ trace:
 
 unittest:
 	$(DMDMAKE) "OPT=-o" "DEBUG=" "DDEBUG=-debug -g -unittest -cov" "DOPT=" "LFLAGS=-L/ma/co/delexe/la" $(TARGETEXE)
+	$(TARGETEXE)
 
 ################################ Libraries ##################################
 
--- a/src/win32.mak
+++ b/src/win32.mak
@@ -306,7 +306,7 @@ trace:
 	$(DMDMAKE) "OPT=-o" "DEBUG=-gt -Nc" "DDEBUG=-debug -g -unittest" "DOPT=" "LFLAGS=-L/ma/co/delexe/la" $(TARGETEXE)
 
 unittest:
-	$(DMDMAKE) "OPT=-o" "DEBUG=" "DDEBUG=-debug -g -unittest -cov" "DOPT=" "LFLAGS=-L/ma/co/delexe/la" $(TARGETEXE)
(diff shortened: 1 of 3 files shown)


wilzbach:
```
Error on line 8: expecting target : dependencies
```

Looks like some patching error? :/

wilzbach:
> Looks like some patching error? :/

Yep -> https://github.com/braddr/at-client/blob/master/src/diff-dmd-win64.diff
We should really get rid of these :/

wilzbach:
This is blocked on https://github.com/braddr/at-client/pull/8 due to auto-tester's patching :/

wilzbach:
This is now failing with:

```
..\generated\windows\release\32\dmd.obj(dmd) 
 Error 42: Symbol Undefined ?out_config_debug@@YAX_N000000@Z
```

Has anyone seen this before?

jacob-carlborg:
Unless you already figured this out, `out_config_debug` is defined in `dmd/backend/backconfig.c`, only when the C preprocessor macro `DEBUG` is defined. It's called from `backend_init` in a `debug` statement in `dmd/dmsc.d`.

wilzbach:
Looks like this one was "healed" by the DMD backend conversion to D (:

src/win32.mak
```diff
@@ -306,7 +306,8 @@ trace:
 	$(DMDMAKE) "OPT=-o" "DEBUG=-gt -Nc" "DDEBUG=-debug -g -unittest" "DOPT=" "LFLAGS=-L/ma/co/delexe/la" $(TARGETEXE)
 
 unittest:
-	$(DMDMAKE) "OPT=-o" "DEBUG=" "DDEBUG=-debug -g -unittest -cov" "DOPT=" "LFLAGS=-L/ma/co/delexe/la" $(TARGETEXE)
+	$(DMDMAKE) "OPT=-o" "DEBUG=" "DDEBUG=-g -unittest -cov" "DOPT=" "LFLAGS=-L/ma/co/delexe/la" $(TARGETEXE)
```
wilzbach: Debug builds fail also on Linux at the moment: https://github.com/dlang/dmd/pull/7528

wilzbach:
CC @thewilsonator this is ready to go and passing on the CIs.

thewilsonator (approved):

thewilsonator:
Auto-merge toggled on

thewilsonator:
Auto-merge toggled off

thewilsonator:
Its failing agin now.

RazvanN7 (approved):

rainers:
removing auto-merge because the unittest is failing. Running dmd without args is considered an error, something like `dmd --version` should be used instead.

MoonlightSentinel:
This can be closed beceause #10612 enabled unittesting on windows

wilzbach:
Thanks!

PR #7415, opened 2017-12-09
Start testing the DMD DUB package on Travis

wilzbach:
Supersedes https://github.com/dlang/dmd/pull/6999

Submitting this to see the list of errors we run into. The last time there has been a problem with `gdc`.

See also: https://github.com/dlang/dmd/pull/6771

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -42,7 +42,6 @@ subPackage {
 
   sourceFiles \
     "src/ddmd/astbase.d" \
-    "src/ddmd/astbasevisitor.d" \
     "src/ddmd/parse.d" \
     "src/ddmd/transitivevisitor.d" \
     "src/ddmd/permissivevisitor.d" \
--- a/test/dub_package/.gitignore
+++ b/test/dub_package/.gitignore
@@ -3,3 +3,4 @@ docs.json
 __dummy.html
 *.o
 *.obj
+/parser
--- a/test/dub_package/dub.sdl
+++ b/test/dub_package/dub.sdl
@@ -1,5 +1,0 @@
-name "dmd-dub-test"
-description "Test of the DMD Dub package"
-license "BSL 1.0"
-
-dependency "dmd" path="../../"
--- a/test/dub_package/parser.d
+++ b/test/dub_package/parser.d
@@ -1,3 +1,6 @@
+/+dub.sdl:
+dependency "dmd" path="../.."
++/
 // The tests in this module are highlevel and mostly indented to make sure all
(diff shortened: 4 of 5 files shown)


jacob-carlborg:
So why doesn't it find the Dub command? It works before the tests are started [1]. Are we messing with the `PATH` or something?

[1] https://travis-ci.org/dlang/dmd/jobs/313877077#L903

wilzbach:
> So why doesn't it find the Dub command? It works before the tests are started [1]. Are we messing with the PATH or something?

Ah because we need to explicitly load the compiler into the environment. I think this is done to ensure that the host compiler is only used in the `build` step and not by accident in other steps.

jacob-carlborg:
LGTM 👌.

wilzbach:
@RazvanN7 @jacob-carlborg FYI: I managed to build a DUB package which exposes the entire frontend: https://github.com/dlang/dmd/pull/7425

JackStouffer:
@wilzbach @mdparker BTW, has there been an article on the blog about the dub functionality? If not, we should really advertise this.

wilzbach:
> @wilzbach @mdparker BTW, has there been an article on the blog about the dub functionality?

Yes there was:

https://dlang.org/blog/2017/08/01/a-dub-case-study-compiling-dmd-as-a-library

>  If not, we should really advertise this.

The problem is that's still far from usable. While with https://github.com/dlang/dmd/pull/7425, it will finally be possible to run through all semantic stages with the DMD as a library package, there's still much work that needs to be done in:
- clean, modern D APIs
- overly complicated "hello world" -> add wrapper
- semantic versioning for DMD (at the moment one can only use `master` as other tags are "invalid semver" tags)
- Improve the easiness of writing a custom visitor (@RazvanN7 is working on this -> https://wiki.dlang.org/Projects/Visitormania)
- Introduce more structure in the DMD code base & refactorings (e.g. ddmd -> dmd transition, start to use D strings at more places) -> https://wiki.dlang.org/Projects/Libdcomp
- Provide an example project for a simple use case to get people started (e.g. a DScanner-like style check, C++ header generation, CTags/Etags output, AST XML dump, Find declaration, ...)

jacob-carlborg:
> Provide an example project for a simple use case to get people started (e.g. a DScanner-like style check, C++ header generation, CTags/Etags output, AST XML dump, Find declaration, ...)

I have  started on an AST dumper, but as a debugging tool within DMD and not outputting XML.

https://github.com/jacob-carlborg/dmd/tree/ast-printer

PR #7416, opened 2017-12-10
fix Issue 18053 - Use stdint.h mangling for int64_t/uint64_t when man…

WalterBright:
…gling D long/ulong

I'm tired of expending many hours dealing with this issue again and again. Fortunately, it only changes the mangling on OSX 64.

### DIFF
--- a/changelog/fix18053.dd
+++ b/changelog/fix18053.dd
@@ -1,0 +1,16 @@
+Fix Issue 18053 - mangle long/ulong as int64_t/uint64_t
+
+This is a breaking change (on OSX 64).
+
+Due to the erratic implementation defined behavior of C++ name mangling, it was
+difficult to get D's long/ulong to portably match up with the corresponding
+C++ compiler.
+
+By instead relying on how the corresponding C++ compiler mangled int64_t/uint64_t
+it makes the C++ side of the D<=>C++ interface much simpler.
+
+For the current platforms dmd supports, only the OSX 64 bit mangling changes. In
+this case from 'm' to 'y'.
+
+Note:
+    int64_t and uint64_t are defined in stdint.h
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -815,10 +815,10 @@ public:
             case Tuns32:                c = 'j';        break;
             case Tfloat32:              c = 'f';        break;
             case Tint64:
-                c = (Target.c_longsize == 8 ? 'l' : 'x');
+                c = Target.c_longsize == 8 ? Target.int64Mangle : 'x';
                 break;
             case Tuns64:
-                c = (Target.c_longsize == 8 ? 'm' : 'y');
+                c = Target.c_longsize == 8 ? Target.uint64Mangle : 'y';
                 break;
             case Tint128:                c = 'n';       break;
             case Tuns128:                c = 'o';       break;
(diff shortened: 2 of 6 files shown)


WalterBright:
Note how much nicer the test cases get :-)

WalterBright:
@ibuclaw @klickverbot please check effect on gdc/ldc

src/ddmd/cppmangle.d
```diff
@@ -815,10 +815,12 @@ public:
             case Tuns32:                c = 'j';        break;
             case Tfloat32:              c = 'f';        break;
             case Tint64:
-                c = (Target.c_longsize == 8 ? 'l' : 'x');
+                // Mangle like target's stdint.h int64_t
+                c = (Target.c_longsize == 4 || global.params.isOSX ? 'x' : 'l');
```
ibuclaw: This should be turned into a target hook. The frontend should not be asking questions such as "Am I targeting osx?". I think we already have a mangle type hook already present, the logic can be put there.

ibuclaw (dismissed):

src/ddmd/cppmangle.d
ibuclaw: https://github.com/dlang/dmd/blob/5494dba193d16d6e126af8a12d37567b4d245086/src/ddmd/target.d#L459

Comment would need to be updated to reflect that if the target requires mangling in a specific way, then handle it here.

jacob-carlborg:
I'm not sure which C++ compiler you're looking at but these are the results I got with Xcode 9.0 containing Clang `900.0.37`:

#### 64bit:

| Type               | C++ | D | D After PR |
|--------------------|-----|---|------------|
| `long`               | l   | l | x          |
| `long long`          | x   | l | x          |
| `int64_t`            | x   | l | x          |
| `unsigned long`      | m   | m | y          |
| `unsigned long long` | y   | m | y          |
| `uint64_t`           | y   | m | y          |

#### 32bit:

| Type               | C++ | D | D After PR |
|--------------------|-----|---|------------|
| `long`               | l   | i | i          |
| `long long`          | x   | x | x          |
| `int64_t`            | x   | x | x          |
| `unsigned long`      | m   | j | j          |
| `unsigned long long` | y   | y | y          |
| `uint64_t`           | y   | y | y          |

As can be seen in these tables, D uses the wrong mangling for the `long` type.

I've used the following C++ code:

```cpp
#include <stdint.h>

void a(long){}
void b(long long){}
void c(int64_t){}

void d(unsigned long){}
void e(unsigned long long){}
void f(uint64_t){}
```

And the following D code:

```d
extern (C++):

import core.stdc.config;
import core.stdc.stdint;

void a(c_long){}
void b(long){}
void c(int64_t){}

void d(c_ulong){}
void e(ulong){}
void f(uint64_t){}
```

jacob-carlborg:
Hmm, seems like `__c_long/__c_ulong` needs to be used on macOS 64bit and not only 32bit.

jacob-carlborg:
BTW, why this complicated logic with a struct with a specific name that matches the mangling of C++? Couldn't we make the following to work?

```d
pragma(mangle, "m") alias c_ulong = ulong;
```

jacob-carlborg:
> I'd welcome a better solution for that. But not in this PR.

I have something that is a work in progress.

JinShil:
@WalterBright Can you please rebase?

JinShil (approved):

wilzbach:
This broke the OSX build on Travis, which shows the problems @jacob-carlborg pointed out. Should we revert, s.t. the error can be properly fixed/worked out?

jacob-carlborg:
@wilzbach I have an alternative solution to the existing `__c_long` available here https://github.com/dlang/dmd/pull/7458. It adds support for specifying `pragma(mangle)` on aliases. Hopefully with this PR together with mine we could get the correct mangling without breaking any existing code.

PR #7418, opened 2017-12-11
Cleanup DMD as a library build files

wilzbach:
A few things I found when working on https://github.com/dlang/dmd/pull/7415 and looking at the examples for DMD as a library:

- Allow them to be directly executable (via the DMD DUB package)
- Add a wild rule to avoid redundant logic
- Remove irrelevant link flags like `-L-lstd++`
- Move example binaries to the generated directory
- Start building the examples on auto-tester (Posix-only for now) __IMPORTANT__: the examples are currently not part of our CI
- Make Phobos path in impvisitor example more robust
- Build the examples with the current, generated DMD binary (separate commit, if controversial)

I think this is the coolest bit is that the examples for having DMD as a library are executable via DUB. This also means that we could drop them from the `posix.mak` and `win32.mak` altogether, could move them to a different directory (e.g. the test suite or `samples`) and don't need to worry about writing any cross-platform Makefiles as DUB takes care of this.

CC @RazvanN7


src/posix.mak
```diff
@@ -467,9 +456,23 @@ $G/dmd-unittest: $(DMD_SRCS) $(ROOT_SRCS) $G/newdelete.o $G/lexer.a $(G_GLUE_OBJ
 unittest: $G/dmd-unittest
 	$<
 
+######## DMD as a library examples
+
+EXAMPLES=$(addprefix $G/examples/, avg impvisitor)
+PARSER_SRCS=$(addsuffix .d, $(addprefix $D/,parse astbase parsetimevisitor transitivevisitor permissivevisitor strictvisitor))
+
+$G/parser.a: $(PARSER_SRCS) $(LEXER_SRCS) $(ROOT_SRCS) $G/dmd $(SRC_MAKE)
+	CC="$(HOST_CXX)" $G/dmd -lib -of$@ $(MODEL_FLAG) -L-lstdc++ -J$G $(DFLAGS) $(PARSER_SRCS) $(LEXER_SRCS) $(ROOT_SRCS)
```
wilzbach: FYI: the existing `parser.a` target is only needed to build the examples

WalterBright:
A more general issue - with every PR, `src/ddmd/posix.mak` gets ever more convoluted and impenetrable. It's so awful, another makefile, `src/posix.mak`, was spawned to drive it, and that one over time got successively more convoluted and impenetrable.

I suspect there is a nice makefile in there somewhere trying to get out.

JinShil:
>  I suspect there is a nice makefile in there somewhere trying to get out.

 I never understood why we use something as arcane as `make` when we have this much more capable and expressive language at our disposal.  If you wanna go forward, you gotta put it in D.

wilzbach:
> I never understood why we use something as arcane as make when we have this much more capable and expressive language at our disposal. If you wanna go forward, you gotta put it in D.

See e.g. http://forum.dlang.org/post/mailman.3317.1497594616.31550.digitalmars-d@puremagic.com

tl;dr: this pops up every three to six months. The main problem is DM make on Windows which creates a huge chunk of duplication (it's especially bad for Phobos) as it only supports a limited subset of GNUmake.
While many solutions have been proposed (reggae, meson, cmake, ninja, ...), and you can read the NG for the reasons why this is probably not going to happen in the near future (many build systems require another language and things like reggae aren't considered stable enough).
Imho the more realistic option and is to drop support for building DMD on Windows with DM make and expect people to have GNUmake (we could also add a "bootstrap" script, which will do the installation & setup or maybe even ship it as part of the Windows release). This would already by a huge benefit and as [Rainers recently pointed out](https://github.com/dlang/dmd/pull/7214#discussion_r155941157) the DMD testsuite already only runs with GNUmake and even expects a Unix-like environment.

RazvanN7:
@wilzbach The examples weren't meant to test the parser library, but to offer an example on how the parser visitors can be used. I don't think that moving the sources to the test files is a good idea since those aren't actually tests.

>Allow them to be directly executable (via the DMD DUB package)

This makes it mandatory that you have the dub package installed while there might be a lot of users who do not use it.

wilzbach:
> The examples weren't meant to test the parser library, but to offer an example on how the parser visitors can be used. 

Point being was that the examples currently aren't built as part of the test suite, which can easily lead to them being in a broken / outdated state (like e.g. the DUB package which already doesn't work anymore - see https://github.com/dlang/dmd/pull/7415).

>  I don't think that moving the sources to the test files is a good idea since those aren't actually tests.

Well my point was to move them out of the main Makefile, s.t. the "nice Makefile" can slowly start to appear again ;-)

andralex (commented):
thx! cc @RazvanN7

PR #4924, opened 2015-08-23
ddmd cleanup and fixes

MartinNowak:
- cleanup makefiles
- drop -O -inline by default (too slow)
- remove unused headers and magicport
- update travis-ci test

PR #4535, opened 2015-03-31
[DDMD] Split C++ operator new declaration into a new file

yebblies:
DDMD has a different version of `rmem.c`, but the C++ operators still need to be defined in C++ code.  This split allows the object file to easily be included in ddmd.

@WalterBright This is needed to fix the performance issue with `new` in ddmd.

PR #7419, opened 2017-12-11
Remove special newdelete.o target

wilzbach:
I might be wrong, but AFAICT the D-part of dmd uses just `rmem.d`.

- remove linking newdelete.o to the libs (it's not used)
- remove dead code block

`newdelete.c` has been introduced in https://github.com/dlang/dmd/pull/4535 and added as Makefile target in https://github.com/dlang/dmd/pull/4924

### DIFF
--- a/src/ddmd/root/newdelete.c
+++ b/src/ddmd/root/newdelete.c
@@ -21,8 +21,6 @@
 
 #if !defined(USE_ASAN_NEW_DELETE)
 
-#if 1
-
 extern "C"
 {
     void *allocmemory(size_t m_size);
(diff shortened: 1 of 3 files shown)


WalterBright:
newdelete.c is still in win32.mak

wilzbach:
> newdelete.c is still in win32.mak

Yeah, I wasn't sure whether this will really pass on the auto-tester. Hence, I only tried `posix.mak`. I removed `newdelete.obj` from the `win32.mak` too, but for some reason CircleCi is still failing :/

WalterBright:
Circleci runs on Linux, so win32.mak changes should be irrelevant to that.

wilzbach:
This isn't going anywhere. Closing for now.

PR #6359, opened 2016-12-24
fix Issue 16794 - dmd not working on Ubuntu 16.10

MartinNowak:
- enable PIC by default on amd64 linux (no significant overhead, full
  PIC/PIE support)
- also see https://github.com/dlang/installer/pull/207

PR #7002, opened 2017-07-17
Use -fPIC by default

wilzbach:
The [release dmd.conf](https://github.com/dlang/dmd/blob/master/ini/linux/bin64/dmd.conf#L5) already contains `-fPIC`, but the default generated one doesn't.
This requires `libphobos2.a` to be built with `-fPIC`, but that one is already in the queue: https://github.com/dlang/phobos/pull/5586

Other PRs:

- [x] https://github.com/dlang/druntime/pull/1880
- [x] https://github.com/dlang/phobos/pull/5586

PR #7420, opened 2017-12-11
Allow to run the DMD testsuite on hardened systems

wilzbach:
Because most Linux OSes now use hardening (e.g. Debian, Ubuntu, Arch Linux) and the DMD test suite should work out of the box.
`-fPIC` has been the default on Linux x64 since `2.072.2`.

See also:
- https://github.com/dlang/dmd/blob/master/ini/linux/bin64/dmd.conf
- https://github.com/dlang/dmd/pull/6359
- https://github.com/dlang/installer/pull/207


wilzbach:
Okay so regarding:

```
 ... runnable/test_cdvecfill.d      -O (-mcpu=avx -mcpu=avx2)
../src/dmd -conf= -m64 -Irunnable -O -fPIC  -odgenerated/runnable -ofgenerated/runnable/test_cdvecfill_0 runnable/test_cdvecfill.d
generated/runnable/test_cdvecfill_0
Expected code sequence for load!(ubyte, 16) not found.
  Expected: 0x50 0x66 0x0f 0x6e 0xc7 0x66 0x0f 0x60 0xc0 0x66 0x0f 0x61 0xc0 0x66 0x0f 0x70 0xc0 0x00 0x59 0xc3
  Actual: 0x55 0x48 0x8b 0xec 0x66 0x0f 0x6e 0xc7 0x66 0x0f 0x60 0xc0 0x66 0x0f 0x61 0xc0 0x66 0x0f 0x70 0xc0 0x00 0x5d 0xc3 0x00 0x55 0x48 0x8b 0xec 0x0f 0xb6 0x07 0x66 0x0f 0x6e 0xc0 0x66 0x0f 0x60 0xc0 0x66 0x0f 0x61 0xc0 0x66 0x0f 0x70 0xc0 0x00 0x5d 0xc3 0x00 0x00 0x55 0x48 0x8b 0xec 0x66 0x0f 0x6e 0xc7 0x66 0x0f 0x60 0xc0
core.exception.AssertError@runnable/test_cdvecfill.d(895): Assertion failure
```

-> I updated the `cdvec_fill.d` file (this script is apparently only run on Linux x64)

test/runnable/test_cdvecfill.d
```diff
@@ -55,7 +55,7 @@ version (update)
         auto sink = appender!string();
         foreach (arch; [EnumMembers!Arch])
         {
-            auto args = [dmd, "-c", "-O", "-mcpu=" ~ arch.to!string, "test/runnable/test_cdvecfill.d"];
+            auto args = [dmd, "-c", "-fPIC", "-O", "-mcpu=" ~ arch.to!string, "test/runnable/test_cdvecfill.d"];
```
wilzbach: This line in the update script isn't necessary itself because the default `dmd.conf` for Linux x86_64 already adds `-fPIC` as default flag, but imho explicitness helps.

test/runnable/test_cdvecfill.d
```diff
@@ -123,730 +123,830 @@ alias AliasSeq(Args...) = Args;
...
+        /* push   rbp                     */ 0x55,
+        /* mov    rbp,rsp                 */ 0x48, 0x8b, 0xec,
         /* movd   xmm0,edi                */ 0x66, 0x0f, 0x6e, 0xc7,
         /* punpcklbw xmm0,xmm0            */ 0x66, 0x0f, 0x60, 0xc0,
         /* punpcklwd xmm0,xmm0            */ 0x66, 0x0f, 0x61, 0xc0,
         /* pshufd xmm0,xmm0,0x0           */ 0x66, 0x0f, 0x70, 0xc0, 0x00,
-        /* pop    rcx                     */ 0x59,
+        /* pop    rbp                     */ 0x5d,
         /* ret                            */ 0xc3,
     ]),
```
JinShil: Why was this code change necessary?

JinShil:
Looks like this may address this issue: https://issues.dlang.org/show_bug.cgi?id=18013

test/runnable/test_cdvecfill.d
wilzbach: Apparently when emitting position-independent code, different registers are used. There's an update script included in this file which I used to update the test cases.

JinShil: Would it be prudent to keep the original test for Non-PIC and simply keep this PR as a simple addition?

wilzbach:
> Looks like this may address this issue: https://issues.dlang.org/show_bug.cgi?id=18013

Yes it will. I'm a Arch Linux user myself, so I constantly bump into the `-fPIC` issues with D, but luckily with a rising number of merged PRs the problems have been drastically reduced. This is actually (hopefully) the last bit of a long series (https://github.com/dlang/phobos/pull/5586, https://github.com/dlang/phobos/pull/5750, https://github.com/dlang/druntime/pull/1880, https://github.com/dlang/dmd/pull/7002, https://github.com/dlang/tools/pull/264 ...) in my quest to build and test the main D repos on Arch Linux without needing to patch anything ;-)

test/runnable/test_cdvecfill.d
wilzbach: Yes I could use `version(D_PIC)`, but this would make the update script more complex. And I'm not sure it's necessary, because this script is only run under Linux `x86_64` atm and since 2.072.2 dmd uses `-fPIC` for compiled programs on Linux as default (for Darwin `-fPIC` is even fixed). Anyhow, if I add the old test cases, this would only make sense, if they are run as part of the test suite and then if not in `version(PIE)` we would have to detect if the environment is hardened and then avoid running the test cases.
And in case this will ever be needed, the original code is still saved in git and alternatively it can be easily regenerated with Martin's fancy update script.
So I would only do this extra work if  @MartinNowak or @WalterBright think it's essential.

JinShil (dismissed):

ZombineDev:
Can you add this to the list of bugs fixed:
https://issues.dlang.org/show_bug.cgi?id=18014

My fix is a bit different, I'll push it soon so we can compare our approaches.

ZombineDev:
I don't think your PR fixes all of the issues from 18014. When running:
```sh
docker run -it zombinedev/dmd-test-suite-docker --repos:dmd:pr/7420
```
I get these errors:

```
/usr/bin/ld: test_results/compilable/main.o: relocation R_X86_64_32 against `.rodata.str1.1' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: test_results/compilable/a.a(a_1_149.o): relocation R_X86_64_32 against symbol `_D28TypeInfo_S4tmpl__T4TmplTiZQi6__initZ' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: test_results/compilable/b.a(b_1_149.o): relocation R_X86_64_32 against symbol `_D28TypeInfo_S4tmpl__T4TmplTlZQi6__initZ' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Nonrepresentable section on output
Makefile:164: recipe for target 'test_results/compilable/test6461.sh.out' failed
make[1]: *** [test_results/compilable/test6461.sh.out] Error 1
 ... fail_compilation/diag6707.d    ()
 ... fail_compilation/fail14416.d   ()
 ... fail_compilation/ice8309.d     ()
 ... fail_compilation/ice10076.d    ()
 ... fail_compilation/fail233.d     -o- ()
/usr/bin/ld: test_results/compilable/test14894main.o: relocation R_X86_64_32 against symbol `_Dmain' can not be used when making a shared object; recompile with -fPIC
/usr/bin/ld: final link failed: Nonrepresentable section on output
Makefile:164: recipe for target 'test_results/compilable/test14894.sh.out' failed
make[1]: *** [test_results/compilable/test14894.sh.out] Error 1
```

ZombineDev:
I'll see if with your fix for 18013 my branch finally passes the test suite, and if that's the case I'll open an alternative PR.

ZombineDev:
https://github.com/ZombineDev/dmd/commits/fpic-for-testsuite

wilzbach:
> pros and cons of this vs #7420 ?

- We already cherry-picked the first commit of #7427 in https://github.com/dlang/dmd/pull/7454
- `cdvecfill.d` fix: https://github.com/dlang/dmd/pull/7454

which means that the two PRs are more or less similar. The only difference is that @ZombineDev explicitly added `$(PIC_FLAG)` to some tests, which AFAICT isn't needed as `-fPIC` is always set and the environment variable `REQUIRED_ARGS` is always appended.

`test/runnable/test17559.d` is still failing, but that's the already know [issue of broken line numbers with `-fPIC`](https://issues.dlang.org/show_bug.cgi?id=18068).

wilzbach:
image

We should probably add another CI run that goes without `-fPIC`.

wilzbach:
> 79.152% (-1.841%) compared to cd6b40f

Seems like we have to go with three containers :/

wilzbach:
image


Looks like we are good to go here :)
The [Docker image](https://github.com/wilzbach/dlang-docker), see e.g. https://github.com/dlang/dmd/pull/7579#issuecomment-357877739 for an example of the relocation errors you get.

The docker image is based on this: https://github.com/wilzbach/dlang-docker/blob/master/circleci/dlang.docker

Why 3 jobs and not 2?
------------------------------

Code coverage:

> 79.152% (-1.841%) compared to cd6b40f

Why is there now `circleci: pic` and `circleci: no_pic` appearing?
-------------------------------------------------------------------------------------

It's a feature from CodeCov to send different status notifications for different stages.

https://circleci.com/docs/2.0/workflows/

> Having the ci/circleci checkbox enabled will prevent the status from showing as completed in GitHub when using a workflow because CircleCI posts statuses to Github with a key that includes the job by name.

.circleci/config.yml
```diff
@@ -1,30 +1,34 @@
 version: 2
 jobs:
-  build:
+  no_pic:
     working_directory: ~/dmd
     docker:
       - image: circleci/node:4.8.2
```
wilzbach: We could replace this with a D Docker image that doesn't require `-fPIC`, e.g. https://github.com/wilzbach/dlang-docker/blob/master/circleci/dlang.docker built for Ubuntu 14.04, but this should be more than fine for now (it's the existing status quo after all).

.circleci/config.yml
```diff
@@ -1,30 +1,34 @@
...
     parallelism: 2
-    branches:
-      ignore:
-        - dmd-1.x
```
wilzbach: This didn't work with the new workflow setting and AFAICT is only required for the respective branch anyways.

.circleci/config.yml
```diff
@@ -1,30 +1,34 @@
...
-          command: sudo apt-get update && sudo apt-get install gdb
-          name: Install GDB
-      - run:
-          command: ./.circleci/run.sh setup-repos
-          name: Clone DRuntime & Phobos
-      - run:
-          command: ./.circleci/run.sh coverage
-          name: Run testsuite with -cov
+          command: ./.circleci/run.sh all
+          name: Run all
```
wilzbach: I reduced the commands to "all", s.t. it's easier to maintain both jobs.

test/Makefile
```diff
@@ -178,6 +178,7 @@ else
     endif
     export DFLAGS
 endif
+REQUIRED_ARGS+=$(PIC_FLAG)
```
wilzbach: After the other PRs that's all that's needed to fix the test suite for `-fPIC` :)

.circleci/run.sh
```diff
@@ -145,8 +160,16 @@ codecov()
...
-    setup-repos) setup_repos ;;
-    coverage) coverage ;;
-    check-clean-git) check_clean_git;;
-    codecov) codecov ;;
+    setup-repos) echo "removed" ;;
+    coverage) echo "removed" ;;
+    check-clean-git) echo "removed" ;;
+    codecov)
+        echo "removed - use 'all'"
+        ;&
```
wilzbach: fall-through is required, s.t. existing PRs don't need to be rebased, but will call `all` automatically.

JinShil:
Does the test suite pass?  Even [test/runnable/test17559.d](https://issues.dlang.org/show_bug.cgi?id=18068)

wilzbach:
> Does the test suite pass? Even test/runnable/test17559.d?

Yes:

image

.circleci/run.sh
JinShil: Probably a good idea to add that as a comment to the script.

JinShil:
This is looking great.  Can you give a final summary?  What's remaining before we can merge this?

.circleci/run.sh
wilzbach: Yes! Done.

wilzbach:
> Can you give a final summary?

I try to do an honest summary of the disadvantages I see as you know the advantages of testing automatically for `-fPIC` yourself.

1) One more container run at CircleCi

Con:
- This shouldn't be a huge problem. CircleCi is still one of our fastest CIs and we have four containers (=jobs) per organization available.
- If this really becomes a bottleneck, we can always easily disable the `pic` stage

2) Dependency on the [D docker image](https://github.com/wilzbach/dlang-docker)

Con:
- it's configured with Travis CRON + auto-deploy for new releases, s.t. not much maintenance should be required
- it uses the official CircleCi Dockerfile and just adds the D installer + a few packages on top 
- it's just [29 lines](https://github.com/wilzbach/dlang-docker/blob/master/circleci/dlang.docker)

3) One more status update by CircleCi (`no_pic` and `pic`) = more visual noise

I haven't found a good solution for this :/

4) The status check `ci/circleci` was still set to "enforced". As the newly deployed dlang-bots/dlang-bot#69 feature lets the bot respect the failing CIs properly, I set "ci/circleci" to non-enforced until we reach a decision here.

> What's remaining before we can merge this?

This is ready from my sides. CIs are happy -> I'm happy

JinShil (approved):
Nice work! 👍

wilzbach:
The only thing that is still unclear to me is why the stack traces work on the Docker image. 
It's failing on my local system and clearly I'm not the only one:

http://forum.dlang.org/post/ufpwobqmisamazigcaav@forum.dlang.org
https://issues.dlang.org/show_bug.cgi?id=18068

Maybe it's not as assumed due to PIE, but sth. else. glibc maybe? Arch Linux currently ships `2.26-10`

JinShil:
> It's failing on my local system and clearly I'm not the only one:

Same here.  That's why I asked.

wilzbach:
> Evidence certainly points to something else. I use Arch Linux too. I don't think it should hold up this PR.

I agree, so let's move forward here?

JinShil:
> I agree, so let's move forward here?

Yep, just wanted to give others a few days opportunity to chime it.  I think they've had enough time.  auto-merge toggling on...

JinShil (approved):

PR #7421, opened 2017-12-11
make outbuffer printf nothrow

WalterBright:


WalterBright:
This is failing because va_start, va_end, and va_copy are not nothrow. But nothrow was added nearly 2 years ago:

https://github.com/dlang/druntime/commit/d600fa79b3f45c4bf83afe75fd70f699a849a54d

andralex (approved):

andralex:
@WalterBright then why does the autotester fail?

mathias-lang-sociomantic:
Because we're using a DMD which is older than v2.071.0, when the va_* primitives where made `nothrow`.

andralex:
@mathias-lang-sociomantic oh ok, so I guess this will wait for a while?

braddr:
Waiting won't do much.  The defined bootstrap compiler for DMD is 2.068.2.  That's not going to change without discussion and explicit actions.  There's repercussions to changing the minimum version that go well beyond the scope of this PR.

andralex:
@braddr thanks. Where is the right starting point in moving the version forward?

wilzbach:
>  Where is the right starting point in moving the version forward?

AFAICT Walter wants to use `-betterC` for the backend (https://github.com/dlang/dmd/pull/6907), which will require updating all host compilers.

wilzbach:
Btw what about making these functions templated without args or using `static if`?

braddr:
@andralex a thread (dmd-internal?  main d group?) that gains consensus among the relevant players that it's the right thing to do.  It's not my call, I don't have much of a stake in it, just pointing out the reality is that 2.068.2 is what has been defined as the long term minimum version that's required to be able to build the code.  Changing it isn't the whim of a particular pull but a larger ecosystem issue.

andralex:
cc @WalterBright

WalterBright:
My motivation on making things `nothrow` in dmd is so that the code can use RAII without penalty. Of course, the 'without penalty' bit also depends on dmd changes I made within the last month or so.

Aside from gdc, ldc, and the autotester, who are the other stakeholders in this?

jacob-carlborg:
@WalterBright can you add a try-catch around the call to the functions not marked with `nothrow`?
@WalterBright @andralex we need to decide if we should upgrade the host compilers, there seems to be many different reasons why we want to upgrade.

wilzbach:
As 2.079.0 is now required to build DMD, this might finally pass on our CIs.
I rebased it and let's hope for the best.

jacob-carlborg (approved):

wilzbach:
Hmm so this looks like it's blocked on GDC (we still use GDC 4.8.5 which is based on 2.068).
If we don't manage to get a newer GDC package in the near future, we could `version` this PR...

ibuclaw:
You're on semaphore now, why not just get gdc-7 from this ppa? https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test

wilzbach:
>  You're on semaphore now, why not just get gdc-7 from this ppa? 

Because I didn't know about this PPA. Thanks a lot for the hint!
I'm playing around with it in #8431.

PR #7425, opened 2017-12-12
Expose D frontend via DUB

wilzbach:
Allows to use the entire frontend as a DUB package.

There's still an ugly lot of initialization needed and two hard-coded bits (platform, importPaths), but one can run through all semantic phases which is more than needed if we want to take the DMD as a library project serious ;-)

I will comment on the individual lines for which I would like to find a better solution.

Also note that I didn't add this to the test scripts yet as I am waiting for https://github.com/dlang/dmd/pull/7415 (allows to test the DUB package on Travis).

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -100,6 +100,7 @@ subPackage {
     "src/dmd/escape.d" \
     "src/dmd/expression.d" \
     "src/dmd/expressionsem.d" \
+    "src/dmd/frontend.d" \
     "src/dmd/func.d" \
     "src/dmd/gluelayer.d" \
     "src/dmd/hdrgen.d" \
--- a/dub.sdl
+++ b/dub.sdl
@@ -53,101 +53,16 @@ subPackage {
 subPackage {
   name "frontend"
   targetType "library"
+  preGenerateCommands `cd "$${DUB_PACKAGE_DIR}" && ./config.sh generated/dub VERSION /etc`
+  stringImportPaths "generated/dub"
   stringImportPaths "res"
   versions "NoBackend"
   versions "GC"
   versions "NoMain"
-  sourcePaths
-
-  sourceFiles \
-    "src/dmd/access.d" \
-    "src/dmd/aggregate.d" \
-    "src/dmd/aliasthis.d" \
-    "src/dmd/apply.d" \
-    "src/dmd/argtypes.d" \
-    "src/dmd/arrayop.d" \
-    "src/dmd/arraytypes.d" \
-    "src/dmd/astcodegen.d" \
-    "src/dmd/attrib.d" \
-    "src/dmd/blockexit.d" \
-    "src/dmd/builtin.d" \
-    "src/dmd/canthrow.d" \
-    "src/dmd/clone.d" \
-    "src/dmd/complex.d" \
-    "src/dmd/cond.d" \
-    "src/dmd/constfold.d" \
-    "src/dmd/cppmangle.d" \
-    "src/dmd/cppmanglewin.d" \
-    "src/dmd/ctfeexpr.d" \
-    "src/dmd/dcast.d" \
-    "src/dmd/dclass.d" \
-    "src/dmd/declaration.d" \
-    "src/dmd/delegatize.d" \
-    "src/dmd/denum.d" \
-    "src/dmd/dimport.d" \
-    "src/dmd/dinifile.d" \
-    "src/dmd/dinterpret.d" \
-    "src/dmd/dmacro.d" \
-    "src/dmd/dmangle.d" \
-    "src/dmd/dmodule.d" \
-    "src/dmd/doc.d" \
-    "src/dmd/dscope.d" \
-    "src/dmd/dstruct.d" \
-    "src/dmd/dsymbol.d" \
-    "src/dmd/dsymbolsem.d" \
-    "src/dmd/dtemplate.d" \
-    "src/dmd/dversion.d" \
-    "src/dmd/escape.d" \
-    "src/dmd/expression.d" \
-    "src/dmd/expressionsem.d" \
-    "src/dmd/frontend.d" \
-    "src/dmd/func.d" \
-    "src/dmd/gluelayer.d" \
-    "src/dmd/hdrgen.d" \
-    "src/dmd/id.d" \
-    "src/dmd/impcnvtab.d" \
-    "src/dmd/imphint.d" \
-    "src/dmd/init.d" \
-    "src/dmd/initsem.d" \
-    "src/dmd/inline.d" \
-    "src/dmd/inlinecost.d" \
-    "src/dmd/intrange.d" \
-    "src/dmd/json.d" \
-    "src/dmd/lib.d" \
-    "src/dmd/link.d" \
-    "src/dmd/mars.d" \
-    "src/dmd/mtype.d" \
-    "src/dmd/nogc.d" \
-    "src/dmd/nspace.d" \
-    "src/dmd/objc.d" \
-    "src/dmd/opover.d" \
-    "src/dmd/optimize.d" \
-    "src/dmd/parse.d" \
-    "src/dmd/parsetimevisitor.d" \
-    "src/dmd/printast.d" \
-    "src/dmd/safe.d" \
-    "src/dmd/sapply.d" \
-    "src/dmd/semantic.d" \
-    "src/dmd/sideeffect.d" \
-    "src/dmd/statement.d" \
-    "src/dmd/statement_rewrite_walker.d" \
-    "src/dmd/statementsem.d" \
-    "src/dmd/staticassert.d" \
-    "src/dmd/staticcond.d" \
-    "src/dmd/target.d" \
-    "src/dmd/templateparamsem.d" \
-    "src/dmd/traits.d" \
-    "src/dmd/typesem.d" \
-    "src/dmd/typinf.d" \
-    "src/dmd/utils.d" \
-    "src/dmd/visitor.d"
-
-   sourceFiles "src/dmd/scanelf.d" \
-    "src/dmd/libelf.d" platform="linux"
-
-   sourceFiles "src/dmd/scanmach.d" \
-    "src/dmd/libmach.d" platform="osx"
-
-  dependency "dmd:parser" version="*"
-  dependency "dmd:lexer" version="*"
+  sourcePaths "src/dmd"
+  excludedSourceFiles "src/dmd/backend/*"
+  excludedSourceFiles "src/dmd/{e2ir,eh,glue,iasm,objc_glue,s2ir,tocsym,toctype,toobj,todt,toir}.d"
+  excludedSourceFiles "src/dmd/{scan,lib}{mach,mscoff,omf}.d" platform="linux"
+  excludedSourceFiles "src/dmd/{scan,lib}{elf,mscoff,omf}.d" platform="osx"
+  excludedSourceFiles "src/dmd/{scan,lib}{elf,mach,mscoff}.d" platform="windows"
 }
--- a/dub.sdl
+++ b/dub.sdl
@@ -5,7 +5,7 @@ copyright "Copyright © 1999-2017, The D Language Foundation"
 license "BSL-1.0"
 
 targetType "none"
-dependency ":parser" version="*"
+dependency ":frontend" version="*"
 
 subPackage {
   name "root"
@@ -47,5 +47,106 @@ subPackage {
     "src/dmd/permissivevisitor.d" \
     "src/dmd/strictvisitor.d"
 
+  dependency "dmd:lexer" version="*"
+}
+
+subPackage {
+  name "frontend"
+  targetType "library"
+  stringImportPaths "res"
+  versions "NoBackend"
+  versions "GC"
+  versions "NoMain"
+  sourcePaths
+
+  sourceFiles \
+    "src/dmd/access.d" \
+    "src/dmd/aggregate.d" \
+    "src/dmd/aliasthis.d" \
+    "src/dmd/apply.d" \
+    "src/dmd/argtypes.d" \
+    "src/dmd/arrayop.d" \
+    "src/dmd/arraytypes.d" \
+    "src/dmd/astcodegen.d" \
+    "src/dmd/attrib.d" \
+    "src/dmd/blockexit.d" \
+    "src/dmd/builtin.d" \
+    "src/dmd/canthrow.d" \
+    "src/dmd/clone.d" \
+    "src/dmd/complex.d" \
+    "src/dmd/cond.d" \
+    "src/dmd/constfold.d" \
+    "src/dmd/cppmangle.d" \
+    "src/dmd/cppmanglewin.d" \
+    "src/dmd/ctfeexpr.d" \
+    "src/dmd/dcast.d" \
+    "src/dmd/dclass.d" \
+    "src/dmd/declaration.d" \
+    "src/dmd/delegatize.d" \
+    "src/dmd/denum.d" \
+    "src/dmd/dimport.d" \
+    "src/dmd/dinifile.d" \
+    "src/dmd/dinterpret.d" \
+    "src/dmd/dmacro.d" \
+    "src/dmd/dmangle.d" \
+    "src/dmd/dmodule.d" \
+    "src/dmd/doc.d" \
+    "src/dmd/dscope.d" \
+    "src/dmd/dstruct.d" \
+    "src/dmd/dsymbol.d" \
+    "src/dmd/dsymbolsem.d" \
+    "src/dmd/dtemplate.d" \
+    "src/dmd/dversion.d" \
+    "src/dmd/escape.d" \
+    "src/dmd/expression.d" \
+    "src/dmd/expressionsem.d" \
+    "src/dmd/func.d" \
+    "src/dmd/gluelayer.d" \
+    "src/dmd/hdrgen.d" \
+    "src/dmd/id.d" \
+    "src/dmd/impcnvtab.d" \
+    "src/dmd/imphint.d" \
+    "src/dmd/init.d" \
+    "src/dmd/initsem.d" \
+    "src/dmd/inline.d" \
+    "src/dmd/inlinecost.d" \
+    "src/dmd/intrange.d" \
+    "src/dmd/json.d" \
+    "src/dmd/lib.d" \
+    "src/dmd/link.d" \
+    "src/dmd/mars.d" \
+    "src/dmd/mtype.d" \
+    "src/dmd/nogc.d" \
+    "src/dmd/nspace.d" \
+    "src/dmd/objc.d" \
+    "src/dmd/opover.d" \
+    "src/dmd/optimize.d" \
+    "src/dmd/parse.d" \
+    "src/dmd/parsetimevisitor.d" \
+    "src/dmd/printast.d" \
+    "src/dmd/safe.d" \
+    "src/dmd/sapply.d" \
+    "src/dmd/semantic.d" \
+    "src/dmd/sideeffect.d" \
+    "src/dmd/statement.d" \
+    "src/dmd/statement_rewrite_walker.d" \
+    "src/dmd/statementsem.d" \
+    "src/dmd/staticassert.d" \
+    "src/dmd/staticcond.d" \
+    "src/dmd/target.d" \
+    "src/dmd/templateparamsem.d" \
+    "src/dmd/traits.d" \
+    "src/dmd/typesem.d" \
+    "src/dmd/typinf.d" \
+    "src/dmd/utils.d" \
+    "src/dmd/visitor.d"
+
+   sourceFiles "src/dmd/scanelf.d" \
+    "src/dmd/libelf.d" platform="linux"
+
+   sourceFiles "src/dmd/scanmach.d" \
+    "src/dmd/libmach.d" platform="osx"
+
+  dependency "dmd:parser" version="*"
   dependency "dmd:lexer" version="*"
 }
--- a/test/dub_package/.gitignore
+++ b/test/dub_package/.gitignore
@@ -1,6 +1,13 @@
+# Ignore all binary files (files without an extension)
+# Start: Ignore everything
+*
+# Include directories
+!*/
+# Include all files with extensions
+!*.*
+
 .dub
 docs.json
 __dummy.html
 *.o
 *.obj
-/parser
--- a/test/dub_package/README.md
+++ b/test/dub_package/README.md
@@ -1,0 +1,21 @@
+DMD as a library
+================
+
+The tests in this module are high-level and mostly indented to make sure all
+necessary modules are included in the Dub package.
+
+The tests are executable single-file packages for the D's package manager `dub`
+and can be directly executed:
+
+```bash
+./lexer.d
+```
+
+If you want to see the log output or want to pass additional options, use `--single`:
+
+```bash
+dub --single -v lexer.d
+```
+
+If you don't have `dub` installed on your system,
+[install an official release](https://dlang.org/download.html).
--- a/test/dub_package/frontend.d
+++ b/test/dub_package/frontend.d
@@ -1,0 +1,162 @@
+#!/usr/bin/env dub
+/+dub.sdl:
+dependency "dmd" path="../.."
++/
+import std.stdio;
+
+// add import paths
+void addImports(T)(T path)
+{
+    import dmd.globals : global;
+    import dmd.arraytypes : Strings;
+
+    stderr.writefln("addImport: %s", path);
+
+    Strings* res = new Strings();
+    foreach (p; path)
+    {
+        import std.string : toStringz;
+        Strings* a = new Strings();
+        a.push(p.toStringz);
+        res.append(a);
+    }
+    global.path = res;
+}
+
+// finds a dmd.conf and parses it for import paths
+auto findImportPaths()
+{
+    import std.file : exists, getcwd;
+    import std.string : fromStringz, toStringz;
+    import std.path : buildPath, buildNormalizedPath, dirName;
+    import std.process : env = environment, execute;
+    import dmd.dinifile : findConfFile;
+    import dmd.errors : fatal;
+    import std.algorithm, std.range, std.regex;
+
+    auto dmdEnv = env.get("DMD", "dmd");
+    auto whichDMD = execute(["which", dmdEnv]);
+    if (whichDMD.status != 0)
+    {
+        stderr.writeln("Can't find DMD.");
+        fatal;
+    }
+
+    immutable dmdFilePath = whichDMD.output;
+    string iniFile;
+
+    if (dmdEnv.canFind("ldmd"))
+    {
+        immutable ldcConfig = "ldc2.conf";
+        immutable binDir = dmdFilePath.dirName;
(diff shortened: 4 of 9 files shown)


wilzbach:
> wilzbach requested a review from WalterBright as a code owner just now

Btw @WalterBright this happens automatically because of

https://github.com/dlang/dmd/blob/f27ee714def99771232190f99c5e7193472c9bfe/CODEOWNERS#L34

If you mind about this, we should improve the mappings.

dub.sdl
```diff
@@ -42,11 +42,110 @@ subPackage {
 
   sourceFiles \
     "src/ddmd/astbase.d" \
-    "src/ddmd/astbasevisitor.d" \
```
wilzbach: Same change required as inhttps://github.com/dlang/dmd/pull/7415 (atm moment the DUB file is outdated).

dub.sdl
```diff
@@ -42,11 +42,110 @@ subPackage {
...
+    "src/ddmd/gluelayer.d"
+
+   sourceFiles "src/ddmd/scanelf.d" \
+    "src/ddmd/libelf.d" platform="linux"
+
+   sourceFiles "src/ddmd/scanmach.d" \
+    "src/ddmd/libmach.d" platform="osx"
+
+  dependency "dmd:parser" version="*"
+  dependency "dmd:lexer" version="*"
```
wilzbach: FWIW splitting up dmd in four packages creates a noticeable overhead in terms of DUB compile times (compared to one big library file).

src/ddmd/root/rmem.d
```diff
@@ -55,6 +55,11 @@ version (GC)
         }
     }
 
+    extern (C) void* allocmemory(size_t m_size) nothrow
+    {
+        return GC.malloc(m_size);
+    }
```
wilzbach: It seems like `version(GC)` isn't used actively (at least by anything that is part of the testsuite/CIs).

test/dub_package/source/app.d
```diff
@@ -43,3 +43,119 @@ unittest
...
+
+    // Initialization
+    Type._init();
+    Id.initialize();
+    Module._init();
+    Target._init();
+    Expression._init();
+    Objc._init();
+    builtin_init();
+}
```
wilzbach: Is there any chance we can move a generic initialization function into the DMD codebase?

test/dub_package/source/app.d
```diff
@@ -43,3 +43,119 @@ unittest
...
+    import ddmd.astcodegen : ASTCodegen;
+    import ddmd.dmodule : Module;
+    import ddmd.globals : Loc;
+    import ddmd.parse : Parser;
+    import ddmd.statement : Identifier;
+    import ddmd.tokens : TOKeof;
+    import ddmd.id : Id;
+
+    init;
+    addImports(["/usr/include/dlang/dmd/"]);
```
wilzbach: This is the worst bit about this PR. While, for this test we could mock `object.d`, I think we should show potential users how they can add import paths as they most likely want to do so.
Ideas:
- mock `object.d`
- use sth. like `../../../druntime/import` (i.e. depend on existing repo layout)
- call DMD and search config strings
- refactor config parsing from `mars.d`, s.t. it can be used here

test/dub_package/source/app.d
```diff
@@ -43,3 +43,119 @@ unittest
...
+{
+    import ddmd.astcodegen : ASTCodegen;
+    import ddmd.dmodule : Module;
+    import ddmd.globals : Loc;
+    import ddmd.parse : Parser;
+    import ddmd.statement : Identifier;
+    import ddmd.tokens : TOKeof;
+    import ddmd.id : Id;
+
+    init;
```
wilzbach: We still need to define all predefined versions, but `addDefaultVersionIdentifiers` is `private`.
Imho if I would be about to write an IDE plugin, I wouldn't want to keep this list in sync if DMD already has the logic.
For example, to import Phobos on Linux, I need this list:

```
import ddmd.cond : VersionCondition;
VersionCondition.addPredefinedGlobalIdent("DigitalMars");
VersionCondition.addPredefinedGlobalIdent("Posix");
VersionCondition.addPredefinedGlobalIdent("linux");
VersionCondition.addPredefinedGlobalIdent("ELFv1");
VersionCondition.addPredefinedGlobalIdent("LittleEndian");
VersionCondition.addPredefinedGlobalIdent("CRuntime_Glibc");
VersionCondition.addPredefinedGlobalIdent("D_Version2");
VersionCondition.addPredefinedGlobalIdent("all");
VersionCondition.addPredefinedGlobalIdent("D_InlineAsm_X86_64");
VersionCondition.addPredefinedGlobalIdent("X86_64");
VersionCondition.addPredefinedGlobalIdent("D_HardFloat");
```

So any chance we can refactor `addDefaultVersionIdentifiers` and expose it?

test/dub_package/source/app.d
```diff
@@ -43,3 +43,119 @@ unittest
...
+    Strings* res = new Strings();
+    foreach (p; path)
+    {
+        import std.string : toStringz;
+        Strings* a = new Strings();
+        a.push(p.toStringz);
+        res.append(a);
+    }
+    global.path = res;
+}
```
wilzbach: Well, obviously it would be nice if the `dmd` "library` contains a couple of high-level wrapper, s.t. a potential author only has to interact with its shiny, modern D APIs.

dub.sdl
jacob-carlborg: I thought it would be nice to be able to only pull in the lexer, i.e. for source code highlighting. But since Dub will currently download the whole DMD package anyway it doesn't matter that much.

Although, it would be nice to have some kind of enforcement of the separation between the different phases, to avoid that the lexer suddenly starts to depend on the rest of the front end.

test/dub_package/source/app.d
jacob-carlborg: Currently this is handled in `mars.d`, which contains the `main` function of DMD. This file should not be included in the Dub package, but I don't see why we can't move this to a separate function, called `initFrontEnd` which can be called both from the compiler and the Dub package. Perhaps we can add a new file with a shared constructor that is only part of the Dub package that calls this `initFrontEnd` function.

wilzbach: > Perhaps we can add a new file with a shared constructor that is only part of the Dub package that calls this initFrontEnd function.

Yeah, I think this would be the best way to move forward. This file could also contain other convenience, modern D wrapper. How about `ddmd/libdmd.d`, `ddmd/library.d`, or `ddmd/frontend.d`?

jacob-carlborg: I don't like `ddmd/libdmd.d`, the others are fine, I think I prefer `ddmd/frontend.d`. But I don't want to limit ourselves to this file. I think we should look at what we want to add and add it to the modules where it makes most sense. I don't mind adding several new modules and or packages if that makes more sense.

wilzbach:
OK, so I added a rudimentary `dmd.conf` detection which is a combination from `findConfFile` and a quick regex to parse the import paths.

-> this should now work on Travis

> I don't like ddmd/libdmd.d, the others are fine, I think I prefer ddmd/frontend.d. But I don't want to limit ourselves to this file. I think we should look at what we want to add and add it to the modules where it makes most sense. I don't mind adding several new modules and/or packages if that makes more sense. I don't have any problems with having only one symbol in a module if it doesn't fit anywhere else.

-> I created `ddmd/frontend.d` and moved the initialization calls to this module to prime it.

Other points (now not displayed anymore):

- [We still need to define all predefined versions, but addDefaultVersionIdentifiers is private.](https://github.com/dlang/dmd/pull/7425#discussion_r156264693)
- "obviously it would be nice if the dmd "library` contains a couple of high-level wrapper, s.t. a potential author only has to interact with its shiny, modern D APIs."

dub.sdl
wilzbach: Well I hope that in practice one doesn't need to constantly recompile `libdmd.a`, so it shouldn't matter too much.

test/dub_package/.gitignore
```diff
@@ -1,6 +1,13 @@
+# Ignore all binary files (files without an extension)
```
wilzbach: As hopefully more examples will be added, ignoring all binary files in this directory avoids frequent updates of this file.

dub.sdl
```diff
@@ -49,3 +49,105 @@ subPackage {
...
+    "src/ddmd/nspace.d" \
+    "src/ddmd/objc.d" \
+    "src/ddmd/opover.d" \
+    "src/ddmd/optimize.d" \
+    "src/ddmd/parse.d" \
+    "src/ddmd/sapply.d" \
+    "src/ddmd/templateparamsem.d" \
+    "src/ddmd/semantic.d" \
+    "src/ddmd/sideeffect.d" \
+	"src/ddmd/statement.d" \
```
jacob-carlborg: The indentation is wrong here.

dub.sdl
```diff
@@ -49,3 +49,105 @@ subPackage {
...
+    "src/ddmd/staticassert.d" \
+    "src/ddmd/target.d" \
+    "src/ddmd/typesem.d" \
+    "src/ddmd/traits.d" \
+    "src/ddmd/parsetimevisitor.d" \
+    "src/ddmd/visitor.d" \
+    "src/ddmd/typinf.d" \
+    "src/ddmd/utils.d" \
+	"src/ddmd/statement_rewrite_walker.d" \
+	"src/ddmd/statementsem.d" \
```
jacob-carlborg: The indentation is wrong here.

test/dub_package/README.md
```diff
@@ -0,0 +1,17 @@
...
+necessary modules are included in the Dub package.
+
+The tests are executable single-file packages for the D's package manager `dub`
+and can be directly executed:
+
+   ./lexer.d
+
+If you want to see the log output or want to pass additional options, use `--single`:
+
+  dub --single -v lexer.d
```
jacob-carlborg: This line and line [10](https://github.com/dlang/dmd/pull/7425/files#diff-16b8627c835f07289426c754bd2472efR10) doesn't have the same indentation. For code blocks I recommend the triple backticks.

test/dub_package/README.md
```diff
@@ -0,0 +1,17 @@
+DMD as a library
+================
```
jacob-carlborg: Personally I prefer the hash `#` for headers. Then I only need one character and don't have to add/remove characters if the title changes.

test/dub_package/frontend.d
```diff
@@ -0,0 +1,121 @@
...
+    import ddmd.tokens : TOKeof;
+    import ddmd.id : Id;
+    import ddmd.globals : global;
+
+    global._init;
+    global.params.isLinux = 1;
+    initDMD;
+    findImportPaths.addImports;
+
+    auto parse(Module m, string code) {
```
jacob-carlborg: Opening curly braces on their own lines.

jacob-carlborg (changes requested):

test/dub_package/README.md
wilzbach: Yeah I usually use a mix of these styles in my markdown: `===` for the title and `#` for all headers within the file as the title usually changes seldom, but has more emphasis this way.
Anyhow I wouldn't mind at all to use a `#` if that's important, but I think there are more pressing issues than which flavor of Markdown should be used?

wilzbach:
(addressed the comments & hopefully found the issue for the failing Travis test.)

test/dub_package/README.md
jacob-carlborg: No, it's not that important. That's why I said it's my personal preference.

dub.sdl
```diff
@@ -49,3 +49,105 @@ subPackage {
...
+    "src/ddmd/visitor.d" \
+    "src/ddmd/typinf.d" \
+    "src/ddmd/utils.d" \
+    "src/ddmd/statement_rewrite_walker.d" \
+    "src/ddmd/statementsem.d" \
+    "src/ddmd/staticcond.d" \
+    "src/ddmd/safe.d" \
+    "src/ddmd/blockexit.d" \
+    "src/ddmd/printast.d" \
+    "src/ddmd/gluelayer.d"
```
jacob-carlborg: Minor detail, it would be nice if the list was sorted alphabetically, unless it's sorted in some other way that I fail to see now.

dub.sdl
```diff
@@ -49,3 +49,105 @@ subPackage {
...
+
+   sourceFiles "src/ddmd/scanelf.d" \
+    "src/ddmd/libelf.d" platform="linux"
+
+   sourceFiles "src/ddmd/scanmach.d" \
+    "src/ddmd/libmach.d" platform="osx"
```
jacob-carlborg: The last three files, I know they're not technically not part of the backend but I think they feel more geared towards the backend or code generation. Do we want to include them?

src/ddmd/frontend.d
```diff
@@ -0,0 +1,39 @@
...
+module ddmd.frontend;
+
+// Online documentation: https://dlang.org/phobos/ddmd_frontend.html
+
+/**
+Initializes the DMD compiler
+*/
+void initDMD()
+{
+    import ddmd.globals : global;
```
jacob-carlborg: This is not used as far as I can see.

test/dub_package/frontend.d
```diff
@@ -0,0 +1,124 @@
...
+    import ddmd.globals : global, Loc;
+    import ddmd.frontend : initDMD;
+    import ddmd.parse : Parser;
+    import ddmd.statement : Identifier;
+    import ddmd.tokens : TOKeof;
+    import ddmd.id : Id;
+    import ddmd.globals : global;
+
+    global._init;
+    global.params.isLinux = 1;
```
jacob-carlborg: Do we need to worry about this when the tests are also run on macOS?

jacob-carlborg (changes requested):

dub.sdl
wilzbach: Currently they are required by some bits of the frontend (see below). Though I'm pretty sure, that these parts could `version`ed out in the future.

#### Remove libelf.d

```
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(lib.o):(.data.rel.ro+0x28): undefined reference to `_D4ddmd6libelf12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(lib_532_2d5.o): In function `_D4ddmd3lib7Library7factoryFZCQBcQBaQz':
/home/seb/dlang/dmd/src/ddmd/lib.d:56: undefined reference to `LibElf_factory()'
```

#### Remove `gluelayer.d`

```
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(dclass.o):(.data.rel.ro+0x20): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(denum.o):(.data.rel.ro+0x10): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(dmodule.o):(.data.rel.ro+0x28): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(dsymbol.o):(.data.rel.ro+0x30): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(dsymbolsem.o):(.data.rel.ro+0xd8): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(dsymbolsem_39a_657.o): In function `Semantic3Visitor::visit(FuncDeclaration*)':
/home/seb/dlang/dmd/src/ddmd/dsymbolsem.d:1043: undefined reference to `retStyle(TypeFunction*)'
/home/seb/dlang/dmd/src/ddmd/dsymbolsem.d:1097: undefined reference to `retStyle(TypeFunction*)'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(expression.o):(.data.rel.ro+0x40): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(func.o):(.data.rel.ro+0x28): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(mars.o):(.data.rel.ro+0x20): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(mtype.o):(.data.rel.ro+0x38): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(objc.o):(.data.rel.ro+0x60): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(objc_5ba_3c6.o): In function `_D4ddmd4objc9Supported6__ctorMFZCQBfQBdQBb':
/home/seb/dlang/dmd/src/ddmd/objc.d:185: undefined reference to `objc_initSymbols()'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(statement.o):(.data.rel.ro+0x30): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(statementsem.o):(.data.rel.ro+0xd0): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(statementsem_632_9d9.o): In function `StatementSemanticVisitor::visit(AsmStatement*)':
/home/seb/dlang/dmd/src/ddmd/statementsem.d:3923: undefined reference to `asmSemantic(AsmStatement*, Scope*)'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(typinf.o):(.data.rel.ro+0x48): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(aggregate.o):(.data.rel.ro+0x18): undefined reference to `_D4ddmd9gluelayer12__ModuleInfoZ'
/tmp/.dub/build/dmd-~dub-frontend/library-debug-linux.posix-x86_64-dmd_2077-A550EF4C39A32552F8FE470A957BC323/libdmd_frontend.a(typinf_680_468.o): In function `genTypeInfo(Type*, Scope*)':
/home/seb/dlang/dmd/src/ddmd/typinf.d:67: undefined reference to `toObjFile(Dsymbol*, bool)'
```

jacob-carlborg: I see. I think we can do some refactoring later to remove these dependencies if we want to.

BTW, I think `lib.d` is for generating libraries, which I don't think we need.

jacob-carlborg (approved):

jacob-carlborg:
LGTM 👌.

andralex (approved):
For an initial release, looks great. cc @RazvanN7

dub.sdl
```diff
@@ -49,3 +49,105 @@ subPackage {
...
+subPackage {
+  name "frontend"
+  targetType "library"
+  stringImportPaths "res"
+  versions "NoBackend"
+  versions "GC"
+  versions "NoMain"
+  sourcePaths
+
+  sourceFiles \
```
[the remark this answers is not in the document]
jacob-carlborg: One way is to move every file that is part of the Dub package to its own directory, then the whole directory can be included with `sourcePaths`.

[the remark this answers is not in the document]
wilzbach: DUB also supports `excludedSourceFiles` for removing files from a `sourcePaths` set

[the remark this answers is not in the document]
wilzbach: Give it a try. Seems to be simple enough, though it would have been easier with a better layout of DMD (see the exclude file globs) and I don't know how fast the globbing is implemented in DUB (though it's a handful of files)...

If you want to cleanup the house, some ideas:
- move everything related to the IR -> `dmd.ir`
- move everything related to reading & writing object files `src/dmd/{scan,lib}{elfmach,mscoff,omf}.d` -> `dmd.obj`

test/dub_package/.gitignore
ZombineDev: I think a cleaner solution would be put all dub artifacts in `generated/` as well, using `copyFiles` (see http://code.dlang.org/package-format?lang=sdl).

wilzbach:
> For an initial release,

So I finally got Travis to pass :)
I had to add a small configuration file searcher for LDC to get Travis to pass.
I excluded GDC from checking the dub package for now, because there were too many errors and using Phobos fully is very convenient. Also IIRC a new release for GDC is scheduled to happen sometime in early 2018, so by the time this DUB package is fully usable (i.e. we have proper semantic versioning and have worked out all the other issues), using DMD as a library via DUB with GDC won't be a problem either. From what I can tell are these examples intended to be showcases anyhow and this PR changes the status quo from not usable to compiling & working with a simple example.

Finally the "simple" frontend example now has more than 150 lines, so in a follow-up PR we should try to factor out common logic to `dmd.frontend` (CC @RazvanN7).

wilzbach:
@andralex @jacob-carlborg @RazvanN7 (or other) okay to move forward with this as an initial version? I would like to look into @joakim-noah's [report](http://forum.dlang.org/post/vvhtgzyyofehmrwsfrlq@forum.dlang.org) that we can't due a full semantic on Phobos yet and also start to build stuff on top of this ;-)

jacob-carlborg:
Yes, please move forward 👌.

jacob-carlborg:
Auto-merge toggled on

test/dub_package/frontend.d
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env dub
+/+dub.sdl:
+dependency "dmd" path="../.."
++/
+import std.stdio;
+
+// add import paths
+void addImports(T)(T path)
```
RazvanN7: Why isn't this in frontend.d? (src/dmd/frontend.d)

test/dub_package/frontend.d
```diff
@@ -0,0 +1,162 @@
...
+        import std.string : toStringz;
+        Strings* a = new Strings();
+        a.push(p.toStringz);
+        res.append(a);
+    }
+    global.path = res;
+}
+
+// finds a dmd.conf and parses it for import paths
+auto findImportPaths()
```
RazvanN7: Why isn't this in frontend.d ?

test/dub_package/frontend.d
```diff
@@ -0,0 +1,162 @@
...
+    Identifier id = Identifier.idPool("test");
+    auto m = new Module("test.d", id, 0, 0);
+    m.members = parse(m, q{
+        void foo()
+        {
+            foreach (i; 0..10) {}
+        }
+    });
+
+    void semantic()
```
RazvanN7: This should certainly be in frontend.d and we could call it  `fullSemanticAnalysis` or something along these terms.

RazvanN7 (commented):
Other than some minor comments it looks good. The next step should be to expose more semantic methods in a way that is easy to use, for example: the one which I find very important is the search method which does symbol resolution.

test/dub_package/frontend.d
jacob-carlborg: Unless there's a reason to not run the full semantic analysis I don't think the `full` prefix is necessary. I hope that we can keep the three phases an implementation detail.

wilzbach:
> The next step should be to expose more semantic methods in a way that is easy to use, for example: the one which I find very important is the search method which does symbol resolution.

Yes they should be (I think I mentioned this a couple of times), but for now I didn't want to push anything to `frontend.d` as I didn't want to stall this PR unnecessarily with a discussion of the exposed user API (@RazvanN7 how about having a chat in the next days regarding this?)
Also, once this is in, we can trivially refactor the code, as we will have at least one CI checking it :)

test/dub_package/frontend.d
RazvanN7: I'm not sure this is possible. As a tool developer you may want to visit the AST while it is in a certain state. Given the fact that the semantic methods modify the AST quite a lot you might want to take a look at it before a certain semantic pass is done. For example: enums are evaluated at compile time  (during semantic3) ; in order to do those modifications the semantic3 stage alters the AST; as a tool developer I may want to do something in with the original AST which is not longer possible given that semantic3 has modified it.

wilzbach: Hehe that's exactly why I didn't want to push this into `frontend.d` for now ;-)

RazvanN7:
@wilzbach Sure thing. I was just suggesting the next step after this PR is merged.

test/dub_package/frontend.d
RazvanN7: That's not a problem. You have the individual semantic passes, but if you don't care about that you can just call `FullSemanticAnalysys` or some other name that you find suited.

jacob-carlborg: Please let leave it as is and continue improve it after this is merged.

wilzbach:
> Why isn't this in frontend.d? (src/dmd/frontend.d)
>  I was just suggesting the next step after this PR is merged.

Yes, let's continue the discussion at https://github.com/dlang/dmd/pull/7485

PR #7426, opened 2017-12-12
Cleanup Makefile

wilzbach:
While diving into https://github.com/dlang/dmd/pull/7419, I realized that there are a couple of unused / outdated Makefile targets.

> I suspect there is a nice makefile in there somewhere trying to get out.

Going `-17` towards it.

Please see the individual commits for their motivation.

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -576,7 +576,7 @@ $(TOOLS_DIR)/checkwhitespace.d:
 
 # See https://github.com/dlang/dmd/pull/7358 for why -xc++ is used here
 # -O is needed for FreeBSD's compiler to silence "-Wuninitialized is not supported without -O"
-checkcxxheaders:
+checkcxxheaders: $(ROOT_SRC) $(SRC)
 	$(HOST_CXX) -xc++ -O -fsyntax-only $(ROOT_FLAGS) $(filter %.h,$(ROOT_SRC))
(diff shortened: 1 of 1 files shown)


src/posix.mak
```diff
@@ -290,6 +278,8 @@ endif
 endif
 
 
+######## DMD frontend source files
```
JinShil: I read source code quite often through a portrait-oriented monitor.  I think the headers are an improvement, though I'd prefer something more visually significant like is done at the bottom
```Make
################################################################################
# DDoc documentation generation
################################################################################
```
Or maybe something like:
```Make
# DDoc documentation generation
################################################################################
```
Regardless, it would be nice if they were all consistently the same.  But, I don't want to start a bikeshedding discussion over it.  This PR is still an improvement, as it stands IMO.

JinShil (approved):

src/posix.mak
wilzbach: Hehe I actually like the more significant lines a lot more as well, I just wanted to keep this PR as minimal as possible to avoid potential controversies and allow a quick merge.

JinShil: You could also get rid of the ...
```Make
######################################################
```
lines with no documentation.  They add nothing but noise, IMO.  Perhaps add some text, or remove them.

wilzbach: Yes, good idea. I will do so, but in a separate PR ;-)

andralex (approved):

PR #7427, opened 2017-12-12
[WIP] Alternative approach to allow to running the DMD testsuite on hardened systems

PetarKirov:
Alternative to https://github.com/dlang/dmd/pull/7420.

To reproduce the original problem, simply run:
```
docker run -it zombinedev/dmd-test-suite-docker --repos:dmd:master
```

To verify the changes in this PR, run:
```
docker run -it zombinedev/dmd-test-suite-docker --repos:dmd:pr/7427
```


ZombineDev:
Alternative to https://github.com/dlang/dmd/pull/7420

test/runnable/test_cdvecfill.d
```diff
@@ -123,730 +123,830 @@ alias AliasSeq(Args...) = Args;
...
+        /* push   rbp                     */ 0x55,
+        /* mov    rbp,rsp                 */ 0x48, 0x8b, 0xec,
         /* movd   xmm0,edi                */ 0x66, 0x0f, 0x6e, 0xc7,
         /* punpcklbw xmm0,xmm0            */ 0x66, 0x0f, 0x60, 0xc0,
         /* punpcklwd xmm0,xmm0            */ 0x66, 0x0f, 0x61, 0xc0,
         /* pshufd xmm0,xmm0,0x0           */ 0x66, 0x0f, 0x70, 0xc0, 0x00,
-        /* pop    rcx                     */ 0x59,
+        /* pop    rbp                     */ 0x5d,
         /* ret                            */ 0xc3,
     ]),
```
JinShil: If `PIC=0` won't that cause this test to fail if we accept these changes?

ZombineDev: I'm not sure. I just cherry-picked @wilzbach's fix to try to get the test suite to pass on my machine.
It seems this will need more work.

JinShil: I think you may need to keep both tests:  This one for -fPIC and the test without these changes for Non-PIC.  That is, if you intend to allow both PIC and non-PIC compiles, which appears to be  the case, unless I'm misunderstanding.

ZombineDev:
I'm not satisfied with my PR yet, as in my docker setup I get the following error:
```
Running runnable tests
 ... runnable/gdb1.d                -g -fPIC -fPIC ()
../src/dmd -conf= -m64 -Irunnable -g -fPIC -fPIC  -odtest_results/runnable -oftest_results/runnable/gdb1_0 runnable/gdb1.d
gdb test_results/runnable/gdb1_0 --batch -x test_results/runnable/gdb1_0.gdb
Breakpoint 1 at 0x842: file runnable/gdb1.d, line 15.
Could not trace the inferior process.
Error: ptrace: Operation not permittedtest_results/runnable/gdb1_0.gdb:3: Error in sourced command file:
During startup program exited with code 127.


Test failed: 
GDB regex: 'RESULT=.*ARG1.*ARG2' didn't match output:
----
Breakpoint 1 at 0x842: file runnable/gdb1.d, line 15.
Could not trace the inferior process.
Error: ptrace: Operation not permittedtest_results/runnable/gdb1_0.gdb:3: Error in sourced command file:
During startup program exited with code 127.

----


Makefile:167: recipe for target 'test_results/runnable/gdb1.d.out' failed
make[1]: *** [test_results/runnable/gdb1.d.out] Error 1
```

wilzbach:
> Alternative to #7420.

Okay, I think we should go step by step here and thus as I first step I cherry-picked your first commit and modified it slightly: https://github.com/dlang/dmd/pull/7433

Note that stack traces on Linux are broken in PIE, see e.g. https://issues.dlang.org/show_bug.cgi?id=18068 (a good reason why we should let at least one CI run the test suite with `-fPIC`)

test/runnable/test_cdvecfill.d
wilzbach: How about copy/pasting this file and using `version(D_PIC) {} else` in the other? This has the advantage that the update script would still work.

andralex:
pros and cons of this vs #7420 ?

JinShil:
@ZombineDev I'm leaning more towards #7420.  Any particular drawbacks of #7420 that this PR does not suffer from?  If not, is it OK to close this in favor of #7420?

JinShil:
I'm wondering, however, if we really need to add `${PIC_FLAG}` to these tests.  I think we can just explicitly add `-fPIC` to these tests as non-PIE-hardened systems can still verify the functionality of the test in PIC-mode.

The drawback, however, is there would be no verification of the test in non-PIC mode, but I'm not sure if that's a problem.

ZombineDev:
@JinShil I am only propagating the PIC flag to the shell-based tests, because the rest pick it up from the REQUIRED_ARGS variable. An alternative solution would be propagate the REQUIRED_ARGS for the shell-based tests.

JinShil:
>  I am only propagating the PIC flag to the shell-based tests,

Ah, right.  Sorry, I'm still learning.

>An alternative solution would be propagate the REQUIRED_ARGS for the shell-based tests and not treat PIC specifically.

That actually seems like the best approach.  I'm wondering why it's not done already, as the name implies, it should be required.  Can you give it a try?

JinShil:
Ok, so after thinking about this more, I'm thinking #7420 is the way to go.  @ZombineDev is there any particular strength this PR has over #7420 that we should consider?

ZombineDev:
@wilzbach substantially improved his PR in the meantime to use a docker container (I had the initial idea, but not the time to realize it), so now I'm very happy to go with #7420

PR #7432, opened 2017-12-13
Fix DDoc warnings and add a CI check to avoid regressions

wilzbach:
Individual commits:
- Remove duplicated id.d from FRONT_SRCS (It's already part of LEXER_SRCS)
- Add new 'style' target to check for Ddoc errors
- Run style target on CircleCi
- Fix DDoc warnings

### DIFF
--- a/src/dmd/backend/oper.d
+++ b/src/dmd/backend/oper.d
@@ -272,7 +272,9 @@ int rel_unord(int op)     { return _rel_unord    [op - RELOPMIN]; }
 /****************************************
  * Conversion operators.
  * Convert from conversion operator to conversion index
- * parallel array invconvtab[] in cgelem.c)
+ * parallel array invconvtab[] in cgelem.c
+ * Params:
+ *   op = conversion operator
  */
 int convidx(OPER op) { return op - CNVOPMIN; }
 
--- a/src/dmd/dinifile.d
+++ b/src/dmd/dinifile.d
@@ -169,8 +169,9 @@ void updateRealEnvironment(StringTable* environment)
  *      environment = our own cache of the program environment
  *      filename = name of the file being parsed
  *      path = what @P will expand to
- *      buffer[len] = contents of configuration file
- *      sections[] = section names
+ *      length = length of the configuration file buffer
+ *      buffer = contents of configuration file
+ *      sections = section names
  */
 void parseConfFile(StringTable* environment, const(char)* filename, const(char)* path, size_t length, ubyte* buffer, Strings* sections)
 {
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -6116,7 +6116,6 @@ extern (C++) class TemplateInstance : ScopeDsymbol
      * Compare proposed template instantiation with existing template instantiation.
      * Note that this is not commutative because of the auto ref check.
      * Params:
-     *  this = proposed template instantiation
      *  o = existing template instantiation
      * Returns:
      *  0 for match, 1 for no match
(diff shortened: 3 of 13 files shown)


CyberShadow:
Shouldn't DAutoTest already (be able to) enforce DDoc warnings?

wilzbach:
> Shouldn't DAutoTest already (be able to) enforce DDoc warnings?

Unfortunately is `-de` effectless on Ddoc warnings.
We do the same thing as part of the `style` target @ Phobos:

https://github.com/dlang/phobos/blob/master/posix.mak#L591

The idea is that this check can be run without needing to clone dlang.org
It's also a chicken-egg problem if we want to avoid even more complex Makefile rules at DAutotest, as we e.g. druntime doesn't have this check at all and here it still will take a bit until it is in `stable` (aka part of a release).

CyberShadow:
`-de` is for deprecations, did you mean `-w`?

CyberShadow:
> It's also a chicken-egg problem if we want to avoid even more complex Makefile rules at DAutotest, as we e.g. druntime doesn't have this check at all and here it still will take a bit until it is in stable (aka part of a release).

Yeah but it's wasted work because the documentation is going to be built anyway...

wilzbach:
> Yeah but it's wasted work because the documentation is going to be built anyway...

Ah I see what you are going for. Fair enough. Done.

CyberShadow (approved):
Thanks, makefile stuff looks good. Can't judge correctness of the new documentation though.

wilzbach:
> Thanks, makefile stuff looks good.

Ah I realized that `-w` doesn't trigger a halt, which makes it rather pointless :/
I opened a PR to fix it (-> https://github.com/dlang/dmd/pull/7440), but as we still use 2.072.2 for dlang.org, that's not a solution for now.
Thus I added a grep to the Makefile documentation target. There was also an `-of-` which lead to `src/-` being generated. I replaced it with the proper `-o-`

CyberShadow:
> but as we still use 2.072.2 for dlang.org

Huh? DAutoTest builds DMD then uses it to build everything else, no?

wilzbach:
> Huh? DAutoTest builds DMD then uses it to build everything else, no?

Yes and no - `$(STABLE_DMD)` is used to build the documentation.

wilzbach:
> $(STABLE_DMD) is used to build the documentation.

See also: https://github.com/dlang/dlang.org/pull/1926

wilzbach:
Removed the controversial `grep` to check for deprecation errors and added `-w` instead to unblock this PR. As of now `-w` is defunct, but https://github.com/dlang/dmd/pull/7440 will hopefully take care of this soon.

PetarKirov (approved):

wilzbach:
>  but #7440 will hopefully take care of this soon.

This is in - so we are good to go here (:

> but as we still use 2.072.2 for dlang.org

Okay. I now changed the documentation build to use the currently build DMD. After all, `dlang.org` already passes [`DMD` to the respective Makefile](https://github.com/dlang/dlang.org/blob/master/posix.mak#L265). So it can be easily overridden if needed. Also using an ancient version of Ddoc means that none of the bugs fixed in the Ddoc build will propagate to dlang.org.
DRuntime and Phobos use the latest version of DMD to build the documentation already (see https://github.com/wilzbach/druntime/blob/master/posix.mak#L152 and https://github.com/wilzbach/phobos/blob/udflags/posix.mak#L93).

wilzbach:
BTW as this PR changes the compiler we use to build the docs, this will fix e.g. this recently introduced page:

Online:

image

https://dlang.org/phobos-prerelease/dmd_semantic2.html

----------------

With this PR

image

PR #7435, opened 2017-12-14
Fix issue18076: allow -run to work with - (stdin)

quickfur:
Turns out, only a trivial change is needed to make this work, by hooking into the code implemented in #6880 for reading source code from stdin.

With this change, dmd acquires the functionality:
````
$ echo 'import std.stdio; void main() { writeln("Hello, World!"); }' | dmd -run -
Hello, World!
$
````
which compiles and runs standard input without the user needing to specify any intermediate filenames.

The above is a trivial example, of course.  A more pertinent use case would be executing the source code in an editor's buffer via stdin, and having the output captured by the editor as stdout, without needing to specify intermediate temporary files.  Another use case would be code generation tools that can instantly produce output without needing to manage temporary files.

### DIFF
--- a/src/ddmd/mars.d
+++ b/src/ddmd/mars.d
@@ -2156,7 +2156,10 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                         error("-run must be followed by a source file, not '%s'", arguments[i + 1]);
                         break;
                     }
-                    files.push(arguments[i + 1]);
+                    if (strcmp(arguments[i + 1], "-") == 0)
+                        files.push("__stdin.d");
+                    else
+                        files.push(arguments[i + 1]);
                     params.runargs.setDim(length - 1);
                     for (size_t j = 0; j < length - 1; ++j)
                     {
--- a/test/runnable/test18076.sh
+++ b/test/runnable/test18076.sh
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+dir=${RESULTS_DIR}${SEP}runnable
+output_file=${dir}/test18076.sh.out
+
+echo 'import std.stdio; void main() { writeln("Success"); }' | \
+	$DMD -m${MODEL} -run - > ${output_file} || exit 1
+grep -q '^Success$' ${output_file} || exit 1

### CODE REVIEW

File: test/runnable/test18076.sh [hunk 1]
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+src=runnable${SEP}extra-files
+dir=${RESULTS_DIR}${SEP}runnable
+output_file=${dir}/test18076.sh.out
+
+echo 'import std.stdio; void main() { writeln("Success"); }' | \
+	$DMD -m${MODEL} -run - > ${output_file} || exit 1
```
wilzbach: Don't you have to check that it actually prints "Success" here?

```
[[ $(echo "Success") == "Success" ]] || exit 1
```

test/runnable/test18076.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+src=runnable${SEP}extra-files
```
wilzbach: not needed?

test/runnable/test18076.sh
```diff
@@ -0,0 +1,9 @@
...
+dir=${RESULTS_DIR}${SEP}runnable
+output_file=${dir}/test18076.sh.out
+
+echo 'import std.stdio; void main() { writeln("Success"); }' | \
+	$DMD -m${MODEL} -run - > ${output_file} || exit 1
+
```
wilzbach: Remove the empty, final line?

wilzbach (commented):
Wow. That's a simple fix! A few comments regarding the test case.

test/runnable/test18076.sh
quickfur: You're right, fixed.

test/runnable/test18076.sh
quickfur: Deleted, thanks!

test/runnable/test18076.sh
quickfur: Done.

test/runnable/test18076.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+dir=${RESULTS_DIR}${SEP}runnable
+output_file=${dir}/test18076.sh.out
+
+echo 'import std.stdio; void main() { writeln("Success"); }' | \
+	$DMD -m${MODEL} -run - > ${output_file} || exit 1
+grep -q '^Success$' ${output_file} || exit 1
```
wilzbach: FWIW the `|| exit` is superfluous here.
A bash script will yield the last return value as exit code if not explicitly exited

```bash
sh -c "grep -q 'foo' afile.d" ; echo $? 
```

wilzbach (approved):

PetarKirov (approved):
I really like the new functionality and the changes LGTM too.

test/runnable/test18076.sh
quickfur: True.  I'm wary of relying on these implicit behaviours, though.  Stating `|| exit 1` makes it more obvious what's happening.  Not that it makes a difference, I just prefer to have things up-front than implicit. Implicit things tend to be fragile and unintentionally break later.

wilzbach: That's why one usually does a `set -e` at the beginning: https://unix.stackexchange.com/questions/15998/what-does-the-e-do-in-a-bash-shebang :wink:

src/ddmd/mars.d
```diff
@@ -2156,7 +2156,10 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                         error("-run must be followed by a source file, not '%s'", arguments[i + 1]);
                         break;
                     }
-                    files.push(arguments[i + 1]);
+                    if (strcmp(arguments[i + 1], "-") == 0)
```
wilzbach: This is too simplistic, it won't work if there's a command after `-run`, e.g. `dmd -run -m64 -`

wilzbach (dismissed):

src/ddmd/mars.d
quickfur: Actually, `-run` takes an argument, so your example is invalid.  (I'm not sure why whoever originally implemented `-run` did it this way, but that's how it currently works, and this PR is just following in its steps.)

wilzbach: Oh sorry. I just looked over this again from my phone and imagined that doing `dmd -run -version=Foo myfile.d` is legimate, so yeah that seems like bad design too me too, but that's an opportunity for another PR ;-)

PR #7438, opened 2017-12-14
Make the Visitor class inherit the ParseTimeVisitor

RazvanN7:
@ibuclaw Is there anything else that should be done in order to make the visitor compatible with gdc? Should the header files be updated so that the inheritance is visible?

### DIFF
--- a/src/dmd/visitor.d
+++ b/src/dmd/visitor.d
@@ -19,115 +19,112 @@ import dmd.root.rootobject;
 
 // Online documentation: https://dlang.org/phobos/dmd_visitor.html
 
-
-/** Visitor instantianted with the code generation AST family
+/**
+ * Classic Visitor class which implements visit methods for all the AST
+ * nodes present in the compiler. The visit methods for AST nodes
+ * created at parse time are inherited while the visiting methods
+ * for AST nodes created at semantic time are implemented.
  */
-alias Visitor = GenericVisitor!ASTCodegen;
-
-// Generic visitor which implements a visit method for all the AST nodes.
-private extern (C++) class GenericVisitor(AST) : ParseTimeVisitor!AST
+extern (C++) class Visitor : ParseTimeVisitor!ASTCodegen
 {
-    alias visit = ParseTimeVisitor!AST.visit;
+    alias visit = ParseTimeVisitor!ASTCodegen.visit;
 public:
-    void visit(AST.ErrorStatement s) { visit(cast(AST.Statement)s); }
-    void visit(AST.PeelStatement s) { visit(cast(AST.Statement)s); }
-    void visit(AST.UnrolledLoopStatement s) { visit(cast(AST.Statement)s); }
-    void visit(AST.SwitchErrorStatement s) { visit(cast(AST.Statement)s); }
-    void visit(AST.DebugStatement s) { visit(cast(AST.Statement)s); }
-    void visit(AST.DtorExpStatement s) { visit(cast(AST.ExpStatement)s); }
-    void visit(AST.ForwardingStatement s)
+    void visit(ASTCodegen.ErrorStatement s) { visit(cast(ASTCodegen.Statement)s); }
+    void visit(ASTCodegen.PeelStatement s) { visit(cast(ASTCodegen.Statement)s); }
+    void visit(ASTCodegen.UnrolledLoopStatement s) { visit(cast(ASTCodegen.Statement)s); }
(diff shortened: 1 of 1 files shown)


andralex:
Eh, that's a fair amount of duplication, but I guess it'll do. @RazvanN7 what's with the unittest failures? @ibuclaw does this float your boat?

RazvanN7:
@andralex I don't know what duplication you are talking about.
Those must have been spurious failures since everything seems to be green now.

ZombineDev:
Also CC @kinke @JohanEngelen and @klickverbot.

ibuclaw:
I think that we should hold this off until some tests are in place. Otherwise you're just shooting in the dark here.

RazvanN7:
@ibuclaw Indeed, but the last PR made it so that the Visitor is just an alias to a templated class. This PR makes the visitor an extern C++ class again so it can't make things worse.

andralex:
@ibuclaw I'll merge this under the assumption it'll make things better even before having added tests.

andralex (approved):

ibuclaw (approved):
@andralex @RazvanN7 - OK I can confirm :+1: on both accounts.

First: I was correct in assuming that this broke C++.
Second: This patch fixes it.

ibuclaw:
Here's a small test:
```
#include "expression.h"
#include "globals.h"
#include "module.h"
#include "mtype.h"
#include "objc.h"
#include "target.h"
#include "visitor.h"

extern "C" int rt_init();
extern "C" void gc_disable();

void initialize()
{
    rt_init();
    gc_disable();

    global._init();
    global.params.isLinux = true;

    Type::_init();
    Module::_init();
    Expression::_init();
    Objc::_init();
    Target::_init();
}

class TestVisitor : public Visitor
{
  public:
    void visit(Expression *)
    {
        printf("Success!\n");
    }
};

void test_visitors()
{
    TestVisitor tv;
    IntegerExp *ie = IntegerExp::create(Loc(), 42, Type::tint32);
    ie->accept(&tv);
}

int main(int argc, char **argv)
{
    initialize();

    test_visitors();
}
```

ibuclaw:
Without this patch:
```
$ ../generated/linux/release/64/dmd-cxx
dmd-cxx: dmd/visitor.h:374: virtual void Visitor::visit(Dsymbol*): Assertion `0' failed.
Aborted
```
Backtrace:
```
#0  0x00007ffff6d75428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007ffff6d7702a in __GI_abort () at abort.c:89
#2  0x00007ffff6d6dbd7 in __assert_fail_base (fmt=<optimized out>, assertion=assertion@entry=0x79410e "0", file=file@entry=0x794100 "dmd/
visitor.h", line=line@entry=374, function=function@entry=0x794220 <Visitor::visit(Dsymbol*)::__PRETTY_FUNCTION__> "virtual void Visitor::
visit(Dsymbol*)") at assert.c:92
#3  0x00007ffff6d6dc82 in __GI___assert_fail (assertion=0x79410e "0", file=0x794100 "dmd/visitor.h", line=374, function=0x794220 <Visitor
::visit(Dsymbol*)::__PRETTY_FUNCTION__> "virtual void Visitor::visit(Dsymbol*)") at assert.c:101
#4  0x00000000006a806b in Visitor::visit (Dsymbol *) at dmd/visitor.h:374
#5  0x00000000006a8820 in Visitor::visit (Declaration *) at dmd/visitor.h:417
#6  0x00000000006a88f0 in Visitor::visit (VarDeclaration *) at dmd/visitor.h:421
#7  0x00000000006a898c in Visitor::visit (TypeInfoDeclaration *) at dmd/visitor.h:425
#8  0x00000000006a8a90 in Visitor::visit (TypeInfoArrayDeclaration *) at dmd/visitor.h:430
#9  0x000000000059cf1a in IntegerExp::accept(GenericVisitor<ASTCodegen>*) (this=0x7ffff7ed51e0, v=0x7fffffffdbd0) at dmd/expression.d:254
#10 0x00000000006a7330 in test_visitors () at dmd/cxxtest.c:61
```

With patch:
```
$ ../generated/linux/release/64/dmd-cxx
Success!
```

ibuclaw:
@MartinNowak - We are under freeze for release now right?  If applicable, this patch may need to be applied to stable.

PR #7441, opened 2017-12-15
Issue 9631 - Error message not using fully qualified name (part 2)

ntrel:
@JinShil Following on from #7405.

* Disambiguate incompatible type errors.
* Add "both operands are of type '%s'" error message case.

The latter saves the programmer time reading the error when there are two types with long qualification.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -4827,11 +4827,19 @@ extern (C++) abstract class BinExp : Expression
         TOK thisOp = (op == TOKquestion) ? TOKcolon : op;
         if (e1.op == TOKtype || e2.op == TOKtype)
         {
-            error("incompatible types for ((%s) %s (%s)): cannot use '%s' with types", e1.toChars(), Token.toChars(thisOp), e2.toChars(), Token.toChars(op));
+            error("incompatible types for ((%s) %s (%s)): cannot use '%s' with types",
+                e1.toChars(), Token.toChars(thisOp), e2.toChars(), Token.toChars(op));
+        }
+        else if (e1.type.equals(e2.type))
+        {
+            error("incompatible types for ((%s) %s (%s)): both operands are of type '%s'",
+                e1.toChars(), Token.toChars(thisOp), e2.toChars(), e1.type.toChars());
         }
         else
         {
-            error("incompatible types for ((%s) %s (%s)): '%s' and '%s'", e1.toChars(), Token.toChars(thisOp), e2.toChars(), e1.type.toChars(), e2.type.toChars());
+            auto ts = toAutoQualChars(e1.type, e2.type);
+            error("incompatible types for ((%s) %s (%s)): '%s' and '%s'",
+                e1.toChars(), Token.toChars(thisOp), e2.toChars(), ts[0], ts[1]);
         }
         return new ErrorExp();
     }
--- a/test/fail_compilation/bug9631.d
+++ b/test/fail_compilation/bug9631.d
@@ -1,32 +1,42 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/bug9631.d(17): Error: cannot implicitly convert expression `F()` of type `bug9631.tem!().F` to `bug9631.F`
+fail_compilation/bug9631.d(20): Error: cannot implicitly convert expression `F()` of type `bug9631.T1!().F` to `bug9631.T2!().F`
 ---
 */
 
-struct F { }
-
-template tem()
+template T1()
 {
     struct F { }
 }
 
-void convert()
+template T2()
 {
-    F x = tem!().F();
+    struct F { }
+}
+
+void main()
+{
+    T2!().F x = T1!().F();
 }
 
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/bug9631.d(31): Error: incompatible types for ((x) == (y)): 'bug9631.F' and 'bug9631.tem!().F'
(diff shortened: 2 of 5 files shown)


src/dmd/expression.d
```diff
@@ -4827,11 +4827,19 @@ extern (C++) abstract class BinExp : Expression
         TOK thisOp = (op == TOKquestion) ? TOKcolon : op;
         if (e1.op == TOKtype || e2.op == TOKtype)
         {
-            error("incompatible types for ((%s) %s (%s)): cannot use '%s' with types", e1.toChars(), Token.toChars(thisOp), e2.toChars(), Token.toChars(op));
+            error("incompatible types for ((%s) %s (%s)): cannot use '%s' with types",
+                e1.toChars(), Token.toChars(thisOp), e2.toChars(), Token.toChars(op));
+        }
+        else if (e1.type.equals(e2.type))
+        {
+            error("incompatible types for ((%s) %s (%s)): both operands are of type '%s'",
```
BBasile: I think it would be better to denote that _the operator %s cant be used for two variable of type %s_.
But maybe in another PR ?

src/dmd/expression.d
```diff
@@ -4827,11 +4827,19 @@ extern (C++) abstract class BinExp : Expression
         TOK thisOp = (op == TOKquestion) ? TOKcolon : op;
         if (e1.op == TOKtype || e2.op == TOKtype)
         {
-            error("incompatible types for ((%s) %s (%s)): cannot use '%s' with types", e1.toChars(), Token.toChars(thisOp), e2.toChars(), Token.toChars(op));
+            error("incompatible types for ((%s) %s (%s)): cannot use '%s' with types",
```
JinShil: I think we can change `((%s) %s (%s))` to `` `(%s) %s (%s)` ``.  That is, lose the outer parentheses and surround the expression in back ticks.

JinShil: Also, those single quotes should be backticks.

test/fail_compilation/fail3.d
```diff
@@ -14,21 +14,14 @@ template vector(T)
...
         return r;
     }
-
-    vec2 make2(T x, T y)
-    {
-        vec2 a;
-        a.x = x;
-        a.y = y;
-        return a;
-    }
```
JinShil: Why was this removed?

test/fail_compilation/bug9631.d
```diff
@@ -1,21 +1,32 @@
...
 }
 
-template T2()
+void convert()
 {
-    struct F { }
+    F x = tem!().F();
 }
 
-void main()
```
JinShil: Why the refactoring of this test case.  There should only be changes to the error messages.

test/fail_compilation/fail3.d
ntrel: It doesn't seem relevant to the test - should I restore it?

test/fail_compilation/bug9631.d
ntrel: I thought it was a bit simpler whilst still essentially testing the same thing. But I suppose it may be worth having both the two template version and the new changes.

src/dmd/expression.d
ntrel: I agree, but I was just following the existing format for these errors. I'll do as you suggest for the new error added in this PR, but the other ones can be fixed in a separate PR.

src/dmd/expression.d
ntrel: Reworded in f73a72b.

test/fail_compilation/bug9631.d
ntrel: Restored the two template version.

src/dmd/expression.d
```diff
@@ -4827,11 +4827,19 @@ extern (C++) abstract class BinExp : Expression
...
+                e1.toChars(), Token.toChars(thisOp), e2.toChars(), Token.toChars(op));
+        }
+        else if (e1.type.equals(e2.type))
+        {
+            error("invalid operation for `(%s) %s (%s)` with both operands of type `%s`",
```
BBasile: the error message is still bad. My first review comment was more accurate. Maybe:
> "the  `%s` operator is not supported for two operands of type `%s`  (`(%s) %s (%s)`)".

ntrel: I'm not sure there's much difference, but if necessary I'll update it.

BBasile: necessary

test/fail_compilation/fail3.d
JinShil: Yes, it does seem unnecessary, but I don't think it belongs in this PR.

ntrel: OK, restored.

ntrel:
I've reverted the rewording of the new error message, so it now just follows the existing format, and does not use backticks. The backtick change should be done in a separate PR as it affects about 8 error messages in dmd, I could do it if this gets merged soon.

src/dmd/expression.d
JinShil: This error message doesn't seem to have a test case.

JinShil (dismissed):

src/dmd/expression.d
ntrel: @JinShil What about `fail_compilation/fail54.d` and `fail_compilation/ice8511.d`?

JinShil: Sorry, I realize that error message didn't actually change.

ntrel:
Thanks for reviewing.

> The backtick change should be done in a separate PR

See #7464.

PR #7442, opened 2017-12-15
Fix issue 7184 - parse error on *(x)++

JinShil:
Revival of #7281

@WalterBright Added test for `(x)--`

@ibuclaw  `(int)++x` will result in _Error: semicolon expected, not x_


ibuclaw:
> @ibuclaw (int)++x will result in Error: semicolon expected, not x

Hmm, so this patch causes a diagnostic regression then.

JinShil:
This doesn't seem to be the right approach.  Closing.

PR #7446, opened 2017-12-16
Remove invalid file existence stat before dmd.conf

wilzbach:
dmd.conf solely depends on posix.mak and should always be regenerated if
the Makefile changes

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -518,7 +518,7 @@ endef
 export DEFAULT_DMD_CONF
 
 $G/dmd.conf: $(SRC_MAKE)
-	[ -f $@ ] || echo "$$DEFAULT_DMD_CONF" > $@
+	echo "$$DEFAULT_DMD_CONF" > $@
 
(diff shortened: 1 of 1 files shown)


Geod24:
Any reason to do this change ? Since the `dmd` rule depends on `dmd.conf`, now every time we make DMD, won't it overwrite local `dmd.conf`, even if we edited it ?

wilzbach:
As explained in the full text git commit message: `dmd.comf` only depends on `posix.mak` as that's it's origin. Only if the origin changes (i.e. has a newer timestamp), `dmd.conf` will be regenerated.
I didn't know that people are manually editing this file directly in the `generated` folder as that one gets nuked on a `make clean`.
I am sorry for breaking your use case, I was actually trying to cleanup the Makefile and not regenerating a file if it's origin changed is a bad Makefile practice.
How about preferring to copy a git-ignored, top-level `dmd.conf` (or in `src`) if existent instead of the auto-generated one? This would also solve the problem of the manual file being nuked...

PR #7448, opened 2017-12-16
Issue 9631 - Error message not using fully qualified name (part 3)

ntrel:
Disambiguate cast error messages.

@JinShil another follow up from #7405.

### DIFF
--- a/test/fail_compilation/bug9631.d
+++ b/test/fail_compilation/bug9631.d
@@ -40,3 +40,21 @@ void equal()
     auto y = tem!().S();
     bool b = x == y;
 }
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/bug9631.d(55): Error: cannot cast expression `x` of type `bug9631.S` to `bug9631.tem!().S` because of different sizes
+fail_compilation/bug9631.d(58): Error: cannot cast expression `ta` of type `bug9631.tem!().S[1]` to `bug9631.S[1]` because of different sizes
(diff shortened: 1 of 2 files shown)


JinShil (approved):

test/fail_compilation/bug9631.d
```diff
@@ -19,3 +19,28 @@ void main()
...
+---
+*/
+void test3()
+{
+    S x;
+    auto y = cast(tem!().S)x;
+
+    tem!().S[1] at;
+    S[1] af = cast(S[1])at;
+    auto st = cast(tem!().S[])af;
```
ZombineDev: `cast(tem!().S[])(af[])`

ZombineDev: Actually, you probably want to fix the error message on line 1590 as well: https://github.com/ntrel/dmd/blob/4e86c34dbb0e2032d0dd55bbc30062b9950475dc/src/dmd/dcast.d#L1590

ntrel: Using `af[]` doesn't cause an error.

The *sizes don't line up* error only happens when `if (tob.ty == Tarray && t1b.ty == Tsarray)` is hit, a few lines before, which IIUC would mean `toChars` is never the same for each type (one is a static array, one not), so a fix is redundant.

ntrel: Or this simple `toAutoQualChars` fix is redundant rather. It would be good to show the qualification for `S[1]` vs `S[]` too, but I'm not sure how to best to do that. Same goes for `S()` and `S`, expression vs type, `toChars` is different.

PR #7449, opened 2017-12-16
Improve test suite interface.

marler8997:
Makes the test suite a little more accommodating to developers who don't know how to use it. It now indicates that the suite is meant to be run through the Makefile (not by running d_do_test.d directly) and also prints a nice error message when environment variables are missing that would have been set by the Makefile.  This is a nice contrast to what was done before which would have been to fail at some later point in the test with a cryptic error message as a result of some environment variable being empty.

### DIFF
--- a/test/d_do_test.d
+++ b/test/d_do_test.d
@@ -18,6 +18,9 @@ void usage()
 {
     write("d_do_test <input_dir> <test_name> <test_extension>\n"
           ~ "\n"
+          ~ "   Note: this program is normally called through the Makefile, it"
+          ~ "         is not meant to be called directly by the user.\n"
+          ~ "\n"
           ~ "   input_dir: one of: compilable, fail_compilation, runnable\n"
           ~ "   test_name: basename of test case to run\n"
           ~ "   test_extension: one of: d, html, or sh\n"
@@ -27,15 +30,16 @@ void usage()
           ~ "   relevant environment variables:\n"
           ~ "      ARGS:          set to execute all combinations of\n"
           ~ "      REQUIRED_ARGS: arguments always passed to the compiler\n"
-          ~ "      DMD:           compiler to use, ex: ../src/dmd\n"
+          ~ "      DMD:           compiler to use, ex: ../src/dmd (required)\n"
           ~ "      CC:            C++ compiler to use, ex: dmc, g++\n"
           ~ "      OS:            win32, win64, linux, freebsd, osx, netbsd\n"
           ~ "      RESULTS_DIR:   base directory for test results\n"
+          ~ "      MODEL:         32 or 64 (required)\n"
           ~ "   windows vs non-windows portability env vars:\n"
           ~ "      DSEP:          \\\\ or /\n"
-          ~ "      SEP:           \\ or /\n"
-          ~ "      OBJ:          .obj or .o\n"
-          ~ "      EXE:          .exe or <null>\n");
+          ~ "      SEP:           \\ or / (required)\n"
+          ~ "      OBJ:          .obj or .o (required)\n"
+          ~ "      EXE:          .exe or <null> (required)\n");
 }
 
 enum TestMode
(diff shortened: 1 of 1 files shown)


test/d_do_test.d
```diff
@@ -436,7 +440,25 @@ bool compareOutput(string output, string refoutput)
     }
 }
 
+string envGetRequired(in char[] name)
```
wilzbach: why not `string`?

test/d_do_test.d
```diff
@@ -27,15 +30,16 @@ void usage()
...
           ~ "      RESULTS_DIR:   base directory for test results\n"
+          ~ "      MODEL:         32 or 64 (required)\n"
           ~ "   windows vs non-windows portability env vars:\n"
           ~ "      DSEP:          \\\\ or /\n"
-          ~ "      SEP:           \\ or /\n"
-          ~ "      OBJ:          .obj or .o\n"
-          ~ "      EXE:          .exe or <null>\n");
+          ~ "      SEP:           \\ or / (required)\n"
+          ~ "      OBJ:          .obj or .o (required)\n"
+          ~ "      EXE:          .exe or <null> (required)\n");
```
wilzbach: we could auto-detect the default value here.
For example:

```
version(windows) string exe = ".exe";
else string exe = "";
```

test/d_do_test.d
```diff
@@ -27,15 +30,16 @@ void usage()
           ~ "   relevant environment variables:\n"
           ~ "      ARGS:          set to execute all combinations of\n"
           ~ "      REQUIRED_ARGS: arguments always passed to the compiler\n"
-          ~ "      DMD:           compiler to use, ex: ../src/dmd\n"
+          ~ "      DMD:           compiler to use, ex: ../src/dmd (required)\n"
```
wilzbach: `../src/dmd` no longer exists ;-)

wilzbach (commented):
Why not setting reasonable defaults?

test/d_do_test.d
marler8997: Setting reasonable defaults was actually my first thought.  The thinking was it might be nice to make it easier to call `d_do_test.d` on it's own without the Makefile.  However, I realized that `d_do_test.d` is not designed to run on it's own, it's really just an extension for the Makefile, and the reason it's using environment variables in the first place is because the Makefile is not just using those variables to call d_do_test, but also for other parts of it's configuration.

If we wanted to allow the developer to call `d_do_test.d` on it's own I would say let's put defaults in there, but if it's just an extension for the Makefile then I think it makes more sense to just verify the interface between the Makefile and d_do_test (make sure required variables are given).

test/d_do_test.d
marler8997: `const char[]` conveys that the contents of the string do not need to be preserved. If you look at `environment.get` you'll see it also uses `in char[]`.  A function that accepts `string` implies that it requires the contents of that string to remain unchanged for the life of the program, however, the `environment.get` function, and this function do not have this requirement.

test/d_do_test.d
marler8997: Yes I believe you just recently changed that because of the `ddmd` to `dmd` change :)  A nice change by the way.

test/d_do_test.d
```diff
@@ -17,6 +17,9 @@ import core.sys.posix.sys.wait;
 void usage()
 {
     write("d_do_test <input_dir> <test_name> <test_extension>\n"
+          ~ "\n"
+          ~ "   Note: this program is normally called through the Makefile, it"
+          ~ "         is not meant to be called directly by the user\n"
```
wilzbach: Nitpick: end sentences with a dot.

test/d_do_test.d
wilzbach: Well in this case you only call this function with pre-defined strings which are immutable by definition ;-)

test/d_do_test.d
```diff
@@ -436,7 +440,25 @@ bool compareOutput(string output, string refoutput)
...
+    if(value is null)
+    {
+        writefln("Error: missing environment variable '%s', was this called this through the Makefile?",
+            name);
+        throw new SilentQuit();
+    }
+    return value;
+}
+
+class SilentQuit : Exception { this() { super(null); } }
```
wilzbach: nit: new line

test/d_do_test.d
```diff
@@ -436,7 +440,25 @@ bool compareOutput(string output, string refoutput)
...
 int main(string[] args)
+{
+    try { return tryMain(args); }
+    catch(SilentQuit) { return 1; }
+}
```
wilzbach: nit: newline

wilzbach (approved):
A couple of nits, but I think this is quite helpful as getting the DMD testsuite to run is really painful for newcomers...

PR #7445, opened 2017-12-16
Remove outdated src/dmd.conf target

wilzbach:
Let's see what breaks ...

PR #7451, opened 2017-12-16
remove old dmd rule

MartinNowak:
- can no longer build src/dmd target as that's a folder now


wilzbach (approved):
There is other post-transition cleanup here: https://github.com/dlang/dmd/pull/7445/files

PR #6681, opened 2017-04-09
implement DIP1008 enabling @nogc for 'throw new Throwable;'

WalterBright:
It adds the `-dip1006` to enable it. Of course, throwing that switch won't work until the druntime end is done.

The idea is to recognize `throw new Throwable` and call `_d_newThrowable()` instead of `_d_newclass()`. When caught, add a call to `_d_delThrowable()` at the end of the catch block.

The druntime will use `malloc` and `free` for these. Reference counting is used to handle rethrowing. It should mix naturally with Throwables allocated via other means, like static and the GC.

PR #7453, opened 2017-12-16
enable test case for dip1008

WalterBright:
Now that https://github.com/dlang/druntime/pull/1804 is pulled, we can enable the test case from https://github.com/dlang/dmd/pull/6681


wilzbach (approved):

test/compilable/testdip1008.d
```diff
@@ -1,5 +1,5 @@
 // PERMUTE_ARGS:
-// won't work until druntime is updated, so disable XXXEQUIRED_ARGS: -dip1008
+// REQUIRED_ARGS: -dip1008
 
 int bar()
```
Geod24: Shouldn't this be at least `@nogc` ?

PR #7454, opened 2017-12-17
Fix Issue 18013 - DMD test suite assertion failure in test_cdvecfill.d

JinShil:


test/runnable/test_cdvecfill.d
```diff
@@ -121,736 +121,1580 @@ struct Code(T_, int N_)
...
-    ]),
-    Code!(double*, 32 / double.sizeof)([
-        /* push   rax                     */ 0x50,
-        /* vbroadcastsd ymm0,QWORD PTR [rdi] */ 0xc4, 0xe2, 0x7d, 0x19, 0x07,
-        /* pop    rcx                     */ 0x59,
-        /* ret                            */ 0xc3,
-    ]),
-);
+version(D_PIC)
+{
```
wilzbach: As mentioned at https://github.com/dlang/dmd/pull/7427#discussion_r156567743 running the update script is really useful and we shouldn't break that. Hence, I think it's best to make a copy of this file for `version(D_PIC)`?

JinShil: Specifically, which update script?

wilzbach: The one which can be used to generate this output. Who did you get the `-fPIC` instructions?
Check the header of this file ;-)

JinShil: Ok, I change the update code to generate either PIC or non-PIC depending on what kind of machine it's run on.

JinShil:
cc @MartinNowak

JinShil:
btw, The diff in Github looks really screwy.  It's probably easier to just review without the diff.

test/runnable/test_cdvecfill.d
```diff
@@ -87,39 +87,883 @@ version (update)
             auto content = src.readText;
             auto f = File(src, "w");
             auto orng = f.lockingTextWriter;
-            replaceFirstInto!((_, orng) => formattedWrite(orng, "// dfmt off\n%s// dfmt on", sink.data))(orng, content, ctRegex!(`^// dfmt off[^$]*// dfmt on$`, "m"));
+            version(D_PIC)
+            {
+                replaceFirstInto!((_, orng) => formattedWrite(orng, "// dfmtpic off\n%s// dfmtpic on", sink.data))(orng, content, ctRegex!(`^// dfmtpic off[^$]*// dfmtpic on$`, "m"));
```
wilzbach: Ehm I think you need to leave `// dfmt off` as it's a known directive.
Maybe sth. like:

```
// PIC\n// dfmt off[^$]*// dfmt off
```

See also: https://github.com/dlang-community/dfmt#disabling-formatting

JinShil: 😆 I thought that was a strange moniker to use.  I've updated to code to keep the `// dfmt` directives and use different strings for the codegen.

wilzbach (approved):
LGTM. So if @MartinNowak doesn't complain in a few days, I suggest we move forward with this. After all, we want to get the testsuite running with `-fPIC` ...

ZombineDev:
I would strongly prefer if we could include this in the 2.078.0 release, as not being able to run test suite is a regression. Can you change the target of your PR to `stable`?

PetarKirov (approved):
LGTM and thanks for doing the work @JinShil. I think we should get this in as soon as possible.

MartinNowak:
Would have been sufficient to put `-fPIC` to require args and only keep one branch.

MartinNowak:
I was scratching my head like crazy trying to remember when I added this PIC code and more importantly why, see https://github.com/dlang/dmd/pull/7716/files.

PR #7455, opened 2017-12-17
fix issue 18093: [Reg 2.071] MSCOFF: dmd crashes when overriding a C+…

rainers:
…+ method in a mixin template

Latest dmd no longer builds against VC...

### DIFF
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -614,12 +614,14 @@ extern (C++) class FuncDeclaration : Declaration
      */
     final BaseClass* overrideInterface()
     {
-        ClassDeclaration cd = parent.isClassDeclaration();
-        foreach (b; cd.interfaces)
+        if (ClassDeclaration cd = parent.isClassDeclaration())
         {
-            auto v = findVtblIndex(&b.sym.vtbl, cast(int)b.sym.vtbl.dim);
-            if (v >= 0)
-                return b;
+            foreach (b; cd.interfaces)
+            {
+                auto v = findVtblIndex(&b.sym.vtbl, cast(int)b.sym.vtbl.dim);
+                if (v >= 0)
+                    return b;
+            }
         }
         return null;
     }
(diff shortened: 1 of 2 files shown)


rainers:
Should this be based on stable instead? IIRC stable is supposed to be branched off master anyway and the regression seems pretty old...

wilzbach:
> IIRC stable is supposed to be branched off master anyway

The branch-off should happen in the next few days. We can always rebase if this doesn't make it in before the branch-off.

PetarKirov (approved):

WalterBright:
It seems odd that, despite having a test case, zero lines of the patch are executed?

rainers:
> It seems odd that, despite having a test case, zero lines of the patch are executed?

The changed code is executed under Windows only, coverage is posix only.

JinShil:
> The changed code is executed under Windows only

There doesn't appear to be anything Windows-specific in the code.  Could you elaborate on why this code in only executed in Windows?

rainers:
The function `overrideInterface` is only called from https://github.com/dlang/dmd/blob/master/src/dmd/dsymbolsem.d#L4512 under condition global.params.mscoff, and from the VisualCPPMangler.

JinShil (approved):

rainers:
@MartinNowak AFAICT the next release has been branched off master in the meantime. It would be good to cherry-pick this for stable, too. Otherwise there is no released compiler that can build the win64 version of dmd.

wilzbach:
> @MartinNowak AFAICT the next release has been branched off master in the meantime. It would be good to cherry-pick this for stable, too. Otherwise there is no released compiler that can build the win64 version of dmd.

Yes and cherry-picking is pretty easy: `git cherry-pick <commt-hash>` ;-)
No need to bother Martin for this:

See: https://github.com/dlang/dmd/pull/7481

PR #7456, opened 2017-12-17
Windows: fix range error in dmd with empty configuration file

rainers:
current debug version of dmd fails to build druntime due to "-conf=" causing a range error.

This removes unnecessary conversion wchar* -> wstring -> const(wchar)*

cc @atilaneves


ibuclaw (approved):
Replacing `&p[0]` with `p` looks reasonable to me.

wilzbach (approved):
Now we just need to run CodeCov on AppVeyor to get coverage information here too.

atilaneves:
I'm not even sure right now why I wrote `&p[0]`. LGTM.

PR #7457, opened 2017-12-17
Move Compiler structure to own module.

ibuclaw:
And I propose that this should be the home of `genCmain` (which closes [issue 15374](https://issues.dlang.org/show_bug.cgi?id=15374)), as well as other free functions called by the frontend, but implemented by the driver/backend in a way that isn't compatible across compiler implementations.

### DIFF
--- a/src/dmd/compiler.d
+++ b/src/dmd/compiler.d
@@ -1,0 +1,16 @@
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Copyright:   Copyright (c) 1999-2017 by The D Language Foundation, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/compiler.d, _compiler.d)
+ */
+
+module dmd.compiler;
+
+struct Compiler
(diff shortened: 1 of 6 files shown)


JinShil:
I don't see any reason not to do this, but it could probably use more substance to motivate it.  What's preventing moving `genCmain` now?

ibuclaw:
> I don't see any reason not to do this, but it could probably use more substance to motivate it. What's preventing moving genCmain now?

If someone from core maintainers says OK, then I'll do the work and move functions across.

jacob-carlborg:
I like that idea. Should the new file be added to any makefiles?

ibuclaw:
> Should the new file be added to any makefiles?

Right, once there's actually data in the module that needs compiling. :-)

wilzbach (approved):
I am hugely in favor of this. It would be very helpful for the DMD as a library project.

RazvanN7:
I didn't know there already was a PR doing a similar thing : https://github.com/dlang/dmd/pull/7524 . Any way seems like we need this so we should merge one or the other.

ibuclaw:
> Any way seems like we need this so we should merge one or the other.

Well merge this then as it's one better. ;-)

In any case, point is that nothing in mars.d is useable in other compilers.

ibuclaw:
Added to make files, and split out into own header.

wilzbach:
> Added to make files, and split out into own header.

Thanks! Looks good! It seems that we are all unanimously in favor of this?

ibuclaw:
I hope so, so that I can start populating it.

JinShil (approved):

PR #7458, opened 2017-12-17
Fix issue 18095 - Add support for pragma(mangle) on alias declarations

jacob-carlborg:
This allows to override the mangling of a type in an alias declaration as follows:

```d
pragma(mangle, "foo") alias foo = int;
static assert(foo.mangleof == "foo");
static assert(int.mangleof == "i");
```

Overriding the mangling of a type in an alias allows for more generic solution to get the correct mangling of `c_long/c_ulong` for C++ functions. With the current solution the compiler is looking for a struct with the name `__c_long/__c_ulong` and special cases the mangling for that particular struct. Since D doesn't have implicit conversions to structs one are also forced cast/explicitly construct a `__c_long` struct when calling a function which uses the `__c_long` type as a parameter. With an alias that's not necessary anymore.

### DIFF
--- a/test/runnable/mangle.d
+++ b/test/runnable/mangle.d
@@ -589,6 +589,89 @@ void fooB(void delegate (void delegate()) scope dg)
 //pragma(msg, fooB.mangleof);
 static assert(typeof(fooA).mangleof != typeof(fooB).mangleof);
 
+/***************************************************/
+
+// test overriding mangling for alias declaration
+struct TestOverridingManglingAlias
+{
+    enum baseMangling = "6mangle27TestOverridingManglingAlias";
+
+    // basic type
+
+    // override mangling
+    pragma(mangle, "bar") alias a = int;
+    static assert(a.mangleof == "bar");
+
+    // alias without overriding
+    alias b = int;
+    static assert(b.mangleof == "i");
+
+    // default mangling of basic type
+    static assert(int.mangleof == "i");
+
+    // verify that the alias and the aliased type is the same type
+    pragma(mangle, "bar") alias c = int;
+    static assert(is(int == c));
+
+    // override mangling at intermediate step
+    alias d = int;
+    pragma(mangle, "foo") alias e = d;
+    alias f = e;
+    f g;
+    static assert(g.mangleof == "_D" ~ baseMangling ~ "1g" ~ e.mangleof);
+
+
+
+    // struct
+
+    struct Struct
+    {
+        enum defaultMangling = "S" ~ baseMangling ~"6Struct";
+    }
+
+    // override mangling
+    pragma(mangle, "Bar") alias h = Struct;
+    static assert(h.mangleof == "Bar");
+
+    // alias without overriding
+    alias i = Struct;
+    static assert(i.mangleof == Struct.defaultMangling);
+
+    // default mangling of struct
+    static assert(Struct.mangleof == Struct.defaultMangling);
+
+
+
+    // class
+    struct Class
+    {
+        enum defaultMangling = "S" ~ baseMangling ~ "5Class";
+    }
+
+    // override mangling
+    pragma(mangle, "Bar") alias j = Struct;
+    static assert(j.mangleof == "Bar");
+
+    // alias without overriding
+    alias k = Class;
+    static assert(k.mangleof == Class.defaultMangling);
+
+    // default mangling of class
+    static assert(Class.mangleof == Class.defaultMangling);
+
+
+
+    // function
+
+    // type of parameter is an alias"
+    pragma(mangle, "foo") alias l = int;
+    void m(l);
+    static assert(m.mangleof == "_D" ~ baseMangling ~ "1mMFfooZv");
+
+    // default mangling of function"
+    void n(int);
+    static assert(n.mangleof == "_D" ~ baseMangling ~ "1nMFiZv");
+}
 
 /***************************************************/
 
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -748,7 +748,12 @@ public:
         else
         {
             // For value types, strip const/immutable/shared from the head of the type
-            t.mutableOf().unSharedOf().accept(this);
+            auto unqualified = t.mutableOf().unSharedOf();
+
+            if (unqualified.mangleOverride)
+                buf.writestring(unqualified.mangleOverride);
+            else
+                unqualified.accept(this);
         }
     }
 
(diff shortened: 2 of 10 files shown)


RazvanN7 (commented):
How is the change in the backend related to this PR?

Later edit: Oh, I see, nevermind.

test/runnable/mangle.d
```diff
@@ -589,6 +589,99 @@ void fooB(void delegate (void delegate()) scope dg)
...
+
+@("test overriding mangling for alias declaration")
+struct TestOverridingManglingAlias
+{
+    enum baseMangling = "6mangle27TestOverridingManglingAlias";
+
+    @("basic type")
+    {
+        @("override mangling")
+        {
```
JinShil: What is the purpose of all of these `@("some string")` UDAs?

jacob-carlborg: It adds a description to each test indicating what is tested. If you don't like it I can remove it. I think it adds value.

JinShil: I don't see where those descriptions are being used anywhere.  How is this different from just adding a comment?

jacob-carlborg: They could be used in the future and they allow me to add a form of scope for the description.

rainers:
I like the idea and also tried to implement it when Walter introduced those special structs that don't work well without implicit conversions.

I suspect it's not so easy, though. Doesn't this reintroduce some kind of "typedef", i.e. types are compared according to their mangling so the type with mangle overwrite is no longer the same as the aliased type?
```
pragma(mangle, "foo") alias a = int;
static assert(is(a == int));
```
Could you also add a test case where the mangle override is at some intermediate step?
```
alias a = int;
pragma(mangle, "3foo") alias b = a;
alias c = b;
c x;
static assert(x.mangle == "_D4test1x3foo");
```

test/runnable/mangle.d
JinShil: I think until we have something in our infrastructure that makes use of these descriptions, they're a bit of a distraction.

jacob-carlborg: Fine 😞. I'll remove or move them to comments.

jacob-carlborg: Done.

jacob-carlborg:
@rainers Done.

test/runnable/mangle.d
jacob-carlborg: I think that now it's much more difficult to see which code belongs to which test and which description belongs to which test. In any other D project I would have used a separate unit test block for each test, but the DMD tests are not using unit test blocks.

JinShil: You can easily resolve that with some kind of comment header...

```D
/*******************************************************
*  Section 1
*******************************************************/

// Section 1.1

// Section 1.2

/*******************************************************
* Section 2
*******************************************************/

// Section 2.1

// Section 2.2
```
... or something along those lines

FWIW, I see the value in what you were proposing, but I don't think this PR is the right place to introduce such idioms into the code base unless it solves a specific problem that we're currently facing and/or has broader support/adoption.

jacob-carlborg: I think it solves specific problem we have but nobody else seems to agree with me.

jacob-carlborg:
I need to add a test for C++ mangling as well.

jacob-carlborg:
Do we have any tests for the mangling in the actual binary? `.mangleof` has a slightly different implementation.

jacob-carlborg:
Added tests for C++ mangling as well.

jacob-carlborg:
> I suspect it's not so easy, though. Doesn't this reintroduce some kind of "typedef", i.e. types are compared according to their mangling so the type with mangle overwrite is no longer the same as the aliased type?

@rainers I added a new field to the `Type` base class, `mangleOverride`, which is not used for anything but the mangled name. It's not used in any comparison of types. So no, this is not a new kind of typedef and aliases to the same type with different mangling will be considered the same type.

jacob-carlborg:
Can we move forward with this?

rainers:
> I added a new field to the Type base class, mangleOverride, which is not used for anything but the mangled name.

I noticed, but the Type.merge uses the mangling to fold multiple types into one. I'm not sure this isn't used somewhere else. It even makes the definition of what actually is an identical type non-obvious: Can a function with a different mangling have the same type?

> Can we move forward with this?

I guess it needs approval by @WalterBright or @andralex . A good example to show the merits would be to implement c_long et al. with this.

jacob-carlborg:
> I noticed, but Type.merge() uses the mangling to fold multiple types into one. I'm not sure this isn't used somewhere else.

I'm not sure I understand. I added a new field so I know it's not used anywhere else.

> It even makes the definition of what actually is an identical type non-obvious: Can a function with a different mangling have the same type?

I don't understand this question either. It's already possible today to manually set the mangling of a function.

rainers:
> I'm not sure I understand. I added a new field so I know it's not used anywhere else.

Check the implementation of `Type.merge()`: two types is considered the same if they have the same mangling, i.e. a Type object is thrown away if another identical Type object already exists. Does this imply that types are different if their mangling differs?

Example:
```
pragma(mangle, "3foo") alias foo = int;
void fun1(int x);
void fun2(foo x);
void call1(void function(int));
void call2(void function(foo));

void main()
{
   call1(&foo2);
   call2(&foo1);
}
```
Does this still work? Are the function still compatible although the mangling of the type is different? It might be ok because checks for implicit conversions are satisfied.

> It's already possible today to manually set the mangling of a function.

Yes, but not the mangling of its type.

jacob-carlborg:
> Does this still work? Are the function still compatible although the mangling of the type is different? It might be ok because checks for implicit conversions are satisfied.

@rainers the above works, but the following fails:

```d
static assert(is(void function(int) == void function(foo)))
```

I'll have a look to see if I can figure something out.

andralex:
status?

jacob-carlborg:
@andralex the status is that it wasn't so easy to implement as I thought it would be. Currently I'm not working on this. The issue is, as Rainer pointed out, that the mangling is used for type comparison as well. An alias with `pragma(mangle)` is not intended to introduce a new type.

JinShil:
I suggest this be closed if it's not being actively pursued.  Any objections?

wilzbach:
FWIW this would have been really helpful for https://github.com/dlang/dmd/pull/7893#discussion_r174813000 where we can't do `alias string = immutable(char)[]`, because the compiler doesn't know how to mangle immutable to C++.

jacob-carlborg:
> FWIW this would have been really helpful for #7893 (comment) where we can't do alias string = immutable(char)[], because the compiler doesn't know how to mangle immutable to C++.

Yes, but I haven't come up with a way to solve the type comparison problem that Rainer mentioned.

PR #7459, opened 2017-12-17
Issue 18098 Fix building 64-bit applications on Windows VS2017

skl131313:


rainers:
What problems do you see? AFAICT the x86 linker binary can cross-compile to x64 aswell.

skl131313:
If it can, it doesn't by default and DMD doesn't pass the proper arguments for it to. So it compiles a x86 binary with -m64.

rainers:
It doesn't need any arguments, the object files and libraries determine the target architecture.

skl131313:
It's changed, everything in the x86 folder is for compiling the target arch being x86. 

```
https://blogs.msdn.microsoft.com/vcblog/2016/10/07/compiler-tools-layout-in-visual-studio-15/

“Target” refers to the platform that the compiler builds applications to run on. In the VS 2015 world, amd64_arm included the compilers that were hosted on x64, and targeted ARM. In VS 2017 it’s more structured: all of the x64 hosted compilers live in a directory called HostX64. **In that directory we find x86, x64, etc., indicating the target architecture.**
```

rainers:
Yes, the paths have changed. This is important for the compiler, but the linker in HostX86\x86 still builds x64 executables if the object files are built in that architecture.

skl131313:
It's not 100%, I've come across times when it still tries to link as x86. I don't know why you are fighting this change so badly.. it should be x64 even just on the merit of consistency with the other paths.

rainers:
I'm just wondering whether there is something wrong with using the same linker for all architectures: the x86 version is also used by default by Visual D and also by the new VS auto detection in DMD. (Most settings in sc.ini are likely to be deleted soon, anyway).

skl131313:
Well it doesn't seem to happen often. I know I was having trouble building my project and was getting arch mismatch for x86/x64 and spent a while trying to figure out why. Changing the sc.ini to use the x64 folder ended up fixing the problem. Reverting the change now doesn't make the build fail anymore. I don't know how frequently it happens but I'm at least not the only person that experienced the problem.

WalterBright:
It's hard to see how the linker could behave randomly. There's likely something else in your build environment driving this.

skl131313:
I don't see what else it could be, I never had that problem building until I used DMD with VS2017. My build configuration didn't change at all, so it wasn't actually passing two mismatching objects. I saw what linker path it was using, changed it to be x64 and then it compiled without a hitch. I could have had VS open with another project, but I don't see how that would be a problem. If using the linker that's specified to be for that target is the solution to the problem, I don't know if whatever you are planning to change that would rely on using only the x86 linker. Even if it does work now (most of the time) it might not work in future builds of VS.

WalterBright:
> Reverting the change now doesn't make the build fail anymore.

This is compelling evidence that this change is not what fixed the problem for your system.

I know these kinds of heisenbugs can be very exasperating to find, but there isn't much choice in the matter.

> most of the time

I just cannot see Microsoft releasing a linker that randomly changes the memory model it is trying to link.

rainers:
I just noticed that there are less DLLs alongside the linker binaries when using the cross compiling versions, e.g. mspdb*.* are needed to link with /DEBUG. So you'll need the respective x86/x64 directory in the PATH anyway (as it is done in sc.ini right now).

In VS2013, link.exe is the same binary for all target architectures. That's not true for VS2015 and VS2017, though. 

Setting vars with `vcvarsall x86_x64` adds HostX86\x64 and HostX86\x86 to the PATH variable, so the change in this PR is closer to what MS does. I think it's ok to merge, even though it should not make a difference (and I also suspect that there is some x86/x64 mixture in your failing builds).

I'll have to update the VS auto detection, too. Probably detecting the host allowing to run 64-bit processes and choose the linker accordingly would be good.

rainers (approved):

skl131313:
> This is compelling evidence that this change is not what fixed the problem for your system.

You can say the same for changing the sc.ini file. How can you explain changing the folder path fixing the problem. What else could it have been, which is then solved by changing linkers.

If running the x64 linker was the solution, it could have overwritten a setting/config/registry that is shared with the x86 linker. Which could be the reason why the x86 linker now works after running the x64 linker.

> I just cannot see Microsoft releasing a linker that randomly changes the memory model it is trying to link.

How do you know it's random? We just don't know the cause. If they never ran the x86 linker to link with x64, then anything could be happening. It wasn't tested for such a case.

skl131313:
@rainers 
> Setting vars with vcvarsall x86_x64 adds HostX86\x64 and HostX86\x86 to the PATH variable

It can't use both though. If you do `vsvarsall x86_x64` and then do `which cl` or `which link` (windows equivalent). What path will it be using? So it's not the same thing, you are still using the correct tools even though both paths are included.

> (and I also suspect that there is some x86/x64 mixture in your failing builds).

That's the very first thing I checked, and it was compiling fine until I installed a new DMD and started using VS2017. I didn't change anything in my build, like I said changing the sc.ini file to use the X64 path solved the problem.

PR #7460, opened 2017-12-18
fixup #7447 - reintroduced Wconstant-logical-operand on older clangs

MartinNowak:
- original fix https://github.com/dlang/dmd/pull/7430
- reintroduced here https://github.com/dlang/dmd/commit/4a195a70d349bd0c5807ced4779ad7ca1edb250b#r26333985, maybe because of an outdated source file

### DIFF
--- a/src/dmd/backend/cod1.c
+++ b/src/dmd/backend/cod1.c
@@ -1321,7 +1321,7 @@ void getlvalue(CodeBuilder& cdb,code *pcs,elem *e,regm_t keepmsk)
             }
 #endif
         }
-        if (s->ty() & mTYcs && LARGECODE)
+        if (s->ty() & mTYcs && (bool) LARGECODE)
             goto Lfardata;
         goto L3;
     case FLdata:


JinShil:
> reintroduced here 4a195a7#r26333985, maybe because of an outdated source file

I'm curious why that wasn't caught by the auto-tester .

WalterBright:
I always check that what I'm working on is up to data. Obviously, I missed a spot.

wilzbach:
> I'm curious why that wasn't caught by the auto-tester .

Because `ENABLE_WARNINGS` isn't enabled. Imho this shouldn't be optional to avoid things like this.
FWIW on the D side at this Makefile `-de` and `-w` aren't optional either anymore as people kept ignoring the warnings and deprecation messages.

MartinNowak:
As in the linked original PR, this is only a warning with older clang versions, used on the 10.8 release builder, but not with the newer clang versions on the auto-tester.

PR #7464, opened 2017-12-18
[trivial] Use backticks instead of brackets in error messages

ntrel:
Change some uses of `(expression)` to use backtick quoting.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -4749,11 +4749,11 @@ extern (C++) class UnaExp : Expression
 
         if (e1.op == TOKtype)
         {
-            error("incompatible type for (%s(%s)): cannot use '%s' with types", Token.toChars(op), e1.toChars(), Token.toChars(op));
+            error("incompatible type for `%s(%s)`: cannot use '%s' with types", Token.toChars(op), e1.toChars(), Token.toChars(op));
         }
         else
         {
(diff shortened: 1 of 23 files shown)


marler8997:
The description doesn't explain what this PR is for, could you add a description?

src/dmd/expression.d
```diff
@@ -4749,11 +4749,11 @@ extern (C++) class UnaExp : Expression
 
         if (e1.op == TOKtype)
         {
-            error("incompatible type for (%s(%s)): cannot use '%s' with types", Token.toChars(op), e1.toChars(), Token.toChars(op));
+            error("incompatible type for `%s(%s)`: cannot use '%s' with types", Token.toChars(op), e1.toChars(), Token.toChars(op));
```
WalterBright: Should do the '%s' to `%s` while you're at it!

JinShil: Yes, there are several messages in this PR that use single quotes instead of backticks.  Please change them to backticks also.

ntrel: I find the brackets worse as they nest. There are many uses of '%s', I think they should be replaced separately.

JinShil: So be it.

JinShil (approved):

PR #7466, opened 2017-12-18
use make's sort to remove duplicate documentables

MartinNowak:


wilzbach:
It's probably better to remove the duplicate file (it's `id.d`). When I bumped into this last week, I was surprised that DMD doesn't error when given the same file twice...

MartinNowak:
> It's probably better to remove the duplicate file (it's id.d). When I bumped into this last week, I was surprised that DMD doesn't error when given the same file twice...

It's a lot more that just this file, the SRCS variables are meant for different targets, each of which might include the same sources.
Not sure what values is wanted for `SRC_DOCUMENTABLES`, simply all sources?

wilzbach (approved):

wilzbach:
> Not sure what values is wanted for SRC_DOCUMENTABLES, simply all sources?

Not sure either, but I think so.

wilzbach:
Cherry-picked this to master (https://github.com/dlang/dmd/pull/7478), s.t. we don't see the ugly warnings there either ...

MartinNowak:
> Cherry-picked this to master (#7478), s.t. we don't see the ugly warnings there either ...

@wilzbach It would be nicer to avoid cherry-picks. They just puts more workload onto me, having to resolve unnecessary merge conflicts. Feel free to open merge_stable PR *anytime* though.
I know some people have an aversion against merge commits, but that's an aesthetic preference, while preserving proper history information is essential for automation.

```sh
git fetch upstream
git checkout -B merge_stable upstream/master
git merge --no-ff upstream/stable
git hub pull new -b master
```

MartinNowak:
Maybe it's time to automate the merging of stable into master?
Might be a bit tricky to implement with flaky CI and review required though.

PR #7467, opened 2017-12-19
(Option 1) Cause win32.mak to print error when HOST_DC is not set

marler8997:
I spent way too much time on this one.  The limitations of Digital Mars' `make` program made this one quite difficult, this is the best mechanism I could come up with to cause the make file to fail with a nice error messages when `HOST_DC` is not set.

### DIFF
--- a/src/win32.mak
+++ b/src/win32.mak
@@ -288,10 +288,17 @@ release:
 $G :
 	if not exist "$G" mkdir $G
 
-debdmd:
+check-host-dc:
+	@cmd /c if "$(HOST_DC)" == "" (echo Error: Environment variable HOST_DC is not set & exit 1)
+
+debdmd: check-host-dc debdmd-make
+
+debdmd-make:
 	$(DMDMAKE) "OPT=" "DEBUG=-D -g -DUNITTEST" "DDEBUG=-debug -g -unittest" "DOPT=" "LFLAGS=-L/ma/co/la" $(TARGETEXE)
 
-reldmd:
+reldmd: check-host-dc reldmd-make
+
+reldmd-make:
 	$(DMDMAKE) "OPT=-o" "DEBUG=" "DDEBUG=" "DOPT=-O -release -inline" "LFLAGS=-L/delexe/la" $(TARGETEXE)
 
 profile:


JinShil:
See also #7256

marler8997:
I wanted to point out that a better solution would have been to create a target that verifies the existence of `HOST_DC`, then either sets a default when it doesn't exist, or exits the build with a nice error message.  i.e.
```
sometarget: default-host-dc do-stuff ...

default-host-dc:
    if "$(HOST_DC)" == "" ( set HOST_DC=dmd )
```
OR
```
sometarget: check-host-dc do-stuff ...

check-host-dc:
    if "$(HOST_DC)" == "" ( echo Error: HOST_DC is not set & exit 1)
```
Try as I might, I could not get either of these to work.  As far as I can tell, Digital Mars Make completely ignores any results from the shell commands, even the exit status.  This is odd since the documentation refers to options that allow you to "ignore status" of shell commands, however, I couldn't find any of these "ignore" mechanisms in our Makefile so it must be a bug.

marler8997:
Had to make another change, it appears to be another bug in digital mars make,   I had to split the `debdmd` and `reldmd` targets into one with their dependencies and one with their rules/actions.  If both the dependencies and actions are in the same staement, then make will not execute the actions.

andralex (commented):
I'm unclear about how this works. I'd expect it would be simpler, what's with all the additional work? What does "deb" and "rel" mean? I assume "debug" and "release"? It was confusing as heck because "deb" also stands in for "debian".

marler8997:
I'd expect it to be simpler as well, but this was the simplest I could figure out how to do.  I believe my changes along with the existing complexity in the Makefile is due to limitations and bugs in Digital Mars make.

JinShil:
> What does "deb" and "rel" mean?

Indeed.  Please spell things out.

marler8997:
`debdmd` and `reldmd` are public targets apart of the public interface of the Makefile.  I'm not against changing but I would consider those changes outside of the scope of this PR.  This PR only addresses proper error handling of the HOST_DC environment variable.

src/win32.mak
```diff
@@ -288,10 +288,18 @@ release:
 $G :
 	if not exist "$G" mkdir $G
```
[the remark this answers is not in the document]
marler8997: I believe the stack overflow question you submitted is an issue with GNU make. This makefile uses Digital Mars make and I haven't seen an issue with how it handles directory dependencies.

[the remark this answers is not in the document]
wilzbach: @andralex as I have stated on other pull requests in the past: depending on directories works _only if_ you use them as [order-only dependencies](https://www.gnu.org/software/make/manual/html_node/Prerequisite-Types.html).
Anyhow DigitalMars Make so different from Make, that it shouldn't be called Make :wink:

wilzbach (approved):
I prefer this approach over Option 2, it's a bit more DRY.

src/win32.mak
```diff
@@ -288,10 +288,18 @@ release:
...
 
-debdmd:
+check-host-dc:
+	@cmd /c if "$(HOST_DC)" == "" (echo Error: Environment variable HOST_DC is not set & exit 1)
+
+debdmd: check-host-dc debdmd-make
+
+debdmd-make:
+	@echo reldmd
```
wilzbach: Why is this line needed?

marler8997: I assume you are asking why have the extra target, why not do this:
```Make
debdmd: check-host-dc 
    $(DMDMAKE) "OPT=" "DEBUG=-D -g -DUNITTEST" "DDEBUG=-debug -g -unittest" "DOPT=" "LFLAGS=-L/ma/co/la" $(TARGETEXE)
```
Instead of this:
```Make
debdmd: check-host-dc debdmd-make

debdmd-make:
    $(DMDMAKE) "OPT=" "DEBUG=-D -g -DUNITTEST" "DDEBUG=-debug -g -unittest" "DOPT=" "LFLAGS=-L/ma/co/la" $(TARGETEXE)
```

The answer is because it doesn't work :)  I don't know why, I'm guess it's a bug in Digital Mars Make, but if you combine them into 1 statement, them make ignores the commands.

wilzbach: Oh I was trying to politely ask you to remove the debugging output ;-)
I understand that DigitalMars Make is "rather rudimentary" ...

JinShil (approved):

PR #7469, opened 2017-12-19
chsched.c refactor: #defines => enums, loop variables

WalterBright:


src/dmd/backend/cgsched.c
```diff
@@ -2317,9 +2327,8 @@ int Schedule::insert(Cinfo *ci)
     ic = -1;
     imin = 0;
     for (i = tblmax; i >= 0; i--)
```
wilzbach: You could have moved the declaration of `i` down here too (as you did with the others) :wink:

wilzbach (approved):

src/dmd/backend/cgsched.c
WalterBright: no, it's used outside the loop

PR #7470, opened 2017-12-19
Enable Codecov 'unexpected changes' + threshold

wilzbach:
Experimenting with this a bit. The idea is to avoid showing a red cross for a PR if all tests passed and there was just flaky coverage changes (IIRC CodeCov still doesn't handle well, that we merge the PR into master before running the testsuite).

See also: https://docs.codecov.io/docs/commit-status

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,10 +1,6 @@
-# Documentation: https://github.com/codecov/support/wiki/Codecov-Yaml
+# Documentation: https://docs.codecov.io/docs/codecov-yaml
 
 codecov:
-  # CodeCov should only wait for CI's that run coverage tests
-  # DAutoTester and auto-tester are not in the default list
-  ci:
-    - "circleci.com"
   notify:
     after_n_builds: 1  # send notifications after the first upload
   bot: dlang-bot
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -18,7 +18,9 @@ coverage:
     # Learn more at https://docs.codecov.io/docs/commit-status
     project: true
     patch: true
-    changes: false
+    changes: true
+    # Avoids showing a red cross for most PRs
```
JinShil: 👍

wilzbach: Well that's the theory. I'm not sure whether it will actually work as expected ...

JinShil (approved):

wilzbach:
> Regarding the WIP label: How do we test this, without just merging the PR and reverting later if necessary?

I assumed CodeCov would use the YAML present in this PR, but apparently I was wrong. However, according to their documentation we can create a new branch:

https://docs.codecov.io/docs/codecov-yaml#section-restricting-changes

I will do an experiment on this tomorrow.

wilzbach:
Argh it doesn't seem to work well:

image


https://github.com/wilzbach/phobos/pull/13

and https://github.com/wilzbach/phobos/blob/test-codecov/.codecov.yml

stevepeak:
Please use this yaml format:

```yaml
# Documentation: https://docs.codecov.io/docs/codecov-yaml

codecov:
  notify:
    after_n_builds: 1  # send notifications after the first upload
  bot: dlang-bot

coverage:
  precision: 3
  round: down
  range: "80...100"

  # Learn more at https://docs.codecov.io/docs/commit-status
  status:
    project:
      default:
        informational: true
    patch:
      default:
        informational: true

comment: false
```

wilzbach:
Works like magic @stevepeak - thank you so much!

image

Test: https://github.com/wilzbach/phobos/pull/13

.codecov.yml
```diff
@@ -1,24 +1,30 @@
...
   status:
-    # Learn more at https://docs.codecov.io/docs/commit-status
-    project: true
-    patch: true
-    changes: false
+    project:
+      default:
+        informational: true
+        # Informational doesn't work with project yet
+        threshold: 0.1
```
wilzbach: I think 0.1% is a reasonable threshold to start with. If it's too low, we can always increase it?

PetarKirov (approved):

PR #5682, opened 2016-04-18
Improve FuncExp: Do nothing if type is already set in FuncExp.semantic

9rnsr:
It was a hack when I started to implement complete template lambda behavior, but
it's already replaced with the `FuncExp.matchType` call from `dcast.d`, so
we can follow the usual pattern in `Expression.semantic`.

PR #7472, opened 2017-12-20
Improve FuncExp: Do nothing if type is already set in semantic

JinShil:
Revival of #5682, to give it one last chance before closing.

@9rnsr's comments in the original PR justifying this change. 

> It was a hack when I started to implement complete template lambda behavior, but it's already replaced with the `FuncExp.matchType` call from `dcast.d`, so we can follow the usual pattern in `Expression.semantic`.

> In my first implementation of "template lambda in everywhere", `FuncExp.semantic` could be called twice.

> If `td` is not `null` in the first semantic call, `type` is set to `Type::tvoid`. And then from `cast.c` `FuncExp.semantic` was called again, to instantiated `td` with the required parameter types which were picked from the "cast to" type.

> But today, the post-semantic process is handled in `FuncExp.matchType`, and the `type == Type.tvoid` check is not used anymore. That's why the check can be removed now.

I revived this because I think, before we can determine whether or not this should be closed, we should see how it does against the current implementation of the compiler and the test suite.

This PR will not remain open for long.  If it is not reviewed and/or accepted in a timely manner, or the team deems it not valuable, I will close it.

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -2551,102 +2551,107 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             if (exp.fd.treq)
                 printf("  treq = %s\n", exp.fd.treq.toChars());
         }
+
+        if (exp.type)
+        {
+            result = exp;
+            return;
+        }
+
         Expression e = exp;
+        uint olderrors;
 
         sc = sc.push(); // just create new scope
         sc.flags &= ~SCOPEctfe; // temporary stop CTFE
         sc.protection = Prot(PROTpublic); // https://issues.dlang.org/show_bug.cgi?id=12506
 
-        if (!exp.type || exp.type == Type.tvoid)
+        /* fd.treq might be incomplete type,
+            * so should not semantic it.
+            * void foo(T)(T delegate(int) dg){}
+            * foo(a=>a); // in IFTI, treq == T delegate(int)
+            */
+        //if (fd.treq)
+        //    fd.treq = fd.treq.dsymbolSemantic(loc, sc);
+
+        exp.genIdent(sc);
+
+        // Set target of return type inference
+        if (exp.fd.treq && !exp.fd.type.nextOf())
         {
-            /* fd.treq might be incomplete type,
-             * so should not semantic it.
-             * void foo(T)(T delegate(int) dg){}
-             * foo(a=>a); // in IFTI, treq == T delegate(int)
-             */
-            //if (fd.treq)
-            //    fd.treq = fd.treq.dsymbolSemantic(loc, sc);
-
-            exp.genIdent(sc);
-
-            // Set target of return type inference
-            if (exp.fd.treq && !exp.fd.type.nextOf())
+            TypeFunction tfv = null;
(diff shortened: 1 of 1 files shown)


src/dmd/expressionsem.d
```diff
@@ -2551,14 +2551,20 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
         }
+
+        if (exp.type)
+        {
+            result = exp;
+            return;
+        }
+
         Expression e = exp;
+        uint olderrors;
```
JinShil: `olderrors` was moved here to resolve this error

```
goto skips declaration of variable dmd.expressionsem.ExpressionSemanticVisitor.visit.olderrors
```

src/dmd/expressionsem.d
```diff
@@ -2551,14 +2551,20 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             if (exp.fd.treq)
                 printf("  treq = %s\n", exp.fd.treq.toChars());
         }
+
+        if (exp.type)
```
RazvanN7: Shouldn't this be : 

```d
if (exp.type && exp.type != Type.tvoid)
```

?

JinShil: According to @9rnsr, no.  He addressed the check with this comment.

> But today, the post-semantic process is handled in `FuncExp.matchType`, and the `type == Type.tvoid` check is not used anymore. That's why the check can be removed now.

Can you think of a test case that would justify adding the check against `Type.tvoid`?

RazvanN7: I can't say I understand completely the reason behind the `Type.tvoid` check. Chesterton's fence...

JinShil:
This doesn't appear to have any support.  I will be closing in 14 days, unless it garners some interest.

ZombineDev:
Most probably other reviewers haven't had the time to review it yet (including me).

src/dmd/expressionsem.d
```diff
@@ -2551,14 +2551,20 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
         Expression e = exp;
+        uint olderrors;
 
         sc = sc.push(); // just create new scope
         sc.flags &= ~SCOPEctfe; // temporary stop CTFE
         sc.protection = Prot(PROTpublic); // https://issues.dlang.org/show_bug.cgi?id=12506
 
-        if (!exp.type || exp.type == Type.tvoid)
-        {
             /* fd.treq might be incomplete type,
```
ibuclaw: The indentation should be fixed. (reviewers can use `?w=1` to hide whitespace changes.

ibuclaw:
I think this is fine, just needs whitespace corrections as per comment.

JinShil:
> I think this is fine, just needs whitespace corrections as per comment.

That was actually intentional on my part.  If I correct the whitespace, it turns the diff into a colossal mess, making it very difficult to review.  Since you've voiced your approval, I'll fix the whitespace. in a final commit.

ibuclaw:
> it turns the diff into a colossal mess,

That's what `w=1` is for. 😉

JinShil:
Fixed whitespace.  Best reviewed with https://github.com/dlang/dmd/pull/7472/files?w=1

PR #7473, opened 2017-12-20
[Makefile] Actually use release flags on BUILD=release

wilzbach:
If we already default to `BUILD=release`, we should actually do so. The README at the top states:

> ENABLE_RELEASE:		Optimized release built (set by BUILD=release)

Before:

```
CC="c++" dmd -of../generated/linux/release/64/dmd -m64 -vtls -J../generated/linux/release/64 -J../res -L-lstdc++ -version=MARS -fPIC -w -de 
```

After:

```
CC="c++" dmd -of../generated/linux/release/64/dmd -m64 -vtls -J../generated/linux/release/64 -J../res -L-lstdc++ -version=MARS -fPIC -w -de -O -release -inline 
```

FYI: I'm not a huge fan of using the release mode by default, but that's a battle for another day.

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -59,6 +59,8 @@ ifneq ($(BUILD),release)
         $(error Unrecognized BUILD=$(BUILD), must be 'debug' or 'release')
     endif
     ENABLE_DEBUG := 1
+else
+    ENABLE_RELEASE := 1
 endif
 
 # default to PIC on x86_64, use PIC=1/0 to en-/disable PIC.
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -474,7 +474,7 @@ $G/parser.a: $(PARSER_SRCS) $(LEXER_SRCS) $(ROOT_SRCS) $G/dmd $G/dmd.conf $(SRC_
 	CC="$(HOST_CXX)" $G/dmd -lib -of$@ $(MODEL_FLAG) -L-lstdc++ -J$G $(DFLAGS) $(PARSER_SRCS) $(LEXER_SRCS) $(ROOT_SRCS)
 
 $G/examples/%: $(EX)/%.d $G/parser.a $G/dmd
-	CC="$(HOST_CXX)" $G/dmd -of$@ $(MODEL_FLAG) $(DFLAGS) $G/parser.a $<
+	CC="$(HOST_CXX)" $G/dmd -of$@ $(MODEL_FLAG) $(DFLAGS) -J$G $G/parser.a  $<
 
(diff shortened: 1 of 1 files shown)


wilzbach:
```
CC="c++" ../generated/linux/release/32/dmd -lib -of../generated/linux/release/32/parser.a -m32 -L-lstdc++ -J../generated/linux/release/32 -version=MARS  -w -de -O -release -inline dmd/parse.d dmd/astbase.d dmd/parsetimevisitor.d dmd/transitivevisitor.d dmd/permissivevisitor.d dmd/strictvisitor.d dmd/console.d dmd/entity.d dmd/errors.d dmd/globals.d dmd/id.d dmd/identifier.d dmd/lexer.d dmd/tokens.d dmd/utf.d dmd/root/aav.d dmd/root/array.d dmd/root/ctfloat.d dmd/root/file.d dmd/root/filename.d dmd/root/man.d dmd/root/outbuffer.d dmd/root/port.d dmd/root/response.d dmd/root/rmem.d dmd/root/rootobject.d dmd/root/speller.d dmd/root/stringtable.d dmd/root/hash.d
CC="c++" ../generated/linux/release/32/dmd -of../generated/linux/release/32/examples/avg -m32 -version=MARS  -w -de -O -release -inline ../generated/linux/release/32/parser.a examples/avg.d
CC="c++" ../generated/linux/release/32/dmd -of../generated/linux/release/32/examples/impvisitor -m32 -version=MARS  -w -de -O -release -inline ../generated/linux/release/32/parser.a examples/impvisitor.d
dmd/globals.d(361): Error: need -Jpath switch to import text file VERSION
posix.mak:479: recipe for target '../generated/linux/release/32/examples/impvisitor' failed
make[1]: *** [../generated/linux/release/32/examples/impvisitor] Error 1
dmd/globals.d(361): Error: need -Jpath switch to import text file VERSION
```

Pretty interesting that this didn't show up before. `ENABLE_RELEASE` just _adds_  `-O -inline -release` to the `DFLAGS` ...

WalterBright:
Pls address the various failures.

wilzbach:
Hmm, I'm no longer convinced of this change as it will make the default build of DMD quite slow.

I think it would be better if
a) we switch to `BUILD=debug` being the default (a big switch, probably not worth the cost)
b) make the "default" build the "debug build" by adding ENABLE_DEBUG=1 by default

Both approaches have their downsides and this isn't pressing really and hopefully all downstream packers use `ENABLE_RELEASE=1` (and `HOST_DMD=ldmd`), s.t. doesn't become a problem.
-> closing

PR #7474, opened 2017-12-20
Fix typo in the ModuleInfo changelog

wilzbach:
From https://github.com/dlang/dlang.org/pull/2021

We really should start to generate the changelog incrementally :/


wilzbach:
Force-merging as Jenkins is still broken.

PR #7475, opened 2017-12-20
 Enable C++ warnings by default

wilzbach:
To avoid regressions in the future. 

On the D side this is successfully done with `-w -de`.

What was the motivation for not doing so? Too many warnings?

### DIFF
--- a/src/dmd/backend/blockopt.c
+++ b/src/dmd/backend/blockopt.c
@@ -727,7 +727,7 @@ void brcombine()
                     {
                         if (EOP(b3->Belem))
                             continue;
-                        tym_t ty = (bc2 == BCretexp) ? b2->Belem->Ety : TYvoid;
+                        tym_t ty = (bc2 == BCretexp) ? b2->Belem->Ety : static_cast<regm_t>(TYvoid);
                         e = el_bin(OPcolon2,ty,b2->Belem,b3->Belem);
                         b->Belem = el_bin(OPcond,ty,b->Belem,e);
                     }
--- a/src/dmd/backend/cgcod.c
+++ b/src/dmd/backend/cgcod.c
@@ -595,7 +595,7 @@ tryagain:
 
     // Mask of regs saved
     // BUG: do interrupt functions save BP?
-    sfunc->Sregsaved = (functy == TYifunc) ? mBP : (mfuncreg | fregsaved);
+    sfunc->Sregsaved = (functy == TYifunc) ? static_cast<regm_t>(mBP) : (mfuncreg | fregsaved);
 
     util_free(csextab);
     csextab = NULL;
@@ -2365,10 +2365,11 @@ static void comsub(CodeBuilder& cdb,elem *e,regm_t *pretregs)
   /* create mask of what's in csextab[] */
   csemask = 0;
   for (size_t i = 0; i < cstop; i++)
-  {     if (csextab[i].e)
-            elem_debug(csextab[i].e);
-        if (csextab[i].e == e)
-                csemask |= csextab[i].regm;
+  {
+    if (csextab[i].e)
+        elem_debug(csextab[i].e);
+    if (csextab[i].e == e)
+        csemask |= csextab[i].regm;
   }
   csemask &= ~emask;            /* stuff already in registers   */
 
--- a/src/dmd/backend/cgcod.c
+++ b/src/dmd/backend/cgcod.c
@@ -237,6 +237,7 @@ tryagain:
             {   case SCfastpar:
                 case SCshadowreg:
                     regcon.params |= s->Spregm();
+                    /* FALL-THROUGH */
                 case SCparameter:
                     if (s->Sfl == FLreg)
                         noparams |= s->Sregm;
@@ -337,6 +338,7 @@ tryagain:
                     cdb.genlinnum(b->Bsrcpos);
                     b->Bcode = cdb.finish();
                 }
+                /* FALL-THROUGH */
             case BCretexp:
                 epilog(b);
                 break;
@@ -2677,6 +2679,7 @@ void codelem(CodeBuilder& cdb,elem *e,regm_t *pretregs,bool constflag)
             else
                 *pretregs &= mPSW | s->Sregm;
         }
+        /* FALL-THROUGH */
     case OPconst:
         if (*pretregs == 0 && (e->Ecount >= 3 || e->Ety & mTYvolatile))
         {
--- a/src/dmd/backend/cod3.c
+++ b/src/dmd/backend/cod3.c
@@ -624,7 +624,7 @@ regm_t regmask(tym_t tym, tym_t tyf)
 
         case TYllong:
         case TYullong:
-            return I64 ? mAX : (I32 ? mDX | mAX : DOUBLEREGS);
+            return I64 ? static_cast<regm_t>(mAX) : (I32 ? static_cast<regm_t>(mDX | mAX) : DOUBLEREGS);
 
         case TYldouble:
         case TYildouble:
--- a/src/dmd/backend/cod3.c
+++ b/src/dmd/backend/cod3.c
@@ -592,6 +592,7 @@ regm_t regmask(tym_t tym, tym_t tyf)
                 return mXMM0;
             if (config.exe & EX_flat)
                 return mST0;
+            /* FALL-THROUGH */
         case TYlong:
         case TYulong:
         case TYdchar:
@@ -635,9 +636,11 @@ regm_t regmask(tym_t tym, tym_t tyf)
             if (I32 && tybasic(tyf) == TYnfunc)
                 return mDX | mAX;
 #endif
+            /* FALL-THROUGH */
         case TYcdouble:
             if (I64)
                 return mXMM0 | mXMM1;
+            /* FALL-THROUGH */
         case TYcldouble:
             return mST01;
 
@@ -4907,6 +4910,7 @@ void assignaddrc(code *c)
             case FLpseudo:
                 assert(0);
                 /* NOTREACHED */
+                /* FALL-THROUGH */
             case FLfast:
                 c->IEVpointer2 += s->Soffset + Fast.size + BPoff;
                 break;
@@ -5466,6 +5470,7 @@ void pinholeopt(code *c,block *b)
                     }
                     if ((op & ~0x0F) != 0x70)
                         break;
+                    /* FALL-THROUGH */
                 case JMP:
                     switch (c->IFL2)
                     {   case FLcode:
@@ -6297,6 +6302,7 @@ unsigned codout(int seg, code *c)
                         if (!(issib(rm) && (c->Isib & 7) == 5 ||
                               (rm & 7) == 5))
                             break;
+                        /* FALL-THROUGH */
                     case 0x80:
                     {   int flags = CFoff;
                         targ_size_t val = 0;
@@ -6337,6 +6343,7 @@ unsigned codout(int seg, code *c)
                     case 0:
                         if ((rm & 7) != 6)
                             break;
+                        /* FALL-THROUGH */
                     case 0x80:
                         do16bit(&ggen, (enum FL)c->IFL1,&c->IEV1,CFoff);
                         break;
@@ -6369,6 +6376,7 @@ unsigned codout(int seg, code *c)
                                 do64bit(&ggen, (enum FL)c->IFL2,&c->IEV2,flags);
                                 break;
                             }
+                            /* FALL-THROUGH */
                         case 0xA0:              /* MOV AL,byte ptr []   */
                         case 0xA2:
                             if (c->Iflags & CFaddrsize && !I64)
@@ -6461,6 +6469,7 @@ unsigned codout(int seg, code *c)
                         case CALL:
                         case JMP:
                             flags |= CFselfrel;
+                            /* FALL-THROUGH */
                         default:
                         case_default16:
                             if (c->Iflags & CFopsize)
--- a/src/dmd/backend/el.c
+++ b/src/dmd/backend/el.c
@@ -2375,7 +2375,7 @@ elem *el_ctor_dtor(elem *ec, elem *ed, elem **pedtor)
         union eve c;
         memset(&c, 0, sizeof(c));
         elem *e_flag_0 = el_bin(OPeq, TYvoid, el_var(sflag), el_const(TYbool, &c));  // __flag = 0
-        er = el_bin(OPinfo, ec ? ec->Ety : TYvoid, ector, el_combine(e_flag_0, ec));
+        er = el_bin(OPinfo, ec ? ec->Ety : static_cast<tym_t>(TYvoid), ector, el_combine(e_flag_0, ec));
 
         /* A destructor always executes code, or we wouldn't need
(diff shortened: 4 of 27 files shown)


src/dmd/backend/cgcod.c
```diff
@@ -2365,10 +2365,11 @@ static void comsub(CodeBuilder& cdb,elem *e,regm_t *pretregs)
...
   for (size_t i = 0; i < cstop; i++)
-  {     if (csextab[i].e)
-            elem_debug(csextab[i].e);
-        if (csextab[i].e == e)
-                csemask |= csextab[i].regm;
+  {
+    if (csextab[i].e)
+        elem_debug(csextab[i].e);
+    if (csextab[i].e == e)
+        csemask |= csextab[i].regm;
```
wilzbach:
```
dmd/backend/cgcod.c: In function ‘void comsub(CodeBuilder&, elem*, regm_t*)’:
dmd/backend/cgcod.c:2368:9: error: this ‘if’ clause does not guard... [-Werror=misleading-indentation]
   {     if (csextab[i].e)
         ^~
dmd/backend/cgcod.c:2370:9: note: ...this statement, but the latter is misleadingly indented as if it were guarded by the ‘if’
         if (csextab[i].e == e)
         ^~
```

src/dmd/backend/blockopt.c
```diff
@@ -727,7 +727,7 @@ void brcombine()
                     {
                         if (EOP(b3->Belem))
                             continue;
-                        tym_t ty = (bc2 == BCretexp) ? b2->Belem->Ety : TYvoid;
+                        tym_t ty = (bc2 == BCretexp) ? b2->Belem->Ety : static_cast<regm_t>(TYvoid);
```
wilzbach:
```
dmd/backend/blockopt.c: In function ‘void brcombine()’:
dmd/backend/blockopt.c:730:54: error: enumeral and non-enumeral type in conditional expression [-Werror=extra]
                         tym_t ty = (bc2 == BCretexp) ? b2->Belem->Ety : TYvoid;
                                    ~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```

`-Wno-enum-compare` didn't have any effect.

jacob-carlborg:
Clang on macOS complains about:

* A trailing else here [1] for debug builds
* Linking with `libstdc++` instead of `libc++`. To link with `libc++` one should also bump the minimum deployment target to 10.9.

[1] https://github.com/dlang/dmd/blob/master/src/dmd/backend/cgreg.c#L841-L855

ibuclaw:
> What was the motivation for not doing so? Too many warnings?

Last time I fixed some warnings, I think it was rejected because Walter didn't consider them as bugs (gdc/gcc builds the frontend with more warnings turned on than dmd).

jacob-carlborg:
Walter has recently fixed some warnings: https://github.com/dlang/dmd/pull/7358 and #7354, perhaps he has changed his mind.

ibuclaw:
@jacob-carlborg - I guess there's a difference between _pro-actively_ fixing warning because you are compiling with more warnings turned on than what's in the dmd make files vs. _reactively_ fixing warnings because your locally installed compiler now complains about it.

Walter is only doing the latter.

ibuclaw:
These are the warning flags that could be used for build, depending on configure flags passed.
https://github.com/gcc-mirror/gcc/blob/17187f6a206b467c66904749baf1dc6c9b0130d5/gcc/configure.ac#L467-L517

Would be nice if the C++ code built cleanly with them. (Default is approximately: `-W -Wall -Wno-narrowing -Wwrite-strings -Wcast-qual -Wformat -Wmissing-format-attribute -Woverloaded-virtual -pedantic -Wno-long-long -Wno-variadic-macros -Wno-overlength-strings`)

src/dmd/backend/blockopt.c
ibuclaw: > -Wno-enum-compare didn't have any effect.

That's because it seems to be turned on by `-Wextra`.   Or at least according to this part of the error `[-Werror=extra]`.

src/posix.mak
```diff
@@ -167,7 +165,10 @@ WARNINGS := -Wall -Wextra -Werror \
 	-Wno-unused-label \
 	-Wno-unused-parameter \
 	-Wno-unused-value \
-	-Wno-unused-variable
+	-Wno-unused-variable \
+	-Wno-implicit-fallthrough \
+	-Wno-enum-compare
```
ibuclaw: We should be removing disabled warnings, not adding more.

jacob-carlborg:
There's also a difference in using a recent compiler and a really old compiler. I've always had warnings when compiling DMD as far as I can remember and I have never enabled any additional warnings.

For example, since we recently started treating warnings as errors I cannot compile DMD in debug mode without patching it locally.

wilzbach:
> For example, since we recently started treating warnings as errors I cannot compile DMD in debug mode without patching it locally.

This shouldn't happen! We need to fix this, s.t. this can't happen in the future. What errors/warnings do you get?

See also: https://github.com/dlang/dmd/pull/7477

wilzbach:
> We should be removing disabled warnings, not adding more.

FYI @ibuclaw - I added `-Wno-implicit-fallthrough` to avoid adding a couple of `/* FALLTHROUGH */` comments as I wasn't sure whether this would be acceptable. I added this now as a separate commit to avoid adding another disabled warning.

> I don't think it's a good idea to enable -Werror by default without any user intervention – I've tried that before, but the difference in warnings between different compilers and compiler versions is just too large for it to be workable in an open source setting where the build environment is not controlled. 

I just realized this too. Even with a different `gcc` version on auto-tester one gets a different set of warnings :disappointed: 

> On the other hand, enabling warnings as errors by default on the CI boxes is a great idea. There will be some fallout when the various environments are updated, but that is dealt with easily enough.

Awesome! Done for CircleCi + auto-tester (Posix-only for now).

ibuclaw:
> There's also a difference in using a recent compiler and a really old compiler.

But if you fix a warning that a newer compiler emits, the older compiler should still be able to build it.  Whereas if you disable the warning, the older compiler may not understand the command line argument.

src/dmd/backend/cod3.c
```diff
@@ -591,7 +591,7 @@ regm_t regmask(tym_t tym, tym_t tyf)
             if (I64)
                 return mXMM0;
             if (config.exe & EX_flat)
-                return mST0;
+            /* FALL-THROUGH */
```
ibuclaw: Was this return removed by accident?

src/dmd/backend/el.c
```diff
@@ -2240,6 +2240,7 @@ elem *el_convert(elem *e)
                 break;
             }
 #if 1
+            /* FALL-THROUGH */
```
ibuclaw: Maybe have this before the `#if 1`

ibuclaw (changes requested):
I have no problem with add fall through, as it brings into sharp focus the intent for anyone reviewing the code (eg: fixing a bug)

ibuclaw:
By the way, is the fall through comment consistent with other places it already exists?

src/dmd/backend/cod3.c
wilzbach: Sorry. Good catch.

src/dmd/backend/el.c
wilzbach: I tried this, but `gcc` doesn't seem to like it.

ibuclaw: Is there a break after the `#endif`?

wilzbach:
> By the way, is the fall through comment consistent with other places it already exists?

I think so. I actually first went with `__attribute__ ((fallthrough))` 
as suggested [here](https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7/) until I realized the existing `/* FALL-THROUGH */` pattern.
At `master`:

```d
> grep -C1 -i fall-through
dmd/backend/el.c:2934-                goto Uchar;
dmd/backend/el.c:2935:            /* FALL-THROUGH */
dmd/backend/el.c:2936-        case TYschar:
dmd/backend/cgcs.c:261-        ecom(&e->E2);                   /* eval right first             */
dmd/backend/cgcs.c:262:        /* FALL-THROUGH */
dmd/backend/cgcs.c:263-    case OPucall:
--
dmd/backend/cgcs.c:273-        ecom(&e->E1);
dmd/backend/cgcs.c:274:        /* FALL-THROUGH */
dmd/backend/cgcs.c:275-    case OPasm:
--
dmd/backend/cgcs.c:544-                break;
dmd/backend/cgcs.c:545:            /* FALL-THROUGH */
dmd/backend/cgcs.c:546-        case SCstatic:
--
dmd/backend/cgcs.c:601-                            break;
dmd/backend/cgcs.c:602:                        /* FALL-THROUGH */
dmd/backend/cgcs.c:603-                    case SCstatic:
dmd/backend/debug.c:192-                ferr("#");
dmd/backend/debug.c:193:                /* FALL-THROUGH */
dmd/backend/debug.c:194-            case OPvar:
dmd/backend/cgxmm.c:385-            }
dmd/backend/cgxmm.c:386:            /* FALL-THROUGH */
dmd/backend/cgxmm.c:387-        default:
dmd/backend/cgcv.c:2429-                }
dmd/backend/cgcv.c:2430:                /* FALL-THROUGH */
dmd/backend/cgcv.c:2431-            case SCglobal:
dmd/backend/strtold.c:184-                        sign++;
dmd/backend/strtold.c:185:                        /* FALL-THROUGH */
dmd/backend/strtold.c:186-                case '+':
dmd/backend/cod4.c:967-                        cs.IEV2.Vsize_t = i;
dmd/backend/cod4.c:968:                        // FALL-THROUGH
dmd/backend/cod4.c:969-                    case OPaddass:
--
dmd/backend/cod4.c:2685-                    goto Lload87;
dmd/backend/cod4.c:2686:                /* FALL-THROUGH */
dmd/backend/cod4.c:2687-            case OPs32_d:
--
dmd/backend/cod4.c:2692-                }
dmd/backend/cod4.c:2693:                /* FALL-THROUGH */
dmd/backend/cod4.c:2694-            case OPs16_d:
--
dmd/backend/cod4.c:2725-                    goto Lcnvt87;
dmd/backend/cod4.c:2726:                /* FALL-THROUGH */
dmd/backend/cod4.c:2727-            case OPd_s32:
--
dmd/backend/cod4.c:2732-                }
dmd/backend/cod4.c:2733:                /* FALL-THROUGH */
dmd/backend/cod4.c:2734-            case OPd_s16:
dmd/backend/symbol.c:955-                    vec_free(s->Srange);
dmd/backend/symbol.c:956:                    /* FALL-THROUGH */
dmd/backend/symbol.c:957-#if 0
dmd/backend/out.c:738-                s->Sflags |= SFLlivexit;
dmd/backend/out.c:739:                /* FALL-THROUGH */
dmd/backend/out.c:740-            case SCauto:
dmd/backend/cod2.c:641-                goto L1;                        /* try OP reg,EA        */
dmd/backend/cod2.c:642:        /* FALL-THROUGH */
dmd/backend/cod2.c:643-    default:                                    /* operator node        */
--
dmd/backend/cod2.c:2583-        }
dmd/backend/cod2.c:2584:        // FALL-THROUGH
dmd/backend/cod2.c:2585-    default:
dmd/backend/gloop.c:986-                markinvar(n->E1,rd);
dmd/backend/gloop.c:987:                /* FALL-THROUGH */
dmd/backend/gloop.c:988-        case OPasm:
--
dmd/backend/gloop.c:1972-                    ty = TYllong;
dmd/backend/gloop.c:1973:                /* FALL-THROUGH */
dmd/backend/gloop.c:1974-            case TYint:
dmd/backend/cgelem.c:3353-                    e = el_selecte1(e);
dmd/backend/cgelem.c:3354:                    /* FALL-THROUGH */
dmd/backend/cgelem.c:3355-                default:                /* called by doptelem()         */
--
dmd/backend/cgelem.c:4154-                        break;
dmd/backend/cgelem.c:4155:                    /* FALL-THROUGH */
dmd/backend/cgelem.c:4156-                case OPeqeq:
--
dmd/backend/cgelem.c:4161-                        break;
dmd/backend/cgelem.c:4162:                    /* FALL-THROUGH */
dmd/backend/cgelem.c:4163-                case OPne:
--
dmd/backend/cgelem.c:4511-                    }
dmd/backend/cgelem.c:4512:                    /* FALL-THROUGH */
dmd/backend/cgelem.c:4513-                case OP32_16:
dmd/tk/mem.c:114-#endif
dmd/tk/mem.c:115:                /* FALL-THROUGH */
dmd/tk/mem.c:116-            case MEM_ABORT:
```

src/dmd/backend/el.c
wilzbach: FYI: you can click on "expand" on GitHub diffs, for convenience:

image

jacob-carlborg:
> This shouldn't happen! We need to fix this, s.t. this can't happen in the future. What errors/warnings do you get?

@wilzbach I get this error:

```
dmd/backend/cgreg.c:846:13: error: add explicit braces to avoid dangling else [-Werror,-Wdangling-else]
            else if (s->Sflags & SFLspill)
            ^
```

When I fix the above I get this warning at the end:

```
clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of OS X 10.9 [-Wdeprecated]
```

I'm using Clang 900.0.39.2 on macOS with Xcode 9.2.

WalterBright:
Maybe it's time to stop fiddling about with erratic C++ warnings and just convert the backend to D.

jacob-carlborg:
> Maybe it's time to stop fiddling about with erratic C++ warnings and just convert the backend to D.

I don't mind 😃.

WalterBright:
Looking through the adjustments in this PR it seems clear we've run out of any meat from checking warnings, and are just rearranging the deck chairs. In the past I've done this sort of thing before, and trying to compile with all warnings on across multiple platforms and diverse compilers is a time wasting exercise in futiley chasing rainbows and butterflies. (Hence my resistance to adding warnings to D.)

I suggest we abandon this initiative (Jacob, I still want to to the noreturn thing, that is worthwhile). The dmd C++ code has no future anyway, these changes are irrelevant to its imminent conversion to D.

jacob-carlborg:
> I suggest we abandon this initiative (Jacob, I still want to to the noreturn thing, that is worthwhile)

Yeah, that fixed a bunch of warnings with a very simple and small change. I can compile without any warnings after that tweak.

src/dmd/backend/el.c
ibuclaw: Not on my phone I can't. :-(

ibuclaw:
I'd say that the fall through comments are also useful for clarity sake. Wouldn't it also make conversion to D easier also?

I mean `s#/* FALL-THROUGH */#goto case;#g`

src/dmd/backend/el.c
ZombineDev: At the bottom of every GitHub page there's a "Desktop version" link which you can use to switch to the full desktop version when you're browsing through your phone.

wilzbach:
Closing in favor of #7518. Thanks a lot @ibuclaw.

PR #7476, opened 2017-12-20
Extend the common Makefile between DMD, DRuntime, Phobos etc.

wilzbach:
I'm tired of keeping all the Makefile in sync. The repositories are already interdependent (e.g. we do include `osmodel.mak` and call their respective Makefile targets.

Using a common Makefile has the following advantages:
- automatically keeping everything in sync with _one_ source of truth (don't underestimate this point)
- using a consistent naming scheme (e.g. `DRUNTIME_DIR` vs. `DRUNTIME_PATH`)
- makes the individual Makefiles thinner (DRY!)
- reduces the overhead in maintaining so many Makefiles (e.g. the ddmd -> dmd transition would have been one change in the new `common.mak` file)
- allows to slowly move to target of having _one_ Makefile for Posix & Windows (per repository)

### DIFF
--- a/src/common.mak
+++ b/src/common.mak
@@ -147,7 +147,7 @@ ifeq (,$(findstring win,$(OS)))
 else
 	DOTOBJ:=.obj
 	DOTEXE:=.exe
-	PATHSEP:=$(shell echo "\\")
+	PATHSEP:=$(subst /,/,\)
 endif
 
 ifeq (osx,$(OS))
--- a/src/common.mak
+++ b/src/common.mak
@@ -7,13 +7,13 @@
 # Common directories
 ################################################################################
 
-DMD_DIR = $(D_HOME)/dmd
-DRUNTIME_DIR = $(D_HOME)/druntime
-PHOBOS_DIR = $(D_HOME)/phobos
-DLANG_ORG_DIR = $(D_HOME)/dlang.org
-TOOLS_DIR = $(D_HOME)/tools
-INSTALL_DIR = $(D_HOME)/install
-INSTALLER_DIR = $(D_HOME)/installer
+DMD_DIR=$(D_HOME)/dmd
+DRUNTIME_DIR=$(D_HOME)/druntime
+PHOBOS_DIR=$(D_HOME)/phobos
+DLANG_ORG_DIR=$(D_HOME)/dlang.org
+TOOLS_DIR=$(D_HOME)/tools
+INSTALL_DIR=$(D_HOME)/install
+INSTALLER_DIR=$(D_HOME)/installer
 TMP?=/tmp
 GIT_HOME=https://github.com/dlang
 
@@ -55,7 +55,7 @@ ifneq ($(BUILD),release)
     ifneq ($(BUILD),debug)
         $(error Unrecognized BUILD=$(BUILD), must be 'debug' or 'release')
     endif
-    ENABLE_DEBUG := 1
+    ENABLE_DEBUG:=1
 endif
 
 ################################################################################
@@ -94,16 +94,16 @@ ifneq (,$(DRUNTIME))
 endif
 
 ifeq (,$(findstring win,$(OS)))
-   LIB_DRUNTIME = $(DRUNTIME_BUILD_DIR)/libdruntime.a
-   LIB_DRUNTIMESO = $(basename $(LIB_DRUNTIME)).so.a
+   LIB_DRUNTIME=$(DRUNTIME_BUILD_DIR)/libdruntime.a
+   LIB_DRUNTIMESO=$(basename $(LIB_DRUNTIME)).so.a
 else
-   LIB_DRUNTIME = $(DRUNTIME_PATH)/lib/druntime.lib
+   LIB_DRUNTIME=$(DRUNTIME_PATH)/lib/druntime.lib
 endif
 
 # Set PHOBOS name and full path
 ifeq (,$(findstring win,$(OS)))
-	LIB_PHOBOS = $(PHOBOS_BUILD_DIR)/libphobos2.a
-	LIB_PHOBOSSO = $(PHOBOS_BUILD_DIR)/libphobos2.so
+	LIB_PHOBOS=$(PHOBOS_BUILD_DIR)/libphobos2.a
+	LIB_PHOBOSSO=$(PHOBOS_BUILD_DIR)/libphobos2.so
 endif
 
 # build with shared library support
@@ -116,15 +116,15 @@ LINKDL=$(if $(findstring $(OS),linux),-L-ldl,)
 ################################################################################
 
 ifeq ($(OS),win32wine)
-	CC = wine dmc.exe
-	DMD = wine dmd.exe
-	RUN = wine
+	CC=wine dmc.exe
+	DMD=wine dmd.exe
+	RUN=wine
 else
 	DMD = $(DMD_BUILD_DIR)/dmd
 	ifeq ($(OS),win32)
-		CC = dmc
+		CC=dmc
 	else
-		CC = cc
+		CC=cc
 	endif
 	RUN =
 endif
@@ -171,19 +171,16 @@ endif
 # Default DUB FLAGS
 ################################################################################
 
-DUBFLAGS = --arch=$(subst 32,x86,$(subst 64,x86_64,$(MODEL)))
+DUBFLAGS=--arch=$(subst 32,x86,$(subst 64,x86_64,$(MODEL)))
 
 ################################################################################
 # Automatically create dlang/tools repository if non-existent
 ################################################################################
 
-_all: all
-
 $(TOOLS_DIR):
 	git clone --depth=1 ${GIT_HOME}/$(@F) $@
 
 $(TOOLS_DIR)/checkwhitespace.d: | $(TOOLS_DIR)
-
 
 ################################################################################
 # Common build rules
--- a/src/common.mak
+++ b/src/common.mak
@@ -1,0 +1,204 @@
+# Common Makefile variables (shared between DMD, DRuntime and Phobos)
+# Be careful when updating
+# This file expects the following variables to be defined in the including Makefile:
+# - D_HOME        Root of all D repositories
+
+###############################################################################
+# Common directories
+################################################################################
+
+DMD_DIR = $(D_HOME)/dmd
+DRUNTIME_DIR = $(D_HOME)/druntime
+PHOBOS_DIR = $(D_HOME)/phobos
+DLANG_ORG_DIR = $(D_HOME)/dlang.org
+TOOLS_DIR = $(D_HOME)/tools
+INSTALL_DIR = $(D_HOME)/install
+INSTALLER_DIR = $(D_HOME)/installer
+TMP?=/tmp
+GIT_HOME=https://github.com/dlang
+
+# get OS and MODEL
+include $(D_HOME)/dmd/src/osmodel.mak
+
+###############################################################################
+# Common variables
+################################################################################
+
+# Set VERSION, where the file is that contains the version string
+VERSION=$(DMD_DIR)/VERSION
+MAKEFILE = $(firstword $(MAKEFILE_LIST))
+
+################################################################################
+# Output directories
+################################################################################
+
+DMD_GENERATED=$(DMD_DIR)/generated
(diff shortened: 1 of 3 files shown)


src/common.mak
```diff
@@ -0,0 +1,204 @@
...
+VERSION=$(DMD_DIR)/VERSION
+MAKEFILE = $(firstword $(MAKEFILE_LIST))
+
+################################################################################
+# Output directories
+################################################################################
+
+DMD_GENERATED=$(DMD_DIR)/generated
+DRUNTIME_GENERATED=$(DRUNTIME_DIR)/generated
+PHOBOS_GENERATED=$(PHOBOS_DIR)/generated
```
wilzbach: I wasn't sure whether these should be named `PHOBOS_GENERATED` or `PHOBOS_GENERATED_DIR`.
I think for consistency it would be better to stick the `_DIR` convention. Opinions?

src/common.mak
```diff
@@ -0,0 +1,204 @@
...
+endif
+
+################################################################################
+# Phobos DFLAGS (for linking with the currently built Phobos)
+################################################################################
+
+PHOBOS_DFLAGS=-conf= $(MODEL_FLAG) -I$(DRUNTIME_IMPORT_DIR) -I$(PHOBOS_IMPORT_DIR) -L-L$(PHOBOS_BUILD_DIR) $(PIC)
+ifeq (1,$(SHARED))
+PHOBOS_DFLAGS+=-defaultlib=libphobos2.so -L-rpath=$(PHOBOS_BUILD_DIR)
+endif
```
wilzbach: This is e.g. used by DRuntime, tools and dlang.org

src/common.mak
```diff
@@ -0,0 +1,204 @@
...
+################################################################################
+# Automatically create dlang/tools repository if non-existent
+################################################################################
+
+_all: all
+
+$(TOOLS_DIR):
+	git clone --depth=1 ${GIT_HOME}/$(@F) $@
+
+$(TOOLS_DIR)/checkwhitespace.d: | $(TOOLS_DIR)
```
wilzbach: This section is optional and if people are against it I could remove it, but the fact is that we already have this rule in almost all Makefile, so this stops the redundancy.

src/common.mak
```diff
@@ -0,0 +1,204 @@
...
+	make -C $(DMD_DIR)/src -f posix.mak BUILD=$(BUILD) OS=$(OS) MODEL=$(MODEL)
+	touch $@
+
+$(LIB_DRUNTIME)_build: $(LIB_DRUNTIME)
+	make -C $(DRUNTIME_DIR) -f posix.mak BUILD=$(BUILD) OS=$(OS) MODEL=$(MODEL)
+	touch $@
+
+ifeq (,$(findstring win,$(OS)))
+$(LIB_DRUNTIMESO)_build: $(LIB_DRUNTIME)_build
+endif
```
wilzbach: This is also an optional idea: instead of redefining how DMD and DRuntime should be built over and over again,
the predefined variables with a `_build` suffix could be used.

src/common.mak
```diff
@@ -0,0 +1,204 @@
...
+# Default DUB FLAGS
+################################################################################
+
+DUBFLAGS = --arch=$(subst 32,x86,$(subst 64,x86_64,$(MODEL)))
+
+################################################################################
+# Automatically create dlang/tools repository if non-existent
+################################################################################
+
+_all: all
```
wilzbach: This is necessary to avoid the targets below becoming the default targets.

wilzbach:
As preview for the effect on DRuntime and Phobos, here are the respective PRs:

- https://github.com/dlang/druntime/pull/2004
- https://github.com/dlang/phobos/pull/5949

src/common.mak
wilzbach: There are other target which are common too, e.g. `detab` or `gitzip`, but I wasn't sure whether they are really still in use.

src/common.mak
```diff
@@ -0,0 +1,204 @@
...
+################################################################################
+
+DMD_DIR = $(D_HOME)/dmd
+DRUNTIME_DIR = $(D_HOME)/druntime
+PHOBOS_DIR = $(D_HOME)/phobos
+DLANG_ORG_DIR = $(D_HOME)/dlang.org
+TOOLS_DIR = $(D_HOME)/tools
+INSTALL_DIR = $(D_HOME)/install
+INSTALLER_DIR = $(D_HOME)/installer
+TMP?=/tmp
```
andralex: Here the convention changes from "spaces around =" to "no spaces around ="

src/common.mak
```diff
@@ -0,0 +1,204 @@
...
+GIT_HOME=https://github.com/dlang
+
+# get OS and MODEL
+include $(D_HOME)/dmd/src/osmodel.mak
```
andralex: Prolly we won't need this extra file now? I.e. we can inline its code here.

src/common.mak
andralex: These don't seem common - on the contrary, they're specific to each project. Do we need them here for crosstalk?

src/common.mak
```diff
@@ -0,0 +1,204 @@
...
+
+# Set DOTOBJ and DOTEXE
+ifeq (,$(findstring win,$(OS)))
+	DOTOBJ:=.o
+	DOTEXE:=
+	PATHSEP:=/
+else
+	DOTOBJ:=.obj
+	DOTEXE:=.exe
+	PATHSEP:=$(shell echo "\\")
```
andralex: Isn't there a simple means?

src/common.mak
andralex: Use `.DEFAULT_GOAL`

andralex (changes requested):
There's a bunch of dmd-, druntime-, and phobos-specific stuff in this file so it's more like a union than a common file. I'm not sure the added value is obvious.

src/common.mak
wilzbach: Yes we do. By defining them once here, we don't need to redefine this over and over again whenever we want to access e.g. the generated DMD binary or Phobos library

src/common.mak
wilzbach: > Here the convention changes from "spaces around =" to "no spaces around ="

Ah yes, this was due to the definitions united from multiple Makefiles with no consistent style.

src/common.mak
wilzbach: It's still imported by all repos atm. We can inline it once all repositories are updated and depend on `common.mak`

src/common.mak
wilzbach: This was copied from Phobos: https://github.com/dlang/phobos/blob/master/posix.mak#L160

wilzbach:
> There's a bunch of dmd-, druntime-, and phobos-specific stuff in this file so it's more like a union than a common file. I'm not sure the added value is obvious.

Yes, it's just _one_ file we need to update & maintain instead of seven.
Including dmd/druntime/phobos-specific bits are needed as the other repositories already do this and redefine the specific bits all over again, which is _very annoying_ when you want to change them and now suddenly have to update five repos (and these in a defined order because of dependencies)
I think the added values is obvious:

1) When you look at how much stuff we will be able to remove from Druntime and Phobos without much work:

- https://github.com/dlang/druntime/pull/2004
- https://github.com/dlang/phobos/pull/5949

2) When you remember how much pain it was to do the ddmd -> dmd transition? (I don't want to invest that much work in every simple Makefile change)

3) When we think about moving forward and dropping DigitalMars make for Windows. This common.mak file can include common OS-specific variables like `PATHSEP`.

src/common.mak
rainers: This works without invoking a subprocess:
```
PATHSEP := $(subst /,/,\)
```

MartinNowak:
> I'm tired of keeping all the Makefile in sync. The repositories are already interdependent (e.g. we do include osmodel.mak and call their respective Makefile targets.

In hindsight separate repos weren't such a good idea, and they still constantly cause a ~1.5x overhead in GH PRs, require plenty of cross-links and extra hops by reviewers.
Maybe we should jump through the hoop and (re-)discuss a monorepo for all of dlang?

For now unifying makefiles is a good step.

DmitryOlshansky:
> For now unifying makefiles is a good step.

Long ago I thought having independent repos will cut on untold dependencies between compiler, run-time and phobos. Today I'm not so sure, they are here no matter how hard we wanted to ignore them.

Indeed at least dmd and druntime are super often developed in the lockstep.
Phobos still needs to coordinate on efforts such as -dip1000 and so on and so forth.

andralex:
I continue to be skeptical about this direction. I can see very well how 6 months from now a contrary PR could be made: "This PR modularizes the mishmash grab-bag of variables defined in the common makefile and used only in one project or not at all. 

As a random example, why does anyone but the installer need `INSTALL_DIR`?

There will be no pressure to reduce the size of the common makefile; it will only grow. Anyone's shoe is too tight today? Add a variable to it. I don't think we want to go this way.

wilzbach:
> As a random example, why does anyone but the installer need INSTALL_DIR?

- dlang.org uses it for parsing the changelog entries
- the tools repo uses the installer for parsing the git log for the contributor stats which is then used by dlang.org again

(there's probably more and that's a rare exception of infrequent use)

> There will be no pressure to reduce the size of the common makefile; it will only grow. Anyone's shoe is too tight today? Add a variable to it. I don't think we want to go this way.

Well, one point is to avoid having exactly this happening in each repository.
Or rather cleaning it up as it already has happened, e.g. there's `DRUNTIME_DIR` in dlang.org and `DRUNTIME_PATH` in phobos or `DOC_OUTPUT_DIR` in phobos and `DOCSRC` in druntime (those are actual examples and they aren't the only ones). This requires a lot of special casing (an example: https://github.com/dlang/dlang.org/blob/c8c9b7345c82648f77168dc22346c3d8ba279c8e/posix.mak#L238)

> I continue to be skeptical about this direction. I can see very well how 6 months from now a contrary PR could be made: "This PR modularizes the mishmash grab-bag of variables defined in the common makefile and used only in one project or not at all.

I highly doubt that we would end up with `DRUNTIME_DIR` and `DRUNTIME_PATH` again.
Also note that this PR is not only about having a standard definition for the Makefile variables, but it also adds a few common targets, s.t. all repos can depend on the same target and don't need to redefine it again.

There's also the argument that by having one file (aka standard) file in which all major variables are defined we enforce a consistent Makefile style over all repos, make it easier for people to learn our Makefile style (as there then actually would be one) and avoids that each repository has its custom naming scheme for variables.

Also note that we already have had a common & shared Makefile "library" - `osmodel.mak` is already included in each Makefile and it hasn't grown in a mishmash grab-bag, has it? On the contrary, ~ one year ago I removed all duplicated `osmodel.mak` files because we ran into many problems of these files being inconsistent (even though they are really tiny).

> I can see very well how 6 months from now 

BTW if we ever tackle the annoyance with the win32/win64.mak files this unification will also save a lot of redundant definitions (in fact it's one of the main reasons for this PR)

@CyberShadow @marler8997 @JinShil @MartinNowak @timotheecour you all voted for this. So maybe you have a more convincing argumentation than I do?

marler8997:
Just brainstorming.  Would it work well to keep the compiler/druntime/phobos repos separate, and put all the common stuff in a new repo that they all depend on?  Common code/makefile definitions/etc?  I think the hope would be that you could download just one repo and build/test it without the others.

marler8997:
> I continue to be skeptical about this direction. I can see very well how 6 months from now a contrary PR could be made: "This PR modularizes the mishmash grab-bag of variables defined in the common makefile and used only in one project or not at all.

As far as I understand, this common makefile is meant for variables that are common to one or more repositories.  I can't see someone removing a variable from this file and making multiples copies of it in multiple different repositories.  That would be like someone taking a function out of a common library, and making two identical copies of it...who would do that!  In the other case that something changed to make a variable specific to only 1 repository...then removing it from this makefile sounds fine.  That would only happen to accommodate other changes, not a result of a change to this design.

I do have to point out that this example sounds a bit funny when paired with your second statement about there being "no pressure to reduce the size of the common makefile; it will only grow".  This being one example where the common makefile would be "reduced in size" :)

wilzbach:
It's a shame, but it looks like we can't convince Andrei for now.
Maybe the best solution forward is to convert all build scripts to D (though this doesn't really solve the problem of having one shared file for all common variables).

PR #7482, opened 2017-12-21
Remove a couple temp files created by test suite

JinShil:
Alternative to  #7462

cc @wilzbach

PR #7483, opened 2017-12-21
Ensure that the testsuite doesn't generate unchecked temporary files

wilzbach:
So that #7482 isn't needed anymore in the future.


posix.mak
```diff
@@ -39,6 +39,10 @@ install: all
 	mkdir -p $(INSTALL_DIR)/man
 	cp -r docs/man/* $(INSTALL_DIR)/man/
 
+check-clean-git:
+	@git diff-index --quiet HEAD -- || \
+		(echo "ERROR: You created temporary files. Store them in tests_results or remove them."; git status; exit 1)
```
JinShil: The error message sounds a bit too much like it is accusing the current user, which may not be at fault for the temporary files.  How about "ERROR:  Found residual temporary files.  Temporary files should be stored in `test_results` or explicitly removed."  

Also, the `git status` tends to overwhelm the error message, making it less obvious.  Perhaps use `git status -s` which displays a less verbose output.  In that case you may want to change the error message to "ERROR:  Found the following residual temporary files.  Temporary files should be stored in `test_results` or explicitly removed."

Example output:
```
ERROR:  Found the following residual temporary files.  Temporary files should be stored in `test_results` or explicitly removed.
?? dub.selections.json
```

Sorry for being so picky about such things.  I can't help it.

JinShil (approved):

posix.mak
JinShil: Also, it might be helpful to add a comment with a link to this PR.  It helps when maintaining these makefiles to understand why certain things exist.

JinShil: Also, it might be helpful to add a comment with a link to this PR. It helps when maintaining these makefiles to understand why certain things exist.

posix.mak
```diff
@@ -39,6 +39,10 @@ install: all
 	mkdir -p $(INSTALL_DIR)/man
 	cp -r docs/man/* $(INSTALL_DIR)/man/
 
+check-clean-git:
```
WalterBright: Major targets should have documentation explaining what they do. Otherwise, the user is left to puzzle out the commands executed, which (like all code) isn't always obvious.

WalterBright (dismissed):
document check-clean-git

JinShil (approved):

PR #7484, opened 2017-12-21
Implement initial version of lambda comparison

RazvanN7:
This PR implements the first draft of the lambda comparison project. For the moment, it is possible to compare lambda functions :

-> which use in their body only parameters, IntegerExps and enums (no runtime variables)
-> that have user defined types as parameter types

A lambda function which has runtime variables or function calls in its body will be considered uncomparable for the moment. I plan on working on that case.

The current PR is just a proof of concept and I am sure that there are a lot of corner cases that I've missed and if you identify such cases please report them. I plan an adding some tests in the next commit and also try to see if function calls can be supported by putting the mangled name of the function in the lambda serialization.

### DIFF
--- a/src/dmd/lambdacomp.d
+++ b/src/dmd/lambdacomp.d
@@ -1,0 +1,227 @@
+module dmd.lambdacomp;
+
+import core.stdc.stdio;
+import core.stdc.string;
+
+import dmd.declaration;
+import dmd.denum;
+import dmd.dsymbol;
+import dmd.expression;
+import dmd.func;
+import dmd.dmangle;
+import dmd.mtype;
+import dmd.root.outbuffer;
+import dmd.root.stringtable;
+import dmd.dscope;
+import dmd.statement;
+import dmd.tokens;
+import dmd.visitor;
+
+enum ExpType
+{
+    None,
+    EnumDecl,
+    Arg
+}
+
+extern (C++) class SerializeVisitor : SemanticTimeTransitiveVisitor
+{
+    alias visit = SemanticTimeTransitiveVisitor.visit;
+    OutBuffer buf;
+    StringTable arg_hash;
+    Scope* sc;
+    ExpType et;
+    Dsymbol d;
+
+    this(Scope* sc)
+    {
+        this.sc = sc;
+    }
+
+    override void visit(FuncLiteralDeclaration fld)
+    {
+        if (fld.type.ty == Terror)
+            return;
+
+        TypeFunction tf = cast(TypeFunction)fld.type;
+        uint dim = cast(uint)Parameter.dim(tf.parameters);
+        buf.printf("%d:", dim);
+
+        arg_hash._init(dim + 1);
+        foreach (i; 0 .. dim)
+        {
+            auto fparam = Parameter.getNth(tf.parameters, i);
+            if (fparam.ident !is null)
+            {
+                auto key = fparam.ident.toString().ptr;
+                OutBuffer value;
+                value.writestring("arg");
+                value.print(i);
+                arg_hash.insert(key, strlen(key), value.extractString);
+                fparam.accept(this);
+            }
+        }
+
+        CompoundStatement cs = fld.fbody.isCompoundStatement();
+        Statement s = !cs ? fld.fbody : null;
+        ReturnStatement rs = s ? s.isReturnStatement() : null;
+        if (rs && rs.exp)
+        {
+            rs.exp.accept(this);
+        }
+    }
+
+    override void visit(DotIdExp exp)
+    {
+        if (buf.offset == 0)
+            return;
+
+        exp.e1.accept(this);
+        if (buf.offset == 0)
+            return;
+
+        if (et == ExpType.EnumDecl)
+        {
+            Dsymbol s = d.search(exp.loc, exp.ident);
+            if (s)
+            {
+                if (auto em = s.isEnumMember())
+                {
+                    em.value.accept(this);
+                }
+                et = ExpType.None;
+                d = null;
+            }
+        }
+
+        else if (et == ExpType.Arg)
+        {
+            buf.setsize(buf.offset -1);
+            buf.writeByte('.');
+            buf.writestring(exp.ident.toString());
+            buf.writeByte('_');
+        }
+    }
+
+    override void visit(IdentifierExp exp)
+    {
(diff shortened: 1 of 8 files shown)


JinShil:
What's the use case motivating this?

RazvanN7:
@JinShil In phobos there are a lot of algorithms which may apply some optimizations if the given range is sorted. In order to check whether the range is sorted with a certain predicate a lambda comparison operation is required which is currently achieved by comparing string lambda's; this technique is buggy since it only does string comparison. By adding support in the compiler for lambda comparison this will enable safe comparison of predicates

wilzbach:
> What's the use case motivating this?

A good example of my head is this recent PR which makes {min,max}Element faster https://github.com/dlang/phobos/pull/4265
For an identity map function, the algorithm is 6x faster with LDC.
So how do we check for identity?
Currently the only ways are (a) string lambdas (`enum isIdentity = map == "a";`) or (b) using many overloads (that's what we did with {min,max}Element eventually).
I really look forward to use the new trait for such optimizations :)

RazvanN7:
Does anyone have any idea why this particular line [1] make compiling druntime impossible? 

[1] https://github.com/dlang/dmd/pull/7484/files#diff-b601037010863a0a6d136649d1e4a5d6R195

wilzbach:
> What's the use case motivating this?

Oh there's also the movement to phase out string lambdas, see e.g. [this old NG post](http://forum.dlang.org/post/twwhyhxyfcvojitsbvmy@forum.dlang.org) and IIRC comparability was one of the last blocking points.

jacob-carlborg:
> I figured it out: it seems that in order to do the mangling, semantic3 must have been performed and for some situations that is not the case.

The semantic analysis of the `.mangleof` property is done in the first semantic phase, as far as I understand.

RazvanN7:
@jacob-carlborg The `mangleof` property is present only for types and expressions. For functions there is `mangleString`, but for Dsymbols (or AggregateDeclarations to be more specific) I haven't anything but the mangleToBuffer function.

stefan-koch-sociomantic:
This is not sufficiently tested.

RazvanN7:
It seems like the 32 bit machines compile the test with a version of the compiler which does not include this patch.

wilzbach:
Nope. All CIs compile DMD from source before running the testsuite. So the error is on your side.

Closing and opening for SemaphoreCI to appear.

andralex:
cc @WalterBright

src/dmd/traits.d
```diff
@@ -1419,8 +1419,50 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
 
+
         auto o1 = (*e.args)[0];
         auto o2 = (*e.args)[1];
+
+        FuncLiteralDeclaration isLambda(RootObject oarg)
+        {
+            if (auto t = isDsymbol(oarg))
+            {
+                if(auto td = t.isTemplateDeclaration())
```
andralex: space after if

andralex (approved):
nice work please fix style

andralex:
Will pull this, please add to the spec appropriately.

UplinkCoder:
There are no comments!

wilzbach:
And no changelog entry / spec PR :(

src/dmd/lambdacomp.d
```diff
@@ -0,0 +1,227 @@
+module dmd.lambdacomp;
```
wilzbach: No ddoc header - please have a look at the header files.

src/dmd/lambdacomp.d
```diff
@@ -0,0 +1,227 @@
...
+}
+
+extern (C++) class SerializeVisitor : SemanticTimeTransitiveVisitor
+{
+    alias visit = SemanticTimeTransitiveVisitor.visit;
+    OutBuffer buf;
+    StringTable arg_hash;
+    Scope* sc;
+    ExpType et;
+    Dsymbol d;
```
wilzbach: `private`?

src/dmd/lambdacomp.d
```diff
@@ -0,0 +1,227 @@
...
+        uint dim = cast(uint)Parameter.dim(tf.parameters);
+        buf.printf("%d:", dim);
+
+        arg_hash._init(dim + 1);
+        foreach (i; 0 .. dim)
+        {
+            auto fparam = Parameter.getNth(tf.parameters, i);
+            if (fparam.ident !is null)
+            {
+                auto key = fparam.ident.toString().ptr;
```
wilzbach: FWIW `.ptr` is deprecated in `@safe` code: https://dlang.org/changelog/pending.html#ptr-safe-end-of-deprecation

src/dmd/lambdacomp.d
```diff
@@ -0,0 +1,227 @@
...
+                OutBuffer value;
+                value.writestring("arg");
+                value.print(i);
+                arg_hash.insert(key, strlen(key), value.extractString);
```
wilzbach: You already call `toString` above and know the length of `key`

src/dmd/traits.d
```diff
@@ -1419,8 +1419,50 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         if (!TemplateInstance.semanticTiargs(e.loc, sc, e.args, 0))
             return new ErrorExp();
 
+
         auto o1 = (*e.args)[0];
         auto o2 = (*e.args)[1];
+
+        FuncLiteralDeclaration isLambda(RootObject oarg)
```
wilzbach: `static`

ibuclaw:
And no C++ header updates.

andralex:
@RazvanN7 and I discussed this during our regular meeting. @RazvanN7 in addition to the spec update, please also add a changelog entry. @ibuclaw what C++ header updates are needed?

ibuclaw:
A new field was added to FuncLiteralDeclaration.

https://github.com/dlang/dmd/pull/7484/files#diff-ca782fb8b1ce12422a29fed42ffe0cadR2921

src/dmd/lambdacomp.d
RazvanN7: toString returns a const(char)*

RazvanN7:
Doc update : https://github.com/dlang/dlang.org/pull/2146

PR #7485, opened 2017-12-21
Move frontend utility functions from the DUB package to dmd.frontend

wilzbach:
Following to https://github.com/dlang/dmd/pull/7425, this just move these bits to `dmd.frontend`. Of course, further work is needed to define the API, but I thought having an open PR is the best place to facilitate this discussion.

This is far from perfect, but I won't have time to work much on it until I chat with @RazvanN7.

Anyhow if you have already have an API idea/requirement and want to drop it or just want to subscribe to the discussion, this is the place to be ;-)

### DIFF
--- a/src/dmd/frontend.d
+++ b/src/dmd/frontend.d
@@ -14,36 +14,31 @@
  */
 module dmd.frontend;
 
-/**
-Initializes the DMD compiler
+import dmd.dmodule : Module;
+import std.range.primitives : isInputRange, ElementType;
+import std.traits : isNarrowString;
+
+version (Windows) private enum sep = ";", exe = ".exe";
+version (Posix) private enum sep = ":", exe = "";
+
+/*
+Initializes the global variables of the DMD compiler.
+This needs to be done $(I before) calling any function.
 */
 void initDMD()
 {
+    import dmd.builtin : builtin_init;
     import dmd.dmodule : Module;
+    import dmd.expression : Expression;
     import dmd.globals : global;
     import dmd.id : Id;
+    import dmd.mars : addDefaultVersionIdentifiers;
     import dmd.mtype : Type;
-    import dmd.target : Target;
-    import dmd.expression : Expression;
     import dmd.objc : Objc;
-    import dmd.builtin : builtin_init;
+    import dmd.target : Target;
 
-    global._init;
-
-    version(linux)
-        global.params.isLinux = 1;
-    else version(OSX)
-        global.params.isOSX = 1;
-    else version(FreeBSD)
-        global.params.isFreeBSD = 1;
-    else version(Windows)
-        global.params.isWindows = 1;
-    else version(Solaris)
-        global.params.isSolaris = 1;
-    else version(OpenBSD)
-        global.params.isOpenBSD = 1;
-    else
-        static assert(0, "OS not supported yet.");
+    global._init();
+    addDefaultVersionIdentifiers();
 
     Type._init();
     Id.initialize();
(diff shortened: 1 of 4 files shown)


jacob-carlborg:
What's the idea here, just move common functionality from the test package or to come up new a proper API?

src/dmd/frontend.d
```diff
@@ -54,3 +62,161 @@ void initDMD()
...
+void addImports(T)(T path)
+{
+    import dmd.globals : global;
+    import dmd.arraytypes : Strings;
+
+    Strings* res = new Strings();
+    foreach (p; path)
+    {
+        import std.string : toStringz;
+        Strings* a = new Strings();
```
jacob-carlborg: Why is an array created for each path, why not added the path directly to `res`?

src/dmd/frontend.d
```diff
@@ -54,3 +62,161 @@ void initDMD()
...
+        .map!(l => l.matchAll(`-I[^ "]+`.regex)
+                    .joiner
+                    .map!(a => a.drop(2)
+                                .replace("%@P%", dmdFilePath.dirName)
+                                .replace("%%ldcbinarypath%%", dmdFilePath.dirName)))
+        .joiner
+        .array
+        .sort
+        .uniq
+        .map!buildNormalizedPath;
```
jacob-carlborg: This whole expression doesn't look very nice. Perhaps split it up with some helper functions/aliases. It's not very clear of what the code is actually doing.

src/dmd/frontend.d
```diff
@@ -14,10 +14,18 @@ module dmd.frontend;
...
 // Online documentation: https://dlang.org/phobos/dmd_frontend.html
 
-/**
-Initializes the DMD compiler
+import dmd.dmodule : Module;
+
+shared static this()
+{
+    initDMD();
+}
```
jacob-carlborg: I'm not sure if this is the best idea if we want to allow multiple instances of the compiler in the future.

src/dmd/frontend.d
```diff
@@ -54,3 +62,161 @@ void initDMD()
...
+auto findImportPaths()
+{
+    import std.file : exists, getcwd;
+    import std.string : fromStringz, toStringz;
+    import std.path : buildPath, buildNormalizedPath, dirName;
+    import std.process : env = environment, execute;
+    import dmd.dinifile : findConfFile;
+    import dmd.errors : fatal;
+    import std.algorithm, std.range, std.regex;
+    import std.stdio : File,  stderr;
```
jacob-carlborg: There's an extra space after the comma.

wilzbach:
> What's the idea here, just move common functionality from the test package or to come up new a proper API?

Both, but (B) is definitely the goal. I just moved the bits from the test package over to bootstrap a discussion and avoid this topic sinking in the abyss.
Anyhow I assume that coming up with a proper API from scratch will be tricky and require a couple of iterations (and use cases). So in any case I will put a red reminder in the module header that warns the user about the API's instability. Maybe `dmd.experimental.frontend` would have been a better place for this module?

src/dmd/frontend.d
```diff
@@ -54,3 +62,161 @@ void initDMD()
...
+
+    auto dmdEnv = env.get("DMD", "dmd");
```
jacob-carlborg: I think this function needs to be split into several layers building on top of each other. One where it's possible to give a direct path to the config and another which tries to infer it.

src/dmd/frontend.d
```diff
@@ -54,3 +62,161 @@ void initDMD()
     builtin_init();
 }
 
+/**
+Add import paths to the `global.path`.
+Params:
+    path = range of imports to add
+*/
+void addImports(T)(T path)
```
jacob-carlborg: This should have constraints.

src/dmd/frontend.d
```diff
@@ -54,3 +62,161 @@ void initDMD()
...
+        .map!buildNormalizedPath;
+}
+
+/**
+Parse a module from a string.
+
+Returns: the parsed module object
+*/
+// TODO: allow files?
+auto parseModule(string fileName, string code)
```
jacob-carlborg: I don't think `auto` should be used here. It returns a simple class and not a template, so I don't think it's necessary.

src/dmd/frontend.d
```diff
@@ -54,3 +62,161 @@ void initDMD()
...
+    m.semantic3(null);
+}
+
+/**
+Pretty print a module.
+
+Returns:
+    Pretty printed module as string.
+*/
+auto prettyPrint(Module m)
```
jacob-carlborg: Same here, write out the actual return type instead of `auto`.

jacob-carlborg (changes requested):

jacob-carlborg:
> So in any case I will put a red reminder in the module header that warns the user about the API's instability. Maybe dmd.experimental.frontend would have been a better place for this module?

I'm don't think that's necessary. I do think the version of the first releases should be `0.x.y`, that allows breaking changes at any point according to semantic versioning.

src/dmd/frontend.d
```diff
@@ -54,3 +59,221 @@ void initDMD()
...
+*/
+auto parseImportPathsFromConfig(string iniFile, string execDir)
+{
+    import std.algorithm, std.range, std.regex;
+    import std.stdio : File;
+    import std.path : buildNormalizedPath;
+
+    return File(iniFile, "r")
+        .byLineCopy
+        // search for all -I imports in this file
```
jacob-carlborg: I still prefer separate functions/lambdas instead of comments.

jacob-carlborg (changes requested):

ibuclaw:
Shouldn't this module be named something like `dmd.driver`?  It's pretty useless for other backends.

wilzbach:
> Shouldn't this module be named something like dmd.driver? It's pretty useless for other backends.

It's not intended to be used by _any_ library. It's only intended to be used for the DMD as a library builds - most likely via DUB.

Regarding the name, see: https://github.com/dlang/dmd/pull/7425#discussion_r156359727
I also don't think the name is perfect, so happy to rename.

src/dmd/mars.d
```diff
@@ -1311,7 +1311,7 @@ private void setDefaultLibrary()
  * variable and config file) in order to add final flags (such as `X86_64` or
  * the `CRuntime` used).
  */
-private void addDefaultVersionIdentifiers()
+void addDefaultVersionIdentifiers()
```
wilzbach: See https://github.com/dlang/dmd/pull/7524 for a better approach

src/dmd/frontend.d
```diff
@@ -14,36 +14,31 @@
  */
 module dmd.frontend;
 
-/**
-Initializes the DMD compiler
+import dmd.dmodule : Module;
+import std.range.primitives : isInputRange, ElementType;
+import std.traits : isNarrowString;
```
mathias-lang-sociomantic: Thought Phobos is DMD was a no-no ?

wilzbach: This file isn't used by dmd at all - only by the _DMD as a library_ project.
Should we rename the file or move it to somwhere else, s.t. this gets clearer?

I don't see any reason not to use Phobos here, it allows to write the high-level wrappers and stops us from re-inventing the wheel and blowing up the code with inferior, probably buggy implementations of Phobos.

wilzbach:
BTW anything else blocking this PR except for the name of this module?
(it's already there and we can always rename if after this PR though)

Everything is still marked as `unstable`, so we can always improve/change this, but it should give us sth. to get started.

src/dmd/frontend.d
```diff
@@ -54,3 +49,242 @@ void initDMD()
...
+	import std.range : front, empty, transposed;
+	// adapted from DUB: https://github.com/dlang/dub/blob/350a0315c38fab9d3d0c4c9d30ff6bb90efb54d6/source/dub/dub.d#L1183
+
+	auto compilers = ["dmd", "gdc", "gdmd", "ldc2", "ldmd2"];
+
+	// Search the user's PATH for the compiler binary
+	if ("DMD" in environment)
+		compilers = environment.get("DMD") ~ compilers;
+	auto paths = environment.get("PATH", "").splitter(sep);
+    auto res = compilers.map!(c => paths.map!(p => p.buildPath(c~exe))).joiner.filter!exists;
```
jacob-carlborg: The indentation is wrong.

jacob-carlborg (changes requested):

src/dmd/frontend.d
wilzbach: Fixed.

wilzbach:
Closing & reopening for Semaphore to learn about this PR.

wilzbach:
Anything else blocking this?

jacob-carlborg (approved):

wilzbach:
CC @RazvanN7 are you okay with this too?

RazvanN7:
LGTM

PR #7487, opened 2017-12-21
[WIP] Provide a mechanism to add extra library search paths for OPTLINK

marler8997:
Currently there's no mechanism to add library search paths when using dmd with optlink on windows.

This change allows you to use the following command line to prepend/append library search paths for optlink.
```
dmd -L:libMyHighPriorityPath\ -L:-libMyLowPriorityPath\
```

To explain a bit more.  The problem is that you cannot add library search paths to optlink via command line switches.  The only way to add search paths is to modify the config file, or add them to the `lib` location on the command line:
```
LINK obj[,out[,map[,lib[,def[,res]]]]]
                    ^
     must be added here, and paths must end with a trailing backslash \
```

The problem is that all the `-L<linker-option>` options are added to the end of the command, which does not allow you to add library search paths.  This change treats optlink options prefixed with `:lib` and `:-lib` special by inserting them into the front/back of the `lib` command line location.

### DIFF
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -68,33 +68,48 @@ version (CRuntime_Microsoft)
   }
 }
 
-/****************************************
- * Write filename to cmdbuf, quoting if necessary.
- */
+/**
+Returns: true if the given filename needs a quote
+*/
+private bool optlinkNeedsQuote(const(char)[] filename)
+{
+    foreach (c; filename)
+    {
+        if (!isalnum(c) && c != '_' && c != '.')
+            return true; // needs quote
+    }
+    return false; // does not need quote
+}
+
+/**
+Write filename to cmdbuf, quoting if necessary.
+*/
 private void writeFilename(OutBuffer* buf, const(char)* filename, size_t len)
 {
-    /* Loop and see if we need to quote
-     */
-    for (size_t i = 0; i < len; i++)
+    if (optlinkNeedsQuote(filename[0 .. len]))
     {
-        char c = filename[i];
-        if (isalnum(c) || c == '_')
-            continue;
-        /* Need to quote
-         */
+        // TODO: add some tests in this case to make sure
+        //       that backslashes in the name work (don't need to be escaped)
+        //       especially if the backslash is at the end of the filename because
+        //       that could be interpreted as an escaped quote.
         buf.writeByte('"');
         buf.write(filename, len);
         buf.writeByte('"');
-        return;
     }
-    /* No quoting necessary
-     */
-    buf.write(filename, len);
+    else
+    {
+        buf.write(filename, len);
+    }
 }
 
 private void writeFilename(OutBuffer* buf, const(char)* filename)
 {
     writeFilename(buf, filename, strlen(filename));
+}
+
+private void writeDelimiter(OutBuffer* buf, bool* first)
+{
+    if (*first) *first = false; else buf.writeByte('+');
 }
 
 version (Posix)
@@ -337,11 +352,38 @@ public int runLINK()
             else
                 cmdbuf.writestring("nul");
             cmdbuf.writeByte(',');
-            for (size_t i = 0; i < global.params.libfiles.dim; i++)
+
+            static bool isLibrarySearchPath(const(char)* option)
             {
-                if (i)
-                    cmdbuf.writeByte('+');
-                writeFilename(&cmdbuf, global.params.libfiles[i]);
+                return option[0] == '+' && option[strlen(option) - 1] == '\\';
+            }
+            {
+                bool first = true;
+                // Add library search paths specified by `-L+<path>\`
+                for (size_t i = 0; i < global.params.linkswitches.dim; i++)
+                {
+                    auto option = global.params.linkswitches[i];
+                    if (isLibrarySearchPath(option))
+                    {
+                        writeDelimiter(&cmdbuf, &first);
+                        // Need to ad an EXTRA trailing slash after the quotes if the path uses quotes
+                        if (optlinkNeedsQuote(option[0 .. strlen(option)]))
+                        {
+                            cmdbuf.writeByte('"');
+                            cmdbuf.writestring(option + 1);
+                            cmdbuf.writestring(`"\`);
+                        }
+                        else
+                        {
+                            cmdbuf.writestring(option + 1);
+                        }
+                    }
+                }
+                for (size_t i = 0; i < global.params.libfiles.dim; i++)
+                {
+                    writeDelimiter(&cmdbuf, &first);
+                    writeFilename(&cmdbuf, global.params.libfiles[i]);
+                }
             }
             if (global.params.deffile)
             {
@@ -382,7 +424,8 @@ public int runLINK()
             cmdbuf.writestring("/noi");
             for (size_t i = 0; i < global.params.linkswitches.dim; i++)
             {
-                cmdbuf.writestring(global.params.linkswitches[i]);
+                if (!isLibrarySearchPath(global.params.linkswitches[i]))
+                    cmdbuf.writestring(global.params.linkswitches[i]);
             }
             cmdbuf.writeByte(';');
             char* p = cmdbuf.peekString();
--- a/test/compilable/libpath.sh
+++ b/test/compilable/libpath.sh
@@ -1,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -ueo pipefail
+
+TEMP_LIB_PATH=${RESULTS_DIR}${SEP}compilable${SEP}templibpath
+
+function echorun() {
+    echo $@
+    $@
+}
+
+if [ $OS == "win32" -o  $OS == "win64" ]; then
+    LIB_ARGS=
+    FOOLIB_OUT=foolib.lib
+    FOOLIB_ARG=foolib.lib
+    # TODO: if MSVC; then
+    #    LINKER_FLAG=/LIBPATH:${TEMP_LIB_PATH}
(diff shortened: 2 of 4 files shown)


marler8997:
P.S.  This change was made to enable running the dmd test suite on windows.  By being able to add more library search paths to optlink, you can add the path to the newly built photos library.  This means you don't have to copy `dmd.exe` to the `windows\bin` path along with creating a `sci.ini` file and copying all your libraries along with `phobos.lib` to `windows\lib`.

rainers:
Do we need an extra syntax for this? Maybe it's good enough to move any optlink option that is not an option itself and ends with '\' to the correct location for library paths? Can it be anything else?

rainers:
BTW: I'm passing LIB="path to phobos" as an argument to gnu-make when running the tests. This puts it into the environment when dmd and optlink are invoked.

jacob-carlborg:
I thought library paths were added using `-L+<path>\`.

marler8997:
@jacob-carlborg 

Wow that actually works.  I'm guessing this might have been an undocumented extension added to optlink because someone realized you couldn't add search paths with "unordered command line switches".  Anway, that makes this PR obselete, thanks for bringing this to my attention :)

jacob-carlborg:
It's been used in DWT for ages.

marler8997:
@rainers 

I tried as hard as I could to get the `LIB` environment variable to affect optlink in any way.  I messed with the config file, with the /SCANLIB option and no matter what I did I couldn't get it to work.  Are you sure that you are able to get it to work?  If you set it to some garbage value, does it break optlink?

wilzbach:
I don't use Windows, but it seems like the build guide (https://wiki.dlang.org/Building_under_Windows) needs a few updates?

rainers:
> I thought library paths were added using -L+<path>\.

That doesn't work in combination with a def file, because it adds a comma that moves the library path into the wrong command line section:
```
>dmd -conf= test.d -Ic:\s\d\rainers\druntime\import -L+c:\s\d\rainers\lib\ x.def -v
...
link test,,nul,user32+kernel32,"x.def"/noi+c:\s\d\rainers\lib\;
user32.lib
 Warning 2: File Not Found user32.lib
...
```

rainers:
> I tried as hard as I could to get the LIB environment variable to affect optlink in any way.

This works for me (with dmd and link copied into an empty directory):
```
>set LIB=c:\s\d\rainers\lib
>dmd -conf= test.d -Ic:\s\d\rainers\druntime\import
```
but fails if LIB is not set. Please note that optlink also reads LIB from sc.ini in its directory, section [Environment] which might override your environment.

marler8997:
Ah ok I was seeing the same thing.  The semantics I'm looking for is to ADD more library paths in addition to what is already in the "sc.ini" file.  It looks like in that case the `LIB` environment variable doesn't work.

Also I didn't realize that if you added def files it would break the library path solution.  Reopening this PR.

marler8997:
The problem with only using the trailing backslash is you can't indicate whether you want the path to come before or after the rest of the arguments in the `lib` section.
```
dmd -Lmypath\

link ,,,mypath\+user32+kernel32
OR
link ,,,user32+kernel32+mypath\
```

You could support some character to distinguish this (i.e. `-L+mypath\` or `-L-mypath\`) but then we are back to this solution of supporting a custom syntax.

Also, there's are advantages to supporting a syntax/interface that allows the user to explicitly declare what they want instead of trying to infer it.  If we want to allow users to put arguments in the `lib` section, the option `-L:libfoo` is more intuitive than `-Lfoo\`. 

Note that this interface is also extensible, if we want to support more special options we can do so with `-L:myspecialoption`.  We may never need it, but it does show how this special syntax has its advantages.

rainers:
As there are no other library search paths on the linker command line generated by dmd, the user has full control over the order already. I still think we should not add new syntax, but just ensure placing any linker option starting with `+` into the linker section of the command line (not using the trailing `\` I proposed before). That helps the .def file issue.

BTW: would it be possible to add a test?

marler8997:
Yes I could easily add a test.

I don't quite understand how you've come to your conclusion though.  You're saying you don't want to add a new syntax, but using the `+` sign to indicate that you want to put the option in the lib section is a new syntax.
```
-Lfoo    # Regular linker switch
-L+foo   # Not a regular linker switch, put this in the lib section
```

If we want to use this technique, which I think we should, why not use a clear syntax that is both intuitive and extensible?
```
-L:libfoo   # Put this argument in the 'lib' section of the linker command
```

If you had a developer who is familiar with optlink and they saw `-L+foo`, how would they know that it's not like all the other optlink switches which are just appended to the end of the command line?  If they saw `-L:libfoo`, they would know that optlink doesn't use the colon `:` character for anything and intuitively realize it must have some special meaning.  Hmm, now that I look at the option, maybe it would better to use `-L:lib:foo`?  I imagine a developer who sees that could easily guess exactly what it means.

Also, if we change what the semantics of `+` we could be breaking existing code that depends on it.  Maybe someone is using that syntax to add a def file.  If we went with your suggestion we might want to use both the `+` and the trailing slash.

jacob-carlborg:
My current interpretation, which might be wrong, of the `-L` flag is that everything after it is passed to the linker as is. Perhaps we need a different flag?

rainers:
AFAICT there is no option that start with `+`. The only use case for it is to add files or paths to the list of libraries. So nothing needs to change for anyone using this "trick". In combination with a def file (it's not possible to pass multiple def files) it produces an error, so changing the position of that "option" will only remove that error.

marler8997:
@jacob-carlborg 

Yes that's what this PR is for.  It allows you to prefix your flag with `:lib` which forces the flag to go in the `lib` section instead of the end of the command.

@rainers 

Ah ok, I didn't know you could only pass 1 .def file.  So then the only argument in favor of my current implementation is PLA (principle of least astonishment).  I'd rather make developers' lives easier.  It's more clear that `-L:lib:foo\` would put `foo\` in the "lib" section of optlink's command then `-L+foo\`.

marler8997:
Going through my old PRs.

We have 3 proposals to specify a library search path.  A library search path means you must put a directory (with a trailing backslash `\`) in the "lib" section of OPTLINK's command line.

For example, if you wanted to add the paths "somepath\mylibs" and "anotherpath\morelibs", you would use the following command line:
```
LINK ,,,somepath\mylibs\+anotherpath\morelibs\
```

Note that the trailing slash is required for OPTLINK to distinguish between library paths and library files.


### Proposal 1

`-L+foo\`

This option proposed by @rainers uses the combination of both the `+` character and the trailing backslash to put `foo\` in OPTLINK's lib section. The backslash implies it's a directory, and no other options accept directories so we can assume this must have bee a library search path.  It will automatically remove the `+` prefix if it is the first argument in the "lib section".

### Proposal 2
 `-L:libfoo\`

The `:lib` prefix allows the caller to explicitly declares that whatever follows should be added as an argument in the "lib section".  It will automatically delimit all arguments with a `+` character if necessary.

### Proposal 3
 `-L:lib:foo`

Same as `-L:libfoo` except it's a bit easier to read.

marler8997:
Ok, I've decided to implement Proposal 1 `-L+foo\` so that I can at least get this integrated.  Should be ready to review.

src/dmd/link.d
```diff
@@ -68,35 +68,46 @@ version (CRuntime_Microsoft)
...
- * Write filename to cmdbuf, quoting if necessary.
- */
+/**
+Returns: true if the given filename needs a quote
+*/
+private bool optlinkNeedsQuote(const(char)[] filename)
+{
+    foreach (c; filename)
+    {
+        if (!isalnum(c) && c != '_' && c != '.' && c != '\\')
```
rainers: Still conservative, but the existing version is even more conservative. Especially with respect to `\` string escaping changes with qoutes, so we should probably just stick to the exisiting test.

src/dmd/link.d
```diff
@@ -337,11 +348,38 @@ public int runLINK()
...
+
+            static bool isLibrarySearchPath(const(char)* option)
             {
-                if (i)
-                    cmdbuf.writeByte('+');
-                writeFilename(&cmdbuf, global.params.libfiles[i]);
+                return option[0] == '+' && option[strlen(option) - 1] == '\\';
+            }
+            {
+                bool first = true;
```
rainers: I think it would be better to move this code into a separate function to avoid adding even more code to this large function. It would also get rid of the slightly irritating indentation.

test/compilable/libpath.sh
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env bash
+
+set -ueo pipefail
+
+TEMP_LIB_PATH=${RESULTS_DIR}/compilable/templibpath
```
rainers: You need back slashes for optlink unconditionally. The test log also shows that adding a quote after the trailing back slash escapes the quote, so the trailing backslash needs to be moved after the quote.

rainers (changes requested):
I like the solution  ;-) but it still needs a few fixes.

marler8997:
For now this PR is stalled because I don't have a way in the test to know if `dmd` is using `OPTLINK` or `MSVC`.  PR https://github.com/dlang/dmd/pull/7838 would provide a mechanism for this, so we'll have to wait for that one before continuing with this one (unless anyone has any other ideas).

rainers:
> For now this PR is stalled because I don't have a way in the test to know if dmd is using OPTLINK or MSVC

You can check the MODEL: 32 -> OMF/optlink, 64 or 32mscoff -> COFF/mslink

marler8997:
I'm closing in favor of adding a general `-libpath=` option that works for all linkers.

PR #7488, opened 2017-12-22
dwarf.c refactor: loop variables

WalterBright:
This also creates a new struct `Section` used to consolidate multiple related global variables into aggregates. Names of `Section` instances adjusted to match the names of the corresponding sections names in the object file.

Future updates should focus on greatly reducing the use of global variables in this file.

### DIFF
--- a/src/dmd/backend/dwarf.c
+++ b/src/dmd/backend/dwarf.c
@@ -457,28 +457,40 @@ void dwarf_CFA_args_size(size_t sz)
     cfa_buf.writeuLEB128(sz);
 }
 
-// .debug_pubnames
-static IDXSEC debug_pubnames_secidx;
-static Outbuffer *debug_pubnames_buf;
+struct Section
+{
+    segidx_t seg;
+    IDXSEC secidx;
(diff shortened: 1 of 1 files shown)


JinShil (approved):

WalterBright:
@JinShil I changed the file extensively after you approved it, please take another look. Thanks!

JinShil (approved):
The "loop variables" in the description is awfully misleading, but the refactor is all pretty trivial stuff.  I imagine you are just trying to get things set up so it's easier to convert to D, yes?

JinShil:
btw, this is not the first time I've seen that Darwin failure on the auto-tester (https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=2955137&isPull=true).  It's probably pass if you force the auto-tester to restart.

WalterBright:
It's partly motivated because I want to convert it to D, but am still blocked. So I'm making it more "D-ish". But it's more because the use of global variables is inefficient, excessive, confusing, and just substandard practice.

PR #7489, opened 2017-12-22
Fix DIP urls in dmd help message

JinShil:
See http://forum.dlang.org/post/lrearyeerwmapheeoyjb@forum.dlang.org

The GitHub URLs were a bit long so I used the [GitHub URL shortener](https://github.com/blog/985-git-io-github-url-shortener)

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -131,9 +131,9 @@ Where:
   -deps            print module dependencies (imports/file/version/debug/lib)
   -deps=<filename> write module dependencies to filename (only imports)" ~
   "%s" /* placeholder for fpic */ ~ "
-  -dip25           implement http://wiki.dlang.org/DIP25 (experimental)
-  -dip1000         implement http://wiki.dlang.org/DIP1000 (experimental)
-  -dip1008         implement DIP1008 (experimental)
+  -dip25           implement http://wiki.dlang.org/DIP25
+  -dip1000         implement https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md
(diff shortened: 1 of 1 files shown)


WalterBright:
Please do not use URL shorteners. They are not necessary, and hide where the link is going.

JinShil:
> Please do not use URL shorteners.

Ok, I've updated this PR to use the genuine URL.

src/dmd/mars.d
```diff
@@ -131,9 +131,9 @@ Where:
   -deps            print module dependencies (imports/file/version/debug/lib)
   -deps=<filename> write module dependencies to filename (only imports)" ~
   "%s" /* placeholder for fpic */ ~ "
-  -dip25           implement http://wiki.dlang.org/DIP25 (experimental)
-  -dip1000         implement http://wiki.dlang.org/DIP1000 (experimental)
-  -dip1008         implement DIP1008 (experimental)
+  -dip25           implement DIP25 http://wiki.dlang.org/DIP25 (experimental)
+  -dip1000         implement DIP1000 https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md (experimental)
+  -dip1008         implement DIP1008 https://github.com/dlang/DIPs/blob/master/DIPs/DIP1008.md (experimental)
```
WalterBright: The middle `DIP1008` of the three can be removed, because at some point it is pretty obviously `DIP1008`. Same for the other two. The `(experimental)` is also redundant.

JinShil: Ok, Removed redundancies

wilzbach (approved):

wilzbach:
Now we would just need a pipeline to automatically update
- https://dlang.org/dmd.html
- the MAN pages

:innocent:

PR #7495, opened 2017-12-22
Fix warning about dangling else using Clang 9

jacob-carlborg:
This only occurs when compiling DMD in debug mode.


wilzbach (approved):
So next step is to do a `BUILD=debug` on Travis too, s.t. this can be catched?

jacob-carlborg:
I think it depends on which version of Clang that is used.

jacob-carlborg:
There are still some warnings that are triggered when enabling `ENABLE_WARNINGS`.

PR #7496, opened 2017-12-22
Fix issue 18111: use fully qualified name for modules in unittest names

atilaneves:


src/dmd/dmodule.d
```diff
@@ -1318,6 +1318,21 @@ extern (C++) final class Module : Package
     {
         v.visit(this);
     }
+
+    /***********************************************
+     * Writes this module fully-qualified name to buf
+     */
+    void fullyQualifiedName(ref OutBuffer buf)
+    {
+
```
JinShil: I don't see any reason for the extra newline.

JinShil (approved):

wilzbach:
Shouldn't there be a test that ensures that the issue can't happen in the future?

atilaneves:
@wilzbach I thought of adding a test specific to this, but ended up not doing so because changing the way unit tests are named would at least make some other tests break (since they unfortunately depend on the exact name of unit tests). It's also unlikely to change back so I decided it wasn't worth the effort.

wilzbach:
Ok as 2.078 is about to be released in two days, I think this is too potentially disturbing for a last minute merge and not a regression with absolute priority. Hence, I changed the target branch to master.

@atilaneves I think there is only one nit missing here

atilaneves:
@wilzbach @JinShil I think that should be it. I also rebased.

src/dmd/dmodule.d
```diff
@@ -1319,6 +1319,20 @@ extern (C++) final class Module : Package
     {
         v.visit(this);
     }
+
+    /***********************************************
+     * Writes this module fully-qualified name to buf
+     */
+    void fullyQualifiedName(ref OutBuffer buf)
```
JinShil: Needs a DDoc `Params`.  Sorry I didn't mention it earlier.

src/dmd/func.d
```diff
@@ -3482,6 +3484,21 @@ extern (C++) final class UnitTestDeclaration : FuncDeclaration
         return Identifier.idPool(buf.peekSlice());
     }
 
+    /*************************************************************************
+     * Writes a module name to name a unittest. Tries to use the fully
+     * qualified name if possible to avoid mismatches when compiling separately.
+     * Otherwise uses the file name.
+     */
+    private static void writeModuleNameOrFileName(ref OutBuffer buf, Loc loc, Scope* scope_)
```
JinShil: Needs a DDoc `Params`.

JinShil (changes requested):

atilaneves:
@JinShil Added the DDoc.

JinShil (approved):
LGTM though I'd like to see additional support before merging.

src/dmd/dmodule.d
```diff
@@ -1319,6 +1319,22 @@ extern (C++) final class Module : Package
     {
         v.visit(this);
     }
+
+    /***********************************************
+     * Writes this module fully-qualified name to buf
```
WalterBright: module's

src/dmd/dmodule.d
```diff
@@ -1319,6 +1319,22 @@ extern (C++) final class Module : Package
...
+     *    buf = The buffer to write to
+     */
+    void fullyQualifiedName(ref OutBuffer buf)
+    {
+        buf.writestring(ident.toString());
+
+        for (auto package_ = parent; package_ !is null; package_ = package_.parent)
+        {
+            buf.prependstring(".");
+            buf.prependstring(package_.ident.toChars());
```
WalterBright: `.toString()` should work here, and is much more efficient

WalterBright (approved):
Approved provided my suggestions are addressed.

src/dmd/dmodule.d
atilaneves: Oops. Fixed.

src/dmd/dmodule.d
atilaneves: `prependstring` takes a `const(char)*`. I _could_ call `toString` and pass its pointer, but I don't see why nor do I understand why `toString` would be more efficient since the implementations of `toString` and `toChars` are nearly identical:

```d
    override const(char)* toChars() const pure
    {
        return string;
    }

    extern (D) final const(char)[] toString() const pure
    {
        return string[0 .. len];
    }
```

atilaneves:
Can I change the target branch back to stable now that 2.078.0 has been released?

wilzbach:
> Can I change the target branch back to stable now that 2.078.0 has been released?

Sure. Done.

PR #7497, opened 2017-12-22
Fix warnings about uninitialized variables

jacob-carlborg:
These warnings occur if the `ENABLE_WARNINGS` environment variable is set when compiling.

### DIFF
--- a/src/dmd/backend/cg87.c
+++ b/src/dmd/backend/cg87.c
@@ -680,7 +680,7 @@ __body
     targ_float f;
     targ_double d;
     targ_ldouble ld;
-    int sz;
+    int sz = -1;
     int zero;
     void *p;
     static char zeros[sizeof(longdouble)];
@@ -1955,7 +1955,7 @@ ret0:   return 0;
 void eq87(CodeBuilder& cdb,elem *e,regm_t *pretregs)
 {
         code cs;
-        unsigned op1;
+        unsigned op1 = -1;
         unsigned op2;
 
         //printf("+eq87(e = %p, *pretregs = %s)\n", e, regm_str(*pretregs));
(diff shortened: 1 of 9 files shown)


jacob-carlborg:
I basically initialized everything to `-1`, I'm not sure if that was correct.

WalterBright:
I personally dislike initializing variables to random values just to shut up the warnings. It degrades code quality to do this.

WalterBright:
I disapprove of this change. The purpose of compiling with warnings on is to improve the code. The changes in this PR do not improve the code.

jacob-carlborg:
For all of these cases the variables were initialized in multiple switch cases but it was missing an initialization for the default case where it would it an `assert(0)`. Do you prefer to move the initialization to the default cases?

WalterBright:
> I basically initialized everything to -1, I'm not sure if that was correct.

Please be sure of what you're doing and be able to defend it, before submitting it.

I do appreciate you doing the work to improve the code quality, but this PR isn't ready for review.

WalterBright:
Consider:
```
int i;
if (condition)
    i = 3;
else
    i = 4;
```
If the compiler warns that "i is uninitialized before use" then the compiler's warnings are crappily designed, and that warning should be disabled.

jacob-carlborg:
That's not what is happening. More something like this:

```c
int i;
switch (x)
{
    case 1:
        i = 2;
        break;
    case 2:
        i = 4;
    default:
        foo();
}
printf("%d\n", i); // here i is not initialized
```

WalterBright:
Please point to a specific line? and I'll have a look.

jacob-carlborg:
Ok, I checked and it's the custom assert implementation that causes the problem. It's missing the `__attribute__((noreturn))` attribute here [1]. Is there a reason to not use the one supplied by the system?

[1] https://github.com/dlang/dmd/blob/master/src/dmd/backend/tassert.h#L27

WalterBright:
No reason to not use it. Please add it! This is the right kind of solution. When this is all fixed up, I'll be pleased to reopen.

jacob-carlborg:
@WalterBright I created a new PR https://github.com/dlang/dmd/pull/7498, this one failed to update.

PR #7500, opened 2017-12-23
VS detection: fallback to lld-link and the mingw libs if no SDK/VC installation found

rainers:
- uses lld-link.exe alongside dmd.exe if no MS link.exe found
- switches the default MS C runtime to msvcrt100 instead libcmt if no VC installation found
- adds mingw subdir of lib64/lib32mscoff as a library search path if no Windows SDK found
- uses the version of link.exe that is looks most appropriate regarding host and target architecture
- improves detection whether legacy_stdio_definitions.lib needs to be added to linker options

### DIFF
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -1041,8 +1041,8 @@ version (Windows)
                     // debug info needs DLLs from $(VSInstallDir)\Common7\IDE for most linker versions
                     //  so prepend it too the PATH environment variable
                     const char* path = getenv("PATH");
-                    auto pathlen = strlen(path);
-                    auto addpathlen = strlen(addpath);
+                    const pathlen = strlen(path);
+                    const addpathlen = strlen(addpath);
 
                     char* npath = cast(char*)mem.xmalloc(5 + pathlen + 1 + addpathlen + 1);
                     memcpy(npath, "PATH=".ptr, 5);
@@ -1055,7 +1055,7 @@ version (Windows)
             }
 
             // try lld-link.exe alongside dmd.exe
-            char[MAX_PATH + 1] dmdpath;
+            char[MAX_PATH + 1] dmdpath = void;
             if (GetModuleFileNameA(null, dmdpath.ptr, dmdpath.length) <= MAX_PATH)
             {
                 auto lldpath = FileName.replaceName(dmdpath.ptr, "lld-link.exe");
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -966,7 +966,7 @@ version (Windows)
         const(char)* linkOptions(bool x64)
         {
             OutBuffer cmdbuf;
-            if (auto vclibdir = getVCDir(VCDir.Lib, x64))
+            if (auto vclibdir = getVCLibDir(x64))
             {
                 cmdbuf.writestring(" /LIBPATH:\"");
                 cmdbuf.writestring(vclibdir);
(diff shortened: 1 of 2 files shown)


skl131313:
If lld is going to be included now an alternative to optlink, changing the compiler flags to be uniform might be a good idea. So -m32 and -m64 use msvc/lld-link and then a third -m32optlink using optlink instead of having -m32mscoff.

ZombineDev:
@skl131313 changing the meaning of existing command-line flags is not possible, since it would be a breaking change.

skl131313:
@ZombineDev 
Not being possible and deterring from implementing something are two entirely different meanings. Breaking changes still get implemented (see integer promotion for + - ~). Optlink should be deprecated as no one is even willing to commit fixes for it and as a result is introducing limitations into DMD. This implements an alternative for it that can be shipped with DMD.

wilzbach:
> Optlink should be deprecated as no one is even willing to commit fixes for it and as a result is introducing limitations into DMD. This implements an alternative for it that can be shipped with DMD.

Yes, please. I don't use Windows, but I still hear from many problems about OPTLINK. As an example it only supports 32k symbols and thus we have to resort to crazy workarounds for Phobos: https://github.com/dlang/phobos/pull/5939

wilzbach:
2.078 will be released soon. Should this be part of it? I'm not sure. On one side these fixes look very important for people on Windows, on the other side there is a risk for regressions and we have already passed the beta release phase. Other opinions?

rainers:
Yes, doesn't make much sense without also having https://github.com/dlang/installer/pull/281 which is probably delayed until next release.

src/dmd/link.d
```diff
@@ -1003,30 +1025,39 @@ version (Windows)
...
+                    memcpy(npath + 5, addpath, addpathlen);
+                    npath[5 + addpathlen] = ';';
+                    memcpy(npath + 5 + addpathlen + 1, path, pathlen + 1);
                     putenv(npath);
                 }
                 return cmdbuf.extractString();
             }
 
+            // try lld-link.exe alongside dmd.exe
+            char[MAX_PATH + 1] dmdpath;
```
WalterBright: `= void;`

src/dmd/link.d
```diff
@@ -1003,30 +1025,39 @@ version (Windows)
...
                 cmdbuf.writestring(p);
                 cmdbuf.writestring(r"\link.exe");
-                if (VSInstallDir)
+                if (addpath)
                 {
                     // debug info needs DLLs from $(VSInstallDir)\Common7\IDE for most linker versions
                     //  so prepend it too the PATH environment variable
                     const char* path = getenv("PATH");
-                    const char* idepath = FileName.combine(VSInstallDir, r"Common7\IDE");
                     auto pathlen = strlen(path);
```
WalterBright: `const pathlen =`

src/dmd/link.d
```diff
@@ -1003,30 +1025,39 @@ version (Windows)
...
-                    auto idepathlen = strlen(idepath);
+                    auto addpathlen = strlen(addpath);
```
WalterBright: `const addpathlen =`

src/dmd/mars.d
```diff
@@ -471,8 +471,12 @@ Language changes listed by -transition=id:
     }
     static if (TARGET_WINDOS)
     {
-        if (!global.params.mscrtlib)
-            global.params.mscrtlib = "libcmt";
+        if (global.params.mscoff && !global.params.mscrtlib)
+        {
+            VSOptions vsopt;
+            vsopt.initialize ();
```
WalterBright: no space before `()`

rainers:
Rebased and addressed review comments.

JinShil:
> Yes, doesn't make much sense without also having dlang/installer#281 

Labeling as "Blocked" then.

wilzbach:
@rainers just saw your post on the NG: http://forum.dlang.org/post/p4883m$1nv1$1@digitalmars.com

> This PR https://github.com/dlang/dmd/pull/7500 would have fixed that by changing the detection to the existence of legacy_stdio_definitions.lib in the VC library path. Unfortunately it didn't make it into the release.

So should we go ahead with this PR? It looks like this PR already makes sense - even without the installer modifications.

rainers:
I guess the largest side-effect would be https://github.com/dlang/dmd/pull/7500/files#diff-12efbb2a88b713dbb0fd0a64673ba132R475: if no VC is detected, the default runtime library is switched from libcmt to msvcrt100 which only exists with the updated installer.

I could extract that to a different PR, though.

wilzbach:
Let's ping @MartinNowak to get the installed updated before the 2.079 release then.

src/dmd/link.d
```diff
@@ -1003,30 +1025,39 @@ version (Windows)
...
                     putenv(npath);
                 }
                 return cmdbuf.extractString();
             }
 
+            // try lld-link.exe alongside dmd.exe
+            char[MAX_PATH + 1] dmdpath = void;
+            if (GetModuleFileNameA(null, dmdpath.ptr, dmdpath.length) <= MAX_PATH)
+            {
+                auto lldpath = FileName.replaceName(dmdpath.ptr, "lld-link.exe");
```
MartinNowak: Likely will break if we shipped a 32-bit bin/lld.exe with a 64-bit bin64/dmd.exe.

MartinNowak (approved):

src/dmd/link.d
rainers: We might also add a 64-bit lld.exe in that case.

src/dmd/link.d
```diff
@@ -1260,6 +1365,12 @@ version (Windows)
                 else if (!x64 && FileName.exists(FileName.buildPath(sdk, "kernel32.lib"))) // SDK 7.1 or earlier
                     return sdk;
             }
+
+            // try mingw fallback relative to phobos library folder that's part of LIB
+            Strings* libpaths = FileName.splitPath(getenv("LIB"));
+            if (auto p = FileName.searchPath(libpaths, r"mingw\kernel32.lib", false))
```
rainers: This doesn't match the 2.079-beta1 installer that places all platform libraries alongside phobos64.lib. Was this a deliberate choice?

PR #7501, opened 2017-12-23
Fix test17868 sporadic failures

JinShil:
Followup to #7490

This is an attempt to fix some sporadic failures for the test17868 postscript.

For example:
https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=2958606&isPull=true
https://travis-ci.org/dlang/dmd/jobs/319656240
https://travis-ci.org/dlang/dmd/jobs/319656246
etc... (there are others)

My Diagnosis.
There are actually two tests `runnable/test17868.d` and `runnable/test17868b.d`, but they both use the same postscript.  The postscript is both creating and removing a file with the same name.  Therefore, if the two tests run at approximately the same time, one can remove the file from the other.

My Solution.
I modified the execution of the post script to take the file name to create as one of its arguments.  `test17868.d` will call the postscript with a file name of `test17868.d.out.2` and `test17868b.d` will call the postscript with a file name of `test17868b.d.out.2`.  Therefore, when the postscript from either test deletes its file, it won't impact the other test.

### DIFF
--- a/test/runnable/extra-files/test17868-postscript.sh
+++ b/test/runnable/extra-files/test17868-postscript.sh
@@ -2,11 +2,11 @@
 
 # trim off the first line which contains the path of the file which differs between windows and non-windows
 # also trim off compiler debug message
-grep -v 'runnable\|DEBUG' $1 > ${RESULTS_DIR}/runnable/test17868.d.out.2
+grep -v 'runnable\|DEBUG' $2 > $1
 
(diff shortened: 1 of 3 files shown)


wilzbach (approved):
Nice. Thanks a lot for doing the detective work!

jacob-carlborg:
The post script still doesn't do proper error handling, i.e. `set -e` 😞.

PR #2130, opened 2013-06-04
Issue 1983 - Delegates violate const

9rnsr:
http://d.puremagic.com/issues/show_bug.cgi?id=1983

<del>Depends on #2125, #2123, and https://github.com/D-Programming-Language/phobos/pull/1331</del>.
1. Add _ambiguous type_ to represent the type which comes from overloaded symbol.
2. Carry overload information properly, and use it.

PR #7503, opened 2017-12-23
Add Type.tambig and make ready to use.

ibuclaw:
First commit of #2130.  I've checked the PR, and fixing all mentioned issues ([1983](https://issues.dlang.org/show_bug.cgi?id=1983), [7322](https://issues.dlang.org/show_bug.cgi?id=7322), [7549](https://issues.dlang.org/show_bug.cgi?id=7549), [8868](https://issues.dlang.org/show_bug.cgi?id=8868), [9027](https://issues.dlang.org/show_bug.cgi?id=9027)) depends on this first if we are to take on the original patch.

I've replaced the handling of `isAmbiguous` in dmangle and hdrgen with asserts, as I do not know whether they are reachable.  See that the type is only used for generating errors, I would suspect not.

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -3377,6 +3377,12 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             }
             else if (t1.ty == Tdelegate)
             {
+                if (t1.isAmbiguous() && exp.e1.op == TOK.delegate_)
+                {
+                    auto de = cast(DelegateExp)exp.e1;
+                    exp.e1 = new DotVarExp(de.e1.loc, de.e1, de.func, de.hasOverloads);
+                    goto Lagain;
+                }
                 TypeDelegate td = cast(TypeDelegate)t1;
                 assert(td.next.ty == Tfunction);
                 tf = cast(TypeFunction)td.next;
@@ -3384,6 +3390,12 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             }
             else if (t1.ty == Tpointer && (cast(TypePointer)t1).next.ty == Tfunction)
             {
+                if (t1.isAmbiguous() && exp.e1.op == TOK.symbolOffset)
+                {
+                    auto se = cast(SymOffExp)exp.e1;
+                    exp.e1 = new VarExp(se.loc, se.var, se.hasOverloads);
+                    goto Lagain;
+                }
                 tf = cast(TypeFunction)(cast(TypePointer)t1).next;
                 p = "function pointer";
             }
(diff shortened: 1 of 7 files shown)


src/dmd/mtype.d
```diff
@@ -1161,6 +1165,11 @@ extern (C++) abstract class Type : RootObject
         return true;
     }
 
+    bool isAmbiguous()
```
JinShil: Please add a DDoc header like that done for `isAssignable()`

JinShil (dismissed):

src/dmd/expressionsem.d
```diff
@@ -5345,6 +5368,12 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         Type t1b = exp.e1.type.toBasetype();
         Type tob = exp.to.toBasetype();
 
+        if (t1b.isAmbiguous() && exp.e1.implicitConvTo(exp.to) <= MATCH.nomatch)
+        {
+            exp.error("cannot cast ambiguous expression %s to %s", exp.e1.toChars(), exp.to.toChars());
```
ibuclaw: Note, missing backticks.

src/dmd/mtype.d
```diff
@@ -992,6 +993,8 @@ extern (C++) abstract class Type : RootObject
         twstring = twchar.immutableOf().arrayOf();
         tdstring = tdchar.immutableOf().arrayOf();
         tvalist = Target.va_listType();
+        tambig = new TypeFunction(null, null, 0, LINKd);
+        tambig.deco = cast(char*)"@".ptr;
```
WalterBright: Dedicating a single character to this is bound to cause problems later. Better to be `__ambiguous`

WalterBright:
I don't see in this PR any examples of what an ambiguous type may be, and according to the code no type is ever set to being ambiguous. I can't evaluate what the case for this is.

ibuclaw:
See the referenced PR.

src/dmd/mtype.d
ibuclaw: I'm not convinced, it should only be used in error diagnostics and so never end up in object code.

JinShil:
@ibuclaw Would you mind accommodating @WalterBright  and adding one of #2130's other commits so he can evaluate further?  IMO this is a very important PR for a couple of reasons, and if this is the right approach (sorry I'm not skilled enough to weigh in on that myself) I'd very much like to move it forward.

wilzbach:
> @ibuclaw Would you mind accommodating @WalterBright and adding one of #2130's other commits so he can evaluate this further?

What's wrong with simply opening https://github.com/dlang/dmd/pull/2130?

andralex:
ping @WalterBright

PR #4688, opened 2015-05-30
Current LDC frontend diff

dnadlinger:
This pull request is not meant to be merged. It is merely there
for @yebblies and others to judge the current state.

PR #7506, opened 2017-12-24
[for information only] LDC diff with DMD: 2.077.1

JohanEngelen:
Copy-paste of LDC master 24 Dec (8c0abd5220e4993d66336b57742c1988dfd73d3f) onto DMD 2.077.1

Continuation of #4688 

Note: parts of the diff have already been applied to DMD master.


src/ddmd/aggregate.h
```diff
@@ -191,7 +191,7 @@ class StructDeclaration : public AggregateDeclaration
     void semantic(Scope *sc);
     void semanticTypeInfoMembers();
     Dsymbol *search(Loc, Identifier *ident, int flags = SearchLocalsOnly);
-    const char *kind();
+    const char *kind() const;
```
ibuclaw: Why?

src/ddmd/attrib.d
```diff
@@ -806,6 +812,21 @@ extern (C++) final class PragmaDeclaration : AttribDeclaration
             }
             return createNewScope(sc, sc.stc, sc.linkage, sc.cppmangle, sc.protection, sc.explicitProtection, sc.aligndecl, inlining);
         }
+        else if (IN_LLVM && ident == Id.LDC_profile_instr) {
```
ibuclaw: Propose a way to add pragmas to the frontend via a hook.  (Maybe #7457 might be the place to put it).

src/ddmd/attrib.h
```diff
@@ -186,6 +186,7 @@ class StaticIfDeclaration : public ConditionalDeclaration
     void addMember(Scope *sc, ScopeDsymbol *sds);
     void setScope(Scope *sc);
     void importAll(Scope *sc);
+    void semantic(Scope *sc);
```
ibuclaw: Why?  Semantic passes should have been removed and visitor()-ized.

src/ddmd/attrib.h
```diff
@@ -199,10 +200,10 @@ class StaticForeachDeclaration : public ConditionalDeclaration
     Dsymbols *cache;
 
     Dsymbol *syntaxCopy(Dsymbol *s);
-    bool oneMember(Dsymbol *ps, Identifier *ident);
+    bool oneMember(Dsymbol **ps, Identifier *ident);
     Dsymbols *include(Scope *sc, ScopeDsymbol *sds);
     void addMember(Scope *sc, ScopeDsymbol *sds);
-    void addComment(const char *comment);
+    void addComment(const utf8_t *comment);
```
ibuclaw: Wasn't this committed recently?

src/ddmd/builtin.d
```diff
@@ -157,6 +160,306 @@ extern (C++) Expression eval_bsr(Loc loc, FuncDeclaration fd, Expressions* argum
     return new IntegerExp(loc, k, Type.tint32);
 }
 
+version(IN_LLVM)
+{
+
+private Type getTypeOfOverloadedIntrinsic(FuncDeclaration fd)
```
ibuclaw: Maybe same as per comment on pragmas.  Need a way to extend front-end built-ins for each compiler (although gdc just ignores this module).

src/ddmd/complex_t.h
```diff
@@ -22,7 +22,7 @@ struct complex_t
     real_t re;
     real_t im;
 
-    complex_t(real_t re) : re(re), im(ldouble(0)) {}
+    complex_t(real_t re) : re(re), im(0) {}
```
ibuclaw: Why don't you implement `ldouble()`?

src/ddmd/constfold.d
```diff
@@ -56,6 +61,17 @@ extern (C++) int isConst(Expression e)
     case TOKnull:
         return 0;
     case TOKsymoff:
+version(IN_LLVM)
+{
+        // We don't statically know anything about the address of a weak symbol
+        // if there is no offset. With an offset, we can at least say that it is
+        // non-zero.
+        SymOffExp soe = cast(SymOffExp) e;
+        if (soe.var.llvmInternal == LDCPragma.LLVMextern_weak && !soe.offset)
```
ibuclaw: Bug  / test for this?  Seems innocuous.

src/ddmd/dcast.d
```diff
@@ -2387,7 +2387,16 @@ extern (C++) Expression castTo(Expression e, Scope* sc, Type t)
 
             if (e2c != e.e2)
             {
-                result = new CommaExp(e.loc, e.e1, e2c);
+                version (IN_LLVM)
+                {
+                    // Cast the CommaExp instead of the inner rhs; this fixes
+                    // https://github.com/ldc-developers/ldc/issues/2415.
+                    result = new CastExp(e.loc, e, e2c.type);
```
ibuclaw: Can't you handle this properly in code-gen?

src/ddmd/declaration.h
```diff
@@ -497,6 +504,26 @@ class FuncDeclaration : public Declaration
...
+
+    // true if overridden with the pragma(LDC_allow_inline); statement
+    bool allowInlining;
+
+    // true if set with the pragma(LDC_never_inline); statement
+    bool neverInline;
+
+    // Whether to emit instrumentation code if -fprofile-instr-generate is specified,
+    // the value is set with pragma(LDC_profile_instr, true|false)
+    bool emitInstrumentation;
```
ibuclaw: Can't you do heuristics at code-gen level?

src/ddmd/dmodule.d
```diff
@@ -590,8 +630,17 @@ extern (C++) final class Module : Package
             if (!strcmp(srcfile.toChars(), "object.d"))
             {
                 .error(loc, "cannot find source code for runtime library file 'object.d'");
+version(IN_LLVM)
+{
+                errorSupplemental(loc, "ldc2 might not be correctly installed.");
+                errorSupplemental(loc, "Please check your ldc2.conf configuration file.");
+                errorSupplemental(loc, "Installation instructions can be found at http://wiki.dlang.org/LDC.");
+}
```
ibuclaw: We need a way to have a per-compiler message here, again with `Compiler` hook.

src/ddmd/dmodule.d
```diff
@@ -1308,6 +1360,21 @@ extern (C++) final class Module : Package
...
+        void checkAndAddOutputFile(File* file);
+        void makeObjectFilenameUnique();
+
+        bool llvmForceLogging;
+        bool noModuleInfo; /// Do not emit any module metadata.
+
+        // Coverage analysis
+        void* d_cover_valid;  // llvm::GlobalVariable* --> private immutable size_t[] _d_cover_valid;
+        void* d_cover_data;   // llvm::GlobalVariable* --> private uint[] _d_cover_data;
+        Array!size_t d_cover_valid_init; // initializer for _d_cover_valid
```
ibuclaw: Can't you attach this data to LLVM AST?

ibuclaw (commented):
There's an awful lot here, and I gave up looking at it. :-)

skl131313:
Why is this being merged into dmd? Unless tests get added for it, it'll constantly be broken and just adds bloat otherwise.

JinShil:
@skl131313 It's just informational.  It is not intended to be merged.

WalterBright:
@ibuclaw @JohanEngelen Since dmd doesn't use the C++ .h files anymore, I'll defer the merging of them to you guys.

ibuclaw:
> Since dmd doesn't use the C++ .h files anymore, I'll defer the merging of them to you guys.

Well, any changes to struct/class layout must also be reflected on the D side too. Otherwise there'd be linker problems, or worse, runtime corruption.

src/ddmd/constfold.d
klickverbot: This is related to an LDC-only pragma, `LDC_extern_weak`, which can be used to implement ELF-style weak symbols (on targets that support them). The frontend would constant-fold away the null checks to see whether they are present, since it assumed that symbol addresses are always non-zero (the `2` below). There is a regression test for it in our lit test suite.

(In fact, the pragma was never publicly documented, as I wanted to have something portable across the compilers instead, but I needed it to implement shared libraries in a way that is compatible with `--gc-sections` due to a certain ld.bfd bug.)

src/ddmd/attrib.h
klickverbot: It's just a merge artefact – benign, as the function is not virtual, but fixed in master nevertheless.

src/ddmd/complex_t.h
kinke: Because it's ugly as hell and a `real_t` constructor accepting 0 makes much more sense. I never understood why you readded it.

ibuclaw: real_t is used in a union, and so must be a POD.  You can't add any constructors to the type.

kinke: That was pre-C++11.

ibuclaw: It is still too soon to be breaking C++98 compatibility.

kinke: [LDC doesn't care and requires C++11.]

klickverbot: [LLVM requires C++11, too.]

ibuclaw: You could initialize it with the constant CTFloat::zero.

(GCC requires C++98 for bootstrapped builds)

andralex:
So what's the story here? There seems to be a lot of good stuff here but it's in a form that's difficult to review.

JinShil:
@JohanEngelen Has this PR served its purpose?  Can you update it, or can we close it?

PR #7507, opened 2017-12-24
Start to generate the CLI options dynamically

wilzbach:
Problem:
- https://dlang.org/dmd.html (there are four pages) are nearly __always__ out of sync
- The MAN pages are nearly __always__ out of sync

Solution: Have one source of truth and generate MAN pages + the dlang.org documentation.
This PR just moves the available CLI flags into a separate module + makes them available as an array for consumption.

As I didn't know whether it would be ok to simply loop over them and use `printf` at runtime, I generated one big string in CTFE, s.t. hopefully this isn't is blocking this first step.

Future steps:
- Make `mpcu` and `transitionUsage` `Options` too
- Use this file to generate the documentation on dlang.org
- Use this file to generate the MAN pages
- Combine this file with `globals.d` and automatically generate `parseCommandLine`


JinShil:
> Make mpcu and transitionUsage Options too
> Combine this file with globals.d and automatically generate parseCommandLine

I've had a couple of bad experiences already approving and merging partial solutions.  Can include the above two items into this pull request.

> Use this file to generate the MAN pages

How are the man pages currently generated?

wilzbach:
> I've had a couple of bad experiences already approving and merging partial solutions. Can you include the above two items into this pull request?

And I made bad experiences with doing too much work in one PR ;-)
I purposefully didn't do more work in this PR, because I wasn't sure whether the way I did it, was ok.

Anyhow, it's not a partial solution, it's an incremental improvement.
With this PR approved/merged. I can already generate the MAN page (we don't list the `transitions` nor `mcpu` there atm). 

> How are the man pages currently generated?

They aren't really _generated_:

https://github.com/dlang/dmd/blob/master/docs/man/man1/dmd.1

They are static files which are rarely updated, because people who add new flags forget about them (to be fair - it's very easy to do so).

JinShil:
Does `cli.d` need to also be added to the `src/vcbuild/dmd.visualdproj` file?

wilzbach:
Added a script that generated the MAN page based on the options array in `cli.d`.
-> I pushed the "updated" version of the MAN file in a separate commit
It comes with a simple Makefile that can be used to `build` or `preview` the MAN page. It does no work if `cli.d` nor `gen_man.d` haven't changed, so it will be easy to integrate it with the main Makefile in a later PR.
I hope the vast differences between the MAN page and actual flags show why we want to automate this.

wilzbach:
Anything that's blocking this? I'm getting tired of rebasing this whenever the CLI options change (and we should have updated the MAN page + dlang.org docs too).
Also I just bumped into three bugzilla issues about outdated MAN pages:
- https://issues.dlang.org/show_bug.cgi?id=14475
- https://issues.dlang.org/show_bug.cgi?id=15679
- https://issues.dlang.org/show_bug.cgi?id=15446

This PR is already complete in itself and a good step forward (the MAN pages are generated).
- Dlang.org's docs are more complicated -> I will submit this in a separate PR
- Automatically rebuilding the MAN page as part of the build might be controversial -> I will submit this in a separate PR

docs/Makefile
```diff
@@ -0,0 +1,23 @@
...
+
+preview:
+	man -l $(MAN_PAGE)
+
+################################################################################
+# Actual compilation commands
+################################################################################
+
+$(MAN_PAGE): $G/gen_man
+	$< > $@
```
wilzbach: The main mainfile can call this target to regenerate the man page when `cli.d` changed, but as I said before, I think the discussion whether that's a good idea or not, shouldn't block this PR and thus belongs into another PR.

wilzbach:
There's more:
- https://issues.dlang.org/show_bug.cgi?id=11985
- https://issues.dlang.org/show_bug.cgi?id=8196

src/dmd/cli.d
```diff
@@ -0,0 +1,383 @@
...
+                    {
+                            buf ~= "  ";
+                    }
+                    buf ~= option.helpText;
+                    buf ~= "\n";
+                }
+            }
+            return buf;
+        }();
+        return s.ptr;
```
JinShil: Can we return a `string` and use `.ptr` at the `printf` call.  I'd like to gradually move away from using `char*`.

JinShil (approved):

src/dmd/cli.d
```diff
@@ -0,0 +1,383 @@
...
+    TargetOS os = TargetOS.all;
+}
+
+/*
+Contains all available CLI options
+*/
+struct Usage
+{
+    // Returns all available CLI options
+    static Option[] options()
```
jacob-carlborg: Looks like this can be a variable instead of a function.

src/dmd/cli.d
```diff
@@ -0,0 +1,383 @@
...
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/cli.d, _cli.d)
+ * Documentation:  https://dlang.org/phobos/dmd_cli.html
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/cli.d
+ */
+module dmd.cli;
+
+// Bit decoding of the TargetOS
+enum TargetOS {
```
jacob-carlborg: Opening curly brace on the next line.

jacob-carlborg:
Some of the comments are Ddoc comments and some are regular comments. Is there a reason for that?

src/dmd/cli.d
```diff
@@ -0,0 +1,383 @@
...
+    windows = 2,
+    macOS = 4,
+    freeBSD = 8,
+    solaris = 16,
+}
+
+// Detect the current TargetOS
+version (linux)
+{
+    enum targetOS = TargetOS.linux;
```
jacob-carlborg: Looks like this can be private.

src/dmd/cli.d
```diff
@@ -0,0 +1,383 @@
...
+else version(Solaris)
+{
+    enum targetOS = TargetOS.solaris;
+}
+else
+{
+    enum targetOS = TarggetOS.all;
+}
+
+
```
jacob-carlborg: Remove extra newline.

src/dmd/cli.d
```diff
@@ -0,0 +1,383 @@
...
+    return (os & targetOS) > 0;
+}
+
+// Representation of a CLI option
+struct Option
+{
+    string flag;
+    string helpText;
+
+    TargetOS os = TargetOS.all;
```
jacob-carlborg: Missing documentation for these three fields.

jacob-carlborg (changes requested):

src/dmd/cli.d
jacob-carlborg: Or even better not use `.ptr` at all and instead use the `printf` formatter that allows to specify pointer and length separately.

wilzbach: Me too, I just used `char*` to keep the changes at a minimum. I'm more than happy to use `string` :)

wilzbach:
> Some of the comments are Ddoc comments and some are regular comments. Is there a reason for that? > Otherwise all public symbols should have Ddoc comments.

Oh I wanted to leave everything undocumented for now as this might still change when I integrate it with dlang.org.
(I removed the Ddoc star for the symbol in question).

OTOH we don't give any guarantees about the API stability at the moment, so we could also simply simple start documenting it. (I don't have a strong opinion here)

src/dmd/mars.d
```diff
@@ -376,29 +283,17 @@ private int tryMain(size_t argc, const(char)** argv)
...
-CPU architectures supported by -mcpu=id:
-  =?             list information on all architecture choices
-  =baseline      use default architecture as determined by target
-  =avx           use AVX 1 instructions
-  =avx2          use AVX 2 instructions
-  =native        use CPU architecture that this compiler is running on
-");
+        import dmd.cli : CLIUsage;
+        auto help = CLIUsage.mcpu;
+        printf("%.*s", help.length, &help[0]);
```
jacob-carlborg: Is there a reason why `&help[0]` is used instead of `help.ptr`?. `.ptr` is more readable in my opinion.

jacob-carlborg:
> OTOH we don't give any guarantees about the API stability at the moment, so we could also simply simple start documenting it. (I don't have a strong opinion here)

I'm in favor of documenting it.

src/dmd/mars.d
wilzbach: Yes `help.ptr` isn't `@safe`

wilzbach:
> I'm in favor of documenting it.

Fair enough. Done.

src/dmd/mars.d
jacob-carlborg: Eh, ok. I didn't know there was a difference.

src/dmd/cli.d
```diff
@@ -0,0 +1,381 @@
...
+else version(Solaris)
+{
+    private enum targetOS = TargetOS.solaris;
+}
+else
+{
+    private enum targetOS = TarggetOS.all;
+}
+
+bool isCurrentTargetOS(TargetOS os)
```
jacob-carlborg: Missing documentation.

src/dmd/cli.d
```diff
@@ -0,0 +1,381 @@
...
+    string helpText; /// A detailed description of the flag
+    TargetOS os = TargetOS.all; /// For which `TargetOS` the flags are applicable
+}
+
+/*
+Contains all available CLI $(LREF Option)s.
+
+See_Also: $(LREF Option)
+*/
+struct Usage
```
jacob-carlborg: Not a Ddco comment.

jacob-carlborg (changes requested):

jacob-carlborg:
I would prefer that the existing commits are amended, then we avoid unnecessary commits like `Address review`. Unless you want to squash the whole thing.

wilzbach:
> I would prefer that the existing commits are amended

Sure. Done. Just did individual commits, s.t. it's easier to follow up.

jacob-carlborg:
> Sure. Done. Just did individual commits, s.t. it's easier to follow up.

You could do fixup commits in that case.

jacob-carlborg:
LGTM 👌.

jacob-carlborg (approved):

jacob-carlborg:
Probably someone else should have a look as well.

wilzbach:
> You could do fixup commits in that case.

- Start to generate the CLI options dynamically
<- I wanted to squashed things here
- Add script to automatically generate the MAN pages
- Use the generated MAN pages

jacob-carlborg:
Fixup commits allow you to fixup any commit and not just the last one.

JinShil (approved):

PetarKirov (approved):
Overall looking great! As discussed with Seb, there are various parts that could be improved, but it's much better do move forward with this and improve it over time.

PR #3572, opened 2014-05-23
Mark if(__ctfe) blocks as SCOPEctfe

jpf91:
Allows e.g. allocating in an if(__ctfe) block in a @nogc function

PR #7509, opened 2017-12-24
Fix Issue 18119 - Mark if(__ctfe) blocks as SCOPEctfe

ibuclaw:
Revival of #3572

### DIFF
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -1930,13 +1930,34 @@ package (dmd) extern (C++) final class StatementSemanticVisitor : Visitor
         // Save 'root' of two branches (then and else) at the point where it forks
         CtorFlow ctorflow_root = scd.ctorflow.clone();
 
+        // Recognize if (__ctfe) as a special form condition and allow allocations
+        // in @nogc functions if we're in a __ctfe branch.
+        if (VarExp ve = ifs.condition.isVarExp())
+        {
+            if (ve.var && ve.var.ident == Id.ctfe)
+                scd.flags |= SCOPE.ctfe;
+        }
         ifs.ifbody = ifs.ifbody.semanticNoScope(scd);
         scd.pop();
 
         CtorFlow ctorflow_then = sc.ctorflow;   // move flow results
         sc.ctorflow = ctorflow_root;            // reset flow analysis back to root
         if (ifs.elsebody)
-            ifs.elsebody = ifs.elsebody.semanticScope(sc, null, null, null);
+        {
+            Scope *sce = sc.push();
+            // Recognize if (!__ctfe) as a special form condition and allow allocations
+            // in @nogc functions if we're in a __ctfe branch.
+            if (ifs.condition.op == EXP.not)
+            {
+                if (VarExp ve = ifs.condition.isUnaExp().e1.isVarExp())
+                {
+                    if (ve.var && ve.var.ident == Id.ctfe)
+                        sce.flags |= SCOPE.ctfe;
+                }
+            }
+            ifs.elsebody = ifs.elsebody.semanticNoScope(sce);
+            sce.pop();
+        }
 
         // Merge 'then' results into 'else' results
         sc.merge(ifs.loc, ctorflow_then);
--- a/changelog/fix18119.dd
+++ b/changelog/fix18119.dd
@@ -1,0 +1,41 @@
+fix Issue 18119 - Allow code that may allocate inside __ctfe condition branches in @nogc functions
+
+The following conditions are now recognized as a specialization during compile-time.
+
+---
+if (__ctfe)
+{
+}
+
+if (!__ctfe)
+{
+}
+---
+
+Where any code inside the CTFE branch that can be evaluated at compile-time is now permissible
+in @nogc functions regardless of whether an allocation would occur or not during runtime.
+
+The following is an example of a function that benefits from this enhancement taken from druntime.
+
+---
+@trusted pure nothrow
+const(ubyte)[] toUbyte(T)(T[] arr)
+    if ((is(typeof(toUbyte(arr[0])) == const(ubyte)[])) && (T.sizeof > 1))
+{
+    if (__ctfe)
+    {
+        const(ubyte)[] ret;
+        foreach (cur; arr)
+        {
+            ret ~= toUbyte(cur);
+        }
+        return ret;
+    }
+    else
+    {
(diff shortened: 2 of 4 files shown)


stefan-koch-sociomantic:
this is nasty special casing, then again this is actually useful.
please propose a spec change @ibuclaw.

ibuclaw:
@stefan-koch-sociomantic - I've put in a changelog entry that gives a real world example of who might benefit.

stefan-koch-sociomantic:
please also put that an expression involving __ctfe even it it is `0 + __ctfe` will not mark the scope.

wilzbach:
> please propose a spec change

Imho this is not a spec change, but an improvement of the compiler like e.g. optimization.

src/dmd/statementsem.d
```diff
@@ -2147,7 +2155,19 @@ else
...
         if (ifs.elsebody)
-            ifs.elsebody = ifs.elsebody.semanticScope(sc, null, null);
+        {
+            Scope *sce = sc.push();
+            // Recognize if (!__ctfe) as a special form condition and allow allocations
+            // in @nogc functions if we're in a __ctfe branch.
+            if (ifs.condition.op == TOKnot && (cast(UnaExp)ifs.condition).e1.op == TOKvar)
+            {
+                VarExp ve = cast(VarExp)(cast(UnaExp)ifs.condition).e1;
+                if (ve.var && ve.var.ident == Id.ctfe)
```
JinShil: I'm seeing an opportunity to centralize some of this logic into a function.  It's already used twice in this function scope and also at https://github.com/dlang/dmd/blob/master/src/dmd/expressionsem.d#L7299.  Perhaps a `VarExp.is__ctfe()` member function?

JinShil (changes requested):

jacob-carlborg:
This might make it easier to get allocators to work at CTFE 😃.

JinShil:
I don't have any answers at this time, but there seems to be a fundamental design problem in the language if we have to resort to something like this.  I'd like to see some brainstorming for a more elegant solution.

src/dmd/statementsem.d
ibuclaw: Maybe would be better in the if condition ast, and return a tristate to represent: no, yes if, yes else.

Such a function may benefit from caching also.  But is not necessary.

ibuclaw:
@JinShil - That's part of why rebasing these old PRs is supposed to achieve.  Debate whether they are sound ideas that may need some TLC (or a better implementation), or reject + close related issues.

UplinkCoder:
so .... thinking about this again, in theory we should be able to constant-fold everything around `__ctfe`  in cases where marking the scope can be done and/or simplifying the expression to only depend on `__ctfe`.

ibuclaw:
@UplinkCoder - what do you mean?

stefan-koch-sociomantic:
@jacob-carlborg I'll rather support __ctfeAlloc and __ctfeFree.

jacob-carlborg:
> I'll rather support __ctfeAlloc and __ctfeFree.

@stefan-koch-sociomantic Doesn't matter to me, as long as then end goal works.

ibuclaw:
Why add a new keyword though?

stefan-koch-sociomantic:
it would not be a keyword. rather a very magic function :)

stefan-koch-sociomantic:
and we already have precedent for this. namely builtins.

stefan-koch-sociomantic:
it removes the need gc in ctfe.
a @nogc function can not be ctfe since ctfe as no way of allocating memory.

ibuclaw:
Ctfe should not be a separate language to runtime. I don't think any of these __ctfeXxx function ideas are any good.

schveiguy:
I'd like to see this feature added. Can this be revived? There didn't seem to be any opposition. This would be a first step to opening a lot more things to CTFE, e.g. any functions that use `RefCounted`.

schveiguy:
> but there seems to be a fundamental design problem in the language if we have to resort to something like this.

I don't think so. Anything run at compile-time *can't* use the GC, as it doesn't exist. This is, IMO, more of a correction than a kludge. Really, the only kludgy part is that it needs a specialized form (but then again, the specialized form is nearly always what anyone uses).

UplinkCoder:
@ibuclaw since you don't like ctfe to be another langauge, we could "fake" a malloc call at ctfe.

n8sh:
> it would not be a keyword. rather a very magic function :)

You don't need any magic to implement CTFE alloc:
https://github.com/dlang/druntime/blob/246d15ac85b2bc7ae7d7a7f61df2ceb80377ed8c/src/core/internal/convert.d#L13-L34
```d

/+
A @nogc function can allocate memory during CTFE.
+/
@nogc nothrow pure @trusted
private ubyte[] ctfe_alloc()(size_t n)
{
    if (!__ctfe)
    {
        assert(0, "CTFE only");
    }
    else
    {
        static ubyte[] alloc(size_t x) nothrow pure
        {
            if (__ctfe) // Needed to prevent _d_newarray from appearing in compiled prorgam.
                return new ubyte[x];
            else
                assert(0);
        }
        return (cast(ubyte[] function(size_t) @nogc nothrow pure) &alloc)(n);
    }
}
```

schveiguy:
@n8sh there's a lot happening in there (it's a template, you have a seemingly unnecessary check for CTFE at the beginning, and then again later). I don't know if I'd consider that not magic.

But it does provide a path forward as a workaround. I'd still rather have the requirements for `@nogc` CTFE simply be that you can allocate inside an `if(__ctfe)` block.

schveiguy:
Wait, I didn't notice the cast. That's interesting.

schveiguy:
I tested and it works for my purpose (using GC to allocate in CTFE, C malloc otherwise)

n8sh:
>(it's a template, you have a seemingly unnecessary check for CTFE at the beginning, and then again later).

The second __ctfe check is so it can be used with `-betterC`.

schveiguy:
No, I misunderstood the example. I thought there was some specific magic going on, but really, it is just casting in `@nogc` and ensuring the compiler doesn't actually emit any GC calls using `if(__ctfe)`. The other parts of the example aren't important (that it's a template, that it asserts __ctfe at the beginning). I tested it separately, and it works exactly as I would want.

rainers:
Just stumbled over this, an alternative was merged by #23211

schveiguy:
@rainers awesome! Thanks for the fix.

PR #4200, opened 2014-12-08
Use in contract from base class if not specified in derived

nemanja-boric-sociomantic:
In case when in contract is not specified in the derived class,
and base class does specify a contract, base contract should
not be silently ignored:

```
class Base
{
    void foo()
    in
    {
        assert(false);
    }
    body
    {

    }
}

class Derived: Base
{
    override void foo()
    {
    }
}

void main()
{
    auto o = new Derived();
    o.foo(); // Should fail
}

```

Old behaviour can be brought back with `in{}` in `Derived.foo`.

See https://issues.dlang.org/show_bug.cgi?id=6856

PR #7510, opened 2017-12-24
fix Issue 6856 - Use in contract from base class if not specified in derived

ibuclaw:
Supersedes #4200.

### DIFF
--- a/src/dmd/semantic3.d
+++ b/src/dmd/semantic3.d
@@ -279,9 +279,6 @@ private extern(C++) final class Semantic3Visitor : Visitor
         }
 
         uint oldErrors = global.errors;
-        auto fds = FuncDeclSem3(funcdecl,sc);
-
-        fds.checkInContractOverrides();
 
         // Remember whether we need to generate an 'out' contract.
         immutable bool needEnsure = FuncDeclaration.needsFensure(funcdecl);
@@ -882,6 +879,9 @@ private extern(C++) final class Semantic3Visitor : Visitor
 
             funcdecl.frequire = funcdecl.mergeFrequire(funcdecl.frequire);
             funcdecl.fensure = funcdecl.mergeFensure(funcdecl.fensure, Id.result);
+
+            auto fds = FuncDeclSem3(funcdecl,sc);
+            fds.checkInContractOverrides();
 
             Statement freq = funcdecl.frequire;
             Statement fens = funcdecl.fensure;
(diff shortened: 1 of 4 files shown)


ibuclaw:
I think I've weeded out all failures now.  However in my local environment, it looks like 64bit has some terrible stack corruption bug that did not happen when I initially tested the change with gdc.

ibuclaw:
Hmm, now the testsuite confirms the stack corruption that I'm seeing.

ibuclaw:
@WalterBright - why would there be stack corruption calling the base `in()` contract? And why does no stack corruption happen when you compile with `-g`?

ibuclaw:
Blocked by dmd-specific backend bug.

src/dmd/semantic3.d
```diff
@@ -857,6 +844,20 @@ extern(C++) final class Semantic3Visitor : Visitor
             funcdecl.frequire = funcdecl.mergeFrequire(funcdecl.frequire);
             funcdecl.fensure = funcdecl.mergeFensure(funcdecl.fensure, funcdecl.outId);
 
+            if (funcdecl.fdrequire)
+            {
+                for (size_t i = 0; i < funcdecl.foverrides.dim; i++)
```
wilzbach: Nit: `foreach (i; 0 .. funcdecl.foverrides.dim)` ?

wilzbach (commented):
Wow! This would be really useful to have :)

wilzbach:
(rebased)

wilzbach (commented):
I think the DMD backend error is gone, but it looks like there have been other changes in DMD in the meanwhile too and thus the tests fail.
Also, now that multiple contracts are possible, it probably is a good idea to cover this with tests.

ibuclaw:
Rebased, lets see...

thewilsonator:
The stack corruption is probably https://issues.dlang.org/show_bug.cgi?id=15984

thewilsonator:
@ibuclaw Ping, https://issues.dlang.org/show_bug.cgi?id=15984 was fixed.

ibuclaw:
@thewilsonator - Nope, only seems to be partially fixed.  I no longer get a runtime assert failure if I compile with `-g`, however.

ibuclaw:
Added variable 'x' in the printf, so you can see the runtime corruption.

thewilsonator:
Hmm, oh well.

dkorpel:
This issue came up in Mike's [gripes and wishes](https://forum.dlang.org/post/rcnrixuppxflnrheyvsz@forum.dlang.org) 

@ibuclaw What's the status of this?

PR #7511, opened 2017-12-24
refactor kind(): add attributes

WalterBright:
Note that overriding `kind()` functions will inherit those attributes.

### DIFF
--- a/src/dmd/dimport.d
+++ b/src/dmd/dimport.d
@@ -6,11 +6,11 @@
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
  * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/dimport.d, _dimport.d)
+ * Documentation:  https://dlang.org/phobos/dmd_dimport.html
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/dimport.d
  */
 
 module dmd.dimport;
-
-// Online documentation: https://dlang.org/phobos/dmd_dimport.html
 
 import dmd.arraytypes;
 import dmd.declaration;
@@ -101,7 +101,7 @@ extern (C++) final class Import : Dsymbol
 
     override const(char)* kind() const
     {
-        return isstatic ? cast(char*)"static import" : cast(char*)"import";
+        return isstatic ? "static import" : "import";
     }
 
     override Prot prot()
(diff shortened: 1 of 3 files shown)


src/dmd/dimport.d
```diff
@@ -6,12 +6,12 @@
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
  * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/dimport.d, _dimport.d)
+ * Documentation:  https://dlang.org/phobos/ddmd_dimport.html
```
BBasile: currently the link works but since the package was renamed _dmd_ a few days ago this is _dmd_dimport.html_.

WalterBright: dmd_dimport.html currently gives "404 Not Found"

wilzbach: Yes, the transition just happened a few days ago and 2.078.0 hasn't been released yet. The link will work in [seven days](https://dlang.org/changelog/2.078.0.html).

JinShil (approved):

src/dmd/dimport.d
```diff
@@ -101,7 +101,7 @@ extern (C++) final class Import : Dsymbol
 
     override const(char)* kind() const
     {
-        return isstatic ? cast(char*)"static import" : cast(char*)"import";
+        return isstatic ? &"static import"[0] : &"import"[0];
```
jacob-carlborg: String literals are implicitly converted to `const(char)*`.

src/dmd/func.d
```diff
@@ -3027,7 +3027,7 @@ extern (C++) final class FuncLiteralDeclaration : FuncDeclaration
     override const(char)* kind() const
     {
         // GCC requires the (char*) casts
-        return (tok != TOKfunction) ? cast(char*)"delegate" : cast(char*)"function";
+        return (tok != TOKfunction) ? &"delegate"[0] : &"function"[0];
```
jacob-carlborg: Same here, string literals are implicitly converted to `const(char)*`.

jacob-carlborg (dismissed):

src/dmd/dimport.d
JinShil: Same thing here: https://github.com/dlang/dmd/blob/master/src/dmd/target.d#L9-L10

src/dmd/dsymbol.d
```diff
@@ -585,7 +585,7 @@ extern (C++) class Dsymbol : RootObject
         return s;
     }
 
-    const(char)* kind() const
+    const(char)* kind() const pure nothrow @nogc @safe
```
ibuclaw: Are the c++ headers marked as `const() const` too?

WalterBright: This PR does not change the `const` attributes on `kind()`.

JinShil (approved):

PR #7513, opened 2017-12-25
Refactor the Objective-C glue layer

jacob-carlborg:
This is a followup on the refactoring of the frontend part. This moves the interface of the Objective-C glue layer to a class to make it possible to determine at runtime (of the compiler) if Objective-C is
supported or not. This allows cross-compiling and the `objc_glue_stubs.d` file can be removed.

I split the PR in several commits that can be reviewed individually, which should be much easier then reviewing everything at once.


ibuclaw:
Could use a little squashing of commits.

src/dmd/objc_glue.d
```diff
@@ -39,273 +37,343 @@ import dmd.backend.type;
...
-        sy = objc_getCString(s, len, namestr.ptr, ObjcSegment.SEGmethname);
-        sv.ptrvalue = sy;
+        auto segmentsPtr = segments.ptr;
+
+        if (segmentsPtr[id] != 0)
+            return segmentsPtr[id];
+
+        foreach (i, seg ; segmentData)
+        {
+            segmentsPtr[i] = MachObj.getsegment(
```
ibuclaw: Do you really need to use `MachObj` here? I thought that dmd encapsulated all object related things into `Obj`.

ibuclaw (changes requested):

src/dmd/objc_glue.d
jacob-carlborg: No, I forgot to change that. That's also why the tests are failing.

jacob-carlborg:
> Could use a little squashing of commits.

I could squash the commits but I deliberately split the changes into multiple commits to make it easier to review. But if you prefer them squash I'll do that.

src/dmd/objc_glue.d
jacob-carlborg: @ibuclaw  Ok, I see now that `getsegment` is declared in the subclasses, like `MachObj` and `ElfObj`, and not in the base class `Obj`. `getsegment` has different signatures for each object format so it looks like it cannot be moved to the base class. Any suggestions?

ibuclaw:
> I could squash the commits but I deliberately split the changes into multiple commits to make it easier to review. But if you prefer them squash I'll do that.

Of what I had looked at, each individual commit seemed reasonable.

src/dmd/objc_glue.d
ibuclaw: As discussed, best course of immediate action would be to use `version(OSX)` to enclose the call to `MachObj` behind.  The `else` branch can just be a runtime `assert(0)` as no other target platform of dmd will get this far.

jacob-carlborg: That's a good idea.

jacob-carlborg: Done.

jacob-carlborg:
@ibuclaw Please let me know if/when I should squash the commits.

jacob-carlborg:
I created a PR [1] for removing `objc_glue.d` from the list of excluded files when generating the documentation.

[1] https://github.com/dlang/dlang.org/pull/2039

jacob-carlborg:
@ibuclaw All the tests (except the documentation) are green and I squashed the commits.

wilzbach:
I just merged the respective dlang.org PR, so after rebasing your PR DAutoTest should be happy too.
Regarding #7567 as discussed on Slack, simply remove `version(OSX)` for `dmd/backend/mach.d` (the backend is already platform-independent).

jacob-carlborg:
@wilzbach any advice on the failing documentation build? Was the PR for dlang.org not correct?

wilzbach:
> @wilzbach any advice on the failing documentation build? Was the PR for dlang.org not correct?

Yep, the exclusion check uses `DMD_DIR` and not `DMD_LATEST_DIR`. I'm sorry that I didn't notice this:
Fix -> https://github.com/dlang/dlang.org/pull/2041

jacob-carlborg:
@ibuclaw All tests are passing now, please have a look.

src/dmd/objc_glue.d
```diff
@@ -38,273 +37,353 @@ import dmd.backend.type;
...
+            }
+        }
+
+        return segmentsPtr[id];
     }
-    return sy;
 }
 
-Symbol *objc_getMethVarName(Identifier *ident)
+struct Symbols
```
ibuclaw: Since `Symbol` is already a type meaning in dmd (`Symbols` being an `Array!Symbol` by convention), maybe call this `ObjcSymbols`?

ibuclaw: Probably ok though as it's private, just a note to say that it could have a grep friendly name.

src/dmd/objc_glue.d
```diff
@@ -38,273 +37,353 @@ import dmd.backend.type;
...
-        objc_segList[s] = 0;
+    override void generateModuleInfo()
+    {
+        if (Symbols.hasSymbols)
+            Symbols.getModuleInfo();
+    }
 }
 
-Symbol *objc_getCString(const(char)* str, size_t len, const(char)* symbolName, ObjcSegment segment)
+struct Segments
```
ibuclaw: Could have an Objc prefix too, but not necessary.

ibuclaw (commented):
Needs rebasing again.

src/dmd/objc_glue.d
jacob-carlborg: That kind of defeats the purpose of the D modules system.

ibuclaw: As I said, something that is grep-able or simple to do automated translation with is a nice to have. If another module declares the same struct name, it's not so trivial to determine who uses which version.

But meh.

jacob-carlborg:
Thanks.

PR #7514, opened 2017-12-25
fix warnings about bad pointer related conversions

rainers:
VC build for Win64: disables Win32 C code that is not used by dmd and is wrong for Win64.

### DIFF
--- a/src/dmd/backend/os.c
+++ b/src/dmd/backend/os.c
@@ -80,7 +80,7 @@ void os_error(int line)
 #pragma noreturn(os_error)
 #endif
 
-#if _WIN32
+#if _WIN32 && !_WIN64
 /*********************************
  * Allocate a chunk of memory from the operating system.
  * Bypass malloc and friends.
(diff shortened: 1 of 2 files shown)


src/dmd/backend/os.c
```diff
@@ -80,7 +80,7 @@ void os_error(int line)
 #pragma noreturn(os_error)
 #endif
 
-#if _WIN32
+#if _WIN32 && !_WIN64
```
JinShil: Why are `_WIN32` and `_WIN64` both `true`?

ibuclaw: https://msdn.microsoft.com/en-us/library/b0084kay.aspx

> _WIN32 Defined as 1 when the compilation target is 32-bit ARM, 64-bit ARM, *x86*, or *x64*. Otherwise, undefined.

> _WIN64 Defined as 1 when the compilation target is 64-bit ARM or *x64*. Otherwise, undefined.

JinShil (approved):

wilzbach (approved):

PR #7505, opened 2017-12-24
Revert "fix Issue 18053 - Use stdint.h mangling for int64_t/uint64_t when man…"

wilzbach:
Reverts dlang/dmd#7416 due to the Travis error on OSX, see e.g. https://travis-ci.org/dlang/dmd/jobs/320910100

```
Undefined symbols for architecture x86_64:
  "mem_malloc(unsigned long long)", referenced from:
      array_toDarray(Type*, elem*) in dmd.o
      toElem::ToElemVisitor::visit(StringExp*) in dmd.o
  "symboldata(unsigned long, unsigned int)", referenced from:
      el_ptr(Symbol*) in backend.a(el.o)
      el_convstring(elem*) in backend.a(el.o)
      out_readonly_sym(unsigned int, void*, int) in backend.a(out.o)
      Obj::sym_cdata(unsigned int, char*, int) in backend.a(machobj.o)
  "_align(unsigned long, unsigned long)", referenced from:
      codgen(Symbol*) in backend.a(cgcod.o)
      stackoffsets(int) in backend.a(cgcod.o)
      outjmptab(block*) in backend.a(cod3.o)
      outswitab(block*) in backend.a(cod3.o)
      type_paramsize(TYPE*) in backend.a(type.o)
      alignOffset(int, unsigned long) in backend.a(out.o)
      out_readonly_sym(unsigned int, void*, int) in backend.a(out.o)
      ...
  "ispow2(unsigned long long)", referenced from:
      asmSemantic(AsmStatement*, Scope*) in dmd.o
  "outthunk(Symbol*, Symbol*, unsigned int, unsigned int, unsigned long long, int, unsigned long long)", referenced from:
      toThunkSymbol(FuncDeclaration*, int) in dmd.o
  "Obj::ehtables(Symbol*, unsigned long long, Symbol*)", referenced from:
      except_gentables() in dmd.o
  "DtBuilder::size(unsigned long long)", referenced from:
      Initializer_toDt::InitToDt::visit(ArrayInitializer*) in dmd.o
      Expression_toDt::ExpToDt::visit(StringExp*) in dmd.o
      Expression_toDt::ExpToDt::visit(ArrayLiteralExp*) in dmd.o
      cpp_type_info_ptr_toDt(ClassDeclaration*, DtBuilder*) in dmd.o
[...]

PR #7515, opened 2017-12-25
Remove default clang-specific warnings and add '-xc++'

ibuclaw:
These aren't needed any more for the remainder sources.

`-xc++` is required as:
```
clang: error: treating 'c' input as 'c++' when in C++ mode, this behavior is deprecated
```

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -188,13 +188,6 @@ endif
 else
 # Default Warnings
 WARNINGS := -Wno-deprecated -Wstrict-aliasing -Werror
-# Clang Specific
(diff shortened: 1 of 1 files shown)


JinShil (approved):

wilzbach (approved):

wilzbach:
Travis is still broken, see e.g. https://github.com/dlang/dmd/pull/7505

ibuclaw:
Hmm, apparently I messed something up in my initial local testing.  As now I'm able to get those errors.

These global environment vars are really finicky to use.  I'll have a look later...

PR #7517, opened 2017-12-25
Remove unneeded clang warning flags.

ibuclaw:
Fixed version of #7515

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -176,15 +176,6 @@ WARNINGS += \
 	-Wno-unused-but-set-variable \
 	-Wno-uninitialized
 endif
-# Clang Specific
-ifeq ($(HOST_CXX_KIND), clang++)
-WARNINGS += \
-	-Wno-tautological-constant-out-of-range-compare \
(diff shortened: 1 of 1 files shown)


JinShil (approved):

wilzbach (approved):

src/posix.mak
```diff
@@ -211,6 +200,11 @@ ifeq ($(CXX_KIND), g++)
 CXXFLAGS += \
     -std=gnu++98
 endif
+# Clang Specific
+ifeq ($(CXX_KIND), clang++)
```
wilzbach: FYI `g++` supports this flag too (and actually my local `clang++` doesn't trigger a warning without it). In any case I don't think there's any harm in always letting `-xc++` be part of the `CXXFLAGS`.

ibuclaw: I must have been doing something weird the first time I ran make with cxx=clang because I couldn't reproduce after the a couple runs, maybe I added or removed something before resetting the branch, such as deprecation warnings.

PR #7518, opened 2017-12-25
Support latest g++ when ENABLE_WARNINGS is on

ibuclaw:
Minimum changes needed to build.  Slightly borrowed from #7475.

### DIFF
--- a/src/dmd/backend/blockopt.c
+++ b/src/dmd/backend/blockopt.c
@@ -713,7 +713,7 @@ void brcombine()
                     {
                         if (EOP(b3->Belem))
                             continue;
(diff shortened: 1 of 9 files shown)


JinShil (approved):

PetarKirov (approved):

wilzbach:
Thanks a lot @ibuclaw!
Should we add `ENABLE_WARNINGS` now to CircleCi (or others) or is the conversion of the backend really so imminent?

jacob-carlborg:
I think we should give it a try.

wilzbach:
> I think we should give it a try.

https://github.com/dlang/dmd/pull/7529

PR #7520, opened 2017-12-25
Replace C++ syntax check with dummy frontend

ibuclaw:
On closer inspection the Visitor interface broken by #7397 still wasn't fixed by #7438.

The entire C++ Visitor class has been re-layed out to match the order of virtuals in the D implementation.

Have exported `Id::initialize` to C++ as fundamental components don't work if you can't call it.

Replaced cxxcheckheaders with cxx-unittest, a small program that links against the front-end in the same way that LDC or GDC would to test that interaction works.

I anticipate to build on this test file later.

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -266,8 +266,8 @@ endif
 endif
 
 # Unique extra flags if necessary
-DMD_FLAGS  := -I$(ROOT) -Wuninitialized
-GLUE_FLAGS := -I$(ROOT) -I$(TK) -I$(C)
+DMD_FLAGS  := -I$(D) -I$(ROOT) -Wuninitialized
+GLUE_FLAGS := -I$(D) -I$(ROOT) -I$(TK) -I$(C)
 BACK_FLAGS := -I$(ROOT) -I$(TK) -I$(C) -I$G -I$D -DDMDV2=1
 ROOT_FLAGS := -I$(ROOT)
 
@@ -396,7 +396,7 @@ TK_SRC = \
 	$(TK)/filespec.h $(TK)/mem.h $(TK)/list.h $(TK)/vec.h \
 	$(TK)/filespec.c $(TK)/mem.c $(TK)/vec.c $(TK)/list.c
 
-######## CXX header files (only needed for checkcxxheaders)
+######## CXX header files (only needed for cxx-unittest)
 
 SRC = $(addprefix $D/, aggregate.h aliasthis.h arraytypes.h	\
 	attrib.h complex_t.h cond.h ctfe.h ctfe.h declaration.h dsymbol.h	\
@@ -423,7 +423,7 @@ DEPS = $(patsubst %.o,%.deps,$(DMD_OBJS) $(GLUE_OBJS) $(BACK_OBJS))
 
 all: $G/dmd
 
-auto-tester-build: $G/dmd checkwhitespace checkcxxheaders $G/dmd_frontend
+auto-tester-build: $G/dmd checkwhitespace cxx-unittest $G/dmd_frontend
 .PHONY: auto-tester-build
 
 toolchain-info:
@@ -569,11 +569,14 @@ $(TOOLS_DIR)/checkwhitespace.d:
 
 ######################################################
 
-# See https://github.com/dlang/dmd/pull/7358 for why -xc++ is used here
-# -O is needed for FreeBSD's compiler to silence "-Wuninitialized is not supported without -O"
-checkcxxheaders: $(ROOT_SRC) $(SRC)
-	$(HOST_CXX) -xc++ -O -fsyntax-only $(ROOT_FLAGS) $(filter %.h,$(ROOT_SRC))
-	$(HOST_CXX) -xc++ -O -fsyntax-only $(DMD_FLAGS) $(filter %.h,$(SRC))
+$G/cxxfrontend.o: $G/%.o: tests/%.c $(SRC) $(ROOT_SRC)
+	$(CXX) -c -o$@ $(CXXFLAGS) $(DMD_FLAGS) $(MMD) $<
+
+$G/cxx-unittest: $G/cxxfrontend.o $(DMD_SRCS) $(ROOT_SRCS) $G/newdelete.o $G/lexer.a $(G_GLUE_OBJS) $(G_OBJS) $(STRING_IMPORT_FILES) $(HOST_DMD_PATH)
+	CC=$(HOST_CXX) $(HOST_DMD_RUN) -of$@ $(MODEL_FLAG) -vtls -J$G -J../res -L-lstdc++ $(DFLAGS) -version=NoMain $(filter-out $(STRING_IMPORT_FILES) $(HOST_DMD_PATH),$^)
+
+cxx-unittest: $G/cxx-unittest
+	$<
 
 ######################################################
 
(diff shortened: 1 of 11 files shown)


ibuclaw:
@MartinNowak - this should go in stable + 2.078, as the C++<->D visitor interface is utterly unusable without this, and breaks all downstream.

ibuclaw:
Looks like I should partially revert #7005 also.

wilzbach:
> @MartinNowak - this should go in stable + 2.078, as the C++<->D visitor interface is utterly unusable without this, and breaks all downstream.

Simple target `stable`, then it will be part of 2.078 (you can change the targeted branch directly via the GitHub UI, but you will need to do the `git rebase` too).
Once merge we can easily merge `stable` back into `master`.

ibuclaw:
Rebased against stable.  I think the tests should pass now once I fixed up `ForwardingStatement` visit methods - @tgehr.

tgehr:
I'm not sure why #7005 needs to be reverted. (Or why it is at all virtuous to not be "lazy" -- there is exactly one sensible default implementation, therefore #7005 prevents all further bugs like the one it fixed.)

src/dmd/sapply.d
```diff
@@ -170,6 +170,12 @@ public:
     {
         doCond(s.statement) || applyTo(s);
     }
+
+    override void visit(ForwardingStatement s)
+    {
+        if (s.statement)
+            doCond(s.statement) || applyTo(s.statement);
```
tgehr: I don't know the exact purpose of the walkPostorder visitor, but this seems fishy to me as ForwardingStatement is not supposed to have behaviour different from the wrapped statement.

ibuclaw: This is correct, as you want to walk over the wrapped statement, not the statement itself.

tgehr: If I have a ForwardingStatement wrapping a LabelStatement, doCond is called with different arguments than if I have only a LabelStatement. Before your changes, the behaviour was identical for both cases. Is this change intentional?

ibuclaw:
>  there is exactly one sensible default implementation, therefore #7005 prevents all further bugs like the one it fixed.

The base visitor should just forward onto the next method in the inheritance chain. Any special logic above that should go in overriding visitors.

tgehr:
Well, then the best fix might be to add a DefaultVisitor in-between Visitor and all its subclasses. To forward to `visit(Statement)` by default is to do the wrong thing by default. Everybody who wants to introduce a new subclass of Visitor now needs to be aware of ForwardingStatement. This is a trap that can lead to bugs.

src/dmd/sapply.d
ibuclaw: Do you have code example that changes behaviour? Or is that just hypothetical?

The order that it walks the AST should be the same as the original.

Should probably add `applyTo(s)` if the underlying statement is null, but that does not happen in the current.

tgehr: Running the visitor on a semantically-analyzed version of the following code will result in changed behaviour:
```d
static foreach(i;0..1) L: writeln(i);
```
`L: writeln(i)` will be visited by the `StoppableVisitor v` twice now, even though before it was visited only once.

I don't think `applyTo(s)` should be called if `s.statement` is `null`.

ibuclaw:
> To forward to visit(Statement) by default is to do the wrong thing by default. Everybody who wants to introduce a new subclass of Visitor now needs to be aware of ForwardingStatement. This is a trap that can lead to bugs.

Having it in visitor just hid bugs. As it wasn't forwarding onto Statement in the event that the underlying wrapped statement was null.

If your writing a visitor, that you should be aware of all classes is a given. If the entire point of ForwardingStatement is to not exist outside semantic then that needs to be looked at and fixed.  If we can do away with ForwardingStatement in any way (I don't know what it's used for, so I'll arbitrarily suggest adding a Scope flag) then all the better.

src/dmd/sapply.d
ibuclaw: So then, `doCond(s.statement) || applyTo(s);`

tgehr:
Which bugs? If it is sensible to ignore a null Statement, it is sensible to ignore a ForwardingStatement wrapping null.

ForwardingStatement is used to insert wrapped symbols into a different scope than the current one.
It could probably be removed during semantic analysis, if that is a useful thing to do.

src/dmd/sapply.d
tgehr: Only if it makes sense to Visit the ForwardingStatement by default. (I don't see why it would.)
Is this Visitor used for anything?

ibuclaw:
> It could probably be removed during semantic analysis, if that is a useful thing to do.

It seems to me that this is not something that should be leaked outside of the static foreach semantic, so yes.

ibuclaw:
> Which bugs? If it is sensible to ignore a null Statement, it is sensible to ignore a ForwardingStatement wrapping null.

If I have a visitor with the following method.
```
class MV : public Visitor
{
  void visit(Statement *) { puts("have statement"); }
}
```
Passing this a ForwardingStatement with interior null would print nothing, which is an unexpected behaviour.

src/dmd/sapply.d
ibuclaw: It's used by `comeFrom` and/or `blockExit` for determining reachable statements.

tgehr:
I'll try to create a pull request that removes ForwardingStatements after they are no longer required.
Knowing the semantics of ForwardingStatement, that behaviour is expected, but point taken.

src/dmd/sapply.d
tgehr: Visiting the ForwardingStatement will break the `hasCode` method.

ibuclaw: Right, that sounds like what I saw last night when seeing unreachable warnings.

So just the one `applyTo(s.statement` then to elide double visiting.

tgehr: I think it should just be
```
if(s.statement)
    s.statement.accept(this);
```
(Otherwise, e.g. bodies of while loops inside `static foreach` will not be visited.)

ibuclaw:
I moved the ForwardingStatement fix to #7526

ibuclaw:
Cherry-picked other PR into this one.

ZombineDev:
@ibuclaw #7526 has been merged. Rebase?

ibuclaw:
@ZombineDev already rebased.

ZombineDev:
> @ZombineDev already rebased.

Hmm, I missed that #7526 was targeting master instead of stable, unlike this PR.

src/posix.mak
```diff
@@ -266,8 +266,8 @@ endif
 endif
 
 # Unique extra flags if necessary
-DMD_FLAGS  := -I$(ROOT) -Wuninitialized
-GLUE_FLAGS := -I$(ROOT) -I$(TK) -I$(C)
+DMD_FLAGS  := -I$(D) -I$(ROOT) -Wuninitialized
+GLUE_FLAGS := -I$(D) -I$(ROOT) -I$(TK) -I$(C)
```
wilzbach: The common style of this Makefile is to skip the parentheses for one letter variables, i.e. `$D`

src/tests/cxxfrontend.c
```diff
@@ -0,0 +1,199 @@
+/**
```
wilzbach: Creating the folder `src/tests` might lead to confusions for newcomers.

src/posix.mak
ibuclaw: All I have to say is `$(C)`.

src/tests/cxxfrontend.c
ibuclaw: I wasn't sure the best place for this, as the main source folder doesn't seem right.

MartinNowak:
> Cherry-picked other PR into this one.

Please avoid cherry-picking if possible to not unnecessary merge conflicts.

ibuclaw:
@MartinNowak so having the source in  tests is fine?

PR #7521, opened 2017-12-26
Implement -probe for showing supported features

wilzbach:
DUB creates a special [platform probe file _on every invocation_](https://github.com/dlang/dub/blob/cea2f15c34ac9493c335f39b8c0cfcaabd31a597/source/dub/compilers/compiler.d#L130), ([more details](https://github.com/dlang/dub/blob/cea2f15c34ac9493c335f39b8c0cfcaabd31a597/source/dub/compilers/utils.d#L245)).

(`determinePlatform` is called by [`setupPackage`]( https://github.com/dlang/dub/blob/cea2f15c34ac9493c335f39b8c0cfcaabd31a597/source/dub/commandline.d#L622) in [`execute`](https://github.com/dlang/dub/blob/cea2f15c34ac9493c335f39b8c0cfcaabd31a597/source/dub/commandline.d#L1032))

While DUB should cache this invocation, I think there should be a "standard" format for feature detection instead of using this hack to generate a D file with `version` identifiers.
This PR generates the same output as the generated DUB probing file, e.g.:

```json
{

    "compiler": "dmd",
    "frontendVersion": 2078,
    "compilerFrontend": "v2.078.0-beta.1-113-g057aa4442",
    "config": null,
    "binary": "/home/seb/dlang/dmd/generated/linux/release/64/dmd",
    "platform": [
        "posix",
        "linux"
    ],
    "architecture": [
        "x86_64"
    ],
    "predefinedVersions": [
        "DigitalMars",
        "Posix",
        "linux",
        "ELFv1",
        "LittleEndian",
        "D_Version2",
        "all",
        "D_SIMD",
        "D_InlineAsm_X86_64",
        "X86_64",
        "CRuntime_Glibc",
        "D_LP64",
        "D_PIC",
        "assert",
        "D_HardFloat"
    ]
}
```

(I didn't add this flag to the documentation, s.t. it can stay an experiment)

### DIFF
--- a/src/dmd/probing.d
+++ b/src/dmd/probing.d
@@ -1,0 +1,161 @@
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * This modules helps to create a list of supported features (aka probing output).
+ * The probing information is returned in JSON.
+ *
+ * Copyright:   Copyright (c) 1999-2018 by The D Language Foundation, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/probing.d, _probing.d)
+ * Documentation:  https://dlang.org/phobos/dmd_probing.html
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/probing.d
+ */
+
+module dmd.probing;
+
+/**
+Prints a list of supported features (aka probing output) to stdout.
+The output is in JSON.
+*/
+void printProbingInfo()
+{
+    import core.stdc.stdio : printf;
+    import dmd.globals : global;
+
+    printf("{\n");
+    bool needsComma;
+    void printKey(const(char*) key, void function() value)
+    {
+        if (needsComma)
+            printf(",");
+        printf("\n");
+        needsComma = true;
+        printf("    \"%s\": ", key);
+        value();
+    }
+
+    printKey("compiler", (){
+        printf(`"%s"`, determineCompiler());
+    });
+    printKey("frontendVersion", (){
+        printf("%d", versionXX());
+    });
+    printKey("compilerFrontend", (){
+        printf(`"%s"`, global._version);
+    });
+    printKey("config", (){
+        if (*global.inifilename != '\0')
+            printf(`"%s"`, global.inifilename);
+        else
+            printf("null");
+    });
+    printKey("binary", (){
+        printf(`"%s"`, global.params.argv0);
+    });
+    printKey("platform", (){
+        printJSONArray(&determinePlatform);
+    });
+    printKey("architecture", (){
+        printJSONArray(&determineArchitecture);
+    });
+    printKey("predefinedVersions", (){
+        printJSONArray(&predefinedVersions);
+    });
+    printf("\n}\n");
+}
+
+private long versionXX()
+{
+    import dmd.lexer : parseVersionXX;
+    import dmd.globals : global;
+    return global._version.parseVersionXX;
+}
+
+private const(char*) determineCompiler()
+{
+    import core.stdc.string : strcmp;
+    import dmd.globals : global;
+
+    if (strcmp(global.compiler.vendor, "Digital Mars D") == 0)
+        return "dmd";
+    else if (strcmp(global.compiler.vendor, "LDC") == 0)
+        return "ldc";
+    else if (strcmp(global.compiler.vendor, "GNU") == 0)
+        return "gdc";
+    else if (strcmp(global.compiler.vendor, "SDC") == 0)
+        return "sdc";
+
+    return null;
+};
+
+private void printJSONArray(void function(void delegate(string)) fun)
+{
+    import core.stdc.stdio : printf;
+    static immutable const(char*) indent = "    ";
+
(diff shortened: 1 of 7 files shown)


wilzbach:
(CC @s-ludwig who is using this in DUB)

src/dmd/probing.d
```diff
@@ -0,0 +1,175 @@
...
+        printf(`%s%s"%s"`, indent, indent, s.ptr);
+        isFirst = false;
+    }
+    fun(&print);
+    printf("\n%s]", indent);
+}
+
+private void determinePlatform(void delegate(string) print)
+{
+	version(Windows) print("windows");
```
ibuclaw: This is wrong.  Don't you want to print *target* information, not host?

src/dmd/probing.d
```diff
@@ -0,0 +1,175 @@
...
+	version(SysV4) print("sysv4");
+	version(Hurd) print("hurd");
+	version(Android) print("android");
+	version(Cygwin) print("cygwin");
+	version(MinGW) print("mingw");
+}
+
+private void determineArchitecture(void delegate(string) print)
+{
+	version(X86) print("x86");
```
ibuclaw: Likewise here.

src/dmd/probing.d
```diff
@@ -0,0 +1,207 @@
...
+ *
+ * This modules helps to create a list of supported features (aka probing output).
+ * The probing information is returned in JSON.
+ *
+ * Copyright:   Copyright (c) 1999-2017 by The D Language Foundation, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/mars.d, _mars.d)
+ * Documentation:  https://dlang.org/phobos/dmd_mars.html
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/mars.d
```
JinShil: This documentation header seems to be cut and pasted from `mars.d`, but needs to be updated for this file.

src/dmd/probing.d
```diff
@@ -0,0 +1,207 @@
...
+}
+
+// TODO: unite with global.compiler.vendor
+private const(char*) determineCompiler()
+{
+	version(DigitalMars) return "dmd";
+	else version(GNU) return "gdc";
+	else version(LDC) return "ldc";
+	else version(SDC) return "sdc";
+	else return null;
```
mathias-lang-sociomantic: That also looks wrong

src/dmd/probing.d
```diff
@@ -0,0 +1,207 @@
...
+	version(SPARC64) print("sparc64");
+	version(S390) print("s390");
+	version(S390X) print("s390x");
+	version(HPPA) print("hppa");
+	version(HPPA64) print("hppa64");
+	version(SH) print("sh");
+	version(SH64) print("sh64");
+	version(Alpha) print("alpha");
+	version(Alpha_SoftFP) print("alpha_softfp");
+	version(Alpha_HardFP) print("alpha_hardfp");
```
mathias-lang-sociomantic: Then leave it out ? No point in faking support, especially given those targets are more likely to be subject to cross compilation.

s-ludwig:
Wouldn't slightly a more generic functionailty make more sense, where simply all predefined versions are dumped to stdout?

ibuclaw:
> Wouldn't slightly a more generic functionailty make more sense, where simply all predefined versions are dumped to stdout?

They are with `-v`.

wilzbach:
> They are with -v.

It doesn't do this anymore: https://github.com/dlang/dmd/pull/7549

> Wouldn't slightly a more generic functionailty make more sense, where simply all predefined versions are dumped to stdout?

Good idea (done), but I think you will still need things like `frontendVersion` for DUB too, right?

s-ludwig:
> Good idea (done), but I think you will still need things like frontendVersion for DUB too, right?

True, would be good to avoid another process invocation for that (`--version`), at least on Windows...

wilzbach:
@s-ludwig so `-probe` would work for you? Anything else that's missing/blocking?

timotheecour:
how about a more flexible way to output exactly what we want:
* -probe=version,import,default_version,deps,file=out.txt (etc, easily extensible without breaking client clode)
* it would output in json (easier to parse) (or maybe allow -probe=...,format=txt)
* if file is not specified, goes to stdout
advantage: if some probes are large or slow down compilation, could be wasteful to add them unless they're explicitly asked for.

NOTE: -v can't change too much unfortunately for backward compatibility reasons, but -probe could be well designed since we're starting from scratch and future proof.

marler8997:
graphql for dmd :)

timotheecour:
could -Xf (json output) used instead of introducing another -probe flag? (cf https://github.com/dlang/dmd/pull/7757)

CyberShadow:
I like Timothee's idea from [here](https://github.com/dlang/dmd/pull/7521#issuecomment-359508204), where `dmd -Xf` without input files produces the same information as `-probe` would. It would mean one less switch and one less format, and would let programs that need to compile something and get compiler information to do so with one compiler invocation.

timotheecour:
the link u gave is wrong, I think you meant https://github.com/dlang/tools/pull/292#issuecomment-359532273 :)

marler8997:
Alternative to `-probe` submitted here: https://github.com/dlang/dmd/pull/7838

wilzbach:
Superseded by https://github.com/dlang/dmd/pull/7894

PR #7522, opened 2017-12-26
ispow2: change parameter to uint64_t

WalterBright:


WalterBright:
Of course, this doesn't work because the bootstrap dmd uses the old scheme. Sigh.

WalterBright:
Worked around by having a separate `ispow2` on the D side.

src/dmd/iasm.d
```diff
@@ -4558,3 +4558,20 @@ AFTER_EMIT:
     //return asmstate.bReturnax;
     return s;
 }
+
+/**********************
+ * If c is a power of 2, return that power else -1.
+ */
+
+private int ispow2(uint c)
```
wilzbach: D shouldn't be about duplicating code :(

wilzbach (approved):
I assume we can get rid of this duplication with the "D-ification" of the backend soon.

src/dmd/iasm.d
WalterBright: There's a lot of hackish code that the conversion to D will eliminate.

PR #7524, opened 2017-12-26
Move addDefaultVersionIdentifiers from mars.d to compiler.d

RazvanN7:
The frontend interface of the compiler library needs to do some initialization which is done in dmd/mars.d. In order to use initialization methods from mars.d there are 3 options:

1. Duplicate the code in frontend.d
2. Import mars.d and use the method.
3. Put it in another file and make frontend.d and mars.d use it

(1), (2) are obviously bad solutions, so this PR implements (3)

### DIFF
--- a/src/dmd/compiler.d
+++ b/src/dmd/compiler.d
@@ -12,7 +12,143 @@
 
 module dmd.compiler;
 
+import core.stdc.stdio;
+import dmd.cond;
+import dmd.globals;
+
 struct Compiler
 {
     const(char)* vendor; // Compiler backend name
+}
+
+/**
+ * Add default `version` identifier for dmd, and set the
+ * target platform in `global`.
+ * https://dlang.org/spec/version.html#predefined-versions
+ *
+ * Needs to be run after all arguments parsing (command line, DFLAGS environment
+ * variable and config file) in order to add final flags (such as `X86_64` or
+ * the `CRuntime` used).
+ */
+void addDefaultVersionIdentifiers()
+{
+    VersionCondition.addPredefinedGlobalIdent("DigitalMars");
+    static if (TARGET_WINDOS)
+    {
+        VersionCondition.addPredefinedGlobalIdent("Windows");
+        global.params.isWindows = true;
+    }
+    else static if (TARGET_LINUX)
+    {
+        VersionCondition.addPredefinedGlobalIdent("Posix");
+        VersionCondition.addPredefinedGlobalIdent("linux");
+        VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        global.params.isLinux = true;
+    }
+    else static if (TARGET_OSX)
+    {
+        VersionCondition.addPredefinedGlobalIdent("Posix");
+        VersionCondition.addPredefinedGlobalIdent("OSX");
+        global.params.isOSX = true;
+        // For legacy compatibility
+        VersionCondition.addPredefinedGlobalIdent("darwin");
+    }
+    else static if (TARGET_FREEBSD)
+    {
+        VersionCondition.addPredefinedGlobalIdent("Posix");
+        VersionCondition.addPredefinedGlobalIdent("FreeBSD");
+        VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        global.params.isFreeBSD = true;
+    }
+    else static if (TARGET_OPENBSD)
+    {
+        VersionCondition.addPredefinedGlobalIdent("Posix");
+        VersionCondition.addPredefinedGlobalIdent("OpenBSD");
+        VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        global.params.isOpenBSD = true;
+    }
+    else static if (TARGET_SOLARIS)
+    {
+        VersionCondition.addPredefinedGlobalIdent("Posix");
+        VersionCondition.addPredefinedGlobalIdent("Solaris");
+        VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        global.params.isSolaris = true;
+    }
+    else
+    {
+        static assert(0, "fix this");
+    }
+    VersionCondition.addPredefinedGlobalIdent("LittleEndian");
+    VersionCondition.addPredefinedGlobalIdent("D_Version2");
+    VersionCondition.addPredefinedGlobalIdent("all");
+
+    if (global.params.cpu >= CPU.sse2)
+    {
+        VersionCondition.addPredefinedGlobalIdent("D_SIMD");
+        if (global.params.cpu >= CPU.avx)
+            VersionCondition.addPredefinedGlobalIdent("D_AVX");
+        if (global.params.cpu >= CPU.avx2)
+            VersionCondition.addPredefinedGlobalIdent("D_AVX2");
+    }
+
+    if (global.params.is64bit)
+    {
+        VersionCondition.addPredefinedGlobalIdent("D_InlineAsm_X86_64");
+        VersionCondition.addPredefinedGlobalIdent("X86_64");
+        static if (TARGET_WINDOS)
+        {
+            VersionCondition.addPredefinedGlobalIdent("Win64");
+        }
+    }
+    else
+    {
+        VersionCondition.addPredefinedGlobalIdent("D_InlineAsm"); //legacy
+        VersionCondition.addPredefinedGlobalIdent("D_InlineAsm_X86");
+        VersionCondition.addPredefinedGlobalIdent("X86");
+        static if (TARGET_WINDOS)
+        {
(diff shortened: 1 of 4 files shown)


src/dmd/compiler.d
```diff
@@ -0,0 +1,188 @@
...
+    if (global.params.useAssert == CHECKENABLE.on)
+        VersionCondition.addPredefinedGlobalIdent("assert");
+    if (global.params.useArrayBounds == CHECKENABLE.off)
+        VersionCondition.addPredefinedGlobalIdent("D_NoBoundsChecks");
+    if (global.params.betterC)
+        VersionCondition.addPredefinedGlobalIdent("D_BetterC");
+
+    VersionCondition.addPredefinedGlobalIdent("D_HardFloat");
+
+    printPredefinedVersions();
```
wilzbach: Hmmm we should we really do this by default in this method?
How about moving `printPredefinedVersions` back to `mars.d`?

src/dmd/frontend.d
```diff
@@ -45,6 +46,8 @@ void initDMD()
     else
         static assert(0, "OS not supported yet.");
 
+    addDefaultVersionIdentifiers();
```
wilzbach: You can now remove the OS detection above :)
I had to make a copy of this as `addDefaultVersionIdentifiers` was private and I wanted to keep the changes to a minimum in the initial version.

src/dmd/compiler.d
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * Set the default and debug libraries to link against, if not already set
+ *
+ * Must be called after argument parsing is done, as it won't
+ * override any value.
+ * Note that if `-defaultlib=` or `-debuglib=` was used,
+ * we don't override that either.
+ */
+void setDefaultLibrary()
```
wilzbach: Not sure whether this can be re-used by other compilers or the frontend. It looks fairly specific to DMD.

jacob-carlborg:
Note that there's this PR as well https://github.com/dlang/dmd/pull/7485.

src/dmd/compiler.d
RazvanN7: It's ok to let it there since the printPredefined versions has all it's code in a block nested by a `if (global.params.verbose)`

wilzbach:
@RazvanN7 this needs another rebase

src/dmd/compiler.d
WalterBright: `printPredefinedVersions()` has nothing to do with `addDefaultVersionIdentifiers()` - note that no mention of it is made in the Ddoc comment - it should not be called here.

src/dmd/frontend.d
```diff
@@ -27,24 +27,10 @@ void initDMD()
...
-    else version(Windows)
-        global.params.isWindows = 1;
-    else version(Solaris)
-        global.params.isSolaris = 1;
-    else version(OpenBSD)
-        global.params.isOpenBSD = 1;
-    else
-        static assert(0, "OS not supported yet.");
-
+    addDefaultVersionIdentifiers();
```
WalterBright: `printPredefinedVersions()` call should go here.

src/dmd/frontend.d
```diff
@@ -27,24 +27,10 @@ void initDMD()
...
-    else version(FreeBSD)
-        global.params.isFreeBSD = 1;
-    else version(Windows)
-        global.params.isWindows = 1;
-    else version(Solaris)
-        global.params.isSolaris = 1;
-    else version(OpenBSD)
-        global.params.isOpenBSD = 1;
-    else
-        static assert(0, "OS not supported yet.");
```
WalterBright: I see all this is deleted, but I don't see where it is moved to?

RazvanN7: Now that `addDefaultVersionIdentifiers` is called there is no need to do that since the code was duplicated from it initially.

src/dmd/frontend.d
RazvanN7: Done

RazvanN7:
@wilzbach I addressed the issues. I think we're good to go.

jacob-carlborg:
> I had to add -J../res to the parser and lexer library builds otherwise the tests would fail

Hmm, that's strange. The only file in the `res` directory is `default_ddoc_theme.ddoc` which is only used in `dmd.doc`.

RazvanN7:
Furthermore, globals.d imports compiler.d which for the sake of `addDefaultVersionIdentifiers` imports cond.d which is a gateway to other non parser related files. I guess `compiler.d` should not be part of the parser.

ibuclaw:
I'll remove the import of dmd.compiler from the globals module tomorrow.

ibuclaw:
Shouldn't this be moved to a module like `dmd.driver`?  These functions are useless for other backends.

(Saying that, however, the compiler module won't be shared code).

src/dmd/frontend.d
```diff
@@ -27,24 +27,11 @@ void initDMD()
...
-
+    addDefaultVersionIdentifiers();
+    printPredefinedVersions();
```
wilzbach: Are we sure that we want to call `printPredefinedVersions` here at all?
After all, this is the `frontend` for DMD as a _library_ code.

wilzbach:
> @wilzbach I addressed the issues. I think we're good to go. I had to add -J../res to the parser and lexer library builds otherwise the tests would fail

You shouldn't need to add `-J../res` - this means that your `parser.a` now imports `doc.d`.
See also:

https://github.com/dlang/dmd/pull/7534#issuecomment-354275802
https://github.com/dlang/dmd/pull/7534#issuecomment-355833853

> Shouldn't this be moved to a module like dmd.driver? These functions are useless for other backends.

`dmd.driver` sounds a bit misleading. How about `mars.compiler`, `mars.versions` or `compiler/mars.d`?

andralex:
@WalterBright ok with you?

src/dmd/mars.d
```diff
@@ -2436,3 +2301,5 @@ private void parseModulePattern(const(char)* modulePattern, MatcherNode* dst, us
         }
     }
 }
+=======
+>>>>>>> Move addDefaultVersionIdentifiers to frontend.d
```
andralex: looks like some conflict made it into the pr

andralex (changes requested):
see merge conflict

src/dmd/compiler.d
```diff
@@ -12,7 +12,143 @@
...
+ * variable and config file) in order to add final flags (such as `X86_64` or
+ * the `CRuntime` used).
+ */
+void addDefaultVersionIdentifiers()
+{
+    VersionCondition.addPredefinedGlobalIdent("DigitalMars");
+    static if (TARGET_WINDOS)
+    {
+        VersionCondition.addPredefinedGlobalIdent("Windows");
+        global.params.isWindows = true;
```
WalterBright: This line should really not be here. "Setting predefined identifiers" and "setting the target platform in global" should be in different functions.

src/dmd/compiler.d
```diff
@@ -12,7 +12,143 @@
...
+    }
+    else
+    {
+        static assert(0, "fix this");
+    }
+    VersionCondition.addPredefinedGlobalIdent("LittleEndian");
+    VersionCondition.addPredefinedGlobalIdent("D_Version2");
+    VersionCondition.addPredefinedGlobalIdent("all");
+
+    if (global.params.cpu >= CPU.sse2)
```
WalterBright: Note how this section is using `global` to set versions, whereas the previous section used `TARGET` to set it.  Using TARGET to set global should be in a separate function, then this function should rely only on global to set versions.

src/dmd/compiler.d
```diff
@@ -12,7 +12,143 @@
...
+    else static if (TARGET_SOLARIS)
+    {
+        VersionCondition.addPredefinedGlobalIdent("Posix");
+        VersionCondition.addPredefinedGlobalIdent("Solaris");
+        VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        global.params.isSolaris = true;
+    }
+    else
+    {
+        static assert(0, "fix this");
```
WalterBright: Fixing a static assert failure is implied, so the message adds no value. "unsupported target" might be better.

src/dmd/compiler.d
WalterBright: Furthermore, instead of accessing `global` as a global variable, consider passing a const reference to it to the function.

RazvanN7:
Moving those functions to compiler.d is not a good idea. `globals.d` (which is part of the parser library) imports `compiler.d`. Moving these functions to `compiler.d` will introduce semantic dependencies to the parser library. Closing this PR until I will come up with a better solution.

PR #3611, opened 2014-06-01
[enh] fix issue 6549 - Implement contracts without implementation

WalterBright:
This is a rebase of https://github.com/D-Programming-Language/dmd/pull/2281

PR #7527, opened 2017-12-26
fix issue 6549 - Implement contracts without implementation

ibuclaw:
Supersedes #3611


test/fail_compilation/fail12901.d
```diff
@@ -1,14 +0,0 @@
...
----
-*/
-
-struct S
-{
-    int a;
-    this(int n)
-    in { a = n; }
-    // no body
-}
```
wilzbach: What's the motivation for removing this? Shouldn't it still be forbidden?

test/runnable/testcontracts.d
```diff
@@ -421,6 +421,56 @@ void test6417()
...
+    abstract int foo(int)
+    in { ++icount; }
+    out { ++ocount; }
+}
+
+class CD6549 : C6549
+{
+    override int foo(int)
+    in { assert(false); }
+    body { return 10; }
```
BBasile: use `do` since soon or later someone will have to patch the test.

test/runnable/testcontracts.d
```diff
@@ -421,6 +421,56 @@ void test6417()
...
+    int foo(int)
+    in { ++icount; }
+    out { ++ocount; }
+}
+
+class DD6549 : D6549
+{
+    override int foo(int)
+    in { assert(false); }
+    body { return 10; }
```
BBasile: use `do`.

test/fail_compilation/fail12901.d
ibuclaw: Not sure as this is would be the third person to rebase and submit this patch.

It certainly seems reasonable to me to not include body-less ctors, dtors, and postblits to have contracts in the relaxation of this rule.

wilzbach: Hmm according to the comment:

> Contracts can only appear without a body when they are virtual interface functions or abstract

and its respective if condition, this should still be forbidden.
I just checked locally and it doesn't throw an error anymore.
The problem is this jump:

```d
if (StructDeclaration sd = parent.isStructDeclaration())
{
    if (funcdecl.isCtorDeclaration())
    {
        goto Ldone;
     }
}
```

ibuclaw: Ah, interesting. I wonder the reason for that.

andralex:
@ibuclaw no mercy

wilzbach:
Ehm you just broke [12901](https://issues.dlang.org/show_bug.cgi?id=12901) without an argument for doing so. The issue is titled like the test case that was removed:

> `in`/`out` contracts on struct constructor must require function body

ibuclaw:
Right, I don't remember seeing anyone actually approve this, and was waiting until I had sorted that test out.

quickfur:
Huh, issue 12901 is really more of a problem of allowing arbitrary assignments inside the contract body.  The original fix was invalid, since it doesn't stop you from writing `this(int n) in { a = n; } do {}`, which is still contract abuse. I'm not sure what the correct fix should be, perhaps require the contracts be pure or something.

PR #7530, opened 2017-12-26
Fix HOST_CSS typo in the Makefile documentation header

wilzbach:


src/posix.mak
```diff
@@ -2,7 +2,7 @@
 # Important variables:
 # --------------------
 #
-# HOST_CSS:				Host C++ compiler to use (g++,clang++)
+# HOST_CXX:				Host C++ compiler to use (g++,clang++)
```
JinShil: While you're at it, please fix the excessive spacing between the variable and the description so it lines up with the others.

JinShil (approved):

src/posix.mak
wilzbach: Those were tabs -> I "retabbed" them with spaces.

JinShil (approved):

PR #6727, opened 2017-04-24
Fix issue 16995 - __traits(getUnitTests) works with separate compilation

atilaneves:
Before this patch, `__traits(getUnitTests)` only worked when compiling all source files at once. This fix enables it to work for separate compilation as well.

PR #7531, opened 2017-12-27
Fix invalid characters in mangling

jacob-carlborg:
Running the DMD test suite with a compiler compiled in debug mode fails due to invalid characters, `-`, exist in the mangled names of some unittest blocks.

This fix centralizes the validation of mangled characters and properly replaces `-` in the mangled names of unittest blocks with `_`.

The issue was introduced in https://github.com/dlang/dmd/pull/6727.

### DIFF
--- a/src/dmd/dmangle.d
+++ b/src/dmd/dmangle.d
@@ -591,12 +591,8 @@ public:
             assert(slice.length);
             foreach (const char c; slice)
             {
-                assert(c == '_' ||
-                       c == '@' ||
-                       c == '?' ||
-                       c == '$' ||
-                       isalnum(c) ||
-                       c & 0x80);
+                assert(c.isValidMangling, "The mangled name '" ~ slice ~ "' " ~
+                    "contains an invalid character: " ~ c);
             }
         }
     }
@@ -1101,6 +1097,34 @@ public:
     }
 }
 
+/// Returns: `true` if the given character is a valid mangled character
+package bool isValidMangling(dchar c)
+{
+    return
+        c >= 'A' && c <= 'Z' ||
+        c >= 'a' && c <= 'z' ||
+        c >= '0' && c <= '9' ||
+        c != 0 && strchr("$%().:?@[]_", c);
+}
+
+// valid mangled characters
+unittest
+{
+    assert('a'.isValidMangling);
+    assert('B'.isValidMangling);
+    assert('2'.isValidMangling);
+    assert('@'.isValidMangling);
+    assert('_'.isValidMangling);
+}
+
+// invalid mangled characters
+unittest
+{
+    assert(!'-'.isValidMangling);
+    assert(!0.isValidMangling);
+    assert(!'/'.isValidMangling);
+    assert(!'\\'.isValidMangling);
+}
 
 /******************************************************************************
  * Returns exact mangled name of function.
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -28,6 +28,7 @@ import dmd.declaration;
 import dmd.denum;
 import dmd.dimport;
 import dmd.dinterpret;
+import dmd.dmangle;
 import dmd.dmodule;
 import dmd.dscope;
 import dmd.dstruct;
@@ -1744,7 +1745,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                     dchar c = p[i];
                     if (c < 0x80)
                     {
-                        if (c >= 'A' && c <= 'Z' || c >= 'a' && c <= 'z' || c >= '0' && c <= '9' || c != 0 && strchr("$%().:?@[]_", c))
+                        if (c.isValidMangling)
                         {
                             ++i;
                             continue;


wilzbach:
CIs are failing with:

```
 ... fail_compilation/ice14424.d    -o- -unittest ()
../generated/linux/release/64/dmd -conf= -m64 -Ifail_compilation -verrors=0 -o- -unittest  -odtest_results/fail_compilation -oftest_results/fail_compilation/ice14424_0.o -c fail_compilation/ice14424.d
fail_compilation/ice14424.d(12): Error: `tuple(__unittest_fail_compilation_imports_a14424.d_3_0)` has no effect


Test failed: 
expected:
----
fail_compilation/ice14424.d(12): Error: `tuple(__unittest_fail_compilation_imports_a14424_d_3_0)` has no effect
----
actual:
----
fail_compilation/ice14424.d(12): Error: `tuple(__unittest_fail_compilation_imports_a14424.d_3_0)` has no effect
----

Makefile:204: recipe for target 'test_results/fail_compilation/ice14424.d.out' failed
```

Also mind the `d-unit` failure:

```
dunit 1.0.14: building configuration "dunit-test-library"...

source/dunit/mockable.d-mixin-156(156,78): Error: semicolon expected, not `.`

source/dunit/mockable.d-mixin-156(156,104): Error: no identifier for declarator `.d_431_0FZ1T7method1MxFiZi`

source/dunit/mockable.d-mixin-156(157,79): Error: semicolon expected, not `.`

source/dunit/mockable.d-mixin-156(157,105): Error: no identifier for declarator `.d_431_0FZ1T7method2MxFiZv`

source/dunit/mockable.d-mixin-156(158,97): Error: semicolon expected, not `.`

source/dunit/mockable.d-mixin-156(158,129): Error: no identifier for declarator `.d_431_0FZ1T7method3MxFNaNbNfiZi`

source/dunit/mockable.d-mixin-156(159,98): Error: semicolon expected, not `.`

source/dunit/mockable.d-mixin-156(159,130): Error: no identifier for declarator `.d_431_0FZ1T7method4MxFNaNbNfiZv`

source/dunit/mockable.d-mixin-156(160,100): Error: semicolon expected, not `.`

source/dunit/mockable.d-mixin-156(160,132): Error: no identifier for declarator `.d_431_0FZ1T7method5MxFNaNbNeiZi`

source/dunit/mockable.d-mixin-156(161,101): Error: semicolon expected, not `.`

source/dunit/mockable.d-mixin-156(161,133): Error: no identifier for declarator `.d_431_0FZ1T7method6MxFNaNbNeiZv`

source/dunit/mockable.d(111,14): Error: template instance dunit.mockable.__unittest_source_dunit_mockable.d_431_0.T.Mockable!(T).Mock!(T) error instantiating

source/dunit/mockable.d(447,23):        instantiated from here: getMock!()

source/dunit/mockable.d-mixin-156(156,78): Error: semicolon expected, not `.`

source/dunit/mockable.d-mixin-156(156,104): Error: no identifier for declarator `.d_464_1FZ1T7method1MxFiZi`

source/dunit/mockable.d-mixin-156(157,79): Error: semicolon expected, not `.`

source/dunit/mockable.d-mixin-156(157,105): Error: no identifier for declarator `.d_464_1FZ1T7method2MxFiZv`

source/dunit/mockable.d-mixin-156(158,97): Error: semicolon expected, not `.`

source/dunit/mockable.d-mixin-156(158,129): Error: no identifier for declarator `.d_464_1FZ1T7method3MxFNaNbNfiZi`

source/dunit/mockable.d-mixin-156(159,98): Error: semicolon expected, not `.`
```

WalterBright:
This is a regression fix, and so should have a bugzilla entry for it.

src/dmd/dmangle.d
```diff
@@ -584,7 +584,8 @@ public:
             assert(slice.length);
             foreach (const char c; slice)
             {
-                assert(c == '_' || c == '@' || c == '?' || c == '$' || isalnum(c) || c & 0x80);
+                assert(c.isValidMangling, "The mangled name '" ~ slice ~ "' " ~
+                    "contains an invalid character: " ~ c);
```
WalterBright: Just leave it as:
```
assert(c.isValidMangling);
```
Anyone who cannot look at the file/line of the assert in the source code is not going to be helped by adding a message. Asserts in the compiler need to be utilitarian, so there is no incentive to leave them out of the release build.

jacob-carlborg:
I looked at one of the failing tests in Phobos. The test inspects the demangle of a symbol. Before this PR the mangled name was `_D3std6traits32__unittest_std_traits_d_7443_193FZ3fooi`, with this PR the mangled name is `_D3std6traits32__unittest_std_traits.d_7443_193FZ3fooi`. Note the dot between `traits` and `d`. The demangler fails to demangle this name with the dot.

I did what @ibuclaw suggested on Slack and took the validation that is performed on `pragma(mangle)` and reused that, which allows a dot, while the code before replaced any dots with an underscore.

rainers:
I would restrain from putting any special character into the mangling as this will byte you back later. For example  `@` has some special meaning on linux (version information IIRC).

So, `.` in the name should be avoided, it also doesn't fit the current mangling grammar.

The validation in dmangle is not strict enough for most D symbols and too strict for symbols in general, as any characters can creep into the mangled string as part of a template alias parameter.

jacob-carlborg:
@rainers so perhaps we shouldn't use the exact same validation everywhere. Because with `pragma(mangle)` one should be able to use all the special symbols, in my opinion.

rainers:
Yes, I support the comment in dymbolsem.d: "Ideally pragma(mangle) can accept a string of any content."
That would mean we cannot make any simple reliable validation, though. Running core.demangle on the mangled name might work, but will cause a bad performance penalty.

jacob-carlborg:
I would expect `core.demangle` to only work on the D mangling, not arbitrary mangled name put in `pragma(mangle)`.

rainers:
Yes, core.demangle doesn't work for the plain symbol set by pragma(mangle), but should be able to handle "embedded" symbols, i.e. as template alias parameters. That means it skips them correctly while decoding the rest.

jacob-carlborg:
So what kind of validation would we like to have in the case of unittest blocks?

rainers:
> So what kind of validation would we like to have in the case of unittest blocks?

Not sure it needs any validation, because I don't think the filename is a good identifier to begin with. It might contain relative or absolute path components which makes it rather unpredictable.

I'd recommend using the module name instead. If you want to replace the dots with an underscore and disambiguate with a module name already containing the underscore, use the mangling scheme for the module name, e.g. `3pkg7modname`.

Long term I think all those internal identifiers should follow the normal naming scheme, e.g. `3pkg7modname13__unittest_L2` or even the line number somehow not part of the identifier, but for eexample as an (template) argument. Typeinfo is another example that decodes badly.

wilzbach:
Rebased and interestingly modifying the test was no longer required.

jacob-carlborg:
> Rebased and interestingly modifying the test was no longer required.

The tests are still not passing. I think Atila's PR https://github.com/dlang/dmd/pull/7761 will fix this as well. But this might be good to have anyway.

wilzbach (approved):
Rebased again. As #7761 was merged, the modification in of `UnitTestDeclaration::createIdentifier` was no longer needed and maybe this finally passes the testsuite.

src/dmd/func.d
```diff
@@ -22,6 +22,7 @@ import dmd.dclass;
 import dmd.declaration;
 import dmd.delegatize;
 import dmd.dinterpret;
+import dmd.dmangle;
```
Geod24: Doesn't seem used anywhere ?

ibuclaw: Indeed. Most likely the original patch had the semantic pass in this module.  Now it's been moved to dsymbolsem.

ibuclaw:
Why is this still open?  Valid code doesn't assemble because of this.

wilzbach:
> Why is this still open? Valid code doesn't assemble because of this.

Because it was failing on auto-tester for a long time and then was having merge conflicts.
I don't see any other reason.

Geod24 (approved):

JinShil (approved):

PR #7532, opened 2017-12-27
[CodeCov] Fix 'No parent commit was found' by allowing coverage offsets

wilzbach:
At CircleCi, the PR is merged into `master` before the testsuite is run.
This allows CodeCov to adjust the resulting coverage diff, s.t. it matches
with the GitHub diff.

This should fix the "Unable to determine changes, no report found" messages.

See also:
- https://docs.codecov.io/v4.3.6/docs/comparing-commits
- https://github.com/codecov/support/issues/363
- https://github.com/dlang/dmd/blob/master/.circleci/run.sh#L74

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -39,4 +39,11 @@ coverage:
     - "test/.*/::src/"
     - "fail_compilation/::src/"
 
+  # At CircleCi, the PR is merged into `master` before the testsuite is run.
+  # This allows CodeCov to adjust the resulting coverage diff, s.t. it matches
+  # with the GitHub diff.
+  # https://github.com/codecov/support/issues/363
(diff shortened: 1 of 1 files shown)


PetarKirov (approved):

ZombineDev:
Hmm, CodeCov report is not showing, any ideas why?

wilzbach:
Hmm, CodeCov reports: "CI failed" 

https://codecov.io/github/dlang/dmd/commit/32f0439daf29eedf6861bcc8bb92e7c900cda895

I thought that we ignored Travis, but maybe there's another flag which we need to set :/

wilzbach:
As I'm in lack of an environment to test this (CodeCov ignores the YAML of the PR) and this won't have any negative effect, let's give this a try.

PR #7533, opened 2017-12-27
cherry picked: Visual D project: add missing files

rainers:
Would be nice to not break building the release using the VS projects


wilzbach (approved):
Yeah let's get the AppVeyor script in ASAP and then let it use VisualD to build?

PR #7534, opened 2017-12-27
fix Issue 15374 - Move genCmain into Compiler struct.

ibuclaw:
Probably a good time to think about what to do with `entrypoint` and `rootHasMain`.  Keep as top-level vars? Or move into `Global`.

### DIFF
--- a/src/dmd/compiler.d
+++ b/src/dmd/compiler.d
@@ -12,10 +12,80 @@
 
 module dmd.compiler;
 
+import dmd.astcodegen;
+import dmd.dmodule;
+import dmd.dscope;
+import dmd.dsymbolsem;
+import dmd.globals;
+import dmd.id;
+import dmd.identifier;
+import dmd.parse;
+import dmd.semantic2;
+import dmd.semantic3;
+import dmd.tokens;
+
+/// DMD-generated module `__entrypoint` where the C main resides
+package __gshared Module entrypoint = null;
+/// Module in which the D main is
+package __gshared Module rootHasMain = null;
+
 /**
  * A data structure that describes a back-end compiler and implements
  * compiler-specific actions.
  */
 struct Compiler
 {
+    /**
+     * Generate C main() in response to seeing D main().
+     *
+     * This function will generate a module called `__entrypoint`,
+     * and set the globals `entrypoint` and `rootHasMain`.
+     *
+     * This used to be in druntime, but contained a reference to _Dmain
+     * which didn't work when druntime was made into a dll and was linked
+     * to a program, such as a C++ program, that didn't have a _Dmain.
+     *
+     * Params:
+     *   sc = Scope which triggered the generation of the C main,
+     *        used to get the module where the D main is.
+     */
+    extern (C++) static void genCmain(Scope* sc)
+    {
+        if (entrypoint)
+            return;
+        /* The D code to be generated is provided as D source code in the form of a string.
+         * Note that Solaris, for unknown reasons, requires both a main() and an _main()
+         */
+        immutable cmaincode =
+        q{
+            extern(C)
+            {
+                int _d_run_main(int argc, char **argv, void* mainFunc);
+                int _Dmain(char[][] args);
+                int main(int argc, char **argv)
+                {
+                    return _d_run_main(argc, argv, &_Dmain);
+                }
+                version (Solaris) int _main(int argc, char** argv) { return main(argc, argv); }
+            }
+        };
+        Identifier id = Id.entrypoint;
+        auto m = new Module("__entrypoint.d", id, 0, 0);
+        scope p = new Parser!ASTCodegen(m, cmaincode, false);
+        p.scanloc = Loc.initial;
+        p.nextToken();
+        m.members = p.parseModule();
+        assert(p.token.value == TOK.endOfFile);
+        assert(!p.errors); // shouldn't have failed to parse it
+        bool v = global.params.verbose;
+        global.params.verbose = false;
+        m.importedFrom = m;
+        m.importAll(null);
+        m.dsymbolSemantic(null);
+        m.semantic2(null);
+        m.semantic3(null);
+        global.params.verbose = v;
+        entrypoint = m;
+        rootHasMain = sc._module;
+    }
 }
--- a/src/dmd/compiler.h
+++ b/src/dmd/compiler.h
@@ -13,6 +13,9 @@
 // This file contains a data structure that describes a back-end compiler
 // and implements compiler-specific actions.
 
+struct Scope;
+
 struct Compiler
 {
+    static void genCmain(Scope *);
 };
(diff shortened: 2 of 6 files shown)


wilzbach:
Regarding:

```
CC="c++" /home/braddr/sandbox/at-client/release-build/install/freebsd/bin32/dmd -lib -of../generated/freebsd/release/32/lexer.a -m32 -J../generated/freebsd/release/32 -L-lstdc++ -version=MARS  -w -de -O -release -inline dmd/console.d dmd/entity.d dmd/errors.d dmd/globals.d dmd/id.d dmd/identifier.d dmd/lexer.d dmd/tokens.d dmd/utf.d dmd/root/array.d dmd/root/ctfloat.d dmd/root/file.d dmd/root/filename.d dmd/root/outbuffer.d dmd/root/port.d dmd/root/rmem.d dmd/root/rootobject.d dmd/root/stringtable.d dmd/root/hash.d
dmd/doc.d(356): Error: file "default_ddoc_theme.ddoc" cannot be found or not in a path specified with -J
posix.mak:445: recipe for target '../generated/freebsd/release/32/lexer.a' failed
```

The other targets do seem to have ` -J$G -J../res` (instead of just `-J$G`), but the error suggests that now `doc.d` is imported.

ibuclaw:
I'm not sure what that has to do with this PR.

It built locally just fine for me.

wilzbach:
> It built locally just fine for me.

Maybe you didn't `clean` your build directory. I can reproduce the failure locally and as said this will fix it:

```diff
diff --git a/src/posix.mak b/src/posix.mak
index a025bb42f..9a42370ab 100644
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -442,7 +442,7 @@ $G/backend.a: $(G_OBJS) $(SRC_MAKE)
 	$(AR) rcs $@ $(G_OBJS)
 
 $G/lexer.a: $(LEXER_SRCS) $(LEXER_ROOT) $(HOST_DMD_PATH) $(SRC_MAKE)
-	CC="$(HOST_CXX)" $(HOST_DMD_RUN) -lib -of$@ $(MODEL_FLAG) -J$G -L-lstdc++ $(DFLAGS) $(LEXER_SRCS) $(LEXER_ROOT)
+	CC="$(HOST_CXX)" $(HOST_DMD_RUN) -lib -of$@ $(MODEL_FLAG) -J$G -J../res -L-lstdc++ $(DFLAGS) $(LEXER_SRCS) $(LEXER_ROOT)
 
 $G/dmd_frontend: $(FRONT_SRCS) $D/gluelayer.d $(ROOT_SRCS) $G/newdelete.o $G/lexer.a $(STRING_IMPORT_FILES) $(HOST_DMD_PATH)
 	CC="$(HOST_CXX)" $(HOST_DMD_RUN) -of$@ $(MODEL_FLAG) -vtls -J$G -J../res -L-lstdc++ $(DFLAGS) $(filter-out $(STRING_IMPORT_FILES) $(HOST_DMD_PATH),$^) -version=NoBackend
```

But I'm still not sure why your changes are leading to the `lexer` importing `doc.d`

wilzbach:
Problem is due to `globals.d` now importing `compiler.d`:

```
> ag "import dmd.compiler"
dmd/mars.d
30:import dmd.compiler;

dmd/globals.d
22:import dmd.compiler;

dmd/dsymbolsem.d
30:import dmd.compiler;
```

ibuclaw:
Rebased.

src/dmd/compiler.d
```diff
@@ -10,7 +10,81 @@
...
 
+import dmd.astcodegen;
+import dmd.dmodule;
+import dmd.dscope;
+import dmd.dsymbolsem;
+import dmd.globals;
+import dmd.id;
+import dmd.identifier;
+import dmd.parse;
+import dmd.semantic;
```
ZombineDev: Import `dmd.semantic2` and `dmd.semantic3` instead.

PetarKirov (commented):
LGTM, just fix the build errors.

ibuclaw:
```
object.Error@src/rt/minfo.d(371): Cyclic dependency between module dmd.lexer and dmd.traits
dmd.lexer* ->
dmd.globals ->
dmd.compiler ->
dmd.dsymbolsem ->
dmd.expressionsem ->
dmd.traits* ->
dmd.dsymbol ->
dmd.lexer*
```
:fearful:

jacob-carlborg:
Perhaps the module constructor in `lexer.d` can be replaced with a CTFE function.

wilzbach:
> Perhaps the module constructor in lexer.d can be replaced with a CTFE function.

Yes, let's kill the module constructor with CTFE :fire: 
-> https://github.com/dlang/dmd/pull/7623

ibuclaw:
Regarding the requirement of `-J../res`, this should not flag up if all imports are nested in functions.

wilzbach:
> Regarding the requirement of -J../res, this should not flag up if all imports are nested in functions.

DMD _only_ does lazy imports for templates. Anyhow we shouldn't need to add `-J ../res` because that means that `parser.a` library depends on `doc.d`.

ibuclaw:
What about the (currently unused) dmd.frontend?  `initDMD()` sure sounds like a tempting place to initialize dmd-specific global fields.

jacob-carlborg:
` initDMD` is intended to be used by when using the DMD library (Dub package).

wilzbach:
> initDMD is intended to be used when using the DMD library (Dub package).

Yep and we want to be able to use Phobos for that to provide nice high-level wrappers.

> Anyhow we shouldn't need to add -J ../res because that means that parser.a library depends on doc.d, which it didn't before this PR.

OTOH considering how much pain `lexer.a` is bringing us - compared to the little gain (people will use the DUB package anyways, I doubt anyone will use `lexer.a` as generated by the Makefile) - you might also want to ignore my statement.

jacob-carlborg:
> OTOH considering how much pain lexer.a is bringing us - compared to the little gain (people will use the DUB package anyways, I doubt anyone will use lexer.a as generated by the Makefile

I thought that was removed.

wilzbach:
> I thought that was removed.

Still there:

https://github.com/dlang/dmd/blob/master/src/posix.mak#L432
https://github.com/dlang/dmd/blob/master/src/posix.mak#L452

CC @RazvanN7

RazvanN7:
If `globals.d` imports `compiler.d` then `compiler.d` should contain only files present in the parser library, otherwise we will end up adding semantic dependencies for the parser library.

ibuclaw:
Indeed, however there are a couple settings that should be set up outside of global module, notably the vendor and version string. Both of which are used by the parser.

Perhaps now there's a frontend module I can dissolve `global._init()` and move the remaining field initialization to more relevant places.

ibuclaw:
Rebased.

jacob-carlborg (approved):

RazvanN7 (approved):

ibuclaw:
Rebased again, reapplying auto-merge.

PR #7535, opened 2017-12-27
[refactor] Replace 'goto Ldefault' with 'goto default'

ibuclaw:
Some artefact of the old C++ code.

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -2759,25 +2759,25 @@ final class Parser(AST) : Lexer
 
                 case TOKconst:
                     if (peek(&token).value == TOKlparen)
-                        goto Ldefault;
+                        goto default;
                     stc = AST.STCconst;
                     goto L2;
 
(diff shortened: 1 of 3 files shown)


wilzbach (approved):
Nice!

src/dmd/parse.d
```diff
@@ -6162,7 +6160,7 @@ final class Parser(AST) : Lexer
                     {
                         t = peek(t);
                         if (t.value != TOKsemicolon && t.value != TOKcomma && t.value != TOKrbracket && t.value != TOKrcurly)
-                            goto Lexpression;
+                            goto default;
```
wilzbach: The CI error is probably due to the `default` being at L6171.

wilzbach (dismissed):

src/dmd/parse.d
ibuclaw: Ah, I had a nagging doubt that this one might be too ambitious.

PR #4136, opened 2014-11-12
fix Issue 519 - Invariant not called from autogenerated class/struct ctor/dtor

WalterBright:
fix https://issues.dlang.org/show_bug.cgi?id=519

Will follow up with change to spec.

PR #7536, opened 2017-12-27
fix Issue 519 - Invariant not called from autogenerated class/struct ctor/dtor

ibuclaw:
Rebase of #4136

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -3878,6 +3878,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
          */
         sd.aggNew = cast(NewDeclaration)sd.search(Loc(), Id.classNew);
         sd.aggDelete = cast(DeleteDeclaration)sd.search(Loc(), Id.classDelete);
+        sd.inv = buildInv(sd, sc2);
 
         // Look for the constructor
         sd.ctor = sd.searchCtor();
@@ -3894,8 +3895,6 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             sd.xcmp = buildXopCmp(sd, sc2);
             sd.xhash = buildXtoHash(sd, sc2);
         }
-
-        sd.inv = buildInv(sd, sc2);
 
         Module.dprogress++;
         sd.semanticRun = PASS.semanticdone;
@@ -4435,6 +4434,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         // Can be in base class
         cldec.aggNew = cast(NewDeclaration)cldec.search(Loc(), Id.classNew);
         cldec.aggDelete = cast(DeleteDeclaration)cldec.search(Loc(), Id.classDelete);
+        cldec.inv = buildInv(cldec, sc2);
 
         // Look for the constructor
         cldec.ctor = cldec.searchCtor();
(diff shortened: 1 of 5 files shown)


wilzbach:
FYI: I am about to change the warning at dlang-bot, so that it isn't displayed for major issues anymore: https://github.com/dlang-bots/dlang-bot/pull/152

src/dmd/e2ir.d
```diff
@@ -5686,6 +5686,18 @@ private elem *toElemStructLit(StructLiteralExp sle, IRState *irs, TOK op, Symbol
         e = el_combine(e, e1);
     }
 
+    if (global.params.useInvariants && sle.sd.inv)
+    {
+        /* If there are no arguments, then this is default initialization and shouldn't call invariant.
+         */
+        if (dim)
```
JinShil: Took me a little bit to work out the reverse logic of that comment and the `if` condition.

src/dmd/dsymbolsem.d
```diff
@@ -6076,6 +6075,29 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             }
         }
 
+        version (none) // Controversial - see Bugzilla 519
```
JinShil: So are we going to do it, or not?

src/dmd/e2ir.d
wilzbach: Yeah. Sth. like "only call invariant for one or more arguments would be easier to read."

test/runnable/testinvariant.d
```diff
@@ -25,6 +25,76 @@ int testinvariant()
...
+    int y;
+}
+
+void test519()
+{
+    //printf("test1\n");
+    {
+        auto s = new C519();
+        assert(C519.x == 0);
+        delete s;
```
wilzbach: `delete` is scheduled for deprecation.

ibuclaw: Even so, for this test to pass, you need to force call the dtor somehow, maybe by using `scope class`.  But that is allegedly deprecated also.  So ¯\_(ツ)_/¯

test/runnable/testinvariant.d
```diff
@@ -25,6 +25,76 @@ int testinvariant()
...
+        assert(C519.x == 1);
+    }
+
+    //printf("test2\n");
+    {
+        auto foo = new S519();
+        printf("lifetime of foo\n");
+        delete foo;
```
JinShil: `destroy(*foo)` should work instead of `delete foo`

test/runnable/testinvariant.d
```diff
@@ -25,6 +25,76 @@ int testinvariant()
...
+        auto foo = S519();
+        printf("lifetime of foo\n");
+    }
+    assert(S519.x == 2);
+
+    //printf("test3\n");
+    {
+        auto foo = new S519c(1);
+        printf("lifetime of foo\n");
+        delete foo;
```
JinShil: `destroy(*foo)` should work instead of `delete foo`

test/runnable/testinvariant.d
JinShil: I suspect `destroy(*foo)` should work instead of `delete foo` here, but this test seems to be failing regardless, so I couldn't verify.

Also, if you want to put the class on the stack, `scope auto s = new C519()` should work.  I'm currently contending that [`scope` in this use case is not being deprecated](https://github.com/dlang/dlang.org/pull/2027).

JinShil (changes requested):

test/runnable/testinvariant.d
ibuclaw: Ah, I see that the original patch was incomplete.  `buildInv()` should be called before `buildDtor()` for classes too, not just structs.

test/runnable/testinvariant.d
ibuclaw: Ah, I had forgotten that there's no pointer-to-struct overload for destroy.

ibuclaw:
Hmm, I think I've managed to turn [4995](https://issues.dlang.org/show_bug.cgi?id=4995) into an error by moving the `buildInv()` call for classes.

src/dmd/dsymbolsem.d
wilzbach: Seems like this decision is the only thing holding this PR of. How about splitting it of into a separate PR?

src/dmd/e2ir.d
```diff
@@ -5678,6 +5678,18 @@ private elem *toElemStructLit(StructLiteralExp sle, IRState *irs, TOK op, Symbol
         e = el_combine(e, e1);
     }
 
+    if (global.params.useInvariants && sle.sd.inv)
```
WalterBright: Inserting the invariant call here seems strange. Everywhere else it is controlled by `addPostInvariant`

RazvanN7:
What is stalling this PR, besides the rebase? @ibuclaw will you take this to the finish line? If not, I volunteer to adopt it.

thewilsonator:
https://github.com/dlang/dmd/pull/7536#pullrequestreview-85749000

RazvanN7:
For anyone who is interested in this patch, I have created a rebase of this PR with the appropriate modifications https://github.com/dlang/dmd/pull/10022

PR #7543, opened 2017-12-28
Re-fix Issue 12901 in/out contracts on struct constructor must require function body

ibuclaw:
Fix as per #7527, but doesn't revert fix in #3664.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -151,6 +151,28 @@ const(char)* getMessage(DeprecatedDeclaration dd)
     return dd.msgstr;
 }
 
+// Returns true if a contract can appear without a function body.
+package bool allowsContractWithoutBody(FuncDeclaration funcdecl)
+{
+    assert(!funcdecl.fbody);
+
+    /* Contracts can only appear without a body when they are virtual
+     * interface functions or abstract.
+     */
+    Dsymbol parent = funcdecl.toParent();
+    InterfaceDeclaration id = parent.isInterfaceDeclaration();
+
+    if (!funcdecl.isAbstract() &&
+        (funcdecl.fensure || funcdecl.frequire) &&
+        !(id && funcdecl.isVirtual()))
+    {
+        auto cd = parent.isClassDeclaration();
+        if (!(cd && cd.isAbstract()))
+            return false;
+    }
+    return true;
+}
+
 private extern(C++) final class DsymbolSemanticVisitor : Visitor
 {
     alias visit = Visitor.visit;
@@ -3149,16 +3171,9 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         // Reflect this.type to f because it could be changed by findVtblIndex
         f = funcdecl.type.toTypeFunction();
 
-        /* Contracts can only appear without a body when they are virtual interface functions or abstract
-         */
-        if (!funcdecl.fbody && !funcdecl.isAbstract() &&
-            (funcdecl.fensure || funcdecl.frequire) &&
-            !(id && funcdecl.isVirtual()))
-        {
-            auto cd = parent.isClassDeclaration();
-            if (!(cd && cd.isAbstract()))
-                funcdecl.error("in and out contracts on non-virtual/non-abstract functions require function body");
-        }
+    Ldone:
+        if (!funcdecl.fbody && !funcdecl.allowsContractWithoutBody())
+            funcdecl.error("in and out contracts can only appear without a body when they are virtual interface functions or abstract");
 
         /* Do not allow template instances to add virtual functions
          * to a class.
@@ -3189,7 +3204,6 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         if (funcdecl.isMain())
             funcdecl.checkDmain();       // Check main() parameters and return type
 
-    Ldone:
         /* Purity and safety can be inferred for some functions by examining
          * the function body.
          */
--- a/src/dmd/semantic3.d
+++ b/src/dmd/semantic3.d
@@ -1074,7 +1074,7 @@ extern(C++) final class Semantic3Visitor : Visitor
                 }
 
                 // If declaration has no body, don't set sbody to prevent incorrect codegen.
-                if (funcdecl.fbody || funcdecl.fdensure || funcdecl.fdrequire)
+                if (funcdecl.fbody || funcdecl.allowsContractWithoutBody())
                     funcdecl.fbody = sbody;
             }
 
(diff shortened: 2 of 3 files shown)


src/dmd/dsymbolsem.d
```diff
@@ -4780,16 +4803,9 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
-            (funcdecl.fensure || funcdecl.frequire) &&
-            !(id && funcdecl.isVirtual()))
-        {
-            auto cd = parent.isClassDeclaration();
-            if (!(cd && cd.isAbstract()))
-                funcdecl.error("in and out contracts on non-virtual/non-abstract functions require function body");
-        }
+    Ldone:
+        if (!funcdecl.fbody && !funcdecl.allowsContractWithoutBody())
+            funcdecl.error("in and out contracts on non-virtual/non-abstract functions require function body");
```
wilzbach:
```
Test failed: 
expected:
----
fail_compilation/fail12901.d(11): Error: constructor fail12901.S.this in and out contracts require function body
----
actual:
----
fail_compilation/fail12901.d(11): Error: constructor fail12901.S.this in and out contracts on non-virtual/non-abstract functions require function body
----
```

How about:

```d
if (parent.isStructDeclaration()))
    funcdecl.error("in and out contracts require function body");
else
   funcdecl.error(""in and out contracts on non-virtual/non-abstract functions require function body");
```

src/dmd/dsymbolsem.d
```diff
@@ -569,6 +569,29 @@ const(char)* getMessage(DeprecatedDeclaration dd)
     return dd.msgstr;
 }
 
+// Returns true if a contact can appear with a function body.
```
wilzbach: s/contact/contract/
s/with/without/

src/dmd/dsymbolsem.d
```diff
@@ -569,6 +569,29 @@ const(char)* getMessage(DeprecatedDeclaration dd)
...
+        InterfaceDeclaration id = parent.isInterfaceDeclaration();
+
+        if (!funcdecl.isAbstract() &&
+            (funcdecl.fensure || funcdecl.frequire) &&
+            !(id && funcdecl.isVirtual()))
+        {
+            auto cd = parent.isClassDeclaration();
+            if (!(cd && cd.isAbstract()))
+                return false;
+        }
```
wilzbach: +1
This is a lot better in terms of readability!

wilzbach (commented):
Thanks for the quick follow-up!

JinShil:
Please heed [this comment](https://github.com/dlang/dmd/pull/7527#issuecomment-354373982) from @quickfur 

> Huh, issue 12901 is really more of a problem of allowing arbitrary assignments inside the contract body. The original fix was invalid, since it doesn't stop you from writing `this(int n) in { a = n; } do {}`, which is still contract abuse. I'm not sure what the correct fix should be, perhaps require the contracts be pure or something.

src/dmd/dsymbolsem.d
JinShil: The function comment header should use DDoc formatting with `Params` and `Returns`.  See https://dlang.org/spec/ddoc.html#standard_srction

src/dmd/dsymbolsem.d
```diff
@@ -569,6 +569,29 @@ const(char)* getMessage(DeprecatedDeclaration dd)
     return dd.msgstr;
 }
 
+// Returns true if a contact can appear with a function body.
+private bool allowsContractWithoutBody(FuncDeclaration funcdecl)
```
WalterBright: the name of the function appears to contradict the `Returns ...` in the function comment.

src/dmd/dsymbolsem.d
WalterBright: Actually, I think the readability is impaired by the extensive use of `!`. Not a big deal, but I'd De Morgan the sense around so it matches the comment.

src/dmd/dsymbolsem.d
```diff
@@ -1570,7 +1593,7 @@ extern(C++) final class Semantic3Visitor : Visitor
                 }
 
                 // If declaration has no body, don't set sbody to prevent incorrect codegen.
-                if (funcdecl.fbody || funcdecl.fdensure || funcdecl.fdrequire)
+                if (funcdecl.fbody || funcdecl.allowsContractWithoutBody())
```
WalterBright: The test of `funcdecl.fbody` here is redundant with what `allowsContractWithoutBody` does.

src/dmd/dsymbolsem.d
WalterBright: The negation in the error message can be rewritten as a positive:
```
Contracts can  only appear without a body when they are virtual interface functions or abstract.
```
:-)

ibuclaw:
> Please heed this comment from @quickfur

I would open a new issue with a better worded title.  The original fix by Kenji does match the issue description.

src/dmd/dsymbolsem.d
ibuclaw: Not if it's private IMO.

src/dmd/dsymbolsem.d
ibuclaw: I've turned it into an assert in the function.

src/dmd/dsymbolsem.d
jacob-carlborg: I think it's appropriate to use the standard sections but not make it a DDoc comment.

quickfur:
The bug title can be (and arguably *should* be) edited.  The complaint really isn't about whether or not to allow bodies in a ctor with contracts, but about being able to initialize aggregate members from inside a contract, which becomes a bug when compiling without contracts.  So the current bug title is misleading, and trying to "fix" the bug according to the title is a wrong solution.  The real problem here is that contracts should not be able to do things like initialize aggregate members.  The part about empty bodies is a red herring.

ibuclaw:
> The bug title can be (and arguably should be) edited.

The bug was closed for nearly 4 years.  No point bringing it back from the dead.  Raise a new issue with a clear description of what should be fixed.

JinShil:
> The bug was closed for nearly 4 years. No point bringing it back from the dead.

Right.  This was broken accidentally when another PR was merged too early.  Let's put things back together first and fix the underlying problem in another PR.  @quickfur, could you please file a separate issue to document the fundamental problem and your proposed solution, so we don't further sidetrack/delay this PR?

ibuclaw:
Rebased. Hopefully no more surprise rebasing needed.

JinShil (approved):

ibuclaw:
New issue raised for other problem. https://issues.dlang.org/show_bug.cgi?id=18143

PR #7550, opened 2017-12-29
error syntax highlighting in func.d

WalterBright:


src/dmd/func.d
```diff
@@ -6,12 +6,12 @@
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
  * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/func.d, _func.d)
+ * Documentation:  https://dlang.org/phobos/dmd_func.html
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/func.d
```
wilzbach: Ddoc's "auto-escaping" rears its head here:

image

http://dtest.dlang.io/artifact/website-d4a3e415e1933b1bf95375258832afc0fa2d38d9-fee3713e1490cd01808e897fa1057418/web/phobos-prerelease/dmd_func.html

BTW what's the motivation for including `Documentation` and `Coverage` on the Ddoc page?
- The user is already on this page
- For Ddox or nightly builds its even a different URL
- Coverage information is not something I expect to be part of the docs?
- If you want something to be done automatic, let us know. It might be more effective.

WalterBright: That's fixed in later versions of dmd.

WalterBright: I find those links very convenient when editing source code.

wilzbach: > That's fixed in later versions of dmd.

Oh yes. Actually there's a fix for this setting in the queue for a while already:

https://github.com/dlang/dmd/pull/7432

> I find those links very convenient when editing source code.

I didn't question this, just the fact that you expose them to the user documentation by adding them to the Ddoc header. What was wrong with using `//` for these two links?

WalterBright: > //

?

wilzbach: You used `// Online documentation: <url>` or `// Coverage: <url>` as comments before - this had the advantage of not being displayed on dlang.org.

WalterBright: Ok. I prefer to have them in one Ddoc block.

JinShil (approved):

PR #6116, opened 2016-09-09
Trivial refactor: Move `@disable` check out of checkDeprecated

mathias-lang-sociomantic:
This separate the check for disabled from the check from deprecated, to (hopefully) make the code clearer.

PR #7552, opened 2017-12-30
Refactor: Move @disable check out of checkDeprecated

JinShil:
Revival of #6116 

This separates the check for `@disabled` from the check from `deprecated`, to (hopefully) make the code clearer.

Courtesy of @mathias-lang-sociomantic

### DIFF
--- a/src/dmd/declaration.d
+++ b/src/dmd/declaration.d
@@ -292,6 +292,51 @@ extern (C++) abstract class Declaration : Dsymbol
         return type.size();
     }
 
+    /**
+     * Issue an error if an attempt to call a disabled method is made
+     *
+     * If the declaration is disabled but inside a disabled function,
+     * returns `true` but do not issue an error message.
+     *
+     * Params:
+     *   loc = Location information of the call
+     *   sc  = Scope in which the call occurs
+     *
+     * Returns:
+     *   `true` if this `Declaration` is `@disable`d, `false` otherwise.
+     */
+    final bool checkDisabled(Loc loc, Scope* sc, bool isAliasedDeclaration = false)
+    {
+        if (storage_class & STCdisable)
+        {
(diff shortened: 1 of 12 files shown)


rainers:
Unfortunately Appveyor doesn't cancel builds automatically for PRs that have been changed after the build has been queued. I hope it is ok to cancel these manually to reduce the queue.
@wilzbach Do you know if that can be configured?

JinShil:
> I hope it is ok to cancel these manually to reduce the queue.

Sorry, I wasn't aware of that.  Cancel away.  I'm still trying to work things out, hence the WIP tag.

wilzbach:
I'm not aware of it, it seems to be an open feature request: http://help.appveyor.com/discussions/suggestions/1180-cancel-build-of-pull-request-updated-via-force-push

Until Travis supported auto-cancelling we implemented this at dlang-bot. However, with Travis supporting this directly, we removed it:
https://github.com/dlang-bots/dlang-bot/pull/75

If this becomes a problem, we could revert this and use the AppVeyor API instead...

rainers:
> If this becomes a problem, we could revert this and use the AppVeyor API instead...

Thanks for the info. If Appveyor doesn't come up with a solution, resurrecting that bot feature might be interesting. There are quite some projects competing for the same resources ATM.

wilzbach:
> There are quite some projects competing for the same resources ATM.

We could also use a different account for each project. Might be the quicker solution and helpful anyway...

rainers:
> We could also use a different account for each project. Might be the quicker solution and helpful anyway...

Not sure if that is ok with their terms of use. I've started a test lld build with my account, though.

JinShil:
I'm still trying to work out a few things with this. Closing for now.

JinShil:
Another revival attempt at #7583

PR #7553, opened 2017-12-30
fix Issue 18143 - in/out contracts should be implicitly 'this' const

ibuclaw:
As noted by @quickfur.

Actual error message should probably be ratified.  I just borrowed the same format as the error above for modifying other parameters in contracts.

### DIFF
--- a/src/dmd/semantic3.d
+++ b/src/dmd/semantic3.d
@@ -885,7 +885,6 @@ private extern(C++) final class Semantic3Visitor : Visitor
                 sc2.flags = (sc2.flags & ~SCOPEcontract) | SCOPErequire;
 
                 // BUG: need to error if accessing out parameters
-                // BUG: need to treat parameters as const
                 // BUG: need to disallow returns and throws
                 // BUG: verify that all in and ref parameters are read
                 freq = freq.statementSemantic(sc2);
@@ -908,7 +907,6 @@ private extern(C++) final class Semantic3Visitor : Visitor
                 sc2 = scout; //push
                 sc2.flags = (sc2.flags & ~SCOPEcontract) | SCOPEensure;
 
-                // BUG: need to treat parameters as const
                 // BUG: need to disallow returns and throws
 
                 if (funcdecl.fensure && f.next.ty != Tvoid)
--- a/src/dmd/declaration.d
+++ b/src/dmd/declaration.d
@@ -315,6 +315,20 @@ extern (C++) abstract class Declaration : Dsymbol
             }
         }
 
+        if (e1 && e1.op == TOKthis && isField())
+        {
+            VarDeclaration vthis = (cast(ThisExp)e1).var;
+            for (Scope* scx = sc; scx; scx = scx.enclosing)
+            {
+                if (scx.func == vthis.parent && (scx.flags & SCOPEcontract))
+                {
+                    if (!flag)
+                        error(loc, "cannot modify parameter 'this' in contract");
+                    return 2; // do not report type related errors
+                }
+            }
+        }
+
         if (v && (isCtorinit() || isField()))
         {
             // It's only modifiable if inside the right constructor
--- a/test/fail_compilation/fail18143.d
+++ b/test/fail_compilation/fail18143.d
@@ -1,0 +1,43 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail18143.d(20): Error: variable fail18143.S.a cannot modify parameter 'this' in contract
+fail_compilation/fail18143.d(21): Error: variable fail18143.S.a cannot modify parameter 'this' in contract
+fail_compilation/fail18143.d(25): Error: variable fail18143.S.a cannot modify parameter 'this' in contract
+fail_compilation/fail18143.d(26): Error: variable fail18143.S.a cannot modify parameter 'this' in contract
+fail_compilation/fail18143.d(35): Error: variable fail18143.C.a cannot modify parameter 'this' in contract
+fail_compilation/fail18143.d(36): Error: variable fail18143.C.a cannot modify parameter 'this' in contract
+fail_compilation/fail18143.d(40): Error: variable fail18143.C.a cannot modify parameter 'this' in contract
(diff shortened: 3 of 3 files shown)


src/dmd/semantic3.d
```diff
@@ -875,7 +875,6 @@ extern(C++) final class Semantic3Visitor : Visitor
                 sc2.flags = (sc2.flags & ~SCOPEcontract) | SCOPErequire;
 
                 // BUG: need to error if accessing out parameters
-                // BUG: need to treat parameters as const
```
quickfur: Hmm. This PR seems to only disable writing to `this`, but does it also prohibit modifying function parameters?

iain-buclaw-sociomantic: @quickfur that was implemented in #1569.  Look at the _whole_ of checkModify.

quickfur: Ah, ok. Nevermind then. :-)   I was just confused because deleting this comment in this PR makes it seem like this PR is fixing the entire issue. I suppose making `this` const is the last remaining issue as far as this comment is concerned?

quickfur:
LGTM.

JinShil:
Looks like this [may have found a bug](https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fdmd/detail/PR-7553/4/pipeline) in [Sociomantic's Ocean](https://github.com/sociomantic-tsunami/oceanf)

cc @leandro-lucarella-sociomantic @mathias-lang-sociomantic

quickfur:
Very nice! I love compiler changes that discover bugs in existing code!

ibuclaw:
It certainly looks suspicious.
https://github.com/sociomantic-tsunami/ocean/blob/84612d09b3ede91ef27333be72b78836997e58e3/src/ocean/net/http/HttpResponse.d#L466

wilzbach:
> Looks like this may have found a bug in Sociomantic's Ocean

Thanks to the work of @iain-buclaw-sociomantic and @mihails-strasuns-sociomantic Ocean was fixed and upgraded:

https://github.com/sociomantic-tsunami/ocean/pull/427
https://github.com/dlang/ci/pull/108

-> Jenkins should be passing now :)

src/dmd/semantic3.d
```diff
@@ -898,7 +897,6 @@ extern(C++) final class Semantic3Visitor : Visitor
                 sc2 = scout; //push
                 sc2.flags = (sc2.flags & ~SCOPEcontract) | SCOPEensure;
 
-                // BUG: need to treat parameters as const
                 // BUG: need to disallow returns and throws
```
wilzbach: We shouldn't document bugs in the source code like this, they should be in Bugzilla:

https://issues.dlang.org/show_bug.cgi?id=18195

wilzbach: Also `return` is already forbidden: https://run.dlang.io/is/A0JWgT

src/dmd/semantic3.d
```diff
@@ -875,7 +875,6 @@ extern(C++) final class Semantic3Visitor : Visitor
                 sc2.flags = (sc2.flags & ~SCOPEcontract) | SCOPErequire;
 
                 // BUG: need to error if accessing out parameters
```
wilzbach: Isn't this already forbidden?

https://run.dlang.io/is/NacqYo

wilzbach (approved):
We should really stop documenting bugs in the source code where they are just forgotten or confuse people because it has been fixed. Should we add the fixed "BUGS" as test cases to the testsuite?

src/dmd/semantic3.d
iain-buclaw-sociomantic: I didn't verify any of the comments, so no idea what the state is.

src/dmd/semantic3.d
iain-buclaw-sociomantic: But is that accessing the `out` parameter, or `out` parameters?

wilzbach:
(rebased to hopefully fix Jenkins)

JinShil (approved):

quickfur:
Excellent!

src/dmd/semantic3.d
ibuclaw: Actually, I've worked out what this means:
```
static int cache;

int foo(out int bar)
in
{
    cache = bar;  // This should be an error.
}
```

ibuclaw: https://issues.dlang.org/show_bug.cgi?id=18203

src/dmd/semantic3.d
```diff
@@ -885,7 +885,6 @@ private extern(C++) final class Semantic3Visitor : Visitor
...
-                // BUG: need to treat parameters as const
                 // BUG: need to disallow returns and throws
                 // BUG: verify that all in and ref parameters are read
```
ibuclaw: This comment as seen today appeared in D1.011.

Before it used to be:
```
BUG: verify that all in and inout parameters are read
```
Which has been here since the dawn of ~time~ svn history.

ibuclaw: It seems to allude to this:
```
int foo(in int a, ref int b)
in
{
    // Should emit error: 'a' is not read in contract.
    // Should emit error: 'b' is not read in contract.
}
do
{
    return a + b;
}
```
But I don't think that is a bug at all, and so this comment should be removed.  @WalterBright do I understand the context correctly?

quickfur: Why should an error be emitted if a parameter isn't read in the in-contract? How else would you express that any possible value is permitted for that parameter?

mathias-lang-sociomantic: Some languages enforce that some parameter are used, e.g. C# forces you to assign `out` parameters. One way to express this in an `in` contract would be to do `cast(void)a;` for example. But that's the kind of enforcement which IMO goes against D's approach, as many times you'll end up just trying to satisfy the compiler.

ibuclaw: I'm saying I don't think it should be considered as a bug.  Also the comments doesn't say anything about all parameters.  Just `in` and `ref` parameters.

JinShil: > verify that all in and ref parameters are read

It may also be implying that all input parameters should actually be used in the implementation.

PR #7554, opened 2017-12-30
Fix Issue 15613, 11529: Show parameter mismatch and rvalue/lvalue ref message

ntrel:
When there's only one overload to consider:
* Add a supplemental message showing the first argument that failed to match the corresponding function parameter.
* Alter the messsage to show if an rvalue is being passed to a ref/out parameter.
* When necessary for disambiguation, show the qualified type of the parameter as well.

### DIFF
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -2329,8 +2329,10 @@ extern (C++) final class TypeDeduced : Type
  *      tiargs          initial list of template arguments
  *      tthis           if !NULL, the 'this' pointer argument
  *      fargs           arguments to function
+ *      failedIndex     address to store argument index of first type mismatch
  */
-void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiargs, Type tthis, Expressions* fargs)
+void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiargs,
+    Type tthis, Expressions* fargs, size_t* failedIndex = null)
 {
     version (none)
     {
@@ -2416,7 +2418,7 @@ void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiar
             else
                 return 0;   // MATCH.nomatch
         }
-        MATCH mfa = tf.callMatch(tthis_fd, fargs);
+        MATCH mfa = tf.callMatch(tthis_fd, fargs, 0, failedIndex);
         //printf("test1: mfa = %d\n", mfa);
         if (mfa > MATCH.nomatch)
         {
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -2330,9 +2330,10 @@ extern (C++) final class TypeDeduced : Type
  *      tthis           if !NULL, the 'this' pointer argument
  *      fargs           arguments to function
  *      failedIndex     address to store argument index of first type mismatch
+ *      pMessage        address to store error message, or null
  */
 void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiargs,
-    Type tthis, Expressions* fargs, size_t* failedIndex = null)
+    Type tthis, Expressions* fargs, const(char)** pMessage = null)
 {
     version (none)
     {
@@ -2418,7 +2419,7 @@ void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiar
             else
                 return 0;   // MATCH.nomatch
         }
-        MATCH mfa = tf.callMatch(tthis_fd, fargs, 0, failedIndex);
+        MATCH mfa = tf.callMatch(tthis_fd, fargs, 0, pMessage);
         //printf("test1: mfa = %d\n", mfa);
         if (mfa > MATCH.nomatch)
         {
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -2321,16 +2321,15 @@ extern (C++) final class TypeDeduced : Type
 /*************************************************
  * Given function arguments, figure out which template function
  * to expand, and return matching result.
- * Input:
- *      m               matching result
- *      dstart          the root of overloaded function templates
- *      loc             instantiation location
- *      sc              instantiation scope
- *      tiargs          initial list of template arguments
- *      tthis           if !NULL, the 'this' pointer argument
- *      fargs           arguments to function
- *      failedIndex     address to store argument index of first type mismatch
- *      pMessage        address to store error message, or null
+ * Params:
+ *      m           = matching result
+ *      dstart      = the root of overloaded function templates
+ *      loc         = instantiation location
+ *      sc          = instantiation scope
+ *      tiargs      = initial list of template arguments
+ *      tthis       = if !NULL, the 'this' pointer argument
+ *      fargs       = arguments to function
+ *      pMessage    = address to store error message, or null
  */
 void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiargs,
     Type tthis, Expressions* fargs, const(char)** pMessage = null)
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -3430,7 +3430,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 return setError();
             }
 
-            if (!tf.callMatch(null, exp.arguments))
+            size_t failIndex;
+            if (!tf.callMatch(null, exp.arguments, 0, &failIndex))
             {
                 OutBuffer buf;
                 buf.writeByte('(');
@@ -3440,7 +3441,9 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                     tthis.modToBuffer(&buf);
 
                 //printf("tf = %s, args = %s\n", tf.deco, (*arguments)[0].type.deco);
-                .error(exp.loc, "%s `%s%s` is not callable using argument types `%s`", p, exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
+                .error(exp.loc, "%s `%s%s` is not callable using argument types `%s`",
+                    p, exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
+                showArgMismatch(exp.loc, exp.arguments, tf, failIndex);
                 return setError();
             }
             // Purity and safety check should run after testing arguments matching
@@ -3500,7 +3503,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             {
                 exp.f = exp.f.toAliasFunc();
                 TypeFunction tf = cast(TypeFunction)exp.f.type;
-                if (!tf.callMatch(null, exp.arguments))
+                size_t failIndex;
+                if (!tf.callMatch(null, exp.arguments, 0, &failIndex))
                 {
                     OutBuffer buf;
                     buf.writeByte('(');
@@ -3508,8 +3512,9 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                     buf.writeByte(')');
 
                     //printf("tf = %s, args = %s\n", tf.deco, (*arguments)[0].type.deco);
-                    .error(exp.loc, "`%s%s` is not callable using argument types `%s`", exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
-
+                    .error(exp.loc, "`%s%s` is not callable using argument types `%s`",
+                        exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
+                    showArgMismatch(exp.loc, exp.arguments, tf, failIndex);
                     exp.f = null;
                 }
             }
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -3430,8 +3430,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 return setError();
             }
 
-            size_t failIndex;
-            if (!tf.callMatch(null, exp.arguments, 0, &failIndex))
+            const(char)* failMessage;
+            if (!tf.callMatch(null, exp.arguments, 0, &failMessage))
             {
                 OutBuffer buf;
                 buf.writeByte('(');
@@ -3443,7 +3443,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 //printf("tf = %s, args = %s\n", tf.deco, (*arguments)[0].type.deco);
                 .error(exp.loc, "%s `%s%s` is not callable using argument types `%s`",
                     p, exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
-                showArgMismatch(exp.loc, exp.arguments, tf, failIndex);
+                if (failMessage)
+                    errorSupplemental(exp.loc, failMessage);
                 return setError();
             }
             // Purity and safety check should run after testing arguments matching
@@ -3503,8 +3504,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             {
                 exp.f = exp.f.toAliasFunc();
                 TypeFunction tf = cast(TypeFunction)exp.f.type;
-                size_t failIndex;
-                if (!tf.callMatch(null, exp.arguments, 0, &failIndex))
+                const(char)* failMessage;
+                if (!tf.callMatch(null, exp.arguments, 0, &failMessage))
                 {
                     OutBuffer buf;
                     buf.writeByte('(');
@@ -3514,7 +3515,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                     //printf("tf = %s, args = %s\n", tf.deco, (*arguments)[0].type.deco);
                     .error(exp.loc, "`%s%s` is not callable using argument types `%s`",
                         exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
-                    showArgMismatch(exp.loc, exp.arguments, tf, failIndex);
+                    if (failMessage)
+                        errorSupplemental(exp.loc, failMessage);
                     exp.f = null;
                 }
             }
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -3498,7 +3498,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             assert(exp.f);
             tiargs = null;
 
-            if (ve.hasOverloads)
+            if (exp.f.overnext)
                 exp.f = resolveFuncCall(exp.loc, sc, exp.f, tiargs, null, exp.arguments, 2);
             else
             {
@@ -3513,8 +3513,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                     buf.writeByte(')');
 
                     //printf("tf = %s, args = %s\n", tf.deco, (*arguments)[0].type.deco);
-                    .error(exp.loc, "`%s%s` is not callable using argument types `%s`",
-                        exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
+                    .error(exp.loc, "%s `%s%s` is not callable using argument types `%s`",
+                        exp.f.kind(), exp.f.toPrettyChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
                     if (failMessage)
                         errorSupplemental(exp.loc, failMessage);
                     exp.f = null;
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -2481,8 +2481,8 @@ extern (C++) FuncDeclaration resolveFuncCall(Loc loc, Scope* sc, Dsymbol s,
     Match m;
     m.last = MATCH.nomatch;
 
-    size_t failIndex;
-    functionResolve(&m, s, loc, sc, tiargs, tthis, fargs, &failIndex);
+    const(char)* failMessage;
+    functionResolve(&m, s, loc, sc, tiargs, tthis, fargs, &failMessage);
 
     if (m.last > MATCH.nomatch && m.lastf)
     {
@@ -2600,7 +2600,8 @@ extern (C++) FuncDeclaration resolveFuncCall(Loc loc, Scope* sc, Dsymbol s,
                     .error(loc, "%s `%s%s%s` is not callable using argument types `%s`",
                         fd.kind(), fd.toPrettyChars(), parametersTypeToChars(tf.parameters, tf.varargs),
                         tf.modToChars(), fargsBuf.peekString());
-                    showArgMismatch(loc, fargs, tf, failIndex);
+                    if (failMessage)
+                        errorSupplemental(loc, failMessage);
                 }
             }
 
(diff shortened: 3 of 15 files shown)


nordlow:
Better diagnostics is exactly what D needs right now! Great work!

src/dmd/func.d
```diff
@@ -3618,3 +3620,21 @@ extern (C++) final class DeleteDeclaration : FuncDeclaration
         v.visit(this);
     }
 }
+
+void showArgMismatch(Loc loc, Expressions* fargs, TypeFunction tf, size_t failIndex)
```
WalterBright: Please add Ddoc style function comment.

src/dmd/func.d
```diff
@@ -3618,3 +3620,21 @@ extern (C++) final class DeleteDeclaration : FuncDeclaration
...
+    {
+        auto arg = (*fargs)[failIndex];
+        auto par = (*tf.parameters)[failIndex];
+        const rv = (!arg.isLvalue() && par.storageClass & (STCref | STCout)) ? "rvalue " : "";
+        const pc = parameterToChars(par, tf.varargs);
+        const ts = toAutoQualChars(arg.type, par.type);
+        auto msg = "cannot pass %sargument `%s` of type `%s` to parameter `%s`";
+        // don't print parameter type if it's already in the parameter string
+        if (strcmp(par.type.toChars(), ts[1]) != 0)
+            msg ~= " of type `%s`";
```
WalterBright: This isn't guaranteed to produce a 0 terminated array.

WalterBright: Just use two different `msg` strings.

WalterBright: i.e.:
```
msg = strcmp() ? "cannot pass ..." : "cannot pass ... of type ...";
```

src/dmd/hdrgen.d
```diff
@@ -3433,3 +3433,17 @@ extern (C++) const(char)* parametersTypeToChars(Parameters* parameters, int vara
     v.parametersToBuffer(parameters, varargs);
     return buf.extractString();
 }
+
+extern (C++) const(char)* parameterToChars(Parameter parameter, int varargs)
```
WalterBright: Add Ddoc function comment.

src/dmd/mtype.d
```diff
@@ -5935,6 +5936,7 @@ extern (C++) final class TypeFunction : TypeNext
             }
         }
 
+        auto u = size_t.max; // used for failedIndex if no match
```
JinShil: I would prefer having a `bool failed` and a `size_t failedIndex` and use the `failed` to determine whether or not `failedIndex` holds a valid value.

src/dmd/hdrgen.d
```diff
@@ -3433,3 +3437,21 @@ extern (C++) const(char)* parametersTypeToChars(Parameters* parameters, int vara
     v.parametersToBuffer(parameters, varargs);
     return buf.extractString();
 }
+
+/** Params:
+ *  parameter = parameter to print.
+ *  varargs = Kind of varargs, see TypeFunction.varargs.
+ * Returns: NT string representing parameter. */
+extern (C++) const(char)* parameterToChars(Parameter parameter, int varargs)
```
JinShil: DDoc looks good, but it also needs a summary

src/dmd/hdrgen.d
```diff
@@ -3425,6 +3425,10 @@ extern (C++) void arrayObjectsToBuffer(OutBuffer* buf, Objects* objects)
     }
 }
 
+/** Params:
+ *  parameters = parameters to print, such as TypeFunction.parameters.
+ *  varargs = Kind of varargs, see TypeFunction.varargs.
+ * Returns: NT string representing parameters. */
 extern (C++) const(char)* parametersTypeToChars(Parameters* parameters, int varargs)
```
JinShil: DDoc looks good, but it also needs a summary

JinShil (changes requested):

src/dmd/hdrgen.d
JinShil: Please don't abbreviate null-terminated.

src/dmd/func.d
ntrel: Fixed by adding `\0` to 2nd string, hope that's OK?

test/fail_compilation/bug9631.d
```diff
@@ -58,3 +58,54 @@ void test3()
     S[1] sa = cast(S[1])ta;
     auto t2 = cast(tem!().S[])sa;
 }
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/bug9631.d(79): Error: function bug9631.arg.f `(int i, S s)` is not callable using argument types `(int, S)`
+fail_compilation/bug9631.d(79):        cannot pass argument `y` of type `bug9631.tem!().S` to parameter `S s` of type `bug9631.S`
```
JinShil: > function bug9631.arg.f \`(int i, S s)\` is not callable using argument types \`(int, S)\`

1. Tick marks should surround the entire function declaration (e.g \`bug9631.arg.f (int i, S s)\`).  That may be a difficult thing to do and I'd be willing to pass on the tick marks for now
2.  The function accepts `(int, S)` but the error message says it's not callable using `(int, S)`.  Does this need fully qualified names or something?

JinShil: > cannot pass argument `y` of type `bug9631.tem!().S` to parameter `S s` of type `bug9631.S`

I suspect this line is supposed the clarify the first, but it's awfully confusing.  Can't we have just one line with fully qualified type names?

test/fail_compilation/bug9631.d
```diff
@@ -58,3 +58,54 @@ void test3()
...
+fail_compilation/bug9631.d(80): Error: function literal `__lambda2(S s)` is not callable using argument types `(S)`
+fail_compilation/bug9631.d(80):        cannot pass argument `x` of type `bug9631.S` to parameter `S s` of type `bug9631.tem!().S`
+fail_compilation/bug9631.d(86): Error: constructor bug9631.arg.A.this `(S _param_0)` is not callable using argument types `(S)`
+fail_compilation/bug9631.d(86):        cannot pass argument `S(0)` of type `bug9631.tem!().S` to parameter `S _param_0` of type `bug9631.S`
```
JinShil: > cannot pass argument \`S(0)\`

`S(0)` doesn't look like an argument; it looks like a function invocation

> to parameter `S _param_0` of type `bug9631.S`

I think this should read "... to parameter \`bug9631.S _param_0\` without redundantly specifying one unqualified type anem and one qualified type name

JinShil (changes requested):

test/fail_compilation/diag8101b.d
```diff
@@ -1,14 +1,15 @@
...
-fail_compilation/diag8101b.d(28): Error: function diag8101b.S.bar `(int _param_0)` is not callable using argument types `(double)`
-fail_compilation/diag8101b.d(31): Error: none of the overloads of `foo` are callable using a const object, candidates are:
-fail_compilation/diag8101b.d(17):        `diag8101b.S.foo(int _param_0)`
-fail_compilation/diag8101b.d(18):        `diag8101b.S.foo(int _param_0, int _param_1)`
-fail_compilation/diag8101b.d(33): Error: mutable method `diag8101b.S.bar` is not callable using a const object
+fail_compilation/diag8101b.d(27): Error: none of the overloads of `foo` are callable using argument types `(double)`, candidates are:
+fail_compilation/diag8101b.d(18):        `diag8101b.S.foo(int _param_0)`
+fail_compilation/diag8101b.d(19):        `diag8101b.S.foo(int _param_0, int _param_1)`
+fail_compilation/diag8101b.d(29): Error: function diag8101b.S.bar `(int _param_0)` is not callable using argument types `(double)`
+fail_compilation/diag8101b.d(29):        cannot pass argument `1.00000` of type `double` to parameter `int _param_0`
```
JinShil: Can we shoot for something like:

> function `diag8101b.S.bar (int _param_0)` is not callable using argument `1.00000` of type `double` to parameter `int _param_0`

test/fail_compilation/diag8101b.d
```diff
@@ -1,14 +1,15 @@
...
 ---
-fail_compilation/diag8101b.d(26): Error: none of the overloads of `foo` are callable using argument types `(double)`, candidates are:
-fail_compilation/diag8101b.d(17):        `diag8101b.S.foo(int _param_0)`
-fail_compilation/diag8101b.d(18):        `diag8101b.S.foo(int _param_0, int _param_1)`
-fail_compilation/diag8101b.d(28): Error: function diag8101b.S.bar `(int _param_0)` is not callable using argument types `(double)`
-fail_compilation/diag8101b.d(31): Error: none of the overloads of `foo` are callable using a const object, candidates are:
-fail_compilation/diag8101b.d(17):        `diag8101b.S.foo(int _param_0)`
-fail_compilation/diag8101b.d(18):        `diag8101b.S.foo(int _param_0, int _param_1)`
-fail_compilation/diag8101b.d(33): Error: mutable method `diag8101b.S.bar` is not callable using a const object
+fail_compilation/diag8101b.d(27): Error: none of the overloads of `foo` are callable using argument types `(double)`, candidates are:
```
JinShil: The term "argument types" seems odd here.  What about...

> none of the overloads of `foo` are callable parameter list `(double)`, candidates are:

This also removes the plural "types" with the singular parameter type oddity.

test/fail_compilation/bug9631.d
ntrel: > bug9631.arg.f `(int i, S s)`

This was introduced by Walter before this PR. I tried to fix it but can't work out why exactly it happens.

test/fail_compilation/diag8101b.d
ntrel: Well, it's talking about arguments so "argument list" would work. But this is not part of the PR, I expect *argument types* is used elsewhere too.

test/fail_compilation/diag8101b.d
ntrel: I think it's useful to see the full parameter list, so we know which function the compiler means. If we combine both messages I think there's too much information in one sentence.

test/fail_compilation/bug9631.d
ntrel: Done.

test/fail_compilation/diag8101b.d
ntrel: I think it's useful to see all the *argument types* too.

ntrel:
BTW, I also have code to do the same for deduced template function calls (if there are no overloads), but I'll wait until this is merged to submit.

src/dmd/mtype.d
ntrel: We can just use check the return value against `MATCH.nomatch` to determine if `failedIndex` is valid.

src/dmd/hdrgen.d
```diff
@@ -3433,3 +3438,24 @@ extern (C++) const(char)* parametersTypeToChars(Parameters* parameters, int vara
...
+{
+    OutBuffer buf;
+    HdrGenState hgs;
+    hgs.fullQual = fullQual;
+    scope PrettyPrintVisitor v = new PrettyPrintVisitor(&buf, &hgs);
+
+    parameter.accept(v);
+    if (varargs)
+    {
+        buf.writestring("...");
```
WalterBright: need test case to cover this new line

src/dmd/hdrgen.d
```diff
@@ -3433,3 +3438,24 @@ extern (C++) const(char)* parametersTypeToChars(Parameters* parameters, int vara
     v.parametersToBuffer(parameters, varargs);
     return buf.extractString();
 }
+
+/** Pretty print function parameters.
+ * Params:
+ *  parameter = parameter to print.
+ *  varargs = kind of varargs, see TypeFunction.varargs.
+ *  fullQual = whether to fully qualify types.
+ * Returns: Null-terminated string representing parameters. */
```
WalterBright: Nit: Please conform to the existing way Ddoc comments are formatted in the front end. To wit:
```
/****
 * blah blah
 */
```
not:
/** blah blah
   * blah blah */
```

File: src/dmd/mtype.d [hunk 13]
```diff
@@ -9224,7 +9228,7 @@ const(char*)[2] toAutoQualChars(Type t1, Type t2)
 {
     auto s1 = t1.toChars();
     auto s2 = t2.toChars();
-    if (strcmp(s1, s2) == 0)
+    if (!t1.equals(t2) && strcmp(s1, s2) == 0)
```
WalterBright: when would the type strings be the same but they are different types?

src/dmd/hdrgen.d
```diff
@@ -3433,3 +3438,24 @@ extern (C++) const(char)* parametersTypeToChars(Parameters* parameters, int vara
     v.parametersToBuffer(parameters, varargs);
     return buf.extractString();
 }
+
+/** Pretty print function parameters.
```
WalterBright: This has the same description and return value as the previous function - what does it do different?

src/dmd/expressionsem.d
```diff
@@ -3518,16 +3521,18 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
                     buf.writeByte('(');
                     argExpTypesToCBuffer(&buf, exp.arguments);
                     buf.writeByte(')');
 
                     //printf("tf = %s, args = %s\n", tf.deco, (*arguments)[0].type.deco);
-                    .error(exp.loc, "`%s%s` is not callable using argument types `%s`", exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
-
+                    .error(exp.loc, "`%s%s` is not callable using argument types `%s`",
+                        exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
+                    showArgMismatch(exp.loc, exp.arguments, tf, failIndex);
```
WalterBright: It'd be nice to have test coverage for the new code.

src/dmd/dtemplate.d
```diff
@@ -2325,8 +2325,10 @@ extern (C++) final class TypeDeduced : Type
  *      tiargs          initial list of template arguments
  *      tthis           if !NULL, the 'this' pointer argument
  *      fargs           arguments to function
+ *      failedIndex     address to store argument index of first type mismatch
```
WalterBright: note that `failedIndex` can be `null`

src/dmd/dtemplate.d
```diff
@@ -2325,8 +2325,10 @@ extern (C++) final class TypeDeduced : Type
...
  */
-void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiargs, Type tthis, Expressions* fargs)
+void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiargs,
+    Type tthis, Expressions* fargs, size_t* failedIndex = null)
```
WalterBright: Please avoid having default arguments. Explicit is better.

src/dmd/expressionsem.d
```diff
@@ -3448,7 +3448,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 return setError();
             }
 
-            if (!tf.callMatch(null, exp.arguments))
+            size_t failIndex;
```
WalterBright: It's called `failIndex` here, but `failedIndex` further up. Please be consistent. Also, above the `failedIndex` is a pointer and here it is a value. Either name the one above `pFailedIndex' or make it a ref type. Probably the latter is better.

test/fail_compilation/bug9631.d
```diff
@@ -58,3 +58,54 @@ void test3()
     S[1] sa = cast(S[1])ta;
     auto t2 = cast(tem!().S[])sa;
 }
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/bug9631.d(79): Error: function `bug9631.arg.f(int i, S s)` is not callable using argument types `(int, S)`
+fail_compilation/bug9631.d(79):        cannot pass argument `y` of type `bug9631.tem!().S` to parameter `bug9631.S s`
```
WalterBright: Note that line 66 calls it `bug9631.S` but line 65 calls it just `S`. Perhaps it would be better to just call it `S` on line 66, too.

src/dmd/mtype.d
ntrel: When the strings aren't the full qualified type, e.g. `S` vs `tem!().S`.

src/dmd/hdrgen.d
ntrel: Added test, turns out it was wrong for variadic parameters :-/

src/dmd/hdrgen.d
ntrel: Done for this module.

src/dmd/hdrgen.d
ntrel: Typo, should be *parameter*. Fixed.

src/dmd/expressionsem.d
ntrel: BTW, what does TOKvar mean? I'm trying to hit this case.

src/dmd/dtemplate.d
ntrel: It is used elsewhere, e.g. `Parameter.getNth(Parameters* parameters, size_t nth, size_t* pn = null)`. It avoids having to touch existing lines where we want `null`, maybe avoiding merge conflicts.

src/dmd/expressionsem.d
ntrel: If it's a `ref` parameter, what do I pass when I don't care about `failedIndex`?

test/fail_compilation/bug9631.d
ntrel: The point is to disambiguate between S and tem!().S. You're right that the module name isn't needed, but it makes it clear that the compiler has fully qualified the type. In the first line, there's no qualification so that it's easier to read the function parameters. Given this, if I remove the module name, it might not be so clear.

ntrel:
@WalterBright Thanks for reviewing. I think this PR might be better if `TypeFunction.callMatch` handles the logic of determining the error messages, setting a new message parameter. The caller then prints that message. There are some subtleties that we would have to duplicate the logic of in `showArgMismatch` here otherwise.

src/dmd/mtype.d
WalterBright: Please add comment to that effect.

src/dmd/expressionsem.d
WalterBright: `TOKvar` distinguishes a `VarExp` from other Expression types.

src/dmd/expressionsem.d
WalterBright: Pass a dummy variable.

test/fail_compilation/bug9631.d
WalterBright: ok

src/dmd/mtype.d
ntrel: done.

src/dmd/expressionsem.d
ntrel: I don't know how to test for the case when TOKvar is a function. Function pointers are handled differently.

src/dmd/dtemplate.d
ntrel: I've noted this for `pMessage` (which replaced `failedIndex`).

src/dmd/expressionsem.d
ntrel: Then we get memory allocated for the unwanted error message whenever an overload fails to match, even though one overload does match, so no error message is needed.

test/fail_compilation/bug15613.d
```diff
@@ -0,0 +1,18 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/bug15613.d(16): Error: function `bug15613.f(int...)` is not callable using argument types `(typeof(null))`
+fail_compilation/bug15613.d(16):        cannot pass argument `null` of type `typeof(null)` to parameter `int...`
```
JinShil: I suggest special casing `null` so it reads:  "cannot pass `null` to parameter `int...`".  "`null` of type `typeof(null)` is quite an odd thing to say.  I'll still approve this PR without this change, as this PR is a huge improvement over the current implementation, so it's your call.

src/dmd/expressionsem.d
```diff
@@ -3503,16 +3507,19 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
                 {
                     OutBuffer buf;
                     buf.writeByte('(');
                     argExpTypesToCBuffer(&buf, exp.arguments);
                     buf.writeByte(')');
 
                     //printf("tf = %s, args = %s\n", tf.deco, (*arguments)[0].type.deco);
-                    .error(exp.loc, "`%s%s` is not callable using argument types `%s`", exp.e1.toChars(), parametersTypeToChars(tf.parameters, tf.varargs), buf.peekString());
-
+                    .error(exp.loc, "`%s%s` is not callable using argument types `%s`",
```
JinShil: I'm not seeing any test coverage for this error message.

JinShil (changes requested):

src/dmd/expressionsem.d
ntrel: I'm not sure how to hit it - when is a `VarExp` a function declaration?

JinShil: Yeah, I'm trying to figure that out too.  test/fail_compilation/ice11919.d hits it with `doBar`, but that's as far as I've got.  If I come up with more info, I'll let you know.

JinShil: This will hit the scope just outside of the error.

```D
void doBar(T)(T t)  { }

void main()
{
    int x;
    doBar(x);
}
```
But I still can't figure out how to hit the error.  I'm beginning to wonder if it's even possible.

ntrel: I've come close with this: 
```d
void g()
{
    (g(5));
}
```
But the `ve.hasOverloads` test is true, despite `g` not having overloads, so the error branch is not hit.

test/fail_compilation/bug15613.d
ntrel: Ideally, maybe. This is a general issue though, e.g. `int x = null;` gives:

    funcvar.d(5): Error: cannot implicitly convert expression `null` of type `typeof(null)` to `int`

src/dmd/expressionsem.d
JinShil: This is clearly a non-trivial problem.  I don't think it should hold up this PR.

JinShil (approved):
Definitely an improvement.

wilzbach:
Rebased to `upstream/master`

src/dmd/dtemplate.d
wilzbach: This should be in Ddoc style:
- `Params:` 
- use `=` - then Ddoc can would throw an error about the non-existing `failedIndex`

wilzbach (approved):
Really cool!

ntrel:
@wilzbach Thanks for rebasing. I changed the ddoc comments to use `Params:` as suggested.

wilzbach:
Rebased again ;-)

Geod24:
So yesterday I dusted off an old project and tried to compile. Sure thing, I ran out of memory. This project used to take 2, max 3 GBs of memory. Now it runs out of memory after reaching 7 GBs.

As a result I started to bisect, which led me to this commit after a bit of gymnastic (having `-de` in the Makefile is really annoying when doing that exercise).
Trying to bisect this PR was rather annoying: the first commit (48c6b49605e265e043e9a37fe6064ec603679619) segfault on this code:
```D
class Bar { this(int) {} }
void main () { scope b = new Bar; }
```
but that was fixed by later refactoring. Second commit (ff99241a38068e85af6e2625e10ce78b2201de57) doesn't compile because it was blindly rebased after the `STCref` => `STC.ref_` refactoring, and so everything before commit 6 (15b1b503440caab6ba8d2879b80d3946d4d5ac23) does not compile. That makes solving issues more complicated than it needs to be.

Now the cherry on the top: this pull request (or more precisely, [commit 6](https://github.com/dlang/dmd/pull/7554/commits/15b1b503440caab6ba8d2879b80d3946d4d5ac23#diff-6fb7b3e81a4136ed3f4cc3992fa2643fR6177)) allocates an error message for **every matching failure during call resolution**. And that memory is allocated with `mem.xmalloc` and never freed. I added `scope (exit) rmem.xfree` here and there but it did not help (because of the bump the pointer I guess, or the very high fragmentation it creates?).

It comes down to commit 6, which eagerly generates error message. @ntrel what was the motivation for it ?

ntrel:
@Geod24 Thanks for finding this. `failedIndex` is not enough information to generate the (updated) error message on the caller side.

I think the problem is the call to `functionResolve` - it won't immediately error after that call so `failMessage` should not be passed there. Instead we can re-call the function where `failMessage` is actually needed (warning: not tested), please try:
https://github.com/ntrel/dmd/tree/resolve-msg

PR #4174, opened 2014-11-28
Partial fix for 13761 - Fix code which triggers warnings

Temtaime:
https://issues.dlang.org/show_bug.cgi?id=13761

PR #7557, opened 2017-12-31
Partial fix for Issue 13761 - Fix code which triggers warnings

ibuclaw:
The bits of #4174 that are worth reviving.  Though now that the majority of sources are converted to D, open bugs relating to C++ static analysis tools should be closed, as they either don't apply to D, or cannot be verified.


ibuclaw:
FYI @Temtaime

JinShil (approved):

PR #7560, opened 2017-12-31
Add GDC stubs to gluelayer module

ibuclaw:
One part of #2194

DMD is built with `-version=MARS`, GDC is built with `-version=IN_GCC`, so separating differences between the two is easy.


PetarKirov (approved):
Let's minimize the front-end diff so we can work towards (semi-)automatic PR merging with GDC (and hopefully LDC too).

ZombineDev:
(Restarted the project tester.)

PR #6827, opened 2017-05-24
remove .sort / .reverse properties and adjust tests

UplinkCoder:
as per title
This is a revamp of #6554

PR #5232, opened 2015-10-26
Strive for an ABI-agnostic handling of TypeInfo_Struct.xopEquals and xopCmp

kinke:
While trying to use a non-reversed parameter order for `extern (D)` in LDC on x86_64, I stumbled upon 2 ugly places assuming a reversed parameter order:
1. druntime's `TypeInfo_Struct.compare()` (see https://github.com/D-Programming-Language/druntime/pull/1417)
2. DMD front-end when generating the static `__xopEquals`/`__xopCmp` functions

The problem is that a pointer to an extern(D) function taking 2 struct references is used to jump either to:
- an extern(D) method `int S.opCmp(ref const S q)`  with LL signature `int opCmp(S* this = p, S* q)`, or
- an extern(D) static function `static int S.__xopCmp(ref const S p, ref const S q)` with LL signature `int __xopCmp(S* q, S* p)` if reversing the args. This function is generated by the front-end in case the `opCmp` method doesn't take its single argument by reference.

For this to work, `TypeInfo_Struct.compare()` assumed reversed order for extern(D) and called the extern(D) function pointer `TypeInfo_Struct.xopCmp` with reversed args `(q, p)`, to get the direct method call right (=> LL order `(p, q)`).
In the DMD front-end, the `static int S.__opCmp(ref const S p, ref const S q)` wrapper function needed to take this into account. So it also assumed a reversed arguments order for extern(D) and re-swapped p and q back again (=> `return q.opCmp(p)`).

When not reversing the args for `extern (D)` (as it should be on all non-Win32 platforms, according to the docs!), there's no need for any swapping at all.
I opted for an `extern (C)` approach for function pointer and static function, as that is lowered to a LL signature identical to the method's, as long as the `this` pointer is passed before the first explicit arg, another (but not new) assumption. I'm not sure whether that's free of any undesired effects though, e.g., for exception handling; name mangling seems to be fine with LDC.

At least these 2 places don't have to assume anything about the parameter order of `extern (D)` anymore.

PR #7561, opened 2017-12-31
Reverse parameter order of xopCmp to match extern(C) ABI

ibuclaw:
Required to support dlang/druntime#2025  (and tests will fail until it is pulled also).

### DIFF
--- a/src/dmd/clone.d
+++ b/src/dmd/clone.d
@@ -510,10 +510,12 @@ extern (C++) FuncDeclaration buildXopEquals(StructDeclaration sd, Scope* sc)
     }
     Loc declLoc = Loc(); // loc is unnecessary so __xopEquals is never called directly
     Loc loc = Loc(); // loc is unnecessary so errors are gagged
+    // Use C linkage: we need to make sure the arguments order (p, q) isn't reversed, in
+    // order to obtain a signature identical to that of method `S.opEquals(ref const S)`.
     auto parameters = new Parameters();
     parameters.push(new Parameter(STC.ref_ | STC.const_, sd.type, Id.p, null));
     parameters.push(new Parameter(STC.ref_ | STC.const_, sd.type, Id.q, null));
-    auto tf = new TypeFunction(parameters, Type.tbool, 0, LINK.d);
+    auto tf = new TypeFunction(parameters, Type.tbool, 0, LINK.c);
     Identifier id = Id.xopEquals;
     auto fop = new FuncDeclaration(declLoc, Loc(), id, STC.static_, tf);
     fop.generated = true;
@@ -524,7 +526,7 @@ extern (C++) FuncDeclaration buildXopEquals(StructDeclaration sd, Scope* sc)
     uint errors = global.startGagging(); // Do not report errors
     Scope* sc2 = sc.push();
     sc2.stc = 0;
-    sc2.linkage = LINK.d;
+    sc2.linkage = LINK.c;
     fop.dsymbolSemantic(sc2);
     fop.semantic2(sc2);
     sc2.pop();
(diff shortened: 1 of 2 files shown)


ibuclaw:
Hmm that's interesting, all tests pass.

I'll have to dig up the original patch which reversed these arguments.  Because *something* should have broke.

ibuclaw:
IIRC it was the sort property, which may mean that some test was removed that validates this particular call works.

ibuclaw:
Yep, test was removed (test 5854) in #6827.

ibuclaw:
Re-added test which validates the call to compiler-generated  `__xopCmp()`.

jpf91:
> IIRC it was the sort property, which may mean that some test was removed that validates this particular call works.

The sort property has been removed: https://dlang.org/blog/2017/07/24/new-d-compiler-release-dmd-2-075-0/

kinke:
So this is a rebase of https://github.com/dlang/dmd/pull/5232.

ibuclaw:
@kinke - right, I missed that.  I wrote this independently, but will have a look at yours.

kinke:
It appears to be exactly the same minus a few comments of mine and plus your test. ;)

IIRC, this approach doesn't work for LDC on 32-bit x86. A method `extern(D) int opCmp([ref T this,] ref T rhs)` and a static function `extern(C) int opCmp(ref T this, ref T rhs)` aren't equivalent ABI-wise - the method's `this` would be passed in EAX, while the static function expects both pointers to be passed on the stack. See https://github.com/dlang/dmd/pull/5232#issuecomment-177538912.

The real problem is that a function pointer is used to invoke either an `extern(D)` method or a static function. That's why my last suggestion was using a static function in any case for the TypeInfo's function pointer, even if `opCmp([ref] const T rhs)` does take its argument by reference (=> trampoline).

ibuclaw:
> IIRC, this approach doesn't work for LDC on 32-bit x86. 

Right, but you would have the same problem calling a nested function from a delegate, then using that same delegate to call a method also.

As I said in the other PR, I've chosen to ignore supporting `thiscall` for the moment until a satisfactory solution comes up.

kinke:
As this isn't about `thiscall` but rather about the D ABI on Win32, DMD should be affected by this too and so shouldn't work with the druntime patch.

ibuclaw:
There is no special Windows ABI, unless you mean the non-standard optlink calling convention that DMD unfavourably implements, which would be better off being dropped altogether.

You shouldn't support that in LDC either.

kinke:
Don't know if it's Optlink specific, but it's D's official ABI for Win32 (https://dlang.org/spec/abi.html#function_calling_conventions) and as such also implemented by LDC (for all 32-bit x86 platforms, most likely because that's what inline assembly in druntime/Phobos expects...).

Anyway, I just wanted to emphasize that the revived changes most likely still don't work for DMD on Win32 as long as that ABI isn't ditched, and that something like the enforced static function would be required in the meantime.

ibuclaw:
> Don't know if it's Optlink specific,

The similarities are too uncanny for it to be a coincidence.

https://www.ibm.com/support/knowledgecenter/SSQ2R2_9.5.0/com.ibm.etools.cbl.win.doc/topics/rpsubw30.htm

https://dlang.org/spec/abi.html#function_calling_conventions

wilzbach:
(rebased)

wilzbach:
(rebased)

WalterBright:
> The similarities are too uncanny for it to be a coincidence.

I've never heard of IBM's "optlink" convention, and it has nothing to do with our Optlink linker. A linker does not define a calling convention anyway.

The win32 calling convention goes back to Microsoft's C++ compiler on win32. I tried to be compatible with it.

WalterBright:
> Hmm that's interesting, all tests pass.

Probably because the test cases do not have side effects in the arguments, so reversing the order of non-existent side effects will have no effect.

Geod24:
@ibuclaw : Do you want to pursue this now that GDC is up to date with master ?

Geod24:
Ping @ibuclaw

ibuclaw:
> Ping @ibuclaw 

Well I still need to patch dmd to reverse the arguments. I assume ldc is the same too.

ibuclaw:
Superseded by #13590

PR #7563, opened 2017-12-31
Upstream GDC patch to mangle va_list as a pointer.

ibuclaw:
The default behaviour is to raise an internal compiler error and exit if cppmangle encounters a static array parameter.  This is not compatible with how GDC represents va_lists internally.

In the codegen pass, GDC will saturate a va_list to a pointer if its underlying type is a static array.

Patch taken from #2194 

However... I think that a better way to do this would be to move it to a backend hook, such as `Target.cppTypePassedAs(Type t)`.

### DIFF
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -657,6 +657,19 @@ private final class CppMangleVisitor : Visitor
                 td = new TypeDelegate(td);
                 t = merge(t);
             }
+            static if (IN_GCC)
+            {
+                // Could be a va_list, which we mangle as a pointer.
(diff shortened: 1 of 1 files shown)


ZombineDev:
@klickverbot Obviously we should all aim at reducing the need for `IN_LLVM` and `IN_GCC`, but shorter-term and more important goal for me is to make making merging upstream frontend changes in LDC and GDC as easy as possible and one day testing each dmd/druntime/phobos PR on our CI infrastructure with the LLVM and GCC backends.

ibuclaw:
This wouldn't be so arbitrary if it was a condition-less runtime call.

PR #4072, opened 2014-10-17
Issue 12004 and issue 13174: Disallow shared destructors and always call the unshared one.

mleise:
When using shared structs, the compiler asks for a shared ~this() which should not exist in the language at all. The destructor for these value types must be run when all threads that could reference it have finished and only the owner thread is left. It is a logical error to have a destructor run in a "shared" environment.

https://issues.dlang.org/show_bug.cgi?id=12004
https://issues.dlang.org/show_bug.cgi?id=13174

It's only 6 lines of changes, but the indentation level change in src/template.c makes it look wild. What it says is basically:

``` D
if (fd->isDtorDeclaration() && tthis_fd->isShared())
{
    tthis_fd = tthis_fd->unSharedOf();
}
```

Since I hit the bug once again and there was now someone else struggling with it and `shared` needs some love, I decided to peek into the dmd code and make some quick adjustments. There is one thing I'm not sure about: Is calling an unshared dtor with a `shared this` kosher? It all works as expected, and I ran the test suite and unittests, so I'm quite confident it didn't cause major shockwaves inside the compiler.

PR #7566, opened 2017-12-31
fix Issue 12004 and Issue 13174: Disallow shared destructors and always call the unshared one.

ibuclaw:
Rebase of  #4072 on ddmd.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -3410,7 +3410,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             dd.type = new TypeFunction(null, Type.tvoid, false, LINK.d, dd.storage_class);
 
         sc = sc.push();
-        sc.stc &= ~STC.static_; // not a static destructor
+        sc.stc &= ~(STC.static_|STC.shared_); // not a static destructor
         if (sc.linkage != LINK.cpp)
             sc.linkage = LINK.d;
 
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -2397,24 +2397,33 @@ void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiar
         /* For constructors, qualifier check will be opposite direction.
          * Qualified constructor always makes qualified object, then will be checked
          * that it is implicitly convertible to tthis.
+         *
+         * For destructors, the shared qualifier is stripped from the object.
          */
         Type tthis_fd = fd.needThis() ? tthis : null;
         bool isCtorCall = tthis_fd && fd.isCtorDeclaration();
-        if (isCtorCall)
+        if (tthis_fd)
         {
-            //printf("%s tf.mod = x%x tthis_fd.mod = x%x %d\n", tf.toChars(),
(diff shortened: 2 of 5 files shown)


ibuclaw:
There seem to be a few breaking problems with this patch.

ibuclaw:
Also added test for 8295, which 12004 and 13174 were marked a duplicate of.

ibuclaw:
Last review by @WalterBright:

> I agree that a shared destructor probably doesn't make sense. Not sure about the implementation yet.

ibuclaw:
Rebased against master.

ibuclaw:
Rebased against master.

wilzbach:
Rebased to master (without conflicts).

wilzbach (approved):
Looks reasonable to me. Let's not let this die again?

wilzbach:
From the previous thread:

Walter:
> I agree that a shared destructor probably doesn't make sense. Not sure about the implementation yet.

Andrei:
> Qualified cdtors are needed because objects may make different layout/flow decisions depending on the qualifier. (This is e.g. a classic problem in C++, which cannot distinguish in a constructor whether an actual const object is being constructed.)

> This all is in fact useful even if indeed shared destructors are never called simultaneously from distinct threads.

Marco:

> Now to get back to the issue at hand, consider you are building an aggregate (classes, structs) from other aggregates and decide to call the shared ctor to build a "shared" instance. I believe this doesn't fly in praxis, because now everything that aggregate is made of needs to work as a "shared" instance, too. If nothing else it now requires an explicit or compiler-generated shared ~this for all structs. But it usually makes the most vetted D coders cry and surrender when they try to pass a shared(HANDLE) into a WinAPI.

Is this a stale mate or can we resolve this to a consensus?

CC @andralex

andralex:
Well, let's consider a couple of examples:

```d
shared struct A
{
    ...
    ~this() { ... }
}
```

In this case it's virtually impossible to define the destructor as nonshared, so it needs to be an exception. That's not necessarily bad, but it is a complication.

The more problematic example is:

```
struct A {
    void* opaque;
    this(int x) { ... initialize opaque in some way ... }
    this(int x) shared { ... initialize opaque in a very different way way ... }
    ~this() { ... we know the non-shared ctor was called ... }
    ~this() shared { ... we know the shared ctor was called ... }
}
```

The scheme will be made impossible by this PR. User code will need to remember at runtime which constructor was called, which seems wasteful.

This may seem like an academic problem but we've hit a version of it in @edi33416's collection code. There, a collection needed to remember which constructor was called - immutable vs const vs mutable - so as to initialize its allocator properly. Then, every time an allocator is used a dynamic decision needs to be made.

The politically correct thing to do is allow `shared` constructors and destructors but typecheck them differently in recognition of the fact that there's no contention during construction and destruction. I reckon that's an arduous path.

I'm not vetoing this, just clarifying we're not looking at a simple problem. Opinions welcome.

jacob-carlborg:
> The politically correct thing to do is allow shared constructors and destructors but typecheck them differently in recognition of the fact that there's no contention during construction and destruction. I reckon that's an arduous path.

Would a shared constructor require a shared destructor, or can it fallback on a non-shared destructor?

andralex:
@jacob-carlborg in all likelihood

PR #7567, opened 2017-12-31
Version platform-specific files

wilzbach:
This allows to simplify the build system.

### DIFF
--- a/src/dmd/libelf.d
+++ b/src/dmd/libelf.d
@@ -11,6 +11,9 @@
 module dmd.libelf;
 
 // Online documentation: https://dlang.org/phobos/dmd_libelf.html
+version(Windows) {}
+else version(OSX) {}
+else:
 
 import core.stdc.time;
 import core.stdc.string;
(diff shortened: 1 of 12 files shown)


wilzbach:
See also https://github.com/dlang/dlang.org/pull/2040 for the respective dlang.org cleanup.

jacob-carlborg:
I think this is a bad idea. We should strive do the opposite: include all these platform specific files on all platforms and decide which object format/backend should be used at runtime (of the compiler). This will make it possible to cross-compile.

WalterBright:
@jacob-carlborg I think you're right as an eventual goal, but in the meantime this is a small improvement.

src/dmd/libelf.d
```diff
@@ -11,6 +11,9 @@
 module dmd.libelf;
 
 // Online documentation: https://dlang.org/phobos/dmd_libelf.html
+version(Windows) {}
+else version(OSX) {}
+else:
```
ibuclaw: Isn't there version identifiers for binary formats?

jacob-carlborg: As far as I can see there are `TARGET_*` for the target platform and `OBJ_*` for the object format. But the latter is set at runtime.

jacob-carlborg:
@wilzbach you made it a lot more complicated for me with this PR. https://github.com/dlang/dmd/pull/7513.

PR #1499, opened 2013-01-17
Fix Issue 3632 - modify float is float to do a bitwise compare

yebblies:
Currently floating point types are the only types where 'is' is not a bitwise comparison. This patch makes floating is floating perform a bitwise comparison, at runtime and compile-time.

http://d.puremagic.com/issues/show_bug.cgi?id=3632

PR #7568, opened 2017-12-31
fix Issue 3632 - modify float is float to do a bitwise compare

ibuclaw:
Currently floating point types are the only types where `is` is not a bitwise comparison. This patch makes floating is floating perform a bitwise comparison, at runtime and compile-time.

Rebase of #1499

### DIFF
--- a/src/dmd/constfold.d
+++ b/src/dmd/constfold.d
@@ -938,29 +938,56 @@ UnionExp Identity(TOK op, const ref Loc loc, Type type, Expression e1, Expressio
         SymOffExp es2 = cast(SymOffExp)e2;
         cmp = (es1.var == es2.var && es1.offset == es2.offset);
     }
-    else
+    else if (e1.op == TOKstructliteral && e2.op == TOKstructliteral)
     {
-        if (e1.type.isreal())
-        {
-            cmp = RealEquals(e1.toReal(), e2.toReal());
-        }
-        else if (e1.type.isimaginary())
-        {
-            cmp = RealEquals(e1.toImaginary(), e2.toImaginary());
-        }
-        else if (e1.type.iscomplex())
-        {
-            complex_t v1 = e1.toComplex();
-            complex_t v2 = e2.toComplex();
-            cmp = RealEquals(creall(v1), creall(v2)) && RealEquals(cimagl(v1), cimagl(v1));
-        }
+        StructLiteralExp es1 = cast(StructLiteralExp)e1;
+        StructLiteralExp es2 = cast(StructLiteralExp)e2;
+
+        if (es1.sd != es2.sd)
+            cmp = 0;
+        else if ((!es1.elements || !es1.elements.dim) &&
+            (!es2.elements || !es2.elements.dim))
+            cmp = 1;            // both arrays are empty
+        else if (!es1.elements || !es2.elements)
+            cmp = 0;
+        else if (es1.elements.dim != es2.elements.dim)
+            cmp = 0;
         else
         {
-            ue = Equal((op == TOK.identity) ? TOK.equal : TOK.notEqual, loc, type, e1, e2);
-            return ue;
+            cmp = 1;
+            for (size_t i = 0; i < es1.elements.dim; i++)
+            {
+                Expression ee1 = (*es1.elements)[i];
+                Expression ee2 = (*es2.elements)[i];
+
+                if (ee1 == ee2)
+                    continue;
+                if (!ee1 || !ee2)
+                {
+                    cmp = 0;
+                    break;
+                }
+                ue = Identity(TOKequal, loc, Type.tint32, ee1, ee2);
+                if (ue.exp().op == TOKcantexp)
+                    return ue;
+                cmp = cast(int)ue.exp().toInteger();
+                if (cmp == 0)
+                    break;
+            }
         }
     }
-    if (op == TOK.notIdentity)
+    else if (e1.type.isfloating() && e2.type.isfloating())
+    {
+        complex_t v1 = e1.toComplex();
+        complex_t v2 = e2.toComplex();
+        cmp = RealEquals(creall(v1), creall(v2)) && RealEquals(cimagl(v1), cimagl(v2));
+    }
+    else
+    {
+        ue = Equal((op == TOKidentity) ? TOKequal : TOKnotequal, loc, type, e1, e2);
+        return ue;
+    }
+    if (op == TOKnotidentity)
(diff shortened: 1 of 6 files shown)


JinShil (dismissed):

src/dmd/constfold.d
```diff
@@ -940,28 +940,55 @@ extern (C++) UnionExp Identity(TOK op, Loc loc, Type type, Expression e1, Expres
...
+
+        if (es1.sd != es2.sd)
+            cmp = 0;
+        else if ((!es1.elements || !es1.elements.dim) &&
+            (!es2.elements || !es2.elements.dim))
+            cmp = 1;            // both arrays are empty
+        else if (!es1.elements || !es2.elements)
+            cmp = 0;
+        else if (es1.elements.dim != es2.elements.dim)
+            cmp = 0;
```
WalterBright: The above code is hard to read. How about:
```
const length1 = es1.elements ? es1.elements.dim : 0;
const length2 = es2.elements ? es2.elements.dim : 0;
cmp = length1 = length2;
```

src/dmd/constfold.d
```diff
@@ -940,28 +940,55 @@ extern (C++) UnionExp Identity(TOK op, Loc loc, Type type, Expression e1, Expres
...
+                {
+                    cmp = 0;
+                    break;
+                }
+                ue = Identity(TOKequal, loc, Type.tint32, ee1, ee2);
+                if (ue.exp().op == TOKcantexp)
+                    return ue;
+                cmp = cast(int)ue.exp().toInteger();
+                if (cmp == 0)
+                    break;
```
WalterBright: None of the above new code is covered by any tests.

src/dmd/ctfeexpr.d
```diff
@@ -1384,6 +1384,43 @@ extern (C++) int ctfeIdentity(Loc loc, TOK op, Expression e1, Expression e2)
         complex_t v2 = e2.toComplex();
         cmp = RealEquals(creall(v1), creall(v2)) && RealEquals(cimagl(v1), cimagl(v1));
     }
+    else if (e1.op == TOKstructliteral)
+    {
+        StructLiteralExp es1 = cast(StructLiteralExp)e1;
+        StructLiteralExp es2 = cast(StructLiteralExp)e2;
+
+        if (es1.sd != es2.sd)
+            cmp = 0;
```
WalterBright: Please add tests for red cases.

src/dmd/e2ir.d
```diff
@@ -2512,6 +2512,36 @@ elem *toElem(Expression e, IRState *irs)
                 // we can skip the compare if the structs are empty
                 e = el_long(TYbool, ie.op == TOKidentity);
             }
+            else if (t1.ty == Tcomplex80 && Target.realpad != 0)
+            {
+                /* creal has padding in the middle on some platforms,
+                 * so do identity comparison on real and imaginary parts
+                 * separately.
```
WalterBright: Please add comment illustrating what the following generated code looks like as an expression.

src/dmd/expression.d
```diff
@@ -1246,7 +1246,7 @@ private:
  */
 extern (C++) int RealEquals(real_t x1, real_t x2)
```
WalterBright: Calling this function `Equals` and then doing an `Identity` comparison is vastly confusing. `==` and `is` are fundamentally different operations for floating point, and we need to be careful not to conflate them.

test/runnable/ufcs.d
```diff
@@ -128,8 +128,8 @@ void test3()
     assert(1    .init == 0);
     assert([1]  .init == null);
     assert([1:1].init == null);
-    assert(1.0  .init is double.nan);
-    assert(10i  .init is idouble.nan);
+    assert(1.0  .init is double.init);
```
WalterBright: why change this?

ibuclaw: Init and nan have different bit patterns. I imagine was the reason why the original patch changed this.

WalterBright: They should have the same bit patterns. We tried to make .init a signalling nan, but this was a complete failure. .init and .nan should both be a quiet nan.

ibuclaw: Is there an issue for this? Also detailing the why and how it is a failure. (I've never been involved in any discussion that made that claim).

Sounds like such a language change should be a prerequisite for this anyhow, and not made here.

WalterBright: Because a signalling nan converts to a quiet nan when used. The problem was nobody could come up with a consistent definition of "used".

Geod24:
Ping @ibuclaw

test/runnable/ufcs.d
thewilsonator: >They should have the same bit patterns.

They don't
```d
double a, b = double.nan;
    writeln(*cast(ulong*)&a, "\n", *cast(ulong*)&b);
```
prints
```
9219994337134247936
9221120237041090560
```

ibuclaw: They currently don't. They should though.

I take this to mean that the signalling nan init should be deprecated somehow.

thewilsonator: Ah, right.

thewilsonator: > is there an issue for this?

There is now: https://issues.dlang.org/show_bug.cgi?id=19905

thewilsonator: This is fixed.

thewilsonator:
Rebased.

src/dmd/expression.d
thewilsonator: #9918

ZombineDev:
Should this PR be closed, now that #9897 has been merged?

thewilsonator:
Yes, thanks.

n8sh:
>Should this PR be closed, now that #9897 has been merged?

Why? That PR doesn't do anything to change the underlying issue that this PR addresses. Expressions like `real.nan is -real.nan` still evaluate to true despite not being bitwise equivalent.

PetarKirov:
> > Should this PR be closed, now that #9897 has been merged?

> Why? 

Reading my message 1-2 years later, I was also unsure. Perhaps my thinking was that changing the `T.init` value for floating point types will also the issue regarding bitwise comparison with `is`.

@ibuclaw was is left to be done, in addition to rebasing?

RazvanN7:
ping @ibuclaw

RazvanN7:
Closing in favor of: #13780

PR #7569, opened 2017-12-31
fix issue 18149 - Add a compiler trait to detect if a function is @disable

ghost:
Although there's not much concrete application (since a call to a `@disable` function is already statically checked) for now there is no way to know if a function is `@disable`.

### DIFF
--- a/changelog/disabled-trait.dd
+++ b/changelog/disabled-trait.dd
@@ -1,0 +1,43 @@
+A compiler trait used to detect if a function is marked with `@disable` has been added.
+
+Prior to this release is was impossible to filter out `@disable` functions without
+using `__trait(compiles)`, which was less than ideal since `false` could be
+returned for other reasons.
+
+Now, in metaprogramming code, `@disable` functions can be detected accurately,
+using `__traits(isDisabled)` and even in overload sets:
+
+---
+module runnable;
+
+struct Foo
+{
+    import std.stdio;
+    @disable static void foo() {__PRETTY_FUNCTION__.writeln;}
+    static void foo(int v) {__PRETTY_FUNCTION__.writeln;}
+    static void bar() {__PRETTY_FUNCTION__.writeln;}
+    @disable static void bar(int v) {__PRETTY_FUNCTION__.writeln;}
+}
+
+void test(T)()
+{
+    foreach (member; __traits(allMembers, T))
+        foreach (overload; __traits(getOverloads, T, member))
+            static if (!__traits(isDisabled, overload))
+    {
+        static if (is(typeof(&overload) == void function()))
+            overload();
+        else static if (is(typeof(&overload) == void function(int)))
+            overload(42);
+    }
+}
+
+void main(){test!Foo;}
+---
+
+prints:
+
+$(CONSOLE
+void runnable.Foo.foo(int v)
+void runnable.Foo.bar()
+)
--- a/src/dmd/declaration.d
+++ b/src/dmd/declaration.d
@@ -480,6 +480,11 @@ extern (C++) abstract class Declaration : Dsymbol
         return (storage_class & STC.deprecated_) != 0;
     }
 
+    final bool isDisabled() const pure nothrow @nogc @safe
+    {
+        return (storage_class & STC.disable) != 0;
+    }
+
     final bool isOverride() const pure nothrow @nogc @safe
     {
         return (storage_class & STC.override_) != 0;
--- a/src/dmd/id.d
+++ b/src/dmd/id.d
@@ -354,6 +354,7 @@ immutable Msgtable[] msgtable =
     { "isTemplate" },
     { "isPOD" },
     { "isDeprecated" },
+    { "isDisabled" },
     { "isFuture" },
     { "isNested" },
     { "isFloating" },
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -99,6 +99,7 @@ shared static this()
         "isAbstractClass",
         "isArithmetic",
         "isAssociativeArray",
+        "isDisabled",
         "isDeprecated",
         "isFuture",
         "isFinalClass",
@@ -577,6 +578,13 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
 
         e.error("aggregate or function expected instead of `%s`", o.toChars());
         return new ErrorExp();
+    }
+    if (e.ident == Id.isDisabled)
+    {
+        if (dim != 1)
+            return dimError(1);
+
+        return isFuncX(f => f.isDisabled());
     }
     if (e.ident == Id.isAbstractFunction)
     {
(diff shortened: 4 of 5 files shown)


JinShil:
Is there any particular reason why `@disabled` can't be added to the `getAttributes` trait?

BBasile:
> Adding @disable to getAttributes would introduce a breaking change (and may not even be 

Yes. This and also it'd be for the `getFunctionAttributes` verb (the one you ask for is  for the UDAs). There's a lot of code in `std.typecons` that relies on them and `@disable` would not being usable in the same fashion anyway (typically all the wrappers things).

JinShil:
I'm itching to approve this, but before then it needs a [changelog entry](https://github.com/dlang/dmd/tree/master/changelog) in this PR and a documentation update at https://github.com/dlang/dlang.org.

BBasile:
Yes, I was waiting for the next  cleanup of the folder containing the entries. Probably only a matters of days.

BBasile:
@JinShil, a changelog entry is not needed. An issue is referenced so it would appear even without.

JinShil:
> @JinShil, a changelog entry is not needed. An issue is referenced so it would appear even without.

Yes, I know, and I generally agree, but I've been asked to add changelog entries for certain Bugzilla issues, even though they appear in list of fixed issues.  The following are examples:

1.  https://dlang.org/changelog/2.078.0.html#fix16997
2.  https://dlang.org/changelog/2.078.0.html#fix11006

This is a new feature and I think it is appropriate to do some advertising for it.

ZombineDev:
We have had examples of both adding a dedicated changelog entry and relying on Bugzilla (e.g. https://dlang.org/changelog/2.078.0.html#bugfix-list - `isFuture`). IMO we should add changelog entries for all enhancement requests which allow expressing new things in the language, that were not easily doable before.

@RazvanN7 can you add a changelog entry for `isFuture`?

BBasile:
Okay i 've just added the changelog entry and the good news is that this has revealed that the trait is much more useful than i thought.

JinShil (approved):
Thanks @BBasile.  I'm going to leave this open for a little while to give others an opportunity to weigh in.

wilzbach:
> I'm going to leave this open for a little while to give others an opportunity to weigh in.

Yes this is a language change and thus needs approval from either @WalterBright or @andralex.

src/dmd/declaration.d
```diff
@@ -417,6 +417,11 @@ extern (C++) abstract class Declaration : Dsymbol
         return (storage_class & STCdeprecated) != 0;
     }
 
+    final bool isDisabled()
```
WalterBright: Add `const pure nothrow @nogc @safe`

changelog/disabled-trait.dd
```diff
@@ -0,0 +1,43 @@
+A compiler trait used to detect if a function is marked with `@disable` has been added.
+
+Prior to this release is was impossible to filter out `@disable` functions without
+using `__trait(compiles)`, which was less than ideal since `false` could be
+returned for other reasons.
+
+Now, in your metaprogramming code, `@disable` functions can be detected accurately,
```
WalterBright: Remove `your`

WalterBright (changes requested):
see my comments

BBasile:
Okay done. Note that I still think that the changelog entry is a bit too much. It's really a minor addition.

JinShil:
@WalterBright Are you willing to explicitly mark this "Approved"?

JinShil:
@BBasile Please rebase.  It won't elicit any reviews unless it's passing.

wilzbach:
> @BBasile Please rebase. It won't elicit any reviews unless it's passing.

What do you mean by this? I don't see any merge conflicts and only auto-tester has one spurious failure, but auto-tester automatically invalidates and restarts all jobs if at any repo a PR is merged to master.

@WalterBright needs to dismiss his review (I could do that as well, but I'm not sure he explicitly approved the addition).

JinShil:
> What do you mean by this?

It appears this code is using `STCdisabled` instead of the recently refactored `STC.disabled`.  See the failure.  That is why I assumed it needed to be rebased.

wilzbach:
> It appears this code is using STCdisabled instead of the recently refactored STC.disabled. See the failure. That is why I assumed it needed to be rebased.

Ah I see. Though we can easily do that and reduce the latency / "ping-pong" time ;-)

wilzbach:
(I just rebased this PR)

BBasile:
Looks like there's a regression under OSX 32 bits, that happened between the 2018-01-10 22:21:41 and the 2018-01-10.

JackStouffer:
>Looks like there's a regression under OSX 32 bits

Can we please stop spending precious dev time and stalling PRs to support this dead platform?

wilzbach:
Oh that's interesting. It looks similar to what @marler8997 is experiencing.
Yeah we definitely should switch that machine to do `64_64` testing only. CC @braddr 
Alternatively we can also try to ping Brad here: https://github.com/braddr/d-tester/issues

marler8997:
Looking at the logs it is the same issue.  I was able to reproduce on my macbook pro and got a stack trace.  Note that this seg fault only occurs because the autotester is using dmd version 2.068.2, I was able to fix the problem on my macbook by using the next version, 2.069.0.

Here's the stack trace, maybe we should consider patching dmd 2.068.2?
```
Obj::term(char const*) + 2285
obj_end(Library*, File*) + 37
tryMain(unsigned long, char const**) + 12066
_start + 203
start + 33
```

marler8997:
P.S. I worked around the issue by adding some printf statements to a certain location in the code :)  It was hours of painstaking trial and error to massage the code to not cause the segfault.  Tedious work, but no one else is going to do it for me.

braddr:
reminder: I do not define the supported platforms.  I only implement them and remind people when they want them changed for whatever reason that those changes must go through the community and leadership team.  I'm part of the former but not the latter.  So, go, get approval for changing the supported platform set.

BBasile:
@wilzbach, circle ci here too doesn't work anymore, although [it used to](https://circleci.com/gh/dlang/dmd/tree/pull%2F7569). I did rebase or amend several times yesterday and it's not triggered anymore (as for semaphore).

wilzbach:
> circle ci here too doesn't work anymore

Interesting that that only happens with your account, but it seems like closing & reopening worked here...

BBasile:
Recently i revoked circle-ci from my account...but i don't think this could be the cause.

wilzbach:
> but i don't think this could be the cause.

It could also be related to the partial GitHub outage a couple of days ago:

https://status.github.com/messages

BBasile:
@WalterBright, you took a look at it 10 days ago, i've adressed your review comments. Can you take a decision quickly ?

andralex (approved):

PR #5189, opened 2015-10-11
Issue 8765, 9255 - Show informative assert errors

AndrejMitrovic:
Fixes: [Issue 8765](https://issues.dlang.org/show_bug.cgi?id=8765) and [Issue 9255](https://issues.dlang.org/show_bug.cgi?id=9255)

Recreated from https://github.com/D-Programming-Language/dmd/pull/1426

Requires https://github.com/D-Programming-Language/phobos/pull/3730

PR #1426, opened 2013-01-01
Issue 8765 - Assert should show failing expression. + Issue 9255.

ghost:
https://github.com/D-Programming-Language/dmd/pull/1203 had to be closed because Github has some bugs with rebasing. Unlike the previous pull this one has unittests.

Fixes [Issue 8765](http://d.puremagic.com/issues/show_bug.cgi?id=8765) - Assert should print failing expression.
Fixes [Issue 9255](http://d.puremagic.com/issues/show_bug.cgi?id=9255) - Assert should print file name rather than module name, to be consistent with other Exception types.

Note that fixing 9255 is automatic after Fixing 8765 due to the change in code paths (a message now always exists and therefore the if statement is always taken).

PR #7575, opened 2018-01-01
Issue 8765, 9255 - Show informative assert errors

wilzbach:
Rebased https://github.com/dlang/dmd/pull/5189 (orginates from: https://github.com/dlang/dmd/pull/1203 via https://github.com/dlang/dmd/pull/1426)

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -4371,6 +4371,9 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             printf("AssertExp::semantic('%s')\n", exp.toChars());
         }
 
+        // save expression as a a string before any semantic expansion
+        auto assertExpMsg = exp.msg ? null : exp.toChars();
+
         if (Expression ex = unaSemantic(exp, sc))
         {
             result = ex;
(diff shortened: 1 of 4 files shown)


ZombineDev:
@nomad-software it looks like your testing library needs to be updated to not rely on the default assert error message.

src/dmd/expressionsem.d
```diff
@@ -4391,6 +4401,16 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         result = exp;
     }
 
+    private Expression ensureValidMsg(Scope* sc, Expression msg)
```
ibuclaw: Could this be inlined into its one and only caller?

src/dmd/expressionsem.d
```diff
@@ -4333,6 +4333,9 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             printf("AssertExp::semantic('%s')\n", exp.toChars());
         }
 
+        // save expression as a a string before any semantic expansion
+        auto assertExpMsg = exp.msg ? null : exp.toChars();
```
ibuclaw: Right, this should be done before semantic, but is using `toChars` correct here? What happens with multiline expressions?

JinShil:
> Issue 8765 - assert should print the source code for the condition when no message argument present

I side with @WalterBright and his comments in the  Bugzilla issue.  I think if  the developer chose to leave off the assert message, then that was their deliberate choice, and I don't think the compiler should override that choice.  If we want to print the expression by default, then I recommend introducing an `assert` overload or a different `assertXXX` function.

Furthermore, this does have the potential to be a disruptive change for test suites that analyze assert outputs.

I vote to close Issue 8765 with won't fix, and instead encourage an alternate solution as I mentioned above.

src/dmd/expressionsem.d
JinShil: Could `exp.copy()` be used instead?

timotheecour: isn't `exp.copy()` shallow, so that the copy will still be affected by semantic analysis?

wilzbach:
https://github.com/dlang/dmd/pull/8517 supersedes this.

PR #4783, opened 2015-06-28
UDAs for function arguments

IgorStepanov:
Implementation of subj.

This topic was partially discussed ~year ago in [NG](http://forum.dlang.org/post/pepqfcalhuonfszpmooc@forum.dlang.org).
The general idea:

```
class MyController
{
     View foo(@postParam("a") int p_a, @getParam("b") int p_b);
}

....
static if (is(typeof(MyController.foo) PAR == __parameters))
{
   enum attr_0 = __traits(getAttributes, PAR[0 .. 1]);
   enum attr_1 = __traits(getAttributes, PAR[1 .. 2]);
   static assert(attr_0 == postParam("a"));
   static assert(attr_1 == getParam("b"));
}
```

This feature is very usefull for different bindings generators like Web MVC or REST bindings like [Java Spring MVC](http://docs.spring.io/spring/docs/current/spring-framework-reference/html/mvc.html#mvc-ann-requestmapping).

PR #6589, opened 2017-03-04
fix Issue 13855 - multi-module selective import statements

MartinNowak:
- allow to add further `mod : sym1, sym2` blocks to import statement
- only allow non-ambiguous variations (requiring look-ahead of 2 tokens)

PR #6161, opened 2016-09-28
fix issue 9701 - allow UDAs to be attached to enum values

ghost:
Only allows the UDA to be defined before the enum value, in such a way:

```
enum Enum
{
    @("value") value,
}
```

Not sure if the following should be allowed but I don't see why not, though would require reworking a bit of code. Not currently implemented with this commit:

```
enum Enum
{
    @("all"):
    hasAll,
    thisOneToo,

    @("extra"):
    hasAllAndExtra,
} 
```

PR #7576, opened 2018-01-02
UDAs for function arguments

wilzbach:
#4783 revived from the dead.
See https://github.com/dlang/DIPs/pull/105 for a respective DIP.

Official spelling is ["user-defined attributes"](https://github.com/dlang/dlang.org/pull/2003#discussion_r157333463).

### DIFF
--- a/changelog/uda-function-arguments.dd
+++ b/changelog/uda-function-arguments.dd
@@ -1,0 +1,10 @@
+UDAs on function arguments are now supported
+
+User-defined attributes on function arguments behave analogous to existing UDAs:
+
+---
+void test(A)(@(22) A a)
+{
+    static assert([__traits(getAttributes, a)] == [22]);
+}
+---
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -1734,15 +1734,17 @@ struct ASTBase
         Type type;
         Identifier ident;
         Expression defaultArg;
+        UserAttributeDeclaration userAttribDecl; // user defined attributes
 
         extern (D) alias ForeachDg = int delegate(size_t idx, Parameter param);
 
-        final extern (D) this(StorageClass storageClass, Type type, Identifier ident, Expression defaultArg)
+        final extern (D) this(StorageClass storageClass, Type type, Identifier ident, Expression defaultArg, UserAttributeDeclaration userAttribDecl)
         {
             this.storageClass = storageClass;
             this.type = type;
             this.ident = ident;
             this.defaultArg = defaultArg;
+            this.userAttribDecl = userAttribDecl;
         }
 
         static size_t dim(Parameters* parameters)
@@ -1809,7 +1811,7 @@ struct ASTBase
 
         Parameter syntaxCopy()
         {
-            return new Parameter(storageClass, type ? type.syntaxCopy() : null, ident, defaultArg ? defaultArg.syntaxCopy() : null);
+            return new Parameter(storageClass, type ? type.syntaxCopy() : null, ident, defaultArg ? defaultArg.syntaxCopy() : null, userAttribDecl ? cast(UserAttributeDeclaration) userAttribDecl.syntaxCopy(null) : null);
         }
 
         void accept(Visitor v)
@@ -3623,7 +3625,7 @@ struct ASTBase
                     Expression e = (*exps)[i];
                     if (e.type.ty == Ttuple)
                         e.error("cannot form tuple of tuples");
-                    auto arg = new Parameter(STC.undefined_, e.type, null, null);
+                    auto arg = new Parameter(STC.undefined_, e.type, null, null, null);
                     (*arguments)[i] = arg;
                 }
             }
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -1428,7 +1428,7 @@ final class Parser(AST) : Lexer
                             // Disallow:
                             //      void function() @uda fp;
                             //      () @uda { return 1; }
-                            error("user defined attributes cannot appear as postfixes");
+                            error("user-defined attributes cannot appear as postfixes");
                         }
                         continue;
                     }
@@ -2765,9 +2765,10 @@ final class Parser(AST) : Lexer
             StorageClass storageClass = 0;
             StorageClass stc;
             AST.Expression ae;
-
+            AST.Expressions* udas = null;
             for (; 1; nextToken())
             {
+            L3:
                 switch (token.value)
                 {
                 case TOK.rightParentheses:
@@ -2801,7 +2802,27 @@ final class Parser(AST) : Lexer
                         goto default;
                     stc = AST.STC.wild;
                     goto L2;
-
+                case TOK.at:
+                    {
+                        AST.Expressions* exps = null;
+                        StorageClass stc2 = parseAttribute(&exps);
+                        if (stc2 == AST.STC.property || stc2 == AST.STC.nogc ||
+                            stc2 == AST.STC.disable || stc2 == AST.STC.safe ||
+                            stc2 == AST.STC.trusted || stc2 == AST.STC.system)
+                        {
+                            error("`@%s` attribute for function parameter is not supported", token.toChars());
+                        }
+                        else
+                        {
+                            udas = AST.UserAttributeDeclaration.concat(udas, exps);
+                        }
+                        if (token.value == TOK.dotDotDot)
+                            error("variadic parameter cannot have user-defined attributes");
+                        if (stc2)
+                            nextToken();
+                        goto L3;
+                        // Don't call nextToken again.
+                    }
                 case TOK.in_:
                     stc = AST.STC.in_;
                     goto L2;
@@ -2918,6 +2939,30 @@ final class Parser(AST) : Lexer
                             if (hasdefault)
                                 error("default argument expected for `%s`", ai ? ai.toChars() : at.toChars());
                         }
+                        auto param = new AST.Parameter(storageClass, at, ai, ae, null);
+                        if (udas)
+                        {
+                            auto a = new AST.Dsymbols();
+                            auto udad = new AST.UserAttributeDeclaration(udas, a);
+                            param.userAttribDecl = udad;
+                        }
+                        if (token.value == TOK.at)
+                        {
+                            AST.Expressions* exps = null;
+                            StorageClass stc2 = parseAttribute(&exps);
+                            if (stc2 == AST.STC.property || stc2 == AST.STC.nogc ||
+                                stc2 == AST.STC.disable || stc2 == AST.STC.safe ||
+                                stc2 == AST.STC.trusted || stc2 == AST.STC.system)
+                            {
+                                error("`@%s` attribute for function parameter is not supported", token.toChars());
+                            }
+                            else
+                            {
+                                error("user-defined attributes cannot appear as postfixes", token.toChars());
+                            }
+                            if (stc2)
+                                nextToken();
+                        }
                         if (token.value == TOK.dotDotDot)
                         {
                             /* This is:
@@ -2926,11 +2971,11 @@ final class Parser(AST) : Lexer
                             if (storageClass & (AST.STC.out_ | AST.STC.ref_))
                                 error("variadic argument cannot be `out` or `ref`");
                             varargs = 2;
-                            parameters.push(new AST.Parameter(storageClass, at, ai, ae));
+                            parameters.push(param);
                             nextToken();
                             break;
                         }
-                        parameters.push(new AST.Parameter(storageClass, at, ai, ae));
+                        parameters.push(param);
                         if (token.value == TOK.comma)
                         {
                             nextToken();
@@ -4303,7 +4348,7 @@ final class Parser(AST) : Lexer
                         parseStorageClasses(storage_class, link, setAlignment, ealign, udas);
 
                         if (udas)
-                            error("user defined attributes not allowed for `%s` declarations", Token.toChars(tok));
+                            error("user-defined attributes not allowed for `%s` declarations", Token.toChars(tok));
 
                         t = parseType();
                         v = new AST.AliasDeclaration(loc, ident, t);
@@ -4476,7 +4521,7 @@ final class Parser(AST) : Lexer
                  */
 
                 if (udas)
-                    error("user defined attributes not allowed for `%s` declarations", Token.toChars(tok));
+                    error("user-defined attributes not allowed for `%s` declarations", Token.toChars(tok));
 
                 if (token.value == TOK.assign)
                 {
@@ -4706,7 +4751,7 @@ final class Parser(AST) : Lexer
                 parameters = new AST.Parameters();
                 Identifier id = Identifier.generateId("__T");
                 AST.Type t = new AST.TypeIdentifier(loc, id);
-                parameters.push(new AST.Parameter(0, t, token.ident, null));
+                parameters.push(new AST.Parameter(0, t, token.ident, null, null));
 
                 tpl = new AST.TemplateParameters();
                 AST.TemplateParameter tp = new AST.TemplateTypeParameter(loc, id, null, null);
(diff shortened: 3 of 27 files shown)


wilzbach:
Jenkins failure looks unrelated: https://github.com/dlang/ci/pull/105

src/dmd/astbase.d
```diff
@@ -1773,15 +1773,17 @@ struct ASTBase
...
         extern (D) alias ForeachDg = int delegate(size_t idx, Parameter param);
 
-        final extern (D) this(StorageClass storageClass, Type type, Identifier ident, Expression defaultArg)
+        final extern (D) this(StorageClass storageClass, Type type, Identifier ident, Expression defaultArg, UserAttributeDeclaration userAttribDecl)
         {
             this.storageClass = storageClass;
             this.type = type;
             this.ident = ident;
             this.defaultArg = defaultArg;
+            this.userAttribDecl = userAttribDecl;
```
wilzbach: @RazvanN7 do you have any plans on removing the giant code duplication in `astbase`?

jacob-carlborg:
Will this work for template parameters?

wilzbach:
> Will this work for template parameters?

Not this version. 
AFAICT is the proposed DIP also only about parameter attributes (and attaching attributes to enums).
It does work with template _d_ parameters - I just added test cases for that.

test/fail_compilation/udaparams.d
```diff
@@ -0,0 +1,24 @@
...
+void vararg1(int a, @(10) ...);
+extern(C) void vararg2(int a, @(10) ...);
+
+void rhsuda(int a @(10));
+void rhsuda2(int @(10));
+void rhsuda3(int[] arr @(10) ...);
+
+void wrongAttr1(@safe int);
+void wrongAttr2(@safe void function());
+void wrongAttr3(@safe void delegate());
```
mathias-lang-sociomantic: That isn't going to be confusing at all :D

test/runnable/uda.d
```diff
@@ -467,6 +467,113 @@ static assert(__traits(getAttributes, FileData11844.member)[0](new FileData11844
...
+{
+    void test17x(A)(@(23) A a)
+    {
+        static assert([__traits(getAttributes, a)] == [23]);
+    }
+    static assert([ParameterUDA!(0, test17x!int)] == [23]);
+}
+
+/************************************************/
+
```
mathias-lang-sociomantic: Probably could use a test with a tuple as well

jacob-carlborg:
> It does work with template d parameters - I just added test cases for that.

@wilzbach that do you mean with "template _d_ parameters"?

test/fail_compilation/udaparams.d
skl131313: Those are invalid though attributes, so it shouldn't be too big of a problem.

src/dmd/astbase.d
```diff
@@ -1773,15 +1773,17 @@ struct ASTBase
         Type type;
         Identifier ident;
         Expression defaultArg;
+        UserAttributeDeclaration userAttribDecl; // user defined attributes
 
         extern (D) alias ForeachDg = int delegate(size_t idx, Parameter param);
 
-        final extern (D) this(StorageClass storageClass, Type type, Identifier ident, Expression defaultArg)
+        final extern (D) this(StorageClass storageClass, Type type, Identifier ident, Expression defaultArg, UserAttributeDeclaration userAttribDecl)
```
skl131313:
```D
Expression defaultArg, UserAttributeDeclaration userAttribDecl = null)
```

Giving this a default value would save on a lot of code edits that just add a null as the last parameter. Not sure if default values are being avoided here on purpose or not.

test/fail_compilation/udaparams.d
MetaLang: The following works currently:
```
void wrongAttr2(void function() @safe);
void wrongAttr3(void delegate() @safe);
```

And has the expected semantics. The question is, will it cause confusion that putting the `@safe` before the function literal type doesn't work?

John-Colvin:
I'm trying to get a hold on the scope of this PR.
Can I do this?

```D
auto a(int x)(void function(@(x) int) f);
auto b = a((@(2) int){ });
```

skl131313:
@John-Colvin 
Not I don't think that's possible with the current implementation, also not sure if that is desired.

src/dmd/astbase.d
wilzbach: AFAICT is the prevalent coding style in the DMD source code to prefer explicitness.

wilzbach:
So before we let this die again, how about making it available under `-transition=uda` (or `-dip=uda` or `-experimental=uda`)?
It won't get tested without being exposed and easily testable...

Other compilers have tons of experimental features that users can use to test upcoming features and even dmd has `-dip25`, `-dip1008` and `-dip1000` ...

E.g. rust has a lot more experiments going on: https://doc.rust-lang.org/stable/unstable-book/

MetaLang:
We should only add such a switch for something that is almost certain to get into the language. Otherwise, people will begin using it in their code and it won't be able to removed later down the road, even if we want to.

wilzbach:
Experimental switches would be like std.experimental - zero guarantees. No deprecations.

Remember that this addition has been submitted __three years ago__. Imho this level of progress for such tiny things (yes tiny compared to the other big fires D has) is really frustrating and from what I read so far from the responses in the State of D survey, many many other people feel the same. That's why they switched to Rust or kept using C++.

Experimental flags would also solve the chicken&egg problem of showing that it's useful without having a possibility to use it. We can of course also fork the compiler and implement new features there, but I doubt that such a split is a good idea.

MetaLang:
> Experimental switches would be like std.experimental - zero guarantees. No deprecations.
I think we've found with std.experimental that telling people this just doesn't work. They will still use it in their code and will still get mad if we remove the feature and break their code.

> Remember that this addition has been submitted three years ago. Imho this level of progress for such tiny things (yes tiny compared to the other big fires D has) is really frustrating
I completely agree.

> Experimental flags would also solve the chicken&egg problem of showing that it's useful
I disagree.

wilzbach:
> I disagree.

Well, do you have better ideas (except for forking the compiler?)

The only other idea that comes to my mind are feature branches. Martin used to popularize this concept (see e.g. http://nightlies.dlang.org), but they never gained traction because (1) they require manual builds which can only be done by him and (2) they feature branch still can get easily unmergeable into master (see eg. `newCTFE`) and (3) they aren't well-known and exposed.

MetaLang:
> Do you have any better ideas

No I don't, which is frustrating, but my belief that experimental switches won't work is only becoming stronger; especially since people complained after (expected and announced) breaking changes to `std.experimental.allocator` were made. If a big "customer" such as Sociomantic, Weka, etc. start using the feature before it's fully fleshed out and we decide that it's not a good fit for D, good luck _ever_ getting it removed. If it ships with the compiler, people will use it, and they'll be upset when it breaks.

The only thing I can think of off the top of my head is compiler plugins that could be hosted on code.dlang.org. However, dmd currently doesn't support this kind of pluggable architecture, and I have no idea as to whether this could be made to work with LDC/GDC.

jacob-carlborg:
> Well, do you have better ideas (except for forking the compiler?)

As Andrei said in another PR [1]:

"there's no reason for this limitation to exist and it breaks the rule of least astonishment. no dip needed, please make part of the changelog".

I would say we should go for it without a DIP.

[1] https://github.com/dlang/dmd/pull/6589#issuecomment-353998482

MetaLang:
Thanks @jacob-carlborg, I didn't realize Andrei had already pre-approved this; that's great.

jacob-carlborg:
>  I didn't realize Andrei had already pre-approved this; that's great.

@MetaLang he hasn't explicitly approved **this** change. But he thought it was fine to bypass the DIP process for another change: https://github.com/dlang/dmd/pull/6589#issuecomment-353998482.

MetaLang:
Ha, I somehow thought the comment you linked to was on https://github.com/dlang/DIPs/pull/105. Considering that there _is_ a DIP for this feature, maybe we should wait until W&A render their verdict and then come back to this PR.

skl131313:
That didn't turn out so great for that feature, it ended up getting removed. I don't think it is ready yet either. There aren't tests for failed compilation, and there isn't a test for lambda's either.

```D
// something like this I think ?

void test(alias func)()
{
    static if(is(typeof(func) PT == __parameters))
    {
        writeln(__traits(getAttributes, PT[0 .. 1]));
    }
}

test!((@("attribute") int param) => 0)();

```

jacob-carlborg:
> That didn't turn out so great for that feature, it ended up getting removed.

True, but I don't think that was a useful addition to begin with.

> I don't think it is ready yet either. There aren't tests for failed compilation, and there isn't a test for lambda's either.

I didn't say we should merge it this second. Of course it needs to have the proper tests.

JinShil:
> As Andrei said in another PR [1]:
>> "there's no reason for this limitation to exist and it breaks the rule of least astonishment. no dip needed, please make part of the changelog".

> I would say we should go for it without a DIP.

A little history on the DIP requirement:

1.  I posted to the forum asking for an official decision (https://forum.dlang.org/post/cltyrthdxkkfvuxqasqw@forum.dlang.org)
2.  @andralex said not to ask for official decisions on the forum (https://forum.dlang.org/post/ovhj5k$2131$1@digitalmars.com)
3.  I sent a private e-mail to @andralex and @WalterBright 
4.  @WalterBright replied saying a DIP is required.
5.  I posted to the forum stating that I had received an official decision https://forum.dlang.org/post/eiulsnlumpifcxwctnzj@forum.dlang.org
6.  I updated this PR with the "Pending DIP Approval" label

I consider the matter of whether a DIP is required for this or not a settled matter.  If you want the decision to be reconsidered, you'll have to ask @andralex and @WalterBright to do so.  Until then, a DIP is required.

jacob-carlborg:
I just don’t think it’s fair that Walter and Andrei can bypass the DIP process while the rest of us have to struggle so much to get something through.

Also, as I’ve mentioned before, it doesn’t look like depreciations need to go through a similar process, which I think is weird.

wilzbach:
> There aren't tests for failed compilation, and there isn't a test for lambda's either.

I'm aware, but as long as it's not foreseeable that this will be merged, I don't really feel inclined in investing more time in this. Though I don't want to stop other people from working with this (PR to this PR, directly pushing, just commenting with a diff, ...)

andralex:
>I just don’t think it’s fair that Walter and Andrei can bypass the DIP process while the rest of us have to struggle so much to get something through.

That's... odd. Folks, no need to get agitated here. I'm trying to avoid bureaucracy where possible, not pull a Cambridge Analytica on a rainy country. Also note that the proposal and work don't originate from us...

jacob-carlborg:
> I'm trying to avoid bureaucracy where possible

So why is bureaucracy needed for this change? Without this change UDA's are not consistent, not turtles all the way down.

andralex:
>So why is bureaucracy needed for this change? Without this change UDA's are not consistent, not turtles all the way down.

My point exactly! So why are you reversing your position now? Are you running a Socratic discourse? Please clarify, thanks.

jacob-carlborg:
> My point exactly! So why are you reversing your position now? Are you running a Socratic discourse? Please clarify, thanks.

I have not reversed my position. I don't think a DIP is required for this. Mike claims that Walter requires a DIP for this change, see his comment, point 4 https://github.com/dlang/dmd/pull/7576#issuecomment-374807297.

JinShil:
@andralex What about #6161?  Is a DIP no longer required for that as well?

src/dmd/parse.d
```diff
@@ -2896,6 +2915,28 @@ final class Parser(AST) : Lexer
...
+                            auto udad = new AST.UserAttributeDeclaration(udas, a);
+                            param.userAttribDecl = udad;
+                        }
+                        if (token.value == TOK.at)
+                        {
+                            AST.Expressions* exps = null;
+                            StorageClass stc2 = parseAttribute(&exps);
+                            if (stc2 == AST.STC.property || stc2 == AST.STC.nogc || stc2 == AST.STC.disable || stc2 == AST.STC.safe || stc2 == AST.STC.trusted || stc2 == AST.STC.system)
+                            {
+                                error("@%s attribute for function parameter is not supported", token.toChars());
```
JinShil: Tick marks?

src/dmd/traits.d
```diff
@@ -715,8 +717,12 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         Identifier id;
         if (auto po = isParameter(o))
         {
+            if (!po.ident)
+            {
+                e.error("argument %s has no identifier", po.type.toChars());
```
JinShil: Tick marks?

andralex:
@JinShil I see a good DIP has been submitted, the plan is to approve that without too much paperwork.

wilzbach:
Finally got around submitting the dlang.org PR: https://github.com/dlang/dlang.org/pull/2363

test/runnable/uda.d
```diff
@@ -467,6 +467,128 @@ static assert(__traits(getAttributes, FileData11844.member)[0](new FileData11844
...
+void test14x(@Test14UDA1 int, @Test14UDA2("1") int, @test14uda3("2") int, @Test14UDA4!"3" int) {}
+void test14()
+{
+    static assert(is(ParameterUDA!(0, test14x)[0] == Test14UDA1));
+    static assert(ParameterUDA!(1, test14x)[0] == Test14UDA2("1"));
+    static assert(ParameterUDA!(2, test14x)[0] == test14uda3("2"));
+    static assert(is(ParameterUDA!(3, test14x)[0] == Test14UDA4!"3"));
+}
+
+void test15x(@(20) void delegate(int) @safe dg)
```
jacob-carlborg: What about the parameters on delegates, are UDAs on those supported? Either way I think it's worth having a test case, showing either that it works or that it doesn't work.

jacob-carlborg (changes requested):

IgorStepanov:
Ping. What status of that?

jacob-carlborg:
> Ping. What status of that?

I asked for some more tests, which I don't think has been added. Not sure what else.

@wilzbach ping.

wilzbach:
> I asked for some more tests, which I don't think has been added.

Sorry. I haven't found much time for this, but now managed to add these tests. See `test15y` and `test15z`.

> Not sure what else.

Same here.

test/runnable/uda.d
wilzbach: Added the tests below.

changelog/uda-function-arguments.dd
```diff
@@ -0,0 +1,10 @@
+UDAs on function arguments are now supported
+
+User-defined attributes on function arguments behave analogous to existing UDAs:
+
+---
+void test(A)(@(22) A a)
+{
+    static assert([__traits(getAttributes, a)] == [22]);
+}
+---
```
jacob-carlborg: Perhaps include an example of how to get the UDAs of a parameter from outside of a function. Or should that be left as an exercise for the reader 😃.

wilzbach:
> Lambdas, both with and without parentheses: alias a = @(3) b => 1 + 2 and alias b = (@(3) c, @(4) d) => 1 + 2 and usage like: map!(@(3) a => 1 + 2)

Only the latter was supported (i.e. with parentheses). I added support for `alias a = @(3) b => 1 + 2`.
Though I'm not sure that `alias a = @(3) b => 1 + 2 ` will be used in practice.
I added the tests to `test19`, but I'm not sure whether there's currently a way to query the attributes.

> Using a delegate with inferred parameter type: foo((@(3) a) { return 1 + 2; })

Added to `test20`.

> Template parameter: void foo(@(3) T)(T t)

That's not supported as of now. I'm not sure it's that useful, but if people run into use cases for it, it could always be added?
(I still added it to `fail_compilation`)

jacob-carlborg:
> That's not supported as of now

I’m not asking for adding new functionality, just to add tests that shows it’s either compiles or fails to compile.

jacob-carlborg (approved):

changelog/uda-function-arguments.dd
wilzbach: That's a bit more complicated and I think we will provide an interface in `std.traits` anyhow.

jacob-carlborg: Ok.

PR #4396, opened 2015-02-08
Issue 14147 - Compiler crash on identical functions in a single module

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=14147

Partial fix for [issue 2789](https://issues.dlang.org/show_bug.cgi?id=2789).

PR #7577, opened 2018-01-02
Issue 14147 - Compiler crash on identical functions in a single module

wilzbach:
Revived #4396

- [x] https://github.com/dlang/druntime/pull/2027
- [x] https://github.com/dlang/druntime/pull/2033


test/compilable/staticforeach.d
```diff
@@ -120,6 +120,7 @@ static foreach(member;__traits(allMembers,S)){
 }
 
 // print prime numbers using overload sets as state variables.
+/+
 static assert(is(typeof(bad57)));
 static assert(!is(typeof(bad53)));
```
wilzbach: I'm not sure how I can "fix" this test. It generated `bad57` a couple of times and AFAICT there's no `break` for `static foreach` yet.

CC @tgehr

src/dmd/backend/el.d
```diff
@@ -185,7 +185,7 @@ elem *el_var(Symbol *);
 elem *el_settype(elem *,type *);
 elem *el_typesize(type *);
 elem *el_ptr(Symbol *);
-void el_replace_sym(elem *e,Symbol *s1,Symbol *s2);
+//void el_replace_sym(elem *e,Symbol *s1,Symbol *s2);
```
ibuclaw: I don't think body-less function declarations should conflict, unless their signature is ambiguous.

ibuclaw (approved):
Its about time the frontend did this.  We've had some similar logic in GDC codegen for a while now to prevent multiple symbols from being emitted to assembly.  Maybe this patch allows room for simplifying that on our side.

src/dmd/dmangle.d
```diff
@@ -1133,3 +1138,31 @@ extern (C++) void mangleToBuffer(TemplateInstance ti, OutBuffer* buf)
     v.mangleTemplateInstance(ti);
 }
 
+/******************************************************************************
+ * Mangle function signatures ('this' qualifier, and parameter types)
+ * to check conflicts in function overloads.
+ * It's different from fd.type.deco. For example, fd.type.deco would be null
+ * if fd is an auto function.
+ */
+void mangleToFuncSignature(OutBuffer* buf, FuncDeclaration fd)
```
JinShil: Need `Params` DDoc

src/dmd/semantic2.d
```diff
@@ -349,6 +350,88 @@ private extern(C++) final class Semantic2Visitor : Visitor
...
+
+                // extern (C) functions always conflict each other.
+                if (f1.ident == f2.ident &&
+                    f1.toParent2() == f2.toParent2() &&
+                    (f1.linkage != LINKd && f1.linkage != LINKcpp) &&
+                    (f2.linkage != LINKd && f2.linkage != LINKcpp))
+                {
+                    /* Allow the hack that is actually used in druntime,
+                     * to ignore function attributes for extern (C) functions.
+                     * TODO: Must be reconsidered in the future.
```
JinShil: File an issue for this, and add link to the issue?

JinShil (changes requested):

src/dmd/semantic2.d
wilzbach: -> https://issues.dlang.org/show_bug.cgi?id=18206

test/compilable/extra-files/header1.d
```diff
@@ -25,7 +25,7 @@ body
 {
     float f = float.infinity;
     int i = cast(int) f;
-    writeln((i,1),2);
+    writeln(i,1,2);
```
wilzbach: I don't know why the trailing comma deprecation wasn't triggered before.

src/dmd/backend/el.d
wilzbach: It's implemented in exactly the opposite way as DRuntime abused the non-existence of such a check. See the "hack"  below.

JinShil (approved):

src/dmd/dmangle.d
```diff
@@ -1143,17 +1143,21 @@ extern (C++) void mangleToBuffer(TemplateInstance ti, OutBuffer* buf)
  * to check conflicts in function overloads.
  * It's different from fd.type.deco. For example, fd.type.deco would be null
  * if fd is an auto function.
+ *
+ * Params:
+ *    buf = `OutBuffer` to write the mangled function signature to
+*     fd  = `FuncDeclaration` to mangle
  */
-void mangleToFuncSignature(OutBuffer* buf, FuncDeclaration fd)
+void mangleToFuncSignature(ref OutBuffer buf, FuncDeclaration fd)
```
stefan-koch-sociomantic: please keep the pointer if you can

stefan-koch-sociomantic (changes requested):

src/dmd/dmangle.d
```diff
@@ -1133,3 +1138,35 @@ extern (C++) void mangleToBuffer(TemplateInstance ti, OutBuffer* buf)
...
+    //    buf.writeByte(Type.needThisPrefix());
+    MODtoDecoBuffer(&buf, tf.mod);
+    //switch (tf.linkage)
+    //{
+    //    case LINKd:         buf.writeByte('F');     break;
+    //    case LINKc:         buf.writeByte('U');     break;
+    //    case LINKwindows:   buf.writeByte('W');     break;
+    //    case LINKpascal:    buf.writeByte('V');     break;
+    //    case LINKcpp:       buf.writeByte('R');     break;
+    //    default:            assert(0);
```
ibuclaw: Remove these comments?

src/dmd/dmangle.d
wilzbach: Why? This is a D-only function.

src/dmd/dmangle.d
wilzbach: Yeah they were part of the PR I revived and I wanted to port it as with minimal changes to have a good history, but now that I'm thinking about this: no one needs the history of a PR anyways, so I am learning to squash his work into an entire commit and maybe squash it entirely. Do we have any guidelines on this?

klickverbot: I tend to amend the original author's commits accordingly myself (i.e. separate commits by semantics, not authorship), and add a note on the provenance to the commit message. I don't think we have any dlang-specific guidelines.

jacob-carlborg: It's possible to have separate authors and committers as well.

wilzbach: Finally got around removing the comments. For simplicity I amended this to the Fixup commit (the one after Kenji's commits).

wilzbach:
So we are all good here? Anything blocking this?

JinShil:
> So we are all good here? Anything blocking this?

I'm just waiting for the tests to pass before giving this one last review

wilzbach:
> I'm just waiting for the tests to pass before giving this one last review

FYI the Jenkins failure just means that the GCE has terminated the runner - it has nothing to do with this PR:

```
remote file operation failed: /var/lib/jenkins/dlang_projects@6 at hudson.remoting.Channel@4cea69e6:JNLP4-connect connection from 128.115.198.35.bc.googleusercontent.com/35.198.115.128:56402: hudson.remoting.ChannelClosedException: Remote call on JNLP4-connect connection from 128.115.198.35.bc.googleusercontent.com/35.198.115.128:56402 failed. The channel is closing down or has closed down
```

dlang/ci uses cheap preemptible GCE instances: https://cloud.google.com/compute/docs/instances/preemptible

> A preemptible VM is an instance that you can create and run at a much lower price than normal instances. However, Compute Engine might terminate (preempt) these instances if it requires access to those resources for other tasks. Preemptible instances are excess Compute Engine capacity so their availability varies with usage.

BBasile:
cause of a regression in IZ: https://semaphoreci.com/bbasile/iz/branches/master/builds/35

```diff
digger: Finding shortest path between e76ec94e6319212f5fcd379310ec17c2a6bc3661 and 27f0cc399eeaac691f42949b7343baac6b0cdca1...
digger: 0 commits (about 0 tests) remaining.
digger: 27f0cc399eeaac691f42949b7343baac6b0cdca1 is the first bad commit
commit 27f0cc399eeaac691f42949b7343baac6b0cdca1
Author: Sebastian Wilzbach <seb@wilzba.ch>
Date:   Wed Jan 24 12:39:13 2018 +0100

    dmd: Issue 14147 - Compiler crash on identical functions in a single module (#7577)
    
    * fix Issue 2789 - Functions overloads are not checked for conflicts
    
    * Allow an overload declaration hack, which is actually used in druntime
    
    If two extern (C) functions are just declared with different signatures,
    they don't conflict.
    
     extern(C):
     alias sigfn_t  = void function(int);
     alias sigfn_t2 = void function(int) nothrow @nogc;
     sigfn_t  bsd_signal(int sig, sigfn_t  func);
     sigfn_t2 bsd_signal(int sig, sigfn_t2 func) nothrow @nogc;  // no error
    
    This behavior must be reconsidered in the future.
    
    * fix Issue 14147 - Compiler crash on identical functions in a single module
    
    *  Allow an overload declaration hack, which is actually used in druntime
    
    * Fix the testsuite
    
    * PR 4396 Fixup
    
    * Remove duplicate definition in the DMD backend

diff --git a/dmd b/dmd
index c9556e4ef..ab47b8b0c 160000
--- a/dmd
+++ b/dmd
@@ -1 +1 @@
-Subproject commit c9556e4ef7233daac57aa81e6d8aa8c8320c0338
+Subproject commit ab47b8b0cf18d14c1ffbc97df6c8fb299971170a
digger: Bisection completed successfully.                     
```

BBasile:
https://issues.dlang.org/show_bug.cgi?id=18410

MartinNowak:
This broke a subtler usage of `extern(C)` to just select the calling convention but not the mangling.
[18385 – [REG 2.079] method cannot be overloaded with another extern(C) method](https://issues.dlang.org/show_bug.cgi?id=18385)

jacob-carlborg:
Can we revert this, due to issue 18385?

wilzbach:
> Can we revert this, due to issue 18385?

> Walter: Yes, it should never have compiled, and it cannot work, as two functions with the same mangled name cannot both exist in the executable.

but yeah, it looks like this needs a deprecation cycle, so we should probably revert it for now and re-add it with deprecations.

jacob-carlborg:
> but yeah, it looks like this needs a deprecation cycle, so we should probably revert it for now and re-add it with deprecations.

I don't see why it has to be deprecated. The actual code that hit the regression have different mangled names:

```d
class Foo
{
    extern (C) static void callback(int) {}
    extern (C) static void callback(char*) {}
}
```

In the above example the mangled names of `callback` is the same as for `extern (D)`. I reduced the test case a bit too much.

wilzbach:
Well, let's go with https://github.com/dlang/dmd/pull/7969 for now. Once, that's in we can can still decide what is going to be deprecated with 2.080 and what not (the test cases of https://github.com/dlang/dmd/pull/7969 are definitely sth. that should be deprecated).

> The actual code that hit the regression have different mangled names:

AFAICT the check already mangles the function signature and checks for that, but as said: "revert first, talk (in-depth) later". Sounds fair?

timotheecour:
@wilzbach if we're not deprecating, can we at least print something (at least as option) when duplicate functions are seen by compiler? this helps catch bugs

jacob-carlborg:
> Isn't the regression caused by your special handling of extern(C) symbols? Therefore that special handling should just be removed outright.

Yeah, I think so. IIRC, Walter has also, in the `extern(C++)` discussions, been talking about that D should follow the same overloading rules for all symbols, i.e. no special treatment for `extern(C++)` symbols.

PR #7578, opened 2018-01-02
Bump OSX deployment target to 10.9

wilzbach:
This warning is getting annoying in the logs:

```
clang: warning: libstdc++ is deprecated; move to libc++ with a minimum deployment target of OS X 10.9 [-Wdeprecated]
```

Let's see what breaks.

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -82,7 +82,7 @@ G = $(GENERATED)/$(OS)/$(BUILD)/$(MODEL)
 $(shell mkdir -p $G)
 
 ifeq (osx,$(OS))
-    export MACOSX_DEPLOYMENT_TARGET=10.7
+    export MACOSX_DEPLOYMENT_TARGET=10.9
 endif
 
 HOST_CXX=c++


jacob-carlborg:
* Will this automatically switch to `libc++`?
* We should do some form of announcement of this since it's a breaking change

wilzbach:
> Will this automatically switch to libc++?

I think the warning was about an old `libc++`

> We should do some form of announcement of this since it's a breaking change

Well, this needs the approval of our release tzar @MartinNowak. I was just interested to see if any CI breaks, but apparently we're all good on the CI side :)

jacob-carlborg:
> I think the warning was about an old libc++

To me it looks like we're using `libstdc++` instead of `libc++`.

MartinNowak:
Unfortunately, release building is tied to a VM that's on 10.8.
I'll try to update this soon, but if it breaks the nightlies I'd rather revert it.

jacob-carlborg:
You would need 10.9 to run DMD, not sure about building. But I highly recommend upgrading to 10.13 as soon as possible.

MartinNowak:
Well, dmd is our app, so that means all our users would need at least 10.9 now.
Please add at least a changelog notice.
https://dlang.org/changelog/2.072.0.html#native_tls_osx

While it's generally good to stay up to date as a user, there are various reasons why it can be tricky, think of slow 3.5G internet in developing countries (or just slow DSL in various other parts). Also OSX is no longer available as offline medium. Granted those are vague arguments for tech savvy compiler users, but generally we shouldn't lightheartedly drop support. For OSX it's definitely a plus to finally move fully to clang/libstdc++. On FreeBSD it's also a pain that our releases are still build with GCC and require libstdc++.

> You would need 10.9 to run DMD, not sure about building.

I don't see any hard-coded deployment target/minos in dmd, so it'll use clang/CC's defaults.

MartinNowak:
Works, apparently the toolchain is new enough to be forward compatible.

```
Darwin-12:~ vagrant$ otool -L dmd.2.078.0
dmd.2.078.0:
	/usr/lib/libstdc++.6.dylib (compatibility version 7.0.0, current version 56.0.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 169.3.0)
Darwin-12:~ vagrant$ otool -L dmd         
dmd:
	/usr/lib/libc++.1.dylib (compatibility version 1.0.0, current version 65.1.0)
	/usr/lib/libSystem.B.dylib (compatibility version 1.0.0, current version 169.3.0)
Darwin-12:~ vagrant$
```

MartinNowak:
Might still want to add a brief changelog entry, instead of surprising people.

jacob-carlborg:
> Works, apparently the toolchain is new enough to be forward compatible.

@MartinNowak not sure what the above shows. Is `libc++` available on macOS 10.8?

jacob-carlborg:
> Might still want to add a brief changelog entry instead of surprising people.

Of course. Perhaps even a newsgroup post.

wilzbach:
> Might still want to add a brief changelog entry instead of surprising people.

https://github.com/dlang/dmd/pull/7643

PR #7580, opened 2018-01-02
highlight dclass.d error messages

WalterBright:


PetarKirov (approved):

ZombineDev:
BTW, it looks like some of those error messages are not covered by any tests, but that's a battle for another day.

PR #7583, opened 2018-01-02
Refactor: Move @disable check out of checkDeprecated

JinShil:
Revival of #6116 

This separates the check for `@disabled` from the check for `deprecated` to (hopefully) make the code clearer.

Courtesy of @mathias-lang-sociomantic 

@ibuclaw  C++ interface change

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -1931,8 +1931,7 @@ extern (C++) abstract class Expression : RootObject
 
     final void checkDisabled(Scope* sc, Dsymbol s)
     {
-        auto d = s.isDeclaration();
-        if (d)
+        if (auto d = s.isDeclaration())
             d.checkDisabled(loc, sc);
     }
 
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -202,7 +202,11 @@ Lagain:
     else
     {
         if (!s.isFuncDeclaration()) // functions are checked after overloading
+        {
             s.checkDeprecated(loc, sc);
+            if (d)
+                d.checkDisabled(loc, sc);
+        }
 
         // https://issues.dlang.org/show_bug.cgi?id=12023
         // if 's' is a tuple variable, the tuple is returned.
@@ -210,7 +214,11 @@ Lagain:
 
         //printf("s = '%s', s.kind = '%s', s.needThis() = %p\n", s.toChars(), s.kind(), s.needThis());
         if (s != olds && !s.isFuncDeclaration())
+        {
             s.checkDeprecated(loc, sc);
+            if (d)
+                d.checkDisabled(loc, sc);
+        }
     }
 
     if (auto em = s.isEnumMember())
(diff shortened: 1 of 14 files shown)


wilzbach (approved):
Nice. We get rid of three `goto`s :)

src/dmd/expressionsem.d
```diff
@@ -2221,6 +2221,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                     return setError();
 
                 exp.checkDeprecated(sc, f);
+                exp.checkDisabled(sc, f);
                 exp.checkPurity(sc, f);
                 exp.checkSafety(sc, f);
                 exp.checkNogc(sc, f);
```
wilzbach: Looking at the diff, it seems like these five function calls are a good candidate to group together in a `doChecks` function? (not required for this PR)

wilzbach: -> https://github.com/dlang/dmd/pull/7628

src/dmd/expression.d
```diff
@@ -1921,6 +1929,13 @@ extern (C++) abstract class Expression : RootObject
         s.checkDeprecated(loc, sc);
     }
 
+    final void checkDisabled(Scope* sc, Dsymbol s)
+    {
+        auto d = s.isDeclaration();
+        if (d)
```
skl131313: You can make this line a bit cleaner:

```D
if (auto d = s.isDeclaration())
    d.checkDisabled(loc, sc);
```

src/dmd/expression.d
```diff
@@ -202,15 +202,23 @@ Lagain:
     else
     {
         if (!s.isFuncDeclaration()) // functions are checked after overloading
+        {
             s.checkDeprecated(loc, sc);
+            if (d)
+                d.checkDisabled(loc, sc);
```
skl131313: Couldn't you do `s.checkDisabled(...)` instead, without having to check `d`?

src/dmd/mtype.d
```diff
@@ -1175,7 +1175,9 @@ extern (C++) abstract class Type : RootObject
     {
         Dsymbol s = toDsymbol(sc);
         if (s)
+        {
             s.checkDeprecated(loc, sc);
+        }
```
skl131313: Think the source prefs no brackets for single-line if statements. Could also move the `Dsymbol s = toDsymbol(sc)` into the if statement.

src/dmd/expression.d
JinShil: `checkDisabled` is a method of the `Declaration` class.  `s` is a `DSymbol` class.  Therefore, `s` must first be converted into `d` with the `isDeclaration` function.

skl131313: You could add the function into DSymbol, then override it in the Declaration class. It'd be cleaner then having all these if statements everywhere.

JinShil: That is correct, but it would depart from the patterns DMD is currently using.  grep the source code for `isDeclaration` and many other `isXXX` functions and you'll see.  Also, if I make such systematic changes to this PR, it will turn into a monstrosity and noone will want to review it.  Such a change could be a candidate for a different PR. though.

src/dmd/mtype.d
JinShil: Done.  Thank you.

src/dmd/expression.d
JinShil: Done.  Thank you.

PR #7584, opened 2018-01-02
 Fix Issue 16165 - Show expected number of function arguments on mismatch

ntrel:
This is ready now #7554 has been merged.

* Show expected number of arguments when too few/too many arguments are passed.
* Show expected number of variadic static array arguments on mismatch.

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -4567,15 +4567,14 @@ extern (C++) final class TypeFunction : TypeNext
 
         size_t nparams = Parameter.dim(parameters);
         size_t nargs = args ? args.dim : 0;
-        if (nparams == nargs)
-        {
-        }
-        else if (nargs > nparams)
+        if (nargs > nparams)
         {
             if (varargs == 0)
             {
-                if (pMessage) *pMessage = getMatchError("expected %d argument(s), not %d", nparams, nargs);
-                goto Nomatch;
+                // suppress early exit if an error message is wanted,
+                // so we can check any matching args are valid
+                if (!pMessage)
+                    goto Nomatch;
             }
             // too many args; no match
             match = MATCH.convert; // match ... with a "conversion" match level
@@ -4624,8 +4623,8 @@ extern (C++) final class TypeFunction : TypeNext
             {
                 if (p.defaultArg)
                     continue;
-                goto L1;
                 // try typesafe variadics
+                goto L1;
             }
             {
                 Expression arg = (*args)[u];
@@ -4814,6 +4813,12 @@ extern (C++) final class TypeFunction : TypeNext
         }
 
     Ldone:
+        if (pMessage && !varargs && nargs > nparams)
+        {
+            // all parameters had a match, but there are surplus args
+            *pMessage = getMatchError("expected %d argument(s), not %d", nparams, nargs);
+            goto Nomatch;
+        }
         //printf("match = %d\n", match);
         return match;
 
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -4744,8 +4744,15 @@ extern (C++) final class TypeFunction : TypeNext
                         sz = tsa.dim.toInteger();
                         if (sz != nargs - u)
                         {
-                            if (pMessage) *pMessage = getMatchError("expected %d variadic argument(s), not %d",
-                                sz, nargs - u);
+                            if (pMessage)
+                            {
+                                // Windows (Vista) OutBuffer.vprintf issue? 2nd argument always zero
+                                //*pMessage = getMatchError("expected %d variadic argument(s), not %d", sz, nargs - u);
+                                OutBuffer buf;
+                                buf.printf("expected %d variadic argument(s)", sz);
+                                buf.printf(", not %d", nargs - u);
+                                *pMessage = buf.extractString();
+                            }
                             goto Nomatch;
                         }
                         goto case Tarray;
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -4514,6 +4514,13 @@ extern (C++) final class TypeFunction : TypeNext
         return buf.extractString();
     }
 
+    private const(char)* getMatchError(A...)(const(char)* format, A args)
+    {
+        OutBuffer buf;
+        buf.printf(format, args);
+        return buf.extractString();
+    }
+
     /********************************
      * 'args' are being matched to function 'this'
      * Determine match level.
@@ -4566,7 +4573,10 @@ extern (C++) final class TypeFunction : TypeNext
         else if (nargs > nparams)
         {
             if (varargs == 0)
+            {
+                if (pMessage) *pMessage = getMatchError("expected %d argument(s), not %d", nparams, nargs);
                 goto Nomatch;
+            }
             // too many args; no match
             match = MATCH.convert; // match ... with a "conversion" match level
         }
@@ -4733,7 +4743,11 @@ extern (C++) final class TypeFunction : TypeNext
                         tsa = cast(TypeSArray)tb;
                         sz = tsa.dim.toInteger();
                         if (sz != nargs - u)
+                        {
+                            if (pMessage) *pMessage = getMatchError("expected %d variadic argument(s), not %d",
+                                sz, nargs - u);
                             goto Nomatch;
+                        }
                         goto case Tarray;
                     case Tarray:
                         {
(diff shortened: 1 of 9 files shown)


marler8997:
Thank you very much for doing this work.  I think this is VERY IMPORTANT.  That being said, I've thought about this and wanted to share my thoughts on how we could improve error messages in the best possible way.  I haven't looked at the implementation enough to know whether or not this would be easy, this is just based on what I'd like to see as a developer.

I think the first thing we should check is if there is a mismatch in argument count.  If so, then it likely means the developer either missed or added an extra argument.  In this case, I think the most helpful message would be:
```
Error: function `foo` takes 10 arguments but got 9
       argument 4 of type `string` is not convertible to `int`
```
Note that in the case there are no type mismatches, this likely means they just forgot the last argument, so the first line of the previous error message alone should suffice.  If there is more than 1 type mismatch, then it's probably better to only print the first because missing or adding an extra argument is likely to propogate "red herring" errors down the line.

If the argument count is correct, then we must have one or more type mismatches.  In the case of 1 type mismatch, a short error message like this should work well:
```
Error: expected argument 4 of function `foo` to be `int` but got `string`
```
If there is more than one type mismatch, then a multiline error message would probably be better:
```
Error: function `foo` was called with invalid arguments
       argument 4 of type `string` is not convertible to `int`
       argument 8 of type 'MyCoolType!(int,byte)` is not convertible to `AnotherType!(int,byte)`
```

In the case of overloads, just seeing the types is often too little information, however, you also don't want to print the full error message for each overload.  I think a good comprimise would be a small message for each overload that gives the most important reason for the mismatch.  If the argument count is incorrect, just print the count mismatch, otherwise, print the first type mismatch, i.e.
```
Error: none of the overloads of `foo` are callable using argument types `(int,string)`, candidates are:
       foo() requires 0 arguments but got 2
       foo(int) requires 1 argument but got 2
       foo(! short) argument 1 of `short` does not accept type `int`
       foo(int, ! int) argument 2 of type `int` does not accept type `string`
       foo(short, byte, short) requires 3 arguments but got 2
```
Also note that I included a ! in the location where a type mismatch occurs.

timotheecour:
> Note that type mismatches are indicated by a ! in the location where a type mismatch occurs.

and color-highlight mis-matches when color=on

marler8997:
I'm not sure if @ntrel is for or against the suggestions I've made.  I'll wait to see if/when this PR gets integrated and afterwards probably look into making those changes myself.

ntrel:
@marler8997: Here I'm avoiding changing the signature of `callMatch`, so I'm stuck with a single specific error message. I do think we should keep the printing of function parameters and provided argument types - this is useful.

Given that, I've added 2 commits to help the user pin-point the failing parameter match, rather than just saying the argument/parameter count.

For overloads, one option would be to print the specific error only when `dmd -v` is used - this keeps the number of output lines as is. Personally I'd be happy if they were always printed.

ntrel:
@wilzbach:
> are you aware of the CI failures? Though I don't know either what could have introduced them..
> Assertion failed: (t), function source_name, file cppmangle.c, line 185.

I don't know why that's happening.

wilzbach:
No idea about the auto-tester failures, looks like an issue with 2.068.2. If your OS doesn't use hardening, you can set the host compiler with `HOST_DMD`. Download with e.g.

```
curl https://dlang.org | bash -s dmd-2.068.2
```

CircleCi + Semaphore are known - I haven't had time to look into to that though :/

> test_results/d_do_test: symbol lookup error: test_results/d_do_test: undefined symbol: _D3std9exception__T7enforceZ__TQmTbZQrFNaNfbLAxaAyamZb
Makefile:209: recipe for target 'test_results/runnable/helloUTF16BE.d.out' failed

marler8997:
@ntrel  Ok it sounds like my suggestions would require a bigger set of changes.  Go ahead with this, any improvement in the current situation is good.  If I have time after this PR is merged I might look into this.

wilzbach:
Okay looks like we got over almost all failures, though the auto-tester is still complaining:

```
Assertion failed: (t), function source_name, file cppmangle.c, line 185.
gmake[1]: *** [../generated/freebsd/release/32/dmd_frontend] Abort trap: 6 (core dumped)
gmake[1]: *** Waiting for unfinished jobs....
Assertion failed: (t), function source_name, file cppmangle.c, line 185.
Assertion failed: (t), function source_name, file cppmangle.c, line 185.
```

RazvanN7:
@ntrel do you plan on pursuing this? I think it is very important that this gets merged eventually, so will you take it to completion or should someone adopt it?

wilzbach:
You could please do a rebase instead?
We try to keep the git history as short as possible.

jacob-carlborg:
Are all these commits necessary, would it be better to squash them instead?

ntrel:
@jacob-carlborg f286ceb and 3437c00 are cherry-picked from `stable` (they're necessary to reduce unnecessary memory usage that was probably causing a timeout in a test), probably they shouldn't be squashed. The existing ones I thought I'd leave to show I've not changed anything, but could be squashed.

jacob-carlborg:
ok

wilzbach:
FYI the SemaphoreCI failure is related to GDC having issues when building this:

```
dmd/mtype.d: In member function 'callMatch':
dmd/mtype.d:5116:21: internal compiler error: Aborted
                     *pMessage = getMatchError("missing argument for parameter #%d: `%s`",
```

ibuclaw:
@wilzbach you forgot the ice line in the compiler itself.

```
cppmangle.c:185: void CppMangleVisitor::source_name(Dsymbol*, bool): Assertion `t' failed.
dmd/mtype.d: In member function 'callMatch':
dmd/mtype.d:4953:21: internal compiler error: Aborted
                     *pMessage = getMatchError("missing argument for parameter #%d: `%s`",
                     ^
```

ibuclaw:
The frontend doesn't support mangling that sort of template in C++. As it's private anyway, you should explicitly mark it as `extern(D)`

src/dmd/mtype.d
```diff
@@ -4656,6 +4656,15 @@ extern (C++) final class TypeFunction : TypeNext
         return buf.extractString();
     }
 
+    private const(char)* getMatchError(A...)(const(char)* format, A args)
```
ibuclaw: The frontend doesn't support mangling that sort of template in C++. As it's private anyway, you should explicitly mark it as extern(D)

ibuclaw (dismissed):

test/fail_compilation/fail332.d
```diff
@@ -1,15 +1,56 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/fail332.d(14): Error: function `fail332.foo(int _param_0, ...)` is not callable using argument types `()`
+fail_compilation/fail332.d(22): Error: function `fail332.foo(int _param_0, ...)` is not callable using argument types `()`
+fail_compilation/fail332.d(22):        missing argument for parameter #1: `int _param_0`
```
jacob-carlborg: Why is `_param_0` mentioned sometimes and sometimes it's not, i.e.: https://github.com/dlang/dmd/pull/7584/files#diff-534158a8eaceb0162bf8ef17709be089R33

test/fail_compilation/diag8101.d
```diff
@@ -1,29 +1,30 @@
...
-fail_compilation/diag8101.d(45):        `diag8101.t_1(T1, T2)()`
-fail_compilation/diag8101.d(62): Error: template `diag8101.t_2` cannot deduce function from argument types `!()()`, candidates are:
-fail_compilation/diag8101.d(47):        `diag8101.t_2(T1)()`
-fail_compilation/diag8101.d(48):        `diag8101.t_2(T1, T2)()`
-fail_compilation/diag8101.d(49):        `diag8101.t_2(T1, T2, T3)()`
-fail_compilation/diag8101.d(50):        `diag8101.t_2(T1, T2, T3, T4)()`
-fail_compilation/diag8101.d(51):        `diag8101.t_2(T1, T2, T3, T4, T5)()`
-fail_compilation/diag8101.d(62):        ... (1 more, -v to show) ...
+fail_compilation/diag8101.d(57): Error: function `diag8101.f_0(int)` is not callable using argument types `()`
+fail_compilation/diag8101.d(57):        missing argument for parameter #1: `int`
```
jacob-carlborg: How does this error message look like with several missing arguments?

thewilsonator:
@ntrel this is green. Anything more you want to add? Otherwise please squash.

ntrel:
@thewilsonator That's great, I think this is good to go. I don't have my laptop where I am for the next few days, so it might take me a while to squash commits (unless there's a way to do it in a browser?).

ntrel:
@thewilsonator I've now squashed 4 simple commits into one.

thewilsonator (approved):

PR #7586, opened 2018-01-02
highlight typesem.d error messages

WalterBright:


src/dmd/typesem.d
```diff
@@ -205,7 +207,7 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
         mtype.basetype = mtype.basetype.toBasetype().mutableOf();
         if (mtype.basetype.ty != Tsarray)
         {
-            mtype.error(loc, "T in __vector(T) must be a static array, not %s", mtype.basetype.toChars());
+            mtype.error(loc, "T in __vector(T) must be a static array, not `%s`", mtype.basetype.toChars());
```
JinShil: Shouldn't `__vector(T)` be surrounded by tick marks also?

src/dmd/typesem.d
```diff
@@ -906,7 +908,7 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
                 {
                     if (ubyte m = fparam.type.mod & (MODimmutable | MODconst | MODwild))
                     {
-                        mtype.error(loc, "cannot have %s out parameter of type %s", MODtoChars(m), t.toChars());
+                        mtype.error(loc, "cannot have %s out parameter of type `%s`", MODtoChars(m), t.toChars());
```
JinShil: I think "%s out" should also be surrounded by tick marks:  "cannot have \`%s out\` parameter of type \`%s\`"

src/dmd/typesem.d
```diff
@@ -1062,7 +1064,7 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
 
         if (wildreturn && !wildparams)
         {
-            mtype.error(loc, "inout on return means inout must be on a parameter as well for %s", mtype.toChars());
+            mtype.error(loc, "inout on return means inout must be on a parameter as well for `%s`", mtype.toChars());
```
JinShil: Tick marks around "inout" and "return"

src/dmd/typesem.d
```diff
@@ -795,15 +797,15 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
             errors |= tf.checkRetType(loc);
             if (tf.next.isscope() && !(sc.flags & SCOPEctor))
             {
-                mtype.error(loc, "functions cannot return scope %s", tf.next.toChars());
+                mtype.error(loc, "functions cannot return scope `%s`", tf.next.toChars());
```
JinShil: Should the tick marks surround "return scope" also? "functions cannot \`return scope %s\`".  I'm not sure.

src/dmd/typesem.d
WalterBright: There are many such decisions, and they don't always have clear answers. In this case, `T` is not a user defined type, it never appears in his source code.

I opted instead to make these changes a bulk "first approximation", taking the most straightforward approach in order to get the bulk of it done.

JinShil (dismissed):

src/dmd/typesem.d
JinShil: Ok, I understand.

src/dmd/typesem.d
WalterBright: I opted out of `MODtoChars` results being highlighted for the first approximation. There are thousands more error messages to go.

src/dmd/typesem.d
WalterBright: Again, this is just a first approximation. There are thousands of messages to fix. The last time I tried to do this, the PRs all got derailed into debates about such things, and the result was zero progress for the following 6 months.

wilzbach: > There are thousands of messages to fix. 

Are you sure you don't want to automate this?

JinShil (approved):

src/dmd/typesem.d
WalterBright: Others have said they were going to automate it. I'm not waiting any longer.

PR #7593, opened 2018-01-04
fix Issue 18190 - [asan] heap-buffer-overflow in Module.load.checkMod…

WalterBright:
…FileAlias

I believe the problem is that `memcmp` is allowed to read `nbytes` of data, even past the byte where a mismatch occurs. The code assumed it did not read past the mismatch.

### DIFF
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -6,11 +6,11 @@
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
  * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/dmodule.d, _dmodule.d)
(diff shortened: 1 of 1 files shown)


JinShil (approved):

JohanEngelen:
> I believe the problem is that memcmp is allowed to read nbytes of data, even past the byte where a mismatch occurs. The code assumed it did not read past the mismatch.

Indeed it appears so: https://trust-in-soft.com/memcmp-requires-pointers-to-fully-valid-buffers/
This is also a problem then in the cmdline args matching code in LDC's LDMD and in mars.d.

WalterBright:
@JohanEngelen thanks for doing the research to confirm. I had assumed forever that it wouldn't read past the mismatch, but when I read the memcmp description last night, it didn't actually say that. We should always assume the worst when things are not explicitly stated.

This also means searching the source code for `memcmp` and double checking them.

JohanEngelen:
And @kinke has found the trivial replacement for the string checking memcmp: `strncmp`.
https://stackoverflow.com/questions/38878195/does-this-usage-of-strncmp-contain-an-out-of-bounds-read

PR #7594, opened 2018-01-04
highlight dsymbolsem.d error messages part 2

WalterBright:
Another batch of 20 messages


wilzbach (approved):
Half of these message aren't even covered :scream:

WalterBright:
I know. Much of it predates having coverage testing.

PR #7596, opened 2018-01-04
[CodeCov] Don't wait for other CIs before sending the notifications

wilzbach:
Huge thanks to the awesome guys from CodeCov (@stevepeak).


JinShil:
Will the change be apparent in this PR?

wilzbach:
> Will the change be apparent in this PR?

Nope, but I got the advice from @stevepeak so I think it's reasonable to give it a try.

JinShil (approved):

PR #7599, opened 2018-01-04
avoid LOOP instruction when passing fat struct arguments

MartinNowak:
- LOOP is extremely slow on Intel and older AMD CPUs
- partly addresses Issue 18114 (4s -> 3.3s)
  
I've tried to allocate a register for the counter instead of hard-coding CX, but couldn't make enough sense of allocreg, getregs_imm, movregconst et.al.. Feel free to do that yourself if it matters enough.

How to self-update the asm test is mentioned here.
https://github.com/dlang/dmd/pull/7599/files#diff-fb66f095cf343c50d9e46185589f04f3R9

### DIFF
--- a/src/dmd/backend/cod1.c
+++ b/src/dmd/backend/cod1.c
@@ -4035,15 +4035,24 @@ void pushParams(CodeBuilder& cdb,elem *e,unsigned stackalign)
                     if (!doneoff)
                     {   // This should be done when
                         // reg is loaded. Fix later
-                                                        // ADD reg,sz-2
+                                                        // ADD reg,sz-pushsize
                         cdb.genc2(0x81,grex | modregrmx(3,0,reg),sz-pushsize);
                     }
                     getregs(cdb,mCX);                       // the LOOP decrements it
-                    cdb.gen2(0xFF,buildModregrm(0,6,rm));           // PUSH [reg]
+                    cdb.gen2(0xFF,buildModregrm(0,6,rm));   // PUSH [reg]
                     cdb.last()->Iflags |= seg | CFtarg2;
                     code *c3 = cdb.last();
-                    cdb.genc2(0x81,grex | buildModregrm(3,5,reg),pushsize);  // SUB reg,2
-                    genjmp(cdb,0xE2,FLcode,(block *)c3);        // LOOP c3
+                    cdb.genc2(0x81,grex | buildModregrm(3,5,reg),pushsize);  // SUB reg,pushsize
+                    if (I16 || config.flags4 & CFG4space)
+                        genjmp(cdb,0xE2,FLcode,(block *)c3);// LOOP c3
+                    else
+                    {
+                        if (I64)
+                            cdb.gen2(0xFF,modregrm(3,1,CX));// DEC CX
+                        else
+                            cdb.gen1(0x48 + CX);            // DEC CX
+                        genjmp(cdb,JNE,FLcode,(block *)c3); // JNE c3
+                    }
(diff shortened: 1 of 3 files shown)


src/dmd/backend/cod1.c
```diff
@@ -4035,15 +4035,19 @@ void pushParams(CodeBuilder& cdb,elem *e,unsigned stackalign)
...
                     }
                     getregs(cdb,mCX);                       // the LOOP decrements it
-                    cdb.gen2(0xFF,buildModregrm(0,6,rm));           // PUSH [reg]
+                    cdb.gen2(0xFF,buildModregrm(0,6,rm));   // PUSH [reg]
                     cdb.last()->Iflags |= seg | CFtarg2;
                     code *c3 = cdb.last();
-                    cdb.genc2(0x81,grex | buildModregrm(3,5,reg),pushsize);  // SUB reg,2
-                    genjmp(cdb,0xE2,FLcode,(block *)c3);        // LOOP c3
+                    cdb.genc2(0x81,grex | buildModregrm(3,5,reg),pushsize);  // SUB reg,pushsize
+                    if (I64)
```
WalterBright: Prepend to line 4046:
```
if (I16 || config.flags4 & CFG4space)
    genjmp(cdb,0xE2,FLcode,(block *)c3);        // LOOP c3
else
```

test/runnable/test_cdstrpar.d
```diff
@@ -0,0 +1,266 @@
...
+                if (tc.code[i] != code[i])
+                    fprintf(stderr, " \033[31m0x%02x\033[0m", d);
+                else
+                    fprintf(stderr, " 0x%02x", d);
+            }
+            fprintf(stderr, "\n");
+            failure = true;
+        }
+        assert(!failure);
+    }
```
wilzbach: There's a lot of code duplication from `test_cdvecfill.d`. Any reason why this can't be shared?

MartinNowak: It's yet unclear how this could be made into a general framework to test codegen, or whether better tools already exist for that.
I've tried to keep them very similar to facilitate future generalization, but for the time being those are ad-hoc solutions and I wouldn't want to waste more time at generalizing sth. that's only used twice.

jacob-carlborg: Perhaps look into how LLVM/GCC are testing this.

WalterBright:
@MartinNowak  There are a couple other cases where LOOPs are generated, if you're interested!

MartinNowak:
> if you're interested!

I'm not ;), fixing even the smallest backend thing takes ages. This was an exception because it has such a massive impact.

WalterBright:
For reference: https://issues.dlang.org/show_bug.cgi?id=18114

Thanks for identifying the LOOP as a problem. I did not realize that.

MartinNowak:
I didn't know either, but [perf](https://perf.wiki.kernel.org/index.php/Main_Page) did ;).
Some trivia, TL;DR too complex instruction for early out-of-order CPUs, so people stopped using it and hence got no love in later HW optimizations.
[performance - Why is the loop instruction slow? Couldn't Intel have implemented it efficiently? - Stack Overflow](https://stackoverflow.com/questions/35742570/why-is-the-loop-instruction-slow-couldnt-intel-have-implemented-it-efficiently)
[x86: Why LOOPxx instruction is so slow?–Google Groups](https://groups.google.com/forum/#!msg/comp.arch/5RN6EegUxE0/mh6lLRc9_4MJ)

PR #7603, opened 2018-01-05
mars.d: replace memcmp() with startsWith()

WalterBright:
The `memcmp()` calls were not memory safe. See https://github.com/dlang/dmd/pull/7593 for explanation.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1532,6 +1532,24 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
         return (overflow || value > max || *p) ? uint.max : value;
     }
 
+    /********************************
+     * Params:
+     *  p = 0 terminated string
+     *  s = string
+     * Returns:
+     *  true if `p` starts with `s`
+     */
+    static pure bool startsWith(const(char)* p, string s)
+    {
+        foreach (const c; s)
+        {
+            if (c != *p)
+                return false;
+            ++p;
+        }
+        return true;
+    }
+
     version (none)
     {
         for (size_t i = 0; i < arguments.dim; i++)
@@ -1554,7 +1572,7 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 params.useDeprecated = 2;
             else if (strcmp(p + 1, "c") == 0)  // https://dlang.org/dmd.html#switch-c
                 params.link = false;
-            else if (memcmp(p + 1, cast(char*)"color", 5) == 0) // https://dlang.org/dmd.html#switch-color
+            else if (startsWith(p + 1, "color")) // https://dlang.org/dmd.html#switch-color
             {
                 params.color = true;
                 // Parse:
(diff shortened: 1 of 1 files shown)


src/dmd/mars.d
```diff
@@ -1554,7 +1572,7 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 params.useDeprecated = 2;
             else if (strcmp(p + 1, "c") == 0)  // https://dlang.org/dmd.html#switch-c
                 params.link = false;
-            else if (memcmp(p + 1, cast(char*)"color", 5) == 0) // https://dlang.org/dmd.html#switch-color
+            else if (startsWith(p + 1, "color") == 0) // https://dlang.org/dmd.html#switch-color
```
JinShil: If you're returning a `bool`, then you should be able to use `if (startsWith(p + 1, "color"))` without the comparison to `0`.

JinShil (changes requested):

src/dmd/mars.d
WalterBright: yes, I know :-)

JinShil (approved):
Yeah!  The fewer dependencies on C std library the better, IMO.

src/dmd/mars.d
```diff
@@ -1532,6 +1532,24 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
...
+    static pure bool startsWith(const(char)* p, string s)
+    {
+        foreach (const c; s)
+        {
+            if (c != *p)
+                return false;
+            ++p;
+        }
+        return true;
+    }
```
jacob-carlborg: I'm note sure if this is a problem, but since you're looping through `s` and not `p` and if `s` is longer than `p` and ends with `\0` it would continue the loop beyond `p`. Example:

```d
import std.stdio : writeln;

static bool startsWith(const(char)* p, string s)
{
    writeln(s.length);
    foreach (i, const c; s)
    {
        writeln(*p);
        if (c != *p)
            return false;
        ++p;
    }
    return true;
}

void main()
{
    auto r = startsWith("foo", "foo\0\0\0\0\0\0\0\0\0\0\0\0\0\0");
    writeln(r);
}
```

The above code prints the following on my machine:

```
17
f
o
o













`
false
```

WalterBright: It's not an issue because none of the strings passed to `s` are user-defined, they're all string literals.

JinShil: @WalterBright  When can we stop using `char*` for strings and start using D's string facilities?

jacob-carlborg: It still feels wrong to have a broken function. What if someone copies it and starts to use it somewhere else with user-defined strings.

WalterBright:
> When can we stop using char* for strings and start using D's string facilities?

There are the C++ back ends which only use char*. Internally to the front end, I've already moved a lot to D strings, but certainly more can be done.

WalterBright:
> What if someone copies it and starts to use it somewhere else with user-defined strings.

It's a nested function, specific to the nestee.

jacob-carlborg:
> It's a nested function, specific to the nestee.

That's how it's used now. But since it's `static` and `pure` it doesn't depend on its surrounding context and can easily be copied/moved somewhere else.

PR #7607, opened 2018-01-05
[Trivial] add func attribs to class Declaration final helpers

ghost:
The same was asked in #7569 so there no reason not to add the attribs to the other helpers.

### DIFF
--- a/src/dmd/declaration.d
+++ b/src/dmd/declaration.d
@@ -337,7 +337,7 @@ extern (C++) abstract class Declaration : Dsymbol
         return s;
     }
 
-    final bool isStatic()
+    final bool isStatic() const pure nothrow @nogc @safe
     {
         return (storage_class & STCstatic) != 0;
     }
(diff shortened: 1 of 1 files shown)


PetarKirov (approved):
LGTM

JinShil:
Does the `const` on these affect the C++ interface?

BBasile:
@ibuclaw, about the c++ interface ?

ZombineDev:
I missed that, because I thought the auto-tester will check the C++ headers, courtesy of https://github.com/dlang/dmd/pull/7520. @ibuclaw how is that we're successfully linking (as seen in https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=2980879&dataid=21094560&isPull=true):
```
c++ -c -o../generated/linux/release/64/cxxfrontend.o -Wno-deprecated -Wstrict-aliasing -Werror -fno-exceptions -fno-rtti -D__pascal= -DMARS=1 -DTARGET_LINUX=1 -DDM_TARGET_CPU_X86=1 -m64 -fPIC -std=gnu++98 -O2 -Idmd -Idmd/root -Wuninitialized -MMD -MF ../generated/linux/release/64/cxxfrontend.deps tests/cxxfrontend.c
```
)
while having a CV-qualifier mismatch?

ZombineDev:
Actually, since the functions are `final` (i.e. non-virtual) they are not part of the vtable, hence why we have implementations of those functions both in C++ and D. @BBasile you don't need to update anything on the C++ side, sorry for the confusion.

BBasile:
NP. The other PR is here in case of.

ibuclaw:
You need to use the function in c++ in order to see the mismatch.

PR #7609, opened 2018-01-05
fixup #7607 - set const the helpers in the cpp header too

ghost:


ZombineDev:
Since those methods are non-virtual you don't strictly need to update the C++ headers, but it's a nice improvement for the C++ header clients nonetheless.

PetarKirov (approved):

PetarKirov (approved):

PR #7595, opened 2018-01-04
Update dmd-cxx glue layer minimally to get dmd building with the 2.071 backend

joakim-noah:
Minimal fixes to get it building with the 2.071 backend, plus updated Travis to build the 2.076.1 stdlib.

PR #7610, opened 2018-01-05
Disable AppVeyor for dmd-cxx

wilzbach:


MartinNowak:
Let's try blacklisting instead
https://www.appveyor.com/docs/branches/

wilzbach:
I tried that before, but maybe it didn't take immediate effect:

image

See also: https://github.com/dlang/dmd/pull/7595

MartinNowak:
And while we're at it, let's remove the dummy Jenkinsfile as well.

MartinNowak:
> I tried that before, but maybe it didn't take immediate effect:

It's puzzling me how codecov and appveyor came up with the weird idea to configure in the repo and on their servers. Also why would it test anything if there is no appveyor config? Exactly because of this server-side default config nonsense, requiring us now to override it with a blacklist dummy :o.

MartinNowak:
Not to mention AppVeyors weird login logic, where I have to select your account b/c they haven't thought of orgs beforehand.
screenshot-2018-1-5 login - appveyor

jacob-carlborg:
@MartinNowak @wilzbach it's possible to configure AppVeyor to not build if there's no `appveyor.yml`. It's on the General settings page at further down, called `Skip branches without appveyor.yml`.

screen shot 2018-01-05 at 17 28 11

PR #7613, opened 2018-01-06
highlight opover.d error messages

JinShil:


src/dmd/opover.d
```diff
@@ -1992,7 +1992,7 @@ private Dsymbol inferApplyArgTypesX(Expression ethis, FuncDeclaration fstart, Pa
         inferApplyArgTypesY(cast(TypeFunction)fd_best.type, parameters);
         if (fd_ambig)
         {
-            .error(ethis.loc, "%s.%s matches more than one declaration:\n%s:     %s\nand:\n%s:     %s",
+            .error(ethis.loc, "`%s.%s` matches more than one declaration:\n`%s`:     `%s`\nand:\n`%s`:     `%s`",
```
ibuclaw: Shouldn't errorSupplemental be used instead of embedded new lines in error strings?

I really dislike using loc.toChars in a message, as it bypasses the diagnostic infrastructure[1] and all the bells and whistles that offers on top of the stock frontend errors.

[1]  gdc hooks into gcc, so you get source file printing, caret for column start, and maybe in future fixits for spelling mistakes and location ranges for underlining entire expressions.

ibuclaw (approved):
We really need to add more fail compilation tests for error coverage.

src/dmd/opover.d
WalterBright: There are many things wrong with the error messages; for these PRs I asked @JinShil to focus on just the highlighting.

ibuclaw: All I'm doing is reminding you again that I still disagree with certain errors (I have raised this before :-)

wilzbach (approved):
@JinShil and I just fixed Jenkins in https://github.com/dlang/ci/pulls/111 and I restarted the build here... hopefully that works.

PR #7559, opened 2017-12-31
Move loadModule and paintAsType from Target to Compiler

ibuclaw:
These are a couple of funny little functions that are compiler-specific, but are not directly related to the target platform.

PR #7615, opened 2018-01-06
Remove import of dmd.compiler from dmd.globals.

ibuclaw:
This is blocking #7534 and #7559, as the D implementation is less encapsulated than C++ version.

### DIFF
--- a/src/dmd/compiler.d
+++ b/src/dmd/compiler.d
@@ -12,7 +12,12 @@
 
 module dmd.compiler;
 
+import dmd.globals;
+
 struct Compiler
 {
-    const(char)* vendor; // Compiler backend name
+    extern(C++) static void _init()
+    {
+        global.vendor = "Digital Mars D";
+    }
 }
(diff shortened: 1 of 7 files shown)


src/dmd/compiler.d
```diff
@@ -12,7 +12,12 @@
...
+import dmd.globals;
+
 struct Compiler
 {
-    const(char)* vendor; // Compiler backend name
+    extern(C++) static void _init()
+    {
+        global.vendor = "Digital Mars D";
+    }
 }
```
wilzbach: As discussed on IRC: how about making `Compiler` an interface instead?

ibuclaw: I've given it some thought and I'd rather avoid this if possible.  Having to maintain interfaces seems like a last resort solution.   For sure though `Global._init` is a _big_ problem to tackle, as a big chunk of it should go into `Target`, but the globals module cannot import target!

I'll start off with adding `initGlobals` and see where we go from there.

PR #7619, opened 2018-01-07
Remove IN_GCC condition for catch in finally blocks.

ibuclaw:
This is a nonsense error.  It's only because of a dumb limitation in DMD that it even exists.

Infact, now that DMD supports DwarfEH, this error should not even apply anymore on those platforms which support.

But just so that all compilers are on the same page...


ibuclaw:
Rebased to fix Jenkins CI.

dnadlinger (approved):

PR #7620, opened 2018-01-07
fix Issue 8207 - extern(D) symbols should include another underscore

ibuclaw:
And put all static vars into ABI "sections" in the target struct whilst I was modifying it.

The extra underscore comes from platforms who symbols visible to C are prefixed with `_`, such as OSX and Win32.  Having double underscores means that DMD now matches ABI of GDC and LDC on those platforms, and removes "special" logic that removed or added the extra `_` in the frontend.

Just to document the list of changes made before getting to the final point:
1. Added a new hook `Target.cManglePrefix` for setting the backend symbol prefix.  Removing some special logic in mtype.d and the `static if (!IN_GCC)` condition.
Builds succeed, but on second thought, it just wasn't pretty.  It was decided to instead make DMD compatible with GDC and LDC, both of which don't have special branching.

2. Removed `Target.cManglePrefix` and ABI hack completely
Win32 returns `Symbol Undefined __D10TypeInfo_i6__initZ` (probably OSX too, but autotester stops processing PRs on first error, Travis has been busy for the last 6 hours building other things).

3. Remove `Target.prefixName` and C++ tests for matching `__Z`
Same result (what do you have to do in order to get a Darwin build round here)

4. Prepend prefixes for C++ and D symbols in the backend only
OSX returns:
```
  "_D9invariant12_d_invariantFC6ObjectZv", referenced from:
      __D8foreach59test6659bFZv in foreach5_0.o
      __D8foreach59test6659cFZv in foreach5_0.o
```

`_d_invariant` is a library function that's `extern(D)`, so I can only assume that `extern(C)` functions are prefixed correctly.

5. Remove `RTLSYM__DINVARIANT`, prefixing underscore should always be present.
Discovered that many C++ symbols were being marked as `extern(D)` in the codegen.  I assume this is the work of Daniel and Igor when they ported the MSVC++ mangler to the front-end.

6. Correctly set C++ mangling on `LINKcpp` symbols.
Darwin x86 fails to build phobos because of missing references to `___tls_get_addr`.

7. Added workaround for `extern(D) ___tls_get_addr`, setting the exact symbol name that should be written to object file.  Raised a PR in druntime to switch linkage to `extern(C)`.


ibuclaw:
Just doing a quick grep, it looks like all uses of `global.params.isXXX` are all contained to either the target module or the glue layer, except for a couple uses of `isWindows`.

Could probably start by moving `TARGET_XXX` enums to the target module, and replace all uses of `global.params.isXXX` with them.

src/dmd/mtype.d
```diff
@@ -2805,18 +2805,15 @@ extern (C++) abstract class Type : RootObject
         auto name = namelen <= namebuf.length ? namebuf.ptr : cast(char*)malloc(namelen);
         assert(name);
 
-        const length = sprintf(name, "_D%lluTypeInfo_%.*s6__initZ",
-                cast(ulong)(9 + slice.length), cast(int)slice.length, slice.ptr);
+        // If C mangling will add the prefix in, don't add our own.
+        const (char)* prefix = Target.cManglePrefix[0] ? "" : "_";
```
wilzbach: Why not go the full chain and add `Target.customCManglePrefix()`?

wilzbach (approved):
I like your direction :)

src/dmd/mtype.d
ibuclaw: I have no idea where DMD does mangling in the backend, I'm not sure whether there's scope to re-use this field in other parts of the compiler also, maybe even expose it to the user code somehow (traits, pragma, mangleof?).

In GCC and LLVM internals expose this as a `__USER_LABEL_PREFIX__` [macro](https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html).

ibuclaw: I'm not aware of any target defining this as anything other than `"_"` or `""`.  So it could be a `char` instead.  It's a string here though just to be compatible with GCC and LLVM.

ibuclaw:
Rebased to fix failing Jenkins CI.

src/dmd/mtype.d
klickverbot: You are confusing Clang and LLVM here, but either way, this change doesn't really work for LDC. We always keep the extra underscore on OS X, to be compatible with what GCC/Clang do (and what GDB expects).

dnadlinger (dismissed):

src/dmd/mtype.d
ibuclaw: Right, and so does GDC for the exact same reasons.  If you have any better ideas that don't involve changing the dmd backend, then send them my way.

ibuclaw: Such as, removing this ABI hack altogether.

ibuclaw: Also, doesn't LLVM expose a baseline ABI?  Not that I'm making accusations over what they consider best practices (you may notice that I get confused by a lot of things that LLVM does).

ibuclaw:
[testing, comment will be updated]
1. Removed `Target.cManglePrefix` ABI hack completely -> Waiting result.

ibuclaw:
Looks like tests are passing, so will rebase.

test/compilable/cppmangle.d
```diff
@@ -364,13 +364,7 @@ extern(C++)
     }
 }
 
-version (OSX)
-{
-    static assert(T.foo.mangleof == "__ZNK1T3fooEi");
-    static assert(T.bar.mangleof == "__ZN1T3barEi");
-    static assert(T.boo.mangleof == "__ZN1T3booE");
-}
```
wilzbach: What's the motivation for removing the OSX test completely?

ibuclaw: Matches gdc and ldc behaviour.

wilzbach: Ah sorry. I just saw the removed test, not that it will still be executed.

WalterBright (changes requested):
Fix failing Win32 test

ibuclaw:
Only done a spot check, but seems like some C++ symbols were being marked as the equivalent of `LINKd` in the backend.

ibuclaw:
And now it is mangling it twice. :-)

src/dmd/target.d
```diff
@@ -427,23 +432,6 @@ struct Target
     {
     }
 
-    /******************************
-     * For the given symbol written to the OutBuffer, apply any
-     * target-specific prefixes based on the given linkage.
-     */
-    extern (C++) static void prefixName(OutBuffer* buf, LINK linkage)
```
klickverbot: +1, unused in LDC as well.

ibuclaw:
Why is there a special case for `__tls_getaddr`?

This means that druntime needs changing in lockstep.

ibuclaw:
Maybe have a way to provide a clean transition from `extern(D)` to `extern(C)`.

ibuclaw:
@klickverbot / @wilzbach / @WalterBright - I think it should be passing all tests on all platforms now.

wilzbach:
> +1 for the mangling change, but note that this (obviously) breaks the ABI on macOS in a spectacular fashion (everything except for extern(C)).

This probably needs a changelog entry. See for example:

https://dlang.org/changelog/2.072.0.html#native_tls_osx

ibuclaw:
Removing easy review, as I guess the implications are anything but.

ibuclaw:
@WalterBright - input would be nice once you're feeling better.

ibuclaw:
@WalterBright - any chance of a review?

ibuclaw:
Rebased again... moved the removal on `enum IN_GCC` to this PR, so CI won't succeed until #7621 is merged.

timotheecour:
on OSX:
* does that mean using lldb or gdb will show properly demangled symbols ?
* will  `echo symbol | ddemangle` start working without having to add an _?

jacob-carlborg:
> does that mean using lldb or gdb will show properly demangled symbols ?

@timotheecour LLDB cannot demangle D symbols. GDB should be able to do that if you have a recent enough version. But Apple has replaced GDB with LLDB. So you would need to install that using Homebrew or MacPorts.

ibuclaw:
> does that mean using lldb or gdb will show properly demangled symbols ?

I'm pretty sure it is mentioned in the issue. On OSX, gdb cannot demangle dmd-generated D symbols because there is a missing prefixing underscore.

timotheecour:
> I'm pretty sure it is mentioned in the issue

well I'm not seeing gdb mentioned in the top-level commit msg https://github.com/dlang/dmd/pull/7620#issue-286533771 so was just making sure I understood correctly. maybe it'd be worth mentioning gdb specifically since it's a major use case of that PR

ibuclaw:
Rebased and unblocked.

WalterBright (approved):
Nice work!

ibuclaw:
@WalterBright - Excellent!  Now let's remove this `___tls_get_addr` special case from both druntime and the compiler. ;-)

https://github.com/dlang/druntime/pull/2034

PR #7627, opened 2018-01-07
Fixup C++ headers (based on v2.078.0)

kinke:
E.g., reflecting the D `Visitor` class hierarchy with base class `ParseTimeVisitor` also in C++ is crucial for Visual C++.

### DIFF
--- a/src/dmd/visitor.h
+++ b/src/dmd/visitor.h
@@ -297,7 +297,7 @@ class StaticIfCondition;
 
 class Parameter;
 
-class Visitor
+class ParseTimeVisitor
 {
 public:
     virtual void visit(Dsymbol *) { assert(0); }
@@ -557,6 +557,12 @@ public:
     virtual void visit(StructInitializer *i) { visit((Initializer *)i); }
     virtual void visit(ArrayInitializer *i) { visit((Initializer *)i); }
     virtual void visit(VoidInitializer *i) { visit((Initializer *)i); }
+};
+
+class Visitor : public ParseTimeVisitor
+{
+public:
+    using ParseTimeVisitor::visit;
 
     // Miscellaneous
     virtual void visit(ErrorStatement *s) { visit((Statement *)s); }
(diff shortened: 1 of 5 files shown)


src/dmd/visitor.h
```diff
@@ -557,6 +557,12 @@ class Visitor
     virtual void visit(StructInitializer *i) { visit((Initializer *)i); }
     virtual void visit(ArrayInitializer *i) { visit((Initializer *)i); }
     virtual void visit(VoidInitializer *i) { visit((Initializer *)i); }
+};
+
+class Visitor : public ParseTimeVisitor
+{
+public:
+    using ParseTimeVisitor::visit;
```
ibuclaw: Is this really necessary?

kinke: Yep, at least for C++11+ compilers, otherwise there are warnings as the base class overloads are hidden.

ibuclaw:
Also  #7334

src/dmd/visitor.h
ibuclaw: Hmm, that's what I'm worried about. Breaking C++98

WalterBright: Eh, just switch to using D already :+1:

ibuclaw: This visitor class is used by all codegen routines, it can't be in any other language. Breaking compatibility is not an option. I think it bad enough that this class was rewritten for the sake of trying to make the frontend a better library. Had I not spotted, it would be @kinke having to spend a day trying to work out why nothing works at runtime, then doing 300 lines worth of rewriting the Visitor classes.

Sometimes I think it's forgotten that there are two major projects written in C++ that have been using DMD as if it was a library for over 14 years now.

kinke: Yeah thanks a lot Iain for that, I know how tedious figuring out header divergences are.

Wrt. the `using`, I don't know how to express it in pre-C++11 or whether it's even required.

ibuclaw: It should be fine, as the testsuite compiles with `-std=c++98`.  I'm just being overly cautious here, and I feel I at least have good reason to be.

PR #7628, opened 2018-01-08
Move common function attributes check into a function

wilzbach:
From: https://github.com/dlang/dmd/pull/7583/files#r160068084

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -2221,11 +2221,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 if (!f || f.errors)
                     return setError();
 
-                exp.checkDeprecated(sc, f);
-                exp.checkDisabled(sc, f);
-                exp.checkPurity(sc, f);
-                exp.checkSafety(sc, f);
-                exp.checkNogc(sc, f);
+                checkFunctionAttributes(exp, sc, f);
                 checkAccess(cd, exp.loc, sc, f);
 
                 TypeFunction tf = cast(TypeFunction)f.type;
@@ -2251,11 +2247,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 if (!f || f.errors)
                     return setError();
 
-                exp.checkDeprecated(sc, f);
-                exp.checkDisabled(sc, f);
-                exp.checkPurity(sc, f);
-                exp.checkSafety(sc, f);
-                exp.checkNogc(sc, f);
+                checkFunctionAttributes(exp, sc, f);
                 checkAccess(cd, exp.loc, sc, f);
 
                 TypeFunction tf = cast(TypeFunction)f.type;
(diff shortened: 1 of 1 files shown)


src/dmd/expression.d
```diff
@@ -2308,6 +2308,24 @@ extern (C++) abstract class Expression : RootObject
...
+     * and usage of `deprecated` symbols are checked.
+     *
+     * Params:
+     *  sc = scope of the function
+     *  f  = function to be checked
+     * Returns: `true` if error occur.
+     */
+    final bool checkFunctionAttributes(Scope* sc, FuncDeclaration f)
+    {
+        return checkDeprecated(sc, f) ||
```
JinShil: Note that `checkDeprecated` does not return a `bool` (https://github.com/wilzbach/dmd/blob/226152b34ef76ce8439cc615dc09e5fc9b2d334c/src/dmd/expression.d#L1918).  Perhaps it should, though.  Maybe, as part of this refactoring, add the `bool` return value to make it consistent with other checks.

JinShil (changes requested):

src/dmd/expression.d
JinShil: Or that might be a better change for #7583 to include.  I'll take a look at that.

JinShil: FYI, I tried to make the change to return `bool` in #7583, but it's too systematic and turns the PR into a monstrosity.  I suggest that after #7583 is merged, we could make an intermediate PR that changes the return type to `bool` and then apply PR on top of that.

wilzbach: BTW it seems like we should stick with returning `void` for now:

```
Test failed: 
expected:
----
fail_compilation/diag10319.d(25): Error: pure function 'D main' cannot call impure function 'diag10319.foo'
fail_compilation/diag10319.d(25): Error: @safe function 'D main' cannot call @system function 'diag10319.foo'
fail_compilation/diag10319.d(26): Error: pure function 'D main' cannot call impure function 'diag10319.bar!int.bar'
fail_compilation/diag10319.d(26): Error: @safe function 'D main' cannot call @system function 'diag10319.bar!int.bar'
fail_compilation/diag10319.d(25): Error: function `diag10319.foo` is not nothrow
fail_compilation/diag10319.d(26): Error: function `diag10319.bar!int.bar` is not nothrow
fail_compilation/diag10319.d(23): Error: nothrow function `D main` may throw
----
actual:
----
fail_compilation/diag10319.d(25): Error: pure function 'D main' cannot call impure function 'diag10319.foo'
fail_compilation/diag10319.d(26): Error: pure function 'D main' cannot call impure function 'diag10319.bar!int.bar'
fail_compilation/diag10319.d(25): Error: function `diag10319.foo` is not nothrow
fail_compilation/diag10319.d(26): Error: function `diag10319.bar!int.bar` is not nothrow
fail_compilation/diag10319.d(23): Error: nothrow function `D main` may throw
----
```

src/dmd/expression.d
```diff
@@ -2320,6 +2320,28 @@ extern (C++) abstract class Expression : RootObject
...
+     * Checks the attributes of a function.
+     * Purity (`pure`), safety (`@safe`), no GC allocations(`@nogc`)
+     * and usage of `deprecated` and `@disabled`-ed symbols are checked.
+     *
+     * Params:
+     *  sc = scope of the function
+     *  f  = function to be checked
+     * Returns: `true` if error occur.
+     */
+    final bool checkFunctionAttributes(Scope* sc, FuncDeclaration f)
```
JinShil: I think this would be better as a `private static` function in the `expressionsem.d` module.

JinShil (changes requested):

src/dmd/expression.d
wilzbach: Done.

JinShil (approved):

PetarKirov (approved):

ZombineDev:
Nice, I was thinking of the same refactoring while looking at @JinShil's PR.

PR #7631, opened 2018-01-08
highlight scanmscoff.d error messages

JinShil:


JinShil:
Just realized this is wrong  "object module" is not a D module.

wilzbach:
@JinShil imho the point about colorizing error messages was to highlight important parts of the message. It doesn't need to D syntax errors.

wilzbach (approved):

PR #7635, opened 2018-01-08
[Trivial] Fix super.visit issue in SemanticTimeTransitiveVisitor

RazvanN7:
The methods which are mixed in aren't included in the overload set of the scope. This needs to be done manually. In my opinion, the compiler should do this automatically, but this PR only fixes the current issue so that calls to super.visit will be resolved correctly.

### DIFF
--- a/src/dmd/visitor.d
+++ b/src/dmd/visitor.d
@@ -112,7 +112,8 @@ extern (C++) class SemanticTimeTransitiveVisitor : SemanticTimePermissiveVisitor
 {
     alias visit = SemanticTimePermissiveVisitor.visit;
 
-    mixin ParseVisitMethods!ASTCodegen;
+    mixin ParseVisitMethods!ASTCodegen __methods;
+    alias visit = __methods.visit;
 
     override void visit(ASTCodegen.PeelStatement s)
     {


src/dmd/visitor.d
```diff
@@ -112,7 +112,8 @@ extern (C++) class SemanticTimeTransitiveVisitor : SemanticTimePermissiveVisitor
 {
     alias visit = SemanticTimePermissiveVisitor.visit;
 
-    mixin ParseVisitMethods!ASTCodegen;
+    mixin ParseVisitMethods!ASTCodegen methods;
```
jacob-carlborg: Perhaps prefix `methods` with two underscores or something. Because this symbol will be visible outside.

JinShil: > Because this symbol will be visible outside.

I don't like it.  With as powerful a language as D is, is there no way to make it private?

RazvanN7: Putting private before the mixin makes all the visit methods private. I don't know of any workaround.

RazvanN7: @jacob-carlborg Done, thanks!

JinShil: > In my opinion, the compiler should do this automatically

Maybe file a bug about this, and put a comment with a link to that bug in the source code.

RazvanN7: Already did it : https://issues.dlang.org/show_bug.cgi?id=18132

jacob-carlborg:
👌.

andralex (approved):

UplinkCoder:
*double underscores are reserved for compiler internal symbols!*
please use a single underscore.

Thanks!

PR #7636, opened 2018-01-08
[Do not merge] Parser experiment

RazvanN7:
This is an experiment to see how creating a new AST family would work. The API for dmd as a library is currently reduced to a few (parse time and semantic time) visitors and the semantic methods. This is insufficient and the compiler should be organized to accept families of ASTs since that would enable the user to modify any part of the compiler that he/she finds suitable.

The implementation duplicates the code in ASTCodegen which are all imports and inherits the AST node that needs to be modified (ClassDeclaration) and adds functionality in the method which needs to be modified (in this case, the constructor). The tool simply counts the number of class declarations in the code. Ideally, the AST classes should contain only fields and getter/setter methods and all functionally methods should be moved out of the compiler and be templated with an AST.

### DIFF
--- a/src/dmd/astcodegen.d
+++ b/src/dmd/astcodegen.d
@@ -33,6 +33,70 @@ struct ASTCodegen
     import dmd.typesem;
     import dmd.ctfeexpr;
 
+//    alias initializerToExpression   = dmd.initsem.initializerToExpression;
+    static initializerToExpression(Initializer i, Type t = null)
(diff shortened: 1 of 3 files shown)


JinShil:
@RazvanN7  Has this experiment concluded?

jacob-carlborg:
@RazvanN7 curios question, what's the purpose of pushing these experiments, to run it in the autotester?

PR #7637, opened 2018-01-08
Add 'man' target to build the MAN pages

wilzbach:
We should be a bit careful here to avoid breaking the nightly builds.

- Remove static dmd.1 in favor of the auto-generated one
- Add 'man' target (builds the MAN pages)
- Build MAN pages to generated
- Document Makefile targets

CC other package maintainers: @Vild @jordisayol @WebDrake (who knows more?)
 
See also: https://github.com/dlang/dmd/pull/7507

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -548,9 +548,20 @@ $G/newdelete.o: $G/%.o: $(ROOT)/%.c $(SRC_MAKE)
 	@echo "  (CC)  ROOT_OBJS  $<"
 	$(CXX) -c -o$@ $(CXXFLAGS) $(ROOT_FLAGS) $(MMD) $<
 
+################################################################################
+# Generate the man pages
+################################################################################
+
+DMD_MAN_PAGE = $(GENERATED)/docs/man1/dmd.1
+
+$(DMD_MAN_PAGE): dmd/cli.d
+	${MAKE} -C ../docs DMD=$(HOST_DMD_PATH) build
+
+man: $(DMD_MAN_PAGE)
+
 ######################################################
 
-install: all
+install: all $(DMD_MAN_PAGE)
 	$(eval bin_dir=$(if $(filter $(OS),osx), bin, bin$(MODEL)))
 	mkdir -p $(INSTALL_DIR)/$(OS)/$(bin_dir)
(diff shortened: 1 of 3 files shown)


src/posix.mak
```diff
@@ -25,6 +25,20 @@
...
+#
+# Targets
+# -------
+#
+# all					Build dmd
+# unittest              Run all unittest blocks
+# cxx-unittest          Check conformance of the C++ headers
+# build-examples        Build DMD as library examples
+# clean                 Remove all generated files
+# man                   Build the MAN pages
```
JinShil: nit: "Generate the man pages" as that is the text used in the comment below.

JinShil (changes requested):

JinShil:
Is it possible to add a test for this to our CIs?  I'm thinking something simple that generates the man page and then verifies that the file actually contains something sane.  I'm also worried that this can get broken somehow, and we'd only know about it after the release when one of our maintainers tries to create their package, or worse one of the users opens a man page with nothing or nonsense.

wilzbach:
> Is it possible to add a test for this to our CIs?

FYI: by adding it to the `install` target, which is run on e.g. auto-tester, we at least check that script doesn't error.

> UPDATE: Or perhaps add the output to CyberShadow/DAutoTest

Yep, excellent idea: https://github.com/dlang/dlang.org/pull/2067

wilzbach:
As dlang/dlang.org#2067 is in, I rebased this PR and now we should see the changes on DAutoTest.

wilzbach:
So it turns out that dlang/dlang.org#2067 broke DAutoTest (sorry @CyberShadow) because we modify versioned files.
As this PR fixes this and was intended to be merged directly after dlang/dlang.org#2067, I suggest we give this a higher priority. Hence, I dropped the only potentially controversial commit - the removal of the static man page file (I will do that in a separate file).
@JinShil - you still have "requested changes" here. Are you okay with merging this? Before I rebased DAutoTest showed two lines changed - the year changes (`DIFFABLE` will only take effect with this PR).

JinShil (approved):

wilzbach:
Great it seems like the vibe.d and dub testsuite depend on code.dlang.org being online. I wonder dub doesn't fallback to use its mirrors there. Anyhow this change doesn't even touch the compiler source code. and unbreaks DAutoTest

PR #7638, opened 2018-01-08
Refactor: make checkDisabled and checkDeprecated return bool

JinShil:
This is a followup to #7583 and a proposed prerequisite for #7628.

This also makes the interface consistent with all the other `checkXXX` methods.

It also contains a fix to the C++ interface that I missed in #7583 

cc @ibuclaw C++ interface change
cc @wilzbach

### DIFF
--- a/src/dmd/dsymbol.d
+++ b/src/dmd/dsymbol.d
@@ -305,7 +305,7 @@ extern (C++) class Dsymbol : RootObject
         va_end(ap);
     }
 
-    final void checkDeprecated(Loc loc, Scope* sc)
+    final bool checkDeprecated(Loc loc, Scope* sc)
     {
         if (global.params.useDeprecated != 1 && isDeprecated())
         {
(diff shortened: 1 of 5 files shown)


src/dmd/dsymbol.d
```diff
@@ -305,24 +305,24 @@ extern (C++) class Dsymbol : RootObject
...
             for (Scope* sc2 = sc; sc2; sc2 = sc2.enclosing)
             {
                 if (sc2.scopesym && sc2.scopesym.isDeprecated())
-                    return;
+                    return true;
 
                 // If inside a StorageClassDeclaration that is deprecated
                 if (sc2.stc & STC.deprecated_)
-                    return;
+                    return true;
```
wilzbach: Not sure about this. The definition for the other symbol is that `true` is returned if errors occur. In these cases error don't occur because it's allowed to e.g. use deprecated from deprecated.

JinShil: Gah! you're right.

JinShil: Fixed.  Thanks.

wilzbach (approved):

ibuclaw (approved):
Fine

PR #7639, opened 2018-01-08
Group scattered AST visitors into a `visitors` package

wilzbach:
I started off by moving the four visitors (`parsetime`, `permissive`, `strict`, `transitive`)
to their own package.
Then `visitor.d` and `visitors` looked a bit strange, so I moved `visitor.d` to `visitors/package.d`

As there is also `astXXXX`, `dmd.ast.visitors` would work too.

CC @RazvanN7

### DIFF
--- a/dub.sdl
+++ b/dub.sdl
@@ -44,7 +44,7 @@ subPackage {
     "src/dmd/astbase.d" \
     "src/dmd/parse.d" \
     "src/dmd/visitor/package.d" \
-    "src/dmd/visitor/parsetime.d" \
+    "src/dmd/visitor/parse_time.d" \
     "src/dmd/visitor/permissive.d" \
     "src/dmd/visitor/strict.d" \
     "src/dmd/visitor/transitive.d"
--- a/dub.sdl
+++ b/dub.sdl
@@ -43,9 +43,11 @@ subPackage {
   sourceFiles \
     "src/dmd/astbase.d" \
     "src/dmd/parse.d" \
-    "src/dmd/transitivevisitor.d" \
-    "src/dmd/permissivevisitor.d" \
-    "src/dmd/strictvisitor.d"
+    "src/dmd/visitor/package.d" \
+    "src/dmd/visitor/parsetime.d" \
+    "src/dmd/visitor/permissive.d" \
+    "src/dmd/visitor/strict.d" \
+    "src/dmd/visitor/transitive.d"
 
   dependency "dmd:lexer" version="*"
 }
--- a/dub.sdl
+++ b/dub.sdl
@@ -43,9 +43,10 @@ subPackage {
   sourceFiles \
     "src/dmd/astbase.d" \
     "src/dmd/parse.d" \
-    "src/dmd/visitor/package.d" \
     "src/dmd/visitor/parse_time.d" \
     "src/dmd/visitor/permissive.d" \
+    "src/dmd/visitor/semantic.d" \
+    "src/dmd/visitor/stoppable.d" \
     "src/dmd/visitor/strict.d" \
     "src/dmd/visitor/transitive.d"
 
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -53,7 +53,7 @@ import dmd.statement;
 import dmd.target;
 import dmd.tokens;
 import dmd.typesem;
-import dmd.visitor;
+import dmd.visitor.semantic;
 
 /*****************************************
  * CTFE requires FuncDeclaration::labtab for the interpretation.
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -127,9 +127,9 @@ extern(C++) Statement statementSemantic(Statement s, Scope* sc)
     return v.result;
 }
 
-private extern (C++) final class StatementSemanticVisitor : Visitor
+private extern (C++) final class StatementSemanticVisitor : SemanticVisitor
 {
-    alias visit = Visitor.visit;
+    alias visit = SemanticVisitor.visit;
 
     Statement result;
     Scope* sc;
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -300,7 +300,7 @@ FRONT_SRCS=$(addsuffix .d, $(addprefix $D/,access aggregate aliasthis apply argt
 	sideeffect statement staticassert target typesem traits \
 	typinf utils scanelf scanmach statement_rewrite_walker statementsem staticcond safe blockexit printast \
 	semantic2 semantic3 \
-	$(addprefix visitor/, package parsetime permissive transitive)))
+	$(addprefix visitor/, package parse_time permissive transitive)))
 
 LEXER_SRCS=$(addsuffix .d, $(addprefix $D/, console entity errors globals id identifier lexer tokens utf))
 
@@ -467,7 +467,7 @@ unittest: $G/dmd-unittest
 ######## DMD as a library examples
 
 EXAMPLES=$(addprefix $G/examples/, avg impvisitor)
-PARSER_SRCS=$(addsuffix .d, $(addprefix $D/,parse astbase $(addprefix visitor/, parsetime transitive permissive strict)))
+PARSER_SRCS=$(addsuffix .d, $(addprefix $D/,parse astbase $(addprefix visitor/, parse_time transitive permissive strict)))
 
 $G/parser.a: $(PARSER_SRCS) $(LEXER_SRCS) $(ROOT_SRCS) $G/dmd $G/dmd.conf $(SRC_MAKE)
 	CC="$(HOST_CXX)" $G/dmd -lib -of$@ $(MODEL_FLAG) -L-lstdc++ -J$G $(DFLAGS) $(PARSER_SRCS) $(LEXER_SRCS) $(ROOT_SRCS)
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -296,10 +296,11 @@ FRONT_SRCS=$(addsuffix .d, $(addprefix $D/,access aggregate aliasthis apply argt
 	dinifile dinterpret dmacro dmangle dmodule doc dscope dstruct dsymbol dsymbolsem	\
 	dtemplate dversion escape expression expressionsem func			\
 	hdrgen id impcnvtab imphint init initsem inline inlinecost intrange	\
-	json lambdacomp lib libelf libmach link mars mtype nogc nspace objc opover optimize parse permissivevisitor sapply templateparamsem	\
-	sideeffect statement staticassert target typesem traits transitivevisitor parsetimevisitor visitor	\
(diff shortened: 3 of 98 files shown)


src/tests/cxxfrontend.c
```diff
@@ -54,7 +54,7 @@
 #include "template.h"
 #include "tokens.h"
 #include "version.h"
-#include "visitor.h"
+#include "visitors/package.h"
```
ibuclaw: Can this just be kept in toplevel?  I'd be ok with a rename to `visitors.h`

wilzbach: Sure, I wasn't fully sure about this anyways (that's why I kept this change as a separate commit).

jacob-carlborg (approved):

WalterBright:
A little exposition on how this improves encapsulation would be nice. It's hard to tell from the diffs.

wilzbach:
> A little exposition on how this improves encapsulation would be nice. It's hard to tell from the diffs.

This group all the various visitors in the codebase neatly together, s.t. they are easy to find for an reader looking to learn more about DMD's visitors:

image

- the `dmd.visitors` module package is now the go-to place on the documentation and can e.g. contain a class hierarchy diagram (see below) and further explanations
- it highlights that `dmd.visitors.strict` is currently unused

Hierarchy in `dmd.visitors`
-----------------------------------

`extern (C++) class ParseTimeVisitor(AST)`
- `extern (C++) class Visitor : ParseTimeVisitor!ASTCodegen`
  - `extern (C++) class StoppableVisitor : Visitor`
- `extern (C++) class SemanticTimePermissiveVisitor : Visitor`
  - `extern (C++) class SemanticTimeTransitiveVisitor : SemanticTimePermissiveVisitor`
- `class PermissiveVisitor(AST): ParseTimeVisitor!AST`
  - `ParseTimeTransitiveVisitor(AST) : PermissiveVisitor!AST`
    - `StrictVisitor(AST) : ParseTimeVisitor!AST`

(simplified)

src/vcbuild/dmd.visualdproj
```diff
@@ -561,12 +559,14 @@
    <File path="..\dmd\tokens.d" />
    <File path="..\dmd\toobj.d" />
    <File path="..\dmd\traits.d" />
-   <File path="..\dmd\transitivevisitor.d" />
    <File path="..\dmd\typesem.d" />
    <File path="..\dmd\typinf.d" />
    <File path="..\dmd\utf.d" />
    <File path="..\dmd\utils.d" />
-   <File path="..\dmd\visitor.d" />
+   <File path="..\dmd\visitors\package.d.d" />
```
ZombineDev: `..\dmd\visitors\package.d`

src/posix.mak
```diff
@@ -296,10 +296,11 @@ FRONT_SRCS=$(addsuffix .d, $(addprefix $D/,access aggregate aliasthis apply argt
...
 	dtemplate dversion escape expression expressionsem func			\
 	hdrgen id impcnvtab imphint init initsem inline inlinecost intrange	\
-	json lib libelf libmach link mars mtype nogc nspace objc opover optimize parse permissivevisitor sapply templateparamsem	\
-	sideeffect statement staticassert target typesem traits transitivevisitor parsetimevisitor visitor	\
+	json lib libelf libmach link mars mtype nogc nspace objc opover optimize parse sapply templateparamsem	\
+	sideeffect statement staticassert target typesem traits \
 	typinf utils scanelf scanmach statement_rewrite_walker statementsem staticcond safe blockexit printast \
-	semantic2 semantic3))
+	semantic2 semantic3 \
+	$(addprefix visitors/, package parsetime permissive transitive)))
```
ZombineDev: What about `vistors.strict`? Why is it only part of the `PARSER_SRCS` list? (here and in win32.mak and the VS project).

dub.sdl
```diff
@@ -43,9 +43,9 @@ subPackage {
   sourceFiles \
     "src/dmd/astbase.d" \
     "src/dmd/parse.d" \
-    "src/dmd/transitivevisitor.d" \
-    "src/dmd/permissivevisitor.d" \
-    "src/dmd/strictvisitor.d"
+    "src/dmd/visitors/transitive.d" \
+    "src/dmd/visitors/permissive.d" \
+    "src/dmd/visitors/strict.d"
```
ZombineDev: Add `"src/dmd/visitors/package.d"`

ZombineDev: And also `parsetime`.

ZombineDev:
BTW, the convention for naming modules composed of two or more words is to separate the parts with an underscore. So how about renaming `parsetime` to `parse_time`?

dub.sdl
wilzbach: The reason why it didn't break, is because this package isn't used anymore. There's `frontend` now...

src/posix.mak
wilzbach: `visitors.strict` is a dead file - it isn't used by anything. I didn't want to modify the status quo in this PR.

ibuclaw:
> A little exposition on how this improves encapsulation would be nice. It's hard to tell from the diffs.

Just generally moving all common related actions into packages would be nice.  Another example would be semantic analysis, which is now scattered across many different modules that are intermixed with the other toplevel sources.  Having a semantic package (`dmd.semantic.dsymbol`, `dmd.semantic.statement`) is a much tidier way to organise them.

wilzbach:
> Having a semantic package (dmd.semantic.dsymbol, dmd.semantic.statement) is a much tidier way to organise them.

Absolutely agreed. This PR is just the start of a cleanup series :)

andralex:
cc @RazvanN7

andralex:
I have a bit of difficulty assessing the value here. At best things would go in a package that are highly likely to be used together. That way importing the entire package is a natural and desirable thing to do.

Here, using several visitors in the same module is the rare case; the frequent case is just using one, and of those, the frequent case is to just need `Visitor`. It strikes me as odd the reorganization makes it virtually impossible to simply import the one module that defines `Visitor`.

There seems to be merit in putting the entire visitation hierarchy in a separate package. But it seems the arrangement should make it such that `import visitor;` (singular, not plural) fetches only the base `Visitor` class and `import visitor.strict;` fetches the strict one etc.

ibuclaw (commented):
> I have a bit of difficulty assessing the value here. At best things would go in a package that are highly likely to be used together. That way importing the entire package is a natural and desirable thing to do.

https://dlang.org/phobos/std_container.html

There's half a dozen containers, but I'm not going to want to use all of them at once.

> Here, using several visitors in the same module is the rare case; the frequent case is just using one, and of those, the frequent case is to just need Visitor. It strikes me as odd the reorganization makes it virtually impossible to simply import the one module that defines Visitor.

I had a look at the implementation (rather than the comments that discussed the high level design), and indeed it seems odd that selective imports were added when they shouldn't be necessary.

> There seems to be merit in putting the entire visitation hierarchy in a separate package. But it seems the arrangement should make it such that import visitor; (singular, not plural) fetches only the base Visitor class and import visitor.strict; fetches the strict one etc.

Right, so the plural naming convention is not doing anything advantageous, and this should stick to the singular.

@wilzbach, any chance of doing one visitor per module and keeping the singular naming?

ie:
```
module dmd.visitor.stoppable;

import dmd.visitor; // for Visitor only.

...
```
It does strike me that ParseTimeVisitor is a terrible name (BaseVisitor? RootVisitor?), and SemanticTimeTransitive and SemanticTimePermissive seem a bit pointless.

I am going to assume that the distinction was made because overrides in the "classic" Visitor caused linker errors in the simple build examples that link only against lexer.a. Which would be unfortunate if I'm correct as there should be no distinction between parse and semantic passes as far as visitors go. But I digress...

RazvanN7:
>It does strike me that ParseTimeVisitor is a terrible name (BaseVisitor? RootVisitor?), and SemanticTimeTransitive and SemanticTimePermissive seem a bit pointless.

@ibuclaw The AST resulted after semantic analysis contains AST nodes that are never created at parse time. Given the fact that one can build the parser library without any semantics it is only natural that a visitor for that particular AST is offered. On the other hand, when one uses the complete frontend library, another visitor is offered which inherits the parse time visitor. Even though the semantic time visitors are not yet spread throughout the compiler, they are momentarily the interface to the compiler library that we have.

src/dmd/statementsem.d
```diff
@@ -53,7 +53,7 @@ import dmd.statement;
 import dmd.target;
 import dmd.tokens;
 import dmd.typesem;
-import dmd.visitor;
+import dmd.visitors : Visitor;
```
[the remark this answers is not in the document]
wilzbach: > Out of curiosity why selective import here?

I thought it's a good practice to show which parts are used.
I wasn't aware of this, but as long as some have it fixed, it will trigger errors.
Anyways, this has now been replaced with `dmd.visitor.visitor` and `dmd.visitor.stoppable`.

wilzbach:
> @wilzbach, any chance of doing one visitor per module and keeping the singular naming?

Done. I moved `Visitor` to `dmd.visitor.semantic` - as atm it should be called `SemanticVisitor`.
Should we rename `Visitor` to `SemanticVisitor`

>>  That way importing the entire package is a natural and desirable thing to do.
> There's half a dozen containers, but I'm not going to want to use all of them at once.

I agree with @ibuclaw here, we even enforce at Phobos that people can't import `std.algorithm`, but must be explicit.

> It strikes me as odd the reorganization makes it virtually impossible to simply import the one module that defines Visitor.

Done. Based on the comments of @ibuclaw and @RazvanN7 I renamed it to `SemanticVisitor` as far as I can tell, its proper name in the class hiearchy (it inherits from `ParseTimeVisitor`)

> There seems to be merit in putting the entire visitation hierarchy in a separate package. But it seems the arrangement should make it such that import visitor; (singular, not plural) fetches only the base Visitor class and import visitor.strict; fetches the strict one etc.

The convention at Phobos and other libraries is to avoid putting logic in `package.d` and only use `public import`.
Also
- The "base `Visitor` class" is `ParseTimeVisitor` - see the discussion above or below.
- `strict` isn't used anywhere at the moment

> It does strike me that ParseTimeVisitor is a terrible name (BaseVisitor? RootVisitor?)

Agreed! I like `BaseVisitor`.

wilzbach:
(rebased)

> Andrei: There seems to be merit in putting the entire visitation hierarchy in a separate package.

@andralex @WalterBright so can we move forward here? It looks like the PR is blocked on your approval.

WalterBright:
> This PR is just the start of a cleanup series :)

Oh gawd, please no. No packages, no glorious shuffling everything around.

WalterBright:
Let's close this.

I've pointed out several times a list of things that will substantively improve DMD's encapsulation, readability, maintainability, etc. I will prepare a better list and post it to the n.g. That's what we should be spending time on.

wilzbach:
> I've pointed out several times a list of things that will substantively improve DMD's encapsulation, readability, maintainability, etc. I will prepare a better list and post it to the n.g. That's what we should be spending time on.

Fair enough. Looking forward to the list!

PR #7640, opened 2018-01-08
Deprecate imaginary and complex types

wilzbach:
Turn the complex deprecation on by default.
Follow-up to https://github.com/dlang/dmd/pull/7081

This might need some work, depending on what Jenkins tells us.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1718,7 +1718,15 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                             foreach (t; Usage.transitions)
                             {
                                 if (t.bugzillaNumber !is null)
-                                    buf ~= `case `~t.bugzillaNumber~`: params.`~t.paramName~` = true;break;`;
+                                {
+                                    buf ~= `case `~t.bugzillaNumber~`:`;
+                                    if (t.deprecated_)
+                                        buf ~= `fprintf(global.stdmsg, "-transition=`~t.bugzillaNumber~` has no effect anymore\n");`;
+                                    else
+                                        buf ~= `params.`~t.paramName~` = true;`;
+
+                                    buf ~= "break;";
+                                }
                             }
                             return buf;
                         }
@@ -1738,12 +1746,19 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                             import dmd.cli : Usage;
                             string buf = `case "all":`;
                             foreach (t; Usage.transitions)
-                                buf ~= `params.`~t.paramName~` = true;`;
+                                if (!t.deprecated_)
+                                    buf ~= `params.`~t.paramName~` = true;`;
                             buf ~= "break;";
 
                             foreach (t; Usage.transitions)
                             {
-                                buf ~= `case "`~t.name~`": params.`~t.paramName~` = true;break;`;
+                                buf ~= `case "`~t.name~`":`;
+                                if (t.deprecated_)
+                                    buf ~= `fprintf(global.stdmsg, "-transition=`~t.name~` has no effect anymore\n");`;
+                                else
+                                    buf ~= `params.`~t.paramName~` = true;`;
+
+                                buf ~= "break;";
                             }
                             return buf;
                         }
--- a/changelog/complex-deprecated.dd
+++ b/changelog/complex-deprecated.dd
@@ -1,0 +1,21 @@
+Imaginary and complex types types have been deprecated - use `std.complex`
+
+D supported imaginary and complex versions of all floating point types:
+
+---
+float a = 2;
+ifloat b = 4i;
+cfloat c = a + b;
+assert(c == 2 + 4i);
+---
+
+Corrective action: Use the library types in $(MREF std, complex)
+
+---
+import std.complex, std.conv;
+float a = 2;
+auto b = complex(0f, 4);
+static assert(is(typeof(b) == Complex!float));
+auto c = a + b;
+assert(c.to!string == "2+4i");
+---
--- a/docs/gen_man.d
+++ b/docs/gen_man.d
@@ -127,6 +127,8 @@ void main()
 Language changes listed by \fB-transition=id\fR:`);
     foreach (transition; Usage.transitions)
     {
+        if (transition.deprecated_)
+            continue;
         string additionalOptions;
         if (transition.bugzillaNumber)
             additionalOptions = "," ~ transition.bugzillaNumber;
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -328,8 +328,9 @@ struct Usage
     {
         string bugzillaNumber; /// bugzilla issue number (if existent)
         string name; /// name of the transition
-        string paramName; // internal transition parameter name
-        string helpText; // detailed description of the transition
+        string paramName; /// internal transition parameter name
+        string helpText; /// detailed description of the transition
+        bool deprecated_; /// the flag isn't in use anymore
     }
 
     /// Returns all available transitions
@@ -341,7 +342,7 @@ struct Usage
         Transition(null, "checkimports", "check10378",
             "give deprecation messages about 10378 anomalies"),
         Transition("14488", "complex", "vcomplex",
-            "give deprecation messages about all usages of complex or imaginary types"),
+            "give deprecation messages about all usages of complex or imaginary types", true),
         Transition("16997", "intpromote", "fix16997",
             "fix integral promotions for unary + - ~ operators"),
         Transition(null, "tls", "vtls",
@@ -416,6 +417,8 @@ CPU architectures supported by -mcpu=id:
                 "list information on all language changes")] ~ Usage.transitions;
             foreach (t; allTransitions)
             {
+                if (t.deprecated_)
+                    continue;
                 buf ~= "  =";
                 buf ~= t.name;
                 auto lineLength = 3 + t.name.length;
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -87,7 +87,7 @@ struct Param
     bool vtls;              // identify thread local variables
     bool vgc;               // identify gc usage
     bool vfield;            // identify non-mutable field variables
-    bool vcomplex;          // identify complex/imaginary type usage
+    bool vcomplex = true;   // identify complex/imaginary type usage
     ubyte symdebug;         // insert debug symbolic information
     bool symdebugref;       // insert debug information for all referenced types, too
     bool alwaysframe;       // always emit standard stack frame
--- a/test/compilable/sw_transition_complex.d
+++ b/test/compilable/sw_transition_complex.d
@@ -1,5 +1,5 @@
 // PERMUTE_ARGS:
-// REQUIRED_ARGS: -c -transition=complex
+// REQUIRED_ARGS: -c
 
 /*
 TEST_OUTPUT:
--- a/test/fail_compilation/b3841.d
+++ b/test/fail_compilation/b3841.d
@@ -4,21 +4,34 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/b3841.d-mixin-31(31): Warning: `char += float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `int += float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `long += double` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `char -= float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `int -= float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `long -= double` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `char *= float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `int *= float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `long *= double` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `char /= float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `int /= float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `long /= double` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `char %= float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `int %= float` is performing truncating conversion
-fail_compilation/b3841.d-mixin-31(31): Warning: `long %= double` is performing truncating conversion
+fail_compilation/b3841.d(42): Deprecation: use of complex type `cfloat` is deprecated, use `std.complex.Complex!(float)` instead
+fail_compilation/b3841.d-mixin-44(44): Warning: `char += float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `int += float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `long += double` is performing truncating conversion
+fail_compilation/b3841.d(42): Deprecation: use of complex type `cfloat` is deprecated, use `std.complex.Complex!(float)` instead
+fail_compilation/b3841.d-mixin-44(44): Warning: `char -= float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `int -= float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `long -= double` is performing truncating conversion
+fail_compilation/b3841.d(42): Deprecation: use of complex type `cfloat` is deprecated, use `std.complex.Complex!(float)` instead
+fail_compilation/b3841.d-mixin-44(44): Warning: `char *= float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `int *= float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `long *= double` is performing truncating conversion
+fail_compilation/b3841.d(42): Deprecation: use of complex type `cfloat` is deprecated, use `std.complex.Complex!(float)` instead
+fail_compilation/b3841.d-mixin-44(44): Warning: `char /= float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `int /= float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `long /= double` is performing truncating conversion
+fail_compilation/b3841.d(42): Deprecation: use of complex type `cfloat` is deprecated, use `std.complex.Complex!(float)` instead
+fail_compilation/b3841.d-mixin-44(44): Warning: `char %= float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `int %= float` is performing truncating conversion
+fail_compilation/b3841.d-mixin-44(44): Warning: `long %= double` is performing truncating conversion
+fail_compilation/b3841.d(42): Deprecation: use of imaginary type `idouble` is deprecated, use `double` instead
+fail_compilation/b3841.d(43): Deprecation: use of imaginary type `ifloat` is deprecated, use `float` instead
+fail_compilation/b3841.d(42): Deprecation: use of imaginary type `ifloat` is deprecated, use `float` instead
+fail_compilation/b3841.d(43): Deprecation: use of imaginary type `idouble` is deprecated, use `double` instead
+fail_compilation/b3841.d(42): Deprecation: use of imaginary type `idouble` is deprecated, use `double` instead
+fail_compilation/b3841.d(43): Deprecation: use of imaginary type `ifloat` is deprecated, use `float` instead
+fail_compilation/b3841.d(42): Deprecation: use of imaginary type `ifloat` is deprecated, use `float` instead
+fail_compilation/b3841.d(43): Deprecation: use of imaginary type `idouble` is deprecated, use `double` instead
 ---
 */
 
@@ -47,7 +60,7 @@ void main()
         f!(op, float, long)();
         f!(op, cfloat, long)();
         f!(op, double, float)();
-        
+
         // Should that really be OK ?
         f!(op, short, int)();
         f!(op, float, double)();
@@ -66,7 +79,7 @@ void main()
         // Should that really be OK ?
         f!(op, ifloat, idouble)();
     }
-    
+
     // OK
     f!("^^=", int, int)();
     f!("^^=", long, int)();
--- a/test/fail_compilation/fail101.d
+++ b/test/fail_compilation/fail101.d
@@ -1,7 +1,8 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/fail101.d(8): Error: cannot implicitly convert expression `1` of type `int` to `creal`
+fail_compilation/fail101.d(9): Deprecation: use of complex type `creal` is deprecated, use `std.complex.Complex!(real)` instead
+fail_compilation/fail101.d(9): Error: cannot implicitly convert expression `1` of type `int` to `creal`
 ---
 */
 
--- a/test/fail_compilation/fail303.d
+++ b/test/fail_compilation/fail303.d
@@ -1,13 +1,14 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/fail303.d(19): Error: `double /= cdouble` is undefined. Did you mean `double /= cdouble.re`?
-fail_compilation/fail303.d(20): Error: `ireal *= ireal` is an undefined operation
-fail_compilation/fail303.d(21): Error: `ireal *= creal` is undefined. Did you mean `ireal *= creal.im`?
-fail_compilation/fail303.d(22): Error: `ireal %= creal` is undefined. Did you mean `ireal %= creal.im`?
-fail_compilation/fail303.d(23): Error: `ireal += real` is undefined (result is complex)
-fail_compilation/fail303.d(24): Error: `ireal -= creal` is undefined (result is complex)
-fail_compilation/fail303.d(25): Error: `double -= idouble` is undefined (result is complex)
+fail_compilation/fail303.d(18): Deprecation: use of imaginary type `ireal` is deprecated, use `real` instead
+fail_compilation/fail303.d(20): Error: `double /= cdouble` is undefined. Did you mean `double /= cdouble.re`?
+fail_compilation/fail303.d(21): Error: `ireal *= ireal` is an undefined operation
+fail_compilation/fail303.d(22): Error: `ireal *= creal` is undefined. Did you mean `ireal *= creal.im`?
+fail_compilation/fail303.d(23): Error: `ireal %= creal` is undefined. Did you mean `ireal %= creal.im`?
+fail_compilation/fail303.d(24): Error: `ireal += real` is undefined (result is complex)
+fail_compilation/fail303.d(25): Error: `ireal -= creal` is undefined (result is complex)
+fail_compilation/fail303.d(26): Error: `double -= idouble` is undefined (result is complex)
 ---
 */
 


wilzbach:
Okay after looking a bit into this (see https://github.com/dlang/phobos/pull/6014) I found the following problems:
- How can `typeid(cfloat)` by called without triggering a deprecation warning?
- This already triggers a deprecation warning:

```
deprecated foo(creal r) {}
```

So usage of `creal` etc. needs to be allowed in `deprecated` (-> https://issues.dlang.org/show_bug.cgi?id=18212)

- there are quite a bunch of [deprecation messages in druntime](https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=2985474&dataid=21127003&isPull=true) to to all `typeinfo`, it's spamy but shouldn't be an issue  

- `std.conv`, `std.format` and `std.traits` will be tricky, e.g.

```
std/conv.d(205): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(205): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(456): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(570): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(586): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(637): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(686): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(759): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(880): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(1024): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(1400): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(1493): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(1575): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(1818): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(1897): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(1935): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
std/conv.d(1400): Deprecation: use of complex type cdouble is deprecated, use std.complex.Complex!(double) instead
```

For example:

- [std.format's deprecation messages](http://ix.io/DOD)
- [std.traits's deprecation messages](http://ix.io/DOC)

ibuclaw:
Start by removing all tests using complex and imaginary types?

wilzbach:
> Could help by removing all tests using complex and imaginary types?

Do you mean for Phobos? Yep, that's one idea, but I guess improving this deprecation check to check whether's in a `deprecated` scope would be better, because we can ensure that no user code is broken during the deprecation phase.

ibuclaw:
Yeah, should be simple.  Pass the `sc` parameter to the function.

src/dmd/mars.d
```diff
@@ -1706,6 +1706,7 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                             params.bug10378 = true;
                             break;
                         case 14488:
+                            printf("-transition=14488 is no longer needed.\n");
```
ibuclaw: Use `global.stdmsg`?

wilzbach: I have seen `printf` being used here a couple of times, but sure `global.stdmsg` looks cleaner.

wilzbach:
image

@ibuclaw looks like we can finally start the official deprecation cycle :tada:

src/dmd/mars.d
```diff
@@ -1738,12 +1746,19 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
...
+                                if (!t.deprecated_)
+                                    buf ~= `params.`~t.paramName~` = true;`;
                             buf ~= "break;";
 
                             foreach (t; Usage.transitions)
                             {
-                                buf ~= `case "`~t.name~`": params.`~t.paramName~` = true;break;`;
+                                buf ~= `case "`~t.name~`":`;
+                                if (t.deprecated_)
+                                    buf ~= `fprintf(global.stdmsg, "-transition=`~t.name~` has no effect anymore\n");`;
```
wilzbach: We can't trigger a real deprecation error because Phobos uses `-transition=complex` for now.

andralex:
What happened with @9il 's concerns?

9il:
https://godbolt.org/g/xvEpNH - The same link about performance issues.

ibuclaw:
You're coming about 6 years too late @9il, but that's fine.

A small penalty moving to soft type of course expected. And std.complex is not a module that I've seen change much from its original implementation, so there's bound to be things brought into sharp focus that need to be tended to.

Are your concerns even well founded though? I seem to recall at the one of the first dconfs the question posed "have you ever used complex types?" being asked, all came back with blank faces. I even asked the same question at GNU Hackers meeting a couple months later, and out of the many dozens only one came back with an answer: "I used it once to encode colour sequences, but didn't really need to do it that way."

Complex numbers have been deprecated for more than twice as long as Mirs entire existence. There's nothing that should be possible with complex that can't be done with packed doubles and still retain more or less the same cost.

So, what's your use case?

9il:
OK, I will do a copy with required fixes whrn I need it. Bb

jacob-carlborg:
What about interfacing with C (`_Complex`)? Does/did D support that?

andralex:
It would be interesting to see what the causes of performance loss are and how to mitigate them by language features such as `@inherit`. Compatibility with `_Complex` is also important.

First off, @9il can you provide insight on the performance loss? We really need detailed information here; naively one would think one complex is exactly two floating-point numbers, and grouping those in a `struct` should be easy to destructure by the optimizer, thus getting back to the constituents. Where is this reasoning going wrong? Can you produce a brief Wiki page describing in detail what the issues are and how language features and/or optimizer improvements can help?

andralex:
Higher-level comment: I don't want to rush these things, in fact I'm actively opposed to moving along deprecation.

* Complex numbers are no bother for those not using them. The last time "complex" appears in the title of a forum post is 2015/03/21. So removal of the feature won't improve their life.
* Conversely, those who do use complex numbers would be adversely affected.
* However poorly explained, there seems to be evidence removing complex numbers has non-obvious consequences related to efficiency and C compatibility. These are important and deserve further investigation. Just deprecating and then removing complex numbers and letting users deal with it should be preceded by a careful investigation.
* Complex numbers are not poorly defined and do not occupy large real estate in the language. Removing them would reduce the size of D only nominally.
* In contrast, there are other areas of the language that are poorly defined and that do hurt the language size, complexity, and rigor adversely. Any time spent on this by our collaborators is time not spent there, creating a large opportunity cost.

A simple way to put it is focus on impact. Of the things we can do, what effort has maximum impact? This is definitely not close to that.

JinShil:
IMO the controversy surrounding this deprecation is not because the deprecation is currently being followed through on, but because it wasn't followed through on years earlier.  Not following up on this unfinished business is creating technical debt and encouraging bad choices by users who don't believe deprecation means much in the development of D.

9il:
> First off, @9il can you provide insight on the performance loss? 

@andralex Well defined (academic) description can be done by LLVM optimiser contributors. I only did hundreds of LLVM IR and assembler output reviews and performance tests. It is 99% experience rather then knowledge.

Definitely it is not a problem to group data into a struct (maybe except for C support). The problem is how inlining works in pair with optimiser.

There are two issues.
1) Data flow based optmization. LDC does not use DMD FE inlining, it uses LLVM inlining. A hypotetic `@feinline` attribute (or `pragma(feinline, true)`) for example can solve this issue as well as the second one. Why this is important? It is not important for common use case, but very important for small math structures like complex numbers and for _many_ math function that compute multiple values. `a * b` is such function, it compute two values `re` and `im`. Looks like that there are few stages of optimization. Looks like the optimiser _frequently_ ignores following information when do inlining
   a) specific input values for a function (for example `Complex(2, 1)` or `Complex(a.re, -a.im)`).
   b) specific operation that are done with the result of computation (you may need only a `re` part for example).

Even if the optimizer can optimize well the function for specific input it may not do some after inlining data flow optimization, because the optimizer _do not generate important information (based on function input arguments) about how a function result is interconnected with other computations_.

2) LLVM floating-point optimisation attributes. To be honest they are not well defined by LLVM spec. But I figured out how they affect the output and can say that with some experience and new LDC features we can combine IEEE and non-IEEE code very well.

We can chat in hangouts about other details.

ibuclaw:
I'll give an attempt at counter-arguments to some points:

>  Complex numbers are no bother for those not using them. The last time "complex" appears in the title of a forum post is 2015/03/21. So removal of the feature won't improve their life.

Two notable open bugs on complex numbers, that I think show that there is in fact a little bit of bother using them.

https://issues.dlang.org/show_bug.cgi?id=692
https://issues.dlang.org/show_bug.cgi?id=5570

The latter one is a serious ABI bug which _is_ affecting industry code.  The workaround was to use inline assembler to put the complex number in the correct registers before calling the function.

 >   Conversely, those who do use complex numbers would be adversely affected.

The same argument could have been made for NCEG operators.  Removing them in favour of std.math functions _also_ had a performance impact.  But nothing that can't be solved as a library solution.

>    However poorly explained, there seems to be evidence removing complex numbers has non-obvious consequences related to efficiency and C compatibility. These are important and deserve further investigation. Just deprecating and then removing complex numbers and letting users deal with it should be preceded by a careful investigation.

@leandro-lucarella-sociomantic, what impact would this have on you?

>    Complex numbers are not poorly defined and do not occupy large real estate in the language. Removing them would reduce the size of D only nominally.

I would put combined complex expression/type handling at around 2-3k lines of code as a conservative estimate for gdc (front and backend).  There's plenty of special handling that needs to be done for them, such as supporting the (needless) distinction between real and imaginary types.

ibuclaw:
> A hypotetic @feinline attribute (or pragma(feinline, true)) for example can solve this issue as well as the second one

Nobody in their right mind would use the inliner included with dmd. :-)

Regardsless, there's already a pragma for inlining.  One is more than enough.

ibuclaw:
> Compatibility with _Complex is also important.

Where does `core.complex` fall into this?  Is there to be a new "special" type along the lines of `c_long_double` for ABI compatibility?

https://github.com/dlang/druntime/blob/cf41932a8e1fa6ff01a3c96cbaae8742b716ab91/src/core/stdc/config.d#L98-L115

andralex:
>Definitely it is not a problem to group data into a struct (maybe except for C support). 

So then we're back to square one because my understanding was that's the main problem. This is completely confusing. I read the sentences and they parse, and at the and I have no idea of the semantics and pragmatics.

I don't even know how to ask more clearly: can we switch complex from a built-in type to a struct with two members? In doing so, can we count on current and future optimizers to work with the struct just as well as they work now with the built-in type?

If not, what steps do we need to take to improve the language? Is a DIP necessary for new features such as `@inherit`?

ibuclaw:
>  can we switch complex from a built-in type to a struct with two members? In doing so, can we count on current and future optimizers to work with the struct just as well as they work now with the built-in type?

Yes, they should. The difference is a matter of ABI.  On x86_64 `cdouble` is passed and returned in XMM registers, but a `struct {double, double }` is passed and returned on the stack.

andralex:
@ibuclaw ok, ok, that's good stuff. Question for @ibuclaw and @9il: is that likely to cause big troubles? I suspect not, because inlining should eliminate the need for passing individual complex values to small functions. Right?

@ibuclaw what's the calling convention for C's `_Complex`?

ibuclaw:
> I suspect not, because inlining should eliminate the need for passing individual complex values to small functions. Right?

Its just the cost of copying to and from xmm registers and the stack whenever you use or modify the struct type.  I see no inlining problems, particularly if the stuct type is templated.

ibuclaw:
> @ibuclaw what's the calling convention for C's _Complex?

You mean, how is it passed?  And on which target?

andralex:
@ibuclaw I wonder if _Complex would be compatible with a struct layout.

ibuclaw:
So, `Complex!double s = *cast(Complex!double*)&c`? 

The layout is the same, they just live in different places.

ibuclaw:
Just for comparison.

---
Native Multiplication.

LDC:
```
cfloat example.nativeMul(cfloat, cfloat):
  vmovq rax, xmm1
  vmovd xmm1, eax
  shr rax, 32
  vmovd xmm2, eax
  vmovq rax, xmm0
  vmovd xmm0, eax
  shr rax, 32
  vmovd xmm3, eax
  vmulss xmm4, xmm0, xmm2
  vmulss xmm2, xmm3, xmm2
  vfmsub213ss xmm0, xmm1, xmm2
  vfmadd231ss xmm4, xmm3, xmm1
  vmovss dword ptr [rsp - 8], xmm0
  vmovss dword ptr [rsp - 4], xmm4
  vmovsd xmm0, qword ptr [rsp - 8]
  ret
```
GDC:
```
cfloat example.nativeMul(cfloat, cfloat):
  vmovq QWORD PTR -24[rsp], xmm1
  vmovss xmm3, DWORD PTR -20[rsp]
  vmovss xmm2, DWORD PTR -24[rsp]
  vmovq QWORD PTR -16[rsp], xmm0
  vmovss xmm1, DWORD PTR -12[rsp]
  vmovss xmm0, DWORD PTR -16[rsp]
  jmp __mulsc3@PLT
```

---
std.complex Multiplication.

LDC:
```
std.complex.Complex!(float).Complex example.stdMul(std.complex.Complex!(float).Complex, std.complex.Complex!(float).Complex):
  vmovq rax, xmm1
  vmovd xmm1, eax
  shr rax, 32
  vmovd xmm2, eax
  vmovq rax, xmm0
  vmovd xmm0, eax
  vmulss xmm3, xmm1, xmm0
  shr rax, 32
  vmovd xmm4, eax
  vmulss xmm5, xmm2, xmm4
  vsubss xmm3, xmm3, xmm5
  vmulss xmm0, xmm2, xmm0
  vmulss xmm1, xmm1, xmm4
  vaddss xmm0, xmm1, xmm0
  vmovss dword ptr [rsp - 4], xmm0
  vmovss dword ptr [rsp - 8], xmm3
  vmovsd xmm0, qword ptr [rsp - 8]
  ret
```
GDC:
```
std.complex.Complex!(float).Complex example.stdMul(std.complex.Complex!(float).Complex, std.complex.Complex!(float).Complex):
  vmovq rdx, xmm1
  vmovq rcx, xmm0
  vmovdqa xmm3, xmm1
  shr rdx, 32
  shr rcx, 32
  vmovdqa xmm1, xmm0
  vmovd xmm2, edx
  vmovd xmm4, ecx
  vmulss xmm0, xmm4, xmm2
  vfmsub231ss xmm0, xmm3, xmm1
  vmulss xmm1, xmm1, xmm2
  vmovd eax, xmm0
  vfmadd231ss xmm1, xmm3, xmm4
  mov eax, eax
  vmovq rdx, xmm1
  sal rdx, 32
  or rax, rdx
  vmovq xmm0, rax
  ret
```

---
Native Divide.

LDC:
```
cfloat example.nativeDiv(cfloat, cfloat):
  vmovq rax, xmm1
  vmovd xmm1, eax
  shr rax, 32
  vmovd xmm2, eax
  vmovq rax, xmm0
  vmovd xmm0, eax
  shr rax, 32
  vmovd xmm3, eax
  vmulss xmm4, xmm0, xmm1
  vfmadd231ss xmm4, xmm3, xmm2
  vmulss xmm1, xmm3, xmm1
  vfmsub231ss xmm1, xmm0, xmm2
  vmulss xmm0, xmm0, xmm0
  vfmadd231ss xmm0, xmm3, xmm3
  vmovss xmm2, dword ptr [rip + .LCPI2_0]
  vdivss xmm0, xmm2, xmm0
  vmulss xmm2, xmm4, xmm0
  vmulss xmm0, xmm1, xmm0
  vmovss dword ptr [rsp - 8], xmm2
  vmovss dword ptr [rsp - 4], xmm0
  vmovsd xmm0, qword ptr [rsp - 8]
  ret
```
GDC:
```
cfloat example.nativeDiv(cfloat, cfloat):
  vmovq QWORD PTR -24[rsp], xmm1
  vmovss xmm3, DWORD PTR -20[rsp]
  vmovss xmm2, DWORD PTR -24[rsp]
  vmovq QWORD PTR -16[rsp], xmm0
  vmovss xmm1, DWORD PTR -12[rsp]
  vmovss xmm0, DWORD PTR -16[rsp]
  jmp __divsc3@PLT
```

---
std.complex Divide.

LDC:
```
std.complex.Complex!(float).Complex example.stdDiv(std.complex.Complex!(float).Complex, std.complex.Complex!(float).Complex):
  push rbx
  sub rsp, 32
  vmovq rax, xmm1
  vmovd xmm1, eax
  vmovd dword ptr [rsp + 12], xmm1
  mov dword ptr [rsp + 16], eax
  shr rax, 32
  vmovd xmm1, eax
  vmovd dword ptr [rsp + 8], xmm1
  mov dword ptr [rsp + 20], eax
  vmovq rbx, xmm0
  vmovd xmm0, ebx
  vmovd dword ptr [rsp + 4], xmm0
  call pure nothrow @nogc @safe float std.math.fabs(float)@PLT
  vmovd dword ptr [rsp + 28], xmm0
  shr rbx, 32
  vmovd xmm0, ebx
  vmovd dword ptr [rsp], xmm0
  call pure nothrow @nogc @safe float std.math.fabs(float)@PLT
  vmovss xmm1, dword ptr [rsp + 28]
  vucomiss xmm0, xmm1
  jbe .LBB3_2
  vmovss xmm0, dword ptr [rsp + 4]
  vmovss xmm1, dword ptr [rsp]
  vdivss xmm2, xmm0, xmm1
  vmulss xmm0, xmm2, xmm0
  vaddss xmm0, xmm0, xmm1
  vmovss xmm3, dword ptr [rsp + 12]
  vmulss xmm1, xmm2, xmm3
  vmovss xmm4, dword ptr [rsp + 8]
  vaddss xmm1, xmm1, xmm4
  vdivss xmm1, xmm1, xmm0
  vmulss xmm4, xmm2, xmm4
  jmp .LBB3_3
.LBB3_2:
  vmovss xmm1, dword ptr [rsp + 4]
  vmovss xmm0, dword ptr [rsp]
  vdivss xmm2, xmm0, xmm1
  vmulss xmm0, xmm2, xmm0
  vaddss xmm0, xmm0, xmm1
  vmovss xmm4, dword ptr [rsp + 8]
  vmulss xmm1, xmm2, xmm4
  vmovss xmm3, dword ptr [rsp + 12]
  vaddss xmm1, xmm1, xmm3
  vdivss xmm1, xmm1, xmm0
  vmulss xmm3, xmm2, xmm3
.LBB3_3:
  vsubss xmm2, xmm4, xmm3
  vdivss xmm0, xmm2, xmm0
  vmovss dword ptr [rsp + 20], xmm0
  vmovss dword ptr [rsp + 16], xmm1
  vmovsd xmm0, qword ptr [rsp + 16]
  add rsp, 32
  pop rbx
  ret
```
GDC:
```
std.complex.Complex!(float).Complex example.stdDiv(std.complex.Complex!(float).Complex, std.complex.Complex!(float).Complex):
  vmovq rax, xmm1
  sub rsp, 40
  vmovdqa xmm4, xmm1
  shr rax, 32
  vmovd DWORD PTR 16[rsp], xmm0
  mov DWORD PTR 12[rsp], eax
  vmovq rax, xmm0
  vmovaps xmm0, xmm4
  shr rax, 32
  vmovss DWORD PTR 24[rsp], xmm4
  mov DWORD PTR 28[rsp], eax
  call float std.math.fabs(float)@PLT
  vmovss DWORD PTR 20[rsp], xmm0
  vmovss xmm0, DWORD PTR 12[rsp]
  call float std.math.fabs(float)@PLT
  vucomiss xmm0, DWORD PTR 20[rsp]
  vmovss xmm1, DWORD PTR 28[rsp]
  ja .L67
  vmovss xmm7, DWORD PTR 12[rsp]
  vmovss xmm5, DWORD PTR 24[rsp]
  vmovss xmm6, DWORD PTR 16[rsp]
  vdivss xmm3, xmm7, xmm5
  vmovaps xmm0, xmm6
  vfmadd132ss xmm7, xmm5, xmm3
  vfmadd231ss xmm0, xmm1, xmm3
  vfnmadd231ss xmm1, xmm3, xmm6
  vdivss xmm0, xmm0, xmm7
  vdivss xmm1, xmm1, xmm7
.L63:
  vmovd edx, xmm0
  vmovd eax, xmm1
  add rsp, 40
  sal rax, 32
  mov edx, edx
  or rdx, rax
  vmovq xmm0, rdx
  ret
.L67:
  vmovss xmm4, DWORD PTR 24[rsp]
  vmovss xmm5, DWORD PTR 12[rsp]
  vmovss xmm6, DWORD PTR 16[rsp]
  vdivss xmm2, xmm4, xmm5
  vmovaps xmm0, xmm6
  vfmadd132ss xmm4, xmm5, xmm2
  vfmadd132ss xmm0, xmm1, xmm2
  vfmsub132ss xmm1, xmm6, xmm2
  vdivss xmm0, xmm0, xmm4
  vdivss xmm1, xmm1, xmm4
  jmp .L63
```

ibuclaw:
Notable observations / tl;dr.

1. LDC native complex is inline.
2. GDC native complex uses libgcc optimized functions.
3. GDC std.complex is actually comparable to LDC native.
4. LDC std.complex sadly is leaving us wanting.
5. std.complex operator divide could benefit from having less branching, or...
6. It might be better to use `core.math.fabs` directly to explicitly use the intrinsic.

9il:
I can not discus this more. Sorry. My arguments still valid.

ibuclaw:
> I can not discus this more. Sorry. My arguments still valid.

In what sense though? Just to paraphrase your argument into a digestible form:

1. D compilers are not able to optimize simple functions in std.complex such as `a * b` because [reasons/limitations] in inliner.
2. Even if D compilers can optimize well, it still produces bad code in more complex cases because [reasons/limitations] in optimizer.
3. We need to add new attributes and semantics to the language frontend in order to overcome these problems in the backend, so that std.complex can produce equivalent code to native complex.


Whereas from what I can see based on your examples, points 1 and 2 are simply not true (see codegen of GDC). It is not outside the capability of a D compiler to generate optimal code using std.complex. 

And so rather than point 3, it looks like only minor improvements to std.complex to encourage better codegen is needed in some places of phobos. I can comb through this later.

leandro-lucarella-sociomantic:
>> However poorly explained, there seems to be evidence removing complex numbers has non-obvious consequences related to efficiency and C compatibility. These are important and deserve further investigation. Just deprecating and then removing complex numbers and letting users deal with it should be preceded by a careful investigation.
>
> @leandro-lucarella-sociomantic, what impact would this have on you?

I'm not sure what do you mean by that, but what I would agree that this should not be deprecated until there is some certainty that the alternative is ready to fully replace the deprecated functionality. There is no point in deprecating something when the alternative is half done, because then you let the user in a wasteland where you either get spurious deprecation messages or you have to start using a half-baked alternative.

ibuclaw:
Functionality has been there for years in std.complex.  I wouldn't call it half done, but increased use/reliance on it will bring into sharp focus areas that could be improved, either at compile or runtime. Both are not blockers per say, but it will reflect on std.complex positively as far as performance goes.

Only open question I see that has not been discussed is what a hypothetical core.complex would look like and why, and what precisely would be exposed by the compiler @andralex.

andralex:
>Only open question I see that has not been discussed is what a hypothetical core.complex would look like and why, and what precisely would be exposed by the compiler @andralex.

Intuitively programs should be able to use the complex type, at least the basic (layout and fundamental operations) without needing the standard library.

wilzbach:
> Intuitively programs should be able to use the complex type, at least the basics (layout and fundamental operations) without needing the standard library.

We ship the standard library with all releases - it's after all the _standard library_, so I think you were referring to "usable in -betterC". As mentioned before by @ibuclaw nearly all methods are already templates and thus can be used in `-betterC`. We can even add a [testsuite for `-betterC`](https://github.com/dlang/phobos/pull/5952) to Phobos, so I don't see the reason for duplicating `std.complex` to `core.complex`.

ibuclaw:
@kinke you wanted to know some examples, here is one.

wilzbach:
> Complex numbers are no bother for those not using them. The last time "complex" appears in the title of a forum post is 2015/03/21. So removal of the feature won't improve their life.
> Two notable open bugs on complex numbers, that I think show that there is in fact a little bit of bother using them.

As long as we don't pull the switch here, people will use the built-in complex types and not `std.complex` or a third-party DUB package (see the recent bug report below).

> In contrast, there are other areas of the language that are poorly defined and that do hurt the language size, complexity, and rigor adversely. Any time spent on this by our collaborators is time not spent there, creating a large opportunity cost.

Yes, but the implementation of complex numbers in DMD isn't well tested because not so many people use it and hence people run into bugs:

https://issues.dlang.org/show_bug.cgi?id=18464

I don't think that finding all the places in DMD which need to be special-cased for complex number isn't time well spent either.

> Intuitively programs should be able to use the complex type, at least the basics (layout and fundamental operations) without needing the standard library.

Why does it need to be in DRuntime then? Can't we just ensure that the basics work with `-betterC`? After all, it's already templated...

Also we have made a lot of bad experiences with adding modules to druntime and phobos + the interest of using complex numbers without Phobos seems to be limited to math libraries like mir, so why couldn't this be a DUB package?

ibuclaw:
> limited to math libraries like mir

Well, if you are a maintainer of a specialization library geared towards math, you probably won't use half of what's available in std anyway.

I still think the only outstanding thing here is ABI compatibility with C.  We already have `struct __c_long`, `struct __c_ulong`, and `struct __c_long_double` that are a special library defined type recognized by the compiler for ABI conformity.  Why not `struct __c_complex_double` also?

p0nce:
Hello, I'm upgrading to 2.079 and I see builtin complexs continue their descent into deprecation. Many functions in std.math have been deprecated.

Please reconsider, builtin complexes are extremely useful.

I heavily use builtin complex types for signal processing, where they are both more readable and **faster**.
Made a benchmark to illustrate the sometimes catastrophic performance profile of std.complex "zero-cost" abstraction, that turns out to have a real cost once measured.

BENCHMARK:
https://issues.dlang.org/show_bug.cgi?id=18627

Key takeaway:
> We are looking to a guaranteed 2x to 10x performance regression when using complex number arithmetic + DMD

I use builtin complex originally because they are faster (in that I've also found what @9il  has found) but also the syntax is an improvement over just a struct. I think this realization will be found by anyone doing heavy number crunching with D.

In fact most of my revenue comes from a product that does 1000+ 2048 point FFT and would get noticeably slower if std.complex were to be used instead. It's really depressing to me, D is supposed to be about top performance, but when you are not looking the rug is pulled under you because "progress" is made without measurements. Not everyone can monitor the bugtracker all the time. I was thinking complex numbers would never be dropped.

p0nce:
> Yes, they should. The difference is a matter of ABI. For example, on x86_64 cdouble is passed and returned in XMM registers, but a struct {double, double } is passed and returned on the stack.

In other words, returning a `cdouble` will be two indirections instead of returning in registers. This is not necessarily the same performance profile. (even assuming the stack is hot, that's still a store/load if not inlined).

ibuclaw:
> I use builtin complex originally because they are faster (in that I've also found what @9il has found) but also the syntax is an improvement over just a struct. I think this realization will be found by anyone doing heavy number crunching with D.

And just for balance, I use std.complex because I found the _opposite_ to be true. :-)

I've already stated [https://github.com/dlang/dmd/pull/7640#issuecomment-359207269](here) that std.complex is not inherently slow, and the compiler can generate performant code with it.

p0nce:
@ibuclaw In your comment you showed that (at least in GDC), builtin complex division and multiplication jump to a library, but there are no measurements that it's faster or slower that way.

Builtin complexes are the thing that is being replaced. It's the baseline.
`std.complex` has to be measured against this baseline, else how do we know we really care about performance. @9il also had a set of more sophisticated remarks.

From the user point of view I don't care if they are kept or not, I only care that the replacement is _at least as fast_ guaranteed. Else it's a performance regression, and when using DMD this performance regression is (currently) guaranteed.

I get it. Most people don't need complex numbers at all. So everyone want them dead, and assume noone is using them (again without asking since 6 years). But some of us are. I've become increasingly admirative of their design since doing a bit of math. It's not like they were added for no reason. Complex numbers are used thoughout numerical programming.

Mostly I trust the inliner, but using LDC there is nothing to trust or not being that builtin complexes are always inlined. Not having to trust > trust.

wilzbach:
>  DMD this performance regression is (currently) guaranteed.

With this argument, we wouldn't be allowed to use ranges. They are usually 2-3x slower than their naive loop replacement in DMD.

> Mostly I trust the inliner, but using LDC there is nothing to trust or not being that builtin complexes are always inlined. Not having to trust > trust.

That's what `pragma(inline, true)` was made for, no?
It could be easily added to more functions in `std.complex`.

p0nce:
Well, I certainly don't use any range in sensitive numerical processing functions. That's just basic de-risking to me. If you opitmize you just can't assume "zero-cost-abstractions" are zero-cost because of their name.

Geod24:
@ibuclaw has taken this over, so I think we can close this.

PR #7641, opened 2018-01-08
dmd.transitivevisitor: Make non-overriding methods private

kinke:
... of mixin template `ParseVisitMethods`, instead of introducing new virtual functions.

This works around https://issues.dlang.org/show_bug.cgi?id=18200, i.e., makes DMD compile for MSVC targets with compilers < 2.078 again.

@rainers came up with the workaround, thanks again!

### DIFF
--- a/src/dmd/transitivevisitor.d
+++ b/src/dmd/transitivevisitor.d
@@ -56,7 +56,7 @@ package mixin template ParseVisitMethods(AST)
         }
     }
 
-    void visitVarDecl(AST.VarDeclaration v)
+    private void visitVarDecl(AST.VarDeclaration v)
     {
         //printf("Visiting VarDeclaration\n");
         if (v.type)
@@ -168,7 +168,7 @@ package mixin template ParseVisitMethods(AST)
             s.elsebody.accept(this);
     }
 
-    void visitArgs(AST.Expressions* expressions, AST.Expression basis = null)
+    private void visitArgs(AST.Expressions* expressions, AST.Expression basis = null)
     {
         if (!expressions || !expressions.dim)
             return;
@@ -298,7 +298,7 @@ package mixin template ParseVisitMethods(AST)
             imp.accept(this);
     }
 
-    void visit(AST.Catch c)
+    private void visit(AST.Catch c)
     {
         //printf("Visiting Catch\n");
         if (c.type)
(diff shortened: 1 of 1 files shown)


WalterBright:
Looks good, except for the test failures :-(

kinke:
Yeah, because a Visitor example (src/examples/avg.d) already relies on a function being virtual. :/

ibuclaw:
Or make it `final`?

kinke:
That won't help with people wanting to override these functions. The new mixin is fine to avoid code duplication (in C++, you'd probably use multiple inheritance), it's just bad luck that the class it is mixed into has a C++ base class with virtual methods, which is required to trigger this long-slumbering segfault with host compilers < 2.078 [mscoff/MSVC targets only].

It's not crucial though; I'll keep this patch in LDC v1.7 so that the Windows version (with the 2.078 fix) can be bootstrapped with all DMD and LDC versions < 2.078.

wilzbach:
> Yeah, because a Visitor example (src/examples/avg.d) already relies on a function being virtual. :/

CC @RazvanN7

rainers:
> Or make it final?

That still checked whether the function overrides a base class function and crashed. The check is skipped for private, but I'm not sure whether that should be an error, too, i.e. to overload (not override) a virtual function with a private function.

RazvanN7:
Those methods should not be made private or final. It is essential for the dmd library to be able to override those methods. I suggest fixing the actual bug (which I understand is a MSVC specific check ) instead of creating workarounds which may damage the further development of the dmd compiler library.

kinke:
> I suggest fixing the actual bug (which I understand is a MSVC specific check ) instead of creating workarounds which may damage the further development of the dmd compiler library.

There is no proper way to fix this, as all D compilers < 2.078 feature this bug, so that 2.078 cannot be bootstrapped for MSVC using an older host compiler, especially not with 2.068, the last one not requiring a D compiler.
As I said, I'm fine with keeping this in LDC, but please make sure such bootstrapping issues don't happen again; next time there may be no easy workaround.

rainers:
> As I said, I'm fine with keeping this in LDC, but please make sure such bootstrapping regressions don't happen again; next time there may be no easy workaround.

We now have an Appveyor build for dmd that uses VC/COFF, too, both via the makefile (64-bit) and via VS (32-bit).

kinke:
Nice, that's a big relief.

PR #7642, opened 2018-01-08
[Trivial] Revise the fix for issue #18093

kinke:
Make sure to skip over template mixin parents etc., all the way up to the containing class, instead of returning null if the immediate parent isn't a class.

### DIFF
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -616,7 +616,7 @@ extern (C++) class FuncDeclaration : Declaration
      */
     final BaseClass* overrideInterface()
     {
-        if (ClassDeclaration cd = parent.isClassDeclaration())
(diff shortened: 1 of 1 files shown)


ibuclaw:
Can you provide a test?

WalterBright:
Yes, a test would be nice, but this was already not tested.

kinke:
AFAICT, `FuncDeclaration.overrideInterface()` is only used once for Windows/MSVC targets during semantic and once more for Windows name mangling, so a test would have been rather challenging.

rainers:
Thanks @kinke. I stumbled over this the other day, too. I wasn't sure whether toParent or toParent2 is appropriate here, but it now seems obvious ;-)

PR #7099, opened 2017-08-25
Enable automatic compiling of imports

marler8997:
An implementation of an idea I had (forum discussion here: http://forum.dlang.org/thread/tcrdpvqvwxffnewzohuj@forum.dlang.org)

I added a compiler option `-ci` which stands for "compile imports".

When this option is enabled, the compiler will automatically treat imported modules as if they were specified on the command line so long as it doesn't think they exist in any library.  For now it uses the same logic as rdmd to determine whether a module is in a library (see the `inALibrary` function), which is not full proof, but works well enough for a prototype.  With this change you should now be able to omit all the imported module files from the command line, i.e.

Instead of:
```
dmd prog.d -Isomelib somelib\foo\module1.d somelib\foo\module2.d somelib\foo\module3.d somelib\foo\module4.d somelib\foo\module5.d somelib\foo\module6.d -Ianotherlib anotherlib\bar\module1.d anotherlib\bar\module2.d anotherlib\bar\module3.d anotherlib\bar\module4.d anotherlib\bar\module5.d
```
Use:
```
dmd -ci prog.d -Isomelib -Ianotherlib
```

Should this feature be desirable, I recommend coming up with a more deterministic way of determining what modules exist in object/library files.

Also note that this feature would simplify rdmd and remove the need for it to invoke the compiler twice.  This would fix the current performance problem with rdmd and would have even better performance than the previous solution (https://github.com/dlang/tools/pull/194).

PR #7644, opened 2018-01-08
Move d_do_test.d combination into a function.

marler8997:
This is in preparation for #7099.  By moving the code for a single test combination into a function, the logic that drives all the combinations can be enhanced without having to increase the nesting of the combination code itself.

### DIFF
--- a/test/d_do_test.d
+++ b/test/d_do_test.d
@@ -570,9 +570,12 @@ int tryMain(string[] args)
 
     auto f = File(output_file, "a");
 
-    foreach (i, c; combinations(testArgs.permuteArgs))
+    enum Result { continue_, return0, return1 }
+
+    // Runs the test with a specific combination of arguments
+    Result testCombination(size_t permuteIndex, string permutedArgs)
     {
-        string test_app_dmd = test_app_dmd_base ~ to!string(i) ~ envData.exe;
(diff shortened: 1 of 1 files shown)


wilzbach (approved):
Looks great. Thanks!

test/d_do_test.d
```diff
@@ -668,7 +669,7 @@ int tryMain(string[] args)
                 }
                 else version (linux)
                 {
-                    auto script = test_app_dmd_base ~ to!string(i) ~ ".gdb";
+                    auto script = test_app_dmd_base ~ to!string(permuteIndex) ~ ".gdb";
```
wilzbach: You need to update the `version(Windows)` block above too.

wilzbach (dismissed):
Sorry - that was a bit hard to catch.

test/d_do_test.d
marler8997: done

wilzbach (approved):
I will wait for another pairs of eyes then.

stefan-koch-sociomantic:
that code is rather unclear due to the lack of documentation. 
If a function is doing something that is not trivial please share it in a sentence or two.

marler8997:
Ok I added a comment for the `testCombination` function.  Since this PR is fairly trivial and blocking #7099, could we get this merged sooner rather than later?

PR #7647, opened 2018-01-09
Fix Issue 18212 - Usage of cfloat,cdouble,cfloat,ifloat,idouble,ireal shouldn't trigger an error in deprecated code

wilzbach:
The following still triggers a deprecation warning:

```d
deprecated creal foo() { return 2 + 2i; }
```

In other words a `creal` in a FuncDeclaration doesn't check its function yet.
However, this is good enough to push the complex transition at Phobos as `auto`
works just fine.

See also: https://github.com/dlang/dmd/pull/7640

### DIFF
--- a/src/dmd/dscope.d
+++ b/src/dmd/dscope.d
@@ -807,4 +807,28 @@ struct Scope
         else
             return STRUCTALIGN_DEFAULT;
     }
+
+    /**********************************
+    * Checks whether the current scope (or any of its parents) is deprecated.
+    *
+    * Returns: `true` if this or any parent scope is deprecated, `false` otherwise`
+    */
+    extern(C++) bool isDeprecated()
+    {
+        for (Dsymbol sp = this.parent; sp; sp = sp.parent)
+        {
+            if (sp.isDeprecated())
+                return true;
+        }
+        for (Scope* sc2 = &this; sc2; sc2 = sc2.enclosing)
+        {
+            if (sc2.scopesym && sc2.scopesym.isDeprecated())
+                return true;
+
+            // If inside a StorageClassDeclaration that is deprecated
+            if (sc2.stc & STC.deprecated_)
+                return true;
+        }
+        return false;
+    }
 }
--- a/src/dmd/dsymbol.d
+++ b/src/dmd/dsymbol.d
@@ -310,20 +310,9 @@ extern (C++) class Dsymbol : RootObject
         if (global.params.useDeprecated != 1 && isDeprecated())
         {
             // Don't complain if we're inside a deprecated symbol's scope
-            for (Dsymbol sp = sc.parent; sp; sp = sp.parent)
-            {
-                if (sp.isDeprecated())
-                    return false;
-            }
-            for (Scope* sc2 = sc; sc2; sc2 = sc2.enclosing)
-            {
-                if (sc2.scopesym && sc2.scopesym.isDeprecated())
-                    return false;
+            if (sc.isDeprecated())
+                return false;
 
-                // If inside a StorageClassDeclaration that is deprecated
-                if (sc2.stc & STC.deprecated_)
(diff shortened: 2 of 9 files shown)


src/dmd/dsymbol.d
```diff
@@ -2146,3 +2135,31 @@ extern (C++) final class DsymbolTable : RootObject
...
+
+/**********************************
+* Checks whether the current scope (or any of its parents) is deprecated.
+*
+* Params:
+*     sc = scope to check for
+*
+* Returns: `true` if any parent scope is deprecated, `false` otherwise`
+*/
+final bool isInDeprecatedScope(Scope* sc)
```
JinShil: In general I prefer not to split things out into functions unless they have more than one caller, or the caller is excessively large.  Also, in general, I think this refactoring should be in a separate PR as it is not essential to the bug-fix at hand, and should be argued on its own individual merits.

JinShil (changes requested):

src/dmd/dsymbol.d
JinShil: Another problem with this method is it's modifying the public interface of the class, yet there is no public caller or potential caller.  If it is decided to keep this, it should probably be in its own private static function within the module.

wilzbach: There are two callers: `checkComplexTransition` and `checkDeprecated` - it just felt wrong to duplicate code that has a certain functionality.

JinShil: Gah! Sorry.  Totally missed that.  You're right on here.

src/dmd/mtype.d
```diff
@@ -3010,15 +3010,18 @@ extern (C++) abstract class Type : RootObject
      * Params:
      *  loc = The source location.
      */
-    final void checkComplexTransition(Loc loc)
+    final bool checkComplexTransition(Loc loc, Scope* sc)
```
JinShil: C++ interface change.  See mtype.h.

JinShil (changes requested):

src/dmd/dsymbol.d
JinShil: I believe this also needs to be reflected in the C++ interface dsymbo.h

wilzbach:
> I believe this also needs to be reflected in the C++ interface dsymbo.h

I thought we only need to add methods to the header files that are used by the LDC or GDC?
BTW I wasn't really sure where the right place to put is. After thinking a bit about it I moved it in `Scope` at `dscope.d`.

src/dmd/dscope.d
```diff
@@ -807,4 +807,28 @@ struct Scope
         else
             return STRUCTALIGN_DEFAULT;
     }
+
+    /**********************************
+    * Checks whether the current scope (or any of its parents) is deprecated.
+    *
+    * Returns: `true` if this or any parent scope is deprecated, `false` otherwise`
+    */
+    extern(D) bool isDeprecated()
```
JinShil: Isn't `extern(D)` the default?  I think that is redundant.

JinShil (dismissed):

JinShil:
>After thinking a bit about it I moved it in Scope at dscope.d.

Seems logical.

src/dmd/mtype.d
wilzbach: (done)

ibuclaw (approved):
Ok.

JinShil (approved):

wilzbach:
Somehow DAutoTest is under heavy stress recently and it hasn't picked up this PR :/
@CyberShadow is there anything easy that could be done to speed up DAutoTest? With our transition to SemaphoreCI, DAutoTest is now the slowest CI.
Cleaning up the Makefile hell at dlang.org will probably take a couple of weeks.

PR #7654, opened 2018-01-09
fix Issue 13742 - undefined reference to __coverage

MartinNowak:
- Nested template functions may reference coverage symbols of other modules.
  Those coverage symbols are either still `null` or already `symbol_reset`, in
  the former case coverage instrumentation is silently skipped, in the later
  case the linker will cause an error. This is because the current object file
  ends up with two `__coverage` symbols, one local BSS symbols and an undefined
  external symbol (for the other module's coverage).

- Fixed by mangling the symbols as `_D3pkg3mod__coverageZ` (similar to
  `_D3pkg3mod__ModuleInfoZ`) and making them global.

- Ideally the symbols would use -visibility=hidden, but that ELF backend feature
  isn't yet accessible from glue.d.

- This does not fix missing coverage when the referenced module hasn't yet
  been codegen'ed (e.g. because it follows on the command line or is part of
  another separate compilation). Using weak linkage to make this heuristically
  work would incurr some overhead for the coverage instrumentation.
  Unconditionally referencing coverage info OTOH might be too excessive,
  e.g. when the template function is in phobos, which would require to recompile
  phobos with -cov for successful linking.

### DIFF
--- a/test/runnable/test13742.sh
+++ b/test/runnable/test13742.sh
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+set -x # debug windows issues
+set -ueo pipefail
+
+src=runnable${SEP}extra-files
+dir=${RESULTS_DIR}${SEP}runnable
+output_file=${dir}/test13742.sh.out
+
+if [ $OS == "win32" -o  $OS == "win64" ]; then
+    LIBEXT=.lib
+    OBJ=.obj
+else
+    LIBEXT=.a
+    OBJ=.o
+fi
+
+$DMD -m${MODEL} -I${src} -lib -cov -of${dir}${SEP}test13742${LIBEXT} ${src}${SEP}lib13742a.d ${src}${SEP}lib13742b.d
+$DMD -m${MODEL} -I${src} -cov -of${dir}${SEP}test13742${EXE} ${src}${SEP}test13742.d ${dir}${SEP}test13742${LIBEXT}
+
+${RESULTS_DIR}/runnable/test13742${EXE} --DRT-covopt=dstpath:${dir}${SEP}
+
+rm ${RESULTS_DIR}/runnable/{runnable-extra-files-{lib13742a,lib13742b,test13742}.lst,test13742{${OBJ},${LIBEXT},${EXE}}}
+
+echo Success > ${output_file}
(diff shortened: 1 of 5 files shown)


rainers (approved):
LGTM. Fortunately, it's _D3pkg3mod10__coverageZ, not _D3pkg3mod__coverageZ.

test/runnable/test13742.sh
```diff
@@ -0,0 +1,20 @@
...
+if [ $OS == "win32" -o  $OS == "win64" ]; then
+	LIBEXT=.lib
+else
+	LIBEXT=.a
+fi
+
+$DMD -m${MODEL} -I${src} -lib -cov -of${dir}${SEP}lib13774${LIBEXT} ${src}${SEP}lib13742a.d ${src}${SEP}lib13742b.d
+$DMD -m${MODEL} -I${src} -cov -of${dir}${SEP}test13742${EXE} ${src}${SEP}test13742.d ${dir}${SEP}lib13774${LIBEXT}
+
+${dir}${SEP}test13742${EXE} --DRT-covopt=dstpath:${dir}${SEP}
```
wilzbach: Windows ain't happy about this:

```
 ... runnable/test13742.sh
./runnable/test13742.sh: line 18: test_results\runnable\test13742.exe: command not found
Makefile:197: recipe for target 'test_results/runnable/test13742.sh.out' failed
gmake[1]: *** [test_results/runnable/test13742.sh.out] Error 127
```

rainers:
Tried this locally:
- coverage options don't parse well if the path is absolute
- rt.cover crashes if it cannot create the file
- if I avoid both (as the auto-tester does) it works correctly for me

MartinNowak:
Nice, passing Windows, please ignore Travis-CI failures from gdb missing PTRACE capabilities.

MartinNowak:
Auto-merge toggled on

wilzbach:
> please ignore Travis-CI failures from gdb missing PTRACE capabilities.

:scream: 
FWIW SemaphoreCI uses Ubuntu 14.04 LTS, so gdb 7.7.1 is available (https://packages.ubuntu.com/trusty/gdb), so you might be lucky on your next `merge_stable`.

MartinNowak:
It's an access rights issue of the OS, not a problem with gdb. For some understandable reason they disallow PTRACE attach on some (most) Travis runs. Wouldn't be surprised if they slowly rolled out mitigations/fixes behind the scene.

MartinNowak:
Looks like this triggered or caused [18296 – [Reg2.078.1] invalid code with coverage and copy construction](https://issues.dlang.org/show_bug.cgi?id=18296).

PR #7658, opened 2018-01-10
[refactor] Remove ABSxxx aliases for enum

ibuclaw:


wilzbach (approved):

JinShil (approved):

marler8997:
Was curious what the reasons were behind this change.  You've removed the alias's and switched to using qualified enum values.

Removing the `alias`s reduces the number of symbols in the module's global namespace.  Is that one of the reasons?  Are we trying to minimize the number of symbols in our module's global namespaces?

Or maybe this change was just trying to make the enum style match conventions of the rest of the project?

wilzbach:
> Was curious what the reasons were behind this change. You've removed the alias's and switched to using qualified enum values.

It's about removing ugly, redundant relicts from the old C++ days.
Apart from the fact that D enum look a lot cleaner and are better to read, they also add more value thanks to D's great `final switch`, see e.g. https://github.com/dlang/dmd/pull/7661/files#r160580318

marler8997:
Gotcha, that makes sense.  For the sake of documentation I'll summarize what I understand to be the reasons.

This PR has 3 changes.  

1. Remove the redundant `ABS` prefix on the enum identifiers themselves (left over from the C++ days)
2. Remove the aliases which were likely only there because `ABS.ABS___` looked ugly.  Now that the redundant prefix is removed, `ABS.___` looks fine.  
3. Renamed `ABS` to `Abstract`...easier to grok.

ibuclaw:
All reasons given are side effects. Motivation is it makes conversion to C++ easier.

marler8997:
Sorry I'm not familiar. What conversion are you referring to? Now I'm glad I asked the question since I have a feeling I've missed hearing about a potentially big feature or change.

ibuclaw:
One of the zombie PRs that translates dmd sources to C++ headers is in the process of being revived - albeit as a standalone tool rather than new compiler switch.  Inconsistencies in how enums are declared is a problem for generating something that's compilable.

ibuclaw:
Other than that, no details. Don't think it's anything
too important.  ☺️

marler8997:
Interesting...

jacob-carlborg:
> Remove the aliases which were likely only there because ABS.ABS___ looked ugly. Now that the redundant prefix is removed, ABS.___ looks fine

I think the aliases were there originally for the backwards compatibility with the rest of the code.

ibuclaw:
The original code was an automated translation from C++ to D.

marler8997:
How ironic. Aliases added to support translating from C++, but now they are removed to facilitate translating the code back to C++ in header form :)

PR #7665, opened 2018-01-10
Fix Issue 18218 - __traits(isDeprecated, creal) should return true

wilzbach:
This is needed to be able to build Phobos with `-transition=complex`

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -502,6 +502,11 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
     }
     if (e.ident == Id.isDeprecated)
     {
+        if (global.params.vcomplex)
+        {
+            if (isTypeX(t => t.iscomplex() || t.isimaginary()))
+                return True();
+        }
         return isDsymX(t => t.isDeprecated());
     }
     if (e.ident == Id.isFuture)
--- a/test/compilable/sw_transition_complex.d
+++ b/test/compilable/sw_transition_complex.d
@@ -154,3 +154,11 @@ deprecated struct Foo
     ifloat a = 2i;
     cfloat b = 2f + 2i;
 }
+
+// https://issues.dlang.org/show_bug.cgi?id=18218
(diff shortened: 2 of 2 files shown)


src/dmd/traits.d
```diff
@@ -502,6 +502,11 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
     }
     if (e.ident == Id.isDeprecated)
     {
+        if (global.params.vcomplex)
+        {
+            if (isTypeX(t => t.iscomplex() || t.isimaginary()))
+                return True();
+        }
         return isDsymX(t => t.isDeprecated());
```
JinShil: Why doesn't `return isDsymX(t => t.isDeprecated());` return true?

wilzbach: Because `checkDeprecated` in `DSymbol` doesn't know about `checkComplexTransition` from `Type`. A `Type` also doesn't have the `isDeprecated` method, only `DSymbols` do. Most notably `AggregateDeclaration`.

JinShil: Ok, I looked closer at DMD's code and I don't see a straight-forward way to give `isDeprecated` semantics to `Type`, as it always seems to rely  on `Scope`.  Anyway, won't your proposed changes be removed when the feature is removed?  If so, please add one of your @@DEPRECATED@@ comments or something like that so we know it has an expiration date.

wilzbach: Well for as long as we support complex numbers, this will need to be true. Unfortunately no expiration date is known for that. When support for complex numbers is finally removed from the compiler the global parameter and the `iscomplex` function will go, so the compiler will tell you about this block here. Imho a lot better than a deprecated comment.
Though adding a comment doesn't hurt and I can so later.

JinShil: > Though adding a comment doesn't hurt and I can so later.

No need.  Thanks for clarifying.

JinShil (approved):

ibuclaw:
Ok

ibuclaw (approved):

PR #7666, opened 2018-01-10
Automatically convert PROTXXX into a D enum

wilzbach:
I wasn't sure if it should be named `PROTKIND` or just `PROT` -
there were some incosistencies.

```shell
replacements=(
"PROTundefined undefined"
"PROTnone none"
"PROTprivate private_"
"PROTpackage package_"
"PROTprotected protected_"
"PROTpublic public_"
"PROTexport export_"
)

    w=($r)
    sed "s/${w[0]}/Protection.${w[1]}/g" -i **/*.d
done
sed "s/PROTKIND/Protection/g" -i **/*.d
```

### DIFF
--- a/src/dmd/access.d
+++ b/src/dmd/access.d
@@ -53,9 +53,10 @@ private Prot getAccess(AggregateDeclaration ad, Dsymbol smember)
         {
             BaseClass* b = (*cd.baseclasses)[i];
             Prot access = getAccess(b.sym, smember);
-            switch (access.kind)
+            final switch (access.kind)
             {
             case Protection.none:
+            case Protection.undefined:
                 break;
             case Protection.private_:
                 access_ret = Prot(Protection.none); // private members of base class not accessible
@@ -71,8 +72,6 @@ private Prot getAccess(AggregateDeclaration ad, Dsymbol smember)
                 if (access_ret.isMoreRestrictiveThan(access))
                     access_ret = access;
                 break;
-            default:
-                assert(0);
             }
         }
     }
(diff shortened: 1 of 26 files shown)


ibuclaw:
Could name it PROT I guess.

jacob-carlborg:
Is `PROT` shortening of `Protection`? If that's the case I vote for `Protection`.

jacob-carlborg (changes requested):
I suggest renaming `PROTKIND` to `Protection` or `PROT`.

wilzbach:
> I suggest renaming PROTKIND to Protection or PROT.

Good idea! Done.

JinShil:
I think we should be careful with the naming conventions.  There is also a `struct Prot` which adds some semantics to the protection enum.  I think we should try to find a consistent convention distinguishing the classes/structs that add semantics to the concept represented by the enums.  

FWICT there are other enums (e.g. `BUILTIN`, `SCOPE`, `PASS`, etc...) that are all caps.  The convention is not strictly followed, but given the presence of `struct Prot`, I recommend going with `PROT` for consistency.  I can imagine a desire to rename `struct Prot` to `struct Protection`, so I'm looking for a consistent convention to follow to distinguish the `struct` from the `enum`.

jacob-carlborg:
> I can imagine a desire to rename struct Prot to struct Protection, so I'm looking for a consistent convention to follow to distinguish the struct from the enum.

Perhaps a nested enum:

```d
struct Protection
{
    enum Kind {}
}
```

ibuclaw:
There's StructFlags and ClassFlags, but not sure if that convention is worth repeating.

wilzbach:
> Perhaps a nested enum:

I made a new PR -> https://github.com/dlang/dmd/pull/7674

PR #7674, opened 2018-01-10
Automatically convert PROTXXX into a D enum (Prot.Kind)

wilzbach:
```d
replacements=(
"PROTundefined undefined"
"PROTnone none"
"PROTprivate private_"
"PROTpackage package_"
"PROTprotected protected_"
"PROTpublic public_"
"PROTexport export_"
)

    w=($r)
    sed "s/${w[0]}/Prot.Kind.${w[1]}/g" -i **/*.d
done
sed "s/PROTKIND/Prot.Kind/g" -i **/*.d
```

See https://github.com/dlang/dmd/pull/7666 for a discussion on why `Prot.Kind`
is the best choice.

In short:
- it avoids the conflicts of `Prot` vs. `PROT`
- it avoids the conflicts of potentially renaming `Prot` to `Protection`

Why `Prot.Kind` and not `Prot.KIND`?
------------------------------------

Because if we already do these conversion, we can also start following the
[_official DStyle_](https://dlang.org/dstyle.html#naming_classes):

> Classes, Interfaces, Structs, Unions, Enums, Non-Eponymous Templates
> The names of user-defined types should be PascalCased, which is the same as camelCased except that the first letter is uppercase.

### DIFF
--- a/src/dmd/dsymbol.d
+++ b/src/dmd/dsymbol.d
@@ -66,28 +66,21 @@ struct Ungag
     }
 }
 
-enum PROTKIND : int
-{
-    PROTundefined,
-    PROTnone,           // no access
-    PROTprivate,
-    PROTpackage,
-    PROTprotected,
-    PROTpublic,
-    PROTexport,
-}
-
-alias PROTundefined = PROTKIND.PROTundefined;
-alias PROTnone = PROTKIND.PROTnone;
-alias PROTprivate = PROTKIND.PROTprivate;
-alias PROTpackage = PROTKIND.PROTpackage;
-alias PROTprotected = PROTKIND.PROTprotected;
-alias PROTpublic = PROTKIND.PROTpublic;
-alias PROTexport = PROTKIND.PROTexport;
-
 struct Prot
 {
-    PROTKIND kind;
+    ///
+    enum Kind : int
+    {
+        undefined,
+        none,           // no access
+        private_,
+        package_,
+        protected_,
+        public_,
+        export_,
+    }
+
+    Kind kind;
     Package pkg;
 
     extern (D) this(PROTKIND kind)
--- a/src/dmd/dsymbol.d
+++ b/src/dmd/dsymbol.d
@@ -83,7 +83,7 @@ struct Prot
     Kind kind;
     Package pkg;
 
-    extern (D) this(PROTKIND kind)
+    extern (D) this(Prot.Kind kind)
     {
         this.kind = kind;
     }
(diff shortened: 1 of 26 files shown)


src/dmd/dsymbol.d
```diff
@@ -66,31 +66,24 @@ struct Ungag
...
-alias PROTpackage = PROTKIND.PROTpackage;
-alias PROTprotected = PROTKIND.PROTprotected;
-alias PROTpublic = PROTKIND.PROTpublic;
-alias PROTexport = PROTKIND.PROTexport;
-
 struct Prot
 {
-    PROTKIND kind;
+    ///
+    enum Kind : int
```
klickverbot: Should probably have a comment with the non-obvious C++ mapping explained (same in the header file).

wilzbach: Would updating the C++ header to match it work too? CC @ibuclaw

src/dmd/dsymbol.d
```diff
@@ -66,31 +66,24 @@ struct Ungag
...
+    {
+        undefined,
+        none,           // no access
+        private_,
+        package_,
+        protected_,
+        public_,
+        export_,
```
jacob-carlborg: Documentation of the enum members would be nice, at least empty Ddoc comments so they show up in the generated documentation.

jacob-carlborg (approved):

jacob-carlborg:
I'll leave the C++ headers to Iain and David.

src/dmd/dsymbol.h
```diff
@@ -94,20 +94,19 @@ void semantic3(Dsymbol *dsym, Scope* sc);
...
-    PROTpublic,
-    PROTexport,
-};
-
 struct Prot
 {
-    PROTKIND kind;
+    enum Kind
+    {
+        PROTundefined,
```
iain-buclaw-sociomantic: Remove `PROT` prefix.

src/dmd/dsymbol.d
iain-buclaw-sociomantic: You need to update the header file to match because putting it here changes mangling of the type.

iain-buclaw-sociomantic: What I see so far in this PR is fine, however.

src/dmd/dsymbol.h
JinShil: @wilzbach I think this PR is ready to go, with the exception of this change.

wilzbach: OK done.

wilzbach:
> @wilzbach I think this PR is ready to go, with the exception of this change.

I think the LDC devs haven't explicitly said that they are okay.
CC @klickverbot @kinke

ibuclaw (approved):

ibuclaw:
I don't consider anything here that requires explicit confirmation here.  If either one of LDC or GDC maintainers say :+1: then its fine.

PR #7676, opened 2018-01-10
Enable automatic compiling of imports (DMD ONLY)

marler8997:
Based on concerns from @ibuclaw this is a refactor of #7099 which keeps the features out of the front end and moves it to the driver by using a callback function.

### DIFF
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -31,7 +31,6 @@ import dmd.globals;
 import dmd.id;
 import dmd.identifier;
 import dmd.parse;
-import dmd.root.array;
 import dmd.root.file;
 import dmd.root.filename;
 import dmd.root.outbuffer;
@@ -302,6 +301,12 @@ extern (C++) final class Module : Package
     extern (C++) static __gshared Dsymbols deferred2;   // deferred Dsymbol's needing semantic2() run on them
     extern (C++) static __gshared Dsymbols deferred3;   // deferred Dsymbol's needing semantic3() run on them
     extern (C++) static __gshared uint dprogress;       // progress resolving the deferred list
+    /**
+     * A callback function that is called once an imported module is
+     * parsed. If the callback returns true, then it tells the
+     * frontend that the driver intends on compiling the import.
+     */
+    extern (C++) static __gshared bool function(Module mod) onImport;
 
     static void _init()
     {
@@ -380,14 +385,14 @@ extern (C++) final class Module : Package
     Dsymbol searchCacheSymbol;  // cached value of search
     int searchCacheFlags;       // cached flags
 
-    // module from command line we're imported from,
-    // i.e. a module that will be taken all the
-    // way to an object file
+    /**
+     * A root module is one that will be compiled all the way to
+     * object code.  This field holds the root module that caused
+     * this module to be loaded.  If this module is a root module,
+     * then it will be set to $(D this).  This is used to determine
+     * ownership of template instantiation.
+     */
     Module importedFrom;
-
-    // indicates this module should be compiled even though
-    // it wasn't given on the command line
-    bool isCompiledImport;
 
     Dsymbols* decldefs;         // top level declarations for this Module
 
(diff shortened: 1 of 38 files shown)


src/dmd/dmodule.d
```diff
@@ -379,9 +380,11 @@ extern (C++) final class Module : Package
     Dsymbol searchCacheSymbol;  // cached value of search
     int searchCacheFlags;       // cached flags
 
-    // module from command line we're imported from,
-    // i.e. a module that will be taken all the
-    // way to an object file
+    // A root module is one that will be compiled all the way to
+    // object code.  This field holds the root module that caused
+    // this module to be loaded.  If this module is a root module,
+    // then it will be set to $(D this).  This is used to determine
```
klickverbot: Stylistic nitpick: I'd just use single backticks instead of `$(D …)`, especially in non-DDoc comments.

marler8997: ok

src/dmd/dmodule.d
```diff
@@ -301,6 +301,7 @@ extern (C++) final class Module : Package
     extern (C++) static __gshared Dsymbols deferred2;   // deferred Dsymbol's needing semantic2() run on them
     extern (C++) static __gshared Dsymbols deferred3;   // deferred Dsymbol's needing semantic3() run on them
     extern (C++) static __gshared uint dprogress;       // progress resolving the deferred list
+    extern (C++) static __gshared bool function(Module mod) onImport; // callback function when module is loaded
```
klickverbot: Nitpick: What precisely does "loaded" mean here? (Obvious with the context in this PR, of course, but perhaps not to somebody looking at this later on.)

marler8997: how about `// callback function once the module is parsed` ?

src/dmd/dmodule.d
```diff
@@ -529,6 +532,16 @@ extern (C++) final class Module : Package
             fprintf(global.stdmsg, "%s\t(%s)\n", ident.toChars(), m.srcfile.toChars());
         }
         m = m.parse();
+
+        if(onImport !is null)
```
klickverbot: Nitpick: Space between `if` and `(` – it would be more common to read `if (onImport)` in DMD.

marler8997: ugh, I always forget that one

src/dmd/dmodule.d
```diff
@@ -529,6 +532,16 @@ extern (C++) final class Module : Package
...
+        {
+            if(!m.isRoot && onImport(m))
+            {
+                m.importedFrom = m;
+                assert(m.isRoot);
```
klickverbot: Nitpick: Parentheses for function calls are rarely (if ever) omitted in the rest of DMD, especially with all the public member variables.

src/dmd/dmodule.d
klickverbot: Documenting the return value would also be useful.

src/dmd/dmodule.d
```diff
@@ -301,6 +301,9 @@ extern (C++) final class Module : Package
     extern (C++) static __gshared Dsymbols deferred2;   // deferred Dsymbol's needing semantic2() run on them
     extern (C++) static __gshared Dsymbols deferred3;   // deferred Dsymbol's needing semantic3() run on them
     extern (C++) static __gshared uint dprogress;       // progress resolving the deferred list
+    extern (C++) static __gshared bool function(Module mod) onImport; // callback function called once an imported module is
```
jacob-carlborg: All new public symbols should have Ddoc comments.

src/dmd/dmodule.d
```diff
@@ -379,9 +382,11 @@ extern (C++) final class Module : Package
...
     int searchCacheFlags;       // cached flags
 
-    // module from command line we're imported from,
-    // i.e. a module that will be taken all the
-    // way to an object file
+    // A root module is one that will be compiled all the way to
+    // object code.  This field holds the root module that caused
+    // this module to be loaded.  If this module is a root module,
+    // then it will be set to `this`.  This is used to determine
+    // ownership of template instantiation.
```
jacob-carlborg: Why not turn this into a Ddoc comment while modifying it.

jacob-carlborg (changes requested):

marler8997:
I've decided to select this version of the PR in favor of the current state of #7099.  I'm going to close this PR and push the branch to that one.

timotheecour:
* That comment is confusing since you seem to have done the opposite (close 7676 and keep 7099 open)
* with the path you choose, will all compilers (including ldc) benefit from `-i` or just dmd ?

marler8997:
I figured it would be better to keep the old PR open so it keeps the whole history of the PR that actually gets merged.

As for LDC/GDC, it's up to them if they want to support the feature or not.  If they do, the pattern matching code can be moved into a shared module and they can set the callback function accordingly.  However, currently only DMD will be using it so it makes sense to keep it in the DMD specific module for now.  If an LDC or GDC maintainer indicates they's like to use it, I can put the code in a shared module now.

marler8997:
Also note that the way the solution is now designed, the interface is just a callback function.  So the drivers for LDC and GDC could compile imports using whatever mechanism they choose. Like DMD they could supply a list of patterns to include/exclude modules, or they could use another mechanism.  Or they could just leave the callback function set to null which means no imports will be included in the compilation.

timotheecour:
> As for LDC/GDC, it's up to them if they want to support the feature or not. If they do, the pattern matching code can be moved into a shared module and they can set the callback function accordingly. However, currently only DMD will be using it so it makes sense to keep it in the DMD specific module for now. If an LDC or GDC maintainer indicates they would like to use it, I can put the code in a shared module now.

@klickverbot I see no reason to restrict this feature to dmd and not ldc; should we move the code in a shared module? which one do you suggest?

PR #4351, opened 2015-01-27
Add Compile-Time Error-Checking and Avoid Boundschecking in Slice Expressions

nordlow:
The goal here is to avoid boundschecking for slice expression whose lower or/and upper bounds can never be outside of `[0 .. $]`. Further, if a slice index can be proven to be outside of bounds an error is issued.

_Step 1_:

``` D
x[0 .. $]
x[0 .. 0]
x[$ .. $]
```

_Step 2_:

``` D
x[0 .. $/n]
x[$/n .. $]
```

where `n` is a compile-time constant >= 1

_Step 3_:

``` D
x[$/m .. $/n]
```

where `m` and `n` are compile-time constants both >= 1 and `m >= n`.

_Step 4_:

``` D
x[$*p/q .. $*r/s]
```

where `p`, `q`, `r` and `s` are compile-time constants both >= 1 and `p/q <= r/s`, `p <= q`, `r <= s`.

_Step 5 (error reporting only)_:

``` D
x[$+o .. $+s]
```

I put this out in the open in a very early stage to get feedback as soon as possible. This so I don't diverge in a direction that you dislike.

I use a warning for now for debug purpose.

My prel tests seems to show that the current logic holds water and at least doesn't crash DMD :)

Most of my questions are given as TODO comments.

Specifically I'm curious
- Why isn't there a specific `class DollarExp` instead of `VarExp`. That would have made `isOpDollar` trivial.
- If `(x->op == TOKdiv)` is it guaranteed that `x` is of type `DivExp`? This seems like an insecure pattern. I see that DMD has `isVarDeclaration` so why isn't there corresponding `is.*Exp`?
- Could any explain what resolveOpDollar is used for. Have any use of it here?

Note that the naming of the variable `lowerIsLessThanUpper` is misleading as it's value is true iff `lwr <= upr`. `lowerIsLessThanOrEqualToUpper` or `lowerIsLTEToUpper` would be a better name.

Update: Another key issue is whether this should include analysis of immutable slice initializations such as, for instance, line 37 to to 46 in std.ascii.

Destroy and thanks in advance!

PR #4293, opened 2015-01-13
Issue 13976 - Value range propagation to disable some slice bound tests

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=13976

Example:

``` d
void main()
{
    int[10] a;
    size_t n;
    auto s = a[n%3 .. n%3 + 3];
    assert(s.length == 3);
}
```

With 2.066:

``` iasm
c:\d\test.d:1 void main()
00402010: c8380000                enter 0x38, 0x0
00402014: 53                      push ebx
00402015: 57                      push edi
c:\d\test.d:3     int[10] a;
00402016: b90a000000              mov ecx, 0xa
0040201b: 31c0                    xor eax, eax
0040201d: 8d7dc8                  lea edi, [ebp-0x38]
00402020: f3ab                    rep stosd
c:\d\test.d:4     size_t n;
00402022: 8945f0                  mov [ebp-0x10], eax
c:\d\test.d:5     auto s = a[n%3 .. n%3 + 3];
00402025: b103                    mov cl, 0x3
00402027: 31d2                    xor edx, edx
00402029: f7f1                    div ecx
0040202b: 8d5a03                  lea ebx, [edx+0x3]
0040202e: 83fb0a                  cmp ebx, 0xa
00402031: 7704                    ja 0x402037   D main c:\d\test.d:5
00402033: 39d3                    cmp ebx, edx
00402035: 730a                    jae 0x402041  D main c:\d\test.d:5
00402037: b805000000              mov eax, 0x5
0040203c: e823000000              call 0x402064 test.__array c:\d\test.d:7
00402041: 2bda                    sub ebx, edx
00402043: 8d5495c8                lea edx, [ebp+edx*4-0x38]
00402047: 895df8                  mov [ebp-0x8], ebx
0040204a: 8955fc                  mov [ebp-0x4], edx
c:\d\test.d:6     assert(s.length == 3);
0040204d: 394df8                  cmp [ebp-0x8], ecx
00402050: 740a                    jz 0x40205c   D main c:\d\test.d:6
00402052: b806000000              mov eax, 0x6
00402057: e824000000              call 0x402080 test.__assert c:\d\test.d:7
0040205c: 31c0                    xor eax, eax
c:\d\test.d:7 }
0040205e: 5f                      pop edi
0040205f: 5b                      pop ebx
00402060: c9                      leave
00402061: c3                      ret
```

With this PR:

``` iasm
c:\d\test.d:1 void main()
00402010: 55                      push ebp
[...]

PR #7677, opened 2018-01-11
Use some more value range propagation to elide  array bound tests.

ibuclaw:
Extends #4293, and inspired by #4351.

I will add tests later, just going to stand back and think about whether or not I'm missing something really obvious and any approach used here is wrong.

The additional branches match the following code:

---

1. `auto x = arr[n .. 0]`
Upper always in bounds if is proven to be `0` at compile time.

---

2. `auto x = arr[n .. $]`
Upper is always in bounds if is just a dollar var.

---

3. `auto x = arr[n .. $/m]`
Upper is using `$`.  Check whether `IntRange(upr).max < IntRange($).max`.
If `m` is a compile-time constant, this would affect the inferred range.
For example, on 64bit:
```
// The length.max integer range is 18446744073709551615.

// These slices don't generate a bounds check.
[n .. $/4] => uprRange.max is 4611686018427387903
[n .. $%4] => uprRange.max is 3

// Bounds check still occurs here.
[n .. 4/$] => uprRange.max is 18446744073709551615
```

---

4. `auto x = arr[$/m .. $]`
Both lower and upper are using `$`.  Check whether `IntRange(lwr).max < IntRange(upr).max`.
If `m` is a compile-time constant, this would affect the inferred int range.
For example, on 64bit:
```
// These slices don't generate a bounds check.
[$/4 .. $] => lwrRange.max is 4611686018427387903, uprRange.max is 18446744073709551615
[$%4 .. $/7] => lwrRange.max is 3, uprRange.max is 2635249153387078802

// Bounds check still occurs here.
[$*10/20 .. $*30/40] => lwrRange.max is 922337203685477580, uprRange.max is 461168601842738790
```

---

In the cases for 1,2,3, it sets `upperIsInBounds = true`.
In the case for 4, it sets `lowerIsLessThanUpper = true`.

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -18,6 +18,7 @@ import core.stdc.string;
 import dmd.access;
 import dmd.aggregate;
 import dmd.aliasthis;
+import dmd.apply;
 import dmd.arrayop;
 import dmd.arraytypes;
 import dmd.attrib;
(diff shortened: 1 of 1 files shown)


nordlow:
Thanks for taking over this!

ibuclaw:
On some thought.

Case 1 and 2 are just a front-end optimization that the backend optimizer should be able to handle (gdc doesn't produce a bounds check during codegen).

Case 3 can't be validated using VRP alone, especially when the expression gets complex.  Consider `$*5/4` where `arr.length = 4` (but we don't know that at compile time).

Case 4 can't be validated at all!  If arr.length is `0` then lower is _never_ less than upper.

So overall, while a nice idea that _seems_ feasible.  Unless _everything_ is known at compile-time, the worst case scenario can only be assumed.

@nordlow ^^

On some inspection though, I see that there's a missed opportunity to correctly set these two fields correctly, as the `lengthVar` is not determined until the const folder runs.

Haven't checked if dmd does array bounds where it shouldn't because of this, but will sort that out later.

ibuclaw:
Maybe if VRP detected also if an overflow could happen.  Then we might have something workable here.

ibuclaw:
I'll have a look into whether an overflow flag could be added to the vrp code.

PR #7683, opened 2018-01-12
highlight mtype.d error messages

JinShil:


src/dmd/mtype.d
```diff
@@ -4683,13 +4683,13 @@ extern (C++) final class TypeSArray : TypeArray
...
             {
-                e.error("%s is not an expression", e.toChars());
+                e.error("`%s` is not an expression", e.toChars());
                 return new ErrorExp();
             }
             else if (!(flag & DotExpFlag.noDeref) && sc.func && !sc.intypeof && sc.func.setUnsafe())
             {
                 // MAINTENANCE: turn into error in 2.073
-                e.deprecation("%s.ptr cannot be used in @safe code, use &%s[0] instead", e.toChars(), e.toChars());
+                e.deprecation("`%s.ptr` cannot be used in `@safe` code, use `&%s[0]` instead", e.toChars(), e.toChars());
```
wilzbach: Hehe that's why we should maintain a file with a list of deprecations :scream: 

I will submit a PR for this then.

wilzbach (approved):

src/dmd/mtype.d
JinShil: huh.  I totally overlooked the comment.  I guess I had tunnel vision on tick marks.

wilzbach: https://github.com/dlang/dmd/pull/7688

PR #7685, opened 2018-01-12
Cache D installer download on SemaphoreCI

wilzbach:
code.dlang.org is often down and there's no mirror for
http://code.dlang.org/download/LATEST yet

See also:
- https://github.com/dlang/dub/issues/1290
- https://github.com/dlang/installer/pull/273
- https://github.com/dlang/installer/pull/291

CC @ibuclaw

### DIFF
--- a/semaphoreci.sh
+++ b/semaphoreci.sh
@@ -22,6 +22,14 @@ export FULL_BUILD="${PULL_REQUEST_NUMBER+false}"
 source ci.sh
 
 ################################################################################
+# Cache the compiler installation on Semaphore
+# See also: https://semaphoreci.com/docs/caching-between-builds.html#additional-dir-caching
+################################################################################
+
+mkdir -p "$SEMAPHORE_CACHE_DIR/dlang"
+ln -s "$SEMAPHORE_CACHE_DIR/dlang" "$HOME/dlang"
+
+################################################################################
 # Always source a DMD instance
 ################################################################################
 


ibuclaw:
Hmm, I'd consider it safe to cache the installation script and tarballs, and check for their existence before downloading.  I wouldn't be so sure about extracted files though.

wilzbach:
Hmm, I still see this failure very often for the gdc builds:

```
++ CURL_USER_AGENT='DMD-CI curl 7.35.0 (x86_64-pc-linux-gnu) libcurl/7.35.0 OpenSSL/1.0.1f zlib/1.2.8 libidn/1.28 librtmp/2.3'
++ bash install.sh gdc --activate
curl: (28) Operation too slow. Less than 1024 bytes/sec transferred the last 30 seconds
curl: (28) Operation too slow. Less than 1024 bytes/sec transferred the last 30 seconds
curl: (28) Operation too slow. Less than 1024 bytes/sec transferred the last 30 seconds
curl: (28) Operation too slow. Less than 1024 bytes/sec transferred the last 30 seconds
curl: (28) Operation too slow. Less than 1024 bytes/sec transferred the last 30 seconds
Failed to download 'http://code.dlang.org/download/LATEST'
```

Fast-failing is activated, hence the last build fails too.

Note that since https://github.com/dlang/installer/pull/218, we don't download `dub` automatically, but use the bundled version if it's provided (DMD + LDC do).
Well, Martin wasn't a fan of dlang/installer#273, but we could do sth. similar and simply fetch dub 1.6.0 from GitHub for gdc builds.

wilzbach:
Downloading the install script is quite reliable with our four mirrors.
As we don't even need the `dub` binary, I think https://github.com/dlang/dmd/pull/7692 is the best way to move forward.

PR #7686, opened 2018-01-12
fix issue 18233 - building with -m64 doesn't work with sc.ini from zip

rainers:
… distribution and VS2017

These entries are auto-detected and should be removed.

It's especially LINKCMD that doesn't work with VS2017.

### DIFF
--- a/ini/windows/bin/sc.ini
+++ b/ini/windows/bin/sc.ini
@@ -17,124 +17,12 @@ LINKCMD=%@P%\link.exe
 
 
 [Environment64]
-LIB="%@P%\..\lib64"
-
+LIB=%@P%\..\lib64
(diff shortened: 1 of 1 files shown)


wilzbach:
```
Root package assert_writeln_magic references unknown package libdparse
posix.mak:818: recipe for target '.generated/assert_writeln_magic' failed
```

FYI: this is due to code.dlang.org's instability. The upgrade to 2.077.1 is already in master - the only way to restart DAutotest is by new commits or force-pushing ...

wilzbach (approved):
Noice. It's finally getting similar to dmd.conf

wilzbach:
>  the only way to restart DAutotest is by new commits or force-pushing ...

FYI: I just did so.

PR #7687, opened 2018-01-12
Fix issue 18232: local vars in union member functions needs initializer.

quickfur:
The `sc.isunion` condition is too broad, because it also excludes local variables declared inside union member functions. Moreoever, it is not necessary, because the `fd` condition later on already excludes union fields, since they would not be inside a function declaration.


JinShil (approved):

wilzbach:
Seems like "old" PRs (i.e. since the introduction of the hook) need to be closed and re-opened once for SemaphoreCI to catch up.

wilzbach (approved):

PR #7688, opened 2018-01-12
Turn .ptr deprecation into an error

wilzbach:
This was scheduled to happen in 2.073 - we really need a file or wiki to list
these things, s.t. they aren't forgotten.

I'm abusing auto-tester for testing as due to `-fPIC` I still
can't run the testsuite locally.

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -4688,9 +4688,8 @@ extern (C++) final class TypeSArray : TypeArray
             }
             else if (!(flag & DotExpFlag.noDeref) && sc.func && !sc.intypeof && sc.func.setUnsafe())
             {
-                // MAINTENANCE: turn into error in 2.073
-                e.deprecation("`%s.ptr` cannot be used in `@safe` code, use `&%s[0]` instead", e.toChars(), e.toChars());
-                // return new ErrorExp();
+                e.error("`%s.ptr` cannot be used in `@safe` code, use `&%s[0]` instead", e.toChars(), e.toChars());
+                return new ErrorExp();
             }
             e = e.castTo(sc, e.type.nextOf().pointerTo());
         }
@@ -4950,9 +4949,8 @@ extern (C++) final class TypeDArray : TypeArray
         {
             if (!(flag & DotExpFlag.noDeref) && sc.func && !sc.intypeof && sc.func.setUnsafe())
             {
-                // MAINTENANCE: turn into error in 2.073
-                e.deprecation("`%s.ptr` cannot be used in `@safe` code, use `&%s[0]` instead", e.toChars(), e.toChars());
-                // return new ErrorExp();
+                e.error("`%s.ptr` cannot be used in `@safe` code, use `&%s[0]` instead", e.toChars(), e.toChars());
+                    return new ErrorExp();
             }
             e = e.castTo(sc, next.pointerTo());
             return e;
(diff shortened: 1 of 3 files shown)


quickfur:
LGTM

ibuclaw (approved):

ibuclaw:
_finger hovers on merge button_

JinShil (dismissed):

JinShil:
Why is this labeled "WIP"?

ibuclaw:
I assume because @wilzbach can't test locally, so using ci to tell him/us what breaks with this change.

I'll block on missing changelog anyhow.

ibuclaw (changes requested):
Changelog entry

JinShil:
Curious that this didn't require a change to any test cases, yet coverage shows it's covered. 🤔

wilzbach:
> I assume because @wilzbach can't test locally, so using ci to tell him/us what breaks with this change.

Yeah, I often patch the test suite, but sometimes I'm in a hurry.
(reminder to myself: I should really look into getting the `-fPIC` PRs merged)

> Curious that this didn't require a change to any test cases, yet coverage shows it's covered. :thinking:

Found the test.

> I'll block on missing changelog anyhow.

Added.
(Just in case someone didn't know this, the rendered changelog can be previewed at DAutoTest.)

ibuclaw (approved):

jacob-carlborg:
`.ptr` can still be used in `@system` code? If that's the case, is it worth mentioning explicitly in the changelog?

wilzbach:
>  If that's the case, is it worth mentioning explicitly in the changelog?

I didn't judge it as important enough, because it's more or less obvious, but that you are wondering about it, is reason enough to include it (I just did so).

test/fail_compilation/test11176.d
```diff
@@ -1,5 +1,5 @@
 /*
-REQUIRED_ARGS: -de
+REQUIRED_ARGS:
 ---
 fail_compilation/test11176.d(12): Deprecation: b.ptr cannot be used in @safe code, use &b[0] instead
 fail_compilation/test11176.d(16): Deprecation: b.ptr cannot be used in @safe code, use &b[0] instead
```
JinShil: This isn't making any sense to me.  Why are the tests passing?  The output of this test should be
```
Error: `b.ptr` cannot be used in `@safe` code, use `&b[0]` instead
```

wilzbach: Excellent catch - the test helped me to realize that there was another Deprecation message that needed maintenance (i.e. the one for dynamic arrays).

ibuclaw:
I'm going to say all good to go.  @JinShil ?

src/dmd/mtype.d
```diff
@@ -4950,9 +4949,8 @@ extern (C++) final class TypeDArray : TypeArray
         {
             if (!(flag & DotExpFlag.noDeref) && sc.func && !sc.intypeof && sc.func.setUnsafe())
             {
-                // MAINTENANCE: turn into error in 2.073
-                e.deprecation("`%s.ptr` cannot be used in `@safe` code, use `&%s[0]` instead", e.toChars(), e.toChars());
-                // return new ErrorExp();
+                e.error("`%s.ptr` cannot be used in `@safe` code, use `&%s[0]` instead", e.toChars(), e.toChars());
+                 return new ErrorExp();
```
ibuclaw: Funny whitespace

ibuclaw (changes requested):

JinShil (approved):
Just waiting on whitespace fix.  I really need to figure out how to do such changes as part of reviewing/merging.

JinShil:
Fixed whitespace with `pr_push`.  Thanks @wilzbach!

PR #5626, opened 2016-04-03
Issue 7625 - inlining only works with explicit else branch

9rnsr:
Consider subsequent statements of `IfStatement`.

PR #7690, opened 2018-01-12
fix Issue 7625 - inlining only works with explicit else branch

ibuclaw:
**UPDATE**: According to digger bisect, Issue 7625 got fixed by https://github.com/dlang/dmd/pull/11236 - this PR now introduces test cases only.

#5626 stalled because the new logic added to the testsuite was rejected.

This only sets `REQUIRED_ARGS: -inline`.  I tested dmd, and without this patch, and it fails to compile with:
```
$ dmd -inline test/runnable/inline7625.d 
test/runnable/inline7625.d(65): Error: function inline7625.tembo cannot inline function
test/runnable/inline7625.d(117): Error: function inline7625.inlineMe15483b cannot inline function
```

### DIFF
--- a/test/runnable/inline7625.d
+++ b/test/runnable/inline7625.d
@@ -1,0 +1,185 @@
+// REQUIRED_ARGS: -inline
+
+/***************************************************/
+
+pragma(inline, true)
+{
+    int foo1(int v)
+    {
+        return bar1(2 * v);
+    }
+
+    int bar1(int a)
+    {
+        if (a > 0)
+            return 1;
+        else
+            return baz1(a);
+    }
+}
+
+int baz1(int a)
+{
+    if (a > 0)
+        throw new Exception("a > 0");
+    return a - 1;
+}
+
+// ---
+
+pragma(inline, true)
+{
+    int foo2(int v)
+    {
+        return bar2(2 * v);
+    }
+
+    int bar2(int a)
+    {
+        if (a > 0)
+            return 1;
+        // else
+            return baz2(a);
+    }
+}
+
+int baz2(int a)
+{
+    if (a > 0)
+        throw new Exception("a > 0");
+    return a - 1;
+}
+
+// ---
+
+pragma(inline, true)
+{
+    int foo3(int v)
+    {
+        return bar3(2 * v);
+    }
+
+    int bar3(int a)
+    {
+        if (a > 0)
+            a = 1;
+        else
+            return baz3(a);
+        return a;
+    }
+}
+
+int baz3(int a)
+{
+    if (a > 0)
+        throw new Exception("a > 0");
+    return a - 1;
+}
+
+void test7625a()
+{
+    assert(foo1(1) ==  1);
+    assert(foo1(0) == -1);
+    assert(foo2(1) ==  1);
+    assert(foo2(0) == -1);
+    assert(foo3(1) ==  1);
+    assert(foo3(0) == -1);
+}
+
+/***************************************************/
+
+@safe pragma(inline, true)
+{
+    int tembo(int x, int y)
+    {
+        if (y == 0)
+            return 0;
+        x++;
+        return x / y;
+    }
+    int pembo(int x, int y)
+    {
+        if (y == 0)
+            return 0;
+        else
+        {
+            x++;
+            return x / y;
+        }
+    }
+
+    int twiga(int x, int y, int z)
+    {
+        auto w = tembo(x, y);
+        return w * z;
+    }
+
+    int simba(int x, int y, int z)
+    {
+        auto w = pembo(x, y);
+        return w * z;
+    }
+}
+
+void test7625b()
+{
+    assert(twiga(5, 3, 4) == 8);
+    assert(twiga(5, 0, 4) == 0);
+
+    assert(simba(5, 3, 4) == 8);
+    assert(simba(5, 0, 4) == 0);
+}
+
+/***************************************************/
+
+@safe pragma(inline, true)
+{
+    bool inlineMe15483a(bool left)
+    {
+        if (left)
+            return true;
+
+        return false;
+    }
+
+    bool inlineMe15483b(bool left)
+    {
+        if (left)
+            return true;
+
+        static if (false)
+        {
+            /* https://issues.dlang.org/show_bug.cgi?id=15483
+               Even though it does absolutely nothing,
+               the mere presence of this block prevents inlining
+               of this function.
+            */
+        }
+
+        return false;
+    }
+}
+
+int foo15483()
+{
+    auto r1 = inlineMe15483a(true);
+    auto r2 = inlineMe15483b(true); // OK <- NG
+
+    return 128;
+}
+
+void test15483()
+{
+    // Prevent inlining of test function call.
+    auto fp = &foo15483;
+    assert(fp() == 128);
+}
+
+/***************************************************/
+
+void main()
+{
+    test7625a();
+    test7625b();
+    test15483();
+}


test/runnable/inline7625.d
```diff
@@ -0,0 +1,150 @@
...
+        if (left)
+            return true;
+
+        static if (false)
+        {
+            /*
+                Even though it does absolutely nothing,
+                the mere presence of this block prevents inlining
+                of this function.
+            */
```
JinShil: Hmm.  Is that by design?

ibuclaw: This is the bug in question being tested.

ibuclaw: Or at least one aspect of it if I understand right.

ibuclaw:
Added link to bugzilla issue in noted comment.

test/runnable/inline7625.d
JinShil: IMO this needs a better comment.  It should read something to the effect of "Ensure that the presence of a `static if` block does not prevent inlining.  See {insert bug URL}."  I understand that the original comment was to clarify the problem, but now we need to clarify what we're testing.

test/runnable/inline7625.d
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        return false;
+    }
+}
+
+int foo15483()
+{
+    auto r1 = inlineMe15483a(true);
+    auto r2 = inlineMe15483b(true); // OK <- NG
```
JinShil: This also needs a better comment.  What does "OK <- NG" mean?

src/dmd/inline.d
```diff
@@ -226,25 +216,54 @@ public:
     }
 
     override void visit(IfStatement s)
+    {
+        visitIfStatement(s, null);
+    }
+
+    extern (D) private void visitIfStatement(IfStatement s, void delegate(DoInlineAs!Result) visitSubseq)
```
JinShil: Does this really need to be declared `extern(D)`?

src/dmd/inlinecost.d
```diff
@@ -206,6 +204,11 @@ public:
     }
 
     override void visit(IfStatement s)
+    {
+        visitIfStatement(s, null);
+    }
+
+    extern (D) private void visitIfStatement(IfStatement s, void delegate(InlineCostVisitor) visitSubseq)
```
JinShil: Same here.  Why the `extern (D)` attribution?

src/dmd/inline.d
ibuclaw: Yes, because C++ does not have delegates.

test/runnable/inline7625.d
ibuclaw: Was "No Good", now "OK"

src/dmd/inline.d
```diff
@@ -123,69 +123,59 @@ public:
     override void visit(CompoundStatement s)
     {
         //printf("CompoundStatement.doInlineAs!%s() %d\n", Result.stringof.ptr, s.statements.dim);
+        visitCompoundStatement(s, 0);
+    }
+
+    private void visitCompoundStatement(CompoundStatement s, size_t n)
```
WalterBright: Document what `n` is.

src/dmd/inline.d
```diff
@@ -123,69 +123,59 @@ public:
...
-                    assert(econd);
-                    auto e1 = doInlineAs!Expression(ifs.ifbody, ids);
-                    assert(ids.foundReturn);
-                    auto e2 = doInlineAs!Expression(s3, ids);
-
-                    Expression e = new CondExp(econd.loc, econd, e1, e2);
-                    e.type = e1.type;
-                    if (e.type.ty == Ttuple)
+                    scope DoInlineAs!Result isv = new DoInlineAs!Result(ids);
+                    bool subseqIsVisited = false;
```
WalterBright: What does `subseq` refer ti>

src/dmd/inline.d
```diff
@@ -226,25 +216,54 @@ public:
...
             alias e2 = elsebody;
+
+            if (thenReturn != elseReturn && visitSubseq)
+            {
+                scope DoInlineAs!Result v = new DoInlineAs!Result(ids);
+                visitSubseq(v);
+                if (!thenReturn)
+                {
+                    e1 = Expression.combine(e1, v.result);
+                    thenReturn = v.ids.foundReturn;
```
WalterBright: These rewrites are pretty important, and need a test case to cover them.

src/dmd/inlinecost.d
```diff
@@ -216,42 +219,49 @@ public:
...
+             *
+             * and:
+             *  if (condition) { ...; return exp1; } else { ...; }
+             *  ...; return exp2;   // subsequent of 'else'.
+             */
+            scope InlineCostVisitor v = new InlineCostVisitor();
+            visitSubseq(v);
+            cost += v.cost;
+            if (!thenReturn)
+                thenReturn = v.foundReturn;
```
WalterBright: need coverage test

src/dmd/inline.d
WalterBright: This would appear to be better served as a function nested within `visit(CompoundStatement)`.

jacob-carlborg: And perhaps give it a more clear.

src/dmd/inlinecost.d
ibuclaw: Adding a third test where there is no return in the if branch.
```
pragma(inline, true)
{
    int foo3(int v)
    {
        return bar3(2 * v);
    }

    int bar3(int a)
    {
        if (a > 0)
            a = 1;
        else
            return baz3(a);
        return a;
    }
}

int baz3(int a)
{
    if (a > 0)
        throw new Exception("a > 0");
    return a - 1;
}
```
Check latest dmd release, and it can't inline it.
```
inline7625.d(62): Error: function inline7625.bar3 cannot inline function
```

```

File: src/dmd/inline.d (hunk 7 above)
Comment (ibuclaw) [reply src/dmd/inline.d]: Test I pasted covers this too.

File: src/dmd/inline.d (hunk 6 above)
Comment (ibuclaw) [reply src/dmd/inline.d]: According to Kenji:
> Subsequent statements of IfStatement.

Judging from the control flow, whilst inlining all CompoundStatements from `n .. dim` (starting at 0), the context is switched over to visitIfStatement if we come across an IfStatement.  It is passed a delegate that switches it back to visitCompoundStatement to inline the remaining statements that follow.

It looks like it will remain false if there were returns found in both branches of the IfStatement.
```
if (foo)
  return 1;
else
  return 2;
// subsequent statements are unreachable.
```

Comment (ibuclaw) [reply src/dmd/inline.d]: If only one of the branches returns, then subsequent statements are moved into the other (the one that doesn't return).

So in the example I gave:
```
pragma(inline, true)
{
    int foo3(int v)
    {
        return bar3(2 * v);
    }

    int bar3(int a)
    {
        if (a > 0)
            a = 1;
        else
            return baz3(a);
        return a;
    }
}
```
Function `foo3` would inline `bar3` as:
```
    int foo3(int v)
    {
        int a = 2 * v;
        if (a > 0)
        {
            a = 1;
            return a;
        }
        else
            return baz3(a);
    }
```
Actual dump from `-vcg-ast` is a bit more succinct however.
```
    int foo3(int v)
    {
        return (int a = 2 * v;) , a > 0 ? a = 1 , a : baz3(a);
    }
```

**Comment by ibuclaw** [comment]
> codecov/patch — 100% of diff hit (target 78.598%)

File: src/dmd/inline.d (hunk 5 above)
Comment (ibuclaw) [reply src/dmd/inline.d]: @WalterBright - I don't think so.  `visitCompoundStatement` can be called from a delegate passing a different `this` context, see `visitIfStatement`.

Comment (ibuclaw) [reply src/dmd/inline.d]: @jacob-carlborg - A more clear what?  Name?  It's just a counter var... startFrom?

File: src/dmd/inline.d (hunk 6 above)
Comment (WalterBright) [reply src/dmd/inline.d]: When I ask for clarification on something, I'm implicitly suggesting it needs more explanation in the comments, not just on github.

File: src/dmd/inline.d (hunk 5 above)
Comment (WalterBright) [reply src/dmd/inline.d]: Ok, but that means even more that `n` needs to be documented. Might also consider making it an actual slice!

File: src/dmd/inline.d (hunk 6 above)
Comment (ibuclaw) [reply src/dmd/inline.d]: They are notes for me to look at later.  I'm trying to understand the original just as much as you are.

**Comment by wilzbach** [comment]
> CyberShadow/DAutoTest — Build failed

that's due to https://github.com/dlang/dub/issues/1336

File: src/dmd/inline.d (hunk 5 above)
Comment (jacob-carlborg) [reply src/dmd/inline.d]: > A more clear what? Name?

@ibuclaw yes, a more clear name.

> It's just a counter var... startFrom?

I don't know what it's counting. I don't like having to read the whole function to figure out what the parameters are used for. If the parameters have descriptive names it a lot easier to figure out what they're used for just by looking at the signature.

**Comment by wilzbach** [comment]
(rebased to remove the failing TravisCI build from the CI list and thus make this PR green)

File: src/dmd/inline.d (hunk 6 above)
Comment (ibuclaw) [reply src/dmd/inline.d]: > They are notes for me to look at later.

As I'm about to pick this up again and I for sure would have forgotten had I not noted it down. 😉

**Comment by wilzbach** [comment]
Jenkins is failing with:

```
allunittests: ocean.math.IEEE: testing ... FAIL [28.5ms]
ocean.core.Test.TestException@./src/ocean/math/IEEE.d(1465): unit test has failed
----------------
./src/ocean/core/Enforce.d:164 void ocean.core.Enforce.enforceImpl!(bool).enforceImpl(lazy Exception, bool, lazy immutable(char)[], immutable(char)[], int) [0xad457f]
./src/ocean/core/Enforce.d:73 void ocean.core.Enforce.enforceImpl!(ocean.core.Test.TestException, bool).enforceImpl(bool, lazy immutable(char)[], immutable(char)[], int) [0xad449e]
./src/ocean/core/Test.d:73 void ocean.core.Test.test!(bool).test(bool, const(char)[], immutable(char)[], int) [0xacdc5a]
./src/ocean/math/IEEE.d:1465 void ocean.math.IEEE.__unittest_ocean_math_IEEE_1429_19() [0xb60500]
??:? void ocean.math.IEEE.__modtest() [0xb620b3]
./src/ocean/core/UnitTestRunner.d:642 ocean.core.UnitTestRunner.UnitTestRunner.Result ocean.core.UnitTestRunner.UnitTestRunner.timedTest(object.ModuleInfo*, out core.sys.posix.sys.time.timeval, ref char[]) [0xabef81]
./src/ocean/core/UnitTestRunner.d:261 int ocean.core.UnitTestRunner.UnitTestRunner.run().__foreachbody2(object.ModuleInfo*) [0xabd01e]
??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)).__lambda2(immutable(object.ModuleInfo*)) [0xfcabd6]
??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))).__foreachbody2(ref rt.sections_elf_shared.DSO) [0xfe7601]
??:? int rt.sections_elf_shared.DSO.opApply(scope int delegate(ref rt.sections_elf_shared.DSO)) [0xfe7a64]
??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))) [0xfe758d]
??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)) [0xfcabad]
./src/ocean/core/UnitTestRunner.d:205 int ocean.core.UnitTestRunner.UnitTestRunner.run() [0xabca57]
./src/ocean/core/UnitTestRunner.d:961 _Dmain [0xabfdd2]
```

Let's see whether there's a persistent failure.

ibuclaw:
Rebased, again.  Added more documentation, renamed `n` parameter to `pos`.

thewilsonator:
Hmm, ocean still fails with `ocean.core.Test.TestException@./src/ocean/math/IEEE.d(1465):`

offending line `   test(feqrel(real.nan,real.nan)==0);`

Geod24:
Can we revive this now that `pragma(inline, true)` has been fixed ?

ibuclaw:
According to digger, Issue 7625 got fixed by #11236.

Rebasing with just the test cases.

Ping @Geod24

thewilsonator (approved):

PR #5428, opened 2016-02-09
Use native AAs for Scope.anchorCounts

WalterBright:
Time to get rid of all uses of root/aav.d

PR #7693, opened 2018-01-13
Add documentation and implementation reference for internal AA hash.

ibuclaw:
Information gathered from comments in #2436 after algorithm was introduced.

### DIFF
--- a/src/dmd/root/aav.d
+++ b/src/dmd/root/aav.d
@@ -15,15 +15,31 @@ module dmd.root.aav;
 import core.stdc.string;
 import dmd.root.rmem;
 
+/****************************************************
+ * Apply supplemental hash function to key in the same manner Java hashes references.
+ * http://hg.openjdk.java.net/jdk8/jdk8/jdk/file/37a05a11f281/src/share/classes/java/util/HashMap.java#l264
+ *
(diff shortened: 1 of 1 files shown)


JinShil (approved):

src/dmd/root/aav.d
```diff
@@ -15,29 +15,46 @@ module dmd.root.aav;
...
 struct AA
 {
     aaA** b;
     size_t b_length;
-    size_t nodes; // total number of aaA nodes
-    aaA*[4] binit; // initial value of b[]
-    aaA aafirst; // a lot of these AA's have only one entry
+    size_t nodes;   // total number of aaA nodes
+    aaA*[4] binit;  // initial value of b[]
+    aaA aafirst;    // a lot of these AA's have only one entry
```
jacob-carlborg: I prefer to not align. The problem is if/when a new line is added it's necessary to realign all lines. Which is more work and creates more noisy diffs.

ibuclaw: Doesn't make for easy reading though having it unaligned.

ibuclaw:
Although on the flip side, I just came across #5428.  So maybe should be doing the opposite of documenting this.

ibuclaw:
Yeah, I changed my mind. #7694

PR #7694, opened 2018-01-13
Use native AAs for Mangler.types and idents

ibuclaw:
Continuation of #5428.

### DIFF
--- a/src/dmd/dmangle.d
+++ b/src/dmd/dmangle.d
@@ -31,7 +31,6 @@ import dmd.identifier;
 import dmd.mtype;
 import dmd.root.ctfloat;
 import dmd.root.outbuffer;
-import dmd.root.aav;
 import dmd.target;
 import dmd.tokens;
 import dmd.utf;
@@ -170,9 +169,8 @@ private extern (C++) final class Mangler : Visitor
 {
     alias visit = Visitor.visit;
 public:
-    static assert(Key.sizeof == size_t.sizeof);
-    AA* types;
-    AA* idents;
+    size_t[void*] types;
+    size_t[void*] idents;
     OutBuffer* buf;
 
     extern (D) this(OutBuffer* buf)
@@ -225,13 +223,13 @@ public:
     {
         if (!t.isTypeBasic())
         {
-            auto p = cast(size_t*)dmd_aaGet(&types, cast(Key)t);
-            if (*p)
+            auto p = cast(void*)t in types;
+            if (p && *p)
             {
                 writeBackRef(buf.offset - *p);
                 return true;
             }
-            *p = buf.offset;
+            types[cast(void*)t] = buf.offset;
         }
         return false;
     }
@@ -251,13 +249,13 @@ public:
     */
     final bool backrefIdentifier(Identifier id)
     {
-        auto p = cast(size_t*)dmd_aaGet(&idents, cast(Key)id);
-        if (*p)
+        auto p = cast(void*)id in idents;
+        if (p && *p)
         {
             writeBackRef(buf.offset - *p);
             return true;
         }
-        *p = buf.offset;
+        idents[cast(void*)id] = buf.offset;
         return false;
     }
 


src/dmd/dmangle.d
```diff
@@ -225,13 +223,13 @@ public:
     {
         if (!t.isTypeBasic())
         {
-            auto p = cast(size_t*)dmd_aaGet(&types, cast(Key)t);
-            if (*p)
+            auto p = cast(void*)t in types;
+            if (p)
```
ibuclaw: Blasts (to put it nicely).  I've just realised why its failing. (check should be `p && *p` :-)

JinShil (approved):

src/dmd/dmangle.d
```diff
@@ -170,9 +169,8 @@ private extern (C++) final class Mangler : Visitor
 {
     alias visit = Visitor.visit;
 public:
-    static assert(Key.sizeof == size_t.sizeof);
-    AA* types;
-    AA* idents;
+    size_t[void*] types;
+    size_t[void*] idents;
```
jacob-carlborg: Why is `void*` used instead of the actual types inserted in the AA?

jacob-carlborg: Found the answer in another PR.

jacob-carlborg (approved):

src/dmd/dmangle.d
ibuclaw: Also Walter explained in the linked PR.

ibuclaw: That it is an optimization is not too obvious unless you know. ☺️

rainers: > That it is an optimization is not too obvious unless you know.

That's not just an optimization, but a necessity as using C++ classes as key just crashes when trying to access the type info.

src/dmd/dmangle.d
```diff
@@ -225,13 +223,13 @@ public:
...
-            auto p = cast(size_t*)dmd_aaGet(&types, cast(Key)t);
-            if (*p)
+            auto p = cast(void*)t in types;
+            if (p && *p)
             {
                 writeBackRef(buf.offset - *p);
                 return true;
             }
-            *p = buf.offset;
+            types[cast(void*)t] = buf.offset;
```
rainers: Not too happy about the duplicate lookups here...

src/dmd/dmangle.d
jacob-carlborg: Perhaps that should be an error instead?

rainers: >Perhaps that should be an error instead?

Or just fall back to void* typeinfo. See also https://issues.dlang.org/show_bug.cgi?id=13875 and https://issues.dlang.org/show_bug.cgi?id=15813.

src/dmd/dmangle.d
WalterBright: Me neither. I suspect it can cause serious slowdowns.

src/dmd/dmangle.d
```diff
@@ -251,13 +249,13 @@ public:
...
-        auto p = cast(size_t*)dmd_aaGet(&idents, cast(Key)id);
-        if (*p)
+        auto p = cast(void*)id in idents;
+        if (p && *p)
         {
             writeBackRef(buf.offset - *p);
             return true;
         }
-        *p = buf.offset;
+        idents[cast(void*)id] = buf.offset;
```
WalterBright: More duplicate lookups.

WalterBright:
May need to revert this due to slowdowns caused by duplicate lookups.

ibuclaw:
@WalterBright - maybe you can think of an alternative?  You introduced duplicate lookups in #5428 too.

WalterBright:
This was reverted and added to the Phantom Zone because it is a good idea, but awaits a resolution of the double lookup problem which affects most of the uses of built-in arrays in the dmd front end.

ibuclaw:
@WalterBright could root.aav perhaps be converted into a thin wrapper around druntime AAs in the meantime perhaps? 

The one difference would be to use aaGet instead of aaIn for the overload of operator `in`.

PR #7695, opened 2018-01-13
VS: choose the default compiler automatically

rainers:
so you don't have to switch it manually from within Visual Studio

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -44,8 +44,6 @@ build_script:
 
   # build via VS projects (needs dmd downloaded in appveyor.sh)
   - cd c:\projects\dmd\src
-  # upgrade used compiler version to VS2015
-  - sed -i s/v120/v140/ vcbuild\dmd_backend.vcxproj
   - devenv /Build "Release|Win32" /Project dmd vcbuild\dmd.sln
 
   # sanity check: build druntime unittests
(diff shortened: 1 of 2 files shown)


wilzbach (approved):

wilzbach:
Now if there was an easy way to avoid manually specifying all D files in this solution, you would make my day :)
Is there no support for globbing or folders?

rainers:
> Is there no support for globbing or folders?

MSBuild can do that for VC projects, but not Visual D projects. So it might work if we switch to the D integration with MSBuild.
I never tried that because there have been too many d files in the source directory that should not be part of the build (e.g. idgen or impconvtab). These seem to have been replaced, though.

wilzbach:
> I never tried that because there have been too many d files in the source directory that should not be part of the build (e.g. idgen or impconvtab). These seem to have been replaced, though.

Yep, these files have been replaced with CTFE as part of the work for the DMD as a DUB package.
We also applied a couple of `version(XX) {} else` on top of files, s.t. all we need for DUB is:

https://github.com/dlang/dmd/blob/b1f6aec796286a08395d689b54dacdd01d141f9e/dub.sdl#L62-L64

(the three packages above aren't in any anymore)

It would be really cool to be able to do the same for for Visual Studio ;-)

jacob-carlborg:
Since the Visual Studio project seems to be just an XML file, can't generate it from some other source? We can do something very simple to begin with. Have a template file that contains everything needed for the Visual Studio project except for the source code files. Then have a tool that inserts the source code files into the XML file.

wilzbach:
IIRC that wasn't possible because people would need to run this before opening the project in Visual Studio?

jacob-carlborg:
You could still commit the generated file and then as a test run the tool again and compare the committed file against the new generated file. That way the test would fail if the files are out of sync.

jacob-carlborg:
Seems like Rainer already has a different solution https://github.com/dlang/dmd/pull/7706 by compiling all D files.

PR #4569, opened 2015-04-07
[refactor] Move Type::ctype into an external AA

yebblies:
The idea is that backend-specific passes shouldn't need to add methods to the frontend ast classes.  Using a hash scales much better, if the performance is acceptable.

PR #7696, opened 2018-01-13
Use native AAs for DsymbolTable.tab

ibuclaw:
Continuation of #5428.

### DIFF
--- a/src/dmd/dsymbol.d
+++ b/src/dmd/dsymbol.d
@@ -43,7 +43,6 @@ import dmd.lexer;
 import dmd.mtype;
 import dmd.nspace;
 import dmd.opover;
-import dmd.root.aav;
 import dmd.root.rmem;
 import dmd.root.rootobject;
 import dmd.root.speller;
@@ -2078,33 +2077,30 @@ extern (C++) final class ForwardingScopeDsymbol : ScopeDsymbol
  */
 extern (C++) final class DsymbolTable : RootObject
 {
-    AA* tab;
+    Dsymbol[void*] tab;
 
     // Look up Identifier. Return Dsymbol if found, NULL if not.
     Dsymbol lookup(const Identifier ident)
     {
         //printf("DsymbolTable::lookup(%s)\n", (char*)ident.string);
-        return cast(Dsymbol)dmd_aaGetRvalue(tab, cast(void*)ident);
+        auto ps = cast(void*)ident in tab;
+        return ps ? *ps : null;
     }
 
     // Insert Dsymbol in table. Return NULL if already there.
     Dsymbol insert(Dsymbol s)
     {
         //printf("DsymbolTable::insert(this = %p, '%s')\n", this, s.ident.toChars());
-        const ident = s.ident;
-        Dsymbol* ps = cast(Dsymbol*)dmd_aaGet(&tab, cast(void*)ident);
-        if (*ps)
+        if (cast(void*)s.ident in tab)
             return null; // already in table
-        *ps = s;
+        tab[cast(void*)s.ident] = s;
         return s;
     }
 
     // Look for Dsymbol in table. If there, return it. If not, insert s and return that.
     Dsymbol update(Dsymbol s)
     {
-        const ident = s.ident;
-        Dsymbol* ps = cast(Dsymbol*)dmd_aaGet(&tab, cast(void*)ident);
-        *ps = s;
+        tab[cast(void*)s.ident] = s;
         return s;
     }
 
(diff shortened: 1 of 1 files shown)


wilzbach (approved):

rainers:
Have you measured the impact on performance? I opted for the dmd specific ones in dmd.dmangle because they were slightly faster in my benchmarks.

ibuclaw:
Well I'd imagine that you'd find the internal compiler AA to be a little bit faster.  Being able to inline the implementation helps greatly in that matter.

In terms of what this is doing vs. the internal AA, there won't be much to it.  `void*` is used as the key type so that we're not calling the `getHash()`/`toHash()` implementation for classes.

rainers:
Yeah, knowing to deal with void* only helps to avoid all the indirections through the type info.

When I tried it, I was hoping for the quadratic probing of the native AAs to help, but even the compiler internal AAs (that still use linked lists per bucket) didn't benefit from a rewrite in that direction.

ibuclaw:
Things might get more interesting anyway.  I would like to see some progression in this area also (https://github.com/dlang/dmd/pull/4569/commits/e6c838925665039a2cb98a9ff5cdc6a6313571a9)

JinShil (approved):

src/dmd/dsymbol.d
```diff
@@ -2078,44 +2077,40 @@ extern (C++) final class ForwardingScopeDsymbol : ScopeDsymbol
...
     Dsymbol insert(Dsymbol s)
     {
         //printf("DsymbolTable::insert(this = %p, '%s')\n", this, s.ident.toChars());
-        const ident = s.ident;
-        Dsymbol* ps = cast(Dsymbol*)dmd_aaGet(&tab, cast(void*)ident);
-        if (*ps)
+        if (cast(void*)s.ident in tab)
             return null; // already in table
-        *ps = s;
+        tab[cast(void*)s.ident] = s;
```
WalterBright: The trouble here is this adds a second lookup. This is not acceptable (performance reasons).

src/dmd/dsymbol.d
```diff
@@ -2078,44 +2077,40 @@ extern (C++) final class ForwardingScopeDsymbol : ScopeDsymbol
...
     // when ident and s are not the same
     Dsymbol insert(const Identifier ident, Dsymbol s)
     {
         //printf("DsymbolTable::insert()\n");
-        Dsymbol* ps = cast(Dsymbol*)dmd_aaGet(&tab, cast(void*)ident);
-        if (*ps)
+        if (cast(void*)ident in tab)
             return null; // already in table
-        *ps = s;
+        tab[cast(void*)ident] = s;
```
WalterBright: Another double lookup.

WalterBright (changes requested):
Cannot accept unless the redundant lookups are fixed.

src/dmd/dsymbol.d
ibuclaw: You were happy to add one to #5428.

WalterBright: Yes, and that wasn't in a hot section of code. The dmd symbol table is very hot.

ibuclaw: Well in the absence of a getOrCreate property for built-in AAs, not much can really be done to help that there.

rainers: Well, there is _aaGetY which could be called directly, but I'm not sure we should go that road. It is already used by `dup`, maybe a similar small helper in object.d would be a nice extension. IIRC there are also a couple of enhancement requests in bugzilla for this.

ibuclaw: Well, `_aaGetY` does not have enough information to say whether or not a value existed before a lookup was requested.  One would have to either define a new internal function, move the common bits into new helpers, and add the difference

i.e:
```
void* _aaGetY(...)
{
  auto p = aaLookup(...);
  if (!p)
    p = aaInsert(...);
  return p;
}

void* _aaGetOrInsert(..., ref bool existed)
{
  existed = false;
  auto p = aaLookup(...);
  if (!p)
    p = aaInsert(...);
  else
    existed = true;
  return p;
}
```

rainers: In general the information `existed` is desirable, but for dmd, _aaGetY has the same semantics as dmd_aaGet. The zero initialization is used to detect whether the key has been added.

WalterBright:
I'm going to Phantom Zone this until a solution is found for the double lookups, at which point it can be revived. The penalty of double lookup outweighs the advantage of dogfooding.

PR #7698, opened 2018-01-13
Replace real_t = 0 with real_t = CTFloat.zero

ibuclaw:
Remove the last few places where a scalar is used to initialize real_t (which is a struct type in other compilers).


wilzbach (approved):
👍

JinShil (approved):

PR #6743, opened 2017-05-06
Make DMD work with a struct as the main integral type

yebblies:
@redstar

PR #7699, opened 2018-01-13
Use sinteger_t and uinteger_t for Token.int64value and uns64value

ibuclaw:
Remove the encoded size from the field name.

One part of #6743, and there are plenty more to come as I cherry-pick parts that have nothing to do with adding a synthetic integer type itself.


JinShil:
Quite a simple change, but I'm afraid I don't understand what's motivating this.  Care to elaborate?

ibuclaw:
Eventually, support for cent and ucent in the compiler.

JinShil (approved):

PR #7701, opened 2018-01-13
fix Issue 18057 - [ICE] Segmentation fault (stack overflow) in Expression::ctfeInterpret()

ibuclaw:
Inside `AggregateDeclaration::fill` looks to be the better place to have this error rather than `VarDeclaration::getConstInitializer`.  In the latter location you'd get the same error printed twice.

### DIFF
--- a/test/compilable/interpret3.d
+++ b/test/compilable/interpret3.d
@@ -7731,3 +7731,14 @@ bool foo17407()
 
 static assert(!foo17407);
 
+/**************************************************/
+// https://issues.dlang.org/show_bug.cgi?id=18057
+// Recursive field initializer causes segfault.
+
+struct RBNode(T)
+{
+    RBNode!T *copy = new RBNode!T;
+}
+
+static assert(!__traits(compiles, { alias bug18057 = RBNode!int; }));
+
--- a/test/fail_compilation/fail18057.d
+++ b/test/fail_compilation/fail18057.d
@@ -1,0 +1,16 @@
+/**
+TEST_OUTPUT:
+---
+fail_compilation/fail18057.d(16): Error: template instance RBNode!int `RBNode` is not a template declaration, it is a struct
+fail_compilation/fail18057.d(13): Error: variable fail18057.RBNode.copy recursive initialization of field
+---
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=18057
+// Recursive field initializer causes segfault.
+struct RBNode
+{
+    RBNode *copy = new RBNode;
+}
+
+alias bug18057 = RBNode!int;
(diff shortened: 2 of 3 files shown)


JinShil (approved):

test/compilable/interpret3.d
```diff
@@ -7731,3 +7731,14 @@ bool foo17407()
...
+/**************************************************/
+// https://issues.dlang.org/show_bug.cgi?id=18057
+// Recursive field initializer causes segfault.
+
+struct RBNode(T)
+{
+    RBNode!T *copy = new RBNode!T;
+}
+
+static assert(!__traits(compiles, { alias bug18057 = RBNode!int; }));
```
wilzbach: Shouldn't we test for the exact error message here?

ibuclaw: If you insist.  Adding more files to the testsuite has been complained about in the past, as it is gradually slowing down the time it takes to run the testsuite vs. if all tests were kept in one mega file.  (Not saying everything should go in one mega file, but logically grouped may be a start :-)

ibuclaw: I created an issue to note this, incase anyone wants to take the baton: https://issues.dlang.org/show_bug.cgi?id=18235

WalterBright: The exact error message only needs to be tested once by the test suite, the others can use `__traits(compiles, ... )`. So, the coverage results can be used to determine if the exact message is ever tested or not.

test/compilable/interpret3.d
```diff
@@ -1,4 +1,4 @@
-﻿// PERMUTE_ARGS: -inline
+// PERMUTE_ARGS: -inline
```
ibuclaw: What's going on with this?

ibuclaw:
Gah, I was just about to do this. :-)

test/fail_compilation/fail18057.d
```diff
@@ -0,0 +1,16 @@
...
+TEST_OUTPUT:
+---
+fail_compilation/fail18057.d(13): Error: variable fail18057.RBNode!int.RBNode.copy recursive initialization of field
+fail_compilation/fail18057.d(16): Error: template instance fail18057.RBNode!int error instantiating
+---
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=18057
+// Recursive field initializer causes segfault.
+struct RBNode(T)
```
ibuclaw: Doesn't need to be a template.  Only did that for the benefit of `__traits(compiles)`.

test/compilable/interpret3.d
wilzbach: I just moved it to `fail_compilation`. It's just that adding `static assert(!__traits(compiles,` to `compilable` tests strikes me as an anti-pattern

test/compilable/interpret3.d
wilzbach: Trailing whitespace - my editor automatically removes it. Sorry

test/fail_compilation/fail18057.d
wilzbach: OK. Removed.

PR #4882, opened 2015-08-12
Issue 14907 - DMD crash when using template name as a default value of template's typed argument

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=14907

Add `TemplateDeclaration::inuse`, and use it to detect recursive template expansion during the template parameter matching.

PR #7702, opened 2018-01-13
fix Issue 14907 - DMD crash when using template name as a default value of template's typed argument

ibuclaw:
Add TemplateDeclaration::inuse, and use it to detect recursive template expansion during the template parameter matching.

Revival of #4882.

### DIFF
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -518,6 +518,7 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
     bool ismixin;           // template declaration is only to be used as a mixin
     bool isstatic;          // this is static template declaration
     Prot protection;
+    int inuse;              /// for recursive expansion detection
 
     // threaded list of previous instantiation attempts on stack
     TemplatePrevious* previous;
@@ -855,7 +856,9 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
                     printf("\tparameter[%d] is %s : %s\n", i, tp.ident.toChars(), ttp.specType ? ttp.specType.toChars() : "");
             }
 
+            inuse++;
             m2 = tp.matchArg(ti.loc, paramscope, ti.tiargs, i, parameters, dedtypes, &sparam);
+            inuse--;
             //printf("\tm2 = %d\n", m2);
             if (m2 == MATCH.nomatch)
             {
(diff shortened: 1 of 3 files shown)


JinShil (approved):

src/dmd/dtemplate.d
```diff
@@ -519,6 +519,7 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
     bool ismixin;           // template declaration is only to be used as a mixin
     bool isstatic;          // this is static template declaration
     Prot protection;
+    int inuse;              // for recursive expansion detection
```
jacob-carlborg: Please add a Ddoc comment and I suggest naming it `inUse` to follow standard D naming conventions.

jacob-carlborg (changes requested):

src/dmd/dtemplate.d
ibuclaw: I veto renaming, as it doesn't follow the internal compiler convention, and conflicts with naming convention for functions - there should be a distinction between what is and is not a side effect, especially as not everyone is using parentheses anymore.

JinShil: I think it would improve the DMD code base to establish a style guide of some kind.  [The official style guide](https://dlang.org/dstyle.html) is a good place to start.  If it is insufficient in some way, let's amend it for DMD.  DMD's code has clearly gone through a few stages, and is showing its age.  It would be nice to set some code-quality expectations for new code, and encourage refactorings that would help remove some of the oddities.

Perhaps now is not the right time with the backend still written in C+.  but I hope we can come to some agreement about conventions and document them somewhere. 

 The point about side effects is interesting, but isn't `pure` supposed to help with that?

jacob-carlborg: I don't see how the backend being written in C++ affects the style guide of the D code.

jacob-carlborg: I wouldn't say that renaming is required but I do require the Ddoc comment. Or make it `private`, it's not used anywhere outside of the class as far as I can see.

ibuclaw: > The point about side effects is interesting, but isn't pure supposed to help with that?

Most methods in the compiler are pure, because they only change internal state.  (Pure in the D sense is not to be confused with pure in the C sense, or absent of side effects).

ibuclaw:
Added `///` to comment.

jacob-carlborg (approved):

wilzbach:
(rebased to remove the failing Travis)

JinShil (approved):

ibuclaw:
This PR introduced a regression https://issues.dlang.org/show_bug.cgi?id=18646

PR #7703, opened 2018-01-14
Add ddoc fullText to the CLI switches

wilzbach:
Required for automatically generating the CLI options at dlang.org, see https://github.com/dlang/dlang.org/pull/2068
There were the following differences:

- `-gc` (only on dlang.org - the flag has been removed)
- `profile` (not mentioned on dlang.org)
- `-run` (not mentioned at the DMD output)
- `--version` (not mentioned on dlang.org)

The DMD CLI wasn't alphabetical ordered for a few cases (https://dlang.org/dmd.html is).
(in the transition from the manual CLI output to the dynamically generated one, I kept the ordering to allow easy diffing of the output).

Generating the language transitions automatically depends on https://github.com/dlang/dmd/pull/7648
Anyways, the available language transitions aren't shown on dlang.org at the moment.

### DIFF
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -109,6 +109,7 @@ struct Usage
             this.os = os;
         }
 
+        /// ditto
         this(string flag, string helpText, string ddocText, TargetOS os = TargetOS.all)
         {
             this.flag = flag;
@@ -154,7 +155,7 @@ struct Usage
         ),
         Option("boundscheck=[on|safeonly|off]",
             "bounds checks on, in @safe only, or off",
-            `controls if bounds checking is enabled.
+            `Controls if bounds checking is enabled.
                 $(UL
                     $(LI $(I on): Bounds checks are enabled for all code. This is the default.)
                     $(LI $(I safeonly): Bounds checks are enabled only in $(D @safe) code.
@@ -222,15 +223,15 @@ dmd -cov -unittest myprog.d
         ),
         Option("debug",
             "compile in debug code",
-            `compile in $(LINK2 spec/version.html#debug, debug) code`,
+            `Compile in $(LINK2 spec/version.html#debug, debug) code`,
         ),
         Option("debug=<level>",
             "compile in debug code <= level",
-            `compile in $(LINK2 spec/version.html#debug, debug level) &lt;= $(I level)`,
+            `Compile in $(LINK2 spec/version.html#debug, debug level) &lt;= $(I level)`,
         ),
         Option("debug=<ident>",
             "compile in debug code identified by ident",
-            `compile in $(LINK2 spec/version.html#debug, debug identifier) $(I ident)`,
+            `Compile in $(LINK2 spec/version.html#debug, debug identifier) $(I ident)`,
         ),
         Option("debuglib=<name>",
             "set symbolic debug library to name",
@@ -270,13 +271,13 @@ dmd -cov -unittest myprog.d
         Option("g",
             "add symbolic debug info",
             `$(WINDOWS
-                add CodeView symbolic debug info with
+                Add CodeView symbolic debug info with
                 $(LINK2 $(ROOT_DIR)spec/abi.html#codeview, D extensions)
                 for debuggers such as
                 $(LINK2 http://ddbg.mainia.de/releases.html, Ddbg)
             )
             $(UNIX
-                add symbolic debug info in Dwarf format
+                Add symbolic debug info in Dwarf format
                 for debuggers such as
                 $(D gdb)
             )`,
@@ -292,11 +293,11 @@ dmd -cov -unittest myprog.d
         ),
         Option("H",
             "generate 'header' file",
-            `generate $(RELATIVE_LINK2 $(ROOT_DIR)interface-files, D interface file)`,
+            `Generate $(RELATIVE_LINK2 $(ROOT_DIR)interface-files, D interface file)`,
         ),
         Option("Hd=<directory>",
             "write 'header' file to directory",
-            `write D interface file to $(I dir) directory. $(SWLINK -op)
+            `Write D interface file to $(I dir) directory. $(SWLINK -op)
             can be used if the original package hierarchy should
             be retained.`,
         ),
@@ -320,13 +321,13 @@ dmd -cov -unittest myprog.d
         ),
         Option("inline",
             "do function inlining",
-            `inline expand functions at the discretion of the compiler.
+            `Inline expand functions at the discretion of the compiler.
             This can improve performance, at the expense of making
             it more difficult to use a debugger on it.`,
         ),
         Option("J=<directory>",
             "look for string imports also in directory",
-            `where to look for files for
+            `Where to look for files for
             $(LINK2 $(ROOT_DIR)spec/expression.html#ImportExpression, $(I ImportExpression))s.
             This switch is required in order to use $(I ImportExpression)s.
             $(I path) is a ; separated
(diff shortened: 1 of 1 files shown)


src/dmd/cli.d
```diff
@@ -88,19 +106,60 @@ struct Usage
...
+                are not inserted into `eh_frame`.)
+                )
+                $(LI Module constructors and destructors are not generated meaning that
+                $(LINK2 $(ROOT_DIR)spec/class.html#StaticConstructor, static) and
+                $(LINK2 $(ROOT_DIR)spec/class.html#SharedStaticConstructor, shared static constructors) and
+                $(LINK2 $(ROOT_DIR)spec/class.html#StaticDestructor, destructors)
+                will not get called.)
+                $(LI `ModuleInfo` is not generated.)
+                $(LI $(LINK2 $(ROOT_DIR)phobos/object.html#.TypeInfo, `TypeInfo`)
+                instances will not be generated for structs.)
```
wilzbach: I'm not so sure that the CLI ages are the best place to document this. Imho we should just link to https://dlang.org/spec/betterc.html and improve this page if needed.

src/dmd/cli.d
```diff
@@ -88,19 +106,60 @@ struct Usage
...
+                                        This is the default for $(SWLINK -release) builds.)
+                    $(LI $(I off): Bounds checks are disabled completely (even in $(D @safe)
+                                   code). This option should be used with caution and as a
+                                   last resort to improve performance. Confirm turning off
+                                   $(D @safe) bounds checks is worthwhile by benchmarking.)
+                )`
         ),
         Option("c",
-            "do not link"
+            "compile only, do not link"
```
wilzbach: That's the text from dlang.org - imho a bit more descriptive.

src/dmd/cli.d
```diff
@@ -109,19 +168,33 @@ struct Usage
...
+---
+dmd -cov -unittest myprog.d
+---
+            )`,
         ),
         Option("D",
-            "generate documentation"
+            "generate documentation",
+            `$(P Generate $(LINK2 $(ROOT_DIR)spec/ddoc.html, documentation) from source.)
+            $(P Note: mind the $(LINK2 $(ROOT_DIR)spec/ddoc.html#security, security considerations).)
```
wilzbach: I don't think that the paragraphs are needed here, but it's a 1:1 from dlang.org

src/dmd/cli.d
```diff
@@ -194,90 +295,206 @@ struct Usage
             "ignore unsupported pragmas"
         ),
         Option("inline",
-            "do function inlining"
+            "do function inlining",
+            `inline expand functions at the discretion of the compiler.
```
wilzbach: The "inline expand" looks out of place. I would just say "Inline functions"

src/dmd/cli.d
```diff
@@ -73,6 +73,24 @@ struct Option
     string flag; /// The CLI flag without leading `-`, e.g. `color`
     string helpText; /// A detailed description of the flag
     TargetOS os = TargetOS.all; /// For which `TargetOS` the flags are applicable
+    string ddocText; /// Detailed description of the flag (in Ddoc)
+
+    ///
+    this(string flag, string helpText, TargetOS os = TargetOS.all)
```
jacob-carlborg: Ideally all these parameters should be documented.

JinShil:
I'm a little skeptical about this documentation being in the binary unless it actually serves some for users of the compiler.  At the moment I think this is only used as a datastore to keep all related information about the compiler flags in one place.  That's perfectly fine, but I think we should version it out (may actually need string mixins) so it is only used for generating the documentation, but doesn't end up in the published binary.

wilzbach:
> I'm a little skeptical about this documentation being in the binary unless it actually serves some for users of the compiler. 

It's needed for dlang/dlang.org#2068

> but doesn't end up in the published binary.

That's why the CLI text is generated at CTFE, so these strings shouldn't end up in the binary.
I made this more explicit with `version(DdocOptions)` as I checked the binary and DMD didn't seem to be smart enough. It would be cool to avoid the need for `version` though, but apparently DMD judges a template instantiation at CTFE reson enough to send a templated function to object code.

src/dmd/cli.d
wilzbach: Done.

WalterBright:
> That's why the CLI text is generated at CTFE, so these strings shouldn't end up in the binary.
I made this more explicit with version(DdocOptions) as I checked the binary and DMD didn't seem to be smart enough. It would be cool to avoid the need for version though, but apparently DMD judges a template instantiation at CTFE reson enough to send a templated function to object code.

Please file a bugzilla for this. But it still needs to be versioned out so it doesn't appear in the dmd binary, because updating the dmd that builds dmd may be a looong wait.

wilzbach:
> Please file a bugzilla for this.

https://issues.dlang.org/show_bug.cgi?id=18238

wilzbach:
So as this is now using `version(DdocOptions)` and won't add up to the executable size. Is everyone OK with moving forward?
As mentioned the main benefit here is that there will be only _one_ source of truth for the compiler arguments and dlang.org/dmd.html will automatically be in sync.

JinShil:
> Is everyone OK with moving forward?

LGTM

I don't see the output in DAutoTest.  Please advise.

wilzbach:
> I don't see the output in DAutoTest. Please advise.

Yes, this PR is a dependency of https://github.com/dlang/dlang.org/pull/2068
I could probably change https://github.com/dlang/dlang.org/pull/2068 to use `static if` to softly fail if no ddoxText attribute is available, but unfortunately one of the two PRs has to be merged without the preview and I thought this PR here is easier to review, because it just adds text.

JinShil (approved):

src/dmd/cli.d
```diff
@@ -68,18 +68,66 @@ bool isCurrentTargetOS(TargetOS os)
...
     {
         string flag; /// The CLI flag without leading `-`, e.g. `color`
         string helpText; /// A detailed description of the flag
         TargetOS os = TargetOS.all; /// For which `TargetOS` the flags are applicable
+
+    version(DdocOptions)
+    {
+        string ddocText; /// Detailed description of the flag (in Ddoc)
+
+        this(string flag, string helpText, TargetOS os = TargetOS.all)
```
marler8997: Maybe pull this method out of the version blocks since it appears to be the same in both cases?

src/dmd/cli.d
```diff
@@ -68,18 +68,66 @@ bool isCurrentTargetOS(TargetOS os)
...
+        {
+            this.flag = flag;
+            this.helpText = helpText;
+            this.os = os;
+        }
+
+        this(string flag, string helpText, string ddocText, TargetOS os = TargetOS.all)
```
marler8997: Maybe
```
version(DdocOptions) this.ddocText = ddocText;
```
Then remove the other instance of this constructor?

marler8997 (approved):

JinShil:
@wilzbach This has been open for a week, which I think is ample time for others to weigh in.  I'll probably merge this if you address @marler8997's comments above and no other objections arise.

wilzbach:
Cool! I will do so once I am back and have access to my laptop.

wilzbach:
(Rebased, addressed the review and resolved the merge conflict.)

src/dmd/cli.d
JinShil: I agree, but let's leave that to another PR.

src/dmd/cli.d
JinShil: I agree.  Sorry, I didn't comment about  that earlier.  Care to make the change to "Inline functions..." before we merge?

PR #7706, opened 2018-01-14
VS: use dmd.vcxproj instead of visuald project to allow wildcards

rainers:
This uses msbuild integration of Visual D into VC projects. This format is probably not as well tested as the Visual D projects, but also allows building through msbuild instead of devenv.

compiles all D files dmd/**/*.d


wilzbach (approved):
Cool!

wilzbach:
AppVeyor passes, so I assume this is good to go?

PR #7707, opened 2018-01-14
fix issue Issue 15482 - new uuid.d forbids to link statically with other libraries

Temtaime:
new uuid.d forbids to link statically with other libraries

Part of https://github.com/dlang/druntime/pull/2040

### DIFF
--- a/samples/win32.mak
+++ b/samples/win32.mak
@@ -51,12 +51,12 @@ winsamp:
 # COM client/server example
 # dclient will fail unless run with administrator rights
 dserver32:
-	$(DMD) dserver.d chello.d $(DFLAGS) dserver.def advapi32.lib ole32.lib user32.lib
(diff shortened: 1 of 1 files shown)


WalterBright:
> new uuid.d forbids to link statically with other libraries

Is this a licensing issue or a technical one?

n8sh:
Per Temtaime, a technical one https://github.com/dlang/druntime/pull/2040#issue-288363344
>All of those symbols are stored inside uuid.lib on windows.

JinShil:
@rainers Could you please review this?

rainers:
As https://github.com/dlang/druntime/pull/2063 was merged instead of https://github.com/dlang/druntime/pull/2040, this change is not needed. 
Maybe we should rather remove this strange dependency on uuid.lib for DLLs: https://github.com/dlang/dmd/blob/master/src/dmd/glue.d#L939.

PR #6090, opened 2016-08-26
refactor: move safe code to safe.d

WalterBright:
expression.d is getting too big. Move `checkUnsafeAccess()` and `isSafeCast()`

PR #5085, opened 2015-09-17
[REG2.066] Issue 14886 - std.parallelism.parallel with large static array seems to hang compile

9rnsr:
https://issues.dlang.org/show_bug.cgi?id=14886

PR #7708, opened 2018-01-14
[refactor] Remove unused local variables.

ibuclaw:
Noticed during some patch rebasing, so decided to compile dmd using `gdc -Wunused` and cleaned up what was flagged.

I think I can only see two places where an assignment to a dead var is done _only_ for the side-effect.  If that is the case, then those checks should really be moved elsewhere/handled better.

Let's see what the CI says...

### DIFF
--- a/src/dmd/ctfeexpr.d
+++ b/src/dmd/ctfeexpr.d
@@ -328,7 +328,6 @@ extern (C++) UnionExp copyLiteral(Expression e)
     if (e.op == TOKarrayliteral)
     {
         auto ale = cast(ArrayLiteralExp)e;
-        auto basis = ale.basis ? copyLiteral(ale.basis).copy() : null;
         auto elements = copyLiteralArray(ale.elements, ale.basis);
 
         emplaceExp!(ArrayLiteralExp)(&ue, e.loc, elements);
--- a/src/dmd/e2ir.d
+++ b/src/dmd/e2ir.d
@@ -1720,8 +1720,6 @@ elem *toElem(Expression e, IRState *irs)
                 }
                 else
                 {
-                    d_uns64 elemsize = sd.size(ne.loc);
-
                     // call _d_newitemT(ti)
                     e = getTypeInfo(ne.newtype, irs);
 
@@ -1816,7 +1814,6 @@ elem *toElem(Expression e, IRState *irs)
             else if (t.ty == Tpointer)
             {
                 TypePointer tp = cast(TypePointer)t;
-                Expression di = tp.next.defaultInit();
                 elem *ezprefix = ne.argprefix ? toElem(ne.argprefix, irs) : null;
 
                 // call _d_newitemT(ti)
@@ -2145,7 +2142,6 @@ elem *toElem(Expression e, IRState *irs)
 
                 ev = el_una(OPind, tym, ev);
 
-                CastExp ce = cast(CastExp)e1;
                 for (size_t d = depth; d > 0; d--)
                 {
                     e1 = be.e1;
@@ -2627,7 +2623,6 @@ elem *toElem(Expression e, IRState *irs)
             {
                 SliceExp are = cast(SliceExp)ae.e1;
                 Type t1 = t1b;
-                Type t2 = ae.e2.type.toBasetype();
                 Type ta = are.e1.type.toBasetype();
 
                 // which we do if the 'next' types match
@@ -5280,7 +5275,6 @@ elem *toElem(Expression e, IRState *irs)
             assert(dim);
 
             Type telem = ((*exps)[0] ? (*exps)[0] : basis).type;
-            Type tsarray = telem.sarrayOf(dim);
             targ_size_t szelem = telem.size();
             .type *te = Type_toCtype(telem);   // stmp[] element type
 
--- a/src/dmd/safe.d
+++ b/src/dmd/safe.d
@@ -110,9 +110,9 @@ bool isSafeCast(Expression e, Type tfrom, Type tto)
     auto tfromb = tfrom.toBasetype();
     auto ttob = tto.toBasetype();
 
-    if (ttob.ty == Tclass && tfrom.ty == Tclass)
+    if (ttob.ty == Tclass && tfromb.ty == Tclass)
     {
-        ClassDeclaration cdfrom = tfrom.isClassHandle();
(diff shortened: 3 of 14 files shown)


src/dmd/safe.d
```diff
@@ -107,7 +107,6 @@ bool isSafeCast(Expression e, Type tfrom, Type tto)
     if (!tto.hasPointers())
         return true;
 
-    auto tfromb = tfrom.toBasetype();
     auto ttob = tto.toBasetype();
 
     if (ttob.ty == Tclass && tfrom.ty == Tclass)
```
ibuclaw: This might actually be a typo.

src/dmd/e2ir.d
```diff
@@ -1816,7 +1814,6 @@ elem *toElem(Expression e, IRState *irs)
             else if (t.ty == Tpointer)
             {
                 TypePointer tp = cast(TypePointer)t;
-                Expression di = tp.next.defaultInit();
```
ibuclaw: Calling `defaultInit` _might_ produce an error.

src/dmd/e2ir.d
```diff
@@ -1720,8 +1720,6 @@ elem *toElem(Expression e, IRState *irs)
                 }
                 else
                 {
-                    d_uns64 elemsize = sd.size(ne.loc);
```
ibuclaw: Calling `size` _might_ produce an error.

wilzbach (approved):
CI likes it, how about adding this to `posix.mak` to prevent regressions?

```
ifeq ($(HOST_DMD_KIND,gdc)
 DFLAGS += -Wunused
endif
```

jacob-carlborg (approved):

ibuclaw:
> CI likes it, how about adding this to posix.mak to prevent regressions?

Actually, it's a WIP feature I'm testing, but maybe in the future once I iron things out. :-)

There's at least one case that I'm not able to catch:
```
const(uint) foo = 6;
int bar = foo + 4;  // front-end optimizer will wipe "foo" out of AST before reaching codegen.
```

src/dmd/e2ir.d
ibuclaw: Error should be caught by the front-end here:
https://github.com/dlang/dmd/blob/74285e41e670bf7e2ae7d83bc9b1617e12f93a65/src/dmd/expressionsem.d#L2273-L2276

ibuclaw:
This is good to go as far as I can tell.  I've done a spot check of the areas I thought were most worrisome.

src/dmd/safe.d
ibuclaw: @wilzbach - worth replacing `tfrom` with `tfromb` here?  PR that introduced was #6090

andralex (approved):

src/dmd/safe.d
wilzbach: https://github.com/dlang/dmd/blob/74285e41e670bf7e2ae7d83bc9b1617e12f93a65/src/dmd/mtype.d#L2293-L2296

https://github.com/dlang/dmd/blob/74285e41e670bf7e2ae7d83bc9b1617e12f93a65/src/dmd/mtype.d#L7966-L7972

1) It definitely was a typo
2) `toBasetype` only has an effect for `TypeEnum` 
3) The following isn't legal anyways:

```d
enum foo = new Object();
```
https://run.dlang.io/is/F0dHH5

4) We should be consistent -> let's rename `tfrom` to `tfromb` (or drop `ttob`)

wilzbach:
(removed "auto-merge" until we have reached a decision regarding the typo in `safe.d`)

ibuclaw:
I fixed up `isSafeCast` to use `tfromb` everywhere applicable.  Should be unneeded to have either, as both should have already been stripped of outer types.  But I guess better to not assume.

src/dmd/ctfeexpr.d
```diff
@@ -328,7 +328,6 @@ extern (C++) UnionExp copyLiteral(Expression e)
     if (e.op == TOKarrayliteral)
     {
         auto ale = cast(ArrayLiteralExp)e;
-        auto basis = ale.basis ? copyLiteral(ale.basis).copy() : null;
         auto elements = copyLiteralArray(ale.elements, ale.basis);
```
WalterBright: I'm suspicious that the `basis` was intended to be used on line 331 instead of `ale.basis`. This is a serious issue.

WalterBright (changes requested):
The `ale.basis` comment. I suggest altering line 331 to use `basis` instead and running the test suite.

src/dmd/ctfeexpr.d
ibuclaw: Introducing PR was #5085. I believe your concerns are unfounded based on the change there (it's always been dead).

WalterBright: It looks to me like the bug was Kenji forgetting to use `basis` in the next line. The code sure looks like it was intended to. The fact that it did not trigger a failure of the test suite doesn't mean it's not a bug. I bet if you make the change to use it in the next like, it will also pass the test suite.

I cannot approve this PR with that line deleted without some investigation into what was the intended use of it. If you don't want to investigate it, and move ahead with this PR, I suggest flagging it with a comment and leave the line as is, and I'll approve it.

WalterBright: @wilzbach issues like this one are why having a warning on unused variables would be problematical.

ibuclaw: I'm just saying let's not discount an early version of the PR needed it, but later changes made it irrelevant.

As for warnings such as -Wunused, it's only useful for new PRs, particularly refactorings.

ibuclaw: OK, I re-reviewed #5085, and this line is not hiding a bug, it genuinely is dead.

Reference, implementation of `copyArrayLiteral()`
https://github.com/dlang/dmd/pull/5085/files#diff-a86b7363d77bfa5ce809fa1d3985a086R309

https://github.com/dlang/dmd/blob/716303a41a642fb3ce2c8d0d95b0a7df4047713a/src/dmd/ctfeexpr.d#L299-L305

The call to `copyLiteral().copy()` was moved here.

WalterBright: Thanks for doing the research.

WalterBright (approved):

PR #7710, opened 2018-01-14
Move glue layer to `dmd.ir`

wilzbach:
Motivation
----------

- Makes it clear what the glue layer is and is to find where it's called
- The DMD as a library shouldn't need to access `dmd.ir` - this makes it explicit
- To keep things simple, this PR just moves the files - `dmd.ir.glue_layer` and `dmd.ir.objc_glue` are still required by the frontend

Script
------

```bash
replacements=(
"dmsc"
"e2ir"
"eh"
"glue"
"gluelayer"
"iasm"
"irstate"
"objc_glue"
"s2ir"
"tocsym"
"toctype"
"tocvdebug"
"todt"
"toir"
"toobj"
)
shopt -s globstar

    w=($r)
    sed "s/dmd\.${w[0]}/dmd.ir.${w[0]}/g" -i **/*.d
done
```


rainers:
You seem to have missed dmd.irstate in your replacement list.

WalterBright:
No, I do not wish to split dmd up into lots of packages. I find it far more convenient to use a flat structure. The only reason `root` is separate is because it was originally intended to be a separate library. `backend` simply has to be separated.

PR #7712, opened 2018-01-15
Don't overwrite the codecov/project status to green

wilzbach:
Now that we CodeCov properly find the base commit, it will only error if the patch reduces the coverage of a file, which shouldn't happen too often.

Follow-up to https://github.com/dlang/dmd/pull/7711

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -42,7 +42,7 @@ coverage:
         threshold: 0.1
     patch:
       default:
-        informational: true
+        informational: false
     changes:
       default:
         informational: true


wilzbach:
So I went through page 1 of the open PRs and those two PRs would have been flagged:

- https://github.com/dlang/dmd/pull/7639
- https://github.com/dlang/dmd/pull/7648

So it doesn't work well for refactorings or work with CTFE, so maybe sticking with `informational: true` and just being a bit more careful when looking at the diff is the best?

JinShil:
I'm torn on this.  I want it to be a requirement, but that red X is going to be just an annoyance if I have to visit each PR to see if it is only the codecov that's failing.  I have to make a binary choice, so I vote "No".

WalterBright:
As noted, it is impractical to require this. There's no substitute for the reviewer checking the uncovered new lines, and requiring that the ones that could be covered be covered, but these are a judgement call.

PR #6013, opened 2016-08-03
Update .codecov.yml

stevepeak:
Fixes mapping coverage reports back to src not test

PR #7713, opened 2018-01-15
[CodeCov] Remove fixing paths

wilzbach:
It seems that Codecov [now does check for the YAML of a PR](https://github.com/dlang/dmd/pull/7711#issuecomment-357589675).
Hence, we can give removing `fixes` a try - it looks like it's not needed anymore.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -47,9 +47,4 @@ coverage:
       default:
         informational: true
 
-  # https://docs.codecov.io/docs/fixing-paths
-  fixes:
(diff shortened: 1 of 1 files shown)


wilzbach:
Ah I just found https://github.com/dlang/dmd/pull/6013 and https://github.com/dlang/dmd/pull/6013/files#r73435640

> The issue is that codecov will resolve the path `file.lst` to `test/blah/file.d`.

Sorry about the noise then.

PR #7716, opened 2018-01-15
fixup for #7454 - always run asm tests with PIC

MartinNowak:
- as the purpose is really not testing call and symobl ABIs
- always use -fPIC so we're compatible with PIE-by-default platforms


JinShil:
This appears to be trying to solve a non-existent problem.  The test works for both PIC and non-PIC builds.  If we pull this we lose the test on non-PIC builds and all the work and knowledge that went into creating it.  I don't see that there's anything to "fix".

Furthermore, @MartinNowak, where were you a month ago when we were working on this and pinging you about it?

wilzbach:
How about we making the testsuite _runnable_ on a `PIE` hardened system first?
See e.g. https://github.com/dlang/dmd/pull/7420

MartinNowak:
> This appears to be trying to solve a non-existent problem. The test works for both PIC and non-PIC builds. If we pull this we lose the test on non-PIC builds and all the work and knowledge that went into creating it.

It's a redundant test that checks the same thing, b/c vector filling doesn't depend on PIC.
Hence it's confusing that we make this distinction here. Mainly the asm compare method is still flawed and contain the pre- and postambles of the test functions, that what differs between the two.
The next person (or me) will just copy and paste the PIC stuff turning it into useless cargo cult.

> I don't see that there's anything to "fix".

Check again, I've stripped PIC from the newly added test_cdstrpar because I couldn't see any sense in it. It is likely suffering from the same problem you were solving for test_cdvecfill.

JinShil:
> How about we making the testsuite runnable on a PIE hardened system first?
> See e.g. #7420

Exactly! I'll revisit this PR when we have the test suite running on PIE hardened systems and added to the CIs.

JinShil (dismissed):
Ok, It finally dawned on me what @MartinNowak is saying with regard to the redundancy of the existing implementation.  Therefore approving.

JinShil:
Auto-merge toggled off.  Waiting for feedback from others on https://github.com/dlang/dmd/pull/7427#issuecomment-357837366

JinShil (approved):

PR #7717, opened 2018-01-16
Fix __traits(isDeprecated) for complex types

wilzbach:
Introduced in https://github.com/dlang/dmd/pull/7665


src/dmd/traits.d
```diff
@@ -504,7 +504,7 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
     {
         if (global.params.vcomplex)
         {
-            if (isTypeX(t => t.iscomplex() || t.isimaginary()))
+            if (isTypeX(t => t.iscomplex() || t.isimaginary()).value == 1)
```
wilzbach: `isTypeX` returns a class instance (now `IntegerExp`) which will always be `true` ...

JinShil: Gah!

JinShil (approved):

src/dmd/traits.d
ibuclaw: And you couldn't have used `isTypeX(...).isBool(true)`?

wilzbach: -> https://github.com/dlang/dmd/pull/7722

PR #7720, opened 2018-01-16
Don't permutate through ARGS on AppVeyor

wilzbach:
We don't do this on CircleCi or SemaphoreCI either.
Auto-tester will do this anyways for correctness and AppVeyor is only intended to give us a feedback about MSCOFF builds and building DMD with VisualD.

### DIFF
--- a/appveyor.sh
+++ b/appveyor.sh
@@ -80,5 +80,5 @@ export MODEL="64"
 export MODEL_FLAG="-m64"
 
 cd /c/projects/dmd/test
-../../gnumake/make -j2 all MODEL=$MODEL MODEL_FLAG=$MODEL_FLAG LIB="../../phobos;$LIB"
+../../gnumake/make -j3 all MODEL=$MODEL ARGS="-O -inline -g" MODEL_FLAG=$MODEL_FLAG LIB="../../phobos;$LIB"
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -13,6 +13,10 @@ cache:
   - C:\projects\gnumake\make.exe
   - C:\projects\VisualD-v0.45.1-rc2.exe
 
+skip_commits:
+  # Avoid retesting the merged PR on `master` or `stable`
+  message: /^Merge pull request/
+
 artifacts:
   - path: src/dmd.exe
     name: dmd


wilzbach:
We save ~9 minutes :tada: 

image

(this adds/minus up over a day and leads to a thinner queue)

PetarKirov (approved):

appveyor.sh
```diff
@@ -80,5 +80,5 @@ export MODEL="64"
 export MODEL_FLAG="-m64"
 
 cd /c/projects/dmd/test
-../../gnumake/make -j2 all MODEL=$MODEL MODEL_FLAG=$MODEL_FLAG LIB="../../phobos;$LIB"
+../../gnumake/make -j3 all MODEL=$MODEL ARGS="-O -inline -release" MODEL_FLAG=$MODEL_FLAG LIB="../../phobos;$LIB"
```
ZombineDev: On second thoughts, I would suggest ARGS="-O -inline -g -debug" as it would have a slightly higher chance of catching bugs.

wilzbach: Hmm you mean better error messages on failures? I think knowing which test failed is already enough info and also `-release` does more interesting things, so isn't the chance that it will catch more bugs higher with `-release`?

MartinNowak: The testee here is dmd built as 64-bit COFF binary so that we can ship that at some point.
From that perspective we want to maximize coverage in dmd itself, but don't care too much about the output, other than it should be correct.
I think we definitely want to exercise the C++ backend parts, that implies `-O` and `-g`.
Whether or not `-release` or `-debug` hardly matters, as it has small impact on codegen, and we already know from the Win32/64 auto-test that dmd can generate working COFF code.

In fact the output of Win32/64 cross-compilation and Win64/64 should be identical, no?
Not sure what the state of reproducible builds on Windows is, but if we have a chance of comparing builds from the 2 host compilers, that would be the most ensuring test. Even though a hard to debug one if sth. differs.

skl131313:
Do you have [rolling builds](https://www.appveyor.com/docs/build-configuration/#rolling-builds) enabled? It's in the settings tab for the project in appveyor if you don't. It's nice for pull requests, so that only the latest commit is tested. But it is too bad it can't be disabled just for the master branch.

wilzbach:
> Do you have rolling builds enabled? It's in the settings tab for the project in appveyor if you don't. It's nice for pull requests, so that only the latest commit is tested. 

Nice. I was looking for exactly this! Thanks!!

> But it is too bad it can't be disabled just for the master branch.

I'm actually thinking about disabling builds on `master` too. Every PR is merged with master already and thus the benefit of running the build again at `master` is very slim - especially because AppVeyor isn't intended to be a replacement for auto-tester, but just a CI to test the MSCOFF builds of the compiler + VisualD build.
As apparently excluding branches on the web interface doesn't work, I will append this to the YAML here.

appveyor.sh
rainers: With `-release`, we disable 99% of the tests as no asserts are checked anymore. `-debug` doesn't have an impact because the debug condition is hardly ever used.
`-g` makes sense as it covers debug info generation.

wilzbach:
> With -release, we disable 99% of the tests as no asserts are checked anymore. -debug doesn't have an impact because the debug condition is hardly ever used.
-g makes sense as it covers debug info generation.

Ok. I removed `-release` and added `-g` - so are we all good here?
The speedup isn't that huge anymore. The current run took 21min 39s - it's not as good as 17min, but still better then 26-27min.

wilzbach:
> As apparently excluding branches on the web interface doesn't work, I will append this to the YAML here.

Adding branches to the YAML will probably prevent them from being run when a PR is opened.
Hence, I went with [skipping commits](https://www.appveyor.com/docs/how-to/filtering-commits/):

```YAML
skip_commits:
  # Avoid retesting the merged PR on `master` or `stable`
  message: /^Merge pull request/
```

PR #7724, opened 2018-01-16
Automatically convert TRUSTXXX -> TRUST

wilzbach:
```bash
replacements=(
"TRUSTdefault default_"
"TRUSTsystem system"
"TRUSTtrusted trusted"
"TRUSTsafe safe"
)
shopt -s globstar
    w=($r)
    sed "s/${w[0]}/TRUST.${w[1]}/g" -i **/*.d
done
```

### DIFF
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -3859,13 +3859,13 @@ struct ASTBase
             if (stc & STC.scope_)
                 this.isscope = true;
 
-            this.trust = TRUSTdefault;
+            this.trust = TRUST.default_;
             if (stc & STC.safe)
-                this.trust = TRUSTsafe;
+                this.trust = TRUST.safe;
             if (stc & STC.system)
-                this.trust = TRUSTsystem;
+                this.trust = TRUST.system;
             if (stc & STC.trusted)
-                this.trust = TRUSTtrusted;
+                this.trust = TRUST.trusted;
         }
 
         override Type syntaxCopy()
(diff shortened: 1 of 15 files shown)


PetarKirov (approved):

PetarKirov (approved):

src/dmd/mtype.h
```diff
@@ -549,10 +549,10 @@ enum RET
...
 enum TRUST
 {
-    TRUSTdefault = 0,
-    TRUSTsystem = 1,    // @system (same as TRUSTdefault)
-    TRUSTtrusted = 2,   // @trusted
-    TRUSTsafe = 3,      // @safe
+    default_ = 0,
+    system = 1,    // @system (same as TRUST.default)
+    trusted = 2,   // @trusted
+    safe = 3,      // @safe
```
ZombineDev: Fails with:
```
dmd/mtype.h:553: error: 'system' redeclared as different kind of symbol
/usr/include/stdlib.h:113: error: previous declaration of 'int system(const char*)'
```

Maybe suffix all enum members with an underscore?

wilzbach: The problem was that enum doesn't create its own scope in C++ 98.

ibuclaw (dismissed):
Don't bother updating the headers, it's doing no one any favours. Recent changes should be reverted.

wilzbach:
> Don't bother updating the headers, it's doing no one any favours. Recent changes should be reverted.

Cool! Done.

PR #7727, opened 2018-01-16
Automatically convert PASSXXX -> PASS

wilzbach:
```bash
replacements=(
"PASSinit init"
"PASSsemantic semantic"
"PASSsemanticdone semanticdone"
"PASSsemantic2 semantic2"
"PASSsemantic2done semantic2done"
"PASSsemantic3 semantic3"
"PASSsemantic3done semantic3done"
"PASSinline inline"
"PASSinlinedone inlinedone"
"PASSobj obj"
)
shopt -s globstar
    w=($r)
    sed "s/${w[0]}/PASS.${w[1]}/g" -i **/*.d
done
```

### DIFF
--- a/src/dmd/access.d
+++ b/src/dmd/access.d
@@ -559,7 +559,7 @@ private Dsymbol mostVisibleOverload(Dsymbol s)
         {
             assert(ad.isOverloadable, "Non overloadable Aliasee in overload list");
             // Yet unresolved aliases store overloads in overnext.
-            if (ad.semanticRun < PASSsemanticdone)
+            if (ad.semanticRun < PASS.semanticdone)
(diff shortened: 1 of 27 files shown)


ZombineDev:
You also need to update:
https://github.com/dlang/dmd/blob/4c0f3e29507e53d2f631259db9f85123efb81a4d/src/dmd/dsymbol.h#L122-L134

wilzbach:
> You also need to update:

I didn't because C++ 98 (the version GDC uses) doesn't support scoped enums at the root level. However, I now used the namespace trick. CC @ibuclaw

JinShil (approved):

PR #7728, opened 2018-01-16
Automatically convert RECXXXX -> AliasThisRec

wilzbach:
```
replacements=(
"RECno no"
"RECyes yes"
"RECfwdref fwdref"
"RECtypeMask typeMask"
"RECtracing tracing"
"RECtracingDT tracingDT"
)
shopt -s globstar
    w=($r)
    sed "s/${w[0]}/AliasThisRec.${w[1]}/g" -i **/*.d
done
```

### DIFF
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -298,20 +298,13 @@ struct ASTBase
 
     enum AliasThisRec : int
     {
-        RECno           = 0,    // no alias this recursion
-        RECyes          = 1,    // alias this has recursive dependency
-        RECfwdref       = 2,    // not yet known
-        RECtypeMask     = 3,    // mask to read no/yes/fwdref
-        RECtracing      = 0x4,  // mark in progress of implicitConvTo/deduceWild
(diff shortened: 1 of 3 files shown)


ZombineDev:
You also need to update this: https://github.com/dlang/dmd/blob/254c9634f8498c3d7eb38f37de23b14f1c509e36/src/dmd/mtype.h#L716-L725

wilzbach:
> You also need to update this:

I didn't because C++ 03 (the version GDC uses) doesn't support scoped enums at the root level. However, I now used the namespace trick. CC @ibuclaw

jacob-carlborg (approved):

ibuclaw (dismissed):
Why do you need to change the headers at all here?

wilzbach:
> Why do you need to change the headers at all here?

I don't. That's why I left them alone initially, but there was https://github.com/dlang/dmd/pull/7728#issuecomment-357923118

PetarKirov (approved):

PR #7730, opened 2018-01-16
Revert "Use native AAs for Mangler.types and idents"

WalterBright:
Reverts dlang/dmd#7694 because of the slowdown for double lookups.


wilzbach:
SemaphoreCI didn't like the `revert` branches (they weren't on the whitelist) - they are now.

wilzbach (approved):
Going to merge this -> a revert of this revert (aka fix) can be put into the the PR loop again.

PR #7731, opened 2018-01-16
highlight expression.d error messages

JinShil:


wilzbach (approved):

wilzbach:
Is this the last one?

image

Noice!

ZombineDev:
Nice work @JinShil, project complete!

JinShil:
> Is this the last one?

Unfortunately, no.  While I was updating the tests, I found some error messages that weren't yet highlighted, so I made a list.  I know need to grep with that list and figure out where they are and update them.  I also have a few inconsistencies that I would like to resolve.  Finally, the `verror()` code needs to be changed to solve a couple of unique cases.  So sorry, but expect 2 or 3 more PRs from me this week.

PR #7734, opened 2018-01-17
AssertExp: add links to spec

WalterBright:
This is a sample of spec URLs I'd like to add to the code. It makes comparing the implementation with the spec very convenient, so that they can be kept in sync.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -5149,6 +5149,7 @@ extern (C++) final class ImportExp : UnaExp
 }
 
 /***********************************************************
+ * https://dlang.org/spec/expression.html#assert_expressions
  */
 extern (C++) final class AssertExp : UnaExp
 {
--- a/src/dmd/e2ir.d
+++ b/src/dmd/e2ir.d
@@ -1953,6 +1953,7 @@ elem *toElem(Expression e, IRState *irs)
 
         override void visit(AssertExp ae)
         {
+            // https://dlang.org/spec/expression.html#assert_expressions
             //printf("AssertExp.toElem() %s\n", toChars());
             elem *e;
             if (global.params.useAssert == CHECKENABLE.on)
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -4315,6 +4315,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
 
     override void visit(AssertExp exp)
     {
+        // https://dlang.org/spec/expression.html#assert_expressions
         static if (LOGSEMANTIC)
         {
             printf("AssertExp::semantic('%s')\n", exp.toChars());
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -7596,6 +7596,7 @@ final class Parser(AST) : Lexer
             }
         case TOKassert:
             {
+                // https://dlang.org/spec/expression.html#assert_expressions
                 AST.Expression msg = null;
 
                 nextToken();


src/dmd/expression.d
```diff
@@ -5152,6 +5152,7 @@ extern (C++) final class ImportExp : UnaExp
  */
 extern (C++) final class AssertExp : UnaExp
 {
+    // https://dlang.org/spec/expression.html#assert_expressions
```
wilzbach: Why don't you add this in the Ddoc header of `AssertExp`?

wilzbach (approved):
> so that they can be kept in sync.

Hehe experience shows that we are pretty bad in manually keeping things in sync.
It would be cooler if we can generate the spec from the frontend, see e.g. https://github.com/dlang/dmd/pull/6877 for a proof of concept.

Or for the more complex cases verify the compiler against a formal spec.

JinShil:
@WalterBright, you need a modern IDE.

image

The URL is only specified in one place:  the type declaration, but is easily visible anywhere the type used.  No need to litter the code with redundant URLs that will likely need updating in a few years time.

src/dmd/expression.d
WalterBright: ok

WalterBright:
@JinShil What file does that URL exist in?

JinShil:
> What file does that URL exist in?

The URL can exist in any file.  When the file is imported, the code-completion/intellisense logic of the editor reads the imported file and and parses the DDoc comments.  Therefore, the URL just needs to exist in the file containing the type declaration.

For the D ecosystem, the primary utility providing this feature is @Hackerpilot's fantastic [D Completion Daemon](https://github.com/dlang-community/DCD).  It's general enough to incorporate in any code editor (if you know how to hack the editor).  And it's open source and written in D, so you can customize it yourself using your favorite language.

WalterBright:
> Therefore, the URL just needs to exist in the file containing the type declaration.

Except it currently does not exist anywhere in the source tree. In any case, you're assuming that browsing source code is only done using the Daemon. This is far from the case - github itself does not do this, and github is majorly used for file browsing. It doesn't work when typing files, when using a file comparator, any editor that you don't have control over, etc.

Links in the source code have already proved very useful - the header links, and the bugzilla links.

If the location of the spec moves, the URLs can be mass updated with a simple script. That's a small price to pay for the convenience.

As additional evidence that your proposed method is ineffective, I'll point out that the spec and the implementation have frequently diverged, implying that few ever compare the spec with the code. I want to make it *easy* to compare the two.

JinShil:
> Except it currently does not exist anywhere in the source tree.

All you have to do is add the URL to the type declaration's ddoc header.

> In any case, you're assuming that browsing source code is only done using the Daemon

You don't browse the source code with the daemon.  The daemon is accessed through the features of your editor.  If your editor doesn't support adding extensions like this, you need a better editor.

> github is majorly used for file browsing

Browsing code with GitHub is like swimming in maple syrup.  If you're using GitHub to browse source code, you're doing it wrong.  Clone the repository locally, open the folder in your editor, and browse it there, if you want to be productive and efficient.

> Links in the source code have already proved very useful

Indeed, the URLs are very useful, but they should exist in one centralized place. and code editing/browsing utilities should be leveraged rather than redundantly copying the same URL at points of usage.

> As additional evidence that your proposed method is ineffective, I'll point out that the spec and the implementation have frequently diverged, implying that few ever compare the spec with the code. I want to make it easy to compare the two.

Once again, I'm not objecting to adding the URL to the code;  adding the URL to the code is a great idea.  I'm objecting to repeatedly adding the same URL in the code.  Just put it in the DDoc header of the type declaration; your editor should do the rest.

JinShil:
> If the location of the spec moves, the URLs can be mass updated with a simple script. That's a small price to pay for the convenience.

Your call, I've had my say.

JinShil:
Here's an example using [Visual Studio Code](https://code.visualstudio.com/) and the [code-d serve-d extension](https://marketplace.visualstudio.com/items?itemName=webfreak.code-d-beta) made by fine members of the D community (namely @WebFreak001).  It leverages the aforementioned D Completion Daemon.

Here's an example I just through together

dmd/expression.d
```D
module dmd.expression;

abstract class expression{}

/*********************************
* See the spec at http://dlang.org/spec/expression.html#expression
*/
class assertExp : expression {}
```

dmd/expressionsem.d
```D
module dmd.expressionsem;

import dmd.expression;

class visitor
{
    void visit(assertExp exp) {}
}
```

The previous image I posted was using Visual Studio 2017 on Windows.  This on uses Visual Studio Code on Linux with the serve-d/code-d plugin.

image

WalterBright:
Those look like nice projects. But I don't use an IDE, and don't want the source code to be organized as "best with an IDE". The URLs are something I'd like to have for my own use. It's just a comment here and there, if it's not useful to others they can just ignore them.

JinShil:
> But I don't use an IDE, and don't want the source code to be organized as "best with an IDE"

Understood, but also understand that I don't want the source code to be organized as "best used with Walter's lack of an IDE".

Just to clarify, you don't need an IDE.  You just need the D Completion Daemon, and some integration code with your editor.  Don't you already customize and build your own editor, anyway?  You should give the D Completion Daemon a try; I'm certain you won't be able to go back once you've got it up and running.

WalterBright:
Maybe some day, but not now. I will reiterate that it's just a comment here and there. Nobody's use of an IDE is degraded by it in any way.

Why do I use microemacs? Because it is small, lightweight, and works everywhere. Including in a tty window into a remote machine, which I use daily. I could use vim or emacs instead, but those are monster programs. Nothing else I'm aware of has the ubiquity.

JinShil:
Anyway, Walter, I've had my say and said more than I intended to.  Your judgement trumps mine, and rightly so,  I'm not trying to veto this.  If you're determined, let's merge this so we can get back to the other stuff in the queue.

wilzbach:
> let's merge this so we can get back to the other stuff in the queue.

Yes, literally _every_ PR in the queue is more important than this ;-)

WalterBright:
thank you

PR #7735, opened 2018-01-18
Automatically convert LINKXXX to LINK + use final switch

wilzbach:
```bash
replacements=(
"LINKdefault default_"
"LINKd d"
"LINKc c"
"LINKcpp cpp"
"LINKwindows windows"
"LINKpascal pascal"
"LINKobjc objc"
"LINKsystem system"
)
shopt -s globstar
    w=($r)
    sed "s/${w[0]}/LINK.${w[1]}/g" -i **/*.d
done
```

### DIFF
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -485,7 +485,7 @@ struct ASTBase
             super(id);
             storage_class = STC.undefined_;
             protection = Prot(Prot.Kind.undefined);
-            linkage = LINKdefault;
+            linkage = LINK.default_;
         }
 
         override final inout(Declaration) isDeclaration() inout
@@ -6280,19 +6280,19 @@ struct ASTBase
     {
         switch (linkage)
         {
-        case LINKdefault:
+        case LINK.default_:
             return null;
-        case LINKd:
+        case LINK.d:
             return "D";
-        case LINKc:
+        case LINK.c:
             return "C";
-        case LINKcpp:
+        case LINK.cpp:
             return "C++";
-        case LINKwindows:
+        case LINK.windows:
             return "Windows";
-        case LINKpascal:
+        case LINK.pascal:
             return "Pascal";
-        case LINKobjc:
+        case LINK.objc:
             return "Objective-C";
         default:
             assert(0);
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -6278,9 +6278,10 @@ struct ASTBase
 
     extern (C++) static const(char)* linkageToChars(LINK linkage)
     {
-        switch (linkage)
+        final switch (linkage)
         {
         case LINK.default_:
+        case LINK.system:
             return null;
         case LINK.d:
             return "D";
@@ -6294,8 +6295,6 @@ struct ASTBase
             return "Pascal";
         case LINK.objc:
             return "Objective-C";
-        default:
-            assert(0);
         }
     }
 
--- a/src/dmd/backend/ty.d
+++ b/src/dmd/backend/ty.d
@@ -74,7 +74,7 @@ enum
     TYifunc             = 0x2E, // interrupt func
     TYptr               = 0x33, // generic pointer type
     TYmfunc             = 0x37, // NT C++ member func
-    TYjfunc             = 0x38, // LINKd D function
+    TYjfunc             = 0x38, // LINK.d D function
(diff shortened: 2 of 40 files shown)


src/dmd/dmangle.d
```diff
@@ -543,25 +543,23 @@ public:
...
+                case LINK.windows:
+                case LINK.pascal:
+                case LINK.objc:
                     return d.ident.toChars();
-                case LINKcpp:
+                case LINK.cpp:
                     return Target.toCppMangle(d);
-                case LINKdefault:
+                case LINK.def:
+                case LINK.system:
```
wilzbach: I wasn't sure what the behavior of `LINK.system` should be here. FWIW `LINK.system` feels like a hack.

src/dmd/json.d
```diff
@@ -293,30 +293,35 @@ public:
...
             property(name, "windows");
             break;
-        case LINKpascal:
+        case LINK.pascal:
             property(name, "pascal");
             break;
-        default:
-            assert(false);
+        case LINK.objc:
+            property(name, "objc");
```
wilzbach: `final switch` for the win!

src/dmd/target.d
```diff
@@ -478,7 +484,7 @@ struct Target
      */
     extern (C++) static LINK systemLinkage()
     {
-        return global.params.isWindows ? LINKwindows : LINKc;
+        return global.params.isWindows ? LINK.windows : LINK.c;
```
wilzbach: This is why `LINK.system` feels like a hack. Should `systemLinkage` be called everywhere instead of using `LINK.system`?

wilzbach: Hmm `LINK.system` is only set once by the parser:

https://github.com/wilzbach/dmd/blob/089241b1359d1a20f19dd1b589ff0922854b16bc/src/dmd/parse.d#L2208-L2211

JinShil (changes requested):

src/dmd/astbase.d
```diff
@@ -6278,24 +6278,23 @@ struct ASTBase
 
     extern (C++) static const(char)* linkageToChars(LINK linkage)
     {
-        switch (linkage)
+        final switch (linkage)
         {
-        case LINKdefault:
+        case LINK.def:
```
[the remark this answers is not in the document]
wilzbach: I didn't change that:

https://github.com/dlang/dmd/blob/d9421bde847c24a8f8402b885e594859c394513c/src/dmd/globals.d#L415-L425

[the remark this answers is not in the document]
JinShil: Yes, I know.  But this is the opportune time to change it.  `def` is a pretty lousy name, and `default_` is also used elsewhere.  While we're refactoring these we have the opportunity to make them all consistent.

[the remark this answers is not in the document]
JinShil: I'll submit a change for #7737 if you do this one.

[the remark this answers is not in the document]
wilzbach: Ok changed, I can change #7737 too - I don't mind.

> But this is the opportune time to change it.

Shouldn't we rename the enum to PascalCase too then?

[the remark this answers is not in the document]
wilzbach: > the same should apply for #7737

https://github.com/dlang/dmd/pull/7745

src/dmd/backend/ty.d
```diff
@@ -74,7 +74,7 @@ enum
     TYifunc             = 0x2E, // interrupt func
     TYptr               = 0x33, // generic pointer type
     TYmfunc             = 0x37, // NT C++ member func
-    TYjfunc             = 0x38, // LINKd D function
+    TYjfunc             = 0x38, // LINK.d D function
```
ibuclaw: I wonder if the `j` stands for Jupiter also.

src/dmd/json.d
ibuclaw: Add test. ;-)

ibuclaw (approved):
Ok

jacob-carlborg (approved):
Approved assuming the test that Iain mentioned is added.

wilzbach:
> final switch for the win!
> Add test. ;-)

Urgh. The JSON output didn't print `linkage` information at all. Glad that we have Coverage annotation :)
-> it was trivial to fix & I added a test for it.

JinShil (approved):

PR #7737, opened 2018-01-18
Convert PINLINEXXX to PINLINE and use final switch

wilzbach:
```bash
replacements=(
"PINLINEdefault def"
"PINLINEnever never"
"PINLINEalways always"
)
shopt -s globstar
    w=($r)
    sed "s/${w[0]}/PINLINE.${w[1]}/g" -i **/*.d
done
```

### DIFF
--- a/src/dmd/inline.d
+++ b/src/dmd/inline.d
@@ -1513,7 +1513,7 @@ bool canInline(FuncDeclaration fd, bool hasthis, bool hdrscan, bool statementsTo
         break;
     }
 
-    switch (fd.inlining)
+    final switch (fd.inlining)
     {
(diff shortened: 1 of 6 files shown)


src/dmd/inline.d
```diff
@@ -1513,16 +1513,14 @@ bool canInline(FuncDeclaration fd, bool hasthis, bool hdrscan, bool statementsTo
         break;
     }
 
-    switch (fd.inlining)
+    final switch (fd.inlining)
```
wilzbach: `final switch` for the win!

JinShil (approved):

PR #7738, opened 2018-01-18
Automatically convert TARGET_XXXX to a TARGET enum

wilzbach:
```bash
replacements=(
"TARGET_LINUX linux"
"TARGET_OSX OSX"
"TARGET_FREEBSD FreeBSD"
"TARGET_OPENBSD OpenBSD"
"TARGET_SOLARIS Solaris"
"TARGET_WINDOS Windows"
)
shopt -s globstar
    w=($r)
    sed "s/${w[0]}/TARGET.${w[1]}/g" -i **/*.d
done
```

### DIFF
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -27,12 +27,15 @@ template xversion(string s)
 
 enum IN_GCC     = xversion!`IN_GCC`;
 
-enum TARGET_LINUX   = xversion!`linux`;
-enum TARGET_OSX     = xversion!`OSX`;
-enum TARGET_FREEBSD = xversion!`FreeBSD`;
-enum TARGET_OPENBSD = xversion!`OpenBSD`;
-enum TARGET_SOLARIS = xversion!`Solaris`;
-enum TARGET_WINDOS  = xversion!`Windows`;
+enum TARGET : bool
+{
+    Linux   = xversion!`linux`,
+    OSX     = xversion!`OSX`,
+    FreeBSD = xversion!`FreeBSD`,
+    OpenBSD = xversion!`OpenBSD`,
+    Solaris = xversion!`Solaris`,
+    Windows = xversion!`Windows`,
+}
 
 enum CHECKENABLE : ubyte
 {
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -293,11 +293,11 @@ struct Global
 
     extern (C++) void _init()
     {
-        static if (TARGET_WINDOS)
+        static if (TARGET.Windows)
         {
             obj_ext = "obj";
         }
-        else static if (TARGET_LINUX || TARGET_OSX || TARGET_FREEBSD || TARGET_OPENBSD || TARGET_SOLARIS)
+        else static if (TARGET.Linux || TARGET.OSX || TARGET.FreeBSD || TARGET.OpenBSD || TARGET.Solaris)
         {
             obj_ext = "o";
         }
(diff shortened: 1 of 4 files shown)


src/dmd/globals.d
```diff
@@ -27,12 +27,15 @@ template xversion(string s)
...
 
-enum TARGET_LINUX   = xversion!`linux`;
-enum TARGET_OSX     = xversion!`OSX`;
-enum TARGET_FREEBSD = xversion!`FreeBSD`;
-enum TARGET_OPENBSD = xversion!`OpenBSD`;
-enum TARGET_SOLARIS = xversion!`Solaris`;
-enum TARGET_WINDOS  = xversion!`Windows`;
+enum TARGET
+{
+    linux   = xversion!`linux`,
```
JinShil: Capital `L' for Linux?

wilzbach: I kept the same smelling as `version(...)` uses to avoid confusions.

JinShil: All other `version`s and `Target`s use proper case.  It appears that `version(linux)` was a mistake.  I vote to not proliferate that mistake.

wilzbach: Ok. Fair enough ;-)

src/dmd/globals.d
```diff
@@ -27,12 +27,15 @@ template xversion(string s)
...
-enum TARGET_SOLARIS = xversion!`Solaris`;
-enum TARGET_WINDOS  = xversion!`Windows`;
+enum TARGET
+{
+    Linux   = xversion!`linux`,
+    OSX     = xversion!`OSX`,
+    FreeBSD = xversion!`FreeBSD`,
+    OpenBSD = xversion!`OpenBSD`,
+    Solaris = xversion!`Solaris`,
+    Windows = xversion!`Windows`,
```
ibuclaw: At some point these should be user defined versions passed by the make file.

src/dmd/globals.d
```diff
@@ -27,12 +27,15 @@ template xversion(string s)
 
 enum IN_GCC     = xversion!`IN_GCC`;
 
-enum TARGET_LINUX   = xversion!`linux`;
-enum TARGET_OSX     = xversion!`OSX`;
-enum TARGET_FREEBSD = xversion!`FreeBSD`;
-enum TARGET_OPENBSD = xversion!`OpenBSD`;
-enum TARGET_SOLARIS = xversion!`Solaris`;
-enum TARGET_WINDOS  = xversion!`Windows`;
+enum TARGET
```
ibuclaw: `enum : bool`

src/dmd/backend/cc.d
```diff
@@ -58,7 +58,7 @@ enum WM
     WM_ccast        = 25,
     WM_obsolete     = 26,
 
-    // if TARGET_LINUX || TARGET_OSX || TARGET_FREEBSD || TARGET_OPENBSD || TARGET_SOLARIS
+    // if TARGET.linux || TARGET.OSX || TARGET.FreeBSD || TARGET.OpenBSD || TARGET.Solaris
```
jacob-carlborg: `TARGET.linux` -> `TARGET.Linux`.

ibuclaw (commented):
I just wonder if these are better as manifest constants.

jacob-carlborg (changes requested):

src/dmd/globals.d
ibuclaw: (nothing for you to do now).

jacob-carlborg: At some point the compiler should support cross-compiling then all these checks need to be runtime checks 😉.

ibuclaw: That's Phase two. ☺️

jacob-carlborg (approved):

src/dmd/globals.d
JinShil: Perhaps a nested enum: `Target.Kind`?

wilzbach: Hmm. As @ibuclaw mentioned `Target` needs to be refactored anyways and at the moment it would be rather empty:

```
struct Target
{
    enum Kind { ...}
}


Also I recall that @ZombineDev wanted to introduce TargetTriplets.
-> I think we can do this change once the time comes?

JinShil: > I think we can do this change once the time comes?

Ok.

JinShil (approved):

wilzbach:
> enum : bool

Oh I only saw this now -> updated.

PR #7742, opened 2018-01-18
Convert TXXXX to an enum

wilzbach:
```bash
replacements=(
"Tarray array"
"Tsarray sarray"
"Taarray aarray"
"Tpointer pointer"
"Treference reference"
"Tfunction function_"
"Tident ident"
"Tclass class_"
"Tstruct struct_"
"Tenum enum_"
"Tdelegate delegate_"
"Tnone none"
"Tvoid void_"
"Tint8 int8"
"Tuns8 uns8"
"Tint16 int16"
"Tuns16 uns16"
"Tint32 int32"
"Tuns32 uns32"
"Tint64 int64"
"Tuns64 uns64"
"Tfloat32 float32"
"Tfloat64 float64"
"Tfloat80 float80"
"Timaginary32 imaginary32"
"Timaginary64 imaginary64"
"Timaginary80 imaginary80"
"Tcomplex32 complex32"
"Tcomplex64 complex64"
"Tcomplex80 complex80"
"Tbool bool_"
"Tchar char_"
"Twchar wchar_"
"Tdchar dchar_"
"Terror error"
"Tinstance instance"
"Ttypeof typeof_"
"Ttuple tuple"
"Tslice slice"
"Treturn return_"
"Tnull null_"
"Tvector vector"
"Tint128 int128"
"Tuns128 uns128"
"TMAX MAX"
)
shopt -s globstar
    w=($r)
    sed -E "s/([^a-zA-Z])${w[0]}/TY.${w[1]}/g" -i **/*.d
done
```

~~~- This should be the last PR in this conversion series~~~
~~~- I'm not too happy about the name `ENUMTY`, but it was the existing one. Better ideas?~~~
- I could use `with(ENUMTY)` for all the `switch` loops. This would probably also reduce the size of the diff
- The last commit with a few initial `final switch` shows why `enums` are cool

### DIFF
--- a/src/dmd/glue.d
+++ b/src/dmd/glue.d
@@ -738,11 +738,11 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
     if (fd.semanticRun >= PASS.obj) // if toObjFile() already run
         return;
 
-    if (fd.type && fd.type.ty == Tfunction && (cast(TypeFunction)fd.type).next is null)
+    if (fd.type && fd.type.ty == TY.function_ && (cast(TypeFunction)fd.type).next is null)
         return;
 
     // If errors occurred compiling it, such as https://issues.dlang.org/show_bug.cgi?id=6118
-    if (fd.type && fd.type.ty == Tfunction && (cast(TypeFunction)fd.type).next.ty == Terror)
+    if (fd.type && fd.type.ty == TY.function_ && (cast(TypeFunction)fd.type).next.ty == TY.error)
         return;
 
     if (fd.semantic3Errors)
@@ -990,7 +990,7 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
     //printf("linkage = %d, tyf = x%x\n", linkage, tyf);
     int reverse = tyrevfunc(s.Stype.Tty);
 
-    assert(fd.type.ty == Tfunction);
+    assert(fd.type.ty == TY.function_);
     TypeFunction tf = cast(TypeFunction)fd.type;
     RET retmethod = retStyle(tf);
     if (retmethod == RET.stack)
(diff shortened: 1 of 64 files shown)


wilzbach:
>  maybe just Ty

Already taken. `Tz` or `Te` is still free.

> or ... (I hesitate to say it) ... T.

Yes, that's too generic.

wilzbach:
>> maybe just Ty
> Already taken.

Correction - only `TY` and `ty` are taken.

https://github.com/dlang/dmd/blob/e1051b0fe9eed1985c296edc0e0be0fd7b0c5cdb/src/dmd/mtype.d#L463

> Anyway, definitely get rid of the T prefix on all the member names. 

Okay. done.

JinShil:
> Already taken.

Where is `Ty` taken?  I can't find it.  I think that might be the best choice.

jacob-carlborg:
How about `Kind` nested in `Type`?

jacob-carlborg:
> Correction - only TY and ty are taken

@wilzbach as far as I can see `Type.ty` is of type `TY`, which is an alias to `ubyte`. But all the values that are used for `Type.ty` are of type `ENUMTY`.

wilzbach:
> @wilzbach as far as I can see Type.ty is of type TY, which is an alias to ubyte. But all the values that are used for Type.ty are of type ENUMTY.

Great catch. Let's go with `TY` then?

src/dmd/glue.d
```diff
@@ -1495,7 +1495,15 @@ uint totym(Type tx)
...
         }
-        default:
+        case TY.none:
+        case TY.error:
+        case TY.instance:
+        case TY.tuple:
+        case TY.slice:
+        case TY.return_:
+        case TY.int128:
+        case TY.uns128:
```
wilzbach: Looks like the cent PRs need to touch this bit.

src/dmd/impcnvtab.d
```diff
@@ -14,17 +14,17 @@ module dmd.impcnvtab;
...
+immutable TY[TY.MAX][TY.MAX] impcnvType2 = impCnvTab.impcnvType2Tab;
 
 private:
 
 struct ImpCnvTab
 {
-    ENUMTY[TMAX][TMAX] impcnvResultTab;
-    ENUMTY[TMAX][TMAX] impcnvType1Tab;
-    ENUMTY[TMAX][TMAX] impcnvType2Tab;
+    TY[TY.MAX][TY.MAX] impcnvResultTab;
```
jacob-carlborg: `ENUMTY` was `int`, `TY` is now `ubyte`, meaning the size of `ImpCnvTab` is now changed. Not sure if that's a problem.

jacob-carlborg:
I think `Tag` is another common name of these kind of things.

src/dmd/impcnvtab.d
wilzbach: I don't think it's a problem because
- the `impcntab` tables are generated at CTFE and `immutable`
- C++ already sees them as `ubyte` tables:

https://github.com/dlang/dmd/blob/b8642c4f7d62cc94d12fee36fbce1f7617692e3c/src/dmd/mtype.h#L129-L134

JinShil:
> Correction - only TY and ty are taken.

Sorry @wilzbach, but what about `Ty` (capital T, lower-case y).  And I do prefer `Type.Kind`, if that's a possibility.

src/dmd/impcnvtab.d
jacob-carlborg: As long as it works I'm happy. Otherwise `alias ENUMTY = int;` can be added.

wilzbach:
> And I do prefer Type.Kind, if that's a possibility.

https://github.com/dlang/dmd/pull/7747

jacob-carlborg:
> #7747

Is #7747 replacing this PR?

wilzbach:
> Is #7747 replacing this PR?

Yes.

PR #7743, opened 2018-01-18
Convert PSXXXX to ParseStatementFlags

wilzbach:
```bash
replacements=(
"PSscope scope_"
"PScurlyscope curlyScope"
"PSsemi_ok semiOk"
"PScurly curly"
"PSsemi semi"
)
shopt -s globstar
    w=($r)
    sed "s/${w[0]}/ParseStatementFlags.${w[1]}/g" -i **/*.d
done
```

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -193,18 +193,12 @@ __gshared PREC[TOKMAX] precedence =
 
 enum ParseStatementFlags : int
 {
-    PSsemi          = 1,        // empty ';' statements are allowed, but deprecated
-    PSscope         = 2,        // start a new scope
-    PScurly         = 4,        // { } statement is required
-    PScurlyscope    = 8,        // { } starts a new scope
-    PSsemi_ok       = 0x10,     // empty ';' are really ok
+    semi          = 1,        // empty ';' statements are allowed, but deprecated
+    scope_        = 2,        // start a new scope
+    curly         = 4,        // { } statement is required
+    curlyScope    = 8,        // { } starts a new scope
+    semiOk        = 0x10,     // empty ';' are really ok
 }
-
-alias PSsemi = ParseStatementFlags.PSsemi;
-alias PSscope = ParseStatementFlags.PSscope;
-alias PScurly = ParseStatementFlags.PScurly;
-alias PScurlyscope = ParseStatementFlags.PScurlyscope;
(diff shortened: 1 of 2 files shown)


src/dmd/parse.d
```diff
@@ -193,19 +193,13 @@ __gshared PREC[TOKMAX] precedence =
...
 {
-    PSsemi          = 1,        // empty ';' statements are allowed, but deprecated
-    PSscope         = 2,        // start a new scope
-    PScurly         = 4,        // { } statement is required
-    PScurlyscope    = 8,        // { } starts a new scope
-    PSsemi_ok       = 0x10,     // empty ';' are really ok
+    semi          = 1,        // empty ';' statements are allowed, but deprecated
+    scope_        = 2,        // start a new scope
+    curly         = 4,        // { } statement is required
+    curlyscope    = 8,        // { } starts a new scope
```
JinShil: We should consider if we want to enforce the style guide here and make it `curlyScope`.

wilzbach: It would also be `semiOk` ...

JinShil: Yes.

wilzbach: Ok changed.

jacob-carlborg: Not that I'm complaining, but we haven't done that in any other of these PRs. All enum names are still all uppercase.

jacob-carlborg (approved):

src/dmd/parse.d
```diff
@@ -2640,7 +2634,7 @@ final class Parser(AST) : Lexer
             check(TOKrparen);
         }
 
-        auto fbody = parseStatement(PScurly);
+        auto fbody = parseStatement(ParseStatementFlags.curly);
         auto f = new AST.InvariantDeclaration(loc, token.loc, stc, null, fbody);
```
ZombineDev: `ParseStatementFlags` is quite long. How about adding a private `alias` at the top of the module:
```D
private alias PS = ParseStatementFlags;
//...
auto fbody = parseStatement(PS.curly);
```

jacob-carlborg: I disagree, it's not used *that* much.

JinShil: I'm OK with or without the alias.

JinShil (approved):

PR #7746, opened 2018-01-18
Allow verbose output to be directed to a file

marler8997:
Provide an option to direct verbose output to a file.  This can be used to allow tools to invoke the compiler and capture the verbose output without having to redirect stdout/stderr.  This is a potential solution to the rdmd "single invocation problem" and can also be used by other tools to gain information from the compiler without having to redirect stdout/stderr so that error messages and other information still gets back to the user.

EDIT: had originally pushed this to allow rdmd to support single invocation, but now I have another tool that can use it.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -143,8 +143,8 @@ extern (C++) void genCmain(Scope* sc)
     m.members = p.parseModule();
     assert(p.token.value == TOK.endOfFile);
     assert(!p.errors); // shouldn't have failed to parse it
-    bool v = global.params.verbose;
-    global.params.verbose = false;
+    auto v = global.params.verbose;
+    global.params.verbose = null;
     m.importedFrom = m;
     m.importAll(null);
     m.dsymbolSemantic(null);
@@ -474,9 +474,9 @@ private int tryMain(size_t argc, const(char)** argv)
 
     if (global.params.verbose)
     {
-        fprintf(global.stdmsg, "binary    %s\n", global.params.argv0);
-        fprintf(global.stdmsg, "version   %s\n", global._version);
-        fprintf(global.stdmsg, "config    %s\n", global.inifilename ? global.inifilename : "(none)");
+        fprintf(global.params.verbose, "binary    %s\n", global.params.argv0);
+        fprintf(global.params.verbose, "version   %s\n", global._version);
+        fprintf(global.params.verbose, "config    %s\n", global.inifilename ? global.inifilename : "(none)");
     }
     //printf("%d source files\n",files.dim);
 
@@ -668,7 +668,7 @@ private int tryMain(size_t argc, const(char)** argv)
     {
         Module m = modules[modi];
         if (global.params.verbose)
-            fprintf(global.stdmsg, "parse     %s\n", m.toChars());
+            fprintf(global.params.verbose, "parse     %s\n", m.toChars());
         if (!Module.rootModule)
             Module.rootModule = m;
         m.importedFrom = m; // m.isRoot() == true
@@ -725,7 +725,7 @@ private int tryMain(size_t argc, const(char)** argv)
         foreach (m; modules)
         {
             if (global.params.verbose)
-                fprintf(global.stdmsg, "import    %s\n", m.toChars());
+                fprintf(global.params.verbose, "import    %s\n", m.toChars());
             genhdrfile(m);
         }
     }
@@ -736,7 +736,7 @@ private int tryMain(size_t argc, const(char)** argv)
     foreach (m; modules)
     {
         if (global.params.verbose)
-            fprintf(global.stdmsg, "importall %s\n", m.toChars());
+            fprintf(global.params.verbose, "importall %s\n", m.toChars());
         m.importAll(null);
     }
     if (global.errors)
@@ -748,7 +748,7 @@ private int tryMain(size_t argc, const(char)** argv)
     foreach (m; modules)
     {
         if (global.params.verbose)
-            fprintf(global.stdmsg, "semantic  %s\n", m.toChars());
+            fprintf(global.params.verbose, "semantic  %s\n", m.toChars());
         m.dsymbolSemantic(null);
     }
     //if (global.errors)
@@ -769,7 +769,7 @@ private int tryMain(size_t argc, const(char)** argv)
     foreach (m; modules)
     {
         if (global.params.verbose)
-            fprintf(global.stdmsg, "semantic2 %s\n", m.toChars());
+            fprintf(global.params.verbose, "semantic2 %s\n", m.toChars());
         m.semantic2(null);
     }
     Module.runDeferredSemantic2();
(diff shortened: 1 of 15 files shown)


src/dmd/mars.d
```diff
@@ -1640,6 +1638,22 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
...
+                if (p[2] == '\0')
+                    params.verbose = stdout;
+                else
+                {
+                    auto filename = p + 2 + (p[2] == '=');
+                    params.verbose = fopen(filename, "w");
+                    if (!params.verbose)
+                    {
+                        import core.stdc.errno : errno;
+                        error("failed to open verbose output file '%s' (errno %d)", filename, errno);
```
ibuclaw: You know that there's a libc function that returns an error message for the given errno.

ibuclaw (commented):
I'm not sure how this solves anything (whether rdmd reads from stdout or reads from file its still one compiler invocation).

marler8997:
https://github.com/dlang/tools/pull/271#issuecomment-358536008

ibuclaw:
I'm not sure how introducing a regression warrants a new feature.

marler8997:
Ugh, not going to deal with you again. Closing and washing my hands of this.

ibuclaw:
@marler8997 - Please be positive about this.

Please be aware that things are still up in the air regarding your linked discussion thread. Please understand that any sweeping changes whilst discussions are ongoing are to be considered in haste.

marler8997:
> I'm not sure how this solves anything (whether rdmd reads from stdout or reads from file its still one compiler invocation).

I've edited the description to explain why it's useful to allow the verbose output to go to a file.

WalterBright:
What's wrong with:
```
dmd -verbose 2> file
```
? It's what I do.

marler8997:
> What's wrong with:
```
dmd -verbose 2> file
```

I'm a bit confused; `-verbose` isn't a dmd flag, I assume you meant `-v` correct?  Also, verbose output goes to `stdout` but you've redirected `stderr`.  Did you mean this?
```
dmd -v > file
```
the point of this PR is to allow a tool to invoke `dmd` while keeping error messages and other compiler information going to `stdout` and `stderr` while at the same time also capturing verbose output.  In your example, since OPTLINK prints errors to `stdout`, all error messages from OPTLINK will go into `file` instead of to the console for the user to see.

Another solution to this problem would be to redirect `stdout` as you have in your example and then for the tool to read `file` and try to separate what output came from dmd's verbose information and whatever else happened to go to stdout (such as OPTLINK output), however this solution is more brittle and requires the tool to have alot more logic about the compiler in order to filter the messages properly.

timotheecour:
@WalterBright @marler8997 I'd like this PR regardless of above mentioned OPTLINK bug, but is this related? https://github.com/DigitalMars/optlink/issues/20

andralex:
>OPTLINK prints errors to stdout

This is definitely a matter with optlink. What other output goes to stdout?

marler8997:
Other than some special categories of verbose output (i.e. -vls, -vgc, etc) that's all that I know of. If a user wanted to use rdmd and see output from those arguments then we'd have to address those as well.
It's also possible that other linker's output their error message to stdout as well but I can't say that I know.

andralex:
optlink would need to be fixed to output to stderr, in all likelihood that shouldn't be too hard. @marler8997 can you please file an issue?

There would also be output of `pragma(msg, ...)` in the compiled program that also go to stdout. I wonder whether that creates a potential security risk as pragmas could output messages that confuse rdmd. cc @CyberShadow @MartinNowak @John-Colvin If that's the case, that's solid motivation for this PR.

marler8997:
why would you think it's a security risk?

marler8997:
`rdmd` ignores any lines that don't match the regex it is expecting so it's unlikely to confuse rdmd

andralex:
I'm not an expert but I figure pragmas could instruct dmd to load dependencies from paths and modules that aren't supposed to be looked at.

marler8997:
Interesting.  Yes you could most certainly use pragmas to cause dmd to load other files (at least with the current design, not with the new `-i` interface).  I didn't think of that....

WalterBright:
You can also use https://dlang.org/dmd-windows.html#switch-Xf 

The JSON output is supposed to be so other tools can get at the information.

marler8997:
I'm not that familiar with what the JSON file contains, are you saying the you think rdmd could use the JSON file to get all the information it needs?  It needs to know all the modules that were imported, the compiler config file, and all the library files that are included in the compilation.

WalterBright:
> I'm not that familiar with what the JSON file contains, are you saying the you think rdmd could use the JSON file to get all the information it needs?

I suggest generating the file, and see if it fills the need or falls short. If the latter, it can be improved.

marler8997:
Will do

wilzbach:
There is also a dedicated JSON output in the works, because dub has similar problems and already generates such a probing file:

https://github.com/dlang/dmd/pull/7521

timotheecour:
* @WalterBright sorry but the json output is not relevant at all for rdmd. It doesn't contain imports. It also contains a lot of irrelevant info (eg serialized function decls), so even if we added the imports (a much bigger PR than this PR), it could make things slower than to just used -v=file. Also it wouldn't solve backward compatibility problem if it expected json output to have this added field
* ldmd2|grep json is empty, so not a good solution

* pragma(msg,..) goes to stderr BUT __ctfeWrite (whenever gets implemented) could write to stdout, and also other fields may print to stdout (and maybe also printf while hacking/debugging the compiler), and also if/when we use the compiler as a library

* This PR is so simple and makes sense, it's inline with other things like -dep/-deps=file

Let's get it in please, I've been wanting it for a while.

ibuclaw:
> All in all, -v=file would avoid all these corner cases AND make tooling based on parsing -v robust without having to know such intricacies.

Different compiler implementations send output to different locations also, which was the entire point of having `global.stdmsg`.  i.e: gdc sends all verbose messages to stderr because stdout is used for `-pipe`.

marler8997:
Per Walter's suggestion, I've started adding code to extend the existing JSON format to contain the relevant information needed by rdmd.  It's not done yet but hopefully will be soon. (https://github.com/dlang/dmd/pull/7757).

WalterBright:
> It doesn't contain imports. It also contains a lot of irrelevant info for rdmd (eg serialized function decls), so even if we added the imports (a much bigger PR than this PR), it could make things slower than to just used -v=file.

The converse argument is providing information to tools via json:

1. Json pretty much removes the requirement to define yet another file format, of which dmd already has several. Defining file formats is seductively appealing, but they always wind up as a time consuming quagmire. Not the least of which is they tend to get hastily and sloppily designed, much to the frustration of the poor slob coming along later who has to add capability to it while being backwards compatible. The poster boy for that is the awful dmd.conf file, for which I bear full responsibility for its slapdash and incompetent design, and being a compiler guy I have no excuse.
2. Json makes it trivial for a tool to read and decode the file, in a correct and robust manner. Phobos has a json parser. Let's leverage it instead of every tool developing their own (inevitably slapdash & incompetent) parser.

I agree that speed may be an issue, as the json file can be large with all the other stuff in it. But I would support a switch to pare that down for specific purposes as needed, but it still should be a json file.

timotheecour:
adding the missing info to json and adding a way to control what json outputs is indeed the best solution. We probably need just 1 json output though (-Xf, see https://github.com/dlang/dmd/pull/7757) instead of 2 (-probe, see https://github.com/dlang/dmd/pull/7521)

marler8997:
I did want to mention that there is one drawback to the JSON format.  It's not a weakness of JSON itself, more that the current JSON format has an organized structure.  This organized structure has a performance cost to the compiler.  With the verbose output, you can print any information you like at any time and then throw away that information away if you no longer need it.  With the JSON format, you need to save the data until the JSON is generated whether or not you need it for anything else.

In the case of the new data I've added for RDMD (https://github.com/dlang/dmd/pull/7757/files), this means that we now have to save a list of all the files that were imported using the "import expression".  Before this, that information was "thrown away" after the file was imported, however, now it is saved to an array of "contentImports" in a field added to the Module class, so that the JSON generation code will have that information.  In this particular case it's not a big deal since import expressions are so rare, but it is a limitation to consider when deciding whether we want rdmd to use the JSON format instead of the current "verbose" output.

marler8997:
Going to close.  In my opinion, allowing verbose output to be directed to a file *might* be a nice feature even if it's not necessary for `rdmd`, but I don't really have any use cases to determine this.  Also since this PR has no activity it's safe to assume no one else see's good justification.  If someone does come up with a justification we can revisit.

marler8997:
?

ibuclaw:
I'll just put this in the phantom zone.

PR #7744, opened 2018-01-18
Automatically convert TOKXXXX to the TOK enum

wilzbach:
I found one more.

```bash
replacements=(
"TOKreserved reserved"
"TOKlparen leftParentheses"
"TOKrparen rightParentheses"
"TOKlbracket leftBracket"
"TOKrbracket rightBracket"
"TOKlcurly leftCurly"
"TOKrcurly rightCurly"
"TOKcolon colon"
"TOKneg negate"
"TOKsemicolon semicolon"
"TOKdotdotdot dotDotDot"
"TOKeof endOfFile"
"TOKcast cast_"
"TOKnull null_"
"TOKassert assert_"
"TOKtrue true_"
"TOKfalse false_"
"TOKcall call"
"TOKaddress address"
"TOKdelete delete_"
"TOKstar star"
"TOKsymoff symbolOffset"
"TOKvar variable"
"TOKdotvar dotVariable"
"TOKdotid dotIdentifier"
"TOKdotti dotTemplateInstance"
"TOKdottype dotType"
"TOKslice slice"
"TOKarraylength arrayLength"
"TOKversion version_"
"TOKdollar dollar"
"TOKtemplate template_"
"TOKdottd dotTemplateDeclaration"
"TOKdeclaration declaration"
"TOKpragma pragma_"
"TOKdsymbol dSymbol"
"TOKuadd uadd"
"TOKremove remove"
"TOKnewanonclass newAnonymousClass"
"TOKcomment comment"
"TOKarrayliteral arrayLiteral"
"TOKassocarrayliteral assocArrayLiteral"
"TOKstructliteral structLiteral"
"TOKclassreference classReference"
"TOKthrownexception thrownException"
"TOKdelegateptr delegatePointer"
"TOKdelegatefuncptr delegateFunctionPointer"
"TOKlt lessThan"
"TOKgt greaterThan"
"TOKge greaterOrEqual"
"TOKequal equal"
"TOKnotequal notEqual"
"TOKidentity identity"
"TOKnotidentity notIdentity"
"TOKindex index"
"TOKis is_"
"TOKunord unord"
"TOKlg lg"
"TOKleg leg"
"TOKule ule"
"TOKul ul"
"TOKuge uge"
"TOKug ug"
"TOKue ue"
"TOKshlass leftShiftAssign"
"TOKshrass rightShiftAssign"
"TOKushrass unsignedRightShiftAssign"
"TOKcatass concatenateAssign"
"TOKcatelemass concatenateElemAssign"
[...]

PR #7747, opened 2018-01-18
Convert TXXXX to Type.Kind

wilzbach:
```bash
replacements=(
"Tarray array"
"Tsarray staticArray"
"Taarray associativeArray"
"Tpointer pointer"
"Treference reference"
"Tfunction function_"
"Tident identifier"
"Tclass class_"
"Tstruct struct_"
"Tenum enum_"
"Tdelegate delegate_"
"Tnone none"
"Tvoid void_"
"Tint8 int8"
"Tuns8 uint8"
"Tint16 int16"
"Tuns16 uint16"
"Tint32 int32"
"Tuns32 uint32"
"Tint64 int64"
"Tuns64 uint64"
"Tfloat32 float32"
"Tfloat64 float64"
"Tfloat80 float80"
"Timaginary32 imaginary32"
"Timaginary64 imaginary64"
"Timaginary80 imaginary80"
"Tcomplex32 complex32"
"Tcomplex64 complex64"
"Tcomplex80 complex80"
"Tbool bool_"
"Tchar char_"
"Twchar wchar_"
"Tdchar dchar_"
"Terror error"
"Tinstance instance"
"Ttypeof typeof_"
"Ttuple tuple"
"Tslice slice"
"Treturn return_"
"Tnull null_"
"Tvector vector"
"Tint128 int128"
"Tuns128 uint128"
"TMAX max_"
)
shopt -s globstar
    w=($r)
    sed -E "s/([^a-zA-Z])${w[0]}/\1Type.Kind.${w[1]}/g" -i **/*.d
done
```

Alternative to https://github.com/dlang/dmd/pull/7742

### DIFF
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -135,105 +135,6 @@ struct ASTBase
     extern (C++) __gshared const(StorageClass) STCStorageClass =
         (STC.auto_ | STC.scope_ | STC.static_ | STC.extern_ | STC.const_ | STC.final_ | STC.abstract_ | STC.synchronized_ | STC.deprecated_ | STC.override_ | STC.lazy_ | STC.alias_ | STC.out_ | STC.in_ | STC.manifest | STC.immutable_ | STC.shared_ | STC.wild | STC.nothrow_ | STC.nogc | STC.pure_ | STC.ref_ | STC.return_ | STC.tls | STC.gshared | STC.property | STC.safe | STC.trusted | STC.system | STC.disable);
 
-    enum ENUMTY : int
-    {
-        Tarray,     // slice array, aka T[]
-        Tsarray,    // static array, aka T[dimension]
-        Taarray,    // associative array, aka T[type]
-        Tpointer,
-        Treference,
-        Tfunction,
-        Tident,
-        Tclass,
-        Tstruct,
-        Tenum,
-
-        Tdelegate,
-        Tnone,
-        Tvoid,
-        Tint8,
-        Tuns8,
-        Tint16,
-        Tuns16,
-        Tint32,
-        Tuns32,
-        Tint64,
-
-        Tuns64,
-        Tfloat32,
-        Tfloat64,
-        Tfloat80,
-        Timaginary32,
-        Timaginary64,
-        Timaginary80,
-        Tcomplex32,
-        Tcomplex64,
-        Tcomplex80,
-
-        Tbool,
-        Tchar,
-        Twchar,
-        Tdchar,
-        Terror,
-        Tinstance,
-        Ttypeof,
-        Ttuple,
-        Tslice,
-        Treturn,
-
-        Tnull,
-        Tvector,
-        Tint128,
-        Tuns128,
-        TMAX
-    }
-
-    alias Tarray = ENUMTY.Tarray;
-    alias Tsarray = ENUMTY.Tsarray;
-    alias Taarray = ENUMTY.Taarray;
-    alias Tpointer = ENUMTY.Tpointer;
-    alias Treference = ENUMTY.Treference;
-    alias Tfunction = ENUMTY.Tfunction;
-    alias Tident = ENUMTY.Tident;
-    alias Tclass = ENUMTY.Tclass;
-    alias Tstruct = ENUMTY.Tstruct;
-    alias Tenum = ENUMTY.Tenum;
-    alias Tdelegate = ENUMTY.Tdelegate;
-    alias Tnone = ENUMTY.Tnone;
-    alias Tvoid = ENUMTY.Tvoid;
-    alias Tint8 = ENUMTY.Tint8;
-    alias Tuns8 = ENUMTY.Tuns8;
-    alias Tint16 = ENUMTY.Tint16;
-    alias Tuns16 = ENUMTY.Tuns16;
-    alias Tint32 = ENUMTY.Tint32;
-    alias Tuns32 = ENUMTY.Tuns32;
-    alias Tint64 = ENUMTY.Tint64;
-    alias Tuns64 = ENUMTY.Tuns64;
-    alias Tfloat32 = ENUMTY.Tfloat32;
-    alias Tfloat64 = ENUMTY.Tfloat64;
-    alias Tfloat80 = ENUMTY.Tfloat80;
-    alias Timaginary32 = ENUMTY.Timaginary32;
-    alias Timaginary64 = ENUMTY.Timaginary64;
-    alias Timaginary80 = ENUMTY.Timaginary80;
-    alias Tcomplex32 = ENUMTY.Tcomplex32;
-    alias Tcomplex64 = ENUMTY.Tcomplex64;
-    alias Tcomplex80 = ENUMTY.Tcomplex80;
-    alias Tbool = ENUMTY.Tbool;
-    alias Tchar = ENUMTY.Tchar;
-    alias Twchar = ENUMTY.Twchar;
-    alias Tdchar = ENUMTY.Tdchar;
-    alias Terror = ENUMTY.Terror;
-    alias Tinstance = ENUMTY.Tinstance;
-    alias Ttypeof = ENUMTY.Ttypeof;
-    alias Ttuple = ENUMTY.Ttuple;
-    alias Tslice = ENUMTY.Tslice;
-    alias Treturn = ENUMTY.Treturn;
-    alias Tnull = ENUMTY.Tnull;
-    alias Tvector = ENUMTY.Tvector;
-    alias Tint128 = ENUMTY.Tint128;
-    alias Tuns128 = ENUMTY.Tuns128;
-    alias TMAX = ENUMTY.TMAX;
-
     alias TY = ubyte;
 
     enum TFlags
@@ -1778,7 +1679,7 @@ struct ASTBase
                 Parameter p = (*parameters)[i];
                 Type t = p.type.toBasetype();
 
-                if (t.ty == Ttuple)
+                if (t.ty == Type.Kind.tuple)
                 {
                     TypeTuple tu = cast(TypeTuple)t;
                     result = _foreach(tu.arguments, dg, &n);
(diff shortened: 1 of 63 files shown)


src/dmd/mtype.d
```diff
@@ -481,7 +380,60 @@ alias MOD = ubyte;
...
+        typeof_,
+        tuple,
+        slice,
+        return_,
+
+        null_,
+        vector,
+        int128,
+        uns128,
+        MAX,
```
JinShil: I suggest `max_`

src/dmd/mtype.d
```diff
@@ -481,7 +380,60 @@ alias MOD = ubyte;
  */
 extern (C++) abstract class Type : RootObject
 {
-    TY ty;
+    enum Kind : ubyte
+    {
+        array,     // slice array, aka T[]
+        sarray,    // static array, aka T[dimension]
+        aarray,    // associative array, aka T[type]
```
JinShil: If it needs a comment, it's probably a bad name.  `staticArray` and `associativeArray` don't seem all that bad.

src/dmd/mtype.d
```diff
@@ -481,7 +380,60 @@ alias MOD = ubyte;
...
+        ident,
+        class_,
+        struct_,
+        enum_,
+
+        delegate_,
+        none,
+        void_,
+        int8,
+        uns8,
```
JinShil: Can't we use `uint8` here and for others without causing any trouble?

JinShil (changes requested):

src/dmd/astbase.d
```diff
@@ -2526,12 +2425,65 @@ struct ASTBase
...
+            struct_,
+            enum_,
+
+            delegate_,
+            none,
+            void_,
+            int8,
+            uint8,
+            int16,
+            uns16,
```
JinShil: `uint16`

JinShil (changes requested):

src/dmd/mtype.d
```diff
@@ -481,7 +380,60 @@ alias MOD = ubyte;
...
-    TY ty;
+    enum Kind : ubyte
+    {
+        array,     // slice array, aka T[]
+        staticArray,      // static array, aka T[dimension]
+        associativeArray, // associative array, aka T[type]
+        pointer,
+        reference,
+        function_,
+        ident,
```
jacob-carlborg: This is shortened as well, but I don't mind it.

JinShil: Let's get'r done.  I think that should be `identifier`.  The fact that I'm not sure is justification.

jacob-carlborg: `identifier` is shortened to `ident` in other places in the code.

wilzbach: Done.

JinShil (approved):

JinShil:
Rebased

jacob-carlborg (approved):

JinShil:
@wilzbach is going to be away for few days, so I want to try to get this in because it is holding up some other work I'm trying to do.  That is, it's such a large refactor it will clobber any other PRs, so I want to get this in first so we don't have rebase so many other PRs.

#7744 really clobbered this one.  There were too many conflicts, so I just reapplied @wilzbach's script.  It could probably use another review.  (Sorry @wilzbach, I know I didn't do the git stuff right)

cc @jacob-carlborg

jacob-carlborg (approved):

WalterBright:
This makes things much more verbose. Please no.

WalterBright:
@jacob-carlborg Please hold off on rebasing, as I have issued a revert PR on this.

andralex:
What is the motivation behind putting the enum behind two types? It looks like a heavy uptick in verbosity for a modest improvement in readability.

JinShil:
it avoids a name collision between `class Type` and `enum Type`.  we could have used something like  `TYPE` for the enum, but we were attempting to follow the [style guide](https://dlang.org/dstyle.html).  The verbosity can be handle easily using D's other fantastic features such as 

```D
with(Type.Kind) 
{  
    auto t = uint8; 
}
```
or
```D
alias T = Type.Kind;
auto t = T.uint8;
```

andralex:
@JinShil it really seems to me going overboard with engineering. I don't see a reason to touch 63 files for this rename.

JinShil:
Please see the larger picture of what's trying to be accomplished here:  https://github.com/dlang/dmd/projects/4.

andralex:
@JinShil very nice but it should be done with sense and sensibility.

JinShil:
> but it should be done with sense and sensibility.

@andralex It was.

andralex:
@JinShil sorry if it seems we have different sensibilities. I'm sure something can be worked out.

JinShil:
@andralex If verbosity is primary concern.   How about if I submit a PR that reduces the verbosity with an alias in the modules where it is used (e.g. `alias T = Type.Kind`)?  Then it would be `T.uint8` instead of `Type.Kind.uint8`.

JinShil:
The other possibility would be to establish a different style guide for DMD that use all-cap enums (e.g. `enum TYPE`) to distinguish from other types (e.g. `class Type`).  Though, IMO, with DMD now being a library in the ecosystem, I think it's a good idea for the organization to follow the same style guide for all user-facing APIs.

Keep in mind also that DMD is old code that has gone through many transitions, and it shows.  We're trying to also remove some of the cruft.

andralex:
@JinShil are you on slack so we can talk?

andralex:
This diff touches about 1% of the entire codebase, which is surprisingly high for a large project. Such heavy changes should be well motivated, approached with caution, and leave a palpable improvement behind. Walter's and my take on this:

* Adds some type safety - enums are better than ints.
* Safety added is not demonstrated - per @WalterBright there has never been a bug caused by this.
* Touches very much code, leading to names that are longer and more cumbersome.
* The `_` warts don't help either.

We'd like to take advantage of the better typing of enums, but also avoid the massive surgery, eyesore, and carpal-tunnel-syndrome-inducing verboseness. How about this: create an enum and also define aliases for its values? That way most of the code is unaffected.

ibuclaw:
Note, D style enums makes for better automatic translation of D -> C++ with the current dtoh tool I have,  particularly as enums are unscoped in C++98.  However whether or not I'm doing things in the right way could be up for debate also. :-)

JinShil:
@andralex If that's how you feel about this one, you're really going to hate #7744 and the 17 other already merged PRs [here](https://github.com/dlang/dmd/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+author%3Awilzbach+convert)

andralex:
Many of those are actually okay. What we need to develop is a common sense of values, so each of the core contributors can trust one another for making changes they themselves would make. It's easy to take a nice simple goal "use D enums instead of C++ enums" and exaggerate in applying it. 

(A meta-comment - the kind of refactoring that can be easily automated is also likely to have low yield. The real leverage is in intelligent refactorings that use the D language features creatively to improve code quality. We have many clowny things in our current codebase, nobody would deny that, but enums are probably not near the top of the list.)

JinShil:
Since @wilzbach was the primary champion behind this, I'm not going to do anything at the moment until he gets back in a few days, so we can all get on the same page.

wilzbach:
> there has never been a bug caused by this.

Just something I found by _accident_:

https://github.com/dlang/dmd/pull/7735

`Final switch` is great because it means that we can't forget to handle cases and the compiler will automatically tell you which places need to be updated when you add a new option.

> DMD now being a library in the ecosystem,

Yes, for example, no one reviewing this PR knew why it is Tuns8 and not Tuint8. People could already use DMD as a library today - the entire frontend is exposed via dub after all, but as we know they don't. My bet is that they don't like dealing with this codebase when for most things libdparse works too and provides readable APIs that are easy to use.

Also as no one has mentioned this so far, you get nice auto-completion when using an enum.

> D style enums makes for better automatic translation of D -> C++ with the current dtoh tool I have

Isn't this a reason enough?

> How about this: create an enum and also define aliases for its values? That way most of the code is unaffected.

That's the status quo. It doesn't help Iain with the automatic header generation, is still confusing for newcomers and FWIW I haven't seen many `final switch`es being used before I started this series.

> also likely to have low yield. 
> ...
> exaggerate in applying it.

Well once two or three of these conversion PRs were merged we realized that we have to rebase the active PRs anyways (that's also due to the backtick conversion of the error messages which happened at the same time), so we continued to make all changes once and for all in this window.
The other motivation was that once people start using DMD as a library, we can't make such changes easily anymore.

> until he gets back in a few days, so we can all get on the same page.

Yeah bad timing for a weekend trip, but even with bad internet I still can't stop reading my mails on my phone :innocent:

Anyhow how do we go from here?

> If verbosity is primary concern, how about if I submit a PR that reduces the verbosity with an alias in the modules where it is used (e.g. alias T = Type.Kind)? 

What happened to this? I didn't see a reply.

jacob-carlborg:
@WalterBright I have not rebased any PRs, not sure what you're talking about.

jacob-carlborg:
I think it's a big disappointment that this was reverted. If we want to attract more contributors we need to be able to adopt and change. We cannot have the whole code base written and organized in a way that is only suited for one specific person. We should always maximize for clarity to make it as easy as possible to understand the code.

Talking about verbosity. I don't think this change make the code more verbose, it makes it easier to read. One don't have to guess the meaning of the names. On the other what I think *is* verbose is to have a single file with 10k+ lines code in it [1].

[1] https://github.com/dlang/dmd/blob/master/src/dmd/expressionsem.d#L10056

andralex:
>Anyhow how do we go from here?

I mentioned this in the revert PR I think - a simple way to go about this is to define the enum and then alias the olden symbols to it. I think what really made this worse is the doubly-nested names and the many underscore warts. (The compiler is in the unique position of dealing with its own keywords, which makes this a more frequent issue than otherwise.) Looking at the code before and after just doesn't make @WalterBright or me go, "yes this is better".

So: please define a non-nested enum e.g. `Tp.xxx` and then alias the existing `Txxx` to its values. Thanks!

>I think it's a big disappointment that this was reverted. If we want to attract more contributors we need to be able to adopt and change. 

The underlying assumption is you believe most others have your sensibilities and not ours. That may as well be true but it's an assumption.

As I said elsewhere: there are many many MANY clowny things in our codebase - phobos, druntime, and dmd. Of these, it seems odd that we choose some of the least hurting pieces to debate about.

JinShil:
@andralex If you want to put an end to this discussion quickly and move on to other things, I'll go along with your suggestion above, and move on to other things myself.

However, if you'd like to take this opportune time to establish a common set of values, then I suggest we take this discussion a little further.  Your call.

If you choose to have the discussion, I suggest, before offering or agreeing to suggestions, we should try to establish a few principles, starting with this:

Should new or significantly modified code attempt to follow the [Style Guide](https://dlang.org/dstyle.html), especially with regard to naming conventions in the context of this discussion?

jacob-carlborg:
> So: please define a non-nested enum e.g. Tp.xxx and then alias the existing Txxx to its values. Thanks!

If the enum is only called `Tp` and in addition to that we have aliases as well, then the aliases would only be two characters shorter compared to referring to the enum member through the enum.

PR #7750, opened 2018-01-19
Use more final switches

wilzbach:
Now that these types are enums, we can make the ignored cases in a `switch` explicit.

### DIFF
--- a/src/dmd/hdrgen.d
+++ b/src/dmd/hdrgen.d
@@ -1227,7 +1227,7 @@ public:
     override void visit(LinkDeclaration d)
     {
         const(char)* p;
-        switch (d.linkage)
+        final switch (d.linkage)
         {
         case LINK.d:
             p = "D";
@@ -1247,7 +1247,8 @@ public:
         case LINK.objc:
             p = "Objective-C";
             break;
-        default:
+        case LINK.default_:
+        case LINK.system:
             assert(0);
         }
         buf.writestring("extern (");
@@ -3279,7 +3280,7 @@ extern (C++) void trustToBuffer(OutBuffer* buf, TRUST trust)
 
 extern (C++) const(char)* trustToChars(TRUST trust)
 {
-    switch (trust)
+    final switch (trust)
     {
     case TRUST.default_:
         return null;
@@ -3289,8 +3290,6 @@ extern (C++) const(char)* trustToChars(TRUST trust)
         return "@trusted";
     case TRUST.safe:
         return "@safe";
-    default:
-        assert(0);
     }
 }
 
--- a/src/dmd/json.d
+++ b/src/dmd/json.d
@@ -246,7 +246,7 @@ public:
 
     void property(const(char)* name, TRUST trust)
     {
-        switch (trust)
+        final switch (trust)
         {
         case TRUST.default_:
             // Should not be printed
@@ -261,14 +261,12 @@ public:
         case TRUST.safe:
             property(name, "safe");
             break;
-        default:
-            assert(false);
         }
     }
 
     void property(const(char)* name, PURE purity)
     {
-        switch (purity)
+        final switch (purity)
         {
         case PURE.impure:
             // Should not be printed
@@ -286,8 +284,6 @@ public:
         case PURE.fwdref:
             property(name, "fwdref");
             break;
-        default:
-            assert(false);
         }
     }
 


src/dmd/cppmanglewin.d
```diff
@@ -500,30 +500,38 @@ private:
...
                     break;
                 case Prot.Kind.protected_:
                     buf.writeByte('M');
                     break;
-                default:
+                case Prot.Kind.undefined:
+                case Prot.Kind.none:
+                case Prot.Kind.package_:
+                case Prot.Kind.public_:
+                case Prot.Kind.export_:
```
wilzbach: Those look like uncovered cases. Anyone familiar with `cppmanglewin`?

JinShil (dismissed):

jacob-carlborg (approved):

src/dmd/cppmanglewin.d
ibuclaw: Nope, but this should have been a translation of the dmd backend mangler. I wouldn't expect C++ to mangle any of those anyway

src/dmd/hdrgen.d
```diff
@@ -1227,8 +1227,9 @@ public:
     override void visit(LinkDeclaration d)
     {
         const(char)* p;
-        switch (d.linkage)
+        final switch (d.linkage)
         {
+        case LINK.default_:
```
ibuclaw: Where does this come from?

ibuclaw: Looks like case is in wrong place

ibuclaw (changes requested):

src/dmd/cppmanglewin.d
WalterBright: The coverage analyzer only runs on Linux, so Windows code won't get executed.

WalterBright: Note that the mangler is performance critical. Adding all these extra cases that do nothing just slows it down.

src/dmd/astbase.d
```diff
@@ -4073,7 +4073,11 @@ struct ASTBase
                         e = new ArrayExp(loc, e, cast(Expression)id);
                         break;
 
-                    default:
+                    case DYNCAST.object:
+                    case DYNCAST.tuple:
+                    case DYNCAST.parameter:
+                    case DYNCAST.statement:
+                    case DYNCAST.condition:
```
WalterBright: I don't see the value in adding these cases just to `assert(0)`. The default does that just fine.

src/dmd/json.d
```diff
@@ -261,14 +261,12 @@ public:
...
             break;
-        default:
-            assert(false);
         }
     }
 
     void property(const(char)* name, PURE purity)
     {
-        switch (purity)
+        final switch (purity)
```
WalterBright: this change looks like an improvement.

src/dmd/mtype.d
```diff
@@ -5594,7 +5594,7 @@ extern (C++) final class TypeFunction : TypeNext
                 purity = PURE.weak;
                 break;
             }
-            switch (purityOfType((fparam.storageClass & STC.ref_) != 0, t))
+            final switch (purityOfType((fparam.storageClass & STC.ref_) != 0, t))
```
WalterBright: not this one.

src/dmd/json.d
```diff
@@ -246,7 +246,7 @@ public:
 
     void property(const(char)* name, TRUST trust)
     {
-        switch (trust)
+        final switch (trust)
```
WalterBright: looks good

src/dmd/hdrgen.d
```diff
@@ -3279,7 +3280,7 @@ extern (C++) void trustToBuffer(OutBuffer* buf, TRUST trust)
 
 extern (C++) const(char)* trustToChars(TRUST trust)
 {
-    switch (trust)
+    final switch (trust)
```
WalterBright: looks good

src/dmd/hdrgen.d
```diff
@@ -1227,8 +1227,9 @@ public:
     override void visit(LinkDeclaration d)
     {
         const(char)* p;
-        switch (d.linkage)
+        final switch (d.linkage)
```
WalterBright: looks good

src/dmd/dmangle.d
```diff
@@ -391,15 +391,16 @@ public:
                 buf.writestring("Nj");
             if (ta.isscope && !ta.isreturn && !ta.isscopeinferred)
                 buf.writestring("Nl");
-            switch (ta.trust)
+            final switch (ta.trust)
```
WalterBright: not worth it

src/dmd/cppmanglewin.d
```diff
@@ -588,15 +600,19 @@ private:
         }
         else
         {
-            switch (d.protection.kind)
+            final switch (d.protection.kind)
```
WalterBright: not worth it

src/dmd/cppmanglewin.d
```diff
@@ -542,15 +550,19 @@ private:
         else if (d.isMember2()) // static function
         {
             // <flags> ::= <virtual/protection flag> <calling convention flag>
-            switch (d.protection.kind)
+            final switch (d.protection.kind)
```
WalterBright: not worth it

src/dmd/cppmanglewin.d
```diff
@@ -500,30 +500,38 @@ private:
...
+                case Prot.Kind.public_:
+                case Prot.Kind.export_:
                     buf.writeByte('U');
                     break;
                 }
             }
             else
             {
-                switch (d.protection.kind)
+                final switch (d.protection.kind)
```
WalterBright: not worth it

src/dmd/hdrgen.d
wilzbach: https://github.com/dlang/dmd/blob/1018db023831a8c348bb8f9121b08fabb6e32fc2/src/dmd/globals.d#L418-L428

Isn't `Link.default` equal to `LINK.d`?

wilzbach:
(Removed the final switches for switches where cases are missing)

JinShil (approved):

src/dmd/hdrgen.d
ibuclaw: It should be considered as a no linkage, or unset.

wilzbach: Imho the best solution would be to avoid having `LINK.default_` and `LINK.system` in the first place, but that's something for another PR..
-> I moved it to the position of the previous `default` case to avoid the PR being stalled. In other words: it doesn't change the status quo.

PR #7751, opened 2018-01-19
Experiment with running Code Coverage on AppVeyor

wilzbach:
This will probably increase the build time heavily, but let's see first
how much more Windows would cover.

### DIFF
--- a/appveyor.sh
+++ b/appveyor.sh
@@ -66,13 +66,20 @@ done
 
 # build via makefile
 cd /c/projects/dmd/src
-make -f win64.mak reldmd DMD=../src/dmd
+make -f win64.mak reldmd
 
 cd /c/projects/druntime
-make -f win64.mak DMD=../dmd/src/dmd
+make -f win64.mak
 
 cd /c/projects/phobos
-make -f win64.mak DMD=../dmd/src/dmd
+make -f win64.mak
+
+cd /c/projects/dmd/
+mv generated _generated
+cd src
+# Inject -cov into the Makefile
+sed -i 's/DOPT=-O -release -inline"/DOPT=-O -release -inline -cov"/' -i win32.mak
+make -f win64.mak reldmd DMD=../_generated/windows/release/64/dmd.exe
 
 cp /c/projects/phobos/phobos64.lib /c/projects/dmd/
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -28,6 +28,9 @@ artifacts:
 init:
   - git config --global core.autocrlf input
 
+before_build:
+  - choco install codecov
+
 build_script:
   - cd c:/projects/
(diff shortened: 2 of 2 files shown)


RazvanN7:
What's the status on this?

wilzbach:
I haven't had time to look into why the coverage files don't get generated and as I don't have a Windows machine it's not at the top of my priority list. Sorry.

wilzbach:
> Even if it's not, would it be problematic to close this and reopen once you work on it?

I don't mind per se, but once this is closed it will probably be forgotten and sink in the abyss.

> At this point, this PR is just eating autotester time.

That's not true. The auto-tester uses a priority list based on the latest push date of a PR which means old PRs are never ever tested ;-)

wilzbach:
BTW for reference, the status of this PR is the following, `-cov` gets injected correctly into DMD

```
dmd  -of..\generated\windows\release\64\dmd.exe -vtls -J..\generated\windows\release\64 -J../res -L/STACK:8388608 -O -release -inline -cov
```

But for some reason:

```
2018-01-23 01:19:31 [Fatal] No Report detected.
   at Codecov.Coverage.Tool.Coverage.LoadReportFile() in C:\Users\Administrator\github\codecov-exe\Source\Codecov\Coverage\Tool\Coverage.cs:line 29
```

Geod24:
Where does Codecov expect the coverage reports ?

wilzbach:
The good news. The coverage files are generated and detected by CodeCov:

image

The bad news:
-------------------

1) Built time increased from to ~28-32min to ~40 min

2) CodeCov doesn't seem to merge the reports yet

image

wilzbach:
This looks a lot better now:

image

https://codecov.io/gh/dlang/dmd/compare/504767e31b2be62f324daf4b973e89ee1dd53a16...c0c4dd17b27a617f7d0a8ee536d7f191837148e8

The runtime of the last build was 44 minutes. So we are looking at ~15 minutes increase in build time of the AppVeyor machine (the maximum build time is 60 minutes).
What do you think: is it worth it?

Geod24:
> What do you think: is it worth it?

Yes. Will merge in 72h unless someone object.

rainers:
AppVeyor only covers COFF so far, shouldn't there also be some coverage for OMF?

The phobos unittests are also not run, not sure if they take part in coverage elsewhere.

rainers:
>  I don't see `'/delexe/la'` anywhere.

It's also shown for the regular build: https://ci.appveyor.com/project/greenify/dmd/builds/20315804#L307 These are options for optlink that mslink ignores.

I think this is failing because some files (ROOT_SRCS) are compiled both into lexer.lib and dmd.exe. This happens to work for the normal build, but coverage creates other dependencies so the linker drags in both versions.

Geod24:
> Looks like this didn't happen.

Indeed. Want to rebase & ping me ?

wilzbach:
Well, we don't use AppVeyor anymore :man_shrugging:

PR #7752, opened 2018-01-19
Remove unused FLAGS aliases

wilzbach:
They are only used once below. We can even use `with` to make it look nicer.

### DIFF
--- a/src/dmd/lexer.d
+++ b/src/dmd/lexer.d
@@ -1942,18 +1942,13 @@ class Lexer
         }
         enum FLAGS : int
         {
-            FLAGS_none = 0,
-            FLAGS_decimal = 1, // decimal
-            FLAGS_unsigned = 2, // u or U suffix
-            FLAGS_long = 4, // L suffix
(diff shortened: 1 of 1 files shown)


JinShil (approved):

jacob-carlborg (approved):

src/dmd/lexer.d
```diff
@@ -1988,9 +1983,10 @@ class Lexer
         if (base == 8 && n >= 8)
             error("octal literals `0%llo%.*s` are no longer supported, use `std.conv.octal!%llo%.*s` instead", n, p - psuffix, psuffix, n, p - psuffix, psuffix);
         TOK result;
+        with(FLAGS)
```
WalterBright: I'd rather not start using `with`.

WalterBright (dismissed):
No `with`

wilzbach:
(removed `with)

wilzbach:
Argh Jenkins seems to be losing PRs more and more frequently :/

PR #7753, opened 2018-01-19
Rename <pkg> to <pattern> in dmd usage.

marler8997:
I've landed on the term "module pattern" for each item in the list of strings provided for option `-i`.  Given that, the abbreviation `<ptrn>` seems to make more sense than `<pkg>`.


jacob-carlborg:
Why use an abbreviation at all, why not just say `pattern` or `module pattern`?

marler8997:
ok

jacob-carlborg (approved):

JinShil (approved):

PR #7754, opened 2018-01-19
some stylistic improvements to toTypeFunction()

WalterBright:
https://github.com/dlang/dmd/pull/7749 got merged before I reviewed it, so my review is coming in the form of a touch-up PR.

1. Please don't invent a new formatting style for function Ddoc comments.

2. I find default arguments make code harder to understand. Besides, this is an almost ideal use case for an `out` parameter. `out` parameters eliminate the null pointer issue, are more checkably memory safe, and the code is simpler.

3. Reorganized the code a bit to eliminate variable `tf` and reduce the scope of `s`.

cc @UplinkCoder @JinShil

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -445,49 +445,46 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         return new IntegerExp(e.loc, false, Type.tbool);
     }
 
-    /**
-       Gets the function type from a given AST node
-       if the node is a function of some sort.
-
-     Params:
-        o = an AST node to check for a `TypeFunction`
-        fdp = optional pointer to a function declararion, to be set
-          if `o` is a function declarartion.
-
-     Returns:
-        a type node if `o` is a declaration of
-            a delegate, function, function-pointer
-          or a variable of the former.  Otherwise, `null`.
-    */
-    static TypeFunction toTypeFunction(RootObject o, FuncDeclaration* fdp = null)
+    /********
+     * Gets the function type from a given AST node
+     * if the node is a function of some sort.
+     * Params:
+     *   o = an AST node to check for a `TypeFunction`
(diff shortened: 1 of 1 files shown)


src/dmd/traits.d
```diff
@@ -445,49 +445,46 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
         {
             if (t.ty == Type.Kind.function_)
-                tf = cast(TypeFunction)t;
+                return cast(TypeFunction)t;
             else if (t.ty == Type.Kind.delegate_)
-                tf = cast(TypeFunction)t.nextOf();
+                return cast(TypeFunction)t.nextOf();
             else if (t.ty == Type.Kind.pointer && t.nextOf().ty == Type.Kind.function_)
-                tf = cast(TypeFunction)t.nextOf();
+                return cast(TypeFunction)t.nextOf();
```
andralex: two redundant elses, but I'll let them be

andralex (approved):

JinShil:
@WalterBright, I'm sorry if you felt this was merged prematurely.  But please note that the PR is actually almost 9 months old, and I assumed you were OK with it due to [this comment from you](https://github.com/dlang/dmd/pull/6835#issuecomment-304488000) in the original PR.

> LGTM provided the function is documented

WalterBright:
I reserve the right to learn how to do things better over time :-)

src/dmd/traits.d
WalterBright: You're right, I missed that.

stefan-koch-sociomantic:
The comment was written a way I would write code: with indentation. I don't mind the change too much.
The default parameter was used in order to not force the caller to pass in a variable when it's not needed.
Since it was only needed in one case.

also after your changes toTypeFunction will be unable to go through the inliner; which I also considered when writing it.

PR #7756, opened 2018-01-20
Add asttypename again

UplinkCoder:
This re-adds the debugging facility `asttypename.d`
It does not only provide an easy way to get a node's name;
It does also provide a rudimentary test of the visitor interface.

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -299,7 +299,7 @@ FRONT_SRCS=$(addsuffix .d, $(addprefix $D/,access aggregate aliasthis apply argt
 	json lib libelf libmach link mars mtype nogc nspace objc opover optimize parse permissivevisitor sapply templateparamsem	\
 	sideeffect statement staticassert target typesem traits transitivevisitor parsetimevisitor visitor	\
 	typinf utils scanelf scanmach statement_rewrite_walker statementsem staticcond safe blockexit printast \
-	semantic2 semantic3))
+	semantic2 semantic3 asttypename))
 
 LEXER_SRCS=$(addsuffix .d, $(addprefix $D/, console entity errors globals id identifier lexer tokens utf))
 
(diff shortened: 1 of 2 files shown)


JinShil:
I'd like to see a link to the pull request that initially added this and a link to the pull request that removed it.

JinShil:
It looks like this was removed in https://github.com/dlang/dmd/pull/7397/commits/1ee14af58a4dd0e1ea925699b238dec5ec8ba2b2 due to build failures, and it still fails to build.

@UplinkCoder do you plan on fixing the build failures?

stefan-koch-sociomantic:
@JinShil there failures where already fixed by the guy who introduced them and on newer compilers  this works.

JinShil:
I assume when you say "new compilers" you mean "newer host compilers", correct?

With each PR I encounter, I want to try to move it in one of two directions:  to merge, or to close.  I can't merge this because it fails to build, so I'm left with only the option to close it.  Perhaps you can resubmit it at a later date when the host compilers are upgraded.

stefan-koch-sociomantic:
@JinShil perhaps https://github.com/dlang/dmd/commit/1ee14af58a4dd0e1ea925699b238dec5ec8ba2b2 should be reverted ?
It purposely removed a file which was flagging an error.

wilzbach:
If you want to this to be re-enabled, you need to find a workaround against older host compilers - [it doesn't look like we can bump the minimal version required to build D anytime soon](https://forum.dlang.org/post/zbxaviowooiaenaawmgs@forum.dlang.org).

@RazvanN7 might be able to help

JinShil:
> perhaps 1ee14af should be reverted ?

Then I encourage you to submit a PR to do that.  As for this PR, I can't do anything with it, but close it.

RazvanN7:
The issue with asttype-name was that it needed a newer version of the BOOTSTRAPPING compiler to compile. Since asttype-name isn't needed anywhere in the compiler I suggest it is left outside of the standard build and added whenever somebody wants to do some debugging.

JinShil:
This PR fails to build.  Right or wrong, I can't do anything with it but close it.  Feel free to re-open if you can find a workaround for the existing host compilers, or resubmit when the host compilers are upgraded.

stefan-koch-sociomantic:
But in case I want to use if for debugging I need it to be working.

PR #7757, opened 2018-01-20
Add more stuff to JSON file

marler8997:
This PR is based on recommendation from @WalterBright (https://github.com/dlang/dmd/pull/7746#issuecomment-359153438).  In order for `rdmd` to support single-invocation, we need a way to get the information it currently gets from "verbose output", but without losing error messages and other stdout messages.  Walter recommended rdmd switch to using the JSON file that dmd generates.  Currently the JSON file is missing most of the information that rdmd requires, so this PR is adding that information to it.

Most of the information needed by rdmd is "semantic data" whereas the JSON file seems to contain "syntax data".  Because of this, I've added the "semantic data" to the end of the JSON file in it's own object called `semantics`.  Here's a list of the new additions and the current status:
* non-root imported modules (ADDED)
* file content imports (i.e. `import("somefile")`) (ADDED)

Along with semantic data, there is some build information that rdmd uses.  I've put this in the beginning of the JSON file in an object named "buildInfo". It contains the following:
* target library file if `-lib` is used (ADDED)
* the compiler config file (ADDED)
* the compiler binary file itself (ADDED)

### DIFF
--- a/src/dmd/json.d
+++ b/src/dmd/json.d
@@ -34,6 +34,12 @@ import dmd.mtype;
 import dmd.root.outbuffer;
 import dmd.visitor;
 
+version(Windows) {
+    extern (C) char* getcwd(char* buffer, size_t maxlen);
+} else {
+    import core.sys.posix.unistd : getcwd;
+}
+
 private extern (C++) final class ToJsonVisitor : Visitor
 {
     alias visit = Visitor.visit;
@@ -789,12 +795,70 @@ public:
         jsonProperties(d);
         objectEnd();
     }
+
+    private void generateCompilerInfo()
+    {
+        objectStart();
+        property("kind", "compilerInfo");
+        property("binary", global.params.argv0);
+        property("version", global._version);
+        propertyBool("supportsIncludeImports", true);
+        objectEnd();
+    }
+    private void generateBuildInfo()
+    {
+        objectStart();
+        property("kind", "buildInfo");
+        property("cwd", getcwd(null, 0));
+        property("config", global.inifilename ? global.inifilename : null);
+        if (global.params.lib) {
+            property("library", global.params.libname);
+        }
+        propertyStart("importPaths");
+        arrayStart();
+        foreach (importPath; *global.params.imppath)
+        {
+            item(importPath);
+        }
+        arrayEnd();
+        objectEnd();
+    }
+    private void generateSemantics()
+    {
+        objectStart();
+        property("kind", "semantics");
+        propertyStart("modules");
+        arrayStart();
+        foreach (m; Module.amodules)
+        {
+            objectStart();
+            if(m.md)
+                property("name", m.md.toChars());
+            property("file", m.srcfile.toChars());
+            propertyBool("isRoot", m.isRoot());
+            if(m.contentImportedFiles.dim > 0)
+            {
+                propertyStart("contentImports");
+                arrayStart();
+                foreach (file; m.contentImportedFiles)
+                {
+                    item(file);
+                }
+                arrayEnd();
+            }
+            objectEnd();
+        }
+        arrayEnd();
+        objectEnd();
+    }
 }
 
 extern (C++) void json_generate(OutBuffer* buf, Modules* modules)
 {
     scope ToJsonVisitor json = new ToJsonVisitor(buf);
     json.arrayStart();
+    json.generateCompilerInfo();
+    json.generateBuildInfo();
     for (size_t i = 0; i < modules.dim; i++)
     {
         Module m = (*modules)[i];
@@ -802,6 +866,7 @@ extern (C++) void json_generate(OutBuffer* buf, Modules* modules)
             fprintf(global.stdmsg, "json gen %s\n", m.toChars());
         m.accept(json);
     }
+    json.generateSemantics();
     json.arrayEnd();
     json.removeComma();
 }
--- a/test/sanitize_json.d
+++ b/test/sanitize_json.d
@@ -1,0 +1,151 @@
+import std.exception : assumeUnique;
+import std.string : startsWith, replace;
+import std.format : formattedWrite, format;
+import std.json;
+import std.getopt;
+import std.file : readText;
+import std.stdio;
+
+bool keepDeco = false;
+
+void usage()
+{
+    writeln("Usage: santize_json [--keep-deco] <input-json> [<output-json>]");
+}
+int main(string[] args)
+{
+    getopt(args,
+        "keep-deco", &keepDeco);
+    args = args[1 .. $];
+    if (args.length == 0)
+    {
+        usage();
+        return 1;
+    }
+    string inFilename = args[0];
+    File outFile;
+    if(args.length == 1)
+    {
+        outFile = stdout;
+    }
+    else if(args.length == 2)
+    {
+        outFile = File(args[1], "w");
+    }
+    else
+    {
+        writeln("Error: too many command line arguments");
+        return 1;
+    }
+
+    auto json = parseJSON(readText(inFilename));
+    sanitize(json.array);
+
+    outFile.write(json.toJSON(true));
+    return 0;
+}
+
+void sanitize(JSONValue[] rootArray)
+{
+    foreach (ref obj; rootArray)
+    {
+        auto kind = obj.object["kind"].str;
+        if (kind == "compilerInfo")
+            sanitizeCompilerInfo(obj.object);
+        else if (kind == "buildInfo")
+            sanitizeBuildInfo(obj.object);
+        else if(kind == "module")
+            sanitizeSyntaxNode(obj);
+        else if(kind == "semantics")
+            sanitizeSemantics(obj.object);
+    }
+}
+
+void removeString(JSONValue* value)
+{
+    assert(value.type == JSON_TYPE.STRING);
+    *value = JSONValue("VALUE_REMOVED_FOR_TEST");
+}
+void removeNumber(JSONValue* value)
+{
+    assert(value.type == JSON_TYPE.INTEGER || value.type == JSON_TYPE.UINTEGER);
+    *value = JSONValue(0);
+}
+void removeStringIfExists(JSONValue* value)
+{
+    if (value !is null)
+        removeString(value);
+}
+
+void sanitizeCompilerInfo(ref JSONValue[string] buildInfo)
+{
+    removeString(&buildInfo["binary"]);
+    removeString(&buildInfo["version"]);
+}
+void sanitizeBuildInfo(ref JSONValue[string] buildInfo)
+{
+    removeString(&buildInfo["cwd"]);
+    removeStringIfExists("config" in buildInfo);
+    removeStringIfExists("lib" in buildInfo);
+    {
+        auto importPaths = buildInfo["importPaths"].array;
+        foreach(ref path; importPaths)
+        {
+            path = JSONValue(normalizeFile(path.str));
+        }
+    }
+}
+void sanitizeSyntaxNode(ref JSONValue value)
+{
+    if (value.type == JSON_TYPE.ARRAY)
+    {
+        foreach (ref element; value.array)
+        {
+            sanitizeSyntaxNode(element);
+        }
+    }
+    else if(value.type == JSON_TYPE.OBJECT)
+    {
+        foreach (name; value.object.byKey)
+        {
+            if (name == "file")
+                removeString(&value.object[name]);
+            else if (name == "offset")
+                removeNumber(&value.object[name]);
+            else if (!keepDeco && name == "deco")
+                removeString(&value.object[name]);
+            else
+                sanitizeSyntaxNode(value.object[name]);
+        }
+    }
+}
+
+void sanitizeSemantics(ref JSONValue[string] semantics)
+{
+    import std.array : appender;
+
+    auto modulesArrayPtr = &semantics["modules"].array();
+    auto newModules = appender!(JSONValue[])();
+    foreach (ref semanticModuleNode; *modulesArrayPtr)
+    {
+        auto semanticModule = semanticModuleNode.object();
+        auto moduleName = semanticModule["name"].str;
+        if(moduleName.startsWith("std.", "core.", "etc.") || moduleName == "object")
+        {
+           // remove druntime/phobos modules since they can change for each
+           // platform
+           continue;
+        }
+        auto fileNode = &semanticModule["file"];
+        *fileNode = JSONValue(normalizeFile(fileNode.str));
+        newModules.put(JSONValue(semanticModule));
+    }
+    *modulesArrayPtr = newModules.data;
+}
+
+auto normalizeFile(string file)
+{
+    version(Windows)
+        return file.replace("\\", "/");
+    return file;
+}
--- a/changelog/more_json.dd
+++ b/changelog/more_json.dd
@@ -1,0 +1,37 @@
+Added more information to the json output
+
+The following objects were to the json output:
+---
+[
+  {
+    "kind" : "compilerInfo",
+    "binary" : "<filename-of-compiler-binary>",
+    "version" : "<compiler-version>",
+    "supportsIncludeImports" : true
+  },
+  {
+    "kind" : "buildInfo",
+    "config": "<filename-of-config-file>",
+    "cwd": "<cwd-during-compiler-invocation>",
+    "importPaths": [
+      "<import-path1>",
+      "<import-path2>",
+      //...
+    ]
+  },
+  // ...
+  {
+    "kind": "semantics",
+    "modules": [
+      {
+        "name": "<module-name>"
+        "file": "<module-filename>",
+        // isRoot is true if the module is going to be taken
+        // to object code.
+        "isRoot": true|false,
+      },
+      // ...
+    ]
+  }
+]
+---
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -322,6 +322,7 @@ extern (C++) final class Module : Package
     uint numlines;              // number of lines in source file
     int isDocFile;              // if it is a documentation input file, not D source
     bool isPackageFile;         // if it is a package.d
+    Strings contentImportedFiles; // array of files whose content was imported
     int needmoduleinfo;
     /**
        How many unit tests have been seen so far in this module. Makes it so the
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -4282,6 +4282,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             return setError();
         }
 
+        sc._module.contentImportedFiles.push(name);
         if (global.params.verbose)
             fprintf(global.stdmsg, "file      %.*s\t(%s)\n", cast(int)se.len, se.string, name);
         if (global.params.moduleDeps !is null)
--- a/src/dmd/module.h
+++ b/src/dmd/module.h
@@ -85,6 +85,7 @@ public:
     unsigned numlines;  // number of lines in source file
     int isDocFile;      // if it is a documentation input file, not D source
     bool isPackageFile; // if it is a package.d
+    Strings contentImportedFiles;  // array of files whose content was imported
     int needmoduleinfo;
     /**
        How many unit tests have been seen so far in this module. Makes it so the
--- a/test/Makefile
+++ b/test/Makefile
@@ -203,24 +203,26 @@ fail_compilation_test_results=$(addsuffix .out,$(addprefix $(RESULTS_DIR)/,$(fai
 
 all: run_tests
 
-$(RESULTS_DIR)/runnable/%.d.out: runnable/%.d $(RESULTS_DIR)/.created $(RESULTS_DIR)/d_do_test$(EXE) $(DMD)
+test_tools: $(RESULTS_DIR)/d_do_test$(EXE) $(RESULTS_DIR)/sanitize_json$(EXE)
+
+$(RESULTS_DIR)/runnable/%.d.out: runnable/%.d $(RESULTS_DIR)/.created test_tools $(DMD)
 	$(QUIET) $(RESULTS_DIR)/d_do_test $(<D) $* d
 
-$(RESULTS_DIR)/runnable/%.sh.out: runnable/%.sh $(RESULTS_DIR)/.created $(RESULTS_DIR)/d_do_test$(EXE) $(DMD)
+$(RESULTS_DIR)/runnable/%.sh.out: runnable/%.sh $(RESULTS_DIR)/.created test_tools $(DMD)
 	$(QUIET) echo " ... $(<D)/$*.sh"
 	$(QUIET) ./$(<D)/$*.sh
 
-$(RESULTS_DIR)/compilable/%.d.out: compilable/%.d $(RESULTS_DIR)/.created $(RESULTS_DIR)/d_do_test$(EXE) $(DMD)
+$(RESULTS_DIR)/compilable/%.d.out: compilable/%.d $(RESULTS_DIR)/.created test_tools $(DMD)
 	$(QUIET) $(RESULTS_DIR)/d_do_test $(<D) $* d
 
-$(RESULTS_DIR)/compilable/%.sh.out: compilable/%.sh $(RESULTS_DIR)/.created $(RESULTS_DIR)/d_do_test$(EXE) $(DMD)
+$(RESULTS_DIR)/compilable/%.sh.out: compilable/%.sh $(RESULTS_DIR)/.created test_tools $(DMD)
 	$(QUIET) echo " ... $(<D)/$*.sh"
 	$(QUIET) ./$(<D)/$*.sh
 
-$(RESULTS_DIR)/fail_compilation/%.d.out: fail_compilation/%.d $(RESULTS_DIR)/.created $(RESULTS_DIR)/d_do_test$(EXE) $(DMD)
+$(RESULTS_DIR)/fail_compilation/%.d.out: fail_compilation/%.d $(RESULTS_DIR)/.created test_tools $(DMD)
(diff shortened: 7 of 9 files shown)


marler8997:
Hmmm, this is going to make testing more difficult. Now that the JSON file contains semantic data, it's contents are different for each platform, and also changes depending on the order that files were passed to the compiler.

timotheecour:
1 quick sanity check: seeing whether on a large project, adding  -Xf=file
slows down compilation (or consumes more memory ?) than without. Because
each function declaeation (+other things) gets serialized. If so, we'd need
a syntax like -X=imports,versions,libs to control what gets exported.

in any case, its a less gacky version than -v

marler8997:
Based on the code, I don't anticipate json code generation having a big performance impact.  However, if it does then filtering can always be added later as an after thought.  Maybe:
```
-X (enable json code generation)
-Xf=<file> (set file to output json to)
-Xo=-functions  (An example interface to filter out function from json code generation)
```

CyberShadow:
> `-Xo=-functions  (An example interface to filter out function from json code generation)`

BTW, if we ever do need to add that, it should probably be a list of things to include rather than a list of things to exclude. Tools that would make use of this option will know exactly what they want, so any information that's added later will not be useful to them.

marler8997:
yeah good point

marler8997:
@WalterBright We may want to try to give this PR some attention early on because if this PR along with the changes to rdmd don't get in before the next release, then we will have an extra use case to handle, namely, the case where you have a compiler that supports `-i` but doesn't generate the JSON output that is needed.  I'd like to avoid that if possible.

src/dmd/json.d
```diff
@@ -793,19 +793,62 @@ public:
...
+    {
+        objectStart();
+        property("kind", "semantics");
+        propertyStart("modules");
+        arrayStart();
+        foreach (m; Module.amodules)
+        {
+            objectStart();
+            if(m.md)
+                property("name", m.md.toChars());
```
timotheecour: how about:
property("name", m.md ? m.md.toChars() : "");
to make each entry have each field?

src/dmd/json.d
```diff
@@ -793,19 +793,62 @@ public:
...
+            property("file", m.srcfile.toChars());
```
timotheecour: how about `absolutePath ( m.srcfile.toChars() )` so that if rdmd is called from a different directory, it'll still work?

src/dmd/json.d
marler8997: I was following the convention established here (https://github.com/dlang/dmd/pull/7757/files/a90c983d6e436e46dde3b1ee07fdd280ea1396e7#diff-1d73c979f233c43e47fb88fac32c21b3L475)

I don't really see much difference whether or not there is an empty name or no name at all, but following the existing convention seems the better choice.

src/dmd/json.d
marler8997: hmmmm, interesting thought.  I'd have to think about that one.  This is mimicking the current functionality from the verbose output, but maybe it can be improved.  I'll come back to this.

timotheecour: NOTE: "contentImports" appears to use absolutePath (i tried with -J. and `import("foo.txt")` ), so I would make sense to be consistent. I suggest using absolutePath for `file` as well (and wherever else it makes sense)

marler8997: I believe that would be changing behavior though...we could end up with a regression if we change the behavior.

timotheecour: well, that info wasn't in json output before... so it's one of those gray area things.
one option is to allow configuring json output (as I suggested somewhere else), which is OK if we have good way to customize, eg:

`-Xf=file -Xopt=abspaths,-functions`

marler8997: What I mean is that `rdmd` will be using that field the same way it was using the `import   ` line from the verbose output.  And in that case it was using the same `m.srcfile.toChars()` expression.  If we use the absolute path instead moving forward, I just want to make sure it won't break anything.

marler8997: Now that I think about it, there might be a bug in the current design.  If you build with rdmd, and then you change directories and build again, then all the relative module filenames will be incorrect.  It might just force you to do a rebuild, but that's no good.  We could fix this using absolute paths, or we could save off the original CWD that rdmd was using when it first performed the build.

src/dmd/json.d
```diff
@@ -793,19 +793,62 @@ public:
...
+            propertyBool("isRoot", m.isRoot());
+            if(m.contentImportedFiles.dim > 0)
+            {
+                propertyStart("contentImports");
+                arrayStart();
+                foreach (file; m.contentImportedFiles)
+                {
+                    value(file);
```
timotheecour: BUG! missing separaters in generated JSON
"contentImports" : [
"/tmp/myfile.txt""/tmp/myfile.txt"
    ]

src/dmd/json.d
CyberShadow: There should be a test for that in the rdmd test suite, FWIW.

timotheecour: there may even be corner cases where it picks up wrong file eg:

ls
main.d
fun.d
temp/fun.d // slightly different version main.d

rdmd main.d
cd temp
rdmd ../main.d
=> can it get confused and get the wrong fun.d?

Here's a good compromise:

add a CWD field, and when rdmd reads a path (any path) call  buildPath(json.CWD, json.path) (does the right thing w absolute paths)

and allow for:
rdmd main.d
cd temp
rdmd ../main.d => should reuse cache and not rebuild

timotheecour: but we should definitely make behavior consistent between contentImports and modules.file:
import("fun.txt") should not get auto-transformed in an absolute path either.

marler8997: I'm thinking that the JSON file saves the CWD in the new `buildInfo` object.

wilzbach (commented):
Do we really need to generate the JSON file?
It makes future updates of the JSON file more difficult as the line numbers on the diff won't reflect the ones in the generated script. Also it's easy to screw up the JSON generation. How about moving the things that are platform-specific into a different test?

test/compilable/extra-files/gen_json_out.d
```diff
@@ -1,4 +1,14 @@
-[
+import std.stdio : write;
+void main()
+{
+    write(
+`[
+ {
+  "kind" : "buildInfo",
+  "binary" : "../generated/linux/release/64/dmd",
+  "version" : "v2.078.1-beta.1-494-g479722a-dirty",
```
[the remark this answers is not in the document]
marler8997: they are "grepped" out along with some other types of lines (see json-postscript.sh).  I thought keeping an example in the actual expected results didn't hurt though.

wilzbach:
>> Do we really need to generate the JSON file?
>> It makes future updates of the JSON file more difficult as the line numbers on the diff won't reflect the ones in the generated script. Also it's easy to screw up the JSON generation. How about moving the things that are platform-specific into a different test?
> Could you elaborate on your idea a bit more?

Here's what I would do:
- Filter out all dynamic content from the main `json.d` test - I think you already realized that `compilerBinary` and `version` are rather tricky to handle
- Create a special test for tests, which are a bit more error tolerant, e.g. by using `sed` to replace the version to a static string

> I'm thinking a better idea might be to create a small json comparer program. Since we are adding a list of all the modules that are being imported (directly or indirectly) this means that dmd and druntime/phobos will have to have a knife edge change whenever that import structure changes

I would keep it simply and do checks like:
- `object` is part of the list (the path might be different for LDC/GDC)
- the number of modules is > 0
- maybe a sanity check with `-v`

marler8997:
I'm thinking we create a "sanitize" program that takes a JSON file, and sanitizes it removing all the environment specific stuff.  Then we can remove the JSON generation code ("gen_json_out.d") and bring back the raw `json.out`.  Then just "sanitize" the JSON generated by the test.  The sanitizer will need to remove all the druntime/phobos modules from the semantic module list.

timotheecour:
the json file comparison should be done by parsing the json, then filtering out nodes you want to compare, seems more robust than grep / sed based tools

marler8997:
I've created a `sanitize_json` tool that takes care of removing all the environment specific stuff.  Note that I included a json parser in the test suite because I needed the parser to maintain the original order of the object members.   This allowed me to bring back the original `json.out` and we can still use `diff` to compare the expected vs actual json output.

src/dmd/json.d
```diff
@@ -34,6 +34,12 @@ import dmd.mtype;
 import dmd.root.outbuffer;
 import dmd.visitor;
 
+version(Windows) {
+    extern (C) char* getcwd(char* buffer, size_t maxlen);
```
timotheecour: is that a concern?
https://msdn.microsoft.com/en-us/library/ms235450.aspx
This POSIX function is deprecated. Use the ISO C++ conformant _getcwd instead.

marler8997: This declaration was copied from `dmodule.d`.  It it's a problem then the fix should be a separate PR.

wilzbach: Always open an issue about such stuff, otherwise it well get forgotten. 
I just did so for this: https://issues.dlang.org/show_bug.cgi?id=18291

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+import std.bigint;
+import std.internal.cstring : tempCString;
+import std.string : startsWith, replace;
+import std.format : formattedWrite, format;
+import std.file : readText;
+import std.stdio;
+
+import json;
+
+__gshared File outFile;
```
wilzbach: Official code should use `shared` whenever possible, though in this case neither is needed.

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+}
+int main(string[] args)
+{
+    args = args[1 .. $];
+    if (args.length == 0)
+    {
+        usage();
+        return 1;
+    }
+    string inFilename = args[0];
```
wilzbach: Why not use `stdin` if no arguments are provided?
If you don't like this, "-" is a common choice. If `-Xf` would support using stdout as output, this would allow to completely avoid the need of a temporary file.

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+        else if (kind == "buildInfo")
+            sanitizeBuildInfo(obj);
+        else if(kind == "module")
+            sanitizeSyntaxNode(objNode);
+        else if(kind == "semantics")
+            sanitizeSemantics(obj);
+    }
+}
+
+void removeString(JsonValue* value)
```
wilzbach: `ref`

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+{
+    assert(value.type == JsonType.string);
+    *value = JsonValue("VALUE_REMOVED_FOR_TEST");
+}
+void removeNumber(JsonValue* value)
```
wilzbach: `ref`

wilzbach: You could use templates here, e.g. `remove!string`

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+    auto modulesNode = semantics.getRef("modules");
+    auto modulesArray = modulesNode.value.array;
+    auto newModules = appender!(JsonValue[])();
+    foreach (ref semanticModuleNode; modulesArray)
+    {
+        auto semanticModule = semanticModuleNode.object;
+        auto moduleName = semanticModule.getAs!string("name");
+        if(moduleName.startsWith("std.") ||
+           moduleName.startsWith("core.") ||
+           moduleName.startsWith("etc.") ||
```
wilzbach: `startsWith` is variadic and supports multiple needles.

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+           moduleName == "object")
+        {
+           continue;
+        }
```
wilzbach: Without the special casing for Windows:

```d
modulesArray
.map!(m => object)
.filter!(m => !m.getAs!string("name").startsWith("std.", "core.", "etc."))
.each!(m => newModules.put(JsonValue(m)));
```

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+        newModules.put(JsonValue(semanticModule));
+    }
+    *modulesNode.value = JsonValue(newModules.data);
+}
+
+// Mimics the json format generated by DMD
+// so that the resulting JSON can be compared
+// using a simple diff
+struct Writer
```
wilzbach: Maybe `JsonWriter`

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+    string jsonText;
+    {
+        auto inFile = File(inFilename, "r");
+        auto inFileSizeLong = inFile.size;
+        assert(inFileSizeLong < size_t.max, "file too large");
+        size_t inFileSize = cast(size_t)inFileSizeLong;
+        char[] buffer = new char[inFileSize + 1];
+        assert(inFileSize == inFile.rawRead(buffer[0..inFileSize]).length);
+        buffer[inFileSize] = '\0';
+        jsonText = cast(string)buffer;
```
wilzbach: check out `assumeUnique` from `std.exception`.

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+
+    string jsonText;
+    {
+        auto inFile = File(inFilename, "r");
+        auto inFileSizeLong = inFile.size;
+        assert(inFileSizeLong < size_t.max, "file too large");
+        size_t inFileSize = cast(size_t)inFileSizeLong;
+        char[] buffer = new char[inFileSize + 1];
+        assert(inFileSize == inFile.rawRead(buffer[0..inFileSize]).length);
+        buffer[inFileSize] = '\0';
```
wilzbach: Why is this necessary. You don't use any C function?

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
+import std.typecons : Flag, Yes, No;
+import std.array : Appender, appender;
+import std.bigint;
+import std.internal.cstring : tempCString;
```
wilzbach: Not used.

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
+import std.typecons : Flag, Yes, No;
```
wilzbach: Not used

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+import std.array : Appender, appender;
```
wilzbach: Only `appender` is used. It's generally a best practice to do local imports - that way it's easy not to forget to update them when code gets modified / refactored.

test/sanitize_json.d
```diff
@@ -0,0 +1,231 @@
...
+import std.bigint;
+import std.internal.cstring : tempCString;
+import std.string : startsWith, replace;
+import std.format : formattedWrite, format;
+import std.file : readText;
```
wilzbach: Not used.

test/json.d
```diff
@@ -0,0 +1,308 @@
...
+        {
+            if(*next == '"')
+            {
+               auto str = start[0 .. next - start];
+               next++;
+               return str;
+            }
+        }
+    }
+    BigInt parseNumber()
```
wilzbach: Json only supports floating point numbers:

> A number is a sequence of decimal digits with no superfluous leading zero. It may have a preceding minus
sign (U+002D). It may have a fractional part prefixed by a decimal point (U+002E). It may have an exponent,
prefixed by e (U+0065) or E (U+0045) and optionally + (U+002B) or – (U+002D). The digits are the code
points U+0030 through U+0039.

http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf

test/json.d
```diff
@@ -0,0 +1,308 @@
...
+    {
+        if(*str != *check)
+            return false;
+    }
+    return true;
+}
+class JsonParseException : Exception
+{
+    this(string msg, string file, size_t line)
+    { super(msg, file, line); }
```
wilzbach: Allman braces

test/json.d
```diff
@@ -0,0 +1,308 @@
...
+    immutable(char)* next;
+    string filenameForErrors;
+    size_t line(immutable(char)* src)
+    {
+        size_t lineCount = 1;
+        for (auto countPtr = start; countPtr < next; countPtr++)
+        {
+            if (*countPtr == '\n')
+                lineCount++;
+        }
```
wilzbach: How about: `text.filter!(a == '\n').walkLength`

Also there is `newline` in `std.ascii` and `std.uni`

test/json.d
```diff
@@ -0,0 +1,308 @@
...
+    parser.skipWhitespace();
+    if (*parser.next != '\0')
+        throw parser.parseError(parser.line(parser.next), "invalid json, multiple root values");
+    return value;
+}
+
+private bool isspace(char c)
+{
+    return c == ' ' || c == '\t' || c == '\r' || c == '\n';
+}
```
wilzbach: Why not use `std.ascii`?

wilzbach (commented):
I think we should try to avoid re-doing a custom Json parser if possible. Are you sure simply CLI tools don't work?

marler8997:
> I think we should try to avoid re-doing a custom Json parser if possible. Are you sure simply CLI tools don't work?

@timotheecour and I discussed the best solution here.

The first question was whether we should compare the JSON via data structures or at the file level.  I've concluded that performing the final diff on the file level is the most helpful.  Since the `json.out` file is what needs to be maintained, then showing exactly where there are differences in the file itself is more useful than trying to showing a diff between 2 JSON data structures.

Given this, we need a way to sanitize a JSON file (remove/normalize environment specific stuff).  Up to this point we could do this by simply removing the lines that contained environment specific stuff, however, with the newly added data, we need to remove objects that can span multiple lines. Given this, moving to a sanitizer that operates on JSON data structures makes sense here.  However, when we load in the JSON write it back out, we want to preserve the original order for object members.  This is a problem since all the JSON parsers that I know of don't preserve this order, hence the need for a new JSON parser.  Note that I have made an attempt to add support for this in `std.json` (https://github.com/dlang/phobos/pull/6059), if that gets accepted we can swap in that implementation and remove this one.

Also note that by using a small D program to sanitize the JSON, we can keep more of the information in the file to compare.  For example, we are now keeping the module filenames by normalizing the directory separators to `/`, whereas before we were completely removing them.  Since we are adding new JSON, having the sanitation logic in D also makes it easier to maintain moving forward.

I don't think adding a JSON parser is ideal but here are some thoughts that help justify it
1. JSON is very simple, adding a parser for a specific purpose is quite easy.  I wrote this one from scratch in less than an hour
2. JSON is old now and is not going to change.
3. This JSON parser only has to support the JSON that dmd outputs, which is subset of the already small JSON spec.

test/sanitize_json.d
marler8997: with stdin you can't use the file size to allocate the buffer to read the string into.   also by reading in a file you can use that filename in error messages :)

test/sanitize_json.d
marler8997: The JSON parser relies on the file content being null-terminated.  Makes the parser simpler.

test/sanitize_json.d
marler8997: My brain doesn't think in these algorithm functions very well :(  Is this easier for you to understand than the current loop?

test/sanitize_json.d
marler8997: This is just personal preference, but when a function is modifying a value that's passed into it, I prefer to see some indication of this when calling the function, i.e.
```
// I prefer
removeString(&myValue);
// rather than
removestring(myValue);
```
To me this make it obvious that the function is modifying the value.  Note that I only use this convention for value types.  For example if you had a function that modified an integer as a parameter such as `square`, I would rather have `void square(int* i)` rather than `void square(ref int i)` because I would prefer to see `square(&myInt)` rather than `square(myInt)`.

wilzbach: How about `out` then?

See also: https://github.com/dlang/dmd/pull/7754/files

marler8997: That has the same problem doesn't it?  You can't see from the call that the value is being modified.
```D
void squareUsingOut(out int i);
// vs
void squareUsingPtr(int* i);

int i = 100;
squareUsingOut(i); // not obvious that i is being modified
squareUsingPtr(&i); // obvious that square could be modifying i
```

CyberShadow:
Was the option to add a secret switch to DMD to only emit JSON fields that are worth testing considered?

marler8997:
Hmmmm, I'm sure the argument against this would be that this adds logic to the compiler that doesn't add any value to the compiler itself.  It just works around an issue with the test suite.

timotheecour:
> Was the option to add a secret switch to DMD to only emit JSON fields that are worth testing considered?

plus there's issue of certain fields like `contentImports` that always output using `absolutePath` (which IMO is a design flaw but that's another topic), so that field would have to be skipped from testing (ie testing won't catch regressions related to `contentImports`)

timotheecour:
* top-level msg: `s/target library file is -lib is used /target library file if -lib is used /`
* how about adding `-I` (import search paths) to the json?

marler8997:
> * how about adding -I (import search paths) to the json?

Yeah I could see that being useful for sure. I'll add em.

marler8997:
@andralex has determined that adding policies to `std.json` is not a desirable enhancement (https://github.com/dlang/phobos/pull/6059#issuecomment-361430810).  Given that we can't use `std.json` to sanitize the file, using the provided JSON parser seems like a solid option to me.

CyberShadow:
> Given that now we can't use std.json to sanitize the file,

Sorry, I don't think we finished [the previous conversation](https://github.com/dlang/phobos/pull/6084#issuecomment-361068225). Why not use `std.json`, and rely on its current behavior of sorting key names lexicographically? Won't that allow achieving the same goal without adding a JSON parser to the DMD repository?

marler8997:
That would work "functionally".  The only drawback is that the two files won't be in the same order that was generated from DMD, which would look odd.  The order of the fields as generated by DMD makes sense, i.e. name first, followed by other identifying properties, followed by more verbose fields such as large objects.  Would be nice to keep it in the same order.  I refuse to accept that sanitizing a JSON file while preserving the original object field order is an insurmountable task in D :)

CyberShadow:
> The only drawback is that the two files won't be in the same order that was generated from DMD, which would look odd.

IMHO that's fine, considering the circumstances, but it's not up to me to decide.

> I refuse to accept that sanitizing a JSON file while preserving the original object field order is an insurmountable task in D :)

Certainly not insurmountable, but it may be worth keeping our feet on the ground in this case. It's just a test case file that hopefully won't be edited / looked at often at all, so maybe just go for the simplest solution.

marler8997:
> Certainly not insurmountable, but it may be worth keeping our feet on the ground in this case. It's just a test case file that hopefully won't be edited / looked at often at all, so maybe just go for the simplest solution.

Yes I see your point.  I'll concede this point and go with your solution.  Be prepared for a large `json.out` diff :)

marler8997:
Ok I caved, the JSON sanitizer uses `std.json` so all the object fields are sorted alphabetically.

marler8997:
Ping.  Trying to get this PR finalized and integrated so we have some time before the 2.079 release to modify rdmd to use it (https://github.com/dlang/tools/pull/292).

JinShil:
Restarted Jenkins

test/sanitize_json.d
```diff
@@ -0,0 +1,146 @@
...
+        foreach (name; value.object.byKey)
+        {
+            if (name == "file")
+                removeString(&value.object[name]);
+            else if (name == "deco")
+                removeString(&value.object[name]);
+            else if (name == "offset")
+                removeNumber(&value.object[name]);
+            else
+                sanitizeSyntaxNode(value.object[name]);
```
wilzbach: Not a fan of this (I know that's how the status quo is).
Ideally this would be configurable via a flag, s.t. special platform-specific, short tests which test that the `deco` is correctly set in the json output can be added, but that's definitely not something which shouldn't block this PR.

wilzbach (approved):
Looks good, but please add a short changelog entry.

test/sanitize_json.d
marler8997: Ok, I added a `--keep-deco` flag that can be used by tests in the future.

marler8997:
Added the changelog

src/dmd/json.d
```diff
@@ -789,19 +795,78 @@ public:
...
+            property("file", m.srcfile.toChars());
+            propertyBool("isRoot", m.isRoot());
+            if(m.contentImportedFiles.dim > 0)
+            {
+                propertyStart("contentImports");
+                arrayStart();
+                foreach (file; m.contentImportedFiles)
+                {
+                    item(file);
+                }
```
wilzbach: This isn't covered.

wilzbach (approved):
Thanks a lot for putting all the hard work into this!
I'm going to merge this now to unblock your work on rdmd - could you do a quick follow up to cover the uncovered output of the imported files? Thanks!

test/sanitize_json.d
timotheecour: > with stdin you can't use the file size to allocate the buffer to read the string into
just consume all stdin and that solves this problem; doesn't need to be processing input line by line for this use case.

> also by reading in a file you can use that filename in error messages :)
```
error: at __stdin.d:(12,10) { "foo": bar"} expected `"` got `b`
```

NOTE: __stdin.d is what is used in src/dmd/mars.d for same use case:
```
 if (strcmp(arguments[i + 1], "-") == 0)
*                        files.push("__stdin.d");
```

test/sanitize_json.d
```diff
@@ -0,0 +1,151 @@
...
+    {
+        outFile = stdout;
+    }
+    else if(args.length == 2)
+    {
+        outFile = File(args[1], "w");
+    }
+    else
+    {
+        writeln("Error: too many command line arguments");
```
timotheecour: stderr?

PR #7758, opened 2018-01-21
Highlight error messages:  Fix a few leftovers and inconsistencies

JinShil:
I may submit a couple more PRs like this.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -607,12 +607,12 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         if (StorageClass stc = dsym.storage_class & (STC.synchronized_ | STC.override_ | STC.abstract_ | STC.final_))
         {
             if (stc == STC.final_)
-                dsym.error("cannot be final, perhaps you meant const?");
+                dsym.error("cannot be `final`, perhaps you meant `const`?");
             else
             {
                 OutBuffer buf;
                 stcToBuffer(&buf, stc);
-                dsym.error("cannot be %s", buf.peekString());
+                dsym.error("cannot be `%s`", buf.peekString());
             }
             dsym.storage_class &= ~stc; // strip off
         }
@@ -692,14 +692,14 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
 
         if ((dsym.storage_class & (STC.ref_ | STC.parameter | STC.foreach_ | STC.temp | STC.result)) == STC.ref_ && dsym.ident != Id.This)
         {
-            dsym.error("only parameters or foreach declarations can be ref");
+            dsym.error("only parameters or `foreach` declarations can be `ref`");
         }
 
         if (dsym.type.hasWild())
         {
             if (dsym.storage_class & (STC.static_ | STC.extern_ | STC.tls | STC.gshared | STC.manifest | STC.field) || dsym.isDataseg())
             {
-                dsym.error("only parameters or stack based variables can be inout");
+                dsym.error("only parameters or stack based variables can be `inout`");
             }
             FuncDeclaration func = sc.func;
             if (func)
@@ -717,7 +717,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                 }
                 if (!isWild)
                 {
-                    dsym.error("inout variables can only be declared inside inout functions");
+                    dsym.error("`inout` variables can only be declared inside `inout` functions");
                 }
             }
         }
@@ -745,12 +745,12 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         {
             if (dsym.storage_class & (STC.field | STC.out_ | STC.ref_ | STC.static_ | STC.manifest | STC.tls | STC.gshared) || !fd)
             {
-                dsym.error("globals, statics, fields, manifest constants, ref and out parameters cannot be scope");
+                dsym.error("globals, statics, fields, manifest constants, ref and out parameters cannot be `scope`");
             }
             if (!(dsym.storage_class & STC.scope_))
             {
                 if (!(dsym.storage_class & STC.parameter) && dsym.ident != Id.withSym)
-                    dsym.error("reference to scope class must be scope");
+                    dsym.error("reference to `scope class` must be `scope`");
             }
         }
 
@@ -760,14 +760,14 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             if (dsym._init && dsym._init.isVoidInitializer() && dsym.type.hasPointers()) // get type size
             {
                 if (sc.func.setUnsafe())
-                    dsym.error("void initializers for pointers not allowed in safe functions");
+                    dsym.error("`void` initializers for pointers not allowed in safe functions");
             }
             else if (!dsym._init &&
                      !(dsym.storage_class & (STC.static_ | STC.extern_ | STC.tls | STC.gshared | STC.manifest | STC.field | STC.parameter)) &&
                      dsym.type.hasVoidInitPointers())
             {
                 if (sc.func.setUnsafe())
-                    dsym.error("void initializers for pointers not allowed in safe functions");
+                    dsym.error("`void` initializers for pointers not allowed in safe functions");
             }
         }
 
@@ -1423,7 +1423,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             }
             if (se.sz != 1)
             {
-                pd.error("mangled name characters can only be of type char");
+                pd.error("mangled name characters can only be of type `char`");
                 goto Ldecl;
             }
             version (all)
@@ -1500,7 +1500,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             goto Lnodecl;
         }
         else
-            pd.error("unrecognized `pragma(%s)`", pd.ident.toChars());
+            error(pd.loc, "unrecognized `pragma(%s)`", pd.ident.toChars());
     Ldecl:
         if (pd.decl)
         {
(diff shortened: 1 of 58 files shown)


src/dmd/expression.d
```diff
@@ -2018,7 +2018,7 @@ extern (C++) abstract class Expression : RootObject
             FuncDeclaration ff = outerfunc;
             if (sc.flags & SCOPE.compile ? ff.isPureBypassingInference() >= PURE.weak : ff.setImpure())
             {
-                error("`pure` %s `%s` cannot call impure %s `%s`",
+                error("`pure` %s `%s` cannot call im`pure` %s `%s`",
```
WalterBright: I don't think highlighting parts of words improves readability. It just looks wrong.

src/dmd/func.d
```diff
@@ -2410,19 +2410,19 @@ void MODMatchToBuffer(OutBuffer* buf, ubyte lhsMod, ubyte rhsMod)
     bool sharedMismatchOnly = ((lhsMod ^ rhsMod) == MODFlags.shared_);
 
     if (lhsMod & MODFlags.shared_)
-        buf.writestring("shared ");
+        buf.writestring("`shared` ");
     else if (sharedMismatch && !(lhsMod & MODFlags.immutable_))
-        buf.writestring("non-shared ");
+        buf.writestring("non-`shared` ");
```
WalterBright: please no highlighting parts of words

JinShil: Would you prefer \`non-shared`, or just not highlighted at all?

JinShil: I went with `non-shared`, `impure`, `non-@nogc`, etc..

WalterBright: Just non-shared, impure, etc., please

src/dmd/expression.d
```diff
@@ -2018,7 +2018,7 @@ extern (C++) abstract class Expression : RootObject
             FuncDeclaration ff = outerfunc;
             if (sc.flags & SCOPE.compile ? ff.isPureBypassingInference() >= PURE.weak : ff.setImpure())
             {
-                error("`pure` %s `%s` cannot call impure %s `%s`",
+                error("`pure` %s `%s` cannot call `impure` %s `%s`",
```
WalterBright: impure is not D code, please do not highlight it

src/dmd/expression.d
```diff
@@ -2082,7 +2082,7 @@ extern (C++) abstract class Expression : RootObject
                     break;
                 if (sc.flags & SCOPE.compile ? ff.isPureBypassingInference() >= PURE.weak : ff.setImpure())
                 {
-                    error("`pure` %s `%s` cannot access mutable static data `%s`",
+                    error("`pure` %s `%s` cannot access `mutable` static data `%s`",
```
WalterBright: mutable is not code, please do not highlight it

src/dmd/expression.d
```diff
@@ -2217,7 +2217,7 @@ extern (C++) abstract class Expression : RootObject
             {
                 if (loc.linnum == 0) // e.g. implicitly generated dtor
                     loc = sc.func.loc;
-                error("`@nogc` %s `%s` cannot call non-`@nogc` %s `%s`",
+                error("`@nogc` %s `%s` cannot call `non-@nogc` %s `%s`",
```
WalterBright: Please just leave it as non-@nogc

src/dmd/func.d
```diff
@@ -2410,21 +2410,21 @@ void MODMatchToBuffer(OutBuffer* buf, ubyte lhsMod, ubyte rhsMod)
...
+        buf.writestring("`immutable` ");
     else if (lhsMod & MODFlags.const_)
-        buf.writestring("const ");
+        buf.writestring("`const` ");
     else if (lhsMod & MODFlags.wild)
-        buf.writestring("inout ");
+        buf.writestring("`inout` ");
     else
-        buf.writestring("mutable ");
+        buf.writestring("`mutable` ");
```
WalterBright: mutable is not D code

src/dmd/func.d
```diff
@@ -2410,21 +2410,21 @@ void MODMatchToBuffer(OutBuffer* buf, ubyte lhsMod, ubyte rhsMod)
     bool sharedMismatchOnly = ((lhsMod ^ rhsMod) == MODFlags.shared_);
 
     if (lhsMod & MODFlags.shared_)
-        buf.writestring("shared ");
+        buf.writestring("`shared` ");
     else if (sharedMismatch && !(lhsMod & MODFlags.immutable_))
-        buf.writestring("non-shared ");
+        buf.writestring("`non-shared` ");
```
WalterBright: non-shared is not D code

JinShil:
Removed all highlighting of negative/opposite attributes (e.g. non-shared, impure, mutable, etc..)

src/dmd/dsymbolsem.d
```diff
@@ -2743,44 +2743,44 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
             const(char)* sfunc;
             if (funcdecl.isStatic())
-                sfunc = "static";
+                sfunc = "`static`";
             else if (funcdecl.protection.kind == Prot.Kind.private_ || funcdecl.protection.kind == Prot.Kind.package_)
                 sfunc = protectionToChars(funcdecl.protection.kind);
             else
                 sfunc = "non-virtual";
-            funcdecl.error("%s functions cannot be abstract", sfunc);
+            funcdecl.error("%s functions cannot be `abstract`", sfunc);
```
ibuclaw: Could `non-virtual` be inverted as `final`? So that you don't just highlight one of the cases here.

src/dmd/dsymbolsem.d
```diff
@@ -1530,7 +1530,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
     Lnodecl:
         if (pd.decl)
         {
-            pd.error("pragma is missing closing `;`");
+            pd.error("is missing an ending `;`");
```
ibuclaw: I can't think of any better way to phrase this error. A word such as terminating might be better, but don't jump to that suggestion.

src/dmd/dsymbolsem.d
```diff
@@ -2991,13 +2991,13 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
                     if (fdv.isFinalFunc())
-                        funcdecl.error("cannot override final function `%s`", fdv.toPrettyChars());
+                        funcdecl.error("cannot override `final` function `%s`", fdv.toPrettyChars());
 
                     if (!funcdecl.isOverride())
                     {
                         if (fdv.isFuture())
                         {
-                            deprecation(funcdecl.loc, "@future base class method %s is being overridden by %s; rename the latter", fdv.toPrettyChars(), funcdecl.toPrettyChars());
+                            deprecation(funcdecl.loc, "`@future` base class method `%s` is being overridden by `%s`; rename the latter", fdv.toPrettyChars(), funcdecl.toPrettyChars());
```
ibuclaw: @future or @__future? (I wasn't in discussions when this feature was implemented).

src/dmd/dsymbolsem.d
```diff
@@ -3319,9 +3319,9 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                     // if the first parameter has a default argument, then the rest does as well
                     if (ctd.storage_class & STC.disable)
                     {
-                        ctd.deprecation("@disable'd constructor cannot have default "~
+                        ctd.deprecation("`@disable`'d constructor cannot have default "~
```
ibuclaw: Does @disable'd highlight really work here? Maybe could be rephrased to keep it.

ibuclaw (commented):
Would like a second opinion though. I don't think diagnostics is really my thing.

src/dmd/dsymbolsem.d
WalterBright: You're right, it's `@__future`. Good catch!

JinShil: Fixed

src/dmd/dsymbolsem.d
JinShil: Yeah, it's odd.  I reworded it.  Also because it calls the `.deprecated` member function it already fills in the "constructor x.y.this()`, so there's no need to redundantly repeat it in the error message.  There are many other error messages like that, but that's another PR.

src/dmd/dsymbolsem.d
JinShil: I like "terminating".  I think it's consistent with the nomenclature used in programming (terminating null character) and programming language theory (terminals).

src/dmd/dsymbolsem.d
JinShil: I like it.  Done.

WalterBright:
LGTM. If you've got more pickups, just do another PR.

src/dmd/dsymbolsem.d
```diff
@@ -2743,44 +2743,44 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
             if (funcdecl.isStatic())
-                sfunc = "static";
+                sfunc = "`static`";
             else if (funcdecl.protection.kind == Prot.Kind.private_ || funcdecl.protection.kind == Prot.Kind.package_)
                 sfunc = protectionToChars(funcdecl.protection.kind);
             else
-                sfunc = "non-virtual";
-            funcdecl.error("%s functions cannot be abstract", sfunc);
+                sfunc = "`final`";
+            funcdecl.error("%s functions cannot be `abstract`", sfunc);
```
ibuclaw: Does protectionToChars return backticked string? I was thinking ```"`%s`" function..."```

ibuclaw: Fix up can be done in next pr though.

JinShil: Yeah, I tried to do that, but `protectionToChars` is also used in one of the tests in the test suite to generate code, so I couldn't make the change without clobbering that test.  I have to find another way.  The `.error` and `.deprecation` methods also automatically prepend the `.toChars()` to the error message, and none of those strings are highlighted yet.  This is all coming in a future PR, but I've got to do some figuring first.

JinShil: I moved the tick marks to the error message in #7763.

PR #7760, opened 2018-01-22
Fix Issue 17630 - selective imports find symbols in private imports o…

RazvanN7:
…f other modules

### DIFF
--- a/changelog/fix17630.dd
+++ b/changelog/fix17630.dd
@@ -1,0 +1,21 @@
+Fix Issue 17630 - selective imports find symbols in private imports of other modules
+
+Selectively importing a symbol should work only if the symbol
+imported is defined or publicly imported in the imported module.
+Due to a compiler bug, selectively importing a symbol works even
+if the symbol is defined in a privately imported module in the
+imported module.
+-----------------------------------------------------------
+//a.d
+int bar;
+
+//b.d
+import a;
+
+//c.d
+import b : bar;
+------------------------------------------------------------
+The above code will now result in a deprecation message
+which states that `bar` cannot be accessed since it is
+privately imported in `b`.
+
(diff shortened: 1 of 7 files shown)


mathias-lang-sociomantic:
Sounds like this needs to be a deprecation first :(

wilzbach:
Thanks a lot for moving this forward, but I fear that this needs a deprecation message instead of a hard error.

andralex:
I'm discussing with @RazvanN7, he just pushed a change that issues a deprecation.

andralex (approved):

changelog/fix17630.dd
```diff
@@ -0,0 +1,5 @@
+Fix Issue 17630 - selective imports find symbols in private imports of other modules
+
+A private module-level import should not have any symbols
+visible outside the module scope. For more information
+see https://issues.dlang.org/show_bug.cgi?id=17630
```
jacob-carlborg: I prefer that the changelog entry is extended to explain the issue and the solution inline without having to read the issue.

jacob-carlborg (changes requested):

src/dmd/dsymbolsem.d
```diff
@@ -1145,10 +1145,12 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             sc.protection = imp.protection;
             for (size_t i = 0; i < imp.aliasdecls.dim; i++)
             {
+                bool flag;
                 AliasDeclaration ad = imp.aliasdecls[i];
                 //printf("\tImport %s alias %s = %s, scope = %p\n", toPrettyChars(), aliases[i].toChars(), names[i].toChars(), ad._scope);
-                if (imp.mod.search(imp.loc, imp.names[i]))
+                if (imp.mod.search(imp.loc, imp.names[i]) /*, IgnorePrivateImports*/)
```
jacob-carlborg: Why is this comment here?

jacob-carlborg (changes requested):

src/dmd/dsymbolsem.d
RazvanN7: When the deprecation is over, that argument should ne uncommented

jacob-carlborg: Ok.

jacob-carlborg (approved):

andralex:
Nice work!

PR #7762, opened 2018-01-22
Add support for Objective-C class/static methods

jacob-carlborg:
A few missing things:

- [x] The header files need to be updated
- [x] Could use some more documentation
- [x] Need to add changelog
- [x] A PR to update the documentation in dlang.org

Documentation of the Objective-C ABI has been started, available in a separate branch: https://github.com/jacob-carlborg/dmd/blob/objc_abi_docs/docs/objective-c_abi.md.

The dlang.org part https://github.com/dlang/dlang.org/pull/2246.

### DIFF
--- a/src/dmd/objc.d
+++ b/src/dmd/objc.d
@@ -12,9 +12,11 @@
 
 module dmd.objc;
 
+import dmd.aggregate;
 import dmd.arraytypes;
 import dmd.cond;
 import dmd.dclass;
+import dmd.declaration;
 import dmd.dmangle;
 import dmd.dmodule;
 import dmd.dscope;
@@ -131,6 +133,20 @@ Objc objc()
     return _objc;
 }
 
+
+/**
+ * Contains all data for a class declaration that is needed for the Objective-C
+ * integration.
+ */
+struct ObjcClassDeclaration
+{
+    /// `true` if this class is a metaclass.
+    bool isMeta = false;
+
+    /// The metaclass of this class.
+    ClassDeclaration metaclass;
+}
+
 // Should be an interface
 extern(C++) abstract class Objc
 {
@@ -144,9 +160,54 @@ extern(C++) abstract class Objc
 
     abstract void setObjc(ClassDeclaration cd);
     abstract void setObjc(InterfaceDeclaration);
+
     abstract void setSelector(FuncDeclaration, Scope* sc);
     abstract void validateSelector(FuncDeclaration fd);
     abstract void checkLinkage(FuncDeclaration fd);
+
+    /**
+     * Returns the `this` pointer of the given function declaration.
+     *
+     * This is only used for class/static methods. For instance methods, no
+     * Objective-C specialization is necessary.
+     *
+     * Params:
+     *  funcDeclaration = the function declaration to get the `this` pointer for
+     *
+     * Returns: the `this` pointer of the given function declaration, or `null`
+     *  if the given function declaration is not an Objective-C method.
+     */
+    abstract inout(AggregateDeclaration) isThis(inout FuncDeclaration funcDeclaration) const;
+
+    /**
+     * Creates and sets the metaclass on the given class/interface declaration.
+     *
+     * Will only be performed on regular Objective-C classes, not on metaclasses.
+     *
+     * Params:
+     *  classDeclaration = the class/interface declaration to set the metaclass on
+     */
+    abstract void setMetaclass(InterfaceDeclaration interfaceDeclaration) const;
+
+    /// ditto
+    abstract void setMetaclass(ClassDeclaration classDeclaration) const;
+
+
+    /**
+     * Returns Objective-C runtime metaclass of the given class declaration.
+     *
+     * `ClassDeclaration.ObjcClassDeclaration.metaclass` contains the metaclass
+     * from the semantic point of view. This function returns the metaclass from
+     * the Objective-C runtime's point of view. Here, the metaclass of a
+     * metaclass is the root metaclass, not `null`. The root metaclass's
+     * metaclass is itself.
+     *
+     * Params:
+     *  classDeclaration = The class declaration to return the metaclass of
+     *
+     * Returns: the Objective-C runtime metaclass of the given class declaration
+     */
+    abstract ClassDeclaration getRuntimeMetaclass(ClassDeclaration classDeclaration) const;
 }
 
 extern(C++) private final class Unsupported : Objc
@@ -179,6 +240,26 @@ extern(C++) private final class Unsupported : Objc
     override void checkLinkage(FuncDeclaration)
     {
         // noop
+    }
+
+    override inout(AggregateDeclaration) isThis(inout FuncDeclaration funcDeclaration) const
+    {
+        return null;
+    }
+
+    override void setMetaclass(InterfaceDeclaration) const
+    {
+        // noop
+    }
+
+    override void setMetaclass(ClassDeclaration) const
+    {
+        // noop
+    }
+
+    override ClassDeclaration getRuntimeMetaclass(ClassDeclaration classDeclaration) const
+    {
+        assert(0, "Should never be called when Objective-C is not supported");
     }
 }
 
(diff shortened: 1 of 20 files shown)


stefan-koch-sociomantic:
There is not a word about what this actually does.

jacob-carlborg:
> There is not a word about what this actually does.

Yeah, that's why the unchecked list item, "Could use some more documentation", in the description.

jacob-carlborg:
@stefan-koch-sociomantic  On a high level the Objective-C ABI looks something like this:

```d
extern (Objective-C) interface NSObject
{
    NSObject init() @selector("init");
}
NSObject o;
o.init();
```

The above method call is translated to a regular C function call in the form of: `objc_msgSend(o, "init")`. `o` would be the `this` pointer and `"init"` is what was specified in the `@selector` UDA in the method declaration. That is for a regular instance method. For a class method it works the same way, it's just a different `this` pointer. In Objective-C, a class is an object and it's an instance of a metaclass. An instance of `ClassDeclaration` is used as the Objective-C metaclass.

docs/objective-c_abi.md
```diff
@@ -0,0 +1,15 @@
+# Objective-C ABI
+
+There are several Objective-C ABIs available:
+
+* Apple legacy ABI (also know as fragile ABI) - used on macOS 32bit and the old
```
ibuclaw: Just so I have the link correct in my head, this is also known as the Nextstep abi?

jacob-carlborg: Yes. I've moved the work on the ABI documentation to a separate branch until it's ready [1]. The description in the separate branch is a bit more clear and thorough.

[1] https://github.com/jacob-carlborg/dmd/blob/objc_abi_docs/docs/objective-c_abi.md

ibuclaw: OK.

I've just checked, and gcc objc seems to support all of these ABIs.  I should see if I can re-use the existing code sometime in gdc.

jacob-carlborg: Before Apple used Clang, they used GCC with this ABI that is implemented here. I don't think anything has changed since then. I'm not sure if that was only supported in their own version of GCC or in the standard version as well.

jacob-carlborg: BTW, some code review would be nice 😉.

JinShil:
I don't see any issue with the code in this PR, but I don't know anything about Objective-C, or what D's goals are in supporting it.  Isn't Objective-C reaching end of life with the emergence of Swift?

jacob-carlborg:
> I don't see any issue with the code in this PR, but I don't know anything about Objective-C, or what D's goals are in supporting it.

D's goals are to support as much of Objective-C as possible or what's necessary to implement meaningful Cocoa applications [1].

>  Isn't Objective-C reaching end of life with the emergence of Swift?

All of the Apple frameworks are still implemented in Objective-C. Apple is not, most likely, going to replace them in the near future. Swift can interface with Objective-C as well, which is a crucial part since all the frameworks are written in Objective-C. That means that D can interface with Swift code as well, assuming both use the Objective-C interface. Apple might implement new frameworks using Swift and possible rewrite existing frameworks as well, but they would need to support an Objective-C interface for a significant future.

Why was the `WIP` label added? I think it's possible to review even if some documentation is missing.

[1] https://github.com/dlang/DIPs/blob/master/DIPs/archive/DIP43.md

JinShil:
> Why was the WIP label added? I think it's possible to review even if some documentation is missing.

Documentation helps the review, and if, from the author's perspective, a PR is not complete enough to be merged, it is indeed a WIP.

jacob-carlborg:
> Documentation helps the review, and if, from the author's perspective, a PR is not complete enough to be merged, it is indeed a WIP.

Well, it depends. I think the code is ready as is. Documentation is missing yes, but keep in mind that most code in DMD was reviewed and merged without documentation. I mean, I don't see any documentation describing any of the C++ ABIs.

src/dmd/objc.d
```diff
@@ -265,3 +362,49 @@ extern(C++) private final class Supported : Objc
...
+                assert(baseCd.objc.metaclass.type.ty == Tclass);
+
+                auto metaBase = new BaseClass(baseCd.objc.metaclass.type);
+                metaBase.sym = baseCd.objc.metaclass;
+                metaBases.push(metaBase);
+            }
+            else
+            {
+                error("base " ~ errorType ~ " for an Objective-C " ~
+                      errorType ~ " must be extern (Objective-C)");
```
JinShil: tick marks around `extern (Objective-C)`, I believe.

jacob-carlborg: Yes, good catch.

jacob-carlborg: Fixed.

src/dmd/objc_glue.d
```diff
@@ -384,8 +475,21 @@ static:
...
-    Symbol* getMethVarRef(Identifier ident)
+    Symbol* getMethVarRef(const Identifier ident)
     {
         return getMethVarRef(ident.toString());
     }
 }
+
+private:
+
+char[] format(size_t bufLength, Args...)(return ref char[bufLength] buffer,
```
WalterBright: Ddoc comment

src/dmd/objc_glue.d
```diff
@@ -352,6 +391,58 @@ static:
...
+        if (symbol)
+            return symbol;
+
+        symbol = getGlobal(name);
+        stringValue.ptrvalue = symbol;
+
+        return symbol;
+    }
+
+    Symbol* getClassReference(const ClassDeclaration classDeclaration)
```
WalterBright: Ddoc comment

src/dmd/objc_glue.d
```diff
@@ -352,6 +391,58 @@ static:
         return moduleInfo;
     }
 
+    Symbol* getClassName(const ClassDeclaration classDeclaration)
```
WalterBright: Ddoc comment

src/dmd/objc_glue.d
```diff
@@ -249,6 +273,21 @@ static:
         return false;
     }
 
+    Symbol* symbolName(const(char)[] name, int sclass, type* t)
```
WalterBright: Ddoc comment for these three

src/dmd/objc.d
```diff
@@ -265,3 +362,49 @@ extern(C++) private final class Supported : Objc
         return _module && _module.isCoreModule(Id.attribute);
     }
 }
+
+private void setMetaclass(alias newMetaclass, T)(T classDeclaration, Scope* sc)
```
WalterBright: Ddoc comment

src/dmd/objc.d
```diff
@@ -131,6 +133,20 @@ Objc objc()
     return _objc;
 }
 
+
+/**
+ * This struct contains all data for a class declaration that is needed for the
```
WalterBright: remove `This struct contains`. It's redundant.

src/dmd/objc.d
```diff
@@ -131,6 +133,20 @@ Objc objc()
...
+ * Objective-C integration.
+ */
+struct ObjcClassDeclaration
+{
+    /// `true` if this class is a metaclass.
+    bool isMeta = false;
+
+    /// The metaclass of this class.
+    ClassDeclaration metaclass;
```
WalterBright: Why have `isMeta`? Isn't it locked to `metaclass !is null` ?

src/dmd/objc.d
jacob-carlborg: Ok.

src/dmd/objc_glue.d
jacob-carlborg: Yes, that's why the "Could use some more documentation" checkbox in the PR description is not checked 😃.

src/dmd/objc.d
jacob-carlborg: I'll investigate if `isMeta` can be removed.

jacob-carlborg:
Please see https://github.com/dlang/dmd/pull/7923 as well.

src/dmd/objc.d
jacob-carlborg: @WalterBright as far as I can see, `isMeta` is necessary. I need to be differentiate on when a class is a metaclass and when a class is a regular class but where the metaclass hasn't been set yet.

jacob-carlborg:
Added documentation.

jacob-carlborg:
Created a PR for the dlang.org part now as well: https://github.com/dlang/dlang.org/pull/2246.

JinShil (approved):

jacob-carlborg:
@JinShil more than 72 hours have passed.

PR #7763, opened 2018-01-23
highlight error messages:  misc. changes

JinShil:
Should be only a few subjective ones remaining after this.

I'll have at least one more PR after this to address the `.error` member functions that automatically prepend the output of `.toChars()` to the error message.

### DIFF
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -646,7 +646,7 @@ private extern (C++) final class StatementSemanticVisitor : Visitor
         else enum skipCheck = false;
         if (!skipCheck && (dim < 1 || dim > 2))
         {
-            fs.error("only one (value) or two (key,value) arguments for tuple foreach");
+            fs.error("only one (value) or two (key,value) arguments for tuple `foreach`");
             setError();
             return returnEarly();
         }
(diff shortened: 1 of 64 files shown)


src/dmd/statementsem.d
```diff
@@ -3795,7 +3795,7 @@ else
 
         if (!global.params.useExceptions)
         {
-            ts.error("Cannot use throw statements with -betterC");
+            ts.error("Cannot use `throw` statements with -betterC");
```
wilzbach: OT: Now that `betterC` is split up and the parameters might be exposed separately, this error message isn't ideal.

test/fail_compilation/betterc.d
```diff
@@ -1,7 +1,7 @@
 /* REQUIRED_ARGS: -betterC
  * TEST_OUTPUT:
 ---
-fail_compilation/betterc.d(11): Error: Cannot use throw statements with -betterC
+fail_compilation/betterc.d(11): Error: Cannot use `throw` statements with -betterC
 fail_compilation/betterc.d(16): Error: Cannot use try-catch statements with -betterC
```
wilzbach: No highlighting for `try-catch`?

wilzbach (approved):

src/dmd/statementsem.d
JinShil: Noted here: https://github.com/dlang/dmd/projects/3#card-6872539

wilzbach:
Jenkins seems to be loosing quite a lot of PRs lately :/

PR #7764, opened 2018-01-23
(WIP) Markdown support for DDoc

dgileadi:
This is a work in progress; please do not merge.


jacob-carlborg:
Would be nice with some examples with screenshots showing the Markdown and how the rendered result looks like.

src/dmd/doc.d
```diff
@@ -1844,6 +1856,53 @@ extern (C++) const(char)* skipwhitespace(const(char)* p)
     return p;
 }
 
+/************************************************
+ * Scan past the given characters.
+ */
```
WalterBright: Please use Ddoc style comments.

WalterBright:
> Support backtick-delimited code blocks  …

Please no.

WalterBright:
We're going to need a spec for all this.

src/dmd/doc.d
jacob-carlborg: It is a Ddoc comment, just no sections.

jacob-carlborg:
>> Support backtick-delimited code blocks …

> Please no.

If we don't, we shouldn't call it Markdown.

jacob-carlborg (changes requested):
There seems to be quite a few utility functions, please make them all private.

src/dmd/doc.d
dgileadi: I was attempting to match the comment style used by the rest of doc.d. I'd be happy to change it.

dgileadi:
As far as a spec, I've created [this project](https://github.com/dgileadi/ddoc-commonmark-spec) which generates a DDoc page from the CommonMark spec. I plan to add that page to the existing set of DDoc specs.

src/dmd/doc.d
```diff
@@ -1823,6 +1823,18 @@ extern (C++) bool isDitto(const(char)* comment)
...
+bool isWhitespace(const(char) c)
+{
+// TODO: unicode whitespace: Zs.
+    return c == ' ' || c == '\t' || c == '\r' || c == '\n';
+}
+
+bool isPunctuation(const(char) c)
+{
+// TODO: unicode punctuation: Pc, Pd, Pe, Pf, Pi, Po, or Ps.
+    return c == '!' || c == '"' || c == '#' || c == '$' || c == '%' || c == '&' || c == '\'' || c == '(' || c == ')' || c == '*' || c == '+' || c == ',' || c == '-' || c == '.' || c == '/' || c == ':' || c == ';' || c == '<' || c == '=' || c == '>' || c == '?' || c == '@' || c == '[' || c == '\\' || c == ']' || c == '^' || c == '_' || c == '`' || c == '{' || c == '|' || c == '}' || c == '~';
```
jacob-carlborg: Please break this line into multiple lines. Ideally a line should not exceed 80 columns.

src/dmd/doc.d
```diff
@@ -1844,6 +1856,53 @@ extern (C++) const(char)* skipwhitespace(const(char)* p)
     return p;
 }
 
+/************************************************
+ * Scan past the given characters.
+ */
+size_t skipchars(OutBuffer* buf, size_t i, string chars)
+{
+    const slice = buf.peekSlice();
+    for (; i < slice.length; i++)
```
jacob-carlborg: Prefer to use a `foreach`-loop.

src/dmd/doc.d
```diff
@@ -1948,6 +2007,201 @@ Lno:
     return i;
 }
 
+/****************************************************
+ * Replace the inline link portion (the part in parentheses) of a Markdown link.
+ */
+size_t replaceMarkdownInlineLink(OutBuffer* buf, size_t i, MarkdownDelimiter delimiter)
+{
+    if (buf.data[i+1] == '(')
```
jacob-carlborg: I suggest negating the condition and do an early return to save some indentation for the test of the code.

jacob-carlborg (changes requested):

jacob-carlborg:
I would like this PR to be split up into multiple smaller PRs. Ideally each of the commits in the PRs should have a meaning, i.e. avoid commits like "Fix bugs with quotes, lists and emphasis ".

src/dmd/doc.d
```diff
@@ -1844,6 +1856,53 @@ extern (C++) const(char)* skipwhitespace(const(char)* p)
...
+/************************************************
+ * Get the indent from one index to another, counting a tab as four spaces.
+ */
+size_t getIndent(OutBuffer* buf, size_t from, size_t to)
+{
+    const slice = buf.peekSlice();
+    if (to > slice.length)
+        to = slice.length;
+    size_t indent = 0;
+    for (; from < to; from++)
```
wilzbach: `foreach`

src/dmd/doc.d
```diff
@@ -1948,6 +2007,201 @@ Lno:
...
+}
+
+/****************************************************
+ * End a Markdown heading, if inside one.
+ */
+bool endMarkdownHeading(OutBuffer *buf, ref size_t i, ref int headingLevel, size_t iHeadingStart)
+{
+    if (headingLevel)
+    {
+        char* heading = cast(char*) "$(H0 ".dup;
```
wilzbach: `.ptr` will allow you to avoid the cast

src/dmd/doc.d
```diff
@@ -2145,20 +2399,148 @@ extern (C++) bool isReservedName(const(char)* str, size_t len)
...
+* A delimiter for Markdown inline content like emphasis and links.
+*/
+struct MarkdownDelimiter
+{
+    size_t iStart;
+    char type;
+    int count;
+    int macroLevel;
+    bool leftFlanking;
+    bool rightFlanking;
```
wilzbach: Ideally all members of a public date type should have a short Ddoc description.

src/dmd/doc.d
```diff
@@ -1844,6 +1856,53 @@ extern (C++) const(char)* skipwhitespace(const(char)* p)
     return p;
 }
 
+/************************************************
+ * Scan past the given characters.
+ */
+size_t skipchars(OutBuffer* buf, size_t i, string chars)
```
wilzbach: - use CamelCase: `skipChars`
- why not `ref OutBuffer`?

dgileadi:
FYI I won't be working on this for a few days—I'm helping with my son's first pinewood derby.

src/dmd/doc.d
WalterBright: It's missing `Params:` and `Returns:` sections.

src/dmd/doc.d
```diff
@@ -1844,6 +1856,53 @@ extern (C++) const(char)* skipwhitespace(const(char)* p)
...
+            }
+        }
+        if (!contains)
+            break;
+    }
+    return i;
+}
+
+/************************************************
+ * Get the indent from one index to another, counting a tab as four spaces.
```
WalterBright: Type a text file, tabs are 8. Print a text file, tabs are 8. Counting tabs as 4 doesn't work. I recommend treating tabs as 8, and for anyone who complains, suggest they use spaces. Note that because of this difficulty, no tabs are allowed in dmd source code.

Also, what do you mean by "indent"? The number of columns?

src/dmd/doc.d
```diff
@@ -1844,6 +1856,53 @@ extern (C++) const(char)* skipwhitespace(const(char)* p)
...
+ */
+size_t getIndent(OutBuffer* buf, size_t from, size_t to)
+{
+    const slice = buf.peekSlice();
+    if (to > slice.length)
+        to = slice.length;
+    size_t indent = 0;
```
WalterBright: one wonders if the indent could ever be larger than 4 gigs :-)

src/dmd/doc.d
WalterBright: or make it an array. Then line 2192 is safe, and use .ptr in the only other use, line 2193.

WalterBright: On further review, there is no need for `heading` to be allocated on the heap. It's a short, fixed size. Make it a static array on the stack.

src/dmd/doc.d
WalterBright: Also reverse sort the fields by size.

wilzbach:
@dgileadi I see that you are actively working on this - did you already try to separate your changes into smaller chunks (PRs)? Then we could move incrementally and merge/review it step by step.

dgileadi:
Not yet, but I do plan to do that. Right now I’m working on getting DDoc’s Markdown to better match the spec. I feel like I may have opened this PR too soon, although it did result in valuable feedback.

wilzbach:
> I feel like I may have opened this PR too soon,

I was just seeing you work on this & asked whether we can already help ;-)

dgileadi:
And I certainly appreciate it, don’t mistake me!

src/dmd/doc.d
```diff
@@ -1823,6 +1833,32 @@ extern (C++) bool isDitto(const(char)* comment)
...
+/*****************************************
+ * Return true if the given character is punctuation.
+ */
+private bool isPunctuation(const(char) c)
+{
+// TODO: unicode punctuation: Pc, Pd, Pe, Pf, Pi, Po, or Ps.
+// However, unicode punctuation should not be included for Markdown backslash
+// escapes, so when it's implemented be sure to make the unicode punctuation
+// check optional via a parameter
+    return c == '!' || c == '"' || c == '#' || c == '$' || c == '%' || c == '&'
```
BBasile: Use a static array of booleans, indexed by char (as ubyte), for O(1) lookups:

```d
static immutable bool[char.max] punctLUT= ['!' : true, ',': true, /*etc*/ ];
```

Only `true` members needs to be initialized. Refactoring for unicode, if necessary will be easy. (if ... > 0x80)...
Also these are rather called symbols but let's  not bikeshed for now.

dgileadi:
Thanks, I suspect you’re right that the O(1) lookup is worth the small memory hit.

src/dmd/doc.d
BBasile: I think that there's been a slight misunderstanding: you finally use an associative array but i proposed a a simple static immutable array, meaning that the data don't need to be initialized.
Look at the difference here: https://godbolt.org/g/VMrxDj

dgileadi:
That makes more sense; the associative array seemed strange. Thanks for taking the time to suggest the change and for correcting me when I got it wrong!

dgileadi:
Closing in favor of individual PRs.

PR #7765, opened 2018-01-23
Remove useless runtime code

JinShil:
The code being removed in this PR is already in druntime (object.d).  I don't see any reason for it to be in the compiler, and it's interfering with some no-runtime code I'm working on.  

~Let's see what the test suite says.~

Interesting!  Everything passes.  I'm shocked, disappointed, relieved and happy all at the same time.  I'm forced to assume this is just left-over junk.  I wonder how much more of this is hiding in the code.


wilzbach (approved):
For reference, it's here: https://github.com/dlang/druntime/blob/cf41932a8e1fa6ff01a3c96cbaae8742b716ab91/src/object.d#L3369

PetarKirov (approved):

PR #7766, opened 2018-01-23
Fix Issue 18243 - selective import + overload = private visibility

RazvanN7:
I fixed the issue by modifying the overloadApply method. The problem was that when resolving overloads the visibility attribute is not checked at all, so I had to add the scope as an optional parameter so that when an alias declaration is encountered (from imports or plain alias declarations) a check is made to make sure that the called function is indeed visible.

### DIFF
--- a/test/fail_compilation/imports/a18243.d
+++ b/test/fail_compilation/imports/a18243.d
@@ -1,0 +1,5 @@
+module a18243;
+
+import std.math : isNaN;
+
+public bool isNaN() { return false; }
--- a/src/dmd/access.d
+++ b/src/dmd/access.d
@@ -512,6 +512,20 @@ extern (C++) bool symbolIsVisible(Dsymbol origin, Dsymbol s)
 extern (C++) bool symbolIsVisible(Scope *sc, Dsymbol s)
 {
     s = mostVisibleOverload(s);
+    return checkSymbolAccess(sc, s);
+}
+
+/**
+ * Check if a symbol is visible from a given scope without taking
+ * into account the most visible overload.
+ *
+ * Params:
+ *  sc = lookup scope
+ *  s = symbol to check for visibility
+ * Returns: true if s is visible by origin
+ */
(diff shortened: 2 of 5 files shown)


test/fail_compilation/imports/a18243.d
```diff
@@ -0,0 +1,5 @@
+module a18243;
+
+import std.math : isNaN;
```
mathias-lang-sociomantic: Selective imports creating`alias` in the current scope is really an annoyance, and violates the principle of least surprise. I wish we could change that.

RazvanN7: That is a small project on its own which will propagate cascade changes throughout a large number of components.

RazvanN7: Although I agree that the current state is hackish and messy.

mathias-lang-sociomantic: Last time I checked with Walter, it was by design. Maybe that changed since. If you have any way to find out, it would be appreciated.

wilzbach:
@RazvanN7 It looks like you need to fix a couple of projects. However, they are all in `dlang-comunity` or `dlang-tour`, so I have merge rights & thus I will be able to help you ;-)

wilzbach:
@RazvanN7 making the PRs isn't too hard:

Fork and clone the respective repo

```
git checkout fix-imports-for-dmd-7766
dub test --compiler=/home/seb/dlang/dmd/generated/linux/release/64/dmd 
# fix the error & run until it passes
git add -u
git commit "Fix imports for dmd 7766"
pr_send
```

[`pr_send`](https://github.com/wilzbach/git-tools/blob/master/pr_send)

RazvanN7:
The jenkins failure is unrelated to this PR. Can we move forward with this? ping @andralex

stefan-koch-sociomantic:
Is this in the changelog ?

jacob-carlborg:
> Is this in the changelog ?

Not in this PR.

PR #7769, opened 2018-01-24
[trivial] Improve error message when aggregate template is used as a type

ntrel:
Make it clear with aggregate templates that they are templates and haven't been instantiated.

Note: `s.kind` is e.g. just "struct" for aggregate templates. Non-aggregates say "template".
Note: When `td.onemember` is an alias, we don't know if it resolves to a type or not until the template is instantiated.

http://forum.dlang.org/post/p49c55$m65$1@digitalmars.com

### DIFF
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -1156,7 +1156,12 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
         {
             if (s)
             {
-                s.error(loc, "is used as a type");
+                auto td = s.isTemplateDeclaration;
+                if (td && td.onemember && td.onemember.isAggregateDeclaration)
+                    mtype.error(loc, "%s template `%s` is used as a type without instantiation",
+                        s.kind, s.toPrettyChars);
+                else
+                    mtype.error(loc, "%s `%s` is used as a type", s.kind, s.toPrettyChars);
                 //assert(0);
             }
             else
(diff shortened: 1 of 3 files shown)


quickfur (commented):
LGTM.

jacob-carlborg (approved):

wilzbach:
CircleCi failure is unrelated - https://github.com/dlang/dmd/pull/7770

src/dmd/typesem.d
```diff
@@ -1156,7 +1156,12 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
         {
             if (s)
             {
-                s.error(loc, "is used as a type");
+                auto td = s.isTemplateDeclaration;
+                if (td && td.onemember && td.onemember.isAggregateDeclaration)
+                    mtype.error(loc, "%s template `%s` is used as a type without instantiation",
```
WalterBright: I suggest an improvement:
```
template %s `%s` is used as a type without instantiation; to instantiate it use `%s!(arguments)`
```

wilzbach:
> CircleCi failure is unrelated - #7770 

In lack of someone wanting merging #7770, I rebased the PR. This will fix the CircleCi failure too.

WalterBright (changes requested):
Please make the change I suggested. A good error message should say what is wrong, and suggest corrective action.

src/dmd/typesem.d
MetaLang: What if the user has a variable called `arguments` already in scope? They might get confused.

WalterBright (approved):

src/dmd/typesem.d
ntrel: @MetaLang Maybe *use `%s!()`* instead? That way if the template requires arguments they will get an error to that effect, and this error may come up when the template doesn't need arguments anyway, as in the forum example.

PR #7771, opened 2018-01-24
Fix Issue 18295 -  `scope class` check too conservative under -dip1000

JinShil:
See issue for the details.  This PR also adds coverage for previously uncovered case.

FYI: This is in direct support of https://github.com/dlang/dlang.org/pull/2027

### DIFF
--- a/test/compilable/test18295.d
+++ b/test/compilable/test18295.d
@@ -1,0 +1,19 @@
+// REQUIRED_ARGS: -dip1000
+
+// https://issues.dlang.org/show_bug.cgi?id=18295
+
+// See test/fail_compilation/test18295.d for the non-`-dip1000` version`
+
+scope class C { int i; }    // Notice the use of `scope` here
+
+C increment(scope return C c) @safe  // Prior to the fix for 18295 an error would be emitted here
+{                                    // which is too conservative
+    c.i++;
+    return c;
+}
+
+void main() @safe
+{
+    scope C c = new C();
+    c.increment();
+}
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -795,10 +795,13 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
             tf.next = tf.next.typeSemantic(loc, sc);
             sc = sc.pop();
             errors |= tf.checkRetType(loc);
-            if (tf.next.isscope() && !(sc.flags & SCOPE.ctor))
+            if (!global.params.vsafe) // https://issues.dlang.org/show_bug.cgi?id=18295
             {
-                mtype.error(loc, "functions cannot return `scope %s`", tf.next.toChars());
-                errors = true;
+                if (tf.next.isscope() && !(sc.flags & SCOPE.ctor))
+                {
+                    mtype.error(loc, "functions cannot return `scope %s`", tf.next.toChars());
+                    errors = true;
+                }
             }
             if (tf.next.hasWild())
                 wildreturn = true;


test/compilable/test18295.d
```diff
@@ -0,0 +1,19 @@
+// REQUIRED_ARGS: -dip1000
+
+// https://issues.dlang.org/show_bug.cgi?id=18295
+
+// See test/fail_compilation/test18295.d for the non-`-dip1000` version`
+
+scope class C { int i; }    // Notice the use of `scope` here
+
+C increment(C c)            // Should not emit error here
```
MartinNowak: As this is a function there is no inferrence, and if the `scope` parameter is not `return`, it cannot be returned.

MartinNowak: Yes it should error, you need `return scope` on the parameter for this to pass.

test/compilable/test18295.d
```diff
@@ -0,0 +1,19 @@
...
+{
+    c.i++;
+    return c;
+}
+
+void main()
+{
+    scope C c = new C();
+    c.increment();
```
MartinNowak: Should error here, cannot pass scope to non-scope parameter `c`.
```
bug.d(10): Error: scope variable `c` assigned to non-scope parameter `c` calling bug.increment
```

test/compilable/test18295.d
```diff
@@ -0,0 +1,19 @@
+// REQUIRED_ARGS: -dip1000
+
+// https://issues.dlang.org/show_bug.cgi?id=18295
+
+// See test/fail_compilation/test18295.d for the non-`-dip1000` version`
+
```
MartinNowak: Without `@safe:` there are no scope checks.

test/fail_compilation/test18295.d
```diff
@@ -0,0 +1,24 @@
...
+C increment(C c)            // Not compiling under -dip1000, so this should emit an error
+{
+    c.i++;
+    return c;
+}
+
+void main()
+{
+    scope C c = new C();
+    c.increment();
```
MartinNowak: That simply won't catch the escaping of a scope variable, checks are only enabled with `-dip1000`.

MartinNowak (commented):
Right, all `scope class` means is enforcing the usage of
`scope var = new Clazz`, it has no implications on the scope checks besides that.
In particular you still need to annotate parameters as `return scope`. Remember, scope is a storage class, not a type qualifier, so it doesn't simply carry to every usage of that type.
I'm inclined to say that forcing a particular storage on users of a class is a weird use-case and unprecedented for other storage options (GC, RC). Maybe we should deprecate `scope class C {}` declarations in the long-run? Not sure why this mechanism was introduced? To catch unwanted GC allocations?

wilzbach:
>  Maybe we should deprecate scope class C {} declarations in the long-run? Not sure why this mechanism was introduced? To catch unwanted GC allocations?

FYI: ` scope A a = new A(1);` is already deprecated:

https://dlang.org/deprecate.html#scope%20for%20allocating%20classes%20on%20the%20stack

JinShil:
> Maybe we should deprecate scope class C {} declarations in the long-run?
> FYI: scope A a = new A(1); is already deprecated:

The de-deprecation has been approved.  See dlang/dlang.org#2027

> Remember, scope is a storage class, not a type qualifier, so it doesn't simply carry to every usage of that type.

When it's attributed to the class declaration, it serves as a type qualifier and does carry to every usage of that type, which is why the following emits an error:

```D
scope class A { }

void main()
{
    A a = new A();   // Error: variable main.a reference to scope class must be scope
}
```

wilzbach:
Jenkins failure is unrelated and should be gone soon (-> https://github.com/dlang/ci/pull/132)

test/compilable/test18295.d
JinShil: Added `@safe` attributes

test/compilable/test18295.d
JinShil: Added `return scope` to the parameter

test/compilable/test18295.d
JinShil: Added `return scope` to the parameter so there should be no error.

test/fail_compilation/test18295.d
JinShil: That's not the point of this test.  The point is to test the current behavior for which there currently is no test.  See https://codecov.io/gh/dlang/dmd/src/master/src/dmd/typesem.d#L800...801  If you believe that is an error, please file a bug in bugzilla and submit a fix in a separate PR.

test/compilable/test18295.d
JinShil: Added `return scope` to the parameter, there should be no error.

JinShil:
> I'm inclined to say that forcing a particular storage on users of a class is a weird use-case and unprecedented for other storage options (GC, RC).  Not sure why this mechanism was introduced?

I find it to be rather useful as a way of enforcing that all instance of the class are stack allocated.  It's consistent with constraints such as `final` in that it places restrictions on how the class is employed.

MartinNowak:
Well, final allows you to restrict any compatibility-relevant usage.
What's use-case for restricting heap allocation of classes?

```d
/+
dub.sdl:
dependency "automem" version="~>0.0.10"
+/
import automem, std.typecons : scoped;

scope class C {}

void main()
{
    auto c1 = scoped!C();
    auto c2 = RefCounted!C();
}
```
If there was a clear need for local only classes, that restriction might be warranted, I just don't see any.

test/fail_compilation/test18295.d
```diff
@@ -0,0 +1,24 @@
...
+---
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=18295
+
+// See test/compilable/test18295.d for the `-dip1000` version`
+
+scope class C { int i; }    // Notice the use of `scope` here
+
+C increment(C c)            // Not compiling under -dip1000, so this should emit an error
```
MartinNowak: The comment seems to be wrong.

JinShil: Right or wrong, that's the way it currently works.  All this test is doing is adding coverage to the existing implementation.  The `compilable/test18295.d` file is the test for the new functionality in this PR.

test/compilable/test18295.d
```diff
@@ -0,0 +1,19 @@
...
+scope class C { int i; }    // Notice the use of `scope` here
```
MartinNowak: `scope class` is irrelevant for this test, please remove the misleading comment.
Only the fact that `c` is a scoped class reference matters for the return scope check.

test/compilable/test18295.d
```diff
@@ -0,0 +1,19 @@
...
+
+C increment(return scope C c) @safe      // Should not emit error here
```
MartinNowak: Ideally we'd want to add this `-dip1000` test as well.

```d
scope class C { int i; }

C increment(return ref C c) @safe // check implicit scope from class type
{
    c.i++;
    return c;
}

void main() @safe
{
    scope C c = new C();
    c.increment();
}
```

Doesn't work though cause `scope class` isn't a real class reference.
```
bug.d(3): Error: variable bug.increment.c globals, statics, fields, manifest constants, ref and out parameters cannot be `scope`
```

MartinNowak (commented):
Looks OKish. Though I don't think anyone wants to define those half-implemented `scope class`es and their interaction with DIP1000 while we're working on `@safe` manual memory management and the whole topic is still in flux.

test/fail_compilation/test18295.d
JinShil: This test has been merged in a separate PR (https://github.com/dlang/dmd/pull/7791), so we can get off this tangent.

test/compilable/test18295.d
JinShil: Then I suggest merging this and filing a bug.

JinShil:
>  What's use-case for restricting heap allocation of classes?

That's a good question for the folks at Sociomantic who have clearly found a use for it in [Ocean](https://github.com/sociomantic-tsunami/ocean/search?p=1&q=%22scope+class%22&type=&utf8=%E2%9C%93). @mathias-lang-sociomantic?

mihails-strasuns-sociomantic:
(answering @mathias-lang-sociomantic summons)

@JinShil it is a legacy of D1 which doesn't have struct constructors/destructors, which means `scope class` was the only form of non-trivial stack-allocated aggregates available. Because we have to keep compatibility with both language versions it was not possible to switch to structs.

In the long term there is nothing preventing deprecation of scope classes and switching to structs instead, but that is both non-trivial and somewhat low priority (there are more important problems with D2 migration that are still unresolved).

Important to note that we only care about actual `scope instance = new Thing` allocation - none of dip1000 constraints will matter because we never use `@safe` which enables them.

JinShil:
> In the long term there is nothing preventing deprecation of scope classes and switching to structs instead

So be it.  Then I leave it to @WalterBright and @andralex  to decide the fate of the `scope class` in dlang/dlang.org#2027

JinShil:
@WalterBright  has made his view known at https://github.com/dlang/dlang.org/pull/2027#issuecomment-374493221

>1. Deprecate `scope class C { }`. It's kind of a bizarre, neglected quirk. It forces all allocations of `C` to go on the stack. It's hard to see what the point of it is these days.
>2. Retain `scope C c = new C();`. The `C` gets allocated on the stack instead of the heap, but pedantically this is an optimization. The -dip1000 checks ensure it works in that `c` is not allowed to escape the lifetime of its stack frame.

Therefore, instead of fixing this issue (Issue 18295), I'll move ahead deprecating `scope`-as-a-type-modifier, which will eventually make the issue irrelevant.  Closing.

PR #7772, opened 2018-01-24
fix Issue 18296 - make __coverage a hidden symbol

MartinNowak:
- external linkage only for current DSO, but not across DSOs
- avoids GOT indirection on every access
- thus avoids codegen bugs related to those
- uses SCstatic with SFLhidden flag as the access is the same as for
  static symbols, even though linkage is between SCglobal and SCstatic
  In the long-run (and with more usage) a separate SChidden might be cleaner.
- no changes to OMF code as LPUBDEF isn't used

### DIFF
--- a/src/dmd/backend/cc.d
+++ b/src/dmd/backend/cc.d
@@ -1185,6 +1185,7 @@ enum
     SFLnodebug      = 0x20000,     // don't generate debug info
     SFLwasstatic    = 0x800000,    // was an uninitialized static
     SFLweak         = 0x1000000,   // resolve to NULL if not found
+    SFLhidden       = 0x2000000,   // not visible outside of DSOs (-fvisibility=hidden)
     SFLartifical    = 0x4000000,   // compiler generated symbol
 
     // CPP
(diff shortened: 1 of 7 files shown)


rainers:
> Still needs equivalent fixes for Mach-O and COFF/OMF.

I haven't been able to reproduce on Windows, probably because there is no -fPIC there.

MartinNowak:
> I haven't been able to reproduce on Windows, probably because there is no -fPIC there.

Yes, linkage on Win is a very different topic.

MartinNowak:
This is useful work anyhow, as we want to make non-exported symbols hidden by default in the long-run. Still hanging on the unfinished export work/fixes though.

ibuclaw:
> This is useful work anyhow, as we want to make non-exported symbols hidden by default in the long-run.

Yes, but such a change may end up being the birth of S3.

MartinNowak:
Iain speaking in riddles again. If we can leverage explicit exports in D to gain a speed-up over C/C++ by default, then we should take the opportunity.

rainers:
> no changes to OMF code as LPUBDEF isn't used

It seems you still need to update cgobj.c, though, to fix the Win32 failures.

ibuclaw:
> Iain speaking in riddles again. If we can leverage explicit exports in D to gain a speed-up over C/C++ by default, then we should take the opportunity.

Whoops, looks like phone auto correct got the better of me. I corrected it (should have been D3). 

What I meant of course was that such a change may be disruptive enough to warrant a new fork in the language similar to what const did to warrant D2.

MartinNowak:
> What I meant of course was that such a change may be disruptive enough to warrant a new fork in the language similar to what const did to warrant D2.

About a small linkage behavior change? In a mostly statically linked language (which isn't affected by visibility). Seems very unreasonable.

MartinNowak:
> It seems you still need to update cgobj.c, though, to fix the Win32 failures.

There are no Win32 errors (any more), can we merge this?

PR #7774, opened 2018-01-25
show -fPIC on all relevant platforms

MartinNowak:
- it's always enabled on OSX, so no need to add the flag there


MartinNowak:
Interesting, if you edit the labels while the bot removes some, you'll readd them on submit. Weird conflict resolution by github, but anyhow.

JinShil (approved):

PR #7777, opened 2018-01-25
Add ddoc for -i

marler8997:


src/dmd/cli.d
```diff
@@ -314,7 +314,42 @@ dmd -cov -unittest myprog.d
             "same as -i=-std,-core,-etc,-object"
         ),
         Option("i=[-]<pattern>,[-]<pattern>,...",
-            "include/exclude imported modules whose name matches one of <pattern>"
+            "include/exclude imported modules whose name matches one of <pattern>",
+            `this option enables "include import" mode, where the compiler will
```
wilzbach: While the dlang.org script will capitalize the first word automatically, I think it's still good to do here too (to avoid confusion etc.).

src/dmd/cli.d
```diff
@@ -314,7 +314,42 @@ dmd -cov -unittest myprog.d
...
+             include imported modules in the compilation, as if they were given
+             on the command line.
+
+             Determining which modules to include is based on a set of "module patterns".
+             These patterns are optionally included after the "-i=" option as a comma
+             separated list.  A pattern matches any module whose fully qualified name
```
wilzbach: Double whitespace

src/dmd/cli.d
```diff
@@ -314,7 +314,42 @@ dmd -cov -unittest myprog.d
...
+             starts with the pattern. The pattern "foo.bar" matches all modules in the
+             "foo.bar" package, such as "foo.bar" itself, "foo.bar.baz" or even
+             "foo.bar.package".  Note that matching is based on the full component name so
+             "foo.bar" would not match a module named "foo.barx".
+
+             The presence of the "-" prefix in a module pattern indicates that matching-modules
+             should be excluded from compilation.  Patterns without this indicate that
```
wilzbach: double ws

src/dmd/cli.d
```diff
@@ -314,7 +314,42 @@ dmd -cov -unittest myprog.d
...
+
+             The default behavior to include modules or exclude them is based on whether the user
+             has provided any "inclusionary" module patterns.  If no patterns are given, or they
+             are all exclusionary, then modules that don't match any pattern will be included
+             by default.  However, if at least one "inclusionary" module pattern is given, then
+             modules that don't match any pattern will be excluded by default.  Should the user
+             want to override this behavior, they can use the special "." pattern to "include by
+             default" or "-." to "exclude by default".
+
+             Note: multiple "-i=..." options is allowed, each one just adds more patterns.`
```
wilzbach: are

wilzbach: Avoid "just" as it colloquial.

src/dmd/cli.d
```diff
@@ -314,7 +314,42 @@ dmd -cov -unittest myprog.d
...
         ),
         Option("i=[-]<pattern>,[-]<pattern>,...",
-            "include/exclude imported modules whose name matches one of <pattern>"
+            "include/exclude imported modules whose name matches one of <pattern>",
+            `this option enables "include import" mode, where the compiler will
+             include imported modules in the compilation, as if they were given
+             on the command line.
+
+             Determining which modules to include is based on a set of "module patterns".
+             These patterns are optionally included after the "-i=" option as a comma
```
wilzbach: How about using a [delimited string](https://dlang.org/spec/lex.html#delimited_strings) and use back-ticks for command line things like `-i=`?

src/dmd/cli.d
```diff
@@ -314,7 +314,42 @@ dmd -cov -unittest myprog.d
...
+             "foo.bar" would not match a module named "foo.barx".
+
+             The presence of the "-" prefix in a module pattern indicates that matching-modules
+             should be excluded from compilation.  Patterns without this indicate that
+             matching-modules should be included in compilation.
+
+             Note that along with the module patterns given by the user, there is a standard
+             set of exclusionary module patterns that are always used, namely:
+                "-std,-core,-etc,-object"
+             Note that these can be overriden (i.e. -i=std,core,etc,object).
```
wilzbach: Double "Note", maybe "However" ?

src/dmd/cli.d
```diff
@@ -314,7 +314,42 @@ dmd -cov -unittest myprog.d
...
+
+             Since module patterns can either be inclusionary or exclusionary, a priority
+             must be established to handle the case where a module matches multiple patterns.
```
wilzbach: I think typically the motivation for a feature isn't stated in its CLI docs.
@ others agree or disagree?

src/dmd/cli.d
```diff
@@ -314,7 +314,42 @@ dmd -cov -unittest myprog.d
...
+             starts with the pattern. The pattern "foo.bar" matches all modules in the
+             "foo.bar" package, such as "foo.bar" itself, "foo.bar.baz" or even
+             "foo.bar.package".  Note that matching is based on the full component name so
+             "foo.bar" would not match a module named "foo.barx".
+
+             The presence of the "-" prefix in a module pattern indicates that matching-modules
+             should be excluded from compilation.  Patterns without this indicate that
+             matching-modules should be included in compilation.
+
+             Note that along with the module patterns given by the user, there is a standard
```
wilzbach: The style of addressing the user as "the user" is completely new and doesn't appear anywhere at https://dlang.org/dmd.html

I think sth. like this would fit better with the existing style:

> The standard set of exclusionary module patterns is "-std,-core,-etc,-object" and can be overridden with i.e. -i=std,core,etc,object).

src/dmd/cli.d
```diff
@@ -314,7 +314,42 @@ dmd -cov -unittest myprog.d
...
+             A pattern's priority is determined by it's component length, where more components
+             means higher priority, i.e. foo.bar.baz is a higher priority match than foo.bar.
+
+             The default behavior to include modules or exclude them is based on whether the user
+             has provided any "inclusionary" module patterns.  If no patterns are given, or they
+             are all exclusionary, then modules that don't match any pattern will be included
+             by default.  However, if at least one "inclusionary" module pattern is given, then
+             modules that don't match any pattern will be excluded by default.  Should the user
+             want to override this behavior, they can use the special "." pattern to "include by
+             default" or "-." to "exclude by default".
```
wilzbach: This paragraph is a bit too complicated for what it tries to tell. How about:

> By default modules that don't match any pattern will be included.
> However, if at least one inclusionary module pattern is given, then modules that don't match any pattern will be excluded.
> This can be overridden by `-i=-.` (exclude by default) or `-i=.` (include by default).

wilzbach:
Oh and congrats to submitting 7777 :four_leaf_clover:  :1st_place_medal:

src/dmd/cli.d
```diff
@@ -314,7 +314,37 @@ dmd -cov -unittest myprog.d
             "same as -i=-std,-core,-etc,-object"
         ),
         Option("i=[-]<pattern>,[-]<pattern>,...",
-            "include/exclude imported modules whose name matches one of <pattern>"
+            "include/exclude imported modules whose name matches one of <pattern>",
+            q"{$(P Enables "include imports" mode, where the compiler will include imported
+	     modules in the compilation, as if they were given on the command line.)
```
timotheecour: > as if they were given on the command line

there seems to be subtle differences: are these a feature or a bug? should they be documented here?
```
main.d:
import foo.fun;
foo/fun.d:
```
dmd -i -o- main.d
#ok
dmd -o- main.d foo/fun.d
Error: module fun from file foo/fun.d must be imported with 'import fun;'

NOTE: if we add `module foo.fun;` to foo/fun.d the problem goes away
NOTE: if we add `module asdf.fun;` to foo/fun.d, `dmd -i` accepts the code but should not. This one is a clear bug.

src/dmd/cli.d
```diff
@@ -314,7 +314,37 @@ dmd -cov -unittest myprog.d
...
+
```
timotheecour: another subtle difference, again not clear if feature or bug but should be documented:

```
dmd -od=generated -op -c -Ifoo foo/fun1.d foo/bar/fun4.d foo/main.d
dmd -od=generated_i -op -c -i -Ifoo foo/main.d
```

```
foo/main.d: void main(){import bar.fun4; import fun1;}
foo/fun1.d:
foo/bar/fun4.d:module bar.fun4;
```

```
├── generated
│   └── foo
│       ├── bar
│       │   └── fun4.o
│       ├── fun1.o
│       └── main.o
├── generated_i
│   ├── bar
│   │   └── fun4.o
│   ├── foo
│   │   └── main.o
│   └── fun1.o

```

timotheecour (commented):
bugs or features? need doc

src/dmd/cli.d
marler8997: yeah looks like a bug, I"ll see if I can fix this one

src/dmd/cli.d
marler8997: bug

marler8997: Fix here: https://github.com/dlang/dmd/pull/7778

src/dmd/cli.d
```diff
@@ -314,7 +314,37 @@ dmd -cov -unittest myprog.d
...
-            "include/exclude imported modules whose name matches one of <pattern>"
+            "include/exclude imported modules whose name matches one of <pattern>",
+            q"{$(P Enables "include imports" mode, where the compiler will include imported
+             modules in the compilation, as if they were given on the command line.)
+
+             $(P Determining which modules to include is based on a set of "module patterns".
+             These patterns are optionally included after `-i=` as a comma separated list.
+             A pattern matches any module whose fully qualified name starts with the pattern.
+             The pattern "foo.bar" matches all modules in the `foo.bar` package, such as
+             "foo.bar" itself, "foo.bar.baz" or even "foo.bar.package". Note that each component
```
wilzbach: As you escape package names below, I would do it here too for the sake of consistency.

src/dmd/cli.d
```diff
@@ -314,7 +314,37 @@ dmd -cov -unittest myprog.d
...
+             of the fully qualified name must match completely, so the pattern "foo.bar" would
+             not match a module named `foo.barx`.)
+
+             $(P The presence of the "-" prefix in a module pattern indicates that matching-modules
+             should be excluded from compilation. Patterns without this indicate that
+             matching-modules should be included in compilation.)
```
wilzbach: Why so you use a hyphen here?

src/dmd/cli.d
```diff
@@ -314,7 +314,37 @@ dmd -cov -unittest myprog.d
...
+
+             $(P Along with the module patterns supplied, there is a standard set of exclusionary
+             patterns that are always included:
+                "-std,-core,-etc,-object"
+             Note that these can be overriden (i.e. -i=std,core,etc,object).)
```
wilzbach: For consistency: use backticks

src/dmd/cli.d
```diff
@@ -314,7 +314,37 @@ dmd -cov -unittest myprog.d
...
+             of the fully qualified name must match completely, so the pattern "foo.bar" would
+             not match a module named `foo.barx`.)
+
+             $(P The presence of the "-" prefix in a module pattern indicates that matching-modules
+             should be excluded from compilation. Patterns without this indicate that
+             matching-modules should be included in compilation.)
+
+             $(P Along with the module patterns supplied, there is a standard set of exclusionary
+             patterns that are always included:
+                "-std,-core,-etc,-object"
```
wilzbach: You use backticks for CLI arguments above.

wilzbach (commented):
There is still a bit of an inconsistency between backticks and quotes which I assume to look a bit weird on the rendered docs, apart from that it looks really good. Thanks!

src/dmd/cli.d
marler8997: I was attempting to escape package names, but not module patterns...there are other places in the help where module patterns are quoted, but I thought that package names themselves should definitely be escaped.  looking at the final generated docs will likely indicate which method looks the best.

JinShil:
@wilzbach Why isn't this showing up in DAutoTest's rendered docs?

marler8997:
I think we're waiting on this: https://github.com/dlang/dlang.org/pull/2068

wilzbach:
As https://github.com/dlang/dlang.org/pull/2068 is in, I rebased this PR to restart DAutoTest.

marler8997:
Hmmm, the `-i=[-]<pattern>,[-]<pattern,..` doesn't seem to show up correctly.  Maybe commas are messing it up?

marler8997:
Yeah I think it's the comma.  I think the dlang.org generator needs to be modified to handle commas in the `flag` field.

marler8997:
Looks like https://github.com/dlang/dlang.org/pull/2188 did the trick.  This PR should be ready for merge.

JinShil (approved):

src/dmd/cli.d
```diff
@@ -314,7 +314,37 @@ dmd -cov -unittest myprog.d
...
+             not match a module named `foo.barx`.)
+
+             $(P The presence of the `-` prefix in a module pattern indicates that matching modules
+             should be excluded from compilation. Patterns without this indicate that
+             matching-modules should be included in compilation.)
+
+             $(P Along with the module patterns supplied, there is a standard set of exclusionary
+             patterns that are always included:
+                `-std,-core,-etc,-object`
+             Note that these can be overriden (i.e. `-i=std,core,etc,object`).)
```
wilzbach: I think this needs a new line or at least a dot:
image
Also you use "Note:" later (with colon) - consistency and all ;-)

wilzbach (approved):
LGTM module nits.

marler8997:
Wait for https://github.com/dlang/dmd/pull/7857 before commit this one.  If that one gets accepted then it will change this one a bit.

marler8997:
Walter has merged #7857 so I've reworded this one a bit...

src/dmd/cli.d
```diff
@@ -310,11 +310,32 @@ dmd -cov -unittest myprog.d
...
+             
+             $(P When a module matches multiple patterns, matches are prioritized by their component length, where
+             a match with more components takes priority (i.e. pattern `foo.bar.baz` has priority over `foo.bar`).)
+
+             $(P By default modules that don't match any pattern will be included. However, if at
+             least one inclusive pattern is given, then modules not matching any pattern will
+             be excluded. This behavior can be overriden by usig `-i=.` to include by default or `-i=-.` to
+             exclude by default.)
+
+             $(P Note that multiple `-i=...` options are allowed, each one adds more patterns.)}"
```
wilzbach: How about: each one adds a pattern?

Though, I think now you can even remove this last paragraph completely as you have a few examples of multiple arguments in the paragraphs before.

src/dmd/cli.d
```diff
@@ -310,11 +310,32 @@ dmd -cov -unittest myprog.d
...
+             of the form `-i=-<package>` is an "exclusive pattern". Inclusive patterns will include
+             all module's whose names match the pattern, whereas exclusive patterns will exclude them.
+             For example. all modules in the package `foo.bar` can be included using `-i=foo.bar` or excluded
+             using `-i=-foo.bar`. Note that each component of the fully qualified name must match the
+             pattern completely, so the pattern `foo.bar` would not match a module named `foo.barx`.)
+
+             $(P The default behavior of excluding druntime/phobos is accomplished by internally adding a
+             set of standard exclusions, namely, `-i=-std -i=-core -i=-etc -i=-object`. Note that these
+             can be overriden with `-i=std -i=core -i=etc -i=object`.)
+
```
wilzbach: file 'dmd/cli.d' contains trailing whitespace at line 329

JinShil:
It looks like this is causing the semaphore failure:

> ../dmd/generated/linux/release/32/dmd -conf= -I../druntime/import  -w -de -dip25 -m32 -fPIC -transition=complex -O -release -unittest -version=StdUnittest -c -ofgenerated/linux/release/32/unittest/std/base64.o std/base64.d
src/rt/tracegc.d(43): Deprecation: The delete keyword has been deprecated.  Use object.destroy() instead.

I don't understand why this wasn't caught by the test suite in #7342.

marler8997:
Are you sure that's the cause of the failure?  It looks like it was a unit test failure in `std.datetime`.
```
****** FAIL release64 std.datetime.stopwatch
core.exception.AssertError@std/datetime/stopwatch.d(452): unittest failure
----------------
??:? _d_unittestp [0x20b672a1]
??:? nothrow @safe void std.datetime.stopwatch.__unittest_L442_C15() [0x205bfe83]
??:? void std.datetime.stopwatch.__modtest() [0x205c0314]
??:? [0x4014ab]
??:? [0x4013e3]
??:? [0x4012bb]
??:? runModuleUnitTests [0x20b67ddd]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [0x20b85558]
??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x20b854e3]
??:? _d_run_main [0x20b8544e]
??:? [0x4017cb]
??:? __libc_start_main [0x217bef44]
make[1]: *** [unittest/std/datetime/stopwatch.run] Error 1
0.203s PASS release64 std.datetime
stderr test, please ignore
6.569s PASS release64 std.process
3.202s PASS release64 std.datetime.systime
make[1]: Leaving directory `/home/runner/phobos'
make: *** [unittest-release] Error 2
make: Leaving directory `/home/runner/phobos'
```

changelog/includeimports.dd
```diff
@@ -6,4 +6,4 @@ $(CONSOLE dmd -i=foo -i=-foo.bar)
 
 The option -i by itself is equivalent to:
 
-$(CONSOLE dmd -i=-std,-core,-etc,-object)
+$(CONSOLE dmd -i=-std -i=-core -i=-etc -i=-object)
```
wilzbach: We probably need to cherry-pick this to `stable`.

PR #7778, opened 2018-01-25
Deprecate import/module name mismatch (ALL CASES)

marler8997:
This PR modifies dmd to verify that the name of a module always matches the name that was used to import it.  Depending on the situation, if the names do not match then this PR will either provide a deprecation message or assert an error.

1. _DEPRECATED_: import a module with a qualified name but the module being imported has no module declaration
```D
// main.d
import foo.bar;
// foo/bar.d
// note the missing module declaration
```
```
 Deprecated: module 'bar' from file foo/bar.d should be imported with 'import bar;'
```

2. _DEPRECATED_: import a module with a qualified name that partially matches the name of the module being imported

```D
// main.d
import foo.bar;
import foo.baz.buz;
// foo/bar.d
module bar;
// foo/baz/buz.d
module baz.buz;
```
```
 Deprecated: module 'bar' from file foo/bar.d should be imported with 'import bar;'
 Deprecated: module 'baz.buz' from file foo/baz/buz.d should be imported with 'import baz.buz;'
```
> NOTE: for this rule to apply, the ends of the imported name and the module name itself must match.

3. _ERROR_: importing a module that matches the filename but does not match the module name
```D
// main.d
import foo;
// foo.d
module bar;
```
```
 Error: module 'bar' from file foo.d must be imported with 'import bar;'
```

4. NOT APPLICABLE/STILL WORKS:  import a module with the correct module name but does not match the filename
```D
// main.d
import bar;
// foo.d
module bar;
```

NOTE: this also addresses a bug found by @timotheecour 

https://github.com/dlang/dmd/pull/7777#discussion_r163939438

NOTE: The DMD test suite itself was using some of this functionality in a fair amount of tests, i.e.
* `test/runnable/imports/link7745b.d`
* `test/fail_compilation/imports/a11919.d`

### DIFF
--- a/changelog/deprecate_bad_module_name.dd
+++ b/changelog/deprecate_bad_module_name.dd
@@ -1,0 +1,48 @@
+Deprecate import module name mismatch
+
+Some instances of module name/import name mismatches were deprecated while others are now considered errors.
+
+Case 1: DEPRECATED: import a module name with a qualified name but the module being imported has no module declaration.
+---
+// main.d
+import foo.bar;
+
+// foo/bar.d
+// this file is empty, it has no module declaration
+---
+The above code will now print:
+
+$(CONSOLE Deprecated: module bar from file foo/bar.d should be imported with 'import bar;' )
+
+Case 2: DEPRECATED: import a module with a qualified name that partially matches the name of the module being imported.
+---
+// main.d
+import foo.bar;
+import foo.baz.buz;
+
+// foo/bar.d
+module bar;
+
+// foo/baz/buz.d
+module baz.buz;
+---
+The above code will now print:
+
+$(CONSOLE Deprecated: module bar from file foo/bar.d should be imported with 'import bar;'
+Deprecated: module baz.buz from file foo/baz/buz.d should be imported with 'import baz.buz;' )
+
+Note that for this rule to apply, the shorter name must completely match the end of the longer name.
+
+Case 3: ERROR: import a module that matches the filename but does not match the module name.
+---
+// main.d
+import foo;
+
+// foo.d
+module bar;
+---
+The above code will now fail and print:
+
+$(CONSOLE Error: module bar from file foo.d must be imported with 'import bar;')
+
+Note that importing a module whose module name does not match its filename is still supported.
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -31,6 +31,7 @@ import dmd.globals;
 import dmd.id;
 import dmd.identifier;
 import dmd.parse;
+import dmd.root.array;
 import dmd.root.file;
 import dmd.root.filename;
 import dmd.root.outbuffer;
@@ -514,7 +515,49 @@ extern (C++) final class Module : Package
             buf.printf("%s\t(%s)", ident.toChars(), m.srcfile.toChars());
             message("import    %s", buf.peekString());
         }
-        m = m.parse();
+        {
+            auto originalModule = m;
+            m = m.parse();
+
+            // verify the module name matches the imported module name
+            if (m is originalModule)
+            {
+                if (m.md is null)
+                {
+                    // in this case there was no module declaration, this means the import
+                    // should have no package prefix, but for now we support this with a
+                    // depecration message for backwards compatibility
+                    if (packages && packages.dim > 0)
+                        m.deprecation(loc, "from file %s should be imported with 'import %s;'",
+                            m.srcfile.name.toChars(), m.toPrettyChars());
+                }
+                else if (!packages.opEquals(m.md.packages) || ident != m.md.id)
+                {
+                    // deprecate the case when the ends of the names match
+                    bool deprecate = false;
+                    if (ident == m.md.id)
+                    {
+                        deprecate = true;
+                        auto mdPackageCount = (m.md.packages == null) ? 0 : m.md.packages.dim;
+                        auto packageCount   = (packages      == null) ? 0 : packages.dim;
+                        for (size_t i = 1; i <= mdPackageCount && i <= packageCount; i++)
+                        {
+                            if (m.md.packages[mdPackageCount - i] != packages[packageCount - i])
+                            {
+                                deprecate = false;
+                                break;
+                            }
+                        }
+                    }
+                    if (deprecate)
+                        m.deprecation(loc, "from file %s should be imported with 'import %s;'",
+                            m.srcfile.name.toChars(), m.toPrettyChars());
+                    else
+                        m.error(loc, "from file %s must be imported with 'import %s;'",
+                            m.srcfile.name.toChars(), m.toPrettyChars());
+                }
+            }
+        }
 
         // Call onImport here because if the module is going to be compiled then we
         // need to determine it early because it affects semantic analysis. This is
--- a/src/dmd/root/array.d
+++ b/src/dmd/root/array.d
@@ -313,3 +313,22 @@ unittest
     array[2] = 910;
     assert([123, 421, 910, 123, 1, 2, 8, 20, 4, 3] == array.asDArray);
 }
+
+/**
+ * Returns true if every element of $(D left) and $(D right) are equal.
+ * Note that an array of size 0 is considered equivalent to a `null` pointer.
+ * Params:
+ *  left = pointer to array to compare
+ *  right = pointer to array to compare
+ */
+bool opEquals(T)(const(Array!T)* left, const(Array!T)* right)
+{
+    if (left is null || left.dim == 0)
+        return right is null || right.dim == 0;
+    if (right is null || left.dim != right.dim)
+        return false;
+    foreach (i; 0 .. left.dim)
+        if ((*left)[i] != (*right)[i])
+            return false;
+    return true;
+}
--- a/test/compilable/imports/a313templatemixin1.d
+++ b/test/compilable/imports/a313templatemixin1.d
@@ -1,3 +1,4 @@
+module imports.a313templatemixin1;
 void bug()
 {
 }
--- a/test/fail_compilation/badimport.d
+++ b/test/fail_compilation/badimport.d
@@ -1,0 +1,10 @@
+/*
+EXTRA_SOURCES: imports/nomodname.d
+REQUIRED_ARGS: -Ifail_compilation -de
+PERMUTE_ARGS:
+TEST_OUTPUT:
+---
+fail_compilation/badimport.d(10): Error: module `nomodname` from file fail_compilation/imports/nomodname.d must be imported with 'import nomodname;'
+---
+*/
+import imports.nomodname;
--- a/test/fail_compilation/imports/nomodname.d
+++ b/test/fail_compilation/imports/nomodname.d
@@ -1,0 +1,1 @@
+// intentionally no module declaration
--- a/test/compilable/extra-files/test14894a.d
+++ b/test/compilable/extra-files/test14894a.d
@@ -1,4 +1,4 @@
-module imports.test14894a;
+module test14894a;
 
 mixin template Protocol()
 {
--- a/test/compilable/imports/a313templatemixin2.d
+++ b/test/compilable/imports/a313templatemixin2.d
@@ -1,3 +1,4 @@
+module imports.a313templatemixin2;
 void bug()
 {
 }
--- a/test/compilable/imports/checkimports3a.d
+++ b/test/compilable/imports/checkimports3a.d
@@ -1,1 +1,2 @@
+module imports.checkimports3a;
 void foo() {}
--- a/test/compilable/imports/checkimports3b.d
+++ b/test/compilable/imports/checkimports3b.d
@@ -1,1 +1,2 @@
+module imports.checkimports3b;
 private void foo(int) {}
--- a/test/compilable/imports/checkimports3c.d
+++ b/test/compilable/imports/checkimports3c.d
@@ -1,1 +1,2 @@
+module imports.checkimports3c;
 void foo(string) {}
--- a/test/compilable/imports/fwdref12201a.d
+++ b/test/compilable/imports/fwdref12201a.d
@@ -1,1 +1,2 @@
+module imports.fwdref12201a;
 alias int FILE;
--- a/test/compilable/imports/fwdref2_test17548.d
+++ b/test/compilable/imports/fwdref2_test17548.d
@@ -1,9 +1,9 @@
-module fwdref2_test17548;
+module imports.fwdref2_test17548;
 
 import test17548;
 
 struct S2 {
     void bar(int arg = .test17548.cnst) {}
     S1 s;
-    import fwdref2_test17548;
+    import imports.fwdref2_test17548;
 }
--- a/test/compilable/imports/fwdref9514.d
+++ b/test/compilable/imports/fwdref9514.d
@@ -1,3 +1,4 @@
+module imports.fwdref9514;
 bool find9514(alias pred, R)(R range)
 {
     return true;
--- a/test/compilable/imports/g313public.d
+++ b/test/compilable/imports/g313public.d
@@ -1,3 +1,4 @@
+module imports.g313public;
 void bug()
 {
 }
--- a/test/compilable/imports/g313staticif.d
+++ b/test/compilable/imports/g313staticif.d
@@ -1,3 +1,4 @@
+module imports.g313staticif;
 void bug()
 {
 }
--- a/test/compilable/imports/g313stringmixin.d
+++ b/test/compilable/imports/g313stringmixin.d
@@ -1,3 +1,4 @@
+module imports.g313stringmixin;
 void bug()
 {
 }
--- a/test/compilable/imports/g313templatemixin.d
+++ b/test/compilable/imports/g313templatemixin.d
@@ -1,3 +1,4 @@
+module imports.g313templatemixin;
 void bug()
 {
 }
--- a/test/compilable/imports/imp16085.d
+++ b/test/compilable/imports/imp16085.d
@@ -1,3 +1,4 @@
+module imports.imp16085;
 struct Pass
 {
 }
--- a/test/compilable/imports/imp16085b.d
+++ b/test/compilable/imports/imp16085b.d
@@ -1,4 +1,5 @@
-import imp16085 : S;
+module imports.imp16085b;
+import imports.imp16085 : S;
 
 struct Fail
 {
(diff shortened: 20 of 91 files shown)


test/fail_compilation/badimport.d
```diff
@@ -0,0 +1,4 @@
+// COMPILED_IMPORTS: imports/nomodname.d
+// REQUIRED_ARGS: -Ifail_compilation
+// PERMUTE_ARGS:
+import imports.nomodname;
```
timotheecour: no newline ad EOF

marler8997: Is that a bad thing?

test/fail_compilation/imports/nomodname.d
```diff
@@ -0,0 +1 @@
+// the point of this module is that it has no module name
```
timotheecour: // intentionally no module declaration

src/dmd/root/array.d
```diff
@@ -313,3 +313,13 @@ unittest
...
 }
+
+bool equals(T)(const(Array!T)* left, const(Array!T)* right)
+{
+    if(left is null || left.dim == 0)
+        return right is null || right.dim == 0;
+    foreach (i; 0 .. left.dim)
+        if ((*left)[i] != (*right)[i])
+	    return false;
+    return true;
```
timotheecour: bug? with right.dim > left.dim

test/fail_compilation/badimport.d
timotheecour: i think it's a convention (lots of editor show an annoying marker when we forget one

src/dmd/root/array.d
marler8997: shoot, woops

src/dmd/dmodule.d
```diff
@@ -536,6 +537,10 @@ extern (C++) final class Module : Package
         }
         m = m.parse();
 
+        // verify the module name matches the imported module name
+        if (!packages.equals((m.md is null) ? null : m.md.packages))
+            m.error(loc, "from file %s must be imported with 'import %s;'", m.srcfile, m.toPrettyChars());
```
JinShil: Sadly, I think this will introduce a regression, but probably one D needs.

 - This needs to specified in the specification, if it's not already.  
 - This probably needs to go through the deprecation process.  That stinks, but it probably needs to be done.
 - I think this will require a changelog entry

That's just my opinion/perception.  I'd be happy to fast-track this as I think it's the right thing to do to enforce reliable usage of the language, but that's not my call.

JinShil:
Also, I think we require bugzilla entries for all bug fixes.

timotheecour:
@JinShil  found this old bug which might be related: https://issues.dlang.org/show_bug.cgi?id=15086

JinShil:
> found this old bug

Ok, if this PR addresses that bug, prepend "Fix Isssue xxxx" to this PR's title and to the commit message so @dlang-bot will pick it up.

jacob-carlborg:
Is this anything related to declaring a module with a different name than the file? Because that was used as a feature in DMD, before the Objective-C integration was refactored. Previously there were two files, `objc.d` and `objc_stubs.d`, both used the same module declaration: `module ddmd.objc`. This allowed to swap the implementation in the makefiles based on the platform without having to change any imports. See https://github.com/dlang/dmd/commit/9fe35b144f8ad5fbd96489597ebaf979e9bb064a for reference. This is not used anymore and only one `objc.d` file exists.

marler8997:
It's actually about the import name matching the module name.  You can still have a module name that's different from the file name, but this PR makes it so the import name has to match the module name.

ibuclaw:
Yes, I've always considered this to be a feature.

marler8997:
This PR is unrelated to @jacob-carlborg's example (I assume that's what you were referring to @ibuclaw).

src/dmd/root/array.d
```diff
@@ -313,3 +313,15 @@ unittest
     array[2] = 910;
     assert([123, 421, 910, 123, 1, 2, 8, 20, 4, 3] == array.asDArray);
 }
+
```
UplinkCoder: why this ?

marler8997: Allows you to compare to pointers to array and handles the `null` case for you as well

marler8997: ok, changed it to `opEquals`.

timotheecour:
so that module name == import name == _MODULE__ is guaranteed.

On Jan 25, 2018 11:46 PM, "Jonathan Marler" <notifications@github.com>
wrote:

It's actually about the import name matching the module name. You can still
have a module name that's different from the file name, but this PR makes
it so the import name has to match the module name.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/dlang/dmd/pull/7778#issuecomment-360707335>, or mute
the thread

src/dmd/dmodule.d
```diff
@@ -536,6 +537,20 @@ extern (C++) final class Module : Package
...
+        {
+            // in this case there was no module declaration, this means the import
+            // should have no package prefix, but for now we support this with a
+            // depecration message backwards compatibility
+            if (packages && packages.dim > 0)
+                m.deprecation(loc, "from file %s should be imported with 'import %s;'",
+                    m.srcfile.name.toChars(), m.toPrettyChars());
+        }
+        else if (!packages.opEquals(m.md.packages) || ident != m.md.id)
+            m.error(loc, "from file %s must be imported with 'import %s;'",
```
RazvanN7: This should probably go through a deprecation cycle.

src/dmd/root/array.d
wilzbach: Why not `left.asDArray == right.asDArray`?

src/dmd/root/array.d
```diff
@@ -313,3 +313,15 @@ unittest
...
+bool opEquals(T)(const(Array!T)* left, const(Array!T)* right)
```
wilzbach: As always - new symbols:
- should have docstrings
- should use `ref` or `out` instead of pointers (except there's a good reason not to)

src/dmd/dmodule.d
marler8997: Maybe.  I think that some variations are more likely to be used out in the wild than others.  Here's my categorization of the various cases:

1. import a module with a qualified name and the module being imported has no module declaration

```D
// main.d
import foo.bar;
// foo/bar.d
// note the missing module declaration
```
This PR doesn't break this case but it does print a deprecation message.

2. import a module with a qualified name that partially matches the name of the module being imported

```D
// main.d
import foo1.bar;
import foo2.bar.baz.
// foo1/bar.d
module bar;
// foo2/baz/buz.d
module baz.buz;
```
Currently this PR makes these cases errors, however, I could see an argument to deprecate these instead.  doing so would be a simple modification to this PR, what do people think?

3. importing a module that matches the filename but does not match the module name
```D
// main.d
import foo;
// foo.d
module bar;
```
This is one case that I don't think we need to deprecate. I say kill it with fire and extreme prejudice.

marler8997: Note that this doesn't break the following case:
```
// main.d
import bar;
// foo.d
module bar;
```

src/dmd/root/array.d
marler8997: since `left` and `right` could be null, using ref seems odd to me in this case

src/dmd/root/array.d
marler8997: Would have to be `(*left).asDArray` and `(*right).asDArray`, and this segfaults when left or right is null.

timotheecour:
@jacob-carlborg  @ibuclaw 

> Is this anything related to declaring a module with a different name than the file? Because that was used as a feature in DMD

This is what the flag `-mv=` is for, to allow customizing the file <=> module mapping. This PR doesn't break that.

jacob-carlborg:
> This is what the flag -mv=<package.module>=<filespec> is for, to allow customizing the file <=> module mapping. This PR doesn't break that.

This was done in DMD without that flag.

marler8997:
No activity for 11 days.  I haven't tried to push anyone else to look at this since I'd rather my other PR's receive attention since they need to be merged quickly before the next release, but I have plenty of time if anyone has time to review/approve/request changes.

wilzbach (approved):
@marler8997 this LGTM. There's still a small nit from @timotheecour, but apart from that I'm just waiting here for @andralex or @WalterBright to have a brief look as seen on the testsuite it might trigger a couple of deprecation messages. However, it's important to realize that (1) it's not breaking any code yet (we never need to follow through with the deprecation and (2) this is in accordance with the D spec.

test/compilable/imports/a313templatemixin1.d
```diff
@@ -1,3 +1,4 @@
+module imports.a313templatemixin1;
```
andralex: Why is this necessary? We should continue to work if the module declaration is not present.

andralex (commented):
It seems this PR forces all modules imported via a path to use a module declaration. Is this correct? If so, it's too much.

test/compilable/imports/a313templatemixin1.d
marler8997: It's one way to go.  You could also add the `imports` directory to the "import path list", i.e.
```
dmd -Icompilable/imports ...
```
then no module declaration is required.

andralex: @marler8997 can't we just penalize people who enter wrong module information? It seems this PR does that but also adds another restriction, which is more difficult to justify.

marler8997:
> It seems this PR forces all modules imported via a path to use a module declaration. Is this correct? If so, it's too much.

I'm not sure exactly what you mean.  There's a couple different cases.  If you import a module using an "unqualified name", then no module declaration is required.  However, if you import a module with a "qualified name", and that module has no module declaration, then this PR will print a deprecation message. i.e.
```D
// main.d
import foo.bar.baz.buzz;

// foo/bar/baz/buzz.d
// no module declaration (deprecation message)
```

andralex:
@marler8997 if I understand correctly you impose that no module could be imported through two different paths. Isn't that a bit excessive?

marler8997:
> @marler8997 if I understand correctly you impose that no module could be imported through two different paths. Isn't that a bit excessive?

2 different paths is ok, it's 2 different module names that this is preventing.

marler8997:
```D
// main.d
import foo.bar.baz.buz;
// other.d
import baz.buz;
//buz.d

```
Are you saying you think this is OK?

test/compilable/imports/a313templatemixin1.d
marler8997: Sorry, I'm not sure exactly what you're trying to say.  There's 2 cases listed in the original PR message that have been deprecated, and 1 that has been declared to be an error.  Do those use cases categorize the cases you are thinking of?  If so, which ones do you think should not be deprecated/made errors?

andralex: @marler8997 yah let me discuss exactly on the cases you present in the changelog. Thanks for the neat organization.

changelog/deprecate_bad_module_name.dd
```diff
@@ -0,0 +1,47 @@
...
+
+// foo/bar.d
+module bar;
+
+// foo/baz/buz.d
+module baz.buz;
+---
+The above code will now print:
+
+$(CONSOLE Deprecated: module bar from file foo/bar.d should be imported with 'import bar;'
```
andralex: Alternatively: replace module declaration with `module foo.bar;`

We need to show advantages of this as well. For example show how the current rules lead to a breach in modularity.

My current mental model goes like this: `main.d` uses a dotted import to guide path selection, and that also confers a local name for the imported module. The fact that the module gave itself a non-dotted name is irrelevant to `main.d`. I see nothing wrong in the fact that the module name from the perspective of code inside `foo/bar.d` is different from the module name from the perspective of code inside `main.d`. In fact we have a language construct that messes exactly with that:

```d
import takezat = foo.bar;
```

We'd need a good argument that the above is a bad thing.

changelog/deprecate_bad_module_name.dd
```diff
@@ -0,0 +1,47 @@
...
+Deprecated: module baz.buz from file foo/baz/buz.d should be imported with 'import baz.buz;' )
```
andralex: A similar line of reasoning would apply here. We need to build a case on why that's bad.

changelog/deprecate_bad_module_name.dd
```diff
@@ -0,0 +1,47 @@
...
+
+Note that for this rule to apply, the shorter name must completely match the end of the longer name.
+
+Case 3: ERROR: import a module that matches the filename but does not match the module name.
+---
+// main.d
+
```
andralex: Did you mean to add some content to main.d?

changelog/deprecate_bad_module_name.dd
```diff
@@ -0,0 +1,47 @@
...
+// foo.d
+module bar;
+---
+The above code will now fail and print:
+
+$(CONSOLE Error: module bar from file foo.d must be imported with 'import bar;')
```
andralex: I don't understand how the file would ever be found. So the compiler is supposed to compile `main.d`. It sees `import bar;`. How does it know it's supposed to look for a file called `foo.d`?

changelog/deprecate_bad_module_name.dd
```diff
@@ -0,0 +1,47 @@
+Deprecate import module name mismatch
+
+Some instances of module name/import name mismatches were deprecated while others are now considered errors.
+
+Case 1: DEPRECATED: import a module name with a qualified name but the module being imported has no module declaration.
```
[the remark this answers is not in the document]
marler8997: > This effectively imposes that modules inside directories (frequent situation) must have a module declaration.

Actually it just means that they must be imported with a "non-qualified" name.

andralex:
>Are you saying you think this should be OK?

@marler8997 what would be the directory structure and compiler flags? Would those be in the same project? (If different projects, I can definitely say that should be okay.)

Overall: this diff renders a use of module names used in a bunch of our own tests as an antipattern. I think this is a serious sign we need to scrutinize the matter appropriately.

changelog/deprecate_bad_module_name.dd
marler8997: Interesting.  That's not how I originally understood the module system to work, but your description matches the implementation that I found.

One disadvantage of the current mechanism is that this causes the compiler to load the same module multiple times.  Each unique path used to import the module will result in another instance of the module being instantiated.  I think there's more to say but I'll leave it at that for now.

marler8997:
> Overall: this diff renders a use of module names used in a bunch of our own tests as an antipattern. I think this is a serious sign we need to scrutinize the matter appropriately.

Definitely.  I welcome the scrutiny.  It looks like we'll have to explore this more, maybe I can come up with some use cases that demonstrate the advantages of these restrictions.

changelog/deprecate_bad_module_name.dd
marler8997: The file is give on the command line.  Module names not matching file names is a technique already used, though the import statement will match the module name (just no the filename).  You have 3 different names:

1. Filename
2. Import Name
3. Module Name

This PR enforces that "import name" and "module name" should match, but filename does not necessarily need to.  @jacob-carlborg provided an example where this is useful (https://github.com/dlang/dmd/pull/7778#issuecomment-360706239)

marler8997:
> @marler8997 what would be the directory structure and compiler flags? Would those be in the same project? (If different projects, I can definitely say that should be okay.)

I've created the small example project I mentioned and depending on how I compile it/if I add/remove module names, etc. I end up with all kinds of weird errors.  Some cases I get a linker error, some cases I get a compilation error, and some cases I don't get any error, but each case is very non-intuitive and requires intimiate knowledge of how the compiler works to understand what's going on.

I argue that the ability to import a module using different names severely breaks the module system.  Normally if people play nice and don't do weird things then it's ok, but this PR is preventing those weird abuses which in my opinion have to utility and only serve to add confusion to the module system.

marler8997:
Let's look at one of the weird things that can happen with the current state of the "unverified" module system. Take the following 2 files:
```D
// main.d
import foo.bar;
void main() { barfunc(); }

// foo/bar.d
void barfunc() { }
```

If you compile "foo/bar.d" seperately like this:
```
dmd -c foo/bar.d
```

Then you can successfully compile the full program via
```
dmd main.d bar.o
```

However, if you compile it fully from source you will get a compilation error:
```
dmd main.d foo/bar.d
```
```
Error: module bar from foo/bar.d must be imported with `import bar;`
```

This error occurs becase the module `foo/bar.d` is being loaded twice with 2 different names.  The first time it is loaded because it is given on the command line, and since it has no module declaration, the compiler gives it the name `bar`.  But then the module gets imported with the name `foo.bar`, however, the compiler cannot change the name of the module and all existing reference to it so an error must be asserted.  In the first case where the module is precompiled, the module only gets loaded once from the import so it will be given the proper name of `foo.bar` and the compilation succeeds.

This difference in behavior couples the module system with how the files are compiled.  Compiling some combinations of files will succeed while others will not.

This PR breaks this coupling by forbidding this behavior altogether.  No matter what combination you compile the files in, you will always get the following error:
```
Error: module bar from foo/bar.d must be imported with `import bar;` or given a name with `module foo.bar;`
```

This example demonstrates the core problem with supporting the first 2 cases that this PR aims to deprecate. However, note that this example is just one manifestation of the core problem.  The same core problem could also cause different files to be imported from the same import statement, which can cause nonsensical compilation/linker errors which are very difficult to root cause.

marler8997:
P.S. Many of the tests in the DMD test suite will only compile if you compile the imports separately from the test file itself.  If you compile them all together in one invocation, you will get an import error.  This demonstrates a real example where not verifying import/module names puts a requirement on what order/how you compile your modules together.

WalterBright:
The fact that this PR requires 80-90 files in the test suite to be edited casts serious doubt on the viability of this change w.r.t. breaking existing code.

marler8997:
Editing all those files is not required. I'll modify the PR to remove those changes to demonstrate.

wilzbach:
Better submit a new PR, s.t. you don't loose these changes ;-)

marler8997:
Since it looks like it's going to take some work to reach consensus on cases 1 and 2, I'm going to split this PR.  I'm going to push a new PR that only addresses case 3 (which addresses issue 15086).  Once that is merged I can create a new PR that addresses case 1 and 2.  This limits the scope of each PR making them easier to discuss and prevents conflation of the individual points for each case.

marler8997:
Closing this one in favor of (https://github.com/dlang/dmd/pull/7878) and potentially a new PR in the future for cases 1 and 2.

PR #7779, opened 2018-01-25
Ddoc: Autodetected URLs should use different macro than LINK.

quickfur:
So that users can override / suppress URL autodetection if necessary.  The default style, of course, simply redirects `DDOC_LINK_AUTODETECT` to `LINK`.

Implementation of: https://github.com/dlang/dlang.org/pull/2127


jacob-carlborg (approved):
Looks good. I would like someone else to have a look as well.

andralex (approved):
ah there we go

PR #7781, opened 2018-01-26
Highlight error messages:  dsymbol.d auto-appended toPrettyChars and a few leftovers

JinShil:
Code changes are pretty small, but there's a LOT of tests affected.

This should be the last major PR for the campaign of highlighting error messages.  There should only be a few left-overs and subjective error messages that remain.


test/fail_compilation/ice15441.d
```diff
@@ -1,8 +1,8 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/ice15441.d(24): Error: variable ice15441.main.__front$n$ type `void` is inferred from initializer `__r$n$.front()`, and variables cannot be of type `void`
-fail_compilation/ice15441.d(24): Error: expression `__r$n$.front()` is `void` and has no value
+fail_compilation/ice15441.d(24): Error: variable `ice15441.main.__front3` type `void` is inferred from initializer `__r2.front()`, and variables cannot be of type `void`
+fail_compilation/ice15441.d(24): Error: expression `__r2.front()` is `void` and has no value
```
JinShil: This change troubled me.  I don't understand what I did in dsybmol.d to cause this, and I'm not sure what's right.

src/dmd/dsymbol.d
```diff
@@ -259,31 +259,39 @@ extern (C++) class Dsymbol : RootObject
     {
         va_list ap;
         va_start(ap, format);
-        .verror(loc, format, ap, kind(), toPrettyChars());
+        auto cstr = toPrettyChars();
```
WalterBright: I'd use const, not auto

src/dmd/dsymbol.d
```diff
@@ -259,31 +259,39 @@ extern (C++) class Dsymbol : RootObject
...
+        scope auto pretty = '`' ~ cstr[0 .. strlen(cstr)] ~ "`\0";
```
WalterBright: const again

src/dmd/dsymbol.d
JinShil: Done.

src/dmd/dsymbol.d
JinShil: Done.

test/fail_compilation/ice15441.d
rainers: `$n$` is a d_do_test placeholder for any number to avoid having to adjust error messages whenever the compiler generates different unique numbers due to changes elsewhere. The test suite should pass with `$n$`, too.

JinShil: I wonder if the tick marks interfere with the d_do_test's text substitution.

jacob-carlborg:
I recommend squashing all commits. The three extra commits don't add any value.

JinShil:
> I recommend squashing all commits. The three extra commits don't add any value.

Squashed!

jacob-carlborg (approved):

test/fail_compilation/ice15441.d
rainers: It's agnostic to anything but `$n$`: https://github.com/dlang/dmd/blob/master/test/d_do_test.d#L447

JinShil: Thanks @rainers, `$n$` is back in.

JinShil:
Ok, I'm finished interrupting @dlang-bot.  This should be ready for merge now.

PR #7782, opened 2018-01-26
Add a test for building the dmd frontend with dmd -i

wilzbach:
Building the dmd frontend with `dmd -i` already works, so better add a simple CI check to ensure that it continues to work.


jacob-carlborg:
The PR, the commit message and the comment mention "rdmd" but I don't see rdmd being used.

wilzbach:
Sorry I used `rdmd` initially, but then realized that we don't even need `rdmd` here, but can use vanilla `dmd` with the new `-i` feature.

jacob-carlborg (approved):

marler8997 (approved):

PR #7783, opened 2018-01-26
Fix the deprecation warning in the DMD DUB package + build with `-de` + CI check

wilzbach:
`de` is used already for the DMD build, tools, phobos etc. It was simply forgotten for the new DMD DUB package. The idea of building with `-de` is that whenever a new deprecation message gets introduced it should be ensured that all core packages can use the replacement. It's also bad for the image if user see deprecation warnings in a library they use.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -103,7 +103,7 @@ test_dub_package() {
             # build with host compiler
             dub --single "$file"
             # build with built compiler (~master)
-            dub --single --compiler="${abs_build_path}" "$file"
+            DFLAGS="-de" dub --single --compiler="${abs_build_path}/dmd" "$file"
         done
         popd
     fi
--- a/ci.sh
+++ b/ci.sh
@@ -10,6 +10,7 @@ if [ -z ${MODEL+x} ] ; then echo "Variable 'MODEL' needs to be set."; exit 1; fi
 if [ -z ${DMD+x} ] ; then echo "Variable 'DMD' needs to be set."; exit 1; fi
 
 CURL_USER_AGENT="DMD-CI $(curl --version | head -n 1)"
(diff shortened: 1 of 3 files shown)


jacob-carlborg (approved):

JinShil (approved):

dub.sdl
```diff
@@ -60,6 +60,7 @@ subPackage {
   versions "GC"
   versions "NoMain"
   sourcePaths "src/dmd"
+  buildRequirements "disallowDeprecations"
```
MartinNowak: Please don't hardcode this into the package, it incurs immediate breakage on deprecations for people that cannot update their compiler.

MartinNowak:
> The idea of building with -de is that whenever a new deprecation message gets introduced it should be ensured that all core packages can use the replacement.

Yes, we should test that `-de` compiles and works for us, but we shouldn't enforce package users to use it as well, as this removes the deprecation phase out of deprecations.

ci.sh
```diff
@@ -97,9 +97,13 @@ test_dub_package() {
     if [ "${DMD:-dmd}" == "gdmd" ] ; then
         echo "Skipping DUB examples on GDC."
     else
+        local abs_build_path="$PWD/$build_path"
         pushd test/dub_package
         for file in *.d ; do
+            # build with host compiler
             dub --single "$file"
+            # build with built compiler (~master)
+            dub --single --compiler="${abs_build_path}" "$file"
```
MartinNowak: You could either add sub-configurations to dub.sdl or build with DFLAGS to check that -de succeeds.

wilzbach: Ok.

RazvanN7:
LGTM

PR #7785, opened 2018-01-26
Fix Issue 17625 - Confusing error message for private functions in different modules

RazvanN7:
The problem was that whenever a symbol is searched in different modules if the same symbol is found in 2 or more different modules an overload set is created no matter the privacy attribute of the symbol. I added a check to make sure that the symbol found is accessible, however that made it so that some deprecations where not issued anymore on some fail_compilation tests. I deleted the tests since it wasn't possible to modify them to get rid of the deprecation. I don't know what the purpose of those tests was, but with the new patch the code does not deprecate anymore. Ideas?


test/fail_compilation/checkimports3.d
```diff
@@ -1,15 +0,0 @@
-/*
```
wilzbach: > [Thou shalt not](https://en.wikipedia.org/wiki/Thou_Shalt_Not) delete existing tests :imp: 

If they really pass now, they should be moved to `compilable`.

wilzbach (commented):
.

andralex (approved):

test/fail_compilation/fail17625.d
```diff
@@ -0,0 +1,17 @@
...
+TEST_OUTPUT:
+---
+fail_compilation/fail17625.d(16): Deprecation: `b17625.boo` is not visible from module `fail17625`
+fail_compilation/fail17625.d(16): Error: function `b17625.boo` is not accessible from module `fail17625`
+---
+*/
+
+module fail17625;
+
+import imports.a17625;
```
marler8997: FYI: You are importing the modules using `imports.a17625` but then you declare the module name in the file to be `a17625`.  I'm hoping to deprecate this via https://github.com/dlang/dmd/pull/7778

PR #7786, opened 2018-01-26
Don't generate exception handling code if `Throwable` is not declared

JinShil:
This PR is a follow up to #7395 and #7768.  It is to `Throwable` what #7395 and #7768 were to `ModuleInfo`.

In D, anything that can be `throw`n must inherit from `object.Throwable`.  Therefore, if `object.Throwable` does not exist, no exception handling code should be generated, and any attempt to `throw` should emit compilation error.

With #7395, #7768, and this PR, we can now generate a minimal executable with almost no boilerplate and produce very small resulting binaries.

Example 1
---------
object.d
```D
module object;

private alias extern(C) int function(char[][] args) MainFunc;
private extern (C) int _d_run_main(int argc, char** argv, MainFunc mainFunc)
{
    return mainFunc(null);
}
```

main.d
```D
void main() { }
```

Compile
```
dmd -conf= -defaultlib= main.d object.d -of=main
```
Result
```
size main
   text	   data	    bss	    dec	    hex	filename
   1403	    584	     16	   2003	    7d3	main
```

Example 2
---------
object.d
```D
module object;

extern(C) void __d_sys_exit(long arg1)
{
    asm
    {
        mov RAX, 60;
        mov RDI, arg1;
        syscall;
    }
}

extern void main();
private extern(C) void _start()
{
    main();
    __d_sys_exit(0);
}
```

main.d
```D
void main() { }
```

Compile and link
```
dmd -c -lib main.d object.d -of=main.o
ld main.o -o main
```

Result
```
size main
   text	   data	    bss	    dec	    hex	filename
     56	      0	      0	     56	     38	main
```

Again, this will be of interest to users wishing to incrementally or partially port D to new platforms, users hoping to target resource constrained platforms, and users wishing to use D as a library from other languages without druntime.


src/dmd/dclass.d
```diff
@@ -194,7 +194,7 @@ enum ClassKind : int
  */
 extern (C++) class ClassDeclaration : AggregateDeclaration
 {
-    extern (C++) __gshared
+    extern (C++) static __gshared
```
wilzbach: It's already `__gshared`. DMD should emit a warning when both are used as this doesn't make any sense.

wilzbach (commented):
How about adding the second example to the testsuite too?

src/dmd/dclass.d
JinShil: I was following the example here: https://github.com/dlang/dmd/blob/e17992f4b7707b6ce6ba26921cb7420732fe5afd/src/dmd/dmodule.d#L312

But, I'll change it.

JinShil:
> How about adding the second example to the testsuite too?

There's a few concerns I have about that:
 - It requires compiling and linking separately due to all the hard-coded linking logic in DMD.  I'm not sure how to do that in the test suite.
 - It's very Linux 64-bit specific.
 - I think verifying the size will be quite brittle and too sensitive to small insignificant changes.

But I'll add it if it's desired.

src/dmd/backend/mscoffobj.c
```diff
@@ -1043,12 +1043,17 @@ void MsCoffObj::startaddress(Symbol *s)
 
 bool MsCoffObj::includelib(const char *name)
 {
-    int seg = seg_drectve();
-    //dbg_printf("MsCoffObj::includelib(name *%s)\n",name);
-    SegData[seg]->SDbuf->write(" /DEFAULTLIB:\"", 14);
-    SegData[seg]->SDbuf->write(name, strlen(name));
-    SegData[seg]->SDbuf->writeByte('"');
-    return true;
+    if (name && strlen(name) > 0)
```
WalterBright: This isn't checked in cgobj.c, why here? What does it have to do with Throwable?

src/dmd/link.d
```diff
@@ -220,9 +220,14 @@ public int runLINK()
...
-                cmdbuf.writestring("/DEFAULTLIB:");
-                writeFilename(&cmdbuf, global.params.libfiles[i]);
+                // The `/DEFAULTLIB:` argument cannot be empty
+                const f = global.params.libfiles[i];
+                if (f && strlen(f) > 0)
+                {
+                    cmdbuf.writeByte(' ');
+                    cmdbuf.writestring("/DEFAULTLIB:");
+                    writeFilename(&cmdbuf, f);
+                }
```
WalterBright: This looks like it belongs in a separate PR?

src/dmd/backend/mscoffobj.c
JinShil: It's because the tests are compiled with `defaultlib=`, and the mircosoft linker rejects empty arguments to `/DEFAULTLIB:`.  It's just necessary to get the tests to pass on Windows.

src/dmd/statementsem.d
```diff
@@ -3737,7 +3743,8 @@ else
 
         auto blockexit = tfs._body.blockExit(sc.func, false);
 
-        if (!global.params.useExceptions)       // if not worrying about exceptions
+        // if not worrying about exceptions
+        if (!global.params.useExceptions || !ClassDeclaration.throwable)
```
WalterBright: Elsewhere is written `useExceptions && throwable`, so I suggest rewriting this to be consistent:
```
if (!(global.params.useExceptions && ClassDeclaration.throwable))
```

src/dmd/link.d
JinShil: I could do that, but I don't have a test to verify it.  This is PR has the logic necessary to create a test that actually benefits from the `defaultlib=` (i.e. no defaultlib).  So, I think they need to go together.

src/dmd/backend/mscoffobj.c
WalterBright: BTW, in glue.d where this function is called:
```
if (global.params.mscrtlib && global.params.mscrtlib[0])
      objmod.includelib(global.params.mscrtlib);
```
so now the check is being done twice. Also, the `strlen` can be safely replaced with `[0]`.

src/dmd/link.d
```diff
@@ -220,9 +220,14 @@ public int runLINK()
             }
             for (size_t i = 0; i < global.params.libfiles.dim; i++)
             {
-                cmdbuf.writeByte(' ');
-                cmdbuf.writestring("/DEFAULTLIB:");
-                writeFilename(&cmdbuf, global.params.libfiles[i]);
+                // The `/DEFAULTLIB:` argument cannot be empty
+                const f = global.params.libfiles[i];
+                if (f && strlen(f) > 0)
```
WalterBright: replace `strlen(f) > 0` with `f[0]`

WalterBright: also, perhaps this check should be done where `libfiles[]` is filled, rather than where it is used.

WalterBright:
Good idea!

src/dmd/link.d
WalterBright: Should clearly set up the code so that no default lib has `defaultlib` set to null, not a 0 length string.

src/dmd/statementsem.d
JinShil: Done

JinShil:
The tests in this PR require `defaultlib=` (i.e. no defaultlib).  Prior to this PR, that parameter set `defaultlibname` to an empty string, which was then passed to the `/DEFAULTLIB:` argument of Microsoft's linker on Windows.  However, Microsoft's linker throws an error if the argument to  `/DEFAULTLIB:` is empty.

Per @WalterBright's suggestion in prior comments in this PR, `defaultlibname` is now set to `null` in mars.d.  However, that means logic in link.d and glue.d now needs to check `defaultlibname` for `null` to ensure there are no segfaults and that the `/DEFAULTLIB:` argument is not empty.

@WalterBright, this is ready for another look.  Thanks.

PR #7788, opened 2018-01-26
Add dmd.frontend to the documentation build

wilzbach:
As `dmd.frontend` isn't part of the default sources, it needs to be added manually.
This depends on https://github.com/dlang/dmd/pull/7783 as `dlang.org` builds the documentation with `-de`.

-- edit:

FYI: ddox doesn't have this problem as it doesn't need the Makefiles -> https://dlang.org/library-prerelease/dmd/frontend.html


WalterBright (approved):

jacob-carlborg (approved):

wilzbach:
Removed auto-merge to save resources. This won't pass atm as it needs #7783  (dlang.org builds the documentation with -de)

PR #7789, opened 2018-01-26
Use lazy initialization for dmd.frontend

wilzbach:
Lazy initialization seems to be the new game in town.
`dmd.frontend` wants to play with the big kids too!

### DIFF
--- a/src/dmd/frontend.d
+++ b/src/dmd/frontend.d
@@ -27,6 +27,7 @@ This needs to be done $(I before) calling any function.
 */
 void initDMD()
 {
+    import std.concurrency : initOnce;
     import dmd.builtin : builtin_init;
     import dmd.dmodule : Module;
     import dmd.expression : Expression;
@@ -37,16 +38,21 @@ void initDMD()
     import dmd.objc : Objc;
     import dmd.target : Target;
 
-    global._init();
-    addDefaultVersionIdentifiers();
+    static shared bool initialized;
+    initOnce!initialized((){
+        global._init();
+        addDefaultVersionIdentifiers();
 
-    Type._init();
-    Id.initialize();
-    Module._init();
-    Target._init();
-    Expression._init();
-    Objc._init();
-    builtin_init();
+        Type._init();
+        Id.initialize();
+        Module._init();
+        Target._init();
+        Expression._init();
+        Objc._init();
+        builtin_init();
+
+        return true;
+    }());
 }
 
 /**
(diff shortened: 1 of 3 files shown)


src/dmd/frontend.d
```diff
@@ -27,6 +27,7 @@ This needs to be done $(I before) calling any function.
 */
 void initDMD()
 {
+    import std.concurrency : initOnce;
```
WalterBright: We're not linking with Phobos - just druntime.

WalterBright:
I'm afraid I don't see how this improves anything.

src/dmd/frontend.d
wilzbach: This file uses Phobos as it isn't part of the compiler.

wilzbach:
The `dmd.frontend` is user-facing and this is one small step into making it easier for people to use as it reduces the number of things people can do wrong. `initOnce` is used very generously in Phobos and this file is very similar to Phobos - it's a _user_ API after all.

jacob-carlborg:
What happens if `initDMD` is called more than once?

rainers:
> What happens if initDMD is called more than once?

Reinitialization is not so easy due to "inaccessible" globals: here's what I had to do in my experiments for Visual D some time ago: 

https://github.com/rainers/visuald/blob/dmdserver/vdc/dmdserver.d#L1009
https://github.com/rainers/visuald/blob/dmdserver/vdc/dmdserver.d#L667

src/dmd/frontend.d
```diff
@@ -37,16 +38,21 @@ void initDMD()
...
-    addDefaultVersionIdentifiers();
-
-    Type._init();
-    Id.initialize();
-    Module._init();
-    Target._init();
-    Expression._init();
-    Objc._init();
-    builtin_init();
+    static shared bool initialized;
```
rainers: The front end is not thread safe, no need to pretend it here.

jacob-carlborg:
> Reinitialization is not so easy due to "inaccessible" globals: here's what I had to do in my experiments for Visual D some time ago.

I was thinking if something bad can happen if `initDMD` is called more than once.

src/dmd/frontend.d
```diff
@@ -60,6 +66,8 @@ void addImport(string path)
     import dmd.arraytypes : Strings;
     import std.string : toStringz;
 
+    initDMD;
```
WalterBright: Please let's not introduce a new style for calling functions - append the `()`.

WalterBright:
This just seems like over-engineering. Just use:
```
__gshared bool initialized;
assert(!initialized);
initialized = true;
```
Done - no import, no delegate, no dependencies, no "what does this do", etc.

PR #7790, opened 2018-01-26
[No Pull] Trace symbol time

UplinkCoder:
This is just to use the tester.


UplinkCoder:
Why does this thing automatically request freaking reviews.

UplinkCoder:
you know why this fail, right ?
Because during the compilation of phobos there are more then 64M calls to semantic.

UplinkCoder:
@quickfur you might be interested in this :)

quickfur:
Thanks! I'll git pull this to my local repo and see if I can find out what's going on with my low-memory system.

UplinkCoder:
@quickfur you'll have to modify root/rdmd.d and add a static varaible to sum up the allocated mem, put that into the symbol-trace-record, and you should be able to reconstruct the allocation patterns from there.

UplinkCoder:
What is going on here ?
I never closed this.

quickfur:
Are you sure somebody didn't hijack your github account?  I've never had this happen to me.

PR #7791, opened 2018-01-27
Add test coverage for `scope class`

JinShil:
All this does is add test coverage for this:  https://codecov.io/gh/dlang/dmd/src/master/src/dmd/typesem.d#L800...801

### DIFF
--- a/test/fail_compilation/scope_class.d
+++ b/test/fail_compilation/scope_class.d
@@ -1,0 +1,20 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/scope_class.d(10): Error: functions cannot return `scope scope_class.C`
+---
+*/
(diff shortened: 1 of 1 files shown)


WalterBright (approved):

jacob-carlborg:
All the red, is that lines that are not covered? Do you have another PR for those or do we settle with this?

JinShil:
> All the red, is that lines that are not covered? Do you have another PR for those or do we settle with this?

It's out of scope for this PR.  I need this for an upcoming another PR (#7771), to keep the review focused on the problem it's addressing.  Right now that discussion is taking too many tangents, and this will remove one of those tangents.

jacob-carlborg (approved):

PR #7792, opened 2018-01-27
Add Trace

UplinkCoder:
This is a #7790 all over again because github is being annoying.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -936,8 +936,32 @@ private int tryMain(size_t argc, const(char)** argv)
         char[255] fileNameBuffer;
         import core.stdc.time : ctime, time;
         auto now = time(null);
+        auto timeString = ctime(&now);
+        // replace the ' ' by _ and '\n' or '\r' by '\0'
+        {
+            int len = 0;
+            char c = void;
+            for(;;)
+            {
+                c = timeString[len++];
+                // break on null, just to be safe;
+                if (!c)
+                    break;
+
+                if (c == ' ')
+                    timeString[len - 1] = '_';
+
+                if (c == '\r' || c == '\n')
+                {
+                    timeString[len - 1] = '\0';
+                    break;
+                }
+            }
+        }
+
         sprintf(&fileNameBuffer[0],
-             "%s-%s.%s".ptr, "symbol".ptr, ctime(&now), "log".ptr);
+             "symbol-%s.log".ptr, timeString);
+
         auto f = File(&fileNameBuffer[0]);
 
         bufferPos += sprintf(cast(char*) bufferPos, "//");
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -62,6 +62,8 @@ import dmd.target;
 import dmd.tokens;
 import dmd.utils;
 
+import dmd.trace;
+
 /**
  * Print DMD's logo on stdout
  */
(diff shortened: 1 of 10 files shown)


wilzbach:
> because github is being annoying.

Huh? You can force-push to commits ;-)
And if you accidentally pushed an empty commit to a PR or deleted your branch, you can restore the commit sha in your git history...

UplinkCoder:
I did force-push but it's not reopening

MartinNowak:
Just a bit of explanation, we're trying to trace more precise compile times (on a per symbol scope with self-time and children-time for things like template instantiation and semantic3). This should eventually help to gain a much better understanding of what's causing slow compile times.

UplinkCoder:
The reason this fails are most likely memory constraints.
@quickfur If you can share your code-example with me (uplink.coder@gmail.com)
then I can profile the memory usage of it.

quickfur:
Actually, I was having trouble compiling Phobos. I managed to hack through it by splitting up the object files into 3 batches then linking it as a final step. But afterwards I continued to have trouble building other D projects that passed all source files on the command-line; I had to split it up.

This was on a (relatively) low-memory system (1GB RAM total), so anything that goes above that in your profile would be problematic.

UplinkCoder:
@quickfur the worst one would be std.traits or std.regex. it depends on how they are used as well.
and the recent move from druntime functions to templates was not a good one.

quickfur:
The same low-memory machine used to be able to compile Phobos and other small-ish D projects no problem. But recently it seems that the memory consumption has increased, and not just a little.  I think we need a major project aimed at improving the way dmd deals with templates.  I believe there's plenty of room for improvement there. (But before that, can we pretty please have newCTFE merged? ;-) )

UplinkCoder:
@quickfur The major blocker on newCTFE right now are still classes. And the fact that I have vastly less time for it. After classes I'll have to deal with delegates whose behavior is going to be difficult to mimic.
b/c of this captures or sometimes on the stack and sometimes on the scope wishy washy.

quickfur:
@UplinkCoder If you have any pointers as to how the rest of us can chip in to prod newCTFE along, I'm all ears.  I'm highly interested in newCTFE, and if I can would contribute to making it see the light of day.

src/dmd/mars.d
```diff
@@ -923,6 +925,55 @@ private int tryMain(size_t argc, const(char)** argv)
...
+        // then 32 Megabyte of log-buffer;
+        char* fileBuffer = cast(char*)malloc(ushort.max * 512 * 64);
+        char* bufferPos = fileBuffer;
+
+        char[255] fileNameBuffer;
+        import core.stdc.time : ctime, time;
+        auto now = time(null);
+        sprintf(&fileNameBuffer[0],
+             "%s-%s.%s".ptr, "symbol".ptr, ctime(&now), "log".ptr);
+        auto f = File(&fileNameBuffer[0]);
```
dkgroot: Your filename contains spaces and unprintable character '\n'
```
-        import core.stdc.time : ctime, time;
-        auto now = time(null);
-        sprintf(&fileNameBuffer[0],
-             "%s-%s.%s".ptr, "symbol".ptr, ctime(&now), "log".ptr);
+
+        import std.datetime : Clock;
+        auto dateTime = Clock.currTime();
+        auto dateStr = dateTime.toISOExtString();
+
+        sprintf(&fileNameBuffer[0], "%s-%s.%s".ptr, "symbol".ptr, dateStr.ptr, "log".ptr);
```
Would make it look like: 'symbol-2018-02-14T00:19:39.28761.log'

src/dmd/trace.d
```diff
@@ -0,0 +1,71 @@
...
+    ulong begin_mem;
+    ulong end_mem;
+
+    string kind;
+    string fn;
+}
+
+extern (C) __gshared uint dsymbol_profile_array_count;
+extern (C) __gshared SymbolProfileEntry* dsymbol_profile_array;
+enum dsymbol_profile_array_size = ushort.max * 1024; // 64 million entries should do, no ?
```
dkgroot: I had some trouble with memory consumption. When i changed this to
```enum dsymbol_profile_array_size = ushort.max * 512;```
I was still able to compile everytning and run the unittests. You might want to think about dynamically resizing on demand, instead of hardcoding this.

**Review by dkgroot — CHANGES_REQUESTED** [review]

File: src/dmd/mars.d [hunk 3]
```diff
@@ -923,6 +925,79 @@ private int tryMain(size_t argc, const(char)** argv)
...
+                if (c == ' ')
+                    timeString[len - 1] = '_';
+
+                if (c == '\r' || c == '\n')
+                {
+                    timeString[len - 1] = '\0';
+                    break;
+                }
+            }
+        }
```
dkgroot: I think this might be the nicer solution:
```
    import core.stdc.time : localtime, time, strftime;
    import core.stdc.stdio : printf;
    import core.thread : getpid;

    char[40] buff;
    auto now = time(null);
    strftime (buff.ptr, 40, "%Y-%m-%dT%H:%M:%S", localtime(&now));
    printf ("symbol_trace_%s_%d.log\n", buff.ptr, getpid());
```
Sorry for using phobos functions in my previous comment about this.
Note this solutiion also add's the PID, so that symbol trace files can be run in parallel, without clobering the output file.

dkgroot (changes requested):

WalterBright:
dmd has a builtin profiler. Why not use it?

UplinkCoder:
@WalterBright the built-in profiler will tell me which function is hot.
But it will not tell me which particular code caused the compiler to spend the time in those functions.
Using this profiling tool you can determine what parts of your code cause long compile times.
Rather then just knowing templateInstance.semantic takes all the time you can now know which particular templates take the time.

RazvanN7:
Is this PR still needed or can we close it?

UplinkCoder:
close it if you like/

RazvanN7:
Thanks for the reply. Reopen anytime you need it

UplinkCoder:
@RazvanN7 I cannot reopen myself since I am _still_ not part of dmd-team.

RazvanN7:
@UplinkCoder Even if you close it by yourself?

UplinkCoder:
if I close by myself I can reopen it.

PR #7793, opened 2018-01-27
add test for issue 16739 - switch ignores case

aG0aep6G:
The issue has apparently been fixed in 2.078.0.

### DIFF
--- a/test/runnable/testswitch.d
+++ b/test/runnable/testswitch.d
@@ -729,6 +729,20 @@ void test15538()
     assert(f15538(s) == 10); /* fails */
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=16739
+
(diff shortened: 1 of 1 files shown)


schveiguy:
Wait, this was a druntime bug, no?

schveiguy:
Actually, yes, it was a druntime bug. PR that fixed it was here: https://github.com/dlang/druntime/pull/1952

Not really an issue where this goes, but given that the behavior comes from druntime and not the compiler, it might be better that this test goes there.

aG0aep6G:
> Not really an issue where this goes, but given that the behavior comes from druntime and not the compiler, it might be better that this test goes there.

Sure thing. https://github.com/dlang/druntime/pull/2064

PR #7795, opened 2018-01-27
replace 'Loc loc' with 'const ref Loc loc'

WalterBright:
It's more efficient and safe this way. Should do it for all `Loc` parameters.


UplinkCoder:
I'd rather wonder in you need them at all.

JinShil (approved):

PR #7797, opened 2018-01-28
Convert Loc() to Loc.init

WalterBright:
This is necessary because `const ref Loc` needs an lvalue, and `Loc()` is an rvalue.

### DIFF
--- a/src/dmd/clone.d
+++ b/src/dmd/clone.d
@@ -211,7 +211,7 @@ extern (C++) FuncDeclaration buildOpAssign(StructDeclaration sd, Scope* sc)
     //printf("StructDeclaration::buildOpAssign() %s\n", sd.toChars());
     StorageClass stc = STC.safe | STC.nothrow_ | STC.pure_ | STC.nogc;
     Loc declLoc = sd.loc;
-    Loc loc = Loc(); // internal code should have no loc to prevent coverage
+    Loc loc; // internal code should have no loc to prevent coverage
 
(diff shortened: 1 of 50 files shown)


src/dmd/clone.d
```diff
@@ -211,7 +211,7 @@ extern (C++) FuncDeclaration buildOpAssign(StructDeclaration sd, Scope* sc)
     //printf("StructDeclaration::buildOpAssign() %s\n", sd.toChars());
     StorageClass stc = STC.safe | STC.nothrow_ | STC.pure_ | STC.nogc;
     Loc declLoc = sd.loc;
-    Loc loc = Loc(); // internal code should have no loc to prevent coverage
+    Loc loc; // internal code should have no loc to prevent coverage
```
JinShil: Can't this be removed and replace usages of this with `Loc.initial`?

JinShil (approved):

ibuclaw:
Wouldn't `Loc.unknown` or `Loc.builtin` be closer to the intent?

ibuclaw:
Then you can have the distinction between `loc.isValid()` and `loc.isUnknown()`/`loc.isBuiltin()`

src/dmd/clone.d
WalterBright: That was my first thought. But `Loc.initial` is a global, and `loc` can be 0-initialized if left alone, so I thought default initializing it would be better.

WalterBright:
dmd doesn't really have a use for having both concepts.

PR #7798, opened 2018-01-28
Fix issue 18322: __FILE_FULL_PATH__ doesn't work as a template parameter

marler8997:


test/runnable/imports/test18322import.d
```diff
@@ -0,0 +1,20 @@
...
+        assert(fileFullPath[1..3] == ":\\");
+        assert(templateFileFullPath[1..3] == ":\\");
+        enum lastPart = "test\\runnable\\test18322.d";
+    }
+    else
+    {
+        assert(fileFullPath[0] == '/');
+        assert(templateFileFullPath[0] == '/');
+        enum lastPart = "test/runnable/test18322.d";
+    }
```
wilzbach: Hmm, how about just checking for the end of the file name?
- I'm not sure whether LDC or GDC use a similar setup to run the test suite
- It would allow convenient execution of the test with `rdmd` or `dmd -i`

wilzbach: Also your test at https://github.com/dlang/dmd/blob/master/test/compilable/line.d doesn't uses the `test` directory.

src/dmd/expression.d
```diff
@@ -7073,9 +7073,17 @@ extern (C++) final class FileInitExp : DefaultInitExp
     override Expression resolveLoc(Loc loc, Scope* sc)
     {
         //printf("FileInitExp::resolve() %s\n", toChars());
-        const(char)* s = loc.isValid() ? loc.filename : sc._module.ident.toChars();
+        const(char)* s;
         if (subop == TOK.fileFullPath)
-            s = FileName.combine(sc._module.srcfilePath, s);
+        {
+            if (loc.isValid())
+                s = FileName.toAbsolute(loc.filename);
```
wilzbach: I'm not sure that `getcwd` will work in the general case. With `-i` and `-I` the file could be anywhere. As a start, we might want to add a test for more complicated cases?

marler8997: Ok, I've refactored the change a bit.  I've added a couple new fields to `Loc`.  One called `path` that can hold a path to `filename`.  Note that the path can be `null` if `filename` is absolute or relative to `CWD`.

I also added a field called `filenameType` which tracks whether `filename` came from an actual file or was generated from a `mixin` or `#line`.  This makes the code simpler in general for `__FILE_FULL_PATH__` because now it only generates an absolute path if the name represents a real file.

src/dmd/expression.d
```diff
@@ -7073,9 +7073,17 @@ extern (C++) final class FileInitExp : DefaultInitExp
...
+            else
+                s = FileName.combine(sc._module.srcfilePath, sc._module.srcfile.name.toChars());
+        }
+        else
+           s = loc.isValid() ? loc.filename : sc._module.ident.toChars();
```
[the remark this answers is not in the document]
marler8997: done

src/dmd/root/filename.d
```diff
@@ -98,6 +99,19 @@ nothrow:
         }
     }
 
+    /********************************
```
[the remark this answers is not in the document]
marler8997: not a problem, I've modified it to use the recommended style.

test/runnable/imports/test18322import.d
marler8997: ok, I removed the `test` part of the path.  I think having `runnable/<test-name>` is probably ok.

src/dmd/globals.h
```diff
@@ -295,18 +295,30 @@ struct DArray
     T *ptr;
 };
 
+enum FilenameType
```
timotheecour: why not C++11 enum class?
```c++
enum class FilenameType{
  default_, pound_line_, mixin_
};
```

src/dmd/globals.h
```diff
@@ -295,18 +295,30 @@ struct DArray
...
     unsigned charnum;
 
+    const char *path; // path to filename
+    FilenameType filenameType;
+
     Loc()
     {
         linnum = 0;
         charnum = 0;
         filename = NULL;
```
timotheecour: s/NULL/nullptr/g

src/dmd/globals.h
marler8997: I don't see this feature used anywhere else in the code base.

src/dmd/expression.d
marler8997: Ok, we've gone back to the original implementation.  @ibuclaw has assured me that `filename` will always either be absolute or relative to CWD.  He says this MUST be the case otherwise it will break the debugger.  Also, since I couldn't find a counter-example it sounds like a reasonable thing to assume for now.  If we ever do find a counter-example we can address it later (with a test!).

ibuclaw:
Is the enum still necessary after simplifying logic?

marler8997:
The requirement is to be able to tell when `Loc.filename` is an actual file, or it's been overriden via `#line` or mixin code.  Maybe there's another solution, but I thought the additional field in `Loc` was nice to have anyway.

marler8997:
Ok, I went with a different solution.  I've modified `__FILE_FULL_PATH__` to always return an absolute path whether or not `Loc.filename` is a mixin or a `#line`.  I don't this this will be a big deal since `#line` and `__FILE_FULL_PATH__` don't really make sense together no matter how you implement it.

test/runnable/test18322.d
```diff
@@ -0,0 +1,21 @@
+/*
+REQUIRED_ARGS: -Irunnable/imports
+COMPILED_IMPORTS: imports/test18322import.d
+PERMUTE_ARGS:
+*/
+import test18322import;
+void main(){
+    version(Windows)
```
timotheecour:
```
import std.path: buildPath;
auto filename = "runnable".buildPath("test18322.d");
```

test/runnable/imports/test18322import.d
```diff
@@ -0,0 +1,14 @@
+module test18322import;
+void fun(string templateFileFullPath = __FILE_FULL_PATH__,
+    string templateFile = __FILE__)(string expectedFilename, string fileFullPath = __FILE_FULL_PATH__)
+{
+    // make sure it is an absolute path
+    version(Windows)
```
timotheecour: `import std.path:isAbsolute;`

test/runnable/imports/test18322import.d
```diff
@@ -0,0 +1,14 @@
...
+        assert(fileFullPath[1..3] == ":\\");
+    else
+        assert(fileFullPath[0] == '/');
+
+    assert(templateFileFullPath == fileFullPath);
+    assert(fileFullPath[$ - expectedFilename.length .. $] == expectedFilename);
```
timotheecour: std.string: endsWith

marler8997:
Hmmm, it seems that alot of the DMD tests suite avoids using phobos.  I'm fine using it but I think this PR is more likely to be accepted if I try to mimic the way things are done in a typical test.  I'll leave it how it is unless a maintaner would prefer me to use the phobos functions.

src/dmd/parse.d
```diff
@@ -7270,11 +7270,11 @@ final class Parser(AST) : Lexer
             {
                 const(char)* srcfile = mod.srcfile.name.toChars();
                 const(char)* s;
-                if(loc.filename && !FileName.equals(loc.filename, srcfile)) {
-                    s = loc.filename;
-                } else {
+                if (loc.isValid())
+                    s = FileName.toAbsolute(loc.filename);
+                else
```
timotheecour: srcfile only used in this branch so should be declared here

marler8997: it's used on line 7273

marler8997:
Closing and reopening since DAutoTest failed to download the compiler.

marler8997:
Hmmm, closing and reopening didn't trigger DAutoTest to re-run...is there a way to tell it to retry?

It would be great if certain kinds of errors (like curl commands) could somehow cause the tester to retry later....

wilzbach:
> Hmmm, closing and reopening didn't trigger DAutoTest to re-run...is there a way to tell it to retry?

It restarts itself from time to time, but typically `git commit --amend && git push -f` is the only way to trigger a manual restart.
However, it seems it's passing now :)

marler8997:
This PR should be ready for merge unless there are any objections?

src/dmd/root/filename.d
```diff
@@ -98,6 +99,20 @@ nothrow:
...
+    /**
+    Return the given name as an absolute path
+
+    Params:
+        name = path
+        base = the absolute base to prefix name with if it is relative
+
+    Returns: name as an absolute path relative to base
+    */
+    static const(char)* toAbsolute(const(char)* name, const(char)* base = getcwd(null, 0))
```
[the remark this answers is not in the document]
marler8997: done

JinShil:
I restarted Jenkins

marler8997:
It appears there are no objections, merge?

src/dmd/expression.d
```diff
@@ -1823,7 +1823,7 @@ extern (C++) abstract class Expression : RootObject
     }
 
     /****************************************
-     * Resolve __FILE__, __LINE__, __MODULE__, __FUNCTION__, __PRETTY_FUNCTION__ to loc.
+     * Resolve __FILE__, __LINE__, __MODULE__, __FUNCTION__, __PRETTY_FUNCTION__, __FILE_FULL_PATH__ to loc.
```
JinShil: I'd prefer this if it was named `__ABSOLUTE_FILE_PATH__`

timotheecour:
thats orthogonal to this bug fixing pr though

JinShil (approved):

timotheecour:
ping?

marler8997:
can I get a merge up in here?

test/runnable/test18322.d
```diff
@@ -0,0 +1,21 @@
...
+        auto sep = "\\";
+    else
+        auto sep = "/";
+
+    auto filename = "runnable" ~ sep ~ "test18322.d";
+
+    fun(filename);
+    mixin(`fun(filename ~ "-mixin-16");`);
```
wilzbach: That was unexpected for me. I didn't expected that `__FILE_FULL_PATH__` changes for mixins to sth. like `test18322.d-mixin-16`, but it's really the current behavior - https://run.dlang.io/is/F4ixWD

wilzbach (approved):
OK, I'm putting this on the 72h merge if no further objection are raised queue.

test/runnable/test18322.d
marler8997: Hmm, yeah this is an interesting case.  I would expect `__FILE__` to match `__FILE_FULL_PATH__` though, and `__FILE__` goes to the mixin.  I think if we want to change it we should probably change both...

src/dmd/parse.d
```diff
@@ -7270,11 +7270,11 @@ final class Parser(AST) : Lexer
...
                     s = FileName.combine(mod.srcfilePath, srcfile);
```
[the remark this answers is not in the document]
marler8997: I wasn't actually able to create a test that exercised this code path.  I added some debugging to see if any files ever had `loc.invalid()` as false, and it was only with `object.d`/`__entrypoint.d` and also a couple phobos modules like `std.array` but the code was at line 0.  I'm guessing it was parsing code that might have been generated internally (not from a file).  So for now I've just added an assert that prints a nice error message if this code path ever does get taken.

marler8997:
Note: I made an additional change.  I realized that the new field I added a while back `Module.srcfilePath` was actually unnecessary.  Given @ibuclaw's new information that relative filenames will always be relative to CWD (and I verified this in all the code paths that I saw), this field was rendered useless.  So I've removed it in this PR as well.  Note that this additional change affected about a dozen lines of code.

src/dmd/dmodule.d
```diff
@@ -59,9 +59,8 @@ version(Windows) {
  * Returns:
  *      NULL if it's not different from filename.
  */
-private const(char)* lookForSourceFile(const(char)** path, const(char)* filename)
+private const(char)* lookForSourceFile(const(char)* filename)
```
ibuclaw: You removed a parameter, but didn't update the documentation.

ibuclaw (changes requested):

src/dmd/globals.d
```diff
@@ -378,6 +378,8 @@ alias d_uns64 = uint64_t;
 // file location
 struct Loc
 {
+    // Note: it is assumed that filename is always either absolute or
+    //       relative to CWD
     const(char)* filename;
```
ibuclaw: Maybe you should make this a comment against the field, such as:
```
const(char)* filename;  // either absolute or relative to cwd
```
Make it a definitive comment, rather than a presumptuous one.

ibuclaw:
Fine after nits.

src/dmd/dmodule.d
marler8997: nice catch...fixed

src/dmd/globals.d
marler8997: ok

ibuclaw (approved):

PR #6561, opened 2017-02-23
generate TypeInfo for structs only on demand

WalterBright:
The original code always generates them when a struct declaration was found. But they were also generated again by other object modules whenever the typeinfo was needed. This relied on COMDATs to cull the duplicates. But with such, the typeinfo does not need to be generated just because the struct is declared - only when it is actually used.

PR #7799, opened 2018-01-28
Don't require TypeInfo in a minimal runtime if it's not needed

JinShil:
This is a followup to the following PRs

 * #7786 for `Throwable`

It is to `TypeInfo` what the above PRs were for `ModuleInfo` and `Throwable`

With this PR, it is now possible to have the same functionality as -betterC (without the -betterC compiler switch) by simply adding the following minimal runtime for executables...

```D
module object;

private alias extern(C) int function(char[][] args) MainFunc;
private extern (C) int _d_run_main(int argc, char** argv, MainFunc mainFunc)
{
    return mainFunc(null);  // assuming `void main()`
}
```

... or the following minimal runtime for libraries:

```D
module object;
```

For libraries it would be nice to not require an empty object.d file, but I'm leaving that for another potential PR.      UPDATE: See  #7825

Again, this will be of interest to users wishing to incrementally or partially port D to new platforms, users hoping to target resource constrained platforms, and users wishing to use D as a library from other languages without druntime.

### DIFF
--- a/src/dmd/e2ir.d
+++ b/src/dmd/e2ir.d
@@ -823,10 +823,10 @@ elem *sarray_toDarray(const ref Loc loc, Type tfrom, Type tto, elem *e)
 /************************************
  */
 
-elem *getTypeInfo(Type t, IRState *irs)
+elem *getTypeInfo(Loc loc, Type t, IRState *irs)
 {
     assert(t.ty != Terror);
-    genTypeInfo(t, null);
+    genTypeInfo(loc, t, null);
     elem *e = el_ptr(toSymbol(t.vtinfo));
     return e;
 }
@@ -865,17 +865,17 @@ StructDeclaration needsDtor(Type t)
  * Set an array pointed to by eptr to evalue:
  *      eptr[0..edim] = evalue;
  * Params:
- *      eptr =    where to write the data to
- *      edim =    number of times to write evalue to eptr[]
- *      tb =      type of evalue
- *      evalue =  value to write
- *      irs =     context
- *      op =      TOK.blit, TOK.assign, or TOK.construct
+ *      exp    = the expression for which this operation is performed
+ *      eptr   = where to write the data to
+ *      edim   = number of times to write evalue to eptr[]
+ *      tb     = type of evalue
+ *      evalue = value to write
+ *      irs    = context
+ *      op     = TOK.blit, TOK.assign, or TOK.construct
  * Returns:
  *      created IR code
  */
-
-private elem *setArray(elem *eptr, elem *edim, Type tb, elem *evalue, IRState *irs, int op)
+private elem *setArray(Expression exp, elem *eptr, elem *edim, Type tb, elem *evalue, IRState *irs, int op)
 {
     assert(op == TOK.blit || op == TOK.assign || op == TOK.construct);
     const sz = cast(uint)tb.size();
@@ -949,7 +949,7 @@ Lagain:
                     r = (op == TOK.construct) ? RTLSYM_ARRAYSETCTOR : RTLSYM_ARRAYSETASSIGN;
                     evalue = el_una(OPaddr, TYnptr, evalue);
                     // This is a hack so we can call postblits on const/immutable objects.
-                    elem *eti = getTypeInfo(tb.unSharedOf().mutableOf(), irs);
+                    elem *eti = getTypeInfo(exp.loc, tb.unSharedOf().mutableOf(), irs);
                     elem *e = el_params(eti, edim, evalue, eptr, null);
                     e = el_bin(OPcall,TYnptr,el_var(getRtlsym(r)),e);
                     return e;
@@ -1323,7 +1323,7 @@ elem *toElem(Expression e, IRState *irs)
             //printf("TypeidExp.toElem() %s\n", e.toChars());
             if (Type t = isType(e.obj))
             {
-                result = getTypeInfo(t, irs);
+                result = getTypeInfo(e.loc, t, irs);
                 result = el_bin(OPadd, result.Ety, result, el_long(TYsize_t, t.vtinfo.offset));
                 return;
             }
@@ -1722,7 +1722,7 @@ elem *toElem(Expression e, IRState *irs)
                 else
                 {
                     // call _d_newitemT(ti)
-                    e = getTypeInfo(ne.newtype, irs);
+                    e = getTypeInfo(ne.loc, ne.newtype, irs);
 
                     int rtl = t.isZeroInit(Loc.initial) ? RTLSYM_NEWITEMT : RTLSYM_NEWITEMIT;
                     ex = el_bin(OPcall,TYnptr,el_var(getRtlsym(rtl)),e);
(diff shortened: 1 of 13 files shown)


src/dmd/typinf.d
```diff
@@ -29,9 +29,16 @@ import dmd.visitor;
 extern (C++) void genTypeInfo(Type torig, Scope* sc)
 {
     //printf("Type::genTypeInfo() %p, %s\n", this, toChars());
+
+    if (!global.params.useTypeInfo)
```
JinShil: This block of code was modified to make it consistent with the following in statementsem.d

https://github.com/dlang/dmd/blob/ec4534ec63958b17167d636cd8b98824ce52f7d3/src/dmd/statementsem.d#L3626-L3636

src/dmd/typinf.d
```diff
@@ -29,9 +29,16 @@ import dmd.visitor;
...
+    {
+        torig.error(Loc.initial, "`TypeInfo` cannot be used with -betterC");
+        fatal();
+    }
```
JinShil: This block of code was modified to make it consistent with the following in statementsem.d

https://github.com/dlang/dmd/blob/ec4534ec63958b17167d636cd8b98824ce52f7d3/src/dmd/statementsem.d#L3626-L3636

test/compilable/minimal.d
```diff
@@ -2,7 +2,10 @@
 // POST_SCRIPT: compilable/extra-files/minimal/verify_symbols.sh
 // REQUIRED_ARGS: -c -defaultlib= runnable/extra-files/minimal/object.d
 
-// This test ensures an empty main built with a minimal runtime does not generate
-// ModuleInfo and exception handling code
+// This test ensures an empty main with a struct, built with a minimal runtime,
+// does not generate ModuleInfo or exception handling code, and does not
+// require TypeInfo
```
wilzbach: Don't you have to modify `verify_symbols` to check for `TypeInfo`?

JinShil: Yeah, best to be thorough.  Done.

JinShil:
Ping @WalterBright, @ZombineDev, @ibuclaw... anyone, really.

jacob-carlborg (approved):

src/dmd/typinf.d
ibuclaw: Can we have a valid line for the error?

ibuclaw: Does this error apply for both explicit and implicit uses of RTTI?

ibuclaw (changes requested):

src/dmd/typinf.d
JinShil: > Does this error apply for both explicit and implicit uses of RTTI?

It applies to anything that ends up calling `genTypeInfo`, which, under the current implementation could be implicit or explicit uses of `TypeInfo`.  For example, trying to do a string concatenation will eventually call runtime hooks like `_d_arraycatnTX` which takes `TypeInfo` as one of its arguments.  See https://issues.dlang.org/show_bug.cgi?id=18312

IMO all of those runtime hooks should be changed to templates, that don't require `TypeInfo`, so we shouldn't need `TypeInfo` for string concatenation as long as the runtime hooks are implemented.  That's something I hope to champion after this PR (maybe 1 or 2 more) are merged.

TL;DR: Yes implicit and explicit uses of `TypeInfo` will trigger the error, but by replacing runtime hooks with templates (later) we may be able to eliminate such errors.

ibuclaw: Ok, but I'd still feel better if explicit uses got a file/line location to highlight where you're using RTTI.

JinShil: By "explicit uses" are you referring to `typeid`?  That's the only case I can think of where `TypeInfo` is explicitly used, and even then, it's indirect.  Please let me know if you have something else in mind.

rainers: I think showing the location of implicit uses is even more desirable, as you cannot grep for these as with explicit uses of `typeid`. That would mean to pass the location of an expression via semanticTypeInfo and getTypeInfoType.

That will probably also be helpful when generating type info more lazily as in https://github.com/dlang/dmd/pull/6561

JinShil: OK, I'll explore this.  Marking this as WIP while I try.

JinShil:
I add a `Loc` parameter to `genTypeInfo` for reporting line numbers in the error messages.  This increased the diff, of course.  

I also add DDoc header comments to the function signatures I modified, and a couple more tests to increase coverage.

There is somewhat of a drawback to this, however.  For example, the following code will produce an error message about `TypeInfo` when it is not apparent, at all, why `TypeInfo` is needed.

```D
void test()
{
    string s = "abc"
    s = "(" ~ s ~ ")";  // main.d(8): Error: `object.TypeInfo` was not found
}
```

That is not actually a problem with this PR, but a general problem with D's ubiquitous use of runtime type information (i.e. `TypeInfo`) for things that are known at compile-time.  As I mentioned previously, I intend to champion an effort replace such uses of `TypeInfo` with templates, which should actually eliminate all such errors, and produce messages that are more meaningful.

cc @ibuclaw @rainers

src/dmd/e2ir.d
```diff
@@ -949,7 +949,8 @@ Lagain:
                     r = (op == TOK.construct) ? RTLSYM_ARRAYSETCTOR : RTLSYM_ARRAYSETASSIGN;
                     evalue = el_una(OPaddr, TYnptr, evalue);
                     // This is a hack so we can call postblits on const/immutable objects.
-                    elem *eti = getTypeInfo(tb.unSharedOf().mutableOf(), irs);
+                    AggregateDeclaration ad = cast(AggregateDeclaration)tb;
+                    elem *eti = getTypeInfo(ad.loc, tb.unSharedOf().mutableOf(), irs);
```
rainers: This location is the one of the aggregate declaration, not the using expression. For completeness, setArray should probably receive the expression to forward it here.

src/dmd/typinf.d
```diff
@@ -24,14 +24,26 @@ import dmd.mtype;
...
+    if (!global.params.useTypeInfo)
+    {
+        torig.error(loc, "`TypeInfo` cannot be used with -betterC");
+        fatal();
+    }
+
     if (!Type.dtypeinfo)
     {
-        torig.error(Loc.initial, "TypeInfo not found. object.d may be incorrectly installed or corrupt, compile with -v switch");
+        torig.error(loc, "`object.TypeInfo` was not found");
```
rainers: Maybe add "but implicitly used" to the error message.

rainers (commented):
Apart from nits LGTM

src/dmd/e2ir.d
JinShil: Done.  Thanks for the help!

src/dmd/typinf.d
JinShil: Done.

rainers (approved):

WalterBright:
This produced a regression https://issues.dlang.org/show_bug.cgi?id=18905

PR #7801, opened 2018-01-28
[Trivial] Fix test/fail_compilation/impconv.d (Issue 17437 / PR#6831)

dkgroot:
Improved error message was implemented causing this check to fail, because of the new error message content. 
Note: This is test is excluded on most platforms (not sure why), causing it to go unnoticed (Not sure if this should be corrected/changed).


wilzbach:
See https://github.com/dlang/dmd/pull/6831

JinShil (approved):

PR #7806, opened 2018-01-29
Fix Issue 17663 - header generation (-H) is broken with public override of private default

RazvanN7:
Whenever there's a block preceded by a visibility attribute and the block contains a single declaration : 

```d
private:
public struct Export {}
```

the generated header is going to be `private public struct Export {} `. This happens because
the header generation does not check for consecutive ProtDeclaration AST nodes.

For blocks that contain more than 1 declaration, the generated header is fine because it is treated
as a block:

```d
private:
public struct Export {}
private struct Import {}
```
Will result in the generated header file:

```d
private
{
    public struct Export {}
    private struct Import {}
}
```

I fixed the issue by adding a check that ensures that if there are 2 ProtDeclaration AST nodes consecutively, the second one is skipped.

### DIFF
--- a/src/dmd/hdrgen.d
+++ b/src/dmd/hdrgen.d
@@ -1281,7 +1281,11 @@ public:
     {
         protectionToBuffer(buf, d.protection);
         buf.writeByte(' ');
-        visit(cast(AttribDeclaration)d);
+        AttribDeclaration ad = cast(AttribDeclaration)d;
+        if (ad.decl.dim == 1 && (*ad.decl)[0].isProtDeclaration)
+            visit(cast(AttribDeclaration)(*ad.decl)[0]);
+        else
+            visit(cast(AttribDeclaration)d);
     }
 
     override void visit(AlignDeclaration d)
--- a/src/dmd/attrib.d
+++ b/src/dmd/attrib.d
@@ -590,6 +590,11 @@ extern (C++) final class ProtDeclaration : AttribDeclaration
         return buf.extractString();
     }
 
+    override final inout(ProtDeclaration) isProtDeclaration() inout
+    {
+        return this;
+    }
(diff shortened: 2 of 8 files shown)


src/dmd/hdrgen.d
```diff
@@ -1281,7 +1281,11 @@ public:
     {
         protectionToBuffer(buf, d.protection);
         buf.writeByte(' ');
-        visit(cast(AttribDeclaration)d);
+        AttribDeclaration ad = cast(AttribDeclaration)d;
+        if (ad.decl.dim == 1 && strcmp((*ad.decl)[0].kind, "protection attribute") == 0)
```
EyalIO: Is there no enum or other way to compare rather than comparing strings?

RazvanN7: Nope. I looked for a `isProtDeclaration()` function, but there is no such thing.

jacob-carlborg: How about adding that function? Or can you visit `(*ad.decl)[0]`?

RazvanN7: Visiting is not a possibility since that will make it visit ProtDeclaration again and print `private public`. I will add the function.

RazvanN7: Is this ok ?

stefan-koch-sociomantic: you should be able to check the tok

jacob-carlborg: @stefan-koch-sociomantic there's no token associated with a `AttribDeclaration`. It's only expressions that stores a token. I would be nice if all AST nodes store a kind or type, similar to the token in expression. So it's possible to use the same way for all AST nodes to identify subclasses.

PR #7807, opened 2018-01-29
Experiment with setting after_n_builds to 3 to see CodeCov status updates again

wilzbach:
For some yet to be determined reason CodeCov doesn't show up in the status update.
However, the reports are still properly processed and the [browser extension](https://github.com/codecov/browser-extension) still works nicely.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -6,7 +6,7 @@ codecov:
     # We don't want to wait for the CodeCov report
     # See https://github.com/codecov/support/issues/312
     require_ci_to_pass: false
-    after_n_builds: 1  # send notifications after the first upload
+    after_n_builds: 3  # send notifications after all CircleCi builds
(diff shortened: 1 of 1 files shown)


wilzbach:
FYI: From the CodeCov support:

> I see an issue with the bot that cannot create statuses on the project
> I got a 403 using the bot when trying to post the status
> Your admin will need to add the project to the Codecov integration list.
> Can you add the Codecov GitHub App?
> ...
> I see it is installed, but maybe the project is not enabled.
> It may have worked before because the bot on file had authentication. But it does not now....

(I have pinged @andralex and @CyberShadow)

wilzbach:
Thanks @CyberShadow for the amazingly fast reaction and action :)

andralex:
Awesome you guys!

PR #7809, opened 2018-01-30
Comments/C++headers/Changelog entry for lambda comparison

RazvanN7:
Part 2 of #7484 

cc @ibuclaw @wilzbach @UplinkCoder

### DIFF
--- a/changelog/lambdacomp.dd
+++ b/changelog/lambdacomp.dd
@@ -1,0 +1,23 @@
+Lambda comparison using __traits(isSame, ...)
+
+It is now possible to compare two lambda functions, under certain
+constraints, using __traits(isSame, lamda1, lambda2). In order to
+correctly compare two lambdas, the following conditions must be
+satisfied:
+
+$(UL
+    $(LI the lambda function arguments must not have a template instantiation
+    as an explicit argument type. Any other argument types (basic,
+    user-defined, template) are supported.)
+
+    $(LI the lambda function body must contain a single expression (no return statement)
+    which contains only numeric values, manifest constants, enum values and arguments.
+    If the expression contains local variables, function calls or return statements,
+    the function is considered uncomparable.)
+)
+
+These limitations might be lifted in the next release version.
+
+Whenver a lambda is considered uncomparable, the __traits(isSame, ...) expression
+in which it's used will return false, no matter what other lambda is used in
+the comparison.
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -2946,9 +2946,15 @@ extern (C++) final class FuncLiteralDeclaration : FuncDeclaration
     TOK tok;        // TOK.function_ or TOK.delegate_
     Type treq;      // target of return type inference
 
+    /**
+     * The serialization of this. It is used to compare
+     * lambda functions.
+     * != null only if tok == TOK.reserved.
+     */
+    const(char)* serialization;
+
     // backend
     bool deferToObj;
-    const(char)* serialization;
 
     extern (D) this(const ref Loc loc, const ref Loc endloc, Type type, TOK tok, ForeachStatement fes, Identifier id = null)
     {
--- a/src/dmd/lambdacomp.d
+++ b/src/dmd/lambdacomp.d
@@ -1,3 +1,20 @@
+/***
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * This modules implements the serialization of a lambda function. The serialization
+ * is computed by visiting the abstract syntax subtree of the given lambda function.
+ * The serialization is a string which contains the type of the parameters and the
+ * string represantation of the lambda expression.
+ *
+ * Copyright:   Copyright (C) 1999-2018 by The D Language Foundation, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/lamdbacomp.d, _lambdacomp.d)
+ * Documentation:  https://dlang.org/phobos/dmd_lambdacomp.html
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/lambdacomp.d
+ */
+
 module dmd.lambdacomp;
 
 import core.stdc.stdio;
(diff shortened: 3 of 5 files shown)


src/dmd/lambdacomp.d
```diff
@@ -108,35 +154,43 @@ extern (C++) class SerializeVisitor : SemanticTimeTransitiveVisitor
...
             {
+                // If it's a VarDeclaration
                 if (auto v = s.isVarDeclaration)
                 {
+                    // It must be a manifest constant
                     if (v.storage_class & STC.manifest)
                     {
                         v.getConstInitializer.accept(this);
                     }
+                    // Otherwise, the function is uncomparable
```
stefan-koch-sociomantic: This flow is a little confusing.
please let the buf.reset be a fall-through case.

src/dmd/lambdacomp.d
```diff
@@ -108,35 +154,43 @@ extern (C++) class SerializeVisitor : SemanticTimeTransitiveVisitor
...
                     else
                         buf.reset();
                 }
+                // Or it may be an enum
```
stefan-koch-sociomantic: redundant comment

changelog/lambdacomp.dd
```diff
@@ -0,0 +1,10 @@
+Lambda comparison using __traits(isSame, ...)
+
+It is now possible to compare two lambda functions, under certain
+constraints, using __traits(isSame, lamda1, lambda2). In order to
+correctly compare two lambdas, the following conditions must be
+satisfied:
+
+-> the lambda function arguments must not have a template instantiation as an explicit argument type. Any other argument types (basic, user-defined, template) are supported.
+
+-> the lambda function body must contain a single expression (no return statement) which contains only numeric values, manifest constants, enum values and arguments. If the expression contains local variables, function calls or return statements, the function is considered uncomparable.
```
stefan-koch-sociomantic: more rational wanted.
why are the limitations the way they are ?

stefan-koch-sociomantic (changes requested):
it's better but could be improved.

meta concern, 
why would you serialize to a string and then compare that ?
it's expensive.

RazvanN7:
>why would you serialize to a string and then compare that ?

To do parameter renaming and hopefully, later on, do some other computation (e.g. to be able to distinguish between `a+b` and `b+a` ). What is the alternative you are thinking?

changelog/lambdacomp.dd
RazvanN7: That's the point I got to. Later on, the limitations will be lifted (where possible).

src/dmd/lambdacomp.d
RazvanN7: it is a fall-through case, I just need it to be reset inside the first if-block if the var declaration isn't a manifest constant. I don't know now to make differently without adding a label, which in my opinion would be worse.

src/dmd/lambdacomp.d
```diff
@@ -17,51 +29,81 @@ import dmd.statement;
 import dmd.tokens;
 import dmd.visitor;
 
+/**
+ * This enum is used to report the type of the expression
+ * found when visiting the AST nodes to higher levels of the
+ * AST.
+ */
 enum ExpType
```
jacob-carlborg: Seems like this can be private.

src/dmd/func.d
```diff
@@ -2946,9 +2946,14 @@ extern (C++) final class FuncLiteralDeclaration : FuncDeclaration
     TOK tok;        // TOK.function_ or TOK.delegate_
     Type treq;      // target of return type inference
 
+    /**
+     * The serialization of this.
+     * != null only if tok == TOKreserved.
+     */
+    const(char)* serialization;
```
jacob-carlborg: Could have a better description of what it's used for.

changelog/lambdacomp.dd
```diff
@@ -0,0 +1,17 @@
...
+-> the lambda function arguments must not have a template instantiation
+as an explicit argument type. Any other argument types (basic,
+user-defined, template) are supported.
+
+-> the lambda function body must contain a single expression (no return statement)
+which contains only numeric values, manifest constants, enum values and arguments.
+If the expression contains local variables, function calls or return statements,
+the function is considered uncomparable.
+
+These limitation will be lifted in the next release version.
```
jacob-carlborg: > These limitation will be lifted in the next release version.

I would recommend avoiding being this specific. There are not guarantees that those changes will make it into the next release. I would say something like: "These limitations might be lifted in the future".

jacob-carlborg (changes requested):

jacob-carlborg:
* What happens if two lambdas are not comparable? Will compiler give an error or will you get back false in the comparison?

* Does `SerializeVisitor` need to be exposed in a C++ header?

src/dmd/lambdacomp.d
stefan-koch-sociomantic: it could also be auto v = s.isVarDeclaration if (v && v.storage_class & STC.manifest)

In general if (a) {if (b) {c}} maps exactly to if (a && b) {c}

src/dmd/lambdacomp.d
```diff
@@ -17,51 +29,81 @@ import dmd.statement;
 import dmd.tokens;
 import dmd.visitor;
 
+/**
+ * This enum is used to report the type of the expression
```
WalterBright: `This enum is used to report` can be excised with no change in meaning to the sentence.

src/dmd/lambdacomp.d
```diff
@@ -17,51 +29,81 @@ import dmd.statement;
...
+ * found when visiting the AST nodes to higher levels of the
+ * AST.
```
WalterBright: I don't know what `visiting the AST nodes to the higher levels of the AST` means.

src/dmd/lambdacomp.d
```diff
@@ -17,51 +29,81 @@ import dmd.statement;
...
 enum ExpType
 {
     None,
     EnumDecl,
     Arg
 }
 
+/**
+ * The serialize visitor visits a FuncLiteralDeclaration
+ * AST node, serializing the implementation of the lambda
```
WalterBright: What does `serialization of the implementation` mean? Is a string produced? A hash?

wilzbach:
> Does SerializeVisitor need to be exposed in a C++ header?

I don't see any reason why it should. It's more or less a "private function of the DMD frontend implemented as a visitor" for the other backends. They only care about the IR.

> What happens if two lambdas are not comparable? Will compiler give an error or will you get back false in the comparison?

`false` - it's the `__traits(isSame)`
In fact at the moment the comparability is severely limited and returning `false` if unsure is a very reasonable decision.

I'm going to merge this now, s.t. it appears on the 2.079 changelog.
Great work again @RazvanN7!

PR #7810, opened 2018-01-30
Fix issue 18335 - The D_ObjectiveC version identifier is not printed in verbose mode

jacob-carlborg:
The `D_ObjectiveC` version identifier is set when the `Objc` is initialized. That means that the predefined version identifiers need to be printed after `Objc` has been initialized.

### DIFF
--- a/test/runnable/test18335.sh
+++ b/test/runnable/test18335.sh
@@ -1,0 +1,10 @@
+#!/usr/bin/env bash
+
+output_file="${RESULTS_DIR}/runnable/$(basename $0 .sh)"
+set -ueo pipefail
+
+if [ "${OS}" == "osx" ] && [ "${MODEL}" == "64" ]; then
+    echo "void main(){}" | "${DMD}" -o- -v - | grep predefs | grep -q "D_ObjectiveC"
+fi
+
+echo Success > "${output_file}"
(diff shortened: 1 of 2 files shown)


JinShil (approved):

wilzbach:
Shouldn't this have a test?
After all, we don't want to reintroduce bugs ...

Disclaimer: I don't have OSX nor used the `ObjectiveC`, but something like this should work:

```sh
#!/usr/bin/env bash

name=`basename $0 .sh`
dir=${RESULTS_DIR}/compilable

set -ueo pipefile

if [ "${OS}" == "osx" ]; then
    echo "void main(){}" | dmd -v - | head -n1 | grep -v "Objective_C"
fi

echo Success >${dir}/`basename $0`.out
```

jacob-carlborg:
> Shouldn't this have a test?

Yes.

jacob-carlborg:
I've added a test.

wilzbach (approved):
Thanks!

test/runnable/test18335.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
```
wilzbach:
```
/bin/bash: ./runnable/test18335.sh: Permission denied
```

jacob-carlborg: Of course, I always forget that 😞.

PR #7811, opened 2018-01-30
[WIP] function argument stringification: __ARG_STRING__ (replaced by 7821)

timotheecour:
Stringify caller's function argument, analog to C's `#` stringification macro (except it's type safe).
A main use case is to allow finally writing nice logging and debug tools.

Purely library based alternatives all have drawbacks, see https://github.com/dlang/dmd/pull/7811#issuecomment-361772960 for example.

It also works with UFCS.

Eg:
```d
string log(T)(T a, string name=__ARG_STRING__!a)
{
  import std.conv;
  return text(name, ": ", a);
}
assert(log(1+2) == "1 + 2: 3");
int x=3;
assert(log(x) == "x: 3");
```

NOTE: 
* code calls `earg.stringified = earg.toChars;` for all function arguments only when attempting to match (via `callMatch`) against a function that has `__ARG_STRING__`. Caveat: `enum folding` will happen before that, so `enum x=3; assert(log(x) == "3: 3");` instead of `assert(log(x) == "x: 3");` ; but the behavior is correct for non-enum's (eg `static const`)
* currently only allows binding to a function parameter (not template parameter, and not variadic parameters); this could be lifted in future PR without breaking backward compatibility

EDIT: closing in favor of https://github.com/dlang/dmd/pull/7821


wilzbach:
Wow, I wanted to add this to DMD to, but you beat me to it.
However, I can add further motivation for this. With `__ARG_NAME__` we can finally write __nice logging and debug tools__. 
Consider, for example, my stalled PR to Phobos (https://github.com/dlang/phobos/pull/4318) for adding a user-friendly dump method, which never went anywhere because the best we can do at the moment was:

```
dump!(x, y) // x == 5, y == 6
dump(x, y) // arg.1= 5, arg.2 = 6
```

Both have severe downsides,  `__ARG_NAME__`  is ideal for this use case.
Other potential areas of use are of course at least:
- logging (e.g. stdx.logger, vibe.core.logger, ..)
- better exception messages ("`myFancyVar` is null" is a lot more helpful than "`r` is null")

timotheecour:
yes, that's exactly of of the main use cases i had in mind :) will update commit msg to make it clearer

jacob-carlborg:
Please no. This is yet another hack because we don't have AST macros. Yes, I know that we most likely will not get those.

changelog/argname.dd
```diff
@@ -0,0 +1,13 @@
+Stringify caller's function argument, analog to C's `#` stringification macro (except it's type safe).
+
+NOTE: it also works with UFCS.
+
+Eg:
+---
+string log(T)(T a, string name=__ARG_NAME__!a)
```
jacob-carlborg: The code implements `__ARG_STRING__`.

src/dmd/expression.h
```diff
@@ -129,6 +129,7 @@ class Expression : public RootObject
     TOK op;                     // to minimize use of dynamic_cast
     unsigned char size;         // # of bytes in Expression so we can copy() it
     unsigned char parens;       // if this is a parenthesized expression
+    const(char)* stringified;  // for function calls (in case of __ARG_NAME__)
```
jacob-carlborg: Same here, the code implements `__ARG_STRING__`.

test/runnable/testargnames.d
```diff
@@ -0,0 +1,62 @@
...
+  (1+1).fun_UFCS("1 + 1");
+
+  fun_template(1+3, `1 + 3`);
+
+  fun1(a+a+a, `a + a + a`);
+
+  // Checks that no constant folding happens, cf D20180130T161632.
+  fun1(1+1+2, `1 + 1 + 2`);
+  enum t=44;
+  fun1(t+t+t, `t + t + t`); // want: t+t+t
```
jacob-carlborg: Operators should have spaces on each side.

jacob-carlborg (changes requested):

jacob-carlborg:
I would like to see examples/tests what happens in the following scenarios:

* `__ARG_STRING__` is used in a non-parameter context
* More complex expressions
* Expressions with user defined types
* Passing a type to `__ARG_STRING__`
* Using `__ARG_STRING__` as the default value of a template argument

src/dmd/expression.d
```diff
@@ -7096,6 +7104,42 @@ extern (C++) final class FileInitExp : DefaultInitExp
     }
 }
 
+/***********************************************************
+ */
+extern (C++) final class ArgnameInitExp : DefaultInitExp
```
jacob-carlborg: Now there are three different names in play here: `ArgnameInitExp`, `__ARG_STRING__` and `argumentName`. Pick one of them and use it consistently. Please also avoid abbreviations.

src/dmd/expressionsem.d
```diff
@@ -598,8 +598,44 @@ private bool functionParameters(Loc loc, Scope* sc, TypeFunction tf, Type tthis,
...
+                {
+                  if(argexp.ident is null)
+                  {
+                    error(argexp.loc, "argexp.ident null");
+                    return true;
+                  }
+                  assert(nargs>0);
+                  bool found=false;
+                  // TODO: should Parameters._foreach be used?
+                  for(int u=0;u<nparams;u++)
```
jacob-carlborg: Use `foreach` instead of `for`.

src/dmd/expressionsem.d
```diff
@@ -2767,6 +2803,12 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             return;
         }
 
+        for(int i=0;i<(exp.arguments ? exp.arguments.dim : 0) ; i++){
```
jacob-carlborg: `asDArray` can be used instead.

jacob-carlborg (changes requested):
There are many issues with the code style. These are the corrections:

* Space after `if`
* Spaces around operators
* `else` on its own line
* `for` -> `foreach`
* Four spaces for indentation

jacob-carlborg:
There are commits from other people that are already merged.

changelog/argname.dd
timotheecour: done

src/dmd/expression.h
timotheecour: done

test/runnable/testargnames.d
timotheecour: done

src/dmd/tokens.d
```diff
@@ -447,6 +448,7 @@ extern (C++) struct Token
         TOK.overloadSet: "__overloadset",
         TOK.file: "__FILE__",
         TOK.fileFullPath: "__FILE_FULL_PATH__",
+        TOK.argumentName: "__ARG_STRING__",
```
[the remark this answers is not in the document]
timotheecour: done

src/dmd/expression.d
timotheecour: changed to more consistent naming:
```
__ARG_STRING__
argString
ArgStringInitExp
isArgStringInitExp
resolveArgString
```
I did use commonly used abbreviation though, `__ARGUMENT_STRING__` would be too long IMO (and some ppl had suggesed ARG also)

src/dmd/expressionsem.d
timotheecour: done

src/dmd/expressionsem.d
timotheecour: not used anymore

src/dmd/parse.d
```diff
@@ -6249,7 +6253,7 @@ final class Parser(AST) : Lexer
 
     /*****************************************
      * Parses default argument initializer expression that is an assign expression,
-     * with special handling for __FILE__, __FILE_DIR__, __LINE__, __MODULE__, __FUNCTION__, and __PRETTY_FUNCTION__.
+     * with special handling for __FILE__, __FILE_DIR__, __LINE__, __MODULE__, __FUNCTION__, __PRETTY_FUNCTION__, __ARG_NAMES__.
```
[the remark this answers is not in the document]
timotheecour: done

timotheecour:
> There are commits from other people that are already merged.

fixed

timotheecour:
>  __ARG_STRING__ is used in a non-parameter context
I am disallowing this use case; __ARG_STRING__ is useful mostly as a default function argument. If you have a good use case for that (which one, I'd be curious?) we could always add that later.

jacob-carlborg:
> I am disallowing this use case; ARG_STRING is useful mostly as a default function argument. If you have a good use case for that (which one, I'd be curious?) we could always add that later.

I'm not arguing for or a against, you just need a test that enforces the current behavior.

> as noted in PR msg, this is on TODO list but could be done in subsequent PR while being backward compatible

Same here, test.

jacob-carlborg:
I don't want to completely ruin this PR, but what about a `__traits` instead of this magic template or whatever it is?

src/dmd/tokens.d
```diff
@@ -302,7 +303,7 @@ enum TOK : int
     vector,
     pound,
 
-    // 239
+    // 241
```
thewilsonator: You only added one token, why does the count go up by two?

thewilsonator:
How does this work with varadic parameters? what about getting stringification of multiple parameters in a `string[]`?

rainers:
> what about a __traits instead

There's even `parameters` reserved as a __traits parameter, but isn't filled with functionality: https://run.dlang.io/gist/d9ab0db568456f102a1e39af574eae4c?compiler=dmd

timotheecour:
closing in favor of #7821

src/dmd/tokens.d
timotheecour: IIRC `fileFullPath` had been added without upading these numbers in the past, so i also correct that; however I'd super like to just get rid of these numbers, they serve no purpose and get out of sync easily

thewilsonator: Fair enough, just making sure you hadn't typo'd. I think they have some use when trying to look up what Tok you have in some function but I'll leave it up to other to determine what to do.

timotheecour: I see; in this case one could use instead arbitrary increasing numbers at sparse intervals, it's much more robust (and more stable across different compiler versions)
```
enum TOK : int
{
    reserved = 0,
    // we can reset numbers arbitrarily
    leftParentheses = 100,
    rightParentheses, // will be 101count)
    leftBracket = 200,
}
```

PR #7813, opened 2018-01-31
End the deprecation period of using the result of a comma expression.

wilzbach:
The deprecation period could have ended a lot earlier.
DMD deprecations should really be tracked at a common place, s.t. for every
release cycle old deprecations can be bumped accordingly.

dlang.org: https://github.com/dlang/dlang.org/pull/2151

### DIFF
--- a/changelog/comma-deprecation-error.dd
+++ b/changelog/comma-deprecation-error.dd
@@ -1,0 +1,63 @@
+The deprecation period of using the result of comma expression has ended
+
+Comma expressions have proven to be a frequent source of confusion, and bugs.
+Using their result will now trigger an error message.
+
+$(P The comma operator (`,`) allows executing multiple expressions and
+    discards the result of them except for the last which is returned.
+
+    ---
+    int a = 1;
+    int b = 2;
+    bool ret = a == 2, b == 2; // true
+    ---
+
+    It's also common to use the comma operator in for-loop increment
+    statements to allow multiple expressions.
+
+    ---
+    for (; !a.empty && !b.empty; a.popFront, b.popFront)
+    ---
+
+    Hence, using the comma operator in for-loop increment statements is still allowed.
+)
+$(H4 Corrective Action)
+$(P If possible, split the comma operator in two statements. Otherwise use
+    lambdas.
+
+    ---
+    auto result = foo(), bar();
+
+    // split off in two statements
+    foo();
+    auto result = bar();
+
+    // or use lambdas
+    auto result = {foo(); return bar();}();
+    ---
+)
+$(H4 Rationale)
+$(P The comma operator leads to unintended behavior (see below for a selection)
+    Moreover it is not commonly used and it blocks the ability to implement tuples
+    as a language features using commas.
+
+    A selection of problems through the accidental use of the comma operator:
+    ---
+    writeln( 6, mixin("7,8"), 9 ); // 6, 8, 9
+
+    template vec(T...)(T args) { ... }
+    vec v = (0, 0, 3); // 3, because vec is variadic
+
+    int a = 0;
+    int b = 2;
+    if (a == 1, b == 2) {
+        // will always be reached
+    }
+
+    void foo(int x, int y=0) {}
+    foo((a, b)); // Oops, foo is called with x=b
+
+    synchronized (lockA, lockB) {}
+    // isn't currently implemented, but still compiles due to the comma operator
+    ---
+)
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -5844,7 +5844,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
 
         e.type = e.e2.type;
         if (e.type !is Type.tvoid && !e.allowCommaExp && !e.isGenerated)
-            e.deprecation("Using the result of a comma expression is deprecated");
+            e.error("Using the result of a comma expression is not allowed");
         result = e;
     }
 
@@ -6272,7 +6272,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             /* Rewrite to get rid of the comma from rvalue
              */
             if (!(cast(CommaExp)exp.e2).isGenerated)
-                exp.deprecation("Using the result of a comma expression is deprecated");
+                exp.error("Using the result of a comma expression is not allowed");
             Expression e0;
             exp.e2 = Expression.extractLast(exp.e2, &e0);
             Expression e = Expression.combine(e0, exp);
--- a/test/fail_compilation/commaexp.d
+++ b/test/fail_compilation/commaexp.d
@@ -1,13 +1,16 @@
-/* REQUIRED_ARGS: -o- -de
+/* REQUIRED_ARGS: -o-
 TEST_OUTPUT:
 ---
-fail_compilation/commaexp.d(24): Deprecation: Using the result of a comma expression is deprecated
-fail_compilation/commaexp.d(36): Deprecation: Using the result of a comma expression is deprecated
-fail_compilation/commaexp.d(37): Deprecation: Using the result of a comma expression is deprecated
-fail_compilation/commaexp.d(38): Deprecation: Using the result of a comma expression is deprecated
-fail_compilation/commaexp.d(39): Deprecation: Using the result of a comma expression is deprecated
-fail_compilation/commaexp.d(41): Deprecation: Using the result of a comma expression is deprecated
-fail_compilation/commaexp.d(42): Deprecation: Using the result of a comma expression is deprecated
+fail_compilation/commaexp.d(27): Error: Using the result of a comma expression is not allowed
+fail_compilation/commaexp.d(39): Error: Using the result of a comma expression is not allowed
+fail_compilation/commaexp.d(40): Error: Using the result of a comma expression is not allowed
+fail_compilation/commaexp.d(41): Error: Using the result of a comma expression is not allowed
+fail_compilation/commaexp.d(42): Error: Using the result of a comma expression is not allowed
+fail_compilation/commaexp.d(44): Error: Using the result of a comma expression is not allowed
+fail_compilation/commaexp.d(45): Error: Using the result of a comma expression is not allowed
+fail_compilation/commaexp.d(56): Error: Using the result of a comma expression is not allowed
+fail_compilation/commaexp.d(69): Error: Using the result of a comma expression is not allowed
+fail_compilation/commaexp.d(81): Error: Using the result of a comma expression is not allowed
 ---
 */
 
@@ -40,4 +43,40 @@ int main () {
     assert(!ok);
     ok = true, (ok = (true, false));
     return 42, 0;
+}
+
+
+/***************************************************/
+// 16022
+
+bool test16022()
+{
+    enum Type { Colon, Comma }
+    Type type;
+    return type == Type.Colon, type == Type.Comma;
+}
+
+bool test16022_structs()
+{
+    struct A
+    {
+        int i;
+        string s;
+    }
+
+    enum Type { Colon = A(0, "zero"), Comma = A(1, "one") }
+    Type type;
+    return type == Type.Colon, type == Type.Comma;
+}
+
+/********************************************/
+
+
+void bar11(int*, int*) { }
+
+void test11()
+{
+    static int* p;
+    static int i;
+    bar11((i,p), &i);
 }
--- a/test/fail_compilation/diag10862.d
+++ b/test/fail_compilation/diag10862.d
@@ -57,7 +57,7 @@ fail_compilation/diag10862.d(77): Error: assignment cannot be used as a conditio
 fail_compilation/diag10862.d-mixin-80(80): Error: assignment cannot be used as a condition, perhaps `==` was meant?
 fail_compilation/diag10862.d-mixin-81(81): Error: assignment cannot be used as a condition, perhaps `==` was meant?
 fail_compilation/diag10862.d-mixin-82(82): Error: assignment cannot be used as a condition, perhaps `==` was meant?
-fail_compilation/diag10862.d-mixin-83(83): Deprecation: Using the result of a comma expression is deprecated
+fail_compilation/diag10862.d-mixin-83(83): Error: Using the result of a comma expression is not allowed
 fail_compilation/diag10862.d-mixin-83(83): Error: assignment cannot be used as a condition, perhaps `==` was meant?
 fail_compilation/diag10862.d-mixin-86(86): Error: `a + b` is not an lvalue
 fail_compilation/diag10862.d-mixin-87(87): Error: undefined identifier `c`
--- a/test/fail_compilation/ice10949.d
+++ b/test/fail_compilation/ice10949.d
@@ -1,7 +1,7 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/ice10949.d(12): Deprecation: Using the result of a comma expression is deprecated
+fail_compilation/ice10949.d(12): Error: Using the result of a comma expression is not allowed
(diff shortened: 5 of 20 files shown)


quickfur:
I wonder if there's a way to use `libdparse`, say, to collect all `deprecated` symbols into a list somewhere, so that we can review which symbols need to be bumped to the next stage of deprecation.

Depending on what we find in such a list, we might even be able to automatically parse the deprecation dates and output a list of symbols that need to be updated.  In the distant (or not-so-distant?) future we could even have a script/bot that automatically submits PRs to this effect.  Automation FTW!

changelog/comma-deprecation-error.dd
```diff
@@ -0,0 +1,4 @@
+The deprecation period of using the result of comma expression has ended
+
+Comma expressions have proven to be a frequent source of confusion, and bug.
```
quickfur: Nitpick: `bug` > `bugs`.

src/dmd/expressionsem.d
```diff
@@ -5844,7 +5844,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
 
         e.type = e.e2.type;
         if (e.type !is Type.tvoid && !e.allowCommaExp && !e.isGenerated)
-            e.deprecation("Using the result of a comma expression is deprecated");
+            e.deprecation("The result of a comma expression cant be used");
```
quickfur: Nitpick: `cant` > `can't`.

src/dmd/expressionsem.d
```diff
@@ -6272,7 +6272,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             /* Rewrite to get rid of the comma from rvalue
              */
             if (!(cast(CommaExp)exp.e2).isGenerated)
-                exp.deprecation("Using the result of a comma expression is deprecated");
+                exp.error("The result of a comma expression cant be used");
```
quickfur: ditto

test/fail_compilation/commaexp.d
```diff
@@ -1,13 +1,13 @@
...
-fail_compilation/commaexp.d(39): Deprecation: Using the result of a comma expression is deprecated
-fail_compilation/commaexp.d(41): Deprecation: Using the result of a comma expression is deprecated
-fail_compilation/commaexp.d(42): Deprecation: Using the result of a comma expression is deprecated
+fail_compilation/commaexp.d(24): Error: The result of a comma expression cant be used
+fail_compilation/commaexp.d(36): Error: The result of a comma expression cant be used
+fail_compilation/commaexp.d(37): Error: The result of a comma expression cant be used
+fail_compilation/commaexp.d(38): Error: The result of a comma expression cant be used
+fail_compilation/commaexp.d(39): Error: The result of a comma expression cant be used
+fail_compilation/commaexp.d(41): Error: The result of a comma expression cant be used
+fail_compilation/commaexp.d(42): Error: The result of a comma expression cant be used
```
quickfur: ditto

test/fail_compilation/diag10862.d
```diff
@@ -57,7 +57,7 @@ fail_compilation/diag10862.d(77): Error: assignment cannot be used as a conditio
 fail_compilation/diag10862.d-mixin-80(80): Error: assignment cannot be used as a condition, perhaps `==` was meant?
 fail_compilation/diag10862.d-mixin-81(81): Error: assignment cannot be used as a condition, perhaps `==` was meant?
 fail_compilation/diag10862.d-mixin-82(82): Error: assignment cannot be used as a condition, perhaps `==` was meant?
-fail_compilation/diag10862.d-mixin-83(83): Deprecation: Using the result of a comma expression is deprecated
+fail_compilation/diag10862.d-mixin-83(83): Error: The result of a comma expression cant be used
```
quickfur: ditto

test/fail_compilation/ice10949.d
```diff
@@ -1,7 +1,7 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/ice10949.d(12): Deprecation: Using the result of a comma expression is deprecated
+fail_compilation/ice10949.d(12): Error: the result of a comma expression cant be used
```
quickfur: ditto

test/runnable/ctorpowtests.d
```diff
@@ -136,7 +136,8 @@ static assert(bazra(14)==64);
 
 void moreCommaTests()
 {
-   auto k = (containsAsm(), containsAsm());
+   (containsAsm(), containsAsm());
+   auto k = containsAsm();
```
quickfur: Huh? Shouldn't this be:
````
containsAsm();
auto k = containsAsm();
````
instead?  This change doesn't make sense to me.

File: src/dmd/expressionsem.d (hunk 3 above)
Comment (wilzbach) [reply src/dmd/expressionsem.d]: Well cant is the existing DMD style:

```
> ag "cant" | wc -l
466 
> ag "can't" | wc -l
128
```

**Comment by wilzbach** [comment]
> I wonder if there's a way to use libdparse, say, to collect all deprecated symbols into a list somewhere, so that we can review which symbols need to be bumped to the next stage of deprecation.

I like the idea, but it would be a lot easier for Phobos. For dmd you are basically left with `grep` for now.
OTOH even for Phobos `grep` works fine - we even have the `@@@DEPRECATED_` pattern for Phobos, which we don't have here.
Hence, my proposal to have a simple file or wiki in which this is being kept track of for dmd - essentially it's the table at https://dlang.org/deprecate.html, but just with more love (not many deprecations actually update this file and sometimes there are compiler bugs).

https://github.com/dlang/dmd/pull/7760/files is one of the first PRs to use the deprecated pattern in dmd too.

File: src/dmd/expressionsem.d (hunk 3 above)
Comment (quickfur) [reply src/dmd/expressionsem.d]: Existing bad grammar should not be justification for adding more bad grammar!

File: test/runnable/ctorpowtests.d (hunk 7 above)
Comment (quickfur) [reply test/runnable/ctorpowtests.d]: Ah, I see why now. Nevermind, carry on. :-P

File: src/dmd/expressionsem.d (hunk 3 above)
Comment (wilzbach) [reply src/dmd/expressionsem.d]: I have seen @WalterBright repeatedly use "cant" - even in recent PRs. That was my motivation to use "cant".

Comment (wilzbach) [reply src/dmd/expressionsem.d]: Examples:
- https://github.com/dlang/dmd/pull/6092
- https://github.com/dlang/dmd/pull/7307

Comment (JinShil) [reply src/dmd/expressionsem.d]: "cant" is not a word.  I believe @WalterBright made a mistake in those PRs.

Comment (andralex) [reply src/dmd/expressionsem.d]: Yah let's fix those. Thanks!

**Review by andralex — APPROVED** [review]

File: src/dmd/expressionsem.d (hunk 3 above)
Comment (wilzbach) [reply src/dmd/expressionsem.d]: Fair enough, I changed it to can't.

**Comment by wilzbach** [comment]
So this is finally passing the testsuite :rocket: and just in case someone was wondering obviously my motivation is to make this an error with 2.079, s.t. comma expressions can be re-purposed for built-in tuples.
See also: [upcoming tuple DIP](https://forum.dlang.org/post/p3bdp1$2b4e$1@digitalmars.com)

File: test/runnable/testscope.d [hunk 8]
```diff
@@ -294,7 +294,7 @@ void test11()
     static int i;
     bar11(p, &i);
 
-    bar11((i,p), &i);  // comma expressions are deprecated, but need to test them
+    bar11(p, &i);
```
JinShil: This test was specifically testing the comma operator.  Now that it produces an error, it should probably be moved to the fail_compilation tests.  However, given that the equivalent expression already exists on line 295, it would probably be ok, to just delete it.

test/runnable/xtest46.d
```diff
@@ -7866,7 +7867,7 @@ bool test16022()
 {
```
JinShil: This [issue 16022](https://issues.dlang.org/show_bug.cgi?id=16022) was very specific to the comma operator.  Since the result of a comma expression now emits an error, this test should no longer compile, and should probably be moved to the fail_compilation test suite.

src/dmd/expressionsem.d
```diff
@@ -6272,7 +6272,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             /* Rewrite to get rid of the comma from rvalue
              */
             if (!(cast(CommaExp)exp.e2).isGenerated)
-                exp.deprecation("Using the result of a comma expression is deprecated");
+                exp.error("The result of a comma expression can't be used");
```
JinShil: I'd prefer this to say "The result of a comma expression is deprecated and can no longer be used".  But, it's ok the way it is; I'll pull it regardless.

JinShil (changes requested):

test/runnable/xtest46.d
wilzbach: Ok. I copied it to fail compilation. I don't see any reason to remove it though.

test/runnable/testscope.d
wilzbach: Yeah, I was very careful not to delete anything, but yeah in this case it does make sense. Thanks!

src/dmd/expressionsem.d
wilzbach: Hmm it's not deprecated anymore, it's a real error and should be displayed to the user as such.

changelog/comma-deprecation-error.dd
```diff
@@ -0,0 +1,4 @@
+The deprecation period of using the result of comma expression has ended
+
+Comma expressions have proven to be a frequent source of confusion, and bugs.
```
jacob-carlborg: Please include an example that now fails to compile and show how to correct it.

jacob-carlborg (changes requested):

src/dmd/expressionsem.d
jacob-carlborg: I agree with @wilzbach.

changelog/comma-deprecation-error.dd
jacob-carlborg: Perhaps include an example showing how the comma expression can still be used without using the result.

wilzbach: I was hoping I could avoid the redundancies from https://dlang.org/deprecate.html#Using%20the%20result%20of%20a%20comma%20expression, but okay copying doesn't cost us much.
I copied the text I have written for dlang.org a while ago.

changelog/comma-deprecation-error.dd
```diff
@@ -0,0 +1,59 @@
...
+    int b = 2;
+    if (a == 1, b == 2) {
+        // will always be reached
+    }
+
+    void foo(int x) {}
+    foo((++a, b));
+
+    synchronized (lockA, lockB) {}
+    // isn't currently implemented, but still compiles "thanks" to the comma operator
```
jacob-carlborg: `"thanks"` sounds a bit condescending. Perhaps replace with `due`.

test/runnable/mangle.d
```diff
@@ -324,23 +324,25 @@ void test9525()
...
-    void test2(alias a)() pure nothrow
+    bool test2(alias a)() pure nothrow
     {
         static struct S {}
         static assert(S.mangleof == result2);
         S s;
         a(&s);  // Error: Cannot convert &S to const(S*) at compile time
+        return true;
     }
-    static assert((test2!f9525(), true));
```
jacob-carlborg: Are the `static assert` not necessary anymore? Or will the `static assert` inside the function take care of that?

jacob-carlborg (approved):

test/runnable/mangle.d
wilzbach: I think the `static assert` was only there, s.t. the function got evaluated and the `static assert` inside of it could trigger, hence I added `evalTest1` and `evalTest2`.

changelog/comma-deprecation-error.dd
wilzbach: Also submitted for dlang.org: https://github.com/dlang/dlang.org/pull/2152

test/runnable/mangle.d
jacob-carlborg: Ok.

changelog/comma-deprecation-error.dd
```diff
@@ -0,0 +1,59 @@
...
+    discards the result of them except for the last which is returned.
+
+    ---
+    int a = 1;
+    int b = 2;
+    bool ret = a == 2, b == 2; // true
+    ---
+
+    It's also common to use the comma operator in for-loop increment
+    statements to allow multiple expressions.
```
CyberShadow: I think it should be clearer that this use case is not affected.

changelog/comma-deprecation-error.dd
```diff
@@ -0,0 +1,59 @@
...
+    foo();
+    auto result = bar();
+
+    // or use lambdas
+    auto result = {foo(); return bar();}();
+    ---
+)
+$(H4 Rationale)
+$(P The comma operator leads to unintended behavior (see below for a selection)
+    Moreover it is not commonly used and it blocks the ability to implement tuples.
```
CyberShadow: It blocks implementing tuples as a language feature using commas, not in general

changelog/comma-deprecation-error.dd
```diff
@@ -0,0 +1,59 @@
...
+
+    A selection of problems through the accidental use of the comma operator:
```
CyberShadow: Honestly all of these examples are entirely unconvincing because they could be caught by a "discarding result of side-effect-free subexpression" warning.

changelog/comma-deprecation-error.dd
```diff
@@ -0,0 +1,59 @@
...
+
+    template vec(T...)(T args) { ... }
+    vec v = (0, 0, 3); // 3, because vec is variadic
+
+    int b = 2;
+    if (a == 1, b == 2) {
+        // will always be reached
+    }
+
+    void foo(int x) {}
```
CyberShadow: I don't understand this example. Perhaps it would make sense for `void foo(int x, int y=0)` ?

src/dmd/expressionsem.d
CyberShadow: Would be nice to distinguish "can't be used" (due to some technical limitation) from "can't be used" (because we decided to forbid it). How about "Using the result of a comma expression is not allowed" ?

src/dmd/expressionsem.d
quickfur: @JinShil Actually, "cant" *is* a word... but it has a [different meaning](https://en.wiktionary.org/wiki/cant). :-D 

@wilzbach The next time Walter submits a PR with the wrong use of "cant", somebody should point it out so that it can be fixed. :-P

changelog/comma-deprecation-error.dd
wilzbach: It took them from here: https://dlang.org/deprecate.html#Using%20the%20result%20of%20a%20comma%20expression
https://github.com/dlang/dlang.org/pull/1308

If you have better examples, just let me know.

wilzbach:
Addressed all comments except for:

> Honestly all of these examples are entirely unconvincing because they could be caught by a "discarding result of side-effect-free subexpression" warning.

It took them from here: https://dlang.org/deprecate.html#Using%20the%20result%20of%20a%20comma%20expression
dlang/dlang.org#1308
If you have better examples, just let me know. And we can always add them to changelog later.

wilzbach:
@JinShil you still have this on "requested changes". I addressed your comments and copied the files over to `fail_compilation`, but kept the modified version in `compilation`. Are you okay with this?
I would like to move on with this, s.t. we can sip this with 2.079 and can introduce tuples into D as soon as possible ...

JinShil (approved):

quickfur:
Woohoo!!!! The comma operator is officially dead!!! *celebration*

PR #7814, opened 2018-01-31
[refactor] Replace fprintf(global.stdmsg) with message()

ibuclaw:
Although I have a funny deja vu feeling here that I've tried to do this before in a former PR.

Removes the `stdmsg` field from `Global` as where you want to send verbose messages is encapsulated in `vmessage()`.

The name of the actual functions are up for debate if you have a better name.

Alternatives could be `verbose()` or `verbatim()`.

### DIFF
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -523,16 +523,18 @@ extern (C++) final class Module : Package
             return null;
         if (global.params.verbose)
         {
-            fprintf(global.stdmsg, "import    ");
+            OutBuffer buf;
             if (packages)
             {
                 for (size_t i = 0; i < packages.dim; i++)
                 {
                     Identifier pid = (*packages)[i];
-                    fprintf(global.stdmsg, "%s.", pid.toChars());
+                    buf.writestring(pid.toChars());
+                    buf.writeByte('.');
                 }
             }
-            fprintf(global.stdmsg, "%s\t(%s)\n", ident.toChars(), m.srcfile.toChars());
+            buf.printf("%s\t(%s)", ident.toChars(), m.srcfile.toChars());
+            message("import    %s", buf.peekString());
         }
         m = m.parse();
 
@@ -553,12 +555,7 @@ extern (C++) final class Module : Package
         {
             if (!m.isRoot() && onImport)
             {
-                static __gshared bool bodge = false;
-                if(bodge)
-                    fprintf(global.stdmsg, "onImport\n");
                 auto onImportResult = onImport(m);
-                if(bodge)
-                    fprintf(global.stdmsg, "onImport %d\n", onImportResult);
                 if(onImportResult)
                 {
                     m.importedFrom = m;
--- a/src/dmd/errors.d
+++ b/src/dmd/errors.d
@@ -163,6 +163,21 @@ extern (C++) void deprecationSupplemental(const ref Loc loc, const(char)* format
 }
 
 /**
+ * Print a verbose message.
+ * Doesn't prefix or highlight messages.
+ * Params:
+ *      format = printf-style format specification
+ *      ...    = printf-style variadic arguments
+ */
+extern (C++) void message(const(char)* format, ...)
+{
+    va_list ap;
+    va_start(ap, format);
+    vmessage(format, ap);
+    va_end(ap);
+}
+
+/**
  * Just print to stderr, doesn't care about gagging.
  * (format,ap) text within backticks gets syntax highlighted.
  * Params:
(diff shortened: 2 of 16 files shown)


src/dmd/errors.d
```diff
@@ -104,6 +104,20 @@ extern (C++) void deprecationSupplemental(const ref Loc loc, const(char)* format
     va_end(ap);
 }
 
+/**
+ * Print a verbose message.
+ * Doesn't prefix or highlight messages.
```
ibuclaw: > Doesn't prefix or highlight messages.

... however there's no reason why it couldn't to make it more useful.

src/dmd/dmodule.d
```diff
@@ -553,12 +555,7 @@ extern (C++) final class Module : Package
         {
             if (!m.isRoot() && onImport)
             {
-                static __gshared bool bodge = false;
-                if(bodge)
-                    fprintf(global.stdmsg, "onImport\n");
                 auto onImportResult = onImport(m);
-                if(bodge)
-                    fprintf(global.stdmsg, "onImport %d\n", onImportResult);
```
wilzbach: :+1:

src/dmd/link.d
```diff
@@ -853,10 +855,14 @@ public int runProgram()
     //printf("runProgram()\n");
     if (global.params.verbose)
     {
-        fprintf(global.stdmsg, "%s", global.params.exefile);
+        OutBuffer buf;
+        buf.writestring(global.params.exefile);
         for (size_t i = 0; i < global.params.runargs.dim; ++i)
```
wilzbach: As I see this - FYI: there's as `asDArray` since a few weeks:

```d
foreach (param; global.params.runargs.asDArray)
    ...
```

src/dmd/link.d
```diff
@@ -670,9 +670,13 @@ public int runLINK()
         if (global.params.verbose)
         {
             // Print it
+            OutBuffer buf;
             for (size_t i = 0; i < argv.dim; i++)
```
wilzbach: As I see this - FYI: there's as `asDArray` since a few weeks:

```d
foreach (arg; argv)
    ...
```

src/dmd/globals.d
```diff
@@ -251,7 +250,6 @@ struct Global
     Param params;
     uint errors;            // number of errors reported so far
     uint warnings;          // number of warnings reported so far
-    FILE* stdmsg;           // where to send verbose messages
```
wilzbach: I always hated that - :+1:

wilzbach (approved):
Huge fan of this. I never liked the hard-coded `fprintf(global.stdmsg)`

src/dmd/dmodule.d
ibuclaw: I wasn't sure about this.  Will wait for the testsuite to give an answer...

src/dmd/globals.d
ibuclaw: GDC sends verbose messages to stderr because stdout is reserved for the use of `-pipe`, which was the rationale for having it in the first place.

If verbose messaging is encapsulated in `vmessage()` - and I implement my own versions of all these diagnostic functions anyway - then its use is no longer required.

ibuclaw:
> Huge fan of this. I never liked the hard-coded fprintf(global.stdmsg)

GDC sends verbose messages to stderr because stdout is reserved for the use of -pipe, which was the rationale for having it in the first place.

If verbose messaging is encapsulated in vmessage() - and I implement my own versions of all these diagnostic functions anyway - then its use is no longer required.

JinShil:
Restarted Jenkins

WalterBright (approved):

wilzbach:
I have set Jenkins to non required, if it's failing with spurious failures, it's still possible to merge.
The bot won't merge no longer automatically with "auto-merge", but if it's set to auto-merge it's a lot easier to find the PR and merge it manually.

If there are specific failures, of course, they can also be disabled directly at dlang/ci, but most errors I have seen so far, are related due to worker being terminated by the GCE data center etc.

wilzbach:
The CircleCi failure looks related though:

```
In file included from /usr/include/stdio.h:27:0,
                 from dmd/backend/optabgen.c:17:
/usr/include/features.h:374:25: fatal error: sys/cdefs.h: No such file or directory
 #  include <sys/cdefs.h>
                         ^
compilation terminated.
In file included from /usr/include/stdio.h:27:0,
                 from dmd/root/newdelete.c:10:
/usr/include/features.h:374:25: fatal error: sys/cdefs.h: No such file or directory
 #  include <sys/cdefs.h>
```

ibuclaw:
No, it should keep failing until I rebase it. 😉

src/dmd/link.d
WalterBright: Sure, but that's outside the scope of this PR.

PR #7815, opened 2018-01-31
Remove deprecated -dylib CLI flag

wilzbach:
They aren't documented either.

Requires:
- [ ] https://github.com/dlang/dub/pull/1361
- [x] https://github.com/dlang-community/D-YAML/pull/87

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1545,19 +1545,6 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
             }
             else if (arg == "-shared")
                 params.dll = true;
-            else if (arg == "-dylib")
-            {
-                static if (TARGET.OSX)
-                {
-                    Loc loc;
-                    deprecation(loc, "use -shared instead of -dylib");
-                    params.dll = true;
-                }
-                else
-                {
-                    goto Lerror;
-                }
-            }
             else if (arg == "-fPIC")
             {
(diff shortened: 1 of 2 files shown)


JinShil (approved):

jacob-carlborg (changes requested):
Please add a changelog entry and include an example that now fails to compile and show how to correct it.

src/dmd/mars.d
```diff
@@ -1573,12 +1560,6 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 params.multiobj = true;
             else if (arg == "-g") // https://dlang.org/dmd.html#switch-g
                 params.symdebug = 1;
-            else if (arg == "-gc")  // https://dlang.org/dmd.html#switch-gc
-            {
-                Loc loc;
-                deprecation(loc, "use -g instead of -gc");
-                params.symdebug = 2;
```
rainers: I always wondered whether this deprecation is desirable. -gc is still used by the COFF backend in case you don't want/cannot use mago with the Visual Studio debugger.

jacob-carlborg:
There's no test. Do we usually test these kind of things?

wilzbach:
> I always wondered whether this deprecation is desirable. -gc is still used by the COFF backend in case you don't want/cannot use mago with the Visual Studio debugger.

I have no idea about this. I haven't been around long enough.
-> as `-gc` seems controversial, I removed it from this PR
(if someone knows the history of `-gc`, we can easily make a new PR for it)

mathias-lang-sociomantic:
You might have forgotten to push :)

jacob-carlborg (approved):

mathias-lang-sociomantic (approved):

PR #7816, opened 2018-01-31
End deprecation phase of read-modify-write on shared variables.

wilzbach:
This has been added four years ago - https://github.com/dlang/dmd/commit/7535f3dc86e9e88c4ee609dd98b5305224c3b88a.
There really needs to be a better concept for deprecations.

(This has never been added to the deprecation page.)

Requires

- [x] https://github.com/dlang/druntime/pull/2072 
- [x] https://github.com/vibe-d/vibe.d/pull/2060
- [x] https://github.com/dlang/druntime/pull/2076

### DIFF
--- a/changelog/rwm-shared-error.dd
+++ b/changelog/rwm-shared-error.dd
@@ -1,0 +1,16 @@
+The deprecation for read-modify-write operations on `shared` variables has ended
+
+Read-modify-write operations are not allowed for `shared` variables:
+
+---
+shared int i;
+i++; // Error: read-modify-write operations are not allowed for shared variables
+---
+
+Use $(REF core, atomic, atomicOp) instead:
+
+---
+import core.atomic : atomicOp;
+shared int i;
+atomicOp!"+="(i, 1);
+---
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -2369,11 +2369,9 @@ extern (C++) abstract class Expression : RootObject
             break;
         }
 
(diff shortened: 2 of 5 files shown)


changelog/rwm-shared-error.dd
```diff
@@ -0,0 +1,3 @@
+The deprecation for read-modify-write operations on `shared` variables has ended
+
+Read-modify-write operations are not allowed for `shared` variables. Use $(REF core, atomic, atomicOp) instead.
```
jacob-carlborg: Please include an example that now fails to compile and show how to correct it.

jacob-carlborg (changes requested):

changelog/rwm-shared-error.dd
wilzbach: ok added

jacob-carlborg (approved):

wilzbach:
Argh people were depending on the following code:

```d
shared int i;
getopt(args, "i", &int);
````

https://issues.dlang.org/show_bug.cgi?id=18375

Seems like we need to fix `std.getopt` to be compatible with `shared` first :/

**Comment by wilzbach** [comment]
Looks like we can finally move ahead with this :)

**Comment by wilzbach** [comment]
The auto-tester failure on Windows 32_64 unrelated:

```
The system cannot find the path specified.
failed to execute '\"Program Files (x86)"\"Microsoft Visual Studio 10.0"\VC\bin\amd64\cl.exe /c /nologo runnable\extra-files\cpp_abi_tests.cpp /Fotest_results\runnable\cpp_abi_tests.cpp.obj'
Makefile:133: recipe for target 'test_results/runnable/cpp_abi_tests.d.out' failed
gmake[1]: *** [test_results/runnable/cpp_abi_tests.d.out] Error 1
gmake[1]: *** Waiting for unfinished jobs....
```

See also: https://github.com/braddr/d-tester/issues/71

PR #7818, opened 2018-01-31
Experiment with using only two CircleCi images

wilzbach:
Follow-up to https://github.com/dlang/dmd/pull/7420

In https://github.com/dlang/dmd/pull/7420 we added the `-fPIC` run as a new image
as CodeCov was heavily complaining, this simply tests whether this is still the case.

In other words: this PR removes the no-pie x86_64 CircleCi image as in theory x86_32 + x86_64 in pie should cover almost everything that is covered by the three of them.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ jobs:
     working_directory: ~/dmd
     docker:
       - image: circleci/node:4.8.2
-    parallelism: 2
+    parallelism: 1
     steps:
       - checkout
       - run:
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -23,10 +23,7 @@ case "${CIRCLE_STAGE}" in
         ;;
     no_pic)
         PIC=0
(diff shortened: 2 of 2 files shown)


wilzbach:
Thanks a lot for merging this @WalterBright - unfortunately I hit the bed before seeing the result here:

image

https://codecov.io/gh/dlang/dmd/compare/7fa030612d7dcb84645cc389714b0268363327eb...2a2444f2d3f2e73e7bf2ed25187566d43e2442e8

So it doesn't look good - we loose about 1.8% global coverage :/

(We recently https://github.com/dlang/dmd/pull/7711 disabled the display of the overall project coverage change as it (1) the backend has too much global state (i.e. random coverage state) too be informative and (2) it's not relevant to reviewers typically when they look at the diff. The important bit is that the diff is covered and the overall coverage changes are only in a few exceptions (like this one) interesting to look at, but then can be easily accessed by simply clicking on the CodeCov "Details" link in the CI status overview.)

PR #7820, opened 2018-01-31
recognize erroneous uses of C types

WalterBright:
Should do more of these to aid in converting C code to D.

### DIFF
--- a/src/dmd/dscope.d
+++ b/src/dmd/dscope.d
@@ -675,6 +675,8 @@ struct Scope
             tok = TOK.false_;
         else if (ident == Id.unsigned)
             tok = TOK.uns32;
+        else if (ident == Id.wchar_t)
+            tok = global.params.isWindows ? TOK.wchar_ : TOK.dchar_;
         else
             return null;
         return Token.toChars(tok);
--- a/src/dmd/lexer.d
+++ b/src/dmd/lexer.d
@@ -1020,16 +1020,25 @@ class Lexer : ErrorHandler
                     p++;
                     Token n;
                     scan(&n);
-                    if (n.value == TOK.identifier && n.ident == Id.line)
+                    if (n.value == TOK.identifier)
                     {
-                        poundLine();
-                        continue;
+                        if (n.ident == Id.line)
+                        {
+                            poundLine();
+                            continue;
+                        }
+                        else
+                        {
+                            const locx = loc();
+                            warning(locx, "C preprocessor directive `#%s` is not supported", n.ident.toChars());
+                        }
                     }
-                    else
+                    else if (n.value == TOK.if_)
                     {
-                        t.value = TOK.pound;
-                        return;
+                        error("C preprocessor directive `#if` is not supported, use `version` or `static if`");
                     }
+                    t.value = TOK.pound;
+                    return;
                 }
             default:
                 {
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -3394,7 +3394,19 @@ final class Parser(AST) : Lexer
 
         case TOK.int64:
             t = AST.Type.tint64;
-            goto LabelX;
+            nextToken();
+            if (token.value == TOK.int64)   // if `long long`
+            {
+                error("use `long` for a 64 bit integer instead of `long long`");
+                nextToken();
+            }
+            else if (token.value == TOK.float64)   // if `long double`
+            {
+                error("use `real` instead of `long double`");
+                t = AST.Type.tfloat80;
+                nextToken();
+            }
+            break;
 
         case TOK.uns64:
             t = AST.Type.tuns64;
--- a/src/dmd/id.d
+++ b/src/dmd/id.d
@@ -416,6 +416,7 @@ immutable Msgtable[] msgtable =
     { "TRUE" },
     { "FALSE" },
     { "unsigned" },
+    { "wchar_t" },
 ];
 
 
--- a/test/fail_compilation/cerrors.d
+++ b/test/fail_compilation/cerrors.d
@@ -1,0 +1,15 @@
+/* REQUIRED_ARGS: -wi
+TEST_OUTPUT:
+---
+fail_compilation/cerrors.d(11): Error: C preprocessor directive `#if` is not supported, use `version` or `static if`
+fail_compilation/cerrors.d(11): Error: declaration expected, not `#`
+fail_compilation/cerrors.d(15): Warning: C preprocessor directive `#endif` is not supported
+fail_compilation/cerrors.d(15): Error: declaration expected, not `#`
+---
+*/
+
+#if 1
+
+void test(wchar_t u);
+
+#endif
--- a/test/fail_compilation/ctypes.d
+++ b/test/fail_compilation/ctypes.d
@@ -1,0 +1,13 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/ctypes.d(11): Error: use `real` instead of `long double`
+fail_compilation/ctypes.d(12): Error: use `long` for a 64 bit integer instead of `long long`
+---
+*/
+
+void test()
+{
+    long double r;
+    long long ll;
+}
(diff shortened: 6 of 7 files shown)


src/dmd/dscope.d
```diff
@@ -673,6 +673,8 @@ struct Scope
             tok = TOK.false_;
         else if (ident == Id.unsigned)
             tok = TOK.uns32;
+        else if (ident == Id.wchar_t)
+            tok = global.params.isWindows ? TOK.wchar_ : TOK.dchar_;
```
JinShil: This doesn't appear to have any coverage.

WalterBright: That's because it'll produce different results on windows and non-windows machines.

JinShil: There are ways to limit tests in the test suite to specific platforms, so you could add a Windows test and non-Windows test to the test suite.  See the `DISABLED` qualifier e.g. https://github.com/dlang/dmd/blob/15c8da123eddff3b3113349cc968290532d9a9a3/test/runnable/test6423.d#L1

rainers: Does this make `wchar_t` part of the language? Even `size_t` is just defined in object.d. What about name mangling?

WalterBright: > Does this make wchar_t part of the language?

No. The error message happens only if it is an undefined identifier, i.e. is already an error. The idea is to produce better error messages when someone tries to compile a C file with the D compiler.

rainers: Ah, sorry, only saw the diff and failed to look at the wider context.

src/dmd/parse.d
```diff
@@ -3402,7 +3402,19 @@ final class Parser(AST) : Lexer
...
-            goto LabelX;
+            nextToken();
+            if (token.value == TOK.int64)   // if `long long`
+            {
+                error("use `long` for a 64 bit integer instead of `long long`");
+                nextToken();
+            }
+            else if (token.value == TOK.float64)   // if `long double`
+            {
+                error("use `real` for an 80 bit floating point type instead of `long double`");
```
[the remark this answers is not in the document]
klickverbot: Yes. The easiest fix is to pull the stringified type from `Target`.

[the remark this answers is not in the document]
klickverbot: (Note that this also concerns DMD/VC++ on Win64.)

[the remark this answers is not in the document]
ibuclaw: Or word it in such a way that omits the type size.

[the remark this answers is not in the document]
rainers: > (Note that this also concerns DMD/VC++ on Win64.)

DMD has 80-bit floats with VC, too, but LDC does not.

> Or word it in such a way that omits the type size.

I think this would be best, as it avoids a dependency on the backend.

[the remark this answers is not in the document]
klickverbot: > DMD has 80-bit floats with VC, too, but LDC does not.

That's precisely why this concerns DMD as well – on LDC/Win64, `long double -> real` would be just fine.

> I think this would be best, as it avoids a dependency on the backend.

Really, we should be pointing to `core.stdc.config.c_long_double`, though, as that stays portable across architectures.

WalterBright:
The semaphoreci test helpfully says "Failed" in red. That's it. No log, no clue, sigh.

wilzbach:
> The semaphoreci test helpfully says "Failed" in red. That's it. No log, no clue, sigh.

Do we have different browsers?
Here's what I see:

image

(you need to click on expand at the failing tab)


Note that this is due to the testsuite not using `-conf=`, see also https://github.com/dlang/dmd/pull/7848#issuecomment-363941920

WalterBright:
I don't see any [expand] tab. I did get the green and orange bars, but not the text underneath it. I'm using Chrome.

JinShil:
> I don't see any [expand] tab. I did get the green and orange bars, but not the text underneath it. I'm using Chrome.

Click on the pinkish bar directly under the green bar.  I made the same mistake as it's not at all apparent that the pink and green bars are clickable.

src/dmd/dscope.d
wilzbach: It's covered for me:

image


FYI: there seems to be a bug in CodeCov' browser extension that requires a refresh once you switch to the diff page to see the coverage information.
See also: https://github.com/codecov/browser-extension/issues/53

wilzbach (approved):

wilzbach:
So the Jenkins failure is a genuine one:

```
source/contentprovider.d(500,20): Error: C preprocessor directive `#Hello` is not supported
source/contentprovider.d(500,20): Error: C preprocessor directive `#First` is not supported
source/contentprovider.d(500,20): Error: C preprocessor directive `#Guten` is not supported
```

Reproduce with:

```d
string markdown = q{
#Hello World
};
```

WalterBright:
The error seems genuine, as q strings are supposed to be valid D tokens.

wilzbach:
> The error seems genuine, as q strings are supposed to be valid D tokens.

Fixed in the library and tagged as new version, s.t. Jenkins picks it up:

https://github.com/dlang-tour/core/pull/685

src/dmd/lexer.d
```diff
@@ -1064,16 +1064,25 @@ class Lexer : ErrorHandler
...
+                            continue;
+                        }
+                        else
+                        {
+                            const locx = loc();
+                            warning(locx, "C preprocessor directive `#%s` is not supported", n.ident.toChars());
+                        }
                     }
-                    else
+                    else if (n.value == TOK.if_)
```
ibuclaw: Should `#pragma` have special handling here also?

ibuclaw (approved):
Looks fine, bar handling of TOKpragma.

src/dmd/lexer.d
```diff
@@ -1064,16 +1064,25 @@ class Lexer : ErrorHandler
...
-                        continue;
+                        if (n.ident == Id.line)
+                        {
+                            poundLine();
+                            continue;
+                        }
+                        else
+                        {
+                            const locx = loc();
+                            warning(locx, "C preprocessor directive `#%s` is not supported", n.ident.toChars());
```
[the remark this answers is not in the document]
WalterBright: It's on purpose. One has a D analog, the other does not.

src/dmd/lexer.d
WalterBright: I don't see a point to it.

src/dmd/lexer.d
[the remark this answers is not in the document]
WalterBright: I don't want to turn it back into an error because it breaks existing code. A warning is fine for now.

llucenic:
I had `#x000D` as a Unicode char reference (among simliar others) in my comment block and I was not able to compile my source code since DMD 2.080.0 (I guess). Was this PR meant to parse the comments? I guess comment blocks should be left as-is. Anyway, this PR broke my code (comments in code, actually) when upgrading to the newer DMD (using dub to build my project).
Should I file a bug report or am I missing something on the PR intention here?

WalterBright:
Please file a regression bug report on https://issues.dlang.org/ so we can get it fixed!

WalterBright:
Thank you! Can you please add a chunk of code which, when compiled, illustrates the issue?

llucenic:
This is the code excerpt (comment):

	/**
	 * Grammar:
	 * 	source code characters are defined by the following grammar rules
	 * ---
	 *Character ::= LineCharacter | EndOfLineCharacter
	 *
	 *LineCharacter ::= Space
	 *		| [#x0021-#x2027] | [#x202A-#xD7FF]
	 *		| [#xE000-#xFFFD] | [#x10000-#x10FFFF]
	 *
	 *
	 *Space ::= #x0020 | #x0009 | #x000B | #x000C
	 *
	 *
	 *EndOfLineCharacter ::= #x000D | #x000A | #x2028 | #x2029
	 *
	 *
	 *EndOfLine ::= #x000D #x000A | EndOfLineCharacter | EndOfFile
	 *
	 *
	 *EndOfFile ::= #x0000 | #x001A | < physical-end-of-file >
	 * ---
	 * See_also:
	 * 	$(D isNewline()), $(D isWhitespace())
	 */

I had to remove all `#`s from in front of the x0000s in order to compile with 2.080.0.

PR #7821, opened 2018-02-01
function argument stringification: `__traits(getCallerSource, symbol)` (variadics work too)

timotheecour:
Stringify caller's function argument, analog to C's `#` stringification macro (except it's type safe).

A main use case is to allow finally writing nice logging and debug tools.

Purely library based alternatives all have drawbacks, see https://github.com/dlang/dmd/pull/7811#issuecomment-361772960 for example.

It also works with UFCS, special tokens __FILE__ etc, see examples in `test/runnable/testargnames.d`

Eg:
```d
string log(T...)(T a, string[T.length] names = __traits(getCallerSource, a)){
  import std.conv;
  return text(name, ": ", a);
}
log(1+2, "foo") // names = ["1+2", `"foo"`]
```

## EDIT
* added variadic support with clean syntax and semantics
* I've change from `__ARG_STRING__!a` to `__traits(getCallerSource, a)` after feedback from reviewers [2]

## NOTES
* this is a new implementation that avoids some of the issues in https://github.com/dlang/dmd/pull/7811 ; in this new implementation, parsed D source file contents are kept in memory until after end of semantic analysis [1]

* currently only allows binding to a function parameter (variadic or not); other cases such as template parameters could be handled in future PR without breaking backward compatibility

* i’ve dfmt-ted `test/runnable/testargnames.d` but not the rest (https://github.com/dlang-community/dfmt/issues/159 would help…) ; please ignore whitespace issues for now

[1] I've measured that impact on memory is minimal [3] Optimizations are possible if needed, if we find use cases where the memory increase is more significant. Use new hidden flag `-disposesrc` to reproduce prexisting behavior of disposing of file contents right after parsing (`getSource` will in this case resort to using `expr.toChars` which was similar to the previous implementation, cf https://github.com/dlang/dmd/pull/7811); you can measure memory with and without `-disposesrc`

[2] pros/cons of `__traits(getCallerSource, a)` vs `__ARG_STRING__`:
* no new keyword
* that seems like the 1st use of `__trait` inside a function declaration
pros/cons of `__ARG_STRING__`:
* syntax is less verbose, which matters if it's used often
* __ARG_STRING__ is natural along the line of other tokens that are used for default initialization in caller’s namespace (__FILE__ + friends)
* but these other tokens don't use template arguments, unlike `__ARG_STRING__`

NOTE: using getCallerSource instead of getSource to avoid conflating with a future potential `getSource`

[3] detailed reproducible memory tests

```
brew install gnu-time

gtime -v generated/osx/release/64/dmd -unittest -version=StdUnittest -version=import_std -main -o- -i=std test/runnable/imports/import_world.d
#+ same with added `-disposesrc`
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:52.96
Maximum resident set size (kbytes): 7711924
```
time is always very similar (with +/- 1 second)
Maximum resident set size varies quite a bit between runs (high variance) but the extra memory involved (at most 10M from `du -sh phobos/std`) is just `10M/7711924k = 1/771` of the total memory usage, so negligible.

On an extreme, unrealistic example that imports all of phobos and does 0 work (ie no unittests or main code), the total memory usage will be less so the fraction from keeping file contents in memory  relative to total memory consumption will be higher, but still acceptable:
```
gtime -v generated/osx/release/64/dmd -version=import_std -main -o- -i=std  test/runnable/imports/import_world.d
#+ same with -disposesrc
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:02.75 # same regardless
Maximum resident set size (kbytes): 663340 # without -disposesrc
Maximum resident set size (kbytes): 655432 # with -disposesrc
```
this 2nd testing scenario has very low variance, and we get:
`(663340-655432)/655432.=1% `
which is acceptable (and again, this is the unfavorable case as explained)


src/dmd/dmodule.d
```diff
@@ -883,10 +883,15 @@ extern (C++) final class Module : Package
             if (p.errors)
                 ++global.errors;
         }
+        // PRTEMP
```
timotheecour: this part is temporary

test/runnable/testargnames.d
```diff
@@ -0,0 +1,94 @@
...
+
+  enum t = 44;
+  fun1(t + t, `t + t`);
+
+  // parenthesis are removed:
+  fun1((t + t), `t + t`);
+
+  // NOTE: __ARG_STRING__ will pretty-print the expression so passing `t+t`
+  // or `t + t` won't make a difference:
+  // TODO: how to do formatting ignoring with dfmt?
```
jacob-carlborg: See the documentation: https://github.com/dlang-community/dfmt#disabling-formatting

timotheecour: thanks, done!

wilzbach:
> I'm ok with doing that change if there's consensus [2]

I think we can get there quickly:

- Vote here with "UP" if you like `__traits`
- Vote with "DOWN" if you like `__ARG_STRING__`
- Vote with "Confused" if you prefer a third, yet to be named option (or don't like this in general)
- Vote with "Heart" if you don't care about `__traits` vs. `__ARG_STRING__`, but just want to finally use this in your code

marler8997:
Note that keeping the files in memory may open up other optimizations and features that weren't available before.  This allows strings to be used directly from source.  This enables better error messages in the semantic phase since the original source is available to pull context from.  The current design which frees source files before semantic analysis may have been holding back superior designs for various solutions.  I pose this question to the more seasoned dmd developers who may know of different ways this change could be taken advantage of.

wilzbach:
One thing that comes to my mind is `__traits(documentation)` - https://github.com/dlang/dmd/pull/6872
It was closed because of the slight increase in memory consumption :/

timotheecour:
> One thing that comes to my mind is __traits(documentation) - #6872
It was closed because of the slight increase in memory consumption :/

added evaluation (+ how to reproduce) of effect on max memory usage + time, see updated PR description. Seems like 0 effect on time and negligible effect on memory.

ibuclaw:
> This enables better error messages in the semantic phase since the original source is available to pull context from.

You have file, line, and column information, you can pull this out of the file at any time during the error diagnostic and read the slice it points to.

ibuclaw:
I think a traits would be better. `__ARG_STRING__` doesn't really look natural alongside other tokens.

You seem to have an unrelated change included here also (`-vbytes`), but I'll give the benefit of the doubt that this is still wip.

d-random-contributor:
`__traits(source)` - it can be subsequently extended to get source of other things like lambda passed as template argument, IIRC somebody asked for such feature.

d-random-contributor:
@ZombineDev was it you?

timotheecour:
> You have file, line, and column information, you can pull this out of the file at any time during the error diagnostic and read the slice it points to.

* the file contents (before this PR) is cleared from memory right after parsing, and before semantic analysis, so it was preventing to do that. With this PR, we have access to it; 

* file,line,col (even with sources still available) doesn't allow O(1) access in a slice; with this PR, we have O(1) access via `bytes` offset

timotheecour:
> You seem to have an unrelated change included here also (-vbytes), but I'll give the benefit of the doubt that this is still wip.

the change is related to this PR: this PR introduces `bytes` offset which was needed for O(1) access to source content at a given `Loc` (again, line+col wasn't enough for O(1) access). `-vbytes` just exposes this `bytes` offset in compiler messages and is the analog to `-vcolumns`. Right now it's set as undocumented flag but I can make it a regular documented flag.

ntrel:
Just wanted to mention that a possible enhancement would be to support multiple arguments using `,` as a separator e.g.:
```d
void f(int a, int b, string args = __traits(getArgument, a, b)) {writeln("Arguments: ", args);}
f(1, n % 3);
```
`args` would be `1, n % 3`. In fact, using `__traits(getArgument)` without passing any symbols could be equivalent to passing the corresponding parameter names for each argument provided at the call site - default arguments would not be included. That would be a useful shortcut for logging.

timotheecour:
> I think a traits would be better

 done, changed to `__traits(getSource, symbol)`

* added variadic support (this was tricky to do but it works)

PTAL (minus whitespace issues)

jacob-carlborg:
* I would like the test cases are clearly separated an a comment is added for each case explaining exactly what's tested

* There are no failing tests, i.e. testing things that shouldn't compile

jacob-carlborg (changes requested):
* I would like the test cases are clearly separated an a comment is added for each case explaining exactly what's tested

* There are no failing tests, i.e. testing things that shouldn't compile

* There should be a PR for updating the spec

* The internal names don't match anymore, i.e. `ArgStringInitExp`

timotheecour:
> Just wanted to mention that a possible enhancement would be to support multiple arguments using , as a separator e.g.:

separators are not good because they can lead to ambiguity, and moves on the burden on user to reparse into separate arguments; now that variadics work (and produce a static array), it'd not be hard to allow multiple fields. this syntax could work:
`__traits(getCallerSource, [symbol1, symbol2])` to leave space for other arguments eg:
`__traits(getCallerSource, [symbol1, symbol2], AST)`
`__traits(getCallerSource, [symbol1, symbol2], json)`

timotheecour:
I would like the test cases are clearly separated an a comment is added for each case explaining exactly what's tested

There are no failing tests, i.e. testing things that shouldn't compile

There should be a PR for updating the spec

The internal names don't match anymore, i.e. ArgStringInitExp

changelog/argname.dd
```diff
@@ -0,0 +1,16 @@
+Stringify caller's function argument, analog to C's `#` stringification macro (except it's type safe).
```
WalterBright: What exactly are you adding? `__traits(getCallerSource, a)` ? Or is it more than that?

src/dmd/dmodule.d
```diff
@@ -309,6 +309,15 @@ extern (C++) final class Module : Package
         modules = new DsymbolTable();
     }
 
+    extern (C++) void releaseResources()
```
WalterBright: Global comment: Add Ddoc comment for all new functions.

src/dmd/expression.d
```diff
@@ -1614,6 +1616,26 @@ extern (C++) abstract class Expression : RootObject
             va_end(ap);
         }
     }
+    else
+    {
+        pragma(inline, true)
```
WalterBright: why force inline this?

src/dmd/expression.d
```diff
@@ -7096,6 +7124,57 @@ extern (C++) final class FileInitExp : DefaultInitExp
     }
 }
 
+/***********************************************************
+ */
+extern (C++) final class ArgStringInitExp : DefaultInitExp
```
WalterBright: What is an ArgStringInitExp? There's no documentation here.

src/dmd/expression.d
```diff
@@ -1614,6 +1616,26 @@ extern (C++) abstract class Expression : RootObject
...
     }
+    else
+    {
+        pragma(inline, true)
+        extern(D)
+        final void error(string file=__FILE__, int line=__LINE__)(const(char)* format, ...) const
+        {
+            version(with_debug){
+                import std.conv:text;
+                writelnL(file, ":", line);
```
WalterBright: where/what is writelnL()?

src/dmd/expression.d
```diff
@@ -1614,6 +1616,26 @@ extern (C++) abstract class Expression : RootObject
             va_end(ap);
         }
     }
+    else
+    {
+        pragma(inline, true)
+        extern(D)
+        final void error(string file=__FILE__, int line=__LINE__)(const(char)* format, ...) const
+        {
+            version(with_debug){
```
WalterBright: where/what is with_debug?

src/dmd/expressionsem.d
```diff
@@ -598,8 +598,96 @@ private bool functionParameters(Loc loc, Scope* sc, TypeFunction tf, Type tthis,
                 }
                 arg = p.defaultArg;
                 arg = inlineCopy(arg, sc);
-                // __FILE__, __LINE__, __MODULE__, __FUNCTION__, and __PRETTY_FUNCTION__
-                arg = arg.resolveLoc(loc, sc);
+
+                if(auto argexp = arg.isArgStringInitExp())
+                {
+                  if(argexp.ident is null)
```
WalterBright: Please don't invent a different indentation style. It's 4 columns.

src/dmd/expressionsem.d
```diff
@@ -598,8 +598,96 @@ private bool functionParameters(Loc loc, Scope* sc, TypeFunction tf, Type tthis,
...
+                            error(argexp.loc, "cannot refer to default arg  `%s` not explicitly set by caller", argexp.ident.toChars);
+                            return true;
+                        }
+
+                        auto argname = getArgSring(argu, sc).idup;
+                        values~=argname;
+                    }
+                    if(is_variadic_param){
+                        arg=argexp.resolveArgStrings(loc, sc, values);
+                    } else{
```
WalterBright: please don't invent new formatting styles.

src/dmd/expressionsem.d
```diff
@@ -598,8 +598,96 @@ private bool functionParameters(Loc loc, Scope* sc, TypeFunction tf, Type tthis,
                 }
                 arg = p.defaultArg;
                 arg = inlineCopy(arg, sc);
-                // __FILE__, __LINE__, __MODULE__, __FUNCTION__, and __PRETTY_FUNCTION__
-                arg = arg.resolveLoc(loc, sc);
+
```
WalterBright: What does the following big block of code do? I don't have a clue, so I have nothing to compare the code to. Also, it's so large consider breaking it out into a separate function.

src/dmd/expressionsem.d
```diff
@@ -598,8 +598,96 @@ private bool functionParameters(Loc loc, Scope* sc, TypeFunction tf, Type tthis,
...
+                if(auto argexp = arg.isArgStringInitExp())
+                {
+                  if(argexp.ident is null)
+                  {
+                    error(argexp.loc, "argexp.ident null");
```
WalterBright: This error message refers to compiler internals and will mean nothing to the user.

src/dmd/expressionsem.d
```diff
@@ -598,8 +598,96 @@ private bool functionParameters(Loc loc, Scope* sc, TypeFunction tf, Type tthis,
...
+                            is_variadic_param=false;
+                            break;
+                        }
+                        else
+                        {
+                            // TODO: factor with where this magic string is defined
+                            enum vararg_name = "_param_";
+                            // TODO: define/use startsWith
+                            auto ident2b=ident2.toString;
+                            if(ident2b.length>vararg_name.length && ident2b[0..vararg_name.length]==vararg_name)
```
WalterBright: Global comment: Please use a space before and after operators. Please format code like in the rest of the front end.

src/dmd/expressionsem.d
```diff
@@ -10062,3 +10193,38 @@ private bool checkFunctionAttributes(Expression exp, Scope* sc, FuncDeclaration
         return error;
     }
 }
+
+// Get source code for `expr`
```
WalterBright: Why is this different from the code in hdrgen.d which turns ASTs back into strings?

src/dmd/globals.d
```diff
@@ -378,15 +380,18 @@ struct Loc
     const(char)* filename;
     uint linnum;
     uint charnum;
+    // 0-based offset from origin in bytes (allows O(1) access)
+    uint bytes;
```
WalterBright: Adding this field, since Loc is attached to every AST node, will result in a very large increase in memory consumption.

src/dmd/mars.d
```diff
@@ -1648,6 +1656,10 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 params.vtls = true;
             else if (arg == "-vcolumns") // https://dlang.org/dmd.html#switch-vcolumns
                 params.showColumns = true;
+            else if (arg == "-vbytes") // https://dlang.org/dmd.html#switch-bytes
+                params.showBytes = true;
+            else if (arg == "-disposesrc")
+                params.disposeSrcContent = true;
```
WalterBright: why is this optional under user control?

src/dmd/mtype.d
```diff
@@ -5494,6 +5494,14 @@ extern (C++) final class TypeFunction : TypeNext
             this.trust = TRUST.trusted;
     }
 
+    extern(D) Parameter searchByName(const(char)[] name){
```
WalterBright: Global comment: Need Ddoc comments for all new functions.

src/dmd/tokens.d
```diff
@@ -81,7 +81,7 @@ enum TOK : int
     delegatePointer,
     delegateFunctionPointer,
 
-    // 54
+    // 54 // TODO: instead of these non robust numbers, use `cast(TOK) number`
```
WalterBright: The numbers are there because when I print out a TOK value, in order to figure out which one it is, I have to count the tokens.

src/dmd/traits.d
```diff
@@ -412,6 +413,31 @@ extern (C++) Expression pointerBitmap(TraitsExp e)
...
+* __traits(getCallerSource, symbol) => get symbol source code (string[T.length] or string)
+*/
+extern (C++) Expression traitsGetSource(TraitsExp e)
+{
+    if (!e.args || e.args.dim < 1|| e.args.dim > 2)
+    {
+        error(e.loc, "TODO");
+        return new ErrorExp();
+    }
+    auto ret = new ArgStringInitExp(e.loc);
```
WalterBright: This is allocated here, but is not initialized or used until the end of the function. It should be allocated there. Also, convention is to use `result` as the return value, not `ret`.

src/dmd/traits.d
```diff
@@ -412,6 +413,31 @@ extern (C++) Expression pointerBitmap(TraitsExp e)
...
+    auto ident_expr=(*e.args)[0];
```
WalterBright: calling it `arg0` would make more sense, especially since the next line verifies that it is a type, not an identifier.

src/dmd/traits.d
```diff
@@ -412,6 +413,31 @@ extern (C++) Expression pointerBitmap(TraitsExp e)
...
+    auto t=isType(ident_expr);
+    if(!t){
+        error(e.loc, "TODO.2");
```
WalterBright: Internal errors should use `assert`.

src/dmd/traits.d
```diff
@@ -412,6 +413,31 @@ extern (C++) Expression pointerBitmap(TraitsExp e)
...
+        return new ErrorExp();
+    }
+    auto name=ident_expr.toChars();
+    // TODO: lookup?
+    auto id = Identifier.idPool(name, name.strlen);
+    ret.setIdent(id);
+    ret.exp=e;
```
WalterBright: Perhaps ArgStringInitExp should have a constructor that takes these two expressions as arguments.

src/dmd/traits.d
```diff
@@ -1599,6 +1626,12 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         return pointerBitmap(e);
     }
 
+    if (e.ident == Id.getCallerSource)
+    {
+        e.error("`%s` unexpected in this context", Id.getCallerSource);
```
WalterBright: Should this be an `assert` instead?

changelog/argname.dd
```diff
@@ -0,0 +1,16 @@
...
+
+NOTE: it also works with UFCS, with tuples (string static array) and non-tuples (string)
+
+Eg:
+---
+string logSimple(T...)(T a, string[T.length] names = __traits(getCallerSource, a))
+{
+  import std.conv;
+  return text(names, ": ", a);  // see testargnames for a better log function
```
wilzbach: Use four spaces for indentation + don't reference to DMD's testsuite in the changelog.

wilzbach:
ping @timotheecour

timotheecour:
I don't have time to work on this at the moment but anyone is welcome to reopen a PR based on this;
a few notes:
* a prerequisite for this PR is to add `bytes` field, if that's a problem it'll block this PR.
* if adding that field is unacceptable, an alternative is to store byte offset of each line start by adding an array bytes_of_line[] in class `File` or `Module` to allow fast access.

PR #7822, opened 2018-02-01
Fix Issue 18348 - Color highlighting of backticks doesn't work

wilzbach:
tl;dr: backticks aren't highlighted.

Reproduce:

```d
echo "aa" | dmddev -c -
```

Or: https://run.dlang.io/is/ioU894

I would also prefer when I manually set `-color=on` to true, DMD would accept that.
It requires ugly workarounds otherwise.

### DIFF
--- a/test/compilable/test18348.sh
+++ b/test/compilable/test18348.sh
@@ -1,0 +1,44 @@
+#!/usr/bin/env bash
+
+set -eu pipefail
+
+dir=${RESULTS_DIR}${SEP}compilable
+output_file=${dir}/test18348.sh.out
+
+faketty () { script -qfc "$(printf "%q " "$@")" /dev/null ; }
+export TERM="faketerm"
+
+check()
+{
+    # the first sed removes remove carriage returns
+    out=$(echo "$1" |
+        faketty "$DMD" -c -o- -color=on - | tail -n+2 | \
+        sed 's/\x0d//g' | \
+        sed 's/\x1b\[m/|RESET|/g' |
+        sed 's/\x1b\[1m/|BOLD|/g' | \
+        sed 's/\x1b\[0;\([0-9]*\)m/|+++\1|/g' | \
+        sed 's/\x1b\[1;\([0-9]*\)m/|BOLD++++\1|/g' | \
+        sed 's/+++30|/BLACK|/g' | \
+        sed 's/+++31|/RED|/g' | \
+        sed 's/+++32|/GREEN|/g' | \
+        sed 's/+++33|/YELLOW|/g' | \
+        sed 's/+++34|/BLUE|/g' | \
+        sed 's/+++35|/PURPLE|/g' | \
+        sed 's/+++36|/CYAN|/g' | \
+        sed 's/+++37|/WHITE|/g')
+    expected="$2"
+    if [ "$out" != "$expected" ] ; then
+        echo "Expected: $expected"
+        echo "Actual:   $out"
+        exit 1
+    fi
+}
+
+if ! [ $OS == "win32" -o  $OS == "win64" ]; then
+    check "test" \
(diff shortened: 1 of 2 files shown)


test/compilable/test18348.sh
```diff
@@ -0,0 +1,34 @@
...
+
+check()
+{
+    # the first sed removes remove carriage returns
+    out=$(echo "$1" |
+        faketty "$DMD" -c -o- -color=on - | tail -n+2 | \
+        sed 's/\x0d//g' | \
+        sed 's/\x1b\[m/.empty./g' |
+        sed 's/\x1b\[1m/.reset./g' | \
+        sed 's/\x1b\[\([0-9;]*\)m/_set:\1_/g')
```
wilzbach: Sanitizes the output, s.t. it's easy to look at without bumping your head against weird escape codes.

test/compilable/test18348.sh
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+set -eu pipefail
+
+dir=${RESULTS_DIR}${SEP}runnable
+output_file=${dir}/test13774.sh.out
+
+faketty () { script -qfc "$(printf "%q " "$@")" /dev/null ; }
```
wilzbach: Required because `-color=on` because DMD requires a TERM to exist for colorized output:

https://github.com/dlang/dmd/blob/419f13596a8fd44fd5a5301cb8db454d61be2b29/src/dmd/console.d#L172-L176

timotheecour:
could we just do the way many tools do? -color=[on|off|auto]
auto would switch on/off depending on `isatty`

eg: git help diff
```
--color[=<when>]
           Show colored diff.  --color (i.e. without =<when>) is the same as --color=always.  <when> can be one of always, never, or auto. It can be changed by the color.ui and
           color.diff configuration settings.
```

wilzbach:
> could we just do the way many tools do? -color=[on|off|auto]
> auto would switch on/off depending on isatty

Yes -> https://github.com/dlang/dmd/pull/7823

test/compilable/test18348.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env bash
+
+set -eu pipefail
+
+dir=${RESULTS_DIR}${SEP}compilable
+output_file=${dir}/test18348.sh.out
+
+faketty () { script -qfc "$(printf "%q " "$@")" /dev/null ; }
+export TERM="faketerm"
```
wilzbach: Apparently this trick doesn't on some CI. Setting this to blocked as https://github.com/dlang/dmd/pull/7823 is a lot nicer.

wilzbach:
and just for reference - here's how things look for me before this change:

image

And after:

image

The test case makes it clearer what happens:

```
Before: |BOLD|__stdin.d(2): |BOLD+RED|Error: |RESET|no identifier for declarator |CYAN||RESET||BOLD|test|CYAN||RESET|
After: |BOLD|__stdin.d(2): |BOLD+RED|Error: |RESET|no identifier for declarator |CYAN||BOLD+WHITE|test|CYAN||RESET|
```

Note that there weren't any tests for this.

aG0aep6G:
@wilzbach: So the escape sequences are emitted as intended, but your terminal doesn't show bold font. Note that "__stdin.d(2):" is also supposed to be bold.

If this is a common restriction among terminals, then color is in order, I guess. But you should specify a background color as well. White text on default background is unreadable in a terminal with a light color scheme.

wilzbach:
urxvt (my terminal)

image

Terminator:

image

Yeah boldness is disabled in my terminal, but I would still be a lot better to be able to use white as a color.
If we can't do that (or use a different color), maybe it's a good idea to allow user-overwrites of the DMD color codes?
I opened an issue for this: https://issues.dlang.org/show_bug.cgi?id=18391

Geod24:
Just tested this with the latest DMD and it works.

PR #7823, opened 2018-02-01
-color=on should overwrite the default detection

wilzbach:
Imho if I as a user tell DMD that it should do colorized output, I expect it to obey my command.
This would avoid the faketty hack in https://github.com/dlang/dmd/pull/7822 and run.dlang.io (https://github.com/dlang-tour/core-exec/blob/master/entrypoint.sh)

### DIFF
--- a/src/dmd/console.d
+++ b/src/dmd/console.d
@@ -57,6 +57,35 @@ struct Console
 
         @property FILE* fp() { return _fp; }
 
+        /**
+        Tries to detect whether DMD has been invoked from a terminal.
+        Returns: `true` if a terminal has been detected, `false` otherwise
+        */
+        static bool detectTerminal()
+        {
+            auto h = GetStdHandle(STD_OUTPUT_HANDLE);
+            CONSOLE_SCREEN_BUFFER_INFO sbi;
+            if (GetConsoleScreenBufferInfo(h, &sbi) == 0) // get initial state of console
+                return false; // no terminal detected
+
+            version (CRuntime_DigitalMars)
+            {
+                if (!isatty(stdout._file))
+                    return false; // no terminal detected
+            }
+            else version (CRuntime_Microsoft)
+            {
+                if (!isatty(fileno(stdout)))
+                    return false; // no terminal detected
+            }
+            else
+            {
+                static assert(0, "Unsupported Windows runtime.");
+            }
+
+            return true; // terminal detected
+        }
+
         /*********************************
          * Create an instance of Console connected to stream fp.
          * Params:
@@ -67,23 +96,6 @@ struct Console
          */
         static Console* create(FILE* fp)
         {
-            /* Determine if stream fp is a console
-             */
-            version (CRuntime_DigitalMars)
-            {
-                if (!isatty(fp._file))
-                    return null;
-            }
-            else version (CRuntime_Microsoft)
-            {
-                if (!isatty(fileno(fp)))
-                    return null;
-            }
-            else
-            {
-                return null;
-            }
-
             DWORD nStdHandle;
             if (fp == stdout)
                 nStdHandle = STD_OUTPUT_HANDLE;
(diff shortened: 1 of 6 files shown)


timotheecour:
moved from https://github.com/dlang/dmd/pull/7822#issuecomment-362234898

could we just do the way many tools do? -color=[on|off|auto]
auto would switch on/off depending on isatty
```
eg: git help diff

--color[=<when>]
           Show colored diff.  --color (i.e. without =<when>) is the same as --color=always.  <when> can be one of always, never, or auto. It can be changed by the color.ui and
           color.diff configuration settings.
```
ie, could we make `bool color` be a 3 valued enun insetead

test/compilable/testcolor.sh
```diff
@@ -0,0 +1,23 @@
+#!/usr/bin/env bash
+
+set -eu pipefail
```
marler8997: `set -euo pipefail` ?

src/dmd/console.d
```diff
@@ -57,6 +57,21 @@ struct Console
...
         @property FILE* fp() { return _fp; }
 
+        /**
+        Tries to detect whether DMD has been invoked from a terminal.
+        Returns: `true` if a terminal has been detect, `false` otherwise
+        */
+        static bool detectedTerminal()
+        {
+            auto nStdHandle = STD_OUTPUT_HANDLE;
+            auto h = GetStdHandle(nStdHandle);
```
marler8997:
```D
auto h = GetStdHandle(STD_OUTPUT_HANDLE);
```

src/dmd/console.d
```diff
@@ -57,6 +57,21 @@ struct Console
...
+            CONSOLE_SCREEN_BUFFER_INFO sbi;
+            if (GetConsoleScreenBufferInfo(h, &sbi) == 0) // get initial state of console
+                return false;
```
marler8997: This is probably just my personal style, but I like to include a comment for raw boolean return statements,
```D
    return false; // no terminal detected

return true; // terminal detected
```

src/dmd/console.d
marler8997: Or even
```D
if (GetConsoleScreenBufferInfo(GetStdHandle(STD_OUTPUT_HANDLE), &sbi) == 0)
```

marler8997 (approved):

wilzbach:
BTW I still have the Windows failure to deal with:

```
Expected: [1m__stdin.d(2): [1;31mError: [mno identifier for declarator [0;36m[1mtest[0;36m
Actual:   __stdin.d(2): Error: no identifier for declarator `test`
Makefile:223: recipe for target 'test_results/compilable/testcolor.sh.out' failed
gmake[1]: *** [test_results/compilable/testcolor.sh.out] Error 1
Makefile:258: recipe for target 'start_compilable_tests' failed
gmake: *** [start_compilable_tests] Error 2
gmake: *** Waiting for unfinished jobs....
```

I assume this is due to the Windows `console.d` logic being different, but my Windows Foo is almost non-existent. I added `set +x` :/

wilzbach:
Finally got around rebasing/reworking this. Let's hope that Windows finally passes this time (I replaced the `strings` command with a crude `sed` call.

src/dmd/console.d
```diff
@@ -57,33 +57,45 @@ struct Console
...
-         * Params:
-         *      fp = io stream
-         * Returns:
-         *      pointer to created Console
-         *      null if failed
-         */
-        static Console* create(FILE* fp)
+        /**
+        Tries to detect whether DMD has been invoked from a terminal.
+        Returns: `true` if a terminal has been detect, `false` otherwise
```
rainers: s/detect/detected/

src/dmd/console.d
```diff
@@ -57,33 +57,45 @@ struct Console
...
+        */
+        static bool detectedTerminal()
```
rainers: `detectTerminal` seems to fit better, as there is no cached result involved.

src/dmd/globals.d
```diff
@@ -360,6 +360,11 @@ struct Global
         }
         _version = (import("VERSION") ~ '\0').ptr;
         compiler.vendor = "Digital Mars D";
+
+        // -color=auto is the default value
+        import dmd.console : Console;
+        if (Console.detectedTerminal)
```
rainers: dmd code does not omit `()` in unction calls AFAICT.

src/dmd/mars.d
```diff
@@ -1579,7 +1579,9 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 //      -color=on|off
                 if (p[6] == '=')
                 {
-                    if (strcmp(p + 7, "off") == 0)
+                    if (strcmp(p + 7, "auto") == 0)
+                        params.color = false;
```
rainers: This assumes the default to be `true`. You should check `on` here instead and error out if not `auto` below.

rainers (changes requested):
Nits + logic in cli parsing slightly off.

GoaLitiuM:
Any updates on this? Before I saw this PR, I wanted to do a similar change for `-color` to add a new option 'always' which would work like the option 'on' in this PR, but I think this change makes more sense.

ZombineDev:
Yes, this PR is definitely a step in the right direction. I think @wilzbach said that he's going to be busy this week, so hopefully next week he will be able to rebase his PR, so we can go ahead with merging it.

PR #7825, opened 2018-02-01
Don't require object.d for -lib if code does not utilize runtime features

JinShil:
This is a followup to the following PRs

 * #7786 for `Throwable`
 * #7799 for `TypeInfo`

Currently, if one wants to create a library in D, that doesn't utilize any runtime features, and use said library from another language, they will need to create an empty `object.d` file to get a build.  

This PR removes that limitation, so users will only need the .d source files relevant to their library to get a build.

### DIFF
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -634,9 +634,11 @@ extern (C++) final class Module : Package
         {
             if (!strcmp(srcfile.toChars(), "object.d"))
             {
-                .error(loc, "cannot find source code for runtime library file 'object.d'");
-                errorSupplemental(loc, "dmd might not be correctly installed. Run 'dmd -man' for installation instructions.");
-                errorSupplemental(loc, "config file: %s", FileName.canonicalName(global.inifilename));
+                // object.d is automatically imported.  If object.d doesn't exist, it's not necessarily an error.
+                // For example, the user could be creating a library in D that is intened to be called from
+                // another language and, therefore, does not wish to use features of the D runtime.  We
+                // return `false` here and rely on future semantic processing to catch any errors.
+                return false;
             }
             else
             {
--- a/test/compilable/minimal_lib.d
+++ b/test/compilable/minimal_lib.d
@@ -1,0 +1,11 @@
+// PERMUTE_ARGS:
+// REQUIRED_ARGS: -conf= -lib
+
+// This test ensures that a library created in D that does not rely on any
+// runtime features can be compiled without object.d.  The `-conf=` compiler
+// flag ensures the default object.d is not in the import path.
+
+extern(C) int add(int a, int b)
+{
+    return a + b;
+}


test/compilable/minimal_lib.d
```diff
@@ -0,0 +1,9 @@
+// REQUIRED_ARGS: -lib
+
+// This test ensures that a library created in D that does not rely on any
+// runtime features can be compiled without object.d
```
jacob-carlborg: How does this test ensures what's described here?

JinShil: Actually, I had an error in the test:  `-conf=` must be added to avoid importing object.d from the default runtime.  Therefore, there is no object.d in the import path.  The fact that this test compiles with the `-conf=` and `-lib` flags is enough to ensure it the test passes.  Go ahead and try it for yourself locally without this change, and you will find it doesn't compile.  However, after apply this PR, it will.

jacob-carlborg: The compiler will not try to find a config globally somewhere in `/etc` with the `-conf=` flag?

jacob-carlborg:
You could add a test that accesses something in `object.d` and makes sure it does not compile. It would be a bit more clear, at least to me.

test/compilable/minimal_lib.d
JinShil: That's correct.  `-conf=` basically means "don't use any .conf file".  Because there is no .conf file being used the are no import directories like `-I/usr/include`.  Therefore the `object.d` file that typically ships with DMD is not imported.

JinShil:
> You could add a test that accesses something in object.d and makes sure it does not compile. It would be a bit more clear, at least to me.

That would not remove the necessity of the current test.  

Currently if `object.d` cannot be found, dmd emits an error saying...

```
Error:  cannot find source code for runtime library file 'object.d' 
          dmd might not be correctly installed. Run 'dmd -man' for installation instructions.
```
... even if the code being compiled uses nothing from `object.d` or anything else in the runtime.  This PR removes that limitation so users don't have to add a phony `object.d` to get a build.

The test in this PR proves that the compiler can still build the library even in the absence of `object.d`, which isn't currently possible.

timotheecour:
is that related in any way to https://issues.dlang.org/show_bug.cgi?id=18301 : 
"Issue 18301 - `Error: no object files to link` when passing a ".a" instead of a ".o" => should be supported"
?

JinShil:
> is that related in any way to https://issues.dlang.org/show_bug.cgi?id=18301 ?

No, this is totally unrelated.

src/dmd/dmodule.d
```diff
@@ -634,9 +634,11 @@ extern (C++) final class Module : Package
...
             if (!strcmp(srcfile.toChars(), "object.d"))
             {
-                .error(loc, "cannot find source code for runtime library file 'object.d'");
-                errorSupplemental(loc, "dmd might not be correctly installed. Run 'dmd -man' for installation instructions.");
-                errorSupplemental(loc, "config file: %s", FileName.canonicalName(global.inifilename));
+                // object.d is automatically imported.  If object.d doesn't exist, it's not necessarily an error.
+                // For example, the user could be creating a library in D that is intened to be called from
+                // another language and, therefore, does not wish to use features of the D runtime.  We
+                // return `false` here and rely on future semantic processing to catch any errors.
+                return false;
```
rainers: I suspect this will make bad installations less obvious, especially to newcomers, that don't know anything about what's to be expected in object.d.

Also, in combination with optional TypeInfo, ModuleInfo and Throwable, this might silently and advertently build unexpected code due to missing import paths.

If someone is actually starting the adventure of porting to another platform, adding a switch or even creating an empty object.d should not be that much of an issue.

jacob-carlborg:
> That is already thoroughly tested elsewhere, when symbols are not found in imported files; object.d is no different

@JinShil What I'm saying is that, as far as I can see, if `object.d` is imported in your test, it will still pass. Is there a test that makes sure that `object.d` is not available when `-conf=` is used?

src/dmd/dmodule.d
JinShil: > I suspect this will make bad installations less obvious, especially to newcomers, that don't know anything about what's to be expected in object.d.

I doubt it.  With basic things like `Object`, `size_t`, and `string` defined in `object.d`, they'll know right away that something's not right.  The existing error message is quite useless to a newcomer anyway. Because `object.d` is imported behind the scenes, they won't even know what the compiler's talking about, let alone what to do about it.  This PR is no worse than the status quo in the case of a bad installation, IMO.

> Also, in combination with optional TypeInfo, ModuleInfo and Throwable, this might silently and advertently build unexpected code due to missing import paths.

I'm not sure what you mean here.  The compiler will fail if it can't find needed symbols, and the linker will fail if it can't find needed references.  The compiler should succeed if it found everything it needed.  The likelihood of an build that shouldn't have succeeded as a result of this PR seems extremely rare to me.

> If someone is actually starting the adventure of porting to another platform, adding a switch or even creating an empty object.d should not be that much of an issue.

Agreed, but this PR is primarily for those who are writing libraries in D to be consumed by other languages.  There is no benefit that I can see to forcing them to add an empty `object.d` to their source code.  It's just unnecessary friction and cruft.

rainers: > This PR is no worse than the status quo in the case of a bad installation, IMO.

I disagree: without the current error, you are likely to get undefined identifiers/unknown properties for symbols mostly considered part of the language and no hint what actually is wrong.



> The compiler will fail if it can't find needed symbols,

For example, this silently produces different results when compiled separately with or without object.d:
```
module noobj;
extern(C) int printf(const char*, ...);
extern(C) void goo();

void foo()
{
	scope(exit) printf("exit\n");
	printf("foo\n");
	goo();
}
```
when linked afterwards with this compiled with object.d:
```
import noobj;

extern(C) void goo()
{
	throw new Exception("ho");
}

void main()
{
	foo();
}
```

> There is no benefit that I can see to forcing them to add an empty object.d to their source code.

Agreed, but a command line switch might do the trick here. Maybe even couple with -betterC?

rainers: > Because object.d is imported behind the scenes, they won't even know what the compiler's talking about

A slightly modified error message including "implicitly imported" might help in this case.

BTW: object.d can also be imported explicitly, and failing to find it should still raise an error.

PR #7827, opened 2018-02-01
fix issue 18352 - [REG 2.078] dmd can't generate 64-bit binaries on Windows

rainers:
… 10 with VS 2015

checking the KitsRoot10 lib folder is not good enough to detect the SDK, it might just contain the UCRT. Mimick vcvarsall.bat and verify that windows.h exists aswell.

### DIFF
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -1046,7 +1046,7 @@ version (Windows)
             if (WindowsSdkDir is null)
             {
                 WindowsSdkDir = GetRegistryString(r"Microsoft\Windows Kits\Installed Roots", "KitsRoot10");
-                if (WindowsSdkDir && !FileName.exists(FileName.combine(WindowsSdkDir, "Lib")))
+                if (WindowsSdkDir && !findLatestSDKDir(FileName.combine(WindowsSdkDir, "Include"), r"um\windows.h"))
                     WindowsSdkDir = null;
             }
             if (WindowsSdkDir is null)


JinShil (approved):

src/dmd/link.d
```diff
@@ -1046,7 +1046,7 @@ version (Windows)
             if (WindowsSdkDir is null)
             {
                 WindowsSdkDir = GetRegistryString(r"Microsoft\Windows Kits\Installed Roots", "KitsRoot10");
-                if (WindowsSdkDir && !FileName.exists(FileName.combine(WindowsSdkDir, "Lib")))
+                if (WindowsSdkDir && !findLatestSDKDir(FileName.combine(WindowsSdkDir, "Include"), r"um\windows.h"))
```
wilzbach: All the hipsters use backticks nowadays ;-)

(FWIW the two alternative ways of doing WYSIWYG strings seem to be yet another historical baggage of D)

wilzbach: Also the same logic is used here too:

https://github.com/dlang/dmd/blob/38355439498d9e7606bb590e92d3bf4bec372b25/src/dmd/link.d#L1080-L1084

Might make sense to avoid duplications as this code seems to be highly susceptible to failures.

wilzbach:
Any way we can ensure on AppVeyor on auto-tester that this doesn't regress in the future?

src/dmd/link.d
rainers: WindowsSdkVersion is no longer used after https://github.com/dlang/dmd/pull/7828, so this copy can be removed later.

rainers:
> Any way we can ensure on AppVeyor on auto-tester that this doesn't regress in the future?

Only by uninstalling all Windows 10 SDKs from the Appveyor image, so not really realistic.

PR #7828, opened 2018-02-01
fix issue 18354 - [Reg 2.078] Building fails with VC 2015 Build Tools

rainers:
replace VS version check with check for existence of legacy_stdio_definitions.lib

cherry-picked from https://github.com/dlang/dmd/pull/7500

### DIFF
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -960,21 +960,23 @@ version (Windows)
         const(char)* linkOptions(bool x64)
         {
             OutBuffer cmdbuf;
-            if (auto p = getVCDir(VCDir.Lib, x64))
+            if (auto vclibdir = getVCDir(VCDir.Lib, x64))
             {
                 cmdbuf.writestring(" /LIBPATH:\"");
-                cmdbuf.writestring(p);
+                cmdbuf.writestring(vclibdir);
                 cmdbuf.writeByte('\"');
-            }
-            if (VisualStudioVersion && strcmp(VisualStudioVersion, "14") >= 0)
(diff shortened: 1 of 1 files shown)


JinShil (approved):

src/dmd/link.d
```diff
@@ -960,21 +960,24 @@ version (Windows)
         const(char)* linkOptions(bool x64)
         {
             OutBuffer cmdbuf;
-            if (auto p = getVCDir(VCDir.Lib, x64))
+            if (auto vclibdir = getVCDir(VCDir.Lib, x64))
+            if (vclibdir)
```
wilzbach: Isn't this `if` unnecessary?

wilzbach:
Shouldn't this have been caught by AppVeyor? Any way we can ensure that it doesn't regress in the future?

JinShil:
> Shouldn't this have been caught by AppVeyor?

AppVeyor doesn't appear to be running on stable.

wilzbach:
> AppVeyor doesn't appear to be running on stable.

Ah that's probably due to 

image

src/dmd/link.d
rainers: Ooops, merge error. Removed and amended.

rainers:
> Shouldn't this have been caught by AppVeyor? Any way we can ensure that it doesn't regress in the future?

No, this only happens if no Visual Studio is installed, but the VC++ Build Tools. Last time I tried, the latter couldn't even be installed if there is already a VS installation on the system.

PR #7830, opened 2018-02-02
[WIP] Investigate GraphQL for JSON introspection/filtering

marler8997:
DISCLAIMER:  Note that supporting the full GraphQL spec would surely be overkill for this application.  This PR is just exploring whether we could leverage a subset of GraphQL to take advantage of the solutions it provides for what we need.  Also note that I'm not convinced myself whether this is a good fit for dmd, but I thought seeing if we could leverage the latest technology was worth looking into.

Seeing the new `-probe` flag being proposed (https://github.com/dlang/dmd/pull/7521) and the new data being added to the JSON file (https://github.com/dlang/dmd/pull/7757#pullrequestreview-93510556), I noticed that these use cases and more were just a subset of what you get from using graphql (http://graphql.org/).  I decided to look into using a "subset" of graphql for both "filtering json output" and also supporting "introspection" on the json format.  For those familair with graphql, I only plan on supporting SelectionSets with Fields (no fragments) (see http://facebook.github.io/graphql/October2016/#sec-Language.Operations).  The advantage of graphql is that it will keep the JSON format specification in sync with the JSON that is generated, this being a good application of the DRY principle (see http://graphql.org/learn/schema and http://graphql.org/learn/execution for more info).  Here are some examples of it's use:
```
dmd -Xg=compilerInfo
```
> Note: Assume `-Xg=<graphql>`
This would output:
```
{
  "data" : {
    "compilerInfo" : {
      "binary" : "dmd",
      "version" : "2.068.1"
      "config": null,
      "predefinedVersions": [
        "DigitalMars",
        "Posix",
        ..
      ]
    }
  }
}
```
Here's an example of deeper filtering:
```
dmd -Xg=compilerInfo{version}
```
```
{
  "data" : {
    "compilerInfo" : {
      "version" : "2.068.1"
    }
  }
}
```

Note that for backwards compatibility, if no graphql query is given, it will default to the original format , namely, an array of objects representing the module files.  However, if you want to select that data in graphql, you could do:
```
dmd -Xg=modules
```
and get
```
{
  "data" : {
    "modules" : [
       ...
    ]
  }
}
```

The filtering is nice, but the main benefit of using graphql will be the introspection.  You will be able to use it to query the JSON type system, to know what fields are available.  I still need to learn a bit more about GraphQL to provide a good example, so I'll try to update this as I learn a bit more.

### DIFF
--- a/src/dmd/graphql/core.d
+++ b/src/dmd/graphql/core.d
@@ -1,0 +1,340 @@
+module dmd.graphql.core;
+
+enum TypeKind
+{
+    boolean,
+    string_,
+    object_,
+}
+
+struct QueryDataObject
+{
+    struct Field
+    {
+        string name;
+        QueryData value;
+    }
+    Field[] fields;
+}
+struct QueryData
+{
+    union
+    {
+        QueryDataObject object_;
+        string str;
+        bool bool_;
+    }
+    TypeKind kind;
+    this(QueryDataObject object_)
+    {
+        this.object_ = object_;
+        this.kind = TypeKind.object_;
+    }
+    this(string str) { this.str = str; this.kind = TypeKind.string_; }
+    this(bool bool_) { this.bool_ = bool_; this.kind = TypeKind.boolean; }
+    QueryDataObject asObject()
+    {
+        assert(kind == TypeKind.object_);
+        return object_;
+    }
+}
+
+
+interface IQueryDataHandler
+{
+    void errorSelectOnValue(const(QuerySelectionSet), TypeKind);
+
+    void objectStart();
+    void objectEnd();
+    void string_(string str);
+    void boolean(bool b);
+}
+
+
+//
+// NOTE: the following data structures are generated from
+//       the grammar specification of a graphql query
+struct QueryValue
+{
+}
+struct QueryArgument
+{
+    string name;
+    QueryValue value;
+}
+struct QueryDirective
+{
+    string name;
+    QueryArgument[] arguments;
+}
+struct QueryField
+{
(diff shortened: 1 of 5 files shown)


wilzbach:
That looks like over-engineering and re-inventing the wheel. I would generally recommend to ask for feedback on the NG or Slack before you start such big projects like this one.

Except for the `-probe` use case (where low overhead is wanted), what's wrong with:

```sh
dmd -Xf- | jq .modules
```

assuming we teach dmd to output the JSON on stdout if requested.

OTOH I have to admit that I like `dmd -Xg=compilerInfo` ;-)

marler8997:
> That looks like over-engineering and re-inventing the wheel. 

Well, I was kinda seeing if we could prevent "re-inventing" the wheel by leveraging the solutions that other engineers have already found to these problems.  GraphQL gives complete freedom to any application to not only query the exact data it needs, but also to introspect on the data.  With a flag like `-probe`, the responsibility is on the compiler to provide all the information, but to also find a balance between to little or too much.  And then invent interfaces to allow applications to enable/disable subsets of that data.  With GraphQL, the interface is defined from the get-go, it never changes, and it supports everything. It might be overkill, but I'm not sure.

timotheecour:
maybe add an example with field1 and field2?

timotheecour:
looks like superseeded by https://github.com/dlang/dmd/pull/7838 ?

marler8997:
This is an optional enhancement of #7838, though, I'm leaning on the side of this being "more trouble than it's worth".

PR #7831, opened 2018-02-03
allow default valued parameters after template variadics

timotheecour:
Function parameters with default values are now allowed after variadic template parameters, and always take their default values. This allows using special tokens (eg __FILE__) after variadic parameters, which was previously impossible.

Eg:
```d
string log(T...)(T a, string file = __FILE__, int line = __LINE__)
{
  return text(file, ":", line, " ", a);
}

assert(log(10, "abc") == text(__FILE__, ":", __LINE__, " 10abc"));
```

This should be preferred to the previous workaround, which caused template bloat:
```d
string log(string file = __FILE__, int line = __LINE__, T...)(T a);
```

### DIFF
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -1349,7 +1349,7 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
                         for (size_t j = parami + 1; j < nfparams; j++)
                         {
                             Parameter p = Parameter.getNth(fparameters, j);
-                            if(p.defaultArg)
+                            if (p.defaultArg)
                             {
                                break;
                             }
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -1342,13 +1342,17 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
                         declaredTuple = new Tuple();
                         (*dedargs)[parameters.dim - 1] = declaredTuple;
 
-                        /* Count function parameters following a tuple parameter.
-                         * void foo(U, T...)(int y, T, U, int) {}  // rem == 2 (U, int)
+                        /* Count function parameters with no defaults following a tuple parameter.
+                         * void foo(U, T...)(int y, T, U, double, int bar = 0) {}  // rem == 2 (U, double)
                          */
                         size_t rem = 0;
                         for (size_t j = parami + 1; j < nfparams; j++)
                         {
                             Parameter p = Parameter.getNth(fparameters, j);
+                            if(p.defaultArg)
+                            {
+                               break;
+                            }
                             if (!reliesOnTident(p.type, parameters, inferStart))
                             {
                                 Type pt = p.type.syntaxCopy().typeSemantic(fd.loc, paramscope);
--- a/changelog/default_after_variadic.dd
+++ b/changelog/default_after_variadic.dd
@@ -1,4 +1,4 @@
-Function parameters with default values are now allowed after variadic template parameters, and always take their default values. This allows using special tokens (eg __FILE__) after variadic parameters, which was previously impossible
+Function parameters with default values are now allowed after variadic template parameters, and when IFTI is used, always take their default values. This allows using special tokens (eg __FILE__) after variadic parameters, which was previously impossible. See also test/runnable/testdefault_after_variadic.d
 
 Eg:
 ---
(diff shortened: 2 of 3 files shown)


wilzbach (commented):
Huge fan of this and I'm impressed that's a two line change.
People have been complaining about this on the NG forever and it was cited as one of the reason why logging in D "sucks".
Also
1) You can add at least https://issues.dlang.org/show_bug.cgi?id=8687 to the fixed bugs.
2) This needs a PR to dlang.org to update the spec

CC @MartinNowak as he mentioned this on Bugzilla:

> While working on limited lifetimes (@safe RC/Uniq) I ran into another use-case where a default argument is needed.

src/dmd/dtemplate.d
```diff
@@ -1342,13 +1342,17 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
...
-                        /* Count function parameters following a tuple parameter.
-                         * void foo(U, T...)(int y, T, U, int) {}  // rem == 2 (U, int)
+                        /* Count function parameters with no defaults following a tuple parameter.
+                         * void foo(U, T...)(int y, T, U, double, int bar = 0) {}  // rem == 2 (U, double)
                          */
                         size_t rem = 0;
                         for (size_t j = parami + 1; j < nfparams; j++)
                         {
                             Parameter p = Parameter.getNth(fparameters, j);
+                            if(p.defaultArg)
```
ibuclaw: Style

jacob-carlborg:
What happens with code like this, will it compile?

```d
string foo(T...)(T a, string file = __FILE__) if (T.length == 1)
{
    return null;
}
foo(1, "bar");
```

timotheecour:
@jacob-carlborg added a test. it does not compile as expected (and doesn't get confused by the fact "bar" is of same type as "file").

src/dmd/dtemplate.d
timotheecour: done

timotheecour:
PTAL, all comments addressed.

> You can add at least https://issues.dlang.org/show_bug.cgi?id=8687 to the fixed bugs.

done

> This needs a PR to dlang.org to update the spec

will do after this is merged (or do you recommend another way?)

WalterBright (approved):

jmdavis:
Yay! Thank you, @timotheecour for doing this.

MartinNowak:
> While working on limited lifetimes (@safe RC/Uniq) I ran into another use-case where a default argument is needed.

Could be helpful would have been the better wording, mainly to pass a default allocator as last argument and keep tracking it's scopeness.

schveiguy:
This PR changed the behavior of valid code. Even though default parameters didn't make sense (because you had to specify them), it didn't stop actual code from using them. See [issue 19057](https://issues.dlang.org/show_bug.cgi?id=19057).

I think the better fix would be to make default parameters actually work (and match the parameters if the trailing arguments have the same type).

In other words:
```D
void foo(A...)(A a, int x = 1, int y = 2)
{
    writeln(a, x, y);
}

void main()
{
   foo("hello"); // 1
   foo("hello", 5); // 2
   foo("hello", 5, 6); // 3
}
```

Prior to this PR, 1 and 2 were errors, but 3 worked and printed "hello56"
After this PR, 1, 2, and 3 work, but print "hello12", "hello512", "hello5612" which is a change in behavior for at least one previously working call.

My expectation is that it should print "hello12", "hello52", "hello56". I don't get why parameters with default values should never match the actual parameters.

jmdavis:
> I think the better fix would be to make default parameters actually work (and match the parameters if the trailing arguments have the same type).

That would be a disaster in the making for __FILE__ and __LINE__, because it would then be trivial to accidentally match them and thus give them the wrong values as well as then be missing arguments that were then intended to be part of the variadic arguments. And I really don't see much point in even having non-default parameters after variadic parameters. You can just make them part of the variadic parameters. So, I seriously question that even supporting having non-default parameters after variadic parameters is particularly worth it, and what you're suggesting would make having parameters with default arguments after variadic parameters highly risky (not just for __FILE__ and __LINE__ but for pretty much any default arguments), whereas these changes actually made them work.

schveiguy:
> That would be a disaster in the making for `__FILE__` and `__LINE__`

No, it wouldn't. You just wouldn't do that, until we can figure out a better way (see [issue 18919](https://issues.dlang.org/show_bug.cgi?id=18919)). Just go back to the templated versions for now.

> And I really don't see much point in even having non-default parameters after variadic parameters. You can just make them part of the variadic parameters.

It's pretty obvious to me, to cut down on complexity and template bloat. What's easier/more understandable?

```D
void foo(A...)(A args, int timeout)
// or
void foo(A...)(A args) if (A.length > 0 && is(A[$-1] == int))
{
    int timeout = a[$-1];
}
```

It's not for us to question how people have written their code. `__FILE__` and `__LINE__` are not the only use cases for default parameters, and those use cases don't always match the pattern that `__FILE__` and `__LINE__` do.

PR #7835, opened 2018-02-04
Fix Issue 18364 - header file generation doesn't print the package name in package(XXX)

wilzbach:
Blocking https://github.com/dlang/phobos/pull/6123


wilzbach:
This is a pretty straight-forward bug fix that has been sitting in the queue for a while. Anything blocking it?
It passes on all CIs and comes with a test ...

JinShil (approved):

jacob-carlborg (approved):

PR #7836, opened 2018-02-04
Fix Issue 18365 - header file generation doesn't include the return attribute

wilzbach:
This is a first approach to the bug that's blocking https://github.com/dlang/phobos/pull/6123.
I'm not a huge fan of working around against the existing code and I will look into accepting `return` as postfix argument too, but maybe someone has a better idea?

### DIFF
--- a/src/dmd/hdrgen.d
+++ b/src/dmd/hdrgen.d
@@ -921,6 +921,12 @@ public:
         }
         t.inuse++;
         PrePostAppendStrings pas;
+        extern(C++) int ignoreReturn(void* param, const(char)* name)
+        {
+            if (strcmp(name, "return") != 0)
+                PrePostAppendStrings.fp(param, name);
+            return 0;
+        }
         pas.buf = buf;
         pas.isCtor = (ident == Id.ctor);
         pas.isPostfixStyle = false;
@@ -931,7 +937,7 @@ public:
             MODtoBuffer(buf, t.mod);
             buf.writeByte(' ');
         }
-        t.attributesApply(&pas, &PrePostAppendStrings.fp);
+        t.attributesApply(&pas, &ignoreReturn);
         if (t.linkage > LINK.d && hgs.ddoc != 1 && !hgs.hdrgen)
         {
             linkageToBuffer(buf, t.linkage);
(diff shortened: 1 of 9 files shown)


src/dmd/hdrgen.d
```diff
@@ -963,6 +969,9 @@ public:
             buf.writeByte(')');
         }
         parametersToBuffer(t.parameters, t.varargs);
+        buf.writeByte(' ');
```
RazvanN7: This should be inside the if statement, otherwise you will have a space even if you don't have a `return` attribut. That's what makes the tests to fail.

test/compilable/extra-files/header18365.d
```diff
@@ -0,0 +1,6 @@
+module foo.bar.ba;
+nothrow pure @nogc @safe package(foo) 
+{
+	void foo();
+	nothrow pure @nogc @safe package(foo.bar) void foo2();
+}
```
RazvanN7: why is this necessary? You can add your test cases in the compilable/extra-files/header1.d or header2.d or header3.d.
There is no need to add an extra file.

RazvanN7 (dismissed):
Looks good, except for the space error.

test/compilable/extra-files/header18365.d
wilzbach: Because of `module foo.bar.ba;`

ZombineDev:
@wilzbach you need to fix the tests

RazvanN7:
ping @wilzbach . What's the status on this?

wilzbach:
Ping @RazvanN7 @JinShil @ZombineDev  - this is read now and finally passing all CIs.

test/runnable/testscope2.d
```diff
@@ -42,7 +42,7 @@ void test3()
         assert(SS.foo4.mangleof == "_D10testscope22SS4foo4MFNcNkKNgPiZi");
 
         // Test scope pretty-printing
-        assert(typeof(SS.foo1).stringof == "ref return ulong(return ref int* delegate() return p)");
+        assert(typeof(SS.foo1).stringof == "ref ulong(return ref int* delegate() return p) return");
```
jacob-carlborg: I hope this will not break downstream projects.

JinShil (approved):

PR #7837, opened 2018-02-04
Fix Issue 18365 - allow return storage class to be set in prefix style

wilzbach:
As an alternative to https://github.com/dlang/dmd/pull/7836.

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -744,6 +744,10 @@ final class Parser(AST) : Lexer
                 stc = AST.STC.gshared;
                 goto Lstc;
 
+            case TOK.return_:
+                stc = AST.STC.return_;
+                goto Lstc;
+
             //case TOK.manifest:   stc = STC.manifest;     goto Lstc;
 
             case TOK.at:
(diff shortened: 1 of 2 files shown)


wilzbach:
Note that the D spec is out of date anyhow:

https://dlang.org/spec/grammar.html#declarations

`return` is not a storage class

https://dlang.org/spec/grammar.html#attributes

but a valid attribute.

rainers:
Doesn't this produce ambiguities for nested functions or doesn't allow it there?

wilzbach:
I'm not sure - imho `return` is highly under-documented.
I assume it will just ignore the storage class or error:

```
struct Foo
{
    char function(char a) return value2()
    {
    	static char foo(char a)
    	{
            return a;
    	}
    	return &foo;
    }
}
```

will error with:

```
compilable/test18365.d(8): Error: const/immutable/shared/inout/return attributes are only valid for non-static member functions
```

BTW I'm not saying this is the best way to move forward, but when trying to generate header files for Phobos I hit the problem that `return` is the only attribute that is only allowed postfix, but emitted in prefix by the header generation.

RazvanN7:
Since https://github.com/dlang/dmd/pull/7836 was merged. Can you close this one please?

PR #7838, opened 2018-02-04
Added -Xi to include new fields in JSON output

marler8997:
Adds a new option (hidden for now) that allows an application to select what it wants in the JSON output, i.e.
```
dmd -Xi=compilerInfo -Xi=semantics
```

This would produce a JSON file looking like this:
```
{
    "compilerInfo" : {
        ...
    },
    "semantics": {
        ...
    }
}
```

The option is hidden because the only users of the option (currently known) will be `rdmd` and `dub`.

NOTE: the `-probe` equivalent would be:
```
dmd -Xf=- -Xi=compilerInfo
```
** json output now works even with no input files


marler8997:
NOTE: I also plan on adding the extra fields that were added in (https://github.com/dlang/dmd/pull/7521), however, I thought that should be it's own set of changes.

Hopefully we can get this one merged quickly so I can add the new fields from `-probe` with a good amount of time before the next release.

src/dmd/json.d
```diff
@@ -857,17 +858,165 @@ public:
...
+        for (auto start = next;;)
+        {
+            next++;
+            if (!isNameChar(*next))
+                return start[0 .. next - start];
+        }
+    }
+
+    /**
+    Returns true if `c` is a valid character for the start of a name..
```
wilzbach: double `.`

wilzbach (approved):
Nice. That looks _a lot_ better than your previous approach. 
I'm fully in favor of this PR and hope we can get it as soon as possible ;-)

src/dmd/json.d
```diff
@@ -857,17 +858,165 @@ public:
...
-        Module m = (*modules)[i];
-        if (global.params.verbose)
-            message("json gen %s", m.toChars());
-        m.accept(json);
-    }
-    json.generateSemantics();
-    json.arrayEnd();
-    json.removeComma();
+
+    if (global.params.jsonQuery is null)
```
timotheecour:
```
if global.params.jsonQuery is null
  global.params.jsonQuery = "modules"; // or whatever we want default query to be; maybe put all the queries here

# then the rest can assume we always have `jsonQuery`
```

src/dmd/json.d
```diff
@@ -857,17 +858,165 @@ public:
...
+        json.removeComma();
+    }
+    else
+    {
+        // Generate the "query format" which is an object where each
+        // output option is it's own field.
+
+        json.objectStart();
+        auto parser = QueryParser(global.params.jsonQuery);
+        for (;;)
```
timotheecour: `while(true)` seems more idiomatic?

src/dmd/json.d
```diff
@@ -857,17 +858,165 @@ public:
...
+        {
+            parser.skipTrivial();
+            if (*parser.next == '\0')
+                break;
+            auto fieldName = parser.tryParseName();
+            if (fieldName is null)
+            {
+                //json.generateError(format("invalid query: expected name but got '%s' (0x%02x)",
+                //    *parser.next, cast(ubyte)*parser.next));
+                json.generateError("invalid query: expected name or EOF but got something else".ptr);
```
timotheecour: most useful diagnostic would be a to show byte offset from start of string where parser got confused

src/dmd/json.d
wilzbach: FYI: `for (;;)` is the "idiomatic" way of Phobos. We should probably put such trivias in the style guide, s.t. they don't constantly pop up...

src/dmd/json.d
wilzbach: (also this is uncovered)

src/dmd/mars.d
```diff
@@ -360,6 +360,11 @@ private int tryMain(size_t argc, const(char)** argv)
     }
     if (files.dim == 0)
     {
+        if (global.params.jsonQuery)
+        {
+            generateJson(null);
+            return EXIT_SUCCESS;
```
wilzbach: Ah nice. I didn't see this initially. 

This is more or less what I am trying to achieve in https://github.com/dlang/dmd/pull/7840

src/dmd/mars.d
```diff
@@ -998,6 +970,45 @@ private int tryMain(size_t argc, const(char)** argv)
...
+    OutBuffer buf;
+    json_generate(&buf, modules);
+
+    // Write buf to file
+    const(char)* name = global.params.jsonfilename;
+    if (name && name[0] == '-' && name[1] == 0)
+    {
+        // Write to stdout; assume it succeeds
+        size_t n = fwrite(buf.data, 1, buf.offset, stdout);
+        assert(n == buf.offset); // keep gcc happy about return values
```
wilzbach: Needs a test.
Example:

```sh
#!/usr/bin/env bash

set -ueo pipefail

name="$(basename "$0" .sh)"
dir="${RESULTS_DIR}/compilable/"
out="$dir/${name}.json.out"

"$DMD" -X | ${RESULTS_DIR}/sanitize_json > "$out"
diff "$out" compilable/extra-files/$name.json
rm "$out"

"$DMD" -Xf=- | ${RESULTS_DIR}/sanitize_json > "$out"
diff "$out" compilable/extra-files/$name.json
rm "$out"
```

(`sanitize_json` currently doesn't accept stdin)

src/dmd/mars.d
```diff
@@ -1869,6 +1880,11 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                         goto Lnoarg;
                     params.jsonfilename = p + 3 + (p[3] == '=');
                     break;
+                case 'q':
+                    if (!p[3])
+                        goto Lnoarg;
```
wilzbach: test

src/dmd/mars.d
```diff
@@ -998,6 +970,45 @@ private int tryMain(size_t argc, const(char)** argv)
...
 }
 
+private void generateJson(Modules* modules)
+{
+    OutBuffer buf;
+    json_generate(&buf, modules);
+
+    // Write buf to file
+    const(char)* name = global.params.jsonfilename;
+    if (name && name[0] == '-' && name[1] == 0)
```
timotheecour: `strcmp`

src/dmd/json.d
```diff
@@ -857,17 +858,165 @@ public:
...
-    json.arrayEnd();
-    json.removeComma();
+
+    if (global.params.jsonQuery is null)
+    {
+        // Generate the original "non-query format, which is just an array
+        // of modules representing their syntax.
+
+        json.arrayStart();
+        for (size_t i = 0; i < modules.dim; i++)
```
timotheecour: foreach

src/dmd/json.d
```diff
@@ -857,17 +858,165 @@ public:
...
+            {
+                //json.generateError(format("invalid query: expected name but got '%s' (0x%02x)",
+                //    *parser.next, cast(ubyte)*parser.next));
+                json.generateError("invalid query: expected name or EOF but got something else".ptr);
+                break;
+            }
+            auto fieldID = tryParseJsonField(fieldName);
+            if (!fieldID.hasValue)
+            {
+                json.generateError(("invalid query: unknown field name '" ~ fieldName ~ "'\0").ptr);
```
timotheecour: aren't backticks "`" used instead of "'"  in messages ?

timotheecour: `.ptr` is not needed; D string litterals are null terminated and can decay to `const(char)*`
+everywhere relevant

src/dmd/json.d
```diff
@@ -857,17 +858,165 @@ public:
...
+    modules,
+    semantics,
+}
+struct OptionalJsonField
+{
+    bool hasValue;
+    JsonField value;
+}
+OptionalJsonField tryParseJsonField(const(char)[] fieldName)
+{
```
timotheecour: DRY:
```
foreach(a; __traits(allMembers, JsonField))
  if(fieldName == a) return OptionalJsonField(true, mixin(`JsonField.`~a));
```

src/dmd/mars.d
timotheecour: diff can take `-`: https://stackoverflow.com/a/9847527/1426932

```
"$DMD" -X | ${RESULTS_DIR}/sanitize_json  | diff - compilable/extra-files/$name.json
``

File: src/dmd/mars.d [hunk 12]
```diff
@@ -998,6 +970,45 @@ private int tryMain(size_t argc, const(char)** argv)
...
+         */
+        const(char)* jsonfilename;
+        if (name && *name)
+        {
+            jsonfilename = FileName.defaultExt(name, global.json_ext);
+        }
+        else
+        {
+            // Generate json file name from first obj name
+            const(char)* n = global.params.objfiles[0];
```
timotheecour: `assert(global.params.objfiles.length > 0);` ?

src/dmd/mars.d
```diff
@@ -998,6 +970,45 @@ private int tryMain(size_t argc, const(char)** argv)
...
+    const(char)* name = global.params.jsonfilename;
+    if (name && name[0] == '-' && name[1] == 0)
+    {
+        // Write to stdout; assume it succeeds
+        size_t n = fwrite(buf.data, 1, buf.offset, stdout);
+        assert(n == buf.offset); // keep gcc happy about return values
+    }
+    else
+    {
+        /* The filename generation code here should be harmonized with Module::setOutfile()
```
timotheecour: yup!

test/compilable/extra-files/json2.out
```diff
@@ -0,0 +1,62 @@
+{
+    "buildInfo": {
+        "config": "VALUE_REMOVED_FOR_TEST",
+        "cwd": "VALUE_REMOVED_FOR_TEST",
+        "importPaths": [
+            "compilable",
+            "..\/..\/druntime\/import",
```
timotheecour: how about making sanitize replace `\/` with `/` ?

test/sanitize_json.d
```diff
@@ -39,25 +39,61 @@ int main(string[] args)
...
-        else if (kind == "buildInfo")
-            sanitizeBuildInfo(obj.object);
-        else if(kind == "module")
-            sanitizeSyntaxNode(obj);
-        else if(kind == "semantics")
-            sanitizeSemantics(obj.object);
+        foreach (ref obj; root.array)
+        {
+            auto kind = obj.object["kind"].str;
+            if (kind == "compilerInfo")
```
timotheecour: DRY + type safety:
s/"compilerInfo"/JsonField.compilerInfo.stringof/
+ everywhere relevant

src/dmd/mars.d
```diff
@@ -1869,6 +1880,11 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                         goto Lnoarg;
                     params.jsonfilename = p + 3 + (p[3] == '=');
                     break;
+                case 'q':
+                    if (!p[3])
+                        goto Lnoarg;
+                    params.jsonQuery = p + 3 + (p[3] == '=');
```
timotheecour: check p[3] == '=' and `params.jsonQuery = p + 4`
We really should not allow things like `-Xqfoo` in new flags; this is old baggage from dmd cmd line (eg `-offile` instead of (`-of=file`), the downsides have these have been discussed: it's not standard, hard to read visually, and disallows future new flags (eg `-Xquery=baz`)

timotheecour:
* I suggest making flag `-Xf-` documented instead of undocumented since it's used externally (rdmd and especially dub are external to dmd)
* toplevel commit msg should mention new behaviors:
** `-Xf=-` means stdout
** json output now works even with no input files

src/dmd/json.d
marler8997: Its definitely better to not break backwards compatibility.  If at some point in the future someone determines it's OK in this case, then it's simple to remove the special case.

wilzbach:
> toplevel commit msg should mention new behaviors:

What do you mean with "top-level"  - wouldn't be the merge commit ?
A (short) changelog entry is much more valuable to the user.

src/dmd/json.d
marler8997: Doesn't appear to work in this case

src/dmd/mars.d
marler8997: This is existing code, I think any changes to it would be a seperate PR.

src/dmd/mars.d
marler8997: > This is existing code, I think any changes to it would be a seperate PR.
ditto

src/dmd/mars.d
marler8997: > This is existing code, I think any changes to it would be a seperate PR.

ditto

src/dmd/mars.d
marler8997: Yes I agree with this, I'll remove support for omitting `=`.

test/compilable/extra-files/json2.out
marler8997: This is the JSON generated by `std.json` (not dmd).  Dmd doesn't generate these weird escaped slashes.  To fix this we would need to post-process the JSON after it is sanitized, basically another JSON parser that sanitizes after `std.json`...lol :)

test/sanitize_json.d
marler8997: This is the test suite, not dmd.  It doesn't have access to the `JsonField` type defined in the compiler.

src/dmd/json.d
marler8997: workin on this one

src/dmd/mars.d
marler8997: done

src/dmd/json.d
marler8997: done, I was thinking I needed to use phobos `EnumMembers`...I forgot I could do this with `__traits`.

src/dmd/json.d
marler8997: ok, I ended up using vsnprintf, also added tests.

test/compilable/extra-files/json2.out
timotheecour: lol indeed... std.json really sux

test/sanitize_json.d
timotheecour: `import dmd.json : JsonField;` ?

src/dmd/json.d
timotheecour: looks like it did :) (not seeing .ptr after string literals in your latest commit) 
if not, i'm curious in what case it didn't work? i thought string literals would convert?

marler8997: The latest doesn't using `~` to append any strings.  Looks like appending strings breaks implicit conversion from `string` to `const(char)*`.

marler8997:
Ok I changed `-Xf-` to `-Xf=-`.  However, `-Xf=-` was existing functionality not introduced by this PR.

src/dmd/json.d
```diff
@@ -857,17 +880,161 @@ public:
...
+        {
+            parser.skipTrivial();
+            if (*parser.next == '\0')
+            {
+                break;
+            }
+            auto fieldName = parser.tryParseName();
+            if (fieldName is null)
+            {
+                json.generateError("invalid query at offset %d: expected name but got '%c' (0x%02x)",
```
timotheecour:
```
'%c' => `%c`
```

marler8997: I think the single-quotes makes more sense when printing a raw character.

changelog/json_query.dd
```diff
@@ -0,0 +1,23 @@
+Added the `-Xq=...` option to support limiting JSON output to a set of fields.
+
+Using the `-Xq=...` option limits the JSON output to a set of fields, i.e.
+---
```
wilzbach: The `CONSOLE` macro is specifically made for this:

```
$(CONSOLE dmd -Xq=compilerInfo,semantics)
```

see e.g. https://dlang.org/changelog/pending.html#deprecation-dylib

src/dmd/json.d
wilzbach: Just "name" isn't so helpful, how about generating an up-to-date list?

```d
enum names = () {
    string s;
    static immutable delim = ", ";
    foreach (member; __traits(allMembers, JsonField))
        s ~= member.stringof ~ delim;
    return s[0 .. $ - delim.length];
}();
```

test/sanitize_json.d
wilzbach: Yeah, the testsuite shouldn't (directly) depend on what it is testing.

wilzbach (approved):
Nice! I only found a few more nits, but it looks more than ready to go to me.
I really like the extra effort you made with adding all these additional test files :+1:

test/compilable/extra-files/json2.out
marler8997: wait, there already is an option for this! https://dlang.org/phobos/std_json.html#.JSONOptions.doNotEscapeSlashes

I'll updated the PR to use it :)

src/dmd/json.d
marler8997: The first error is caused by an invalid character, providing all the field names might be too much information in this case.  However, providing the names for the the 2nd error, which is an invalid field name definitely makes sense.  I've included this change for the 2nd error.

timotheecour: > I think the single-quotes makes more sense when printing a raw character.

then how about doing like:
src/dmd/parse.d:7425:77:                                error("mismatched string literal postfixes `'%c'` and `'%c'`", postfix, token.postfix);

 the point is it'll get syntax highlighted

timotheecour (approved):

marler8997:
I've thought about it and I'm going to make a small change.  I'm going to check the JSON query during command line parse time instead of JSON generation time.  I wasn't sure which would be better but for some reason I've changed my mind...

timotheecour:
agreed, we need to fail early

marler8997:
Hmmm, getting a bunch of failures with this message:
```
test_results/d_do_test: symbol lookup error: test_results/d_do_test: undefined symbol: _D3std9exception__T7enforceZ__TQmTbZQrFNaNfbLAxaAyamZb
```
Looks like some of the CI builders aren't able to build the `d_do_test` tool because of a linker error...I wasn't able to reproduce this in my box...

wilzbach:
Uhm I have seen this one on https://github.com/dlang/dmd/pull/7845 too and I suspect it has something to do with the recently merged https://github.com/dlang/phobos/pull/6086

wilzbach:
FYI I'm proposing to revert the Phobos PR for now (-> revert: https://github.com/dlang/phobos/pull/6130), s.t. we have time to figure it out and check why the CIs didn't catch it.

marler8997:
Ok, sounds like a good idea.  We're still getting that OOM issue on linux intermittently...wondering how we can address that one.  The autotester has 2GB of memory according to @braddr, and I think that should be enough without intermittently failing.  Maybe we can wrap the processes that are running the tests in phobos, and on OOM errors retry the test...? Not sure.

wilzbach:
>  We're still getting that OOM issue on linux intermittently...wondering how we can address that one.

Repeating what I did here: https://github.com/dlang/phobos/pull/6061 or fix the CTFE parser :P 
`std/regex/internal/tests.d` is still a really, really bad guy in terms of regex performance...

marler8997:
Ah ok, sounds like a plan.

marler8997:
@wilzbach looks like you fixed it :)

wilzbach:
BTW @marler8997 I added support for `-Xf=-` to run.dlang.io 

https://run.dlang.io/is/mA9krz - you don't even need to enter it. There's a button.

image

marler8997:
Let's get this one merged soon.  After this one I'll create a new PR that adds the rest of the fields from the originally proposed `-probe` flag and finalize the field names.  After that PR is done that should be everything to allow me to finish the `rdmd` change for single invocation.  If we get this PR and the next one done soon we can make it before the next release.

wilzbach:
> Let's get this one merged soon.

I just waiting for someone like @WalterBright or @andralex to have a brief look and approve this direction. I doubt that Walter cares much about the Json output (as long as it doesn't affect codegen) nor break anything.
If no one raises any voice until this Friday evening, I will go ahead and merge this.

> After this one I'll create a new PR that adds the rest of the fields from the originally proposed -probe flag and finalize the field names. After that PR is done that should be everything to allow me to finish the rdmd change for single invocation. If we get this PR and the next one done soon we can make it before the next release.

:+1:

src/dmd/globals.d
```diff
@@ -182,6 +182,7 @@ struct Param
 
     bool doJsonGeneration;              // write JSON file
     const(char)* jsonfilename;          // write JSON file to jsonfilename
+    uint jsonQueryFlags;                // flags for JSON query
```
WalterBright: Shouldn't this be typed as `JsonQueryFlags` ?

src/dmd/json.d
```diff
@@ -857,17 +871,178 @@ public:
...
+        s ~= prefix ~ enumName;
+        prefix = ", ";
+    }
+    s ~= "\0"; // make sure it is null-terminated
+    return s;
+}();
+
+/**
+Parses a JSON query into a set of flags.
+
```
WalterBright: `Params:` ?

src/dmd/json.d
```diff
@@ -857,17 +871,178 @@ public:
...
+Returns: false on error, prints its own error messages.
+*/
+bool parseJsonQuery(const(char)* query, uint* outQueryFlags)
```
WalterBright: `out uint outQueryFlags`

src/dmd/mars.d
```diff
@@ -1869,6 +1885,16 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
...
                     params.jsonfilename = p + 3 + (p[3] == '=');
                     break;
+                case 'q':
+                    if (!p[3])
+                        goto Lnoarg;
+                    if (p[3] != '=')
+                        goto Lerror;
+                    if (!p[4])
+                        goto Lnoarg;
+                    if (!parseJsonQuery(p + 4, &global.params.jsonQueryFlags))
```
WalterBright: I see this is cumulative to jsonQueryFlags. Should it be? There's no way to turn off a flag with them being cumulative.

src/dmd/mars.d
```diff
@@ -1869,6 +1885,16 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
...
+                        return true; // error
```
WalterBright: should do a `continue` like the rest, not `return`

src/dmd/json.d
```diff
@@ -857,17 +871,178 @@ public:
...
+{
+    auto parser = QueryParser(query);
```
WalterBright: This seems like a lot of effort to simply turn a comma-separated list of 4 identifiers into 4 flags? In fact, this could be avoided entirely by just supporting multiple -Xq=identifier queries, which you already do:
```
dmd -Xq=compilerInfo,buildInfo,semantics,modules
```
becomes:
```
dmd -Xq=compilerInfo -Xq=buildInfo -Xq=semantics -Xq=modules
```
and all this code goes away. Note we do the same thing for multiple `-version=identifier` switches, etc.

src/dmd/json.d
```diff
@@ -797,19 +797,33 @@ public:
         objectEnd();
     }
 
+    private void generateModules(Modules* modules)
```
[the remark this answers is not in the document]
timotheecour: even private? 
from https://dlang.org/dstyle.html: 
`All public declarations will be documented in Ddoc format and should have at least Params and Returns sections.` => says public

src/dmd/json.d
timotheecour: there's a precedent for that with:
* `-i=+core,-std.stdio`   
* and in clang, `-Wl,-lfoo,-lbar`
* and the suggested -unittest=std.stdio,foo.bar

maybe exposing that logic (parsing comma separated list) into reusable code would be an option

src/dmd/json.d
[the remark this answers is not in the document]
WalterBright: If they're non-trivial, yes. I don't know what `generateModules()` does. How can it generate a module when it is passed modules? Where do these generated modules go? I'm sure this is obvious to you since you wrote it, but I submit if you looked at it a year from now, you wouldn't know what it did. (This happens to me all the time.)

src/dmd/json.d
WalterBright: or `ref` as it is cumulative.

src/dmd/json.d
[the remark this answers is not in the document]
timotheecour: I'm not questioning whether it should have a comment, I'm questioning whether it should have a comment formatted as Ddoc

src/dmd/json.d
WalterBright: I would have argued against the -i syntax, on the same grounds. What clang does is not that compelling.

In any case, reinventing another parser for CSV options is not the best plan. As noted in the comments, dmd already has several different ways of specifying a file, which is a miserably stupid design, and is really hard to unwind because of subtle breakage. Let's not do that with multiple CSV parsers with same-but-maddeningly-different behaviors.

marler8997: If you prefer using multiple arguments, i.e.
```
-i=foo -i=-foo.bar -i=std -i=core
```
 I don't think it's too late to change `-i`.  It hasn't showed up in a major release yet.  I agree that consistency is virtue we should aspire to.  Just say the word and I'll create a PR for this.

src/dmd/mars.d
marler8997: All flags are off by default, you "turn then off" by not specifying them.  Are you saying that we should have a way to allow a user to turn it on in one argument, and they specify another argument to turn it off?

src/dmd/json.d
marler8997: Ugh, I hate that now the caller won't know their "value" argument could be modified, i.e.
```D
ubyte myFlags = 0;
parseJsonQuery(myQuery, myFlags);
```
Now `myFlags` was modified, but it's not clear from the call that this happened.  I would prefer to see:
```D
ubyte myFlags = 0;
parseJsonQuery(myQuery, &myFlags);
```
Now you definitely know that `myFlags` could have been modified.  That's just my preference, but this is not my code so I'll do whatever the core team prefers for `dmd`.

marler8997:
Ok changes made.

src/dmd/json.d
marler8997: I've created a PR to remove the "comma support" for `-i`: https://github.com/dlang/dmd/pull/7857/files

marler8997:
Note, now that we moved from one query (`-Xq=<query>`) to multiple fields (`-Xi=<field1> -Xi=<field2>`), I've changed the wording from "query" to "include field", hence the change from `-Xq` to `-Xi`.

src/dmd/json.d
```diff
@@ -857,17 +891,77 @@ public:
...
+
+/**
+A string listing the name of each JSON field. Useful for errors messages.
+*/
+enum jsonFieldNames = () {
+    string s;
+    string prefix = "";
+    foreach (enumName; __traits(allMembers, JsonFieldFlags))
+    {
+        if (enumName != "none")
```
wilzbach: How about:

```d
foreach (idx, enumName; __traits(allMembers, JsonFieldFlags))
{
    if (idx > 0)
    {
        s ~= ", " ~ enumName;
    }
}
```

src/dmd/json.d
```diff
@@ -857,17 +891,77 @@ public:
...
+
+Returns: JsonFieldFlags.none on error, otherwise the JsonFieldFlags value
+         corresponding to the given fieldName.
+*/
+JsonFieldFlags tryParseJsonField(const(char)* fieldName)
+{
+    auto fieldNameString = fieldName[0 .. strlen(fieldName)];
+    foreach (enumName; __traits(allMembers, JsonFieldFlags))
+    {
+        if (enumName != "none" && fieldNameString == enumName)
```
wilzbach: You could move `if (enumName != "none")` out of the loop.
That saves a few duplicated comparisons.

test/sanitize_json.d
```diff
@@ -39,25 +39,50 @@ int main(string[] args)
...
+                sanitizeSyntaxNode(modules);
+            }
+        }
+        {
+            auto semantics = rootObject.get("semantics", JSONValue.init);
+            if (semantics.type != JSON_TYPE.NULL)
+            {
+                sanitizeSemantics(semantics.object);
+            }
+        }
```
wilzbach: DRY idea: 

```d
static foreach (e; [tuple("compilerInfo", "sanitizeCompilerInfo")
                    tuple("buildInfo", "sanitizeBuildInfo")])])
{{
    auto infonfo = rootObject.get(e[0], JSONValue.init);
    if (compilerInfo.type != JSON_TYPE.NULL)
        mixin(e[1] ~ "(info.object);");
}}
```

It's okay to use newer compiler features here as DMD + DRuntime + Phobos are just freshly built from master.

src/dmd/json.d
marler8997: Hmmm, that produces a different string ", compilerInfo, ..." instead of "compilerInfo, ..."

wilzbach: Ah it's too late here. I meant:

```d
if (idx > 0)
    s ~= ", ";
s ~= enumName;
```

(the main point was about not hard-coding that the first `enumName` is named "none")

marler8997: Oh...but now you're hardcoding that the first member should be ignored...is that better than hardcoding that the value named "none" should be ignored?

marler8997: Eh...I suppose I don't have a preference for either...so we'll go with your preference :)

test/sanitize_json.d
```diff
@@ -39,25 +39,44 @@ int main(string[] args)
...
 }
 
-void sanitize(JSONValue[] rootArray)
+string capitalize(string s)
 {
-    foreach (ref obj; rootArray)
+    if (s.length == 0 || (s[0] < 'a' || s[0] > 'z'))
+        return s;
+    return (cast(char)(s[0] - ('a' - 'A'))) ~ s[1 .. $];
+}
```
wilzbach: https://dlang.org/phobos/std_string.html#.capitalize ?

wilzbach (approved):
I still like it :+1:

test/sanitize_json.d
marler8997: Looks like that one won't work, it converts the rest to lower-case...which is not what is wanted.

wilzbach: Ah that's unexpected. I thought there's a function for this in Phobos :/
Well, you could do this:

```d
import std. unit : asCapitalized;
return s.take(1).asCapitalized.chain(s.drop(1));
```

It would be more idiomatic, but put more stress on the CTFE (though it's a small standalone script for which this is probably not a concern);

marler8997: Ugh...I can't get it to work.  I added `.array` to the end but it ends up returning `dchar[]`...

wilzbach: Interesting it worked for me (https://run.dlang.io/is/xP19Bv), 

>  the end but it ends up returning dchar[]

You can use `.byCodeUnit` to opt-out of auto-decoding, but `to!string` or `text` from `std.conv` are both smart enough (e.g. https://run.dlang.io/is/zv8RlX)

For pure ASCII, there's also `std.ascii.toLower`:

```d
s.enumerate.map!(a => a.index == 0 ? a.value.toUpper : a.value).text;
```

https://run.dlang.io/is/rHnADh

But this is definitely nit-picking here ;-)

marler8997: looks like `std.conv.text` did the trick.  My `std.range` / `std.algorithm` 'fu could use work.

JinShil:
This PR could use a statement regarding motivation (at least if you need my review).  I don't have an intimate understanding of the problem this solves or the use case.  Also, why do we need multiple options?  Is there a problem with just generating everything?

Also, I don't see why it shouldn't be documented.  What's the motivation for keeping it hidden?  Are we not sure it's a good idea or not?

The code is no 100% covered by tests.  Can the lack of coverage be justified?

Overall, though, I don't have any issue with this PR, and I'm not objecting to it;  I'm actually trying to move it forward by adding my support.  But I need some understanding (answers to the questions above) before I can support it.

marler8997:
> This PR could use a statement regarding motivation (at least if you need my review)

This PR is a result of a suggestion from @WalterBright .  We needed a way of capturing information from the compiler without redirecting stdout/stderr.  I suggested providing a way to redirect verbose output to a file but Walter suggested we add the information to the JSON file, so this PR is the result.

> Also, why do we need multiple options? Is there a problem with just generating everything?

Generating everything is fine, but the output is meant for tools that can process the JSON output.  So it makes sense to allow the tool to specify what it wants.  This also has the advantage that the tool won't need to pay the overhead to generate everything if they don't need it.  Also it provides a mechanism to allow a tool to run the compiler with no source files, something DUB will do to query the compiler information.  It also maintains backwards compatibility. Existing use cases where JSON is generated is unnafected, it generates the original format, but if a tool uses `-Xi` then it will generate the new format with the requested fields.  And it will remain forward compatible because any new fields that might be added later will not be included since they won't be requested via `-Xi`.

> Also, I don't see why it shouldn't be documented. What's the motivation for keeping it hidden? Are we not sure if it's a good idea?

The only reason for not documenting it is no one has thought of a good use case for this feature to be used out in the wild.  The only current use cases are `dub` and `rdmd`, both of which are internal tools.  If it deemed to be useful outside of internal use, then documenting it makes sense.  I'm basically taking the conservative approach where I'm not including any changes that can always be decided later.

> The code is not 100% covered by tests. Can the lack of coverage be justified?

I haven't checked the code coverage. Depends on which code paths aren't taken, I'll see if I can take a look.

JinShil:
Thanks @marler8997 

Who are the primary consumers of this JSON output and for what purpose?  You mentioned `dub` and `rdmd`, so what do they need with this information?

What about feature overlap?  Isn't this information something 3rd party tools can obtain for themselves using DMD-as-a-library?

marler8997:
rdmd will use it to output all the files used during a compilation.  It uses this for dependency analysis. It will also use compilerInfo to know what features a particular compiler binary supports. Same with dub.

Dmd as a library really isn't applicable for these use cases. The JSON output is used to get information about a particular compiler binary or a particular build, whereas dmd as a library shows an application to have access to compiler functionality.

src/dmd/mars.d
```diff
@@ -998,6 +970,50 @@ private int tryMain(size_t argc, const(char)** argv)
...
+        const(char)* jsonfilename;
+        if (name && *name)
+        {
+            jsonfilename = FileName.defaultExt(name, global.json_ext);
+        }
+        else
+        {
+            if (global.params.objfiles.dim == 0)
+            {
+                error(Loc.initial, "cannot determine JSON filename, use -Xf=<file> or provide a source file");
```
JinShil: Tick marks:  e.g. \`-Xf=<file>\`

test/fail_compilation/jsonBadField1.d
```diff
@@ -0,0 +1,9 @@
+/*
+PERMUTE_ARGS:
+REQUIRED_ARGS: -Xi=UNKNOWN_FIELD_NAME
+TEST_OUTPUT:
+---
+Error: unknown JSON field `-Xi=UNKNOWN_FIELD_NAME`, expected one of compilerInfo, buildInfo, modules, semantics
```
JinShil: Please modify the code to surround each expected option in tick marks, e.g. \`compilerInfo\`, \`buildInfo\`, \`modules\`, \`semantics\`

JinShil (changes requested):

wilzbach:
> This also has the advantage that the tool won't need to pay the overhead to generate everything if they don't need it. For example, dub only needs to generate compilerInfo and generating it quickly will likely be a requirement. 

Yes, at the moment Dub creates and compiles a probing file and invocates DMD on it to parse the `pragma(msg)` output  _every time_ before it builds something  (see https://github.com/dlang/dmd/pull/7521).

> For example, 

There are many other potential use cases - run.dlang.io is another or basically any build tool that needs more information about the currently installed D compiler.

> Also, I don't see why it shouldn't be documented. What's the motivation for keeping it hidden? Are we not sure if it's a good idea?

Yep this was done, s.t. we have one or two releases to experiment with until we set it into stone.

test/fail_compilation/jsonBadField1.d
marler8997: done, thanks for the suggestion

JinShil:
Thank you @marler8997 and @wilzbach.  I see where this is going now.  I'm just waiting on coverage.

JinShil (approved):

PR #7839, opened 2018-02-04
Fix Issue 18367 - dmd should not segfault on -X with libraries, but no source files

wilzbach:
I wish all segfaults would be that easy.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -880,6 +880,12 @@ private int tryMain(size_t argc, const(char)** argv)
             }
             else
             {
+                if (global.params.objfiles.dim == 0)
+                {
+                    error(Loc.initial, "-X requires source files");
+                    fatal();
+                }
+
                 // Generate json file name from first obj name
                 const(char)* n = global.params.objfiles[0];
                 n = FileName.name(n);
--- a/test/compilable/test18367.sh
+++ b/test/compilable/test18367.sh
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -u -o pipefail
+
+name=$(basename "$0" .sh)
+dir=${RESULTS_DIR}/compilable/
+
+# dmd should not segfault on -X with libraries, but no source files
+out=$("$DMD" -conf= -X foo.a 2>&1)
+[ $? -eq 1 ] || exit 1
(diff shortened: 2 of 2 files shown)


src/dmd/mars.d
```diff
@@ -880,6 +880,9 @@ private int tryMain(size_t argc, const(char)** argv)
             }
             else
             {
+                if (global.params.objfiles.dim == 0)
```
UplinkCoder: @wilzbach indentation seems to suggest you missed braces.

UplinkCoder (changes requested):

UplinkCoder (approved):
LGTM

wilzbach:
Argh this is failing on Windows :/

timotheecour:
https://github.com/dlang/dmd/pull/7838 will supersede this behavior right? (but still good to submit in meantime)

wilzbach:
I thought you don't want `-X` to emit JSON, but just `-Xf=-`?

timotheecour:
ya, sorry ignore my comment

marler8997:
PR https://github.com/dlang/dmd/pull/7838 addresses this.  I defaulted to writing to `info.json`, however, your approach of asserting an error works as well.  Let me know what you want to do.

wilzbach:
I think producing an error is saner then writing to a default info.json file. This would just need extra documentation and the case isn't very popular anyways.

marler8997:
Ok I changed my PR to issue an error in this case:
```
Error: cannot determine JSON filename, specify using -Xf=<file> or provide at least one source file
```

test/compilable/test18367.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
```
marler8997: bash wizadry

marler8997: I copied this test to my PR, and added another case to it

timotheecour:
> I defaulted to writing to info.json, however, your approach of asserting an error works as well. Let me know what you want to do.

indeed, IMO as a general rule it's always better to be explicit (let user choose file name or give error); it reduces bugs (eg what if a user already had an info.json and didn't intend it to be deleted)

wilzbach:
Closing this in favor of #7838 then. Thanks @marler8997 !

PR #7840, opened 2018-02-04
Issue 18368 - -X should print all static information on stdout if no files are given

wilzbach:
Based on the discussion in https://github.com/dlang/dmd/pull/7521, https://github.com/dlang/tools/pull/292 and https://github.com/dlang/dmd/pull/7838 the conclusion was the following:

- `-Xq` will allow querying (see https://github.com/dlang/dmd/pull/7838)
- `-X` should support printing to stdout if no files are given to satisfy the probing use case from https://github.com/dlang/dmd/pull/7521

This is a simple version to get approval of the concept.
Moving over the full attributes from https://github.com/dlang/dmd/pull/7521 will be easy and can be done as a follow-up as for now this feature is undocumented.

As you might have noticed, this would also require refactoring the JsonWriter.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -360,8 +360,33 @@ private int tryMain(size_t argc, const(char)** argv)
     }
     if (files.dim == 0)
     {
-        usage();
-        return EXIT_FAILURE;
+        // -X without arguments is the JSON analog of --version
+        if (global.params.doJsonGeneration)
+        {
+            OutBuffer buf;
+            json_info(&buf);
+            buf.writestring("\n");
+            // write to stdout for -X and -Xf=-
+            if (!global.params.jsonfilename || strcmp(global.params.jsonfilename, "-") == 0)
+            {
+                fputs(buf.peekString(), stdout);
+                fflush(stdout);
+            }
+            else
+            {
+                ensurePathToNameExists(Loc.initial, global.params.jsonfilename);
+                auto jsonfile = new File(global.params.jsonfilename);
+                jsonfile.setbuffer(buf.data, buf.offset);
+                jsonfile._ref = 1;
+                writeFile(Loc.initial, jsonfile);
+            }
+            return EXIT_SUCCESS;
+        }
+        else
+        {
+            usage();
+            return EXIT_FAILURE;
+        }
     }
     static if (TARGET.OSX)
     {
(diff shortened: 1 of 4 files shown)


timotheecour:
We should keep things orthogonal: where we output json (file vs stdout) and what (with or without input files):
* `-Xf=outputfile` should also print all static info to outputfile if no input files are given (same as` -X`)
* outputting to stdout (eg via `-X` or whatever syntax is used) should behave exactly as `-Xf=outputfile` (except outputting to stdout) even when input files are given

* There's a change of behavior involved: `-X` currently outputs json to mainfile.json; 

* Why not make things simple and use `-Xf=-` to mean stdout? following a common convention: https://unix.stackexchange.com/questions/16357/usage-of-dash-in-place-of-a-filename

wilzbach:
> There's a change of behavior involved: -X currently outputs json to mainfile.json;

No. `dmd -X` will just print the help page, but if `-Xf=-` makes everyone happy, I don't mind to change it do this. 

> -Xf=outputfile should also print all static info to outputfile if no input files are given (same as-X)

Ok. I will change to this soon.

src/dmd/mars.d
```diff
@@ -360,8 +360,33 @@ private int tryMain(size_t argc, const(char)** argv)
...
-        usage();
-        return EXIT_FAILURE;
+        // -X without arguments is the JSON analog of --version
+        if (global.params.doJsonGeneration)
+        {
+            OutBuffer buf;
+            json_info(&buf);
+            buf.writestring("\n");
+            // write to stdout for -X and -Xf=-
+            if (!global.params.jsonfilename || strcmp(global.params.jsonfilename, "-") == 0)
```
wilzbach: If wanted I can split this if statement and print the manual for `-X` (previous behavior).

timotheecour:
what i meant was:
`dmd -X main.d` currently writes `main.json` so this PR introduces a change of behavior if `-X` were to mean `stdout`; hence the new syntax `-Xf=-` is nice because it doesn't introduce a change of behavior, and is also intuitive and standard

wilzbach:
FYI https://github.com/dlang/dmd/pull/7838 comes with the same behavior and actually already does what this PR was trying to do, so let's better move the discussion to that PR.

PR #7841, opened 2018-02-04
fix issue 18315 - wrong code for `int.min > 0`

aG0aep6G:
~~Removing the wrong optimization, falling back to cmp instruction.~~

----

I have no idea what I'm doing when it comes to dmd code, and this seems like a place where one can introduce subtle bugs. Please review carefully.

### DIFF
--- a/src/dmd/backend/cod4.c
+++ b/src/dmd/backend/cod4.c
@@ -2064,16 +2064,20 @@ void cdcmp(CodeBuilder& cdb,elem *e,regm_t *pretregs)
                     goto oplt;
                 case OPgt:
                     cdb.gen2(0xF7,grex | modregrmx(3,3,reg));         // NEG reg
-#if TARGET_WINDOS
-                    // What does the Windows platform do?
-                    //  lower INT_MIN by 1?   See test exe9.c
-                    // BUG: fix later
+                        /* Flips the sign bit unless the value is 0 or int.min.
+                        Also sets the carry bit when the value is not 0. */
                     code_orflag(cdb.last(), CFpsw);
                     cdb.genc2(0x81,grex | modregrmx(3,3,reg),0);  // SBB reg,0
-#endif
+                        /* Subtracts the carry bit. This turns int.min into 0,
+                        flipping the sign bit.
+                        For other negative and positive values, subtracting 1
+                        doesn't affect the sign bit.
+                        For 0, the carry bit is not set, so this does nothing
+                        and the sign bit is not affected. */
                     goto oplt;
                 case OPlt:
                 oplt:
+                    // Get the sign bit, i.e. 1 if the value is negative.
                     if (!I16)
                         cdb.genc2(0xC1,grex | modregrmx(3,5,reg),sz * 8 - 1); // SHR reg,31
                     else
--- a/src/dmd/backend/cod4.c
+++ b/src/dmd/backend/cod4.c
@@ -2068,8 +2068,8 @@ void cdcmp(CodeBuilder& cdb,elem *e,regm_t *pretregs)
                         Also sets the carry bit when the value is not 0. */
                     code_orflag(cdb.last(), CFpsw);
                     cdb.genc2(0x81,grex | modregrmx(3,3,reg),0);  // SBB reg,0
-                        /* Subtracts the carry bit. This turns int.min into 0,
-                        flipping the sign bit.
+                        /* Subtracts the carry bit. This turns int.min into
+                        int.max, flipping the sign bit.
                         For other negative and positive values, subtracting 1
                         doesn't affect the sign bit.
                         For 0, the carry bit is not set, so this does nothing
--- a/test/runnable/mars1.d
+++ b/test/runnable/mars1.d
@@ -1725,6 +1725,17 @@ void test16997()
 
 ////////////////////////////////////////////////////////////////////////
 
+void test18315() // https://issues.dlang.org/show_bug.cgi?id=18315
+{
+    int i = int.min;
+    bool b = i > 0;
+    assert(!b);
+    b = 0 < i;
+    assert(!b);
+}
+
+////////////////////////////////////////////////////////////////////////
+
 int main()
 {
     testgoto();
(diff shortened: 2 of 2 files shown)


timotheecour:
I think a benchmark showing effect on performance would be nice to know how much this affects dmd.
Since there was no bug with `-O` (and ldc2 isn't affected) my guess is shouldn't affect much.

aG0aep6G:
> I think a benchmark showing effect on performance would be nice to know how much this affects dmd.

For this D code:

```d
bool f(int i) { return i > 0; }
```

dmd 2.078.0 generates this:

```asm
   0:   push   rbp
   1:   mov    rbp,rsp
   4:   mov    rax,rdi
   7:   neg    eax
   9:   shr    eax,0x1f
   c:   pop    rbp
   d:   ret
```

with this PR:

```asm
   0:   push   rbp
   1:   mov    rbp,rsp
   4:   sub    rsp,0x10
   8:   mov    DWORD PTR [rbp-0x8],edi
   b:   cmp    DWORD PTR [rbp-0x8],0x0
   f:   setg   al
  12:   mov    rsp,rbp
  15:   pop    rbp
  16:   ret
```

LDC:

```asm
   0:   test   edi,edi
   2:   setg   al
   5:   ret
```

(All with `-O`.)

I'm not sure if a fair comparison includes the prologue/epilogue in this case. Anyway, here are some times with function calls:
```bash
# Compile old version.
echo 'bool f_old(int i) { return i > 0; }' | dmd2.078 - -c -ofold.o -O -betterC
# Compile new version.
echo 'bool f_new(int i) { return i > 0; }' | dmd2.git - -c -ofnew.o -O -betterC
# Try LDC too, for reference.
echo 'bool f_ldc(int i) { return i > 0; }' | ldc2 - -c -ofldc.o -O -betterC
# Compile timing code, link, run.
dmd -O old.o new.o ldc.o -run - <<EOF
bool f_old(int);
bool f_new(int);
bool f_ldc(int);
void main()
{
    import std.conv;
    import std.datetime;
    import std.datetime.stopwatch: benchmark; /* Really? */
    import std.stdio;
    auto ts = benchmark!(
        () { f_old(int.min); },
        () { f_new(int.min); },
        () { f_ldc(int.min); },
    )(10_000_000);
    writeln("old: ", ts[0].to!Duration);
    writeln("new: ", ts[1].to!Duration);
    writeln("LDC: ", ts[2].to!Duration);
}
EOF
```

Typical output:
```
old: 28 ms, 140 μs, and 2 hnsecs
new: 31 ms, 107 μs, and 2 hnsecs
LDC: 23 ms, 451 μs, and 6 hnsecs
```

> Since there was no bug with -O (and ldc2 isn't affected) my guess is shouldn't affect much.

The bug exists with `-O`. It's just a bit harder to trigger:

```d
void main(string[] args)
{
    int i = args.length != 0 ? int.min : 1;
    bool* b = new bool(i > 0);
    if (*b) assert(false);
}
```

schveiguy:
IMO, the performance loss is irrelevant (well, if it was like 10x slower, maybe we could work on it a bit). A compiler should never emit code that doesn't do what the source is telling it to do.

wilzbach (approved):
Thanks a lot. I absolutely agree that DMD shouldn't emit the optimization until its bugs have been fixed.

wilzbach:
(The Travis failure can be safely ignored - we have disabled it for `master` already)

WalterBright:
Don't pull this until I have a chance to work on it.

src/dmd/backend/cod4.c
```diff
@@ -2062,16 +2067,6 @@ void cdcmp(CodeBuilder& cdb,elem *e,regm_t *pretregs)
...
                     cdb.genc2(0x81,grex | modregrmx(3,2,reg),0);          // ADC reg,0
                     goto oplt;
-                case OPgt:
-                    cdb.gen2(0xF7,grex | modregrmx(3,3,reg));         // NEG reg
-#if TARGET_WINDOS
-                    // What does the Windows platform do?
-                    //  lower INT_MIN by 1?   See test exe9.c
-                    // BUG: fix later
-                    code_orflag(cdb.last(), CFpsw);
-                    cdb.genc2(0x81,grex | modregrmx(3,3,reg),0);  // SBB reg,0
```
WalterBright: I don't know why this section was #if'd only for Windows. The SBB instruction makes it work. No need to disable the optimization.

src/dmd/backend/cod4.c
```diff
@@ -2050,7 +2050,12 @@ void cdcmp(CodeBuilder& cdb,elem *e,regm_t *pretregs)
         if (!tyuns(tym) && !tyuns(e2->Ety) &&
             !boolres(e2) && !(*pretregs & mPSW) &&
             (sz == REGSIZE || (I64 && sz == 4)) &&
-            (!I16 || op == OPlt || op == OPge))
+            (!I16 || op == OPlt || op == OPge) &&
+            op != OPgt)
+                /*NOTE: OPgt is left out because there is currently no
+                optimization implemented for it. There used to be one but it
+                was wrong.
+                https://issues.dlang.org/show_bug.cgi?id=18315 */
```
WalterBright: No need to change this.

test/runnable/test18315.d
```diff
@@ -0,0 +1,8 @@
+void main()
+{
+    int i = int.min;
+    bool b = i > 0;
+    assert(!b);
+    b = 0 < i;
+    assert(!b);
+}
```
WalterBright: Please just add this test case to test/runnable/mars1.d, instead of putting it in its own file. The autotester runs faster with fewer files.

aG0aep6G:
Followed @WalterBright's advice and fixed the optimization by enabling the `SBB` instruction. I've also added some comments to explain how all this works. Also moved the test to mars1.d.

MartinNowak:
Why don't we just implement the `test Xdi, Xdi; setg al` solution?
https://godbolt.org/g/8gCBrV

WalterBright:
> Why don't we just implement the test Xdi, Xdi; setg al solution?

Great question.

1. The fix here is simple and not risky, and it gets the bug fixed.
2. The code gen still supports processors that don't have the `setg` instruction.
3. I plan to support `setg` when I have some time to spend on it.

PR #7843, opened 2018-02-05
Revival of DCache, now with SHA1 as hash function + dependencies

DmitryOlshansky:
Getting closer to building Phobos with DCache enabled

Reboot of https://github.com/dlang/dmd/pull/7239 moved here because the original discussion got super long.

### DIFF
--- a/src/dmd/dcache.d
+++ b/src/dmd/dcache.d
@@ -1,0 +1,376 @@
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Copyright:   Copyright (c) 1999-2017 by Digital Mars, All Rights Reserved
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/dcache.d, _dcache.d)
+ *
+ * Dcache - an implementation if a distributed cache over memory-mapped file.
+ * It doesn't try to be smart and has most basic interface in the way of get/put,
+ * both operations are synchronious and "atomic" by using a simple spin-lock.
+ * 
+ * Key is assumed to be a strong hash (e.g. MD5) of some object and its first bits are used directly.
+ *
+ * Intended use
+ *
+ * Allow DMD instances (and frontends on its base) to share CTFE computation results and such,
+ * importantly a key must reflect the state of all dependent entities such as contents of files.
+ *
+ * TODOs:
+ * - eviction policy (e.g. time based expiration)
+ * - sizing policy (is 128 Mb enough for everybody?)
+ * - better hash map algorithm (it's really dumb)
+ * - better block management (does a linear pass to clean up entries ATM)
+ */
+module dmd.dcache;
+
+import core.sys.posix.sys.mman;
+import core.sys.posix.fcntl;
+import core.sys.posix.semaphore;
+import core.sys.posix.time;
+import core.sys.posix.unistd;
+import core.stdc.errno;
+import core.stdc.stdio;
+import core.stdc.stdlib;
+import core.stdc.string;
+import core.atomic;
+
+version(Posix)
+{
+    struct DCache
+    {
+        /// Load the cache identified by 'keyspace'
+        void initialize(const(char)[] keyspace)
+        {
+            char[512] buf;
+            const(char)* home = getenv("HOME");
+            if (home == null) home = "/tmp";
+            snprintf(buf.ptr, buf.length-1, "%s/.cache", home);
+            mkdir(buf.ptr, S_IRWXU | S_IRWXG | S_IRWXO);
+            snprintf(buf.ptr, buf.length-1, "%s/.cache/dcache-%s", home, keyspace.ptr);
+            printf("Using DCache at %s\n", buf.ptr);
+            fd = open(buf.ptr, O_CREAT | O_RDWR, S_IRWXU);
+            if (fd < 0) perror("Failed to open DCache mapping");
+            if (posix_fallocate(fd, 0, DATA_SIZE + TOC_SIZE + META_SIZE) < 0) perror("Failed to fallocate DCache mapping");
+            
+            mapping = mapRange(META_SIZE + DATA_SIZE + TOC_SIZE, fd);
+            owner = cast(shared(pid_t)*)mapping.ptr;
+            nextAlloc = cast(uint*)(mapping.ptr + CACHE_LINE);
+            toc = cast(TocEntry[])mapping[META_SIZE .. META_SIZE + TOC_SIZE];
+            data = cast(Data*)(mapping.ptr + META_SIZE + TOC_SIZE);
+            if (toc.length & (toc.length-1)) abort();
+            ourPid = getpid();
+            // Lock before modification of shared data
+            {
+                lock();
+                scope(exit) unlock();
+                // If first entry is all zeros - new mapping
+                if (data.occupied == 0 && data.keySize == 0)
+                {
+                    data.keySize = DATA_SIZE - Data.sizeof;
+                    foreach (ref t; toc) t.index = uint.max;
+                }
+            }
+        }
+
+        /// Clean the cache
+        void clear()
+        {
+            lock();
+            scope(exit) unlock();
+            (cast(ubyte[])mapping)[]  = 0;
+            data.keySize = DATA_SIZE - Data.sizeof;
+            foreach (ref t; toc) t.index = uint.max;
+        }
+
+        static struct TocEntry
+        {
+            uint hash32;
+            uint index; // == uint.max if empty
+            bool empty(){ return index == uint.max; }
+        }
+        static assert(TocEntry.sizeof == 8);
+
+        static struct Data
+        {
+            uint occupied; // 0 - free block, 1 - occupied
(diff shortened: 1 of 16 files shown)


src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+import core.sys.posix.semaphore;
+import core.sys.posix.time;
+import core.sys.posix.unistd;
+import core.stdc.errno;
+import core.stdc.stdio;
+import core.stdc.stdlib;
+import core.stdc.string;
+import core.atomic;
+
+version(Posix)
```
WalterBright: Windows has memory-mapped files, too.

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+{
+    struct DCache
+    {
+        /// Load the cache identified by 'keyspace'
+        void initialize(const(char)[] keyspace)
+        {
+            char[512] buf;
```
WalterBright: = void;

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+            const(char)* home = getenv("HOME");
```
WalterBright: I have no idea what all the following string manipulation and file calls are doing without very carefully reading the code. Please add a block comment.

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+import core.stdc.stdlib;
+import core.stdc.string;
+import core.atomic;
+
+version(Posix)
+{
+    struct DCache
+    {
+        /// Load the cache identified by 'keyspace'
+        void initialize(const(char)[] keyspace)
```
WalterBright: Global comment: Please add Ddoc headers for all functions.

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+                // If first entry is all zeros - new mapping
+                if (data.occupied == 0 && data.keySize == 0)
+                {
+                    data.keySize = DATA_SIZE - Data.sizeof;
+                    foreach (ref t; toc) t.index = uint.max;
+                }
+            }
+        }
+
+        /// Clean the cache
```
WalterBright: What does "cleaning" the cache mean? Deleting it? Filling it with 0's?

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+
+        static struct TocEntry
+        {
+            uint hash32;
+            uint index; // == uint.max if empty
+            bool empty(){ return index == uint.max; }
+        }
+        static assert(TocEntry.sizeof == 8);
+
+        static struct Data
```
WalterBright: `Data` is very generic. What sort of data is it?

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+                    if (p.occupied && p.key == key)
+                    {
+                        return p.value;
+                    }
+                }
+                offset = (offset + 1) & (toc.length-1);
+            }
+        }
+
+        bool put(const(ubyte)[] key, const(ubyte)[] value)
```
WalterBright: Ddoc comment needed. What the function does, what the parameters are, what the return value means.

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+        {
+            uint h = first32bits(key);
+            size_t offset = h & (toc.length-1);
+            uint itemSize = cast(uint)(Data.sizeof + key.length + value.length);
```
WalterBright: Any possibility of integer overflow or truncation?

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+                    }
+                }
+                offset = (offset + 1) & (toc.length-1);
+            }
+        }
+
+        bool put(const(ubyte)[] key, const(ubyte)[] value)
+        {
+            uint h = first32bits(key);
+            size_t offset = h & (toc.length-1);
```
WalterBright: I assume this means `toc.length` must be a power of 2. I don't see where this is documented.

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+            uint itemSize = cast(uint)(Data.sizeof + key.length + value.length);
+            lock();
+            scope(exit) unlock();
+            for (;;)
+            {
```
WalterBright: There are lots of hashing algorithms. It'd be nice if there was an explanation of which one this is, maybe with a URL to an explanation if it's a standard one.

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+        Data* dataEnd(){ return cast(Data*)(cast(void*)data + DATA_SIZE); }
+
+        void[] mapping;
+        int fd; // file descriptor of shared memory mapping
+        shared(pid_t)* owner; // owner of a lock, points to shared memory
+        uint* nextAlloc; // last allocated block to avoid linear scan
+        pid_t ourPid; // pid of this process
+        TocEntry[] toc; // table of contents, fixed-size open-addressing hash map
```
WalterBright: Should document that the length must be a power of 2.

src/dmd/dcache.d
```diff
@@ -0,0 +1,376 @@
...
+            }
+        }
+
+        void unlock()
+        {
+            // we hold the lock, so no races
+            *owner = 0;
+        }
+
+        Data* dataEnd(){ return cast(Data*)(cast(void*)data + DATA_SIZE); }
```
WalterBright: Investigate using attributes. This one can be `pure`, for example.

src/dmd/dscope.d
```diff
@@ -16,10 +16,12 @@ import core.stdc.stdio;
 import core.stdc.string;
 import dmd.aggregate;
 import dmd.attrib;
+import dmd.dcache;
 import dmd.dclass;
 import dmd.declaration;
 import dmd.dmodule;
 import dmd.doc;
+import dmd.dimport;
```
WalterBright: why is this being imported?

src/dmd/hdrgen.d
```diff
@@ -1535,7 +1537,10 @@ public:
 
     override void visit(TemplateInstance ti)
     {
-        buf.writestring(ti.name.toChars());
+        if (ti.aliasdecl)
+            buf.writestring(ti.aliasdecl.toChars());
+        else
+            buf.writestring(ti.name.toChars());
```
WalterBright: This looks like part of another PR?

src/dmd/mars.d
```diff
@@ -1543,6 +1546,10 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 else if (p[4])
                     goto Lerror;
             }
+            else if (strcmp(p + 1, cast(char*)"cache=mmap") == 0)
```
WalterBright: please don't use strcmp in mars.d anymore.

src/dmd/dinterpret.d
```diff
@@ -720,6 +731,171 @@ extern (C++) Expression ctfeInterpretForPragmaMsg(Expression e)
...
+    else
+    {
+        SHA1 hasher;
+        hasher.start();
+        // predefined versions indicate target, bitness and asserts
+        if (global.versionids)
+        {
+            foreach (id; *global.versionids)
+            {
+                const(char)* s = id.toChars();
```
WalterBright: use `id.toString()` and don't use `strlen`

src/dmd/dinterpret.d
```diff
@@ -836,6 +1012,53 @@ private Expression interpret(FuncDeclaration fd, InterState* istate, Expressions
         }
         eargs[i] = earg;
     }
+ 
+    // Skip generated symbols and only permit caching of top-level CTFE calls
+    bool cacheable = (istate == null || istate.caller == null)
+        && !thisarg // if have this then can mutate the object as "side-effect"
+        && !fd.isNogc() // compiler bug - should not consider GC in CTFE closures
+        && strncmp(fd.ident.toChars(), "__", 2) != 0;
```
WalterBright: strncmp, strlen, strcmp, etc., must die.

src/dmd/dinterpret.d
```diff
@@ -836,6 +1012,53 @@ private Expression interpret(FuncDeclaration fd, InterState* istate, Expressions
         }
         eargs[i] = earg;
     }
+
```
WalterBright: The following large chunks of code should be factored out into separate functions.

src/dmd/dinterpret.d
```diff
@@ -836,6 +1012,53 @@ private Expression interpret(FuncDeclaration fd, InterState* istate, Expressions
         }
         eargs[i] = earg;
     }
+ 
+    // Skip generated symbols and only permit caching of top-level CTFE calls
+    bool cacheable = (istate == null || istate.caller == null)
```
WalterBright: How about having `cacheable` be off if `global.params.cache` is off? I.e. why test both repeatedly in subsequent code?

ibuclaw:
As a small security detail, shouldn't you be using SHA2?

MartinNowak:
> As a small security detail, shouldn't you be using SHA2?

Indeed would make sense, given that such a cache could be used on sites like run.dlang.io.

MartinNowak:
TODO:
- [ ] fix visibility issues when reifying cached literals
- [ ] switch to SHA2

PR #7844, opened 2018-02-05
Improve documentation of SwitchStatement

wilzbach:
Follow-up to https://github.com/dlang/dmd/pull/7842

### DIFF
--- a/src/dmd/statement.d
+++ b/src/dmd/statement.d
@@ -1559,17 +1559,17 @@ extern (C++) final class StaticAssertStatement : Statement
  */
 extern (C++) final class SwitchStatement : Statement
 {
-    Expression condition;
-    Statement _body;
-    bool isFinal;
+    Expression condition;           /// switch(condition)
+    Statement _body;                ///
+    bool isFinal;                   ///
 
-    DefaultStatement sdefault;
-    TryFinallyStatement tf;
-    GotoCaseStatements gotoCases;   // array of unresolved GotoCaseStatement's
-    CaseStatements* cases;          // array of CaseStatement's
-    int hasNoDefault;               // !=0 if no default statement
-    int hasVars;                    // !=0 if has variable case values
-    VarDeclaration lastVar;
+    DefaultStatement sdefault;      /// default:
+    TryFinallyStatement tf;         ///
+    GotoCaseStatements gotoCases;   /// array of unresolved GotoCaseStatement's
+    CaseStatements* cases;          /// array of CaseStatement's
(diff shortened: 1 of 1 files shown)


src/dmd/statement.d
```diff
@@ -1591,6 +1591,15 @@ extern (C++) final class SwitchStatement : Statement
...
     {
+        /*
+        * Checks the scope of a label for existing variable declaration.
+        *
+        * Params:
+        *   vd = variable declaration to check
+        *
+        * Returns: `true` if the variables declared in this label would be skipped
+                   `false` otherwise
+        */
```
ntrel: Seems a bit verbose, especially for a non-ddoc symbol. Does  `Params` really add anything?

ntrel (approved):

src/dmd/statement.d
jacob-carlborg: I think it _should_ have a Ddoc comment.

WalterBright: The "`false` otherwise" is redundant and can be removed.

src/dmd/statement.d
```diff
@@ -1591,6 +1591,15 @@ extern (C++) final class SwitchStatement : Statement
 
     final bool checkLabel()
     {
+        /*
+        * Checks the scope of a label for existing variable declaration.
+        *
```
WalterBright: These blank lines unnecessarily take up vertical space.

wilzbach:
Comments addressed - this adds more comments to an existing function, it's not rocket science :/
(FWIW I didn't even add the function ...)

src/dmd/statement.d
```diff
@@ -1591,6 +1591,12 @@ extern (C++) final class SwitchStatement : Statement
 
     final bool checkLabel()
     {
+        /*
+        * Checks the scope of a label for existing variable declaration.
+        * Params:
+        *   vd = variable declaration to check
+        * Returns: `true` if the variables declared in this label would be skipped.
```
JinShil: `grep`ping dmd source code revealed the prevailing style is to put the `Returns` description on the following line.

JinShil (approved):

src/dmd/statement.d
wilzbach: I like that too, but Walter's comment was:

> These blank lines unnecessarily take up vertical space.

JinShil: 😆  My comment was on Walter's behalf.  Ok, then I'm good.

PR #7845, opened 2018-02-06
Add ability to override DFLAGS env variable in test suite

JinShil:
The test suite makefile adds the import directories for druntime and phobos to the `DFLAGS` environment variable:
https://github.com/dlang/dmd/blob/c55c9be607f6939c6212759c054880b7074bff6d/test/Makefile#L173-L177
This makes it impossible to write tests for no- or minimal-runtime programming, due to the implicit import of object.d.  See #7825 for evidence.

This PR adds a `DFLAGS:` test variable that clears the `DFLAGS` environment variable enabling us to write no- or minimal-runtime tests among other potential possibilities.

### DIFF
--- a/test/d_do_test.d
+++ b/test/d_do_test.d
@@ -14,6 +14,11 @@ import std.stdio;
 import std.string;
 import core.sys.posix.sys.wait;
 
+version(Win32)
+{
+    extern(C) int putenv(const char*);
+}
+
 void usage()
 {
     write("d_do_test <input_dir> <test_name> <test_extension>\n"
@@ -56,6 +61,7 @@ struct TestArgs
     bool     compileSeparately;
     bool     link;
     string   executeArgs;
+    string   dflags;
     string[] sources;
     string[] compiledImports;
     string[] cppSources;
@@ -186,6 +192,8 @@ void replaceResultsDir(ref string arguments, const ref EnvData envData)
 bool gatherTestParameters(ref TestArgs testArgs, string input_dir, string input_file, const ref EnvData envData)
 {
     string file = cast(string)std.file.read(input_file);
+
+    findTestParameter(envData, file, "DFLAGS", testArgs.dflags);
 
     findTestParameter(envData, file, "REQUIRED_ARGS", testArgs.requiredArgs);
     if (envData.required_args.length)
@@ -564,6 +572,24 @@ int tryMain(string[] args)
 
     if (!gatherTestParameters(testArgs, input_dir, input_file, envData))
         return 0;
+
+    // Clear the DFLAGS environment variable if it was specified in the test file
+    if (testArgs.dflags !is null)
+    {
+        if (testArgs.dflags != "")
+            throw new Exception("The DFLAGS test argument must be empty: It is '" ~ testArgs.dflags ~ "'");
+
+        // `environment["DFLAGS"] = "";` doesn't seem to work on Win32 (might be a bug
+        // in std.process). So, resorting to `putenv` in snn.lib
+        version(Win32)
+        {
+            putenv("DFLAGS=");
+        }
+        else
+        {
+            environment["DFLAGS"] = "";
+        }
+    }
 
     //prepare cpp extra sources
     if (testArgs.cppSources.length)
--- a/test/fail_compilation/no_object.d
+++ b/test/fail_compilation/no_object.d
@@ -1,0 +1,15 @@
+/*
+DFLAGS:
+REQUIRED_ARGS:
+TEST_OUTPUT:
+---
+Error: cannot find source code for runtime library file 'object.d'
+       dmd might not be correctly installed. Run 'dmd -man' for installation instructions.
+       config file: (null)
+import path[0] = fail_compilation
+---
+*/
+
+// Due to the empty D FLAGS test variable specified above, only the current
+// directory is in the import path.  Therefore, compilation fails because the compiler
+// cannot locate object.d
--- a/test/Makefile
+++ b/test/Makefile
@@ -36,6 +36,10 @@
 #                        once by explicitly passing the modules to the compiler.
 #                        default: (none)
 #
+#   DFLAGS:              Overrides the DFLAGS environment variable if specified in the test.
+#                        No values are permitted; an error will be emitted if the value is not
+#                        empty.
+#
 #   EXTRA_SOURCES:       list of extra files to build and link along with the test
 #                        default: (none)
 #
--- a/test/compilable/minimal.d
+++ b/test/compilable/minimal.d
@@ -1,3 +1,4 @@
+// DFLAGS:
 // PERMUTE_ARGS:
 // POST_SCRIPT: compilable/extra-files/minimal/verify_symbols.sh
 // REQUIRED_ARGS: -c -defaultlib= runnable/extra-files/minimal/object.d
--- a/test/runnable/minimal.d
+++ b/test/runnable/minimal.d
@@ -1,3 +1,4 @@
+// DFLAGS:
 // PERMUTE_ARGS:
 // REQUIRED_ARGS: -defaultlib= runnable/extra-files/minimal/object.d
 


wilzbach:
The CI failures are correct but unrelated to this PR - see https://github.com/dlang/dmd/pull/7838#issuecomment-363388022

wilzbach:
> `DFLAGS=-I$(DRUNTIME_PATH)/import -I$(PHOBOS_PATH) -L-L$(PHOBOS_PATH)/generated/$(OS)/$(BUILD)/$(MODEL)`

Interesting. Shouldn't it overwrite the default `dmd.conf` with `-conf=` too? I guess it didn't make a difference to far because the auto-generated one is more or less identical to this command.

I fear a bit that exposing `DFLAGS` could be abused or confused with `REQUIRED_ARGS` or `PERMUTE_ARGS`. How about a `NO_DFLAGS` setting that - if set - would set the `DFLAGS` to `-conf=`? Then `REQUIRED_ARGS` / `PERMUTE_ARGS` / `ARG_SETS` could be used as usual.

JinShil:
> How about a `NO_DFLAGS` setting that - if set - would set the `DFLAGS` to `-conf=`?

`-conf=` is already being set:

https://github.com/dlang/dmd/blob/c55c9be607f6939c6212759c054880b7074bff6d/test/d_do_test.d#L677

A `NO_DFLAGS:` test variable that simply wipes out the `DFLAGS` environment variable would also work for my specific needs, but I thought the `DFLAGS:` test variable might open up few more possibilities.  I'm impartial; either way will work for me.

wilzbach:
>  I'm impartial; either way will work for me.

Same here. It was just an idea. Has someone a (strong) opinion about this?

test/d_do_test.d
```diff
@@ -565,6 +569,10 @@ int tryMain(string[] args)
     if (!gatherTestParameters(testArgs, input_dir, input_file, envData))
         return 0;
 
+    // Clear the DFLAGS environment variable if it was specified in the test file
+    if (testArgs.noDflags)
+        environment.remove("DFLAGS");
```
wilzbach: Shouldn't this set `-conf=` too? Otherwise the DFLAGS from the dmd.conf would be used.

wilzbach (approved):
Looks reasonable to me.

WalterBright:
Shouldn't
```
DFLAGS=
```
do the trick of setting it to nothing?

JinShil:
> Shouldn't `DFLAGS=` do the trick of setting it to nothing?

Where?  In `Makefile`?  If you set `DFLAGS` to nothing in `Makefile` then it applies to all tests.  The goal with this PR is to set it to nothing for only specific tests.  To do that it must be done in the `testxxx.d` file, and in there `DFLAGS=` syntax is not supported.

Originally, I created a `DFLAGS:` test variable that would override the `DFLAGS` environment variable, but then that leaves us with too many different ways to set compiler flags:  `DFLAGS`, `REQUIRED_ARGS`, `PERMUTE_ARGS`, and `ARGS_SETS`.  That will only lead to inconsistency and unpredictability in how the compiler flags are set for the test suite.  The `NO_DFLAGS` test argument allows us to clear the `DFLAGS` environment variable individually for each test, without opening the flood gates to all the different *creative* ways one might choose to set compiler flags.

It could be argued that we could simplify the test suite by removing all the other test variables and only allowing `DFLAGS` to be overridden or appended, but that would require changing every test file in the repository, and you already scolded me twice for pulling PRs that touch too many files.

WalterBright:
> The goal with this PR is to set it to nothing for only specific tests. To do that it must be done in the testxxx.d file, and in there DFLAGS= syntax is not supported.

Sorry, the `DFLAGS:` syntax. Since these are specific to each test file, there is no earthly reason to have:
```
DFLAGS:
```
in there unless one wants to clear DFLAGS.

> The NO_DFLAGS test argument allows us to clear the DFLAGS environment variable individually for each test, without opening the flood gates to all the different creative ways one might choose to set compiler flags.

Good point. Just don't allow anything for `DFLAGS:` but a blank line.

I.e. just rename `NO_DFLAGS` to `DFLAGS`. (You already must have dealt with the issue of what to do with: `NO_DFLAGS: foo`).

WalterBright:
> It could be argued that we could simplify the test suite by removing all the other test variables and only allowing DFLAGS to be overridden or appended

Not by me.

wilzbach:
> Good point. Just don't allow anything for DFLAGS: but a blank line.

I bet this will confuse people and they will waste time wondering why `DFLAGS` doesn't take effect or even assume that something is checked which isn't.
Please let's use `NO_DFLAGS` - it's absolutely clear what's going on and can't be accidentally misused.

JinShil:
> I bet this will confuse people and they will waste time wondering why `DFLAGS` doesn't take effect or even assume that something is checked which isn't.
>Please let's use `NO_DFLAGS` - it's absolutely clear what's going on and can't be accidentally misused.

I agree.  @WalterBright, can we convince you to keep this PR as is and support `NO_DFLAGS` for the reason given above?

test/d_do_test.d
```diff
@@ -187,6 +188,9 @@ bool gatherTestParameters(ref TestArgs testArgs, string input_dir, string input_
 {
     string file = cast(string)std.file.read(input_file);
 
+    string ignore;
```
marler8997: I would put `ignore` in it's own scope
```
{
    string ignore;
    testArgs.noDflags = findTestParameter(envData, file, "NO_DFLAGS", ignore);
}
```

marler8997 (approved):

marler8997:
I wonder how many tests don't need druntime/phobos?  Probably the majority correct?  Maybe the better solution is to have DFLAGS be "opt-in" instead of "opt-out"?

JinShil:
> I wonder how many tests don't need druntime/phobos? 

I bet most, if not all, don't need Phobos, but I'd be willing to bet 99.9% need druntime simply because  of `object.d`.  Afterall that's where fundamentals of the language like `Object`, `string`, and `size_t` are defined are declared.

WalterBright:
> I bet this will confuse people

How? I set/reset env variables all the time, and setting them to nothing is `VAR=`. There's nothing confusing about it.

> and they will waste time wondering why DFLAGS doesn't take effect or even assume that something is checked which isn't.

If they set `DFLAGS=something` then they should get an error message. I don't see how that is confusing.

marler8997:
> I bet most, if not all, don't need Phobos, but I'd be willing to bet 99.9% need druntime simply because of object.d.

Ah ok, nevermind then :)

WalterBright:
@wilzbach Keep in mind that you have another PR with `Use -conf= to overwrite the default dmd config`, not `-no_conf` :-)

marler8997:
`DFLAGS=` works in .sh files, but doesn't work in `.d` files does it?

JinShil:
Yes, I believe we'll have to use `DFLAGS=` in .sh files and `DFLAGS:` in .d files if we go with @WalterBright's preference.

marler8997:
`DFLAGS:` for .d files and `DFLAGS=` for .sh files seems more consistent.  But `NO_DFLAGS:` seems more intuitive than `DFLAGS:`...sounds like a wash to me

wilzbach:
> How? I set/reset env variables all the time, and setting them to nothing is VAR=. There's nothing confusing about it.
> ...

I was talking about things like:

```d
// DFLAGS: -version=Foo

version(Foo) {
 ... // this is never executed, but assumed by contributor and reviewers to be executed & tested
}
```

> But NO_DFLAGS: seems more intuitive than DFLAGS:...sounds like a wash to me

Maybe it should have been named `NO_CONF`
Well, there are already these accepted in-test boolean variables:

- `COMPILE_SEPARATELY` (set to enable)
- `LINK`  (set to enable)

> If they set DFLAGS=something then they should get an error message. I don't see how that is confusing.

Ok. We should do this no matter how the variable is named.

> `@wilzbach Keep in mind that you have another PR with Use -conf= to overwrite the default dmd config, not -no_conf
```

That would imply we should use `CONF:`, no?
Not a bad idea!

CC @CyberShadow who has always shown a lot of wisdom in such CLI usability questions.

**Comment by JinShil** [comment]
OK, I've implemented `DFLAGS:`.  `d_do_test.d` will throw an error if it is not empty.

I still need to figure out why this isn't working for Windows though.

**Comment by wilzbach** [comment]
> I still need to figure out why this isn't working for Windows though.

Looks strange:

```
..\generated\windows\release\32\dmd.exe -conf= -m32 -Ifail_compilation -verrors=0    -odtest_results\fail_compilation -oftest_results\fail_compilation\no_object_0.obj  -c fail_compilation\no_object.d  -map nul.map
Test failed: expected rc == 1, but exited with rc == 0
```

Maybe add `-v` to the arguments to see more on the log?

File: test/d_do_test.d [hunk 3]
```diff
@@ -565,6 +568,16 @@ int tryMain(string[] args)
...
         return 0;
 
+    // Clear the DFLAGS environment variable if it was specified in the test file
+    if (testArgs.dflags !is null)
+    {
+        if (testArgs.dflags != "")
+            throw new Exception("The DFLAGS test argument must be empty: It is '" ~ testArgs.dflags ~ "'");
+
+        environment["DFLAGS"] = "";
+        writefln("DFLAGS='%s'", environment["DFLAGS"]);
```
marler8997: This print statement is going to go to the main test log, which may not be what was intended.

WalterBright:
> `// DFLAGS: -version=Foo`

And that would give an error message: "no argument allowed for DFLAGS"

WalterBright (approved):

test/d_do_test.d
marler8997: Ah OK.  I find it helpful to include a comment in these cases so I don't forget.
```
// REMOVE ME BEFORE MERGE
writefln("...");
```

It's funny because I mainly develop on windows, but after hours of failure I still haven't been able to get the DMD test suite running on windows.  So I have to SSH into my linux box to run the test suite.  I don't think there is even a WIKI or documentation on how to get the test suite running on windows.  Definitely something we should setup...then maybe I could finally get it to work :)

marler8997: https://issues.dlang.org/show_bug.cgi?id=18413

JinShil:
Interesting:
Win32:  `import    object	(..\..\druntime\import\object.d)`  --> WRONG!
Win64:  `import path[0] = fail_compilation`  --> Correct!

rainers:
Clearing the environment variable with `environment["DFLAGS"] = ""` doesn't work. Using `putenv` helps, but the actual issue is probably in std.process.

JinShil:
I tried calling Win32 API function `SetEnvironmentVariable`, but that didn't work.  @braddr, is the DFLAGS variable set in the "System" environment variables on the Win32 machine?

wilzbach:
FYI: That's all I know about:

https://github.com/braddr/at-client/blob/master/src/do_test_dmd.sh
https://github.com/braddr/at-client/blob/master/src/diff-dmd-win64.diff

marler8997:
It looks like `d_do_test` is using the `system` function to invoke the compiler.  I believe that function will just inherit its environment variables from the parent process, however, I don't think setting `environment["DFLAGS"] = ""` will actually change the process's environment variables.  Did you try `putenv`?  If that doesn't work, we could probably use a different function to invoke the compiler (other than `system`) that could accept environment variables to pass to the child process.

JinShil:
> It looks like `d_do_test` is using the `system` function to invoke the compiler. I believe that function will just inherit its environment variables from the parent process

Isn't the process executing `d_do_test` the parent process?  If so, the environment variables are cleared there before the call to `system`, so it should work, at least as I understand it.

>   Did you try putenv?

`putenv` is a POSIX function, isn't it?  The only platform that doesn't work is Win32.  Even Win32_64 works.  There's something odd about Win32.

marler8997:
Ah, it looks like druntime actually does call SetEnvironmentVariable when you set `environment["DFLAGS"] = ""`.  Maybe `win32` isn't trying to import `object.d`.  Maybe try adding something into the test file that explicitly depends on it?
```D
// string is defined in object.d, this should guarantee that object.d is getting loaded
static assert(string.sizeof > 0);
```

test/fail_compilation/no_object.d
```diff
@@ -0,0 +1,15 @@
+/*
+DFLAGS:
+REQUIRED_ARGS:
```
wilzbach: Try `-v` here to debug win32

JinShil: I did.  See the result here: https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3027610&isPull=true

> import    object	(..\..\druntime\import\object.d)

I don't know where that's coming from.

JinShil: > (edit: I just tried this for you)

You'll probably want to disable other platforms or the auto-tester will fail and will never get around to running the test in question.

wilzbach: Oh sorry. Feel free to drop my commit. 
We should really print `D_FLAGS` with `-v` too ...

wilzbach: > We should really print D_FLAGS with -v too ...

https://github.com/dlang/dmd/pull/7865 - this might help you.

rainers:
> putenv is a POSIX function, isn't it?

The OMF library snn.lib has putenv, too, it just isn't declared in core.stdc.stdlib. Using `putenv("DFLAGS=");` with an appropriate declaration worked for me for yesterday.

JinShil:
Thanks @rainers.  Using `putenv` from snn.lib seems to have worked.  I'll do some more testing on `std.process.environment` and either submit a PR to phobos, or file a bug.  For now, I've implemented a workaround just for Win32.

This should now be ready to go.

marler8997:
Should a bug be filed for `std.process`?

JinShil:
> Should a bug be filed for std.process?

Yes.  If I can create an isolated test case, I'll file a bug.

JinShil:
Phew! What a chore that turned out to be.  I had to make a couple of changes (it worked once and then started failing again... weird).  This should be ready to go now.

I'll file a bug for std.process after I get in front of a Windows machine, and assuming I can reproduce the problem in an isolated test.

test/d_do_test.d
```diff
@@ -565,6 +573,25 @@ int tryMain(string[] args)
...
+    {
+        if (testArgs.dflags != "")
+            throw new Exception("The DFLAGS test argument must be empty: It is '" ~ testArgs.dflags ~ "'");
+
+        // `environment["DFLAGS"] = "";` doesn't seem to work on Win32 (might be a bug
+        // in std.process). So, resorting to `putenv` in snn.lib
+        version(Win32)
+        {
+            char[] varExpression = cast(char[])"DFLAGS=\0";
+            putenv(varExpression.ptr);
```
rainers: Declaring `extern(C) int putenv(const char *);` globally and calling `putenv("DFLAGS=");` worked for me. The literal already has the trailing zero and converts implicitly.

rainers:
I suspect the issue is that SetEnvironmentVariable is called with an empty string, while removal should be done with a null argument. Not sure why this works for Win64, though.

marler8997:
Hmmm, but deleting DFLAGS or setting it to an empty string should still produce the same result...

JinShil:
> I suspect the issue is that SetEnvironmentVariable is called with an empty string, while removal should be done with a null argument

I tried calling `SetEnvironmentVariable` directly with both null and empty strings, and neither worked.

marler8997:
Might have something to do with mixing native windows functions with posix. The system posix function may not pick up environment variables from SetEnvironmentVariable

test/d_do_test.d
JinShil: Done.

rainers:
> Hmmm, but deleting DFLAGS or setting it to an empty string should still produce the same result...

Maybe the empty string is silently ignored. Both snn.lib and libcmt.lib implement putenv via SetEnvironmentVariable and convert it to the null parameter for removal.

wilzbach (approved):

wilzbach:
Thanks a lot @JinShil @rainers @marler8997 for this diligent debugging work. We should really look into fixing `std.process` :O

PR #7846, opened 2018-02-06
Use -conf= to overwrite the default dmd config

wilzbach:
`-conf=` should almost always be used in any of our Makefiles to avoid any issues with user-defined configuration files.

Note: there's currently an permananet failure on some CIs. See also https://github.com/dlang/dmd/pull/7838#issuecomment-363388022

### DIFF
--- a/test/compilable/issue17167.sh
+++ b/test/compilable/issue17167.sh
@@ -14,6 +14,6 @@ src="$bin_base.d"
 echo 'void main() {}' > "${src}"
 
 # Only compile, not link, since optlink can't handle long file names
-$DMD -m"${MODEL}" "${DFLAGS}" -c -of"${bin}" "${src}"
+$DMD -m"${MODEL}" -c -of"${bin}" "${src}"
 
 rm -rf "${OUTPUT_BASE}"
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -119,14 +119,8 @@ coverage()
     make -j$N -C ../druntime -f posix.mak MODEL=$MODEL PIC="$PIC"
     make -j$N -C ../phobos -f posix.mak MODEL=$MODEL PIC="$PIC"
 
-    # FIXME
-    # Building d_do_test currently uses the host library for linking
-    # Remove me after https://github.com/dlang/dmd/pull/7846 has been merged (-conf=)
-    if [ -f ~/dlang/install.sh ] ; then
-        deactivate
-    else
-        sudo rm -rf /dlang
-    fi
+    # We don't want to end up using the host compiler in the testsuite
+    deactivate
 
     # FIXME
     # Temporarily the failing long file name test has been removed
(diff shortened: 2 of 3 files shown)


JinShil:
`-conf=` is already set in the execution of each test.  See https://github.com/dlang/dmd/blob/c55c9be607f6939c6212759c054880b7074bff6d/test/d_do_test.d#L677

marler8997:
> `-conf=` is already set in the execution of each test...

I believe that only affects `.d` tests, not `.sh` tests.  Setting it in the Makefile should affect ALL tests.

JinShil:
Then, shouldn't this PR also remove `conf=` from `d_do_test.d`?

marler8997:
> Then, shouldn't this PR also remove conf= from d_do_test.d?

I can't think of a reason that `d_do_test` would need to redundantly set `-conf=`, so I suppose that should work.

marler8997:
One thought (somewhat unrelated).  Sometimes it's nice to be able to copy the command line and run it on it's own (add/remove arguments like `-v`) and the more arguments that depend on the Makefile setting up the environment, the more arguments you have to manually add back into the command line.  Maybe redundancy is ok in this case.

test/compilable/issue17167.sh
```diff
@@ -14,6 +14,6 @@ src="$bin_base.d"
 echo 'void main() {}' > "${src}"
 
 # Only compile, not link, since optlink can't handle long file names
-$DMD -m"${MODEL}" "${DFLAGS}" -c -of"${bin}" "${src}"
```
wilzbach: It should be `"${DFLAGS[@]}" if expansion to multiple arguments is wanted, but DFLAGS is already an environment variable, so this isn't needed.
BTW it might be worthwhile to set `-m"${MODEL}"` in dflags too to avoid accidentally forgetting `-m"${MODEL}"` when it matters in the bash scripts.

wilzbach:
> I can't think of a reason that d_do_test would need to redundantly set -conf=, so I suppose that should work.

I left it in d_do_test because the environment flags are printed on a failure.
So when a test failed, you get this nice output with all parameters that were used to run the file:

```
Test compilable/test10056.d failed.  The logged output:
../generated/linux/release/64/dmd -conf= -m64 -Icompilable  -fPIC  -odtest_results/compilable -oftest_results/compilable/test10056_0.o  -c compilable/test10056.d 
```

This isn't so important for the bash scripts, because you can't run them directly anyways and you see all variables used thanks to xtrace.

wilzbach:
Nope and we shouldn't be depending on the user to set environment variables to repeat the run of a test. I know in theory we could do `DFLAGS="..." generated...`, but that I think the current behavior of d_do_test is rather nice.
BTW d_do_test enforces DFLAGS to be an empty string to avoid exactly these problems:

https://github.com/dlang/dmd/blob/e1f2be6f7103a0ac35bd7e37c8be4404bb4ad1fc/test/d_do_test.d#L603

So the "redundant" `-conf=` is really required.

marler8997 (approved):

JinShil:
> Then, shouldn't this PR also remove conf= from d_do_test.d?

Please address this.

Geod24:
Ping @wilzbach

RazvanN7:
The Makefile for running tests is deprecated. Closing this.

PR #7847, opened 2018-02-06
Fix indentation of the lambda comparison changelog

wilzbach:
Fixes this issue:

image

at https://dlang.org/changelog/pending.html#lambdacomp found by @schveiguy in https://github.com/dlang/phobos/pull/6132.

The changelog tool currently automatically inserts `$(P)` (like Markdown).
It can be argued whether this is a good or bad behavior, but for now, let's fix the changelog.

### DIFF
--- a/changelog/lambdacomp.dd
+++ b/changelog/lambdacomp.dd
@@ -9,7 +9,6 @@ $(UL
     $(LI the lambda function arguments must not have a template instantiation
     as an explicit argument type. Any other argument types (basic,
     user-defined, template) are supported.)
-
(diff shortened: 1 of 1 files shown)


schveiguy:
ah, now I see. I was working on getting changed.d to run locally to see WTF was happening :)

wilzbach:
Works:

image

>  I was working on getting changed.d to run locally to see WTF was happening :)

It should be as easy as running `make -f posix.mak pending_changelog` in `dlang.org` - everything else should automatically boostrap itself.

PR #7853, opened 2018-02-09
refactor func.d: add const

WalterBright:
Progress in using `const`

### DIFF
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -1054,17 +1054,17 @@ extern (C++) class FuncDeclaration : Declaration
         return buf.extractString();
     }
 
-    final bool isMain()
+    final bool isMain() const
     {
         return ident == Id.main && linkage != LINK.c && !isMember() && !isNested();
     }
 
-    final bool isCMain()
+    final bool isCMain() const
     {
         return ident == Id.main && linkage == LINK.c && !isMember() && !isNested();
     }
 
-    final bool isWinMain()
+    final bool isWinMain() const
     {
         //printf("FuncDeclaration::isWinMain() %s\n", toChars());
         version (none)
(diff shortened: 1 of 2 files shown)


src/dmd/func.d
```diff
@@ -1561,21 +1561,26 @@ extern (C++) class FuncDeclaration : Declaration
...
                 printf("\tmember of %s\n", cd.toChars());
-        }
-        return isMember() && (Declaration.isFinal() || ((cd = toParent().isClassDeclaration()) !is null && cd.storage_class & STC.final_));
+        }}
+        if (!isMember())
+            return false;
+        if (Declaration.isFinal())
+            return true;
+        auto cd = toParent().isClassDeclaration();
+        return (cd !is null && cd.storage_class & STC.final_);
```
marler8997: *NIT
unhelpful parens? 
```D
return cd !is null && cd.storage_class & STC.final_;
```
 Maybe you meant to do:
```D
return (cd !is null) && (cd.storage_class & STC.final_);
```

marler8997 (approved):

marler8997 (approved):

JinShil:
Doesn't this also need to be reflected in the C++ headers?

WalterBright:
For virtual functions, not really since they are accessed through the vtbl[], but I added them anyway.

JinShil (approved):

JinShil:
Looks like all tests in Jenkins passed, but the GitHub status didn't update.  Therefore, I'm manually merging.

wilzbach:
> Looks like all tests in Jenkins passed, but the GitHub status didn't update. 

FYI: https://github.com/dlang/ci/issues/133 (the GH API rate limit sometimes gets exceeded, it's probably due to Martin using the same token as it's used for dlang-bot)

PR #7855, opened 2018-02-09
SemaphoreCI: Don't activate the host compiler

wilzbach:
Maybe that fixes the SemaphoreCI issue.


wilzbach:
this has been fixed in https://github.com/dlang/dmd/pull/7848 - thanks @MartinNowak :)

PR #7857, opened 2018-02-09
Removed comma support for -i (-i=a,b becomes -i=a -i=b)

marler8997:
@WalterBright has suggested that supporting multiple arguments for a single option using the comma character is "a lot of effort".  He also mentions other command-line options already support multiple arguments because they can be given multiple times.  This PR removes comma support for `-i` so the following
```
-i=foo,-foo.bar,baz
```
must now be specified via:
```
-i=foo i=-foo.bar -i=baz
```

### DIFF
--- a/changelog/includeimports.dd
+++ b/changelog/includeimports.dd
@@ -1,10 +1,9 @@
-Automatically include imports via -i command line option
+Added the -i command line option to automatically include imports
 
-Added the command line option -i which causes the compiler to treat imported modules as if they were given on the command line.  The option also accepts "module patterns" that include/exclude modules based on their name.  For example, the following will include all modules whose names start with "foo", except for those that start with "foo.bar":
----
-dmd -i=foo,-foo.bar
----
+Added the command line option -i which causes the compiler to treat imported modules as if they were given on the command line.  The option also accepts "module patterns" that include/exclude modules based on their name.  For example, the following will include all modules whose names start with `foo`, except for those that start with `foo.bar`:
+
+$(CONSOLE dmd -i=foo -i=-foo.bar)
+
 The option -i by itself is equivalent to:
----
-dmd -i=-std,-core,-etc,-object
----
+
+$(CONSOLE dmd -i=-std,-core,-etc,-object)
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -311,10 +311,10 @@ dmd -cov -unittest myprog.d
             "look for imports also in directory"
         ),
         Option("i",
-            "same as -i=-std,-core,-etc,-object"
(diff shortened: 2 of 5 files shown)


wilzbach:
(If we go with this, changelog and documentation should be updated too.)
Luckily everything is now in this repo.

marler8997:
I've modified the changelog, and we can delay integrating the documentation PR until this PR is settled.

JinShil (approved):
Comma or no comma is fine with me.

thewilsonator:
> @WalterBright has suggested that supporting multiple arguments for a single option using the comma character is "a lot of effort".

Really? It shouldn't be hard to do what we have in LDC, namely a generic comma separated command line option parser.

timotheecour:
totally agree and strongly support comma separated list of arguments for this flag and others. I planned to add one for Comma separated linker flags to ease adapting these from, say, clang or gcc -Wl,-lfoo1,-lfoo2

ldc, clang and others do it for good reason. Some push back desirable here. Original cmdline design had a number of issues (eg flags without = separator eg -ofbar is just bad). Allowing comma separated flags improve things a bit. If parsing for all comma separated flags was reusing some common code, maybe it would ease some of walters concerns.

marler8997:
> If parsing for all comma separated flags was reusing some common code, maybe it would ease some of walters concerns.

I'll create a PR to see how complex this would be and we can see what Walter thinks.

marler8997:
Ok, PR is here: https://github.com/dlang/dmd/pull/7863/files

The implementation creates a struct called `OptionRange` that takes care of splitting an option's arguments into a range of null-terminated strings, so then when you parse the option all you have to do is pass the argument to this `OptionRange` and iterate over all the values.

changelog/includeimports.dd
```diff
@@ -1,10 +1,9 @@
...
-Added the command line option -i which causes the compiler to treat imported modules as if they were given on the command line.  The option also accepts "module patterns" that include/exclude modules based on their name.  For example, the following will include all modules whose names start with "foo", except for those that start with "foo.bar":
----
-dmd -i=foo,-foo.bar
----
 The option -i by itself is equivalent to:
----
-dmd -i=-std,-core,-etc,-object
----
+
+$(CONSOLE dmd -i=-std,-core,-etc,-object)
```
ZombineDev: I thought this syntax was removed with this PR.

timotheecour: indeed, unless https://github.com/dlang/dmd/pull/7863 gets merged before we release

marler8997: Shoot, missed that one

marler8997: @ZombineDev do you want to make a PR for that one or shall I?  I think at this point it's gotta be merged to the `stable` branch

PR #7858, opened 2018-02-09
CircleCi: deactivate the host environment before running the testsuite

wilzbach:
See also: https://github.com/dlang/dmd/pull/7848 and https://github.com/dlang/dmd/pull/7846

### DIFF
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -120,6 +120,15 @@ coverage()
     make -j$N -C ../phobos -f posix.mak MODEL=$MODEL PIC="$PIC"
 
     # FIXME
+    # Building d_do_test currently uses the host library for linking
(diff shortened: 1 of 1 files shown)


wilzbach:
Ping @JinShil or @CyberShadow or other people with merge rights lurking around.
We managed to break one of our CIs again :/

I think @marler8997 has summarized the motivation of this nicely in one sentence:

> Lol, you can't merge the fix [#7846]...so you have to workaround it temporarily

tl;dr: it turns out that we only resolved the SemaphoreCi issues in #7848, but not the CircleCi ones.

The cause of this was again https://github.com/dlang/phobos/pull/6086 and this time I would prefer to use the temporary workaround and _not_ revert it again.
Once the real fix (#7846) is in, we can easily remove the workaround here and even prove that the proper fix works as expected.

JinShil (approved):

PR #7859, opened 2018-02-09
fix issue 18403 - [REG2.078.2] Access violation when dmd tries to pri…

aG0aep6G:
…nt highlighted code

... by moving the array declaration before the static constructor. This
is a workaround for issue 18412 - [REG2.077.0] immutable array in library
becomes null when referenced in static constructor.


aG0aep6G:
I don't know how to add a test for this. Maybe #7823 is needed first?

marler8997:
What's the problem with adding a test?  Is the syntax highlighting not turned on on the autotester...and if so, do we know why it isn't turned on?

marler8997:
Oh I see, yes looks like you would need to wait for #7823

aG0aep6G:
It seems impossible to add a test for this that would fail without this PR.

The auto tester uses 2.068.2 to build dmd. But issue 18412 only strikes when building with 2.077.0 or newer. So any test is going to pass the auto tester with and without this PR.

Issue 18403 (the one being fixed here) is also triggered by code blocks in Ddoc comments. There are already tests for that in the suite (e.g. test/compilable/ddoc1.d), and they fail when a new dmd is used to build the one that's tested. They succeed on the auto tester because 2.068.2 is used to build dmd there.

ibuclaw:
Have you tried bisecting which was the first bad commit?  Digger could help you in automating that.

ibuclaw:
Without test should be fine, as this is a release regression, and fixing the binaries as soon as should be first priority. Ping @MartinNowak.

JinShil:
I suggest one of the following options to deal with the lack of a test:
1.  Add a bugzilla entry with a link to this PR saying that this needs a test when the host compilers are updated
2.  Add test in this PR, but disable it for all platforms, and add a comment explaining that the test should be enabled when the host compiler is upgraded.

Would be very helpful to explicitly declare which version of the host compiler the test can be added/enabled for.

WalterBright (approved):

aG0aep6G:
> I suggest one of the following options to deal with the lack of a test:
> 
> 1. Add a bugzilla entry with a link to this PR saying that this needs a test when the host compilers are updated
> 2. Add test in this PR, but disable it for all platforms, and add a comment explaining that the test should be enabled when the host compiler is upgraded.
> 
> Would be very helpful to explicitly declare which version of the host compiler the test can be added/enabled for.

Issue 18403 is triggered by (A) colored compiler messages, (B) code blocks in documentation.

For B, the tests are already there. They would have caught issue 18403 if the host compiler were newer. There's no point in adding more tests for this.

For A, #7823 is going to add a test, if they can figure out how to do it. It's clear that colored messages are completely untested without #7823.

I think we're good test-wise.

JinShil (approved):

ibuclaw:
@JinShil -
There is already a bug.  This just hot fixes the compiler.

https://issues.dlang.org/show_bug.cgi?id=18412

ibuclaw (approved):

MartinNowak:
> The auto tester uses 2.068.2 to build dmd. But issue 18412 only strikes when building with 2.077.0 or newer. So any test is going to pass the auto tester with and without this PR.

Travis-CI (now Semaphore CI) does a recursive (self-hosted) build and runs the test suite afterwards.

aG0aep6G:
> Travis-CI (now Semaphore CI) does a recursive (self-hosted) build and runs the test suite afterwards.

But it doesn't test Windows, or why did it not catch the bug?

MartinNowak:
> But it doesn't test Windows, or why did it not catch the bug?

Right, only saw later that it's a Windows only thing, just bisected the underlying issue.

PR #7860, opened 2018-02-10
Removed set -x in test 13742.sh

marler8997:
I think @MartinNowak may have accidentally left in a `set -x` in the test13742.sh BASH script (https://github.com/dlang/dmd/pull/7719/commits/69e7bd18892df18f682906916bf8c8ad87fb30e5).  This PR removes it.

### DIFF
--- a/test/runnable/test13742.sh
+++ b/test/runnable/test13742.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 
-set -x # debug windows issues
 set -ueo pipefail
 
 src=runnable${SEP}extra-files


wilzbach (approved):
Yes, but do you care?
I mean, you don't see the log if the test doesn't fao; anyhow...

marler8997:
Actually I thought I was seeing it show up in the main test log...if it doesn't then I suppose it doesn't really matter...let me confirm

marler8997:
Yeah, confirmed the logging is showing up in the main test log even when the test is passing.   Which is odd...maybe `set -x` is going to `stderr` and `stderr` from BASH tests is going to the main log (unlike `stdout`)

marler8997:
https://auto-tester.puremagic.com/show-run.ghtml?projectid=14&runid=3027053&dataid=21429848&isPull=true

Search for "test13742.sh"

wilzbach:
> .maybe set -x is going to stderr and stderr fro

[Yes](https://askubuntu.com/questions/629705/how-to-pipe-output-of-traced-bash-debugging)

marler8997:
Actually it looks like both `stdout` and `stderr` go to the main test log for BASH tests...

marler8997:
Looking at a test like `compilable/test9680.sh`, it looks like BASH scripts are responsible for directing their own output to `test_results/<category>/<test>.sh.out`

PR #7861, opened 2018-02-10
optimize.d: small refactor

WalterBright:
Minor stuff.

### DIFF
--- a/src/dmd/optimize.d
+++ b/src/dmd/optimize.d
@@ -42,24 +42,44 @@ extern (C++) Expression expandVar(int result, VarDeclaration v)
 {
     //printf("expandVar(result = %d, v = %p, %s)\n", result, v, v ? v.toChars() : "null");
 
+    /********
+     * Params:
+     *  e = initializer expression
+     */
+    Expression initializerReturn(Expression e)
+    {
+        if (e.type != v.type)
+        {
+            e = e.castTo(null, v.type);
+        }
+        v.inuse++;
+        e = e.optimize(result);
+        v.inuse--;
+        //if (e) printf("\te = %p, %s, e.type = %d, %s\n", e, e.toChars(), e.type.ty, e.type.toChars());
+        return e;
+    }
+
+    static Expression nullReturn()
+    {
+        return null;
+    }
+
     static Expression errorReturn()
     {
         return new ErrorExp();
     }
 
-    Expression e = null;
     if (!v)
-        return e;
+        return nullReturn();
     if (!v.originalType && v.semanticRun < PASS.semanticdone) // semantic() not yet run
         v.dsymbolSemantic(null);
-    if (v.isConst() || v.isImmutable() || v.storage_class & STC.manifest)
+    if (v.type &&
+        (v.isConst() || v.isImmutable() || v.storage_class & STC.manifest))
     {
-        if (!v.type)
-        {
-            return e;
-        }
         Type tb = v.type.toBasetype();
-        if (v.storage_class & STC.manifest || v.type.toBasetype().isscalar() || ((result & WANTexpand) && (tb.ty != Tsarray && tb.ty != Tstruct)))
+        if (v.storage_class & STC.manifest ||
+            tb.isscalar() ||
+            ((result & WANTexpand) && (tb.ty != Tsarray && tb.ty != Tstruct)))
         {
             if (v._init)
             {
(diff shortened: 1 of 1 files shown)


src/dmd/optimize.d
```diff
@@ -1167,15 +1178,15 @@ extern (C++) Expression Expression_optimize(Expression e, int result, bool keepL
...
     }
 
-    scope OptimizeVisitor v = new OptimizeVisitor(result, keepLvalue);
-    Expression ex = null;
-    v.ret = e;
+    scope OptimizeVisitor v = new OptimizeVisitor(e, result, keepLvalue);
 
     // Optimize the expression until it can no longer be simplified.
-    while (ex != v.ret)
+    while (1)
```
JinShil: `while(true)` please.

JinShil (changes requested):

src/dmd/optimize.d
```diff
@@ -188,18 +188,29 @@ extern (C++) Expression fromConstInitializer(int result, Expression e1)
     return e;
 }
 
+/*********************************
+ * Constant fold an Expression.
+ * Params:
+ *      e = expression to const fold; this may get modified in-place
+ *      result = WANTvalue, WANTexpand, or both
+ *      keepLvalue = `e` is an lvalue, and keep it as an lvalue since it is
+ *                   an argument to a `ref` or `out` parameter, or the operand of & operator
```
JinShil: tick marks around & maybe?

JinShil (changes requested):

src/dmd/optimize.d
WalterBright: `while (1)` is extensively used in dmd. `while (true)` is used 5 times.

src/dmd/optimize.d
WalterBright: ok

src/dmd/optimize.d
JinShil: Then maybe it's time to change the others in another PR.  Let's try to get rid of some of this legacy cruft.

WalterBright:
While waiting, I decided to eliminate the `goto` statements in `expandVar()`, too.

src/dmd/optimize.d
WalterBright: No, I prefer `while (1)`. Short & sweet.

WalterBright: The reason I redid that loop in the first place is it took me some time to verify that the loop was correct. I redid it so that it was quicker to see that it was correct. The latter is the kind of refactorings we need in dmd - ones that make the code flow in a simpler, less convoluted manner.

The same goes for redoing `expandVar()`. The use of nested functions makes the logic flow much more understandably.

WalterBright: For another example, see line 225. I relaid out the fields and the constructor parameters and the `Params:` to match the sequence of the parameters in the declaration. These little things just make the code easier to read.

src/dmd/optimize.d
```diff
@@ -42,24 +42,44 @@ extern (C++) Expression expandVar(int result, VarDeclaration v)
...
+
     static Expression errorReturn()
     {
         return new ErrorExp();
     }
 
-    Expression e = null;
     if (!v)
-        return e;
+        return nullReturn();
```
JinShil: It looks like you're just replacing `return null` with function that always returns `null` called "nullReturn".  I'm a little dumbfounded.  What am I missing?

WalterBright: It gives me an easy place to hook in things like a printf or other debug code. The inliner will inline it anyway, so it's cost free. And it lines up neatly with the Ddoc description of the 3 things that can be returned.

(When I debug code, I'll often add/remove printf's on the return statements. It's a bit of a hassle when there are multiple return points, and annoying when I miss one.)

JinShil (approved):
The coding style choices are odd, but I don't think I'm going to convince @WalterBright to change it, nor do I think it's a deal-breaker.  So, so be it.

WalterBright:
Thanks for indulging me with this. None of us will ever agree 100% on stylistic issues, because personal taste plays a role. But I do hope that once you get used a bit to my admittedly unusual style, you'll see the merit in it even if you don't care for it.

wilzbach:
> But I do hope that once you get used a bit to my admittedly unusual style, you'll see the merit in it even if you don't care for it.

Or it's just https://en.m.wikipedia.org/wiki/Stockholm_syndrome ... ;-)

PR #7682, opened 2018-01-12
Set length variable for slice and index expressions during semantic

ibuclaw:
Optimization opportunity noticed in #7677.

For static arrays, slice and indexes using values known at compile time should now never generate array bounds checks at codegen.

Before:
```
int[10] sa;
enum n = 2;                  // upperInRange
enum m = 5;                  // | lowerLessThan
                             // | |  check code
{ auto s = sa[0 .. 0];     } // 1 1  NULL
{ auto s = sa[0 .. $];     } // 0 1  ($ <= 10           )
{ auto s = sa[1 .. $];     } // 0 0  ($ <= 10 && 1 <= $ )
{ auto s = sa[$ .. $];     } // 0 0  ($ <= 10 && $ <= $ )
{ auto s = sa[0 .. $/n];   } // 0 1  ($/n <= 10         )
{ auto s = sa[$/n .. $];   } // 0 0  ($ <= 10 && 5 <= $ )
{ auto s = sa[$/m .. $/n]; } // 0 0  ($/n <= 10 && $/n <= $/m)
```
After
```
int[10] sa;
enum n = 2;                  // upperInRange
enum m = 5;                  // | lowerLessThan
                             // | |  check code
{ auto s = sa[0 .. 0];     } // 1 1  NULL
{ auto s = sa[0 .. $];     } // 1 1  NULL
{ auto s = sa[1 .. $];     } // 1 1  NULL
{ auto s = sa[$ .. $];     } // 1 1  NULL
{ auto s = sa[0 .. $/n];   } // 1 1  NULL
{ auto s = sa[$/n .. $];   } // 1 1  NULL
{ auto s = sa[$/m .. $/n]; } // 1 1  NULL
```

For reviewing the test changes (mostly just whitespace alignment with the rest of the function).
https://github.com/dlang/dmd/pull/7682/files?w=1

PR #7862, opened 2018-02-10
Recognize upper bounds values '0' and '$' in slices.

ibuclaw:
Both can never result in a bounds error, and so don't generate code to check for it.

Follow-up of #7682.

This means that code like:
```
auto sl = arr[n .. $])
```
Will only check that `n <= $`.

For smart optimizers, the conditions `0 <= $` and `$ <= $` would be removed as dead code anyway.  This just means that the glue/backend never generates the test in the first place.


JinShil (approved):

d-random-contributor:
Do we have xfail tests for
```d
a[1..0]
a[$..0]
```
to not overdo with shortcuts?

PR #7864, opened 2018-02-10
export  HOST_DMD=$DMD for druntime and phobos

timotheecour:
needed for https://github.com/dlang/druntime/pull/2084

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -46,8 +46,8 @@ clone() {
 build() {
     source ~/dlang/*/activate # activate host compiler
     make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=$DMD ENABLE_RELEASE=1 ENABLE_WARNINGS=1 all
-    make -j$N -C ../druntime -f posix.mak MODEL=$MODEL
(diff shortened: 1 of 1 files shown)


wilzbach:
As noted on Slack, there are many other scripts that would need to be updated too. Not only the five scripts I linked on Slack, but potentially also distro packing scripts and other tooling like digger.
Imho it makes sense to move in this direction and use D at DRuntime and Phobos as build script to cut the hugely redundant and ugly Makefile and a D compiler is already required to build DMD, so I don't think there's any general reason not to require it for building DRuntime, but we need to make sure that @andralex, @WalterBright, @MartinNowak and @CyberShadow are on board with this.

BTW to avoid any breakage a bootstrapping trick like the one already used in DMD's Makefile, DLang.org's Makefile can be applied. A simplified version of it can be found [here](https://github.com/wilzbach/d-bootstrap/blob/master/Makefile).

JinShil:
@timotheecour Is this still relevant given that the druntime PR has been closed?

JinShil:
The DRuntime PR was closed, so I'm assuming this is no longer needed.  Closing.

PR #7865, opened 2018-02-11
Print DFLAGS on -v to ease debugging

wilzbach:
Motivation: it can be tricky to figure out what dflags DMD has read. This helps debugging at almost no cost (and `-v` is only used for debugging anyhow).

See also: https://github.com/dlang/dmd/pull/7845

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -486,6 +486,31 @@ private int tryMain(size_t argc, const(char)** argv)
         message("binary    %s", global.params.argv0);
         message("version   %s", global._version);
         message("config    %s", global.inifilename ? global.inifilename : "(none)");
+        // Print DFLAGS environment variable
+        {
+            OutBuffer buf;
+            foreach (flag; dflags.asDArray)
+            {
+                bool needsQuoting;
+                for (auto flagp = flag; flagp; flagp++)
+                {
+                    auto c = flagp[0];
+                    if (!(isalnum(c) || c == '_'))
+                    {
+                        needsQuoting = true;
+                        break;
+                    }
+                }
+
+                if (flag.strchr(' '))
+                    buf.printf("'%s' ", flag);
+                else
+                    buf.printf("%s ", flag);
+            }
+
+            auto res = buf.peekSlice() ? buf.peekSlice()[0 .. $ - 1] : "(none)";
+            message("DFLAGS    %.*s", res.length, res.ptr);
+        }
     }
     //printf("%d source files\n",files.dim);
 
--- a/test/compilable/testclidflags.sh
+++ b/test/compilable/testclidflags.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+dir=${RESULTS_DIR}${SEP}compilable
+output_file=${dir}/testclidflags.sh.out
+
+unset DFLAGS
+
+# Force DMD to print the -v menu by passing an invalid object file
+# It will fail with "no object files to link", but print the log
+# On OSX DMD terminates with a successful exit code, so `|| true` is used.
+( "$DMD" -v foo.d 2> /dev/null || true) | grep -q "DFLAGS    (none)"
+( DFLAGS="-O -D" "$DMD" -v foo.d 2> /dev/null || true) | grep -q "DFLAGS    -O -D"
+( DFLAGS="-O '-Ifoo bar' -c" "$DMD" -v foo.d 2> /dev/null || true) | grep -q "DFLAGS    -O '-Ifoo bar' -c"
+
+echo Success >${output_file}


test/compilable/testclidflags.sh
```diff
@@ -0,0 +1,16 @@
...
+
+dir=${RESULTS_DIR}${SEP}compilable
+output_file=${dir}/testclidflags.sh.out
+
+unset DFLAGS
+
+# Force DMD to print the -v menu by passing an invalid object file
+# It will fail with "no object files to link", but print the log
+( ! "$DMD" -v foo.a 2> /dev/null ) | grep -q "dflags    (none)"
+( ! DFLAGS="-O -D" "$DMD" -v foo.a 2> /dev/null ) | grep -q "dflags    -O -D"
```
wilzbach: Imho it sucks that we have to workaround DMD so hard just to get it's config output. Shouldn't we make DMD's raw config settings easier to access? Most of this will be solved by the improved Json output, but there are still many cases where just looking what flags, args and precompiled versions DMD's as helps.
Ideas:
- default for `-v` (instead of showing the help page)
- `--version=detailed`

src/dmd/mars.d
```diff
@@ -486,6 +486,11 @@ private int tryMain(size_t argc, const(char)** argv)
         message("binary    %s", global.params.argv0);
         message("version   %s", global._version);
         message("config    %s", global.inifilename ? global.inifilename : "(none)");
+        OutBuffer buf;
```
JinShil: I recommend adding a header comment like "Print DFLAGS environment variable" for this block of code.

src/dmd/mars.d
```diff
@@ -486,6 +486,11 @@ private int tryMain(size_t argc, const(char)** argv)
...
+        foreach (flag; dflags.asDArray)
+            buf.printf("%s ", flag);
+        auto res = buf.peekSlice() ? buf.peekSlice()[0 .. $ - 1] : "(none)";
+        message("dflags    %.*s", res.length, res.ptr);
```
JinShil: Please use "DFLAGS" in all caps.

JinShil (changes requested):

src/dmd/mars.d
wilzbach: I used the existing convention of lower-casing everything - the entire output of `-v` starts with lower-cases tags.

JinShil: Environment variables are case-sensitive (at least on Linux).  Therefore, I think the case is important here.

wilzbach: What do other people think? (I don't mind changing it, but for now I focus on getting it to pass on auto-tester, so I won't change it immediately)

src/dmd/mars.d
```diff
@@ -486,6 +486,12 @@ private int tryMain(size_t argc, const(char)** argv)
         message("binary    %s", global.params.argv0);
         message("version   %s", global._version);
         message("config    %s", global.inifilename ? global.inifilename : "(none)");
+        // Print DFLAGS environment variable
+        OutBuffer buf;
```
marler8997: Should `buf` and `res` go in their own scope?

src/dmd/mars.d
marler8997: doesn't matter to me, both have small pros/cons.  Lower case is consistent with all other verbose output, CAPITAL make it consistent with the actual variable name.

src/dmd/mars.d
marler8997: Actually I didn't realize this was inside such a small `if` block...nevermind.

src/dmd/mars.d
```diff
@@ -486,6 +486,12 @@ private int tryMain(size_t argc, const(char)** argv)
...
+        foreach (flag; dflags.asDArray)
+            buf.printf("%s ", flag);
```
marler8997: I think individual dflags may contain spaces.  If so, we may want to surround them with quotes when we see that they contain whitespace.

wilzbach: So you want me to detect whether there's a space in the flag and then use quotes for it?
Most flags don't use spaces and it would look pretty ugly if this is done by default.

wilzbach:
Okay let's make a vote:
- :+1:  - `dflags`
- :-1:  - `DFLAGS`
- :heart: - I don't care just get this merged
- :confused: What are you talking about?

src/dmd/mars.d
timotheecour: DFLAGS=-I"path to foo" -Ibar
=>
`'-Ipath to foo' -Ibar`
so only `''` where needed;

also, using
`-Ipath\ to\ foo -Ibar`
is another option

wilzbach:
Yes of course, but we are currently blocked on Walter for the JSON output: https://github.com/dlang/dmd/pull/7838

src/dmd/mars.d
marler8997: Actually, a better solution might just be to print the original unprocessed value of `DFLAGS` because that one will already be quoted properly

wilzbach:
> Okay let's make a vote:

Looks like `DFLAGS` is more popular. Changed accordingly.

> shouldn't be by default, just if there's space:

Okay done.

> Actually, a better solution might just be to print the original unprocessed value of DFLAGS because that one will already be quoted properly

Not sure that's a good idea. There might potentially be some bugs or other problems with the processing of DFLAGS, so in my experience it's better to dump the actual value the program uses.

JinShil (approved):

test/compilable/testclidflags.sh
```diff
@@ -0,0 +1,17 @@
...
+output_file=${dir}/testclidflags.sh.out
+
+unset DFLAGS
+
+# Force DMD to print the -v menu by passing an invalid object file
+# It will fail with "no object files to link", but print the log
+# On OSX DMD terminates with a successful exit code, so `|| true` is used.
+( "$DMD" -v foo.d 2> /dev/null || true) | grep -q "DFLAGS    (none)"
+( DFLAGS="-O -D" "$DMD" -v foo.d 2> /dev/null || true) | grep -q "DFLAGS    -O -D"
+( DFLAGS="-O '-Ifoo bar' -c" "$DMD" -v foo.d 2> /dev/null || true) | grep -q "DFLAGS    -O '-Ifoo bar' -c"
```
rainers: I wonder why this works on the auto tester. For me, this test fails without any helpful output and if I run it explicitly it seems DFLAGS is picked up as a test parameter and d_do_test complains:
```
object.Exception@d_do_test.d(584): The DFLAGS test argument must be empty: It is '   ="-O -D" "$DMD" -v foo.d 2> /dev/null || true) | grep -q "DFLAGS    -O -D" ="-O '-Ifoo bar' -c" "$DMD" -v foo.d 2> /dev/null || true) | grep -q "DFLAGS    -O '-Ifoo bar' -c"'
```

JinShil: You running Windows?

rainers: Yes, but picking up DFLAGS should be OS platfrom-agnostic.

wilzbach: Maybe something to do with the binutils or bash? What's the actual output of `dmd` for you?

rainers: It's not dmd complaining, but d_do_test because `findTestParameter(envData, file, "DFLAGS", testArgs.dflags);` results in the shown string.

wilzbach: Sorry, that's weird. Bash parameters don't outlive their scope/script.
Does `unset DFLAGS` at the end help?

marler8997: It doesn't have to do with BASH.

`d_do_test` parses test parameters from the file and it looks like it is "accidently" picking up the new `DFLAGS` parameter from the BASH script itself...which it shouldn't be doing.

marler8997: Oh wait...`d_do_test` isn't involved in BASH tests at all...

@rainers are you sure it's this test that is failing?  If so, then the Makefile is probably using the wrong rule to run this test.

rainers: Ouch, totally missed that shell scripts are (no longer?) executed through d_do_test. 
There is zero output from the script itself, will have to investigate further...

marler8997: We just merged a chnage to BASH scripts which adds `sh_do_test.sh` that automatically redirects output from all BASH tests to the test output file, i.e. `test_reuslts/compilable/mytest.sh.out`

marler8997: https://github.com/dlang/dmd/pull/7928

rainers: Found the issue: dmd still used an sc.ini which overrides DFLAGS. Better to pass `-conf=` to dmd in these tests, too?

rainers: https://github.com/dlang/dmd/pull/7966

PR #7866, opened 2018-02-11
Fix rm: cannot remove ‘test_results/runnable/test13742.exe’: Device or resource busy

wilzbach:
Hopefully this prevents this spurious failure from showing up:

```
rm: cannot remove ‘test_results/runnable/test13742.exe’: Device or resource busy
```

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3028674&isPull=true

### DIFF
--- a/test/runnable/test13742.sh
+++ b/test/runnable/test13742.sh
@@ -19,6 +19,8 @@ $DMD -m${MODEL} -I${src} -cov -of${dir}${SEP}test13742${EXE} ${src}${SEP}test137
 
 ${RESULTS_DIR}/runnable/test13742${EXE} --DRT-covopt=dstpath:${dir}${SEP}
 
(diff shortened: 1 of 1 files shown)


marler8997:
Ugh, apparently this is a common problem in windows with no easy automatic solution.   I don't think `-f` is going to work.  However, ignoring the failure should be OK as long as we are OK with the files staying around.  If not, then in Windows we either need to kill the process that is using the file and then remove it, or rename it, possibly move it to a `junk` directory.

wilzbach:
> Ugh, apparently this is a common problem in windows with no easy automatic solution. I don't think -f is going to end up removing the file. However, ignoring the failure (like you've done here) should be OK

Yes, my idea was to be better safe than sorry and try both approaches.

JinShil (approved):

mathias-lang-sociomantic (approved):

PR #7867, opened 2018-02-11
fix Issue 18010 - Undefined reference to _d_arraycopy when copying ar…

WalterBright:
…rays in -betterC

This works by essentially inlining the code that is in _d_arraycopy(). It's faster, too. The downside is you just get a generic array bounds error when making a mistake rather than the detailed one _d_arraycopy() would emit.

### DIFF
--- a/src/dmd/e2ir.d
+++ b/src/dmd/e2ir.d
@@ -638,6 +638,11 @@ elem *addressElem(elem *e, Type t, bool alwaysCopy = false)
 
 /*****************************************
  * Convert array to a pointer to the data.
+ * Params:
+ *      t = array type
+ *      e = array to convert, it is "consumed" by the function
+ * Returns:
+ *      e rebuilt into a pointer to the data
  */
 
 elem *array_toPtr(Type t, elem *e)
@@ -659,8 +664,19 @@ elem *array_toPtr(Type t, elem *e)
             }
             else if (e.Eoper == OPpair)
             {
-                e.Eoper = OPcomma;
-                e.Ety = TYnptr;
+                if (el_sideeffect(e.EV.E1))
+                {
+                    e.Eoper = OPcomma;
+                    e.Ety = TYnptr;
+                }
+                else
+                {
+                    auto r = e;
+                    e = e.EV.E2;
+                    e.Ety = TYnptr;
+                    r.EV.E2 = null;
+                    el_free(r);
+                }
             }
             else
             {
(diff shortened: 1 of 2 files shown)


JinShil:
I have some reservations about maintaining two version of this logic: one inline in the compiler and one in the runtime.  I suggest either removing the dependency on `_d_arraycopy` altogether and keep the inline implementation, or modify the runtime hooks (maybe using templates) to make it compatible with -betterC usage.

WalterBright:
I generally prefer to wait a bit before removing a no-longer-needed druntime function, as people don't always update druntime and dmd simultaneously. There's also gdc/ldc.

wilzbach:
> I generally prefer to wait a bit before removing a no-longer-needed druntime function, as people don't always update druntime and dmd simultaneously. There's also gdc/ldc.

No need. Everyone does this. All CI scripts (and know of) depend on this. Also you very very often you get errors otherwise, because someone is already using a new compiler feature in DRuntime or Phobos (or simply depending on a fixed bug).
Also GDC and LDC update DRuntime + Phobos in lockstep.

WalterBright:
Can't remove _d_arraycopy with this pr anyway, so it's beside the point.

WalterBright:
@wilzbach please pull.

src/dmd/e2ir.d
```diff
@@ -669,8 +674,19 @@ elem *array_toPtr(Type t, elem *e)
             }
             else if (e.Eoper == OPpair)
             {
-                e.Eoper = OPcomma;
-                e.Ety = TYnptr;
+                if (el_sideeffect(e.EV.E1))
```
wilzbach: Any chance to add a test where this is actually true?

wilzbach (commented):
> @wilzbach please pull.

Ehm I feel uncomfortable with signing of changes I don't fully understand the effect and might have bigger consequences. I'm simply not involved long enough with D to review this properly. Sorry.

src/dmd/e2ir.d
```diff
@@ -2758,30 +2774,74 @@ elem *toElem(Expression e, IRState *irs)
...
                     {
                         elem *ex = el_same(&eto);
 
-                        // Determine if elen is a constant
-                        elem *elen;
-                        if (eto.Eoper == OPpair &&
-                            eto.EV.E1.Eoper == OPconst)
+                        /* Returns: length of array ex
+                         */
+                        static elem *getDotLength(elem *eto, elem *ex)
```
n8sh: Note: the logic of `getDotLength` is directly copied from the source, just moved to an inner function.

src/dmd/e2ir.d
```diff
@@ -2758,30 +2774,74 @@ elem *toElem(Expression e, IRState *irs)
...
+                         *   memcpy(ex.ptr, ey.ptr, nbytes)[0..elen]
+                         */
+                        e = el_params(nbytes, epfr, epto, null);
                         e = el_bin(OPcall,TYnptr,el_var(getRtlsym(RTLSYM_MEMCPY)),e);
                         e = el_pair(eto.Ety, el_copytree(elen), e);
-                        e = el_combine(eto, e);
+
+                        /* Combine: eto, efrom, echeck, e
+                         */
+                        e = el_combine(el_combine(eto, efrom), el_combine(echeck, e));
```
n8sh: Why `el_combine(el_combine(eto, efrom), el_combine(echeck, e))` instead of `el_combine(eto, el_combine(echeck, e))`?

src/dmd/e2ir.d
```diff
@@ -2758,30 +2774,74 @@ elem *toElem(Expression e, IRState *irs)
...
-                            elen = el_copytree(eto.EV.E1);
+                            auto ey = el_same(&efrom);
+                            auto eleny = getDotLength(efrom, ey);
+                            epfr = array_toPtr(ae.e2.type, ey);
+
+                            // length check: (eleny == elen)
+                            auto c = el_bin(OPeqeq, TYint, eleny, el_copytree(elen));
+
+                            /* Don't check overlap if epto and epfr point to different symbols
+                             */
```
n8sh: This appears to be vulnerable to pointer aliasing (epto and epfr could be pointing to arguments of an enclosing function, right?).

src/dmd/e2ir.d
WalterBright: Because that would omit `efrom`.

src/dmd/e2ir.d
WalterBright: Backend symbols themselves are always distinct, so no.

src/dmd/e2ir.d
```diff
@@ -659,8 +664,19 @@ elem *array_toPtr(Type t, elem *e)
...
+                    e.Eoper = OPcomma;
+                    e.Ety = TYnptr;
+                }
+                else
+                {
+                    auto r = e;
+                    e = e.EV.E2;
+                    e.Ety = TYnptr;
+                    r.EV.E2 = null;
+                    el_free(r);
```
n8sh: Note: it appears that this branch skips evaluating the length of the array and just returns the array's pointer directly.

src/dmd/e2ir.d
n8sh: I mean why does `efrom` need to be included apparently only in the bound-schecking case? In the non-bound-schecking case it gets set to null so `el_combine(eto, efrom)` is just `eto`.

src/dmd/e2ir.d
n8sh: When this condition is true this function uses the same behavior as prior to this PR.

n8sh (approved):
Most of the code here is not new, merely rearranged. The newly added bounds-checking code has explanatory comments that make it straightforward to see it is correct.

CyberShadow:
https://issues.dlang.org/show_bug.cgi?id=23911

RazvanN7:
Yes, this is a hack. The way to fix this issue (18010) is to template the _d_arraycopy hook. That way we get rid of the linker errors and also maintain the informative runtime assert message. cc @teodutu

PR #7868, opened 2018-02-11
fix Issue 18312 - string concatenation with -betterC fails with linke…

WalterBright:
…r errors

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -7745,7 +7745,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
 
     override void visit(CatExp exp)
     {
-        //printf("CatExp::semantic() %s\n", toChars());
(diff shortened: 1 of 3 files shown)


jacob-carlborg (approved):

src/dmd/expressionsem.d
```diff
@@ -7764,6 +7765,12 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             return;
         }
 
+        if (global.params.betterC)
```
wilzbach: Someone using `betterC`, might still choose to provide a custom `object.d` file in which he defines e.g. `RCString` and thus _can_ use array concatenation.
Thus, wouldn't it be better to use `global.params.useTypeInfo` ?

WalterBright: That would only work if array concatenation was implemented via a template in object.d. While this is a good idea, it is way beyond what this PR is.

wilzbach (approved):

PR #7870, opened 2018-02-12
Compute lazy lambda serialization

RazvanN7:
This is an optimization to compute the lambda serialization only when __traits(isSame) is used on lambda functions.

### DIFF
--- a/src/dmd/lambdacomp.d
+++ b/src/dmd/lambdacomp.d
@@ -23,6 +23,7 @@ import core.stdc.string;
 import dmd.declaration;
 import dmd.denum;
 import dmd.dsymbol;
+import dmd.dtemplate;
 import dmd.expression;
 import dmd.func;
 import dmd.dmangle;
@@ -33,6 +34,8 @@ import dmd.dscope;
 import dmd.statement;
 import dmd.tokens;
 import dmd.visitor;
+
+enum LOG = false;
 
 /**
  * The type of the visited expression.
@@ -83,6 +86,8 @@ public:
     override void visit(FuncLiteralDeclaration fld)
     {
         assert(fld.type.ty != Terror);
+        static if (LOG)
+            printf("FuncLiteralDeclaration: %s\n", fld.toChars());
 
         TypeFunction tf = cast(TypeFunction)fld.type;
         uint dim = cast(uint)Parameter.dim(tf.parameters);
@@ -111,9 +116,7 @@ public:
         }
 
         // Now the function body can be serialized.
-        CompoundStatement cs = fld.fbody.isCompoundStatement();
-        Statement s = !cs ? fld.fbody : null;
-        ReturnStatement rs = s ? s.isReturnStatement() : null;
+        ReturnStatement rs = fld.fbody.isReturnStatement();
         if (rs && rs.exp)
         {
             rs.exp.accept(this);
@@ -122,6 +125,8 @@ public:
 
     override void visit(DotIdExp exp)
     {
+        static if (LOG)
+            printf("DotIdExp: %s\n", exp.toChars());
         if (buf.offset == 0)
             return;
 
@@ -156,13 +161,9 @@ public:
         }
     }
 
-    override void visit(IdentifierExp exp)
+    bool checkArgument(const(char)* id)
     {
-        if (buf.offset == 0)
-            return;
-
         // The identifier may be an argument
-        auto id = exp.ident.toChars;
         auto stringtable_value = arg_hash.lookup(id, strlen(id));
         if (stringtable_value)
         {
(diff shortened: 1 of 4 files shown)


src/dmd/lambdacomp.d
```diff
@@ -262,9 +270,34 @@ public:
...
+        }
+        Dsymbol scopesym;
+        Dsymbol s = sc.search(t.sym.loc, t.sym.ident, &scopesym);
+        writeMangledName(s);
+    }
+
+    override void visit(TypeClass t)
     {
-        buf.reset();
+        if (LOG)
```
stefan-koch-sociomantic: static if

src/dmd/lambdacomp.d
```diff
@@ -83,6 +86,8 @@ public:
     override void visit(FuncLiteralDeclaration fld)
     {
         assert(fld.type.ty != Terror);
+        if (LOG)
```
stefan-koch-sociomantic: static if

andralex (approved):
cc @WalterBright

JinShil:
There aren't any tests and coverage is only 80%.  Is that deliberate?

RazvanN7:
@JinShil This PR does not introduce any new functionality, but optimizes the lambda comparison. The code is not covered because the struct and class cases are identical and I only added examples for one of them. I will add another few tests so that coverage is 100%

timotheecour:
@RazvanN7  speaking of __traits(isSame) and lambdas, there's a regression in git master:
https://issues.dlang.org/show_bug.cgi?id=18430 Issue 18430 - isSame is wrong for non global lambdas; obviously this PR couldn't have caused it since it's not merged but maybe you'd have some idea? (sorry for the drop in)

src/dmd/lambdacomp.d
```diff
@@ -34,6 +35,8 @@ import dmd.statement;
 import dmd.tokens;
 import dmd.visitor;
 
+enum LOG = false;
```
wilzbach: Out of interest: why not `version= LOG`? This is commonly used pattern in DRuntime.

src/dmd/traits.d
```diff
@@ -1453,8 +1453,24 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
 
         if (l1 && l2)
         {
-            import core.stdc.string;
+            const(char)* getSerialization(FuncLiteralDeclaration fld)
+            {
+                import dmd.lambdacomp;
```
wilzbach: Nit: local imports should be selective.

src/dmd/traits.d
```diff
@@ -1453,8 +1453,24 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
 
         if (l1 && l2)
         {
-            import core.stdc.string;
+            const(char)* getSerialization(FuncLiteralDeclaration fld)
```
wilzbach: Why don't you return `string` here? Comparing D strings is more efficient (no `\0` checking)
You already know the length of the buffer - no need to throw this valuable information away.

src/dmd/lambdacomp.d
RazvanN7: This is the pattern used in dmd from what I could obeserve :+1: 

https://github.com/dlang/dmd/blob/master/src/dmd/mtype.d#L61
https://github.com/dlang/dmd/blob/master/src/dmd/expression.d#L67
https://github.com/dlang/dmd/blob/master/src/dmd/expressionsem.d#L73

And the list could continue.

src/dmd/traits.d
RazvanN7: I know it's easier with strings, but the serialization field is is `const(char)*` and I really haven't seen AST members declared as strings. All of them are `const(char)*`

wilzbach: Quoting a conversation with Walter here:

>> On 12/5/2017 11:20 AM, Andrei Alexandrescu wrote:
Hi Walter, Seb noticed that the dmd code is going through unpleasant hoops caused by C-style strings. Do you think incrementally switching to D-style strings would improve the compiler? -- Andrei

> Yes, but we have to be mindful of gdc and ldc both relying on the C++ interface to use dmd, and D strings don't work on that interface.
> I've already converted a bunch of internal stuff to D strings where the interface to the back end was not involved.

(we can even convert the C++ interfaces, but that's significantly more effort)

RazvanN7: So how do we proceed here? That was my problem also: I could turn the field into a string, but what should I do regarding the header file?

RazvanN7:
I discovered some problems. Please do not merge.

src/dmd/traits.d
timotheecour: better yet: for all the apis used by ldc, gdc
use a compiler dependent (via eg version(LDC)) type Dstring, which for dmd is alias to char[] and for ldc, dmd is alias to extern(C++) struct with size and char* head, along with conversions functions.
that way, dmd code stays clean, safe and keeps length aroud

timotheecour: how about this:
```d
extern(C++) Dstring funLib(Dstring input);
```
see full working example here showing using such API from C++ with the function defined in D.

https://github.com/timotheecour/dtools/blob/master/tests/d01_dmd_string/

andralex: @timotheecour #7893  is a good step in the right direction, we need to start using D strings in the compiler. That would be an effort separated from this.

andralex (approved):

PR #7873, opened 2018-02-12
cc.d: harmonize with backend

WalterBright:


WalterBright:
The semaphoreci failure:
```
../../src/dmd/backend/cc.d(166,12): Error: module dtoken is in file 'dtoken.d' which cannot be read
```
is on the lines:
```
version (MARS)
    struct token_t;
else
    import dtoken : token_t;
```
Compiling these need to be done with `-version=MARS`. Looks like semaphoreci is broken.

wilzbach:
> Compiling these need to be done with -version=MARS. Looks like semaphoreci is broken.

All it does, is running the build scripts with `HOST_DMD=ldmd` (among other things).
And of course that can be reproduced locally:

image

### So what's the problem?

The DMD as a library package doesn't set `MARS`. It can easily be fixed with:

```diff
diff --git a/dub.sdl b/dub.sdl
index 641ab187d..bb6179b79 100644
--- a/dub.sdl
+++ b/dub.sdl
@@ -59,6 +59,7 @@ subPackage {
   versions "NoBackend"
   versions "GC"
   versions "NoMain"
+  versions "MARS"
   sourcePaths "src/dmd"
   excludedSourceFiles "src/dmd/backend/*"
   excludedSourceFiles "src/dmd/{e2ir,eh,glue,iasm,objc_glue,s2ir,tocsym,toctype,toobj,todt,toir}.d"
```

wilzbach:
FYI: In the interest of moving ahead with speed I took the liberty of applying the fix for you.

WalterBright:
Thanks, @wilzbach !

ibuclaw (approved):

PR #7874, opened 2018-02-12
Fix Issue 17970 - shared struct destructor doesn't compile anymore

RazvanN7:
The problem here is that if a struct is declared as shared the dtor is automatically considered to be shared, but when the struct is instantiated the instance is no longer considered to be shared when the function call matching is done. The fix makes it so that if a struct declaration is shared, when the destructor is called, the instantiated struct is also considered shared.

### DIFF
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -2422,6 +2422,24 @@ void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiar
             else
                 return 0;   // MATCH.nomatch
         }
+        /* Fix Issue 17970:
+           If a struct is declared as shared the dtor is automatically
+           considered to be shared, but when the struct is instantiated
+           the instance is no longer considered to be shared when the
+           function call matching is done. The fix makes it so that if a
+           struct declaration is shared, when the destructor is called,
+           the instantiated struct is also considered shared.
+        */
+        if (auto dt = fd.isDtorDeclaration())
+        {
+            auto dtmod = dt.type.toTypeFunction();
+            auto shared_dtor = dtmod.mod & MODFlags.shared_;
+            auto shared_this = tthis_fd.mod & MODFlags.shared_;
+            if (shared_dtor && !shared_this)
+                tthis_fd = dtmod;
+            else if (shared_this && !shared_dtor)
+                tf.mod = tthis_fd.mod;
+        }
         MATCH mfa = tf.callMatch(tthis_fd, fargs, 0, pMessage);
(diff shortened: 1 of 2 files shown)


wilzbach:
See also: https://github.com/dlang/dmd/pull/7566 (for disallowing shared destructors)

src/dmd/dtemplate.d
```diff
@@ -2422,6 +2422,17 @@ void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiar
             else
                 return 0;   // MATCH.nomatch
         }
+        auto dt = fd.isDtorDeclaration();
+        if (dt)
```
wilzbach: `if (auto dt = fd.isDtorDeclaration())`

src/dmd/dtemplate.d
```diff
@@ -2422,6 +2422,17 @@ void functionResolve(Match* m, Dsymbol dstart, Loc loc, Scope* sc, Objects* tiar
...
+        {
+            auto dtmod = dt.type.toTypeFunction();
+            auto shared_dtor = dtmod.mod & MODFlags.shared_;
+            auto shared_this = tthis_fd.mod & MODFlags.shared_;
+            if (shared_dtor && !shared_this)
+                tthis_fd = dtmod;
+            else if (shared_this && !shared_dtor)
+                tf.mod = tthis_fd.mod;
```
wilzbach: What's about the case of `shared_this && shared_dtor`?

wilzbach:
> The problem here is that if a struct is declared as shared the dtor is automatically considered to be shared, but when the struct is instantiated the instance is no longer considered to be shared when the function call matching is done. The fix makes it so that if a struct declaration is shared, when the destructor is called, the instantiated struct is also considered shared.

This explanation should be in the source code ;-)

src/dmd/dtemplate.d
RazvanN7: If either the struct/class declaration is shared or the destructor is shared, it makes sure that when the matching is done, both are considered shared.

wilzbach: (I asked because such motivation belongs into the source code - it will help people in the future and also the review of this PR)

andralex (approved):

andralex:
I'll approve this because it doesn't compete with #7566 and unblocks other students' work.

jacob-carlborg:
> This explanation should be in the source code ;-)

Putting it in the commit message is also good. When a PR only consists of one commit GitHub will automatically use the commit messages as the title and description for the PR.

PR #7878, opened 2018-02-12
Interpret imports the same regardless of other modules passed on the command line

marler8997:
This PR unifies the behavior of imports regardless of what modules were given on the command line (https://issues.dlang.org/show_bug.cgi?id=15086).  Consider the following:

```D
// main.d
import foo;
```
```D
// foo.d
module bar;
```

The above code will fail if compiled together, i.e.
```
dmd -c main.d foo.d
```
```
Error: module `bar` from file foo.d must be imported with 'import bar;'
```

However, if you compile them using 2 invocations, it will succeed:
```
dmd -c main.d
dmd -c foo.d
```

This PR unifies both cases by making them both fail with the same error message.

### DIFF
--- a/changelog/deprecate_compile_seperately_inconsistency.dd
+++ b/changelog/deprecate_compile_seperately_inconsistency.dd
@@ -1,0 +1,21 @@
+Make imports behave the same whether or not both modules were given on the command line
+
+Before, if an import statement matches a filename but does not match its module name, an error is asserted, but only if both modules have been given on the command line. Now it will be an error even if the modules are not both given on the command line, i.e.
+---
+// main.d
+import foo;
+---
+---
+// foo.d
+module bar;
+---
+Before, this module would fail if compiled like this:
+$(CONSOLE dmd -c main.d foo.d)
+
+and pass if compiled like this:
+$(CONSOLE
+dmd -c main.d
+dmd -c foo.d
+)
+
+Now it will fail in both cases.
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -514,7 +514,36 @@ extern (C++) final class Module : Package
             buf.printf("%s\t(%s)", ident.toChars(), m.srcfile.toChars());
             message("import    %s", buf.peekString());
         }
-        m = m.parse();
+        {
+            auto originalModule = m;
+            m = m.parse();
+
+            // verify the module name matches the imported module name
+            if (m is originalModule && m.md !is null)
+            {
+                bool mismatch = false;
+                if (!ident.equals(m.md.id))
+                {
+                    mismatch = true;
+                }
+                else
+                {
+                    auto mdPackageCount = (m.md.packages == null) ? 0 : m.md.packages.dim;
+                    auto packageCount   = (packages      == null) ? 0 : packages.dim;
+                    for (size_t i = 1; i <= mdPackageCount && i <= packageCount; i++)
+                    {
+                        if (!(*m.md.packages)[mdPackageCount - i].equals((*packages)[packageCount - i]))
+                        {
+                            mismatch = true;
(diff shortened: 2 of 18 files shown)


aG0aep6G:
> Note that this will not be an error if the import name matches the module name at least partially, i.e.
> 
> // main.d
> import foo.bar1
> import foo.bar2;
> 
> // foo/bar1.d
> module pkg.foo.bar1; // OK (foo.bar1 partially matches pkg.foo.bar1)
> 
> // foo/bar2.d
> module bar2; // OK (foo.bar2 partially matches bar2)

I don't see how this makes sense. The modules are `pkg.foo.bar1` and `bar2`, and they must be imported as such.

Fixing just the test case of 15086 is fine by me, though. As long as this doesn't allow more broken stuff than we currently have. I can just file another issue for the partial matching thing.

marler8997:
> I don't see how this makes sense. The modules are pkg.foo.bar1 and bar2, and they must be imported as such.

This is currently being debated between @WalterBright, @andralex and myself (see #7778).  They currently don't see why this is a problem.

> Fixing just the test case of 15086 is fine by me, though. As long as this doesn't allow more broken stuff than we currently have. I can just file another issue for the partial matching thing.

This PR does not allow anything that wasn't already allowed, it is only restricting the use case described.

marler8997:
I should mention that importing a module using it's filename, when it's module name does not match is already an error, it just doesn't get detected if you don't pass both files to the compiler on the same invocation.  i.e.
```D
// main.d
import foo;
```
```D
// foo.d
module bar;
```

If you run
```D
dmd -c main.d
```
It compiles fine, but if you run
```D
dmd -c main.d foo.d
```
it will fail.  This should not be.

andralex:
Had a long discussion with @WalterBright. Of the cases initially presented, this is the one with most merit. The following will focus on it, although certain considerations apply beyond.

Meta: the incumbent behavior has extra weight by means of being incumbent. We need to make a strong case to displace it.

History: initially module declarations were a lot stricter. That turned out early on to be too rigid to be workable. That doesn't dismiss revisiting it outright, but does stand as a warning.

The case for this PR rests on the notion there should be some invariance of build results with changing build conditions. E.g. "Either both cases should fail or both should pass" (and let me add: "... and create the same binary"). That actually doesn't seem to be a strong argument: any number of changes in the way binaries are built influence the result, and some are not workable. Furthermore, the added value of changing that should be argued (see "Meta"): what problem is this fixing? The error message plainly tells what must be done. We should enhance it.

A stronger case would be made by showing how unwitting code compiles in two distinct and plausible cases, yet behaves differently at runtime. "Unwitting" is an important detail because matters can be manipulated on the filesystem side - see below.

The second point made is that simple imports such as `import foo;` leave detail out. First off, a brief detour - in languages like C and C++, the directive `#include "foo.h"` is similarly vague. Both can mean many things by means of file and directory links, mounted drives, uses of `-I`, etc etc. 

What `import foo;` does mean is that once the import succeeds, the importing module has access to the artifacts imported under the name `foo`. Intuitively it makes a lot of sense that the importing module has some flexibility in choosing the imported module name, otherwise `import xyz = foo;` would be fraught with trouble.

Another good argument in favor of this PR could be constructed by showing that semantics of a successful build depend on the order in which files are specified. Order dependency is easy to characterize as undesirable.

marler8997:
I've presented what I think to be the best argument.  It appears it wasn't enough.  Closing.

marler8997:
In the future, I think we can be more efficient with our time if concerns about a bug not being a bug are voiced before a PR is created for it.  Especially when that bug has been around for 2 and a half years.

andralex:
@marler8997 You are certainly right, and I apologize. This has been off my radar.

PR #7886, opened 2018-02-13
Fix Issue 18427 - Improve deprecation message for importing privately imported symbol

RazvanN7:
The compiler cannot show how to import it because there is no way to import a symbol which is privately declared or imported in another module. The only way is to make the symbol/import public, but the compiler cannot advise that, so I pointed out where the variable is declared/imported.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -1094,7 +1094,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         if (imp.mod)
         {
             // Modules need a list of each imported module
-            //printf("%s imports %s\n", sc.module.toChars(), mod.toChars());
+            //printf("%s imports %s\n", sc._module.toChars(), imp.mod.toChars());
             sc._module.aimports.push(imp.mod);
 
             if (sc.explicitProtection)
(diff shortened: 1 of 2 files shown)


src/dmd/dsymbolsem.d
```diff
@@ -1145,13 +1145,20 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
             {
-                bool flag;
                 AliasDeclaration ad = imp.aliasdecls[i];
                 //printf("\tImport %s alias %s = %s, scope = %p\n", toPrettyChars(), aliases[i].toChars(), names[i].toChars(), ad._scope);
                 Dsymbol sym = imp.mod.search(imp.loc, imp.names[i] /*, IgnorePrivateImports */);
                 if (sym)
                 {
-                    flag = true;
+                    // Deprecated in 2018-01.
+                    // Change to error in 2019-01 by deleteting the following 3 lines and uncommenting
```
JinShil: This comment needs to be updated to "following 5 lines", or wordsmithed in a way to where the number of lines doesn't matter.

JinShil (approved):
LGTM other than the comment nit.

wilzbach (approved):
LGTM too (module @JinShil's nit)

src/dmd/dsymbolsem.d
RazvanN7: Oh, missed that, sorry. Thanks for the awesome attention to details. I addressed it now.

RazvanN7:
ping @wilzbach @JinShil I think this is good to go now. Merging this will make the PR queue < 140  :dagger:

JinShil:
It's relatively new PR, so I just want to give it at least 24 hours before I put my name next to the merge button.  If someone else wants to cowboy it in right now, I won't object, but I want to make sure there's sufficient time for others to have their say.  Also Jenkins is failing for some reason.  I'll restart it.

PR #7888, opened 2018-02-13
Fix Issue 18429 - alias this enum causes segmentation fault

RazvanN7:
If the identifier in the alias this declaration is defined later and is a voldemort type, we must perform semantic on the declaration to deduce the type.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -245,6 +245,15 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         Declaration d = sx.isDeclaration();
         if (d && !d.isTupleDeclaration())
         {
+            /* https://issues.dlang.org/show_bug.cgi?id=18429
+             *
+             * If the identifier in the AliasThis declaration
+             * is defined later and is a voldemort type, we must
+             * perform semantic on the declaration to deduce the type.
+             */
+            if (!d.type)
+                d.dsymbolSemantic(sc);
+
             Type t = d.type;
             assert(t);
             if (ad.type.implicitConvTo(t) > MATCH.nomatch)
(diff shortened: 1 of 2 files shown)


src/dmd/dsymbolsem.d
```diff
@@ -245,6 +245,11 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         Declaration d = sx.isDeclaration();
         if (d && !d.isTupleDeclaration())
         {
+            // Fix Issue 18429 : if the identifier in the alias this declaration
```
JinShil: Instead of writing "Fix Issue 18429", I suggest just pasting the URL.  It has the issue number in it, and gives the reader of the code immediate access the the issue.

JinShil (changes requested):

src/dmd/dsymbolsem.d
RazvanN7: Done.

stefan-koch-sociomantic:
This might be indicative of other problems, I'd be on the look out for other places where  the type is not set is not called.

WalterBright (approved):

WalterBright:
BTW @RazvanN7 thanks for the quick action on this bug!

JinShil:
The error in `automem` may be legitimate.  I restarted Jenkins to see if it's reproducible.

RazvanN7:
@WalterBright 

>It would be nice to merge the test case into another file in the compilable directory, since fewer files make the autotester go faster.

Done.

@JinShil 

Indeed, it was a regression, but I fixed it now.

JinShil (approved):

wilzbach:
>> UPDATE: Yep, look likes this may cause a regression. See the automem test in Jenkins.
> Indeed, it was a regression, but I fixed it now.

Wow. We added automem to the Projects Tester three days ago. I think @atilaneves will like this ;-)

atilaneves:
@wilzbach Can't complain ;)

PR #7817, opened 2018-01-31
End deprecation phase for `switch` cases which skip the declaration of a variable

wilzbach:
This never made it to the deprecation page.

- [x] https://github.com/dlang/dmd/pull/7842

PR #7891, opened 2018-02-14
Fix Issue 16578 - bogus deprecation - switch skips declaration of variable

JinShil:
The fix for this was modeled after the code here:  https://github.com/dlang/dmd/blob/56940531b005903a90ecaf0314e9198812bc69db/src/dmd/statement.d#L2264-L2267

I'm wondering if both the of these should be the same, that is

```D
else if (vd.storage_class & STC.temp || vd.storage_class & STC.exptemp)
{
    // Lifetime ends at end of expression, so no issue with skipping the statement
}
```

### DIFF
--- a/src/dmd/statement.d
+++ b/src/dmd/statement.d
@@ -1610,6 +1610,11 @@ extern (C++) final class SwitchStatement : Statement
                 // All good, the label's scope has no variables
                 return false;
             }
+            else if (vd.storage_class & STC.temp)
+            {
+                // Lifetime ends at end of expression, so no issue with skipping the statement
+                return false;
+            }
             else if (vd.ident == Id.withSym)
             {
                 deprecation("`switch` skips declaration of `with` temporary at %s", vd.loc.toChars());
--- a/test/compilable/test16578a.d
+++ b/test/compilable/test16578a.d
@@ -1,0 +1,16 @@
+// REQUIRED_ARGS: -debug
+
+// https://issues.dlang.org/show_bug.cgi?id=16578
+
+string[string] opts;
+
+void main()
+{
+    string arg;
+    switch (arg)
+    {
+        case "-f": opts["fore"] = ""; break;
+        debug { case "-throw": opts["throw"] = ""; break; }
+        default:
+    }
+}
--- a/test/compilable/test16578b.d
+++ b/test/compilable/test16578b.d
@@ -1,0 +1,16 @@
+// https://issues.dlang.org/show_bug.cgi?id=16578
+
+void main()
+{
+    string[string] opts;
+    switch (2)
+    {
+    case 0:
+        opts["a"] = "";
+        {
+    case 1:
+            break;
+        }
+    default:
+    }
+}


JinShil:
I think #7817 is waiting on this, so I'm marking this as blocking other work.

JinShil:
@dkgroot This is a very simple PR, but is failing on BSD.  Do you have local resources for debugging this?  Is this something you could look into?

src/dmd/statement.d
```diff
@@ -1603,6 +1603,10 @@ extern (C++) final class SwitchStatement : Statement
             {
```
wilzbach:
```
dmd/statement.d(1600): Error: function dmd.statement.SwitchStatement.checkLabel.checkVar no return exp; or assert(0); at end of function
posix.mak:450: recipe for target '../generated/freebsd/release/32/dmd_frontend' failed
```

I think this is a bug in the host compiler. I guess combining the `if`s to a joint expression could be worth a try:

```d
if (last == vd ||
    vd.storage_class & STC.temp)
``

**Comment by wilzbach** [comment]
> I'm wondering if both the of these should be the same, that is
> else if (vd.storage_class & STC.temp || vd.storage_class & STC.exptemp)

According to the docs `exptemp` is:

> temporary variable that has lifetime restricted to an expression

and in fact it's only used once when rewriting an opDollar expression:

https://github.com/dlang/dmd/blob/491a98693aab6b57c0c8a4099a9e881e84c348e1/src/dmd/expression.d#L1348

It's probably a relict from a hack against some bug and we don't need to check it here.

**Comment by wilzbach** [comment]
I was able to reproduce this with `digger build "master + dmd#7891"`.
The following works though:

```d
if (last == vd || vd.storage_class & STC.temp)
{
    // All good, the label's scope has no variables
    return false;
```

JinShil:
I don't get it.  That is already implemented as...

```D
if (last == vd)
{
    // All good, the label's scope has no variables
}
else if (vd.storage_class & STC.temp)
{
    // Lifetime ends at end of expression, so no issue with skipping the statement
}
...
return false;
```

Is that not the same as...
```D
if (last == vd || vd.storage_class & STC.temp)
{
    // All good, the label's scope has no variables
    return false;
}
```
... or do I need another cup of coffee?

JinShil:
Ok, so the problem I encountered with BSD appears to be a bug in the host compiler.  It may still exist in current master, but, despite and hour or so of trying, I have not succeeded in creating an isolated test to reproduce the bug so I can document it in Bugzilla.  Therefore, I went ahead and combined the two checks in to one `else if (condition1 || condition2)`, which avoids the bug, and allows this PR to move forward.

dkgroot:
@JinShil I only just noticed you request. Looks like you already resolved it (partially). If you like i can have a look. Does it only happen when compiling for 32bit ?

JinShil:
@dkgroot Thanks for getting back to me.  The problem occurs if we use this...

```
if (last == vd)
{
}
else if (vd.storage_class & STC.temp)
{
}
...
return false
```

... but doesn't occur if we write this:

```
if (last == vd || vd.storage_class & STC.temp)
{
    return false;
}
```

It only occurs in statement.d.  I tried to make an isolated test, but I couldn't reproduce it.

I can't remember whether or not it was only 32 or 64 bit.  I vaguely remember 32-bit being the failure, but it never got around to the 64-bit test because once a failure occurs the autotester seems to make it a lower priority over other PRs.

Any information you can provide would be most appreciated.

dkgroot:
@JinShil Ok will fire up a FreeBSD VM and have look

dkgroot:
> Thanks for getting back to me. The problem occurs if we use this...
if (last == vd)
{
}
else if (vd.storage_class & STC.temp)
{
}
...
return false
... but doesn't occur if we write this:
if (last == vd || vd.storage_class & STC.temp)
{
    return false;
}

Tested both on DragonFly64, FreeBSD64, FreeBSD32 and Linux64, and they work as expected. 
I do get however the same error message, when i leave out the 'return false' at the end of the function, which makes sense. Because your PR has been rebased, i cannot tell if that was the mistake you made when running into the issue. Maybe you can use 'git reflog' to checkout an older version of the PR and see if you may have forgotten 'return false' at the end.

dkgroot:
@JinShil 
> Thanks for getting back to me. The problem occurs if we use this...
if (last == vd)
{
}
else if (vd.storage_class & STC.temp)
{
}
...
return false
... but doesn't occur if we write this:
if (last == vd || vd.storage_class & STC.temp)
{
    return false;
}

Tested both on DragonFly64, FreeBSD64, FreeBSD32 and Linux64, and they work as expected. 
I do get however the same error message, when i leave out the 'return false' at the end of the function, which makes sense. Because your PR has been rebased, i cannot tell if that was the mistake you made when running into the issue. Maybe you can use 'git reflog' to checkout an older version of the PR and see if you may have forgotten 'return false' at the end.

wilzbach:
@dkgroot _please please_ don't waste your time. The issue has long being resolved and occurred just because of the old host compiler used. For this PR there are two workarounds
1) use return
2) put everything into the same if condition (currently done)

I would have preferred (1), because it allows to immediately see that the new code is covered, but that's not blocking this PR.

dkgroot:
@wilzbach Thanks for the heads up.

JinShil:
@dkgroot I'm terribly sorry for the wild goose chase.  I was under the impression from @wilzbach's comment...

> I was able to reproduce this with digger build "master + dmd#7891".

... that it was still an issue.

> I would have preferred (1), because it allows to immediately see that the new code is covered

@wilzbach I've updated the source code per your preference

dkgroot:
@JinShil No problem.
BTW: I used dmd-cxx as the bootstrap compiler on Linux / FreeBSD, which did not show this issue. 

@wilzbach mentioned this was an issue when using older bootstrap compilers (guessing something like v2.067.1). So there is a bit of good news, we could/should move to using dmd-cxx to be the bootstrap compiler on CI's instead of earlier versions. Positive side effect: It will/would (at least) make sure dmd-cxx does get a bit of attention every now and then when it breaks a bootstrap cycle.

src/dmd/statement.d
```diff
@@ -1610,6 +1610,11 @@ extern (C++) final class SwitchStatement : Statement
                 // All good, the label's scope has no variables
                 return false;
             }
+            else if (vd.storage_class & STC.temp)
+            {
+                // Lifetime ends at end of expression, so no issue with skipping the statement
+                return false;
```
wilzbach: That's why I wanted to have coverage info for this.
It looks like the test case is already fixed - even with the latest dmd + ldc release: https://run.dlang.io/is/ZoH9Fs

JinShil: Weird.  Something recent must have changed the logic path.  Shall we just close this and the bug as "Works for me"?

JinShil:
Updated the tests.  Coverage is now 100%.

wilzbach (approved):
Thanks!

jacob-carlborg (approved):

PR #7892, opened 2018-02-14
Fix Issue 17570 - Misleading error message illegal conditional function definition

RazvanN7:


JinShil:
There's something screwy about the 3 error messages displayed for 1 problem.  I think there's a way to prevent that by checking to see if an error has already been emitted.  I remember @WalterBright  telling me how to do that once.  I'll see if I can find it.

RazvanN7:
>There's something screwy about the 3 error messages displayed for 1 problem.

I completely agree, but you don't want the parser to stop parsing once it encounters one error; on the other hand you have to skip the whole if constraint which is not a predictable thing.

JinShil:
Is there any wisdom in [this comment](https://github.com/dlang/dmd/pull/7199#issuecomment-335272119) from @WalterBright  that can be applied to this PR?

> The theory of dealing with compilation errors is not to attempt to recover from them, but to 'poison' anything that depends on an AST node that has an error in it. This is the rationale of TypeError and similar in the compiler. Recovery attempts (i.e. guess what the user intended and attempt to repair the error) rarely work out well, and result in cascaded misleading error messages.

> This PR is trying to recover from the error by putting the erroneous AnonDeclaration back into a known state, which later passes depend on.

> A fix more in keeping with the error design in the compiler is to 'poison' the AnonDeclaration node by setting AnonDeclaration.errors. Then, when the node is used, errors is checked and the error is propagated instead. Making this work in the most effective manner is to put the errors check inside the various include() member functions, having them return null if errors is set.

src/dmd/parse.d
```diff
@@ -4770,6 +4770,8 @@ final class Parser(AST) : Lexer
                     error("'%s' cannot be placed after a template constraint", token.toChars);
                 else if (t == TOK.at)
                     error("attributes cannot be placed after a template constraint");
+                else if (t == TOK.if_)
+                    error("cannot use conditional definition for non-template functions. Use `static if` instead");
```
JinShil: I don't think "conditional definition" is the right terminology here.  Shouldn't it be "conditional expression"?

RazvanN7: I think that "function constraint" would be best.

RazvanN7:
@JinShil 

>Is there any wisdom in this comment from @WalterBright that can be applied to this PR?

There is a lot of wisdom in that comment, but I fail to see how it can be applied to this PR.

JinShil (approved):

PR #7894, opened 2018-02-15
Add more JSON

marler8997:
Added fields to `compilerInfo` object in JSON object.

Note that this is based on https://github.com/dlang/dmd/pull/7521 

It differs a bit, electing to go for a json representation matching the internals rather than trying to exactly match DUB probe files.  Not sure which approach is better for DMD, this one is easier to implement so I went with this one for now.

Example Output:
```json
{
 "compilerInfo" : {
  "vendor" : "Digital Mars D",
  "version" : "v2.078.2-742-gd6b8376-dirty",
  "__VERSION__" : 2078,
  "compilerInterface" : "dmd",
  "bitWidth" : 64,
  "platforms" : [
   "posix"
  ],
  "architectures" : [
   "x86_64"
  ],
  "predefinedVersions" : [],
  "supportsIncludeImports" : true
 },
 "buildInfo" : {
  "cwd" : "/home/marlerj/ddev1/dmd",
  "argv0" : "./generated/linux/release/64/dmd",
  "config" : "./generated/linux/release/64/dmd.conf",
  "importPaths" : [
   "./generated/linux/release/64/../../../../../druntime/import",
   "./generated/linux/release/64/../../../../../phobos"
  ],
  "objectFiles" : [],
  "libraryFiles" : [],
  "ddocFiles" : [],
  "mapFile" : null,
  "resourceFile" : null,
  "defFile" : null
 }
}
```

### DIFF
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -362,6 +362,44 @@ struct Global
         _version = (import("VERSION") ~ '\0').ptr;
         compiler.vendor = "Digital Mars D";
     }
+
+    /**
+    Returns: the version as the number that would be returned for __VERSION__
+    */
+    extern(C++) uint versionNumber()
+    {
+        import core.stdc.ctype;
+        __gshared static uint cached = 0;
+        if (cached == 0)
+        {
+            //
+            // parse _version
+            //
+            uint major = 0;
+            uint minor = 0;
+            bool point = false;
+            for (const(char)* p = _version + 1;; p++)
+            {
+                const c = *p;
+                if (isdigit(cast(char)c))
+                {
+                    minor = minor * 10 + c - '0';
+                }
+                else if (c == '.')
+                {
+                    if (point)
+                        break; // ignore everything after second '.'
+                    point = true;
+                    major = minor;
+                    minor = 0;
+                }
+                else
+                    break;
+            }
+            cached = major * 1000 + minor;
+        }
+        return cached;
+    }
 }
 
 // Because int64_t and friends may be any integral type of the
--- a/src/dmd/json.d
+++ b/src/dmd/json.d
@@ -228,12 +228,36 @@ public:
         buf.writestring(" : ");
     }
 
+    /**
+    Write the given string object property only if `s` is not null.
+
+    Params:
+     name = the name of the object property
+     s = the string value of the object property
+    */
     void property(const(char)* name, const(char)* s)
     {
         if (s is null)
             return;
         propertyStart(name);
         value(s);
+        comma();
+    }
+
+    /**
+    Write the given string object property.
+
+    Params:
+     name = the name of the object property
+     s = the string value of the object property
+    */
+    void requiredProperty(const(char)* name, const(char)* s)
+    {
+        propertyStart(name);
+        if (s is null)
+            buf.writestring("null");
+        else
+            value(s);
         comma();
     }
 
@@ -826,8 +850,61 @@ public:
     private void generateCompilerInfo()
     {
         objectStart();
-        property("binary", global.params.argv0);
-        property("version", global._version);
+        requiredProperty("vendor", global.compiler.vendor);
+        requiredProperty("version", global._version);
+        property("__VERSION__", global.versionNumber());
+        requiredProperty("interface", determineCompilerInterface());
+        property("size_t", size_t.sizeof);
+        propertyStart("platforms");
+        arrayStart();
+        if (global.params.isWindows)
+        {
+            item("windows");
+        }
+        else
+        {
+            item("posix");
+            if (global.params.isLinux)
+                item("linux");
+            else if (global.params.isOSX)
+                item("osx");
+            else if (global.params.isFreeBSD)
+            {
+                item("freebsd");
+                item("bsd");
+            }
+            else if (global.params.isOpenBSD)
+            {
+                item("openbsd");
+                item("bsd");
+            }
+            else if (global.params.isSolaris)
+            {
+                item("solaris");
+                item("bsd");
+            }
+        }
+        arrayEnd();
+
+        propertyStart("architectures");
+        arrayStart();
+        if (global.params.is64bit)
+            item("x86_64");
+        else
+            version(X86) item("x86");
+        arrayEnd();
+
+        propertyStart("predefinedVersions");
+        arrayStart();
+        if (global.versionids)
+        {
+            foreach (const versionid; *global.versionids)
+            {
+                item(versionid.toChars());
+            }
+        }
+        arrayEnd();
+
         propertyBool("supportsIncludeImports", true);
         objectEnd();
     }
@@ -839,18 +916,50 @@ public:
     private void generateBuildInfo()
     {
         objectStart();
-        property("cwd", getcwd(null, 0));
-        property("config", global.inifilename ? global.inifilename : null);
-        if (global.params.lib) {
-            property("library", global.params.libname);
-        }
+        requiredProperty("cwd", getcwd(null, 0));
+        requiredProperty("argv0", global.params.argv0);
+        requiredProperty("config", global.inifilename);
+        requiredProperty("libName", global.params.libname);
+
         propertyStart("importPaths");
         arrayStart();
-        foreach (importPath; *global.params.imppath)
+        if (global.params.imppath)
         {
-            item(importPath);
+            foreach (importPath; *global.params.imppath)
+            {
+                item(importPath);
+            }
         }
         arrayEnd();
+
+        propertyStart("objectFiles");
+        arrayStart();
+        foreach (objfile; global.params.objfiles)
+        {
+            item(objfile);
+        }
+        arrayEnd();
+
+        propertyStart("libraryFiles");
+        arrayStart();
+        foreach (lib; global.params.libfiles)
+        {
+            item(lib);
+        }
+        arrayEnd();
+
+        propertyStart("ddocFiles");
+        arrayStart();
+        foreach (ddocFile; global.params.ddocfiles)
+        {
+            item(ddocFile);
+        }
+        arrayEnd();
+
+        requiredProperty("mapFile", global.params.mapfile);
+        requiredProperty("resourceFile", global.params.resfile);
+        requiredProperty("defFile", global.params.deffile);
+
         objectEnd();
     }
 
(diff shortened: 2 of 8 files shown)


wilzbach (commented):
Wow that was quick! Thanks!
~~~`config` and `compilerFrontend` are globals, so they could be easily added from this PR.~~~

`compiler` and `frontendVersion` are a bit more effort, so it probably makes sense to do them in a quick followup (I can do this too).

Otherwise LGTM!

wilzbach:
Ah I see:
- `config` is already part of `buildInfo`
-  `version` is part of `compilerInfo`

src/dmd/json.d
```diff
@@ -826,8 +826,60 @@ public:
     private void generateCompilerInfo()
     {
         objectStart();
-        property("binary", global.params.argv0);
+        property("vendor", global.compiler.vendor);
         property("version", global._version);
+        property("argv0", global.params.argv0);
```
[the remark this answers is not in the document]
marler8997: Well, I was thinking that `argv0` is more true to what it actually is.  "binary" sounds more like a filename but argv0 can be an alias/string/filename.  It may be helpful to include both "binary" and "argv0"...

src/dmd/json.d
```diff
@@ -826,8 +826,60 @@ public:
     private void generateCompilerInfo()
     {
         objectStart();
-        property("binary", global.params.argv0);
+        property("vendor", global.compiler.vendor);
         property("version", global._version);
```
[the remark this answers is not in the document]
marler8997: inifilename is already in the `buildInfo` field

If we add DFLAGS...wouldn't we want to add all the command line arguments?

src/dmd/json.d
```diff
@@ -826,8 +826,60 @@ public:
...
+        property("argv0", global.params.argv0);
+
```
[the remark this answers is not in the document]
marler8997: Can you elaborate what `frontendVersion` and `compilerFrontend` are for?

src/dmd/json.d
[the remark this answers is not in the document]
wilzbach: > If we add DFLAGS...wouldn't we want to add all the command line arguments?

Yes, but again doesn't have to be in this PR.

src/dmd/json.d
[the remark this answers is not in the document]
wilzbach: (these variables are currently used by DUB in its platform probing)

- `frontendVersion`: 2071, 2072, known in D as `__VERSION__` (and in DMD as `VERSIONX` token). This is important for feature and compatibility detection
- `compiler`, e.g. `dmd`, `ldc` or `gdc` - as mentioned the vendor might be different to the CLI interface (maybe `compilerInterface` is a better name)
- `compilerFrontend` - same as `version` (already existent)

timotheecour:
definitely not PR blockers (could be done in future ones) but for reference:
* json output shd probably be somewhere in docs (maybe in https://dlang.org/dmd-osx.html) ; real useful to have
* linker flags
* example with architectures.length > 1 
* predefinedVersions shouldn’t be empty (bug?)
* stringImportPaths (on that note IIRC there was a bug where these would become absolute paths)

marler8997:
Ok, I think this is a good start.  I've included "frontendVersion" as "__VERSION__" and also included "compilerInterface".  This should definitely have everything `rdmd` needs.  I didn't add `DFLAGS`...I think I'll let @wilzbach figure out the best way to add that.  Since this is experimental, if there's something wrong we can always fix it later.

Note for future changes.  I opted for the compilerInfo/buildInfo objects to ALWAYS include fields even if their values are null/empty.  This way a user can still see the format of the document even if a value is null (see `resourceFile`/`defFile`/etc as examples).  However, to have these semantics with string fields, you need to use the `propertyOrNull` function instead of `property`.  The `property` function will just leave the field out if the value is null.

marler8997:
NOTE:  If you copy the following script to the test directory, you can run it after building DMD to automatically update the json tests with the new expected JSON output:

https://gist.github.com/marler8997/f3973cc6bcde7a88d04f48140ea8fe6b
```BASH
## Script to update json test files
set -x

make test_results/compilable/json2.d.out
cp test_results/compilable/json2.out.sanitized compilable/extra-files/json2.out

make test_results/compilable/jsonCompilerInfo.d.out
cp test_results/compilable/jsonCompilerInfo.out.sanitized compilable/extra-files/jsonCompilerInfo.out

make test_results/compilable/json_nosource.sh.out
cp test_results/compilable/json_nosource.out.sanitized compilable/extra-files/json_nosource.out
```

src/dmd/json.d
```diff
@@ -826,8 +836,61 @@ public:
     private void generateCompilerInfo()
     {
         objectStart();
-        property("binary", global.params.argv0);
-        property("version", global._version);
+        propertyOrNull("vendor", global.compiler.vendor);
+        propertyOrNull("version", global._version);
+        property("__VERSION__", global.versionNumber());
+        propertyOrNull("interface", determineCompilerInterface());
+        property("bitWidth", size_t.sizeof * 8);
```
wilzbach: Not sure if `bitWidth` is a good name.

src/dmd/json.d
```diff
@@ -867,9 +962,8 @@ public:
         foreach (m; Module.amodules)
         {
             objectStart();
-            if(m.md)
-                property("name", m.md.toChars());
-            property("file", m.srcfile.toChars());
+            propertyOrNull("name", m.md ? m.md.toChars() : null);
+            propertyOrNull("file", m.srcfile.toChars());
```
wilzbach: Can this even be `null`?

src/dmd/json.d
```diff
@@ -968,3 +1062,20 @@ JsonFieldFlags tryParseJsonField(const(char)* fieldName)
...
     return JsonFieldFlags.none;
 }
+
+/**
+Determines and returns the compiler interface which is one of `dmd`, `ldc`,
+`gdc` or `sdc`. Returns `null` if no interface can be determined.
+*/
+private const(char)* determineCompilerInterface()
+{
+    if (0 == strcmp(global.compiler.vendor, "Digital Mars D"))
```
wilzbach: Pattern is `actual == expected`  (grep through the source code).
Of course `!strcmp` would work too.

src/dmd/globals.d
```diff
@@ -477,3 +515,4 @@ enum PINLINE : int
 alias StorageClass = uinteger_t;
 
 extern (C++) __gshared Global global;
+
```
wilzbach: Unrelated?

src/dmd/json.d
```diff
@@ -839,18 +902,50 @@ public:
...
         objectStart();
-        property("cwd", getcwd(null, 0));
-        property("config", global.inifilename ? global.inifilename : null);
-        if (global.params.lib) {
-            property("library", global.params.libname);
-        }
+        propertyOrNull("cwd", getcwd(null, 0));
+        propertyOrNull("argv0", global.params.argv0);
+        propertyOrNull("config", global.inifilename);
+        propertyOrNull("libname", global.params.libname);
```
wilzbach: Everything else uses camelCase, might be a good idea to use `libName` too?
Precedent is above: we don't use `inifilename`, but `config`

src/dmd/json.d
marler8997: yeah, any other ideas?  It's the same number that's printed in the --version output.

src/dmd/json.d
marler8997: no, but it seemed like consistently using `propertyOrNull` was better.  The new objects should use `propertyOrNull` and the old `modules` object uses property....maybe there is a better name....

src/dmd/globals.d
```diff
@@ -362,6 +362,44 @@ struct Global
...
         compiler.vendor = "Digital Mars D";
     }
+
+    /**
+    Returns: the version as the number that would be returned for __VERSION__
+    */
+    extern(C++) uint versionNumber()
+    {
+        import core.stdc.ctype;
+        static uint cached = 0;
```
WalterBright: Don't use thread local storage for dmd. Use `__gshared`

src/dmd/json.d
```diff
@@ -237,6 +237,16 @@ public:
         comma();
     }
 
+    void propertyOrNull(const(char)* name, const(char)* s)
```
WalterBright: No Ddoc header

src/dmd/json.d
WalterBright: `size_tSizeof`

src/dmd/json.d
```diff
@@ -826,8 +836,61 @@ public:
...
+        if (global.params.isWindows)
+        {
+            item("windows");
+        }
+        else
+        {
+            item("posix");
+            if (global.params.isLinux)
+                item("linux");
+            if (global.params.isOSX)
```
WalterBright: `else if`, etc.

src/dmd/globals.d
marler8997: good catch

src/dmd/json.d
marler8997: ok, `size_t` seems reasonable.  With `size_t` I also changed it to the "byte-width" instead of "bit-width".   Let me know if someone prefers it to be the bit-width instead.

src/dmd/json.d
marler8997: The name `requiredProperty` came to mind and seems like a good idea.

src/dmd/globals.d
```diff
@@ -362,6 +362,44 @@ struct Global
         _version = (import("VERSION") ~ '\0').ptr;
         compiler.vendor = "Digital Mars D";
     }
+
+    /**
+    Returns: the version as the number that would be returned for __VERSION__
+    */
+    extern(C++) uint versionNumber()
```
WalterBright: This code duplicates code in the lexer https://github.com/dlang/dmd/blob/master/src/dmd/lexer.d#L518

In fact, it's a cut&paste of it. It should take an argument instead of relying on a global, and moved to another module (lexer.d?).

marler8997: As you can see later on in this PR, the code was removed from lexer.d and replaced with a call to this function.  I could move this function to lexer.d and have globals import lexer (instead of the other way around).  Would you prefer that?

marler8997: Yes it could take an argument, but then the value won't be cached.  This current implementation allows the value to be cached in case `__VERSION__` is used multiple times along with outputing the `__VERSION__` json value.

WalterBright: I see you are correct. I missed that. I with draw both suggestions.

WalterBright:
Is this no longer a WIP?

marler8997:
Correct no longer WIP.  Though this does not "finalize" all the JSON fields...this should be ready for merge so others can build upon it.

PR #7896, opened 2018-02-15
fix DMDSRC link

timotheecour:
cc @wilzbach a more robust fix could be to use DMDROOT with:

in dlang.org/dlang.org.ddoc:
s/
DMDSRC=$(HTTPS github.com/dlang/dmd/blob/master/src/dmd/$0, $0)
/
DMDROOT=$(HTTPS github.com/dlang/dmd/blob/master/$0, $0)
DMDSRC=$(DMDROOT src/dmd/$0) #not sure what syntax here ?
/

### DIFF
--- a/changelog/default_after_variadic.dd
+++ b/changelog/default_after_variadic.dd
@@ -21,4 +21,4 @@ for every invocation:
 string log(string file = __FILE__, int line = __LINE__, T...)(T a);
 ---
 
-For more details, see $(DMDSRC test/runnable/testdefault_after_variadic.d).
+For more details, see $(DMDSRC ../../test/runnable/testdefault_after_variadic.d).
--- a/changelog/default_after_variadic.dd
+++ b/changelog/default_after_variadic.dd
@@ -20,5 +20,3 @@ for every invocation:
 ---
 string log(string file = __FILE__, int line = __LINE__, T...)(T a);
 ---
-
-For more details, see $(DMDSRC ../../test/runnable/testdefault_after_variadic.d).


wilzbach:
> cc @wilzbach a more robust fix could be to use DMDROOT with:

Or you could simply remove the link to the test. Changelog entries typically don't link to the DMD testsuite.

> could be to use DMDROOT with:

For reference these already exist:

```
PHOBOSSRC=$(SPANC phobos_src, $(AHTTPS github.com/dlang/phobos/blob/master/$0, $0))
DRUNTIMESRC=$(HTTPS github.com/dlang/druntime/blob/master/src/$0, $0)
```

(these are needed for the automatically generated source lines from Ddoc)

I'm not sure about `DMDROOT` as I do see that its' useful to have, but on the other hand the only use of it is here where it's probably not a good idea to link to the testsuite anyhow.

timotheecour:
> Or you could simply remove the link to the test. Changelog entries typically don't link to the DMD testsuite.

well you add it in eaf851253403e0a6ee72990b19492af42540d6be ...  ;-)

actually ideal would be:

* fixing this:
https://issues.dlang.org/show_bug.cgi?id=18442 Issue 18442 - documented unittests don't have anchors (so we can't link to them, eg in changelogs) (needs to be fixed regardless of this)

* and making sure theres a link like https://dlang.org/phobos/std_complex.html generated for testsuite, eg https://dlang.org/phobos/some_link_to_dmd_testsuite.html
(which would show only pages w at least some ddoc comment); that way we automatically benefit from all the goodies like runnable unittests

feel free to remove the link altogether (you can edit PR iirc)

wilzbach (approved):

wilzbach:
> I've removed the link altogether in the meantime so we can merge this to avoid broken link

:+1: 

> well you add it in eaf8512 ... ;-)

Yeah sorry about this. `PHOBOSSRC` works doesn't use this prefixing :/

> fixing this:
https://issues.dlang.org/show_bug.cgi?id=18442 Issue 18442 - documented unittests don't have anchors (so we can't link to them, eg in changelogs) (needs to be fixed regardless of this)

Replied on the issue.

> And making sure theres a link like https://dlang.org/phobos/std_complex.html generated for testsuite, eg https://dlang.org/phobos/some_link_to_dmd_testsuite.html
(which would show only pages w at least some ddoc comment); that way we automatically benefit from all the goodies like runnable unittests; then it can be linked in changelog (and future changelog entries can benefit from same stuff)

We already have a linter for HTML pages:

https://github.com/dlang/dlang.org/blob/master/chmgen.d#L361

It just finds too many broken links that it's not active at the moment.
Though we could probably modify it to use a blacklist and then slowly weed out that blacklist...

PR #7900, opened 2018-02-15
Fix module/import order invariance

marler8997:
Verify that all imports in the same compilation are order invariant.

1. Compiler sees import
2. Check if import is in module symbol table
3. __CHANGE:__ Before this change, if the module is found just return it.  After this change, we verify that the import name is valid (i.e. the import name has been used to load this module before).  If it's not validated then we throw away the match and continue this algorithm by loading whatever module matches this import name (this is the key to fixing import order invariance).
3. Convert import name to filename and look for it on the file system
4. __CHANGE:__ Using filename, check if a module has already been loaded with that filename.  If so, return it.  This prevents the same module file from being loaded twice.
5. Otherwise, continue loading the module as normal.

NOTE: the following tests were previously NOT COMPILABLE but now are:
```
test/fail_compilation/ice14446.d
test/fail_compilation/diag4479.d
```


test/fail_compilation/ice14446.d
```diff
@@ -1,14 +0,0 @@
-// REQUIRED_ARGS: -o-
```
marler8997: NOTE: this test is now passing...moved to `test/compilable/ice14446.d`.  It now passes because the compiler now fully supports importing via filename so long as it doesn't break import order invariance.

test/fail_compilation/diag4479.d
```diff
@@ -1,11 +0,0 @@
-// REQUIRED_ARGS: fail_compilation/imports/fail4479.d
```
marler8997: NOTE: this test is now passing...moved to `test/compilable/diag4479.d`. It now passes because the compiler now fully supports importing via filename so long as it doesn't break import order invariance.

marler8997:
I'm not sure how much I'll be able to work on this in the next few days.  However, I plan on adding tests (maybe with some help from @timotheecour) but the current implementation is ready for review.

marler8997:
@WalterBright this PR is the result of many hours of work, trial and error trying to find the right solution to solve this invariance problem.   This was not an easy problem to solve. This PR is definitely going to need your eyes to approve.  Please take a look when you have a chance.

andralex:
cc @WalterBright

JinShil:
Assuming this is no longer a WIP.  Removing the label.  (We need a way for contributors to do that).

WalterBright:
General comment: Please use Ddoc comments for new functions. This means having a `Params:` section if there are parameters, and a `Returns:` section for non-void return values.

src/dmd/dimport.d
```diff
@@ -135,10 +135,39 @@ extern (C++) final class Import : Dsymbol
...
+        //printf("import %s\n", importName.toChars());
         if (s)
         {
             if (s.isModule())
-                mod = cast(Module)s;
+            {
+                auto match = cast(Module)s;
+                //printf("  - matched %s\n", match.toChars());
+
+                // We need to make sure that this match would occur the same way whether or not the module
```
WalterBright: s/We need to make sure/Ensure/

src/dmd/dimport.d
```diff
@@ -135,10 +135,39 @@ extern (C++) final class Import : Dsymbol
...
+                // was previously loaded by another import.  To ensure this, if the current import name matches the
```
WalterBright: no double space after .

src/dmd/module.h
```diff
@@ -53,6 +53,13 @@ class Package : public ScopeDsymbol
     Module *isPackageMod();
 };
 
+struct QualifiedID
```
WalterBright: No Ddoc comment

src/dmd/module.h
```diff
@@ -53,6 +53,13 @@ class Package : public ScopeDsymbol
...
+{
+    Identifiers *packages;
+    Identifier *ident;
+    const char *toChars();
```
WalterBright: `const(char)* toChars() const;`

src/dmd/module.h
```diff
@@ -75,6 +82,7 @@ class Module : public Package
 
 
     const char *arg;    // original argument name
+    Array<QualifiedID> importNames; // a list of names that can be used to import this module
```
WalterBright: s/a list of names/names/

as it's obviously an array of names, not a list of names

src/dmd/root/array.d
```diff
@@ -313,3 +313,22 @@ unittest
     array[2] = 910;
     assert([123, 421, 910, 123, 1, 2, 8, 20, 4, 3] == array.asDArray);
 }
+
+/**
+ * Returns true if every element of $(D left) and $(D right) are equal.
```
WalterBright: It's `Returns:` note the colon, and it goes after the `Params:` section

src/dmd/root/filename.d
```diff
@@ -64,11 +64,28 @@ nothrow:
         return compare(str, (cast(FileName*)obj).str);
     }
 
+    private static char normalizeWinPathChar(char c) pure
```
WalterBright: No Ddoc comment

src/dmd/root/filename.d
```diff
@@ -64,11 +64,28 @@ nothrow:
...
+    {
+        if (c >= 'a' && c <= 'z') return cast(char)(c - ('a' - 'A'));
```
WalterBright: Uppercase file name characters are annoying, better to normalize to lower case. Also, this should be a nested function within `compare()`.

marler8997:
Are you wanting ddoc comments for new private functions as well?

src/dmd/root/filename.d
```diff
@@ -64,11 +64,28 @@ nothrow:
...
     {
         version (Windows)
         {
-            return stricmp(name1, name2);
+            for (size_t i = 0;; i++)
+            {
+                if (name1[i] == '\0')
+                    return name2[i] == '\0' ? 0 : -1;
+
+                auto left = normalizeWinPathChar(name1[i]);
```
WalterBright: `const`, not `auto`

src/dmd/module.h
marler8997: I think you meant `const char *toChars() const` ?

src/dmd/root/array.d
```diff
@@ -313,3 +313,22 @@ unittest
...
+ * Params:
+ *  left = pointer to array to compare
+ *  right = pointer to array to compare
+ */
+bool opEquals(T)(const(Array!T)* left, const(Array!T)* right)
+{
+    if (left is null || left.dim == 0)
+        return right is null || right.dim == 0;
+    if (right is null || left.dim != right.dim)
+        return false;
```
WalterBright: This code is hard on the eyes. How about:
```
const leftdim = left is null ? 0 : left.dim;
const rightdim = right is null ? 0 : right.dim;
if (leftdim != rightdim)
    return false;
foreach (i; 0 .. leftdim)
    ...
return true;

File: src/dmd/dmodule.d [hunk 13]
```diff
@@ -280,6 +281,48 @@ extern (C++) class Package : ScopeDsymbol
     }
 }
 
+extern (C++) struct QualifiedID
```
WalterBright: No Ddoc comment

src/dmd/dmodule.d
```diff
@@ -280,6 +281,48 @@ extern (C++) class Package : ScopeDsymbol
...
+{
+    Identifiers* packages;
+    Identifier ident;
+    extern(D) bool opEquals(const QualifiedID right) const
+    {
+        return this.ident == right.ident &&
+            packages.opEquals(right.packages);
+    }
+    extern (C++) const(char)* toChars()
```
WalterBright: toChars() should be const

src/dmd/dmodule.d
```diff
@@ -280,6 +281,48 @@ extern (C++) class Package : ScopeDsymbol
...
+        if (packages && packages.dim)
+        {
+            for (size_t i = 0; i < packages.dim; i++)
+            {
+                Identifier pid = (*packages)[i];
+                buf.writestring(pid.toChars());
+                buf.writeByte('.');
+            }
+        }
+        buf.writestring(ident.toChars());
```
WalterBright: use `ident.toString()`

src/dmd/dmodule.d
```diff
@@ -280,6 +281,48 @@ extern (C++) class Package : ScopeDsymbol
...
+        if (!ident || ident.toChars() == null)
+            return "<no-name>".ptr;
+
+        OutBuffer buf;
+        if (packages && packages.dim)
+        {
+            for (size_t i = 0; i < packages.dim; i++)
+            {
+                Identifier pid = (*packages)[i];
+                buf.writestring(pid.toChars());
```
WalterBright: use `pid.toString()`

src/dmd/dmodule.d
```diff
@@ -280,6 +281,48 @@ extern (C++) class Package : ScopeDsymbol
...
+    }
+    extern (C++) const(char)* toChars()
+    {
+        if (!ident || ident.toChars() == null)
+            return "<no-name>".ptr;
+
+        OutBuffer buf;
+        if (packages && packages.dim)
+        {
+            for (size_t i = 0; i < packages.dim; i++)
```
WalterBright: use foreach

src/dmd/dmodule.d
```diff
@@ -280,6 +281,48 @@ extern (C++) class Package : ScopeDsymbol
...
+                buf.writeByte('.');
+            }
+        }
+        buf.writestring(ident.toChars());
+        return buf.extractString();
+    }
+}
+
+// Take the module and create a new class that points to the
+// same module but with a different id.
```
WalterBright: Ddoc comment

src/dmd/dmodule.d
```diff
@@ -306,6 +349,7 @@ extern (C++) final class Module : Package
     extern (C++) static __gshared AggregateDeclaration moduleinfo;
 
     const(char)* arg;           // original argument name
+    Array!QualifiedID importNames; // a list of names that can be used to import this module
```
WalterBright: It's an array, not a list
s/a list of names/names/

src/dmd/dmodule.d
```diff
@@ -427,91 +471,126 @@ extern (C++) final class Module : Package
         return new Module(filename, ident, doDocComment, doHdrGen);
     }
 
-    static Module load(Loc loc, Identifiers* packages, Identifier ident)
+    static const(char)* importToFilename(QualifiedID importName)
```
WalterBright: Ddoc comment

src/dmd/dmodule.d
```diff
@@ -427,91 +471,126 @@ extern (C++) final class Module : Package
...
+                        --qlen;             // remove trailing separator
+                    buf.writestring(q[1 .. qlen + 1]);
+                    break Lmain;            // last matching entry in ms[] wins
                 }
-                dotmods.writeByte('.');
+             }
+             dotmods.writeByte('.');
+         }
+
+        for (size_t i = 0; i < importName.packages.dim; i++)
```
WalterBright: foreach

src/dmd/dmodule.d
```diff
@@ -983,20 +1098,32 @@ extern (C++) final class Module : Package
...
                 assert(global.errors);
         }
-        else
+        return null;
+    }
+
+    /**
+    Adds the new importName to the list of valid importNames, and then
+    adds the module to the symbol table in the new import location.
+    */
```
WalterBright: Sigh. Ddoc comments.

src/dmd/root/filename.d
marler8997: The character being returned isn't used to build a new filename string, it's just used as a temporary value for comparision.

From what I understand, comparison functions usually convert to upper case because it will typically result in less if statements.  Because the lower-case letters are at the end of the ASCII table, most characters will be less than `'a'` which means they will only perform one comparison.  If we go to lower-case instead, then you'll get have
```D
if (c >= 'A' && c <= 'Z') ...
```
then you'll almost always perform 2 comparisons whether or not c is upper case or lower case, resulting in more total comparisons.

WalterBright:
> Are you wanting ddoc comments for new private functions as well?

Yes, unless they're nested.

src/dmd/module.h
marler8997: Do you want ddoc comments for the C++ headers?

WalterBright: No, sorry, I didn't realize it was in a .h file.

marler8997:
NOTE: I'm planning on adding more tests but I was waiting to see if https://github.com/dlang/dmd/pull/7905 will be merged first.  These tests will contain a bunch of small files and it would be alot easier to write/maintain them with that PR.

marler8997:
Ping

PR #7902, opened 2018-02-16
[WIP] Refactor mtype.d

RazvanN7:
This PR is aimed at refactoring mtype.d. It includes:

- Making functions private, where possible, or moving them where they are used
- Creating visitors for semantic time methods so that encapsulation is increased
- Separating semantic time methods from AST nodes so that AST nodes contain the minimum necessary information


WalterBright:
> Making function private, where possible, or moving them where they are used

Review is easier if independent changes like this are done in a separate PR.

RazvanN7:
I thought that making a single PR where the independent changes are commit-based would imply less "pollution" for the PR queue. After all, the changes are not so independent; it's a refactorization of mtype.d

RazvanN7:
I broke down this PR into smaller PR's : #7932 , #7934 .

PR #7904, opened 2018-02-16
Add an AA range (debug mode only)

marler8997:
Made this input range for dmd associative arrays for debug purposes.  Decided to make a PR for it if other people think it's useful to include.

### DIFF
--- a/src/dmd/dsymbol.d
+++ b/src/dmd/dsymbol.d
@@ -2114,6 +2114,22 @@ extern (C++) final class DsymbolTable : RootObject
         return s;
     }
 
+    debug
+    {
+        /**
+        print the symbol table contents
+        */
+        override void print()
+        {
+            printf("SYMBOL TABLE (%d entries)\n------------------------------\n", dmd_aaLen(tab));
+            foreach (keyValue; tab.asRange)
+            {
+                auto ident = cast(Identifier)keyValue.key;
+                printf("%s\n", ident.toChars());
+            }
+        }
+    }
+
     // Look for Dsymbol in table. If there, return it. If not, insert s and return that.
     Dsymbol update(Dsymbol s)
     {
--- a/src/dmd/root/aav.d
+++ b/src/dmd/root/aav.d
@@ -24,11 +24,17 @@ extern (C++) size_t hash(size_t a)
 alias Key = void*;
 alias Value = void*;
 
+struct KeyValue
+{
+    Key key;
+    Value value;
+}
+
 struct aaA
 {
     aaA* next;
-    Key key;
-    Value value;
+    KeyValue keyValue;
+    alias keyValue this;
 }
 
 struct AA
@@ -119,6 +125,92 @@ extern (C++) Value dmd_aaGetRvalue(AA* aa, Key key)
         }
     }
     return null; // not found
+}
+
+debug
+{
+    /**
+    Gets a range of key/values for `aa`.
+
+    Returns: a range of key/values for `aa`.
+    */
+    @property auto asRange(AA* aa)
+    {
+        return AARange(aa);
+    }
+
+    private struct AARange
+    {
+        AA* aa;
+        // current index into bucket array `aa.b`
+        size_t bIndex;
+        aaA* current;
+
+        this(AA* aa)
+        {
+            if (aa)
+            {
+                this.aa = aa;
+                toNext();
+            }
+        }
+
+        @property bool empty() const { return current is null; }
+
+        @property auto front() { return current.keyValue; }
+
+        void popFront()
+        {
+            if (current.next)
+                current = current.next;
+            else
+            {
+                bIndex++;
+                toNext();
+            }
+        }
+
+        private void toNext()
+        {
+            for (; bIndex < aa.b_length; bIndex++)
+            {
+                if (auto next = aa.b[bIndex])
+                {
+                    current = next;
+                    return;
+                }
+            }
+            current = null;
+        }
+    }
+    unittest
+    {
+        AA* aa = null;
+        foreach(keyValue; aa.asRange)
+            assert(0);
+
+        enum totalKeyLength = 50;
+        foreach (i; 1 .. totalKeyLength + 1)
+        {
+            auto key = cast(void*)i;
+            {
+                auto valuePtr = dmd_aaGet(&aa, key);
+                assert(valuePtr);
+                *valuePtr = key;
+            }
+            bool[totalKeyLength] found;
+            size_t rangeCount = 0;
+            foreach (keyValue; aa.asRange)
+            {
+                assert(keyValue.key <= key);
+                assert(keyValue.key == keyValue.value);
+                rangeCount++;
+                assert(!found[cast(size_t)keyValue.key - 1]);
+                found[cast(size_t)keyValue.key - 1] = true;
+            }
+            assert(rangeCount == i);
+        }
+    }
 }
 
 /********************************************


src/dmd/root/aav.d
```diff
@@ -121,6 +121,61 @@ extern (C++) Value dmd_aaGetRvalue(AA* aa, Key key)
...
+        Value value;
+    }
+    /**
+    Returns an input range of key/values from `aa`.
+    Returns: an input range of key/values from `aa`.
+    */
+    auto aaRange(AA* aa)
+    {
+        return AARange(aa);
+    }
```
jacob-carlborg: Please add an empty newline between functions.

JinShil:
What's the problem that this solves?  What's the use case and how does it help?  If this were not merged, would DMD be in any kind of deficit?

marler8997:
My use case was debugging the contents of symbol tables.

src/dmd/dsymbol.d
```diff
@@ -2114,6 +2114,22 @@ extern (C++) final class DsymbolTable : RootObject
         return s;
     }
 
+    debug
+    {
+        /**
+        dump the symbol table contents
+        */
+        void dump()
```
JinShil: This could probably be renamed to something like "printSymbolTable".

marler8997: seems redundant...
```
symbolTable.printSymbolTable
```

What's the argument for the redundancy?

JinShil: Sorry. I didn't realize it was a member.  Still "dump" is quite informal (slang in a way); perhaps "print"

marler8997: ok

JinShil:
Is there any way to test this in the test suite to keep it from getting accidentally breaking, and provide a way to verify the output without having to test it locally?

marler8997:
The logic that would need to be tested is AARange. Would a unnittest suffice?

marler8997:
> I'd also like to verify the output. Here's and idea, but I don't know if it will work.

I'm all for testing...but, I do think you can "overtest" as well.  The `print` function itself is just a `foreach` loop with a print statement.  All it's doing is printing the contents of the associative array.

That being said, the underlying logic in `AARange` would definitely warrant a unittest to prevent regression.

This is of course my opinion, maybe I'm wrong and this is worth extra testing.  If it helps the likelihood that it will be accepted than I'm willing to accommodate, however, I have a feeling that it actually might hurt the odds rather than help them.

src/dmd/root/aav.d
```diff
@@ -121,6 +127,63 @@ extern (C++) Value dmd_aaGetRvalue(AA* aa, Key key)
...
+    */
+    @property auto asRange(AA* aa)
+    {
+        return AARange(aa);
+    }
+
+    private struct AARange
+    {
+        AA* aa;
+        size_t b_index;
```
wilzbach: DStyle: `dIndex` (and a comment wouldn't hurt too)

src/dmd/root/aav.d
```diff
@@ -121,6 +127,63 @@ extern (C++) Value dmd_aaGetRvalue(AA* aa, Key key)
     return null; // not found
 }
 
+debug
+{
+    /**
+    Returns a range of key/values for `aa`.
+    Returns: a range of key/values for `aa`.
```
wilzbach: duplicate?

wilzbach:
> Decided to make a PR for it if other people think it's useful to include.

Yes, it would be quite useful. I have missed this many times already. Thanks a lot for submitting this.

> The logic that would need to be tested is AARange. Would a unnittest suffice?

I think so too. It's not like it would be a big deal if an opt-in debug function stops to work and I don't think we have a precedent of testing `debug` code in DMD except for using D's built-in unittest blocks.
Also it looks like rather big modifications to the Makefile are required to be able to test this in the testsuite which I think is an overkill.

src/dmd/root/aav.d
```diff
@@ -121,6 +127,63 @@ extern (C++) Value dmd_aaGetRvalue(AA* aa, Key key)
...
+                toNext();
+            }
+        }
+
+        private void toNext()
+        {
+            for (; b_index < aa.b_length; b_index++)
+            {
+                auto next = aa.b[b_index];
+                if (next)
```
wilzbach: One line: `if (auto next = ...)`

src/dmd/root/aav.d
marler8997: Actually it was intentional, I was trying to make a valid ddoc comment.  A description and a return value...maybe one of them is not necessary?

src/dmd/root/aav.d
marler8997: actually that logic doesn't work...that assignment can't be done until after the loop condition is satisfied.

src/dmd/dsymbol.d
marler8997: uh...it looks like the `print` function is already taken...switched it back to `dump` unless you have another name you'd prefer.

JinShil: What does `print()` currently do?  Maybe it should be overridden.

marler8997: ok, fine with me :)

jacob-carlborg:
I don't mind some extra debug functionality in DMD. I think it's very useful, especially for someone not that familiar with the code base.

src/dmd/root/aav.d
```diff
@@ -121,6 +127,93 @@ extern (C++) Value dmd_aaGetRvalue(AA* aa, Key key)
...
+                toNext();
+            }
+        }
+
+        private void toNext()
+        {
+            for (; bIndex < aa.b_length; bIndex++)
+            {
+                auto next = aa.b[bIndex];
+                if (next)
```
wilzbach: > One line: if (auto next = ...)
> actually that logic doesn't work...that assignment can't be done until after the loop condition is satisfied.

I don't follow you here - is this a remark on previous code?

marler8997: Woops, I thought you were talking about putting `auto next = ...` in the for loop...will change

wilzbach (approved):

wilzbach:
BTW I'm wondering whether it would make sense to start abstracting over the relicts from the C/C++ age (like it's done for arrays) and use D-syntax for associative arrays, e.g. sth. like this:

```d
struct DAa(K, V)
{
    AA* aa;
    size_t length()
    {
        return dmd_aaLen(aa);
    }

    V* opIndex(K key)
    {
        return cast(V*) dmd_aaGet(&aa, cast(void*) key);
    }

    auto opIndexAssign(V value, K key)
    {
        auto ptr = dmd_aaGet(&aa, cast(void*) key);
        *ptr = cast(void*) value;
        return *ptr;
    }
}
```

marler8997:
Yeah I think that's a great idea.

src/dmd/root/aav.d
```diff
@@ -121,6 +127,92 @@ extern (C++) Value dmd_aaGetRvalue(AA* aa, Key key)
     return null; // not found
 }
 
+debug
+{
+    /**
+    Returns a range of key/values for `aa`.
+
+    Returns: a range of key/values for `aa`.
```
JinShil: Nit:  Could use "Gets a range..." for the summary to avoid the redundancy.

JinShil (approved):

wilzbach (approved):

timotheecour:
@marler8997 

Thanks for dong this PR, something I also wished existed; however, one thing i don't like is:
`override void print()` 
how about instead having:
`char* toString()` or `extern(D) string toString()` or another name like `toStringDebug`.

forcing print to stdout is not flexible, eg:
* in my case I want to log through a function that prepends  __FILE__:__LINE__ before calling toString; * another case is recursive dumping using indentation that depends on nesting level
* another case is dumping to a file insetead of stdout

also, it's much more standard to output to a string/char* than to dump to stdout

PR #7905, opened 2018-02-16
Isolate tests to one file using HAR

marler8997:
The HAR format allows a test with multiple files to be contained within a single file.  This makes it easy to see all the files involved in a single test, making it easier to understand the test and modify it. This also allows tests to use simple filenames without having to worry about stomping on files from other tests.

I've noticed that many of the tests are made up of multiple small files...having the ability to manage them in one file seems easier than dealing with a bunch of little files.

Also note that using the HAR format would allow one to copy/paste a tests with multiple files to/from online posts. A nice convenience.

HAR: https://github.com/marler8997/har
forum discussion about HAR: https://forum.dlang.org/thread/ndgdqraxjkuvfsjhewgy@forum.dlang.org


MartinNowak:
We do use `----` in ddoc to delimit code examples, any chance of conflict?
Somewhat of a nice idea, but it's also a new format that needs to be learned, and using files from imports and extra-files has hardly been a problem so far.
I do like the perspective of self-contained test-cases that might be copy+pasted from Bugzilla, and maybe publishing d_do_test to a small dub app to run them.
So if you can teach&convince lots of people to use this format (maybe a forum post), that would definitely pan out in the long-term.

marler8997:
> We do use ---- in ddoc to delimit code examples, any chance of conflict?

The delimiter requires 3 dashes, a space (ascii 0x20) and then a filename character.  This doesn't conflict with ddoc as far as I can tell.  Also, if a test did have a conflict, HAR allows for a file to use it's own delimiter (https://github.com/marler8997/har#custom-delimiters).

> Somewhat of a nice idea, but it's also a new format that needs to be learned, and using files from imports and extra-files has hardly been a problem so far.

Not a big problem but a small annoyance for me.  Managing tests with multiple files is a pain sometimes.  A pain to know all the files involved in the test, a pain to make tests with multiple files.  I'm going to need to add some tests pretty soon with up to 5 small files (2 to 5 lines each) and I'm not looking forward to that.  I'll probably need to add 5 new tests, each with their own set of files...that's like 25 files...bleh (this is to test the new import order invariance stuff...requires test cases with multiple files).

> So if you can teach&convince lots of people to use this format (maybe a forum post), that would definitely pan out in the long-term.

Yeah there's a forum link in the description.  @timotheecour and I just came up with the format a couple days ago so it's too soon to know what people think.

marler8997:
Not bad ideas. But I do want to say that this is definitely not a preprocessor, the compiler knows nothing about the HAR format, it only exists as a tool in the test suite.

Changing the language to support multiple modules in a file is a much larger endeavor...something that would be great to investigate, but I don't want to be conflated woth this effort. Even if that feature was implemented, it would add a new mechanism to pass modules to the compiler, but we would still want to test cases with multiple files still making the HAR format useful. These are essentially orthogonal features, even though both mechanisms can be used to put multiple modules in a single file.

marler8997:
BTW, if the language were changed to support multiple modules per file, your approach makes much more sense that using HAR as a preprocessor. Your approach limits the feature to only allow nested modules, which provides more flexibility without losing the intuitive organization which makes it easier to find source code. It also wouldn't allow weird stuff like making files whose module names are different from their filenames. I would probably go with something like this:

JinShil:
What about modifying HAR to use the syntax I proposed earlier?

marler8997:
HAR is a general syntax for representing multiple files in a single block of text. Not specific to any language or format. The syntax you presented is for representing D modules in a single file.

JinShil:
Ok, then, perhaps we could make DAR (D programming language Archive Format) and use that instead of HAR.

marler8997:
Sorry I'm failing to see the reason...are you wanting to add a new file format that the compiler can read multiple modules from? or are you talking about a file format for the test suite to represent multiple files.  Keep in mind, this isn't just for D files, you could put JSON files/html files/BASH files, etc

JinShil:
What I'm proposing is to replace the  `--- modulename.d` syntax with the syntax `module modulename {   }`, but otherwise keep everything else basically the same.  Let's call this syntax "DAR" for the sake of this discussion.

I realize that implementing the DAR syntax in the language would be a major undertaking requiring a DIP and a number of other hurdles.  So, to keep with the spirit of this PR, while also setting a precedent for a potential language modification, change the implementation, currently in `har.d` to support the DAR syntax instead.

That is, instead of the `---- modulename.d` syntax, modify `har.d` (could be renamed to `dar.d`)  to support the syntax `module modulename {   }`.  It should produce the same output as any equivalent HAR file.

The parser would probably be more difficult to implement, but it would set a precedent that could motivate a future DIP for incorporating the syntax into the language with potential future benefits such as having accurate line numbers in error messages, just to name one.  

The syntax can still be used for the same use cases as HAR (at least in the context of D) For example:  improving our test implementation as demonstrated in this PR, sharing code in forum posts, illustrating concepts in DLang Tour, and sharing code and ideas through http://run.dlang.io, just to name a few.

Yes, DAR would only be useful for D source code, unfortunately.  Though HAR could still be used as an archive DAR files together with the other file formats HAR would support.

marler8997:
I'm not necessarily against these ideas, but they seem orthogonal to me.  These 2 formats can coexist and serve different purposes. I don't want this idea to be conflated with a much larger endeaver.

JinShil:
The two ideas (HAR and DAR) are orthogonal in that they can coexist with one another, but, in the context of D and in the spirit of this PR as demonstrated `importorder.har`, they overlap in use case and purpose (e.g. single-file tests, sharing code in forum posts, illustrating concepts in DLang Tour, and sharing code and ideas through http://run.dlang.io, to name a few)

marler8997:
They would be testing different things. All har does is archive files. Your vision for DAR would be to have the compiler read it directly, but thats different than testing multiple files. These are different use cases.

marler8997:
I'm going to have to respectfully disagree on this one. I think your syntax idea could be a good addition to the D language, but I dont see any benefit in its use here.

wilzbach:
@JinShil have a look at the [HAR parser](https://github.com/marler8997/har). It's super-simple at the moment and still general purpose.  Your idea of DAR requires at least a D lexer.

Anyhow, I like the idea and imho it would be nice to have it on dlang.org, DFeed and Bugzilla, s.t. all these sites have an easy way to send multiple files to run.dlang.io while still being readable for humans.

(and of course on the test suite here too)

wilzbach:
> Anyhow, I like the idea and imho it would be nice to have it on dlang.org, DFeed and Bugzilla, s.t. all these sites have an easy way to send multiple files to run.dlang.io while still being readable for humans.

Turns out it's really simple to do it for run.dlang.io (and thus the DTour, DSpec, and Phobos documentation:

https://run.dlang.io/is/uearnw

-> I vote in favor of giving this a try. How else are people going to know about it and use it?

Also the implementation of `har` is really simple and it wouldn't be much work to extract all har tests to their files in the source code if we ever decide to drop it in the future.

JinShil:
I don't object at all to implementing this feature, and am even willing to pull it as is, but I recognize that the only reason this is required is because of a flaw in the design of the language.  

The fact that the source code must mirror the way it is physically stored on the file system is an arbitrary limitation.  HAR would not be required if the syntax I proposed earlier were incorporated into the language.  Even the shell script tests would be able to generate their requisite files using the shell language itself.

I am well aware that doing the work to incorporate said syntax into the language would be a major undertaking, and the HAR solution is vastly simpler by comparison, but that is the cost of due diligence.  

If I'm the only one that sees this, then so be it; it is on me to do something about it, and I'll support this HAR format as it does indeed provide a simple and immediate solution to a number of inconveniences caused by the aforementioned design flaw.

marler8997:
HAR is useful even if the language supported multiple modules per file. It allows you to create tests with mutiple files.  This is useful even if your syntax was integrated.

JinShil:
I noticed this PR creates a new directory for the HAR code:  `test/archive/har.d`  Can we just put `har.d` in the root of `test`, instead?  I don't see the need for a separate directory to house one file.

wilzbach:
We could create `tools` that then can also house `sanitize_json.d`, `combinations.d`, `d_do_test.d`, `do_test.sh` and potentially other, future tools.
Would that work for everyone?

marler8997:
Fine with me, but I put har in archive since its in the "archive" package. Could go in tools/archive

JinShil:
> Would that work for everyone?

OK with me.

test/archive/har.d
```diff
@@ -0,0 +1,176 @@
+module archive.har;
```
JinShil: I think it would be helpful to add a documentation header to this file summarizing the HAR format. Perhaps some cut and paste from the readme at https://github.com/marler8997/har.  A link to that repository would also be helpful to know where it came from to add to the maintainer's historical and contextual understanding.

marler8997: ok

JinShil (approved):
If the arbitrary limitation of having modules and packages tied to directories and files were removed, this PR would have much less value.  But, since that is not likely to happen any time in the near future, I'll settle for something like this, as I'd rather have it than not have it.

test/Makefile
```diff
@@ -199,13 +199,13 @@ DEBUG_FLAGS=$(PIC_FLAG) -g
 
 export DMD_TEST_COVERAGE=
 
-runnable_tests=$(wildcard runnable/*.d) $(wildcard runnable/*.sh)
+runnable_tests=$(wildcard runnable/*.d) $(wildcard runnable/*.har) $(wildcard runnable/*.sh)
```
wilzbach: You can do `$(wildcard runnable/*.har runnable/*.sh)`

test/archive/har.d
```diff
@@ -0,0 +1,198 @@
...
+                scope(exit) currentOutputFile.close();
+                for (;;)
+                {
+                    lineRange.popFront();
+                    if (lineRange.empty)
+                        break LfileLoop;
+                    line = lineRange.front;
+                    if (line.startsWith(delimiter))
+                        break;
+                    currentOutputFile.write(line);
```
wilzbach: You need to write the delimiter too, see https://github.com/marler8997/har/pull/4

test/archive/har.d
```diff
@@ -0,0 +1,198 @@
...
+                }
+            }
+        }
+        return Yes.success;
+    }
+}
+
+@property auto formatDir(const(char)[] dir)
```
wilzbach: The `@property` seems unnecessary here.

test/archive/har.d
```diff
@@ -0,0 +1,198 @@
...
+  in { assert(file.length > 0); } do
+{
+    return formatQuotedIfSpaces(file);
+}
+
+/**
+Returns a formatter that will print the given string.  it will print
+it surrounded with quotes if the string contains any spaces.
+*/
+@property auto formatQuotedIfSpaces(T...)(T args) if (T.length > 0)
```
wilzbach: inconsistent constant style. the if constraint is typically on a separate new line, see https://dlang.org/dstyle.html#phobos_declarations

test/archive/har.d
```diff
@@ -0,0 +1,198 @@
...
+            {
+                if (arg.indexOf(' ') >= 0)
+                {
+                    useQuotes = true;
+                    break;
+                }
+            }
+
+            if (useQuotes)
+                sink("\"");
```
wilzbach: You could have used backticks

test/archive/har.d
```diff
@@ -0,0 +1,198 @@
...
+
+            foreach (arg; args)
+            {
+                sink(arg);
+            }
+            if (useQuotes)
+                sink("\"");
```
wilzbach: You could have used backticks

test/archive/har.d
```diff
@@ -0,0 +1,198 @@
...
+            mkdirRecurse(dir);
+        }
+        return Yes.success;
+    }
+
+    Flag!"success" extract()
+    {
+        if (outputDir is null)
+        {
+            outputDir = harFilename[0 .. $ - ".har".length];
```
wilzbach: `".har"` could have been a global constant.

test/archive/har.d
```diff
@@ -0,0 +1,198 @@
...
+                    if (!extractMkdir(fullDir))
+                        return No.success;
+                }
+            }
+            auto fullFileName = buildPath(outputDir, fileName);
+            infoFile.writefln("creating %s", fullFileName.formatFile);
+            extractedFilenames.put(fullFileName.idup);
+            {
+                auto currentOutputFile = File(fullFileName, "w");
+                scope(exit) currentOutputFile.close();
```
wilzbach: Redundant. `File` is refCounted and will be closed once the scope is left, so the behavior is already exactly what you manually defined here.

test/archive/har.d
```diff
@@ -0,0 +1,198 @@
...
+            auto spaceIndex = line.indexOf(' ');
+            if (spaceIndex < 0)
+            {
+                infoFile.writefln("%s(1) Error: first line does not contain a space", harFilename);
+                return No.success;
+            }
+            delimiter = line[0 .. spaceIndex + 1].idup;
+        }
+    
+    LfileLoop:
```
wilzbach: This could have been a separate function.

test/archive/har.d
marler8997: This one uses `byLine(Yes.keepTerminator)` so it should work correctly.  This maintains the original line terminator of the HAR file.

test/archive/har.d
marler8997: What criteria do you use to determine when to use `@property`?

wilzbach: Never.
At the moment it has nearly no effect. It just enforces that a member function without argument is called without parenthesis. https://github.com/dlang/DIPs/pull/97 might fix this, but it's still a long journey until then.

See also: https://dlang.org/spec/function.html#property-functions

marler8997: I currently use it as a means to indicate that a "free function" (non-member function) should always be called with UFCS with no arguments.  A way to indicate it should always be called like accessing a "member field/property" indicating it doesn't do much but instead is meant to expose data from the object (which is the first argument).

marler8997:
Copied the latest version of `archive/har.d` from https://github.com/marler8997/har/pull/5/files

test/archive/har.d
jacob-carlborg: @wilzbach doesn't seem to have much affect at all: https://run.dlang.io/is/8WPlhI.

wilzbach: Then it's even more useless than I thought :/

jacob-carlborg: It did have a functionality with the `-property` flag. But that's now deprecated and has now affect.

marler8997: ok, well looks like no one thinks my use has much value so I've removed `@property`

marler8997:
Looks like we've hit the OSX segfault on this PR.

ibuclaw:
Well this will not (initially) work on my side, but manually splitting files is not a problem per say.

But do bear in mind that this will just add more time onto the testsuite run when - depending on the bug - doing as per this issue might be better.

https://issues.dlang.org/show_bug.cgi?id=18235

> Looks like we've hit the OSX segfault on this PR.

Why does it seem like you're the only person who gets this after making a change. ;-)

wilzbach:
FWIW the spurious failures of osx with its segfaults even appear on PRs to phobos now: https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3081734&isPull=true

marler8997:
It probably seems like Im the only one that gets them because I spent about 4 hours reproducing it in a macbook and know what it looks like. Others have definitely gotten it, but I assume most failures go unreported since its so intermittent, most people proba ly just rebase without ever knowing what actually failed.

marler8997:
Rebased to see if the OSX failure goes away.  This one is more curious since only test files were modified but the failure occurs when compiling `dmd-unittest` which doesn't use any test files as far as I'm aware.

@ibuclaw I didn't quite understand what you were trying to say.  You mentioned something wouldn't work on your side..what were you referring to?  I also wasn't sure what you were trying to say when talking about this adding more time to the test suite...?

marler8997:
Rebase didn't fix it...it looks like there's a bunch of failures with OSx 32 bit right now.  Maybe some change got into the compiler that usually fails OSx 32 bit but got lucky and made it through integration testing...

ibuclaw:
> @ibuclaw I didn't quite understand what you were trying to say. You mentioned something wouldn't work on your side..what were you referring to?

GDC, what else would I be referring to?

> I also wasn't sure what you were trying to say when talking about how this is adding more time to the test suite...?

This will slow down the testsuite run time. We've already dumped TravisCI because the time it took to finish slowed down so much we couldn't function anymore. There's just a wish of not repeating the same with SemaphoreCI as there probably won't be a faster platform to turn to next time.

marler8997:
> GDC, what else would I be referring to?

I don't appreciate the "snarky" response.  What specifically does this break in GDC?

> This will slow down the testsuite run time.

By far the biggest time sink with the test suite is invoking the compiler.  Especially when a test is permuting it's arguments.  I don't anticipate this having a noticeable impact on overall performance, but it is better to measure than theorize.

ibuclaw:
This is shared code and every change potentially breaks another compilers workflow. Just be aware of that.

marler8997:
> This is shared code and every change potentially breaks another compilers workflow. Just be aware of that.

Oh thanks for telling me that...I had no idea.

marler8997:
I think you just like trolling my pull requests making bogus claims like "this is going to break things for me".  Then when I ask what you to explain your claim you go silent or ignore the question.

Then you come back with a comment like "be aware that this code is shared between code bases". 
 How is it helpful to tell me something so obvious? Seems like you're just sidestepping the question because you know there's no substance behind your claim.  Again...how is this breaking GDC?  You made the claim, back up what you say or don't say it at all.

marler8997:
Ping

JinShil:
@ibuclaw it appears you have reservations about this and had labeled it as "Need Work"  What specifically needs to be done to make this acceptable, or are you against this approach in general.

It will likely add time to the test suite, but I don't know if the time would be significant.

Assigning to @ibuclaw for further consideration and clarification; how do we get to a yay or nay?

wilzbach:
> But do bear in mind that this will just add more time onto the testsuite run when - depending on the bug - doing as per this issue might be better.
> https://issues.dlang.org/show_bug.cgi?id=18235

FWIW I think the costly bit are the `runnable` tests.
All 1283 `fail_compilation` tests compile in ~11s on my machine.
The 774 `compilable` tests are slightly slower with ~25s ... and you don't want to know how long `runnable` takes :/
So if someone wants to make the testsuite faster, the `runnable` tests are the real bad boys.
A simple optimization could be to only link druntime if nothing from Phobos is imported.

Anyhow, I don't think that a creating a few files will lead to a measurable effect - compared to all the crazy things DMD and the linker do. I think @marler8997 even measured it once (?), but I can't find the reference right now.

marler8997:
I analyzed timing on the autotester once upon a time (in #7099).  The trick was to make sure you were comparing the same machine.

marler8997:
@ibuclaw could you let me know what you were wanting to be changed?  This PR is one that I'm waiting for before I add more testing (i.e. tests for import order invariance which require many small files per test). Thanks.

marler8997:
Pinging @ibuclaw 

You last activity on this PR was 20 days ago where you added the "Needs Work" label without clarifying what needed work.  Now this PR has been assigned you to and I'm waiting for your feedback.  This is blocking other work, please let us know what you'd like changed.

ibuclaw:
I could see this being useful on run.dlang.io, so you should try submitting it there.

Regarding inclusion here, the testsuite is not the place to add more tooling into the mix without good reason, to which I can only see negative points to this.

1. Testing single test that has been "packed" becomes non-trivial.

2. Makes the test run more complex by adding in a new step, rather than taking one away.

3. Said step exposes a new point of failure that didn't exist before.

4. Intention for the change is sketchy.
  - Is the idea is to make it more clearer to a reviewer what are all the files are in a single test?
    - Can be solved in another way that doesn't alter the way tests are written.
 - Is the idea to make it easier to write/manage tests?
    - This is not a problem that needs solving.
  - Is the idea to obsolete the `EXTRA_SOURCES` test runner option?
    - No, no, NO!

5. Breaks downstream for no reason or gain. Files that were once useable can no longer be tested within the testing frameworks used by downstream compilers, i.e: this bespoke format will have to be implemented in dejagnu so that remote execution and emulation continue to work.

This is not something worth pursuing.

andralex:
Hi folks, please don't close one another's PRs. I know from personal experience it creates a lot of animosity instantly, even when done by folks with authority such as Walter himself. What I do after a number of mistakes was to discuss with people so they close their PRs instead of me closing them. Thanks. I'm reopening this as a courtesy.

That said, I agree with @ibuclaw 's assessment. This PR is illustrative of a trend that goes somewhat sideways relative to what we want. The trend has some of the following:

* Large code reorganizations of which benefit is tenuous to argue, and therefore needs vigorous argumentation.
* Change is labor-intensive but of the mechanical kind, often doable with a simple tool.
* Change reflects more of a preference of the author, than a universally desirable trait that all reasonable people agree with.

We are looking for the opposite:

* Small code changes of obvious and high impact
* Change is creative and ingenious
* Change is motherhood and apple pie: make all of this state private, improve information hiding, improve modularity, decouple modules, measurably improves a desirable metric, etc.

I've reopened this in the spirit of fair play but suggest it be closed by the author.

marler8997:
@andralex thanks for taking a look.  I think all arguments have been presented so I'll accept your decision to close.

PR #7906, opened 2018-02-17
correct namespace mangling for Issue 17772

WalterBright:
This fixes the lack of namespace mangling for templates in https://issues.dlang.org/show_bug.cgi?id=17772

Fixing the `T_` mangling is much more problematic, I don't know how to do that yet. This namespace issue is independent, so it is convenient to do it separately.

### DIFF
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -584,7 +584,20 @@ private final class CppMangleVisitor : Visitor
              */
             TemplateInstance ti = d.parent.isTemplateInstance();
             assert(ti);
-            source_name(ti);
+            Dsymbol p = ti.toParent();
+            if (p && !p.isModule() && tf.linkage == LINK.cpp)
+            {
+                buf.writeByte('N');
+                CV_qualifiers(d.type);
+                prefix_name(p);
+                if (d.isDtorDeclaration())
+                    buf.writestring("D1");
+                else
+                    source_name(ti);
+                buf.writeByte('E');
+            }
+            else
+                source_name(ti);
             headOfType(tf.nextOf());  // mangle return type
         }
         else
--- a/test/compilable/cppmangle.d
+++ b/test/compilable/cppmangle.d
@@ -438,3 +438,15 @@ version (linux)
 {
     static assert(test36.mangleof == "_Z6test36PPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPiPS12_");
 }
+
+/*****************************************/
+// https://issues.dlang.org/show_bug.cgi?id=17772
+
+extern(C++, SPACE)
+int test37(T)(){ return 0;}
+
+version (Posix) // all non-Windows machines
+{
+    static assert(test37!int.mangleof == "_ZN5SPACE6test37IiEEiv");
+}
+


src/dmd/cppmangle.d
```diff
@@ -584,7 +584,20 @@ private final class CppMangleVisitor : Visitor
...
+                CV_qualifiers(d.type);
+                prefix_name(p);
+                if (d.isDtorDeclaration())
+                    buf.writestring("D1");
+                else
+                    source_name(ti);
+                buf.writeByte('E');
+            }
+            else
+                source_name(ti);
```
ibuclaw: This just seems to be a duplicate of below, maybe it could be refactored in such a way that they are made a bit more common?

Perhaps it could be benefited from splitting into a new function.

WalterBright: It's not quite a duplicate. I also wanted to make this easy to review, and including a refactoring of other code would make it harder.

test/compilable/cppmangle.d
```diff
@@ -438,3 +438,15 @@ version (linux)
...
     static assert(test36.mangleof == "_Z6test36PPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPPiPS12_");
 }
+
+/*****************************************/
+// https://issues.dlang.org/show_bug.cgi?id=17772
+
+extern(C++, SPACE)
+int test37(T)(){ return 0;}
+
+version (Posix)
```
JinShil: Why only POSIX?  Is the mangling different on Windows?

ibuclaw: Well, strictly we should be using another version identifier here, as the difference is IA64 vs. MSVC++, not Posix vs. Windows.

There's also nothing stopping you from supporting other C++ compilers, or varying versions of their mangling standard.

ibuclaw (approved):
Seems fine, but I still think that this could be moved to a function. Each part of the C++ Abi should be a trivial component, and that can be easily mirrored in the implementation also.

test/compilable/cppmangle.d
JinShil: If that's the case, please add a comment, at a minimum.

JinShil (dismissed):

test/compilable/cppmangle.d
WalterBright: Add a comment saying what?

JinShil: Saying something about this:

> strictly we should be using another version identifier here, as the difference is IA64 vs. MSVC++, not Posix vs. Windows.

WalterBright: The name mangling is standardized across non-Windows machines, it is not an IA64 thing. But I added a comment anyway.

src/dmd/cppmangle.d
```diff
@@ -584,7 +584,20 @@ private final class CppMangleVisitor : Visitor
...
             assert(ti);
-            source_name(ti);
+            Dsymbol p = ti.toParent();
+            if (p && !p.isModule() && tf.linkage == LINK.cpp)
+            {
+                buf.writeByte('N');
+                CV_qualifiers(d.type);
+                prefix_name(p);
+                if (d.isDtorDeclaration())
+                    buf.writestring("D1");
```
wilzbach: I thought we don't merge PRs anymore if the introduce uncovered code (and don't have a good reason to do so)?

ibuclaw: If it was refactored to be common code, then it would show as being  covered.

test/compilable/cppmangle.d
```diff
@@ -438,3 +438,15 @@ version (linux)
...
+
+/*****************************************/
+// https://issues.dlang.org/show_bug.cgi?id=17772
+
+extern(C++, SPACE)
+int test37(T)(){ return 0;}
+
+version (Posix) // all non-Windows machines
+{
+    static assert(test37!int.mangleof == "_ZN5SPACE6test37IiEEiv");
```
Geod24: Next time we introduce an *incorrect* mangling in the test suite, please either:
- Add a comment (e.g.: "This should be `_ZN5SPACE6test37IiEET_v` but this test a partial fix for issue 17772")
- Just test the relevant part (e.g. static assert(test37!int.mangleof[0 .. 16] == "_ZN5SPACE6test37")

Introducing tests that are testing a value is wrong is just plain confusing for the people actually fixing the issue, because suddenly the test suite fails and there's no indication the test is not correct.

PR #6246, opened 2016-11-06
Pass static arrays in registers when possible.

LemonBoy:
While investigating [why](https://github.com/ldc-developers/ldc/issues/960) LDC sometimes didn't eliminate a tail call I stumbled across this.

PR #7907, opened 2018-02-17
refactor argtypes.d

WalterBright:
Simplify the logic and separate the 32 and 64 bit paths instead of interleaving them.


WalterBright:
I'd like to get this pulled so I can go about replacing #6246

JinShil (approved):

WalterBright:
@braddr The Freebsd 64 machine appears to be dead? Is it on the right pills?

PR #7912, opened 2018-02-18
fix Issue 18459 - The static array ABI doesn't match the struct ABI

WalterBright:
The idea is to have static array use the same code as structs, so they will behave the same. It's a replacement for #6246


src/dmd/argtypes.d
```diff
@@ -41,6 +42,28 @@ TypeTuple toArgTypes(Type t)
...
+        }
+
+        ///
+        void oneType(Type t)
+        {
+            result = new TypeTuple(t);
+        }
+
+        ///
+        void twoTypes(Type t1, Type t2)
```
ibuclaw: Am I having deja vu? I thought this was already added as part of another refactoring pr. Must have been somewhere else.

src/dmd/argtypes.d
```diff
@@ -256,64 +260,124 @@ TypeTuple toArgTypes(Type t)
...
+                offset = cast(uint)(n * tn.size());
+                alignsize = tn.alignsize();
+                return tn;
             }
 
-            void oneType(Type t)
+            aggregate(t.size(Loc.initial), cast(size_t)t.dim.toInteger(), &getNthField);
+        }
+/+
+        override void visit(TypeSArray t)
```
ibuclaw: Dead code should be removed (this is commented out entirely)

src/dmd/argtypes.d
```diff
@@ -434,8 +502,10 @@ TypeTuple toArgTypes(Type t)
                      * Cannot find any documentation on it.
                      */
 
-                    VarDeclaration f = t.sym.fields[0];
-                    TypeTuple tup = toArgTypes(f.type);
+                    uint foffset;
+                    uint falignsize;
+                    Type ftype = getFieldInfo(0, foffset, falignsize);
+                    TypeTuple tup = toArgTypes(ftype);
```
ibuclaw: I wonder who originally added this. It seems a bit odd to have a FreeBSD special case here.

src/dmd/argtypes.d
```diff
@@ -29,6 +29,7 @@ import dmd.visitor;
  * Returns:
  *      tuple of types, each element can be passed in a register.
  *      A tuple of zero length means the type cannot be passed/returned in registers.
+ *      null indicates a `void`.
```
ibuclaw: But none of the callers of toArgTypes check for `null` as far as I recall.

ibuclaw (commented):
The aggregate function itself seems reasonable.

src/dmd/argtypes.d
WalterBright: I moved them up a level so they would be accessible from other functions.

src/dmd/argtypes.d
WalterBright: I think it was me. Nevertheless, it is for real.

src/dmd/argtypes.d
WalterBright: Good reason to document it!

src/dmd/argtypes.d
ibuclaw: https://github.com/dlang/dmd/pull/7230

Found it and commented there. Very certain that this should be turned on for other non-Windows x86 platforms too.  I've been looking at the gcc x86 backend recently to implement what this module effectively does in gcc itself.

test/runnable/testargtypes.d
```diff
@@ -104,6 +104,13 @@ version (X86_64)
         struct S9 { int a,b,c; }
         chkArgTypes!(S9, long, int)();
 
+        struct S10 { int[3] a; }
+        chkArgTypes!(S10, long, int)();
+
+        chkArgTypes!(int[3], long, int)();
+
+        chkArgTypes!(S9[1], long, int)();
```
ibuclaw: Are you checking `S9` or `S10` here?  Tests against `S9` should be grouped together to prevent confusion.

ibuclaw (dismissed):

test/runnable/test_cdstrpar.d
```diff
@@ -259,6 +259,6 @@ void main()
             fprintf(stderr, "\n");
             failure = true;
         }
-        assert(!failure);
+//        assert(!failure);
```
JinShil: This makes me nervous.  Why is the assertion commented out?

WalterBright: Because it's a long series of tests of exact code generation. Perturb that and it all fails. It's a nuisance.

WalterBright:
Please can we move forward with this. There are several other ABI issues that will build on this.

test/runnable/test_cdstrpar.d
wilzbach: That's why there's an update script for it - it even has the added advantage that the exact effect of the code gen changes can be inspected ...

JinShil: @WalterBright Please see https://github.com/dlang/dmd/blob/ad598ecaf3eae57531c848e133fd70b2d85d19c0/test/runnable/test_cdstrpar.d#L8-L9

```D
// Run this after codegen changes:
// env DMD=generated/linux/release/64/dmd rdmd -fPIC -version=update test/runnable/test_cdstrpar.d
```

If you do that, you can keep this test in tact and I can then approve this PR.

JinShil: Either that or, if this test has no value, then it should be deleted.  I'm not sure why it's necessary to test precise code gen rather than just testing behavior.  @MartinNowak What was your justification for this?

test/runnable/testargtypes.d
WalterBright: this was fixed.

WalterBright:
@wilzbach @JinShil ping

JinShil (approved):

ibuclaw (approved):

PR #7916, opened 2018-02-19
Set CallExp's function declaration immediately for criticalEnter/Exit

JinShil:
This is an alternative to #6840.  Please see the discussion there for context.

Essentially, this is a courtesy PR for LDC to reduce DMD/LDC diff and relieve a minor inconvenience [described here](https://github.com/dlang/dmd/pull/6840#issuecomment-304742970).

This requires @thewilsonator's approval.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -5439,6 +5439,19 @@ extern (C++) final class CallExp : UnaExp
         this.arguments = arguments;
     }
 
+    /***********************************************************
+    * Instatiates a new function call expression
+    * Params:
+    *       loc   = location
+    *       fd    = the declaration of the function to call
+    *       earg1 = the function argument
+    */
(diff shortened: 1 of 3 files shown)


thewilsonator:
Looks fine.

WalterBright (approved):

src/dmd/expression.d
```diff
@@ -5454,6 +5467,18 @@ extern (C++) final class CallExp : UnaExp
...
+    * Creates a new function call expression
+    * Params:
+    *       loc   = location
+    *       fd    = the declaration of the function to call
+    *       earg1 = the function argument
+    */
+    static CallExp create(Loc loc, FuncDeclaration fd, Expression earg1)
+    {
+        return new CallExp(loc, fd, earg1);
+    }
```
wilzbach: Not really sure why we need this `create` methods and can't let C++ call the constructor, but well that's a matter for another day ...

wilzbach (approved):

PR #7917, opened 2018-02-19
Add changelog entry for -dip1008 + fix typo in the lambda comparison changelog entry

wilzbach:


changelog/dip1008.dd
```diff
@@ -0,0 +1,12 @@
...
+[DIP1008](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1008.md) has been merged and it can be previewed under the experimental `-dip1008` flag:
+
+---
+void main() @nogc
+{
+    throw new Exception("I'm fully @nogc now");
+}
+---
+
+$(CONSOLE rdmd -dip1008 app.d)
```
jacob-carlborg: I would prefer a more descriptive changelog entry, or a link to the specification. The linked DIP doesn't even say it's accepted. I think it should mentioned how the exception is allocated and that it's reference counted. The DIP also mentions breaking changes. It would be nice to know the solutions to those breaking changes. Or do we wait until this is the default behavior?

jacob-carlborg (changes requested):

changelog/dip1008.dd
wilzbach: > I would prefer a more descriptive changelog entry,

Hmm, it's not "officially" released (and as you noticed the DIP isn't approved either).
The changelog already mentions so with "experimental" and "preview" and I think it's reasonable to expect the alpha warriors to read the DIP, but if you 

> or a link to the specification.

The spec doesn't "know" about DIP1008. The DIP is the best we got.

>  think it should mentioned how the exception is allocated and that it's reference counted. 

That's an implementation detail and might change if more work happens in DIP1008.
Using reference counting for this sole instance isn't universally loved.

> It would be nice to know the solutions to those breaking changes. 

I'm the wrong person to ask.
I'm just trying to address Martin's comment "There is experimental `@nogc` exception throwing `-dip1008` (still
missing changelog)."

https://forum.dlang.org/post/dd713ad8-a985-9881-5c5d-4966815f3b9b@dawg.eu

> Or do we wait until this is the default behavior?

I think so.

jacob-carlborg: Fine, but I don't like it.

jacob-carlborg (approved):

PR #7918, opened 2018-02-19
Fix Issue 18468 - cannot use `synchronized {}` in @safe code

wilzbach:
DRuntime already uses `shared` internally: https://github.com/dlang/druntime/blob/master/src/rt/critical_.d#L32

### DIFF
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -3460,14 +3460,15 @@ else
         else
         {
             /* Generate our own critical section, then rewrite as:
-             *  __gshared align(D_CRITICAL_SECTION.alignof) byte[D_CRITICAL_SECTION.sizeof] __critsec;
-             *  _d_criticalenter(__critsec.ptr);
-             *  try { body } finally { _d_criticalexit(__critsec.ptr); }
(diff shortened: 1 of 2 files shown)


src/dmd/statementsem.d
```diff
@@ -3460,14 +3460,15 @@ else
         else
         {
             /* Generate our own critical section, then rewrite as:
-             *  __gshared align(D_CRITICAL_SECTION.alignof) byte[D_CRITICAL_SECTION.sizeof] __critsec;
+             *  shared align(D_CRITICAL_SECTION.alignof) byte[D_CRITICAL_SECTION.sizeof] __critsec;
              *  _d_criticalenter(__critsec.ptr);
              *  try { body } finally { _d_criticalexit(__critsec.ptr); }
```
thewilsonator: Update the comment to 
```d
*  _d_criticalenter(&__critsec[0]);
*  try { body } finally { _d_criticalexit(&__critsec[0]); }
```

wilzbach:
Changed the comment.

Anything else blocking this? Imho `synchronized` should really, really work in `@safe`.
Imho we should really try to get it in with 2.079.

Anyone around for a quick pull?

WalterBright (approved):

PR #7924, opened 2018-02-20
Type-Safe template wrapper for AA

marler8997:


src/dmd/root/aav.d
```diff
@@ -170,3 +170,42 @@ unittest
...
+
+    Params:
+        key = key to lookup the value for
+
+    Returns: the value associated with `key` if present, otherwise, null.
+    */
+    V opIndex(const(K) key)
+    {
+        return cast(V)dmd_aaGetRvalue(aa, cast(void*)key);
+    }
```
wilzbach: How about adding `opIndexAssign` too?

```d
    auto opIndexAssign(V value, K key)
    {
        auto ptr = dmd_aaGet(&aa, cast(void*) key);
        *ptr = cast(void*) value;
        return *ptr;
    }
```

src/dmd/root/aav.d
```diff
@@ -170,3 +170,42 @@ unittest
...
+}
```
wilzbach: How about adding a public example:

```d
///
unittest
{
    AssocArray!(string, int) aa;

    assert(aa["foo"] == 0);
    aa["foo"] = 2;
    assert(aa["foo"] == 2);
    assert(aa.length == 1);
}
```

wilzbach (commented):
Nice. I was about to add this too, but you beat me to it :)

src/dmd/root/aav.d
marler8997: Currently no code uses it this way and I don't anticipate code ever doing this.
```
aa[key] = value;
```
This being the more "intuitive" usage, I have to assume that Walter deliberately chose not to support this.

timotheecour: maybe it's just because it wasn't implemented? maybe ask @WalterBright ?

marler8997: I don't think it's something worth bugging him about.  Currently there's no use for it.  We can always add it later if a use comes to light.

JinShil: Keep in mind that the current code base was translated from C+ (not a typo).  I don't think there was any deliberate effort by Walter to avoid it.

marler8997: I'm also basing this on comments he made about one of @ibuclaw's PRs where he made a change that ended up causing multiple hashes of the same key.  The current implementation uses `getLvalue` which allows the developer to do multiple operations on a value with only one lookup.  I think the API was designed to encourage this behavior, and adding `opIndexAssign` would provide a way to circumvent it, making it easy for developers to accidentally write code that has to hash/lookup multiple times.

marler8997: Here's Walter's comment: https://github.com/dlang/dmd/pull/7696#discussion_r161407629

JinShil (approved):

wilzbach (approved):
Nice. I added this to the "no objection -> merge" queue.

nordlow:
Whats the purpose with this addition?

marler8997:
I thought the purpose was self evident from the change...what do you think?

nordlow:
Is the long-term goal with this change to make the default AA-implementation user-configurable?

marler8997:
No long term goal in mind for this change, the change is its own end goal, a type safe template wrapper around AA.

wilzbach:
This has nothing to do with druntime. It's about introducing more type safety into the dmd codebase which uses its own AA-implementation back from the C++ times.

I think what you are looking for is something like https://github.com/dlang/druntime/pull/1282

nordlow:
Ahh, now I get it. Thanks, @wilzbach

PR #7928, opened 2018-02-20
Move common BASH test logic to wrapper

marler8997:
This PR creates a wrapper to hold common logic/settings for all BASH tests.  Currently, this wrapper will pipe `stdout`/`stderr` to the test output file, and when a failure occurs, will dump the output file with a message saying the test failed (just like `d_do_test` does with `.d` tests).

It also exports common variables to the test script, currently
* `${TEST_DIR}` (i.e. compilable/runnable)
* `${TEST_NAME}` being the base name of the test.

TODO:

Maybe `set -euo pipefail` and/or `set -x` in wrapper?

### DIFF
--- a/test/compilable/crlf.sh
+++ b/test/compilable/crlf.sh
@@ -2,9 +2,8 @@
 
 # Test CRLF and mixed line ending handling in D lexer.
 
-name=`basename $0 .sh`
 dir=${RESULTS_DIR}/compilable
-fn=${dir}/${name}.d
+fn=${TEST_DIR}/${TEST_NAME}.d
 
 printf '%s\r\n' \
        '#!/usr/bin/env dmd -run' \
@@ -46,8 +45,8 @@ printf 'static assert(wstr == "%s");\n' 'foo\nbar\nbaz\n' >> ${fn}
 printf 'enum dstr = q"(\r\nfoo\r\nbar\nbaz\r\n)";\n' >> ${fn}
 printf 'static assert(dstr == "%s");\n' '\nfoo\nbar\nbaz\n' >> ${fn}
 
-$DMD -c -D -Dd${dir} -m${MODEL} -of${dir}/${name}a${OBJ} ${fn} || exit 1
+$DMD -c -D -Dd${TEST_DIR} -m${MODEL} -of${TEST_DIR}/${TEST_NAME}a${OBJ} ${fn} || exit 1
 
-rm -f ${dir}/${name}a${OBJ} ${dir}/${name}.html ${fn}
+rm -f ${TEST_DIR}/${TEST_NAME}a${OBJ} ${TEST_DIR}/${TEST_NAME}.html ${fn}
 
-echo Success >${dir}/`basename $0`.out
+echo Success
--- a/test/compilable/ddoc9764.sh
+++ b/test/compilable/ddoc9764.sh
@@ -1,11 +1,9 @@
 #!/usr/bin/env bash
 
-name=`basename $0 .sh`
-dir=${RESULTS_DIR}/compilable
-output_file=${dir}/${name}.html
+output_file=${RESULTS_DIR}/${TEST_DIR}/${TEST_NAME}.html
 
 rm -f ${output_file}
 
 $DMD -m${MODEL} -D -o- compilable/extra-files/ddoc9764.dd -Df${output_file}
 
-compilable/extra-files/ddocAny-postscript.sh 9764 && touch ${dir}/`basename $0`.out
+compilable/extra-files/ddocAny-postscript.sh 9764 && touch ${TEST_DIR}/${TEST_NAME}.out
--- a/test/sh_do_test.sh
+++ b/test/sh_do_test.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+if [ "${RESULTS_DIR}" == "" ]; then
+    echo Note: this program is normally called through the Makefile, it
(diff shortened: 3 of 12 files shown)


test/bash_test_wrapper.sh
```diff
@@ -0,0 +1,22 @@
...
+echo " ... /${script_name}"
+
+output_file=${RESULTS_DIR}/${category}/${test_name}.out
+
+./${script_name} > ${output_file}  2>&1
+if [ $? -ne 0 ]; then
+    # duplicate d_do_test output
+    echo >> ${output_file}
+    echo ============================== >> ${output_file}
+    echo Test failed >> ${output_file}
```
wilzbach: It's good to state here which test failed as the tests might run in high parallelism.

test/compilable/crlf.sh
```diff
@@ -2,9 +2,8 @@
 
 # Test CRLF and mixed line ending handling in D lexer.
 
-name=`basename $0 .sh`
 dir=${RESULTS_DIR}/compilable
```
wilzbach: this is no longer used?

test/compilable/crlf.sh
```diff
@@ -2,9 +2,8 @@
...
-fn=${dir}/${name}.d
+fn=${TEST_DIR}/${TEST_NAME}.d
```
wilzbach: In a next step `TEST_FILE` could be exported as well.

test/compilable/ddoc9764.sh
```diff
@@ -1,11 +1,9 @@
 #!/usr/bin/env bash
 
-name=`basename $0 .sh`
-dir=${RESULTS_DIR}/compilable
-output_file=${dir}/${name}.html
+output_file=${RESULTS_DIR}/${TEST_DIR}/${TEST_NAME}.html
```
wilzbach: This seems to be quite frequent enough.
```
export OUTPUT_FILE="${RESULTS_DIR}/${TEST_DIR}/${TEST_NAME}"
```

could be part of a next step.

test/sh_do_test.sh
```diff
@@ -0,0 +1,29 @@
...
+
+if [ "${RESULTS_DIR}" == "" ]; then
+    echo Note: this program is normally called through the Makefile, it
+    echo is not meant to be called directly by the user.
+    exit 1
+fi
+
+# TEST_DIR should be one of compilable, fail_compilation or runnable
+export TEST_DIR=$1
+export TEST_NAME=$2
```
wilzbach: There are a few more variables which I frequently need, but I think we can easily go step by step here.

test/sh_do_test.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
+
```
wilzbach: `set -u -o pipefail` (doesn't have to be done in this PR)

test/sh_do_test.sh
```diff
@@ -0,0 +1,29 @@
...
+export TEST_DIR=$1
+export TEST_NAME=$2
+script_name=${TEST_DIR}/${TEST_NAME}.sh
+
+echo " ... ${script_name}"
+
+output_file=${RESULTS_DIR}/${script_name}.out
+rm -f ${output_file}
+
+./${script_name} > ${output_file}  2>&1
```
wilzbach: A future extension could use `source` here.
Advantages:
- use a trap, to show the error of the failure
- activate `set -ue -o pipefail` by default
- activate `set +x` by default, redirect the debug log into a `XYZ.debug` file and only show it when there's a failure.

wilzbach (approved):
I'm strongly voting for this. Redefining the same variables in the bash scripts has become quite a pain. 
Thanks a lot @marler8997 for pushing this!

I even have a few idea to make this more useful (a few more default exported variables, a default debug trap, default `set -eu o pipefail`, ...), but I'm more than happy to do this in a quick follow-up.

CC @MartinNowak @CyberShadow

wilzbach:
> Maybe set -euo pipefail and/or set -x in wrapper?

Yes, we should do this, but I think your PR is already a good start and makes a good base for future improvements like this one (or the other potential ones I have mentioned).

test/compilable/crlf.sh
marler8997: yeah, replaced with `${TEST_NAME}`

test/compilable/ddoc9764.sh
marler8997: this one is common enough, I'll add it to this PR

marler8997: actually, wait, this is specific to this test only...it has the `.html` extension

marler8997:
Ok yeah, we can merge as is and I'll let you enhance it (you're better at BASH anyway)

test/compilable/ddoc9764.sh
wilzbach: Yes, that's why `export OUTPUT_FILE="${RESULTS_DIR}/${TEST_DIR}/${TEST_NAME}"` which allows the user to pick the extension, but yeah, let's think about this in the next step.

marler8997: Yeah, I wasn't sure the best variables to expose...there's combinations of the source dir vs results dir and different extensions...we can figure it out in the next step like you said

wilzbach:
No one has objected to this change within one week, it's not critical DMD code, but just affects the test suite and we need to move on -> merging.

PR #7929, opened 2018-02-20
Fix Issue 18460 - Improve error message for missing 'new' for class instantiation

wilzbach:
One small step on the long road to better diagnostics.

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -2560,7 +2560,12 @@ extern (C++) abstract class Type : RootObject
                 if (s)
                     error(loc, "no property `%s` for type `%s`, did you mean `%s`?", ident.toChars(), toChars(), s.toPrettyChars());
                 else
-                    error(loc, "no property `%s` for type `%s`", ident.toChars(), toChars());
+                {
+                    if (ident == Id.call && ty == Tclass)
+                        error(loc, "no property `%s` for type `%s`, did you mean `new %s`?", ident.toChars(), toChars(), toPrettyChars());
+                    else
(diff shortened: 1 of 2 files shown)


src/dmd/mtype.d
```diff
@@ -2560,7 +2560,12 @@ extern (C++) abstract class Type : RootObject
                 if (s)
                     error(loc, "no property `%s` for type `%s`, did you mean `%s`?", ident.toChars(), toChars(), s.toPrettyChars());
                 else
-                    error(loc, "no property `%s` for type `%s`", ident.toChars(), toChars());
+                {
+                    if (ident == Id.call && ty == Tclass)
+                        error(loc, "no property `%s` for type `%s`, did you mean `new %s`?", ident.toChars(), toChars(), toPrettyChars());
```
wilzbach: BTW any reason why `toChars` is always fully qualified for classes, but not for structs?

JinShil: I don't know.  I suppose that should also be addressed (in another PR).

JinShil (approved):

jacob-carlborg (approved):

test/fail_compilation/diag18460.d
```diff
@@ -0,0 +1,13 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/diag18460.d(12): Error: no property `opCall` for type `diag18460.Foo`, did you mean `new Foo`?
```
ZombineDev: ``Error: no `static` property `opCall` for...``

PR #7931, opened 2018-02-21
statementsem.d: add helpful spec URLs

WalterBright:
In general, we should put links to the spec where code that implements it is.

### DIFF
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -164,6 +164,9 @@ private extern (C++) final class StatementSemanticVisitor : Visitor
 
     override void visit(ExpStatement s)
     {
+        /* https://dlang.org/spec/statement.html#expression-statement
+         */
+
         if (s.exp)
         {
             //printf("ExpStatement::semantic() %s\n", exp.toChars());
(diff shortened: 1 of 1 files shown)


src/dmd/statementsem.d
```diff
@@ -2074,6 +2091,9 @@ else
 
     override void visit(IfStatement ifs)
     {
+        /* https://dlang.org/spec/statement.html#IfStatement
+         */
```
JinShil: Do we really need to increase the height of the source code with the trailing `*/`.  Can we at least put it on one line or use the `//` comment syntax?

JinShil (changes requested):

src/dmd/statementsem.d
WalterBright: I like using /* */ for comments on their own line, and // for comments appended to a line.

WalterBright: The use of a block comment increases the "weight" of a comment, which is appropriate for one that is intended to document a block of code rather than a single line. // implies lightweight, line-only.

JinShil (approved):

PR #7932, opened 2018-02-21
[Refactor] Create visitor for resolve method of type AST nodes

RazvanN7:
cc @WalterBright

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -25,7 +25,6 @@ import dmd.arrayop;
 import dmd.arraytypes;
 import dmd.gluelayer;
 import dmd.complex;
-import dmd.dcast;
 import dmd.dclass;
 import dmd.declaration;
 import dmd.denum;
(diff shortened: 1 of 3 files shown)


RazvanN7:
The jenkins failure is most likely unrelated since this PR does not modify any code logic.

wilzbach (approved):
LGTM, though `resolve.d` would have been a real option here.

JinShil (approved):

andralex:
@ibuclaw ping just to make sure we're good here

ibuclaw:
@andralex so you merged it anyway?

andralex:
@ibuclaw yah, under the assumption we can easily undo the merge or make the necessary changes. Optimistic concurrency control :)

ibuclaw:
The resolve function is probably a candidate for making extern(D), I can think of no reason off the top of my head why it needs to be publicly visible (outside of D code).

PR #7934, opened 2018-02-21
[Refactor] Create visitor for dotExp method of type AST nodes

RazvanN7:
cc @WalterBright


jacob-carlborg:
This doesn't need to be exposed in the C++ headers?

RazvanN7:
>This doesn't need to be exposed in the C++ headers?

When I created visitors for the semantic methods I didn't have to so I assumed that's fine not to in this case.

wilzbach (approved):
Can we please go with `dotexp.d` if we already move code around?
(ideally all of these would be in a `visitors` package, but well seems like [this](https://github.com/dlang/dmd/pull/7639) ain't happening)

JinShil:
@RazvanN7 Please rebase if you want this to move forwarrd.

RazvanN7:
@JinShil done, although I don't know why it fails on Darwin

jacob-carlborg:
If you only moved some code you might have missed the changes that the latest objc PR did to in the `dotExp` method. This PR was recently merged.

jacob-carlborg:
@RazvanN7 This code needs to be included: https://github.com/dlang/dmd/blob/bf824c6170cab5a152668fe488a87954fea07415/src/dmd/mtype.d#L8411-L8419

RazvanN7:
@jacob-carlborg thank you very much! Done.

src/dmd/typesem.d
```diff
@@ -2543,3 +2548,1371 @@ private extern(C++) final class ResolveVisitor : Visitor
...
+ *  resulting expression with e.ident resolved
+ */
+extern(C++) Expression dotExp(Type mt, Scope* sc, Expression e, Identifier ident, int flag)
+{
+    scope v = new DotExpVisitor(sc, e, ident, flag);
+    mt.accept(v);
+    return v.result;
+}
+
+private extern(C++) final class DotExpVisitor : Visitor
```
andralex: Make extern(D) since it's not private anyway

src/dmd/typesem.d
```diff
@@ -2543,3 +2548,1371 @@ private extern(C++) final class ResolveVisitor : Visitor
...
+ *  mt = type for which the dot expression is used
+ *  sc = instantiating scope
+ *  e = expression to convert
+ *  ident = identifier being used
+ *  flag = DotExpFlag bit flags
+ *
+ * Returns:
+ *  resulting expression with e.ident resolved
+ */
+extern(C++) Expression dotExp(Type mt, Scope* sc, Expression e, Identifier ident, int flag)
```
andralex: cc @ibuclaw does the C++ interface need this at all? If not let's make it `extern(D)` for better modularity.

andralex (approved):
I'd need to understand better what the C++ interface needs and what it doesn't so we minimize the surface. cc @ibuclaw @JohanEngelen @klickverbot @redstar

PR #7935, opened 2018-02-21
__traits(isSame) for lambda functions which contain function calls

RazvanN7:
This PR adds the capability of comparing lambda functions which contain calls to other functions.

Two functions are considered equal if their serializations is equal. 2 function serializations are equal if : 

-> the mangled names of the functions are equal
-> the serializations of the parameters with which the function is called are equal

### DIFF
--- a/src/dmd/lambdacomp.d
+++ b/src/dmd/lambdacomp.d
@@ -48,7 +48,33 @@ private enum ExpType
 }
 
 /**
- * The serialize visitor computes the string representation of a
+ * Compares 2 lambda functions described by their serialization.
+ *
+ * Params:
+ *  l1 = first lambda to be compared
+ *  l2 = second lambda to be compared
+ *  sc = the scope where the lambdas are compared
+ *
+ * Returns:
+ *  `true` if the 2 lambda functions are equal, `false` otherwise
+ */
+bool isSameFuncLiteral(FuncLiteralDeclaration l1, FuncLiteralDeclaration l2, Scope* sc)
+{
+    if (auto ser1 = getSerialization(l1, sc))
+    {
+        //printf("l1 serialization: %s\n", &ser1[0]);
+        if (auto ser2 = getSerialization(l2, sc))
+        {
+            //printf("l2 serialization: %s\n", &ser2[0]);
+            if (ser1 == ser2)
+                return true;
+        }
+    }
+    return false;
+}
+
+/**
+ * Computes the string representation of a
  * lambda function described by the subtree starting from a
  * $(REF dmd, func, FuncLiteralDeclaration).
  *
@@ -59,8 +85,26 @@ private enum ExpType
  * variable or a template instance is encountered, the
  * serialization is dropped and the function is considered
  * uncomparable.
+ *
+ * Params:
+ *  fld = the starting AST node for the lambda function
+ *  sc = the scope in which the lambda function is located
+ *
+ * Returns:
+ *  The serielization of `fld`.
  */
-extern (C++) class SerializeVisitor : SemanticTimeTransitiveVisitor
+private string getSerialization(FuncLiteralDeclaration fld, Scope* sc)
+{
+    scope serVisitor = new SerializeVisitor(sc);
+    fld.accept(serVisitor);
+    const len = serVisitor.buf.offset;
+    if (len == 0)
+        return null;
+
+    return cast(string)serVisitor.buf.extractString()[0 .. len];
+}
+
+private extern (C++) class SerializeVisitor : SemanticTimeTransitiveVisitor
 {
 private:
     StringTable arg_hash;
@@ -120,6 +164,10 @@ public:
         if (rs && rs.exp)
         {
             rs.exp.accept(this);
+        }
+        else
+        {
+            buf.offset = 0;
         }
     }
 
@@ -206,6 +254,10 @@ public:
                     d = em;
                     et = ExpType.EnumDecl;
                 }
+                else if (auto fd = s.isFuncDeclaration())
+                {
+                    writeMangledName(fd);
+                }
                 // For anything else, the function is deemed uncomparable
                 else
                 {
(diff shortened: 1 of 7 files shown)


RazvanN7:
CyberShadow/DAutoTest fails because the documentation contains a test which was supposed to fail but now passes. PR to fix documentation : https://github.com/dlang/dlang.org/pull/2231

test/compilable/imports/testlambda2.d
```diff
@@ -0,0 +1,3 @@
+module imports.testlambda2;
+
+int bar() { return 7; }
```
wilzbach: I assume it's not smart enough to do:

```
enum seven = 7;
int bar() { return seven; }
```

yet?

test/compilable/testlambdacomp.d
```diff
@@ -38,7 +39,7 @@ void test2()
     static assert(!__traits(isSame, a => a + b, a => a + b));
 
     int f() { return 3;}
-    static assert(!__traits(isSame, a => a + f(), a => a + f()));
+    static assert(__traits(isSame, a => a + f(), a => a + f()));
```
wilzbach: :+1:

test/compilable/testlambdacomp.d
```diff
@@ -132,10 +133,75 @@ void test4()
...
+    // testlambdacomp.bar != testlambdacomp.test5.bar
+    static assert(!__traits(isSame, pred, (int a) => a + bar()));
+
+    // imports.testlambda1.bar != testlambdacomp.test5.bar
+    import imports.testlambda1 : bar;
+    static assert(!__traits(isSame, pred, (int a) => a + bar()));
+
+    // functions imported from different modules are not equal
+    foobar2!((int a) => a + bar())();
+}
```
wilzbach: Ehm that's hard to read - I suggest the following:

- better names than `foobar` and `foobar2`
- put the test that calls `foobar` etc directly below their declaration (maybe even put the methods in the function itself)

wilzbach: Alternatively, simply don't declare a function which you just call once, but call it e.g. `test6` and do `alias pred = (int a) => a + bar()` in the beginning.

test/compilable/testlambdacomp.d
```diff
@@ -132,10 +133,75 @@ void test4()
...
+    // functions in different scopes
+    foobar!((int a) => a + bar())();
+
+    int foo(int a, int b)
+    {
+        return 2 + a + b;
+    }
+
+    // functions with different kind of parameters
+    alias pred3 = a => a + foo(2, 3);
```
wilzbach: `preda23`

test/compilable/testlambdacomp.d
```diff
@@ -132,10 +133,75 @@ void test4()
...
+    alias pred4 = b => b + foo(2, 3);
```
wilzbach: `predb23` etc.

wilzbach (commented):
A few comments regarding the tests - remember that humans should read them - not machines ;-)

wilzbach:
(anyhow looks generally good and I already pulled your dlang.org PR)

test/compilable/imports/testlambda2.d
RazvanN7: The body of the function which is called is not serialized. When `bar` is called in a lambda the mangled name of the function is put in the serialization + the serialization of the function arguments (the function body is not touched).

RazvanN7:
@wilzbach Thanks for the review. Hope this covers it

src/dmd/lambdacomp.d
```diff
@@ -59,8 +59,26 @@ private enum ExpType
...
+ *
+ * Returns:
+ *  The serielization of `fld`.
  */
-extern (C++) class SerializeVisitor : SemanticTimeTransitiveVisitor
+string getSerialization(FuncLiteralDeclaration fld, Scope* sc)
+{
+    auto serVisitor = new SerializeVisitor(sc);
+    fld.accept(serVisitor);
+    auto len = serVisitor.buf.offset;
```
WalterBright: `const len`, not `auto len`

src/dmd/lambdacomp.d
```diff
@@ -59,8 +59,26 @@ private enum ExpType
...
+ *  fld = the starting AST node for the lambda function
+ *  sc = the scope in which the lambda function is located
+ *
+ * Returns:
+ *  The serielization of `fld`.
  */
-extern (C++) class SerializeVisitor : SemanticTimeTransitiveVisitor
+string getSerialization(FuncLiteralDeclaration fld, Scope* sc)
+{
+    auto serVisitor = new SerializeVisitor(sc);
```
WalterBright: `scope serVisitor`, not `auto serVisitor`. The `scope` will cause it to be allocated on the stack.

src/dmd/lambdacomp.d
```diff
@@ -59,8 +59,26 @@ private enum ExpType
...
+    fld.accept(serVisitor);
+    auto len = serVisitor.buf.offset;
+    if (len == 0)
+        return null;
+    else
+        return serVisitor.buf.peekString()[0 .. len].idup;
```
WalterBright: `return serVisitor.buf.extractString()[0 .. len];`
The extract takes ownership of the buffer, so it does not need to be duplicated.

src/dmd/traits.d
```diff
@@ -1453,28 +1453,16 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
-            //printf("serialization1: %s\n", l1.serialization);
-            //printf("serialization2: %s\n", l2.serialization);
-            if (strcmp(l1.serialization, l2.serialization) == 0 &&
-                strcmp(l1.serialization, "uncomparable") != 0)
-            {
-                return True();
+                //printf("l1 serialization: %s\n", ser1);
+                if (auto ser2 = getSerialization(l2, sc))
+                {
+                    //printf("l2 serielization: %s\n", ser2);
```
WalterBright: serial, not seriel

src/dmd/traits.d
```diff
@@ -1453,28 +1453,16 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+                    if (ser1 == ser2)
+                        return True();
```
WalterBright: Note that if both are incomparable, this returns true. Which is a bug. Which means this particular case needs to be added to the tests.

src/dmd/traits.d
```diff
@@ -1453,28 +1453,16 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+                }
```
WalterBright: Note that the allocated memory returned by `getSerialization` is no longer needed here. This suggests a further encapsulation of the lines 1456 to here, create a new function `isSameFuncLiteral()` in `lambdacomp.d` that returns `true` if they are the same. In that function, put the `OutBuffer` declarations, so the memory is recovered at the function end.

andralex (approved):
great!

PR #7937, opened 2018-02-21
Add [Environment{arch}] DFLAGS to -v output

JinShil:
Followup to #7865

#7865 printed DFLAGS from the `dflags` local variable.  The `dflags` local variable was populated from the `[Environment]` section of the config file, but was never populated from the architecture-specific section of the config file (i.e. `[Environment64]`)

This PR defers reading DFLAGS from the `environment` string table until needed in the `-v` output, by which time the architecture-specific section of the config file has already been parsed.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -242,19 +242,26 @@ private int tryMain(size_t argc, const(char)** argv)
      */
     sections.push("Environment");
     parseConfFile(&environment, global.inifilename, inifilepath, inifile.len, inifile.buffer, &sections);
-    Strings dflags;
-    getenv_setargv(readFromEnv(&environment, "DFLAGS"), &dflags);
-    environment.reset(7); // erase cached environment updates
+
     const(char)* arch = global.params.is64bit ? "64" : "32"; // use default
     arch = parse_arch_arg(&arguments, arch);
-    arch = parse_arch_arg(&dflags, arch);
+
+    // parse architecture from DFLAGS read from [Environment] section
+    {
+        Strings dflags;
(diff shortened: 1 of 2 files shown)


src/dmd/mars.d
```diff
@@ -242,18 +242,25 @@ private int tryMain(size_t argc, const(char)** argv)
...
-    arch = parse_arch_arg(&dflags, arch);
+
+    // parse architecture from DFLAGS read from [Environment] section
+    {
+        Strings dflags;
+        getenv_setargv(readFromEnv(&environment, "DFLAGS"), &dflags);
+        environment.reset(7); // erase cached environment updates
+        arch = parse_arch_arg(&dflags, arch);
+    }
+
```
JinShil: `dflags` is never used after this point, and will actually contain incomplete information after the `[Environment{arch}]` section is parsed, so it is placed in it's own scope.

src/dmd/mars.d
```diff
@@ -242,18 +242,25 @@ private int tryMain(size_t argc, const(char)** argv)
...
     bool is64bit = arch[0] == '6';
 
     version(Windows) // delete LIB entry in [Environment] (necessary for optlink) to allow inheriting environment for MS-COFF
         if (is64bit || strcmp(arch, "32mscoff") == 0)
             environment.update("LIB", 3).ptrvalue = null;
 
+    // read from DFLAGS in [Environment{arch}] section
     char[80] envsection;
```
WalterBright: `= void;`

WalterBright (approved):

src/dmd/mars.d
JinShil: Done

rainers (approved):

PR #7938, opened 2018-02-22
[Refactor] Create Visitor for getProperty method of type AST nodes

RazvanN7:


src/dmd/typesem.d
```diff
@@ -1504,3 +1506,548 @@ static Type merge(Type type)
...
+        else
+        {
+            error(loc, "no property `%s` for tuple `%s`", ident.toChars(), mt.toChars());
+            e = new ErrorExp();
+        }
+        result = e;
+    }
+}
+
+
```
wilzbach: Trailing lines at the end.

wilzbach (approved):
OK, but could we use the opportunity to move this into a separate file. `property.d` maybe?

WalterBright (approved):

src/dmd/typesem.d
RazvanN7: I think that the right place for this is in typesem since getProperty, dotExp and resolve are all part of the semantic phase.

PR #7940, opened 2018-02-22
Fixup C++ headers

kinke:
Based on 2.079.0-beta.1. Many `Loc -> const Loc &` transitions were missing; I may have missed some, as not all `Loc` by-value params transitioned to const refs, so I had to check manually.

### DIFF
--- a/src/dmd/declaration.d
+++ b/src/dmd/declaration.d
@@ -475,7 +475,7 @@ extern (C++) abstract class Declaration : Dsymbol
         return (storage_class & STC.parameter) != 0;
     }
 
-    override final bool isDeprecated() const pure nothrow @nogc @safe
+    override final bool isDeprecated() pure nothrow @nogc @safe
     {
         return (storage_class & STC.deprecated_) != 0;
     }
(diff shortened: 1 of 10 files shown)


src/dmd/declaration.d
```diff
@@ -475,7 +475,7 @@ extern (C++) abstract class Declaration : Dsymbol
         return (storage_class & STC.parameter) != 0;
     }
 
-    override final bool isDeprecated() const pure nothrow @nogc @safe
+    override final bool isDeprecated() pure nothrow @nogc @safe
```
wilzbach: ?

kinke: Yep, the frontend apparently doesn't care about this invalid override: https://github.com/dlang/dmd/blob/master/src/dmd/dsymbol.d#L782

ibuclaw:
Many of these could be added to the C++ frontend test.

src/dmd/declaration.d
ibuclaw: Could the base method have const added instead?

WalterBright: > invalid override

It's a feature, not a bug. It's the miracle of covariance :-)

WalterBright (approved):

PR #7942, opened 2018-02-23
Assume that monitor functions have the nothrow attribute.

FeepingCreature:
Depends on https://github.com/dlang/druntime/pull/2105 and communicates the change to DMD.

### DIFF
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -3484,12 +3484,12 @@ else
                 auto args = new Parameters();
                 args.push(new Parameter(0, ClassDeclaration.object.type, null, null));
 
-                FuncDeclaration fdenter = FuncDeclaration.genCfunc(args, Type.tvoid, Id.monitorenter);
+                FuncDeclaration fdenter = FuncDeclaration.genCfunc(args, Type.tvoid, Id.monitorenter, STC.nothrow_);
                 Expression e = new CallExp(ss.loc, fdenter, new VarExp(ss.loc, tmp));
                 e.type = Type.tvoid; // do not run semantic on e
 
                 cs.push(new ExpStatement(ss.loc, e));
-                FuncDeclaration fdexit = FuncDeclaration.genCfunc(args, Type.tvoid, Id.monitorexit);
+                FuncDeclaration fdexit = FuncDeclaration.genCfunc(args, Type.tvoid, Id.monitorexit, STC.nothrow_);
                 e = new CallExp(ss.loc, fdexit, new VarExp(ss.loc, tmp));
                 e.type = Type.tvoid; // do not run semantic on e
                 Statement s = new ExpStatement(ss.loc, e);


FeepingCreature:
Er, hang on, no. DRuntime depends on *this* change.

wilzbach:
Re this Jenkins failure:

```
../../.dub/packages/libasync-0.8.3/libasync/source/libasync/dns.d(120,3): Warning: statement is not reachable
../../.dub/packages/libasync-0.8.3/libasync/source/libasync/file.d(73,3): Warning: statement is not reachable
dmd failed with exit code 1.
```

Could be something permanent. I restarted.

mathias-lang-sociomantic:
Last time we tried this it blew up in Vibe.d, big time.
I would really like to have @s-ludwig 's word before we merge anything.

mathias-lang-sociomantic:
Also, little to no rationale, and no changelog entry. This is a big change, please treat it as such.
Here's what happen last time it landed (v2.067): https://github.com/vibe-d/vibe.d/pull/972

FeepingCreature:
I didn't realize this would be controversial, or actually that people were using non-nothrow monitors at all.

If this is an issue, the more annoyingly complex solution would be maybe a sub-interface of Monitor that has nothrow? So that the stock monitor could inherit from it, and at least synchronized (this) would no longer violate nothrow.

wilzbach:
> Could be something permanent. I restarted.

Yes, looks like the libasync failure is genuine and introduced by this PR:

image

FeepingCreature:
In that case, I'd probably retract this pr and file a bug for it.

It's kind of ridiculous that D ostensibly has all these nice features like nothrow, but you can't use synchronized(this) with it even though the default monitor evidently does not throw. You end up deciding to use a feature, then a few days later you have to undo the feature because it doesn't work with some other feature and nobody gave any thought to the combination.

FeepingCreature:
Oh, it seems it was filed in 2015...

https://issues.dlang.org/show_bug.cgi?id=11216

s-ludwig:
vibe.d should be fine. After the last attempt, I split up the `TaskMutex` types into one that can throw and another one that is `nothrow`. Only the latter derives from `Mutex` to be able to act as a replacement.

FeepingCreature:
I don't know DMD. Is there any way to ensure that Monitor has nothrow on lock/unlock before setting those attributes on the _d_monitorenter/leave externs we generate?

WalterBright (approved):

s-ludwig:
The failure appears to be directly in libasync, which is optionally used by vibe.d. All other configurations are fine, though, and the fix should be trivial.

FeepingCreature:
I am increasingly not confident in this pr. The proper solution would really be to find some way to ensure that the monitor lock/unlock functions actually do have nothrow. Is that possible somehow? Otherwise I guess I'll reopen the associated druntime pr.

ibuclaw:
The path of least resistance may be to first deprecate if you are able to detect the wrong use case before we add nothrow.

thewilsonator:
This sound like something that should be solved by proto-object. Closing

PR #7946, opened 2018-02-23
fix issue 18510 - lld-link.exe fails to open obj file in subpath

rainers:
LLD doesn't assume .obj automatically for files with `\` in the name, but there is no good reason to strip the extension to begin with

### DIFF
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -161,22 +161,12 @@ public int runLINK()
         if (global.params.mscoff)
         {
             OutBuffer cmdbuf;
-            cmdbuf.writestring("/NOLOGO ");
+            cmdbuf.writestring("/NOLOGO");
             for (size_t i = 0; i < global.params.objfiles.dim; i++)
             {
-                if (i)
-                    cmdbuf.writeByte(' ');
+                cmdbuf.writeByte(' ');
                 const(char)* p = global.params.objfiles[i];
-                const(char)* basename = FileName.removeExt(FileName.name(p));
-                const(char)* ext = FileName.ext(p);
-                if (ext && !strchr(basename, '.'))
-                {
-                    // Write name sans extension (but not if a double extension)
-                    writeFilename(&cmdbuf, p, ext - p - 1);
-                }
-                else
-                    writeFilename(&cmdbuf, p);
-                FileName.free(basename);
+                writeFilename(&cmdbuf, p);
             }
             if (global.params.resfile)
             {


WalterBright:
> no good reason

Command line length maximums can get hit. Microsoft tends to increase those limits with every new version of Windows, but there is one. I ran into it just the other day. The symptom is it just truncates it. Arggh.

rainers:
> Command line length maximums can get hit.

We are writing the command line into a response file anyway at a certain threshold (7000 characters ATM).

wilzbach (approved):
Is there an easy way to add tests for this?
Maybe a Windows-only bash test that greps the `-v` output?

rainers:
> Is there an easy way to add tests for this?

All the tests against the COFF object format show it didn't break anything.
It might be nice to add an lld-link test, but we would have to add it to the autotesters (or download it before testing). For full coverage, it would have to uninstall VS first, though.

src/dmd/link.d
```diff
@@ -161,22 +161,12 @@ public int runLINK()
...
             {
-                if (i)
-                    cmdbuf.writeByte(' ');
+                cmdbuf.writeByte(' ');
                 const(char)* p = global.params.objfiles[i];
-                const(char)* basename = FileName.removeExt(FileName.name(p));
-                const(char)* ext = FileName.ext(p);
-                if (ext && !strchr(basename, '.'))
-                {
-                    // Write name sans extension (but not if a double extension)
```
marler8997: Do we know why this was being done in the first place?

marler8997:
Do we have a test that verifies we use a response file when command line hits the limit?

rainers:
> Do we have a test that verifies we use a response file when command line hits the limit?

The druntime unittest does that, the list of source files in mak/SRCS exceeds 11k, so about the same size when changing the extension.

> Actually, does lld-link support response files as well?

It does, it's meant as a replacement for link.exe.

marler8997:
Great.  Do we know what the exact limit is?  Maybe we should add a test that verifies one under and one over the limit both work?

marler8997:
>> LLD doesn't assume .obj automatically for files with \ in the name
>...it's meant as a replacement for link.exe.

It sounds like this is also demonstrating a bug in lld-link?  If `link.exe` assumes .obj automatically for files with `\` in the name, shouldn't `lld-link`?  Not saying we should integrate this PR, but do you know how to file a bug for `lld-link`?  I can look into this if you don't know.

rainers:
> Do we know what the exact limit is?

For CreateProcess, it's 32768, but if it goes through a shell command it's 8191 minus what the shell needs.

> Maybe we should add a test that verifies "one under" and "one at" the limit both work?

We've had the pretty large limit 7000 since forever, testing it seems excessive.

marler8997:
Oh I see, we just opt to use a response file way before the limit...ok sounds good

marler8997 (approved):

rainers:
> It sounds like this is also demonstrating a bug in lld-link?

Yes, I've filed one: https://bugs.llvm.org/show_bug.cgi?id=36502

PR #7948, opened 2018-02-24
build gen_man with the same model as dmd

braddr:
freebsd 11.1 32 is failing due to what looks like crossing 32 and 64 bit builds.


braddr:
Nope, not the issue.

PR #7952, opened 2018-02-24
Temporary workaround for Issue 16243

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=16243

In order to get the autotester passing again.


braddr:
Force pulled.  This disables a failing freebsd/32 test, but there's others in the way still so it won't pass cleanly yet.

PR #7954, opened 2018-02-24
cleaner fix for Issue 18480 - dmd hangs with self-alias declaration

MartinNowak:
- fixed by detecting `alias sym = sym;` during semantic
- small adoption of assertion in mostVisibleOverload to deal with
  alias error expression (type = Type.terror) in overload

see #7930 for previous workaround


MartinNowak:
FYI @timotheecour and @RazvanN7, also see https://github.com/dlang/dmd/pull/7930#issuecomment-368267710.

JinShil (approved):

JinShil:
Thanks, @MartinNowak

ibuclaw (approved):

PR #7958, opened 2018-02-25
[dmd-cxx] Fix warnings when building with all enabled in GCC

ibuclaw:
In GDC tree, the default compiler warnings are:
```
-W -Wall \
-Wno-narrowing \
-Wwrite-strings \
-Wcast-qual \
-Wno-long-long \
-Wno-variadic-macros \
-Wno-overlength-strings \
-fno-common
```
And no complaints come from the compiler.

In DMD tree, building with `ENABLE_WARNINGS=1` is now mostly silent, save for the backend and other non-shared frontend sources.


src/ctfeexpr.c
```diff
@@ -1562,7 +1560,7 @@ Expression *ctfeIndex(Loc loc, Type *type, Expression *e1, uinteger_t indx)
         StringExp *es1 = (StringExp *)e1;
         if (indx >= es1->len)
         {
-            error(loc, "string index %llu is out of bounds [0 .. %llu]", indx, (ulonglong)es1->len);
+            error(loc, "string index %llu is out of bounds [0 .. %llu]", (ulonglong)indx, (ulonglong)es1->len);
             return CTFEExp::cantexp;
```
dkgroot: @ibuclaw 
For size_t related elements it is better to use %zu / %zd etc.
ie:
```
size_t x = ...;
ssize_t y = ...;
printf("%zu\n", x);  // prints as unsigned decimal
printf("%zx\n", x);  // prints as hex
printf("%zd\n", y);  // prints as signed decimal
```
These will adapt automatically to the platform dependent size of size_t.

dkgroot (changes requested):
@ibuclaw Thanks a lot for these changes !

src/ctfeexpr.c
dkgroot: @ibuclaw: I did the same kind of fixes you did (using %zd in my case), just moments before you posted this PR. If you like i can attach the diff here.

wilzbach (approved):
Whenever you are ready.
Do you want to add the appropriate flags to `posix.mak`?
(For `dmd-cxx` being bug fixes only this is quite an effort though)

src/ctfeexpr.c
ibuclaw: @dkgroot - Lets just stick to `%l` or `%ll`.  The explicit casts are just a formality anyway.

src/expression.c
```diff
@@ -1928,7 +1888,7 @@ bool functionParameters(Loc loc, Scope *sc, TypeFunction *tf,
 
                 /* Declare temporary 'auto __pfx = arg' (needsDtor) or 'auto __pfy = arg' (!needsDtor)
                  */
-                VarDeclaration *tmp = tmp = copyToTemp(0,
+                VarDeclaration *tmp = copyToTemp(0,
```
ibuclaw: The typo's that warnings found. :-)

ibuclaw:
> (For dmd-cxx being bug fixes only this is quite an effort though)

Bug fixes slash gdc mirror. ;-)

src/ctfeexpr.c
dkgroot: Ok, it was just a suggestion.
Getting those compiler warnings silenced is the main objective

ibuclaw:
@wilzbach - Added warnings as per 18126.  There's a slight tightening of warnings for the frontend sources, and a loosening of the glue sources, that is loosened further with the backend.

Also added `ENABLE_WERROR` and `ENABLE_WFORMAT`.

Tested locally using `ENABLE_WERROR=1` and `ENABLE_WARNINGS=1 ENABLE_WERROR=1` with both g++ and clang++ as the set compiler.

ibuclaw:
@wilzbach - Shall we merge and move on?

PetarKirov (approved):

PR #7959, opened 2018-02-26
Fix Issue 17373 -  traits getOverloads + multiple interface inheritan…

RazvanN7:
…ce only see one of the interfaces' overloads

Interface inheritance does not create an OverloadSet for overloaded methods in different inherited interfaces. This caused traits(getOverloads) to report incorrect answers. In order to fix this I added a test so that if the symbol passed to traits is a interface declaration, all the inherited interfaces are checked to see if they add a new overload. In order to avoid counting functions with the same signature, but located in different inherited interfaces, I am using a hashtable which stores the function mangled signature.

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -923,7 +923,33 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
                     ex = dve.e1;
             }
 
-            overloadApply(f, (Dsymbol s)
+            bool[string] funcTypeHash;
+
+            /* Compute the function signature and insert it in the
+             * hashtable, if not present. This is needed so that
+             * traits(getOverlods, F3, "visit") does not count `int visit(int)`
+             * twice in the following example:
+             *
+             * =============================================
+             * interface F1 { int visit(int);}
+             * interface F2 { int visit(int); void visit(); }
+             * interface F3 : F2, F1 {}
+             *==============================================
+             */
+            void insertInterfaceInheritedFunction(FuncDeclaration fd, Expression e)
+            {
+                auto funcType = fd.type.toChars();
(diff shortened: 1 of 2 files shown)


src/dmd/traits.d
```diff
@@ -923,7 +923,34 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+             * interface F3 : F2, F1 {}
+             *==============================================
+             */
+            void insertInterfaceInheritedFunction(FuncDeclaration fd, Expression e)
+            {
+                bool yes = true;
+                auto funcType = fd.type.toChars();
+                auto len = strlen(funcType);
+                //printf("%s - %s\n", fd.toChars, fd.type.toChars());
+                if (!funcTypeHash.lookup(funcType, len))
```
wilzbach: 1) Why don't you use D strings and a normal hashmap?
2) `toChars` is a bit of an anti-pattern. Why don't you just insert the `type` directly?

RazvanN7: Type is a class that does not have `opEquals` and `toHash` defined.

JinShil:
@RazvanN7 Please look into the Jenkins failures.

RazvanN7:
@JinShil Jenkins is green now.

JinShil (approved):

wilzbach:
This PR broke Vibe.d's complex `InterfaceProxy` generation and thus breaks Vibe.d in general.

https://github.com/vibe-d/vibe.d/issues/2181

It's a shame no one noticed this before 2.081 (Vibe.d was disabled on the project tester due to issues with the dub registry at the time of this PR).

wilzbach:
See also: https://issues.dlang.org/show_bug.cgi?id=19064

PR #7961, opened 2018-02-27
[changelog] Argument mismatch errors

ntrel:
Also tweak some other changelog formatting.

### DIFF
--- a/changelog/argument-mismatch.dd
+++ b/changelog/argument-mismatch.dd
@@ -1,0 +1,22 @@
+Argument mismatch errors have been improved
+
+$(P `dmd` now shows which argument failed to match a parameter and why.)
+---
+void fill(ref char[16] buf, char c);
+
+void main()
+{
+    fill("1234567890123456", '*');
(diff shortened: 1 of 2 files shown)


wilzbach:
image

Maybe a shorter example would have been better to prevent the wraps?

changelog/argument-mismatch.dd
```diff
@@ -0,0 +1,22 @@
+Argument mismatch errors have been improved
+
+$(P `dmd` now shows which argument failed to match a parameter with an explanatory error message.)
```
wilzbach: FYI: this `P` isn't necessary for changelog entries. The wrapper is smart enough to add them automatically for you to provide a "Markdownesque experience".

wilzbach (approved):
In the interest of time and our CI problems at the moment, I added a small nit to your changelog entry and I'm now directly going ahead and merge this as 2.079 is coming up very soon.
Thanks a lot for the follow-up!

PR #7963, opened 2018-02-28
fix Issue 16243 - wrong C++ argument passing with empty struct when i…

WalterBright:
…nterfacing with Clang

### DIFF
--- a/src/dmd/backend/cgcod.c
+++ b/src/dmd/backend/cgcod.c
@@ -9,7 +9,7 @@
  * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/cgcod.c, backend/cgcod.c)
  */
 
-#if !SPP
+#if (SCPP && !HTOD) || MARS
 
 #include        <stdio.h>
 #include        <string.h>
@@ -1128,25 +1128,32 @@ void stackoffsets(int flags)
     for (int si = 0; si < globsym.top; si++)
     {   symbol *s = globsym.tab[si];
 
-        if (s->Sisdead(anyiasm))
+        /* Don't allocate space for dead or zero size parameters
+         */
+        switch (s->Sclass)
         {
-            /* The variable is dead. Don't allocate space for it if we don't
-             * need to.
-             */
-            switch (s->Sclass)
-            {
-                case SCfastpar:
-                    if (!(funcsym_p->Sfunc->Fflags3 & Ffakeeh))
-                        continue;   // don't need consistent stack frame
-                    break;
+            case SCfastpar:
+                if (!(funcsym_p->Sfunc->Fflags3 & Ffakeeh))
+                    goto Ldefault;   // don't need consistent stack frame
+                break;
 
-                case SCshadowreg:
-                case SCparameter:
-                    break;          // have to allocate space for parameters
+            case SCparameter:
+                if (type_zeroSize(s->Stype, tybasic(funcsym_p->Stype->Tty)))
+                {
+                    Para.offset = _align(REGSIZE,Para.offset); // align on word stack boundary
+                    s->Soffset = Para.offset;
+                    continue;
+                }
+                break;          // allocate even if it's dead
 
-                default:
+            case SCshadowreg:
+                break;          // allocate even if it's dead
+
+            default:
+            Ldefault:
+                if (s->Sisdead(anyiasm))
                     continue;       // don't allocate space
-            }
+                break;
         }
 
         targ_size_t sz = type_size(s->Stype);
--- a/src/dmd/backend/type.c
+++ b/src/dmd/backend/type.c
@@ -209,10 +209,63 @@ L1:
     return sz;
 }
 
+/***********************************
+ * Compute special zero sized struct.
+ * Params:
+ *      t = type of parameter
+ *      tyf = function type
+ * Returns:
+ *      true if it is
+ */
+bool type_zeroSize(type *t, tym_t tyf)
+{
+    if (tyf != TYjfunc && config.exe & (EX_FREEBSD | EX_OSX))
+    {
+        /* Use clang convention for 0 size structs
+         */
+        if (t && tybasic(t->Tty) == TYstruct)
+        {
+            type *ts = t->Ttag->Stype;     // find main instance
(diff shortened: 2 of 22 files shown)


src/dmd/target.d
```diff
@@ -728,6 +728,36 @@ struct Target
...
+    /***
+     * Determine the size a value of type `t` will be when it
+     * is passed on the function parameter stack.
+     * Params:
+     *  loc = location to use for error messages
+     *  t = type of parameter
+     * Returns:
+     *  size used on parameter stack
+     */
+    extern (C++) ulong parameterSize(const ref Loc loc, Type t)
```
ibuclaw: Add entry to target.h

src/dmd/target.d
```diff
@@ -728,6 +728,36 @@ struct Target
...
+    {
+        if (!global.params.is64bit &&
+            (global.params.isFreeBSD || global.params.isOSX))
+        {
+            /* These platforms use clang, which regards a struct
+             * with size 0 as being of size 0 on the parameter stack,
+             * even while sizeof(struct) is 1.
+             * It's an ABI incompatibility with gcc.
+             */
+            if (t.ty == Tstruct)
```
ibuclaw: Are you sure this is correct for both C and C++?

I ask because gcc and g++ has a discrepancy between empty aggregates. One reports sizeof == 0 while the other reports 1.

src/dmd/backend/cgcod.c
```diff
@@ -1128,25 +1128,32 @@ void stackoffsets(int flags)
...
 
-                case SCshadowreg:
-                case SCparameter:
-                    break;          // have to allocate space for parameters
+            case SCparameter:
+                if (type_parameterSize(s->Stype) == 0)
+                {
+                    Para.offset = _align(REGSIZE,Para.offset); // align on word stack boundary
+                    s->Soffset = Para.offset;
+                    continue;
```
ibuclaw: So how does passing around an empty aggregate change here?  I'd expect the equivalent of replacing `foo(s)` with `foo({})`.
That is to say that empty aggregate vars and params aren't passed around directly - having nothing themselves to pass around - they are just an alias for a field-less struct literal.

ibuclaw (changes requested):
Looks fine except for comments.

src/dmd/backend/cgcod.c
WalterBright: I did some experiments. You can take the address of a parameter that is an empty struct. It is the start of the next parameter. So it definitely has an offset.

src/dmd/target.d
WalterBright: You're right, gcc and g++ do behave differently. gcc, clang, clang++ have 0 sized structs on the parameter stack taking up no space, g++ they take up space.

WalterBright: With g++, `extern "C" { ... }` has no effect on the empty struct size, it stays at 1. So g++ cannot interface with code compiled with gcc. (!)

WalterBright: Which leads us to a hopeless problem - we cannot be compatible with both g++ in `extern "C"` and with gcc.

stefan-koch-sociomantic: clang++ should give you a non-zero sized struct in cpp code.
If it does not it has a bug.

src/dmd/backend/type.c
```diff
@@ -209,10 +209,62 @@ unsigned type_alignsize(type *t)
...
+ * Params:
+ *      t = type of parameter
+ * Returns:
+ *      true if it is
+ */
+bool type_zeroSize(type *t)
+{
+    if (config.exe & (EX_FREEBSD | EX_OSX))
+    {
+        /* Use clang convention for 0 size structs
```
jacob-carlborg: Isn't the auto-tester using GCC?

src/dmd/target.d
ibuclaw: > You're right, gcc and g++ do behave differently. gcc, clang, clang++ have 0 sized structs on the parameter stack taking up no space, g++ they take up space.

As per Stefan's comment.  The distinction here is C vs. C++.  In C empty structs have no size, but in C++, the standard dictates that an empty struct is equivalent to `struct S { char pad; }`.

Citation:
https://gcc.gnu.org/onlinedocs/gcc-7.3.0/gcc/Empty-Structures.html#Empty-Structures
http://www.stroustrup.com/bs_faq2.html#sizeof-empty

ibuclaw: Now, as per 14.1.3 of the spec (https://dlang.org/spec/struct.html#struct_layout)

> The default layout of the fields of a struct is an exact match with the associated C compiler.

Ignoring the phrasing of the word "associated C compiler" and taking that to just mean "D structs are C ABI compatible".

That means that all structs should have a size `0`, therefore we can never be compatible with C++.  But you can however define `struct MyCppStruct { char pad; }` to mimic the equivalent behaviour.

src/dmd/backend/type.c
wilzbach: AFAICT is clang used by FreeBSD:

https://github.com/braddr/d-tester/issues/70#issuecomment-368116189
https://github.com/braddr/d-tester/issues/70#issuecomment-368199964

src/dmd/target.d
WalterBright: The clang++ equivalent behavior cannot be mimicked by adding a dummy field. clang++ returns sizeof==1 for empty structs, but they use 0 size on the parameter stack.

WalterBright: The whole thing is a mess.

src/dmd/backend/type.c
jacob-carlborg: What about macOS?

src/dmd/target.d
ibuclaw: Or at least perhaps gdc will be g++ compatible, ldc clang++ compatible. You're just stuck in the middle.

src/dmd/target.d
WalterBright: done

WalterBright:
ping @ibuclaw

ibuclaw (approved):
:see_no_evil:

PR #7964, opened 2018-02-28
testsuite: avoid rebuilding all tests every time

rainers:
Might just be my version of make, but the testsuite has recently lost its make functionality, it rebuilds everything with each invocation. Introduced by 24d9d18f75ba6ae6ea17e6da6d00573264a2e7de


ibuclaw (approved):
No you are right. The recipe test_tools should be a file that exists (a stamp, if you would call it that) to not rebuild the other recipes it depends on.

wilzbach (approved):

PR #7966, opened 2018-02-28
make testclidflags independent of sc.ini/dmd.conf

rainers:


PetarKirov (approved):

JinShil:
This was also done in #7937.  We could just merge that.

rainers:
Ok, closing in favor #7937

wilzbach:
I wonder why this is necessary though. `dmd.conf` is auto generated on Posix ...

rainers:
> I wonder why this is necessary though. dmd.conf is auto generated on Posix ...

That doesn't happen on Windows AFAICT. My build script copies the compiler and the libraries elsewhere anyway so I can use it for other stuff besides testing, too.

PR #7968, opened 2018-02-28
appveyor: build test suite for coff32

rainers:


wilzbach:
Looks like it's definitely worthwhile to do this:

```
Test runnable/test17559.d failed.  The logged output:
..\generated\Windows\Release\Win32\dmd.exe -conf= -m32mscoff -Irunnable -g    -odtest_m32mscoff\runnable -oftest_m32mscoff\runnable\test17559_0.exe  runnable\test17559.d  -map nul.map
test_m32mscoff\runnable\test17559_0.exe
object.Exception@runnable\test17559.d(48): fun
----------------
0x003DADEE in test17559 at C:\projects\dmd\test\runnable\test17559.d(48)
0x003DAD3B in D at C:\projects\dmd\test\runnable\test17559.d(32)
0x003DBE67 in void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll().__lambda1()
0x003DBDA9 in void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll()
0x003DBB68 in _d_run_main
0x003DB2BC in __entrypoint
0x0040CC8B in __scrt_common_main_seh at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl(253)
0x76E77C04 in BaseThreadInitThunk
0x77D5AD8F in RtlInitializeExceptionChain
0x77D5AD5A in RtlInitializeExceptionChain
core.exception.AssertError@runnable\test17559.d(55): Assertion failure
----------------
0x003DB463 in _d_assertp
0x003DAD3B in D at C:\projects\dmd\test\runnable\test17559.d(32)
0x003DBE67 in void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll().__lambda1()
0x003DBDA9 in void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll()
0x003DBB68 in _d_run_main
0x003DB2BC in __entrypoint
0x0040CC8B in __scrt_common_main_seh at f:\dd\vctools\crt\vcstartup\src\startup\exe_common.inl(253)
0x76E77C04 in BaseThreadInitThunk
0x77D5AD8F in RtlInitializeExceptionChain
0x77D5AD5A in RtlInitializeExceptionChain
```

skl131313:
Does the VisualD build for win32 use `-m32` or `-m32mscoff`?

rainers:
> Does the VisualD build for win32 use -m32 or -m32mscoff?

It links with a VC built backend, so it has to use -m32mscoff.

JinShil (approved):

test/runnable/test15779.d
```diff
@@ -18,5 +18,6 @@ int main()
 
 void bar()
 {
+    version(CRuntime_Microsoft) {} else // TODO: fails for -O -inline, also disabled for the auto-tester
```
JinShil: What's the problem here?  What's the plan for eventually enabling it for the auto tester?  Is a bug report filed for it?

rainers: Not sure, maybe related: https://issues.dlang.org/show_bug.cgi?id=13821 and https://issues.dlang.org/show_bug.cgi?id=16511.
I have added a new one: https://issues.dlang.org/show_bug.cgi?id=18547

JinShil: In that case I recommend adding links to those issues in the comments and come clean about you're understanding of what's going on.  Something like:

"Fibers appear to be broken in Windows.  We don't yet have a solution, see these issues...."  or something like that.  My motivations is to ensure that there is some kind of "paper trail" that one can follow to eventually solve the issue.

skl131313:
Using `DISABLED: win32` in the test settings at the top would help for that a bit, so that the test output doesn't show it's successful rather than being skipped.

test/runnable/test15779.d
WalterBright: I agree with the "paper trail" proposal, otherwise such comments just mystify the reader.

ZombineDev: @rainers my guess is that this test passes on LDC. Can you scope this to CRuntime_Microsoft and DigitalMars (i.e. dmd)?

rainers: I think I figured it out: https://github.com/dlang/druntime/pull/2129. All windows builds should be affected, but happens spuriously.

rainers: I have removed the workaround in the hope that the bugfix will be accepted.

ZombineDev: Thanks!

rainers:
Appveyor even passed without https://github.com/dlang/druntime/pull/2129, but I guess that can change spuriously (but also for the Win64 check on Appveyor or the Windows checks on the auto tester).

JinShil:
Restarted Semaphore CI.  All CIs are passing now.

PetarKirov (approved):

PR #7969, opened 2018-02-28
Temporarily disable #7577, s.t. a proper deprecation cycle can be started

wilzbach:
Disables https://github.com/dlang/dmd/pull/7577 as it looks like a proper deprecation cycle is needed and 2.079 is right around the corner.

### DIFF
--- a/src/dmd/semantic2.d
+++ b/src/dmd/semantic2.d
@@ -352,6 +352,10 @@ private extern(C++) final class Semantic2Visitor : Visitor
         fd.semanticRun = PASS.semantic2;
 
         //printf("FuncDeclaration::semantic2 [%s] fd0 = %s %s\n", loc.toChars(), toChars(), type.toChars());
+
+        // https://issues.dlang.org/show_bug.cgi?id=18385
+        // Disable for 2.079, s.t. a deprecation cycle can be started with 2.080
+        if (0)
         if (fd.overnext && !fd.errors)
         {
             OutBuffer buf1;
--- a/test/fail_compilation/fail17492.d
+++ b/test/fail_compilation/fail17492.d
@@ -1,4 +1,8 @@
-/* TEST_OUTPUT:
+/*
+https://issues.dlang.org/show_bug.cgi?id=18385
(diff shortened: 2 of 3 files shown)


JinShil:
I support "revert first, talk (in-depth) later".  We might want to add DWT to the CI, after this is done so future fix can be tested.

JinShil:
It appears `DISABLED` isn't working for `win`.

timotheecour:
@wilzbach if we're not deprecating, can we at least print something (at least as option) when duplicate functions are seen by compiler? this helps catch bugs

wilzbach:
> It appears DISABLED isn't working for win.

Yeah looks like all other platform use e.g. `linux` for `OS`, but Windows sets `win32` and `win64` for `OS` (please don't ask me why!).
That means that this fails:

```
if (testArgs.disabledPlatforms.canFind(envData.os, envData.os ~ envData.model))
```

The problem is that the variadic is the wrong way around - it should be sth. like `[envData.os, envData.os ~ envData.model].canFind(testArgs)`. I will submit this as separate PR though.

wilzbach:
The Windows failure on auto-tester is unrelated - https://github.com/braddr/d-tester/issues/71

wilzbach:
Going to merge this now as the 2.079 release is approaching soon and we want to avoid releasing any potential regressions and this change seems like it needs more time and discussions.
For the record: the auto-tester failures are unrelated and due to https://github.com/braddr/d-tester/issues/71 and https://github.com/braddr/d-tester/issues/70 and FreeBSD 32-64 is just randomly failing on opening a file in druntime...

jacob-carlborg:
> We might want to add DWT to the CI, after this is done so future fix can be tested

I’ve been planning to do that. But I think there are some other depreciations that I thought it would be good to fix first.

jacob-carlborg:
@JinShil If we can get this merged into master as well I can added DWT to the CI. I've already have an open PR https://github.com/dlang/ci/pull/183.

JinShil:
Cherry-pick to master: #7974

PR #7971, opened 2018-02-28
 Fix Issue 18516: Add -vnrvo switch 

ibuclaw:
And replace those fragile pointer comparison tests with verbatim output tests.

### DIFF
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -566,6 +566,9 @@ dmd -cov -unittest myprog.d
         Option("vgc",
             "list all gc allocations including hidden ones"
         ),
+        Option("vnrvo",
+            "list all functions that return using named return value optimization"
+        ),
         Option("vtls",
             "list all variables going into thread local storage"
         ),
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -100,6 +100,7 @@ struct Param
     bool vgc;               // identify gc usage
     bool vfield;            // identify non-mutable field variables
     bool vcomplex;          // identify complex/imaginary type usage
+    bool vnrvo;             // identify calls returning with nrvo
     ubyte symdebug;         // insert debug symbolic information
     bool symdebugref;       // insert debug information for all referenced types, too
     bool alwaysframe;       // always emit standard stack frame
--- a/src/dmd/globals.h
+++ b/src/dmd/globals.h
@@ -77,6 +77,7 @@ struct Param
     bool vgc;           // identify gc usage
     bool vfield;        // identify non-mutable field variables
     bool vcomplex;      // identify complex/imaginary type usage
+    bool vnrvo;         // identify calls returning with nrvo
     char symdebug;      // insert debug symbolic information
     bool symdebugref;   // insert debug information for all referenced types, too
     bool alwaysframe;   // always emit standard stack frame
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1695,6 +1695,8 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 params.showColumns = true;
             else if (arg == "-vgc") // https://dlang.org/dmd.html#switch-vgc
                 params.vgc = true;
+            else if (arg == "-vnrvo") // https://dlang.org/dmd.html#switch-vnrvo
+                params.vnrvo = true;
             else if (startsWith(p + 1, "verrors")) // https://dlang.org/dmd.html#switch-verrors
             {
                 if (p[8] == '=' && isdigit(cast(char)p[9]))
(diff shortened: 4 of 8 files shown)


src/dmd/cli.d
```diff
@@ -566,6 +566,9 @@ dmd -cov -unittest myprog.d
         Option("vgc",
             "list all gc allocations including hidden ones"
         ),
+        Option("vnrvo",
+            "list all functions that return using named return value optimization"
+        ),
```
wilzbach: If we expose it publicly, we should probably add it a full changelog entry for it?

wilzbach (approved):
I like this. Looks a lot cleaner than before!

src/dmd/cli.d
ibuclaw: Probably, I also put in a link that doesn't exist yet either.

WalterBright:
I don't see where the "verbatim output tests" are?

ibuclaw:
@WalterBright search for `TEST_OUTPUT` in the testnrvo.d file.

WalterBright:
@ibuclaw ok I see it.

This PR looks good except for one thing - adding a new command line switch. We seem to be adding them like rabbits. This one appears to have no value to anyone but the compiler tester, and yet it's going to take up space in any command line description, and contributes to a daunting list of them. This list of switches is off-putting to anyone starting with D, and also doubles the number of tests needed to be run if one is looking at it from a combinatorial perspective.

I suggest instead making it a `__traits(isNRVO, functionName)`, which:

1. moves it to a rarely read section of the spec
2. makes the test program self-contained
3. makes it available for introspection by future metaprogrammers

The idea is analogous to why I proposed `__traits(isReturnOnStack, functionName)`.

ibuclaw:
A few problems having it as a trait:
- Needs full semantic, possibly of both nested and parent functions also.
- Closures may turn off the optimization, and we don't know until the codegen pass if one will be created.
- Because struct literals can be returned this way, the only feasible implementation is walking the body to pull out and analyse all return statements.
- Became of this, may want to cache the results for multiple invocations, but cache might be invalidated by closures, or other reasons the target may decide during codegen.
- The value this has for introspection is low, as there is no semantic guarantee about nrvo in the language. And in codegen it generates a move assigned rather than copy to/from temporary.

Initially I had wanted to underline all calls that returned this way, however there is no way to distinguish between RVO and NRVO. I'm not sure of the distinction either in dmd, but perhaps noting any calls that return via hidden pointer may be more useful.

Again though, we can't introspect if a function call returns by hidden pointer until the codegen stage when the hidden slot is set up.

ibuclaw:
> The idea is analogous to why I proposed __traits(isReturnOnStack, functionName).

I'm not even sure if that can be reliably done also. A backend optimizer could decide during finalising codegen that a nested function that doesn't escape no longer has an ABI and so can return in whatever way it deems the cheapest.

Not a problem for dmd though where language and target are meshed together.

andralex:
I really think we must do something to stop inflation of command line switches. If we need to add one for every small thing we can think of listing, then we've lost.

andralex (changes requested):
Asking for changes - can this please be done without a new compiler switch?

jacob-carlborg:
Can we do something like Clang (and I think GCC) is doing. That is, to have a driver and a frontend. The frontend has significantly more flags available than the driver.

ibuclaw:
Actually what gcc and probably clang do are dumping a specific ast pass and scanning for a special pattern on the line it expects it.

WalterBright:
> Needs full semantic, possibly of both nested and parent functions also.

Only of the function itself and its nested functions, though the latter are naturally done to do semantic of the function anyway.

> Closures may turn off the optimization

They may indeed, and this is calculated in toir.d. But I think this can be moved the semantic phase. Ditto for where it is also turned off in glue.d. (Such moving should be done in separate PRs.)

With this, there are these states:

1. does nrvo
2. does not do nrvo
3. may do nrvo

which may be good enough for test porpoises.

ibuclaw:
> which may be good enough for test porpoises

@WalterBright - and about returning true for all functions that returns via hidden pointer (both RVO and NRVO).

I ask because you only need to analyse the return type to detect RVO, rather than walk the function body and make adjustments for closures to determine NRVO.

WalterBright:
I don't understand what RVO has to do with this PR?

ibuclaw:
@WalterBright - Because it wouldn't be expensive to just determine if a function returns by hidden pointer, compared to whether that hidden pointer is masking a named variable.

Never mind then, will do it the hard way.

WalterBright:
> Because it would be inexpensive to just determine if a function returns by hidden pointer, compared to whether that hidden pointer is masking a named variable.

There already is `__traits(isReturnOnStack, ...)`

wilzbach:
> I really think we must do something to stop inflation of command line switches. If we need to add one for every small thing we can think of listing, then we've lost.

Not sure whether it's better to have an inflation of `__traits`. Command line switches can be private, excluded from the docs and even excluded from the release build.

PR #7972, opened 2018-03-01
Improve d_do_test DISABLED to allow true submatches

wilzbach:
Allows to use `win` for `DISABLED` as it now searches for truw submatches (what was probably intended in the first place).

https://github.com/dlang/dmd/pull/7969#issuecomment-369425112

I'm thinking about supporting `all` too. Anyone is favor or against `all`?


JinShil (approved):

JinShil:
This needs to now go into stable.

PR #7973, opened 2018-03-01
Allow getting type from template declarations.

skl131313:
```D
struct TypeObj
{
    alias This = typeof(this);

    mixin template MixinTempl()
    {
        int value;
    }
}

ref TypeObj Obj()
{
    static TypeObj a;
    return a;
}

void main()
{
    mixin Obj.This.MixinTempl; // ok
    mixin Obj.MixinTempl;      // error: "MixinTempl!()" is not defined
}

```

Couldn't access templates like in the above example, though other symbols such as types are accessible. This allows the second line to be usable.

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -2798,6 +2798,10 @@ extern (C++) abstract class Type : RootObject
                 s = getDsymbol(e);
                 break;
 
+            case TOK.dotTemplateDeclaration:
+                s = (cast(DotTemplateExp)e).td;
+                break;
+
             //case TOK.this_:
             //case TOK.super_:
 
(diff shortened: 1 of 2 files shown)


src/dmd/mtype.d
```diff
@@ -2798,6 +2798,10 @@ extern (C++) abstract class Type : RootObject
                 s = getDsymbol(e);
                 break;
 
+            case TOK.dotTemplateDeclaration:
+                s = (cast(DotTemplateExp)e).td;
+                break;
```
wilzbach: image

uncovered!

WalterBright:
Thanks for your contribution. This needs an issue raised in Bugzilla which this PR can then resolve.

wilzbach (approved):

PR #7974, opened 2018-03-02
Cherry-pick #7969 from stable - Temporarily disable #7577, s.t. a proper deprecation cycle can be started

JinShil:
See #7577 and #7969

Also see dlang/ci#183

@jacob-carlborg


wilzbach:
No cherry-picking from stable to master please. This will only lead to more merge conflicts.
Martin or I will run `merge_stable` soon.
Adding DWT to the project tester is nice, but can wait for one day.

If you really want to get the changes back into master immediately, run this script: https://github.com/wilzbach/git-tools/blob/master/merge_stable

PR #7976, opened 2018-03-03
fix Issue 5212 - no escape analysis for typesafe variadic function ar…

WalterBright:
…guments

The fix is implicitly adding the `scope` annotation.

### DIFF
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -739,7 +739,12 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
                 fparam.storageClass &= (STC.in_ | STC.out_ | STC.ref_ | STC.lazy_ | STC.final_ | STC.TYPECTOR | STC.nodtor);
                 fparam.storageClass |= STC.parameter;
                 if (fvarargs == 2 && i + 1 == nfparams)
+                {
                     fparam.storageClass |= STC.variadic;
+                    /* Don't need to set STC.scope_ because this will only
+                     * be evaluated at compile time
+                     */
+                }
             }
             for (size_t i = 0; i < fparameters.dim; i++)
             {
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -604,7 +604,7 @@ private bool functionParameters(Loc loc, Scope* sc, TypeFunction tf, Type tthis,
                 nargs++;
             }
 
-            if (tf.varargs == 2 && i + 1 == nparams)
+            if (tf.varargs == 2 && i + 1 == nparams) // https://dlang.org/spec/function.html#variadic
             {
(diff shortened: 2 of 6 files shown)


src/dmd/dtemplate.d
```diff
@@ -739,7 +739,12 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
                 fparam.storageClass &= (STC.in_ | STC.out_ | STC.ref_ | STC.lazy_ | STC.final_ | STC.TYPECTOR | STC.nodtor);
                 fparam.storageClass |= STC.parameter;
                 if (fvarargs == 2 && i + 1 == nfparams)
+                {
                     fparam.storageClass |= STC.variadic;
+                    /* Don't need to set STC.scope_ because this will only
+                     * be evaluated at compile time
+                     */
```
ibuclaw: Shouldn't comments come before code in a block?

src/dmd/expressionsem.d
```diff
@@ -604,7 +604,7 @@ private bool functionParameters(Loc loc, Scope* sc, TypeFunction tf, Type tthis,
                 nargs++;
             }
 
-            if (tf.varargs == 2 && i + 1 == nparams)
+            if (tf.varargs == 2 && i + 1 == nparams) // https://dlang.org/spec/function.html#variadic
```
ibuclaw: Maybe the link should be put on the line before this condition. Perhaps with a slight introduction also, eg: `Handle variadic arguments as per spec <link>`.

src/dmd/semantic3.d
```diff
@@ -442,7 +442,15 @@ private extern(C++) final class Semantic3Visitor : Visitor
...
                     stc |= STC.parameter;
                     if (f.varargs == 2 && i + 1 == nparams)
+                    {
                         stc |= STC.variadic;
+                        auto vtypeb = vtype.toBasetype();
+                        if (vtypeb.ty == Tsarray || vtypeb.ty == Tarray)
+                            /* Since it'll be pointing into the stack for the array
+                             * contents, it needs to be `scope`
+                             */
+                            stc |= STC.scope_;
```
ibuclaw: Multiline indent should probably be given braces.

ibuclaw (approved):
Implementation is fine, though not sure on your comment style and positioning.

src/dmd/dtemplate.d
WalterBright: In the other change, the setting of STC.scope_ came after.

src/dmd/expressionsem.d
WalterBright: The spec link implies just that.

ibuclaw: The comment still - to me at least - looks to be in an awkward place.  But well, eye of the beholder and all.

PR #5452, opened 2016-02-14
Issue 15653 - IFTI fails for immutable parameter

9rnsr:

PR #7977, opened 2018-03-03
Fix issue 15653 - IFTI fails for immutable parameter

JinShil:
Revival of #5452

Special thanks to @9rnsr

### DIFF
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -4324,6 +4324,16 @@ MATCH deduceType(RootObject o, Scope* sc, Type tparam, TemplateParameters* param
                 }
             }
 
+            /* Returns `true` if `t` is a reference type, or an array of reference types
+             */
+            bool isTopRef(Type t)
+            {
+                auto tb = t.baseElemOf();
+                return tb.ty == Tclass ||
+                       tb.ty == Taarray ||
+                       tb.ty == Tstruct && tb.hasPointers();
+            }
+
             Type at = cast(Type)(*dedtypes)[i];
             Type tt;
             if (ubyte wx = deduceWildHelper(e.type, &tt, tparam))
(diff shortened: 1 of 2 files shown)


src/dmd/dtemplate.d
```diff
@@ -4324,6 +4324,14 @@ MATCH deduceType(RootObject o, Scope* sc, Type tparam, TemplateParameters* param
                 }
             }
 
+            bool isTopRef(Type t)
+            {
+                auto tb = t.baseElemOf();
+                return tb.ty == Tclass ||
+                       tb.ty == Taarray ||
+                       tb.ty == Tstruct && tb.hasPointers();
```
ibuclaw: Do we need fail tests for these types?

src/dmd/dtemplate.d
```diff
@@ -4324,6 +4324,14 @@ MATCH deduceType(RootObject o, Scope* sc, Type tparam, TemplateParameters* param
                 }
             }
 
+            bool isTopRef(Type t)
```
ibuclaw: Maybe just need a small comment on what this function returns.  True if the `t` is a reference type, or an array of reference types.

ibuclaw (changes requested):
As per comments

src/dmd/dtemplate.d
JinShil: The failure conditions are covered using the `compiles` traits (e.g. `static assert(!__traits(compiles, bar!U(u)), U.stringof);`), in the runnable test.  Is that not sufficient?

ibuclaw: On second look, ok, there are three types.

- `Object` covers `tb.ty == Tclass`
- `Y` covers `tb.ty == Tstruct && tb.hasPointers()`
- `Y[2]` covers `t.baseElemOf()`

So only missing a check for AAs, adding `int[int]` should be enough by the looks.

JinShil: Done.  Thank you!

ibuclaw (approved):

wilzbach (approved):

PR #7981, opened 2018-03-04
fix Issue 16037 - assigning delegate to a scope variable shouldn't al…

WalterBright:
…locate closure

Improves the reach of @nogc code.

### DIFF
--- a/src/dmd/escape.d
+++ b/src/dmd/escape.d
@@ -46,7 +46,8 @@ import dmd.arraytypes;
  */
 bool checkParamArgumentEscape(Scope* sc, FuncDeclaration fdc, Identifier par, Expression arg, bool gag)
 {
-    //printf("checkParamArgumentEscape(arg: %s par: %s)\n", arg.toChars(), par.toChars());
+    enum log = false;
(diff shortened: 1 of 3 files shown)


mathias-lang-sociomantic:
Why is it a DIP1000-only fix ?

WalterBright:
So it doesn't break existing code.

mathias-lang-sociomantic:
Do you have an example of code that would break with this fix ? It looks like a fix to a reject-valid, which should have a limited, if any, impact.

WalterBright:
The thing is, the old behavior was still correct, because it generated a closure. It only failed when the function was marked `@nogc`.

So I was not willing to gratuitously break the old behavior. Recall that a lot of the "don't break working code" is because Sociomantic has strongly advocated that.

I recommend switching to -dip1000 as soon as you can. It has already found some expensive bugs in existing code bases.

mathias-lang-sociomantic:
> The thing is, the old behavior was still correct, because it generated a closure. It only failed when the function was marked `@nogc`.

That is actually correct, and the point I was missing. So objection withdrawn on my side.

JinShil (approved):

PR #7982, opened 2018-03-04
fix Issue 17284 - Template function attribute inference wrongly infer…

WalterBright:
…s @safe for accessing overlapping pointer fields in unions

### DIFF
--- a/test/fail_compilation/test17284.d
+++ b/test/fail_compilation/test17284.d
@@ -1,0 +1,20 @@
+/* REQUIRED_ARGS:
+TEST_OUTPUT:
+---
+fail_compilation/test17284.d(16): Error: field `U.c` cannot access pointers in `@safe` code that overlap other fields
+pure nothrow @safe void(U t)
+---
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=17284
+
+class C { }
+union U { C c; int i; }
+
+@safe void func(T)(T t)
+{
+        t.c = new C;
+}
+
+pragma(msg, typeof(func!U));
+
--- a/src/dmd/safe.d
+++ b/src/dmd/safe.d
@@ -41,6 +41,7 @@ import dmd.tokens;
 
 bool checkUnsafeAccess(Scope* sc, Expression e, bool readonly, bool printmsg)
 {
+    //printf("checkUnsafeAccess(e: '%s', readonly: %d, printmsg: %d)\n", e.toChars(), readonly, printmsg);
     if (e.op != TOK.dotVariable)
         return false;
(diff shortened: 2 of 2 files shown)


WalterBright:
AppVeyor crashed with a linker seg fault.

rainers:
> AppVeyor crashed with a linker seg fault.

I tried to reproduce the crash locally with the same linker version but it worked for me. Maybe just try retriggering the build by pushing again.

WalterBright:
Now it's showing as passing AppVeyor.

JinShil:
I don't know why our CIs are so unreliable.  I restarted AppVeyor and Semaphore (a few times) and all CIs are passing now.

test/fail_compilation/test17284.d
```diff
@@ -0,0 +1,20 @@
+/* REQUIRED_ARGS:
+TEST_OUTPUT:
+---
+fail_compilation/test17284.d(16): Error: field `U.c` cannot access pointers in `@safe` code that overlap other fields
+pure nothrow @safe void(U t)
```
ZombineDev: How about actually verifying that attribute inference actually considers this function `@system`?
Also In the bug report me and H. S. Teoh posted several more test cases, which doesn't seem to be addressed here.

WalterBright: > How about actually verifying that attribute inference considers this function @system?

The message does that.

> Also In the bug report me and H. S. Teoh posted several more test cases, which doesn't seem to be addressed here.

This is the root cause. If you find more cases after this one is pulled, please point them out.

test/fail_compilation/test17284.d
```diff
@@ -0,0 +1,20 @@
...
+
+class C { }
+union U { C c; int i; }
+
+@safe void func(T)(T t)
+{
+        t.c = new C;
+}
+
+pragma(msg, typeof(func!U));
```
wilzbach: Could be converted into a `static assert`

wilzbach (approved):

MoonlightSentinel:
This apparently caused a regression: https://issues.dlang.org/show_bug.cgi?id=20661

schveiguy:
This also has never fixed the true issue.

This code has compiled since 2.072 (October of 2016), and segfaults:

```d
struct UnionOf(T1, T2)
{
    union Values
    {
        T1 t1;
        T2 t2;
    }
    Values values;
    this(T1 t1)
    {
        values.t1 = t1;
    }
    T2 get()
    {
        return values.t2;
    }
}

void main() @safe
{
    alias U = UnionOf!(int, int*);
    U u;
    u = U(1);
    import std.stdio;
    writeln(*u.get);
}
```

PR #7984, opened 2018-03-05
Fix Issue 18551 - Improve hint for does not override any function

RazvanN7:


src/dmd/dsymbolsem.d
```diff
@@ -3164,8 +3164,19 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
 
                 if (s)
-                    funcdecl.error("does not override any function, did you mean to override `%s%s`?",
-                        bc.sym.isCPPclass() ? "extern (C++) ".ptr : "".ptr, s.toPrettyChars());
+                {
+                    auto fd = s.isFuncDeclaration();
+                    HdrGenState hgs;
+                    OutBuffer buf1, buf2;
+
+                    functionToBufferFull(cast(TypeFunction)(fd.type), &buf1,
```
BBasile: Unless `fd` is guaranteed not to be `null` can you assert it is not ?

RazvanN7: Done.

JinShil (approved):

PR #7985, opened 2018-03-05
[Refactor] Move buildPostBlit from dmd/clone.d to dmd/dsymbolsem.d

RazvanN7:


JinShil:
@RazvanN7 Please provide some rationale and motivation for these refactorings.  I don't know what you @WalterBright and @andralex have discussed.  If this is a part of a larger goal, please articulate that larger goal and what role this PR plays in that.  Thanks!

andralex:
@JinShil notice the moved function is `private`, i.e. it was moved to the (only) module that uses it.

PR #7986, opened 2018-03-06
fix Issue 18557 - Types with 0 size should not be usable as aa key types

WalterBright:


WalterBright:
Spec: https://github.com/dlang/dlang.org/pull/2265

WalterBright:
blocked by https://github.com/dlang/phobos/pull/6248

jacob-carlborg:
Can we make this a deprecation instead of an error?

WalterBright:
> Can we make this a deprecation instead of an error?

No, because of the glorious clang++ idea that `struct S { }` occupies zero space on the parameter stack despite `sizeof(S)==1`, but sits at the same address as the next parameter. This means that all copying of such structs must not happen. Meaning that AAs now get uninitialized data errors. Meaning this PR has to happen.

src/dmd/typesem.d
```diff
@@ -13,6 +13,7 @@
 module dmd.typesem;
 
 import core.checkedint;
+import core.stdc.stdio;
```
UplinkCoder: remove ?

WalterBright: No, because I keep having to add it back in.

wilzbach: How about adding a `base` file that public imports `core.stdc.stdio` and other debug goodies when compiled in debug mode?

WalterBright: Why make things more complicated? just import `core.stdc.stdio`

wilzbach (approved):

WalterBright:
Blocked by https://github.com/dlang/druntime/pull/2131

dnadlinger (changes requested):
I'm not sure this is the best solution. Sure, it fixes the issue, but at the expense of introducing a special case – and especially with the ubiquity of generic programming in D, those are almost never a good idea.

There are two alternatives that come to mind:
 - Make sure the one data byte is zeroed whenever such a value is actually "materialized" in memory (e.g. in a stack frame), even if passing it as a parameter is a no-op on the ABI level.
 - Special-case the AA code to always hash zero-size values to the same key, irrespective of the memory contents of the dummy byte.

Either of these options would seem preferable to outright disallowing zero-size keys – and that's not even considering that this PR would be a breaking change without a deprecation period.

WalterBright:
The clang++ convention of 0 sized objects sometimes being 1 byte in size and sometimes having the same address as another random variable is a special case, any solutions I can think of are hackish.

Passing the parameter as a no-op is not the problem, the problem is its location is on top of another variable. So, all assignments are turned off. Now there's another problem - the optimizer fails because a variable with no assignments to it gets used to pass to the AAs.

src/dmd/typesem.d
jacob-carlborg: Like a `debug` statement. But I do like the possibility to only enable debug output in specific areas in the code.

WalterBright:
> You could just allocate a one-byte stack slot for parameters as if you had a local variable declaration instead. This way, the ABI hack wouldn't spill over into other bits of the code. (Incidentally, I think this is what Clang does.)

That has another problem - what about variadic argument lists? That depends on being able to take the address of the last named parameter.

WalterBright:
I moved the test for 0 size key to the expression semantics. Doing it in the type semantics causes forward reference problems.

WalterBright:
> Does it? va_start can just use the internal frame information, disregarding the second argument.

I suppose if you want to make va_start a magic compiler function for 32 bits. There comes a point where "is it worth it" for AAs of 0 key size that won't work anyway (without the special case hack for them you suggested earlier).

> va_arg is more complex than just bumping a pointer on Posix x86_64 anyway.

The problem is also on 32 bit clang++.

WalterBright:
> how would any problems encountered with this be specific to AAs?

AAs would have the key size listed as being 1, but the size on the parameter stack is 0. The size being 0 means that the compiler does not do assignments to it. Then, the value gets passed to the AA functions. The optimizer sees that no value was ever assigned to the key, yet it gets passed to the AA. This results in "use before set" optimizer errors.

The data flow analysis is pretty sacred, it's very very dangerous to try and special case it. When you talk about "cracks", that's one terrible one by accepting 0 sized structs.

0 size key types are pointless anyway, and I don't see any problem with disallowing them.

WalterBright:
As the autotester shows, I have yet to find *any* instance of a zero sized AA key that was actually used (as opposed to just being a dummy type).

WalterBright:
Put in Phantom Zone

PR #7987, opened 2018-03-08
Fix issue 18574: Unclear error message when trying to inherit from mu…

thewilsonator:
…ltiple classes

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -4397,6 +4397,8 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
 
             // Treat the remaining entries in baseclasses as interfaces
             // Check for errors, handle forward references
+            bool multiClassError = false;
+
             for (size_t i = (cldec.baseClass ? 1 : 0); i < cldec.baseclasses.dim;)
             {
                 BaseClass* b = (*cldec.baseclasses)[i];
@@ -4404,8 +4406,29 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                 TypeClass tc = (tb.ty == Tclass) ? cast(TypeClass)tb : null;
                 if (!tc || !tc.sym.isInterfaceDeclaration())
                 {
-                    if (b.type != Type.terror)
-                        cldec.error("base type must be `interface`, not `%s`", b.type.toChars());
+                    // It's a class
+                    if (tc)
+                    {
+                        if (!multiClassError)
+                        {
+                            error(cldec.loc,"`%s`: multiple class inheritance is not supported." ~
+                                  " Use multiple interface inheritance and/or composition.", cldec.toPrettyChars());
+                            multiClassError = true;
+                        }
+
+                        if (tc.sym.fields.dim)
+                            errorSupplemental(cldec.loc,"`%s` has fields, consider making it a member of `%s`",
+                                              b.type.toChars(), cldec.type.toChars());
+                        else
+                            errorSupplemental(cldec.loc,"`%s` has no fields, consider making it an `interface`",
+                                              b.type.toChars());
+                    }
+                    // It's something else: e.g. `int` in `class Foo : Bar, int { ... }`
+                    else if (b.type != Type.terror)
+                    {
+                        error(cldec.loc,"`%s`: base type must be `interface`, not `%s`",
+                              cldec.toPrettyChars(), b.type.toChars());
+                    }
                     cldec.baseclasses.remove(i);
                     continue;
                 }
--- a/test/fail_compilation/diag18574.d
+++ b/test/fail_compilation/diag18574.d
@@ -1,0 +1,17 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/diag18574.d(16): Error: `diag18574.Test`: multiple class inheritance is not supported. Use multiple interface inheritance and/or composition.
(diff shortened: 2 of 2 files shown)


MetaLang:
Wow, that was fast. Thanks @thewilsonator!

src/dmd/dsymbolsem.d
```diff
@@ -4397,15 +4397,36 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
                 if (!tc || !tc.sym.isInterfaceDeclaration())
                 {
-                    if (b.type != Type.terror)
+                    // It's a class
+                    if (tc)
+                    {
+                        if (!multiClassError)
+                        {
+                            cldec.error("D does not support multiple class inheritance." ~
+                                        " Use multiple interface inheritance and/or membership.");
```
JinShil: s/membership/composition

JinShil (changes requested):

src/dmd/dsymbolsem.d
```diff
@@ -4397,15 +4397,36 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
                 TypeClass tc = (tb.ty == Tclass) ? cast(TypeClass)tb : null;
                 if (!tc || !tc.sym.isInterfaceDeclaration())
                 {
-                    if (b.type != Type.terror)
+                    // It's a class
+                    if (tc)
+                    {
+                        if (!multiClassError)
+                        {
+                            cldec.error("D does not support multiple class inheritance." ~
```
JinShil: You might  want to use `error(cldec.loc, "D does not..")` instead of `cldec.error()` because the `class `diag18574.Test` prefix that `cldec.error` adds to the error message makes it read weird.

thewilsonator: I See what you mean, but it is nice to have the symbol for where the source location doesn't correspond well to the error (cf. templates mixins), what about `Error: class diag18574.Test multiple class inheritance is not supported. Use ...`?

JinShil: I think if there was simply a colon after the prefix, it would read a lot better.  What about `error(cldec.loc, "%s: D does not support...,", cldec.toPrettyChars())`?  It's on my list to fix the prefix oddity (https://github.com/dlang/dmd/projects/3#card-6750742).

thewilsonator:
Will do.

src/dmd/dsymbolsem.d
```diff
@@ -4397,15 +4397,40 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
-                        cldec.error("base type must be `interface`, not `%s`", b.type.toChars());
+                    // It's a class
+                    if (tc)
+                    {
+                        if (!multiClassError)
+                        {
+                            error(cldec.loc,"`%s`: D does not support multiple class inheritance." ~
+                                  " Use multiple interface inheritance and/or composition.",
+                                  cldec.toPrettyChars()
+                                  );
```
BBasile: can you put this on the previous line ?

thewilsonator: Will do, just waiting for checks before I squash the lot.

src/dmd/dsymbolsem.d
```diff
@@ -4397,15 +4397,38 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                    {
+                        if (!multiClassError)
+                        {
+                            error(cldec.loc,"`%s`: D does not support multiple class inheritance." ~
+                                  " Use multiple interface inheritance and/or composition.", cldec.toPrettyChars());
+                            multiClassError = true;
+                        }
+
+                        if (tc.sym.fields.dim)
+                            errorSupplemental(cldec.loc,"`%s` has fields: consider making it a member of `%s`",
```
JinShil: I believe this this should have a semicolon, not a colon "... has fields; consider..."

thewilsonator: A semicolon delimits a list, thats not really a list. Also there would no cooling preceding it which would be more weird. (Also also, when I'm programming semicolons mean something completely different ;))

MetaLang: I agree with @JinShil  that a colon is not really appropriate here.

thewilsonator: A comma or a full stop then? 
A semicolon is much worse than a colon as the `errorSupplemental`s form items in a list. delimiting the propositions(?) _within_ the items with semicolons is just wrong.

JinShil: I think a comma would be fine.

thewilsonator: Done

src/dmd/dsymbolsem.d
```diff
@@ -4397,15 +4397,38 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
                 if (!tc || !tc.sym.isInterfaceDeclaration())
                 {
-                    if (b.type != Type.terror)
-                        cldec.error("base type must be `interface`, not `%s`", b.type.toChars());
+                    // It's a class
+                    if (tc)
+                    {
+                        if (!multiClassError)
+                        {
+                            error(cldec.loc,"`%s`: D does not support multiple class inheritance." ~
```
WalterBright: The `D` is implied and can be omitted, use "multiple class inheritance is not supported."

WalterBright (approved):

PR #6703, opened 2017-04-15
RFC: Support for interpolated strings in D

jcd:
Support for interpolated strings like e.g. C# has. It will let you write code like:

```
void main()
{
  auto name = "John Doe";
  auto halfAge = 42;
  writeln( $"The age of {name} is {2 * halfAge}");
}
```

This PR is a request for comments on the approach taken and to see if this is something that has a chance of being included.

It needs the addition of the following function to object.d :

```
string _interp_text(T...)(T args)
{
    static if(T.length == 0)
        return null;

    // Check if phobos std.conv is present and use if so
    // else make a best effort to convert to string.
    static if (__traits(compiles, { import std.conv; return 0; }()))
    {
        static import std.conv;
        return std.conv.text(args);
    }
    else static if (__traits(compiles, args[0].toString()))
        return args[0].toString() ~ _interp_text(args[1..$]);
    else static if (is(args[0] == string))
        return args[0] ~ _interp_text(args[1..$]);
    else
        return "?" ~ _interp_text(args[1..$]);
}
```

PR #7988, opened 2018-03-08
Interpolated Strings

marler8997:
Added interpolated strings.  Syntax is to prefix any kind of string literal with the letter `i`, i.e.
```D
int a = 100;
writeln(i"a is $(a)");
writeln(i`a + 56 = $(a + 56)`);
writeln(iq{ $(a) * $(a) = $(a*a) });
```

Note that an interpolated string does not generate code to build a string at runtime, instead, lowers to a tuple.
```D
i"a is $(a) and b is $(b)"

// IS THE SAME AS

tuple("a is ", a, " and b is ", b);
```

> Note: it's not quite the same as `tuple(...)`.  It creates the same kind of tuple you would get from a variadic template `template Foo(T...)`, the key difference being that it "expands" its elements to separate arguments when passed to a function call.

A good way to think about an interpolated string is that it's just another way of creating a tuple.  It's a nice way to write a tuple if the tuple is mostly made of raw characters (i.e. mixin code, HTML generation, etc). Generating a tuple gives full control to the application for determining what to do with it. To build a runtime string from an interpolated string, you can use the `text` function, i.e.
```D
import std.conv : text;

int a = 32;
assert("a is 32" == text(i"a is $(a)"));

// equivalent to
assert("a is 32" == text(tuple("a is ", a)));
```

Here's an example where you can generate code using an interpolated `iq{ ... }` string:
```D
string generateFunction(string attributes, string returnType, string name, string args, string body)
{
    import std.conv : text;
    return text(iq{
        $(attributes) $(returnType) $(name)($(args))
        {
            $(body)
        }
    });
}
mixin(generateFunction("pragma(inline)", "int", "add", "int a, int b", "return a + b;"));
assert(100 == add(25, 75));
```

This example demonstrates how interpolated strings enable using the `q{...}` string literal for code generation even if the string itself contains curly braces.  This wasn't possible before.

### A Note on Performance

Theoretically (though I haven't measured yet) this form of string should have very good compile-time and run-time performance when it comes to creating strings.  Since it lowers to a tuple, it allows a print or string building function to instantiate the minimal code needed to build/write the string.  It should have the same run-time performance as `format!"..."(...)` but should also have MUCH better compile-time performance.

To add a bit more detail, this PR performs string interpolation in the "parse stage".  This means it is done after the string literal has been processed/escaped.  This also means it is done before the semantic stage which makes it more performant than any library solution could be allowing it to be used with mixins or low-level code without having to give up performance or sacrifice more overhead.  It also allows error messages to point directly to source code instead of pointing to code that is "mixed in" in the semantic stage.

Previous attempt: https://github.com/dlang/dmd/pull/6703

### DIFF
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -684,6 +684,8 @@ dmd -cov -unittest myprog.d
             "list all variables going into thread local storage"),
         Feature("vmarkdown", "vmarkdown",
             "list instances of Markdown replacements in Ddoc"),
+        Feature("interpolate", "interpolateStrings",
+            "enable interpolated string support via the 'i' prefix on string literals"),
     ];
 
     /// Returns all available reverts
--- a/test/runnable/istring.d
+++ b/test/runnable/istring.d
@@ -1,0 +1,72 @@
+/*
+PERMUTE_ARGS:
+REQUIRED_ARGS: -transition=interpolate
+*/
+import std.conv : text;
+
+static assert(i"$()".length == 0);
+static assert(i"$(/* a comment!*/)".length == 0);
+static assert(i"$(// another comment)".length == 0);
+static assert(i"$(/+ yet another comment+/)".length == 0);
+
+void main()
+{
+    int a = 42;
+    assert("a is 42" == text(i"a is $(a)"));
+    assert("a + 23 is 65" == text(i"a + 23 is $(a + 23)"));
+
+    // test each type of string literal
+    int b = 93;
+    assert("42 + 93 = 135" == text(  i"$(a) + $(b) = $(a + b)"));  // double-quote
+    assert("42 + 93 = 135" == text( ir"$(a) + $(b) = $(a + b)"));  // wysiwyg
+    assert("42 + 93 = 135" == text(  i`$(a) + $(b) = $(a + b)`));  // wysiwyg (alt)
+    assert("42 + 93 = 135" == text( iq{$(a) + $(b) = $(a + b)}));  // token
+    assert("42 + 93 = 135" == text(iq"!$(a) + $(b) = $(a + b)!")); // delimited (char)
+    assert("42 + 93 = 135\n" == text(iq"ABC
+$(a) + $(b) = $(a + b)
+ABC")); // delimited (heredoc)
+
+    // Escaping double dollar
+    assert("$" == i"$$"[0]);
+    assert(" $ " == i" $$ "[0]);
+    assert(" $(just raw string) " == i" $$(just raw string) "[0]);
+    assert("Double dollar $$ becomes $" == text(  i"Double dollar $$$$ becomes $$"));  // double-quote
+    assert("Double dollar $$ becomes $" == text( ir"Double dollar $$$$ becomes $$"));  // wysiwyg
+    assert("Double dollar $$ becomes $" == text(  i`Double dollar $$$$ becomes $$`));  // wysiwyg (alt)
+    assert("Double dollar $$ becomes $" == text( iq{Double dollar $$$$ becomes $$}));  // token
+    assert("Double dollar $$ becomes $" == text(iq"!Double dollar $$$$ becomes $$!")); // delimited
+
+    assert(928 == add(900, 28));
+}
+
+string funcCode(string attributes, string returnType, string name, string args, string body)
+{
+    return text(iq{
+    $(attributes) $(returnType) $(name)($(args))
+    {
+        $(body)
+    }
+    });
+}
+mixin(funcCode("pragma(inline)", "int", "add", "int a, int b", "return a + b;"));
+
+// Test interpolated strings with escape sequences
+static assert(i" foo \n bar".length == 1);
+static assert(i"foo \x0a bar".length == 1);
+static assert(i"foo \xC2\xA2 bar".length == 1);
+static assert(i"foo \u042f bar".length == 1);
+static assert(i"foo \U00010f063 bar".length == 1);
+static assert(i"foo \0 bar".length == 1);
+static assert(i"foo \1 bar".length == 1);
+static assert(i"foo \7 bar".length == 1);
+static assert(i"foo \01 bar".length == 1);
+static assert(i"foo \001 bar".length == 1);
+static assert(i"foo \377 bar".length == 1);
+static assert(i"foo &quot; bar".length == 1);
+
+// Test string literals with odd newlines
+static assert(i"
+".length == 1);
+// test carriage return
+static assert(i"
+".length == 1);
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -173,7 +173,7 @@ struct Param
 
     bool markdown;          // enable Markdown replacements in Ddoc
     bool vmarkdown;         // list instances of Markdown replacements in Ddoc
-
+    bool interpolateStrings;// Enable interpolated string support via the 'i' prefix on string literals
     bool showGaggedErrors;  // print gagged errors anyway
     bool printErrorContext;  // print errors with the error context (the error line in the source file)
     bool manual;            // open browser on compiler manual
--- a/src/dmd/globals.h
+++ b/src/dmd/globals.h
@@ -141,6 +141,7 @@ struct Param
     CppStdRevision cplusplus;  // version of C++ name mangling to support
     bool markdown;          // enable Markdown replacements in Ddoc
     bool vmarkdown;         // list instances of Markdown replacements in Ddoc
+    bool interpolateStrings;// Enable interpolated string support via the 'i' prefix on string literals
     bool showGaggedErrors;  // print gagged errors anyway
     bool printErrorContext;  // print errors with the error context (the error line in the source file)
     bool manual;            // open browser on compiler manual
--- a/src/dmd/lexer.d
+++ b/src/dmd/lexer.d
@@ -420,7 +420,7 @@ class Lexer
     bool commentToken;      // comments are TOK.comment's
     int lastDocLine;        // last line of previous doc comment
 
-    private DiagnosticReporter diagnosticReporter;
+    protected DiagnosticReporter diagnosticReporter;
 
     private Token* tokenFreelist;
 
@@ -646,6 +646,44 @@ class Lexer
             case '"':
                 escapeStringConstant(t);
                 return;
+            case 'i':
+                if (global.params.interpolateStrings)
+                {
+                    if (p[1] == 'r')
+                    {
+                        if (p[2] == '"')
+                        {
+                            p += 2;
+                            goto case '`';
+                        }
+                    }
+                    else if (p[1] == '`')
+                    {
+                        p++;
+                        goto case '`';
+                    }
+                    else if (p[1] == '"')
+                    {
+                        p++;
+                        goto case '"';
+                    }
+                    else if (p[1] == 'q')
+                    {
+                        if (p[2] == '"')
+                        {
+                            p += 2;
+                            delimitedStringConstant(t);
+                            return;
+                        }
+                        else if (p[2] == '{')
+                        {
+                            p += 2;
+                            tokenStringConstant(t);
+                            return;
+                        }
+                    }
+                }
+                goto case_ident;
             case 'a':
             case 'b':
             case 'c':
@@ -654,7 +692,6 @@ class Lexer
             case 'f':
             case 'g':
             case 'h':
-            case 'i':
             case 'j':
             case 'k':
             case 'l':
@@ -2491,6 +2528,13 @@ class Lexer
         diagnosticReporter.error(loc, format, args);
         va_end(args);
     }
+    // temporary hack to handle r-value references to Loc structures
+    pragma(inline)
+    final void error(T...)(const Loc loc, const(char)* format, T args)
+    {
+        error(loc, format, args);
+    }
+
 
     final void errorSupplemental(const ref Loc loc, const(char)* format, ...)
     {
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -24,6 +24,7 @@ import dmd.root.outbuffer;
 import dmd.root.rmem;
 import dmd.root.rootobject;
 import dmd.tokens;
+import dmd.utf;
 
 // How multiple declarations are parsed.
 // If 1, treat as C.
@@ -7745,6 +7746,13 @@ final class Parser(AST) : Lexer
             break;
 
         case TOK.string_:
+            if (global.params.interpolateStrings && token.ptr[0] == 'i')
+            {
+                e = parseInterpolatedString(token);
+                nextToken();
+                break;
+            }
+            goto case TOK.hexadecimalString;
         case TOK.hexadecimalString:
             {
                 // cat adjacent strings
@@ -7762,6 +7770,10 @@ final class Parser(AST) : Lexer
                             if (token.postfix != postfix)
                                 error("mismatched string literal postfixes `'%c'` and `'%c'`", postfix, token.postfix);
                             postfix = token.postfix;
+                        }
+                        if (token.ptr[0] == 'i')
+                        {
+                            error("cannot implicitly concatenate interpolated strings with non-interpolated strings");
                         }
 
                         error("Implicit string concatenation is deprecated, use %s ~ %s instead",
@@ -8995,6 +9007,305 @@ final class Parser(AST) : Lexer
             s.addComment(combineComments(blockComment, token.lineComment, true));
             token.lineComment = null;
         }
+    }
+
+    /**
+    Parse the given interpolated string into a tuple of expressions.
+
+    Params:
+        token = the interpolated string token
+
+    Returns:
+        A tuple of expressions from the interpolated string.
+    */
+    AST.TupleExp parseInterpolatedString(Token token)
+    in { assert(token.value == TOK.string_ && token.ptr[0] == 'i'); } body
+    {
+        //fprintf(stderr, "parseInterpolatedString `%.*s`\n", token.len, token.ustring);
+
+        auto parts = new AST.Expressions();
+
+        // Used to map string contents back to original source location
+        auto sourcePos = StringSourcePos(token.ptr);
+
+        auto str = token.ustring;
+        auto len = token.len;
+        size_t mark = 0;
+        auto markSourcePos = sourcePos;
+        size_t next = 0;
+        size_t doubleDollar = size_t.max;
+
+    KmainLoop:
+        while(true)
+        {
+            auto endOfRawCharacters = next;
+            if (next < len)
+            {
+                auto nextChar = sourcePos.scan(str, next);
+                if (nextChar != '$')
+                    continue;
+
+                if (next >= len)
+                {
+                    error(sourcePos.loc.from(token.loc), "unfinished interpolated string expression '$'");
+                    break;
+                }
+                if (str[next] == '$') // handle $$
+                {
+                    if (doubleDollar == size_t.max)
+                        doubleDollar = next - 1;
+                    sourcePos.scan(str, next);
+                    continue;
+                }
+            }
+
+            // Add next string expression
+            if (endOfRawCharacters > mark)
+            {
+                auto markLoc = markSourcePos.loc.from(token.loc);
+                if (doubleDollar == size_t.max)
+                {
+                    parts.push(new AST.StringExp(markLoc, cast(char*)str + mark, endOfRawCharacters - mark, token.postfix));
+                }
+                else
+                {
+                    auto buffer = cast(char*)mem.xmalloc(endOfRawCharacters - 1 - mark);
+                    size_t offset;
+                    {
+                        auto length = doubleDollar + 1 - mark;
+                        buffer[0 .. length] = str[mark .. doubleDollar + 1];
+                        offset = length;
+                    }
+                    for (size_t i = doubleDollar + 2; i < endOfRawCharacters; i++)
+                    {
+                        auto c = str[i];
+                        if (c == '$')
+                        {
+                            i++;
+                            assert(i < endOfRawCharacters && str[i] == '$');
+                        }
+                        buffer[offset++] = c;
+                    }
+                    parts.push(new AST.StringExp(markLoc, buffer, offset, token.postfix));
+                    doubleDollar = size_t.max;
+                }
+            }
+
+            if (next >= len)
+                break;
+
+            // Process the '$' expression
+            if (str[next] == '(')
+            {
+                sourcePos.scan(str, next);
+                mark = next;
+                for(uint depth = 1;;)
+                {
+                    if (next >= len)
+                    {
+                        error(sourcePos.loc.from(token.loc), "unfinished interpolated string expression '$(...)'");
+                        break KmainLoop;
+                    }
+                    auto nextChar = sourcePos.scan(str, next);
+                    if (nextChar == ')')
+                    {
+                        depth--;
+                        if (depth == 0)
+                            break;
+                    }
+                    else if (nextChar == '(')
+                    {
+                        depth++;
+                    }
+                }
+                {
+                    auto writeableStr = cast(char*)str;
+                    // Need to null-terminate so the parser does not scan past the end of
+                    // the expression.  A case has been found where the parser will scan
+                    // past the expression without this, namely, i"$(var)'"
+                    writeableStr[next - 1] = '\0';
+                    scope(exit) writeableStr[next - 1] = ')';
+
+                    auto expr = str[mark .. next - 1];
+                    //printf("parsing the expression '%s'\n", expr.ptr);
+                    scope tempParser = new Parser!AST(mod, expr, false, diagnosticReporter);
+                    tempParser.scanloc = sourcePos.loc.from(token.loc);
+                    tempParser.nextToken();
+                    if (tempParser.token.value != TOK.endOfFile)
+                    {
+                        auto result = tempParser.parseExpression();
+                        if (tempParser.token.value != TOK.endOfFile)
+                        {
+                            error(sourcePos.loc.from(token.loc), "invalid expression '%s' inside interpolated string", expr.ptr);
+                            break;
+                        }
+                        parts.push(result);
+                    }
+                }
+                mark = next;
+                markSourcePos = sourcePos;
+            }
+            else
+            {
+                // TODO: if we want to support `$` expressions without parentheses, this is
+                //       where we would add support for it. Maybe a good grammar node for this
+                //       would be DotIdentifier.
+                error(sourcePos.loc.from(token.loc), "missing parentheses in interpolated string expression '$(...)'");
+                break;
+            }
+        }
+
+        return new AST.TupleExp(token.loc, parts);
+    }
+
+    /**
+    Represents a location offset from a `Loc`.
+    */
+    static struct LocOffset
+    {
+        uint offset;         /// byte offset from base loc
+        uint line;           /// line offset from base loc
+        uint lastLineOffset; /// byte offset of the start of the last line
+
+        /**
+        Indicates 'offset' has been moved to the next line.
+        */
+        void atNextLine()
+        {
+            line++;
+            lastLineOffset = offset;
+        }
+
+        /**
+        Get location relative to `baseLoc`.
+        Params:
+            baseLoc = relative base location
+        Returns:
+            location relative to `baseLoc`
+        */
+        Loc from(ref const Loc baseLoc) const
+        {
+            return Loc(baseLoc.filename,
+                       baseLoc.linnum + line,
+                       (line == 0) ? baseLoc.charnum + offset : offset - lastLineOffset);
+        }
+    }
+
+    /**
+    Used to map offsets in a processed string back to the source location
+    */
+    private static struct StringSourcePos
+    {
+        private const(char)* sourcePtr;
+        private bool wysiwyg;
+        LocOffset loc;
+
+        this(const char* sourcePtr)
+        {
+            this.sourcePtr = sourcePtr;
+            loc.offset = 2;
+            if (sourcePtr[1] == '"')
+                wysiwyg = false;
+            else
+            {
+                wysiwyg = true;
+                if (sourcePtr[1] == 'r') // ir"
+                    loc.offset++;
+                else if (sourcePtr[1] == '`') // i`
+                { }
+                else
+                {
+                    assert(sourcePtr[1] == 'q', "code bug");
+                    loc.offset = 3;
+                    if (sourcePtr[2] != '{')
+                    {
+                        assert(sourcePtr[2] == '"', "code bug");
+                        char c = sourcePtr[3];
+                        bool isheredoc;
+                        if (c >= 0x80)
+                        {
+                            import dmd.utf : utf_decodeChar, isUniAlpha;
+                            size_t tempOffset = 3;
+                            dchar fullChar;
+                            assert(!utf_decodeChar(sourcePtr, size_t.max, tempOffset, fullChar), "code bug");
+                            isheredoc = isUniAlpha(fullChar);
+                        }
+                        else
+                        {
+                            import core.stdc.ctype : isalpha;
+                            isheredoc = (isalpha(c) || c == '_');
+                        }
+                        if (isheredoc)
+                            loc.offset = 3 + cast(uint)(strchr(sourcePtr + 3, '\n') - (sourcePtr + 3)) + 1;
+                        else
+                            loc.offset = 4;
+                    }
+                }
+            }
+        }
+
+        /**
+        Read and move past the next character both in the source string and
+        in the processed string.
+        Params:
+            str = the processed string
+            ridx = in/out index into the processed string
+        Returns:
+            the next character
+        */
+        dchar scan(const(char)* str, ref size_t ridx)
+        {
+            dchar sourceChar = sourcePtr[loc.offset++];
+            dchar strChar;
+            if (sourceChar == '\r')
+            {
+                if (sourcePtr[loc.offset] == '\n')
+                {
+                    loc.offset++;
+                    loc.atNextLine();
+                    sourceChar = '\n';
+                }
+                strChar = str[ridx++];
+            }
+            else if (wysiwyg || sourceChar != '\\')
+            {
+                strChar = str[ridx++];
+                if (sourceChar == '\n')
+                    loc.atNextLine();
+            }
+            else
+            {
+                scope ignore = new IgnoreDiagnosticReporter();
+                auto escapeStart = sourcePtr + loc.offset;
+                if (escapeStart[0] == 'u' || escapeStart[0] == 'U' || escapeStart[0] == '&')
+                    assert(!utf_decodeChar(str, size_t.max, ridx, strChar), "code bug");
+                else
+                    strChar = str[ridx++];
+
+                auto escapeEnd = escapeStart;
+                auto tempLoc = Loc(); // TODO: get the current location correctly
+                sourceChar = Lexer.escapeSequence(tempLoc, ignore, escapeEnd);
+                loc.offset += (escapeEnd - escapeStart);
+            }
+            assert(strChar == sourceChar/*, "strChar `" ~ strChar ~ "' != sourceChar '" ~ sourceChar ~ "'"*/);
+            return strChar;
+        }
+    }
+    private static class IgnoreDiagnosticReporter : DiagnosticReporter
+    {
+        import core.stdc.stdarg : va_list;
+        private int _errorCount;
+        private int _warningCount;
+        private int _deprecationCount;
+        override int errorCount() { return _errorCount; }
+        override int warningCount() { return _warningCount; }
+        override int deprecationCount() { return _deprecationCount; }
+        override void error(const ref Loc loc, const(char)* format, va_list args) { _errorCount++; }
+        override void errorSupplemental(const ref Loc loc, const(char)* format, va_list) { _errorCount++; }
+        override void warning(const ref Loc loc, const(char)* format, va_list args) { _warningCount++; }
+        override void warningSupplemental(const ref Loc loc, const(char)* format, va_list) { _warningCount++; }
+        override void deprecation(const ref Loc loc, const(char)* format, va_list args) { _deprecationCount++; }
+        override void deprecationSupplemental(const ref Loc loc, const(char)* format, va_list) { _deprecationCount++; }
     }
 }
 
--- a/test/Makefile
+++ b/test/Makefile
@@ -202,8 +202,8 @@ $(RESULTS_DIR)/d_do_test$(EXE): tools/d_do_test.d $(RESULTS_DIR)/.created
 	@echo "OS: '$(OS)'"
 	@echo "MODEL: '$(MODEL)'"
 	@echo "PIC: '$(PIC_FLAG)'"
-	$(DMD) -conf= $(MODEL_FLAG) $(DEBUG_FLAGS) -unittest -run $<
-	$(DMD) -conf= $(MODEL_FLAG) $(DEBUG_FLAGS) -od$(RESULTS_DIR) -of$(RESULTS_DIR)$(DSEP)d_do_test$(EXE) $<
+	$(DMD) -conf= $(MODEL_FLAG) $(DEBUG_FLAGS) -transition=interpolate -unittest -run $<
+	$(DMD) -conf= $(MODEL_FLAG) $(DEBUG_FLAGS) -transition=interpolate -od$(RESULTS_DIR) -of$(RESULTS_DIR)$(DSEP)d_do_test$(EXE) $<
 
 $(RESULTS_DIR)/sanitize_json$(EXE): tools/sanitize_json.d $(RESULTS_DIR)/.created
 	@echo "Building sanitize_json tool"
--- a/test/fail_compilation/istring1.d
+++ b/test/fail_compilation/istring1.d
@@ -1,0 +1,33 @@
+/*
+PERMUTE_ARGS:
+REQUIRED_ARGS: -transition=interpolate
+TEST_OUTPUT:
+---
+fail_compilation/istring1.d(16): Error: missing parentheses in interpolated string expression '$(...)'
+fail_compilation/istring1.d(23): Error: unfinished interpolated string expression '$(...)'
+fail_compilation/istring1.d(26): Error: unfinished interpolated string expression '$'
+fail_compilation/istring1.d(29): Error: invalid expression '1 + 2;' inside interpolated string
+fail_compilation/istring1.d(32): Error: undefined escape sequence \c
+fail_compilation/istring1.d(33): Error: unterminated named entity &quot";
+---
+*/
+enum s1 = i`
+
+    $!
+
+`;
+enum s2 = i`
+
+    $(
+
+`;
+enum s3 = i`
+
+    $`;
+enum s4 = i`
+
+    $(1 + 2;)`;
+
+// Test that bad escape sequences are handled sanely
+enum s5 = i"\c";
+enum s6 = i"\&quot";
--- a/test/fail_compilation/istring2.d
+++ b/test/fail_compilation/istring2.d
@@ -1,0 +1,11 @@
+/*
+PERMUTE_ARGS:
+REQUIRED_ARGS: -transition=interpolate
+TEST_OUTPUT:
+---
+fail_compilation/istring2.d(11): Error: undefined identifier `a`
+---
+*/
+enum s1 = i`
+
+    $(a)`;
--- a/test/run.d
+++ b/test/run.d
@@ -88,6 +88,7 @@ Options:
 
     // bootstrap all needed environment variables
     auto env = getEnvironment;
+    hostDMD = buildPath("..", "generated", env["OS"], env["BUILD"], env["MODEL"], "dmd" ~ env["EXE"]);
 
     if (runUnitTests)
     {
@@ -162,6 +163,7 @@ void ensureToolsExists(const TestTool[] tools ...)
             const command = [
                 hostDMD,
                 "-of"~targetBin,
+                "-transition=interpolate",
                 sourceFile
             ] ~ tool.extraArgs;
 
--- a/test/tools/d_do_test.d
+++ b/test/tools/d_do_test.d
@@ -519,8 +519,7 @@ string envGetRequired(in char[] name)
     auto value = environment.get(name);
     if(value is null)
     {
-        writefln("Error: missing environment variable '%s', was this called this through the Makefile?",
-            name);
+        writeln(i"Error: missing environment variable '$(name)', was this called this through the Makefile?");
         throw new SilentQuit();
     }
     return value;
@@ -733,13 +732,13 @@ int tryMain(string[] args)
                 string objfile = output_dir ~ envData.sep ~ test_name ~ "_" ~ to!string(permuteIndex) ~ envData.obj;
                 toCleanup ~= objfile;
 
-                command = format("%s -conf= -m%s -I%s %s %s -od%s -of%s %s %s%s %s", envData.dmd, envData.model, input_dir,
-                        reqArgs, permutedArgs, output_dir,
-                        (testArgs.mode == TestMode.RUN || testArgs.link ? test_app_dmd : objfile),
-                        argSet,
-                        (testArgs.mode == TestMode.RUN || testArgs.link ? "" : "-c "),
-                        join(testArgs.sources, " "),
-                        (autoCompileImports ? "-i" : join(testArgs.compiledImports, " ")));
+                command = text(
+                    i"$(envData.dmd) -conf= -m$(envData.model) -I$(input_dir) $(reqArgs) ",
+                    i"$(permutedArgs) -od$(output_dir) -of",
+                    (testArgs.mode == TestMode.RUN || testArgs.link) ? test_app_dmd : objfile,
+                    i` $(argSet) $(testArgs.mode == TestMode.RUN || testArgs.link ? "" : "-c ") `,
+                    join(testArgs.sources, " "), " ",
+                    (autoCompileImports ? "-i" : join(testArgs.compiledImports, " ")));
                 version(Windows) command ~= " -map nul.map";
 
                 compile_output = execute(fThisRun, command, testArgs.mode != TestMode.FAIL_COMPILE, result_path);
@@ -751,18 +750,21 @@ int tryMain(string[] args)
                     string newo= result_path ~ replace(replace(filename, ".d", envData.obj), envData.sep~"imports"~envData.sep, envData.sep);
                     toCleanup ~= newo;
 
-                    command = format("%s -conf= -m%s -I%s %s %s -od%s -c %s %s", envData.dmd, envData.model, input_dir,
-                        reqArgs, permutedArgs, output_dir, argSet, filename);
+                    command = text(
+                        i"$(envData.dmd) -conf= -m$(envData.model) -I$(input_dir) $(reqArgs) ",
+                        i"$(permutedArgs) -od$(output_dir) -c $(argSet) $(filename)");
                     compile_output ~= execute(fThisRun, command, testArgs.mode != TestMode.FAIL_COMPILE, result_path);
                 }
 
                 if (testArgs.mode == TestMode.RUN || testArgs.link)
                 {
                     // link .o's into an executable
-                    command = format("%s -conf= -m%s%s%s %s %s -od%s -of%s %s", envData.dmd, envData.model,
+                    command = text(
+                        i"$(envData.dmd) -conf= -m$(envData.model)",
                         autoCompileImports ? " -i" : "",
                         autoCompileImports ? "extraSourceIncludePaths" : "",
-                        envData.required_args, testArgs.requiredArgsForLink, output_dir, test_app_dmd, join(toCleanup, " "));
+                        i" $(envData.required_args) $(testArgs.requiredArgsForLink) -od$(output_dir)",
+                        i" -of$(test_app_dmd) $(join(toCleanup, ` `))");
                     version(Windows) command ~= " -map nul.map";
 
                     execute(fThisRun, command, true, result_path);
@@ -870,11 +872,10 @@ int tryMain(string[] args)
             }
             f.writeln();
             f.writeln("==============================");
-            f.writef("Test %s failed: ", input_file);
-            f.writeln(e.msg);
+            f.writeln(i"Test $(input_file) failed: $(e.msg)");
             f.close();
 
-            writefln("\nTest %s failed.  The logged output:", input_file);
+            writeln("\nTest $(input_file) failed.  The logged output:");
             auto outputText = output_file.readText;
             writeln(outputText);
             output_file.remove();


wilzbach:
See also: https://github.com/dlang/dmd/pull/6703

wilzbach:
Regarding syntax. I had a short look in this a couple of months ago and here's what other languages use:

- "I have $apples apples" - Bash, Dart, E, Haxe, Frink, FunL, Julia, Perl, PHP
- "I have ${apples} apples" - Bash, Dart, Groovy, Haxe, Kotlin, PHP
- $"I have {apples} apples" - C#
- "I have #{apples} apples" - CoffeeScript, Elixir, OOC, Ruby
- `I have ${apples} apples` - JavaScript
- f"I have {apples} apples" - Python
- s"I have $apples apples" - Scala
- s"I have ${apples + bananas} fruits" - Scala
- "I have \(apples) apples" - Swift

More details: https://en.wikipedia.org/wiki/String_interpolation and https://rosettacode.org/wiki/String_interpolation_(included)

marler8997:
I think a big use case for interpolated strings in D is code generation for use with mixins. This makes me lean towards starting with an uncommon character like `$(...)` as opposed to `(...)` or `{...}`. `$` seems to be the most popular character to represent interpolation and it works well for D.

MetaLang:
I really like this, but it seems a little strange to me that 'i' is a prefix while 'c', 'w', and 'd' are all suffixes (now that x has been removed, I think 'i' being a prefix is a special case).

JackStouffer:
This needs to be done with a DIP.

src/dmd/astbase.d
```diff
@@ -4470,6 +4470,7 @@ struct ASTBase
         size_t len;         // number of code units
         ubyte sz = 1;       // 1: char, 2: wchar, 4: dchar
         char postfix = 0;   // 'c', 'w', 'd'
+        char interpolate = 0;
```
jacob-carlborg: Is there a reason this isn’t a `bool`?

test/runnable/istring.d
```diff
@@ -0,0 +1,25 @@
...
+}
+
+string funcCode(string attributes, string returnType, string name, string args, string body)
+{
+    return text(iq{
+    $(attributes) $(returnType) $(name)($(args))
+    {
+        $(body)
+    }
+    });
```
jacob-carlborg: Perhaps add tests with other string literal syntaxes as well.

jacob-carlborg:
> Regarding syntax. I had a short look in this a couple of months ago and here's what other languages use

For Swift it's `"I have \(apples) apples"`.

I like the following syntax:

* Variables: `"I have $apples apples"`
* Expressions: `"I have $(apples + oranges) apples and oranges"`

src/dmd/astbase.d
marler8997: probably should be `ubyte`, but it's to match the `C++` ABI

marler8997: nevermind, I see `bool` is used elsewhere in D/C++ files...switching to `bool`

src/dmd/astbase.d
```diff
@@ -4470,6 +4470,7 @@ struct ASTBase
         size_t len;         // number of code units
         ubyte sz = 1;       // 1: char, 2: wchar, 4: dchar
         char postfix = 0;   // 'c', 'w', 'd'
+        bool interpolate = false; // true if prefixed with 'i'
```
jacob-carlborg: Please make this a Ddoc comment. I know the other ones are not, but we're trying to improve. I would also flesh out the comment a bit, something like: `true if  an interpolated string, i.e. prefixed with 'i'`. Please also use backticks for code in comments.

jacob-carlborg (changes requested):

marler8997:
I can see arguments for and against supporting a shorthand escape syntax like `$id`.  To support it we would need to establish what "grammar node" is supported after a `$`.  Maybe `DotIdtifier` (https://dlang.org/spec/grammar.html#DotIdentifier)?

The good thing is that we can always delay this decision as it is a simple addition that doesn't affect the mechanics of interpolated strings without it.

src/dmd/astbase.d
marler8997: Should I also add ddoc comments to expression.d, or is that supposed to go away at some point?

jacob-carlborg: Hmm, that's a good question. I see that I added the comment to `astbase.d`. But I would say aim to add the comments in other files, i.e. not in `astbase.d`.

test/d_do_test.d
```diff
@@ -497,8 +497,7 @@ string envGetRequired(in char[] name)
     auto value = environment.get(name);
     if(value is null)
     {
-        writefln("Error: missing environment variable '%s', was this called this through the Makefile?",
-            name);
+        writeln(i"Error: missing environment variable '$(name)', was this called this through the Makefile?");
```
marler8997: All changes in this file are just "dog fooding" interpolated strings.

JackStouffer:
>I can see arguments for and against supporting a shorthand escape syntax like $id. To support it we would need to establish what "grammar node" to use. Maybe DotIdentifier (https://dlang.org/spec/grammar.html#DotIdentifier)?

Please, let's stop the syntax discussion and move this proposal into a DIP. 

Firstly, I really don't want you to do a ton on work on this and then have to change a lot or throw it away due to feedback. The DIP review fleshes-out potential pitfalls in design and implementation, and also involves more community opinions. Secondly, I'd like to see some justification for this addition, because I'm really not sold that this provides enough usefulness over `writefln` or the proposed dump [1] to warrant inclusion.

[1] https://github.com/dlang/phobos/pull/4318 if it's moved to the suggested alias parameters approach it can be merged.

jacob-carlborg:
> I can see arguments for and against supporting a shorthand escape syntax like $id. To support it we would need to establish what "grammar node" to use. Maybe DotIdentifier (https://dlang.org/spec/grammar.html#DotIdentifier)?
The good thing is that we can always delay this decision as it is a simple addition that doesn't affect the mechanics of interpolated strings without it.

What do template instantiations use, the shorter syntax, i.e. `foo!bar`?

marler8997:
> What do template instantiations use, the shorter syntax, i.e. foo!bar?

A single token.  Which would mean `i"$foo.bar"` would be the same as `i"$(foo).bar"` instead of `i"$(foo.bar)"`.

marler8997:
> Please, let's stop the syntax discussion and move this proposal into a DIP.

DIPs is where good work goes to die.  This is a simple addition and if we can get it in without the red-tape/formality then we would be saving alot of people's time/effort to work on other things.

> Firstly, I really don't want you to do a ton on work on this and then have to change a lot or throw it away due to feedback

Telling me you don't want me to do a ton of work, and then telling me I need to do a DIP are two completely contradictory statements :)

> Secondly, I'd like to see some justification for this addition, because I'm really not sold that this provides enough usefulness over writefln or the proposed dump [1] to warrant inclusion.

Interpolated strings are pretty nice in some use cases.  Code generation is a great example which makes it a perfect complement with D's mixin.  Most other modern languages have also added support for string interpolation and being able to say that D has it as well is a huge plus, along with the fact that D's version has NO RUNTIME OVERHEAD making it better than everything else currently out there!

marler8997:
inTUPLEated strings!! :)

JackStouffer:
>This is a simple addition and if we can get it in without the red-tape/formality then we would be saving alot of people's time/effort to work on other things.

That's what many people thought when they added https://github.com/dlang/dmd/pull/6589. Turns out even "simple" changes need more eyes on them. Let's not do that again.

I understand why some people have lost faith in the DIP process, but Mike has been working with Walter and Andrei on the DIPs behind the scenes, I assure you. And it does help your case to already have an implementation ready. I'm not saying that this should be closed or anything. What I'm saying is we have a process in place to approve language additions in order to weigh the benefits and the costs and to work out any issues. Let's use that system to our advantage.

marler8997:
If Walter/Andrei think a DIP is required then so be it.  However, you're muddying the waters and getting in the way of developers by asserting they need to do more work than is currently being asked for.  Work that arguably has little to no value and IMO overkill for this particular change.  I don't mind you suggesting that a DIP could be useful, but just because you think something MUST have a DIP doesn't mean that it should.

You opinion is encouraged and welcome, but asserting your opinions on others derails the conversation and distracts from getting work done.  Leave that decision up to Walter/Andrei.

JinShil:
@WalterBright and @andralex, Can you please explicitly state if this requires a DIP or not?

Marking this as "Need Approval" and assigning to @WalterBright and @andralex in an effort to invoke a decision.

timotheecour:
* How is `$` itself escaped in each string syntax? Couldn't find it discussed in PR description or unittests:

how about using `\$`:
```
auto apples=10;
writeln(i"I have $(apples) apples\$"); =>I have 10 apples$
writeln(i"I have $(apples) apples\\\$"); =>I have 10 apples\$
```
would that be the same with iq{} and i``?

* would it make more sense to have:
```
i""
ir""
iq{}
i``
```
instead of:
```
i""
iq{}
i``
```
?

marler8997:
The \$ can conflict with the string escape sequences. `$$` works fine. You can currently do it with $("$") too, but the shorthand is probably worth a bit of extra code to support :)

jacob-carlborg:
If we’re satisfied with only supporting `$(foo)` and not `$foo` then I think a single `$` should work without requiring any escaping. That’s how it works in Ruby.

timotheecour:
no, I think we should at least leave the door open for supporting $foo meaning $(foo) and $foo.bar meaning $(foo.bar) (ie `DotIdentifier`) in a future PR.

marler8997:
We would still want something like `$$` in both cases so you can represent `$(` in your string.

timotheecour:
I'd really like for this PR to move forward without having to go through yet another DIP red tape

timotheecour:
```
// IS THE SAME AS
tuple!("a is ", a, " and b is ", b);
```

shouldn't that be:
```
std.meta.AliasSeq!("a is ", a, " and b is ", b);
```
?

ntrel:
```d
int a = 100;
alias seq = i`a + 56 = $(a + 56)`;
alias seq = AliasSeq!(`a + 56 = `, a + 56);
```
`AliasSeq` above cannot accept a runtime expression such as `a + 56` as an argument, so this lowering is inexact (as I see your tests support this).

I do think we need `i"$var"` syntax as it avoids some bracket nesting. I would limit `$var` to an identifier, i.e. `[a-z_][a-z0-9_]+`. Expressions should use curly braces: `i"${var+1}"` - this makes the expression stand out as code more - `(bracketed text)` is quite common in strings, braces are not.

I would like expressions such as `i"${var+1}"` be required to be strongly pure. This is a useful guarantee for translating interpolated strings - changing the order of expressions in a translation (which is often needed) won't have side effects. Translations are a key motivating factor for interpolated strings vs `text("I have ", n, " apples")`.

This pull adds quite a lot of code to dmd. I'm not convinced it's worth supporting interpolation string literals other than `i"$arr[\$-1]"` backslash-escaping (at least for now). Otherwise Wysiwyg strings then have to have escapes: `iq{$arr[$$-1]}` - we would at least need to call them something else other than Wysiwyg.

`generateFunction` is interesting but it might be better to enhance `mixin` instead of having `iq{}` interpolated strings, e.g. (not actually equivalent):
https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1010.md#mixin-identifiers

marler8997:
I'll defer any syntax decisions and if the shorthand `$id` should be supported to Walter/Andrei, but thanks for sharing your opinion. 

> I would like expressions such as i"${var+1}" be required to be strongly pure. 

I'm not sure what benefits this restriction yields.  You mentioned an example about how this can help "translation" but I'm not sure what you meant by that.

> I'm not convinced it's worth supporting interpolation string literals other than i"$arr[\$-1]" backslash-escaping (at least for now). Otherwise Wysiwyg strings then have to have escapes: iq{$arr[$$-1]} - we would at least need to call them something else other than Wysiwyg.

I'm a bit confused by this sentence.  But I'm strongly in favor of supporting wysiwyg strings.  The biggest use case I see for this is code generation.  Try writing `generateFunction` using a double-quoted string, it's ugly :)

Note that the interpolation does not happen until after the string is processed, not while it is being processed.  By the time you interpolate the string, all the escape sequences and modifications are already done.  So a wysiwyg string is still wysiwyg, it's just that after it's created it then goes through interpolation. Also note that you don't have to escape the `$` character in the "code" parts of the string, only the "raw string" parts.  So you would write `iq{$(arr[$-1])}`.  When I write this `iq{$(arr[$$-1])}` I get an error.

wilzbach:
FYI: in the State of D survey string interpolation was the third most-missed language feature with 31% of _all_ respondents missing it:

image

https://rawgit.com/wilzbach/state-of-d/master/report.html

marler8997:
Those top 5 items are the perfect list of the things I miss in D!

ntrel:
> I'm not sure what benefits this restriction yields.

Here's an example:
```d
    text(translate(i"$x and then $y"))
```

Imagine evaluating `x` has side effects, changing the value of `y`. If the translation is `i"$y after $x"`, then an unintended value for `y` will be put into the string. Requiring strong purity would forbid this. Actually even weak purity would prevent this case, but not the case `i"${x++} becomes $x"` - although this is gratuitous anyway, and such translations could be machine checked for inconsistent ordering by a simple utility.

>> Otherwise Wysiwyg strings then have to have escapes: iq{$arr[$$-1]} - we would at least need to call them something else other than Wysiwyg.

Note: here `[$$-1]` is part of the string literal, not the interpolated expression (i.e. `iq{${arr}[$$-1]}`).

> I'm a bit confused by this sentence.

If you see `$$`, you don't get `$$`. (Same applies for `$`). Therefore it's no longer WYSIWYG. I suppose it could just be accepted as an exception though.

> Try writing generateFunction using a double-quoted string, it's ugly :)

```d
return text(i"
        $(attributes) $(returnType) $(name)($(args))
        {
            $(body)
        }
    ");
```
Maybe I'm missing something?

marler8997:
> Imagine evaluating x has side effects, changing the value of y. If the translation is i"$y after $x", then an unintended value for y will be put into the string. Requiring strong purity would forbid this. Actually even weak purity would prevent this case, but not the case i"${x++} becomes $x" - although this is gratuitous anyway, and such translations could be machine checked for inconsistent ordering by a simple utility.

I'm still not seeing the benefit.  Remember, an interpolated string is just another way of writing a tuple.  Would you put the same restriction on all tuples?  Every expression in every tuple needs to be pure? I don't see any problem with the example you gave.  If a program wants side effects why disallow this?

> If you see $$, you don't get $$. (Same applies for $). Therefore it's no longer WYSIWYG. I suppose it could just be accepted as an exception though.

It's a WYSIWYG string that's being "interpolated".  You don't get the full string as you see it anyway.  It gets turned into a tuple.  Think of the `i` prefix as an operator, that's taking the WYSIWYG string and turning it into a tuple.  It's still a WYSIWYG string, it's just getting passed to an operator that modifies it.

As for the double-quoted string example.  I forgot that double-quoted string literals can use raw newlines so I guess it works fine.  Still, I don't see any reason not to support WYSIWYG strings.

ntrel:
> Remember, an interpolated string is just another way of writing a tuple. Would you put the same restriction on all tuples?

Are we talking about runtime tuples or CT sequences? I'm assuming the latter.

First, no, because CT sequences don't have any particular link with translations. Interpolated strings are required by translators unless they fall back on printf format strings, so there is a link here.

Second, as I mentioned earlier, `x++` on a runtime variable is not compilable as an argument to `AliasSeq`. So what is the actual lowering - wrap the runtime expression in a property?

> I don't see any problem with the example you gave.

The problem is that translators should not be able to change program behaviour. Translators might want to reorder expressions, especially as they would have access to the expressions with an interpolated string. They would assume it's OK to rearrange them. A simple build utility could detect that expressions have been reordered, but I was thinking that requiring expression purity would allow them to safely be reordered. Some translation languages might need expression reordering for good translations. Also integrating yet another build tool will in practice not be done usually. I thought it was worth bringing up, but maybe no one else thinks it's a significant problem.

ntrel:
>  You could use the same argument you gave to say that you shouldn't be able to pass WYSIWYG strings to a functions like capitalize because they "change" the string thus making it "no longer WYSIWYG".

No, because that is clearly a function call. `ir""` looks like a string literal. It's a sequence literal defined with a WYSIWYG string, but even the string literal components of the resulting sequence are not WYSIWYG strings.

marler8997:
@ntrel  are you on slack?  Let's discuss on there.

ntrel:
I don't have an invitation (AFAICT) ;-)

marler8997:
I'll invite you, I just need your email, you can post it here or email it to me at `johnnymarler@gmail.com`

ntrel:
Thanks, emailed you.

marler8997:
> No, because that is clearly a function call. ir"" looks like a string literal. It's a sequence literal defined with a WYSIWYG string, but even the string literal components of the resulting sequence are not WYSIWYG strings.

Like I said, `i` is equivalent to an "operator" that operates on the string.   It works the same as a function.  If it wasn't an operator then it would be a function like this:
```D
interpolate(q{ ... })
```

It's jut that using `iq{...}` is less verbose just like operators on less verbose than functions.  The WYSIWYG string is still created as any WYSIWYG string, it's just passed to the internal interpolate function and returned as a tuple.  This is the same thing as passing it to any function, like `capitalize` for example:
```D
enum s1 = capitalize(q{ hello, world });
enum s2 = iq{ hello, world };
enum s3 = interpolate(q{ hello, world }); // would be the same thing if it was implemented as a function
```

ibuclaw (changes requested):
> DIPs is where good work goes to die. This is a simple addition and if we can get it in without the red-tape/formality then we would be saving alot of people's time/effort to work on other things.

Adding a monitor attribute is a simple addition also, but we still need a DIP for it.

It's not a lot to ask for a DIP such as this, and a problem may come up that you didn't think about.

wilzbach:
Can't we implement this under a `--experiment=string` flag?
Rust and Node have tons of experimental flags, e.g.

https://doc.rust-lang.org/stable/unstable-book
https://github.com/thlorenz/v8-flags/blob/master/flags-0.11.md

marler8997:
> Can't we implement this under a --experiment=interpolation flag?

Not a bad idea.  We've done similar in the past as well.  If Walter or Andrei is ok with this I'm more than willing to go along.

ibuclaw:
> Can't we implement this under a --experiment=interpolation flag?

There's been an aversion towards adding new compiler flags recently. Not saying no though.

wilzbach:
> There's been an aversion towards adding new compiler flags recently. Not saying no though.

Well clearly `-dip25`, `-dip1000`, `-dip1008` show that this doesn't scale, but I was thinking about something grouped like the existing `-transition`.
Heck, if there's such a strong aversion against new flags, even `-transition=interpolation` would work fine.

wilzbach:
BTW @marler8997 `d_do_test` seems to segfault under LDC here:

```
../generated/linux/release/64/dmd -conf= -m64 -fPIC -g -unittest -run d_do_test.d -unittest
make: *** [test_results/d_do_test] Segmentation fault (core dumped)
```

(I restarted, maybe it's just spurious)

marler8997:
Ok, added `-transition=interpolate` option to enable this feature.  It was pretty easy to put the feature behind a flag.  It's checked once in the lexer and once in the parser.

JinShil:
> Ok, added -transition=interpolate option to enable this feature.

This doesn't change the fact that we still need a policy decision from @WalterBright or @andralex about whether or not this feature requires a DIP.

skl131313:
Just how much code is actually usable in the string?

You have operators like so:

```D
int a = 100;
writeln(i"a is $(a)");
writeln(i`a + 56 = $(a + 56)`);
writeln(iq{ $(a) * $(a) = $(a*a) });
```

Just how much code is actually valid since operators reduce to function calls, what about the below case?

```D
int var = 100;
writeln(i`var is $(mixin("var"))"`);
```

marler8997:
Yup that works (tested it just to make sure).  An interpolated string is just another way of writing a tuple.  Currently I don't think D has a way of creating tuples like this, but if you imagine a builtin function `tuple` existed, your example would be equivalent to this:
```D
int var = 100;
writeln(tuple("var is ", mixin("var")));
```

timotheecour:
@skl131313 you can just try out yourself simply with:
`dub run digger -- build "master +dmd#7988"`

marler8997:
That's cool. Also don't forget the -transition=interpolate parameter if you try it.

andralex:
Thanks for this work. @WalterBright and I discussed it. The first question is what would be a viable library-based solution. Far as I can tell this is as far as a library would go:

```D
int a = 100;
writeln(mixin(i!"a is $(a)"));
writeln(mixin(i!`a + 56 = $(a + 56)`));
writeln(mixin(i!q{ $(a) * $(a) = $(a*a) }));
```
```D
import std.conv : text;

int a = 32;
assert("a is 32" == text(mixin(i!"a is $(a)")));

// same as
assert("a is 32" == text(tuple("a is ", a)));
```

```D
string generateFunction(string attributes, string returnType, string name, string args, string body)
{
    import std.conv : text;
    return mixin(i!q{
        $(attributes) $(returnType) $(name)($(args))
        {
            $(body)
        }
    }).text;
}
mixin(generateFunction("pragma(inline)", "int", "add", "int a, int b", "return a + b;"));
assert(100 == add(25, 75));
```

Pros of a library-based solution:
* No need for a language change
* Less need to invent syntax
* Can be readily made available on dub for experimentation
* Allows us to accumulate experience cheaply
* Its popularity will inform future language improvements (maybe we don't need "i" for interpolated strings, but a nicer mixin syntax, e.g. make it UFCS-able a la `i!"stuff".mixin` or make it an operator (no parens needed), or even create an operator such as `##` that mixes stuff in etc.
* No trouble of a language feature depending on a standard library feature (one would need to import `std.typecons` in order to use interpolated strings, which is surprising)

Cons:
* Syntax is more awkward. This is pretty much the only con but it's a pretty big one given that the entire feature is saving syntax over the more awkward mix of string literals, commas, and expressions.

At any rate, we should experiment with this as a library before considering building it into the language.

wilzbach:
@andralex FYI: We had a library solution for more than three years and imho this experiment concluded that this doesn't work:

1) The syntax is really awkward and inconvenient: https://github.com/abscissa/scriptlike#string-interpolation
-> an option here would be to allow auto-mixins, i.e. mixin's that don't require the user to type `mixin`, but I doubt that you would accept sth. like this
2) The library solution (even though it existed for a long time) it's not very known nor used (I haven't seen anyone using this so far)
(-> yes, adding this to Phobos might help)
Also since 2.079 there's `substitute` and often I just use `"{template}".substitute("{template}","<replacement>")`, but that's just an ugly workaround and also has horrible CTFE performance.
3) A mixin-based library solution really slows down the build when you use this more than a few times (built-in string interpolation allows a lot more optimizations)
4) Almost all other modern languages have built-in string interpolation (see the list above) - on RosettaCode I count __82__ languages having string interpolation.

> Can be readily made available on dub for experimentation

It has been on dub for more than three years.
Also there are many cases where you don't want to (or can't) import a dub package just to have interpolated strings (though `std.experimental` might solve this).

> Allows us to accumulate experience cheaply

I get the point that you want to move everything that's possible to do in a library away from the compiler, but the "experience" is quite different here. If the current "experience" with dub would be great, it wouldn't have been on the number 3 top-priority item in the State of D survey with 31% of _all_ respondents mentioning missing string interpolation as "negatively affecting their D experience".

> Its popularity will inform future language improvements

As mentioned, this is a _very popularly_ requested feature. Only tuples and named arguments are more popular.

> No trouble of a language feature depending on a standard library feature (one would need to import std.typecons in order to use interpolated strings, which is surprising)

That's just an implementation detail and I'm pretty sure that with a bit of more work this could be avoided.
After all, I think this PR is just a PoC.

JackStouffer:
I'm not seeing the issue with script-like's implementation. `mixin(interp!"The number is ${num}!"));` is really not that much worse than `i"The number is ${num}!";`. Longer? Yes. A little more cluttered? Definitely. More difficult? Hardly.

>The library solution (even though it existed for a long time) it's not very known nor used (I haven't seen anyone using this so far)

This is honestly evidence against the desire for the feature more than anything else.  If a workable solution to this issue has existed in dub for years and no one who wants interpolation cares enough to use it, that's an indication that the feature wasn't that important. Either that, or it's an indicator about dub's popularity. If dub was such a blocker, than the code for `interp` can be copy/pasted into your own code base easily.

Someone please explain why this is a big sticking point for some people. It's such a minor feature that can already be covered by existing code.

MetaLang:
> Someone please explain why this is a big sticking point for some people.

For me personally, it drives me insane that `mixin` for strings requires parentheses. The fewer levels of nesting required, the better. A lot of the time I create mixin template wrappers just for the more convenient syntax:
```
template generateFunction1(...) { ... }
mixin template generateFunction2(...) { mixin(generateFunction1!(...)); }

struct Test
{
    mixin(generateFunction1!(...)); //naughty
    mixin generateFunction2!(...);  //nice
}
```
I've always wished we could use the mixin template syntax for both types of mixins. That would make the `interp` solution more ergonomic, at least for me:
```
int num = 21;
writeln(mixin interp!"The number ${num} doubled is ${num * 2}!");

// Output: Empty braces output nothing.
writeln(mixin interp!"Empty ${}braces ${}output nothing.");

// Output: Multiple params: John Doe.
auto first = "John", last = "Doe";
writeln(mixin interp!`Multiple params: ${first, " ", last}.`);
```
Of course, there's still the issue that `interp` itself is slow and causes a lot of template bloat.

Ultimately, string interpolation isn't that high on my list of things I want added to D, but it's one of those quality of life features that's very annoying to live without once you've used another language that _has_ it (and is one of those things I _would_ use all the time if D had a better solution for it). Look at how many comments we get from C++/Java/C# programmers who complain about how D has ruined their primary language for them. A large part of that is due to these little ergonomic features, I suspect.

marler8997:
@andralex 

The biggest con that this PR introduces is requiring a new string literal prefix characer `i` be added to the language syntax.  I did have one correction in your "CON" list, interpolated strings don't require you to import anything. They DO NOT require you to import `std.typecons`.  The tuple is created internally and has no dependency on any external library.  

There are 3 cons to the library solution.  The first you identified (awkward syntax).  However, this could be mitigated if a better syntax for "mixins" was implemented (like you suggested).

The second con is performance.  Because of D's powerful template/mixin metaprogramming, the need for a fast and pretty syntax for code generation becomes much more justified.  By incorporating interpolated strings directly into the parser, you don't have to choose between fast code and pretty code...you can have both.  If we implement a library solution, then I anticipate interpolated strings being rejected for much of the standard/library and druntime because it will kill compile-time performance (even though it should have the same runtime performance).

The third con is error messages.  The way that interpolated string support has been added guarantees that the strings being interpolated must be raw string literals.  This allows error messages to point directly to source code, rather than pointing to a "mixin" location.

To summarize, the trade off is having to support a new string literal prefix character `i` for concise syntax, fast performance, and nice error messages.  I'm leaning toward this feature being worth the cost, but I also don't think the right answer is black and white.  To me, the 3 benefits, concise syntax/performance and error messages hold alot of weight, but they may not be as important to others.

andralex:
Thanks for the replies. 

Regarding "the library solution doesn't work" - we have the paradoxical situation whereby the poll indicates a healthy request for interpolation, whereas the results in practice are that people can't seem to bring themselves to use a library-based solution as a workaround. I'm no psychologist but it seems a stretch to conclude that a language feature is the answer. What I can offer is a bundle of anecdotes from the C++ community - `enable_if` and `enable_if_t` (there are others similar, too). Boost has had `enable_if` for a while, which turned out to be immensely useful. (No surprise, they really are template constraints that are so useful in D.) So C++11 introduced `enable_if`, used as such: `enable_if<condition, T>::type`. Since C++14, `enable_if_t` has been introduced, which is a mere convenience allowing the alternative spelling `enable_if_t<condition, T>`. The rationale is that `enable_if` turned out to be so heavily used, the sugar was deemed worthy of introduction in the standard library. This would be solid precedent: popular library solution gets standardized and is subsequently even more popular so it gets supplemental sugar also standardized. This kind of arc is not present with string interpolation.

Regarding "really awkward and inconvenient" - this is only reframing my already strong mention "...the only con but it's a pretty big one..." as a vivid argument but without adding factual value. We'd do good to stay away from such lest we get into a contest of who can produce the most flowery argument. The example given is similar to the ones already presented here by @marler8997 and reworked by myself.

Regarding "an option here would be to allow auto-mixins, i.e. mixin's that don't require the user to type mixin, but I doubt that you would accept sth. like this" - this paints the matter as an arbitrary ruling on my side, so I feel compelled to refute this. In reality it's about our values - we want D to adhere to good programming language design theory and practice. If we allow templates to use `mixin` internally and capture names from their environment, we break [hygienic expansion](https://en.wikipedia.org/wiki/Hygienic_macro), which would allow a variety of breakages of encapsulations. We do care about encapsulation and we want to avoid non-hygienic expansion (though we allow it in very very very limited cases such as `__FILE` etc in default arguments). So it's not that I don't like it.

Regarding the build speed - indeed, a library solution would be slower. This should be mentioned as a disadvantage. (It seems the best angle here is focusing on speeding up template processing. That would accelerate a bunch of facilities, so the win scales up nicely.)

Regarding "82 languages have this" - this would need a good amount of additional information. 82 out of how many? How are they related to D's values? How are languages closest to D doing wrt string interpolation? (C no, C++ no, Rust [no](https://stackoverflow.com/questions/21324657/does-rust-support-ruby-like-string-interpolation)...) Etc. We really need to get scientific about constructing arguments on this kind of stuff.

Regarding "31% of all respondents mentioning missing string interpolation as negatively affecting their D experience": this is good data, thanks for the survey! I recall there were some 540 respondents, so the level of confidence on the 31% is small with so few samples. It would be great if that were accurate because with only a small effort we could improve the lot of D users. 

Regarding popularity: the item has been popularly asked for, but not the library implementation not popularly used. It is seldom discussed in the forums. It is not clear where this leaves matters.

andralex:
@marler8997 thanks for the explanation. OK, so the tuple is not the one in std.typecons.

I'm not sure how the error messages play out. The interpolation location is the same, whether done by a built-in facility or by `mixin`.

marler8997:
Oh another thing.  As far as I know, D does not have a way to create tuples in the way that this PR has done for interpolated strings...so a library solution that could do this would require new tuple semantics to be built into the language.

As for the error messages, consider the following example. `generatePage1` uses this PR, `generatePage2` uses a library solution:
```D
import std.stdio, std.conv;

void main()
{
    generatePage1("MyCoolPage");
    generatePage2("MyCoolPage");
}

void generatePage1(string title)
{
    writeln(i`
<html>
 <head>
  <title>$(title)</title>
 </head>
 <body>
  <h1>$(title)</h1>
  <pre>
1 + 3 = $(1 _ 3)
  </pre>
 </body>
</html>
`);
}

void generatePage2(string title)
{
    writeln(mixin(interpolate!`
<html>
 <head>
  <title>$(title)</title>
 </head>
 <body>
  <h1>$(title)</h1>
  <pre>
1 + 3 = $(1 _ 3)
  </pre>
 </body>
</html>
`));
}
```

There is a small error in this code.  With this PR I get the following error:
```
example.d(19): Error: invalid expression '1 _ 3' inside interpolated string
```

With a library solution you would get an error inside "mixin code" so it would look something like this:
```
example.d-mixin-28(36): Error: ...
```

The error messages get much worse as you add more complex logic to build up your string.  For example, if you had a "generateHeader' function and a "generateBody" function, then finding where your mixin code actually is becomes much more cryptic.

JinShil:
> Its popularity will inform future language improvements (maybe we don't need "i" for interpolated strings, but a nicer mixin syntax, e.g. make it UFCS-able a la i!"stuff".mixin or make it an operator (no parens needed), or even create an operator such as ## that mixes stuff in etc.

Is the keyword `mixin` even required to instantiate a mixin template?  Doesn't the template instantiation operator, `!`, provide everything need to mixin a mixin template?

FYI: This same issue came up at https://forum.dlang.org/post/mqveusvzkmkshrzwsgjy@forum.dlang.org when exploring a library implementation for properties.

marler8997:
> Is the keyword mixin even required to instantiate a mixin template?

Not that I know of, but you need a `mixin`, not just a `mixin template`.

timotheecour:
> Is the keyword mixin even required to instantiate a mixin template?

mixin template Foo()
{
    int x = 5;
}

mixin Foo;

JinShil:
What I mean was is the `minin` keyword necessary to instantiate a mixin template?  In other words, could the language be modified to not require it?

marler8997:
> What I mean was is the minin keyword necessary to instantiate a mixin template? In other words, could the language be modified to not require it?

Yes, but [copy/paste Andre's comment]...

> Regarding "an option here would be to allow auto-mixins, i.e. mixin's that don't require the user to type mixin, but I doubt that you would accept sth. like this" - this paints the matter as an arbitrary ruling on my side, so I feel compelled to refute this. In reality it's about our values - we want D to adhere to good programming language design theory and practice. If we allow templates to use mixin internally and capture names from their environment, we break hygienic expansion, which would allow a variety of breakages of encapsulations. We do care about encapsulation and we want to avoid non-hygienic expansion (though we allow it in very very very limited cases such as __FILE etc in default arguments). So it's not that I don't like it.

andralex:
Regarding the ability to generate several values as a single expression, I just tried this:

```d
import std.experimental.scripting;
void main()
{
    writeln(mixin(`tuple("a", 2).expand`));
}```

which prints `a2`, meaning the pack-unpack sequence `tuple(e1, e2,).expand` is an effective way to represent multiple expressions as a single expressions.

How about this: we develop `interp` (or even `i`!) as a library facility in `std.experimental.typecons` and take it from there? @marler8997 since you've designed it as a compiler facility already I assume there's a good amount of intellectual overlap, would this be something you'd be interested in doing?

**Comment by marler8997** [comment]
Yeah I can look into this.  I know @timotheecour has an implementation as well, we could probably use that as a starting point.

**Comment by wilzbach** [comment]
> Regarding "82 languages have this" - this would need a good amount of additional information. 82 out of how many? How are they related to D's values? How are languages closest to D doing wrt string interpolation? (C no, C++ no, Rust no...) Etc. We really need to get scientific about constructing arguments on this kind of stuff.

Not so sure whether we want to consider C as the base line for modern languages?
I posted a list [above](https://github.com/dlang/dmd/pull/7988#issuecomment-371451743):

Bash, Dart, E, Haxe, Julia, Groovy, PHP, C#, Elixir, Ruby, JavaScript, Nim, Python, Scala, Swift, ...
I think Rust is the only modern language that I know that doesn't has this yet, but they are discussing it too, e.g. https://github.com/rust-lang/rfcs/issues/1250
(and Rust does have named parameters, so their library-version of format isn't too bad)

> we have the paradoxical situation whereby the poll indicates a healthy request for interpolation, whereas the results in practice are that people can't seem to bring themselves to use a library-based solution as a workaround.

Well, I think because it's not a life breaker and just a constant annoyance and in most cases `"text.text"~mycode.to!string~"text.text"` does work too (it's just not as clean as string interpolation would be).

> Regarding "31% of all respondents mentioning missing string interpolation as negatively affecting their D experience": this is good data, thanks for the survey! I recall there were some 540 respondents, so the level of confidence on the 31% is small with so few samples. It would be great if that were accurate because with only a small effort we could improve the lot of D users.

Not everyone answered the question (it was optional, so we even have fewer replies).
Though does it really matter whether 20% or 40% of all respondents mentioned this as a pain point? I think (even without the survey) we all agree on that the current status with neither a language nor library solution readily available is sub optimal.We just don't agree on what the solution should be ;-)

**Comment by marler8997** [comment]
I don't mind implementing the library solution, if nothing else would be good to have as a comparison.

**Comment by wilzbach** [comment]
> Yeah I can look into this. I know @timotheecour has an implementation as well, we could probably use that as a starting point.

See also: https://github.com/Abscissa/scriptlike/blob/master/src/scriptlike/core.d#L141

(though it's not Boost licensed, but maybe @Abscissa could help if that would be a problem)

**Comment by Abscissa** [comment]
On March 23, 2018 7:10:15 PM EDT, Sebastian Wilzbach <notifications@github.com> wrote:
>> Yeah I can look into this. I know @timotheecour has an implementation
>as well, we could probably use that as a starting point.
>
>See also:
>https://github.com/Abscissa/scriptlike/blob/master/src/scriptlike/core.d#L141
>
>(though it's not Boost licensed, but maybe @Abscissa could help if that
>would be a problem)

I'm more than willing to relicense anything to Boost for dlang/phobos/etc.

**Comment by marler8997** [comment]
@andralex 

https://run.dlang.io/gist/marler8997/f24c0ed63f078c99332d629b5f5314aa?compiler=dmd

I copied my prototype implementation...but I can't get a simple example to work as expected (see link).  I believe it is because of the "tuple problem" which is that D's current semantics have no way of creating and returning "real tuples" that "collapse" when passed to function calls.

The problem is simple...create the string "a is 42" from the interpolated string "a is $(a)"...none of the following worked:
```D
    string attempt1 = text(mixin(interp("a is $(a)")));
    writefln("attempt1 '%s'", attempt1); // NOPE
    //assert(attempt1 == "a is 42"); // Fails

    string attempt2 = text(interp("a is $(a)"));
    writefln("attempt2 '%s'", attempt2); // NOPE
    //assert(attempt2 == "a is 42"); // Fails

    string attempt3 = text(mixin("a is $(a)".interp));
    writefln("attempt3 '%s'", attempt3); // NOPE
    //assert(attempt3 == "a is 42"); // Fails
```

jacob-carlborg:
> Regarding "the library solution doesn't work" - we have the paradoxical situation whereby the poll indicates a healthy request for interpolation, whereas the results in practice are that people can't seem to bring themselves to use a library-based solution as a workaround.

In my opinion the point of string interpolation is to have a short and convenience syntax. If it's not, I don't see much value over using `format` or manually concatenating strings. I think `mixin(i!"foo $(bar)");` is too verbose. To solve that in a library solution AST macros or some form of auto-mixin would be required. If AST macros where available the following syntax could be used `i!"foo$(bar)"`, which is just one character more than the built-in solution. But I know that many don't like AST macros.

> Regarding "an option here would be to allow auto-mixins, i.e. mixin's that don't require the user to type mixin, but I doubt that you would accept sth. like this" - this paints the matter as an arbitrary ruling on my side, so I feel compelled to refute this. In reality it's about our values - we want D to adhere to good programming language design theory and practice. If we allow templates to use mixin internally and capture names from their environment, we break hygienic expansion, which would allow a variety of breakages of encapsulations.

In my opinion string mixins are non-hygienic regardless if the `mixin` keyword is required or not.

marler8997:
I don't want my comment to get lost.  @andralex, based on the code I posted in the link above, I believe we have determined that a library solution for "interpolated strings" is not viable without adding new tuple support to the language.  That is unless you or someone else can modify the code I posted to work?

andralex:
@marler8997 I assumed we'd use `std.experimental.tuple` and `std.experimental.Tuple`, not other artifacts. Did you try them? Per my example, using `.expand` seems to work properly.

andralex:
>In my opinion the point of string interpolation is to have a short and convenience syntax. If it's not, I don't see much value over using `format` or manually concatenating strings. I think `mixin(i!"foo $(bar)");` is too verbose. To solve that in a library solution AST macros or some form of auto-mixin would be required. If AST macros where available the following syntax could be used `i!"foo$(bar)"`, which is just one character more than the built-in solution. But I know that many don't like AST macros.

Ack. Clearly the relative overhead is large with short strings. My hope is that a library for interpolation will do wonders for code generation, which is very awkward right now. E.g. probably `bitfields` would look a lot better.

>In my opinion string mixins are non-hygienic regardless if the `mixin` keyword is required or not.

With `mixin` the lack of hygiene is opt-in; if we allow it without the keyword it's... nobody knows. Big difference. Night and day.

marler8997:
> @marler8997 I assumed we'd use std.experimental.tuple and std.experimental.Tuple, not other artifacts. Did you try them? Per my example, using .expand seems to work properly.

Yes, the interactive online example I posted uses `std.typecons.tuple`.  I also tried `std.experimental.typecons.tuple` and got the same result.

https://run.dlang.io/gist/marler8997/f24c0ed63f078c99332d629b5f5314aa?compiler=dmd

Your example seems to be using a different design for interpolated strings.  It seems like you're trying to implement a library function that returns a single string, instead of a tuple.  Such a design would carry with it alot of overhead and also not allow functions to operate on the interpolated "parts", since all they would get is the resulting string.

marler8997:
> My hope is that a library for interpolation will do wonders for code generation, which is very awkward right now. E.g. probably bitfields would look a lot better.

https://github.com/marler8997/interpolated_strings/blob/master/phobos_example.d#L36

andralex:
@marler8997 I can't see your code for some reason, but my experiments indicate that using `tuple(expressions).expand` should work properly. Here's a longer example. It doesn't deal with parsing the string, instead it just expands some variable `x` inside the string:

```d
import std.experimental.scripting;

template interp(string s)
{
    enum interp = "tuple(`" ~ s[0 .. 2] ~ "`,x,`" ~ s[2 .. $] ~ "`).expand";
}

void main()
{
    pragma(msg, interp!"[[]]");
    int x = 42;
    writeln(mixin(interp!"[[]]"));
    auto t = tuple(mixin(interp!"===="));
    writeln(t);
}
```

The example prints:
```
tuple(`[[`,x,`]]`).expand
[[42]]
Tuple!(string, int, string)("==", 42, "==")
```

which indicates that the expansion works correctly - and indeed not to `string`, which would be inefficient.

marler8997:
@andralex 
Thanks...I see the missing "secret sauce" I needed was the `.expand` function.  Ok, back on track.

marler8997:
Phobos PR (WIP) will be here: https://github.com/dlang/phobos/pull/6339/files

wilzbach:
BTW one pro point for built-in string interpolation that no one mentioned so far is that IDEs can support this a lot better (think auto-completion or more simple things like syntax highlighting).
`mixin(interp!"keyword=${keyword}"` won't use syntax coloring (`mixin(text(mixin(interp!q{...})))` only works for a small subset of the use cases where you actually want to mixin the interpolated string and even then not for the interpolated expressions.)

Also here's a screenshot of what Google Chrome provides out of the box:

image

Yep not even an IDE - but still auto-completion for string interpolation.

rracariu:
Watching this PR with interest for a while.
Looks that it got side tracked to "let's build a library solution" mantra.

Some points here, some are overlapping with other's views:
- library solution existed for a while now, there are dub packages and the D forums have several threads on how to do it/use it as library (user code)
- I'd say that there are an irrelevant number of users of those library solutions, and I'll explain why:
- string interpolation is a language feature, from all languages I know of - they are build-in, you learn about them when studying the language, one time, just need to remember the syntax. Done!
- the reasons for having it build-in are ergonomics: less typing, less clutter and less details need to remembered (like what imports are need it). But the biggest one is aesthetics, all Dlang library proposals are UGLY! This single point has killed, and will kill, any interest in a library solution, unless changes for mixins or some form of macros are added to the language.
- there is of course a missed opportunity for better IDE support and easier to understand error messages when going to a library solution using current Dlang semantics and syntax.

Also, with current proposal for a library solution, there is no concern to betterC usage, for example the phobos PR uses exceptions, I'd say that any library solution must work in betterC mode!

IMHO, this PR should go to a DIP and be voted. I see no serious issues with it, except the requirement for the `i` (I would preferred it without it, but it's a small compromise to bet by), and looks that it addresses the performance concerns some brought forward. Any other way is a waste of time and talent.

Thank you!

ntrel:
>> ir"" looks like a string literal. It's a sequence literal defined with a WYSIWYG string

> i is equivalent to an "operator" that operates on the string. It works the same as a function. If it wasn't an operator then it would be a function like this:
> interpolate(q{ ... })

I would like the tuple literal to look less like a string literal. I had an idea, as `__interpolate(StringLiteral)` is indeed verbose, adding more bracket nesting. We could have `__interpolate!StringLiteral`, or shorter still, make interpolation a form of mixin: `mixin StringLiteral`:
```d
writeln(i"a is $(a)");      // looks like a string, i is not that intuitive
writeln(mixin "a is $(a)"); // looks like an operation, we can google mixin
```
Note: For hygiene, you can't use `mixin EnumString`, it has to be a string literal. This also avoids confusion with template mixins (`mixin Identifier;`).

(This avoids the problem with Wysiwyg interpolation literals having dollar escaping. E.g. `ir"$$"`, which becomes `"$"`, which is not Wysiwyg. With `mixin`, the escaping is more obviously part of the operation, not the string literal. I accept this is a somewhat aesthetic argument).

If `mixin` is not liked, I think using `$"..."` is better than `i"..."`, as at least the dollar can be thought of as an operator rather than the `i` appearing to denote a special kind of string, as all other char prefixes do. Googling when unfamiliar would be "dollar string". C# uses `$"..."` for interpolation. But `mixin` is clearer that something special is produced, not a string.

skl131313:
It's not that much more difficult to google what "i" does when it is prefixed with a string.

https://www.google.ca/search?q=C%2B%2B+string+starts+with+r

Using `mixin` for that changes the meaning of what mixin does. I'd much rather see `mixin "int a;";` actually do the same thing as `mixin("int a;")` than mean an interpolated string.

ntrel:
@skl131313 
>  I'd much rather see mixin "int a;"; actually do the same thing as mixin("int a;")

Would that allow enum strings? If not then it's less useful than the status quo. If yes, then there would be no way of knowing at the use site whether a string is being mixed in or a template mixin is being instantiated. It also does not compose as an expression, it's only useful as a statement. Anyway, I reiterate, that `$"..."` is better than `i"..."` because the latter looks like a string literal. The dollar can be thought of as a tuple literal operator.

*/all* I think `text(i"...")` is something that may deserve extra support, particularly as it requires importing std.conv.text. Given this pull is a language feature, maybe the tuple result can implicitly convert to a string, using `text`? It could even go further and do this whenever the tuple is not passed to a variadic function or aliased.

skl131313:
> Would that allow enum strings? If not then it's less useful than the status quo. If yes, then there would be no way of knowing at the use site whether a string is being mixed in or a template mixin is being instantiated.

What benefit is there knowing whether a mixin template is used or a string? It doesn't make that big of a difference, you can do a string mixin within a mixin template. It depends on how it is evaluated but it even still, when I see a mixin I go to see what is being mixed in irregardless of what it actually is.

```D
string genSomeString(T)()
{
    return "...";
}

mixin template mt(T)
{
    mixin(genSomeString!T);
}

void main()
{
    mixin mt!int;
}
```
> The dollar can be thought of as a tuple literal operator.

The dollar is already used as the range length operator.

ntrel:
@skl131313:
> What benefit is there knowing whether a mixin template is used or a string?

It's not a big benefit, but it does help a bit in working out which symbol definition you're using.

> The dollar is already used as the range length operator.

People would not confuse `$StringLiteral` with `Expression[$]`.

skl131313:
> It's not a big benefit, but it does help a bit in working out which symbol definition you're using.

That is pretty much useless, using a good code editor you can just instantly go to the symbol and see what it is.

futscdav:
It's really sad that this has developed in the way it has developed so far. Several very good reasons why this feature should be compiler-supported have been laid out: efficiency, ergonomics, aesthetics, playing nice with code generation, error messages, ease of integration with IDEs, ability to rely on it inside phobos implementation..., while basically **one reason against** it was given: we need to change the language.

So what? This is not a breaking change, you need to actively opt-in to use the feature by prefixing your strings with i, this breaks NO old code. This should be a no-brainer accepted feature, proven to be useful and used in many other languages. Instead, there is resistance, likely from people who have never used it  before and cannot see the benefit. D has prided itself on implementing language features that make it easier to use for day-to-day programming, this is such a feature.

Abscissa:
I strongly agree with @futscdav here, and it's a general pattern that's been chafing me for some time: The whole key reason D was ever good in the first place was because it was willing to do what C++ wouldn't: Bite the bullet, adding not-*technically*-necessary features and changes to create a better language. But then several years ago, in the name of "maturing", the D language took a 180 and started aiming itself directly at C++'s exact coordinates: "avoid changes at all costs". That robs D of exactly what makes it D. And slowly, but surely, every year, D inches closer and closer, bit by bit, to being another C++ all over again. Minus the market/mind share. Sh(i|oo)t, these days even *C++* is more willing to add non-breaking language-level features than D!!! "The last thing D needs" warning came too late.

This pattern is a constant frustration.

marler8997:
Yes I think you're right.  D has ALOT of friction when it comes to any kind of change regardless of the benefits.  That is, unless you are yourself part of the leadership.  I think this is a purposeful decision to decrease turmoil and attempt to make the language stable.  I personally think that stifling change and innovation hurts the language but that I don't think the leadership feels the same way.  Not much to do about it.  You can either try to change the leadership's minds (a very difficult task in general) or I suppose you could fork the code.  I found out it was much easier to fork `rdmd` than it was to constantly battle the friction to get anything done. Things to consider...

JackStouffer:
This has been said many times before, but I want to make this absolutely clear: 

Writing a DIP guarantees you a response from the leadership.

Just write one. It will take you maybe two hours, less time than you've spent writing and discussing this implementation.

>So what? 

Adding in language changes by bypassing existing channels willy-nilly because they have popular support seems like a poor long term strategy.

Additionally, when the main driver behind this seems to be efficiency over the library solutions it would make more sense to speed up the library solution, rather than add new features. Especially since this would mean speeding up templates and CTFE, which would benefit a lot more areas of D.

> the D language took a 180 and started aiming itself directly at C++'s exact coordinates: "avoid changes at all costs"

The implication that we aren't willing to make breaking changes anymore is ludicrous. We add deprecations to language and library features all the time.

marler8997:
> Writing a DIP guarantees you a response from the leadership.
> Just write one. It will take you maybe two hours, less time than you've spent writing and discussing this implementation.

We already got a response from leadership. As someone who has done alot of discussion on github and also gone through the DIP processs, I can say that you can get alot more done in a few days on github then you can get done in over a year with a DIP.  I spent over a year on a DIP and got 2 responses from leadership...and am still awaiting more discussion...but my last message was sent about 8 months ago so I'm not holding my breath.

Requiring people to write a formal proposal just so they can get a response is not reasonable.  If you have an idea, you should be able to casually discuss it before investing alot of time into a proposal.  And the DIP process takes ALOT of time...it's not 2 hours...it's hours split over the span of months to a year before the leadership even responds to it.

> Adding in language changes by bypassing existing channels willy-nilly because they have popular support seems like a poor long term strategy.

It's been the only strategy that's worked for me...and I've tried everything I can think of.  I'm speaking from "experience" here, not what "seems" like the right strategy.

wilzbach:
Well it all depends on someone pushing this through the DIP process. I'm more than happy to help and work on a DIP, but for now I'm a bit too occupied with other things. Anyone?

andralex:
It would be great to focus effort on finishing and improving the features we have, so working on a language proposal risks of diluting focus. For interpolation, the scant use of, and experience with, `mixin`-based workarounds is a negative, not a positive, so that doesn't help the case of a built-in feature. A better direction here would be to improve the syntax of `mixin` - either paren-less or as a suffix `.mixin`. That has better chances to argue successfully.

marler8997:
Cleaning up my old PRs that aren't going anywhere...this one goes to the chopping block

aberba:
String interpolation is mostly done with curly braces.

Flatllin:
@marler8997 Can we please reopen this? This would make D so much better!

@wilzbach I would love to help create a DIP if you could help guide me

marler8997:
Your welcome to the code but I can't work on D while my PRs are ignored for months and even years.

timotheecour:
@marler8997 see slack msg

Flatllin:
What is the reason to _not_ merge this PR? @marler8997 has already done all the work, and since it really wont affect anyone who doesn't want to use interpolated strings, why not just add it? On a side note, I don't thing interpolation should be applied to WYSIWYG strings, as it kind of defeats the purpose of WYSIWYG.

Flatllin:
@marler8997 why do you keep closing your PRs? I think you should leave them open so someone will review them.

marler8997:
I closed all my pending pull requests to the various dlang repositories back in May, I realized that all the work I was doing was starting to get unhealthy.  I decided to make a clean break and have been much happier since I stopped having to deal with all the apathy and disengagement and constant frustration when contributing to these repositories. But I still want D to succeed, if this code helps anyone, they are free to copy it and improve upon it as they see fit.

Geod24:
> What is the reason to not merge this PR? @marler8997 has already done all the work, and since it really wont affect anyone who doesn't want to use interpolated strings, why not just add it? On a side note, I don't thing interpolation should be applied to WYSIWYG strings, as it kind of defeats the purpose of WYSIWYG.

That's a bit OT for this PR, but I see this point coming up frequently and I'd like to address it.

When it comes to the core of the language, the reasoning should not be "why should we not include it?", but "why should we include it?", as the burden of proof falls on the people trying to get the feature in, not the other way around. Any new feature have to be maintained, tested, and cared for once it's in. As we cannot *just* remove an existing feature, we cannot add just anything.

@marler8997 : I'm sorry to hear that, yet not surprised. I believe your frustration was experienced by most, if not all regular contributors, myself included. I still hope to see some contributions from you in the future, as we do need passionate people.

jash11:
@marler8997 could you please re-open this pull request? Not that you have to continue to work on it yourself, but leaving it open would make it more accessible. This feature is taking far too long to get added to D, and I would put at least $150 bountysource on it if someone could go through the DIP process for it.

marler8997:
If someone wants to they can fetch the branch and create another pull request.
```
git fetch https://github.com/marler8997/dmd interpolatedStrings:interpolatedStrings
git push https://github.com/<MYUSER>/dmd interpolatedStrings
```

test/runnable/istring.d
adamdruppe: and prolly the suffixes. `i"foo $(bar)"w` should translate to `("foo "w, bar)` or something; just repeat the suffixes an the generated string literals inside.

jash11:
As this is obviously going nowhere without a DIP, here it is: https://github.com/jash11/DIPs/blob/master/Drafts/1NNN-JSH.md

jash11:
test

marler8997:
Because of renewed interest I've decided to rebase this PR and reopen it so that those working on the DIP can try out the implementation with the latest version of code.

jsatellite:
I've been playing with this and really like it. However, instead of returning a TupleExp, which means users have to import std.format/conv to get a string, what if `parseInterpolatedString` returned called a function in druntime/phobos that creates a struct wrapping the tuple and overriding `toString` (which is aliased to this)? This would allow those who wanted to access the tuple could do so, but the most common use of getting a string would be the default.

So parseInterpolatedString would end with something like this (alternatively instantiate the Interpolation instance but CallExp was easier to implement):
```d
auto te = new AST.TupleExp(token.loc, parts);
AST.Expression e = new AST.IdentifierExp(te.loc, Id.empty);
e = new AST.DotIdExp(te.loc, e, Id.object);
e = new AST.DotIdExp(te.loc, e, Identifier.idPool("__interpolation"));
return new AST.CallExp(te.loc, e, te);
```
And then you'd have this in object.d:
```d
struct Interpolation(T...) {

  T args;
  private this()(auto ref T args) { this.args = args; }
  string toString() { 
    import std.conv : text;
    return text(args);
  }
  alias toString this;

}

auto __interpolation(T...)(auto ref T args) { return Interpolation!T(args); }
```

marler8997:
An interesting idea.  One thought is that if we have a `string toString()` method we'd definitely also want a `void toString(scope void delegate(const(char)[]) sink)` overload.  If we look at the `writeln` example:

```D
writeln(i"a is $(a)");

// instead of becoming
writeln("a is ", a);

// now becomes
writeln(Interpolate!(string, typeof(a))("a is ", a));
```

We don't want it converting that struct to a string before calling `writeln`.  I'm not sure if I'd want to make `toString` an `alias this` or not.  I'd worry about passing the interpolated string to a function and having it unknowingly convert it to a string and calling the wrong overload.  This point is up for debate.

I'd have to think on this more to cover all the consequences but it seems promising.  It would also open up ways to do things like nice asserts.  If we added a special field to the struct like:
```D
struct Interpolated(T...)
{
    enum isInterpolatedString = true;
    //...
}
enum isInterpolatedString(T) = hasMember!(T, "isInterpolatedString");
```
Then you could make functions specifically for interpolated strings like this:
```D
void foo(T)(T istr) if (isInterpolatedString!T)
{
    ...
}
```
Once you can do that, you can expose more information about the interpolated string.  Like you could expose the original string parts before interpolation.   You could make a nice assert like this:
```D
void assert(T)(T istr) if (isInterpolatedString!T)
{
    assert(istr.args[0], "assert(" ~ istr.partsBeforeInterpolation[0] ~ ") failed");
}
assert(i"$(a == b)");
// if fails, prints 'assert(a == b) failed'
```

jash11:
@andralex Can we PLEASE get this merged? :pray:

jash11:
Also see: https://www.strawpoll.me/17590116

marler8997:
@jash11 I've rebased and updated the implementation to work with the latest branch.

jash11:
@WalterBright

thewilsonator:
@jash11 I hope to get progress on this at dconf.

pbbob:
@thewilsonator any progress?

jacob-carlborg:
> any progress?

From DConf? Not that I heard of.

wilzbach:
According to Andrei @marler8997 was working on a DIP for this? Is this true @marler8997?

Anyhow, I fully agree that we should just merge it under the new `-preview` flag as otherwise there's no way to get proper feedback on this. After all, that's what the `-preview` flags were created for.

src/dmd/cli.d
```diff
@@ -684,6 +684,8 @@ dmd -cov -unittest myprog.d
             "list all variables going into thread local storage"),
         Feature("vmarkdown", "vmarkdown",
             "list instances of Markdown replacements in Ddoc"),
+        Feature("interpolate", "interpolateStrings",
+            "enable interpolated string support via the 'i' prefix on string literals"),
```
wilzbach: Should be preview - not a transition - as this allows earlier merging.

test/runnable/istring.d
```diff
@@ -0,0 +1,72 @@
+/*
+PERMUTE_ARGS:
+REQUIRED_ARGS: -transition=interpolate
+*/
+import std.conv : text;
```
wilzbach: We should try to avoid using Phobos in the DMD testsuite. Maybe we can add a very reduced and minimalistic version of `text`?

WalterBright:
This does need a DIP. Issues that need covering:

1. need a grammar
2. what if the user wants to print a `$`
3. specifying things like field width, padding, hex output, etc.
4. mixed cases, like `i"... $(a) ... %s ... $(c)", b`
5. comparing `i" ... $(a) ..."` generating `" ... ", a, " ..."` versus `" ... %s ...", a`
6. supporting `$a` as well as `$(a)` (as suggested above)

WalterBright:
> we should just merge it under the new -preview flag as otherwise there's no way to get proper feedback on this.

I listed a number of design issues what would not be helped along with the current state as a preview. They're not difficult to resolve, they just need attention.

thewilsonator:
> This does need a DIP. 

https://github.com/dlang/DIPs/pull/140 the author seems to be missing but the DIP is there.

andralex:
> > This does need a DIP.
> 
> [dlang/DIPs#140](https://github.com/dlang/DIPs/pull/140) the author seems to be missing but the DIP is there.

A DIP would need a point of contact (usually but not necessarily the first author) stated in the header. If none is specified then it's the first author.

thewilsonator:
> A DIP would need a point of contact (usually but not necessarily the first author) stated in the header. If none is specified then it's the first author.

Totally, I'm just linking for those who might not know it is there.

WalterBright:
@marler8997 thanks for the detailed response. I think this makes it clear exactly why a DIP is needed - the design needs to be complete and all in one document. You have a great start on the implementation, people can play with it as they like (I do suggest at least rebasing it so it's easier for people to try out), but I think it is more productive at the moment to work on the DIP.

jacob-carlborg:
> Are you asking how this changes the language grammar? If so, it doesn't change it. It only changes the tokenization (i.e., allows the i character before string literals)

That needs to be updated here: https://dlang.org/spec/lex.html.

marler8997:
I will write a DIP so long as @WalterBright  and/or @andralex are willing to provide guidance and feedback at reasonable times or when requested.  I'm am very glad that I have received such feedback on my current DIP (DIP1011) which gives me some confidence that this will be the case.  However, I do worry because this goes against the guidelines of the DIP process.

I have a pretty good idea of how I'd like interpolated strings to be implemented but there are pieces that I will need to know your preferences on (i.e. which grammar node to use for non-parenthetical expressions).  So long as we are on the same page, I'll find some time soon to write the DIP and rebase/refresh this PR for experimentation.

wilzbach:
> I will write a DIP so long as @WalterBright and/or @andralex are willing to provide guidance and feedback at reasonable times or when requested.

@marler8997 One key announcement from DConf AGM which I think Mike Parker will publicly announce on the D Blog soon, is that Andrei passed down his top-level leadership functions to Atila.

You can watch this here for the full annoucement: https://www.youtube.com/watch?v=cpTAtiboIDs&t=3015s 

tl;dr: there will be a few big changes to the DIP process announced very soon. The biggest one is that you can ask Atila for guidance too. I don't know more than that, but I think more timely reviews and shorter DIP turnaround times are under discussion ;-)

pbbob:
@marler8997 any progress on the DIP?

marler8997:
No sorry. I'm still working on my current DIP, DIP 1011. Trying to finish my new changes to it before I take on another.

denizzzka:
There is a chance to replace .toString or .to!string by some user-defined method?

marler8997:
Cleaning my old PRs that aren't going to be integrated to clean up my Pull Request list.

PR #7989, opened 2018-03-09
fix Issue 18575 - making delegate from member function can lead to un…

WalterBright:
…safe code


JinShil (approved):
Looks ok to me, but I would rather see less code duplication.

WalterBright:
I considered refactoring, but that would make the changes I made harder to sort out from the original code. I figure refactoring should be done separately.

PR #7990, opened 2018-03-09
fix Issue 18576 - Compiler not doing RVO with auto returns

WalterBright:


src/dmd/s2ir.d
```diff
@@ -754,6 +766,16 @@ private extern (C++) class S2irVisitor : Visitor
...
+                        Type t = ce.e1.type.toBasetype();
+                        if (t.ty == Tdelegate)
+                            t = t.nextOf();
+                        if (t.ty == Tfunction && retStyle(cast(TypeFunction)t) == RET.stack)
+                        {
+                            irs.ehidden = el_var(irs.shidden);
+                            e = toElemDtor(s.exp, irs);
+                            e = el_una(OPaddr, TYnptr, e);
+                            goto L1;
+                        }
```
JinShil: These code blocks look identical.  Can they be combined into a reusable entity somehow?

test/runnable/test28.d
```diff
@@ -1247,6 +1247,49 @@ void test65()
...
+{
+    return S66(100);
+}
+
+void test18576()
+{
+    auto s = f66();
+    printf("%p vs %p\n", &s, callback.ptr);
+    callback();
+    printf("s.x = %d\n", s.x);
```
nemanja-boric-sociomantic: `printf` leftovers in tests.

WalterBright: They're not a problem. Lots of tests in the test suite are chatty.

src/dmd/s2ir.d
WalterBright: I thought I'd leave refactoring for another PR, especially if there turns out to be a problem with this code.

JinShil (approved):

test/runnable/test28.d
```diff
@@ -1247,6 +1247,49 @@ void test65()
...
+    assert(s.x == 101);
+    assert(&s == callback.ptr);
```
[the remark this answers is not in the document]
WalterBright: The point is, with RVO, it is not out of scope.

PR #7991, opened 2018-03-10
Fix Issue 18584 - Undefined identifier when not specifying 'this'

JinShil:


test/fail_compilation/test17380spec.d
```diff
@@ -1,9 +1,11 @@
 /* REQUIRED_ARGS: -verrors=spec
 TEST_OUTPUT:
 ---
-(spec:1) fail_compilation/test17380spec.d(14): Error: cannot resolve identifier `ThisTypeDoesNotExistAndCrashesTheCompiler`
-(spec:1) fail_compilation/test17380spec.d(14): Error: no property `ThisTypeDoesNotExistAndCrashesTheCompiler` for type `Uint128`
-fail_compilation/test17380spec.d(14): Error: undefined identifier `ThisTypeDoesNotExistAndCrashesTheCompiler`
+(spec:1) fail_compilation/test17380spec.d(16): Error: cannot resolve identifier `ThisTypeDoesNotExistAndCrashesTheCompiler`
+(spec:1) fail_compilation/test17380spec.d(16): Error: no property `ThisTypeDoesNotExistAndCrashesTheCompiler` for type `Uint128`
+(spec:1) fail_compilation/test17380spec.d(16): Error: cannot resolve identifier `ThisTypeDoesNotExistAndCrashesTheCompiler`
+(spec:1) fail_compilation/test17380spec.d(16): Error: no property `ThisTypeDoesNotExistAndCrashesTheCompiler` for type `Int128`
```
JinShil: I don't understand what these speculative errors are all about, and I'm not sure why the changes made affect this test.  Reviewer beware.

src/dmd/expressionsem.d
```diff
@@ -1328,6 +1328,17 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                     return;
                 }
             }
+
+            // https://issues.dlang.org/show_bug.cgi?id=16657
```
WalterBright: should this be bug 18584?

JinShil: Yes, sorry!.  Fixed.

test/fail_compilation/test17380spec.d
ibuclaw: Speculative are gagged errors. Makes sense to me. That extra error is related to your change (it now checks explicit this.doesnotexist as well as trying to resolve doesnotexist).

wilzbach (approved):
Looks like you managed to find the related bit :+1:

RazvanN7:
I don't think that this is the right fix. From what I could notice, the problem is in Scope.search since it doesn't follow the scopes correctly. Try this example:
```d
struct S { 
    int n;
    auto fun() { return tmp!(a => k)(); }
}

struct tmp(alias fns) {
    int k;
    alias fun = fns!int;
}
```

test.d(5): Error: need this for k of type int is not correct. It implies that you may use k.

Logging the scope search when compiling the above code looks like this:

IdentifierExp::semantic('k')
Scope.search(0x7f3777208820, 'k' flags=x0)
	scope __anonymous
	scope __anonymous
	scope tmp
	scope tmp!((a) => k)
	scope __anonymous
	scope test
	scope __anonymous

It looks like it erroneously searches tmp scope although it should not and it never gets to search the S scope.

JinShil:
@RazvanN7 I think you're right, but I'm having one hell of a time trying to figure this one out.

RazvanN7:
@JinShil I made an attempt to correclt fix this issue. Please take a look whenever you have time and sorry for the late response.

JinShil:
Closing in favor of #8159.  Thanks @RazvanN7!

PR #7992, opened 2018-03-10
Make Lexer.escapeSequence more reusable and add unittests

marler8997:
Moved the logic of the member function `Lexer.escapeSequence` into a static function with the same name.  Allows code to reuse the logic by calling `Lexer.escapeSequence(errorHandler, mystring)` instead of having to create a full lexer to re-use this logic.

I also added unittests for this function that both demonstrate how the function could be used without creating a full lexer.  It tests valid escape sequences as well as invalid ones making sure the correct error message is generated.

### DIFF
--- a/src/dmd/lexer.d
+++ b/src/dmd/lexer.d
@@ -194,9 +194,32 @@ unittest
     }
 }
 
+/**
+Handles error messages
+*/
+interface ErrorHandler
+{
+    /**
+    Report an error message.
+    Params:
+        format = format string for error
+        ... = format string arguments
+    */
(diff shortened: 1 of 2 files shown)


WalterBright (approved):

jacob-carlborg (approved):

src/dmd/lexer.d
```diff
@@ -194,9 +194,15 @@ unittest
     }
 }
 
+interface ErrorHandler
```
JinShil: DDoc for interface and methods.

JinShil (changes requested):

PR #7993, opened 2018-03-10
fix Issue 17892 - Scope analysis with -dip1000 fails for templated st…

WalterBright:
…ructs

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -1072,7 +1072,13 @@ extern (C++) int expandAliasThisTuples(Expressions* exps, size_t starti = 0)
 }
 
 /****************************************
- * Get TemplateDeclaration enclosing FuncDeclaration.
+ * If `s` is a function template, i.e. the only member of a template
+ * and that member is a function,
+ * return that template.
+ * Params:
+ *      s = symbol that might be a function template
+ * Returns:
+ *      template for that function, otherwise null
  */
 extern (C++) TemplateDeclaration getFuncTemplateDecl(Dsymbol s)
 {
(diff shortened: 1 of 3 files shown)


src/dmd/expression.d
```diff
@@ -1531,6 +1537,47 @@ StringExp semanticString(Scope *sc, Expression exp, const char* s)
...
+ * Given an Expression, find the variable it really is.
+ * for example, `a[index]` is really `a`, and `s.f` is really `s`.
+ * Params:
+ *      e = Expression to look at
+ * Returns:
+ *      variable if there is one, null if not
+ */
+VarDeclaration expToVariable(Expression e)
+{
+    while (1)
```
UplinkCoder: `while(true)`

src/dmd/expression.d
```diff
@@ -1072,7 +1072,13 @@ extern (C++) int expandAliasThisTuples(Expressions* exps, size_t starti = 0)
 }
 
 /****************************************
- * Get TemplateDeclaration enclosing FuncDeclaration.
+ * If s is a function template, i.e. the only member of a template
```
jacob-carlborg: Will `s` be highlighted here?

src/dmd/expression.d
```diff
@@ -1531,6 +1537,47 @@ StringExp semanticString(Scope *sc, Expression exp, const char* s)
     return se;
 }
 
+/***************************************************
+ * Given an Expression, find the variable it really is.
+ * for example, `a[index]` is really `a`, and `s.f` is really `s`.
```
jacob-carlborg: Capital `f` after period. I also recommend having adding a newline before this line. That would allow the first line to be used as a summary.

test/fail_compilation/test17892.d
```diff
@@ -0,0 +1,28 @@
...
+---
+fail_compilation/test17892.d(25): Error: returning `s.pointer1()` escapes a reference to local variable `s`
+fail_compilation/test17892.d(27): Error: returning `s.pointer2()` escapes a reference to local variable `s`
+---
+*/
+
+
+// https://issues.dlang.org/show_bug.cgi?id=17892
+
+struct S
```
mathias-lang-sociomantic: Why isne the struct templated?

nordlow: Why isn't this struct templatized as mentioned in the bugzilla issue?

WalterBright: because the problem is with inference, and `auto` will do that just as well and is simpler.

src/dmd/expression.d
JinShil: Yeah, I tried that once, too:  https://github.com/dlang/dmd/pull/7861#discussion_r167387243

JinShil (approved):

wilzbach:
> Merge branch 'master' into fix17892

FWIW (and maybe the future) we usually do rebases of a PR. The GitHub UI to resolve conflicts uses such a merge.

PR #7994, opened 2018-03-10
Fix issue 18034 - Bad codegen for OPvecfill

LemonBoy:
Do not replace vector values if that implies a change of Ety, doing so
may end up initializing the vector with the wrong content.

Another idea would be to special-case `OPvecfill` in the `Ety` assignment below to take advantage of the value replacement.

### DIFF
--- a/test/runnable/b18034.d
+++ b/test/runnable/b18034.d
@@ -1,0 +1,19 @@
+version (D_SIMD)
+{
+import core.simd;
+
+void check(void16 a) {
+    foreach (x; (cast(ushort8)a).array) {
+	assert(x == 1);
+    }
+}
+
+void make(ushort x) {
+    ushort8 v = ushort8(x);
+    check(v);
+}
+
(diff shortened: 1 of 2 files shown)


JinShil:
Welcome back!

test/runnable/b18034.d
```diff
@@ -0,0 +1,23 @@
+version (D_SIMD)
```
wilzbach: Should this have `// REQUIRED_ARGS: -O`  or maybe both variants with `ARG_SETS`?

LemonBoy: Good catch, I forgot about the `-O`.

JinShil:
Assigning to @WalterBright

JinShil (approved):

ibuclaw:
This PR introduced a regression https://issues.dlang.org/show_bug.cgi?id=23084

PR #7995, opened 2018-03-10
dmd part of fix for 18559

thewilsonator:
Add a bunch of overloads for intrinsics for in `core.math`.

### DIFF
--- a/src/dmd/toir.d
+++ b/src/dmd/toir.d
@@ -492,7 +492,7 @@ int intrinsic_op(FuncDeclaration fd)
         "5bitop7_popcntFNaNbNiNfmxx", // don't find 64 bit version in 32 bit code
         "5bitop7_popcntFNaNbNiNftZt",
     ];
-    __gshared immutable char*[46] core_namearray64 =
+    __gshared immutable char*[62] core_namearray64 =
     [
         //cos
         "4math3cosFNaNbNiNfdZd",
--- a/src/dmd/toir.d
+++ b/src/dmd/toir.d
@@ -413,19 +413,44 @@ int intrinsic_op(FuncDeclaration fd)
         OPyl2xp1,
     ];
 
-    __gshared immutable char*[46] core_namearray =
+    __gshared immutable char*[62] core_namearray =
     [
+        //cos
+        "4math3cosFNaNbNiNfdZd",
         "4math3cosFNaNbNiNfeZe",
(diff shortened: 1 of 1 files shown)


thewilsonator:
Druntime & phobos parts to follow.

quickfur:
Dumb question: is there a way to check that this emits the expected machine code? Or we simply don't have the facilities for that in place?

wilzbach:
> Dumb question: is there a way to check that this emits the expected machine code? Or we simply don't have the facilities for that in place?

We do, it's just not very well-known and at the moment the it contains quite a lot of duplication:

https://github.com/dlang/dmd/blob/master/test/runnable/test_cdvecfill.d
https://github.com/dlang/dmd/blob/master/test/runnable/test_cdcmp.d
https://github.com/dlang/dmd/blob/master/test/runnable/test_cdstrpar.d

At some point this should probably be factored out into _one_ small library and _one_ update script ...

JackStouffer:
What's the status of this? What still needs to be done in order to be considered?

thewilsonator:
I'd prefer #8002 to go instead of this but it fails for 32-bit, idk why, I won't have time to work on this until dconf. In the meantime this should suffice for the druntime and phobos parts assuming I have the mangling correct, but should probably be tested in conjunction with the associated druntime and phobos PRs to make sure.

JinShil:
This PR:

> I'd prefer #8002 to go instead of this

#8002:

> in the interest of moving forward with 18559 please review #7995

I'm in an infinite loop.

thewilsonator:
Haha. 
This one works, is not the solution I'd prefer, but isn't necessarily bad (more of the same code) . 
The other one is a better solution (removes use of strlen & friends and goes by symbol name rather then mangled) but doesn't pass 32-bit. 

Go with this one, unless you want to figure out why the other one fails for 32-bit. I plan to get the other one working at dconf.

WalterBright:
Can we get this one pulled? https://github.com/dlang/dmd/pull/8071

WalterBright (approved):

ibuclaw:
Possibly a regression introduced by this PR.

https://issues.dlang.org/show_bug.cgi?id=24040

There are float and double overloads for these core.math intrinsics, but they all still map to x87 instructions, and get treated as if the precision is 80-bit.

PR #7996, opened 2018-03-10
Fix Issue 18578 - First enum value assigned 0 instead of EnumBaseType.init

LemonBoy:
> The value of an EnumMember is given by its AssignExpression. If there is no AssignExpression and it is the first EnumMember, its value is the .init property of the EnumMember's type.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -2311,7 +2311,22 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                 if (!em.ed.isAnonymous())
                     em.ed.memtype = t;
             }
-            Expression e = new IntegerExp(em.loc, 0, t);
+
+            Expression e;
+
+            if (!t.isZeroInit(em.loc))
+            {
+                deprecation(em.loc,
+                            "Enum `%s` has base type `%s` with non-zero .init; " ~
+                            "initialize the enum member explicitly",
+                            em.ed.toPrettyChars(), t.toPrettyChars());
+                e = new IntegerExp(em.loc, 0, t);
+            }
+            else
+            {
+                e = t.defaultInitLiteral(em.loc);
+            }
+
             e = e.ctfeInterpret();
 
             // save origValue for better json output
--- a/test/fail_compilation/diag7477.d
+++ b/test/fail_compilation/diag7477.d
@@ -1,8 +1,10 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/diag7477.d(13): Error: integral constant must be scalar type, not `Foo`
-fail_compilation/diag7477.d(20): Error: integral constant must be scalar type, not `string`
+fail_compilation/diag7477.d(15): Deprecation: Enum `diag7477.Bar` has base type `Foo` with non-zero .init; initialize the enum member explicitly
+fail_compilation/diag7477.d(15): Error: integral constant must be scalar type, not `Foo`
+fail_compilation/diag7477.d(20): Error: no property `max` for type `string`
+fail_compilation/diag7477.d(23): Error: incompatible types for `(null) + (1)`: `Baz` and `int`
 ---
 */
 


thewilsonator:
This changes the value of `a` in and subsequently causes `b` to give an error. This is the cause of the Semaphore failure.
```
enum X : char
{
    a,
    b
}
```
ditto for `wchar` and `dchar`.

LemonBoy:
I see, the use of `float` as base type also gives different results since `float.init` is `NaN`.
I wonder how many dub packages are broken by this change, the testing pipeline shows a single [case](https://github.com/etcimon/libasync/blob/ffbc5ee5e001c984f20dc4eed02b8288774eec7d/source/libasync/posix.d#L3456) so far.

wilzbach:
> I wonder how many dub packages are broken by this change, the testing pipeline shows a single case so far.

According to Jenkins (the project tester) this triggers a segfault in libasync.

wilzbach:
Also fails in Phobos with:

```
std/stdio.d(3818): Error: enum member `std.stdio.__unittest_L3808_C9.EC.B` initialization with `EC.A+1` causes overflow for type `char`
```

marler8997:
https://issues.dlang.org/show_bug.cgi?id=18578

LemonBoy:
Hmm, this change may end up breaking some stuff...let's do a deprecation cycle first (or add some warnings/errors) ?

wilzbach:
> let's do a deprecation cycle first (or add some warnings/errors)

Warnings are often set to errors in some downstream projects, so deprecations are probably better here.

WalterBright:
It clearly will need a deprecation cycle. If `defaultInitLiteral` does not return `0`, then issue the deprecation.

WalterBright:
Please include test case from https://github.com/dlang/dmd/pull/8090/files

LemonBoy:
I've rebased this on top of #8090, what's missing is to adjust the failures in phobos.

marler8997:
This PR should fix the phobos issues: https://github.com/dlang/phobos/pull/6368

n8sh:
Maybe it would be better to change the language specification than to change this, because this utterly breaks for floating point types. [See further on in the specification](https://dlang.org/spec/enum.html):

>If there is no AssignExpression and it is not the first EnumMember, it is given the value of the previous EnumMember+1. If the value of the previous EnumMember is EnumBaseType.max, it is an error. If the value of the previous EnumMember+1 is the same as the value of the previous EnumMember, it is an error. (This can happen with floating point types.)

So you'll get an enum whose every member is NaN or just a compile-time error, depending on which kind of equality is meant by "the same as".

It also breaks for any `char` enum with more than one member, since `char.max` is `char.init`. Same for `wchar`.

marler8997:
Yeah maybe there is a better way to define how enum values are determined.  However, we are talking about a case that doesn't make much sense in the first place, i.e.
```D
enum Foo : float
{
    a, b
}
```
What would you expect the value of `a` and `b` to be?  This is such an odd usage of enums that having them default to `NaN` may actually be fine since this doesn't really make sense in the first place.

n8sh:
>This is such an odd usage of enums that having them default to NaN may actually be fine since this doesn't really make sense in the first place.

That is worse than failing to compile IMO.

marler8997:
> That is worse than failing to compile IMO.

Maybe...having all the members of an enum default to the same values does sound counter-intuitive to what an enum does, namely, enumerate a set of values.  But I can't think of a use case where you would ever enumerate a set of values where float is the base type, unless you were explicitly assigning values to them, can you?

marler8997:
Oh wait...I think I misunderstood you a bit.  I guess you were saying it would be better to fail in this case rather than assigning every value to `NaN`.  Yes I agree that failing in that case would be better.

JackStouffer:
Please don't forget to update dlang.org's deprecation page as well. Please see https://github.com/dlang/DIPs/pull/108 for an in draft process for deprecations of this nature.

RazvanN7:
ping @LemonBoy . what's the status on this?

thewilsonator:
@RazvanN7 status is rebased and green.

RazvanN7 (approved):

src/dmd/dsymbolsem.d
```diff
@@ -2311,7 +2311,22 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
             }
-            Expression e = new IntegerExp(em.loc, 0, t);
+
+            Expression e;
+
+            if (!t.isZeroInit(em.loc))
+            {
+                deprecation(em.loc,
+                            "Enum `%s` has base type `%s` with non-zero .init; " ~
+                            "initialize the enum member explicitly",
```
Geod24: The message should indicate why. Namely that the default value will change in a later release.
Also, there's no changelog entry.

Geod24:
Removed auto-merge to give time to address feedback. I think it's important to tell the user why they need to take a certain action when something is deprecated. Also, deprecations need a changelog entry.

n8sh:
I urge you all again to reconsider this. "The first element of an enum with base type T defaults to T.init" is very reasonable. "For every T, T.init should be an invalid value if possible" is reasonable. But combining the two leads to something very unreasonable.

>Removed auto-merge to give time to address feedback. I think it's important to tell the user why they need to take a certain action when something is deprecated. Also, deprecations need a changelog entry.

This is going to break existing code. People will right ask why a char enum's first element being 0xff is useful or helps anyone write better programs. This new behavior is a bear trap.

Geod24:
>  People will right ask why a char enum's first element being 0xff is useful or helps anyone write better programs.

At the same time, I don't think enums based on `char` are that common. Why wouldn't you use `[u]byte` instead ? The language does support it, but it's by no means a first class citizen.

WalterBright:
I suspect we should just fix the spec to match the behavior. This won't break existing code, and seems to be what people expect anyway.

marler8997:
Write a DIP

PR #7998, opened 2018-03-10
add expToVariable() function

WalterBright:
Factor out common code into new function.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -1072,7 +1072,13 @@ extern (C++) int expandAliasThisTuples(Expressions* exps, size_t starti = 0)
 }
 
 /****************************************
- * Get TemplateDeclaration enclosing FuncDeclaration.
+ * If `s` is a function template, i.e. the only member of a template
+ * and that member is a function,
+ * return that template.
+ * Params:
+ *      s = symbol that might be a function template
+ * Returns:
+ *      template for that function, otherwise null
  */
 extern (C++) TemplateDeclaration getFuncTemplateDecl(Dsymbol s)
 {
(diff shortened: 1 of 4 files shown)


src/dmd/expression.d
```diff
@@ -1072,7 +1072,13 @@ extern (C++) int expandAliasThisTuples(Expressions* exps, size_t starti = 0)
 }
 
 /****************************************
- * Get TemplateDeclaration enclosing FuncDeclaration.
+ * If `s` is a function template, i.e. the only member of a template
+ * and that member is a function,
+ * return that template.
```
jacob-carlborg: Why is there a newline after the comma? Looks like `return that template` would fit on the previous line.

src/dmd/expressionsem.d
```diff
@@ -5571,6 +5542,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 /* Slicing a static array is like taking the address of it.
                  * Perform checks as if e[] was &e
                  */
+version (none)
```
JinShil: Why not just delete this code?

WalterBright: Because I'm testing it.

WalterBright:
@JinShil It's ready now.

JinShil (approved):

wilzbach:
Out of interest: any reason why this isn't a member function or just `toVariable`?
`exp.expToVariable` is a bit redundant and reminds me of the old C-style function naming like `atoi`.

WalterBright:
Too many member functions create a 'kitchen sink' class which tends to be a mess.

PR #7999, opened 2018-03-11
fix Issue 17423 - pointer assignment to  in member function is not ac…

WalterBright:
…counted for

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -4964,11 +4964,12 @@ extern (C++) final class TypeFunction : TypeNext
      * the value of p can 'escape' the scope of the function.
      * This is useful to minimize the needed annotations for the parameters.
      * Params:
+     *  tthis = type of `this` parameter, null if none
      *  p = parameter to this function
      * Returns:
      *  true if escapes via assignment to global or through a parameter
      */
-    bool parameterEscapes(Parameter p)
+    bool parameterEscapes(Type tthis, Parameter p)
     {
         /* Scope parameters do not escape.
          * Allow 'lazy' to imply 'scope' -
@@ -4976,11 +4977,10 @@ extern (C++) final class TypeFunction : TypeNext
          * as lazy parameters to the next function, but that isn't
          * escaping.
          */
-        if (parameterStorageClass(p) & (STC.scope_ | STC.lazy_))
+        if (parameterStorageClass(tthis, p) & (STC.scope_ | STC.lazy_))
             return false;
         return true;
     }
-
 
     /************************************
      * Take the specified storage class for p,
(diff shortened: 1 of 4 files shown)


src/dmd/mtype.d
```diff
@@ -5033,6 +5035,24 @@ extern (C++) final class TypeFunction : TypeNext
...
+                if (tb.ty == Tclass)
+                    ad = (cast(TypeClass)tb).sym;
+                else if (tb.ty == Tstruct)
+                    ad = (cast(TypeStruct)tb).sym;
+                else
+                    assert(0);
+                foreach (VarDeclaration v; ad.fields)
+                {
+                    if (v.hasPointers())
+                        return stc;
```
wilzbach: This looks suspicious.

test/fail_compilation/test17423.d
```diff
@@ -0,0 +1,29 @@
...
+struct Foo
+{
+    Bar o;
+    int i = 3;
+
+    this(int x) @safe {
+        foreach(_; o) { i = 0; }
+        i = x;
+    }
+}
```
wilzbach: The test file from the bug report still triggers in a segfault (with this PR).

src/dmd/mtype.d
WalterBright: don't know what you mean

wilzbach: image

test/fail_compilation/test17423.d
WalterBright: It fails to compile with `-dip1000`

src/dmd/mtype.d
WalterBright: What about it? Please don't make me guess.

wilzbach: Oh my point was only that these two lines are uncovered, thus `return stc` is never called and thus all code that you added has never an effect for the testcase.

test/fail_compilation/test17423.d
wilzbach: Oh my point was that 17423 isn't fixed by this PR because the segfault still exists and with `-dip1000` it currently doesn't segfault: https://run.dlang.io/is/W1DGlg

src/dmd/mtype.d
WalterBright: never mind, I see

test/fail_compilation/test17423.d
WalterBright: My point is -dip1000 is required to fix it. The whole point of dip1000 is to fix these sorts of issues.

src/dmd/mtype.d
WalterBright: Should be covered now.

WalterBright:
Why are the coverage test results not showing up in the code view anymore?

wilzbach:
> Why are the coverage test results not showing up in the code view anymore?

They are for me:

image

Though I needed to hit refresh after I went to the diff view, maybe that's a new bug from CodeCov ...

WalterBright:
Yes, it worked for me too after I hit refresh. Thanks for the tip! Anyhow, it's all green now and good to go.

wilzbach:
FYI: I opened an issue (https://github.com/codecov/browser-extension/issues/53), but the extension seems to be unmaintained (https://github.com/codecov/browser-extension/issues/52).

WalterBright:
@JinShil @wilzbach please approve.

JinShil (approved):

JinShil:
Done.  Please consider returning the favor by reviewing some of the pull request that require someone with your unique skills and position.

PR #8000, opened 2018-03-11
supplemental fix for PR 7993

WalterBright:
Added the slice case to https://github.com/dlang/dmd/pull/7993 for https://issues.dlang.org/show_bug.cgi?id=17892

### DIFF
--- a/test/fail_compilation/test17892.d
+++ b/test/fail_compilation/test17892.d
@@ -1,8 +1,9 @@
 /* REQUIRED_ARGS: -dip1000
 TEST_OUTPUT:
 ---
-fail_compilation/test17892.d(25): Error: returning `s.pointer1()` escapes a reference to local variable `s`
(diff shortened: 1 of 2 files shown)


test/fail_compilation/test17892.d
```diff
@@ -16,13 +17,16 @@ struct S
...
 @safe int* testPointer(bool b)
 {
     S s;
     if (b)
         return s.pointer1();
-    else
+    else if (!b)
         return s.pointer2();
+    else
+        return s.pointer3();
```
ibuclaw: Use a switch if you'll be adding more cases? This branch should really error that its unreachable.

WalterBright: Doesn't matter if it is unreachable or not. The purpose of the test is to get it to fail to compile with the correct message. Any change that changes that will not pass the autotester.

wilzbach (approved):

PR #8002, opened 2018-03-11
refactor intrinsic_op

thewilsonator:
Also allows to add overloads of intrinsics without updating intrinsic_op, this supersedes #7995 as a more elegant solution to the dmd part of the fix for 18559.

it should also be faster on account of not computing the mangle, using `Identifier` comparisons and early exits.


ibuclaw:
Well, not testing the mangle deco is no good, as not all overrides are intrinsics.

~I'm not sure if this is faster, now you have linear search, rather than~ ... for some reason I thought you were using binary search for matching intrinsics in your front-end.

src/dmd/toir.d
```diff
@@ -361,311 +361,93 @@ elem *setEthis(const ref Loc loc, IRState *irs, elem *ey, AggregateDeclaration a
...
+            if (id3 == Id.inp)  op = OPinp;
+            if (id3 == Id.inpl) op = OPinp;
+            if (id3 == Id.inpw) op = OPinp;
+
+            if (id3 == Id.outp)  op = OPoutp;
+            if (id3 == Id.outpl) op = OPoutp;
+            if (id3 == Id.outpw) op = OPoutp;
+
+            if (id3 == Id.bswap)   op = OPbswap;
+            if (id3 == Id._popcnt) op = OPpopcnt;
```
ibuclaw: It would be better to check the identifier length first before the pointer comparison.
i.e: (pseudo-code)
```
switch (id3.length)
{
  case 3:
    if (id3 == Id.bsf)
      op = OPbsf;
    else if (...)
      // ...
    break;
  case ...:
    // ...
}
```

thewilsonator: true. Will do.

thewilsonator:
Its faster because there is no need to touch strings, just pointer compares with identifiers.

> Well, not testing the mangle deco is no good, as not all overrides are intrinsics.

What do you mean by that? I've been mostly looking at the `core.math` ones. The one (well three) special cases are `bsf`, `bsr` and `va_start` which are handled.

src/dmd/toir.d
thewilsonator: Hmm, LDC/GDC should be able to vectorise it pretty well without the switch as a masked gather , DMD obviously won't do that. Switching on the length requires loading it, its probably cold: I don't think it would buy all that much over the if chains because they are pretty short.

ibuclaw: Well, there were once only half a dozen or so intrinsics in the past, that list has grown considerably since then, and it will only continue to grow.

src/dmd/toir.d
```diff
@@ -361,311 +361,93 @@ elem *setEthis(const ref Loc loc, IRState *irs, elem *ey, AggregateDeclaration a
...
-
-    static assert(std_namearray.length == std_namearray64.length);
-    static assert(std_namearray.length == std_ioptab.length);
-    static assert(core_namearray.length == core_namearray64.length);
-    static assert(core_namearray.length == core_ioptab.length);
-    debug
+    id3 = fd.ident;
+    auto md = fd.getModule().md;
+    if (!md)
+        return op;
```
ibuclaw: Regarding segfault, I'm going to make a guess here.

Try:
```
auto m = fd.getModule();
if (!m || !m.md)
  return op;
```

src/dmd/toir.d
```diff
@@ -361,311 +361,95 @@ elem *setEthis(const ref Loc loc, IRState *irs, elem *ey, AggregateDeclaration a
...
+    auto m = fd.getModule();
+    if (!m || !m.md)
+        return op;
+
+    auto md = m.md;
+
+    id2 = md.id;
+
+    if (md.packages.dim != 1)
+        return op;
```
ibuclaw: `md.packages` could also be `null` here, e.g: for `module object`.

src/dmd/toir.d
```diff
@@ -361,311 +361,102 @@ elem *setEthis(const ref Loc loc, IRState *irs, elem *ey, AggregateDeclaration a
...
+
+    Lva_start:
+    if (global.params.is64bit &&
+        fd.toParent().isTemplateInstance() &&
+        fd.ident == Id.va_start)
+    {
+        OutBuffer buf;
+        mangleToBuffer(m, &buf);
+        const s = buf.peekString();
+        if (!strcmp(s, "4core4stdc6stdarg"))
```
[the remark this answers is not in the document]
thewilsonator: This is just a relocation of existing code. I want to get this working first :)

src/dmd/toir.d
```diff
@@ -361,311 +361,105 @@ elem *setEthis(const ref Loc loc, IRState *irs, elem *ey, AggregateDeclaration a
...
-        "5bitop3bsrFNaNbNiNfkZi",
-        "5bitop3bsrFNaNbNiNfmZi",
-        "5bitop3btcFNaNbNiPkkZi",
-        "5bitop3btrFNaNbNiPkkZi",
-        "5bitop3btsFNaNbNiPkkZi",
-        "5bitop3inpFNbNikZh",
-        "5bitop4inplFNbNikZk",
-        "5bitop4inpwFNbNikZt",
-        "5bitop4outpFNbNikhZh",
-        "5bitop5bswapFNaNbNiNfkZk",
```
thewilsonator: Why is only the 64-bit bswap an intrinsic (for both 64 and 32 bit)?

thewilsonator:
Hmm, semaphore seems to fail on 32-bit on [this line](https://github.com/dlang/phobos/blob/master/std/bitmanip.d#L4334) ( with `T == long` ? ).

thewilsonator:
I have no idea why this is not working on 32-bit, so in the interest of moving forward with 18559 please review #7995

PR #8003, opened 2018-03-11
Update runnable/test4.d for AArch64 and PPC64.

redstar:
Both architectures have alignment similar to x86_64.


ibuclaw (approved):

wilzbach (approved):
Seems like an awful lot of duplication just to distinguish for 32-bit, so imho these versions could be grouped together...

redstar:
@wilzbach Yes, I thought about this, too. I try to check the values for MIPS and PPC and then regroup the check.

redstar:
See https://github.com/dlang/dmd/pull/8004 for an update.

PR #8007, opened 2018-03-11
Ternary codegen fixes/optimization

LemonBoy:
Fixes 18534 and adds some niceties.
Adding more optimizations to the dmd backend feels like putting lipstick on a pig but those were low-hanging fruits.

### DIFF
--- a/src/dmd/backend/cod2.c
+++ b/src/dmd/backend/cod2.c
@@ -2011,8 +2011,7 @@ void cdcond(CodeBuilder& cdb,elem *e,regm_t *pretregs)
 
         opcode = 0x81;
         switch (sz2)
-        {   case 1:     opcode--;
-                        v1 = (signed char) v1;
+        {   case 1:     v1 = (signed char) v1;
                         v2 = (signed char) v2;
                         break;
             case 2:     v1 = (short) v1;
--- a/src/dmd/backend/cod2.c
+++ b/src/dmd/backend/cod2.c
@@ -2054,6 +2054,20 @@ void cdcond(CodeBuilder& cdb,elem *e,regm_t *pretregs)
             cgstate.stackclean--;
             return;
 	}
+        else if (v1 == v2)
+        {
+            unsigned reg = findreg(retregs);
+
+            movregconst(cdb, reg, v1, 0);
+
+            freenode(e21);
+            freenode(e22);
+            freenode(e2);
+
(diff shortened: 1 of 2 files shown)


wilzbach:
Have a look at Jenkins (seems like the first CI that completed), there are quite a few projects that break with illegal instructions.

WalterBright:
> lipstick on a pig

Haha, nice to have you back, @LemonBoy :-)

Anyhow, since this is multiple commits, I suggest trying them one at a time to see which one(s) are causing the test failures.

src/dmd/backend/cod2.c
```diff
@@ -2031,6 +2030,44 @@ void cdcond(CodeBuilder& cdb,elem *e,regm_t *pretregs)
         {
             // only sign-extension from 32-bits is available for 'and'
         }
+        else if (v1 == v2)
```
WalterBright: this optimization is better done in cgelem.c

src/dmd/backend/cod2.c
```diff
@@ -2011,8 +2011,7 @@ void cdcond(CodeBuilder& cdb,elem *e,regm_t *pretregs)
 
         opcode = 0x81;
         switch (sz2)
-        {   case 1:     opcode--;
```
WalterBright: This has been this way for 30 years. I know of no bugs it has caused.

WalterBright:
In fact, I'd like to have the independent changes in independent PRs. This means debugging (i.e. using git bisect) is much more effective at finding the culprit.

src/dmd/backend/cod2.c
LemonBoy: To be honest I expected the front-end to eliminate such a trivial case, maybe it's worth to move such an optimization there (iif it can be proved the comparison is side-effect free).

src/dmd/backend/cod2.c
LemonBoy: See 18534 and the first commit.

LemonBoy:
> cmov is usually a slow instruction only to be used when i-cache is tight.

cmov is useful when the and/add trick becomes too large and if you assume both the branches are equiprobable.
Since no branch weights are propagated to the backend I'd suggest to follow GCC/clang and use cmovs for "big" constants (whenever it's supported of course, I assumed all the x64 machines would support it but Jenkins proved me wrong).
I'll split the PRS asap, I just wanted to give this patch a run on some real world code.

PR #8008, opened 2018-03-12
fix Issue 17959 - [DIP1000] Can store scope delegate in non-scope member

WalterBright:


src/dmd/escape.d
```diff
@@ -216,6 +216,18 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
         va = null;
     }
 
+    if (va && e1.op == TOK.dotVariable && va.type.toBasetype().ty == Tclass)
```
wilzbach: could be:

```d
if (va)
switch (e.op)
{
  case TOK.concatenateElemAssign:
...
  case TOK.dotVariable:
...
  default: assert(0);
}
```

wilzbach (approved):

mathias-lang-sociomantic (approved):
LGTM

PR #8009, opened 2018-03-12
Issue 18594 - Replace lvalue in error messages with 'cannot modify'

wilzbach:
lvalue is a compiler developer term and I have seen quite a few people struggling with this term.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -1829,9 +1829,9 @@ extern (C++) abstract class Expression : RootObject
             loc = e.loc;
 
         if (e.op == TOK.type)
-            error("%s `%s` is a type, not an lvalue", e.type.kind(), e.type.toChars());
+            error("`%s` is a `%s` definition and cannot be modified", e.type.toChars(), e.type.kind());
         else
-            error("`%s` is not an lvalue", e.toChars());
+            error("`%s` is not an lvalue and cannot be modified", e.toChars());
 
         return new ErrorExp();
     }
@@ -2614,7 +2614,7 @@ extern (C++) final class IntegerExp : Expression
             e = this;
         else if (!loc.isValid())
             loc = e.loc;
-        e.error("constant `%s` is not an lvalue", e.toChars());
+        e.error("cannot modify constant `%s`", e.toChars());
         return new ErrorExp();
     }
 
@@ -4290,22 +4290,22 @@ extern (C++) final class VarExp : SymbolExp
     {
         if (var.storage_class & STC.manifest)
         {
-            error("manifest constant `%s` is not lvalue", var.toChars());
+            error("manifest constant `%s` cannot be modified", var.toChars());
             return new ErrorExp();
         }
         if (var.storage_class & STC.lazy_)
         {
-            error("lazy variables cannot be lvalues");
+            error("lazy variable `%s` cannot be modified", var.toChars());
             return new ErrorExp();
         }
         if (var.ident == Id.ctfe)
         {
-            error("compiler-generated variable `__ctfe` is not an lvalue");
+            error("cannot modify compiler-generated variable `__ctfe`");
             return new ErrorExp();
         }
         if (var.ident == Id.dollar) // https://issues.dlang.org/show_bug.cgi?id=13574
         {
-            error("`$` is not an lvalue");
+            error("cannot modify operator `$`");
             return new ErrorExp();
         }
         return this;
--- a/test/fail_compilation/checkimports1b.d
+++ b/test/fail_compilation/checkimports1b.d
@@ -3,7 +3,7 @@
 TEST_OUTPUT:
 ---
 fail_compilation/checkimports1b.d(16): Deprecation: local import search method found struct `imports.diag12598a.lines` instead of variable `checkimports1b.C.lines`
-fail_compilation/checkimports1b.d(16): Error: struct `lines` is a type, not an lvalue
+fail_compilation/checkimports1b.d(16): Error: `lines` is a `struct` definition and cannot be modified
 ---
 */
 
--- a/test/fail_compilation/checkimports1c.d
+++ b/test/fail_compilation/checkimports1c.d
@@ -3,7 +3,7 @@
 TEST_OUTPUT:
 ---
 fail_compilation/checkimports1c.d(16): Deprecation: local import search method found struct `imports.diag12598a.lines` instead of variable `checkimports1c.C.lines`
-fail_compilation/checkimports1c.d(16): Error: struct `lines` is a type, not an lvalue
+fail_compilation/checkimports1c.d(16): Error: `lines` is a `struct` definition and cannot be modified
 ---
 */
 
--- a/test/fail_compilation/diag10862.d
+++ b/test/fail_compilation/diag10862.d
@@ -1,24 +1,36 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/diag10862.d(28): Error: assignment cannot be used as a condition, perhaps `==` was meant?
-fail_compilation/diag10862.d(29): Error: assignment cannot be used as a condition, perhaps `==` was meant?
-fail_compilation/diag10862.d(30): Error: assignment cannot be used as a condition, perhaps `==` was meant?
-fail_compilation/diag10862.d(31): Error: assignment cannot be used as a condition, perhaps `==` was meant?
-fail_compilation/diag10862.d(32): Error: assignment cannot be used as a condition, perhaps `==` was meant?
-fail_compilation/diag10862.d(34): Error: assignment cannot be used as a condition, perhaps `==` was meant?
-fail_compilation/diag10862.d(35): Error: assignment cannot be used as a condition, perhaps `==` was meant?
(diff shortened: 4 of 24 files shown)


jacob-carlborg (approved):

src/dmd/expression.d
```diff
@@ -4285,22 +4285,22 @@ extern (C++) final class VarExp : SymbolExp
...
         if (var.storage_class & STC.manifest)
         {
-            error("manifest constant `%s` is not lvalue", var.toChars());
+            error("manifest constant `%s` cannot be modified", var.toChars());
             return new ErrorExp();
         }
         if (var.storage_class & STC.lazy_)
         {
-            error("lazy variables cannot be lvalues");
+            error("lazy variables cannot be modified");
```
WalterBright: While you're modifying the line anyway, how about upgrading it to:
```
error("lazy variable `%s` cannot be modified", var.toChars());
```

WalterBright (changes requested):
see comment

RazvanN7:
LGTM except Walter's comment.

src/dmd/expression.d
wilzbach: Good idea! Thanks! I also added a test case for it as it was uncovered.

src/dmd/expression.d
```diff
@@ -1834,9 +1834,9 @@ extern (C++) abstract class Expression : RootObject
             loc = e.loc;
 
         if (e.op == TOK.type)
-            error("%s `%s` is a type, not an lvalue", e.type.kind(), e.type.toChars());
+            error("cannot modify `%s`. It is a %s", e.type.toChars(), e.type.kind());
```
wilzbach: Not sure about this, maybe "cannot modify `%s`, it is a %s`" is better?

WalterBright (approved):

test/fail_compilation/checkimports1b.d
```diff
@@ -3,7 +3,7 @@
 TEST_OUTPUT:
 ---
 fail_compilation/checkimports1b.d(16): Deprecation: local import search method found struct `imports.diag12598a.lines` instead of variable `checkimports1b.C.lines`
-fail_compilation/checkimports1b.d(16): Error: struct `lines` is a type, not an lvalue
+fail_compilation/checkimports1b.d(16): Error: cannot modify `lines`. It is a struct
```
mathias-lang-sociomantic: Maybe:
```
`lines` is a struct definition and cannot be modified 
```

Basically holding the information that it's a type / definition makes this message clearer IMHO.

src/dmd/expression.d
mathias-lang-sociomantic: Ah, see my comment https://github.com/dlang/dmd/pull/8009#discussion_r173763618

wilzbach: Good idea. Thanks!

wilzbach:
(this is ready)

JinShil:
I think just stating that something cannot be modified without explaining why leads to frustration for those that do not understand.  I suggest saying something like "variable `whatever` is not an lvalue, so it cannot be modified" (wordsmithed to suit the precise situation, of course).

wilzbach:
>  "variable whatever is not an lvalue, so it cannot be modified" (wordsmithed to suit the precise situation, of course).

Not sure about this, I wanted to avoid lvalue because people have repeatedly told me that they have a hard time with this word (it's a compiler term for them).
Maybe "XX cannot be modified. Consider declaring it as mutable variable"?

JinShil:
| Maybe "XX cannot be modified. Consider declaring it as mutable variable"?

Better, but not by much.  In some cases, declaring a mutable variable is not what they need to do.

| I wanted to avoid lvalue because people have repeatedly told me that they have a hard time with this word (it's a compiler term for them).

`lvalue` is not just a compiler term, it's a programming concept (though I admit the nomenclature is quite poor), so it's in the user's interest to learn what it is. If they know, they will immediately understand the problem and likely know what to do about it.  If they don't know, then they would be prompted to look it up.  Without the term *lvalue* in the error message, it would be difficult for the user to know what to search for, but with the term *lvalue*, they'll immediately know what to search for, and will likely find their answer quickly.  

Letting the user know that they are trying to modify something that can't be modified is great, but we should also explain *why* it can't be modified if possible.  If you can explain why the code doesn't compile without using the word *lvalue*, that'd be great, but I can't think of a way to say it so precisely.

jacob-carlborg:
> Not sure about this, I wanted to avoid lvalue because people have repeatedly told me that they have a hard time with this word (it's a compiler term for them).
Maybe "XX cannot be modified. Consider declaring it as mutable variable"?

To me that sounds like `const` or `immutable` is somehow involved. Which would be an even more confusing error message.

wilzbach:
FYI: I just found a four year old issue about this (https://issues.dlang.org/show_bug.cgi?id=12663)

> DMD 2.066alpha gives an error message that is both wrong in contents and refers to the wrong line:
> test2.d(1,31): Error: constant 1u is not an lvalue
>Expected is an error message like:
> test2.d(3,5): Error: cannot modify immutable expression data

Funnily, it's exactly what I proposed here :P (i.e. without lvalue)

src/dmd/expression.d
```diff
@@ -1834,9 +1834,9 @@ extern (C++) abstract class Expression : RootObject
             loc = e.loc;
 
         if (e.op == TOK.type)
-            error("%s `%s` is a type, not an lvalue", e.type.kind(), e.type.toChars());
+            error("`%s` is a `%s` definition and cannot be modified", e.type.toChars(), e.type.kind());
```
JinShil: This is ok because we know a definition cannot be modified.

src/dmd/expression.d
```diff
@@ -1834,9 +1834,9 @@ extern (C++) abstract class Expression : RootObject
...
         else
-            error("`%s` is not an lvalue", e.toChars());
+            error("cannot modify `%s`", e.toChars());
```
JinShil: This is not ok.  Why can't it be modified?

src/dmd/expression.d
```diff
@@ -2614,7 +2614,7 @@ extern (C++) final class IntegerExp : Expression
             e = this;
         else if (!loc.isValid())
             loc = e.loc;
-        e.error("constant `%s` is not an lvalue", e.toChars());
+        e.error("cannot modify constant `%s`", e.toChars());
```
JinShil: This is ok.  We know that constant's can't be modified.

src/dmd/expression.d
```diff
@@ -4295,22 +4295,22 @@ extern (C++) final class VarExp : SymbolExp
     {
         if (var.storage_class & STC.manifest)
         {
-            error("manifest constant `%s` is not lvalue", var.toChars());
+            error("manifest constant `%s` cannot be modified", var.toChars());
```
JinShil: This is ok. We know that manifest constants cannot be modified.

src/dmd/expression.d
```diff
@@ -4295,22 +4295,22 @@ extern (C++) final class VarExp : SymbolExp
...
             return new ErrorExp();
         }
         if (var.storage_class & STC.lazy_)
         {
-            error("lazy variables cannot be lvalues");
+            error("lazy variable `%s` cannot be modified", var.toChars());
```
JinShil: Never used `lazy` before, but if the variable being `lazy` is the reason it can't be modified.  Then this is ok too.

src/dmd/expression.d
```diff
@@ -4295,22 +4295,22 @@ extern (C++) final class VarExp : SymbolExp
...
             return new ErrorExp();
         }
         if (var.ident == Id.ctfe)
         {
-            error("compiler-generated variable `__ctfe` is not an lvalue");
+            error("cannot modify compiler-generated variable `__ctfe`");
```
JinShil: This is ok, because `__ctfe` is special, and we know we can't modify things at runtime that only exist in the compiler's memory at compile-time.

src/dmd/expression.d
```diff
@@ -4295,22 +4295,22 @@ extern (C++) final class VarExp : SymbolExp
...
             return new ErrorExp();
         }
         if (var.ident == Id.dollar) // https://issues.dlang.org/show_bug.cgi?id=13574
         {
-            error("`$` is not an lvalue");
+            error("cannot modify `$`");
```
JinShil: This is ok, too, but I would prefer it say something like "cannot modify operator \`$\`", implying that the reason it can't be modified is because it is an operator.

JinShil (changes requested):

src/dmd/expression.d
wilzbach: Ok reverted this.

JinShil: I think it would be good to say "cannot modify \`%s\` because it is not an lvalue", or something like that.

JinShil: Or to keep with the pattern of the other error messages "\`%s\` cannot is not an lvalue, so it cannot be modified"

wilzbach: Oh sorry wrong wording. I meant I changed to:

> "`%s` is not an lvalue and cannot be modified

JinShil (approved):

PR #8010, opened 2018-03-12
Add AUTO_UPDATE=1 to the testsuite which automatically updates mismatching test output

wilzbach:
I have missed this for so long that I finally decided to do something about it.
Example usage:

```
make run_fail_compilation_tests -j20 AUTO_UPDATE=1
```

It's not 100% perfect as it doesn't work for test files with multiple, separated test output comment, but it's entirely opt-in and already a huge help for me.

### DIFF
--- a/test/d_do_test.d
+++ b/test/d_do_test.d
@@ -45,7 +45,8 @@ void usage()
           ~ "      DSEP:          \\\\ or /\n"
           ~ "      SEP:           \\ or / (required)\n"
           ~ "      OBJ:          .obj or .o (required)\n"
-          ~ "      EXE:          .exe or <null> (required)\n");
+          ~ "      EXE:          .exe or <null> (required)\n"
+          ~ "      AUTO_UPDATE:   set to 1 to auto-update mismatching test output\n");
 }
 
 enum TestMode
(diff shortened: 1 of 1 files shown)


wilzbach:
CC @CyberShadow as one of our test suite experts.

BTW with [`runnable/test_cdvecfill.d`](https://github.com/dlang/dmd/blob/master/test/runnable/test_cdvecfill.d), [`runnable/test_cdcmp.d`](https://github.com/dlang/dmd/blob/master/test/runnable/test_cdcmp.d), and [`runnable/test_cdstrpar.d`](https://github.com/dlang/dmd/blob/master/test/runnable/test_cdstrpar.d) there are a few precedents to this already.

test/d_do_test.d
```diff
@@ -792,14 +809,23 @@ int tryMain(string[] args)
             if (testArgs.disabled)
                 return Result.return0;
 
+            if (envData.autoUpdate)
```
jacob-carlborg: Looks like this is missing curly braces.

jacob-carlborg (changes requested):

test/d_do_test.d
wilzbach: No that's a short-hand syntax. Unfortunately assignments in `if` are quite limited, so `if (envData.autoUpdate && auto ce = cast(CompareException) e)` doesn't work.

WalterBright: I understand why you're doing that, but such syntax is always going to raise an eyebrow because it looks like a bug. I'll approve it because it is in the test harness and I'll defer to you on style in it - in the compiler/phobos I'd also ask to change it.

PR #8011, opened 2018-03-12
fix Issue 18000 - [scope] auto-generated opAssign not scope aware

WalterBright:
Also added `notMaybeScope()` which makes debugging inference problems much easier.

### DIFF
--- a/src/dmd/escape.d
+++ b/src/dmd/escape.d
@@ -85,7 +85,7 @@ bool checkParamArgumentEscape(Scope* sc, FuncDeclaration fdc, Identifier par, Ex
 
         Dsymbol p = v.toParent2();
 
-        v.storage_class &= ~STC.maybescope;
+        notMaybeScope(v);
 
         if (v.isScope())
         {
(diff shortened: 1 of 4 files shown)


src/dmd/escape.d
```diff
@@ -1292,3 +1292,26 @@ void findAllOuterAccessedVariables(FuncDeclaration fd, VarDeclarations* vars)
...
+ *      v = variable
+ */
+version (none)
+{
+    void notMaybeScope(string file = __FILE__, int line = __LINE__)(VarDeclaration v)
+    {
+        printf("%.*s(%d): notMaybeScope('%s')\n", cast(int)file.length, file.ptr, line, v.toChars());
+        v.storage_class &= ~STC.maybescope;
+    }
+}
```
wilzbach: How about defining a special `opAssign` on STC which might be a more generic debugging facility?

wilzbach (approved):
Not too happy about `notMaybeScope` as logging all assignments should be possible with a dedicated `opAssign`.

src/dmd/escape.d
WalterBright: Because I don't want a blizzard of messages to sort through, and because I've had repeated issues only with maybescope, not others.

PR #8012, opened 2018-03-12
[stable] issue 18515: Fix it so that neither g++ nor CC is required.

wilzbach:
Cherry-picked https://github.com/dlang/dmd/pull/7949, s.t. win32_64 hopefully passes on auto-tester for `stable` again.

See also: https://github.com/braddr/d-tester/issues/71

----

It was previously fixed so that cc is used if CC isn't set when
compiling dmd, but the test suite still had g++ hardcoded, so if the
system was set up with clang and not gcc, the dmd test suite barfed
midway through. This fixes it so that it uses c++ if CC isn't set
instead of using g++.


wilzbach:
So apparently just cherry-picking that patch doesn't work.

PR #8013, opened 2018-03-12
fix Issue 12486 - Function returning struct isn't called if `enum` of its result is accessed

FeepingCreature:
Fix issue 12486

When accessing an enum member of a struct returned by an expression with side effects, don't discard the expression but evaluate it anyways. (No, this is not safe even if the expression is pure: it might loop infinitely.)

### DIFF
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -3223,6 +3223,10 @@ private extern(C++) final class DotExpVisitor : Visitor
                 }
                 checkAccess(e.loc, sc, null, v);
                 Expression ve = new VarExp(e.loc, v);
+                if (!isTrivialExp(e))
+                {
+                    ve = new CommaExp(e.loc, e, ve);
+                }
                 ve = ve.expressionSemantic(sc);
                 result = ve;
                 return;
--- a/test/runnable/test12486.d
+++ b/test/runnable/test12486.d
@@ -1,0 +1,18 @@
+module test12486;
+
+struct S { enum a = 1; } // or `const` but not for all types
+
+S f(ref int i)
+{
+    ++i;
+    return S();
+}
+
+void main()
+{
+    int i = 2;
+    assert(f(i).a == 1);
+    // ensure that f(i) was actually called, even though
+    // a is a statically known property of the returned type
+    assert(i == 3);
+}


FeepingCreature:
Ah, I see, it has to be an expression but it is a struct.
Wait, huh?

FeepingCreature:
Is there a way to say "the expression is trivial"? I don't like enumerating all the different kinds of trivial expressions manually. Maybe optimize just the comma expression, somehow?

FeepingCreature:
Weird, auto-tester seems to be running the wrong version of the code. Gonna wait for everything to finish, then kick it off again.

wilzbach:
> Weird, auto-tester seems to be running the wrong version of the code. Gonna wait for everything to finish, then kick it off again.

No need to wait. auto-tester automatically refreshes itself though it might take a bit until it reruns your PR as there's usually a lot of traffic.

FeepingCreature:
Okay. `e.optimize` is _definitely_ the wrong way to go about this. Is there any way to decide whether an Expression is relatively trivial?

FeepingCreature:
_:stares at ci Jenkins:_

Do you ever get that feeling that, while all the tests are green, they really *shouldn't* be?

wilzbach:
> Do you ever get that feeling that, while all the tests are green, they really shouldn't be?

Jenkins doesn't run the testsuite. It just compiles dmd and then builds ~40 projects and runs their testsuite.

FeepingCreature:
:stares at AppVeyor:

Do you ever get that feeling that, while all the tests are green, they really shouldn't be?

FeepingCreature:
Is it stuck? auto-tester has been at pending: 8 for hours now.

wilzbach:
> Is it stuck? auto-tester has been at pending: 8 for hours now.

No, auto-tester deletes all builds whenever a new PR is merged at DMD, druntime or phobos. Click on the link to see the history.
Don't worry too much about it. Once a PR is approved, it gets priority and lands in a special auto-merge queue.

FeepingCreature:
How do I find a reviewer?

src/dmd/typesem.d
```diff
@@ -3223,7 +3223,16 @@ private extern(C++) final class DotExpVisitor : Visitor
                 }
                 checkAccess(e.loc, sc, null, v);
                 Expression ve = new VarExp(e.loc, v);
-                ve = ve.expressionSemantic(sc);
+                if (e.op == TOK.type || e.op == TOK.this_ || e.op == TOK.super_ || e.op == TOK.variable)
+                {
+                    ve = ve.expressionSemantic(sc);
```
FeepingCreature: The point here is to avoid being forced to optimize the expression if we know the lhs of the comma expression would be trivial, since optimization may change the meaning but is sometimes necessary so we don't have a pointless comma expression lying around. I wish there was a better way to do this.

ibuclaw:
> How do I find a reviewer?

Ping or assign me and I'll get round to it. 😉

FeepingCreature:
Um ... @ibuclaw ping? I don't know how to assign people.

FeepingCreature:
I debugged what's happening with the Phobos tests.

The Cartesian products being computed in std.setops are ridiculously too large. They *look* small, but due to the fact that 3D nested Cartesian product is strongly biased towards Z spread, actually take billions to trillions of cycles to run.

This was not noticed before because due to the bug, they essentially passed by coincidence. (`canFind` calls `any`.)

src/dmd/typesem.d
RazvanN7: Please add this explanation as a comment in the code. It will be useful in the future when people will try to understand the code.

RazvanN7 (approved):

src/dmd/typesem.d
FeepingCreature: Documentation added.

src/dmd/typesem.d
```diff
@@ -3223,7 +3223,24 @@ private extern(C++) final class DotExpVisitor : Visitor
                 }
                 checkAccess(e.loc, sc, null, v);
                 Expression ve = new VarExp(e.loc, v);
-                ve = ve.expressionSemantic(sc);
+                // In case the expression is something known trivial, like this.enum, super.enum,
+                // variable.enum or type.enum, where we know the lhs does not have side effects,
+                // just use the enum directly.
+                // It would be preferable if we could just directly check if the left hand side of the
+                // expression is obviously side-effect free, but there does not seem to be a way to do that.
```
WalterBright: See https://github.com/dlang/dmd/blob/master/src/dmd/sideeffect.d

src/dmd/typesem.d
```diff
@@ -3223,7 +3223,24 @@ private extern(C++) final class DotExpVisitor : Visitor
...
+                if (e.op == TOK.type || e.op == TOK.this_ || e.op == TOK.super_ || e.op == TOK.variable)
+                {
+                    ve = ve.expressionSemantic(sc);
+                }
+                else
+                {
```
WalterBright: This fix looks like it's in the wrong place. `expressionSemantic()` should *not* be removing expressions with side effects. Putting the fix here is papering over the problem for this case, and it'll just cause other problems. Need to find the root problem.

WalterBright (changes requested):
Fix is in the wrong place.

src/dmd/typesem.d
FeepingCreature: Right, the problem is if I keep the side effects via comma expression, as I have to in order to get correct code in the case call().ENUM, then I end up with unusable comma expressions downstream. I can't fix it earlier, because this is where the access resolves.

WalterBright: > what DMD does right now is removing side effects from expressions in that code: call().ENUM becomes ENUM, with call() being discarded.

That's a bug, as call() has side effects. That's where it needs to be fixed.

> The problem is fixing that without breaking enum access for trivial lhs, say alias foo = this.bar.

Maybe that's there the special case should go.

FeepingCreature: I don't understand what you mean. This is where that happens.

src/dmd/typesem.d
FeepingCreature: As I understand it, a function that never returns but is pure will still be considered as not having side effects by this code? But we must not optimize it away: we may optimize >1 calls to 1 call, but we mustn't optimize 1 call to 0 calls.

FeepingCreature:
Ping?

quickfur:
@FeepingCreature  I think what Walter meant is that the proper fix should probably be somewhere in `sideeffect.d` so that the code here, unchanged, will do the right thing.  In other words, the fact that the code here is discarding side effects when it shouldn't, means that `sideeffect.d` (or whatever else is the root cause of this problem) isn't doing its job properly.

quickfur:
More specifically, there may be a defect somewhere in `hasSideEffect` or `callSideEffect` that isn't detecting the side-effect when it should be.  Or perhaps `hasSideEffect` or `callSideEffect` isn't getting called when it should be, resulting in the compiler throwing away code thinking that there are no side-effects, when in fact there are.

FeepingCreature:
Well, are infinite loops *supposed* to count as side effects? If not, istm there is a lack of an idiom here.

ibuclaw:
> Well, are infinite loops supposed to count as side effects? If not, istm there is a lack of an idiom here. isDisposable or such.

All calls should be considered as a side effect until inlined. There is `isTrivialExp` in the sideeffects module that likely fits your needs.

FeepingCreature:
Thanks, that seems sensible.

I still don't get why Walter thinks the fix is in the wrong place. This is the place that throws away the side effect right now.

FeepingCreature:
@WalterBright I tried to keep it as a DotExp instead of forcing it into a CommaExp, but if I do so all the way through codegen, it produces invalid code because it seems to try and treat the enum access as a struct read. ISTM that CommaExp is just how enum dot accesses to expressions with side effects have to look right now. If it's possible to make a better fix for this, I don't know how. I don't think I can get the code materially better than this.

FeepingCreature:
ping

ibuclaw (approved):
Seems reasonable to me.

@FeepingCreature - please rebase to restart failing tests.
@WalterBright - check this again.

wilzbach:
> please rebase to restart failing tests.

No need. auto-tester will automatically restart itself and all other CIs are passing.
Regarding the failures on auto-tester - they are spurious and I just submitted:

https://github.com/dlang/phobos/pull/6347

which hopefully helps to weed them out.

FeepingCreature:
ping

ibuclaw:
I'll compile this patch in locally and inspect the change done, then will try to break it. 😉

If all is OK I see no problem with committing this.

src/dmd/typesem.d
```diff
@@ -3223,6 +3223,10 @@ private extern(C++) final class DotExpVisitor : Visitor
                 }
                 checkAccess(e.loc, sc, null, v);
                 Expression ve = new VarExp(e.loc, v);
+                if (!isTrivialExp(e))
```
ibuclaw: @FeepingCreature - you could also use `hasSideEffects(e)` (or `e.hasSideEffects()` if you prefer UFCS).  When having an initial glance at the implementation, it didn't seem like it was a strong enough guarantee, but having a closer look, only strongly pure functions would have their calls omitted here.  Which is harmless.

FeepingCreature: As far as I can tell, `pure void foo() { while (true) { } }` is strongly pure, but must under no circumstances be optimized away.

ibuclaw: Ah, bugger those looping pure functions. :-)

Yep, makes sense, and LGTM.

ibuclaw:
@WalterBright - Any last comments before I dismiss your stale review?

src/dmd/typesem.d
ibuclaw: @FeepingCreature - For brownie points, you can add a comment with a link to the bugzilla reference.

PR #8014, opened 2018-03-12
Fix issue 18534 - Correct codegen of some ternary sequences

LemonBoy:
Changing the base OPcode from 0x81 to 0x80 also changes how the MOD/RM
field is interpreted: if reg > 3 end up accessing the [8..16] bits of
{eax,ecx,edx,ebx} instead of the [0..8] bits of {esp,ebp,esi,edi} as the
code expects.

### DIFF
--- a/src/dmd/backend/cod2.c
+++ b/src/dmd/backend/cod2.c
@@ -2011,8 +2011,7 @@ void cdcond(CodeBuilder& cdb,elem *e,regm_t *pretregs)
 
         opcode = 0x81;
         switch (sz2)
-        {   case 1:     opcode--;
-                        v1 = (signed char) v1;
+        {   case 1:     v1 = (signed char) v1;
                         v2 = (signed char) v2;
                         break;
             case 2:     v1 = (short) v1;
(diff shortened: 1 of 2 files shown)


stefan-koch-sociomantic:
:heart: 
I've been hunting for this bug suspecting it in the optimizer, I should have known better and check codegen :)
Thanks for fixing!

JinShil:
Another PR for which @WalterBright is uniquely qualified to review.  Assigning to him.

WalterBright:
I believe this fix is incorrect. The 0x80 opcodes are the byte versions of the instructions, which is what is desired here. The problem is when the register used is not a byte register. The solution for 16 and 32 bit compilations is to restrict the register used to be BYTEREGS, and for 64 bit code use the 8 bit addressing mode for registers via the REX prefix.

LemonBoy:
Sure thing, but I'm not a huge fan of partial register updates myself.
If you want to keep those then I think that relaxing the `code_orrex` check to something like `(I64 && reg >= 4)` suffices.

WalterBright:
It doesn't suffice for 32 bit code generation, which has the bug as well, just that it's hard to generate a test case for it. I'll just go ahead and fix it. Thanks for pinpointing where the problem is, though! That's very helpful.

WalterBright:
Rebooted as https://github.com/dlang/dmd/pull/8052

PR #8015, opened 2018-03-12
Fix issue 18461 - Do not remove assignments to ambiguous symbols

LemonBoy:
If a value is reached through a pointer don't consider it as definitely
dead.

cc @WalterBright

### DIFF
--- a/src/dmd/backend/gother.c
+++ b/src/dmd/backend/gother.c
@@ -1488,6 +1488,7 @@ STATIC void accumda(elem *n,vec_t DEAD, vec_t POSS)
                 vec_free(Pr); vec_free(Dr);
                 break;
 
+            case OPrelconst:
             case OPvar:
             {   symbol *v = n->EV.sp.Vsym;
                 targ_size_t voff = n->EV.sp.Voffset;
--- a/test/runnable/b18461.d
+++ b/test/runnable/b18461.d
@@ -1,0 +1,10 @@
+// REQUIRED_ARGS: -O -inline -release
+import core.bitop;
+
+void main()
+{
+    size_t test_val = 0b0001_0000;
+
+    if(bt(&test_val, 4) == 0)
+        assert(false);
+} 


src/dmd/backend/gother.c
```diff
@@ -1324,11 +1324,18 @@ void rmdeadass()
...
+                        if (!vec_testbit(j,DEAD))
+                        {
+                            /* If not positively dead but v is live on a    */
+                            /* successor to b, then v is live.              */
+                            //printf("\tDEAD=%d, live=%d\n",vec_testbit(j,DEAD),vec_testbit(v->Ssymnum,b->Boutlv));
+                            if (vec_testbit(v->Ssymnum,b->Boutlv))
+                                    continue;
+                            /* Do not remove assignments to ambiguous values as
+                             * they might be reached trough a pointer */
+                            if (!(v->Sflags & SFLunambig))
```
WalterBright: I suspect the bug is elsewhere, as the testbit on b->Boutlv should take into account ambiguous values

LemonBoy: The variable doesn't escape the BB, that's why it's missing from there.
I've pushed another fix that's simpler and hopefully more correct.

src/dmd/backend/gother.c
```diff
@@ -1488,6 +1488,7 @@ STATIC void accumda(elem *n,vec_t DEAD, vec_t POSS)
                 vec_free(Pr); vec_free(Dr);
                 break;
 
+            case OPrelconst:
```
WalterBright: This is close, but not quite right. The fix is to add a case for `OPbt`. I'll work on that tomorrow.

WalterBright:
Turns out there was a bit more to fixing it. I rebooted it as https://github.com/dlang/dmd/pull/8065

Thank you for identifying where things were going wrong in the optimizer. That's 90% of the work in fixing this problem. This made it easy for me.

PR #8019, opened 2018-03-13
fix Issue 18597 - more unsafe unaligned pointer errors

WalterBright:


src/dmd/dstruct.d
```diff
@@ -479,6 +479,19 @@ extern (C++) class StructDeclaration : AggregateDeclaration
             Type origType = t;
             Type tb = t.toBasetype();
 
+            const hasPointers = tb.hasPointers();
```
stefan-koch-sociomantic: seems to me like this block would be big enough to put it into a function.
This would help preventing diverging error messages.

WalterBright:
The autotester on Darwin_64_32 is seg faulting when compiling dmd. Is it using the same version of dmd as the other test platforms?

wilzbach:
Maybe related: https://github.com/dlang/dmd/pull/8022#comment-373617747

WalterBright:
ping @wilzbach

JinShil (approved):

PR #8020, opened 2018-03-13
Add Musl flag -libcmusl.

joakim-noah:
Dmd 2.079 and ldc 1.8 shipped with [@yshui's Musl port](https://github.com/dlang/druntime/pulls?q=is%3Apr+author%3Ayshui+is%3Aclosed), which is useful for running D apps in Alpine linux containers for microservices, though only ldc 1.8 has a flag to enable it, ldc-developers/ldc#2373.  Dmd 2.079 has to be manually modified to invoke `CRuntime_Musl`, hence this pull to finally enable it.

All ~but one~ of the stdlib unit tests pass with Musl, ~an assert related to `core.stdc.time.tzdata` not having a time zone set by default with Musl, dlang/phobos#6193~ (fixed in dlang/phobos#6548).  Also, a few of the additional tests in druntime trip some asserts related to shared libraries, dlang/druntime#2071, and a couple tests fail in the dmd testsuite, ldc-developers/ldc#2604.

This pull was tested both on a linux/x64/glibc distro with a Musl package installed, [Arch linux](https://www.archlinux.org/packages/community/x86_64/musl/), followed by running those same cross-compiled tests on Alpine, and dmd building natively on the latest Alpine linux/x64 3.7.0, by bootstrapping with the native Alpine build of ldc 1.8. I also used ldc's cross-compilation tool `ldc-build-runtime` and the Musl package to cross-compile the tests with the 1.8 release, no problem.  I was only unable to cross-compile a D compiler itself for Alpine, because Arch's Musl package doesn't support C++.

The runtime flag in this pull is merely a placeholder till we decide how we want to deal with cross-compiling from a non-Musl distro. I see three possibilities:

1. Take out the runtime flag and make them manually change it, ie the status quo before this pull.
2. A runtime flag specific to Musl, ie this pull.
3. A runtime flag that lets the user choose from several libc's, which might be particularly useful since @rracariu is finishing up his uClibc port.

I implemented the second option here, but don't care what's done.  I will amend this pull, including updating `dmd.cli`, once the dmd devs decide on an approach.

### DIFF
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -114,6 +114,7 @@ struct Param
     bool isOpenBSD;         // generate code for OpenBSD
     bool isDragonFlyBSD;    // generate code for DragonFlyBSD
     bool isSolaris;         // generate code for Solaris
+    bool isLibcMusl;        // Use the Musl C Runtime
     bool hasObjectiveC;     // target supports Objective-C
     bool mscoff = false;    // for Win32: write MsCoff object files instead of OMF
     // 0: don't allow use of deprecated features
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1383,7 +1383,15 @@ void addDefaultVersionIdentifiers()
     }
     else static if (TARGET.Linux)
     {
-        VersionCondition.addPredefinedGlobalIdent("CRuntime_Glibc");
+        version (CRuntime_Musl)
+            VersionCondition.addPredefinedGlobalIdent("CRuntime_Musl");
+        else
+        {
+            if (global.params.isLibcMusl)
+                VersionCondition.addPredefinedGlobalIdent("CRuntime_Musl");
+            else
+                VersionCondition.addPredefinedGlobalIdent("CRuntime_Glibc");
+        }
     }
 
     if (global.params.isLP64)
@@ -1675,6 +1683,17 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
                 else
                 {
                     error("-mscrtlib");
+                }
+            }
+            else if (arg == "-libcmusl")
+            {
+                static if (TARGET.Linux)
+                {
+                    params.isLibcMusl = true;
+                }
+                else
+                {
+                    error("-libcmusl is only supported on linux");
                 }
             }
             else if (startsWith(p + 1, "profile")) // https://dlang.org/dmd.html#switch-profile


jacob-carlborg:
I think the third option makes most sense.

wilzbach:
> A runtime flag that lets the user choose from several libc's, which might be particularly useful since @rracariu is finishing upd his uClibc port.

(runtime is a bit confusing in this context as the druntime accepts flags with `--DRT` too, maybe "compiler flag" is a better word)

I think that's the best way forward. Just `-libcmusl` doesn't scale well (and both @WalterBright and @andralex have repeatedly called for being careful when adding new CLI flags).

--------

(a bit of explanation that's just partially related)

IIRC @ZombineDev wanted to add a target-triple like GDC and LDC support them (see e.g. https://clang.llvm.org/docs/CrossCompilation.html) though that's just a side note.

LLVM for example supports `x86_64-pc-linux-musl` (see https://github.com/llvm-mirror/llvm/commit/4c93fe068533cf9568fc1c62fb0b24fd463dfdcf)

FYI: since a few weeks DMD also outputs the compilation targets in the JSON and verbose output.
Examples: https://run.dlang.io/is/dZ5sl2 and https://run.dlang.io/is/rc7Oz6 where a triple (or individual fields like `architecture`) would be better than just the current messy list of `predefinedVersions`.

rracariu:
Implementing something akin to LDC `-mtriple` would make more sense in the long run.

jacob-carlborg:
A flag to specify a triple would be even better.

joakim-noah:
>(runtime is a bit confusing in this context as the druntime accepts flags with --DRT too, maybe "compiler flag" is a better word)

I was trying to differentiate the runtime flag from the first version block, where Musl is set when the compiler itself is compiled, for a native Musl compiler.

As for adding a cross-compilation triple, which you all seem to prefer, I don't see the point of it for dmd.  You can usually set the Arch, OS, and ABI using a triple, but dmd only supports two CPU architectures and doesn't currently support cross-compiling to other OS's.  So that only leaves the ABI, which is essentially what this libc flag would be adding.

jacob-carlborg:
I would like to see DMD be able to cross-compile in the future.

joakim-noah:
I would too- I've even suggested an ARM backend for dmd on the forum a couple times- but isn't it premature to add a triple before dmd can do any of that? :smile: Also, ldc/llvm and presumably gdc use their own triples, I don't see them switching to some new dmd scheme.

I'd rather add a libc flag now to enable the limited "cross-compilability" that's already implemented, then worry about generalizing once we actually support cross-compiling to anything else.  We can always deprecate the libc flag _if and when_ that happens.

jacob-carlborg:
> We can always deprecate the libc flag if and when that happens.

Good luck with that 😉.

src/dmd/globals.d
```diff
@@ -114,6 +114,7 @@ struct Param
     bool isOpenBSD;         // generate code for OpenBSD
     bool isDragonFlyBSD;    // generate code for DragonFlyBSD
     bool isSolaris;         // generate code for Solaris
+    bool isLibcMusl;        // Use the Musl C Runtime
```
ibuclaw: We need a new place to put this information. Only the modules that take care of compilation steps care about this.

joakim-noah: I could have just defined a variable in the `dmd.mars` module, as it's only used there.  I simply added it to this global list, because it's where everything else is added.  Anyway, let's hold off on where to put the variable until we decide whether we want a runtime flag and what it's supposed to do first.

ibuclaw: Yes, hence just a comment from me. Generally speaking, all params that reference a platform or target should ideally not be here.

joakim-noah:
@wilzbach, worth setting up a Circle CI running Alpine once this is in? I'd need to look into the few remaining test failures, but might be worth keeping this Must port up to date.

wilzbach:
Yes that would be pretty cool!
Here's the way I currently auto-build the Docker images used for dmd's CircleCi:

https://github.com/wilzbach/dlang-docker/tree/master/circleci

and how it's configured:

https://github.com/dlang/dmd/blob/master/.circleci/config.yml

In theory it's as easy as building the Docker image once and then specifying the docker image at CircleCi. However, there's one problem with this: druntime and phobos could in theory still introduce regressions.
There are at least three options:
(1) just add it at dmd (disadvantage: regressions might pop in unnoticed and no way to test whether druntime/phobos PRs break musl)
(2) add the setup to all three repos (disadvantage: more maintenance)
(3) add it to the [Jenkins](https://github.com/dlang/ci) which runs on all platforms (disadvantage: probably a bit complicated to add docker support to it)

tl;dr: I guess we have to go with (2)

CC
@MartinNowak (who maintains dlang/ci)
@dkgroot (who maintains his own SemaphoreCI for DragonFlyBSD)
@CyberShadow (who wanted to build a more sophisticated CI solution for dlang a while ago)

joakim-noah:
I could merge this now, but I'd like someone with decision-making authority on dmd CLI flags to choose one of the three options listed in the top post. @andralex or @Jinshil, since you focus on cross-compilation?

wilzbach (changes requested):
I thought we go with a more generic `-libc` / `-mlibc` flag?
(Or even sth like `-mtriple`)

Also this should be added to the documentation which is in `cli.d`

JinShil:
> I'd like someone with decision-making authority on dmd CLI flags to choose one of the three options listed in the top post. @andralex or @JinShil, since you focus on cross-compilation?

Sorry, I don't have any decision making authority, but since you brought me into the conversation, I'll say that it's quite unfortunate that D has such strong coupling to the C standard library.  If it were up to me, I'd demand that all C and C++ standard library bindings be removed from druntime and delegated to Deimos or Dub, and then privately imported by the druntime ports that need them.  Applications would need to explicitly import and link to them just like any other 3rd party library.  I guess that means option (1) is the closest to my preference.

wilzbach:
>  Applications would need to explicitly import and link to them just like any other 3rd party library. 

Of DUB would be better, I would advocate to do put Phobos on DUB too ;-)

> I'll say that it's quite unfortunate that D has such strong coupling to the C standard library

Yeah, but I don't think this is going to change in the near future :/

> I guess that means option (1) is the closest to my preference.

FWIW I still vote for (3) - the more generic `-mlibc` flag that can be used for uclibc (and other runtimes) too.

Maybe we can involve @andralex and @WalterBright in this decision process?

andralex:
I know nothing about libcmusl. Who around here is a domain expert? Can we avoid dedicating a flag to it?

jacob-carlborg:
@andralex If we want to support swapping which version of libc we're linking against when running the compiler I think a new flag is required. If we instead require the users to compile druntime, Phobos and perhaps the compiler as well to be able to use a non-default libc it would be enough with specifying a `-version`.

I prefer that the user can choose this when running the compiler.

joakim-noah:
I believe the D stdlib will need to be recompiled or downloaded separately. This is strictly about being able to use the same compiler, ie not having to recompile DMD too when using a different libc.

jacob-carlborg:
I we have a dedicated flag we could ship pre-compiled libraries with different names that the compiler can recognize if this flag is passed.

ibuclaw:
@jacob-carlberg - what if one version path implements a function that is otherwise extern on other runtimes? Or low level types that are of different sizes between musl and glibc? Is shared implemented in the same way? Etc... 

If anything differs in our implementation then you need two different druntime libraries anyway.

jacob-carlborg:
If we go with the version approach one needs to recompile druntime and Phobos, and perhaps the compiler as well to use a different libc than the default one. I don't think it's a good idea, I'm just saying it might be an option.

wilzbach:
> Currently, we use flags like -m32mscoff

FWIW there's already `-mscrtlib` with which you can specify the C runtime to link with... (it's Windows only).

joakim-noah:
ping @WalterBright, any input? This is an important compiler flag for microservices, a market that D should target. The [ldc builds for Alpine containers have been downloaded close to a 100 times](http://www.somsubhra.com/github-release-stats/?username=ldc-developers&repository=ldc), with little advertising.

jacob-carlborg:
Can we please go with a flag to specify a target triple? At least I would like DMD to be able to cross-compile, Manu has shown interest in this as well.

wilzbach:
So to summarize the discussion.
AFAICT these are the two most popular options for this PR:

A) `-mlibc`
B) `-mtriple` (LLVM-like)

As multiple people have expressed interest in DMD being able to cross-compile, it's probably better to go with the more generic `-mtriple`.
If there's interest, a more specialized `-mlibc` flag can always be added later for better convenience.
If no one is objecting to this in the next three days, I take this as a silent agreement for `-mtriple`.

chadjoan:
I just ran into this thread while trying to update DMD on my musl system.

I second JinShil's comments on libc dependencies: don't have them. At least whenever possible.

If you push DMD into a lot of the future useful targets, then you will run into places where there is no libc or it doesn't make sense for libc to exist.

Imagine making DMD emit javascript, Python (ex: D code callable from Python that runs on the python interpreter itself, without requiring host system integration), .NET bytecode, Haxe environment / Neko bytecode, legacy systems (anyone else ever do migration work? >:) ), and so on.

In those cases, you'll want things like qsort and some math functions and whatnot to have pure-D paths. They do computation only (and not of great complexity, either) and no systems integration, so they should never be a hard-dependency. It makes sense to version them out when runtime optimization calls for delegating them to the host environment, but please make it easy to get things into the "It Works" state on an unexplored system as quickly as possible. Some libc things aren't purely computation, like backtrace support, and the most helpful thing would be to have those be as modular as possible, and to have the ability to turn them off if they aren't absolutely necessary for making "Hello World" run on a new system.

This is all very out-in-the-future, and some possible targets (ex: javascript) would require a lot more work than merely removing libc dependencies, but it'd be a step forward and would remove a particularly un-fun (at least IMO) obstacle for people working on such things. And then there are things like ulibc, which would also benefit greatly from decoupling D from libc in a more immediate fashion, just like musl would (have).

Hope that helps. Thanks for reading.

joakim-noah:
That would be nice, but as you can imagine, nobody wants to do that "un-fun" work of removing the libc dependency, so it hasn't happened.

joakim-noah:
Not going to work on this, so closing.

PR #8023, opened 2018-03-13
Emit a newline at the end of the JSON output file

wilzbach:
Many editors automatically insert a trailing newline,
so it's a bit annoying to manually edit the JSON output files.
Also I don't see any reason why we shouldn't emit a trailing CR at the end, so I think the best way forward is to just write `\n` at the end of the JSON generation.

CC @marler8997 @timotheecour


src/dmd/json.d
```diff
@@ -999,6 +999,8 @@ public:
 extern (C++) void json_generate(OutBuffer* buf, Modules* modules)
 {
     scope ToJsonVisitor json = new ToJsonVisitor(buf);
```
wilzbach: Why I didn't add this in in the destructor:
`scope` with classes is planned to be deprecated. While that's controversial (see e.g. https://github.com/dlang/dlang.org/pull/2027), imho it's best not to rely on a feature that might will be removed soon.

marler8997 (approved):

JinShil (approved):

PR #8025, opened 2018-03-13
sh_do_test: verbose trace + start to remove inferred output_file

wilzbach:
Now that a dedicated Bash wrapper exists, a lot of boilerplate code can be removed from the bash scripts.

This is a start which saves an verbose trace of the bash log in `.log` and all stdout and stddev in `output_file`. A trap will print both output when an error occurs.

There's a lot more that can be done, see https://github.com/dlang/dmd/pull/7928 for ideas.

CC @marler8997 @CyberShadow

### DIFF
--- a/test/compilable/crlf.sh
+++ b/test/compilable/crlf.sh
@@ -2,6 +2,8 @@
 
 # Test CRLF and mixed line ending handling in D lexer.
 
+set -e
+
 dir=${RESULTS_DIR}/compilable
 fn=${TEST_DIR}/${TEST_NAME}.d
 
@@ -45,8 +47,6 @@ printf 'static assert(wstr == "%s");\n' 'foo\nbar\nbaz\n' >> ${fn}
 printf 'enum dstr = q"(\r\nfoo\r\nbar\nbaz\r\n)";\n' >> ${fn}
 printf 'static assert(dstr == "%s");\n' '\nfoo\nbar\nbaz\n' >> ${fn}
 
-$DMD -c -D -Dd${TEST_DIR} -m${MODEL} -of${TEST_DIR}/${TEST_NAME}a${OBJ} ${fn} || exit 1
+$DMD -c -D -Dd${TEST_DIR} -m${MODEL} -of${TEST_DIR}/${TEST_NAME}a${OBJ} ${fn}
 
 rm -f ${TEST_DIR}/${TEST_NAME}a${OBJ} ${TEST_DIR}/${TEST_NAME}.html ${fn}
-
-echo Success
--- a/test/compilable/issue17167.sh
+++ b/test/compilable/issue17167.sh
@@ -16,8 +16,6 @@ src="$bin_base.d"
 echo 'void main() {}' > "${src}"
 
 # Only compile, not link, since optlink can't handle long file names
-$DMD -m"${MODEL}" "${DFLAGS}" -c -of"${bin}" "${src}" || exit 1
+$DMD -m"${MODEL}" "${DFLAGS}" -c -of"${bin}" "${src}"
 
 rm -rf "${dir:?}"/"$TEST_NAME"
-
-echo Success
--- a/test/compilable/jsonNoOutFile.sh
+++ b/test/compilable/jsonNoOutFile.sh
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 set -euo pipefail
 dir=${RESULTS_DIR}/compilable
-output_file=${dir}/jsonNoOutFile.sh.out
-$DMD -c -od=${dir} -Xi=compilerInfo compilable/extra-files/emptymain.d > ${output_file}
-rm -f emptymain.json || true
+$DMD -c -od=${dir} -Xi=compilerInfo compilable/extra-files/emptymain.d
+rm -f emptymain.json
--- a/test/compilable/test18367.sh
+++ b/test/compilable/test18367.sh
@@ -1,13 +1,10 @@
 #!/usr/bin/env bash
 
-set -u -o pipefail
+set -ueo pipefail
 
 # dmd should not segfault on -X with libraries, but no source files
-out=$("$DMD" -conf= -X foo.a 2>&1)
-[ $? -eq 1 ] || exit 1
-echo "$out" | grep -q 'Error: cannot determine JSON filename, specify using `-Xf=<file>` or provide at least one source file'
-[ $? -eq 1 ] || exit 1
+out=$(! "$DMD" -conf= -X foo.a 2>&1)
+echo "$out" | grep 'Error: cannot determine JSON filename, use `-Xf=<file>` or provide a source file'
 
-out=$("$DMD" -conf= -Xi=compilerInfo 2>&1)
-[ $? -eq 1 ] || exit 1
-echo "$out" | grep -q 'Error: cannot determine JSON filename, use `-Xf=<file>` or provide a source file'
+out=$(! "$DMD" -conf= -Xi=compilerInfo 2>&1)
+echo "$out" | grep 'Error: cannot determine JSON filename, use `-Xf=<file>` or provide a source file'
--- a/test/runnable/depsprot.sh
+++ b/test/runnable/depsprot.sh
@@ -3,23 +3,19 @@
 name=`basename $0 .sh`
 dir=${RESULTS_DIR}/runnable
 dmddir=${RESULTS_DIR}${SEP}runnable
-output_file=${dir}/${name}.sh.out
 deps_file="${dmddir}${SEP}${name}.deps"
 
 die()
 {
-    cat ${output_file}
     echo "---- deps file ----"
     cat ${deps_file}
     echo
     echo "$@"
-    rm -f ${output_file} ${deps_file}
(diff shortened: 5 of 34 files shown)


test/compilable/jsonNoOutFile.sh
```diff
@@ -1,6 +1,5 @@
 #!/usr/bin/env bash
 set -euo pipefail
 dir=${RESULTS_DIR}/compilable
-output_file=${dir}/jsonNoOutFile.sh.out
-$DMD -c -od=${dir} -Xi=compilerInfo compilable/extra-files/emptymain.d > ${output_file}
-rm -f emptymain.json || true
+$DMD -c -od=${dir} -Xi=compilerInfo compilable/extra-files/emptymain.d
+rm -f emptymain.json
```
wilzbach: Not sure why `|| true` was added here. `rm -f` always exits with `0` and also this tests doesn't check the generated output.

test/compilable/test18367.sh
```diff
@@ -1,13 +1,13 @@
 #!/usr/bin/env bash
 
-set -u -o pipefail
+set -u -o pipefail -e
 
 # dmd should not segfault on -X with libraries, but no source files
 out=$("$DMD" -conf= -X foo.a 2>&1)
-[ $? -eq 1 ] || exit 1
```
wilzbach: That's exactly what `-e` is for. `set -euo pipefail` should probably be set by default, but that's a task for another PR.

test/compilable/issue17167.sh
```diff
@@ -16,8 +18,6 @@ src="$bin_base.d"
 echo 'void main() {}' > "${src}"
 
 # Only compile, not link, since optlink can't handle long file names
-$DMD -m"${MODEL}" "${DFLAGS}" -c -of"${bin}" "${src}" || exit 1
+$DMD -m"${MODEL}" "${DFLAGS}" -c -of"${bin}" "${src}"
 
 rm -rf "${dir:?}"/"$TEST_NAME"
-
-echo Success
```
wilzbach: We no longer need to write sth. to the output file to create it as it's now created by default and only removed in the error case. Of course, that behavior could be easily changed in the `sh_do_test` script, but the point is that this is superfluous, was often forgotten in the past and should be handled by the wrapper.

test/runnable/test16096.sh
```diff
@@ -4,23 +4,13 @@ set -e
...
 if [ "$OS" != 'osx' ] || [ "$MODEL" != '64' ]; then
     exit 0
 fi
 
-function teardown {
-    if [ "$?" -ne 0 ] && [ -r "$output_file" ]; then
-        cat "${output_file}" 1>&2
-    fi
-    rm -f "${output_file}"
-}
```
wilzbach: No need for custom rolled error handling anymore. This is the sane default behavior.

test/runnable/test2.sh
```diff
@@ -2,33 +2,20 @@
...
 for x in "${a[@]}"; do
-    echo "executing with args: $x" >> ${output_file}
+    echo "executing with args: $x"
 
-    $DMD -m${MODEL} $x -unittest -od${dmddir} -of${dmddir}${SEP}test2${EXE} runnable/extra-files/test2.d >> ${output_file}
-    if [ $? -ne 0 ]; then
-        cat ${output_file}
-        rm -f ${output_file}
-        exit 1
-    fi
```
wilzbach: Kill the custom error handling!

test/compilable/jsonNoOutFile.sh
marler8997: Cause I didn't know `rm -f` always returned 0 :)

marler8997: I remember I was having troubles trying to check the generated output, I can't remember why though...

marler8997 (approved):
Definitely an improvement :)

wilzbach:
Pinging @CyberShadow as you seem to be the only one being able to approve this.

CyberShadow:
> Pinging @CyberShadow as you seem to be the only one being able to approve this.

What do you mean? Any committer can approve it.

wilzbach:
> What do you mean? Any committer can approve it.

Yes, but there aren't so many who feel comfortable with Bash - you and Martin are the only one who I know and Martin is traditionally very busy.
Sorry for the poor wording.

test/runnable/depsprot.sh
```diff
@@ -3,23 +3,19 @@
 name=`basename $0 .sh`
 dir=${RESULTS_DIR}/runnable
 dmddir=${RESULTS_DIR}${SEP}runnable
-output_file=${dir}/${name}.sh.out
```
CyberShadow: Looks like this script wanted to print its output only on failure? Is its output overly verbose, or something?

test/compilable/crlf.sh
```diff
@@ -2,6 +2,8 @@
 
 # Test CRLF and mixed line ending handling in D lexer.
 
+set -e
```
CyberShadow: 👍

I'd go for the whole `-euo pipefail` trio. Variables that might be unset should be "declared" at the top of the file, i.e. replace unset variables with empty ones.

test/runnable/gdb15729.sh
```diff
@@ -2,12 +2,13 @@
 
 src=runnable${SEP}extra-files
 dir=${RESULTS_DIR}${SEP}runnable
-output_file=${dir}/gdb15729.sh.out
+
+set -e
```
CyberShadow: Flags should probably come right after the shebang, since `-u` could be added later, and those variable definitions would benefit from `-u`'s checks.

test/runnable/test16096.sh
CyberShadow: Depending on how the makefile is written, `.DELETE_ON_ERROR` might be required.

test/sh_do_test.sh
```diff
@@ -30,14 +30,36 @@ output_file=${RESULTS_DIR}/${script_name}.out
...
+
+        rm -rf ${output_file}
+        exit $?
+    fi
+}
+trap 'finish $?' INT TERM EXIT
+
+# redirect stdout + stderr to the output file + keep a reference to the std{out,err} streams for later
+exec 40>&1
+exec 41>&2
```
CyberShadow: bash has a syntax to assign a free file descriptor and store it in a variable (`exec {name_of_fd_var}>&1`), which should guarantee that there will not be collisions, but the feature is "relatively" new and not sure if it's available on all test systems.

test/sh_do_test.sh
```diff
@@ -30,14 +30,36 @@ output_file=${RESULTS_DIR}/${script_name}.out
...
+function finish {
+    # reset output stream
+    set +x
+    exec 1>&40
+    exec 2>&41
+
+    if [ $1 -ne 0 ]; then
+        echo ==============================
+        echo Test ${script_name} failed. The logged output:
+        cat ${output_file}
```
CyberShadow: General note, would be nice to see a lot more quoting. This line (and many many others) will fail if the variable contains spaces or some other special characters. ShellCheck helps greatly in identifying such moments.

test/sh_do_test.sh
```diff
@@ -30,14 +30,36 @@ output_file=${RESULTS_DIR}/${script_name}.out
...
-    exit 1
-fi
+function finish {
+    # reset output stream
+    set +x
+    exec 1>&40
+    exec 2>&41
+
+    if [ $1 -ne 0 ]; then
+        echo ==============================
```
CyberShadow: General note, `printf` > `echo` in scripts.

CyberShadow (approved):
Thanks! Looks good overall.

test/sh_do_test.sh
marler8997: FYI, I think if tests contained spaces, there's alot of places that would fail.  But I suppose that's not a good reason to perpetuate bad practice.

test/runnable/depsprot.sh
wilzbach: All output is already printed on a failure, so this is just redundant.

test/runnable/gdb15729.sh
wilzbach: Done.
FYI: My plan is to add `set -eu` and maybe even `set -o pipefail` by default for ALL bash scripts as that would be a sane default.

test/runnable/test16096.sh
wilzbach: Oh this is now handled by the `sh_do_test` wrapper

test/compilable/crlf.sh
wilzbach: Yes, my plane is to enable the entire `-euo pipefail` trio by default for all bash scripts in `sh_do_test.sh`.
This is just an intermediate PR that should lay the ground work for this.
I just added `-e` here as it's an almost hard-coded reaction whenever I see `|| exit 1`

test/sh_do_test.sh
wilzbach: Nice! According to [this SO answer](https://stackoverflow.com/a/41620630/3026245) it's available since available since bash 4.1+ (2009-12-31), so I just gave it a try.
Who knows, maybe it works ...

PR #8026, opened 2018-03-13
sh_do_test: Export LIBEXT

wilzbach:
Define `LIBEXT` and `OBJ` once instead of redefining it over and over again.

CC @marler8997 @CyberShadow

### DIFF
--- a/test/compilable/test6461.sh
+++ b/test/compilable/test6461.sh
@@ -3,12 +3,6 @@
 dir=${RESULTS_DIR}/compilable
 src=compilable/extra-files/${TEST_NAME}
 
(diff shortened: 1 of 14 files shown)


marler8997 (approved):

wilzbach:
> Do we need shared library extensions as well?

It's currently not used by any bash script. Of course, if there's demand it could be easily added, but I believe in the "on demand" strategy for creating global variables.

There are a few more variables that I would like to add, but I will add them incrementally over other PRs to keep the scope minimal and make it easier to review.

jacob-carlborg (approved):

PR #8027, opened 2018-03-13
Fix issue 18504 - Generate unwind tables for halting functions

LemonBoy:
I think #7333 heuristic is too eager.


JinShil (approved):

JinShil:
LGTM, but I'm not really qualified.  @WalterBright or @ibuclaw, please review.

WalterBright:
When creating a PR to fix a bugzilla issue, please post a link to the PR in the issue. This is so that someone browsing bug reports will know if and where a PR for it is.

WalterBright (approved):

ibuclaw:
Why is this information in the frontend? This is a backend check, and dmd specific.

WalterBright:
All it does is set a flag, which is then used by the glue code and back end.

PR #8028, opened 2018-03-13
Optimize away trivial CondExp expressions

LemonBoy:


ibuclaw:
Wouldn't the backend do this anyway?

LemonBoy:
> Wouldn't the backend do this anyway?

It depends on which backend we're talking about, as shown in #8007 dmd doesn't do that and I think this optimization is better suited in the frontend since we have more informations about e1/e2/econd.

WalterBright:
As I recommended before, this should go in the backend in `cgelem.c`. Reasons:

1. cgelem.c is packed with such patterns.
2. compilers normally do this sort of thing in the back end - as I'm sure gdc and ldc do already
3. the optimizer reshuffles code a lot, and these usually produce new opportunities for optimization. Doing it in the front end will not cover new opportunities.
4. the intermediate representation is far simpler than D AST code is, hence far less opportunities for bugs
5. do we really want some patterns in the front end and some in the back end? I don't think so.
6. the back end has enough information to do this.

WalterBright:
BTW, I made a huge mistake with dmd by putting the inliner in the front end. It should go in the back end. Reasons 1..6 apply in spades.

LemonBoy:
Not so fast, don't forget that the intermediate representations is much simpler as is dumber.
Consider the following piece of code:
```d
bool f1(ulong i) pure nothrow
{
    return (i != 0);
}
int f2(ulong i)
{
    return f1(i) ? 'A' : 'A';
}
```
While `hasSideEffect` is able to infer `f1` is side-effect free thanks to the additional metadata available at the AST level you can also notice `el_sideeffect` is not able to do so (Unless you manage to have the compiler to inline it).

You may also notice how well `f1` is optimized :)

```objdump
48 89 7d f8          mov    QWORD PTR [rbp-0x8],rdi
48 89 7d f8          mov    QWORD PTR [rbp-0x8],rdi
48 0f 95 c0          rex.W setne al
```

WalterBright:
> el_sideeffect is not able to do so

It is. There's a `OPcallns` in the intermediate code which is a function call with no side effects.

LemonBoy:
I just checked and here's the result:
```
el:0x55b6ae3cd140 cnt=0 call  TYbool 0x55b6ae3cd010 0x55b6ae3cd1a0
 el:0x55b6ae3cd010 cnt=0 var  TYD func  _D5r_QKa2f1FNaNbmZb
 el:0x55b6ae3cd1a0 cnt=0 var  TYuns long long  i
el_sideeffect(e->E1) = 1
el_match(e->E2->E1, e->E2->E2) = 1
```

> the optimization of f1(i) ? 'A' : 'A' should be f1(i), 'A'

Yes, I agree.

LemonBoy:
After some digging the `OPcall` there is due to the `global.params.useAssert` check in `e2ir.d`, that's why I think this small optimization can be done in the frontend.

WalterBright:
Yes, it can be done in the front end. But there needs to be a compelling reason why it should be, as opposed to where all the other optimization patterns are done.

PR #8031, opened 2018-03-14
[WIP] fix issue 7804 - Allow __traits(getMember) as a BasicType

ghost:
To allow `alias AggrMember = __traits(getMember, Aggr, "member");` The `BasicType` grammar must be changed, since it's used by `AliasDeclarationY`.

```diff
BasicType:
    BasicTypeX
    . IdentifierList
    IdentifierList
    Typeof
    Typeof . IdentifierList
    TypeCtor ( Type )
    TypeVector
+    Traits
```

As a bonus, declarations work with a `__traits` as type.
Another more limited option would have been to change the grammar of `AliasDeclarationY`:

```diff
AliasDeclarationY:
    Identifier TemplateParametersopt = StorageClassesopt Type
+    Identifier TemplateParametersopt = StorageClassesopt Traits
    Identifier TemplateParametersopt = FunctionLiteral
```

For now, only Types are handled. A second part will allow this for Symbols.

### DIFF
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -4085,6 +4085,24 @@ struct ASTBase
         }
     }
 
+    extern (C++) class TypeTraits : Type
+    {
+        TraitsExp exp;
+        Loc loc;
+
+        extern (D) this(Loc loc, TraitsExp exp)
+        {
+            super(Tident);
+            this.loc = loc;
+            this.exp = exp;
+        }
+
+        override void accept(Visitor v)
+        {
+            v.visit(this);
+        }
+    }
+
     extern (C++) final class TypeIdentifier : TypeQualified
     {
         Identifier ident;
--- a/src/dmd/dmangle.d
+++ b/src/dmd/dmangle.d
@@ -102,6 +102,7 @@ private immutable char[TMAX] mangleChar =
     Tslice       : '@',
     Treturn      : '@',
     Tvector      : '@',
+    Ttraits      : '@',
 ];
 
 unittest
--- a/src/dmd/hdrgen.d
+++ b/src/dmd/hdrgen.d
@@ -784,6 +784,12 @@ public:
         buf.writestring(t.dstring);
     }
 
+    override void visit(TypeTraits t)
+    {
+        //printf("TypeBasic::toCBuffer2(t.mod = %d)\n", t.mod);
+        visit(t.exp);
+    }
+
     override void visit(TypeVector t)
     {
         //printf("TypeVector::toCBuffer2(t.mod = %d)\n", t.mod);
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -334,6 +334,7 @@ enum ENUMTY : int
     Tvector,
     Tint128,
     Tuns128,
+    TTraits,
     TMAX,
 }
 
@@ -381,6 +382,7 @@ alias Tnull = ENUMTY.Tnull;
 alias Tvector = ENUMTY.Tvector;
 alias Tint128 = ENUMTY.Tint128;
 alias Tuns128 = ENUMTY.Tuns128;
+alias Ttraits = ENUMTY.TTraits;
 alias TMAX = ENUMTY.TMAX;
 
 alias TY = ubyte;
@@ -512,6 +514,7 @@ extern (C++) abstract class Type : RootObject
             sizeTy[Terror] = __traits(classInstanceSize, TypeError);
             sizeTy[Tnull] = __traits(classInstanceSize, TypeNull);
             sizeTy[Tvector] = __traits(classInstanceSize, TypeVector);
+            sizeTy[Ttraits] = __traits(classInstanceSize, TypeTraits);
             return sizeTy;
         }();
 
@@ -5232,6 +5235,41 @@ extern (C++) final class TypeDelegate : TypeNext
     override bool hasPointers() const
     {
         return true;
+    }
+
+    override void accept(Visitor v)
+    {
+        v.visit(this);
+    }
+}
+
+/**
+ * This is a shell containing a TraitsExp that can be
+ * either resolved to a type or to a symbol.
+ *
+ * The point is to allow AliasDeclarationY to use `__traits(getMember)`
+ * directly (VS using a library helper in the past).
+ */
+extern (C++) final class TypeTraits : Type
+{
+    Loc loc;
+
+    /// The expression to resolve as type or symbol.
+    TraitsExp exp;
+
+    final extern (D) this(Loc loc, TraitsExp exp)
+    {
+        super(Ttraits);
+        this.loc = loc;
+        this.exp = exp;
+    }
+
+    final override Type syntaxCopy()
+    {
+        TraitsExp te = cast(TraitsExp) exp.syntaxCopy();
+        TypeTraits tt = new TypeTraits(loc, te);
+        tt.mod = mod;
+        return tt;
     }
 
     override void accept(Visitor v)
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -513,6 +513,7 @@ final class Parser(AST) : Lexer
             case TOK.union_:
             case TOK.class_:
             case TOK.interface_:
+            case TOK.traits:
             Ldeclaration:
                 a = parseDeclarations(false, pAttrs, pAttrs.comment);
                 if (a && a.dim)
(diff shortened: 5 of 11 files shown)


src/dmd/parse.d
```diff
@@ -3497,6 +3498,25 @@ final class Parser(AST) : Lexer
...
+            AST.Expression e = parsePrimaryExp();
+            AST.TraitsExp te = cast(AST.TraitsExp) e;
+            if (!te)
+            {
+                error("invalid `__traits` expression");
+                break;
+            }
+            else if (strcmp(te.ident.toChars, "getMember") != 0)
+            {
+                error("invalid `__traits`, only `getMember` can be aliased and not `%s`",
```
jacob-carlborg: Should this really be handled by the parser?

BBasile: No. It's a semantic thing that's handled eagerly.

AndrejMitrovic:
Nice work on handling such an ancient issue!

src/dmd/dmangle.d
```diff
@@ -102,6 +102,7 @@ private immutable char[TMAX] mangleChar =
     Tslice       : '@',
     Treturn      : '@',
     Tvector      : '@',
+    TTraits      : '@',
```
jacob-carlborg: I think this should be `Ttraits` to follow the naming conventions.

src/dmd/parse.d
```diff
@@ -3497,6 +3498,27 @@ final class Parser(AST) : Lexer
             t = parseVector();
             break;
 
+        case TOK.traits:
+        {
+            AST.TraitsExp te;
+            if (AST.Expression e = parsePrimaryExp())
+                te = cast(AST.TraitsExp) e;
```
RazvanN7: How about : 

```d
AST.Traits te = cast(AST.TraitsExp)parsePrimaryExp();
```

src/dmd/parse.d
```diff
@@ -3497,6 +3498,27 @@ final class Parser(AST) : Lexer
...
+                t = new AST.TypeError;
+            }
+            else if (strcmp(te.ident.toChars, "getMember") != 0)
+            {
+                // even if this is not a grammar error, it's not worth continuing.
+                error("invalid `__traits`, only `getMember` can be aliased and not `%s`",
+                    te.ident.toChars);
+                t = new AST.TypeError;
+            }
+            else t = new AST.TypeTraits(loc, te);
```
RazvanN7: Brackets for if/else bodies should be consistent (use `{}` for all or for none).

src/dmd/hdrgen.d
```diff
@@ -784,6 +784,12 @@ public:
         buf.writestring(t.dstring);
     }
 
+    override void visit(TypeTraits t)
```
RazvanN7: A similar visit method should be added in the `ParseTimeTransitiveVisitor` [1] otherwise the tree traversal will be truncated.

[1] https://github.com/dlang/dmd/blob/master/src/dmd/transitivevisitor.d

src/dmd/mtype.d
```diff
@@ -5220,6 +5224,35 @@ extern (C++) final class TypeDelegate : TypeNext
     }
 }
 
+/**
+ */
+extern (C++) final class TypeTraits : Type
```
RazvanN7: Please add some documentation for this class : why is it needed, when it is created and what information is kept in each field. Hope that is the intent with the empty comments :+1:

RazvanN7 (changes requested):
Cool addition! Thanks!

src/dmd/mtype.d
BBasile: Indeed i planned to do it but forgot to.

BBasile:
First part, handling of types, is done now. Will start the second soon.

src/dmd/parse.d
```diff
@@ -3497,6 +3498,28 @@ final class Parser(AST) : Lexer
...
 
+        case TOK.traits:
+        {
+            AST.TraitsExp te = cast(AST.TraitsExp) parsePrimaryExp();
+            if (!te)
+            {
+                // error already emitted while parsing primary
+                t = new AST.TypeError;
+            }
+            else if (strcmp(te.ident.toChars, "getMember") != 0)
```
WalterBright: strcmp must die. Really, everyone, please stop using it! Use `if (te.ident == Id.getMember)`.

src/dmd/typesem.d
```diff
@@ -1307,6 +1307,21 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
...
+        import dmd.traits : semanticTraits;
+        import dmd.dtemplate : getType;
+
+        result = null;
+        if (Expression e = semanticTraits(mtype.exp, sc))
+            if (Type t = getType(e))
+                result = t.addMod(mtype.mod);
+        //if (result) printf("TypeTraits found to be %s\n".ptr, result.toChars);
+        if (result is null)
+            mtype.error(loc, "`%s` cannot be resolved to a valid type",
```
WalterBright: definitely need a test case for this

src/dmd/mtype.d
```diff
@@ -334,6 +334,8 @@ enum ENUMTY : int
     Tvector,
     Tint128,
     Tuns128,
+
```
WalterBright: no purpose to blank line

src/dmd/parse.d
wilzbach: Would be good to approve https://github.com/dlang/dmd/pull/7893 then, s.t. we can start using DStrings everywhere.

src/dmd/hdrgen.d
```diff
@@ -784,6 +784,12 @@ public:
...
+    {
+        //printf("TypeBasic::toCBuffer2(t.mod = %d)\n", t.mod);
+        visit(t.exp);
```
WalterBright: need test case

src/dmd/mtype.d
```diff
@@ -5220,6 +5224,42 @@ extern (C++) final class TypeDelegate : TypeNext
     }
 }
 
+/**
+ * This type is a shell containing a TraitsExp that can be
```
WalterBright: `This type is a` is completely redundant and can be elided.

wilzbach:
(@reviewers: don't forget that this requires a PR to dlang.org with the spec update before this can be merged)

src/dmd/mtype.d
```diff
@@ -5220,6 +5224,42 @@ extern (C++) final class TypeDelegate : TypeNext
...
+    Loc loc;
+
+    /// The expression to resolve as type or symbol.
+    TraitsExp exp;
+
+    final extern (D) this(Loc loc, TraitsExp exp)
+    {
+        this.loc = loc;
+        this.exp = exp;
+        super(Ttraits);
```
WalterBright: super call should go first.

src/dmd/mtype.d
```diff
@@ -5220,6 +5224,42 @@ extern (C++) final class TypeDelegate : TypeNext
...
+    }
+
+    final override Type syntaxCopy()
+    {
+        TypeTraits tt;
+        TraitsExp te = cast(TraitsExp) exp.syntaxCopy();
+        tt = new TypeTraits(loc, te);
```
WalterBright: Declare `tt` here, not on line 5250

src/dmd/astbase.d
```diff
@@ -4085,6 +4085,24 @@ struct ASTBase
         }
     }
 
+    extern (C++) class TypeTraits : Type
```
WalterBright: Needs an entry in `visitor.d`, too.

src/dmd/parse.d
```diff
@@ -3497,6 +3498,28 @@ final class Parser(AST) : Lexer
             t = parseVector();
             break;
 
+        case TOK.traits:
```
WalterBright: Also needs entry in `isBasicType()`, `parseMixin()`, `parseStatement()`, `parsePrimaryExp()` and `parseUnaryExp()`. Basically, just look at where `TOK.typeof_` appears, which is very similar.

WalterBright:
I approve of the concept, but this needs much work. It'll need 100% test coverage. It also needs a changlog entry, and a spec change.

src/dmd/astbase.d
BBasile: If i do so i get a conflict with the entry in `transitivevisitor.d`

BBasile:
I give the hand if someone wants to continue. It's not too hard to finish but i don't want to work anymore.

BBasile:
This was almost finished actually, the second part didn't require to add a new AST node.
Unfortunately i have to open another PR because i have forced pushed.

PR #8032, opened 2018-03-14
Fix Issue 18417 - Make const and immutable postblit constructors illegal

RazvanN7:


src/dmd/parse.d
```diff
@@ -2365,8 +2365,11 @@ final class Parser(AST) : Lexer
             check(TOK.rightParentheses);
 
             stc = parsePostfix(stc, &udas);
-            if (stc & AST.STC.static_)
-                error(loc, "postblit cannot be static");
+            if (stc & AST.STC.immutable_ || stc & AST.STC.shared_ ||
+                stc & AST.STC.const_ || stc & AST.STC.static_)
+            {
+                error(loc, "postblit cannot be `const`/`immutable`/`shared`/`static`");
```
jacob-carlborg: * Should probably be a depreciation
* Specifying the exact attribute used in the source code in the message would be nicer

jacob-carlborg (changes requested):

src/dmd/parse.d
RazvanN7: It doesn't have to be a deprecation since marking a postblit as `const/immutable/shared` makes it impossible to call the copy constructor, so no code can be broken whatsoever.

mathias-lang-sociomantic: Except it only happen if the code is called. If the postblit is not called, you're breaking the build.

RazvanN7: Indeed, but having a const/immutable postblit which is not called is in itself an indication of design issues. Anyway, I don't mind making it a deprecation, rather I honestly think it's useless.

andralex: This is an accept-invalid bug, and virtually any occurrence of the syntax indicates a bug. I'll allow without a deprecation cycle. @MartinNowak agree?

andralex (approved):

jacob-carlborg:
Should at least have a changelog entry.

src/dmd/parse.d
jacob-carlborg: I don't see the harm in deprecating it, even if it's an accept-invalid bug. It's trivial to do and, perhaps, we don't need a full deprecation cycle. How about deprecating it in for one release and make it an error in the next?

What does we gain from _not_ deprecating it?

mathias-lang-sociomantic: @andralex : That's based on the assumption that the error is beneficial to the user, which often it is not.

Example:
Vibe.d has this code somewhere in a module, on a type that's not used.
1) You make this a deprecation: I get a deprecation from Vibe.d but my project keeps building. By the time the deprecation period ends, the deprecation will be fixed. Everyone is happy.
2) You make it an error: My project does not build anymore because of a code I do not use nor care about. I cannot use this compiler release until Vibe.d gets fixed. In addition, forcing me to upgrade Vibe.d might force me to upgrade other dependencies, if the same philosophy is applied. I spend the whole day upgrading the dependencies of my project, trying to find a compatible subset, because someone decided that "it's a bug anyway".

andralex: @mathias-lang-sociomantic I very well understand. I just want to avoid bureaucracy unless truly needed.

andralex:
@RazvanN7 can you please add a changelog

RazvanN7:
@jacob-carlborg Added a changelog entry and made the error specify the exact attribute.

RazvanN7:
@andralex Indeed, it looks like there are 2 projects in jenkins that have a const, respectively shared postblit. Should we first deprecate this or announce people they have a bug in their code?

changelog/postblitQualifiers.dd
```diff
@@ -0,0 +1,3 @@
+Make const/immutable/shared postblit an error
+
+Before this patch, if a postblit was declared const/immutable/shared the compiler would have accepted the declaration but there would have been no way of calling the postblit succesfully (e.g. whenever the postblit was called, an error was issued). The patch makes it so that if a postblit contains const/immutable/shared in its signature an error will occur.
```
jacob-carlborg: I would like the changelog entry to explicitly mention that this is a breaking change.

jacob-carlborg (changes requested):

src/dmd/parse.d
mathias-lang-sociomantic: This is a breaking change. We have a process for handling breaking change. Why is that so hard to follow ?

RazvanN7:
Errors are now deprecation. @jacob-carlborg @andralex

jacob-carlborg (approved):

andralex (approved):

andralex:
@RazvanN7 Cool, can you check the autotester?

andralex:
OK, will force pull. cc @braddr

MartinNowak:
Not allowing shared postblits is a bit less obvious and would have deserved an explanation in the changelog.

Also wouldn't `shared(RC!T)` seems like a reasonable type, using atomic ref-counting, while the normal `RC!T` would use non-atomic ref-counting? This relies on the bit-blit copy to be atomic, but I ruling this out seems a bit harsh. In particular since many lock-free algorithms rely on some sort of lazy fixup.

@andralex @RazvanN7 please check whether disallowing shared postblit overloads is warranted.

andralex:
@MartinNowak My understanding is shared postblit could not even be called successfully unless the object was a field.

MartinNowak:
> My understanding is shared postblit could not even be called successfully unless the object was a field.

Not sure I understand what you're trying to say.

I think the proper reason is that shared postblit implies read-modify-write, hence was disallowed as the other read-modify-write operations.
Also my example above would imply read-modify-write and hence is incorrect. `shared(RC!T)` should still be implementable through copy construction (`this(shared RC!T other)`) and `opAssign(shared RC!T other)`.

MartinNowak:
> post-blit doing atomic increments of the reference count should be used for the case RC!(shared T).

Doesn't make sense, the ref-count is part of the `RC` not the value of type `T`, hence `shared(RC)` should get the atomic ref-counting.
Also an `RC!(shared T)` can't be shared among threads, e.g. by storing it in a shared static variable.

ZombineDev:
> Also an `RC!(shared T)` can't be shared among threads, e.g. by storing it in a shared static variable.

Exactly. It shouldn't be directly sharable across threads - only the payload is typed as `shared`. To *share* a reference to the payload you could send `RC!(shared T)` by value to other threads through message passing.

> shared(RC!T) should get the atomic ref-counting

Yes, but that's because `shared` is transitive and the payload would be typed as `shared`.

Think about it - if `T` is a RAII wrapper for some resource we need to ensure that `T`'s destructor is called only by a single thread. If the smart pointer doesn't use necessary precautions (like atomic ref counting) multiple thread could enter `T` destructor concurrently.

Let's say we have a global variable `shared(RC!T) global;` which is obviously shared by multiple threads. When a thread wants to operate with the current payload that `global` is pointing to, it needs to atomically load the pointer, otherwise on some architectures the read may be teared. It could do so like this:
```D
RC!(shared T)` local = global.atomicLoad();
```
I think you would agree that `local` should continue to use **the same** synchronization mechanism as `global`, otherwise we would have an obvious data race.

MartinNowak:
Thanks, good explanation.

> I think you would agree that local should continue to use the same synchronization mechanism as global, otherwise we would have an obvious data race.

Yes of course, I wouldn't argue though that atomic loading would strip the head shared.
Plain reading of a global `shared(T)` remains a `shared(T)` and `atomicLoad` was recently adjusted to not strip tail shared https://github.com/dlang/druntime/pull/1605.

OT: Our type system doesn't currently allow for qualifiers to move from `qualifier(RC!(T))` to `RC!(qualifier T)`, e.g. see the dreaded inout range discussion.

PR #8034, opened 2018-03-14
Fix issue 15206 - SROA doesn't play well with double slices

LemonBoy:
The double-access pattern in the example makes the SROA pass generate a first slice as `TYdouble` and another one as `TYnptr`, the latter should be `TYdouble` as well.

I'm not sure this is the right fix so let's see what Jerkins thinks of it.
CC @WalterBright

### DIFF
--- a/src/dmd/backend/gsroa.d
+++ b/src/dmd/backend/gsroa.d
@@ -113,6 +113,11 @@ private void sliceStructs_Gather(SymInfo *sia, elem *e)
                             {
                                 sia[si].canSlice = false;
                             }
+                            // Disable SROA on OSX32 (because XMM registers?)
+                            else if (!(config.exe & EX_OSX))
+                            {
+                                sliceStructs_Gather(sia, e.EV.E1);
+                            }
                         }
                         return;
                     }
--- a/test/compilable/b15206.d
+++ b/test/compilable/b15206.d
@@ -1,0 +1,19 @@
+// PERMUTE_ARGS:
+// REQUIRED_ARGS: -O
+
+void main()
+{
+}
+
+struct Line
+{
+    double slope;
+    double intercept;
+}
+
+Line nLineProjection(double[] historicData)
+{
+    Line projLine;
+    projLine.intercept = historicData[$-1] + projLine.slope;
+    return projLine;
+}


WalterBright:
When creating a PR to fix a bugzilla issue, please post a link to the PR in the issue. This is so that someone browsing bug reports will know if and where a PR for it is.

JinShil:
@LemonBoy Is this still a WIP?

LemonBoy:
@JinShil Yes, I'm not sure about what the best way to handle this case is...
Paging @WalterBright since this is backend stuff.

LemonBoy:
Moreover the SROA pass is also responsible for [18489](https://issues.dlang.org/show_bug.cgi?id=18489).

WalterBright (approved):

WalterBright:
Looks like it's failing in test42.d on 32 bit OSX.

LemonBoy:
No idea why it's failing (and don't have a mac)

WalterBright:
It probably has something to do with OSX 32 being the only 32 bit platform that uses XMM registers.

WalterBright:
I can't reproduce bug 15206, either.

wilzbach:
Cross-posting from Bugzilla

1) John Colvin posted another test case that reproduces this:

```d
alias A = Complex!double;

struct Complex(T)
{
    double re, im;

    Complex!double baz()
    {
        return Complex!double(re).blah;
    }

    ref Complex!double blah()
    {
        im = re;
        return this;
    }
}
```
(with `-O -inline -c`)

2) This regression was introduced in 2.077 by https://github.com/dlang/dmd/pull/7175
3) Reproducing this is easy e.g. https://run.dlang.io/is/REkj3X (John's test case)

jacob-carlborg:
I can reproduce the test case in this PR on macOS 64-bit and when the compiler is compiled as 32-bit. But not when cross-compiling from 64-bit to 32-bit. I don't see the issue that the auto-tester is having. I get a different failed test: cpp_abi_tests.d(94). Might be due to I'm using Clang instead of GCC.

jacob-carlborg (approved):

wilzbach:
Removing auto-merge as OSX_64_32 is a permanently failing.

n8sh:
```diff
+                            else
+                            {
+                                sliceStructs_Gather(sia, e->E1);
+                            }
```
Is it an option to disable this branch and the test case on OSX_64_32 only?

thewilsonator:
Rebased. Disabling this on OSX32.

PR #8036, opened 2018-03-15
fix Issue 18484 - [dip1000] Subtype allows reference to escape with i…

WalterBright:
…mplicit casting

### DIFF
--- a/test/fail_compilation/test18484.d
+++ b/test/fail_compilation/test18484.d
@@ -1,0 +1,26 @@
+/* REQUIRED_ARGS: -dip1000
+TEST_OUTPUT:
+---
+fail_compilation/test18484.d(19): Error: returning `x.bar()` escapes a reference to local variable `x`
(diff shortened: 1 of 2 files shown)


test/fail_compilation/test18484.d
```diff
@@ -0,0 +1,26 @@
...
+
+int* test1()
+{
+    auto x = S(); return x.bar();  // error
+}
+
+int* test2()
+{
+    return S().bar();  // error
+}
```
mathias-lang-sociomantic: This code is not marked as `@safe`, how come it errors out ?

WalterBright: Because it's obviously wrong. Even unsafe code gets some checks!

JinShil (approved):

test/fail_compilation/test18484.d
mathias-lang-sociomantic: So far, DIP1000 only affected `@safe` code (or at least that was the state intention). It looks like this change also change behavior in non-`@safe` code, which is very odd to me.

PR #8037, opened 2018-03-15
Improve error message 'for {mutable,non-shared} method is not callable using X'

wilzbach:
Mutable method is not callable using const object is a common message (see e.g. https://forum.dlang.org/post/blwbenjxxblercjphozz@forum.dlang.org), so let's be a bit more friendly to newbies and provide helpful advice directly in the error message.

### DIFF
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -2436,8 +2436,33 @@ extern (C++) int overloadApply(Dsymbol fstart, void* param, int function(void*, 
     return overloadApply(fstart, s => (*fp)(param, s));
 }
 
-void MODMatchToBuffer(OutBuffer* buf, ubyte lhsMod, ubyte rhsMod)
+/**
+Checks for mismatching modifiers between `lhsMod` and `rhsMod` and prints the
+mismatching modifiers to `buf`.
+
+The modifiers of the `lhsMod` mismatching the ones with the `rhsMod` are printed, i.e.
+lhs(shared) vs. rhs() prints "`shared`", wheras lhs() vs rhs(shared) prints "non-shared".
+
+Params:
+    buf = output buffer to write to
+    lhsMod = modifier on the left-hand side
+    lhsMod = modifier on the right-hand side
+
+Returns:
+
+A tuple with `isMutable` and `isNotShared` set
+if the `lhsMod` is missing those modifiers (compared to rhs).
+*/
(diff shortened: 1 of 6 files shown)


src/dmd/func.d
```diff
@@ -2615,9 +2615,14 @@ extern (C++) FuncDeclaration resolveFuncCall(const ref Loc loc, Scope* sc, Dsymb
                 }
                 else
                 {
+                    auto fullFdPretty = fd.toPrettyChars();
                     .error(loc, "%smethod `%s` is not callable using a %sobject",
-                        funcBuf.peekString(), fd.toPrettyChars(),
+                        funcBuf.peekString(), fullFdPretty,
                         thisBuf.peekString());
+                    if (funcBuf.peekSlice() == "mutable ")
```
jacob-carlborg: Isn’t there a better way to do this than comparing strings?

wilzbach: Then I would have to duplicate this logic:

https://github.com/dlang/dmd/blob/be225ecd2d7dab2dfa878e3ed2c2fc0b39edbdab/src/dmd/func.d#L2439-L2461

wilzbach: though `MODMatchToBuffer` returns `void` and I could modify it to return something more useful like:

```d
struct Errors {
    bool isNotShared;
    bool isMutable;
}
```

jacob-carlborg: Yes, exactly.

wilzbach: OK done.

jacob-carlborg (approved):

src/dmd/func.d
```diff
@@ -2436,16 +2436,44 @@ extern (C++) int overloadApply(Dsymbol fstart, void* param, int function(void*,
...
+    buf = output buffer to write to
+    lhsMod = modifier on the left-hand side
+    lhsMod = modifier on the right-hand side
+
+Returns:
+
+A tuple with `isMutable` and `isNotShared` set
+if the `lhsMod` is missing those modifiers (compared to rhs).
+*/
+auto MODMatchToBuffer(OutBuffer* buf, ubyte lhsMod, ubyte rhsMod)
```
JinShil: `ref OutBuffer` instead of `OutBuffer*`?

wilzbach: One change at a time?
(This is an existing function)

JinShil: Ah, I overlooked it was existing.

JinShil (approved):

PR #8038, opened 2018-03-15
Add and use RESULTS_TEST_DIR + OUTPUT_BASE + EXTRA_FILES in the bash scripts of DMD's testsuite

wilzbach:
Turns out we can save quite a lot when we move the redundant definitions in the bash files to the new wrapper.
This also opens the opportunity to run the shell script in a temporary directory [1] or do other things as now most scripts don't have any hard-coded information anymore.

I'm not fixed about the names `EXTRA_FILES` could also be named `SRC` for example.

[1] this would mean e.g. that no manual cleaning of the generated files is necessary anymore, but it's just one potential opportunity. The big motivation of this PR is to remove the redundant declarations and thus let the test scripts focus on what's important: testing and not startup/cleanup code.

CC @marler8997 @timotheecour @CyberShadow

PR #8040, opened 2018-03-16
Add a convenience script for running individual tests

wilzbach:
After typing `test_results` and adding `.out` over an over again, I was so annoyed that I finally automated this and though this is worth sharing as others might miss this too.

Usage:

```sh
./run_single runnable/template2962.d fail_compilation/fail282.d
```

Reasons:
- safe time
- auto-completion in CLI works

If we plan to move the documentation to a README.md as proposed [here](https://github.com/dlang/dmd/pull/8038#issuecomment-373524698), this could also be documented though I think it's rather self-explanatory.

This might be a small addition, but it saves a bit of time for me every time and that quickly adds up.

### DIFF
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -164,6 +164,12 @@ check_clean_git()
     make -f posix.mak check-clean-git
 }
 
+# sanitycheck for the run_individual_tests script
+check_run_individual()
+{
+	./test/run_individual_tests test/runnable/template2962.d ./test/compilable/test14275.d
+}
+
 codecov()
 {
     # CodeCov gets confused by lst files which it can't matched
@@ -185,6 +191,7 @@ case $1 in
     all)
         coverage;
         check_clean_git;
+        check_run_individual;
         codecov;
     ;;
 esac
--- a/test/run_individual_tests
+++ b/test/run_individual_tests
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Allows to run a individual tests
+# Allows running tests individually
 # Usage:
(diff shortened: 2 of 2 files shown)


test/run_single
```diff
@@ -0,0 +1,37 @@
+#!/usr/bin/env bash
+# Allows to quickly run a single tests
```
RazvanN7: "Allows to run single tests". The use of quickly is redundant

RazvanN7 (commented):
Loooks good, but I would rename it `run_custom_tests`

wilzbach:
> I would rename it run_custom_tests

Ok. (I don't really care about the name of this script)

jacob-carlborg:
> Loooks good, but I would rename it run_custom_tests

`custom` sounds to me like my own test that I personally want to run that no one else runs.

wilzbach:
> custom sounds to me like my own test that I personally want to run that no one else runs.

(I'm happy about better name suggestion - as mentioned I don't really care about the name)

RazvanN7:
>  custom sounds to me like my own test that I personally want to run that no one else runs.

`run_selected_tests` , `run_hand_picked_tests` , `run_chosen_tests` ?

JinShil:
Yeah, `run_custom_tests` isn't right, `run_selected_tests` isn't bad, but I would prefer `run_individual_tests`.

wilzbach:
> I would prefer run_individual_tests.

OK. Renamed. Anything else blocking this?

JinShil (approved):

PR #8041, opened 2018-03-16
Rename onlysafeaccess to onlyunsafeaccess to avoid confusion

wilzbach:
From: https://github.com/dlang/dmd/pull/8035#discussion_r174677577

CC @JinShil

### DIFF
--- a/src/dmd/access.d
+++ b/src/dmd/access.d
@@ -189,9 +189,9 @@ extern (C++) bool checkAccess(AggregateDeclaration ad, Loc loc, Scope* sc, Dsymb
             printf("result4 = %d\n", result);
         }
     }
-    if (!result && (!(sc.flags & SCOPE.onlysafeaccess) || sc.func.setUnsafe()))
+    if (!result && (!(sc.flags & SCOPE.onlyunsafeaccess) || sc.func.setUnsafe()))
     {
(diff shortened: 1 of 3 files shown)


JinShil (dismissed):

WalterBright:
I have a hard time seeing an improvement. "onlysafeaccess" means that only safe accesses are allowed. "onlyunsafeaccess" means only unsafe accesses are allowed???

wilzbach:
> "onlysafeaccess" means that only safe accesses are allowed. "onlyunsafeaccess" means only unsafe accesses are allowed???

Yes, currently the code uses this check to error in @safe, hence only @system is allowed.

I don't care about the exact naming, another one would be onlysystemaccess, but using onlysafeaccess when the code actually errors with @safe is going to hit people in the future.

rainers:
FWIW: I agree with Walter, `onlysafeaccess` allows only safe access and rejects accesses that need `@system`. What is unclear about that?

JinShil:
> Yes, currently the code uses this check to error in @safe, hence only @System is allowed.

Right.  Look at this code:  `!(sc.flags & SCOPE.onlysafeaccess) || sc.func.setUnsafe())`.  One condition uses the term *unsafe* and the other users *onlysafe*.  It's like it was written by a politician.

JinShil:
Comment above without the typos:  I had to twist my mind a little, but now I see it. I'm withdrawing my approval. @wilzbach, I think we can close this, it's going to be consusing either way depending on how one thinks about it.

PR #8042, opened 2018-03-16
Deprecate class allocators and deallocators

JinShil:
`delete` was deprecated in 2.079.  Before the `delete` keyword can be removed it will be necessary to also deprecate and remove class allocators and deallocators.  They have been planned for deprecation for years; this PR finally puts the deprecation process in motion.

See https://dlang.org/deprecate.html#Class%20allocators%20and%20deallocators

dlang.org PR:  https://github.com/dlang/dlang.org/pull/2277

### DIFF
--- a/changelog/deprecate_class_allocators.dd
+++ b/changelog/deprecate_class_allocators.dd
@@ -17,7 +17,7 @@ class C
 }
 ---
 
-For more information see https://dlang.org/deprecate.html#Class%20allocators%20and%20deallocators
+See the $(LINK2 $(ROOT_DIR)deprecate.html#Class%20allocators%20and%20deallocators, Deprecated Features) for more information.
 
 Many alternatives for class allocators/deallcators exist. Among them is the generic $(REF make, std,experimental,allocator) and $(REF dispose, std,experimental,allocator) from the allocator package. For other alternatives, see $(LINK2 https://dlang.org/blog/the-gc-series, the recent article about memory allocation on the DBlog) and $(LINK2 https://wiki.dlang.org/Memory_Management, the D Wiki memory management article).
 
--- a/changelog/deprecate_class_allocators.dd
+++ b/changelog/deprecate_class_allocators.dd
@@ -1,0 +1,50 @@
+Class allocators and deallocators have been deprecated
+
+Class allocators and deallocators have been planned for deprecation for years.  Starting with this release the following code will emit deprecation messages.
+
+---
+class C
+{
+    new(size_t size)         // deprecation message
+    {
+        return malloc(size);
+    }
+
+    delete(void* obj)        // deprecation message
+    {
+        free(obj);
+    }
+}
+---
+
+For more information see https://dlang.org/deprecate.html#Class%20allocators%20and%20deallocators
+
+Many alternatives for class allocators/deallcators exist. Among them is the generic $(REF make, std,experimental,allocator) and $(REF dispose, std,experimental,allocator) from the allocator package. For other alternatives, see $(LINK2 https://dlang.org/blog/the-gc-series, the recent article about memory allocation on the DBlog) and $(LINK2 https://wiki.dlang.org/Memory_Management, the D Wiki memory management article).
+
+Users have leveraged allocators in order to disable GC allocation, as illustrated in the following example:
+
+---
+class C
+{
+    @disable new(size_t size);
+}
+
+void main()
+{
+    auto c = new C();  // Error: allocator `new` is not callable because it is annotated with `@disable`
+}
+---
+
+That idiom will remain, but has been enhanced with this release to no longer require the `size_t` argument.  That is, starting with this release, the following syntax is also permitted:
+
+---
+class C
+{
+    @disable new();
+}
+
+void main()
+{
+    auto c = new C();  // Error: allocator `new` is not callable because it is annotated with `@disable`
+}
+---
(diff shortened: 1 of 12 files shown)


changelog/deprecate_class_allocators.dd
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    delete(void* obj)        // deprecation message
+    {
+        free(obj);
+    }
+}
+---
+
+For more information see https://dlang.org/deprecate.html#Class%20allocators%20and%20deallocators
```
wilzbach: Probably a good idea to point users to alternatives like std.experimental.allocator and other RAII articles.
See also: https://issues.dlang.org/show_bug.cgi?id=18613

src/dmd/dsymbolsem.d
```diff
@@ -3897,6 +3897,12 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
     override void visit(NewDeclaration nd)
     {
         //printf("NewDeclaration::semantic()\n");
+
+        // @@@DEPRECATED_2019-05@@@
```
wilzbach: We recently switched to using an explicit version, see also https://github.com/dlang/DIPs/pull/108 or https://github.com/dlang/phobos/pull/6286 for an example

ghost (changes requested):
Please don't do this for class allocators. We have recently [discovered](https://github.com/dlang/phobos/pull/6003) that they can be used to prevent the usage of new (when marked `@disable` hence forcing people to use custom  phobos-style allocators.

I planned to add this to my user library but forgot to. This user library is now in the project tester so this change would really be embarassing.

For deallocators ok since `delete` is deprecated.

changelog/deprecate_class_allocators.dd
JinShil: Please enumerate specifically which articles I should included and I'll add them.

JinShil:
I've added support for the syntax `@disable new();`.  You won't even need to add the useless `size_t` argument.  That should prevent any breaking changes.

ghost (approved):
This new version allows the `@disable` trick so it's good.

changelog/deprecate_class_allocators.dd
```diff
@@ -0,0 +1,20 @@
...
+        return malloc(size);
+    }
+
+    delete(void* obj)        // deprecation message
+    {
+        free(obj);
+    }
+}
+---
+
```
wilzbach: The new `@disable new();` syntax should be mentioned here.

changelog/deprecate_class_allocators.dd
wilzbach: Probably this page is the best we have at the moment - https://dlang.org/blog/the-gc-series

It doesn't really mention stdx.allocator, so I would probably mentioned that separately.

How about:

> Many alternatives for class allocators/deallcators exist, among many is the generic $(REF make, std,experimental,allocator) and $(REF dispose, std,experimental,allocator) from the allocator package. For other alternatives, see $(LINK2 https://dlang.org/blog/the-gc-series, the recent article about memory allocation on the DBlog)

(just an idea)

JinShil: Done.  Thanks!

changelog/deprecate_class_allocators.dd
JinShil: Done.

wilzbach (approved):
Nice!

JinShil:
I'm labeling this as "Vision", as I think it corresponds to item 1 in the stated vision.

> Lock down the language definition:   D is a powerful language but its definition is not precise enough. A number of subtleties can only be assessed only by running the compiler, not by perusing the specification. This semester we are pushing for a better, more precise, and more complete specification of the D language. 

While that statement was probably intended largely for documentation, there are two sides to that coin:  1. ensuring the specification documents the implementation, 2. ensuring the code implements the documentation; they are one and the same.  

Deprecations are updates to the specification.  Documenting that something is deprecated without actually implementing said deprecation is to disregard the specification.

Moving these deprecations forward is locking down the language, and doing it sooner rather than later avoids problems that crop up when users choose to leverage a deprecated feature in some unexpected way, forcing us to [make accomodations](https://github.com/dlang/dmd/pull/8042#issuecomment-373834698) later.

wilzbach:
@WalterBright @andralex I think your approval for this deprecation is required?
(it's more a formal thing because it has been on the list of to be deprecated for a long time)

RazvanN7 (approved):

andralex:
Affirmative. Let's move with deprecation. Thanks.

andralex (approved):

wilzbach:

There was one spurious failure on OSX in druntime

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3098596&isPull=true

````
Testing invalid_memory_operation
timelimit -t 10 ./generated/osx/debug/32/invalid_memory_operation  2>&1 1>/dev/null | grep -qF "InvalidMemoryOperationError"
timelimit -t 10 generated/osx/release/32/unittest/test_runner core.sys.dragonflybsd.netinet.in_
make[2]: *** [generated/osx/debug/32/invalid_memory_operation.done] Error 1
rm generated/osx/debug/32/unknown_gc generated/osx/debug/32/stderr_msg generated/osx/debug/32/static_dtor generated/osx/debug/32/future_message generated/osx/debug/32/invalid_memory_operation
make[1]: *** [test/exceptions/.run] Error 2
make: *** [unittest-debug] Error 2
```

Let's see whether this is repeats itself - otherwise "merge ahoy!"

PR #8043, opened 2018-03-17
Support Markdown backslash escapes

dgileadi:
This is the first of several PRs to support Markdown syntax in DDoc.

There are corresponding PRs for [dlang.org](https://github.com/dlang/dlang.org/pull/2283), [phobos](https://github.com/dlang/phobos/pull/6303) and [druntime](https://github.com/dlang/druntime/pull/2150).

### DIFF
--- a/res/default_ddoc_theme.ddoc
+++ b/res/default_ddoc_theme.ddoc
@@ -3,7 +3,6 @@ RPAREN = )
 BACKTICK = `
 DOLLAR = $
 COMMA = ,
-QUOTE = &quot;
 LF =
 $(LF)
 
--- a/res/default_ddoc_theme.ddoc
+++ b/res/default_ddoc_theme.ddoc
@@ -2,6 +2,8 @@ LPAREN = (
 RPAREN = )
 BACKTICK = `
 DOLLAR = $
+COMMA = ,
+QUOTE = &quot;
 LF =
 $(LF)
 
--- a/src/dmd/doc.d
+++ b/src/dmd/doc.d
@@ -146,8 +146,8 @@ extern (C++) class Section
     L1:
         size_t o = buf.offset;
         buf.write(_body, bodylen);
-        escapeStrayParenthesis(loc, buf, o);
-        highlightText(sc, a, buf, o);
+        escapeStrayParenthesis(loc, buf, o, true);
+        highlightText(sc, a, loc, buf, o);
         buf.writestring(")");
     }
 }
@@ -252,7 +252,7 @@ extern (C++) final class ParamSection : Section
                             }
                             buf.write(namestart, namelen);
                         }
-                        escapeStrayParenthesis(loc, buf, o);
+                        escapeStrayParenthesis(loc, buf, o, true);
                         highlightCode(sc, a, buf, o);
                     }
                     buf.writestring(")");
@@ -260,8 +260,8 @@ extern (C++) final class ParamSection : Section
                     {
                         size_t o = buf.offset;
                         buf.write(textstart, textlen);
-                        escapeStrayParenthesis(loc, buf, o);
-                        highlightText(sc, a, buf, o);
+                        escapeStrayParenthesis(loc, buf, o, true);
+                        highlightText(sc, a, loc, buf, o);
                     }
                     buf.writestring(")");
                 }
@@ -419,6 +419,8 @@ extern (C++) void gendocfile(Module m)
     if (m.isDocFile)
     {
         Loc loc = m.md ? m.md.loc : m.loc;
+        if (!loc.filename)
+            loc.filename = srcfilename;
         size_t commentlen = strlen(cast(char*)m.comment);
         Dsymbols a;
         // https://issues.dlang.org/show_bug.cgi?id=9764
@@ -429,7 +431,7 @@ extern (C++) void gendocfile(Module m)
             dc.macros.write(loc, dc, sc, &a, &buf);
         }
         buf.write(m.comment, commentlen);
-        highlightText(sc, &a, &buf, 0);
+        highlightText(sc, &a, loc, &buf, 0);
     }
     else
     {
@@ -551,7 +553,7 @@ extern (C++) void escapeDdocString(OutBuffer* buf, size_t start)
  *
  * Fix by replacing unmatched ( with $(LPAREN) and unmatched ) with $(RPAREN).
  */
-extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start, bool respectMarkdownEscapes = true)
+extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start, bool respectMarkdownEscapes)
 {
     uint par_open = 0;
     bool inCode = 0;
@@ -1754,8 +1756,8 @@ struct DocComment
                 buf.writestring("$(DDOC_SUMMARY ");
                 size_t o = buf.offset;
                 buf.write(sec._body, sec.bodylen);
-                escapeStrayParenthesis(loc, buf, o);
-                highlightText(sc, a, buf, o);
+                escapeStrayParenthesis(loc, buf, o, true);
+                highlightText(sc, a, loc, buf, o);
                 buf.writestring(")");
             }
             else
@@ -1786,7 +1788,7 @@ struct DocComment
                     buf.writestring("----\n");
                     buf.writestring(codedoc);
                     buf.writestring("----\n");
-                    highlightText(sc, a, buf, o);
+                    highlightText(sc, a, loc, buf, o);
                 }
                 buf.writestring(")");
             }
@@ -1835,25 +1837,6 @@ extern (C++) bool isDitto(const(char)* comment)
             return true;
     }
     return false;
-}
-
-/*****************************************
- * Return true if the given character is punctuation.
- */
-private bool isPunctuation(const(char) c)
-{
-    static immutable bool[char.max+1] asciiPunctuation =
-       ['!': true, '"': true, '#': true,
-        '$': true, '%': true, '&': true, '\'': true, '(': true, ')': true,
-        '*': true, '+': true, ',': true, '-': true, '.': true, '/': true,
-        ':': true, ';': true, '<': true, '=': true, '>': true, '?': true,
-        '@': true, '[': true, '\\': true, ']': true, '^': true, '_': true,
-        '`': true, '{': true, '|': true, '}': true, '~': true];
-    return asciiPunctuation[c];
-// TODO: unicode punctuation: Pc, Pd, Pe, Pf, Pi, Po, or Ps.
-// However, unicode punctuation should not be included for Markdown backslash
-// escapes, so when it's implemented be sure to make the unicode punctuation
-// check optional via a parameter
 }
 
 /**********************************************
@@ -2181,9 +2164,13 @@ extern (C++) bool isReservedName(const(char)* str, size_t len)
 /**************************************************
  * Highlight text section.
  */
-extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t offset)
+extern (C++) void highlightText(Scope* sc, Dsymbols* a, Loc loc, OutBuffer* buf, size_t offset)
 {
     Dsymbol s = a.dim ? (*a)[0] : null; // test
+    bool incrementLoc = loc.linnum == 0;
+    if (incrementLoc)
+        ++loc.linnum;
+    loc.charnum = 0;
     //printf("highlightText()\n");
     int leadingBlank = 1;
     int inCode = 0;
@@ -2222,6 +2209,8 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
             }
             leadingBlank = 1;
             iLineStart = i + 1;
+            if (incrementLoc)
+                ++loc.linnum;
             break;
         case '<':
             {
@@ -2447,12 +2436,17 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
 
         case '\\':
         {
-            /* Escape Markdown special characters */
+            leadingBlank = false;
             if (inCode || i+1 >= buf.offset)
                 break;
+
+            /* Escape Markdown special characters */
             char c1 = buf.data[i+1];
-            if (isPunctuation(c1))
+            if (ispunct(c1))
             {
+                if (global.params.vmarkdown)
+                    message(loc, "Ddoc: backslash-escaped %c", c1);
+
                 buf.remove(i, 1);
 
                 auto se = sc._module.escapetable.escapeChar(c1);
@@ -2463,10 +2457,9 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
                     const len = strlen(se);
                     buf.remove(i, 1);
                     i = buf.insert(i, se, len);
-                    i--; // point to ';'
+                    i--; // point to escaped char
                 }
             }
-            leadingBlank = false;
             break;
         }
 
@@ -2555,7 +2548,7 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
         }
     }
     if (inCode)
-        error(s ? s.loc : Loc.initial, "unmatched `---` in DDoc comment");
+        error(loc, "unmatched `---` in DDoc comment");
 }
 
 /**************************************************
--- a/src/dmd/doc.d
+++ b/src/dmd/doc.d
@@ -136,7 +136,7 @@ extern (C++) class Section
                 char c = name[u];
                 buf.writeByte((c == '_') ? ' ' : c);
             }
-            escapeStrayParenthesis(loc, buf, o);
+            escapeStrayParenthesis(loc, buf, o, false);
             buf.writestring(")");
         }
         else
@@ -551,7 +551,7 @@ extern (C++) void escapeDdocString(OutBuffer* buf, size_t start)
  *
  * Fix by replacing unmatched ( with $(LPAREN) and unmatched ) with $(RPAREN).
  */
-extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start)
+extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start, bool respectMarkdownEscapes = true)
 {
     uint par_open = 0;
     bool inCode = 0;
@@ -599,6 +599,20 @@ extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start)
             }
             if (numdash >= 3)
                 inCode = !inCode;
+            break;
+        case '\\':
+            if (!inCode && respectMarkdownEscapes && u+1 < buf.offset)
+            {
+                if (buf.data[u+1] == '(' || buf.data[u+1] == ')')
+                {
+                    string paren = buf.data[u+1] == '(' ? "$(LPAREN)" : "$(RPAREN)";
+                    buf.remove(u, 2); //remove the \)
+                    buf.insert(u, paren); //insert this instead
+                    u += 8; //skip over newly inserted macro
+                }
+                else if (buf.data[u+1] == '\\')
+                    ++u;
+            }
             break;
         default:
             break;
@@ -1823,6 +1837,25 @@ extern (C++) bool isDitto(const(char)* comment)
     return false;
 }
 
+/*****************************************
+ * Return true if the given character is punctuation.
+ */
+private bool isPunctuation(const(char) c)
+{
+    static immutable bool[char.max+1] asciiPunctuation =
+       ['!': true, '"': true, '#': true,
+        '$': true, '%': true, '&': true, '\'': true, '(': true, ')': true,
+        '*': true, '+': true, ',': true, '-': true, '.': true, '/': true,
+        ':': true, ';': true, '<': true, '=': true, '>': true, '?': true,
+        '@': true, '[': true, '\\': true, ']': true, '^': true, '_': true,
+        '`': true, '{': true, '|': true, '}': true, '~': true];
+    return asciiPunctuation[c];
+// TODO: unicode punctuation: Pc, Pd, Pe, Pf, Pi, Po, or Ps.
+// However, unicode punctuation should not be included for Markdown backslash
+// escapes, so when it's implemented be sure to make the unicode punctuation
+// check optional via a parameter
+}
+
 /**********************************************
  * Skip white space.
  */
@@ -2411,6 +2444,31 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
                 }
             }
             break;
+
+        case '\\':
+        {
+            /* Escape Markdown special characters */
+            if (inCode || i+1 >= buf.offset)
+                break;
+            char c1 = buf.data[i+1];
+            if (isPunctuation(c1))
+            {
+                buf.remove(i, 1);
+
+                auto se = sc._module.escapetable.escapeChar(c1);
+                if (!se)
+                    se = c1 == '$' ? "$(DOLLAR)".ptr : c1 == ',' ? "$(COMMA)".ptr : null;
+                if (se)
+                {
+                    const len = strlen(se);
+                    buf.remove(i, 1);
+                    i = buf.insert(i, se, len);
+                    i--; // point to ';'
+                }
+            }
+            leadingBlank = false;
+            break;
+        }
 
         case '$':
         {
--- a/test/compilable/extra-files/ddoc_markdown_escapes.html
+++ b/test/compilable/extra-files/ddoc_markdown_escapes.html
@@ -1,0 +1,510 @@
+
+<!DOCTYPE html>
+<html>
+  <head>
+    <meta charset="UTF-8">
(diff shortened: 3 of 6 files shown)


src/dmd/doc.d
```diff
@@ -551,7 +551,7 @@ extern (C++) void escapeDdocString(OutBuffer* buf, size_t start)
  *
  * Fix by replacing unmatched ( with $(LPAREN) and unmatched ) with $(RPAREN).
  */
-extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start)
+extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start, bool respectMarkdownEscapes = true)
```
jacob-carlborg: I don't think Walter like default values.

src/dmd/doc.d
```diff
@@ -1823,6 +1837,25 @@ extern (C++) bool isDitto(const(char)* comment)
...
+private bool isPunctuation(const(char) c)
+{
+    static immutable bool[char.max+1] asciiPunctuation =
+       ['!': true, '"': true, '#': true,
+        '$': true, '%': true, '&': true, '\'': true, '(': true, ')': true,
+        '*': true, '+': true, ',': true, '-': true, '.': true, '/': true,
+        ':': true, ';': true, '<': true, '=': true, '>': true, '?': true,
+        '@': true, '[': true, '\\': true, ']': true, '^': true, '_': true,
+        '`': true, '{': true, '|': true, '}': true, '~': true];
+    return asciiPunctuation[c];
```
jacob-carlborg: If the given character is not in the associative array an error will be thrown.

src/dmd/doc.d
```diff
@@ -1823,6 +1837,25 @@ extern (C++) bool isDitto(const(char)* comment)
     return false;
 }
 
+/*****************************************
+ * Return true if the given character is punctuation.
+ */
```
jacob-carlborg: The comment should have proper `Returns:` and `Params:` sections.

src/dmd/doc.d
```diff
@@ -2412,6 +2445,31 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
...
+                if (se)
+                {
+                    const len = strlen(se);
+                    buf.remove(i, 1);
+                    i = buf.insert(i, se, len);
+                    i--; // point to ';'
+                }
+            }
+            leadingBlank = false;
+            break;
```
jacob-carlborg: Can this be moved to a separate function?

jacob-carlborg (changes requested):
* There are no tests
* There is no changelog entry
* There should be a PR for dlang.org that updates the spec. Link that PR from this one

jacob-carlborg:
A generic comment about the Markdown support. Can that be added separately as a form of prerpocessor? That would avoid mixing code for handling the Markdown and the regular Ddoc in the same functions.

src/dmd/doc.d
dgileadi: Except this is not an associative array, it's a regular array of length 256 as suggested by @BBasile. So no danger.

dgileadi:
I thought about putting Markdown processing into its own function. I decided not to for two reasons:

1. It's more efficient to process everything in one loop than in two. However this isn't a very strong reason.
2. The `highlightText` function was created to process Markdowny things—its pre-Markdown job is to handle paragraphs, to handle HTML tags/comments/entities, and to handle code fences and backtick-escaped code [1]. These are Markdown features, albeit with slightly different syntax and semantics [2]. So to me it doesn't make sense to put additional Markdown features anywhere else; it's what the function is designed to do.

[1] Plus the iffy addition of auto-highlighting D symbols.
[2] Indeed the most difficulty I've had in conforming to the CommonMark spec is due to mismatches in existing features and the spec's implementation.

src/dmd/doc.d
WalterBright: In this case, no I don't :-)

src/dmd/doc.d
WalterBright: Is this reinventing core.stdc.ctype.ispunct?

http://www.cplusplus.com/reference/cctype/

dgileadi: After testing it, yes. Yes it is.

src/dmd/doc.d
dgileadi: There are several sections like this in `highlightText` that could be moved to separate functions. I generally went with the principle that if it was called in multiple places I made it a separate function and otherwise left it alone. However I'd be happy to pull it out if that's preferred.

test/compilable/extra-files/ddoc_markdown_escapes.html
```diff
@@ -0,0 +1,504 @@
+
+<!DOCTYPE html>
```
wilzbach: Maybe we should have something like dlang.org's verbatim.ddoc file here too?

https://github.com/dlang/dlang.org/blob/master/verbatim.ddoc

Might making updating and debugging easier.

res/default_ddoc_theme.ddoc
```diff
@@ -2,6 +2,8 @@ LPAREN = (
 RPAREN = )
 BACKTICK = `
 DOLLAR = $
+COMMA = ,
+QUOTE = &quot;
```
wilzbach: Requires an update to:
- https://github.com/dlang/dlang.org/blob/master/verbatim.ddoc
- https://github.com/dlang/dlang.org/blob/master/spec/latex.ddoc

CyberShadow:
So what's the compatibility story? Is the idea that users will just have to deal with any changes that break existing documentation?

Shouldn't this have a DIP or something?

jmdavis:
My main concern with all of this is that right now, I can write ddoc without worrying about stray parts of text being screwed with, because they're treated as special in some way. The only real concern is that you have to worry about text that's not going to play well in the target format, and then you have to use a macro in those cases so that it can be converted to the appropriately escaped piece of text for the target format. With markdown stuff being added, am I suddenly going to have perfectly reasonable text being mucked with, because markdown thinks that something in there should be treated as special? The reason that I hate markdown is precisely because stray pieces of text get treated as magic, and I have to constantly fight it so that what I'm trying to display doesn't get screwed up.

dgileadi:
I'm not the right person to discuss process (DIP etc.). What I can speak to is what I've noticed with Phobos/dlang.org breakage due to adding Markdown support.

Some of the breakage comes from supporting backslash escapes, which is this PR and the PRs it links to. With "full" Markdown support I've noticed additional breakage from lines that start with `-` which are treated as list items, lines that start with `>` which are treated as block quotes, and paired `*` characters embedded in text which are treated as emphasis.

In Phobos and dlang.org there are about 30 files that change. Of these, maybe a quarter would be considered desired changes, like changing words that uses `*` emphasis to actual emphasis, or changing numbered or bulleted list text to actual lists. There's even [a Markdown-style link in the latest 2.079.0 changelog](https://dlang.org/changelog/2.079.0.html#dip1008) that gets converted to an actual link.

CyberShadow:
Thanks for the analysis. I think that's sufficient grounds that we may want to think of a design before rushing in with an implementation.

This should probably be discussed more formally as a DIP or elsewhere, but perhaps making Markdown syntax opt-in would be a better path, e.g. it could be enabled if the comment body begins with `!` (i.e. `/**! ...markdown follows ... */`).

wilzbach:
> perhaps making Markdown syntax opt-in would be a better path

How about `-transition=markdown` for now?
This should unblock this PR. After all, both Walter and Andrei have pre-approved this and small details can be figured out while the `-transition=markdown` flag gets more complete and it allows to work gradually on improving it.

dgileadi:
I just added support for `-transition=markdown`. I added that flag to the makefiles in dlang.org, phobos and druntime to make sure we eat our own dogfood.

jmdavis:
Honestly, I think that adding markdown support to ddoc is a huge mistake. It's going to make it a royal pain to just write documentation (or web pages or anything else that one might want to generate with ddoc) without having problems due to pieces of text suddenly being treated as special by the markdown processing and getting screwed up in the process. Having a transition flag doesn't help with that in any way shape or form. Anyone who's been doing anything serious with ddoc is going to get screwed by having stray pieces of text treated as markdown.

IMHO, if we're going to add support for any kind of markdown stuff, it needs to be in a controlled manner and not just treating normal ddoc as if it potentially contained markdown.

CyberShadow:
> I just added support for -transition=markdown

Sorry, I don't think that's the expected behavior for a transition flag. A more useful meaning for that flag would be for the compiler to list all places that will be interpreted differently with Markdown parsing enabled. As the plan seems to be to introduce Markdown syntax support gradually, there would need to be one transition flag per change.

But before you spend any more time on the implementation, it's probably better to figure out how we want to go about doing this whole thing.

@WalterBright / @andralex ?

res/default_ddoc_theme.ddoc
andralex: This definition seems out of place. All others are universal - plain paren, plain backtick, plain dollar. This is HTML-specific so its hould go in an HTML-specific file. Granted, all others work out in HTML...

andralex (commented):
I'm not particularly concerned about compatibility with our previous document formatting. A careful transition would be nice, but one flag per change is definitely too much.

I'm okay with one flag enabling the new interpretation.

jmdavis:
> I'm not particularly concerned about compatibility with our previous document formatting. A careful transition would be nice, but one flag per change is definitely too much.

Remember that there are no unit tests for anything generated with ddoc, and there generally aren't compilation errors when generating anything from ddoc. If we're going to add markdown support we need to be _very_ clear about what's being changed and remember that any changes made are likely to result in existing ddoc being screwed up and/or folks having to read through all of their ddoc (be it for documentation, a website, or whatever) and manually find and fix every problem caused by the compiler treating portions of text as markdown whereas before it was just text. This is in stark contrast to a language or library change where you get a deprecation message or compiler error message for everywhere that you have to change.

CyberShadow:
> A careful transition would be nice, but one flag per change is definitely too much.

One thing about doing this as a series of steps is that we commit ourselves to something we can't predict the outcome of. This change or the next few might have negligible side effects, but at some point we will possibly want to support some Markdown feature that is going to break enough DDoc comments that it will matter. And, at that point, it possibly will already be too late to reconsider options like making Markdown syntax opt-in (via syntax or a switch) or have some other more rigorous transition plan.

dgileadi:
Regarding enabling markdown step-by-step rather than all at once, my initial PR did enable all Markdown features at once. Since it added a fair bit of code, @WalterBright and @jacob-carlborg requested that I create individual PRs for each feature to make them easier to manage. If you'd like to see all the changes at once you can check out [the defunct PR](https://github.com/dlang/dmd/pull/7764) and/or try the branch out on your own docs. I'll try to keep that branch up to date with requested changes in each PR like this one.

Since it's all already implemented, I was able to compare the differences in the generated dlang.org, druntime and phobos docs pre- and post-markdown and see what changed. In a previous comment I mentioned:

> In Phobos and dlang.org there are about 30 files that change. Of these, maybe a quarter would be considered desired changes, like changing words that uses * emphasis to actual emphasis, or changing numbered or bulleted list text to actual lists. There's even a Markdown-style link in the latest 2.079.0 changelog that gets converted to an actual link.

That's around 8% of files that it affected, and I'd expect similar numbers for other repos. Hopefully this can help with planning how best to introduce Markdown support.

res/default_ddoc_theme.ddoc
dgileadi: I'll remove the `QUOTE` macro until it's needed, which is for URIs. It is indeed specific to HTML, but will have to be one of the built-in macros because it'll be generated by the ddoc processor.

andralex:
@CyberShadow up to you

wilzbach:
> @CyberShadow up to you

Ping.

> I'm thinking that perhaps dmd would log messages by default, and that -transition=markdown would suppress those messages. How does that sound?

I like the idea (though we should probably enable the messages only iff dmd supports markdown "fully"?)

CyberShadow:
Well, I'm not sure about making any unilateral decisions here, as I don't really have a dog in this race, and the compiler isn't really a part of D I'm too involved with. Though... @dgileadi How would you feel about writing a DIP so that the exact syntax can be formalized, and to provide for further grounds for discussing the compatibility aspects?

dgileadi:
That brings up the content of the messages. I see three options:

1. Log a message every time a markdown feature takes effect, like `path/to/file.d(32): Ddoc: backslash-escaped $ (suppress this message with -transition=markdown)`.
2. Log a message for each markdown block that had markdown features applied to it, like `path/to/file.d(32): Ddoc: applied the following markdown features: backslash-escaped punctuation, emphasized text (suppress this message with -transition=markdown)`.
3. Log a message for each file that had markdown features applied to it, like `path/to/file.d: Ddoc: applied some markdown features (suppress this message with -transition=markdown)`.

Note that for both the first and second options, the location `(32)` is the location of the symbol that the Ddoc is documenting, *not* the line that the markdown text was on. If option 1 is preferred and the markdown-text line is desired, I can see if I can make it work.

dgileadi:
@CyberShadow I've never written a DIP and am a little intimidated by it. In any case I guess the real question is, will going through the DIP process improve this feature? I don't know if the answer is clearly "yes".

andralex:
@dgileadi thanks for this work. I can tell this DIP is unlikely to encounter any serious opposition so no need to consider it a difficult process. It would have a number of benefits:

* We can introduce the semantics with the flag -dipXXXX, which is nice and uniform. Later releases will do it automatically
* The DIP will serve as clear, crisp documentation of the changes and how they affect existing code

So I'd say this is not much more work than actually documenting the feature, and in a more official format.

wilzbach:
Guys. You have approved this already half a year ago when @dgileadi asked on the mailing list. It would take another half a year for the DIP to reach the phase where it will be decided, just because of:

> We can introduce the semantics with the flag -dipXXXX, which is nice and uniform. 

Which is _not_ so nice because at the moment because there's no way to group all these dips and we have to add each and everyone to the precious command line interface. Remember that you recently blocked a PR that adds Makefile-compatible  dependency output to DMD and another one that would allow cross-compilation with DMD to another runtime, because you didn't want to commit to a new CLI switch.
Also note that we humans can't possibly remember what all the cryptic numbers mean, so a lot of folks already have to lookup what -dip1008 does and you want to add more instead of -transition=markdown?

I get that you don't want to commit to a decision that could backfire and are this pushing this off into the DIP process.
However, please don't make me list how many breaking changes are merged here on a regular basis. I know that is not a good argument, but I still don't like the "some sheep are more equal" strategy. Also note that e.g. dip1008 has been merged long before it got "accepted", which imho is fine because it takes time for a feature to mature, but we should allow others to have similar capabilities. It will be easy for you to "approve" dip1008, because it's already implemented and might soon even be used by Phobos. Can't we give this a similar possibility? After all, most issues are going to be worked out when people test this, i.e. they can use it easily as part of dmd.
Also the general transition to allowing Markdown has already approved been by Walter himself.

Now it definitely would be interesting to have a DIP that explores what subset of Markdown can be kept etc., but note that a few months ago everyone was fine with "step by step pull requests". I understand if for formal purposes you decide that a DIP would be better for the grammar changes, but please please don't block @dgileadi's great work by requiring him to write a DIP, and wait another six months before doing any more real work on this.

Hence, I am strongly voting for merging this behind the -transition=markdown flag and while this feature gets developed further and we have real-world feedback, a DIP can materialize and define the exact grammar changes.
It's not realistic to review a > 5k LoC PR nor even maintain it. Let's learn from the mistakes of the past please.

andralex:
>Hence, I am strongly voting for merging this behind the -transition=markdown flag and while this feature gets developed further and we have real-world feedback, a DIP can materialize and define the exact grammar changes.

I'm on board with this or anything that makes things faster. Just trying to simplify everyone's life and delegate to @CyberShadow. FWIW I don't consider documentation generation breakages as unpleasant as code breakages.

jmdavis:
> FWIW I don't consider documentation generation breakages as unpleasant as code breakages.

In some respects, documentation generation breakage isn't as bad, since it doesn't prevent any programs from running, but it's also often hard to detect and easy to miss, whereas code breakage often involves compilation errors, which makes it very easy to know that you have a problem that needs fixing. Also, remember that ddoc is used for more than just documentation. So, any changes (be they positive or negative) will affect more than what the documentation for a particular library looks like.

CyberShadow:
OK, how about this:

- We skip the DIP for this one
- We create a new branch in this repository, `markdown` (just so there's no risk of this project spanning a release branch-off - then there is no risk of commitment)
- Markdown changes are introduced as a series of small pull requests targeting the `markdown` feature branch, so that they can be reviewed & merged
- Each PR should update `-transition=markdown` correspondingly
- Once all changes are in, there is one final review/discussion and the `markdown` branch is merged in `master`.
- The introduction of Markdown support in `master` is accompanied by a changelog entry that instructs people who worry about their documentation looking OK to use `-transition=markdown` to help them update it.

CyberShadow:
Two more things about the implementation:

- It would be great if it was possible to update the documentation in such a way that it results in identical output before and after these changes. This could be done by wrapping special characters into macros, which should work the same with and without Markdown support.

- It's very important that merely reflowing text does not affect its structure. I.e. if a dash in the middle of a line is reflowed to the beginning of the line, it shouldn't be reinterpreted as a list item (lists ought to be recognized only if they're on a "paragraph" - block of text delimited by empty lines - on their own).

dgileadi:
I'm definitely on board with the `markdown` branch approach. I was worried about crossing release boundaries too, given that at least this PR has taken a while and it's the smallest of the Markdown-related changes.

As for your two additional items:

> - It would be great if it was possible to update the documentation in such a way that it results in identical output before and after these changes. This could be done by wrapping special characters into macros, which should work the same with and without Markdown support.

The "Markdown" way is, of course, backslash escapes, which is this PR. That's one of the reasons I made this PR the first one, since it allows undoing Markdown features generated by subsequent PRs. If I understand correctly, are you suggesting something like `$(HYPHEN)` for `-`, which people could use to avoid accidentally starting a bullet list? This is certainly doable (entirely within the existing Ddoc feature set) but in my opinion `\-` is simpler.

Please let me know if I misunderstood your proposal.

> - It's very important that merely reflowing text does not affect its structure. I.e. if a dash in the middle of a line is reflowed to the beginning of the line, it shouldn't be reinterpreted as a list item (lists ought to be recognized only if they're on a "paragraph" - block of text delimited by empty lines - on their own).

I followed the CommonMark spec for this matter. The spec has [an in-depth discussion about it](http://spec.commonmark.org/0.28/#example-266), including why they chose to support starting a list without requiring a blank line ahead of it.

I'm open to debating it, but I lean toward supporting the CommonMark spec as much as is feasible since it's emerging as the most widely-supported standard (and therefore what people are likeliest to expect).

Note that source code text doesn't "reflow" by accident—whenever a dash is at the start of a line it's because someone hit the enter key to put it there. There are likely tools that automatically reflow documentation comments in source code, but I'd expect such tools to typically not start a line with punctuation like a dash. However I'm interested to hear about such tools and how they behave in the real world.

There are a couple of places in the Phobos docs that currently do start a line with a dash and so are unexpectedly treated as lists. These join the other unintended markdown features in the standard docs that I'll create a fixing PR for when the `markdown` branch is ready to merge into `master`.

dgileadi:
I'd really like the desired meaning of `-transition=markdown` to get nailed down in the near future. Things about it that I think need solving are:

1. Does providing `-transition=markdown` on the CLI turn on markdown-related messages or does it turn them off (i.e. are markdown-related messages on by default)?
2. How granular are the markdown-related messages? For further discussion about this see [my previous comment](#issuecomment-377960957).

dgileadi:
I'm up for doing the Phobos/dlang.org transition the way you describe, assuming we go with the separate `markdown` branch you proposed earlier. Speaking of which, can someone with the appropriate rights create that branch for me?

I'll make `-transition=markdown` turn on messages about each Markdown replacement.

After some effort I've discovered that it's a more difficult task to output the correct line and column number in the messages than I expected. This is for a couple reasons:

1. DMD provides `highlightText` with the location of the symbol being documented, not the location of the start of the preceding documentation. I may be able to work backwards, but I'm not convinced it'll always be accurate.
2. `highlightText` processes Markdown (like existing Ddoc features) as it encounters them, rather than the slower approach of using multiple passes to tokenize the doc text and then do replacements on it afterwards. This means that by the time the processor arrives at a line/column, previous text replacements cause it to differ from the original source location.

I could do some sort of pre-processing step to discover locations, but this would effectively mean parsing the documentation twice and it would be tricky to accurately sync the state between passes. I'll give it some more effort, but so far it appears likely that I'll have to give up on column numbers, and it may well be that if I'm trying to be accurate, all I can print is the line number of the symbol being documented instead of the line number that the Markdown feature appears on.

CyberShadow:
> Speaking of which, can someone with the appropriate rights create that branch for me?

I created the branch and changed the PR's target branch to it.

> I could do some sort of pre-processing step to discover locations, but this would effectively mean parsing the documentation twice and it would be tricky to accurately sync the state between passes. I'll give it some more effort, but so far it appears likely that I'll have to give up on column numbers, and it may well be that if I'm trying to be accurate, all I can print is the line number of the symbol being documented instead of the line number that the Markdown feature appears on.

Ah, too bad. The symbol's line number will have to do, I guess. Having the exact line:column number would have allowed easily going through the output and fixing it very quickly, if the editor could place the cursor right on the symbol in question.

dgileadi:
Thanks for making the branch!

I'll work on getting line numbers correct.

dgileadi:
After some looking into getting the line numbers correct, I've realized that I can't just count backwards from where the symbol is defined, because inside doc.d it isn't possible to distinguish between the following situations:

```d
/// doc
enum symbol;
```

```d
/// doc

enum symbol;
```

```d
enum symbol; /// doc
```

One alternative is just to print the line number of the symbol that the documentation applies to, not the line number where the Markdown text is transformed. However this isn't friendly in situations where the documentation comments are long.

One possible solution is to add a `Loc commentStart` member to the `Token` and `DSymbol` structs. This would allow the comment location to be preserved. However it would break when comments are concatenated, such as:

```d
/// doc

/// more doc
enum symbol;
```

Also adding a member to `Token` and `DSymbol` is a bit of a fundamental change. Any suggestions?

CyberShadow:
> Any suggestions?

I suggest to not worry about it, and go with the simplest solution for now (print the location of the documented symbol).

I think ideally, strings and DDoc comments would be represented in the compiler as ropes, wherein each rope segment stores the location in the source code from whence it came from. Said data would survive the usual rope operations like slicing and concatenation, so it's possible to know which file, line and column each character in the rope came from. This would also allow providing accurate error location for string mixins, as the quality of their error messages is a common complaint with them. But this is a topic to discuss for another day.

Another idea is to keep the location of everything not as a point, but as a range (start and end), which sounds like it could be useful here. It would also be useful for editors/IDEs which support drawing red squiggly lines to highlight bad code, as well as command-line output if we ever decide to switch to a format more like clang and Rust.

dgileadi:
Okay, I went with the simple solution of just printing the line of the documented symbol. However if the source file is a .dd file then it is able to print the actual line number of the replacement in the source file, so at least there's that.

CyberShadow:
@dgileadi Have a look at the CI failures, they seem pertinent to the changes here.

@jacob-carlborg Have your objections been adequately addressed?

jacob-carlborg:
> Have your objections been adequately addressed?

No. Still now changelog entry and no spec update.

src/dmd/doc.d
```diff
@@ -600,6 +602,20 @@ extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start)
             if (numdash >= 3)
                 inCode = !inCode;
             break;
+        case '\\':
+            if (!inCode && respectMarkdownEscapes && u+1 < buf.offset)
```
jacob-carlborg: Spaces around operators, `u+1` -> `u + 1`.

src/dmd/doc.d
```diff
@@ -2412,6 +2434,35 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, OutBuffer* buf, size_t o
...
+            if (ispunct(c1))
+            {
+                if (global.params.vmarkdown)
+                    message(loc, "Ddoc: backslash-escaped %c", c1);
+
+                buf.remove(i, 1);
+
+                auto se = sc._module.escapetable.escapeChar(c1);
+                if (!se)
+                    se = c1 == '$' ? "$(DOLLAR)".ptr : c1 == ',' ? "$(COMMA)".ptr : null;
```
jacob-carlborg: `.ptr` is not necessary.

jacob-carlborg (changes requested):

jacob-carlborg:
> We never added a changelog entry for something that isn't finished before and neither did we update the spec.

It didn't noticed it was behind a transition flag.

> we didn't add the creation of newCTFE to the changelog, did we?

newCTFE isn't merged into master, is it?

jacob-carlborg (approved):

jacob-carlborg:
@dgileadi what's the status of this? Two CI systems fail, seems it could be related.

dgileadi:
Yes, I suspect it's related. I haven't taken the time to look into it yet—all my evenings have recently been taken by installing a sprinkler system in our yard. It's time/season sensitive. I'll see if I can sneak some time to take a look at the CI failure. Otherwise it may be another week or two.

dgileadi:
Yay, got the build fixed!

jacob-carlborg:
I've already given my approvement. I would like someone else to approve this before merging.

andralex:
OK I assume everybody's on board so I'll pull the trigger. Thanks!

PR #8045, opened 2018-03-17
fix Issue 18282 part 2

WalterBright:


WalterBright:
@CyberShadow looks like the jenkins failure is in your ae project, which looks like a bug in it.

CyberShadow:
> which looks like a bug in it.

I don't see how. Could you please elaborate?

WalterBright:
Here's the Jenkins error report:
```
sys/persistence/core.d(106,9): Error: function type `return @system void()` has `return` but does not return any indirections

sys/install/common.d(242,13): Error: template instance `ae.sys.persistence.memoize.PersistentMemoized!(resolveRedirectImpl, cast(FlushPolicy)3)` error instantiating
```
and the project is `Test Projects / CyberShadow/ae`

CyberShadow:
Yes, I saw that.

Those error messages make no sense for that situation.

I also don't see any bugs in the code (that the compiler would be able to detect, anyway).

As far as I can see, the code should compile, and the problem is with this pull request.

WalterBright:
Could you please produce a test case that fails?

CyberShadow:
> Could you please produce a test case that fails?

You mean, a *reduced* test case, right? Because the project itself is *a* test case :)

I've started a DustMite reduction, but it will take a bit to finish.

CyberShadow:
Reduced test case:

```d
struct S()
{
	void fun()
	{
		arr ~= &this;
	}

	typeof(this)*[] arr;
}

alias T = S!();
```

WalterBright:
Thanks!

WalterBright:
@CyberShadow while the code snippet is not a bug in current D, with dip1000 it will be. The trouble is that `this` is a reference, and it will not be allowed to escape the reference. An alternative is to write the function as:
```
void fun(S* s) { arr ~= s; }
```

CyberShadow:
Yep, I noticed that, and it does make sense. Thanks for the tip.

BTW, not sure if you noticed, but the error message is still misleading. With this PR, `-dip1000`, and the reduced example above, I get:

```
test.d(5): Error: returning &this escapes a reference to parameter this, perhaps annotate with return
```

but if I do what the error message suggests and add a `return` annotation, I get:

```
test.d(3): Error: function type return void() has return but does not return any indirections
```

WalterBright:
> error message

It's been filed already:

https://issues.dlang.org/show_bug.cgi?id=17784

JinShil (approved):

marler8997:
Ah, thanks.  I should have searched the PR list for this commit, I was just looking at the commit list and didn't see a merge commit.

Anyway, looks like this PR did have the intermittent failure.  I don't think the previous one (or any other previous ones) did.

marler8997:
Way to go DustMite...brilliant tool

PR #8046, opened 2018-03-17
refactor: add foreachBodyToFunction()

WalterBright:
Pulling it out into a separate function makes it easier to understand. Reducing the scope of various variables is a plus, too.


JinShil (approved):

src/dmd/statementsem.d
```diff
@@ -1912,6 +1859,79 @@ else
         result = s;
     }
 
+    /*************************************
+     * Turn foreach body into the function literal:
+     *  int delegate(ref T param) { body }
```
jacob-carlborg: Can be highlighted as code.

PR #8047, opened 2018-03-17
fix Issue 15624 - opApply with @safe and @system variants can't be us…

WalterBright:
…ed with foreach syntax

### DIFF
--- a/src/dmd/opover.d
+++ b/src/dmd/opover.d
@@ -2014,12 +2014,20 @@ private FuncDeclaration findBestOpApplyMatch(Expression ethis, FuncDeclaration f
             fd_ambig = null;
             match = m;
         }
-        else if (m == match)
-            fd_ambig = f;
+        else if (m == match && m > MATCH.nomatch)
+        {
+            assert(fd_best);
+            /* Ignore covariant matches, as later on it can be redone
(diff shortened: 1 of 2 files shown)


jacob-carlborg (approved):

JinShil (approved):

marler8997:
This PR may have exposed a bug in the 2.068 compiler on the autotester.  Most PRs are failing in the "test dmd" stage after this commit was merged (this commit even failed but succeeded on another run, definitely intermittent).

This link shows where OSx 32 bit started failing:
https://auto-tester.puremagic.com/platform-history.ghtml?projectid=1&os=Darwin_64_32

It's running into a segfault bug that only happens on OSx 32-bit.  This bug has been encountered many times in the past.  I offered to create a patch for 2.068 to fix it but was told the autotester would be updating their compiler version so it was not needed.  Howver that was back in January and the compiler version has not been updated because @braddr has not been told by dlang leadership to update.

marler8997:
Wait, looks like this failure happened on older commits than this one.

Looks like it might be this one: https://github.com/dlang/dmd/commit/2dbc2bd85d5e5a55a3bcbf362c60cb0648102a73

I don't see a PR associated with this one so I'm not sure if the autotester has tested it.

PR #8049, opened 2018-03-18
Fix Issue 18620 - error cannot be interpreted at compile time is missing context where error occurs

RazvanN7:
…ing context where error occurs.

The fix currently produces redundant output for some cases, but that is because I have not found a simple, direct way to pas information from the ctfe interpreting visitor to the starting declaration. Suggestions welcome. cc @UplinkCoder 

Ideally, I would send a token from the interpret visitor all the way to the the point in the code after interpret has been run on the Initializer, but that cannot be done by simply sending ErrorExp

### DIFF
--- a/src/dmd/ctfeexpr.d
+++ b/src/dmd/ctfeexpr.d
@@ -200,14 +200,15 @@ extern (C++) final class ThrownExceptionExp : Expression
     }
 }
 
+
 /***********************************************************
  * This type is only used by the interpreter.
  */
-extern (C++) final class CTFEExp : Expression
+extern (C++) class CTFEExp : Expression
 {
-    extern (D) this(TOK tok)
+    extern (D) this(TOK tok, int size = __traits(classInstanceSize, CTFEExp))
     {
-        super(Loc.initial, tok, __traits(classInstanceSize, CTFEExp));
+        super(Loc.initial, tok, size);
         type = Type.tvoid;
     }
 
@@ -244,6 +245,30 @@ extern (C++) final class CTFEExp : Expression
     static bool isGotoExp(const Expression e)
     {
         return e && e.op == TOK.goto_;
+    }
+}
+
+enum CTE : int
+{
+    missingFuncBody,
+}
+
+/* This class is used to pass CTFE errors to
+ * higher levels in the AST so that error
+ * messages can be written on the call site
+ */
+extern (C++) final class CTErrorExp : CTFEExp
+{
+    CTE err;
+    extern (D) this(CTE err)
+    {
+        super(TOK.cantExpression, __traits(classInstanceSize, CTErrorExp));
+        this.err = err;
+    }
+
+    override inout(CTErrorExp) isCTErrorExp() inout
+    {
+        return this;
     }
 }
 
--- a/src/dmd/ctfeexpr.d
+++ b/src/dmd/ctfeexpr.d
@@ -200,15 +200,14 @@ extern (C++) final class ThrownExceptionExp : Expression
     }
 }
 
-
 /***********************************************************
  * This type is only used by the interpreter.
  */
-extern (C++) class CTFEExp : Expression
+extern (C++) final class CTFEExp : Expression
 {
-    extern (D) this(TOK tok, int size = __traits(classInstanceSize, CTFEExp))
+    extern (D) this(TOK tok)
     {
-        super(Loc.initial, tok, size);
+        super(Loc.initial, tok, __traits(classInstanceSize, CTFEExp));
         type = Type.tvoid;
     }
 
@@ -236,6 +235,10 @@ extern (C++) class CTFEExp : Expression
     extern (C++) static __gshared CTFEExp breakexp;
     extern (C++) static __gshared CTFEExp continueexp;
     extern (C++) static __gshared CTFEExp gotoexp;
+    /* Used when additional information is needed regarding
+     * a ctfe error.
+     */
+    extern (C++) static __gshared CTFEExp showcontext;
 
     static bool isCantExp(const Expression e)
     {
@@ -248,34 +251,10 @@ extern (C++) class CTFEExp : Expression
     }
 }
 
-enum CTE : int
-{
-    missingFuncBody,
-}
-
-/* This class is used to pass CTFE errors to
- * higher levels in the AST so that error
- * messages can be written on the call site
- */
-extern (C++) final class CTErrorExp : CTFEExp
-{
-    CTE err;
-    extern (D) this(CTE err)
-    {
-        super(TOK.cantExpression, __traits(classInstanceSize, CTErrorExp));
-        this.err = err;
-    }
-
-    override inout(CTErrorExp) isCTErrorExp() inout
-    {
-        return this;
-    }
-}
-
 // True if 'e' is CTFEExp::cantexp, or an exception
 extern (C++) bool exceptionOrCantInterpret(const Expression e)
 {
-    return e && (e.op == TOK.cantExpression || e.op == TOK.thrownException);
+    return e && (e.op == TOK.cantExpression || e.op == TOK.thrownException || e.op == TOK.showCtfeContext);
 }
 
 /************** Aggregate literals (AA/string/array/struct) ******************/
--- a/src/dmd/dinterpret.d
+++ b/src/dmd/dinterpret.d
@@ -4997,7 +4997,7 @@ public:
         if (!fd.fbody)
         {
             e.error("`%s` cannot be interpreted at compile time, because it has no available source code", fd.toChars());
-            result = new CTErrorExp(CTE.missingFuncBody);
+            result = CTFEExp.showcontext;
             return;
         }
 
--- a/src/dmd/dinterpret.d
+++ b/src/dmd/dinterpret.d
@@ -4997,7 +4997,7 @@ public:
         if (!fd.fbody)
         {
             e.error("`%s` cannot be interpreted at compile time, because it has no available source code", fd.toChars());
-            result = CTFEExp.cantexp;
+            result = new CTErrorExp(CTE.missingFuncBody);
             return;
         }
 
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -22,6 +22,7 @@ import dmd.astcodegen;
 import dmd.attrib;
 import dmd.blockexit;
 import dmd.clone;
+import dmd.ctfeexpr;
 import dmd.dcast;
 import dmd.dclass;
 import dmd.declaration;
@@ -560,10 +561,10 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
     {
         version (none)
         {
-            printf("VarDeclaration::semantic('%s', parent = '%s') sem = %d\n", toChars(), sc.parent ? sc.parent.toChars() : null, sem);
-            printf(" type = %s\n", type ? type.toChars() : "null");
+            printf("VarDeclaration::semantic('%s', parent = '%s')\n", dsym.toChars(), sc.parent ? sc.parent.toChars() : null);
+            printf(" type = %s\n", dsym.type ? dsym.type.toChars() : "null");
             printf(" stc = x%x\n", sc.stc);
-            printf(" storage_class = x%llx\n", storage_class);
+            printf(" storage_class = x%llx\n", dsym.storage_class);
             printf("linkage = %d\n", sc.linkage);
             //if (strcmp(toChars(), "mul") == 0) assert(0);
         }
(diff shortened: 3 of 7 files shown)


src/dmd/dsymbolsem.d
```diff
@@ -502,10 +503,10 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
     {
         version (none)
         {
-            printf("VarDeclaration::semantic('%s', parent = '%s') sem = %d\n", toChars(), sc.parent ? sc.parent.toChars() : null, sem);
-            printf(" type = %s\n", type ? type.toChars() : "null");
+            printf("VarDeclaration::semantic('%s', parent = '%s')\n", dsym.toChars(), sc.parent ? sc.parent.toChars() : null);
+            printf(" type = %s\n", dsym.type ? dsym.type.toChars() : "null");
```
UplinkCoder: printf will ignore nulls, no need for ternary here.

src/dmd/dsymbolsem.d
```diff
@@ -1157,9 +1158,20 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                    toCBuffer(dsym, &buff, &hgs);
+                    buff.setsize(buff.offset-1);
+
                     // https://issues.dlang.org/show_bug.cgi?id=14166
                     // Don't run CTFE for the temporary variables inside typeof
                     dsym._init = dsym._init.initializerSemantic(sc, dsym.type, sc.intypeof == 1 ? INITnointerpret : INITinterpret);
+                    const init_err = dsym._init.isExpInitializer();
+                    if (init_err && init_err.exp.op == TOK.error)
+                    {
+                        errorSupplemental(dsym.loc, "statical context created here: `%s`", buff.extractString());
```
UplinkCoder: statical => compile time

UplinkCoder:
Well you could derive a CTErrorExp from ErrorExp and have that store more context.

src/dmd/dsymbolsem.d
wilzbach: Maybe instead of here, use the actual line number?

RazvanN7: that would be redundant since errorSupplemental already provides the line number.

RazvanN7:
@UplinkCoder I gave it another shot creating CTErrorExp. Now the redundant error messages have disappeared. I also created an enum which accounts for different error messages which may be used in the future if other supplemental error message are needed.

src/dmd/dsymbolsem.d
wilzbach: Hmm imho you should still remove or replace "here" as it's rather a bad word for an error message. Maybe "with"?

RazvanN7:
It looks like the compiler version used in the test bots is outdated. I cannot the failures locally.

wilzbach:
There's no way to update gdc

stefan-koch-sociomantic:
@RazvanN7 This is not what I meant please extend `ErrorExp` and not `CTFEExp`.

wilzbach:
Also still failing auto-tester:

```
Assertion failed: (tv.expr == true), function test_visitors, file tests/cxxfrontend.c, line 159.
gmake[1]: *** [posix.mak:588: cxx-unittest] Abort trap (core dumped)
```

JonathanWilbur:
@RazvanN7 are you still working on this? I am having the problem that necessitated this PR myself. Developing a compiler is way outside my experience, but if I can do anything to help expedite this, you have my full support.

RazvanN7:
@JonathanWilbur From what I can remember, the fix couldn't be integrated because the testers use older versions of the compiler that do not accommodate these changes. I will take a look at it the following days and see how I can work around it.

RazvanN7:
@JonathanWilbur @UplinkCoder I gave this another try. What are your thoughts on this?

RazvanN7:
It  looks like AppVeyor fails because the error message is formatted differently. Any ideas on how to bypass this?

src/dmd/dsymbolsem.d
```diff
@@ -1215,9 +1215,21 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                    if (buff.offset != 0)
+                        buff.setsize(buff.offset-1);
+
                     // https://issues.dlang.org/show_bug.cgi?id=14166
                     // Don't run CTFE for the temporary variables inside typeof
                     dsym._init = dsym._init.initializerSemantic(sc, dsym.type, sc.intypeof == 1 ? INITnointerpret : INITinterpret);
+                    const init_err = dsym._init.isExpInitializer();
+                    if (init_err && init_err.exp.op == TOK.errExtraInfo)
+                    {
+                         errorSupplemental(dsym.loc, "compile time context created here: `%s`", buff.extractString());
```
jacob-carlborg: Looks like `buff.extractString()` contains a trailing newline on Windows, for some reason.

src/dmd/dsymbolsem.d
```diff
@@ -1215,9 +1215,26 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                 }
                 else
                 {
+                    import dmd.hdrgen : toCBuffer, HdrGenState;
+                    OutBuffer buff;
+                    HdrGenState hgs;
+                    toCBuffer(dsym, &buff, &hgs);
+                    if (buff.offset != 0)
+                    {
+                        version(Windows)
```
RazvanN7: I had to do this to get around the extra newline that appears on Windows when using Outbuffer.extractString

RazvanN7: If anybody has some knowledge on a better approach I will happily substitute this noobish approach

src/dmd/dsymbolsem.d
```diff
@@ -1215,9 +1215,26 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                            buff.setsize(buff.offset-1);
+                    }
+
                     // https://issues.dlang.org/show_bug.cgi?id=14166
                     // Don't run CTFE for the temporary variables inside typeof
                     dsym._init = dsym._init.initializerSemantic(sc, dsym.type, sc.intypeof == 1 ? INITnointerpret : INITinterpret);
+                    const init_err = dsym._init.isExpInitializer();
+                    if (init_err && init_err.exp.op == TOK.errExtraInfo)
+                    {
+                         errorSupplemental(dsym.loc, "compile time context created here: `%s`", buff.peekString());
```
JinShil: I think it's somewhat of an abuse to use the `hdrgen` code for this purpose, as that's not what it was intended for.

I don't think it's absolutely necessary to print the statement; the line number should be reasonably sufficient.  There are other places in the compiler where the line number alone does the job:

```
src/dmd/parse.d:4207:                        error("C++ name spaces not allowed here");
src/dmd/parse.d:4211:                        error("C++ mangle declaration not allowed here");
src/dmd/ctfeexpr.d:194:            errorSupplemental(loc, "thrown from here");
src/dmd/dinterpret.d:5404:            errorSupplemental(result.loc, "originally uninitialized here");
```

stefan-koch-sociomantic: Seconded!
there's no need to use the hdrgen here.

RazvanN7: Addressed!

RazvanN7:
Maybe we can finally merge this?

stefan-koch-sociomantic:
I am still wondering why you inherit from `Expression` rather then ErrorExp.

RazvanN7:
`ErrorExp` is a final class and it wasn't meant to be inherited.

stefan-koch-sociomantic:
Then definalize it ?

src/dmd/ctfeexpr.d
```diff
@@ -247,10 +247,26 @@ extern (C++) final class CTFEExp : Expression
...
 }
 
+/*
+ * This class is used to pass additional information
+ * about ctfe errors. For now it is used solely to
+ * signal a missing function body error but it can be
+ * extended to support more errors in the future, if needed,
+ * by adding a field which will represent the error.
+ */
+extern (C++) class CTFEErrorExtraInfoExp : Expression
```
JinShil: (1) I don't think creating a new expression is necessary....

src/dmd/dinterpret.d
```diff
@@ -4997,7 +4997,7 @@ public:
         if (!fd.fbody)
         {
             e.error("`%s` cannot be interpreted at compile time, because it has no available source code", fd.toChars());
-            result = CTFEExp.cantexp;
+            result = new CTFEErrorExtraInfoExp();
```
JinShil: (2)... I think you can just `return CTFEExp.cantexp;` here...

src/dmd/dsymbolsem.d
```diff
@@ -1218,6 +1218,11 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                     // https://issues.dlang.org/show_bug.cgi?id=14166
                     // Don't run CTFE for the temporary variables inside typeof
                     dsym._init = dsym._init.initializerSemantic(sc, dsym.type, sc.intypeof == 1 ? INITnointerpret : INITinterpret);
+                    const init_err = dsym._init.isExpInitializer();
+                    if (init_err && init_err.exp.op == TOK.errExtraInfo)
```
JinShil: (3) ...then check `init_err.exp.op == TOK.cantExpression`

src/dmd/tokens.d
```diff
@@ -693,6 +694,7 @@ extern (C++) struct Token
         TOK.interval: "interval",
         TOK.voidExpression: "voidexp",
         TOK.cantExpression: "cantexp",
+        TOK.errExtraInfo : "errExtraInfo",
```
JinShil: (4) ...and then you shouldn't need this anymore.

JinShil (changes requested):

src/dmd/tokens.d
JinShil: If that is not feasible, what about just adding a new `__gshared CTFEExp` here?  https://github.com/dlang/dmd/blob/ed6386ada4549d70797c3d09e51e05756de174d4/src/dmd/ctfeexpr.d#L233-L237

src/dmd/ctfeexpr.d
RazvanN7: I think it's better to have a class that can be easily extended for other ctfe error types that need extra information.

src/dmd/dinterpret.d
RazvanN7: CTFEExp.cantexp can't be used because the ctfe code usually returns when cantexp is encountered and we need to transport the extra information outside the ctfe context.

src/dmd/dsymbolsem.d
RazvanN7: cantExpression can be throwed from a lot of situations that don;t require additional information

src/dmd/tokens.d
RazvanN7: This seems like the right way to go. Thanks!

RazvanN7: Done.

RazvanN7:
ping @JinShil

src/dmd/ctfeexpr.d
```diff
@@ -235,6 +235,7 @@ extern (C++) final class CTFEExp : Expression
     extern (C++) static __gshared CTFEExp breakexp;
     extern (C++) static __gshared CTFEExp continueexp;
     extern (C++) static __gshared CTFEExp gotoexp;
+    extern (C++) static __gshared CTFEExp errinfoexp;
```
JinShil: I think we should consider if this name is still appropriate.  I suggest something link `showcontext`, `displaycontext`, `printcontext`, or something like that, though `errinfoexp` isn't bad.  Also, since this is a new symbol, it should have a documentation comment.

src/dmd/tokens.d
JinShil: Consider a different name, but like `showCtfeContext`, or something like that.  Though I'd be willing to merge this without a name change.

JinShil (changes requested):
Just add a documentation comment for any new public symbols and I think this is ready to go.

src/dmd/tokens.d
```diff
@@ -283,6 +283,7 @@ enum TOK : int
     interval = 231,
     voidExpression,
     cantExpression,
+    errExtraInfo,
```
jacob-carlborg: Please don’t abbreviate, you’re only saving two characters.

src/dmd/ctfeexpr.d
jacob-carlborg: Or even better using the D naming conventions.

RazvanN7: I would rather have the code I add be consistent with existing code in the same block.

RazvanN7:
@JinShil Done.

JinShil (approved):

src/dmd/ctfeexpr.d
```diff
@@ -235,6 +235,10 @@ extern (C++) final class CTFEExp : Expression
     extern (C++) static __gshared CTFEExp breakexp;
     extern (C++) static __gshared CTFEExp continueexp;
     extern (C++) static __gshared CTFEExp gotoexp;
+    /* Used when additional information is needed regarding
+     * a ctfe error.
+     */
+    extern (C++) static __gshared CTFEExp showcontext;
```
jacob-carlborg: Still not using camel case.

wilzbach:
> Please don’t abbreviate, you’re only saving two characters.

Considering how likely it is that we won't change the name again, it would be really nice not to save those to characters.

>> Still not using camel case.
>> Or even better using the D naming conventions.
> I would rather have the code I add be consistent with existing code in the same block.

You gotta start somewhere ;-)

RazvanN7:
ping

ibuclaw:
This caused a regression:

https://issues.dlang.org/show_bug.cgi?id=19140

PR #8050, opened 2018-03-18
Fix issue 18242 - Warn if object.Exception is missing

LemonBoy:
Let's not segfault if no `Exception` class is defined.

### DIFF
--- a/test/compilable/b18242.d
+++ b/test/compilable/b18242.d
@@ -6,7 +6,7 @@ class Object { }
 
 class TypeInfo { }
 class TypeInfo_Class : TypeInfo { 
-    version(D_LP64) { ubyte[136] _x; } else { ubyte[68] _x };
+    version(D_LP64) { ubyte[136] _x; } else { ubyte[68] _x; }
 }
 
 class Throwable { }
--- a/test/compilable/b18242.d
+++ b/test/compilable/b18242.d
@@ -1,0 +1,15 @@
+// REQUIRED_ARGS: -c
+// PERMUTE_ARGS:
+module object;
+
+class Object { }
+
+class TypeInfo { }
+class TypeInfo_Class : TypeInfo { ubyte[136] _x; }
+
+class Throwable { }
+
+int _d_run_main() {
+    try { } catch(Throwable e) { return 1; }
+    return 0;
+}
--- a/test/compilable/b18242.d
+++ b/test/compilable/b18242.d
@@ -5,13 +5,15 @@ module object;
 class Object { }
 
 class TypeInfo { }
-class TypeInfo_Class : TypeInfo { 
+class TypeInfo_Class : TypeInfo 
+{ 
     version(D_LP64) { ubyte[136] _x; } else { ubyte[68] _x; }
 }
 
 class Throwable { }
 
-int _d_run_main() {
+int _d_run_main() 
+{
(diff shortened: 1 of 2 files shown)


wilzbach:
How about adding a test like it was done in https://github.com/dlang/dmd/pull/7799 or https://github.com/dlang/dmd/pull/7786?

JinShil:
If `Throwable` is declared, but `Exception` isn't, one should still be able to use try-catch statements.  I think the fix should be placed elsewhere where the `Exception` symbol/identifier is used (perhaps when searching the symbol table for `Exception` in the semantic processing of `Identifier` or `IdentifierExp`).

WalterBright:
When creating a PR to fix a bugzilla issue, please post a link to the PR in the issue. This is so that someone browsing bug reports will know if and where a PR for it is. I did it for this one.

jacob-carlborg:
> When creating a PR to fix a bugzilla issue, please post a link to the PR in the issue. This is so that someone browsing bug reports will know if and where a PR for it is. I did it for this one.

Isn’t there a bot that does this automatically?

LemonBoy:
> I think the fix should be placed elsewhere where the Exception symbol/identifier is used

Let's try with this other approach then.

WalterBright:
> Isn’t there a bot that does this automatically?

No.

wilzbach:
> Isn’t there a bot that does this automatically?

https://github.com/dlang-bots/dlang-bot/issues/135

JinShil:
I just tested this and it looks good to me.  It just needs a test case.

test/compilable/b18242.d
```diff
@@ -0,0 +1,15 @@
+// REQUIRED_ARGS: -c
```
JinShil: If you add `DFLAGS:` (without anything following it) it will prevent all the default flags for importing the runtime, etc.. from being added implicitly to compiler invocation.  Then you shouldn't need to declare the `TypeInfo` stuff and have to deal with all object mismatch errors.

See #7845

JinShil: Sorry, Ignore the comment above.  It appears `TypeInfo` is required because `Throwable` has been declared.  This should get you passed the 32-bit tests:

```D
module object;

class Object { }

class Throwable { }

class TypeInfo { }
class TypeInfo_Class : TypeInfo 
{ 
    version(D_LP64)
    {
        ubyte[136] _x; 
    }
    else
    {
        ubyte[68] _x;
    }
}

int _d_run_main() 
{
    try { } catch(Throwable e) { return 1; }
    return 0;
}
```

LemonBoy: Thanks for the hint, I always forget about 32bit platforms.

test/compilable/b18242.d
```diff
@@ -0,0 +1,17 @@
...
+// PERMUTE_ARGS:
+module object;
+
+class Object { }
+
+class TypeInfo { }
+class TypeInfo_Class : TypeInfo { 
+    version(D_LP64) { ubyte[136] _x; } else { ubyte[68] _x };
```
JinShil: I think you need to remove the semicolon at the end of this line.

JinShil: I modified it (though I misspelled *fix* in the commit message) 😢

JinShil (approved):

wilzbach (approved):

PR #8051, opened 2018-03-19
fix Issue 18199 - Error with lambda in struct initializer

belm0:
Allow struct initializers to include all function literal forms.

Previously, a function literal containing semicolon or return tokens
would cause the containing struct initializer to be incorrectly
treated as a parameterless funtion literal (i.e. {statements...}).

The issue is resolved by only aborting the struct initializer
lookahead when these tokens appear at the top curly bracket scope.

Also, document another ambiguous case in function literal vs.
struct initializer, as well as explain why these are resolved as
struct initializer.

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -6029,19 +6029,59 @@ final class Parser(AST) : Lexer
         switch (token.value)
         {
         case TOK.leftCurly:
-            /* Scan ahead to see if it is a struct initializer or
-             * a function literal.
-             * If it contains a ';', it is a function literal.
-             * Treat { } as a struct initializer.
+            /* Scan ahead to discern between a struct initializer and
+             * parameterless function literal.
+             *
+             * We'll scan the topmost curly bracket level for statement-related
+             * tokens, thereby ruling out a struct initializer.  (A struct
+             * initializer which itself contains function literals may have
+             * statements at nested curly bracket levels.)
+             *
+             * It's important that this function literal check not be
+             * pendantic, otherwise a function having the slightest syntax
+             * error would emit confusing errors when we proceed to parse it
+             * as a struct initializer.
+             *
+             * The following two ambiguous cases will be treated as a struct
+             * initializer (best we can do without type info):
+             *     {}
+             *     {{statements...}}  - i.e. it could be struct initializer
+             *        with one function literal, or function literal having an
+             *        extra level of curly brackets
+             * If a function literal is intended in these cases (unlikely),
+             * source can use a more explicit function literal syntax
+             * (e.g. prefix with "()" for empty parameter list).
              */
             braces = 1;
             for (t = peek(&token); 1; t = peek(t))
             {
                 switch (t.value)
                 {
+                /* Look for a semicolon or keyword of statements which don't
+                 * require a semicolon (typically containing BlockStatement).
+                 * Tokens like "else", "catch", etc. are omitted where the
+                 * leading token of the statement is sufficient.
+                 */
+                case TOK.asm_:
+                case TOK.class_:
+                case TOK.debug_:
+                case TOK.enum_:
+                case TOK.if_:
+                case TOK.interface_:
+                case TOK.pragma_:
+                case TOK.scope_:
                 case TOK.semicolon:
-                case TOK.return_:
-                    goto Lexpression;
+                case TOK.struct_:
+                case TOK.switch_:
+                case TOK.synchronized_:
+                case TOK.try_:
+                case TOK.union_:
+                case TOK.version_:
+                case TOK.while_:
+                case TOK.with_:
+                    if (braces == 1)
+                        goto Lexpression;
+                    continue;
 
                 case TOK.leftCurly:
                     braces++;
--- a/test/compilable/test18199.d
+++ b/test/compilable/test18199.d
@@ -1,0 +1,87 @@
+// https://issues.dlang.org/show_bug.cgi?id=18199
+
+//
+// struct initializer cases
+//
+
+// original error report
+struct Bla
+{
+    int delegate(int, int) fun;
+}
+Bla bla1 = Bla((int a, int b) { return a + b; });
+Bla bla2 = {(int a, int b) { return a + b; }};  // yielded error
+
+// additional error report with memberName:expression syntax
+struct Foo
+{
+    int function(int) bar;
+    int function(int) bar2;
+}
+Foo foo =
+{
+    bar : function(x) { return 2 * x; },  // yielded error
+    bar2 : (x) => 2 * x,
+};
+
+struct MyStruct
+{
+    int function() f;
+    int delegate() d;
+}
+
+// confirm that ambiguous cases assume struct initializer
+MyStruct ambiguous_1 = {};
+MyStruct ambiguous_2 =
+{
+    { return 1 + 1; }
+};
+
+// statement-holding function literal variants not covered above
+static MyStruct function_and_delegate_keywords =
+{
+    function () { return 1 + 1; },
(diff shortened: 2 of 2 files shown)


belm0:
Test Projects caught a regression.  In [this code](https://github.com/etcimon/libasync/blob/master/source/libasync/socket.d#L545) from vibed/libasync, RHS is getting incorrectly treated as a struct initializer:
```
auto socketType = {
	switch(type) {
		case SOCK_STREAM:    return SocketType.STREAM;
		case SOCK_SEQPACKET: return SocketType.SEQPACKET;
		...
	}
}();
```
I.e. since there are no comma or return tokens at top level, RHS is parsed as struct.

Well, actually this points us to another way to break the production dmd:
```
alias MyFun = void function();
static MyFun test2 = {
  final switch(5) {
  }
};
----
Error: expression expected, not final
Error: comma expected separating field initializers
Error: expression expected, not switch
Error: comma expected separating field initializers
```
I.e. there are many ways to have a function body with no comma or return tokens, which causes the existing code to incorrectly parse RHS as a struct initializer.

Back to the drawing board.  Either we need an isFunctionLiteral or isStructInitializer in parser.d, or need to punt somehow to a higher semantic pass?

test/compilable/test18199.d
```diff
@@ -0,0 +1,19 @@
...
+static MyStruct test1 = {
+    { return 1 + 1; }
+};
+
+alias MyFun = int function();
+
+// RHS is function literal
+static MyFun test2 = {
+    return 1 + 1;
+};
```
WalterBright: Missing test case for https://issues.dlang.org/show_bug.cgi?id=18199#c2, and comment 5

belm0:
@WalterBright I'd like to consult with you on how to proceed.  Summary:

  * the parser's existing use of semicolon / return as function literal test is not a robust approach-- it can have both positive and negative failures.  My attempt to salvage it has failed (see regression).
  * faced with implementing a robust lookahead for either FunctionLiteralBody or struct initializer, the latter seems easier.  However I'm concerned with the resulting "fall back to function literal" behavior, since users will be faced with confusing error messages when the compiler tries to parse their malformed struct initializer as a function literal.
  * I don't know enough about the parsing levels to know if there's an easier way out (e.g. can we get type info)

Thank you

test/compilable/test18199.d
belm0: added tests covering bug report, thank you

belm0:
I was able to salvage the "lookahead for evidence of function literal" approach of the original code.

test/compilable/test18199.d
```diff
@@ -0,0 +1,62 @@
+// REQUIRED_ARGS:
+// PERMUTE_ARGS:
+// https://issues.dlang.org/show_bug.cgi?id=18199
+
+//
+// initializer is a struct initializer
+//
+
+// original error report
+struct Bla {
```
JinShil: Although I don't think it should prevent this PR from being pulled, I'd prefer if the code followed the general D coding style, that is curly braces on their own line.

belm0: done, thank you for raising

test/compilable/test18199.d
```diff
@@ -0,0 +1,73 @@
+// REQUIRED_ARGS:
+// PERMUTE_ARGS:
```
wilzbach: Those two lines aren't required. On the contrary, `PERMUTATE_ARGS` shouldn't be specified manually (except if there's a good reason to do so)

src/dmd/parse.d
```diff
@@ -6029,19 +6029,56 @@ final class Parser(AST) : Lexer
...
-             * Treat { } as a struct initializer.
+            /* Scan ahead to discern between struct initializer and
+             * parameterless function literal.
+             *
+             * We'll scan the topmost curly bracket level for statement-related
+             * tokens, thereby ruling out a struct initializer.  (A struct
+             * initializer which itself containins function literals may have
+             * statements at nested curly bracket levels.)
+             *
+             * It's important that this functional literal check not be
```
wilzbach: It's important for this function literal check to be

src/dmd/parse.d
```diff
@@ -6029,19 +6029,56 @@ final class Parser(AST) : Lexer
...
+             * pendantic, otherwise a function having the slightest syntax
+             * error would emit confusing errors when we proceed to parse it
+             * as a struct initializer.
```
wilzbach: This paragraph doesn't help the reader as that's really obvious. I suggest to drop it.

test/compilable/test18199.d
belm0: remove, thank you

src/dmd/parse.d
belm0: Not sure I understand.  You'd like to change the polarity of the statement, from "the check should not be pedantic" to "the check should be pedantic"?

Responding to your other comment, this matter was indeed not obvious to me so seems worthy of source code comment.  If the author _intended_ a function literal (implies heuristics as employed here), then even when that literal has a syntax error we want to parse it as a function and emit understandable errors.

belm0: Ah I think you just meant "functional" --> "function" here.   Done.

I'll move the other matter to your corresponding comment.

src/dmd/parse.d
belm0: [copied from other thread] This matter was indeed not obvious to me so seems worthy of source code comment. If the author intended a function literal (implies heuristics as employed here), then even when that literal has a syntax error we want to parse it as a function and emit understandable errors.

src/dmd/parse.d
wilzbach: Oh I was just trying to point out that "It's important that XXX not be" is missing a verb or in other words "not be" sounded really strange.

src/dmd/parse.d
```diff
@@ -6029,19 +6029,56 @@ final class Parser(AST) : Lexer
...
-             * Treat { } as a struct initializer.
+            /* Scan ahead to discern between struct initializer and
+             * parameterless function literal.
+             *
+             * We'll scan the topmost curly bracket level for statement-related
+             * tokens, thereby ruling out a struct initializer.  (A struct
+             * initializer which itself contains function literals may have
+             * statements at nested curly bracket levels.)
+             *
+             * It's important that this function literal check not be
```
wilzbach: So I was trying to say that this should probably be e.g. "check isn't" as "not be" sounds really strange here and I'm not sure that actually works with English grammar.

belm0: I believe the grammar is correct either way.  Pondering why I worded it this way, it was to avoid the uncomfortable pattern "it is [...] is not".  E.g.:
>     "It is important that this elephant is not in the room." vs.
>     "It is important that this elephant not be in the room."

wilzbach (approved):
Looks reasonable to me.

src/dmd/parse.d
```diff
@@ -6029,19 +6029,56 @@ final class Parser(AST) : Lexer
...
+                case TOK.pragma_:
                 case TOK.semicolon:
-                case TOK.return_:
-                    goto Lexpression;
+                case TOK.scope_:
+                case TOK.switch_:
+                case TOK.synchronized_:
+                case TOK.try_:
+                case TOK.while_:
+                case TOK.with_:
```
JinShil: I think each of these should have a test case.

belm0: Done, thanks for raising.  Actually, reviewing the possible no-semicolon statements for this exercise exposed a lot of missing tokens.

JinShil (approved):

NilsLankila:
This causes a regression in StructInitializers, see https://issues.dlang.org/show_bug.cgi?id=21163.

PR #8052, opened 2018-03-19
fix Issue 18534 - Wrong code for ?: operator when compiling with -O

WalterBright:
This is a reboot of https://github.com/dlang/dmd/pull/8014


JinShil (approved):
I don't quite understand what this code is doing, but I don't see anything alarming about it.  Approving on faith in Walter.

WalterBright:
1. for 32 bit platforms, it made sure to select a byte register
2. for 64 bit platforms, it was using the wrong variable to test for a byte operation

wilzbach:
The OSX failure is unrelated.

PR #8053, opened 2018-03-19
Remove NCEG operators from the language

JinShil:
These were made an error long ago in 2.072 (See https://dlang.org/deprecate.html).  It's always nice to keep the shop clean.

Spec PR: https://github.com/dlang/dlang.org/pull/2280


jacob-carlborg (approved):

ibuclaw:
Not so sure about removing support from the backend though, there may be another frontend that uses it.

src/dmd/backend/evalu8.c
```diff
@@ -1599,63 +1599,55 @@ elem * evalu8(elem *e, goal_t goal)
...
         // BUG: complex numbers
         i ^= d1 !<>= d2;
         e->EV.Vint = i;
         break;
 
     case OPnlg:
         i++;
-    case OPlg:
         // BUG: complex numbers
         i ^= d1 <> d2;
```
ibuclaw: Infact this suggests that dmc supports nceg operators still, so definitely don't remove backend support.

ibuclaw (dismissed):

JinShil:
Backend support has been restored.

src/dmd/tokens.d
```diff
@@ -94,18 +94,6 @@ enum TOK : int
...
-    unord,
-    lg,
-    leg,
-    ule,
-    ul,
-    uge,
-    ug,
-    ue,
-
     // 72
```
wilzbach: I guess this number should be removed too as it gets outdated too easily.

wilzbach (approved):
Nice! Thanks a lot for cleaning up the legacy baggage.

src/dmd/tokens.d
JinShil: I subtracted 8 from each of the numbers to try and keep the status quo.

WalterBright:
> Backend support was restored

Thank you, it's used by DMC/C++.

JinShil:
Closing and reopening in an attempt to wake up Semaphore CI.

PR #8054, opened 2018-03-19
fix Issue 17784 - [scope][DIP1000] Confusing error message for escapi…

WalterBright:
…ng local via new-expression

This splits off `checkNewEscape()` from `checkReturnEscapeImpl()` which enables customizing the error messages and removes confusing logic trying to deal with both Return and New.

### DIFF
--- a/src/dmd/escape.d
+++ b/src/dmd/escape.d
@@ -585,8 +585,142 @@ bool checkThrowEscape(Scope* sc, Expression e, bool gag)
 bool checkNewEscape(Scope* sc, Expression e, bool gag)
 {
     //printf("[%s] checkNewEscape, e = %s\n", e.loc.toChars(), e.toChars());
-    return checkReturnEscapeImpl(sc, e, false, gag, false);
+    enum log = false;
+    if (log) printf("[%s] checkNewEscape, e: `%s`\n", e.loc.toChars(), e.toChars());
+    EscapeByResults er;
+
+    escapeByValue(e, &er);
+
+    if (!er.byref.dim && !er.byvalue.dim && !er.byexp.dim)
+        return false;
+
+    bool result = false;
+    foreach (VarDeclaration v; er.byvalue)
+    {
+        if (log) printf("byvalue `%s`\n", v.toChars());
+        if (v.isDataseg())
+            continue;
+
+        Dsymbol p = v.toParent2();
+
+        if (v.isScope())
+        {
+            if (sc._module && sc._module.isRoot() &&
+                /* This case comes up when the ReturnStatement of a __foreachbody is
+                 * checked for escapes by the caller of __foreachbody. Skip it.
+                 *
+                 * struct S { static int opApply(int delegate(S*) dg); }
+                 * S* foo() {
+                 *    foreach (S* s; S) // create __foreachbody for body of foreach
+                 *        return s;     // s is inferred as 'scope' but incorrectly tested in foo()
+                 *    return null; }
+                 */
+                !(p.parent == sc.func))
+            {
+                // Only look for errors if in module listed on command line
+                if (global.params.vsafe) // https://issues.dlang.org/show_bug.cgi?id=17029
+                {
+                    if (!gag)
+                        error(e.loc, "scope variable `%s` may not be copied into allocated memory", v.toChars());
+                    result = true;
+                }
+                continue;
+            }
+        }
+        else if (v.storage_class & STC.variadic && p == sc.func)
+        {
+            Type tb = v.type.toBasetype();
+            if (tb.ty == Tarray || tb.ty == Tsarray)
+            {
+                if (!gag)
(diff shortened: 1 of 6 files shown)


src/dmd/escape.d
```diff
@@ -585,9 +585,142 @@ bool checkThrowEscape(Scope* sc, Expression e, bool gag)
...
+                 *        return s;     // s is inferred as 'scope' but incorrectly tested in foo()
+                 *    return null; }
+                 */
+                !(p.parent == sc.func))
+            {
+                // Only look for errors if in module listed on command line
+                if (global.params.vsafe) // https://issues.dlang.org/show_bug.cgi?id=17029
+                {
+                    if (!gag)
+                        error(e.loc, "scope variable `%s` may not be copied into allocated memory", v.toChars());
```
JinShil: I suggest using the term "GC-allocated memory" to be more precise.  The problem isn't about memory (per se) or its allocation but about the lifetimes of objects under the jurisdiction of the GC.  I suggest something like:

> scope variable `%s` may not be copied into GC-allocated memory because doing so would cause it to escape

JinShil:
Looks like this needs additional test cases to increase coverage.

JinShil:
We run the risk of turning D into Rust, where instead of users fighting the borrow checker, they'll be fighting the scope checker.  The wording of the error messages will become increasingly important to help users understand precisely what the problem is.

src/dmd/escape.d
```diff
@@ -585,9 +585,142 @@ bool checkThrowEscape(Scope* sc, Expression e, bool gag)
...
+                continue;
+            }
+        }
+        else if (v.storage_class & STC.variadic && p == sc.func)
+        {
+            Type tb = v.type.toBasetype();
+            if (tb.ty == Tarray || tb.ty == Tsarray)
+            {
+                if (!gag)
+                    error(e.loc, "copying `%s` into allocated memory escapes a reference to variadic parameter `%s`", e.toChars(), v.toChars());
```
JinShil: My suggestion:

> copying `%s` into GC-allocated memory escapes a reference to variadic parameter `%s`

src/dmd/escape.d
```diff
@@ -585,9 +585,142 @@ bool checkThrowEscape(Scope* sc, Expression e, bool gag)
...
+    foreach (VarDeclaration v; er.byref)
+    {
+        if (log) printf("byref `%s`\n", v.toChars());
+
+        void escapingRef(VarDeclaration v)
+        {
+            if (!gag)
+            {
+                const(char)* kind = (v.storage_class & STC.parameter) ? "parameter" : "local";
+                error(e.loc, "copying `%s` into allocated memory escapes a reference to %s variable `%s`",
```
JinShil: I suggest:

> copying `%s` into GC-allocated memory escapes a reference to %s variable `%s`

src/dmd/escape.d
```diff
@@ -585,9 +585,142 @@ bool checkThrowEscape(Scope* sc, Expression e, bool gag)
...
+                }
+                // Don't need to be concerned if v's parent does not return a ref
+                FuncDeclaration fd = p.isFuncDeclaration();
+                if (fd && fd.type && fd.type.ty == Tfunction)
+                {
+                    TypeFunction tf = cast(TypeFunction)fd.type;
+                    if (tf.isref)
+                    {
+                        if (!gag)
+                            error(e.loc, "storing reference to outer local variable `%s` into allocated memory", v.toChars());
```
JinShil: Not sure I understand what the intention is, but if my understanding is correct I suggest:

> storing a reference to outer local variable `%s` into GC-allocated memory causes it to escape

src/dmd/escape.d
```diff
@@ -585,9 +585,142 @@ bool checkThrowEscape(Scope* sc, Expression e, bool gag)
...
+
+            }
+        }
+    }
+
+    foreach (Expression ee; er.byexp)
+    {
+        if (log) printf("byexp %s\n", ee.toChars());
+        if (!gag)
+            error(ee.loc, "storing reference to stack allocated value returned by `%s` into allocated memory",
```
JinShil: I suggest:

> storing a reference to stack allocated value returned by `%s` into GC-allocated memory causes it to escape

JinShil (dismissed):
Please add additional tests to increase coverage.

ZombineDev:
In principle I agree with @JinShil, but I think that the source of the momery allocation (e.g. GC vs malloc vs static) doesn't matter, so it's better to simply say "storing / copying [..] to an object with greater lifetime".

src/dmd/escape.d
WalterBright: I didn't use that term because the message could apply to any allocated memory.

WalterBright:
> with greater lifetime

I tend to think of lifetimes in relationship to lexical scopes, so I prefer using "allocated memory" instead.

JinShil (approved):

PR #8056, opened 2018-03-19
[WIP] Fix issue 17770 - Null pointer access in CTFE code

LemonBoy:
The lwr/upr fields may be null if the default values are to be used.

CC The CTFE Man @UplinkCoder

### DIFF
--- a/test/compilable/b17770.d
+++ b/test/compilable/b17770.d
@@ -1,0 +1,24 @@
+/*
+REQUIRED_ARGS: -c
+PERMUTE_ARGS:
+TEST_OUTPUT:
+---
+0
+---
+*/
+struct S { T* t; }
+struct T { string name; }
+
+S foo(string name)
+{
+    return S(new T(name[0 .. $]));
+}
+
+int bar(string name)
+{
+    return cast(int)name.length;
+}
+
+const S s = foo("");
+enum b = bar(s.t.name);
+pragma(msg, b);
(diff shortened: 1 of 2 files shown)


WalterBright:
@LemonBoy please add a link to this PR to https://issues.dlang.org/show_bug.cgi?id=17770

stefan-koch-sociomantic:
LGTM nice catch! @LemonBoy

stefan-koch-sociomantic:
Tough there still seems to be an error here ...
I would propose this: If you detect a [0 .. $] Slice, return the original expression.
Which will get you out of the business of this hack.

LemonBoy:
Hmm, that's a nice solution and it also allows this snippet to compile, let's see what the testing pipeline thinks of it.
In case you're wondering why the `lwr`/`uwr` are null that's because of `optimize.d` .

LemonBoy:
Back to the drawing board, we must handle the endpoints being null since `optimize.d` may do so for strings (or call/replicate `Expression_optimize` at the end of the `visit` function).

test/compilable/b17770.d
```diff
@@ -0,0 +1,24 @@
+/*
+REQUIRED_ARGS: -c
+PERMUTE_ARGS:
```
wilzbach: FYI: -c is redundant and `PERMUTE_ARGS` doesn't need to be defined as well.

RazvanN7:
There doesn't seem to be any activity on this PR so closing for now.

PR #8058, opened 2018-03-20
outbuf: check for overflow errors

WalterBright:
This is an attempt to track down the dmd crash in OSX 32. Don't pull it.


wilzbach:
We don't seem to get the OSX failure anymore... (looks like we were lucky).
Is this still active?

RazvanN7:
ping @WalterBright . Can we close this?

PR #8059, opened 2018-03-21
Verify consecutive token values

marler8997:
Added a `static assert` in tokens.d that checks that all token enum members are consecutive.  It does this by verifying that each consecutive member values matches it's index.  This caught a mistake on line 284 so that was also fixed.

### DIFF
--- a/src/dmd/tokens.d
+++ b/src/dmd/tokens.d
@@ -81,9 +81,8 @@ enum TOK : int
     delegatePointer,
     delegateFunctionPointer,
 
-    // 54
     // Operators
-    lessThan,
+    lessThan = 54,
     greaterThan,
     lessOrEqual,
     greaterOrEqual,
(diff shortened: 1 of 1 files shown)


src/dmd/tokens.d
```diff
@@ -281,8 +271,7 @@ enum TOK : int
     ref_,
     macro_,
 
-    // 221
```
marler8997: This should have been `// 220`.  This mistake is no longer possible with this PR.

src/dmd/tokens.d
```diff
@@ -312,6 +300,18 @@ enum TOK : int
     max_,
 }
 
+// Assert that all token enum members have consecutive values and
+// that none of them overlap
+static assert(() {
```
JinShil: This will get executed with every compiler invocation, correct?  Can we maybe put it in a `unittest` block (or is there some other solution)?

JinShil: Don't know what I was thinking.  This will only be executed when DMD is built.  It's good!

JinShil (approved):

src/dmd/tokens.d
marler8997: Every time dmd is "compiled", but not when the resulting compiler is invoked.

WalterBright: Clever - nice work!

PR #8062, opened 2018-03-21
fix Issue 18644 - [dip1000] escape of outer local not detected

WalterBright:


JinShil (approved):
I would prefer to avoid the code duplication.

WalterBright:
> duplication

It used to be all together, with a maze of conditional statements. It was impossible to deal with.

PR #8065, opened 2018-03-21
fix Issue 18461 - codegen bug - OPbt expressions and assignments to a…

WalterBright:
…mbiguous symbols

Reboot of https://github.com/dlang/dmd/pull/8015


ibuclaw (approved):
Walter's backend, trust he knows what he is doing.

PR #8066, opened 2018-03-21
Fix Issue 18647 - Use of delete should be allowed without a deprecation in a deprecated scope

wilzbach:
Targetting stable as this is rather relevant to the `delete` deprecation that was part of 2.079

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -5130,12 +5130,15 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
 
     override void visit(DeleteExp exp)
     {
-        // @@@DEPRECATED_2019-02@@@
-        // 1. Deprecation for 1 year
-        // 2. Error for 1 year
-        // 3. Removal of keyword, "delete" can be used for other identities
-        if (!exp.isRAII)
-            deprecation(exp.loc, "The `delete` keyword has been deprecated.  Use `object.destroy()` (and `core.memory.GC.free()` if applicable) instead.");
+        if (!sc.isDeprecated)
+        {
+            // @@@DEPRECATED_2019-02@@@
+            // 1. Deprecation for 1 year
+            // 2. Error for 1 year
+            // 3. Removal of keyword, "delete" can be used for other identities
+            if (!exp.isRAII)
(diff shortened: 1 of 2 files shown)


JinShil (approved):

src/dmd/expressionsem.d
```diff
@@ -5130,12 +5130,15 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
-            deprecation(exp.loc, "The `delete` keyword has been deprecated.  Use `object.destroy()` (and `core.memory.GC.free()` if applicable) instead.");
+        if (!sc.isDeprecated)
+        {
+            // @@@DEPRECATED_2019-02@@@
+            // 1. Deprecation for 1 year
+            // 2. Error for 1 year
+            // 3. Removal of keyword, "delete" can be used for other identities
+            if (!exp.isRAII)
+                deprecation(exp.loc, "The `delete` keyword has been deprecated.  Use `object.destroy()` (and `core.memory.GC.free()` if applicable) instead.");
+        }
```
mathias-lang-sociomantic: The bug report was more general (before you changed the title).
I took delete as an example, because it's what hits me, but I'm pretty sure it affects other deprecations.
For example:
```D
deprecated void main ()
{
    int i, o;
    switch (i)
    {
    case 1:
        o = 1;
    case 2:
        o = 2;
        break;
    default:
    }
}
```

wilzbach: OKay I created a new subissue 18647 of 17906 which just addresses the delete deprecation.
Maybe `deprecation` should require the current scope, s.t. `sc.isDeprecated` can automatically be called (and we don't need to patch this in at all places)?

WalterBright:
@wilzbach In the future, please note in bugzilla that this PR exists. I did it for this one.

test/compilable/test17906.d
```diff
@@ -0,0 +1,6 @@
+// REQUIRED_ARGS: -de
```
WalterBright: Please add the corresponding bugzilla issue as a comment.

WalterBright: I added it.

WalterBright (approved):

PR #8069, opened 2018-03-22
Set -euo pipefail by default for all bash scripts

wilzbach:
- `-e` - abort if a command fails
- `-u` - abort if an undefined variable is accessed
- `-o pipefail` - abort if a command in a pipe fails

These settings are commonly known as "Safe Bash" or "Strict Bash" and imho there's no reason why these shouldn't be the default.

CC @marler8997 @CyberShadow

### DIFF
--- a/test/runnable/depsprot.sh
+++ b/test/runnable/depsprot.sh
@@ -1,9 +1,12 @@
 #!/usr/bin/env bash
 
-name=`basename $0 .sh`
+name="depsprot"
 dir=${RESULTS_DIR}/runnable
 dmddir=${RESULTS_DIR}${SEP}runnable
 deps_file="${dmddir}${SEP}${name}.deps"
+
+# custom error handling
+set +eo pipefail
 
 die()
 {
--- a/test/sh_do_test.sh
+++ b/test/sh_do_test.sh
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -euo pipefail
+
 if [ "${RESULTS_DIR}" == "" ]; then
     echo Note: this program is normally called through the Makefile, it
     echo is not meant to be called directly by the user.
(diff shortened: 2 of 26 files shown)


test/sh_do_test.sh
```diff
@@ -60,6 +62,11 @@ exec 2>&1
 # log all a verbose xtrace to a temporary file which is only displayed when an error occurs
 exec 42> "${output_file}.log"
 export BASH_XTRACEFD=42
+
+# fail on errors and undefined variables
+set -euo pipefail
```
wilzbach: This is duplicated with the header for explicitness.

test/runnable/depsprot.sh
```diff
@@ -1,10 +1,13 @@
 #!/usr/bin/env bash
 
-name=`basename $0 .sh`
+name="depsprot"
 dir=${RESULTS_DIR}/runnable
 dmddir=${RESULTS_DIR}${SEP}runnable
 deps_file="${dmddir}${SEP}${name}.deps"
 
+# custom error handling
+set +eo pipefail
```
jacob-carlborg: What’s the difference between `-` and `+`?

jacob-carlborg (approved):

test/runnable/depsprot.sh
wilzbach: This disables these settings as this test currently uses a custom approach to error handling. I might remove the custom approach here too, but I thought I leave that for another PR (and it's not really important if one file doesn't use pipefail. It's important that it's the default).

CyberShadow:
>  Revert to non automatic file-descriptor allocation 

How did that end up in master and worked? Did it fail silently?

wilzbach:
> How did that end up in master and worked? Did it fail silently?

Yep. I think it just piped the file stream into the nowhere, which wasn't noticed because for the normal run without errors it passed.
When I enabled `set -u` for `sh_do_test` I got the following error for macOS:

```
./sh_do_test.sh: line 57: exec: {stdout_copy}: not found
./sh_do_test.sh: line 38: stdout_copy: unbound variable
```

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3087217&isPull=true

And now I look at the bash version:

```
SHELL(/bin/sh): GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin16) Copyright (C) 2007 Free Software Foundation, Inc.
```

(I'm sorry - I should have looked at that earlier and not just assumed that green means green):

CyberShadow (approved):

PR #8070, opened 2018-03-22
Run individual tests with all available cores by default

wilzbach:
I just realized that we can use this wrapper to conveniently run all shell scripts, e.g.

```bash
./run_individual_tests compilable/*.sh
./run_individual_tests **/*.sh # requires globbing support by your shell
```

For this and imho in general, it's a sane default to use all available cores.
It can be opted-out with `PROCS=1` (though I doubt that this is a frequent need).

### DIFF
--- a/test/run_individual_tests
+++ b/test/run_individual_tests
@@ -9,6 +9,9 @@
 
 set -euo pipefail
 
+# if PROCS is not set, auto-detect or fallback to 4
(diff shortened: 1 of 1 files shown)


WalterBright:
I'm concerned this will interleave the output of multiple threads, making it impractical to view the log files when the tests fail.

wilzbach:
This isn't run on any CI, but just for local convenience use where one can simply rerun the specific failing file. It has no effect when the script is run with only one file (except that building the build tools initially is faster)
Also all CIs already use -j4, -j8 or similar when they run the testsuite.

PR #8071, opened 2018-03-23
fix Issue 5227 - X ^^ FP at compile-time

WalterBright:
Adds the CTFE functions:
```
real_t nearbyint(real_t x);
real_t rint(real_t x);
real_t round(real_t x);
real_t floor(real_t x);
real_t ceil(real_t x);
real_t trunc(real_t x);
real_t log(real_t x);
real_t log2(real_t x);
real_t log10(real_t x);
real_t pow(real_t x, real_t y);
real_t expm1(real_t x);
real_t exp2(real_t x);
real_t fmin(real_t x, real_t y);
real_t fmax(real_t x, real_t y);
real_t copysign(real_t x, real_t s);
real_t fma(real_t x, real_t y, real_t z);
```

### DIFF
--- a/src/dmd/builtin.d
+++ b/src/dmd/builtin.d
@@ -103,6 +103,116 @@ extern (C++) Expression eval_ldexp(Loc loc, FuncDeclaration fd, Expressions* arg
     return new RealExp(loc, CTFloat.ldexp(arg0.toReal(), cast(int) arg1.toInteger()), arg0.type);
 }
 
+extern (C++) Expression eval_log(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.log(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_log2(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.log2(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_log10(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.log10(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_expm1(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.expm1(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_exp2(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.exp2(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_round(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.round(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_floor(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.floor(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_ceil(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.ceil(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_trunc(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.trunc(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_copysign(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    Expression arg1 = (*arguments)[1];
+    assert(arg1.op == TOK.float64);
+    return new RealExp(loc, CTFloat.copysign(arg0.toReal(), arg1.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_pow(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    Expression arg1 = (*arguments)[1];
+    assert(arg1.op == TOK.float64);
+    return new RealExp(loc, CTFloat.pow(arg0.toReal(), arg1.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_fmin(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    Expression arg1 = (*arguments)[1];
+    assert(arg1.op == TOK.float64);
+    return new RealExp(loc, CTFloat.fmin(arg0.toReal(), arg1.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_fmax(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    Expression arg1 = (*arguments)[1];
+    assert(arg1.op == TOK.float64);
+    return new RealExp(loc, CTFloat.fmax(arg0.toReal(), arg1.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_fma(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    Expression arg1 = (*arguments)[1];
+    assert(arg1.op == TOK.float64);
+    Expression arg2 = (*arguments)[2];
+    assert(arg2.op == TOK.float64);
+    return new RealExp(loc, CTFloat.fma(arg0.toReal(), arg1.toReal(), arg2.toReal()), arg0.type);
+}
+
 extern (C++) Expression eval_isnan(Loc loc, FuncDeclaration fd, Expressions* arguments)
 {
     Expression arg0 = (*arguments)[0];
@@ -226,7 +336,7 @@ public extern (C++) void builtin_init()
     add_builtin("_D4core4math4sqrtFNaNbNiNfeZe", &eval_sqrt);
     add_builtin("_D4core4math4fabsFNaNbNiNfeZe", &eval_fabs);
     add_builtin("_D4core4math5expm1FNaNbNiNfeZe", &eval_unimp);
-    add_builtin("_D4core4math4exp21FNaNbNiNfeZe", &eval_unimp);
+    add_builtin("_D4core4math4exp2FNaNbNiNfeZe", &eval_unimp);
     // @trusted @nogc pure nothrow real function(real)
     add_builtin("_D4core4math3sinFNaNbNiNeeZe", &eval_sin);
     add_builtin("_D4core4math3cosFNaNbNiNeeZe", &eval_cos);
@@ -234,7 +344,6 @@ public extern (C++) void builtin_init()
     add_builtin("_D4core4math4sqrtFNaNbNiNeeZe", &eval_sqrt);
     add_builtin("_D4core4math4fabsFNaNbNiNeeZe", &eval_fabs);
     add_builtin("_D4core4math5expm1FNaNbNiNeeZe", &eval_unimp);
-    add_builtin("_D4core4math4exp21FNaNbNiNeeZe", &eval_unimp);
     // @safe @nogc pure nothrow double function(double)
     add_builtin("_D4core4math4sqrtFNaNbNiNfdZd", &eval_sqrt);
     // @safe @nogc pure nothrow float function(float)
@@ -266,15 +375,14 @@ public extern (C++) void builtin_init()
     add_builtin("_D3std4math4sqrtFNaNbNiNfeZe", &eval_sqrt);
     add_builtin("_D3std4math4fabsFNaNbNiNfeZe", &eval_fabs);
     add_builtin("_D3std4math5expm1FNaNbNiNfeZe", &eval_unimp);
-    add_builtin("_D3std4math4exp21FNaNbNiNfeZe", &eval_unimp);
     // @trusted @nogc pure nothrow real function(real)
     add_builtin("_D3std4math3sinFNaNbNiNeeZe", &eval_sin);
     add_builtin("_D3std4math3cosFNaNbNiNeeZe", &eval_cos);
     add_builtin("_D3std4math3tanFNaNbNiNeeZe", &eval_tan);
     add_builtin("_D3std4math4sqrtFNaNbNiNeeZe", &eval_sqrt);
     add_builtin("_D3std4math4fabsFNaNbNiNeeZe", &eval_fabs);
-    add_builtin("_D3std4math5expm1FNaNbNiNeeZe", &eval_unimp);
-    add_builtin("_D3std4math4exp21FNaNbNiNeeZe", &eval_unimp);
+    add_builtin("_D3std4math5expm1FNaNbNiNeeZe", &eval_expm1);
+    add_builtin("_D3std4math4exp2FNaNbNiNeeZe", &eval_exp2);
     // @safe @nogc pure nothrow double function(double)
     add_builtin("_D3std4math4sqrtFNaNbNiNfdZd", &eval_sqrt);
     // @safe @nogc pure nothrow float function(float)
@@ -305,6 +413,39 @@ public extern (C++) void builtin_init()
     add_builtin("_D3std4math5ldexpFNaNbNiNfeiZe", &eval_ldexp);
     add_builtin("_D3std4math5ldexpFNaNbNiNfdiZd", &eval_ldexp);
     add_builtin("_D3std4math5ldexpFNaNbNiNffiZf", &eval_ldexp);
+
+    add_builtin("_D3std4math3logFNaNbNiNfeZe", &eval_log);
+
+    add_builtin("_D3std4math4log2FNaNbNiNfeZe", &eval_log2);
+
+    add_builtin("_D3std4math5log10FNaNbNiNfeZe", &eval_log10);
+
+    add_builtin("_D3std4math5roundFNbNiNeeZe", &eval_round);
+    add_builtin("_D3std4math5roundFNaNbNiNeeZe", &eval_round);
+
+    add_builtin("_D3std4math5floorFNaNbNiNefZf", &eval_floor);
+    add_builtin("_D3std4math5floorFNaNbNiNedZd", &eval_floor);
+    add_builtin("_D3std4math5floorFNaNbNiNeeZe", &eval_floor);
+
+    add_builtin("_D3std4math4ceilFNaNbNiNefZf", &eval_ceil);
+    add_builtin("_D3std4math4ceilFNaNbNiNedZd", &eval_ceil);
+    add_builtin("_D3std4math4ceilFNaNbNiNeeZe", &eval_ceil);
+
+    add_builtin("_D3std4math5truncFNaNbNiNeeZe", &eval_trunc);
+
+    add_builtin("_D3std4math4fminFNaNbNiNfeeZe", &eval_fmin);
+
+    add_builtin("_D3std4math4fmaxFNaNbNiNfeeZe", &eval_fmax);
+
+    add_builtin("_D3std4math__T8copysignTfTfZQoFNaNbNiNeffZf", &eval_copysign);
+    add_builtin("_D3std4math__T8copysignTdTdZQoFNaNbNiNeddZd", &eval_copysign);
+    add_builtin("_D3std4math__T8copysignTeTeZQoFNaNbNiNeeeZe", &eval_copysign);
+
+    add_builtin("_D3std4math__T3powTfTfZQjFNaNbNiNeffZf", &eval_pow);
+    add_builtin("_D3std4math__T3powTdTdZQjFNaNbNiNeddZd", &eval_pow);
+    add_builtin("_D3std4math__T3powTeTeZQjFNaNbNiNeeeZe", &eval_pow);
+
+    add_builtin("_D3std4math3fmaFNaNbNiNfeeeZe", &eval_fma);
 
     // @trusted @nogc pure nothrow bool function(T)
     add_builtin("_D3std4math__T5isNaNTeZQjFNaNbNiNeeZb", &eval_isnan);
--- a/src/dmd/root/ctfloat.d
+++ b/src/dmd/root/ctfloat.d
@@ -1,4 +1,4 @@
-/**
+/***
  * Compiler implementation of the
  * $(LINK2 http://www.dlang.org, D programming language).
  *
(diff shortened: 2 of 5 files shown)


ibuclaw:
Wouldn't this just add more discrepancies between compile time and runtime?

WalterBright:
These are only executed during CTFE, not optimized constant folding.

But certainly all the `std.math` functions need to be gone over again to ensure the ULPs are correct.

kinke:
[I circumvented this problem in LDC by using the D host compiler's `std.math`, as we've already been depending on Phobos.]

UplinkCoder:
more builtins .... sigh.
But what does fma do there, when infact it does not produce a fused-multiply-add.

test/compilable/test5227.d
```diff
@@ -0,0 +1,161 @@
...
+
+pragma(msg, "copysign()");
+enum csf = copysign(2.5f, -3.0f); pragma(msg, csf); static assert(csf == -2.5);
+enum csd = copysign(2.5 , -3.0 ); pragma(msg, csd); static assert(csd == -2.5);
+enum csr = copysign(2.5L, -3.0L); pragma(msg, csr); static assert(csr == -2.5);
+
+pragma(msg, "pow()");
+enum powf = pow(2.5f, -3.0f); pragma(msg, powf);
+enum powd = pow(2.5 , -3.0 ); pragma(msg, powd);
+enum powr = pow(2.5L, -3.0L); pragma(msg, powr);
```
TurkeyMan: There's no test here for the problem case. Please add a test for fractional exponent, eg: `pow(2, 3.5);`
That's the case that doesn't work in CTFE. Integer exponent seems to be handled as a different case.

TurkeyMan (changes requested):
I'd love to see a test for the problem case... although I'm sure it works ;)

TurkeyMan:
OMFG @WalterBright is my homeboy!

Yeah, even if you guys reckon this isn't the most sanitary solution, compile errors when you type 2^^1.5 has never been okay.

WalterBright:
> But what does fma do there, when infact it does not produce a fused-multiply-add.

Neither does the std.math one, which is why I suggested a review of std.math.

> I circumvented this problem in LDC by using the D host compiler's std.math

std.math may be better off in druntime. In any case, that's a separate issue. The ctfloat.d implementations are not terrible, they just rely on the C functions which aren't the best.

test/compilable/test5227.d
WalterBright: added

ibuclaw:
@TurkeyMan - you'll start complaining again when `2^^1.5` yields a different result at compile-time vs. run-time. ;-)

There are a couple of functions that should not be here in my opinion, I'll pick them out if they haven't been removed since I started writing.

src/dmd/builtin.d
```diff
@@ -103,6 +103,130 @@ extern (C++) Expression eval_ldexp(Loc loc, FuncDeclaration fd, Expressions* arg
...
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.exp2(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_nearbyint(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.nearbyint(arg0.toReal()), arg0.type);
```
ibuclaw: This makes no sense, because there is no rounding mode at compile time.

src/dmd/builtin.d
```diff
@@ -103,6 +103,130 @@ extern (C++) Expression eval_ldexp(Loc loc, FuncDeclaration fd, Expressions* arg
...
+}
+
+extern (C++) Expression eval_rint(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.rint(arg0.toReal()), arg0.type);
```
ibuclaw: Likewise having this as a built-in too.

ibuclaw (changes requested):

src/dmd/builtin.d
TurkeyMan: Can 'compile time' specify a static rounding mode? Presumably 'nearest', since it's the hardware default...?

TurkeyMan:
@ibuclaw And I'll be happy to complain when that comes up! Current situation is much worse that that hypothetical (albeit probable) moment.
None of the cases where I'm blocked on this will compare compile time results against runtime results that I'm aware of.
I need to build shit-loads of look-up tables.

TurkeyMan:
Incidentally; I'm used to building with fast-math and the like. I'm accustomed to CTFE not precisely matching runtime results ;)

wilzbach:
> @TurkeyMan - you'll start complaining again when 2^^1.5 yields a different result at compile-time vs. run-time. ;-)

Yes different behavior between CTFE and runtime is horrible to debug, because once you add `enum` somewhere, your tests fail (or worse: they still pass, but the small difference might add up in your production program.)
For many applications that's okay and they can opt-in with fast-math, but I fully agree with @ibuclaw that this shouldn't be the default.

WalterBright:
> Yes different behavior between CTFE and runtime is horrible to debug

There are *always* going to be slight differences between compile time and run time. Java early on tried to define away the difference, and they were forced to relent. If your algorithm cares about ULP, you *really* *really* need to pay close attention to how your expressions are constructed and compiled.

It isn't integer math and cannot be treated like integer math.

Anyhow, again, issues with accuracy are off topic for this PR.

JackStouffer:
This should get a changlog entry.

src/dmd/builtin.d
WalterBright: These two are there for compatibility with LDC, and also as a simple matter of convenience so the user doesn't have to use two paths for the same code.

src/dmd/builtin.d
```diff
@@ -306,6 +430,56 @@ public extern (C++) void builtin_init()
     add_builtin("_D3std4math5ldexpFNaNbNiNfdiZd", &eval_ldexp);
     add_builtin("_D3std4math5ldexpFNaNbNiNffiZf", &eval_ldexp);
 
+    add_builtin("_D3std4math3logFNaNbNiNfeZe", &eval_log);
+    add_builtin("_D3std4math3logFNaNbNiNfeZe", &eval_log);
+    add_builtin("_D3std4math3logFNaNbNiNfeZe", &eval_log);
```
rainers: You are adding the same function 3 times here. If the float and double overloads don't exist one builtin should be enough. Same for a number of other functions.

src/dmd/builtin.d
ibuclaw: Having a look at my end, no, gcc emulated floating point does not.

And hardware default for who? SuperH? Motorola 6800? Soft processors on an FPGA? :-)

ibuclaw: You can however convert the float to mpfr and explicitly request a rounding mode.

src/dmd/builtin.d
ibuclaw: Well I'd rather not add builtins that aren't common for all, as it means you will _have to_ use two paths anyway.  For example y2lx comes with version(INLINE_Y2LX).

test/compilable/test5227.d
```diff
@@ -0,0 +1,162 @@
...
+-12.84L
+---
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=5227
+
+import std.math;
+
+pragma(msg, "log()");
+enum logf = log(5.5f); pragma(msg, logf);
```
jacob-carlborg: How come `pragma(msg)` is used to test this instead of `static assert`?

src/dmd/builtin.d
```diff
@@ -226,15 +350,15 @@ public extern (C++) void builtin_init()
     add_builtin("_D4core4math4sqrtFNaNbNiNfeZe", &eval_sqrt);
     add_builtin("_D4core4math4fabsFNaNbNiNfeZe", &eval_fabs);
     add_builtin("_D4core4math5expm1FNaNbNiNfeZe", &eval_unimp);
-    add_builtin("_D4core4math4exp21FNaNbNiNfeZe", &eval_unimp);
+    add_builtin("_D4core4math4exp2FNaNbNiNfeZe", &eval_unimp);
```
kinke: `&eval_unimp` => `&eval_exp2`

src/dmd/builtin.d
```diff
@@ -226,15 +350,15 @@ public extern (C++) void builtin_init()
...
     // @trusted @nogc pure nothrow real function(real)
     add_builtin("_D4core4math3sinFNaNbNiNeeZe", &eval_sin);
     add_builtin("_D4core4math3cosFNaNbNiNeeZe", &eval_cos);
     add_builtin("_D4core4math3tanFNaNbNiNeeZe", &eval_tan);
     add_builtin("_D4core4math4sqrtFNaNbNiNeeZe", &eval_sqrt);
     add_builtin("_D4core4math4fabsFNaNbNiNeeZe", &eval_fabs);
     add_builtin("_D4core4math5expm1FNaNbNiNeeZe", &eval_unimp);
-    add_builtin("_D4core4math4exp21FNaNbNiNeeZe", &eval_unimp);
+    add_builtin("_D4core4math4exp2FNaNbNiNeeZe", &eval_unimp);
```
kinke: ditto

src/dmd/builtin.d
```diff
@@ -266,15 +390,15 @@ public extern (C++) void builtin_init()
     add_builtin("_D3std4math4sqrtFNaNbNiNfeZe", &eval_sqrt);
     add_builtin("_D3std4math4fabsFNaNbNiNfeZe", &eval_fabs);
     add_builtin("_D3std4math5expm1FNaNbNiNfeZe", &eval_unimp);
-    add_builtin("_D3std4math4exp21FNaNbNiNfeZe", &eval_unimp);
+    add_builtin("_D3std4math4exp2FNaNbNiNfeZe", &eval_unimp);
```
kinke: ditto

test/compilable/test5227.d
kinke: Right. The current tests also test that CTFE `log(float|double)` etc. return a `real`, i.e., that there are no float/double overloads, which I'm not sure is desirable.

WalterBright: The `std.math.log` function does not have overloads, and passing float/double to it will return a real. I don't think CTFE should do something different than the corresponding std.math function.

src/dmd/builtin.d
WalterBright: There isn't an `exp2` function in `core.math`. We could add to it, but that discussion is beyond the scope of this PR.

src/dmd/builtin.d
WalterBright: ditto too :-)

src/dmd/builtin.d
WalterBright: ditto tree

ibuclaw: Likewise, not ditto. :-)

test/compilable/test5227.d
kinke: > I don't think CTFE should do something different than the corresponding std.math function.

That isn't the question, it's whether the float/double tests should be there if there's no corresponding function. As soon as Phobos gets proper overloads, these DMD tests will fail, so DMD has to be patched in tandem. I'm in favor of adding corresponding CTFE tests after extending Phobos.

src/dmd/builtin.d
kinke: So you noticed the wrong mangle & fixed it instead of just removing that useless entry?

test/compilable/test5227.d
WalterBright: Put the functions in Phobos first, then in dmd. Not the other way around, because they cannot be tested if put in dmd first.

src/dmd/builtin.d
WalterBright: Yes. I don't know why the eval_unimp entries are there, I didn't write that code. But I felt it was beyond the scope of this PR to address that.

test/compilable/test5227.d
rainers: I'd also recommend skipping these tests for now: Testing them here to return real will block adding the overloads to phobos.

p0nce:
Just chiming in: I don't get most of implementation details, but precision across runtime and compile-time is not really a concern for us. I'm much more concerned from variations across bitness (typically 32-bit benefitting from 80-bit FPU operations but not 64-bit, happens once a year). If all compilers have the same values at CTFE, that's already great. That they would have the same values at runtime is already untrue for different compiler+arch combinations. 

So I don't get how runtime and compile-time values could ever match, since runtime values already don't match (examples on demand).

src/dmd/builtin.d
WalterBright: What's the issue with adding them to gdc?

src/dmd/builtin.d
rainers: std.math has annotated exp2 with `@trusted`, so this overload can go in favor of the one a few lines below.

src/dmd/builtin.d
thewilsonator: FWIW, the same goes for the `@trusted`/`@safe` overloads, the functions are all `@safe` IIRC.

WalterBright:
semaphoreci fails with no indication whatsoever of anything gone wrong.

wilzbach:
@WalterBright you still need to expand the failing tabs (the one with a  red X):

image

The log when building with LDC:

```
 ... fail_compilation/fail24.d       -fPIC ()
Test compilable/test5227.d failed.  The logged output:
../generated/linux/release/64/dmd -conf= -m64 -Icompilable  -fPIC  -odtest_results/compilable -oftest_results/compilable/test5227_0.o  -c compilable/test5227.d 
log()
1.70475L
1.70475L
1.70475L
log2()
2.45943L
2.45943L
2.45943L
log10()
0.740363L
0.740363L
0.740363L
nearbyint()
6.00000L
6.00000L
6.00000L
rint()
6.00000F
6.00000
6.00000L
round()
6.00000L
6.00000L
6.00000L
floor()
5.00000F
5.00000
5.00000L
ceil()
6.00000F
6.00000
6.00000L
trunc()
5.00000L
5.00000L
5.00000L
expm1()
243.692L
243.692L
243.692L
exp2()
45.2548L
45.2548L
45.2548L
fmin()
-3.2L
-3.2L
-3.2L
fmax()
5.2L
5.2L
5.2L
copysign()
-2.5F
-2.5
-2.5L
pow()
../../phobos/std/math.d(6884): Error: cannot evaluate unimplemented builtin `yl2x` at compile time
../../phobos/std/math.d(6897):        called from here: `impl(cast(real)x, cast(real)y)`
compilable/test5227.d(151):        called from here: `pow(2.5F, 2.5F)`
compilable/test5227.d(151):        while evaluating `pragma(msg, powf)`
../../phobos/std/math.d(6884): Error: cannot evaluate unimplemented builtin `yl2x` at compile time
../../phobos/std/math.d(6897):        called from here: `impl(cast(real)x, cast(real)y)`
compilable/test5227.d(152):        called from here: `pow(2.5, 2.5)`
compilable/test5227.d(152):        while evaluating `pragma(msg, powd)`
../../phobos/std/math.d(6884): Error: cannot evaluate unimplemented builtin `yl2x` at compile time
../../phobos/std/math.d(6897):        called from here: `impl(x, y)`
compilable/test5227.d(153):        called from here: `pow(2.5L, 2.5L)`
compilable/test5227.d(153):        while evaluating `pragma(msg, powr)`
../../phobos/std/math.d(6884): Error: cannot evaluate unimplemented builtin `yl2x` at compile time
../../phobos/std/math.d(6897):        called from here: `impl(cast(real)x, cast(real)y)`
compilable/test5227.d(154):        called from here: `pow(2.5, 2.5)`
compilable/test5227.d(154):        while evaluating `pragma(msg, powctfe)`
fma()
-12.84L
-12.84L
-12.84L


Test compilable/test5227.d failed: expected rc == 0, exited with rc == 1

make[1]: *** [test_results/compilable/test5227.d.out] Error 1
make: *** [start_compilable_tests] Error 2
```

kinke:
This is due to this: https://github.com/dlang/dmd/blob/728a18e8faec9d91b3b7ce1c3f86e405cb0aaa31/src/dmd/root/ctfloat.d#L42-L67

These bools determine whether the compiler being built will offer yl2x[p1] evaluation during CTFE. For this to work, the host compiler building DMD needs to support core.math.yl2x[p1]. Until recently, this was DMD-only; luckily, LDC supports them for x86[_64] targets too since v1.6 IIRC.
So the above `version(DigitalMars)` should probably be replaced by `static if (__traits(compiles, core.math.yl2x(1.0L, 2.0L)))` to fix the issue for the LDC 1.8 host compiler.

kinke:
The other way of looking at it is that it shouldn't even come that far (evaluating the pow() body at CTFE) - the mangle appears wrong/old, for `pow!(real, real)`:
`_D3std4math12__T3powTeTeZ3powFNaNbNiNeeeZe` => `_D3std4math__T3powTeTeZQjFNaNbNiNeeeZe`

test/compilable/test5227.d
kinke: > Put the functions in Phobos first, then in dmd. Not the other way around, because they cannot be tested if put in dmd first.

Absolutely, but this patch here puts them in dmd first, that's what I've been trying to tell the whole time.

src/dmd/builtin.d
ibuclaw: These are non existent builtins in gcc.

More generally though adding them as builtins would make them more susceptible to wrong code in the const folder/optimizer given.
```
setRoundingMode(down);
immutable x = rint(1.8);
setRoundingMode(up);
immutable y = rint(1.8);
```

WalterBright:
When I compile the program:
```
import std.math;
real test(real x, real y) { return pow(x, y); }
```
I get:
```
_D5test44testFeeZe	comdat
	assume	CS:_D5test44testFeeZe
L0:		push	dword ptr 018h[ESP]
		push	dword ptr 018h[ESP]
		push	dword ptr 018h[ESP]
		push	dword ptr 018h[ESP]
		push	dword ptr 018h[ESP]
		push	dword ptr 018h[ESP]
		call	near ptr _D3std4math__T3powTeTeZQjFNaNbNiNeeeZe
		ret	018h
```

WalterBright:
> So the above version(DigitalMars) should probably be replaced by static if (__traits(compiles, core.math.yl2x(1.0L, 2.0L))) to fix the issue with the LDC 1.8 host compiler.

@kinke @ibuclaw  now it fails semphoreci with GDC. What do you suggest, Iain?

src/dmd/root/ctfloat.d
```diff
@@ -39,28 +39,20 @@ private
...
-    {
-        static __gshared bool yl2x_supported = true;
-        static __gshared bool yl2xp1_supported = true;
-    }
-    else
-    {
-        static __gshared bool yl2x_supported = false;
-        static __gshared bool yl2xp1_supported = false;
-    }
+    enum yl2x_supported = __traits(compiles, core.math.yl2x(1.0L, 2.0L));
```
rainers: The declaration exists in core.math unconditionally, so it always compiles. For DMD (and LDC?), it is converted to an intrinsic. I think it should be excluded in builtin_init for appropriate versions. This would also produce better error messages for GDC instead of asserting here.

kinke: For DMD it's a builtin and always declared.
For LDC, it's a regular inline asm function, defined for x86[_64] targets using 80-bit `real` only, and only since a couple of versions.
For GDC, it's apparently just a declaration without implementation => linker error.

So one possibility is:
````D
version(GNU)
    enum yl2x_supported = false;
else
    enum yl2x_supported = __traits(compiles, core.math.yl2x(1.0L, 2.0L));
enum yl2xp1_supported = yl2x_supported; // most likely redundant
```

[This automatically makes sure the CTFloat functions aren't called, so no asserts for GDC.]

RazvanN7:
Tests are all green.  ping @ibuclaw

WalterBright:
@ibuclaw please respond

rainers:
I agree with Ian that `rint` and `nearbyint` don't make much sense without also supporting `fesetround` or similar. You can always use floor, ceil, trunc or round as explicit replacements.

Please also avoid testing the return type of fallbacks for non-existing functions, these block adding them to core/std.math. You could replace them with static asserts checking the value, but not the type.

WalterBright:
@rainers ok, done. Except I didn't do the static assert, it seemed pointless.

src/dmd/builtin.d
```diff
@@ -103,6 +103,116 @@ extern (C++) Expression eval_ldexp(Loc loc, FuncDeclaration fd, Expressions* arg
...
+    return new RealExp(loc, CTFloat.trunc(arg0.toReal()), arg0.type);
+}
+
+extern (C++) Expression eval_copysign(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    Expression arg1 = (*arguments)[1];
+    assert(arg1.op == TOK.float64);
+    return new RealExp(loc, CTFloat.copysign(arg0.toReal(), cast(int) arg1.toInteger()), arg0.type);
```
rainers: LDC fails to comple this line (no implicit conversion to real_t). Why do you cast to int here? copysign expects a real parameter.

rainers:
Apart from the failure LGTM

ibuclaw (approved):

src/dmd/builtin.d
ibuclaw: @rainers - Doesn't ldc implement their own builtins module independent of dmd?

ibuclaw:
I guess the build failure is for platforms where dmd uses `struct longdouble`.  Check and use explicit casts to `real_t`

src/dmd/builtin.d
rainers: I think so, but as noted, this is a usage of struct longdouble introduced by compilation with LDC. AFAICT passing `int` here is wrong anyway, should just be `arg1.toReal()`.

WalterBright:
@rainers I'm shooting in the dark here with appveyor. Now I'm getting the error:
```
..\dmd\root\ctfloat.d(88): error : cannot cast expression `roundl(cast(real)x.opCast())` of type `real` to `longdouble`
```
What do you advise I do to make this work?

rainers:
It needs a small part of https://github.com/dlang/dmd/pull/8169:
``` src/dmd/root/longdouble.d | 12 ++++++++++++
diff --git a/src/dmd/root/longdouble.d b/src/dmd/root/longdouble.d
index 4b3cedc..0ce7acd 100644
--- a/src/dmd/root/longdouble.d
+++ b/src/dmd/root/longdouble.d
@@ -45,6 +45,13 @@ nothrow:
         else
             ld_set(&this, d);
     }
+    this(real r)
+    {
+        static if (real.sizeof > 8)
+            *cast(real*)&this = r;
+        else
+            this(cast(double)r);
+    }
 
     void opAssign(float f) { ld_set(&this, f); }
     void opAssign(double f) { ld_set(&this, f); }
@@ -71,6 +78,11 @@ nothrow:
         else static if(is(T == double)) return cast(T)ld_read(&this);
         else static if(is(T == long))   return ld_readll(&this);
         else static if(is(T == ulong))  return ld_readull(&this);
+        else static if(is(T == real))
+        {
+            static if (real.sizeof > 8) return *cast(real*)&this;
+            else                        return cast(T)ld_read(&this);
+        }
         else static assert(false, "usupported type");
     }
```
Then this works:
```
        static real_t round(real_t x) { return real_t(core.stdc.math.roundl(cast(real)x)); }
        static real_t floor(real_t x) { return real_t(core.stdc.math.floor(cast(real)x)); }
        static real_t ceil(real_t x) { return real_t(core.stdc.math.ceil(cast(real)x)); }
        static real_t trunc(real_t x) { return real_t(core.stdc.math.trunc(cast(real)x)); }
        static real_t log(real_t x) { return real_t(core.stdc.math.logl(cast(real)x)); }
        static real_t log2(real_t x) { return real_t(core.stdc.math.log2l(cast(real)x)); }
        static real_t log10(real_t x) { return real_t(core.stdc.math.log10l(cast(real)x)); }
        static real_t pow(real_t x, real_t y) { return real_t(core.stdc.math.powl(cast(real)x, cast(real)y)); }
        static real_t expm1(real_t x) { return real_t(core.stdc.math.expm1l(cast(real)x)); }
        static real_t exp2(real_t x) { return real_t(core.stdc.math.exp2l(cast(real)x)); }
        static real_t copysign(real_t x, real_t s) { return real_t(core.stdc.math.copysignl(cast(real)x, cast(real)s)); }
```
and could even be used for the part `is(real_t == real)`.

WalterBright:
@rainers sigh, now I get this error with appveyor:
```
lib\druntime32mscoff.lib: Error: MS-Coff object module `errno_c_32mscoff.obj` has magic = 5ad5, should be 8664
lib\druntime32mscoff.lib: Error: MS-Coff object module `msvc_32mscoff.obj` has magic = 5ad5, should be 8664
lib\druntime32mscoff.lib: Error: MS-Coff object module `msvc_math_32mscoff.obj` has magic = 5ad5, should be 8664
```

rainers:
Sorry, atm I have no idea why the object files start with 0x5ad5. Locally the files have magic 0x14c (the message is still for x64 only) and building works fine.

Maybe just another spurious failure? You can restart the build by amending the same commit again.

MartinNowak:
> You can restart the build by amending the same commit again.

`git commit --allow-empty --amend` is handy for that purpose, basically just updates the commit date, make sure you haven't staged anything though.

MartinNowak:
Thanks for fixing this @WalterBright, missing log/exp support in CTFE has been a long-standing annoyance, in particular for precomputing scientific tables.

TurkeyMan:
Thanks @WalterBright! This is the most exciting patch in years! :P
Looking forward to the next release.

PR #8072, opened 2018-03-23
Fix Issue 18645 - DMD segmentation fault (Enum Initialization)

JinShil:


mihails-strasuns-sociomantic:
Thank you!

mihails-strasuns-sociomantic:
Tested on internal project - fixes the same problem there too.

jacob-carlborg (approved):

src/dmd/dsymbolsem.d
```diff
@@ -2039,7 +2039,10 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                 return errorReturn();
 
             Expression eprev = emprev.value;
-            Type tprev = eprev.type.equals(em.ed.type) ? em.ed.memtype : eprev.type;
+            Type tprev =
+                eprev.type.ty == em.ed.type.ty    // https://issues.dlang.org/show_bug.cgi?id=18645
```
WalterBright: How does this prevent a seg fault? Why is `equals` not working?

JinShil: It's trying to compare `immutable(A)` to `A`, causing `equals` to evaluate to false.  That eventually cases problems later in the semantic pass resulting in a segfault at `BinExp::setNoderefOperands`.  

When I compare the behavior of the compiler before the regression with the the current compiler, the change above is the precise point of digression.

WalterBright: Thanks, that make sense. But the proposed fix doesn't work if the type is not a leaf type. I propose instead to use:
```
eprev.type.toHeadMutable().equals(em.ed.type.toHeadMutable())
```
This technique is used elsewhere in the front end.

JinShil: Changed.  Thank you!

PR #8073, opened 2018-03-23
refactor: dmangle.d

WalterBright:
Refactor for clarity and improve performance:

* remove redundant tests
* remove redundant `strlen` (remember, `strlen/strcmp` must die)
* making helper functions `final` so they are called directly rather than indirectly
* use `const`
* use `foreach`

### DIFF
--- a/src/dmd/dmangle.d
+++ b/src/dmd/dmangle.d
@@ -275,14 +275,14 @@ public:
     final void mangleIdentifier(Identifier id, Dsymbol s)
     {
         if (!backrefIdentifier(id))
-            toBuffer(id.toChars(), s);
+            toBuffer(id.toString(), s);
     }
 
     ////////////////////////////////////////////////////////////////////////////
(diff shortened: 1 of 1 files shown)


src/dmd/dmangle.d
```diff
@@ -979,9 +988,9 @@ public:
             break;
         case 4:
             m = 'd';
-            for (size_t u = 0; u < e.len; u++)
+            foreach (u; 0 .. e.len)
```
wilzbach: I could automatically do this replacement for all the easy cases over the entire codebase if you want to.

wilzbach (approved):
LGTM

src/dmd/dmangle.d
WalterBright: Don't do it automatically. Some of those loops change the length of the loop in the loop body in unexpected ways - weird bugs result.

PR #8074, opened 2018-03-24
fix Issue 16189 - Optimizer bug, with simple test case

WalterBright:


test/runnable/mars1.d
```diff
@@ -1675,6 +1675,23 @@ void testdivcmp()
...
+void test16189()
+{
+    ubyte[9][1] data;
+    uint a = 0;
+  loop:
+    data[0] = data[a];
+    a--;
+    bool b = false;
+    if (b) goto loop;
+    assert(a == -1); // Fails with -O
```
jacob-carlborg: Should this comment still be here? I assume the assert does *not* fail now.

ibuclaw: Past tense "Failed"?  Doesn't bother me really.

ibuclaw (approved):
Walter's backend, trust he knows what he is doing.

PR #8076, opened 2018-03-24
Fix Issue 16317 - Wrong binop evaluation/load order when optimizing

LemonBoy:
Nothing to see here (for now), move along.
Let's see if 16317 is gone for good.

### DIFF
--- a/src/dmd/backend/cgelem.c
+++ b/src/dmd/backend/cgelem.c
@@ -3450,25 +3450,40 @@ STATIC elem * eleq(elem *e, goal_t goal)
         if (OTop(op2) && !el_sideeffect(e1)
             && op2 != OPdiv && op2 != OPmod
            )
-        {   tym_t ty;
+        {
+            elem *lhs, *rhs;
+
+            // Turn the expression into x = (x op e) if possible
+            if (el_match(e1, e2->E1))
+            {
+                lhs = e2->E1;
+                rhs = e2->E2;
+            }
+            else if (OTcommut(op2) && el_match(e1, e2->E2))
+            {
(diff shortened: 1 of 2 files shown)


WalterBright:
Please use the format: "fix Issue NNNN ..." for PR subjects that fix Issue NNNN. This triggers the dlangbot to do the right thing and connect it. It's also necessary to insert the link to this PR as a comment on the bugzilla issue. (I have already done it for this PR.) All this is necessary for cross referencing purposes.

WalterBright:
https://issues.dlang.org/show_bug.cgi?id=16317

wilzbach:
(FYI @LemonBoy: @dlang-bot looks at the commit message - not the title)

LemonBoy:
@WalterBright as you may have noticed there's no description of the problem nor of the solution implemented, there's no proper commit message but there's a suggestive "Nothing to see here (for now), move along." in the first message.
I only needed to run this commit trough the testing pipeline to see whether my assumptions about the code are correct or not, I'll follow up with a proper patch and commit message soon.

WalterBright:
> I only needed to run this commit trough the testing pipeline to see whether my assumptions about the code are correct or not, I'll follow up with a proper patch and commit message soon.

Thanks for the explanation (and the work!). In general, if a PR is not ready for review, put WIP in the title and we won't bother you until it's ready :-)

RazvanN7:
@LemonBoy what's the status on this?

RazvanN7:
I will close this for now as it appears that the purpose of it was to test some assumptions. @LemonBoy feel free to reopen if you will start working again on it.

PR #8080, opened 2018-03-25
Fix CLang 6.0 build errors

wilzbach:
(when building under Linux with `HOST_CXX=clang++`)

```
dmd/backend/go.c:123:18: error: case value evaluates to -1, which cannot be narrowed to type 'unsigned int' [-Wc++11-narrowing]
            case -1:                    /* not in flagtab[]     */
                 ^
  (CC)  BACK_OBJS  dmd/backend/dt.c
dmd/backend/go.c:157:18: error: case value evaluates to -1, which cannot be narrowed to type 'unsigned int' [-Wc++11-narrowing]
            case -1:                    /* not in flagtab[]     */
                 ^

dmd/backend/cod3.c:2707:22: error: case value evaluates to -1, which cannot be narrowed to type 'targ_size_t' (aka 'unsigned long') [-Wc++11-narrowing]
                case -1:
                     ^
```

I'm not so familiar with C++.
`flag` is `char`, but looking at the error message it seems that C++
does type widening automatically, so maybe `UINT_MAX` needs to be used
there too?

### DIFF
--- a/src/dmd/backend/cod3.c
+++ b/src/dmd/backend/cod3.c
@@ -2704,7 +2704,7 @@ L1:
                         goto L4;
                     genregs(cdb,0x31,reg,reg);
                     goto inc;
-                case -1:
+                case ~(targ_size_t)0:
                     if (I64)
                         goto L4;
                     genregs(cdb,0x31,reg,reg);
--- a/src/dmd/backend/go.c
+++ b/src/dmd/backend/go.c
@@ -65,7 +65,7 @@ void dbg_optprint(char *);
 
 int go_flag(char *cp)
 {   unsigned i;
-    unsigned flag;
+    int flag;
 
     enum GL     // indices of various flags in flagtab[]
     {


src/dmd/backend/cod3.c
```diff
@@ -2704,7 +2705,7 @@ void movregconst(CodeBuilder& cdb,unsigned reg,targ_size_t value,regm_t flags)
                         goto L4;
                     genregs(cdb,0x31,reg,reg);
                     goto inc;
-                case -1:
+                case UINT_MAX:
```
WalterBright: Try `~(targ_size_t)0` instead.

src/dmd/backend/cod3.c
```diff
@@ -12,6 +12,7 @@
 
 #if !SPP
 
+#include        <climits>
```
WalterBright: Pleased don't use the `.h`-less versions. `limits.h` will do fine. Also, it's unnecessary due to below suggestion.

src/dmd/backend/go.c
```diff
@@ -120,7 +120,7 @@ int go_flag(char *cp)
             case GLspace:
                 go.mfoptim |= MFtime;      /* inverse of -time     */
                 break;
-            case -1:                    /* not in flagtab[]     */
+            case 255:                    /* not in flagtab[]     */
```
WalterBright: Better to change the declaration of `flag` from `unsigned` to `int`.

src/dmd/backend/cod3.c
wilzbach: Cool! Done

src/dmd/backend/go.c
wilzbach: Cool! Done

PR #8082, opened 2018-03-25
Fix Issue 18197 - Correct optimization for OPpair in x87 mode

LemonBoy:
Attempting to push a non-fp value into mST0 results in a ICE.

The patch assumes `OPpair` is homogeneous, I hope that invariant holds (and should probably slip an assertion just to be sure).

### DIFF
--- a/src/dmd/backend/cgelem.c
+++ b/src/dmd/backend/cgelem.c
@@ -3579,6 +3579,7 @@ STATIC elem * eleq(elem *e, goal_t goal)
             !el_appears(e2, e1->EV.sp.Vsym) &&
             // Disable this rewrite if we're using x87 and `e1` is a FP-value
             // but `e2` is not, or vice versa
+            // https://issues.dlang.org/show_bug.cgi?id=18197
             (config.fpxmmregs ||
              (tyfloating(e2->E1->Ety) != 0) == (tyfloating(e2->Ety) != 0))
            )
--- a/src/dmd/backend/cgelem.c
+++ b/src/dmd/backend/cgelem.c
@@ -3578,9 +3578,9 @@ STATIC elem * eleq(elem *e, goal_t goal)
             goal == GOALnone &&
             !el_appears(e2, e1->EV.sp.Vsym) &&
             // Disable this rewrite if we're using x87 and `e1` is a FP-value
-            // but `e2` is not
+            // but `e2` is not, or vice versa
             (config.fpxmmregs ||
-             tyfloating(e2->E1->Ety) != 0 == tyfloating(e2->Ety) != 0)
+             (tyfloating(e2->E1->Ety) != 0) == (tyfloating(e2->Ety) != 0))
            )
         {
             // printf("** before:\n"); elem_print(e); printf("\n");
--- a/src/dmd/backend/cgelem.c
+++ b/src/dmd/backend/cgelem.c
@@ -3576,10 +3576,14 @@ STATIC elem * eleq(elem *e, goal_t goal)
             e1->Eoper == OPvar &&
             (e2->Eoper == OPpair || e2->Eoper == OPrpair) &&
             goal == GOALnone &&
-            !el_appears(e2, e1->EV.sp.Vsym)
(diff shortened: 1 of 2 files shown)


ibuclaw:
I wonder why dlang bot hasn't picked this up. @LemonBoy spell Issue with a capital I in the commit?

ibuclaw (commented):
Seems reasonable, but ping @WalterBright though.

https://issues.dlang.org/show_bug.cgi?id=18197

wilzbach:
> I wonder why dlang bot hasn't picked this up. @LemonBoy spell Issue with a capital I in the commit?

This first time when dlang-bot was called it got a 404 from the GH API:

```
GET https://api.github.com/repos/dlang/dmd/pulls/8082/commits failed; Not Found 404.
```

Not sure why though. It sometimes does run into the rate-limit, but then it gets a 403.
(Triggering an action, e.g. by changing the title is a workaround in these cases.)

src/dmd/backend/cgelem.c
```diff
@@ -3576,10 +3576,13 @@ STATIC elem * eleq(elem *e, goal_t goal)
             e1->Eoper == OPvar &&
             (e2->Eoper == OPpair || e2->Eoper == OPrpair) &&
             goal == GOALnone &&
-            !el_appears(e2, e1->EV.sp.Vsym)
+            !el_appears(e2, e1->EV.sp.Vsym) &&
+            // Disable this rewrite if we're using x87 and the OPpair isn't made
+            // of FP values
+            (config.fpxmmregs || tyfloating(e2->E1->Ety) == tyfloating(e2->Ety))
```
WalterBright: tyfloating doesn't return a bool:
```
inline unsigned tyfloating(tym_t ty) { return tytab[ty & 0xFF] & (TYFLreal | TYFLimaginary | TYFLcomplex); }
```
so the test should be `tyfloating() != 0`

src/dmd/backend/cgelem.c
```diff
@@ -3576,10 +3576,13 @@ STATIC elem * eleq(elem *e, goal_t goal)
             e1->Eoper == OPvar &&
             (e2->Eoper == OPpair || e2->Eoper == OPrpair) &&
             goal == GOALnone &&
-            !el_appears(e2, e1->EV.sp.Vsym)
+            !el_appears(e2, e1->EV.sp.Vsym) &&
+            // Disable this rewrite if we're using x87 and the OPpair isn't made
+            // of FP values
```
WalterBright: I think you mean OPpair isn't made of mixed FP and non-FP values?

src/dmd/backend/cgelem.c
LemonBoy: Fixed.

src/dmd/backend/cgelem.c
LemonBoy: I reworded the comment a bit, I hope it's clearer now.

src/dmd/backend/cgelem.c
```diff
@@ -3576,10 +3576,14 @@ STATIC elem * eleq(elem *e, goal_t goal)
             e1->Eoper == OPvar &&
             (e2->Eoper == OPpair || e2->Eoper == OPrpair) &&
             goal == GOALnone &&
-            !el_appears(e2, e1->EV.sp.Vsym)
+            !el_appears(e2, e1->EV.sp.Vsym) &&
+            // Disable this rewrite if we're using x87 and `e1` is a FP-value
+            // but `e2` is not
```
WalterBright: or vice versa

src/dmd/backend/cgelem.c
```diff
@@ -3576,10 +3576,14 @@ STATIC elem * eleq(elem *e, goal_t goal)
...
+            (config.fpxmmregs ||
+             tyfloating(e2->E1->Ety) != 0 == tyfloating(e2->Ety) != 0)
```
WalterBright: Need parens here:
```
(tyfloating(e2->E1->Ety) != 0) == (tyfloating(e2->Ety) != 0)
```

PR #8083, opened 2018-03-26
Add postscript.sh run environment

wilzbach:
Basically just removing superfluous `|| exit 1` as now `set -euo pipefail` is the default for these postscripts too.

CC @marler8997


jacob-carlborg (approved):

marler8997 (approved):

braddr:
Broke win 64 bit builds due to nearly pointless file movement.  I've fixed up the windows auto-tester diffs to accommodate, but really, bad prioritization of effort and deflection of blame that resulted from the pull.

PR #8084, opened 2018-03-26
test18367: use proper library extension for Windows

rainers:
Please also note that the failing test doesn't stop the auto-tester from reporting success! Not sure when this has been introduced.

### DIFF
--- a/test/compilable/test18367.sh
+++ b/test/compilable/test18367.sh
@@ -2,7 +2,7 @@
 
 
 # dmd should not segfault on -X with libraries, but no source files
-out=$(! "$DMD" -conf= -X foo.a 2>&1)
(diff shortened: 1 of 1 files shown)


wilzbach:
Introduced in https://github.com/dlang/dmd/commit/7c0999581c1abb1f82874c67a06c15c23cdd41e0

wilzbach (approved):
That auto-tester doesn't stop if a tests fails on Windows seems like a serious flaw. It does so on the other platforms - do you have an idea why it doesn't on Windows? Is this related to DigitalMars make?

rainers:
> That auto-tester doesn't stop if a tests fails on Windows seems like a serious flaw. It does so on the other platforms - do you have an idea why it doesn't on Windows? Is this related to DigitalMars make?

https://github.com/dlang/dmd/pull/8085

PR #8087, opened 2018-03-26
Rename potentially conflicting output files + small bash cleanup

wilzbach:
- Make output_file + script_name "private" (to avoid any accidental dependence on them)
- Rename potentially conflicting output files (it's not a good idea to meddle with the actual output file)
- Small bash cleanup (removed a few `|| exit 1`)


RazvanN7 (approved):
Squash and we're good to go

jacob-carlborg (approved):
I think the commits look fine. They're separate and doing different things.

PR #8089, opened 2018-03-26
Allow passing additional DFLAGS to the DMD build

wilzbach:
There are many predefined ways to build DMD, but AFAIK one can't easily
pass in its own `DFLAGS` without modifying the Makefile.
With this change the following is now possible:

```
make -f posix.mak DFLAGS="-version=FooBar"
```

Learn more about `override`: https://www.gnu.org/software/make/manual/make.html#Override-Directive

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -221,9 +221,11 @@ ifeq ($(CXX_KIND), clang++)
 CXXFLAGS += \
     -xc++
 endif
-DFLAGS := -version=MARS $(PIC)
+
+DFLAGS=
+override DFLAGS += -version=MARS $(PIC)
 # Enable D warnings
-DFLAGS += -w -de
+override DFLAGS += -w -de
 
 ifneq (,$(DEBUG))
 ENABLE_DEBUG := 1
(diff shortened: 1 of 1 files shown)


wilzbach:
Hmm, environment variables are respected too and I think we don't want a global DFLAGS setting to affect the DMD build, so it looks like it's not as simple as this :/

wilzbach:
So looking at the manual there's a good solution:

> Variables in make can come from the environment in which make is run. Every environment variable that make sees when it starts up is transformed into a make variable with the same name and value. However, an explicit assignment in the makefile, or with a command argument, overrides the environment.

https://www.gnu.org/software/make/manual/html_node/Environment.html#Environment

which means that:

```make
DFLAGS=
override DFLAGS += -version=MARS $(PIC)
```

will ignore environment variables, but allow user-defined command line arguments:

```
> pmake -j20 DFLAGS="-version=FooBar"
CC="c++" dmd -of../generated/linux/release/64/dmd -m64 -vtls -J../generated/linux/release/64 -J../res -L-lstdc++ -version=FooBar -version=MARS -fPIC -w -de
```

```
> DFLAGS="-version=FooBar" pmake -j20
CC="c++" dmd -of../generated/linux/release/64/dmd -m64 -vtls -J../generated/linux/release/64 -J../res -L-lstdc++  -version=MARS -fPIC -w -de 
```

PR #8090, opened 2018-03-26
Fix Issue 18578 - First enum value assigned 0 instead of EnumBaseType.init

RazvanN7:


marler8997:
Also this: https://github.com/dlang/dmd/pull/7996

src/dmd/dsymbolsem.d
```diff
@@ -2232,8 +2232,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                 if (!em.ed.isAnonymous())
                     em.ed.memtype = t;
             }
-            Expression e = new IntegerExp(em.loc, 0, Type.tint32);
-            e = e.implicitCastTo(sc, t);
+            Expression e = new IntegerExp(em.loc, 0, t);
```
marler8997: Not too familair with the AST, but @LemonBoy's change "looks" more correct:
```D
Expression e = t.defaultInitLiteral(em.loc);
```

RazvanN7: That was my first option also, but for some reason it led to segfaults when running the tests. From my point of view, enums are an enumeration of constants that are represented by integers behind the scenes, so this solution looks correct (notice that the cast is not needed anymore)

WalterBright: The trouble is there is significant code depending on it being 0. A deprecation cycle will be needed. @lemonboy 's solution is more correct, but your test case is better. Both of you fixing the same bug is a pity, as now I have to close one. :-(

WalterBright:
Better yet, I'll just approve this one, and regard @LemonBoy 's fix as an enhancement.

src/dmd/dsymbolsem.d
marler8997: Yeah merging this one first seems reasonable.

@RazvanN7 The language spec doesn't say anything about enums being "integers behind the scenes" (https://dlang.org/spec/enum.html). In fact, if enums had this restriction you couldn't do things like this:
```D
enum message = "MyMessage"; // not an integer behind the scenes
```

Making the default initialization always an integer is an example of breaking "turtles all the way down".

PR #8091, opened 2018-03-26
Use --retry for AUTO_BOOTSTRAP=1

wilzbach:
Make `AUTO_BOOTSTRAP=1` more resilient to temporary network errors.
This is useful when used in CI scripts, e.g. the Jenkins project tester uses `AUTO_BOOTSTRAP=1` to install DMD:

https://github.com/dlang/ci/blob/8fc7ddd221b96fbe715b4c0f9775a7ff9e8cbbdd/vars/runPipeline.groovy#L375

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -486,12 +486,13 @@ clean:
 ######## Download and install the last dmd buildable without dmd
 
 ifneq (,$(AUTO_BOOTSTRAP))
+CURL_FLAGS:=-fsSL --retry 5 --retry-max-time 120 --connect-timeout 5 --speed-time 30 --speed-limit 1024
 $(HOST_DMD_PATH):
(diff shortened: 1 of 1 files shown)


src/posix.mak
```diff
@@ -486,12 +486,13 @@ clean:
 ######## Download and install the last dmd buildable without dmd
 
 ifneq (,$(AUTO_BOOTSTRAP))
+CURL_FLAGS:=-fsSL --retry 5 --retry-max-time 120 --connect-timeout 5 --speed-time 30 --speed-limit 1024
```
wilzbach: A short explanation:
Retry max 5 times or max. for 120 seconds. The last three arguments are to prevent stalled connects or infinitesimal small download speed (the speed is measured in byes/second)

PR #8095, opened 2018-03-28
Fix Issue 18670 - compiler segfault if new on a union type with dip1000

RazvanN7:


wilzbach (approved):

wilzbach:
Hmm SemaphoreCi seems to have troubles with this:

```
std/path.d(1740): Error: scope variable path assigned to chained with longer lifetime
std/file.d(3003): Error: template instance `std.path.buildNormalizedPath!char` error instantiating
```

(I restarted the build)

test/compilable/test18670.d
```diff
@@ -0,0 +1,12 @@
+// REQUIRED_ARGS: -dip1000
+
+// https://issues.dlang.org/show_bug.cgi?id=18670
+
+void foo() {
```
mathias-lang-sociomantic: Missing `@safe`

wilzbach: The ICE happened without @safe and I thought the general policy is to copy/paste the submitted samples to the testsuite?

RazvanN7:
@wilzbach I have no idea why that error might occur. The code changes affect compilation only if -dip1000 is used and all the check does is avoid calling a method on an empty object.

wilzbach:
> The only thing that comes to mind is a situation in which the patch is applied on a different version of dmd/druntime/phobos then the one I am using.

Yep, SemaphoreCi seems to checkout `master` - I think https://github.com/dlang/dmd/pull/8097 should (hopefully) fix this.

test/compilable/test18670.d
mathias-lang-sociomantic: Damn, another case were DIP1000 is not doing what it's supposed to do (nothing DIP1000 related should be checked outside of `@safe` code).

PR #8096, opened 2018-03-28
Fix issue 17918 - Stop if the late semantic pass fails

LemonBoy:
No matter what the outcome of finishVtbl was the old code kept building
the rest of the class. If the late semantic pass over the
FunctionDeclaration fails we should abort the codegen phase instead of
going on.

### DIFF
--- a/src/dmd/toobj.d
+++ b/src/dmd/toobj.d
@@ -1379,8 +1379,11 @@ void toObjFile(Dsymbol ds, bool multiobj)
 
 
 /*********************************
- * Finish semantic analysis of functions in vtbl[],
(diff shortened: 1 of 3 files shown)


src/dmd/toobj.d
```diff
@@ -1374,10 +1380,12 @@ void toObjFile(Dsymbol ds, bool multiobj)
 
 /*********************************
  * Finish semantic analysis of functions in vtbl[],
- * check vtbl[] for errors.
+ * check vtbl[] for errors and returns false if it does.
```
WalterBright: Use standard Ddoc `Returns:` form.

PR #8097, opened 2018-03-28
SemaphoreCi: fix BRANCH detection

wilzbach:
Apparentely is SemaphoreCI checking out the wrong branch as $BRANCH_NAME doesn't refer to the base branch, but the PR branch.

This dirty `git` hack should be able to detect that the commit is based on `stable` without needing to query the GitHub API.

### DIFF
--- a/semaphoreci.sh
+++ b/semaphoreci.sh
@@ -16,8 +16,20 @@ export DMD=${DMD:-dmd}         # can be {dmd,ldc,gdc}
 
 export N=4
 export OS_NAME=linux
-export BRANCH=$BRANCH_NAME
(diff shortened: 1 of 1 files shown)


wilzbach:
Looks like the `git describe -all` tricks works:

```
++ git describe --all
++ sed -E 's/.*\/([^-/]*)-.*/\1/'
+ BRANCH=stable
...
+ local branch=stable
+ for i in '{0..4}'
+ git clone --depth=1 --branch stable https://github.com/dlang/phobos.git ../phobos --quiet
```

RazvanN7 (approved):

andralex (approved):
great thanks for moving fast

PR #8098, opened 2018-03-28
Fix Issue 18474 - Postblit not working in shared structs

RazvanN7:


andralex (approved):
thx! cc @WalterBright

PR #8100, opened 2018-03-29
Fix Issue 18688 - Constructors shouldn't have implicit super call if it throws

RazvanN7:
The problem with this is that if an `assert(false);` is encountered the scope of the function in which the assert is located is marked as `halt` no matter where the assert is located. Let's take for example the bug report:

```d
class A {
    this(int x){}
    @disable this();
}
class B: A {
    this(int x) {
        super(x);
    }
    this(string b) {
        switch(b) {
            case "a":break;
            default: assert(false);
        }
        this(1);
    }
}
``` 

When `this(1)` is semantically analyzed, the scope is not updated to reflect that a constructor call is issued because it is considered that the line is not reachable, due to the `assert(false)`. This is obviously wrong since the code is reachable in some situations. The fix makes it so that the scope is updated to reflect that a this/super call is issued even if inside a halt scope.

### DIFF
--- a/test/compilable/test18688.d
+++ b/test/compilable/test18688.d
@@ -1,0 +1,43 @@
+// https://issues.dlang.org/show_bug.cgi?id=18688
+
+class A
+{
+    this(int x){}
+    @disable this();
+}
+
+class B: A
+{
+    this(int x)
+    {
+        super(x);
+    }
+
+    this(string b)
+    {
+        switch(b)
+        {
+            case "a":break;
+            default: assert(false);
+        }
+        this(1);
+    }
+}
+
+class C: A
+{
+    this(int x)
+    {
+        super(x);
+    }
+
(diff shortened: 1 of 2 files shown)


test/compilable/test18688.d
```diff
@@ -0,0 +1,43 @@
...
+        switch(b)
+        {
+            case "a":break;
+            default: assert(false);
+        }
+        super(1);
+    }
+}
+
+void main() {}
```
WalterBright: Not needed for compile-only tests

RazvanN7:
@WalterBright Done. Thank you!

RazvanN7:
jenkins is failing, but when inspecting the problem looks like all the tests pass. I find the matter puzzling...

wilzbach:

PR #8101, opened 2018-03-29
Fix Issue 18694 - pure functions using static/global structs with no fields should pass compilation

RazvanN7:


andralex:
Thanks!

andralex (approved):

ibuclaw:
This PR has caused a cascade of regressions upon regressions.
https://issues.dlang.org/show_bug.cgi?id=21464
https://issues.dlang.org/show_bug.cgi?id=23589

PR #8102, opened 2018-03-30
Put DMD's binary in 'generated' instead of TMP when using AUTO_BOOTSTRAP=1

wilzbach:
Motivation:

```
/bin/sh: 1: /tmp/.host_dmd-2.072.2/dmd2/linux/bin64/dmd: Text file busy
```

In CIs with multiple workers on the same machine.

e.g. https://ci.dlang.io/blue/organizations/jenkins/dlang-org%2Fphobos/detail/PR-6384/2/pipeline

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -137,7 +137,7 @@ else
   # Auto-bootstrapping, will download dmd automatically
   # Keep var below in sync with other occurrences of that variable, e.g. in circleci.sh
   HOST_DMD_VER=2.072.2
-  HOST_DMD_ROOT=$(TMP)/.host_dmd-$(HOST_DMD_VER)
+  HOST_DMD_ROOT=$(GENERATED)/host_dmd-$(HOST_DMD_VER)
   # dmd.2.072.2.osx.zip or dmd.2.072.2.linux.tar.xz
   HOST_DMD_BASENAME=dmd.$(HOST_DMD_VER).$(OS)$(if $(filter $(OS),freebsd),-$(MODEL),)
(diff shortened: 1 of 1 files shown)


src/posix.mak
```diff
@@ -137,7 +137,7 @@ else
   # Auto-bootstrapping, will download dmd automatically
   # Keep var below in sync with other occurrences of that variable, e.g. in circleci.sh
   HOST_DMD_VER=2.072.2
-  HOST_DMD_ROOT=$(TMP)/.host_dmd-$(HOST_DMD_VER)
+  HOST_DMD_ROOT=$(GENERATED)/host_dmd-$(HOST_DMD_VER)
```
RazvanN7: `make clean` deletes the directory. This means that every time you do it you will have to download the bootstrapping version. I suggest you choose a different directory (or create one) so that the compiler is downloaded only once.

wilzbach: I'm aware, but
- remember that you don't need to call `clean` because `make` is smart enough
- `$(TMP)` gets nuked too
- I don't think anyone developing on the compiler doesn't have a DMD compiler installed
- If I create a new top-level directory, people will complain too
- If I change `clean` to only nuke `generated/$(OS)` people will complain too
- we do this already for `dlang.org` ([source](https://github.com/dlang/dlang.org/blob/master/posix.mak#L185))

That being said, I honestly don't care which path is taken except that it can't be one outside of this directory as this will lead to CI errors when running in parallel.

RazvanN7: Fair enough

wilzbach:

PR #8103, opened 2018-03-30
 Postscripts: export common makefile variables + use them

wilzbach:
Part 2 of https://github.com/dlang/dmd/pull/8083

This exports the same variables as the bash test files already use to the environment and then in the second commit makes use of them.

The removed files like `sieve-postscript.sh`  have been generalized to e.g `coverage-postscript.sh`

CC @marler8997


wilzbach:
Windows failure is unrelated - see https://github.com/braddr/d-tester/issues/71

PR #8107, opened 2018-04-01
refactor: split specialFunctions() out of FuncDeclaration_toObjFile()

WalterBright:
A copy-paste job because FuncDeclaration_toObjFile() is unreasonably long.

### DIFF
--- a/src/dmd/glue.d
+++ b/src/dmd/glue.d
@@ -880,96 +880,7 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
     }
     else
     {
-        const(char)* libname = (global.params.symdebug)
-                                ? global.params.debuglibname
-                                : global.params.defaultlibname;
-
-        // Pull in RTL startup code (but only once)
-        if (fd.isMain() && onlyOneMain(fd.loc))
-        {
-            if (global.params.isLinux || global.params.isOSX || global.params.isFreeBSD ||
-                global.params.isOpenBSD || global.params.isDragonFlyBSD || global.params.isSolaris)
-            {
-                objmod.external_def("_main");
-            }
-            else if (global.params.mscoff)
-            {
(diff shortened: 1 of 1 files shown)


src/dmd/glue.d
```diff
@@ -1344,6 +1256,112 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
...
+        }
+        if (libname)
+            objmod.includelib(libname);
+        s.Sclass = SCglobal;
+    }
+    else if (fd.isRtInit())
+    {
+        if (global.params.isLinux || global.params.isOSX || global.params.isFreeBSD ||
+            global.params.isOpenBSD || global.params.isDragonFlyBSD || global.params.isSolaris ||
+            global.params.mscoff)
```
wilzbach: I know that this is just a split-out refactoring, but wouldn't `globals.useExceptions` be better here?

wilzbach (approved):

jacob-carlborg (approved):

src/dmd/glue.d
```diff
@@ -1344,6 +1256,112 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
...
 
 
+/*******************************************
+ * Detect special functions like `main()` and do special handling for them,
+ * like special mangling, including libraries, setting the storage class, etc.
+ * `objmod` and `fd` are updated.
+ *
+ * Params:
+ *      objmod = object module
+ *      fd = function symbol
```
jacob-carlborg: I would prefer to see some more documentation of the parameters.

src/dmd/glue.d
WalterBright: Perhaps, but this PR is much easier to review as a straight cut&paste job. The minute the pasted code starts getting rewritten, too, makes it much much harder.

PR #8109, opened 2018-04-01
refactor: use CSX instead of uint

WalterBright:
which will reduce the brittleness of the code. Also refactored `mergeFieldInit()` a bit to reduce unnecessary dependencies.

### DIFF
--- a/src/dmd/dscope.d
+++ b/src/dmd/dscope.d
@@ -36,7 +36,15 @@ import dmd.tokens;
 
 //version=LOGSEARCH;
 
-private bool mergeFieldInit(ref uint fieldInit, uint fi, bool mustInit)
+/****************************************
+ * Merge fi flow analysis results into fieldInit.
+ * Params:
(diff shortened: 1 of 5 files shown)


src/dmd/dscope.d
```diff
@@ -54,36 +62,37 @@ private bool mergeFieldInit(ref uint fieldInit, uint fi, bool mustInit)
         }
         else if (!aHalt && aRet)
         {
-            ok = !mustInit || (fi & CSX.this_ctor);
+            ok = (fi & CSX.this_ctor);
```
wilzbach: The parenthesis are now redundant (though they were "redundant" even before).

wilzbach (approved):
FYI: Due to the unrelated Jenkins failure (https://github.com/dlang/ci/issues/190), this will need to be manually merged once the other CIs pass. Still giving it auto-merge for higher priority on the auto-tester.

PR #8110, opened 2018-04-02
add ctorflow.d - begin encapsulation of flow analysis of constructors

WalterBright:
The logic of this is spread out over 5 files or so. This begins to pull it into one file, ctorflow.d. I haven't changed any logic in this, just moved code around and combined `superCall` and `fieldinit` into a struct.

### DIFF
--- a/src/dmd/ctorflow.d
+++ b/src/dmd/ctorflow.d
@@ -1,0 +1,122 @@
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Manage flow analysis for constructors.
+ *
+ * Copyright:   Copyright (C) 1999-2018 by The D Language Foundation, All Rights Reserved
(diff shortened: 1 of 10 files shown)


src/dmd/ctorflow.d
```diff
@@ -0,0 +1,122 @@
...
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/ctorflow.d
+ */
+
+module dmd.ctorflow;
+
+import core.stdc.stdio;
+
+import dmd.root.rmem;
+
+enum CSX : ubyte
```
wilzbach: If you want this to be visible on the docs, you will need to put `///` on it.

wilzbach (approved):
Nice! Modularization for the win!

wilzbach:
SemaphoreCI failure is due the newly added divcoeff.d - fix: https://github.com/dlang/dmd/pull/8111

PR #7714, opened 2018-01-15
convert divcoeff.c to D rebooted

WalterBright:
Reboot of https://github.com/dlang/dmd/pull/6907 because `git rebase` simply would not work with files that were renamed.

PR #8111, opened 2018-04-02
Fix GDC build on SemaphoreCI

wilzbach:
Follow-up to https://github.com/dlang/dmd/pull/7714

For some reason SemaphoreCI didn't want to run on this PR and it turns out that `gdmd` doesn't support `-lib`.

See e.g. https://semaphoreci.com/dlang/dmd-2/branches/master/builds/391

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -452,8 +452,7 @@ $G/glue.a: $(G_GLUE_OBJS) $(SRC_MAKE)
 	$(AR) rcs $@ $(G_GLUE_OBJS)
 
 $G/backend.a: $(G_OBJS) $(G_DOBJS) $(SRC_MAKE)
-	$(HOST_DMD_RUN) -lib -of$G/backend.a $(G_OBJS)
-	$(HOST_DMD_RUN) -lib -of$G/backend.a $@ $(G_DOBJS)
+	$(AR) rcs $@ $(G_OBJS) $(G_DOBJS)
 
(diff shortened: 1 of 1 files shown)


WalterBright (approved):

wilzbach:
FreeBSD failure looks unrelated -> https://issues.dlang.org/show_bug.cgi?id=18704

ibuclaw:
> it turns out that gdmd doesn't support -lib.

It does support -lib, so maybe you could be a bit more specific or have a look at the - vdmd output of the script and fix whatever is wrong. ;-)

wilzbach:
This is the failing line for `gdmd`:

> /home/seb/dlang/gdc-4.8.5/bin/gdmd -lib -of../generated/linux/release/64/backend.a abc.o

> or have a look at the -vdmd output of the script and fix whatever is wrong

Sure - https://github.com/D-Programming-GDC/GDMD/pull/5

Though we should still merge this PR here as the gdc release we use here (aka the "latest") is from 2016.

ibuclaw:
The gdmd wrapper is only superficial, and currently the ci pulls the latest copy anyway.

wilzbach:
> The gdmd wrapper is only superficial, and currently the ci pulls the latest copy anyway.

It's locked to a specific commit:

https://github.com/dlang/installer/blob/2e3086a02305637d119e252e539e47f365228ec0/script/install.sh#L554

So an update is needed -> https://github.com/dlang/installer/pull/316

While I could merge it at the installer, I don't have the permissions to deploy the installer script.
So I think we should still go with `$(AR)` for now ;-)

PR #8113, opened 2018-04-02
refactor: more use of ctorflow

WalterBright:
Move more implementation details of `CtorFlow` into `ctorflow.d`


andralex (approved):
sweet

wilzbach:
FYI: auto-merge won't work if one CI fails and Jenkins is currently "permanently" failing - https://github.com/dlang/ci/issues/190

PR #8116, opened 2018-04-02
expression.d: Use foreach

wilzbach:
~~~While this is straight-forward,~~~ I have to admit that it can't be automatized (easily).

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -535,9 +535,8 @@ extern (C++) Expression resolvePropertiesOnly(Scope* sc, Expression e1)
         os = (cast(OverExp)e1).vars;
     Los:
         assert(os);
-        for (size_t i = 0; i < os.a.dim; i++)
+        foreach (s; os.a)
         {
-            Dsymbol s = os.a[i];
             fd = s.isFuncDeclaration();
             td = s.isTemplateDeclaration();
             if (fd)
@@ -932,16 +931,15 @@ extern (C++) bool arrayExpressionSemantic(Expressions* exps, Scope* sc, bool pre
     bool err = false;
     if (exps)
     {
-        for (size_t i = 0; i < exps.dim; i++)
+        foreach (ref e; *exps)
         {
-            Expression e = (*exps)[i];
             if (e)
             {
-                e = e.expressionSemantic(sc);
-                if (e.op == TOK.error)
+                auto e2 = e.expressionSemantic(sc);
+                if (e2.op == TOK.error)
                     err = true;
-                if (preserveErrors || e.op != TOK.error)
-                    (*exps)[i] = e;
+                if (preserveErrors || e2.op != TOK.error)
+                    e = e2;
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/dmd/expression.d
```diff
@@ -1769,9 +1764,8 @@ extern (C++) abstract class Expression : RootObject
         {
             a = new Expressions();
             a.setDim(exps.dim);
-            for (size_t i = 0; i < a.dim; i++)
+            foreach (i, e; *exps)
             {
-                Expression e = (*exps)[i];
                 (*a)[i] = e ? e.syntaxCopy() : null;
```
wilzbach: Would be could to have `lockstep` / `zip` here, s.t. we can drop the iteration index in favor of the second range to be iterated.

src/dmd/expression.d
```diff
@@ -3499,9 +3492,8 @@ extern (C++) final class TupleExp : Expression
                 return false;
             if (e0 && !e0.equals(te.e0) || !e0 && te.e0)
                 return false;
-            for (size_t i = 0; i < exps.dim; i++)
+            foreach (i, e1; *exps)
             {
-                Expression e1 = (*exps)[i];
                 Expression e2 = (*te.exps)[i];
```
wilzbach: Another potential winner for `lockstep`

src/dmd/expression.d
```diff
@@ -3578,9 +3570,8 @@ extern (C++) final class ArrayLiteralExp : Expression
             {
                 return false;
             }
-            for (size_t i = 0; i < elements.dim; i++)
+            foreach (i, e1; *elements)
             {
-                Expression e1 = (*elements)[i];
                 Expression e2 = (*ae.elements)[i];
```
wilzbach: ... and another potential `lockstep` one.

src/dmd/expression.d
```diff
@@ -3840,9 +3831,8 @@ extern (C++) final class StructLiteralExp : Expression
                 return false;
             if (elements.dim != se.elements.dim)
                 return false;
-            for (size_t i = 0; i < elements.dim; i++)
+            foreach (i, e1; *elements)
             {
-                Expression e1 = (*elements)[i];
                 Expression e2 = (*se.elements)[i];
```
wilzbach: Yep, `lockstep` is a common pattern.

src/dmd/expression.d
```diff
@@ -4791,8 +4779,8 @@ extern (C++) final class IsExp : Expression
         {
             p = new TemplateParameters();
             p.setDim(parameters.dim);
-            for (size_t i = 0; i < p.dim; i++)
-                (*p)[i] = (*parameters)[i].syntaxCopy();
+            foreach (i, el; *parameters)
```
wilzbach: `foreach (e, ref e2; *parameters.lockstep(*p))` ...

jacob-carlborg (approved):

RazvanN7 (approved):

WalterBright:
It's not straightforward at all. Many of the loops change the array they are looping over, which doesn't work if it's converted to foreach. That's likely why this PR is failing.

PR #8118, opened 2018-04-03
fix Issue 18708 - Flow analysis in constructors not done correctly fo…

WalterBright:
…r if, ||, &&

### DIFF
--- a/src/dmd/ctorflow.d
+++ b/src/dmd/ctorflow.d
@@ -85,6 +85,22 @@ struct CtorFlow
         foreach (ref u; fieldinit)
             u |= csx;
     }
+
+    /******************************
+     * OR CSX bits to `this`
+     * Params:
+     *  ctorflow = bits to OR in
+     */
+    void OR(const ref CtorFlow ctorflow)
+    {
+        callSuper |= ctorflow.callSuper;
+        if (fieldinit.length && ctorflow.fieldinit.length)
+        {
+            assert(fieldinit.length == ctorflow.fieldinit.length);
+            foreach (i, u; ctorflow.fieldinit)
+                fieldinit[i] |= u;
+        }
+    }
 }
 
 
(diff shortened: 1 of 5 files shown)


src/dmd/ctorflow.d
```diff
@@ -85,6 +85,22 @@ struct CtorFlow
         foreach (ref u; fieldinit)
             u |= csx;
     }
+
+    /******************************
+     * OR CSX bits to `this`
+     * Params:
+     *  from = bits to OR in
```
jacob-carlborg: The name of the parameter is `ctorflow`.

test/fail_compilation/test18708.d
```diff
@@ -0,0 +1,64 @@
+/* TEST_OUTPUT:
+---
+fail_compilation/test18708.d(24): Error: one path skips field `s`
```
jacob-carlborg: Can the missing path be included in the error message?

jacob-carlborg (changes requested):

src/dmd/dscope.d
```diff
@@ -209,21 +209,11 @@ struct Scope
...
-                if (enclosing.ctorflow.fieldinit.length)
-                {
-                    assert(ctorflow.fieldinit.ptr != enclosing.ctorflow.fieldinit.ptr);
-                    foreach (i, u; ctorflow.fieldinit)
-                        enclosing.ctorflow.fieldinit[i] |= u;
-                }
-                ctorflow.freeFieldinit();
-            }
-        }
+            enclosing.ctorflow.OR(ctorflow);
```
jacob-carlborg: Indentention looks inconsistent.

jacob-carlborg (changes requested):

src/dmd/dscope.d
RazvanN7: It isn't. It looks inconsistent with the deleted lines

jacob-carlborg: Ok, I’m a bit limited here on my phone.

test/fail_compilation/test18708.d
WalterBright: Changing error messages is beyond the scope of this PR.

jacob-carlborg (approved):

PR #8119, opened 2018-04-03
Fix Issue 17626 - Same name variable assignment should raise a compile-time warning

wilzbach:
This is a rather common error:

```d
struct Foo
{
    this(int f)
    {
        f = f; // oops, but no warning
    }

    int f;
}
```

~~~and I wanted to see if how much would be affected by this.~~~
~~~That's why it's using `error` instead of `deprecation` (for now) and has  some tests just commented out.~~~

edit: experimented with this check and it found three real bugs (one in dmd and two in the tested projects).

### DIFF
--- a/src/dmd/ctorflow.d
+++ b/src/dmd/ctorflow.d
@@ -115,7 +115,6 @@ bool mergeFieldInitX(ref CSX fieldInit, const CSX fi)
         else if (!aHalt && aRet)
         {
             ok = (fi & CSX.this_ctor);
-            fieldInit = fieldInit;
         }
         else if (!bHalt && bRet)
         {
@@ -125,7 +124,6 @@ bool mergeFieldInitX(ref CSX fieldInit, const CSX fi)
         else if (aHalt)
         {
             ok = (fieldInit & CSX.this_ctor);
-            fieldInit = fieldInit;
         }
         else if (bHalt)
         {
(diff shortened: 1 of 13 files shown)


JinShil:
Currently, similar error messages in the compiler use "...has no effect".

https://github.com/dlang/dmd/search?utf8=%E2%9C%93&q=%22has+no+effect%22&type=

wilzbach:
BTW here's what I found from looking at the all the ~40 projects that we are testing.
1) Only 4 projects were affected
2) From these four failures - we catch two real bugs :)
- https://github.com/higgsjs/Higgs/blob/dbe1fb993addfef98e900df6fe2e1e3afe8ecae4/source/parser/lexer.d#L1021 (real bug!)
- https://github.com/atilaneves/automem/blob/5dff40105e3fdce78c5a1a513d6febdf8719d363/source/automem/ref_counted.d#L540 (unittest dummy)
- https://github.com/d-gamedev-team/gfm/blob/master/core/gfm/core/queue.d#L163 (real bug!)
- https://github.com/libmir/mir-algorithm/blob/f22937fe70970a220d970d27df1026becdf63f5a/source/mir/ndslice/sorting.d#L181
(`enum size_t naive = 32 > naive_est ? 32 : naive_est;` - a bug in this PR/implementation)

LemonBoy:
The implementation is too simplistic, consider the following snippet
```d
struct S
{
    int c = 0;
    this(this) { c++; }
}

void main()
{
    S a;
    assert(a.c==0);
    a = a;
    assert(a.c==0);
}
```
The second `assert` fails because the self-assignment **does** have a side-effect even though both the LHS and the RHS are `a`.

JinShil:
What about calling `hasSideEffect` on the expression, and if that doesn't produce the desired results, modify `hasSideEffect` until it does?

JinShil:
This should probably be closed.  See @WalterBright's comments in #8208.

wilzbach:
FWIW I disagree as this is a very common error in constructors, but well ...

PR #8120, opened 2018-04-03
Fix issue 18716: type `const(char)[]` can not be mapped to C++

jacob-carlborg:
D arrays don't have any corresponding type in C++. Instead we mangle it as a templated struct with the name `__dslice`, i.e. `struct __dslice(T)`, where `T` is the element type of the array. For an array of ints it would be mangled as the following type: `__dslice!int`.

Reboot of #7893.

@WalterBright this should interesting for you since you want to kill C string in DMD.


src/dmd/root/array.d
```diff
@@ -29,6 +29,21 @@ private:
 public:
     @disable this(this);
 
+    /**
+     * Connivence constructor to add the given elements to this array.
```
wilzbach: Convenience

src/dmd/typesem.d
```diff
@@ -497,6 +497,83 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
...
+         * by `dsliceTemplateDeclaration`.
+         *
+         * The template is instantiated with the element type of this array. For
+         * an array of ints it would correspond to the following D code:
+         * `__dslice!int`.
+         *
+         * Returns: a template instance
+         *
+         * See_Also: `dsliceTemplateDeclaration`
+         * See_Also: `typeForCppMangling`
```
wilzbach: `$(LREF dsliceTemplateDeclaration)` would work here.

WalterBright:
Thanks for doing the work on this, it's a nice job.

I've thought about doing this now and then, having run into the issue myself, but I have reservations:

* The user has to declare the array type in the C++ code. I run into strong resistance whenever I suggest something like this. If anything, they want an `std::vector<T>` interface.
* There's already a solution - make the function `extern (C)`.
* Or just pass the array parameter as `.ptr` and `.length`.

The conclusion I always come to is it is not worth the extra complexity in the D core language.

src/dmd/typesem.d
jacob-carlborg: Would that macro work for a nested function?

jacob-carlborg:
@WalterBright This primarily for use within DMD. 

Based on your conclusion, you don’t want this feature?

wilzbach:
> The user has to declare the array type in the C++ code. I run into strong resistance whenever I suggest something like this.

As @jacob-carlborg mentioned this is primarily intended for DMD, s.t. D strings can be used everywhere and passed to LDC/GDC.
LDC even has something like this on their side (see https://github.com/dlang/dmd/pull/7893#issuecomment-365767743)

See also: https://github.com/dlang/dmd/pull/7893 (another attempt at making the interoperability possible)

> If anything, they want an std::vector<T> interface.

FWIW: Last time I checked, this hit multiple dead-ends (see: https://github.com/dlang/druntime/pull/1316) :/

> The conclusion I always come to is it is not worth the extra complexity in the D core language.

Hmm what's your plan on passing D strings to LDC/GDC then?

Geod24:
> Or just pass the array parameter as `.ptr` and `.length`.

I think this is a much more sensible approach. Instead of introducing another symbol, could we turn `extern (C++) void func(const(char)[] slice)` into `extern (C++) void func(const(char)* ptr, size_t length)` in the backend ?

jacob-carlborg:
> I think this is a much more sensible approach. Instead of introducing another symbol, could we turn extern (C++) void func(const(char)[] slice) into extern (C++) void func(const(char)* ptr, size_t length) in the backend ?

I disagree.

It would be more complicated to rewrite the AST or the IR. I think it’s too late to do it in the backend. My approach is ABI compatible so no rewriting is required.

jacob-carlborg:
@wilzbach auto-tester is reporting that all tests passed here, but looking at the details, one test failed.

WalterBright:
> This primarily for use within DMD.

Then I definitely feel this is not worth it. It would be simple enough to manually rewrite:

```
foo(T[]);
foo(a[])
```
as:
```
foo(Array!(T[]));
foo(Array(a[]);
```
> Based on your conclusion, you don’t want this feature?

Right, not worth it.

ibuclaw:
Hmm. When I think about moving the dmd implementation from `char*` to `char[]`, two things come to mind.

1. Wrap it.
```
extern(C++) class SomeType
{
  extern(D) char[] doSometing(char[] input);

  char* doSomething(char* input)
  {
    // or .toStringz() instead of ptr.
    return doSomething(input[0..strlen(input)]).ptr;
  }
}
```
2. Remove it. There are many methods that are unused outside the frontend and so better off internal to the parser/semantic as far as C++ api is concerned.

jacob-carlborg:
> Remove it. There are many methods that are unused outside the frontend and so better off internal to the parser/semantic as far as C++ api is concerned.

How can we know which methods are used in GDC/LDC?

ibuclaw:
> How can we know which methods are used in GDC/LDC?

We will tell you. 😉

...

Let's start with a rule of thumb. The role of the interface is to initialize the frontend, call entrypoints to read, parse, semantic, and optionally generate header/ddoc/json files for all modules.
It may also provide its own interface for various tasks such as CTFloat, Target, Port, and builtins, and may require access to ctfeInterpret or new'ing objects via Foo::create.
Finally it traverses the AST, _gets_ information out of the object/trees (hopefully without rewriting any part of in-place) and generates code.

Whatever doesn't fall into any of those categories should be good for the taking. i.e: AST types that only live during semantic analysis, anything that sets or mutates state, and resolver methods and functions.

Off the top of my head, the only functions that need to continue returning a `char*` are mangleExact, toChars, and toPrettyChars. All of which could have an `extern(D)` or have a new name toString, toPrettyString, etc... for the string variant.

ibuclaw:
> optionally generate header/ddoc/json files for all modules.

For example (this might already have been done), the only function that needs to be exposed to C++ for header generation is `genhdrfile`, everything else in the hdr module can be `extern(D)`.

thewilsonator:
@WalterBright with the (questionable) power of C++ implicit conversions this _is_ a `std::vector<T>` interface, behold:
```C++
#include <vector>

template<typename T> struct __dslice { 
    size_t length; T* ptr;
    __dslice(const std::vector<T>& x) 
    {
         length = x.size();
         ptr = (T*)&x[0]; // seems to default to const not sure why hence the cast
    }
};
void foo(__dslice<int> a) {};

int main(int a, char** b)
{
    std::vector<int> x;
    foo(x);
}
```
Proof this compiles: https://godbolt.org/g/VFA1xV

If somebody is using the code then they are already `#include`ing something, include the definition as part of that (maybe adding an include guard to avoid multiple definitions).

jacob-carlborg:
> Wrap it

@ibuclaw Unfortunately it's not possible to overload a function taking `const(char)*` and one taking `const(char)[]` when passing a string literal:

```d
void symbol_name(const(char)* name) {}
void symbol_name(const(char)[] name) {}

void main()
{
    symbol_name("foo");
}
```

```
main.d(6): Error: main.symbol_name called with argument types (string) matches both:
main.d(1):     main.symbol_name(const(char)* name)
and:
main.d(2):     main.symbol_name(const(char)[] name)
```

ibuclaw:
@jacob-carlborg - then make the C++ overload include a length parameter?

Identifier::idPool is a good example of this, the `const(char)*` override is only declared in C++ headers.

D:
https://github.com/dlang/dmd/blob/f1effe0b30bbb9392597278c27e24fec0446a2a6/src/dmd/identifier.d#L175-L202

C++:
https://github.com/dlang/dmd/blob/f1effe0b30bbb9392597278c27e24fec0446a2a6/src/dmd/identifier.h#L42-L47

jacob-carlborg:
> then make the C++ overload include a length parameter?

@ibuclaw there’s already one overload with and one without the length parameter. 

https://github.com/dlang/dmd/blob/master/src/dmd/backend/global.d#L335

jacob-carlborg:
@thewilsonator rebased.

thewilsonator:
Hmm, it seems to be segfaulting on windows? No idea why though.

See e.g. https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3231832&isPull=true

thewilsonator:
Also with the `typeForCppMangling()` done as part of semantic it seems that `__dslice!T` will be instantiated for every `T[]`. I think it would make more sense as a private method of `TypeDArray` so that it is done lazily only when `typeForMangling(LINK.cpp)` is called.

jacob-carlborg:
> Also with the typeForCppMangling() done as part of semantic it seems that __dslice!T will be instantiated for every T[]. I think it would make more sense as a private method of TypeDArray so that it is done lazily only when typeForMangling(LINK.cpp) is called.

@thewilsonator to do that I need access to a scope, the `sc` variable here: https://github.com/dlang/dmd/pull/8120/files#diff-64d4abe08ba2355c2b95cb22e46b8572R928. This is not available in `cppmangle` where `typeForMangling` is called. What do you suggest?

thewilsonator:
That seems to work (apart from the segfault, which I still can't see why it would fail only on windows). For compiler synthesised stuff its okay to cheat and not run semantic e.g. https://github.com/dlang/dmd/blob/master/src/dmd/statementsem.d#L3528 (for expressions) not sure how to do that for types though.

test/compilable/cppmangle.d
```diff
@@ -760,3 +760,37 @@ version(Windows)
...
+extern(C++) void test18716a(char[]);
+extern(C++) void test18716b(__dslice!char);
+extern(C++) void test18716c(const(char)[]);
+extern(C++) void test18716d(Test18716[]);
+
+void test18716()
+{
+    test18716a("foo".dup);
+    test18716c("foo");
+    test18716d([Test18716()]);
```
thewilsonator: These three lines are causing the segfault, I've still no idea where in the code though. Likely not in the sections you're changed :(

thewilsonator: Got it!

```diff
    void mangleIdent(Dsymbol sym, bool dont_use_back_reference = false)
    {
        // <qualified name> ::= <sub-name list> @
        // <sub-name list>  ::= <sub-name> <name parts>
        //                  ::= <sub-name>
        // <sub-name> ::= <identifier> @
        //            ::= ?$ <identifier> @ <template args> @
        //            :: <back reference>
        // <back reference> ::= 0-9
        // <template args> ::= <template arg> <template args>
        //                ::= <template arg>
        // <template arg>  ::= <type>
        //                ::= $0<encoded integral number>
        //printf("mangleIdent('%s')\n", sym.toChars());
        Dsymbol p = sym;
-        if (p.toParent() && p.toParent().isTemplateInstance())
+        if (p && p.toParent() && p.toParent().isTemplateInstance())
        {
            p = p.toParent();
        }
        while (p && !p.isModule())
        {
            mangleName(p, dont_use_back_reference);
            p = p.toParent();
            if (p.toParent() && p.toParent().isTemplateInstance())
            {
                p = p.toParent();
            }
        }
        if (!dont_use_back_reference)
            buf.writeByte('@');
    }
```

jacob-carlborg:
> For compiler synthesised stuff its okay to cheat and not run semantic

IIRC, I did has little as possible to make the code compile.

thewilsonator:
Should work either way. Add a null check to the start of `mangleIdent` and I think we should be all good to go.

jacob-carlborg:
@thewilsonator All green for the auto-tester.

thewilsonator:
Sweet.

thewilsonator:
@WalterBright Your original reservations have been fixed, please re-evaluate. See https://github.com/thewilsonator/interop/blob/master/c++/dslice.h

WalterBright:
@thewilsonator you have indeed created a C++ type that can interface to some D arrays. I don't think it's a direction we should be going in, though. My original suggestion for people interfacing D with C++ is to be flexible on the C++ side - and incorporating these C++ types is being flexible. But they aren't flexible. It's a hard sell. I still don't think it is worth it.

src/dmd/root/array.h
```diff
@@ -240,4 +240,17 @@ struct BitArray
     BitArray(const BitArray&);
 };
 
+// This is the type that the D compiler will mangle D slices as when mangling
+// for C++. This type is ABI compatible with native D slices.
```
kinke: Is ABI compatibility actually tested somewhere? I seriously doubt it is on Win64 for example. LDC on Win64 returns and passes slices as pairs (in 2 registers if available), whereas a 128-bit struct is passed indirectly by value (hidden ref), and there is no way for a struct to emulate that behaviour.

[Btw, the `__dslice` ctors below make MSVC++ treat this as a non-POD and return it via hidden pointer, but that's irrelevant here, as it's > 64 bits anyway.]

thewilsonator:
> My original suggestion for people interfacing D with C++ is to be flexible on the C++ side - and incorporating these C++ types is being flexible. But they aren't flexible. It's a hard sell.

I don't follow, who is "they"? C++ types? C++ people? 

The former is a contradiction in terms, it cannot be both flexible and not flexible.
The latter, either they are none the wiser because it works: I give you some header files and a library, you use a `std::vector` or whatever, and you don't even know (unless you bother looking an the source instead of the docs that say "it effectively takes a `std::span<T,dynamic_extent>`") that it takes a `__dslice<T>`. Or they are flexible enough to be using D in the first place. Is the harness of the sell any harder than using D in the first place? I think this makes it easier, no need to pass the pointer and length to a separate function, less stuff to think about.

src/dmd/root/array.h
jacob-carlborg: > Is ABI compatibility actually tested somewhere?

No.

jacob-carlborg: > LDC on Win64 returns and passes slices (and delegates IIRC) as pairs (in 2 registers if available), whereas a 128-bit struct is passed/returned indirectly by value (hidden ref), and there is no way for a struct to emulate that behavior.

That's unfortunate.

wilzbach:
FWIW in the state of D survey this has been requested a few times too.
The way I see it, people have been able to expose pointer ABIs to C++ for a lone time and not many people did so, because it means losing all the beauty of D's arrays.
Sure, one could wrap natives arrays and strings in a wrapper type, but that's another step of effort and often people just want to use a few of their methods in C++ and they really don't want to add additional overloads to libraries.

With this change, people can choice whether they want to adapt the C++ side or the D side, before there was no choice. If someone prefers to add pointer+length overloads or wrapping it in custom types, he can already do this and won't be affected by this improvement. However, those who prefer the D style will get an additional reason to like D, no?

thewilsonator:
@WalterBright ahh I get you. 

> Passing arrays as a special __dslice they have to add a special code for on the C++ side is not going to go over well

The point of [its constructor](https://github.com/thewilsonator/interop/blob/master/c%2B%2B/dslice.h#L66) is that given 
```
void foo(__dslice<int> a);
```
you can pass a `std::vector`, a `std::span`, a `gsl::span`, `std::string_view` or any custom container `U`, that supports `U u; size_t i = u.size(); T* = u.data();` to it.

Likewise  
```C++
template<typename T struct __dslice
{
    size_t length; T* ptr;
...
operator gsl::span<T>(){ return gsl::span<T>(ptr,length);}
};
__dslice<int> bar();
...
gsl::span<int> c = bar();
```
works. This means that you don't ever have to _use_ `__dslice` directly, use a`gsl::span`/`std::span` which the C++ core guidelines encourage people to use.

We can synthesise D `T[]` wrappers for any code that uses `gsl::span<T>` or `std::span<T>`

thewilsonator:
Thanks. @WalterBright can we please get the green light for this?

kinke:
I don't see how you can simply discard the ABI mismatch on Win64. If the purpose is direct slices interop between D and C++, but it doesn't work on all platforms, it's useless to me.

test/compilable/cppmangle.d
```diff
@@ -976,3 +976,38 @@ else version(Posix)
     static assert(test19278_4.mangleof == "_ZN5hello5world3yay11test19278_4Ev");
     static assert(test19278_var.mangleof == "_ZN5hello5world13test19278_varE");
 }
+
+/*****************************************/
+// https://issues.dlang.org/show_bug.cgi?id=18716
+// D slices as parameters in extern(C++) functions
+
+struct Test18716 {}
+struct __dslice(T) {}
```
Geod24: The tests needs to be in runnable

jacob-carlborg: The ABI doesn't match, so there's no point, at least not yet.

Geod24: The point is exactly that the ABI does not match.

jacob-carlborg: I'm not following. This whole PR is moot as long as the ABI problem exists.

thewilsonator:
Sorry for the confusion, I meant green light as in he's be happy with the feature, so that it would then not be a waste of time fixing ABI issues for a PR that would be rejected anyway.

test/compilable/cppmangle.d
Geod24: Yep, I was just saying that it's not obvious, since there are tests, and the PR is green.
But IMO, this PR should ultimately be accepted, but it looks like you fear it to be vetoed.

thewilsonator: Yes. It should, but Walter hasn't responded to the reasoning why his concerns are unfounded.

jacob-carlborg: I’m not going to spend time on moving the tests if/until I fix the ABI issue.

jacob-carlborg: Walter did already vetoed it.

thewilsonator: But not sensibly.

MoonlightSentinel: #11073 contains a `runnable_cxx` test (if someone wants to use it)

kinke:
I'm not familiar with the DMD ABI details, but at least for LDC, a half-feasible option would be mangling a slice parameter as 2 separate C++ parameters - `size_t` + `T*`. For Itanium, that should work, as the return type isn't part of the mangle; MSVC++ does encode it, and well, there's only a single return value. :/

Imperatorn:
Any update on this?

RazvanN7:
@WalterBright how should we proceed here?

ljmf00:
Ping @WalterBright . This is a good step to make the compiler `strlen`-free.

ljmf00:
> > Ping @WalterBright . This is a good step to make the compiler `strlen`-free.
> 
> How? Many things are still being passed to the front-end from C++ land.

I'm sure there is a lot of benefits on make this step, especially making the compiler use more `memcpy`'s and less `strcpy`'s, that surely benefits better from SIMD implementations. The `strlen` part is another bonus that basically removes some `O(N)` `strlen` to `O(1)` accesses.

PR #8121, opened 2018-04-03
fix Issue 18729 - dmd -run executes in different environment

marler8997:
TLDR: the following fails on windows
```
echo void main(){} > voidmain.d
dmd -run rdmd.d -m64 voidmain.d

OPTLINK : Warning 9: Unknown Option : OUT
OPTLINK : Warning 9: Unknown Option : LIBPATH
OPTLINK : Warning 9: Unknown Option : LIBPATH
OPTLINK : Warning 9: Unknown Option : LIBPATH
...
```

I ran into this problem when I was running a D program via `dmd -run` that itself invoked DMD.  The resulting program was invoking the compiler with `-m64`, but the original `dmd -run` did not use `-m64`.  This caused a problem because the first invocation of the compiler set the `LINKCMD` environment variable to the OPTLINK executable, which caused the second invocation to use the OPTLINK executable but it treated it like the MSVC linker because it was running in 64-bit mode.

The solution here is to maintain the original environment (the environment before it was changed by the compiler) when `dmd -run` executes the compiled program.  This PR accomplishes this by saving the original values of environment variables when they are overwritten and then restoring these values before executing the compiled program.

### DIFF
--- a/src/dmd/dinifile.d
+++ b/src/dmd/dinifile.d
@@ -19,6 +19,7 @@ import core.sys.posix.stdlib;
 import core.sys.windows.winbase;
 import core.sys.windows.windef;
 
+import dmd.env;
 import dmd.errors;
 import dmd.globals;
 import dmd.root.rmem;
@@ -28,7 +29,6 @@ import dmd.root.port;
 import dmd.root.stringtable;
 import dmd.utils;
 
-version (Windows) extern (C) int putenv(const char*) nothrow;
 private enum LOG = false;
 
 /*****************************
@@ -145,20 +145,11 @@ void updateRealEnvironment(ref StringTable environment)
     static int envput(const(StringValue)* sv) nothrow
     {
         const name = sv.toDchars();
-        const namelen = strlen(name);
         const value = cast(const(char)*)sv.ptrvalue;
         if (!value) // deleted?
             return 0;
-        const valuelen = strlen(value);
-        auto s = cast(char*)malloc(namelen + 1 + valuelen + 1);
-        if (!s)
-            Mem.error();
-        memcpy(s, name, namelen);
-        s[namelen] = '=';
-        memcpy(s + namelen + 1, value, valuelen);
-        s[namelen + 1 + valuelen] = 0;
-        //printf("envput('%s')\n", s);
-        putenv(s);
+        if (putenvRestorable(name.toDString, value.toDString))
+            assert(0);
         return 0; // do all of them
     }
 
--- a/src/dmd/env.d
+++ b/src/dmd/env.d
@@ -1,0 +1,77 @@
+module dmd.env;
+
+import core.stdc.string;
+import core.sys.posix.stdlib;
+import dmd.globals;
+import dmd.root.array;
+import dmd.root.rmem;
+import dmd.utils;
+
+version (Windows)
+    private extern (C) int putenv(const char*) nothrow;
+
+/**
+Construct a variable from `name` and `value` and put it in the environment while saving
+the previous value of the environment variable into a global list so it can be restored later.
+Params:
+    name = the name of the variable
+    value = the value of the variable
+Returns:
+    true on error, false on success
+*/
+bool putenvRestorable(const(char)[] name, const(char)[] value) nothrow
+{
+    saveEnvVar(name);
+    const nameValue = allocNameValue(name, value);
+    const result = putenv(cast(char*)nameValue.ptr);
+    version (Windows)
+        mem.xfree(cast(void*)nameValue.ptr);
+    else
+    {
+        if (result)
+            mem.xfree(cast(void*)nameValue.ptr);
+    }
+    return result ? true : false;
+}
+
+/**
+Allocate a new variable via xmalloc that can be added to the global environment. The
+resulting string will be null-terminated immediately after the end of the array.
+Params:
+    name = name of the variable
+    value = value of the variable
+Returns:
+    a newly allocated variable that can be added to the global environment
+*/
+string allocNameValue(const(char)[] name, const(char)[] value) nothrow
+{
+    const length = name.length + 1 + value.length;
+    auto str = (cast(char*)mem.xmalloc(length + 1))[0 .. length];
+    str[0 .. name.length] = name[];
+    str[name.length] = '=';
+    str[name.length + 1 .. length] = value[];
+    str.ptr[length] = '\0';
+    return cast(string)str;
+}
+
+/// Holds the original values of environment variables when they are overwritten.
+private __gshared string[string] envNameValues;
+
+/// Restore the original environment.
+void restoreEnvVars()
+{
+    foreach (var; envNameValues.values)
+    {
+        if (putenv(cast(char*)var.ptr))
+            assert(0);
+    }
+}
+
+/// Save the environment variable `name` if not saved already.
+void saveEnvVar(const(char)[] name) nothrow
+{
+    if (!(name in envNameValues))
+    {
+        envNameValues[name.idup] = allocNameValue(name, name.toCStringThen!(n => getenv(n.ptr)).toDString);
+    }
+}
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -21,6 +21,7 @@ import core.sys.posix.unistd;
 import core.sys.windows.winbase;
 import core.sys.windows.windef;
 import core.sys.windows.winreg;
+import dmd.env;
 import dmd.errors;
 import dmd.globals;
 import dmd.root.file;
@@ -30,7 +31,6 @@ import dmd.root.rmem;
 import dmd.utils;
 
 version (Posix) extern (C) int pipe(int*);
-version (Windows) extern (C) int putenv(const char*);
 version (Windows) extern (C) int spawnlp(int, const char*, const char*, const char*, const char*);
 version (Windows) extern (C) int spawnl(int, const char*, const char*, const char*, const char*);
 version (Windows) extern (C) int spawnv(int, const char*, const char**);
@@ -769,17 +769,11 @@ version (Windows)
         {
             if ((len = strlen(args)) > 255)
             {
-                char* q = cast(char*)alloca(8 + len + 1);
-                sprintf(q, "_CMDLINE=%s", args);
-                status = putenv(q);
+                status = putenvRestorable("_CMDLINE", args[0 .. len]);
                 if (status == 0)
-                {
                     args = "@_CMDLINE";
-                }
                 else
-                {
                     error(Loc.initial, "command line length of %d is too long", len);
-                }
             }
         }
         // Normalize executable path separators
@@ -897,6 +891,7 @@ public int runProgram()
         argv.push(a);
     }
     argv.push(null);
+    restoreEnvVars();
     version (Windows)
     {
         const(char)[] ex = FileName.name(global.params.exefile);
@@ -1053,16 +1048,18 @@ version (Windows)
                 {
                     // debug info needs DLLs from $(VSInstallDir)\Common7\IDE for most linker versions
                     //  so prepend it too the PATH environment variable
-                    const char* path = getenv("PATH");
+                    const path = getenv("PATH");
                     const pathlen = strlen(path);
                     const addpathlen = strlen(addpath);
 
-                    char* npath = cast(char*)mem.xmalloc(5 + pathlen + 1 + addpathlen + 1);
-                    memcpy(npath, "PATH=".ptr, 5);
-                    memcpy(npath + 5, addpath, addpathlen);
-                    npath[5 + addpathlen] = ';';
-                    memcpy(npath + 5 + addpathlen + 1, path, pathlen + 1);
-                    putenv(npath);
+                    const length = addpathlen + 1 + pathlen;
+                    char* npath = cast(char*)mem.xmalloc(length);
+                    memcpy(npath, addpath, addpathlen);
+                    npath[addpathlen] = ';';
+                    memcpy(npath + addpathlen + 1, path, pathlen);
+                    if (putenvRestorable("PATH", npath[0 .. length]))
+                        assert(0);
+                    mem.xfree(npath);
                 }
                 return cmdbuf.extractChars();
             }
--- a/src/win32.mak
+++ b/src/win32.mak
@@ -139,7 +139,7 @@ FRONT_SRCS=$D/access.d $D/aggregate.d $D/aliasthis.d $D/apply.d $D/argtypes.d $D
 	$D/cond.d $D/constfold.d $D/cppmangle.d $D/cppmanglewin.d $D/ctfeexpr.d $D/ctorflow.d $D/dcast.d $D/dclass.d		\
 	$D/declaration.d $D/delegatize.d $D/denum.d $D/dimport.d $D/dinifile.d $D/dinterpret.d	\
 	$D/dmacro.d $D/dmangle.d $D/dmodule.d $D/doc.d $D/dscope.d $D/dstruct.d $D/dsymbol.d $D/dsymbolsem.d		\
-	$D/lambdacomp.d $D/dtemplate.d $D/dversion.d $D/escape.d			\
+	$D/lambdacomp.d $D/dtemplate.d $D/dversion.d $D/env.d $D/escape.d			\
 	$D/expression.d $D/expressionsem.d $D/func.d $D/hdrgen.d $D/id.d $D/imphint.d	\
 	$D/impcnvtab.d $D/init.d $D/initsem.d $D/inline.d $D/inlinecost.d $D/intrange.d $D/json.d $D/lib.d $D/link.d	\
 	$D/mars.d $D/mtype.d $D/nogc.d $D/nspace.d $D/objc.d $D/opover.d $D/optimize.d $D/parse.d	\
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -317,7 +317,7 @@ FRONT_SRCS=$(addsuffix .d, $(addprefix $D/,access aggregate aliasthis apply argt
 	arraytypes astcodegen ast_node attrib builtin canthrow cli clone compiler complex cond constfold	\
 	cppmangle cppmanglewin ctfeexpr ctorflow dcast dclass declaration delegatize denum dimport	\
 	dinifile dinterpret dmacro dmangle dmodule doc dscope dstruct dsymbol dsymbolsem	\
-	dtemplate dversion escape expression expressionsem func			\
+	dtemplate dversion env escape expression expressionsem func			\
 	hdrgen id impcnvtab imphint init initsem inline inlinecost intrange	\
 	json lambdacomp lib libelf libmach link mars mtype nogc nspace objc opover optimize parse permissivevisitor sapply templateparamsem	\
 	sideeffect statement staticassert target typesem traits transitivevisitor parsetimevisitor visitor	\
--- a/test/dshell/extra-files/printenv.d
+++ b/test/dshell/extra-files/printenv.d
@@ -1,0 +1,11 @@
+import std.array, std.stdio, std.process, std.algorithm;
+void main()
+{
+    foreach (varPair; environment.toAA().byKeyValue.array.sort!"a.key < b.key")
+    {
+        if (varPair.key != "_")
+        {
+            writeln(varPair.key, "=", varPair.value);
+        }
+    }
+}
--- a/test/dshell/sameenv.d
+++ b/test/dshell/sameenv.d
@@ -1,0 +1,23 @@
+import dshell;
+void main()
+{
+    const envFromExe = shellExpand("$OUTPUT_BASE/envFromExe.txt");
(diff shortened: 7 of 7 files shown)


test/runnable/sameenv.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+
+$DMD -m${MODEL} -of${OUTPUT_BASE}/printenv${EXE} ${EXTRA_FILES}/printenv.d
+${OUTPUT_BASE}/printenv${EXE} > ${OUTPUT_BASE}/envFromExe.txt
+
+$DMD -m${MODEL} -run ${EXTRA_FILES}/printenv.d > ${OUTPUT_BASE}/envFromRun.txt
+
+diff ${OUTPUT_BASE}/envFromExe.txt ${OUTPUT_BASE}/envFromRun.txt
```
wilzbach: I always found adding `-p` more helpful when looking at errors. Maybe we should overwrite add an `alias diff = diff -p` to the "bash runtime environment"?

wilzbach: Also maybe you want to compare it with the output from `env`?

marler8997: output from env might be different, it sorts differently I believe.

wilzbach: Then normalize both by  `| sort`?

marler8997: I get different outputs from `env` and this program on windows.  It sorts a bit different.  In any case, we don't really care that the output exactly matches the output of `env`, we just want to make sure running the executable directly matches `dmd -run`.

src/dmd/dinifile.d
```diff
@@ -153,7 +153,7 @@ void updateRealEnvironment(StringTable* environment)
         memcpy(s + namelen + 1, value, valuelen);
         s[namelen + 1 + valuelen] = 0;
         //printf("envput('%s')\n", s);
-        putenv(s);
```
marler8997: TODO: Not sure why this call to `putenv` exists...we should determine why this was put here.

src/dmd/link.d
```diff
@@ -151,6 +172,37 @@ version (Posix)
...
+
+    /**
+    Access the saved environment variable value.
+    */
+    string saved() const { return _saved; }
+
+    /**
+    Call before the overriding the environment variable to save and restore it later.
+    */
+    void saveToRestoreLater()
```
wilzbach: How about wrapping `EnvRestore` as a voldemort type in a function and thus doing this in at construction?

src/dmd/link.d
```diff
@@ -151,6 +172,37 @@ version (Posix)
     }
 }
 
+/**
+Use to restore an environment variable when this structure goes out of scope.
+*/
+struct EnvRestore(string varName)
+{
+    private string _saved;
+    private bool restore;
```
wilzbach: `isSaved` / `hasSaved`?

src/dmd/mars.d
```diff
@@ -266,7 +266,8 @@ private int tryMain(size_t argc, const(char)** argv)
     sections.push(envsection.ptr);
     parseConfFile(&environment, global.inifilename, inifilepath, inifile.len, inifile.buffer, &sections);
     getenv_setargv(readFromEnv(&environment, "DFLAGS"), &arguments);
-    updateRealEnvironment(&environment);
+    // TODO: not sure why dmd is doing this
+    //updateRealEnvironment(&environment);
```
wilzbach: Because it has the "great concept" of replacing `DFLAGS` completely with the content of the INI file.
See also: https://github.com/dlang/tools/pull/343

src/dmd/link.d
```diff
@@ -32,6 +32,27 @@ version (Windows) extern (C) int putenv(const char*);
...
+int putenv2(const char[] name, const char[] value)
+{
+    auto fullLength = name.length + 1 + value.length;
+    auto buffer = cast(char*)alloca(fullLength + 1);
+    buffer[0               .. name.length] = name[];
+    buffer[name.length]                    = '=';
+    buffer[name.length + 1 .. fullLength ] = value[];
+    buffer[fullLength] = '\0';
+    return putenv(buffer);
+}
```
wilzbach: Imho `{get,put}env2` is a genuinely bad name for functions for D-ified versions.
Maybe `dgetenv` or simply use overloading?

src/dmd/link.d
```diff
@@ -738,15 +791,16 @@ version (Windows)
...
             message("%s %s", cmd, args);
         if (!global.params.mscoff)
         {
             if ((len = strlen(args)) > 255)
             {
-                char* q = cast(char*)alloca(8 + len + 1);
-                sprintf(q, "_CMDLINE=%s", args);
-                status = putenv(q);
+                restore_CMDLINE.saveToRestoreLater();
+                status = putenv2("_CMDLINE", args);
```
wilzbach: Use D strings -> `args[0 .. len]`

src/dmd/link.d
```diff
@@ -32,6 +32,27 @@ version (Windows) extern (C) int putenv(const char*);
 version (Windows) extern (C) int spawnlp(int, const char*, const char*, const char*, const char*);
 version (Windows) extern (C) int spawnl(int, const char*, const char*, const char*, const char*);
 version (Windows) extern (C) int spawnv(int, const char*, const char**);
+
+string getenv2(const(char)* name)
+{
+    auto result = getenv(name);
+    return cast(string)result[0 .. strlen(result)];
+}
+int putenv2(const char[] name, const(char)* value)
```
wilzbach: Do we really need the C-string overload? Remember they "must die" ;-)

src/dmd/link.d
marler8997: it doesn't always save the environment variable, only in some cases.

src/dmd/globals.d
```diff
@@ -68,6 +73,75 @@ enum CPU
     native              // the machine the compiler is being run on
 }
 
+/**
+Used to store an environment variable including its value.
+*/
+extern (C++) struct EnvironmentVariable
```
ibuclaw: Why does this need to be in global? Nobody apart from dmd would use this.

src/dmd/globals.d
```diff
@@ -219,6 +294,7 @@ struct Param
 
     bool run; // run resulting executable
     Strings runargs; // arguments for executable
+    Array!EnvironmentVariable envToRestoreBeforeRun;
```
ibuclaw: This is only accessed by private functions, and shouldn't be in global params accessible by all.

ibuclaw (dismissed):
For better encapsulation, these routines should be free functions and declared `extern(D)`, if dinifile or utils isn't a good place, then it's own module, where both EnvironmentVariable struct and array are private.

marler8997:
@ibuclaw good suggestions.  Thanks for taking the time to review.

src/win32.mak
```diff
@@ -153,7 +153,7 @@ FRONT_SRCS=$D/access.d $D/aggregate.d $D/aliasthis.d $D/apply.d $D/argtypes.d $D
 	$D/cond.d $D/constfold.d $D/cppmangle.d $D/cppmanglewin.d $D/ctfeexpr.d $D/ctorflow.d $D/dcast.d $D/dclass.d		\
 	$D/declaration.d $D/delegatize.d $D/denum.d $D/dimport.d $D/dinifile.d $D/dinterpret.d	\
 	$D/dmacro.d $D/dmangle.d $D/dmodule.d $D/doc.d $D/dscope.d $D/dstruct.d $D/dsymbol.d $D/dsymbolsem.d		\
-	$D/lambdacomp.d $D/dtemplate.d $D/dversion.d $D/escape.d			\
+	$D/lambdacomp.d $D/dtemplate.d $D/dversion.d $D/env.d $D/escape.d			\
```
stefan-koch-sociomantic: positional difference when compared to posix makefile.

stefan-koch-sociomantic (changes requested):
Please keep the make files as much as possible in sync

src/win32.mak
wilzbach: @stefan-koch-sociomantic this is fine. The windows Makefile is only semi-ordered anyhow ...

marler8997:
Ping

marler8997:
Ping

marler8997:
Re-opening, I'm hopeful that with the re-organization, that this one will be able to get in.

JinShil:
Forgive me if I'm missing the point, but shouldn't `dmd -m64 -run rdmd.d -m64 voidmain.d` do the right thing?  That is, the position of the switches relative to the `-run` switch determines which switches are passed to which invocation?

marler8997:
Yeah that command would work. Though, it's different than the problem command that this PR fixes. Dmd uses environment variables to pass settings to other programs, but then doesn't clean up those environment variables before invoking the resulting executable with -run.

marler8997:
By the way, if you're sitting in the dlang tools repo, you can do this:
```bash
./rdmd.d rdmd.d rdmd.d rdmd.d ...[repeat forver]
```
You can just keep going, and eventually it will just print out the rdmd usage...lol :)

thewilsonator (approved):

src/dmd/env.d
```diff
@@ -0,0 +1,134 @@
...
+struct Env
+{
+    private const(char)* setString; /// null-terminated string of the form "VAR=VALUE"
+    private uint length;            /// length of setString
+    private uint equalsIndex;       /// index of '=' in setString
+
+    /**
+    Access the name of the environment variable as a D string.
+    */
+    auto nameString() const { return setString[0 .. equalsIndex]; }
```
jacob-carlborg: Why the `String` suffix?

src/dmd/env.d
```diff
@@ -0,0 +1,134 @@
...
+import core.sys.posix.stdlib;
+import dmd.globals;
+import dmd.root.array;
+import dmd.root.rmem;
+
+version (Windows) extern (C) int putenv(const char*) nothrow;
+
+/**
+Used to store an environment variable including its value.
+*/
```
jacob-carlborg: You can save two lines by using the `///` syntax.

src/dmd/env.d
```diff
@@ -0,0 +1,134 @@
...
+struct Env
+{
+    private const(char)* setString; /// null-terminated string of the form "VAR=VALUE"
```
jacob-carlborg: Can we use a D string?

marler8997: We could but I don't see the benefit here.

marler8997: Right but this isn't being used as a string.  It's just being passed to the C function `putenv`.

marler8997: I could still make it a string, but this isn't really being used like other strings we want to convert to D strings, hence why I don't see the benefit.  Maybe you can show me a place where using a D string would prevent an error?  I don't see any.

marler8997: This member MUST be a null-terminated string.  And I've noticed that when DMD has to use a null-terminated string, it usually uses `char*` to signify that.  If I use `char[]` would that be confusing?

marler8997: Ok I've noticed other places where DMD uses D arrays for strings even when they need to be null-terminated.  So I've gone ahead and made it a `string`.

WalterBright (dismissed):
First off, bug fixes should have a bugzilla issue filed for them. Second, I'm not sure this is even a bug. Why should running a compiled program inherit the environment set by the compiler? The two have no necessary connection between them. I think it's a bug if it **did** inherit the environment variables set by the compiler. This shouldn't be polymorphism.

I.e. need a much better justification.

WalterBright:
Ok, I see it does have a bugzilla issue. I fixed the title to follow the usual form for bug fixes.

marler8997:
> Why should running a compiled program inherit the environment set by the compiler?

It shouldn't that's the bug.

> The two have no necessary connection between them. I think it's a bug if it did inherit the environment variables set by the compiler. This shouldn't be polymorphism.

Agreed.  You're describing what the compiler is doing today.  This PR fixes it so that the executable is run from the original environment before the compiler changed it during compilation.

marler8997:
To be fair, I read the description and could see how the purpose could be confused.  I updated the description to be more clear.

> The solution here is to maintain the original environment (the environment before it was changed by the compiler) when dmd -run executes the compiled program. This PR accomplishes this by saving the original values of environment variables when they are overwritten and then restoring these values before executing the compiled program.

src/dmd/link.d
```diff
@@ -1053,12 +1048,15 @@ version (Windows)
...
 
-                    char* npath = cast(char*)mem.xmalloc(5 + pathlen + 1 + addpathlen + 1);
+                    const length = 5 + addpathlen + 1 + pathlen;
+                    char* npath = cast(char*)mem.xmalloc(length + 1);
                     memcpy(npath, "PATH=".ptr, 5);
                     memcpy(npath + 5, addpath, addpathlen);
                     npath[5 + addpathlen] = ';';
                     memcpy(npath + 5 + addpathlen + 1, path, pathlen + 1);
-                    putenv(npath);
+                    if (0 != LocalEnvVar(cast(string)npath[0 .. length], 4).putenvRestorable)
```
thewilsonator: please don't use the yodel style

thewilsonator:
Apart from the nit, this looks good.

src/dmd/link.d
marler8997: Actually in this cast, `putenvRestorable` is a member of `LocalEnvVar`

thewilsonator: I meant remove the `0 !=`

marler8997: gotcha, done

marler8997:
I emailed Walter and he just got back to me saying he's going to take another look at this PR after I clarified what it was doing.
> Thanks for clearing up my confusion. I'll see about getting to it. -Walter

marler8997:
Still waiting on @WalterBright  to review.  I've emailed him a couple times and he says he hasn't had time to review yet.

In the meantime, since #9848 was merged, I've converted the new BASH test I added in this PR to D.

RazvanN7 (approved):

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+*/
+int putenvRestorable(const(char)[] name, const(char)[] value) nothrow
+{
+    auto var = LocalEnvVar.xmalloc(name, value);
+    const result = var.putenvRestorable();
+    var.xfree(result);
+    return result;
+}
+
+/// Holds a `VAR=value` string that can be put into the global environment.
```
WalterBright: It's "name=value".

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+struct LocalEnvVar
+{
+    string nameValueCStr;        // The argument passed to `putenv`.
```
WalterBright: Note it's called "nameValue" here, not "VARvalue"

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+import dmd.root.array;
+import dmd.root.rmem;
+import dmd.utils;
+
+version (Windows)
+    private extern (C) int putenv(const char*) nothrow;
+
+/**
+Construct a variable from `name` and `value` and put it in the environment while saving
+the current value of the environment variable.
```
WalterBright: No `Params:` section. Function does not construct a variable. Should return a bool.

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+Returns:
+    0 on success, non-zero on failure
+*/
+int putenvRestorable(const(char)[] name, const(char)[] value) nothrow
+{
+    auto var = LocalEnvVar.xmalloc(name, value);
+    const result = var.putenvRestorable();
```
WalterBright: Having two different functions with the same name is confusing.

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+    auto name() const { return nameValueCStr[0 .. equalsIndex]; }
+
+    /// The value of the variable
+    auto value() const { return nameValueCStr[equalsIndex + 1 .. $]; }
+
+    /**
+    Put this variable in the environment while saving the current value of the
+    environment variable.
+    Returns:
+        0 on success, non-zero on failure
```
WalterBright: use bool

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+
+    /// The name of the variable
+    auto name() const { return nameValueCStr[0 .. equalsIndex]; }
+
+    /// The value of the variable
+    auto value() const { return nameValueCStr[equalsIndex + 1 .. $]; }
+
+    /**
+    Put this variable in the environment while saving the current value of the
+    environment variable.
```
WalterBright: save it where?

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+        0 on success, non-zero on failure
+    */
+    int putenvRestorable() const nothrow
+    {
+        RestorableEnv.save(name);
+        return .putenv(cast(char*)nameValueCStr.ptr);
+    }
+
+    /**
+    Allocate a new variable via xmalloc that can be promoted to the global environment.
```
WalterBright: `promoted`? more like `added`

src/dmd/env.d
WalterBright: Note in the comment that it has a terminating 0 past the end of the string, not as part of the name.

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+        version (Windows)
+            doFree = true;
+        else
+        {
+            // on posix, when putenv succeeds ownership of the memory is transferred
+            // to the global environment
+            doFree = (putenvResult != 0);
+        }
+        if (doFree)
+            mem.xfree(cast(void*)nameValueCStr.ptr);
```
WalterBright: get rid of `doFree` and you can collapse this to 4 lines of code instead of 7.

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+    /// Holds the original values of environment variables when they are overwritten.
+    private __gshared Array!LocalEnvVar originalVars;
+
+    /// Restore the original environment.
+    static void restore()
+    {
+        foreach (var; originalVars)
+        {
+            if (0 != putenv(cast(char*)var.nameValueCStr))
+                assert(0, "putenv failed");
```
WalterBright: Really don't need elaborate assert's in DMD. If people can't find where an assert failed from the stack trace that gives file/line, they shouldn't be working on DMD.

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+{
+    /// Holds the original values of environment variables when they are overwritten.
+    private __gshared Array!LocalEnvVar originalVars;
+
+    /// Restore the original environment.
+    static void restore()
+    {
+        foreach (var; originalVars)
+        {
+            if (0 != putenv(cast(char*)var.nameValueCStr))
```
WalterBright: Not a fan of `0 !=`. We can expect some minimal knowledge of how boolean conditions work in D for people working on DMD.

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+
+    /**
+    Allocate a new variable via xmalloc that can be promoted to the global environment.
+    Params:
+        name = name of the variable
+        value = value of the variable
+    Returns:
+        a newly allocated variable that can be promoted to the global environment
+    */
+    static LocalEnvVar xmalloc(const(char)[] name, const(char)[] value) nothrow
```
WalterBright: Calling this `xmalloc` is confusing it with `Mem.xmalloc`. Same with `xfree`.

src/dmd/env.d
```diff
@@ -0,0 +1,113 @@
...
+    /// Save the environment variable `name` if not saved already.
+    static void save(const(char)[] name) nothrow
+    {
+        foreach (var; originalVars)
+        {
+            if (name == var.name)
+                return; // already saved
+        }
+        originalVars.push(LocalEnvVar.xmalloc(name,
+            name.toCStringThen!(n => getenv(n.ptr)).toDString));
```
WalterBright: There seems to be a lot of converting `char*` to `[]` and back again. Wouldn't it be simpler to just leave them as `char*`?

marler8997: > There seems to be a lot of converting char* to [] and back again
Alot? This is the one and only place in this change that we're calling `toCStringThen`.

We're using the `char[]` version of `name` for both the comparison and to pass to the `LocalEnvVar` allocation function.  We only need the `char*` version to call `getenv`.  I'm not sure what you'd like me to change here.

marler8997:
@WalterBright I've integrated all your suggestions except the one about `toCStringThen` since I'm not sure what you want for that suggestion.  Please let me know what you'd like me to do next here.

WalterBright:
Thank you. In thinking about it some more, what's happening is simply storing `name=value` pairs, and avoiding making duplicate entries for `name`. This is exactly what D associative arrays do. You can dispense with the new structs, and just use:
```
__gshared string[string] envNameValues;
```
Use the `in` operator to test for membership, and `foreach` to iterate over `envNameValues`.

marler8997:
Ok, I've changed the code per your suggestions.

WalterBright:
Much better.

marler8997:
Thanks alot for seeing this through to the finish line.

PR #8123, opened 2018-04-04
[Experiment] Use auto-bootstrapping on the auto-tester

wilzbach:
This is an experiment to test whether this mechanism would work to be able to bump the host version.

See also: https://github.com/dlang/dmd/pull/8112


wilzbach:
Experiment successful - so the auto-bootstrapping would work (on Posix systems at least):

image

Sorry for the noise.

PR #8124, opened 2018-04-04
[BLOCKED][WIP] Don't instantiate unittest templates in libraries

marler8997:
> NOTE: This change no longer has a functional impact on users other than they should see a performance increase when compiling with `-unittest`.

### DIFF
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -1441,6 +1441,21 @@ struct ASTBase
             srcfile = new File(srcfilename);
         }
 
+        /**
+        A root module is one that will be compiled to object code.
+        */
+        final bool isRoot() const
+        {
+            return true;
+        }
+
+        /**
+        Called by the parser after the module declaration has been determined.
+        */
+        final void afterModuleDeclaration(ModuleDeclaration* md)
+        {
+        }
+
         override void accept(Visitor v)
         {
             v.visit(this);
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -292,10 +292,10 @@ extern (C++) final class Module : Package
     extern (C++) static __gshared Dsymbols deferred3;   // deferred Dsymbol's needing semantic3() run on them
     extern (C++) static __gshared uint dprogress;       // progress resolving the deferred list
     /**
-     * A callback function that is called once an imported module is
-     * parsed. If the callback returns true, then it tells the
-     * frontend that the driver intends on compiling the import.
-     */
+    A callback function that is called when a module is loaded as soon as its
+    module declaration has been parsed. If the callback returns true, then it tells the
+    frontend that the driver intends on compiling the import.
+    */
     extern (C++) static __gshared bool function(Module mod) onImport;
 
     static void _init()
@@ -370,12 +370,10 @@ extern (C++) final class Module : Package
     int searchCacheFlags;       // cached flags
 
     /**
-     * A root module is one that will be compiled all the way to
-     * object code.  This field holds the root module that caused
-     * this module to be loaded.  If this module is a root module,
-     * then it will be set to `this`.  This is used to determine
-     * ownership of template instantiation.
-     */
+    This field holds the root module that caused this module
+    to be loaded.  If this module is a root module, then it will
+    be set to `this`.
+    */
     Module importedFrom;
 
     Dsymbols* decldefs;         // top level declarations for this Module
@@ -515,41 +513,6 @@ extern (C++) final class Module : Package
             message("import    %s", buf.peekString());
         }
         m = m.parse();
-
-        // Call onImport here because if the module is going to be compiled then we
-        // need to determine it early because it affects semantic analysis. This is
-        // being done after parsing the module so the full module name can be taken
-        // from whatever was declared in the file.
-
-        //!!!!!!!!!!!!!!!!!!!!!!!
-        // Workaround for bug in dmd version 2.068.2 platform Darwin_64_32.
-        // This is the compiler version that the autotester uses, and this code
-        // has been carefully crafted using trial and error to prevent a seg fault
-        // bug that occurs with that version of the compiler.  Note, this segfault
-        // does not occur on the next version of dmd, namely, version 2.069.0. If
-        // the autotester upgrades to that version, then this workaround can be removed.
-        //!!!!!!!!!!!!!!!!!!!!!!!
-        version(OSX)
-        {
-            if (!m.isRoot() && onImport)
-            {
-                auto onImportResult = onImport(m);
-                if(onImportResult)
-                {
-                    m.importedFrom = m;
-                    assert(m.isRoot());
-                }
-            }
-        }
-        else
-        {
-            if (!m.isRoot() && onImport && onImport(m))
-            {
-                m.importedFrom = m;
-                assert(m.isRoot());
-            }
-        }
-
         Target.loadModule(m);
         return m;
     }
@@ -853,7 +816,6 @@ extern (C++) final class Module : Package
             scope p = new Parser!ASTCodegen(this, buf[0 .. buflen], docfile !is null);
             p.nextToken();
             members = p.parseModule();
-            md = p.md;
             numlines = p.scanloc.linnum;
             if (p.errors)
                 ++global.errors;
@@ -1288,9 +1250,33 @@ extern (C++) final class Module : Package
         return false;
     }
 
-    bool isRoot()
+    /**
+    A root module is one that will be compiled to object code.  This is used to
+    determine ownership of template instantiation and when to compile unittests.
+    Returns:
+        true if this is a root module
+    */
+    final bool isRoot() const
     {
         return this.importedFrom == this;
+    }
+
+    /**
+    Called by the parser after the module declaration has been determined.
+    */
+    final void afterModuleDeclaration(ModuleDeclaration* md)
+    {
+        this.md = md;
+
+        // Call onImport here because if the module is going to be compiled then we
+        // need to determine it early because it affects whether unittests will be parsed
+        // and semantic analysis. This is being done right after parsing the module delclaration
+        // so the full module name can be taken from whatever was declared in the file.
+        if (!isRoot() && onImport && onImport(this))
+        {
+            importedFrom = this;
+            assert(isRoot());
+        }
     }
 
     // true if the module source file is directly
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -6360,8 +6360,7 @@ extern (C++) class TemplateInstance : ScopeDsymbol
          * is behind a conditional debug declaration.
          */
         // workaround for https://issues.dlang.org/show_bug.cgi?id=11239
-        if (global.params.useUnitTests ||
-            global.params.debuglevel)
+        if (global.params.debuglevel)
         {
             // Prefer instantiations from root modules, to maximize link-ability.
             if (minst.isRoot())
@@ -7354,7 +7353,7 @@ extern (C++) class TemplateInstance : ScopeDsymbol
     {
         Module mi = minst; // instantiated . inserted module
 
-        if (global.params.useUnitTests || global.params.debuglevel)
+        if (global.params.debuglevel)
         {
             // Turn all non-root instances to speculative
             if (mi && !mi.isRoot())
--- a/test/runnable/extra-files/printVersionUnittest.d
+++ b/test/runnable/extra-files/printVersionUnittest.d
@@ -1,0 +1,6 @@
+module printVersionUnittest;
+
+version(unittest)
+{
+    pragma(msg, "version unittest");
+}
--- a/test/runnable/selectiveUnittests.sh
+++ b/test/runnable/selectiveUnittests.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+
+obj_file=${OUTPUT_BASE}/tempobj${OBJ}
+exe_file=${OUTPUT_BASE}/tempexe${EXE}
+
+# Compile both modules, make sure unittest compiles/runs
+unittest_output="unittesting printingUnittest"
+$DMD -m${MODEL} -of=${exe_file} -I=${EXTRA_FILES} -unittest ${EXTRA_FILES}/printingUnittest.d ${EXTRA_FILES}/selectiveUnittests.d
+${exe_file} | grep -q "${unittest_output}"
+./runnable/extra-files/check_bin.sh ${exe_file} HAS __unittest
+
+# Precompile the module and make sure it's unittest doesn't get compiled/run
+$DMD -m${MODEL} -c -of=${obj_file} ${EXTRA_FILES}/printingUnittest.d
+./runnable/extra-files/check_bin.sh ${exe_file} DOES_NOT_HAVE __unittest
+
+rm -f ${exe_file}
+$DMD -m${MODEL} -of=${exe_file} -I=${EXTRA_FILES} -unittest ${obj_file} ${EXTRA_FILES}/selectiveUnittests.d
+${exe_file} | grep -q -v "${unittest_output}"
+
+# Make sure that version(unittest) still get analyzed
+unittest_output="version unittest"
+$DMD -m${MODEL} -I=${EXTRA_FILES} -o- -unittest ${EXTRA_FILES}/printVersionUnittest.d ${EXTRA_FILES}/selectiveUnittests2.d 2>&1 | grep -q "${unittest_output}"
+
+# Precompile the module and make sure version(unittest) is still enabled
+$DMD -m${MODEL} -c -of=${obj_file} ${EXTRA_FILES}/printVersionUnittest.d
+$DMD -m${MODEL} -I=${EXTRA_FILES} -o- -unittest ${obj_file} ${EXTRA_FILES}/selectiveUnittests2.d 2>&1 | grep -q "${unittest_output}"
+
+rm -f ${obj_file}
+rm -f ${exe_file}
--- a/src/dmd/glue.d
+++ b/src/dmd/glue.d
@@ -752,8 +752,17 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
     if (!fd.fbody)
         return;
 
+    // Find module m for this function
+    Module m = null;
+    for (Dsymbol p = fd.parent; p; p = p.parent)
+    {
+        m = p.isModule();
+        if (m)
+            break;
+    }
+
     UnitTestDeclaration ud = fd.isUnitTestDeclaration();
-    if (ud && !global.params.useUnitTests)
+    if (ud && (!global.params.useUnitTests || !m.isRoot))
         return;
 
     if (multiobj && !fd.isStaticDtorDeclaration() && !fd.isStaticCtorDeclaration())
@@ -885,15 +894,6 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
 
     symtab_t *symtabsave = cstate.CSpsymtab;
     cstate.CSpsymtab = &f.Flocsym;
-
-    // Find module m for this function
-    Module m = null;
-    for (Dsymbol p = fd.parent; p; p = p.parent)
-    {
-        m = p.isModule();
-        if (m)
-            break;
-    }
 
     Dsymbols deferToObj;                   // write these to OBJ file later
     Array!(elem*) varsInScope;
--- a/src/dmd/module.h
+++ b/src/dmd/module.h
@@ -144,7 +144,7 @@ public:
     static void clearCache();
     int imports(Module *m);
 
-    bool isRoot() { return this->importedFrom == this; }
+    bool isRoot() const { return this->importedFrom == this; }
     // true if the module source file is directly
     // listed in command line.
     bool isCoreModule(Identifier *ident);
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -377,6 +377,8 @@ final class Parser(AST) : Lexer
             }
         }
 
+        mod.afterModuleDeclaration(md);
+
         decldefs = parseDeclDefs(0, &lastDecl);
         if (token.value != TOK.endOfFile)
         {
@@ -547,7 +549,7 @@ final class Parser(AST) : Lexer
                     break;
                 }
             case TOK.unittest_:
-                if (global.params.useUnitTests || global.params.doDocComments || global.params.doHdrGeneration)
+                if ( (global.params.useUnitTests || global.params.doDocComments || global.params.doHdrGeneration) && mod.isRoot)
                 {
                     s = parseUnitTest(pAttrs);
                     if (*pLastDecl)
--- a/test/runnable/extra-files/check_bin.sh
+++ b/test/runnable/extra-files/check_bin.sh
@@ -1,0 +1,17 @@
+#!/usr/bin/env bash
+
+if [ "$2" == "HAS" ]; then
+    GREP_OPTION=
+elif [ "$2" == "DOES_NOT_HAVE" ]; then
+    GREP_OPTION=-v
+else
+    echo Unknown option "$2"
+    exit 1
+fi
+
+if [ ${OS} != "linux" ]; then
+    echo Skipping checkbin $2 $3 because os $OS is not linux
+    exit 0
+fi
+
+objdump -t "$1" | grep -q $GREP_OPTION "$3"
--- a/test/runnable/extra-files/printingUnittest.d
+++ b/test/runnable/extra-files/printingUnittest.d
@@ -1,0 +1,7 @@
+module printingUnittest;
+
+unittest
+{
+    import core.stdc.stdio;
+    printf("unittesting printingUnittest\n");
+}
--- a/test/runnable/extra-files/selectiveUnittests.d
+++ b/test/runnable/extra-files/selectiveUnittests.d
@@ -1,0 +1,6 @@
+import printingUnittest;
+void main()
+{
+    // have to print something so grep works
+    import core.stdc.stdio; printf("done\n");
+}
--- a/test/runnable/extra-files/selectiveUnittests2.d
+++ b/test/runnable/extra-files/selectiveUnittests2.d
@@ -1,0 +1,4 @@
+import printVersionUnittest;
+void main()
+{
+}


src/dmd/dmodule.d
```diff
@@ -377,6 +377,7 @@ extern (C++) final class Module : Package
      * ownership of template instantiation.
      */
     Module importedFrom;
+    bool isCompiledImport;
```
marler8997: This PR determines if an import should be compiled earlier than before (right after parsing the module declaration instead of after parsing the whole file).  This way, we know if we are going to compile unittests before parsing the file.  To that end, I found out that the parser depends on `importedFrom` being set correctly meaning we can no longer "hijack" this field by resetting it when an imported module is compiled.  The solution is to add a new field that indicates an imported module should be compiled, and to leave the `importedFrom` field intact so it can be used during the parse stage.

andralex:
cc @WalterBright @atilaneves @schveiguy

schveiguy:
Very cool! A few things:

1. How does this affect `version(unittest)`? My understanding is that this will still compile those in, as it seems you are only skipping parsing/semantic on `unittest` blocks.
2. It has been suggested in the forums that `__traits(getUnitTests)` override the fact that it's not a root, but the way you have done it makes this very difficult. But I may not understand the way the compiler works here.

On the other hand, `__traits(getUnitTests)` may work just fine for modules that are actually passed on the command line (which may be all @atilaneves needs).

schveiguy:
> How does this affect `version(unittest)`?

Did more research, indeed the unittest version is more blunt, it determines if the scope should be included based on the global flag only. I can think of a way to make it work more like I had expected, but now I'm not sure if that's the correct course.

There are cases when `version(unittest)` alters the layout of a type, but this really is a bad idea in most cases. `version(unittest)` is bound to be rare, and it may be bad to set the precedent of having a version that isn't global for the compilation.

The one case where I'm concerned is when `version(unittest)` does extra imports, and therefore causes a slowdown. However, we can migrate code away from that and towards using local imports inside unittests (which would be elided by this PR). It's a different case for the unittests themselves, as those are functions that definitely aren't going to be emitted for modules that are simply imported.

marler8997:
We could make `version(unittest)` be true only when the module's unittests are being compiled, and then create a new version for the global unittest flag `version(some_unittests_enabled)`?

jacob-carlborg:
If I understand correctly, the `unittest` blocks are basically implemented as functions. At some point the compiler will merge all those functions into one function and make that, combined unit test function, available in the module info for later use at runtime.

An alternative approach that, I think, would not affect `__traits(getUnitTests)` would be to compile everything as usual, but don't create the combined unit test function.

marler8997:
The slowdown is in analyzing the unittests, not "combining them as you describe".  And I don't think it "combines" them into one function.  I think it just calls all of them in succession.

jacob-carlborg:
> The slowdown is in analyzing the unittests, not "combining them/generating code" as you describe.

Yes. Not sure what I was thinking.

> And I don't think it "combines" them into one function. I think it leaves them as separate functions and just calls them in succession.

That might be the case.

marler8997:
Or maybe we reverse the logic, have `version(unittest)` mean that we are compiling with the `-unittest` flag...and then create `version(unittesting_this_module)` mean that the compiler is compiling unittests for the current module.  This is a distinction we should support because there are times when many unittests share the same common functions or data structures.

jmdavis:
I don't know exactly what we want to do with `version(unittest)`, but I can guarantee that if we make it so that `version(unittest)` is not compiled in when the module is imported, then existing code will break - e.g. when I've had helpers for unit tests, I've sometimes put them in a separate module where the unit tests can then import them, and I've generally put them in a `version(unittest)` block, because they're intended to only be used as part of the unit tests. So, blindly making it so that `version(unittest)` isn't compiled in when the module is imported will break code.

Now, we can choose to do something like print a deprecation message if you ever import anything that's in a `version(unittest)` block (though I don't know how easy that would be to implement), and then later, we can treat `version(unittest)` as only applying when the module is directly compiled with `-unittest`, but I'm opposed to simply changing the behavior of `version(unittest)` without some sort of deprecation mechanism.

Regardless, if we want to make it so that `version(unittest)` only affects the module being compiled and not those it's importing, I don't think that we then need to add a version identifier for unit tests that affects all modules. Any project looking to do something like that can create its own version identifier, and in that case, it won't spill over into other projects.

marler8997:
@jmdavis 

Yeah, this PR takes the conservative approach, leaving in`version(unittest)` for now.  The idea is maybe later we could add another `version(unittesting_this_module)` that would only get instantiated if both the `-unittest` flag is provide AND unittests are being compiled for this module.  This is where you could define code that is common between unittests in the same module.

jmdavis:
> The slowdown is in analyzing the unittests, not "combining them/generating code" as you describe. And I don't think it "combines" them into one function. I think it leaves them as separate functions and just calls them in succession.

If you have an assertion fails in a `unittest` block, and you look at the stack, then you can see the functions involved, and the function for the `unittest` block includes the line number, clearly indicating that there is indeed a function specifically for that specific `unittest block`. e.g.

    core.exception.AssertError@q.d(8): unittest failure
    ----------------
    ??:? _d_unittestp [0x42be75]
    ??:? void q.__unittest_L5_C1() [0x42bced]
    ??:? void q.__modtest() [0x42bd90]
    ??:? int core.runtime.runModuleUnitTests().__foreachbody2(object.ModuleInfo*) [0x42e77b]
    ??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)).__lambda2(immutable(object.ModuleInfo*)) [0x4351a6]
    ??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))).__foreachbody2(ref rt.sections_elf_shared.DSO) [0x444a75]
    ??:? int rt.sections_elf_shared.DSO.opApply(scope int delegate(ref rt.sections_elf_shared.DSO)) [0x4329e0]
    ??:? int rt.minfo.moduleinfos_apply(scope int delegate(immutable(object.ModuleInfo*))) [0x444a01]
    ??:? int object.ModuleInfo.opApply(scope int delegate(object.ModuleInfo*)) [0x43517d]
    ??:? runModuleUnitTests [0x42e559]
    ??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).runAll() [0x42c190]
    ??:? void rt.dmain2._d_run_main(int, char**, extern (C) int function(char[][])*).tryExec(scope void delegate()) [0x42c11b]
    ??:? _d_run_main [0x42c086]
    ??:? main [0x42bda1]
    ??:? _start [0x42bbae]
    ??:? ??? [0x68dfff]
    1/1 unittests FAILED

src/dmd/dmodule.d
```diff
@@ -370,12 +370,10 @@ extern (C++) final class Module : Package
...
-     * A root module is one that will be compiled all the way to
-     * object code.  This field holds the root module that caused
-     * this module to be loaded.  If this module is a root module,
-     * then it will be set to `this`.  This is used to determine
-     * ownership of template instantiation.
-     */
+    This field holds the root module that caused this module
+    to be loaded.  If this module is a root module, then it will
+    be set to `this`.
+    */
```
atilaneves: Unless it's no longer true, I'd keep the part of the comment that says "This is used to determine the ownership of templaet instantiation".

test/runnable/extra-files/printVersionUnittest.d
```diff
@@ -0,0 +1,6 @@
+module printVersionUnittest;
```
atilaneves: If the next module is `printingUnittest`, this one should be `pragmaMsgVersionUnittest`, no?

test/runnable/selectiveUnittests.sh
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+unittest_output="unittesting printingUnittest"
+
+$DMD -m${MODEL} -I=${EXTRA_FILES} -unittest ${EXTRA_FILES}/printingUnittest.d -run ${EXTRA_FILES}/selectiveUnittests.d | grep -q "${unittest_output}"
```
atilaneves: The error code for this isn't 0, but it won't cause the whole script to fail.

atilaneves:
There is no documentation (or, as far as I can see, tests) that show how this impacts `__traits(getUnittests)`. Does it still work? Does it still work with separate compilation? I assume so because the current tests still pass. I'm going to build this branch and play with it.

src/dmd/dmodule.d
marler8997: The information was just moved to the `isRoot` function.

test/runnable/selectiveUnittests.sh
marler8997: ah...I see you're not familiar with the latest work on DMD's Bash tests...see `test/tools/sh_do_test.sh` :)

test/runnable/extra-files/printVersionUnittest.d
marler8997: meh, not a big deal, kind of a long name

marler8997:
Haven't tried `__traits(getUnittests)` yet, come back with what you find

schveiguy:
> how this impacts `__traits(getUnittests)`

My expectation is that because the adjustment is at the same place where -unittest is checked, it will be as if that module was compiled without -unittest (as far as I can tell, the whole AST branch is not processed).

In other words, I don't think the unittest will be properly compiled, so I don't think `__traits(getUnittests)` will work for modules that are simply imported.

However, because of the way this is done, if the module in question *is* being compiled by DMD, then the unittests will be available. I think separate compilation will work in some cases but not in others. Depending on how you implement your testing framework, it might be OK. If you instantiate a template that uses `__traits(getUnittests)` inside the module you are currently compiling, it should be fine.

I suppose the best way to figure it out is to try it :)

schveiguy:
I think `version(unittest)` should be left the way it is -- compile it in if -unittest is passed on the command line, regardless of whether the module is a root. This is a change from what I originally thought, but now I think it's more dangerous to create a situation where a version statement compiles or not based on whether the module is passed on the command line.

atilaneves:
I tried this build and it doesn't seem to do anything that I can notice. __traits(getUnitTests) still works in separate compilation, and std.path still takes forever to compile with `-unittest`.

src/dmd/astbase.d
```diff
@@ -1441,6 +1441,21 @@ struct ASTBase
             srcfile = new File(srcfilename);
         }
 
+        /**
+        A root module is one that will be compiled to object code.
+        */
+        final bool isRoot() const
+        {
+            return true;
```
schveiguy: When does isRoot return false for a module such that its unittests don't compile?

marler8997: isRoot will return false if the module isn't "taken" to object code (see comment).  This will happen if the module is not passed on the command line and is not included as a "compiled import"

marler8997:
@atilaneves 

What I'm guessing is happening is that the unittest "functions" are still getting generated, but are empty. This would mean that `__traits(getUnitTests)` would still contain the list of all unittest functions from all modules, it's just that most of them would be empty.

As to why the performance is still bad, my best guess is that `version(unittest)` blocks are to blame.  The bad compile-time performance is likely due to a few symbols being loaded, so a single import can be the cause of massive slowdown.  We've been discussing ways to solve that problem as well.  One idea is to introduce a new version, something like `version(unittesting_this_module)` that is only true if both unittests are enabled AND the unittests for the current module are being compiled.  This version could be used if the only dependencies of the code within the block are unittests in the current module.  That would make `version(unittest)` useful if a module was providing "unittest support" for code outside the module.

Atila I'm not sure if you tried this, but were you able to see if unittests from precompiled modules were included or not?  With this PR they definitely should not be.  You could add a "print" statement inside one of them to check.

test/runnable/selectiveUnittests.sh
```diff
@@ -0,0 +1,20 @@
...
+
+obj_file=${OUTPUT_BASE}/temp${OBJ}
+
+# Precompile the module and make sure it's unittest doesn't get run
+$DMD -m${MODEL} -c -of=${obj_file} ${EXTRA_FILES}/printingUnittest.d
+$DMD -m${MODEL} -I=${EXTRA_FILES} -unittest ${obj_file} -run ${EXTRA_FILES}/selectiveUnittests.d | grep -q -v "${unittest_output}"
```
wilzbach: Maybe add something along the lines of:

compilable/extra-files/minimal/verify_symbols.sh
compilable/extra-files/test11237.sh
runnable/test17619.sh

Just grep for `__unittest` in the output of `nm` or `objdump`

jacob-carlborg:
If `version(unittest)` might be the problem it should be straight forward to modify the compiler to not set that version and see if it makes any difference.

marler8997:
@jacob-carlborg good suggestions.

It looks like removing the `version(unittest)` blocks helped a little bit, but still doesn't get us all the way there:

`importpath.d`:
```
import std.path;
```

```
dmd importpath.d
user .276s
(with version(unittest)) dmd -unittest importpath.d
user .648s
(without version(unittest)) dmd -unittest importpath.d
user .756s
```

marler8997:
Found the problem. In `dtemplate.d` around line 7360, there is a different code path taken when unittests are enabled.  It looks like this code "moves" template instances from "non-root" modules to "root" modules so code will be generated for them.  What's interesting is that this is only done when `-unittest` is enabled or when `-debug` is enabled.

atilaneves:
@marler8997 I just updated to your last commit from 3 hours ago, rebuilt dmd, druntime and phobos, and everything still works with __traits(getUnitTests) and separate compilation. I created a [gist](https://gist.github.com/atilaneves/760cfc46f81005ed0d9503cd89c49acd) with the files I used.

Don't get me wrong, breaking that gist would annoy the hell out of me. I'm just not sure what this PR *does*.

wilzbach:
> breaking that gist would annoy the hell out of me.

That's Atila-speak for "please add my gist to the testsuite" ;-)

atilaneves:
@wilzbach Hah! It's already there as [a dmd test](https://github.com/dlang/dmd/blob/master/test/runnable/issue16995.d).

src/dmd/astbase.d
schveiguy: I guess I'm just asking because this looks pretty constant (final, returns true). What is the type of AST class that represents an import module?

schveiguy: OK, I'm seeing that this is duplicated in `dmodule.d`, but with a different implementation (one that makes sense). I'm not sure what the `astbase.d` file does, but it sure seems like it's doing something different than `dmodule.d`. What is this for?

schveiguy:
huh... So I was downloading this PR to see if I can play with the differences, and I've discovered a few things with :

1. unittests in imported modules are ALREADY NOT COMPILED. That is, they are parsed, but not actually semantic analyzed. This is what I had assumed the issue was, but it's obviously not.
2. version(unittest) is compiled (obviously), so that will add more code/imports.

I feel it's strange that 1 occurs, I thought that was the point of checking for "isRoot"?

The rudimentary test I did was create 2 modules, one imports the other. In the imported module, I create a unittest that simply imports std.stdio. If you compile the first module with -unittest, std.stdio is NOT processed. If compile both, it IS processed.

Now I'm not sure why the original problem occurs. Do we have that many `version(unittest)` items in phobos?

> Found the problem. In dtemplate.d around line 7360, there is a different code path taken when unittests are enabled.

I don't know what this means. It moves templates to the root module? I know that if you instantiate a template, it needs to move the whole template to your local module, just in case it's not instantiated anywhere else.

At this point, I'm super-confused. I thought I knew what the problem is, but obviously I don't.  So I'm going to fade into the background here.

marler8997:
> @marler8997 I just updated to your last commit from 3 hours ago, rebuilt dmd, druntime and phobos, and everything still works with __traits(getUnitTests) and separate compilation. I created a gist with the files I used.

Yes, I can see why `__traits(getUnitTest)` still works. Checkout `parse.d` around line 590.
```D
                    // Workaround 14894. Add an empty unittest declaration to keep
                    // the number of symbols in this scope independent of -unittest.
                    s = new AST.UnitTestDeclaration(loc, token.loc, AST.STC.undefined_, null);
```

marler8997:
> Now I'm not sure why the original problem occurs. Do we have that many version(unittest) items in phobos?

It only takes one extra import in a `version(unittest)` to kill compilation time.

schveiguy:
And most of them can be eliminated. This pattern is in there a lot:

```D
version(unittest) // can be replaced easily with unittest
{
    static assert(localTemplate!(args) == expected);
}
```

atilaneves:
@schveiguy Yeah, those should definitely just be replaced with a `unittest` block.

schveiguy:
I'm working on eliminating all version(unittest) imports in Phobos. I think this is going to have a good effect, even without this PR. Will let you know.

marler8997:
With the latest change, I'm getting some undefined symbol errors during the "test phobos" stage.

Missing Symbol (demangled):
```D
pure nothrow @nogc @safe std.uni.SliceOverIndexed!(std.uni.Grapheme).SliceOverIndexed std.uni.SliceOverIndexed!(std.uni.Grapheme).SliceOverIndexed.opSlice()
```

If I dump the symbols in "uni.o" I can see this symbol, the only difference being that it is decorated with the `return` attribute
```D
pure nothrow @nogc return @safe std.uni.SliceOverIndexed!(int[]).SliceOverIndexed std.uni.SliceOverIndexed!(int[]).SliceOverIndexed.opSlice()
```

it's returning `auto` so it looks like the compiler might be interpreting it differently in the "template instantiation" and where it's used in the unittest...hmmmm

wilzbach:
That looks like the issue with `-dip1000` changing the mangling of the symbols.
Many unittests are already compiled with `-dip1000`, so that's my bet.

See also: https://github.com/dlang/phobos/pull/5915#issuecomment-350425051

schveiguy:
Welp, I changed all the version(unittest) so that they don't include any extra files. And it had zero effect. From what I can tell, it imports the same number of files, and the time for compiling the std.path unittests are just as bad as before.

I'm completely stumped, my simple test shows that it doesn't follow imports inside unit tests. I suppose the only thing left is to rip apart std.path until it starts making a difference, just to figure out what the trigger is.

marler8997:
@schveiguy 

It's already figured out.

You need 3 things.

1. Don't compile unittests (this PR does that)
2. Don't move template instances to root modules (this PR does that, make sure you have the current version)
3. Don't compile `version(unittest)` (there's a line in mars.d you can comment out to do this)

If you do this, then importing `std.path` will perform the same with and without `-unittest`.

src/dmd/dmodule.d
```diff
@@ -1288,11 +1250,35 @@ extern (C++) final class Module : Package
...
+
+        // Call onImport here because if the module is going to be compiled then we
+        // need to determine it early because it affects whether unittests will be parsed
+        // and semantic analysis. This is being done right after parsing the module delclaration
+        // so the full module name can be taken from whatever was declared in the file.
+        if (!isRoot() && onImport && onImport(this))
+        {
+            importedFrom = this;
+            assert(isRoot());
+        }
```
ibuclaw: You could move `Target.loadModule` here also.

schveiguy:
@marler8997 What I'm stumped on is how to recreate the problem in a simple test.

1. How do I make the (original) compiler compile unittests in imported modules? I can't seem to do it with a simple test.
2. How do I make the compiler "move template instances to root modules"? I don't even know what you mean by that. What I'm looking for is a small example that shows what you mean (i.e. forces the compiler to compile imported unit tests).
3. I'm very concerned about not compiling version(unittest) code. I think we should instead focus on making sure phobos version(unittest) code doesn't do anything unexpected (what I've been trying to do, but I'm not seeing any affect).

src/dmd/dtemplate.d
```diff
@@ -6360,8 +6360,7 @@ extern (C++) class TemplateInstance : ScopeDsymbol
          * is behind a conditional debug declaration.
          */
         // workaround for https://issues.dlang.org/show_bug.cgi?id=11239
-        if (global.params.useUnitTests ||
-            global.params.debuglevel)
+        if (global.params.debuglevel)
```
ibuclaw: Did you test the linked bug after making this change? Doesn't look like the CI gets so far along to run it itself.

marler8997:
@schveiguy 

I'm missing some details, could you provide the compiler commands you used so I can understand what you're doing?

marler8997:
> 1. How do I make the (original) compiler compile unittests in imported modules? I can't seem to do it with a simple test.

Don't know the answer to this.  It could be that unittests were already not being compiled...but you need all 3 criteria I listed in order to have good "unittest import" performance (not just disabling unittest compilation).

> 2. How do I make the compiler "move template instances to root modules"? I don't even know what you mean by that. What I'm looking for is a small example that shows what you mean (i.e. forces the compiler to compile imported unit tests).

See this comment (https://github.com/dlang/dmd/pull/8124#issuecomment-379164250).  What's happening is when `-unittest` is enabled, template instances from imported modules are being moved into root modules, meaning there's ALOT more code being compiled when `-unittest` is enabled, not just the code inside `version(unittest)` and `unittest`.

> 3. I'm very concerned about not compiling version(unittest) code. I think we should instead focus on making sure phobos version(unittest) code doesn't do anything unexpected (what I've been trying to do, but I'm not seeing any affect).

Agreed.  I was just showing that if you disable `version(unittest)` then you completely solve the performance problem.  I'm working on ways to solve this `version(unittest)` problem (see https://github.com/dlang/dmd/pull/8147 for one idea)

src/dmd/dtemplate.d
marler8997: This is being tested and is passing on the autotester.  It's not until the "test phobos" stage that we get linker errors.

src/dmd/dmodule.d
marler8997: Does the module need to be fully parsed before calling that, or does it just need the module declaration parsed, or does it not need anything to be parsed?

marler8997: Wait...I see from the comment it says "post parsing".  Which means it should be called in this function since this function is calling "during parsing" right after parsing the module declaration.

marler8997: Might be good to rename this function to something like `onModuleParsed` in another PR?

schveiguy:
I'm trying to show a simple case where "look, this is how imports sneak into your compilation when you use -unittest". It's got to be something simpler than `std.path` and all of phobos to trigger the problem.

I'm not well versed in how the compiler works, so I'm very much struggling to understand what your changes are doing -- or that they are necessarily the correct solution. What I thought they were doing doesn't happen with my small example.

So here is my attempt to show that unittests increase compilation time by adding more imports:

importme.d:
```D
module importme;

unittest
{
    static assert(0, "bad");
    import std.stdio;
    writlen("testing!");
}
```

importer.d:
```D
import importme;
```

Now, I compile like this:

```
dmd -v -c importer.d
predefs   DigitalMars Posix OSX darwin LittleEndian D_Version2 all D_SIMD D_InlineAsm_X86_64 X86_64 D_LP64 D_PIC assert D_HardFloat D_ObjectiveC
binary    dmd
version   v2.079.0
config    /Users/steves/.dvm/compilers/dmd-2.079.0/osx/bin/dmd.conf
DFLAGS    -I/Users/steves/.dvm/compilers/dmd-2.079.0/osx/bin/../../src/phobos -I/Users/steves/.dvm/compilers/dmd-2.079.0/osx/bin/../../src/druntime/import -L-L/Users/steves/.dvm/compilers/dmd-2.079.0/osx/bin/../lib
parse     importer
importall importer
import    object	(/Users/steves/.dvm/compilers/dmd-2.079.0/osx/bin/../../src/druntime/import/object.d)
import    core.attribute	(/Users/steves/.dvm/compilers/dmd-2.079.0/osx/bin/../../src/druntime/import/core/attribute.d)
import    importme	(importme.d)
semantic  importer
semantic2 importer
semantic3 importer
code      importer
```
Notice how the static assert is NOT reached, importme isn't even compiled, and std.stdio isn't imported.

With -unittest, I get EXACTLY the same output and result.

So why does importing `std.path` not ignore unittests like this simple example? What is it that I have to add to `importme.d` such that compiling with -unittest makes a difference? That's what I'm struggling with. I thought it was adding a `version(unittest)` block, but I literally removed nearly ALL of them from Phobos, and have not seen any changes. Not a single import changed. Perhaps your knowledge of what the problems are can help you show me what to add.

jacob-carlborg:
@schveiguy I think you need to add a lot of templates to `importme`.

marler8997:
Looks like my phobos PR https://github.com/dlang/phobos/pull/6427 wasn't enough to fix all the mangling issues.  So this PR will be stalled until the DIP 1000 bugs are addressed.

schveiguy:
So, I have figured out (somewhat) the issues here, and we have this peculiar notion that when `-unittest` is passed on the command line, templates that appear to be handled by imported modules are "moved" to a root module, exactly as @marler8997 was saying.

An example:

importme.d:
```D
module importme;
struct S(T)
{
   T t;
   void method() {}
}

void baz(S!int) {}

void foo()() {
    S!int x;
    x.method();
}
```

importer.d:
```D
import importme;

void main() { foo(); }
```

Compile with `dmd importer.d`, and you get a linker failure -- `method` cannot be found. What I'm guessing happens here is that the compiler assumes that since importme has instantiated `S!int`, it doesn't need to put the code for `method` inside the importer.o object, since it should technically be in importme.o.

But when I execute `dmd -unittest importer.d` it links! This is because even though importme.d instantiates `S!int`, it's "moved" to the root module, meaning importer now contains the object code for `method`.

Indeed, compiling separately, both .o modules contain a definition of `method`. The linker simply discards the extra definition (I guess?).

Doing some research, I found that this "moving" actually ALWAYS used to happen without -unittest or -debug (even before this if statement was introduced). Here is the original pull that removed the "moving" situation for all code but certain cases (including unittests inexplicably): #2550

And here is the one that adds moving for `-debug` back into the mix: #2661

So IMO, I think we need to do 2 things:

1. Remove all cases in debug and unittesting code where druntime/phobos is instantiated with:
  a) different layout
  b) different symbols
2. Remove the special case "moving" of template instantiations into root modules for both -unittest and -debug.

These 2 fixes will help fix the extra imports due to template unittests in some cases, but not all. In cases where the instantiation is happening inside a root module, you will still see unittests be compiled (and even included).

Then I think we can move forward with a PR that kills semantic/emitting of unittests for imported modules. Preferably, this would allow `__traits(getUnittest)` to override this setting for a specific import.

I don't think we should disable `version(unittest)` parsing, as I think the effect would be too detrimental -- often the declarations inside these are helper functions that are useful outside the module that defines them. All we can do is clean our own house (phobos) and show people the correct way to do things.

src/dmd/dtemplate.d
schveiguy: @ibuclaw the original bug only had to do with `-debug` switch. As far as I can tell, we have never tried (before this PR) compiling code that doesn't move templates into root modules when `-unittest` is specified.

schveiguy:
Yeah, I think the mangling issue has to do with the lack of "moving" to root modules. Note that the debug build works! This is because -debug is passed, and you haven't turned off the moving for -debug.

What it looks like is happening is the "moving" is re-compiling the code as if it were in the root module, and this masks any incorrect symbols in the library object. So @marler8997 is correct that this is "uncovering" dip1000 issues, albeit probably only when doing separate compilation. It's kind of alarming that we have bugs in separate compilation, which is how we compile unittests for phobos, but we ship the code compiled all-at-once! And all of it is masked over by moving all template instantiations local 😝 

I wonder how much the autotester will speed up when it's properly ignoring imported templates it shouldn't be compiling.

wilzbach:
>  So @marler8997 is correct that this is "uncovering" dip1000 issues, albeit probably only when doing separate compilation.

For the record, this issue of separate mangling for dip1000 is "well-known" (at least among [Phobos developers](https://github.com/dlang/phobos/pull/5915#issuecomment-350425051)) and definitely a blocker for it (sadly not the only one), because it means that we once we ship Phobos compiled with -dip1000 everyone has to use -dip1000 too (and of course in the other way around: as long as we don't ship a -dip1000 version of Phobos, it's really difficult to us Phobos with -dip1000).

marler8997:
> because it means that we once we ship Phobos compiled with -dip1000 everyone has to use -dip1000 too

I'm sure the path would be to make dip 1000 enabled by default before shipping phobos with dip1000 enabled.

schveiguy:
Am I understanding that we are running the auto tester with dip1000, but shipping without it? Is this really a good thing? I mean, dip1000 isn't being shipped, isn't properly working, yet it's impeding other work (such as this PR). Worse yet, if we have issues that are somehow masked by dip1000, they will get through the tester.

I understand and agree that the dip1000 issues are important, but at the cost of preventing other things from being worked on, I'd prefer it be tested in the background instead of being a blocker. You could run only the master branch with -dip1000 and still see when things went bad. Most of the time, dip1000 issues aren't going to regress anyway.

wilzbach:
> Am I understanding that we are running the auto tester with dip1000, but shipping without it?

Partially. Only on some machines.

> Is this really a good thing?

Yes (see answers below).

> Worse yet, if we have issues that are somehow masked by dip1000, they will get through the tester.

That's why it's only run on some CIs.

> I understand and agree that the dip1000 issues are important, but at the cost of preventing other things from being worked on

If that's really _blocking_ this PR, you could remove the `$(aa[$(subst /,.,$(basename $<))]) ` substitution from [this line](https://github.com/dlang/phobos/blob/ac1dc46f4111d8e2e3b38bf2c8b9c0ff2a097d1f/posix.mak#L351). Though we would need to ensure that then the *.test targets are run at dlang/dmd too (at the moment they are only run at dlang/phobos via CircleCi). Shouldn't be a big deal though.

> I'd prefer it be tested in the background instead of being a blocker.

No warnings don't work. Everyone ignores them.
A recent example: https://github.com/dlang/dmd/pull/8018

> You could run only the master branch with -dip1000 and still see when things went bad.

Not sure how this be different from the status quo? (`dip1000.mak` isn't in stable yet).
Also `-dip1000` modules needs to be enforced, because (1) otherwise it will regress and (2) it makes approving -dip1000 PRs a lot easier.

> Most of the time, dip1000 issues aren't going to regress anyway.

No. People love to break -dip1000, because no one (except Walter maybe), really understands it.
There were quite a few regressions until https://github.com/dlang/phobos/pull/6278.

schveiguy:
> because no one (except Walter maybe), really understands it.

This isn't encouraging when it's a blocker in some cases. How can there be regressions if it wasn't being tested before that PR?

> > You could run only the master branch with -dip1000 and still see when things went bad.
> 
> Not sure how this be different from the status quo?

I mean not run it on any PRs, just the master build. And I don't mean the real master, I mean like a separate build, similar to how @yebblies had his own build branch until we flipped the DDMD switch.

This is similar to the broken cycle situation we had. Until the cycle tester was fixed, any PR could mysteriously break the build for weird reasons, and then the PR is rejected for reasons having nothing to do with the PR changes. Having the status quo be a broken build is not conducive to productivity.

> That's why it's only run on some CIs.

Which ones? The only ones that matter here are the autotester (where it seems every platform but Windows is using -dip1000). It'd be fine if there was a different CI that tested dip1000 issues and we could override it individually, but the time to build Phobos with separate compilation probably rules that out.

wilzbach:
> Though we would need to ensure that then the *.test targets are run at dlang/dmd too (at the moment they are only run at dlang/phobos via CircleCi). Shouldn't be much work though.

-> https://github.com/dlang/phobos/pull/6443
@marler8997  - this should unblock your PR?

changelog/selectiveUnittests.dd
```diff
@@ -0,0 +1,3 @@
+Only compile unittests in compiled modules
+
+The compiler was modified to only compile unittests in modules that are being compiled.  The most common impact is that unittests from druntime/phobos will not be included when compiling other programs that use these libraries.
```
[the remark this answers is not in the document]
marler8997: Actually...I should update this.  Turns out the unittests already weren't being compiled in (thanks @schveiguy for looking into this).  What's actually happening is that `version(unittest)` blocks are importing and instantiating templates that aren't instantiated without `-unittest`.  Then these instantiations are being moved to root modules.  With this new information, as far as I know this change shouldn't have any noticeable impact on users.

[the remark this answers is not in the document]
CyberShadow: So this just fixes sync between whether `unittest` blocks and `version(unittest)` blocks are being compiled, right?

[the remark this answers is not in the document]
marler8997: Actually no.  This is preventing template instances inside libraries from being moved to root modules when compiling with `-unittest`.  This is what's causing the compilation slowdown because the template instances that are being moved cause ALOT of phobos modules to be loaded and analyzed that wouldn't be otherwise.

`version(unittest)` blocks still get analyzed, however, @schveiguy is working on a phobos PR that should prevent those blocks from causing too much slowdown.  This PR along with the phobos PR should be everything we need to make `-unittest` compile just as fast as non `-unittest` compilations (when the app has no unittesting itself).

src/dmd/astbase.d
marler8997: Was just looking through the comments and saw that your question wasn't answered.  The module class you are looking at is missing ALOT of information, it's just as STUB as far as I know probably meant to hold a future refactored version of the current `Module` class in `dmodule.d`.

schveiguy: I had already figured it out. it seems it's there for when you want to instantiate the parser only, and not actually compile anything. Perhaps if you aren't generating object code.

schveiguy:
> Any progress with phobos?

I want to split my single giant commit into smaller PRs for easier review. Will do soon.

schveiguy:
@klickverbot I think I understand what you are saying. An imported module may instantiate more templates when compiled with -unittest. The mechanism that says "I don't need to instantiate this template locally because it's already done inside the import" will NOT move the instantiation local (given this PR is accepted). So if you haven't *actually* compiled the imported module with -unittest, then there is no instantiation anywhere, and you get linker errors.

BUT, with this PR, the only place this should happen is in `version(unittest)` blocks, since unittests in imported modules aren't parsed or examined. So what we need to ensure is that `version(unittest)` blocks do not instantiate any templates that aren't already instantiated. I personally think we should get rid of as much `version(unittest)` as possible in any case. I've been doing so, but without this understanding, so may need a revisit once I'm done.

I'm very concerned about eliminating `version(unittest)` in imported modules, as it's used in some places to provide common harnesses for unittesting. For example, `allocator.common.testAllocator`.

marler8997:
The demangled symbol looks like this:
```D
pure nothrow @nogc @safe void std.conv.toChars!(10, char, 1, uint).toChars(uint).Result.popFront()
```
Something to look into...

schveiguy:
A shame we can't get this to work.

Likely the symbol failure is because DMD gets confused about which symbols are already generated. The fact that it's a uint, and a 64-bit system has size_t being ulong, makes me think that in the 32-bit case it doesn't fail because the uint and size_t versions are the same, and size_t is used somewhere else that makes it link.

PR #8125, opened 2018-04-04
[WIP] Fix Issue 18718 - ICE in dmd/traits.d:417

wilzbach:
I know that this isn't "the right fix", but I didn't know DMD protects itself from self-recursion.
Are there any precendents I can follow?

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -1332,7 +1332,17 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
             return 0;
         }
 
-        ScopeDsymbol._foreach(sc, sds.members, &pushIdentsDg);
+        __gshared int counter;
+        if (counter > 1000)
+        {
+            e.error("Maximal recursion depth hit");
+        }
+        else
+        {
+            counter++;
+            ScopeDsymbol._foreach(sc, sds.members, &pushIdentsDg);
+        }
+        counter = 0;
         auto cd = sds.isClassDeclaration();
         if (cd && e.ident == Id.allMembers)
         {
--- a/test/fail_compilation/test18718.d
+++ b/test/fail_compilation/test18718.d
@@ -1,0 +1,18 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/test18718.d(15): Error: Maximal recursion depth hit
+fail_compilation/test18718.d(11): Error: mixin `test18718.World.BuildStuff!()` error instantiating
+---
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=18718
+struct World {
+	mixin BuildStuff;
+}
+
+template BuildStuff() {
+	static foreach(elem; __traits(allMembers, typeof(this))) {
+
+	}
+}


src/dmd/traits.d
```diff
@@ -1332,7 +1332,17 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
             return 0;
         }
 
-        ScopeDsymbol._foreach(sc, sds.members, &pushIdentsDg);
+        __gshared int counter;
+        if (counter > 1000)
```
bausshf: How is this counter calculated, like why is it exactly when it's above 1000?

wilzbach: Well, it's my noobish attempt at fixing this stack overflow and completely arbitrary. This function (`semanticTraits`) recursively calls itself and I bet there's a more sophisticated way in DMD to protect itself against such recursive overflows, but I couldn't find one.

RazvanN7: Actually, there isn't. Look how CTFE detects infinite recursion:
https://github.com/dlang/dmd/blob/master/src/dmd/dinterpret.d#L923
where CTFE_RECURSION_LIMIT is ... 1000.  

Or template instantiation: https://github.com/dlang/dmd/blob/master/src/dmd/dsymbolsem.d#L5640 

I guess this fix is fine for now, until a better mechanism is found.

RazvanN7 (approved):

src/dmd/traits.d
JinShil: Would it make sense to leverage the same limit, rather than create a new one?

src/dmd/traits.d
```diff
@@ -1332,7 +1332,17 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        {
+            e.error("Maximal recursion depth hit");
```
JinShil: I recommend "recursion limit exceeded" to remain consistent with the wording used at https://github.com/dlang/dmd/blob/master/src/dmd/dinterpret.d#L927

src/dmd/traits.d
RazvanN7: I guess we could add a symbol recursionLimit which can be used throughout the compiler

src/dmd/traits.d
stefan-koch-sociomantic: it can't hurt to make it explicit.

src/dmd/traits.d
```diff
@@ -1332,7 +1332,17 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        }
+        else
+        {
+            counter++;
+            ScopeDsymbol._foreach(sc, sds.members, &pushIdentsDg);
+        }
```
stefan-koch-sociomantic: decrease counter here, otherwise you are counting static_foreach frequency and not just recustion-depth.

jacob-carlborg:
@wilzbach ping, what's the status of this?

wilzbach:
Superseded by https://github.com/dlang/dmd/pull/8213

PR #8126, opened 2018-04-04
Fix Issue 17991 - ICE with imports without module

wilzbach:
Not sure about the policy of catching/catching null-pointers.
Is this the job of the caller or callee here?

### DIFF
--- a/src/dmd/dsymbol.d
+++ b/src/dmd/dsymbol.d
@@ -1487,6 +1487,10 @@ public:
 
     final void addAccessiblePackage(Package p, Prot protection)
     {
+        // https://issues.dlang.org/show_bug.cgi?id=17991
+        // An import of truly empty file/package can happen
+        if (p is null)
+            return;
         auto pary = protection.kind == Prot.Kind.private_ ? &privateAccessiblePackages : &accessiblePackages;
         if (pary.length <= p.tag)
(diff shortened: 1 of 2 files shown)


JinShil (approved):

src/dmd/dsymbol.d
```diff
@@ -1487,6 +1487,8 @@ public:
 
     final void addAccessiblePackage(Package p, Prot protection)
     {
+        if (p is null)
```
JinShil: Could use a comment with a link to the bug URL here.

RazvanN7:
> Not sure about the policy of catching/catching null-pointers.
Is this the job of the caller or callee here?

It depends on what the null pointer might mean. For example, `addAccessiblePackage` is only used 3 times in this block: https://github.com/dlang/dmd/blob/master/src/dmd/dsymbolsem.d#L1375 . At a first glance it seems that p is null if it is not inserted in the package symtab. I don't know if this is intended behavior or this should be fixed (i.e. insert it in the symtab). If p being null can be ignored, then this PR correctly fixes it, otherwise it's the caller's job to verify this and take action accordingly

RazvanN7:
I guess we can merge it as is and fix it later if we can come up with better ideas. At least we'll get rid of the ice

PR #8130, opened 2018-04-04
More test documentation

marler8997:


test/README.md
```diff
@@ -150,8 +160,10 @@ In-test variables
                          Optionally a MODEL suffix can used for further filtering, e.g.
                          win32 win64 linux32 linux64 osx32 osx64 freebsd32 freebsd64
 
-Relevant environment variables
+Environment variables
 ------------------------------
+Tests are run through the Makefile, which uses environment variables to store test settings and to pass these settings to the `d_do_test` tool and to the BASH tests.
```
wilzbach: Spelling is Bash ?

https://en.wikipedia.org/wiki/Bash_(Unix_shell)

test/README.md
```diff
@@ -83,8 +83,20 @@ Makefile targets
...
+All tests defined within `.d` source files may be configured using the settings listed below, i.e.
+
+mytest.d
+```
+/*
+REQUIRED_ARGS: -version=Foo
+EXTRA_SOURCES: imports/foo.d
+*/
+// ... test code ...
+```
```
wilzbach: Three idea that come to my mind:
- mention difference between `extra-files` and `imports` (i.e. `imports` is automatically added to the imports path)
- mention `TEST_OUTPUT` as it's probably the most commonly used option
- `EXTRA_SOURCES` is one of the things that gets "abused" easily and maybe we shouldn't recommend (i.e. if the file is in `imports`, you can simply `import` it, it's only required if you explicitly need the file as DMD argument and even then is your `-i` easier. So this leaves it useful only to e.g. ddoc files)

test/README.md
```diff
@@ -168,3 +182,22 @@ Windows vs non-windows portability env vars:
...
+
+    RESULTS_TEST_DIR   the results directory for tests of this type, e.g. test_results/runnable
+
+    OUTPUT_BASE        the prefix used for test output files, e.g. test_results/runnable/mytest
+
+    EXTRA_FILES        directory for extra files of this test type, e.g. runnable/extra-files
+
+    LIBEXT             platform-specific extension for library files, e.g. .a or .lib
+
+    OBJ                platform-specific extension for object files, e.g. .obj or .o
```
wilzbach: :+1:

test/README.md
```diff
@@ -168,3 +182,22 @@ Windows vs non-windows portability env vars:
...
     EXE:          .exe or <null> (required)
+
+BASH Tests
+------------------------------
+Inside BASH tests, there are a set of predefined variables that can be used to access common test data.  These are defined in `tools/exported_vars.h`.
+
+    TEST_DIR           the name of the test directory
+                       (one of compilable, fail_compilation or runnable)
+
+    TEST_NAME          the base name of the test file without the extension
```
wilzbach: Maybe add "e.g.  test15897" ?

test/README.md
```diff
@@ -168,3 +182,22 @@ Windows vs non-windows portability env vars:
     SEP:           \ or / (required)
     OBJ:          .obj or .o (required)
     EXE:          .exe or <null> (required)
+
+BASH Tests
+------------------------------
+Inside BASH tests, there are a set of predefined variables that can be used to access common test data.  These are defined in `tools/exported_vars.h`.
```
wilzbach: `.sh`

test/README.md
```diff
@@ -150,8 +162,10 @@ In-test variables
                          Optionally a MODEL suffix can used for further filtering, e.g.
                          win32 win64 linux32 linux64 osx32 osx64 freebsd32 freebsd64
 
-Relevant environment variables
+Environment variables
 ------------------------------
```
wilzbach: FWIW I tried to align the MarkDown header, but if it's meh to you, maybe use `##` ?

test/README.md
```diff
@@ -168,3 +182,20 @@ Windows vs non-windows portability env vars:
     SEP:           \ or / (required)
     OBJ:          .obj or .o (required)
     EXE:          .exe or <null> (required)
+
+BASH Tests
+------------------------------
+Along with environment variables defined by the Makefile (see above), there is another a set of variables that can be used inside BASH tests that are defined in `tools/exported_vars.h`.
```
wilzbach: - `.h` -> `.sh`
- BTW technically these variables are also exported in the postscript tests.

test/README.md
```diff
@@ -168,3 +182,20 @@ Windows vs non-windows portability env vars:
     SEP:           \ or / (required)
     OBJ:          .obj or .o (required)
     EXE:          .exe or <null> (required)
+
+BASH Tests
+------------------------------
```
wilzbach: Some markdown engines require a newline here (and you might want to align the length of the bar with heading or use `##` if you don't care)

wilzbach (commented):
Thanks a lot for the initiative!

test/README.md
```diff
@@ -150,8 +162,10 @@ In-test variables
                          Optionally a MODEL suffix can used for further filtering, e.g.
                          win32 win64 linux32 linux64 osx32 osx64 freebsd32 freebsd64
 
-Relevant environment variables
+Makefile Environment variables
 ------------------------------
+Tests are run through the Makefile, which uses environment variables to store test settings and to pass these settings to the `d_do_test` tool and to the BASH tests.
```
wilzbach: Spelling is Bash ?

https://en.wikipedia.org/wiki/Bash_(Unix_shell)

test/README.md
marler8997: Actually, `imports` is not added to the import paths

marler8997: Importing modules from the `imports` directory ONLY works if your import matches the filename, i.e.
```
--- test.d
version(WontWork)
    import imports.foo;
else
    import foo;
--- imports/foo.d
module foo;
```

test/README.md
marler8997: but it looks like they are "defined" in tools/exported_vars.h

wilzbach: Point was that the file ending is .sh -> https://github.com/dlang/dmd/blob/master/test/tools/exported_vars.sh

marler8997: yeah got that one

test/README.md
```diff
@@ -83,8 +83,27 @@ Makefile targets
...
+---
+*/
+import std.stdio;
+void main(string[] args)
+{
+    writeln("Hello, World!");
+}
+```
+
+The most commonly used settings is probably `TEST_OUTPUT:`. The following is a list of all available settings:
```
wilzbach: Oh I was just using this as a reasoning why it should be part of the example. Maybe just remove the first sentence, it sounds a bit informal?

test/README.md
marler8997: oh woops...I mispelled it again in my response! :)

test/README.md
```diff
@@ -83,8 +83,27 @@ Makefile targets
...
 
-In-test variables
+Test Configuration
 -----------------
+All tests defined within `.d` source files may be configured using the settings listed below, i.e.
+
+`mytest.d`:
+```
+/*
+REQUIRED_ARGS: -version=Foo
```
wilzbach: This has no effect - maybe add a `version(Foo)` in the example?

test/README.md
```diff
@@ -83,8 +83,27 @@ Makefile targets
     test_results/compilable/json.d.out      runs an individual test
                                             (run log of the test is stored)
 
-In-test variables
+Test Configuration
 -----------------
```
wilzbach: As mentioned I think some Markdown engines don't like this and require a newline after the heading.

test/README.md
```diff
@@ -150,9 +173,13 @@ In-test variables
                          Optionally a MODEL suffix can used for further filtering, e.g.
                          win32 win64 linux32 linux64 osx32 osx64 freebsd32 freebsd64
 
-Relevant environment variables
+Makefile Environment variables
 ------------------------------
 
+The Makefile uses environment variables to store test settings and as a way to pass these settings to the test wrapper tool `d_do_test`.
+
+> Note: These variables are available in any Bash test via ${VAR_NAME}
```
wilzbach: Note: Technically there's no difference between `$VAR_NAME` and `${VAR_NAME}` (for strings at least) except when that `${foo}${bar}` doesn't work with `$foo$bar`.
It's one of these legacy things that no one dares to deprecate because it would affect millions of millions of lines of code.

https://stackoverflow.com/questions/18135451/what-is-the-difference-between-var-var-and-var-in-the-bash-shell

So maybe we should just say: "These variables are available in the Bash tests too" ?

test/README.md
```diff
@@ -168,3 +195,21 @@ Windows vs non-windows portability env vars:
     SEP:           \ or / (required)
     OBJ:          .obj or .o (required)
     EXE:          .exe or <null> (required)
+
+Bash Tests
+----------
+
+Along with the environment variables provided by the Makefile (see above), an additional set of environment variables are made available to Bash tests.  These variables are defined in `tools/exported_vars.sh`:
```
wilzbach: Double whitespace here.

wilzbach (approved):
Final nits ;-)

JinShil (approved):

PR #6103, opened 2016-09-02
Custom assert handler to point to bugzilla

mihails-strasuns:
Makes it more obvious that any failed assertion is a compiler
bug to be reported. Inspired by http://forum.dlang.org/thread/qkxyfiwjwqklftcwtcik@forum.dlang.org

PR #8132, opened 2018-04-04
Automatically print the segtrace for segfaulting tests

wilzbach:
Useful because
- CIs: if a test segfaults on a CI --> we automatically get the stack trace
- local development: when something segfaults, you probably want to look at the stack trace

An alternative would be to use `coredumpctl gdb -1`, but that's systemd-only and I don't know whether all CIs support this.

So what do you think about this?

### DIFF
--- a/test/tools/d_do_test.d
+++ b/test/tools/d_do_test.d
@@ -675,7 +675,7 @@ int tryMain(string[] args)
     Result testCombination(bool autoCompileImports, string argSet, size_t permuteIndex, string permutedArgs)
     {
         string test_app_dmd = test_app_dmd_base ~ to!string(permuteIndex) ~ envData.exe;
-
+        string command; // copy of the last executed command so that it can be re-invoked on failures
         try
         {
             string[] toCleanup;
@@ -706,7 +706,7 @@ int tryMain(string[] args)
                 string objfile = output_dir ~ envData.sep ~ test_name ~ "_" ~ to!string(permuteIndex) ~ envData.obj;
                 toCleanup ~= objfile;
 
-                string command = format("%s -conf= -m%s -I%s %s %s -od%s -of%s %s %s%s %s", envData.dmd, envData.model, input_dir,
+                command = format("%s -conf= -m%s -I%s %s %s -od%s -of%s %s %s%s %s", envData.dmd, envData.model, input_dir,
                         reqArgs, permutedArgs, output_dir,
                         (testArgs.mode == TestMode.RUN || testArgs.link ? test_app_dmd : objfile),
                         argSet,
@@ -724,7 +724,7 @@ int tryMain(string[] args)
                     string newo= result_path ~ replace(replace(filename, ".d", envData.obj), envData.sep~"imports"~envData.sep, envData.sep);
                     toCleanup ~= newo;
 
-                    string command = format("%s -conf= -m%s -I%s %s %s -od%s -c %s %s", envData.dmd, envData.model, input_dir,
+                    command = format("%s -conf= -m%s -I%s %s %s -od%s -c %s %s", envData.dmd, envData.model, input_dir,
                         reqArgs, permutedArgs, output_dir, argSet, filename);
                     compile_output ~= execute(fThisRun, command, testArgs.mode != TestMode.FAIL_COMPILE, result_path);
                 }
@@ -732,7 +732,7 @@ int tryMain(string[] args)
                 if (testArgs.mode == TestMode.RUN || testArgs.link)
                 {
                     // link .o's into an executable
-                    string command = format("%s -conf= -m%s%s%s %s %s -od%s -of%s %s", envData.dmd, envData.model,
+                    command = format("%s -conf= -m%s%s%s %s %s -od%s -of%s %s", envData.dmd, envData.model,
                         autoCompileImports ? " -i" : "",
                         autoCompileImports ? "extraSourceIncludePaths" : "",
                         envData.required_args, testArgs.requiredArgsForLink, output_dir, test_app_dmd, join(toCleanup, " "));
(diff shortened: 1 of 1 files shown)


test/tools/d_do_test.d
```diff
@@ -835,6 +840,17 @@ int tryMain(string[] args)
             writefln("Test %s/%s.%s failed.  The logged output:", input_dir, test_name, test_extension);
             writeln(output_file.readText);
             output_file.remove();
+
+            // automatically rerun a segfaulting test and print its stack trace
+            version(linux)
```
jacob-carlborg: Why only Linux?

jacob-carlborg:
There's `registerMemoryErrorHandler` as well. But that is only implemented for Linux.

test/tools/d_do_test.d
wilzbach: Because I didn't know whether it would work on other platforms and I thought it's good to start with something that I can test. Also most of our CIs are Linux ones, so it would already have a measurable impact.
IIRC OSX uses `lldb` and Windows has who knows what debugger available?

(+I didn't feel like fighting the auto-tester for something that hasn't received feedback yet.)

wilzbach:
> There's registerMemoryErrorHandler as well. But that is only implemented for Linux.

DMD is built in release mode on e.g. the auto-tester, so AFAICT unfortunately this won't work.
FYI: sth. similar has been tried before (see e.g. https://github.com/dlang/dmd/pull/6103), but this was blocked on the output of DIP1006 :/

test/tools/d_do_test.d
```diff
@@ -674,7 +674,7 @@ int tryMain(string[] args)
     Result testCombination(bool autoCompileImports, string argSet, size_t permuteIndex, string permutedArgs)
     {
         string test_app_dmd = test_app_dmd_base ~ to!string(permuteIndex) ~ envData.exe;
-
+        string lastCommand;
```
JinShil: I suggest adding a simple comment here describing what this is for, so the next person to modify this code doesn't have to do so much investigation.

JinShil:
I like it.

wilzbach:
BTW here's how this will look like:


```
 ... compilable/foo.d                -fPIC ()
sh: line 1:  6086 Segmentation fault      (core dumped) ../generated/linux/release/64/dmd -conf= -m64 -Icompilable -fPIC -odtest_results/compilable -oftest_results/compilable/foo_0.o -c compilable/foo.d > test_results/fa6211c06986167f92e582fe24165c671874856a5ffc0bae595cf8f133ed6bf 2>&1
Test compilable/foo.d failed.  The logged output:
../generated/linux/release/64/dmd -conf= -m64 -Icompilable  -fPIC  -odtest_results/compilable -oftest_results/compilable/foo_0.o  -c compilable/foo.d 


Test compilable/foo.d failed: expected rc == 0, exited with rc == 139

[Thread debugging using libthread_db enabled]
Using host libthread_db library "/usr/lib/libthread_db.so.1".

Program received signal SIGSEGV, Segmentation fault.
0x000055555589c9a7 in semanticTraits(TraitsExp*, Scope*) ()
#0  0x000055555589c9a7 in semanticTraits(TraitsExp*, Scope*) ()
#1  0x00005555558157cd in ExpressionSemanticVisitor::visit(TraitsExp*) ()
#2  0x000055555580527d in TraitsExp::accept(Visitor*) ()
#3  0x00005555558246e7 in expressionSemantic(Expression*, Scope*) ()
#4  0x000055555578c525 in _D3dmd4cond13StaticForeach7prepareMFPSQBk6dscope5ScopeZv ()
#5  0x0000555555785b5c in StaticForeachDeclaration::include(Scope*) ()
#6  0x00005555557d57e9 in _D3dmd7dsymbol12ScopeDsymbol8_foreachFPSQBm6dscope5ScopePSQCe4root5array__T5ArrayTCQDdQDc7DsymbolZQxMDFmQwZiPmZi ()
#7  0x000055555589eeb3 in semanticTraits(TraitsExp*, Scope*) ()
#8  0x00005555558157cd in ExpressionSemanticVisitor::visit(TraitsExp*) ()
#9  0x000055555580527d in TraitsExp::accept(Visitor*) ()
#10 0x00005555558246e7 in expressionSemantic(Expression*, Scope*) ()
#11 0x000055555578c525 in _D3dmd4cond13StaticForeach7prepareMFPSQBk6dscope5ScopeZv ()
#12 0x0000555555785b5c in StaticForeachDeclaration::include(Scope*) ()
...
```

Luckily there are enough ICEs available, this one was "tested" with https://issues.dlang.org/show_bug.cgi?id=18718

JinShil:
So no longer an RFC?

JinShil:
Also, if a stack trace is added to #8138 will this even be necessary?

I like the idea, I'm not convinced of the implementation yet.  I though DMD had a way to output the stack trace without using GDB.  Am I mistaken?

test/tools/d_do_test.d
```diff
@@ -674,7 +674,7 @@ int tryMain(string[] args)
     Result testCombination(bool autoCompileImports, string argSet, size_t permuteIndex, string permutedArgs)
     {
         string test_app_dmd = test_app_dmd_base ~ to!string(permuteIndex) ~ envData.exe;
-
+        string command; // copy of the last executed command, s.t. it can be re-invoked on failures
```
JinShil: I would prefer if we didn't start using informal chatting abbreviations.  Please spell out "so that".

JinShil (approved):

nordlow:
How can I make use of this idea during local development of my dub projects in Linux?

wilzbach:
Not sure what you are referring to?
You want dub to automatically start gdb?
Then just define a simple bash script that does so and pass it it to dub with --compiler?

nordlow:
I want dub run to print the stacktrace of my app when it segfaults.

PR #8134, opened 2018-04-04
d_do_test: Move 'msc' to envData

wilzbach:
- `msc` only depends on `envData.ccompiler`
- I wasn't quite sure what `msc` stands for ([Meteorological Service of Canada?, Missionaries of the Sacred Heart?](https://en.wikipedia.org/wiki/MSC)), so I hope that Microsoft's compiler is referred to here

edit: I just learned that according to https://en.wikipedia.org/wiki/Microsoft_Visual_C%2B%2B#Internal_version_numbering the initial name of MSVC++ was MSC (well, until 1993)

### DIFF
--- a/test/tools/d_do_test.d
+++ b/test/tools/d_do_test.d
@@ -99,6 +99,7 @@ struct EnvData
     bool dobjc;
     bool coverage_build;
     bool autoUpdate;
+    bool usingMicrosoftCompiler;
 }
 
 bool findTestParameter(const ref EnvData envData, string file, string token, ref string result, string multiLineDelimiter = " ")
@@ -434,7 +435,7 @@ unittest
         == `fail_compilation\diag.d(2): Error: fail_compilation\imports\fail.d must be imported`);
 }
 
-bool collectExtraSources (in string input_dir, in string output_dir, in string[] extraSources, ref string[] sources, bool msc, in EnvData envData, in string compiler)
+bool collectExtraSources (in string input_dir, in string output_dir, in string[] extraSources, ref string[] sources, in EnvData envData, in string compiler)
 {
     foreach (cur; extraSources)
     {
@@ -443,7 +444,7 @@ bool collectExtraSources (in string input_dir, in string output_dir, in string[]
         string command = compiler;
         if (envData.compiler == "dmd")
         {
-            if (msc)
+            if (envData.usingMicrosoftCompiler)
             {
                 command ~= ` /c /nologo `~curSrc~` /Fo`~curObj;
             }
@@ -590,7 +591,8 @@ int tryMain(string[] args)
             default:      envData.ccompiler = "c++"; break;
         }
     }
-    bool msc = envData.ccompiler.toLower.endsWith("cl.exe");
+
+    envData.usingMicrosoftCompiler = envData.ccompiler.toLower.endsWith("cl.exe");
 
     if (!gatherTestParameters(testArgs, input_dir, input_file, envData))
         return 0;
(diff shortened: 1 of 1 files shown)


test/tools/d_do_test.d
```diff
@@ -99,6 +99,13 @@ struct EnvData
     bool dobjc;
     bool coverage_build;
     bool autoUpdate;
+    bool hasMicrosoftStudioCompiler() const
+    {
+        import std.uni : sicmp;
+        import std.range : only;
+        alias lowerCasePred = (a, b) => a.only.sicmp(b.only) == 0;
+        return ccompiler.endsWith!lowerCasePred("cl.exe");
+    }
```
wilzbach: Wasn't sure about this. I think there are pros (lazy, encapsulated) and cons (only set once) to it.
I'm happy to drop this bit if it's controversial.

test/tools/d_do_test.d
```diff
@@ -590,7 +597,6 @@ int tryMain(string[] args)
             default:      envData.ccompiler = "c++"; break;
         }
     }
-    bool msc = envData.ccompiler.toLower.endsWith("cl.exe");
```
marler8997: I would just change this line to `envData.hasMicrosoftCompiler = envData.ccompiler.toLower.endsWith("cl.exe");`

test/tools/d_do_test.d
marler8997: This variable is checked many times (especially when there are alot of test combinations) and is ALWAYS checked at least once, so lazy doesn't really make sense.

I'd just make this a field like everything else and assign it below.

marler8997: Also, maybe change the field to `usingMicrosoftCompiler`

marler8997: Wait...I misread the code...looks like lazy might make sense...

wilzbach: From https://forum.dlang.org/post/p6oibo$1lmi$1@digitalmars.com

> 19. Try to use function prefixes:
> 'is' is the parameter in a certain category?
> 'has' does the parameter have a certain feature?
> 'can' for can I do X with the parameter?

There are multiple features (aka compilers: D, C++, Objective C, ...), so "has" sounded to me like the best name to query for the feature's existence.

marler8997: But I guess the method parameter is really the "environment", not the compiler itself....  I must say that "has" is a bit confusing...

The least confusing might be `usingMicrosoftCompiler`...saying that  the environment is "using" the Microsoft compiler.

wilzbach: Yes, but you query `envData` which has multiple compilers (`compiler`, `ccompiler`, and `dobjc`), thus my reasoning, but I don't feel strongly about it. I think both are a lot better than `msc` ;-)

wilzbach: Ok changed.

marler8997: Not sure about the laziness though...I'll let you decide but it makes me a bit uneasy :)

marler8997 (approved):

test/tools/d_do_test.d
```diff
@@ -590,7 +591,10 @@ int tryMain(string[] args)
...
+
+    // can't modify the lines above due to auto-tester's patching
```
marler8997: Can you surround it with comments?
```D
/* CAN'T REMOVE BECAUSE OF AUTOTESTER PATCHING
bool msc = envData.ccompiler.toLower.endsWith("cl.exe");
*/

wilzbach: Cool. Removed then :)

marler8997:
merge it

PR #8117, opened 2018-04-03
Allow to run_individual_tests from any directory

wilzbach:
If the `run_individual_tests` is used from another directory (e.g. the root or `src`), it currently doesn't allow to use `runnable/aliasthis.d`.
It's easy to check whether the file exists and I don't see any reason why we shouldn't.

### Current behavior:

```
cd ~/dlang/dmd
./test/run_individual_tests runnable/aliasthis.d # fails
./test/run_individual_tests test/runnable/aliasthis.d # works
./test/run_individual_tests foo # fails
cd test
./run_individual_tests runnable/aliasthis.d # works (main usage)
```

#### New behavior:

```
cd ~/dlang/dmd
./test/run_individual_tests runnable/aliasthis.d # works (changed behavior)
./test/run_individual_tests test/runnable/aliasthis.d # works
./test/run_individual_tests foo # fails
cd test
./run_individual_tests runnable/aliasthis.d # works (main usage)
```

PR #8135, opened 2018-04-04
run_individual_tests print make command

marler8997:
`run_individual_tests` was previously printing a single number which was confusing.  Turns out it was the number of processors.  Useful info, but makes more sense when paired with the full make command that ends up being executed.

### DIFF
--- a/test/run_individual_tests
+++ b/test/run_individual_tests
@@ -37,5 +37,8 @@ done
 # allows the script to be called from anywhere
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 cd $DIR
-echo ${PROCS}
-${MAKE:-make} "${files[@]}" -j${PROCS}
+function echorun {
+    echo $@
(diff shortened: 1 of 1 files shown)


wilzbach:
> run_individual_tests was previously printing a single number which was confusing. 

Yeah that one ended up accidentally in there. Sorry about that.
I was actually thinking to just remove it (as e.g. proposed in https://github.com/dlang/dmd/pull/8117).

> Useful info, but makes more sense when paired with the full make command that ends up being executed.

Not sure about this:
- does this extra information help the user in any way?
- the `d_do_test.d` already prints which tests were executed (and with which params)
- we don't log any calls of `d_do_test.d` from the Makefile either
- The long-term plan is to replace the Makefile with a pure D test runner and drop the dependency on Make, so maybe it's a good idea to start to "abstract" things behind `run_individual_tests`

marler8997:
Sounds good, closing in favor of #8117

PR #8129, opened 2018-04-04
Route all tests through d_do_test

marler8997:
Provides a common entry point for all tests to perform common setup/operations.  Also simplifies the interface to run a test (resulting in a simplified Makefile).

PR #8136, opened 2018-04-04
Port the individual test script runner to D

wilzbach:
- The individual test script runner is still "in development" (it has been added two weeks ago in https://github.com/dlang/dmd/pull/8040)
- As this is the "main repo" of the D programming language, we should be able to write our scripts in D too
- `rdmd` caches the build and subsequent executions don't require a compilation step of this tool (so there shouldn't be a noticeable slowdown in comparison to bash)
- The long-term goal is to skip the Makefile middleman, s.t. executing the testsuite on Windows doesn't depend on GNUmake

See also: #8129 (for another step on the journey from GNUMake to D)


test/run_individual_tests.d
```diff
@@ -0,0 +1,47 @@
...
+/**
+Allows running tests individually
+
+Usage:
+  ./run_individual_tests.d <test-file>...
+
+Example:
+  ./run_individual_tests.d runnable/template2962.d fail_compilation/fail282.d
+
+See the README.md for all available test targets
```
wilzbach: Maybe drop this header in favor of the built-in help and README.md?

test/run_individual_tests.d
```diff
@@ -0,0 +1,47 @@
...
+    string[] targets;
+    foreach (file; args[1 .. $])
+    {
+        auto normalizedDir = file.absolutePath.dirName;
+        targets ~= "test_results/%s/%s.out".format(normalizedDir.baseName, file.baseName);
+    }
+    auto scriptDir = __FILE_FULL_PATH__.dirName;
+    auto makeArguments = targets;
+    makeArguments ~= "--jobs=".text(jobs);
+    spawnProcess("make" ~ makeArguments, null, Config.none, scriptDir).wait;
```
wilzbach: With #8129, all that is left to replace the Makefile is:
- only add targets which have a target with an older or non-existing timestamp
- replace jobs with `foreach (target; targets.parallel(1))`
- ensure the `test_results` directories exists
- set the environment flags
- resolve shorthands like `all` or `run_fail_compilation_tests`
- ensure that the test sanitizers are build
- add `if (targets.canFind("clean") { return removeRecursive(targetDir);}`

test/run_individual_tests.d
```diff
@@ -0,0 +1,47 @@
...
+`, res.options);
+        return;
+    }
+
+    string[] targets;
+    foreach (file; args[1 .. $])
+    {
+        auto normalizedDir = file.absolutePath.dirName;
+        targets ~= "test_results/%s/%s.out".format(normalizedDir.baseName, file.baseName);
+    }
```
jacob-carlborg: Replace with `map`?

test/run_individual_tests.d
```diff
@@ -0,0 +1,47 @@
...
+    auto scriptDir = __FILE_FULL_PATH__.dirName;
+    auto makeArguments = targets;
```
jacob-carlborg: Why the extra variable?

test/run_individual_tests.d
```diff
@@ -0,0 +1,47 @@
...
+    makeArguments ~= "--jobs=".text(jobs);
```
jacob-carlborg: Perhaps concatenate this on the previous line.

jacob-carlborg (changes requested):

test/run_individual_tests.d
wilzbach: Ok. Done.

jacob-carlborg (approved):

wilzbach:
> So this supposed to be a shorthand form of running dcompiler -flags test1234.d yourself? :-)

Yes, but do you have _auto-completion_ for that? :stuck_out_tongue: 

> but the lengths that some people will go to in order to avoid typing...

This actually originates from the shell scripts which you can't easily run with `dcompiler -flags`.
Also there are many tests which do more than just compiling a file (ddoc, json, imports, other postprocessing) which you don't want to do by hand.
Other goodies:
- Run all diagnostics tests -> `./run_individual_tests.d ./fail_compilation/diag*`
- full features of the test suite like e.g. setting up all environment variables, diffing the output, maybe [auto-displaying a stack trace soon](https://github.com/dlang/dmd/pull/8132) ...
- a step towards replacing the Makefile which apparently ([1], [2], [3], [4]) is a huge problem for people on Windows

See https://github.com/dlang/dmd/pull/8040 for more details/motivation.

[1]  https://forum.dlang.org/post/tulbhulbeqqzofdxevcg@forum.dlang.org
[2] https://forum.dlang.org/post/gxxfmrnezfrlodlhpiwe@forum.dlang.org
[3] https://forum.dlang.org/post/uyzscudefkhqyjdzvczc@forum.dlang.org
[4] https://issues.dlang.org/show_bug.cgi?id=18413

JinShil (approved):

PR #8137, opened 2018-04-05
Apply auto-tester's patching to upstream and avoid the need for it

wilzbach:
Applies https://github.com/braddr/at-client/blob/master/src/diff-dmd-win64.diff to master.
and thus ideally this removes the need for the auto-tester to do its patching.

I wasn't sure what to do about via patching disabled [`runnable/test15779.d`](https://github.com/dlang/dmd/blob/master/test/) test as it passes on AppVeyor. Is this still an issue?

@braddr - I guess at the moment this will fail and that's why https://github.com/braddr/at-client/pull/8 would be useful, s.t. (1) this can be tested before and (2) `master` and `stable` don't need to be upgraded in lockstep.
However, I guess simply disabling the patching would work fine too. `stable` isn't particularily busy and I can back-port these changes to `stable` in a quick follow-up PR.
If you want we can also do all three repos in one stroke (making similar PRs for druntime + phobos's patching shouldn't be too hard once it's disabled).

-----

@ other reviewers: this PR depends on Brad and changes in the auto-tester infrastructure. Please do _not_ merge without his feedback and approval.

edit: don't get confused by "Pass: 8" on the auto-tester - it didn't run the tests on Windows:

image

### DIFF
--- a/test/tools/d_do_test.d
+++ b/test/tools/d_do_test.d
@@ -586,7 +586,7 @@ int tryMain(string[] args)
         switch (envData.os)
         {
             case "win32": envData.ccompiler = "dmc"; break;
-            case "win64": envData.ccompiler = `\"Program Files (x86)"\"Microsoft Visual Studio 10.0"\VC\bin\amd64\cl.exe`; break;
+            case "win64": envData.ccompiler = environment.get("VCBIN_DIR") ~ `\cl.exe`; break;
             default:      envData.ccompiler = "c++"; break;
         }
     }
--- a/win32.mak
+++ b/win32.mak
@@ -1,13 +1,12 @@
-MAKE=make
 
 auto-tester-build:
 	cd src
-	$(MAKE) -f win32.mak auto-tester-build
+	make -f win32.mak auto-tester-build
 	cd ..
 
 auto-tester-test:
 	cd test
-	$(MAKE)
+	gmake -j$(PARALLELISM)
 	cd ..
 	cd samples
 	gmake -f win32.mak DMD=..\src\dmd.exe MODEL=$(MODEL) "LIB=..\..\phobos;$(LIB)" \


win32.mak
```diff
@@ -1,13 +1,12 @@
-MAKE=make
 
 auto-tester-build:
 	cd src
-	$(MAKE) -f win32.mak auto-tester-build
+	make -f win32.mak auto-tester-build
```
marler8997: What's the reason for this change?

test/tools/d_do_test.d
```diff
@@ -586,7 +586,7 @@ int tryMain(string[] args)
         switch (envData.os)
         {
             case "win32": envData.ccompiler = "dmc"; break;
-            case "win64": envData.ccompiler = `\"Program Files (x86)"\"Microsoft Visual Studio 10.0"\VC\bin\amd64\cl.exe`; break;
+            case "win64": envData.ccompiler = environment.get("VCBIN_DIR") ~ `\cl.exe`; break;
```
marler8997: Why was this done as a patch instead of pushing directly to the repo?

Anyway, we might want the logic to be a bit more "forgiving" and helpful in "bad environment" cases.  Maybe something like this:
```D
auto vcbinDir = environment.get("VCBIN_DIR", null);
if (!vcbinDir)
{
    // NOTE: I added "C:" because windows only installs Visual Studio to the C drive and without this
    //       you break 64-bit testing if dmd is on another drive.  I actually have dmd on another drive
    //       but I've never run into this problem since I've never gotten the dmd test suit to work on windows :)
    vcbinDir = `"C:\Program Files (x86)\Microsoft Visual Studio 10.0\VC\bin\amd64"`;
}
if (!exists(vcbinDir))
{
    writefln("Error: the Visual Studio C compiler bin directory does not exist (%s)", vcbinDir);
    return 1; // fail
}
envData.ccompiler = vcbinDir ~ `\cl.exe`;
```

rainers:
> I wasn't sure what to do about via patching disabled runnable/test15779.d test as it passes on AppVeyor. Is this still an issue?

should be fixed by https://github.com/dlang/druntime/pull/2129, see https://issues.dlang.org/show_bug.cgi?id=18547

braddr:
Essentially all of this pull request is obsolete at this point.

The only file still being patched locally to the auto-tester in the dmd package is win32.mak, and I really hate those changes.  Dancing between versions and implementations of make is... horrible.  I'm very open to suggestions on how to make that suck less.

JinShil:
> Dancing between versions and implementations of make is... horrible. I'm very open to suggestions on how to make that suck less.

IMO, the way to make the whole *make* system cross platform, portable, and easily maintainable by all is to incrementally convert it all to D.

wilzbach:
> Essentially all of this pull request is obsolete at this point.

Cool! Thanks a lot!

PR #8138, opened 2018-04-05
Point users to Bugzilla on ICEs

wilzbach:
Revival of https://github.com/dlang/dmd/pull/6103.

While there's still a bit of hold-up on [DIP1006](https://github.com/dlang/DIPs/pull/54) being implemente see e.g. https://github.com/dlang/dmd/pull/7980, we can already prepare the codebase today.

Also LDC already allows to selectively enable assert's in `-release` today.

ICEs to test this:

Frontend
--------

```
///
template allSameType()
{
    static foreach (idx; T)
            enum allSameType ;
}
```

(from: https://issues.dlang.org/show_bug.cgi?id=18721)

```
> dmd -D
This is a compiler bug, please report it via https://issues.dlang.org/enter_bug.cgi
core.exception.AssertError@dmd/cond.d(378): Assertion failure
----------------
??:? _d_assertp [0x4f6c81b5]
??:? pure nothrow @nogc @safe void dmd.cond.StaticForeach.prepare(dmd.dscope.Scope*).__require() [0x4f4bb637]
??:? void dmd.cond.StaticForeach.prepare(dmd.dscope.Scope*) [0x4f4bb4f4]
??:? _ZN24StaticForeachDeclaration7includeEP5Scope [0x4f4b4b5b]
...
```

Backend
--------

```d
alias A = Complex!double;

struct Complex(T)
{
    double re, im;

    Complex!double baz()
    {
        return Complex!double(re).blah;
    }

    ref Complex!double blah()
    {
        im = re;
        return this;
    }
}
```

(from: https://issues.dlang.org/show_bug.cgi?id=15206)

```
> dmd -c -inline -O -m64
tym = x1d
Internal error: dmd/backend/cgxmm.c 684
This is a compiler bug, please report it via https://issues.dlang.org/enter_bug.cgi
```

Other ideas:
-----------

How about dumping all `global` data too? (or just a long variant of `--version`)?

### DIFF
--- a/src/dmd/backend/util2.c
+++ b/src/dmd/backend/util2.c
@@ -38,6 +38,8 @@ void *ph_calloc(size_t nbytes);
 void ph_free(void *p);
 void *ph_realloc(void *p , size_t nbytes);
 
+extern "C" void printInternalFailure(FILE* stream); // from dmd/mars.d
+
 
 void util_exit(int exitcode);
 
@@ -52,6 +54,7 @@ void file_progress()
 void util_assert(const char *file, int line)
 {
     fflush(stdout);
+    printInternalFailure(stdout);
     printf("Internal error: %s %d\n",file,line);
     err_exit();
 #if __clang__
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -70,6 +70,25 @@ private void logo()
     printf("DMD%llu D Compiler %s\n%s %s\n", cast(ulong)size_t.sizeof * 8, global._version, global.copyright, global.written);
 }
 
+/**
+Print DMD's logo with more debug information and error-reporting pointers.
+
+Params:
+    stream = output stream to print the information on
+*/
+extern(C) void printInternalFailure(FILE* stream)
+{
+    fputs(("---\n" ~
+    "ERROR: This is a compiler bug.\n" ~
+            "Please report it via https://issues.dlang.org/enter_bug.cgi\n" ~
+            "with, preferably, a reduced, reproducible example and the information below.\n" ~
+    "DustMite (https://github.com/CyberShadow/DustMite/wiki) can help with the reduction.\n" ~
+    "---\n").ptr, stream);
+    stream.fprintf("DMD %s\n", global._version);
+    stream.printPredefinedVersions;
+    stream.printGlobalConfigs();
+    fputs("---\n".ptr, stream);
+}
 
 /**
  * Print DMD's usage message on stdout
@@ -495,40 +514,10 @@ private int tryMain(size_t argc, const(char)** argv)
     Objc._init();
     builtin_init();
 
-    printPredefinedVersions();
-
     if (global.params.verbose)
     {
-        message("binary    %s", global.params.argv0);
-        message("version   %s", global._version);
-        message("config    %s", global.inifilename ? global.inifilename : "(none)");
-        // Print DFLAGS environment variable
-        {
-            Strings dflags;
-            getenv_setargv(readFromEnv(&environment, "DFLAGS"), &dflags);
-            OutBuffer buf;
-            foreach (flag; dflags.asDArray)
-            {
-                bool needsQuoting;
-                for (auto flagp = flag; flagp; flagp++)
-                {
-                    auto c = flagp[0];
-                    if (!(isalnum(c) || c == '_'))
-                    {
-                        needsQuoting = true;
-                        break;
-                    }
-                }
-
-                if (flag.strchr(' '))
-                    buf.printf("'%s' ", flag);
-                else
-                    buf.printf("%s ", flag);
-            }
-
-            auto res = buf.peekSlice() ? buf.peekSlice()[0 .. $ - 1] : "(none)";
-            message("DFLAGS    %.*s", res.length, res.ptr);
-        }
+        stdout.printPredefinedVersions();
+        stdout.printGlobalConfigs();
     }
     //printf("%d source files\n",files.dim);
 
@@ -1094,6 +1083,8 @@ int main()
         dmd_coverSetMerge(true);
     }
 
+    scope(failure) stderr.printInternalFailure;
+
     auto args = Runtime.cArgs();
     return tryMain(args.argc, cast(const(char)**)args.argv);
 }
@@ -1406,9 +1397,9 @@ void addDefaultVersionIdentifiers()
     VersionCondition.addPredefinedGlobalIdent("D_HardFloat");
 }
 
-private void printPredefinedVersions()
+private void printPredefinedVersions(FILE* stream)
 {
-    if (global.params.verbose && global.versionids)
+    if (global.versionids)
     {
         OutBuffer buf;
         foreach (const str; *global.versionids)
@@ -1416,10 +1407,45 @@ private void printPredefinedVersions()
             buf.writeByte(' ');
             buf.writestring(str.toChars());
         }
-        message("predefs  %s", buf.peekString());
+        stream.fprintf("predefs  %s", buf.peekString());
     }
 }
 
+extern(C) void printGlobalConfigs(FILE* stream)
+{
+    stream.fprintf("binary    %s\n", global.params.argv0);
+    stream.fprintf("version   %s\n", global._version);
+    stream.fprintf("config    %s\n", global.inifilename ? global.inifilename : "(none)");
+    // Print DFLAGS environment variable
+    {
+        StringTable environment;
+        environment._init(0);
+        Strings dflags;
+        getenv_setargv(readFromEnv(&environment, "DFLAGS"), &dflags);
+        environment.reset(1);
+        OutBuffer buf;
+        foreach (flag; dflags[])
+        {
+            bool needsQuoting;
+            foreach (c; flag[0 .. strlen(flag)])
+            {
+                if (!(isalnum(c) || c == '_'))
+                {
+                    needsQuoting = true;
+                    break;
+                }
+            }
+
+            if (flag.strchr(' '))
+                buf.printf("'%s' ", flag);
+            else
+                buf.printf("%s ", flag);
+        }
+
+        auto res = buf.peekSlice() ? buf.peekSlice()[0 .. $ - 1] : "(none)";
+        stream.fprintf("DFLAGS    %.*s\n", res.length, res.ptr);
+    }
+}
 
 /****************************************
  * Determine the instruction set to be used.
--- a/test/runnable/test18141.sh
+++ b/test/runnable/test18141.sh
@@ -6,4 +6,5 @@ else
     expected="Posix"
 fi
 
-echo "void main(){}" | "${DMD}" -v -o- - | grep "predefs" | grep "${expected}"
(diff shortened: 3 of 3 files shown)


JinShil:
> How about dumping all global data too? (or just a long variant of --version)?

Yeah, I think it would be helpful to output some more information.  Compiler version would be good.  I'm not sure what you mean by `global`.

JinShil:
I think you should also consider that this code may be shared by LDC and GDC which use different bug trackers.  It'd be nice to make is easy for them to adapt it to their workflows.

JinShil:
```
> echo "foo" | dmd -v -
predefs   DigitalMars Posix linux ELFv1 LittleEndian D_Version2 all D_SIMD D_InlineAsm_X86_64 X86_64 CRuntime_Glibc D_LP64 D_PIC assert D_HardFloat
binary    dmd
version   v2.079.0
config    /etc/dmd.conf
DFLAGS    (none)
parse     __stdin
...
```

I think I'd prefer that format over the json.

JinShil:
Is a stack trace going too far?

JinShil:
Also, I suggest "This is a compiler bug, please report it via https://issues.dlang.org/enter_bug.cgi with, preferably, a reduced, reproducible example." or something like that to encourage the user to do a little more than just cut and paste the compiler output into the bug tracker.

wilzbach:
Okay here's how it will look with the new changes:

```
---
ERROR: This is a compiler bug.
Please report it via https://issues.dlang.org/enter_bug.cgi
with, preferably, a reduced, reproducible example and the information below.
DustMite (https://github.com/CyberShadow/DustMite/wiki) can help with the reduction.
---
DMD v2.079.0-372-gc16061fe4
predefs   DigitalMars Posix linux ELFv1 LittleEndian D_Version2 all D_SIMD D_InlineAsm_X86_64 X86_64 CRuntime_Glibc D_LP64 D_PIC assert D_HardFloatbinary    /home/seb/dlang/bin/dmddev
version   v2.079.0-372-gc16061fe4
config    /home/seb/dlang/dmd/generated/linux/release/64/dmd.conf
DFLAGS    -I/home/seb/dlang/dmd/generated/linux/release/64/../../../../../druntime/import -I/home/seb/dlang/dmd/generated/linux/release/64/../../../../../phobos -L-L/home/seb/dlang/dmd/generated/linux/release/64/../../../../../phobos/generated/linux/release/64 -L--export-dynamic -fPIC
---
core.exception.AssertError@dmd/mars.d(1091): Assertion failure
```

> Is a stack trace going too far?

Assertion failures already have a stack trace right now (well, if assert's are enabled like in the release build) and they are very useful, so I don't think there's a reason not to show them.
If you were just asking about requesting to include it from the user I think "and the information below" catches this.

src/dmd/mars.d
```diff
@@ -70,6 +70,25 @@ private void logo()
...
+
+Params:
+    stream = output stream to print the information on
+*/
+extern(C) void internalFailureLogo(FILE* stream)
+{
+    stream.fprintf("---\n");
+    stream.fprintf("ERROR: This is a compiler bug.\n" ~
+            "Please report it via https://issues.dlang.org/enter_bug.cgi\n" ~
+            "with, preferably, a reduced, reproducible example and the information below.\n");
```
wilzbach: Maybe: "Include a a reduced, reproducible ..." ?

src/dmd/mars.d
```diff
@@ -70,6 +70,25 @@ private void logo()
     printf("DMD%llu D Compiler %s\n%s %s\n", cast(ulong)size_t.sizeof * 8, global._version, global.copyright, global.written);
 }
 
+/**
+Print DMD's logo with more debug information and error-reporting pointers.
+
+Params:
+    stream = output stream to print the information on
+*/
+extern(C) void internalFailureLogo(FILE* stream)
```
wilzbach: The backend was printing to `stdout` and I didn't want to change this behavior in this PR.

ibuclaw:
> Is a stack trace going too far?

No, stack trace is fine - and infact ideally all bug reports should have the file/line info of the ICE as it makes it easy to find possibly related issues.

src/dmd/mars.d
```diff
@@ -70,6 +70,25 @@ private void logo()
...
+{
+    stream.fprintf("---\n");
+    stream.fprintf("ERROR: This is a compiler bug.\n" ~
+            "Please report it via https://issues.dlang.org/enter_bug.cgi\n" ~
+            "with, preferably, a reduced, reproducible example and the information below.\n");
+    stream.fprintf("DustMite (https://github.com/CyberShadow/DustMite/wiki) can help with the reduction.\n");
```
ibuclaw: Time to make dustmite a dlang project? 😉

src/dmd/backend/util2.c
```diff
@@ -38,6 +38,8 @@ void *ph_calloc(size_t nbytes);
 void ph_free(void *p);
 void *ph_realloc(void *p , size_t nbytes);
 
+extern "C" void internalFailureLogo(FILE* stream); // from dmd/mars.d
```
JinShil: "Logo" or "Log"?

wilzbach: DMD already had a function called "logo" that printed its version. Hence "Logo" as this is an extended version of it, but I really don't care about the name ;-)

JinShil: Ok, I understand why it was called logo, but this isn't that.  Also, I suggested making methods begin with a verb, so my preference would be something like "printInternalFailure".

src/dmd/mars.d
wilzbach: Yeah :+1: 
FYI: it gets checked out as a "full source" to the tools repo already (https://github.com/dlang/tools/tree/master/DustMite), which is a bit useless as no one is developing on this copy.

CC @andralex

andralex (approved):
This is a great idea, looking forward to it!

src/dmd/mars.d
```diff
@@ -1406,20 +1397,56 @@ void addDefaultVersionIdentifiers()
...
+    if (global.versionids)
     {
         OutBuffer buf;
         foreach (const str; *global.versionids)
         {
             buf.writeByte(' ');
             buf.writestring(str.toChars());
         }
-        message("predefs  %s", buf.peekString());
+        stream.fprintf("predefs  %.*s", buf.peekString());
```
WalterBright: Use `%*s` for 0 terminated strings.

src/dmd/mars.d
```diff
@@ -1406,20 +1397,56 @@ void addDefaultVersionIdentifiers()
...
+        StringTable environment;
+        environment._init(0);
+        Strings dflags;
+        getenv_setargv(readFromEnv(&environment, "DFLAGS"), &dflags);
+        environment.reset(1);
+        OutBuffer buf;
+        foreach (flag; dflags.asDArray)
+        {
+            bool needsQuoting;
+            for (auto flagp = flag; flagp; flagp++)
```
WalterBright: Wat? Since when does incrementing a pointer until it is null work?

src/dmd/mars.d
```diff
@@ -1406,20 +1397,56 @@ void addDefaultVersionIdentifiers()
...
+    stream.fprintf("config    %s\n", global.inifilename ? global.inifilename : "(none)");
+    // Print DFLAGS environment variable
+    {
+        StringTable environment;
+        environment._init(0);
+        Strings dflags;
+        getenv_setargv(readFromEnv(&environment, "DFLAGS"), &dflags);
+        environment.reset(1);
+        OutBuffer buf;
+        foreach (flag; dflags.asDArray)
```
WalterBright: This `asDArray` seems mighty peculiar, and is in fact not necessary.

src/dmd/mars.d
jacob-carlborg: For null terminated strings?

src/dmd/mars.d
wilzbach: FWIW I just _moved_ the code - check the red part on the left.

src/dmd/mars.d
wilzbach: FWIW I just _moved_ the code - check the red part on the left.

src/dmd/mars.d
```diff
@@ -1406,20 +1397,56 @@ void addDefaultVersionIdentifiers()
...
+            {
+                auto c = flagp[0];
+                if (!(isalnum(c) || c == '_'))
+                {
+                    needsQuoting = true;
+                    break;
+                }
+            }
+
+            if (flag.strchr(' '))
```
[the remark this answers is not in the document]
wilzbach: I agree, but I just _moved_ the code - check the red part on the left.

src/dmd/mars.d
```diff
@@ -1406,20 +1397,55 @@ void addDefaultVersionIdentifiers()
...
+        Strings dflags;
+        getenv_setargv(readFromEnv(&environment, "DFLAGS"), &dflags);
+        environment.reset(1);
+        OutBuffer buf;
+        foreach (flag; dflags[])
+        {
+            bool needsQuoting;
+            foreach (c; flag[0 .. strlen(flag)])
+            {
+                if (!(isalnum(c) || c == '_' || c == '-' || c == '/' || c == '.'))
```
wilzbach: Example:
```
DFLAGS    -I/home/seb/dlang/dmd/generated/linux/release/64/../../../../../druntime/import -I/home/seb/dlang/dmd/generated/linux/release/64/../../../../../phobos -L-L/home/seb/dlang/dmd/generated/linux/release/64/../../../../../phobos/generated/linux/release/64 -L--export-dynamic -fPIC
```

(AFAICT no quoting needed)

src/dmd/mars.d
```diff
@@ -1406,20 +1397,55 @@ void addDefaultVersionIdentifiers()
...
+        StringTable environment;
+        environment._init(0);
+        Strings dflags;
+        getenv_setargv(readFromEnv(&environment, "DFLAGS"), &dflags);
+        environment.reset(1);
+        OutBuffer buf;
+        foreach (flag; dflags[])
+        {
+            bool needsQuoting;
+            foreach (c; flag[0 .. strlen(flag)])
```
wilzbach: Sadly `readFrom` calls `getenv` which returns a boring, old C-string.

wilzbach (commented):
> I understand, and normally I would let things slide for that reason. But this code is so clearly wrong I don't think we should defer fixing it.

Fair enough. Changed the code.

src/dmd/mars.d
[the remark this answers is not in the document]
wilzbach: Apparently it's not that easy (breaks a test), so I will modify this in a different PR.

JinShil:
@wilzbach Is this ready to go?  Do you plan on adding anything else?

wilzbach:
AFAICT this is ready (at least I don't plan on adding sth. else).
It would be great to combine this with DIP1006 (see https://github.com/dlang/dmd/pull/7980) and leave the asserts enabled, because at the moment this will only have effect for the backend code (the backend asserts stay enabled)

andralex (approved):

PR #8140, opened 2018-04-05
[trivial] d_do_test.d: show a warning when auto-updating the TEST_OUTPUT failed.

wilzbach:
It's non-trivial to automatically update the `TEST_OUTPUT` block when the test file has multiple blocks (i.e. a new error message appeared - to which block should it go?)
Sure, in theory that could be solved, but for now I think it's better to display a nice warning, s.t. a "user" isn't confused.

### DIFF
--- a/test/tools/d_do_test.d
+++ b/test/tools/d_do_test.d
@@ -817,13 +817,20 @@ int tryMain(string[] args)
             if (envData.autoUpdate)
             if (auto ce = cast(CompareException) e)
             {
+                // remove the output file in test_results as its outdated
+                output_file.remove();
+
                 auto existingText = input_file.readText;
                 auto updatedText = existingText.replace(ce.expected, ce.actual);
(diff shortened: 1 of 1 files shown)


test/tools/d_do_test.d
```diff
@@ -816,13 +816,20 @@ int tryMain(string[] args)
             if (envData.autoUpdate)
             if (auto ce = cast(CompareException) e)
             {
-                auto existingText = input_file.readText;
-                auto updatedText = existingText.replace(ce.expected, ce.actual);
-                std.file.write(input_file, updatedText);
                 // remove the output file in test_results as its outdated
-                f.close();
```
wilzbach: N.B.: It's not necessary to close the file manually - it's ref-counted and will be closed as soon as the scope is left.

JinShil (approved):

wilzbach:
(merge conflicts resolved)

PR #8142, opened 2018-04-05
fix issue 18730 - dmd miscompiles core.bitop.bt with -O

aG0aep6G:


src/dmd/backend/cod4.c
```diff
@@ -3604,6 +3604,8 @@ void cdbt(CodeBuilder& cdb,elem *e, regm_t *pretregs)
         cs.Iop = 0x0F00 | op;                     // BT rm,reg
         code_newreg(&cs,reg);
         cs.Iflags |= CFpsw | word;
+        if (I64)
```
WalterBright: I think this should be:
```
if (_tysize[ty1] == 8 && I64)
``

Comment (aG0aep6G) [reply src/dmd/backend/cod4.c]: `ty1` is the type of the pointer, isn't it? Shouldn't we check the type of the offset?
I.e.: `if (_tysize[ty2] == 8 && I64)` with `tym_t ty2 = tybasic(e2->Ety);`.

Comment (WalterBright) [reply src/dmd/backend/cod4.c]: No, E1 is an lvalue, not a pointer. The size of E1 and E2 should match, so we only need to test one of them, and the rest of the function tests the size of ty1.

Comment (aG0aep6G) [reply src/dmd/backend/cod4.c]: > The size of E1 and E2 should match

When I change `bitnum`'s type to `uint`, DMD still calls `cdbt` and the sizes don't match.

If we keep using `rdi` in that case, we get bad code:

```d
void main()
{
    size_t s;
    foo(&s, uint.max + 1L); /* Segfault because bt32 is being miscompiled. */
}
void foo(size_t* p, ulong x)
{
    bt32(p, cast(uint) x);
        /* The high bits of rdi are not being zeroed here. This is ok. bt32
        should not access them. */
}
int bt32(in size_t* p, uint bitnum) pure @system
{
    return ((p[bitnum >> 6] & (1L << (bitnum & 63)))) != 0;
        /* If rdi is used in the bt instruction, it will see the garbage high
        bits of rdi. With edi it would work correctly, as far I see. */
}
```

But maybe it's another bug that DMD calls `cdbt` with mismatching sizes? Why do they have to match, though?

WalterBright: See http://www.felixcloutier.com/x86/BT.html

The instructions available are for `mem16,r16`, `mem32,r32`, and `mem64,r64` so the lvalue size needs to match the register size.

There could be a bug in the generation of the OPbt instruction if it allows a mismatch.

aG0aep6G: > No, E1 is an lvalue, not a pointer.

I think you're mistaken. When I add

```c
printf("%X %d\n", ty1, _tysize[ty1]);
```

in `cdbt` and then compile core.bitop.bt, it prints "1D 8". That's [TYnptr](https://github.com/dlang/dmd/blob/3cf5d36578ff612654fddb8f0bb679257e5e4f3f/src/dmd/backend/ty.d#L66). And it stays "1D 8" even when I change it to `uint p` in the `bt` signature.

WalterBright: Aggh, you're right. BTW, when dmd is compiled with debug, you can use the `--b` `--f` flags to get the following output:
```
el:0150B1AC cnt=0 bt  TYint 0150AD9C 0150B06C
 el:0150AD9C cnt=0 var  mTYconst|TY*  p
 el:0150B06C cnt=0 var  TYunsigned  bitnum
```

WalterBright: with --b --o -O -m64:
```
el:014BB23C cnt=0 bt  TYint 014BADDC 014BB0AC
 el:014BADDC cnt=0 var  mTYconst|TY*  p
 el:014BB0AC cnt=0 var  TYuns long long  bitnum
```
so looks like you're right - check the size of E2.

aG0aep6G: > so looks like you're right - check the size of E2.

Unfortunately, that's not enough. We can't use the 64-bit instruction for `uint* p, ulong bitnum`. The machine would read 64 bits at a time, but only 32 might be accessible.

So I've also added a condition in `elbitwise` so that it doesn't emit `OPbt` in that case. This way we get:

D | asm
---- | -----
`ulong* p, ulong bitnum` | `bt QWORD PTR [rsi],rdi`
`ulong* p, uint bitnum`  | `bt DWORD PTR [rsi],edi`
`uint* p, uint bitnum`   | `bt DWORD PTR [rsi],edi`
`uint* p, ulong bitnum`  | not optimized

aG0aep6G:
While looking into this, I've found some more issues with the bt code. Enough to warrant a tracker issue, I think: [issue 18750](https://issues.dlang.org/show_bug.cgi?id=18750)

src/dmd/backend/cgelem.c
```diff
@@ -1391,6 +1391,7 @@ STATIC elem * elbitwise(elem *e, goal_t goal)
             ELCONST(e1->E1,1) &&
             (((e12 = e1->E2)->Eoper == OP64_32 ? (e12 = e12->E1) : e12)->Eoper == OPand) &&
             ELCONST(e12->E2,sz * 8 - 1) &&
+            tysize(e12->Ety) <= sz &&
```
WalterBright: What case is this necessary for?

aG0aep6G: That's so that the version with `uint* p, ulong bitnum` doesn't get optimized. The third test is about this.

RazvanN7:
@aG0aep6G The windows and semaphore failures are unrelated to this PR and happen because you are targeting `stable`. The stable-master branchoff happens in 3 days so you can retarget master.

I tried to retarget master for you but it seems that this will result in some conflicts. Can you please rebase this over master ?

aG0aep6G:
> Can you please rebase this over master ?

Done.

JinShil:
It appears that Appveyor isn't picking up the latest runtime code:

```
object.d(1456): Deprecation: `const` postblit is deprecated. Please use an unqualified postblit.
```
Closing and reopening in an attempt to remedy that.

PR #8143, opened 2018-04-06
fix Issue 18736 - constructor calls allowed after case labels

WalterBright:
Case labels are also labels, and break the simple flow analysis done by the constructor logic. The trouble is that labels mean more complex flow analysis is needed.

Blocked by https://github.com/dlang/phobos/pull/6425


jacob-carlborg:
If this require changes in Phobos it’s most likely that other projects are affected too. Can we make this a deprecation?

WalterBright:
There's no easy way to make that happen.

WalterBright:
The jenkins failure is:
```
utils/zlib.d(43,9): Error: constructor calls not allowed in loops or after labels
```
which is the same file as was corrected in https://github.com/dlang/phobos/pull/6425 @CyberShadow can you fix this, please?

jacob-carlborg:
> There's no easy way to make that happen.

@WalterBright a simple solution, perhaps a bit ugly, is to and a new entry to `CSX`, called `deprecatedLabel` or similar. Then where the error occurs check for both `CSX.label` and `CSX.deprecatedLabel`, if it's `deprecatedLabel` issue a deprecation, otherwise an error. Alternatively a bool can be added to `CtorFlow` if that's better.

CyberShadow:
> @CyberShadow can you fix this, please?

Done, but please keep in mind that the project tester is still meant as a canary for finding breakages affecting D code we don't know about.

If you conclude that this is not feasible to fix as a deprecation, then there should be a changelog entry describing the breakage, the rationale, and what actions need to be taken to fix the code.

CyberShadow (changes requested):
Adding a "Request changes" review so adding a changelog entry (and/or turning this into a deprecation) isn't forgotten - ping me when this has been taken care of.

WalterBright:
Ping @CyberShadow added changelog

CyberShadow (dismissed):

changelog/fix18736.dd
```diff
@@ -0,0 +1,23 @@
+Issue 18736 - constructor calls allowed after case labels
+
+Delegating constructor calls are not allowed after labels,
+but case labels and default labels should also count as labels.
+
+  class A
```
CyberShadow: This looks like MediaWiki syntax? I think you want to add `---` fences to make this a DDoc code block.

wilzbach: Yep this is still Ddoc and at the moment it doesn't look so great:

http://dtest.dlang.io/artifact/website-f34af6020cfa432dd56bfd6d65ffcd5e6ed2bce4-8920e4bd03f4a6c8df6bd757a2b9f733/web/changelog/pending.html#fix18736

WalterBright: fixed

RazvanN7 (approved):

MoonlightSentinel:
This intoduced a regression: https://issues.dlang.org/show_bug.cgi?id=21080

PR #8147, opened 2018-04-06
[RFC] Added __traits(isRootModule)

marler8997:
Exploring solutions to support faster compilation times.  The proposed trait `isRootModule` trait would return `true` if and only if the module is currently being "compiled" all the way to "object code".

This trait allows a module to include code ONLY when the module is being compiled.  One example of how this could be used is with #8124, you could create "unittest support code" that ONLY gets analyzed/instantiated when the module is being compiled, i.e.
```D
static if (__traits(isRootModule)) version (unittest)
{
    // a bunch of shared code that's only used by unittests in this module
}

unittest
{
   // use the code defined above
}
unittest
{
   // use the code defined above
}
unittest
...
```

To show a "real-world" example of where these semantics would be used, I looked at the `version(unittest)` blocks in `std.path`.  It has 3 of these blocks that define code to support various unittests in the module.  These blocks instantiate templates from other modules which be the cause of slow compilations times.  By prefixing these `version(unittest)` blocks with `static if (__traits(isRootModule))` it would allow applications to unittest their own code without having to analyze all the `version(unittest)` blocks in phobos and all their dependencies.

### DIFF
--- a/src/dmd/id.d
+++ b/src/dmd/id.d
@@ -373,6 +373,7 @@ immutable Msgtable[] msgtable =
     { "isOverrideFunction" },
     { "isStaticFunction" },
     { "isRef" },
+    { "isRootModule" },
     { "isOut" },
     { "isLazy" },
     { "hasMember" },
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -20,6 +20,7 @@ import dmd.arraytypes;
 import dmd.canthrow;
 import dmd.dclass;
 import dmd.declaration;
+import dmd.dmodule;
 import dmd.dscope;
 import dmd.dsymbol;
 import dmd.dsymbolsem;
@@ -122,6 +123,7 @@ shared static this()
         "isOut",
         "isLazy",
         "isReturnOnStack",
+        "isRootModule",
         "hasMember",
         "identifier",
         "getProtection",
@@ -1075,6 +1077,23 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
 
         bool value = Target.isReturnOnStack(tf);
         return new IntegerExp(e.loc, value, Type.tbool);
+    }
+    if (e.ident == Id.isRootModule)
+    {
+        Module mod;
+        if (dim == 0)
+        {
+            mod = sc.instantiatingModule();
+        }
+        else if (dim == 1)
+        {
+            e.error("__traits(isRootModule, <module>) not implemented");
+            return new ErrorExp();
+        }
+        else
+            return dimError(1);
+
+        return new IntegerExp(e.loc, mod.isRoot(), Type.tbool);
     }
     if (e.ident == Id.getFunctionVariadicStyle)
     {
--- a/test/compilable/extra-files/rootmodimport.d
+++ b/test/compilable/extra-files/rootmodimport.d
@@ -1,0 +1,10 @@
+module rootmodimport;
+
+static if (__traits(isRootModule))
+{
+    pragma(msg, "isRootModule 1");
+}
+else
+{
+    pragma(msg, "isRootModule 0");
+}
--- a/test/compilable/extra-files/rootmodmain.d
+++ b/test/compilable/extra-files/rootmodmain.d
@@ -1,0 +1,1 @@
+import rootmodimport;
--- a/test/compilable/isrootmod.sh
+++ b/test/compilable/isrootmod.sh
@@ -1,0 +1,5 @@
+#!/usr/bin/env bash
+
+$DMD -m$MODEL -o- -I=$EXTRA_FILES $EXTRA_FILES/rootmodmain.d 2>&1 | grep -q "isRootModule 0"
+$DMD -m$MODEL -o- -I=$EXTRA_FILES $EXTRA_FILES/rootmodmain.d $EXTRA_FILES/rootmodimport 2>&1 | grep -q "isRootModule 1"
+$DMD -m$MODEL -o- -I=$EXTRA_FILES $EXTRA_FILES/rootmodmain.d -i 2>&1 | grep -q "isRootModule 1"
--- a/test/fail_compilation/isrootmod.d
+++ b/test/fail_compilation/isrootmod.d
@@ -1,0 +1,10 @@
+/*
+PERMUTE_ARGS:
+TEST_OUTPUT:
+---
+fail_compilation/isrootmod.d(10): Error: __traits(isRootModule, <module>) not implemented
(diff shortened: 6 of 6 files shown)


marler8997:
Pinging people involved in https://github.com/dlang/dmd/pull/8124 to get some discussion going.

@wilzbach @andralex @WalterBright @atilaneves @schveiguy @jacob-carlborg @quickfur @jmdavis @ibuclaw 

Huh....more people than I thought....

Anyway, I'm not sure if this PR is the right way forward.  I think the real question is if we want to support the proposed semantics (I'd rather not bikeshed the syntax).

ibuclaw:
Has it been determined what is the reason for compilation slowdown? From what I've seen all attempts to invoke what was thought to be the behaviour in isolation have failed.

Thereby any stab-in-the-dark attempts to would just be wasted time until it is clear what needs fixing (and where).

Has any analysis been made with gdb or callgrind?

marler8997:
> Has it been determined what is the reason for compilation slowdown? From what I've seen all attempts to invoke what was thought to be the behaviour in isolation have failed.

It's pretty well understood at this point. If you take the following:
```D
import std.path;
```

Compilation time will currently slow down by about a factor of 5 if you compile this with `-unittest`.  The ideas is that `-unittest` performance should match non-unittest performance in this case since the app we are compiling isn't doing any unittesting.  There are 3 reasons for this:

1. `unittest` block analysis
2. `version(unittest)` block analysis
3. template instance movement

PR https://github.com/dlang/dmd/pull/8124 addresses 1 and 3 but doesn't disable `version(unittest)` blocks since this results in code breakage.  I can see that implementing each criteria on its own has positive affects on performance with `-unittest`, but it's not until I enable all 3 of these criteria that it matches "non-unittest" performance.

If you want to understand more about why it's these 3 criteria that cause the compilation slowdown you can compile with and without these features along with verbose `-v` enabled.  You can see in the "semantic 3" stage when template instances are analyzed a bunch more modules get imported due to template instances inside `version(unittest)` blocks.  The `version(unittest)` blocks are getting analyzed in imported modules, and then the template instances inside those modules are getting moved to the root module.

schveiguy:
@marler8997 I think what @ibuclaw is saying is, can we find a test case that's smaller than "import a module that imports half of phobos"? I'm going to make another attempt today at finding a small test case that shows where the issue is.

marler8997:
```D
version(unittest)
{
    // import modules and define data structures for use with unittests
    // in this module
}
```

schveiguy:
The results of my investigation are that there are several items contributing to extra imports, the weirdest one being that when compiling with -debug or -unittest, templates are instantiated as if they were in the module being compiled, even if not instantiated there. This results in code that normally is excluded now adding more items to be analyzed, including template unittests. I believe this is the largest source of "extra imports". In addition to the version(unittest) issue we need to deal with this problem.

But in any case, I think we should be very cautious in providing a mechanism that changes how code is generated depending on whether the compiler is generating code for a module or not. One thing to be cautious about is that if an import module isn't being compiled, it doesn't mean the template itself isn't being compiled. I feel we can solve the problems of version(unittest) in different ways.

For instance, std.path's version(unittest) blocks I found to be quite harmless. It includes a struct which is very simple,2 templated types and one templated function that shouldn't affect anything as long as they are only instantiated inside unit tests.

None of the types or blocks include extra imports. There are a few static asserts, but those can be moved inside unittests also.

marler8997:
> But in any case, I think we should be very cautious in providing a mechanism that changes how code is generated depending on whether the compiler is generating code for a module or not. 

Agreed, we should be careful with this.

The `version(unittest)` blocks in `std.path` don't include any imports, so omitting them *might* not change too much.  Note that they still could if they cause certain templates to be instantiated, even if we fix the template instantiation move issue.  However, if you look at other modules in phobos...you can find cases where there are imports inside of `version(unittest) blocks`

```D
// std/ascii.d
version(unittest)
{
    // FIXME: When dmd bug #314 is fixed, make these selective.
    import std.meta; // : AliasSeq;
    import std.range; // : chain;
    import std.traits; // : functionAttributes, FunctionAttribute, isSafe;
}
```

The point is that when you have a module that contains alot of unittesting...you may want to provide common code/imports just for those unittests.  Without support for the proposed semantics, the only way to do this is with `version(unittest)` or copy/pasting the common code in each `unittest` block.  This is not what we want.  I'm not saying this PR is the only solution, but we need to find a solution for this use case.

marler8997:
@schveiguy 

Awesome, you'll have to let me know when your PR is merged and I can mess with the results...see if it solves our performance problems.

Your template wrapper idea is interesting.  Maybe we could make it a pattern, something like
```D
version(unittest) template Unittest()
{
    // imports/functions/data structures just for unittests
    import std.path;
    struct Foo
    {
        void doStuff();
    }
}

unittest
{
    auto foo = Unittest.Foo();
    foo.doStuff();
}
```

schveiguy:
Yep, except it would be `Unittest!().Foo` 😝 

I don't know if we need that level of cleverness yet. I'm still not 100% sure when imports will happen and when they won't regarding unittests.

marler8997:
Yea, `Unittest!().Foo` is pretty ugly...I edited my comment to include an alias version :)  `ut.Foo`.  I'm very confident that imports inside a template will be loaded "if and only if" the template instantiated, which is exactly the semantics we want in this case.

marler8997:
We are essentially using `templates` to cause "lazy" loading of code/imports for unittests.  Similar to the `from` template that was talked about in the D blog a while back.

schveiguy:
> I edited my comment to include an alias version

Wouldn't that have the same problem as today (assuming we didn't touch version(unittest) semantics)? I mean, an importing module sees the alias instantiation. I don't know if private affects it or not.

A horrid hack workaround could be to put all the declarations inside a voldemort type, and then you use IFTI to access it (don't need explicit instantiation then).

```D
version(unittest) private auto Unittest()()
{
   static struct UT
   {
      struct Foo
      {
         void doStuff() {}
      }
   }
   return UT.init;
}

unittest
{
    auto foo = Unittest.Foo()
    foo.doStuff();
}
```

marler8997:
> Wouldn't that have the same problem as today (assuming we didn't touch version(unittest) semantics)? I mean, an importing module sees the alias instantiation.

Not sure...would need to mess with it.

jacob-carlborg:
> Another possibility is to define templates inside version(unittest) that will only do the imports when unittests are being compiled.

So the problem is templates and the solution is to add more templates :joy:.

marler8997:
Cleaning up old PRs that aren't going anywhere

PR #8148, opened 2018-04-06
refactor: redo CSX merge logic layout and names

WalterBright:
The idea is to:

* push CSX logic into ctorflow.d
* push error handling into dscope.d
* use `a` and `b` for left and right branches

### DIFF
--- a/src/dmd/ctorflow.d
+++ b/src/dmd/ctorflow.d
@@ -91,7 +91,7 @@ struct CtorFlow
      * Params:
      *  ctorflow = bits to OR in
      */
-    void OR(const ref CtorFlow ctorflow)
+    void OR(const ref CtorFlow ctorflow) pure nothrow
     {
         callSuper |= ctorflow.callSuper;
         if (fieldinit.length && ctorflow.fieldinit.length)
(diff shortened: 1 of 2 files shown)


src/dmd/ctorflow.d
```diff
@@ -105,53 +105,118 @@ struct CtorFlow
...
- *      fieldInit = the path to merge fi into
- *      fi = the other path
+ *      a = the path to merge `b` into
+ *      b = the other path
  * Returns:
- *      false means either fieldInit or fi skips initialization
+ *      false means one of the paths skips construction
  */
-bool mergeFieldInitX(ref CSX fieldInit, const CSX fi)
+bool mergeCallSuper(ref CSX a, const CSX b) pure nothrow
```
jacob-carlborg: Why not call it `leftBranch` and `rightBranch`, if that is what it means?

wilzbach: Maybe `lhs` and `rhs` if brevity is really needed?

WalterBright: Please, no.

* What follows in the implementation is essentially math, and in math one uses one letter variable names for good reason. We write `sin x`, not `sin theAngleInRadians`.

* One letter names are easiest to comprehend when one has complex math expressions.

* The algorithm is commutative, i.e. there is no `left` or `right`. Giving such names implies there is an asymmetry when there is none.

* Also, it's good style to use shorter names the more local they are.

src/dmd/ctorflow.d
```diff
@@ -105,53 +105,118 @@ struct CtorFlow
...
         {
-            ok = true;
-            fieldInit = CSX.halt;
+            a = CSX.halt;
         }
-        else if (!aHalt && aRet)
+        else if ((!bHalt && bRet && !bAny && aAny) || (!aHalt && aRet && !aAny && bAny))
+        {
+            // If one has returned without a constructor call, there must be never
+            // have been ctor calls in the other.
```
JinShil: "there must be never have been" doesn't make sense to me.

src/dmd/ctorflow.d
andralex: The names are fine. The code is wanting. Reviewers should suggest material improvement, not trivial pursuits.

src/dmd/ctorflow.d
```diff
@@ -105,53 +105,118 @@ struct CtorFlow
...
-bool mergeFieldInitX(ref CSX fieldInit, const CSX fi)
+bool mergeCallSuper(ref CSX a, const CSX b) pure nothrow
 {
-    if (fi != fieldInit)
+    // This does a primitive flow analysis to support the restrictions
+    // regarding when and how constructors can appear.
+    // It merges the results of two paths.
+    // The two paths are `a` and `b`; the result is merged into `a`.
+    bool ok = true;
+    if (b != a)
```
andralex: The function does nothing and returns `true` if `a == b`, yet it takes the reader a whopping 47 lines to figure that out. Can we please please please with Truvia on top renounce this crap:

```d
bool ok = true;
if (condition)
{
    ... so many lines ...
}
return ok;
```

This antipattern is all over the compiler code. It's insufferable in old code, and definitely not tolerable in the sligthest in new code. We're not coding Fortran over here. Please:

```d
if (!condition)
{
    return true;
}
... knock yourself out ...
```

src/dmd/ctorflow.d
```diff
@@ -105,53 +105,118 @@ struct CtorFlow
...
     {
+        // Have ALL branches called a constructor?
+        int aAll = (a & (CSX.this_ctor | CSX.super_ctor)) != 0;
+        int bAll = (b & (CSX.this_ctor | CSX.super_ctor)) != 0;
```
andralex: These are `bool` really, not `int`. And they should be immutable. Just use `immutable` or `const` and let type deduction say what you mean.

src/dmd/ctorflow.d
```diff
@@ -105,53 +105,118 @@ struct CtorFlow
...
+        // Have ANY branches called a constructor?
+        bool aAny = (a & CSX.any_ctor) != 0;
+        bool bAny = (b & CSX.any_ctor) != 0;
```
andralex: immutable throughout

src/dmd/ctorflow.d
```diff
@@ -105,53 +105,118 @@ struct CtorFlow
...
-            ok = true;
-            fieldInit = CSX.halt;
+            a = CSX.halt;
         }
-        else if (!aHalt && aRet)
+        else if ((!bHalt && bRet && !bAny && aAny) || (!aHalt && aRet && !aAny && bAny))
+        {
+            // If one has returned without a constructor call, there must not
+            // be ctor calls in the other.
+            ok = false;
```
andralex: Just return - all of this `ok` variable nonsense is another antipattern

src/dmd/ctorflow.d
```diff
@@ -105,53 +105,118 @@ struct CtorFlow
...
+ * Merge `b` flow analysis results into `a`.
+ * Params:
+ *      a = the path to merge `b` into
+ *      b = the other path
+ * Returns:
+ *      false means either `a` or `b` skips initialization
+ */
+bool mergeFieldInit(ref CSX a, const CSX b) pure nothrow
+{
+    if (b != a)
```
andralex: here we go again

andralex (changes requested):

src/dmd/ctorflow.d
WalterBright: Well, I beg that that code was all copy-pasta from the original.

RazvanN7 (approved):

andralex (approved):
Nice, thx! So what's with the failures?

RazvanN7:
@andralex the circleCi failure seems unrelated. @WalterBright please give it another push so that you will restart the failing test.

wilzbach:
You can restart CircleCi directly by going there, logging in via GitHub and hitting restart.

wilzbach:
(I just did so)

PR #8150, opened 2018-04-06
Do not run auto-tester build by default on Windows

MetaLang:


MetaLang:
I called it "defaulttarget" because that's what it is in win64.mak.

wilzbach (approved):
Meh in src/win32.mak too

MetaLang:
Thanks!

PR #8151, opened 2018-04-06
Determine module isRoot earlier

marler8997:
This PR is in preparation for other functionality (such as skipping unittests in imported modules).  It modifies the code to determine if a module is a "root module" (a module being compiled to object code) as soon as the module name has been determined which is just after the module declaration has been parsed.


marler8997:
Ping

PR #8152, opened 2018-04-09
use special enums for __c_long, etc. instead of special structs

WalterBright:
The existing mechanism is in core.stdc.config, which uses a struct named __c_long. This change enables using an enum instead, which has some advantages:

* enums are a lot closer to typedefs
* we can allow implicit conversion of other integers to these special enums
* linking with druntime will not be required

Later we can do away with core.stdc.config entirely and just predefine these enums in the compiler.

The tests will be when core.stdc.config is redone using enums, which will be the egg (or the chicken?).

### DIFF
--- a/src/dmd/glue.d
+++ b/src/dmd/glue.d
@@ -1434,8 +1434,19 @@ uint totym(Type tx)
             break;
 
         case Tenum:
-            t = totym(tx.toBasetype());
+        {
+            Type tb = tx.toBasetype();
+            const id = tx.toDsymbol(null).ident;
+            if (id == Id.__c_long)
+                t = tb.ty == Tint32 ? TYlong : TYllong;
+            else if (id == Id.__c_ulong)
+                t = tb.ty == Tuns32 ? TYulong : TYullong;
+            else if (id == Id.__c_long_double)
+                t = TYdouble;
+            else
+                t = totym(tb);
             break;
+        }
 
         case Tident:
         case Ttypeof:
--- a/test/runnable/cppa.d
+++ b/test/runnable/cppa.d
@@ -1,4 +1,4 @@
-// PERMUTE_ARGS: -g
+// PERMUTE_ARGS: -g -version=PULL8152
 // EXTRA_CPP_SOURCES: cppb.cpp
 
 import core.stdc.stdio;
@@ -605,13 +605,19 @@ extern(C++)
 
 version (CRuntime_Microsoft)
 {
-    struct __c_long_double
+    version (PULL8152)
     {
-        this(double d) { ld = d; }
-        double ld;
-        alias ld this;
+        enum __c_long_double : double;
     }
-
+    else
+    {
+        struct __c_long_double
+        {
+            this(double d) { ld = d; }
+            double ld;
+            alias ld this;
+        }
+    }
     alias __c_long_double myld;
 }
 else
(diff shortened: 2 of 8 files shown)


jacob-carlborg:
How will the usage of this look like?

kinke:
I guess something like
```D
enum __c_long : long;
void foo(__c_long c);
void bar() { foo(123); }
```

jacob-carlborg:
```d
enum __c_long : long;
alias c_long = __c_long;

void foo(c_long a){};

void main()
{
    foo(3L);
}
```

The above results in:

```
Error: function main.foo (__c_long a) is not callable using argument types (long)
```

jacob-carlborg:
@WalterBright would you consider trying to get this working instead https://github.com/dlang/dmd/pull/7458. That is, adding support for `pragma(mangle)` to alias declarations.

WalterBright:
> That is, adding support for pragma(mangle) to alias declarations.

It's completely antiethical to the way alias works. Aliases are internally instantly replaced with their aliased type all through the compiler. Enums are not, which is why doing special enums is a very small change.

jacob-carlborg:
What's the advantage of using an enum instead of a struct?

jacob-carlborg:
> It's completely antiethical to the way alias works. Aliases are internally instantly replaced with their aliased type all through the compiler. Enums are not, which is why doing special enums is a very small change.

That might be the case but it would be very convenient.

WalterBright:
> What's the advantage of using an enum instead of a struct?

* enums are a lot closer to typedefs
* we can allow implicit conversion of other integers to these special enums
* linking with druntime will not be required

WalterBright:
Ok I added some tests to cppa.d

jacob-carlborg (changes requested):
Can you add a usage test as well that verifies that the implicit conversions work too.

src/dmd/glue.d
```diff
@@ -1434,8 +1434,18 @@ uint totym(Type tx)
...
         case Tenum:
-            t = totym(tx.toBasetype());
+        {
+            const id = tx.toDsymbol(null).ident;
+            if (id == Id.__c_long)
+                t = TYlong;
+            else if (id == Id.__c_ulong)
+                t = TYulong;
+            else if (id == Id.__c_long_double)
+                t = TYdouble;
```
ibuclaw: I take it that tydouble is correct here, and that there is no long double ty.

ibuclaw (commented):
One thing that I don't think is handled here but probably will need to be in druntime are type properties. Wouldn't you need to declare enum members named init, max, min, etc... Unless there is special handling in the compiler that I'm not seeing here.

jacob-carlborg (approved):

test/runnable/cppa.d
```diff
@@ -605,13 +605,19 @@ extern(C++)
 
 version (CRuntime_Microsoft)
 {
-    struct __c_long_double
+    version (PULL8152)
```
jacob-carlborg: What's the reasoning for the version statement? Will this test be run on a compiler that doesn't contain this change?

WalterBright: Can't have two global declarations of __c_long, etc.

WalterBright:
> Unless there is special handling in the compiler that I'm not seeing here.

The `.init` was there, but not `.max`/`.min`. I added them.

PR #8154, opened 2018-04-09
[RFC] Debug Logging in DMD

marler8997:
NOTE: this PR does not contain any changes...it's just a placeholder for discussion

DMD currently uses commented `printf`/`fprintf` statements for "debug logging".
```D
// printf("here's some info about %s: %s", foo.toChars(), foo.getInfo());
```
Because these log statements are commented, they tend to break since don't get checked on normal builds.  Another annoyance is that many times `printf` isn't defined so you need to add `import core.std.stdio` somewhere in the module any time you want to enable some of these for debugging.

To make this less of a pain, I propose we define a function used for these types of log messages that still get analyzed but don't actually do anything unless the developer changes something in the code.  Something like this:
```D
module dmd.log;

// Set to true to enable all debug logging
private enum enableAllDebug = false;

// print debug message to stderr only if `enableAllDebug` is true
pragma(inline) void debugf(T...)(const(char)* format, T args)
{
    static if (enableAllDebug)
    {
        fprintf(stderr, format, args);
    }
}
// print debug message to stderr even if `enableAllDebug` is false
pragma(inline) void forcedebugf(T...)(const(char)* format, T args)
{
    fprintf(stderr, format, args);
}
```
So...
```D
// instead of this
// printf("here's some info about %s: %s", foo.toChars(), foo.getInfo());

// we would have this
debugf("here's some info about %s: %s", foo.toChars(), foo.getInfo());
```
This addresses the 2 issues, `debugf` calls won't break since they aren't commented and they don't require adding an import statement to turn on.   To turn on individual statements, you can prefix the function call with `force`, i.e.
```D
debugf("here's some info about %s: %s", foo.toChars(), foo.getInfo());
// prefix the call with "force" to forcefully enable it...
forcedebugf("here's some info about %s: %s", foo.toChars(), foo.getInfo());
```

# Debug Kinds

In addition to this, we could also create "categories" for our debugging.  Maybe a developer wants to print debug related to "optimization" or "glue layer details".  By creating categories, a developer could enable all debug logging for particular categories.  Maybe something like this:
```D
enum DebugKind
{
    optimization,
    glue,
    semantic2,
   ...
}
immutable debugEnabled = [
    DebugKind.optimization: false,
    DebugKind.glue: false,
    DebugKind.semantic2: false,
    ...
];


pragma(inline) void debugf(T...)(DebugKind kind, const(char)* format, T args)
{
    static if (debugEnabled[kind])
    {
        fprintf(stderr, format, args);
    }
}

```
So...
```D
debugf(DebugKind.glue, "my cool info %s", glueObject.toChars());
```


jacob-carlborg:
One advantage of the current approach is that I can enable a single `printf` to avoid a lot of unwanted output.

marler8997:
> One advantage of the current approach is that I can enable a single printf to avoid a lot of unwanted output.

Yes this is a critical use case. Still works with this proposal, i.e.
```D
// Change from this
debugf("here's some info about %s: %s", foo.toChars(), foo.getInfo());
// to this
forcedebugf("here's some info about %s: %s", foo.toChars(), foo.getInfo());
```

The proposal is to prefix the function name with `force` but we could of course have it be a parameter to the function as well.  Whatever people like more.

UplinkCoder:
Since I don't have to work on recent dmd versions, do whatever.
But I would not approve. and I don't think walter would either.

marler8997:
@UplinkCoder 

Thanks for the input.  I was also thinking a "performance" category could be used to print output similar to what I saw in your youtube video.  Care to elaborate on why you would not approve?

wilzbach:
> Care to elaborate on why you would not approve?

:+1: 
If you already take the time to comment, please do so in a productive and constructive way.
Just "I hate this" doesn't help anyone.

ibuclaw:
FYI: https://github.com/dlang/dmd/pull/7198#discussion_r143397225

ibuclaw:
> NOTE: this PR does not contain any changes...it's just a placeholder for discussion

Gah... Use the NG then, which is the proper place to discuss this, or better yet raise a bugzilla report.  Github is not the place for general chit-chat.

marler8997:
Thanks for the reference @ibuclaw 

So Walter indicates there he likes this pattern:
```D
enum log= false;
...
if (log) printf("blah blah %s", something.toChar());
```

This is an improvement to the current situation.  So we have 2 proposals on the table

1. Introduce dmd.log
2. Use the "enum log" pattern in each module going forward.

marler8997:
> Gah... Use the NG then, which is the proper place to discuss this, or better yet raise a bugzilla report. Github is not the place for general chit-chat.

I chose to use Github for this since the conversation tends to be more productive on this platform.

marler8997:
@klickverbot 

lol...was planning on pushing the changes to this PR when a design was agreed upon.  I was actually in the middle of putting something together when you closed it :)

marler8997:
Odd, it won't let me reopen it....?

marler8997:
Are you able to re-open it @klickverbot ?

marler8997:
Weird...it made me create a whole new PR....hmmm

I guess I'll just reference this one so the conversation does not get lost.

ibuclaw:
> I guess I'll just reference this one so the conversation does not get lost.

Exactly the reason for opening a bugzilla issue first to discuss details.

marler8997:
You're saying I shouldn't use github to discuss features because someone might close the PR and I wont be able to reopen it, resulting in some of the conversation being split between 2 places?

And how is that different from some of the conversation being on bugzilla? Also, I'm suppose to assume that maintainers will close my PRs so I shouldnt push anything to github unless there full discussion and consensus on bugzilla first? I wasnt aware of this...thanks for letting me know.

marler8997:
@ibuclaw please stop talking down to me. Its extremely disrespectful and unhelpful. Taking any chance you get to make a snide comment is getting very old.

ibuclaw:
Bugzilla is _the_ tool we use to keep track of any work that needs to be done. Be it a bug, a feature, needed refactoring, or just an administration task.

The NG can be used for preliminary discussions, it doesn't have to be in general, which I think you alluded to not being "productive",  we do have other channels to discuss compiler internals.

ibuclaw:
We have had platforms and processes in place for well over a decade now to open such discussions. The friendly reminder is to use them as other active core contributors do, rather than reinventing the wheel.

wilzbach:
FWIW I fully understand why Jonathan opened this as a PR:

- not many people look at Bugzilla issues actively (except for a few exceptions of famous, 10-year-old bugs) and I think if this would have been created as a Bugzilla issue that there wouldn't have been a "reply" within the first two weeks
- the NG will be off-topic with the first page. And even if there's some consensus like for the recent Markdown thread, people who don't read the NG will still object here.
- not everyone uses Slack - especially Walter whose approval is required

marler8997:
And thats the 3rd time you've mentioned it. I understand we disagree on my choice of platform for this situation...drop it and move on to prductive discussion.

PR #8155, opened 2018-04-09
Add Expression-Based Contract Syntax

tgehr:
Implementation of [DIP 1009](https://github.com/dlang/DIPs/blob/master/DIPs/accepted/DIP1009.md).

### DIFF
--- a/src/dmd/arraytypes.h
+++ b/src/dmd/arraytypes.h
@@ -67,4 +67,6 @@ typedef Array<class GotoStatement *> GotoStatements;
 
 typedef Array<class TemplateInstance *> TemplateInstances;
 
+typedef Array<struct Ensure> Ensures;
+
 #endif
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -44,6 +44,7 @@ struct ASTBase
     alias Catches               = Array!(Catch);
     alias Identifiers           = Array!(Identifier);
     alias Initializers          = Array!(Initializer);
+    alias Ensures               = Array!(Ensure);
 
     enum Sizeok : int
     {
@@ -674,13 +675,18 @@ struct ASTBase
         }
     }
 
+    extern (C++) struct Ensure
+    {
+        Identifier id;
+        Statement ensure;
+    }
+
     extern (C++) class FuncDeclaration : Declaration
     {
         Statement fbody;
-        Statement frequire;
-        Statement fensure;
+        Statements* frequires;
+        Ensures* fensures;
         Loc endloc;
-        Identifier outId;
         StorageClass storage_class;
         Type type;
         bool inferRetType;
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -153,12 +153,47 @@ enum FUNCFLAG : uint
 }
 
 /***********************************************************
+ * Tuple of result identifier (possibly null) and statement.
+ * This is used to store out contracts: out(id){ ensure }
+ */
+extern (C++) struct Ensure
+{
+    Identifier id;
+    Statement ensure;
+
+    Ensure syntaxCopy()
+    {
+        return Ensure(id, ensure.syntaxCopy());
+    }
+
+    /*****************************************
+     * Do syntax copy of an array of Ensure's.
+     */
+    static Ensures* arraySyntaxCopy(Ensures* a)
+    {
+        Ensures* b = null;
+        if (a)
+        {
+            b = a.copy();
+            foreach (i, e; *a)
+            {
+                (*b)[i] = e.syntaxCopy();
+            }
+        }
+        return b;
+    }
+
+}
+
+/***********************************************************
  */
 extern (C++) class FuncDeclaration : Declaration
 {
     Types* fthrows;                     /// Array of Type's of exceptions (not used)
-    Statement frequire;                 /// in contract body
-    Statement fensure;                  /// out contract body
+    Statements* frequires;              /// in contracts
+    Ensures* fensures;                  /// out contracts
+    Statement frequire;                 /// lowered in contract
+    Statement fensure;                  /// lowered out contract
     Statement fbody;                    /// function body
 
     FuncDeclarations foverrides;        /// functions this function overrides
@@ -167,8 +202,7 @@ extern (C++) class FuncDeclaration : Declaration
 
     const(char)* mangleString;          /// mangled symbol created from mangleExact()
 
-    Identifier outId;                   /// identifier for out statement
-    VarDeclaration vresult;             /// variable corresponding to outId
+    VarDeclaration vresult;             /// result variable for out contracts
     LabelDsymbol returnLabel;           /// where the return goes
 
     // used to prevent symbols in different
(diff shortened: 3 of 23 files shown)


src/dmd/arraytypes.h
```diff
@@ -67,4 +67,6 @@ typedef Array<class GotoStatement *> GotoStatements;
 
 typedef Array<class TemplateInstance *> TemplateInstances;
 
+typedef Array<class Ensure> Ensures;
```
thewilsonator: Should this not be `struct Ensure`?

src/dmd/func.d
```diff
@@ -154,11 +154,40 @@ enum FUNCFLAG : uint
 
 /***********************************************************
  */
+
+extern (C++) struct Ensure
```
ibuclaw: Wouldn't this be better as `class EnsureStatement : Statement`? Then it can go in the statement module.

src/dmd/func.d
```diff
@@ -167,7 +196,13 @@ extern (C++) class FuncDeclaration : Declaration
 
     const(char)* mangleString;          /// mangled symbol created from mangleExact()
 
-    Identifier outId;                   /// identifier for out statement
+    @property Identifier outId(){       /// identifier for result
```
ibuclaw: This would be dead code now, so remove?

src/dmd/parse.d
```diff
@@ -4715,36 +4740,90 @@ final class Parser(AST) : Lexer
             }
 
         case TOK.in_:
+            auto loc = token.loc;
```
ibuclaw: Add a couple comments presenting the supported `in` syntaxes as per `invariant` above.

src/dmd/parse.d
```diff
@@ -4715,36 +4740,90 @@ final class Parser(AST) : Lexer
...
+            else
+            {
+                f.frequires.push(parseStatement(ParseStatementFlags.curly | ParseStatementFlags.scope_));
+                requireDo = true;
+            }
             goto L1;
 
         case TOK.out_:
             // parse: out (identifier) { statement }
+            auto loc = token.loc;
```
ibuclaw: Likewise here, add comments for syntaxes that are now handled.

src/dmd/statement.d
```diff
@@ -85,6 +85,19 @@ extern (C++) abstract class Statement : RootObject
         assert(0);
     }
 
+    static Statements* arraySyntaxCopy(Statements* a){
```
ibuclaw: Document

src/dmd/astbase.d
```diff
@@ -674,13 +675,18 @@ struct ASTBase
...
+        Statement ensure;
+    }
+
     extern (C++) class FuncDeclaration : Declaration
     {
         Statement fbody;
-        Statement frequire;
-        Statement fensure;
+        Statements* frequires;
+        Ensures* fensures;
```
ibuclaw: Document fields.

ibuclaw (changes requested):
The most notable thing missing here is documentation, both for new code blocks added and ddoc entries for functions and types introduced.

Other than that, just a couple notes on implementation detail that are trivial or could be explained away by adding documentation for why it's needed.

ibuclaw:
I think this could be ironed out within a week ready for inclusion.

src/dmd/func.d
```diff
@@ -2031,14 +2065,55 @@ extern (C++) class FuncDeclaration : Declaration
...
+            {
+                if (r.id)
+                {
+                    auto rloc = r.ensure.loc;
+                    auto a = new VarDeclaration(rloc, null, r.id, new ExpInitializer(rloc, new IdentifierExp(rloc, Id.result)), STC.ref_ | STC.temp | STC.result);
+                    auto o = new Objects;
+                    o.push(new TypeIdentifier(rloc,Id.result));
+                    import dmd.cond: StaticIfCondition;
+                    auto c = new StaticIfCondition(rloc, new TraitsExp(loc, Id.compiles, o));
+                    auto result = new ConditionalStatement(rloc, c, new ExpStatement(rloc, a), null);
```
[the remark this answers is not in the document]
tgehr: This was a hack to improve diagnostics. (`__result` should not appear in the output.)
The `static if` condition always passes in well-formed D programs, but will fail if out contracts of `void` functions specify a result identifier. I'd prefer to remove this. I can just not semantically analyze the out contracts if one of them specifies an output but the result is `void`, I guess.

src/dmd/func.d
tgehr: It is not really a statement, it's just a way to create a tuple of result identifier and out contract.
I can of course declare it as a statement if that is better.

src/dmd/func.d
tgehr: It is not dead code, but I can turn it into a more appropriately-named `bool` function.

src/dmd/arraytypes.h
tgehr: It does not make a difference in code semantics, but I guess using `struct` is better style.

src/dmd/statement.d
```diff
@@ -85,6 +85,22 @@ extern (C++) abstract class Statement : RootObject
         assert(0);
     }
 
+    /*************************************
+     * Do syntax copy of an array of Statement's.
+     */
+    static Statements* arraySyntaxCopy(Statements* a){
```
ibuclaw: Curly braces on next line, check for this style everywhere.

src/dmd/func.d
```diff
@@ -152,13 +152,48 @@ enum FUNCFLAG : uint
...
+    /*****************************************
+     * Do syntax copy of an array of Ensure's.
+     */
+    static Ensures* arraySyntaxCopy(Ensures* a)
+    {
+        Ensures* b = null;
+        if (a)
+        {
+            b = a.copy();
+            foreach(i,e;*a)
```
ibuclaw: Add spaces in foreach (`i, e; *a`) - check for this style everywhere as I've seen it more than once.

src/dmd/func.d
```diff
@@ -1891,6 +1924,17 @@ extern (C++) class FuncDeclaration : Declaration
         return false;
     }
 
+    /****************************************************
+     * Check whether result variable can be built.
+     * Returns:
+     *     `true` if the function has a return type that
+     *     is different from `void`.
+     */
+    final bool canBuildResultVar(){
```
ibuclaw: I see that this is only used by the module its declared in. So make it private, and extern(D).

tgehr:
I have addressed all of @ibuclaw 's comments and added a changelog entry. The spec pull request is https://github.com/dlang/dlang.org/pull/2339

Is there anything else left to do? @ibuclaw @WalterBright @wilzbach

JinShil:
Can you add more test coverage?  codecov only reports 73% with the largest void being in the hdrgen code.

tgehr:
@JinShil Probably not, but I can delete code. (I think most of what is not covered in hdrgen it is currently dead code.)

JinShil:
I'm referring to coverage within this PR itself.  The hdrgen code that isn't covered is new code in this PR.  See https://codecov.io/gh/dlang/dmd/compare/b4d5005db0fd3581789d8a0942583f7c13efbc82...772309cbccb1c0d3e3cd9892f1ebc36024e5f7b7/diff

Ideally, we should have 100% coverage for all code in this PR.

tgehr:
So am I. I know.

JinShil:
If its dead code, why was it added?

tgehr:
There is an obvious way to handle the case other than an assertion failure. But if coverage is more important than robustness to changes to invariants in other parts of the code, I can change that.

JinShil:
Sorry, I don't understand what you're implying.  I think all that is required is a test to generate a .di file and verify that the result is correct.  That should exercise the new code in hdrgen.d

tgehr:
Currently, a contract is always either a ScopeStatement or an AssertStatement. The header generation code also handles the case when it is a different statement. But yes, you are right, the AssertStatement case can still be covered. I'll add a test and delete the dead code.

tgehr:
@JinShil @ibuclaw @klickverbot @RazvanN7 @WalterBright Gentle reminder. (The remaining few lines of uncovered code cannot be covered in an end-to-end test as far as I can tell, but they should be there for completeness.)

RazvanN7 (approved):

andralex (approved):
Cool, thanks! Please follow with a spec PR.

andralex:
cc @ibuclaw

wilzbach:
@andralex FYI: there's already a PR to dlang.org: https://github.com/dlang/dlang.org/pull/2339

ibuclaw (approved):
:see_no_evil:

PR #8156, opened 2018-04-09
[WIP][Needs Approval] Debug Logging in DMD

marler8997:
Reopened https://github.com/dlang/dmd/pull/8154

Current implementation is based on Walter's preference for logging described here: 

https://github.com/dlang/dmd/pull/7198#discussion_r143397225

This PR introduces following logging pattern:
```D
import core.stdc.stdio;

private enum LOG = false;

if (LOG) printf(...);
```

This replaces the current pattern of commented printf statements:
```D
// printf(...);
```

The new pattern makes it easier to turn log statements on and keeps log statements from breaking since having them uncommented forces their arguments to be analyzed.  It also keeps developers from removing the `import core.stdc.stdio` because doing so will break the log statements now. Note that `if (LOG)` should be used instead of `static if (LOG)`, otherwise, the statements don't get analyzed when they are disabled.

PR currently introduces the pattern in the first 3 modules `access.d` and `aggregate.d`  and `apply.d` (determined alphabetically) to show proof of concept.  Note that 2 out of the first 3 modules had errors in their printf statements...

### DIFF
--- a/src/dmd/access.d
+++ b/src/dmd/access.d
@@ -12,6 +12,7 @@
 
 module dmd.access;
 
+import core.stdc.stdio;
 import dmd.aggregate;
 import dmd.dclass;
 import dmd.declaration;
@@ -34,7 +35,7 @@ private enum LOG = false;
 private Prot getAccess(AggregateDeclaration ad, Dsymbol smember)
 {
     Prot access_ret = Prot(Prot.Kind.none);
-    static if (LOG)
+    if (LOG)
     {
         printf("+AggregateDeclaration::getAccess(this = '%s', smember = '%s')\n", ad.toChars(), smember.toChars());
     }
@@ -75,7 +76,7 @@ private Prot getAccess(AggregateDeclaration ad, Dsymbol smember)
             }
         }
     }
-    static if (LOG)
+    if (LOG)
     {
         printf("-AggregateDeclaration::getAccess(this = '%s', smember = '%s') = %d\n", ad.toChars(), smember.toChars(), access_ret);
     }
@@ -139,14 +140,14 @@ extern (C++) bool checkAccess(AggregateDeclaration ad, Loc loc, Scope* sc, Dsymb
 {
     FuncDeclaration f = sc.func;
     AggregateDeclaration cdscope = sc.getStructClassScope();
-    static if (LOG)
+    if (LOG)
     {
         printf("AggregateDeclaration::checkAccess() for %s.%s in function %s() in scope %s\n", ad.toChars(), smember.toChars(), f ? f.toChars() : null, cdscope ? cdscope.toChars() : null);
     }
     Dsymbol smemberparent = smember.toParent();
     if (!smemberparent || !smemberparent.isAggregateDeclaration())
     {
-        static if (LOG)
+        if (LOG)
         {
             printf("not an aggregate member\n");
         }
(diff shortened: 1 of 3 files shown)


marler8997:
Old PR was closed and I'm unable to reopen...pinging reviewers from original PR:

@jacob-carlborg @UplinkCoder @wilzbach @ibuclaw @klickverbot

JinShil:
Forgive the silly question, but why `if` instead of `static if`?  Won't that cause it to evaluate at runtime?

marler8997:
The log statements don't get analyzed when using 'static if' defeating the whole purpose of the PR. However, if the optimizer can't handle removing these log statements....then we should hang our heads in shame

jacob-carlborg:
> Forgive the silly question, but why if instead of static if? Won't that cause it to evaluate at runtime?

Yes. The assumption is that the compiler will optimized that away. The reason is to make sure the compiler will to semantic analysis on the code in the if statement. Currently there are a lot of debug calls that are commented out that contains invalid code.

WalterBright:
For similar reasons to the other logging PR I oppose this one. In addition, I have a hard time seeing the utility in a 40 line module for one function declaration and 2 constants. It's just not in proportion.

marler8997:
@WalterBright thanks for chiming in.  I'm not married to any particular implementation, but I do want to make it easier to turn debugging on.  The problem is that today you have to

1. uncomment the log statement
2. fix it (they are often broken since they aren't analyzed)
3. add `import core.std.stdio;`

I'd prefer if we had a way to make sure log statements didn't break (were analyzed) and were easy to turn on.  Do you have a solution to solve these issues that you would endorse?

marler8997:
Ok, I've updated the PR, no more `dmd.log`.  Now it establishes the following pattern for logging:
```D
import core.stdc.stdio;

enum LOG = false;

if (LOG) printf(...);
```
This pattern makes sure that log statements are analyzed and also makes them easy to turn on both individually and at the module level (if on is so inclined).  I'd even be fine with this:
```D
if (false) printf(...);
```

WalterBright:
> uncomment and fix them since they are often broken from not being analyzed

I run across that too, and just fix 'em when I need 'em.

> add import core.stdc.stdio;

I keep adding that back in, someone keeps deleting it :-)
Please, folks, just leave it in.

> I'd prefer if we had a way to make sure log statements didn't break (were analyzed) and were easy to turn on without having to add extra imports. Do you have a solution to solve these issues that you would endorse?

I just don't think these issues rise to the level of being a problem.

I merely comment/uncomment single line debugging statements, and use `enum log` for a group of related ones. I like your idea of making `printf` trusted, but I think that would be better implemented as allowing system code in `debug` statements. Might be a bit tricky to implement.

I'm also a rather negative curmudgeon when it comes to logging packages, as you might suspect, so I stayed out of any `std.log` discussions.

marler8997:
>> uncomment and fix them since they are often broken from not being analyzed

> I run across that too, and just fix 'em when I need 'em.

Like you say, this isn't a big problem but it's an annoyance nonetheless.  Why not solve this small annoyance with a simple change if we can?  The PR no longer contains `dmd.log`, the proposal is simply to uncomment log statements and put them behind an `if(LOG)` or `if(false)` (whichever you prefer).

>> add import core.stdc.stdio;

> I keep adding that back in, someone keeps deleting it :-)
Please, folks, just leave it in.

With this PR, if someone removes it you will get a build error.  Now people have to leave them in :)

JinShil:
We need to be trying to move PRs to acceptance or closure.  Leaving them in limbo is the *worst* thing we can do.  

I don't think this is such a bad idea.  I don't think it's a priority, but some probably think may of the PRs I submit are not a priority, too.  If it were up to me, I'd merge this, but if @WalterBright opposes this, I don't think there's much I can do.  

So, IMO, we either need to establish some protocol for overriding @WalterBright, or we need to close this.

marler8997:
Keep in mind that the goal of this PR is to establish a consistent logging pattern throughout dmd where the logging doesnt break (meaning its analyzed by the compiler). The changes exist as a demonstration in a few modules, but the goal is to have leadersip either accept this design for all of dmd or come up with another solution to this annoyance.

PR #8157, opened 2018-04-10
[RFC] Debug Logging in DMD (Option 2)

marler8997:
The "more liberal" alternative to https://github.com/dlang/dmd/pull/8156

This one defines the function `logf` as a wrapper around logging instead of having other modules call `printf` directly.

Instead of:
```D
if (LOG) printf(...);
```
You have
```D
logf(...);
```

In the case that you have "setup" to do for the log, you can do this
```D
if (logEnabled)
{
    // ... setup
    logf(...);
}
```

If you want to enable ALL log messages, you change the `allLoggingEnabled` enum in `dmd/log.d` to `true`.  If you want to turn on individual log statements you can change `logf` to `forcelogf`.
```D
forcelogf(...);
```

If you want to enable all the log statements in particular modules, you can do this in the `logEnabled` function in `log.d`.

Note that one advantage of this PR over #8156 is that it allows developers to customize all logging by changing code in a single location rather than having to change it in every call.  For example, sometimes you may want to log to `stderr` instead of `stdout` so that the log messages are in "order" with error messages, with this PR you could change this in one place rather than finding and changing every call.

### DIFF
--- a/src/dmd/access.d
+++ b/src/dmd/access.d
@@ -23,10 +23,9 @@ import dmd.errors;
 import dmd.expression;
 import dmd.func;
 import dmd.globals;
+import dmd.log;
 import dmd.mtype;
 import dmd.tokens;
-
-private enum LOG = false;
 
 /****************************************
  * Return Prot access for Dsymbol smember in this declaration.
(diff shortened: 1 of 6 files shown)


JinShil:
I think I prefer this over #8156 but I like the idea of enumerating the different kinds of log output (e.g. `semantic`, `scope`, etc...)  

This is definitely an improvement over the status quo, and I had also been thinking about something like this myself; glad you took it on instead 👍

marler8997:
Maybe the final design will be a combination of the two. I have a feeling maintainers will be picky over this...so I'm going to be as flexible as possible here so we can make everyone's lives easier

wilzbach:
FWIW I also like this approach over #8156 and I don't like the current status quo either, so thanks a lot for taking this up and trying to push it!
And for the record I also agree that GitHub is the place where something like this should be discussed, because no one looks at Bugzilla issues and the NG will be off-topic with the first page. And even if there's some consensus like for the recent Markdown thread, people who don't read the NG will still object here.

WalterBright:
I'm sorry I just cannot see the value in this PR or the other one. I've used these things before and eventually move away from them. The compiler's logic can be fairly complex, and "turn on logging" usually results in a hopeless blizzard of haystacks with the pin lost in it.

Logging usually need to be customized to the particular problem that needs finding. A general logging scheme just is not helpful.

Some of this is also reinventing the `debug` conditional. For example, code in a `debug` version is not checked for purity. (Perhaps that can be extended to also allow system code in a safe function.)

PR #8158, opened 2018-04-10
appveyor: build Win32 dmd executable with LDC

rainers:
Unfortunately https://github.com/dlang/dmd/pull/7339 wasn't good enough for Win32.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -8,10 +8,12 @@ environment:
       D_VERSION:           2.077.1
       C_COMPILER:          MSVC
       VISUALD_VER:         v0.45.1-rc2
+      LDC_VERSION:         1.8.0
 
 cache:
   - C:\projects\gnumake\make.exe
   - C:\projects\VisualD-v0.45.1-rc2.exe
+  - C:\projects\ldc2-1.8.0-windows-multilib.7z
 
 skip_commits:
   # Avoid retesting the merged PR on `master` or `stable`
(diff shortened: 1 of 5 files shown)


WalterBright:
`longdouble.c` should be rewritten in betterC

WalterBright (approved):

rainers:
There is one failing test:
```
Test fail_compilation\ice1358.d failed.  The logged output:
..\generated\Windows\Release\Win32\dmd.exe -conf= -m32mscoff -Ifail_compilation -verrors=0   -L/OPT:NOICF  -odtest_m32mscoff\fail_compilation -oftest_m32mscoff\fail_compilation\ice1358_0.obj  -c fail_compilation\ice1358.d  -map nul.map
fail_compilation\ice1358.d(29): Error: invalid UTF character \U00d30762
Test fail_compilation\ice1358.d failed: 
expected:
----
fail_compilation\ice1358.d(29): Error: invalid UTF character \U80000000
----
actual:
----
fail_compilation\ice1358.d(29): Error: invalid UTF character \U00d30762
```
It is caused by a wrong value being passed through Lexer.error to .verror. Seems to be an issue with va_args. @kinke I suspect you have seen this with LDC before?

kinke:
Hmm nope. That test passes for LDC (CI-compiled using 64-bit LDC v1.6) on both Win32 and Win64 and I don't recall seeing it fail.

rainers:
@kinke Just wait until you update to latest dmd sources (before the latest commit in this PR): the problem is that passing variadic arguments through an interface function doesn't work with LDC because the interface thunk **calls** the class member function instead of just **jump** to it avoiding a copy of the arguments on the stack.

kinke:
Oh I just fixed that for LDC: https://github.com/ldc-developers/ldc/pull/2630

WalterBright:
> but it involves quite some changes so I think it should be better made in a separate PR.

Ehhhkselllehnt!

PR #8160, opened 2018-04-10
Add RUN_OUTPUT to test suite

marler8997:


JinShil:
I'm looking for motivation/rationale.  Why are asserts not sufficient for verifying runnable tests?

marler8997:
Good question...

I implemented this to support a new test I was going to add that verifies module import order invariance...but I could just use asserts for that.

Hmmmm, maybe the feature is not very useful after all?  What do you think @wilzbach ?  It's pretty simple to support, but is it justified?  I suppose verifying the output of runnable tests is "extra verification" that things are working correctly.

marler8997:
Feature doesn't seem to be as useful as I thought. Closing.

PR #8161, opened 2018-04-11
Fix Issue 18753 - chunkBy compile error causes ICE

RazvanN7:
I haven't added a test since the bug report contains code from phobos and the ice is present when the code errors (wrongfully). The test case errors on phobos lines of code which may differ depending on the version of phobos used so I cannot safely add the test in the fail_compilation directory. Any ideas on how we can test this ice?

Later edit: Thanks @wilzbach . I added the test.

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -3985,7 +3985,11 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             default:
                 assert(0);
             }
-            goto Lyes;
+
+            // https://issues.dlang.org/show_bug.cgi?id=18753
+            if (tded)
+                goto Lyes;
+            goto Lno;
         }
         else if (e.tspec && !e.id && !(e.parameters && e.parameters.dim))
         {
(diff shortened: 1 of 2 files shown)


wilzbach:
>  The test case errors on phobos lines of code which may differ depending on the version of phobos used so I cannot safely add the test in the fail_compilation directory. Any ideas on how we can test this ice?

Dustmite + reduce it. I just did that for you.
It's really easy though:

1) copy the Phobos module to a separate directory (it can also take the entire phobos dir, but that takes longer to reduce)
2) Add the failing example
3) Run dustmite

```
dustmite --no-redirect . "gdb -batch -ex run -ex bt -args ~/dlang/bin/dmddev -c -o- iteration.d 2>&1 | grep 'declaration.d'"
```

4) Depending on what you want to reduce, it can be a lot faster to just copy the imported Phobos modules over than to throw entire Phobos at DustMite - in your case I did that (i.e. I copied std.range.primitives + std.traits over). I do this as additional step, because sometimes dustmite can reduce the imports too.

Here's the result:

```d
struct ChunkByImpl{
    struct Group
    {
    }
    static assert(isForwardRange!Group);

}

enum isInputRange(R) =
ReturnType;


enum isForwardRange(R) = isInputRange!R
is(ReturnType!(() => r) );


template ReturnType(func...)
{
    static if (is(FunctionTypeOf!func R == return))
        ReturnType R;
}


template FunctionTypeOf(func...)
{
    static if (is(typeof(func[0]) T))
        static if (is(T Fptr ) )
        alias FunctionTypeOf = Fptr;
}


template Select()
{
}
```

JinShil (approved):

PR #8162, opened 2018-04-11
Use D to run the testsuite

wilzbach:
So there have been many complaints about how hard running the test suite on Windows is.
Here's an excerpt:

- https://forum.dlang.org/post/tulbhulbeqqzofdxevcg@forum.dlang.org
- https://forum.dlang.org/post/gxxfmrnezfrlodlhpiwe@forum.dlang.org
- https://forum.dlang.org/post/uyzscudefkhqyjdzvczc@forum.dlang.org
- https://issues.dlang.org/show_bug.cgi?id=18413

This PR allows to run the testsuite _without_ GNUMake by making the `run_individual_tests.d` wrapper no longer dependent on the Makefile.
It's not perfect yet and I still have to test a few things, i.e.
- [x] can allow environment variables that can be overwritten in the Makefile, overwritten here too (e.g. `results_dir`)
- [x] use `$HOST_DMD` instead of `dmd` for building the tools
- [x] ensure that the correct `$DMD` and `$MODEL` is passed in everywhere
- [x] normalize the file names, s.t. this script can be run from anywhere
- [x] make sure it actually runs on Windows

but apart from that I think this already shows the general direction and it can already execute all tests, e.g.

```sh
./run_individual_tests.d fail_compilation
./run_individual_tests.d clean
./run_individual_tests.d all
./run_individual_tests.d fail_compilation -j10
./run_individual_tests.d fail_compilation/fail11562.d
./run_individual_tests.d fail_compilation AUTO_UPDATE=1
```

(I opened this already, s.t. no duplicate work on this happens)


test/run.d
```diff
@@ -0,0 +1,308 @@
...
+        scope(exit) taskPool.finish();
+        ensureToolsExists;
+        foreach (target; taskPool.parallel(targets, 1))
+        {
+            auto makeArguments = [scriptDir.buildPath("test_results", "d_do_test"), target];
+            spawnProcess(makeArguments, env, Config.none, scriptDir).wait;
+        }
+    }
+}
+
```
marler8997:
```D
immutable testDirs = ["runnable", "compilable", "fail_compilation"];
```
Can be used in a few places.

test/run.d
```diff
@@ -0,0 +1,308 @@
...
+        else
+        {
+            auto command = ["dmd", "-of"~targetBin, sourceFile];
+            writefln("Executing: %-(%s %)", command);
+            spawnProcess(command).wait;
+        }
+    }
+
+    // ensure output directories exist
+    foreach (dir; ["compilable", "fail_compilation", "runnable"])
```
marler8997:
```D
foreach (dir; testDirs)
```

test/run.d
```diff
@@ -0,0 +1,308 @@
...
+    {
+        return scriptDir.buildPath(dir).dirEntries("*{.d,.sh}", SpanMode.shallow).map!(e => e.name);
+    }
+
+    Appender!(string[]) newTargets;
+    foreach (t; targets)
+    {
+        switch (t)
+        {
+            case "clean":
```
marler8997: I had a problem with this, my tab-autocomplete added a trailing slash to the test directory (i.e. `compilable/`).  We could handle this by removing the trailing slash, maybe this?
```D
if (t.endsWith(dirSeparator))
    t = t[0 .. $ - 1];
```

test/run.d
```diff
@@ -0,0 +1,308 @@
...
+            case "run_fail_compilation_tests", "fail_compilation", "fail":
+                newTargets.put(findFiles("fail_compilation"));
+                break;
+
+            case "run_compilable_tests", "compilable", "compile":
+                newTargets.put(findFiles("compilable"));
+                break;
+
+            case "all":
+                newTargets.put(findFiles("runnable"));
```
marler8997:
```D
foreach (testDir; testDirs)
    newTargets.put(findFinds(testDir));
```

test/run.d
```diff
@@ -0,0 +1,308 @@
...
+                newTargets.put(findFiles("fail_compilation"));
+                newTargets.put(findFiles("compilable"));
+                break;
+
+            default:
+                newTargets ~= t;
```
marler8997: This results in an odd error in some cases.  Maybe the following check would work well?
```D
if (!testDirs.canFind(t.dirName))
{
    writefln("Error: invalid target '%s'", t);
    exit(1);
}
```

test/run.d
```diff
@@ -0,0 +1,308 @@
...
+        return "openbsd";
+    else version(NetBSD)
+        return "netbsd";
+    else version(DragonFlyBSD)
+        return "dragonflybsd";
+    else version(Solaris)
+        return "solaris";
+    else version(SunOS)
+        return "solaris";
+    else
```
marler8997: No windows!??!! :)

I went ahead and tested that this runs on windows and it looks like it does.  Just had to add the `version(Windows)` case here.

marler8997:
Could we add a "v|verbose" flag to see things like the `d_do_test` commands that are being run?

marler8997:
Another thought...now that we have full control of the test suite, maybe it would be helpful to print a summary of the results at the end of the test:
```
----------------------------------------------
100 Passed 20 Failed
```

test/run.d
wilzbach: We already check for file existence in `filterTargets`:

```
./run.d compilable/foo
Warning: compilable/foo can't be found
```

What's the odd error message you get?

wilzbach:
> Could we add a "v|verbose" flag to see things like the d_do_test commands that are being run?

Done. Though I only added it in a very naive way for now. We can always expand on this later.

> Another thought...now that we have full control of the test suite, maybe it would be helpful to print a summary of the results at the end of the test. Something like:

Good idea, but an overkill for this PR. Let's focus on getting this to a mergeable state and let's expand on it afterwards?
It's already a rather large PR.

test/run.d
wilzbach: (windows is on the top of this list now)

wilzbach (commented):
Okay I finally found time to work more on this and it should be ready now to replace the `run_individual_tests.d` script.
This improvement is a huge superset of the functionality of this test runner that we added a few weeks ago and hasn't been integrated in any CI process (except for a single test at CircleCi).
The long run goal is to replace the Makefile with `run.d`, but for now `run.d` should already allow running the tests on Windows and remove the dependence on GNUmake.

tl;dr: nothing depends on it + we can slowly improve `run.d` if there are still some slight differences between it and the Makefile (these probably won't be caught in the review anyhow).

test/run.d
```diff
@@ -0,0 +1,343 @@
...
+    // allow overwrites from the environment
+    resultsDir = environment.get("RESULTS_DIR", resultsDir);
+    hostDMD = environment.get("HOST_DMD", "dmd");
+
+    // bootstrap all needed environment variables
+    auto env = getEnvironment;
+
+    // default target
+    if (!args.length)
+        args = ["all"];
```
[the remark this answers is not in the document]
RazvanN7: I think that running all the tests by default is a more intuitive approach

RazvanN7:
Looks like you forgot to delete the check for the run individual test. Delete it and we are ready to go.

RazvanN7 (approved):

RazvanN7:
Do we still need the Makefile after this? Maybe we should consider deleting it.

wilzbach:
> Do we still need the Makefile after this? Maybe we should consider deleting it.

No we don't :)
But currently a lot of CIs still depend on it, so the idea is slowly to migrate away from it.
Also, this new `run.d` test runner probably still needs a bit of real life testing and feedback.

> Looks like you forgot to delete the check for the run individual test. Delete it and we are ready to go. (this will solve the circleci failure)

The failure was due to change of the file name. Changing it made the CI happy again and I think it doesn't hurt to keep testing this in the CI. After all, eventually we want all CIs to use it.

wilzbach:
@marler8997 @JinShil are you two also okay with finally moving forward with this?
We can always work out the subtleties in follow-ups.

marler8997 (approved):

test/run.d
```diff
@@ -0,0 +1,342 @@
...
+
+    alias normalizeTestName = f => format!"%s/%s"(f.absolutePath.dirName.baseName, f.baseName);
```
JinShil: `std.path.pathSeparator` instead of `/`?

test/tools/d_do_test.d
```diff
@@ -895,11 +897,11 @@ int runBashTest(string input_dir, string test_name)
     version(Windows)
     {
         auto process = spawnShell(format("bash %s %s %s",
-            buildPath("tools", "sh_do_test.sh"), input_dir, test_name));
+            buildPath(scriptDir, "tools", "sh_do_test.sh"), input_dir, test_name));
     }
     else
     {
-        auto process = spawnProcess(["./tools/sh_do_test.sh", input_dir, test_name]);
+        auto process = spawnProcess([scriptDir ~ "/tools/sh_do_test.sh", input_dir, test_name]);
```
JinShil: The `/` will probably work fine in the posix/Bash environment.  I don't know if it will or won't cause problems on Windows if we ever get to a point where we can run this outside of that environment.  I suggest using `std.path.pathSeparator`.

test/run.d
```diff
@@ -0,0 +1,342 @@
...
+    {
+        t = t.buildNormalizedPath; // remove trailing slashes
+        switch (t)
+        {
+            case "clean":
+                resultsDir.rmdirRecurse;
+                exit(0);
+                break;
+
+            case "run_runnable_tests", "runnable":
```
JinShil: The "run_" prefix seems redundant, but I'm just being picky.

JinShil (approved):
LGTM, but I'm wondering if the use of `/` as a delimiter will cause problems on Windows if run outside of a Posix/Bash environment.  I'm also not sure how DMD handles `/` and `\` when used as a separator in its arguments on Windows

test/run.d
```diff
@@ -0,0 +1,342 @@
+#!/usr/bin/env rdmd
+/**
+Allows running tests individually
```
JinShil: Probably not an entirely accurate summary anymore.  It may be prudent to elaborate.

test/tools/d_do_test.d
wilzbach: I went with `buildPath`

test/run.d
wilzbach: I went with buildPath

test/run.d
wilzbach: `run_runnable_tests` is the name of the existing Makefile target and at least for the transition I wanted to keep the most commonly used Makefile targets

PR #8163, opened 2018-04-11
[D1] Run sema3 for all modules when generating deps

stefan-koch-sociomantic:
This makes sure we don't miss depdenecies introduces imports functions or unittests

### DIFF
--- a/src/mars.c
+++ b/src/mars.c
@@ -1531,6 +1531,20 @@ int main(int iargc, const char *argv[])
     {
         assert(global.params.moduleDepsFile != NULL);
 
+        // when generating depdencies we have to run sema3
+        // because otherwise we will miss some depdencies
+        // introdcued in function bodies
+        for (size_t i = 0; i < Module::amodules.dim; i++)
+        {
+            m = Module::amodules[i];
+            if (global.params.verbose)
+                printf("[deps] semantic(2/3) %s\n", m->toChars());
+            // not sure if it can happen that we havn't run sema2 ?
+            // better safe then sorry, it does not cost more then a few cycles
+            m->semantic2();
+            m->semantic3();
+        }
+
         File deps(global.params.moduleDepsFile);
         OutBuffer* ob = global.params.moduleDeps;
         deps.setbuffer((void*)ob->data, ob->offset);


stefan-koch-sociomantic:
backport of: https://github.com/dlang/dmd/pull/6748

src/mars.c
```diff
@@ -1531,6 +1531,20 @@ int main(int iargc, const char *argv[])
     {
         assert(global.params.moduleDepsFile != NULL);
 
+        // when generating depdencies we have to run sema3
+        // because otherwise we will miss some depdencies
+        // introdcued in function bodies
+        for (size_t i = 0; i < Module::amodules.dim; i++)
```
RazvanN7: Couldn't this start from 1 since the first module is the module being compiled and semantic has already been performed on it ?

RazvanN7 (approved):
Looks good.

src/mars.c
stefan-koch-sociomantic: It could but who cares. I'd rather give no guarantees about order.
That's how bugs are born :)

jacob-carlborg: Please make it a `foreach` loop.

jacob-carlborg: Never mind.

jacob-carlborg (approved):

WalterBright:
I object to this because:

* it is not fixing any known bug
* it will substantially slow down the compiler
* the compiler has moved towards lazily doing all sorts of analyses - for speed purposes
* if there are specific problems found where semantic3() isn't being run, that's where it should be run. There are many such places already in the compiler

src/mars.c
```diff
@@ -1531,6 +1531,20 @@ int main(int iargc, const char *argv[])
     {
         assert(global.params.moduleDepsFile != NULL);
 
+        // when generating depdencies we have to run sema3
```
WalterBright: please don't abbreviate `semantic3` - such will make grep useless

also, `dependencies`

src/mars.c
```diff
@@ -1531,6 +1531,20 @@ int main(int iargc, const char *argv[])
...
+        // because otherwise we will miss some depdencies
```
WalterBright: dependencies

src/mars.c
```diff
@@ -1531,6 +1531,20 @@ int main(int iargc, const char *argv[])
...
+        // introdcued in function bodies
```
WalterBright: introduced

src/mars.c
```diff
@@ -1531,6 +1531,20 @@ int main(int iargc, const char *argv[])
...
+        for (size_t i = 0; i < Module::amodules.dim; i++)
+        {
+            m = Module::amodules[i];
+            if (global.params.verbose)
+                printf("[deps] semantic(2/3) %s\n", m->toChars());
+            // not sure if it can happen that we havn't run sema2 ?
+            // better safe then sorry, it does not cost more then a few cycles
```
WalterBright: than

Need a better rationale than "not sure" and "better safe than sorry"

mihails-strasuns-sociomantic:
@WalterBright this only targets dmd-1.x branch and is intended to fix problem already fixed in main D2 branch. Is that still a no-go?

WalterBright:
Sorry, I missed that it was D1 only. It's good to go!

ibuclaw:
Shouldn't the spelling fixes been done prior to merge? We should still try to keep a reasonable code quality regardless of the branch.

mihails-strasuns-sociomantic:
Yes, I was intending to merge this myself after doing necessary manual testing (there are no auto-tests for D1 branch to rely on) ;)

Will make separate fixup PR if I find something.

stefan-koch-sociomantic:
I am going to fix the spelling of the comments.
@WalterBright thanks for pointing them out

src/mars.c
```diff
@@ -1531,6 +1531,20 @@ int main(int iargc, const char *argv[])
...
 
+        // when generating depdencies we have to run sema3
+        // because otherwise we will miss some depdencies
+        // introdcued in function bodies
+        for (size_t i = 0; i < Module::amodules.dim; i++)
+        {
+            m = Module::amodules[i];
+            if (global.params.verbose)
+                printf("[deps] semantic(2/3) %s\n", m->toChars());
+            // not sure if it can happen that we havn't run sema2 ?
```
gautam-kotian-sociomantic: typo `havn't`

stefan-koch-sociomantic: fixed in the fixup #8166

PR #8164, opened 2018-04-11
Fix Issue 7443 - Better diagnostic on wrongly written static constructor

RazvanN7:
This should probably go through a deprecation cycle. Opinions?

### DIFF
--- a/changelog/deprecate_ctor_in_static_block.dd
+++ b/changelog/deprecate_ctor_in_static_block.dd
@@ -1,0 +1,54 @@
+Deprecate allowing a constructor declaration in a static block
+
+Before patch:
+The language specification states that a static constructor is
+defined using the construction "static this()". Defining a
+constructor inside a static block does not have any effect
+on the constructor. The following code samples do not affect
+the constructor in any way:
+
+---
+static:
+    this() {}
+---
+
+---
+static
+{
+    this() {}
+}
+---
+
+The compiler does not issue any warning/error on the above code
+samples and generates a normal constructor which is not ran before
+the main() function. This leads to situations in which the
+compiler is not able to correctly indicate the problem:
+
+---
+class A
+{
+    static
+    {
+        this() {}
+    }
+
+    this() {}
+}
+
+void main()
+{
+    new A();
+}
+---
+
+This code will result in an error message indicating that
+there is a multiple definition of the constructor this()
+which is a misleading message.
+
+After patch:
+Whenever a constructor is encountered in a static context a
+deprecation message is issued stating that the `static`
+keyword does not have any effect on the constructor.
+The solution is to declare the constructor outside the
+static block either as a normal constructor or a static
+one (`static this()`).
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -3564,6 +3564,18 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         }
 
         sc = sc.push();
+
+        if (sc.stc & STC.static_)
+        {
+            // Deprecated in 2018-04.
+            // Change to error in 2019-04.
+            // @@@DEPRECATED_2019-04@@@.
+            if (sc.stc & STC.shared_)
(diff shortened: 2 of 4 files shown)


jacob-carlborg (changes requested):
Yes, please make it a deprecation instead of an error. Especially since it was necessary to change a test.

RazvanN7:
@jacob-carlborg Thanks for the answer. Done.

jacob-carlborg (changes requested):
A changelog is required as well, describing the problem and what to do to solve it.

RazvanN7:
@jacob-carlborg Done.

changelog/deprecate_ctor_in_static_block.dd
```diff
@@ -0,0 +1,54 @@
...
+    samples and generates a normal constructor which is not ran before
+    the main() function. This leads to situations in which the
+    compiler is not able to correctly indicate the problem:
+
+---
+class A
+{
+    static
+    {
+        this()
```
jacob-carlborg: Missing semicolon?

changelog/deprecate_ctor_in_static_block.dd
```diff
@@ -0,0 +1,54 @@
...
+    there is a multiple definition of the constructor this()
+    which is a misleading message.
+
+After patch:
+    Whenever a constructor is encountered in a static context a
+    deprecation message is issued stating that the `static`
+    keyword does not have any effect on the constructor.
+    The solution is to declare the constructor outside the
+    static block either as a normal constructor or a static
+    one (`static this()`).
```
jacob-carlborg: Why the indentation for many of the paragraphs?

changelog/deprecate_ctor_in_static_block.dd
RazvanN7: Yes, added.

changelog/deprecate_ctor_in_static_block.dd
RazvanN7: Fixed the indentation

jacob-carlborg (approved):

jacob-carlborg:
Same question here, do we need a spec update?

RazvanN7:
I don't think so : https://dlang.org/spec/class.html#static-constructor . The spec does not explicitly say it is an error, just that it's not a static constructor. At most, the example in the spec could be updated to pinpoint that the code will issue an error once the deprecation is over.

jacob-carlborg:
Ok, so the spec was missing this. Or rather there was a bug in the compiler and shouldn't have worked in the first place.

jacob-carlborg:
LGTM :ok_hand:. Going to leave this open a bit longer for others to get a chance to look as well.

JinShil:
From the spec:

> "The static in the static constructor declaration is not an attribute, it must appear immediately before the this"

Why does a static constructor need to be given extra special treatment?  Can't the compiler be fixed to recognize that the constructor is in a static block and define it as a static constructor instead of emitting an error?

RazvanN7:
@JinShil
 
>Why does a static constructor need to be given extra special treatment? Can't the compiler be fixed to recognize that the constructor is in a static block and define it as a static constructor instead of emitting an error?

Although I might admit that this is not the best solution, I think that these are the 2 reasons why the static
constructor needs extra treatment : 

1. Implementation specific: at parse time there is no way to know whether the declaration is inside a static block so the parser cannot create a StaticCtorDeclaration AST node. The information is known at semantic time by checking the scope in which the constructor is declared. This can be changed by rewriting the AST and turning the constructor into a static constructor (with a bit of overhead, of course).

2. User specific : the meaning of `static` in a static constructor is slightly different from the meaning of static in a static method. While the first means "run this method once for initialization before running the `main` function" , the second has a different meaning. Issuing an error when a constructor is defined in a static block is a way to ensure that the user knows exactly the difference between the 2 and rules out cases in which a `static:` definition accidentally includes a constructor definition.

andralex (approved):

andralex:
@JinShil changing behavior from one working semantics to a different working semantics is the most perilous so we want to avoid that.

PR #8165, opened 2018-04-11
Add broken import order invariance tests

marler8997:
Added tests to the DMD test suite that will pass once https://issues.dlang.org/show_bug.cgi?id=18517 is fixed (PR here https://github.com/dlang/dmd/pull/7900).  Added the `BROKEN:` setting to these tests to indicate that they are currently broken.

The tests that were added currently fail because import order invariance is broken.  One test has an example where imports in a different order in the same file will produce different results, while another has an example where modules passed in a different order on the command line will produce different results.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -176,6 +176,8 @@ The following is a list of all available settings:
                          Optionally a MODEL suffix can used for further filtering, e.g.
                          win32 win64 linux32 linux64 osx32 osx64 freebsd32 freebsd64
 
+    BROKEN:              the test is broken. failure won't cause test suite to fail.
+
 Makefile Environment variables
 ------------------------------
 
--- a/test/fail_compilation/imports/import18517a.d
+++ b/test/fail_compilation/imports/import18517a.d
@@ -1,0 +1,1 @@
+module import18517b;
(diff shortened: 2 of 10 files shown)


marler8997:
Ping

RazvanN7:
Instead of adding `Broken:`, why not move these tests to : https://github.com/dlang/dmd/pull/7900 ?

marler8997:
#7900 is a pretty big change and I'm not sure if it's going to get integrated.  There are also other solutions to import order invariance besides what I've done in #7900 (i.e. I made another pull requests before #7900 that fixed the issue but that one was rejected).  In any case, I'd like to have the tests integrated so that when this does get fixed we will already have tests for them.  Note that a "BROKEN" tests will cause the test suite to fail if the test itself passes, so as soon as this issue gets fixed, the developer will already have tests for their fix.

RazvanN7:
I don't see much benefit in having tests that don't pass in the test suite. This puts a burden on the autotester and encourages people to  make PRs just with tests creating a separation between the fix and the test. I think that any of the following solutions are better:

-> close this PR and link it on the bug report page so that anyone who will fix it may use the tests
-> move the tests to https://github.com/dlang/dmd/pull/7900 and link the PR to the bug report so that
    anyone who wants to fix the issue will see this past attempt and the associated tests

PR #8167, opened 2018-04-12
Fix Issue 18719 - Doubly-called constructor against member when using forwarding constructors

RazvanN7:
Whenever a call expression to a constructor is encountered, update the scope ctorflow so that it is known that all member fields have been initialized

### DIFF
--- a/changelog/deprecate_double_initialization.dd
+++ b/changelog/deprecate_double_initialization.dd
@@ -1,0 +1,47 @@
+Deprecate double initialization of immutable fields inside constructor
+
+Inside a constructor scope, assigning to aggregate declaration (class/struct)
+members is done by considering the first assignment as initialization and
+subsequent assignments as modifications of the initially constructed object.
+For `const`/`immutable` fields the initialization is accepted in the constructor,
+but subsequent modifications are not. Example:
+
+---
+struct A
+{
+    int a;
+    immutable int b;
+    this(int a, int b)
+    {
+        this.a = a;
+        this.b = b;
+
+        this.a = 7; // OK, a is mutable
+        this.b = 9; // Error: immutable field b initialized multiple times
+    }
+}
+---
+
+However, $(BUGZILLA 18719) shows that this rule does not apply when inside
+a constructor scope there is a call to a different constructor:
+
+---
+struct A
+{
+    immmutable int a;
+    this()
+    {
+        this(42);
+        this.a = 5;  // second initialization of immutable field
+    }
+
+    this(int a)
+    {
+        this.a = a;
+    }
+}
+---
+
+The above code wrongfully compiled succesfully before this patch, accepting the double
+initialization of the `immutable` field `a`. After this patch, `this.a = 5` will issue
+a deprecation warning stating that `a` is initialized multiple times.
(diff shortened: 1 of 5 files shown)


src/dmd/expressionsem.d
```diff
@@ -3271,11 +3271,22 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
                 {
                     exp.error("constructor call must be in a constructor");
                     return setError();
                 }
+
+                // https://issues.dlang.org/show_bug.cgi?id=18719
+                // If `exp` is a call expression to another constructor
+                // then it means that all struct/class fields will be
+                // initialized after this call.
+                for (size_t i = 0; i < sc.ctorflow.fieldinit.length; i++)
```
UplinkCoder: foreach ?

src/dmd/expressionsem.d
```diff
@@ -3271,11 +3271,22 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             }
             else
             {
+                // `this` call expression must be inside a
+                // constructor
                 if (!ad || !sc.func.isCtorDeclaration())
                 {
                     exp.error("constructor call must be in a constructor");
```
UplinkCoder: does this ever happen ?

RazvanN7: it might if  one is noobish enough

RazvanN7:
@CyberShadow I transformed the error into a deprecation and added a changelog entry.

src/dmd/ctorflow.d
```diff
@@ -29,6 +29,7 @@ enum CSX : ubyte
     return_         = 0x20,     /// seen a return statement
     any_ctor        = 0x40,     /// either this() or super() was called
     halt            = 0x80,     /// assert(0)
+    deprecate_18719 = 0x100,    // issue deprecation for Issue 18719 - delete when deprecation period is over
```
jacob-carlborg: Can we call it `deprecate` so it can be used for other deprecations?

RazvanN7: It is highly unlikely that this is going to be used for something else. even so, the modification can be done when it is necessary

changelog/deprecate_double_initialization.dd
```diff
@@ -0,0 +1,48 @@
...
+        this.a = a;
+        this.b = b;
+
+        this.a = 7; \\ OK, a is mutable
+        this.b = 9; \\ Error: immutable field b initialized multiple times
+    }
+}
+---
+
+However, issue 18719 ( https://issues.dlang.org/show_bug.cgi?id=18719 ) shows
```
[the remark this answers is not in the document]
andralex: Better yet: `$(BUGZILLA 18719)`

andralex (approved):

src/dmd/ctorflow.d
jacob-carlborg: https://github.com/dlang/dmd/pull/8143#issuecomment-379533909

wilzbach:
This introduced a regression as it only checks whether a `this()` call is present, but not what the this call actually does. For example, it is totally valid to have a `this` constructor that does nothing..

https://issues.dlang.org/show_bug.cgi?id=19030

PR #8168, opened 2018-04-13
Preserve the `__debug_line` section on macOS in the final executable

jacob-carlborg:
The linker on macOS will remove any debug info, i.e. every section with the `S_ATTR_DEBUG` flag, this includes everything in the `__DWARF` section. By using the `S_REGULAR` flag the linker will not remove this section. This allows to get the filenames and line numbers for backtraces from the executable.

Normally the linker removes all the debug info but adds a reference to the object files. The debugger can then read the object files to get filename and line number information. It's also possible to use an additional tool that generates a separate `.dSYM` file. This file can then later be deployed with the application if debug info is needed when the application is deployed.

### DIFF
--- a/src/dmd/backend/dwarf.c
+++ b/src/dmd/backend/dwarf.c
@@ -11,6 +11,21 @@
 
 // Emit Dwarf symbolic debug info
 
+/*
+Some generic information for debug info on macOS:
+
+The linker on macOS will remove any debug info, i.e. every section with the
+`S_ATTR_DEBUG` flag, this includes everything in the `__DWARF` section. By using
+the `S_REGULAR` flag the linker will not remove this section. This allows to get
+the filenames and line numbers for backtraces from the executable.
+
+Normally the linker removes all the debug info but adds a reference to the
+object files. The debugger can then read the object files to get filename and
+line number information. It's also possible to use an additional tool that
+generates a separate `.dSYM` file. This file can then later be deployed with the
+application if debug info is needed when the application is deployed.
+*/
+
 #if !SPP
 #include        <stdio.h>
 #include        <string.h>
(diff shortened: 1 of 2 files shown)


jacob-carlborg:
I've not included a changelog entry here because it's not a user facing change.

WalterBright (approved):

WalterBright:
Please add your opening comment in as a comment in the code.

jacob-carlborg:
> Please add your opening comment in as a comment in the code.

It's already in the commit message, do you want it in the code as well?

jacob-carlborg:
Is the failing DAutoTest related to this change? Nothing else fails.

wilzbach:
Yeah the error is

```
---
void main() {
import std.stdio;
        struct S { int x; float y; }

        int foo(S s) { return s.x; }

        foo( S(1, 2) ); // set field x to 1, field y to 2
        
}
---
/usr/bin/ld: error in __stdin.o(.eh_frame); no .eh_frame_hdr table will be created.
/usr/bin/ld: __stdin.o: invalid string offset 1096055152 >= 528 for section `.strtab'
/usr/bin/ld: unknown operator '(' in complex symbol
/usr/bin/ld: final link failed: Invalid operation
```

But it's unrelated to this PR.

wilzbach:
> Is the failing DAutoTest related to this change? Nothing else fails.

-> https://github.com/dlang/dlang.org/pull/2345

jacob-carlborg:
Everything is green.

jacob-carlborg:
Ping, this is pretty trivial.

jacob-carlborg:
@wilzbach @JinShil ping

JinShil (approved):

WalterBright:
Suggestion to revert this:

https://www.digitalmars.com/d/archives/digitalmars/D/Not_Working_on_the_Mac_369782.html#N369787

jacob-carlborg:
@WalterBright 

There are a couple of options:

* You can choose to revert this. Then filenames and line numbers will not be available in stack traces of exceptions (unless LDC has upstreamed its changes in druntime. The changes require an external tool, shipped with Xcode, which might not be available when an application is deployed)
* Try to fix the alignment (don't know if that's possible)
* Revert this change but duplicate the debug info in a custom section. I don't know if the linker will strip custom sections with the `S_ATTR_DEBUG` attribute

WalterBright:
@jacob-carlborg  I'd prefer to fix the problem rather than revert this. Except nobody seems to know what to do about it.

WalterBright:
Can you try fixing the alignment and see if it works?

jacob-carlborg:
> Can you try fixing the alignment and see if it works?

Unfortunately that's out of my league. It would be more suitable for someone that actually knows the compiler.

PR #8169, opened 2018-04-14
reimplement longdouble in D

rainers:
This gets rid of the split implementation in C and D and asm.

Unfortunately, I had to introduce another name for the software implementation, because we now have to cover the case where D code is compiled with dmd using native reals, but needs to supply symbols for the backend that is compiled with VC.

The respective type name aliases are now:
VisualC backend compiler:  `targ_ldouble` typedefs `longdouble` typedefs `longdouble_soft`
other C backend compilers: `targ_ldouble` typedefs `longdouble` typedefs `long double`

dmd frontend: `real_t` aliases `longdouble` aliases `real`
LDC/win frontend: `real_t` aliases `longdouble` aliases `longdouble_soft`

The GDC frontend can replace longdouble_soft with a complete software emulation.


src/dmd/root/longdouble.d
```diff
@@ -6,29 +6,30 @@
  * https://github.com/dlang/dmd/blob/master/src/root/longdouble.d
  */
 
-// 80 bit floating point value implementation for LDC compiler targetting MSVC
+// 80-bit floating point value implementation if the C/D compiler do not support them natively
```
WalterBright: s/do/does/

src/dmd/root/longdouble.d
```diff
@@ -6,29 +6,30 @@
...
 extern(C++):
 nothrow:
+@nogc:
 
-align(2) struct longdouble
+align(2) struct longdouble_soft
 {
-nothrow:
+nothrow @nogc:
     ulong mantissa = 0xC000000000000001UL; // default to snan
```
WalterBright: I gave up on snan a while ago. Nothing supports it, and it just causes problems when people compare nan bits. Use qnan instead.

src/dmd/root/longdouble.d
```diff
@@ -45,18 +46,23 @@ nothrow:
...
+    longdouble_soft opNeg() const { return longdouble_soft(mantissa, exp_sign ^ 0x8000); }
 
-    bool opEquals(const longdouble rhs) const { return this.ld_cmpe(rhs); }
-    int  opCmp(longdouble rhs) const { return this.ld_cmp(rhs); }
-    longdouble opAdd(longdouble rhs) const { return this.ld_add(rhs); }
-    longdouble opSub(longdouble rhs) const { return this.ld_sub(rhs); }
-    longdouble opMul(longdouble rhs) const { return this.ld_mul(rhs); }
-    longdouble opDiv(longdouble rhs) const { return this.ld_div(rhs); }
-    longdouble opMod(longdouble rhs) const { return this.ld_mod(rhs); }
+    bool opEquals(const longdouble_soft rhs) const { return this.ld_cmpe(rhs); }
```
WalterBright: hmm, on this `rhs` is `const`, but not so for the other functions.

src/dmd/root/longdouble.d
```diff
@@ -6,29 +6,30 @@
...
+else
+    alias longdouble = longdouble_soft;
 
-static if (!hasNativeFloat80):
+// longdouble_soft needed when building the backend with
+// Visual C or the frontend with LDC on Windows
+version(CRuntime_Microsoft):
 extern(C++):
 nothrow:
+@nogc:
```
WalterBright: Investigate using `pure` and `@safe` as much as practical.

ibuclaw:
> The GDC frontend can replace longdouble_soft with a complete software emulation.

Or I just ignore it completely as neither ctfloat.d nor longdouble.d is shareable code. :-)

rainers:
> I used the "Paranoia" test suite to make the DMC soft implementation of floating point arithmetic correct. It lives up to its name, you have to get it right for Paranoia to pass!

The Paranoia test only seems to fail for the overflow detection (using SIGFPE signals), but I guess we don't need that.

src/dmd/root/longdouble.d
rainers: Ok, updated.

src/dmd/root/longdouble.d
rainers: Added `pure` and `@safe`, but it has `pure @trusted` has to be verified manually on all the asm.

src/dmd/root/longdouble.d
rainers: Removed `const` for consistency as the argumets are passed by value anyway. I abuse the non-constness in some operations to avoid RVO ignoring the asm modifications.

WalterBright: > RVO ignoring the asm modifications.

Please file a bug report about that.

WalterBright:
> The Paranoia test only seems to fail for the overflow detection (using SIGFPE signals), but I guess we don't need that.

It's great news that Paranoia is working! Perhaps add it to the test suite? Maintenance on longdouble may accidentally break it. Did you translate it to D?

I know we don't need the signals, but is there a way to get it to pass anyway?

ibuclaw:
>> The Paranoia test only seems to fail for the overflow detection (using SIGFPE signals), but I guess we don't need that.

> It's great news that Paranoia is working! Perhaps add it to the test suite?

Are you testing compile time or runtime here?

rainers:
> Are you testing compile time or runtime here?

paranoia tests the C++ interface linked with the D implementation of longdouble, so I'd consider it a test of the compile time evaluation of dmd, but not some D operator overloads.

rainers:
Unfortunately the tests fail because LDC translates all types of FPU instructions to 64-bit, so
```
    fld real ptr [EAX];
    fld double ptr [EAX];
    fld extended ptr [EAX];
```
are all the same.
@kinke I see some unexpected translation for `extended` to `double` in asm-x86.h for MSVC. I'd rather expect it to be taken literally, but `double` to be used for `real` types. Or is this an LLVM limitation?

ibuclaw:
> paranoia tests the C++ interface linked with the D implementation of longdouble, so I'd consider it a test of the compile time evaluation of dmd, but not some D operator overloads.

So it could be part of the c++ test source then.  Or it could be converted into a ctfe test for the testsuite.

In any case, @WalterBright no chance of supporting all the paranoia tests because there is no fpu at compile time.

src/dmd/root/longdouble.d
rainers: > Please file a bug report about that.

https://issues.dlang.org/show_bug.cgi?id=18758

WalterBright:
> In any case, @WalterBright no chance of supporting all the paranoia tests because there is no fpu at compile time.

The point of paranoia is to test the "no fpu" soft float implementation. I'm not understanding your comment.

ibuclaw:
Why would it raise a signal if floating point is emulated?

kinke:
> I see some unexpected translation for extended to double in asm-x86.h for MSVC. I'd rather expect it to be taken literally, but double to be used for real types. Or is this an LLVM limitation?

That's a bug I introduced a while back, thanks for reporting => https://github.com/ldc-developers/ldc/pull/2653.

rainers:
>>I know we don't need the signals, but is there a way to get it to pass anyway?

> Not sure. Maybe it just needs enabling approriate FPU exceptions...

I was wrong about the actual problem due to the rather confusing output: the actual "DEFECT" was an inaccuracy of the pow function, so pretty expected because we are just using the double version.

I then noticed that the FPU precision hasn't been changed at all because I missed to call initFPU() :-/

With that I get 1 "failure" and 3 "defects". Quite ok when compared to dmc (1 failure, 1 serious defect, 6 defects) and gcc (2 failures, 1 serious defect, 6 defects). The latter is rather old, but version 4.9.1 is the latest I found on my system.

rainers:
> That's a bug I introduced a while back, thanks for reporting => ldc-developers/ldc#2653.

Thanks for the quick action. While trying to workaround I noticed
- `.bytes` or `db` not available to emit the opcodes myself. Would have been nice...
- EIP not available as a register (RIP is ok). Is this deliberate?

kinke:
> EIP not available as a register (RIP is ok).

Thx => https://github.com/ldc-developers/ldc/pull/2654

> .bytes or db not available to emit the opcodes myself. Would have been nice...

I'm not too familiar with the DMD-style inline asm parser code, so I don't know if that would be possible.

kinke:
[In LLVM asm, loading an 80-bit value would look something like this: `import ldc.llvmasm;  __asm("fldt $0", "*m,~{st}", ptr);`.]

WalterBright:
> Why would it raise a signal if floating point is emulated?

Because an emulator should emulate the real thing.

rainers:
> [In LLVM asm, loading an 80-bit value would look something like this: import ldc.llvmasm; __asm("fldt $0", "*m,~{st}", ptr);.]

I couldn't get this to work (keeps spoiling the values as arguments are passed by reference), but it could also be my ignorance of LLVM assembly. As the next LDC version will support sane syntax, I'll keep the byte emission hack for now.

JinShil:
What's the status with this?  @rainers, are you still working on this, or is it ready to go in your opinion?  Are there any outstanding issues?  What are they?

wilzbach:
> A possible complication is that the backend now depends on D code which might make linking a bit more troublesome on non-Windows-platforms (though not used there for now).

We are about to start converting the backend to D soon hopefully anyhow, so this shouldn't be a big problem.

src/dmd/root/longdouble.d
```diff
@@ -6,29 +6,92 @@
...
+else
+    static assert(false, "longdouble_soft not supported on this platform");
+
+bool initFPU()
 {
-nothrow:
+    version(D_InlineAsm_X86_64)
+    {
+        // set precision to 64-bit mantissa and rounding control to nearest
+        asm nothrow @nogc pure
```
JinShil: `nothrow` and `@nogc` are already declared above.  Is this redundant?  If so, I suggest choosing a coding style:  either add the attributes explicitly at each declaration, or group them under one attribution, but don't mix the two styles.

JinShil: Not confident enough for `@trusted`?

src/dmd/root/longdouble.d
```diff
@@ -89,35 +184,658 @@ nothrow:
...
+        return "asm nothrow @nogc pure @trusted { mov EAX, " ~ arg ~ "; " ~ fld_eax!arg ~ " }";
+    }
+    string fstp_parg(string arg)()
+    {
+        return "asm nothrow @nogc pure @trusted { mov EAX, " ~ arg ~ "; " ~ fstp_eax ~ " }";
+    }
+    alias fld_local = fld_arg;
+}
+
+double ld_read(const longdouble_soft* pthis)
```
JinShil: Can these be attributed with `@safe`?

JinShil (approved):
It's quite beyond me to verify the semantics of this code, but the craftsmanship and translation looks good.  Nice work!

src/dmd/root/longdouble.d
```diff
@@ -6,29 +6,92 @@
...
+            or      EAX, 0x300;
+            mov     dword ptr [RSP],EAX;
+            fldcw   word ptr [RSP];
+            pop     RAX;
+        }
+    }
+    else version(D_InlineAsm_X86)
+    {
+        // set precision to 64-bit mantissa and rounding control to nearest
+        asm nothrow @nogc
```
JinShil: Not `pure` like the one above?

JinShil:
@adamdruppe I understand you're quite good with Intel ASM.  Care to lend your expertise here, perhaps even verifying safety and purity?

src/dmd/root/longdouble.d
rainers: > nothrow and @nogc are already declared above. Is this redundant?

These attributes have to be repeated for each asm statement. otherwise the compiler errors out.

src/dmd/root/longdouble.d
rainers: `pure` is wrong above, the FPU control word is changed. I'll remove that.

src/dmd/root/longdouble.d
JinShil: Seems like a bug.

rainers: Before dmd 2.067, you couldn't write asm in pure/safe code: https://dlang.org/changelog/2.067.0.html#asm-attributes

src/dmd/root/longdouble.d
```diff
@@ -6,29 +6,92 @@
...
+shared static this()
+{
+    initFPU();
+}
+
+void ld_clearfpu()
+{
+    version(AsmX86)
+    {
+        asm nothrow @nogc pure @trusted
```
adamdruppe: I probably wouldn't call this pure since it explicitly is clearing flags; if this function were optimized out it would probably be wrong most times.

adamdruppe:
So I'm not in love with making `mov` instruction wrappers `pure`... in isolation, it seems silly, but in context it seems to work. Maaaybe slap `private` on it just to make it clear this is a bit of internal cheating/convenience, though tbh I am a bit meh on it and wouldn't hold up over this little feeling.

Resetting *flag* registers while calling it pure rubs me more the wrong way though. I guess it would pass the test if they are pushed and popped inside a pure function, but I don't think they consistently are.

For the `@trusted` parts.... so it looks right, but I gotta warn my fpu asm experience is fairly limited so I might be missing something too.... just it looks right eyeballing it.

rainers:
Thanks @adamdruppe for review. Indeed, ld_clearfpu() is not pure, so the asm inside should also not be marked that way. I've also added `private` to the asm mixins.

BTW: there is little new about the asm, it was already there with ldfpu.asm (for Win64) and longdouble.d (for Win32), but sure, merging them might have introduced regressions. Appveyor tests both versions, though.

dnadlinger (approved):

RazvanN7 (approved):

RazvanN7:
@WalterBright I think we are ready to merge to this. Do you have anything to add?

PR #5501, opened 2016-03-04
Add `bool gototarget` to `CaseStatement` and `DefaultStatement`

JohanEngelen:
(First commit is a small cleanup in SwitchStatement.semantic.)

Add `bool gototarget` to `CaseStatement` and `DefaultStatement` that is set to true when the switch contains a `goto case` targetting that `CaseStatement` (or `DefaultStatement`)
`gototarget` is used in LDC for PGO.
The variable is not currently used in DMD, but may be useful in the future, e.g. for control flow analysis.

Note that I also had to create SwitchStatement.hasGotoDefault variable. I did not find a satisfactory way to avoid creating that variable.

PR #8171, opened 2018-04-15
improve try-finally handling of nofallthru

WalterBright:
If the try-body does not fall-through to the finally-body, we can improve the code generation by marking that block as an exit node (treated as a does-not-get-here node). Saves a few instructions in generated code.

### DIFF
--- a/src/dmd/statement.d
+++ b/src/dmd/statement.d
@@ -2027,11 +2027,14 @@ extern (C++) final class TryFinallyStatement : Statement
     Statement _body;
     Statement finalbody;
 
+    bool bodyFallsThru;         // true if _body falls through to finally
+
     extern (D) this(const ref Loc loc, Statement _body, Statement finalbody)
     {
         super(loc);
         this._body = _body;
         this.finalbody = finalbody;
+        this.bodyFallsThru = true;      // assume true until statementSemantic()
     }
 
     static TryFinallyStatement create(Loc loc, Statement _body, Statement finalbody)
--- a/src/dmd/s2ir.d
+++ b/src/dmd/s2ir.d
@@ -1381,8 +1381,31 @@ private extern (C++) class S2irVisitor : Visitor
              *  BC_ret
              *  breakblock
              */
-            blx.curblock.appendSucc(breakblock);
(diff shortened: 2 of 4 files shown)


WalterBright:
Important to pull this because it makes my keynote presentation look better :-)

WalterBright:
On windows 32 and 64:
```
core.exception.AssertError@std\regex\package.d(1237): unittest failure
```
Doesn't seem to have anything to do with this change?

RazvanN7 (approved):

braddr:
Removing auto-merge tag to de-prioritize the build.  The windows platforms are failing consistently.  There isn't a systemic windows platform failing on any other builds, which points strongly to this pull being broken in some way still.

src/dmd/statement.d
```diff
@@ -2027,6 +2027,8 @@ extern (C++) final class TryFinallyStatement : Statement
     Statement _body;
     Statement finalbody;
 
+    bool bodyFallsThru;         // true if _body falls through to finally
```
klickverbot: Note that this is very similar in spirit to https://github.com/dlang/dmd/pull/5501, which was promptly shot down back then without offering an alternative solution.

Both add extra fields for metadata that can totally be inferred from the after-semantic AST, but is more convenient to collect during semantic analysis, and that is used by one particular backend implementation.

@ibuclaw, time to get on your soapbox again and settle this debate once and for all?

WalterBright:
> There isn't a systemic windows platform failing on any other builds, which points strongly to this pull being broken in some way still.

There was at the time, but now it is clear that there is something wrong with this PR.

src/dmd/statement.d
ibuclaw: Indeed this field does look pretty pointless.

ibuclaw: There should ideally be a region that's reserved for per compiler specific information. For arguments sake, let's give it a void pointer type in the frontend. Here is stored information that can be read or set by the backend before, in-between, or after the semantic passes. In the case of LDC, I'd be surprised if you couldn't add your own PGO pass to achieve the same result - just at the slight cost of compile time.

PR #8172, opened 2018-04-15
Fix issue 15574 - wrong order of link arguments

rtbo:
This essentially places *.a arguments before the -l and -L switches in
the linker command line.

This allows to link applications that link against a static archive *.a
that depends on -l lib switches.
Such dependency scheme is typical of the following situation:
  - have a D package that contains binding to a C library.
  - this D package is itself built as a static library (libdpackage.a).
  - the C library is built in the same build process as a static library
    (pathtoClib/libclib.a)
  - dmd is invoked in the following way:
```
  $ dmd ... libdpackage.a -L-LpathtoClib -L-lclib
```
Currently dmd reorders arguments and places *.a after -l linker switches
which causes link errors if the -l points to a static library.

This is especially useful with Dub, which can be configured to probe for
the C library, and build it as a static lib if not found system-wide.
In both cases, the C library is provided with "libs" json entry.

So this is not about keeping relative order of linker arguments, it is
about providing an order of linker arguments that is more sound than
the current one.

### DIFF
--- a/test/compilable/issue15574.sh
+++ b/test/compilable/issue15574.sh
@@ -1,0 +1,58 @@
+#! /usr/bin/env bash
+
+
+if [[ $OS = *"win"* ]]; then exit 0; fi
+
+
+if [ $LIBEXT = ".a" ]; then
+    LIB_PREFIX=lib
+else
+    LIB_PREFIX=
+fi
+
+TEST_DIR=${OUTPUT_BASE}
+C_FILE=$TEST_DIR/square.c
+C_LIB=$TEST_DIR/${LIB_PREFIX}csquare${LIBEXT}
+D_FILE=$TEST_DIR/square.d
+D_LIB=$TEST_DIR/${LIB_PREFIX}dsquare${LIBEXT}
+APP_FILE=$TEST_DIR/app.d
+APP=$TEST_DIR/app${EXE}
+
+mkdir -p $TEST_DIR
+
+cat >$C_FILE <<EOF
+int square(int x) { return x*x; }
+EOF
+
+cat >$D_FILE <<EOF
+module square;
+
+extern(C) nothrow {
+    int square (int x);
+}
+
+bool testSquare() {
+    return square(2) == 4 && square(5) == 25;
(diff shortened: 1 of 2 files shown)


wilzbach:
@rtbo have a look at the existing bash scripts in the testsuite for examples on how this can be added to testsuite.

RazvanN7 (approved):

RazvanN7:
The documentation failure is unrelated to this PR.

test/compilable/issue15574.sh
```diff
@@ -0,0 +1,56 @@
...
+EOF
+
+if [ -z ${CC+x} ]; then CC=cc; fi
+
+$CC -m${MODEL} -c -o ${C_FILE}${OBJ} $C_FILE
+ar rcs ${C_LIB} ${C_FILE}${OBJ}
+
+${DMD} -m${MODEL} -lib -of${D_LIB} ${D_FILE}
+
+${DMD} -m${MODEL} -of${APP} ${APP_FILE} -I${TEST_DIR} ${D_LIB} -L-L${TEST_DIR} -L-lcsquare
```
RazvanN7: You should also run the executable to make sure that everything is fine.

rtbo: Done. (and squashed)

test/compilable/issue15574.sh
```diff
@@ -0,0 +1,58 @@
...
+
+${APP}
```
RazvanN7: It would be best if you could check the application actually runs correctly i.e. it finished with exit code 0

rtbo: Doesn't the test fail if a command returns non-zero?

RazvanN7: you are correct.

RazvanN7:
@rtbo I added the `merge` label on this PR. Thank you for your contribution

CyberShadow:
@rtbo This pull request introduced a regression:
https://issues.dlang.org/show_bug.cgi?id=19243

PR #8173, opened 2018-04-16
 Partial fix for Issue 18472 - permit using typeid at compile-time in betterC

JinShil:
When compiling with `-betterC`, `TypeInfo` is not usable at runtime, but it should still be usable at compile-time.

### DIFF
--- a/src/dmd/typinf.d
+++ b/src/dmd/typinf.d
@@ -22,6 +22,7 @@ import dmd.globals;
 import dmd.gluelayer;
 import dmd.mtype;
 import dmd.visitor;
+import core.stdc.stdio;
 
 /****************************************************
  * Generates the `TypeInfo` object associated with `torig` if it
@@ -35,10 +36,16 @@ extern (C++) void genTypeInfo(Loc loc, Type torig, Scope* sc)
 {
     //printf("Type::genTypeInfo() %p, %s\n", this, toChars());
 
-    if (!global.params.useTypeInfo)
+    // Even when compiling without `useTypeInfo` (e.g. -betterC) we should
+    // still be able to evaluate `TypeInfo` at compile-time, just not at runtime.
+    // https://issues.dlang.org/show_bug.cgi?id=18472
+    if (!sc || !(sc.flags & SCOPE.ctfe))
     {
-        torig.error(loc, "`TypeInfo` cannot be used with -betterC");
-        fatal();
+        if (!global.params.useTypeInfo)
+        {
+            torig.error(loc, "`TypeInfo` cannot be used with -betterC");
+            fatal();
+        }
     }
 
     if (!Type.dtypeinfo)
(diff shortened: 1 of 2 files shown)


RazvanN7 (approved):

JinShil:
While this fixes the test case `enum b = typeid(size_t) is typeid(uint);`, unfortunately it still fails for...

```D
import std.format;
import std.stdio;
void main()
{
    enum s = "%1$s,%2$s".format("foo","bar");
    writeln(s);
}

// phobos/std/array.d(2850): Error: TypeInfo cannot be used with -betterC
```

src/dmd/typinf.d
```diff
@@ -35,10 +36,16 @@ extern (C++) void genTypeInfo(Loc loc, Type torig, Scope* sc)
 {
     //printf("Type::genTypeInfo() %p, %s\n", this, toChars());
 
-    if (!global.params.useTypeInfo)
+    // Even when compiling without `useTypeInfo` (a.k.a. `-betterC`) we should
```
WalterBright: Code should appear in backticks, not 'betterC' which is not code.

src/dmd/typinf.d
```diff
@@ -35,10 +36,16 @@ extern (C++) void genTypeInfo(Loc loc, Type torig, Scope* sc)
...
+    // still be able to evaluate `TypeInfo` at compile-time, just not at runtime.
+    // See https://issues.dlang.org/show_bug.cgi?id=18472
```
WalterBright: The 'See ' is redundant.

WalterBright (approved):
Please fix my suggestions, then I approve.

jacob-carlborg:
Does this fix the whole issue?

JinShil:
> Does this fix the whole issue?

No, it only fixes...

```D
enum b = typeid(size_t) is typeid(uint);
```

The following will still fail to compile...

```D
import std.format;
import std.stdio;
void main()
{
    enum s = "%1$s,%2$s".format("foo","bar");
    writeln(s);
}

// phobos/std/array.d(2850): Error: TypeInfo cannot be used with -betterC
```

jacob-carlborg:
So, if we merge this, won't that close the issue and show up in the next changelog?

JinShil:
> So, if we merge this, won't that close the issue and show up in the next changelog?

Yes, but I plan to reopen the issue when that happens (I don't know how else to deal with it unless I change the title and commit messages). The `enum` test case was a test added to that issue, so the issue is partially addressed by this PR.

jacob-carlborg:
I think you should change the title. Even if the issue is reopened and there's no changelog entry the commit message would still be misleading from someone reading the git log.

jacob-carlborg (changes requested):
Please change the commit message since it the PR doesn't fix the whole issue and the commit message would be misleading as is.

jacob-carlborg (approved):

PR #8174, opened 2018-04-16
Fix Issue 18762 - DMD should use a unique path/filename for __stdin.o

CyberShadow:
On my machine, this seems to allow building dlang.org reliably again.

### DIFF
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -579,6 +579,16 @@ extern (C++) final class Module : Package
         else
         {
             const(char)* argdoc;
+            OutBuffer buf;
+            if (!strcmp(arg, "__stdin.d"))
+            {
+                version (Posix)
+                    import core.sys.posix.unistd : getpid;
+                else version (Windows)
+                    import core.sys.windows.windows : getpid = GetCurrentProcessId;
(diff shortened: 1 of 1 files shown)


src/dmd/dmodule.d
```diff
@@ -579,6 +579,16 @@ extern (C++) final class Module : Package
...
         {
             const(char)* argdoc;
+            OutBuffer buf;
+            if (!strcmp(arg, "__stdin.d"))
+            {
+                version (Posix)
+                    import core.sys.posix.unistd : getpid;
+                else version (Windows)
+                    import core.sys.windows.windows : getpid = GetCurrentProcessId;
+                buf.printf("__stdin_%d.d", getpid());
```
CyberShadow: Poor man's `mkstemp`.

src/dmd/dmodule.d
```diff
@@ -579,6 +579,16 @@ extern (C++) final class Module : Package
         else
         {
             const(char)* argdoc;
+            OutBuffer buf;
```
CyberShadow: This (and the string inside) is scoped, but the string should always be copied (by `forceExt`) before the scope exits.

andralex (approved):

PR #8176, opened 2018-04-16
evalu8.c: refactor to remove shadowing of 'i'

WalterBright:
Replaced it with `b` and did some localization to reduce the scope of the variable.

(Shadowing is warned about by VC, and have to fix it anyway for converting to D, because D doesn't just warn about shadowing, it does not allow it.)


JinShil (approved):
I don't see anything alarming in the code.  Approving on faith in @WalterBright

PR #8177, opened 2018-04-16
cg87.c: remove shadowing of 'retregs'

WalterBright:


src/dmd/backend/cg87.c
```diff
@@ -9,7 +9,7 @@
  * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/cg87.c, backend/cg87.c)
  */
 
-#if !SPP
+#if (SCPP && !HTOD) || MARS
```
JinShil: I don't know what this is, but I trust @WalterBright does.

JinShil (approved):

src/dmd/backend/cg87.c
WalterBright: It was the result of some DMC work :-)

PR #8178, opened 2018-04-17
refactor: remove more shadowing declarations

WalterBright:


JinShil (approved):
Again, I don't understand all the semantics of this code, but I can see there's nothing special in this change; just some simple refactoring, so approving.

jacob-carlborg (approved):

PR #8179, opened 2018-04-17
[D1] Backport __ctfe

UplinkCoder:
Provides `__ctfe` in D1 as well

### DIFF
--- a/src/declaration.c
+++ b/src/declaration.c
@@ -1555,6 +1555,10 @@ void VarDeclaration::checkNestedReference(Scope *sc, Loc loc)
 
 int VarDeclaration::isDataseg()
 {
+    if (ident == Id::ctfe)
+    {
+        return true;
+    }
 #if 0
     printf("VarDeclaration::isDataseg(%p, '%s')\n", this, toChars());
     printf("%llx, %p, %p\n", storage_class & (STCstatic | STCconst), parent->isModule(), parent->isTemplateInstance());
--- a/src/e2ir.c
+++ b/src/e2ir.c
@@ -721,6 +721,10 @@ elem *VarExp::toElem(IRState *irs)
 
     //printf("VarExp::toElem('%s') %p\n", toChars(), this);
     //printf("\tparent = '%s'\n", var->parent ? var->parent->toChars() : "null");
+    if (var->ident == Id::ctfe)
+    {
+        return el_long(TYint, 0);
+    }
     if (var->needThis())
     {
         error("need 'this' to access member %s", toChars());
--- a/src/expression.c
+++ b/src/expression.c
@@ -2092,7 +2092,6 @@ Expression *IdentifierExp::semantic(Scope *sc)
         }
         return e->semantic(sc);
     }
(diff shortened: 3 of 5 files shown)


mihails-strasuns-sociomantic:
As usual, please link commit in D2 branch that introduced similar functionality for comparison.

stefan-koch-sociomantic:
@mihails-strasuns-sociomantic Apparently it was introduced in 
https://github.com/dlang/dmd/commit/b0a2add98184c972a0668d460d5d51b7eec37c8c

ibuclaw:
@stefan-koch-sociomantic - That is not the D2 branch?

ibuclaw:
This change looks about right anyway for `__ctfe` support.  Generates 1 during interpret, and 0 during codegen.

src/declaration.c
```diff
@@ -1555,6 +1555,10 @@ void VarDeclaration::checkNestedReference(Scope *sc, Loc loc)
 
 int VarDeclaration::isDataseg()
 {
+    if (ident == Id::ctfe)
+    {
+        return true;
+    }
```
ibuclaw: Except this, why is this needed?

ibuclaw (changes requested):

src/declaration.c
stefan-koch-sociomantic: because otherwise the error handling below this is triggered.

ibuclaw: Doesn't that indicate that something is missing though?  This variable should never be written to the data segment.

stefan-koch-sociomantic: it won't be.
Because it does not exist.
e2ir transforms the variable into a literal `0`.

mihails-strasuns-sociomantic: Maybe better make it assert in that case?

stefan-koch-sociomantic: Where would I assert ? `__ctfe` is not a real variable hence it does not have complete type information.
Because it does not have complete type information it would error in `isDataseg` if we would not intercept it.

stefan-koch-sociomantic:
wait ... why does @UplinkCoder want to merge it ?
I should want to merge it ?

Burgos:
Original use case comes from `enforce` not being CTFE-able IIRC. So, piece of code that in some cases _could_ be executed in CTFE, suddenly couldn't, because `enforce` is not allowed to be used. In that case it's possible to `if (__ctfe) assert(...); else enforce(...);`, but sadly not in D2.

Of course, this doesn't answer more philosophical question _why_ do we want this in D1 :-)

mihails-strasuns-sociomantic:
@leandro-lucarella-sociomantic basically combination of very low effort (most of implementation code is already present) and large potential benefit (enables compatible D1/D2 patterns like @nemanja-boric-sociomantic has explained)

leandro-lucarella-sociomantic:
OK, I would be very careful in backporting features unless there is a very good need for them, because it might be hard to asses it doesn't introduce any regressions. At this stage I think we need to focus in moving to D2 rather than backporting features to D1.

PR #8189, opened 2018-04-18
Fix Issue 14245 - Immutable reference to immutable field in constructor allows breaking type system

RazvanN7:
```d
immutable(int)* g;

struct X {
    int a = 10;
    immutable this(int x) {
        g = &a;
        a = 42;
    }
}

void main() {
    auto x = immutable X();
}
```
Taking the address of an immutable field before initialization should error since it will cause the violation of immutable once the field is initialized (g will point to a memory location which will first have the value of 10 and then 42).

This patch checks that if the address of an immutable field is taken before initialization (no matter what the reason is) the compiler will issue an error ("the address of an immutable field may not be taken before initialization").

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -4761,6 +4761,48 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         if (exp.e1.op == TOK.dotVariable)
         {
             DotVarExp dve = cast(DotVarExp)exp.e1;
+
+            // Fix Issue 14245 : https://issues.dlang.org/show_bug.cgi?id=14245
+            // The address of an uninitialized immutable field cannot be taken
+            // inside a constructor because it will break immutability once
+            // the field is assigned
+            if (sc.func)
+            {
+                // if inside a constructor
+                if(const ctd = sc.func.isCtorDeclaration())
+                {
+                    const decl = dve.var;
+
+                    // The below condition checks for (&this.var && var is immutable). Note that
+                    // var might not be declared immutable, but if the address is taken inside an
+                    // immutable constructor then var will be part  of an immutable object, thus
+                    // being immutable. This case can be identified by checking if the constructor
+                    // is immutable.
+                    if (dve.e1.op == TOK.this_ && decl.isField() && (decl.isImmutable() || ctd.type.isImmutable()))
+                    {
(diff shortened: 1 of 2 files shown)


src/dmd/expressionsem.d
```diff
@@ -4761,6 +4761,48 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+                    // being immutable. This case can be identified by checking if the constructor
+                    // is immutable.
+                    if (dve.e1.op == TOK.this_ && decl.isField() && (decl.isImmutable() || ctd.type.isImmutable()))
+                    {
+                        const dim = sc.ctorflow.fieldinit.length;
+                        auto ad = ctd.isMember2();
+
+                        size_t i;
+                        // obtain the position of the variable in the fieldinit array
+                        for (i = 0; i < dim; i++)
```
JinShil: Can we add an `indexOf` method to the `dmd.root.Array`?  I think I could have used it recently too.

jacob-carlborg: Fine by me.

JinShil:
I think there should probably be something about this added to the spec.

src/dmd/expressionsem.d
wilzbach: -> https://github.com/dlang/dmd/pull/8228

RazvanN7:
I think that this PR is now obsolete since [1] has accidentally solved this too by considering that a variable is initialized if the address for it is taken. While this PR would have rejected taking the address of an immutable before initialization, the code at [1] accepts the code but considers that if the address is taken then the variable is initialized (issuing an error on subsequent assignments):

```d
immutable(int)* g;

struct X {
    int a = 10;
    immutable this(int x) {
        g = &a; // this patch : error : cannot take address of uninitialized variable;
        a = 42; // [1] patch : error : field a initialized multiple times;
    }
}

void main() {
    auto x = immutable X();
}
```
Although I think that Walter's accidental fix is better, my only concern is that the error message is not that obvious (you end up initializing a variable without actually using it as the `lhs` of an AssignExp). I think that the best solution would be to close this PR and the filed bug report and update the docs to highlight the fact that when the address of a field is taken, the field is actually initialized to its current value. Thoughts?

[1] https://github.com/dlang/dmd/pull/8200/files

PR #8193, opened 2018-04-19
add paranoia test for floating point

rainers:
This is a semi-automatic translation of http://www.netlib.org/paranoia/paranoia.c to D (without any additional reformatting, so diffing files is still feasible).

As there are detected failures and defects in existing floating point operations I'm unsure how to handle these. Maybe downgrade to "flaws"? These don't cause the tests to fail ATM.

### DIFF
--- a/test/runnable/extra-files/paranoia.d
+++ b/test/runnable/extra-files/paranoia.d
@@ -1,0 +1,2284 @@
+module paranoia;
+
+/*	A D version of Kahan's Floating Point Test "Paranoia"
+
+			Rainer Schuetze, April 2018
+
+	This is a translation from the C version 
+	(http://www.netlib.org/paranoia/paranoia.c) by
+
+			Thos Sumner, UCSF, Feb. 1985
+			David Gay, BTL, Jan. 1986
+
+	This is a rewrite from the Pascal version by
+
+			B. A. Wichmann, 18 Jan. 1985
+
+	(and does NOT exhibit good C programming style).
+
+	Adjusted to use Standard C headers 19 Jan. 1992 (dmg);
+	compile with -DKR_headers or insert
+#define KR_headers
+	at the beginning if you have an old-style C compiler.
+
+(C) Apr 19 1983 in BASIC version by:
+	Professor W. M. Kahan,
+	567 Evans Hall
+	Electrical Engineering & Computer Science Dept.
+	University of California
+	Berkeley, California 94720
+	USA
+
+converted to Pascal by:
+	B. A. Wichmann
+	National Physical Laboratory
+	Teddington Middx
+	TW11 OLW
+	UK
+
+converted to C by:
+
+	David M. Gay		and	Thos Sumner
+	AT&T Bell Labs			Computer Center, Rm. U-76
+	600 Mountain Avenue		University of California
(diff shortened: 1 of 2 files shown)


WalterBright:
I recommend having paranoia return success even though it detected errors. The errors should then be filed in bugzilla.

wilzbach:
> I recommend having paranoia return success even though it detected errors. The errors should then be filed in bugzilla.

Doesn't this removes of the value of adding the testsuite to the CI setup in the first place?
Did you maybe mean setting the currently failing tests _individually_ to pass / allowing only them to fail?
(this way regressions can be prevented and detected by our CIs)

don-clugston-sociomantic:
This is clearly a program intended to be used by humans  --- hence the masses of print statements.
I think if you want to put it in a test suite, all that output should be suppressed (version statement, maybe?)

But...
Looks as though it doesn't like `sqrt`. That's interesting.

wilzbach:
There's no need to suppress the output. That's already handled by the testsuite. The output will only be shown on a failure, i.e. when you want to look at it and are happy to receive more than just failed from the CI.

rainers:
I've changed the failure and the serious error to "defect" and allowed these with the existing 6 others for extended precision.
I guess I'll have to debug this quite a bit to get understandable test cases to put into bugzilla.

WalterBright:
> Doesn't this removes of the value of adding the testsuite to the CI setup in the first place?Did you maybe mean setting the currently failing tests individually to pass / allowing only them to fail? (this way regressions can be prevented and detected by our CIs)

Ideally, Paranoia should be added to the test suite along with fixes for all the errors it detects. But the latter may be a long term issue, and I felt it was important to simply get Paranoia into the test suite as a baseline.

Modifying Paranoia diverges it from the original sacred text :-) and runs the risk of introducing bugs in the tests. Having the baseline in the repository helps head off such problems, or at least makes it easier to find and fix introduced problems.

It's similar to how I translated code from C to D. The baseline is a rote conversion. I NEVER fix bugs or refactor. I establish a baseline conversion, and then fix/refactor the baseline. This makes tracking down the cause of a bug infinitely easier.

WalterBright:
@rainers thanks for doing this work! I am impressed that longdouble did so well in it. Congratulations!

jacob-carlborg:
The DMD tests failed for one of my pull requests: 

```
Makefile:133: recipe for target 'test_results/runnable/paranoia.sh.out' failed
```

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3131202&isPull=true.

wilzbach:
That's weird:

```
The arithmetic diagnosed appears to be Excellent!
END OF TEST.
LINK : fatal error LNK1104: cannot open file 'test_results\runnable\paranoia.exe'
Error: linker exited with status 1104
```

rainers:
I suspect that there is some program still sitting on the just created file, e.g. a virus scanner, while the next test is compiled. It's probably more reliable to use different file names for consecutive tests. I'll update it in https://github.com/dlang/dmd/pull/8199.

WalterBright:
> I suspect that there is some program still sitting on the just created file, e.g. a virus scanner, while the next test is compiled.

Windows 7 is particularly bad about that, even if the virus scanner is disabled.

PR #8194, opened 2018-04-19
Fix Issue 18775, 12807: gently resolve alias this

FeepingCreature:
Only emit warnings during alias this resolution if the alias this later actually turns out to lead to a valid property access.

This fixes issues where the impropriety of the alias this is reported as warnings or errors, even if the alias this doesn't actually end up being taken.

The underlying problem is that alias this competes with UFCS for property resolution, and is evaluated first. So UFCS to an alias this type always triggers the alias this, even if it doesn't match call types.

~~Maybe the better course of action would be to check UFCS before alias this, but I don't know how to do that.~~ No, alias this should correctly take priority.

### DIFF
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -2997,6 +2997,8 @@ private extern(C++) final class DotExpVisitor : Visitor
     {
         //printf("Type.noMember(e: %s ident: %s flag: %d)\n", e.toChars(), ident.toChars(), flag);
 
+        bool gagError = flag & 1;
+
         static __gshared int nest;      // https://issues.dlang.org/show_bug.cgi?id=17380
 
         static Expression returnExp(Expression e)
@@ -3008,7 +3010,7 @@ private extern(C++) final class DotExpVisitor : Visitor
         if (++nest > 500)
         {
             .error(e.loc, "cannot resolve identifier `%s`", ident.toChars());
-            return returnExp(flag & 1 ? null : new ErrorExp());
+            return returnExp(gagError ? null : new ErrorExp());
         }
 
 
@@ -3062,13 +3064,12 @@ private extern(C++) final class DotExpVisitor : Visitor
                 auto dti = new DotTemplateInstanceExp(e.loc, e, Id.opDispatch, tiargs);
                 dti.ti.tempdecl = td;
                 /* opDispatch, which doesn't need IFTI,  may occur instantiate error.
-                 * It should be gagged if flag & 1.
                  * e.g.
                  *  template opDispatch(name) if (isValid!name) { ... }
                  */
-                uint errors = flag & 1 ? global.startGagging() : 0;
+                uint errors = gagError ? global.startGagging() : 0;
                 e = dti.semanticY(sc, 0);
-                if (flag & 1 && global.endGagging(errors))
+                if (gagError && global.endGagging(errors))
                     e = null;
                 return returnExp(e);
             }
@@ -3080,9 +3081,20 @@ private extern(C++) final class DotExpVisitor : Visitor
                 /* Rewrite e.ident as:
                  *  e.aliasthis.ident
                  */
-                e = resolveAliasThis(sc, e);
-                auto die = new DotIdExp(e.loc, e, ident);
-                return returnExp(die.semanticY(sc, flag & 1));
+                auto alias_e = resolveAliasThis(sc, e, gagError);
+
+                if (!alias_e)
+                    return returnExp(null);
+
+                auto die = new DotIdExp(e.loc, alias_e, ident);
+                auto exp = die.semanticY(sc, gagError);
+
+                if (exp && gagError)
+                    // now that we know that the alias this leads somewhere useful,
+                    // go back and print deprecations/warnings that we skipped earlier due to the gag
+                    resolveAliasThis(sc, e, false);
+
+                return returnExp(exp);
             }
         }
         visit(cast(Type)mt);
--- a/test/compilable/test12807.d
+++ b/test/compilable/test12807.d
@@ -1,0 +1,16 @@
+void foo(T)(ref T t)
+{
+}
+
+struct S
+{
+    int impure() {assert(0);}
+    alias impure this;
+}
+
+void main() pure
+{
+    S s;
+    foo(s);
+    s.foo(); // triggering alias this violates purity, but ufcs matches
+}
--- a/test/compilable/test18775.d
+++ b/test/compilable/test18775.d
@@ -1,0 +1,20 @@
+// REQUIRED_ARGS: -de
+
+struct Foo { }
+
+struct Bar {
+    deprecated
+    @property Foo foo() { return Foo.init; }
+
+    alias foo this;
+}
+
+void test(Bar bar) { }
+
+void main()
+{
+    Bar bar;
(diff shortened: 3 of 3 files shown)


FeepingCreature:
How do I get the bot to track two issues...?

test/compilable/test12807.d
```diff
@@ -0,0 +1,16 @@
+void foo(T)(ref T t)
```
RazvanN7: please add the test case from Issue 18775 also.

RazvanN7 (commented):
I am not convinced that the alias this should not report the deprecation. FWIW aliasing this a deprecated method implies that the line of code which implements the alias uses the deprecated function. I would argue that this calls for a deprecation message even if the alias this does not actually get to be used and it should be reported at the line where the alias this is created to a deprecated symbol.

test/compilable/test12807.d
FeepingCreature: How do I add a testcase that ensures that a deprecation message does not get printed?

FeepingCreature:
Unfortunately, at present deprecating the alias itself does nothing. However, even so, precedent seems to be that a deprecated function only triggers a warning when it is actually used.

In my opinion, an alias of a function should not in itself qualify as a use.

test/compilable/test12807.d
RazvanN7: add a comment like this : 

// REQUIRED_ARGS: -de

It tells the tester to treat deprecations as errors

FeepingCreature: Added.

jacob-carlborg:
The commits say: "Mathis Beer authored and Mathis Beer committed". It would be great if you could commit as the same user as the GitHub account.

FeepingCreature:
I'm working on this from my work pc, which is configured for my irl name. I'm not actually sure how to change my git identity; however, if anything, my irl name is more appropriate because it's work I'm doing "on company time."

jacob-carlborg:
I think it's based on the email address and the name used to commit. See https://help.github.com/articles/setting-your-commit-email-address-in-git
https://help.github.com/articles/setting-your-username-in-git/

jacob-carlborg:
You can also add your real name to your GitHub account without changing the account name. Have a look at my profile: https://github.com/jacob-carlborg

jacob-carlborg:
Even though you're not committing using the same email address as your GitHub account, there should be no reason for the committer and author to be different.

FeepingCreature:
Aah! The problem is that I really want to keep this _account_ largely separate from my work, but I also don't want to make a new Github account just for work. It's a bit awkward.

There should be no reason for the committer and author to be different. In fact, they are not different. I looked at the commit; they're exactly the same. I have no idea why Github shows them separately.

jacob-carlborg:
> There should be no reason for the committer and author to be different. In fact, they are not different. I looked at the commit; they're exactly the same. I have no idea why Github shows them separately.

Hmm, yes. Strange that GitHub thinks the committer and author are not the same.

FeepingCreature:
I think that semaphoreci failure is spurious.

FeepingCreature:
Yep!

FeepingCreature:
How do I request review again?

src/dmd/typesem.d
```diff
@@ -3080,9 +3081,30 @@ private extern(C++) final class DotExpVisitor : Visitor
...
-                return returnExp(die.semanticY(sc, flag & 1));
+                // gag always: we don't know if we'll actually
+                // succeed on the DotIdExp access later on, and if we fail it, we
+                // want to suppress warnings arising from the alias this access.
+                auto alias_e = resolveAliasThis(sc, e, true);
+
+                if (!alias_e)
+                {
+                    if (!gagError)
+                        resolveAliasThis(sc, e, false); // but we do want to report errors.
```
RazvanN7: Why is this necessary? Won't the errors be reported at line 3104?

RazvanN7: It it's necessary, why is `if (!gagError)` necessary?

FeepingCreature: No, because we're bailing here - we can't continue because alias_e is null, so we can't make a DotIdExp of it. This can happen if the target of the alias is itself invalid. This is the part that fixes Issue 12807.

FeepingCreature: I guess I could set e to the ungagged expression because it'll know that it's erroring.

FeepingCreature: No I can't! Because even so, the later property access could still fail a la 18775. So I cannot error that early ever, anyways.

I stand by the current impl.

FeepingCreature: ... No, we can keep going if we're not gagged. Because we'll be gagged via flag for the test to decide whether we can do ufcs after? ....? ???

FeepingCreature: Oh, fuck. Strictly speaking I can only emit deprecations if they are *not* errors.

FeepingCreature: Because if they *are* errors, we may want to fall through into ufcs?

FeepingCreature: ... I'll stick my neck out and say that the code as written _possibly_ does the Right Thing.

I think it definitely does a significantly righter thing than DMD does at present.

FeepingCreature: Hm. Maybe I've been coming at this wrong.

FeepingCreature: Okay! Here's the __issue__.

The problem is **warnings**. If I get an error in the alias resolve, I can just keep going with the errored exp; if it returns null, I can just return null since we'd be gagged. The problem is if the initial alias resolve prints a **warning**, then later on the dot lookup **fails** and we fall through to UFCS, we'll have printed a warning spuriously. That's the case that I have to workaround with the repeated resolveAliasThis calls.

FeepingCreature: I actually managed to find an improvement.

This code is fucking *hellish*.

RazvanN7 (approved):
Thanks for the effort. Looks good to me. I'll wait a few days before tagging auto-merge to offer the possibility to other reviewers to jump in.

FeepingCreature:
Cheers. <3

RazvanN7:
Looks like you shot 2 "rabbits" on your first shot in dmd :drum:

FeepingCreature:
Third or fourth shot, actually! And yeah, somebody pointed it out in the comments of the bug report, thankfully.

src/dmd/typesem.d
```diff
@@ -3080,9 +3081,20 @@ private extern(C++) final class DotExpVisitor : Visitor
...
+                if (!alias_e)
+                    return returnExp(null);
+
+                auto die = new DotIdExp(e.loc, alias_e, ident);
+                auto exp = die.semanticY(sc, gagError);
+
+                if (exp && gagError)
+                    // now that we know that the alias this leads somewhere useful,
+                    // go back and print deprecations/warnings that we skipped earlier due to the gag
+                    resolveAliasThis(sc, e, false);
```
ibuclaw: Makes sense, but pretty awful that you have to call this twice.

FeepingCreature: I agree. I don't know a better way though.

FeepingCreature:
Monday morning ping! :-)

PR #8195, opened 2018-04-19
Fix Issue 16206 - traits getOverloads fails when one of the overload is a templatized function

Biotronic:
This adds templates to the result of `__traits(getOverloads)`, allowing them to be iterated and inspected.

There's a few issues other than 16206 that will be fixed by this - I'll have a look-see through bugzilla tomorrow to find them.

It's possible this breaks code by considering template overloads where only function overloads were expected earlier. I expect this to happen rarely in real code, but it's certainly possible. It might be that some templates in Phobos will require updating to handle mixed function/template overload sets.

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -824,7 +824,7 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         e.ident == Id.getVirtualMethods ||
         e.ident == Id.getVirtualFunctions)
     {
-        if (dim != 2)
+        if (dim != 2 && !(dim == 3 && e.ident == Id.getOverloads))
             return dimError(2);
 
         auto o = (*e.args)[0];
@@ -835,6 +835,19 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
             return new ErrorExp();
         }
         ex = ex.ctfeInterpret();
+
+        bool includeTemplates = false;
+        if (dim == 3 && e.ident == Id.getOverloads)
+        {
+            auto b = isExpression((*e.args)[2]);
+            b = b.ctfeInterpret();
+            if (!b.type.equals(Type.tbool))
+            {
+                e.error("`bool` expected as third argument of `__traits(getOverloads)`, not `%s` of type `%s`", b.toChars(), b.type.toChars());
+                return new ErrorExp();
+            }
+            includeTemplates = b.isBool(true);
+        }
 
         StringExp se = ex.toStringExp();
         if (!se || se.len == 0)
@@ -908,7 +921,7 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
             /* Create tuple of functions of ex
              */
             auto exps = new Expressions();
-            FuncDeclaration f;
+            Dsymbol f;
             if (ex.op == TOK.variable)
             {
                 VarExp ve = cast(VarExp)ex;
@@ -924,9 +937,23 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
                 else
                     ex = dve.e1;
             }
+            else if (ex.op == TOK.template_)
+            {
+                VarExp ve = cast(VarExp)ex;
+                auto td = ve.var.isTemplateDeclaration();
+                f = td;
+                if (td && td.funcroot)
+                    f = td.funcroot;
+                ex = null;
+            }
 
             overloadApply(f, (Dsymbol s)
             {
+                if (includeTemplates)
+                {
+                    exps.push(new DsymbolExp(Loc.initial, s, false));
+                    return 0;
+                }
                 auto fd = s.isFuncDeclaration();
                 if (!fd)
                     return 0;
--- a/test/compilable/Test16206.d
+++ b/test/compilable/Test16206.d
@@ -1,0 +1,28 @@
+struct S {
+    static int foo()() { return 0; }
+    static int foo()(int n) { return 1; }
+    static int foo(string s) { return 2; }
+    enum foo(int[] arr) = arr.length;
+}
+
+alias AliasSeq(T...) = T;
+
+alias allFoos = AliasSeq!(__traits(getOverloads, S, "foo", true));
+
+static assert(allFoos.length == 4);
+
+static assert(allFoos[0]("") == 2);
+static assert(allFoos[1]() == 0);
+static assert(allFoos[2](1) == 1);
+alias foo3 = allFoos[3];
+static assert(foo3!([]) == 0);
+
+static assert(S.foo() == 0);
+static assert(S.foo(1) == 1);
+static assert(S.foo("") == 2);
+static assert(S.foo!([]) == 0);
+
+
+alias fooFuns = AliasSeq!(__traits(getOverloads, S, "foo"));
+static assert(fooFuns.length == 1);
+static assert(fooFuns[0]("") == 2);
--- a/test/fail_compilation/fail16206b.d
+++ b/test/fail_compilation/fail16206b.d
@@ -1,0 +1,12 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail16206b.d(12): Error: expected 2 arguments for `hasMember` but had 3
+---
+*/
+
+struct S {
+    static int foo()() { return 0; }
+}
+alias AliasSeq(T...) = T;
+alias allFoos = AliasSeq!(__traits(hasMember, S, "foo", true));
--- a/changelog/fix16206.dd
+++ b/changelog/fix16206.dd
@@ -1,0 +1,24 @@
+Support for iterating template overloads
+
+`__traits(getOverloads)` has been extended to return template overloads when passed an optional parameter with a true value.
+
+---
+struct S {
+    static int foo()() { return 0; }
+    static int foo()(int n) { return 1; }
+    static int foo(string s) { return 2; }
+    enum foo(int[] arr) = arr.length;
+}
+
+alias AliasSeq(T...) = T;
+
+alias allFoos = AliasSeq!(__traits(getOverloads, S, "foo", true));
+
+static assert(allFoos.length == 4);
+
+static assert(allFoos[0]("") == 2);
+static assert(allFoos[1]() == 0);
+static assert(allFoos[2](1) == 1);
+alias foo3 = allFoos[3];
+static assert(foo3!([]) == 0);
+---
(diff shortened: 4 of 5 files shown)


jacob-carlborg (approved):

jacob-carlborg:
It would be nice if the commit message contained the same info as the PR description.

Perhaps worth a changelog entry.

JinShil:
@BBasile Can you please investigate the iz failure in Jenkins?

BBasile:
@JinShil: It's something done with the overloads that doesn't work anymore when the overload is a template. Probably just a `static if (isTemplate` to add somewhere.

BBasile:
@JinShil, iz fix + tag pushed. Not sure of what was the last change in this PR, so maybe it wasn't necessary to do anything from my side ?

BBasile:
I don't know if adding the templates to the overloads is a good idea. The problem was mostly that the count was different. The fix, as done right now, is actually a breaking change.

Biotronic:
So I was proven wrong immediately:  
> I expect this to happen rarely in real code

The other option would be to create a new trait, something like `getOverloadsIncludingTemplates` (or some even better name :p). `getTemplateOverloads` would also be possible, but it seems likely that users will often be interested in not only templates but all symbols in an overload set.

For that matter, the issue of `getOverloads` returning nothing when the lexically first overload is a template, is just one part of this fix, and could perfectly well be done without introducing templates in the result.

RazvanN7:
@Biotronic I worked on this PR a couple of months ago : https://github.com/dlang/dmd/pull/7959/files . Maybe it will help you if you plan on fixing traits(getOverloads) entirely

BBasile:
> For that matter, the issue of getOverloads returning nothing when the lexically first overload is a template, is just one part of this fix, and could perfectly well be done without introducing templates in the result.

Yes. This in a first time sounds more reasonable.

src/dmd/traits.d
```diff
@@ -924,12 +924,34 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
                 else
                     ex = dve.e1;
             }
+            else if (ex.op == TOK.template_)
+            {
+                VarExp ve = cast(VarExp)ex;
+                auto td = ve.var.isTemplateDeclaration();
+                f = td;
```
RazvanN7: `f = ve.var.isTeplateDeclaration` and you can get rid of `td`

BBasile:
Possible trick to prevent the breaking change: add an optional bool parameter (which would means "include templatized funcs or not").

src/dmd/traits.d
```diff
@@ -924,12 +924,34 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+                if (td && td.funcroot)
+                    f = td.funcroot;
+                ex = null;
+            }
 
             overloadApply(f, (Dsymbol s)
             {
                 auto fd = s.isFuncDeclaration();
                 if (!fd)
```
RazvanN7: I think the code would look nicer like this:

```d
auto fd = s.isFuncDeclaration();
auto td = s.isTemplateDeclaration();

if (!fd && !td)
    return 0;
if (td) { /* do template stuff */}
else { /* do function stuff */}          // fd and td are mutually exclusive
```

test/compilable/Test16206.d
```diff
@@ -0,0 +1,17 @@
...
+
+alias foos = AliasSeq!(__traits(getOverloads, S, "foo"));
+
+static assert(foos[0]("") == 2);
+static assert(foos[1]() == 0);
+static assert(foos[2](1) == 1);
+
+static assert(S.foo() == 0);
+static assert(S.foo(1) == 1);
+static assert(S.foo("") == 2);
```
RazvanN7: add newline

RazvanN7 (changes requested):

src/dmd/traits.d
Biotronic: No can do - I use `td.funcroot` later, and that's defined in `TemplateDeclaration`, not in `Dsymbol`.

Biotronic:
Removed templates from result, since it makes Jenkins sad. Will make a separate PR for template overloads.

BBasile:
> Removed templates from result, since it makes Jenkins sad. Will make a separate PR for template overloads.

You misunderstand the problem. The problem is that you didn't only fix the issue, you also broke the semantic of getOverloads. Your separate PR will also make Jenkins sad. This is why i reverted https://github.com/BBasile/iz/commit/6cacdec694aef27e20ef98dcbd235d16f056d2df.

Biotronic:
I'm still somewhat confused. My second PR will be on the form `__traits(getAllOverloads)`, and will not change the semantics of `getOverloads` at all. Are you saying that having `getOverloads` return templates is OK?

Biotronic:
Seems sensible. As shown in Test16206.d, `__traits(getOverloads)` now takes an optional third argument. If this evaluates to true, all overloads are returned - no matter what sort of construct it is. If this evaluates to false, or is not provided, all regular function overloads will be returned. This still fixes the issue of order-of-definition.

ghost (approved):
With the hope that the optional param was not a bad idea.
Once something optional is added it's locked forever.

src/dmd/traits.d
```diff
@@ -836,6 +836,19 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
         ex = ex.ctfeInterpret();
 
+        bool includeTemplates = false;
+        if (dim == 3 && e.ident == Id.getOverloads)
+        {
+            auto b = isExpression((*e.args)[2]);
+            b = b.ctfeInterpret();
+            if (!b.type.equals(Type.tbool))
+            {
+                e.error("bool expected as third argument of __traits(getOverloads).");
```
JinShil: tick marks around `bool` and `__traits(getOverloads)` for highlighting in the error message.

JinShil: Also, we need a test case for this error message.

JinShil (changes requested):

src/dmd/traits.d
```diff
@@ -824,7 +824,7 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         e.ident == Id.getVirtualMethods ||
         e.ident == Id.getVirtualFunctions)
     {
-        if (dim != 2)
+        if (dim != 2 && !(dim == 3 && e.ident == Id.getOverloads))
             return dimError(2);
```
JinShil: codecov shows this line isn't covered in the test suit.

JinShil (changes requested):

JinShil:
This will require a documentation update at https://dlang.org/spec/traits.html#getOverloads

JinShil:
And it should probably have a changelog entry.

src/dmd/traits.d
Biotronic: Question: Should this be covered by a separate test file in the fail_compilation folder, or is a `static assert(!__traits(compiles))` in Test16206.d good enough?

Biotronic: Scratch that. Looking at the existing tests, I was able to notice a pattern. :p

Biotronic:
Added changelog entry, and a spec PR in https://github.com/dlang/dlang.org/pull/2351

RazvanN7 (approved):

aliak00:
Is adding an unnamed boolean argument preferred over having a new trait such as ‘getTemplateOverloads’. Personally I think at the call site the API without the “magic” bool would be quite a bit more readable.

jacob-carlborg:
Since `__traits` is kind of like a magic function, is it possible to support the following syntax `__traits(getOverloads, S, "foo", includeTemplates)` without `includeTemplates` having to be defined?

Biotronic:
Sorry for taking so long to look into this. Of course it's possible - it's basically what `__traits(compiles)` does. However, it would require much greater changes and make the code harder to follow.

I'm much more partial to aliak00's suggestion of `getTemplateOverloads`, though I feel that name is somewhat misleading, since we want it to return non-template overloads too. I argue that the best behavior is to return every element of what the compiler considers the 'overload set', which may be a mix of templates and functions, a set of templates, or one of a number of different things. `getAllOverloads` is a better match, but feels a lot like calling it `getOverloads2`. 😛 Maybe `getOverloadSet`? Though that begets the question of how it differs from `getOverloads`. Ah well.

I've already written and tested that code, so changing it to `getAllOverloads` would be easy.

JinShil:
@WalterBright, any concerns about this?

aliak00:
`getTemplateOverloads` would be misleading if it returned the whole set yeah (while typing I was thinking it would only return template overloads but i failed to type out my entire thought :p). 

`getOverloadsEx` ? ala MS APIs 😱 

But all suggestions IMO so far are better than a bool at least.

Biotronic:
`getSymbolOverloads`?

jacob-carlborg:
> getSymbolOverloads?

I think `Symbols` is too generic. There are other kind of symbols that cannot be overloaded.

RazvanN7:
@JinShil anything else blocking the merge of this PR?

JinShil:
I was hoping @WalterBright or @andralex would weigh in, since it affects the public API.

JinShil:
I was also thinking about maybe using a bitArray enum for the selection criteria.

```D
enum OverloadSelection
{
    nonTemplates = 1;
    templates = 2;
    couldThereBeOthers = 4;
}

__traits(getOverloads, mixin(__MODULE__), "S", OverloadSelection.Templates);
```

But that may be totally unnecessary.  Thoughts?

RazvanN7:
@JinShil I think that what you propose may be a candidate for std.traits . For now, I think that this PR is ready to go.

JinShil (approved):

test/fail_compilation/fail16206b.d
```diff
@@ -0,0 +1,12 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail16206b.d(12): Error: expected 2 arguments for `hasMember` but had 3
+---
+*/
+
+struct S {
```
jacob-carlborg: Curly braces on their own lines. Applies to all test files and the changelog entry.

wilzbach: OT:
Btw I am not sure whether we need to be so picky for the test files, because it mostly doesn't matter there.
No one looks actively at the tests anyhow, but only when they fail.

jacob-carlborg: BTW, I think that tests should be treated just as the rest of the code. Same standards, style, refactoring and so on.

PR #8196, opened 2018-04-20
better error message for else without if

WalterBright:


JinShil (approved):

src/dmd/parse.d
```diff
@@ -5523,6 +5523,11 @@ final class Parser(AST) : Lexer
                     s = null; // don't propagate parsing errors
                 break;
             }
+
+        case TOK.else_:
+            error("found `else` without a corresponding `if`, `version` or `debug` statement");
```
jacob-carlborg: Do we want to include `static if` in the error message?

WalterBright: `static if` is covered by `if`

PR #8198, opened 2018-04-20
Fix Issue 11742 - cannot inizialize void initialized static variable in static constructor

RazvanN7:
If a non-mutable variable is initialized at module level, it cannot be modified in the module static constructor because that would invalidate the initial initialization. However, if the non-mutable variable is initialized with `void` this should not count as an initialization and the static constructor should be able to initialize it.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -1019,7 +1019,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             }
         }
 
-        if (!dsym._init && !fd)
+        if ((!dsym._init || dsym._init.isVoidInitializer) && !fd)
         {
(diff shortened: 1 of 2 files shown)


jacob-carlborg (approved):

test/runnable/test11742.d
```diff
@@ -0,0 +1,10 @@
+// https://issues.dlang.org/show_bug.cgi?id=11742
```
WalterBright: Can this be merged into one of the "lots of test cases" files, like test42.d? The reason is then the autotester runs faster.

WalterBright (changes requested):
See comment. Otherwise, looks good!

RazvanN7:
@WalterBright done. thanks!

WalterBright (approved):

PR #8199, opened 2018-04-21
run floating point paranoia test via CTFE

rainers:
While investigating the FP failures in #8193 I noticed that they go away when replacing core.stdc.math with std.math. There is a single "flaw" left for `double` with respect to rounding.

Running paranoia at compile time shows unexpected precision (all FP types calculated as reals) and similar results as the runtime tests before because dmd uses core.stdc.math for `sqrt` and `pow`.

### DIFF
--- a/test/runnable/extra-files/paranoia.d
+++ b/test/runnable/extra-files/paranoia.d
@@ -229,16 +229,23 @@ q
 import core.stdc.stdio;
 import core.stdc.signal;
 import core.stdc.stdlib;
-import core.stdc.math;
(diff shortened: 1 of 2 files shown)


WalterBright:
We don't really care about the precision of core.stdc.math anyway, because that's the C compiler vendor's responsibility. I've found that the precision of that is erratic across different platforms - an assertion that arose the ire of a professor I talked to about this. He couldn't believe that modern C compilers had sloppy precision in their math.h implementations.

PR #8200, opened 2018-04-22
fix Issue 15869 - RVO can overwrite argument

WalterBright:


src/dmd/expression.d
```diff
@@ -5321,6 +5322,30 @@ extern (C++) final class DotVarExp : UnaExp
...
+        if (e1.op == TOK.this_ && sc.ctorflow.fieldinit.length && !(sc.ctorflow.callSuper & CSX.any_ctor))
+        {
+            if (VarDeclaration vd = var.isVarDeclaration())
+            {
+                auto ad = vd.isMember2();
+                if (ad && ad.fields.dim == sc.ctorflow.fieldinit.length)
+                {
+                    foreach (i, f; ad.fields)
+                    {
+                        if (f == vd)
```
ibuclaw: If (f != vd) continue? As its quite a nest you have here of statements.

src/dmd/expression.d
```diff
@@ -5321,6 +5322,30 @@ extern (C++) final class DotVarExp : UnaExp
...
+                        {
+                            if (!(sc.ctorflow.fieldinit[i] & CSX.this_ctor))
+                            {
+                                /* If the address of vd is taken, assume it is thereby initialized
+                                 * https://issues.dlang.org/show_bug.cgi?id=15869
+                                 */
+                                modifyFieldVar(loc, sc, vd, e1);
```
RazvanN7: what if the value isn't actually initialized? Maybe it should be check that `vd._init !is null` otherwise issue an error that the address cannot be taken if the variable isn't initialized and the default constructor is disabled

WalterBright: It not being initialized should only be possible in @system code, where anything goes.

RazvanN7: I find it odd that the compiler forces you to initialize the field but lets you use it before initialization. I agree that if the struct does have a default constructor it should be allowed, but when it's missing why am I forced to initialize it if I can use the struct anyway?
For example, this code compiles:

```d
struct A
{
    @disable this();
    this(string) {}
    int c;
}

struct B
{
    A a;
    int c;
    this(int)
    {   
        c = a.c; 
        a = A("string"); // if this is commented, error : field `a` must be initialized in constructor
       // why? if I can use the fields anyway...
    }   
}

void main() {}
```

WalterBright: By having a constructor, and default construction disabled, you have to initialize it explicitly rather than it being default initialized. It's the whole point of disabling default construction. The initial value, however, is still defined and is still useful.

RazvanN7 (approved):

src/dmd/expression.d
```diff
@@ -5321,6 +5322,30 @@ extern (C++) final class DotVarExp : UnaExp
...
+                auto ad = vd.isMember2();
+                if (ad && ad.fields.dim == sc.ctorflow.fieldinit.length)
+                {
+                    foreach (i, f; ad.fields)
+                    {
+                        if (f == vd)
+                        {
+                            if (!(sc.ctorflow.fieldinit[i] & CSX.this_ctor))
+                            {
+                                /* If the address of vd is taken, assume it is thereby initialized
```
RazvanN7: This needs to be documented somewhere, otherwise you may end up with surprising error messages. For example, this code:

immutable(int)* g;

```d
struct X { 
    int a = 10; 
    immutable this(int x)
    {   
            g = &a; 
            a = 42;    // error
    }   
}

void main()
{
    auto x = immutable X();
}
```

Leads to : `test.d(7): Error: immutable field a initialized multiple times` which is correct but not obvious

yshui: I think we should point to the first initialization in the error message. And maybe a special message explaining why taking address is initialization.

PR #8201, opened 2018-04-22
Fix Issue 12764 - Disabled struct default construction circumvented when field is written to

RazvanN7:
```d
struct S
{
    @disable this();

    this(string) { }
    int f;
}

class C
{
    this(int)
    {
        s.f = 1;  // circumvents default ctor!
    }

    S s;
}
```
`f` cannot be initialized before the object was constructed and since no call to the constructor was made, `s` can be partially initialized. The fix makes it so that if `s` does not have a default initializer (`S s = S("some string");` ) or a constructor call wasn't made in the constructor an error is issued.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -5313,6 +5313,47 @@ extern (C++) final class DotVarExp : UnaExp
         if (e1.op == TOK.this_)
             return var.checkModify(loc, sc, e1, flag);
 
+        /* https://issues.dlang.org/show_bug.cgi?id=12764
+         * If inside a constructor and an expression of type `this.field.var`
+         * is encountered, where `field` is a struct declaration with
+         * default construction disabled, we must make sure that
+         * assigning to `var` does not imply that `field` was initialized
+         */
+        if (sc.func)
+        {
+            auto ctd = sc.func.isCtorDeclaration();
+
+            // if inside a constructor scope and e1 of this DotVarExp
+            // is a DotVarExp, then check if e1.e1 is a `this` identifier
+            if (ctd && e1.op == TOK.dotVariable)
+            {
+                scope dve = cast(DotVarExp)e1;
+                if (dve.e1.op == TOK.this_)
+                {
+                    scope v = dve.var.isVarDeclaration();
+                    /* if v is a struct member field with no initializer, no default construction
+                     * and v wasn't intialized before
+                     */
+                    if (v && v.isField() && v.type.ty == Tstruct && !v._init && !v.ctorinit)
+                    {
+                        const sd = (cast(TypeStruct)v.type).sym;
+                        if (sd.noDefaultCtor)
+                        {
+                            /* checkModify will consider that this is an initialization
+                             * of v while it is actually an assignment of a field of v
+                             */
+                            scope modifyLevel = v.checkModify(loc, sc, dve.e1, flag);
+                            // reflect that assigning a field of v is not initialization of v
+                            v.ctorinit = false;
+                            if (modifyLevel == 2)
+                                return 1;
+                            return modifyLevel;
+                        }
+                    }
+                }
+            }
+        }
+
         //printf("\te1 = %s\n", e1.toChars());
         return e1.checkModifiable(sc, flag);
     }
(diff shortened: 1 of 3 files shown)


rainers:
> Windows failure seems unrelated

Yes, see https://github.com/dlang/dmd/pull/8202

src/dmd/expression.d
```diff
@@ -5307,6 +5307,49 @@ extern (C++) final class DotVarExp : UnaExp
...
+                    // if v is a struct member field with no initializer and no default construction
+                    if (v && v.isField() && v.type.ty == Tstruct && !v._init)
+                    {
+                        const sd = (cast(TypeStruct)v.type).sym;
+                        if (sd.noDefaultCtor)
+                        {
+                            // check if v was initialized. if not we must error
+                            const ad = ctd.isMember2();
+                            size_t i;
+                            for (i = 0; i < sc.ctorflow.fieldinit.length; i++)
```
JinShil: I think we need an `Array.indexOf` method.

JinShil (approved):

src/dmd/expression.d
```diff
@@ -5307,6 +5307,49 @@ extern (C++) final class DotVarExp : UnaExp
         if (e1.op == TOK.this_)
             return var.checkModify(loc, sc, e1, flag);
 
+        // Issue 12764 : https://issues.dlang.org/show_bug.cgi?id=12764
```
WalterBright: Please delete "Issue 12764 :" as it is redundant with the URL. In general, comments should not contain non-essential verbiage.

src/dmd/expression.d
```diff
@@ -5307,6 +5307,49 @@ extern (C++) final class DotVarExp : UnaExp
...
+
+        // If inside a constructor and an expression of type this.field.var
+        // is encountered, where field is a struct declaration with
+        // default construction disabled, we must check if that field was
+        // properly initialized before being able to actually use its
+        // fields.
```
WalterBright: Please use:
```
/* ...
 * ...
 */
```
for block comments.

src/dmd/expression.d
```diff
@@ -5307,6 +5307,49 @@ extern (C++) final class DotVarExp : UnaExp
...
+                            size_t i;
+                            for (i = 0; i < sc.ctorflow.fieldinit.length; i++)
+                            {
+                                if (ad.fields[i] == v)
+                                    break;
+                            }
+                            const fi = sc.ctorflow.fieldinit[i];
+                            import dmd.ctorflow : CSX;
+                            if (!(fi & CSX.this_ctor))
+                                .error(loc, "`struct` field `%s` must be initialized before accessing any of its members as it cannot be default constructed", v.toChars());
```
WalterBright: The bug is not that it must be initialized, but that it must be initialized *with a constructor call*.

WalterBright: Also, the bug is about `s.f = 1;` which is initialization of `s` by assignment. But this patch is also detecting rvalue usage, such as `x = s.f;`, which should be allowed.

src/dmd/expression.d
wilzbach: -> https://github.com/dlang/dmd/pull/8228

src/dmd/expression.d
RazvanN7: >Also, the bug is about s.f = 1; which is initialization of s by assignment. But this patch is also detecting rvalue usage, such as x = s.f;, which should be allowed.
 

That is not true. The current patch does not affect the case of `x = s.f;` because checkModifiable is called inside the AssignExp semantic visitor node for `lhs` of the AssignExp. checkModifiable is never called for a `rhs`.

>The bug is not that it must be initialized, but that it must be initialized with a constructor call.

It looks like you are right, although I find it harder and harder to understand why it should be allowed to change the value of `s.f` (which is equivalent to partially initializing the struct) **before** calling the constructor

andralex:
We have one immutable rule here to obey: "All calls to constructors start from an object in its initial state". If this fix still allows calling the constructor on top of a partially initialized object, it needs to be modified.

JinShil:
> "All calls to constructors start from an object in its initial state"

I don't see that being violated with this fix.  @RazvanN7, is this good to go?

RazvanN7:
@JinShil From my point of view, yes.

MoonlightSentinel:
This PR introduced a regression: https://issues.dlang.org/show_bug.cgi?id=22118

PR #8202, opened 2018-04-22
fix spurious test failures for test18322.d

rainers:
by using index over all tests to avoid identical output names in consecutive tests. Output files might still be locked by bacground pograms (e.g. virus scanner).

### DIFF
--- a/test/tools/d_do_test.d
+++ b/test/tools/d_do_test.d
@@ -862,19 +862,21 @@ int tryMain(string[] args)
         }
     }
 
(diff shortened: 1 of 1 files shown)


RazvanN7 (approved):

wilzbach:
Probably the biggest win will come from running the bash scripts in a temporary directory and thus not running into these problems.
I recently went through all bash scripts and prepared them for this switch by introducing the unique `OUTPUT_BASE`, so probably not much more than setting that one to a temporary directory will be required.

PR #8203, opened 2018-04-23
Fix Issue 5153 - Struct pointer to struct variable assign error message

RazvanN7:


test/fail_compilation/fail5153.d
```diff
@@ -0,0 +1,27 @@
+// https://issues.dlang.org/show_bug.cgi?id=5153
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail5153.d(25): Error: cannot initialize a `struct` object using `new`. Remove the `new` keyword.
```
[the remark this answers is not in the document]
ZombineDev: It may be an overkill, but what about something along the lines of:
```
The type of `new Foo` is `Foo*` (pointer to `Foo`), whereas the typeof the variable `f` is `Foo`. Consider removing the `new` keyword before the constructor call, or change the type of `f` to `Foo*`.
```

src/dmd/expressionsem.d
```diff
@@ -6774,6 +6774,28 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+                        if (exp.e2.op == TOK.new_)
+                        {
+                            auto newExp = cast(NewExp)(exp.e2);
+                            if (newExp.newtype && newExp.newtype == t1)
+                            {
+                                error(exp.loc, "cannot initialize a `struct` object using `new`. Remove the `new` keyword.");
+                                return setError();
+                            }
+                        }
+
```
WalterBright: The error message should also print the offending expression, which will make it easier to find in complex code. Something along the lines of:
```
cannot implicitly convert expression `new Foo(0)` of type `Foo*` to `Foo`, perhaps remove the `new` keyword?
```

src/dmd/expressionsem.d
```diff
@@ -6774,6 +6774,28 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+                        // Using `new` to initialize a struct object is a common mistake, but
+                        // the error message from the compiler is not very helpful in that
+                        // case. If exp.e2 is a NewExp and the type of new is the same as
+                        // the type as exp.e1 (struct in this case), then we know for sure
+                        // that the user wants to instantiate a struct. This is done to avoid
+                        // issuing an error when the user actually wants to call a constructor
+                        // which receives a class object.
+                        //
+                        // Foo f = new Foo2(0); is a valid expression if Foo has a constructor
+                        // which receives an instance of a Foo2 class
```
WalterBright: Please use:
```
/* ...
 * ...
 */
```
for multiline comments.

test/fail_compilation/fail5153.d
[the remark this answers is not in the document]
RazvanN7: `lhs` of the AssignExp may not necessarily be a variable declaration, it might be just `f` or `some.object.f`. In order to put the above mentioned message, which I certainly think is an overkill, I would have to specifically test for this particular case (where `lhs` is a declaration expression). I don't think that that level of specificity is required and from my experience when a user uses `Foo f = new Foo(0)` in the vast majority of cases the fix is `Foo f = Foo(0)` and not `Foo *f = new Foo(0)`

src/dmd/expressionsem.d
```diff
@@ -6774,6 +6774,31 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+                         * which receives an instance of a Foo2 class
+                         */
+                        if (exp.e2.op == TOK.new_)
+                        {
+                            auto newExp = cast(NewExp)(exp.e2);
+                            if (newExp.newtype && newExp.newtype == t1)
+                            {
+                                error(exp.loc, "cannot implicitly convert expression `%s` of type `%s` to `%s`",
+                                      newExp.toChars(), newExp.type.toChars(), t1.toChars());
+                                errorSupplemental(exp.loc, "Perhaps remove the `new` keyword?");
```
RazvanN7: What do you guys think about adding a function called `errorFixSuggestion` which would be used to output the suggestions on how to fix the error and would color the message in green? This would naturally organize the output error message for both the programmer and the user.

JinShil: I won't object to it, but I'm OK without it, too.

JinShil (approved):

jacob-carlborg (approved):

PR #8204, opened 2018-04-24
Add ability to use interfaces and classes without the runtime if they only contain static members

JinShil:
This is a followup to the PRs previously released in 2.079 that enable users to use D without the runtime in a pay-as-you-go fashion

 * #7786 for `Throwable`
 * #7799 for `TypeInfo`

Currently, even with the above PRs, the following code will cause the compiler to emit an error about `TypeInfo` if compiled without the runtime.  However, `TypeInfo` is in no way required for such code because all interface/class members are static.

```D
interface I
{
    static int i;
}

class A : I
{
    static int a;
}

class B : A
{
    static int b;
}

void main() 
{ 
    B.i = 32;
    B.a = 42;
    B.b = 52;
}
```
This PR removes this arbitrary limitation enabling more features of D to be used without the runtime.

### DIFF
--- a/changelog/min_runtime_classes.dd
+++ b/changelog/min_runtime_classes.dd
@@ -1,0 +1,100 @@
+Interfaces and classes can be used without the runtime if only static fields are utilized
+
+Prior to this release any attempt to use interfaces and/or classes without the runtime would result
+in compile-time errors due to the lack of `TypeInfo`.  However, as long as classes are not instantiated
+there is no need for `TypeInfo`.
+
+Beginning with this release the compiler will no longer emit errors about missing `TypeInfo` when
+using interfaces and/or classes as long as they are not instantiated and only `shared static` members are
+utilized.
+
+$(B Example 1)
+---
+module object
+
+private alias extern(C) int function(char[][] args) MainFunc;
+private extern (C) int _d_run_main(int argc, char** argv, MainFunc mainFunc)
+{
+    return mainFunc(null);  // assumes `void main()` for simplicity
+}
+---
+
+---
+interface I
+{
+    shared static int i;
+}
+
+class A : I
+{
+    shared static int a;
+}
+
+class B : A
+{
+    shared static int b;
+}
+
+void main()
+{
+    B.b = B.a + B.i;
+}
+---
+
+$(CONSOLE
+dmd -conf= -defaultlib= main.d object.d -of=main
+size main
+   text    data     bss     dec     hex filename
+   1867    1208      32    3107     c23 main
+)
+
+Non-shared static members can also be used, but will require a thread-local storage (TLS) implementation.
+For example, on Linux the TLS implementation is already supplied by the C runtime and C
+standard library, so, since dmd automatically calls gcc to link the final executable, it will automatically
+bring in the TLS implementation.
+
+$(B Example 2)
+---
+module object
+
+private alias extern(C) int function(char[][] args) MainFunc;
+private extern (C) int _d_run_main(int argc, char** argv, MainFunc mainFunc)
+{
+    return mainFunc(null);  // assumes `void main()` for simplicity
+}
+---
+
+---
+interface I
+{
+    static int i;
+}
+
+class A : I
+{
+    static int a;
+}
+
+class B : A
+{
+    static int b;
+}
+
+void main()
+{
+    B.b = B.a + B.i;
+}
+---
+
+$(CONSOLE
+dmd -conf= -defaultlib= main.d object.d -of=main
+size main
+   text    data     bss     dec     hex filename
+   2123    1296      28    3447     d77 main
+)
+
+Some platforms may require some TLS implementation code or some specialized build procedures to link
+in a TLS implementation.
+
+This will hopefully reduce friction for those using D without the runtime, porting D to new platforms,
+or using D from other langauges, while enabling more features and idioms of D to be used in those use cases.
--- a/test/runnable/minimal2.d
+++ b/test/runnable/minimal2.d
@@ -1,0 +1,44 @@
+// DFLAGS:
+// REQUIRED_ARGS: -defaultlib= runnable/extra-files/minimal/object.d
+
+// This test ensures that interfaces and classes can be used in a minimal
+// runtime as long as they only contain shared static members.  Non-shared
+// static members would require a thread-local storage (TLS) implementation.
+
+interface I
+{
+    shared static int i;
+}
+
+class A : I
+{
+    shared static int a;
+}
+
+class B : A
+{
+    shared static int b;
+
+    static int sumAll()
+    {
+        return b + a + i;
+    }
+}
+
+void poorMansAssert(bool condition)
+{
+    if (!condition)
+    {
+        asm {hlt;}
+    }
+}
+
+void main()
+{
+    B.i = 32;
+    B.a = 42;
+    B.b = 52;
+
+    poorMansAssert(B.i == 32 || B.a == 42 || B.b == 52);
+    poorMansAssert(B.sumAll() == (32 + 42 + 52));
+}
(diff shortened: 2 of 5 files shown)


jacob-carlborg:
There might some confusion around which part of a feature can be used where.

WalterBright (approved):

JinShil:
@jacob-carlborg Are you a Mac user?  If so, any chance I can persuade you to look into the Darwin failure?  I'm mostly concerned with why it behave differently from Linux, and would like to reduce the discrepancies.

jacob-carlborg:
I’ll have a look.

jacob-carlborg:
@JinShil macOS 32-bit is using a custom implementation of TLS. For every access to a TLS variable, the compiler generates a call to the `___tls_get_addr` function which is defined in druntime. The `-defaultlib=` flag will cause druntime to not be linked meaning this symbol won't exist.

This does not occur on macOS 64-bit since it's using the system provided functionality for TLS (implemented in the dynamic loader). The reason why it's not used on 32-bit is because it did not work out of the box (I don't see a reason why it wouldn't, they both use the same system). Because macOS 32-bit is a dead platform we felt it wasn't worth spending time trying to fix the issue.

JinShil:
Thank you @jacob-carlborg, very helpful. I've disabled the runnable test for osx32 and added a compilable test.  See my comments in the test files for justification.

@rainers Do you have any insight to offer for the Win32_64 error?  The exit code is 5, which, based on what I could find, is an access denied error.

jacob-carlborg:
> I've disabled the runnable test for osx32 and added a compilable test.

@JinShil an alternative would be to not use TLS variables, or do you want to explicitly test that?

JinShil:
> an alternative would be to not use TLS variables, or do you want to explicitly test that?

I specifically want to test TLS members as I think that is the most common usage.

rainers:
> @rainers Do you have any insight to offer for the Win32_64 error? The exit code is 5, which, based on what I could find, is an access denied error.

@JinShil 5 seems an abbreviation of 0xc0000005 which is an access violation. The crash happens in the `B.a = 42` assignment:

```
    39:     B.i = 32;
00007FF704BC1087 B8 20 00 00 00       mov         eax,20h  
00007FF704BC108C 65 48 8B 0C 25 58 00 00 00 mov         rcx,qword ptr gs:[58h]  
00007FF704BC1095 48 8B 11             mov         rdx,qword ptr [rcx]  
00007FF704BC1098 BB 00 00 00 00       mov         ebx,0  
    38: { 
    39:     B.i = 32;
00007FF704BC109D 89 04 1A             mov         dword ptr [rdx+rbx],eax  
    40:     B.a = 42;
00007FF704BC10A0 BE 2A 00 00 00       mov         esi,2Ah  
00007FF704BC10A5 65 48 8B 3C 25 58 00 00 00 mov         rdi,qword ptr gs:[58h]  
00007FF704BC10AE 48 8B 0F             mov         rcx,qword ptr [rdi]  <---- crashes here
00007FF704BC10B1 BA 04 00 00 00       mov         edx,4  
00007FF704BC10B6 89 34 11             mov         dword ptr [rcx+rdx],esi  
```

TLS seems broken: the first entry points to the tls_array table itself, so the assignemnt before overwrites the pointer. I guess it misses some of the magic from druntime/src/rt/msvct_stub.c

JinShil:
@jacob-carlborg and @rainers, Thank you for your help.

OSX32 and Windows have at least part of their TLS implementation in the D runtime, so if users opt into using TLS variables and are not using a runtime, or creating their own runtime, they will need to link to an existing TLS implementation or build their own.  I've, therefore, moved the TLS test to the compilable folder and modified the runnable test so it executes only for `shared static` members, no longer requiring a TLS implementation.

I would prefer to emit a compile-time error if a TLS implementation is not found, but that would require reworking a lot of code, and probably be too disruptive to support this niche use case.

With this PR, users will at least no longer be obstructed with an irrelevant TypeInfo error.

It's all green now, and ready to go..

test/runnable/minimal2.d
```diff
@@ -0,0 +1,34 @@
+// DFLAGS:
+// REQUIRED_ARGS: -defaultlib= runnable/extra-files/minimal/object.d
+
+// This test ensures that interfaces and classes can be used in a minimal
+// runtime as long as they only contain shared static members.  Non-shared
+// static members would a thread-local storage (TLS) implementation.
```
jacob-carlborg: The last sentence sounds a bit weird. Is it missing a “required”?

jacob-carlborg (dismissed):
There’s no changelog entry.

rainers:
The necessary TLS magic for Windows is in the C runtime, too. What is missing is a reference to the respective symbols, but it is optimized away under the assumption that the code ends up in the executable image. If you compile it separately, `_tls_index` will be used and the code should be fine. The test than fails due to missing `_d_run_main`, though.

jacob-carlborg:
> The test than fails due to missing _d_run_main, though.

That symbol is available in the `runnable/extra-files/minimal/object.d` file, which is included in the test as well [1].

[1] https://github.com/dlang/dmd/pull/8204/files#diff-26f06e2cdfaed27b14b8a004281f0798R2

jacob-carlborg:
For macOS 32-bit you can add a dummy implementation:

```d
extern(D) void* ___tls_get_addr( void* p )
{
    return p;
}
```

It won't be a true TLS but you're not using threads so...

rainers:
I have just added COMPILE_SEPARATELY to the test, but that didn't pick up object.obj (for obvious reasons). There doesn't seem to be a simple way to add it to the link command line.

JinShil:
@jacob-carlborg and @rainers, are you concerned about having a runnable test for osx32 and Windows?

jacob-carlborg:
>  are you concerned about having a runnable test for osx32

No.

JinShil:
> I have just added COMPILE_SEPARATELY to the test, but that didn't pick up object.obj (for obvious reasons). There doesn't seem to be a simple way to add it to the link command line.

You add the `_d_run_main` implementation directly to the `minimal2.d` file and it should work then. But still, I'm not sure how to link once the `.obj` file is created.

rainers:
> You add the _d_run_main implementation directly to the minimal2.d file and it should work then

Yes, that works with COMPILE_SEPARATELY and runs successfully.

rainers:
> are you concerned about having a runnable test for osx32 and Windows?

Not really ;-)

changelog/min_runtime_classes.dd
```diff
@@ -0,0 +1,100 @@
...
+
+$(CONSOLE
+dmd -conf= -defaultlib= main.d object.d -of=main
+size main
+   text    data     bss     dec     hex filename
+   1867    1208      32    3107     c23 main
+)
+
+Non-shared static members can also be used, but will require a thread-local storage (TLS) implementation.
+For Linux, the TLS implementation is already supplied by the C runtime and C
```
jacob-carlborg: Don't we know that there are only two platforms where it doesn't work out of the box?

JinShil: All the other tests in the auto-tester passed except osx32 and Win32_64.  So those are the only supported platforms I know of that don't work out of the box.

jacob-carlborg: Exactly, so why only mentioning `Linux`?

jacob-carlborg:
Do we want/is there a need to test static methods as well?

JinShil:
> Do we want/is there a need to test static methods as well?

Added a method to the runtime test.

changelog/min_runtime_classes.dd
JinShil: I'm just using Linux and an illustrative example.  How else should I say it?  "For *nix platforms..."?

JinShil: I'll add "For example.."

jacob-carlborg (approved):

12345swordy:
Wait hold on, does this mean it's possible for determine the attributes for destroy at compile time because of this?

JinShil:
> Wait hold on, does this mean it's possible for determine the attributes for destroy at compile time because of this?

I don't see how it would.

PR #8208, opened 2018-04-25
Fix Issue 11970 - Self-assigned variable expression with no effect doesn't have compiler error

JinShil:
~Please review carefully.~

~I'm not sure if identity assignment is a thing in D.  It certainly *feels* wrong, but there were a few tests I had to modify because they were explicitly testing for it.  Perhaps it may be necessary to restrict the error in this PR to basic types or other more specific conditions.~

There have been some significant changes in the implementation since I wrote the comments above.  The current implementation is much more selective and precise about which expressions have no side effects.

To emphasize the motivation for this PR, please see https://github.com/vibe-d/vibe.d/issues/406 as an example of a production bug that could have been prevented with this change.


scarface-one:
I think it would make more sense to have compiler error only for types that define no `postblit` or `opAssign`.

JinShil:
> I think it would make more sense to have compiler error only for types that define no postblit or opAssign.

Actually, this change currently does that.  I'm currently scouring the test suite to see if that behavior is already tested.  If not I'll add some tests to confirm that that behavior is verified.

JinShil:
> Is this an alternative fix for : #8119 ?

Gah! I'm drowning in the PR sea.  Yes, this PR addresses the same problem, but should also satisfy [@LemonBoy's concern](https://github.com/dlang/dmd/pull/8119#issuecomment-378180553)

Everything is working except arrays of struct with postblits.  Once I get that sorted out, this should be good to go.

WalterBright:
I'm not so sure this is a good idea, because of metaprogramming. It may "fall out" of some generic code, like `x+0` might fall out. Making it an error then would burden the programmer to try to find a workaround.

It's sort of like the "identity function" which does nothing but return its parameter.

wilzbach:
Duplicate of https://github.com/dlang/dmd/pull/8119?

test/runnable/builtin.d
```diff
@@ -45,10 +45,6 @@ void test2()
     i = i ^^ 2;
     assert(i == 9);
 
-    int j = 2;
-    j = j ^^ 1;
```
JinShil: I think this fails with this PR due to some aggressive const folding.  Interestingly `j = j + 0` does not generate an error, but `j ^^ 1` does.  I would prefer to add this back into the test suite, but a decision needs to be made about how it *should* behave.

JinShil:
This needs to be changed to a deprecation message, but I'm going to leave it as an error for a while so you all can see all the interesting places where this catches potential mistakes in users' code.

WalterBright:
> The primary benefit of this PR is to catch legitimate errors in users' code.

Yes, like C's `lint` and the many C static checkers. They do find legitimate errors. But they are not part of the language because they have too many false positives.

test/runnable/builtin.d
WalterBright: The question "how far should constant folding be done" is a hopeless one. In this case, no error should have been given.

Also,
```
int j = j;
```
must not be an error, because the `j` on the right is looked up in the enclosing scope - it's not the same `j`.

JinShil: I just tested the following with the current implementation:

test.d
```D
module test;

int t;
```

main.d
```D
import test;

void main()
{
    int j = j;
}
```

Under the current implementation `int j = j` will not generate an error.  The implementation is quite precise.

JinShil: > In this case, no error should have been given.

I'll work on implementing a fix for `j = j ^^ 1`

WalterBright:
> If the current implementation has even 1 false positive, let me know and I'll try to fix it.

It's not fixable - it comes about from legitimate use cases of `x=x;`.

I hate to see you spend more time on this, as it has a very dim chance of being approved. Even if it didn't have the negatives, it is of very low impact. High impact work would be things like getting Phobos to compile with -dip1000.

JinShil:
So be it.  Closing and will mark issue as WONTFIX.

PR #8209, opened 2018-04-25
Support Markdown inline emphasis

dgileadi:
This is the second of several PRs to support Markdown syntax in DDoc.

### DIFF
--- a/src/dmd/doc.d
+++ b/src/dmd/doc.d
@@ -1861,6 +1861,41 @@ extern (C++) const(char)* skipwhitespace(const(char)* p)
 }
 
 /************************************************
+ * Scan past all instances of the given characters.
+ * Params:
+ *  buf =           an OutBuffer containing the DDoc
+ *  i =             the index within `buf` to start scanning from
+ *  chars =         the characters to skip; order is unimportant
+ * Returns: the index after skipping characters.
+ */
+private size_t skipChars(OutBuffer* buf, size_t i, string chars)
+{
+    Outer:
+    foreach (j, c; buf.peekSlice()[i..$])
+    {
+        foreach (d; chars)
+        {
+            if (d == c)
+                continue Outer;
+        }
+        return i + j;
+    }
+    return buf.offset;
+}
+
+unittest {
+    OutBuffer buf;
+    string data = "test ---\r\n\r\nend";
+    buf.write(data.ptr, data.length);
+
+    assert(skipChars(&buf, 0, "-") == 0);
+    assert(skipChars(&buf, 4, "-") == 4);
+    assert(skipChars(&buf, 4, " -") == 8);
+    assert(skipChars(&buf, 8, "\r\n") == 12);
+    assert(skipChars(&buf, 12, "dne") == 15);
+}
+
+/************************************************
  * Scan forward to one of:
  *      start of identifier
  *      beginning of next line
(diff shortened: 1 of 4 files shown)


src/dmd/doc.d
```diff
@@ -1860,6 +1860,34 @@ extern (C++) const(char)* skipwhitespace(const(char)* p)
...
+            if (slice[i] == c)
+            {
+                contains = true;
+                break;
+            }
+        }
+        if (!contains)
+            break;
+    }
+    return i;
```
WalterBright: Just for fun:
```
Outer:
foreach (j, c; buf.peekSlice()[i .. length])
{
    foreach (d; chars)
    {
        if (d == c)
              continue Outer;
    }
    return i + j;
}
return buf.offset;

Comment (WalterBright) [reply src/dmd/doc.d]: Eliminating state flags (i.e. `contains`) using control flow generally makes the code easier to understand.

File: src/dmd/doc.d [hunk 2]
```diff
@@ -1964,6 +1992,102 @@ Lno:
     return i;
 }
 
+/****************************************************
+ * Process Markdown emphasis
```
WalterBright: Add brief explanation of what "markdown emphasis" is.

src/dmd/doc.d
```diff
@@ -1964,6 +1992,102 @@ Lno:
...
+ * Params:
+ *  buf =               an OutputBuffer containing the DDoc
+ *  i =                 the index within `buf` to end the Markdown heading at. Its value changes if any emphasis was replaced.
```
WalterBright: Its value changes to what?

src/dmd/doc.d
```diff
@@ -1964,6 +1992,102 @@ Lno:
     return i;
 }
 
+/****************************************************
+ * Process Markdown emphasis
+ * Params:
+ *  buf =               an OutputBuffer containing the DDoc
```
WalterBright: OutBuffer, not OutputBuffer

src/dmd/doc.d
dgileadi: Nice—I had no idea a label could work in front of a foreach loop like that. Thanks!

src/dmd/doc.d
```diff
@@ -1964,6 +1992,102 @@ Lno:
...
+
+        if (!start.count)
+            start.type = 0;
+        if (!end.count)
+            end.type = 0;
+
+        if (global.params.vmarkdown)
+        {
+            buf.data[iEnd] = '\0';
+            char *s = cast(char*)(buf.data + iStart + count);
```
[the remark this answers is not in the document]
dgileadi: You can't delete line 2026 without messing up the output, because `message` takes a C-style zero-terminated string. Luckily `buf.data[iEnd]` gets overwritten by the emphasis replacements so setting it to `\0` here doesn't hurt anything.

dgileadi:
Here's my suggested documentation for this feature, to be added to ddoc.dd when all the Markdown features are ready to be pulled into master:

First, the existing **Emphasis** section would be renamed **Identifier Emphasis**. Second, the following section would be added:

### Text Emphasis

A span of text wrapped in asterisks (`*`) is emphasized, and text wrapped in two asterisks (`**`) is strongly emphasized:

`*single asterisks*` is rendered as *single asterisks*.

`**double asterisks**` is rendered as **double asterisks**.

You can insert a literal asterisk by [backslash-escaping](https://dlang.org/spec/ddoc.html#punctuation_escapes) it: `\*` is rendered as \*.

Unlike [Markdown](https://daringfireball.net/projects/markdown/syntax), underscores (`_`) are not supported for emphasizing text because it would break snake_case names and underscore prefix processing in [identifier emphasis](https://dlang.org/spec/ddoc.html#emphasis).

src/dmd/doc.d
```diff
@@ -1964,6 +1999,100 @@ Lno:
...
+        if (!start.count)
+            start.type = 0;
+        if (!end.count)
+            end.type = 0;
+
+        if (global.params.vmarkdown)
+        {
+            buf.data[iEnd] = '\0';
+            const char *s = &buf.peekString()[iStart + count];
+            message(loc, "Ddoc: emphasized text '%s'", s);
```
wilzbach: You could use dstrings here with "%.*s"

src/dmd/doc.d
```diff
@@ -1964,6 +1999,100 @@ Lno:
...
+            continue;
+        }
+
+        // multiple-of-3 rule
+        if (((inlineDelimiters[start].leftFlanking && inlineDelimiters[start].rightFlanking) ||
+                (inlineDelimiters[end].leftFlanking && inlineDelimiters[end].rightFlanking)) &&
+            (inlineDelimiters[start].count + inlineDelimiters[end].count) % 3 == 0)
+        {
+            --start;
+            continue;
```
wilzbach: uncovered

src/dmd/doc.d
```diff
@@ -2483,8 +2662,16 @@ extern (C++) void highlightText(Scope* sc, Dsymbols* a, Loc loc, OutBuffer* buf,
...
                 break;
-            if (inMacro)
-                --inMacro;
+            if (parenLevel > 0)
+                --parenLevel;
+            else if (macroLevel)
+            {
+                i += replaceMarkdownEmphasis(buf, loc, inlineDelimiters);
+                while (inlineDelimiters.length && inlineDelimiters[$-1].macroLevel >= macroLevel)
+                    --inlineDelimiters.length;
```
wilzbach: uncovered

test/compilable/extra-files/ddoc_markdown_emphasis.html
```diff
@@ -0,0 +1,513 @@
...
+        color: rgba(0, 0, 0, 1);
+      }
+
+      .psymbol {
+        color: rgba(0, 0, 0, 1);
+      }
+
+      .ddoc_member_header .ddoc_header_anchor .code {
+        font-size: 1em;
+      }
```
wilzbach: Maybe it's a good idea to use a raw template or sanitize this away.
Alternatively, maybe use something like https://github.com/dlang/dlang.org/blob/master/verbatim.ddoc which will yield the expanded DDoc macros and should probably be a lot easier to look at for humans.

wilzbach:
The DAutoTest failure looks unrelated. I hope it was only a spurious one and is gone on the next push.

test/compilable/extra-files/ddoc_markdown_emphasis.html
dgileadi: It probably would be a good idea, but I figured the correct thing was to copy what the other ddoc tests do, which is this.

Interestingly some upcoming Markdown features like headers and lists will require changes to the default CSS, and that will require changing all of the existing HTML files for compilable tests to match. That should be... interesting...

For now I think it's best to not go in a different direction with my HTML test files than the rest of the ddoc tests do, even if it duplicates all that boilerplate ugliness.

wilzbach: > That should be... interesting...

Check out the new AUTO_UPDATE=1 feature. It allows you to automatically update all files depending on the generated files.

src/dmd/doc.d
dgileadi: Added a test case for that (the "more tricky unspaced" one).

dgileadi:
Could the build failures be due to the markdown branch becoming out of date from master?

jacob-carlborg:
The failure doesn't look related to this PR.

dgileadi:
Is there anything else I need to do to get this merged?

jacob-carlborg (approved):

RazvanN7 (approved):

RazvanN7:
The AppVeyor seems unrelated. @dgileadi Can you push again your latest commit to restart AppVeyor?

dgileadi:
Well, that failed hard. Do we need to bring the markdown branch up to date with master?

RazvanN7:
@wilzbach any thoughts on this?

wilzbach:
An attempt at merging master into markdown -> https://github.com/dlang/dmd/pull/8325

dgileadi:
Closing in favor of PRs directly against `master`, starting with #8870.

PR #8211, opened 2018-04-27
Fix issue 18803: when failing to parse a module without errors, don't call semantic3

FeepingCreature:
The existing code is trivially wrong; it constructs with null but checks for `TOK.error`.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -1333,9 +1333,12 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         imp.semanticRun = PASS.semantic;
 
         // Load if not already done so
+        bool loadErrored = false;
         if (!imp.mod)
         {
+            const errors = global.errors;
             imp.load(sc);
+            loadErrored = global.errors != errors;
             if (imp.mod)
                 imp.mod.importAll(null);
         }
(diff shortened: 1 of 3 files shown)


src/dmd/dsymbolsem.d
```diff
@@ -1333,9 +1333,13 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         imp.semanticRun = PASS.semantic;
 
         // Load if not already done so
+        bool load_errored = false;
         if (!imp.mod)
         {
+            uint errors = global.errors;
```
WalterBright: `const errors = ...`

src/dmd/dsymbolsem.d
```diff
@@ -1333,9 +1333,13 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
             imp.load(sc);
+            if (global.errors != errors)
+                load_errored = true;
```
WalterBright:
```
load_errored = global.errors != errors;
```

WalterBright (approved):
Yes, I like this better.

FeepingCreature:
Woo!

src/dmd/dsymbolsem.d
```diff
@@ -1333,9 +1333,12 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         imp.semanticRun = PASS.semantic;
 
         // Load if not already done so
+        bool load_errored = false;
```
jacob-carlborg: Please use `loadErrored` to follow the standard D style guide.

PetarKirov (approved):
Looks good to me.

FeepingCreature:
Ping: SemaphoreCI seems hung.

PR #8213, opened 2018-05-02
fix Issue 17819 - static foreach segfaults on __traits(allMembers)

tgehr:
https://issues.dlang.org/show_bug.cgi?id=17819

This makes `static if` and `static foreach` handle "circular" dependencies the same way as string `mixin`s do it. (I think DMD uses a somewhat arbitrary definition of what constitutes a "circular" dependency.)
It is quite obvious that neither the current `static if` behavior nor the current string `mixin` behavior was consciously designed, so I don't know if this is a good fix. It is however the fix that was asked for.


src/dmd/attrib.d
```diff
@@ -929,6 +928,7 @@ extern (C++) final class StaticIfDeclaration : ConditionalDeclaration
 {
     ScopeDsymbol scopesym;
     bool addisdone;
+    bool onStack = false;
```
JinShil: It took me a little while to understand what "onStack" means, but I see now.

JinShil (approved):

src/dmd/attrib.d
jacob-carlborg: All new public symbols should have documentation.

tgehr: Added a comment.

tgehr: It does not need to be public, but neither does addisdone. Made both private and added comments.

wilzbach:
(rebased to stable as it's a critical bug and should be included in the next point release)

src/dmd/attrib.d
```diff
@@ -47,7 +47,6 @@ extern (C++) abstract class AttribDeclaration : Dsymbol
     {
         if (errors)
             return null;
-
```
RazvanN7: This change is unrelated to this PR

RazvanN7 (approved):

src/dmd/attrib.d
tgehr: Which change?

wilzbach:
How about adding the test from 18718?
https://github.com/wilzbach/dmd/blob/0abcf79eda89cea2409b484b07a84d33affd0a94/test/fail_compilation/test18718.d

tgehr:
@wilzbach: This is already in there.

wilzbach (approved):
Sorry @tgehr I guess I need more sleep :O

src/dmd/attrib.d
```diff
@@ -928,7 +927,8 @@ extern (C++) class ConditionalDeclaration : AttribDeclaration
 extern (C++) final class StaticIfDeclaration : ConditionalDeclaration
 {
     ScopeDsymbol scopesym;
-    bool addisdone;
+    private bool addisdone = false; // stores whether members have been added to scope
+    private bool onStack = false;   // stores whether a call to `include` is currently active
```
WalterBright: s/stores whether/true if/

WalterBright (approved):

src/dmd/attrib.d
tgehr: Fixed.

PR #8214, opened 2018-05-02
fix Issue 18211 - Access violation when generating JSON on static foreach

tgehr:
https://issues.dlang.org/show_bug.cgi?id=18211

This makes `static foreach` behave the same as `static if` with the -X flag.

### DIFF
--- a/src/dmd/attrib.d
+++ b/src/dmd/attrib.d
@@ -1080,7 +1080,10 @@ extern (C++) final class StaticForeachDeclaration : AttribDeclaration
         {
             return cache;
         }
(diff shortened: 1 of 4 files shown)


JinShil (approved):

wilzbach:
FWIW it would be awesome if the following would work:

```d
static foreach (e; ["foo", "bar"])
{
    mixin("/// my doc
    void "~e~"(){}");
}
```

wilzbach:
(rebased to `stable` as it's a critical bug)

RazvanN7 (approved):

PR #8215, opened 2018-05-03
Auto-update ddoc/json diffs

wilzbach:
I'm lazy and there's no reason why `AUTO_UPDATE=1` can't be more powerful.
As for the current `AUTO_UPDATE`, you can always see the changes in your local `git diff` and then decide whether to stage these changes automatically.

Also, of course, the `AUTO_UPDATE=1` needs to be set.

At the moment the shell scripts are quite simplistic and there will only be one f (as this terminates the bash test immediately).


wilzbach:
My use case was https://github.com/dlang/dmd/pull/7836, but essentially any PR author that touches the DDoc or Json files will save time with this.

wilzbach:
FWIW I just needed this on another PR.
For me this feature would be very useful, has no one else ever run into the problem of e.g. needing to update a bunch of generated header files or documentation files?

jacob-carlborg:
> For me this feature would be very useful, has no one else ever run into the problem of e.g. needing to update a bunch of generated header files or documentation files?

If this is what I think it is, I see it as being useful as well. When I changed the default ddoc theme it took a lot of time to update the existing tests.

jacob-carlborg (approved):

wilzbach:
> If this is what I think it is, I see it as being useful as well. When I changed the default ddoc theme it took a lot of time to update the existing tests.

Yep, that's what it's supposed to do now :)

PR #8216, opened 2018-05-04
Stop eagerly generating error message for mismatch arguments

Geod24:
```
This mostly reverts commit 15b1b503440caab6ba8d2879b80d3946d4d5ac23.

The aforementioned commit eagerly allocated the error message during call resolution,
resulting in a lot of garbage being generated for no reason.
Said garbage was never freed, and freeing it seemingly has no effect,
maybe because of DMD's memory allocation pattern.
```

Testing here since the testsuite doesn't pass on my machine ¯\_(ツ)_/¯
Also, targeting stable because I expect large project to suffer from this a lot.

### DIFF
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -2571,8 +2571,8 @@ extern (C++) FuncDeclaration resolveFuncCall(const ref Loc loc, Scope* sc, Dsymb
     Match m;
     m.last = MATCH.nomatch;
 
-    const(char)* failMessage;
-    functionResolve(&m, s, loc, sc, tiargs, tthis, fargs, &failMessage);
+    size_t failIndex;
+    functionResolve(&m, s, loc, sc, tiargs, tthis, fargs, &failIndex);
 
     if (m.last > MATCH.nomatch && m.lastf)
     {
@@ -2699,8 +2699,9 @@ extern (C++) FuncDeclaration resolveFuncCall(const ref Loc loc, Scope* sc, Dsymb
                     .error(loc, "%s `%s%s%s` is not callable using argument types `%s`",
                         fd.kind(), fd.toPrettyChars(), parametersTypeToChars(tf.parameters, tf.varargs),
                         tf.modToChars(), fargsBuf.peekString());
-                    if (failMessage)
-                        errorSupplemental(loc, failMessage);
+                    // If no argument is provided, there's nothing to show
+                    if (fargs)
+                        showArgMismatch(loc, fargs, tf, failIndex);
                 }
             }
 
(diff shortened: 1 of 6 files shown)


wilzbach:
> Testing here since the testsuite doesn't pass on my machine ¯_(ツ)_/¯

Interesting. What error do you get?
Maybe the stack traces without line numbers?
https://github.com/dlang/druntime/pull/2172

Geod24:
It's green :)

ntrel:
This pull degrades error messages significantly. What memory usage do you get with my fix? https://github.com/dlang/dmd/pull/7554#issuecomment-386818232

If my fix is not sufficient, the proper solution would be to use`free` and find out why dmd doesn't free memory properly (it could have an alternative strategy that works better with`free`, enabled with a new compiler switch for intensive projects). This would greatly help a wider scope of projects besides yours.

Geod24:
> This pull degrades error messages significantly.

But the test suite is passing. Which means the original pull request was significantly lacking tests. I haven't had a chance to test the fix you proposed yet, but I'm definitely not found of this approach.

src/dmd/func.d
```diff
@@ -3693,3 +3694,27 @@ extern (C++) final class DeleteDeclaration : FuncDeclaration
...
+{
+    if (failIndex < fargs.dim && failIndex < tf.parameters.dim)
+    {
+        auto arg = (*fargs)[failIndex];
+        auto par = (*tf.parameters)[failIndex];
+        const rv = (!arg.isLvalue() && par.storageClass & (STC.ref_ | STC.out_)) ? "rvalue " : "";
+
+        // disambiguate when toChars() is the same
+        auto at = arg.type.toChars();
+        bool qual = !arg.type.equals(par.type) && strcmp(at, par.type.toChars()) == 0;
```
wilzbach: "strcmp must die."
(I am aware that this is moved existing code)

wilzbach (approved):
Let's get this into stable as the performance regressions seems is critical.
@ntrel please feel free to submit a follow-up with more test cases. I will open a PR to merge this to master too.

wilzbach:
(update the C++ headers)

ntrel:
@wilzbach This pull mucks up my [pull here](https://github.com/dlang/dmd/pull/7584) (which I forgot hadn't been merged). That pull is completely incompatible with just returning an index.

ntrel:
@jinshil I've now submitted https://github.com/dlang/dmd/pull/8232. If I have time I might try to improve the test suite.

@Geod24:
> I haven't had a chance to test the fix you proposed yet, but I'm definitely not found of this approach

My pull fixes the only case where an error could be allocated and not used. Returning an error string is necessary, otherwise we would have to combine and return various other data besides just a parameter index.

src/dmd/func.d
ntrel: Not sure what you're quoting. What should it be replaced with?

src/dmd/func.d
```diff
@@ -3693,3 +3694,27 @@ extern (C++) final class DeleteDeclaration : FuncDeclaration
...
+ *  loc = line information for error message
+ *  fargs = arguments to function
+ *  tf = function called
+ *  failIndex = index of first argument mismatch */
+void showArgMismatch(Loc loc, Expressions* fargs, TypeFunction tf, size_t failIndex)
+{
+    if (failIndex < fargs.dim && failIndex < tf.parameters.dim)
+    {
+        auto arg = (*fargs)[failIndex];
+        auto par = (*tf.parameters)[failIndex];
```
ntrel: `failIndex` is the argument index, not parameter, so for typesafe variadics with more than one argument, no supplemental error will be shown. (This could be worked around perhaps). I'll send a PR to test this.

ntrel:
> This pull degrades error messages significantly

The only problem I've found is with typesafe variadics - PR for a test against this branch: https://github.com/Geod24/dmd/pull/2

ntrel:
BTW #8232 has now been merged. I've also made #8263 (against `master` for now) which removes 2 temporary allocations and adds the missing typesafe variadic test I mentioned.

RazvanN7:
Now that @ntrel 's PRs have been merged. Is this needed anymore?

Geod24:
> Now that @ntrel 's PRs have been merged. Is this needed anymore?

Nope

PR #8217, opened 2018-05-05
Adds a proper initializer + deinitializer process for the lexer

rikkimax:
If dmd is to be used as a library, we need a single central point where all initialization + deinitialization is to occur for the separate components of dmd. This PR does it for the lexer.

It should not break dmd as an executable (see mars.d for switch from module constructor to initializer). This should be temporary, and only one call should be made inside that module, once the initializer process is complete. But right now it is as-is compatibility wise.

### DIFF
--- a/src/dmd/compiler.d
+++ b/src/dmd/compiler.d
@@ -12,7 +12,59 @@
 
 module dmd.compiler;
 
+///
 struct Compiler
 {
-    const(char)* vendor; // Compiler backend name
+    /// Compiler backend name
+    const(char)* vendor;
+}
+
+/**
+ * Initializes the dmd lexer and all required components in order
+ *
+ * It is templated (so D only) to prevent
+ *  errors on local imports to things that may not exist.
+ *
+ * See_Also:
+ *     lexerDeinit
+ */
+void lexerInit()()
+{
+    import dmd.globals : global;
+    import dmd.id : Id;
+    import dmd.tokens : Token;
+    import dmd.identifier : Identifier;
+
+    global._init();
+    Identifier.initTable();
+
+    // Id MUST be initialized after Token
+    Token.initialize();
+    Id.initialize();
+}
+
+/**
+ * Deinitializers the dmd lexer and all required components in order
+ *
+ * It is templated (so D only) to prevent
+ *  errors on local imports to things that may not exist.
+ *
+ * See_Also:
+ *     lexerInit
+ */
+void lexerDeinit()()
+{
+    import dmd.globals : global, Global;
+    import dmd.id : Id;
+    import dmd.tokens : Token;
+    import dmd.identifier : Identifier;
+
+    Id.deinitialize();
+    Token.deinitialize();
+    Identifier.deinitTable();
+
+    // we reset this manually, just to prevent any chance of pinning memory
+    // but this approach does currently leak when !version(GC)
+    // which is quite a problem... but out of scope for now
+    global = Global.init;
 }
--- a/src/dmd/id.d
+++ b/src/dmd/id.d
@@ -39,6 +39,11 @@ struct Id
     {
         mixin(msgtable.generate(&initializer));
     }
+
+    extern(C++) void deinitialize()
+    {
+        mixin(msgtable.generate(&deinitializer));
+    }
 }
 
 private:
@@ -476,4 +481,10 @@ string identifier(Msgtable m)
 string initializer(Msgtable m)
 {
     return m.ident ~ ` = Identifier.idPool("` ~ m.name ~ `");`;
+}
+
+// Used to generate the code for each deinitializer, to prevent pinning.
+string deinitializer(Msgtable m)
+{
+    return m.ident ~ ` = null;`;
 }
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -174,6 +174,11 @@ private int tryMain(size_t argc, const(char)** argv)
     Strings files;
     Strings libmodules;
     global._init();
+
+    // Temporary: removed module constructor (dmd.tokens) which initialized Identifier
+    Identifier.initTable();
+    Token.initialize();
+
     debug
     {
         printf("DMD %s DEBUG\n", global._version);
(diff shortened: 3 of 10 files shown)


rikkimax:
I have had to re-add the module constructor for dmd.tokens, otherwise c++ code being tested won't have the lexer initialized.

rikkimax:
The module constructor in dmd.tokens does not retain any state. It configures dmd.identifier which does get uninitialized.

When used as a library the frontend must be able to be fully uninitialized otherwise we could keep stray modules and pointers around leaking memory or worse corrupting its future state. This includes tokens like keywords.

Right now it remains as a workaround, to keep external code working. It should be removed ASAP.

src/dmd/compiler.d
```diff
@@ -12,7 +12,59 @@
...
+/**
+ * Initializes the dmd lexer and all required components in order
+ *
+ * It is templated (so D only) to prevent
+ *  errors on local imports to things that may not exist.
+ *
+ * See_Also:
+ *     dmd_lexer_deinit
+ */
+void dmd_lexer_init()()
```
wilzbach: Nit: why underscores? [DStyle](https://dlang.org/dstyle.html#naming_conventions) is camelCase

src/dmd/compiler.d
```diff
@@ -12,7 +12,59 @@
...
-    const(char)* vendor; // Compiler backend name
+    /// Compiler backend name
+    const(char)* vendor;
 }
+
+/**
+ * Initializes the dmd lexer and all required components in order
+ *
+ * It is templated (so D only) to prevent
+ *  errors on local imports to things that may not exist.
```
wilzbach: Not sure I understand the motivation. Exception for the private import error this compiles fine.

wilzbach (commented):
Probably a good idea to add a test to test/dub_package?

rikkimax:
@wilzbach I won't disagree with you. Right now I only care about the testsuite and making it pass without modifying unrelated files. It's just silly to modify them halfway when the parser will need this done to it as well. Only then to modify it yet again and probably remove their initialization code completely. A lot of extra work for no benefit.

src/dmd/compiler.d
wilzbach: As mentioned this currently fails to compile because of private access which is why using templates is a bad idea, because it will never be instantiated and tested by the compiler.
Hence, I would strongly prefer not to use templates.
If you have import problems, use `frontend.d`

rikkimax: I am not seeing any compile errors. So if you are getting any, I would like to know.

This will be tested and initiated, but not in this PR. Dmd will call it, as part of its main function. I am currently using it locally and is hooked up into Intellij IDEA.

It is not appropriate to put (de)initialization functions into frontend.d. It is meant for helper functions which (de)initialization is not. The current function in there that does this, exists as a crutch, nothing more.

wilzbach: Hmm, sorry I can't reproduce the build failure anymore. Maybe it was just a stale build cache.

src/dmd/id.d
```diff
@@ -39,6 +39,11 @@ struct Id
     {
         mixin(msgtable.generate(&initializer));
     }
+
+    extern(C++) void deinitialize()
```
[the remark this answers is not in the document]
rikkimax: I think we can drop ``extern(C++)`` for both initialize and deinitialize.
However, it might be best to keep it for now. To verify that it isn't needed.

src/dmd/compiler.d
```diff
@@ -12,7 +12,59 @@
...
+void dmdLexerDeinit()()
+{
+    import dmd.globals : global, Global;
+    import dmd.id : Id;
+    import dmd.tokens : Token;
+    import dmd.identifier : Identifier;
+
+    Id.deinitialize();
+    Token.deinitialize();
+    Identifier.deinitTable();
```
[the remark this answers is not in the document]
rikkimax: For the most part I try to be consistent with the surrounding code, in terms of naming. With a default being initialize/deinitialize if it is needed. I'm just hesitant to go do that sort of refactoring here. It has a stylistic purpose, not a functional one.

Most of the global data probably shouldn't be sitting where it is currently. But at least now, we know where it is.

jacob-carlborg (approved):

wilzbach (changes requested):
This still doesn't come with a test. DMD as a library tests can be put in `test/dub_package`

src/dmd/compiler.d
```diff
@@ -12,7 +12,59 @@
...
+/**
+ * Initializes the dmd lexer and all required components in order
+ *
+ * It is templated (so D only) to prevent
+ *  errors on local imports to things that may not exist.
+ *
+ * See_Also:
+ *     dmdLexerDeinit
+ */
+void dmdLexerInit()()
```
JinShil: This function is already in a package called `dmd`.  Is it necessary to have the `dmd` prefix?

jacob-carlborg:

jacob-carlborg:
> Rather than going back on ourselves, why not first create two new extern(C++) functions. Compiler::initialize and Compiler::deinitialize, and move all state construction/destruction there. This first simplifies the call site of those already using dmd as a library.

That already exists [1], but only as part of the Dub package. 

[1] https://github.com/dlang/dmd/blob/master/src/dmd/frontend.d

src/dmd/mars.d
```diff
@@ -174,6 +174,11 @@ private int tryMain(size_t argc, const(char)** argv)
     Strings files;
     Strings libmodules;
     global._init();
+
+    // Temporary: removed module constructor (dmd.tokens) which initialized Identifier
+    Identifier.initTable();
+    Token.initialize();
```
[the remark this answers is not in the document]
rikkimax: As Jacob said, it exists, but this isn't really the place to discuss such an addition.

This PR is the first of many attempting to get dmd-fe ready for use as a library while keeping primary use case for compiler development.

Right now we need to be concerned with discovering what exists e.g. state, before investigating if a better design would be suitable. I would love to do a huge amount of refactoring, to make it significantly easier to work on, i.e. move lexer to its own directory. But without knowing the boundaries of the parser for state, I could do quite a bit of damage for the future.

Small steps now, better results later :)

PR #8218, opened 2018-05-05
fix issue 7184 - parse error on *(x)++

arBmind:
* reduced parser to the specified grammar

see #7281 and #7442 for reference as well

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -7939,7 +7939,7 @@ final class Parser(AST) : Lexer
                                 error("C style cast illegal, use `%s`", e.toChars());
                             }
                             return e;
-						}
+                        }
                     }
                 }
                 e = parsePrimaryExp();
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -7914,113 +7914,32 @@ final class Parser(AST) : Lexer
                     if (isDeclaration(tk, NeedDeclaratorId.no, TOK.rightParentheses, &tk))
                     {
                         tk = peek(tk); // skip over right parenthesis
-                        switch (tk.value)
-                        {
-                        case TOK.not:
-                            tk = peek(tk);
-                            if (tk.value == TOK.is_ || tk.value == TOK.in_) // !is or !in
-                                break;
-                            goto case;
+                        if (tk.value == TOK.dot) {
+                            // (type) una_exp
+                            nextToken();
+                            auto t = parseType();
+                            check(TOK.rightParentheses);
 
-                        case TOK.dot:
-                        case TOK.plusPlus:
-                        case TOK.minusMinus:
-                        case TOK.delete_:
-                        case TOK.new_:
-                        case TOK.leftParentheses:
-                        case TOK.identifier:
-                        case TOK.this_:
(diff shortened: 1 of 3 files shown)


andralex (approved):
clean code!

src/dmd/parse.d
```diff
@@ -7914,112 +7914,31 @@ final class Parser(AST) : Lexer
...
+                            // if .identifier
+                            // or .identifier!( ... )
+                            if (token.value == TOK.dot)
+                            {
+                                if (peekNext() != TOK.identifier && peekNext() != TOK.new_)
                                 {
-                                    e = parseUnaryExp();
-                                    e = new AST.CastExp(loc, e, t);
-                                    error("C style cast illegal, use `%s`", e.toChars());
+                                    error("identifier or new keyword expected following `(...)`.");
```
JinShil: This does not appear to be covered by any test.  Please add coverage.  You can add a test to the fail_compilation directory to verify compile errors.

src/dmd/parse.d
```diff
@@ -7914,112 +7914,31 @@ final class Parser(AST) : Lexer
...
                                 }
-                                return e;
+                                e = new AST.TypeExp(loc, t);
+                                e = parsePostExp(e);
                             }
-                        default:
-                            break;
+                            else
+                            {
+                                e = parseUnaryExp();
```
JinShil: This block does not appear to be covered by any test.  Please add coverage.  You can add a test to the fail_compilation directory to verify compile errors.

JinShil (changes requested):

src/dmd/parse.d
arBmind: This line comes from line 8015.

arBmind:
To make this clear. I have not added any logic. This patch only removes code. This broke no existing unit test. I added a now test that shows that the wrong behavior was fixed.

If there was logic without a test before there is no test now. These should be added in a different PR.

JinShil:
@arBmind We need a copyright assignment declaration from you.  Please see instructions in the Readme of this repository.

src/dmd/parse.d
JinShil: I added a test for this at https://github.com/dlang/dmd/pull/8256.  It should be pulled before this PR to ensure there isn't any breakage.

arBmind:
@JinShil Sorry this is legally not possible in Germany. The law here is basically "the Copyright is always by the author". I can only give away all kinds of "usage rights". If this is not compatible with this project, feel free to close this PR and use my work (the diff) on your own behalf or ignore it entirely.

If you want to be compatible with every law, take a look at the CLA of Microsoft. https://cla.microsoft.com - The only CLA I was able to sign so far.

WalterBright:
If you prefer, it works if you take the Microsoft CLA, s/Microsoft/D Language Foundation/, sign it and email it to us. Thanks!
(The license in the CLA allows us to make those changes!)

arBmind:
@WalterBright the email was sent.

src/dmd/parse.d
arBmind: This new test is now broken by this PR.

PR #8222, opened 2018-05-06
Fix issue 17968 - Use typeid(Object) instead of typeid(T) to avoid pulling in unused symbols

schveiguy:
The root cause of this problem still isn't fixed, I'll open a new bug report to demonstrate the original issue. This is a workaround, but since it's generated code by the compiler, it shouldn't really affect anything (except allow linking).

Note that using `typeid(Object)` instead of `typeid(T)` on the tuple member is fine as long as it's really an `Object`. `TypeInfo_Class.getHash` *always* casts its argument to `Object` and calls `toHash` on it, no matter what the actual type is.

ping @MartinNowak this is the bug that prevents `iopipe` from using `io`. Thanks to @tgehr, @jacob-carlborg , @thewilsonator, @rainers, @RazvanN7 for helping me figure this out at dconf!

### DIFF
--- a/src/dmd/clone.d
+++ b/src/dmd/clone.d
@@ -765,6 +765,7 @@ extern (C++) FuncDeclaration buildXtoHash(StructDeclaration sd, Scope* sc)
     const(char)* code =
         "size_t h = 0;" ~
         "foreach (i, T; typeof(p.tupleof))" ~
+        // workaround https://issues.dlang.org/show_bug.cgi?id=17968
         "    static if(is(T* : const(Object)*)) " ~
         "        h = h * 33 + typeid(const(Object)).getHash(cast(const void*)&p.tupleof[i]);" ~
         "    else " ~
--- a/src/dmd/clone.d
+++ b/src/dmd/clone.d
@@ -765,7 +765,10 @@ extern (C++) FuncDeclaration buildXtoHash(StructDeclaration sd, Scope* sc)
     const(char)* code =
         "size_t h = 0;" ~
         "foreach (i, T; typeof(p.tupleof))" ~
-        "    h = h * 33 + typeid(T).getHash(cast(const void*)&p.tupleof[i]);" ~
+        "    static if(is(T* : const(Object)*)) " ~
+        "        h = h * 33 + typeid(const(Object)).getHash(cast(const void*)&p.tupleof[i]);" ~
+        "    else " ~
+        "        h = h * 33 + typeid(T).getHash(cast(const void*)&p.tupleof[i]);" ~
         "return h;";
     fop.fbody = new CompileStatement(loc, new StringExp(loc, cast(char*)code));
     Scope* sc2 = sc.push();
--- a/test/runnable/extra-files/test17968.d
+++ b/test/runnable/extra-files/test17968.d
@@ -1,0 +1,12 @@
+import test17968a;
+
+void main()
+{
+    auto r = fun2.fun1;
+    // just check that getHash works (doesn't throw).
+    typeid(r).getHash(&r);
+
+    // try gethash when member is null.
+    r.t = null;
+    typeid(r).getHash(&r);
+}
--- a/test/runnable/extra-files/test17968a.d
+++ b/test/runnable/extra-files/test17968a.d
@@ -1,0 +1,16 @@
+struct S(T)
+{
+    T t;
+}
+
+class C(T) { }
+
+auto fun1(T)(T t)
+{
+    return S!T(t);
+}
+
+auto fun2()
+{
+    return new C!int;
+}
--- a/test/runnable/test17968.sh
+++ b/test/runnable/test17968.sh
@@ -1,0 +1,9 @@
+#!/usr/bin/env bash
+
+
+$DMD -m${MODEL} -I${EXTRA_FILES} -of${OUTPUT_BASE}a${OBJ} -c ${EXTRA_FILES}${SEP}test17968a.d
+$DMD -m${MODEL} -I${EXTRA_FILES} -of${OUTPUT_BASE}${EXE} ${EXTRA_FILES}${SEP}test17968.d ${OUTPUT_BASE}a${OBJ}
+
+${OUTPUT_BASE}${EXE}
+
+rm ${OUTPUT_BASE}{a${OBJ},${EXE}}


jacob-carlborg (commented):
@wilzbach is there a wrapper script that takes care of exiting early on failure?

jacob-carlborg:
@wilzbach I’m talking about `set -e`. Is that still required in the scripts? You were working on something similar if I recall correctly.

wilzbach:
Yep, `set -euo pipefail` is set by default.

schveiguy:
FYI I discovered some more details about this bug on the plane. Will update

schveiguy:
OK, so something I didn't realize originally, and it explains a lot of the weird behavior we were seeing: the members are all considered `const`.

This is why `is(T : Object)` did not work. It also explains the bizarre extra TypeInfo symbol. We never demangled it fully, it's actually the `Typeinfo_Const` for the object! I have a super-reduced test case for that one now, will add the bug report soon.

The updated fix I think should work and is still better than the original (even if we fix the other bug), as it doesn't add the extra Typeinfo unnecessarily. At least this proves that my test case I added is properly being run ;)

schveiguy:
woohoo, it's starting to pass now.

schveiguy:
New bug report: https://issues.dlang.org/show_bug.cgi?id=18833

jacob-carlborg (approved):

JinShil:
Forgive my ignorance on this, but I'm having a hard time understanding the fundamental problem in this issue.  Why is `TypeInfo` needed for this?  I'm currently undertaking an effort to replace runtime calls that require `TypeInfo` with templates, so `TypeInfo` no longer be required for many things.  Would that help resolve this issue?

JinShil:
Please look into the Darwin failure.  If it's spurious just create a new git hash for this PR with a new push.  It may need some babysitting to ensure it stays green until its reviewed and merged.

JinShil:
If this does not fix the root cause, what does it fix?  Is it just a temporary workaround to be backed out when the root cause is fixed.  I'm having trouble understanding this PR.  What specifically is the problem, not the symptom, but the problem?

schveiguy:
> Please look into the Darwin failure. If it's spurious

It was spurious. It's a test to see if the GC throws an InvalidMemoryOperationError when you allocate during a destructor. No way that was affected ONLY on osx 32-bit.

> just create a new git hash for this PR

Note: as a member of druntime, you can click on the test, and click on the "deprecate" link, and the tester will retest the PR. You need to click on the login link at the top of the autotester screen.

I did that for this test.

schveiguy:
> If this does not fix the root cause, what does it fix?

The root cause is outlined more clearly in the new bug report: https://issues.dlang.org/show_bug.cgi?id=18833

> Is it just a temporary workaround to be backed out when the root cause is fixed?

I would say no, because using `typeid(const(Object))` generates less TypeInfo requirements, and provides the same exact call to `obj.toHash` that the original did.

> What specifically is the problem, not the symptom, but the problem?

The problem is that the compiler requires the TypeInfo of `const(SomeClass)` in order to generate the `xtoHash` function for a struct that contains that class, and the compiler fails to generate the TypeInfo in certain circumstances. This is all outlined in the new bug report, please ask more questions there if you still don't grok it.

schveiguy:
FYI, I was working personally with @jacob-carlborg on this at dconf, so there is a lot he knows that is not in this PR conversation.

jacob-carlborg:
As far as I understand it's kind of a workaround. In the sense that if the proper typeinfo was generated there wouldn't be a problem. But I wouldn't say that the workaround is any more wrong than the existing code, only slightly more complicated.

src/dmd/clone.d
```diff
@@ -765,7 +765,10 @@ extern (C++) FuncDeclaration buildXtoHash(StructDeclaration sd, Scope* sc)
     const(char)* code =
         "size_t h = 0;" ~
         "foreach (i, T; typeof(p.tupleof))" ~
-        "    h = h * 33 + typeid(T).getHash(cast(const void*)&p.tupleof[i]);" ~
+        "    static if(is(T : const(Object))) " ~
```
n8sh: Would `is(T == class) || is(T == interface)` be better? It would avoid false positives for non-objects that implicitly convert to objects, such as:

https://run.dlang.io/is/xlZFQA
```d
struct S
{
    size_t toHash() const { return 3; }
    Object what;
    alias what this;
}

void main()
{
	S s;
    static assert(is(S : const(Object)));
    static assert(!is(S == class));
    assert(typeid(S).getHash(&s) == 3);
}
```

n8sh:
When a field is an Object can't we write `h = h * 33 + (p.tupleof[i] is null ? 0 : p.tupleof[i].toHash())` and avoid `TypeInfo` altogether? Is there a benefit to doing that betterC-wise?

schveiguy:
`xtoHash` I believe is only for the struct's TypeInfo anyway. I'm not sure exactly what happens on betterC when you don't specify toHash. AA's certainly don't work, so that's a moot point.

src/dmd/clone.d
schveiguy: OK, what about this:

```D
is(T* : const(Object)*)
```

I tested it against types with alias this, and it seems to work.

jacob-carlborg: `__traits(getLinkage)` works on classes and interfaces now https://github.com/dlang/dmd/commit/971777de5a19b960c343265efca3f9833ea071bc.

schveiguy:
> avoid TypeInfo altogether?

Note that we would have to cast away const to do that (as Object.toHash is not const), and it is subject to alias this trickery as well.

It is a good idea, though...

jacob-carlborg:
>> When a field is an Object can't we write h = h * 33 + (p.tupleof[i] is null ? 0 : p.tupleof[i].toHash()) and avoid TypeInfo altogether? Is there a benefit to doing that betterC-wise?

> Note that we would have to cast away const to do that (as Object.toHash is not const)

This is basically what `TypeInfo_Class.getHash` is doing. Casting away `const`, null check and calling `toHash`.

schveiguy:
> This is basically what TypeInfo_Class.getHash is doing.

Not exactly -- it's casting it FROM const(void)* to Object directly. Then calling it. I think I'd have to go through the same set of casts in order to get it to work.

However, one thing I'd rather not do is put more complication into this generated function -- nobody sees the code anywhere. I'd rather have the breaking of const guarantees inside the runtime.

schveiguy:
Alternatively, we can find a way to fix the root cause and avoid all this BS :)

src/dmd/clone.d
jacob-carlborg: Can you please add the test case from n8sh.

schveiguy: @jacob-carlborg ok

n8sh:
By "fix the root cause" do you mean avoid using `TypeInfo` as @JinShil in https://github.com/dlang/dmd/pull/7799#issuecomment-365048319:

>That is not actually a problem with this PR, but a general problem with D's ubiquitous use of runtime type information (i.e. `TypeInfo`) for things that are known at compile-time. As I mentioned previously, I intend to champion an effort replace such uses of `TypeInfo` with templates, which should actually eliminate all such errors, and produce messages that are more meaningful.

schveiguy:
@JinShil I think it's a good idea, but I don't want to make this PR into that change.

I'm going to add the @n8sh's case as a test case, but don't have time this moment.

schveiguy:
@n8sh That's not what I meant, I mean fix issue 18833 -- the compiler should be generating typeinfo that is being used.

src/dmd/clone.d
```diff
@@ -765,7 +765,10 @@ extern (C++) FuncDeclaration buildXtoHash(StructDeclaration sd, Scope* sc)
     const(char)* code =
         "size_t h = 0;" ~
         "foreach (i, T; typeof(p.tupleof))" ~
-        "    h = h * 33 + typeid(T).getHash(cast(const void*)&p.tupleof[i]);" ~
+        "    static if(is(T* : const(Object)*)) " ~
+        "        h = h * 33 + typeid(const(Object)).getHash(cast(const void*)&p.tupleof[i]);" ~
+        "    else " ~
+        "        h = h * 33 + typeid(T).getHash(cast(const void*)&p.tupleof[i]);" ~
```
MartinNowak: OT: We should look at whether Knuth hashing is actually a sane choice here.
[Multiplicative hashing](https://en.wikipedia.org/wiki/Hash_function#Multiplicative_hashing)

src/dmd/clone.d
```diff
@@ -765,7 +765,10 @@ extern (C++) FuncDeclaration buildXtoHash(StructDeclaration sd, Scope* sc)
     const(char)* code =
         "size_t h = 0;" ~
         "foreach (i, T; typeof(p.tupleof))" ~
-        "    h = h * 33 + typeid(T).getHash(cast(const void*)&p.tupleof[i]);" ~
+        "    static if(is(T* : const(Object)*)) " ~
```
MartinNowak: The need for comparing pointers is slightly subtle.

MartinNowak (approved):
Let's try this, it's a weird workaround though.

src/dmd/clone.d
jacob-carlborg: Checking for a class and using `__traits(getLinkage)` to exclude C++ and Objective-C classes might have been more obvious.

src/dmd/clone.d
n8sh: @MartinNowak I share your hope that this will be subject to further examination. I changed to this in April (https://github.com/dlang/dmd/pull/8175) because it was a small improvement that could pass quickly through the approval pipeline, not because it was the final word. In particular I'd like if we could use compile-time information about the types and order of the fields to make better choices.

src/dmd/clone.d
schveiguy: Really, this is never seen by anyone. So I don't know that it matters what technique is used. I'm hoping that at some point instead of adjusting this, we simply move the function generation into an object.d template.

I'm also hoping that the underlying issue is fixed, and this workaround becomes moot.

PR #8224, opened 2018-05-06
getLinkage on class

TurkeyMan:
Fix issue [17972 __traits(getLinkage) doesn't work for classes](https://issues.dlang.org/show_bug.cgi?id=17972)

Support `__traits(getLinkage)` on classes.
We have no way to detect if a class is `extern(C++)` or not.

And with this, I can fix this: [14536 - Calling destroy() on a on an extern(C++) class causes a segfault](https://issues.dlang.org/show_bug.cgi?id=14536)

Doco: dlang/dlang.org#2361

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -1284,12 +1284,26 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         {
             auto s = getDsymbol(o);
             Declaration d;
-            if (!s || (d = s.isDeclaration()) is null)
+            ClassDeclaration c;
+            if (!s || ((d = s.isDeclaration()) is null && (c = s.isClassDeclaration()) is null))
             {
                 e.error("argument to `__traits(getLinkage, %s)` is not a declaration", o.toChars());
                 return new ErrorExp();
             }
-            link = d.linkage;
+            if (d !is null)
+                link = d.linkage;
+            else final switch (c.classKind)
+            {
+                case ClassKind.d:
+                    link = LINK.d;
+                    break;
+                case ClassKind.cpp:
+                    link = LINK.cpp;
+                    break;
+                case ClassKind.objc:
+                    link = LINK.objc;
+                    break;
+            }
         }
         auto linkage = linkageToChars(link);
         auto se = new StringExp(e.loc, cast(char*)linkage);
--- a/test/compilable/test17419.d
+++ b/test/compilable/test17419.d
@@ -35,3 +35,19 @@ void bar()
     void nested() { }
     static assert(__traits(getLinkage, typeof(&nested)) == "D");
 }
+
+class FooD {}
+interface FooDInterface {}
+extern (C++) class FooCpp {}
+extern (C++) interface FooCppInterface {}
+
+static assert(__traits(getLinkage, FooD) == "D");
+static assert(__traits(getLinkage, FooDInterface) == "D");
+static assert(__traits(getLinkage, FooCpp) == "C++");
+static assert(__traits(getLinkage, FooCppInterface) == "C++");
+
+version (D_ObjectiveC)
+{
+    extern (Objective-C) interface FooObjC {}
+    static assert(__traits(getLinkage, FooObjC) == "Objective-C");
+}
--- a/changelog/getlinkage_for_classes.dd
+++ b/changelog/getlinkage_for_classes.dd
@@ -1,0 +1,12 @@
+__traits(getLinkage, ...) now works on classes and interfaces.
+
+It is now possible to detect the language ABI specified for a class or interface.
+
+---
+class MyClass {}
+extern (C++) class MyCPPClass {}
+extern (C++) interface MyCPPInterface {}
+static assert(__traits(getLinkage, MyClass) == "D");
+static assert(__traits(getLinkage, MyCPPClass) == "C++");
+static assert(__traits(getLinkage, MyCPPInterface) == "C++");
+---


TurkeyMan:
Dunno what to do about this spec thing...

JinShil:
> Dunno what to do about this spec thing...

Update this page https://dlang.org/spec/traits.html.  In other words, submit a PR to https://github.com/dlang/dlang.org/blob/master/spec/traits.dd adding sufficient detail about this feature.

TurkeyMan:
Doco: https://github.com/dlang/dlang.org/pull/2361

jacob-carlborg (changes requested):
I think this will cover interfaces as well. If that’s the case, it needs to be reflected in the tests, spec and changelog.

jacob-carlborg:
If the commit message starts with “Fix issue XXX” where XXX is the bugzilla issue number, dlang-bot will be able to link the PR and the issue.

src/dmd/traits.d
```diff
@@ -1284,12 +1284,26 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+            else final switch (cd.classKind)
+            {
+                case ClassKind.d:
+                    link = LINK.d;
+                    break;
+                case ClassKind.cpp:
+                    link = LINK.cpp;
+                    break;
+                case ClassKind.objc:
+                    link = LINK.objc;
```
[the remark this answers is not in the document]
TurkeyMan: It does not cover interfaces.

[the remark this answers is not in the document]
TurkeyMan: That would be a further development, and I'm not even sure it makes sense. I'd rather not address that here.

TurkeyMan:
Okay. Although I didn't initially seek to 'fix that issue', that was just incidental...

src/dmd/traits.d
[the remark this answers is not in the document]
TurkeyMan: I could look into that as a follow up if it makes sense... but I'd rather this stand alone.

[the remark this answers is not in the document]
TurkeyMan: Dun it.

jacob-carlborg:
> Okay. Although I didn't initially seek to 'fix that issue', that was just incidental...

If the issue and PR are linked the issue will automatically be closed when the PR is merged.

TurkeyMan:
Okay. I'll remember that in future.
Updated the comment.

jacob-carlborg (approved):

jacob-carlborg:
Hmm, there are COM interfaces as well [1]. Not sure if that should be part of this trait.

[1] https://github.com/dlang/dmd/blob/f94107e121bf6f76983d14dfa1528ab637b5b632/src/dmd/dclass.d#L228

TurkeyMan:
Can I work on that in a follow-up?

WalterBright:
> Your PR doesn't reference any Bugzilla issue.

It will if you entitle it in the form "Fix Issue NNNN ..." where "..." is the title of the bugzilla issue. Please use this format, as the tools rely on it.

test/compilable/test17419.d
```diff
@@ -35,3 +35,19 @@ void bar()
...
+
+class FooD {}
+interface FooDInterface {}
+extern (C++) class FooCpp {}
+extern (C++) interface FooCppInterface {}
+
+static assert(__traits(getLinkage, FooD) == "D");
+static assert(__traits(getLinkage, FooDInterface) == "D");
+static assert(__traits(getLinkage, FooCpp) == "C++");
+static assert(__traits(getLinkage, FooCppInterface) == "C++");
```
WalterBright: Please test all the linkages: https://dlang.org/spec/attribute.html#linkage

src/dmd/traits.d
```diff
@@ -1284,12 +1284,26 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         {
             auto s = getDsymbol(o);
             Declaration d;
-            if (!s || (d = s.isDeclaration()) is null)
+            ClassDeclaration c;
+            if (!s || ((d = s.isDeclaration()) is null && (c = s.isClassDeclaration()) is null))
             {
                 e.error("argument to `__traits(getLinkage, %s)` is not a declaration", o.toChars());
                 return new ErrorExp();
```
WalterBright: Since linkage applies to structs, functions, and variables (not just classes), it makes sense to include them here, or there'll just be another inevitable bug report.

WalterBright:
BTW, does this actually fix the issue? If not, it needs another issue as an "enhancement request" that is specific to this PR.

src/dmd/traits.d
TurkeyMan: Can I do that as follow-up? Surely moving towards the target is better than not moving towards the target?

TurkeyMan: I need to do investigation to understand the struct case, since they don't have vtables and stuff.

test/compilable/test17419.d
TurkeyMan: I tested all the linkages that are supported... what do you want me to test?

TurkeyMan:
I believe it fixes the issue... At least, the title of the issue, Jacob's use case (ObjC), and mine (C++)?

src/dmd/traits.d
TurkeyMan: `struct`'s don't appear to have a linkage the same way. Let me address that in a follow-up if it needs to be addressed (I'm not so sure)?

jacob-carlborg:
> Can I work on that in a follow-up?

I’m fine with that. I’ve already given my approval.

JinShil:
@jacob-carlborg The ObjectiveC branch isn't covered by the tests.  Is that expected, or does something need to be done to run that branch on our CIs?

jacob-carlborg:
> The ObjectiveC branch isn't covered by the tests. Is that expected, or does something need to be done to run that branch on our CIs?

@JinShil the test is guarded by `version (D_ObjectiveC)` since Objective-C interfaces only work on macOS. The tests should run on macOS but I guess the code coverage tool doesn't run on macOS.

test/compilable/test17419.d
```diff
@@ -35,3 +35,19 @@ void bar()
...
+
+version (D_ObjectiveC)
```
JinShil: Is this guard necessary?  Should this work regardless of whether we're compiling for ObjectiveC or not?

jacob-carlborg: Yes it's necessary, this is what happens when compiling an Objective-C interface on Linux: `Error: interface foo.Foo Objective-C interfaces not supported`.

JinShil (approved):
LGTM, but it is incomplete; the bug report specifically mentions structs as well.  I'll leave it to others to decide whether to demand those be done now or as a followup.

src/dmd/traits.d
jacob-carlborg: https://github.com/dlang/dmd/blob/1ea9308b9eeb268a264c559c11dda1b0064ede6b/src/dmd/declaration.d#L281

jacob-carlborg: @TurkeyMan do you want to handle structs in this PR or in a follow-up?

TurkeyMan:
I don't actually think struct's are a related development, if this work is even applicable to struct's at all. The linkage affects the ABI of classes and interfaces. struct's have no such ABI that's affected by linkage.

If someone can show me how it makes sense for struct's, I can do it in a follow-up.

PR #8226, opened 2018-05-07
Promote run.d as the default testsuite runner in README.md

wilzbach:
As we eventually want to move over to `run.d` and avoid needing to maintain both systems for a long time, it's probably a good idea to only mention `run.d` in the test README.md instructions.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -10,26 +10,25 @@ Quick guide
 ### Run all tests
 
 ```sh
-make -j10
+./run.d
 ```
 
 Note:
 
-- the exact parallelism parameter `-j` depends on your system,
-but if you have enough memory, a high `-j` allows a significant speed-up.
+- `run.d` will automatically use all available threads. Use e.g. `-j4` if you need a lower parallelism
 - all commands below assume that you are in the `test` directory
 
 ### Run only a specific subset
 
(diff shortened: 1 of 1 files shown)


test/README.md
```diff
@@ -10,26 +10,25 @@ Quick guide
...
-but if you have enough memory, a high `-j` allows a significant speed-up.
+- `run.d` will automatically use all available threads. Use e.g. `-j4` if you need a lower parallelism
 - all commands below assume that you are in the `test` directory
 
 ### Run only a specific subset
 
 ```sh
-make -j10 run_fail_compilation_tests
-make -j10 run_compilable_tests
+./run.d fail
```
wilzbach: Could be `fail_compilation` too, but I prefer the short alias

test/README.md
```diff
@@ -97,12 +97,11 @@ TEST_OUTPUT:
 Hello, World!
 ---
 */
-import std.stdio;
```
wilzbach: Using Phobos in the testsuite is more or less an anti-pattern, so imho it's a good idea if the example doesn't use Phobos.

JinShil (approved):

jacob-carlborg (approved):

PR #8228, opened 2018-05-08
Add indexOf to root.array

wilzbach:
For use in e.g. https://github.com/dlang/dmd/pull/8189#discussion_r182639975 or https://github.com/dlang/dmd/pull/8201/files#r183921790

### DIFF
--- a/src/dmd/root/array.d
+++ b/src/dmd/root/array.d
@@ -146,6 +146,35 @@ public:
         dim++;
     }
 
+    /**
+    Search the array for an element and returns its index.
+    Params:
+        el = element to search for
+
+    Returns: index of the element, -1 otherwise
+    */
+    ptrdiff_t indexOf(const T el) const nothrow pure @nogc
+    {
+        foreach (i, const v; this)
+        {
+            if (v == el)
+                return i;
+        }
+        return -1;
+    }
+
+    ///
+    unittest
+    {
+        Array!int array;
+        array.push(1);
+        assert(array.indexOf(2) == -1);
+        array.push(2);
+        assert(array.indexOf(2) == 1);
+        array.push(3);
+        assert(array.indexOf(2) == 1);
+    }
+
     void setDim(size_t newdim) nothrow
     {
         if (dim < newdim)
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -2,7 +2,7 @@
 
 set -uexo pipefail
 
-HOST_DMD_VER=2.079.1 # same as in dmd/src/posix.mak
+HOST_DMD_VER=2.080.1 # same as in dmd/src/posix.mak
 CURL_USER_AGENT="CirleCI $(curl --version | head -n 1)"
 N=4
 CIRCLE_NODE_INDEX=${CIRCLE_NODE_INDEX:-0}


src/dmd/root/array.d
```diff
@@ -146,6 +146,35 @@ public:
...
+    /**
+    Search the array for an element and returns its index.
+    Params:
+        el = element to search for
+
+    Returns: index of the element, -1 otherwise
+    */
+    ptrdiff_t indexOf(T el) nothrow
+    {
+        foreach (i; 0 .. dim)
```
jacob-carlborg: Why not a regular `foreach` instead of a `foreach` range loop?

wilzbach: No reason. Thanks for the reminder!

jacob-carlborg (approved):

src/dmd/root/array.d
```diff
@@ -146,6 +146,35 @@ public:
...
     }
 
+    /**
+    Search the array for an element and returns its index.
+    Params:
+        el = element to search for
+
+    Returns: index of the element, -1 otherwise
+    */
+    ptrdiff_t indexOf(T el) nothrow
```
WalterBright: `el` should be `const`, function should be `@nogc`

src/dmd/root/array.d
```diff
@@ -146,6 +146,35 @@ public:
...
     }
 
+    /**
+    Search the array for an element and returns its index.
+    Params:
+        el = element to search for
+
+    Returns: index of the element, -1 otherwise
+    */
+    ptrdiff_t indexOf(const T el) nothrow @nogc
```
jacob-carlborg: `pure` as well?

JinShil:
Grrrr!  DAutoTest uses a newer compiler version than auto-tester, and it throws the following error:

```
dmd/root/array.d(160): Error: none of the overloads of `opEquals` are callable using argument types `(Statement, const(Statement))`, candidates are:
```

I can't seem to produce this locally, either.

jacob-carlborg:
Perhaps adding `const` to `v` in the `foreach` loop will help.

jacob-carlborg:

JinShil:
> dmd/root/array.d(160): Error: none of the overloads of `opEquals` are callable using argument types `(const(Statement), const(Statement))`, candidates are:
object.d(140):        `object.opEquals(Object lhs, Object rhs)`
object.d(165):        `object.opEquals(const(Object) lhs, const(Object) rhs)`

Great!

wilzbach:
@JinShil looks like with the upgrade to 2.079, this is no longer blocked :)
Are you okay with squashing the commits into one?

JinShil:
@wilzbach, It's actually your PR, so squash away.

wilzbach:
> I wonder if it's a quirk of how DAutoTest chooses the version of Phobos to test with.

I think it's using `master`. I tried another repush.

> Would it help to update markdown-master to the tip of master, if it isn't already there?

This PR already tries to merge master into `markdown`.

CC @CyberShadow any other ideas what could have gone wrong here?

CyberShadow:
Sorry, what do you mean? This PR is targeting branch master, not markdown.

wilzbach:
Sorry this comment was supposed to be for https://github.com/dlang/dmd/pull/8325 - not sure how it ended up here :/

wilzbach:
This is still blocked and I don't need it anymore :/

PR #8230, opened 2018-05-08
Add debug symbols for all non-release builds

wilzbach:
Currently DMD has the following concepts and options:

- BUILD=release|debug (default: release)
- ENABLE_RELEASE=1
- ENABLE_DEBUG=1
- DEBUG=1 (legacy alias)
- RELEASE=1 (legacy alias)

Only with BUILD=debug or (ENABLE_DEBUG=1) the debug build gets triggered.
OTOH the release build only gets activated with ENABLE_RELEASE=1 (or RELEASE=1).

This means that there's this weird intermediate, default state which is neither a debug nor release build.
This PR adds -g to the default state as that's imho a sane default and often very useful.
This doesn't affect release builds.

I have already quickly talked with @MartinNowak about this at DConf and he agreed that this is a sensible thing to do.

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -245,6 +245,9 @@ endif
 ifdef ENABLE_RELEASE
 CXXFLAGS += -O2
 override DFLAGS += -O -release -inline
+else
+# Add debug symbols for all non-release builds
+override DFLAGS += -g
 endif
 ifdef ENABLE_PROFILING
 CXXFLAGS  += -pg -fprofile-arcs -ftest-coverage


jacob-carlborg (approved):

JinShil:
Sorry, but I don't see the need for this tertiary, neither debug nor release state.  Wouldn't it be more prudent to force select either "ENABLE_DEBUG" or "ENABLE_RELEASE" by default?  And why do we need so many different ways to select debug or release builds?

wilzbach:
> Sorry, but I don't see the need for this tertiary, neither debug nor release state. Wouldn't it be more prudent to force select either "ENABLE_DEBUG" or "ENABLE_RELEASE" by default?

I agree that this _current_ intermediate state is weird and I would love for the debug build to become the default.
FWIW I once even tried to make the release build the default (see https://github.com/dlang/dmd/pull/7473), but that's a bit tricky because then the default build time will go up from ~2s to >60s.
Anyhow, I thought that making the debug build the default might be more controversial (and would require a lot of changes to our CI scripts), so I went with this poor solution of at least getting debug symbols by default.

> And why do we need so many different ways to select debug or release builds?

I guess historical baggage?
See e.g. https://github.com/dlang/dmd/pull/8231 for trying to cleanup at least a part of this.

JinShil:
> Anyhow, I thought that making the debug build the default might be more controversial

What's controversial about it?

> and would require a lot of changes to our CI scripts

Good work often requires such drudgery.  If we want debug builds to be the default, let's do muscle through it and get it done.

WalterBright:
The state of the debug selection as shown in the first comment is abominable. However, I'll pull this for the moment.

PR #8232, opened 2018-05-08
Avoid eager error message allocation with `functionResolve`

ntrel:
* Don't set `failMessage` with `functionResolve` until it's needed.

This is a less drastic fix than #8216. In particular, this pull does not cause problems for #7584. The latter pull is a good example of why `functionResolve` cannot just return a parameter index and create the error message on the caller side.

### DIFF
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -2570,9 +2570,8 @@ extern (C++) FuncDeclaration resolveFuncCall(const ref Loc loc, Scope* sc, Dsymb
 
     Match m;
     m.last = MATCH.nomatch;
-
-    const(char)* failMessage;
-    functionResolve(&m, s, loc, sc, tiargs, tthis, fargs, &failMessage);
+    functionResolve(&m, s, loc, sc, tiargs, tthis, fargs, null);
+    auto orig_s = s;
 
     if (m.last > MATCH.nomatch && m.lastf)
     {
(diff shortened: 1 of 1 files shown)


JinShil:
Since this is a critical regression, please target stable instead of master.

JinShil:
@Geod24 Could you please test this locally for your use case and let us know if it resolves the issue you encountered?

src/dmd/func.d
```diff
@@ -2570,9 +2570,8 @@ extern (C++) FuncDeclaration resolveFuncCall(const ref Loc loc, Scope* sc, Dsymb
 
     Match m;
     m.last = MATCH.nomatch;
-
-    const(char)* failMessage;
-    functionResolve(&m, s, loc, sc, tiargs, tthis, fargs, &failMessage);
+    functionResolve(&m, s, loc, sc, tiargs, tthis, fargs, null);
+    auto orig_s = s;
```
RazvanN7: Why the extra reference? You can just use `s`.

RazvanN7 (changes requested):

RazvanN7:
ping @ntrel . What's the status on this?

src/dmd/func.d
ntrel: @RazvanN7 What about line 2609?
https://github.com/dlang/dmd/pull/8232/files/8facce8f4138910a600f3580dba0279666573e21#diff-ca782fb8b1ce12422a29fed42ffe0cadL2609

RazvanN7: Oh, ok missed that

RazvanN7 (approved):

PR #8235, opened 2018-05-09
Fix Issue 18828 - [-betterC] helpless error in object.d

JinShil:
This won't pass tests without https://github.com/dlang/druntime/pull/2178


JinShil:
Test was added to druntime PR.

PR #8237, opened 2018-05-10
Fix Issue 18743 - ConditionalExpression with AssignExpression should require parens

ntrel:
Depends on https://github.com/dlang/phobos/pull/6505.

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -8494,7 +8494,7 @@ final class Parser(AST) : Lexer
             // Change to error in 2019-05.
             // @@@DEPRECATED_2019-05@@@.
             if (e.op == TOK.question && !e.parens && precedence[token.value] == PREC.assign)
-                dmd.errors.deprecation(e.loc, "`%s` must be parenthesized when next to operator `%s`",
+                dmd.errors.deprecation(e.loc, "`%s` must be surrounded by parentheses when next to operator `%s`",
                     e.toChars(), Token.toChars(token.value));
 
         const loc = token.loc;
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -8489,6 +8489,11 @@ final class Parser(AST) : Lexer
     AST.Expression parseAssignExp()
     {
         auto e = parseCondExp();
+            // require parens for e.g. a ? a : a = a;
+            if (precedence[e.op] == PREC.cond && !e.parens && precedence[token.value] == PREC.assign)
+                dmd.errors.deprecation(e.loc, "`%s` must be parenthesized when next to operator `%s`",
+                    e.toChars(), Token.toChars(token.value));
+
         const loc = token.loc;
         switch (token.value)
         {
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -6444,7 +6444,7 @@ final class Parser(AST) : Lexer
     void checkParens(TOK value, AST.Expression e)
     {
         if (precedence[e.op] == PREC.rel && !e.parens)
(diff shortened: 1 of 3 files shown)


src/dmd/parse.d
```diff
@@ -8483,6 +8483,11 @@ final class Parser(AST) : Lexer
         auto e = parseCondExp();
         while (1)
         {
+            // require parens for e.g. a ? a : a = a;
```
WalterBright: This example doesn't seem to be in the test cases.

src/dmd/parse.d
```diff
@@ -8483,6 +8483,11 @@ final class Parser(AST) : Lexer
...
+            if (precedence[e.op] == PREC.cond && !e.parens && precedence[token.value] == PREC.assign)
```
WalterBright: Replace `precedence[e.op] == PREC.cond` with `e.op == TOK.question`

jacob-carlborg:
If this requires a change in Phobos it means it's a breaking change and requires going through the regular deprecation phase.

src/dmd/parse.d
ntrel: Done.

src/dmd/parse.d
ntrel: I've now updated this line to be equivalent to a test case.

ntrel:
@jacob-carlborg Added changelog, deprecation date comment and spec pull: https://github.com/dlang/dlang.org/pull/2366. Also addressed Walter's comments.

src/dmd/parse.d
```diff
@@ -8483,6 +8483,14 @@ final class Parser(AST) : Lexer
         auto e = parseCondExp();
         while (1)
         {
+            // require parens for e.g. `t ? a = 1 : b = 2`
+            // Deprecated in 2018-05.
+            // Change to error in 2019-05.
+            // @@@DEPRECATED_2019-05@@@.
+            if (e.op == TOK.question && !e.parens && precedence[token.value] == PREC.assign)
+                dmd.errors.deprecation(e.loc, "`%s` must be parenthesized when next to operator `%s`",
+                    e.toChars(), Token.toChars(token.value));
```
jacob-carlborg: I suggest that the deprecation message contains a suggestion on how to add the parentheses to get the same behavior as before the deprecation was added. The changelog should reflect this as well.

jacob-carlborg (changes requested):

jacob-carlborg (approved):
LGTM, although I still think it needs approval.

RazvanN7 (approved):

RazvanN7:
ping @andralex @WalterBright. This PR is ready to merge. Are you against this addition?

wilzbach (approved):
FWIW I'm in favor of this. Thanks a lot for making this happen!
@WalterBright or @andralex can we make this happen?

Geod24 (approved):
LGTM

src/dmd/parse.d
JinShil: > assuming the reader knows what parenthesized means

I had to refer to the changelog example to understand what it meant.

jacob-carlborg: Would "surrounded by parentheses" be better?

JinShil: @andralex and @WalterBright aren't showing any interest in this, so I'm ready to move forward without their approval, but could you please consider @jacob-carlborg's suggestion with regard to the error message?

wilzbach: > Edit: Made the change myself.

Ah I think we pushed in the same moment. Anyhow, I force-pushed my variant as the tests need updating too. Sorry.

jacob-carlborg:
We could still include the fix in the error message or as a supplement message.

PR #8238, opened 2018-05-10
Fix more cases of bug 15475

dgileadi:
It turns out that parentheses in Ddoc code spans weren't properly escaped in a couple of other cases than those the original PR addressed. One is when they were in backtick-quoted code spans, and another is when they weren't part of some other D construct. This PR fixes these cases.

### DIFF
--- a/src/dmd/doc.d
+++ b/src/dmd/doc.d
@@ -554,7 +554,8 @@ extern (C++) void escapeDdocString(OutBuffer* buf, size_t start)
 extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start)
 {
     uint par_open = 0;
-    bool inCode = 0;
+    char inCode = 0;
+    bool atLineStart = true;
     for (size_t u = start; u < buf.offset; u++)
     {
         char c = buf.data[u];
@@ -563,6 +564,7 @@ extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start)
         case '(':
             if (!inCode)
                 par_open++;
+            atLineStart = false;
             break;
         case ')':
             if (!inCode)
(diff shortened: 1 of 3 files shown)


src/dmd/doc.d
```diff
@@ -598,7 +598,12 @@ extern (C++) void escapeStrayParenthesis(Loc loc, OutBuffer* buf, size_t start)
                 u++;
             }
             if (numdash >= 3)
-                inCode = !inCode;
+                inCode = inCode == c ? 0 : c;
```
ibuclaw: What is this change in logic doing?

dgileadi: Not what I expected :)

Since there are actually two types of code sections, \`backtick-quoted spans\` and `---` fenced blocks, I meant to change the toggling logic to only toggle `inCode` off if it was the same type of code section as was started. However I did it wrong. I'll make a fix.

RazvanN7:
Does this fix Issue 15475 ?
If (yes)
    please add Fix Issue 15475 in your commit message.
else
    what other cases need to be addressed to fix the issue?

dgileadi:
I don't know whether it fixes all cases—as @quickfur says in the bug report:

> doc.d right now is a fragile tangle of spaghetti code, full of special cases and unexpected gotchas.

It fixed the instances I noticed. I'll add that to the commit message, but I'm not certain it fixes all cases.

RazvanN7:
@dgileadi then add "Fix Issue 15475" so that the bug is closed and if other specific failures are found then other bug reports may be opened.

dgileadi:
I indeed amended the commit to include that message. Does that work?

RazvanN7:
>I indeed amended the commit to include that message. Does that work?

Hmm, what command did you use? It looks like the commit is still called "Fix more cases of bug 15475".

I suggest using : 

```
git commit --amend -am "Fix Issue 15475 - Fix more cases"
git push -f $fork_name $branch_name 
```

dgileadi:
Ah, I put the message on the 2nd line of the commit (just expand it above). Does that work?

RazvanN7:
>Ah, I put the message on the 2nd line of the commit (just expand it above). Does that work?

Apparently not. Please see my previous comment

PR #8141, opened 2018-04-05
Fix Issue 18228 - this(this a){} doesn't generate postblit ctor; this(this){} does

RazvanN7:

PR #8242, opened 2018-05-12
Issue 12228 & Issue 13943 - Deprecate usage of 'this' and 'super' as types

JinShil:
This is also a followup to https://github.com/dlang/dmd/pull/8141

### DIFF
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -1510,7 +1510,7 @@ extern(C++) Expression getProperty(Type t, const ref Loc loc, Identifier ident, 
 
 private extern (C++) final class GetPropertyVisitor : Visitor
 {
-    alias visit = super.visit;
+    alias visit = typeof(super).visit;
     Loc loc;
     Identifier ident;
     int flag;
@@ -2056,7 +2056,7 @@ extern(C++) void resolve(Type mt, const ref Loc loc, Scope* sc, Expression* pe, 
 
 private extern(C++) final class ResolveVisitor : Visitor
 {
-    alias visit = super.visit;
+    alias visit = typeof(super).visit;
     Loc loc;
     Scope* sc;
     Expression* pe;
@@ -2238,6 +2238,16 @@ private extern(C++) final class ResolveVisitor : Visitor
         //printf("TypeIdentifier::resolve(sc = %p, idents = '%s')\n", sc, mt.toChars());
         if ((mt.ident.equals(Id._super) || mt.ident.equals(Id.This)) && !hasThis(sc))
         {
+            // @@@DEPRECATED_v2.086@@@.
+            if (mt.ident.equals(Id._super))
+            {
+                deprecation(mt.loc, "Using `super` as a type is deprecated. Use `typeof(super)` instead");
+            }
+            // @@@DEPRECATED_v2.086@@@.
+            if (mt.ident.equals(Id.This))
+            {
+                deprecation(mt.loc, "Using `this` as a type is deprecated. Use `typeof(this)` instead");
+            }
             AggregateDeclaration ad = sc.getStructClassScope();
             if (ad)
             {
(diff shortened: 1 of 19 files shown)


jacob-carlborg (approved):

JinShil:
@CyberShadow, can you please advise why DAutoTest is failing while others are passing?

CyberShadow:
Could be due to a difference in the DMD version used to compile DMD (DAutoTest is trying to build the DMD source documentation at the point where it fails).

Does the error message make sense to you?

```
dmd/sapply.d(30): Error: typeof(super).visit is used as a type
dmd/nogc.d(32): Error: typeof(super).visit is used as a type
dmd/typesem.d(1513): Error: typeof(super).visit is used as a type
dmd/typesem.d(2059): Error: typeof(super).visit is used as a type
dmd/typesem.d(2562): Error: typeof(super).visit is used as a type
dmd/func.d(71): Error: typeof(super).visit is used as a type
dmd/apply.d(32): Error: typeof(super).visit is used as a type
```

If it's an easy fix, it might as well be done, to avoid unnecessarily restraining the range of versions that DMD can be built with.

JinShil:
> Does the error message make sense to you?

Somewhat.  I did modify those lines, but it appears that the compiler being used to build the documentation has a bug.

>  DAutoTest: `HOST_DC=/home/dtest/DAutoTest/work/dl/dmd-2.067.1/dmd2/linux/bin64/dmd`
>  auto-tester: `HOST_DC=/home/braddr/sandbox/at-client/release-build/dmd-2.079.0/linux/bin64/dmd`

So DAutoTest is using 2.067.1?  Is that necessary?  Can it be upgraded?

src/dmd/typesem.d
```diff
@@ -2238,6 +2238,16 @@ private extern(C++) final class ResolveVisitor : Visitor
         //printf("TypeIdentifier::resolve(sc = %p, idents = '%s')\n", sc, mt.toChars());
         if ((mt.ident.equals(Id._super) || mt.ident.equals(Id.This)) && !hasThis(sc))
         {
+            // @@@DEPRECATED_2019-05@@@.
```
wilzbach: The new deprecation DIP suggests to use versions instead.

wilzbach (approved):

CyberShadow:
Found the issue.

DAutoTest and related software will use 2.079.0 once #8112 is merged (see https://github.com/dlang/dmd/pull/8112#issuecomment-379728815). However, that PR is still not merged, so the old default (2.067.1) remains in force.

There are a few ways to proceed:
- Wait until #8112 is merged
- Work around the compiler bug by adjusting the code in this PR
- Add a second condition for detecting when a newer compiler is needed to ae.sys.d.manager to allow the code as it is in this PR right now to compile in Digger and DAutoTest.

I would obviously prefer one of the first two :)

wilzbach:
> deprecate/remove calling static methods via instances?

AFAICT that's a feature by design. I at least remember Walter defending this a few months ago as it allows for more generic code.

> Or at least opCall

Not sure even this will be easy.
It's still used in Phobos: https://github.com/dlang/phobos/search?utf8=✓&q="static+opcall"&type=

CyberShadow:
Being able to forbid static or non-static calls in a contract would work too.

JinShil:
Will wait on #8112 or alternative.

WalterBright:
> AFAICT that's a feature by design. I at least remember Walter defending this a few months ago as it allows for more generic code.

Yes, it's deliberate. Removing it would require a compelling argument.

CyberShadow:
> Removing it would require a compelling argument.

The reason I asked is that I ran into a problem caused by it when trying to implement reference-counted classes.

The problem occurs when the wrapped class defines `opCall`. In this case, we cannot use `static opCall` in the reference-counted wrapper to wrap the class constructor (and we can't use normal struct constructors because `this()` without parameters is forbidden on structs).

In the end I had to move all construction to factory functions because of this issue.

Do you have a better solution?

JinShil:
DAutoTest has been updated.  This PR is ready to go.

Geod24:
There's a slight problem with this: https://issues.dlang.org/show_bug.cgi?id=18976

PR #8243, opened 2018-05-12
account for 'unsigned long' name mangling in OSX 32

WalterBright:


WalterBright:
ping @braddr

wilzbach:
If this is required for the auto-tester to upgrade to a newer DMD version, targeting stable might help (and tests of course, too).

WalterBright:
Blocking https://github.com/dlang/dmd/pull/8112

JinShil:
I'd be happy to review this, but I need some context.  What's the problem and how does this fix it?

WalterBright:
The problem comes when using dmd 2.079 as the boot compiler on OSX:

https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3161515&isPull=true

```
  "IntegerExp::create(Loc, unsigned long, Type*)", referenced from:
ld: symbol(s) not found for architecture x86_64
```
I have fixes for the unsigned long mangling issue itself, but those aren't in the 2.079 yet. Hence this PR which is a workaround.

schveiguy:
> but those aren't in the 2.079 yet

should we target a later base compiler then?

WalterBright:
> should we target a later base compiler then?

It's so hard getting this moving forward with all the hand-wringing, I mean months and months and months of delays, I do not wish to start over with a new version. Let's just get this done.

wilzbach:
FWIW if this is already fixed in 2.080, then there's nothing stopping Brad from using that one...
(the only guarantee we agreed to give is that the latest release can compile DMD)

Of course I don't object to this PR.

WalterBright:
I'll take that as approval :-)

PR #8244, opened 2018-05-13
reduce global access in s2ir and e2ir

WalterBright:
This makes access to global.params through a const reference and a local.


JinShil:
What problem does this solve?  Are you just trying to reduce coupling?

WalterBright:
The problem is accessing mutable global state through the side. This PR makes the access through the parameter list, and it is const.

It's about encapsulation, and moving to a more functional style.

JinShil (approved):

PR #8245, opened 2018-05-14
Lower array literals to object.__arrayLiteral template

JinShil:
@andralex @WalterBright et al.

This is my attempt at picking up where @somzzz left off, trying to replace runtime hooks that depend on runtime `TypeInfo` with templates.

For now `__arrayLiteral` is just a replacement for `_d_arrayLiteralTX` to verify the compiler/runtime interface is correct.  `__arrayLiteral` needs to be refactored to no longer rely on `TypeInfo`, but I want to get the compler/runtime interface implemented first and passing all tests.  Then I can refactor `__arrayLiteral` to no longer rely on runtime information.

I could use some direction on this.  I'm not sure about naming conventions, where these templates should reside in the source tree, etc...

This requires https://github.com/dlang/druntime/pull/2181

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -7301,6 +7301,24 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
 
             // Remains valid array assignments
             //  d = d[], d = [1,2,3], etc
+
+            if (exp.e2.op == TOK.arrayLiteral && !exp.e1.type.isImmutable() && exp.e2.type.ty != Tsarray && !(sc.flags & SCOPE.ctfe))
+            {
+                ArrayLiteralExp ale = cast(ArrayLiteralExp)exp.e2;
+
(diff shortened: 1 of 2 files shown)


schveiguy:
Why is this passing the auto tester when the required druntime call hasn't been merged yet?

UplinkCoder:
Please don't do this without thinking about the impact.

JinShil:
> Why is this passing the auto tester when the required druntime call hasn't been merged yet?

Because I made a mistake.  It's fixed now (at least to the best of my ability).

> Please think about the impact of this change.

🤔 🤔 🤔 🤔 🤔   Yep, it's a good idea.

Seriously, @UplinkCoder, your comment is most unhelpful.  If you have concerns, make them explicitly known.

JinShil:
Gah! I screwed up the first commit creating a false positive.  This PR is much too naive. I need to spend more time on it.  Don't waste your time looking at it.

PR #8247, opened 2018-05-14
Fix Issue 18859 - Silence class allocator/deallocator deprecation war…

CyberShadow:
…ning if they are marked "deprecated"

What do you think?


wilzbach (approved):
Makes sense to me!

CyberShadow:
Aha, looks like conveniently there was already something for this in the test suite, so no new test cases are needed - just update the existing ones.

Geod24:
Another instance of https://issues.dlang.org/show_bug.cgi?id=17906#c1
We should really do something about it, with the current state, it's very easy to forget this check.

PR #8248, opened 2018-05-15
adjust d_size_t for OSX 64 bit

WalterBright:
Because of:
```
```
https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3165089&isPull=true


WalterBright:
Abandoning this. Fixing size_t on OSX is whack-a-mole absurdity. Replacing with https://github.com/dlang/dmd/pull/8249

PR #8250, opened 2018-05-16
Add support for IRETQ instruction in inline assembler

leitimmel:
DMD currently supports `IRET` and `IRETD`, with `IRETQ` missing for no apparent reason. Adding it allows writing interrupt service routines for AMD64-compatible CPUs with nothing more than D and it's inline assembler.

### DIFF
--- a/changelog/iretq.dd
+++ b/changelog/iretq.dd
@@ -1,0 +1,17 @@
+`IRETQ` is now supported in inline assembler.
+
+`IRETQ` is the 64-bit version of the already supported `IRET` instruction. With
+its inclusion, it is now possible to write 64-bit interrupt service routines
+in pure D.
+
+---
+void isr()
+{
+    asm
+    {
+        naked;
+        // ...
+        iretq;
+    }
+}
+---
--- a/test/runnable/iasm64.d
+++ b/test/runnable/iasm64.d
@@ -3605,6 +3605,10 @@ void test50()
         0xCD, 0x67,     // int  067h
 //      0xCE,           // into
         0x66, 0xCF,     // iret
+        0x48, 0xCF,     // iretq
+        0x90, 0x90,     // nop;nop - put instructions above this or L10 changes
+                        //           address, which changes all the jump
+                        //           instructions and breaks the test case.
         0x77, 0xFC,     // ja   L30
         0x77, 0xFA,     // ja   L30
         0x73, 0xF8,     // jae  L30
@@ -3727,7 +3731,11 @@ void test50()
         int     3       ;
         int     0x67    ;
         //into          ;
-L10:    iret            ;
+        iret            ;
+        iretq           ;
+L10:    nop; nop;         // put instructions above this or L10 changes
+                          // address, which changes all the jump instructions
+                          // and breaks the test case.
         ja      L10     ;
         jnbe    L10     ;
         jae     L10     ;
--- a/src/dmd/backend/ptrntab.c
+++ b/src/dmd/backend/ptrntab.c
@@ -64,6 +64,7 @@ OPTABLE0(INTO,    0xce,_i64_bit);
 OPTABLE0(INVD,    0x0f08,_I386);               // Actually a 486 only instruction
 OPTABLE0(IRET,    0xcf,_16_bit);
 OPTABLE0(IRETD,   0xcf,_32_bit | _I386);
+OPTABLE0(IRETQ,   0xcf,_64_bit | _I386);
 OPTABLE0(LAHF,    0x9f,_modax);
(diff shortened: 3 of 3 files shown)


WalterBright (changes requested):
Looks good! Please add test case: https://github.com/dlang/dmd/blob/master/test/runnable/iasm64.d#L3607

jacob-carlborg (changes requested):
This PR doesn’t reference an existing issue and therefore should include a changelog entry.

test/runnable/iasm64.d
```diff
@@ -3605,6 +3605,7 @@ void test50()
         0xCD, 0x67,     // int  067h
 //      0xCE,           // into
         0x66, 0xCF,     // iret
+		0x48, 0xCF,		// iretq
```
jacob-carlborg: Please use spaces instead of tabs.

jacob-carlborg:
Please squashing the commits when done.

changelog/iretq.dd
```diff
@@ -0,0 +1,15 @@
+`IRETQ` now supported in inline assembler.
```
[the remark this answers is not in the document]
leitimmel: You are right, it looked like a commit message.

leitimmel:
I'm unsure what to do with the failing test case on 64-bit Windows since I don't have a Windows installation set up for testing. Is it possible to obtain or inspect the test machine's build products somehow?

jacob-carlborg:
It's possible to give access to the AppVeyor build machine. Although I'm not sure of our policies for this.

WalterBright:
Also needs to be added to `test/runnable/iasm.d` (the 32 bit version of the asm tests).

leitimmel:
Does it, though? The test case rightfully complains about a 64-bit instruction in 32-bit mode because `IRETQ` is `IRET` with the AMD64-specific `REX.W` prefix.

WalterBright:
> a 64-bit instruction in 32-bit mode

Looks like I was wrong and it doesn't belong in iasm.d. Sorry for the inconvenience.

WalterBright (approved):

test/runnable/iasm64.d
```diff
@@ -3728,6 +3729,7 @@ void test50()
         int     0x67    ;
         //into          ;
 L10:    iret            ;
+        iretq           ;
```
rainers: Not sure why the CIs other tha Win64 are not complaining, but by inserting the instruction here the jumps to L10 below change. It passes for me when moving `L10` to the `iretq` instruction.

jacob-carlborg (approved):
I good with this assuming the commits are squashed.

leitimmel:
I now have reverted the 32-bit test case, adjusted the changelog item to standard formatting, and squashed the commits.

test/runnable/iasm64.d
leitimmel: Okay, that's tricky. Should we instead move the label to a separate `nop` instruction and add a comment or would that require a new pull request?

wilzbach:
Regarding the Jenkins failure - the Ocean failure is unrelated.
See https://github.com/dlang/ci/pull/208 (after the next push, Jenkins should be green again).

jacob-carlborg:
Please update the commit message. The idea of squashing the commits is when this PR is merge there should be no trace of the fixes that was made in the PR due to comments. It clutters the history and doesn't add any value. Using the same message as the title of this PR seems appropriate.

test/runnable/iasm64.d
rainers: What's tricky bout it. Just change it to
```
        iret            ;
L10:    iretq           ;
```

BTW: I suspect these tests are run only on Win64 because they are disabled for version D_PIC at the top.

leitimmel: The fix is not tricky, neither is the mistake I made, and of course I should have realised that the jumps change. The tricky part about the test case is that one has to realise this for oneself, which is why I would like to add a comment that explains what is going on, because in my opinion that's not something one *should* have to figure out.

test/runnable/iasm64.d
```diff
@@ -3727,7 +3731,11 @@ void test50()
         int     3       ;
         int     0x67    ;
         //into          ;
-L10:    iret            ;
+        iret            ;
+        iretq           ;
+L10:    nop             ; // put instructions above this or L10 changes
```
rainers: You will have to add two `nop`s to simulate the length of the `iret` instruction while avoiding changing all jumps.

leitimmel: Oh man, wrong again. Sorry :|

PR #8251, opened 2018-05-16
fix Issue 18821 - DMD segfault when doing unsafe operation outside of any function

WalterBright:


JinShil (approved):

jacob-carlborg (approved):

RazvanN7 (approved):

RazvanN7:
@Geod24 From what I know master gets merged into stable every 3 days so it's really not that important to target stable

nemanja-boric-sociomantic:
I think it is important that this bugfix targets stable branch because the bug is introduced in v2.080 (judging solely by commit message) and you do want this to hit 2.080.1, and not 2.081, right?

Geod24:
> master gets merged into stable every 3 days

What? Of course not. If anything, it's the other way around. Otherwise it wouldn't be stable at all.

RazvanN7:
Oh, sorry, I misread a message from Seb and got the wrong idea. You are right @Geod24

Geod24:
That's why that big dlang-bot warning is useless @wilzbach . But I didn't expect reviews to be as useless...

wilzbach:
What can I say that at this point?
It's a really simple rule:

__Regression fixes should ALWAYS ALWAYS go to stable__

And dlang-bot even reminds one about this...
Should we maybe add a CI that always fails for such cases, s.t. a PR can't be merged?
(then again Walter does have admin rights and could override that restriction too)

jacob-carlborg:
It’s all about making it easy to do the correct thing and difficult to do the wrong thing. Not being perfect.

nemanja-boric-sociomantic:
> And dlang-bot even reminds one about this...

To be fair, dlang bot didn't warned this time ;-) (no reference to `stable` branch in its comment).

jacob-carlborg:
> To be fair, dlang bot didn't warn this time ;-) (no reference to stable branch in its comment).

Perhaps because the target branch changed and then changed back?

MartinNowak:
Would be an idea to try addressing this problem with a PR linter/CI.

PR #8252, opened 2018-05-16
test name mangling of core.stdc.stdint types

WalterBright:
Types from `core.stdc.stdint` should mangle to the same as the corresponding C++ compiler does.

### DIFF
--- a/test/compilable/teststdint.d
+++ b/test/compilable/teststdint.d
@@ -1,0 +1,262 @@
+
+import core.stdc.stdint;
+
+extern (C++):
+
+void int8(int8_t i) { }
+void uint8(uint8_t i) { }
+
+void int16(int16_t i) { }
+void uint16(uint16_t i) { }
+
+void int32(int32_t i) { }
+void uint32(uint32_t i) { }
+
+void int64(int64_t i) { }
+void uint64(uint64_t i) { }
+
+void int_least8(int_least8_t i) { }
+void uint_least8(uint_least8_t i) { }
+
+void int_least16(int_least16_t i) { }
+void uint_least16(uint_least16_t i) { }
+
+void int_least32(int_least32_t i) { }
+void uint_least32(uint_least32_t i) { }
+
+void int_least64(int_least64_t i) { }
+void uint_least64(uint_least64_t i) { }
+
+void int_fast8(int_fast8_t i) { }
+void uint_fast8(uint_fast8_t i) { }
+
+void int_fast16(int_fast16_t i) { }
+void uint_fast16(uint_fast16_t i) { }
+
+void int_fast32(int_fast32_t i) { }
+void uint_fast32(uint_fast32_t i) { }
+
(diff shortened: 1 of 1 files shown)


WalterBright:
Blocked by https://github.com/dlang/druntime/pull/2183

rainers:
Shouldn't this actually test against a C++ compiler? Could be placed into externmangle.d/externmangle.cpp.

> I thought about it, but core.stdc.stdint is really a "header only" module, and it just didn't seem right to put tests in it.

I agree, but you can also put tests into druntime/test.

WalterBright:
> Shouldn't this actually test against a C++ compiler?

It's not really necessary since C++ compilers pretty much never change their mangling. The test cases were made from cut&pasting the output of the C++ compiler. Testing `.mangleof` is simpler (all in one file), and easier to debug when it goes wrong (no linker errors to figure out - the error prints what the string is and what it should be).

> I agree, but you can also put tests into druntime/test.

It could, but I like it here.

JinShil:
> It could, but I like it here.

The problem with the tests being here, is that https://github.com/dlang/druntime/pull/2183 can't be fully verified.  It would be best to add the tests to https://github.com/dlang/druntime/pull/2183, show that it's green, and then after that is merged, move them here to DMD.

WalterBright:
Getting the name mangling correct is the job of the compiler, not druntime.

Inevitably there isn't a sharp line separating druntime from dmd, and some judgement will be necessary now and then. This is one such case.

JinShil:
Druntime PR was merged.  Should no longer be blocked.

WalterBright:
Blocked by https://github.com/dlang/druntime/pull/2188

JinShil (approved):

wilzbach:
FYI this test is now failing on the self-bootstrapped SemaphoreCI:

```
Test compilable/teststdint.d failed.  The logged output:
../generated/linux/release/64/dmd -conf= -m64 -Icompilable  -fPIC  -odtest_results/compilable -oftest_results/compilable/teststdint_0.o  -c compilable/teststdint.d 
compilable/teststdint.d(211): Error: static assert:  `"_Z10int_fast16i" == "_Z10int_fast16l"` is false
```

WalterBright:
On my Ubuntu linux box, the mangling from the C version of the code is _Z10int_fast16l. The line in core.stdc.stdint for fast16 is:
```
        alias int_fast16_t  = cpp_long;
```
which all looks right. Anyhow, why would it pass all the other linux tests? What is different about semaphoreci?

wilzbach:
SemaphoreCI uses a compiler built from the PR, to build itself (instead of just using the host compiler as the other CIs do):

Use host DMD to build DMD ->use this built DMD with fresh DMD -> run tests

This means that one should be able to reproduce this if with dmd-nightly or using a freshly built host compiler via `HOST_DMD`.

I'm sorry that I can't be of more help for the next days as I am still traveling with my phone only.

PR #8255, opened 2018-05-16
fix Issue 18868 - nondeterministic static ctor/dtor

JohanEngelen:
Instead of using a counter to create unique static ctor and dtor function identifiers, use the deterministic line+column location that is also used for generating unittest function identifiers.

### DIFF
--- a/src/dmd/identifier.d
+++ b/src/dmd/identifier.d
@@ -138,6 +138,40 @@ nothrow:
         return idPool(buf.peekSlice());
     }
 
+    /***************************************
+     * Generate deterministic named identifier based on a source location,
+     * such that the name is consistent across multiple compilations.
+     * A new unique name is generated. If the prefix+location is already in
+     * the stringtable, an extra suffix is added (starting the count at "_1").
+     *
+     * Params:
+     *      prefix      = first part of the identifier name.
+     *      loc         = source location to use in the identifier name.
+     * Returns:
+     *      Identifier (inside Identifier.idPool) with deterministic name based
+     *      on the source location.
+     */
+    extern (D) static Identifier generateIdWithLoc(string prefix, const ref Loc loc)
+    {
+        OutBuffer buf;
+        buf.writestring(prefix);
+        buf.writestring("_L");
+        buf.print(loc.linnum);
+        buf.writestring("_C");
+        buf.print(loc.charnum);
+        auto basesize = buf.peekSlice().length;
+        uint counter = 1;
+        while (stringtable.lookup(buf.peekSlice().ptr, buf.peekSlice().length) !is null)
+        {
+            // Strip the extra suffix
+            buf.setsize(basesize);
+            // Add new suffix with increased counter
+            buf.writestring("_");
+            buf.print(counter++);
+        }
+        return idPool(buf.peekSlice());
+    }
+
     /********************************************
      * Create an identifier in the string table.
      */
--- a/test/runnable/imports/test18868_fls.d
+++ b/test/runnable/imports/test18868_fls.d
@@ -1,0 +1,33 @@
+module imports.test18868_fls;
+
+template FLS(T)
+{
+    int ctorcount = 0;
+    int dtorcount = 0;
+    int sharedctorcount = 0;
+    int shareddtorcount = 0;
+
+    static this()
+    {
+        assert(ctorcount == 0);
+        ctorcount += 1;
+    }
+
+    static ~this()
+    {
+        assert(dtorcount == 0);
+        dtorcount += 1;
+    }
+
+    shared static this()
+    {
+        assert(sharedctorcount == 0);
+        sharedctorcount += 1;
+    }
+
+    shared static ~this()
+    {
+        assert(shareddtorcount == 0);
+        shareddtorcount += 1;
+    }
+}
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -864,9 +864,7 @@ struct ASTBase
 
         extern (D) this(Loc loc, Loc endloc, StorageClass stc, char* codedoc)
         {
-            OutBuffer buf;
-            buf.printf("__unittestL%u_", loc.linnum);
-            super(loc, endloc, Identifier.generateId(buf.peekString()), stc, null);
+            super(loc, endloc, Identifier.generateIdWithLoc("__unittest", loc), stc, null);
             this.codedoc = codedoc;
         }
 
@@ -914,11 +912,11 @@ struct ASTBase
     {
         final extern (D) this(Loc loc, Loc endloc, StorageClass stc)
         {
-            super(loc, endloc, Identifier.generateId("_staticCtor"), STC.static_ | stc, null);
+            super(loc, endloc, Identifier.generateIdWithLoc("_staticCtor", loc), STC.static_ | stc, null);
         }
-        final extern (D) this(Loc loc, Loc endloc, const(char)* name, StorageClass stc)
+        final extern (D) this(Loc loc, Loc endloc, string name, StorageClass stc)
         {
-            super(loc, endloc, Identifier.generateId(name), STC.static_ | stc, null);
+            super(loc, endloc, Identifier.generateIdWithLoc(name, loc), STC.static_ | stc, null);
         }
 
         override void accept(Visitor v)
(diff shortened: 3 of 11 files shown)


Shachar:
If the same static this is instantiated several times due to different template extraction, using the line number would create false-positive on the duplicate detection.

src/dmd/identifier.d
```diff
@@ -138,6 +138,22 @@ nothrow:
         return idPool(buf.peekSlice());
     }
 
+    /***********************************************************
+     * Generate deterministic named Id based on the `loc` such that
+     * the naming is consistent across multiple compilations.
+     * `prefix` must be a null-terminated string.
+     */
```
jacob-carlborg: New functions should have standard Ddoc sections like `Returns:`  and `Params`.

test/runnable/imports/test18868_fls.d
```diff
@@ -0,0 +1,28 @@
+module imports.test18868_fls;
+
+template FLS(T) {
```
jacob-carlborg: Curly braces on their own lines.

jacob-carlborg (changes requested):

src/dmd/identifier.d
JohanEngelen: fixed

jacob-carlborg (approved):

wilzbach:
I restarted Jenkins, let's hope it was just a spurious failure in Ocean.

JohanEngelen:
@wilzbach Thanks, but it still fails :(

```
/var/lib/jenkins/dlang_projects@4/sociomantic-tsunami/ocean/integrationtest/unixlistener/main.d(281): Error: incompatible types for `(handleEcho) : (handleShutdown)`: `void function(const(char)[] args, void delegate(const(char)[] response) send_response) @system` and `void delegate(const(char)[] args, void delegate(const(char)[] response) send_response) @system`
```
In https://github.com/sociomantic-tsunami/ocean/blob/v4.1.x/integrationtest/unixlistener/main.d#L281

JohanEngelen:
Confirmed not related to this PR: https://github.com/dlang/dmd/pull/8257 has the same problem.

src/dmd/identifier.d
```diff
@@ -138,6 +138,27 @@ nothrow:
...
+     * Generate deterministic named identifier based on a source location,
+     * such that the name is consistent across multiple compilations.
+     * Params:
+     *      prefix      = first part of the identifier name, null-terminated.
+     *      loc         = source location to use is the identifier name.
+     * Returns:
+     *      Identifier (inside Identifier.idPool) with deterministic name based
+     *      on the source location.
+     */
+    static Identifier generateIdWithLoc(const(char)* prefix, const ref Loc loc)
```
WalterBright: Any reason this can't be `string prefix` ?

JohanEngelen: No, this should be `string`. Except when I do that:
`dmd/identifier.d(151): Error: string is used as a type`

Turns out `Identifier` has a `const char*` member called `string`!  >_<

JohanEngelen: @WalterBright I've changed it to string, by prefixing with `.`. Ugly, but this PR shouldn't become a refactoring of `Identifier`. I'll do that now in another PR.

jacob-carlborg: The refactoring PR has been merged now, the leading dot can now be removed?

JohanEngelen: @jacob-carlborg Yes, done.

src/dmd/identifier.d
```diff
@@ -138,6 +138,27 @@ nothrow:
         return idPool(buf.peekSlice());
     }
 
+    /***************************************
+     * Generate deterministic named identifier based on a source location,
+     * such that the name is consistent across multiple compilations.
+     * Params:
+     *      prefix      = first part of the identifier name.
+     *      loc         = source location to use is the identifier name.
```
jacob-carlborg: `is` doesn't sound correct. Should it be `in`?

JohanEngelen: should be `in`.

JohanEngelen:
all done

jacob-carlborg (approved):

jacob-carlborg:
Should this go to the stable branch?

JohanEngelen:
> Should this go to the stable branch?

It's a fairly critical bug, I find.
But because it depends on the refactor in identifier.d ....................................

JohanEngelen:
Rebased and retargeted to  `stable`

JohanEngelen:
I don't understand why it is failing after rebasing onto stable. It's failing on a "__unittest*" declaration, but that name has not changed with this PR (the code is almost identical, replacing a printf with a number of print statements...). Don't know what's happening here...

jacob-carlborg:
Is an older compiler being used without this PR https://github.com/dlang/dmd/pull/8224?

JohanEngelen:
Don't pull.
I've discovered an issue with this patch when mixins are involved: location information is not working correctly (location inside the mixin is used...).

JohanEngelen:
https://issues.dlang.org/show_bug.cgi?id=18880

JinShil:
Maybe add an counter to compilation unit that can be incremented each time an id is generated and appended to the location-based identifier.

JohanEngelen:
> Maybe add a counter to compilation unit that can be incremented each time an id is generated and appended to the location-based identifier.

That won't be deterministic because it will depend on the order in which things are compiled.

Shachar:
The only thing that comes to mind is a "static foreach iteration counter".

They will have to be recursivable (i.e.: if there is a static foreach inside another static foreach).

jacob-carlborg:
What about using both? Both file name plus line number and a counter. Try to have a local counter so it will be 0 in most cases.

JohanEngelen:
> Scope in this sense means everything that is possibly an "entry point" for semantic analysis, i.e. modules, template instances, etc. The scope name would be part of the mangled symbol name, and within each such scope care would have to be taken that semantic analysis always proceeds linearly to make sure the symbol indices are stable.

I think this won't work unless you define a new scope for each _different_ particular template instantiation of the _same_ source template. Because the order in which those are SemA'd is not deterministic.

An easy approach for this is (using nameclash as poor man's "scope detection" ;-)) : use linenumber+columnnumber. If a new symbol is generated and that name already exists: suffix with ever increasing counter. I'll implement this scheme and see if that does work well:
1. name = "<prefix>_L<line>_C<column>", if (!name.alreadyInIdPool) return name;
2. tryname = name ~ "_1"; if (!tryname.alreadyInIdPool) return tryname;
3. tryname = name ~ "_2; if (!tryname.alreadyInIdPool) return tryname;
4. ...

Extra problematic issue is that the current mixin linenumber scheme is already flawed:
- two mixins on same line will have same line number _and same columnnumber_ 
- two mixins on different lines can be crafted to result in same linenumber...
Resolution of these more advanced issues are for another PR/fix.

JohanEngelen:
This PR also fixes https://issues.dlang.org/show_bug.cgi?id=18880 which is the same issue that is plaguing the static ctor/dtors.

src/dmd/identifier.d
```diff
@@ -138,6 +138,39 @@ nothrow:
...
+    {
+        OutBuffer buf;
+        buf.writestring(prefix);
+        buf.writestring("_L");
+        buf.print(loc.linnum);
+        buf.writestring("_C");
+        buf.print(loc.charnum);
+        auto basesize = buf.peekSlice().length;
+        uint counter = 1;
+        while (stringtable.lookup(buf.peekSlice().ptr, buf.peekSlice().length) !is null) {
```
jacob-carlborg: Curly braces on their own lines.

jacob-carlborg (commented):
The code changes look good to me, not sure if it will be enough to fix the problems.

JohanEngelen:
All good now, ready for merging.

RazvanN7 (approved):

JohanEngelen:
Squashed and rebased into one commit onto current master.

JohanEngelen:
The semaphoreci failure is unrelated (the semaphore testing is very flaky, do more people experience that?).

JohanEngelen:
Would appreciate if someone could restart the semaphoreci tester.

wilzbach:
Restarted SemaphoreCi (and added you to the SemaphoreCI collaborators)
For the record the failure was:

```
../generated/linux/release/64/dmd -conf= -m64 -Irunnable  -fPIC -L-lstdc++  -odtest_results/runnable -oftest_results/runnable/externmangle_0  runnable/externmangle.d test_results/runnable/externmangle.cpp.o 
runnable/externmangle.d(128): Error: undefined identifier `cpp_ulong`, did you mean alias `c_ulong`?
runnable/externmangle.d(128): Error: undefined identifier `cpp_long`, did you mean alias `c_long`?
runnable/externmangle.d(128): Error: undefined identifier `cpp_ulong`, did you mean alias `c_ulong`?
runnable/externmangle.d(129): Error: undefined identifier `cpp_ulonglong`
runnable/externmangle.d(129): Error: undefined identifier `cpp_longlong`
runnable/externmangle.d(129): Error: undefined identifier `cpp_ulonglong`
runnable/externmangle.d(130): Error: undefined identifier `cpp_size_t`
runnable/externmangle.d(130): Error: undefined identifier `cpp_ptrdiff_t`
runnable/externmangle.d(130): Error: undefined identifier `cpp_size_t`


Test runnable/externmangle.d failed: expected rc == 0, exited with rc == 1
```

Seems like https://github.com/dlang/dmd/blob/master/test/runnable/externmangle.d was recently modified...

jacob-carlborg (approved):

PR #5716, opened 2016-04-27
[REG2.069] Issue 15962 - Don't strip off asserts to check internal compiler errors

9rnsr:

PR #8257, opened 2018-05-17
convert vec.c to dvec.d

WalterBright:


src/dmd/backend/dvec.d
```diff
@@ -0,0 +1,603 @@
...
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Copyright:   Copyright (c) 2013-2018 by The D Language Foundation, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/dvec.d, backend/dvec.d)
+ */
+
+
```
jacob-carlborg: Should there be a module declaration?

src/dmd/backend/dvec.d
```diff
@@ -0,0 +1,603 @@
...
+/****************
+ * Set vector.
+ */
+
+pure
+void vec_set(vec_t v)
+{
+    if (v)
+    {   memset(v, ~0, v[0].sizeof * vec_dim(v));
```
jacob-carlborg: Braces on their own lines.

jacob-carlborg:
The Visual Studio project file is not updated.

src/dmd/backend/dvec.d
jacob-carlborg: I know that but all other D files in DMD has a module declaration.

src/dmd/backend/dvec.d
```diff
@@ -0,0 +1,605 @@
...
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Copyright:   Copyright (c) 2013-2018 by The D Language Foundation, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/dvec.d, backend/dvec.d)
+ */
+
+module dvec;
```
rainers: Should be module dmd.backend.dvec, otherwise separate compilation fails.

src/vcbuild/dmd_backend.vcxproj.filters
```diff
@@ -1,446 +1,443 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
rainers: This is probably converting CRLF to LF. This should not happen for the VC projects.

jacob-carlborg:
Is there a reason for changing the name from `vec` to `dvec`?

src/dmd/backend/dvec.d
```diff
@@ -0,0 +1,605 @@
...
+                            break;
+                default:    memset(v,0,dim * vec_base_t.sizeof);
+                            break;
+            }
+            goto L1;
+        }
+        else
+        {
+            v = cast(vec_t) calloc(dim + 2, vec_base_t.sizeof);
+            assert(v);
```
rainers: This should better be reporting "out of memory", the backend uses `err_nomem()` for this.

src/dmd/backend/dvec.d
```diff
@@ -0,0 +1,605 @@
...
+        vec_t result;
+        if (dim < freelist.length && (vc = freelist[dim]) != null)
+        {
+            freelist[dim] = *cast(vec_t *)vc;
+            goto L1;
+        }
+        else
+        {
+            vc = cast(vec_t) calloc(nbytes, 1);
+            assert(vc);
```
rainers: Same here.

Otherwise, translation looks good.

src/dmd/backend/dvec.d
```diff
@@ -0,0 +1,605 @@
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
```
RazvanN7: I suggest adding a description of the module. I have no idea what the code in this file does. Also, a more descriptive file name wouldn't hurt

src/dmd/backend/dvec.d
WalterBright: Aye, there's the rub. I cannot make this work for both dmc and dmd, because they use separate directory structures. Upgrading the boot compiler to 2.079 is required. Unfortunately, this has been blocked for a year now, and is currently blocked by:
https://github.com/dlang/dmd/pull/8249
https://github.com/dlang/dmd/pull/8112

src/dmd/backend/dvec.d
WalterBright: It uses the assert because the file is supposed to be independent of any other project modules. The C++ version it was translated from did this, too.

src/vcbuild/dmd_backend.vcxproj.filters
WalterBright: Changed it back to CRLF.

src/dmd/backend/dvec.d
rainers: I haven't yet heard a convincing reason why dmc and dmd should not share a common directory structure, e.g. use module name `tk.dvec` and put it into the respective folder. Not using package names is unlikely to be a good idea for shared modules.

src/dmd/backend/dvec.d
rainers: The C++ version uses mem_calloc/mem_exception from mem.c.

`assert(v)` has the problem that it is likely to go away in a release build, so if you want no dependencies I'd recommend `if (!v) assert(false, "out of memory")`.

src/dmd/backend/dvec.d
WalterBright: The reason is:

1. It's a lot of nuisance work for me to crowbar dmc into a totally different directory structure - messing up the git history, changing all the imports, etc.
2. I'm the one doing the work, and this is the way I want to do the work, because mashing it into multiple directories is unproductive for me.
3. All the blocking items for making this work are inconsequential and do not negatively impact anything else.
4. All the blocking items are things we need to do anyway (like support the -mv command). I can hardly be the only one who wants to share the same modules among multiple projects.
5. assert() support in 2.068 is so heavy it is disabled in dmd, with various negative consequences that show up now and then. The back end relies heavily on asserts being ON. 2.079 has much better support for lightweight assert().

klickverbot: >  I cannot make this work for both dmc and dmd, because they use separate directory structures.

src/dmd/backend/dvec.d
WalterBright: See my other comment about why assert not being disabled is required in the back end.

src/dmd/backend/dvec.d
rainers: > It's a lot of nuisance work for me to crowbar dmc into a totally different directory structure - messing up the git history, changing all the imports, etc.

You'll have to change .c files to .d in your build system anyway. Is it any worse to prepend `tk/` or similar, too?

>     I'm the one doing the work, and this is the way I want to do the work, because mashing it into multiple directories is unproductive for me.

Valid point. Don't consider my comments as blockers.

>     All the blocking items for making this work are inconsequential and do not negatively impact anything else.
>     All the blocking items are things we need to do anyway (like support the -mv command). I can hardly be the only one who wants to share the same modules among multiple projects.

That's what packages are designed for. Adding -mv showed that you seem to think that it has failed, but I don't see how. BTW: if you don't want to use package names anyway, -I is good enough. IMO the existence of -mv just adds another point of failure and confusion.

>     assert() support in 2.068 is so heavy it is disabled in dmd, with various negative consequences that show up now and then. The back end relies heavily on asserts being ON. 2.079 has much better support for lightweight assert().

AFAICT the difference in code size/performance is negligible between C and D asserts (just strings instead of raw pointers). I just tested removing C asserts from the backend: compiling release-dmd itself with a VC built 64-bit version took 110s instead of 114s with asserts enabled (dmc built dmd2.080 took 137s). IIRC the change in the frontend was in the same ballpark.

WalterBright: See https://github.com/dlang/dmd/pull/5716 for more discussion on the effects of `assert` on the compiler.

WalterBright: > Is it any worse

Yes. I generally try to make the first .d version close enough that git will show the diffs.

> but I don't see how

For one thing, the unzip utility will no longer unzip subdirectories that are not in the cwd. So if you want to zip up the project, you're hosed unless they share subdirectories. But dmd and dmc use different hierarchies.

I understand you don't like `-mv`. Nobody is making anyone use it, and so it doesn't negatively impact them. It's also TRIVIAL. I simply cannot understand the resistance to it. (I've implemented a lot of stuff that I don't personally find useful, but others do.) It's a heluva lot more work for me to disrupt all the layouts on the various machines I use, and the other git repositories that dmc resides in, than just add `-mv=` to the command line, and I'm done.

Your proposal is like what happens when your credit card gets a fraud alert. You then have to notify and change the number with every vendor you've got an account with, and then for months afterwards you get dinged on the ones you overlooked. I've suffered through that many times, nobody does that for fun if they don't have to.

It's not just me, it's not a nice thing to inflict on our users, either.

I've been repeating these arguments for a year. I could have had the back end converted a year ago. Instead I spend time on this back and forth over a trivial blocker.

klickverbot: > I could have had the back end converted a year ago. Instead I spend time on this back and forth over a trivial blocker.

You keep repeating this claim, yet all these "blockers" are trivially worked around. If you were really interested in converting the backend, you could have done just that a year ago. Your earlier statement applies verbatim: "It's also TRIVIAL. I simply cannot understand the resistance to it."

As I've pointed out elsewhere, I believe the only such workaround that would require code changes is that to support "release `assert`s". Any time you might need to spend writing a sed script/… to convert those `assert` calls to a custom function is entirely dwarfed by the time it takes to do the conversion itself, or indeed the required changes to all the various build systems that go with it.

So, can we please stop pretending that "I need it to convert the backend to D" is a valid argument in favour of bumping the bootstrap compiler version requirement? If you'd like to change the version requirements to make your life easier, I'm not sure why you would need to make up bogus reasons to ask for that. Your word still carries weight here.

WalterBright: > sed scripts

Adding more technical debt is to be avoided, not embraced. Never mind that sed isn't a standard tool on Windows, so I'd have to write a program to do that, and check it into the repositories. I know what I think when I see such things in other projects.

> trivial

Yes, I could write a script to copy directory structures around before each compile. Is that something you'd want to do for your project? I prefer to avoid rube goldberg build setups. (Heaven knows the build system is fragile enough as it is.)

> pretending

I'm used to doing ugly things to get things to work when I have no choice. I've even resorted to writing programs to patch object files. But there's no reason to do that here. The advantage to owning the tool chain is the fix can be put in the right spot.

> Your word still carries weight here.

Then please pull the blocking PRs.

klickverbot: >> sed scripts
> Adding more technical debt is to be avoided, not embraced. Never mind that sed isn't a standard tool on Windows

Replacing `assert` with `my_assert` is a one-time thing, to be done along with many other mechanical changes during the C -> D conversion.

> > trivial

WalterBright: I had my own non-standard `assert` for years in the back end. I finally removed it along with a lot of other garbage I used to think was a good idea.

WalterBright: > One is failing,

Because it's blocked by the other.

> and @kinke seems to have a valid suggestion for the other one.

Regardless of the merit of his suggestion, it requires delaying for multiple release cycles.

WalterBright:
@CyberShadow can you please upgrade the boot compiler on CyberShadow/DAutoTest to 2.079 so this passes?

CyberShadow:
DAutoTest is already set up to use 2.079 once #8112 is merged.

Do you want to merge this PR before #8112? If so, I could change the detection logic enabling a newer host compiler to check for `dvec.d` instead of `-mv` in the makefile.

WalterBright:
> Do you want to merge this PR before #8112?

Yes, please. #8112 is not a blocker for this, it's just a canary for needed features in the boot compiler.

CyberShadow:
Done, should take effect on the next rebuild.

WalterBright:
@CyberShadow thanks! It's working now, ready to pull!

JinShil (approved):
There are a lot of valid concerns voiced in the review of this PR, but Walter has addressed most of it and made his decision.  I think it's time to move on; some things can even be addressed after the conversion to D.

quickfur:
This PR broke my local dmd build, which bootstraps with 2.069:
````
../generated/linux/release/64/backend.a(dvec.o): In function `_D4dvec9VecGlobal9terminateMUNbNiZv':
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:65: undefined reference to `_D4dvec8__assertFiZv'
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:67: undefined reference to `_D4dvec7__arrayZ'
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:76: undefined reference to `_D4dvec7__arrayZ'
../generated/linux/release/64/backend.a(dvec.o): In function `_D4dvec9VecGlobal8allocateMUNbNimZPm':
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:87: undefined reference to `_D4dvec7__arrayZ'
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:89: undefined reference to `_D4dvec7__arrayZ'
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:106: undefined reference to `_D4dvec8__assertFiZv'
../generated/linux/release/64/backend.a(dvec.o): In function `_D4dvec9VecGlobal3dupMUNbNixPmZPm':
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:131: undefined reference to `_D4dvec7__arrayZ'
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:133: undefined reference to `_D4dvec7__arrayZ'
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:138: undefined reference to `_D4dvec8__assertFiZv'
../generated/linux/release/64/backend.a(dvec.o): In function `_D4dvec9VecGlobal4freeMUNbNiPmZv':
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:162: undefined reference to `_D4dvec7__arrayZ'
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:163: undefined reference to `_D4dvec7__arrayZ'
../generated/linux/release/64/backend.a(dvec.o): In function `vec_setbit':
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:269: undefined reference to `_D4dvec8__assertFiZv'
../generated/linux/release/64/backend.a(dvec.o): In function `vec_clearbit':
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:280: undefined reference to `_D4dvec8__assertFiZv'
../generated/linux/release/64/backend.a(dvec.o): In function `vec_testbit':
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:299: undefined reference to `_D4dvec8__assertFiZv'
../generated/linux/release/64/backend.a(dvec.o): In function `vec_andass':
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:348: undefined reference to `_D4dvec8__assertFiZv'
/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:349: undefined reference to `_D4dvec8__assertFiZv'
../generated/linux/release/64/backend.a(dvec.o):/mnt/1/usr/src/d/dmd/src/dmd/backend/dvec.d:355: more undefined references to `_D4dvec8__assertFiZv' follow
--- errorlevel 1
make[1]: *** [posix.mak:475: ../generated/linux/release/64/dmd] Error 1
make[1]: Leaving directory '/mnt/1/usr/src/d/dmd/src'
make: *** [posix.mak:8: all] Error 2
````

Checking out the revision before this PR was merged fixes the problem.

WalterBright:
You'll need to upgrade your boot compiler.

PR #8258, opened 2018-05-17
[Trivial] Rename Identifier.string to Identifier.name to not clash with the `string` type.

JohanEngelen:
This will make it easier in future to refactor code to use D strings instead of C-style strings.


WalterBright (approved):

JinShil:
Jenkins failure appears to be caused by Sociomantic Ocean's latest release.  Merging anyway.

JinShil (approved):

PR #8260, opened 2018-05-18
Fix Issue 18871 & 18819 - DMD illegal hardware instruction crash

JinShil:
I'm still trying to create a reduced test case.  When I print out `ea.toChars()`, I get this:

```
CallExp::interpret() _ArrayDtor([Vector(), Vector()])
1 ea = Vector[] [Vector(), Vector()]
2 ea = Vector[], arrayliteral [Vector(), Vector()]
```

If someone knows how to get a reduced test case from that.  Please let me know.

### DIFF
--- a/src/dmd/dinterpret.d
+++ b/src/dmd/dinterpret.d
@@ -4827,17 +4827,23 @@ public:
             {
                 assert(e.arguments.dim == 1);
                 Expression ea = (*e.arguments)[0];
-                //printf("1 ea = %s %s\n", ea.type.toChars(), ea.toChars());
+                // printf("1 ea = %s %s\n", ea.type.toChars(), ea.toChars());
                 if (ea.op == TOK.slice)
                     ea = (cast(SliceExp)ea).e1;
                 if (ea.op == TOK.cast_)
                     ea = (cast(CastExp)ea).e1;
 
-                //printf("2 ea = %s, %s %s\n", ea.type.toChars(), Token::toChars(ea.op), ea.toChars());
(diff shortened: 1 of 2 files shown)


JinShil:
Figured it out.  This is ready for review.

dnadlinger (commented):
Is _ArrayDtor supposed to be a public API?

JinShil:
> Is _ArrayDtor supposed to be a public API?

I don't think so, but so far, that's the only way I've been able to reproduce the bug without adding dependencies on `std.experimental.allocator`.  I created that test based on what I saw the compiler's lowering.

JinShil:
Actually, I should probably wait until https://github.com/dlang/druntime/pull/2147 and https://github.com/dlang/dmd/pull/8067 are pulled and test this again.

JinShil:
Sorry, I don't understand.  I'm not seeing the special case.

JinShil:
> Calling destructors directly on an ArrayLiteralExp seems somewhat weird

Yeah, that doesn't seem right, does it.  I'll use digger to see if I can identify what caused the regression.

JinShil:
It appears the regression was caused by https://github.com/dlang/phobos/pull/6411

I don't think this is the right fix.  Calling a constructor on an array literal is nonsense.

JinShil:
Ok, reopened in case someone wants to merge it.

JinShil:
Fixed test case thanks to reduction in given in https://issues.dlang.org/show_bug.cgi?id=18819#c4

jacob-carlborg (approved):

PR #8261, opened 2018-05-18
add =void, cantExp(), remove dead code

WalterBright:
Makes the code simpler and slightly more efficient. The error message printing turns out to be dead code - the interpreter checks it before calling the function.

### DIFF
--- a/src/dmd/constfold.d
+++ b/src/dmd/constfold.d
@@ -43,7 +43,10 @@ private Expression expType(Type type, Expression e)
     return e;
 }
 
-/* ================================== isConst() ============================== */
+/************************************
+ * Returns:
+ *    true if e is a constant
+ */
 extern (C++) int isConst(Expression e)
 {
     //printf("Expression::isConst(): %s\n", e.toChars());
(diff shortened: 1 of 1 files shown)


src/dmd/constfold.d
```diff
@@ -1358,13 +1378,10 @@ extern (C++) UnionExp Slice(Type type, Expression e1, Expression lwr, Expression
     else if (e1.op == TOK.arrayLiteral && lwr.op == TOK.int64 && upr.op == TOK.int64 && !hasSideEffect(e1))
     {
         ArrayLiteralExp es1 = cast(ArrayLiteralExp)e1;
-        uinteger_t ilwr = lwr.toInteger();
-        uinteger_t iupr = upr.toInteger();
-        if (iupr > es1.elements.dim || ilwr > iupr)
-        {
-            e1.error("array slice `[%llu .. %llu]` is out of bounds", ilwr, iupr);
```
jacob-carlborg: What happens with this error?

JinShil: > The error message printing turns out to be dead code

... from the description in this PR, I believe.

JinShil (approved):
Not 100% coverage, but I doubt I'll be able to persuade you to add the tests.  Feel free to prove me wrong.

PetarKirov (approved):
LGTM. I think we're good about the test coverage, as the changes don't alter the behavior of non-covered core in a significant way.

jacob-carlborg (approved):

ibuclaw:
This PR caused a regression (the error is not dead code after all).

https://issues.dlang.org/show_bug.cgi?id=22881

PR #8263, opened 2018-05-19
[refactor] Simplify parameter mismatch error code

ntrel:
* Use one `OutBuffer` instead of 3, reducing memory allocations.
* Avoid repeating common string messages.
* Add a missing test for typesafe variadics.

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -4780,19 +4780,20 @@ extern (C++) final class TypeFunction : TypeNext
     }
 
     // arguments get specially formatted
-    private const(char)* getParamError(const(char)* format, Expression arg, Parameter par)
+    private const(char)* getParamError(Expression arg, Parameter par)
     {
         // show qualification when toChars() is the same but types are different
         auto at = arg.type.toChars();
         bool qual = !arg.type.equals(par.type) && strcmp(at, par.type.toChars()) == 0;
         if (qual)
             at = arg.type.toPrettyChars(true);
-        OutBuffer as;
-        as.printf("`%s` of type `%s`", arg.toChars(), at);
-        OutBuffer ps;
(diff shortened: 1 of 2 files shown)


src/dmd/mtype.d
```diff
@@ -4780,19 +4780,20 @@ extern (C++) final class TypeFunction : TypeNext
...
-        as.printf("`%s` of type `%s`", arg.toChars(), at);
-        OutBuffer ps;
-        ps.printf("`%s`", parameterToChars(par, this, qual));
         OutBuffer buf;
-        buf.printf(format, as.peekString(), ps.peekString());
+        // only mention rvalue if it's relevant
+        const rv = !arg.isLvalue() && par.storageClass & (STC.ref_ | STC.out_);
+        const char* fmt = rv ?
+            "cannot pass rvalue argument `%s` of type `%s` to parameter `%s`" :
+            "cannot pass argument `%s` of type `%s` to parameter `%s`";
```
jacob-carlborg: Seems a bit unnecessary to duplicate the whole format string when the only difference is `rvalue`. Perhaps `rvalue` can be passed as a parameter to `printf` as well?

RazvanN7 (approved):

src/dmd/mtype.d
wilzbach: Yep, last time I tried something in this direction, it was complained that this increases the binary size.
It's just "cannot pass%s" and then `rv ? " rvalue" : ""`, no?

ntrel: Fix is here: https://github.com/ntrel/dmd/commit/6219e5e189ad7b758488fc751187ddbe6b792909

wilzbach: Cool. Would you mind submitting it as PR too? Thanks!

PR #8265, opened 2018-05-20
Fix issue 14086 - Invalid extern C++ name for constructor

TurkeyMan:
Destructor mangles correctly, but constructor does not.
This has wasted a lot of time with people trying to understand link problems with `extern(C++)`.

As with the dtor, the ABI appears to be perfectly compatible... so, mangling correctly.
Also, fixed the virtual-ness mangling of the destructor. Windows dtor mangling only half-worked.

### DIFF
--- a/changelog/mangle_ctor.dd
+++ b/changelog/mangle_ctor.dd
@@ -1,0 +1,4 @@
+`extern (C++)` now mangles class constructor's correctly.
+
+`extern (C++)` mangling of class constructor is working, also virtual and non-virtual destructors mangle correctly too.
+C++ semantics are not yet perfectly matching, but some simple cases are working.
--- a/src/dmd/backend/dt.d
+++ b/src/dmd/backend/dt.d
@@ -41,7 +41,7 @@ private:
     dt_t** pTail;
 
 public:
-    this()
+    extern (D) this()
     {
         pTail = &head;
     }
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -590,7 +590,9 @@ private final class CppMangleVisitor : Visitor
                 buf.writeByte('N');
                 CV_qualifiers(d.type);
                 prefix_name(p);
-                if (d.isDtorDeclaration())
+                if (d.isCtorDeclaration())
+                    buf.writestring("C1");
+                else if (d.isDtorDeclaration())
                     buf.writestring("D1");
                 else
                     source_name(ti);
@@ -621,7 +623,11 @@ private final class CppMangleVisitor : Visitor
                 prefix_name(p);
                 //printf("p: %s\n", buf.peekString());
 
-                if (d.isDtorDeclaration())
+                if (d.isCtorDeclaration())
+                {
+                    buf.writestring("C1");
+                }
+                else if (d.isDtorDeclaration())
                 {
                     buf.writestring("D1");
                 }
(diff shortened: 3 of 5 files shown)


JinShil:
Please rename the commit messaged to ensure the dlang-bot establishes a relationship between this PR and the bugzilla issue.  I usually make the PR title and commit message the same.

JinShil:
Adding the "Vision" label as this is congruent with item 4 of [Vision 2018H1](https://wiki.dlang.org/Vision/2018H1)

TurkeyMan:
It's in the PR header... that is included in the comment when pulled.
dbot will see that won't he?

TurkeyMan:
Why is DAutoTest red? I had this problem last PR too...

rainers:
It complains about your terse changelog entry:
```
object.Exception@../tools/changed.d(216): Changelog entries should consist of one title line, a blank separator line, and a description.
```

kinke:
When calling an extern(C++) ctor implemented in D from C++, one needs to be aware that the ctor assumes the instance is pre-initialized with `T.init` and may not initialize all fields, so one may end up with garbage.

Wrt. destruction, the ABIs diverge when it comes to destructing by-value arguments - for C++, the caller destructs it (and passes a ref to the callee), while D compilers move the argument and let the callee destruct the param. This also happens when using `extern(C++)` for the D code, with DMD, LDC and GDC, so D/C++ interop for non-POD values passed by value is broken on most platforms anyway. On Win64 (where non-PODs are passed by ref), when passing a non-POD by value from C++ to a D function, I expect a double-free (and D's callee doesn't reset it to `T.init` after the first destruction).

TurkeyMan:
@kinke I've been desperately searching for this document!!
Where is all the very important information documented?
I can't even find docs that talk about `__xdtor`, or something so simple as the vtable layout! >_<

It's really hard to start to address these issues with no information available.

TurkeyMan:
Next port of call is to attempt to make the behaviours of `extern(C++)` stuff precisely match C++.

TurkeyMan:
But for D? The details for D should be published somewhere.

TurkeyMan:
@kinke Are you on IM? I have a zillion questions I need quick answers to :/

jacob-carlborg:
It’s usually quite difficult to find documentation for these kinds of things. You usually find out when implementing a compiler.

WalterBright:
> Nope, MSVC++ on Win64 seems to do it exactly like D (tested with LDC), letting the callee destruct it. It's still broken on POSIX systems, at least for x86_64.

No surprise there. The original VC++ was based on the ABI used by Zortech C++ :-)

We sold a lot of Zortech C++ compilers into Microsoft before they started implementing C++. The COM ABI is directly the ABI used by ZTC++ for virtual functions.

@kinke is quite right that not only is there no document for this, when such documents exist they tend to be all lies. The only reliable way is to write test cases, compile them, and examine the object file.

BTW, I didn't realize that Posix does these things differently.

WalterBright (approved):
Approving despite the problems @kinke reported. At least matching the mangling is a good first step.

TurkeyMan:
Thanks. My next struggle is trying to work out how __xdtor is fabricated, and mangle THAT symbol as C++ dtor instead of the explicit dtor. __xdtor does what C++ expects.

kinke:
@TurkeyMan: That's [here](https://github.com/dlang/dmd/blob/master/src/dmd/clone.d#L795).  `__xdtor` seems to be an alias for `__aggrDtor`.

TurkeyMan:
Huh... I don't understand why semaphore says it failed?

TurkeyMan:
Okay, seriously... after how ever many _hours_ waiting for the CI to build, then it fails.
This CI situation is really not great.

jacob-carlborg (approved):

PR #8268, opened 2018-05-20
add constructor to ArrayExp

WalterBright:


RazvanN7 (approved):

src/dmd/expression.d
```diff
@@ -5681,6 +5681,12 @@ extern (C++) final class AddrExp : UnaExp
         super(loc, TOK.address, __traits(classInstanceSize, AddrExp), e);
     }
 
+    extern (D) this(const ref Loc loc, Expression e, Type t)
+    {
+        super(loc, TOK.address, __traits(classInstanceSize, AddrExp), e);
```
JinShil: I think I would have called `this` instead of `super` here.

WalterBright: You're right. Feel free to do a PR to improve that!

PR #8270, opened 2018-05-20
dinterpret.d: move public interface to top of file

WalterBright:
C/C++ code tends to fall an inverted pattern, with the private code on top and the public at the bottom. This comes about because of no forward referencing.

But D has forward referencing, so the public interface should be at the top of the file and everything else "below the fold" which is completely hidden from other modules.

### DIFF
--- a/src/dmd/dinterpret.d
+++ b/src/dmd/dinterpret.d
@@ -43,6 +43,121 @@ import dmd.tokens;
 import dmd.utf;
 import dmd.visitor;
 
+/*************************************
+ * Entry point for CTFE.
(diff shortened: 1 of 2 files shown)


RazvanN7 (approved):

nemanja-boric-sociomantic:
A suggestion unrelated to the changes done: if your PR has only one commit, instead of writing the description in the github's PR description, you can include it into the commit's body and GitHub will automatically populate the PR description with it. That way the information is included in the git repository as well.

stefan-koch-sociomantic:
If ever I saw churn ....

PR #8271, opened 2018-05-20
add numberOfElems() to mtype.d

WalterBright:
Pulling out common code into a function.

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -2594,6 +2594,33 @@ extern (C++) abstract class Type : RootObject
         return t;
     }
 
+    /*******************************************
+     * Compute number of elements for a (possibly multidimensional) static array,
+     * or 1 for other types.
+     * Params:
+     *  loc = for error message
+     * Returns:
+     *  number of elements, uint.max on overflow
+     */
+    final uint numberOfElems(const ref Loc loc)
+    {
+        //printf("Type::numberOfElems()\n");
+        uinteger_t n = 1;
+        Type tb = this;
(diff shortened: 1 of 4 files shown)


src/dmd/mtype.d
```diff
@@ -2594,6 +2594,32 @@ extern (C++) abstract class Type : RootObject
...
+        //printf("Type::numberOfElems()\n");
+        uinteger_t n = 1;
+        Type tb = this;
+        while ((tb = tb.toBasetype()).ty == Tsarray)
+        {
+            bool overflow = false;
+            n = mulu(n, (cast(TypeSArray)tb).dim.toUInteger(), overflow);
+            if (overflow || n > uint.max)
+            {
+                error(loc, "1static array `%s` size overflowed to %llu", toChars(), cast(ulong)n);
```
klickverbot: "1"?

src/dmd/mtype.d
```diff
@@ -2594,6 +2594,32 @@ extern (C++) abstract class Type : RootObject
         return t;
     }
 
+    /*******************************************
+     * Compute number of elements in type.
```
klickverbot: What is "the number of elements in type"?

Maybe something like "Returns the total number of elements for a (possibly multidimensional) static array, or 1 otherwise." would be clearer?

dnadlinger (approved):

src/dmd/mtype.d
WalterBright: debug string left in, oops

PR #8273, opened 2018-05-21
improve buildOpAssign()

WalterBright:
Should make it easier to read.

### DIFF
--- a/src/dmd/clone.d
+++ b/src/dmd/clone.d
@@ -37,12 +37,19 @@ import dmd.tokens;
 
 /*******************************************
  * Merge function attributes pure, nothrow, @safe, @nogc, and @disable
+ * from f into s1.
+ * Params:
+ *      s1 = storage class to merge into
+ *      f = function
+ * Returns:
+ *      merged storage class
  */
-extern (C++) StorageClass mergeFuncAttrs(StorageClass s1, FuncDeclaration f)
+extern (C++) StorageClass mergeFuncAttrs(StorageClass s1, const FuncDeclaration f) pure
 {
     if (!f)
         return s1;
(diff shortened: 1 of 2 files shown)


src/dmd/expression.d
```diff
@@ -1722,6 +1722,16 @@ extern (C++) abstract class Expression : RootObject
         return e1;
     }
 
+    static Expression combine(Expression e1, Expression e2, Expression e3)
+    {
+        return combine(combine(e1, e2), e3);
+    }
+
+    static Expression combine(Expression e1, Expression e2, Expression e3, Expression e4)
```
JinShil: These two functions seem a little weird.  Can we make it a single variadic; maybe a template?

src/dmd/clone.d
```diff
@@ -141,20 +145,32 @@ extern (C++) FuncDeclaration hasIdentityOpAssign(AggregateDeclaration ad, Scope*
 private bool needOpAssign(StructDeclaration sd)
 {
     //printf("StructDeclaration::needOpAssign() %s\n", sd.toChars());
-    if (sd.isUnionDeclaration())
+
+    static bool need()
```
JinShil: Naming conventions, as far as I'm aware, would prefer `isNeeded()` and `isNotNeeded`.

jacob-carlborg: Do we really need both?

src/dmd/expression.d
WalterBright: It's obvious what they do with a quick glance. Using variadics or templates, not so easy.

RazvanN7 (approved):
I'll merge this and make a new PR which addresses the reviewers comments

PR #8275, opened 2018-05-21
Fix issue 18891 - extern(C++) destructor prototype should just link

TurkeyMan:
This patch inhibits generation of __xdtor in the presence of extern(C++) destructor prototypes.
Destructors supplied by C++ code already do full destruction, so the field/aggregate destructors are not necessary.

### DIFF
--- a/src/dmd/clone.d
+++ b/src/dmd/clone.d
@@ -802,87 +802,92 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
     Loc declLoc = ad.dtors.dim ? ad.dtors[0].loc : ad.loc;
     Loc loc; // internal code should have no loc to prevent coverage
 
-    Expression e = null;
-    for (size_t i = 0; i < ad.fields.dim; i++)
+    // if the dtor is an extern(C++) prototype, then we expect it performs a full-destruction; we don't need to build a full-dtor
+    const bool dtorIsCppPrototype = ad.dtors.dim == 1 && ad.dtors[0].linkage == LINK.cpp && !ad.dtors[0].fbody;
+    if (!dtorIsCppPrototype)
     {
-        auto v = ad.fields[i];
-        if (v.storage_class & STC.ref_)
-            continue;
-        if (v.overlapped)
-            continue;
-        auto tv = v.type.baseElemOf();
-        if (tv.ty != Tstruct)
-            continue;
-        auto sdv = (cast(TypeStruct)tv).sym;
-        if (!sdv.dtor)
-            continue;
-        sdv.dtor.functionSemantic();
-
-        stc = mergeFuncAttrs(stc, sdv.dtor);
-        if (stc & STC.disable)
+        Expression e = null;
+        for (size_t i = 0; i < ad.fields.dim; i++)
         {
-            e = null;
-            break;
-        }
-
-        Expression ex;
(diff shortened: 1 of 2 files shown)


src/dmd/clone.d
```diff
@@ -802,87 +802,92 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
     Loc declLoc = ad.dtors.dim ? ad.dtors[0].loc : ad.loc;
     Loc loc; // internal code should have no loc to prevent coverage
 
-    Expression e = null;
-    for (size_t i = 0; i < ad.fields.dim; i++)
+    // if the dtor is an extern(C++) prototype, then we expect it performs a full-destruction; we don't need to build a full-dtor
+    bool dtorIsCppPrototype = ad.dtors.dim == 1 && ad.dtors[0].linkage == LINK.cpp && !ad.dtors[0].fbody;
```
jacob-carlborg: `const` please.

test/compilable/cppmangle.d
```diff
@@ -463,3 +463,32 @@ version (Windows)
...
+
+/**************************************/
+// https://issues.dlang.org/show_bug.cgi?id=18891
+
+extern (C++) class C18891
+{
+    ~this();
+    extern (C++) struct Agg
+    {
+        ~this() {}
```
jacob-carlborg: I'm not entirely sure of the DMD AST but won't this be a body in the AST? I'm thinking about this line https://github.com/dlang/dmd/pull/8275/files#diff-12b8cdd12102ad3f468b3a4faa04f0aeR806

TurkeyMan: Oh? I wasn't sure about that line.
I couldn't work out how to identify if a function is a prototype or not. That's what I came up with.

TurkeyMan: Oh? I wasn't sure about that line.
I couldn't work out how to identify if a function is a prototype or not. That's what I came up with.

src/dmd/clone.d
```diff
@@ -802,87 +802,92 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
     Loc declLoc = ad.dtors.dim ? ad.dtors[0].loc : ad.loc;
     Loc loc; // internal code should have no loc to prevent coverage
 
-    Expression e = null;
-    for (size_t i = 0; i < ad.fields.dim; i++)
+    // if the dtor is an extern(C++) prototype, then we expect it performs a full-destruction; we don't need to build a full-dtor
+    const bool dtorIsCppPrototype = ad.dtors.dim == 1 && ad.dtors[0].linkage == LINK.cpp && !ad.dtors[0].fbody;
```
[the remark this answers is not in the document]
jacob-carlborg: I think so.

test/compilable/cppmangle.d
jacob-carlborg: The point of the test is that there should be one destructor without a body and one with?

TurkeyMan: Yes, it should be structured such that the class *wants* to produce an aggregate destructor, but doesn't.

TurkeyMan: The tests confirm that `__dtor` is `__xdtor` (alias), ie, they are the same function; no aggregate destructor was generated.

PR #8276, opened 2018-05-22
Start using stack temporaries instead of allocated data

WalterBright:
Convert interpreter to using stack temporaries.

### DIFF
--- a/src/dmd/dinterpret.d
+++ b/src/dmd/dinterpret.d
@@ -749,7 +749,7 @@ private void ctfeCompile(FuncDeclaration fd)
  * Return result expression if successful, TOK.cantExpression if not,
  * or CTFEExp if function returned void.
  */
-private Expression interpret(FuncDeclaration fd, InterState* istate, Expressions* arguments, Expression thisarg)
+private Expression interpretFunction(FuncDeclaration fd, InterState* istate, Expressions* arguments, Expression thisarg)
 {
     debug (LOG)
     {
@@ -1008,9 +1008,11 @@ public:
     InterState* istate;
     CtfeGoal goal;
     Expression result;
+    UnionExp* pue;              // storage for `result`
 
-    extern (D) this(InterState* istate, CtfeGoal goal)
+    extern (D) this(UnionExp* pue, InterState* istate, CtfeGoal goal)
     {
+        this.pue = pue;
         this.istate = istate;
         this.goal = goal;
     }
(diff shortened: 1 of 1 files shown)


WalterBright:
The purpose of this is to speed up CTFE and reduce its memory consumption. There's a lot more to come, this is just the first step.

src/dmd/dinterpret.d
```diff
@@ -6230,11 +6249,26 @@ public:
     }
 }
 
-private Expression interpret(Expression e, InterState* istate, CtfeGoal goal = ctfeNeedRvalue)
+/********************************************
+ * Interpret the expression.
+ * Returns:
+ *    resulting expression
+ */
+
```
jacob-carlborg: No `Params` section.

src/dmd/dinterpret.d
```diff
@@ -6230,11 +6249,26 @@ public:
...
+Expression interpret(Expression e, InterState* istate, CtfeGoal goal = ctfeNeedRvalue)
+{
+    UnionExp ue = void;
+    auto result = interpret(&ue, e, istate, goal);
+    if (result == ue.exp())
+        result = ue.copy();
+    return result;
+}
+
+Expression interpret(UnionExp* pue, Expression e, InterState* istate, CtfeGoal goal = ctfeNeedRvalue)
```
jacob-carlborg: No documentation.

src/dmd/dinterpret.d
```diff
@@ -6248,11 +6282,20 @@ private Expression interpret(Expression e, InterState* istate, CtfeGoal goal = c
...
+Expression interpret(Statement s, InterState* istate)
+{
+    UnionExp ue = void;
+    auto result = interpret(&ue, s, istate);
+    if (result == ue.exp())
+        result = ue.copy();
+    return result;
+}
+
+Expression interpret(UnionExp* pue, Statement s, InterState* istate)
```
jacob-carlborg: No documentation.

jacob-carlborg (dismissed):

RazvanN7 (approved):

JinShil:
Just for my own education, how were any allocations saved with this PR?  I can only see that output was returned via a pointer parameter instead of returning it through `result`.

WalterBright:
> how were any allocations saved with this PR?

Given `enum e = a + b + c;`, one allocation is saved.

PR #8277, opened 2018-05-22
Fix issue 15589 - extern(C++) virtual destructors are not put in vtbl[]

TurkeyMan:
I'm plumbing in support for virtual destructors in extern(C++) classes.
Can I please get some initial review to know that I'm on the right track, and that the approach is okay?

With this initial bit of code, `__xdtor` is made virtual and placed into the vtable.
Calling `__xdtor` correctly performs a virtual call and destructs derived types (supplied by C++).
Is it also possible to declare the a base-class in D, derive from it in C++, and the virtual destruction will correctly destruct the mixed-language hierarchy. 🎉

I haven't supported overriding yet (for derived types in D). That should be straight-forward.

### DIFF
--- a/src/dmd/clone.d
+++ b/src/dmd/clone.d
@@ -15,6 +15,7 @@ module dmd.clone;
 import core.stdc.stdio;
 import dmd.aggregate;
 import dmd.arraytypes;
+import dmd.dclass;
 import dmd.declaration;
 import dmd.dscope;
 import dmd.dstruct;
@@ -881,6 +882,34 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
                 ex = new CallExp(loc, new IdentifierExp(loc, Id.__ArrayDtor), ex);
             }
             e = Expression.combine(ex, e); // combine in reverse order
+        }
+
+        /* extern(C++) destructors call into super to destruct the full hierarchy
+        */
+        ClassDeclaration cldec = ad.isClassDeclaration();
+        if (cldec && cldec.classKind == ClassKind.cpp && cldec.baseClass && cldec.baseClass.dtor)
+        {
+            // WAIT BUT: do I need to run `cldec.baseClass.dtor` semantic? would it have been run before?
+            cldec.baseClass.dtor.functionSemantic();
+
+            stc = mergeFuncAttrs(stc, cldec.baseClass.dtor);
+            if (!(stc & STC.disable))
+            {
+                // super.__xdtor()
+
+                Expression ex = new SuperExp(loc);
+
+                // This is a hack so we can call destructors on const/immutable objects.
+                // Do it as a type 'paint'.
+                ex = new CastExp(loc, ex, cldec.baseClass.type.mutableOf());
+                if (stc & STC.safe)
+                    stc = (stc & ~STC.safe) | STC.trusted;
+
+                ex = new DotVarExp(loc, ex, cldec.baseClass.dtor, false);
+                ex = new CallExp(loc, ex);
+
+                e = Expression.combine(e, ex); // super dtor last
+            }
         }
 
         /* Build our own "destructor" which executes e
--- a/src/dmd/dclass.d
+++ b/src/dmd/dclass.d
@@ -214,6 +214,9 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
     /// true if this is a scope class
     bool stack;
 
+    /// if this is a C++ class, this is the slot reserved for the virtual destructor
+    int cppDtorVtblIndex = -1;
+
     /// to prevent recursive attempts
     private bool inuse;
 
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -4783,6 +4783,8 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
                         // reserve the dtor slot for the destructor (which we'll create later)
                         cldec.cppDtorVtblIndex = cast(int)cldec.vtbl.dim;
                         cldec.vtbl.push(s);
+                        if (Target.cppDeletingDestructor)
+                            cldec.vtbl.push(s); // deleting destructor uses a second slot
                     }
                 }
             }
@@ -4878,6 +4880,13 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             // now we've built the aggregate destructor, we'll make it virtual and assign it to the reserved vtable slot
             cldec.dtor.vtblIndex = cldec.cppDtorVtblIndex;
             cldec.vtbl[cldec.cppDtorVtblIndex] = cldec.dtor;
+
+            if (Target.cppDeletingDestructor)
+            {
+                // TODO: create a C++ compatible deleting destructor (call out to `operator delete`)
+                //       for the mooment, we'll call the non-deleting destructor and leak
+                cldec.vtbl[cldec.cppDtorVtblIndex + 1] = cldec.dtor;
+            }
         }
 
         if (auto f = hasIdentityOpAssign(cldec, sc2))
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -3735,7 +3735,17 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         if (dd.ident == Id.dtor && dd.semanticRun < PASS.semantic)
             ad.dtors.push(dd);
         if (!dd.type)
-            dd.type = new TypeFunction(null, Type.tvoid, false, LINK.d, dd.storage_class);
+        {
+            Parameters* params = null;
+            if (ad.classKind == ClassKind.cpp && !Target.cppDeletingDestructor)
+            {
+                // Windows doesn't use a deleting destructor, it takes an argument instead!
+                Parameter param = new Parameter(STC.undefined_, Type.tuns32, new Identifier("del"), new IntegerExp(Loc.initial, 0, Type.tuns32));
+                params = new Parameters;
+                (*params).push(param);
+            }
+            dd.type = new TypeFunction(params, Type.tvoid, false, LINK.d, dd.storage_class);
+        }
 
         sc = sc.push();
         sc.stc &= ~STC.static_; // not a static destructor
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -4767,6 +4767,25 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         {
             auto s = (*cldec.members)[i];
             s.dsymbolSemantic(sc2);
+
+            if (cldec.classKind == ClassKind.cpp)
+            {
+                const DtorDeclaration dtor = s.isDtorDeclaration();
+                if (dtor && cldec.cppDtorVtblIndex == -1)
+                {
+                    if (cldec.baseClass && cldec.baseClass.cppDtorVtblIndex != -1)
+                    {
+                        // override the base virtual
+                        cldec.cppDtorVtblIndex = cldec.baseClass.cppDtorVtblIndex;
+                    }
+                    else if (!dtor.isFinal())
+                    {
+                        // reserve the dtor slot for the destructor (which we'll create later)
+                        cldec.cppDtorVtblIndex = cast(int)cldec.vtbl.dim;
+                        cldec.vtbl.push(s);
+                    }
+                }
+            }
         }
 
         if (!cldec.determineFields())
@@ -4853,6 +4872,13 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         }
 
         cldec.dtor = buildDtor(cldec, sc2);
+
+        if (cldec.classKind == ClassKind.cpp && cldec.cppDtorVtblIndex != -1)
+        {
+            // now we've built the aggregate destructor, we'll make it virtual and assign it to the reserved vtable slot
+            cldec.dtor.vtblIndex = cldec.cppDtorVtblIndex;
+            cldec.vtbl[cldec.cppDtorVtblIndex] = cldec.dtor;
+        }
 
         if (auto f = hasIdentityOpAssign(cldec, sc2))
         {
--- a/src/dmd/target.d
+++ b/src/dmd/target.d
@@ -60,6 +60,7 @@ struct Target
         bool cppExceptions;       /// set if catching C++ exceptions is supported
         char int64Mangle;         /// mangling character for C++ int64_t
         char uint64Mangle;        /// mangling character for C++ uint64_t
+        bool cppDeletingDestructor; /// target C++ ABI uses deleting destructor
     }
 
     /**
@@ -115,6 +116,9 @@ struct Target
         ptrsize = 4;
         classinfosize = 0x4C; // 76
 
+        // all systems but windows use deleting destructors
+        cppDeletingDestructor = true;
+
         /* gcc uses int.max for 32 bit compilations, and long.max for 64 bit ones.
          * Set to int.max for both, because the rest of the compiler cannot handle
          * 2^64-1 without some pervasive rework. The trouble is that much of the
@@ -148,6 +152,7 @@ struct Target
             realpad = 0;
             realalignsize = 2;
             reverseCppOverloads = true;
+            cppDeletingDestructor = false;
             c_longsize = 4;
             if (ptrsize == 4)
             {
--- a/test/runnable/cppa.d
+++ b/test/runnable/cppa.d
@@ -1265,6 +1265,49 @@ void test16536()
 }
 
 /****************************************/
+// 15589 - extern(C++) virtual destructors are not put in vtbl[]
+
+extern(C++)
+{
+    class A15589
+    {
+        extern(D) static int[] dtorSeq;
+        struct S
+        {
+            this(int x) { this.x = x; }
+            ~this() { dtorSeq ~= x; }
+            int x;
+        }
+        int foo() { return 100; } // shift dtor to slot 1
+        ~this() { dtorSeq ~= 10; }
+        S s1 = S(1);
+        S s2 = S(2);
+    }
+    class B15589 : A15589
+    {
+        int bar() { return 200;} // add an additional function AFTER the dtor at slot 2
+        ~this() { dtorSeq ~= 20; }
+        S s3 = S(3);
+    }
+
+    void test15589b(A15589 p);
+}
+
+void test15589()
+{
+    A15589 c = new B15589;
+    assert(A15589.dtorSeq == null);
+    assert(c.foo() == 100);
+    assert((cast(B15589)c).bar() == 200);
+    c.__xdtor(); // virtual dtor call
+    assert(A15589.dtorSeq[] == [ 20, 3, 10, 2, 1 ]); // destroyed full hierarchy!
+
+    A15589.dtorSeq = null;
+    test15589b(c);
+    assert(A15589.dtorSeq[] == [ 20, 3, 10, 2, 1 ]); // destroyed full hierarchy!
+}
+
+/****************************************/
 
 void main()
 {
@@ -1308,6 +1351,7 @@ void main()
     test15372();
     test15802();
     test16536();
+    test15589();
 
     printf("Success\n");
 }
--- a/src/dmd/aggregate.h
+++ b/src/dmd/aggregate.h
@@ -285,6 +285,7 @@ public:
     TypeInfoClassDeclaration *vclassinfo;       // the ClassInfo object for this ClassDeclaration
     bool com;                           // true if this is a COM class (meaning it derives from IUnknown)
     bool stack;                         // true if this is a scope class
+    int cppDtorVtblIndex;               // slot reserved for the virtual destructor [extern(C++)]
     bool inuse;                         // to prevent recursive attempts
     bool isActuallyAnonymous;           // true if this class has an identifier, but was originally declared anonymous
                                         // used in support of https://issues.dlang.org/show_bug.cgi?id=17371
--- a/src/dmd/cppmanglewin.d
+++ b/src/dmd/cppmanglewin.d
@@ -603,7 +603,7 @@ private:
             // <flags> ::= Y <calling convention flag>
             buf.writeByte('Y');
         }
-        const(char)* args = mangleFunctionType(cast(TypeFunction)d.type, d.needThis(), d.isCtorDeclaration() || d.isDtorDeclaration());
+        const(char)* args = mangleFunctionType(cast(TypeFunction)d.type, d.needThis(), d.isCtorDeclaration() || d.isDtorDeclaration(), d.isDtorDeclaration() !is null);
         buf.writestring(args);
     }
 
@@ -1144,7 +1144,7 @@ private:
         cur.accept(this);
     }
 
-    const(char)* mangleFunctionType(TypeFunction type, bool needthis = false, bool noreturn = false)
+    const(char)* mangleFunctionType(TypeFunction type, bool needthis = false, bool noreturn = false, bool noargs = false)
     {
         scope VisualCPPMangler tmp = new VisualCPPMangler(this);
         // Calling convention
@@ -1204,7 +1204,7 @@ private:
             rettype.accept(tmp);
             tmp.flags &= ~MANGLE_RETURN_TYPE;
         }
-        if (!type.parameters || !type.parameters.dim)
+        if (!type.parameters || !type.parameters.dim || noargs)
         {
             if (type.varargs == 1)
                 tmp.buf.writeByte('Z');
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -3372,8 +3372,9 @@ extern (C++) final class DtorDeclaration : FuncDeclaration
 
     override bool isVirtual() const
     {
-        // false so that dtor's don't get put into the vtbl[]
-        return false;
+        // D dtor's don't get put into the vtbl[]
+        // this is a hack so that extern(C++) destructors report as virtual, which are manually added to the vtable
+        return vtblIndex != -1;
     }
 
     override bool addPreInvariant()
--- a/src/dmd/target.h
+++ b/src/dmd/target.h
@@ -45,6 +45,7 @@ struct Target
     static bool cppExceptions;          // set if catching C++ exceptions is supported
     static char int64Mangle;            // mangling character for C++ int64_t
     static char uint64Mangle;           // mangling character for C++ uint64_t
+    static bool cppDeletingDestructor;  // target C++ ABI uses deleting destructor
 
     template <typename T>
     struct FPTypeProperties
--- a/test/runnable/extra-files/cppb.cpp
+++ b/test/runnable/extra-files/cppb.cpp
@@ -519,7 +519,7 @@ void test14200b(float a, int b, double c) {};
 
 namespace std {
     namespace N14956 {
-	struct S14956 { };
+    struct S14956 { };
     }
 }
 
@@ -762,7 +762,7 @@ int foo15372(int value)
 
 void test15372b()
 {
-	int t = foo15372<int>(1);
+    int t = foo15372<int>(1);
 }
 
 /****************************************/
@@ -793,7 +793,7 @@ public:
 
 void test15802b()
 {
-	int t = Foo15802<int>::boo(1);
+    int t = Foo15802<int>::boo(1);
 }
 
 
@@ -806,3 +806,34 @@ uint64_t pass16536(uint64_t a)
     return a;
 }
 #endif
+
+/****************************************/
+// 15589 - extern(C++) virtual destructors are not put in vtbl[]
+
+class A15589
+{
+public:
+    struct S
+    {
+    public:
+        int x;
+    };
+    virtual int foo();
+    virtual ~A15589();
+    S s1;
+    S s2;
+};
+class B15589 : public A15589
+{
+public:
+    virtual int bar();
+    virtual ~B15589();
+    S s3;
+};
+
+void test15589b(A15589 *p)
+{
+    assert(p->foo() == 100);
+    assert(((B15589*)p)->bar() == 200);
+    p->~A15589();
+}


src/dmd/dclass.d
```diff
@@ -233,6 +233,9 @@ extern (C++) class ClassDeclaration : AggregateDeclaration
     /// specifies whether this is a D, C++, Objective-C or anonymous class/interface
     ClassKind classKind;
 
+    /// if this is a C++ class, this is the slot reserved for the virtual destructor
+    int cppDtorVtblIndex = -1;
```
JinShil: I'm not sure about the approach in this PR, as I'm not all that familiar with the way vtables are built in D.  However, we nee to maintain both a C++ and D interface to DMD (unless I'm mistaken both LDC and GDC use the C++ interface).  If you're adding a field to `ClassDelcaration`, I think you'll also need to update the same class in `aggregate.h` to keep the C++ interface compatible.  cc @ibuclaw and @kinke.

src/dmd/dsymbolsem.d
```diff
@@ -4851,6 +4863,13 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
 
         cldec.dtor = buildDtor(cldec, sc2);
 
+        if (cldec.classKind == ClassKind.cpp && cldec.cppDtorVtblIndex != -1)
+        {
+            // now we've built the aggregate destructor, we'll make it virtual and assign it to the reserved vtable slot
+            cldec.dtor.vtblIndex = cldec.cppDtorVtblIndex;
```
jacob-carlborg: Why is `cppDtorVtblIndex` needed, why can't `cldec.dtor.vtblIndex` be used directly?

TurkeyMan:
I need to hold space for the v-destructor at the offset where I encounter __dtor for __xdtor which is generated much later... because __dtor is not the destructor.
It will also make more sense when I support inherentance to identify if the base has virtual dtor already.

I'm sure it's possible to write the code without it, but I think it will simplify the logic immensely when it's all done. I'll revise the value of it later?

src/dmd/dsymbolsem.d
TurkeyMan: I mean, you can see on this line that `cldec.dtor.vtblIndex` is being assigned :)
That function was only created on the line above, but its virtual slot was reserved way back when processing members...

jacob-carlborg:
Is this testable?

kinke:
Assuming we don't call that 2nd destructor from D, you may get away with a weak reference to the external C++ symbol, just to have the vtable on the D side complete in case that dtor is provided by C++ when linking, and otherwise resulting in a segfault if called by C++.

TurkeyMan:
"the external symbol", there's no reason to believe the external symbol exists though... of the derived class is in D, and C++ tries to destruct, then it shouldn't crash.

TurkeyMan:
I'm pretty sure the deleting destructor is all about `operator delete`. It is so that a derived class can override `operator delete`, and then the deleting destructor will call the correct delete, rather than the global one.
It's a shame that they conflated the deletion of memory into the destructor.

What I think is safe to say though, is that we won't support `operator delete` (at least, not eagerly), and that no such mischief should appear in our extern(C++) hierarchies. Only the default deleting destructor, which presumably calls global operator delete... which we definitely can extern to.

kinke:
> I'm pretty sure the deleting destructor is all about operator delete as class members. It is so that a derived class can override operator delete, and then the deleting destructor will call the correct delete, rather than the global one.

Apparently so: https://eli.thegreenplace.net/2015/c-deleting-destructors-and-virtual-operator-delete/

As custom [class (de)allocators](https://dlang.org/deprecate.html#Class%20allocators%20and%20deallocators) are deprecated in D, a custom C++ delete operator cannot be modeled in D anymore (I don't think that's an issue though, and the signature wouldn't match either). So calling the global delete operator in the deleting dtor (for C++ `delete obj` statements) seems reasonable. C++-deleting is undefined if the object hasn't been allocated via C++ `new` as well (see point 1 in http://en.cppreference.com/w/cpp/memory/new/operator_delete), so one needs to be careful anyway.

src/dmd/dclass.d
TurkeyMan: updated it.

src/dmd/dsymbolsem.d
```diff
@@ -4764,6 +4764,29 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         {
             auto s = (*cldec.members)[i];
             s.dsymbolSemantic(sc2);
+
+            if (cldec.classKind == ClassKind.cpp)
+            {
+                DtorDeclaration dtor = s.isDtorDeclaration();
```
jacob-carlborg: Can this variable be `const`?

TurkeyMan: probably... is this another case where i'm expected to write code differently than all the surrounding code, and I'm just meant to know that? :P

wilzbach:
> So, at this point, the code successfully destructs a full hierarchy from a base class, including when the hierarchy is mixed languages!

Test or it didn't happen! 😉

TurkeyMan:
Maybe the order isn't quite right, but it's all there!

src/dmd/clone.d
```diff
@@ -875,6 +876,34 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
             e = Expression.combine(ex, e); // combine in reverse order
         }
 
+        /* extern(C++) destructors call into super to destruct the full hierarchy
+        */
+        ClassDeclaration cldec = ad.isClassDeclaration();
+        if (cldec && cldec.classKind == ClassKind.cpp && cldec.baseClass && cldec.baseClass.dtor)
+        {
+            // WAIT BUT: do I need to run `cldec.baseClass.dtor` semantic? would it have been run before?
+            cldec.baseClass.dtor.functionSemantic();
```
TurkeyMan: This line.... do I need it?
When I step in there, it looks like it's always done it before.

src/dmd/dsymbolsem.d
jacob-carlborg: Yes :smiley:.

test/runnable/cppa.d
```diff
@@ -1264,6 +1264,34 @@ void test16536()
     version(OSX) assert(pass16536(123) == 123);
 }
 
+/****************************************/
+// 15589 - extern(C++) virtual destructors are not put in vtbl[]
+
+extern(C++) {
```
jacob-carlborg: Curly braces on their own lines.

TurkeyMan: Like the test immediately above... ;)

jacob-carlborg: Yes, like the rest of the file :smiley:.

TurkeyMan: Except the one immediately above, which i used as reference..

jacob-carlborg: Ah, you mean the test for 15802? Half of that is using the correct style, the other half is not. Then there's the test for 16536 in between, as far as I can see.

src/dmd/dsymbolsem.d
```diff
@@ -4851,6 +4874,92 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+//                    if (!cpp_delete)
+//                    {
+//                        auto parameters = new Parameters();
+//                        parameters.push(new Parameter(0, Type.tvoidptr, Id.p, null));
+//                        auto tf = new TypeFunction(parameters, Type.tvoid, 0, LINK.cpp);
+//                        cpp_delete = new FuncDeclaration(Loc.initial, Loc.initial, Id.__cpp_operator_delete, STC.static_ | STC.extern_ | STC.nothrow_ | STC.trusted | STC.nogc, tf);
+//
+////                        cldec.dtors.push(deldtor);
+////                        cldec.members.push(deldtor);
+//                        cpp_delete.dsymbolSemantic(sc2);
```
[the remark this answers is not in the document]
jacob-carlborg: As far as I know there is no global scope as you're describing it.

Not sure if there's a better way but you can either find the `object` module in the array of modules [1]. Or you can store the `object` module in a new variable when that is first seen by the compiler, similar to ho `ClassDeclaration` is doing with some classes in the `object` module [2].

[1] https://github.com/dlang/dmd/blob/b67227300b1e90f6e348e65f2a5f5cb0cb2e4464/src/dmd/dmodule.d#L289

[2] https://github.com/dlang/dmd/blob/b67227300b1e90f6e348e65f2a5f5cb0cb2e4464/src/dmd/dclass.d#L397-L427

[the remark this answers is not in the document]
TurkeyMan: Is `object` available in betterc?
I feel like this stuff shouldn't depend on the D runtime at all.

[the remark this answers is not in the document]
JinShil: > Is object available in betterc?

Not it isn't.  object.d is imported in -betterC, but the druntime is never linked, so things like templates work, but any runtime implementation will not.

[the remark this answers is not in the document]
jacob-carlborg: Perhaps I'm completely wrong here but can you generate one function per module, put it in a COMDAT section and the linker will just pick one [1]? Perhaps you can even use what's mention in the reply of [1]: "make sure all dups. have the same content".

[1] https://stackoverflow.com/questions/1834597/what-is-the-comdat-section-used-for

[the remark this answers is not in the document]
TurkeyMan: I'm not generating any function, only an extern declaration to the C++ function.
It's also a global function (in C++ terms), it feels really weird to emit one symbol per module. Module's are already cluttered, and I do stuff like __traits(allMembers) on modules all the time. I don't wanna start seeing `operator delete` appearing in there.

src/dmd/dsymbolsem.d
```diff
@@ -4851,6 +4874,92 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                    deldtor.generated = true;
+                    deldtor.storage_class |= STC.inference;
+                    deldtor.fbody = new ExpStatement(dtloc, dtor_exp);
+//                    deldtor.fbody = new ExpStatement(dtloc, Expression.combine(dtor_exp, delete_exp));
+                }
+
+                // add it to the class and run semantic
+                cldec.dtors.push(deldtor);
+                cldec.members.push(deldtor);
+                deldtor.dsymbolSemantic(sc2);
```
TurkeyMan: Just FYI, someone might have more better idea here...  
`dsymbolSemantic` goes all wrong if I give it any hint that this function is virtual/override when it runs. It can't deal with reserved vtable slots like this here. The simplest solution seemed to fix up the relevant state (immediately below) after the function processes as if a normal non-virtual method. I didn't want to mess with the chaos in `dsymbolSemantic` to try and teach it about reserved vtable slots. And that would also add extra state to node structures too... so this hack(-ish) just below... looks okay?

src/dmd/dsymbolsem.d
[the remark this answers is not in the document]
jacob-carlborg: This code specifies which C function to call for an Objective-C method call, most likely `objc_msgSend` [1]. Te point here is that `objc_msgSend` is never declared. But this is at a later stage in the compilation process.

https://github.com/dlang/dmd/blob/master/src/dmd/e2ir.d#L339-L342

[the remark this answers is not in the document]
TurkeyMan: @JinShil it just feels like literally the only place that system level stuff like this belongs... like, this is related to implementation of classes.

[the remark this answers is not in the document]
jacob-carlborg: What about using this function [1], can that work?

[1] https://github.com/dlang/dmd/blob/master/src/dmd/func.d#L2295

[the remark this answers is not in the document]
jacob-carlborg: @TurkeyMan found an example, at a higher level, where it calls `__cxa_end_catch` [1].

[1] https://github.com/dlang/dmd/blob/master/src/dmd/s2ir.d#L1176-L1191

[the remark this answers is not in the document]
jacob-carlborg: I made a quick test, I replaced this line [1] with:

```d
cldec.dtor = FuncDeclaration.genCfunc(null, Type.tvoid, "__cxa_end_catch");
```

The compiler doesn't complain about missing symbols and if I provide a C function with that name it correctly calls it when the destructor of the class is run.

[1] https://github.com/dlang/dmd/blob/master/src/dmd/dsymbolsem.d#L4852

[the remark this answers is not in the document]
kinke: As Jacob pointed out, static `FuncDeclaration.genCfunc()` is used for similar C function declarations, so adding a similar `genCppFunc()` (or passing the desired function linkage as optional `genCfunc` parameter) should do.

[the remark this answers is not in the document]
TurkeyMan: Ah, I had no idea that existed. Easy! Thanks!

"and if I provide a C function with that name" <-- and if you *don't* provide a C function with that name? Weak stub,  or link error? I don't really want a link error when someone makes an extern(C++) class and _doesn't_ link to C++.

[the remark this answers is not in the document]
jacob-carlborg: Currently it will generate a link error. What do you want the behavior to be? Can it not be generated only when necessary?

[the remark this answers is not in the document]
kinke: Indeed. A weak stub seems necessary and sadly can't live in druntime if `-betterC` is to be supported. But you may opt to postpone that problem, it doesn't say `-betterCpp` after all. ;)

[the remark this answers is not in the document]
TurkeyMan: I'm feeling like I might make the deleting destructor leak, and I'll follow up in another PR...
That puts us in an infinitely better position than we're already in.

This PR is getting overwhelmed by detail relating to calling `operator delete`, which won't be used 99.99% of the time.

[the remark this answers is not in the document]
jacob-carlborg: Smaller PRs are better so I'm all for that.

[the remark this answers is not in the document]
TurkeyMan: kk, i'll strip this noise out.

TurkeyMan:
Okay, I removed the deleting destructor complexity.
Code populated dtor slots, virtual destruction works.
This patch can be reviewed/merged I think.

TurkeyMan:
So, how's this one? Is it good?

jacob-carlborg:
I don’t feel l'm that familiar with C++ to review this.

src/dmd/target.d
```diff
@@ -60,6 +60,7 @@ struct Target
         bool cppExceptions;       /// set if catching C++ exceptions is supported
         char int64Mangle;         /// mangling character for C++ int64_t
         char uint64Mangle;        /// mangling character for C++ uint64_t
+        bool deletingDestructor;  /// target ABI uses deleting destructor
```
klickverbot: This should mention C++, probably even in the member name.

TurkeyMan: Haha, I just remembered why I didn't call it `cppDeletingDestructor` (which was the original name), it's because adding 3 characters caused the commend to move to the right... and then there was the terrible tension; do I move ALL the comments to the right to maintain the line-up, or do I let my comment be the only one out of line? ;)

TurkeyMan: I opted for the conservative resolution of; use a shorter variable name...

TurkeyMan:
So is there any more I need to do on this one? I think it's good for the now.
The implementation of the deleting destructor has some complex (possibly controversial) details (purely extern? weak local implementation? which scope should it live in?), so I'll address it in a follow-up.

jacob-carlborg:
Ping. Someone feeling comfortable about merging this?

src/dmd/dsymbolsem.d
```diff
@@ -4764,6 +4764,29 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         {
             auto s = (*cldec.members)[i];
             s.dsymbolSemantic(sc2);
+
+            if (cldec.classKind == ClassKind.cpp)
+            {
+                const DtorDeclaration dtor = s.isDtorDeclaration();
+                if (dtor && !dtor.isFinal())
```
rainers: A final function can still override a base class implementation.
It seems you cannot override non-virtual destructors anyway, so it should also check for the existence of a base class destructor?

src/dmd/dsymbolsem.d
```diff
@@ -4764,6 +4764,29 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                {
+                    assert(cldec.cppDtorVtblIndex == -1); // how does a class declare 2 destructors?
```
rainers: This code can get executed again if semantic was deferred due to forward references (search for `addDeferredSemantic`). Instead of an assert, you could use the state `cppDtorVtblIndex` to avoid duplicate vtbl entries.

test/runnable/cppa.d
```diff
@@ -1264,6 +1264,39 @@ void test16536()
...
+        }
+        ~this() { dtorSeq ~= 10; }
+        S s1 = S(1);
+        S s2 = S(2);
+    }
+    class B15589 : A15589
+    {
+        ~this() { dtorSeq ~= 20; }
+        S s3 = S(3);
+    }
```
rainers: Would be good to test whether C++ and D agree on the position of the dtor in the vtable. Can you add some virtual functions, matching declarations in cppb.cpp and call the dtor from there?

TurkeyMan:
@rainers I think I addressed all that feedback...

TurkeyMan:
There's a test failure because DMC doesn't compile C++ properly :/

12345swordy:
/cc @WalterBright 
Can you help TurkeyMan on this @WalterBright?

rainers:
> There's a test failure because DMC doesn't compile C++ properly :/

The test also fails for VC on 32-bit. Appveyor performs two builds:
- builds the 64-bit compiler with dmd and runs tests with -m64
- builds the 32-bit compiler with LDC and runs tests with -m32mscoff

rainers:
LGTM but the failing test

TurkeyMan:
I'm trying to work out how that extra 4 bytes of stack is ingested in a normal C++ program, but the C++ destructor doesn't seem to do anything with it, then I spotted this:
```asm
Fooo::`vector deleting destructor':   ; <--- C++ destructor
0104FF10  push        ebp             ; <--- function entry
0104FF11  mov         ebp,esp  

...        ; function code here, leaves the stack exactly how it found it

0104FF64  mov         esp,ebp         ; <--- function exit  
0104FF66  pop         ebp  
0104FF67  ret         4               ; <--- what is this? O_O
```
I noticed that the return statement doesn't just fetch 4 bytes from the stack and return, the C++ destructor consumes 8 bytes when it returns (which consumes the 4 bytes the calling code pushed).

Looking at the D destructor:
```asm
B15589::~B15589:    ; <--- extern(C++) destructor
0102B128  enter       4,0  

...   ; destructor code, doesn't upset the stack

0102B13F  leave  
0102B140  ret       ; <--- there's no `4` after the ret...
```
So... I didn't know that x86 `ret` took an integer argument instructing to pop extra bytes from the stack when it returns. DMD doesn't add that, and the stack is out by 4 bytes.

I have no idea why VC does this. It seems when calling destructors (not other functions) that it pushes 4 extra bytes, and then the destructor pops it when it returns.

Perhaps the solution goes: calling `extern(C++)` `DtorDeclaration` functions should `ret 4` when they return, and also call with a bonus `push 0` immediately before the call.

I don't know how to approach this in DMD. I'd really appreciate if someone who understands the DMD backend or calling convention code could add a patch to my PR here to address this detail?

TurkeyMan:
Checking x64 to understand differences, and why it doesn't need to do this, I spotted this:
```asm
	p->~A15589();
00007FF7CC3B6622  mov         rax,qword ptr [p]  
00007FF7CC3B6629  mov         rax,qword ptr [rax]  
00007FF7CC3B662C  xor         edx,edx      ; <--- destructor doesn't receive any arguments!
00007FF7CC3B662E  mov         rcx,qword ptr [p]  
00007FF7CC3B6635  call        qword ptr [rax+8]   ; <--- call virtual dtor
```
I noticed that when x64 code calls destructors, it sets EDX (not RDX!) to zero immediately before the call. I guess this parallel's the `push 0` in x86.
Just like x86, C++ destructors completely ignore it, and since it's not on the stack, there's no crash!

rainers:
If I dump an empy detructor I get:
```
??_GCC@@QEAAPEAXI@Z (public: void * __cdecl CC::`scalar deleting destructor'(unsigned int)):
  0000000000000000: 89 54 24 10        mov         dword ptr [rsp+10h],edx
  0000000000000004: 48 89 4C 24 08     mov         qword ptr [rsp+8],rcx
  0000000000000009: 48 83 EC 28        sub         rsp,28h
  000000000000000D: 48 8B 4C 24 30     mov         rcx,qword ptr [rsp+30h]
  0000000000000012: E8 00 00 00 00     call        ??1CC@@QEAA@XZ
  0000000000000017: 8B 44 24 38        mov         eax,dword ptr [rsp+38h]
  000000000000001B: 83 E0 01           and         eax,1
  000000000000001E: 85 C0              test        eax,eax
  0000000000000020: 74 0F              je          0000000000000031
  0000000000000022: BA 01 00 00 00     mov         edx,1
  0000000000000027: 48 8B 4C 24 30     mov         rcx,qword ptr [rsp+30h]
  000000000000002C: E8 00 00 00 00     call        ??3@YAXPEAX_K@Z
  0000000000000031: 48 8B 44 24 30     mov         rax,qword ptr [rsp+30h]
  0000000000000036: 48 83 C4 28        add         rsp,28h
  000000000000003A: C3                 ret
```
Which shows in the demangled name that the destructor receives an argument. This is then used to determine whether `??3@YAXPEAX_K@Z` (`void __cdecl operator delete(void * __ptr64,unsigned __int64)`) is called. So this seems VCs version of the deleting destructor.
Calling the destructor explicitly via `obj->~Classname()` passes 0, `delete obj` passes 1 as the argument.

TurkeyMan:
Well, the code you pasted doesn't quite read that way...
```asm
  000000000000001E: 85 C0              test        eax,eax   ; <- (true) unconditionally jump
  0000000000000020: 74 0F              je          00000031  ; <- jump over delete
  0000000000000022: BA 01 00 00 00     mov         edx,1     ; <- overwrite EDX with 1
  0000000000000027: 48 8B 4C 24 30     mov         rcx,qword ptr [rsp+30h]
  000000000000002C: E8 00 00 00 00     call        ??3@YAXPEAX_K@Z ; <- can't get called
```
That destructor you pasted tests `eax == eax` (true), then unconditionally jumps over that whole block.
Then, within that block (which it can't enter under any condition), it assigns 1 to `EDX`.

The value of `EDX` that the caller passed in is never read anywhere inside that function.
There must be a construction where it isn't ignored though. I just don't know how to invoke it.

Safe to say, `extern(C++)` destructors from D don't need to handle this case, since we have simple class hierarchy and no `operator delete` overloading. So we can ignore the argument, but we must still respect the calling convention...

rainers:
> That destructor you pasted tests eax == eax (true)

It tests whether eax != 0, with eax being read from the lowest bit of the argument passed to the destructor.

rainers:
> The value of EDX that the caller passed in is never read anywhere inside that function.

It's saved to [rsp+10h] and then read back via [rsp+38h] later after adjusting rsp.

edx = 1 passed to `??3@YAXPEAX_K@Z` is the size argument for global delete, probably used as "unknown size".

I think the destructor should be declared as receiving an `int` argument, and always be called with 0 to avoid memory deallocation for now (similar to the posix situation without the second dtor).

TurkeyMan:
`test eax,eax` is `eax == eax`; unconditional true?

rainers:
> test eax,eax is eax == eax; unconditional true?

No, `test` is an `and` operation that throws away the result but keeps the flags.

TurkeyMan:
Oh right, I didn't spot it written to the stack on the first mine! >_<

> I think the destructor should be declared as receiving an int argument, and always be called with 0 to avoid memory deallocation for now (similar to the posix situation without the second dtor).

That's a very idea... although that feels super-hacky... I'm not sure how to make all calls to destructors magically call with a zero argument for windows `extern(C++)`...

TurkeyMan:
Oh right! `test`, not `cmp` >_<
It's been so long since I had to worry about x86 asm!

rainers:
> That's a very idea... although that feels super-hacky... I'm not sure how to make all calls to destructors magically call with a zero argument for windows extern(C++)...

Yeah, it gets a bit more complicated. I think what could work is to add the destructor with the extra argument to the vtbl (it should forward the argument to the base class), but use a stub for `xdtor` that calls it with argument `0`.

TurkeyMan:
But yeah, it's all good. That's the right approach.
I'll find a way to do it without too much ugly.

TurkeyMan:
Okay, I pushed a patch that works for classes. I made the dtor receive a `uint del = 0` default arg, so it automatically passes everywhere it's called.

struct dtor's also need this though, and I can't work out how to detect if a struct is `extern(C++)`... can anyone give me any hints?
The only way to know appears to be if they're declared in an `extern(C++)` scope, and I don't know how to find that given a `StructDeclaration`...?

src/dmd/dsymbolsem.d
```diff
@@ -3735,7 +3735,18 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         if (dd.ident == Id.dtor && dd.semanticRun < PASS.semantic)
             ad.dtors.push(dd);
         if (!dd.type)
-            dd.type = new TypeFunction(null, Type.tvoid, false, LINK.d, dd.storage_class);
+        {
+            Parameters* params = null;
+            ClassDeclaration cd = ad.isClassDeclaration();
+            if (cd && cd.classKind == ClassKind.cpp && !Target.cppDeletingDestructor)
```
TurkeyMan: How can I also detect if `ad` is an `extern(C++) struct`??

jacob-carlborg: Perhaps add a `linkage` instance variable to `StructDeclaration` just as it exists for `Declaration` and `classKind` exists for `ClassDeclaration`.

TurkeyMan:
I have no idea what's going wrong with the tests now! >_<

TurkeyMan:
Okay, I couldn't get away without correctly detecting struct linkage.
I had to hoist `ClassKind` to `AggregateDeclaration` so it applies to struct's too.

This has actually been a point of friction for other tasks too, so some follow-up PR's can come from this change.

TurkeyMan:
What's the deal with semaphoreci? It doesn't say anything in the log; I don't know how to reproduce it... :/

JinShil:
image

... but I don't know what the problem is.  May be just a quirk.

TurkeyMan:
On the plus side, this finally works properly! :)

TurkeyMan:
I rearranged to sit in top of #8336.

test/compilable/test17419.d
```diff
@@ -39,11 +39,13 @@ void bar()
 class FooD {}
 interface FooDInterface {}
 extern (C++) class FooCpp {}
+extern (C++) struct FooCppStruct {}
 extern (C++) interface FooCppInterface {}
 
 static assert(__traits(getLinkage, FooD) == "D");
 static assert(__traits(getLinkage, FooDInterface) == "D");
 static assert(__traits(getLinkage, FooCpp) == "C++");
+static assert(__traits(getLinkage, FooCppStruct) == "C++");
```
jacob-carlborg: When this now works for struct, the documentation should be updated [1].

[1] https://dlang.org/spec/traits.html#getLinkage

TurkeyMan:
Rebased on master to reduce the diff footprint since the parent PR was merged.
All green, finally! ;)

src/dmd/dsymbolsem.d
```diff
@@ -3735,7 +3735,17 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
         if (dd.ident == Id.dtor && dd.semanticRun < PASS.semantic)
             ad.dtors.push(dd);
         if (!dd.type)
-            dd.type = new TypeFunction(null, Type.tvoid, false, LINK.d, dd.storage_class);
+        {
+            Parameters* params = null;
+            if (ad.classKind == ClassKind.cpp && !Target.cppDeletingDestructor)
+            {
+                // Windows doesn't use a deleting destructor, it takes an argument instead!
+                Parameter param = new Parameter(STC.undefined_, Type.tuns32, new Identifier("del"), new IntegerExp(Loc.initial, 0, Type.tuns32));
```
rainers: I don't think it can work so easily (AFAIU you are adding a parameter with a default value so it can still be called without arguments). This helps for explicit calls, but not for the dtor written to the TypeInfo. The latter needs a shim that adds the default value as an argument. Here's an example that calls the dtor this way:
```
import core.stdc.stdio;

extern(C++) 
struct S
{
	int x;
	~this() { printf("~S %x\n", x); }
}

void main()
{
	S[] arr = new S[3];
	arr[1].x = 1;
	arr[2].x = 2;
	arr.length = 1;
	assumeSafeAppend(arr); // destroys arr[1] and arr[2]
	printf("done\n");
}
```
Output is:
```
~S 2
~S 1
done
~S 0
```
Works for Win64, but not for Win32 in dmd 2.080. (Falsely assumes that C++ dtor does not have an argument, though).

src/dmd/clone.d
```diff
@@ -883,6 +884,34 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
...
+                Expression ex = new SuperExp(loc);
+
+                // This is a hack so we can call destructors on const/immutable objects.
+                // Do it as a type 'paint'.
+                ex = new CastExp(loc, ex, cldec.baseClass.type.mutableOf());
+                if (stc & STC.safe)
+                    stc = (stc & ~STC.safe) | STC.trusted;
+
+                ex = new DotVarExp(loc, ex, cldec.baseClass.dtor, false);
+                ex = new CallExp(loc, ex);
```
rainers: Would be nice if you could forward the dtor argument here, so it doesn't get lost when `delete` is called from C++ and the base class is also in C++ and handles it as expected.

src/dmd/dsymbolsem.d
TurkeyMan: Oh no, you're killing me bro! >_<

src/dmd/clone.d
TurkeyMan: If C++ ever calls a destructor with non-0, it means an operator delete override exists beneath it in the hierarchy, and therefore is not supported by D.
There is no case where it's valid that any argument other than 0 is used, and D is also involved. I dont think 'we can support this, so I'm not sure fixing this is worth the complexity no?

TurkeyMan: Also, I think it's only the top-level destructor that calls delete, you don't pass the request to delete down the hierarchy, otherwise it would call delete at each level of destruction... only the top-level destruction should finally delete.

src/dmd/dsymbolsem.d
TurkeyMan: Here I was thinking this was a super clever and elegant solution :(

TurkeyMan: Perhaps I could address this in a follow-up? Situation in this PR is immeasurably better than what we had before, and I have people here that want to be using this code already.
Obviously, I'm happy to continue to work on this issue until it's correct.

rainers: Maybe druntime can handle that case, too. Any user of the dtor member of TypeInfo_Struct/Class needs to know about it then, though.

rainers: > Perhaps I could address this in a follow-up?

Sounds reasonable.

rainers (approved):
Mixing C++ and D constructors will probably still have issues (e.g. mangling), but this is already a step forward.

src/dmd/clone.d
TurkeyMan: Right. If we add the enhancement to support calling C++ delete, then we should be semantically compatible I think.

TurkeyMan:
Mixing constructors is broken right now, because there's a semantic difference with vtable assignment.
That's a further development which I haven't begun yet. This destructor case was more important (and also simpler).
Constructors are working fine right now when there is no mixed-hierarchy, and C++ deriving from D works fine too. D deriving from C++ doesn't work because of vtable assignment ordering.

src/dmd/dsymbolsem.d
TurkeyMan: Can you put this snippet in an issue?
I'm not sure if I'll work on this case, or the constructor issue first...

TurkeyMan:
> (e.g. mangling)

I'm not aware of a constructor mangling issue though... what do you have in mind?

Geod24:
@TurkeyMan : Closing/Reopening to re-trigger a build. Since semaphoreci was disabled, I think it just got stuck...

src/dmd/dsymbolsem.d
rainers: > Why is it that druntime isn't calling the destructor with the default arg supplied like everywhere else?

Because it uses a function pointer assuming it does not need an extra argument. See
https://github.com/dlang/druntime/blob/master/src/rt/lifetime.d#L1347
https://github.com/dlang/druntime/blob/master/src/object.d#L2010

> Can you put this snippet in an issue?

https://issues.dlang.org/show_bug.cgi?id=18953

I just realized that calling the C++ dtor is pretty much broken for Win32 anyway because it doesn't use the __thiscall ABI.

TurkeyMan: We'll need that hack you suggest to generate a shim when assigning to a delegate I guess.

rainers:
> I'm not aware of a constructor mangling issue though... what do you have in mind?

Sorry, I meant destructors. These are currently mangled without arguments AFAICT, but have one now.

TurkeyMan:
I fixed the mangling issue. It's in this PR.
Look for `noargs`.

TurkeyMan:
There are unit-tests that prove mangled ctor/dtor names from a prior PR too.

rainers:
> I fixed the mangling issue. It's in this PR. Look for noargs.

That's what I meant: the mangling is enforced to have no argument, but that's not what happens in C++. The function in the vtable also demangles having the extra argument, but the dtor used for chaining to the base class does not contain it. 
I suspect that the dtor in the TypeInfo would have been fine before adding that extra argument (but the calling convention for Win32), and that the vtable entry needs the shim function.

TurkeyMan:
The C++ mangling has no argument, even though the function DOES have an argument.
The function demangles to show the argument, but that argument is not actually present in the mangled name in any of my tests. I think the demangler just shows it to be helpful.
I never observed a destructor with the argument mangled into the name in any C++ test, even though the function was definitely called with the argument...

Geod24:
Since auto-merge doesn't seem to have effect, and this has been approved, doing it manually.

PR #8278, opened 2018-05-22
dtemplate.d: add some inout and const to utility functions

WalterBright:


src/dmd/dtemplate.d
```diff
@@ -186,16 +179,15 @@ extern (C++) Dsymbol getDsymbol(RootObject oarg)
 
 private Expression getValue(ref Dsymbol s)
 {
-    Expression e = null;
     if (s)
```
jacob-carlborg: This could do an early return and save one indentation.

jacob-carlborg (approved):

PR #8274, opened 2018-05-21
Fix issue 18890 - extern(C++) mangles all destructors the same

TurkeyMan:
With `extern(C++)`, classes with multiple destructors were all mangling the same.
This patch only mangles the full-destructor as extern(C++), which is the behaviour C++ expects.

PR #8280, opened 2018-05-22
isPrimaryDtor: add const

WalterBright:
A followup to https://github.com/dlang/dmd/pull/8274

1. use Ddoc style for new functions
2. use const wherever possible
3. use isMember() to find the enclosing aggregate (just using .parent falls afoul of template mixins)

### DIFF
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -65,16 +65,21 @@ const(char)* cppTypeInfoMangleItanium(Dsymbol s)
 }
 
 /******************************
-* Determine if a symbol is the 'primary' destructor, that is, the most-aggregate destructor (the one that is defined as __xdtor)
-*/
-package bool isPrimaryDtor(Dsymbol sym)
+ * Determine if sym is the 'primary' destructor, that is,
+ * the most-aggregate destructor (the one that is defined as __xdtor)
+ * Params:
+ *      sym = Dsymbol
+ * Returns:
+ *      true if sym is the primary destructor for an aggregate
+ */
+bool isPrimaryDtor(const Dsymbol sym)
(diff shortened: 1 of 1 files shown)


src/dmd/cppmangle.d
```diff
@@ -65,16 +65,21 @@ const(char)* cppTypeInfoMangleItanium(Dsymbol s)
...
-*/
-package bool isPrimaryDtor(Dsymbol sym)
+ * Determine if sym is the 'primary' destructor, that is,
+ * the most-aggregate destructor (the one that is defined as __xdtor)
+ * Params:
+ *      sym = Dsymbol
+ * Returns:
+ *      true if sym is the primary destructor for an aggregate
+ */
+bool isPrimaryDtor(const Dsymbol sym)
```
JinShil: I'd like to know your justification for making this `public` rather than `package` so I can apply such consideration in future PR reviews.  I would have expected that we'd prefer `package` over `public`;  that is, we restrict access as much as possible until justification is provided to expose a symbol more broadly.  In other words, the burden of proof is on the desire to expose access, not restrict access.  It's always possible to expose access to a symbol, but once it's exposed, it's a breaking change to restrict access.

jacob-carlborg (approved):
LGTM except making the symbol public.

src/dmd/cppmangle.d
WalterBright: We don't use packages in dmd.

JinShil: Actually we do.  `dmd` is a package, and if we don't take care about how much we expose publicly, we'll just be creating unnecessary restrictions and maintenance for ourselves.

PR #8281, opened 2018-05-23
[perf] Don't allocate argument mismatch string when errors gagged

ntrel:
@Geod24 This should fix the high memory use found [here](https://github.com/dlang/dmd/pull/7554#issuecomment-386625022) (Note: #8232 already helped this somewhat).


RazvanN7 (approved):

Geod24:
@ntrel : Thanks for taking care of this! Glad you could come up with a solution. I didn't have much time to look into D since DConf, but I'll try to build my project from `stable` this weekend and see if memory usage has returned to normal.

PR #8282, opened 2018-05-23
Fix Issue 18892 -  Wrong type in error message for static members and alias this

RazvanN7:


src/dmd/expressionsem.d
```diff
@@ -9510,7 +9510,7 @@ Expression semanticX(DotIdExp exp, Scope* sc)
 
 // Resolve e1.ident without seeing UFCS.
 // If flag == 1, stop "not a property" error and return NULL.
-Expression semanticY(DotIdExp exp, Scope* sc, int flag)
+Expression semanticY(DotIdExp exp, Scope* sc, int flag, bool calledForAliasThis = false)
```
jacob-carlborg: Walter won’t like this optional argument.

RazvanN7: I haven't found a better way to signal that the function was called from resolveAliasThis

JinShil:
No test?

RazvanN7:
@JinShil Forgot to git add it.

src/dmd/expressionsem.d
```diff
@@ -9788,7 +9788,9 @@ Expression semanticY(DotIdExp exp, Scope* sc, int flag)
     }
     else
     {
-        if (exp.e1.op == TOK.type || exp.e1.op == TOK.template_)
+        if (calledForAliasThis)
```
JinShil: It appears we first need and `AggregateDeclaration` and then it's `aliasthis` property.  Can't the `AggregateDeclaration` be determined from `exp` or `sc` to avoid having to pass in the `calledForAliasThis` parameter.  Looking at the other code in this function, it looks like we just need to find a way to obtain the `AggreegateDeclaration` from the existing parameters.

RazvanN7: I don't think the aggregate declaration can be obtained from the parameters. Consider that the DotIdExp passed to semanticY is the one which is constructed **with** the resolved aliasthis as it's `lhs` and the `sc` parameter represents the scope of the DotIdExp. The only part of the code which links the expression to the aggregate declaration is the one present in typesem.d where the aliasthis is resolved and then tried by calling semanticY

JinShil: Ok, then what about setting `flag` before calling `semanticY`?  And is there a constant defined somewhere so we don't have to use the magic number `1`?

RazvanN7: No there isn't. The dmd code that uses flag checks for it like this : `if (flag & 1)`, even the documentation says : `if flag == 1, stop error and return null` .

RazvanN7:
@JinShil Can you take another look at it? I managed to eliminate the extra parameter

src/dmd/typesem.d
```diff
@@ -3066,7 +3066,10 @@ private extern(C++) final class DotExpVisitor : Visitor
                     return returnExp(null);
 
                 auto die = new DotIdExp(e.loc, alias_e, ident);
-                auto exp = die.semanticY(sc, gagError);
+                /* second argument of semanticY is 2 to signal
+                 * that it was called for an aliasThis
+                 */
+                auto exp = die.semanticY(sc, 2);
```
JinShil: Why can't we just past `1` here to get the same effect?

RazvanN7: Because the type check in expressionsem may or may not be passed (depending on the type of the alias this) and passing 1 does not affect it anyhow.

RazvanN7: The fix is to pass 3, because that way we make sure that the first bit is set for other branches.

RazvanN7: It's ugly, but at least we get rid of the extra parameter.

JinShil:
I submitted an alternative fix at https://github.com/dlang/dmd/pull/8303

PR #7721, opened 2018-01-16
Remove Travis testing

wilzbach:
- Travis got super slow and unreliable over the last weeks
- Even reduced to [only one OSX job](https://github.com/dlang/dmd/pull/7653) it still takes hours for a PR to build
- We often run into failures as the job reaches the maximal 50 minutes build limit
- Travis was only intended to test bootstrapping with the latest DMD + GDC + LDC (SemaphoreCI does this [now pretty reliable](https://github.com/dlang/dmd/pull/7617#issuecomment-357567565).
- If https://github.com/dlang-bots/dlang-bot/pull/69 gets merged, we would need to wait for hours or even days until a PR gets merged (even if all other CI pass)

Or in Martin's word:

> What's the worth of a CI that doesn't report back in 1 day.

See also:
- https://github.com/dlang-bots/dlang-bot/pull/69#issuecomment-357854277
- https://github.com/dlang/dmd/pull/7617#issuecomment-357567938
- https://travis-ci.org/dlang/dmd/builds
- http://forum.dlang.org/post/ronypzvsafxpqlpnxlsf@forum.dlang.org (Heads-up for pending auto-merge change)
- http://forum.dlang.org/post/zbxaviowooiaenaawmgs@forum.dlang.org (Re: Bump the minimal version required to compile DMD to 2.076.1)

CC @MartinNowak

PR #8285, opened 2018-05-23
Revert mangling of D `long` back to C++ `long` on 64-bit OS X

kinke:
As before 2.079, which allowed for direct interop of D `size_t` with C++ `size_t` on all platforms except for 32-bit OS X (which is about to reach end of life AFAIK).
The 2.079 mangling change to C++ `long long` allows for direct interop of D `long` with C++ `int64_t` on all platforms (which was already the case, with the notable 64-bit OS X exception). My argument is that (the now correctly defined) `core.stdc.stdint.int64_t` alias should be used for that purpose (names matching perfectly), and that `size_t` interop is far more important and breaking it in 2.079+ by no means worth the questionable gain.
E.g., there are about 90 occurrences of `size_t` in the DMD C++ headers included by LDC. Most of them affect the function name mangling and would thus need to be changed to the ugly `d_size_t` typedef to fix all potential linking errors. So almost every mixed D/C++ code base supporting 64-bit macOS will have to adapt the code on the C++ and/or D side due to this breaking 2.079 ABI change, so I'd rather revert it while we still can.
The new/fixed aliases in druntime (`cpp_size_t`, `int64_t`, `cpp_long`, `cpp_longlong` etc.), coupled with the magic enums (`__c_long` etc.) for implicit convertibility, now give the user the right tools for proper portable C++ interop wrt. integers in new code.

### DIFF
--- a/changelog/osx64_long_cppmangling.dd
+++ b/changelog/osx64_long_cppmangling.dd
@@ -1,0 +1,11 @@
+64-bit OS X: Revert C++ mangling of `long` to pre-2.079 to restore `size_t` interop
+
+Direct interop of D `size_t` and C++ `size_t` was working before 2.079, on all platforms
+except for 32-bit OS X. By mangling D `long` as C++ `long long` on 64-bit OS X starting
+with 2.079, `size_t` interop broke on a more relevant platform.
+With new/fixed aliases, e.g., $(REF int64_t, core, stdc, stdint),
+$(REF uint64_t, core, stdc, stdint) and $(REF cpp_size_t, core, stdc, config), there are
+now proper tools for portable C++ interop wrt. integers.
+Reverting to the previous C++ mangling on 64-bit OS X (C++ `long`) may save mixed D/C++
+code bases from the need of manual adaptations by skipping the 2.079 and 2.080 DMD
+versions.
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -1179,10 +1179,10 @@ extern(C++):
             case Tuns32:                c = 'j';        break;
             case Tfloat32:              c = 'f';        break;
             case Tint64:
-                c = Target.c_longsize == 8 ? Target.int64Mangle : 'x';
+                c = Target.c_longsize == 8 ? 'l' : 'x';
                 break;
             case Tuns64:
-                c = Target.c_longsize == 8 ? Target.uint64Mangle : 'y';
+                c = Target.c_longsize == 8 ? 'm' : 'y';
                 break;
             case Tint128:                c = 'n';       break;
             case Tuns128:                c = 'o';       break;
--- a/src/dmd/target.d
+++ b/src/dmd/target.d
@@ -59,8 +59,6 @@ struct Target
         // C++ ABI
         bool reverseCppOverloads; /// set if overloaded functions are grouped and in reverse order (such as in dmc and cl)
         bool cppExceptions;       /// set if catching C++ exceptions is supported
-        char int64Mangle;         /// mangling character for C++ int64_t
-        char uint64Mangle;        /// mangling character for C++ uint64_t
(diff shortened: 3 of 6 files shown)


kinke:
Requires https://github.com/dlang/druntime/pull/2191.

kinke:
Mangling D `long` as C++ `long` consistently on all 64-bit Posix platforms allows for these tiny `core.stdc.config` adaptations:
```diff
@@ -117,14 +117,11 @@ else version( Posix )
 {
   static if( (void*).sizeof > int.sizeof )
   {
-    enum __c_long  : long;
-    enum __c_ulong : ulong;
-
     enum __c_longlong  : long;
     enum __c_ulonglong : ulong;

-    alias __c_long   cpp_long;
-    alias __c_ulong  cpp_ulong;
+    alias long  cpp_long;
+    alias ulong cpp_ulong;

     alias long  c_long;
     alias ulong c_ulong;
@@ -198,17 +195,10 @@ else version( SDC )

 static assert(is(c_long_double), "c_long_double needs to be declared for this platform/architecture.");

-version (Darwin)
-{
+version (Darwin) { version (D_LP64) {} else version = Darwin_32; }
+version (Darwin_32)
     alias cpp_size_t = cpp_ulong;
-    version (D_LP64)
-        alias cpp_ptrdiff_t = cpp_long;
-    else
-        alias cpp_ptrdiff_t = ptrdiff_t;
-}
 else
-{
     alias cpp_size_t = size_t;
-    alias cpp_ptrdiff_t = ptrdiff_t;
-}
+alias cpp_ptrdiff_t = ptrdiff_t;
 }
```

jacob-carlborg (approved):

WalterBright:
> As before 2.079, which allowed for direct interop of D size_t with C++ size_t on all platforms except for 32-bit OS X (which is about to reach end of life AFAIK).

The purpose of this change seems to be all about `size_t`. As the next step, my plan was to fix the declaration of `size_t` in object.d to match the one used by the corresponding C++ compiler, so there shouldn't be further issues with it. That would seem to negate the point of this PR. This PR doesn't solve the size_t problem either, as while 32 bit OSX may be obsolete, there's nothing preventing another C++ platform having the same issue.

The bottom line is there is no "correct" solution to this problem as long as C++ has multiple types that map to the same size.

kinke:
I fully agree (and my main point being that the ABI change required not just a few **manual** adaptations for no real gain). It'd be nice (and the problem solved) if you can come up with something better than `alias size_t = cpp_size_t`; having such a core type defined as magic `__c_long` enum on OSX64 doesn't feel right to me.

WalterBright:
It doesn't feel right to me, either, but there's nothing we can really do about it. Having `size_t` not always be a core type on current D platforms forces us to find and fix any issues that may produce. If we can't fix those issues, then this PR is a reasonable alternative.

I'm pretty pleased with the enum hack to deal with this, wish I'd thought of it years ago. It's sooo much better than the struct hack!

wilzbach (approved):

wilzbach:
Rebased this to master. As the druntime PR has been merged, I don't see anything blocking this?

MartinNowak:
What nobody did though is testing/fixing dmd's C++ <-> D interop based on this. I'm now getting linkage errors when building dmd with dmd-2.082.0. Guess that's the price of having dropped the OSX self-hosted tests on Travis-CI [¹](https://github.com/dlang/dmd/pull/7721), but I wouldn't expect it to be of much use despite the integer mangling struggles.

jacob-carlborg:
>  I'm now getting linkage errors when building dmd with dmd-2.082.0

Yeah, I run into the same problem today.

kinke:
> but I wouldn't expect it to be of much use despite the integer mangling struggles

The LDC self-hosting CI tests (on Windows/Linux/Mac) have proven quite useful, exactly to prevent C++ interop regressions.
LDC even has a special case for DMD host compiler versions in [2.079, 2.081] due to the temporary mangling change...

PR #8286, opened 2018-05-24
Fix issue 18902 - -lib crashes on static libraries in the command line

belka-ew:


JinShil:
What about just passing the .a file to the linker without generating an error?

JinShil (approved):

test/runnable/test18902.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+$DMD -m${MODEL} -lib 18902.a || test $? = 1
```
wilzbach: This doesn't check for the exit code.
If DMD exists with 0, it still will pass.
I recommend the following:
- use grep to check for the error message (check the other bash tests for examples)
- use `! $DMD` to ensure that only a failing DMD exit code will pass the test

belka-ew: I don't see a need to check the error message since the bug can be identified with the return code.
What about:
```d
if $DMD -m${MODEL} -lib 18902.a; then
	exit 1
else
	test $? = 1
fi
```
?
But if `grep` is a preferd way for this kind of tests, I can rewrite it with grep, it doesn't matter that much.

RazvanN7 (approved):

test/runnable/test18902.sh
MartinNowak: I agree with Sebastian, you want to check for an error message and exit code 1, so that both cannot regress.

belka-ew: But the modified code does `exit 1` and exits with a non-zero exit code if the result becomes zero.

test/runnable/test18902.sh
```diff
@@ -0,0 +1,8 @@
+#!/usr/bin/env bash
+
+# dmd -lib should fail without input sources/object files
+if $DMD -m${MODEL} -lib 18902.a; then
+    exit 1
+else
+    [ $? -eq 1 ]
+fi
```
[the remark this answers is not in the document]
MartinNowak: Negating doesn't seem to allow you to check the exit code though, and there is quite a difference between 139 and 1.

PR #8288, opened 2018-05-25
Fix C++ mangling for substitution of pointers with mismatching qualifier.

skl131313:
https://run.dlang.io/gist/9821400cfd8ae64eba7f964994096aa3
https://godbolt.org/g/81fNBg

### DIFF
--- a/changelog/mangle_substitute.dd
+++ b/changelog/mangle_substitute.dd
@@ -1,0 +1,12 @@
+`extern (C++)` now mangles substitution with const pointers correctly for Itanium mangling (Windows mangling was unaffected).
+
+`extern (C++)` mangling of function definitions which contained similar pointer parameters but differed because of a const wouldn't be substituted correctly.
+This has been correctly and now substitutes as expected, generating the correct C++ mangling.
+
+---
+extern(C++, NameSpace) void SubstitueFailure(const(char)*, const(char**));
+
+
+pragma(msg, SubstitueFailure.mangleof) // Before: _ZN9NameSpace16SubstitueFailureEPKcPKPS0_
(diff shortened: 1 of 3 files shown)


jacob-carlborg:
Is the mangling on Windows already correct?

skl131313:
Windows uses Microsoft's specific mangling, it doesn't have this issue as far as I am aware. Its a completely different format anyways.

WalterBright:
Need to create a bug report in bugzilla explaining what the problem is this PR addresses.

skl131313:
I'd rather not deal with that rat's nest, you can close this pull request if you won't accept it without one.

WalterBright:
> rat's nest

I'm curious what the barrier there is for you posting on bugzilla?

WalterBright:
A search on bugzilla of 'skl131313' turns up nothing:
https://issues.dlang.org/buglist.cgi?bug_status=NEW&bug_status=ASSIGNED&bug_status=REOPENED&email1=skl131313&emailreporter1=1&emailtype1=substring&list_id=221733&query_format=advanced

jacob-carlborg (changes requested):
A bugzilla issue or changelog entry is required, otherwise this won’t show up in the changelog.

skl131313:
@WalterBright I don't use the same username for an email and I definitely wouldn't give my email to be publicly viewed on something like bugzilla.

Geod24:
@skl131313 : Thanks for taking the time to find and fix this!
Unfortunately the comparison code has changed significantly since you've made your PR, so I've submitted a reworked version (#8919 ) using your test cases as a base.

PR #8289, opened 2018-05-25
dinterpret.d: more use of stack temporaries

WalterBright:


src/dmd/dinterpret.d
```diff
@@ -1089,11 +1089,11 @@ public:
         if (istate.start == s)
             istate.start = null;
 
-        size_t dim = s.statements ? s.statements.dim : 0;
-        for (size_t i = 0; i < dim; i++)
+        const dim = s.statements ? s.statements.dim : 0;
+        foreach (i; 0 .. dim)
```
jacob-carlborg: Please use a regular foreach loop instead of a range foreach if possible.

src/dmd/dinterpret.d
```diff
@@ -1112,11 +1112,11 @@ public:
...
```
jacob-carlborg: Same here.

jacob-carlborg (changes requested):

src/dmd/dinterpret.d
WalterBright: I didn't because look at the declaration for `dim`, there's a null check that wouldn't be there in the foreach.

jacob-carlborg: Make sense.

jacob-carlborg (approved):

PR #8290, opened 2018-05-25
[WIP] Mangle all the C++ operators good and proper

TurkeyMan:
We can map the operators just fine, so let's.

### DIFF
--- a/src/dmd/cppmanglewin.d
+++ b/src/dmd/cppmanglewin.d
@@ -684,22 +684,138 @@ private:
             buf.writestring("?1");
             return;
         }
+        else if (sym.ident)
+        {
+            switch (sym.ident.toString)
+            {
+                case "opAssign":    buf.writestring("?4");  return;
+                case "opEquals":    buf.writestring("?8");  return;
+                case "opIndex":     buf.writestring("?A");  return;
+                case "opCall":      buf.writestring("?R");  return;
+                default:            break;
+            }
+        }
         if (TemplateInstance ti = sym.isTemplateInstance())
         {
+            auto id = ti.tempdecl.ident;
+            const(char)* ident = id.toChars;
+
+            // test for special symbols
+            int firstTemplateArg = 0;
+            int whichOp = -1;
+            switch (ti.name.toString)
+            {
+                case "opCast":
+                    buf.writestring("?B");
+                    return;
+                case "opUnary":     whichOp = 0;    break;
+                case "opBinary":    whichOp = 1;    break;
+                case "opOpAssign":  whichOp = 2;    break;
+                case "opAssign":    whichOp = 3;    break;
+                case "opEquals":    whichOp = 4;    break;
+                case "opIndex":     whichOp = 5;    break;
+                case "opCall":      whichOp = 6;    break;
+                default:            break;
+            }
+            if (whichOp != -1)
+            {
+                TemplateDeclaration td = ti.tempdecl.isTemplateDeclaration();
+                assert(td);
+                TemplateParameter tp = (*td.parameters)[0];
+                TemplateValueParameter tv = tp.isTemplateValueParameter();
+                if (tv)
+                {
+                    if (tv.valType.isString())
+                    {
+                        Expression exp = (*ti.tiargs)[0].isExpression();
+                        StringExp str = exp.toStringExp();
+                        switch (whichOp)
+                        {
+                            case 0: // opUnary
+                                switch (str.peekSlice)
+                                {
+                                    case "*":   ident = "?D";   goto continue_template;
+                                    case "++":  ident = "?E";   goto continue_template;
+                                    case "--":  ident = "?F";   goto continue_template;
+                                    case "-":   ident = "?G";   goto continue_template;
+                                    case "+":   ident = "?H";   goto continue_template;
+                                    case "~":   ident = "?S";   goto continue_template;
+                                    default:    break;
+                                }
+                                break;
+                            case 1: // opBinary
+                                switch (str.peekSlice)
+                                {
+                                    case ">>":  ident = "?5";   goto continue_template;
+                                    case "<<":  ident = "?6";   goto continue_template;
+                                    case "*":   ident = "?D";   goto continue_template;
+                                    case "-":   ident = "?G";   goto continue_template;
+                                    case "+":   ident = "?H";   goto continue_template;
+                                    case "&":   ident = "?I";   goto continue_template;
+                                    case "/":   ident = "?K";   goto continue_template;
+                                    case "%":   ident = "?L";   goto continue_template;
+                                    case "^":   ident = "?T";   goto continue_template;
+                                    case "|":   ident = "?U";   goto continue_template;
+                                    default:    break;
+                                }
+                                break;
+                            case 2: // opOpAssign
+                                switch (str.peekSlice)
+                                {
+                                    case "/=":  ident = "?_0";  goto continue_template;
+                                    case "%=":  ident = "?_1";  goto continue_template;
+                                    case ">>=": ident = "?_2";  goto continue_template;
+                                    case "<<=": ident = "?_3";  goto continue_template;
+                                    case "&=":  ident = "?_4";  goto continue_template;
+                                    case "|=":  ident = "?_5";  goto continue_template;
+                                    case "^=":  ident = "?_6";  goto continue_template;
+                                    case "*=":  ident = "?X";   goto continue_template;
+                                    case "+=":  ident = "?Y";   goto continue_template;
+                                    case "-=":  ident = "?Z";   goto continue_template;
+                                    default:    break;
+                                }
+                                break;
+                            case 3: // opAssign
+                                ident = "?4";
+                                break;
+                            case 4: // opEquals
+                                ident = "?8";
+                                break;
+                            case 5: // opIndex
+                                ident = "?A";
+                                break;
+                            case 6: // opCall
+                                ident = "?R";
+                                break;
+                            default:
+                                break;
+                            continue_template:
+                                if (ti.tiargs.dim == 1)
+                                {
+                                    buf.writestring(ident);
+                                    return;
+                                }
+                                firstTemplateArg = 1;
+                                break;
+                        }
+                    }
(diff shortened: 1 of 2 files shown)


TurkeyMan:
Still gotta do posix...

src/dmd/cppmanglewin.d
```diff
@@ -684,22 +684,138 @@ private:
...
+                    return;
+                case "opUnary":     whichOp = 0;    break;
+                case "opBinary":    whichOp = 1;    break;
+                case "opOpAssign":  whichOp = 2;    break;
+                case "opAssign":    whichOp = 3;    break;
+                case "opEquals":    whichOp = 4;    break;
+                case "opCall":      whichOp = 5;    break;
+                case "opIndex":     whichOp = 6;    break;
+                default:            break;
+            }
```
jacob-carlborg: Why splitting this in two switches? If we have to, I prefer an enum over an int.

TurkeyMan: Because the block of work immediately below shouldn't be entered at all if the symbol name isn't a special name, so the names need to be considered up front.  
From there, I'd rather not repeat the string comparisons inside the block below, so I cache the comparison as an int.

I felt like defining an enum for a trivial thing was overkill, but I can do it if people feel strongly about it.

The enum definition will be about as many extra lines as the switch that will use it... I don't feel like those extra lines bring value to carry their weight.

jacob-carlborg: No, I don't feel strongly about it.

TurkeyMan:
Incidentally, I was surprised how well the D operators and C++ operators can map to eachother. I thought this wasn't supported because of poor compatibility of expression, but turns out the only one that's awkward are the comparisons.

src/dmd/cppmanglewin.d
```diff
@@ -684,22 +684,138 @@ private:
...
             scope VisualCPPMangler tmp = new VisualCPPMangler((flags & IS_DMC) ? true : false);
             tmp.buf.writeByte('?');
             tmp.buf.writeByte('$');
-            auto id = ti.tempdecl.ident;
-            tmp.buf.writestring(id.toChars());
-            tmp.saved_idents[0] = id.toChars();
-            tmp.buf.writeByte('@');
+            tmp.buf.writestring(ident);
+            tmp.saved_idents[0] = ident;
+            if (ident == id.toChars)
```
WalterBright: Please don't drop the () on the `toChars`

src/dmd/cppmanglewin.d
```diff
@@ -684,22 +684,138 @@ private:
...
+        }
         if (TemplateInstance ti = sym.isTemplateInstance())
         {
+            auto id = ti.tempdecl.ident;
+            const(char)* ident = id.toChars;
+
+            // test for special symbols
+            int firstTemplateArg = 0;
+            int whichOp = -1;
+            switch (ti.name.toString)
```
WalterBright: Resorting to string compares is terribly inefficient. DMD puts all identifiers into a hash table, and then comparing identifiers becomes a simple comparison of pointers.

src/dmd/cppmanglewin.d
```diff
@@ -684,22 +684,138 @@ private:
...
+                TemplateDeclaration td = ti.tempdecl.isTemplateDeclaration();
+                assert(td);
+                TemplateParameter tp = (*td.parameters)[0];
+                TemplateValueParameter tv = tp.isTemplateValueParameter();
+                if (tv)
+                {
+                    if (tv.valType.isString())
+                    {
+                        Expression exp = (*ti.tiargs)[0].isExpression();
+                        StringExp str = exp.toStringExp();
```
WalterBright: Please use the operator, not the string representation of the operator.

src/dmd/cppmanglewin.d
```diff
@@ -684,22 +684,138 @@ private:
...
+                        switch (whichOp)
+                        {
+                            case 0: // opUnary
+                                switch (str.peekSlice)
+                                {
+                                    case "*":   ident = "?D";   goto continue_template;
```
WalterBright: Please do not use `ident` as a pointer to a 0 terminated string. Use `name` instead. Use of `ident` should be restricted to an instance of `Identifier`.

TurkeyMan:
@WalterBright Thanks Walter, these are all great points!

TurkeyMan:
@WalterBright: While I have you, there was a small part of me that wanted to close the loop by addressing the comparison operators... Unless D is considering having an optional alternative to opCmp which would be allowing users to define the complete set explicitly, I toyed with the idea that `extern(C++)` classes with opCmp defined would emit 4 magic functions that each wrap the 4 comparisons on the result of opCmp, and mangle them for C++ to call.

Then I decided to not... since D can't link the other way, and C++ can fairly easily just extern to opCmp and wrap them in local inlines.

TurkeyMan:
Problem is that, currently, we have no good way to link to a C++ class that implements comparisons without invading the C++ code, which I'm interested in minimising/eliminating.

TurkeyMan:
Perhaps we could introduce `bool opCmpExplicit(string op)(...)`, which the compiler would attempt to instantiate like `opCmpExplicit!("<=")` prior to calling existing `opCmp`...

For cases other than extern(C++), there are cases where opCmp results on a performance disadvantage compared to explicit comparisons, so niche cases even just within D might want that too (I certainly have!).

src/dmd/cppmanglewin.d
TurkeyMan: Why?

src/dmd/cppmanglewin.d
TurkeyMan: I'm not sure what this means?

WalterBright: `exp.op`

src/dmd/cppmanglewin.d
WalterBright: Because the `( )` are there on every other use in dmd. It's jarring to see it removed, and makes it look like a typo.
I prefer the paren-less versions only when using the . operator to chain function calls.

TurkeyMan: Okay. I have had other PR push-backs because they wanted me to NOT match the surrounding code, and move slowly towards modern D style. It's hard to know the rules here ;)

src/dmd/cppmanglewin.d
TurkeyMan: I don't understand... `exp.op == 121`, that is `TOK.string_`... That is, the expression is a StringExp, which I get on the next line.  
Is `toStringExp()` expensive?

Or are you saying there's some way for me to get the operator that the string represents without testing the instantiation string? How can I know what the string value of the first template arg is without evaluating it? What if someone instantiates with ctfe?

src/dmd/cppmanglewin.d
```diff
@@ -684,22 +684,148 @@ private:
...
         {
+            auto id = ti.tempdecl.ident;
+            const(char)* symName = id.toChars();
+
+            int firstTemplateArg = 0;
+
+            // test for special symbols
+            import std.algorithm : countUntil;
+            auto operators = [Id._cast, Id.assign, Id.eq, Id.index, Id.call, Id.opUnary, Id.opBinary, Id.opOpAssign];
+            sizediff_t whichOp = operators.countUntil!((o, n) => n.equals(o))(ti.name);
```
TurkeyMan: @WalterBright does this make you upset? :)
Without this, I return to `if`/`else if`, and the thing gets a lot longer and uglier.

WalterBright: We don't use phobos in the dmd source, mainly to try to keep dependencies down when dmd is ported to a new machine.

src/dmd/cppmanglewin.d
```diff
@@ -684,22 +684,148 @@ private:
             buf.writestring("?1");
             return;
         }
+        else if (sym.ident)
+        {
+            if (sym.ident.equals(Id.assign))
```
WalterBright: `sym.ident == Id.assign` should be just fine.

src/dmd/cppmanglewin.d
```diff
@@ -684,22 +684,148 @@ private:
...
+                    TemplateParameter tp = (*td.parameters)[0];
+                    TemplateValueParameter tv = tp.isTemplateValueParameter();
+                    if (!tv || !tv.valType.isString())
+                        break; // expecting a string argument to operators!
+                    Expression exp = (*ti.tiargs)[0].isExpression();
+                    StringExp str = exp.toStringExp();
+                    switch (whichOp)
+                    {
+                        case 5: // opUnary
+                            switch (str.peekSlice)
```
WalterBright: `switch (exp.op)`

src/dmd/cppmanglewin.d
```diff
@@ -684,22 +684,148 @@ private:
...
+                            {
+                                case "*":   symName = "?D";     goto continue_template;
```
WalterBright: `case TOK.mul:`

src/dmd/cppmanglewin.d
TurkeyMan: I still don't get it... `exp.op == TOK.string_`. The template arg to the operators is a string of the operator.

src/dmd/cppmanglewin.d
TurkeyMan: `exp.op == TOK.string_`, it doesn't distinguish operators. The value of the string is the operator on question.

src/dmd/cppmanglewin.d
WalterBright: Sorry, I was wrong. You are correct.

WalterBright (approved):

src/dmd/cppmanglewin.d
TurkeyMan: Fortunately, this deep code will never be executed by not-extern(c++)-not-operators... it's not a perf hazard ;)

TurkeyMan:
Not done... Posix >_<

TurkeyMan:
The other one's done though ;)

WalterBright:
Please don't add Posix to this PR. Make another one! This one is complex enough.

PR #8292, opened 2018-05-25
dinterpret.d: more use of stack temporaries pt 4

WalterBright:


jacob-carlborg (approved):

ibuclaw:
This caused a regression.

https://issues.dlang.org/show_bug.cgi?id=19074

PR #8293, opened 2018-05-25
Initial pure D build script

wilzbach:
(inspired by @JinShil's [recent dream on the NG](https://forum.dlang.org/post/dwnirthzzwtzmxypvizp@forum.dlang.org) about being able to build DMD easily on any platform)

Similarly to the `test/run.d`, this allows building DMD from source without depending on the Makefile.
At the moment it's a simplistic translation of more or the less the bare minimum of the `posix.mak` which is needed to build DMD, but the following is intended (not in this PR though):
- support all OS (getting rid of the win32/win64.mak files is a major motivation)
- allow this script to be used by dub (this should allow to build/expose full DMD via DUB)
- target compatibility with the posix.mak Makefile (i.e. all targets that are supported in posix.mak should be supported here)

I opened this PR to avoid duplicated efforts and give others already a sneak-peek at this.
This already compiles DMD on Linux and even supports partial and parallel compilation (almost as good the Make build).

I kept the rule system as simplistic as possible to avoid adding a huge chunk of code that we would have to maintain + it's not really needed here anyhow.
Also the builder script got quite long, because instead of using one text blob for the source files, I used single lines for each file to avoid the merge conflicts that sometimes happen (though I have no strong feeling on this style).

Why such a naive approach and not cmake, ninja, reggae, ...?
-----------------------------------------------------------------------------------

There has never been a consensus on which third-party build system should be used.
Using a pure D script avoids any further dependencies and after all, D should be powerful to handle any task!
Also the dependency flow is rather simplistic here. The backend and lexer need to be built before the main DMD built, but they can be built in parallel once the initial dependencies have been generated.

But the Makefile is working so nicely
------------------------------------------------

Well, I guess I'm one of the very few persons who understands the `posix.mak` (it looks more scary than it is), but I doubt many others do.
Also we there have been numerous complaints about the more or less unmaintained win32.mak and win64.mak files.
For now, like `run.d` this is an experiment and only intended to slowly replace the Makefiles once it has been thoroughly tested.

But isn't depending on D another dependency?
--------------------------------------------------------------

Nope, D requires a host D compiler since 2015. The Makefile comes with an AUTO_BOOTSTRAP option to automatically download a D compiler. While I added support for this in `build.d` too (it makes sense when you want to lock the D compiler version to a specific one), it's only of limited help to someone on a fresh machine without a D compiler.
Imho we can expect a D compiler dev to have a working D installation, but if this AUTO_BOOTSTRAP option is convenient for some (I think the Jenkins CI uses it), we could opt to add a small Bash bootstrapping script around it (or keep this part of the Makefile).

What's missing from this PR?
---------------------------------------

This PR contains everything essential to use the `build.d` script on x86_64 Linux.
Though there's still much work that needs to be done in future PRs.
Check the TODO header.

What won't be part of this PR?
----------------------------------------

I don't plan to support Windows nor all targets of posix.mak in this initial PR.

### DIFF
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -2,7 +2,7 @@
 
 set -uexo pipefail
 
-HOST_DMD_VER=2.074.1 # same as in dmd/src/posix.mak
+HOST_DMD_VER=2.079.1 # same as in dmd/src/posix.mak
 CURL_USER_AGENT="CirleCI $(curl --version | head -n 1)"
 N=4
 CIRCLE_NODE_INDEX=${CIRCLE_NODE_INDEX:-0}
@@ -171,6 +171,25 @@ check_run_individual()
 	"${build_path}/dmd"  -i -run ./test/run.d test/runnable/template2962.d ./test/compilable/test14275.d
 }
 
+# Checks the D build.d script
+check_d_builder()
+{
+    echo "Testing D build"
+    # load environment for bootstrap compiler
+    if [ -f ~/dlang/install.sh ] ; then
+        source "$(CURL_USER_AGENT=\"$CURL_USER_AGENT\" bash ~/dlang/install.sh dmd-$HOST_DMD_VER --activate)"
+    fi
+    ./src/build.d clean
+    rm -rf generated # just to be sure
+    # TODO: add support for 32-bit builds
+    ./src/build.d MODEL=64
+    ./generated/linux/release/64/dmd --version | grep -v "dirty"
+    ./src/build.d clean
+    if [ -f ~/dlang/install.sh ] ; then
+        deactivate
+    fi
+}
+
 codecov()
 {
     # CodeCov gets confused by lst files which it can't matched
@@ -183,13 +202,8 @@ codecov()
 case $1 in
     install-deps) install_deps ;;
     setup-repos) setup_repos ;;
-    coverage) echo "removed" ;;
-    check-clean-git) echo "removed" ;;
-    codecov)
-        echo "removed - use 'all'"
-        # Fall-through is used to maintain compatibility with the existing PRs
-        ;&
     all)
+        check_d_builder;
         coverage;
         check_clean_git;
         check_run_individual;
--- a/src/build.d
+++ b/src/build.d
@@ -1,0 +1,892 @@
+#!/usr/bin/env rdmd
+/**
+DMD builder
+
+Usage:
+  ./build.d dmd
+
+TODO:
+- add different ENABLE modes
+- add all posix.mak Makefile targets
+- support 32-bit builds
+- test on OSX
+- test on Windows
+- allow appending DFLAGS via the environment
+- test the script with LDC or GDC as host compiler
+*/
+
+version(CoreDdoc) {} else:
+
+import std.algorithm, std.conv, std.datetime, std.exception, std.file, std.format,
+       std.getopt, std.parallelism, std.path, std.process, std.range, std.stdio, std.string;
+import core.stdc.stdlib : exit;
+
+const thisBuildScript = __FILE_FULL_PATH__;
+const srcDir = thisBuildScript.dirName.buildNormalizedPath;
+shared bool verbose; // output verbose logging
+shared bool force; // always build everything (ignores timestamp checking)
+
+__gshared string[string] env;
+__gshared string[][string] flags;
+__gshared typeof(sourceFiles()) sources;
+
+void main(string[] args)
+{
+    int jobs = totalCPUs;
+    auto res = getopt(args,
+        "j|jobs", "Specifies the number of jobs (commands) to run simultaneously (default: %d)".format(totalCPUs), &jobs,
+        "v", "Verbose command output", (cast(bool*) &verbose),
+        "f", "Force run (ignore timestamps and always run all tests)", (cast(bool*) &force),
+    );
+    if (res.helpWanted)
+    {
+        defaultGetoptPrinter(`./build.d <targets>...
+
+Examples:
+
+    ./build.d dmd                                                  # build DMD
+    ./build.d unittest                                             # runs internal unittests
+    ./build.d clean                                                # remove all generated files
+
+Options:
+`, res.options);
+        "\nSee the README.md for a more in-depth explanation of the build script.".writeln;
+        return;
+    }
+
+    // parse arguments
+    args.popFront;
+    args2Environment(args);
+
+    // bootstrap all needed environment variables
+    env = getEnvironment;
+
+    // get all sources
+    sources = sourceFiles;
+
+    // default target
+    if (!args.length)
+        args = ["all"];
+
+    auto targets = args
+        .predefinedTargets // preprocess
+        .array;
+
+    if (targets.length > 0)
+    {
+        if (verbose)
+        {
+            log("================================================================================");
+            foreach (key, value; env)
+                log("%s=%s", key, value);
+            log("================================================================================");
+        }
+    }
+}
+
+/**
+D build dependencies
+====================
+
+The strategy of this script is to emulate what the Makefile is doing,
+but without a complicated dependency and dependency system.
+The "dependency system" used here is rather naive and only parallelizes the
+build of the backend and lexer (writing a few config files doesn't take much time).
+However, it does skip steps when the source files are younger than the target
+and thus supports partial rebuilds.
+
+Below all individual dependencies of DMD are defined.
+They have a target path, sources paths and an optional name.
+When a dependency is needed either its command or custom commandFunction is executed.
+A dependency will be skipped if all targets are older than all sources.
+This script is by default part of the sources and thus any change to the build script,
+will trigger a full rebuild.
+
+The function buildDMD defines the build order of its dependencies.
+*/
+
+// TODO: newdelete is probably not needed anymore
+auto newDelete()
+{
+    Dependency dependency = {
+        target: env["G"].buildPath("newdelete").objName,
+        sources: [env["ROOT"].buildPath("newdelete.c")],
+        name: "(CC) NEW_DELETE",
+        command: [env["HOST_CXX"], "-c", "-o$@", "$<"]
+    };
+    return dependency;
+}
+
+// Builds the lexer as a separate library
+auto lexer()
+{
+    Dependency dependency = {
+        target: env["G"].buildPath("lexer").libName,
+        sources: sources.lexer,
+        rebuildSources: configFiles,
+        name: "(CC) D_LEXER_OBJ",
+        command: [
+            env["HOST_DMD_RUN"],
+            "-of$@",
+            "-lib",
+            "-J"~env["G"], "-J../res",
+            "-L-lstdc++",
+        ].chain(flags["DFLAGS"], "$<".only).array
+    };
+    return dependency;
+}
+
+// Generates a dmd.conf file in the generated folder
+auto dmdConf()
+{
+    // TODO: add support for Windows
+    string exportDynamic;
+    version(OSX) {} else
+        exportDynamic = " -L--export-dynamic";
+
+    auto conf = `[Environment32]
+DFLAGS=-I%@P%/../../../../../druntime/import -I%@P%/../../../../../phobos -L-L%@P%/../../../../../phobos/generated/$(OS)/$(BUILD)/32{exportDynamic}
+
+[Environment64]
+DFLAGS=-I%@P%/../../../../../druntime/import -I%@P%/../../../../../phobos -L-L%@P%/../../../../../phobos/generated/$(OS)/$(BUILD)/64{exportDynamic} -fPIC`.replace("{exportDynamic}", exportDynamic);
+
+    auto target = env["G"].buildPath("dmd.conf");
+    auto commandFunction = (){
+        conf.toFile(target);
+    }; // defined separately to support older D compilers
+    Dependency dependency = {
+        target: target,
+        name: "(TX) DMD_CONF",
+        commandFunction: commandFunction,
+    };
+    return dependency;
+}
+
+/*
+optabgen generates a few C++ files.
+Thus it first needs to be built and the executed.
+*/
+auto opTabGen()
+{
+    auto opTabFiles = ["debtab.c", "optab.c", "cdxxx.c", "elxxx.c", "fltables.c", "tytab.c"];
+    auto opTabFilesBin = opTabFiles.map!(e => env["G"].buildPath(e)).array;
+    auto opTabBin = env["G"].buildPath("optabgen").exeName;
+    auto opTabSourceFile = env["C"].buildPath("optabgen.c");
+
+    auto commandFunction = (){
+        auto args = [env["HOST_CXX"], "-I"~env["TK"], opTabSourceFile, "-o", opTabBin];
+        args ~= flags["CXXFLAGS"];
+
+        writefln("(CC) BUILD_OPTABGEN");
+        args.runCanThrow;
+
+        writefln("(CC) RUN_OPTABBIN %-(%s, %)", opTabFiles);
+        [opTabBin].runCanThrow;
+
+        // move the generated files to the generated folder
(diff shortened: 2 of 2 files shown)


src/build.d
```diff
@@ -0,0 +1,1105 @@
...
+        targetCH = "code_stub.h";
+        targetObjs = ["platform_stub"];
+    }
+    else
+    {
+        assert(0, "Unknown TARGET_CPU: " ~ env["TARGET_CPU"]);
+    }
+    Sources sources = {
+        frontend: [
+	        "arraytypes",
```
JinShil: Do the source files need to be explicitly enumerated?  I envisioned something more like `sourceDir.dirEntries("*.d", SpanMode.depth).join(" ")`

wilzbach: I tried to stay as close as possible to the Makefile, but in general there's no big reason not search for them though  there are a few dead files which are in the codebase, but never used.

src/build.d
```diff
@@ -0,0 +1,1105 @@
...
+        "v", "Verbose command output", (cast(bool*) &verbose),
+        "f", "Force run (ignore timestamps and always run all tests)", (cast(bool*) &force),
+    );
+    if (res.helpWanted)
+    {
+        defaultGetoptPrinter(`./build.d <targets>...
+
+Examples:
+
+    ./run.d dmd                                                  # build DMD
```
JinShil: I think this should be "build.d" instead of "run.d"

.circleci/run.sh
```diff
@@ -183,13 +202,8 @@ codecov()
 case $1 in
     install-deps) install_deps ;;
     setup-repos) setup_repos ;;
-    coverage) echo "removed" ;;
-    check-clean-git) echo "removed" ;;
-    codecov)
-        echo "removed - use 'all'"
-        # Fall-through is used to maintain compatibility with the existing PRs
-        ;&
```
wilzbach: As I was already touching this code, I removed these long-dead targets.

.circleci/run.sh
```diff
@@ -171,6 +171,25 @@ check_run_individual()
...
+    echo "Testing D build"
+    # load environment for bootstrap compiler
+    if [ -f ~/dlang/install.sh ] ; then
+        source "$(CURL_USER_AGENT=\"$CURL_USER_AGENT\" bash ~/dlang/install.sh dmd-$HOST_DMD_VER --activate)"
+    fi
+    ./src/build.d clean
+    rm -rf generated # just to be sure
+    # TODO: add support for 32-bit builds
+    ./src/build.d MODEL=64
+    ./generated/linux/release/64/dmd --version | grep -v "dirty"
```
wilzbach: I will fix the 32-bit build in a follow-up PR.

src/build.d
```diff
@@ -0,0 +1,1143 @@
...
+TODO:
+- build.d should be a default dependency to all rules
+- improve rebuild dependencies (e.g. header files should trigger a rebuild)
+- add different ENABLE modes
+- add all posix.mak Makefile targets
+- support 32-bit builds
+- test on OSX
+- test on Windows
+- allow appending DFLAGS via the environment
+- evaluate whether to use globbing for the file listings
```
wilzbach: I know it's a long list, but imho it's better to have something as an initial base on which incremental reviews can be done.
Also with an existing build.d other people can more easily join this effort.

src/build.d
```diff
@@ -0,0 +1,1131 @@
...
+    auto conf = `[Environment32]
+DFLAGS=-I%@P%/../../../../../druntime/import -I%@P%/../../../../../phobos -L-L%@P%/../../../../../phobos/generated/$(OS)/$(BUILD)/32{exportDynamic}
+
+[Environment64]
+DFLAGS=-I%@P%/../../../../../druntime/import -I%@P%/../../../../../phobos -L-L%@P%/../../../../../phobos/generated/$(OS)/$(BUILD)/64{exportDynamic} -fPIC`.replace("{exportDynamic}", exportDynamic);
+
+    auto target = env["G"].buildPath("dmd.conf");
+    auto commandFun = (){
+        conf.toFile(target);
+    }; // defined separately to support older D compilers
```
wilzbach: Due to a recently DMD bug declaring the function directly doesn't work.
See also:

https://issues.dlang.org/show_bug.cgi?id=18199
https://github.com/dlang/dmd/pull/8051

wilzbach (commented):
Okay, so while there's still a ton of things that should be done with this script, it's already quite usable on x86_64 Linux (to verify this I added this script to CircleCi).
Also perfect is the enemy of progress and this PR is already quite big.
So I would suggest we use this as a base for future improvements and PR which should then be a lot easier to review and manage.
Also with an existing build.d other people can more easily join this effort.

Everyone ok with starting of with the minimal subset and slowly increasing it?

src/build.d
```diff
@@ -0,0 +1,1131 @@
...
+                "TODO: html".writeln; // TODO
+                break;
+
+            case "install":
+                "TODO: install".writeln; // TODO
+                break;
+
+            case "man":
+                "TODO: man".writeln; // TODO
+                break;
```
wilzbach: All these targets are nice-to-have (as we eventually want to get rid of the Makefile), but obviously not required for building dmd. I added them here, s.t. adding them doesn't get forgotten.

src/build.d
```diff
@@ -0,0 +1,1131 @@
...
+
+            case "clean":
+                env["G"].rmdirRecurse;
+                exit(0);
+                break;
+
+            default:
+            case "all":
+                goto dmd;
```
wilzbach: This is to maintain "compatibility" with the Makefile whose default goal was "all" and is just an alias for "dmd".

src/build.d
```diff
@@ -0,0 +1,1131 @@
...
+
+    env.getDefault("GIT_HOME", "https://github.com/dlang");
+    env.getDefault("SYSCONFDIR", "/etc");
+    env.getDefault("TMP", "/tmp");
+    env.getDefault("PGO_DIR", srcDir.buildPath("pgo"));
+    auto d = env.getDefault("D", "dmd");
+    env.getDefault("C", d.buildPath("backend"));
+    env.getDefault("TK", d.buildPath("tk"));
+    env.getDefault("ROOT", d.buildPath("root"));
+    env.getDefault("EX", d.buildPath("examples"));
```
wilzbach: This variable isn't used yet, but it will very likely soon be needed (i.e. when the example goal gets added)

src/build.d
```diff
@@ -0,0 +1,1131 @@
...
+    }
+    else
+    {
+        env["PIC_FLAG"] = "";
+    }
+
+    env.getDefault("GIT_HOME", "https://github.com/dlang");
+    env.getDefault("SYSCONFDIR", "/etc");
+    env.getDefault("TMP", "/tmp");
+    env.getDefault("PGO_DIR", srcDir.buildPath("pgo"));
```
wilzbach: This variable isn't used yet, but once `ENABLE_PGO` gets added, it should be used.

src/build.d
```diff
@@ -0,0 +1,1131 @@
...
+            auto curlFlags = "-fsSL --retry 5 --retry-max-time 120 --connect-timeout 5 --speed-time 30 --speed-limit 1024";
+            hostDMDRoot.mkdirRecurse;
+            ("curl " ~ curlFlags ~ " " ~ hostDMDURL~".tar.xz | tar -C "~hostDMDRoot~" -Jxf - || rm -rf "~hostDMDRoot).spawnShell.wait;
+        }
+    }
+    else
+    {
+        env["HOST_DMD_PATH"] = ["which", env["HOST_DMD"]].execute.output.strip;
+        env["HOST_DMD_RUN"] = env["HOST_DMD"];
+    }
```
wilzbach: Auto-bootstrapping doesn't make so much sense anymore as it did with the Makefile (after all a D compiler is already required to run this script).
However, the use case of looking the compilation to a specific D version still exists.

src/build.d
```diff
@@ -0,0 +1,1131 @@
...
+// D source files
+////////////////////////////////////////////////////////////////////////////////
+
+auto sourceFiles()
+{
+    struct Sources
+    {
+        // TODO: remove unused file lists
+        string[] frontend, lexer, lexerRoot, root, glue, dmd, backend;
+        string[] backendHeaders, tkHeaders, backendC, tkC, backendObjects;
```
wilzbach: We could use some globbing here to find all the files which would probably cut down half of the size of this script.
The problem is that the frontend is separated in multiple parts and we would have to do a bit of filtering and thus maybe not save so much.

jacob-carlborg:
> Everyone ok with starting of with the minimal subset and slowly increasing it?

Please do.

jacob-carlborg:
Do not work on macOS:

```
object.Exception@./build.d(1135): dmd/backend/cgelem.c:1253:35: error: '&&' within '||' [-Werror,-Wlogical-op-parentheses]
                    op == OPu8_16 && (e->Eoper == OPand || !(i & ~0xFF)) ||
                    ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ ~~
```

Looks like it's missing some flags to compile the C++ code.

src/build.d
```diff
@@ -0,0 +1,1137 @@
...
+            warnings ~= [
+                "-Wno-logical-op",
+                "-Wno-narrowing",
+                "-Wno-unused-but-set-variable",
+                "-Wno-uninitialized",
+                "-Wno-class-memaccess",
+                "-Wno-implicit-fallthrough",
+            ];
+        else if (env["CXX_KIND"] == "clang++")
+            warnings ~= "-Wno-logical-op-parentheses";
```
jacob-carlborg: The check for `clang++` should be moved to the else branch of `env["ENABLE_WARNINGS"] != "0"`.

JinShil:
> Everyone ok with starting of with the minimal subset and slowly increasing it?

Yes, though I would like to see a brief description of the strategy.

wilzbach:
> Do not work on macOS:

Yep, sorry but I don't have a macOS machine for testing.
Though I do plan to add this script to `auto-tester-test` target in the main `posix.mak`, s.t. we/I can test it for OSX.

> Yes, though I would like to see a brief description of the strategy.

Here's how I plan to proceed over the next weeks:

1) Fix 32-bit build mode
2) Add different ENABLE modes
3) Test with LDC as host
4) OSX + Windows support
5) Remaining TODO tasks (e.g. allow appending DFLAGS from the environment)
6) Add other Makefile targets

(2), (4) and (6) might result in multiple PRs.
Is this what you were asking for or did you ask for the strategy of this build script?

----

The strategy of this script is to emulate what the Makefile is doing, but without a complicated dependency and dependency system.
The "dependency system" used here is rather naive and only parallelizes the build of the backend and lexer (writing a few config files doesn't take much time).
However, it does skip steps when the source files are younger than the target and thus supports partial rebuilds.

All individual dependencies of DMD are defined. They have a target path, sources paths and an optional name. When a dependency is needed either its command or custom commandFunction is executed.
A dependency will be skipped if all targets are older than all sources. The build script is by default part of the sources and thus any change to the build script, will trigger a full rebuild.

The function buildDMD defines the build order of its dependencies.

JinShil:
> Is this what you were asking for or did you ask for the strategy of this build script?

The steps 1~6 you described is what I was looking for.  Thanks.

jacob-carlborg:
> Yep, sorry but I don't have a macOS machine for testing.

I haven't tested everything but by fixing the warnings it will now at least build DMD successfully.

src/build.d
```diff
@@ -0,0 +1,1178 @@
+#!/usr/bin/env rdmd
+/**
+DMD builder
+
+Usage:
+  ./build.d dmd
+
+TODO:
+- add different ENABLE modes
```
JinShil: What are "ENABLE modes"?

JinShil:
Its appears that we will be bringing up build.d in parallel with the existing makefiles.  We are using Circle CI as the CI for build.d.  At what point will we replace posix.mak with build.d?  After steps 1~3 are completed, can we make the switch?

JinShil:
Finally, as much as I really want this to happen, I think we should wait to pull this until @WalterBright  is finished porting the backend to D.  That process seems to involve a lot of makefile changes, and putting the burden on @WalterBright and reviewers to modify and verify two different builds scripts is going to be a bit of a nuisance.

wilzbach:
> At what point will we replace posix.mak with build.d? After steps 1~3 are completed, can we make the switch?

In theory we could already make the switch for e.g. CircleCi (only the enable modes are missing and they are quick to add), though I wanted to wait a few months to make sure there aren't any sneaky bugs in the build script somewhere and as you mentioned yourself it makes sense to wait with adding this build script until the backend conversion is complete.

> As much as I really want this to happen, I think we should wait to pull this until @WalterBright is finished porting the backend to 

Fair enough, but how about adding the script without adding it to the CIs (for now)?
That would still come with the advantage of being able to make incremental progress on it and expose it for other people who want to play with/test it.

Though of course, we could replace the hard-coded file list with file globbing and thus eliminate the need for manually updating this build script during the backend conversion.
Maybe I try this as a follow-up.

JinShil:
> Fair enough, but how about adding the script without adding it to the CIs (for now)?

But we need to test it as we go.

> Though of course, we could replace the hard-coded file list with file globbing and thus eliminate the need for manually updating this build script during the backend conversion.

That would be great.  If you can pull it off, I'd support adding it to the CI now, but we should be at @WalterBright's beckon call should he run into any issues.

wilzbach:
>>  we could replace the hard-coded file list with file globbing and thus eliminate the need for manually updating this build script during the backend conversion.
> That would be great. If you can pull it off

I think I managed to do so.
It now builds with only a minimally hard-coded lists and globs for all other source files.

src/build.d
```diff
@@ -0,0 +1,889 @@
...
+        return "freebsd";
+    else version(OpenBSD)
+        return "openbsd";
+    else version(NetBSD)
+        return "netbsd";
+    else version(DragonFlyBSD)
+        return "dragonflybsd";
+    else version(Solaris)
+        return "solaris";
+    else version(SunOS)
```
jacob-carlborg: Is this a predefined version identifier?

src/build.d
```diff
@@ -0,0 +1,889 @@
...
+}
+
+/*
+Detects the host model
+
+Returns: 32, 64 or throws an Exception
+*/
+auto detectModel()
+{
+    version(Solaris)
```
jacob-carlborg: Should `SunOS` be used here as well? Or should `detectOS` be used instead?

wilzbach:
BTW when the recent `tocvdebug.d` PR was merged no modifications where required, which is a promising sign that the newly adding globbing does its job.
I addressed @jacob-carlborg's points. Anything else you want to see in this initial version?
(as [mentioned](https://github.com/dlang/dmd/pull/8293#issuecomment-394089370) there's a plan to extend this step by step over the next weeks).

JinShil (approved):
It's a big step in the right direction.  Let's move forward.  I don't want this to hang around in an unfinished state for years, though.  I hope we can keep at it regularly until it's done.

jacob-carlborg (approved):

src/build.d
```diff
@@ -0,0 +1,892 @@
...
+shared bool verbose; // output verbose logging
+shared bool force; // always build everything (ignores timestamp checking)
+
+__gshared string[string] env;
+__gshared string[][string] flags;
+__gshared typeof(sourceFiles()) sources;
+
+void main(string[] args)
+{
+    int jobs = totalCPUs;
```
andralex: This variable is never used. A primary feature of a build system is to facilitate parallel builds. Are there plans to implement such?

wilzbach: It always was (though of course the user should have been able to regulate this with `-j`): 

https://github.com/wilzbach/dmd/blob/e417f0e0dfd3ea117609cd48a9de53695aaaa3ad/src/build.d#L244

This is still the case for master:

https://github.com/dlang/dmd/blob/9e877e08d9933449380721d85e29b0f71a567cae/src/build.d#L141-L142

https://github.com/dlang/dmd/blob/9e877e08d9933449380721d85e29b0f71a567cae/src/build.d#L1317-L1320

PR #8294, opened 2018-05-25
Fix typo in dmd.expression apparently causing sporadic frontend crashes

kinke:
Seems to fix sporadic segfaults reported in https://issues.dlang.org/show_bug.cgi?id=18864.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -2120,7 +2120,7 @@ extern (C++) abstract class Expression : RootObject
 
         if (v.type.ty == Tstruct)
         {
-            StructDeclaration sd = (cast(TypeStruct)type).sym;
+            StructDeclaration sd = (cast(TypeStruct)v.type).sym;
             if (sd.hasNoFields)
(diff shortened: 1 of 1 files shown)


dnadlinger (approved):

kinke:
That was quick, thanks; I should have targeted stable though. :/ Pinging @MartinNowak.

wilzbach:
Ah we typically only do cherry-picking if it's important, so if you feel like this really should be part of 2.080.1, simply cherry-pick and open a new PR.
OTOH the next release will be in one month.

kinke:
@wilzbach: So a fix for random DMD segfaults isn't important enough to be cherry-picked by one of you guys? What is then? ;)

wilzbach:
Ehm I'm currently traveling and without laptop, otherwise I would have opened the cherry-pick for you.
Well, Martin always complains about this because these cherry-picks can result in ugly merge conflicts, so yep we often don't bother except for really important things :/

PR #8299, opened 2018-05-26
typedefs to make transition of backend from C to D easier

WalterBright:


wilzbach (approved):

jacob-carlborg (approved):

WalterBright:
Thanks. This means I can make the transition to D with fewer diffs.

PR #8304, opened 2018-05-29
fix Issue 18905 - [Reg 2.079] C++ classes can no longer be used with …

WalterBright:
…-betterC

### DIFF
--- a/src/dmd/toobj.d
+++ b/src/dmd/toobj.d
@@ -358,10 +358,14 @@ void toObjFile(Dsymbol ds, bool multiobj)
                 return;
             }
 
+            const bool gentypeinfo = global.params.useTypeInfo && Type.dtypeinfo;
+            const bool genclassinfo = gentypeinfo || !(cd.isCPPclass || cd.isCOMclass);
+
             // Generate C symbols
-            toSymbol(cd);
-            toVtblSymbol(cd);
-            Symbol *sinit = toInitializer(cd);
+            if (genclassinfo)
+                toSymbol(cd);                           // __ClassZ symbol
+            toVtblSymbol(cd);                           // __vtblZ symbol
+            Symbol *sinit = toInitializer(cd);          // __initZ symbol
 
             //////////////////////////////////////////////
 
@@ -376,16 +380,18 @@ void toObjFile(Dsymbol ds, bool multiobj)
                 outdata(sinit);
             }
 
+            if (genclassinfo)
+            {
             //////////////////////////////////////////////
 
             // Put out the TypeInfo
-            if (global.params.useTypeInfo && Type.dtypeinfo)
+            if (gentypeinfo)
                 genTypeInfo(cd.loc, cd.type, null);
             //toObjFile(cd.type.vtinfo, multiobj);
 
             //////////////////////////////////////////////
 
-            // Put out the ClassInfo
+            // Put out the ClassInfo, which will be the __ClassZ symbol in the object file
             cd.csym.Sclass = scclass;
             cd.csym.Sfl = FLdata;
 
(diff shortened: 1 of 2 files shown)


WalterBright:
The code between the new `{ ... }` is not indented in order to keep the diffs simple. This section of code should be pulled out into a separate function, which I will do as a separate refactoring PR.

test/compilable/test18905.d
```diff
@@ -0,0 +1,3 @@
+// https://issues.dlang.org/show_bug.cgi?id=18905
```
JinShil: Don't you need -betterC compiler flags here?  Also, you might want to add this to the existing betterc.d test file.

JinShil (dismissed):

src/dmd/toobj.d
```diff
@@ -358,10 +358,14 @@ void toObjFile(Dsymbol ds, bool multiobj)
                 return;
             }
 
+            const bool gentypeinfo = global.params.useTypeInfo && Type.dtypeinfo;
```
JinShil: If you merge https://github.com/dlang/druntime/pull/2184, then you'll only need to check `Type.dtypeinfo` because the runtime will no longer import the `TypeInfo` classes when compiling with -betterC.  There are other places in the DMD source code that can be simplified due to that PR.

JinShil:
> The code between the new { ... } is not indented in order to keep the diffs simple. 

I don't think you need to worry about that anymore.  GitHub now has a "Hide whitespace changes" option under the "Diff Settings" on the review tab to aid with that:

image

...but it's no big deal either way.

src/dmd/toobj.d
WalterBright: I'm not so sure about that. 2184 seems to be more concerned with the nothrow issue, which is unrelated to this PR's fix. It isn't ready yet, either, and this PR fix works and is pretty simple.

test/compilable/test18905.d
WalterBright: Yes, good catch!

WalterBright:
I know about the no whitespace view, but I wanted to yank all that code out into a separate function anyway, and that view won't help with that. If I yank it out now, then nobody will want to review this because it's multiple screens of diffs.

src/dmd/toobj.d
JinShil: 2184 is actually more concerned with preventing the import of code which is not relevant to -betterC build.  The `nothrow` addition was added because -betterC code should be `nothrow` at all times.  If you want me to separate the two into different PRs, I can.  Just let me know.  And, it is all green and ready to go.

WalterBright: I'm not too happy about the nothrow business, as I mentioned. But we should discuss that elsewhere, not here. I also do not believe that 2184 will resolve this issue, as note it took two different tests to get it through the test suite, not just checking typeinfo.

JinShil (approved):

WalterBright:
@CyberShadow the following error appears in the CyberShadow/DAutoTest:
```
posix.mak:760: recipe for target 'dpl-docs' failed
```
but scanning the voluminous log, I can't see what the actual error is. Please advise!

CyberShadow:
It looks like a problem with Dub. It seems to have exited with a non-zero exit code without printing an error message.

Try rebasing and force-pushing to force a rebuild.

CyberShadow:
@s-ludwig @wilzbach Any idea?

CyberShadow:
It still looks like an intermittent failure. See http://dtest.dlang.io/history/master - a build of the master branch recently failed with a similar problem, but got fixed by itself.

src/dmd/toobj.d
```diff
@@ -503,12 +512,13 @@ void toObjFile(Dsymbol ds, bool multiobj)
...
-                            goto L2;
+                        {
+                            flags &= ~ClassFlags.noPointers;  // not no-how, not no-way
+                            break Louter;
+                        }
                     }
                 }
             }
-            flags |= ClassFlags.noPointers;
-          L2:
```
MartinNowak: Please avoid seemingly unrelated refactorings in bugfixes or mention why they were necessary.

MartinNowak (commented):
Should have targeted stable.

PR #8305, opened 2018-05-29
toobj.d: make genClassInfoForInterface() a separate function

WalterBright:
Straightforward cut&paste job.

### DIFF
--- a/src/dmd/toobj.d
+++ b/src/dmd/toobj.d
@@ -664,8 +664,6 @@ void toObjFile(Dsymbol ds, bool multiobj)
             else if (global.params.symdebug)
                 toDebug(id);
 
-            enum_SC scclass = SCcomdat;
-
             // Put out the members
             for (size_t i = 0; i < id.members.dim; i++)
             {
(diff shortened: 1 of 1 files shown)


JinShil:
Generally, I don't really like code that is factored into a function unless it has more than one caller.  Large functions are generally a symptom of some deeper code craftsmanship problem.  The `ClassDeclaration` and `InterfaceDeclaration` code look awfully similar.  It would probably be a more productive refactor to combine the two, introduce a `visit(AggregateDeclaration)`, or maybe even create an `AggregateDeclarationVisitor` class.

WalterBright:
> Generally, I don't really like code that is factored into a function unless it has more than one caller. Large functions are generally a symptom of some deeper code craftsmanship problem.

I decided to break it out after the nth time of scrolling down and getting lost.

> The ClassDeclaration and InterfaceDeclaration code look awfully similar.

They are.

> It would probably be a more productive refactor to combine the two, introduce a visit(AggregateDeclaration), or maybe even create an AggregateDeclarationVisitor class.

They started out the same in the distant past, but the branching and ifs turned out to make it hard to reason about.

jacob-carlborg (approved):

PR #8307, opened 2018-05-29
toobj.d: make genClassInfoForClass() a separate function

WalterBright:
Cut & paste job analogous to https://github.com/dlang/dmd/pull/8305 but for classes.


JinShil (approved):
At least it's now consistent with `InterfaceDeclaration`

PR #8308, opened 2018-05-30
Add more test coverage (runtime lowerings) for betterC

JinShil:
This adds test coverage to betterC to ensure https://github.com/dlang/druntime/pull/2194 doesn't break code.  It will probably need to be expanded, but this is what I have for now.

### DIFF
--- a/test/runnable/betterc.d
+++ b/test/runnable/betterc.d
@@ -25,6 +25,7 @@ extern (C) void main()
 {
     test(1);
     test18472();
+    testRuntimeLowerings();
 }
 
 /*******************************************/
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
We can’t do this in druntime? Didn’t you already create a betterC test in druntime?

JinShil:
This is not actually testing the druntime implementation.  What this test does is ensure DMD lowers the expression to the runtime implementation and calls it properly.  Well, actually, this does test the runtime implementation too, but that is by accident.

IMO this is the correct spot for this test as -betterC is a compiler flag, and we want to make sure -betterC code is making the right calls to the runtime implementations.

That being said, we do need runtime implementation tests as well, but those should be in the form of `unittest`s and they should call the runtime implementations directly (e.g. `__equals(x, y)`, as opposed to `x == y`.  I've asked @wilzbach for help, privately, to modify the druntime makefiles so we can run the druntime unittests under the -betterC compiler flag.  So, the runtime CIs would then run the unittests twice:  once without -betterC, and once with.  Then we can add `version(D_BetterC)` conditional compilations to the druntime unittests to verify the druntime PRs.

I hope that make sense.

PR #8315, opened 2018-05-31
__mutable

tgehr:
Implementation of `__mutable` from the DConf hackathon.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -1036,6 +1036,29 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             }
         }
 
+        if (dsym.storage_class & STC.mutable)
+        {
+            if (!dsym.isField())
+            {
+                dsym.error("only fields can be `__mutable`");
+            }
+            else
+            {
+                if (dsym.storage_class & STC.immutable_)
+                {
+                    dsym.error("cannot be both `__mutable` and `immutable`");
+                }
+                if (dsym.storage_class & STC.const_)
+                {
+                    dsym.error("cannot be both `__mutable` and `const`");
+                }
+                if (dsym.protection.kind != Prot.Kind.private_)
+                {
+                    dsym.error("`__mutable` fields must be `private`");
+                }
+            }
+        }
+
         // Calculate type size + safety checks
         if (sc.func && !sc.intypeof)
         {
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -1962,6 +1962,38 @@ extern (C++) abstract class Type : RootObject
     }
 
     /************************************
+     * Add MODxxxx bits to type of __mutable member variable
+     */
+
+    final Type mutableAddMod(MOD mod)
+    {
+        if (isShared())
+        {
+            return this;
+        }
+        if (mod & MODFlags.immutable_)
+        {
+            mod &= ~MODFlags.immutable_;
+            mod |= MODFlags.shared_;
+        }
+        if (mod & MODFlags.wild)
+        {
+            // TODO: even /reading/ the field is actually not okay.
+            mod &= ~MODFlags.wild;
+            mod |= MODFlags.const_;
+        }
+        if (mod & MODFlags.shared_)
+        {
+            mod &= ~MODFlags.const_;
+        }
+        if (mod & MODFlags.const_)
+        {
+            mod &= ~MODFlags.const_; // TODO: best behavior?
+        }
+        return addMod(mod);
+    }
+
+    /************************************
      * Add storage class modifiers to type.
      */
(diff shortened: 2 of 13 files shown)


leandro-lucarella-sociomantic:
Hi, I'm not trying to be a PITA but am I the only one that think adding `__mutable` to the language if a profoundly bad idea?

leandro-lucarella-sociomantic:
BTW, a change in the language that is this big certainly needs a DIP, right?

tgehr:
It does and I currently have only a rough draft, but Walter told me to create this pull request anyway.

tgehr:
For some reason my first comment is not here anymore, so again: this is not C++ `mutable`, `__mutable` is unsafe, because it does not disable rewrites based on `immuable`/`const`/`pure`. The reason this is being suggested at all is that Andrei wants the language to be able to store allocators within `immutable` containers. I don't think there is a better solution.

tgehr:
Also @leandro-lucarella-sociomantic : Maybe put forward some arguments in favor of your position. Do you think it is unreasonable to store references to allocators in `immutable` collections, or do you think there is a better solution?

src/dmd/mtype.d
```diff
@@ -1961,6 +1961,38 @@ extern (C++) abstract class Type : RootObject
         return t;
     }
 
+    /************************************
+     * Add MODxxxx bits to type of __mutable member variable
```
WalterBright:
```
Returns:
    type with added bits
```

WalterBright (approved):

leandro-lucarella-sociomantic:
I agree with @WalterBright that is better to leave the discussion about language change to the DIP and their review process, and keep only discussion about the implementation here. So I will follow that.

I think when submitting PRs for a new language feature that doesn't even have a DIP, it would be nice to make it clear to avoid hard attacks on some people :smiley: 

Like having a `[Proposal] Draft implementation of __mutable` in the title and something like `This PR is a reference implementation for an upcoming DIP, please don't merge yet` in the issue description.

test/compilable/__mutable.d
```diff
@@ -0,0 +1,92 @@
+// REQUIRED_ARGS: -o-
```
wilzbach: Why disable the object generation?
I think it wouldn't hurt to have at least one test that does test that `__mutable` interacts fine without the backend failing with it.

src/dmd/traits.d
```diff
@@ -699,6 +700,14 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
 
         return isDeclX(d => (d.storage_class & STC.lazy_) != 0);
     }
+    if (e.ident == Id.isUnderUnderMutable)
+    {
+        if (dim != 1)
+            return dimError(1);
```
wilzbach: Not covered.

src/dmd/dsymbolsem.d
```diff
@@ -1036,6 +1036,29 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             }
         }
 
+        if (dsym.storage_class & STC.mutable)
+        {
+            if (!dsym.isField())
+            {
+                dsym.error("only fields can be `__mutable`");
```
wilzbach: This results in hard to understand error message:

```
Error: variable `__mutable2.foo.p` only fields can be `__mutable`
```

maybe "cannot be __mutable. Only fields can."?

src/dmd/dsymbolsem.d
```diff
@@ -1036,6 +1036,29 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                {
+                    dsym.error("cannot be both `__mutable` and `immutable`");
+                }
+                if (dsym.storage_class & STC.const_)
+                {
+                    dsym.error("cannot be both `__mutable` and `const`");
+                }
+                if (dsym.protection.kind != Prot.Kind.private_)
+                {
+                    dsym.error("`__mutable` fields must be `private`");
```
wilzbach: Same thing here about the error message:

```
Error: variable `__mutable1.S.pm` `__mutable` fields must be `private`
```

maybe "can only be `private`"? Or triggering a full error with: "__mutable field X can only be private"?

src/dmd/mtype.d
ntrel: Redundancy in documentation is sometimes useful, e.g. if the `Returns` information was also mentioned in a *long* description, so the return can be quickly looked up. Here, reading the *short* description is clearly enough IMO.

src/dmd/dsymbolsem.d
bausshf: Or maybe "cannot be declared `__mutable`. `__mutable` can only be applied to fields."

RazvanN7:
Closing, as this endeavor is not going to be pursued.

PR #8317, opened 2018-06-01
convert list.c to dlist.d

WalterBright:
Git diff doesn't seem to render diffs when files get renamed. :-(


jacob-carlborg:
> Git diff doesn't seem to render diffs when files get renamed. :-(

You could do the rename in on commit and the actual changes in a second commit. The it's possible to diff the individual commits, but perhaps not directly on GitHub,

src/dmd/backend/dlist.d
```diff
@@ -0,0 +1,571 @@
...
+
+void list_term()
+{
+    if (list_inited)
+    {
+        debug printf("Max # of lists = %d\n",nlist);
+        while (list_freelist)
+        {
+            list_t list = list_next(list_freelist);
+            list_freelist = list;
```
JinShil: The C implementation had a call to `list_delete` here.

src/dmd/backend/dlist.d
```diff
@@ -0,0 +1,571 @@
...
+        }
+        debug if (nlist)
+            printf("nlist = %d\n",nlist);
+        assert(nlist == 0);
+        list_inited = 0;
+    }
+}
+
+
+list_t list_alloc();
```
JinShil: Why this declaration?

src/dmd/backend/dlist.d
```diff
@@ -0,0 +1,571 @@
...
+/* **************** PUBLIC FUNCTIONS ********************* */
+
+/********************************
+ * Create link to existing list, that is, share the list with
+ * somebody else.
+ *
+ * Returns:
+ *    pointer to that list entry.
+ */
+
```
JinShil: I'm not sure why you always have a newline between the DDoc and the code.  I recommend removing it, but it's not a problem.

src/dmd/backend/dlist.d
```diff
@@ -0,0 +1,571 @@
...
+ *      null if out of memory
+ */
+
+list_t list_prepend(list_t *plist, void *ptr)
+{
+    list_t list = list_alloc();
+    if (list)
+    {
+        list.next = *plist;
+        list.ptr = ptr;
```
JinShil: C implementation used the `list_next` and `list_ptr` wrappers.  It seems to be the same functionality, but throughout this file you're sometimes using the wrapper and sometimes not.  It makes it a little hard to review and the code is inconsistent.

wilzbach (commented):
Not much more I can add besides @JinShil's nice review.

src/dmd/backend/dlist.d
WalterBright: `list_next` and `list_ptr` don't work as lvalues in D.

src/dmd/backend/dlist.d
WalterBright: Yes, it does seem to be missing!

src/dmd/backend/dlist.d
```diff
@@ -0,0 +1,571 @@
...
+}
+
+list_t list_alloc(const(char)* file, int line)
+{
+    return list_alloc();
+}
+
+
+list_t list_new() { return cast(list_t)malloc(LIST.sizeof); }
+void list_delete(list_t list) { free(list); }
```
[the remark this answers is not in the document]
WalterBright: They're both used.

src/dmd/backend/dlist.d
WalterBright: Leftover from the conversion. Removed.

src/dmd/backend/dlist.d
WalterBright: The newline is the style commonly used. Others have started using the no newline approach. Besides, I try to minimize diffs on initial conversions.

src/vcbuild/dmd_backend.vcxproj.filters
```diff
@@ -174,15 +174,9 @@
...
     </ClCompile>
-    <ClCompile Include="..\dmd\tk\list.c">
-      <Filter>dmd\tk</Filter>
-    </ClCompile>
     <ClCompile Include="..\dmd\tk\mem.c">
       <Filter>dmd\tk</Filter>
     </ClCompile>
-    <ClCompile Include="..\dmd\tk\vec.c">
-      <Filter>dmd\tk</Filter>
-    </ClCompile>
```
wilzbach: It might make sense to move the `vec.h` header file to the backend, s.t. D code and header files are "in sync".

src/dmd/backend/dlist.d
```diff
@@ -0,0 +1,570 @@
...
+        if (list)
+        {
+            list.next = null;
+            list.ptr = p;
+            list.count = 1;
+            *pe = list;
+            pe = &list.next;
+        }
+    }
+    //va_end(ap);
```
wilzbach: What's the reason for not calling `va_end` here? 
(in C it's called)

wilzbach (approved):
Let's move finally forward with this!
I'm just unsure about the commented `va_end` call.

src/dmd/backend/dlist.d
```diff
@@ -0,0 +1,570 @@
...
+
+alias LIST* list_t;             /* pointer to a list entry              */
+
+/* FPNULL is a null function pointer designed to be an argument to
+ * list_free().
+ */
+
+alias void function(void*) @nogc nothrow list_free_fp;
+
+enum FPNULL = cast(list_free_fp)null;
```
rainers: You can avoid the cast with `enum list_free_fp FPNULL = null`. But defining a new symbol here doesn't really buy a lot in comparison to just using `null`.

src/dmd/backend/dlist.d
```diff
@@ -0,0 +1,570 @@
...
+
+list_t list_alloc()
+{
+    list_t list;
+
+    if (list_freelist)
+    {
+        list = list_freelist;
+        list_freelist = list_next(list);
+        //mem_setnewfileline(list,file,line);
```
rainers: Keeping this as a comment doesn't really help as all the (conditional) parameter declarations have gone.

src/dmd/backend/dlist.d
```diff
@@ -0,0 +1,570 @@
...
+
+nothrow @nogc
+{
+
+/* **************** TYPEDEFS AND DEFINES ****************** */
+
+struct LIST
+{
+        /* Do not access items in this struct directly, use the         */
+        /* functions designed for that purpose.                         */
```
rainers: Can you make them private now?

WalterBright:
Fellas, these are good suggestions but please keep in mind the #1 rule of translating code - do not attempt to refactor, fix, enhance, improve, rename, reorganize, etc. the code. Do as little as possible to get it to work. I have learned this bitter lesson over and over and over.

src/dmd/backend/dlist.d
WalterBright: There was some problem with it being @nogc and nothrow. core.stdc.stdarg apparently needs some additional work.

src/dmd/backend/dlist.d
WalterBright: FPNULL dates from the olden days when function pointers were different from regular pointers due to the wacky DOS memory models. It doesn't make sense any more, as you point out, but changing it here violates the #1 rule of translation.

JinShil (approved):
> do not attempt to refactor, fix, enhance, improve, rename, reorganize, etc. the code

Yeah, let's get this done.

src/dmd/backend/dlist.d
WalterBright: Gawd will smite me if I violate the #1 rule of translation.

WalterBright: Note that if I make it private now, then I have to add to this PR all the other files that are modified to make that work. Of course, that will inevitably cause more problems. Such changes need to be separate PRs.

src/dmd/backend/dlist.d
wilzbach: Compiles fine for me -> https://github.com/dlang/dmd/pull/8367

PR #8319, opened 2018-06-01
Fix Issue 9701 - allow UDAs to be attached to enum values

JinShil:
This is a reboot of https://github.com/dlang/dmd/pull/6161

I've also added the ability to mark enum members as `deprecated` and `@disabled`.

It was decided a while ago that a proper implementation would be accepted without a DIP.  Here's the forum announcement:  https://forum.dlang.org/post/gcgtotrhuirhcvvbxrfm@forum.dlang.org

This only implements the feature for enum members.  For function arguments, see https://github.com/dlang/dmd/pull/7576

### DIFF
--- a/test/compilable/test9701.d
+++ b/test/compilable/test9701.d
@@ -1,0 +1,56 @@
+// https://issues.dlang.org/show_bug.cgi?id=9701
+
+template AliasSeq(TList...)
+{
+    alias AliasSeq = TList;
+}
+
+enum
+{
+    uda4,
+    uda5,
+    uda6,
+    uda8,
+    uda9
+}
+
+enum Enum
+{
+    value0,
+    @("uda1") value1,
+    @("uda2", "uda3", 42) value2,
+    @uda4 value3,
+    @uda5 @uda6 value4,
+    @("uda7") @uda8 value5,
+    @uda9 @("uda10") value6,
+    deprecated value7
+}
+
+@("uda0")
+enum
+{
+    value0,
+    @("uda1") value1,
+    @("uda2") @("uda3") value2,
+    @uda4 value3,
+    @uda5 @uda6 value4,
+    @("uda7") @uda8 value5,
+    @uda9 @("uda10") value6
+}
+
+static assert(__traits(getAttributes, Enum.value0).length == 0);
+static assert(__traits(getAttributes, Enum.value1) == AliasSeq!("uda1"));
+static assert(__traits(getAttributes, Enum.value2) == AliasSeq!("uda2", "uda3", 42));
+static assert(__traits(getAttributes, Enum.value3) == AliasSeq!(uda4));
+static assert(__traits(getAttributes, Enum.value4) == AliasSeq!(uda5, uda6));
+static assert(__traits(getAttributes, Enum.value5) == AliasSeq!("uda7", uda8));
+static assert(__traits(getAttributes, Enum.value6) == AliasSeq!(uda9, "uda10"));
+static assert(__traits(isDeprecated, Enum.value7));
+
+static assert(__traits(getAttributes, value0) == AliasSeq!("uda0"));
+static assert(__traits(getAttributes, value1) == AliasSeq!("uda0", "uda1"));
+static assert(__traits(getAttributes, value2) == AliasSeq!("uda0", "uda2", "uda3"));
+static assert(__traits(getAttributes, value3) == AliasSeq!("uda0", uda4));
+static assert(__traits(getAttributes, value4) == AliasSeq!("uda0", uda5, uda6));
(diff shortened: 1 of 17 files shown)


test/compilable/test9701.d
```diff
@@ -0,0 +1,33 @@
...
+    @disable disabled,
+    deprecated deprecated_
+}
+
+@("uda0")
+enum
+{
+    value0,
+    @("uda1") value1,
+    @("uda2") @("uda3") value2,
```
jacob-carlborg: Should be a test with the `@uda` syntax as well.

JinShil:
This is currently held up on a problem with the `VarExp` semantic visitor.  It seems to be needlessly called multiple times.  So if I put a deprecation message there, it gets printed multiple times.  I'm trying to resolve that in a separate PR, but I'm having a hell of a time getting it right.

src/dmd/expression.d
```diff
@@ -4299,7 +4299,6 @@ extern (C++) final class VarExp : SymbolExp
         super(loc, TOK.variable, __traits(classInstanceSize, VarExp), var, hasOverloads);
         //printf("VarExp(this = %p, '%s', loc = %s)\n", this, var.toChars(), loc.toChars());
         //if (strcmp(var.ident.toChars(), "func") == 0) assert(0);
-        this.type = var.type;
```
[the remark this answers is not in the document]
jacob-carlborg: Is this really the correct fix?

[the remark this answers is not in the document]
JinShil: I'm not 100% sure, but if you search expressionsem.d for `if (e.type)` you will see it is a common idiom.  The expression classes don't seem to have a flag indicating whether or not semantic has been run, so it looks like the `type` property is (ab)used for that.

That being said, I don't see my `if (type)` check in the diff, so I may have neglected to add it.  It doesn't work at the moment anyway because there are other places in the code that don't call `expressionSemantic` on the `VarExp` class, so I need to submit a separate PR to remedy that.  I worked on it all day today, and it's proving to be quite difficult.

If I don't this, then a deprecation message is printed every type `VarExp.expressionSemantic` is run, and that's suboptimal.  I need some way to know whether or it has already been run, and the dominating idiom for that seems to be `if (e.type)` in the semantic visitor functions.

test/compilable/test9701.d
JinShil: Done

test/compilable/test9701.d
```diff
@@ -0,0 +1,56 @@
...
+enum Enum
+{
+    value0,
+    @("uda1") value1,
+    @("uda2", "uda3", 42) value2,
+    @uda4 value3,
+    @uda5 @uda6 value4,
+    @("uda7") @uda8 value5,
+    @uda9 @("uda10") value6,
+    deprecated value7
```
jacob-carlborg: Please add a test for `deprecated` with a message as well, i.e. `deprecated("foo")`.

JinShil:
I'm closing this.  My frustration with D has hit a new high, and I need to walk away for a while.

IgorStepanov:
@JinShil Hi. Why do you closed this PR? Are there any troubles in implementation?

IgorStepanov:
You may call `e.checkDeprecated(sc, e.var);` only if `e.type` is null. (before the first semantic of VarExp) 
Or you may return from `visit` if e.type is not null (if you are sure that there should be only one semantic for VarExp).

JinShil:
> Are there any troubles in implementation?

Hi @IgorStepanov, Yes, there is one fundamental issue that I can't find a solution for.  The semantic pass for `VarExp` is called multiple times.  Therefore, when I add the code to display an error message for the `@disable` attribute, the error message prints at least 3 times for each `VarExp`.  I tried to solve it by checking if the `type` property is null to determine if semantic had already been run (that pattern exists elsewhere for other derived `Expression` types), but while it solved the multiple error message problem, it caused a cascade of other failures.  I tried to work through it little-by-little solving each failure, but the problem was so systematic, it became hopeless.

[I asked for help on the forum](https://forum.dlang.org/post/pqlimjsieoapmdtzblou@forum.dlang.org) but never received a response.

IgorStepanov:
@JinShil we may add the field with flag only for attribute errors. Ensure that this field properly copies in `copy` `syntaxCopy` functions. This is an ugly solution, but it doesnt block a very useful feature. 
And why do you check deprecation in VarExp semantic, not in EnumMember semantic? 
I'll try to fix it in a week, I hope.

JinShil:
> And why do you check deprecation in VarExp semantic, not in EnumMember semantic?

Because we want the error message at the usage site, not the declaration site.

PR #4862, opened 2015-08-05
issue 14564 - [REG2.067] dmd -property -unittest combination causes compiler error

aG0aep6G:
Was: #4850. Differences: targeting master now; removed `test/fail_compilation/diag8629.d` entirely; fixed `-property -w`.
## 

fixes issue 14564 - [REG2.067] dmd -property -unittest combination causes
compiler error

https://issues.dlang.org/show_bug.cgi?id=14564

The problem is that phobos isn't compatible with `-property`. And it looks like `-property`'s current behaviour isn't desired anymore.

The plan (also in `mars.c`):
- Step 1 (2.069): Ignore -property and warn about it.
- Step 2: Deprecate -property.
- Step 3: Remove -property. Throw an error when it's set.
- Step 4: Possibly reintroduce -property with different semantics.

To do if this gets agreed upon:
- update http://dlang.org/dmd-windows.html and friends,
- changelog entry.

An alternative way to fix issue 14564 would be to make phobos compatible with the current `-property`.

PR #8320, opened 2018-06-01
Fix Issue 8161 - deprecate invalid property function definitions

ntrel:
Add error for void getter or aggregate setter method with 2 params.

### DIFF
--- a/src/dmd/dscope.d
+++ b/src/dmd/dscope.d
@@ -597,7 +597,7 @@ struct Scope
     }
 
     /********************************************
-     * Search enclosing scopes for ClassDeclaration.
+     * Search enclosing scopes for Struct/ClassDeclaration.
      */
     extern (C++) AggregateDeclaration getStructClassScope()
     {
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -1150,10 +1150,16 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
         {
             const np = Parameter.dim(tf.parameters);
             if (tf.varargs || np > 2 || (np == 2 && sc.getStructClassScope))
+            {
                 mtype.error(loc, "setter property can only have one %s",
-                    sc.getStructClassScope ? "parameter".ptr : "or two parameters".ptr), errors = true;
-            else if (np == 0 && tf.next && tf.next.ty == Tvoid)
-                mtype.error(loc, "getter properties must not return void"), errors = true;
+                    sc.getStructClassScope ? "parameter".ptr : "or two parameters".ptr);
+                errors = true;
+            }
+            else if (np == 0 && tf.next && tf.next.ty == Tvoid && !tf.isref)
+            {
+                mtype.error(loc, "getter properties must not return void");
+                errors = true;
+            }
         }
 
         if (tf.varargs == 1 && tf.linkage != LINK.d && Parameter.dim(tf.parameters) == 0)
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -1148,17 +1148,20 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
 
         if (tf.isproperty)
         {
+            // Deprecated in 2018-06.
+            // Change both to error in 2019-06 and uncomment the `errors =` lines.
+            // @@@DEPRECATED_2019-06@@@.
             const np = Parameter.dim(tf.parameters);
             if (tf.varargs || np > 2 || (np == 2 && sc.getStructClassScope))
             {
-                mtype.error(loc, "setter property can only have one %s",
+                deprecation(loc, "setter property can only have one %s",
                     sc.getStructClassScope ? "parameter".ptr : "or two parameters".ptr);
-                errors = true;
+                //errors = true;
             }
             else if (np == 0 && tf.next && tf.next.ty == Tvoid && !tf.isref)
             {
-                mtype.error(loc, "getter properties must not return void");
-                errors = true;
+                deprecation(loc, "getter properties must not return void");
+                //errors = true;
             }
         }
 
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -1146,10 +1146,14 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
             errors = true;
         }
 
-        if (tf.isproperty && (tf.varargs || Parameter.dim(tf.parameters) > 2))
+        if (tf.isproperty)
         {
-            mtype.error(loc, "properties can only have zero, one, or two parameter");
-            errors = true;
+            const np = Parameter.dim(tf.parameters);
+            if (tf.varargs || np > 2 || (np == 2 && sc.getStructClassScope))
+                mtype.error(loc, "setter property can only have one %s",
+                    sc.getStructClassScope ? "parameter".ptr : "or two parameters".ptr), errors = true;
+            else if (np == 0 && tf.next && tf.next.ty == Tvoid)
+                mtype.error(loc, "getter properties must not return void"), errors = true;
         }
 
         if (tf.varargs == 1 && tf.linkage != LINK.d && Parameter.dim(tf.parameters) == 0)
--- a/test/fail_compilation/fail334.d
+++ b/test/fail_compilation/fail334.d
@@ -1,11 +1,28 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/fail334.d(10): Error: properties can only have zero, one, or two parameter
+fail_compilation/fail334.d(15): Error: setter property can only have one parameter
+fail_compilation/fail334.d(16): Error: setter property can only have one parameter
+fail_compilation/fail334.d(17): Error: setter property can only have one parameter
+fail_compilation/fail334.d(19): Error: getter properties must not return void
+fail_compilation/fail334.d(26): Error: setter property can only have one or two parameters
+fail_compilation/fail334.d(27): Error: getter properties must not return void
 ---
 */
 
 struct S
 {
     @property int foo(int a, int b, int c) { return 1; }
+    @property void set(int, int) {}
+    @property void set(...) {}
+    @property void set(int) {} // OK
+    @property void get() {}
 }
+
+// OK
+@property ufcs_set(Object obj, int val) {}
+@property void set(int) {}
+
+@property void set(int[]...) {}
+@property void get() {}
+
--- a/test/fail_compilation/fail334.d
+++ b/test/fail_compilation/fail334.d
@@ -1,12 +1,13 @@
+// REQUIRED_ARGS: -de
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/fail334.d(15): Error: setter property can only have one parameter
-fail_compilation/fail334.d(16): Error: setter property can only have one parameter
-fail_compilation/fail334.d(17): Error: setter property can only have one parameter
-fail_compilation/fail334.d(19): Error: getter properties must not return void
-fail_compilation/fail334.d(27): Error: setter property can only have one or two parameters
-fail_compilation/fail334.d(28): Error: getter properties must not return void
+fail_compilation/fail334.d(16): Deprecation: setter property can only have one parameter
+fail_compilation/fail334.d(17): Deprecation: setter property can only have one parameter
+fail_compilation/fail334.d(18): Deprecation: setter property can only have one parameter
+fail_compilation/fail334.d(20): Deprecation: getter properties must not return void
+fail_compilation/fail334.d(28): Deprecation: setter property can only have one or two parameters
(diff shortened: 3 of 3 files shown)


jacob-carlborg:
I'm not sure if this is a good idea because I'm not sure what we want to do with `@property`.

ntrel:
@jacob-carlborg I'm not aware of any plans for `@property`, could you elaborate? I think getters not returning void is intuitive. Setters accepting more than one value should either be illegal, or otherwise, we remove the limit on accepting a sequence *iff* it only has two elements (see [here](https://issues.dlang.org/show_bug.cgi?id=8161#c4)).

jacob-carlborg:
Currently `@property` doesn't give any value. The question is if we should keep it or not.

wilzbach (approved):
No matter how our plans for @property are, I think it's more than reasonable to forbid invalid cases like @property getters which return void and show a better error message, so I like this idea.

BTW a potential DIP is here: https://github.com/dlang/DIPs/pull/97

Though I think we should update the [spec](https://dlang.org/spec/function.html#property-functions) to match the compiler error.

wilzbach:
Looks like this would need a deprecation period though:

```
std/parallelism.d(577): Error: getter properties must not return void
std/parallelism.d(605): Error: getter properties must not return void
std/parallelism.d(650): Error: getter properties must not return void
std/parallelism.d(934): Error: template instance `std.parallelism.Task!(run, void delegate())` error instantiating
std/parallelism.d(3587):        instantiated from here: `scopedTask!(void delegate())`
std/parallelism.d(577): Error: getter properties must not return void
std/parallelism.d(605): Error: getter properties must not return void
std/parallelism.d(650): Error: getter properties must not return void
std/parallelism.d(934): Error: template instance `std.parallelism.Task!(run, void delegate())` error instantiating
std/parallelism.d(3587):        instantiated from here: `scopedTask!(void delegate())`
std/regex/internal/thompson.d(73): Error: getter properties must not return void
```

jacob-carlborg:
> Looks like this would need a deprecation period though

That should be a given 😊.

wilzbach:
> Move test to compilable now we use deprecation

FWIW we typically use `-de` in the test, so that it can reside in fail_compilation. Though I don't think there's any style guide on this.

jmdavis:
Personally, I don't think that we should do anything with `@property` until it's officially decided what the overall plan for it is. There's no guarantee at this point that `@property` is even going to be left in the language. So, forcing anyone to change code related to `@property` which may or may not need to be changed later based on what is done with `@property` doesn't seem like a great idea - even more so if we make them "fix" code that uses `@property` only to then end up deprecating the feature entirely, resulting in the fix having been pointless.

First, we need to decide what we're actually going to do with `@property`. After that, maybe changes like this will be necessary and desirable, but for now, it would just be forcing folks to change code that relates to a half-baked feature which both Andrei and Walter have previously stated they thought was a mistake.

ntrel:
@jmdavis:
> There's no guarantee at this point that `@property` is even going to be left in the language

Removing `@property` **would break code**, *without any workaround* because there is no other way to make `typeof(someFunction)` not have a function type. For this reason I can't see W & A doing it with D2.

jmdavis:
> Removing @property would break code, without any workaround because there is no other way to make typeof(someFunction) not have a function type. For this reason I can't see W & A doing it with D2.

Personally,  I think that that aspect of `@property` was a huge mistake. It makes type introspection harder and has caused problems - especially because property functions can only act like variables in a very limited sense. Having the symbols lie about what they are just makes life harder.

But regardless, the point is that `@property` is in limbo. The original plans for what to do with it have been dropped, and Walter and Andrei have both expressed that they think that the feature was a mistake. At some point here, they're going to need to make a decision about what to do with it. That may involve some sort of enforcement about what sort of function it's put on, and it may not. It may involve solving the problem of properties that return callables, or it could involve dropping the feature entirely. We don't know. We can argue until the cows come home about what we think should or shouldn't be done with `@property`, but we don't know what Walter and Andrei are actually going to decide to do other than that they dislike the feature and probably want to get rid of it but also don't want to break code, making the course forward anything but clear. And as such, I think that it's a terrible idea to be mucking around with `@property` until a decision has actually been made. Writing a DIP on the matter could force the issue (and the current DIP on trying to add binary assignment to property functions may end up forcing the issue), but until then, actually making changes in the compiler would mean that we'd be changing a feature for which there is no real plan and no clear future.

Once that decision has been made, then maybe these changes will fit in with that wonderfully, and we can have these extra checks on `@property` functions, but until then, we'd just be causing problems for folks by telling them to change code that uses a half-baked feature that's really only used for documentation purposes at this point.

JinShil:
> Once that decision has been made

If @WalterBright and @andralex believe `@property` is a mistake, then it is on them to official declare it so by either updating the deprecated features table, or otherwise updating the documentation to discourage its use (Or asking one of us to do it).  Until then, `@property` is a legitimate feature of D, and we are obligated to support it.

This PR helps users employ `@property` correctly, and having D code out there with fewer oddities like a `void` getter is a good thing for everyone.  Having the compiler notify users when somethings wrong or odd helps them learn the language and makes for a more pleasant coding experience.  D already has enough oddities competing for mindshare; we need less, not more.

> Though I think we should update the spec to match the compiler error.

Yes, @ntrel Please add a spec update to ensure these new enforcements are properly documented.

jmdavis:
>  Until then, @property is a legitimate feature of D, and we are obligated to support it.

What feature? Basically all it does is affect the result of `typeof` and disallow you to overload functions that aren't marked with `@property` with those that are. It's just documented wishful thinking right now, not really an implemented feature. What it was originally intended to do was never actually implemented.

jmdavis:
Honestly, I think that this is one of those cases where @WalterBright just needs to weigh in. If he's fine with adding these checks, then okay, but otherwise, they shouldn't go in. They're mucking around with an incomplete and pseudo-canceled feature, so they're really more in DIP territory rather than a bug fix or simple improvement that we clearly want.

JinShil:
> but it was pretty much decided that we're not going to add property enforcement

No, it wasn't decided, because that phantom decision that you claim exists is not currently reflected in the documentation.  If you want any of that hearsay to hold water, it needs to be officially added somewhere in the documentation.

> These changes are forcing people to change their @property code on the assumption that @property is actually going to mean something and be enforced in some manner,

No.  It's forcing people to change their currently incorrect `@property` code based on the feature's current design and intended use.

> For now, it's forcing folks to change their code for an essentially unimplemented feature.

It's not unimplemented.  It's partially implemented.  And this PR provides enforcement for that part which is implemented.

> Honestly, I think that this is one of those cases where @WalterBright just needs to weigh in.

I would prefer that too, but @WalterBright and @andralex rarely weigh in on these *little* things.  And, until they do, and their decision is somehow reflected in the documentation, `@property` is a feature of D that should be supported.

> so they're really more in DIP territory rather than a bug fix or simple improvement that we clearly want.

No.  Deprecating `@property` and choosing to not support it any longer is DIP territory, and until then, or until @WalterBright or @andralex say otherwise, it should be supported.

jmdavis:
> No, it wasn't decided, because that phantom decision that you claim exists is not currently reflected in the documentation. If you want any of that hearsay to hold water, it needs to be officially added somewhere in the documentation.

It's not hearsay. It was decided quite some time ago that `@property` would not be implemented as originally intended. That's why the -property flag was removed. I won't disagree that the documentation should be improved, but claiming that the documentation is the end all and be all of the state of things is sadly not true much as work has been done to improve that situation. The documentation talks as if `@property` was how it was decided if a function can be used with property syntax and that's patently false. It doesn't have anything to do with it, much as that was at one time the intended design.

The problem with `@property` is that is was decided partway through that the intended design for it was being dropped, but it was never decided what exactly we're going to do with it, and it's been left in limbo for years, because it hasn't been a high enough priority for Walter and Andrei to actually decide what to do with it.

JinShil:
> It's not hearsay.

Then please prove it with something concrete.

Geod24:
I have to side with @jmdavis here. In my opinion, inaction is currently preferable to enforcing some rules that have been designed years ago.

As mentioned by @jmdavis , [there is previous work to go in exactly the other direction this PR is going](https://github.com/dlang/dmd/pull/4862)

JinShil:
For the record, I don't like the `@property` feature, I very much wish it were implemented differently or not implemented at all. That is why I have placed dlang/DIPs#97 on hold.  I also very much dislike the -betterC feature.  However, as a member of this team and steward of the language, I support these features, fix bugs with them, and try to improve their implementation despite my feelings about them, as doing so reflects positively on the language as a whole.

If the preference is for `@property` to be deprecated, discouraged, redesigned, discouraged, or otherewise changed, we need a concrete, documented decision saying so.  Until then `@property` is a feature of D that wall members of team DMD are obligated to support.

ntrel:
@Geod24:
>  there is previous work to go in exactly the other direction this PR is going

No, that pull just affected `-property`, not `@property`.

AndrewEdwards:
@WalterBright @andralex @atilaneves

Gentlemen, the mail is far overdue on @property. Request a discussion amongst yourselves and a formal decision made regarding the fate of this non-feature. Seven years is way more than enough time to make a decision. It's far better to [kill it](https://forum.dlang.org/post/kdqrnl$13ft$1@digitalmars.com) than to remain utterly indecisive about it.

ntrel:
@AndrewEdwards you can't kill property without breaking code using typeof. Breaking D2 code with no error for previously correct code is unacceptable.

AndrewEdwards:
@ntrel, point taken. In your opinion, what is the correct/acceptable solution moving forward? Is that to find a solution to improve `typeof` such that it no longer depends on `@property`? As far as I see it, the decision was already made. Accept the status quo! The problem is that because the leaders did not explicitly say so, the world is at a standstill. Therefore, I am asking for that decision.

12345swordy:
I opt to finished this implementation as @andralex still uses the property code in the druntime recently. Property needs a major code overall here to be useful as c# properties.

AndrewEdwards:
@WalterBright @andralex @atilaneves

Gentlemen, your input is requested.

wilzbach:
> @WalterBright @andralex @atilaneves
> 
> Gentlemen, your input is requested.

FYI: most people have GitHub notifications turned off as there's too much noise. So please don't assume that your pings will reach people. For certainty, old-school email is required.

atilaneves:
I always get notifications if I'm pinged.

In this instance I'll defer to @WalterBright

WalterBright:
@property is more or less a failure. But I don't want to break peoples' code that uses it in good faith. I recommend not doing anything with @property, document properly what it does now, and recommend people not use it for new code.

Fixing problems with it like this PR both breaks existing code and appears to encourage new code to be written with it.

12345swordy:
I see andralax write new code using @property in the druntime. I hate it to see it in a half bake state with no attempts to fix it.

andralex:
@12345swordy You mean the typeinfo stuff? I think it was more to stay in keeping with the existing code. Not to worry about.

12345swordy:
Real talk, why can't we just mark @property as deprecate with no remove date? I want actual c# like properties here.
If I have to write a dip on this, then I will do it myself.

andralex:
IMHO the best way (and the obvious first step) is to marginalize in the documentation.

AndrewEdwards:
CC @WalterBright @andralex @atilaneves @ntrel 

In order to summarize my understanding fully. The decision is made to:
    1) Cease all attempts to modify/update/upgrade the current behavior of User-Defined Properties
    2) Properly documentation the behavior as it is currently implemented in the language
    3) Suggest to the community NOT to use `@property` for future code.
    4) Issue a warning on use of `@property` ("Warning: Current implantation of @property is incomplete and is only retained for backward compatibility. Not to be used in new code!")
    5) Remove all uses of `@property` in the documentation where it is not absolutely necessary

To move forward we will need to update the content of the following pages:
    https://dlang.org/spec/function.html#property-functions
    https://dlang.org/spec/declaration.html#typeof
    https://dlang.org/spec/traits.html#getFunctionAttributes
    https://dlang.org/spec/grammar.html#Property
    https://dlang.org/spec/operatoroverloading.html#array-ops
    https://dlang.org/spec/abi.html#FuncAttrProperty

@wilzbach @jmdavis, request your assistance in making this happen. Once completed, this pull request will be closed.

ntrel:
Feel free to close this pull. I don't see any good argument for deprecating `@property`. It causes no harm to users, remove it only if D3 happens. For the toolchain, document that no further work will be done on it. Don't break people's code with no workaround for (possibly external) library code using typeof.

wilzbach:
There's nothing I can do here, hence I unassigned myself. This needs a decision from leadership.

If it's decided to deprecate @property, I'm happy to help.

Geod24:
As mentioned, there's no clear path for property. If a champion comes with a proper plan, timeline, and way to manage breakage, great. In the meantime, let's leave it be, there are bigger fish to fry.

PR #8321, opened 2018-06-01
[RFC] Allow passing a string literal to an extern(C) variadic parameter

ntrel:
As a string literal can be implicitly passed to a `const(char)*` parameter, passing one to an `extern(C)` variadic parameter can also be supported. In both cases, the string is cast to a pointer. (String literals are always `\0`-terminated). This is handy for C string compatibility.

I'll update the spec and changelog if comments are positive.

### DIFF
--- a/test/runnable/varargstringlit.d
+++ b/test/runnable/varargstringlit.d
@@ -1,0 +1,19 @@
+extern(C) void vf(int x, ...)
+{
+    import core.stdc.stdarg;
+    va_list args;
+
+    va_start(args, x);
+    auto s = va_arg!(const(char)*)(args);
+    assert(s[0..2] == "hi");
+    va_end(args);
+}
+
+void main()
+{
+    vf(0, "hi");
+
+    string s;
+    static assert(!__traits(compiles, vf(0, s)));
+}
+
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -880,6 +880,10 @@ private bool functionParameters(Loc loc, Scope* sc, TypeFunction tf, Type tthis,
                 }
                 if (tf.varargs == 1)
                 {
+                    // string literal -> implicit .ptr
+                    if (arg.toStringExp)
+                        arg = arg.castTo(sc, Type.tstring).castTo(sc, Type.tvoidptr);
+
                     const(char)* p = tf.linkage == LINK.c ? "extern(C)" : "extern(C++)";
                     if (arg.type.ty == Tarray)
                     {


kinke:
[This currently applies to non-`extern(D)`, e.g., `extern(C++)` as well.]

JinShil:
Seems like an obvious benefit to me, as long as it passes the test suite and has the appropriate doc updates.

test/runnable/varargstringlit.d
```diff
@@ -0,0 +1,19 @@
...
+    import core.stdc.stdarg;
+    va_list args;
+
+    va_start(args, x);
+    const(char)* s;
+    va_arg(args, s);
+    assert(s[0..2] == "hi");
+}
+
+unittest
```
wilzbach: Any reason you don't use `void main()` here?
`unittest` blocks don't automatically get run here.

ntrel: Oops, I have my editor set up to pass `-unittest -main`, and got in the habit of doing that instead. Fixed.

wilzbach: FYI you can do `./run.d runnable/varargstringlit.d` (almost as good)

ntrel:
The 64-bit targets are failing with this error:
```
Error: undefined identifier `__va_list_tag`
```

I don't know how to fix that.

ntrel:
@klickverbot:
> This change instead adds a special case to the D language

I disagree - from the compiler's point of view, it adds an extra line of code. But from the user and spec's point of view, [a `"string literal"` can implicitly convert to `immutable(char)*`](https://dlang.org/spec/expression.html#string_literals), so this removes an arbitrary restriction.

schveiguy:
Prior to 2.065, arrays were passed to C varargs functions as length/pointer pairs, such that they could be used with printf:
```D
import core.stdc.stdio;
void main()
{
   printf("hello, %*s\n", "world");
}
```
I'm not sure we should introduce a new behavior that is also surprising. The current restriction is meant to make you think how you really want to send an array into a C varags function, and is easily worked around by saying `.ptr`.

On the other hand, if we are being honest, the only purpose for C vararg support in D is for `printf`, and there's no other valid reason to pass in a string to printf other than to use it as a null-terminated `char *`.

ZombineDev:
The behavior prior to 2.065 (passing length and pointer) makes the most sense, IMO.

ntrel:
@klickverbot I've closed this as although slices are disallowed, a user could be unaware of this and pass a string literal expecting ptr + length. I accept the argument that this is a special case which could cause confusion. Also not having this means we are more flexible in case we decide to somehow allow passing slices again.

PR #8322, opened 2018-06-02
convert evalu8.c to evalu8.d

WalterBright:


rainers:
Building on Appveyor fails because the D part assumes `real` exists on the C++ side, but that uses longdouble_soft.
As this is only a temporary problem until the other C files are translated aswell, it's probably easiest to just force the correct mangling with `pragma(mangle)` on a couple of functions.

rainers:
Affected functions seem to be `el_toldouble`, `_modulo` and `Complex_ld::abs`.

rainers:
Another option is to use `longdouble_soft` in D instead of `real` whenever version CRuntime_Microsoft is active. That needs a tiny change for y2lx, too.

WalterBright:
Since it'll all be converted to D anyway, I figured might as well go ahead and convert bcomplex.c to D in this PR.

WalterBright:
Appveyor failures:
```
unittest
core.exception.AssertError@src\core\math.d(116): unittest failure
----------------
0x01114497
0x011AA5D3 in gc_clrProxy
core.exception.AssertError@src\rt\util\typeinfo.d(250): Assertion failure
----------------
0x01114441
0x011AA5D3 in gc_clrProxy
2/42 unittests FAILED
--- errorlevel 1
```
which have nothing to do with this PR. @rainers what's going on?

WalterBright:
well, maybe they do :-(

src/dmd/backend/bcomplex.d
```diff
@@ -24,28 +22,291 @@ struct Complex_f
...
+        }
+        return c;
+    }
 }
 
+
 struct Complex_ld
 {
-    targ_ldouble re, im;
+    real re, im;
```
rainers: You have to keep `targ_ldouble` here because when built with LDC on Windows, `real` is the same as `double`.

WalterBright:
@rainers are you able to replicate the AppVeyor failures? I can't on my machine.

src/dmd/backend/bcomplex.d
WalterBright: ok

WalterBright:
> You have to keep targ_ldouble here because when built with LDC on Windows, real is the same as double.

Making that change causes semaphoreci to fail with:

```
gdmd -c -of../generated/linux/release/64/divcoeff.o  -version=MARS -fPIC -w -de -O -release -inline -m64  dmd/backend/divcoeff.d
dmd/backend/bcomplex.d:217:1: error: struct dmd.backend.bcomplex.Complex_ld no size yet for forward reference
 struct Complex_ld
 ^
dmd/backend/cdef.d:798:1: error: union dmd.backend.cdef.eve no size yet for forward reference
 union eve
 ^
````

File: src/dmd/e2ir.d [hunk 2]
```diff
@@ -1479,8 +1479,8 @@ elem *toElem(Expression e, IRState *irs)
                     break;
 
                 case TYcldouble:
-                    e.EV.Vcldouble.re = re;
-                    e.EV.Vcldouble.im = im;
+                    e.EV.Vcldouble.re = cast(real)re;
+                    e.EV.Vcldouble.im = cast(real)im;
```
rainers: This cast is no longer necessary and is harmful to the LDC build, too.

WalterBright: ok, gone

src/dmd/backend/evalu8.d
```diff
@@ -1876,4 +1898,51 @@ elem * evalu8(elem *e, goal_t goal)
   return e;
 }
 
-#endif /* !SPP */
+/******************************
+ * This is the same as the one in el.c, but uses native D reals
+ * instead of the soft long double ones.
+ */
+
+extern (D) real el_toldoubled(elem *e)
```
rainers: Sorry, still more uses of `real` instead of `targ_ldouble`...

WalterBright: fixed

src/dmd/backend/evalu8.d
```diff
@@ -1876,4 +1898,51 @@ elem * evalu8(elem *e, goal_t goal)
...
+    }
+    return result;
+}
+
+/***************************************
+ * Copy of _modulo from fp.c. Here to help with linking problems.
+ */
+version (CRuntime_Microsoft)
+{
+    extern (D) private real _modulo(real x, real y)
```
rainers: I think the casts should go into this function instead of the call.

src/dmd/backend/evalu8.d
```diff
@@ -6,46 +6,51 @@
...
-#if _MSC_VER
-#define isnan _isnan
-#endif
-
-#include        "cc.h"
-#include        "oper.h"                /* OPxxxx definitions           */
-#include        "global.h"
-#include        "el.h"
-#include        "type.h"
+version (SPP)
```
rainers: BTW: no module name here? (I won't object to module names without package if that's that's how you want to do it ;-) )

src/dmd/backend/evalu8.d
WalterBright: all right

src/dmd/backend/evalu8.d
WalterBright: No imports of it yet, I'll probably worry about that when imports happen. Right now, I just want to get it to work :-)

rainers:
Still too many uses of `real`. Shall I push a fix?

WalterBright:
Yes, I could use your help. I'm mostly just floundering about on these platforms I have little knowledge of.

rainers:
> Yes, I could use your help. I'm mostly just floundering about on these platforms I have little knowledge of.

Done. Now let's see the other platforms complain...

rainers:
This failed because comparison of two `targ_ldouble`s is forwarded to `longdouble_soft.opCmp`, but that cannot return "false" for all comparisons involving `nan`. Now added explicit checks.

WalterBright:
It's all green and ready to pull! Thanks @rainers for your invaluable help. ping @JinShil @wilzbach @jacob-carlborg ?

src/dmd/backend/bcomplex.d
```diff
@@ -4,14 +4,15 @@
  *
  * Copyright:   public domain
  * License:     public domain
- * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/bcomplex.d, backend/_bcomplex.d)
+ * Source:      $(DMDSRC backend/_bcomplex.d)
```
wilzbach: FYI underscore escaping is no longer needed. We enabled the opt-out about two months ago.
See: https://github.com/dlang/dlang.org/pull/2307

src/dmd/backend/evalu8.d
```diff
@@ -82,9 +89,7 @@ int boolres(elem *e)
                 case TYbool:
                 case TYwchar_t:
                 case TYenum:
-#if !MARS
                 case TYmemptr:
-#endif
```
wilzbach: Shouldn't this result in a `version(MARS){} else { ... }` ?

src/dmd/backend/evalu8.d
```diff
@@ -275,33 +282,37 @@ int iffalse(elem *e)
...
     targ_ldouble d1,d2;
-    elem esave;
+    elem esave = void;
+
+    static bool unordered(T)(T d1, T d2) { return isnan(d1) || isnan(d2); }
 
     assert((statusFE() & 0x3800) == 0);
-    assert(e && EOP(e));
-    op = e->Eoper;
+    assert(e && !OTleaf(e.Eoper));
```
wilzbach: Why not define EOP or EBIN as a function?

JinShil (approved):

src/dmd/backend/bcomplex.d
wilzbach: Also last year you made a point of using absolute links here as they are clickable (see https://github.com/dlang/dmd/pull/7030).
Though I was a fan of DMDSRC, so I don't mind ;-)

wilzbach (approved):
Nice!

src/dmd/backend/evalu8.d
WalterBright: I don't think those macros paid off.

src/dmd/backend/evalu8.d
WalterBright: I decided that having too many version statements was clutter, and so removed the ones that weren't really necessary.

src/dmd/backend/bcomplex.d
WalterBright: yes, it should be the full version.

src/dmd/backend/bcomplex.d
```diff
@@ -24,28 +25,291 @@ struct Complex_f
...
+    {
+        Complex_f q;
+        targ_ldouble r;
+        targ_ldouble den;
+
+        if (fabs(y.re) < fabs(y.im))
+        {
+            r = y.re / y.im;
+            den = y.im + r * y.re;
+            q.re = cast(float)((x.re * r + x.im) / den);
```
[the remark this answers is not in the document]
WalterBright: they make it clearer what is happening

JinShil:
It also appears bcomplex.c was not deleted in this PR.  Shouldn't it be?

src/dmd/backend/bcomplex.d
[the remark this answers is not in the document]
rainers: and the casts are necessary if targ_ldouble resolves to longdouble_soft, as there is no implicit conversion to float/double then.

PR #8324, opened 2018-06-02
Allow classes and interfaces with only static members to be used in -betterC

JinShil:
This is a followup to https://github.com/dlang/dmd/pull/8204, but for -betterC.

https://github.com/dlang/druntime/pull/2194 was merged, which prevents -betterC from importing symbols that -betterC cannot utilize.  This means the logic added for https://github.com/dlang/dmd/pull/8204 can now be utilized for -betterC builds as well.

However, with -betterC builds, object.d is still imported and parsed in its entirety; `TypeInfo`, `ModuleInfo`, and other unsupported features are just versioned out.  Therefore, I had to move the logic for setting the special `ClassDeclaration`s, out of the parsing phase and into the symbol semantic phase.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -783,8 +783,6 @@ private int tryMain(size_t argc, const(char)** argv)
     if (global.errors)
         fatal();
 
-    backend_init();
-
     // Do semantic analysis
     foreach (m; modules)
     {
@@ -792,6 +790,9 @@ private int tryMain(size_t argc, const(char)** argv)
             message("semantic  %s", m.toChars());
         m.dsymbolSemantic(null);
     }
+
+    backend_init();
+
     //if (global.errors)
     //    fatal();
     Module.dprogress = 1;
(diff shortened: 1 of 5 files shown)


src/dmd/mars.d
```diff
@@ -783,15 +783,16 @@ private int tryMain(size_t argc, const(char)** argv)
...
-
     // Do semantic analysis
     foreach (m; modules)
     {
         if (global.params.verbose)
             message("semantic  %s", m.toChars());
         m.dsymbolSemantic(null);
     }
+
+    backend_init();
```
JinShil: The backend is dependent on whether `Throwable` is declared, so we have to move its initialization until after `dsymbolSemantic`.

JinShil:
Gah! This is not going to work if we want to allow using `TypeInfo` at compile-time.  https://issues.dlang.org/show_bug.cgi?id=18472

I think if we had a `version(CTFE)` feature, it could be done, but until then, I'm giving up on this.

PR #8326, opened 2018-06-03
fix Issue 18936 - Internal error: dmd/backend/cgxmm.c 684

WalterBright:
This code fails to compile with 2.079, see if the failure still exists in master.

### DIFF
--- a/src/dmd/backend/gsroa.c
+++ b/src/dmd/backend/gsroa.c
@@ -63,6 +63,8 @@ static void sliceStructs_Gather(SymInfo *sia, elem *e)
                     {
                         if (!sia[si].accessSlice)
                         {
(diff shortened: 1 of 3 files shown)


WalterBright:
Filed as https://issues.dlang.org/show_bug.cgi?id=18936

RazvanN7:
Since this is filed as a bug report and the test is contained in the OP, can we close this? The person that will solve the bug will surely add this test in the fixing PR

WalterBright:
I'm not in a rush to close it, I'd rather wait until there's a fix for it and then decide what to do about it.

WalterBright:
Added the fix.

RazvanN7 (approved):

PR #8329, opened 2018-06-04
relax assert about return value from GetFullPathNameW

rainers:
This assert triggers if the processed filename contains relative path components, e.g. `c:\a\b\c\d\..\e.d`. The the initial call to GetFullPathNameW that returns the required buffer length seems to keep them, while the actual call to retrieve the full path eliminates `d\..\`.


RazvanN7:
Can you add a test that failed before this addition but is now passing?

RazvanN7 (approved):

PR #8330, opened 2018-06-04
fix issue 18928 - extern(C++) bad codegen on win64, wrong calling convention

rainers:
With the VC x64 ABI, a member function always returns a struct on the stack no matter its size.

This doesn't yet take into account that delegates are still incompatible. To avoid having to create another linkage attribute, a stub should be generated in that case to convert the ABIs.

### DIFF
--- a/src/dmd/s2ir.d
+++ b/src/dmd/s2ir.d
@@ -719,7 +719,7 @@ private extern (C++) class S2irVisitor : Visitor
             assert(func.type.ty == Tfunction);
             TypeFunction tf = cast(TypeFunction)(func.type);
 
-            RET retmethod = retStyle(tf);
+            RET retmethod = retStyle(tf, func.needThis());
             if (retmethod == RET.stack)
             {
                 elem *es;
@@ -745,7 +745,7 @@ private extern (C++) class S2irVisitor : Visitor
                         Type t = ce.e1.type.toBasetype();
                         if (t.ty == Tdelegate)
                             t = t.nextOf();
-                        if (t.ty == Tfunction && retStyle(cast(TypeFunction)t) == RET.stack)
+                        if (t.ty == Tfunction && retStyle(cast(TypeFunction)t, ce.f && ce.f.needThis()) == RET.stack)
                         {
                             irs.ehidden = el_var(irs.shidden);
                             e = toElemDtor(s.exp, irs);
@@ -769,7 +769,7 @@ private extern (C++) class S2irVisitor : Visitor
                         Type t = ce.e1.type.toBasetype();
                         if (t.ty == Tdelegate)
                             t = t.nextOf();
-                        if (t.ty == Tfunction && retStyle(cast(TypeFunction)t) == RET.stack)
+                        if (t.ty == Tfunction && retStyle(cast(TypeFunction)t, fd && fd.needThis()) == RET.stack)
                         {
                             irs.ehidden = el_var(irs.shidden);
                             e = toElemDtor(s.exp, irs);
--- a/src/dmd/target.d
+++ b/src/dmd/target.d
@@ -21,6 +21,7 @@ import dmd.dmodule;
 import dmd.dstruct;
 import dmd.dsymbol;
 import dmd.expression;
+import dmd.func;
 import dmd.globals;
 import dmd.id;
 import dmd.identifier;
@@ -589,12 +590,15 @@ struct Target
     }
 
     /**
+     * Determine return style of function - whether in registers or
+     * through a hidden pointer to the caller's stack.
      * Params:
      *   tf = function type to check
+     *   needsThis = true if the function type is for a non-static member function
      * Returns:
      *   true if return value from function is on the stack
      */
-    extern (C++) static bool isReturnOnStack(TypeFunction tf)
+    extern (C++) static bool isReturnOnStack(TypeFunction tf, bool needsThis)
     {
         if (tf.isref)
         {
@@ -621,6 +625,8 @@ struct Target
                 StructDeclaration sd = (cast(TypeStruct)tns).sym;
                 if (sd.ident == Id.__c_long_double)
                     return false;
+                if (tf.linkage == LINK.cpp && needsThis)
+                    return true;
                 if (!sd.isPOD() || sz > 8)
                     return true;
                 if (sd.fields.dim == 0)
@@ -638,6 +644,8 @@ struct Target
                 StructDeclaration sd = (cast(TypeStruct)tb).sym;
                 if (sd.ident == Id.__c_long_double)
                     return false;
+                if (tf.linkage == LINK.cpp && needsThis)
+                    return true;
             }
         }
 
--- a/src/dmd/toir.d
+++ b/src/dmd/toir.d
@@ -1029,9 +1029,14 @@ void buildClosure(FuncDeclaration fd, IRState *irs)
 /***************************
  * Determine return style of function - whether in registers or
  * through a hidden pointer to the caller's stack.
+ * Params:
+ *   tf = function type to check
+ *   needsThis = true if the function type is for a non-static member function
+ * Returns:
+ *   RET.stack if return value from function is on the stack, RET.regs otherwise
  */
-RET retStyle(TypeFunction tf)
+RET retStyle(TypeFunction tf, bool needsThis)
 {
     //printf("TypeFunction.retStyle() %s\n", toChars());
-    return Target.isReturnOnStack(tf) ? RET.stack : RET.regs;
+    return Target.isReturnOnStack(tf, needsThis) ? RET.stack : RET.regs;
 }
--- a/src/dmd/e2ir.d
+++ b/src/dmd/e2ir.d
@@ -267,7 +267,7 @@ private elem *callfunc(const ref Loc loc,
     objc.setupMethodSelector(fd, &esel);
     objc.setupEp(esel, &ep, left_to_right);
 
-    const retmethod = retStyle(tf);
+    const retmethod = retStyle(tf, fd && fd.needThis());
     if (retmethod == RET.stack)
     {
         if (!ehidden)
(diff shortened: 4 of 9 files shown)


TurkeyMan:
> To avoid having to create another linkage attribute...

I've had a bug in for 5 years that `extern(C++) delegate` should be supported. C++ and D methods may have a different calling convention from each other (as demonstrated here!) and as such, delegates will also need to support the calling convention attribute...

TurkeyMan:
Oh no, test doesn't work :(

JinShil:
I think this should target stable.

WalterBright:
> I've had a bug in for 5 years that extern(C++) delegate should be supported.

Last time I checked, C++ doesn't have delegates. Member function pointers are something different.

TurkeyMan:
D has delegates, and you probably want to assign a method to a delegate, since that's what they're for.

TurkeyMan:
Rainers idea about creating a little shim function is probably better though...

kinke:
There's pretty new `Target.isReturnOnStack()`, which gets a `TypeFunction` only right now, which will also need to be adapted.

kinke:
Oh and this also applies to Win32.

rainers:
> Oh no, test doesn't work :(

That was the 32-bit test. Added the condition for m32mscoff, too. Interfacing with dmc seems to be fine without the change.

> The tests here should prove that both cases generate code to match C++'s expectations.

Added tests for virtual, final and static.

rainers:
> I think this should target stable.

I wasn't so sure as it's a bit experimental (e.g. due to lack of delegate support). But I guess Manu is currently working with git-HEAD anyway...

BTW: There is currently no check for extern(C++), so this also changes the ABI for extern(D) in line with "extern(D) matches what the C compiler does (but for Windows x86)", but there is no info about member functions and C++ in the spec. What is the desired behaviour?

kinke:
> There is currently no check for extern(C++)

I was just about to mention that. I'm for not blindly applying legacy C++ ABI stupidities like this one to `extern(D)`. The params order is reversed anyway, so D code interfacing with other languages will almost always imply `extern(C[++])`.

rainers:
> I'm for not blindly applying legacy C++ ABI stupidities like this one to extern(D).

Ok, I've limited it to extern(C++) now.

jacob-carlborg:
Passing `TypeFunction` and `FuncDeclaration` as separate arguments risks to pass two unrelated values. Is it possible to pass only `FuncDeclaration`?

jacob-carlborg:
> Unfortunately not as there are situations where only the type is available

I had a quick look, when it comes to `__traits(isReturnOnStack)`, it has a `FuncDeclaration` [1]. A delegate has a `FuncDeclaration` as well [2]. Not sure what `tytym` is.

[1] https://github.com/dlang/dmd/blob/d94c02a983438f6a2574193e1c8b05455e80eaa4/src/dmd/traits.d#L1143

[2] https://github.com/dlang/dmd/blob/d94c02a983438f6a2574193e1c8b05455e80eaa4/src/dmd/expression.d#L5540

rainers:
> when it comes to __traits(isReturnOnStack), it has a FuncDeclaration

Yes, but it is not always set by `toTypeFunction`, e.g. when you pass a type to the trait.

> A delegate has a FuncDeclaration as well [2]

That covers explicit delegates, but a variable of TypeDelegate has no way to know at compile time what function symbol is actually called.

jacob-carlborg:
Ok.

TurkeyMan:
How's this going? I'd really love to see this in a nightly sooner than later, then I have some colleagues who can try it out.

src/dmd/target.d
```diff
@@ -586,10 +587,11 @@ struct Target
     /**
      * Params:
      *   tf = function type to check
+     *   fd = declaration of function to check (null if unavailable, i.e. just checking type)
      * Returns:
      *   true if return value from function is on the stack
      */
-    extern (C++) static bool isReturnOnStack(TypeFunction tf)
+    extern (C++) static bool isReturnOnStack(TypeFunction tf, FuncDeclaration fd)
```
JinShil: This doesn't seem right; taking both a `TypeFunction` and a `FunctionDeclaration`.  I don't have a solution at the moment, but my intuition tells me there must be a better way.

TurkeyMan: Doesn't really feel very offensive to me. I've seen lots of functions over the years follow this pattern for the same reasons.

jacob-carlborg: We have gone through this above https://github.com/dlang/dmd/pull/8330#issuecomment-394592022. I would like a better solution as well.

rainers: > fd.isMember2() && !fd.isStatic()

Just noticed that there is `fd.needThis()` for this check.

JinShil: > `(fd && Target.isReturnOnStack(fd)) || (!fd && Target.isReturnOnStack(tf))`

That looks very idiomatic-DMD, actually.  I would prefer an even better solution, but if I had to choose between that and the current proposal, I'd go with that.

rainers: If we use overloads, I would replace `retStyle()` with `Target.returnsOnStack()` to avoid having 4 similar functions.

JinShil: Sounds good to me.  I don't see why we need both `retStyle` and `returnsOnStack`.

rainers:
Rebased and changed as recommended. I think it's worse on both caller and callee side, though.

JinShil (dismissed):

kinke:
I'm not a fan of the two `isReturnOnStack()` overloads either.
The `FuncDeclaration` overload is too bloated IMO, needlessly duplicating some logic from the later called `TypeFunction` overload. (I think the complex32 case is wrong, plus the non-power-of-2 sizes aren't handled correctly either, so fixing it in one place only is preferrable).

It also looks like the tests got lost. The previous ones are [here](https://github.com/ldc-developers/dmd-testsuite/commit/905d7139e42e9f415d32a1345e8880311e8a057d), just in case. ;)

rainers:
> It also looks like the tests got lost. The previous ones are here, just in case. ;)

Oops, will add them back.

I'll give this version another chance by moving the previous version with two parameters into a private function to be called by the public functions.

> I think the complex32 case is wrong, plus the non-power-of-2 sizes aren't handled correctly either

Do you mean in the existing implementation? Without the complex32 condition other tests failed, but maybe the tests are broken, too...

rainers:
Tests are now back and code duplication avoided. The actual changes are in the second commit.

kinke:
> There's something wrong with this design if it reqires a function taking TypeFunction and FunctionDeclaration

What's wrong is basically MS' decision to treat C++ methods in a special way ABI-wise, not just wrt. position of `this` pointer (1st arg, even before hidden `sret` pointer, unlike regular functions and unlike all other ABIs supported by LDC; for the 32-bit ABI, it must be in `EAX`, `__thiscall`, another method specialty), but also wrt. this special `sret` rules for *all* struct return types.

This makes a function type (TypeFunction) not suffice to fully establish the ABI in general, but only for MSVC targets and `extern(C++)`. While the FuncDeclaration is available in most cases, a delegate is in bad luck (and will probably need an extra shim function with `extern(D)` ABI-signature forwarding to the actual C++ method), as are all attempts to fake an arbitrary (incl. `extern(C++)`) method call reliably and portably in D via a function pointer.

So I think the latest version now with private impl taking both params is fine and a necessary evil to deal with the ABI mess.

src/dmd/target.d
```diff
@@ -849,3 +719,164 @@ private Expression decodeReal(const ref Loc loc, Type type, ubyte* buffer)
...
 }
+
+/**
+ * Params:
+ *   tf = function type to check
+ *   fd = function to check (optional)
+ * Returns:
+ *   true if return value from function is on the stack
+ */
+private bool isReturnOnStackImpl(TypeFunction tf, FuncDeclaration fd)
```
JinShil: It doesn't look like you need `FunctionDeclaration`, you just need `fd.needThis()`.  You could potentially make it a `bool needsThis` to remove the dependency.

src/dmd/s2ir.d
```diff
@@ -769,7 +769,8 @@ private extern (C++) class S2irVisitor : Visitor
                         Type t = ce.e1.type.toBasetype();
                         if (t.ty == Tdelegate)
                             t = t.nextOf();
-                        if (t.ty == Tfunction && retStyle(cast(TypeFunction)t) == RET.stack)
+                        if (t.ty == Tfunction &&
```
JinShil: You could then obtain do something like this at the call site...

```D
bool needsThis = fd && fd.needThis(();
TypeFunction tf = fd ? cast(TypeFunction)fd.type : t.ty;
Target.isReturnOnStack(tf, needsThis);
```

That's probably not correct (I haven't fully analyzed the code), but I hope it illustrates the idea.  Just get the `TypeFunction` and `needsThis` at the call site, and then make the call to `isReturnOnStack(tf, needsThis)`.

rainers: That mostly brings us back to the initial version ;-) Agreed, passing `fd.needThis` instead of `fd` is the better API as it doesn't pass the same information twice which can cause ambiguities. It has to evaluate `needThis` eagerly, though. Considering all the stuff happening in isReturnOnStack, that's probably not a big deal.
I'll give that version another shot...

TurkeyMan:
Since the initial concern was the function receives `TypeFunction` and `FuncDeclaration`, and that they might mismatch, since it receives both.
It feels like a pretty simple solution to me; first line of the function: `assert(fd.type == tf);`... now the user has not passed disassociated values. Problem solved?

src/dmd/toir.d
```diff
@@ -1030,8 +1030,8 @@ void buildClosure(FuncDeclaration fd, IRState *irs)
  * Determine return style of function - whether in registers or
  * through a hidden pointer to the caller's stack.
  */
-RET retStyle(TypeFunction tf)
+RET retStyle(TypeFunction tf, bool needsThis)
```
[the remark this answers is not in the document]
rainers: The function did not document its parameter or return value so far. I'd rather remove it as in the previous version, but probably better in another PR.

rainers:
> I agree with @jacob-carlborg that the documentation should be completed even if it wasn't done before.

Ok, I took the best parts of the two documentation headers and added them to both functions.

src/dmd/toir.d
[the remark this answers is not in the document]
jacob-carlborg: Ah, ok. I only looked at the diff, I did the review on the subway on my phone. It’s a bit difficult to look at the whole file on the phone.

Geod24 (approved):

PR #8331, opened 2018-06-04
Fix Issue 17580 - Marking methods as synchronized is allowed despite spec

RazvanN7:
Looks like many tests took advantage of this spec violation. Maybe a deprecation cycle first?

### DIFF
--- a/changelog/fix17580.dd
+++ b/changelog/fix17580.dd
@@ -1,0 +1,12 @@
+Deprecate marking class methods as `synchronized` while class is not synchronized
+
+The D language specification states that member functions of non-`synchronized` classes cannot be marked as `synchonized`, however the compiler does not reject such declarations. The following code compiled before the patch:
+
+---
+class S
+{
+    synchronized void fun() { }
+}
+---
+
+After applying the patch, the above code will issue a deprecation message
--- a/changelog/fix17580.dd
+++ b/changelog/fix17580.dd
@@ -1,6 +1,6 @@
 Deprecate marking class methods as `synchronized` while class is not synchronized
 
-The D language specification states that member functions of non-`synchronized` classes cannot be marked as `synchonized`, however the compiler does not reject such declarations. The following code compiled before the patch:
+The D language specification states that member functions of non-`synchronized` classes cannot be marked as `synchronized`, however the compiler does not reject such declarations. The following code compiled before this release:
 
 ---
 class S
@@ -9,4 +9,20 @@ class S
 }
 ---
 
-After applying the patch, the above code will issue a deprecation message
+To fix the deprecation message, make the class synchronized:
+
+---
+synchronized class S
+{
+    void fun() { }
+}
+---
+
+or mark the methods as `shared`:
+
+---
+class S
+{
+    shared void fun() { }
+}
+---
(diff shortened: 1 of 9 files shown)


jacob-carlborg:
> Maybe a deprecation cycle first?

Yes.

RazvanN7:
Transformed error into deprecation and added changelog entry.

changelog/fix17580.dd
```diff
@@ -0,0 +1,12 @@
...
+The D language specification states that member functions of non-`synchronized` classes cannot be marked as `synchonized`, however the compiler does not reject such declarations. The following code compiled before the patch:
+
+---
+class S
+{
+    synchronized void fun() { }
+}
+---
+
+After applying the patch, the above code will issue a deprecation message
```
jacob-carlborg: Should contain an example how to fix the code.

src/dmd/semantic3.d
```diff
@@ -1083,6 +1083,8 @@ private extern(C++) final class Semantic3Visitor : Visitor
                     ClassDeclaration cd = funcdecl.isThis() ? funcdecl.isThis().isClassDeclaration() : funcdecl.parent.isClassDeclaration();
                     if (cd)
                     {
+                        if (!(cd.storage_class & STC.synchronized_))
+                            funcdecl.deprecation("cannot be marked as `synchronized` because it is a member of the non-`synchronized` class `%s`. The `synchronized` attribute must be applied to the class declaration itself", cd.toPrettyChars());
```
wilzbach: Needs a deprecation marker comment.
Check https://github.com/dlang/DIPs/blob/master/DIPs/DIP1013.md

changelog/fix17580.dd
```diff
@@ -0,0 +1,12 @@
+Deprecate marking class methods as `synchronized` while class is not synchronized
+
+The D language specification states that member functions of non-`synchronized` classes cannot be marked as `synchonized`, however the compiler does not reject such declarations. The following code compiled before the patch:
```
wilzbach: Typo: synchronized

wilzbach:
Addressed my review myself.

BTW I think we should fix the spec here instead as DMD already does the correct thing and many people might already depend on it (even I used to do this in the beginning of my D time).
So the following example:

```d
class Foo
{
    synchronized void foo(){}
}
```

Does correctly compile with monitor calls:

```
shared void example.Foo.foo():
  push rbx
  mov rbx, rdi
  call _d_monitorenter@PLT
  mov rdi, rbx
  pop rbx
  jmp _d_monitorexit@PLT
```

https://godbolt.org/g/rz7VCY

JinShil:
@WalterBright or @andralex I think there is some ambiguity about what the intention with the language is here.  In other words, is the spec wrong, or is the implementation wrong.  Please clarify.

FeepingCreature:
Also note that without synchronized methods, invariants in multithreaded programs are basically useless, because even if you synchronize the invariant, the class will release and reacquire the lock, meaning that anybody could do anything to the variable in the meantime.

jmdavis:
@andralex 

> Here we want to allow people the flexibility of defining methods that do their own synchronization, so I suggest we change the spec to match the current semantics. cc @WalterBright

How on earth would anyone be able to use their own synchronization with synchronized methods? synchronized uses the monitor object in such cases. Methods can do their own synchronization internally, but unless I'm missing somethings, we'd have to make a significant change to how synchronized works to allow methods mark with synchronized to do their own synchronization.

PR #8332, opened 2018-06-05
 Fix issue 18922: No substitutions for C++ namespaces in different module/file

Geod24:
```
DMD was just comparing the object by their reference instead of their string value.
Obviously when a namespace is declared in another module, the reference is different.
```

There are two other small refactors in this PR, which can be taken out at will.
But since they are different commits, they shouldn't be a problem either for `git revert` or reviewers (juse review commit by commit and/or ignore whitespace).

Also tested it on a large C++ project I am porting to D (on Linux x64).

### DIFF
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -36,6 +36,7 @@ import dmd.globals;
 import dmd.id;
 import dmd.identifier;
 import dmd.mtype;
+import dmd.nspace;
 import dmd.root.outbuffer;
 import dmd.root.rootobject;
 import dmd.target;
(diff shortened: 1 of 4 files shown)


Geod24:
Added a test case for Windows, which was already working

src/dmd/dsymbol.d
```diff
@@ -243,6 +243,8 @@ extern (C++) class Dsymbol : RootObject
     {
         if (this == o)
             return true;
+        if (o.dyncast() != DYNCAST.dsymbol)
```
rainers: Is this extra test necessary? Your new call for namespaces already verifies that a Dsymbol is passed. Or did you see other failures here?

Geod24: It's not strictly necessary no. But when developing and trying out other solutions, I was surprised that the argument was blindly casted, which resulted in some memory corruption. It wouldn't surprise me if it would hit other people in the future.

rainers: Ok. It seems a similar check exists in about half of the `equals` oerloads, but misses in the other half. This base class is unlikely to be hit that often, so the extra check should be ok.

src/dmd/cppmangle.d
```diff
@@ -124,12 +125,24 @@ private final class CppMangleVisitor : Visitor
 
     /******
      * See if `p` exists in components[]
+     *
+     * Note that components can contains `null` entry,
```
RazvanN7: can **contain** `null` **entries**

src/dmd/cppmangle.d
```diff
@@ -138,6 +151,40 @@ private final class CppMangleVisitor : Visitor
         return -1;
     }
 
+    /**
+     * Overload which accepts a Namespace
+     *
+     * It is very common for large C++ projects to have multiple files sharing
+     * the same `namespace`. If any D project adopts the same approach
+     * (e.g. separating data structures from functions), it will lead to two
+     * `Nspace` object being instantiated, with different address.
```
RazvanN7: `Nspace` **objects** being ... with different **addresses**

src/dmd/cppmangle.d
```diff
@@ -138,6 +151,40 @@ private final class CppMangleVisitor : Visitor
...
+     *
+     * Returns:
+     *  Index of the entry, if found, or `-1` otherwise
+     */
+    int find(Nspace ns)
+    {
+        foreach (i, component; components)
+        {
+            if (ns == component)
+                return cast(int)i;
```
RazvanN7: Is this cast really necessary? I would argue it is not, but maybe I am wrong

RazvanN7 (approved):
LGTM except for the documentation nits

src/dmd/cppmangle.d
Geod24: Well it was there before, I just followed the existing style.
I would much prefer to return `components.length` and use `size_t`, but I'll leave that for another refactor...

Geod24:
Fixed typos

src/dmd/cppmangle.d
```diff
@@ -138,6 +151,40 @@ private final class CppMangleVisitor : Visitor
...
+     * See_Also:
+     *  https://issues.dlang.org/show_bug.cgi?id=18922
+     *
+     * Params:
+     *   ns = C++ namespace to do substitution for
+     *
+     * Returns:
+     *  Index of the entry, if found, or `-1` otherwise
+     */
+    int find(Nspace ns)
```
RazvanN7: It seems that this method is used solely inside nspace.d. If that is the case, make it private. Also, is there a possibility that the parameter can be made `const`? It doesn't seem to be modified.

Geod24: The whole class is `private final`.

Also for the parameter to be `const`, `const` needs to be added to `Nspace.equals`.
I think it should also be added to all base class (in D it wouldn't be a problem, but I'm not sure how that would play with the C++ header).

Geod24:
Looks like SemaphoreCI is broken (@wilzbach )...

PR #8334, opened 2018-06-05
Posix extern(C++) operator mangling

TurkeyMan:
Supplement the prior PR that introduced `extern(C++)` operator mangling.

### DIFF
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -45,6 +45,21 @@ import dmd.visitor;
 
 extern (C++):
 
+// helper to check if an identifier is a C++ operator
+enum CppOperator { Cast, Assign, Eq, Index, Call, Unary, Binary, OpAssign, Unknown }
+package CppOperator isCppOperator(Identifier id)
+{
+    __gshared const(Identifier)[] operators = null;
+    if (!operators)
+        operators = [Id._cast, Id.assign, Id.eq, Id.index, Id.call, Id.opUnary, Id.opBinary, Id.opOpAssign];
+    foreach (i, op; operators)
+    {
+        if (op == id)
+            return cast(CppOperator)i;
+    }
+    return CppOperator.Unknown;
+}
+
 const(char)* toCppMangleItanium(Dsymbol s)
 {
     //printf("toCppMangleItanium(%s)\n", s.toChars());
@@ -163,18 +178,111 @@ private final class CppMangleVisitor : Visitor
     }
 
     /******************************
+     * Write the mangled representation of a template argument.
+     * Params:
+     *  ti  = the template instance
+     *  arg = the template argument index
+     */
+    void template_arg(TemplateInstance ti, size_t arg)
+    {
+        TemplateDeclaration td = ti.tempdecl.isTemplateDeclaration();
+        assert(td);
+        TemplateParameter tp = (*td.parameters)[arg];
+        RootObject o = (*ti.tiargs)[arg];
+
+        if (tp.isTemplateTypeParameter())
+        {
+            Type t = isType(o);
+            assert(t);
+            t.accept(this);
+        }
+        else if (TemplateValueParameter tv = tp.isTemplateValueParameter())
+        {
+            // <expr-primary> ::= L <type> <value number> E  # integer literal
+            if (tv.valType.isintegral())
+            {
+                Expression e = isExpression(o);
+                assert(e);
+                buf.writeByte('L');
+                tv.valType.accept(this);
+                auto val = e.toUInteger();
+                if (!tv.valType.isunsigned() && cast(sinteger_t)val < 0)
+                {
+                    val = -val;
+                    buf.writeByte('n');
+                }
+                buf.print(val);
+                buf.writeByte('E');
+            }
+            else
+            {
+                ti.error("Internal Compiler Error: C++ `%s` template value parameter is not supported", tv.valType.toChars());
+                fatal();
+            }
+        }
+        else if (tp.isTemplateAliasParameter())
+        {
+            Dsymbol d = isDsymbol(o);
+            Expression e = isExpression(o);
+            if (d && d.isFuncDeclaration())
+            {
+                bool is_nested = d.toParent() &&
+                    !d.toParent().isModule() &&
+                    (cast(TypeFunction)d.isFuncDeclaration().type).linkage == LINK.cpp;
+                if (is_nested)
+                    buf.writeByte('X');
+                buf.writeByte('L');
+                mangle_function(d.isFuncDeclaration());
+                buf.writeByte('E');
+                if (is_nested)
+                    buf.writeByte('E');
+            }
+            else if (e && e.op == TOK.variable && (cast(VarExp)e).var.isVarDeclaration())
+            {
+                VarDeclaration vd = (cast(VarExp)e).var.isVarDeclaration();
+                buf.writeByte('L');
+                mangle_variable(vd, true);
+                buf.writeByte('E');
+            }
+            else if (d && d.isTemplateDeclaration() && d.isTemplateDeclaration().onemember)
+            {
+                if (!substitute(d))
+                {
+                    cpp_mangle_name(d, false);
+                }
+            }
+            else
+            {
+                ti.error("Internal Compiler Error: C++ `%s` template alias parameter is not supported", o.toChars());
+                fatal();
+            }
+        }
+        else if (tp.isTemplateThisParameter())
+        {
+            ti.error("Internal Compiler Error: C++ `%s` template this parameter is not supported", o.toChars());
+            fatal();
+        }
+        else
+        {
+            assert(0);
+        }
+    }
+
+    /******************************
      * Write the mangled representation of the template arguments.
      * Params:
      *  ti = the template instance
+     * Returns:
+     *  true if any arguments were written
      */
-    void template_args(TemplateInstance ti)
+    bool template_args(TemplateInstance ti, int firstArg = 0)
     {
         /* <template-args> ::= I <template-arg>+ E
          */
-        if (!ti)                // could happen if std::basic_string is not a template
-            return;
+        if (!ti || ti.tiargs.dim <= firstArg)   // could happen if std::basic_string is not a template
+            return false;
         buf.writeByte('I');
-        foreach (i, o; *ti.tiargs)
+        foreach (i; firstArg .. ti.tiargs.dim)
         {
             TemplateDeclaration td = ti.tempdecl.isTemplateDeclaration();
             assert(td);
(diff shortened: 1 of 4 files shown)


jacob-carlborg:
What about a changelog entry? Did the other PR contain that?

src/dmd/cppmangle.d
```diff
@@ -602,23 +621,145 @@ private final class CppMangleVisitor : Visitor
...
                 else
-                    source_name(ti);
+                {
+                    int firstTemplateArg = 0;
+                    bool isConvertFunc = false;
+                    string symName;
+
+                    // test for special symbols
+                    sizediff_t whichOp = -1;
+                    foreach (i, op; [Id._cast, Id.assign, Id.eq, Id.index, Id.call, Id.opUnary, Id.opBinary, Id.opOpAssign])
```
Geod24: This allocates every time, doesn't it ?

src/dmd/cppmangle.d
```diff
@@ -651,7 +792,36 @@ private final class CppMangleVisitor : Visitor
                 }
                 else
                 {
-                    source_name(d);
+                    // test for special symbols
+                    sizediff_t whichOp = -1;
+                    foreach (i, op; [Id.assign, Id.eq, Id.index, Id.call])
```
Geod24: Ditto allocation

src/dmd/cppmangle.d
```diff
@@ -162,19 +162,112 @@ private final class CppMangleVisitor : Visitor
                 !getQualifier(s));      // at global level
     }
 
+    /******************************
+     * Write the mangled representation of a template argument.
+     * Params:
+     *  ti  = the template instance
+     *  arg = the template argument index
+     */
+    void template_arg(TemplateInstance ti, size_t arg = 0)
```
jacob-carlborg: Try to avoid default arguments, then it will make it easier to get the PR passed Walter.

src/dmd/cppmangle.d
TurkeyMan: I thought that was fixed years ago?

TurkeyMan:
> What about a changelog entry? Did the other PR contain that?

I'll add a big changelog entry aggregating all the extern(C++) stuff I manage to get in when the next build rolls around.

src/dmd/cppmangle.d
stefan-koch-sociomantic: nope. you can fix it though :)
() {
    enum a = [Id._cast, Id.assign, Id.eq, Id.index, Id.call, Id.opUnary, Id.opBinary, Id.opOpAssign];

TurkeyMan: That's terrible.

TurkeyMan: just to be sure... `immutable` will subvert TLS right? Why not `__gshared immutable array...`?

jacob-carlborg: > immutable will subvert TLS right?

Yes, it will be shared.

> Why not __gshared immutable array...

Yes, we're mostly using `__gshared` in DMD.

> () {
    static immutable array = [Id._cast, Id.assign, Id.eq, Id.index, Id.call, Id.opUnary, Id.opBinary, Id.opOpAssign];
    return array;
} ();

I don't see a need to complicate it like this, just declare it like this before the foreach loop:

```d
__gshared immutable ops = [Id._cast, Id.assign, Id.eq, Id.index, Id.call, Id.opUnary, Id.opBinary, Id.opOpAssign];
foreach (i, op; ops)
```

TurkeyMan: Why aren't array literals in the constant data block? I really recall this being fixed forget back?

klickverbot: Unless the array type is const/immutable, the actual runtime slice can't point to the data segment because it could be written to.

However, I believe both DMD and LDC will actually allocate an rvalue literal on the stack if possible, rather than putting it on the GC heap.

TurkeyMan: Oh yeah, I remember why I wrote it this way now... because these `Id.[token]`'s are not static values. They are initialised at runtime, so this:
```d
__gshared immutable ops = [Id._cast, Id.assign, Id.eq, Id.index, Id.call, Id.opUnary, Id.opBinary, Id.opOpAssign];
foreach (i, op; ops)
```
Doesn't compile, because can't evaluate at compile-time.

@klickverbot says those arrays will be on stack, which feels acceptable.

WalterBright: There's no reason to guess where they go. Just compile a snippet and look at the assembler. It's faster than asking, and gives a definitive answer.

test/compilable/cppmangle.d
```diff
@@ -686,6 +717,6 @@ version (Win64)
     static assert(TestOperators.opCast!int.mangleof      == "??BTestOperators@@QEAAHXZ");
     static assert(TestOperators.opAssign.mangleof        == "??4TestOperators@@QEAAHH@Z");
     static assert(TestOperators.opEquals.mangleof        == "??8TestOperators@@QEAA_NH@Z");
-    static assert(TestOperators.opIndex.mangleof         == "??ATestOperators@@QEAAH_K@Z");
+    static assert(TestOperators.opIndex.mangleof         == "??ATestOperators@@QEAAHH@Z");
```
WalterBright: How was this passing before? This PR doesn't fix it.

TurkeyMan: I'm not sure what you mean? All the tests pass...

Before it was `size_t`, I changed it to `int` (look above) because 32/64 mangles `size_t` differently so the posix tests which aren't split on 32/64bit needed a stable type to test.

src/dmd/cppmangle.d
TurkeyMan: It's on the stack.

test/compilable/cppmangle.d
WalterBright: ok

src/dmd/cppmangle.d
WalterBright: Being on the stack makes what should be an immutable array into a performance problem, because it has to be loaded on the stack. You can make it a __gshared array and initialize it as a singleton.

klickverbot: Alternatively, just turn the code below into an if chain instead of a switch?

TurkeyMan:
I put a function that makes the array once.

src/dmd/cppmangle.d
```diff
@@ -45,6 +45,21 @@ import dmd.visitor;
 
 extern (C++):
 
+// helper to check if an identifier is a C++ operator
+enum CppOperator { Cast, Assign, Eq, Index, Call, Unary, Binary, OpAssign, Unknown }
```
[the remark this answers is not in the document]
TurkeyMan: The enum values have to match the array indices, and I think it's awkward to populate the array with a dummy element, and also awkward to perform arithmetic when associating the array element to the enum. I was going to make `Unknown = -1` explicit, but that affects the optimisers ability to transform the switch statement into a jump table.

It's best like this.

src/dmd/cppmangle.d
```diff
@@ -45,6 +45,21 @@ import dmd.visitor;
...
+package CppOperator isCppOperator(Identifier id)
+{
+    __gshared const(Identifier)[] operators = null;
+    if (!operators)
+        operators = [Id._cast, Id.assign, Id.eq, Id.index, Id.call, Id.opUnary, Id.opBinary, Id.opOpAssign].dup;
```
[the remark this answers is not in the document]
TurkeyMan: Because array literals are stack allocated.

src/dmd/cppmangle.d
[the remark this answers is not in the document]
TurkeyMan: This is exactly the only place this enum will ever be used, and it's initialised by the function and returned in valid state. Nobody is going to default-initialise an instance of this.

src/dmd/cppmangle.d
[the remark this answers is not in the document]
Geod24: No they aren't, unless you assign them to a static array.
```D
void foo () @nogc @safe
{
    int[] arr = [1, 2, 3, 4];
}
```
> test.d(3): Error: array literal in @nogc function lol.foo may cause a GC allocation

src/dmd/cppmangle.d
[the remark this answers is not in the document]
Geod24: Didn't consider the index relationship. Makes sense then.

src/dmd/cppmangle.d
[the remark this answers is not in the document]
TurkeyMan: Okay, I wasted 64 bytes of ram ;)

Geod24:
> now we wait another few hours >_<

As soon as it is approved (with `auto-merge`) you'll have priority on the CI anyway

TurkeyMan:
This one's not so important to me, since we don't build Linux anyway ;)
`auto-merge` is on the dtor one, and been there half of today... is there a reason that's stalled?

Geod24:
Looks like there's an issue with std types:
```D
import core.stdcpp.exception;

extern(C++, Namespace):

struct Bar {}

struct Foo
{
    ref Foo opAssign(ref const int) { return this; }
    ref Foo opAssign(ref const exception) { return this; }
    ref Foo opAssign(ref const Bar) { return this; }
}
```

```
../dmd/generated/linux/release/64/dmd -c lol.d && nm lol.o | grep Namespace
0000000000000000 V _D28TypeInfo_S3lol9Namespace3Bar6__initZ
0000000000000000 V _D28TypeInfo_S3lol9Namespace3Foo6__initZ
0000000000000000 B _D3lol9Namespace3Bar6__initZ
0000000000000001 B _D3lol9Namespace3Foo6__initZ
0000000000000000 W _ZN9Namespace3FooaSERKi
0000000000000000 W _ZN9Namespace3FooaSERKNS_3BarE
0000000000000000 W _ZN9Namespace3FooaSERKPKN3std9exceptionE <= No NS substitution for `std`
```

Compared to C++ mangling:
```
#include <exception>

namespace Namespace {
struct Foo
{
    Foo& operator=(const std::exception& s)
    {
        return *this;
    };
};
};
```

```
g++ test.cpp -c && nm test.o | grep Namespace
0000000000000000 W _ZN9Namespace3FooaSERKSt9exception
```

TurkeyMan:
Why do the unit tests all pass?

TurkeyMan:
Are you saying this patch broke it? This patch has nothing to do with namespace mangling...
If I broke something, it's because there's no unittests for the thing ;)

Geod24:
Oh damn it, I thought this was working, but it's only working for mangling the declarations of `std` types, not for using them as parameters...  Time to open a new issue!

Geod24 (approved):

wilzbach:
> auto-merge doesn't work anymore ;)

Well that's not dlang-bot's fault. It waits until all CIs are green. In this case it looks like SemaphoreCi didn't report a success back yet.
Sometimes closing & reopening can help (and looks like it did help here) :/

TurkeyMan:
Last night the dtor one was all green, but it still didn't do the thing. I don't know what's stalled this one. It's been stalled for almost 24 hours.

TurkeyMan:
It was green last night... but then I removed `.dup` and it never finished :(

Geod24:
I think the situation with the `dtor` PR was a bit different. Let's wait&see. If it goes green and doesn't get merged, I'll do it manually like the other one.

wilzbach:
Yeah I don't know what's up with the Semaphore guys, but as far as I can observe every ~30-50th build request to them triggers the build at their site, but they never seem to send a reply back GitHub.
Anyhow, looks like it did work now:

image

TurkeyMan:
Aaaaand `win-farm-1`, `win-farm-2`, and `bellevue` are on it ;) ... those guys all take >50 minutes each!

wilzbach:
BTW I'm sorry, but it might be that Razvan's PR gets merged first and thus the auto-tester build will be restarted:

image

Though it should be merged "soonish (TM)" ;-)
BTW I don't know this for sure, but I think everyone here has learned to do sth. else or work on a new/other PR while a PR is slowly working its way through the auto-tester priority queue ...

TurkeyMan:
It's fine anyway, I won't even use this PR ;)
Rainer's PR is the only one left that I really want merged before I can get people at work looking at it.

PR #8336, opened 2018-06-06
Support __traits(getLinkage) for structs

TurkeyMan:
This is required for the extern(C++) destructors patch, but I felt it was isolated enough to stand as a patch on its own.

### DIFF
--- a/src/dmd/aggregate.d
+++ b/src/dmd/aggregate.d
@@ -50,6 +50,22 @@ enum Baseok : int
     semanticdone,     // all base classes semantic done
 }
 
+/**
+ * The ClassKind enum is used in AggregateDeclaration AST nodes to
+ * specify the linkage type of the struct/class/interface or if it
+ * is an anonymous class. If the class is anonymous it is also
+ * considered to be a D class.
+ */
+enum ClassKind : int
+{
+    /// the aggregate is a d(efault) class
+    d,
+    /// the aggregate is a C++ struct/class/interface
+    cpp,
+    /// the aggregate is an Objective-C class/interface
+    objc,
+}
+
 /***********************************************************
  */
 extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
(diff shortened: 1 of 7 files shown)


TurkeyMan:
semaphoreci thing is here too... :/

src/dmd/aggregate.d
```diff
@@ -50,6 +50,22 @@ enum Baseok : int
     semanticdone,     // all base classes semantic done
 }
 
+/**
+ * The ClassKind enum is used in AggregateDeclaration AST nodes to
+ * specify the linkage type of the struct/class/interface or if it
+ * is an anonymous class. If the class is anonymous it is also
+ * considered to be a D class.
```
rainers: As it wasn't obvious why structs need this information, adding a comment here would be nice. Otherwise LGTM.

TurkeyMan: I totally rewrote that comment to include struct... What should it say?
My feeling is, any reason that class needs it is logically transferred to struct too.

TurkeyMan: I didn't rename the enum though, that would have been a very invasive change...
I started renaming it, and then realised how far it spread.

TurkeyMan:
Right... Like, I feel like it's implied that everything about it should be like C++ :P

jacob-carlborg:
Needs to updated the spec [1].

[1] https://dlang.org/spec/traits.html#getLinkage

WalterBright (approved):

src/dmd/aggregate.d
RazvanN7: >I started renaming it, and then realised how far it spread.

Honestly, I think that it should be renamed even if it's an invasive change. But it can be done in a subsequent PR

RazvanN7 (approved):

RazvanN7:
Spec PR : https://github.com/dlang/dlang.org/pull/2383

PR #8338, opened 2018-06-06
Fix Issue 18951 - package static method masked by public static method in class

FeepingCreature:
When selecting the "most-visible" overload method to check access,
treat package methods as public if they're visible; otherwise, private.
This avoids package methods that are perfectly visible being passed over
in favor of public (numerically more visible) overloads appearing later
in the class.

### DIFF
--- a/src/dmd/access.d
+++ b/src/dmd/access.d
@@ -543,7 +543,7 @@ extern (C++) bool checkSymbolAccess(Scope *sc, Dsymbol s)
  * but doesn't recurse nor resolve aliases because protection/visibility is an
  * attribute of the alias not the aliasee.
  */
-public Dsymbol mostVisibleOverload(Dsymbol s)
+public Dsymbol mostVisibleOverload(Dsymbol s, Module mod = null)
 {
     if (!s.isOverloadable())
         return s;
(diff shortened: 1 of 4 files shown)


src/dmd/access.d
```diff
@@ -543,7 +543,7 @@ extern (C++) bool checkSymbolAccess(Scope *sc, Dsymbol s)
  * but doesn't recurse nor resolve aliases because protection/visibility is an
  * attribute of the alias not the aliasee.
  */
-public Dsymbol mostVisibleOverload(Dsymbol s)
+public Dsymbol mostVisibleOverload(Dsymbol s, Module mod = null)
```
RazvanN7: You can get rid of the optional parameter and obtain the module form s by calling `s.getModule` [1]

[1] https://github.com/dlang/dmd/blob/master/src/dmd/dsymbol.d#L327

FeepingCreature: No? The module is the module the symbol is supposed to be visible *from*.

src/dmd/access.d
```diff
@@ -612,8 +612,24 @@ public Dsymbol mostVisibleOverload(Dsymbol s)
...
     return mostVisible;
 }
+
+/**
+ * Return the "effective" protection attribute of a symbol seen from a module.
+ * The effective protection attribute is the same as the regular protection attribute,
+ * except package() is "private" if the module is outside the package;
+ * otherwise, "public".
+ */
+Prot effectiveProtAt(Dsymbol d, Module mod)
```
JinShil: Please spell this out as "effectiveProtection".  I had to read the comment to understand what "Prot" meant.

JinShil: I also recommend making this a nested function in the `mostVisibleOverload` function, or at a minimum, make it `private`.

FeepingCreature: "protection from" sounds like a D&D spell. :)

FeepingCreature: Done, and renamed to protectionIn

src/dmd/access.d
```diff
@@ -612,7 +612,23 @@ public Dsymbol mostVisibleOverload(Dsymbol s)
...
             break;
 
-        if (next && mostVisible.prot().isMoreRestrictiveThan(next.prot()))
+        /**
+        * Return the "effective" protection attribute of a symbol when accessed in a module.
+        * The effective protection attribute is the same as the regular protection attribute,
+        * except package() is "private" if the module is outside the package;
+        * otherwise, "public".
+        */
+        Prot protectionInModule(Dsymbol d, Module mod = null)
```
jacob-carlborg: Looks like this function can be `static`.

JinShil: I still prefer "From" over "In".

FeepingCreature: fixed?

JinShil (approved):

RazvanN7 (approved):

FeepingCreature:
ping? Fail is almost certainly unrelated.

FeepingCreature:
Woo!

PR #8341, opened 2018-06-08
Fix issue 14178 - extern(C++) doesn't mangle 'std' correctly on posix systems

TurkeyMan:


jacob-carlborg (approved):

Geod24 (approved):
Thanks. Updated the title to reference https://issues.dlang.org/show_bug.cgi?id=14178

PR #8342, opened 2018-06-08
Fix Issue 18958 - extern(C++) wchar, dchar mangling not correct

TurkeyMan:


Geod24:
Does this mean we drop compatibility for C++ < 11 ?

Geod24:
Note there's https://issues.dlang.org/show_bug.cgi?id=14956

Geod24:
[Don't think that's possible yet](https://github.com/dlang/dmd/blob/d1ee5892181fe0c9f69e79c764d7224225ba5362/test/tools/d_do_test.d#L437-L474)

TurkeyMan:
Well it's about time :)

TurkeyMan:
Feel like negotiating/working on a solution to that problem?
I need to go to bed! :P

src/dmd/cppmanglewin.d
```diff
@@ -202,27 +202,26 @@ public:
...
+            buf.writestring("_N");
+            break;
+        case Tchar:
+            buf.writeByte('D');
+            break;
         case Twchar:
             if (flags & IS_DMC)
                 buf.writestring("_Y"); // DigitalMars wchar_t
             else
-                buf.writestring("_W"); // Visual C++ wchar_t
```
jacob-carlborg: This a breaking change.

jacob-carlborg:
> Does anyone know how we build the C++ file with -std=c++11?

As far a I know, DMC does not support C++11.

wilzbach:
You could create a new test file, add the c11 flag and ignore that test for win32?

src/dmd/cppmanglewin.d
TurkeyMan: It's a *fixing* change. It's already broken.

TurkeyMan:
DMC can keep its mangling hacks then, and typedef `char16_t`/`char32_t` to some int types?

TurkeyMan:
> and ignore that test for win32?
What's wrong with win32?

How do I apply the c++11 flag?

src/dmd/cppmanglewin.d
TurkeyMan: Nobody can depend on the extern(C++) string stuff, because it's totally broken! We need to finally straighten this out.

`wchar_t` needs a typedef, like `c_long`.

wilzbach:
> How do I apply the c++11 flag?

Here's one example:

```diff
diff --git a/test/compilable/extra-files/test18958.cpp b/test/compilable/extra-files/test18958.cpp
new file mode 100644
index 000000000..45e1ce1e9
--- /dev/null
+++ b/test/compilable/extra-files/test18958.cpp
@@ -0,0 +1,5 @@
+#include <tuple>
+int hello() {
+    std::tuple<int, double> t(1, 2.0);
+    return std::get<0>(t);
+}
diff --git a/test/compilable/extra-files/test18958.d b/test/compilable/extra-files/test18958.d
new file mode 100644
index 000000000..7da3a94ec
--- /dev/null
+++ b/test/compilable/extra-files/test18958.d
@@ -0,0 +1,5 @@
+extern(C++) int hello();
+
+void main(){
+    assert(hello() == 1);
+}
diff --git a/test/compilable/test18958.sh b/test/compilable/test18958.sh
new file mode 100644
index 000000000..bfaf3cabf
--- /dev/null
+++ b/test/compilable/test18958.sh
@@ -0,0 +1,8 @@
+// DISABLED: win32
+#!/usr/bin/env bash
+
+"${CC}" -std=c++11 -o ${OUTPUT_BASE}cpp${OBJ} -c ${EXTRA_FILES}/${TEST_NAME}.cpp
+
+$DMD -c -m${MODEL} -of${OUTPUT_BASE}${EXE} -I${EXTRA_FILES} ${EXTRA_FILES}/${TEST_NAME}.d ${OUTPUT_BASE}cpp${OBJ}
+
+rm -f ${OUTPUT_BASE}${EXE}
diff --git a/test/tools/exported_vars.sh b/test/tools/exported_vars.sh
index 4f097e92d..5ca796cc0 100644
--- a/test/tools/exported_vars.sh
+++ b/test/tools/exported_vars.sh
@@ -12,3 +12,5 @@ if [ "$OS" == "win32" ] || [ "$OS" == "win64" ]; then
 else
     export LIBEXT=.a
 fi
+
+export CC="${CC:-c++}" # C++ compiler to use
```

An alternative to this would be to add sth. like `CFLAGS` to [`d_do_test`](https://github.com/dlang/dmd/blob/master/test/tools/d_do_test.d).

> What's wrong with win32?

I thought DigitalMars C++ doesn't support C++11?

src/dmd/cppmangle.d
```diff
@@ -1047,8 +1047,8 @@ public:
             case Tfloat80:               c = 'e';       break;
             case Tbool:                  c = 'b';       break;
             case Tchar:                  c = 'c';       break;
-            case Twchar:                 c = 't';       break;  // unsigned short (perhaps use 'Ds' ?
-            case Tdchar:                 c = 'w';       break;  // wchar_t (UTF-32) (perhaps use 'Di' ?
+            case Twchar:        p = 'D'; c = 's';       break;  // since C++11
+            case Tdchar:        p = 'D'; c = 'i';       break;  // since C++11
```
ibuclaw: The TypeBasic visitor guarded by Target.cppTypeMangle, right? Maybe is about time that we allow the user to control the c++ abi version than breaking backwards compatibility.

TurkeyMan:
...no. We already clearly demonstrate that we can't maintain even a single abi revision.
We should just support the 'current' one.

ibuclaw:
> We should just support the 'current' one.

It starts off with no one agreeing what 'current' means. I don't think cppmangle even does C++11 Std strings, even though I raised that bug report probably 2 years ago now. (this is why we can't maintain a single version :-)

At least on the gcc side, abi versions are small additive changes. I can't say that maintaining it would be beyond us.

Also, at least _I_ will have a dependency on C++98 for the foreseeable future, so will end up adding an -stdc++=xx style option anyway even if the frontend doesn't. :-(

TurkeyMan:
It would be really great to move this forward.

Geod24:
@TurkeyMan : You can use [`// CXXFLAGS: -std=c++11` ](https://github.com/dlang/dmd/blob/08e4d8287081a68adf53c405bbd07dea7dd27f75/test/runnable/cpp_stdlib.d#L3)  to compile your `EXTRA_CPP_SOURCES` with this flag.

ibuclaw:
> You can use // CXXFLAGS: -std=c++11 to compile your EXTRA_CPP_SOURCES with this flag.

This just hides that C++98 support is now broken. I won't treat that as anything else other than a regression.

TurkeyMan:
I agree in principle, but we're in a broken position either way.
I think there are 2 options here; we officially support just one C++ version, and in that case, people have generally agreed it should be the 'current' one, or, we add the concept of C++ version to D, and I haven't heard any consensus, or even support for that.

I think The right thing to do would be, in the meantime while we only do support one version, declare that we support the 'current' version rather than a random 'whichever one seems to work', which feels like the only reasonable option. And then in the future if we decide to support C++ versions selectively, then we can include multi-version tests and commit to backwards compatibility then... No?

Geod24:
@TurkeyMan : I haven't looked into it too much yet, but was there a lot of mangling changes between c++11 and the later versions ?

ibuclaw:
>  but was there a lot of mangling changes between c++11 and the later versions ?

It isn't quite so clear cut as a lot of mangling changes between C++ standards.  In g++, ABI version revolves around compiler release (https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Dialect-Options.html).

In this case, `wchar_t` is a basic C++ type, and `char16_t` and `char32_t` are new C++11 extensions.  Both `wchar_t` and `char32_t` are 

So you have:
```
char16_t foo();  // wchar -> 'Ds'
char32_t foo();  // dchar -> 'Di'
wchar_t foo();   // dchar -> 'w' (However... wchar -> 'w' if -fshort-wchar)
```
D's `dchar` can't be both `char32_t` and `wchar_t` at the same time!  `wchar_t` should be a special `enum`, same as `__c_long` and friends.

ibuclaw:
I don't see any problem defaulting to C++11 and later types, as they are more predictable than the earlier types.  However you shouldn't prevent people from being able to link to a C++98 application, whereby you fall back on the closest equivalent as what we currently do.

Having an `enum cxxdialect { cxx98, cxx11, cxx14, cxx17, ... }` is fine by me, dmd can hard code it to whatever it wants, I can just add a command-line option on my side with a default that matches the gcc compiler i'm released with.

TurkeyMan:
> Having an `enum cxxdialect { cxx98, cxx11, cxx14, cxx17, ... }` is fine by me [...]

It's fine by me too, but I don't see that as necessary here, all 3 cases are legitimate bugs that should be corrected.
`wchar_t` is a distinct type with distinct mangling; fixing this doesn't depend on a C++ dialect.
`wchar`/`dchar` mangling are defined by C++11 when they were introduced. Mangling them in any other way is also just plain wrong.
The only thing you might get from a `cxxdialect`, is that you should "error: char16_t/char32_t mangling not supported prior C++11" when dialect is set as such, but that doesn't offer any advantage.

The current situation is just incorrect no matter how you look at it. It needs to be correct-first, and any magical additional cases are strictly opt-in.

thewilsonator:
@TurkeyMan you'll need to update the tests in https://github.com/dlang/dmd/blob/master/test/runnable/extra-files/cppb.cpp#L410-L435

as well

TurkeyMan:
Certainly, I'll fix this as soon as yours merges.

test/runnable/cpp_abi_tests.d
```diff
@@ -94,7 +107,7 @@ template IsBoolean(T)
 
 template IsSomeChar(T)
 {
-    enum IsSomeChar = is(T==char) || is(T==dchar);
+    enum IsSomeChar = is(T==char) || is(T==wchar) || is(T==dchar) || is(T==wchar_t);
```
thewilsonator: use `is(T : wchar)` because an the enum `__c_wchar_t` is not the same (`==`). Ditto for `dchar`.

thewilsonator: Oh wow I can't read. Ignore that.

TurkeyMan:
Oh no... the tests are building with a C compiler!
https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3406683&isPull=true

thewilsonator:
Are you sure its not an old C++ compiler?

thewilsonator:
Oh, you did add `// CXXFLAGS: -std=c++11` to the top of that file. Hmm.

TurkeyMan:
What are you looking at?
```
runnable/extra-files/cpp_abi_tests.cpp:14:1: error: unknown type name 'char16_t'
char16_t           passthrough(char16_t            value)     { return value; }
^
failed to execute 'cc -m64 -c runnable/extra-files/cpp_abi_tests.cpp -o generated/runnable/cpp_abi_tests.cpp.o'
```
Trying to call `cc`...

thewilsonator:
https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3406692&isPull=true

TurkeyMan:
Yeah that's fine, I fixed that one.

TurkeyMan:
I have no idea what went wrong with semaphore either :/

thewilsonator:
See the top of https://github.com/dlang/dmd/blob/master/test/runnable/extra-files/cppb.cpp

TurkeyMan:
I saw that. But there is also this:
```
Fortunately, it is easily possible to disable the new feature
by defining _GLIBCXX_USE_CXX11_ABI as 0 before including any standard
headers.
```
```c++
#define _GLIBCXX_USE_CXX11_ABI 0
```

TurkeyMan:
Anyway, the freebas C compiler should be trivial; I just can't find where the makefile that executes the build is...

thewilsonator:
Indeed maybe its time to split the c++11 stuff into its own test files.

test/runnable/extra-files/cpp_abi_tests.cpp
```diff
@@ -11,6 +11,8 @@ bool               passthrough(bool                value)     { return value; }
 signed char        passthrough(signed char         value)     { return value; }
 unsigned char      passthrough(unsigned char       value)     { return value; }
 char               passthrough(char                value)     { return value; }
+char16_t           passthrough(char16_t            value)     { return value; }
```
thewilsonator: I think you're going to have to wrap these in `#if ! __DMC__`, win32 fails with duplicate definition. Ditto for below.

thewilsonator:
> Anyway, the freebsd C compiler should be trivial; I just can't find where the makefile that executes the build is...

Per https://www.freebsd.org/cgi/man.cgi?query=cc&sektion=1 using `cc` is fine, per the autotester `failed to execute 'cc -m64 -c runnable/extra-files/cpp_abi_tests.cpp -o generated/runnable/cpp_abi_tests.cpp.o'
 I don't see any c++11 flags there.

TurkeyMan:
`cc` is not a C++ compiler... I had the same issue in the std::array<> PR.
The flags are specified in the `// CXXFLAGS: -std=c++11` bits.

`__DMC__` is indeed an ongoing hassle >_<

TurkeyMan:
Anyway, I'm done here. G'nite.

thewilsonator:
> cc is not a C++ compiler...

It is on *BSDs, its all just an alias to clang.

thewilsonator:
Appveyor fails with 
```
unresolved external symbol "void __cdecl fuzz3_cppvararg(char16_t,char16_t,bool)" referenced in function _D4cppa5fuzz3FZv
unresolved external symbol "void __cdecl fuzz3_checkValues(wchar_t,wchar_t,bool)" referenced in function "void __cdecl fuzz3_cppvararg(wchar_t,wchar_t,bool)"
```

TurkeyMan:
Appveyor looks fine to me...
That must be old. I changed those function signatures really early on.

thewilsonator:
Go to sleep!

thewilsonator:
Oh no...  they're adding a `char8_t` to C++20.

TurkeyMan:
Well... at least that will make C++ consistent with D! :)

thewilsonator:
This also needs a changelog entry.

TurkeyMan:
Sure, but it's still nowhere near working ;)

WalterBright:
Regarding the mangling of `wchar_t` - it's already specially handled in D. The trouble is `core.stdc.stddef` was never updated to actually use it. The dmd internal type is `__c_win32_t`.

WalterBright:
I recommend the following:

1. add a PR to fix `core.stdc.stddef` to use the existing special dmd enum `__c_wchar_t`
2. add a `global.params.cpp98` boolean to use C++98 mangling (which DMC++ uses)
3. add a `-transition=cpp98` flag to support older C++98 compilers (DMC++ should default to this)
4. rely on `global.params.cpp98` to use the older C++ name mangling method
5. use `global.params.cpp98` to set a `Cpp98` predefined version

WalterBright:
> a -std-c++ flag is better because there are multiple version of C++

I was thinking of the future, where old versions of C++ will be gone. We should make the modern C++ the default, and then have switches for older implementations. This should only be done as required, and currently C++98 should solve our problems. We don't need one for every version. The intent of the `-transition` set of switches is to encourage people to modernize their code.

WalterBright:
See https://github.com/dlang/dmd/pull/9029 for addition of compiler switch.

thewilsonator:
> I was thinking of the future, where old versions of C++ will be gone. We should make the modern C++ the default, and then have switches for older implementations. 

Yes

> We don't need one for every version. 

Sorry for the confusion, I meant `-std-c++=xx`

> The intent of the -transition set of switches is to encourage people to modernize their code.

This is not something I would intuitively expect to find under `-transition`, also no `-transition` flags are mutually exclusive

thewilsonator:
(just giving this a bit of testing priority)

thewilsonator (approved):

test/runnable/cpp_abi_tests.d
```diff
@@ -1,4 +1,11 @@
 // EXTRA_CPP_SOURCES: cpp_abi_tests.cpp
+// CXXFLAGS: -std=c++11
```
thewilsonator: This looks like it will need to be handled by a makefile. The Windows boxen don't like it.

thewilsonator:
@TurkeyMan The linux problem seems to have been solved. All that needs doing is fixing the flags passed to `cl` and versioning out `dmc`.

TurkeyMan:
Good :)

test/runnable/cpp_abi_tests.d
rainers: You can make the option depend on the OS variable:
```
// CXXFLAGS(linux freebsd osx netbsd dragonflybsd): -std=c++11
```

WalterBright (dismissed):
Still needs changelog.

jacob-carlborg:
> Still needs changelog.

@WalterBright the PR is linked with issue https://issues.dlang.org/show_bug.cgi?id=18958. It will automatically generate a changelog entry. I don't think we need an explicit changelog entry describing the the fix.

test/runnable/cpp_abi_tests.d
thewilsonator: Yeah but windows needs to be tested with c++11+ as well.

lesderid: @thewilsonator MSVC doesn't have a C++11 switch, but it defaults to the latest fully-supported standard

thewilsonator: Ah, good to know.

thewilsonator:
(again)

thewilsonator:
Wonderful VS2013 doesn't support char16_t/char32_t

test/runnable/cpp_abi_tests.d
rainers: cl from VS2019 shows this in the command line help:
```
/std:<c++14|c++17|c++latest> C++ standard version
    c++14 – ISO/IEC 14882:2014 (default)
    c++17 – ISO/IEC 14882:2017
    c++latest – latest draft standard (feature set subject to change)```

thewilsonator: As long as the default is >c++11 it doesn't really matter. What is very annoying is that VS2013 doesn't support char16_t/char32_t, and the autotester machines seem to use that.

TurkeyMan:
You're loving it!
This has been my eternal punishment for trying to do any of the `extern(C++)` stuff!

thewilsonator:
I F$%#^# HATE THIS CI SYSTEM!

wilzbach:
Well, we all would love to replace the Auto-Tester, but so far the problem was always money. Andrei recently told me that the DLF these days would pay for their own CI machines, but unfortunately I don't really have time to the migration these days.
If someone is interested in this, ping me. I'm happy to help. The general idea is pretty simple though: we rent a few cheap Windows/OSX boxes, install tooling and the Buildkite agent and then use them to fully cover all platforms, s.t. we can retire the Auto-Tester.

thewilsonator:
To correct my self slightly, I hate the windows machines because I hate DMC and I hate Microsoft for being inconsistent with their C++11 support on VS2013. We axed OSX32, that made things much nicer. 

Could we just move to Azure for the Windows stuff?

wilzbach:
> Could we just move to Azure for the Windows stuff?

Yes, but we would need to configure it for druntime and Phobos (it's just running the same script though with just the repository cloning being different).

PR #8343, opened 2018-06-08
Export CC to the test shell environment

wilzbach:
Useful for e.g. https://github.com/dlang/dmd/pull/8342

This performs what `d_do_test` already does today:

https://github.com/dlang/dmd/blob/cce909b19a1e6b03a182d2a1d488841e2cc64b53/test/tools/d_do_test.d#L588-L596

(though the win64 case isn't really used anymore)

### DIFF
--- a/test/compilable/issue15574.sh
+++ b/test/compilable/issue15574.sh
@@ -1,8 +1,6 @@
 #! /usr/bin/env bash
 
-
 if [[ $OS = *"win"* ]]; then exit 0; fi
-
 
 if [ $LIBEXT = ".a" ]; then
     LIB_PREFIX=lib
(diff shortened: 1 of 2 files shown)


test/compilable/issue15574.sh
```diff
@@ -46,9 +43,7 @@ int main() {
 }
 EOF
 
-if [ -z ${CC+x} ]; then CC=cc; fi
-
-$CC -m${MODEL} -c -o ${C_FILE}${OBJ} $C_FILE
+cc -m${MODEL} -c -o ${C_FILE}${OBJ} $C_FILE
```
wilzbach: So we managed to reach the state where CC refers to the C++ compiler:

https://github.com/dlang/dmd/blob/cce909b19a1e6b03a182d2a1d488841e2cc64b53/test/tools/d_do_test.d#L588-L596

https://github.com/dlang/dmd/blob/cce909b19a1e6b03a182d2a1d488841e2cc64b53/test/README.md#L188

But in this case CC is used as people normally use it: the C compiler.

In this case it makes a difference whether a C or C++ compiler is used.
I would suggest we introduce a new environment variable (e.g. `CPP`) and let CC point to the system's C compiler.
FWIW in the build make scripts the C++ compiler is at least called `HOST_CXX`.

RazvanN7 (approved):

PR #8344, opened 2018-06-08
Fix Issue 15361 - Incomprehensible error message: function declaration without return type

RazvanN7:


wilzbach (approved):
Nice!

RazvanN7:
The AppVeyor fail seems unrelated :-1:

JinShil:
I restarted AppVeyor.

rainers:
> The AppVeyor fail seems unrelated 👎

LDC forbids assigning to a Token (maybe because it contains a union?).

That's unlikely a good idea to begin with, because it might break the chain of tokens (it has a `next` member). I think you should avoid modifying the tokens, use pointers instead.

RazvanN7:
@rainers is this better?

UplinkCoder:
This seems like it's not too well thought out.

RazvanN7:
@rainers 

>Sorry, I don't think so. You might fool LDC into compling it, but still change the current token of the lexer. I think you should call skipParens instead to keep the lexer state sane.

I am obviously missing something here, but if I don't change the current token to jump over what is inside the `is` expression we will get redundant error information.  

>BTW: I find it a bit ugly to check for some special situations and error out before actually trying to parse the type. (See the C cast debacle wih (x)++).

I agree, but I have found no other way. The problem is that the isExpression is expecting a type so after the parser has encountered `is(` it tries to parse the following tokens as a type and from inside the `parseType` function it is impossible to know whether you're in an is expression or not. So I preferred to check for this case in advance while I still know that I am parsing an isExpression, rather then to add hacks in the parseType function which is used from many other places. If you have a better solution or I am missing something, please share your thoughts. 

>Did you close the wrong bugzilla issue?

Yes, sorry about that. I reopened 15361 and closed 15360

RazvanN7:
@UplinkCoder 

>This seems like it's not too well thought out.

Thanks you! This is really helpful.

src/dmd/parse.d
```diff
@@ -7664,6 +7664,15 @@ final class Parser(AST) : Lexer
                 if (token.value == TOK.leftParentheses)
                 {
                     nextToken();
+                    if (token.value == TOK.identifier && peekNext() == TOK.leftParentheses)
+                    {
+                        error(loc, "unexpected `(` after `%s`, inside `is` expression. Try nesting the contents of `is` with a `typeof` expression", token.toChars());
+                        nextToken();
+                        Token* tempTok = peekPastParen(&token);
+                        memcpy(&token, tempTok, Token.sizeof);
```
WalterBright: What is the purpose of the memcpy()?

WalterBright:
> If you have a better solution

Lookahead. See the lookahead functions isDeclaration, isBasicType, isDeclarator, isParameters, isExpression.

WalterBright (dismissed):
Use lookahead parsing functions.

src/dmd/parse.d
RazvanN7: ldc forbids assigning to a token. before the mempcy the code looked like this:

```d
nextToken();
token = *peekPastParens(&token);
```

I did this in order to avoid further error messages. I don't know how to do it otherwise that to just call nextToken() until the current token matches the one that peekPastParens points to.

rainers:
This is not about lookahead, but discarding the is-expression argument, similar to moving forward to the `;` after an error.

Unfortunately, `skipParens` is not actually skipping the token, it also peeks ahead. I would try something like (untested)
```
Token* paren;
if (skipParens(&token, &paren))
    while(token.next != paren)
        nextToken();
nextToken();
```

rainers:
> If you have a better solution or I am missing something, please share your thoughts.

Not really. One option might be gagging the errors, and checking for possible causes after parsing failed, but that can remove valid messages, too, so it's also not great. Early emission of errors need to be scrutinized to not reject valid code, though.

src/dmd/parse.d
WalterBright: I meant what is the purpose of assigning to token? Lookahead is used elsewhere in the parser to decide what is being parsed.

RazvanN7:
@WalterBright The lookeahead functions are not so helpful in this case.

src/dmd/parse.d
RazvanN7: The purpose of assigning to a token is to skip the list of parameters that is passed to the function call inside the is expression

rainers:
One case that could be considered a diagnostic regression if changed:
```
alias Tint = int;
static if(is(Tint(*)())) {}

test.d(2): Error: instead of C-style syntax, use D-style Tint function()
```
i.e. C style functions are errors, but still recoginized.

andralex (approved):
This looks good to me modulo what seems to be a minor misunderstanding. The code correctly discards `is(abc(.....`, which is lexically incorrect, and skips over the input portion deemed invalid.

andralex:
>One case that could be considered a diagnostic regression if changed:
```
alias Tint = int;
static if(is(Tint(*)())) {}

test.d(2): Error: instead of C-style syntax, use D-style Tint function()
```
>i.e. C style functions are errors, but still recognized.

Interesting case, thanks! Spent a few minutes myself with the spec, couldn't find it. I think it can be safely ignored.

src/dmd/parse.d
```diff
@@ -7664,6 +7664,15 @@ final class Parser(AST) : Lexer
                 if (token.value == TOK.leftParentheses)
                 {
                     nextToken();
+                    if (token.value == TOK.identifier && peekNext() == TOK.leftParentheses)
+                    {
+                        error(loc, "unexpected `(` after `%s`, inside `is` expression. Try nesting the contents of `is` with a `typeof` expression", token.toChars());
```
JinShil: Should it be "Try *enclosing* the contents of \`is\` *in* a \`typeof\` expression"?

src/dmd/parse.d
wilzbach: > ldc forbids assigning to a token. before the mempcy the code looked like this:

Just tested this locally and LDC 1.10 compiles dmd just fine. Maybe you weren't using the latest LDC?

src/dmd/parse.d
```diff
@@ -7664,6 +7664,15 @@ final class Parser(AST) : Lexer
                 if (token.value == TOK.leftParentheses)
                 {
                     nextToken();
+                    if (token.value == TOK.identifier && peekNext() == TOK.leftParentheses)
+                    {
+                        error(loc, "unexpected `(` after `%s`, inside `is` expression. Try enclosing the contents of `is` with a `typeof` expression", token.toChars());
+                        nextToken();
+                        Token* tempTok = peekPastParen(&token);
+                        memcpy(&token, tempTok, Token.sizeof);
```
wilzbach: As mentioned, I don't think you need the `memcpy` here. Assignment should work.

RazvanN7:
ping. I think we can merge this.

rainers (changes requested):
I still think the memcpy is bad, but worse: this will reject valid code (for -m64):
```
static if(is(__vector(double[2]))) pragma(msg, "accepted");
```

RazvanN7:
>I still think the memcpy is bad, but worse: this will reject valid code (for -m64):
static if(is(__vector(double[2]))) pragma(msg, "accepted");

@rainers , I just checked and the result of the above code is the same with this patch as without the patch. Am I missing something?

rainers:
> @rainers , I just checked and the result of the above code is the same with this patch as without the patch.

You are right, I probably thought `__vector` is a special identifier, but it is a keyword.

`memcpy` is also used elsewhere to advance to a peeked token so should be ok, but it leaks the skipped tokens here. Probably not worth the trouble, though.

rainers (approved):

RazvanN7:
cc @thewilsonator

PR #8345, opened 2018-06-08
Fix Issue 15360 - typeof is not a template argument

RazvanN7:
Spec PR: https://github.com/dlang/dlang.org/pull/2387

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -2028,12 +2028,14 @@ final class Parser(AST) : Lexer
         case TOK.functionString:
         case TOK.prettyFunction:
         case TOK.this_:
-            {
-                // Template argument is an expression
-                AST.Expression ea = parsePrimaryExp();
-                tiargs.push(ea);
-                break;
-            }
+            // Template argument is an expression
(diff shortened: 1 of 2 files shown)


src/dmd/parse.d
```diff
@@ -2028,12 +2028,14 @@ final class Parser(AST) : Lexer
         case TOK.functionString:
         case TOK.prettyFunction:
         case TOK.this_:
-            {
-                // Template argument is an expression
-                AST.Expression ea = parsePrimaryExp();
-                tiargs.push(ea);
-                break;
-            }
```
RazvanN7: This code block here was not respecting the norm that the other code blocks introduced, so I fixed the indentation.

jacob-carlborg:
I think this bug should be closed as “invalid” or “won’t fix”.

RazvanN7:
>I think this bug should be closed as “invalid” or “won’t fix”.

That was my first reaction also, but I thought it would be best to make a PR and discuss it. Closing PR and issue

jacob-carlborg:
It can be discussed directly in the issue as well.

PR #8346, opened 2018-06-09
fix Issue 18963 - Relax restrictions on 'return' parameters when para…

WalterBright:
…meter is not a pointer

### DIFF
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -875,7 +875,7 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
 
             if (tf.isreturn && !tf.isref && !tf.next.hasPointers())
             {
-                mtype.error(loc, "function type `%s` has `return` but does not return any indirections", tf.toChars());
+                tf.isreturn = false;
             }
         }
 
@@ -964,9 +964,7 @@ private extern (C++) final class TypeSemanticVisitor : Visitor
                         }
                         else if (tf.next && !tf.next.hasPointers())
                         {
-                            mtype.error(loc, "parameter `%s` is `return` but function does not return any indirections",
-                                fparam.ident ? fparam.ident.toChars() : "");
-                            errors = true;
+                            fparam.storageClass &= ~STC.return_;   // https://issues.dlang.org/show_bug.cgi?id=18963
                         }
                     }
                 }
--- a/test/fail_compilation/test16228.d
+++ b/test/fail_compilation/test16228.d
@@ -1,10 +1,10 @@
 /* REQUIRED_ARGS: -dip25
    TEST_OUTPUT:
 ---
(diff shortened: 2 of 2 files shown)


WalterBright:
This is blocking https://github.com/dlang/phobos/pull/6560

WalterBright:
@jinshil the spec doesn't mention `return` parameters at all.

JinShil:
> the spec doesn't mention return parameters at all.

It mentions `return ref` parameters.  But the lack of documentation is all the more reason to document it now, and to through that documentation justify this PR.

schveiguy:
Just want to voice my support here. I don't have enough experience to review this, to know that it's correctly implemented. But speaking from experience with `inout` (I originally wanted `inout` to have similar errors like "can't have `inout` return without `inout` parameters" and it's horrible for generic code).

WalterBright:
> But the lack of documentation is all the more reason to document it now, and through that documentation justify this PR.

1. Dip1000 is not in the spec. It certainly needs to be done, but I'm pretty reluctant to put the entire Phobos implementation on hold for another year because of this. Making everything serial instead of parallel is not productive.

2. I can't get even simple PRs to the spec pulled. Last January, I resolved to go through and make the spec more accurate and more formal. But this all ground to a halt due to complete disinterest in anyone reviewing or pulling them.

3. This PR is a trivial and obvious improvement. The proposed behavior is the same as for `scope` which is already implemented.

4. This PR is not a dot i and cross t thing. It's *blocking* valuable and important work.

JinShil:
> Dip1000 is not in the spec. It certainly needs to be done, but I'm pretty reluctant to put the entire Phobos implementation on hold for another year because of this. Making everything serial instead of parallel is not productive.

Too often, features are added to D without sufficient documentation, and DIP1000 is the worst offender.  I'm not asking that you remedy that entire problem right now, I'd just like to see a section on Return Parameters with an brief explanation of it's purpose, an bullet point calling out this PR's exception, and perhaps an example.  I don't think that's too much to ask.

> I can't get even simple PRs to the spec pulled. Last January, I resolved to go through and make the spec more accurate and more formal. But this all ground to a halt due to complete disinterest in anyone reviewing or pulling them.

Indeed, and I'm sorry about that.  There's a lot of disinterest in the things that I and others care about too.  But it's not just disinterest.  Many of us are not capable of certifying certain PRs.  I recommend you put a time limit on those PRs that aren't getting attention.  If noone raises any objections after a week, merge it.  Noone has a right to complain if they chose not to participate, and we can always submit corrections if errors are found after the fact.

> This PR is not a dot i and cross t thing. It's blocking valuable and important work.

The sooner you add the docs, the sooner you'll be unblocked.

>  There's an Enhancement Request referenced which describes it.

Bugzilla is not a substitute for proper documentation.

andralex:
Please don't let bureaucracy smother progress. Thanks.

andralex (approved):

PR #8347, opened 2018-06-10
Fix Issue 18964 - `-m32` should mean COFF, default is surprising

TurkeyMan:
I mostly just want to see how fast this gets shut-down.
I feel this is important.

### DIFF
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -377,21 +377,23 @@ dmd -cov -unittest myprog.d
         Option("m32",
             "generate 32 bit code",
             `$(UNIX Compile a 32 bit executable. This is the default for the 32 bit dmd.)
-            $(WINDOWS Compile a 32 bit executable. This is the default.
-            The generated object code is in OMF and is meant to be used with the
-            $(LINK2 http://www.digitalmars.com/download/freecompiler.html, Digital Mars C/C++ compiler)).`,
+            $(WINDOWS Compile 32 bit MS-COFF and is meant to be used with the
+            $(LINK2 https://msdn.microsoft.com/en-us/library/dd831853(v=vs.100).aspx, Microsoft Visual Studio 10)
+            or later compiler. This is the default for the 32 bit dmd.`,
             (TargetOS.all & ~TargetOS.dragonFlyBSD)  // available on all OS'es except DragonFly, which does not support 32-bit binaries
         ),
-        Option("m32mscoff",
-            "generate 32 bit code and write MS-COFF object files",
-            TargetOS.windows
+        Option("m32omf",
+            "generate 32 bit code and write OMF object files",
+            `The generated object code is in OMF and is meant to be used with the
+            $(LINK2 http://www.digitalmars.com/download/freecompiler.html, Digital Mars C/C++ compiler)).`,
+            (TargetOS.windows)  // only applicable to windows
         ),
         Option("m64",
             "generate 64 bit code",
             `$(UNIX Compile a 64 bit executable. This is the default for the 64 bit dmd.)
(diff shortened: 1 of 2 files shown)


WalterBright:
Given the endless problems people have installing VC and setting paths to it, no.

There are the endless issues with trying to simultaneously support 6+ versions of VC, and then everything grinding to a halt when a new VC is released which breaks compatibility. I've been on that treadmill before, it would consume >50% of my dev time for years. No.

Dmd is stable, it works out of the box, and we can fix any issues when they do come up.

For example, what if dmd has a problem with MS-LINK 2? Microsoft fixes it with MS-LINK 3. But that does us no good, because half our users will still have MS-LINK 2, and we get the blame for the bug. We will *always* get the blame, and there's nothing we can do to fix it. Been there, done that, not doing it again.

I personally would be stuck with the bill for this. No.

TurkeyMan:
We have lld already bundled with the distribution, so MS-LINK has been removed from the equation hasn't it?
How is situation any different when -m64 already exists?

I'm not suggesting _changing_ anything with respect to expectation or support. I don't see why this would cause you or anyone any additional work... just suggesting making the defaults not surprising to users.
Literally nobody that isn't already firmly involved in this dev community would ever anticipate OMF popping out the other end. Most users have never encountered the kinds of link errors it produces, and would never have even imagined such a thing was possible or existed. People don't understand it, it's really not a good default.

Is D a language for the future, or a language for a couple of decades ago?

WalterBright:
> How is situation any different when -m64 already exists?

At least one can still build programs in Windows. I recall the year or two spent dealing with "I have a VS 2xxx and the linker can't find the VS libraries", and that's just one issue with relying on an ever-changing system we don't control.

> would ever anticipate OMF

Most don't know what an object file even is. I had a hard lesson with that "can't find the libraries" thing. People want, when they're staring with D,
```
dmd hello.d
```
to produce an executable. Not give errors about can't find the library.

> Most users have never encountered the kinds of link errors it produces

Anyone baffled by "undefined symbol" errors won't be able to configure VC or set the LIBPATH either.
```
C:\bug>..\dmd test
test.obj(test)
 Error 42: Symbol Undefined _D4test3fooFZi

C:\cbx>..\dmd test -m64
test.obj : error LNK2019: unresolved external symbol _D4test3fooFZi referenced in function _Dmain
```
I suppose everyone knows what `LNK2019` means, at least optlink gives a URL for where to look.

TurkeyMan:
>```
>dmd hello.d
>```
>to produce an executable. Not give errors about can't find the library.

Right, but this is the point. Windows libs are COFF. Nobody has OMF libs.

```
dmd hello.d mylib.lib
```
(or whatever the command is to supply a lib to the linker, you get the idea)

It's also that `-m64` works like people expect, and then `-m32` which is the logical compliment goes surprisingly wrong.

> Anyone baffled by "undefined symbol" errors won't be able to configure VC or set the LIBPATH either.

We do very well at detection of correct settings these days. I haven't seen issues for years.
Optlink/OMF issues come up all the time however.

TurkeyMan:
No but he went to do the 32bit complement build.

Surely you must acknowledge the asymmetry and probability for surprises?
Also that OMF has no relevance to Windows system?

TurkeyMan:
> Problem solved :-)

The 'problem' is one of confusion/expectation of users.. Giving them instruction is not a solution, this is about intuitive/expected behaviour, that is all.

jacob-carlborg:
Not sure if this is a good idea or not, but regardless it’s a breaking change that needs to follow the regular depreciation path.

TurkeyMan:
I don't have the energy to pursue this. I do think it's important though.

PR #8349, opened 2018-06-10
fix issue 15589 - cleanup/improve PR #8277 - C++ dtors in vtbl

rainers:
I just wanted to get the mangling right, but noticed a couple of other issues. This fix had to revert a couple of changes from #8277, because
- mangling was wrong on Windows (vtable holds (optional) deleting dtor, this uses "?_G")
- C++ dtors were not handled at all after `public:`

This PR also adds a shim for Win32 that allows the dtor to be called through the function pointer in the TpeInfo.


src/dmd/cppmanglewin.d
```diff
@@ -1193,7 +1198,7 @@ private:
             rettype.accept(tmp);
             tmp.flags &= ~MANGLE_RETURN_TYPE;
         }
-        if (!type.parameters || !type.parameters.dim || noargs)
+        if (!type.parameters || !type.parameters.dim)
```
TurkeyMan: All the code I looked had no args mangled into the destructor... and all my tests link. How did you manage to make it mangle with args?

src/dmd/dsymbolsem.d
```diff
@@ -5989,3 +5986,95 @@ void aliasSemantic(AliasDeclaration ds, Scope* sc)
...
+private DtorDeclaration buildExternDDtor(ClassDeclaration cldec, Scope* sc)
+{
+    auto dtor = cldec.dtor;
+    if (!dtor)
+        return null;
+
+    if (cldec.classKind != ClassKind.cpp || !global.params.isWindows || global.params.is64bit)
+        return dtor;
+
+    // the only ABI known to be incompatible is Windows/x86
```
TurkeyMan: Should this apply to 64bit too?
`cldec.dtor` receives an argument either way. That it's not passed on the stack won't lead to a crash, but if `dtor` takes the argument and decides to call `operator delete` and the calling function didn't supply the arg (just whatever was in RDX), then won't it just randomly delete?

TurkeyMan: Sorry, I think you've rearranged it further than I realised...

src/dmd/dsymbolsem.d
```diff
@@ -5989,3 +5986,95 @@ void aliasSemantic(AliasDeclaration ds, Scope* sc)
...
+private FuncDeclaration buildWindowsCppDtor(ClassDeclaration cldec, Scope* sc)
+{
+    auto dtor = cldec.dtor;
+    if (!dtor)
+        return null;
+
+    Parameter param = new Parameter(STC.undefined_, Type.tuns32, Identifier.idPool("del"), null);
+    Parameters* params = new Parameters;
+    (*params).push(param);
+    auto ftype = new TypeFunction(params, Type.tvoidptr, false, LINK.cpp, dtor.storage_class);
```
TurkeyMan: Does the windows dtor also return a value?

src/dmd/dsymbolsem.d
```diff
@@ -5989,3 +5986,95 @@ void aliasSemantic(AliasDeclaration ds, Scope* sc)
...
+    auto func = new DtorDeclaration(dtor.loc, dtor.loc, dtor.storage_class, Id.cppdtor);
+    func.type = ftype;
+    if (dtor.fbody)
+    {
+        Loc loc = dtor.loc;
+        auto stmts = new Statements;
+        auto call = new CallExp(loc, dtor, null);
```
TurkeyMan: Why don't these calls receive 'this' as an argument?

src/dmd/cppmanglewin.d
rainers: VC builds 2 dtors for class CC:
- one without arguments and return type: `??1CC@@UAE@XZ (public: virtual __thiscall CC::~CC(void))` 
- one with an integer argument `??_ECC@@UAEPAXI@Z (public: virtual void * __thiscall CC::`vector deleting destructor'(unsigned int))`. This is mangled as a regular function, but with a special name. This roughly is represented by the code
```
void* CC::~CC(int del)
{
   this->~CC();
   if (del)
      delete (char*)this;
   return (void*) this;
}
```
The latter is the one added to the vtable or called by an explicite `delete`. The fomer is the one called in the dtor chain. So both are necessary when mixing C++ and D in the class hierarchy.

src/dmd/dsymbolsem.d
rainers: Yeah, `cldec.dtor` no longer receives the argument, only he function in the vtable does. If D destroys a class/struct, it always calls the version without argument.

src/dmd/dsymbolsem.d
rainers: `dtor` is a member function, so semantic figures this out. Not sure I need the ubiquitous hack casting away const/immutable, though.

src/dmd/dsymbolsem.d
TurkeyMan: So `xdtor` is the one without the argument? And the one with the argument is the one that's mangled and added to the vtable, but not callable from D?

src/dmd/dsymbolsem.d
TurkeyMan: I saw that everywhere; just rolled with it ;)

src/dmd/cppmanglewin.d
TurkeyMan: Okay great. Thanks for clarifying that! I didn't do enough testing.

TurkeyMan:
LGTM, what's holding the unit test back?

TurkeyMan:
Kinda looks like link-stipping or some sort of dead-code elimination for some of them... but the Windows one just crashed :(

src/dmd/dsymbolsem.d
rainers: Yes.

rainers:
> but the Windows one just crashed :(

I can reproduce that, investigating...

rainers:
It seems I have to back up a bit: `__xdtor` is used to call the (virtual) dtor in druntime instead of the TypeInfo entry (which would not be virtual). That caused a bad mixture of the wrapped and not-wrapped version.

So I'm back to using a default argument and assigning the resulting dtor to `cldec.dtor` and `__xdtor`, but now the "primaryDtor"-detection no longer works...

TurkeyMan:
Right. I thought I landed on that particular configuration I did because of pretty specific reasons like that. It wasn't my first impulse to produce the configuration in my PR.
You could just detect a special name? :P

src/dmd/clone.d
```diff
@@ -978,6 +982,53 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
...
+ *  the shim destructor, semantically analyzed and added to the class as a member
+ */
+private DtorDeclaration buildWindowsCppDtor(AggregateDeclaration ad, DtorDeclaration dtor, Scope* sc)
+{
+    if (!dtor)
+        return null;
+
+    Parameter param = new Parameter(STC.undefined_, Type.tuns32, Identifier.idPool("del"), new IntegerExp(dtor.loc, 0, Type.tuns32));
+    Parameters* params = new Parameters;
+    (*params).push(param);
```
WalterBright: `params.push(param);` should be sufficient.

src/dmd/clone.d
```diff
@@ -978,6 +982,53 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
...
+    auto ftype = new TypeFunction(params, Type.tvoidptr, false, LINK.cpp, dtor.storage_class);
+    auto func = new DtorDeclaration(dtor.loc, dtor.loc, dtor.storage_class, Id.cppdtor);
+    func.type = ftype;
+    if (dtor.fbody)
+    {
+        Loc loc = dtor.loc;
```
WalterBright: `const`

src/dmd/clone.d
```diff
@@ -978,6 +982,53 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
...
+ *
+ * Returns:
+ *  the shim destructor, semantically analyzed and added to the class as a member
+ */
+private DtorDeclaration buildWindowsCppDtor(AggregateDeclaration ad, DtorDeclaration dtor, Scope* sc)
+{
+    if (!dtor)
+        return null;
+
+    Parameter param = new Parameter(STC.undefined_, Type.tuns32, Identifier.idPool("del"), new IntegerExp(dtor.loc, 0, Type.tuns32));
```
WalterBright: Please rename `param` to `del` or something like that.

src/dmd/clone.d
```diff
@@ -978,6 +982,53 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
...
+        auto stmts = new Statements;
+        auto call = new CallExp(loc, dtor, null);
+        call.directcall = true;                                  // Class.__dtor();
+        stmts.push(new ExpStatement(loc, call));                 // TODO: free mem if del != 0
+        stmts.push(new ReturnStatement(loc, new NullExp(loc)));  // return null; (instead of 'this', fails for structs)
+        func.fbody = new CompoundStatement(loc, stmts);
+    }
+    func.generated = true;
+
+    sc = sc.push();
```
WalterBright: Declare a new variable here instead of reusing `sc`. Say, `sc2`.

src/dmd/clone.d
```diff
@@ -978,6 +982,53 @@ extern (C++) FuncDeclaration buildDtor(AggregateDeclaration ad, Scope* sc)
...
+        return null;
+
+    Parameter param = new Parameter(STC.undefined_, Type.tuns32, Identifier.idPool("del"), new IntegerExp(dtor.loc, 0, Type.tuns32));
+    Parameters* params = new Parameters;
+    (*params).push(param);
+    auto ftype = new TypeFunction(params, Type.tvoidptr, false, LINK.cpp, dtor.storage_class);
+    auto func = new DtorDeclaration(dtor.loc, dtor.loc, dtor.storage_class, Id.cppdtor);
+    func.type = ftype;
+    if (dtor.fbody)
+    {
```
WalterBright: A comment here showing what D code is being generated would be nice, as it is clumsy to try and piece it together in one's mind from the AST generating expressions.

src/dmd/dsymbolsem.d
```diff
@@ -5989,3 +5986,48 @@ void aliasSemantic(AliasDeclaration ds, Scope* sc)
...
+    func.type = ftype;
+
+    //Expression ex = new ThisExp(dtor.loc);
+    //ex = new DotVarExp(dtor.loc, ex, dtor);
+    auto call = new CallExp(dtor.loc, dtor, null);
+    call.directcall = true;                   // non-virtual call Class.__dtor();
+    func.fbody = new ExpStatement(dtor.loc, call);
+    func.generated = true;
+
+    sc = sc.push();
```
WalterBright: `auto sc2 = sc.push();`

WalterBright:
Does this work for 32 bit dmd - both -m32 and -m32mscoff ?

src/dmd/dsymbolsem.d
```diff
@@ -5989,3 +5986,48 @@ void aliasSemantic(AliasDeclaration ds, Scope* sc)
...
+ *  a C++ destructor to a destructor with extern(D) calling convention
+ *
+ * Params:
+ *  cldec = the class that contains the destructor to wrap
+ *  sc = the scope in which to analyze the new function
+ *
+ * Returns:
+ *  the shim destructor, semantically analyzed and added to the class as a member
+ */
+private DtorDeclaration buildExternDDtor(ClassDeclaration cldec, Scope* sc)
```
WalterBright: perhaps this function should go into clone.d along with the other Dtor building functions.

rainers:
> Does this work for 32 bit dmd - both -m32 and -m32mscoff ?

I have a slightly revised version locally, that works for -m64 and -m32mscoff. optlink gives me undefined references even though the C++ files defines the symbols...

src/dmd/dsymbolsem.d
rainers: Ok, will move.

rainers:
I amended what works for me locally (but -m32 as described) with review comments addressed.

WalterBright:
> optlink gives me undefined references even though the C++ files defines the symbols...

I have a hard time believing that optlink has such a bug (optlink has no idea what C++ even is, it has no C++ semantics in it). It just connects references with definitions. The overwhelming likelihood is the reference and the definition are different strings.

WalterBright:
The MSC++ destructor menagerie ABI was replicated in Symantec C++ back in the 90's. Although Microsoft changed object file formats, they didn't change the mangling, and probably did not change the binary API. So the -m32 and m32msoff should still be compatible if one of them is.

rainers:
Yeah, I was probably looking at the wrong symbols. DMC only generates the deleting dtor, not the one without argument. These are the symbols that are complained about by the linker.

WalterBright:
dmc++ creates the primary dtor, the scalar deleting destructor, and the vector destructor. See https://github.com/DigitalMars/Compiler/blob/master/dm/src/dmc/dstruct.d#L5133 etc.

rainers:
> dmc++ creates the primary dtor, the scalar deleting destructor, and the vector destructor.

Thanks, so no extra case should be necessary. It seems they are omitted if the class is defined in the cpp-file, I see them when moving the declaration into a header file.

WalterBright:
It looks good to go to me, except it's failing the autotester :-(

rainers:
> It looks good to go to me, except it's failing the autotester :-(

Works on Windows and Linux now. Is the C++ ABI for OSX and FreeBSD different from Linux?

WalterBright:
> Is the C++ ABI for OSX and FreeBSD different from Linux?

FreeBSD and Linux are pretty close, OSX tends to vary. As for C++ destructors specifically, I've never looked into that. Win64 is in another galaxy.

WalterBright:
I fear the workarounds you added for OSX and FreeBSD leave us in a bad place, in that there seem to be a lot of them and there's no documentation for them. I recommend outlining the workarounds with `version (FreeBSD)` and `version (OSX)` so they aren't a mystery to the next guy. Alternatively, create a bugzilla for the issues and in the test case create:
```
version (OSX) version = Bugzilla18xxx;
...
version (Bugzilla18xxx)
{
    ... workaround code ...
}
```

rainers:
Ok, added https://issues.dlang.org/show_bug.cgi?id=18986 and reversed the logic.

PR #8350, opened 2018-06-11
cod2, remove #if1 #else #endif dead code

ghost:


RazvanN7 (approved):
Looks good to me, but maybe @WalterBright has some objections to it

jacob-carlborg (approved):

WalterBright:
I do not approve of these changes. Sometimes I leave such in as documentation of alternative methods.

BBasile:
okay, actually i proposed this because for some reason i landed here : https://github.com/dlang/dmd/pull/1832#issuecomment-15865077. One of stuff deleted here is deactivated since 5 years.

PR #8351, opened 2018-06-11
Fix Issue 15068 - wrong error message on attempting to use type as template

RazvanN7:


src/dmd/dtemplate.d
```diff
@@ -6597,7 +6597,7 @@ extern (C++) class TemplateInstance : ScopeDsymbol
                     Dsymbol s2 = s.getType().toDsymbol(sc);
                     if (!s2)
                     {
-                        error("`%s` is not a template declaration, it is a %s", id.toChars(), s.kind());
+                        error("`%s` is not a template declaration, it is a type", id.toChars());
```
WalterBright: Technically, it is an alias, as type parameters are accepted by alias. However, it could also be an alias to some variable, making the error message still wrong.
I suggest testing to see of s2 is an alias. If it is, then resolve the alias to what it really is using `toAlias()`, then use the original error message.

test/fail_compilation/fail15068.d
```diff
@@ -0,0 +1,16 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail15068.d(15): Error: template instance `T!int` `T` is not a template declaration, it is a type
+fail_compilation/fail15068.d(11): Error: template instance `fail15068.Stuff!int` error instantiating
+---
+*/
```
WalterBright: Add to the comment the URL of the bugzilla entry.

test/fail_compilation/fail15068.d
```diff
@@ -0,0 +1,16 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail15068.d(15): Error: template instance `T!int` `T` is not a template declaration, it is a type
```
ZombineDev: Can we change the error message to something along the lines of:

```
Error: `T!int` is not a valid template instance, because `T` is not a template declaration, but a type (`` == int`).
```

(i.e. is the `template instance` part after `Error: ` mandatory, can we override it?)

src/dmd/dtemplate.d
RazvanN7: I haven't found any examples where that particular error is triggered for something different than a type. It looks like there is another branch for non-types [1] . In addition, using the original error message will never specify that a type is used instead of a template declaration because `kind` never returns "type" for type AST nodes (instead it returns the actual name of the type for basic types and `struct`, `class` etc. for user defined types)

[1] https://github.com/dlang/dmd/blob/master/src/dmd/dtemplate.d#L6625

test/fail_compilation/fail15068.d
RazvanN7: Nice idea. Done.

src/dmd/dtemplate.d
```diff
@@ -6597,7 +6597,7 @@ extern (C++) class TemplateInstance : ScopeDsymbol
                     Dsymbol s2 = s.getType().toDsymbol(sc);
                     if (!s2)
                     {
-                        error("`%s` is not a template declaration, it is a %s", id.toChars(), s.kind());
+                        .error(loc, "`%s` is not a valid template instance, because `%s` is not a template declaration but a type (`%s == %s`)", toChars(), id.toChars(), id.toChars(), s.getType.kind());
```
wilzbach: Isn't `errorSupplemental` intended for such long error messages?

wilzbach (approved):

WalterBright (approved):

PR #8354, opened 2018-06-12
Fix Issue 18970 - DMD segfault due to opDispatch

wilzbach:
The error doesn't get displayed, because for `opDispatch`  gagging is activated:

https://github.com/dlang/dmd/blob/ecc8d829b473e8477110305a6666759e568d94c5/src/dmd/typesem.d#L3056-L3063

(gets called from `TypeStruct::dotExp(e = 'S()', ident = 'y')` via `semanticY` and `resolveUFCS`)

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -4023,7 +4023,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
 
     override void visit(StructDeclaration sd)
     {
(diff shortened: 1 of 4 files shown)


UplinkCoder:
@wilzbach gagging needs to be active for opDispatch, since it's entirely speculative compilation.
you should be able to fix the segfault without turning off gagging.

wilzbach:
Oh it's working (there was just an accidental line mismatch error due to rebasing).
Anyhow, I was trying to point out why the error message added here isn't shown.

The SemaphoreCI error is unrelated.

PR #8356, opened 2018-06-13
Fix Issue 16088 - Parse error for import expression in statement

RazvanN7:


RazvanN7:
Don't mind my branch name

jacob-carlborg:
Hmm, interesting, I had know idea this is/should be possible.

Is it enough to have a compilable test?

RazvanN7:
@jacob-carlborg  It's actually a runnable test.

test/runnable/test16088.d
```diff
@@ -0,0 +1,17 @@
+// REQUIRED_ARGS: -Jrunnable/imports/
```
WalterBright: This is a compile issue, no need to run it. Please move to `compilable/`. Runnable tests are a lot slower to autotest than compilable ones.

src/dmd/parse.d
```diff
@@ -6076,10 +6076,26 @@ final class Parser(AST) : Lexer
...
             {
-                AST.Dsymbols* imports = parseImport();
-                s = new AST.ImportStatement(loc, imports);
-                if (flags & ParseStatementFlags.scope_)
-                    s = new AST.ScopeStatement(loc, s, token.loc);
+                /* https://issues.dlang.org/show_bug.cgi?id=16088
+                 * At this point it can either be an import expression
+                 * or an import declaration. See if the next token after
+                 * `import` is a `(`; if so, then it is an import expression
+                 * and `parseExpression` is called, otherwise it's a
```
WalterBright: `an` not `a`

src/dmd/parse.d
```diff
@@ -6076,10 +6076,26 @@ final class Parser(AST) : Lexer
             }
         case TOK.import_:
             {
-                AST.Dsymbols* imports = parseImport();
-                s = new AST.ImportStatement(loc, imports);
-                if (flags & ParseStatementFlags.scope_)
-                    s = new AST.ScopeStatement(loc, s, token.loc);
+                /* https://issues.dlang.org/show_bug.cgi?id=16088
+                 * At this point it can either be an import expression
+                 * or an import declaration. See if the next token after
```
WalterBright: Add `https://dlang.org/spec/grammar.html#ImportDeclaration` instead of `import declaration`, and ditto for `https://dlang.org/spec/grammar.html#ImportExpression`

src/dmd/parse.d
WalterBright: Remove "and `parseExpression` is called" - it is redundant as the code itself is pretty obvious on that point.

src/dmd/parse.d
WalterBright: Having links to the corresponding Spec place that the code is implementing is very handy for reading the code and ensuring it is in sync with the Spec.

RazvanN7:
@WalterBright done.

jacob-carlborg:
I think you can use a `static assert` to have the same behavior as the runnable test you had before, if you want to.

jacob-carlborg (approved):

PR #8357, opened 2018-06-14
Go.c to d

WalterBright:


src/dmd/backend/go.d
```diff
@@ -0,0 +1,500 @@
...
+
+extern (C++) void mem_free(void* p);
+
+void go_term()
+{
+    vec_free(go.defkill);
+    vec_free(go.starkill);
+    vec_free(go.vptrkill);
+version (Posix)
+{
```
jacob-carlborg: Seems to be some unusual indentations in some places, like this one.

WalterBright: I was trying to minimize diffs with the C version.

jacob-carlborg: I suspected that.

jacob-carlborg: Ah, looking at the individual commits worked a lot better, now it makes more sense. 👍

RazvanN7:
Are you going to make any measurements  to see how the conversion to D affects the compilation time?

WalterBright:
Don't bother reviewing this till I figure out what is going wrong with the OSX32 build.

WalterBright:
Hmm, looks like `clock()` on OSX32 is the source of the problem.

WalterBright:
Definitely it's clock():
```
clock_t clock(void) __asm("_" "clock" "$UNIX2003");
```
:-(

jacob-carlborg:
Looks like 3fa072618a8d9c45920ec6dea8a58dc3fe9535ad passed on all platforms except Windows where it, timed out?

stefan-koch-sociomantic:
Does this still build with VS?

stefan-koch-sociomantic:
@RazvanN7 I'd say this should not have a serious effect since it's in the optimizer which you should turn off if you want fast compiletime

src/dmd/backend/go.d
```diff
@@ -0,0 +1,511 @@
...
+        //clock_t clock() { return 0; }
+    }
+}
+
+clock_t clock() { return 0; }
++/
+
+extern (C++):
+
+extern (C) clock_t os_clock();
```
rainers: `os_clock` is extern(C++) despite the file extension of os.c.

WalterBright:
Ready to go now.

src/dmd/backend/go.d
```diff
@@ -0,0 +1,513 @@
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Copyright:   Copyright (C) 1986-1998 by Symantec
+ *              Copyright (c) 2000-2017 by Digital Mars, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     Distributed under the Boost Software License, Version 1.0.
+ *              http://www.boost.org/LICENSE_1_0.txt
+ * Source:      https://github.com/dlang/dmd/blob/master/src/dmd/backend/go.c
```
rainers: still links to C file.

src/dmd/backend/go.d
```diff
@@ -0,0 +1,513 @@
...
+{
+    version (X86)
+    {
+        //pragma(mangle, "clock$UNIX2003") clock_t clock();
+        //clock_t clock() { return 0; }
+    }
+}
+
+clock_t clock() { return 0; }
++/
```
rainers: test remainder?

src/dmd/backend/go.d
```diff
@@ -0,0 +1,513 @@
...
+ */
+
+int go_flag(char *cp)
+{
+    enum GL     // indices of various flags in flagtab[]
+    {
+        O,all,cnp,cp,cse,da,dc,dv,li,liv,local,loop,
+        none,o,reg,space,speed,time,tree,vbe,MAX
+    }
+    __gshared const char*[GL.MAX] flagtab =
```
rainers: Can you use `static immutable` here?

src/dmd/backend/go.d
```diff
@@ -0,0 +1,513 @@
...
+    for (block* b = startblock; b; b = b.Bnext)
+    {
+        if (!b.Belem)
+            continue;
+        int d = el_countCommas(b.Belem) + 100;
+        if (d > iterationLimit)
+            iterationLimit = d;
+    }
+
+//    optfunc2(iterationLimit);
```
rainers: Why is this comment added?

src/dmd/backend/global.d
```diff
@@ -140,7 +140,8 @@ void util_set32();
 void util_set64();
 int ispow2(uint64_t);
 
-//#if __GNUC__
+version (Posix)
+{
```
rainers: shouldn't the util_malloc/calloc/free functions use (templated) forwarding functions to not repeat the `version(Posix)` condition everywhere? Or just implement `util_*` in util2.c for Posix, too? The parc_* don't seem to be used anyway.

rainers (commented):
Apart from nits, LGTM

src/dmd/backend/go.d
WalterBright: fixed

src/dmd/backend/go.d
WalterBright: removed

src/dmd/backend/go.d
WalterBright: removed

src/dmd/backend/global.d
WalterBright: The memory allocation functions in the back end are a bit of a mess. Straightening them out is something I plan to do, but it's beyond the scope of this PR.

src/dmd/backend/go.d
WalterBright: sure why not

rainers:
It seems the OSX/32 failure is related to this PR. It does not happen always, but the ICE for std.conv can be found in earlier failures, too.

WalterBright:
@rainers does that mean this problem is also happening with other than this PR?

WalterBright:
It works fine on my OSX machine, sigh.

WalterBright:
Adding instrumentation to the autotester, and now it works. What a surprise! A true heisenbug.

rainers:
I bisected the (not so spurious) ICE, it is caused by a wrong definition of CLOCKS_PER_SEC in core.time. It should be 1_000_000 instead of 100 on the system I tested it on.

Not caused by the translation, but the ICE makes the optimization look brittle. Also, using a timeout is no good for reproducible builds, though a 30 second timeout for a single function is not going to be hit too often.

WalterBright:
@rainers Thanks for figuring this out! Can you please submit a PR for core.stc.time? Also, what are you referring to with "the optimization"?

src/dmd/backend/go.d
```diff
@@ -3,56 +3,129 @@
  * $(LINK2 http://www.dlang.org, D programming language).
  *
  * Copyright:   Copyright (C) 1986-1998 by Symantec
- *              Copyright (C) 2000-2018 by The D Language Foundation, All Rights Reserved
+ *              Copyright (c) 2000-2017 by Digital Mars, All Rights Reserved
```
wilzbach: Is this change on purpose? Looks like an accidentally copied header from another file.

wilzbach (approved):

PR #8359, opened 2018-06-15
fix issue 18953 - 32-bit: extern(C++) struct destructor not called cor…

rainers:
…rectly through runtime

mostly fixed by d08e0fb3bff6767bd28516f53393083d86f63045, but 32-bit platforms other than Win32 are affected, too. Also added function attribute inference for generated shim function.

### DIFF
--- a/src/dmd/clone.d
+++ b/src/dmd/clone.d
@@ -1056,11 +1056,12 @@ DtorDeclaration buildExternDDtor(AggregateDeclaration ad, Scope* sc)
     if (!dtor)
         return null;
 
-    // the only ABI known to be incompatible is Windows/x86
-    if (ad.classKind != ClassKind.cpp || !global.params.isWindows || global.params.is64bit)
+    // ABI incompatible on all (?) x86 32-bit platforms
+    if (ad.classKind != ClassKind.cpp || global.params.is64bit)
         return dtor;
 
-    // generate member function that adjusts calling convention (EAX used instead of ECX for 'this'):
+    // generate member function that adjusts calling convention
+    // (EAX used for 'this' instead of ECX on Windows/stack on others):
     // extern(D) void __ticppdtor()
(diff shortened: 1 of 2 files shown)


jacob-carlborg (approved):

rainers:
Actually other 32-bit platforms are affected, too. Why do we have to deviate from the C++ ABI to begin with?

TurkeyMan:
If reviewers could expedite this PR, it should be in 2.081 along with all the other extern(C++) work this cycle.

JinShil:
@TurkeyMan Knowing that you have reviewed this helps.   Please explicitly mark it "Approved" if that's your determination.

src/dmd/clone.d
```diff
@@ -1081,6 +1083,7 @@ DtorDeclaration buildExternDDtor(AggregateDeclaration ad, Scope* sc)
     ad.members.push(func);
     func.addMember(sc2, ad);
     func.dsymbolSemantic(sc2);
+    func.functionSemantic(); // to infer attributes
```
wilzbach: While this inference change makes sense, I'm not sure whether it's a good idea to combine it with a PR that "just adds a test".

wilzbach (approved):
I don't know much about the C++ ABI incompatibility on 32-bit platforms, but this change just lifts the restriction for 64-bit Windows and the test and attribute inference LGTM too.

src/dmd/clone.d
rainers: The commit message already aid otherwise, I've updated the PR description now, too.

rainers:
Does it have to be rebased to stable to make it into 2.081?

wilzbach:
> Does it have to be rebased to stable to make it into 2.081?

Yup.

rainers:
Rebased to stable.

PR #8360, opened 2018-06-15
Fix Issue 18985 - bad error message for += operation on shared Object

RazvanN7:


src/dmd/expressionsem.d
```diff
@@ -4152,7 +4149,19 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         exp.e1 = exp.e1.optimize(WANTvalue);
         exp.e1 = exp.e1.modifiableLvalue(sc, exp.e1);
         exp.type = exp.e1.type;
-        if (exp.checkScalar())
+
+        AggregateDeclaration ad = isAggregate(exp.e1.type);
+        Dsymbol s;
```
wilzbach: Can be moved into the if clause.

src/dmd/expressionsem.d
```diff
@@ -4152,7 +4149,19 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+        if (ad)
+        {
+            s = search_function(ad, Id.opOpAssign);
+            if (s)
```
wilzbach: Can be moved into one combined line: `if (auto s = ...)`

src/dmd/expressionsem.d
```diff
@@ -4152,7 +4149,19 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+            {
+                error(exp.loc, "none of the `opOpAssign` overloads of `%s` are callable for `%s` of type `%s`", ad.toChars(), exp.e1.toChars(), exp.e1.type.toChars());
+                return setError();
+            }
```
wilzbach: Not sure that is correct. Consider the case where a user defines both a shared and a non-shared opOpAssign.

RazvanN7: If that is the case, then this code block is not going to get reached because op_overload [1] will either find the corresponding opOpAssign or it will error. 

[1] https://github.com/dlang/dmd/blob/master/src/dmd/expressionsem.d#L4109

JinShil (approved):

WalterBright (approved):

PR #8361, opened 2018-06-15
Some foreach related fixes

yshui:
Side note: dmd doesn't generate a call to _d_dynamic_cast when casting a Expression to FuncExp. How do you make it do that? I don't even know you could turn it off.

Edit: OK, never mind. I guess type checking is not done when casting `extern(C++)` classes. Seems really unsafe, though. We should check to make sure there is no more case of blind casting in dmd.

### DIFF
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -1926,6 +1926,8 @@ else
         Expression flde = new FuncExp(fs.loc, fld);
         flde = flde.expressionSemantic(sc);
         fld.tookAddressOf = 0;
+        if (flde.op == TOK.error)
+            return null;
         return cast(FuncExp)flde;
     }
 
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -1907,10 +1907,10 @@ else
             LcopyArg:
                 id = Identifier.generateId("__applyArg", cast(int)i);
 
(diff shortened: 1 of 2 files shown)


src/dmd/statementsem.d
```diff
@@ -1921,11 +1921,13 @@ else
         auto tf = new TypeFunction(params, Type.tint32, 0, LINK.d, stc);
         fs.cases = new Statements();
         fs.gotos = new ScopeStatements();
-        auto fld = new FuncLiteralDeclaration(fs.loc, Loc.initial, tf, TOK.delegate_, fs);
+        auto fld = new FuncLiteralDeclaration(fs.loc, fs.endloc, tf, TOK.delegate_, fs);
         fld.fbody = fs._body;
         Expression flde = new FuncExp(fs.loc, fld);
         flde = flde.expressionSemantic(sc);
         fld.tookAddressOf = 0;
+        if (flde.op == TOK.error)
```
wilzbach: Any chance you can drop a quick test for this to fail_compilation?

wilzbach (commented):
Even though this looks trivial it would be nice to have coverage for it, so that it can't regress in the future.

src/dmd/statementsem.d
yshui: I will add the test case in issue 18994

WalterBright (approved):

jacob-carlborg (approved):

JinShil (approved):

PR #8362, opened 2018-06-16
fix Issue 18966 - extern(C++) constructor should match C++ semantics …

rainers:
…assigning vtable

set vtbl after exlicit or implicit call to super() in C++ constructors

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -3105,6 +3105,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             return f;
         }
 
+        bool isSuper = false;
         if (exp.e1.op == TOK.dotVariable && t1.ty == Tfunction || exp.e1.op == TOK.dotTemplateDeclaration)
         {
             UnaExp ue = cast(UnaExp)exp.e1;
@@ -3250,7 +3251,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             auto ad = sc.func ? sc.func.isThis() : null;
             auto cd = ad ? ad.isClassDeclaration() : null;
 
-            const bool isSuper = exp.e1.op == TOK.super_;
+            isSuper = exp.e1.op == TOK.super_;
             if (isSuper)
             {
                 // Base class constructor call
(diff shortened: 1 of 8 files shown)


TurkeyMan:
LGTM :)

TurkeyMan:
Hey @MartinNowak, I saw you branched the beta for the next release.
I just wanted you to be aware that this PR completes a large body of work, which I think will be one of the biggest headline features for 2.081. Just wanted to check that we'll be able to get it in there, and that the branch isn't frozen?

TurkeyMan:
If reviewers could expedite this PR, it should be in 2.081 along with all the other extern(C++) work this cycle.

wilzbach (commented):
The changes look good to me based on the diff, but I'm not that familiar with C++, s.t. I would feel comfortable with approving it.

rainers:
Rebased to stable.

dnadlinger (approved):

src/dmd/expressionsem.d
```diff
@@ -3588,6 +3589,32 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+
+        if (isSuper)
+        {
+            auto ad = sc.func ? sc.func.isThis() : null;
+            auto cd = ad ? ad.isClassDeclaration() : null;
+            if (cd && cd.classKind == ClassKind.cpp)
+            {
+                // if super is defined in C++, it sets the vtable pointer to the base class
+                // so we have to rewrite it, but still return 'this' from super() call:
+                // (auto tmp = super(), this.__vptr = __vtbl, tmp)
```
kinke: This assignment cannot be interpreted at compile-time (unsupported cast), see https://forum.dlang.org/thread/ghusgzuqpcskhwzmlwbh@forum.dlang.org.

TurkeyMan: How does CTFE assign vtables?

TurkeyMan: Is this actually an issue? Any class with an opaque constructor (ie, is defined in C++) can't possibly CTFE...?

kinke: Yep, if there's an external ctor, CTFE will fail anyway. If they're all in D, it used to work before, so it's clearly a regression, and resetting the vptr can probably be safely skipped during CTFE (something like `__ctfe || (this.__vptr = __vtbl)`).

kinke: => #8533

TurkeyMan: Yeah, if that logic is only performed when the base constructor is extern, then it should still be fine.

PR #8365, opened 2018-06-17
prelude to PR 8357

WalterBright:
Since https://github.com/dlang/dmd/pull/8357 suffers from a heisenbug somewhere in the compiler, this PR pulls out all the non-`go.d` changes of more general interest, so at least we can isolate 8357.

### DIFF
--- a/src/posix.mak
+++ b/src/posix.mak
@@ -300,7 +300,7 @@ endif
 
 ifneq (gdc, $(HOST_DMD_KIND))
  ifeq (,$(findstring 2.068,$(HOST_DMD_VERSION)))
-  BACK_BETTERC = -betterC
+  BACK_BETTERC = -mv=dmd.backend=$C -betterC
  endif
 endif
 
--- a/src/win32.mak
+++ b/src/win32.mak
@@ -524,7 +524,7 @@ $G/dt.obj : $C\dt.h $C\dt.c
 	$(CC) -c -o$@ $(MFLAGS) $C\dt
 
 $G/dvec.obj : $C\dvec.d
-	$(HOST_DC) -c -of$@ $(DFLAGS) -betterC $C\dvec
+	$(HOST_DC) -c -of$@ $(DFLAGS) -betterC -mv=dmd.backend=$C $C\dvec
 
 $G/dwarf.obj : $C\dwarf.h $C\dwarf.c
 	$(CC) -c -o$@ $(MFLAGS) $C\dwarf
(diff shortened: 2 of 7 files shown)


src/posix.mak
```diff
@@ -300,7 +300,7 @@ endif
 
 ifneq (gdc, $(HOST_DMD_KIND))
  ifeq (,$(findstring 2.068,$(HOST_DMD_VERSION)))
-  BACK_BETTERC = -betterC
+  BACK_BETTERC = -mv=dmd.backend=$C -mv=dmd.backend.dlist=dmd/tk/dlist -betterC
```
wilzbach: Consistently fails on FreeBSD with DMD 2.079.0:

```
dmd/backend/el.d(18): Error: module `dlist` is in file 'dmd/tk/dlist.d' which cannot be read
```

Now that the dlist PR has been merged, do we still need `-mv=dmd.backend.dlist=dmd/tk/dlist`?
Also I'm not sure why we even need the `dmd.backend` flag here? Shouldn't the D compiler find the module automatically by being in the `src` directory as current working directory and thus `dmd.backend.dvec` resulting in a lookup of `src/dmd/backend/dvec.d`?

src/win32.mak
```diff
@@ -524,7 +524,7 @@ $G/dt.obj : $C\dt.h $C\dt.c
 	$(CC) -c -o$@ $(MFLAGS) $C\dt
 
 $G/dvec.obj : $C\dvec.d
-	$(HOST_DC) -c -of$@ $(DFLAGS) -betterC $C\dvec
+	$(HOST_DC) -c -of$@ $(DFLAGS) -betterC -mv=dmd.backend=$C $C\dvec
```
jacob-carlborg: Is this necessary when the file has the correct module declaration, `dmd.backend.dvec`?

jacob-carlborg (approved):

src/win32.mak
WalterBright: yes, because otherwise the compiler cannot find it.

klickverbot: The module is in `src/dmd/backend/dvec.d`. Why wouldn't having `src` in the include path be enough for that?

WalterBright: dvec.d is also used by dmc++ which has a different directory structure. The whole point of `-mv` is so different directory structures can be accommodated easily. We've discussed this innumerable times, let's just agree to disagree on the value of `-mv`.

PR #8368, opened 2018-06-18
fix issue 18984 - stack struct's which are returned causes incorrect debuginfo.

rainers:
swap the names of the hidden return value and the NRVO variable and do not emit debug info for the unused original variable.

### DIFF
--- a/src/dmd/glue.d
+++ b/src/dmd/glue.d
@@ -917,8 +917,15 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
         char[5+4+1] hiddenparam = void;
         __gshared int hiddenparami;    // how many we've generated so far
 
-        sprintf(hiddenparam.ptr,"__HID%d",++hiddenparami);
-        shidden = symbol_name(hiddenparam.ptr,SCparameter,thidden);
+        const(char)* name;
+        if (fd.nrvo_can && fd.nrvo_var)
+            name = fd.nrvo_var.ident.toChars();
+        else
+        {
+            sprintf(hiddenparam.ptr, "__HID%d", ++hiddenparami);
+            name = hiddenparam.ptr;
+        }
+        shidden = symbol_name(name, SCparameter, thidden);
         shidden.Sflags |= SFLtrue | SFLfree;
         if (fd.nrvo_can && fd.nrvo_var && fd.nrvo_var.nestedrefs.dim)
             type_setcv(&shidden.Stype, shidden.Stype.Tty | mTYvolatile);
--- a/src/dmd/tocsym.d
+++ b/src/dmd/tocsym.d
@@ -137,24 +137,35 @@ Symbol *toSymbol(Dsymbol s)
             //printf("VarDeclaration.toSymbol(%s)\n", vd.toChars());
             assert(!vd.needThis());
 
-            Symbol *s;
+            const(char)[] id;
+            import dmd.root.outbuffer : OutBuffer;
+            OutBuffer buf;
+            bool isNRVO = false;
             if (vd.isDataseg())
             {
-                import dmd.root.outbuffer : OutBuffer;
-                OutBuffer buf;
                 mangleToBuffer(vd, &buf);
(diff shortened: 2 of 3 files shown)


src/dmd/tocsym.d
```diff
@@ -137,24 +137,35 @@ Symbol *toSymbol(Dsymbol s)
...
             else
             {
-                const id = vd.ident.toChars();
-                s = symbol_calloc(id, cast(uint)strlen(id));
+                id = vd.ident.toString();
+                if (FuncDeclaration fd = vd.toParent2().isFuncDeclaration())
+                {
+                    if (fd.nrvo_can && fd.nrvo_var == vd)
+                    {
+                        buf.write("__HID_".ptr, 6);
```
jacob-carlborg: Please use `writestring`, taking a D string.

src/dmd/tocsym.d
```diff
@@ -137,24 +137,35 @@ Symbol *toSymbol(Dsymbol s)
             //printf("VarDeclaration.toSymbol(%s)\n", vd.toChars());
             assert(!vd.needThis());
 
-            Symbol *s;
+            const(char)[] id;
+            import dmd.root.outbuffer : OutBuffer;
+            OutBuffer buf;
+            bool isnrvo = false;
```
jacob-carlborg: Please use D naming conventions, i.e. `isNrvo` or `isNRVO`.

src/dmd/glue.d
```diff
@@ -917,8 +917,15 @@ void FuncDeclaration_toObjFile(FuncDeclaration fd, bool multiobj)
...
         __gshared int hiddenparami;    // how many we've generated so far
 
-        sprintf(hiddenparam.ptr,"__HID%d",++hiddenparami);
-        shidden = symbol_name(hiddenparam.ptr,SCparameter,thidden);
+        const(char)* name;
+        if (fd.nrvo_can && fd.nrvo_var)
+            name = fd.nrvo_var.ident.toChars();
+        else
+        {
+            sprintf(hiddenparam.ptr, "__HID%d", ++hiddenparami);
```
jacob-carlborg: It would be nice to not have to use C strings. Perhaps you can copy or move this utility function [1].

[1] https://github.com/dlang/dmd/blob/806f105da1d802919f5741ecc1568c9623e5377f/src/dmd/objc_glue.d#L521-L543

jacob-carlborg (changes requested):

TurkeyMan:
Rebase on stable? It would be ideal to not wait another full release cycle for this fix. We have people hitting on this stuff actively.

src/dmd/tocsym.d
rainers: > Please use writestring, taking a D string.
Done.

> Please use D naming conventions, i.e. isNrvo or isNRVO.
Done.

src/dmd/glue.d
rainers: This use of sprintf was there before and ubiquitous in the dmd source code. I'd rather leave it to some other PR to change that.

rainers:
Rebased to stable and added test.

jacob-carlborg:
> Rebase on stable?

As far as I know, only regressions should target stable.

src/dmd/glue.d
jacob-carlborg: Right, didn't look close enough on what was removed.

TurkeyMan:
Yeah, but if I knew that stable was gonna cut yesterday, I would have done everything I could to get this one in there before then. This is a super-important bug fix, and it really won't be good if we have to wait another 2 months for the fix when it was already good to go this time round.

WalterBright (approved):

jacob-carlborg (approved):

rainers:
I had to change the test for Win32, as it passes the hidden pointer in EAX, so the check for a positive offset from EBP no longer works. It now checks that the symbol is at offset -4 as EAX is stored there.

TurkeyMan:
Thanks again Rainer! I owe you a whole lot of beers! :)

PR #8370, opened 2018-06-18
Fix Issue 18266 - ICE: should allow reusing identifier in declaration…

RazvanN7:
…s in disjoint scopes in a function

### DIFF
--- a/test/fail_compilation/fail18266.d
+++ b/test/fail_compilation/fail18266.d
@@ -1,0 +1,28 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail18266.d(22): Error: declaration `fail18266.main.S` is already defined in another scope in `main` at line `14`
+---
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=18266
+
+void main()
+{
+    foreach (i; 0 .. 10)
+    {
+        struct S
+        {
+            int x;
+        }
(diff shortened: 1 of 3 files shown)


wilzbach (approved):

test/fail_compilation/fail18266.d
```diff
@@ -0,0 +1,28 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail18266.d(22): Error: declaration `fail18266.main.S` is already defined in another scope in `main`
```
jacob-carlborg: Would it be complicated to add the line number of where the symbol is first defined to the error message?

RazvanN7: is this better?

jacob-carlborg: Yes, but perhaps as a supplementary error message.

RazvanN7: I think it's better this way, because the supplemental error message would be too much text with little information

jacob-carlborg: This is how the error message would look like with Clang:

```
main.cpp:5:6: error: redefinition of 'foo'
void foo() {}
     ^
main.cpp:3:6: note: previous definition is here
void foo() {}
     ^
1 error generated.
```

PR #8371, opened 2018-06-18
Fix Issue 17656 - Enum member circular reference error

RazvanN7:
```d
enum E {
    AAA = S.BBB
}

struct S {
    enum SZAQ = E.AAA;
    enum BBB = 8080;
}
```

When `E` is semantically analyzed, each enum member has to go through the semantic pass. When `AAA = S.BBB` is reached, the compiler marks that `AAA` has started the first semantic pass [1]; if semantic1 is started again on AAA before the first attempt was finished, the compiler issues a "circular reference error" [2]. This is the correct behavior if we are dealing with circular references inside the enum declaration, but it is wrong on the general case, as this examples shows: part of the semantical analysis of `AAA` is to resolve `S.BBB`; when the compiler tries to resolve `BBB` it notices that semantic1 wasn't performed on S so it starts it. Now semantic1 is done for `enum SZAQ = E.AAA` which calls semantic1 for `E.AAA` so now we end up semantically analyzing AAA again thus hitting the error from [2].

The fix makes it so that the error in [2] is issued only for circular references to enum members inside the same enum, because the other cases are caught by other checks.

[1] https://github.com/dlang/dmd/blob/master/src/dmd/dsymbolsem.d#L2183
[2] https://github.com/dlang/dmd/blob/master/src/dmd/dsymbolsem.d#L2166

### DIFF
--- a/src/dmd/typesem.d
+++ b/src/dmd/typesem.d
@@ -56,6 +56,8 @@ import dmd.target;
 import dmd.tokens;
 import dmd.typesem;
 
+enum LOGDOTEXP = 0;         // log ::dotExp()
+
 /************************************
  * Transitively search a type for all function types.
  * If any function types with parameters are found that have parameter identifiers
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -2165,7 +2165,7 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
 
         if (em.errors || em.semanticRun >= PASS.semanticdone)
             return;
-        if (em.semanticRun == PASS.semantic)
+        if (em.semanticRun == PASS.semantic && sc == em._scope)
         {
             em.error("circular reference to `enum` member");
             return errorReturn();
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -54,7 +54,6 @@ import dmd.tokens;
 import dmd.typesem;
 import dmd.visitor;
 
-enum LOGDOTEXP = 0;         // log ::dotExp()
 enum LOGDEFAULTINIT = 0;    // log ::defaultInit()
 
 extern (C++) __gshared int Tsize_t = Tuns32;
--- a/test/compilable/test17656.d
+++ b/test/compilable/test17656.d
@@ -1,0 +1,12 @@
+// https://issues.dlang.org/show_bug.cgi?id=17656
+
+enum E
+{
+    AAA = S.BBB
+}
+
+struct S
+{
+    enum SZAQ = E.AAA;
+    enum BBB = 8080;
+}


stefan-koch-sociomantic:
This PR needs a proper description in commit message.
The reason why this fixes anything is not oblivious.

src/dmd/typesem.d
```diff
@@ -56,6 +56,8 @@ import dmd.target;
 import dmd.tokens;
 import dmd.typesem;
 
+enum LOGDOTEXP = 0;         // log ::dotExp()
```
wilzbach: Why did you move this declaration?

RazvanN7: because it was a leftover from creating the dotExp visitor. You will notice that it is not used at all in mtype.d . I did some debugging in typesem, so I needed the LOGDOTEXP definition

stefan-koch-sociomantic:
@RazvanN7 Thanks for describing it.
But this is github only to make it part of the commit message you need to reword the commit which you can do with ``git rebase -i @~1` and then putting an `r` instead of the `pick`.

stefan-koch-sociomantic:
@RazvanN7 I cannot come up with an example right now, but this looks like it invites uncaught forward-reference errors.
Please try your best to come up with an example that'll be broken by this change.

RazvanN7:
@stefan-koch-sociomantic I can't find any examples.

Syniurge:
> @RazvanN7 I cannot come up with an example right now, but this looks like it invites uncaught forward-reference errors.

I cannot find an example either, but this is *almost* what happens in this case:
```d
enum E
{
    AAA = S.SZAQ
}

struct S
{
    enum SZAQ = E.AAA;
    enum BBB = 8080;
}
```
`em.error("circular reference to `enum` member");` gets skipped but fortunately another circular referencing error gets triggered later:
```
test17656.d(3): Error: circular reference to variable `test17656.S.SZAQ`
```

-----

@RazvanN7 I don't think the check is correct, it fixes my case because `sc` comes from `expressionSemantic(DotIdExp)`. But `sc` may also be `em._scope` in other cases, for example in:

```d
enum E
{
    AAA = S.BBB
}

alias MLKJ = E.AAA;

struct S
{
    enum SZAQ = MLKJ;
    enum BBB = 8080;
}
```

Thanks for looking into it, those errors are pretty tricky. I recently re-started working on fixing them too, I hope I can see it through its completion this time.

RazvanN7:
@Syniurge You are right, but looking at the way semantic analysis is performed, the strategy of identifying circular references is very weak. I think that the best solution is to not check for circular references during semantic 1 and just return when the error is issued. That way, we can check for circular references during semantic 2 or 3 by just following the chain of assignments.

PR #8373, opened 2018-06-18
reduce reliance on global.errors

WalterBright:


wilzbach (approved):

src/dmd/dimport.d
```diff
@@ -120,18 +120,24 @@ extern (C++) final class Import : Dsymbol
         return si;
     }
 
-    void load(Scope* sc)
+    /*******************************
+     * Load this module.
+     * Returns:
+     *  true for errors, false for success
```
Geod24: `true` for error is quite counter-intuitive.

WalterBright: Normally 0 (false) is success, and !=0 (true) for errors, because of returning error codes.

jacob-carlborg: This is not returning an error code so these conventions don't apply 😉.

PR #8383, opened 2018-06-20
[C++] Put ClassKind enum inside a struct.

ibuclaw:
So its members are properly scoped on the C++ side.

### DIFF
--- a/src/dmd/aggregate.h
+++ b/src/dmd/aggregate.h
@@ -75,14 +75,17 @@ FuncDeclaration *buildPostBlit(StructDeclaration *sd, Scope *sc);
 DtorDeclaration *buildDtor(AggregateDeclaration *ad, Scope *sc);
 FuncDeclaration *buildInv(AggregateDeclaration *ad, Scope *sc);
 
-enum ClassKind
+struct ClassKind
 {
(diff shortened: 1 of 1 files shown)


src/dmd/aggregate.h
```diff
@@ -75,14 +75,18 @@ FuncDeclaration *buildPostBlit(StructDeclaration *sd, Scope *sc);
...
-enum ClassKind
+struct ClassKind
 {
-    /// the aggregate is a d(efault) struct/class/interface
-    d,
-    /// the aggregate is a C++ struct/class/interface
-    cpp,
-    /// the aggregate is an Objective-C class/interface
-    objc
+    typedef int Type;
```
klickverbot: The usual C++03 idiom would be `struct ClassKind { enum Type { d, cpp, objc }; };`. Any reason not to use that?

ibuclaw: No reason whatsoever, this was just a quick copy pasto from ClassFlags.

RazvanN7 (approved):

src/dmd/aggregate.h
ibuclaw: Yes there is a reason. 32bit compilers align this type differently.

PR #8384, opened 2018-06-20
provide location of called function in message

WalterBright:
In a heavily overloaded function, it can be hard to determine which one was matched.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -2303,8 +2303,8 @@ extern (C++) abstract class Expression : RootObject
             {
                 if (!loc.isValid()) // e.g. implicitly generated dtor
                     loc = sc.func.loc;
-                error("`@safe` %s `%s` cannot call `@system` %s `%s`",
-                    sc.func.kind(), sc.func.toPrettyChars(), f.kind(), f.toPrettyChars());
+                error("`@safe` %s `%s` cannot call `@system` %s `%s` declared at %s",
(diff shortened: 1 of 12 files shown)


RazvanN7:
Looks like you need to fix fail_compilation/fail328.d too. LGTM otherwise.

Geod24:
Only for `@safe` ? Not for impure & other attributes ?

WalterBright:
This is trying it out. It can be expanded later.

jacob-carlborg:
> Maybe make this a supplemental line?

Yes, this allows editors and IDEs to create links back to the source code.

wilzbach:
> Looks like you need to fix fail_compilation/fail328.d too. 

FYI as I'm not sure everyone is aware of this recent change. You can do the following to automatically update most of the tests (except they have multiple `TEST_OUTPUT` areas):

```
./run.d AUTO_UPDATE=1
```

This is also supported by the older Makefile:

```
make AUTO_UPDATE=1
```

Or just everything in `fail_compilation`:

```
./run.d fail_compilation AUTO_UPDATE=1
```

WalterBright:
Looks like the most over-tested feature ever :-)

WalterBright:
@RazvanN7 ready now!

Geod24 (approved):

jacob-carlborg:
No supplemental line? 😞

PR #8387, opened 2018-06-20
[C++] Add declaration for ClassDeclaration::vtblSymbol.

ibuclaw:
As per #8362, this function also needs calling from the backend also.


rainers (approved):
Sorry for forgetting this update.

PR #8390, opened 2018-06-21
add some const to mtype.d

WalterBright:


src/dmd/mtype.d
```diff
@@ -6468,7 +6468,7 @@ extern (C++) final class TypeTuple : Type
             {
                 for (size_t i = 0; i < tt.arguments.dim; i++)
                 {
-                    Parameter arg1 = (*arguments)[i];
+                    const Parameter arg1 = (*arguments)[i];
                     Parameter arg2 = (*tt.arguments)[i];
```
jacob-carlborg: Can this not be made `const`?

RazvanN7 (approved):

src/dmd/mtype.d
RazvanN7: why not?

jacob-carlborg: Because it's not marked as `const`?

RazvanN7: Oh sorry, I was under the impression you were asking that arg1 is not made `const`. As for `arg2`, I see no reason why it shouldn't

WalterBright: It can't be until the entire hierarchy of equals() marks it const.

jacob-carlborg: Ok.

PR #8391, opened 2018-06-21
Adapt dmd.root.longdouble to LDC

kinke:
* For LDC, prefer more robust & safe LLVM inline assembly over DMD-style `asm`.
* Fix unittest compile errors (must not be pure as it accesses mutable globals) by moving the block before module-level `pure:`.
* Remove packed-ness of `longdouble_soft` struct (`align(2)`) and prefer natural alignment (`long.alignof` = 8 bytes on Win32/64, so a size of 16 bytes). Just my personal preference wrt. non-bad practices after having had to deal with numerous alignment issues for LDC on non-x86 architectures.

### DIFF
--- a/src/dmd/root/longdouble.d
+++ b/src/dmd/root/longdouble.d
@@ -82,12 +82,13 @@ void ld_clearfpu()
 }
 
 pure:
-@safe:
+@trusted: // LDC: LLVM __asm is @system AND requires taking the address of variables
 
-align(2) struct longdouble_soft
+struct longdouble_soft
 {
 nothrow @nogc pure:
-    ulong mantissa = 0xC000000000000001UL; // default to snan
+    // DMD's x87 `real` on Windows is packed (alignof = 2 -> sizeof = 10).
+    align(2) ulong mantissa = 0xC000000000000001UL; // default to snan
     ushort exp_sign = 0x7fff; // sign is highest bit
 
     this(ulong m, ushort es) { mantissa = m; exp_sign = es; }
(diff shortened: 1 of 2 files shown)


kinke:
Pinging @rainers.

kinke:
Which one, the removal of the little helpers with misleading (because unused) template param, or the usage of LLVM `__asm` in general? The existing DMD `asm` doesn't work with LDC; e.g., we don't handle a param differently depending on whether it is passed indirectly by value (Win64) or on the stack (Win32) (which I consider a DMD bug, it's a param and as such a local lvalue in both cases).

kinke:
And wrt. maintenance overhead, this has got to be an improvement over the replaced LDC workarounds. ;)

WalterBright:
> which I consider a DMD bug, it's a param and as such a local lvalue in both cases

That's the Microsoft Win64 parameter passing protocol. I don't like it, but it's not a dmd bug. I went to considerable effort to match what VC++ does.

kinke:
> but it's not a dmd bug. I went to considerable effort to match what VC++ does.

I think the asm should be ABI agnostic when it comes to accessing a variable, regardless of whether it's a local or a parameter. I'd expect it to represent the lvalue in all cases, regardless of how the protocol looks like. As the MS compiler doesn't support inline asm for x64 AFAIK, there's nothing to match (?).

src/dmd/root/longdouble.d
```diff
@@ -81,10 +81,75 @@ void ld_clearfpu()
...
+    v = cast(ulong)ldu;
+    u = u * (1L << 23);
+    assert(u == v);
+}
+
+//////////////////////////////////////////////////////////////
+
 pure:
-@safe:
+@trusted: // LDC: LLVM __asm is @system AND requires taking the address of variables
```
Geod24: `@trusted:` at the top of a module is just asking for trouble

src/dmd/root/longdouble.d
```diff
@@ -324,7 +388,14 @@ void ld_setull(longdouble_soft* pthis, ulong d)
     d ^= (1L << 63);
     version(AsmX86)
     {
-        mixin(fld_local!("twoPow63"));
+        version(LDC)
+        {
+            // cannot access mutable global in pure function
+            static __gshared immutable _twoPow63 = longdouble_soft(1UL << 63, 0x3fff + 63);
```
rainers: Just make the constants `const`. This also allows keeping the unittest at the bottom (but they need `@system` anyway).

For some reason, this needs an "offset" to be added for "fld_local" for D_InlineAsm_X86_64. "fld_local" is misnamed anyway, it should rather be called "fld_global".

src/dmd/root/longdouble.d
rainers: Not to happy about it, but current state is also stetching the concept of safety a bit, because separate trusted asm sections are combined in each function. So maybe it is ok to mark most of the module as `@trusted`.

You can reenable `@safe` before the constants starting with `ld_qnan`, though.

src/dmd/root/longdouble.d
```diff
@@ -186,63 +251,62 @@ nothrow @nogc pure:
...
-    alias fld_parg = fld_arg;
-    alias fstp_parg = fstp_arg;
-    string fld_local(string arg)()
-    {
-        return "asm nothrow @nogc pure @trusted { lea RAX, " ~ arg ~ "; " ~ fld_eax!arg ~ " }";
-    }
+    string fld_arg  (string arg)() { return `__asm("fldt $0",  "*m,~{st}", &` ~ arg ~ `);`; }
+    string fstp_arg (string arg)() { return `__asm("fstpt $0", "*m,~{st}", &` ~ arg ~ `);`; }
+    string fld_parg (string arg)() { return `__asm("fldt $0",  "*m,~{st}",  ` ~ arg ~ `);`; }
+    string fstp_parg(string arg)() { return `__asm("fstpt $0", "*m,~{st}",  ` ~ arg ~ `);`; }
```
rainers: This is much better than my dirty hack, but I couldn't get it to work when I tried. If we upgrade LDC to 1.9.0, we could also use saner dmd-like asm.

src/dmd/root/longdouble.d
kinke: > Just make the constants const. 

I already tried that, but get this when making `twoPow63` const, with DMD 2.080 and `-m32mscoff`/`-m64`:

```
src\dmd\root\longdouble.d(756): Error: integer constant expression expected instead of longdouble_soft(9223372036854775808LU, cast(ushort)16446u)
src\dmd\root\longdouble.d-mixin-398(398): Error: bad type/size of operands lea
```

src/dmd/root/longdouble.d
kinke: Done.

src/dmd/root/longdouble.d
kinke: With LDC 1.9 and fixed `extended ptr` semantics for DMD `asm` (DMD doesn't support `extended` apparently), the semantics of variables in `asm` apparently still differ. The current asm still doesn't work, I tried that first.

LDC translates these DMD `asm` fragments internally to LLVM `__asm` fragments and passes in variables as LLVM-IR-values, way before register allocation etc., so getting that right is probably not trivial.

`__asm` directly on the other hand is simpler, as you usually explicitly pass in the address of variables to be read from or written to.

src/dmd/root/longdouble.d
rainers: Passed for x64 for me (but no longer now), but this seems to work for all cases:
```
        auto p = &twoPow63;
        mixin(fld_parg!("p"));
```
With this we can get rid of fld_local completely.

kinke: Nice; reworked + squashed.

rainers:
Looks good, but tests fail because of misalignment of longdouble between D and C++ when compiling with dmd, because it uses `real` instead of longdouble_soft i.e. alignment 2.

src/dmd/root/longdouble.d
```diff
@@ -82,12 +82,13 @@ void ld_clearfpu()
 }
 
 pure:
-@safe:
+@trusted: // LDC: LLVM __asm is @system AND requires taking the address of variables
```
wilzbach: Why not
```d
version(LDC)
@trusted:
else
@safe:
```

WalterBright:
I'll defer to @rainers on this one.

rainers (approved):
LGTM, but the unrelated Semaphore failure blocks merging.

PR #8389, opened 2018-06-21
Use supplemental error message for called function

jacob-carlborg:
By using a supplemental error message the compiler will output the location of the called function on a separate line, with the same syntax as a regular error message. This allows editors and IDEs to generate links back to the source code without any special treatment for these messages.

The following example:

```d
void foo() {}

@safe void main()
{
    foo();
}
```

Will now generate the following error message when compiled:

```
main.d(5): Error: @safe function D main cannot call @system function main.foo
main.d(1):        main.foo is declared here
```

PR #8394, opened 2018-06-22
Improve compile_output regex

wilzbach:
See also: https://github.com/dlang/dmd/pull/8389#issuecomment-399076936

CC @jacob-carlborg

I can't reproduce your issue on my Linux system, but maybe this fixes it for you?

### DIFF
--- a/test/tools/d_do_test.d
+++ b/test/tools/d_do_test.d
@@ -747,7 +747,7 @@ int tryMain(string[] args)
             }
 
             compile_output = compile_output.unifyNewLine();
-            compile_output = std.regex.replace(compile_output, regex(`^DMD v2\.[0-9]+.*\n DEBUG$`, "m"), "");
+            compile_output = std.regex.replace(compile_output, regex(`^DMD v2\.[0-9]+.*\n? DEBUG$`, "m"), "");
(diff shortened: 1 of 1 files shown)


jacob-carlborg (approved):

jacob-carlborg:
It fixes the problem for me.

jacob-carlborg:
But the split/grouped output is more annoying since that fails to automatically update.

wilzbach:
> But the split/grouped output is more annoying since that fails to automatically update.

Are you referring to the case when a file has multiple `TEST_OUTPUT` sections?
Yeah I haven't found an easy way to hack support for this in :/

However, there's one "easy" solution: just remove the all `TEST_OUTPUT` sections except the first one and rerun the test file with `AUTO_UPDATE=1` twice.

jacob-carlborg:
> Are you referring to the case when a file has multiple TEST_OUTPUT sections?
Yeah I haven't found an easy way to hack support for this in :/

Yes.

> However, there's one "easy" solution: just remove the all TEST_OUTPUT sections except the first one and rerun the test file with AUTO_UPDATE=1 twice.

Already done that. I tried to keep the sections first, but gave up on that.

PR #8395, opened 2018-06-22
Require test rerun if DMD has been rebuilt

wilzbach:
tl;dr: when DMD is rebuilt, all test results should be invalidated and be rerun.


wilzbach:
Sorry the "[Test]" prefix was to symbolize that this is about the test system, I didn't anticipate it could be misunderstood as this PR being a test.

JinShil (approved):

PR #8396, opened 2018-06-23
Fix Issue 19018 - lexer do not allow invalid integer literal

kubo39:
This is blocked by https://github.com/dlang/druntime/pull/2232

### DIFF
--- a/src/dmd/lexer.d
+++ b/src/dmd/lexer.d
@@ -1803,6 +1803,8 @@ class Lexer : ErrorHandler
         int d;
         bool err = false;
         bool overflow = false;
+        bool anyBinaryDigitsUS = false;
+        bool anyHexDigitsNoSingleUS = false;
         dchar c = *p;
         if (c == '0')
         {
@@ -1861,6 +1863,10 @@ class Lexer : ErrorHandler
             {
             case '0':
             case '1':
+                if (base == 2 && !anyBinaryDigitsUS)
+                    anyBinaryDigitsUS = true;
+                else if (base == 16 && !anyHexDigitsNoSingleUS)
+                    anyHexDigitsNoSingleUS = true;
                 ++p;
                 d = c - '0';
                 break;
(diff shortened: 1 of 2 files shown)


jacob-carlborg:
* Since druntime required changes this should be a deprecation
* Tests need to be added

test/compilable/fix19018.d
```diff
@@ -0,0 +1,22 @@
+/* REQUIRED_ARGS: -dw
```
wilzbach: FWIW we typically use `-de` here and put the test in fail_compilation.
The advantage is that once the deprecation is over, the test doesn't need to be moved, but there's no strict rule on this.

src/dmd/lexer.d
```diff
@@ -1955,6 +1969,11 @@ class Lexer : ErrorHandler
             error("integer overflow");
             err = true;
         }
+        if ((base == 2 && !anyBinaryDigitsUS) ||
+            (base == 16 && !anyHexDigitsNoSingleUS))
+        {
+            deprecation("Invalid integer literal is used, use valid one");
```
wilzbach: Maybe mention the name of the literal here?
E.g.

> `0x` isn't a valid integer literal

jacob-carlborg: Please include what literal to use to get the current behavior.

jacob-carlborg (changes requested):

src/dmd/lexer.d
```diff
@@ -1955,6 +1969,14 @@ class Lexer : ErrorHandler
             error("integer overflow");
             err = true;
         }
+        if ((base == 2 && !anyBinaryDigitsUS) ||
+            (base == 16 && !anyHexDigitsNoSingleUS))
+        {
+            auto tmp = new char[p-start+1];
+            tmp[0 .. p-start] = start[0 .. p-start];
+            tmp[$-1] = cast(char) null;
+            deprecation("%s isn't a valid integer literal", tmp.ptr);
```
JinShil: Code that is quoted in messages should be surrounded by tick marks.  In other words...

```
`%s` isn't a valid integer literal
```

JinShil (changes requested):

kubo39:
I changed commit message, should I change pull request title too?

jacob-carlborg:
That's not necessary.

JinShil (approved):

jacob-carlborg (approved):

kubo39:
Thanks for all reviews!

schveiguy:
Am I missing something or why is `0b_` allowed?

PR #5855, opened 2016-06-09
fix Issue 15831 - Mangle back references to avoid huge symbol names

rainers:
This is another implementation of a mangling scheme that uses back references for reoccuring types and identifiers. It also get's rid of an ambiguity in the existing mangling and encapsulates mangling better in dmangle.d.

see https://issues.dlang.org/show_bug.cgi?id=15831.

Updated:
The bad symbol `1.s.s.s.s.s.foo` mangles as `_D13testexpansion__T1sTSQw__TQjTSQBf__TQtTSQBp__TQBdTSQCa__TQBoTiZQBuFiZ6ResultZQCiFQBfZQqZQCtFQCbZQBbZQDfFQCxZQBnZQDrFQDsZQBz3fooMFNaNfZv` with 138 characters (instead of 1153). Please also note some extra characters due to the gratuitously long `__T` prefix.

The worse symbol `1.s.s.s.s.s.s.s.s.s.s.foo` grows to 253 characters (instead of 37640).

Here is a rough sketch of the implementation:
- for every type that has been mangled before in the same symbol, encode it as `QAAa`, where AAa is the relative position of the first occurrence encoded as a base 26 number with the last character using a lower case letter instead of upper case letters to detect the end of the number
- for every identifier that has been mangled before, encode it as `QAAa` aswell. To distinguish the two types of back references the demangler needs to lookup the character at the referenced position.

To capture all back references it is no longer allowed to build the mangled string from mangled strings of its sub components. It's also not feasible to insert characters into the mangled strings (currently done for the length of a mangled template instance) as it can invalidate back references.

Accompanying PRs: https://github.com/dlang/druntime/pull/1830 and https://github.com/dlang/dlang.org/pull/1664

PR #8398, opened 2018-06-24
Test pointing immutable!T to T types

ibuclaw:
Just seeing the result of it, as https://github.com/dlang/dmd/pull/5855/files#r197637374 introduced an AST bug (and this fixes the bug in question).


ibuclaw:
@WalterBright do you recall the reason why their is no link between `immutable(T)` and `T` in the AST?

ibuclaw:
https://issues.dlang.org/show_bug.cgi?id=19021

PR #8399, opened 2018-06-24
Fix Issue 19022 - CTorFlow: Show the line of the duplicated initialization for const/immutable fields

wilzbach:
I wasn't sure on the best way to save this information.
Is it okay to use an additional array here?

### DIFF
--- a/src/dmd/declaration.d
+++ b/src/dmd/declaration.d
@@ -12,6 +12,7 @@
 
 module dmd.declaration;
 
+import core.stdc.stdio;
 import dmd.aggregate;
 import dmd.arraytypes;
 import dmd.ctorflow;
@@ -112,7 +113,8 @@ bool modifyFieldVar(Loc loc, Scope* sc, VarDeclaration var, Expression e1)
                         break;
                 }
                 assert(i < dim);
-                const fi = sc.ctorflow.fieldinit[i];
+                auto fieldInit = &sc.ctorflow.fieldinit[i];
+                const fi = fieldInit.csx;
 
                 if (fi & CSX.this_ctor)
                 {
@@ -127,9 +129,14 @@ bool modifyFieldVar(Loc loc, Scope* sc, VarDeclaration var, Expression e1)
                         // dmd.ctorflow.CSX enum.
                         // @@@DEPRECATED_2019-01@@@.
                         if (fi & CSX.deprecate_18719)
-                            .deprecation(loc, "%s field `%s` initialized multiple times", modStr, var.toChars());
+                        {
+                            .deprecation(loc, "%s field `%s` was initialized in a previous constructor call", modStr, var.toChars());
+                        }
                         else
+                        {
                             .error(loc, "%s field `%s` initialized multiple times", modStr, var.toChars());
+                            .errorSupplemental(fieldInit.loc, "Previous initialization is here.");
+                        }
                     }
                 }
                 else if (sc.inLoop || (fi & CSX.label))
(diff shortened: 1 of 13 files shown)


src/dmd/root/rmem.d
```diff
@@ -256,4 +256,22 @@ extern (D) static char[] xarraydup(const(char)[] s) nothrow
...
 
+/**
+Makes a copy of the given array on newly allocated memory.
+
+Params:
+    s = array to copy
 
+Returns: A copy of the input array.
+*/
+extern (D) static auto arraydup(T)(T[] s) nothrow
```
JinShil: `const scope` maybe?

jacob-carlborg: * `static` is not necessary
* please don’t use `auto` for the return type
* please use standard D naming conventions

src/dmd/root/rmem.d
```diff
@@ -256,4 +256,22 @@ extern (D) static char[] xarraydup(const(char)[] s) nothrow
...
+{
+    if (s !is null)
```
jacob-carlborg: I suggest an early return.

jacob-carlborg (changes requested):

src/dmd/root/rmem.d
wilzbach: FYI: I copied the declaration of the previous `xarraydup`, hence the name.

> can this be made pure?

Yes, if the `pureMalloc` declarations are added. However, I don't see a big to do this in this PR as `xarraydup` isn't pure either, DMD can't take much advantage of purity anyhow and this PR still has other, real issues.

wilzbach:
Instead of using to separate arrays, I now merged them (CSX + Location info) into one.

JinShil (approved):

test/fail_compilation/fail9665a.d
```diff
@@ -1,15 +1,41 @@
...
-// immutable field
-
 /+
 TEST_OUTPUT:
 ---
-fail_compilation/fail9665a.d(19): Error: immutable field `v` initialized multiple times
+fail_compilation/fail9665a.d(45): Error: immutable field `v` initialized multiple times
+fail_compilation/fail9665a.d(44):        Previous initialization is here.
+fail_compilation/fail9665a.d(55): Error: immutable field `v` initialized multiple times
+       Previous initialization is here.
```
jacob-carlborg: This is missing file and line information as far as I can see.

jacob-carlborg (changes requested):

src/dmd/declaration.d
```diff
@@ -127,9 +129,14 @@ bool modifyFieldVar(Loc loc, Scope* sc, VarDeclaration var, Expression e1)
                         // dmd.ctorflow.CSX enum.
                         // @@@DEPRECATED_2019-01@@@.
                         if (fi & CSX.deprecate_18719)
-                            .deprecation(loc, "%s field `%s` initialized multiple times", modStr, var.toChars());
+                        {
+                            .deprecation(loc, "%s field `%s` was initialized in a previous constructor call", modStr, var.toChars());
```
wilzbach: The fix for 18719 is too aggressive and only checks whether a this call is present.
Hence, there's no way to get the line number of the field with the previous initialization (there might be none after all).
See also: https://issues.dlang.org/show_bug.cgi?id=19030

jacob-carlborg: Ok. Perhaps an intermediate solution can be implemented. For example, printing the line number of the constructor call if there is no line number for the field.

jacob-carlborg (approved):

src/dmd/declaration.d
wilzbach: The problem is the as of now we don't have this information. It currently triggers this deprecation whenever a `this(...)` call has been observed and a immutable field is modified. This means if there are three constructors it could be any of the two other.

PR #8400, opened 2018-06-24
Sanitize Json import path to a relative path

wilzbach:
Depending on the build process, the import paths might be absolute.
The sanitizer can be smart enough to handle this.


wilzbach:
This is finally passing.

jacob-carlborg (approved):

PR #8403, opened 2018-06-25
Fix buffer overflow in test/runnable/cppa.d

kinke:
This fixes reproducible `cppa` segfaults for LDC CI on 32/64-bit macOS.


Geod24:
Restarting SemaphoreCI...

Geod24 (approved):

PR #8404, opened 2018-06-25
Fix Issue 9701 - allow UDAs to be attached to enum values

IgorStepanov:
Reboot of #8319
@JinShil @jacob-carlborg Let finish it.

### DIFF
--- a/changelog/enum_attributes.dd
+++ b/changelog/enum_attributes.dd
@@ -1,0 +1,24 @@
+D now supports deprecated, @disable and user-defined attributes on enum members
+
+Example
+---
+template AliasSeq(TList...)
+{
+    alias AliasSeq = TList;
+}
+
+enum MyEnum
+{
+    @("uda0") value0,
+    @disable value1,
+    deprecated value2  // Deprecation: enum member `main.MyEnum.value2` is deprecated
+}
+
+static assert(__traits(getAttributes, MyEnum.value0) == AliasSeq!("uda0"));
+
+void main()
+{
+    auto v1 = MyEnum.value1;  // Error: enum member `main.MyEnum.value1` cannot be used because it is annotated with `@disable`
+}
+---
+
--- a/src/dmd/denum.d
+++ b/src/dmd/denum.d
@@ -14,6 +14,7 @@ module dmd.denum;
 
 import core.stdc.stdio;
 
+import dmd.attrib;
 import dmd.gluelayer;
 import dmd.declaration;
 import dmd.dscope;
@@ -382,6 +383,15 @@ extern (C++) final class EnumMember : VarDeclaration
         this.origType = origType;
     }
 
+    extern(D) this(Loc loc, Identifier id, Expression value, Type memtype,
+        StorageClass stc, UserAttributeDeclaration uad, DeprecatedDeclaration dd)
+    {
+        this(loc, id, value, memtype);
+        storage_class = stc;
+        userAttribDecl = uad;
+        depdecl = dd;
+    }
+
     override Dsymbol syntaxCopy(Dsymbol s)
     {
         assert(!s);
@@ -396,6 +406,9 @@ extern (C++) final class EnumMember : VarDeclaration
     Expression getVarExp(const ref Loc loc, Scope* sc)
     {
         dsymbolSemantic(this, sc);
+        if (errors)
+            return new ErrorExp();
+        checkDisabled(loc, sc);
         if (errors)
             return new ErrorExp();
         Expression e = new VarExp(loc, this);
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -4295,6 +4295,11 @@ extern (C++) final class SymOffExp : SymbolExp
  */
 extern (C++) final class VarExp : SymbolExp
 {
+    /**
+    * Semantic can be called multiple times for a single expression.
+    * This field is needed to ensure the deprecation message will be printed only once.
+    */
+    bool hasCheckedAttrs;
     extern (D) this(const ref Loc loc, Declaration var, bool hasOverloads = true)
     {
         if (var.isVarDeclaration())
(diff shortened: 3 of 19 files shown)


src/dmd/denum.d
```diff
@@ -374,6 +375,7 @@ extern (C++) final class EnumMember : VarDeclaration
     Type origType;
 
     EnumDeclaration ed;
+    bool isdeprecated;
```
wilzbach: FYI: We now have an official [D style](https://dlang.org/dstyle.html) which we try to follow for new variables.
So this would be `isDeprecated`

wilzbach (commented):
Wow! I hope that the third attempt works :+1:

src/dmd/denum.d
IgorStepanov: Base class Dsymbol already has a virtual method `isDeprecated()`. And some classes e.g. `AggregateDeclaration` has a field `isdeprecated`, which is returned by overridden method `isDeprecated()`.
Maybe stay it as is?

wilzbach: Hmm, I guess `_deprecated` or `_isDeprecated` or `isDeprecatedEnum` would be better than having a symbol which only differences in the spelling.
BTW `isdeprecated` is only set once and never looked at, which looks a bit suspicious.

src/dmd/expression.d
```diff
@@ -4295,6 +4295,7 @@ extern (C++) final class SymOffExp : SymbolExp
  */
 extern (C++) final class VarExp : SymbolExp
 {
+    bool hasCheckedAttrs;
```
JinShil: It's a little naughty, but I don't know of any other solution, and it'd be a shame to hold up this feature on such a minor problem, so OK, let's do it. I do, however, suggest adding a comment explaining why this exists.

IgorStepanov: Done

src/dmd/denum.d
IgorStepanov: Ok, I've just removed it:)

IgorStepanov:
BTW, What purpose of `astbase` module and classes. Looks like `astbase.Dsymbol.deprecatedDeclaration` hasn't used anywhere. Maybe we should skip `dd` parameter of constructor?

JinShil:
> BTW, What purpose of astbase module and classes. Looks like astbase.Dsymbol.deprecatedDeclaration hasn't used anywhere. Maybe we should skip dd parameter of constructor?

Keep in mind that I had been playing with that code for a very long time, trying many different things to work through the problem.  If you find that the `DeprecatedDeclaration` is not needed, then just remove it.

wilzbach:
AstBase is used nowhere in DMD. It was an attempt by @RazvanN7 to separate the AST nodes and create a pure "data" layer.
However, the switch to AstBase has never been made though I don't recall the exact reasons (I think it had sth. to do with the proposal needing too many templates).
See also:
- https://github.com/dlang/dmd/pull/6966
- https://github.com/dlang/dmd/pull/6683

IgorStepanov:
>Keep in mind that I had been playing with that code for a very long time, trying many different things to work through the problem. If you find that the DeprecatedDeclaration is not needed, then just remove it.

`astbase.EnumMember.__ctor` should have the same signature with `denum.EnumMember.__ctor` because they both are used in `Parser(AST)`.
I may just ignore `dd` parameter in `astbase.EnumMember.__ctor`, but in can be needed in `astbase.EnumMembe` in future.

JinShil:
> but in can be needed in astbase.EnumMembe in future.

I suggest adhering to the [YAGNI principle](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

IgorStepanov:
@JinShil 
>remove it from both constructors.

It needed in `denum.EnumMember.__ctor` and isn't needed in `astbase.EnumMember.__ctor`.
Thus it should be in constructor but may be ignored in `astbase.EnumMember`.

IgorStepanov:
>may be ignored in `astbase.EnumMember`

Done

src/dmd/expression.d
```diff
@@ -4295,6 +4295,9 @@ extern (C++) final class SymOffExp : SymbolExp
  */
 extern (C++) final class VarExp : SymbolExp
 {
+    // Semantic can be called many times for a one expression.
+    // This field need to ensure that deprecation message will be printed only one time.
+    bool hasCheckedAttrs;
```
JinShil: Changes for proper English grammar:
```
// Semantic can be called multiple times for a single expression.
// This field is needed to ensure the deprecation message will be printed only once.
```

JinShil (changes requested):

JinShil:
It looks good to me.  I would like to hear from @RazvanN7 about the discrepancy with `ASTBase`.

src/dmd/expression.d
```diff
@@ -4295,6 +4295,9 @@ extern (C++) final class SymOffExp : SymbolExp
  */
 extern (C++) final class VarExp : SymbolExp
 {
+    // Semantic can be called many times for a one expression.
```
jacob-carlborg: Please make this a Ddoc comment.

jacob-carlborg (approved):
Looks good, just a few minor comments.

src/dmd/expression.d
IgorStepanov: Fixed

src/dmd/expression.d
IgorStepanov: Done

changelog/enum_attributes.dd
```diff
@@ -0,0 +1,24 @@
+D now supports attributes on enum members
+
```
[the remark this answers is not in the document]
IgorStepanov: Fixed

IgorStepanov:
@JinShil @jacob-carlborg All comments resolved. Any other objections?

jacob-carlborg:
> It looks good to me. I would like to hear from @RazvanN7 about the discrepancy with ASTBase.

As far as I understand, ASTBase was introduced to allow using the parser without the rest of the frontend. Using the rest of the frontend as a library caused some problems. Those are fixed and the full frontend is available as a library. I don't think there's a need for ASTBase.

jacob-carlborg (approved):
I'll approve this and if there's a need to update ASTBase we can do that in a followup PR.

IgorStepanov:
Thanks!

PR #8405, opened 2018-06-26
[Experiment] Run the testsuite in parallel

wilzbach:
auto-tester would be a lot faster if a higher parallelism is used.
I'm just interested in what would break.


wilzbach:
Closing due to https://github.com/dlang/phobos/pull/6612#issuecomment-400148566

PR #8406, opened 2018-06-26
Deprecate opDot

wilzbach:
I just found this [very old post about `opDot`](https://stackoverflow.com/questions/9880064/d2-what-are-semantics-of-opdot) from 2012.
The gist:

> opDot has been scheduled for deprecation. That's why it isn't documented. Don't use it. Use alias this instead.

I think it's more than time to actually trigger the deprecation!

### DIFF
--- a/changelog/deprecate-opDot.dd
+++ b/changelog/deprecate-opDot.dd
@@ -1,0 +1,51 @@
+Deprecatie usage of `opDot`
+
+`opDot` was the D1 analog to `alias this`.
+However, `alias this` covers all use cases of `opDot`, but ensures safety.
+
+---
+struct S
+{
+    int a, b;
+}
+struct T
+{
+    S s;
+
+    S* opDot()
+    {
+        return &s;
+    }
+}
+
+void main()
+{
+    T t;
+    t.a = 4;
+    assert(t.a == 4);
+    t.b = 5;
+}
+---
+
+With `alias this`:
+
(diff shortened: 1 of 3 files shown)


Geod24:
There's a slight semantic difference between them though: `alias this` does implicit conversion, `opDot` does not. Besides function call, that can make a difference in template specialization.

I'm not sure that's something worth worrying about though, as `opDispatch` covers that case and much more.

It LGTM and I'd merge it, but since it's a feature deprecation which wasn't talked about for a while, let's ask for @WalterBright and/or @andralex for a green light. I don't think there's any rush here.

jacob-carlborg (approved):

WalterBright (approved):

rainers:
I have the impression that deprecation of opDot comes too early: alias this is not a good replacement because it is still too buggy in combination with operator overloads. I've been struggling for hours with replacing opDot here: https://github.com/rainers/visuald/commit/ead3f510b207ae909d74c68f2cd75a160d6cd114, but the `alias this` version still causes crashes. Operators other than opEquals also fail to work as expected, e.g. ++ is always forwarded to the aliased symbol rather than using the explicit opAddAssign overload.

Mimicking opDot with opDispatch kind of works, but is pretty tough to get right, is ugly and has the usual disadvantages of templates, e.g. no implicit conversions for function arguments.

PR #8407, opened 2018-06-26
adjust maybescope when turning off inferring

WalterBright:
`STC.maybescope` should never be on when `doNotInferScope` is on.


Geod24:
Do you have a test case of a bug it might trigger ?

PR #8408, opened 2018-06-27
fix Issue 19035 - Escape in scope inference, improve scope inference

WalterBright:
I.e. when two or more parameters are mutually dependent on each other for `scope` inference, this detects more cases.

### DIFF
--- a/src/dmd/declaration.d
+++ b/src/dmd/declaration.d
@@ -1088,6 +1088,8 @@ extern (C++) class VarDeclaration : Declaration
     Expression edtor;               // if !=null, does the destruction of the variable
     IntRange* range;                // if !=null, the variable is known to be within the range
 
+    VarDeclarations* maybes;        // STC.maybescope variables that are assigned to this STC.maybescope variable
+
     final extern (D) this(Loc loc, Type type, Identifier id, Initializer _init, StorageClass storage_class = STC.undefined_)
     {
         super(id);
@@ -1631,6 +1633,23 @@ extern (C++) class VarDeclaration : Declaration
     final bool enclosesLifetimeOf(VarDeclaration v) const pure
     {
         return sequenceNumber < v.sequenceNumber;
+    }
+
+    /***************************************
+     * Add variable to maybes[].
+     * When a maybescope variable `v` is assigned to a maybescope variable `this`,
+     * we cannot determine if `this` is actually scope until the semantic
+     * analysis for the function is completed. Thus, we save the data
+     * until then.
+     * Params:
+     *  v = an STC.maybescope variable that was assigned to `this`
+     */
+    final void addMaybe(VarDeclaration v)
+    {
+        //printf("add %s to %s's list of dependencies\n", v.toChars(), toChars());
+        if (!maybes)
+            maybes = new VarDeclarations();
+        maybes.push(v);
     }
 }
 
--- a/src/dmd/escape.d
+++ b/src/dmd/escape.d
@@ -96,7 +96,7 @@ bool checkAssocArrayLiteralEscape(Scope *sc, AssocArrayLiteralExp ae, bool gag)
 bool checkParamArgumentEscape(Scope* sc, FuncDeclaration fdc, Identifier par, Expression arg, bool gag)
 {
     enum log = false;
-    if (log) printf("checkParamArgumentEscape(arg: %s par: %s)\n", arg.toChars(), par.toChars());
+    if (log) printf("checkParamArgumentEscape(arg: %s par: %s)\n", arg ? arg.toChars() : "null", par ? par.toChars() : "null");
     //printf("type = %s, %d\n", arg.type.toChars(), arg.type.hasPointers());
 
     if (!arg.type.hasPointers())
@@ -153,14 +153,15 @@ bool checkParamArgumentEscape(Scope* sc, FuncDeclaration fdc, Identifier par, Ex
             /* v is not 'scope', and is assigned to a parameter that may escape.
              * Therefore, v can never be 'scope'.
              */
-            if (log) printf("no infer for %s in %s, fdc %s, %d\n",
-                v.toChars(), sc.func.ident.toChars(), fdc.ident.toChars(),  __LINE__);
+            if (log) printf("no infer for %s in %s loc %s, fdc %s, %d\n",
+                v.toChars(), sc.func.ident.toChars(), sc.func.loc.toChars(), fdc.ident.toChars(),  __LINE__);
             v.doNotInferScope = true;
         }
     }
 
     foreach (VarDeclaration v; er.byref)
     {
+        if (log) printf("byref %s\n", v.toChars());
         if (v.isDataseg())
             continue;
 
@@ -283,6 +284,11 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
     bool inferScope = false;
     if (va && sc.func && sc.func.type && sc.func.type.ty == Tfunction)
         inferScope = (cast(TypeFunction)sc.func.type).trust != TRUST.system;
+    //printf("inferScope = %d, %d\n", inferScope, (va.storage_class & STCmaybescope) != 0);
+
+    // Determine if va is a parameter that is an indirect reference
+    const bool vaIsRef = va && va.storage_class & STC.parameter &&
+        (va.storage_class & (STC.ref_ | STC.out_) || va.type.toBasetype().ty == Tclass);
 
     bool result = false;
     foreach (VarDeclaration v; er.byvalue)
@@ -296,7 +302,17 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
 
         Dsymbol p = v.toParent2();
 
-        if (!(va && va.isScope()))
+        if (va && !vaIsRef && !va.isScope() && !v.isScope() &&
+            (va.storage_class & v.storage_class & (STC.maybescope | STC.variadic)) == STC.maybescope &&
+            p == sc.func)
+        {
+            /* Add v to va's list of dependencies
+             */
+            va.addMaybe(v);
+            continue;
+        }
+
+        if (!(va && va.isScope()) || vaIsRef)
             notMaybeScope(v);
 
         if (v.isScope())
@@ -314,8 +330,9 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
             if (va &&
                 (va.enclosesLifetimeOf(v) && !(v.storage_class & (STC.parameter | STC.temp)) ||
                  // va is class reference
-                 ae.e1.op == TOK.dotVariable && va.type.toBasetype().ty == Tclass && (va.enclosesLifetimeOf(v) || !va.isScope) ||
-                 va.storage_class & STC.ref_ && !(v.storage_class & STC.temp)) &&
+                 ae.e1.op == TOK.dotVariable && va.type.toBasetype().ty == Tclass && (va.enclosesLifetimeOf(v) || !va.isScope()) ||
+                 vaIsRef ||
+                 va.storage_class & (STC.ref_ | STC.out_) && !(v.storage_class & STC.temp)) &&
                 sc.func.setUnsafe())
             {
                 if (!gag)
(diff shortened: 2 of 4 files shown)


jacob-carlborg:
> don't pull needs work

Give us a ping when it's ready for review.

WalterBright:
I'm surprised this slipped through all the tests! I guess it is good to go. Ping @jacob-carlborg

test/fail_compilation/retscope6.d
```diff
@@ -21,3 +21,55 @@ int* test() @safe
...
     return arr[0][0];
 }
+
+/* TEST_OUTPUT:
+---
+fail_compilation/retscope6.d(7034): Error: reference to local variable `i` assigned to non-scope parameter `_param_1` calling retscope6.S.emplace!(int*).emplace
+fail_compilation/retscope6.d(7035): Error: reference to local variable `i` assigned to non-scope parameter `_param_0` calling retscope6.S.emplace2!(int*).emplace2
+fail_compilation/retscope6.d(7024): Error: scope variable `_param_2` assigned to `s` with longer lifetime
+fail_compilation/retscope6.d(7025): Error: scope variable `_param_2` assigned to `t` with longer lifetime
+fail_compilation/retscope6.d(7037): Error: template instance `retscope6.S.emplace4!(int*)` error instantiating
```
jacob-carlborg: I recommend a single section of test output, it will be much easier if the error messages need to be updated. It caused me quite some pain in another PR.

WalterBright: I do multiple blocks with a `#line` precisely because they are easier to update. Adding a line doesn't require updating every message line.

wilzbach: `./run.d fail_compilation AUTO_UPDATE=1` for the win...

src/dmd/declaration.d
```diff
@@ -1088,6 +1088,8 @@ extern (C++) class VarDeclaration : Declaration
     Expression edtor;               // if !=null, does the destruction of the variable
     IntRange* range;                // if !=null, the variable is known to be within the range
 
+    VarDeclarations* maybes;        // STC.maybescope variables that are assigned to this STC.maybescope variable
```
jacob-carlborg: * Can this be a D array?
* The comment should be a Ddoc comment
* Can we call this `maybeScopes`, `possibleScopeVariables` or some variations of those?

src/dmd/escape.d
```diff
@@ -1580,3 +1597,51 @@ else
...
 
+
+/**********************************************
+ * Have some variables that are maybescopes that were
+ * assigned values from other maybescope variables.
+ * Now that semantic analysis of the function is
+ * complete, we can finalize this by turning off
+ * maybescope for array elements that cannot be scope.
+ *
+ *  va      v      =>  va     v
```
jacob-carlborg: Are `va` and `v` referring to the foreach variables in the source below? If that’s the case please quote them with backticks to make that more clear.

jacob-carlborg (changes requested):

test/fail_compilation/retscope6.d
jacob-carlborg: Yeah, this only works with a single output section.

src/dmd/escape.d
```diff
@@ -283,6 +284,11 @@ bool checkAssignEscape(Scope* sc, Expression e, bool gag)
     bool inferScope = false;
     if (va && sc.func && sc.func.type && sc.func.type.ty == Tfunction)
         inferScope = (cast(TypeFunction)sc.func.type).trust != TRUST.system;
+    //printf("inferScope = %d, %d\n", inferScope, (va.storage_class & STCmaybescope) != 0);
+
+    // Determine if va is a parameter that is an indirect reference
+    bool vaIsRef = va && va.storage_class & STC.parameter &&
```
jacob-carlborg: Can be made `const`.

jacob-carlborg (dismissed):

src/dmd/declaration.d
WalterBright: 1. No, as it's variable sized, and we generally avoid D arrays in the front end for that reason.
2. As the rest or not, such a change should await that
3. I'm not a big fan of long names

src/dmd/escape.d
WalterBright: ok

src/dmd/escape.d
WalterBright: ok

JinShil:
Am I the only one drowning in DIP1000 sea?  Can we please document all of this first, and *then* move to implementation?  Or at the very least, provide some formal documentation with these PRs?

This PR could also use an elaborate problem statement.  There's no bugzilla issue, so I don't understand what the problem is that this solves.  How are we supposed to be able to predict the behavior of the compiler with all this inference magic?  We'll be debugging our understanding of the language before we ever get to debugging our code.  (cue Scott Meyers "The last thing D needs")

JinShil:
> I thought the test case was pretty clear.

Sorry, but it's not for me.  Perhaps you could augment it with some helpful comments.

test/fail_compilation/retscope6.d
```diff
@@ -21,3 +21,55 @@ int* test() @safe
...
+
+    static void emplace3(Args...)(S s, Args args) @safe
+    {
+        s.payload = args[0];
+    }
+
+    static void emplace4(Args...)(scope ref S s, scope out S t, scope Args args) @safe
+    {
+        s.payload = args[0];
+        t.payload = args[0];
```
JinShil: Due to the lack of documentation surrounding the semantics of `scope` I'm assuming that in this context, `scope` means that `arg[0]`'s lifetime must not exceed the lifetime of this `emplace4`'s scope.  However, since we are assigning to a `ref` and `out` we are escaping out of `emplace4`'s scope.  Is this correct?

When I analyze the implementation of `foo`, however, there's not escape because everything is local to `foo`.  Perhaps this could use a more compelling test that would actually result in a leak or memory corruption without this PR.

WalterBright: I'll add some more documentation.

src/dmd/declaration.d
jacob-carlborg: > I'm not a big fan of long names

I'm not a fan of names that are not understandable. `maybes`, "maybe" what?

jacob-carlborg: > No, as it's variable sized, and we generally avoid D arrays in the front end for that reason.

There are several cases where D arrays are used [1] [2] and you're usually arguing for D string instead of C strings, which are D arrays.

[1] https://github.com/dlang/dmd/blob/master/src/dmd/dclass.d#L56
[2] https://github.com/dlang/dmd/blob/master/src/dmd/dclass.d#L189

test/fail_compilation/retscope6.d
JinShil: @WalterBright Is it really too much to ask to have you open a dlang.org PR to document the semantics of `scope` and `return` function parameters with illustrative examples so we can refer to it when we review these?  Or are you still trying to figure all this out, and this is your way of working through it?

test/fail_compilation/retscope6.d
```diff
@@ -21,3 +21,120 @@ int* test() @safe
...
+void testfrankly()
+{
+    int* p;
+    int i;
+    frank(p, &i);
+}
+
+/**********************/
+
+void betty()(int* p, int* q)
```
JinShil: It's interesting that you chose to use a template here?  Why is that?  Is it to ensure the body of the function is needed in order to do the inference.  For example, I noticed that Example 1 does not compile, but Example 2 does.

Example 1
```D
@safe:
void betty(int* p, int* q)  // Notice this is not a template
{
     p = q;
}

void testbetty()
{
    int i;
    int j;
    betty(&i, &j); // Error: reference to local variable i assigned to non-scope parameter p calling betty
                   // Error: reference to local variable j assigned to non-scope parameter q calling betty
}
```

Example 2
```D
@safe:
void betty()(int* p, int* q)  // Notice this is a template
{
     p = q;
}

void testbetty()
{
    int i;
    int j;
    betty(&i, &j);
}
```

There's a lot you're not saying here, and I think you're taking a lot for granted expecting us to review the semantics of this PR.

test/fail_compilation/retscope6.d
WalterBright: It's documented in DIP25 and DIP1000. I'd start with a gentler introduction, http://dconf.org/2017/talks/bright.html

test/fail_compilation/retscope6.d
```diff
@@ -21,3 +21,120 @@ int* test() @safe
...
+    {
+        payload = args[0];
+    }
+
+    static void emplace3(Args...)(S s, Args args) @safe
+    {
+        s.payload = args[0];
+    }
+
+    static void emplace4(Args...)(scope ref S s, scope out S t, scope Args args) @safe
```
JinShil: https://dlang.org/spec/attribute.html#scope

> scope cannot be applied to globals, statics, data members, ref or out parameters.

This is violating the spec.  The spec needs an update.

JinShil: https://dlang.org/spec/attribute.html#scope

> Assignment to a scope, other than initialization, is not allowed.

Also not accurate.

WalterBright: As I've mentioned before, dip1000 has not been folded into the spec.

JinShil: Yes, and that's part of the problem.  Why hasn't it been done? And shouldn't it be done either prior to PRs like this or together with PRs like this?

WalterBright: > Why hasn't it been done?

One large reason is because I can't get my spec fixes pulled.

Aside from that, DIP1000 despite being years old has little use, so I don't know how well it works. I can't get people to use DIP1000 features because phobos doesn't compile with dip1000. So I try to get phobos to compile with dip1000. I run into problems. I make corrections (like this PR). And I get blocked.

End result, no progress anywhere.
---
There is another issue. I used to work at Boeing in design. The designs ended up as drawings, using pen and ink on special plastic paper. Unsurprisingly, revising these drawings (including redrawing them) is quite expensive. But the designs changed constantly (bug fixes, improvements, changing requirements, the same issues that affect software). What was done was issuing DCN (Drawing Change Notices), which contained the fixes to the drawing. The official "drawing" was the original drawing plus a stack of DCNs, applied in chronological order (much like git, git is hardly a new concept). Eventually, the drawing may get redone with all the DCNs folded in.

The analogy for us is the spec, plus approved DIPs, plus Bugzilla Enhancement Requests and bug fixes. It's not ideal, but it's pragmatic. Halting all progress because it all hasn't been folded into the spec yet is not pragmatic, especially when my PRs to update the spec just sit there.

JinShil: > One large reason is because I can't get my spec fixes pulled.

I suggest putting a time limit on those PRs that haven't received feedback.  We already have a `72h no objection -> merge" label in the DMD repository, and I don't see why that can't be applied to your dlang.org PRs.  Noone has a right to complain if they chose not to participate.

> There is another issue. I used to work at Boeing in design...

I understand that, but I'm not going to be your rubber stamp.  I need to understand what these PRs are doing and why.

Others on this team are well within their charter to move this PR forward if they understand it and have confidence in the direction you're heading, but I'm not on board yet and have concerns.  I've already sent you an e-mail or two about the way `scope` is being inferred, and I don't like what I'm seeing in the implementation so far.  I think we need a way to set storage classes meta-programmatically, rather than reaching for inference.

WalterBright:
> I need to understand what these PRs are doing and why.

I won't argue that it is simple. But what do you suggest we do with a 2 person rule when nobody else understands the compiler?

jacob-carlborg:
> I won't argue that it is simple. But what do you suggest we do with a 2 person rule when nobody else understands the compiler?

We need to incrementally try to get more people to understand the compiler.

JinShil:
> But what do you suggest we do with a 2 person rule when nobody else understands the compiler?

IMO, if you are determined, highly confident in your solution, and you don't receive any feedback or don't receive feedback that you consider significant to change your position, then after a reasonable amount of time, you're in a unique position to just merge away.

But, I'd prefer if you did more to bring others along with you.  With the lack of documentation about these inference rules (formal or informal) I don't think you're doing enough to justify this PR to the team.  Though it could just be me.

This is an important feature, and I don't recommend anyone hold up this PR based on my objections alone.  I won't protest if my objections are overruled, and may even merge this myself if I can be convinced that it's the right solution.

wilzbach (approved):
I'm going to merge this now on the understanding that DIP1000 is still under work/ in flux and it needs to move forward (and this is indeed a nice improvement).
However, I agree with @JinShil that if we want -dip1000 to be a success there needs to be a better plan and roadmap.

PR #8410, opened 2018-06-27
Deprecate scope as a type constraint

JinShil:
Followup to https://github.com/dlang/dlang.org/pull/2027

See https://dlang.org/deprecate.html#scope%20as%20a%20type%20constraint

Update to the deprecations table: https://github.com/dlang/dlang.org/pull/2397

### DIFF
--- a/changelog/deprecate_scope_type.dd
+++ b/changelog/deprecate_scope_type.dd
@@ -1,0 +1,8 @@
+`scope` as a type constraint is deprecated.
+
+`scope` as a type constraint has been deprecated for quite some time.  However,
+starting with this release, the compiler will emit a deprecation warning if used.
+
+---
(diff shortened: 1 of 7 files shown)


wilzbach (approved):

changelog/deprecate_scope_type.dd
```diff
@@ -0,0 +1,8 @@
+`scope` as a type constraint is deprecated.
+
+`scope` as a type constraint has been deprecated for quite some time.  However,
+starting with this release, the compiler will emit a deprecation warning if used.
```
jacob-carlborg: Should we mention that there is not direct alternative?

JinShil:
This is blocked waiting on the outcome of @WalterBright use of `scope struct` in https://github.com/dlang/dmd/pull/8369#discussion_r198675586

PR #8411, opened 2018-06-27
BUILD=release should set ENABLE_RELEASE

John-Colvin:


wilzbach (changes requested):
Currently we have three states:
- release (default) without the actual optimizations, but with debug symbols
- debug
- release via ENABLE_RELEASE

This is mostly due to the historic layout and there actually has been a failed attempt once to set the default build to debug. Enabling all the optimizations by default is imho not a good idea as the default build time will go up from a few seconds to two minutes and will have severe implications for the out of the box build experience (and CI setup).

WalterBright:
Wouldn't it be better to just remove `BUIILD=release`, instead of trying to maintain two different ways to build the same thing? What are we trying to be backwards compatible with?

wilzbach:
> Wouldn't it be better to just remove BUIILD=release, instead of trying to maintain two different ways to build the same thing?

FWIW `BUILD=release` and `ENABLE_RELEASE=1` aren't exactly the same.
For example, with `BUILD=debug` a _different_ binary gets generated in `generated/linux/debug/...`.
This is often handy if you want to simultaneously have a debug and release build of dmd/druntime/phobos.
`ENABLE_X` only adds the respective flags, but doesn't change the build folder.
Also the actual release flags only get added with `ENABLE_RELEASE=1` as we have these three state (default [aka BUILD=release], BUILD=debug, ENABLE_RELEASE=1).

> What are we trying to be backwards compatible with?

Our CIs - for your removing `BUILD=release` we probably would have to update all CIs here, at druntime, phobos, tools, dlang.org, installer and the release pipeline.
Updating them is a PITA. If we could do this easily, we probably would have switched to `debug` being the default build already.

AndrewEdwards:
@WalterBright, this pull request is pending you input.

wilzbach:
> It probably won't be trivial to look after / fix our CIs

This is the blocker.

> but I'm happy to do so if there's approval.

Another blocker - I don't have much time these days :/

RazvanN7:
This doesn't seem to be going anywhere.

PR #8412, opened 2018-06-27
Changed 'body' to 'do' x2

GabyForceQ:
I found two 'body' keywords in dmd/cond.d. I replaced them with 'do'.

### DIFF
--- a/src/dmd/cond.d
+++ b/src/dmd/cond.d
@@ -106,7 +106,7 @@ extern (C++) final class StaticForeach : RootObject
     {
         assert(!!aggrfe ^ !!rangefe);
     }
-    body
+    do
     {
         this.loc = loc;
         this.aggrfe = aggrfe;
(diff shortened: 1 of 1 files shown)


wilzbach (dismissed):
I'm sorry but as long as we have bootstrapping compilers as old as 2.068 I'm afraid we can't apply this.

GabyForceQ:
You can merge this when 'body' is finally is removed.

wilzbach:
Well, we can do lots of cool upgrades if all host compilers on all CIs have been upgraded, but for now dmd still needs to built with GDC 4.8.5 which is based on the 2.068.2 frontend.
I really appreciate your contribution, this is sadly something we can't do for now & if we ever can do, it will be easy to recreate. In fac as DMD is built with `-de`, if we ever start to deprecate `body`, this change will be required anyways.
I'm going to close this now as we try to keep the PR queue as slim as possible and this would probably linger around for another year.

If you are searching for other ideas to contribute to DMD, you might want to have a look at https://github.com/dlang/dmd/projects or talk with us humans in #d IRC or #dmd on Slack.

wilzbach (approved):

PR #8413, opened 2018-06-28
Fix issue 19038: when performing an op on two nested arrays that differ only in constness, try to make both const.

FeepingCreature:
Is `string[][]` implicitly convertible to `const(string)[][]`? `TypeDArray::implicitConvTo` calls `<string[]>.constConv(<const(string)[]>)`, which checks if `<string>.equals(<const(string)>)`; seeing that it doesn't, it returns `nomatch`.

Why not, at least in the simple case where constness is the same as the comparison type, just recurse instead?

Let's see what the tests say.

### DIFF
--- a/src/dmd/dcast.d
+++ b/src/dmd/dcast.d
@@ -3302,6 +3302,21 @@ LmodCompare:
             e2 = tmp;
         }
     }
+    // constness1(T)[][] op constness2(T)[][]: maybe we can make them both const?
+    else if (t1.ty == Tarray && t2.ty == Tarray // two dynamic arrays
+        && t1.constOf().equals(t2.constOf()) // and they're unifiable as const
+        && (!t1.equals(t1.constOf()) || !t2.equals(t2.constOf())) // but they're not both const yet
+        && e1.implicitConvTo(t1.constOf()) >= MATCH.constant // but we can make them const
+        && e2.implicitConvTo(t2.constOf()) >= MATCH.constant)
+    {
+        e1 = e1.castTo(sc, t1.constOf());
+        e2 = e2.castTo(sc, t2.constOf());
+        t1 = e1.type;
+        t2 = e2.type;
+        t = t1;
+
+        goto Lagain;
+    }
     else
     {
     Lincompatible:
--- a/test/compilable/test19038.d
+++ b/test/compilable/test19038.d
@@ -1,0 +1,18 @@
+module test19038;
+
+void test()
+{
+    string mutableStr = "x";
+    const string constStr = mutableStr;
+
+    if ([[constStr]] == [[mutableStr]])
+    {
+    }
+
+    int mutableInt = 5;
+    const int constInt = mutableInt;
+
+    if ([[constInt]] == [[mutableInt]])
+    {
+    }
+}


FeepingCreature:
Alright, clearly something broke. I'll take a look tomorrow.

yshui:
@FeepingCreature string[][] is not implicitly converible to const(string)[][], and there is good reason why:

https://stackoverflow.com/questions/5055655/double-pointer-const-correctness-warnings-in-c

yshui:
@FeepingCreature Translate that to D:

```d
string[][] a = ...;
const(string)[][] b = a; // assume this works
const(string)[] c = ...;
c[0] = ""; // invalid, because c[0] is const(string)

b[0] = c; // valid, b[0] is const(string)[]
// But now, a[0] has become c
a[0][0] = ""; // valid, because a[0][0] is string
// But this also changes c[0], which shouldn't be possible
```

yshui:
@FeepingCreature In D, you cannot change variable through a const reference.

Suppose you have a function

```d
void foo(const(string)[] c);
```

What ever `foo` does, content of `c[0]` should not change.

But if `T[][]` can be implicitly convert to `const(T)[][]`, by doing the dance I pointed out, `foo` will be able to change `c[0]`, which is not acceptable.

FeepingCreature:
You cannot change a value through a const reference.

But you can absolutely have a const reference to a _value that changes_.

FeepingCreature:
On the other hand, both should be implicitly convertible to const(string[])[].

FeepingCreature:
Second try!

FeepingCreature:
Ah crud. That's weird.

FeepingCreature:
oops, that was silly. fixed?

test/compilable/test19038.d
```diff
@@ -0,0 +1,11 @@
+module test19038;
+
+void test()
+{
+    if ([[cast(const) "x"]] == [["x"]])
+    {
+    }
+    if ([[cast(const) 5]] == [[5]])
+    {
+    }
```
Geod24: It would be better to type out variables instead of casting literals.
```D
const a = "a";
```
Will do it

aG0aep6G:
> oops, that was silly. fixed?

Yeah. The inferred common type is now `const(int[][])`. Ideally, it would be `const(int[])[]`, I think, but I don't know if that's harder to implement. Any way, `const(int[][])` is much better than an error.

Another thing that may be beyond the scope of this PR: The ternary test still fails when `immutable` is involved:

```d
void main()
{
    int[][] m;
    const(int)[][] c;
    immutable(int)[][] i;
    
    const(int[])[] common;
    common = true ? m : c; /* Now accepted. Cool. */
    common = true ? m : i; /* Still rejected. */
    common = true ? c : i; /* Still rejected. */
}
```

src/dmd/dcast.d
```diff
@@ -3302,6 +3302,21 @@ LmodCompare:
             e2 = tmp;
         }
     }
+    // constness1(T)[][] op constness2(T)[][]: maybe we can make them both const?
+    else if (t1.ty == Tarray && t2.ty == Tarray // two dynamic arrays
+        && t1.constOf().equals(t2.constOf()) // and they're unifiable as const
+        && (!t1.equals(t1.constOf()) || !t2.equals(t2.constOf())) // but they're not both const yet
+        && e1.implicitConvTo(t1.constOf()) >= MATCH.constant // but we can make them const
+        && e2.implicitConvTo(t2.constOf()) >= MATCH.constant)
```
Geod24: That looks like an extremely specific check. What about other kind of indirections ? (ptr, `ref`)
```D
    const ca = "a".ptr;
    if ([[ca]] == [["a".ptr]]) {}
```

Does this compile after your change ? Should it ?

Geod24 (changes requested):
This looks way too specific to fix the underlying issue.

Side note: When opening a PR, please put the issue title in the PR in addition to the bugzilla entry. It makes triage way easier.

src/dmd/dcast.d
FeepingCreature: Looking at the rest of the function, admittedly this check is very specific but it's not all that much more specific than the other ten checks already chained in there.

I wanted to specifically nail this one case, because I am still very unconfident about the general case.

The pasted code compiles, as it should. Is there any reason it shouldn't?

test/compilable/test19038.d
FeepingCreature: fixed

WalterBright:
> But if T[][] can be implicitly convert to const(T)[][], by doing the little dance I pointed out, foo will be able to change c[0], which is not acceptable.

@yshui I'd like to see a test case for this added to the test suite, to ensure we don't make this mistake in the future. Thanks for pointing it out!

WalterBright:
What the heck, I added it myself.

FeepingCreature:
Closing cause I've lost track of what even motivated this.

Geod24:
I think you were seeing [issue 20938](https://issues.dlang.org/show_bug.cgi?id=20938) which was fixed in v2.094.0.

FeepingCreature:
Nah, `Error: incompatible types for ([["a"]]) == ([["a"]]): const(string)[][] and string[][]` still happens in master. But I haven't run into this bug in the wild in years, so eh.

PR #7528, opened 2017-12-26
Run CircleCi with BUILD=debug

wilzbach:
From @jacob-carlborg on Slack:

> I’ve noticed that when compiling DMD in debug mode and running the test the following assertion fails: https://github.com/dlang/dmd/blob/master/src/dmd/dmangle.d#L587
> the assertion fails due to one of characters in the mangled name contains a `-`. Is it the assertion that needs to be updated or is the mangling wrong?
> since we’re not testing the compiler in debug mode, as far as I understand, and everything seems to be working I’m guessing the assertion needs to be updated
> feels slightly pointless to me with the debug code if it’s never run

This just opens the PR with `BUILD=debug`. We still need to fix the error in debug mode, but this can be done independently.

PR #8418, opened 2018-06-29
Add different ENABLE + unittest modes to build.d

wilzbach:
Moving more bits from the `posix.mak` to `build.d`.
Depends on https://github.com/dlang/dmd/pull/8417.

### DIFF
--- a/.circleci/run.sh
+++ b/.circleci/run.sh
@@ -113,9 +113,14 @@ coverage()
     if [ -f ~/dlang/install.sh ] ; then
         source "$(CURL_USER_AGENT=\"$CURL_USER_AGENT\" bash ~/dlang/install.sh dmd-$HOST_DMD_VER --activate)"
     fi
+    RDMD="$(type -p rdmd)"
 
     # build dmd, druntime, and phobos
-    make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=$DMD BUILD=$BUILD ENABLE_WARNINGS=1 PIC="$PIC" all
+    if [ "$MODEL" == "64" ] ; then
+        "$RDMD" ./src/build.d MODEL=$MODEL HOST_DMD=$DMD BUILD=$BUILD ENABLE_WARNINGS=1 PIC="$PIC" all
+    else
+        make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=$DMD BUILD=$BUILD ENABLE_WARNINGS=1 PIC="$PIC" all
+    fi
     make -j$N -C ../druntime -f posix.mak MODEL=$MODEL PIC="$PIC"
     make -j$N -C ../phobos -f posix.mak MODEL=$MODEL PIC="$PIC"
 
--- a/src/build.d
+++ b/src/build.d
@@ -6,7 +6,6 @@ Usage:
   ./build.d dmd
 
 TODO:
-- add different ENABLE modes
 - add all posix.mak Makefile targets
 - support 32-bit builds
 - test on OSX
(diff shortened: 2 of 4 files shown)


src/dmd/asttypename.d
```diff
@@ -112,6 +112,7 @@ public :
 unittest
 {
     import dmd.globals : Loc;
-    Expression e = new TypeidExp(Loc.init, null);
+    Loc loc;
+    Expression e = new TypeidExp(loc, null);
```
wilzbach: This file isn't part of the DMD build and I only accidentally ended up testing it.
It still doesn't hurt to fix the build failure though.

.circleci/run.sh
```diff
@@ -113,9 +113,10 @@ coverage()
     if [ -f ~/dlang/install.sh ] ; then
         source "$(CURL_USER_AGENT=\"$CURL_USER_AGENT\" bash ~/dlang/install.sh dmd-$HOST_DMD_VER --activate)"
     fi
+    RDMD="$(type -p rdmd)"
 
     # build dmd, druntime, and phobos
-    make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=$DMD BUILD=$BUILD ENABLE_WARNINGS=1 PIC="$PIC" all
+    "$RDMD" ./src/build.d MODEL=$MODEL HOST_DMD=$DMD BUILD=$BUILD ENABLE_WARNINGS=1 PIC="$PIC" all
```
wilzbach: So we slowly start introducing `build.d` into more parts of the CI infrastructure.

.circleci/run.sh
```diff
@@ -113,9 +113,14 @@ coverage()
...
         source "$(CURL_USER_AGENT=\"$CURL_USER_AGENT\" bash ~/dlang/install.sh dmd-$HOST_DMD_VER --activate)"
     fi
+    RDMD="$(type -p rdmd)"
 
     # build dmd, druntime, and phobos
-    make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=$DMD BUILD=$BUILD ENABLE_WARNINGS=1 PIC="$PIC" all
+    if [ "$MODEL" == "64" ] ; then
+        "$RDMD" ./src/build.d MODEL=$MODEL HOST_DMD=$DMD BUILD=$BUILD ENABLE_WARNINGS=1 PIC="$PIC" all
+    else
+        make -j$N -C src -f posix.mak MODEL=$MODEL HOST_DMD=$DMD BUILD=$BUILD ENABLE_WARNINGS=1 PIC="$PIC" all
```
wilzbach: FYI: MODEL=32 doesn't fully work yet. I will fix this in another PR.

src/build.d
```diff
@@ -564,7 +576,59 @@ string[string] getEnvironment()
     version(OSX) version(X86_64)
         dObjc = true;
 
-    return env;
+    if (env.getDefault("ENABLE_DEBUG", "0") != "0")
```
JinShil: Can we copy the documentation comments for these from the Makefile?  Or maybe put them in the `help` output?  We should probably do the same for the targets as well.  It's not a requirement for this PR, but eventually it'd be nice.

JinShil: Also, some of these are only applicable to posix, right?

src/dmd/asttypename.d
wilzbach: https://github.com/dlang/dmd/blob/67963b95d021e99577b99982d9b1ebd79834ff9a/src/dmd/expression.d#L4758-L4764

That's why `Loc.init` doesn't work. It might be a good idea to use `auto ref` in the constructors.

src/build.d
wilzbach: > Also, some of these are only applicable to posix, right?

Yes, actually I'm not sure all of these modes even work on Posix (i.e. dmd still fails the test suite with `ENABLE_DEBUG` - https://github.com/dlang/dmd/pull/7528).

wilzbach: > Can we copy the documentation comments for these from the Makefile? Or maybe put them in the help output?

Done.

src/dmd/asttypename.d
rainers: `Loc.initial` is used instead.

wilzbach: Thanks!
BTW this looks like a half-finished story -> https://github.com/dlang/dmd/pull/8427

JinShil (approved):

PR #8421, opened 2018-06-30
convert gother.c to D

WalterBright:


wilzbach:
We really need to upgrade gdc:

```
dmd/backend/global.d:71:17: error: variable SCMAX extern symbols cannot have initializers char[SCMAX] sytab;
```

jacob-carlborg (approved):

wilzbach:
Turns out you can workaround the gdc failure with:

```
enum SCMAX_ = SCMAX;
```

(I added this as a commit to check whether the CIs agree.)

wilzbach (approved):

PR #8422, opened 2018-06-30
Refactor cppmanglewin

thewilsonator:
Reduce nesting and break cascades

### DIFF
--- a/src/dmd/cppmanglewin.d
+++ b/src/dmd/cppmanglewin.d
@@ -673,49 +673,281 @@ private:
         buf.writeByte(cv_mod);
     }
 
-    void mangleName(Dsymbol sym, bool dont_use_back_reference = false)
+    /**
+     * Mangles a symbol with a special name, if any
+     *
+     * Params:
+     *      sym = symbol to mangle
+     *
+     * Returns:
+     *      true if sym has no further mangling needed
+     *      false otherwise
+     */
+    bool mangleSpecialName(Dsymbol sym)
+    {
+        const(char)* mangle;
+        if (sym.isCtorDeclaration())
+            mangle = "?0";
+        else if (sym.isPrimaryDtor())
+            mangle = "?1";
+        else if (!sym.ident)
+            return false;
+        else if (sym.ident == Id.assign)
+            mangle = "?4";
+        else if (sym.ident == Id.eq)
+            mangle = "?8";
+        else if (sym.ident == Id.index)
+            mangle = "?A";
+        else if (sym.ident == Id.call)
+            mangle = "?R";
+        else if (sym.ident == Id.cppdtor)
+            mangle = "?_G";
+        else
+            return false;
+
+        buf.writestring(mangle);
+        return true;
+    }
+
+    /**
+     * Mangles an operator, if any
+     *
+     * Params:
+     *      ti                  = associated template instance of the operator
+     *      symName             = symbol name
+     *      firstTemplateArg    = index if the first argument of the template (because the corresponding c++ operator is not a template)
+     * Returns:
+     *      true if sym has no further mangling needed
+     *      false otherwise
+     */
+    bool mangleOperator(TemplateInstance ti, ref const(char)* symName, ref int firstTemplateArg)
+    {
+        auto whichOp = isCppOperator(ti.name);
+        final switch (whichOp)
+        {
+        case CppOperator.Unknown:
+            return false;
+        case CppOperator.Cast:
+            buf.writestring("?B");
+            return true;
+        case CppOperator.Assign:
+            symName = "?4";
+            return false;
+        case CppOperator.Eq:
+            symName = "?8";
+            return false;
+        case CppOperator.Index:
+            symName = "?A";
+            return false;
+        case CppOperator.Call:
+            symName = "?R";
+            return false;
+
+        case CppOperator.Unary:
+        case CppOperator.Binary:
+        case CppOperator.OpAssign:
+            TemplateDeclaration td = ti.tempdecl.isTemplateDeclaration();
+            assert(td);
+            assert(ti.tiargs.dim >= 1);
+            TemplateParameter tp = (*td.parameters)[0];
+            TemplateValueParameter tv = tp.isTemplateValueParameter();
+            if (!tv || !tv.valType.isString())
+                return false; // expecting a string argument to operators!
+            Expression exp = (*ti.tiargs)[0].isExpression();
+            StringExp str = exp.toStringExp();
+            switch (whichOp)
+            {
+            case CppOperator.Unary:
+                switch (str.peekSlice())
+                {
+                    case "*":   symName = "?D";     goto continue_template;
+                    case "++":  symName = "?E";     goto continue_template;
+                    case "--":  symName = "?F";     goto continue_template;
+                    case "-":   symName = "?G";     goto continue_template;
+                    case "+":   symName = "?H";     goto continue_template;
+                    case "~":   symName = "?S";     goto continue_template;
+                    default:    return false;
+                }
+            case CppOperator.Binary:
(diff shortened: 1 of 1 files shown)


src/dmd/cppmanglewin.d
```diff
@@ -669,149 +669,241 @@ private:
         buf.writeByte(cv_mod);
     }
 
-    void mangleName(Dsymbol sym, bool dont_use_back_reference = false)
+    // Return true if we are finished mangling this symbol
```
WalterBright: When adding new functions, please use proper Ddoc style function comments. These should include `Params:` and `Returns:` sections.

src/dmd/cppmanglewin.d
```diff
@@ -669,149 +669,241 @@ private:
...
+        t.accept(tmp);
+    }
+
+    void mangleName(Dsymbol sym, bool dont_use_back_reference = false)
+    {
+        //printf("mangleName('%s')\n", sym.toChars());
+        const(char)* name = null;
+        bool is_dmc_template = false;
+
+        if (mangleSpecialName(sym)
```
WalterBright: seems to be missing a closing paren

src/dmd/cppmanglewin.d
```diff
@@ -669,149 +669,241 @@ private:
...
+    }
+
+    void mangleTemplateType(RootObject o, VisualCPPMangler tmp)
+    {
+        Type t = isType(o);
+        assert(t);
+        t.accept(tmp);
+    }
+
+    void mangleName(Dsymbol sym, bool dont_use_back_reference = false)
```
WalterBright: No need for default values here.

src/dmd/cppmanglewin.d
```diff
@@ -669,149 +669,241 @@ private:
...
+            mangle = "?8";
+        else if (sym.ident == Id.index)
+            mangle = "?A";
+        else if (sym.ident == Id.call)
+            mangle = "?R";
+        else if (sym.ident == Id.cppdtor)
+            mangle = "?_G";
+        
+        buf.writestring(mangle);
+        return true;
```
WalterBright: Since it always returns true, this seems of little value.

src/dmd/cppmanglewin.d
```diff
@@ -669,149 +669,241 @@ private:
...
+
+        if (mangleSpecialName(sym)
+            return;
+            
         if (TemplateInstance ti = sym.isTemplateInstance())
         {
             auto id = ti.tempdecl.ident;
             const(char)* symName = id.toChars();
 
             int firstTemplateArg = 0;
```
WalterBright: looks like firstTemplateArg should be a bool.

src/dmd/cppmanglewin.d
```diff
@@ -669,149 +669,241 @@ private:
...
+            mangle = "?R";
+        else if (sym.ident == Id.cppdtor)
+            mangle = "?_G";
+        
+        buf.writestring(mangle);
+        return true;
+    }
+    
+    // Return true if we are finished mangling this symbol
+    bool mangleOperator(TemplateInstance ti, CppOperator whichOp, ref const(char)* symName, ref int firstTemplateArg)
```
WalterBright: firstTemplateArg should be `out`, not `ref` (`symName` too?)

thewilsonator: Symname should actually be an `out bool` since its only used to see if it changed. `firstTemplateArg` is cleaner as a `ref` since an `out` must be assigned.

src/dmd/cppmanglewin.d
thewilsonator: I realised that, but I stuff up with git locally so will have to force push, so I'm addressing all these comments now.

src/dmd/cppmanglewin.d
thewilsonator: It was like that before (it just got moved around). Noted.

src/dmd/cppmanglewin.d
thewilsonator: It is used as a start index to a for loop just below, I think its more logical to have it be an int.

src/dmd/cppmanglewin.d
```diff
@@ -663,49 +663,268 @@ private:
...
-        //printf("mangleName('%s')\n", sym.toChars());
-        const(char)* name = null;
-        bool is_dmc_template = false;
+        const(char)* mangle;
         if (sym.isCtorDeclaration())
+            mangle = "?0";
+        else if (sym.isPrimaryDtor())
+            mangle = "?1";
+        else if (!sym.ident)
+            return false;
```
Geod24: Out of interest, when can this happen ?

src/dmd/cppmanglewin.d
```diff
@@ -663,49 +663,268 @@ private:
...
+                    case "-":   symName = "?Z";     goto continue_template;
+                    case "/":   symName = "?_0";    goto continue_template;
+                    case "%":   symName = "?_1";    goto continue_template;
+                    case ">>":  symName = "?_2";    goto continue_template;
+                    case "<<":  symName = "?_3";    goto continue_template;
+                    case "&":   symName = "?_4";    goto continue_template;
+                    case "|":   symName = "?_5";    goto continue_template;
+                    case "^":   symName = "?_6";    goto continue_template;
+                    default:    return false;
+                }
```
Geod24: Did you look at the POSIX mangling by any chance ? Just curious to know if we can refactor common code for operators in their own, shared function (In a later PR).

Geod24 (commented):
Would have been useful to have it broken down into commits that either change something, or move something, as it makes it easier to review.

src/dmd/cppmanglewin.d
thewilsonator: No, I've no idea. Manu probably knows if anyone, he wrote that, I just got rid of the cascaded `break`s.

src/dmd/cppmanglewin.d
```diff
@@ -50,6 +50,16 @@ const(char)* cppTypeInfoMangleMSVC(Dsymbol s)
     assert(0);
 }
 
+private bool rejectImmutableShared(Type type)
```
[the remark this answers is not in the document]
thewilsonator: Will do.

src/dmd/cppmanglewin.d
thewilsonator: dunno. The old code checked for it.

src/dmd/cppmanglewin.d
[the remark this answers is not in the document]
thewilsonator: Done.

src/dmd/cppmanglewin.d
thewilsonator: I have now, heh. Manu's got basically the same code in the posix, replete with cascading `break`s.

src/dmd/cppmanglewin.d
```diff
@@ -102,24 +112,18 @@ public:
...
-        if (type.isImmutable() || type.isShared())
-        {
-            type.error(Loc.initial, "Internal Compiler Error: `shared` or `immutable` types can not be mapped to C++ (%s)", type.toChars());
-        }
-        else
-        {
-            type.error(Loc.initial, "Internal Compiler Error: type `%s` can not be mapped to C++\n", type.toChars());
-        }
+        if (rejectImmutableShared(type))
+            return;
```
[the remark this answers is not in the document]
thewilsonator: The old code used `visit(cast(Type)type);` as a way of triggering errors and that was one of the checks, I can't be sure I've got all of them.

Geod24 (approved):

src/dmd/cppmanglewin.d
TurkeyMan: ;) .. Yeah they didn't actually start the same... but they drifted towards each other.

PR #8424, opened 2018-07-01
convert gflow.c to D

WalterBright:


WalterBright:
@CyberShadow looks like a problem with CyberShadow/DAutoTest : 
```
/usr/bin/ld: final link failed: No space left on device
posix.mak:811: recipe for target 'test_dspec' failed
```

src/dmd/backend/gloop.c
```diff
@@ -1268,6 +1268,8 @@ STATIC void markinvar(elem *n,vec_t rd)
  *      e = defnod[] entry that is an assignment to a variable
  */
 
+extern "C"
+{
 void fillInDNunambig(vec_t v, elem *e)
```
wilzbach: Isn't this change unrelated?

src/dmd/backend/gflow.d
```diff
@@ -6,26 +6,50 @@
...
+import dmd.backend.outbuf;
+import dmd.backend.ty;
+import dmd.backend.type;
+
+import dmd.backend.dlist;
+import dmd.backend.dvec;
+
+void vec_setclear(size_t b, vec_t vs, vec_t vc) { vec_setbit(b, vs); vec_clearbit(b, vc); }
+
+bool Eunambig(elem* e) { return OTassign(e.Eoper) && e.EV.E1.Eoper == OPvar; }
```
wilzbach: This is already defined in `gdag.d` - why not import it (or move it to a common bit)?

wilzbach (approved):

src/dmd/backend/gloop.c
WalterBright: No. I set it to C to avoid name mangling issues with `vec_t` which is our annoying friend `size_t`.

src/dmd/backend/gflow.d
WalterBright: Because the 'header' files are a mess at the moment, and I decided to defer dealing with that. My goal is to do the least disruptive thing to get the files converted to D. Then, start doing refactoring and removing technical debt.

PR #8425, opened 2018-07-01
Refactor POSIX's C++ mangling code

Geod24:
Better reviewed commit-by-commit.

While working on https://issues.dlang.org/show_bug.cgi?id=16479 I had a hard time following some of the constructs, so I started adding docs and trying to reduce nesting / cascading.
This is purely the refactor part of it, I am still working on the fix. Given @thewilsonator had submitted a similar PR to cppmanglewin, I figured it would be better to submit it earlier than later, to avoid conflicts / duplication of work.

### DIFF
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -745,21 +745,7 @@ private final class CppMangleVisitor : Visitor
             Dsymbol p = d.toParent();
             if (p && !p.isModule() && tf.linkage == LINK.cpp)
             {
-                /* <nested-name> ::= N [<CV-qualifiers>] <prefix> <unqualified-name> E
-                 *               ::= N [<CV-qualifiers>] <template-prefix> <template-args> E
-                 */
-                buf.writeByte('N');
-                CV_qualifiers(tf);
-
-                /* <prefix> ::= <prefix> <unqualified-name>
-                 *          ::= <template-prefix> <template-args>
-                 *          ::= <template-param>
-                 *          ::= # empty
-                 *          ::= <substitution>
-                 *          ::= <prefix> <data-member-prefix>
-                 */
-                prefix_name(p);
-                //printf("p: %s\n", buf.peekString());
+                this.mangleNestedFuncPrefix(tf, p);
 
                 if (d.isCtorDeclaration())
                     buf.writestring("C1");
@@ -806,9 +792,8 @@ private final class CppMangleVisitor : Visitor
         Dsymbol p = ti.toParent();
         if (p && !p.isModule() && tf.linkage == LINK.cpp)
         {
-            buf.writeByte('N');
-            CV_qualifiers(tf);
-            prefix_name(p);
+            this.mangleNestedFuncPrefix(tf, p);
+
             if (d.isCtorDeclaration())
             {
                 buf.writestring("C1");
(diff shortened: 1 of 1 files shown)


thewilsonator:
I don't think there will be conflicts or duplicated effort if 16479 is purely posix, as 19043 is purely windows. I'm trying to get it in a shape I can understand it so I can fix the mangling of `struct __dslice(T)` for `T == const(char)` (currently drops the const).

Geod24:
@thewilsonator : Pardon my poor phrasing. I didn't want to imply that our PR will conflict, but that I realized waiting too long might lead to other people working on this part of the code as well (there's been quite some interest for C++ interop recently).

JinShil:
LGTM.  There are probably some opportunities to add a few attributes to these functions and their parameters.  See (1) and (7) here:  https://forum.dlang.org/post/p6oibo$1lmi$1@digitalmars.com

thewilsonator:
No problem.

Geod24:
> There are probably some opportunities to add a few attributes to these functions and their parameters.

Good point, I didn't think about this. I'll see if I can spare some time on it in the near future.

src/dmd/cppmangle.d
```diff
@@ -116,22 +179,7 @@ private final class CppMangleVisitor : Visitor
...
-                        write_seq_id(i / 36);
-                        i %= 36;
-                    }
-                    i += (i < 10) ? '0' : 'A' - 10;
-                    buf.writeByte(cast(char)i);
-                }
-
-                write_seq_id(i - 1);
-            }
+            writeSequenceFromIndex(i);
```
WalterBright: I don't see a purpose to making this reusable, because it isn't used anywhere else. The point of nested functions like this is to make it clear it isn't used elsewhere. This refactoring makes it harder to understand, not simpler, because the logic is spread out.

src/dmd/cppmangle.d
```diff
@@ -100,12 +101,74 @@ bool isPrimaryDtor(const Dsymbol sym)
...
+    /**
+     * Write a seq-id from an index number, excluding the terminating '_'
+     *
+     * Params:
+     *   idx = the index in a substitution list.
+     *         Note that index 0 has no value, and `S0_` would be the
+     *         substitution at index 1 in the list.
+     *
+     * See-Also:
+     *  https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangle.seq-id
```
WalterBright: The link comment is good.

WalterBright:
What is the purpose of the reorder (commit 3) ?

src/dmd/cppmangle.d
Geod24: As mentioned in the commit message, this will be reused by the fix to https://issues.dlang.org/show_bug.cgi?id=16479 . Template parameter substitution use the same base36 index as "regular" substitution.

Geod24:
> What is the purpose of the reorder (commit 3) ?

I wanted to get rid of unnecessary `extern (C++)` in this module.
However, `extern(C++)` is necessary for the `visit` overloads.
So I put all the `visit` overloads at the end, after an `extern (C++):`.

In addition, I've moved the entry point of this class to the top, to make it easier to read, since `mangleItanium` calls the ctor then `mangleOf` from this class, and `mangleOf` forwards to either `mangle_variable` or `mangle_function`.

PR #8426, opened 2018-07-01
convert blockopt.c to blockopt.d

wilzbach:


src/dmd/backend/blockopt.d
```diff
@@ -300,20 +336,20 @@ void blocklist_free(block **pb)
...
+    vec_free(b.Boutlv);
+    vec_free(b.Bin);
+    vec_free(b.Bout);
+    vec_free(b.Bgen);
+    vec_free(b.Bkill);
+    vec_free(b.Bout2);
+    vec_free(b.Bgen2);
+    vec_free(b.Bkill2);
+
+    memset(&b,0,b.sizeof);
```
wilzbach: I'm not sure whether this is the correct translation.

src/dmd/backend/blockopt.d
```diff
@@ -652,196 +711,196 @@ void blockopt(int iter)
...
                     }
 
-                    b->BC = b2->BC;
-                    b->BS = b2->BS;
-                    b->Bsucc = b2->Bsucc;
-                    b2->Bsucc = NULL;
-                    b2->BC = BCret;             /* a harmless one       */
-                    cmes3("brcombine(): %p goto %p eliminated\n",b,b2);
+                    b.Bswitch = b2.Bswitch;
+                    b.Bswitch = b2.Bswitch;
```
wilzbach: I'm not sure whether this is the correct translation.

src/dmd/backend/blockopt.d
```diff
@@ -2057,96 +2150,96 @@ STATIC int el_anyframeptr(elem *e)
...
                     maxblks += 3;
                     block *b2 = block_calloc();
-                    b2->Bnext = b->Bnext;
-                    b->Bnext = b2;
-                    b2->BC = b->BC;
-                    b2->BS = b->BS;
+                    b2.Bnext = b.Bnext;
+                    b.Bnext = b2;
+                    b2.BC = b.BC;
+                    b2.Bswitch = b.Bswitch;
```
wilzbach: I'm not sure whether this is the correct translation.

src/dmd/backend/blockopt.d
```diff
@@ -6,80 +6,111 @@
...
 block *curblock;        /* current block being read in                  */
 block *block_last;      // last block read in
 
-static block * block_freelist;
+block * block_freelist;
 
 ////////////////////////////
 // Storage allocator.
 
-static block blkzero;
```
jacob-carlborg: Walter usually coverts `static` to `private`.

wilzbach: This is in a `private __gshared` block.

jacob-carlborg: Ok.

wilzbach:
Not sure why Windows isn't passing.
Auto-tester simply times out for Windows when building sanitize_json:

```
timed out after 3600 seconds, step failed
```

and AppVeyor even fails while linking: 

```
backend.lib(blockopt.obj) : error LNK2001: unresolved external symbol _D3dmd7backend5gflow12__ModuleInfoZ
``

**Comment by rainers** [comment]
> and AppVeyor even fails while linking:
> 
> backend.lib(blockopt.obj) : error LNK2001: unresolved external symbol _D3dmd7backend5gflow12__ModuleInfoZ

This happens because some files are compiled with -betterC, but blockopt is not. I suspect some of the import dependencies have `static this()`.

File: src/dmd/backend/blockopt.d (hunk 2 above)
Comment (rainers) [reply src/dmd/backend/blockopt.d]: This is copying too little, it should be somethink llike
```
                    b.BC = b2.BC;
                    b.cover_BS[] = b2.cover_BS[];
```
with
```
        // add member at least as large as the other elements of this union
        void[jcatchvar.sizeof + Bscope_index.sizeof + Blast_index.sizeof] cover_BS;
```
added to the respective union in `block`.

File: src/dmd/backend/blockopt.d [hunk 5]
```diff
@@ -1519,42 +1592,44 @@ STATIC elem * bl_delist(list_t el)
...
+        for (block *b = startblock; b; b = b.Bnext)
+            b.Blist = bl_enlist(b.Belem);
 
         /* Search for two blocks that have the same successor list.
            If the first expressions both lists are the same, split
            off a new block with that expression in it.
          */
-        for (block *b = startblock; b; b = b->Bnext)
+        static if (SCPP_OR_NTEXCEPTIONS)
+            bool additionalAnd = b.Btry == bn.Btry;
```
rainers: This seems not to be actually compiled, but it has to be inside the inner loop.

rainers:
> Not sure why Windows isn't passing.
> Auto-tester simply times out for Windows when building sanitize_json:

It's not the building of sanitize_json, but the execution of the d_do_test unittests: it waits forever for threads to join when it returns from main. I guess there are still code gen bugs (after my proposed changes).

rainers:
handling of NTEXCEPTIONS was incorect, too. I pushed some fixes.

wilzbach:
>  I pushed some fixes.

Thanks a lot! It's still failing in the Visual Studio build with:

```
  ..\dmd\backend\blockopt.d(55): error : Function type does not match previously declared function with the same mangled name: `mem_ffree` [C:\projects\dmd\src\vcbuild\dmd.vcxproj]
```

I don't see where it's apparently declared previously...

rainers:
> I don't see where it's apparently declared previously...

This is merge conflict because https://github.com/dlang/dmd/pull/8466 now also defines mem_ffree.

wilzbach:
BTW this is passing the CIs now :)

jacob-carlborg:
> BTW this is passing the CIs now :)

No 😊. Merge conflicts.

wilzbach:
> No blush. Merge conflicts.

Resolved.

src/dmd/backend/blockopt.d
WalterBright: It isn't. It's supposed to zero the pointers to the vectors that were just freed. An easy way to do this is to write a nested function:
```
static void vfree(ref vec_t* v) { vec_free(v); v = null; }
```
and call that instead of vec_free.

WalterBright (changes requested):
Nice work! Fix the memset as described, and it's good to go.

WalterBright (approved):

PR #8428, opened 2018-07-01
Fix Issue 19024 - [ICE] AssertError@dmd/dsymbolsem.d(4317): Assertion failure

ibuclaw:
Same fix as #6964, but for structs.

PR #8429, opened 2018-07-01
Start deprecation period of identical functions in a single module

wilzbach:
Initially started in https://github.com/dlang/dmd/pull/7577, but reverted in https://github.com/dlang/dmd/pull/7969 due to the immediate 2.079 release and a reported issue for it (https://issues.dlang.org/show_bug.cgi?id=18385).

This now officially starts the deprecation period of identical functions in a single module.

The examples provided in 18385 are:

```d
extern (C):

void foo(int) { }
void foo(double) { }

struct S
{
    static void foo(int) {}
    static void foo(double) {}
}
```

They clearly can't work because the mangling of these functions is the same.
The deprecation period is ten releases after the new deprecation DIP.

Note that the following will still work:

```d
extern (C):

void foo(int) { }
extern(D) void foo(double) { } // OK as it has a different mangling
```

or:

```d
extern (C):

void foo(int) { }
void foo(double); // duplicate declarations are allowed
```

See also: https://github.com/dlang/dmd/pull/8428

### DIFF
--- a/changelog/duplicate-implementations-deprecation.dd
+++ b/changelog/duplicate-implementations-deprecation.dd
@@ -1,0 +1,23 @@
+Usage of identical functions in a single module
+
+A binary should never contain multiple versions of a symbol.
+Before this version, it was allowed to provide multiple implementation for a function,
+even if they result in the same mangling, which can result in undefined behavior:
+
+---
+extern(C):
+void foo(int) { }
+void foo(double) { } // will error now
+---
+
+However, it's still allowed to have multiple declarations as long as there's at
+most one implementation:
+
+---
+extern(C):
+void foo(int) { }
+void foo(double);
+---
+
+Note that this change is only relevant for mangling schemes that do no support
+overloading like `extern(C)`.
--- a/changelog/duplicate-implementations-deprecation.dd
+++ b/changelog/duplicate-implementations-deprecation.dd
@@ -2,12 +2,15 @@ Usage of identical functions in a single module
 
 A binary should never contain multiple versions of a symbol.
 Before this version, it was allowed to provide multiple implementation for a function,
-even if they result in the same mangling, which can result in undefined behavior:
+even if they result in the same mangling, which can result in undefined behaviour:
 
 ---
 extern(C):
 void foo(int) { }
-void foo(double) { } // will error now
+void foo(double) { } // will issue a deprecation now
+
+void bar() {}
+void bar() {} // will raise an error
 ---
 
 However, it's still allowed to have multiple declarations as long as there's at
--- a/changelog/duplicate-implementations-deprecation.dd
+++ b/changelog/duplicate-implementations-deprecation.dd
@@ -1,26 +1,33 @@
-Usage of identical functions in a single module
+Diagnostics for conflicting function definitions within a module
 
-A binary should never contain multiple versions of a symbol.
-Before this version, it was allowed to provide multiple implementation for a function,
-even if they result in the same mangling, which can result in undefined behaviour:
+Previously, multiple definitions of identical functions within a module were not
+recognized, although they have the same mangling. This was problematic because a
+binary cannot contain multiple definitions of a symbol, which caused undefined
+behavior depending on the compiler backend.
+
+DMD will now raise an error message if there are conflicting implementations
+within a single module:
+
+---
+void foo() {}
+void foo() {} // error
+---
+
+Multiple declarations are still allowed as long as there is at most one definition:
+
+---
+void bar(int);
+void bar(int) { }
+void bar(int);
+---
+
+DMD will issue a deprecation for mangling schemes that don't support overloading
+(`extern(C|Windows|System)`):
 
 ---
 extern(C):
 void foo(int) { }
-void foo(double) { } // will issue a deprecation now
-
-void bar() {}
-void bar() {} // will raise an error
+void foo(double) { } // deprecation
 ---
 
-However, it's still allowed to have multiple declarations as long as there's at
-most one implementation:
-
----
-extern(C):
-void foo(int) { }
-void foo(double);
----
-
-Note that this change is only relevant for mangling schemes that do no support
-overloading like `extern(C)`.
+This deprecation will become an error in 2.105.
--- a/src/dmd/semantic2.d
+++ b/src/dmd/semantic2.d
@@ -371,6 +371,7 @@ private extern(C++) final class Semantic2Visitor : Visitor
             // Always starts the lookup from 'this', because the conflicts with
             // previous overloads are already reported.
             alias f1 = fd;
+            auto tf1 = cast(TypeFunction) f1.type;
             mangleToFuncSignature(buf1, f1);
 
             overloadApply(f1, (Dsymbol s)
@@ -394,13 +395,17 @@ private extern(C++) final class Semantic2Visitor : Visitor
                 if (f1.overrides(f2))
                     return 0;
 
+                auto tf2 = cast(TypeFunction) f2.type;
+
                 // extern (C) functions always conflict each other.
                 if (f1.ident == f2.ident &&
                     f1.toParent2() == f2.toParent2() &&
                     (f1.linkage != LINK.d && f1.linkage != LINK.cpp) &&
-                    (f2.linkage != LINK.d && f2.linkage != LINK.cpp))
+                    (f2.linkage != LINK.d && f2.linkage != LINK.cpp) &&
+
+                    // But allow the hack to declare overloads with different parameters/STC's
+                    (!tf1.attributesEqual(tf2) || tf1.parameterList != tf2.parameterList))
                 {
-                    auto tf2 = cast(TypeFunction)f2.type;
                     // @@@DEPRECATED_2.094@@@
                     // Deprecated in 2020-08, make this an error in 2.104
                     deprecation(f2.loc, "%s `%s%s` cannot be overloaded with %s`extern(%s)` function at %s",
@@ -425,7 +430,6 @@ private extern(C++) final class Semantic2Visitor : Visitor
                 //printf("+%s\n\ts1 = %s\n\ts2 = %s @ [%s]\n", toChars(), s1, s2, f2.loc.toChars());
                 if (strcmp(s1, s2) == 0)
                 {
-                    auto tf2 = cast(TypeFunction)f2.type;
                     error(f2.loc, "%s `%s%s` conflicts with previous declaration at %s",
                             f2.kind(),
                             f2.toPrettyChars(),
--- a/src/dmd/semantic2.d
+++ b/src/dmd/semantic2.d
@@ -359,9 +359,6 @@ private extern(C++) final class Semantic2Visitor : Visitor
 
         //printf("FuncDeclaration::semantic2 [%s] fd0 = %s %s\n", loc.toChars(), toChars(), type.toChars());
 
-        // https://issues.dlang.org/show_bug.cgi?id=18385
-        // Disable for 2.079, s.t. a deprecation cycle can be started with 2.080
-        if (0)
         if (fd.overnext && !fd.errors)
         {
             OutBuffer buf1;
@@ -414,7 +411,8 @@ private extern(C++) final class Semantic2Visitor : Visitor
                         return 0;
 
                     auto tf2 = cast(TypeFunction)f2.type;
-                    error(f2.loc, "%s `%s%s` cannot be overloaded with %s`extern(%s)` function at %s",
+                    // @@@DEPRECATED_2.092@@@
+                    f2.deprecation("%s `%s%s` cannot be overloaded with %s`extern(%s)` function at %s",
                             f2.kind(),
                             f2.toPrettyChars(),
                             parametersTypeToChars(tf2.parameterList),
@@ -435,7 +433,7 @@ private extern(C++) final class Semantic2Visitor : Visitor
                 if (strcmp(s1, s2) == 0)
                 {
                     auto tf2 = cast(TypeFunction)f2.type;
-                    error(f2.loc, "%s `%s%s` conflicts with previous declaration at %s",
+                    f2.error("%s `%s%s` conflicts with previous declaration at %s",
                             f2.kind(),
                             f2.toPrettyChars(),
                             parametersTypeToChars(tf2.parameterList),
--- a/src/dmd/semantic2.d
+++ b/src/dmd/semantic2.d
@@ -408,12 +408,9 @@ private extern(C++) final class Semantic2Visitor : Visitor
                 {
                     // @@@DEPRECATED_2.094@@@
                     // Deprecated in 2020-08, make this an error in 2.104
-                    deprecation(f2.loc, "%s `%s%s` cannot be overloaded with %s`extern(%s)` function at %s",
-                            f2.kind(),
-                            f2.toPrettyChars(),
-                            parametersTypeToChars(tf2.parameterList),
-                            (f1.linkage == f2.linkage ? "another " : "").ptr,
-                            linkageToChars(f1.linkage), f1.loc.toChars());
+                    f2.deprecation("cannot overload `extern(%s)` function at %s",
+                            linkageToChars(f1.linkage),
+                            f1.loc.toChars());
 
                     // Enable this when turning the deprecation into an error
                     // f2.type = Type.terror;
--- a/src/dmd/semantic2.d
+++ b/src/dmd/semantic2.d
@@ -359,7 +359,11 @@ private extern(C++) final class Semantic2Visitor : Visitor
 
         //printf("FuncDeclaration::semantic2 [%s] fd0 = %s %s\n", loc.toChars(), toChars(), type.toChars());
 
-        if (fd.overnext && !fd.errors)
+        // Only check valid functions which have a body to avoid errors
+        // for multiple declarations, e.g.
+        // void foo();
+        // void foo();
+        if (fd.fbody && fd.overnext && !fd.errors)
         {
             OutBuffer buf1;
             OutBuffer buf2;
@@ -411,15 +415,18 @@ private extern(C++) final class Semantic2Visitor : Visitor
                         return 0;
 
                     auto tf2 = cast(TypeFunction)f2.type;
-                    // @@@DEPRECATED_2.092@@@
-                    f2.deprecation("%s `%s%s` cannot be overloaded with %s`extern(%s)` function at %s",
+                    // @@@DEPRECATED_2.094@@@
+                    // Deprecated in 2020-08, make this an error in 2.104
+                    deprecation(f2.loc, "%s `%s%s` cannot be overloaded with %s`extern(%s)` function at %s",
                             f2.kind(),
                             f2.toPrettyChars(),
                             parametersTypeToChars(tf2.parameterList),
                             (f1.linkage == f2.linkage ? "another " : "").ptr,
                             linkageToChars(f1.linkage), f1.loc.toChars());
-                    f2.type = Type.terror;
-                    f2.errors = true;
+
+                    // Enable this when turning the deprecation into an error
+                    // f2.type = Type.terror;
+                    // f2.errors = true;
                     return 0;
                 }
 
@@ -433,7 +440,7 @@ private extern(C++) final class Semantic2Visitor : Visitor
                 if (strcmp(s1, s2) == 0)
                 {
                     auto tf2 = cast(TypeFunction)f2.type;
-                    f2.error("%s `%s%s` conflicts with previous declaration at %s",
+                    error(f2.loc, "%s `%s%s` conflicts with previous declaration at %s",
                             f2.kind(),
                             f2.toPrettyChars(),
                             parametersTypeToChars(tf2.parameterList),
--- a/src/dmd/semantic2.d
+++ b/src/dmd/semantic2.d
@@ -370,7 +370,7 @@ private extern(C++) final class Semantic2Visitor : Visitor
 
             // Always starts the lookup from 'this', because the conflicts with
             // previous overloads are already reported.
-            auto f1 = fd;
+            alias f1 = fd;
             mangleToFuncSignature(buf1, f1);
 
             overloadApply(f1, (Dsymbol s)
@@ -380,7 +380,7 @@ private extern(C++) final class Semantic2Visitor : Visitor
                     return 0;
 
                 // Don't have to check conflict between declaration and definition.
-                if ((f1.fbody !is null) != (f2.fbody !is null))
+                if (f2.fbody is null)
                     return 0;
 
                 /* Check for overload merging with base class member functions.
@@ -400,20 +400,6 @@ private extern(C++) final class Semantic2Visitor : Visitor
                     (f1.linkage != LINK.d && f1.linkage != LINK.cpp) &&
                     (f2.linkage != LINK.d && f2.linkage != LINK.cpp))
                 {
-                    /* Allow the hack that is actually used in druntime,
-                     * to ignore function attributes for extern (C) functions.
-                     * TODO: Must be reconsidered in the future.
-                     *  BUG: https://issues.dlang.org/show_bug.cgi?id=18206
-                     *
-                     *  extern(C):
-                     *  alias sigfn_t  = void function(int);
-                     *  alias sigfn_t2 = void function(int) nothrow @nogc;
-                     *  sigfn_t  bsd_signal(int sig, sigfn_t  func);
-                     *  sigfn_t2 bsd_signal(int sig, sigfn_t2 func) nothrow @nogc;  // no error
-                     */
-                    if (f1.fbody is null || f2.fbody is null)
-                        return 0;
-
                     auto tf2 = cast(TypeFunction)f2.type;
                     // @@@DEPRECATED_2.094@@@
                     // Deprecated in 2020-08, make this an error in 2.104
--- a/test/fail_compilation/fail2789.d
+++ b/test/fail_compilation/fail2789.d
@@ -1,11 +1,18 @@
 /*
-DISABLED: win32 win64 osx linux freebsd dragonflybsd
-https://issues.dlang.org/show_bug.cgi?id=18385
-Disabled for 2.079, s.t. a deprecation cycle can be started with 2.080
+REQUIRED_ARGS: -de
 TEST_OUTPUT:
 ---
-fail_compilation/fail2789.d(15): Error: function `fail2789.A2789.m()` conflicts with previous declaration at fail_compilation/fail2789.d(10)
-fail_compilation/fail2789.d(25): Error: function `fail2789.A2789.m()` conflicts with previous declaration at fail_compilation/fail2789.d(10)
+fail_compilation/fail2789.d(25): Error: function `fail2789.A2789.m` function `fail2789.A2789.m()` conflicts with previous declaration at fail_compilation/fail2789.d(20)
+fail_compilation/fail2789.d(35): Error: function `fail2789.A2789.m` function `fail2789.A2789.m()` conflicts with previous declaration at fail_compilation/fail2789.d(20)
+fail_compilation/fail2789.d(47): Error: function `fail2789.f3` function `fail2789.f3()` conflicts with previous declaration at fail_compilation/fail2789.d(46)
+fail_compilation/fail2789.d(50): Error: function `fail2789.f4` function `fail2789.f4()` conflicts with previous declaration at fail_compilation/fail2789.d(49)
+fail_compilation/fail2789.d(53): Error: function `fail2789.f5` function `fail2789.f5()` conflicts with previous declaration at fail_compilation/fail2789.d(52)
+fail_compilation/fail2789.d(56): Error: function `fail2789.f6` function `fail2789.f6()` conflicts with previous declaration at fail_compilation/fail2789.d(55)
+fail_compilation/fail2789.d(60): Deprecation: function `fail2789.f_ExternC1` function `fail2789.f_ExternC1()` cannot be overloaded with another `extern(C)` function at fail_compilation/fail2789.d(59)
+fail_compilation/fail2789.d(63): Deprecation: function `fail2789.f_ExternC2` function `fail2789.f_ExternC2(int)` cannot be overloaded with another `extern(C)` function at fail_compilation/fail2789.d(62)
+fail_compilation/fail2789.d(66): Deprecation: function `fail2789.f_ExternC3` function `fail2789.f_ExternC3()` cannot be overloaded with another `extern(C)` function at fail_compilation/fail2789.d(65)
+fail_compilation/fail2789.d(69): Error: function `fail2789.f_MixExtern1` function `fail2789.f_MixExtern1()` conflicts with previous declaration at fail_compilation/fail2789.d(68)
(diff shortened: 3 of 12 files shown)


JinShil (approved):

WalterBright:
There's the issue of:
```
void foo();
void foo() { ... }
```
This is a common pattern in C, should it be allowed in D?

wilzbach:
> This is a common pattern in C, should it be allowed in D?

I don't see a reason to forbid it. Note that this PR doesn't touch this part (see the tests for examples), it will only trigger a deprecation warning when the mangling of two function is equivalent AND both have a body.

jacob-carlborg:
It's been previously stated that when D interfaces with another language it should follow it's own overloading and lookup rules. Otherwise we would need implement argument dependent lookup for `extern(C++)` functions.

One use case I could see would be to add overloads so that the D binding allows passing the same values as the C function would allow without requiring casts. For example:

```d
extern (C) void foo(bool);
extern (C) void foo(int);

foo(true);
foo(1);
```

jacob-carlborg:
Another use case is, not sure if you're changes cover that:

```d
template Metaclass
{
    extern (Objective-C) interface Class
    {
        extern (C) pragma(mangle, "objc_lookUpClass") static typeof(this) objc_lookUpClass(const(char)* name);
    }
}

extern (Objective-C) NSObject
{
    mixin Metaclass;
}

extern (Objective-C) NSObject
{
    mixin Metaclass;
}
```

Otherwise calling `objc_lookUpClass` always requires a cast.

jacob-carlborg:
> This PR is just about having two duplicate implementation in the same module.

Aha, ok.

jacob-carlborg:
I'm a bit unclear on the process, when do we add a changelog entry?

wilzbach:
> I'm a bit unclear on the process, when do we add a changelog entry?

Sorry. I expected more outcry / request for changes on this one. Added the changelog entry now.

changelog/duplicate-implementations-deprecation.dd
```diff
@@ -0,0 +1,23 @@
+Usage of identical functions in a single module
+
+A binary should never contain multiple versions of a symbol.
+Before this version, it was allowed to provide multiple implementation for a function,
+even if they result in the same mangling, which can result in undefined behavior:
+
+---
+extern(C):
+void foo(int) { }
+void foo(double) { } // will error now
```
jacob-carlborg: `error` should be `deprecation`?

jacob-carlborg:
Is Jenkins disabled? DWT might be affected. But since it’s only a deprecation I think it shouldn’t be a problem.

wilzbach:
> Is Jenkins disabled? DWT might be affected. But since it’s only a deprecation I think it shouldn’t be a problem.

Yep it's running again, but here's what I locally get:

```
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(397,7): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLayoutLine.is_paragraph_start` function org.eclipse.swt.internal.c.pango._PangoLayoutLine.is_paragraph_start(uint v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(394,7)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(398,7): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLayoutLine.resolved_dir` function org.eclipse.swt.internal.c.pango._PangoLayoutLine.resolved_dir(uint v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(395,7)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(524,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_line_break` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_line_break(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(512,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(525,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_mandatory_break` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_mandatory_break(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(513,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(526,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_char_break` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_char_break(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(514,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(527,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_white` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_white(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(515,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(528,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_cursor_position` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_cursor_position(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(516,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(529,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_word_start` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_word_start(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(517,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(530,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_word_end` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_word_end(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(518,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(531,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_sentence_boundary` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_sentence_boundary(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(519,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(532,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_sentence_start` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_sentence_start(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(520,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(533,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_sentence_end` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_sentence_end(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(521,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(534,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.backspace_deletes_character` function org.eclipse.swt.internal.c.pango._PangoLogAttr.backspace_deletes_character(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(522,6)
org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(535,6): Deprecation: function `org.eclipse.swt.internal.c.pango._PangoLogAttr.is_expandable_space` function org.eclipse.swt.internal.c.pango._PangoLogAttr.is_expandable_space(bool v) cannot be overloaded with another extern(C) function at org.eclipse.swt.gtk.linux.x86/src/org/eclipse/swt/internal/c/pango.d(523,6)
---
ERROR: This is a compiler bug.
Please report it via https://issues.dlang.org/enter_bug.cgi
with, preferably, a reduced, reproducible example and the information below.
DustMite (https://github.com/CyberShadow/DustMite/wiki) can help with the reduction.
---
DMD v2.081.1-beta.1-186-gd4e3895ee
predefs   VibeCustomMain Have_dwt Have_dwt_base DigitalMars Posix linux ELFv1 LittleEndian D_Version2 all D_SIMD D_InlineAsm_X86_64 X86_64 CRuntime_Glibc D_LP64 D_PIC unittest assert D_HardFloat
binary    /home/seb/dlang/dmd/generated/linux/release/64/dmd
version   v2.081.1-beta.1-186-gd4e3895ee
config    /home/seb/dlang/dmd/generated/linux/release/64/dmd.conf
DFLAGS    -I/home/seb/dlang/dmd/generated/linux/release/64/../../../../../druntime/import -I/home/seb/dlang/dmd/generated/linux/release/64/../../../../../phobos -L-L/home/seb/dlang/dmd/generated/linux/release/64/../../../../../phobos/generated/linux/release/64 -L--export-dynamic -fPIC
---
core.exception.AssertError@dmd/glue.d(774): Assertion failure
----------------
??:? _d_assertp [0x559d1005]
dmd/glue.d:774 _Z25FuncDeclaration_toObjFileP15FuncDeclarationb [0x558f3f0c]
dmd/toobj.d:312 _ZN9toObjFile9ToObjFile5visitEP15FuncDeclaration [0x55901eb0]
dmd/func.d:2369 _ZN15FuncDeclaration6acceptEP7Visitor [0x5584c8d9]
dmd/toobj.d:526 _ZN9toObjFile9ToObjFile5visitEP17StructDeclaration [0x5590264c]
dmd/dstruct.d:589 _ZN17StructDeclaration6acceptEP7Visitor [0x557e8e65]
dmd/toobj.d:731 _ZN9toObjFile9ToObjFile5visitEP17AttribDeclaration [0x55902d01]
dmd/parsetimevisitor.d:67 _ZN16ParseTimeVisitorI10ASTCodegenE5visitEP15LinkDeclaration [0x558d2a4a]
dmd/attrib.d:469 _ZN15LinkDeclaration6acceptEP7Visitor [0x55790205]
dmd/toobj.d:1025 _Z9toObjFileP7Dsymbolb [0x55901e19]
dmd/glue.d:403 _Z10genObjFileP6Moduleb [0x558f3739]
dmd/mars.d:927 int dmd.mars.tryMain(ulong, const(char)**) [0x558855ca]
dmd/mars.d:1078 _Dmain [0x55885f9a]
```

I will have a look into the segfault, but the deprecations messages are legit as C doesn't mangle its parameters. Try an `objdump` on this minimal example:

```
long bitfield;

extern(C):
auto is_paragraph_start(){ return ( bitfield >> 0 ) & 0x01; }
auto is_paragraph_start(uint v ){ bitfield &= ~( 0x01 << 0 ); bitfield |= ( v & 0x01 ) << 0; return is_paragraph_start(); }
``

**Comment by jacob-carlborg** [comment]
I think there are several functions that have no reason to be `extern(C)`

**Comment by JinShil** [comment]
Removed the "72h no object -> merge" label as it was applied 7 days ago and this isn't passing the auto tester or Jenkins.  I deprecated the auto tester failing test, so let's see if it passes.

**Comment by Geod24** [comment]
Failure is definitely related:
```

core.exception.AssertError@dmd/glue.d(774): Assertion failure

----------------

??:? _d_assertp [0x7500d5]

dmd/glue.d:774 _Z25FuncDeclaration_toObjFileP15FuncDeclarationb [0x673534]

dmd/toobj.d:312 _ZN9toObjFile9ToObjFile5visitEP15FuncDeclaration [0x6814d8]

dmd/func.d:2369 _ZN15FuncDeclaration6acceptEP7Visitor [0x5cd001]

dmd/toobj.d:526 _ZN9toObjFile9ToObjFile5visitEP17StructDeclaration [0x681c74]

dmd/dstruct.d:589 _ZN17StructDeclaration6acceptEP7Visitor [0x56a031]

dmd/toobj.d:1025 _Z9toObjFileP7Dsymbolb [0x681441]

dmd/glue.d:403 _Z10genObjFileP6Moduleb [0x672d61]

dmd/glue.d:156 _Z18obj_write_deferredP7Library [0x6723b4]

dmd/mars.d:947 int dmd.mars.tryMain(ulong, const(char)**) [0x6055b8]

dmd/mars.d:1078 _Dmain [0x605dca]

/var/lib/jenkins/dlang_projects@7/distribution/bin/dmd failed with exit code 1.

script returned exit code 2
```

**Comment by wilzbach** [comment]
No feel free to update DWT. Once I have time for this, I will manage to reproduce the segfault from DWT, reduce it and it to the testsuite anyhow.

**Comment by JohanEngelen** [comment]
Is there another PR for updating the spec?

File: test/fail_compilation/test18385.d [hunk 2]
```diff
@@ -0,0 +1,31 @@
...
+{
+    static void foo(int) {}
+    static void foo(double) {}
+}
+
+void foo2(int) { }
+extern(D) void foo2(double) { } // OK as it has a different mangling
+
+void foo3(int) { }
+void foo3(double); // duplicate declarations are allowed
```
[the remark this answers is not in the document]
JohanEngelen: (so you probably wanted to write `void foo3(int);` ?)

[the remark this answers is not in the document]
jacob-carlborg: I guess the rational is that this is `extern(C)` which would mangle to the same function.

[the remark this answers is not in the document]
JohanEngelen: argh... disregard my comments on this piece. Perhaps explicitly add `extern(C)` in front of each function here? Makes it easier to read the testcases.

JohanEngelen:
@wilzbach Does this also work for multiply defined `extern(D)` functions? If so: changelog should say so too.

test/fail_compilation/test18385.d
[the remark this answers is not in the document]
kinke: I'm not convinced this avoid-explicit-casts use-case is worth the ugly inconsistent AST, and having to handle incompatible declarations, incl. casts, in the backend.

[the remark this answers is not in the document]
wilzbach: I'm not sure I follow. This PR just tries to enable a deprecation of being more restrictive than dmd is currently. It was originally intended for 2.079, but got disabled/reverted due to people already complaining about it.
I agree that the avoid-explicit-casts use-case is a pretty bad one and I would love to deprecate more ;-)

kinke:
> multiply defined extern(D) functions

For illustration, this compiles and runs successfully:
```D
int foo() { return 0; }
int foo() { return 1; }

int main() {
    void* p = &foo;
    return (cast(int function()) p)();
}
```

DMD emits a foo() function containing both bodies, see https://run.dlang.io/is/JfK9YH.

Geod24:
```
Up to      2.071.2: Failure with output:
-----
onlineapp.o: In function `_D9onlineapp3fooFZi':
/sandbox/onlineapp.d:2: multiple definition of `_D9onlineapp3fooFZi'
onlineapp.o:onlineapp.d:(.text._D9onlineapp3fooFZi+0x0): first defined here
--- errorlevel 1
-----

2.072.2 to 2.074.1: Failure with output:
-----
onlineapp.o: In function `_D9onlineapp3fooFZi':
/sandbox/onlineapp.d:2: multiple definition of `_D9onlineapp3fooFZi'
onlineapp.o:onlineapp.d:(.text._D9onlineapp3fooFZi+0x0): first defined here
-----

Since      2.075.1: Success with output: /usr/bin/ld: Warning: size of symbol `_D9onlineapp3fooFZi' changed from 8 in onlineapp.o to 11 in onlineapp.o
```

Looks like a regression :/

test/fail_compilation/test18385.d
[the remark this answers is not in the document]
kinke: Ah sorry I meant the test; I'd rather have it check that a compatible declaration (`void foo3(int);`) and later definition works, instead of testing that incompatible declarations are allowed.

MoonlightSentinel:
Seems like it wrongly flags duplicate declarations, see e.g. the `fail_compilation/bug19569.d` failure.

MoonlightSentinel:
Reduced the DWT failure to this test case:

```
extern(C)
{
	uint is_paragraph_start(){ return 0; }
	uint is_paragraph_start(){ return 0; }
}
```

jacob-carlborg:
> Reduced the DWT failure to this test case:
>
> ```d
> extern(C)
> {
>     uint is_paragraph_start(){ return 0; }
>     uint is_paragraph_start(){ return 0; }
> }
> ```

Isn't that what supposed to be deprecated? DWT does not contain that exact code. It contains:

```d
guint is_paragraph_start(){ return ( bitfield0 >> 0 ) & 0x01; }
guint is_paragraph_start( guint v ){ bitfield0 &= ~( 0x01 << 0 ); bitfield0 |= ( v & 0x01 ) << 0; return is_paragraph_start(); }
```

The file these functions live in is bindings to a C library. `extern (C):` has been added to the top of the module. Since D doesn't support bitfields, the above functions were added to a struct to handled the bit manipulation. It uses overloading to be able to call the functions with the same syntax as one would do in C, i.e. as a field. I can add `extern(D)` for these functions.

kinke:
>  I can add extern(D) for these functions.

Please do, there's no overloading for C functions, the mangled names are identical.

MoonlightSentinel:
> Isn't that what supposed to be deprecated?

Yes

> DWT does not contain that exact code. It contains:

Should've mentioned that it was a reduction found by `dustmite` which I've fixed up to be compilable again by current master.

> The file these functions live in is bindings to a C library. `extern (C):` has been added to the top of the module. Since D doesn't support bitfields, the above functions were added to a struct to handled the bit manipulation. It uses overloading to be able to call the functions with the same syntax as one would do in C, i.e. as a field. I can add `extern(D)` for these functions.

See @kinke's answer

jacob-carlborg:
> Please do

Will do.

> the mangled names are identical

These functions are part of a struct. Even if those are functions are `extern (C)` they will get D mangling.

kinke:
> These functions are part of a struct. Even if those are functions are extern (C) they will get D mangling.

Brrr... D mangling but using the incompatible C calling convention at the same time...

jacob-carlborg:
> there's no overloading for C functions

This is D, not C. See one of my other posts: https://github.com/dlang/dmd/pull/8429#issuecomment-401703566.

jacob-carlborg:
The deprecations in DWT has been fixed in: https://github.com/d-widget-toolkit/dwt/releases/tag/v1.0.5%2Bswt-3.4.1.

wilzbach:
> SmartUnion generates conflicting opCalls sociomantic-tsunami/ocean#827

FYI: Ocean is no longer maintained by Sociomantic as the company shutdown (https://forum.dlang.org/post/r85v83$72k$1@digitalmars.com). So we might need to remove it from Buildkite.

jacob-carlborg:
> So we might need to remove it from Buildkite.

Unless someone wants to maintain it for the purpose of increasing surface area for the test suite.

MoonlightSentinel:
> FYI: Ocean is no longer maintained by Sociomantic as the company shutdown (https://forum.dlang.org/post/r85v83$72k$1@digitalmars.com). So we might need to remove it from Buildkite.

I would rather implement a crude fix because it apparently contains code not present in other projects.

kinke:
FWIW, I've been waiting for almost a year now for a new Ocean release, blocking https://github.com/dlang/dmd/pull/10124#issuecomment-625373405.

MoonlightSentinel:
@wilzbach Could you rebase the branch? I cannot do that because it introduces the GitHub workflow for `runnable_cxx`:
```
 ! [remote rejected]     duplicate-declaration -> duplicate-declaration (refusing to allow an OAuth App to create or update workflow `.github/workflows/runnable_cxx.yml` without `workflow` scope)
```

MoonlightSentinel:
Fixed the DWT issue.

ibuclaw:
> > SmartUnion generates conflicting opCalls sociomantic-tsunami/ocean#827
> 
> FYI: Ocean is no longer maintained by Sociomantic as the company shutdown (https://forum.dlang.org/post/r85v83$72k$1@digitalmars.com). So we might need to remove it from Buildkite.

Nooooooooooqqqqqqqq

Something is happening, please hold on. :-)

MoonlightSentinel:
Swarm got a new release so this isn't blocked any longer (thanks @Geod24 !).

This PR currently includes some changes/bugfixes for the original implementation:

- Don't issue errors for mutliple declarations
- Properly handle `extern(C)` functions
  - reject identication function definitions
  - allow different parameters/storage classes for now (but issue a deprecation)
- Remove dead code (partially caused by the previous changes)

Can also defer most of them to another PR if preferred.

MoonlightSentinel:
> @wilzbach Could you rebase the branch? I cannot do that because it introduces the GitHub workflow for `runnable_cxx`:
> 
> ```
>  ! [remote rejected]     duplicate-declaration -> duplicate-declaration (refusing to allow an OAuth App to create or update workflow `.github/workflows/runnable_cxx.yml` without `workflow` scope)
> ```

Same issue prevents another rebase to include `frontend.h`...

wilzbach:
> @wilzbach Could you rebase the branch? I cannot do that because it introduces the GitHub workflow for `runnable_cxx`:

Done.

MoonlightSentinel:
Thanks!

ibuclaw (approved):
I guess it won't supplant the existing checks that gdc (and I assume ldc too for the same reason) in the code-gen layer that look for duplication of mangling, and issue a duplicate definition error.

thewilsonator (approved):

changelog/duplicate-implementations-deprecation.dd
```diff
@@ -0,0 +1,26 @@
+Usage of identical functions in a single module
```
Geod24, line 1:
```suggestion
A deprecation is now issued on multiple definitions of identical functions within the same module
```

Rationale: Usually we say "<Feature> has be deprecated" or "Deprecated the ability to <feature>".
However, the previous code was buggy, hence the formulation "A deprecation is now issued"

changelog/duplicate-implementations-deprecation.dd
```diff
@@ -0,0 +1,26 @@
...
+
+A binary should never contain multiple versions of a symbol.
```
Geod24, line 3: It's not that it shouldn't, it simply **cannot**. Also this is a weird sentence to start the longer explanation with.

changelog/duplicate-implementations-deprecation.dd
```diff
@@ -0,0 +1,26 @@
...
+Before this version, it was allowed to provide multiple implementation for a function,
+even if they result in the same mangling, which can result in undefined behaviour:
```
Geod24, line 5: Instead of saying it was "allowed", I would say that it was either "not detected", or similar.
Not a native speaker, but allowed could imply intent, while it was really a missing diagnostic.

My litmus test for this is: "Does this deprecation prevent me from doing something I was able to do before", and here's the answer clearly no.

changelog/duplicate-implementations-deprecation.dd
```diff
@@ -0,0 +1,26 @@
...
+extern(C):
+void foo(int) { }
+void foo(double) { } // will issue a deprecation now
+
+void bar() {}
+void bar() {} // will raise an error
+---
+
+However, it's still allowed to have multiple declarations as long as there's at
+most one implementation:
```
Geod24, line 17:
```suggestion
However, conflicting function declarations are still allowed:
```

~What's the reasoning for this ?~

Looking at the diff, it seems to be only for attributes. Maybe we should restrict it to just attributes, then ?

changelog/duplicate-implementations-deprecation.dd
```diff
@@ -0,0 +1,26 @@
...
+
+---
+extern(C):
+void foo(int) { }
+void foo(double);
+---
+
+Note that this change is only relevant for mangling schemes that do no support
+overloading like `extern(C)`.
```
Geod24, line 26:
```suggestion
Note that this change is only relevant for `extern(C)`, `extern(System)`, and `extern(Windows)` functions,
as other scheme already detected duplicates.
```

Better to enumerate than to let the user guess.

src/dmd/semantic2.d
```diff
@@ -393,35 +395,29 @@ private extern(C++) final class Semantic2Visitor : Visitor
...
-                     *  sigfn_t2 bsd_signal(int sig, sigfn_t2 func) nothrow @nogc;  // no error
-                     */
-                    if (f1.fbody is null || f2.fbody is null)
-                        return 0;
-
-                    auto tf2 = cast(TypeFunction)f2.type;
-                    error(f2.loc, "%s `%s%s` cannot be overloaded with %s`extern(%s)` function at %s",
+                    // @@@DEPRECATED_2.094@@@
+                    // Deprecated in 2020-08, make this an error in 2.104
+                    deprecation(f2.loc, "%s `%s%s` cannot be overloaded with %s`extern(%s)` function at %s",
```
Geod24, line 411: 1) Shouldn't we call  `f2.deprecation`  ?
2) It's a weird phrasing: f2 cannot be overloaded by f1 (which is first in lexical order). Isn't it backwards ?
     I would expect something like "<func name> cannot overload `extern (C)` `function` `<FQN>` defined at `%s`"

test/fail_compilation/fail2789.d
```diff
@@ -60,9 +61,9 @@ auto f6() { return ""; }    // string(), conflict
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/fail2789.d(67): Error: function `fail2789.f_ExternC1()` cannot be overloaded with another `extern(C)` function at fail_compilation/fail2789.d(66)
-fail_compilation/fail2789.d(70): Error: function `fail2789.f_ExternC2(int)` cannot be overloaded with another `extern(C)` function at fail_compilation/fail2789.d(69)
-fail_compilation/fail2789.d(73): Error: function `fail2789.f_ExternC3()` cannot be overloaded with another `extern(C)` function at fail_compilation/fail2789.d(72)
+fail_compilation/fail2789.d(67): Error: function `fail2789.f_ExternC1()` conflicts with previous declaration at fail_compilation\fail2789.d(66)
```
Geod24, line 64: Side note, why isn't the file name / line highlighted ?

Geod24 (changes requested):
Nice, few comments

test/fail_compilation/test18385.d
```diff
@@ -0,0 +1,31 @@
...
+
+extern (C):
+
+void foo(int) { }
+void foo(double) { }
+
+struct S
+{
+    static void foo(int) {}
+    static void foo(double) {}
```
jacob-carlborg, line 18: Please remind me why this shouldn't be allowed? These two symbols are part of a struct and therefore will get mangled as D symbols, even though the linkage is `extern (C)`.

The changelog entry says: "Note that this change is only relevant for mangling schemes that do no support
 overloading". The D mangling scheme **does** support overloading. Make up your minds.

MoonlightSentinel:
@Geod24 Rewrote the changelog and improved the deprecation message.

test/fail_compilation/test18385.d
MoonlightSentinel: As discussed above, the current behaviour of `extern(C)` not changing the mangling inside of a struct is a bug because the function still adheres to the C ABI.

The current implementation could cater for that special case but I don't think that it should. Having the deprecation now will allow for a smoother transition when fixing `extern(C)` inside `struct`s.

test/fail_compilation/fail2789.d
Geod24: Ping @MoonlightSentinel . And why the backslash ?

Geod24 (approved):

BorisCarvajal:
This pull destroyed compiler diagnostic for conflicts in overload sets created by aliases.

```
void bar (int s) {}
void foo (int s) {}
alias bar = foo;

void main ()
{
   bar(0);
}
```
before:
```
test.d(7): Error: test.foo called with argument types (int) matches both:
test.d(1):     test.foo(int s)
and:
test.d(2):     test.bar(int s)
```
now:
`test.d(2): Error: function test.bar(int s) conflicts with previous declaration at test.d(1)`

rainers:
When compiling Visual D with 2.095-beta1 I get a lot of deprecation messages due to this change:
- using std.bitmanip.bitfields with structs compiled as "extern(C)" (from translated Win/VS SDK headers) complain:
```
import std.bitmanip;
extern(C) struct S
{
	mixin(bitfields!(uint,"cloneGroupId", 16 , uint,"sourceModeInfoIdx", 16));
}
bit.d-mixin-6(7): Deprecation: function bit.S.cloneGroupId cannot overload extern(C) function at bit.d-mixin-6(6)
bit.d-mixin-6(9): Deprecation: function bit.S.sourceModeInfoIdx cannot overload extern(C) function at bit.d-mixin-6(8)
```
- classes implementing IUnknown are automatically extern(Windows), and every overload is now deprecated, even for ctors:
```
interface IUnknown {}
class C : IUnknown
{
	this() {}
	this(int) {}
	
	void fun(int) {}
	void fun(string) {}
}
bit.d(5): Deprecation: constructor bit.C.this cannot overload extern(Windows) function at bit.d(4)
bit.d(8): Deprecation: function bit.C.fun cannot overload extern(Windows) function at bit.d(7)
```
Is this expected even for member functions?

jacob-carlborg:
> When compiling Visual D with 2.095-beta1 I get a lot of deprecation messages due to this change:

I have the same problem with some of my projects. The reason for the change was to remove the mangling of `extern(C)` methods and make the transition smoother: https://github.com/dlang/dmd/pull/8429#discussion_r492853471.

But then everyone changed their mind when it comes to remove the actual mangling: https://github.com/dlang/dmd/pull/11786.

I don't know what's going on here, it's a mess.

jacob-carlborg:
> Is this expected even for non-static member functions?

The code is probably just looking at the linkage and not the actual mangling.

MoonlightSentinel:
I'll have a look

rainers:
I'm fine with the change for functions with identical mangling. So the change is good for free functions, not so sure about static member functions (with changed mangling), but non-static member functions are not a thing in C, so don't have to obey the same ABI anyway.

MoonlightSentinel:
> This pull destroyed compiler diagnostic for conflicts in overload sets created by aliases.

The false positive hides the actual error. Fix #12053 

> But then everyone changed their mind when it comes to remove the actual mangling: #11786

Fix to accept the name mangling for (static) member funtions: #12054

ibuclaw:
This PR caused a regression: https://issues.dlang.org/show_bug.cgi?id=22975

PR #8437, opened 2018-07-02
Fixed issue 11934. Range with a non-ref front + foreach(ref...)

trikko:
Patch mostly written by ketmar.


test/runnable/test11934.d
```diff
@@ -0,0 +1,25 @@
...
+    {
+        void popFront() { cnt++; }
+        @property front() { return Struct11934(0); }
+        @property empty() { return cnt >= 10; }
+        size_t cnt;
+    }
+
+    foreach(ref i; Range11934()) { }
+
+    GC.collect();
```
wilzbach: There should be no need for calling the `GC.collect` here. `struct` are allocated on the stack and destroyed deterministically once you leave a scope (i.e. after the foreach).

wilzbach (dismissed):
Mind the CI warnings about the trailing whitespace.

Geod24 (commented):
Would be nice to use a [more descriptive commit message](https://chris.beams.io/posts/git-commit/)

trikko:
@wilzbach is it ok now?

test/runnable/test11934.d
```diff
@@ -0,0 +1,22 @@
...
+    {
+        void popFront() { cnt++; }
+        @property front() { return Struct11934(0); }
+        @property empty() { return cnt >= 10; }
+        size_t cnt;
+    }
+
+    foreach(ref i; Range11934()) { }
+
+	assert(Struct11934.instances == 0);
```
wilzbach: Maybe have three global counters to ensure the {con,de,copy}structors are actually called.

src/dmd/statementsem.d
```diff
@@ -1454,13 +1454,54 @@ private extern (C++) final class StatementSemanticVisitor : Visitor
...
+                    goto Lrangeerr;
+                if (tfront.toBasetype().ty == Tfunction)
+                {
+                    auto ftt = cast(TypeFunction)tfront.toBasetype();
+                    tfront = tfront.toBasetype().nextOf();
+                    if (!ftt.isref)
+                    {
+                        // .front() does not return a ref. We ignore ref on foreach arg.
+                        // see https://issues.dlang.org/show_bug.cgi?id=11934
+                        if (tfront.needsDestruction()) ignoreRef = true;
```
wilzbach: I thought the point of the issue was to error (or `deprecate`) on `ref` if `front` doesn't return by `ref`?

trikko: It seems that consensus is to simply allow silently it. Probably returning an error is going to break a lot of code and to complicate a lot of code with many corner cases (consider f.e. "each" template on Phobos).

test/runnable/test11934.d
trikko: Imo it's not the point of the test for this issue, but I can do it easily of course.

src/dmd/statementsem.d
wilzbach: Hmm did you check how much code in Phobos would be affected?
As long as it only triggers a deprecation, we don't break _any_ code and this seems like one of those silent things that you really want your compiler to warn you about.

trikko: Yes error will brake code but deprecation as said is going to complicate code with many corner cases. Give a look at "each" code. You probably have to duplicate the just-duplicated code. Moreover every single time you use foreach+ref with a range you have to check if front is a function that returns a ref and write a special case for it. It doesn't sound good to me. Anyway I can check deprecation against Phobos unit tests.

test/runnable/test11934.d
trikko: In this example postblit is never called but counter is there to make clear we're not missing any check.

src/dmd/statementsem.d
trikko: Actually phobos unit tests doesn't give any deprecation because - apparently - test aren't done on range returning a struct as front (at least for "each" test).

So I can add a deprecation about this but both Ketmar and @Geod24 told me to be permissive instead. 

What to do?

wilzbach: I guess it's best to open a thread in D.general about this. Imho if we can fix this properly without huge walls of deprecation messages, we should do so.
I also sent out an invitation to our D slack where we sometimes have discussions in #dmd too.

trikko: On D.internals ketmar answer my question stone days ago. On slack geod24 said he agreed. I tried to open a discussion on general but I don't get so much attention. Anyway Steven Schveighoffer seemed to agree too.

See https://forum.dlang.org/post/pg0n5d$110$1@digitalmars.com

https://forum.dlang.org/post/b6f3b32debd34cc8883eec18827758d3@chiroptera

trikko:
ping @wilzbach

rikkimax:
I do not like when the compiler goes around silently "fixing" my code. No, this is a problem with the spec (corner case). Let's deprecate it, have it patch the problem for now and change the spec to reflect it.

It should result in more code, and possibly breaking some. That is why deprecation with a fix now is a good stepping stone.

andralex (approved):
I think it's important for many parts of the language to not naively bind references to rvalues. In `foreach` the most obvious ill effect is that people believe they change elements of a range but they don't. cc @WalterBright and @RazvanN7 for reviewing implementation.

andralex:
please take a look at the autotester failure

src/dmd/statementsem.d
```diff
@@ -1454,13 +1454,54 @@ private extern (C++) final class StatementSemanticVisitor : Visitor
...
+                }
+
                 if (dim == 1)
                 {
                     auto p = (*fs.parameters)[0];
                     auto ve = new VarDeclaration(loc, p.type, p.ident, new ExpInitializer(loc, einit));
                     ve.storage_class |= STC.foreach_;
                     ve.storage_class |= p.storageClass & (STC.in_ | STC.out_ | STC.ref_ | STC.TYPECTOR);
-
+
```
wilzbach: Error - file 'dmd/statementsem.d' contains trailing whitespace at line 1501

RazvanN7 (dismissed):
I am against this solution. I think that the right path is to deprecate the use of foreach(ref..) if range.front does not return a reference.

Imagine this scenario: you have a library defined range with a front that returns by value; a user wants to iterate over the range elements and modify them by using foreach(ref item; range). Here the intention is clearly to modify the elements, but the compiler does not hint that that's not possible and this results in unexpected behavior.

trikko:
@RazvanN7 That was my first idea. But it was rejected.

For example @schveiguy said: "At the very least, the code should allocate a temporary and assign a ref to that, so lifetime management works properly. I agree the better path is to simply disallow ref, but this would be a non-code-breaking step that would fix the issue of lifetime management." 

I think that would break a lot of code. And you need to create a special case (in phobos or your own code) to static check if element type is a ref or not. Every. Single. Time.

RazvanN7:
@andralex Are you approving the current fix which is to silently ignore ref or a deprecation/error if the front does not return a ref?

trikko:
Consider that with this patch foreach(ref x; range) means  "iterate thru range reading elements as ref if possible". And consider that is exactly what D is doing now (but it doesn't free temporary elements!).

If you instead mean "iterate thru range by ref only if front() is ref" you're changing current behaviour.

schveiguy:
So I'm not sure what the current patch is doing, or how it's doing it (I'm super-ignorant of dmd code), but my position is:

1. There is a lifetime issue with the current code -- if you foreach w/ ref on something that doesn't return by ref, if the compiler is going to allow it, it should properly manage the lifetime of the temporary. My understanding (from testing and the report) is that it doesn't call a destructor for the temporary.
2. foreach with ref should really only bind to something that returns ref. This may break a lot of code, but the consequences of leaving it the way it is are horrible. If you are expecting to affect the elements, you will not, and that's not acceptable IMO.

So I think we should patch dmd with a *bug fix* that fixes the lifetime issues right now. I also think we should deprecate the current behavior, and only allow ref foreach on things that can actually be ref.

I'm not sure what this patch does, but it seems from the discussion it is only doing the quick bug fix. In that case, this is NOT fixing issue 11934, but some other issue (is there a bugzilla for it?).

trikko:
@RazvanN7 why can't we simply keep current behaviour? foreach(ref x; range) would mean "iterate range by ref if possible". 

The behaviuor you're proposing make really hard to write generic code that use foreach.

I think it's up to you to know if range you're iterating a range that returns by ref or not. Just like when you use its "front()" method. For example (short and dirty example):

https://run.dlang.io/is/D8DYOc

How do you know here if front is returning by ref or not? I think you should check range documentation. And the same goes for foreach. If you use foreach(ref ...) with a non-ref range, it is lowered to a common foreach, it's up to you to check if front() return ref or not.

schveiguy:
@RazvanN7 I agree with you, but at the same time, this is a breaking change, requiring a deprecation period. While the deprecation period is happening, I think the lifetime issues should be fixed.

Whether it's done via a hidden temporary that you reference, or removal of ref (I think semantically it should be identical), fixing the lifetime problem should be done first, as the deprecation and error will be years away.

> why can't we simply keep current behaviour? foreach(ref x; range) would mean "iterate range by ref if possible".

The problem is this:

```D
foreach(ref item; someRange)
{
    item.x += 5;
}
writeln(someRange); // wait, nothing happened? WTF???!
```

This code is broken. It should be trivial for the compiler to tell you "you know, that ref really doesn't reference anything". It's not a matter of using ref as some optimization, it's a matter of what the intention of the author wants. Right now, using ref means, I want to refer to the original. If it's not possible, it should be discouraged (error).

trikko:
> This code is broken. It should be trivial for the compiler to tell you "you know, that ref really doesn't reference anything". It's not a matter of using ref as some optimization, it's a matter of what the intention of the author was. Right now, using ref means, I want to refer to the original. If it's not possible, it should be discouraged (error).

@schveiguy that code is broken anyway. It doesn't print the updated version of item, but the original one since someRange is copied and then iterated (you're editing the copied version). So nothing happen wtf?! anyway!

schveiguy:
> but the original one since someRange is copied and then iterated

But the range is not the thing that make a difference, it's the items. A range is perfectly able to ref elements even if you copy it.

For example:

```D
struct S
{
   int x;
}

S[] arr = iota(10).map!(x => S(x)).array;

foreach(ref item; arr) // ok!
{
    item.x += 5; // affects original elements
}

foreach(ref item; arr.map!(x => x)) // now items are copies
{
    item.x += 5; // no effect!
}
```

Obviously the example is made up, nobody's going to do `map!(x => x)`, but at some point you may come across this kind of thing. It would be nice for the compiler to let you know something is wrong.

trikko:
I just said that your previous example with a struct-based range (with ref front) won't work, I guess. So probably it's not the right example to show here.

schveiguy:
> I just said that your previous example with a struct-based range

map provides a struct-based range? I'm not sure what you mean.

trikko:
I mean: https://run.dlang.io/is/dj95Md
I think user should be smart enough to understand what is going to happen.

RazvanN7:
@trikko When ref is used (for function parameters or foreach variables), it is explicitly intended that the modifications are propagated. If front is non-ref then the ref in foreach is useless and the user must be informed about that; the solution would be to simply delete ref from foreach.

It's not about if the user is smart enough or not to understand what's happening, it's more about people naturally expecting the language to be self-consistent. Treating ref like this patch does results in another exception from the rule which is easy to overlook and forget and might result in lost time trying to figure out why your ref foreach variable does not actually modify the element.

schveiguy:
> I mean: https://run.dlang.io/is/dj95Md
> I think user should be smart enough to understand what is going to happen.

Sure, but *that* range returning `front` by `ref` is useless. Try using it in any std.algorithm, it's going to be copied, and the `ref` is misleading. In this case, the range author is the one who made the error -- it shouldn't have `ref` on `front`.

There is no rule about this, but ranges are expected to be copied wholesale and do exactly what the original does. Only the iteration state is duplicated, not the data itself, especially if `front` returns by `ref`. If copying a range means it refers to different items by reference, you will not be able to use that range nearly anywhere. When in doubt, see what a slice does.

Bolpat:
@trikko wrote:
> Consider that with this patch `foreach(ref x; range)` means "iterate thru range reading elements by ref if possible".

The main problem is: `ref` does not mean "if possible" anywhere else (except after [DIP1016](https://github.com/dlang/DIPs/blob/master/DIPs/DIP1016.md)). The best way to express "`ref` if possible" is `auto ref`. While `auto ref` on function parameters is limited to templates, it need not be for `foreach` loops, so `foreach(auto ref element; range)` basically is this:
```d
static if (hasLvalueElements!(typeof(range)))
{
    foreach (/*actual*/ ref element; range) body();
}
else
{
    foreach (element; range) body();
}
```
This can be done in non-template functions, too. With `auto ref`, you don't need to spell out this.

Furthermore, the [spec about the `foreach` statement](https://dlang.org/spec/statement.html#foreach-statement) says:
>  2. [...] If the variable is `ref`, it is a reference to the contents of that aggregate.

The fact that is is not is a compiler bug. Nowhere does the spec say, `ref` is okay for any variable nor does it anywhere define `ref` as binding a reference *when possible*. Fixing a compiler bug is by definition not change, therefore not a breaking change -- code that won't compile afterwards *was already broken* according to the spec. The spec does not explicitly say, it's illegal for the compiler to accept `ref` when referencing is not possible, but typically, if the spec cannot be satisfied, the compiler must refuse to compile.

look-at-me:
6+ months and nothing? Shouldn't this functionality be deprecated.

@WalterBright

s-ludwig:
Isn't this quite simple? Fixing the failure to call the destructor is critical and a clear improvement over the status quo. The `ref` foreach didn't ever do something useful for r-value ranges, but that's an orthogonal issue that should simply be discussed in the context of a separate ticket (or DIP).

PR #8438, opened 2018-07-02
Change deprecation to error for C-style array syntax

JinShil:
Documentation update:  https://github.com/dlang/dlang.org/pull/2401

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -5030,10 +5030,7 @@ final class Parser(AST) : Lexer
 
         const(char)* sp = !ident ? "" : " ";
         const(char)* s = !ident ? "" : ident.toChars();
-        if (alt & 1) // contains C-style function pointer syntax
-            error(loc, "instead of C-style syntax, use D-style `%s%s%s`", t.toChars(), sp, s);
-        else
-           dmd.errors.deprecation(loc, "instead of C-style syntax, use D-style syntax `%s%s%s`", t.toChars(), sp, s);
+        error(loc, "instead of C-style syntax, use D-style `%s%s%s`", t.toChars(), sp, s);
     }
 
     /*****************************************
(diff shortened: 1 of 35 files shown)


src/dmd/parse.d
```diff
@@ -5030,10 +5030,8 @@ final class Parser(AST) : Lexer
 
         const(char)* sp = !ident ? "" : " ";
         const(char)* s = !ident ? "" : ident.toChars();
-        if (alt & 1) // contains C-style function pointer syntax
-            error(loc, "instead of C-style syntax, use D-style `%s%s%s`", t.toChars(), sp, s);
-        else
-           dmd.errors.deprecation(loc, "instead of C-style syntax, use D-style syntax `%s%s%s`", t.toChars(), sp, s);
+        // @@@DEPRECATED_2.087_REMOVE@@@
```
wilzbach: Do we plan to ever deprecate this?
It seems like a useful error message to keep around forever (it was very helpful for the C->D conversion).

wilzbach (approved):

src/dmd/parse.d
JinShil: Actually, I hope so.  I don't think we should spend logic detecting invalid C syntax just so we can display an error.  I'm hoping we can just rely on ordinary parsing errors.  But that's just my opinion.  I'll go along with whatever the team decides.

Geod24: I used to think the same as you @JinShil then I ported some C and C++ code to D. If the overhead isn't too much (it's usually a few lines), it's definitely worth it.

jacob-carlborg (approved):

wilzbach:
Fails due to:

```
---
void main() {
import std.stdio;
enum Color { red, blue, green };

int value[Color.max + 1] =
  [ Color.blue :6,
    Color.green:2,
    Color.red  :5 ];

}
---
__stdin.d(3): Deprecation: use `{ }` for an empty statement, not `;`
__stdin.d(5): Error: instead of C-style syntax, use D-style `int[Color.max + 1] value`
```

PR: https://github.com/dlang/dlang.org/pull/2402

PR #8439, opened 2018-07-02
convert compress.c to compress.d

wilzbach:


jacob-carlborg (approved):

wilzbach:
Looks like Windows is consistently failing:

```
backend.lib(cgobj.obj) : error LNK2019: unresolved external symbol "char * __cdecl id_compress(char *,int,unsigned __int64 *)" (?id_compress@@YAPEADPEADHPEA_K@Z) referenced in function "public: virtual unsigned __int64 __cdecl Obj::mangle(struct Symbol *,char *)" (?mangle@Obj@@UEAA_KPEAUSymbol@@PEAD@Z)
```

I tried the typical `extern(C)` mangling hack.

src/dmd/backend/compress.d
```diff
@@ -5,28 +5,20 @@
...
+import core.stdc.string;
 
 /****************************************
  * Find longest match of pattern[0..plen] in dict[0..dlen].
  * Returns:
  *      true if match found
  */
 
-static bool longest_match(char *dict, int dlen, char *pattern, int plen,
+private __gshared bool longest_match(char *dict, int dlen, char *pattern, int plen,
```
WalterBright: functions are not __gshared

src/dmd/backend/global.h
```diff
@@ -483,7 +483,7 @@ void rtlsym_reset();
 void rtlsym_term();
 
 // compress.c
-char *id_compress(char *id, int idlen, size_t *plen);
+extern "C" char *id_compress(char *id, int idlen, size_t *plen);
```
WalterBright: Need `{ }` after the `extern "C"`

WalterBright (changes requested):
see my comments

src/dmd/backend/global.h
wilzbach: BTW there are a few other files that don't use the `{ }`:

```
dmd/backend/os.c
646:extern "C" void * __cdecl _osfhnd[];

dmd/backend/type.h
200:extern "C" type *type_static_array(targ_size_t dim, type *tnext);
203:extern "C" type *type_function(tym_t tyf, type **ptypes, size_t nparams, bool variadic, type *tret);

tests/cxxfrontend.c
61:extern "C" int rt_init();
62:extern "C" void gc_disable();
82:extern "C" int rt_term();
83:extern "C" void gc_enable();
```

WalterBright (approved):

PR #8440, opened 2018-07-02
[RFC] Allow using if ! (condition)

wilzbach:
Timoses suggested allowing `if !(condition)` [on the NG recently](https://forum.dlang.org/post/tugdfxokfvoyfbjjujul@forum.dlang.org).
Instead of this:

```d
if (!(t1.ty == Tarray && t2.ty == Tarray && needsDirectEq(t1, t2))
```

This shorthand would be allowed:

```d
if !(t1.ty == Tarray && t2.ty == Tarray && needsDirectEq(t1, t2))
```

So in other words it allows to avoid this awkward `(!(`.

I know that this might be controversial and will require a DIP, but as its such a simple change, I thought I put the PR forward.
I won't have much time for a DIP on this, so if you like this sugar, please go forward!

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -5653,8 +5653,14 @@ final class Parser(AST) : Lexer
             {
                 AST.Parameter param = null;
                 AST.Expression condition;
+                bool negateIf;
 
                 nextToken();
+                if (token.value == TOK.not)
+                {
+                    negateIf = true;
+                    nextToken();
+                }
                 check(TOK.leftParentheses);
 
                 StorageClass storageClass = 0;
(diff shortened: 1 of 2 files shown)


JinShil:
Looks a little like a template instantiation.

wilzbach:
I'm closing this as I just wanted to show that adding support for `if !(cond)` would be pretty easy implementation-wise.
As mentioned such a change definitely needs a DIP and I encourage everyone who is interested in this (or similar changes) to write one!

mcandre:
We can move further. Prefer `if <boolean expression> { ... } [else ...]` like Go.

PR #8442, opened 2018-07-03
rename md5.d to md5.di

WalterBright:
Because there's an md5.d implementation file coming.


jacob-carlborg (approved):

JinShil:
Just out of curiosity, why can't the .d file serve as both the implementation and the header?

WalterBright:
> Just out of curiosity, why can't the .d file serve as both the implementation and the header?

It can. But it slows down compilation, as every file that imports it will have to slog through it. .di files are also a more obvious way of hiding implementation details.

WalterBright:
@CyberShadow sounds like the documentation tester can't deal with .di files?

wilzbach:
For reference, the full failing command is:

```
find ../dmd/src -name '*.d' > .generated/.prerelease-files.txt
find ../druntime/src -name '*.d' | \
	sed -e /unittest/d >> .generated/.prerelease-files.txt
find ../phobos/etc ../phobos/std -name '*.d' | \
	sed -e /unittest.d/d | sort >> .generated/.prerelease-files.txt
../dmd/generated/linux/release/64/dmd -J../dmd/res -J../dmd/generated/linux/release/64/ -c -o- -version=MARS -version=CoreDdoc \
	-version=StdDdoc -Df.generated/.prerelease-dummy.html \
	-Xf.generated/docs-prerelease.json -I../phobos @.generated/.prerelease-files.txt
../druntime/src/core/simd.d(53): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(54): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(55): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(56): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(57): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(58): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(59): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(60): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(61): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(62): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../druntime/src/core/simd.d(63): Deprecation: 32 byte vector types are only supported with -mcpu=avx
../dmd/src/dmd/e2ir.d(5809): Error: module `md5` is in file 'dmd/backend/md5.d' which cannot be read
```

A quick workaround would be to keep an empty `md5.d` around.

WalterBright:
Trying @wilzbach 's workaround...

WalterBright:
@wilzbach 's workaround didn't work.

CyberShadow:
From a glance, it looks like the DMD source structure (specifically, its property that the module names do not map to filesystem paths, i.e. the same issue requiring `-mv`) isn't compatible with documentation generation + `.di` files.

What's preventing moving the source files to match the module hierarchy?

CyberShadow:
Changing the dlang.org makefile to glob for `.di` files too would probably fix it.

wilzbach:
> Changing the dlang.org makefile to glob for .di files too would probably fix it.

-> https://github.com/dlang/dlang.org/pull/2405

WalterBright:
> What's preventing moving the source files to match the module hierarchy?

The backend code is also used by DMC++, which has a different hierarchy and file organization.

PR #8444, opened 2018-07-03
Bump the version for AUTO_BOOTSTRAPPING to 2.079.1

wilzbach:
All other CIs have been upgraded to 2.079.1 (or newer).
That's why https://github.com/dlang/dmd/pull/8412 failed.
However, it seems that the AUTO_BOOTSTRAPPING=1 is no longer directly tested here.
Thus, this is required to make the other CIs happy (e.g. https://github.com/dlang/dlang.org/pull/2405)


jacob-carlborg (approved):

ibuclaw:
Don't you mean 2.076 or newer?

stefan-koch-sociomantic:
I am impressed ....

PR #8447, opened 2018-07-03
convert md5.c to D

WalterBright:


src/posix.mak
```diff
@@ -397,7 +397,7 @@ BACK_SRC = \
 	$C/machobj.c $C/mscoffobj.c \
 	$C/xmm.h $C/obj.h $C/pdata.c $C/cv8.c $C/backconfig.c $C/divcoeff.d \
 	$C/varstats.c $C/varstats.h $C/dvec.d \
-	$C/md5.c $C/md5.h \
```
wilzbach: You still need to change `BACK_OBJS` + the VisualD filters.
BTW if you want to, we can do globbing over this list, s.t. it doesn't permanently need to be updated.

WalterBright: I fixed the vcproj filters.

WalterBright:
@CyberShadow : 
```
Error: module `dmd.backend.md5` from file ../dmd/src/dmd/backend/md5.di conflicts with another module md5 from file ../dmd/src/dmd/backend/md5.d
```
That's the point. They're supposed to coexist in the same location.

CyberShadow:
@wilzbach We should probably move at least enumeration to the DMD makefile, so things could be updated as needed here.

WalterBright:
@CyberShadow @wilzbach Can you update what DAutoTest uses to build to this will pass? Or should I give up on .di files?

CyberShadow:
https://github.com/dlang/dlang.org/pull/2420 is merged, hopefully this should now work too.

joakim-noah:
Needs to be updated, `posix.mak` cannot be automatically merged on the auto-tester anymore.

andralex (approved):

PR #8449, opened 2018-07-03
Issue 18407 - debug should escape nothrow

wilzbach:
References:
- https://github.com/dlang/dmd/pull/7882
- https://github.com/dlang/dlang.org/pull/2209

I wasn't entirely sure on the best way to do this.
The problem is that `blockexit` checker runs so late in semantic3 that the DebugStatements are already flattened out.

I'm not proud of the current solution, so I'm more than open to better suggestions.

### DIFF
--- a/changelog/debug-nothrow.dd
+++ b/changelog/debug-nothrow.dd
@@ -1,0 +1,14 @@
+Exception-throwing code can now be used in debug blocks
+
+When writing debug code, one isn't interested statically ensuring that a function block doesn't `throw`,
+but a pleasant debugging experience as e.g. writing to the console or logfiles typically can't be `nothrow`.
+DMD already allowed to escape `pure` and `@nogc` within `debug` statement.
+With this release, exception-throwing code can be called from `debug` statements too:
+
+---
+import std.stdio;
+void main() nothrow
+{
+    debug writeln("Your throwing log statement.");
+}
+---
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -4841,6 +4841,7 @@ extern (C++) final class CallExp : UnaExp
     Expressions* arguments; // function arguments
     FuncDeclaration f;      // symbol to call
     bool directcall;        // true if a virtual call is devirtualized
+    bool inDebugStatement;  /// true if this was in a debug statement
     VarDeclaration vthis2;  // container for multi-context
 
     extern (D) this(const ref Loc loc, Expression e, Expressions* exps)
--- a/src/dmd/statement.d
+++ b/src/dmd/statement.d
@@ -1436,7 +1436,10 @@ extern (C++) final class ConditionalStatement : Statement
         {
             DebugCondition dc = condition.isDebugCondition();
             if (dc)
+            {
                 s = new DebugStatement(loc, ifbody);
+                debugThrowWalker(ifbody);
+            }
             else
                 s = ifbody;
         }
@@ -1452,6 +1455,37 @@ extern (C++) final class ConditionalStatement : Statement
     {
         v.visit(this);
     }
+}
+
+/**
+Marks all occurring ThrowStatements as internalThrows.
+This is intended to be called from a DebugStatement as it allows
+to mark all its nodes as nothrow.
+
+Params:
+    s = AST Node to traverse
+*/
+private void debugThrowWalker(Statement s)
+{
+
+    extern(C++) final class DebugWalker : SemanticTimeTransitiveVisitor
+    {
+        alias visit = SemanticTimeTransitiveVisitor.visit;
+    public:
+
+        override void visit(ThrowStatement s)
+        {
+            s.internalThrow = true;
+        }
+
+        override void visit(CallExp s)
+        {
+            s.inDebugStatement = true;
+        }
+    }
+
+    scope walker = new DebugWalker();
+    s.accept(walker);
 }
 
 /***********************************************************
--- a/test/compilable/test16492.d
+++ b/test/compilable/test16492.d
@@ -40,3 +40,48 @@ void unsafeDIP1000Lifetime()(ref char[] p, scope char[] s)
 {
     p = s;
 }
+
+
+void test2() nothrow
+{
+    debug throw new Exception("");
+}
+
+void test3() nothrow
+{
+    debug {
+        foreach (_; 0 .. 10) {
+            if (1) {
+                throw new Exception("");
+            }
+        }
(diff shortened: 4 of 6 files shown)


changelog/debug-nothrow.dd
```diff
@@ -0,0 +1,11 @@
+Unsafe code can now be used in nothrow blocks
```
jacob-carlborg: This isn’t related to safe code?

src/dmd/statement.d
```diff
@@ -1448,6 +1451,45 @@ extern (C++) final class ConditionalStatement : Statement
     }
 }
 
+/**
+Marks all occurring ThrowStatements as as internalThrows.
```
RazvanN7: double **as**

RazvanN7 (approved):
LGTM

src/dmd/statement.d
```diff
@@ -1448,6 +1451,45 @@ extern (C++) final class ConditionalStatement : Statement
...
+    public:
+
+        override void visit(ThrowStatement s)
+        {
+            s.internalThrow = true;
+        }
+
+        override void visit(TryCatchStatement s)
+        {
+            foreach (c; *s.catches)
```
RazvanN7: uncovered

src/dmd/statement.d
```diff
@@ -1448,6 +1451,45 @@ extern (C++) final class ConditionalStatement : Statement
...
+            {
+                c.internalCatch = true;
```
RazvanN7: uncovered

RazvanN7:
Actually, I just thought about something: what about anonymous functions that get called in debug statements?

```d
void foo() nothrow
{
    debug
    {
        () (throw new Exception("");)()
    }
}
```

Does this work with your patch? If, yes please add an example.

changelog/debug-nothrow.dd
```diff
@@ -0,0 +1,14 @@
...
+When writing debug code, one isn't interested in the type safety, but a pleasant
+debugging experience.
+The type checker already allowed to escape `pure` and `@nogc` within `debug`
+statement.
+With this release, `nothrow` code can be called from `debug` statements too:
+
+---
+void main()
+{
+    debug throw new Exception("");
```
joakim-noah: Nothing exceptional about this, did you mean to make the main `nothrow`?

changelog/debug-nothrow.dd
```diff
@@ -0,0 +1,14 @@
+Nothrow code can now be used in debug blocks
+
+When writing debug code, one isn't interested in the type safety, but a pleasant
```
jacob-carlborg: I don’t think `nothrow` has any thing to do with type safety. This is clearly a copy paste from the other PR for `@safe`.

jacob-carlborg (dismissed):

wilzbach:
BTW the following will be invalid after this PR.
I could add `inDebugStatement` to `NewExp` et.al. too, but I already feel like adding it to `CallExp` is wrong.

```d
void test6() nothrow
{
    static struct S
    {
        this(this)
        {
            throw new Exception("BOOM!");
        }
    }

    S    ss;

    // TOKassign
    ss = ss;

    // TOKconstruct
    S    ss2 = ss;
}

void test7() nothrow
{
    static struct S
    {
        this(int k)
        {
            throw new Exception("BOOM!");
        }
    }
    static class C
    {
        this(int k)
        {
            throw new Exception("BOOM!");
        }
    }
    debug S(10);
    debug new C(10);
}
```

src/dmd/mars.d
```diff
@@ -498,6 +498,14 @@ private int tryMain(size_t argc, const(char)** argv)
     Objc._init();
     builtin_init();
 
+    version(CRuntime_Microsoft)
+    {
+        import dmd.root.longdouble;
+        initFPU();
+    }
+    import dmd.root.ctfloat : CTFloat;
+    CTFloat.initialize();
```
Geod24: That doesn't look related ?

WalterBright:
My concern with this is that statements get replaced with other statements in the semantic analysis, and those flags may get lost.

changelog/debug-nothrow.dd
```diff
@@ -0,0 +1,14 @@
+Nothrow code can now be used in debug blocks
```
Geod24: > Code in `debug` blocks will now ignore `nothrow`

Or something related. `nothrow` code can already be used in `debug` blocks :)

wilzbach: Fixed :O

wilzbach:
> My concern with this is that statements get replaced with other statements in the semantic analysis, and those flags may get lost.

Yes, good point, but I didn't know a better way to store this information as the `canthrow` checker runs late in the semantic3 analysis and thus the original `debug` statements have already been replaced.

andralex:
status?

wilzbach:
> status?

This is still ready, but @WalterBright wasn't fully happy with this.
As explained above, I don't know a better way to hack this into DMD as the information gets lost before we reach semantic3.

thewilsonator (approved):

jacob-carlborg (approved):

src/dmd/expression.d
```diff
@@ -4696,6 +4696,7 @@ extern (C++) final class CallExp : UnaExp
     Expressions* arguments; // function arguments
     FuncDeclaration f;      // symbol to call
     bool directcall;        // true if a virtual call is devirtualized
+    bool inDebugStatement;  // true if this was in a debug statement
```
ibuclaw: H...h...h...

ibuclaw (dismissed):

src/dmd/expression.d
WalterBright: I presume you mean fix the corresponding .h file.

quickfur:
ping @WalterBright 

It's been a *little* bit more than 72hrs now. ;-)  Any updates?

12345swordy:
It's green!

Geod24:
Amazing

test/compilable/test16492.d
```diff
@@ -40,3 +40,48 @@ void unsafeDIP1000Lifetime()(ref char[] p, scope char[] s)
...
+    debug {
+        foreach (_; 0 .. 10) {
+            if (1) {
+                throw new Exception("");
+            }
+        }
+    }
+}
+
+void test3() nothrow
```
ibuclaw, line 61: Broke the testsuite (function name conflict).

PR #8450, opened 2018-07-03
Fix Issue 19053 - debug should escape @safe

wilzbach:
- https://github.com/dlang/dmd/pull/7882
- https://github.com/dlang/dlang.org/pull/2209

### DIFF
--- a/changelog/debug-unsafe.dd
+++ b/changelog/debug-unsafe.dd
@@ -1,0 +1,21 @@
+Unsafe code can now be used in debug blocks
+
+When writing debug code, one isn't interested in the type safety, but a pleasant
+debugging experience.
+The type checker already allowed to escape `pure` and `@nogc` within `debug`
+statement.
+With this release, `@system` code can be called from `debug` statements too:
+
+---
+void main()
+{
+    int[] arr = [1, 2];
+    debug unsafeCode(arr);
+}
+
+@system void unsafeCode(T)(T[] arr)
+{
+    import core.stdc.stdio;
+    printf("arr.ptr: %p", arr.ptr);
+}
+---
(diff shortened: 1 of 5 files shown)


changelog/debug-unsafe.dd
```diff
@@ -0,0 +1,13 @@
...
+When writing debug code, one isn't interested in the type safety, but a pleasant
+debugging experience.
+The type checker already allowed to escape `pure` and `@nogc` within `debug`
+statement.
+With this release, `@system` code can be called from `debug` statements too:
+
+---
+debug unsafeCode();
+
+@system void unsafeCode(){}
```
Geod24: I would nest this inside a function for clarity.

Geod24:
Is `-dip1000` checked for in `-debug` ?

wilzbach:
> Is -dip1000 checked for in -debug ?

Well, -dip1000 can mean many things but as long as just `@safe` is concerned safety it should no longer be checked in `debug` too as while the inference will infer a method as unsafe / @system, it should still bypass `checkSafety`. I added an additional test.

test/compilable/test16492.d
```diff
@@ -13,3 +13,26 @@ void test() @nogc pure
...
+void unsafe() @system {}
+void unsafe2()() {
+    int[] arr;
+    auto b = arr.ptr;
+}
+
+void debugSafe2() @safe
+{
+    char[] arr1, arr2;
+    debug unsafe3(arr1, arr2);
```
Geod24: I meant something like:
```D
void debugSafe2() @safe
{
    char*ptr;
    char[] arr;
    debug ptr = arr.ptr;
}
```

wilzbach: Ah sorry. Yeah that was indeed still forbidden. Fixed.

test/compilable/test16492.d
```diff
@@ -13,3 +13,30 @@ void test() @nogc pure
...
+
+    char*ptr;
```
Geod24: Oops I was missing a space. Also, don't make functions needlessly templated.

RazvanN7 (approved):

Geod24:
@wilzbach : Has the specs been updated ?

PR #8451, opened 2018-07-04
Fix Issue 19059 - Disallow `08` and `09` in integer literal

kubo39:


JinShil:
If D no longer supports octals, I think literals beginning with 0 should evaluate to a decimal number.

aG0aep6G:
This PR outlaws `08` and `09`, but still allows `01` through `07`. That's still at odds with the spec (which doesn't allow `01` through `07`), and I can see no practical advantage in disallowing only `08` and `09`.

So I'm against this.

schveiguy:
The last comment concerns octal literals like `019`, which this still will have an odd error. I'll note that this fixes an inconsistency as well, as `009` errors while `09` does not.

I think instead of throwing an error on the new cases, just set the base to 8 and break (don't parse off the `8` or `9`, let the next loop do it).

Finally, you need to suppress the printing of the alternate for an invalid octal literal. e.g. the compiler currently says for `009`:
```
testoctal.d(7): Error: radix 8 digit expected, not 9
testoctal.d(7): Error: octal literals 011 are no longer supported, use std.conv.octal!11 instead
```

You may be able to change the error message to simply "octal literals are no longer supported" and don't offer a substitution when the original literal wasn't valid.

schveiguy:
> I can see no practical advantage

The practical advantage is that it doesn't break existing code that was valid prior to the deprecation of octal literals (and is still valid). Such changes would have to go through a deprecation period. IMO we should change the spec to allow octal literals less than 8, as they do no harm and are unambiguous. But I'm also OK with a deprecation and removal.

`08` and `09` (and even `0900`) were NEVER valid literals, it was a bug in the original parsing algorithm that they were ever accepted. This is fixing that bug.

aG0aep6G:
> IMO we should change the spec to allow octal literals less than 8, as they do no harm and are unambiguous.

`08` and `09` are also unambiguous. How are they doing more harm than `07`?

schveiguy:
`08` and `09` have *always been* invalid literals. The bug that allows these to parse is in the compiler.

`07` was not an invalid literal, and given the way it's parsed, it's not an invalid literal now. The bug for this is in the spec.

If we restored the grammar for octal literals, and said that only octal literals less than 8 are allowed, it fixes the problem and makes everything consistent. We are never going to allow octal literals to mean decimal numbers, as the confusion between C and D would be huge compared to the minimal benefit.

schveiguy:
Grammar was changed in https://github.com/dlang/dlang.org/pull/1079 (long after octal numbers were deprecated in version 2.053), we can revert the octal number grammar changes (which are actually in the lexer still, and need to be for displaying the nice "use std.conv.octal!123 instead" message), and then this PR becomes more palatable.

aG0aep6G:
> 08 and 09 have always been invalid literals. The bug that allows these to parse is in the compiler.
> 
> 07 was not an invalid literal, and given the way it's parsed, it's not an invalid literal now. The bug for this is in the spec.

According to run.dlang.io, DMD used to reject `08` up to 2.064. It always accepted `07`, of course. So the acceptance of `08` can be seen as a regression. I suppose it makes sense to disallow `08` and `09` but keep `07` when viewed this way.

So I can live with this PR, I guess.

I'm still not convinced that it's the best move, though. If we have to touch the spec and the compiler anyway, we might as well go for the best outcome instead of just fixing bugs by the letter.

> We are never going to allow octal literals to mean decimal numbers, as the confusion between C and D would be huge compared to the minimal benefit.

I don't see how the confusion is going to be huge if we allow `08` and `09` alongside `07`. They can't come up when porting from C to D, and their values are obvious.

I agree that larger values like `0800` should probably be rejected. They should follow the same rules as other invalid literals like `01238567`. And those should be rejected, because they look too similar to octals.

src/dmd/lexer.d
```diff
@@ -1824,6 +1824,13 @@ class Lexer : ErrorHandler
                 ++p;
                 base = 8;
                 break;
+            case '8':
+            case '9':
+                ++p;
+                base = 8;
+                error("radix %d digit expected, not `%c`", base, c);
+                n = n * base + (c - '0');
```
schveiguy: n is zero here, so all you need is
```D
n = c - '0'.
```

schveiguy: Make sure to set `err` to true also, so they don't get another error for other digits.

Essentially, I should get the same error messages for `088` as I do for `0088`

schveiguy: Looking at the code here, it's very convoluted. The error messages change not based on the base, but based on the digits.

Like you get a different error message for `0b5` than you do for `0b8`.

I think there are likely better ways to structure this whole thing. But maybe too much for this PR which is focused on the `08` problem.

schveiguy (approved):
I'm OK with this. Will work on a further PR to fix up the error messaging, which I think still is confusing.

PR #8453, opened 2018-07-05
Remove usage of initializerToExpression from parse.d

RazvanN7:
`initializerToExpression` does a lot of type processing (which requires dependencies used at semantic time) so there is no need to be used in the parser. This further separates the parser from the semantic methods and leads the way to getting rid of the duplication in ASTBase

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -6489,7 +6489,9 @@ final class Parser(AST) : Lexer
                     if (token.value == TOK.colon)
                     {
                         nextToken();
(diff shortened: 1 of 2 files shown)


src/dmd/parse.d
```diff
@@ -6488,8 +6488,11 @@ final class Parser(AST) : Lexer
                     value = parseInitializer();
                     if (token.value == TOK.colon)
                     {
+
```
wilzbach: unrelated

jacob-carlborg (approved):

PR #8455, opened 2018-07-05
Fix issue 16479: No namespace substitution for C++ mangling on POSIX

Geod24:
```
The C++ ABI used by POSIX is the Itanium C++ ABI.
Reference document available here: https://itanium-cxx-abi.github.io/cxx-abi/abi.html#mangling

One important (and tricky) part of the ABI is the substitutions being done,
in order to reduce the bloat introduced by long symbol names,
a typical issue when using templates heavily (of which D was not exempt).

There are 2 kinds of substitutions: component substitution and template parameter substitution.
Component substitution replaces repeated parts of the symbol with `S[X]_`,
template parameter substitution replaces occurences of template parameters with `T[X]_`.
`X` represents a base36 index into the array of components or
template parameters already encountered so far, respectively.

This substitution is done on an identity basis, which means that the templated function
`template<typename T> int foo()` instantiated with `int` will be mangled as `_Z3fooIiE*i*v`
(asterisks are emphasis and not part of the mangling) while it would be mangled as `_Z3fooIiE*T_*v`
if the definition was `template<typename T> T foo()`.

Moreover, experience with C++ compilers shows that component substitution is prefered over
template parameter substitution, such as `template<typename T> T foo(T)` is mangled as
`_Z3fooIiET_S0_` when instantiated with `int` and not `_Z3fooIiET_T_` as would be the case
if template substitution was prefered.

This is just brushing the surface of the problem, since only template type parameters have been
mentioned so far, but other kind (aliases, values) are also concerned.
Substitution also needs to happen if a template parameter is part of another type,
such as the `template<typename T> array<T>* foo (T, int)`, which, when instantiated with `int`,
is mangled as `_Z3fooIiEP5arrayIT_ES1_i`.

For more detailed test cases, see `test/compilable/cppmangle.d`.

The main issue encountered while implementing this in DMD is that there's no easy way to know
if a type (which is part of the function's type, e.g. parameters and return value)
was a template parameter or not, as DMD merges types, so in the previously mentioned
`template<typename T> int foo()` vs `template<typename T> T foo()` the template instantiation
will come with the same exact two pointer to the singleton `int` type.

Moreover, DMD does destructive semantic analysis, meaning that objects gets mutated,
pointers get replaced, aliases get resolved, and information gets lost.

After different approaches where taken, the most practical and reliable approach devised was to
provide a `visit` overload for non-resolved AST type `TypeIdentifier` and `TypeInstance`,
and compare the identifier to that of the template definition.
Fallback to post-semantic type when it isn't found.

Note that no attempt has been made whatsoever to handle the mess that would result from
expressions themselves being mangled. The reference doc for the ABI mentions that
"[...] this mangling is quite similar to the source token stream. (C++ Standard reference 14.5.5.1p5.)".
Original quote:
https://itanium-cxx-abi.github.io/cxx-abi/abi.html#expressions (5.1.6 Expressions)
```

~Please don't merge yet, I want to attempt to port some std types to D and see if I'm missing something. But I think this is ready for reviews.~

### DIFF
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -513,48 +513,16 @@ private final class CppMangleVisitor : Visitor
         {
             if (isStd(p))
             {
-                TemplateInstance ti = si.isTemplateInstance();
-                if (ti)
+                bool needsTa;
+                auto ti = si.isTemplateInstance();
+                if (this.writeStdSubstitution(ti, needsTa))
                 {
-                    if (s.ident == Id.allocator)
+                    if (needsTa)
                     {
-                        buf.writestring("Sa");
                         template_args(ti);
                         append(ti);
-                        return;
                     }
-                    if (s.ident == Id.basic_string)
-                    {
-                        // ::std::basic_string<char, ::std::char_traits<char>, ::std::allocator<char>>
-                        if (ti.tiargs.dim == 3 &&
-                            isChar((*ti.tiargs)[0]) &&
-                            isChar_traits_char((*ti.tiargs)[1]) &&
-                            isAllocator_char((*ti.tiargs)[2]))
-
-                        {
-                            buf.writestring("Ss");
-                            return;
-                        }
-                        buf.writestring("Sb");      // ::std::basic_string
-                        template_args(ti);
-                        append(ti);
-                        return;
-                    }
-
-                    // ::std::basic_istream<char, ::std::char_traits<char>>
-                    if (s.ident == Id.basic_istream &&
-                        char_std_char_traits_char(ti, "Si"))
-                        return;
-
-                    // ::std::basic_ostream<char, ::std::char_traits<char>>
-                    if (s.ident == Id.basic_ostream &&
-                        char_std_char_traits_char(ti, "So"))
-                        return;
-
-                    // ::std::basic_iostream<char, ::std::char_traits<char>>
-                    if (s.ident == Id.basic_iostream &&
-                        char_std_char_traits_char(ti, "Sd"))
-                        return;
+                    return;
                 }
                 buf.writestring("St");
             }
@@ -568,6 +536,65 @@ private final class CppMangleVisitor : Visitor
              * https://issues.dlang.org/show_bug.cgi?id=17947
              */
             append(si);
+    }
+
+    /**
+     * Write common substitution for standard types, such as std::allocator
+     *
+     * This function assumes that the symbol `ti` is in the namespace `std`.
+     *
+     * Params:
+     *   ti = Template instance to consider
+     *   needsTa = If this function returns `true`, this value indicates
+     *             if additional template argument mangling is needed
+     *
+     * Returns:
+     *   `true` if a special std symbol was found
+     */
+    bool writeStdSubstitution(TemplateInstance ti, out bool needsTa)
+    {
+        if (!ti)
+            return false;
+
+        if (ti.name == Id.allocator)
+        {
+            buf.writestring("Sa");
+            needsTa = true;
+            return true;
+        }
+        if (ti.name == Id.basic_string)
+        {
+            // ::std::basic_string<char, ::std::char_traits<char>, ::std::allocator<char>>
+            if (ti.tiargs.dim == 3 &&
+                isChar((*ti.tiargs)[0]) &&
+                isChar_traits_char((*ti.tiargs)[1]) &&
+                isAllocator_char((*ti.tiargs)[2]))
+
+            {
+                buf.writestring("Ss");
+                return true;
+            }
+            buf.writestring("Sb");      // ::std::basic_string
+            needsTa = true;
+            return true;
+        }
+
+        // ::std::basic_istream<char, ::std::char_traits<char>>
+        if (ti.name == Id.basic_istream &&
+            char_std_char_traits_char(ti, "Si"))
+            return true;
+
+        // ::std::basic_ostream<char, ::std::char_traits<char>>
+        if (ti.name == Id.basic_ostream &&
+            char_std_char_traits_char(ti, "So"))
+            return true;
+
+        // ::std::basic_iostream<char, ::std::char_traits<char>>
+        if (ti.name == Id.basic_iostream &&
+            char_std_char_traits_char(ti, "Sd"))
+            return true;
+
+        return false;
     }
 
 
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -101,8 +101,52 @@ bool isPrimaryDtor(const Dsymbol sym)
     return dtor == ad.primaryDtor;
 }
 
+/// Context used when processing pre-semantic AST
+private struct Context
+{
+    /// Template instance of the function being mangled
+    TemplateInstance ti;
+    /// Function declaration we're mangling
+    FuncDeclaration fd;
+    /// Current type / expression being processed (semantically analyzed)
+    RootObject res;
+
+    @disable ref Context opAssign(ref Context other);
+    @disable ref Context opAssign(Context other);
+
+    /**
+     * Helper function to track `res`
+     *
+     * Params:
+     *   next = Value to set `this.res` to.
+     *          If `this.res` is `null`, the expression is not evalutated.
+     *          This allow this code to be used even when no context is needed.
+     *
+     * Returns:
+     *   The previous state of this `Context` object
+     */
+    private Context push(lazy RootObject next)
+    {
+        auto r = this.res;
+        if (r !is null)
+            this.res = next;
+        return Context(this.ti, this.fd, r);
+    }
+
+    /**
+     * Reset the context to a previous one, making any adjustment necessary
+     */
+    private void pop(ref Context prev)
+    {
+        this.res = prev.res;
+    }
+}
+
 private final class CppMangleVisitor : Visitor
 {
+    /// Context used when processing pre-semantic AST
+    private Context context;
+
     Objects components;         // array of components available for substitution
     OutBuffer* buf;             // append the mangling to buf[]
     Loc loc;                    // location for use in error messages
@@ -139,6 +183,26 @@ private final class CppMangleVisitor : Visitor
         {
             assert(0);
         }
+    }
+
+    /**
+     * Mangle the return type of a function
+     *
+     * This is called on a templated function type.
+     * Context is set to the `FuncDeclaration`.
+     *
+     * Params:
+     *   preSemantic = the `FuncDeclaration`'s `originalType`
+     */
+    void mangleReturnType(TypeFunction preSemantic)
+    {
+        auto tf = cast(TypeFunction)this.context.res.asFuncDecl().type;
+        Type rt = preSemantic.nextOf();
+        if (tf.isref)
+            rt = rt.referenceTo();
+        auto prev = this.context.push(tf.nextOf());
+        scope (exit) this.context.pop(prev);
+        this.headOfType(rt);
     }
 
     /**
@@ -264,6 +328,17 @@ private final class CppMangleVisitor : Visitor
         TemplateParameter tp = (*td.parameters)[arg];
         RootObject o = (*ti.tiargs)[arg];
 
+        Objects* pctx;
+        auto prev = this.context.push({
+                TemplateInstance parentti;
+                if (this.context.res.dyncast() == DYNCAST.dsymbol)
+                    parentti = this.context.res.asFuncDecl().parent.isTemplateInstance();
+                else
+                    parentti = this.context.res.asType().toDsymbol(null).parent.isTemplateInstance();
+                return (*parentti.tiargs)[arg];
+            }());
+        scope (exit) this.context.pop(prev);
+
         if (tp.isTemplateTypeParameter())
         {
             Type t = isType(o);
@@ -296,20 +371,18 @@ private final class CppMangleVisitor : Visitor
         }
         else if (tp.isTemplateAliasParameter())
         {
+            // Passing a function as alias parameter is the same as passing
+            // `&function`
             Dsymbol d = isDsymbol(o);
             Expression e = isExpression(o);
             if (d && d.isFuncDeclaration())
             {
-                bool is_nested = d.toParent() &&
-                    !d.toParent().isModule() &&
-                    (cast(TypeFunction)d.isFuncDeclaration().type).linkage == LINK.cpp;
-                if (is_nested)
-                    buf.writeByte('X');
-                buf.writeByte('L');
+                // X .. E => template parameter is an expression
+                // 'ad'   => unary operator ('&')
+                // L .. E => is a <expr-primary>
+                buf.writestring("XadL");
                 mangle_function(d.isFuncDeclaration());
-                buf.writeByte('E');
-                if (is_nested)
-                    buf.writeByte('E');
+                buf.writestring("EE");
             }
             else if (e && e.op == TOK.variable && (cast(VarExp)e).var.isVarDeclaration())
             {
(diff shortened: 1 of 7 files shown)


Geod24:
Updated, with loads of tests
@TurkeyMan it might solve some (and hopefully all) of your issues.

I welcome other tests cases, but one of my main target was:
```cpp
#include <vector>

template<typename T>
std::vector<T>* foo() { return 0; }

int main ()
{
    foo<int>();
    return 0;
}
```

```console
$ nm a.out
0000000100000f90 T __Z3fooIiEPNSt3__16vectorIT_NS0_9allocatorIS2_EEEEv
0000000100000000 T __mh_execute_header
0000000100000f60 T _main
                 U dyld_stub_binder
```

```D
extern(C++, `std`) {
    extern (C++, `__1`) {
        struct allocator (T);
        struct vector(T, alloc = allocator!T);
    }
}

extern(C++) vector!(T)* foo(T)();

void main ()
{
    assert(foo!int() is null);
}
```

```console
$ nm abo.o
0000000000000050 s EH_frame0
00000000000000b8 s _TMP0
0000000000000040 S __D3abo12__ModuleInfoZ
0000000000000000 T __Dmain
                 U __Dmain
0000000000000068 S __Dmain.eh
                 U __Z3fooIiEPNSt3__16vectorIT_NS0_9allocatorIS2_EEEEv # Yay
                 U __d_assertp
                 U __d_run_main
                 U _main
0000000000000028 T _main
0000000000000090 S _main.eh
```

src/dmd/cppmangle.d
```diff
@@ -101,8 +101,52 @@ bool isPrimaryDtor(const Dsymbol sym)
...
+    private void reset (ref Context prev)
+    {
+        this.res = prev.res;
+    }
+}
+
 private final class CppMangleVisitor : Visitor
 {
+    /// Context used when processing pre-semantic AST
+    private Context ctx;
```
jacob-carlborg: Can we please avoid abbreviations like this.

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1426,391 @@ extern(C++):
...
+        ta.accept(this_);
+    else if (Expression ea = isExpression(o))
+        ea.accept(this_);
+    else if (Dsymbol sa = isDsymbol(o))
+        sa.accept(this_);
+    else if (TemplateParameter t = isTemplateParameter(o))
+        t.accept(this_);
+    else if (Tuple t = isTuple(o))
+        this_.visit(t);
+    else {
```
jacob-carlborg: A bit inconsistent here that none of the other branches have curly braces.

src/dmd/cppmangle.d
Geod24: I wish you could convince Walter, because `ctx` is much more widespread than the various 2-letters type abbreviations which are so common in DMD. Anyway, I'll update with the full name.

jacob-carlborg: A project wide search for `ctx` gives back 25 results and for `context` it gives back 253 results. This is just a plain text search which includes comments and everything.

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1426,391 @@ extern(C++):
...
+     * because it messes with templates.
+     *
+     * See_Also:
+     *  https://issues.dlang.org/show_bug.cgi?id=18922
+     *
+     * Params:
+     *   ns = C++ namespace to do substitution for
+     *
+     * Returns:
+     *  Index of the entry, if found, or `-1` otherwise
```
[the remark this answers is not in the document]
Geod24: That's a copy-pasta leftover, removed

Geod24:
Updated according to @jacob-carlborg 's suggestions

jacob-carlborg:
> Updated according to jacob-carlborg 's suggestions

I don't think I can be of any more help than for these stylistics issues.

Geod24:
Added some tests to instantiate some stdlib containers and return them (`test/runnable/cpp_stdlib.d`).
It fails on Linux because files need to be compiled with `-std=c++11` and on OSX32 because of `size_t` mangling, but it does pass on OSX64 and FreeBSD.

Geod24:
Should be ready. Extra test (cpp_stdlib) disabled on osx32 and windows.

TurkeyMan:
LGTM, but it's a truly gigantic patch!
The tests appear to prove it works though.

Geod24:
Okay, so this is green and IMO ready, but I don't know who can review / approve / merge this.

The last (and main) commit is indeed gigantic, but I don't see a way to break it down further that would make sense. Fortunately it's backed by an unusual amount of tests, which hopefully will provide some confidence to the poor reviewer(s).

Last but not least, most of the code in that commit should only trigger when templates are used. Given anything template-related is currently broken, even if this PR doesn't manage to capture all the cases, it will already be a big win over something which is currently completely broken.

I'd very much like this to get merged sooner than later, so we can focus on druntime's `stdcpp` bindings and other related issues (e.g. `classinfo`).

ibuclaw:
Can you wait for the weekend? I'll probably have some time then.

Geod24:
Sure can. Thanks!

Geod24:
Ping @ibuclaw

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1453,400 @@ extern(C++):
...
+
+    /**
+     * Performs template parameter substitution
+     *
+     * We perform mangling on a copy of the post-mangling AST,
+     * before any semantic pass is run.
+     * This is because there is no easy way to link a type to the template
+     * parameters once semantic ran, because:
+     * - the `TemplateInstance` installs aliases in its scope to its params
+     * - `AliasDeclaration` are resolved in many places, so we cannot keep a link
```
ibuclaw: I think you mean "`AliasDeclaration`s" as a plural here.

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1453,400 @@ extern(C++):
...
+    }
+
+    void visit(Tuple t)
+    {
+        assert(0);
+    }
+
+    /**
+     * Helper function to go through the `TemplateParameter`s
+     * and perform a substitution, if possible.
```
ibuclaw: What column length are you keeping yourself to here?  I would say 80 is standard, anything shorter just looks out of place.

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1453,400 @@ extern(C++):
...
+     *
+     * Returns:
+     *   The `TemplateParameter` that was used, or `null`.
+     */
+    private bool writeTemplateSubstitution (
+        const ref Identifier ident, TemplateParameters* params, Type type)
```
ibuclaw: Keeping parameters on the same line is more conformant to coding style.

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1453,400 @@ extern(C++):
...
+    {
+        assert(0);
+    }
+
+    /**
+     * Helper function to go through the `TemplateParameter`s
+     * and perform a substitution, if possible.
+     *
+     * Returns:
+     *   The `TemplateParameter` that was used, or `null`.
```
ibuclaw: This comment - `TemplateParameter` or `null` - does not match the return type `bool`.

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1453,400 @@ extern(C++):
...
+            {
+                if (type)
+                    CV_qualifiers(type);
+                if (this.substitute(param))
+                    return true;
+                this.append(param);
+
+                // Since we didn't run semantic yet, all we get are
+                // `TypeIdentifier`, but we need to check if it's an expression
+                // or whatnot
```
ibuclaw: This (and maybe other places, I haven't looked too deep), could be rephrased to use a little less `we`.

For instance (feel free to correct as you see fit):
```
Since semantic hasn't run yet, all that is handled here are `TypeIdentifier`,
but we still need to check if it's an expression or not.
```

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1453,400 @@ extern(C++):
...
+                if (param.isTemplateValueParameter())
+                    buf.writeByte('E');
+                return true;
+            }
+        }
+        return false;
+    }
+
+    /**
+     * Given a template instance `t`, write it's qualified name
```
ibuclaw: `its`

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1453,400 @@ extern(C++):
...
+     * without the template parameter list
+     *
+     * Params:
+     *   t = Post-parsing `TemplateInstance` pointing to the symbol
+     *       to mangle (one level deep)
+     *
+     * Returns:
+     *   `true` if the name was qualified and need an ending `E`
```
ibuclaw: and needs an ending `E`.

You could also say requires here, need sounds urgent. :-)

src/dmd/cppmangle.d
```diff
@@ -1373,4 +1453,400 @@ extern(C++):
...
+            return false;
+        }
+        // Get the template instance
+        sym = getQualifier(sym);
+        auto sym2 = getQualifier(sym);
+        if (sym2)
+        {
+            if (isStd(sym2))
+            {
+                bool useless;
```
ibuclaw: If only used `bool* = null` instead of an `out bool`...

Perhaps `unused` is a better name than `useless` for a local var.

ibuclaw (changes requested):
Few nits, nothing that I feel is preventing merge here.

src/dmd/cppmangle.d
```diff
@@ -606,6 +611,65 @@ private final class CppMangleVisitor : Visitor
             append(si);
     }
 
+    /**
+     * Write common substitution for standard types, such as std::allocator
```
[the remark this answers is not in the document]
Geod24: AFAIK, we don't use full stop for the title as it's supposed to be a one-liner.
It's however used in the description, which can be any number of sentences.

src/dmd/cppmangle.d
```diff
@@ -101,8 +101,52 @@ bool isPrimaryDtor(const Dsymbol sym)
     return dtor == ad.primaryDtor;
 }
 
+/// Context used when processing pre-semantic AST
```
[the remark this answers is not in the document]
Geod24: You mean full stop ? If so see my previous comment.

test/compilable/cppmangle.d
```diff
@@ -771,3 +773,165 @@ version(Windows)
...
 }
 
+// https://issues.dlang.org/show_bug.cgi?id=16479
+//  Missing substitution while mangling C++ template parameter for functions
+version (Posix) extern (C++)
+{
+    // Make sure aliases are still resolved
+    alias Alias16479 = int;
+    Alias16479 func16479_0 (FuncT1) (FuncT1, Alias16479);
+    static assert(func16479_0!(int).mangleof == `_Z11func16479_0IiEiT_i`);
```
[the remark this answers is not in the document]
Geod24: Agreed, I originally went with full mangleof testing (makes development easier), but then later realized I was missing some things, which is why the link testing was added.
I'll keep it in mind in the future, and maybe will move some test cases over in a later iteration.

test/runnable/cpp_stdlib.d
```diff
@@ -0,0 +1,59 @@
+// DISABLED: win32 win64 osx32
+// EXTRA_CPP_SOURCES: cpp_stdlib.cpp
+// CXXFLAGS: -std=c++11
```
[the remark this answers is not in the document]
Geod24: Yeah, I originally went full C++11 but `throwle` failed to link.
Do you have any document about the mangling changes between 98 and 11 ?
I didn't find anything relevant and figured this was too big of a matter to address in this PR.

src/dmd/cppmangle.d
```diff
@@ -101,8 +101,52 @@ bool isPrimaryDtor(const Dsymbol sym)
...
+        auto r = this.res;
+        if (r !is null)
+            this.res = next;
+        return Context(this.ti, this.fd, r);
+    }
+
+    /**
+     * Reset the context to a previous one, making any adjustment necessary
+     */
+    private void reset (ref Context prev)
```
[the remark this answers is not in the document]
Geod24: Changed to `push` / `pop`. Internal chaining sounds nice, but the current pattern is IMO simple enough and allocation-less that it does not provide much of an improvement.

src/dmd/cppmangle.d
Geod24: Yep 80. Some places have seen a few iterations and might look weird. I'm also not sure what the correct rules for breaking a sentence are in English. Fixed this.

src/dmd/cppmangle.d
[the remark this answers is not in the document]
ibuclaw: I'll highlight another example then.

src/dmd/cppmangle.d
```diff
@@ -101,8 +101,52 @@ bool isPrimaryDtor(const Dsymbol sym)
...
+    @disable ref Context opAssign(ref Context other);
+    @disable ref Context opAssign(Context other);
+
+    /**
+     * Helper function to track `res`
+     *
+     * Params:
+     *   next = Value to set `this.res` to
+     *          If `this.res` is `null`, the expression is not evalutated
+     *          This allow this code to be used even when no context is needed.
```
ibuclaw: As the first line didn't have a period, when reading this I started the parsing the second line as part of the same sentence.

Not disputing whether it looks fine in the generated docs, however I won't likely ever read from the web. :-)

src/dmd/cppmangle.d
```diff
@@ -1127,11 +1209,6 @@ extern(C++):
 
     alias visit = Visitor.visit;
 
-    override void visit(Type t)
-    {
-        error(t);
-    }
```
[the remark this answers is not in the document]
ibuclaw: It seemed to be already appropriately grouped with other type visit methods here, which is why I asked.

Geod24:
Updated:
```diff
diff --git a/src/dmd/cppmangle.d b/src/dmd/cppmangle.d
index 2083625fe..b5647e94c 100644
--- a/src/dmd/cppmangle.d
+++ b/src/dmd/cppmangle.d
@@ -118,8 +118,8 @@ private struct Context
      * Helper function to track `res`
      *
      * Params:
-     *   next = Value to set `this.res` to
-     *          If `this.res` is `null`, the expression is not evalutated
+     *   next = Value to set `this.res` to.
+     *          If `this.res` is `null`, the expression is not evalutated.
      *          This allow this code to be used even when no context is needed.
      *
      * Returns:
```

src/dmd/cppmangle.d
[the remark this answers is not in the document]
Geod24: Oups, that's right. It was moved because I ordered them from most specialized to least specialized.
Can revert if you want.

Geod24:
Anything else @ibuclaw ?

src/dmd/cppmangle.d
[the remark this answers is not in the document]
ibuclaw: The less line changes you can fit this in the better. I flagged this because initially I thought you removed it outright.

ibuclaw:
@Geod24 please read through all comments and documentation you've written up - preferably out loud. If you don't pick up any mistakes, then it's fine.

test/compilable/cppmangle.d
```diff
@@ -771,3 +773,165 @@ version(Windows)
...
+                  == `_Z11func16479_8IiEv10Array16479IT_E`);
+
+    // And non-substitution
+    void func16479_9 (FuncT1) (Array16479!(int));
+    static assert(func16479_9!(int).mangleof
+                  == `_Z11func16479_9IiEv10Array16479IiE`);
+
+    // Now let's have a bit of fun with alias parameters,
+    // starting with C functions
+    // TODO: Why is this mangled by g++:
```
thewilsonator: Is this working?

thewilsonator (approved):
Otherwise LGTM, waiting on DAutoTest to go green.

test/compilable/cppmangle.d
Geod24: Re-looking at this, it seems correct.
I'll do what Iain requested in a few hours and will update the comment.

Geod24:
> @Geod24 please read through all comments and documentation you've written up - preferably out loud. If you don't pick up any mistakes, then it's fine.

Thanks, did pick up quite a few. Removed historical artifacts, some typos, and some usage of "we".

thewilsonator:
@ibuclaw merge this?

ibuclaw (approved):

PR #8457, opened 2018-07-05
[Trivial] Fix missing newline regression in -v predefs output

kinke:
Fine with 2.080.0, but like this with 2.081.0:
```
predefs   DigitalMars Windows LittleEndian D_Version2 all D_InlineAsm D_InlineAsm_X86 X86 Win32 CRuntime_DigitalMars assert D_HardFloatbinary    C:\LDC\dmd2\windows\bin\dmd.exe
```

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1412,7 +1412,7 @@ private void printPredefinedVersions(FILE* stream)
             buf.writeByte(' ');
             buf.writestring(str.toChars());
         }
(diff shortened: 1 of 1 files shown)


kinke:
Currently building with qemu for armhf. The newer qemu version shipping with Ubuntu 18.04 is apparently finally able to use multiple host CPU cores; building LLVM 6.0.1 took less than 12 hours, that's a huge improvement. :]

JinShil (approved):

PR #8458, opened 2018-07-05

wilzbach:


wilzbach:
Tagging this with auto-merge as this is just the merge from `stable` to `master` and it prevents Vibe.d from being re-enabled at the Project Tester (which after breaking it with the 2.081 release seems like a reasonable thing to do).

PR #8460, opened 2018-07-06
[PoC] In-place struct initialization

wilzbach:
So this is my WIP hack to have in-place struct initialization into DMD.
As during the parsing stage we don't have much information about the struct and can only parse a `StructInitializer`, I created an intermediate `StructLiteralExp2` that will be converted into a `StructLiteralExp` during expression semantic.
I would like to have a more elegant solution and this still needs a lot of work, but it already passes the test suite for the common cases of this DIP and I thought it's good to show that it's possible to do this in dmd for submitting the DIP. Also I get to see what the CIs say.

What's in-place struct initialization?
-------------------------------------------

Full DIP: https://github.com/dlang/DIPs/pull/71

tl;dr:

```d
struct StructFoo
{
    int a;
}

void bar(StructFoo foo)
{
    assert(foo.a == 3);
}

StructFoo f = {a: 1};
f = StructFoo({a: 2});
assert(f.a == 2);
bar(StructFoo({a: 3}));
```

### DIFF
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -6161,6 +6161,22 @@ struct ASTBase
         }
     }
 
+    extern (C++) final class StructLiteralExp2 : Expression
+    {
+        StructInitializer _init;
+
+        extern (D) this(Loc loc, StructInitializer _init)
+        {
+            super(loc, TOK.structLiteral2, __traits(classInstanceSize, StructLiteralExp2));
+            this._init = _init;
+        }
+
+        override void accept(Visitor v)
+        {
+            v.visit(this);
+        }
+    }
+
     extern (C++) final class ArrayInitializer : Initializer
     {
         Expressions index;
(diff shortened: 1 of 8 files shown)


jacob-carlborg:
Why do you need a new expression type?

RazvanN7:
Can't you just use the existing StructLiteralExp and add the fields fields you want + a bool field that tells you if StructLiteralExp is used for in-place initialization?

andralex:
status?

jacob-carlborg:
> status?

@andralex there's a DIP that hasn't been approved. Personally I don't think a DIP is needed. But we seem to disagree on the syntax.

wilzbach:
> status?

This is a working PoC, but there are multiple ways to integrate this feature in the D grammar (see the [DIP](https://github.com/dlang/DIPs/pull/71) for more details) and as there was never a consensus on which syntax to use. @andralex what are your thoughts on the syntax for this?
Once that's decided, polishing & merging this would be easy...

dkorpel:
This has been superseded by named arguments, which have been implemented for struct constructors.

PR #8461, opened 2018-07-06
Don't fail test9287 with 'Device or resource busy'

wilzbach:
From: https://auto-tester.puremagic.com/show-run.ghtml?projectid=1&runid=3234943&isPull=true

```
Test runnable/test9287.sh failed. The logged output:
Success
rm: cannot remove ‘test_results/runnable/test9287.exe’: Device or resource busy
Test runnable/test9287.sh failed. The xtrace output:
+ source runnable/test9287.sh
++ set -e
++ echo 'import std.stdio; void main() { writeln("Success"); }'
++ ../generated/windows/release/32/dmd.exe -m32 -oftest_results/runnable/test9287.exe -
++ test_results/runnable/test9287.exe
++ rm -f test_results/runnable/test9287.obj test_results/runnable/test9287.exe
+ finish 1
+ set +x
```

I guess the easiest is the `|| true` trick to avoid this spurious failure.

### DIFF
--- a/test/runnable/test9287.sh
+++ b/test/runnable/test9287.sh
@@ -7,4 +7,4 @@ echo 'import std.stdio; void main() { writeln("Success"); }' | \
 
 ${OUTPUT_BASE}${EXE}
 
-rm -f ${OUTPUT_BASE}{${OBJ},${EXE}}
+rm -f ${OUTPUT_BASE}{${OBJ},${EXE}} || true


jacob-carlborg:
Is the executable still running?

wilzbach:
Yeah it looked like it, but don't ask me why. It just prints "Success" to stdout.

jacob-carlborg:
If the executable is still running should we try to force close it?

ZombineDev:
The question is: why is it still running?

wilzbach:
@ZombineDev if I could answer this, I wouldn't have proposed this workaround.

> If the executable is still running should we try to force close it?

@jacob-carlborg I thought that the auto-tester or other CIs do this automatically in the end, but I think even the OS itself should try to kill the process once its parent (the bash script) is terminated. Though I'm not sure what Cygwin on Windows does...

jacob-carlborg:
Most other CIs are running in Docker. For auto-tester, I don't know if it does that cleanup.

PR #8463, opened 2018-07-06
Fix issue 19066 - Allow using symbols Object inside a struct without a toHash override

schveiguy:
#8222 introduced a regression. If you have a struct with a member named `Object`, the generated code that wants to use `typeid(const(Object))` added to work around the linker error actually looks at the local `Object` symbol instead of the global.

I tried `.Object`, but if you define a variable named `Object` outside the struct, it matches there.

I tried `object.Object`, but if you define a local `object` symbol, then it fails.

`.object.Object` works, because defining a variable named `object` outside the struct is always an error (thank god!).

The test added tests for all of these cases. Ping @jacob-carlborg


schveiguy:
ping @wilzbach @CyberShadow what is that failure on the doc build? Is that something new? I've never seen it before.

CyberShadow:
Looks like another Dub issue. No idea. @s-ludwig ?

jacob-carlborg (approved):

PR #8465, opened 2018-07-06
Avoid weird `!<ClassKind> == <ClassKind.cpp>`

kinke:
... which doesn't translate to `<ClassKind> != <ClassKind.cpp>`, but to `<ClassKind> == <ClassKind.d>`! See https://run.dlang.io/is/tuYCPH.


JinShil (approved):

jacob-carlborg:
Is this testable?

dnadlinger (approved):

PR #8466, opened 2018-07-07
convert tk/mem.h to memh.d

WalterBright:
Should use .di files, but gave up on that due to non-makefile build systems can't handle it. So call it memh.d instead.

### DIFF
--- a/src/dmd/backend/memh.d
+++ b/src/dmd/backend/memh.d
@@ -1,0 +1,52 @@
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Copyright:   Copyright (C) 1985-1998 by Symantec
+ *              Copyright (C) 2000-2018 by The D Language Foundation, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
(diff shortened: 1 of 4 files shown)


WalterBright:
@CyberShadow looks like DAutoTest is having some unrelated problem.

src/dmd/backend/memh.d
```diff
@@ -0,0 +1,52 @@
...
+ *
+ * Copyright:   Copyright (C) 1985-1998 by Symantec
+ *              Copyright (C) 2000-2018 by The D Language Foundation, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/memh.d, backend/memh.d)
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/backend/memh.d
+ */
+
+
```
jacob-carlborg: Some extra unnecessary whitespace here.

src/dmd/backend/memh.d
```diff
@@ -0,0 +1,52 @@
...
+module dmd.backend.memh;
+
+
```
jacob-carlborg: Some extra unnecessary whitespace here.

jacob-carlborg (approved):

wilzbach (approved):

PR #8467, opened 2018-07-07
Move check for NRVO from ReturnStatement semantic into FuncDeclaration

ibuclaw:
Due to patches such as #7990, there's a lot of logic for determining NRVO by analysing front-end ASTs in the back-end.

GDC only checks whether `nrvo_can` is set, and then only overrides it if the target does not return the struct/static array on the stack.  So there may be a number of missed cases where NRVO is never done by GDC because the DMD front-end sets `nrvo_can = false` - only for the DMD back-end to ignore that later and return it by RVO in the codegen pass.

This moves the logic for setting `nrvo_can` into a function, then mirrors the logic from the DMD back-end to the front-end.


WalterBright:
This looks ready to me. Why is it labeled as WIP?

wilzbach:
CC @RazvanN7 who is trying to move out logic from the AST nodes.

RazvanN7:
@ibuclaw Is this still something of interest to you? If not, can we close this?

thewilsonator:
Rebased, lets see what CI thinks.

src/dmd/func.d
```diff
@@ -2496,6 +2496,55 @@ extern (C++) class FuncDeclaration : Declaration
             error("parameters must be `main()` or `main(string[] args)`");
     }
 
+    /***********************************************
+     * Check all return statements for a function to verify that returning
+     * using NRVO is possible.
+     *
+     * Returns:
+     *      true if the result cannot be returned by hidden reference.
```
MoonlightSentinel: Returning `true` if NRVO is **not** possible feels kinda wrong

thewilsonator: A lot of the `checkFoo` functions do that. I don't think thats too out of place.

MoonlightSentinel: Fair enough

PR #8472, opened 2018-07-08
ensure this implicit cast cannot happen

WalterBright:
Brought up in https://github.com/dlang/dmd/pull/8413

Ensure we don't fall into this obscure trap again.


WalterBright:
@FeepingCreature thanks for the correction

wilzbach (approved):

PR #8474, opened 2018-07-09
convert ee.c to ee.d

wilzbach:


jacob-carlborg (approved):

wilzbach:
Jenkins still has massive problems. We hopefully can switch to buildkite soon though (it's already green and doing the same work as Jenkins does in less time: https://github.com/dlang/ci/pull/225#issuecomment-403338519).

PR #8475, opened 2018-07-09
Convert outbuf.c to outbuf.d

wilzbach:


src/dmd/backend/outbuf.d
```diff
@@ -152,34 +254,89 @@ struct Outbuffer
...
+    }
 
     /**
      * Writes a 64 bit long.
      */
-    void write64(long v);
+    version (X86_64)
+        private enum write64Mangle = "_ZN9Outbuffer7write64Ex";
+    else
+        private enum write64Mangle = "_ZN9Outbuffer7write64El";
```
wilzbach: I wasn't sure what happened here. Why doesn't `long` mangle to `long long` on x86_64?
The spec says it would do so: https://dlang.org/spec/cpp_interface.html#data-type-compatibility :/

wilzbach:
Looks like I finally got Windows + Linux to pass, but macOS is still not happy about the resulting mangling:

```
  "Outbuffer::writezeros(unsigned long)", referenced from:
      MachObj::getsegment(char const*, char const*, int, int) in machobj.o
      Obj::init(Outbuffer*, char const*, char const*) in machobj.o
      Obj::term(char const*) in machobj.o
      Obj::reftoident(int, unsigned long, Symbol*, unsigned long, int) in machobj.o
  "Outbuffer::create(unsigned long)", referenced from:
      dwarf_initfile(char const*) in dwarf.o
      MachObj::getsegment(char const*, char const*, int, int) in machobj.o
      Obj::init(Outbuffer*, char const*, char const*) in machobj.o
  "Outbuffer::setsize(unsigned long)", referenced from:
      dwarf_initfile(char const*) in dwarf.o
      dwarf_termfile() in dwarf.o
      dwarf_typidx(TYPE*) in dwarf.o
      dwarf_abbrev_code(unsigned char*, unsigned long) in dwarf.o
      Obj::bytes(int, unsigned long, unsigned int, void*) in machobj.o
      MachObj::getsegment(char const*, char const*, int, int) in machobj.o
      Obj::init(Outbuffer*, char const*, char const*) in machobj.o
      ...
  "Outbuffer::position(unsigned long, unsigned long)", referenced from:
      Obj::bytes(int, unsigned long, unsigned int, void*) in machobj.o
      Obj::term(char const*) in machobj.o
      Obj::reftoident(int, unsigned long, Symbol*, unsigned long, int) in machobj.o
```

jacob-carlborg:
The mangling of `size_t` has been change back and forth on macOS, it's a complete mess. It looks like you need a newer host compiler, use a different type, use `extern(C)` or explicitly set the mangling using `pragma(mangle)`. DMD 2.081.0 mangles `size_t` as `y`, while the correct mangling is `m`. DMD master mangles it as `m`.

src/dmd/backend/cgcv.c
```diff
@@ -2263,7 +2263,7 @@ STATIC void cv4_outsym(symbol *s)
             return;
 #if SCPP
         if (CPP && isclassmember(s))            // if method
-        {   Outbuffer buf;
+        {   Outbuffer buf = Outbuffer::create();
```
WalterBright: I don't understand the need for this change.

src/dmd/backend/dwarf.c
```diff
@@ -234,7 +234,7 @@ void dwarf_appreladdr(int seg, Outbuffer *buf, int targseg, targ_size_t val)
     if (I64)
     {
         dwarf_addrel64(seg, buf->size(), targseg, val);
-        buf->write64(0);
+        write64(buf, 0);
```
WalterBright: Why this change?

WalterBright: It should work correctly if you use `stdint.h` and `int64_t` on the C side. I've encountered this problem before with divcoeff.d and cod2.c.

PR #8477, opened 2018-07-09
convert cod5.c to cod5.d

wilzbach:


jacob-carlborg (approved):

wilzbach:
BTW in case someone else wants to have a shot at this conversion project too.
Here are the simple sed replacement rules that I use so far (https://github.com/wilzbach/dmd-scripts/blob/master/convert.sh). They, of course, aren't perfect, but they automate most of the boring replacements.

PR #8478, opened 2018-07-09
Improve error message for invalid binary digit

kubo39:
schveiguy pointed out [this strange difference in error messages](https://github.com/dlang/dmd/pull/8451#discussion_r200650308).

### DIFF
--- a/src/dmd/lexer.d
+++ b/src/dmd/lexer.d
@@ -1880,7 +1880,7 @@ class Lexer : ErrorHandler
                     anyHexDigitsNoSingleUS = true;
                 if (base == 2 && !err)
                 {
-                    error("radix %d digit expected, not `%c`", base, c);
+                    error("radix 2 digit expected, not `%c`", c);
                     err = true;
                 }
                 ++p;
--- a/src/dmd/lexer.d
+++ b/src/dmd/lexer.d
@@ -1880,7 +1880,7 @@ class Lexer : ErrorHandler
                     anyHexDigitsNoSingleUS = true;
                 if (base == 2 && !err)
                 {
-                    error("binary digit expected");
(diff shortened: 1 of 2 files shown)


wilzbach (approved):

schveiguy:
Hm.. I was actually thinking we could simplify the code to be generic in terms of which bases are supported by which digits. In general, you can say a digit is invalid if the resulting integer (e.g. `d - '0'`) is `>= base`.

Otherwise, I think the original message is actually better ('binary digit' is more understandable than 'radix 2 digit'), even if inconsistent.

I'll see if I can put together a PR that shows what I mean.

src/dmd/lexer.d
```diff
@@ -1880,7 +1880,7 @@ class Lexer : ErrorHandler
                     anyHexDigitsNoSingleUS = true;
                 if (base == 2 && !err)
                 {
-                    error("binary digit expected");
+                    error("radix %d digit expected, not `%c`", base, c);
```
WalterBright: The base is always 2, it doesn't need to be an argument.

WalterBright (changes requested):

src/dmd/lexer.d
WalterBright: I don't see anything wrong with `binary` either, instead of `radix 2`.

schveiguy: The issue originally is that 0b5 prints "binary digit expected" while 0b8 prints "radix 2 digit expected not 8". It's the inconsistency that is being fixed.

But like I said, I think the whole thing can be fixed to be completely consistent and future proof. When #8451 is merged, I'll set up a new PR.

schveiguy:
This is obsoleted by #8612

PR #8479, opened 2018-07-09
Move semantic methods out of expression.d

RazvanN7:
Trying to remove semantic dependencies from expression.d. Next step is to create a visitor toBoolean methods and after that, hopefully, we can get rid of the expressionsem and dsymbolsem dependencies.


jacob-carlborg (approved):

wilzbach (approved):

WalterBright:
> Trying to remove semantic dependencies from expression.d.

Rationale?

RazvanN7:
@WalterBright Trying to pull out semantic methods out of AST nodes so that we can have a separation between parsing and semantic analysis. This will enable to replace the massive duplication in ASTBase with simple imports of AST nodes.

PR #8480, opened 2018-07-09
Reduce global state

jacob-carlborg:
This is a WIP PR on reducing the global state. It moves quite a lot of global state as in `__gshared` variables declared in classes to instance variables in a new class called `CompilerInvocation`. For backwards compatibility there's one global instance of `CompilerInvocation`. For each global state that was moved to `CompilerInvocation` property functions are generated to forward to the global instance of `CompilerInvocation`. Ideally this should not be needed at some point in the future. With this approach the global state is located in only one place and can easily be reset, just create a new instance of `CompilerInvocation`.

My plan for this is to hopefully start testing the compiler on a more unit test level, instead of the current end to end testing. For this it needs to be possible to reset the state of the compiler. This will hopefully speed up the testing since a new compiler process doesn't not need to be started for every test.

This will cause quite a lot of breakage for GDC and LDC since C++ doesn't support properties (although they can be emulated).

There are still some things that won't compile with this patch. I wanted some feedback on this to see if it's a good idea and if someone has any ideas why it's failing.

### DIFF
--- a/src/dmd/backend/dwarf.c
+++ b/src/dmd/backend/dwarf.c
@@ -1091,9 +1091,9 @@ void dwarf_initfile(const char *filename)
     }
 #endif
 #if 0 && MARS
-    for (int i = 0; i < global.params.imppath->dim; i++)
+    for (int i = 0; i < global().params.imppath->dim; i++)
     {
-        debug_line.buf->writeString((*global.params.imppath)[i]);
+        debug_line.buf->writeString((*global().params.imppath)[i]);
         debug_line.buf->writeByte(0);
     }
 #endif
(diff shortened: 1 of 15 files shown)


jacob-carlborg:
Oh, this would need a later host compiler due to https://github.com/dlang/dmd/pull/8463. Is there a way around that?

WalterBright:
> if someone has any ideas why it's failing.

The problem is you're attempting to do far, far too much in one PR. The symptom is not being able to find what went wrong. With small PRs, you've got only a small amount of code to inspect for the problem.

src/dmd/backend/dwarf.c
```diff
@@ -1091,9 +1091,9 @@ void dwarf_initfile(const char *filename)
     }
 #endif
 #if 0 && MARS
-    for (int i = 0; i < global.params.imppath->dim; i++)
+    for (int i = 0; i < global().params.imppath->dim; i++)
```
WalterBright: The problem is "side channel access to state" that is not coming through the function's parameters. Writing a function that returns global state rather than accessing global state doesn't improve this at all.

WalterBright:
To clarify, functions can access data through the front door, which is the parameter list, or through the side door, which is globals. Ideally, it should be through the front door only. (This is "weak purity" in D.) Hiding side doors behind a function call does not change this.

jacob-carlborg:
I understand. The main goal here is to be able to reset the global state, not to remove it. By moving it to one place it will make it easier to reset.

I have another patch waiting, which adds location information to all UDAs. Since this is currently not used by the compiler itself (it can be used when using the compiler as a library), I don't know how to test this the traditional way so I planned to use the compiler as a library to test it.

WalterBright:
It's easy enough to change the global state:
```
auto savestate = globalstate;
globalstate = new value;
... code ...
globalstate = savestate;
```

andralex:
This looks like a complicated way to shuffle global state around rather than eliminate it. Something like "it's not a bunch of globals, it's a Singleton!" What am I missing?

jacob-carlborg:
> This looks like a complicated way to shuffle global state around rather than eliminate it. Something like "it's not a bunch of globals, it's a Singleton!" What am I missing?

The idea was to collect the global state into one place to make it easier to reset.

PR #8481, opened 2018-07-10
convert nteh.c to D

WalterBright:
Being the first of the code gen files, this includes some modifications to the headers that will be needed by subsequent code gen conversions.

### DIFF
--- a/src/dmd/backend/nteh.d
+++ b/src/dmd/backend/nteh.d
@@ -6,43 +6,66 @@
  *              Copyright (C) 2000-2018 by The D Language Foundation, All Rights Reserved
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
- * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/nteh.c, backend/nteh.c)
- * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/backend/nteh.c
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/nteh.d, backend/nteh.d)
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/backend/nteh.d
  */
 
 // Support for NT exception handling
 
-#include        <stdio.h>
-#include        <string.h>
-#include        <time.h>
+module dmd.backend.nteh;
 
-#include        "cc.h"
-#include        "el.h"
-#include        "code.h"
-#include        "oper.h"
-#include        "global.h"
-#include        "type.h"
-#include        "dt.h"
-#if SCPP
-#include        "scope.h"
-#endif
-#include        "exh.h"
+version (SPP)
+{
+}
+else
+{
 
-#if !SPP && NTEXCEPTIONS
+import core.stdc.stdio;
+import core.stdc.string;
 
-static char __file__[] = __FILE__;      /* for tassert.h                */
-#include        "tassert.h"
+import dmd.backend.cc;
+import dmd.backend.cdef;
+import dmd.backend.code;
+import dmd.backend.code_x86;
+import dmd.backend.dt;
+import dmd.backend.el;
+import dmd.backend.global;
+import dmd.backend.oper;
+import dmd.backend.rtlsym;
+import dmd.backend.ty;
+import dmd.backend.type;
 
-static symbol *s_table;
-static symbol *s_context;
-static char s_name_context_tag[] = "__nt_context";
-static char s_name_context[] = "__context";
-static char s_name_ecode[] = "__ecode";
+version (SCPP)
+{
+    import scopeh;
+}
+else version (HTOD)
+{
+    import scopeh;
+}
 
-static char text_nt[] =
-    "struct __nt_context {"
-        "int esp; int info; int prev; int handler; int stable; int sindex; int ebp;"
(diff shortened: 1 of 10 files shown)


WalterBright:
The jenkins failure appears to be spurious.

wilzbach:
It's related to https://github.com/vibe-d/vibe.d/issues/2184 though I don't know why the old version is still download (maybe a stale cache somewhere). Anyhow, we're already in the progress of migrating to a new CI system (buildkite) and hopefully can thus finally get a stable project tester.

src/dmd/backend/rtlsym.c
```diff
@@ -23,6 +23,10 @@
 static char __file__[] = __FILE__;      /* for tassert.h                */
 #include        "tassert.h"
 
+#if HTOD
+unsigned ALLREGS;
```
wilzbach: Not `regm_t` ?

https://github.com/dlang/dmd/blob/4cc487f1edb3ed1e91ec94c98407384d3e372ae8/src/dmd/backend/cgcod.c#L65

wilzbach (approved):

src/dmd/backend/rtlsym.c
WalterBright: I think it is because regm_t isn't defined for HTOD. HTOD is the C .h file to D translator, it doesn't generated code.

src/dmd/backend/nteh.d
```diff
@@ -106,116 +131,125 @@ void nteh_filltables()
...
      *  handler address
      */
-    unsigned fsize = 4;             // target size of function pointer
-    DtBuilder dtb;
+    uint fsize = 4;             // target size of function pointer
+    scope dtb = new DtBuilder();
     int sz = 0;                     // size so far
 
-    for (block *b = startblock; b; b = b->Bnext)
+    foreach (b; BlockRange(startblock))
```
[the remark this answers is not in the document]
WalterBright: I couldn't resist.

jmdavis:
@WalterBright Where on earth is `except_gensym` defined? After these changes were committed, I can no longer build dmd on my FreeBSD system, getting

    /usr/bin/ld: Dwarf Error: found dwarf version '4', this reader only handles version 2 information.
    ../generated/freebsd/release/64/backend.a(nteh.o): In function `cdsetjmp(CodeBuilder&, elem*, unsigned int*)':
    dmd/backend/nteh.d:(.text._Z8cdsetjmpR11CodeBuilderP4elemPj[_Z8cdsetjmpR11CodeBuilderP4elemPj]+0x79): undefined reference to `except_gensym()'
    c++: error: linker command failed with exit code 1 (use -v to see invocation)
    Error: /usr/bin/c++ failed with status: 1
    gmake: *** [posix.mak:469: ../generated/freebsd/release/64/dmd] Error 1
    rm optabgen.out
    gmake: dmd failed

Grepping the repo, all I get is

    src/dmd/backend/exh.d:Symbol *except_gensym();
    src/dmd/backend/nteh.d:Symbol* except_gensym();
    src/dmd/backend/nteh.d:        s = except_gensym();
    src/dmd/backend/exh.h:symbol *except_gensym();

none of which look like definitions. Given that the autotester should be running some version of FreeBSD 11.x now (I'm running 11.1), and the autotester is passing, that implies that there's something that I need to install or do to my environment to make `except_gensym` available, but I have no clue what that would be or why it would be necessary after this commit but not before. I did try installing gcc in case this had somehow managed to make us depend on it again, but it didn't help. I can't find anything in /usr/include on my Linux box which has `except_gensym` in it, even though it doesn't seem to have this build problem. So, I have absolutely no clue where the symbol could be supposed to be coming from.

wilzbach:
It's defined in dmc:

https://github.com/DigitalMars/Compiler/blob/4b1eba84d966cdd873bc04dad4edd82ef6d3d999/dm/src/dmc/eh.d#L792

Anyhow, all the code is `neth.d` should only get compiled if `NTEXCEPTIONS` is set (which shouldn't be the case for FreeBSD?).

I assume you purged everything, did a clean rebuild and just used `make -f posix.mak`?
What host compiler are you using?

jmdavis:
I have a program that does all of the `gmake` calls for me for all of the repos and then installs the results. Let's see what it's doing... It looks like it runs

    HOST_DMD=ldmd2 gmake -f posix.mak clean
    HOST_DMD=ldmd2 gmake -f posix.mak

since it detects that ldc is installed. However, if I run

    HOST_DMD=dmd gmake -f posix.mak clean
    HOST_DMD=dmd gmake -f posix.mak

then it works. So, it looks like for some reason, this commit doesn't work with ldc, but it works with dmd. ldc on my system seems to be using version 2.079.1 of the frontend, so if I make sure that the built dmd is the same version... then dmd still works. So, I guess that this commit broke building dmd with ldc, but I don't know if that means that ldc is broken or what, since I don't know enough about how things work when you start dealing with the backend.

wilzbach:
We only test building DMD with LDC 1.10 (latest stable release). Any chance you can do:

```
. $(curl https://dlang.org/install.sh | bash -s ldc --activate)
```

and try to build DMD again?

jmdavis:
If I run that script, it says

    no ldc binaries available for freebsd

So, it's not going to work. If you're testing building dmd with the latest ldc though, that at least gives some hope that this is caused by some issue that has been fixed.

jacob-carlborg:
> no ldc binaries available for freebsd

Yeah, they stopped building binaries for FreeBSD because there were so few downloads.

jmdavis:
Normally, you'd just install it from ports, so there isn't normally much reason to download it unless it's not in ports - but it is in ports, though it's currently version 1.9 and not yet 1.10.

PR #8482, opened 2018-07-10
[RFC] An alternative directory structure

yshui:
Previously package file is looked for at `mod/package.d`. Now,
we will consider `mod.d` as a package file as well if directory
`mod/` exists.

### DIFF
--- a/src/dmd/dimport.d
+++ b/src/dmd/dimport.d
@@ -172,6 +172,7 @@ extern (C++) final class Import : Dsymbol
                     }
                     if (!mod)
                     {
+                        // FIXME confusing error messages
                         .error(loc, "can only import from a module, not from package `%s.%s`", p.toPrettyChars(), id.toChars());
                     }
                 }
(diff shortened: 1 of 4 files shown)


wilzbach:
Without wanting to demotivate you, this change will most likely require a DIP and definitely requires tests.

yshui:
@wilzbach Tests added. This change should have minimal disturbances.

Let's see if others think this needs a DIP or not.

jacob-carlborg:
I don’t think we need more ways to do the same thing. The language is complicated as it is without this.

yshui:
@jacob-carlborg This is a better way than `package.d` (it eliminate a special name `package`). So if this is accepted, I would submit a DIP to deprecate `package.d`

JinShil:
Needs a DIP.

JinShil:
In general I like the idea, but it's such a minor issue, I think it will be a tough sell.  I also agree we don't need more ways to do the same thing.  If this is the preferred way, the other should be deprecated.

Geod24:
@yshui : What's the motivation for this change ? I would really like to know if it solves an actual problem.

jacob-carlborg:
> This is a better way than package.d (it eliminates a special name package)

Isn’t `package` already a name that cannot be used for any other purpose since it’s a keyword?

yshui:
OK, then

PR #8483, opened 2018-07-10
Fix Issue 19074: Internal error: dmd/backend/el.c 1760

ibuclaw:
StructLiteralExp has a self referenced field named `origin`.  If an instance is allocated on the stack, and shallow copied to memory, this reference is garbage by the time it reaches the codegen pass.  In this case, `toElemVisitor::visit(AddrExp)` was attempting to use this data.

The intention of `origin` was that if any copies of a struct literal were made, this is the reference to the original copy.  This patch adds in that missing reference.

### DIFF
--- a/src/dmd/dinterpret.d
+++ b/src/dmd/dinterpret.d
@@ -2906,6 +2906,7 @@ public:
             auto sle = cast(StructLiteralExp)pue.exp();
             sle.type = e.type;
             sle.ownedByCtfe = OwnedBy.ctfe;
+            sle.origin = e.origin;
             result = sle;
         }
         else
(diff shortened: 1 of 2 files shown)


JinShil (approved):

WalterBright:
@ibuclaw @JinShil first off thanks for tracking down this issue, and then resolving it. This is great!

But I do have a gripe. I looked at the bug report, and started work on tracking down the problem. I did not know there was already a fix for it.

PLEASE PLEASE PLEASE when you post a PR, put a link in the bug report for it.

WalterBright (approved):

wilzbach:
Repushed to restart DAutoTest.

wilzbach (approved):

PR #8485, opened 2018-07-11
eliminate global use from compdom()

WalterBright:
Used foreach, too. Eliminating the use of globals is good practice, as well as enabling better code generation from whatever compiler is compiling this code. Now compdom() could be made pure, except for vec_calloc() and vec_free().

Also refactored blockinit() to use foreach.

### DIFF
--- a/src/dmd/backend/gloop.d
+++ b/src/dmd/backend/gloop.d
@@ -244,34 +244,34 @@ private void freeloop(loop **pl)
 
 int blockinit()
 {
-    uint i;
-    block *b;
-    int hasasm = 0;
+    bool hasasm = false;
 
     assert(dfo);
-    for (i = 0, b = startblock; b; i++, b = b.Bnext)
+    uint i = 0;
+    foreach (b; BlockRange(startblock))
     {
         debug                   /* check integrity of Bpred and Bsucc   */
-            for (list_t blp = b.Bpred; blp; blp = list_next(blp))
+          L1:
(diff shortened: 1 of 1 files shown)


jacob-carlborg (approved):

Geod24 (approved):

src/dmd/backend/gloop.d
```diff
@@ -289,54 +289,59 @@ int blockinit()
...
-    t1 = vec_calloc(vec_numbits(sb.Bdom));       // allocate a temporary
     vec_clear(sb.Bdom);
     vec_setbit(0,sb.Bdom);               // starting block only doms itself
-    for (int i = 1; i < dfotop; i++)     // for all except startblock
-        vec_set(dfo[i].Bdom);            // dominate all blocks
-    cntr = 0;                            // # of times thru loop
+    foreach (b; dfo)                     // for all except startblock
+    {
+        if (b != sb)
+            vec_set(b.Bdom);             // dominate all blocks
```
wilzbach: Wouldn't the following be a more literal translation that saves checking the condition on every loop iteration?

```
if (dfo.length > 0)
foreach (b; dfo[1 .. $])
  ...
```

wilzbach (approved):

src/dmd/backend/gloop.d
WalterBright: I was thinking about making the algorithm independent of `dfo` being an array.

PR #8486, opened 2018-07-11
[REG] Fix Issue 19076 - dmd 2.081 crashed by getVirtualFunctions for a interface extended interface

RazvanN7:
When traits(getVirtualMethods) is called, first, semantic is performed on the expression. If the provided aggregate does not contain the provided method an error is issued, but the function does not return [1]. 
I would argue that the function should return an empty tuple, but I do not know the reason why it doesn't, so according to Chesterton's fence, I will leave it be. 

https://github.com/dlang/dmd/blob/master/src/dmd/traits.d#L927

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -1010,7 +1010,7 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
             // If the symbol passed as a parameter is an
             // interface that inherits other interfaces
             overloadApply(f, &dg);
-            if (ifd && ifd.interfaces)
+            if (ifd && ifd.interfaces && f)
             {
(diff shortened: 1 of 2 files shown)


jacob-carlborg (approved):

wilzbach (approved):

wilzbach:
FYI this looks good on buildkite:

image

I know that the log isn't public yet, but I sent invites to all currently active people here (let me know if you haven't gotten one) and I will anyhow monitor buildkite actively for the next two weeks and post the failing log manually if it's really a valid failure.
This seems to me a better approach than having a constantly red Jenkins.

Geod24:
@wilzbach Letting you know

PR #8487, opened 2018-07-11
Generate expressive diagnostic messages with the offending line

wilzbach:
A very naive implementation to ask on feedback on this before I invest more time in this.

Questions
---------

1) Should we save the originally opened file (it's currently thrown away by default)? Re-opening them has the advantage of only paying this overhead in an error case.

2) Under which CLI flag should this be exposed?

3) Is there a chance that we can do sth. similar like we do with the coloring? Only if dmd is run in a terminal, turn of the expressive diagnostics by default if dmd, but still have flags to control the behavior manually

Clang calls this "expressive diagnostics" and [there's a lot more cool stuff that we can do](https://clang.llvm.org/diagnostics.html).

Motivation
----------

[Compiler error for humans](http://elm-lang.org/blog/compiler-errors-for-humans) sums it up pretty nicely.

Also note that all three major C++ compilers (Clang, GCC, MSVC) have such expressive diagnostics [enabled by default](https://easyaspi314.github.io/gcc-vs-clang.html). It's time for DMD to catch up with the modern era.

### DIFF
--- a/src/dmd/errors.d
+++ b/src/dmd/errors.d
@@ -233,6 +233,31 @@ private void verrorPrint(const ref Loc loc, Color headerColor, const(char)* head
     else
         fputs(tmp.peekString(), stderr);
     fputc('\n', stderr);
+
+    if (global.params.printErrorContext &&
+        // ignore invalid files
+        loc != Loc.initial &&
+        // ignore mixins for now
+        !loc.filename.strstr(".d-mixin-"))
+    {
+        import dmd.filecache : FileCache;
+        auto fllines = FileCache.fileCache.addOrGetFile(loc.filename[0 .. strlen(loc.filename)]);
+
+        if (loc.linnum - 1 < fllines.lines.length)
+        {
+            auto line = fllines.lines[loc.linnum - 1];
+            if (loc.charnum < line.length)
+            {
+                fprintf(stderr, "%.*s\n", line.length, line.ptr);
+                foreach (_; 1 .. loc.charnum)
+                    fputc(' ', stderr);
+
+                fputc('^', stderr);
+                fputc('\n', stderr);
+            }
+        }
+    }
+end:
     fflush(stderr);     // ensure it gets written out in case of compiler aborts
 }
 
--- a/src/dmd/errors.d
+++ b/src/dmd/errors.d
@@ -238,7 +238,8 @@ private void verrorPrint(const ref Loc loc, Color headerColor, const(char)* head
         // ignore invalid files
         loc != Loc.initial &&
         // ignore mixins for now
-        !loc.filename.strstr(".d-mixin-"))
+        !loc.filename.strstr(".d-mixin-") &&
+        !global.params.mixinOut)
     {
         import dmd.filecache : FileCache;
         auto fllines = FileCache.fileCache.addOrGetFile(loc.filename[0 .. strlen(loc.filename)]);
--- a/src/dmd/filecache.d
+++ b/src/dmd/filecache.d
@@ -1,0 +1,127 @@
+/**
+ * Compiler implementation of the
+ * $(LINK2 http://www.dlang.org, D programming language).
+ *
+ * Copyright:   Copyright (C) 1999-2018 by The D Language Foundation, All Rights Reserved
+ * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
+ * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/filecache.d, filecache.d)
+ * Documentation:  https://dlang.org/phobos/dmd_filecache.html
+ * Coverage:    https://codecov.io/gh/dlang/dmd/src/master/src/dmd/filecache.d
+ */
+
+module dmd.filecache;
+
+import dmd.root.stringtable;
+import dmd.root.array;
+import dmd.root.file;
+
+import core.stdc.stdio;
+
+/**
+A line-by-line representation of a $(REF File, dmd,root,file).
+*/
+class FileAndLines
+{
+    File* file;
+    const(char[])[] lines;
+
+    /**
+    File to read and split into its lines.
+    */
+    this(const(char)[] filename)
+    {
+        file = new File(filename);
+        readAndSplit();
+    }
+
+    // Read a file and split the file buffer linewise
+    private void readAndSplit()
+    {
+        file.read();
+        auto buf = file.buffer;
+        // slice into lines
+        while (*buf)
+        {
+            auto prevBuf = buf;
+            for (; *buf != '\n' && *buf != '\r'; buf++)
+            {
+                if (!*buf)
+                    break;
+            }
+            // handle Windows line endings
+            if (*buf == '\r' && *(buf + 1) == '\n')
+                buf++;
+            lines ~= cast(const(char)[]) prevBuf[0 .. buf - prevBuf];
+            buf++;
+        }
+    }
+
+    void destroy()
+    {
+        if (file)
+        {
+            file.destroy();
+            file = null;
+            lines.destroy();
+            lines = null;
+        }
+    }
+
+    ~this()
+    {
+        destroy();
+    }
+}
+
+/**
+A simple file cache that can be used to avoid reading the same file multiple times.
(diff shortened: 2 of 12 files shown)


jacob-carlborg:
I like it. Would be nice to have as default. 

I’ve been thinking about fixits that Clang has as well. This would be especially useful for deprecations that usually contains a suggestion how to fix it.

test/fail_compilation/fail_pretty_errors.d
```diff
@@ -0,0 +1,15 @@
...
+    a = 1;
+    ^
+---
+*/
+
+void foo()
+{
+
+    a = 1;
+}
```
Geod24: I'd add a test with `mixin()` and `mixin template`

ZombineDev:
See also https://blog.rust-lang.org/2016/08/10/Shape-of-errors-to-come.html for more inspiration.

wilzbach:
> I'd add a test with mixin() and mixin template

Added. For simplicity, I don't do the pretty printing for mixins for now though.

> Would be nice to have as default.

Yeah that's my goal too!
However, before we make it the default, I think we should have it as an opt-in for two or three releases, s.t. we get enough feedback/testing.

WalterBright:
I want to thank @wilzbach for finding a way to do this that doesn't involve changing 30 or 40 files :-) I like that the implementation is pretty encapsulated. When other PRs cross-cut through the files, it shows how poorly encapsulated the compiler internals are.

1. Reopen the file for printing errors. (Otherwise, keeping the files around consumes a lot of memory.) But once it is reopened for error messages, cache it in a buffer.)

2. I'd rather the switch be "context" rather than "pretty". Pretty-printing usually refers to a function identifier being expanded to printing the whole declaration.

3. I don't think it being in outputting to a tty should change this behavior.

[On a personal note, I am annoyed by this, for the simple reason that the DMC C compiler has done the line with the caret since 1982, and not a soul ever cared for the feature or even mentioned it. So I dropped it for DMD. Suddenly, it's the height of fashion, innovation and modernity to have it, and I'm 180 degrees out of phase. Sigh.]

WalterBright:
In case I wasn't clear, yes, let's move forward with it.

wilzbach (commented):
Added a simple line-by-line cache, but before I figure out how to add it to all the makefile, it would be good to know if this should live in a separate module or somewhere else.

andralex:
cc @WalterBright

src/dmd/filecache.d
```diff
@@ -0,0 +1,137 @@
...
+module dmd.filecache;
+
+import dmd.root.stringtable;
+import dmd.root.array;
+import dmd.root.file;
+
+import core.stdc.stdio;
+
+// A simple wrapper around a string in D for extern(C++) compatibility
+extern(C++) struct Dstring
```
jacob-carlborg: I don't think this need to be exposed to C++. I'm pretty sure that both LDC and GDC are using their own error printing.

src/dmd/filecache.d
```diff
@@ -0,0 +1,137 @@
...
+*/
+class FileAndLines
+{
+    File* file;
+    Array!Dstring* lines;
+
+    /**
+    File to read and split into its lines.
+    */
+    this(const(char)* filename)
```
jacob-carlborg: Please use D strings.

src/dmd/filecache.d
```diff
@@ -0,0 +1,137 @@
...
+    // Read a file and split the file buffer linewise
+    private void readAndSplit()
+    {
+        file.read();
+        auto buf = file.buffer;
+        // slice into lines
+        while (*buf)
+        {
+            auto prevBuf = buf;
+            for (; *buf != '\n'; buf++)
```
jacob-carlborg: Do we need to care about Windows line endings?

src/dmd/filecache.d
```diff
@@ -0,0 +1,137 @@
...
+    If the file has been read before, the cached file object will be returned
+
+    Params:
+        file = file to load in (or get from) the cache
+
+    Returns: a $(LREF FileAndLines) object containing a line-by-line representation of the requested file
+    */
+    FileAndLines addOrGetFile(const(char)[] file)
+    {
+        if (auto payload = files.lookup(file.ptr, file.length))
```
jacob-carlborg: There's an overload taking a D string.

src/dmd/filecache.d
```diff
@@ -0,0 +1,137 @@
...
+            if (payload !is null)
+                return cast(typeof(return)) payload.ptrvalue;
+
+        auto lines = new FileAndLines(file.ptr);
+        auto p = files.insert(file.ptr, file.length, cast(void*) lines);
```
jacob-carlborg: There's an overload taking a D string.

src/dmd/filecache.d
```diff
@@ -0,0 +1,137 @@
...
+
+    void deinitialize()
+    {
+        foreach (sv; files)
+            sv.destroy();
+        files.reset();
+    }
+}
+
+static fileCache = FileCache();
```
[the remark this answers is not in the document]
jacob-carlborg: `__gshared`? All (most?) other module level variables in DMD are `__gshared`.

src/dmd/filecache.d
```diff
@@ -0,0 +1,137 @@
...
+
+shared static this()
```
jacob-carlborg: I think we should avoid module constructors.

src/dmd/globals.d
```diff
@@ -158,6 +158,7 @@ struct Param
      */
 
     bool showGaggedErrors;  // print gagged errors anyway
+    bool contextPrintErrors;  // print errors with the error context (the error line in the source file)
```
jacob-carlborg: Perhaps `printErrorContext` is a better name?

jacob-carlborg (changes requested):

src/dmd/filecache.d
wilzbach: Well, it's a general definition of a string in D and might be useful at other places.

thewilsonator: FWIW LDC already has one of those.

ibuclaw:
Moving forward post this pr, it would be really nice to have location ranges (start and end locations for all expressions).

src/dmd/filecache.d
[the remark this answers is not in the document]
wilzbach: Changed it to `__gshared`

src/dmd/filecache.d
wilzbach: Removed it.

src/dmd/filecache.d
wilzbach: Done.

src/dmd/filecache.d
wilzbach: Used D strings.

src/dmd/filecache.d
wilzbach: Changed to the D string overload.

src/dmd/globals.d
wilzbach: Renamed.

src/dmd/filecache.d
wilzbach: Removed the module constructor, but now we need to manually initialize in `mars.d` / `frontend.d`

src/dmd/filecache.d
wilzbach: Yeah. Added a check for it.

jacob-carlborg (approved):

src/dmd/errors.d
```diff
@@ -233,6 +233,31 @@ private void verrorPrint(const ref Loc loc, Color headerColor, const(char)* head
     else
         fputs(tmp.peekString(), stderr);
     fputc('\n', stderr);
+
+    if (global.params.printErrorContext &&
+        // ignore invalid files
+        loc != Loc.initial &&
+        // ignore mixins for now
+        !loc.filename.strstr(".d-mixin-"))
```
thewilsonator: Since the addition of `-mixin=<filename>` might want to doublecheck that logic.

wilzbach: Good point. For now, I just added `&& !global.params.mixinOut` as that's definitely safe.
In the future, we probably need to call `flushMixins` or make the fileCache aware of mixin files.

wilzbach:
Ping @thewilsonator. Anything blocking this PR?
It's not enabled by default yet and it would be great to gather some real-world feedback with the next release / nightlies.

PR #8488, opened 2018-07-12
Fix Issue 18979 - Template constructor bypasses private

RazvanN7:


src/dmd/access.d
```diff
@@ -144,13 +144,25 @@ extern (C++) bool checkAccess(AggregateDeclaration ad, Loc loc, Scope* sc, Dsymb
         printf("AggregateDeclaration::checkAccess() for %s.%s in function %s() in scope %s\n", ad.toChars(), smember.toChars(), f ? f.toChars() : null, cdscope ? cdscope.toChars() : null);
     }
     Dsymbol smemberparent = smember.toParent();
```
RazvanN7: If it weren't for the deprecation, the single modification for this patch would have been transforming this line into:

`Dsymbol smemberparent = smember.toParent2()`

Explanation:
For template declarations, the parent is the enclosing template declaration; we are not interested in that in this situation, but if the template declaration itself is nested by a struct or not.

src/dmd/access.d
```diff
@@ -144,13 +144,25 @@ extern (C++) bool checkAccess(AggregateDeclaration ad, Loc loc, Scope* sc, Dsymb
...
+    bool deprecate18979;
+    bool checkParent(Dsymbol sym) { return !sym || !sym.isAggregateDeclaration(); }
+    if (checkParent(smemberparent))
     {
         static if (LOG)
         {
             printf("not an aggregate member\n");
         }
-        return false; // then it is accessible
+        // @@@DEPRECATED_2019-02@@@
```
wilzbach: Use versions as the deprecation DIP suggests.

src/dmd/access.d
```diff
@@ -144,13 +144,25 @@ extern (C++) bool checkAccess(AggregateDeclaration ad, Loc loc, Scope* sc, Dsymb
         printf("AggregateDeclaration::checkAccess() for %s.%s in function %s() in scope %s\n", ad.toChars(), smember.toChars(), f ? f.toChars() : null, cdscope ? cdscope.toChars() : null);
     }
     Dsymbol smemberparent = smember.toParent();
-    if (!smemberparent || !smemberparent.isAggregateDeclaration())
+    bool deprecate18979;
+    bool checkParent(Dsymbol sym) { return !sym || !sym.isAggregateDeclaration(); }
```
wilzbach: static

wilzbach:
> Doesn anyone have any ideas why the tester is failing to test druntime? I cannot reproduce locally

Run:

```
cd druntime
make -f posix.mak benchmark-compile-only
```

RazvanN7:
@schveiguy given the code:

```d
// test.d

import imports.imp18979;
 void main()
{
    auto f = Foo(42);
}

// imp198979
struct Foo
{
    private this(A)(A a) {}
}
```
The call to the constructor is lowered the following the way:
 1. with DotTemplateExp:auto f = Foo().__ctor(A)(A a)(42)
 2. with DotIdExp: auto f = Foo().__ctor(42)

The first one bypasses somehow the private check, but after some more processing it gets lowered to the second one. I'm not entirely sure about this fix because I don't completely understand DotTemplateExp is supposed to be used for, but it seems that all tests are green with this change.

schveiguy:
Hm... OK. I can buy that explanation. It's odd that one expansion would bypass protection attributes in a way that is invalid for the language. That's probably something that should be fixed in general.

Still too timid to say I approve for merging, but it makes sense to me.

RazvanN7:
Why does this need rebase? @thewilsonator

PR #8489, opened 2018-07-12
Const add

gapdan:


jacob-carlborg (approved):

src/dmd/link.d
```diff
@@ -745,7 +745,7 @@ version (Windows)
         {
             if ((len = strlen(args)) > 255)
             {
-                char* q = cast(char*)alloca(8 + len + 1);
+                const char* q = cast(char*)alloca(8 + len + 1);
                 sprintf(q, "_CMDLINE=%s", args);
```
wilzbach:
```
dmd\link.d(749): Error: function `core.stdc.stdio.sprintf(scope char* s, scope const(char*) format, ...)` is not callable using argument types `(const(char*), string, const(char)*)`
dmd\link.d(749):        cannot pass argument `q` of type `const(char*)` to parameter `scope char* s`
```

We probably should change the declaration in druntime to reflect this, but I guess for now it looks you can't easily do this here.

src/dmd/lexer.d
```diff
@@ -147,7 +147,7 @@ unittest
      */
     string text = "int"; // We rely on the implicit null-terminator
     scope Lexer lex1 = new Lexer(null, text.ptr, 0, text.length, 0, 0);
-    TOK tok;
+    const TOK tok;
```
Geod24: This won't work, it gets modified afterwards

src/dmd/link.d
```diff
@@ -745,7 +745,7 @@ version (Windows)
         {
             if ((len = strlen(args)) > 255)
             {
-                char* q = cast(char*)alloca(8 + len + 1);
+                const char* q = cast(char*)alloca(8 + len + 1);
```
Geod24: This cannot work because we need to mutate it later

Geod24 (changes requested):
I commented on changes which cannot work.
I think adding `const` to parameters which are value type is not useful. It doesn't add anything to the interface.
Adding `const` inside body is good though.

jacob-carlborg:
> I think adding const to parameters which are value type is not useful. It doesn't add anything to the interface.

No, but it won't be possible to modify the parameter inside the function.

> Adding const inside body is good though.

Why shouldn't it apply to parameters as well? They are basically like local variables.

RazvanN7 (approved):

jacob-carlborg:
Please squash the commits.

Geod24:
@jacob-carlborg : Except that local variables are not part of the function's interface.
It's not a big deal to have those, certainly not something to block a merge over, but it is just at odds with the rest of the codebase, not only dmd, but druntime and Phobos as well (there was a similar discussion on `scope` on value type recently in Phobos).

jacob-carlborg:
> Except that local variables are not part of the function's interface.

Yes. But why should you be allowed to modify the parameters inside a function but not the local variables?

Geod24:
@gapdan : I don't necessarily agree, but not enough to block this PR :)
I'll close/reopen to re-trigger the CI.

wilzbach:
>  I don't necessarily agree, but not enough to block this PR

Same here.

> I'll close/reopen to re-trigger the CI.

Jenkins has now finally been replaced by Buildkite, Force-merging this one ;-)

PR #8490, opened 2018-07-12
Fix Issue 19070 - Deprecate octal literal `00` through `07`

kubo39:


Geod24:
@kubo39 : D has a long history of not imposing unnecessary restrictions. This is a goof example of it IMO.
I don't think we are going to get rid of the error message, since it would break compatibility with C code, but disallowing `01` through `07` doesn't seem to add any value for the end user.

kubo39:
I think so too, okay, close this.

PR #8499, opened 2018-07-15
add type parameter to ArrayLiteralExp constructors

WalterBright:
Most of the time the type parameter is needed, so add it.


wilzbach (approved):
Shouldn't the constructor in astbase be updated too?

wilzbach:
(Jenkins failure is unrelated)

WalterBright:
> Shouldn't the constructor in astbase be updated too?

May break using the AST library.

wilzbach:
FWIW DMD as a library is still declares as unstable and AFAICT no one is using it as of now.

PR #8501, opened 2018-07-16
refactor semanticY()

WalterBright:
Takes advantage of UFCS to reduce lines of code, remove unnecessary variables, reduce scope of `e`. Changes are all local, so should be an easy review. No semantic changes.

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -10783,6 +10783,11 @@ Expression semanticY(DotTemplateInstanceExp exp, Scope* sc, int flag)
         printf("DotTemplateInstanceExpY::semantic('%s')\n", exp.toChars());
     }
 
+    static Expression errorExp()
+    {
+        return new ErrorExp();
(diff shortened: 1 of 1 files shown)


src/dmd/expressionsem.d
```diff
@@ -10836,77 +10841,68 @@ L1:
...
-            Declaration v = s.isDeclaration();
-            if (v)
+            if (Declaration v = exp.ti.toAlias().isDeclaration())
             {
                 if (v.type && !v.type.deco)
                     v.type = v.type.typeSemantic(v.loc, sc);
-                e = new DotVarExp(exp.loc, exp.e1, v);
-                e = e.expressionSemantic(sc);
-                return e;
+                return (new DotVarExp(exp.loc, exp.e1, v))
```
jacob-carlborg: The parentheses around the new expression is not necessary.

RazvanN7 (approved):

PR #8502, opened 2018-07-18
sideeffect.d: small refactor

WalterBright:
Eliminate some unnecessary computation and unnecessary variables.

### DIFF
--- a/src/dmd/sideeffect.d
+++ b/src/dmd/sideeffect.d
@@ -123,6 +123,9 @@ extern (C++) int callSideEffectLevel(Type t)
         assert(t.ty == Tfunction);
         tf = cast(TypeFunction)t;
     }
+    if (!tf.isnothrow)  // function can throw
+        return 0;
+
     tf.purityLevel();
     PURE purity = tf.purity;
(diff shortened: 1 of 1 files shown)


jacob-carlborg (approved):

src/dmd/sideeffect.d
```diff
@@ -364,13 +365,11 @@ extern (C++) bool discardValue(Expression e)
...
 VarDeclaration copyToTemp(StorageClass stc, const char* name, Expression e)
 {
-    assert(name && name[0] == '_' && name[1] == '_');
-    auto id = Identifier.generateId(name);
-    auto ez = new ExpInitializer(e.loc, e);
-    auto vd = new VarDeclaration(e.loc, e.type, id, ez);
-    vd.storage_class = stc;
-    vd.storage_class |= STC.temp;
-    vd.storage_class |= STC.ctfe; // temporary is always CTFEable
+    assert(name[0] == '_' && name[1] == '_');
```
ibuclaw: Let's hope the caller never passes null.

WalterBright: If it is, there'll be a seg fault.

PR #8503, opened 2018-07-18
Fix Issue 19095 - deprecate array length mismatch in global static arrays

kubo39:


src/dmd/initsem.d
```diff
@@ -408,6 +408,10 @@ private extern(C++) final class InitializerSemanticVisitor : Visitor
         if (t.ty == Tsarray)
         {
             uinteger_t edim = (cast(TypeSArray)t).dim.toInteger();
+            if (!i.isAssociativeArray() && i.dim != edim)
+            {
+                deprecation(i.loc, "mismatched array lengths, %llu and %u", edim, i.dim);
```
jacob-carlborg: The message doesn’t say which length is which.

jacob-carlborg (changes requested):
Might need a changelog entry and spec update, I can never remember when these should be added.

src/dmd/initsem.d
wilzbach: Needs a @@@DEPRECATED comment as described in https://github.com/dlang/DIPs/blob/master/DIPs/DIP1013.md

wilzbach: Also shouldn't `i.deprecation` work directly?

test/fail_compilation/fail19095.d
```diff
@@ -0,0 +1,16 @@
...
+fail_compilation/fail19095.d(15): Deprecation: mismatched array lengths, 2 and 1
+---
+ */
+
+// https://issues.dlang.org/show_bug.cgi?id=1XXXX
+
+int[2] ARR = [1];
+static int[2] static_arr = [1];
+__gshared int[2] gshared_arr = [1];
+int[3] asoc_arr = [1:2];
```
wilzbach: Wow, I didn't know that this was legal, but it really is: https://run.dlang.io/is/3ytqf6
I think this should be deprecated too!

wilzbach:
> Might need a changelog entry and spec update, I can never remember when these should be added.

In theory https://github.com/dlang/DIPs/blob/master/DIPs/DIP1013.md should clarify this.
As it doesn't fully do this, here's the gist:
- __every__ new deprecation needs to have a changelog entry explaining the motivation (and how to fix the code)
- __every__ new deprecation needs to have a `@@@DEPRECATED`  pattern, s.t. it can be easily grepped for

Deprecations of language features should be added to the [deprecation page at dlang.org](https://dlang.org/deprecate.html), but this isn't a language feature - just the correction of a small oversight.

jacob-carlborg:
Both the deprecation message and changelog entry should ideally contain an indication how to get back to old behavior.

test/fail_compilation/fail19095.d
ZombineDev: On the contrary, I think this is a handy way to initialize an array with mostly zeroes:
https://dlang.org/spec/arrays#static-init-static

wilzbach:
@kubo39 there are still failures in fail_compilation.

How to run:

````
cd test && ./run.d fail_compilation
```

How to update the test output:

````
./run.d fail_compilation AUTO_UPDATE=1
```

kubo39:
Thank you for some reviews.
I think this provides the least surprising behaviour for the behavior of static arrays of local scope, and it is not necessary to change the behavior of the preceded by an index and a : since it can be determined that this is a special case.

WalterBright:
@kubo39 Please use a PR title that starts with "Fix Issue XXXXX" so the tester will automatically connect to the issue. https://issues.dlang.org/show_bug.cgi?id=19095

wilzbach:
> @kubo39 Please use a PR title that starts with "Fix Issue XXXXX" so the tester will automatically connect to the issue. https://issues.dlang.org/show_bug.cgi?id=19095

The bot looks at commit messages, not PR titles.
Also @kubo39 has closed this PR - though it looks like that might have been accidental?

kubo39:
Sorry, this is an accident.

RazvanN7:
Closing in favor of: https://github.com/dlang/dmd/pull/10585

PR #8506, opened 2018-07-19
Fix Issue 19098 - Improve error for non-assignable struct

RazvanN7:
That line was not covered by any tests until now.

### DIFF
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -948,7 +948,8 @@ extern (C++) abstract class Expression : RootObject
             }
             else if (!type.isAssignable())
             {
-                error("cannot modify struct `%s` `%s` with `immutable` members", toChars(), type.toChars());
+                error("cannot modify struct instance `%s` of type `%s` because it contains `const` or `immutable` members",
+                    toChars(), type.toChars());
                 return new ErrorExp();
             }
         }
--- a/test/fail_compilation/fail19098.d
+++ b/test/fail_compilation/fail19098.d
@@ -1,0 +1,19 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail19098.d(18): Error: cannot modify struct instance `a` of type `A` because it contains `const` or `immutable` members
+---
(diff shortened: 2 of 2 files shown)


src/dmd/expression.d
```diff
@@ -948,7 +948,7 @@ extern (C++) abstract class Expression : RootObject
             }
             else if (!type.isAssignable())
             {
-                error("cannot modify struct `%s` `%s` with `immutable` members", toChars(), type.toChars());
+                error("cannot modify struct instance `%s` of type `%s` because it contains non-mutable members", toChars(), type.toChars());
```
jacob-carlborg: How about "value" instead of `struct instance`, or is it important to know it's a struct in the error message?

RazvanN7: If the struct contains any mutable fields, you can modify the value of the instance by assigning directly to one of the them. Instead, this error message is for instance assignments  (`a = b` for example ) where you modify the value as a whole. In my opinion, this message is better.

stefan-koch-sociomantic: `where you modify the value as a whole` ...

WalterBright: Please use `immutable` not `non-mutable`. The former is the usual term for it, the latter is one I've never seen before.

klickverbot: The distinction is between `immutable` and `const/immutable`. Using the former to refer to a struct without `immutable` members would be confusing.

WalterBright: Good point, but then I'd prefer "contains const or immutable members".

jacob-carlborg:
Is it too much to include the `const`/`immutable` members in the message?

RazvanN7:
>Is it too much to include the const/immutable members in the message?

I can't because : 

1. There might be a lot of const/immutable members and including them might be redundant
2. Type.isAssignable returns bool, so it would have to be refactored to return an array of VarDeclarations

jacob-carlborg (approved):

PR #8507, opened 2018-07-19
Fix Issue 19099 - Struct with field that has postblit or destructor makes struct assignable

RazvanN7:
I know this requires a deprecation, it's just that it's a bit complicated to do it.

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -5509,13 +5509,15 @@ extern (C++) final class TypeStruct : Type
         bool assignable = true;
         uint offset = ~0; // dead-store initialize to prevent spurious warning
 
+        sym.determineSize(sym.loc);
+
         /* If any of the fields are const or immutable,
          * then one cannot assign this struct.
          */
         for (size_t i = 0; i < sym.fields.dim; i++)
         {
             VarDeclaration v = sym.fields[i];
-            //printf("%s [%d] v = (%s) %s, v.offset = %d, v.parent = %s", sym.toChars(), i, v.kind(), v.toChars(), v.offset, v.parent.kind());
+            //printf("%s [%d] v = (%s) %s, v.offset = %d, v.parent = %s\n", sym.toChars(), i, v.kind(), v.toChars(), v.offset, v.parent.kind());
             if (i == 0)
             {
             }
--- a/test/fail_compilation/fail19099.d
+++ b/test/fail_compilation/fail19099.d
@@ -1,0 +1,27 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail19099.d(26): Error: cannot modify struct instance `a` of type `A` because it contains `const` or `immutable` members
+---
+*/
+
+struct B
+{
+    this(this) {}
+    ~this() {}
+    int a;
+}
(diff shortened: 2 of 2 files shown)


test/fail_compilation/fail19099.d
```diff
@@ -0,0 +1,27 @@
...
+TEST_OUTPUT:
+---
+fail_compilation/fail19099.d(26): Error: cannot modify struct `a` `A` with `immutable` members
+---
+*/
+
+struct B
+{
+    //this(this) {}
+    //~this() {}
```
jacob-carlborg: Shouldn't these lines be enabled?

test/fail_compilation/fail19099.d
```diff
@@ -0,0 +1,27 @@
...
+{
+    this(this) {}
+    ~this() {}
+    int a;
+}
+
+struct A
+{
+    B b;
+    immutable int a;
```
jacob-carlborg: Does this error occur because `a` is immutable even though `b` isn’t?

RazvanN7: yes

WalterBright:
@RazvanN7 Please add a link to the PR to the bugzilla entry. Otherwise, someone will waste time fixing a bug that has already been fixed. This has happened.

I already added it for this one.

src/dmd/mtype.d
```diff
@@ -5498,7 +5498,7 @@ extern (C++) final class TypeStruct : Type
             if (i == 0)
             {
             }
-            else if (v.offset == offset)
+            else if (!(sym.postblit || sym.dtor) && v.offset == offset)
```
WalterBright: This is rather mystifying. First off, this clause applies to unions, and the added test has nothing to do with unions. Second, the test, being on `sym`, is loop invariant and so looks like it should not be in this loop at all. Third, it is not checked if it is the first field, and the test case conveniently has the immutable member as the second field - I suspect this fix won't work at all if fields `a` and `b` are swapped. And lastly, this test looks like it should be someplace other than this function.

WalterBright: It seems that a postblit should not even be generated for this struct, and that's where the fix should go.

WalterBright (dismissed):
see my comments.

RazvanN7:
After my first noobish attempt, I gave this another look and found the real problem. When the postblit is generated for a struct, `opAssign` is also generated; `opAssign` is generated only if all the members are assignable; the TypeStruct `isAssignable` function tests if each field is assignable, however it has a hack for anonymous unions members so that if a member is assignable, the union is considered assignable; to implement this, it is checked if fields have the same offset, however, the size for the struct is determined on in the second semantic step (right after semantic 1 finished), therefore at this point, all fields have offset 0 => this bug. To fix this, I called the `determineSize` function in `isAssignable` so that the offsets are correct. This should be the correct course of action, as at the time of `opAssign` generation all fields are already semantically analyzed.

ping @WalterBright @andralex

RazvanN7:
cc @thewilsonator

jacob-carlborg (approved):

thewilsonator (approved):

PR #8508, opened 2018-07-19
[WIP] Activate -dip1008 (@nogc exception handling) by default

wilzbach:
Just to see what would break if `-dip1008` is activated by default.

### DIFF
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -291,10 +291,6 @@ dmd -cov -unittest myprog.d
             "implement https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md",
             "implement $(LINK2 https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md, DIP1000 (Scoped Pointers))"
         ),
-        Option("dip1008",
-            "implement https://github.com/dlang/DIPs/blob/master/DIPs/DIP1008.md",
-            "implement $(LINK2 https://github.com/dlang/DIPs/blob/master/DIPs/DIP1008.md, DIP1008 (@nogc Throwable))"
-        ),
         Option("fPIC",
             "generate position independent code",
             TargetOS.all & ~(TargetOS.windows | TargetOS.macOS)
--- a/test/compilable/testdip1008.d
+++ b/test/compilable/testdip1008.d
@@ -1,5 +1,4 @@
 // PERMUTE_ARGS:
-// REQUIRED_ARGS: -dip1008
 
 int bar()
 {
--- a/changelog/dip1008-default.dd
+++ b/changelog/dip1008-default.dd
@@ -1,0 +1,14 @@
+Exceptions are `@nogc`
+
+DIP 1008 is now activated by default:
+
+---
+void main() @nogc
+{
+    throw new Exception("I'm @nogc now");
+}
+---
+
+$(CONSOLE
+rdmd app.d
+)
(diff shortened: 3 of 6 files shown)


wilzbach:
So this pattern will cause breakage:

```d
Exception collectException(void delegate() dg)
{
    try
        dg();
    catch (Exception e)
        return e.;
    return null;
}
```

According to DIP1008:

> Catch objects being scope will cause problems in that everything done with those objects will also have to be scope. The most likely problem will be printing the objects, which relies on Object.toString(), which is not scope. One possible solution is to force Throwable.toString() to be scope, which will likely cause minimal disruption. Of course, compiling with -dip1000 will disable such checking and can work in the interim. Code that needs to leak the thrown exception object can clone the object.

https://github.com/dlang/DIPs/blob/master/DIPs/DIP1008.md

1) So I guess we should add a deprecation warning to clone the object for such cases?
2) Should we support `.dup` for the Exception class?

jacob-carlborg:
> Should we support .dup for the Exception class?

Ideally we should have a generic (deep) copy function that works for all types.

WalterBright:
Adding a warning wouldn't be helpful, as running the code will corrupt memory. Better to just leave it as an error. Need to fix the errors as they crop up.

wilzbach:
What should we do about:

```d
catch (shared(Abc) b)
```

Add an overload to `_d_delThrowable` or deprecate this?

> Better to just leave it as an error.

This would imply breaking changes for a few projects, no?

jacob-carlborg:
> This would imply breaking changes for a few projects, no?

As far as I understand, yes.

WalterBright:
> Add an overload to _d_delThrowable or deprecate this?

I think we should deprecated it. Throwing shared objects seems like it would be fraught with problems.

> This would imply breaking changes for a few projects, no?

Yes. That's why this should add `revert=dip1008` for those projects that need more time.

WalterBright (changes requested):
Need to add `-revert=dip1008` switch.

rainers:
Please note that the current implementation still allocates from the GC:

```d

int foo(int) @nogc
{
    throw new Exception("huhu");
}

void test() @nogc
{
    try
    {
        foo(1);
    }
    catch(Exception e)
    {
    }
}

void main()
{
    import std.stdio;
    import core.memory;
    auto stats1 = GC.stats();
    test();
    auto stats2 = GC.stats();
    writeln(stats1);
    writeln(stats2);
}
```
https://run.dlang.io/is/qP2hZg

src/dmd/cli.d
```diff
@@ -291,10 +291,6 @@ dmd -cov -unittest myprog.d
             "implement https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md",
             "implement $(LINK2 https://github.com/dlang/DIPs/blob/master/DIPs/DIP1000.md, DIP1000 (Scoped Pointers))"
         ),
-        Option("dip1008",
-            "implement https://github.com/dlang/DIPs/blob/master/DIPs/DIP1008.md",
```
AndrejMitrovic: Btw, this link was a 404. I don't know why there need to be 4 different DIP folders and the DIPs constantly being moved back and forth, it just keeps breaking links everywhere..

wilzbach:
(in case it wasn't clear from the "Orphaned" label: I unfortunately don't have any time in the forseeable future to invest time in this effort. I encourage anyone, however, to help out with this effort.)

test/compilable/testdip1008.d
```diff
@@ -1,5 +1,4 @@
 // PERMUTE_ARGS:
-// REQUIRED_ARGS: -dip1008
```
JohanEngelen: Not future-proof to remove this flag!

RazvanN7:
I have adopted this PR. Check: https://github.com/dlang/dmd/pull/10561

PR #8509, opened 2018-07-20
port blockopt.d to dmc

WalterBright:
blockopt.d has to work with both dmc and dmd.

### DIFF
--- a/src/dmd/backend/blockopt.d
+++ b/src/dmd/backend/blockopt.d
@@ -18,6 +18,11 @@
 version (SPP) {} else
 {
 
+version (SCPP)
+    version = COMPILE;
+else version (HTOD)
+    version = COMPILE;
+
 import core.stdc.stdio;
 import core.stdc.string;
 import core.stdc.time;
@@ -35,10 +40,11 @@ import dmd.backend.global;
 import dmd.backend.goh;
 import dmd.backend.code;
 import dmd.backend.ty;
-version (SCPP)
+version (COMPILE)
 {
 import parser;
 import iasm;
+import precomp;
 }
 
 
@@ -51,12 +57,17 @@ else
 
 extern(C++):
 
-extern(C) void *mem_fcalloc(size_t numbytes); // tk/mem.c
-void *mem_free(size_t numbytes); // tk/mem.c
+extern (C) void *mem_fcalloc(size_t numbytes); // tk/mem.c
+extern (C) void mem_free(void*); // tk/mem.c
 
-import dmd.backend.gflow : util_realloc;
+alias MEM_PH_FREE = mem_free;
 
-private __gshared
+version (HTOD)
+    void *util_realloc(void* p, uint n, uint size);
+else
+    import dmd.backend.gflow : util_realloc;
+
+__gshared
 {
 
 uint numblks;       // number of basic blocks in current function
@@ -86,7 +97,7 @@ pragma(inline, true) block *block_calloc_i()
         *b = blkzero;
     }
     else
-        b = cast(block *) mem_fcalloc(block.sizeof);
+        b = cast(block *) mem_calloc(block.sizeof);
     return b;
 }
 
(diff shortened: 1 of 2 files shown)


jacob-carlborg:
Why are these changes necessary? Is DMC using an older D host compiler?

src/dmd/backend/blockopt.d
```diff
@@ -51,12 +57,17 @@ else
 
 extern(C++):
 
-extern(C) void *mem_fcalloc(size_t numbytes); // tk/mem.c
-void *mem_free(size_t numbytes); // tk/mem.c
+extern (C) void *mem_fcalloc(size_t numbytes); // tk/mem.c
+extern (C) void *mem_free(void*); // tk/mem.c
```
wilzbach: The Visual Studio build doesn't like this:

```
..\dmd\backend\blockopt.d(61): error : Function type does not match previously declared function with the same mangled name: `mem_free` [C:\projects\dmd\src\vcbuild\dmd.vcxproj]
Done Building Project "C:\projects\dmd\src\vcbuild\dmd.vcxproj" (default targets) -- FAILED.
Done Building Project "C:\projects\dmd\src\vcbuild\dmd.vcxproj.metaproj" (default targets) -- FAILED.
Done Building Project "C:\projects\dmd\src\vcbuild\dmd.sln" (dmd target(s)) -- FAILED.
Build FAILED.
"C:\projects\dmd\src\vcbuild\dmd.sln" (dmd target) (1) ->
```

src/dmd/backend/blockopt.d
```diff
@@ -1025,12 +1045,19 @@ private void brrear()
...
+                            static if (NTEXCEPTIONS)
+                                enum additionalAnd = "b.Btry == bt.Btry &&
+                                                  bt.Btry == bt.nthSucc(0).Btry";
+                            else
+                                enum additionalAnd = "b.Btry == bt.Btry";
+                        }
                         else static if (NTEXCEPTIONS)
                             enum additionalAnd = "b.Btry == bt.Btry &&
-                                                  bt.Btry == bt.nthSucc(0).Btry";
+                                              bt.Btry == bt.nthSucc(0).Btry";
```
wilzbach: What's the motivation for this code duplication?

wilzbach:
> Why are these changes necessary?

A few changes are related to code that isn't compiled for DMD (and thus was never test/compiled during the blockopt.c<->blockopt.d conversion), not sure about the "B" prefix though.

src/dmd/backend/blockopt.d
wilzbach: The only other definitions are:

```
src/dmd/backend/go.d
93:extern (C) void mem_free(void* p);

src/dmd/backend/memh.d
26:void mem_free(void *);
```

src/dmd/backend/blockopt.d
```diff
@@ -18,6 +18,11 @@
 version (SPP) {} else
 {
 
+version (SCPP)
+    version = COMPILE;
```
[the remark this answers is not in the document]
WalterBright: It's not exactly redundant with it.

src/dmd/backend/blockopt.d
WalterBright: The lack of `||` for version statements.

src/dmd/backend/blockopt.d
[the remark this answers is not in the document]
thewilsonator: Oh `SPP` vs `S*C*PP`, my bad.

src/dmd/backend/blockopt.d
WalterBright: It shouldn't return a `void*`

WalterBright:
> not sure about the "B" prefix though.

You'll find it in `cc.h`.

src/dmd/backend/blockopt.d
wilzbach: Oh I was wondering about the fact that you duplicated the strings too, because there are a few ways to avoid this like:

```d
version(SCPP)
    enum SCPP = 1;
```

wilzbach (approved):

PR #8511, opened 2018-07-21
convert cgreg.c to D

WalterBright:


WalterBright:
It'd be nice if semaphoreci gave a better message than "execution expired", if that is even what the problem is.

jacob-carlborg (approved):

PR #8514, opened 2018-07-23
Make `TemplateParameter` inherit from `RootObject` + minor fixes

Geod24:
Split from #8455 because it still needs work, but those are rather simple and can already be applied.
The change to `TemplateParameter` should make it easier for users of DMD as a library.


ibuclaw:
Seems reasonable in the context of the other PR, though there are no users of any of the added members here. Are you sure that this will remain in the final version of the implementation you're working on?

Geod24:
Hugh, second time that libmir fails with `Killed`. Is that frequent @wilzbach ?

Geod24:
> though there are no users of any of the added members here

I guess you're mentioning `TemplateParameter`, right ? Since `writeIdentifier` is used in 2 places (although I usually go by the rule of 3, but I'm sure it's going to be used in the final version of the other PR).

I am pretty certain making `TemplateParameter` inherit from `RootObject` is a good move. AFAIK it's the only class in the AST that does not derive from `RootObject`, and it sometimes makes things a bit more complicated (e.g. no `toChars()` member is annoying for debugging).

When I was working with DMD as a library a few years ago, it's also one of the first change I needed to for most use case I had, hence why I decided to make this separate PR.

RazvanN7 (approved):
Makes sense to me.

wilzbach:
> Second time that libmir fails with Killed. Is that frequent @wilzbach ?

Well, DMD requires a ton of memory and I currently use two cheap VPS of mine to host the buildkite agents and memory can often get sparse (though each of them has 6 GB).
We hopefully can add more powerful instances to Buildkite once Martin is okay with terminating the Jenkins server and doing the full switch to Buildkite (i.e. once the logs are fully public :/).

test/runnable/cppa.d
```diff
@@ -1230,7 +1230,7 @@ void test15455()
 /****************************************/
 // 15372
 
-extern(C++) int foo15372(T)(T v);
+extern(C++) int foo15372(T)(int v);
```
wilzbach: Why is this change required?

Geod24: It's not required for `TemplateParameter`, but I figured I'd include it here since it's so trivial. I always make atomic commits, so you can find the explanation in the commit message.

JinShil (approved):

src/dmd/dtemplate.d
```diff
@@ -99,6 +99,13 @@ extern (C++) inout(Parameter) isParameter(inout RootObject o)
     return cast(inout(Parameter))o;
 }
 
+extern (C++) inout(TemplateParameter) isTemplateParameter(inout RootObject o)
+    {
+        if (!o || o.dyncast() != DYNCAST.templateparameter)
+            return null;
+        return cast(inout(TemplateParameter))o;
+    }
```
WalterBright: Hmm, this is completely dead code.

src/dmd/dtemplate.d
```diff
@@ -5095,6 +5102,12 @@ extern (C++) class TemplateParameter
 
     abstract bool hasDefaultArg();
 
+    override const(char)* toChars() { return this.ident.toChars(); }
+    override DYNCAST dyncast() const pure @nogc nothrow @safe
+    {
+        return DYNCAST.templateparameter;
```
WalterBright: also all dead code. Can you write a test case for it?

Geod24: It will be used in a later PR

src/dmd/dtemplate.d
Geod24: Same, it will be used in a later PR.

PetarKirov (approved):
LGTM as well. I trust @Geod24 that he we'll put the currently dead code to a good use in a future PR.

PR #8515, opened 2018-07-23
convert debug.c to debugprint.d

WalterBright:


src/dmd/backend/debugprint.d
```diff
@@ -0,0 +1,486 @@
...
+            printf(" b_ret=B%d", b.b_ret ? b.b_ret.Bnumber : 0);
+version (MARS)
+{
+        if (b.Bsrcpos.Sfilename)
+            printf(" %s(%u)", b.Bsrcpos.Sfilename, b.Bsrcpos.Slinnum);
+}
+        printf("\n");
+        if (b.Belem)
+        {
+            if (debugf)
```
wilzbach: First rule of conversion ;-) (it's originally `0`)

src/dmd/backend/debugprint.d
```diff
@@ -0,0 +1,486 @@
...
+                        printf(" B%d",list_block(bl).Bnumber);
+                    printf("\n");
+                }
+                break;
+            case BCret:
+            case BCretexp:
+            case BCexit:
+                break;
+            default:
+                printf("bc = %d\n", b.BC);
```
wilzbach: First rule of conversion: this isn't part of the CXX version.

src/dmd/backend/debugprint.d
```diff
@@ -0,0 +1,486 @@
...
+        switch (e.Eoper)
+        {   case OPconst:
+                elem_print_const(e);
+                break;
+            case OPrelconst:
+                ferr("#");
+                goto case OPvar;
+
+            case OPvar:
+                printf("%s",e.EV.Vsym.Sident.ptr);
```
wilzbach: This looks a bit dangerous. It doesn't look `char[1]` is null-terminated in the default cause, e.g. https://run.dlang.io/is/VHJOtq

Maybe "%c" or being length-specific with `%.*s`?

src/dmd/backend/debugprint.d
WalterBright: Eh, must be my copy of debug.c differed slightly. No matter.

src/dmd/backend/debugprint.d
WalterBright: same as other comment.

src/dmd/backend/debugprint.d
WalterBright: Sident is always 0 terminated, that's why the original being just Sident worked.

wilzbach (approved):

PR #8516, opened 2018-07-23
fix issue 19112 - Associative array opIn with static array key fails …

aG0aep6G:
…with dynamic array


RazvanN7 (approved):

wilzbach:
(DAutoTest and Jenkins failures are unrelated :/)

jacob-carlborg:
This is a breaking change and should ideally be a deprecation instead.

aG0aep6G:
> This is a breaking change and should ideally be a deprecation instead.

What is it breaking? The code that is now rejected was producing garbage before.

wilzbach:
> This is a breaking change and should ideally be a deprecation instead.

I disagree. Our policy is that we can (and should do) breaking changes for broken code that has never and will never work.

wilzbach (approved):

jacob-carlborg (approved):

PR #8517, opened 2018-07-23
Informative Assert Messages

wilzbach:
So I presume I probably have to write a DIP for this, but as there are many open question I thought I already start hacking with DMD to see what they are and how potential solutions could look like.

Motivation
-------------

__Informative__ assert messages that tell you why sth. failed. This is incredibly useful when the testsuite isn't run on your machine, but on a CI server or the assertion failed in production and you don't have a coredump.

tl;dr:

```d
int a = 1, b = 2;
assert(a == b); // ERROR: 1 != 2 
```

Also better just assert diagnostics is the number #7 features that people miss according to the recent State of D survey:

image

Moreover, a better debugging experience (and in fact DIP83) has been part of the bi-annual high level visions [since 2017](https://wiki.dlang.org/Vision/2017H2).

Lastly, this is based on top of https://github.com/dlang/dmd/pull/7575 which would already solve two very old issues.

What's the status quo?
----------------------------

```
core.exception.AssertError@onlineapp.d(4): Assertion failure
----------------
??:? _d_assertp [0x79a0dad5]
onlineapp.d:4 _Dmain [0x79a0da08]
```

https://run.dlang.io/is/b3RHNb

(for the example above)

Previous work:
------------------

- [Original DIP83](https://wiki.dlang.org/DIP83)
- [Improving assert-printing in DMD NG thread from 2015](https://forum.dlang.org/post/holdxspayjguauomrbcx@forum.dlang.org)
- [Original DMD PR](https://github.com/dlang/dmd/pull/263)
- [Issue 5547 - Improve assert to give information on values given to it when it fails](https://issues.dlang.org/show_bug.cgi?id=5547#c3)
- [assertPred - druntime PR](https://github.com/dlang/druntime/pull/41)

Main problems pointed out in the past
------------------------------------------------

1) druntime can't use `std.conv.to` or anything from Phobos.

Solution: some debug output is a lot better than none. All basic types can be formatted by libc, structs could just be recursively iterated over all their basic fields and for classes `.to!string` only calls `.toString` anyhow. Even just basic types would already be a huge improvement.

2) Performance / Bloat

As all formatting is done only in the non-release build and the actual string formatting happens lazily when the error is triggered, there's no performance overhead.
However, if a templated `_d_assert_fail` is chosen, there might be a bit of template bloat when a project uses many asserts with different types.
If this is really a concern, a minimal implementation could only go for basic types or define a  `extern(C)` interface. Though there's already lots of templated stuff in druntime, e.g. `destroy`,  `__delete`, `__cmp`, `__ArrayEq`, `__ArrayPostblit`, `__ArrayDtor`, `.dup`, `.idup`, `__switch`, `_arrayOp`, `hashOf`, `assumeSafeAppend`, `capacity`, `_postblitRecurse`, `update`, `require`, `get`, `values`, `keys`, `by{Key,Value,KeyValue}`, `rehash`, ...

3) This is yet another compiler magic

Yes, but as `assert` is already baked into the language, let's make use of the fact and by a moderately simple change we can vastly improve the debugging/testing experience of D.

Open questions
--------------------

This PR directly hacks the `lazy` string message that is provided to the `AssertError`. This has the advantage of making things easy. There could be a point made this isn't ideal because:

1) it makes lots of code that is currently `@nogc`, use the garbage collection for string formatting (and is potentially slower than directly printing the warning)

My take on this is that the cost only happens at most once and after which the program will be terminated anyhow (and catching errors is defined to result in UB). In fact, I think that these formatting calls can be marked as "@nogc" because either we can just use malloc and not worry about freeing or simply do the same with D's built-in concatenation.

2) Support/Integration with custom test runners  / custom formatting backends

Currently, custom test runners like [unit-threaded](https://github.com/atilaneves/unit-threaded) or [fluent-asserts](https://github.com/gedaiu/fluent-asserts) come with their own TDD or BDD style to work around the problem of not being able to catch the values of `assert(a == b)`, with e.g. `foo.should.equal(2)`

3) Name of the druntime function

TODO
--------

- [x] Add formatting support for more than ints (it's a PoC for now)
- [x] Add support for more than EqualExp
- [x] For now this is hidden behind a flag to simplify testing
- [x] Pass the token type to the formatting helper e.g. `_d_assert_failed!"=="(a, b)`
- [x] Add more tests
- [x] Check review from https://github.com/dlang/dmd/pull/7575
- [x] Move implementation `_d_assert_fail` to druntime

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -5512,6 +5512,9 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             printf("AssertExp::semantic('%s')\n", exp.toChars());
         }
 
+        // save expression as a string before any semantic expansion
+        auto assertExpMsg = exp.msg ? null : exp.toChars();
+
         if (Expression ex = unaSemantic(exp, sc))
         {
             result = ex;
@@ -5522,9 +5525,84 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         exp.e1 = exp.e1.optimize(WANTvalue);
         exp.e1 = exp.e1.toBoolean(sc);
 
+        if (!exp.msg && global.params.checkAction == CHECKACTION.context)
+        // no message - use assert expression as msg
+        {
+            /*
+            {
+              auto a = e1, b = e2;
+              assert(a == b, _d_assert_fail!"=="(a, b));
+            }()
+            */
+            const tok = exp.e1.op;
+            bool isEqualsCallExpression;
+            if (tok == TOK.call)
+            {
+                const callExp = cast(CallExp) exp.e1;
+                const callExpIdent = callExp.f.ident;
+                isEqualsCallExpression = callExpIdent == Id.__equals ||
+                                         callExpIdent == Id.eq;
+            }
+            if (tok == TOK.equal || tok == TOK.notEqual ||
+                tok == TOK.lessThan || tok == TOK.greaterThan ||
+                tok == TOK.lessOrEqual || tok == TOK.greaterOrEqual ||
+                tok == TOK.identity || tok == TOK.notIdentity ||
+                isEqualsCallExpression)
+            {
+                auto es = new Expressions();
+                auto tiargs = new Objects();
+                Loc loc = exp.e1.loc;
+
+                if (isEqualsCallExpression)
+                {
+                    auto callExp = cast(CallExp) exp.e1;
+                    auto args = callExp.arguments;
+
+                    // template args
+                    static immutable compMsg = "==";
+                    Expression comp = new StringExp(loc, cast(char*) compMsg.ptr);
+                    comp = comp.expressionSemantic(sc);
+                    tiargs.push(comp);
+                    tiargs.push((*args)[0].type);
+                    tiargs.push((*args)[1].type);
+
+                    // runtime args
+                    es.push((*args)[0]);
+                    es.push((*args)[1]);
+                }
+                else
+                {
+                    auto binExp = cast(EqualExp) exp.e1;
+
+                    // template args
+                    Expression comp = new StringExp(loc, cast(char*) Token.toChars(exp.e1.op));
+                    comp = comp.expressionSemantic(sc);
+                    tiargs.push(comp);
+                    tiargs.push(binExp.e1.type);
+                    tiargs.push(binExp.e2.type);
+
+                    // runtime args
+                    es.push(binExp.e1);
+                    es.push(binExp.e2);
+                }
+
+                Expression __assertFail = new IdentifierExp(exp.loc, Id.empty);
+                auto assertFail = new DotIdExp(loc, __assertFail, Id.object);
+
+                auto dt = new DotTemplateInstanceExp(loc, assertFail, Id._d_assert_fail, tiargs);
+                auto ec = CallExp.create(Loc.initial, dt, es);
+                exp.msg = ec;
+            }
+            else
+            {
+                OutBuffer buf;
+                buf.printf("%s failed", assertExpMsg);
+                exp.msg = new StringExp(Loc.initial, buf.extractString());
+            }
+        }
         if (exp.msg)
         {
-            exp.msg = exp.msg.expressionSemantic(sc);
+            exp.msg = expressionSemantic(exp.msg, sc);
             exp.msg = resolveProperties(sc, exp.msg);
             exp.msg = exp.msg.implicitCastTo(sc, Type.tchar.constOf().arrayOf());
             exp.msg = exp.msg.optimize(WANTvalue);
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -185,7 +185,7 @@ struct Usage
                     $(LI $(B off): specified check is disabled.)
                 )`
         ),
-        Option("checkaction=D|C|halt",
+        Option("checkaction=D|C|halt|context",
             "behavior on assert/boundscheck/finalswitch failure",
             `Sets behavior when an assert fails, and array boundscheck fails,
              or a final switch errors.
@@ -193,6 +193,7 @@ struct Usage
                     $(LI $(B D): Default behavior, which throws an unrecoverable $(D Error).)
                     $(LI $(B C): Calls the C runtime library assert failure function.)
                     $(LI $(B halt): Executes a halt instruction, terminating the program.)
+                    $(LI $(B context): Prints the error context as part of the unrecoverable $(D Error).)
                 )`
         ),
         Option("color",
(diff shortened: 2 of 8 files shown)


timotheecour:
in nim, this is relatively easily implemented in library code via macros:

```
inim
INim 0.3.0
Nim Compiler Version 0.18.1 [MacOSX: amd64] at
/Users/timothee/.nimble/bin/nim
>>> import unittest
>>> check 1+3 < 1+1
Check failed: 1 + 3 < 1 + 1
1 + 3 was 4
1 + 1 was 2
Error
```

wilzbach:
> Is it necessary the analyze the expression?
> What I mean is can't you just plop exp.toChars() into the assert message?

That would lead to:

```
assert(a == b); // ERROR: a != b 
```

Not much more helpful than before. You still are in bad luck if the reason for the crash is hard to replicate or simply it happened on a CI test suite or in a production server.

wilzbach:
> Is it necessary the analyze the expression?
> What I mean is can't you just plop exp.toChars() into the assert message?

That would lead to:

```
assert(a == b); // ERROR: a != b 
```

Not much more helpful than before. You still are in bad luck if the reason for the crash is hard to replicate or simply it happened on a CI test suite or in a production server.

> The number of expressions that will need to be analyzed seems enormous.

I will do benchmarking on this once I get back to it.

andralex:
This is nice. What's the status? FWIW it doesn't require a DIP or a flag - we never specified the contents of assertion messages (and we shouldn't).

wilzbach:
> This is nice. What's the status?

I just took some time to rework this and give you all a Christmas present.
(i.e. I addressed the main points that were blocking this from moving forward)

In short: https://github.com/dlang/druntime/pull/2413 needs to be merged (this as `_d_assert_fail` to `object.d`) and then we can move forward here.

This still needs some benchmarking and a bit of further work before it can be activated by default, but I hope it's now in a shape where it can be tested behind the CLI flag and be improved by others as this is a lot better than slowly dying in the PR queue and as long as its behind the flag it won't affect anyone.

jacob-carlborg:
> I addressed the main points that were blocking this from moving forward

Really! Did you implement AST macros? 😃

WalterBright:
Should integrate with `-checkaction` switch:

 https://github.com/dlang/dmd/pull/8718

src/dmd/expressionsem.d
```diff
@@ -5521,13 +5524,87 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+
+                auto dt = new DotTemplateInstanceExp(loc, assertFail, Id._d_assert_fail, tiargs);
+                auto ec = CallExp.create(Loc.initial, dt, es);
+                exp.msg = ec;
+            }
+            else
+            {
+                OutBuffer buf;
+                buf.printf("%s failed", assertExpMsg);
+                exp.msg = new StringExp(Loc.initial, buf.extractString());
```
wilzbach: This originates from https://github.com/dlang/dmd/pull/7575, but leads to the iasm failures on 32-bit as foo isn't on EBP anymore:

```diff
 Disassembly of section .text._D3foo5test1FZv:
@@ -7,23 +7,31 @@ Disassembly of section .text._D3foo5test
 00000000 <_D3foo5test1FZv>:
    0:	55                   	push   ebp
    1:	8b ec                	mov    ebp,esp
-   3:	83 ec 08             	sub    esp,0x8
-   6:	31 c0                	xor    eax,eax
-   8:	89 45 f8             	mov    DWORD PTR [ebp-0x8],eax
-   b:	89 45 fc             	mov    DWORD PTR [ebp-0x4],eax
-   e:	90                   	nop
-   f:	90                   	nop
-  10:	b8 08 00 00 00       	mov    eax,0x8
-  15:	89 45 f8             	mov    DWORD PTR [ebp-0x8],eax
-  18:	83 7d f8 08          	cmp    DWORD PTR [ebp-0x8],0x8
-  1c:	74 0d                	je     2b <_D3foo5test1FZv+0x2b>
-  1e:	6a 0f                	push   0xf
-  20:	b9 00 00 00 00       	mov    ecx,0x0
-  25:	51                   	push   ecx
-  26:	e8 fc ff ff ff       	call   27 <_D3foo5test1FZv+0x27>
-  2b:	c9                   	leave  
-  2c:	c3                   	ret    
-  2d:	00 00                	add    BYTE PTR [eax],al
+   3:	83 ec 10             	sub    esp,0x10
+   6:	89 5d f0             	mov    DWORD PTR [ebp-0x10],ebx
+   9:	89 75 f4             	mov    DWORD PTR [ebp-0xc],esi
+   c:	31 c0                	xor    eax,eax
+   e:	89 45 f8             	mov    DWORD PTR [ebp-0x8],eax
+  11:	89 45 fc             	mov    DWORD PTR [ebp-0x4],eax
+  14:	b8 10 00 00 00       	mov    eax,0x10
+  19:	89 45 f8             	mov    DWORD PTR [ebp-0x8],eax
+  1c:	83 7d f8 08          	cmp    DWORD PTR [ebp-0x8],0x8
+  20:	74 1f                	je     41 <_D3foo5test1FZv+0x41>
+  22:	6a 0f                	push   0xf
+  24:	b9 00 00 00 00       	mov    ecx,0x0
+  29:	bb 05 00 00 00       	mov    ebx,0x5
+  2e:	51                   	push   ecx
+  2f:	53                   	push   ebx
+  30:	ba 06 00 00 00       	mov    edx,0x6
+  35:	be 17 00 00 00       	mov    esi,0x17
+  3a:	52                   	push   edx
+  3b:	56                   	push   esi
+  3c:	e8 fc ff ff ff       	call   3d <_D3foo5test1FZv+0x3d>
+  41:	8b 5d f0             	mov    ebx,DWORD PTR [ebp-0x10]
+  44:	8b 75 f4             	mov    esi,DWORD PTR [ebp-0xc]
+  47:	c9                   	leave  
+  48:	c3                   	ret    
+  49:	00 00                	add    BYTE PTR [eax],al
```

Are we supposed to cater for this?

wilzbach:
> Should integrate with -checkaction switch:

Done.

> In short: dlang/druntime#2413 needs to be merged (this as _d_assert_fail to object.d) and then we can move forward here.

No longer true, as this is now is an entirely experimental, opt-in feature, so this can be merged before druntime (and it's indeed green on the CIs) and then we can enable the actual `_d_assert_fail` logic, tests and changelog at druntime.

Anything else that needs to be done before we can start the first part of this experiment?

thewilsonator (approved):

wilzbach:
For the impatient: changelog entry, examples and the final bits can be found at the respective druntime PR: https://github.com/dlang/druntime/pull/2413 (with `digger` you can even test these druntime changes for playing/experimenting with this locally).

AndrejMitrovic:
Well done on finally getting this feature through. Nice work @wilzbach!

PR #8518, opened 2018-07-24
convert type.c to dtype.d

WalterBright:


src/dmd/backend/dtype.d
```diff
@@ -0,0 +1,1754 @@
+/**
```
jacob-carlborg: Shouldn't this file replace the existing `type.d`?

WalterBright: There's already a type.d.

jacob-carlborg: I know, but looks like bindings to type.c.

WalterBright: It is. Unfortunately, I gave up on using .di/.d files for the compiler source because it broke all the different build systems for dmd. Right now, I'm only interested in getting everything in D in the most straightforward way practical before reorganizing it.

jacob-carlborg: Do we need .di files?

wilzbach (approved):

src/dmd/backend/dtype.d
```diff
@@ -0,0 +1,1754 @@
...
+/*************************************
+ * Allocate a TYtemplate.
+ */
+
+version (SCPP_HTOD)
+{
+type *type_alloc_template(Symbol *s)
+{   type *t;
+
+    t = cast(type *) mem_fmalloc(typetemp_t.sizeof);
```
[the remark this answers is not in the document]
WalterBright: The whole mem_xxxx code in the back end is an utter disaster.

src/dmd/backend/dtype.d
WalterBright: To separate definition from implementation, yes.

jacob-carlborg: We don’t have that for the front end why do you think we need that for the back end?

WalterBright: The most pragmatic answer is that's the way the C++ code was organized, and translating works a lot better if you don't try and refactor it in the process.

PR #8443, opened 2018-07-03
Turn deprecation into error for issue 313

RazvanN7:

PR #8519, opened 2018-07-26
Fix Issue 19107 - -de produces compilation error, -dw does not

JinShil:
Due to the error gagging, deprecations are not being displayed in `typeSemantic`.  This PR remedies that by running `typeSemantic` again without gagging if it previously succeeded.  It's not a very elegant solution, but I don't think there's any other way without a major DMD refactor.

See discussion in the following PRs for more context:
https://github.com/CyberShadow/ae/pull/33
https://github.com/dlang/dmd/pull/8443

cc @CyberShadow

### DIFF
--- a/src/dmd/errors.d
+++ b/src/dmd/errors.d
@@ -299,12 +299,18 @@ extern (C++) void verrorSupplemental(const ref Loc loc, const(char)* format, va_
  */
 extern (C++) void vwarning(const ref Loc loc, const(char)* format, va_list ap)
 {
-    if (global.params.warnings && !global.gag)
+    if (global.params.warnings)
     {
-        verrorPrint(loc, Classification.warning, "Warning: ", format, ap);
-        //halt();
-        if (global.params.warnings == 1)
-            global.warnings++; // warnings don't count if gagged
+        if (!global.gag)
+        {
+            verrorPrint(loc, Classification.warning, "Warning: ", format, ap);
+            if (global.params.warnings == 1)
+                global.warnings++;
+        }
+        else
+        {
+            global.gaggedWarnings++;
+        }
     }
 }
 
@@ -335,8 +341,17 @@ extern (C++) void vdeprecation(const ref Loc loc, const(char)* format, va_list a
     static __gshared const(char)* header = "Deprecation: ";
     if (global.params.useDeprecated == 0)
         verror(loc, format, ap, p1, p2, header);
-    else if (global.params.useDeprecated == 2 && !global.gag)
-        verrorPrint(loc, Classification.deprecation, header, format, ap, p1, p2);
+    else if (global.params.useDeprecated == 2)
+    {
+        if (!global.gag)
+        {
+            verrorPrint(loc, Classification.deprecation, header, format, ap, p1, p2);
+        }
+        else
+        {
+            global.gaggedWarnings++;
+        }
+    }
 }
 
 /**
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -264,6 +264,7 @@ struct Global
     uint warnings;          // number of warnings reported so far
     uint gag;               // !=0 means gag reporting of errors & warnings
     uint gaggedErrors;      // number of errors reported while gagged
+    uint gaggedWarnings;    // number of warnings reported while gagged
 
     void* console;         // opaque pointer to console for controlling text attributes
 
@@ -275,6 +276,7 @@ struct Global
     extern (C++) uint startGagging()
     {
         ++gag;
+        gaggedWarnings = 0;
         return gaggedErrors;
     }
 
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -909,11 +909,22 @@ extern (C++) abstract class Type : RootObject
     final Type trySemantic(const ref Loc loc, Scope* sc)
     {
         //printf("+trySemantic(%s) %d\n", toChars(), global.errors);
+
+        // Needed to display any deprecations that were gagged
+        auto tcopy = this.syntaxCopy();
+
         uint errors = global.startGagging();
         Type t = typeSemantic(this, loc, sc);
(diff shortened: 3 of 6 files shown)


CyberShadow:
Have you checked if this has any impact on performance? Not a dmd expert but performing semantics twice sounds like might be expensive.

As an alternative, would it be possible to count if any deprecation messages *would* have been emitted during gagging, and only run the semantics a second time in that case?

JinShil:
> Have you checked if this has any impact on performance? Not a dmd expert but performing semantics twice sounds like might be expensive.

It appears `Type.trySemantic` is only called in one place in the compiler for `is` expressions:  https://github.com/dlang/dmd/blob/750f0245c2882887cdb830c171e9221719e0da28/src/dmd/expressionsem.d#L4791, so I believe the impact would be relatively low, but I haven't measured it, and not sure if I reliably could.

> As an alternative, would it be possible to count if any warning/deprecation messages would have been emitted during gagging, and only run the semantics a second time in that case?

Yeah, I think I can do that.  Other reviewers, please let me now if that is preferred.

JinShil:
With https://github.com/dlang/phobos/pull/6642 merged (Thank you!), how can I cause this PR to pick up the new bits?  I thought it would do it automatically.  Do I need to wait, or initiate a new commit hash, or just close and re-open the PR?

JinShil:
I just found out this same pattern, running semantic twice, is used elsewhere in the compiler:

https://github.com/dlang/dmd/blob/750f0245c2882887cdb830c171e9221719e0da28/src/dmd/typesem.d#L3439-L3458

wilzbach:
There's `global.showGaggedErrors` why not use it for deprecations too?

wilzbach:
> Do you mean global.params.showGaggedErrors? That's enabled for -verrors=spec, and I think is intended a different purpose. If I'm misunderstanding please clarify.

Sorry, I was trying to suggest instead of counting for the warnings (and then re-running the semantic), you could immediately print them directly if a respective global flag like `global.params. showDeprecationMessages` is set?

(It assumed that was what `showGaggedErrors` was intended to do. Sorry)

wilzbach:
Yes I understood that, but I was talking about the fact that there's only one place where `gaggedWarnings` is checked, it's known before the `startGagging` call a few lines above, that the deprecations should be printed which is why @JinShil currently reruns `typeSemantic` without gagging.
So running `startGagging` once with a special flag should avoid the need to run it twice in this scenario, no?

JinShil:
I don't think that will work because you only want to show deprecation messages if the entire `typeSemantic` call succeeds.  You won't know that until it's finished.

JinShil:
@CyberShadow What's up with DAutoTest?  Is it a dub failure?

CyberShadow:
Definitely a Dub failure, though never saw it print an error in that format before. @wilzbach Do you know?

```
No versions for ddoc_preprocessor for registry at https://code.dlang.org/ (fallback ["registry at http://code.dlang.org/", "registry at https://code-mirror.dlang.io/", "registry at https://code-mirror2.dlang.io/", "registry at https://dub-registry.herokuapp.com/"])
```

It looks like it tried all those mirrors already? Also I thought treating network failures as "No versions" was due to be fixed. Or was it not a network failure?

wilzbach:
No `ddoc_preprocessor` is a local `path` dependency, so it can't even be downloaded from the registry (no idea why dub tries anyhow), but it looks more like a problem with the version resolver:

```
No versions for ddoc_preprocessor for registry at https://code.dlang.org/ (fallback ["registry at http://code.dlang.org/", "registry at https://code-mirror.dlang.io/", "registry at https://code-mirror2.dlang.io/", "registry at https://dub-registry.herokuapp.com/"])
Return for ddoc_preprocessor: [~master]
Search for versions of dmd (1 package suppliers)
Return for dmd: [~master, ~stable, ~newCTFE, ~markdown, ~dcache]
The dependency resolution process is taking too long. The dependency graph is likely hitting a pathological case in the resolution algorithm. Please file a bug report at https://github.com/dlang/dub/issues and mention the package recipe that reproduces this error.
posix.mak:762: recipe for target 'dpl-docs' failed
```

Not sure why it spuriously appears, but maybe https://github.com/dlang/dub/pull/1513 helps at least in the future.

JinShil:
@CyberShadow I've tried re-running DAutoTest several times, and it just won't pass.  Other PRs I've submitted pass fine.  Could you please dig a little bit and see if you can find anything?

wilzbach:
> @wilzbach Any idea if anything between that success/failure could have introduced a regression?

No. Sorry, we haven't done much at dlang.org (https://github.com/dlang/dlang.org/commits/master) in the last weeks.

However, I agree that this is unrelated - I have seen this error way too often over the last week :/
Maybe we can get away with not running `dub upgrade --missing-only --root=dpl-docs` before we run `dub build`?

wilzbach:
FYI: this will also help https://github.com/dlang/dmd/pull/7241 to move forward.

RazvanN7:
> I just found out this same pattern, running semantic twice, is used elsewhere in the compiler:

On the long run we should probably use a more elegant solution, like having a queue of gagged errors/deprecations/warnings and just print them without having to copy parts of the AST and perform semantic twice. This would negligibly  increase the memory usage, but it will increase performance.

wilzbach:
A naive attempt at fixing our problem by bumping the dub version used at dlang.org: https://github.com/dlang/dlang.org/pull/2432

wilzbach:
Seems like DAutoTest is entirely broken for `stable`, but as this should go into 2.082 and not 2.081.2 anyhow, I rebased this, s.t. we can move forward with this.
Reasoning for introducing this in 2.082: we shouldn't deprecate things in patch updates (that's pretty much against SemVer) and a user only expects stability patches/improvements/fixes when they upgrade a patch version.

RazvanN7 (approved):

wilzbach (approved):

PR #8520, opened 2018-07-26
convert symbol.c to D

WalterBright:


src/dmd/backend/symbol.d
```diff
@@ -0,0 +1,2454 @@
...
+            (vol || !(s.Stype.Tty & mTYvolatile)) &&
+
+            (config.flags4 & CFG4optimized || !config.fulltypes));
+}
+
+/****************************************
+ * Determine if symbol needs a 'this' pointer.
+ */
+
+int Symbol_needThis(Symbol* s)
```
JinShil: This looks new.

src/dmd/backend/symbol.d
```diff
@@ -0,0 +1,2454 @@
...
+debug
+{
+    int count = 0;
+
+    if (flag) symbol_tree_check(*parent);
+}
+    for (; *psx; psx = &s.Snext)
+    {
+        //printf("%p ",*psx);
+debug
```
JinShil: Yuck! Needs curly braces.

src/dmd/backend/symbol.d
```diff
@@ -0,0 +1,2454 @@
...
+            c = *p;
+
+            // Put symbol s into symbol table
+
+static if (MMFIO)
+{
+            if (s.Sl || s.Sr)         // avoid writing to page if possible
+                s.Sl = s.Sr = null;
+}
+else
```
JinShil: Again, lack of curly braces here is a little jarring.

JinShil (approved):
That was painful; started losing my patience towards the end.  I don't know what `TX86` is and why it's omitted, but other than that I couldn't find anything to complain about.

src/dmd/backend/symbol.d
WalterBright: That's because the original was Symbol::needThis(). D doesn't allow member functions to be defined outside of the aggregate, so this was the easiest translation.

src/dmd/backend/symbol.d
WalterBright: I try to minimize diffs, leaving reformatting for another PR.

PR #8523, opened 2018-07-28
-betterC: Remove illegitimate dependency on TypeInfo for arrays of structs

JinShil:
When using arrays of structs, the compiler was complaining about implicit usages of `TypeInfo` for array literals and comparisons, but the `TypeInfo` is never actually needed.  This stops the whining.

### DIFF
--- a/src/dmd/dcast.d
+++ b/src/dmd/dcast.d
@@ -144,7 +144,7 @@ extern (C++) Expression implicitCastTo(Expression e, Scope* sc, Type t)
             visit(cast(Expression)e);
 
             Type tb = result.type.toBasetype();
-            if (tb.ty == Tarray)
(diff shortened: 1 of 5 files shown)


jacob-carlborg (approved):
Should there be a changelog entry as well?

wilzbach (approved):
Nice!

wilzbach:
> Should there be a changelog entry as well?

In the past, we grouped a few of such changes together in a changelog, like e.g. https://dlang.org/changelog/2.079.0.html#minimal_runtime or more recently https://dlang.org/changelog/2.081.0.html#min_runtime_classes

tl;dr: A changelog entry never hurts and is free advertisement of the active development, but I don't think it's required for this small change.

ibuclaw:
Small change, big impact.

PR #8525, opened 2018-07-29
Add D_ModuleInfo D_Exceptions and D_TypeInfo version identifiers

JinShil:
I'm going to begin converting some runtime hooks to templates so they can be used in -betterC and other scenarios.  However, some of the existing implementations throw `Error`.  I need to special case those for -betterC to `assert` instead.

Examples: 
* [`core.exception.__switch_errorT`](https://github.com/dlang/druntime/blob/52d3fe02272d16d32c150ce6f78bc00241a9dd5d/src/core/exception.d#L584)
* [`_d_arrayCopy`](https://github.com/dlang/druntime/blob/9a8edfb48e4842180c706ee26ebd8edb10be53f4/src/rt/arraycat.d#L27) which throws through [`enforceRawArraysConformable`](https://github.com/dlang/druntime/blob/9a8edfb48e4842180c706ee26ebd8edb10be53f4/src/rt/util/array.d#L26)

I think it would be best to have a little more precision when doing runtime programming by mimicking the current implementation of the compiler with respect to [`global.params.useModuleInfo`](https://github.com/dlang/dmd/blob/033d324957ff5bee8e25c7335eab2c562d17e2b1/src/dmd/globals.d#L142), [`global.params.useExceptions`](https://github.com/dlang/dmd/blob/033d324957ff5bee8e25c7335eab2c562d17e2b1/src/dmd/globals.d#L144), and [`global.params.useTypeInfo`](https://github.com/dlang/dmd/blob/033d324957ff5bee8e25c7335eab2c562d17e2b1/src/dmd/globals.d#L143)

Also, this may be useful for other D compilers that expose compiler flags such as `-fno-exceptions`, and `fno-rtti`.

Spec change: https://github.com/dlang/dlang.org/pull/2431

cc @ibuclaw @klickverbot

### DIFF
--- a/test/compilable/betterCarray.d
+++ b/test/compilable/betterCarray.d
@@ -15,12 +15,3 @@ int foo(int[] a, int i)
 {
     return a[i];
 }
-
-// https://issues.dlang.org/show_bug.cgi?id=17787
-version (D_BetterC)
-{
-}
-else
-{
-    static assert(0);
-}
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1388,7 +1388,16 @@ void addDefaultVersionIdentifiers()
     if (global.params.useArrayBounds == CHECKENABLE.off)
         VersionCondition.addPredefinedGlobalIdent("D_NoBoundsChecks");
     if (global.params.betterC)
+    {
         VersionCondition.addPredefinedGlobalIdent("D_BetterC");
+    }
+    else
+    {
+        VersionCondition.addPredefinedGlobalIdent("D_ModuleInfo");
+        VersionCondition.addPredefinedGlobalIdent("D_Exceptions");
+        VersionCondition.addPredefinedGlobalIdent("D_TypeInfo");
+    }
+
 
     VersionCondition.addPredefinedGlobalIdent("D_HardFloat");
 }
--- a/test/compilable/betterc.d
+++ b/test/compilable/betterc.d
@@ -1,0 +1,27 @@
+/* REQUIRED_ARGS: -betterC
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=17787
+version (D_BetterC)
+{
+}
+else
+{
+    static assert(0);
+}
+
+// -betterC does not support `ModuleInfo`, `TypeInfo`, or exception handling
+version (D_ModuleInfo)
+{
+    static assert(0);
+}
+
+version (D_Exceptions)
+{
+    static assert(0);
+}
+
+version (D_TypeInfo)
+{
+    static assert(0);
+}
--- a/test/compilable/version.d
+++ b/test/compilable/version.d
@@ -1,0 +1,23 @@
+/* REQUIRED_ARGS:
+*/
+
+version (D_ModuleInfo)
+{ }
+else
+{
+    static assert(0);
+}
+
+version (D_Exceptions)
+{ }
+else
+{
+    static assert(0);
+}
+
+version (D_TypeInfo)
+{ }
+else
+{
+    static assert(0);
+}


test/compilable/betterCarray.d
```diff
@@ -15,12 +15,3 @@ int foo(int[] a, int i)
 {
     return a[i];
 }
-
-// https://issues.dlang.org/show_bug.cgi?id=17787
-version (D_BetterC)
```
JinShil: Moved this to test/compilable/betterc.d so they are all together.

JinShil:
Should `D_NoExceptions` actually be `D_NoThrowables`?

JinShil:
> Why use the negatives? Why not D_ModuleInfo ?

This is just breaking out `D_BetterC` into its subcomponents. `D_BetterC` implies "No ModuleInfo", "No runtime type info" and "No throwing".  

Also, more often than not, users will be compiling **_with_** support for `ModuleInfo`, `TypeInfo`, and `Throwable`s.  So, in code we will want `D_ModuleInfo`, `D_Throw`, and `D_TypeInfo` to be the default, and `D_NoModuleInfo`, `D_NoThrow`, and `D_NoTypeInfo` to be the exception.  With `D_NoModuleInfo`, `D_NoThrow`, and `D_NoTypeInfo` the frequency with which we will need to litter code with `version` statements will be much less.

Also if we were to use `D_ModuleInfo`, `D_Throw`, and `D_TypeInfo` instead, we'd see this pattern much too often...
```D
version(D_Throw)
{ }
else
{
    // implementation code
}
```
... which is a bit of an eyesore.

To offer a demonstrative example, I want to avoid the the same mistake that was made with `@safe`.  `@safe` should be the default and we should be explicitly attributing code that is `@unsafe` instead.

wilzbach:
> Also, this may be useful for other D compilers that expose compiler flags such as -fno-exceptions, and fno-rtti.

Maybe time for DMD to expose the individual steps of betterC too?

wilzbach:
BTW as I was reading Herb's proposal for C++ on [Zero-overhead deterministic exceptions: Throwing values](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0709r0.pdf) the other day, I find their quote pretty adequate to our -betterC status:

> Divergent error handling has fractured the C++ community into incompatible dialects:
C++ projects often ban even turning on compiler support for exception handling, but this means they are not using Standard C++. Exceptions are required to use central C++ standard language features (e.g., constructors) and the C++ standard library. Yet in [SC++F 2018], over half of C++ developers report that exceptions are banned in part (32%) or all (20%) of their code, which means they are using a divergent language dialect with different idioms (e.g., two-phase construction) and either a nonconforming standard library dialect or none at all. We must make it possible for all C++ projects to at least turn on exception handling support and use the standard language and library.

Geod24:
> D_NoThrow

I think `D_NoExceptions` was better. I assume the change to `D_NoThrow` is seen as more correct, but it's way too close to the `nothrow` keyword. `Exception` is not only the type of the class being thrown, but also the mechanism.

Otherwise, LGTM

WalterBright:
@JinShil Thanks for the explanation, but I want to bring up again the well-known cognitive problem with negation. That is why the dmd source code has `useModuleInfo`, etc. Framing names as positives is much easier to read/understand/etc. In my own code, I've been doing this, and find the results to be a significant improvement. (For example, in the C->D translation I've been replacing the `#if !SPP` conditionals with positive ones.)
```
version (D_NoThrow)
{ }
else
{
    // throwing code
}
```
I doubt many people will process that without a brain hiccup.

Yes, D has `nothrow` and `@nogc`. But I cringe every time I see those.

You suggested that `@safe` should have been the default. I agree. However, that does not help this particular case. Note that `@safe`, `@trusted`, `@system` cover the territory without use of negation, and safe code could be the default without changing these names. `@unsafe` was very deliberately not chosen.

JinShil:
@WalterBright I can't say I agree with your preference for positives, I prefer to think in terms of the most common or preferred case as the default so users should only have to specify the exception case, regardless of whether they are worded as positive or negative.

> In my own code, I've been doing this, and find the results to be a significant improvement. (For example, in the C->D translation I've been replacing the #if !SPP conditionals with positive ones.)

Yes, but keep in mind that the C -> D translation is using -betterC, so in that case `nothrow` is the default.  Most users will not be using -betterC, so `throw` will be the default.

Ultimately, I just want to make using arrays possible and -betterC and similar scenarios, and I need to turn the runtime hooks into templates to do that.  However, the runtime hooks currently throw, so I need to conditionally compile out the `throw` statements and replace them with `assert`s.  I can do that easily enough with either `D_NoThrow` or `D_Throw`, so I'll go along with whatever you prefer.

Please confirm you prefer the following names: `D_ModuleInfo`, `D_TypeInfo`, and `D_Throw`, correct? Do you have a preference for `D_Exceptions` or `D_Throw`?

Geod24:
Minor suggestion: `D_UseXXX` (as done in the compiler) would be much more clear IMHO.

ibuclaw:
Note: Another argument against having negative version conditions is that you can do it yourself within the language using `static if` instead. e.g: https://github.com/dlang/dmd/blob/4b4e15a35bf3f1dae4f43ff433bde20ccc3e7a9d/src/dmd/globals.d#L24

JinShil:
> Minor suggestion: D_UseXXX (as done in the compiler) would be much more clear IMHO.

I went with Walter's preference.  @WalterBright, if you want `D_UseXXX`, speak now.

WalterBright (approved):

jacob-carlborg:
@wilzbach can we check for changelog entries?

PR #8526, opened 2018-07-29
Fix Issue 15373 - Segfault when using typeid on extern(C++) class with virtual functions

JinShil:


jacob-carlborg (approved):

wilzbach (approved):
Definitely better than the segfault, but do we really want to disallow `typeid` on all `extern(C++`) classes?
Seems a bit restrictive for people who want to expose their code to a `C++` codebase.

JinShil:
Please note that I'm operating of [this part of the spec](https://dlang.org/spec/cpp_interface.html#classes):

>Unlike classes and interfaces with D linkage, `extern (C++)` classes and interfaces are not rooted in Object and cannot be used with `typeid`.

However, that doesn't appear to be precisely accurate as using the `extern C++` class type, seems to work OK:

```
import std.stdio;

extern(C++) class C
{ }

void main()
{
    auto Cti = typeid(C);
    writeln(Cti.name);    // OK
    
    C c = new C();
    Cti = typeid(c);  
    writeln(Cti.name);    // seg fault
}
```

So I'm not quite sure what the actual intent is.  @WalterBright, what are your thoughts.

test/fail_compilation/test15373.d
```diff
@@ -0,0 +1,16 @@
...
+*/
+
+// https://issues.dlang.org/show_bug.cgi?id=15373
+
+extern(C++) class C { }
+
+void foo()
+{
+    C c = new C();
+    auto ti = typeid(c);
```
Geod24: Can you also add a test that `typeid(C)` works ?

Geod24:
@JinShil : I think the intent is just to have it working at compile time.

test/fail_compilation/test15373.d
JinShil: Done

PR #8528, opened 2018-07-31
Add `deinitializeDMD` function to deinitialize the front end

jacob-carlborg:
This is a start of being able to restore the global state initialized by `initDMD` to its original state. It only deinitializes state initialized by `initDMD`, state store directly inside functions are not restored.

This is useful to invoke the compiler multiple times within the same application. This allows to write more fine grained tests for the compiler, more of a unit test style compared to the current end to end tests.

### DIFF
--- a/src/dmd/builtin.d
+++ b/src/dmd/builtin.d
@@ -425,6 +425,17 @@ public extern (C++) void builtin_init()
         add_builtin("_D4core5bitop7_popcntFNaNbNiNfmZi", &eval_popcnt);
 }
 
+/**
+ * Deinitializes the global state of the compiler.
+ *
+ * This can be used to restore the state set by `builtin_init` to its original
+ * state.
+ */
+public void builtinDeinitialize()
+{
+    builtins = builtins.init;
+}
+
 /**********************************
  * Determine if function is a builtin one that we can
  * evaluate at compile time.
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -299,6 +299,17 @@ extern (C++) final class Module : Package
         modules = new DsymbolTable();
     }
 
+    /**
+     * Deinitializes the global state of the compiler.
+     *
+     * This can be used to restore the state set by `_init` to its original
+     * state.
+     */
+    static void deinitialize()
+    {
+        modules = modules.init;
+    }
+
     extern (C++) __gshared AggregateDeclaration moduleinfo;
 
     const(char)* arg;           // original argument name
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -640,6 +640,22 @@ extern (C++) abstract class Expression : RootObject
         CTFEExp.showcontext = new CTFEExp(TOK.showCtfeContext);
     }
 
+    /**
+     * Deinitializes the global state of the compiler.
+     *
+     * This can be used to restore the state set by `_init` to its original
+     * state.
+     */
+    static void deinitialize()
+    {
+        CTFEExp.cantexp = CTFEExp.cantexp.init;
+        CTFEExp.voidexp = CTFEExp.voidexp.init;
+        CTFEExp.breakexp = CTFEExp.breakexp.init;
+        CTFEExp.continueexp = CTFEExp.continueexp.init;
+        CTFEExp.gotoexp = CTFEExp.gotoexp.init;
+        CTFEExp.showcontext = CTFEExp.showcontext.init;
+    }
+
     /*********************************
      * Does *not* do a deep copy.
      */
--- a/src/dmd/frontend.d
+++ b/src/dmd/frontend.d
@@ -77,6 +77,35 @@ void initDMD()
 }
 
 /**
+Deinitializes the global variables of the DMD compiler.
+
+This can be used to restore the state set by `initDMD` to its original state.
+Useful if there's a need for multiple sessions of the DMD compiler in the same
+application.
+*/
+void deinitializeDMD()
+{
+    import dmd.builtin : builtinDeinitialize;
+    import dmd.dmodule : Module;
+    import dmd.expression : Expression;
+    import dmd.globals : global;
+    import dmd.id : Id;
+    import dmd.mtype : Type;
+    import dmd.objc : Objc;
+    import dmd.target : target;
+
+    global.deinitialize();
+
+    Type.deinitialize();
+    Id.deinitialize();
+    Module.deinitialize();
+    target.deinitialize();
+    Expression.deinitialize();
+    Objc.deinitialize();
+    builtinDeinitialize();
+}
+
+/**
 Add import path to the `global.path`.
 Params:
     path = import to add
--- a/src/dmd/globals.d
+++ b/src/dmd/globals.d
@@ -405,6 +405,17 @@ struct Global
     }
 
     /**
+     * Deinitializes the global state of the compiler.
+     *
+     * This can be used to restore the state set by `_init` to its original
+     * state.
+     */
+    void deinitialize()
+    {
+        this = this.init;
+    }
+
+    /**
     Returns: the version as the number that would be returned for __VERSION__
     */
     extern(C++) uint versionNumber()
--- a/src/dmd/id.d
+++ b/src/dmd/id.d
@@ -39,6 +39,17 @@ struct Id
     {
         mixin(msgtable.generate(&initializer));
     }
+
+    /**
+     * Deinitializes the global state of the compiler.
+     *
+     * This can be used to restore the state set by `initialize` to its original
+     * state.
+     */
+    void deinitialize()
+    {
+        mixin(msgtable.generate(&deinitializer));
+    }
 }
 
 private:
@@ -486,4 +497,10 @@ string identifier(Msgtable m)
 string initializer(Msgtable m)
 {
     return m.ident ~ ` = Identifier.idPool("` ~ m.name ~ `");`;
+}
+
+// Used to generate the code for each deinitializer.
+string deinitializer(Msgtable m)
+{
+    return m.ident ~ " = Identifier.init;";
 }
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -905,6 +905,17 @@ extern (C++) abstract class Type : RootObject
         thash_t = tsize_t;
     }
 
+    /**
+     * Deinitializes the global state of the compiler.
+     *
+     * This can be used to restore the state set by `_init` to its original
+     * state.
+     */
+    static void deinitialize()
+    {
+        stringtable = stringtable.init;
+    }
+
     final d_uns64 size()
     {
         return size(Loc.initial);
--- a/src/dmd/objc.d
+++ b/src/dmd/objc.d
@@ -188,6 +188,17 @@ extern(C++) abstract class Objc
             _objc = new Unsupported;
     }
 
+    /**
+     * Deinitializes the global state of the compiler.
+     *
+     * This can be used to restore the state set by `_init` to its original
+     * state.
+     */
+    static void deinitialize()
+    {
+        _objc = _objc.init;
+    }
+
     abstract void setObjc(ClassDeclaration cd);
     abstract void setObjc(InterfaceDeclaration);
 
--- a/src/dmd/target.d
+++ b/src/dmd/target.d
@@ -181,6 +181,17 @@ struct Target
     }
 
     /**
+     * Deinitializes the global state of the compiler.
+     *
+     * This can be used to restore the state set by `_init` to its original
+     * state.
+     */
+    void deinitialize()
+    {
+        this = this.init;
+    }
+
+    /**
      * Requested target memory alignment size of the given type.
      * Params:
      *      type = type to inspect
--- a/test/unit/deinitialization.d
+++ b/test/unit/deinitialization.d
@@ -1,0 +1,166 @@
+module deinitialization;
+
+@("global.deinitialize")
+unittest
+{
+    import dmd.globals : global;
+
+    static void assertStructsEqual(T)(const ref T a, const ref T b) @nogc pure nothrow
+    if (is(T == struct))
+    {
+        foreach (i, _; typeof(a.tupleof))
+        {
+            enum name = __traits(identifier, a.tupleof[i]);
+
+            static if (is(typeof(a.tupleof[i]) == const(char*)))
+                assert(a.tupleof[i].fromStringz == b.tupleof[i].fromStringz, name);
+            else
+                assert(a.tupleof[i] == b.tupleof[i], name);
+        }
+    }
+
+    immutable init = global.init;
+    assertStructsEqual(global, init);
+
+    global._init();
+    global.deinitialize();
+
+    assert(global == global.init);
+}
+
+@("Type.deinitialize")
+unittest
+{
+    import dmd.mars : addDefaultVersionIdentifiers, setTarget;
+    import dmd.mtype : Type;
+    import dmd.globals : global;
+
+    assert(Type.stringtable == Type.stringtable.init);
+
+    global._init();
+    setTarget(global.params);
+
+    Type._init();
+    Type.deinitialize();
+
+    global.deinitialize();
+
+    assert(Type.stringtable == Type.stringtable.init);
+}
+
+@("Id.deinitialize")
+unittest
+{
+    import dmd.id : Id;
+
+    static void assertInitialState()
+    {
+        foreach (e ; __traits(allMembers, Id))
+        {
+            static if (!__traits(isStaticFunction, mixin("Id." ~ e)))
+                assert(__traits(getMember, Id, e) is null);
+        }
+    }
+
+    assertInitialState();
+
+    Id.initialize();
+    Id.deinitialize();
+
+    assertInitialState();
+}
+
+@("Module.deinitialize")
+unittest
+{
+    import dmd.dmodule : Module;
+
+    assert(Module.modules is Module.modules.init);
+
+    Module._init();
+    Module.deinitialize();
+
+    assert(Module.modules is Module.modules.init);
+}
+
+@("Target.deinitialize")
+unittest
+{
+    import dmd.globals : Param;
+    import dmd.mars : setTarget;
+    import dmd.target : target, Target;
+
+    static bool isFPTypeProperties(T)()
+    {
+        return is(T == const(typeof(Target.FloatProperties))) ||
+            is (T == const(typeof(Target.DoubleProperties))) ||
+            is (T == const(typeof(Target.RealProperties)));
+    }
+
+    static void assertStructsEqual(T)(const ref T a, const ref T b) @nogc pure nothrow
+    if (is(T == struct))
+    {
+        foreach (i, _; typeof(a.tupleof))
+        {
+            alias Type = typeof(a.tupleof[i]);
+            enum name = __traits(identifier, a.tupleof[i]);
+
+            static if (!isFPTypeProperties!Type)
+                assert(a.tupleof[i] == b.tupleof[i], name);
+        }
+    }
+
+    const init = target.init;
+    assertStructsEqual(target, init);
+
+    Param params;
+    setTarget(params);
+
+    target._init(params);
+    target.deinitialize();
+
+    assertStructsEqual(target, init);
+}
+
+@("Expression.deinitialize")
+unittest
+{
+    import dmd.ctfeexpr : CTFEExp;
+    import dmd.expression : Expression;
+
+    static void assertInitialState()
+    {
+        assert(CTFEExp.cantexp is null);
+        assert(CTFEExp.voidexp is null);
+        assert(CTFEExp.breakexp is null);
+        assert(CTFEExp.continueexp is null);
+        assert(CTFEExp.gotoexp is null);
+        assert(CTFEExp.showcontext is null);
+    }
+
+    assertInitialState();
+
+    Expression._init();
+    Expression.deinitialize();
+
+    assertInitialState();
+}
+
+@("Objc.deinitialize")
+unittest
+{
+    import dmd.objc : Objc, objc;
+
+    assert(objc is null);
+
+    Objc._init();
+    Objc.deinitialize();
+
+    assert(objc is null);
+}
+
+private inout(char)[] fromStringz(inout(char)* cString) @nogc @system pure nothrow
+{
+    import core.stdc.string : strlen;
+    return cString ? cString[0 .. strlen(cString)] : null;
+}
--- a/test/unit/self_test.d
+++ b/test/unit/self_test.d
@@ -10,8 +10,8 @@ import support : afterEach, beforeEach, defaultImportPaths;
 
 @afterEach deinitializeFrontend()
 {
-    // import dmd.frontend : deinitializeDMD;
-    // deinitializeDMD();
+    import dmd.frontend : deinitializeDMD;
+    deinitializeDMD();
 }
 
 @("self test")


ci.sh
```diff
@@ -76,12 +76,21 @@ rebuild() {
...
 test() {
+    run_unit_tests
     test_dub_package
     make -j$N -C ../druntime -f posix.mak MODEL=$MODEL unittest
     make -j$N -C ../phobos -f posix.mak MODEL=$MODEL unittest
     test_dmd
 }
 
+run_unit_tests() {
+    source ~/dlang/*/activate # activate host compiler
```
wilzbach: Any reason why you didn't use the `dub_package` folder?

`test_dub_package` does two important things (that you don't do here)
- doesn't run for GDC (as there's no `dub` installed for GDC)
- runs the dub package tests for the host compiler _and_ currently built compiler

src/dmd/builtin.d
```diff
@@ -475,6 +475,17 @@ public extern (C++) void builtin_init()
         add_builtin("_D4core5bitop7_popcntFNaNbNiNfmZi", &eval_popcnt);
 }
 
+/**
+ * Deinitializes the global state of the compiler.
+ *
+ * This can be used to restore the state set by `builtin_init` to its original
+ * state.
+ */
+public void builtinDeinitialize()
```
wilzbach: - Why not `deinitialize` like in the other modules?
- Why not `static` like in the other modules?

test/unit/source/deinitialization.d
```diff
@@ -0,0 +1,150 @@
+// global.deinitialize
+unittest
```
wilzbach: I thought you wanted to add the `unittest` directly in the DMD source code?

test/unit/source/deinitialization.d
```diff
@@ -0,0 +1,150 @@
...
+{
+    import dmd.globals : global;
+
+    static void assertStructsEqual(T)(const ref T a, const ref T b) @nogc pure nothrow
+        if (is(T == struct))
```
wilzbach: As this is new code:

> Constraints on declarations should have the same indentation level as their declaration:

https://dlang.org/dstyle.html#phobos_declarations

test/unit/source/deinitialization.d
```diff
@@ -0,0 +1,150 @@
...
+    {
+        foreach (i, _; typeof(a.tupleof))
```
wilzbach: FWIW `static foreach` is supported by all host compilers.

test/unit/source/deinitialization.d
```diff
@@ -0,0 +1,150 @@
...
+                assert(a.tupleof[i].fromStringz == b.tupleof[i].fromStringz, name);
+            else
+                assert(a.tupleof[i] == b.tupleof[i], name);
+        }
+    }
+
+    immutable init = global.init;
+    assertStructsEqual(global, init);
+
+    global._init();
```
wilzbach: Partially OT: What do you think about renaming this to `initialize`, s.t. the pattern matches with the other modules?

wilzbach (commented):
I think we should add the unittests directly in the source code. That way they are also automatically run on all platforms that the auto-tester supports and are easier to find and we could even shown some of them in the documentation.

test/unit/source/deinitialization.d
ZombineDev: And also for the same reason we renamed `TypeInfo.init` to `TypeInfo.initializer`.

jacob-carlborg:
> I think we should add the unittests directly in the source code. That way they are also automatically run on all platforms that the auto-tester supports and are easier to find and we could even shown some of them in the documentation

I think there will be too much code in the existing modules. If we would covert all the existing tests to unit tests directly in the existing modules it would add another 200k lines. I think it still will be difficult to test individual functions. I'm thinking we need to start with tests that uses the front end to lex and parse and perhaps run semantic analysis. The tests would then do some assertions on the AST.

ci.sh
jacob-carlborg: > Any reason why you didn't use the dub_package folder?

Yes, the `dub_package` was originally intended to test the Dub package, not the front end, not the compiler. That is, basically make sure that all necessary files are included and the correct flags are used. I didn't have to use Dub for the unit tests, I could just have used a simple shell script that collects all the files, compiles them and run the tests. But I consider Dub to be more portable and it will do that all that automatically.

> doesn't run for GDC (as there's no dub installed for GDC - that's why SemaphoreCI is failing)

That's unfortunate. Can we add it?

> runs the dub package tests for the host compiler and currently built compiler

Do you think that is important? Are we doing that for the other tests (not the Dub package)?

This need to be moved from the `ci.sh` file to the makefile and `run.d`. These tests should always be run, also when running the tests locally. This is intended as an alternative to the existing `compilable`, `fail_compilation` and `runnable`. If we can't use Dub for that, I'll have to come up with something else.

src/dmd/builtin.d
jacob-carlborg: > Why not deinitialize like in the other modules?

To better match the existing `builtin_init`. At the call site having something called just `deinitialize` is not very descriptive in my opinion. It works with the other modules because all those deinitialize functions are scoped in a class, so the class name will always be used at the call site.

> Why not static like in the other modules?

It's a free function.

test/unit/source/deinitialization.d
jacob-carlborg: No (already replied elsewhere).

test/unit/source/deinitialization.d
jacob-carlborg: Yes, I would like to rename all `_init` to `initialize`. That's why I named it `deinitialize` and not `deinit`. This will get rid of the ugly underscore prefix and spell out the full name instead of an abbreviation. A double win, in my opinion 😃.

test/unit/source/deinitialization.d
jacob-carlborg: Ok.

ci.sh
jacob-carlborg: I would need some help to add it to `run.d` and the makefile.

jacob-carlborg:
@wilzbach Dub is not installed on the auto-tester machines. Can we installed that or do I need to find a different approach?

wilzbach:
> Dub is not installed on the auto-tester machines. Can we installed that or do I need to find a different approach?

That's a question for @braddr.

thewilsonator:
@jacob-carlborg would it be possible to have the tests be a GitHub project and test it with build kite, substituting dmd master with the merge of the PR being tested?

wilzbach:
> would it be possible to have the tests be a GitHub project and test it with build kite, substituting dmd master with the merge of the PR being tested?

Would be a pain to upgrade them whenever this still unstable frontend API changes.

wilzbach:
I guess the only way for now is to only run these tests on CircleCi.

jacob-carlborg:
I was planning to replace Dub with a D script. That should work everywhere?

wilzbach:
Yes, but what do you mean by replace? Isn't this partially already done by build.d?

jacob-carlborg:
> Yes, but what do you mean by replace? Isn't this partially already done by build.d?

Currently `test/run.d` and `test/Makefile` has been modified to run Dub which will execute all unit tests I've added in this PR. Instead of using Dub, I was going to replace that with a plain D script, since Dub isn't available on the auto-tester. I don't think `build.d` builds and runs the tests?

jacob-carlborg:
Finally auto-tester and AppVeyor is passing. Not sure why CircleCI and SemaphoreCI don't pass. @wilzbach @thewilsonator?

jacob-carlborg:
@wilzbach @thewilsonator all tests are passing now. Please have a look.

thewilsonator (approved):
Looks fine to me.

test/README.md
```diff
@@ -65,6 +65,65 @@ Note:
 - `AUTO_UPDATE` doesn't work with tests that have multiple `TEST_OUTPUT` segments
 - `AUTO_UPDATE` can be set as an environment variable or as Makefile-like argument assignment
 
+### Running the Unit Tests
+
+The unit tests will automatically run when all tests are run using `./run.d` or
+`make`. To only run the unit tests the `./run.d unit_tests` command can be used.
```
wilzbach: Consistency please.

test/README.md
```diff
@@ -65,6 +65,65 @@ Note:
...
+
+For more finer grain control over the unit tests the `./unit_test_runnner.d`
+command can be used:
+
+To run all unit tests:
+
+```sh
+./unit_test_runnner.d
```
wilzbach: In D it's called `unittest`, not `unit_test`

test/run.d
```diff
@@ -157,51 +215,55 @@ auto predefinedTargets(string[] targets)
...
     foreach (t; targets)
     {
-        auto resultRunTime = resultsDir.buildPath(t ~ ".out").timeLastModified.ifThrown(SysTime.init);
-        if (!force && resultRunTime > testPath(t).timeLastModified &&
+        immutable testName = t.normalizedTestName;
+        auto resultRunTime = resultsDir.buildPath(testName ~ ".out").timeLastModified.ifThrown(SysTime.init);
+        if (!force && resultRunTime > testPath(testName).timeLastModified &&
                 resultRunTime > env["DMD"].timeLastModified.ifThrown(SysTime.init))
-            writefln("%s is already up-to-date", t);
+            writefln("%s is already up-to-date", testName);
```
wilzbach: Why wouldn't you want the short name here on the CLI?

```
compilable/ddoc9369.d is already up-to-date
compilable/test6056c.d is already up-to-date
compilable/test12807.d is already up-to-date
compilable/test15056.d is already up-to-date
```

test/run.d
```diff
@@ -145,7 +181,29 @@ auto predefinedTargets(string[] targets)
...
 
-    Appender!(string[]) newTargets;
+    static Target createUnitTestTarget()
+    {
+        Target target = {
+            args: [getEnvironment["DMD"], "-run", "unit_test_runner.d"]
+        };
+
+        return target;
+    }
```
wilzbach: This won't play nice with rebuilds.

test/run.d
```diff
@@ -145,7 +181,29 @@ auto predefinedTargets(string[] targets)
         return testPath(dir).dirEntries("*{.d,.sh}", SpanMode.shallow).map!(e => e.name);
     }
 
-    Appender!(string[]) newTargets;
+    static Target createUnitTestTarget()
+    {
+        Target target = {
+            args: [getEnvironment["DMD"], "-run", "unit_test_runner.d"]
```
wilzbach: Why not use `hostDMD` here like we do to build the other tools? That way you don't need to rebuild the binary every time when the DMD binary changes.

test/run.d
```diff
@@ -145,7 +181,29 @@ auto predefinedTargets(string[] targets)
...
+        };
+
+        return target;
+    }
+
+    static Target createDefaultTarget(string filename)
```
wilzbach: `default` is a very confusing name.
Maybe sth. like `createTestTarget`?

test/run.d
```diff
@@ -145,7 +181,29 @@ auto predefinedTargets(string[] targets)
...
+    {
+        Target target = {
+            filename: filename,
+            args: [
+                resultsDir.buildPath("d_do_test"),
```
wilzbach: Not that we did this in the past, but if we already change this, let's better make an `enum d_do_test = ...`

test/run.d
```diff
@@ -145,7 +181,29 @@ auto predefinedTargets(string[] targets)
...
+                Target.normalizedTestName(filename)
```
wilzbach: This wasn't required before, why is it required now?

test/run.d
```diff
@@ -157,51 +215,55 @@ auto predefinedTargets(string[] targets)
...
+auto filterTargets(Target[] targets, string[string] env)
 {
     bool error;
     foreach (target; targets)
     {
-        if (!testPath(target).exists)
+        if (!target.exists)
         {
-            writefln("Warning: %s can't be found", target);
+            writefln("Warning: %s can't be found", target.normalizedTestName);
```
wilzbach: What's wrong with the status quo?

```
Warning: compilable/test12009i.d can't be found
```

test/run.d
```diff
@@ -133,6 +130,45 @@ void ensureToolsExists()
...
+    }
+
+    /// Returns: `true` if the test exists
+    bool exists()
+    {
+        // This is assumed to be the `unit_tests` target which always exists
+        if (filename.empty)
+            return true;
+
+        return testPath(normalizedTestName).exists;
```
wilzbach: `testPath` already normalizes.

test/unit_test_runner.d
```diff
@@ -0,0 +1,392 @@
...
+else version(SunOS)
+    enum os ="solaris";
+else
+    static assert(0, "Unrecognized or unsupported OS.");
+
+enum scriptDir = __FILE_FULL_PATH__.dirName.buildNormalizedPath;
+enum projectRoot = scriptDir.buildNormalizedPath("..");
+enum unitTestDir = scriptDir.buildNormalizedPath("unit");
+enum resultsDir = testPath("test_results");
+enum generatedDir = projectRoot.buildPath("generated");
```
wilzbach: A lot of duplication happened here. Why not put common parts into `tools`?

test/Makefile
```diff
@@ -158,7 +158,11 @@ $(RESULTS_DIR)/.created:
 	$(QUIET)if [ ! -d $(RESULTS_DIR)/fail_compilation ]; then mkdir $(RESULTS_DIR)/fail_compilation; fi
 	$(QUIET)touch $(RESULTS_DIR)/.created
 
-run_tests: start_runnable_tests start_compilable_tests start_fail_compilation_tests
+run_tests: unit_tests start_runnable_tests start_compilable_tests start_fail_compilation_tests
+
+unit_tests:
+	@echo "Running unit tests"
+	$(DMD) -conf= $(MODEL_FLAG) $(DEBUG_FLAGS) -run unit_test_runner.d
```
wilzbach: The binary will always be rebuilt :/

test/README.md
```diff
@@ -65,6 +65,65 @@ Note:
 - `AUTO_UPDATE` doesn't work with tests that have multiple `TEST_OUTPUT` segments
 - `AUTO_UPDATE` can be set as an environment variable or as Makefile-like argument assignment
 
+### Running the Unit Tests
+
+The unit tests will automatically run when all tests are run using `./run.d` or
+`make`. To only run the unit tests the `./run.d unit_tests` command can be used.
+
+For more finer grain control over the unit tests the `./unit_test_runnner.d`
+command can be used:
```
wilzbach: I would recommend just using `run.d` as an interface (it can pass all arguments to `unit_test_runner` when an e.g. a `-u` shortcut is used) as this should simplify the exposed tools to the user (which we'll need to maintain) and it will make bootstrapping the binaries easier.

wilzbach (commented):
Looks great! Just a few minor nits
- a lot of (probably necessary) normalization happens here
- consistency `unittest` vs. `unit_test` vs. `unit_tests` (this PR uses all three)

jacob-carlborg:
> Looks great! Just a few minor nits

A few? 😃

jacob-carlborg:
> consistency `unittest` vs. `unit_test` vs. `unit_tests` (this PR uses all three)

I don't use `unittest` anywhere except for when using the D keyword `unittest` (or version identifier).

Here's the deal. In English it's called "unit test", it's also what the D spec calls it in the title [1]. Therefore I use "unit test" or "unit tests" (when it's plural) everywhere it's possible. It's not possible to use "unit test" for an identifier in D source code, therefore the convention is to use camel casing, i.e. `unitTest`. It's possible to use spaces in arguments to applications/scripts but that's not common practice so I use `unit_tests` instead. The makefile also uses underscores in all the target names. As for the name "unit_test_runner", I think it sounds weird with "unit_tests_runner", but I'm not a native English speaker so what do I know.

[1] https://dlang.org/spec/unittest.html

test/run.d
jacob-carlborg: It gives the same result as on the master branch. I've not changed the behavior, as far as I can see.

test/run.d
jacob-carlborg: Any suggestions?

test/run.d
jacob-carlborg: > Why not use `hostDMD` here like we do to build the other tools?

I can try `hostDMD`.

> That way you don't need to rebuild the binary every time when the DMD binary changes.

Why would the binary need to be rebuilt when the DMD binary changes?

test/run.d
jacob-carlborg: Yeah, perhaps. I couldn't come up with a good name.

test/run.d
jacob-carlborg: `enum d_do_test = "d_do_test"` or what do you refer to? I would not consider that a good name for the constant.

test/run.d
jacob-carlborg: Because the targets are not normalized at the beginning as before:

https://github.com/dlang/dmd/pull/8528/files#diff-060ed063c39ccc738bd0e384c097973cL70

test/run.d
jacob-carlborg: There's no change, see: https://github.com/dlang/dmd/pull/8528#discussion_r25398297.7

test/run.d
jacob-carlborg: It's not the same thing. `normalizedTestName` does what this line did: https://github.com/dlang/dmd/pull/8528/files#diff-060ed063c39ccc738bd0e384c097973cL67.

Look at the bottom of this failing CircleCI test to see why it's necessary: https://circleci.com/gh/dlang/dmd/25742.

test/unit_test_runner.d
jacob-carlborg: What is `tools`? Is that something that already exists?

test/README.md
jacob-carlborg: I can give that a try.

test/Makefile
jacob-carlborg: I can try to fix that.

test/README.md
jacob-carlborg: Done.

test/Makefile
```diff
@@ -143,6 +143,9 @@ test_tools=$(RESULTS_DIR)/d_do_test$(EXE) $(RESULTS_DIR)/sanitize_json$(EXE)
 $(RESULTS_DIR)/%.out: % $(RESULTS_DIR)/.created $(test_tools) $(DMD)
 	$(QUIET) $(RESULTS_DIR)/d_do_test $<
 
+$(RESULTS_DIR)/%.out: % $(RESULTS_DIR)/.created $(RESULTS_DIR)/unit_test_runner$(EXE) $(DMD)
+	$(QUIET) $(RESULTS_DIR)/d_do_test $<
```
wilzbach: You duplicated the target here (aka overwrite the target two lines above).

test/README.md
```diff
@@ -65,6 +65,64 @@ Note:
...
+To run all unit tests:
+
+```sh
+./run.d -u
+```
+
+To only run the unit tests in one or more specific files:
+
+```sh
+./run.d -u -- unit/deinitialization.d
```
wilzbach: I don't think we need the `--`. It's an anti-pattern that `dub` introduced, there aren't many precedents in other tools for this.

test/run.d
```diff
@@ -56,6 +63,10 @@ Options:
     // allow overwrites from the environment
     resultsDir = environment.get("RESULTS_DIR", resultsDir);
     hostDMD = environment.get("HOST_DMD", "dmd");
+    unitTestRunnerCommand = resultsDir.buildPath("unit_test_runner");
+
+    if (runUnitTests)
+        return spawnProcess(unitTestRunnerCommand ~ args).wait();
```
wilzbach: How do you ensure that this binary actually exists?
`ensureToolsExists` hasn't been called yet.

test/run.d
```diff
@@ -93,13 +102,14 @@ Options:
...
         foreach (target; taskPool.parallel(targets, 1))
         {
-            auto args = [resultsDir.buildPath("d_do_test"), target];
-            log("run: %-(%s %)", args);
-            ret |= spawnProcess(args, env, Config.none, scriptDir).wait;
+            log("run: %-(%s %)", target.args);
+            ret |= spawnProcess(target.args, env, Config.none, scriptDir).wait;
         }
         if (ret)
             exit(1);
```
wilzbach:
```suggestion
            return 1;
```

test/run.d
```diff
@@ -122,7 +133,13 @@ void ensureToolsExists()
             writefln("%s is already up-to-date", tool);
         else
         {
-            auto command = [hostDMD, "-of"~targetBin, sourceFile];
+            auto command = [
+                hostDMD,
+                toolsDir.buildPath("paths"),
```
wilzbach: Why not use `-i` here? In either case, I would make a comment that these are the imported dependencies of the tools.
Also, you now compiler `paths.o` into the other tools as well (though this probably is just a few KB more).

test/run.d
wilzbach: Well my point was that this value is a constant and could be defined as such statically.

test/tools/paths.d
```diff
@@ -0,0 +1,76 @@
...
+import std.path : buildNormalizedPath, buildPath, dirName, setExtension;
+import std.process : environment;
+
+version (Posix)
+    enum exeExtension = "";
+else version (Windows)
+    enum exeExtension = ".exe";
+
+version(Windows)
+    enum os ="windows";
```
wilzbach: Nit: DStyle requires spacing.

test/tools/paths.d
```diff
@@ -0,0 +1,76 @@
...
+import std.file : exists;
+import std.path : buildNormalizedPath, buildPath, dirName, setExtension;
+import std.process : environment;
+
+version (Posix)
+    enum exeExtension = "";
+else version (Windows)
+    enum exeExtension = ".exe";
+
+version(Windows)
```
wilzbach: Since `run.d` has been written it was agreed to use a space after `version ` too :man_shrugging: 
At least be consistent in this file.

test/run.d
```diff
@@ -250,47 +329,35 @@ string[string] getEnvironment()
     string[string] env;
 
     env["RESULTS_DIR"] = resultsDir;
-    auto os = env.getDefault("OS", detectOS);
-    auto build = env.getDefault("BUILD", "release");
+    env["OS"] = os;
+    env["MODEL"] = model;
+    env["BUILD"] = build;
```
wilzbach: Hmm, I would rename these three to make clear to the reader where they come from.
How about `targetOS` or `buildOS`? Or maybe simply doing all UPPERCASE would work?
I know that's a bit against the style guide, but this comes from the Makefiles where they are uppercased.

test/tools/paths.d
```diff
@@ -0,0 +1,76 @@
...
+
+enum projectRootDir = __FILE_FULL_PATH__.dirName.buildNormalizedPath("..", "..");
+enum generatedDir = projectRootDir.buildPath("generated");
+enum resultsDir = testPath("test_results");
+
+enum dmdFilename = "dmd".setExtension(exeExtension);
+
+alias testPath = path => projectRootDir.buildPath("test", path);
+
+string build()
```
wilzbach: `buildType` or maybe simply `BUILD` (like in Make)?

test/tools/paths.d
```diff
@@ -0,0 +1,76 @@
...
+{
+    shared string build;
+    return build = build ? build : environment.get("BUILD", "release");
+}
+
+string buildOutputPath()
+{
+    shared string buildOutputPath;
```
wilzbach: This string is never modified. Did you forget the assignment?

test/run.d
wilzbach: Also to avoid the need for `--` e.g save `args` before doing the `args.popFront` on L60.

test/README.md
ZombineDev: I've seen this style in in several tools, e.g. git - `git log -p master -- posix.mak`. It's mostly useful when the CLI has options (named parameters), non-file arguments and file arguments.

test/Makefile
jacob-carlborg: Is this target not needed? Seems to work without it.

test/README.md
jacob-carlborg: It's common practice among unix tools that invoke a subcommand to use `--` to separate the arguments for the subcommand from the main command. In this case there will be a conflict with the `-f` flag. `run.d` uses `-f` to mean "force" and `unit_test_runner.d` to mean "filter". I can work around that if you insist.

test/run.d
jacob-carlborg: > How do you ensure that this binary actually exists? `ensureToolsExists` hasn't been called yet.

That's a good point. I'll fix that.

> Also to avoid the need for `--` e.g save `args` before doing the `args.popFront` on L60.

That won't work because of the conflict with the `-f` flag.

test/run.d
jacob-carlborg: > Why not use `-i` here?

Does GDC support that flag?

test/tools/paths.d
jacob-carlborg: Hmm, not sure what happened there. I'll fix that.

test/tools/paths.d
jacob-carlborg: Sure, that's what I prefer too.

test/run.d
jacob-carlborg: Is it relevant where they come from? It was called `os` before:

https://github.com/dlang/dmd/blob/a60e9e25f075af0179f521e8e6864d892996a713/test/run.d#L253

test/tools/paths.d
jacob-carlborg: It was called `build` before: https://github.com/dlang/dmd/blob/a60e9e25f075af0179f521e8e6864d892996a713/test/run.d#L254

test/tools/paths.d
jacob-carlborg: Yes, I'll fix that.

test/Makefile
jacob-carlborg: Fixed (I think).

test/run.d
jacob-carlborg: Fixed (ensuring the tool exists).

test/run.d
jacob-carlborg: Fixed.

test/run.d
jacob-carlborg: Fixed (without using the `-i` flag).

test/tools/paths.d
jacob-carlborg: Fixed.

test/tools/paths.d
jacob-carlborg: Fixed.

test/tools/paths.d
jacob-carlborg: Fixed.

jacob-carlborg:
> Well my point was that this value is a constant and could be defined as such statically.

Fixed (not sure why I can't reply inline to https://github.com/dlang/dmd/pull/8528#discussion_r254264348).

test/tools/unit_test_runner.d
```diff
@@ -0,0 +1,326 @@
...
+        .writeRunnerFile(runnerPath, unitTestFilter);
+
+    enum cmdfilePath = resultsDir.buildPath("cmdfile");
+    enum outputPath = resultsDir.buildPath("runner").setExtension(exeExtension);
+    writeCmdfile(cmdfilePath, runnerPath, outputPath, testFiles);
+
+    runConfig();
+    buildStrtold();
+    execute(dmdPath, "@" ~ cmdfilePath);
+    execute(outputPath);
```
wilzbach: So I presume that you save the commands to a file because you want to be able to invoke DMD outside of this tool, but why not add `-run` to the command flags above?

test/tools/unit_test_runner.d
```diff
@@ -0,0 +1,326 @@
...
+        configPath,
+        outputPath,
+        versionPath
+    ];
+
+    version (Posix)
+        cmd ~= "/etc";
+
+    execute(cmd);
+}
```
wilzbach: Why do we need this? This file will be generated as part of either building DMD or building DMD as a library with Dub.

test/run.d
wilzbach: It's not that important as we currently don't have many tests, but in the future we probably have to do:
- write a file after every invocation
- lastEdit  = MAX(timestamp of src/*.d)
- compare lastEdit with the timestamp of the written file

(that's how the current testsuite does it, except that it checks the timestamp of the DMD binary which wouldn't be helpful here.)

wilzbach (approved):
LGTM! (just a few final questions out of interest).

wilzbach:
BTW as this does have an impact on the testsuite and tooling based on it it would be amazing if @CyberShadow or @MartinNowak could have a quick look over this too.

test/tools/unit_test_runner.d
jacob-carlborg: > So I presume that you save the commands to a file because you want to be able to invoke DMD outside of this tool

The assumption is that there might be a lot of files added and I think Windows has a limitation on how many arguments can be passed to an application.

> but why not add `-run` to the command flags above?

It didn't work. Seems to be some kind of bug. I didn't mange to create a small test case.

CyberShadow:
My initial reaction: that's a lot of kinds of changes in a single commit. I would suggest splitting the changes into separate commits (or PRs, as often argued by Walter), with one kind of change in one commit.

test/tools/unit_test_runner.d
jacob-carlborg: This is not using Dub anymore. I failed to get it to install. But it might not be required since the assumption is that DMD is built anyway. I'll see if I can remove it.

test/run.d
jacob-carlborg: That could work, but it would complicate the unit test runner even more. It would also need to check `unit/*.d`.

jacob-carlborg:
> that's a lot of kinds of changes in a single commit.

There are two kinds of changes: a unit test runner is added and a new function to deinitialize the frontend. What complicates things are that we have five CI systems and two different ways to invoke the tests.

CyberShadow:
I haven't really been following this side of things.

As for the components I'm responsible for, Digger just runs `make` (with a properly set up environment) in the `dmd/test` directory. As long as that keeps working, nothing there should break.

> There are two kinds of changes: a unit test runner is added

Can this be done in a separate commit or PR?

jacob-carlborg:
> Can this be done in a separate commit or PR?

Yes, anything is possible.

jacob-carlborg:
Created a new PR with only the unit test runner: https://github.com/dlang/dmd/pull/9333.

jacob-carlborg:
This PR now contains two commits, one which adds the unit test runner and one which adds the deinitialization function. When https://github.com/dlang/dmd/pull/9333 is merged I'll rebase this PR and the unit test runner commit will be gone from this PR.

jacob-carlborg:
Blocked by: https://github.com/dlang/dmd/pull/9333.

jacob-carlborg:
Is blocking: https://github.com/dlang/dmd/pull/9350.

PR #8521, opened 2018-07-27
-betterC: Add support for types when comparing arrays

JinShil:
The actual implementation change is in the druntime PR at https://github.com/dlang/druntime/pull/2255.  This just adds the tests.  The druntime PR will need to be merged first before this will pass the CI.

PR #8529, opened 2018-07-31
[RFC] Move object.d to dmd

wilzbach:
Just as an experiment. 

> If anything makes sense, it would be to remove the compiler-dependencies out of druntime into a smaller runtime library that is included in the dmd project. Most of druntime is dependencies for the platform, not the compiler.

https://forum.dlang.org/post/pjqepc$2sfv$1@digitalmars.com

Required for https://github.com/dlang/druntime/pull/2262


JinShil:
Sorry, but I'm not following how this is going to work.  object.d is automatically imported by every D module.  If it's in DMD, where will D code import from?

wilzbach:
> Sorry, but I'm not following how this is going to work.   object.d is automatically imported by every D module.  If it's in DMD,  where will D code import object.d from?

Nothing changes. For example, here's the development config for dmd:

```
[Environment64]
DFLAGS=-I%@P%/../../../../../druntime/import -I%@P%/../../../../../phobos -L-L%@P%/../../../../../phobos/generated/linux/release/64 -L--export-dynamic -fPIC
```

`druntime/import` is already a copy of druntime's src. However, https://github.com/dlang/druntime/pull/2262 now does this little trick:

```diff
diff --git a/posix.mak b/posix.mak
index e5bf862c..cfcd542f 100644
--- a/posix.mak
+++ b/posix.mak
@@ -188,7 +188,7 @@ $(IMPDIR)/core/sync/%.di : src/core/sync/%.d $(DMD)
 
 copy: $(COPY)
 
-$(IMPDIR)/object.d : src/object.d
+$(IMPDIR)/object.d : $(DMD_DIR)/druntime/object.d
 	@mkdir -p $(dir $@)
 	@rm -f $(IMPDIR)/object.di
 	cp $< $@
```

Note: Druntime's already depends on the DMD repository (e.g. it depends on the built compiler to built itself and there's even a `include` to DMD's Makefile in the same file.)

JinShil:
Then, from the user's perspective, it's still part of druntime, as druntime's build and deployment process copies object.d from DMD to druntime.  I suppose it all a wash in the end, but it makes this PR seem somewhat superficial.

wilzbach:
>  it makes this PR seem somewhat superficial.

You can now edit the compiler and `object.d` in _one_ PR and don't need to make a druntime, dmd, druntime PR series.
For example, the recent PR of yours from this week (https://github.com/dlang/dmd/pull/8521) could have been submitted as a single PR and would have been fully checked by all CIs.

> with the platform vs. compiler dependencies and all that, but I don't think it's quite there yet.

This seems to be the pragmatic compromise that can be made between keeping druntime as a separate repository and reducing the cyclic dependencies.
What's "there" for you?

> Maybe as a first step, but I'd like to have a clear idea of the end goal before taking incremental steps towards it.

Ehm, that's more or less the end goal. There can be a point made to move `rt` too, but some parts of `rt` are platform-depend and also depend on `core`. On the other hand, `rt.typeinfo`, and a few other depend much more on the compiler than on druntime. So it's a possibility that's still discussed on the NG.

JinShil:
> You can now edit the compiler and object.d in one PR and don't need to make a druntime, dmd, druntime PR series.

Yes, I understand that, but that can also be solved in other ways.  We need a more compelling reason to restructure druntime and dmd besides just testing dependent PRs.  I think there are compelling reasons to do it and @schveiguy hit a nerve with his observations about compiler and platform dependencies.  I think there is a structure that makes more sense than what we have now, but I'd like it to be done for reasons other than the limitations of GitHub and our CIs.

PR #8531, opened 2018-08-02
Lower _d_arraycast to object.__ArrayCast

JinShil:
Requires https://github.com/dlang/druntime/pull/2264 to be merged first

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -6272,6 +6272,16 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
             if (exp.to == Type.terror)
                 return setError();
 
+            // Convert an expression like `cast(uint[])[]` or `cast(string[])[null]`
+            // to a typed array literals
+            if (exp.e1.op == TOK.arrayLiteral)
+            {
+                auto al = cast(ArrayLiteralExp)exp.e1.expressionSemantic(sc);
+                al.type = exp.to;
+                result = al;
+                return;
+            }
+
             if (!exp.to.hasPointers())
                 exp.setNoderefOperand();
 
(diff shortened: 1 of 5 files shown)


src/dmd/expressionsem.d
```diff
@@ -6383,6 +6397,42 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+            {
+                // Array element sizes do not match, so we must adjust the dimensions
+
+                if ((fromSize % toSize) != 0)
+                {
+                    // Runtime check needed in case arrays don't line up
+
+                    // lower to `object.__ArrayCast!(TFrom, TTo)(from)`
+
+                    auto id = Identifier.idPool("__ArrayCast");
```
JinShil: How can I fully qualify this to `object.__ArrayCast`?

test/runnable/betterc_array_cast.d
```diff
@@ -0,0 +1,34 @@
...
+    // We can't actually create dynamic arrays in idiomatic D when
+    // compiling with -betterC, so we do it manually.
+    auto b_length = cast(size_t*)&b;
+    auto b_ptr = cast(void*)(b_length + 1);
+    *b_length = int.sizeof * 3;
+    b_ptr = malloc(*b_length);
+
+    // size mismatch, should result in an assertion failure
+    l = cast(long[])b;
+}
```
Geod24: Missing newline

JinShil: Fixed.

src/dmd/expressionsem.d
jacob-carlborg: Look at how the comparison lowering is done to `__cmp`.

https://github.com/dlang/dmd/blob/03e7693150c453055b098768e59ebd31cb3ca52e/src/dmd/expressionsem.d#L9794-L9798

JinShil: But I need an `Identifer`, not an `Expression`, for the `TemplateInstance`, later.  How do I create a fully-qualified `Identifier`?

jacob-carlborg: > How do I create a fully-qualified Identifier?

I don't think that is possible. To get the fully qualified name inside the compiler one needs to walk up the parent chain and build up the name that way. In your case, I think you need to use the overload of the `TemplateInstance` constructor that takes a `TemplateDeclaration`. Perhaps you can store the `TemplateDeclaration` when it's analyzed in the same way as the `Object` class declaration (and others) is stored:

https://github.com/dlang/dmd/blob/03e7693150c453055b098768e59ebd31cb3ca52e/src/dmd/dclass.d#L367-L372

PR #8532, opened 2018-08-02
 Add predefined version identifier `D_BoundsChecks`

Geod24:
```
One stylistic rule that is strictly enforced in D is that
version identifier and CLI switch should be positive.
`D_NoBoundsChecks` stands out as an exception in this regard.
This PR just adds `D_BoundsChecks` so that code have at least
one release to transition to it, before `D_NoBoundsChecks` is deprecated.
```

### DIFF
--- a/changelog/d_boundscheck.dd
+++ b/changelog/d_boundscheck.dd
@@ -1,0 +1,29 @@
+Add version identifier `D_BoundsChecks`
+
+One stylistic rule of D is that version identifier should be positive.
+`D_NoBoundsChecks` stands out as an exception in this regard.
+This version adds the opposite `D_BoundsChecks` so that code have at least
+one release to transition to it, before `D_NoBoundsChecks` is deprecated.
+
+---
+/// Before
+void func (ubyte[] arr, size_t idx)
+{
+  version (D_NoBoundsCheck) {}
+  else
+  {
+    assert(idx < arr.length);
+  }
+  // ...
+}
+
+/// After
+void func (ubyte[] arr, size_t idx)
+{
+  version (D_BoundsCheck)
+  {
+    assert(idx < arr.length);
+  }
+  // ...
+}
+---
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1297,6 +1297,8 @@ void addDefaultVersionIdentifiers(const ref Param params)
         VersionCondition.addPredefinedGlobalIdent("assert");
     if (params.useArrayBounds == CHECKENABLE.off)
         VersionCondition.addPredefinedGlobalIdent("D_NoBoundsChecks");
+    else // Introduced in 2.083.0 - Deprecate `D_NoBoundsChecks` later
+        VersionCondition.addPredefinedGlobalIdent("D_BoundsChecks");
     if (params.betterC)
     {
         VersionCondition.addPredefinedGlobalIdent("D_BetterC");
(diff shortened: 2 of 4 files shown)


changelog/d_boundscheck.dd
```diff
@@ -0,0 +1,29 @@
+Add version identifier `D_BoundsChecks`
+
+One stylistic rule of D is that version identifier should be positive.
+`D_NoBoundsChecks` stands out as an exception in this regard.
+This version adds the opposite `D_BoundsChecks` so that code have at least
```
JinShil: I think we might be subject to the deprecation policy being proposed at https://github.com/dlang/DIPs/blob/master/DIPs/DIP1013.md

src/dmd/mars.d
```diff
@@ -1387,6 +1387,8 @@ void addDefaultVersionIdentifiers()
         VersionCondition.addPredefinedGlobalIdent("assert");
     if (global.params.useArrayBounds == CHECKENABLE.off)
         VersionCondition.addPredefinedGlobalIdent("D_NoBoundsChecks");
+    else // Introduced in 2.082.0 - Deprecate `D_NoBoundsChecks` later
```
JinShil: Use the deprecation comment syntax proposed in https://github.com/dlang/DIPs/blob/master/DIPs/DIP1013.md for easy grepping, and remain consistence with other deprecations in the code base.

JinShil:
Fine with me after we have a spec update and we get deprecation policy sorted out.

WalterBright (approved):
It's not "strictly" enforced, just a strong bias that direction. Also, it should be removed from the documentation (and this added).

Geod24:
@JinShil : Thanks for the link, didn't knew about it.
I wanted to introduce the version identifier for at least one release before deprecating the other identifier, so that transition is smoother. What do you think ?

Geod24:
https://github.com/dlang/dlang.org/pull/2438

JinShil (approved):

Geod24:
Also found https://issues.dlang.org/show_bug.cgi?id=19137 out

changelog/d_boundscheck.dd
wilzbach: I think it should be either code bases or projects (code can't transition itself, can it?)

WalterBright:
> Also found https://issues.dlang.org/show_bug.cgi?id=19137 out

Good catch. I withdraw my approval until this is sorted out.

WalterBright (changes requested):
Need to resolve bug 19137

Geod24:
@WalterBright : Could you give your opinion on said bug ? My original thought was that we could apply `D_BoundsChecks` when the scope is `@safe` but I'm afraid it will introduce a bad precedent. E.g. what about a template specialization which is `@safe` and has invalid code in its `D_BoundsChecks` versioned code ?

On the other hand @JinShil proposed to introduce `D_SafeBoundsChecks`, which is a much simpler approach but makes it more difficult for the user. One has to know which one to use, and it might not be possible (e.g. when templates are used).

Do you have an alternative idea, or do you prefer one of those ?

ghost (changes requested):
This not worth starting a deprecation. New version identifiers however should follow the rule of being "positive". Also even if unlikey, the new reserved version could exist somewhere in some code, so a deprecation would also be required, making this (even if simple) change a long process.

I suggest @Geod24 to close.

Geod24:
Yes the benefit of this is very low, hence why I haven't pursued it. However, one note: version identifiers starting with `D_` are reserved so there wouldn't be a need for a deprecation.

NilsLankila:
ok, note taked.

PR #8534, opened 2018-08-04
Don't pass null pointer as format parameter to errorSupplemental

ibuclaw:
Null pointer causes an ICE when used with gcc's diagnostic routines.

### DIFF
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -624,7 +624,8 @@ extern (C++) final class Module : Package
             {
                 .error(loc, "cannot find source code for runtime library file 'object.d'");
                 errorSupplemental(loc, "dmd might not be correctly installed. Run 'dmd -man' for installation instructions.");
-                errorSupplemental(loc, "config file: %s", FileName.canonicalName(global.inifilename));
+                const dmdConfFile = FileName.canonicalName(global.inifilename);
+                errorSupplemental(loc, "config file: %s", dmdConfFile ? dmdConfFile : "not found".ptr);
             }
             else
             {
--- a/test/fail_compilation/no_object.d
+++ b/test/fail_compilation/no_object.d
@@ -5,7 +5,7 @@ TEST_OUTPUT:
 ---
 Error: cannot find source code for runtime library file 'object.d'
        dmd might not be correctly installed. Run 'dmd -man' for installation instructions.
-       config file: (null)
+       config file: not found
 import path[0] = fail_compilation
 ---
 */


jacob-carlborg:
Testable?

ibuclaw:
> Testable?

Just build gdc and run `touch ice.d && ./gcc/gdc -Bgcc ice.d` and you'll get a lovely backtrace.

ibuclaw:
I also have the C standard on my side. :-)
```
7.1.4 Use of library functions
---
If an argument to a function has an invalid value (such as a value outside the domain of the function, or a pointer outside the address space of the program, or  a  null  pointer,  or  a  pointer  to  non-modifiable  storage  when  the  corresponding parameter  is  not  const-qualified)  or  a  type  (after  promotion)  not  expected  by  a  function with variable number of arguments, the behavior is undefined.
```

ibuclaw:
@jacob-carlborg - Ah, I didn't test with dmd itself, checking the result of `FileName::canonicalName` instead.

jacob-carlborg (approved):

jacob-carlborg:
Alternatively you could provide something more descriptive instead of not printing anything.

src/dmd/dmodule.d
```diff
@@ -624,7 +624,9 @@ extern (C++) final class Module : Package
             {
                 .error(loc, "cannot find source code for runtime library file 'object.d'");
                 errorSupplemental(loc, "dmd might not be correctly installed. Run 'dmd -man' for installation instructions.");
-                errorSupplemental(loc, "config file: %s", FileName.canonicalName(global.inifilename));
+                const dmdConfFile = FileName.canonicalName(global.inifilename);
+                if (dmdConfFile)
+                    errorSupplemental(loc, "config file: %s", dmdConfFile);
```
wilzbach: Why not keep the existing behaviour with sth. like: `dmdConfFile ? dmdConfFile : "(null)".ptr`?

ibuclaw: I'm not fussed either way, just so long as it isn't null.

wilzbach (approved):

src/dmd/dmodule.d
```diff
@@ -624,7 +624,8 @@ extern (C++) final class Module : Package
             {
                 .error(loc, "cannot find source code for runtime library file 'object.d'");
                 errorSupplemental(loc, "dmd might not be correctly installed. Run 'dmd -man' for installation instructions.");
-                errorSupplemental(loc, "config file: %s", FileName.canonicalName(global.inifilename));
+                const dmdConfFile = FileName.canonicalName(global.inifilename);
+                errorSupplemental(loc, "config file: %s", dmdConfFile ? dmdConfFile : "(null)".ptr);
```
WalterBright: The message would be `config file: (null)` which doesn't seem very helpful. How about `configuration file 'dmd.conf' not found` ?

WalterBright (changes requested):
Please make message more informative.

JinShil (approved):

PR #8537, opened 2018-08-05
fix Issue 19140 - AssertError@dmd/ctfeexpr.d(229): Assertion failure

ibuclaw:
https://issues.dlang.org/show_bug.cgi?id=19140

When compiling with `-verrors=spec`
```
(spec:1) test/compilable/interpret3.d(7751): Error: f19140 cannot be interpreted at compile time, because it has no available source code
(spec:1) Error: cannot implicitly convert expression <error> of type void to real
```

### DIFF
--- a/src/dmd/ctfeexpr.d
+++ b/src/dmd/ctfeexpr.d
@@ -219,6 +219,8 @@ extern (C++) final class CTFEExp : Expression
             return "<cant>";
         case TOK.voidExpression:
             return "<void>";
+        case TOK.showCtfeContext:
+            return "<error>";
         case TOK.break_:
             return "<break>";
         case TOK.continue_:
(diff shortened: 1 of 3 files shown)


RazvanN7 (approved):

RazvanN7:
Please re-target stable.

ibuclaw:
No, because this is a regression only in master.

ibuclaw:
Also, I'll be rebasing gdc against master on a weekly basis, so all these implementation regressions will be rooted out long before release. ;-)

Geod24 (approved):

CyberShadow:
> @CyberShadow any idea why DAutoTest is now permanently failing?

It doesn't look like a permanent failure judging from the branch history:
http://dtest.dlang.io/history/master

Perhaps bad dependencies in the makefile causing an unstable build order?

PR #8538, opened 2018-08-05
Implement GccAsmStatement for gcc-style asm statements

ibuclaw:
The AST itself is only used by GDC.

Parser and semantic analysis has been merged into a single function, and put in its own module as per https://github.com/dlang/dmd/pull/7562#issuecomment-360684196.

Lots more documentation could be added to the new functions that implement gcc asm.

Closes #7562.

### DIFF
--- a/src/dmd/iasmgcc.d
+++ b/src/dmd/iasmgcc.d
@@ -361,3 +361,89 @@ public Statement gccAsmSemantic(GccAsmStatement s, Scope *sc)
 
     return s;
 }
+
+unittest
+{
+    uint errors = global.startGagging();
+
+    // Immitates asmSemantic if version = IN_GCC.
+    static int semanticAsm(Token* tokens)
+    {
+        scope gas = new GccAsmStatement(Loc.initial, tokens);
+        scope p = new Parser!ASTCodegen(null, ";", false);
+        p.token = *tokens;
+        p.parseGccAsm(gas);
+        return p.errors;
+    }
+
+    // Immitates parseStatement for asm statements.
+    static void parseAsm(string input, bool expectError)
+    {
+        // Generate tokens from input test.
+        scope p = new Parser!ASTCodegen(null, input, false);
+        p.nextToken();
+
+        Token* toklist = null;
+        Token** ptoklist = &toklist;
+        p.check(TOK.asm_);
+        p.check(TOK.leftCurly);
+        while (1)
+        {
+            if (p.token.value == TOK.rightCurly || p.token.value == TOK.endOfFile)
+                break;
+            *ptoklist = Token.alloc();
+            memcpy(*ptoklist, &p.token, Token.sizeof);
+            ptoklist = &(*ptoklist).next;
+            *ptoklist = null;
+            p.nextToken();
+        }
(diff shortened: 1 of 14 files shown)


ibuclaw:
As title update suggests, renamed class to GccAsmStatement to avoid ambiguity.

JinShil (approved):

src/dmd/iasmgcc.d
```diff
@@ -0,0 +1,362 @@
...
+        s.insn = (cast(CompileExp)s.insn).e1;
+    s.insn = semanticString(sc, s.insn, "asm instruction template");
+
+    if (s.labels && s.outputargs)
+        s.error("extended asm statements with labels cannot have output constraints");
+
+    // Analyse all input and output operands.
+    if (s.args)
+    {
+        for (size_t i = 0; i < s.args.dim; i++)
```
wilzbach: Could be a foreach

wilzbach (approved):
Reviewing this commit by commit was really straight-forward. Thank you!

src/dmd/iasmgcc.d
ibuclaw: If you mean `foreach (i; 0 .. s.args.dim)`, then ok.

ibuclaw:
Rebased; tweaked documented grammar so that there is no distinction between input and output operands; use foreach in a few places.

ibuclaw:
@WalterBright - This is good to review I guess.

jacob-carlborg:
Can tests be added?

ibuclaw:
@jacob-carlborg unittests mimicking gdc? I guess so...

jacob-carlborg:
> jacob-carlborg unittests mimicking gdc? I guess so..

I had a few ideas:

1. Use more of an integration/unit test style approach instead of the current end-to-end
2. Tests that shows what happens if DMD is trying to use this new statement
3. Can we have tests that are only executed by GDC? We are running GDC using SemaphoreCI, as far as I know

Unless somebody has a better idea.

ibuclaw:
>  Use more of an integration/unit test style approach instead of the current end-to-end

See latest commit.

> Tests that shows what happens if DMD is trying to use this new statement

DMD will always treat it like a dmd asm statement, so will error in same way as before.

> Can we have tests that are only executed by GDC? We are running GDC using SemaphoreCI, as far as I know

That doesn't really help at all if we make changes to it, as for proper testing you should build GDC from source.  You should probably have some faith in me that I vet my own compiler. :-)

jacob-carlborg:
> That doesn't really help at all if we make changes to it, as for proper testing you should build GDC from source. You should probably have some faith in me that I vet my own compiler

Yes, but you're making changes to a project without adding tests to the project. But now I see that you've added tests :+1:.

PetarKirov (approved):
Nice work on the code and git commit separation!

wilzbach:
SemaphoreCI failure is due to it checking out the wrong branch. This should be fixed by https://github.com/dlang/dmd/pull/8560 (once it's merged into `stable` and merged to `master`).

ibuclaw:
:-O

PR #8540, opened 2018-08-06
Remove contracts from StaticForEach

ibuclaw:
There's no benefit to having them there over just having the assert as part of the body.

The use of `body` vs. `do` also prevents the frontend from being built by older compilers in the 2.07x series, such as what's used by the pre-built gdc-7 packages in Debian.


jacob-carlborg:
> There's no benefit to having them there over just having the assert as part of the body.

No, not how it currently works. Ideally the contracts should be executed by the caller, then it makes a difference.

Geod24 (approved):

PR #8542, opened 2018-08-07
Fix issue 19058 - __traits(getUnitTests) works again with separate co…

atilaneves:
I tried creating the PR to `stable` instead of `master` but for reasons I don't understand a whole lot of commits came along with it. And yes, I'd rebased to `upstream/stable`.

### DIFF
--- a/src/dmd/identifier.d
+++ b/src/dmd/identifier.d
@@ -23,6 +23,7 @@ import dmd.root.stringtable;
 import dmd.tokens;
 import dmd.utf;
 
+
 /***********************************************************
  */
 extern (C++) final class Identifier : RootObject
@@ -124,6 +125,18 @@ nothrow:
 
     extern (C++) __gshared StringTable stringtable;
 
+    /**
+       A secondary string table is used to guarantee that we generate unique
+       identifiers per module. See generateIdWithLoc and issues
+       https://issues.dlang.org/show_bug.cgi?id=16995
+       https://issues.dlang.org/show_bug.cgi?id=18097
+       https://issues.dlang.org/show_bug.cgi?id=18111
+       https://issues.dlang.org/show_bug.cgi?id=18880
+       https://issues.dlang.org/show_bug.cgi?id=18868
+       https://issues.dlang.org/show_bug.cgi?id=19058.
+     */
+    private extern (C++) __gshared StringTable fullPathStringTable;
+
     static Identifier generateId(const(char)* prefix)
     {
         __gshared size_t i;
@@ -153,23 +166,75 @@ nothrow:
      */
     extern (D) static Identifier generateIdWithLoc(string prefix, const ref Loc loc)
     {
-        OutBuffer buf;
-        buf.writestring(prefix);
-        buf.writestring("_L");
-        buf.print(loc.linnum);
-        buf.writestring("_C");
-        buf.print(loc.charnum);
-        auto basesize = buf.peekSlice().length;
-        uint counter = 1;
-        while (stringtable.lookup(buf.peekSlice().ptr, buf.peekSlice().length) !is null)
+        import dmd.root.filename: absPathThen;
+
+        // see below for why we use absPathThen
+        return loc.filename.absPathThen!((absPath)
         {
-            // Strip the extra suffix
-            buf.setsize(basesize);
-            // Add new suffix with increased counter
-            buf.writestring("_");
-            buf.print(counter++);
-        }
-        return idPool(buf.peekSlice());
+
+            // this block generates the "regular" identifier, i.e. if there are no collisions
+            OutBuffer idBuf;
+            idBuf.writestring(prefix);
+            idBuf.writestring("_L");
+            idBuf.print(loc.linnum);
+            idBuf.writestring("_C");
+            idBuf.print(loc.charnum);
+
+            // This block generates an identifier that is prefixed by the absolute path of the file
+            // being compiled. The reason this is necessary is that we want unique identifiers per
+            // module, but the identifiers are generated before the module information is available.
+            // To guarantee that each generated identifier is unique without modules, we make them
+            // unique to each absolute file path. This also makes it consistent even if the files
+            // are compiled separately. See issues:
+            // https://issues.dlang.org/show_bug.cgi?id=16995
+            // https://issues.dlang.org/show_bug.cgi?id=18097
+            // https://issues.dlang.org/show_bug.cgi?id=18111
+            // https://issues.dlang.org/show_bug.cgi?id=18880
+            // https://issues.dlang.org/show_bug.cgi?id=18868
+            // https://issues.dlang.org/show_bug.cgi?id=19058.
+            OutBuffer fullPathIdBuf;
+
+            if (absPath)
+            {
+                // replace characters that demangle can't handle
+                for (auto ptr = absPath; *ptr != '\0'; ++ptr)
+                {
+                    // see dmd.dmangle.isValidMangling
+                    // Unfortunately importing it leads to either build failures or cyclic dependencies
+                    // between modules.
+                    if (*ptr == '/' || *ptr == '\\' || *ptr == '.' || *ptr == '?' || *ptr == ':')
+                        *ptr = '_';
+                }
+
+                fullPathIdBuf.writestring(absPath);
+                fullPathIdBuf.writestring("_");
+            }
+
+            fullPathIdBuf.writestring(idBuf.peekSlice());
+            const fullPathIdLength = fullPathIdBuf.peekSlice().length;
+            uint counter = 1;
+
+            // loop until we can't find the absolute path ~ identifier, adding a counter suffix each time
+            while (fullPathStringTable.lookup(fullPathIdBuf.peekSlice()) !is null)
+            {
+                // Strip the counter suffix if any
+                fullPathIdBuf.setsize(fullPathIdLength);
+                // Add new counter suffix
+                fullPathIdBuf.writestring("_");
+                fullPathIdBuf.print(counter++);
+            }
+
+            // `idStartIndex` is the start of the "true" identifier. We don't actually use the absolute
+            // file path in the generated identifier since the module system makes sure that the fully
+            // qualified name is unique.
+            const idStartIndex = fullPathIdLength - idBuf.peekSlice().length;
+
+            // Remember the full path identifier to avoid possible future collisions
+            fullPathStringTable.insert(fullPathIdBuf.peekSlice(),
+                                       null);
+
+            return idPool(fullPathIdBuf.peekSlice()[idStartIndex .. $]);
+        });
     }
 
     /********************************************
@@ -240,6 +305,8 @@ nothrow:
 
     static void initTable()
     {
-        stringtable._init(28000);
+        enum size = 28_000;
+        stringtable._init(size);
+        fullPathStringTable._init(size);
     }
 }
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -815,7 +815,7 @@ version(Windows)
      * References:
      *  https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx
      */
-    package auto extendedPathThen(alias F)(const(char*) path)
+    auto extendedPathThen(alias F)(const(char*) path)
     {
         return path.toWStringzThen!((wpath)
         {
@@ -828,7 +828,7 @@ version(Windows)
                                                 null /*filePartBuffer*/);
             if (pathLength == 0)
             {
-                return F(""w.ptr);
+                return F((wchar*).init);
             }
 
             // wpath is the UTF16 version of path, but to be able to use
@@ -851,7 +851,7 @@ version(Windows)
 
             if (absPathRet == 0 || absPathRet > absPath.length - prefix.length)
             {
-                return F(""w.ptr);
+                return F((wchar*).init);
             }
 
             return F(absPath.ptr);
@@ -868,17 +868,18 @@ version(Windows)
      * Returns:
      *  The result of calling F on the UTF16 version of str.
      */
-    private auto toWStringzThen(alias F)(const(char*) str) nothrow
+    private auto toWStringzThen(alias F)(const(char)* str) nothrow
     {
         import core.stdc.string: strlen;
         import core.stdc.stdlib: malloc, free;
+        import core.sys.windows.winnls: MultiByteToWideChar;
 
         wchar[1024] buf;
         // cache this for efficiency
         const int strLength = cast(int)(strlen(str) + 1);
         // first find out how long the buffer must be to store the result
         const length = MultiByteToWideChar(0 /*codepage*/, 0 /*flags*/, str, strLength, null, 0);
-        if (!length) return F(""w.ptr);
+        if (!length) return F((wchar*).init);
 
         auto ret = length > buf.length
             ? (cast(wchar*)malloc(length * wchar.sizeof))
@@ -893,5 +894,81 @@ version(Windows)
         assert(length == length2); // should always be true according to the API
 
         return F(ret);
+    }
+
+}
+
+version(Posix)
+{
+
+    /**************************
+    Takes a callable F and applies it to the result of converting
+    `fileName` to an absolute file path (char*)
+    Params:
+        fileName = The file name to be converted to an absolute path
+    Returns: Whatever `F` returns.
+    */
+    auto absPathThen(alias F)(const(char)* fileName)
+    {
+        import core.sys.posix.stdlib: realpath, free;
+        auto absPath = realpath(fileName, null /* realpath allocates */);
+        scope(exit) free(absPath);
+        return F(absPath);
+    }
+}
+else
+{
+    /**************************
+    Takes a callable F and applies it to the result of converting
+    `fileName` to an absolute file path (char*)
+    Params:
+        fileName = The file name to be converted to an absolute path
+    Returns: Whatever `F` returns.
+     */
+    auto absPathThen(alias F)(const(char)* fileName)
+    {
+        import core.sys.windows.winnls: WideCharToMultiByte;
+        import core.stdc.stdlib: malloc, free;
+
+        return fileName.extendedPathThen!((wpath) {
+                // first find out how long the buffer must be to store the result
+                const length = WideCharToMultiByte(0,    // code page
+                                                   0,    // flags
+                                                   wpath,
+                                                   -1,   // wpath len, -1 is null terminated
+                                                   null, // multibyte output ptr
+                                                   0,    // multibyte output length
+                                                   null, // default char
+                                                   null, // if used default char
+                );
+
+                if(!length) return F((char*).init);
+
+                char[1024] buf;
+
+                scope multibyteBuf = length > buf.length
+                    ? (cast(char*)malloc(length * char.sizeof))
+                    : &buf[0];
+                scope (exit)
+                {
+                    if (multibyteBuf != &buf[0])
+                        free(multibyteBuf);
+                }
+
+                // first find out how long the buffer must be to store the result
+                const length2 = WideCharToMultiByte(0,    // code page
+                                                    0,    // flags
+                                                    wpath,
+                                                    -1,   // wpath len, -1 is null terminated
+                                                    multibyteBuf,
+                                                    length,
+                                                    null, // default char
+                                                    null, // if used default char
+                );
+
+                assert(length == length2);
+
+                return F(multibyteBuf);
+        });
     }
 }
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -957,7 +957,7 @@ else
                         free(multibyteBuf);
                 }
 
-                // first find out how long the buffer must be to store the result
+                // now store the result
                 const length2 = WideCharToMultiByte(0,    // code page
                                                     0,    // flags
                                                     wpath,
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -901,9 +901,10 @@ version(Windows)
 version(Posix)
 {
 
-    /**************************
+    /**
     Takes a callable F and applies it to the result of converting
     `fileName` to an absolute file path (char*)
+
     Params:
         fileName = The file name to be converted to an absolute path
     Returns: Whatever `F` returns.
@@ -918,9 +919,10 @@ version(Posix)
 }
 else
 {
-    /**************************
+    /**
     Takes a callable F and applies it to the result of converting
     `fileName` to an absolute file path (char*)
+
     Params:
         fileName = The file name to be converted to an absolute path
     Returns: Whatever `F` returns.
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -944,7 +944,7 @@ else
 
                 if(!length) return F((char*).init);
 
-                char[1024] buf;
+                char[1024] buf = void;
 
                 scope multibyteBuf = length > buf.length
                     ? (cast(char*)malloc(length * char.sizeof))
--- a/src/dmd/dmangle.d
+++ b/src/dmd/dmangle.d
@@ -1098,7 +1098,7 @@ public:
 }
 
 /// Returns: `true` if the given character is a valid mangled character
-package bool isValidMangling(dchar c)
+package bool isValidMangling(dchar c) nothrow
 {
     return
         c >= 'A' && c <= 'Z' ||
--- a/src/dmd/identifier.h
+++ b/src/dmd/identifier.h
@@ -37,6 +37,7 @@ public:
     int dyncast() const;
 
     static StringTable stringtable;
+    static StringTable fullPathStringTable;
     static Identifier *generateId(const char *prefix);
     static Identifier *generateId(const char *prefix, size_t i);
     static Identifier *idPool(const char *s, unsigned len);
--- a/test/runnable/imports/another_module_with_tests.d
+++ b/test/runnable/imports/another_module_with_tests.d
@@ -1,0 +1,3 @@
+unittest {}
+unittest {}
+unittest { assert(false); }
--- a/test/runnable/issue16995.d
+++ b/test/runnable/issue16995.d
@@ -1,8 +1,9 @@
 // REQUIRED_ARGS: -unittest
 // COMPILE_SEPARATELY
-// EXTRA_SOURCES: imports/module_with_tests.d
+// EXTRA_SOURCES: imports/module_with_tests.d imports/another_module_with_tests.d
 
 import imports.module_with_tests;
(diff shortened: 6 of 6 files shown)


wilzbach:
> I tried creating the PR to stable instead of master but for reasons I don't understand a whole lot of commits came along with it. And yes, I'd rebased to upstream/stable.

Try using `--onto` the next time:

```
git rebase --onto upstream/stable upstream/master
```

I just rebased this to stable for you ;-)

atilaneves:
I did `--onto`. I don't know what happened.

Anyway, the semaphore failure is due to stable/master differences (I got that locally).

DAutoTest seems to be entering a pathological dub dependency graph that has nothing to do with this.

atilaneves:
I'd rather get this merged to master and miss dmd 2.081.2 than deal with buildkite and semaphoreci failing for no good reason.

src/dmd/identifier.d
```diff
@@ -153,23 +161,67 @@ nothrow:
...
+            // 16995, 18097, 18111, 18880, 18868, 19058.
+            OutBuffer fullPathIdBuf;
+            if(absPath)
+            {
+                // replace characters that demangle can't handle
+                for(auto ptr = absPath; *ptr != '\0'; ++ptr)
+                {
+                    if(*ptr == '/' || *ptr == '\\' || *ptr == '.' || *ptr == '?' || *ptr == ':')
+                        *ptr = '_';
+                }
```
jacob-carlborg: Can the linker (or any other tools that deal with mangled symbols) handle these symbols? I'm asking because of the comment and why demangle isn't updated instead.

src/dmd/identifier.d
```diff
@@ -153,23 +161,67 @@ nothrow:
...
+
+            // This block generates an identifier that is prefixed by the absolute path of the file
+            // being compiled. The reason this is necessary is that we want unique identifiers per
+            // module, but the identifiers are generated before the module information is available.
+            // To guarantee that each generated identifier is unique without modules, we make them
+            // unique to each absolute file path. This also makes it consistent even if the files
+            // are compiled separately. See issues
+            // 16995, 18097, 18111, 18880, 18868, 19058.
+            OutBuffer fullPathIdBuf;
+            if(absPath)
```
jacob-carlborg: Space between `if`/`for`/`while` etc. and the opening parenthesis.

src/dmd/identifier.d
```diff
@@ -153,23 +161,67 @@ nothrow:
...
+                fullPathIdBuf.writestring(absPath);
+                fullPathIdBuf.writestring("_");
+            }
+
+            fullPathIdBuf.writestring(idBuf.peekSlice());
+            const fullPathIdLength = fullPathIdBuf.peekSlice().length;
+            uint counter = 1;
+
+            // loop until we can't find the absolute path ~ identifier, adding a counter suffix each time
+            while (fullPathStringTable.lookup(fullPathIdBuf.peekSlice().ptr, fullPathIdBuf.peekSlice().length) !is null)
```
jacob-carlborg: Please add a `const(char)[]` overload to `lookup`.

src/dmd/identifier.d
```diff
@@ -153,23 +161,67 @@ nothrow:
...
+                fullPathIdBuf.print(counter++);
+            }
+
+            // `idStartIndex` is the start of the "true" identifier. We don't actually use the absolute
+            // file path in the generated identifier since the module system makes sure that the fully
+            // qualified name is unique.
+            const idStartIndex = fullPathIdLength - idBuf.peekSlice().length;
+
+            // Remember the full path identifier to avoid possible future collisions
+            fullPathStringTable.insert(fullPathIdBuf.peekString(),
```
jacob-carlborg: Same here, add an overload for `insert` taking `const(char)[]`.

src/dmd/identifier.d
```diff
@@ -124,6 +125,13 @@ nothrow:
 
     extern (C++) static __gshared StringTable stringtable;
 
+    /**
+       A secondary string table is used to guarantee that we generate unique
+       identifiers per module. See generateIdWithLoc and issues
+       16995, 18097, 18111, 18880, 18868, 19058
+     */
+    private extern (C++) static __gshared StringTable fullPathStringTable;
```
jacob-carlborg: If this is `private` there shouldn't be a reason for it being `extern(C++)`.

src/dmd/root/filename.d
```diff
@@ -793,6 +793,7 @@ version(Windows)
      */
     private int _mkdir(const(char)* path) nothrow
     {
+        import core.sys.windows.winbase: CreateDirectoryW;
```
jacob-carlborg: Why is this necessary?

src/dmd/root/filename.d
```diff
@@ -810,8 +811,10 @@ version(Windows)
      * References:
      *  https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx
      */
-    package auto extendedPathThen(alias F)(const(char*) path)
+    auto extendedPathThen(alias F)(const(char*) path)
     {
+        import core.sys.windows.winbase: GetFullPathNameW;
```
jacob-carlborg: Why is this necessary?

src/dmd/root/filename.d
```diff
@@ -889,4 +893,77 @@ version(Windows)
...
+     * `fileName` to an absolute file path (char*)
+     */
+    auto absPathThen(alias F)(const(char)* fileName)
+    {
+        char[1024] realPathBuffer;
+        auto absPath = realpath(fileName, &realPathBuffer[0]);
+        return F(absPath);
+    }
+
+    extern(C) char* realpath(const(char)* file_name, char* resolved_name) @nogc nothrow;
```
jacob-carlborg: 1. Why not import this from druntime?
2. This should be private

src/dmd/root/filename.d
```diff
@@ -889,4 +893,77 @@ version(Windows)
...
+}
+
+version(Posix)
+{
+
+    /**************************
+     * Takes a callable F and applies it to the result of converting
+     * `fileName` to an absolute file path (char*)
+     */
+    auto absPathThen(alias F)(const(char)* fileName)
```
jacob-carlborg: Can a `const(char)[]` parameter be used instead.

src/dmd/root/filename.d
```diff
@@ -889,4 +893,77 @@ version(Windows)
...
+
+                char[1024] buf;
+
+                auto multibyteBuf = length > buf.length
+                    ? (cast(char*)malloc(length * char.sizeof))
+                    : &buf[0];
+                scope (exit)
+                {
+                    if (multibyteBuf != &buf[0])
+                        free(multibyteBuf);
```
jacob-carlborg: Is there anything preventing the parameter to `F` from escaping?

jacob-carlborg (changes requested):

src/dmd/identifier.d
atilaneves: I don't know if the linker can handle it. Maybe updating demangle is a good idea, but even if it is it's not the responsibility of this PR to handle that.

Similar code existed before in the long chain of bug fixes mentioned in the comments.

src/dmd/identifier.d
atilaneves: I thought about it, but again, PRs are supposed to be focused. I only made the changes to `filename.d` because I had to.

src/dmd/identifier.d
atilaneves: Same reply as the previous comment.

src/dmd/identifier.d
atilaneves: I don't know why the other one is. Maybe because of the backend. Speaking of which, I need to add this to the .h file.

I'm following the existing pattern.

src/dmd/root/filename.d
atilaneves: It isn't, but I copied the code around and it failed to compile because of the import. It was one my one concession to "do only what you must". I don't mind deleting it.

src/dmd/root/filename.d
atilaneves: Same reply as above.

src/dmd/root/filename.d
atilaneves: That would complicate the code given that the Win32 API only deals in null-terminated strings.

src/dmd/root/filename.d
atilaneves: 1. That didn't occur to me. I didn't even know importing from druntime was allowed in dmd.
2. Agreed.

src/dmd/root/filename.d
atilaneves: I changed `auto` to `scope`.

wilzbach:
>  I don't know how these CI pipelines are set up. 
> I'd rather get this merged to master and miss dmd 2.081.2 than deal with buildkite and semaphoreci failing for no good reason.

Fix for Buildkite: https://github.com/dlang/ci/pull/271
Not sure how to fix SemaphoreCI as it doesn't expose the target branch and we probably can only find it out via the GitHub API.

> DAutoTest seems to be entering a pathological dub dependency graph that has nothing to do with this either.

Yeah sadly that one is known :/

> > I'd rather get this merged to master and miss dmd 2.081.2 than deal with buildkite and semaphoreci failing for no good reason.

Fair enough, but keep in mind that the master-stable branch-off will happen around 15.8 (i.e. in one week). So if this hasn't been merged by then, you might need to rebase to stable again if you want this to be part of 2.082.

Sorry for the CI annoyances here :/

atilaneves:
I hope it gets merged soon but that's out of my control. It _is_ a regression fix though...

src/dmd/identifier.d
jacob-carlborg: We should avoid adding new code that uses C strings or passing pointer and length separately. We're trying to improve the code, not digging the hole deeper we are already in.

atilaneves: I agree that it's simple. I agree that it makes the code better. I agree that it should be done. However, it goes against what dlang-bot posted above: "My PR is as minimal as possible". 

It also makes the code better to localise imports, yet in this same review you asked me to delete them.

I'm trying to fix a bug, not change the interface to `StringTable`.

Feel free to submit a PR that adds slice overloads.

jacob-carlborg: I've already submitted a PR that adds those overloads https://github.com/dlang/dmd/pull/8547.

jacob-carlborg: An overload taking `const(char)[]` has been added and merged. Please rebase and use the new overload.

src/dmd/identifier.d
jacob-carlborg: An overload taking `const(char)[]` has been added and merged. Please rebase and use the new overload.

src/dmd/identifier.d
atilaneves: It's been merged to master, this is targeting stable.

jacob-carlborg: Oh, my mistake.

atilaneves:
Green and ready to go.

src/dmd/identifier.d
jacob-carlborg: Master has been merged into stable now https://github.com/dlang/dmd/blob/stable/src/dmd/root/stringtable.d#L127.

src/dmd/root/filename.d
JinShil: We would prefer all new functions have full DDoc headers including `Params` and `Returns`.  Also see https://dlang.org/dstyle.html#phobos_documentation

src/dmd/identifier.d
JinShil: I recommend not following existing patterns.  We need to get rid of this cruft somehow.

src/dmd/root/filename.d
JinShil: Although a `const(char)[]` would complicate the code, that complication would be encapsulated in this function, and that's good IMO.

src/dmd/identifier.d
atilaneves: There are header files here for a reason. Unless they disappear, I think `extern(C++)` is the way to go.

src/dmd/root/filename.d
atilaneves: `Loc.filename` isn't `char[]`, it's `const(char)*`. That would require a _lot_ of conversions. `const(char)*` is used all over the codebase from the time it was written in C++. Converting `const(char)*` to `char[]` back to a C string for _both_ the Windows and Posix APIs doesn't make much sense to me.

I'm trying to fix a bug, not refactor dmd. Yes, one should refactor as one goes, but C strings are so pervasive right now I'd need dozens of PRs just to get to the point that I can fix this regression, which _seriously_ affects my workflow and is stopping us from upgrading the compiler version at work.

src/dmd/root/filename.d
```diff
@@ -894,4 +895,82 @@ version(Windows)
...
+version(Posix)
+{
+
+    /**************************
+     * Takes a callable F and applies it to the result of converting
+     * `fileName` to an absolute file path (char*)
+     * Params:
+     *     fileName: The file name to be converted to an absolute path
+     * Returns: Whatever `F` returns.
+     */
```
JinShil: Please remove the "*" prefix, and put a newline above "Params" and "Returns".  See the example at https://dlang.org/dstyle.html#phobos_documentation

JinShil (approved):
LGTM minus nits.

src/dmd/root/filename.d
```diff
@@ -828,7 +828,7 @@ version(Windows)
                                                 null /*filePartBuffer*/);
             if (pathLength == 0)
             {
-                return F(""w.ptr);
+                return F(cast(wchar*) null);
```
Geod24: `(wchar*).init`

JinShil (approved):

src/dmd/identifier.d
wilzbach: `static` and `__gshared` are redundant. Dscanner which is in master would have complained about this.

src/dmd/root/filename.d
```diff
@@ -828,7 +828,7 @@ version(Windows)
                                                 null /*filePartBuffer*/);
             if (pathLength == 0)
             {
-                return F(""w.ptr);
+                return F((wchar*).init);
```
wilzbach: Why not simply `null`?

src/dmd/root/filename.d
```diff
@@ -894,4 +895,82 @@ version(Windows)
...
+    Takes a callable F and applies it to the result of converting
+    `fileName` to an absolute file path (char*)
+    Params:
+        fileName = The file name to be converted to an absolute path
+    Returns: Whatever `F` returns.
+    */
+    auto absPathThen(alias F)(const(char)* fileName)
+    {
+        import core.sys.posix.stdlib: realpath;
+        char[1024] realPathBuffer;
```
wilzbach: The maximal path length on Posix is typically at least 4096!

src/dmd/root/filename.d
```diff
@@ -894,4 +895,82 @@ version(Windows)
...
+                                                   null, // if used default char
+                );
+
+                if(!length) return F((char*).init);
+
+                char[1024] buf;
+
+                scope multibyteBuf = length > buf.length
+                    ? (cast(char*)malloc(length * char.sizeof))
+                    : &buf[0];
```
wilzbach: @ other reviews: I have seen this stack buffer / malloc buffer pattern now three times this week. Maybe it's time to abstract this?

With std.experimental.allocator it would be simply:

```d
StackFront!(1024, Mallocator.instance);
```

See also: https://dlang.org/phobos/std_experimental_allocator_showcase.html

src/dmd/root/filename.d
```diff
@@ -894,4 +895,82 @@ version(Windows)
...
+    Params:
+        fileName = The file name to be converted to an absolute path
+    Returns: Whatever `F` returns.
+     */
+    auto absPathThen(alias F)(const(char)* fileName)
+    {
+        import core.sys.windows.winnls: WideCharToMultiByte;
+        import core.stdc.stdlib: malloc, free;
+        import std.stdio: writeln;
+        import std.conv: text;
```
wilzbach: Why is Phobos imported here??
(looks like it was only used for debugging, but while we do link DMD with Phobos, Walter is still against using any imports as it might make debugging DMD harder).

src/dmd/identifier.d
```diff
@@ -153,23 +161,66 @@ nothrow:
...
+            idBuf.writestring("_C");
+            idBuf.print(loc.charnum);
+
+            // This block generates an identifier that is prefixed by the absolute path of the file
+            // being compiled. The reason this is necessary is that we want unique identifiers per
+            // module, but the identifiers are generated before the module information is available.
+            // To guarantee that each generated identifier is unique without modules, we make them
+            // unique to each absolute file path. This also makes it consistent even if the files
+            // are compiled separately. See issues
+            // 16995, 18097, 18111, 18880, 18868, 19058.
```
wilzbach: Walter likes clickable links.

src/dmd/identifier.d
```diff
@@ -153,23 +161,66 @@ nothrow:
...
+            OutBuffer fullPathIdBuf;
+            if (absPath)
+            {
+                // replace characters that demangle can't handle
+                for (auto ptr = absPath; *ptr != '\0'; ++ptr)
+                {
+                    if (*ptr == '/' || *ptr == '\\' || *ptr == '.' || *ptr == '?' || *ptr == ':')
```
wilzbach: https://github.com/dlang/dmd/blob/2a35f81ff3747a6acee2f23fbe539d8487f9e2ff/src/dmd/dmangle.d#L1101-L1108

uses a different character set. It would be great to avoid having tons of different definitions of "valid mangling" characters in DMD...

src/dmd/root/filename.d
atilaneves: Well spotted, I forgot to delete this.

src/dmd/root/filename.d
atilaneves: Good point.

src/dmd/identifier.d
atilaneves: Fixed.

src/dmd/root/filename.d
atilaneves: Makes sense, but this is a bug fix targeting stable.

src/dmd/root/filename.d
atilaneves: It turns out that realpath can allocate so I changed it to do that.

src/dmd/identifier.d
atilaneves: Fixed by importing that functino from `dmd.dmangle`

src/dmd/root/filename.d
Geod24: @wilzbach : Maybe it's just a good pattern ? :)
There have been talk of making `scope foo = new char[](runtime_length);` work, which is exactly what we need.

wilzbach: Hehe my point was just that this "good pattern" could be reduced from ~10 lines to a single line with sth. like Phobos's `StackFront`

> There have been talk of making scope foo = new char[](runtime_length); work,

Yep, that would be great too (:

atilaneves:
Oh great, a cyclic dependency.

src/dmd/identifier.d
atilaneves: I tried that and import dmd.mangle.isValidMangling was a world of pain that required a change to the makefile _and_ resulted in autotester failures to the module import cyclic dependencies. I'm reverting to how it was.

src/dmd/root/filename.d
atilaneves: Because type inference doesn't work otherwise and it broke the build. Reverting.

atilaneves:
@jacob-carlborg AFAICT I made all the changes you requested.

atilaneves:
buildkite is being weird again but otherwise I think this is good to go.

wilzbach:
> buildkite is being weird again

Sorry, my fault. Should be fixed now. I restarted the build.

MartinNowak:
I'd be happy to see a solution that could also be applied to [14894 – mangling of mixins and lambdas is not unique and depends on compilation flags](https://issues.dlang.org/show_bug.cgi?id=14894), basically the same problem.

Including absolute paths in mangling has a quite a smell (and also bloats symbols sizes, though that's less important for debug builds). It would likely break hard-linked folders and copies of sources (vendored folders). As you already said yourself, module names should be unique already (https://forum.dlang.org/post/krkinfxikztqvudjopao@forum.dlang.org).

Also I'm not sure that we need to support lookup of unittests by their symbol name (hence they start with `__`). We might wanna have a second look at the resolution of [18097 – [REG2.077] Unittest function is undefined identifier](https://issues.dlang.org/show_bug.cgi?id=18097).

Overall we probably should spent more time on this and fix it once and for all, and at best also the similar problem with mixins and lambdas.

Short summary of the suffix approaches and their problems:
Line/Column number: possibly not suitable due to differences in .di/.d files (or only for unittests which are assumed to not be in .di files)
Counter: not suitable as it requires full semantic analysis when importing (but not compiling) a module and also increments depend on exact compilation flags
Hash: would prolly work, but we'd need a consistent and fast hash

At least for unittests the [suggested line/column number + counter on collision approach](https://forum.dlang.org/post/pffwzdqnarmjdckncuzo@forum.dlang.org) might be worth a second thought. If the importer can already see a unittest, why wouldn't it see all the other ones at the same loc/col?

atilaneves:
> I'd be happy to see a solution that could also be applied to 14894

That's great, but that's not what this PR is about. It's about fixing 19058. It's hard enough to keep things focussed as it is.

> Including absolute paths in mangling has a quite a smell (and also bloats symbols sizes, though that's less important for debug builds

This isn't what this PR does. It includes the absolute path iff there's a collision. Notice that none of the Phobos tests that check for error messages that print out unittest names had to be changed. And they didn't have to be changed because the absolute path isn't part of the symbol. Imagine having to write tests so that the absolute paths _were_ part of the symbol that didn't depend on the filesystem layout of the autotester - it wouldn't be fun.

> As you already said yourself, module names should be unique already 

I also said that modules aren't available until semantic, which is far too late to generate the identifier, which was the cause of some of the bugs referenced in the comments of this PR.

Without completely restructuring dmd, I don't see how these multiple issues can be solved simultaneously.

> Also I'm not sure that we need to support lookup of unittests by their symbol name (hence they start with __). We might wanna have a second look at the resolution of 18097 – [REG2.077] Unittest function is undefined identifier.

That wasn't the problem in that bug. It was the fact that linking failed after trying to get a reference to a unittest function. The included implementation-defined names were used as an illustration. There's nothing weird at all in trying to get a reference to a unittest function without using the actual identifier.

> Overall we probably should spent more time on this and fix it once and for all, and at best also the similar problem with mixins and lambdas.

And, in the meanwhile, this bug has completely broken my workflow and is currently preventing us from upgrading the dmd version at work - we're still stuck on 2.080.1 on the figurative eve of the release of 2.082.0.

This bug is not a detail. It's a regression blocking us at work. It breaks legitimate uses of the language's functionality and is currently taking a major vesion bump of 2 (or 3 or more if this carries on long enough) to get fixed despite a small PR with a fix that passes all tests.

> At least for unittests the suggested line/column number + counter on collision approach might be worth a second thought. If the importer can already see a unittest, why wouldn't it see all the other ones at the same loc/col?

This demonstrably does not, and cannot, work. It's literally the "fix" that caused this regression in the first place. The counter would need to be per module, and modules don't exist yet when the identifier is generated.

jacob-carlborg:
Would it be possible to change or set the identifier later? I mean, we do have anonymous functions.

atilaneves:
@jacob-carlborg No, we can't set the identifier later. That's what I did in [my original fix](https://github.com/dlang/dmd/commit/f527810984133fb379f99c4b12a357257a53eefb) which ended up causing [issue 18097](https://issues.dlang.org/show_bug.cgi?id=18097). There's a long list of issues related to this, which I chronicled in the [already mentioned blog post](https://forum.dlang.org/post/krkinfxikztqvudjopao@forum.dlang.org) where I asked for ideas to solve this problem.

This fix works. This fix is small. This fix doesn't affect the vast majority of symbols. I've considered this every which way I can, asked for help, and nobody has come up with a better idea that _also_ works.

I don't understand why this hasn't been merged yet. Did I mention this is a work-blocking regression?

Geod24:
It was broken last night but is fixed now. Retriggering the build did the trick.

JinShil (approved):

src/dmd/root/filename.d
```diff
@@ -894,4 +895,80 @@ version(Windows)
...
+                                                   -1,   // wpath len, -1 is null terminated
+                                                   null, // multibyte output ptr
+                                                   0,    // multibyte output length
+                                                   null, // default char
+                                                   null, // if used default char
+                );
+
+                if(!length) return F((char*).init);
+
+                char[1024] buf;
```
kinke: = void?

src/dmd/identifier.d
```diff
@@ -153,23 +166,75 @@ nothrow:
...
+                        *ptr = '_';
+                }
+
+                fullPathIdBuf.writestring(absPath);
+                fullPathIdBuf.writestring("_");
+            }
+
+            fullPathIdBuf.writestring(idBuf.peekSlice());
+            const fullPathIdLength = fullPathIdBuf.peekSlice().length;
+            uint counter = 1;
```
kinke: I think the counter/StringTable could be simplified and possibly accelerated/requiring less memory via `static uint[string][Loc] counters`, indexed by `counters[prefix][loc]`; checking if there's an existing entry and incrementing/setting it to 1 when done. I might give that a try as a follow-up.

src/dmd/root/filename.d
atilaneves: Why?

atilaneves:
I'm going on holiday today. I'm not going to reply to anything until I'm back on the 12th of September.

src/dmd/root/filename.d
JinShil: It's a minor optimization.  No need to default-initialize it if you're just going to overwrite it.

JinShil (approved):

JinShil:
I would like to merge this before the next release.  If there are any objections, please make them known soon.  If there are no show-stoppers, but you have some recommended changes, let's do them as followups.

MartinNowak:
> > Also I'm not sure that we need to support lookup of unittests by their symbol name (hence they start with __). We might wanna have a second look at the resolution of 18097 – [REG2.077] Unittest function is undefined identifier.
> 
> That wasn't the problem in that bug. It was the fact that linking failed after trying to get a reference to a unittest function. The included implementation-defined names were used as an illustration. There's nothing weird at all in trying to get a reference to a unittest function without using the actual identifier.

What fails with 2.077.0 is `auto a = &mixin(__traits(identifier, __traits(parent, {  })));`, but not `auto b = &__traits(parent, {  });`. So it's only lookup via the internal unittest name (`__unittest_bug_d_5_1`) that has a problem and that might be acceptable.

PR #8544, opened 2018-08-08
aggregate.d, replace `for` loops with `foreach` over aggregates

ghost:
Also addresses a few D-Scanner warnings for the module
- line len over 120
- local var can be const

(https://github.com/dlang/dmd/projects/27)

### DIFF
--- a/src/dmd/aggregate.d
+++ b/src/dmd/aggregate.d
@@ -193,7 +193,7 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
             if (v.aliassym)
                 return 0;   // If this variable was really a tuple, skip it.
 
-            with (STC) if (v.storage_class & (static_ | extern_ | tls | gshared | manifest | ctfe | templateparameter))
+            if (v.storage_class & (STC.static_ | STC.extern_ | STC.tls | STC.gshared | STC.manifest | STC.ctfe | STC.templateparameter))
                 return 0;
             if (!v.isField() || v.semanticRun < PASS.semanticdone)
                 return 1;   // unresolvable forward reference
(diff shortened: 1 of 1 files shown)


src/dmd/aggregate.d
```diff
@@ -193,7 +193,7 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
             if (v.aliassym)
                 return 0;   // If this variable was really a tuple, skip it.
 
-            if (v.storage_class & (STC.static_ | STC.extern_ | STC.tls | STC.gshared | STC.manifest | STC.ctfe | STC.templateparameter))
+            with (STC) if (v.storage_class & (static_ | extern_ | tls | gshared | manifest | ctfe | templateparameter))
```
jacob-carlborg: I have a suspicion Walter won't like this.

BBasile: Okay, cancelled this change.

wilzbach (approved):

WalterBright:
Be extremely careful about converting loops in dmd to foreach. The reason is that some of the routines can add members to the arrays while in the loop. A number of these changes had to be reverted because of that.

Basically, do not do this conversion unless you are absolutely sure that isn't happening.

src/dmd/aggregate.d
```diff
@@ -730,11 +730,8 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
                 }
             }
 
-            for (size_t i = 0; i < members.dim; i++)
-            {
-                auto sm = (*members)[i];
+            foreach (sm; *members)
```
WalterBright: this loop can add members, do not convert

src/dmd/aggregate.d
```diff
@@ -216,9 +216,8 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
             return 0;
         }
 
-        for (size_t i = 0; i < members.dim; i++)
+        foreach (s; *members)
```
WalterBright: This loop can add members, please revert

PR #8547, opened 2018-08-08
Add D string overloads to methods in StringTable

jacob-carlborg:
This will help reduce the surface of where C strings and pointer length pairs are used.

### DIFF
--- a/src/dmd/root/stringtable.d
+++ b/src/dmd/root/stringtable.d
@@ -103,6 +103,18 @@ public:
         pools = null;
     }
 
+    /**
+    Looks up the given string in the string table and returns its associated
+    value.
+
+    Params:
+     s = the string to look up
+     length = the length of $(D_PARAM s)
+     str = the string to look up
+
+    Returns: the string's associated value, or `null` if the string doesn't
+     exist in the string table
+    */
     extern (C++) inout(StringValue)* lookup(const(char)* s, size_t length) inout nothrow pure
     {
         const(hash_t) hash = calcHash(s, length);
@@ -111,6 +123,26 @@ public:
         return getValue(table[i].vptr);
     }
 
+    /// ditto
+    inout(StringValue)* lookup(const(char)[] str) inout nothrow pure
+    {
+        return lookup(str.ptr, str.length);
+    }
+
+    /**
+    Inserts the given string and the given associated value into the string
+    table.
+
+    Params:
+     s = the string to insert
+     length = the length of $(D_PARAM s)
+     ptrvalue = the value to associate with the inserted string
+     str = the string to insert
+     value = the value to associate with the inserted string
+
+    Returns: the newly inserted value, or `null` if the string table already
+     contains the string
+    */
     extern (C++) StringValue* insert(const(char)* s, size_t length, void* ptrvalue) nothrow
     {
         const(hash_t) hash = calcHash(s, length);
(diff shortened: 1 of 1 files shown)


src/dmd/root/stringtable.d
```diff
@@ -111,6 +111,11 @@ public:
         return getValue(table[i].vptr);
     }
 
+    inout(StringValue)* lookup(const(char)[] str) inout nothrow pure
```
JinShil: I hate to nitpick, but new public functions should have full DDoc.

jacob-carlborg: Done.

src/dmd/root/stringtable.d
```diff
@@ -103,6 +103,17 @@ public:
...
+     * Looks up the given string in the string table and returns its associated
+     * value.
+     *
+     * Params:
+     *  s = the string to look up
+     *  length = the length of $(D_PARAM s)
+     *
+     * Returns: the string's associated value, or `null` if the string doesn't
+     *  exist in the string table
+     */
```
wilzbach: FWIW the DStyle we use for new symbols at Druntime/Phobos says:

> Documentation comments should not have leading stars on each line.

https://dlang.org/dstyle.html#phobos_documentation

wilzbach (approved):

src/dmd/root/stringtable.d
jacob-carlborg: 😞

jacob-carlborg: It looks so weird. I have a feeling that most of the code out there is not using this style.

jacob-carlborg: Fixed.

src/dmd/root/stringtable.d
```diff
@@ -103,6 +103,17 @@ public:
         pools = null;
     }
 
+    /**
+    Looks up the given string in the string table and returns its associated
+    value.
+
+    Params:
+     s = the string to look up
+     length = the length of $(D_PARAM s)
```
wilzbach: Technically `str` would need to be documented too (as it gets dittoed into this documentation.)

JinShil (approved):

src/dmd/root/stringtable.d
jacob-carlborg: We need a better way to inherit documentation that `ditto`. Fixed.

jacob-carlborg:
Do these new methods need to be placed at after all C++ methods sine they cannot appear in the C++ header?

jacob-carlborg:
@JinShil @wilzbach ping.

JinShil:
> Do these new methods need to be placed at after all the C++ methods sine they cannot appear in the C++ header?

Yeah, that may be nice.  I won't hold up this PR on it though, and it can always be done as a followup.

JinShil:
Semaphore CI seems stuck, so I'm closing and re-opening.

jacob-carlborg:
>> Do these new methods need to be placed at after all the C++ methods sine they cannot appear in the C++ header?

> Yeah, that may be nice

Actually, since these functions are not virtual and don't occupy a slot in the vtable the order doesn't matter [1]. In that case I think it makes more sense to have the overloads next to each other.

[1] https://github.com/dlang/dmd/pull/8550#issuecomment-411880216

PR #8549, opened 2018-08-08
astbase.d, replace `for` loops with `foreach` over aggregates

ghost:
Also simplify a few cases where Array.setDim can be done directly in array ctor.

### DIFF
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -5102,8 +5102,7 @@ struct ASTBase
         extern (D) this(const ref Loc loc, Expression e, Expression earg1, Expression earg2)
         {
             super(loc, TOK.call, __traits(classInstanceSize, CallExp), e);
-            auto arguments = new Expressions();
-            arguments.setDim(2);
+            auto arguments = new Expressions(2);
             (*arguments)[0] = earg1;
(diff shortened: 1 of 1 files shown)


src/dmd/astbase.d
```diff
@@ -6236,7 +6230,7 @@ struct ASTBase
             const(char)* id;
         }
 
-        static __gshared SCstring* table =
+        static __gshared SCstring[31] table =
```
wilzbach: Why not `static immutable` and letting the compiler infer the size?

BBasile: Yes, done.

jacob-carlborg (approved):

RazvanN7 (approved):

src/dmd/astbase.d
```diff
@@ -330,10 +330,9 @@ struct ASTBase
             Dsymbol s = null;
             if (members)
             {
-                for (size_t i = 0; i < members.dim; i++)
+                foreach (Dsymbol sx; *members)
```
JinShil: @WalterBright  [warned](https://github.com/dlang/dmd/pull/8551#issue-207183029) about `foreach`ing over `members`

PR #8551, opened 2018-08-09
Revert "aggregate.d, replace `for` loops with `foreach` over aggregates"

WalterBright:
Reverts dlang/dmd#8544

Please DO NOT convert for loops over members[] to foreach, because members can be added within the loop.

### DIFF
--- a/src/dmd/aggregate.d
+++ b/src/dmd/aggregate.d
@@ -202,7 +202,7 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
 
             if (v.storage_class & STC.ref_)
                 return 0;
-            const tv = v.type.baseElemOf();
+            auto tv = v.type.baseElemOf();
             if (tv.ty != Tstruct)
                 return 0;
             if (ad == (cast(TypeStruct)tv).sym)
@@ -216,8 +216,9 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
             return 0;
         }
 
-        foreach (s; *members)
+        for (size_t i = 0; i < members.dim; i++)
         {
+            auto s = (*members)[i];
             if (s.apply(&func, cast(void*)this))
             {
                 if (sizeok != Sizeok.none)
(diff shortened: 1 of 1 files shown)


Geod24 (approved):
Well some of these changes are worthwhile, but I personally prefer a simple `git revert`.
How hard do you think it would be to add a test for those issues ? The testsuite should've caught it.

JinShil:
> How hard do you think it would be to add a test for those issues ?

We should be able to catch this at compile time by either leveraging `const` or by implementing a read-only wrapper type that is returned from `members.asReadOnly` or something like that.

Geod24:
I don't think that's possible. Consider:
```D
class Custom
{
  string msg;
  Custom[] arr;
}

void bar (Custom[] arr)
{
  foreach (c; arr)
  {
    c.msg = "Hello world";
    functionThatMightMutateArr(c);
  }
}
```

We don't want to disallow modifying instances of `Custom`.

WalterBright:
It's kind of an awful problem. Whether members are added or not depends on the precise path through semantic analysis, and that changes all the time. Worse, when a member is added and then gets skipped by the loop, often the bug does not manifest itself. It may show up in very puzzling ways.

JinShil:
> We don't want to disallow modifying instances of Custom.

Yes, that would be a problem with leveraging `const` due to its transitivity.  Some `foreach`'s will work with `const`, some won't depending on what they do.  But we should still be able to create a readonly wrapper type:

```d
ReadonlyRange(T) asReadOnly(T)(T[] mutableArray);

void bar (ReadonlyRange(T) r)
{
  foreach (c; r)
  {
    c.msg = "Hello world";          // OK
    functionThatMightMutateArr(c);  // OK
    r.add(new T());                 //  Error:  `add` doesn't even exist
  }
}

void main()
{
    Custom[] arr;
    bar(arr.asReadOnly);
}
```

The semantics of `ReadOnlyRange(T)` will not allow addition or removal of items, but will allow the individual items to be mutable.

It will likely require refactoring more than just the `for` loop itself.

> Whether members are added or not depends on the precise path through semantic analysis, and that changes all the time.

Not a problem if the read-only semantics are enforced through the type system.  The compiler will simply complain if you try to change the collection through a read-only type.  Those `for` loops that do modify the array they are iterating through simply cannot be refactored, or must be refactored in a way that moves the addition or removal of items outside of the loop itself.

This principle works well in C# (See https://msdn.microsoft.com/en-us/library/e78dcd75(v=vs.110).aspx)

PR #8554, opened 2018-08-10
dtemplate Tuple, add a ctor taking num elements

ghost:


src/dmd/dtemplate.d
```diff
@@ -476,6 +476,13 @@ extern (C++) final class Tuple : RootObject
 {
     Objects objects;
 
+    extern (D) this(){}
+
+    extern (D) this(size_t numObjects)
```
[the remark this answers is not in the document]
BBasile: Yes of course. This is just a bit strange to start doing it when the practice is not generalized.
I had to amend anyway since the default ctor can be avoided using a default numObjects.

src/dmd/dtemplate.d
```diff
@@ -476,6 +476,17 @@ extern (C++) final class Tuple : RootObject
 {
     Objects objects;
 
+    /**
+     *  Constructs a new instance.
+     *  Params:
+     *      numObjects = The initial number of objects.
+     */
+    extern (D) this(size_t numObjects = 0)
```
JinShil: See #12 here:  https://forum.dlang.org/post/p6oibo$1lmi$1@digitalmars.com

I don't really like it, but you can avoid the default parameter by forwarding to a parameter-less constructor in the body of this constructor.

src/dmd/dtemplate.d
```diff
@@ -476,6 +476,17 @@ extern (C++) final class Tuple : RootObject
 {
     Objects objects;
 
+    /**
+     *  Constructs a new instance.
```
JinShil: Please remove the prepended `*`.

src/dmd/dtemplate.d
BBasile: Yes but it's either that or a useless default ctor since default ctors ares till not auto generated ;)

src/dmd/dtemplate.d
```diff
@@ -476,6 +476,17 @@ extern (C++) final class Tuple : RootObject
...
+     *  Params:
```
JinShil: Newline before `params`

src/dmd/dtemplate.d
JinShil: I'm guessing the preference would be:

```D
/**
Constructs a new instance.
*/
extern (D) this()
{
    this(0);
}

/**
Constructs a new instance.

Params:
    numObjects = The initial number of objects.
*/
extern (D) this(size_t numObjects)
{
    objects.setDim(numObjects);
}
```

src/dmd/dtemplate.d
```diff
@@ -476,6 +476,19 @@ extern (C++) final class Tuple : RootObject
 {
     Objects objects;
 
+    extern (D) this() {}
+
+    /**
+     *  Constructs a new instance.
```
WalterBright: This line is spurious, as that's what a `this` constructor does. Please delete it.

BBasile: ok done.

jacob-carlborg: I think all documented symbols should have a summary. This allows to automatically generate an index table, i.e. something like the table at the top of [1].

[1] https://dlang.org/phobos/std_array.html

src/dmd/dtemplate.d
jacob-carlborg: We really should have an updated style guide for all this that is linked from the readme or the contribute documentation.

JinShil (approved):

jacob-carlborg (approved):

PR #8555, opened 2018-08-10
Convert the JSON code to use D slices instead of pointers

Geod24:
Since we want to get rid of strlen, I tried to convert a simple global to a D slice and was blocked by the JSON module. So here is the module conversion to D.
Some simple things were added (e.g. `RootObject.toString` which just `strlen` on `toChars`), but I avoided changing the AST as much as possible, so things like `kind()` and `Loc.filename` can be done in another PR.

### DIFF
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -38,6 +38,7 @@ import dmd.root.port;
 import dmd.semantic2;
 import dmd.semantic3;
 import dmd.target;
+import dmd.utils;
 import dmd.visitor;
 
 version(Windows) {
@@ -1337,19 +1338,24 @@ struct ModuleDeclaration
         this.isdeprecated = isdeprecated;
     }
 
-    extern (C++) const(char)* toChars()
+    extern (C++) const(char)* toChars() const
     {
         OutBuffer buf;
         if (packages && packages.dim)
         {
-            for (size_t i = 0; i < packages.dim; i++)
+            foreach (pid; *packages)
             {
-                Identifier pid = (*packages)[i];
-                buf.writestring(pid.toChars());
+                buf.writestring(pid.toString());
                 buf.writeByte('.');
             }
         }
-        buf.writestring(id.toChars());
+        buf.writestring(id.toString());
         return buf.extractString();
+    }
+
+    /// Provide a human readable representation
+    extern (D) const(char)[] toString() const
+    {
+        return this.toChars().toDString;
     }
 }
--- a/src/dmd/hdrgen.d
+++ b/src/dmd/hdrgen.d
@@ -3347,6 +3347,16 @@ extern (C++) const(char)* stcToChars(ref StorageClass stc)
     return null;
 }
 
+
+/**
+ * Returns:
+ *   a human readable representation of `stc`
+ */
+extern (D) const(char)[] stcToString(ref StorageClass stc)
+{
+    return stcToChars(stc).toDString;
+}
+
 extern (C++) void trustToBuffer(OutBuffer* buf, TRUST trust)
 {
     const(char)* p = trustToChars(trust);
@@ -3354,7 +3364,19 @@ extern (C++) void trustToBuffer(OutBuffer* buf, TRUST trust)
         buf.writestring(p);
 }
 
+/**
+ * Returns:
+ *   a human readable representation of `trust`,
+ *   which is the token `trust` corresponds to
+ */
 extern (C++) const(char)* trustToChars(TRUST trust)
+{
+    /// Works because we return a literal
+    return trustToString(trust).ptr;
+}
+
+/// Ditto
+extern (D) string trustToString(TRUST trust)
 {
     final switch (trust)
     {
@@ -3416,7 +3438,18 @@ extern (C++) void protectionToBuffer(OutBuffer* buf, Prot prot)
     }
 }
 
+/**
+ * Returns:
+ *   a human readable representation of `kind`
+ */
 extern (C++) const(char)* protectionToChars(Prot.Kind kind)
+{
+    // Null terminated because we return a literal
+    return protectionToString(kind).ptr;
+}
+
+/// Ditto
+extern (D) string protectionToString(Prot.Kind kind)
 {
     final switch (kind)
     {
--- a/src/dmd/json.d
+++ b/src/dmd/json.d
@@ -33,6 +33,8 @@ import dmd.id;
 import dmd.identifier;
 import dmd.mtype;
 import dmd.root.outbuffer;
+import dmd.root.rootobject;
+import dmd.utils;
 import dmd.visitor;
 
 version(Windows) {
@@ -47,12 +49,13 @@ private extern (C++) final class ToJsonVisitor : Visitor
 public:
     OutBuffer* buf;
     int indentLevel;
-    const(char)* filename;
+    const(char)[] filename;
 
     extern (D) this(OutBuffer* buf)
     {
         this.buf = buf;
     }
+
 
     void indent()
     {
@@ -83,11 +86,10 @@ public:
         buf.writeByte('\"');
     }
 
-    void stringPart(const(char)* s)
+    extern(D) void stringPart(const(char)[] s)
     {
-        for (; *s; s++)
+        foreach (char c; s)
         {
-            char c = cast(char)*s;
             switch (c)
             {
             case '\n':
@@ -128,7 +130,7 @@ public:
     /*********************************
      * Encode string into buf, and wrap it in double quotes.
      */
-    void value(const(char)* s)
+    extern(D) void value(const(char)[] s)
     {
         stringStart();
         stringPart(s);
@@ -153,7 +155,7 @@ public:
     /*********************************
      * Item is an intented value and a comma, for use in arrays
      */
-    void item(const(char)* s)
+    extern(D) void item(const(char)[] s)
     {
         indent();
         value(s);
@@ -221,7 +223,7 @@ public:
     }
 
     // Json object property functions
-    void propertyStart(const(char)* name)
+    extern(D) void propertyStart(const(char)[] name)
     {
         indent();
         value(name);
@@ -235,7 +237,7 @@ public:
      name = the name of the object property
      s = the string value of the object property
     */
-    void property(const(char)* name, const(char)* s)
+    extern(D) void property(const(char)[] name, const(char)[] s)
     {
         if (s is null)
             return;
@@ -251,7 +253,7 @@ public:
      name = the name of the object property
      s = the string value of the object property
     */
-    void requiredProperty(const(char)* name, const(char)* s)
+    extern(D) void requiredProperty(const(char)[] name, const(char)[] s)
     {
         propertyStart(name);
         if (s is null)
@@ -261,21 +263,21 @@ public:
         comma();
     }
 
-    void property(const(char)* name, int i)
+    extern(D) void property(const(char)[] name, int i)
     {
         propertyStart(name);
         value(i);
         comma();
     }
 
-    void propertyBool(const(char)* name, const bool b)
+    extern(D) void propertyBool(const(char)[] name, const bool b)
     {
         propertyStart(name);
         valueBool(b);
         comma();
     }
 
-    void property(const(char)* name, TRUST trust)
+    extern(D) void property(const(char)[] name, TRUST trust)
     {
         final switch (trust)
         {
@@ -295,7 +297,7 @@ public:
         }
     }
 
-    void property(const(char)* name, PURE purity)
+    extern(D) void property(const(char)[] name, PURE purity)
     {
         final switch (purity)
         {
@@ -318,7 +320,7 @@ public:
         }
     }
 
-    void property(const(char)* name, const LINK linkage)
+    extern(D) void property(const(char)[] name, const LINK linkage)
     {
         final switch (linkage)
         {
@@ -352,7 +354,7 @@ public:
         }
     }
 
-    void propertyStorageClass(const(char)* name, StorageClass stc)
+    extern(D) void propertyStorageClass(const(char)[] name, StorageClass stc)
     {
         stc &= STCStorageClass;
         if (stc)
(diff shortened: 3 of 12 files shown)


src/dmd/json.d
```diff
@@ -391,26 +397,26 @@ public:
         }
     }
 
-    void property(const(char)* name, Type type)
+    extern(D) void property(const(char)[] name, Type type)
     {
         if (type)
         {
-            property(name, type.toChars());
+            property(name, RootObject.toString(type));
```
WalterBright: Not too thrilled about this. It should be type.toString(). Yes, the C++ code cannot call it, but all you need to do is put a dummy function declaration there on the C++ side so the vtbl[] entries line up.

Geod24: Fixed

src/dmd/dmodule.d
```diff
@@ -1337,19 +1337,24 @@ struct ModuleDeclaration
...
+                buf.writestring(pid.toString());
                 buf.writeByte('.');
             }
         }
-        buf.writestring(id.toChars());
+        buf.writestring(id.toString());
         return buf.extractString();
     }
+
+    extern (D) const(char)[] toString() const
```
JinShil: DDoc?

src/dmd/hdrgen.d
```diff
@@ -3347,6 +3347,12 @@ extern (C++) const(char)* stcToChars(ref StorageClass stc)
     return null;
 }
 
+extern (D) const(char)[] stcToString(ref StorageClass stc)
```
JinShil: DDoc?

src/dmd/hdrgen.d
```diff
@@ -3355,6 +3361,12 @@ extern (C++) void trustToBuffer(OutBuffer* buf, TRUST trust)
 }
 
 extern (C++) const(char)* trustToChars(TRUST trust)
+{
+    /// Works because we return a literal
+    return trustToString(trust).ptr;
+}
+
+extern (D) string trustToString(TRUST trust)
```
JinShil: DDoc?

src/dmd/hdrgen.d
```diff
@@ -3417,6 +3429,12 @@ extern (C++) void protectionToBuffer(OutBuffer* buf, Prot prot)
 }
 
 extern (C++) const(char)* protectionToChars(Prot.Kind kind)
+{
+    // Null terminated because we return a literal
+    return protectionToString(kind).ptr;
+}
+
+extern (D) string protectionToString(Prot.Kind kind)
```
JinShil: DDoc?

src/dmd/json.d
```diff
@@ -361,22 +368,21 @@ public:
...
+    extern(D) void property(const(char)[] linename, const(char)[] charname, Loc* loc)
     {
         if (loc)
         {
-            const(char)* filename = loc.filename;
-            if (filename)
+            if (auto filename = toDString(loc.filename))
             {
-                if (!this.filename || strcmp(filename, this.filename))
+                if (!this.filename.length || filename != this.filename)
```
jacob-carlborg: I'm guessing you're trying keep the new to code be as close as possible to the old code but this first expression (`!this.filename.length`) isn't needed anymore?

jacob-carlborg (approved):

src/dmd/dmodule.d
WalterBright: As I mentioned before, toString() is used pervasively in dmd, and as long as it does what the other toString()'s do, adding a Ddoc for it is redundant.

Geod24:
Updated according to feedback

JinShil (approved):

src/dmd/dmodule.d
```diff
@@ -1337,19 +1337,25 @@ struct ModuleDeclaration
...
+                buf.writestring(pid.toString());
                 buf.writeByte('.');
             }
         }
-        buf.writestring(id.toChars());
+        buf.writestring(id.toString());
         return buf.extractString();
     }
+
+    /// Provide an human readable representation
```
ibuclaw: Provide a human

src/dmd/hdrgen.d
```diff
@@ -3347,14 +3347,37 @@ extern (C++) const(char)* stcToChars(ref StorageClass stc)
     return null;
 }
 
+
+/**
+ * Returns:
+ *   an human readable representation of `stc`
```
ibuclaw: A human

src/dmd/hdrgen.d
```diff
@@ -3347,14 +3347,37 @@ extern (C++) const(char)* stcToChars(ref StorageClass stc)
...
 extern (C++) void trustToBuffer(OutBuffer* buf, TRUST trust)
 {
     const(char)* p = trustToChars(trust);
     if (p)
         buf.writestring(p);
 }
 
+/**
+ * Returns:
+ *   an human readable representation of `trust`,
```
ibuclaw: A human

src/dmd/hdrgen.d
```diff
@@ -3416,7 +3439,18 @@ extern (C++) void protectionToBuffer(OutBuffer* buf, Prot prot)
     }
 }
 
+/**
+ * Returns:
+ *   an human readable representation of `kind`
```
ibuclaw: A human

src/dmd/json.d
```diff
@@ -83,11 +91,10 @@ public:
         buf.writeByte('\"');
     }
 
-    void stringPart(const(char)* s)
+    extern(D) void stringPart(const(char)[] s)
```
ibuclaw: Wouldn't all non-visit members of this visitor be private? Ergo you can just add `extern(D):` at the top of this body rather than adding it for each individual function.

Think about it for future refactor.

src/dmd/root/outbuffer.d
```diff
@@ -459,4 +459,3 @@ char[] unsignedToTempString(ulong value, char[] buf, uint radix = 10) @safe
     } while (value);
     return buf[i .. $];
 }
-
```
ibuclaw: Seems like a leftover from a previous attempt here.

ibuclaw (changes requested):

src/dmd/root/outbuffer.d
Geod24: It is, but also its own commit

src/dmd/json.d
Geod24: True that. At the moment I'm putting `extern(D)` right and left (got more slice-ification coming), but I also had weird bugs (compiler SEGV) being triggered in the past some I decided to go for verbose and simple. But when enough is converted, we surely can do that.

Geod24:
Fixed typos

src/dmd/dmodule.d
rainers: While I agree that a comment is not really necessary here, you have to add some ddoc comment to make the function appear in the documentation at all. Not sure if that is a good thing as you have to litter the code with empty comments this way. Is there an option to put all (public) symbols into the documentation?

Geod24: For the record I added a (trivial) DDOC comment just for this reason (documentation)

src/dmd/json.d
```diff
@@ -47,13 +48,20 @@ private extern (C++) final class ToJsonVisitor : Visitor
...
     extern (D) this(OutBuffer* buf)
     {
         this.buf = buf;
     }
 
+    /// Helper: Remove when all usages of strlen are gone
+    static extern(D) const(char)[] toDString (const(char)* s)
+    {
+        return s ? s[0 .. strlen(s)] : null;
+    }
```
wilzbach: Why not put this as a global function into `utils` or similar? It's used at other places too and doesn't really fit in `ToJsonVisitor`

src/dmd/hdrgen.d
```diff
@@ -3347,14 +3347,37 @@ extern (C++) const(char)* stcToChars(ref StorageClass stc)
...
 
+
+/**
+ * Returns:
+ *   a human readable representation of `stc`
+ */
+extern (D) const(char)[] stcToString(ref StorageClass stc)
+{
+    auto p = stcToChars(stc);
+    return p[0 .. strlen(p)];
```
wilzbach: Could use `toDString` from below.

src/dmd/dmodule.d
```diff
@@ -1337,19 +1337,25 @@ struct ModuleDeclaration
...
-        buf.writestring(id.toChars());
+        buf.writestring(id.toString());
         return buf.extractString();
     }
+
+    /// Provide a human readable representation
+    extern (D) const(char)[] toString() const
+    {
+        auto p = this.toChars();
+        return p[0 .. strlen(p)];
```
wilzbach: Could use `toDString` from below.

src/dmd/json.d
```diff
@@ -526,20 +532,20 @@ public:
...
             }
         }
-        stringPart(s.id.toChars());
+        stringPart(s.id.toString());
         stringEnd();
         comma();
-        property("kind", s.kind());
-        property("comment", s.comment);
+        property("kind", toDString(s.kind()));
+        property("comment", toDString(s.comment));
```
wilzbach: Nit: another advantage of making `toDString` a global function is that you can do the more natural: `s.comment.toDString()`

src/dmd/root/rootobject.d
```diff
@@ -62,6 +62,14 @@ extern (C++) class RootObject
         assert(0);
     }
 
+    ///
+    extern(D) const(char)[] toString()
+    {
+        import core.stdc.string : strlen;
+        auto p = this.toChars();
+        return p[0 .. strlen(p)];
```
wilzbach: Could use `toDString` from above.

src/dmd/json.d
Geod24: Didn't expect it to be so widely used :(
Also I'm wary that `dmd.utils` will be imported everywhere soon.
But let's give it a shot...

Geod24:
Addressed @wilzbach 's comments

src/dmd/root/rootobject.d
Geod24: Here I choose not to, as we want `rootobject` to depends on as little as possible.

JinShil (approved):
Still looking good.

Geod24:
Since this has gotten a bit more reviews, and one might be wondering where I am going, I added a bit more diff than originally intended:
- Commit 0c91769 moves `DArray` from `globals.h` to a standalone header, because we don't want `object.h` importing `globals.h`
- I reworked `RootObject.toString` to be callable from C++ and added a test.

As for the end goal is, [it's to be able to do this](https://github.com/Geod24/dmd/commit/1a740a2e98294e6d5430659d4ca358d5f05bbdda#diff-3ea076fe941194ddccb18fc686689e4eL148) gradually.

src/tests/cxxfrontend.c
```diff
@@ -212,6 +212,11 @@ void test_semantic()
     semantic2(m, NULL);
     semantic3(m, NULL);
 
+    // Basic test for D slice passing
+    DArray<const char> sl = m->toString();
```
jacob-carlborg: Does this work on Windows? I’ve been told D arrays are returned in registers while structs of this size are not on Windows.

src/dmd/utils.d
```diff
@@ -132,3 +132,9 @@ extern (C++) void escapePath(OutBuffer* buf, const(char)* fname)
         fname++;
     }
 }
+
+/// Slices a `\0`-terminated C-string, excluding the terminator
+const(char)[] toDString (const(char)* s) pure nothrow @nogc
```
jacob-carlborg: Why not `toString`?

Geod24: `toString` is to provide a string representation, this is to convert a `const(char)*` to a `const(char)[]`.

src/tests/cxxfrontend.c
Geod24: I should've known it was too good to be true.
I'll wait for the auto-tester to tell me what's the situation with Windows, but it fails on 32 bits.
So I guess I will take that test out for now :'(

Geod24: Hum, the C++ frontend tests only run on POSIX anyway...

Geod24:
@wilzbach : 
```
dmd/root/rootobject.d(66:29)[warn]: Methods 'opCmp', 'toHash', 'opEquals', 'opCast', and/or 'toString' are non-const.
```

What's the deal with that warning ?? `extern(C++)` classes don't inherit from `Object` so it doesn't make sense to require them to be `const`. And I currently cannot make it `const`.

src/tests/cxxfrontend.c
jacob-carlborg: For reference: https://github.com/dlang/dmd/pull/8120#discussion_r200835776.

wilzbach:
> What's the deal with that warning ?? extern(C++) classes don't inherit from Object so it doesn't make sense to require them to be const. 

Yeah that's the newly added Dscanner static code analysis check and the idea is that the checks slowly allow to modernize DMD / enforce some automatic code checking.
I guess the assumption behind this check was that "opEquals, opCmp, toHash, and toString are either const, immutable or inout" because they shouldn't be changing any data.

> And I currently cannot make it const.

tl;dr: in this case, add your module to this list:

https://github.com/dlang/dmd/blob/48726c875fbfe16fd93051570d149e893d2dc3dc/src/.dscanner.ini#L118

src/dmd/root/object.h
```diff
@@ -57,6 +58,9 @@ class RootObject
     virtual void print();
 
     virtual const char *toChars();
+    /// This function is `extern(D)` and should not be called from C++,
+    /// as the ABI does not match on some platforms
+    virtual DArray<const char> toString();
```
jacob-carlborg: Is it possible to implement this on the C++ and have the implementation assert?

Geod24:
@wilzbach : Thanks, glad to see there's a way out of it!

src/dmd/root/object.h
Geod24: I don't think so.

wilzbach:
@Geod24 I don't know exactly what's up with SemaphoreCI, but https://github.com/dlang/dmd/pull/8560 (once merged back to master) will hopefully help a bit.

Geod24:
Rebased on master, let's see.

JinShil (approved):
Yup, I still like it.

wilzbach (approved):
Yeah same here. One small step forward (:

Geod24:
Anything else ? Got a few changes depending on this PR (to be precise, on `toDString`).

wilzbach:
I think this is just pending/blocked on Ian's change request. @ibuclaw are you okay with moving forward here?

PR #8556, opened 2018-08-10
Remove opDollar from internal array because it's not used at all and doesn't work either

ghost:
…

### DIFF
--- a/src/dmd/root/array.d
+++ b/src/dmd/root/array.d
@@ -211,8 +211,6 @@ public:
         assert(a <= b && b <= dim);
         return data[a .. b];
     }
-
-    alias opDollar = dim;
 }
 
 struct BitArray


jacob-carlborg (changes requested):
Remove the code instead of commenting it out.

BBasile:
@jacob-carlborg ok done.

BBasile:
The problem of this alias is that it gives the illusion that `stuff[$-N]` works, which is not the case.

jacob-carlborg (approved):

n8sh:
For my edification can you explain why`alias opDollar = dim` doesn't work as expected?

BBasile:
Yes, if course @n8sh, a simple usage shows this pretty well : 

dollar_not_working

PR #8552, opened 2018-08-09
Fix Issue 19108 - Unknown pragmas not ignored inside body

JinShil:

PR #8557, opened 2018-08-12
Refactor pragma semantic

thewilsonator:
cc @ibuclaw @JinShil


src/dmd/pragmasem.d
```diff
@@ -0,0 +1,467 @@
...
+
+/**
+ * Semanically analyses a PragmaStatement.
+ * Params:
+ *  sc        = scope of evaluation
+ *  ps        = PragmaStatement to analyse
+ */
+extern(C++) Statement pragmaSemantic(Scope* sc, PragmaStatement ps)
+{
+    auto psa = PragmaSemanticAnalysis(sc,ps,null);
```
ibuclaw: Spaces.

src/dmd/pragmasem.d
```diff
@@ -0,0 +1,467 @@
...
+    return psa.result;
+}
+
+/**
+ * Semanically analyses a PragmaDeclaration.
+ * Params:
+ *  sc        = scope of evaluation
+ *  pd        = PragmaDeclaration to analyse
+ */
+extern(C++) void pragmaSemantic(Scope* sc, PragmaDeclaration pd)
```
ibuclaw: Do these need to be exported to C++?  I mean, we already have dsymbolSemantic and statementSemantic.

src/dmd/pragmasem.d
```diff
@@ -0,0 +1,472 @@
...
+            pragmaLibSemantic();
+        else if (ident == Id.startaddress)
+            pragmaStartAddressSemantic();
+        else if (ident == Id.Pinline)
+            pragmaInlineSemantic();
+        else if (ident == Id.mangle)
+            return pragmaMangleSemantic(); // does semantic on its decls
+        else if (ident == Id.crt_constructor || ident == Id.crt_destructor)
+            pragmaCrtCtorDtorSemantic(ident);
+        else if (!Target.pragmaSemantic(ident,sc,args,ps,pd))
```
ibuclaw: Spaces

src/dmd/pragmasem.d
```diff
@@ -0,0 +1,472 @@
...
+            Scope* sc2 = pd.newScope(sc);
+            foreach(s; *pd.decl)
+                s.dsymbolSemantic(sc2);
+
+            if (sc2 != sc)
+            sc2.pop();
+        }
+    }
+
+    void pragmaMsgSemantic()
```
ibuclaw: Maybe add a little documentation, or a link to the spec for each of these functions?

src/dmd/statementsem.d
```diff
@@ -2234,146 +2234,22 @@ else
...
         else if (!global.params.ignoreUnsupportedPragmas)
         {
-            ps.error("unrecognized `pragma(%s)`", ps.ident.toChars());
-            return setError();
-        }
-
-        if (ps._body)
-        {
-            ps._body = ps._body.statementSemantic(sc);
+            result = null;
```
ibuclaw: Are you sure this is correct? Shouldn't this condition block be removed?

src/dmd/target.d
```diff
@@ -774,6 +779,23 @@ struct Target
...
+     *  sc        = scope of evaluation
+     *  args      = arguments to pass to pragma
+     *  ps        = if != null analysing a PragmaStatement
+     *  pd        = if != null analysing a PragmaDeclaration
+     * Returns:
+     *  true if the pragma is a known pragma
+     *  false otherwise.
+     */
+    extern(C++) static bool pragmaSemantic(Identifier ident, Scope* sc, Expressions* args,
+                                           PragmaStatement ps, PragmaDeclaration pd)
```
ibuclaw: Rather than passing both kinds of pragma to the function, just have a boolean true/false like I mentioned in the issue?

src/dmd/pragmasem.d
```diff
@@ -0,0 +1,472 @@
...
+                if (ps)
+                    error(ps.loc, msg, ps.ident.toChars());
+                else
+                    error(pd.loc, msg, pd.ident.toChars());
+            }
+        }
+
+        if (pd && pd.decl)
+        {
+            pd.error("is missing a terminating `;`");
```
ibuclaw: The provider of the pragma should determine whether this is an error or not.

It should be moved to a separate function and called by `pragmaMsgSemantic` and others.

src/dmd/pragmasem.d
```diff
@@ -0,0 +1,472 @@
...
+        if (ident == Id.msg)
+            pragmaMsgSemantic();
+        else if (ident == Id.lib)
+            pragmaLibSemantic();
+        else if (ident == Id.startaddress)
+            pragmaStartAddressSemantic();
+        else if (ident == Id.Pinline)
+            pragmaInlineSemantic();
+        else if (ident == Id.mangle)
+            return pragmaMangleSemantic(); // does semantic on its decls
```
ibuclaw: This should not be needed if you move the the `missing ;` out of here.

ibuclaw (changes requested):
Nice, but needs some more work to make it usable by compilers implementing their own vendor pragmas.

ibuclaw:
Also, you should reference three other issues here. 3004, 19109, and 19110.

src/dmd/statementsem.d
thewilsonator: That was supposed to be unconditional. I think I messed up a bit trying to squash a conflicting merge (#8552).

src/dmd/target.d
thewilsonator: You can't issue errors for misuse if you don't have the declaration/statement.

src/dmd/pragmasem.d
thewilsonator: True.

src/dmd/target.d
ibuclaw: Hmm, could we do without one or two parameters then? For instance args is part of the pragma declaration/statement itself, no need to pass it twice I guess.

thewilsonator: We could, I have it as a local of `PragmaSemanticAnalysis` because it is so widely used. Looking now I see that LDC does not support (non DMD) PragmaStatements, what is GDC's usage like?

ibuclaw: I have a use case that isn't implemented yet.
```
pragma(GNU_unroll, 20)
for (int i = 0; i < m; i++)
    a[i] = a[i + k] * c;
```

WalterBright (changes requested):
Please do NOT combine large refactorings with bug fixes.

WalterBright:
Please do the bug fix and the refactorings as separate PRs, because 1. that makes them infinitely easier to review and 2. PRs should be one issue per PR.

JinShil:
@thewilsonator Do you plan on heading @WalterBright's advise here and split the refactor out into it's own PR?

thewilsonator:
Sorry I've been sick all week. I will, but If you want to get it done faster, I won't stop you from doing it.

JinShil:
No worries.  I'd rather you did it; you might need me to merge it, and I can't merge my own PRs.

thewilsonator:
Oh wonderful, Im corrupting something somewhere and object file emission segfaults. I guess this will have to wait until I'm not sick.

andralex (changes requested):
This kind of work needs to stop. There's no discernible purpose to this refactoring and no discernible improvement in the code following. It actually adds code!

thewilsonator:
On the contrary, this centralises pragma semantic to one file and make it much easier to be consistent between `PragmaDeclaration`s and `PragmaStatement`s. See 
https://issues.dlang.org/show_bug.cgi?id=19149 https://issues.dlang.org/show_bug.cgi?id=19109

JinShil:
@andralex, `expressionsem.d` is 11,000 lines of code.  I don't know if you spend much time in DMD's codebase, but I do, and it's drudgery navigating files of that size.  Furthermore, since modules are D's unit of encapsulation, there's little encapsulation and separation of concerns in the semantic logic.  That not only makes it difficult to see relationships between functions in _all the noise_, but it invites abuse as some functions cannot be privately restricted to just their intended callers.

> It actually adds code!
That's not necessarily a bad thing, especially if some of that additional code is documentation that didn't previously exist.

@thewilsonator, I don't know if we'll be able to move forward with this given @andralex's comment.  Though I would consider moving this forward if others voiced their support. I haven't actually looked much at the code though; not much point given @andralex's comment.

For now, I suggest focusing on bug fixes without the refactoring.  Also, I recommend not _reusing_ PRs when there is a significant change in its purpose; it tends to poison the conversation and review.

thewilsonator:
@JinShil I take Andrei's comment along the lines of his "Don't let bureaucracy stand in the way of progress" comment, i.e. misinformed at best. The whole point of this refactoring was to allow the bug fix to happen, either I put the bug fix with the PR and Walter tells me to break it apart, or Andrei tells me that the refactoring is useless. 

@ibuclaw sorry, you can fix your pragma issues on your own.

thewilsonator:
Well the comments  in the original that say `//Should be merged with PragmaStatement` and `//Should be merged with PragmaDeclaration` for a start. Having them separate makes is much easier to get out of sync e.g. w.r.t handling "unrecognised" pragmas.

See [issue 19109](https://issues.dlang.org/show_bug.cgi?id=19109),[19110](
https://issues.dlang.org/show_bug.cgi?id=19110) and others in [this list](https://issues.dlang.org/buglist.cgi?quicksearch=pragma&list_id=222897)

src/dmd/pragmasem.d
```diff
@@ -0,0 +1,489 @@
...
+    return psa.result;
+}
+
+/**
+ * Semanically analyses a PragmaDeclaration.
+ * Params:
+ *  sc        = scope of evaluation
+ *  pd        = PragmaDeclaration to analyse
+ */
+void pragmaSemantic(Scope* sc, PragmaDeclaration pd)
```
WalterBright: Note that this function returns a `void`, and the overload of it returns a `Statement`. This certainly looks peculiar, and they shouldn't be overloads if they are doing fundamentally different things.

thewilsonator: Thats because the statement is needed in the `Statement` semantic, but the declaration is not needed for `Declaration` semantic. Why, I don't know.

src/dmd/pragmasem.d
```diff
@@ -0,0 +1,489 @@
...
+            buf.writeByte(')');
+            message("pragma    %s", buf.peekString());
+        }
+        else
+        {
+            enum msg = "unrecognized `pragma(%s)`";
+            if (ps)
+                .error(ps.loc, msg, ps.ident.toChars());
+            else
+                .error(pd.loc, msg, pd.ident.toChars());
```
WalterBright: Not sure why a global `.error` call, given that above there's a custom `error`.

thewilsonator: Good point, I think I was confused about the `loc` parameter. It should fit the custom one.

src/dmd/pragmasem.d
```diff
@@ -0,0 +1,489 @@
...
+ *  sc        = scope of evaluation
+ *  pd        = PragmaDeclaration to analyse
+ */
+void pragmaSemantic(Scope* sc, PragmaDeclaration pd)
+{
+    auto psa = PragmaSemanticAnalysis(sc, null, pd);
+    psa.semantic(pd.ident);
+}
+
+private struct PragmaSemanticAnalysis
```
WalterBright: The purpose of this struct appears to be to simply group together some related functions. Isn't that what modules are for, and the whole purpose of creating this file?

src/dmd/pragmasem.d
WalterBright: Because of the distinctly different interfaces, they should have different names, not be overloads.

Overloads are useful for writing generic code, this isn't generic.

ibuclaw:
> How does it enable fixing @ibuclaw 's pragma issues?

Decentralise handling of D pragmas by enabling compiler vendors to add their own pragmas without modifying the dmd code.

The crux of it comes down to asking "Do you support this pragma?" before issuing an error.

I don't think you *need* to do a large refactoring in order to make this happen, but code to support it will be identical for both pragma declarations and statements.

@thewilsonator same as in your other patch for function semantic. Probably best to move these one at a time, and preferably to free functions so you aren't relying on some internal state.

ibuclaw:
> Sorry if I'm missing context or being hasty folks. What I saw in this PR was a reshuffling of the code that effects no visible improvement. The runes were in some position, they get thrown in the air, they land in a different position. As a litmus test, if the refactoring went the other way, would it be seen as an obvious problem?

@andralex - Under "Modernize DMD"

https://github.com/dlang/dmd/projects

In particular, this point applies:

> Reduce cyclomatic complexity 

I also take that to mean: Moving functionality into leaf modules in order to reduce cyclic dependencies of the compiler implementation.

And if it doesn't, then maybe we should add a new bullet to make that clearer.

An example, modifying `dmd.argtypes` will only trigger a rebuild of `dmd.target`, and modifying `dmd.target` will leave `dmd.argtypes` alone.  I consider this as good encapsulation of a module, however, eyes of the beholder may apply.

andralex:
@ibuclaw totally cool with the idea, and we should avoid entering a pattern whereby the slightest challenge is immediately met with abandoning the pull request.

"Reducing cyclomatic complexity" is a bit vacuous and should probably be changed because we currently have no way of measuring it. (It's more related to control flow paths than module dependencies though.)

Anyhow, unless a refactoring has obvious overwhelming advantages it should not just go through with just the "Refactoring Xxx" title and on the basis it passes tests. The code has a number of issues that do warrant a close review. 

For example, the best thing about this PR is unifying two large horrendous functions. One is `void visit(PragmaDeclaration pd)`, which opens with the comment `// Should be merged with PragmaStatement`, and of course the other corresponding `void visit(PragmaStatement pd)`, which too, opens with the comment `// Should be merged with PragmaDeclaration`. The functions have 230 lines and 142 lines, respectively. Merging them properly together would be an obvious win. (I'm not sure if the right way is to have essentially two references `PragmaStatement ps` and `PragmaDeclaration pd` of which only one can be non-null, and then choose between them at runtime.) At any rate, a good refactor would do this kind of merge and be able to show a clear improvement, and would be the basis of any further work. @thewilsonator try that one?

PR #8561, opened 2018-08-13
final_attribute_check: remove useless usage of the final attribute

wilzbach:
A first and simple example on how DScanner can be incrementally used to upgrade DMD's codebase.

The more advanced checks are, of course, more work, though.
However, with `ModuleFilters` list, it can be done on a module-by-module basis.


JinShil:
Forgive my ignorance, but why are these useless?  I thought D's methods were virtual-by-default.

wilzbach:
They are all on `final` classes or `private` methods. Maybe the check should use the word "redundant" instead.

JinShil (approved):

jacob-carlborg (approved):

PR #8565, opened 2018-08-14
Test with gdc-7 and gdc-8

wilzbach:
See also:
- https://github.com/D-Programming-GDC/GDC/pull/704#issuecomment-410694849
- https://github.com/dlang/dmd/pull/8540 (recent removal of in contracts in DMD for GDC-7 support)

CC @ibuclaw

I already configured another gdc-7 job at SemaphoreCI.
With the free plan we can only run four jobs in parallel, but this seems to be important to the GDC project and also the long-term plan is to migrate all SemaphoreCI tests to Buildkite runners (e.g. https://github.com/dlang/ci/pull/261) where we can just arbitrarily (and cheaply!) add new runners (we already have ten agents there).

I know that this way of hacking with the Ubuntu PPA isn't ideal, but as long as the install.sh doesn't support newer GDC binaries, this is an easy workaround.

### DIFF
--- a/ci.sh
+++ b/ci.sh
@@ -158,17 +158,17 @@ download_install_sh() {
 
 install_d() {
   if [ "${DMD:-dmd}" == "gdc" ] || [ "${DMD:-dmd}" == "gdmd" ] ; then
-    export DMD=gdmd-8
-    if [ ! -e ~/dlang/gdc-8/activate ] ; then
+    export DMD=gdmd-${GDC_VERSION}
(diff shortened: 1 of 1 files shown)


PetarKirov (approved):

ibuclaw:
To clarify my stance, I think it is a reasonable request to ensure that the last two releases of gdc work with compiling dmd.  When gdc-9 is out, you can drop gdc-7.

PR #8566, opened 2018-08-15
Add temp directory detection to build.d for Windows

JinShil:


jacob-carlborg (approved):

src/build.d
```diff
@@ -479,7 +479,7 @@ void parseEnvironment()
 
     env.getDefault("GIT_HOME", "https://github.com/dlang");
     env.getDefault("SYSCONFDIR", "/etc");
-    env.getDefault("TMP", "/tmp");
+    env.getDefault("TMP", detectTemp);
```
ZombineDev: How about: https://dlang.org/phobos/std_file#tempDir ?

jacob-carlborg: Good point, we’re already using Phobos.

PetarKirov (approved):

wilzbach:
SemaphoreCI didn't get triggered, but `build.d` isn't run there anyhow (for now at least.)

PR #8568, opened 2018-08-15
build.d: Add ability to determine the host C++ compiler on Windows

JinShil:


src/build.d
```diff
@@ -820,6 +821,29 @@ auto detectModel()
...
+{
+    version(Posix)
+        return "c++";
+    else version(Windows)
+    {
+        immutable model = detectModel;
+        if (model == "32")
+            return "dmc";
+        else if (model == "64")
+            return `vcbuild\msvc-dmc`;
```
JinShil: See https://github.com/dlang/dmd/blob/fd66f64bf8f8720b5248a8c3f836c899d55e4676/src/win64.mak#L24

`msvc-dmc` can be found in https://github.com/dlang/dmd/tree/master/src/vcbuild

PetarKirov (approved):

src/build.d
JinShil: My understanding is that for 32-bit code, the makefiles us DMC:  https://github.com/dlang/dmd/blob/fd66f64bf8f8720b5248a8c3f836c899d55e4676/src/win32.mak#L88

wilzbach: I think you need to make sure that this file is compiled.

```d
Dependency dependency = {
        target: env["G"].buildPath("msvc-dmc").exeName,
        sources: ["vcbuild\msvc-dmc"],
};
dependency.run();
```

(though if you put this in `$G`, make sure `mkdirRecurse(g);` (also part of `parseEnvironment` is called first)).

JinShil: If you're implying that we should be using MSVC for both 32 and 64-bit code, well, I'm trying not to rock the boat.

JinShil: > I think you need to make sure that this file is compiled.

Yes, I'm just working through the script step by step.  I'll get there.

wilzbach: And looking a bit more at the win64.mak file, it seems like it also requires to build the `vcbuild\msvc-lib` file

src/build.d
```diff
@@ -236,6 +236,19 @@ auto opTabGen()
...
+version(Windows)
+{
+    // Build the msvc-dmc compiler wrapper
+    void buildMsvcDmc()
+    {
+        Dependency dependency = {
+            target: env["G"].buildPath("msvc-dmc").exeName,
+            sources: [`vcbuild\msvc-dmc`],
+        };
+        dependency.run();
```
wilzbach: `return dependency.target`

src/build.d
```diff
@@ -836,7 +849,10 @@ auto getHostCXX()
         if (model == "32")
             return "dmc";
         else if (model == "64")
+        {
+            buildMsvcDmc();
             return `vcbuild\msvc-dmc`;
```
wilzbach: `return buildMsvcDmc()` (otherwise you won't use the built msvc-dmc.exe)

src/build.d
JinShil: Added `msvc-lib`, but they're not building yet, that's a bigger task than I intended to tackle today.

src/build.d
JinShil: I added `msvc-dmd` and `msvc-lib` dependencies for `cxxBackend`.  I think I need `processEnvironment` to run first before those can be built.

src/build.d
```diff
@@ -270,6 +293,15 @@ auto dBackend()
 auto cxxBackend()
 {
     Dependency[] dependencys;
```
wilzbach: `dependencies` - sorry for the typo :O

src/build.d
wilzbach: Okay, but I think you still need sth. like `buildMsvcDmc().target` here to return the correct path.

src/build.d
JinShil: I will fix that later.  There are a few places it needs to change and I'd like to keep the diff low.

wilzbach (approved):

PR #8569, opened 2018-08-15
Fix misspellings in build.d

JinShil:


wilzbach (approved):
Oh I think this is because initially `Dependency` was named `Rule` and I just did a dumb Replace/All at the end.
Thanks for fixing this!

wilzbach:
(This build script run only on CircleCi which has already passed.)

PR #8570, opened 2018-08-16
build.d: Set CXX_KIND variable for Windows

JinShil:


src/build.d
```diff
@@ -876,6 +875,24 @@ auto getHostCXX()
...
+*/
+auto getHostCXXKind()
+{
+    version(Posix)
+    {
+        auto cxxVersion = execute([getHostCXX, "--version"]).output;
+        return !cxxVersion.find("gcc", "Free Software")[0].empty ? "g++" : "clang++";
+    }
+    else version(Windows)
+        return "dmc";
```
wilzbach: Couldn't this be MVSC too?

JinShil: I say no, because `msvc-dmc` is a wrapper around MSVC to make it behave like DMC, so for all intents and purposes we're dealing with DMC.

PetarKirov (approved):

PR #8572, opened 2018-08-17
Refactor and document ensurePathExists

Geod24:
```
- Use early returns
- add DDOC and comments
- return `false` on failure, not `true`
```

### DIFF
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -631,71 +631,72 @@ nothrow:
         }
     }
 
+    /**
+       Ensure that the provided path exists
+
+       Accepts a path to either a file or a directory.
+       In the former case, the basepath (path to the containing directory)
+       will be checked for existence, and created if it does not exists.
+       In the later case, the directory pointed to will be checked for existence
+       and created if needed.
+
+       Params:
+         path = a path to a file or a directory
+
+       Returns:
+         `true` if the directory exists or was successfully created
+     */
     extern (C++) static bool ensurePathExists(const(char)* path)
     {
         //printf("FileName::ensurePathExists(%s)\n", path ? path : "");
-        if (path && *path)
+        if (!path || !(*path))
+            return true;
+        if (exists(path))
+            return true;
+
+        // We were provided with a file name
+        // We need to call ourselves recursively to ensure parent dir exist
+        const(char)* p = FileName.path(path);
+        if (*p)
         {
-            if (!exists(path))
+            version (Windows)
             {
-                const(char)* p = FileName.path(path);
-                if (*p)
+                const len = strlen(path);
+                const plen = strlen(p);
+                if (len == plen ||
+                    (len > 2 && path[1] == ':' && path[2 .. len] == p[0 .. plen]))
                 {
-                    version (Windows)
-                    {
-                        size_t len = strlen(path);
-                        if ((len > 2 && p[-1] == ':' && strcmp(path + 2, p) == 0) || len == strlen(p))
-                        {
-                            mem.xfree(cast(void*)p);
-                            return 0;
-                        }
-                    }
-                    bool r = ensurePathExists(p);
                     mem.xfree(cast(void*)p);
-
-                    if (r)
-                        return r;
-                }
-                version (Windows)
-                {
-                    char sep = '\\';
-                }
-                else version (Posix)
-                {
-                    char sep = '/';
-                }
-                if (path[strlen(path) - 1] != sep)
-                {
-                    version (Windows)
-                    {
-                        int r = _mkdir(path);
-                    }
-                    version (Posix)
-                    {
-                        int r = mkdir(path, (7 << 6) | (7 << 3) | 7);
-                    }
-                    if (r)
-                    {
-                        /* Don't error out if another instance of dmd just created
-                         * this directory
-                         */
-                        version (Windows)
-                        {
-                            // see core.sys.windows.winerror - the reason it's not imported here is because
-                            // the autotester's dmd is too old and doesn't have that module
-                            enum ERROR_ALREADY_EXISTS = 183;
-
-                            if (GetLastError() != ERROR_ALREADY_EXISTS)
-                                return true;
-                        }
-                        version (Posix)
-                        {
-                            if (errno != EEXIST)
-                                return true;
-                        }
-                    }
+                    return true;
                 }
             }
+            const r = ensurePathExists(p);
+            mem.xfree(cast(void*)p);
+
+            if (!r)
+                return r;
+        }
+
+        version (Windows)
+            const r = _mkdir(path);
+        version (Posix)
+            const r = mkdir(path, (7 << 6) | (7 << 3) | 7);
+
+        if (r == 0)
+            return true;
+
+        // Don't error out if another instance of dmd just created
+        // this directory
+        version (Windows)
+        {
+            import core.sys.windows.winerror : ERROR_ALREADY_EXISTS;
+            if (GetLastError() == ERROR_ALREADY_EXISTS)
+                return true;
+        }
+        version (Posix)
+        {
+            if (errno == EEXIST)
+                return true;
         }
 
         return false;
--- a/src/dmd/utils.d
+++ b/src/dmd/utils.d
@@ -95,7 +95,7 @@ extern (C++) void ensurePathToNameExists(Loc loc, const(char)* name)
     const(char)* pt = FileName.path(name);
     if (*pt)
     {
-        if (FileName.ensurePathExists(pt))
+        if (!FileName.ensurePathExists(pt))
         {
             error(loc, "cannot create directory %s", pt);
             fatal();


src/dmd/root/filename.d
```diff
@@ -631,71 +631,75 @@ nothrow:
...
+        version (Windows)
+            int r = _mkdir(path);
+        version (Posix)
+            int r = mkdir(path, (7 << 6) | (7 << 3) | 7);
+
+        if (r == 0)
+            return true;
+
+        /* Don't error out if another instance of dmd just created
+         * this directory
```
JinShil: While you're at it, remove the '*' prefix.

src/dmd/root/filename.d
```diff
@@ -631,71 +631,75 @@ nothrow:
...
+       In the former case, the `basepath` (path to the containing directory)
+       will be checked for existence, and created if it does not exists.
+       In the later case, the directory pointed to will be checked for existence
+       and created if needed.
+
+       Params:
+         path = a path to a file or a directory
+
+       Returns:
+         `true` if the directory exists / was successfully created,
```
JinShil: I would prefer a proper "or" rather than a "/".

src/dmd/root/filename.d
```diff
@@ -631,71 +631,75 @@ nothrow:
...
+        // We need to call ourselves recursively to ensure parent dir exist
+        const(char)* p = FileName.path(path);
+        if (*p)
         {
-            if (!exists(path))
+            version (Windows)
             {
-                const(char)* p = FileName.path(path);
-                if (*p)
+                size_t len = strlen(path);
```
JinShil: I though `strlen` was supposed to die.  Wile you're at it, can't you change the interface to accept a `string` or other `char[]`.

Geod24: I have a WIP which converts everything in filename to use slices, which is how I came across this code. But it's currently blocked on my JSON pr so I decided to submit this refactoring first.

Geod24:
```diff
diff --git a/src/dmd/root/filename.d b/src/dmd/root/filename.d
index 800f0673c..75b594674 100644
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -644,7 +644,7 @@ nothrow:
          path = a path to a file or a directory

        Returns:
-         `true` if the directory exists / was successfully created,
+         `true` if the directory exists or was successfully created,
          `false` otherwise
      */
     extern (C++) static bool ensurePathExists(const(char)* path)
@@ -684,9 +684,8 @@ nothrow:
         if (r == 0)
             return true;

-        /* Don't error out if another instance of dmd just created
-         * this directory
-         */
+        // Don't error out if another instance of dmd just created
+        // this directory
         version (Windows)
         {
             // see core.sys.windows.winerror - the reason it's not imported here is because
```

Updated

JinShil (approved):

src/dmd/root/filename.d
```diff
@@ -631,71 +631,74 @@ nothrow:
...
+
+       Accepts a path to either a file or a directory.
+       In the former case, the `basepath` (path to the containing directory)
+       will be checked for existence, and created if it does not exists.
+       In the later case, the directory pointed to will be checked for existence
+       and created if needed.
+
+       Params:
+         path = a path to a file or a directory
+
```
WalterBright: All these blank lines detract a bit from readability, as one has less code that fits on the screen.

src/dmd/root/filename.d
```diff
@@ -631,71 +631,74 @@ nothrow:
...
+       Returns:
+         `true` if the directory exists or was successfully created,
+         `false` otherwise
```
WalterBright: This second line is redundant.

src/dmd/root/filename.d
```diff
@@ -631,71 +631,74 @@ nothrow:
...
     }
 
+    /**
+       Ensure that the provided path exists
+
+       Accepts a path to either a file or a directory.
+       In the former case, the `basepath` (path to the containing directory)
+       will be checked for existence, and created if it does not exists.
+       In the later case, the directory pointed to will be checked for existence
+       and created if needed.
```
WalterBright: It has nothing to do with whether the path leads to a file or a directory, as that has no effect on the behavior of the function. This detail caused me to spend some time looking at the function implementation, because adding the detail implies it has relevance. Perhaps a better description would be simply:
```
If `path` does not exist, create it.
```

src/dmd/root/filename.d
```diff
@@ -631,71 +631,74 @@ nothrow:
...
+        const(char)* p = FileName.path(path);
+        if (*p)
         {
-            if (!exists(path))
+            version (Windows)
             {
-                const(char)* p = FileName.path(path);
-                if (*p)
+                size_t len = strlen(path);
+                if ((len > 2 && p[-1] == ':' && strcmp(path + 2, p) == 0) || len == strlen(p))
```
WalterBright: The `strcmp` is a giant bug here, because Windows filenames are case insensitive. Since this PR is a refactor, do not fix it here, but in a followup PR would be nice.

src/dmd/root/filename.d
```diff
@@ -631,71 +631,74 @@ nothrow:
...
             }
+            bool r = ensurePathExists(p);
+            mem.xfree(cast(void*)p);
+
+            if (!r)
+                return r;
+        }
+
+        version (Windows)
+            int r = _mkdir(path);
```
WalterBright: r should be const.

src/dmd/root/filename.d
```diff
@@ -631,71 +631,74 @@ nothrow:
...
-                        version (Posix)
-                        {
-                            if (errno != EEXIST)
-                                return true;
-                        }
-                    }
+                    return true;
                 }
             }
+            bool r = ensurePathExists(p);
```
WalterBright: r should be const.

src/dmd/root/filename.d
```diff
@@ -631,71 +631,74 @@ nothrow:
...
+            // see core.sys.windows.winerror - the reason it's not imported here is because
+            // the autotester's dmd is too old and doesn't have that module
+            enum ERROR_ALREADY_EXISTS = 183;
+
+            if (GetLastError() == ERROR_ALREADY_EXISTS)
+                return true;
+        }
+        version (Posix)
+        {
+            if (errno == EEXIST)
```
WalterBright: Another bug. Any test of errno must be preceded by setting errno to 0.

WalterBright: Please fix in followup PR.

src/dmd/root/filename.d
```diff
@@ -631,71 +631,74 @@ nothrow:
...
+            int r = mkdir(path, (7 << 6) | (7 << 3) | 7);
+
+        if (r == 0)
+            return true;
+
+        // Don't error out if another instance of dmd just created
+        // this directory
+        version (Windows)
+        {
+            // see core.sys.windows.winerror - the reason it's not imported here is because
```
WalterBright: Now that we have github, core.sys.windows.winerror can be productively replaced with https://github.com/dlang/druntime/blob/master/src/core/sys/windows/winerror.d#L160

src/dmd/root/filename.d
JinShil: Please keep the newlines, it's consistent with coding conventions used in other dlang repositories, and it prevents the documentation from becoming a blob of text that is difficult to visually parse and glance through.

Geod24:
Updated:

```diff
diff --git a/src/dmd/root/filename.d b/src/dmd/root/filename.d
index 75b594674..7d156e6c6 100644
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -644,8 +644,7 @@ nothrow:
          path = a path to a file or a directory

        Returns:
-         `true` if the directory exists or was successfully created,
-         `false` otherwise
+         `true` if the directory exists or was successfully created
      */
     extern (C++) static bool ensurePathExists(const(char)* path)
     {
@@ -662,14 +661,14 @@ nothrow:
         {
             version (Windows)
             {
-                size_t len = strlen(path);
+                const len = strlen(path);
                 if ((len > 2 && p[-1] == ':' && strcmp(path + 2, p) == 0) || len == strlen(p))
                 {
                     mem.xfree(cast(void*)p);
                     return true;
                 }
             }
-            bool r = ensurePathExists(p);
+            const r = ensurePathExists(p);
             mem.xfree(cast(void*)p);

             if (!r)
@@ -677,9 +676,9 @@ nothrow:
         }

         version (Windows)
-            int r = _mkdir(path);
+            const r = _mkdir(path);
         version (Posix)
-            int r = mkdir(path, (7 << 6) | (7 << 3) | 7);
+            const r = mkdir(path, (7 << 6) | (7 << 3) | 7);

         if (r == 0)
             return true;
@@ -688,10 +687,7 @@ nothrow:
         // this directory
         version (Windows)
         {
-            // see core.sys.windows.winerror - the reason it's not imported here is because
-            // the autotester's dmd is too old and doesn't have that module
-            enum ERROR_ALREADY_EXISTS = 183;
-
+            import core.sys.windows.winerror : ERROR_ALREADY_EXISTS;
             if (GetLastError() == ERROR_ALREADY_EXISTS)
                 return true;
         }
```

I don't know how to address the comment about blank line, as blank lines before a new section is the prevalent style so far (e.g. look at the example [here](https://dlang.org/dstyle.html#phobos_documentation)).

src/dmd/root/filename.d
jacob-carlborg: Technically it could throw which would mean it wouldn't return at all.

Geod24:
Anything else ?

src/dmd/root/filename.d
```diff
@@ -631,71 +631,70 @@ nothrow:
         }
     }
 
+    /**
+       Ensure that the provided path exists
+
+       Accepts a path to either a file or a directory.
+       In the former case, the `basepath` (path to the containing directory)
```
wilzbach: Hmm this could be confusing to people as typically only variables are backticked and `basepath` isn't one. Maybe simply remove the backticks here?

src/dmd/root/filename.d
```diff
@@ -631,71 +631,70 @@ nothrow:
...
+        const(char)* p = FileName.path(path);
+        if (*p)
         {
-            if (!exists(path))
+            version (Windows)
             {
-                const(char)* p = FileName.path(path);
-                if (*p)
+                const len = strlen(path);
+                if ((len > 2 && p[-1] == ':' && strcmp(path + 2, p) == 0) || len == strlen(p))
```
wilzbach: `&& path[2 .. len] == p[0 .. strlen(p)]`

Geod24: Oh I misread. Yeah this part makes sense 👍

Geod24:
Updated:
```diff
diff --git a/src/dmd/root/filename.d b/src/dmd/root/filename.d
index f39f96335..a9ac20a68 100644
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -635,7 +635,7 @@ nothrow:
        Ensure that the provided path exists

        Accepts a path to either a file or a directory.
-       In the former case, the `basepath` (path to the containing directory)
+       In the former case, the basepath (path to the containing directory)
        will be checked for existence, and created if it does not exists.
        In the later case, the directory pointed to will be checked for existence
        and created if needed.
@@ -662,7 +662,8 @@ nothrow:
             version (Windows)
             {
                 const len = strlen(path);
-                if ((len > 2 && p[-1] == ':' && strcmp(path + 2, p) == 0) || len == strlen(p))
+                if ((len > 2 && p[-1] == ':' && path[2 .. len] == p[0 .. strlen(p)])
+                     || len == strlen(p))
                 {
                     mem.xfree(cast(void*)p);
                     return true;
```

src/dmd/root/filename.d
```diff
@@ -631,71 +631,71 @@ nothrow:
...
+        if (*p)
         {
-            if (!exists(path))
+            version (Windows)
             {
-                const(char)* p = FileName.path(path);
-                if (*p)
+                const len = strlen(path);
+                if ((len > 2 && p[-1] == ':' && path[2 .. len] == p[0 .. strlen(p)])
+                     || len == strlen(p))
```
WalterBright: Note that `strlen(p)` appears twice. The `p[-1]` is a buffer overflow bug because `Filename.path()` does not guarantee there is anything there.

Geod24:
Updated:
```diff
diff --git a/src/dmd/root/filename.d b/src/dmd/root/filename.d
index 187fbd6c7..1966f0255 100644
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -662,8 +662,9 @@ nothrow:
             version (Windows)
             {
                 const len = strlen(path);
-                if ((len > 2 && p[-1] == ':' && path[2 .. len] == p[0 .. strlen(p)])
-                     || len == strlen(p))
+                const plen = strlen(p);
+                if (len == plen ||
+                    (len > 2 && path[1] == ':' && path[2 .. len] == p[0 .. plen]))
                 {
                     mem.xfree(cast(void*)p);
                     return true;
```

Geod24:
Ping ? The has been quite heavily reviewed already...

JinShil (approved):

PR #8574, opened 2018-08-18
Fix Issue 19176 - Dmd crashes because of __traits(getUnitTests)

JinShil:


jacob-carlborg (approved):

clYd3r:
I think that this fix hadn't targeted the real issue since it remained some cases of crashes. https://github.com/dlang/dmd/pull/10533 uses another approach

PR #8575, opened 2018-08-18
Turn globals.argv0 into a DArray

Geod24:
This is a step towards converting `globals` to use D array.

There is currently no way to call a function accepting a D string from C++ (thanks to ABI issues on i386 / Win64) but it does not matter for `globals` AFAICS.
So I resorted to providing trivial C++ wrapper to the D functions, and gradually removing the wrapper as they become unused, as can be seen in filename.

### DIFF
--- a/src/dmd/frontend.d
+++ b/src/dmd/frontend.d
@@ -79,11 +79,7 @@ string findDMDConfig(const(char)[] dmdFilePath)
     import dmd.dinifile : findConfFile;
     import std.string : fromStringz, toStringz;
 
-    auto f = findConfFile(dmdFilePath.toStringz, "dmd.conf");
-    if (f is null)
-        return null;
-
-    return f.fromStringz.idup;
+    return findConfFile(dmdFilePath, "dmd.conf").idup;
 }
 
 /**
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -1056,8 +1056,8 @@ version (Windows)
 
             // search PATH to avoid createProcess preferring "link.exe" from the dmd folder
             Strings* paths = FileName.splitPath(getenv("PATH"));
-            if (auto p = FileName.searchPath(paths, "link.exe", false))
-                return p;
+            if (auto p = FileName.searchPath(paths, "link.exe"[], false))
+                return p.ptr;
             return "link.exe";
         }
 
@@ -1359,8 +1359,8 @@ version (Windows)
 
             // try mingw fallback relative to phobos library folder that's part of LIB
             Strings* libpaths = FileName.splitPath(getenv("LIB"));
-            if (auto p = FileName.searchPath(libpaths, r"mingw\kernel32.lib", false))
-                return FileName.path(p);
+            if (auto p = FileName.searchPath(libpaths, r"mingw\kernel32.lib"[], false))
+                return FileName.path(p).ptr;
 
             return null;
         }
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -824,16 +824,15 @@ version (Windows)
 {
     private int executearg0(const(char)* cmd, const(char)* args)
     {
-        const(char)* file;
-        const(char)* argv0 = global.params.argv0;
+        const argv0 = global.params.argv0;
         //printf("argv0='%s', cmd='%s', args='%s'\n",argv0,cmd,args);
         // If cmd is fully qualified, we don't do this
         if (FileName.absolute(cmd))
             return -1;
-        file = FileName.replaceName(argv0, cmd);
+        const file = FileName.replaceName(argv0, cmd.toDString);
         //printf("spawning '%s'\n",file);
         // spawnlp returns intptr_t in some systems, not int
-        return spawnl(0, file, file, args, null);
+        return spawnl(0, file.ptr, file.ptr, args, null);
     }
 }
 
@@ -1049,9 +1048,9 @@ version (Windows)
             char[MAX_PATH + 1] dmdpath = void;
             if (GetModuleFileNameA(null, dmdpath.ptr, dmdpath.length) <= MAX_PATH)
             {
-                auto lldpath = FileName.replaceName(dmdpath.ptr, "lld-link.exe");
+                auto lldpath = FileName.replaceName(dmdpath, "lld-link.exe");
                 if (FileName.exists(lldpath))
-                    return lldpath;
+                    return lldpath.ptr;
             }
 
             // search PATH to avoid createProcess preferring "link.exe" from the dmd folder
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -218,7 +218,7 @@ private int tryMain(size_t argc, const(char)** argv)
     //for (size_t i = 0; i < arguments.dim; ++i) printf("arguments[%d] = '%s'\n", i, arguments[i]);
     files.reserve(arguments.dim - 1);
     // Set default values
-    global.params.argv0 = arguments[0];
+    global.params.argv0 = arguments[0].toDString;
 
     // Temporary: Use 32 bits as the default on Windows, for config parsing
     static if (TARGET.Windows)
@@ -235,11 +235,11 @@ private int tryMain(size_t argc, const(char)** argv)
     {
         version (Windows)
         {
-            global.inifilename = findConfFile(global.params.argv0, "sc.ini");
+            global.inifilename = findConfFile(global.params.argv0, "sc.ini").ptr;
         }
         else version (Posix)
         {
-            global.inifilename = findConfFile(global.params.argv0, "dmd.conf");
+            global.inifilename = findConfFile(global.params.argv0, "dmd.conf").ptr;
         }
         else
         {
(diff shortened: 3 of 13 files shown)


src/dmd/mars.d
```diff
@@ -1418,7 +1415,7 @@ private void printPredefinedVersions(FILE* stream)
 
 extern(C) void printGlobalConfigs(FILE* stream)
 {
-    stream.fprintf("binary    %s\n", global.params.argv0);
+    stream.fprintf("binary    %*s\n", global.params.argv0.length, global.params.argv0.ptr);
```
ibuclaw: `%.*s` - if memory serves me right.

ibuclaw (commented):
Is github loosing my reviews?

ibuclaw:
> Is github loosing my reviews?
Answer, yes. This is twice now I've written something.

JinShil:
> Is github loosing my reviews?

I experienced that yesterday as well; started questioning my sanity.

src/dmd/frontend.d
```diff
@@ -79,7 +79,7 @@ string findDMDConfig(string dmdFilePath)
     import dmd.dinifile : findConfFile;
     import std.string : fromStringz, toStringz;
 
-    auto f = findConfFile(dmdFilePath.toStringz, "dmd.conf");
+    auto f = findConfFile(dmdFilePath.toStringz[0 .. dmdFilePath.length], "dmd.conf");
```
jacob-carlborg: This is converted to a C string then back to a D string. Am I missing something?

Geod24: It was a leftover. `findConfFile` was still dependent on `\0` termination, as one of the function was not converted to `const(char)[]`. I converted it and now `\0` termination isn't required anymore.

WalterBright:
The subject has nothing to do with the code. globals.argv0 is never mentioned in any of the diffs.

src/dmd/dmodule.d
```diff
@@ -943,7 +943,7 @@ extern (C++) final class Module : Package
             assert(prev);
             if (Module mprev = prev.isModule())
             {
-                if (FileName.compare(srcname, mprev.srcfile.toChars()) != 0)
+                if (!FileName.equals(srcname, mprev.srcfile.toChars()))
```
WalterBright: I don't understand. What is being gained here?

src/dmd/root/filename.d
```diff
@@ -49,30 +50,25 @@ nothrow:
...
-    {
         version (Windows)
         {
-            return stricmp(name1, name2);
+            return _strnicmp(name1.ptr, name2.ptr, name1.length) == 0;
         }
         else
         {
-            return strcmp(name1, name2);
+            return strncmp(name1.ptr, name2.ptr, name1.length) == 0;
```
WalterBright: strncmp is the wrong answer here. I quote the C Standard: "The strncmp function compares not more than n characters (characters that follow a null character are not compared) from the array pointed to by s1 to the array pointed to bys 2." I.e. embedded 0's are still a factor. Use memcmp() and memicmp() instead. I've never found a legitimate use for strncmp().

src/dmd/root/filename.d
```diff
@@ -49,30 +50,25 @@ nothrow:
...
-    extern (C++) bool equals(const RootObject obj) const pure
-    {
-        return compare(obj) == 0;
-    }
-
+    /// Compare two name according to the platform's rules (case sensitive or not)
     extern (C++) static bool equals(const(char)* name1, const(char)* name2) pure
     {
-        return compare(name1, name2) == 0;
+        return equals(name1.toDString, name2.toDString);
```
WalterBright: Please use ( ) when calling functions.

src/dmd/utils.d
```diff
@@ -138,3 +139,44 @@ const(char)[] toDString (const(char)* s) pure nothrow @nogc
...
+   Returns:
+     The return value of `T`
+*/
+auto toCStringThen(alias dg)(const(char)[] src)
+{
+    char[512] small = void;
+    scope ptr = (src.length < (small.length - 1))
+                    ? small[0 .. src.length + 1]
+                    : (cast(char*)mem.xmalloc(src.length + 1))[0 .. src.length + 1];
+    scope (exit)
```
WalterBright: Using scope in dmd is problematic - take a look at the code generated for it, it isn't pretty. It's a lot better if the function is marked nothrow.

src/dmd/utils.d
```diff
@@ -138,3 +139,44 @@ const(char)[] toDString (const(char)* s) pure nothrow @nogc
...
+     one has to copy it.
+
+   Params:
+     src = Slice to use to call the C function
+     dg  = Delegate to call afterwards
+
+   Returns:
+     The return value of `T`
+*/
+auto toCStringThen(alias dg)(const(char)[] src)
```
WalterBright: Why add functions that are never called?

Geod24:
@WalterBright : There is a bug with this PR I cannot reproduce locally. So I started to bisect it on the auto-tester. This is not ready for review, and I'll modify the title accordingly.

WalterBright:
Ok. But the size of this is getting large. Please consider more bite size chunks.

Geod24:
> Ok. But the size of this is getting large. Please consider more bite size chunks.

Fair enough, I've took some of those commits to https://github.com/dlang/dmd/pull/8585

Geod24:
This is working now, but based on #8585

Geod24:
@klickverbot : Isn't `DArray` enough ? E.g. can't you access `globals.argv0.{ptr,length}` ?
I should have added a test to the C++ interop.

Geod24:
@wilzbach Any idea what's up with that linker failure ? I see it's during coverage, but I'm not sure whether those two files should be compiled together....

Geod24:
This finally passes and is ready for review

src/dmd/frontend.d
```diff
@@ -79,11 +79,11 @@ string findDMDConfig(const(char)[] dmdFilePath)
     import dmd.dinifile : findConfFile;
     import std.string : fromStringz, toStringz;
 
-    auto f = findConfFile(dmdFilePath.toStringz, "dmd.conf");
+    auto f = findConfFile(dmdFilePath, "dmd.conf");
     if (f is null)
         return null;
 
-    return f.fromStringz.idup;
+    return f.idup;
```
jacob-carlborg: You can `idup` a string that is `null`, which will save you the if statement.

src/dmd/link.d
```diff
@@ -1049,15 +1048,15 @@ version (Windows)
...
                 if (FileName.exists(lldpath))
-                    return lldpath;
+                    return lldpath.ptr;
             }
 
             // search PATH to avoid createProcess preferring "link.exe" from the dmd folder
             Strings* paths = FileName.splitPath(getenv("PATH"));
-            if (auto p = FileName.searchPath(paths, "link.exe", false))
-                return p;
+            if (auto p = FileName.searchPath(paths, "link.exe"[], false))
```
jacob-carlborg: Why is this slice needed?

Geod24: Otherwise the call is ambiguous, since we have a `const(char)*` and a `const(char)[]` overload, and manifest constant convert to both.
In a later PR (because this is already quite big) I'll remove the `const(char)*` overload

src/dmd/frontend.d
Geod24:
```diff
diff --git a/src/dmd/frontend.d b/src/dmd/frontend.d
index 93c5a8a94..9ae64a51a 100644
--- a/src/dmd/frontend.d
+++ b/src/dmd/frontend.d
@@ -79,11 +79,7 @@ string findDMDConfig(const(char)[] dmdFilePath)
     import dmd.dinifile : findConfFile;
     import std.string : fromStringz, toStringz;

-    auto f = findConfFile(dmdFilePath, "dmd.conf");
-    if (f is null)
-        return null;
-
-    return f.idup;
+    return findConfFile(dmdFilePath, "dmd.conf").dup;
 }

 /**
```

Updated thanks

src/dmd/root/filename.d
```diff
@@ -624,20 +705,34 @@ nothrow:
...
+         1 if it exists and is not a directory
+         2 if it exists and is a directory
+     */
     extern (C++) static int exists(const(char)* name)
+    {
+        return exists(name.toDString);
+    }
+
+    /// Ditto
+    extern (D) static int exists(const(char)[] name)
```
jacob-carlborg: Should probably return an enum instead of an int, but that might be something for a separate PR.

src/dmd/root/filename.d
```diff
@@ -938,12 +1053,12 @@ version(Posix)
         fileName = The file name to be converted to an absolute path
     Returns: Whatever `F` returns.
     */
-    auto absPathThen(alias F)(const(char)* fileName)
+    auto absPathThen(alias F)(const(char)[] fileName)
     {
         import core.sys.posix.stdlib: realpath, free;
-        auto absPath = realpath(fileName, null /* realpath allocates */);
+        char* absPath = fileName.toCStringThen!((fn) => realpath(&fn[0], null /* realpath allocates */));
```
jacob-carlborg: `const`?

src/dmd/link.d
jacob-carlborg: Ah, ok.

src/dmd/root/filename.d
jacob-carlborg: Ah, my bad.

jacob-carlborg (approved):

RazvanN7 (approved):

BBasile:
This causes a regression : https://issues.dlang.org/show_bug.cgi?id=19510.

Geod24:
I'll look at it

BBasile:
BTW, sorry since i discovered that your not responsible for the regression. It's more a Windows API call that was not used correctly... the problem was only revealed by your changes.

PR #8579, opened 2018-08-19
build.d: Get host's absolute path to dmd on Windows

JinShil:


jacob-carlborg (approved):

src/build.d
```diff
@@ -893,6 +893,23 @@ auto getHostCXXKind()
         static assert(false, "Unrecognized or unsupported OS.");
 }
 
+/*
+Gets the abosulte path of the host's dmd executable
```
Geod24: `abosulte`

PR #8583, opened 2018-08-20
Fix Issue 19180 - Expose dmd.mtype.Type.isZeroInit as __traits(isZeroInit, T)

n8sh:
It's useful to identify at compile time types with 0 initializers. Doing this with CTFE and templates is slow and memory-hungry[1] and runs into errors[2][3]. The compiler already has its own check for this that we can expose instead.

Proposed syntax is `__traits(isZeroInit, T)` where the second argument must resolve to the type T itself instead of being any expression whose result is an instance of T. This would be to prevent someone from mistakenly thinking he can use `__traits(isInitZero, x)` to test whether some variable `x` was explicitly initialized as zero.

[1] https://github.com/dlang/phobos/pull/6537
[2] https://github.com/dlang/phobos/pull/6670#issuecomment-414111649
[3] https://github.com/dlang/phobos/pull/6461

### DIFF
--- a/src/dmd/dstruct.d
+++ b/src/dmd/dstruct.d
@@ -401,9 +401,13 @@ extern (C++) class StructDeclaration : AggregateDeclaration
         {
             if (vd._init)
             {
-                // Should examine init to see if it is really all 0's
-                zeroInit = false;
-                break;
+                // Examine init to see if it is all 0s.
+                auto exp = vd.getConstInitializer();
+                if (!exp || !_isZeroInit(vd.type.toBasetype(), exp))
+                {
+                    zeroInit = false;
+                    break;
+                }
             }
             else if (!vd.type.isZeroInit(loc))
             {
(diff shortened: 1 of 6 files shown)


test/compilable/isZeroInit.d
```diff
@@ -0,0 +1,53 @@
...
+static assert(__traits(isZeroInit, S1));
+static assert(!__traits(isZeroInit, S2));
+
+void test()
+{
+    int a = 3;
+    float b = 0;
+    // Providing a specific variable rather than a type isn't allowed.
+    static assert(!__traits(compiles, __traits(isZeroInit, a)));
+    static assert(!__traits(compiles, __traits(isZeroInit, b)));
```
jacob-carlborg: We usually put failing tests in the `fail_compilation` directory which allows to verify the error message as well.

jacob-carlborg (dismissed):
Should have a changelog entry and a spec PR as well.

test/compilable/isZeroInit.d
```diff
@@ -0,0 +1,53 @@
...
+struct S3
+{
+    S2 x = { 0 };
+    float f = 0;
+}
+struct S4
+{
+    int a = 1;
+}
+static assert(__traits(isZeroInit, S1));
```
RazvanN7: I'm guessing this should be `S3`

test/compilable/isZeroInit.d
```diff
@@ -0,0 +1,53 @@
...
+static assert(!__traits(isZeroInit, S2));
```
RazvanN7: ...and `S4`

test/fail_compilation/fail_isZeroInit.d
```diff
@@ -0,0 +1,12 @@
+/*
+TEST OUTPUT:
+---
+fail_isZeroInit.d(11): Error: type expected as second argument of __traits isZeroInit instead of a
```
jacob-carlborg: You need to quote `isZeroInit` and `a` in the error message with backticks.

test/fail_compilation/fail_isZeroInit.d
```diff
@@ -0,0 +1,12 @@
...
+---
+*/
+void test()
+{
+    int a = 3;
+    // Providing a specific variable rather than a type isn't allowed.
+    enum bool az = __traits(isZeroInit, a);
+}
```
jacob-carlborg: Missing newline.

n8sh:
The test is failing because `.isZeroInit` is false for a struct with fields explicitly initialized to 0 (!).

```d
struct X1 { int x; }
static assert(__traits(isZeroInit, X1)); // Passes.

struct X2 { int x = 0; }
static assert(__traits(isZeroInit, X2)); // Fails!
```

test/compilable/isZeroInit.d
```diff
@@ -0,0 +1,50 @@
...
+{
+    static assert(__traits(isZeroInit, T) == (T.init is T(0)));
+    static assert(__traits(isZeroInit, T[2]) == (T.init is T(0)));
+
+    static assert(!__traits(isZeroInit, Holder!(T, 1)));
+    static assert(__traits(isZeroInit, Holder!(T, 0)));
+}
+
+static assert(__traits(isZeroInit, void)); // For initializing arrays of element type `void`.
+static assert(__traits(isZeroInit, void*));
```
wilzbach: Could we add a test for `static assert(__traits(isZeroInit, void[]));` too?

n8sh: Do you mean `void[10]`? Every dynamic array has a zero initializer.

src/dmd/dstruct.d
```diff
@@ -590,6 +597,98 @@ extern (C++) class StructDeclaration : AggregateDeclaration
...
+                    return false;
+                }
+            }
+            return true;
+        }
+//        else if (exp.op == TOK.string_ || exp.op == TOK.hexadecimalString)
+//        {
+//            // Don't bother checking for all-zero strings.
+//            return dim == 0;
+//        }
```
JinShil: Do we really want to keep this commented code?

test/fail_compilation/fail_isZeroInit.d
n8sh: This has been fixed.

RazvanN7 (approved):

RazvanN7:
@jacob-carlborg Any more objections before getting this in?

thewilsonator:
72h is up. Merge this?

n8sh:
@thewilsonator I think it's only been 70 hours, yeah?

src/dmd/dstruct.d
```diff
@@ -590,6 +594,100 @@ extern (C++) class StructDeclaration : AggregateDeclaration
     }
 }
 
+private bool _isZeroInit(Type type, Expression exp)
```
WalterBright: @RazvanN7 Functions need Ddoc comments.

PR #8584, opened 2018-08-20
Fix Issue 19181: always error when an UFCS or opDispatch call has syntactically invalid arguments

FeepingCreature:
Repro:

```
$ dmd test.d
---
module test;

struct S {
    void opDispatch(string name, T)(T arg) { }
}

void main() {
    S s;
    s.foo(LanguageError);
}
```

Before:
    test.d(9): Error: no property foo for type S

After:
    test.d(9): Error: undefined identifier LanguageError

### DIFF
--- a/test/fail_compilation/fail19181.d
+++ b/test/fail_compilation/fail19181.d
@@ -1,0 +1,16 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail19181.d(15): Error: undefined identifier `LanguageError`
+---
+*/
+struct S
+{
+    void opDispatch(string name, T)(T arg) { }
+}
+
+void main()
+{
+    S s;
+    s.foo(LanguageError);
+}
--- a/test/fail_compilation/ice14929.d
+++ b/test/fail_compilation/ice14929.d
@@ -75,7 +75,7 @@ template isComponentStorage(CS, C)
     {
         CS cs = CS.init;
         ulong eid;
-        cs.add(eid, c);
+        cs.add(eid, C.init);
     }));
 }
 
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -509,6 +509,10 @@ private Expression resolveUFCS(Scope* sc, CallExp ce)
         }
         else
         {
+            // even opDispatch and ufcs must have valid arguments.
+            if (arrayExpressionSemantic(ce.arguments, sc))
+                return new ErrorExp();
+
             if (Expression ey = die.semanticY(sc, 1))
             {
                 if (ey.op == TOK.error)


jacob-carlborg (approved):

FeepingCreature:
Holy hell, that's the fastest approval I have *ever* gotten on Github, let alone DMD.

jacob-carlborg (changes requested):
Needs to have a test.

JinShil:
Doesn't this need a test case to prevent a regression.

jacob-carlborg:
> Holy hell, that's the fastest approval I have ever gotten on Github, let alone DMD.

Sorry, I had to change my mind.

jacob-carlborg:
> Doesn't this need a test case to prevent a regression.

Yes, I changed my mind and requested changes.

FeepingCreature:
The original behavior is not strictly speaking _wrong_, it's just highly misleading. opDispatch fails to match since the call fails to compile - it just fails to compile due to a call-site error. That said, I'm not sure how you test for a specific error message.

JinShil:
> That said, I'm not sure how you test for a specific error message.

See examples in the `test/fail_compilation` directory for examples.  Also see `TEST_OUTPUT` described in the Readme at https://github.com/dlang/dmd/tree/master/test

FeepingCreature:
Test added.

test/fail_compilation/fail19181.d
```diff
@@ -0,0 +1,14 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail19181.d(13): Error: undefined identifier LanguageError
+---
+*/
+struct S {
```
jacob-carlborg: Please follow the standard D style, curly braces on their own lines.

jacob-carlborg (changes requested):

test/fail_compilation/fail19181.d
FeepingCreature: fixed

jacob-carlborg (approved):

test/fail_compilation/fail19181.d
```diff
@@ -0,0 +1,16 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/fail19181.d(15): Error: undefined identifier LanguageError
```
JinShil: I think the tests suite will probably fail because LanguageError needs to be surrounded with tick marks.  We'll find out soon enough.

FeepingCreature: Since the other tests have that too, I've just gone ahead and added it. Can't hurt.

jacob-carlborg:
Back to approved, still less than 20 minutes.

FeepingCreature:
Still the fastest, I believe!

FeepingCreature:
Again, that is not an error in my change.

JinShil:
> Again, that is not an error in my change.

Are you sure?  Run the test suite locally with `make -C test -f Makefile`.  Or to run just the failing test use `rdmd test/run.d test/fail_compilation/ice14929.d`.

test/fail_compilation/ice14929.d
```diff
@@ -75,7 +75,7 @@ template isComponentStorage(CS, C)
     {
         CS cs = CS.init;
         ulong eid;
-        cs.add(eid, c);
+        cs.add(eid, C.init);
```
jacob-carlborg: Why was this change necessary?

FeepingCreature: Fun reason: Previously, DMD would try to semantically evaluate `cs.add` _before_ figuring out what if anything `c` is, hit the `!in` instantiation error, and bail with the `TEST_OUTPUT` message. WIth this change, it verifies the arguments to cs.add() _first_ and immediately notices that there is no such variable as `c`, causing the `typeof` to fail and, as is `typeof`'s wont, silently return false.

jacob-carlborg: Ah, yes, of course. I was reading the corrected code and tried to figure out what was wrong with it :smiley:.

FeepingCreature: It is a bit weird and inconsistent that `typeof` errors out if only the error that happens in `typeof` is in a different template.

But that's neither what this bug nor the testcase's bug is about, so ... :man_shrugging:

wilzbach (approved):

wilzbach:
> Or to run just the failing test use rdmd test/run.d test/fail_compilation/ice14929.d.

FYI:  for the lazy people like me: `./test/run.d fail_compilation/ice14929.d` and a few other combinations work too.

MoonlightSentinel:
This introduced a regression: https://issues.dlang.org/show_bug.cgi?id=20488

FeepingCreature:
(Regression fixed.)

PR #8585, opened 2018-08-20
Cleanup and refactor some filename-related code to use D slices

Geod24:
As requested [by @WalterBright ](https://github.com/dlang/dmd/pull/8575#issuecomment-414230336).
Best reviewed commit-by-commit. Let me know if it should be broken down further.

### DIFF
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -295,39 +295,66 @@ nothrow:
         return f;
     }
 
+    /**
+       Combine a `path` and a file `name`
+
+       Params:
+         path = Path to append to
+         name = Name to append to path
+
+       Returns:
+         The `\0` terminated string which is the combination of `path` and `name`
+         and a valid path.
+    */
     extern (C++) static const(char)* combine(const(char)* path, const(char)* name)
     {
-        char* f;
-        size_t pathlen;
-        size_t namelen;
-        if (!path || !*path)
-            return cast(char*)name;
-        pathlen = strlen(path);
-        namelen = strlen(name);
-        f = cast(char*)mem.xmalloc(pathlen + 1 + namelen + 1);
-        memcpy(f, path, pathlen);
+        if (!path)
+            return name;
+        return combine(path.toDString, name.toDString).ptr;
+    }
+
+    /// Ditto
+    extern(D) static const(char)[] combine(const(char)[] path, const(char)[] name)
+    {
+        if (!path.length)
+            return name;
+
+        char* f = cast(char*)mem.xmalloc(path.length + 1 + name.length + 1);
+        memcpy(f, path.ptr, path.length);
+        bool trailingSlash = false;
         version (Posix)
         {
-            if (path[pathlen - 1] != '/')
+            if (path[$ - 1] != '/')
             {
-                f[pathlen] = '/';
-                pathlen++;
+                f[path.length] = '/';
+                trailingSlash = true;
             }
         }
         else version (Windows)
         {
-            if (path[pathlen - 1] != '\\' && path[pathlen - 1] != '/' && path[pathlen - 1] != ':')
+            if (path[$ - 1] != '\\' && path[$ - 1] != '/' && path[$ - 1] != ':')
             {
-                f[pathlen] = '\\';
-                pathlen++;
+                f[path.length] = '\\';
+                trailingSlash = true;
             }
         }
         else
         {
             assert(0);
         }
-        memcpy(f + pathlen, name, namelen + 1);
-        return f;
+        const len = path.length + trailingSlash;
+        memcpy(f + len, name.ptr, name.length);
+        // Note: At the moment `const(char)*` are being transitioned to
+        // `const(char)[]`. To avoid bugs crippling in, we `\0` terminate
+        // slices, but don't include it in the slice so `.ptr` can be used.
+        f[len + name.length] = '\0';
+        return f[0 .. len + name.length];
+    }
+
+    unittest
+    {
+        assert(combine("foo"[], "bar"[]) == "foo/bar");
+        assert(combine("foo/"[], "bar"[]) == "foo/bar");
     }
 
     static const(char)* buildPath(const(char)* path, const(char)*[] names...)
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -21,12 +21,13 @@ import core.sys.windows.windows;
 import dmd.root.array;
 import dmd.root.file;
 import dmd.root.outbuffer;
+import dmd.root.port;
 import dmd.root.rmem;
 import dmd.root.rootobject;
+import dmd.utils;
 
 nothrow
 {
-version (Windows) extern (C) int stricmp(const char*, const char*) pure;
 version (Windows) extern (Windows) DWORD GetFullPathNameW(LPCWSTR, DWORD, LPWSTR, LPWSTR*) @nogc;
 version (Windows) extern (Windows) void SetLastError(DWORD) @nogc;
 version (Windows) extern (C) char* getcwd(char* buffer, size_t maxlen);
@@ -49,30 +50,25 @@ nothrow:
         this.str = mem.xstrdup(str);
     }
 
-    extern (C++) bool equals(const RootObject obj) const pure
-    {
-        return compare(obj) == 0;
-    }
-
+    /// Compare two name according to the platform's rules (case sensitive or not)
     extern (C++) static bool equals(const(char)* name1, const(char)* name2) pure
     {
-        return compare(name1, name2) == 0;
+        return equals(name1.toDString, name2.toDString);
     }
 
-    extern (C++) int compare(const RootObject obj) const pure
+    /// Ditto
+    extern (D) static bool equals(const(char)[] name1, const(char)[] name2) pure
     {
-        return compare(str, (cast(FileName*)obj).str);
-    }
+        if (name1.length != name2.length)
+            return false;
 
-    extern (C++) static int compare(const(char)* name1, const(char)* name2) pure
-    {
         version (Windows)
         {
-            return stricmp(name1, name2);
+            return Port.memicmp(name1.ptr, name2.ptr, name1.length) == 0;
         }
         else
         {
-            return strcmp(name1, name2);
+            return name1 == name2;
         }
     }
 
@@ -469,7 +465,7 @@ nothrow:
             return true;
         if (!e || !ext)
             return false;
-        return FileName.compare(e, ext) == 0;
+        return FileName.equals(e, ext);
     }
 
     /******************************
--- a/src/dmd/root/stringtable.d
+++ b/src/dmd/root/stringtable.d
@@ -57,6 +57,12 @@ pure:
     {
         return cast(const(char)*)(&this + 1);
     }
+
+    /// Returns: The content of this entry as a D slice
+    extern (D) inout(char)[] toString() inout
+    {
+        return (cast(inout(char)*)(&this + 1))[0 .. length];
+    }
 }
 
 struct StringTable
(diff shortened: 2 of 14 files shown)


src/dmd/utils.d
```diff
@@ -138,3 +139,44 @@ const(char)[] toDString (const(char)* s) pure nothrow @nogc
...
+    }
+    ptr[0 .. src.length] = src[];
+    ptr[src.length] = '\0';
+    return dg(ptr);
+}
+
+unittest
+{
+    assert("Hello world".toCStringThen!((v) => v == "Hello world\0"));
+    assert("Hello world\0".toCStringThen!((v) => v == "Hello world\0\0"));
```
jacob-carlborg: Why adding an additional null character if one is already present?

src/dmd/root/stringtable.d
```diff
@@ -57,6 +57,11 @@ pure:
     {
         return cast(const(char)*)(&this + 1);
     }
+
+    extern (D) inout(char)[] toString() inout
```
jacob-carlborg: Should at least have an empty Ddoc comment.

src/dmd/utils.d
Geod24: This function does not care about the content of the slice. And slices should not contain the terminating null char, it should be out of bounds.

Geod24:
Updated:
```diff
diff --git a/src/dmd/root/stringtable.d b/src/dmd/root/stringtable.d
index aa73bb82c..2d9b1876c 100644
--- a/src/dmd/root/stringtable.d
+++ b/src/dmd/root/stringtable.d
@@ -58,6 +58,7 @@ pure:
         return cast(const(char)*)(&this + 1);
     }

+    /// Returns: The content of this entry as a D slice
     extern (D) inout(char)[] toString() inout
     {
         return (cast(inout(char)*)(&this + 1))[0 .. length];
diff --git a/src/dmd/root/filename.d b/src/dmd/root/filename.d
index ba34d6e44..e6afe9e7e 100644
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -298,6 +298,10 @@ nothrow:
     /**
        Combine a `path` and a file `name`

+       Params:
+         path = Path to append to
+         name = name to append to path
+
        Returns:
          The `\0` terminated string which is the combination of `path` and `name`
          and a valid path.
```

src/dmd/utils.d
```diff
@@ -138,3 +139,44 @@ const(char)[] toDString (const(char)* s) pure nothrow @nogc
...
+   Note:
+     The argument to `dg` is `scope`. To keep the data around after `dg` exits,
+     one has to copy it.
+
+   Params:
+     src = Slice to use to call the C function
+     dg  = Delegate to call afterwards
+
+   Returns:
+     The return value of `T`
```
wilzbach: Nit: For new Phobos code, we recommend to not use any white space or stars as indentation.

src/dmd/utils.d
```diff
@@ -138,3 +139,44 @@ const(char)[] toDString (const(char)* s) pure nothrow @nogc
...
+*/
+auto toCStringThen(alias dg)(const(char)[] src) nothrow
```
wilzbach: Maybe include "temporary" in the name, s.t. it's always clear that keeping the C string around is dangerous?

src/dmd/utils.d
```diff
@@ -138,3 +139,44 @@ const(char)[] toDString (const(char)* s) pure nothrow @nogc
...
+{
+    char[512] small = void;
```
wilzbach: How about as a simple optimization checking whether the last character (or the one after) is already a `\0` and thus skipping the copying all together?

src/dmd/utils.d
wilzbach: But all static D strings already do contain a null afterwards, e.g. https://run.dlang.io/is/Abty1m so it looks like we could save quite a bit of copying here.

Geod24: >  e.g. https://run.dlang.io/is/Abty1m

Your example is reading past the end of the slice. It's unsafe and cannot be relied on in this function.
Also that copy is unlikely to have any noticeable performance impact, since this function will be called  primarily to wrap C function calls, and we'll be calling mostly I/O functions.

src/dmd/utils.d
Geod24: I used this name because it matches [what's already there](
https://github.com/dlang/dmd/blob/c3f6320d59ec14d6fa81a18f92b59393671b346a/src/dmd/root/filename.d#L818), namely `toWStringzThen` and `toExtendedPathThen`.
Marking those functions `@safe` should enforce that `scope` is checked for, but we cannot (yet) do that.

wilzbach: >  should enforce that scope is checked for, but we cannot (yet) do that.

Well safety is still broken for now too, e.g. https://issues.dlang.org/show_bug.cgi?id=19175

src/dmd/root/filename.d
```diff
@@ -49,30 +50,25 @@ nothrow:
...
-    }
+        if (name1.length != name2.length)
+            return false;
 
-    extern (C++) static int compare(const(char)* name1, const(char)* name2) pure
-    {
         version (Windows)
         {
-            return stricmp(name1, name2);
+            return strnicmp(name1.ptr, name2.ptr, name1.length) == 0;
```
wilzbach: > These POSIX functions are deprecated. Use the ISO C++ conformant _strnicmp, _wcsnicmp, _mbsnicmp, _strnicmp_l, _wcsnicmp_l, _mbsnicmp_l instead.

https://msdn.microsoft.com/en-us/library/ms235324.aspx

src/dmd/root/filename.d
```diff
@@ -49,30 +50,25 @@ nothrow:
...
         }
         else
         {
-            return strcmp(name1, name2);
+            return memcmp(name1.ptr, name2.ptr, name1.length) == 0;
```
wilzbach: Why not simply `name == name2`?

Geod24:
Updated:
```diff
diff --git a/src/dmd/root/filename.d b/src/dmd/root/filename.d
index e6afe9e7e..606cb3107 100644
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -27,7 +27,13 @@ import dmd.utils;

 nothrow
 {
-version (Windows) extern (C) int strnicmp(const char*, const char*, size_t) pure;
+// Win32 prepend an `_`, so this will call the non-deprecated `_strnicmp`
+version (Win32) extern (C) int strnicmp(const char*, const char*, size_t) pure;
+else version (Windows)
+{
+    extern (C) int _strnicmp(const char*, const char*, size_t) pure;
+    alias strnicmp = _strnicmp;
+}
 version (Windows) extern (Windows) DWORD GetFullPathNameW(LPCWSTR, DWORD, LPWSTR, LPWSTR*) @nogc;
 version (Windows) extern (Windows) void SetLastError(DWORD) @nogc;
 version (Windows) extern (C) char* getcwd(char* buffer, size_t maxlen);
@@ -68,7 +74,7 @@ nothrow:
         }
         else
         {
-            return memcmp(name1.ptr, name2.ptr, name1.length) == 0;
+            return name1 == name2;
         }
     }
```

src/dmd/root/filename.d
WalterBright: I see strncmp has reappeared in another guise. Please don't use it. I pointed this out in a previous PR, and why.

src/dmd/root/filename.d
```diff
@@ -299,39 +301,66 @@ nothrow:
...
+                trailingSlash = true;
             }
         }
         else
         {
             assert(0);
         }
-        memcpy(f + pathlen, name, namelen + 1);
-        return f;
+        const len = path.length + (trailingSlash ? 1 : 0);
```
WalterBright: The ?1:0 is redundant, because trailingSlash is a bool that is already 1 or 0.

src/dmd/root/filename.d
WalterBright: Previous review of strncmp:
https://github.com/dlang/dmd/pull/8575#pullrequestreview-147490646
"strncmp is the wrong answer here. I quote the C Standard: "The strncmp function compares not more than n characters (characters that follow a null character are not compared) from the array pointed to by s1 to the array pointed to bys 2." I.e. embedded 0's are still a factor. Use memcmp() and memicmp() instead. I've never found a legitimate use for strncmp()."

src/dmd/root/filename.d
Geod24: Didn't knew it was guaranteed by the language. D never cease to amaze me.

wilzbach:
> Buildkite is being weird @wilzbach 

Sorry, my fault. Should be fixed now. I restarted the build.

src/dmd/utils.d
Geod24: That's absolutely hideous, but if it's the D style...

Geod24:
All comments I did not answer should be addressed.

src/dmd/utils.d
jacob-carlborg: I agree, it's hideous.

jacob-carlborg (approved):

src/dmd/utils.d
```diff
@@ -138,3 +139,44 @@ const(char)[] toDString (const(char)* s) pure nothrow @nogc
...
+dg  = Delegate to call afterwards
+
+Returns:
+The return value of `T`
+*/
+auto toCStringThen(alias dg)(const(char)[] src) nothrow
+{
+    char[512] small = void;
+    scope ptr = (src.length < (small.length - 1))
+                    ? small[0 .. src.length + 1]
```
JinShil: `src.length + 1` is excessively repeated.  I think it can be stored in a constant variable and reused.  It's asking for the next developer to introduce an off-by-1 error.

Geod24:
Updated:
```diff
diff --git a/src/dmd/utils.d b/src/dmd/utils.d
index 55a9bb54d..5747a50a4 100644
--- a/src/dmd/utils.d
+++ b/src/dmd/utils.d
@@ -160,10 +160,11 @@ The return value of `T`
 */
 auto toCStringThen(alias dg)(const(char)[] src) nothrow
 {
+    const len = src.length + 1;
     char[512] small = void;
     scope ptr = (src.length < (small.length - 1))
-                    ? small[0 .. src.length + 1]
-                    : (cast(char*)mem.xmalloc(src.length + 1))[0 .. src.length + 1];
+                    ? small[0 .. len]
+                    : (cast(char*)mem.xmalloc(len))[0 .. len];
     scope (exit)
     {
         if (&ptr[0] != &small[0])
```

src/dmd/root/filename.d
```diff
@@ -299,39 +295,66 @@ nothrow:
         return f;
     }
 
+    /**
+       Combine a `path` and a file `name`
+
+       Params:
+         path = Path to append to
+         name = name to append to path
```
JinShil: Capitalize "name" in the description.

JinShil (approved):

src/dmd/root/filename.d
Geod24: Done

JinShil:
I think this might cause a merge conflict with https://github.com/dlang/dmd/pull/8542.  Let's try to get that in first.

Geod24:
> I think this might cause a merge conflict with #8542. Let's try to get that in first.

It definitely will, as this PR was what originally prompted me to do this.
I'll wait for the stable merge then.

JinShil (approved):

Geod24:
> @JinShil JinShil added the auto-merge  label 10 minutes ago

#8542 was to stable. While it was merged, stable hasn't been merged to stable yet, so this is still blocked. I'll remove the label until then :'(

ibuclaw:
@Geod24 rebase!

Geod24:
Already done and no conflict

Geod24:
Re-adding auto-merge now that stable has been merged and it didn't lead to any code modification.

PR #8586, opened 2018-08-20
build.d: Remove redundant appending of backend object files

JinShil:
Due to file globbing, these object files were already added.  In fact, rather than appending object files, we should probably be removing some.


wilzbach (approved):
Yes this is from the times when `build.d` didn't do globbing (and the `posix.mak` was more or less literally translated). This works because of https://github.com/dlang/dmd/pull/7567 btw.

PR #8587, opened 2018-08-21
build.d: Remove redundant backend target objects

JinShil:
These files are already being added with the code at https://github.com/dlang/dmd/blob/c3f6320d59ec14d6fa81a18f92b59393671b346a/src/build.d#L787-L793

They are currently being built twice.

### DIFF
--- a/src/build.d
+++ b/src/build.d
@@ -720,7 +720,6 @@ auto sourceFiles()
     if (env["TARGET_CPU"] == "X86")
     {
         targetCH = "code_x86.h";
-        targetObjs = ["cgxmm", "cod1", "cod2", "cod3", "cod4", "ptrntab"];
     }
     else if (env["TARGET_CPU"] == "stub")
     {


JinShil:
No, the makefile hard codes each file to build, but build.d does file globbing.  The two are quite different, so we're not actually able to cross-reference them to each other.  It's quite simple to verify, though.  At the end of the `sourceFiles()` function just do `writeln(sources.backendObjects)` and you'll see they've been added twice.

JinShil:
Also I believe these duplications were mistakenly added by @WalterBright  in some of his C to D conversions.  See https://github.com/dlang/dmd/pull/8577/files#diff-9be7d53d2c3189b3d2d85ee53bb2adb0L724

PetarKirov (approved):
LGTM

wilzbach:
> The two are quite different, so we're not actually able to cross-reference them to each other.

Due to the globbing and custom-dependency system, yes, but FWIW here's the respective part from the Makefile:

https://github.com/dlang/dmd/blob/b43d3fc9150aeb1fa7638af746c791fd72d4257b/src/posix.mak#L332-L342

JinShil:
Thanks @wilzbach for clarifying.  Sorry @jacob-carlborg if I led you astray; I'm still learning the build system and that was my understanding at the time I replied.

PR #8588, opened 2018-08-21
[stable] [from master] Issue 19181: when an UFCS or opDispatch call has syntactically invalid arguments, always error.

FeepingCreature:
Cherry-pick from master. Sorry, I didn't understand how stable worked when I made the original PR and I don't particularly want to wait two months.

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -509,6 +509,10 @@ private Expression resolveUFCS(Scope* sc, CallExp ce)
         }
         else
         {
+            // even opDispatch and ufcs must have valid arguments.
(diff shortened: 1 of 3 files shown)


FeepingCreature:
ping

jacob-carlborg:
If I understand correctly we should not cherry pick.

FeepingCreature:
Yeah explanations in slack suggest that adding the same commit in both branches will lead to merge issues later on.

MartinNowak:
If you need to cherry-pick, please prefer to pick the merge commit (c2317ad3b7401dc9fe26a28cca66846a892d5d34), so we can find the original PR from the commit message.

PR #8589, opened 2018-08-21
build.d: Build msvc-dmc and msvc-lib for Windows

JinShil:
I had to move the logic for `HOST_DMD_RUN` to `parseEnvironment` because DMD is needed to build `msvc-dmc` and `msvc-lib`, and those dependencies are used in the logic to determine `HOST_CXX`.

`msvc-dmc` and `msvc-lib` build, but they don't run yet; I need to update the logic for locating the Visual C++ executable.  I think we should use https://github.com/Microsoft/vswhere.  I intend to do that in the next PR.

### DIFF
--- a/src/build.d
+++ b/src/build.d
@@ -246,7 +246,7 @@ version(Windows)
         Dependency dependency = {
             target: env["G"].buildPath(targetName).exeName,
             sources: [`vcbuild\` ~ targetName],
-            name: "(DC) CC " ~ targetName,
+            name: "(DC) MSCV-CC " ~ targetName,
             command: [env["HOST_DMD_RUN"], "-of$@", "$<"]
         };
         return dependency;
@@ -260,7 +260,7 @@ version(Windows)
         Dependency dependency = {
             target: env["G"].buildPath(targetName).exeName,
             sources: [`vcbuild\` ~ targetName],
-            name: "(DC) LIB " ~ targetName,
+            name: "(DC) MSCV-LIB " ~ targetName,
             command: [env["HOST_DMD_RUN"], "-of$@", "$<"]
         };
         return dependency;
--- a/src/build.d
+++ b/src/build.d
@@ -241,9 +241,13 @@ version(Windows)
     // Build the msvc-dmc compiler wrapper
     auto buildMsvcDmc()
     {
+        enum targetName = "msvc-dmc";
+
         Dependency dependency = {
-            target: env["G"].buildPath("msvc-dmc").exeName,
-            sources: [`vcbuild\msvc-dmc`],
+            target: env["G"].buildPath(targetName).exeName,
+            sources: [`vcbuild\` ~ targetName],
+            name: "(DC) CC " ~ targetName,
+            command: [env["HOST_DMD_RUN"], "-of$@", "$<"]
         };
         return dependency;
     }
@@ -251,9 +255,13 @@ version(Windows)
     // Build the msvc-lib linker wrapper
     auto buildMsvcLib()
     {
+        enum targetName = "msvc-lib";
+
         Dependency dependency = {
-            target: env["G"].buildPath("msvc-lib").exeName,
-            sources: [`vcbuild\msvc-lib`],
+            target: env["G"].buildPath(targetName).exeName,
+            sources: [`vcbuild\` ~ targetName],
+            name: "(DC) LIB " ~ targetName,
+            command: [env["HOST_DMD_RUN"], "-of$@", "$<"]
         };
         return dependency;
     }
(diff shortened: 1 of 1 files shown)


src/build.d
```diff
@@ -366,8 +365,21 @@ and afterwards builds the DMD compiler.
...
+        {
+            dependencies = [buildMsvcDmc, buildMsvcLib];
+            foreach (dependency; dependencies.parallel(1))
+                dependency.run;
+        }
+    }
+
     // The string files are required by most targets
-    Dependency[] dependencies = buildStringFiles();
+    dependencies = buildStringFiles();
```
wilzbach: Never a good idea to overwrite variables.

src/build.d
```diff
@@ -366,8 +365,21 @@ and afterwards builds the DMD compiler.
...
 auto buildDMD()
 {
+    Dependency[] dependencies;
+
+    version(Windows)
+    {
+        immutable model = detectModel;
+        if (model == "64")
+        {
+            dependencies = [buildMsvcDmc, buildMsvcLib];
```
wilzbach: I would do sth. like `const windowsDependencies`

src/build.d
```diff
@@ -241,19 +241,27 @@ version(Windows)
...
     auto buildMsvcDmc()
     {
+        enum targetName = "msvc-dmc";
+
         Dependency dependency = {
-            target: env["G"].buildPath("msvc-dmc").exeName,
-            sources: [`vcbuild\msvc-dmc`],
+            target: env["G"].buildPath(targetName).exeName,
+            sources: [`vcbuild\` ~ targetName],
+            name: "Building " ~ targetName,
```
wilzbach: Not saying that it's great, but the convention which was taken over from the Makefile is to do sth. like:

```d
name: "(CC) VCBUILD " ~ targetName
```

Though we probably should start using `DC` wherever a D compiler is used.

src/build.d
```diff
@@ -241,19 +241,27 @@ version(Windows)
...
     {
+        enum targetName = "msvc-lib";
+
         Dependency dependency = {
-            target: env["G"].buildPath("msvc-lib").exeName,
-            sources: [`vcbuild\msvc-lib`],
+            target: env["G"].buildPath(targetName).exeName,
+            sources: [`vcbuild\` ~ targetName],
+            name: "Building " ~ targetName,
+            command: [env["HOST_DMD_RUN"], "-of$@", "$<"]
```
wilzbach: Sorry I thought that was the default. Should we set this as the default command?

JinShil: I didn't know there was a default command.  Please elaborate.

wilzbach:
> I didn't know there was a default command. Please elaborate.

There isn't, but we could do:

```d
diff --git a/src/build.d b/src/build.d
index 0f8db4fa3..bebdab3b6 100755
--- a/src/build.d
+++ b/src/build.d
@@ -1041,6 +1041,9 @@ struct Dependency
         if (name)
             name.writeln;
 
+        if (command is null)
+            command = [env["HOST_DMD_RUN"], "-of$@", "$<"];
+
         command.runCanThrow;
     }
```

Though thinking about it, maybe it's good to be explicit for now.

wilzbach (approved):
Thanks! I appended a slight improvement of the dependency command name myself. Let me know if you're okay with it or feel free to merge directly yourself ;-)

src/build.d
```diff
@@ -366,6 +365,16 @@ and afterwards builds the DMD compiler.
 */
 auto buildDMD()
 {
+    version(Windows)
+    {
+        immutable model = detectModel;
```
wilzbach: BTW as `detectModel` is now called 3x we might want to cache it in the future, but I think the cost of one `uname` is insignificant to invoking DMD many times.

wilzbach:
> I need to update the logic for locating the Visual C++ executable. I think we should use https://github.com/Microsoft/vswhere. I intend to do that in the next PR.

I have no idea about Windows (CC @rainers and @CyberShadow), but I guess you can always use std.net.curl's [download](https://dlang.org/phobos/std_net_curl.html#.download) to fetch the binary from GitHub (though we need to do this in a loop like e.g. we [do in dub](https://github.com/dlang/dub/blob/b7e1e461ea7ecc716a5cc5092d0ae5e3440dabe6/source/dub/packagesuppliers/registry.d#L61-L73) as the `download` method of std.net.curl doesn't allow to use curl's retry feature which is very important for CI usage.

src/build.d
JinShil: Yeah, I'm not too happy about that either.  But, if I'm permitted to have my druthers, after this is working for Windows, added to the CIs, and given a little field testing, I would like to do a major refactor of the code so it no longer resembles a makefile, but rather more of an idiomatic D program.  We'll see, and I'm sure there will be differences of opinions, but I'm sure we'll get something quite nice in the end.

rainers:
> Microsoft has standardized the location of vswhere to %ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe, so I intend to do the following:

I have never used vswhere, but here is what dmd does: https://github.com/dlang/dmd/blob/master/src/dmd/link.d#L943, i.e. there is a lot more to it than just finding the VS installation path.

PR #8591, opened 2018-08-21
[WIP] Generate C++ header files from public extern (C++) declarations

ibuclaw:
Takes #5082 and converts it into a standalone program.

Still needs a bit of work, documentation, command-line option handling maybe to turn on/off certain parts.

Some bugs that I reported are still present, such as https://github.com/dlang/dmd/pull/5082#issuecomment-217176635

Using the auto-generated header in the C++ unittest works at least.


jacob-carlborg:
I would recommend using the Dub package. Also building the new tool as a Dub package.

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
...
+    "json.d",
+    "lambdacomp.d", "lexer.d",
+    "mtype.d",
+    "nogc.d", "nspace.d",
+    "objc.d", "opover.d", "optimize.d",
+    "parse.d", "printast.d",
+    "safe.d", "sapply.d", "semantic2.d", "semantic3.d", "sideeffect.d", "statement.d",
+    "statementsem.d", "staticassert.d", "staticcond.d",
+    "target.d", "templateparamsem.d", "tokens.d", "traits.d", "typesem.d", "typinf.d",
+    "utf.d", "utils.d",
```
wilzbach: Can't we use globbing here like we do for the dub package or `build.d`, s.t. we have less files to update whenever a file gets moved?

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
...
+                auto save = adparent;
+                adparent = sd;
+                foreach (m; *sd.members)
+                {
+                    m.accept(this);
+                }
+                adparent = save;
+                // Generate default ctor
+                buf.printf("    %s(", sd.ident.toChars());
+                buf.printf(") {");
```
wilzbach: AFAICT the header files currently don't have this. Why was generating a default ctor necessary?

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
...
+                adparent = cd;
+                foreach (m; *cd.members)
+                {
+                    m.accept(this);
+                }
+                adparent = save;
+                // Generate special static inline function.
+                if (cd.isIdentifierClass())
+                {
+                    buf.writestring("    static inline Identifier *idPool(const char *s) { return idPool(s, strlen(s)); }\n");
```
wilzbach: Why this special casing?

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
...
+
+            if (td.getModule() && !td.getModule().isFrontendModule())
+                return;
+            if (!td.parameters || !td.onemember || !td.onemember.isStructDeclaration())
+            {
+                visit(cast(AST.Dsymbol)td);
+                return;
+            }
+
+            // Explcitly disallow templates with non-type parameters or specialization.
```
wilzbach: Explicitly

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
...
+    VersionCondition.addPredefinedGlobalIdent("linux");
+    VersionCondition.addPredefinedGlobalIdent("CRuntime_Glibc");
+    VersionCondition.addPredefinedGlobalIdent("NoBackend");
+    VersionCondition.addPredefinedGlobalIdent("NoMain");
+    if (global.params.is64bit)
+        VersionCondition.addPredefinedGlobalIdent("X86_64");
+    else
+        VersionCondition.addPredefinedGlobalIdent("X86");
+    if (global.params.isLP64)
+        VersionCondition.addPredefinedGlobalIdent("D_LP64");
```
wilzbach: The frontend module (available via the dub package) is intended to avoid exactly this boilerplate code.
See https://github.com/dlang/dmd/blob/master/src/dmd/frontend.d and e.g. https://github.com/dlang/dmd/blob/master/test/dub_package/frontend.d

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
...
+
+    GC.disable();
+
+    global._init();
+    global.params.isLinux = true;
+    global.params.is64bit = (size_t.sizeof == 8);
+    global.params.isLP64 = global.params.is64bit;
+
+    global.path = new Strings();
+    global.path.push(__FILE_FULL_PATH__.dirName.buildPath("../../../druntime/src/").toStringz());
```
wilzbach: This is fragile if `druntime` is checkout out in a different directory (I think the LDC people do this).

wilzbach (commented):
Yeah it would save a lot of code in the main method if this would use the DMD dub package.

wilzbach:
SemaphoreCI failure is due to the dub package trying to include `dtoh`:

```
../../src/dmd/dtoh.d(103,23): Deprecation: Using super as a type is deprecated. Use typeof(super) instead
../../src/dmd/dtoh.d(131,23): Error: function `dmd.dtoh.genCppFiles.ToCppBuffer!(ASTCodegen).ToCppBuffer.visit` conflicts with alias dmd.dtoh.genCppFiles.ToCppBuffer!(ASTCodegen).ToCppBuffer.visit at ../../src/dmd/dtoh.d(103,9)
../../src/dmd/dtoh.d(1140,19): Error: template instance `dmd.dtoh.genCppFiles.ToCppBuffer!(ASTCodegen)` error instantiating
```

src/dmd/dtoh.d
ibuclaw: Maybe, this part was carried over from the old PR.

For sure am open to design improvements - the most important part is the blacklisting of certain modules that we retain manual updates for.

src/dmd/dtoh.d
ibuclaw: You're thinking classes.  This is for structs, and we do have this in some places, such as `Loc`.

src/dmd/dtoh.d
ibuclaw: Because this doesn't exist in D for obvious reasons.

wilzbach: Oh I was more asking: shouldn't we do this in a more general-purpose way.

jacob-carlborg:
Related [1]. Perhaps join forces?

[1] https://forum.dlang.org/thread/oikrrejcengbjivsyzws@forum.dlang.org

ibuclaw:
@jacob-carlborg this differs in that it is not intended to be a general purpose tool.  Same as the tool used to convert dmd to D. This handles only dmd specific style and quirks.

kinke:
Nice job Iain, thanks!

src/dmd/dtoh.d
thewilsonator: Yes our path is `$LDCROOT/runtime/druntime`.

src/dmd/intrange.d
```diff
@@ -43,7 +41,7 @@ struct SignExtendedNumber
 
     static SignExtendedNumber max()
     {
-        return SignExtendedNumber(UINT64_MAX, false);
```
WalterBright: Why is this refactoring here? Perhaps make it a separate PR (as this one is very large)?

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
+// Compiler implementation of the D programming language
+// Copyright (c) 1999-2015 by Digital Mars
+// All Rights Reserved
+// written by Walter Bright
+// http://www.digitalmars.com
+// Distributed under the Boost Software License, Version 1.0.
+// http://www.boost.org/LICENSE_1_0.txt
```
WalterBright: Please use comparable header to other dmd source files.

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
...
+    "nogc.d", "nspace.d",
+    "objc.d", "opover.d", "optimize.d",
+    "parse.d", "printast.d",
+    "safe.d", "sapply.d", "semantic2.d", "semantic3.d", "sideeffect.d", "statement.d",
+    "statementsem.d", "staticassert.d", "staticcond.d",
+    "target.d", "templateparamsem.d", "tokens.d", "traits.d", "typesem.d", "typinf.d",
+    "utf.d", "utils.d",
+];
+
+private bool isIdentifierClass(ASTCodegen.ClassDeclaration cd)
```
WalterBright: Since this is not C++, the public functions should come first.

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
...
+            cd.parent.ident == Identifier.idPool("identifier") &&
+            cd.parent.parent && cd.parent.parent.ident == Identifier.idPool("dmd") &&
+            !cd.parent.parent.parent);
+}
+
+private bool isVisitorClass(ASTCodegen.ClassDeclaration cd)
+{
+    for (auto cdb = cd; cdb; cdb = cdb.baseClass)
+    {
+        if (cdb.ident == Identifier.idPool("Visitor") ||
```
WalterBright: This loop invariant inside the loop will be agonizingly slow.

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1256 @@
...
+private bool isIdentifierClass(ASTCodegen.ClassDeclaration cd)
+{
+    return (cd.ident == Identifier.idPool("Identifier") &&
+            cd.parent !is null &&
+            cd.parent.ident == Identifier.idPool("identifier") &&
+            cd.parent.parent && cd.parent.parent.ident == Identifier.idPool("dmd") &&
+            !cd.parent.parent.parent);
+}
+
+private bool isVisitorClass(ASTCodegen.ClassDeclaration cd)
```
WalterBright: In general, this file completely lacks comments of any sort.

TurkeyMan:
What's going on with this? I could really use this like, yesterday.

ibuclaw:
@TurkeyMan - it's hardly meant for general purpose, there are many special cases that look for dmd-specific code.

If it gets used at all, it'll be part of the build process for generating headers that gdc and ldc can include.  No need to check what is and isn't in sync.

TurkeyMan:
Right... I want to do an experiment where I define some struct's and functions in D, but they are extern(C++), and the .h file needs to be emit by the D compiler. Is that the goal here? It sounds like the same goal; keeping the C++ headers in sync.
This would be a massive enabler for my current project.

thewilsonator:
@TurkeyMan I'll do this up as a dub package once #8776 goes in.

WalterBright:
Is this still a WIP?

thewilsonator:
This needs:
* the druntime path to be fixed for LDC
* made to work for non-linux, and 32-bit

It would also be good to have the DMD-specific stuff factored out.

ibuclaw:
> Is this still a WIP?

Yes, as I have done no testing whatsoever against gdc and auto-generated headers.

I've been too busy with work on other things to look back at this (you should know what other things are :-)

The cxxfrontend tests that are in dmd only cover the surface.

As I've said before, this not meant to be general purpose, I would suggest maybe dpp is better suited for you than this utility tool.

thewilsonator:
As a general purpose dub-package-to-be: https://github.com/thewilsonator/dtoh

N.B. also completely untested.

jacob-carlborg:
> I would suggest maybe dpp is better suited for you than this utility tool.

dpp is for the other way.

TurkeyMan:
> As I've said before, this not meant to be general purpose, I would suggest maybe dpp is better suited for you than this utility tool.

dpp is the wrong way around; we want to emit `.h` files with the extern(C++) symbols like `.di` files work today... in this case, the D code should be the authoritative code. I don't want C++ injecting noise into this particular codebase.

src/dmd/dtoh.d
```diff
@@ -0,0 +1,1496 @@
...
+            global.inifilename = findConfFile(global.params.argv0, "dmd.conf").ptr;
+        }
+        else
+        {
+            static assert(0, "fix this");
+        }
+    }
+    // Read the configurarion file
+    auto inifile = File(global.inifilename);
+    inifile.read();
```
wilzbach: I think most of this code in this `main` method could be saved by using DMD's `dmd.frontend`: 
- https://github.com/dlang/dmd/blob/master/src/dmd/frontend.d
- https://github.com/dlang/dmd/blob/master/test/dub_package/frontend.d

thewilsonator:
There is https://github.com/dlang/dmd/pull/9971 now, can this be closed?

ibuclaw:
If they're willing to take the baton in this relay race.

wilzbach:
It's Edi's GSoC project, so he'll be working full-time on this over the next weeks.

PR #8592, opened 2018-08-22
[Refactor] Some more C string removal

Geod24:
A few more commits cleaning up strings while we wait on #8585 .

### DIFF
--- a/src/dmd/doc.d
+++ b/src/dmd/doc.d
@@ -1833,20 +1833,25 @@ extern (C++) bool isDitto(const(char)* comment)
  */
 extern (C++) const(char)* skipwhitespace(const(char)* p)
 {
-    for (; 1; p++)
+    return skipwhitespace(p.toDString).ptr;
+}
+
+/// Ditto
+extern (D) const(char)[] skipwhitespace(const(char)[] p)
+{
+    foreach (idx, char c; p)
     {
-        switch (*p)
+        switch (c)
         {
         case ' ':
         case '\t':
         case '\n':
             continue;
         default:
-            break;
+            return p[idx .. $];
         }
-        break;
     }
-    return p;
+    return p[$ .. $];
 }
 
 /************************************************
--- a/src/dmd/doc.d
+++ b/src/dmd/doc.d
@@ -102,7 +102,7 @@ extern (C++) class Section
         assert(a.dim);
         if (namelen)
         {
-            __gshared const(char)** table =
+            static immutable table =
             [
                 "AUTHORS",
                 "BUGS",
@@ -117,13 +117,12 @@ extern (C++) class Section
                 "STANDARDS",
                 "THROWS",
                 "VERSION",
-                null
             ];
-            for (size_t i = 0; table[i]; i++)
+            foreach (entry; table)
             {
-                if (icmp(table[i], name, namelen) == 0)
+                if (icmp(entry.ptr, name, namelen) == 0)
                 {
-                    buf.printf("$(DDOC_%s ", table[i]);
+                    buf.printf("$(DDOC_%s ", entry.ptr);
                     goto L1;
                 }
             }
--- a/src/dmd/doc.d
+++ b/src/dmd/doc.d
@@ -120,7 +120,7 @@ extern (C++) class Section
             ];
             foreach (entry; table)
             {
-                if (icmp(entry.ptr, name, namelen) == 0)
+                if (iequals(entry, name[0 .. namelen]))
                 {
                     buf.printf("$(DDOC_%s ", entry.ptr);
                     goto L1;
@@ -1431,11 +1431,11 @@ struct DocComment
         for (size_t i = 0; i < dc.sections.dim; i++)
         {
             Section sec = dc.sections[i];
-            if (icmp("copyright", sec.name, sec.namelen) == 0)
+            if (iequals("copyright", sec.name[0 .. sec.namelen]))
             {
                 dc.copyright = sec;
             }
-            if (icmp("macros", sec.name, sec.namelen) == 0)
+            if (iequals("macros", sec.name[0 .. sec.namelen]))
             {
                 dc.macros = sec;
             }
@@ -1519,7 +1519,7 @@ struct DocComment
                 // Output existing macro
             L1:
                 //printf("macro '%.*s' = '%.*s'\n", namelen, namestart, textlen, textstart);
-                if (icmp("ESCAPES", namestart, namelen) == 0)
+                if (iequals("ESCAPES", namestart[0 .. namelen]))
                     parseEscapes(pescapetable, textstart, textlen);
                 else
                     Macro.define(pmacrotable, namestart[0 ..namelen], textstart[0 .. textlen]);
@@ -1698,9 +1698,9 @@ struct DocComment
             if (namelen || pstart < pend)
             {
                 Section s;
-                if (icmp("Params", name, namelen) == 0)
+                if (iequals("Params", name[0 .. namelen]))
                     s = new ParamSection();
-                else if (icmp("Macros", name, namelen) == 0)
+                else if (iequals("Macros", name[0 .. namelen]))
                     s = new MacroSection();
                 else
                     s = new Section();
@@ -1812,14 +1812,6 @@ extern (C++) int cmp(const(char)* stringz, const(void)* s, size_t slen)
     if (len1 != slen)
         return cast(int)(len1 - slen);
     return memcmp(stringz, s, slen);
-}
-
-extern (C++) int icmp(const(char)* stringz, const(void)* s, size_t slen)
-{
-    size_t len1 = strlen(stringz);
-    if (len1 != slen)
-        return cast(int)(len1 - slen);
-    return Port.memicmp(stringz, cast(char*)s, slen);
 }
 
 /*****************************************
--- a/src/dmd/doc.d
+++ b/src/dmd/doc.d
@@ -352,11 +352,11 @@ private TemplateDeclaration getEponymousParent(Dsymbol s)
     return (td && getEponymousMember(td)) ? td : null;
 }
 
-extern (C++) __gshared const(char)* ddoc_default = import("default_ddoc_theme.ddoc");
-extern (C++) __gshared const(char)* ddoc_decl_s = "$(DDOC_DECL ";
-extern (C++) __gshared const(char)* ddoc_decl_e = ")\n";
-extern (C++) __gshared const(char)* ddoc_decl_dd_s = "$(DDOC_DECL_DD ";
-extern (C++) __gshared const(char)* ddoc_decl_dd_e = ")\n";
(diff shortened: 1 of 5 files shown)


src/dmd/doc.d
```diff
@@ -1815,12 +1814,12 @@ extern (C++) int cmp(const(char)* stringz, const(void)* s, size_t slen)
     return memcmp(stringz, s, slen);
 }
 
-extern (C++) int icmp(const(char)* stringz, const(void)* s, size_t slen)
+private bool icmp(const(char)* stringz, const(char)* s, size_t slen)
```
jacob-carlborg: Isn’t the name a bit misleading now when it will only make a comparison.

Geod24: Definitely. Any suggestion ?

src/dmd/doc.d
```diff
@@ -117,13 +117,12 @@ extern (C++) class Section
                 "STANDARDS",
                 "THROWS",
                 "VERSION",
-                null
             ];
-            for (size_t i = 0; table[i]; i++)
+            foreach (const entry; table)
```
WalterBright: The `const` is redundant, as `table` is immutable.

src/dmd/doc.d
```diff
@@ -1815,12 +1814,12 @@ extern (C++) int cmp(const(char)* stringz, const(void)* s, size_t slen)
     return memcmp(stringz, s, slen);
 }
 
-extern (C++) int icmp(const(char)* stringz, const(void)* s, size_t slen)
+private bool icmp(const(char)* stringz, const(char)* s, size_t slen)
 {
     size_t len1 = strlen(stringz);
```
WalterBright: Your fixes make all the first arguments to `icmp` a string, but then they get coerced to pointers by the parameter type, and then strlen is run on them here. Just commit to having string parameters to icmp.

src/dmd/doc.d
```diff
@@ -1815,12 +1814,12 @@ extern (C++) int cmp(const(char)* stringz, const(void)* s, size_t slen)
...
     if (len1 != slen)
-        return cast(int)(len1 - slen);
-    return Port.memicmp(stringz, cast(char*)s, slen);
+        return false;
+    return Port.memicmp(stringz, s, slen) == 0;
```
WalterBright: Give up on pointer-based memicmp and just write a loop.

src/dmd/errors.d
```diff
@@ -73,10 +58,7 @@ extern (D) void error(Loc loc, const(char)* format, ...)
  */
 extern (C++) void error(const(char)* filename, uint linnum, uint charnum, const(char)* format, ...)
 {
-    Loc loc;
-    loc.filename = filename;
-    loc.linnum = linnum;
-    loc.charnum = charnum;
+    Loc loc = Loc(filename, linnum, charnum);
```
WalterBright: const

src/dmd/doc.d
jacob-carlborg: `equal`, `iequal`, `equali`, `equalCaseInsensitive`.

Geod24:
Updated:
- Addressed Walter's comment
- Renamed/refactored `icmp` into `dmd.root.port.Port.iequals` and use slicing at call site

src/dmd/root/port.d
```diff
@@ -54,6 +54,34 @@ extern (C++) struct Port
...
+       Compare two slices for equality, in a case-insensitive way
+
+       Params:
+         s1 = string to compare
+         s2 = string to compare
+
+       Returns:
+         `true` if `s1 == s2` regardless of case
+     */
+    extern(D) static int iequals(const char[] s1, const char[] s2)
```
jacob-carlborg: * Returns `int` instead of `bool`
* Is if worth mentioning in the documentation that this only works with ASCII (as far as I can tell)?

Geod24:
Updated according to Jacob's comments:
```diff
diff --git a/src/dmd/root/port.d b/src/dmd/root/port.d
index ea52b617b..01ef74d53 100644
--- a/src/dmd/root/port.d
+++ b/src/dmd/root/port.d
@@ -57,6 +57,9 @@ extern (C++) struct Port
     /**
        Compare two slices for equality, in a case-insensitive way

+       Comparison is based on `char` and does not do decoding.
+       As a result, it's only really accurate for plain ASCII strings.
+
        Params:
          s1 = string to compare
          s2 = string to compare
@@ -64,7 +67,7 @@ extern (C++) struct Port
        Returns:
          `true` if `s1 == s2` regardless of case
      */
-    extern(D) static int iequals(const char[] s1, const char[] s2)
+    extern(D) static bool iequals(const(char)[] s1, const(char)[] s2)
     {
         if (s1.length != s2.length)
             return false;
```

jacob-carlborg (approved):

src/dmd/doc.d
```diff
@@ -117,13 +117,12 @@ extern (C++) class Section
...
-                null
             ];
-            for (size_t i = 0; table[i]; i++)
+            foreach (entry; table)
             {
-                if (icmp(table[i], name, namelen) == 0)
+                if (Port.iequals(entry, name[0 .. namelen]))
                 {
-                    buf.printf("$(DDOC_%s ", table[i]);
+                    buf.printf("$(DDOC_%.*s ", entry.length, entry.ptr);
```
WalterBright: When using `.*s`, you must cast the `entry.length` parameter to `int`, because `entry.length` is a `size_t`.
Also, this change to `.*s` doesn't buy anything anyway, since the string is guaranteed to be 0 terminated.

Geod24: I reverted this bit to `.ptr`

Printing is the main issue I encountered when converting pointers to slices, and I think we need a more convenient way to print slices for it to move forward.

src/dmd/root/port.d
```diff
@@ -54,6 +54,37 @@ extern (C++) struct Port
...
+       As a result, it's only really accurate for plain ASCII strings.
+
+       Params:
+         s1 = string to compare
+         s2 = string to compare
+
+       Returns:
+         `true` if `s1 == s2` regardless of case
+     */
+    extern(D) static bool iequals(const(char)[] s1, const(char)[] s2)
```
klickverbot: Nitpick: `Port` was originally intended to encapsulate platform-specific code, hence the name.

Geod24: I was a bit surprised too, but decided to go with consistency.
Moved `iequals` to `utils`

src/dmd/dinifile.d
```diff
@@ -351,7 +351,7 @@ void parseConfFile(StringTable* environment, const(char)* filename, const(char)*
                 {
                     if (!writeToEnv(environment, strdup(pn)))
                     {
-                        error(Loc(filename, lineNum, 0), "Use `NAME=value` syntax, not `%s`", pn);
+                        error(filename, lineNum, 0, "Use `NAME=value` syntax, not `%s`", pn);
```
klickverbot: This is imho a regression in legibility/code quality.

Geod24: You're right, I only considered the fact that this overload was used in only one place.
However changing this is a bit more involved and would trigger other changes, so I simply removed this commit from this PR and will do the changes in a separate PR.

klickverbot: Thanks!

src/dmd/doc.d
```diff
@@ -1833,20 +1833,25 @@ extern (C++) bool isDitto(const(char)* comment)
...
+{
+    foreach (idx, char c; p)
     {
-        switch (*p)
+        switch (c)
         {
         case ' ':
         case '\t':
         case '\n':
-            continue;
```
klickverbot: This is personal taste, but I found the `continue` version easier to read, as it can't be confused with breaking out of the loop.

src/dmd/imphint.d
```diff
@@ -23,7 +23,9 @@ import core.stdc.string;
  */
 extern (C++) const(char)* importHint(const(char)* s)
 {
-    return hints.get(cast(string) s[0..strlen(s)], null).ptr;
+    if (auto ptr = s.toDString() in hints)
+        return (*ptr).ptr;
```
klickverbot: Not sure if anything containing the snippet `(*ptr).ptr` can qualify as an improvement…

src/dmd/link.d
```diff
@@ -112,11 +112,11 @@ version (Posix)
     {
         version (OSX)
         {
-            __gshared const(char)* nmeErrorMessage = "`__Dmain`, referenced from:";
+            static immutable nmeErrorMessage = "`__Dmain`, referenced from:";
```
klickverbot: Imho a raw pointer is more appropriate as long as it's only used as a C string as it encodes the presence of a null character by convention. Again just a question of style, though.

dnadlinger (approved):

Geod24:
Updated:
```diff
diff --git a/src/dmd/imphint.d b/src/dmd/imphint.d
index 95f6319f0..57631750a 100644
--- a/src/dmd/imphint.d
+++ b/src/dmd/imphint.d
@@ -23,8 +23,8 @@ import dmd.utils;
  */
 extern (C++) const(char)* importHint(const(char)* s)
 {
-    if (auto ptr = s.toDString() in hints)
-        return (*ptr).ptr;
+    if (auto entry = s.toDString() in hints)
+        return entry.ptr;
     return null;
 }
diff --git a/src/dmd/doc.d b/src/dmd/doc.d
index 60b9b724d..44fc81782 100644
--- a/src/dmd/doc.d
+++ b/src/dmd/doc.d
@@ -1846,7 +1846,7 @@ extern (D) const(char)[] skipwhitespace(const(char)[] p)
         case ' ':
         case '\t':
         case '\n':
-            break;
+            continue;
         default:
             return p[idx .. $];
         }
diff --git a/src/dmd/link.d b/src/dmd/link.d
index 3334d2e0a..2d67e08e0 100644
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -112,11 +112,11 @@ version (Posix)
     {
         version (OSX)
         {
-            static immutable nmeErrorMessage = "`__Dmain`, referenced from:";
+            static immutable(char*) nmeErrorMessage = "`__Dmain`, referenced from:";
         }
         else
         {
-            static immutable nmeErrorMessage = "undefined reference to `_Dmain`";
+            static immutable(char*) nmeErrorMessage = "undefined reference to `_Dmain`";
         }
         FILE* stream = fdopen(fd, "r");
         if (stream is null)
@@ -136,7 +136,7 @@ version (Posix)
             const(char)* lastSep = strrchr(buffer.ptr, '\n');
             if (lastSep)
                 buffer[(end = lastSep - &buffer[0])] = '\0';
-            if (strstr(&buffer[0], nmeErrorMessage.ptr))
+            if (strstr(&buffer[0], nmeErrorMessage))
                 nmeFound = true;
             if (lastSep)
                 buffer[end++] = '\n';
```

JinShil (approved):

Geod24:
Well this is green with auto-merge but dlang bot doesn't care, so merging manually

PR #8593, opened 2018-08-22
simplify loadconst() for LDC build

rainers:


WalterBright:
While this is great, unfortunately there are no tests to ensure it actually works, i.e. are the special x87 instructions actually generated? I don't know at the moment how to put it in the test suite, but if you could write a function that operates on those values, compile it, and dump the assembler to see if the FLDZ, FLD1, FLDPI, FLDL2T, FLDL2E, FLDLG2, FLDLN2 opcodes are generated that would work for now.

rainers:
> While this is great, unfortunately there are no tests to ensure it actually works, i.e. are the special x87 instructions actually generated?

Added test case. 

While this just restores the C version I wonder why your refactoring could be comitted without such a test case, though ;-)

test/runnable/fldconst.d
```diff
@@ -0,0 +1,46 @@
...
+{
+    static T fun() { return v; }
+    T f = fun();
+    assert(f == x); // to!string(x) ~ " " ~ to!string(f - x));
+    static assert(fun() == v);
+    auto fn = &fun;
+    auto s = *cast(ushort*) fn;
+    ushort expected = (op << 8) | 0xd9;
+
+    assert(s == expected); //, T.stringof ~ " " ~ to!string(v) ~ " " ~ to!string(s, 16) ~ " " ~ to!string(expected, 16));
```
WalterBright: sweet!

WalterBright (approved):

PR #8594, opened 2018-08-22
Fix debug build and remove debug messages

rainers:


src/dmd/backend/nteh.d
```diff
@@ -642,7 +642,8 @@ else
          */
         Symbol *s;
 
-        s = except_gensym();
+        version (SCPP) // except_gensym doesn't exist, but this code is not removed in debug builds
```
ZombineDev: Previously this whole block:
https://github.com/dlang/dmd/blob/v2.081.2/src/dmd/backend/nteh.c#L567-L610
was ifdef-ed. I think it would be more correct to wrap it a static if.

PetarKirov (approved):
LGTM with one comment addressed, feel free to auto-merge afterwards:

src/dmd/backend/nteh.d
rainers: > Previously this whole block:
> https://github.com/dlang/dmd/blob/v2.081.2/src/dmd/backend/nteh.c#L567-L610
> was ifdef-ed. I think it would be more correct to wrap it a static if.

Due to the mixture of compile time and runtime conditions in the if statement, adding a static if is not so easy (at least without duplication). I restored the C version but had to simulate the `else` with a `goto`.

RazvanN7 (approved):

PR #8596, opened 2018-08-22
build.d: Get path to vswhere

JinShil:
This gets the path to _vswhere_:  https://github.com/Microsoft/vswhere

_vswhere_ is a single executable for locating Visual Studio installations and components.  It's installed automatically by Visual Studio 2017 v15.2+ to a standard location at `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe`.  It is also downloadable from the GitHub repository linked above.

I intend to use it at https://github.com/dlang/dmd/blob/3d289d3ea100bda6a2b25c5e43779d7dd487281f/src/vcbuild/msvc-dmc.d#L17-L21 and https://github.com/dlang/dmd/blob/3d289d3ea100bda6a2b25c5e43779d7dd487281f/src/vcbuild/msvc-lib.d#L14-L16 I think that code has a typo:  notice that the path begins with `\Program Files` without a drive letter.  Anyway, that code doesn't even work on my computer, so I can't get a build.  I think `vswhere` will be a better solution for locating `cl.exe` and `lib.exe`.

_msvc-dmc_ and _msvc_lib_ will check the environment variable `HOST_VSWHERE` and execute it to locate  `cl.exe` and `lib.exe`.  _build.d_ is responsible for setting `HOST_VSWHERE`, and that's what this PR does.

cc @rainers for your expertise

### DIFF
--- a/src/build.d
+++ b/src/build.d
@@ -809,6 +809,41 @@ auto sourceFiles()
 }
 
 /*
+Downloads a file from a given URL
+
+Params:
+    to    = Location to store the file downloaded
+    from  = The URL to the file to download
+    tries = The number of times to try if an attempt to download fails
+Returns: `true` if download succeeded
+*/
+bool download(string to, string from, uint tries = 3)
+{
+    import std.net.curl : download, HTTPStatusException;
+    
+    foreach(i; 0..tries)
+    {
+        try
+        {
+            log("Downloading %s ...", from);
+            download(from, to);
+            return true;
+        }
+        catch(HTTPStatusException e) 
+        {
+            if (e.status == 404) throw e;
+            else 
+            {
+                log("Failed to download %s (Attempt %s of %s)", from, i + 1, tries);
+                continue;
+            }
+        }
+    }
+
+    return false;
+}
+
+/*
 Detects the host OS.
 
 Returns: a string from `{windows, osx,linux,freebsd,openbsd,netbsd,dragonflybsd,solaris}`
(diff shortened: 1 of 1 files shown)


src/build.d
```diff
@@ -913,6 +916,61 @@ auto getHostDMDPath(string hostDmd)
...
+            catch(HTTPStatusException e) 
+            {
+                if (e.status == 404) throw e;
+                else 
+                {
+                    log("Failed to download %s (Attempt %s of %s)", url, i + 1, tries);
+                    continue;
+                }
+            }
+        }
```
wilzbach: Better wrap this in a function as we can reuse it for a cross-platform download of the bootstrap compiler too, then.

JinShil: Done.  I didn't change the existing bootstrap compiler download to avoid introducing issues.  I can always do that as a followup.

src/build.d
```diff
@@ -913,6 +951,43 @@ auto getHostDMDPath(string hostDmd)
...
+    {
+        // Check if vswhere.exe can be found in the host's PATH
+        const where = ["where", "vswhere"].execute;
+        if (where.status == 0)
+            return where.output;
+
+        // Check if vswhere.exe is in the standard location
+        auto standardPath = ["cmd", "/C", "echo", `%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe`].execute.output;
+        standardPath = standardPath.replace(`"`, "");    // Remove quotes surrounding the path
+        standardPath = standardPath.replace("\r\n", ""); // Remove trailing newline characters
```
jacob-carlborg: If the functions are chained `standardPath` could be made const.

jacob-carlborg (approved):

src/build.d
```diff
@@ -913,6 +951,44 @@ auto getHostDMDPath(string hostDmd)
...
+        const outputPath = outputFolder.buildPath("vswhere").exeName;
+        if (outputPath.exists)
+            return outputPath;
+
+        // try to download it
+        if (download(outputPath, "https://github.com/Microsoft/vswhere/releases/download/2.5.2/vswhere.exe"))
+            return outputPath;
+
+        // Could not find or obtain vswhere.exe
+        throw new Exception("Could not obtain vswhere.exe.  Consider downloading it from https://github.com/Microsoft/vswhere and placing it in your PATH");
```
jacob-carlborg: There's an extra space before `Consider`.

PetarKirov (approved):

PR #8597, opened 2018-08-22
Fix Issue 19185 - [ICE] Nested struct segfaults when using variable from outer scope

RazvanN7:
The context pointer was not initialized when when the nested struct was constructed implicitly.

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -7239,6 +7239,42 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
         result = e.expressionSemantic(sc);
     }
 
+    /*
+     * Get the expression initializer for a specific struct
+     *
+     * Params:
+     *  sd = the struct for which the expression initializer is needed
+     *  loc = the location of the initializer
+     *  sc = the scope where the expression is located
+     *  t = the type of the expression
+     *
+     * Returns:
+     *  The expression initializer or error expression if any errors occured
+     */
+    private Expression getInitExp(StructDeclaration sd, Loc loc, Scope* sc, Type t)
+    {
+        if (sd.zeroInit && !sd.isNested())
+        {
+            // https://issues.dlang.org/show_bug.cgi?id=14606
+            // Always use BlitExp for the special expression: (struct = 0)
+            return new IntegerExp(loc, 0, Type.tint32);
+        }
+
+        if (sd.isNested())
+        {
+            auto sle = new StructLiteralExp(loc, sd, null, t);
+            if (!sd.fill(loc, sle.elements, true))
+                return new ErrorExp();
+            if (checkFrameAccess(loc, sc, sd, sle.elements.dim))
+                return new ErrorExp();
+
+            sle.type = t;
+            return sle;
+        }
+
+        return t.defaultInit(loc);
+    }
+
     override void visit(AssignExp exp)
     {
         static if (LOGSEMANTIC)
@@ -7651,28 +7687,13 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                          * variable with a bit copy of the default
                          * initializer
                          */
-                        Expression einit;
-                        if (sd.zeroInit && !sd.isNested())
+                        Expression einit = getInitExp(sd, exp.loc, sc, t1);
+                        if (einit.op == TOK.error)
                         {
-                            // https://issues.dlang.org/show_bug.cgi?id=14606
-                            // Always use BlitExp for the special expression: (struct = 0)
-                            einit = new IntegerExp(exp.loc, 0, Type.tint32);
+                            result = einit;
+                            return;
                         }
-                        else if (sd.isNested())
-                        {
-                            auto sle = new StructLiteralExp(exp.loc, sd, null, t1);
-                            if (!sd.fill(exp.loc, sle.elements, true))
-                                return setError();
-                            if (checkFrameAccess(exp.loc, sc, sd, sle.elements.dim))
-                                return setError();
 
-                            sle.type = t1;
-                            einit = sle;
-                        }
-                        else
-                        {
-                            einit = t1.defaultInit(exp.loc);
-                        }
                         auto ae = new BlitExp(exp.loc, exp.e1, einit);
                         ae.type = e1x.type;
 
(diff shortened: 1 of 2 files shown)


src/dmd/expressionsem.d
```diff
@@ -6287,6 +6287,42 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+     * Params:
+     *  sd = the struct for which the expression initializer is needed
+     *  loc = the location of the initializer
+     *  sc = the scope where the expression is located
+     *  t1 = the type of the expression
+     *
+     * Returns:
+     *  The expression initializer
+     */
+    private Expression getInitExp(StructDeclaration sd, Loc loc, Scope* sc, Type t)
```
JinShil: I wonder if functions like this should actually be member functions of the class.

src/dmd/expressionsem.d
```diff
@@ -6287,6 +6287,42 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+     *
+     * Params:
+     *  sd = the struct for which the expression initializer is needed
+     *  loc = the location of the initializer
+     *  sc = the scope where the expression is located
+     *  t1 = the type of the expression
+     *
+     * Returns:
+     *  The expression initializer
+     */
```
JinShil: If we want to be strict about the style (https://dlang.org/dstyle.html#phobos_documentation) we should not prepend `*` characters on each line.  But I don't think it should hold up this PR.

jacob-carlborg:
Perhaps because I’m reading this on my phone but I don’t see what actually fixed the issue.

src/dmd/expressionsem.d
RazvanN7: It could be made a nested function inside visit(AssignExp)

src/dmd/expressionsem.d
```diff
@@ -6287,6 +6287,42 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+    private Expression getInitExp(StructDeclaration sd, Loc loc, Scope* sc, Type t)
+    {
+        if (sd.zeroInit && !sd.isNested())
+        {
+            // https://issues.dlang.org/show_bug.cgi?id=14606
+            // Always use BlitExp for the special expression: (struct = 0)
+            return new IntegerExp(loc, 0, Type.tint32);
+        }
+
+        if (sd.isNested())
```
RazvanN7: @jacob-carlborg This block of code is executed. Before this patch, it would have not.

src/dmd/expressionsem.d
```diff
@@ -6287,6 +6287,42 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
     }
 
+    /*
+     * Get the expression initializer for a specific struct
+     *
+     * Params:
+     *  sd = the struct for which the expression initializer is needed
+     *  loc = the location of the initializer
+     *  sc = the scope where the expression is located
+     *  t1 = the type of the expression
```
n8sh: t1 => t

src/dmd/expressionsem.d
n8sh: The remark about not using leading `*` in documentation was added to the style guide in January of this year and reflects Phobos but doesn't reflect the style of the existing DMD codebase. IMO the most important thing is maintaining consistency within a document.

src/dmd/expressionsem.d
```diff
@@ -6833,8 +6851,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                             }
                         }
 
-                        Expression einit;
-                        einit = new BlitExp(exp.loc, e1x, e1x.type.defaultInit(exp.loc));
+                        Expression einit = getInitExp(sd, exp.loc, sc, t1);
+                        einit = new BlitExp(exp.loc, e1x, einit);
```
n8sh: Note to other reviewers: it's fine that this PR replaces `e1x.type` with  `t1` because here `e1x` is `exp.e1` and `t1` is `exp.e1.type.toBasetype()`.

n8sh (approved):

src/dmd/expressionsem.d
WalterBright: Note that `einit` is assigned twice. A better method is to assign it only once.

WalterBright (approved):

src/dmd/expressionsem.d
RazvanN7: Fixed.

src/dmd/expressionsem.d
```diff
@@ -6287,6 +6287,42 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+     * Get the expression initializer for a specific struct
+     *
+     * Params:
+     *  sd = the struct for which the expression initializer is needed
+     *  loc = the location of the initializer
+     *  sc = the scope where the expression is located
+     *  t = the type of the expression
+     *
+     * Returns:
+     *  The expression initializer
```
JinShil: I think this should mention that it might return `null` and why.

JinShil: Or, better yet, shouldn't this return an `ErrorExp` and then check for `TOK.error` in the caller?

MartinNowak (dismissed):
I agree with @JinShil here, getInitExp should return an ErrorExp instead of null.

src/dmd/expressionsem.d
```diff
@@ -6833,8 +6851,7 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                             }
                         }
 
-                        Expression einit;
-                        einit = new BlitExp(exp.loc, e1x, e1x.type.defaultInit(exp.loc));
+                        Expression einit = new BlitExp(exp.loc, e1x, getInitExp(sd, exp.loc, sc, t1));
```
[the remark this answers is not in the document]
MartinNowak: Better yet, `getInitExp` should just return an ErrorExp and thus turn any use of such expressions into an error as well.

src/dmd/expressionsem.d
```diff
@@ -6700,28 +6736,10 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
-                        else if (sd.isNested())
-                        {
-                            auto sle = new StructLiteralExp(exp.loc, sd, null, t1);
-                            if (!sd.fill(exp.loc, sle.elements, true))
-                                return setError();
-                            if (checkFrameAccess(exp.loc, sc, sd, sle.elements.dim))
-                                return setError();
+                        Expression einit = getInitExp(sd, exp.loc, sc, t1);
+                        if (einit.op == TOK.error)
+                            return setError();
```
JinShil: This is redundant.  `getInitExp` returned an `ErrorExp` and then `setError` immediately creates a new `ErrorExp` and assigns it to result.  I see few options:

- create an overload of `setError` to take an `ErrorExp`
- instead of `return setError` do `result = einit; return;`

Maybe there is another solution, but those are the two that come to mind for me.

RazvanN7: Done

JinShil (approved):

PR #8598, opened 2018-08-22
Fix Issue 19182 - missing semicolon crashes compiler

RazvanN7:
The grammar for pragma statements states that a pragma is either followed by a `;` or a statement, however, predefined pragma's like `msg`, `startaddress` and `lib`, as far as I could tell, are always ended with a `;`. The fix eagerly checks that the `;` is presented at parse time.

### DIFF
--- a/test/fail_compilation/fail19182.d
+++ b/test/fail_compilation/fail19182.d
@@ -1,0 +1,18 @@
+// REQUIRED_ARGS: -c
+/*
+TEST_OUTPUT:
+---
+gigi
+fail_compilation/fail19182.d(12): Error: `pragma(msg)` is missing a terminating `;`
+---
+*/
+
+void foo()
+{
+    pragma(msg, "gigi") // Here
+    static foreach (e; [])
+    {
+        pragma(msg, "lili");
+    }
+
+}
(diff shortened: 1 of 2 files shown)


BBasile:
It's an upstream regression (https://run.dlang.io/is/JJ7giv). Also the missing semicolon does not violate the grammar (the first pragma follows this rule https://dlang.org/spec/statement.html#PragmaStatement).
So i think that the problem is not handled as it should.

JinShil:
I'm not sure about this.  I wonder if it would be better to emit an error that the statement following the `pragma` is not relevant/permitted.  In the case of `msg` and friends, it's a bit moot, since no statement would be permitted, but it just seems that this should be done in the semantic pass rather than the parser.

test/fail_compilation/fail19182.d
```diff
@@ -0,0 +1,18 @@
+// REQUIRED_ARGS: -c
+/*
+TEST_OUTPUT:
+---
+gigi
```
jacob-carlborg: What is this doing here?

RazvanN7: Since we are in semantic, the pragma(msg) is ran before checking for the next instruction.

jacob-carlborg: Ah, ok.

jacob-carlborg (approved):

PetarKirov (approved):

PR #8599, opened 2018-08-23
Accept const(char)[] instead of string in dmd.frontend

Geod24:
```
We shouldn't use string when immutability is not required.
In this case, while the functions don't change their input,
they don't require it to never change either.
```

### DIFF
--- a/src/dmd/frontend.d
+++ b/src/dmd/frontend.d
@@ -54,7 +54,7 @@ Add import path to the `global.path`.
 Params:
     path = import to add
 */
-void addImport(string path)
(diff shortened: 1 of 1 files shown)


wilzbach (approved):

JinShil:
I don't know if this is such a good idea.  Doesn't it tell the caller that what they send to the function can potentially change; something they didn't need to consider when the input was `string`?  I think it changes the agreement between the function and it's caller, and opens possibilities that the caller otherwise wouldn't have been required to consider or handle.  Forgive me if I'm missing the point.

PR #8607, opened 2018-08-23
Fix Issue 18672 - Error in @safe transitive propagation with associative arrays

RazvanN7:
When `opAssign` is generated, the destructor has only been through semantic1, therefore no attributes were generated (an empty destructor that is not marked as @safe will be seen as @system). In order to overcome this, when opAssign is in semantic3, a check is performed to see if the destructor has been considered @safe in the meantime.

### DIFF
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -2919,20 +2919,24 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
             if (sc.func)
             {
                 /* If the nesting parent is pure without inference,
-                 * then this function defaults to pure too.
-                 *
-                 *  auto foo() pure {
-                 *    auto bar() {}     // become a weak purity function
-                 *    class C {         // nested class
-                 *      auto baz() {}   // become a weak purity function
-                 *    }
-                 *
-                 *    static auto boo() {}   // typed as impure
-                 *    // Even though, boo cannot call any impure functions.
-                 *    // See also Expression::checkPurity().
-                 *  }
+                   then this function defaults to pure too.
+
+                   auto foo() pure
+                   {
+                       auto bar() {}       // become a weak purity function
+                       class C             // nested class
+                       {
+                           auto baz() {}   // become a weak purity function
+                       }
+
+                       static auto boo() {}   // typed as impure
+                       // Even though, boo cannot call any impure functions.
+                       // See also Expression::checkPurity().
+                   }
+
+                   The same applies for @safe.
                  */
-                if (tf.purity == PURE.impure && (funcdecl.isNested() || funcdecl.isThis()))
+                if ((funcdecl.isNested() || funcdecl.isThis) && !funcdecl.isInstantiated)
                 {
                     FuncDeclaration fd = null;
                     for (Dsymbol p = funcdecl.toParent2(); p; p = p.toParent2())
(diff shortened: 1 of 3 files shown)


src/dmd/semantic3.d
```diff
@@ -241,6 +241,20 @@ private extern(C++) final class Semantic3Visitor : Visitor
         {
             if (funcdecl.storage_class & STC.inference)
             {
+                /* https://issues.dlang.org/show_bug.cgi?id=18672
+                   If the parent struct has a destructor that is
+                   `@safe`, then the `opAssign` function is `@trusted`
+                 */
+                auto sd = funcdecl.parent.isStructDeclaration();
```
jacob-carlborg: Shouldn't it check for classes as well?

src/dmd/semantic3.d
```diff
@@ -241,6 +241,20 @@ private extern(C++) final class Semantic3Visitor : Visitor
...
+                if (auto dtor = sd.dtor)
+                {
+                    if (dtor.isSafe())
```
JinShil: Please correct me if I'm wrong, but I don't think it should check if the destructor is `@safe`, but that the context in which the struct is declared is `@safe`.

JinShil: To elaborate, marking something as `@trusted` involves manual auditing.  How do we know that `opAssign` is memory-safe in this context.  And what if `opAssign` is written by the user?  Will that be automatically annotated as `@trusted`?

src/dmd/semantic3.d
RazvanN7: identity opAssign is not allowed for classes: https://dlang.org/spec/operatoroverloading.html

src/dmd/semantic3.d
RazvanN7: No, this test is solely for the generated opAssign. We know how the generated opAssign looks, because, well, the compiler generates it [1]. The body of the generated function looks like this:

```d
S __swap = void;
 __swap = this;   // bit copy
 this = s;        // bit copy
 __swap.dtor();
```

As you can see, the `S __swap = void` is unsafe, but it can be trusted. So if the destructor is @safe, we can trust the whole opAssign.

[1] https://github.com/dlang/dmd/blob/master/src/dmd/clone.d#L210

RazvanN7: >Please correct me if I'm wrong, but I don't think it should check if the destructor is @safe, but that the context in which the struct is declared is @safe

I see. I, wrongfully, thought that you can declare a `@system` function inside a `@safe` one. For example, on today's compiler, I can do this:

```d
void delegate() fp; 

void fun2() @safe
{

    void fun()
    {
        int a;
        int* b = &a;                  // totally unsafe                                                                                        
        import std.stdio : writeln;
        writeln("called system function inside safe context");
    }

    fp = &fun;
}

void main()
{
    fun2();
    fp();
}
```

However, the spec is very clear on this one: https://dlang.org/spec/function.html#safe-functions (point number 4). So it seems that the bug is not about `opAssign`, but a more general one: if the scope is @safe, then function declarations inside that scope should be @safe.

RazvanN7: After reading the spec more carefully, I could unveil the true nature of this bug: dmd fails to propagate safety to nested functions. It seems that attributes are inferred solely for templates and function literals, so the generated opAssign body should not be analyzed in order to establish if it is @safe or not. Therefore, the current fix is wrong and I will try to fix the bigger problem, which is propagating safety to nested functions. In the case of other attributes, the spec is unclear whether those should be propagated as well or not; I would assume that all attributes should be treated the same.

RazvanN7:
@JinShil I believe this is the right fix.

src/dmd/semantic3.d
jacob-carlborg: `opAssign` is allowed for classes if the parameter is of a different type than the class. I tried this and got the same error as for the struct example.

src/dmd/semantic3.d
jacob-carlborg: > It seems that attributes are inferred solely for templates and function literals

Yeah, that’s excepted, at least for user defined functions.

rainers:
> static assert(__traits(compiles, { fun(); })) // does not fail, but is fixed by this bug

Is this really wrong code? There's nothing wrong with creating a `@system` delegate as long as it is not called from `@safe` code. Maybe you meant to `()` after the function literal?

RazvanN7:
@rainers 

>Is this really wrong code? There's nothing wrong with creating a @system delegate as long as it is not called from @safe code. Maybe you meant to () after the function literal?

Maybe I'm misinterpreting  this, but the spec says: "Functions nested inside safe functions default to being safe functions.".  Even if I'm misinterpreting, the @safe tests in phobos that declare @system delegates still have to fixed.

rainers:
> but the spec says: "Functions nested inside safe functions default to being safe functions."

I might be wrong but I don't think that applies to function literals which have all their attributes inferred. There needs to be no *default* because inference always yields a result.

> Even if I'm misinterpreting, the @safe tests in phobos that declare @System delegates still have to fixed.

Yes, to verify `@safe` they should be calling the delegates by adding `()` after the closing braces.

RazvanN7:
@WalterBright What is your take on this?

PR #8612, opened 2018-08-24
Simplify number parsing in lexer, clean up error messages.

schveiguy:
This is a followup to #8451 to streamline the code in the lexer.

I've factored out all the checking of whether a digit is too big for the base into one place. I also fixed the messages to be always consistent based on the base instead of the digit being examined.

I wanted to go a step further and combine the 2 flags for hex and binary digits missing after the header, but curiously, D currently deprecates `0x_` but not `0b_`, so I'm not sure whether there's a valid reason for this or not. Relevant PR is #8396 

This should obsolete #8478

### DIFF
--- a/test/fail_compilation/fix19059.d
+++ b/test/fail_compilation/fix19059.d
@@ -2,10 +2,11 @@
  * PERMUTE_ARGS:
  * TEST_OUTPUT:
 ---
-fail_compilation/fix19059.d(16): Error: radix 8 digit expected, not `8`
-fail_compilation/fix19059.d(16): Error: octal literals `010` are no longer supported, use `std.conv.octal!10` instead
-fail_compilation/fix19059.d(17): Error: radix 8 digit expected, not `9`
-fail_compilation/fix19059.d(17): Error: octal literals `011` are no longer supported, use `std.conv.octal!11` instead
+fail_compilation/fix19059.d(17): Error: octal digit expected, not `8`
+fail_compilation/fix19059.d(17): Error: octal literals larger than 7 are no longer supported
+fail_compilation/fix19059.d(18): Error: octal digit expected, not `9`
+fail_compilation/fix19059.d(18): Error: octal literals larger than 7 are no longer supported
+fail_compilation/fix19059.d(19): Error: octal literals `010` are no longer supported, use `std.conv.octal!10` instead
 ---
  */
 
@@ -15,4 +16,5 @@ void foo()
 {
     auto a = 08;
     auto b = 09;
+    auto c = 010;
 }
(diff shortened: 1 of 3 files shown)


schveiguy:
Note: I'm expecting failures on the auto tester due to the changed error messages. Still not knowledgeable on how to run the test suite for dmd locally, so I'm waiting to see where the failures occur.

wilzbach:
> Still not knowledgeable on how to run the test suite for dmd locally, so I'm waiting to see where the failures occur.

Just execute `./test/run.d` - see its readme for more details.

schveiguy:
> Just execute ./test/run.d

I get non stop linker errors.

wilzbach:
Which ones?
Alternative you can just run the makefile there (the readme documents the available targets).

schveiguy:
```
ld: unknown option: -rpath=/Users/steves/git/dmd2/src/dmd-dev/test/../../phobos/generated/osx/release/64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I executed via `rdmd ./test/run.d`

test/fail_compilation/fix19059.d
```diff
@@ -2,10 +2,10 @@
...
  * TEST_OUTPUT:
 ---
-fail_compilation/fix19059.d(16): Error: radix 8 digit expected, not `8`
-fail_compilation/fix19059.d(16): Error: octal literals `010` are no longer supported, use `std.conv.octal!10` instead
-fail_compilation/fix19059.d(17): Error: radix 8 digit expected, not `9`
-fail_compilation/fix19059.d(17): Error: octal literals `011` are no longer supported, use `std.conv.octal!11` instead
+fail_compilation/fix19059.d(16): Error: octal digit expected, not `8`
+fail_compilation/fix19059.d(16): Error: octal literals larger than 7 are no longer supported
+fail_compilation/fix19059.d(17): Error: octal digit expected, not `9`
+fail_compilation/fix19059.d(17): Error: octal literals larger than 7 are no longer supported
```
WalterBright: Why did the `use std.conv.octal` disappear?

WalterBright: None of the test cases show the `use` message. Can you please add one?

schveiguy: I can add them. Note that the error messages exist in [other places](https://github.com/dlang/dmd/blob/79c8de1bdddfcd4c0338bde478d17eccb8edcd4b/test/fail_compilation/fail2656.d), this one test was added to check that `08` and `09` do not compile.

schveiguy:
Added test case for valid octal literal that is banned.

WalterBright:
@schveiguy Thanks!

PR #8613, opened 2018-08-24
Fix Issue 19187: check variable type is defined before accessing

FeepingCreature:


FeepingCreature:
Ah! Sorry about that massive review request, I accidentally rebased onto master rather than stable.

FeepingCreature:
Corrected the testcase: the error happens regardless of if the identifier is defined or not. It happens purely because `foo`, as an overload, has no clear type.

jacob-carlborg (approved):

FeepingCreature:
yay! thanks

PR #8618, opened 2018-08-26
Add support for VT100 color sequences on Windows 10

GoaLitiuM:
This addition would be really handy when combined with other build tools, where the color codes gets stripped out due to how traditional console coloring is applied to the console itself and not to the output.

### DIFF
--- a/src/dmd/console.d
+++ b/src/dmd/console.d
@@ -42,17 +42,39 @@ enum Color : int
     white         = bright | lightGray,
 }
 
+/* The ANSI escape codes are used.
+ * https://en.wikipedia.org/wiki/ANSI_escape_code
+ * Foreground colors: 30..37
+ * Background colors: 40..47
+ * Attributes:
+ *  0: reset all attributes
+ *  1: high intensity
+ *  2: low intensity
+ *  3: italic
+ *  4: single line underscore
+ *  5: slow blink
+ *  6: fast blink
+ *  7: reverse video
+ *  8: hidden
+ */
+
 struct Console
 {
     version (Windows)
     {
         import core.sys.windows.windows;
+        import core.sys.windows.wincon;
+        
+        // added in druntime#2285
+        static if (__VERSION__ < 2082)
+            enum ENABLE_VIRTUAL_TERMINAL_PROCESSING = 4;
 
       private:
         CONSOLE_SCREEN_BUFFER_INFO sbi;
         HANDLE handle;
         FILE* _fp;
-
+        bool vtmode;
+        bool resetVtmode;
       public:
 
         @property FILE* fp() { return _fp; }
@@ -97,11 +119,41 @@ struct Console
             if (GetConsoleScreenBufferInfo(h, &sbi) == 0) // get initial state of console
                 return null;
 
+            // try enabling support for ANSI escape codes
+            DWORD mode;
+            bool vtmode;
+            bool resetVtmode;
+            if (GetConsoleMode(h, &mode))
+            {
+                if (mode & ENABLE_VIRTUAL_TERMINAL_PROCESSING)
+                {
+                    // the flag was enabled already, do not touch it afterwards
+                    vtmode = true;
+                }
+                else if (SetConsoleMode(h, mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING))
+                {
+                    vtmode = true;
+                    resetVtmode = true;
+                }
+            }
+
             auto c = new Console();
             c._fp = fp;
             c.handle = h;
             c.sbi = sbi;
+            c.vtmode = vtmode;
+            c.resetVtmode = resetVtmode;
             return c;
+        }
+
+        ~this()
+        {
+            if (resetVtmode)
+            {
+                DWORD mode;
+                if (GetConsoleMode(handle, &mode))
+                    SetConsoleMode(handle, mode & ~ENABLE_VIRTUAL_TERMINAL_PROCESSING);
+            }
         }
 
         /*******************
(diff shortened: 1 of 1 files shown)


GoaLitiuM:
Depends on pending druntime change https://github.com/dlang/druntime/pull/2285

src/dmd/console.d
```diff
@@ -42,17 +42,39 @@ enum Color : int
...
+        static if (__VERSION__ < 2082)
+            enum ENABLE_VIRTUAL_TERMINAL_PROCESSING = 4;
 
       private:
         CONSOLE_SCREEN_BUFFER_INFO sbi;
         HANDLE handle;
         FILE* _fp;
-
+        bool vtmode;
+        bool resetVtmode;
```
WalterBright: What do these names mean?

WalterBright:
I don't get this. ANSI color sequences are already emitted (and they're VT100 sequences already).

src/dmd/console.d
```diff
@@ -42,17 +42,39 @@ enum Color : int
...
+
 struct Console
 {
     version (Windows)
     {
         import core.sys.windows.windows;
+        import core.sys.windows.wincon;
+        
+        // added in druntime#2285
+        static if (__VERSION__ < 2082)
```
CyberShadow: Why not `static if (!is(typeof(ENABLE_VIRTUAL_TERMINAL_PROCESSING)))` ?

CyberShadow:
> I don't get this. ANSI color sequences are already emitted (and they're VT100 sequences already).

This is for Windows 10, which has some support for them. The old code was only for Posix.

src/dmd/console.d
```diff
@@ -121,14 +176,19 @@ struct Console
...
-            enum FOREGROUND_WHITE = FOREGROUND_RED | FOREGROUND_GREEN | FOREGROUND_BLUE;
-            WORD attr = sbi.wAttributes;
-            attr = (attr & ~(FOREGROUND_WHITE | FOREGROUND_INTENSITY)) |
-                   ((color & Color.red)    ? FOREGROUND_RED   : 0) |
-                   ((color & Color.green)  ? FOREGROUND_GREEN : 0) |
-                   ((color & Color.blue)   ? FOREGROUND_BLUE  : 0) |
-                   ((color & Color.bright) ? FOREGROUND_INTENSITY : 0);
-            SetConsoleTextAttribute(handle, attr);
+            if (vtmode)
+                fprintf(_fp, "\033[%d;%dm", color & Color.bright ? 1 : 0, 30 + (color & ~Color.bright));
```
CyberShadow: Not sure about this duplication of code.

Perhaps define a PosixConsole and WindowsConsole (Windows-only), and decide whether to use WindowsConsole or PosixConsole depending on the console flag?

src/dmd/console.d
```diff
@@ -97,21 +119,54 @@ struct Console
...
+            bool vtmode;
+            bool resetVtmode;
+            if (GetConsoleMode(h, &mode))
+            {
+                if (mode & ENABLE_VIRTUAL_TERMINAL_PROCESSING)
+                {
+                    // the flag was enabled already, do not touch it afterwards
+                    vtmode = true;
+                }
+                else if (SetConsoleMode(h, mode | ENABLE_VIRTUAL_TERMINAL_PROCESSING))
```
CyberShadow: Why?

Furthermore, if the compiler crashes, we will have changed the setting of something with a longer lifetime that the compiler (the console window).

src/dmd/console.d
```diff
@@ -97,21 +119,54 @@ struct Console
             if (GetConsoleScreenBufferInfo(h, &sbi) == 0) // get initial state of console
                 return null;
 
+            // try enabling support for ANSI escape codes
+            DWORD mode;
+            bool vtmode;
+            bool resetVtmode;
+            if (GetConsoleMode(h, &mode))
+            {
+                if (mode & ENABLE_VIRTUAL_TERMINAL_PROCESSING)
```
CyberShadow: This looks like a heuristic at best. Your PR description says: "combined with other build tools", which I understood to mean tools that capture output and display it later (like Ninja), or IDEs which don't run things in a console window at all. 

In which case:

- Enabling VT processing, outputting VT codes, disabling VT processing, and allowing the tool to print them will just dump unprocessed VT codes to the user.
- If the output is redirected to a file, then this check (or changing console settings) is pointless.
- Same as when output is sent to a stream, to be displayed in an IDE output pane.

Perhaps allow enabling VT sequence output using `-color=vt` ?

src/dmd/console.d
CyberShadow: `vtmode`: whether the console currently understands VT sequences.
`resetVtmode`: whether we need to turn off VT mode in the console when the compiler exits.

src/dmd/console.d
GoaLitiuM: I just noticed this PR does not help with the color codes with tools like Ninja, because the colors are not added when no valid Console is present (isatty fails, GetConsoleMode fails, etc.). I think the best way to deal with this is to have ``-color=vt`` or ``-color=always`` which enables the colors in all cases.

src/dmd/console.d
GoaLitiuM: Or maybe something like AnsiConsole and WindowsConsole (legacy)?

src/dmd/console.d
GoaLitiuM: I don't think it's required to turn off the `ENABLE_VIRTUAL_TERMINAL_PROCESSING` flag, but I thought it would be common courtesy turn off something we enabled.

src/dmd/console.d
WalterBright: These would certainly be helpful as comments in the code.

PR #8621, opened 2018-08-27
Issue 19193 - deprecated attribute on enum members in templates is ignored

JinShil:


jacob-carlborg (approved):

JinShil:
> Could you please test is this also fixes 19202?

I already did.  They appear to be different issues.  Note that the 19202 regression was introduced long before the enum member attributes feature was added.

PR #8622, opened 2018-08-27
Remove CRLF newlines from cppb.h

ibuclaw:
The mixed used of both CR and CRLF just makes it horrible to edit.


JinShil (approved):

WalterBright:
My git commit script runs `tolf` first.

https://github.com/dlang/tools/blob/master/tolf.d

PR #8624, opened 2018-08-27
build.d: Get paths to MSVC's cl.exe and lib.exe for Windows

JinShil:


src/build.d
```diff
@@ -987,6 +996,38 @@ version(Windows)
         // Could not find or obtain vswhere.exe
         throw new Exception("Could not obtain vswhere.exe. Consider downloading it from https://github.com/Microsoft/vswhere and placing it in your PATH");
     }
+
+    /*
```
Geod24: Nit: This is not DDOC (need at least 2 stars)

Geod24 (approved):
LGTM. Feel free to merge it yourself with or without the nitpick.

src/build.d
```diff
@@ -987,6 +996,38 @@ version(Windows)
...
+            .replace("\r\n", "");
+
+        if (!vsInstallPath.exists)
+            throw new Exception("Could not locate the Visual Studio installation directory");
+
+        const vcVersionFile = vsInstallPath.buildPath("VC", "Auxiliary", "Build", "Microsoft.VCToolsVersion.default.txt");
+        if (!vcVersionFile.exists)
+            throw new Exception(`Could not locate the Visual C++ version file "%s"`.format(vcVersionFile));
+
+        const vcVersion = ["cmd", "/C", "type", `%s`.format(vcVersionFile)].execute.output.replace("\r\n", "");
```
rainers: Why execute a rather expensive `cmd /c type <versionFile>` instead of just `std.file.read(versionFile)`?

src/build.d
JinShil: Actually all functions could use proper DDoc.  I'll address that holistically in a followup PR.

src/build.d
JinShil: Because sometimes I'm a fool.  Fixed.

PR #8626, opened 2018-08-27
Backend: Import dmd.errors.error where it is used

Geod24:
I'm not 100% sure of this one, but I don't see the reason to avoid this import in D.
There is also [this chunk of code](https://github.com/dlang/dmd/blob/ad921f0c42cf8f7cd430503d40402e7bd18d18c2/src/dmd/backend/cgobj.c#L32-L50) in a c++ file but [it's unused](https://github.com/dlang/dmd/blob/ad921f0c42cf8f7cd430503d40402e7bd18d18c2/src/dmd/backend/cgobj.c#L2367) so I didn't bother.

### DIFF
--- a/src/dmd/backend/cgelem.d
+++ b/src/dmd/backend/cgelem.d
@@ -73,7 +73,7 @@ int REGSIZE();
 
 version (MARS)
 {
-extern void error(const(char)* filename, uint linnum, uint charnum, const(char)* format, ...);
+    import dmd.backend.errors;
 }
 
 /*****************************
--- a/src/dmd/backend/cgxmm.d
+++ b/src/dmd/backend/cgxmm.d
@@ -35,15 +35,14 @@ import dmd.backend.xmm;
 
 version (SCPP)
     import dmd.backend.exh;
-
+version (MARS)
+    import dmd.backend.errors;
 
 extern (C++):
 
 int REGSIZE();
 
 uint mask(uint m);
-
-void error(const(char)* filename, uint linnum, uint charnum, const(char)* format, ...);
 
 /*******************************************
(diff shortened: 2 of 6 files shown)


jacob-carlborg (approved):

ZombineDev:
CircleCI no_pic fails with:
```
  (HOST_DMD_RUN)  BACK_DOBJS  dmd/backend/evalu8.d
dmd -c -of../generated/linux/debug/32/evalu8.o  -version=MARS  -J../generated/linux/debug/32 -w -de -g -debug -g -m32 -mv=dmd.backend=dmd/backend -betterC dmd/backend/evalu8.d
dmd/root/array.d(19): Error: `TypeInfo` cannot be used with -betterC
posix.mak:552: recipe for target '../generated/linux/debug/32/evalu8.o' failed
make: *** [../generated/linux/debug/32/evalu8.o] Error 1
make: Leaving directory '/home/circleci/dmd/src'
Exited with code 2
```

I suppose that's because the backend is compiled with `-betterC`, while the frontend without. But I wonder why the issue manifests doesn't manifest itself on the CircleCI PIC - if anything it's a compile-time error, not link time.

Geod24:
That error is actually legit. I am trying another approach, let's see what the CI thinks of it. Will add explanations if it goes ✅

ZombineDev:
I wonder if a selective import would have the same effect as a di header.

PetarKirov (approved):
LGTM, though perhaps @WalterBright should confirm if the import should be `version (MARS)`, as before the extern declaration was not versioned.

Geod24:
> the extern declaration was not versioned.

But its usages are

PR #8627, opened 2018-08-27
Fix Issue 19199 - Use core.bitops intrinsics during CTFE

n8sh:
Currently in `dmd.builtin` there are software implementations of `core.bitop` functions within `eval_bsf`, `eval_bsr`, `eval_bswap`, and `eval_popcnt`. With DMD written in D this is unnecessary: they can use `core.bitop`.

### DIFF
--- a/src/dmd/builtin.d
+++ b/src/dmd/builtin.d
@@ -24,6 +24,7 @@ import dmd.mtype;
 import dmd.root.ctfloat;
 import dmd.root.stringtable;
 import dmd.tokens;
+static import core.bitop;
 
 private:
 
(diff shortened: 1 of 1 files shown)


src/dmd/builtin.d
```diff
@@ -259,45 +253,27 @@ extern (C++) Expression eval_bsr(Loc loc, FuncDeclaration fd, Expressions* argum
...
     Expression arg0 = (*arguments)[0];
     assert(arg0.op == TOK.int64);
     uinteger_t n = arg0.toInteger();
-    int cnt = 0;
-    while (n)
-    {
-        cnt += (n & 1);
-        n >>= 1;
-    }
-    return new IntegerExp(loc, cnt, arg0.type);
```
n8sh: This type for the result of `eval_popcnt` -- `arg0.type` -- was incorrect.  The result of `popcnt` is always `int`. See:

https://github.com/dlang/druntime/blob/db2910ee7425f91dc6fe38f149f4b8eaa3b88f7f/src/core/bitop.d#L561
https://github.com/dlang/druntime/blob/db2910ee7425f91dc6fe38f149f4b8eaa3b88f7f/src/core/bitop.d#L593

JinShil:
cc @UplinkCoder

PetarKirov (approved):

UplinkCoder:
Not sure if the swap code is correct here?
seems to be missing the htons case ?

n8sh:
@UplinkCoder If I understand correctly `eval_bswap` is only used for `core.bitop.bswap` and there isn't an overload of `bswap` that accepts a `ushort` like htons. The only appearance of `eval_bswap` that shows up with _grep_ other than its definition is:
https://github.com/dlang/dmd/blob/63ef858bc76b43b590cd9503ddf841b735312435/src/dmd/builtin.d#L467-L468
I'm not up on my D mangling, but it looks like this might even be just the 32-bit overload of `bswap`.

UplinkCoder:
You are right.
nevermind then ;)

UplinkCoder (approved):
LGTM

jacob-carlborg (approved):

PR #8628, opened 2018-08-28
convert cod2.c to D

WalterBright:


JinShil (approved):
There are a number of things in this PR that are not direct translations, but I have complete faith in Walter.  The direct translations look good.

WalterBright:
> There are a number of things in this PR that are not direct translations

Hmm, example?

WalterBright:
> The changes in code.d (not sure where that code came from) and the additions in code_x86.d, but also a couple things that were uncommented for reasons I don't know.

They all came from their corresponding .h files. The uncommented ones were ones I hadn't converted to D yet, but are now needed. Some of the extra stuff was needed for the cod4.d translation, because I wanted to minimize the need for rebasing after merging this. (It turns out git still did a botched merge of code.d, which I had to fix.)

PR #8630, opened 2018-08-28
Don explains exception chaining

WalterBright:


JinShil (approved):

Geod24 (approved):
Would it be something that we could, potentially, under the right conditions, deprecate in the long run?

PR #8633, opened 2018-08-28
Rework #8542 wrt. deterministic IDs via generateIdWithLoc()

kinke:


src/dmd/globals.d
```diff
@@ -469,6 +470,24 @@ nothrow:
...
+    }
+
+    extern (D) hash_t toHash() const pure
+    {
+        import dmd.root.hash : mixHash;
+
+        auto hash = hashOf(linnum);
+        hash = mixHash(hash, hashOf(charnum));
+        if (filename !is null)
+            hash = mixHash(hash, hashOf(filename[0 .. strlen(filename)]));
```
kinke: Please tell me there's something like `mixHash()` in druntime and an existing helper function converting a C string to a D one.

kinke: Custom hashing is necessary because of mixins, e.g.:
```
test/runnable/test18880.d-mixin-7 (0x558d3e709930)
test/runnable/test18880.d-mixin-7 (0x558d3e70a120)
test/runnable/test18880.d-mixin-7 (0x558d3e70a3f0)
```

Btw, 2.081.1 isn't able to compile master, that's pretty bad.

kinke:
Custom hashing of `Loc` is necessary because of mixins, where the `const(char)* Loc.filename` pointers diverge, e.g.:
```
test/runnable/test18880.d-mixin-7 (0x558d3e709930)
test/runnable/test18880.d-mixin-7 (0x558d3e70a120)
test/runnable/test18880.d-mixin-7 (0x558d3e70a3f0)
```

Btw, 2.081.1 isn't able to compile master, that's pretty bad (I went back to 2.077).

src/dmd/globals.d
```diff
@@ -469,6 +469,24 @@ nothrow:
...
+    }
+
+    extern (D) hash_t toHash() const pure
+    {
+        import dmd.root.hash;
+        import dmd.utils;
+
+        auto hash = hashOf(linnum);
+        hash = mixHash(hash, hashOf(charnum));
+        hash = mixHash(hash, hashOf(filename.toDString));
```
kinke: Please tell me there's something like mixHash() in druntime.

kinke:
Some remarks:

1) Hashing `Loc` is case-sensitive wrt. file path on Windows too, unlike `opEquals()`. So that's not really correct.
2) The existing `equals()` method ignoring charnum without explicit `-vcolumns` is pretty weird IMO; it'd be better renamed to make that clear.
3) `Loc.filename` should probably be a proper `dmd.root.filename.Filename` struct (contains a single `const(char)*` pointer). And that should probably get a proper `opEquals()` and `toHash()` instead of the `Loc`. If it contained a D string instead of the C one, it would probably just work out of the box without explicit implementations, at the (memory) cost of an additional `size_t` per `Loc`.
4) The generated IDs aren't really unique per module, but rather per filename (same as current master; there should be no functional changes at all with this PR). The mixin example above illustrates that (=> artificial filenames with extra suffix). Not sure if there can thus still be collisions between the actual module (`foo.d`) and a mixin (`foo.d-mixin-<N>`); identifier, line and column would have to match for that to happen.

ibuclaw:
> If it contained a D string instead of the C one, it would probably just work out of the box without explicit implementations, at the (memory) cost of an additional size_t per Loc.

I can't see this happening, Loc is a critical object used everywhere. Increasing its size is not really an available option.

src/dmd/identifier.d
```diff
@@ -125,18 +125,6 @@ nothrow:
...
-       A secondary string table is used to guarantee that we generate unique
-       identifiers per module. See generateIdWithLoc and issues
-       https://issues.dlang.org/show_bug.cgi?id=16995
-       https://issues.dlang.org/show_bug.cgi?id=18097
-       https://issues.dlang.org/show_bug.cgi?id=18111
-       https://issues.dlang.org/show_bug.cgi?id=18880
-       https://issues.dlang.org/show_bug.cgi?id=18868
-       https://issues.dlang.org/show_bug.cgi?id=19058.
-     */
-    private extern (C++) __gshared StringTable fullPathStringTable;
```
ibuclaw: This is the fifth time that someone has completely reimplemented this, right?

kinke: Probably. But this is only an attempt to simplify/accelerate the current version, without functional changes.

Geod24:
> I can't see this happening, Loc is a critical object used everywhere. Increasing its size is not really an available option.

Well that's exactly the goal I've been working towards recently...

ibuclaw:
> Well that's exactly the goal I've been working towards recently...
I'll just leave this review here: https://github.com/dlang/dmd/pull/7821#discussion_r167157872

jacob-carlborg:
> I can't see this happening, Loc is a critical object used everywhere. Increasing its size is not really an available option.

Isn't `Loc` mostly passed by `const ref`? Or is that not your concern?

BTW, in Clan the corresponding type is only 32 bit large. The documentation says:

"Technically, a source location is simply an offset into the manager's view of the input source"

[1] https://clang.llvm.org/doxygen/classclang_1_1SourceLocation.html

src/dmd/globals.d
n8sh: druntime's explicit `mixHash` function [was removed October 14, 2014](https://github.com/dlang/druntime/commit/6da8b3b367531670bcc1e8d1ad3579737cfef7bd#diff-1cb07fafd18644ca9974aea7a74483e9) but druntime `hashOf` has a two-argument form that can be used for this:
```d
hash = hashOf(charnum, hash);
```

ibuclaw:
> Isn't Loc mostly passed by const ref? Or is that not your concern?

It's not how it's passed around, increasing the size of Loc increases the size of every object in the front end.

If we want memory consumption to go in any direction, we would want to keep it down.

> BTW, in Clan the corresponding type is only 32 bit large. The documentation says:

The same is also true for GCC's location_t.

Having a location map might be reasonable to ensure that each location is only allocated once. You could make it more useful however by storing location ranges - start and end of all tokens and expressions.

I digress from the topic of this PR however...

kinke:
`toHash()` and `opEquals()` are now consistent (both case-sensitive on Windows). I don't really want to create an uppercase copy each time the hash is requested on Windows, and assume that once we operate with `Loc` instances, there are no case-insensitive duplicate filenames to worry about. That assumption is probably not valid for generic `FileName`s, that's why I left it in `Loc`. `equals()` should probably be case-sensitive then too actually.

If you think modifying ubiquitous `Loc` may have too much impact, I can move custom hashing and equivalence to the `Key` struct.

src/dmd/globals.d
```diff
@@ -462,13 +462,45 @@ nothrow:
         return buf.extractString();
     }
 
+    /* Checks for equivalence,
+     * a) comparing the filename contents (not the pointer), case-
+     *    insensitively on Windows, and
```
jacob-carlborg: FYI, macOS also has a case insensitive filesystem by default.

kinke: Ah, interesting. The `FileName.equals()` function used here only has a Windows special case.

src/dmd/identifier.d
```diff
@@ -166,75 +154,53 @@ nothrow:
...
-            OutBuffer fullPathIdBuf;
-
-            if (absPath)
-            {
-                // replace characters that demangle can't handle
-                for (auto ptr = absPath; *ptr != '\0'; ++ptr)
+            // 2.082+
+            counters.update(Key(loc, prefix),
+                () => 1u,          // insertion
+                (ref uint counter) // update
```
kinke: I had to use `ref` here, although it makes no sense in this case. To be fair, I haven't tried returning void and incrementing the ref directly.

kinke:
Rebased + squashed. Can someone remove the WIP label please?

jacob-carlborg:
> Can someone remove the WIP label please?

Done.

kinke:
@thewilsonator: Please try to push this. It was only supposed to rectify the IMO horrible code introduced with the important fix in #8542; I don't see why it has to rot here for more than half a year now.

> If you think modifying ubiquitous Loc may have too much impact, I can move custom hashing and equivalence to the Key struct.

That 'offer' still stands and is IMO the only question mark; I'll be on vacation though.

thewilsonator (approved):

thewilsonator:
I've got it in the AGM at dconf to make `Loc` an offset into a source manager, that would make the problem of making `Loc` bigger moot.

PR #8634, opened 2018-08-29
Fix issue 19203 - alias this to a bool behaves both inconsistently and incorrectly with static assert

thewilsonator:


src/dmd/staticcond.d
```diff
@@ -93,6 +94,8 @@ bool evalStaticCondition(Scope* sc, Expression exp, Expression e, ref bool error
         return false;
     }
 
+    e = resolveAliasThis(sc,e);
```
ZombineDev: Mind the style: `e = resolveAliasThis(sc, e);`

PetarKirov (approved):

test/compilable/test19203.d
```diff
@@ -0,0 +1,27 @@
...
+        enum hasPopBack = BoolWithErr(false, T.stringof~" does not have popBack");
+    else
+        enum hasPopBack = BoolWithErr(true,"");
+}
+
+void test()
+{
+    static assert( hasPopBack!Foo);
+    static assert(!hasPopBack!Bar);
+    static assert( hasPopBack!Foo && !hasPopBack!Bar);
```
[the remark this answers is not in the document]
thewilsonator: Because before it would complain with `foo && bar` if both were `alias this` to a bool then the result of the expression is a bool, but would fail because `foo` was not a bool.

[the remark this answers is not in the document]
thewilsonator: See the bug report for more detail.

src/dmd/staticcond.d
carun: Why not use dfmt or as part of precommit hook?

ZombineDev: It's a good idea in general, I'm not sure if dfmt is prosuction ready for enforcing dmd's code style. This seems like a good experiment to try if you have time.

carun: I tried this a while ago to get dfmt integrated with phobos. But I stumbled upon an edge case which I couldn't figure out, so I dropped it. Let me try it again.

RazvanN7 (approved):

RazvanN7:
@thewilsonator Please add the whitespace that @ZombineDev requested and we are good to go

PR #8637, opened 2018-08-29
doc: Re-add C++ linkage back to gendocfile

ibuclaw:
As this is called from the compiler main function.

### DIFF
--- a/src/dmd/doc.h
+++ b/src/dmd/doc.h
@@ -1,0 +1,22 @@
+
+/* Compiler implementation of the D programming language
+ * Copyright (C) 1999-2018 by The D Language Foundation, All Rights Reserved
+ * written by Walter Bright
+ * http://www.digitalmars.com
+ * Distributed under the Boost Software License, Version 1.0.
+ * http://www.boost.org/LICENSE_1_0.txt
+ * https://github.com/dlang/dmd/blob/master/src/dmd/doc.h
+ */
+
+#ifndef DMD_DOC_H
+#define DMD_DOC_H
+
+#ifdef __DMC__
+#pragma once
+#endif /* __DMC__ */
+
+class Module;
+
+void gendocfile(Module *m);
+
+#endif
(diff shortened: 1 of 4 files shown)


Geod24 (changes requested):
If it's called from C++, can you add it to a header file?

ibuclaw:
For the docs module, this would be `gendocfile` at the top and public, the rest of the functions and declarations that follow are `package`, everything else that is already `private` gets grouped together in the appropriate place - probably the bottom.

jacob-carlborg:
> In this case, gendocfile is only used by mars.d - which is NOT a front end module. :-)

So why does it need to be `extern(C++)`? I thought `mars.d` was only used by DMD.

ibuclaw:
@jacob-carlborg - Because dmd is not the only compiler using the front end...

jacob-carlborg:
@ibuclaw you said `mars.d` is not a front end module :wink:.

Geod24:
@ibuclaw : If it is `extern(C++)`, it's because it has to be called from C++... If it has to be called from C++, it should be in the header. Yes they are not accurate at the moment, but there's no reason to keep it that way.

ibuclaw:
@Geod24 - created a lonely file.

ibuclaw:
> you said mars.d is not a front end module wink.

That is correct, it is not a front end module.  The best label to give it would be the driver, parts of the compiler that deal with inputs and outputs.  Other modules with the same label include dinifile, libXXX, link, and scanXXX.

src/dmd/doc.h
```diff
@@ -0,0 +1,22 @@
...
+ * Distributed under the Boost Software License, Version 1.0.
+ * http://www.boost.org/LICENSE_1_0.txt
+ * https://github.com/dlang/dmd/blob/master/src/dmd/doc.h
+ */
+
+#ifndef DMD_DOC_H
+#define DMD_DOC_H
+
+#ifdef __DMC__
+#pragma once
```
Geod24: Isn't pragma once implemented by all compilers? :D

Geod24 (approved):
LGTM

src/dmd/doc.h
ibuclaw: Let's continue this in #8641

PR #8638, opened 2018-08-30
Remove C++ linkage from CTFE functions

ibuclaw:
Sizeable exodus of what is an internal implementation detail.

@Geod24 an example of what I was referring to. :-)

### DIFF
--- a/src/dmd/constfold.d
+++ b/src/dmd/constfold.d
@@ -47,7 +47,7 @@ private Expression expType(Type type, Expression e)
  * Returns:
  *    true if e is a constant
  */
-extern (C++) int isConst(Expression e)
+int isConst(Expression e)
 {
     //printf("Expression::isConst(): %s\n", e.toChars());
     switch (e.op)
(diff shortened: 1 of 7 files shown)


jacob-carlborg (approved):

Geod24 (approved):
LGTM in principle, although it would be nice of one of the LDC dev could take a look at this.

ZombineDev:
@kinke what do you think?

kinke:
I guess this is fine for us as well; we have a bunch of additional CTFE builtins in `dmd.builtin`, but that seems unaffected.

Geod24:
@ibuclaw: Feel free to self merge once you have fixed the Windows only failure.

ibuclaw:
@kinke - maybe I should have clarified that these are specific to the CTFE engine.  It's implementation details should not concern you.

ibuclaw:
@Geod24 - Win-what?

Geod24:
@ibuclaw Are you done closing and opening the door ?

ibuclaw:
@Geod24 - Win32 machine seems to have lack of available temporary files. :-)

Geod24:
In this case you can just deprecate the test on the autotester.
AFAIK Open/close is only useful for AppVeyor and semaphoreci if you want to re-trigger a test without pushing (and loosing the auto-tester result).

ZombineDev:
Recently I've had to deprecate the test results on the Win32 power machine on a few other PRs for the same reason - failure to create a temp file during the phobos tests, so it's definitely not a problem with this PR.

@braddr could you take a look if this is something that could be fixed on your side?

PR #8641, opened 2018-08-30
Replace all header guards with pragma once

ibuclaw:
~Added guards where missing, and corrected names to be `DMD_NAME_H`~

~Though maybe we should drop DMC's `#pragma once`, I'm not sure.~

After discussing both here and with gcc maintainers, using `#pragma once` everywhere then.

### DIFF
--- a/src/dmd/aggregate.h
+++ b/src/dmd/aggregate.h
@@ -8,12 +8,7 @@
  * https://github.com/dlang/dmd/blob/master/src/dmd/aggregate.h
  */
 
-#ifndef DMD_AGGREGATE_H
-#define DMD_AGGREGATE_H
-
-#ifdef __DMC__
 #pragma once
-#endif /* __DMC__ */
 
 #include "root/root.h"
 
@@ -350,5 +345,3 @@ public:
     InterfaceDeclaration *isInterfaceDeclaration() { return this; }
     void accept(Visitor *v) { v->visit(this); }
 };
-
-#endif /* DMD_AGGREGATE_H */
--- a/src/dmd/aliasthis.h
+++ b/src/dmd/aliasthis.h
@@ -8,12 +8,7 @@
  * https://github.com/dlang/dmd/blob/master/src/dmd/aliasthis.h
  */
 
-#ifndef DMD_ALIASTHIS_H
-#define DMD_ALIASTHIS_H
-
-#ifdef __DMC__
 #pragma once
-#endif /* __DMC__ */
 
 #include "mars.h"
 #include "dsymbol.h"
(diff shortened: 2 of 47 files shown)


Geod24:
Isn't pragma once prefered nowadays? I don't mind one way or another, especially since those will be auto generated... One day.

dkgroot:
@ibuclaw Why not drop all the "ifdef guards" and only use "pragma once" instead ?
- Shorter :+1: Less Code
- Avoidance of name clashes
- Portable (nowadays): See [Pragma once portability](https://en.wikipedia.org/wiki/Pragma_once#Portability). 
- Doubling pragma and ifdef guards is not a good idea and looks strange/odd.
- Performance: On most compilers that support the once pragma will even improve be able to compiler a little faster by ensuring a file is only opened/loaded once (some compilers are able to do the same for ifdef guards).

If you need some help converting some of the other files, i would be glad to help.

ibuclaw:
I've forwarded the question to the GCC maintainers, first response was that `#pragma once` should be fine for non-default languages such as ourselves.

dkgroot:
gcc 2.95 did have support for "#pragma once" but it was deprecated temporarily (for some reason(s)) until gcc 3.4... So using 2.95 should actually work, but might produce a number of 'deprecation'-warnings. Maybe you should give it a try and see :-)

ibuclaw:
> And if ~14 years is not ample time to update your machine/compiler, nothing will help. Better tell the user that logo might be a better fit for them and be done with it :-)

Directly behind me, is an IBM workstation with AIX, a MIPS workstation with IRIX, and an Xserve G5 1U server.

At some point, I should find them a home and attempt building gdc on them. ;-)

ibuclaw:
> I've forwarded the question to the GCC maintainers, first response was that #pragma once should be fine for non-default languages such as ourselves.

And in general, they try to support 4.1 at least.

ibuclaw:
Remove all header guards from front-end headers.  Leaving the backend alone.

Geod24 (approved):

PetarKirov (approved):

ZombineDev:
@ibuclaw I think this needs a rebase.

ibuclaw:
@ZombineDev - ok. Was waiting for the others to go in first, then get back to this.

Geod24:
I'm not really bothered by this, but it would be nice to respect the 72h label next time.

ZombineDev:
Excuse me @Geod24, I thought that we all reached consensus on the approach, so I didn't see a reason for waiting. I decided to go ahead with merging in order to help move PRs in the queue faster.

ibuclaw:
Can we add time-delayed auto merge functionality to dlang bot when a PR is labelled as 72h no objection?

PR #8644, opened 2018-08-31
Fix Hack: Remove hasCheckedAttrs from VarExp

JinShil:
`hasCheckedAttrs` was a member variable of `VarExp` introduced in https://github.com/dlang/dmd/pull/8404 to avoid having the deprecation message printed multiple times as `expressionSemantic` could have been called on the same `VarExp` multiple times. 

While working on a bug fix recently, I discovered that we could avoid that hack by checking for deprecation in `EnumMember.GetVarExp`, as is done with `checkDisabled`.

This PR removes the hack and makes the code more consistent with the existing implementation.

### DIFF
--- a/src/dmd/denum.d
+++ b/src/dmd/denum.d
@@ -415,6 +415,11 @@ extern (C++) final class EnumMember : VarDeclaration
         if (errors)
             return new ErrorExp();
         checkDisabled(loc, sc);
+
+        if (depdecl && !depdecl._scope)
+            depdecl._scope = sc;
+        checkDeprecated(loc, sc);
+
         if (errors)
             return new ErrorExp();
         Expression e = new VarExp(loc, this);
--- a/src/dmd/expression.d
+++ b/src/dmd/expression.d
@@ -3347,11 +3347,6 @@ extern (C++) final class SymOffExp : SymbolExp
  */
 extern (C++) final class VarExp : SymbolExp
 {
-    /**
-    * Semantic can be called multiple times for a single expression.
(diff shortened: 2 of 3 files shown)


jacob-carlborg (approved):

src/dmd/denum.d
```diff
@@ -415,6 +415,11 @@ extern (C++) final class EnumMember : VarDeclaration
         if (errors)
             return new ErrorExp();
         checkDisabled(loc, sc);
+
+        if (depdecl && !depdecl._scope)
+            depdecl._scope = sc;
```
ZombineDev: Took me a while to see that `DeprecatedDeclaration._scope` is used here:
https://github.com/dlang/dmd/blob/58fde675fb69ea57da8902ddbabf87cf8b74e1f9/src/dmd/dsymbolsem.d#L410-L428
which is called from here:
https://github.com/dlang/dmd/blob/ce5e17727e43d60991091fc96fbe5fc0d7435334/src/dmd/dsymbol.d#L300-L324

PetarKirov (approved):

ZombineDev:
From https://github.com/dlang/dmd/pull/8645#issuecomment-417654394:
> Hmm. Looks suspiciously familiar (#8644). I'd like to wait for that to make it to stable first.

Did you intend to target stable with this PR?

JinShil:
> Did you intend to target stable with this PR?

No, it's not a critical PR.  I know my other PR is waiting on it, but I think it can wait until the next master --> stable merge.  I'm not even sure if there's a real dependency yet.

PR #8645, opened 2018-08-31
Fix Issue 19202 - deprecated eponymous template prints no warning

JinShil:


JinShil:
Hmm. Looks suspiciously familiar (https://github.com/dlang/dmd/pull/8644).  I'd like to wait for that to make it to stable first.

PR #8649, opened 2018-09-01
build.d:  Run vcvars{arch}.bat and copy MSVC environment variables to current environment

JinShil:


JinShil:
ping
cc @rainers

src/build.d
```diff
@@ -1019,13 +1013,64 @@ version(Windows)
...
+        const vcVarsFilename = "vcvars" ~ model ~ ".bat";
+        const vcVars = vcInstallPath.buildPath("Auxiliary", "Build", vcVarsFilename);
+        if (!vcVars.exists)
+            throw new Exception("Could not locate the Visual C++ " ~ vcVarsFilename ~ " file");
+
+        // run vcvars{arch}.bat and return the resulting environment variables with the `set` command
+        const vsVars = ["cmd", "/C", vcVars, "&&", "set"].execute.output;
+
+        // copy the environment variables we need to the current environment
+        const varsToCopy = ["INCLUDE", "LIBPATH"];
```
rainers: For compilation INCLUDE should be enough, but if you want to link aswell, you need LIB, not LIBPATH. You will probably also need to get some updates to PATH as some DLLs might not be found, e.g. with /DEBUG.

Why not just copy the full environment?

src/build.d
```diff
@@ -1019,13 +1013,64 @@ version(Windows)
...
+        // See https://github.com/Microsoft/vswhere/wiki/Find-VC
+
+        const vcVersionFile = vcInstallPath.buildPath("Auxiliary", "Build", "Microsoft.VCToolsVersion.default.txt");
         if (!vcVersionFile.exists)
             throw new Exception(`Could not locate the Visual C++ version file "%s"`.format(vcVersionFile));
 
         const vcVersion = vcVersionFile.readText().replace("\r\n", "");
         const vcArch = model == "64" ? "x64" : "x86";
-        const vcPath = vsInstallPath.buildPath("VC", "Tools", "MSVC", vcVersion, "bin", "Host" ~ vcArch, vcArch);
+        const vcPath = vcInstallPath.buildPath("Tools", "MSVC", vcVersion, "bin", "Host" ~ vcArch, vcArch);
```
rainers: BTW: This path is very specific to VS2017, older versions work differently. The auto-tester runs VS2013.

src/build.d
JinShil: > Why not just copy the full environment?

I don't know if there will be conflicts, and I just want to avoid problems and noise.

src/build.d
JinShil: What do you suggest?

rainers: Using vswhere is specific to VS2017 to begin with, so it's ok to build the path to VC this way. It's a lot simpler to read it (up to vcVersion) from the environment variable `VCToolsInstallDir`, though.

If you want to support more than VS2017, I'd rather use the detection in dmd.link.

PR #8650, opened 2018-09-01
[build.d] Make 'unittest' target pass on OSX

Geod24:
Honestly no idea why it was not failing on POSIX.
The second commit fixes the issue, the first one is just something I came across while looking at the failure.

### DIFF
--- a/src/build.d
+++ b/src/build.d
@@ -367,8 +367,11 @@ auto configFiles()
 Main build routine for the DMD compiler.
 Defines the required order for the build dependencies, runs all these dependency dependencies
 and afterwards builds the DMD compiler.
+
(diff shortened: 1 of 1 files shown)


src/build.d
```diff
@@ -183,7 +183,9 @@ auto dmdConf()
 {
     // TODO: add support for Windows
     string exportDynamic;
-    version(OSX) {} else
+    version (OSX)
+        exportDynamic = " -L--export_dynamic"; // Note the underscore
```
jacob-carlborg: This has never been needed before, as far as I know.

RazvanN7 (approved):

PR #8652, opened 2018-09-02
Fixes Issue 19212 - Add versions for C++ runtimes.

TurkeyMan:
Necessary for `std::` stuff

Spec: https://github.com/dlang/dlang.org/pull/2461

### DIFF
--- a/src/dmd/cond.d
+++ b/src/dmd/cond.d
@@ -653,6 +653,11 @@ extern (C++) final class VersionCondition : DVCondition
             case "CRuntime_Microsoft":
             case "CRuntime_Musl":
             case "CRuntime_UClibc":
+            case "CppRuntime_Clang":
+            case "CppRuntime_DigitalMars":
+            case "CppRuntime_Gcc":
+            case "CppRuntime_Microsoft":
+            case "CppRuntime_Sun":
             case "unittest":
             case "assert":
             case "all":
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1307,18 +1307,30 @@ void addDefaultVersionIdentifiers(const ref Param params)
     if (params.isWindows)
     {
         VersionCondition.addPredefinedGlobalIdent("Windows");
+        if (global.params.mscoff)
+        {
+            VersionCondition.addPredefinedGlobalIdent("CRuntime_Microsoft");
+            VersionCondition.addPredefinedGlobalIdent("CppRuntime_Microsoft");
+        }
+        else
+        {
+            VersionCondition.addPredefinedGlobalIdent("CRuntime_DigitalMars");
+            VersionCondition.addPredefinedGlobalIdent("CppRuntime_DigitalMars");
+        }
     }
     else if (params.isLinux)
     {
         VersionCondition.addPredefinedGlobalIdent("Posix");
         VersionCondition.addPredefinedGlobalIdent("linux");
         VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        VersionCondition.addPredefinedGlobalIdent("CRuntime_Glibc");
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_Gcc");
     }
     else if (params.isOSX)
     {
         VersionCondition.addPredefinedGlobalIdent("Posix");
         VersionCondition.addPredefinedGlobalIdent("OSX");
-
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_Clang");
         // For legacy compatibility
         VersionCondition.addPredefinedGlobalIdent("darwin");
     }
@@ -1327,24 +1339,28 @@ void addDefaultVersionIdentifiers(const ref Param params)
         VersionCondition.addPredefinedGlobalIdent("Posix");
         VersionCondition.addPredefinedGlobalIdent("FreeBSD");
         VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_Clang");
     }
     else if (params.isOpenBSD)
     {
         VersionCondition.addPredefinedGlobalIdent("Posix");
         VersionCondition.addPredefinedGlobalIdent("OpenBSD");
         VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_Gcc");
     }
     else if (params.isDragonFlyBSD)
     {
         VersionCondition.addPredefinedGlobalIdent("Posix");
         VersionCondition.addPredefinedGlobalIdent("DragonFlyBSD");
         VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_Gcc");
     }
     else if (params.isSolaris)
     {
         VersionCondition.addPredefinedGlobalIdent("Posix");
         VersionCondition.addPredefinedGlobalIdent("Solaris");
         VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_Sun");
     }
     else
     {
@@ -1381,18 +1397,6 @@ void addDefaultVersionIdentifiers(const ref Param params)
         {
             VersionCondition.addPredefinedGlobalIdent("Win32");
         }
-    }
-
-    if (params.isWindows)
-    {
-        if (params.mscoff)
-            VersionCondition.addPredefinedGlobalIdent("CRuntime_Microsoft");
-        else
-            VersionCondition.addPredefinedGlobalIdent("CRuntime_DigitalMars");
-    }
-    else if (params.isLinux)
-    {
-        VersionCondition.addPredefinedGlobalIdent("CRuntime_Glibc");
     }
 
     if (params.isLP64)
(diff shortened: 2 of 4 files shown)


thewilsonator:
This will presumably need a spec PR to reserve those versions, and add them to the test that check reserved versions.

TurkeyMan:
Spec where?
I didn't find any such test...

thewilsonator:
`test/fail_compilation/reserved_version.d` and `test/fail_compilation/reserved_version_switch.d`

TurkeyMan:
When you say 'spec', do you mean this: https://github.com/dlang/dlang.org/pull/2461?

jpf91:
Spec: https://dlang.org/spec/version.html#predefined-versions

TurkeyMan:
That was with the initial PR... it's references at the top.

TurkeyMan:
Errr... what's with the CI? O_o

jpf91:
> That was with the initial PR... it's referenced at the top.

Oh right, I didn't see that. Sorry!

test/fail_compilation/reserved_version.d
```diff
@@ -83,6 +83,10 @@ fail_compilation/reserved_version.d(182): Error: version identifier `CRuntime_Gl
 fail_compilation/reserved_version.d(183): Error: version identifier `CRuntime_Microsoft` is reserved and cannot be set
 fail_compilation/reserved_version.d(184): Error: version identifier `CRuntime_Musl` is reserved and cannot be set
 fail_compilation/reserved_version.d(185): Error: version identifier `CRuntime_UClibc` is reserved and cannot be set
+fail_compilation/reserved_version.d(185): Error: version identifier `CppRuntime_DigitalMars` is reserved and cannot be set
+fail_compilation/reserved_version.d(185): Error: version identifier `CppRuntime_LibCpp` is reserved and cannot be set
+fail_compilation/reserved_version.d(185): Error: version identifier `CppRuntime_Microsoft` is reserved and cannot be set
+fail_compilation/reserved_version.d(185): Error: version identifier `CppRuntime_StdCpp` is reserved and cannot be set
```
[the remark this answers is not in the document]
TurkeyMan: Oh wow, solame.

src/dmd/cond.d
```diff
@@ -653,6 +653,10 @@ extern (C++) final class VersionCondition : DVCondition
             case "CRuntime_Microsoft":
             case "CRuntime_Musl":
             case "CRuntime_UClibc":
+            case "CppRuntime_DigitalMars":
+            case "CppRuntime_LibCpp":
```
jacob-carlborg: This is never used as far as I can see.

jacob-carlborg (changes requested):

src/dmd/cond.d
jacob-carlborg: > OSX? Clang distributes libc++ as standard no?

It's used on macOS yes.

> Oh, yeah.. Does DMD have any builds that build against such a C++ toolset?

Yes, on macOS.

jacob-carlborg: But for the C standard library on macOS there is no version identifier. In the code `OSX` is used.

TurkeyMan: Why isn't there some lines configuring DMD for OSX right there next to the lines I edited?
What should the proper patch for OSX look like? Can you add a patch for OSX?

TurkeyMan: Added for OSX + FreeBSD, they seem to be the 2 platforms that use it.

src/dmd/mars.d
```diff
@@ -1361,13 +1361,25 @@ void addDefaultVersionIdentifiers()
...
     }
     else static if (TARGET.Linux)
     {
         VersionCondition.addPredefinedGlobalIdent("CRuntime_Glibc");
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_StdCpp");
+    }
+    else static if (TARGET.OSX || TARGET.FreeBSD)
+    {
+        // why u no CRuntime guys?
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_LibCpp");
```
TurkeyMan: As far as I could tell, these are the only guys that use clang as default toolchain...

TurkeyMan: @jacob-carlborg Does this look correct to you?

TurkeyMan: Probably not on macos, but certainly on linux, and even windows (mingw/cygwin). Toolchains could be built to target all sorts of combinations. LDC/GDC need the freedom to define these however the platform desires.

jacob-carlborg: I'm specifically talking about macOS here. I'll let someone else decide if this is necessary or not.

TurkeyMan: I don't want to be writing:
```D
version(CppRuntime_LibCpp)
  version = CppRuntime_LibCpp_or_OSX;
else version (OSX)
  version = CppRuntime_LibCpp_or_OSX;
version (CppRuntime_LibCpp_or_OSX)
{
  // libc++ code...
}
```

jacob-carlborg: You would be writing:

```d
version (CppRuntime_StdCpp)
else version (OSX)
else version (FreeBSD)
```

That's how it's done for the C standard library: https://github.com/dlang/druntime/blob/master/src/core/stdc/locale.d#L60-L137.

ibuclaw: What about using `CppRuntime_GCC` and `CppRuntime_Clang`?

If I were to review code with `StdCpp` and `LibCpp` I would certainly have a hard time remembering the difference.

ibuclaw: > I'm specifically talking about macOS here.

C++ STL is tied to the compiler toolchain, not the OS.

jacob-carlborg: Fine. What about the rest of the platforms?

TurkeyMan: Rest of the platforms accounted.

TurkeyMan:
I changed the names to Gcc/Clang and populated all targets.

TurkeyMan:
This one is blocking all `std::`, the others are blockers to `vector`/`string`. Can this PR has priority?

Geod24 (approved):
LGTM

jacob-carlborg (approved):

jacob-carlborg:
Since this is a new feature I think this should have an explicit changelog entry. Feel free to add that in a follow up PR.

src/dmd/mars.d
ibuclaw: >  That's how it's done for the C standard library: https://github.com/dlang/druntime/blob/master/src/core/stdc/locale.d#L60-L137.

Speaking of which, why don't we just add `CRuntime_OSX` and `CRuntime_BSD`?

I take it that FreeBSD, DragonFlyBSD, OpenBSD, ..., InsertNameHereBSD all share the same libc away.

RazvanN7:
Resolve the conflict and we're good to go!

thewilsonator:
@jacob-carlborg this still has a merge conflict...

src/dmd/mars.d
```diff
@@ -1307,18 +1307,30 @@ void addDefaultVersionIdentifiers(const ref Param params)
...
+            VersionCondition.addPredefinedGlobalIdent("CppRuntime_DigitalMars");
+        }
     }
     else if (params.isLinux)
     {
         VersionCondition.addPredefinedGlobalIdent("Posix");
         VersionCondition.addPredefinedGlobalIdent("linux");
         VersionCondition.addPredefinedGlobalIdent("ELFv1");
+        VersionCondition.addPredefinedGlobalIdent("CRuntime_Glibc");
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_Gcc");
```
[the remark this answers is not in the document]
ibuclaw: The C standard library is a standalone library, in which you can mix any combination of compiler and runtime - on Linux you also have `CRuntime_Musl`, `CRuntime_UClibc`, and `CRuntime_Bionic`.

The C++ standard library is tied to, and shipped with the compiler.

src/dmd/mars.d
```diff
@@ -1307,18 +1307,30 @@ void addDefaultVersionIdentifiers(const ref Param params)
...
     }
     else if (params.isOSX)
     {
         VersionCondition.addPredefinedGlobalIdent("Posix");
         VersionCondition.addPredefinedGlobalIdent("OSX");
-
+        VersionCondition.addPredefinedGlobalIdent("CppRuntime_Clang");
```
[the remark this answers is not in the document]
TurkeyMan: I wondered that myself.

src/dmd/mars.d
```diff
@@ -1307,18 +1307,30 @@ void addDefaultVersionIdentifiers(const ref Param params)
     if (params.isWindows)
     {
         VersionCondition.addPredefinedGlobalIdent("Windows");
+        if (global.params.mscoff)
+        {
+            VersionCondition.addPredefinedGlobalIdent("CRuntime_Microsoft");
+            VersionCondition.addPredefinedGlobalIdent("CppRuntime_Microsoft");
```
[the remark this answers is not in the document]
TurkeyMan: I initially did use the CRuntime to imply the C++ runtime, but I added the CppRuntime in response to feedback on the `std::array` PR. The reason was that while Windows has a strong connection, other platforms do not.

MSVC C and C++ runtimes are tightly coupled, and glibc typically implies libstdc++, but C runtimes eg; bionic, musl, uclibc, all don't infer a C++ runtime. Android offers 2 or 3 C++ runtimes to select from, and libc++ is relatively new and not mandated by LLVM toolchains.

Embedded toolchains are likely to use some combination of embedded libs; newlib? stlport? etc... I was persuaded by the opinion that these should be separated.

src/dmd/mars.d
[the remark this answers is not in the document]
jacob-carlborg: Clang does not provide a C runtime, the platform does.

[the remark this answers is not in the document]
TurkeyMan: I think he might have rather meant:

> Why is there no CRuntime_AnyRuntimeAtAll ?

At least, that's what I was thinking ;)
There just isn't one defined for most of those platforms for some reason.

PR #8653, opened 2018-09-02
convert cod1.c to D

WalterBright:
It turns out it's a lot easier to diff the C and D versions if you copy the C file to a temp, then do a global search/replace of -> to . and then diff the D version with temp.


WalterBright:
It'd be nice of semaphoreci would put what failed at the bottom of the log. I can't find the reason why it's failing. Searching for the text 'error' is unhelpful. It'd also be nice if the scroll bar tab was wider than 2 pixels.

RazvanN7 (approved):

PR #8654, opened 2018-09-02
Fixes Issue 19195 - Support pragma to specify linker directives

TurkeyMan:
Added pragma(linkerdirective, "string")

Necessary for `std::vector`, `std::string`, etc.

Spec: https://github.com/dlang/dlang.org/pull/2463

### DIFF
--- a/src/dmd/backend/machobj.c
+++ b/src/dmd/backend/machobj.c
@@ -1603,6 +1603,15 @@ bool Obj::includelib(const char *name)
     return false;
 }
 
+/*******************************
+* Output linker directive.
+*/
+
+bool Obj::linkerdirective(const char *name)
+{
+    return false;
+}
+
 /**********************************
  * Do we allow zero sized objects?
  */
--- a/src/dmd/dsymbolsem.d
+++ b/src/dmd/dsymbolsem.d
@@ -1616,6 +1616,24 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
     {
         // Should be merged with PragmaStatement
         //printf("\tPragmaDeclaration::semantic '%s'\n", pd.toChars());
+        if (global.params.mscoff)
+        {
+            if (pd.ident == Id.linkerDirective)
+            {
+                if (!pd.args || pd.args.dim != 1)
+                    pd.error("one string argument expected for pragma(linkerDirective)");
+                else
+                {
+                    auto se = semanticString(sc, (*pd.args)[0], "linker directive");
+                    if (!se)
+                        goto Lnodecl;
+                    (*pd.args)[0] = se;
+                    if (global.params.verbose)
+                        message("linkopt   %.*s", cast(int)se.len, se.string);
+                }
+                goto Lnodecl;
+            }
+        }
         if (pd.ident == Id.msg)
         {
             if (pd.args)
--- a/src/dmd/id.d
+++ b/src/dmd/id.d
@@ -304,6 +304,7 @@ immutable Msgtable[] msgtable =
     // For pragma's
     { "Pinline", "inline" },
     { "lib" },
+    { "linkerDirective" },
     { "mangle" },
     { "msg" },
     { "startaddress" },
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -2300,6 +2300,13 @@ else
                 }
             }
         }
+        else if (ps.ident == Id.linkerDirective)
+        {
+            /* Should this be allowed?
+             */
+            ps.error("`pragma(linkerDirective)` not allowed as statement");
+            return setError();
+        }
         else if (ps.ident == Id.startaddress)
         {
             if (!ps.args || ps.args.dim != 1)
--- a/src/dmd/backend/cgobj.c
+++ b/src/dmd/backend/cgobj.c
@@ -1225,6 +1225,19 @@ bool Obj::includelib(const char *name)
     return true;
 }
 
+/*******************************
+* Output linker directive.
+* Output:
+*      directive is modified
+* Returns:
+*      true if operation is supported
+*/
+
+bool Obj::linkerdirective(const char *name)
+{
+    return false;
+}
+
 /**********************************
  * Do we allow zero sized objects?
  */
--- a/src/dmd/backend/elfobj.c
+++ b/src/dmd/backend/elfobj.c
@@ -1566,6 +1566,15 @@ bool Obj::includelib(const char *name)
     return false;
 }
 
+/*******************************
+* Output linker directive.
+*/
+
+bool Obj::linkerdirective(const char *name)
+{
+    return false;
+}
+
 /**********************************
  * Do we allow zero sized objects?
  */
--- a/src/dmd/backend/mscoffobj.c
+++ b/src/dmd/backend/mscoffobj.c
@@ -1060,6 +1060,19 @@ bool MsCoffObj::includelib(const char *name)
     return true;
 }
 
+/*******************************
+* Output linker directive name.
+*/
+
+bool MsCoffObj::linkerdirective(const char *directive)
+{
+    int seg = seg_drectve();
+    //dbg_printf("MsCoffObj::linkerdirective(directive *%s)\n",directive);
+    SegData[seg]->SDbuf->writeByte(' ');
+    SegData[seg]->SDbuf->write(directive, strlen(directive));
+    return true;
+}
+
 /**********************************
  * Do we allow zero sized objects?
  */
--- a/src/dmd/backend/obj.d
+++ b/src/dmd/backend/obj.d
@@ -61,6 +61,7 @@ version (OMF)
         void dosseg();
         void startaddress(Symbol *);
         bool includelib(const(char)* );
+        bool linkerdirective(const char *);
         bool allowZeroSize();
         void exestr(const(char)* p);
         void user(const(char)* p);
@@ -139,6 +140,7 @@ else version (OMFandMSCOFF)
         void dosseg();
         void startaddress(Symbol *);
         bool includelib(const(char)* );
+        bool linkerdirective(const(char)* );
         bool allowZeroSize();
         void exestr(const(char)* p);
         void user(const(char)* p);
@@ -215,6 +217,7 @@ else version (OMFandMSCOFF)
         //override size_t mangle(Symbol *s,char *dest);
         override void startaddress(Symbol *);
         override bool includelib(const(char)* );
+        override bool linkerdirective(const(char)* );
         override void _alias(const(char)* n1,const(char)* n2);
         override void user(const(char)* p);
     //    void _import(elem *e);
@@ -306,6 +309,7 @@ else version (Posix)
         static void dosseg();
         static void startaddress(Symbol *);
         static bool includelib(const(char)* );
+        static bool linkerdirective(const char *);
         static size_t mangle(Symbol *s,char *dest);
         static void _alias(const(char)* n1,const(char)* n2);
         static void user(const(char)* p);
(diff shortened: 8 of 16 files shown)


TurkeyMan:
Can anyone explain why the tests are failing? It doesn't emit any errors, and I can run this fine locally...

thewilsonator:
Could you add a that checks the linker directive does what its supposed to do?
e.g.

```
pragma(linkerdirective, "thisdoesntexist");
```
```
dmd testlinkerdirective.d &2>testlinkerdirective.log  || /*check for the linker error in testlinkerdirective.log*/
```

Otherwise looks good.

Geod24:
I'm wondering why this has to be a `pragma` as opposed to, say, a compiler flag ?
Can't the same effect be achieved with `-L` ?

TurkeyMan:
This isn't something the user should interact with. It's wrapped up in libs.

thewilsonator:
c.f. `-lfoo`/`pragma(lib,"foo")`

TurkeyMan:
But it's literally the entire point to NOT use `-L`. One should be able to embed directives in object files, since that's a feature of object files.

jacob-carlborg:
Is this something that GDC can support? @ibuclaw ?

TurkeyMan:
@jacob-carlborg What do you mean?
I don't know of any existing need for this outside windows.

src/dmd/dsymbolsem.d
```diff
@@ -1684,6 +1684,28 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                if (!se)
+                    goto Lnodecl;
+                (*pd.args)[0] = se;
+
+                if (global.params.verbose)
+                {
+                    auto directive = cast(char*)mem.xmalloc(se.len + 1);
+                    memcpy(directive, se.string, se.len);
+                    directive[se.len] = 0;
+                    message("linker directive   %s", directive);
```
ibuclaw: I guess this doesn't align with other verbose messages.

src/dmd/statementsem.d
```diff
@@ -2300,6 +2300,35 @@ else
                 }
             }
         }
+        else if (ps.ident == Id.linkerdirective)
+        {
+            version (all)
```
ibuclaw: Rather than using version all, remove the dead code.

src/dmd/backend/machobj.c
```diff
@@ -1603,6 +1603,16 @@ bool Obj::includelib(const char *name)
     return false;
 }
 
+/*******************************
+* Output linker directive.
+*/
+
+bool Obj::linkerdirective(const char *name)
+{
+    // do non-windows platforms support text linker directives?
```
ibuclaw: No, they don't as far as I'm aware.

ibuclaw (dismissed):
Remove version none code.

src/dmd/dsymbolsem.d
TurkeyMan: I'm not sure what you mean? It's a carbon copy of the lined from pragma(lib) just above...?

src/dmd/statementsem.d
TurkeyMan: This is a copy from the code above... I have no idea why it's like that, but I felt I should respect precedent.

ibuclaw:
> Is this something that GDC can support?

It is not something supportable no.  For the same reason as why `pragma(lib)` is not supported either.

From what I recall, you can't `import mod` and expect all `pragma(lib)` in that module to be pulled into the current compilation if you only encode this in the object file somewhere, the same would be true for this PR also.

The most platform agnostic thing to do is to write all information about link time dependencies to another file that is installed alongside the D header, importing then checks for both module source and linkdeps files - the latter being ignored if it doesn't exist.

For ELF, you can write all collected link deps to a custom text section, then when you get to the linking stage, the compiler can check all `*.o` files for this section and read in the linker flags to pass.

src/dmd/dsymbolsem.d
ibuclaw: Try compiling with `-v` to see what I mean. 😉

I don't see it as a problem, the whitespace padding seems quite arbitrary anyway.

src/dmd/statementsem.d
ibuclaw: Then perhaps rather than copy pasting code around, you should commonize it to a function.

TurkeyMan:
I tested this importing into a module being compiled, and it was definitely present in the object.
I don't know any reason for this to be platform agnostic... it's a windows/MSVC thing.

I'm not sure how the rest of your comment is relevant?
I must have missed something; why is everyone talking about `pragma(lib)`, I'm not sure how that came up? Is it because I used `/DEFAULTLIB` in the test? (the most common)
It is possible (and necessary) for .obj files to contain arbitrary linker directives. `/DEFAULTLIB` is just one of many.
`std::` populates .obj files with such bonus material to prove proper binary compatibility at link time.

ibuclaw:
> I don't know of any existing need for this outside windows.

And so rather than treating the pragma as no-op, you should be issuing an error if you attempt to use it on unsupported platforms, unless `-ignore` is passed.

TurkeyMan:
I didn't know it was unsupported...

How do I emit an error from these locations in code?
Can you give me a hint how to properly handle an error like this?

src/dmd/statementsem.d
TurkeyMan: It didn't feel like enough code to warrant that treatment... are you asking me to do that?

ibuclaw:
> I must have missed something; why is everyone talking about pragma(lib)

Because `pragma(lib)` does the same thing, emits a linker directive.

src/dmd/dsymbolsem.d
TurkeyMan: I just made it the same as the code nearby...

TurkeyMan:
Still not sure I see the connection? I guess `pragma(lib)` is one such directive, but that one's special, in that it's implemented to be portable.

I actually thought about refactoring such that `pragma(lib)` call this function... but I favoured minimising my edit footprint. See, my experience with DMD is that literally every line I touch is likely to waste a few extra hours making corrections when reviewers complain about minutia ;)

TurkeyMan:
I think I dun all changes...

TurkeyMan:
Oh... no now it's gonna fail all the unit tests on not-windows! >_<

ibuclaw:
> but that one's special, in that it's implemented to be portable.

Spoiler alert, it isn't - but that's another topic for another chat.  And I guess that because `pragma(lib)` isn't portable was the reason why I was pinged here also.

src/dmd/dsymbolsem.d
```diff
@@ -1616,6 +1616,30 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
     {
         // Should be merged with PragmaStatement
         //printf("\tPragmaDeclaration::semantic '%s'\n", pd.toChars());
+        version (Windows)
```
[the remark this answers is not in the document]
ibuclaw: What?  No one should be using version conditions in dmd unless it concerns *the host*.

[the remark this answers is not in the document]
ibuclaw: In this case, supporting the pragma is a feature of the target, and so the appropriate condition should be used.  Think Linux -> Windows cross compilers.

TurkeyMan:
It's really annoying that I can't see builtkite... whatever that us. It insists I register... and then when I do register, it still won't show me.

src/dmd/dsymbolsem.d
```diff
@@ -1616,6 +1616,30 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                    auto se = semanticString(sc, (*pd.args)[0], "linker directive");
+                    if (!se)
+                        goto Lnodecl;
+                    (*pd.args)[0] = se;
+                    if (global.params.verbose)
+                    {
+                        auto directive = cast(char*)mem.xmalloc(se.len + 1);
+                        memcpy(directive, se.string, se.len);
+                        directive[se.len] = 0;
+                        message("linkopt   %s", directive);
```
WalterBright: You can avoid the memory allocation with:
```
message("linkopt   %.*s", cast(int)se.len, se.string);
```

src/dmd/dsymbolsem.d
```diff
@@ -1616,6 +1616,30 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
     {
         // Should be merged with PragmaStatement
         //printf("\tPragmaDeclaration::semantic '%s'\n", pd.toChars());
+        if (global.params.mscoff)
+        {
+            if (pd.ident == Id.linkerdirective)
+            {
+                if (!pd.args || pd.args.dim != 1)
+                    pd.error("string expected for linker directive");
```
WalterBright: `one string argument expected for pragma(linkerDirective)`

src/dmd/id.d
```diff
@@ -304,6 +304,7 @@ immutable Msgtable[] msgtable =
     // For pragma's
     { "Pinline", "inline" },
     { "lib" },
+    { "linkerdirective" },
```
WalterBright: Should be `linkerDirective`

src/dmd/dsymbolsem.d
TurkeyMan: Like this: https://github.com/dlang/dmd/blob/master/src/dmd/dsymbolsem.d#L1658 ?

src/dmd/id.d
TurkeyMan: Really? Like `pragma(startaddress)`? :/

TurkeyMan: So... I feel worse about breaking precedent. Do you really want me to make this change?

TurkeyMan:
Just for the record. These changes are the definition of trivia. Each is a request to break from precedent. In every case, I rolled with precedent according to code nearby; I have no idea what anyone could be expected to do other than that.

I'm making amendments that I wouldn't usually support in principle; I think conforming to precedent important.

And while this sort of feedback is generally fine, it's disruptive, and I only find it acceptable in the context where turn-around time is measured in seconds... I could stretch for minutes <= 3.
DMD has a turnaround time of closer to an hour, and from that perspective, trivia like this feels repellent to contributors.

Again, I don't have a suggestion; I'm just trying to communicate how this is frustrating, and applies a strong pressure against my willingness to contribute at all.
DMD CI should be <5 minutes. It's a microscopic project by contrast to any software I've ever worked on. If that were the case, then fussing over trivia becomes reasonable.

src/dmd/id.d
WalterBright: The startaddress comes from the very beginning of D. Later on, but long ago, we moved to the other convention.

TurkeyMan:
Then you need to create a patch that updates the user experience atomically, maintaining consistency in one go.
Precedent updates should be atomic operations.

WalterBright:
I see no reason to start with a lesser error message and later replace it with a better one. Changing `startaddress` will break user code, and so is a long process, and probably pointless.

TurkeyMan:
Right, so we accept that precedent as established and not worth changing, and infer `linkerdirective` is the proper name?
Do you really want to establish a discontinuity where one does not already exist?

WalterBright:
__traits uses the newer suggested style: https://dlang.org/spec/traits.html
So, yes, please change linkerdirective to linkerDirective.

WalterBright (approved):

TurkeyMan:
I can't see buildkite, I have no idea what's hidden behind that link.
And I have no idea why that win32 test is failing. Nothing I touched has any interaction with the noise it's making.

rainers (approved):
Here is another use case: https://github.com/dlang/druntime/blob/master/src/rt/msvc.c#L48

TurkeyMan:
Oh yeah, that's MS's special (weird) brand of weak linkage ;)

TurkeyMan:
@ibuclaw I think you're blocking this?

PR #8655, opened 2018-09-02
Fixes Issue 19194 - version for `-mscrtlib` specification

TurkeyMan:
Necessary for `std::vector`, `std::string`, etc.

Spec https://github.com/dlang/dlang.org/pull/2462

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1202,6 +1202,34 @@ void addDefaultVersionIdentifiers(const ref Param params)
         {
             VersionCondition.addPredefinedGlobalIdent("CRuntime_Microsoft");
             VersionCondition.addPredefinedGlobalIdent("CppRuntime_Microsoft");
+            if (!global.params.mscrtlib || !global.params.mscrtlib[0])
+            {
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_None");
+            }
+            else if (FileName.equals(global.params.mscrtlib, "libcmt"))
+            {
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Static");
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Release");
+            }
+            else if (FileName.equals(global.params.mscrtlib, "libcmtd"))
+            {
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Static");
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Debug");
+            }
+            else if (FileName.equals(global.params.mscrtlib, "msvcrt") || FileName.equals(global.params.mscrtlib, "msvcrt100"))
+            {
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Dynamic");
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Release");
+            }
+            else if (FileName.equals(global.params.mscrtlib, "msvcrtd"))
+            {
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Dynamic");
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Debug");
+            }
+            else
+            {
+                assert(0, "unexpected mscrtlib");
+            }
         }
         else
         {
--- a/src/dmd/cond.d
+++ b/src/dmd/cond.d
@@ -658,6 +658,11 @@ extern (C++) final class VersionCondition : DVCondition
             case "CppRuntime_Gcc":
             case "CppRuntime_Microsoft":
             case "CppRuntime_Sun":
+            case "MSRuntime_None":
+            case "MSRuntime_Static":
+            case "MSRuntime_Dynamic":
+            case "MSRuntime_Debug":
+            case "MSRuntime_Release":
             case "unittest":
             case "assert":
             case "all":
--- a/test/fail_compilation/reserved_version.d
+++ b/test/fail_compilation/reserved_version.d
@@ -105,6 +105,11 @@ fail_compilation/reserved_version.d(206): Error: version identifier `CppRuntime_
 fail_compilation/reserved_version.d(207): Error: version identifier `CppRuntime_Gcc` is reserved and cannot be set
 fail_compilation/reserved_version.d(208): Error: version identifier `CppRuntime_Microsoft` is reserved and cannot be set
 fail_compilation/reserved_version.d(209): Error: version identifier `CppRuntime_Sun` is reserved and cannot be set
+fail_compilation/reserved_version.d(210): Error: version identifier `MSRuntime_None` is reserved and cannot be set
+fail_compilation/reserved_version.d(211): Error: version identifier `MSRuntime_Static` is reserved and cannot be set
+fail_compilation/reserved_version.d(212): Error: version identifier `MSRuntime_Dynamic` is reserved and cannot be set
+fail_compilation/reserved_version.d(213): Error: version identifier `MSRuntime_Debug` is reserved and cannot be set
+fail_compilation/reserved_version.d(214): Error: version identifier `MSRuntime_Release` is reserved and cannot be set
 ---
 */
 
@@ -216,6 +221,11 @@ version = CppRuntime_DigitalMars;
 version = CppRuntime_Gcc;
 version = CppRuntime_Microsoft;
 version = CppRuntime_Sun;
+version = MSRuntime_None;
+version = MSRuntime_Static;
+version = MSRuntime_Dynamic;
+version = MSRuntime_Debug;
+version = MSRuntime_Release;
 
 // This should work though
 debug = DigitalMars;
@@ -300,6 +310,11 @@ debug = CppRuntime_DigitalMars;
 debug = CppRuntime_Gcc;
 debug = CppRuntime_Microsoft;
 debug = CppRuntime_Sun;
+debug = MSRuntime_None;
+debug = MSRuntime_Static;
+debug = MSRuntime_Dynamic;
+debug = MSRuntime_Debug;
+debug = MSRuntime_Release;
 debug = D_Coverage;
 debug = D_Ddoc;
 debug = D_InlineAsm_X86;
--- a/test/fail_compilation/reserved_version_switch.d
+++ b/test/fail_compilation/reserved_version_switch.d
@@ -82,6 +82,11 @@
 // REQUIRED_ARGS: -version=CppRuntime_Gcc
 // REQUIRED_ARGS: -version=CppRuntime_Microsoft
 // REQUIRED_ARGS: -version=CppRuntime_Sun
+// REQUIRED_ARGS: -version=MSRuntime_None
+// REQUIRED_ARGS: -version=MSRuntime_Static
+// REQUIRED_ARGS: -version=MSRuntime_Dynamic
+// REQUIRED_ARGS: -version=MSRuntime_Debug
+// REQUIRED_ARGS: -version=MSRuntime_Release
 // REQUIRED_ARGS: -version=D_Coverage
 // REQUIRED_ARGS: -version=D_Ddoc
 // REQUIRED_ARGS: -version=D_InlineAsm_X86
@@ -178,6 +183,11 @@
 // REQUIRED_ARGS: -debug=CppRuntime_Gcc
 // REQUIRED_ARGS: -debug=CppRuntime_Microsoft
 // REQUIRED_ARGS: -debug=CppRuntime_Sun
+// REQUIRED_ARGS: -debug=MSRuntime_None
+// REQUIRED_ARGS: -debug=MSRuntime_Static
+// REQUIRED_ARGS: -debug=MSRuntime_Dynamic
+// REQUIRED_ARGS: -debug=MSRuntime_Debug
+// REQUIRED_ARGS: -debug=MSRuntime_Release
 // REQUIRED_ARGS: -debug=D_Coverage
 // REQUIRED_ARGS: -debug=D_Ddoc
 // REQUIRED_ARGS: -debug=D_InlineAsm_X86
@@ -279,6 +289,11 @@ Error: version identifier `CppRuntime_DigitalMars` is reserved and cannot be set
 Error: version identifier `CppRuntime_Gcc` is reserved and cannot be set
 Error: version identifier `CppRuntime_Microsoft` is reserved and cannot be set
 Error: version identifier `CppRuntime_Sun` is reserved and cannot be set
+Error: version identifier `MSRuntime_None` is reserved and cannot be set
+Error: version identifier `MSRuntime_Static` is reserved and cannot be set
+Error: version identifier `MSRuntime_Dynamic` is reserved and cannot be set
+Error: version identifier `MSRuntime_Debug` is reserved and cannot be set
+Error: version identifier `MSRuntime_Release` is reserved and cannot be set
 Error: version identifier `D_Coverage` is reserved and cannot be set
 Error: version identifier `D_Ddoc` is reserved and cannot be set
 Error: version identifier `D_InlineAsm_X86` is reserved and cannot be set


TurkeyMan:
What is that error on that one build machine?

TurkeyMan:
When I try and run that failing test, I get:
```
Error: undefined identifier double2
```
I'm fairly certain my patch didn't affect that...

src/dmd/mars.d
```diff
@@ -1361,7 +1361,37 @@ void addDefaultVersionIdentifiers()
...
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Release");
+            }
+            else if (!strcmp(global.params.mscrtlib, "msvcrtd"))
+            {
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Dynamic");
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Debug");
+            }
+            else
+            {
+                assert(0, "unexpected mscrtlib, how did the command line accept this?");
```
WalterBright: Please, no snark in the code.

src/dmd/mars.d
```diff
@@ -1361,7 +1361,37 @@ void addDefaultVersionIdentifiers()
     static if (TARGET.Windows)
     {
         if (global.params.mscoff)
+        {
             VersionCondition.addPredefinedGlobalIdent("CRuntime_Microsoft");
+            if (!global.params.mscrtlib || !global.params.mscrtlib[0])
+            {
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_None");
+            }
+            else if (!strcmp(global.params.mscrtlib, "libcmt"))
```
WalterBright: Windows has a case insensitive file system, so `strcmp` is unreliable. Use `FileName.compare`.

src/dmd/mars.d
TurkeyMan: That's not even remotely snark... that's a genuine question.
In the event of that assert, the question is, how did the commend line allow it? (ie, that is the problem source)

src/dmd/mars.d
TurkeyMan: `FileName.equals`?

WalterBright: that works too

src/dmd/mars.d
WalterBright: I'd prefer "the command line should have rejected it"

WalterBright:
When you write a PR for a bugzilla issue, please include a link in the issue to the PR. Otherwise, nobody perusing bugzilla will know that there's work being done on it.

TurkeyMan:
I feel like the same bot that populates that big block of text at the top should probably do that.

WalterBright:
Feel free to file an Enhancement Request on Bugzilla for that!

TurkeyMan:
Did.

WalterBright (approved):

TurkeyMan:
I can't see buildkite... but can anyone speculate what that Win32 build might be complaining about?
I can't imagine how anything I touched intersected the noise it's making...

src/dmd/mars.d
```diff
@@ -1361,7 +1361,37 @@ void addDefaultVersionIdentifiers()
...
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Release");
+            }
+            else if (FileName.equals(global.params.mscrtlib, "msvcrtd"))
+            {
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Dynamic");
+                VersionCondition.addPredefinedGlobalIdent("MSRuntime_Debug");
+            }
+            else
+            {
+                assert(0, "unexpected mscrtlib");
```
rainers: There is also at least `msvcrt100` being used for the mingw based libraries used when no VC is installed. As this is verifying user input it should be an error. (That will probably also make the CI output more readable).

rainers (changes requested):
Sorry for adding to the slow process, but my gut feeling is that this is not the right direction:
* -mscrtlib is currently just embedding a lib with `main`, it does not have to be specified with any other module
* if it has to be specified with every compilation you could also use -version on the command line
* how does `MsRuntime_None` help with selecting code generation path? You will still have to select Static/Dynamic and Debug/Release
* dmc suggests using stlport 4.5 and the commercial version of dmc comes with these precompiled libraries:
  - stlp45dm.lib
  - stlp45dm_debug.lib
  - stlp45dm_debug_static.lib
  - stlp45dm_debug_staticx.lib
  - stlp45dm_static.lib
  - stlp45dm_staticx.lib
  - stlp45dm_stldebug.lib
  - stlp45dm_stldebug_static.lib
  - stlp45dm_stldebug_staticx.lib
  - stlp45dms.lib
  - stlp45dms_debug.lib
  - stlp45dms_stldebug.lib
 
Do you want to add predefined versions for these, too?

TurkeyMan:
> `-mscrtlib` is currently just embedding a lib with main, it does not have to be specified with any other module

Embedding a lib reference is barely the start of it >_<

It has to be available to any module that wants to interact with the standard library.

The ABI changes with respect to MS CRT libs; data structures are different in debug/release, and different functions are available between static/dynamic. I can't do anything useful with stdcpp without this knowledge. Of all my recent PR's, this is the single most important one.

Instantiation of std:: containers also emits some custom linker directives to any obj that they come in contact with to guarantee the object files link correctly.

> if it has to be specified with every compilation you could also use -version on the command line

The user shouldn't be required to manually supply a (hopefully!) matching version whenever they want to use `extern(C++)`. They'll never know the correct one to choose, and it's totally error prone.

I don't think there's any harm in informing which one is being used to compile at compile time, and I don't know how else to generate correct code without this knowledge. It's not reasonable to depend on the user to supply a separate compile flag that repeats another compile flag they already supplied (ie, a matching version for the crt they selected).

> how does MsRuntime_None help with selecting code generation path? You will still have to select Static/Dynamic and Debug/Release

It's useful to give some sort of an error or request for information where the user explicitly requested that the object will not marry to a specific c++rt version. In that case, the user should not be able to instantiate C++ containers, because the ABI is unknown.

> dmc suggests using stlport 4.5 and the commercial version of dmc comes with these precompiled libraries [...]. Do you want to add predefined versions for these, too?

Does the ABI totally change with each selection? If not, then no, it's not useful.
I suspect they don't change the data structures for each version, that's just some madness that MS thought up in the 90's and they've been suffering ever since.

src/dmd/mars.d
TurkeyMan: Oh wow. `msvcrt100`... that's like a decade old.

I'll add a case for that lib. Is the full set available when using mingw, or just that one?
I'm surprised the mingw libs use the MS CRT... thereby expecting to support the MSVC STL ABI? It's super complicated and crap.

jacob-carlborg:
Whatever the decision is, it should have an explicit changelog entry since it adds a feature.

Geod24:
Removing the `auto-merge` label following Rainer's comment.
Also asked @wilzbach on Slack to invite you to Buildkite, so you will see the logs.
For reference, the failure is spurious and I've restarted the build.

src/dmd/mars.d
rainers: > Oh wow. msvcrt100... that's like a decade old.

It's the latest version the mingw .def-files support. Please note that "mingw-based" means that the mingw def files are used to generate import libraries for the most important system DLLs (similar to the libs in the platform SDK) and for a distributable C runtime.
I've chosen the version from VS2010 because later versions can be a hassle to install on Win7 or earlier. `msvcrt100` is a shared release library. It is not the usual mingw runtime.

rainers:
> The ABI changes with respect to MS CRT libs; data structures are different in debug/release

This is true for the C++ runtime, but not for the C runtime that is specified with this option. The same build of phobos links to any of the C runtimes (debug/relase, static/dynamic) for VS versions 2008 to 2017.

> Of all my recent PR's, this is the single most important one.

It's a convenience feature (specifying one option instead of 2). Deciding on how to select the library version should not stop you from working on the actual bindings.

On second thought, I guess it is ok to couple the C++ library with the C library, because that is what VC does aswell. I suspect that you will always need a C++ module to actually instantiate the templates anyway.

I think specifying other runtime libraries should not be forbidden, though. So maybe use `MSRuntime_Unknown` for these libraries and the empty option.

> The user shouldn't be required to manually supply a (hopefully!) matching version whenever they want to use extern(C++). They'll never know the correct one to choose, and it's totally error prone.

That's the same with `-msvcrt=` if using the command line, but it might help that the options look very similar in VS ;-)

>> dmc suggests using stlport 4.5 and the commercial version of dmc comes with these precompiled libraries [...]. Do you want to add predefined versions for these, too?
> 
> Does the ABI totally change with each selection?

There are different implementations for most containers for debug and release. stlport also distinguishes multi-threaded/single-threaded, but that doesn't seem to change for these libs.

I doubt it makes much sense to bind stlport as it is seriously outdated (this version seems from 2004). There are also two other even older stl versions in the commercial distributions of dmc.

src/dmd/mars.d
TurkeyMan: Oh I see. Fair enough.
You don't bundle the full set? That is, the static/dynamic release/debug versions?

TurkeyMan:
> This is true for the C++ runtime, but not for the C runtime that is specified with this option. The same build of phobos links to any of the C runtimes (debug/relase, static/dynamic) for VS versions 2008 to 2017.

There are no ABI problems don't affect the C runtime that I know of. The ABI issues are C++ related.
The C runtime has some difference in public symbols between the static/dynamic versions.
There have been times I've needed to call some functions in the C runtime which are not `export` in the dynamic version, but they are public in the static versions. That's the reason for distinction between static/dynamic, there is actually a slightly different set of functions available (both C and C++).

> It's a convenience feature (specifying one option instead of 2). Deciding on how to select the library version should not stop you from working on the actual bindings.

It's not just convenience, specifying 2 options requires that the user understand both options, and be certain that they correctly match. If the user specifies `-mscrt=mscrtd` and then specifies the incorrect version symbol, there will be runtime crashing due to ABI mismatch.
The `None` option is important, because it suggests that the user has elected to omit affinity to a particular CRT (totally common, it avoids the binary distribution matrix for static libs), but it's not possible to instantiate STL types in that case, and that needs to be correctly explained to them. A quality compile-time error explaining the situation depends on that knowledge.

I think the user should only supply `-mscrt=`, which they understand, and has a direct counterpart in the C++ code, and the library should do correct codegen accordingly, which is the same experience they expect from C++.

It's also important to embed the linker directives which refuse linkage between objects with mismatching CRT spec when C++ is used.
Since this depends on the user _using_ C++ (we don't want to create the typical MS linkage difficulties unnecessarily), the linker directive should be inserted by the library on instantiation of ABI sensitive types, rather than by the compiler. We need the knowledge in this PR to drive that logic.

> On second thought, I guess it is ok to couple the C++ library with the C library, because that is what VC does aswell. I suspect that you will always need a C++ module to actually instantiate the templates anyway.

I'm not sure what you mean? Coupling from a static/dynamic debug/release sense? I mean, they are coupled... that's just that. But what this set of versions means to C (function availability) and to C++ (ABI specification) is different, but in both cases, they contain useful information.

> I think specifying other runtime libraries should not be forbidden, though. So maybe use MSRuntime_Unknown for these libraries and the empty option.

Other runtime libraries are not forbidden.
I think `MSRuntime_None` is quite satisfactory to suggest that no MS runtime is in use. But also, it's reasonable to just not define any of these MSRuntime versions (current behaviour) in that case.
If a GDC toolchain were built for mingw or cygwin let's say, it would specify `CRuntime_GLibc` and `CppRuntime_Gnu` and none of the MS versions would be defined. C++ modules would generate for GNU structures naturally. C bindings would also do the right thing.

> That's the same with -msvcrt= if using the command line, but it might help that the options look very similar in VS ;-)

What I'm saying is, if they're required to supply *two* things independently, now it's the user responsibility to assure that they correctly match. Failure to do so == runtime crashes.

I think that's an unacceptable user experience. We can infer the correct versions from the runtime selection easily. No reason to expect the user to do it.

> There are different implementations for most containers for debug and release. stlport also distinguishes multi-threaded/single-threaded, but that doesn't seem to change for these libs.

Does 'different implementations' == 'different ABI'? Microsoft's STL is the only one I've ever seen that redefined the data structures and makes the whole thing binary incompatible with itself. Performing extra runtime checks and that sort of thing (what sane implementations do) doesn't break the ABI like MS does.

> I doubt it makes much sense to bind stlport as it is seriously outdated (this version seems from 2004). There are also two other even older stl versions in the commercial distributions of dmc.

I'll have to do *something* for DMD + OMF (assume the same runtime as DMC?) ... I don't think I can escape it. But again, I don't think stlport is in the business of producing incompatible ABI's, even if they do have multiple runtime builds.

TurkeyMan:
*sigh*

rainers:
> I'm not sure what you mean?

That I (now) agree that the C++ libraries should be selected by the user with the same switch as the C libraries. It shouldn't be a blocker for you to work on the bindings, though.

> Other runtime libraries are not forbidden.

They are forbidden by the assert.


> What I'm saying is, if they're required to supply two things independently, now it's the user responsibility to assure that they correctly match. Failure to do so == runtime crashes.

What I'm saying is that `-msvcrt` settings must still match across dependent libraries (think dub) and as usually noone will care, the default libcmt will always be used.

> Does 'different implementations' == 'different ABI'?

Yes, I just checked the vector implementation: the debug version has an additional iterator list member. But building these stldebug-libraries is not coupled to the C runtime, they are built with -D_STLP_DEBUG. That kind of corresponds to the `_ITERATOR_DEBUG_LEVEL` definition within the VS libs (are you planning to deal with this?).

src/dmd/mars.d
rainers: > You don't bundle the full set? That is, the static/dynamic release/debug versions?

Only the dynamic/release version is redistributable.

rainers:
I would like to bikeshed the defined versions: we usually have versions if something is enabled (D_Coverage, unnittest, etc.) or one of multiple options (e.g. CRuntime_...). With MSRuntime_Debug/MSRuntime_Release both states of a "switch" are added as versions, one of them only set if MSRuntime_None is not defined. These versions interdependent and not orthogonal.

Wouldn't it be better to just set one of MSRuntime_libcmt, MSRuntime_msvcrt, etc. and let the code sort out the rest?

TurkeyMan:
> They are forbidden by the assert.

Oh, right. But this is just Windows COFF DMD specifically. It can be defined differently for LDC/GDC/cross-compilers, etc. We don't currently have any command line arg to select alternative C or C++ libraries, but this design isn't prohibiting it.
These versions are only applicable when the MS runtime is selected. I don't think they need to be defined when it's not selected, since they're meaningless in those cases.

> What I'm saying is that -msvcrt settings must still match across dependent libraries (think dub) and as usually noone will care, the default libcmt will always be used.

If a dub based project does link against C++ code, then they'll certainly need to specify the proper CRT, just like any MSVC project must. That is simply the unavoidable nature of linking against C++.
What that means for dub-based dependencies is a genuine problem, and this exposes one key impotence of dub, and probably the single greatest reason why I don't use it, and I couldn't even if I wanted to.

I suspect that people that link against C++ will tend not to use DUB; they probably already have some build system to build the C++ code.

> kind of corresponds to the _ITERATOR_DEBUG_LEVEL definition within the VS libs (are you planning to deal with this?).

Absolutely. It's completely inescapable.
Sadly, in the MS case, _ITERATOR_DEBUG_LEVEL *is* strongly coupled to the runtime selection. Handling _ITERATOR_DEBUG_LEVEL is the primary motivator for requiring `MSRuntime_Debug/Release`, and also the `pragma(linkerDirective)` stuff. It's impossible to successfully bind the MS STL without both those tools.

If the STLPort debug stuff is driven by a user-supplied macro, then that's great; it can equally correspond to a user-supplied version and requires no language support. The MS case is not so simple, because it's hard-coded to the CRT selection.

> I would like to bikeshed the defined versions [...] and let the code sort out the rest?

I thought about that... but I think it has a higher maintenance burden. Any time a new lib appears, all that's required is to state whether it's static/dynamic or debug/release, and supporting library code will work properly. The alternative would require a patch to the libs as well each time a runtime appeared, which would add code to resolve them down to these same concepts anyway... that just seems like a greater maintenance burden. But I'm not particularly emotional on this either way.

Perhaps you're suggesting that the compiler would fabricate a version string by concatenating the runtime name, and then it rather _moves_ such maintenance burden to the lib?
That might be fine, but it's unprecedented, and hard to document... hard to reserve those names (ie `MSRuntime_*`)

rainers:
> Oh, right. But this is just Windows COFF DMD specifically. It can be defined differently for LDC/GDC/cross-compilers, etc. We don't currently have any command line arg to select alternative C or C++ libraries, but this design isn't prohibiting it.
> These versions are only applicable when the MS runtime is selected. I don't think they need to be defined when it's not selected, since they're meaningless in those cases.

I think we should not forbid custom libraries for MS-COFF, too. See the `msvcrt100` example.

> Sadly, in the MS case, _ITERATOR_DEBUG_LEVEL is strongly coupled to the runtime selection.

Microsoft distributes 3 versions of the debug libraries and 2 version of the release libraries with different _ITERATOR_DEBUG_LEVELs (libcpmt and libconcrt), see https://docs.microsoft.com/de-de/cpp/standard-library/iterator-debug-level. How are these going to be selected?

> thought about that... but I think it has a higher maintenance burden. Any time a new lib appears, all that's required is to state whether it's static/dynamic or debug/release,

A new library will probably have other properties that need to be distinguished by a new version anyway.



> Perhaps you're suggesting that the compiler would fabricate a version string by concatenating the runtime name, and then it rather moves such maintenance burden to the lib?
> That might be fine, but it's unprecedented, and hard to document... hard to reserve those names (ie MSRuntime_*)

I thought about that to but rejected it for the same reasons. It could also be a special symbol like `__VENDOR__` or `__VERSION__`, though.

TurkeyMan:
> I think we should not forbid custom libraries for MS-COFF, too. See the msvcrt100 example.

Sure... certainly. I'm not quite sure what you mean though? How are we forbidding them?

> [...] _ITERATOR_DEBUG_LEVELs [...]. How are these going to be selected?

1. Allow the user to provide `_ITERATOR_DEBUG_LEVEL_0`, `_ITERATOR_DEBUG_LEVEL_1`, `_ITERATOR_DEBUG_LEVEL_2` as versions when they build. The requested level will be used (and appripriate `/FAILMISMATCH` directive written to the object files.
2. If user did not make an explicit request, infer the appropriate level for the supplied `-mscrt=` (I expect this is the common usage)
3. If the user selected to omit a CRT affinity (this may be to inhibit automatic linkage of CRT), then we have no source of confidence for correct linkage against their extern code. In this niche case, there are 2 options:
- 1. Associate the typical levels 0 or 2 from whether `-release` is supplied, embed `/FAILMISMATCH`, and user will experience a link error if they attempt to link incorrectly. (This is actually very likely to be correct with minimal user intervention, but rare occurrence of link error might be surprising)
- 2. Eagerly emit an error message explaining that the user must manually select an appropriate level (ie, point 1)

I think I lean towards the first option on the final case, but the second option is also reasonable. I bias towards convenience and reduced cognitive load in general; especially in the knowledge that most VS users have *no idea* that this problem even exists, and I wouldn't want to cause users to be required to interact with the problem unless absolutely necessary.

Finally, because pragma(linkerDirective), I'm able to emit the `/FAILMISMATCH` only when the types are instantated by user code, which means D programs that make no use of STL will never be polluted by MS linkage problems.

> A new library will probably have other properties that need to be distinguished by a new version anyway.

Will they? If it's a *different* library, then it might need its own `CppRuntime_xxxx` version so appropriate structures can be generated... but hopefully they don't act so differently.
If it's a different *Microsoft* runtime, then they've had a consistent formula for a very long time. I also think they've learned their lesson here, and they've made strong effort to NOT change the ABI of their libs for the last 3 VS versions now. I expect that'll be ongoing to the best of their ability.

One example of your proposition might be the old single/multithreaded distinction?
I thought about that, but I couldn't imagine any reason to distinguish the single/multi-threaded libs that MS used to roll with. From the calling perspective, they're API and ABI compatible. We can always add a means for distinction if/when that case arises, but I couldn't imagine it.

Point is, I don't think the design in this PR inhibits anything, it just adds necessary information beside the MS CRT's, and no such versions should be defined when not using MS CRT's which don't carry this baggage.

> > Perhaps you're suggesting that the compiler would fabricate a version string by concatenating the runtime name, and then it rather moves such maintenance burden to the lib?
>
> I thought about that, too, but rejected it for the same reasons. It could also be a special symbol like `__VENDOR__` or `__VERSION__`, though.

Yeah, that's a better strategy than generated versions. I'd be happy with either approach (this PR, or a magic string enum).

TurkeyMan:
On further thought, I think I lean towards `__CXXRT_VERSION__` (or `__MSCRT_VERSION__`, if we restrict it to the MS toolchains) or something like that... but it feels more invasive to me, and it also feels like it should be part of a suite of such information-providing symbols. I'm not sure we wanna go that far.
So, maybe if go that way, then `__MSCRT_VER__` and restrict the symbol to toolchains that use MS runtimes to protect against snowball effect?

I'm still not upset by the versions used in this PR though...

MartinNowak:
Without too much inside into msvcrt, this problem seems similar to ABI differences in FreeBSD's libc.
Please check the discussion around version identifiers and platform specific lexer macros here https://github.com/dlang/dmd/pull/8567#issuecomment-421278755.

Also note that such conditional compilation guards must not produce differences in linkable entities, i.e. they must be D header-only changes, as we can only compile and distribute a single release build of phobos/druntime atm.

TurkeyMan:
Here's an alternative: https://github.com/dlang/dmd/pull/8701

TurkeyMan:
> they must be D header-only changes

Check, this is understood, and is indeed the case.

TurkeyMan:
Killing this, since the other approach was accepted.

PR #8658, opened 2018-09-04
[C++]  Include globals.h instead of mars.h 

ibuclaw:
The mars header doesn't need to be included everywhere any more.  Interfacing C++ programs should explicitly include the header if they want access to anything that is declared there.

### DIFF
--- a/src/dmd/aliasthis.h
+++ b/src/dmd/aliasthis.h
@@ -10,7 +10,7 @@
 
 #pragma once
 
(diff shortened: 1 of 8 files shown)


jacob-carlborg (approved):

Geod24 (approved):

dnadlinger (approved):

Geod24:
AppVeyor is being fussy again

PR #8660, opened 2018-09-04
[stable] Fix Issue 19219 - Could not CTFE with std.math.exp

kinke:
The underlying issue is an apparent CTFE bug wrt. passing static immutable static arrays by (const) ref, see https://issues.dlang.org/show_bug.cgi?id=17351#c10. That makes this [poly() optimization](https://github.com/dlang/phobos/commit/e1d987dda6e61596fbdf67ede4f0e76a12badbec) not CTFE-able.

### DIFF
--- a/src/dmd/builtin.d
+++ b/src/dmd/builtin.d
@@ -124,6 +124,13 @@ extern (C++) Expression eval_log10(Loc loc, FuncDeclaration fd, Expressions* arg
     return new RealExp(loc, CTFloat.log10(arg0.toReal()), arg0.type);
 }
 
+extern (C++) Expression eval_exp(Loc loc, FuncDeclaration fd, Expressions* arguments)
+{
+    Expression arg0 = (*arguments)[0];
+    assert(arg0.op == TOK.float64);
+    return new RealExp(loc, CTFloat.exp(arg0.toReal()), arg0.type);
+}
+
 extern (C++) Expression eval_expm1(Loc loc, FuncDeclaration fd, Expressions* arguments)
 {
     Expression arg0 = (*arguments)[0];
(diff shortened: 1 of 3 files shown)


src/dmd/builtin.d
```diff
@@ -124,6 +124,13 @@ extern (C++) Expression eval_log10(Loc loc, FuncDeclaration fd, Expressions* arg
     return new RealExp(loc, CTFloat.log10(arg0.toReal()), arg0.type);
 }
 
+extern (C++) Expression eval_exp(Loc loc, FuncDeclaration fd, Expressions* arguments)
```
jacob-carlborg: Not my expertise but Walter has changed a lot of passing `Loc` by value to pass by `const ref`

jacob-carlborg: Ah, ok.

PetarKirov (approved):

jacob-carlborg (approved):

thewilsonator:
Win32: `rm: cannot remove ‘test_results/runnable/test2.exe’: Device or resource busy` looks like it needs another resuscitation.

src/dmd/builtin.d
```diff
@@ -381,6 +388,7 @@ public extern (C++) void builtin_init()
     add_builtin("_D3std4math3tanFNaNbNiNeeZe", &eval_tan);
     add_builtin("_D3std4math4sqrtFNaNbNiNeeZe", &eval_sqrt);
     add_builtin("_D3std4math4fabsFNaNbNiNeeZe", &eval_fabs);
+    add_builtin("_D3std4math3expFNaNbNiNeeZe", &eval_exp);
```
Geod24: Could that builtin have gone to druntime instead ? AFAIK that's what we are trying to do nowadays (well, since a few years), and this list should probably be purged since many of those builtin just forward to druntime AFAIK.

kinke: I think you mixed something up here. These CTFE builtins are generally used for non-CTFEable functions, e.g., if implemented in asm. None of them use druntime, but the host's C functions.

PR #8661, opened 2018-09-05
Temporarily revert SemaphoreCI to v2.081.2

Geod24:
Because, as explained in the comment, v2.082.0's DUB is broken and is blocking development,
as SemaphoreCI is required to merge a PR.

### DIFF
--- a/semaphoreci.sh
+++ b/semaphoreci.sh
@@ -36,7 +36,14 @@ source ci.sh
 # Always source a DMD instance
 ################################################################################
 
-install_d "$DMD"
+# FIXME: v2.082.0 has a broken DUB which fails the CI
+# Remove this when a fixed v2.082.1 is released
+# See https://github.com/dlang/dub/issues/1551
+if [ "$DMD" == "dmd" ]; then
+    install_d "dmd-2.081.2"
+else
+    install_d "$DMD"
+fi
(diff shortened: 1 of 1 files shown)


semaphoreci.sh
```diff
@@ -36,7 +36,14 @@ source ci.sh
 # Always source a DMD instance
 ################################################################################
 
-install_d "$DMD"
+# FIXME: v2.082.0 has a broken DUB which fails the CI
+# Remove this when a fixed v2.082.1 is released
+# See https://github.com/dlang/dub/issues/1551
+if [ "$DMD" == "dmd" ]; then
+    install_d "dmd-2.081.2"
```
jacob-carlborg: We don’t want to install Dub separately and still use DMD 2.082.0?

Geod24: There's already all the logic in place to get a specific version in `install.sh`, but nothing to get `dub`. It would be a much bigger investment (and waste, IMHO) of time to do the later, and not of much help.

I already submitted a PR to dub to fix the bug, all I'm waiting for is for someone with the right credentials to pull it and make a release. This just unblocks DMD development.

jacob-carlborg:
Approved since it’s blocking PRs.

PR #8663, opened 2018-09-05
add setTarget()

WalterBright:
The setting of `params` should not be intermingled with setting version identifiers. Split the former into a separate function.

### DIFF
--- a/src/dmd/frontend.d
+++ b/src/dmd/frontend.d
@@ -32,12 +32,13 @@ void initDMD()
     import dmd.expression : Expression;
     import dmd.globals : global;
     import dmd.id : Id;
-    import dmd.mars : addDefaultVersionIdentifiers;
+    import dmd.mars : setTarget, addDefaultVersionIdentifiers;
     import dmd.mtype : Type;
     import dmd.objc : Objc;
     import dmd.target : Target;
 
     global._init();
+    setTarget(global.params);
     addDefaultVersionIdentifiers(global.params);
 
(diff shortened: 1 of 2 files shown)


src/dmd/frontend.d
```diff
@@ -32,12 +32,13 @@ void initDMD()
     import dmd.expression : Expression;
     import dmd.globals : global;
     import dmd.id : Id;
-    import dmd.mars : addDefaultVersionIdentifiers;
+    import dmd.mars : setTarget, addDefaultVersionIdentifiers;
```
RazvanN7: This looks like a bug. Isn't set target a private function?

src/dmd/mars.d
```diff
@@ -1261,70 +1261,94 @@ private void setDefaultLibrary()
...
- * target platform in `global`.
+ * target platform in `params`.
  * https://dlang.org/spec/version.html#predefined-versions
  *
  * Needs to be run after all arguments parsing (command line, DFLAGS environment
  * variable and config file) in order to add final flags (such as `X86_64` or
  * the `CRuntime` used).
+ *
+ * Params:
+ *      params = which target to compile for
```
Geod24: Please add a note that `setTarget` needs to be called before this function

src/dmd/frontend.d
WalterBright: not anymore!

RazvanN7 (approved):

PR #8664, opened 2018-09-05
refactor setTargetCPU()

WalterBright:
There is a simpler function lurking within it.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -357,7 +357,8 @@ private int tryMain(size_t argc, const(char)** argv)
     if (global.params.color)
         global.console = Console.create(core.stdc.stdio.stderr);
 
-    global.params.cpu = setTargetCPU(global.params.cpu);
+    setTarget(global.params);           // set target operating system
+    setTargetCPU(global.params);
     if (global.params.is64bit != is64bit)
         error(Loc.initial, "the architecture must not be changed in the %s section of %s", envsection.ptr, global.inifilename);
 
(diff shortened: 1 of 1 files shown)


WalterBright:
This relies on https://github.com/dlang/dmd/pull/8663

jacob-carlborg:
#8663 has been merged, please rebase.

WalterBright:
rebased

Geod24 (changes requested):
Personally I don't find this to be much of an enhancement. Instead of passing `CPU` to `setTargetCPU` we are passing the whole `Param` struct, which contains just about everything passed on the command line.

So as an external person reads the code, it is not clear what this function does at call site. It **could** do about anything: change any value in `Param`, or read any value in `Param`.

On the other hand, when looking at the call site, the current code clearly shows that only `params.cpu` is read, and `params.cpu` is written, and the `setTargetCPU` does not have to know about `Params`.

WalterBright:
> does not have to know about Params

This is incorrect, .cpu, .is64bit and .isOSX fields are read:
```
if (params.is64bit || params.isOSX)
```
Another problem with the original is its use of TARGET.OSX instead of params.isOSX, which is a bit schizophrenic because of the params.is64bit.

A close read of the original shows it takes four parameters:

1. cpu
2. params.is64bit
3. TARGET.OSX
4. cpuid

There is no indication of the latter three in the parameter list for the function. A case could be made for passing cpuid in as an argument, too.

Ideally, params should be passed in as `const ref`, but there's a problem with the design that there should be a globals.cpu to set instead of params.cpu, but that's a refactoring for another day.

It took me several reads of the original tangled logic to figure it all out. I like the new layout much better.

src/dmd/mars.d
```diff
@@ -1474,51 +1475,38 @@ extern(C) void printGlobalConfigs(FILE* stream)
...
- *      value to generate code for
+ *      params = parameters set by command line switch
  */
 
-private CPU setTargetCPU(CPU cpu)
+private void setTargetCPU(ref Param params)
 {
-    // Determine base line for target
-    CPU baseline = CPU.x87;
-    if (global.params.is64bit)
```
Geod24: Right, I missed this

Geod24 (approved):

Geod24:
Thanks for the breakdown, it does make sense

jacob-carlborg:
Isn't better to have the function return the value instead of setting it inside the function?

PR #8665, opened 2018-09-05
Fix Issue 19209 - [ICE] Overriding a field in a baseclass issues an ICE

RazvanN7:


src/dmd/dsymbolsem.d
```diff
@@ -3458,19 +3458,30 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
-                    OutBuffer buf1, buf2;
+                    OutBuffer buf;
 
-                    functionToBufferFull(cast(TypeFunction)(fd.type), &buf1,
-                        new Identifier(fd.toPrettyChars()), &hgs, null);
-                    functionToBufferFull(cast(TypeFunction)(funcdecl.type), &buf2,
+                    auto fd = s.isFuncDeclaration();
+                    functionToBufferFull(cast(TypeFunction)(funcdecl.type), &buf,
                         new Identifier(funcdecl.toPrettyChars()), &hgs, null);
+                    char* funcdeclToString = buf.extractString();
```
WalterBright: extractString() causes a memory leak. Use peekString(). Also declare funcdeclToString as const, and rename it to funcdeclToChars because it is a char*, not a D string.

WalterBright (changes requested):

src/dmd/dsymbolsem.d
```diff
@@ -3458,19 +3458,30 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+                    {
+                        OutBuffer buf1;
 
-                    error(funcdecl.loc, "function `%s` does not override any function, did you mean to override `%s`?",
-                        buf2.extractString(), buf1.extractString());
+                        functionToBufferFull(cast(TypeFunction)(fd.type), &buf1,
+                            new Identifier(fd.toPrettyChars()), &hgs, null);
+
+                        error(funcdecl.loc, "function `%s` does not override any function, did you mean to override `%s`?",
+                            funcdeclToString, buf1.extractString());
```
WalterBright: again, use peekString()

src/dmd/dsymbolsem.d
RazvanN7: extractString was used before .... Chesterton's Fence.

>and rename it to funcdeclToChars

Indeed but I wanted to be consistent with `extractString`/`peekString`

RazvanN7:
@WalterBright Done.

src/dmd/dsymbolsem.d
WalterBright: Yes, but those are named wrongly. The convention should be `String` for char[], and `Chars` for zero terminated char*.

WalterBright (approved):

PR #8666, opened 2018-09-05
Fix issue 19225 - Confusing error message on `static else`

0xEAB:
So users assuming `static else` is fine get a helpful error message.

This might need some further fine-tuning, though.
Since it will trigger on faulty code like this:
```d
module mymod;
static else i = 0;
static int j = 0;
```

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -3686,6 +3686,8 @@ final class Parser(AST) : Lexer
 
         default:
             error("basic type expected, not `%s`", token.toChars());
+            if (token.value == TOK.else_)
+                errorSupplemental(token.loc, "There's no `static else`, use `else` instead.");
             t = AST.Type.terror;
             break;
         }
@@ -4558,7 +4560,7 @@ final class Parser(AST) : Lexer
             bool isThis = (t.ty == AST.Tident && (cast(AST.TypeIdentifier)t).ident == Id.This && token.value == TOK.assign);
             if (ident)
                 checkCstyleTypeSyntax(loc, t, alt, ident);
-            else if (!isThis)
+            else if (!isThis && (t != AST.Type.terror))
                 error("no identifier for declarator `%s`", t.toChars());
 
             if (tok == TOK.alias_)
--- a/test/fail_compilation/diag19225.d
+++ b/test/fail_compilation/diag19225.d
@@ -1,0 +1,15 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/diag19225.d(14): Error: basic type expected, not `else`
+fail_compilation/diag19225.d(14):        There's no `static else`, use `else` instead.
+fail_compilation/diag19225.d(14): Error: found `else` without a corresponding `if`, `version` or `debug` statement
+fail_compilation/diag19225.d(15): Error: unrecognized declaration
+---
+*/
+
+void main()
+{
+    static if (true) {}
+    static else {}
+}


src/dmd/parse.d
```diff
@@ -3684,6 +3684,11 @@ final class Parser(AST) : Lexer
             check(TOK.rightParentheses);
             break;
 
+        case TOK.else_:
+            error("basic type expected, not `else`. Hint: there's no `static else`, use `else` instead.");
```
jacob-carlborg: Perhaps the hint should be a supplemental message.

0xEAB: Should I strip the suffix `Hint:` then?

jacob-carlborg: Not sure.

jacob-carlborg (changes requested):
Missing tests. Have a look in `tests/fail_compilation` for how to test error messages.

src/dmd/parse.d
```diff
@@ -3684,6 +3684,13 @@ final class Parser(AST) : Lexer
             check(TOK.rightParentheses);
             break;
 
+        case TOK.else_:
```
WalterBright: If this is added instead as:
```
if (token.value == TOK.else_)
    errorSupplemental(token.loc, ...)
```
then this change adds only two lines of code instead of 6.

WalterBright: Added under the `default:` section, that is.

0xEAB: oh thanks, makes sense now that the hint is a separate message

0xEAB:
Is the filename of the test (`bug19225.d`) fine?

thewilsonator:
Yes.

WalterBright (approved):

test/fail_compilation/bug19225.d
```diff
@@ -0,0 +1,16 @@
+/*
+TEST_OUTPUT:
+---
+fail_compilation/bug19225.d(15): Error: basic type expected, not `else`
+fail_compilation/bug19225.d(15):        There's no `static else`, use `else` instead.
+fail_compilation/bug19225.d(15): Error: no identifier for declarator `_error_`
```
jacob-carlborg: Are error messages containing `_error_` supposed to be printed?

jacob-carlborg:
Please squash commits.

test/fail_compilation/bug19225.d
0xEAB: I don't know if they're supposed to. But those lines also show up without my patch.
see:
```
$ rdmd --compiler=ldc2 --eval="static if(false){}static else if {}"
(18): Error: basic type expected, not else
(18): Error: no identifier for declarator _error_
(18): Error: found else instead of statement
(19): Error: unrecognized declaration
```

0xEAB:
- Renamed the test as @thewilsonator suggested.
- Squashed everything into one commit

test/fail_compilation/bug19225.d
0xEAB: (To avoid confusion, `ldc` targets to LDC v1.9.0 on my machine.)

thewilsonator: I general it is good to only a one error and hence error message for one error in the code so as to avoid cascaded errors like this.

this is usually fixed by adding a 
```
if (foo.tok == TOK.error) return; /*or otherwise give up further parsing/semantic analysis*/
else checnkTheCondition();
```

0xEAB: (but also with DMD, the `_error_` thingy is still there)

0xEAB: @thewilsonator How would I implement this in this case?
Everything that happens next to the error message is returning `AST.Type.terror`.

thewilsonator: add a guard like [this]( https://github.com/dlang/dmd/blob/4a2805d4986ee85899abeb7903f523526603291c/src/dmd/parse.d#L3116) to [here](https://github.com/dlang/dmd/blob/4a2805d4986ee85899abeb7903f523526603291c/src/dmd/parse.d#L4562)

thewilsonator:
Also I think it would be better to have `There's no `static else`, use `else` instead.` as the error rather than as a supplementary to `basic type expected, no foo`.

0xEAB:
Is is safe to assume `static else` will always occur as this kind of mistake?

jacob-carlborg (approved):

RazvanN7 (approved):

PR #8667, opened 2018-09-05
Implement extern(C++,"string") C++ namespace mangling feature

ghost:
Still needs some checking to ensure the namespace name is valid, right now you can do `"+.$.&"` for a namespace.

Pending DIP, possibly maybe.

### DIFF
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -1159,11 +1159,14 @@ struct ASTBase
 
     extern (C++) final class Nspace : ScopeDsymbol
     {
-        extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members)
+        bool ignoreCppSymbols;
+
+        extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members, bool ignoreCppSymbols)
         {
             super(ident);
             this.loc = loc;
             this.members = members;
+            this.ignoreCppSymbols = ignoreCppSymbols;
         }
 
         override void accept(Visitor v)
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -4540,6 +4540,44 @@ struct ASTBase
             this.sz = 1;                    // work around LDC bug #1286
         }
 
+        /**********************************************
+        * Write the contents of the string to dest.
+        * Use numberOfCodeUnits() to determine size of result.
+        * Params:
+        *  dest = destination
+        *  tyto = encoding type of the result
+        *  zero = add terminating 0
+        */
+        void writeTo(void* dest, bool zero, int tyto = 0) const
+        {
+            int encSize;
+            switch (tyto)
+            {
+                case 0:      encSize = sz; break;
+                case Tchar:  encSize = 1; break;
+                case Twchar: encSize = 2; break;
+                case Tdchar: encSize = 4; break;
+                default:
+                    assert(0);
+            }
+            if (sz == encSize)
+            {
+                memcpy(dest, string, len * sz);
+                if (zero)
+                    memset(dest + len * sz, 0, sz);
+            }
+            else
+                assert(0);
+        }
+
+        extern (D) final const(char)[] toStringz() const
+        {
+            auto nbytes = len * sz;
+            char* s = cast(char*)mem.xmalloc(nbytes + sz);
+            writeTo(s, true);
+            return s[0 .. nbytes];
+        }
+
         override void accept(Visitor v)
         {
             v.visit(this);
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -1162,14 +1162,14 @@ struct ASTBase
         /**
          * Determines whether the symbol for this namespace should be included in the symbol table.
          */
-        bool ignoreCppSymbols;
+        bool mangleOnly;
 
-        extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members, bool ignoreCppSymbols)
+        extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members, bool mangleOnly)
         {
             super(ident);
             this.loc = loc;
             this.members = members;
-            this.ignoreCppSymbols = ignoreCppSymbols;
+            this.mangleOnly = mangleOnly;
         }
 
         override void accept(Visitor v)
--- a/src/dmd/astbase.d
+++ b/src/dmd/astbase.d
@@ -1159,6 +1159,9 @@ struct ASTBase
 
     extern (C++) final class Nspace : ScopeDsymbol
     {
+        /**
+         * Determines whether the symbol for this namespace should be included in the symbol table.
+         */
         bool ignoreCppSymbols;
 
         extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members, bool ignoreCppSymbols)
--- a/src/dmd/nspace.d
+++ b/src/dmd/nspace.d
@@ -30,6 +30,9 @@ private enum LOG = false;
  */
 extern (C++) final class Nspace : ScopeDsymbol
 {
+    /**
+     * Determines whether the symbol for this namespace should be included in the symbol table.
+     */
     bool ignoreCppSymbols;
 
     extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members, bool ignoreCppSymbols)
--- a/src/dmd/nspace.d
+++ b/src/dmd/nspace.d
@@ -30,23 +30,27 @@ private enum LOG = false;
  */
 extern (C++) final class Nspace : ScopeDsymbol
 {
-    extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members)
+    bool ignoreCppSymbols;
+
+    extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members, bool ignoreCppSymbols)
     {
         super(ident);
         //printf("Nspace::Nspace(ident = %s)\n", ident.toChars());
         this.loc = loc;
         this.members = members;
+        this.ignoreCppSymbols = ignoreCppSymbols;
     }
 
     override Dsymbol syntaxCopy(Dsymbol s)
     {
-        auto ns = new Nspace(loc, ident, null);
+        auto ns = new Nspace(loc, ident, null, ignoreCppSymbols);
         return ScopeDsymbol.syntaxCopy(ns);
     }
 
     override void addMember(Scope* sc, ScopeDsymbol sds)
     {
-        ScopeDsymbol.addMember(sc, sds);
+        if(!ignoreCppSymbols)
+            ScopeDsymbol.addMember(sc, sds);
         if (members)
         {
             if (!symtab)
--- a/src/dmd/nspace.d
+++ b/src/dmd/nspace.d
@@ -33,26 +33,26 @@ extern (C++) final class Nspace : ScopeDsymbol
     /**
      * Determines whether the symbol for this namespace should be included in the symbol table.
      */
-    bool ignoreCppSymbols;
+    bool mangleOnly;
 
-    extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members, bool ignoreCppSymbols)
+    extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members, bool mangleOnly)
     {
         super(ident);
         //printf("Nspace::Nspace(ident = %s)\n", ident.toChars());
         this.loc = loc;
         this.members = members;
-        this.ignoreCppSymbols = ignoreCppSymbols;
+        this.mangleOnly = mangleOnly;
     }
 
     override Dsymbol syntaxCopy(Dsymbol s)
     {
-        auto ns = new Nspace(loc, ident, null, ignoreCppSymbols);
+        auto ns = new Nspace(loc, ident, null, mangleOnly);
         return ScopeDsymbol.syntaxCopy(ns);
     }
 
     override void addMember(Scope* sc, ScopeDsymbol sds)
     {
-        if(!ignoreCppSymbols)
+        if(!mangleOnly)
             ScopeDsymbol.addMember(sc, sds);
         if (members)
         {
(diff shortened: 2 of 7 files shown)


12345swordy:
You need a DIP for this as well. Contact @TurkeyMan on this, he more then gladly help you write one, as this is a major issue for him.

TurkeyMan:
Oh... my... god.
Submit this patch to LDC, it will be merged immediately.

test/compilable/cppmangle3.d
```diff
@@ -0,0 +1,25 @@
...
+{
+}
+
+extern(C++, "foo")
+{
+}
+
+int foo; // no name clashing with above namespace
+
+extern(C++, "std.chrono")
```
TurkeyMan: I reckon '::' should probably be the separator in this case, since it specifically describes a C++ namespace, and is not necessarily valid D identifiers (as your example), so it probably shouldn't look like a D identifer.

src/dmd/nspace.d
```diff
@@ -30,23 +30,27 @@ private enum LOG = false;
  */
 extern (C++) final class Nspace : ScopeDsymbol
 {
-    extern (D) this(const ref Loc loc, Identifier ident, Dsymbols* members)
+    bool ignoreCppSymbols;
```
TurkeyMan: Can you explain the implementation? This is a weird bool, and it's not clear to me from the diff how it works... (I'm not very familiar with DMD)

TurkeyMan:
Haha, touche :P

TurkeyMan:
More importantly, who is this @look-at-me guy... he's an enigma. Does he have a name?

jacob-carlborg:
> More importantly, who is this @look-at-me guy... he's an enigma

Joined 22 days ago, hmm. https://github.com/look-at-me

test/compilable/cppmangle3.d
jacob-carlborg: Or it can be separate arguments, then we don't have to worry about which separator to use or to parse a string, i.e. `extern (C++, "std", "chrono")`.

TurkeyMan:
https://youtu.be/IhpCfSslq1w?t=30s <- an enigma :)

src/dmd/expression.d
```diff
@@ -2497,7 +2497,7 @@ extern (C++) final class StringExp : Expression
      * Convert string contents to a 0 terminated string,
      * allocated by mem.xmalloc().
      */
-    extern (D) final const(char)[] toStringz() const
+    extern (D) final char[] toStringz() const
```
ibuclaw: Unrelated change I think.

src/dmd/astbase.d
```diff
@@ -4537,6 +4540,14 @@ struct ASTBase
             this.sz = 1;                    // work around LDC bug #1286
         }
 
+        extern (D) final char[] toStringz() const
```
ibuclaw: Why do you need this?

src/dmd/parse.d
```diff
@@ -2152,6 +2154,35 @@ final class Parser(AST) : Lexer
                             cppmangle = token.value == TOK.class_ ? CPPMANGLE.asClass : CPPMANGLE.asStruct;
                             nextToken();
                         }
+                        else if (token.value == TOK.string_) // extern(C++, "namespaces")
+                        {
+                            ignoreCppSymbols = true;
+                            idents = new AST.Identifiers();
+                            AST.StringExp stringExp = cast(AST.StringExp)parsePrimaryExp();
+                            char[] names = stringExp.toStringz();
```
ibuclaw: This is not the appropriate way to get a copy if you intend to alter it (and you need to remove const from the signature in order to do what you want).

src/dmd/parse.d
```diff
@@ -2152,6 +2154,35 @@ final class Parser(AST) : Lexer
...
+                            {
+                                uint i = startIndex;
+                                foreach(ref char c ; names[startIndex .. $])
+                                {
+                                    i += 1;
+                                    if (c == '.')
+                                    {
+                                        if (i != names.length)
+                                        {
+                                            c = '\0';
```
ibuclaw: Don't do this.

src/dmd/parse.d
```diff
@@ -2152,6 +2154,35 @@ final class Parser(AST) : Lexer
...
+                                            break;
+                                        }
+                                        else
+                                            error("C++ namespace expected after `.` at end of `%s`", names.ptr);
+
+                                    }
+                                }
+                                idents.push(new Identifier(&names[startIndex]));
```
ibuclaw: Send a slice to `new Identifier` instead, then you don't need to make things mutable.

ibuclaw (changes requested):

src/dmd/parse.d
look-at-me: I was going to but all the functions in `Identifer` appear to assume that it has a null terminator.

https://github.com/dlang/dmd/blob/v2.082.0/src/dmd/identifier.d#L59

Including the strncmp, unless that `+ 1` is a bug.

src/dmd/parse.d
look-at-me: I was going to, but see my other comment.

src/dmd/parse.d
look-at-me: Hell even this explicitly uses `'\0'`.
https://github.com/dlang/dmd/blob/v2.082.0/src/dmd/identifier.d#L200

look-at-me:
Not an enigma just got locked out of my account :( RIP @skl131313.

src/dmd/parse.d
look-at-me: NVM, seems idPool() is the correct way to create a new Identifer.

test/compilable/cppmangle3.d
TurkeyMan: I'm with Jacob.

src/dmd/nspace.d
TurkeyMan: Ohhhh, there's the magic.
Okay, got it. Thanks!

TurkeyMan:
So, I guess you're doing training for your upcoming trip to mars ;)

TurkeyMan:
@ibuclaw you're reviewing this... but you know it won't be merged on principle.
would you consider merging this to GDC? If both GDC and LDC merge it, then we're golden.

jacob-carlborg:
Should have a changelog entry.

ibuclaw:
> would you consider merging this to GDC?

I do not have man power to deviate from dmd.

TurkeyMan:
@ibuclaw Well, this won't merge, unless there's a coup, and people just merge it... So why are we reviewing this?

src/dmd/nspace.d
jacob-carlborg: All new public symbols should have Ddoc comments.

jacob-carlborg (changes requested):
There should be tests for the all the failing code paths. Have a look at `tests/fail_compilation`

jacob-carlborg:
I think this is small enough to not need a DIP.

look-at-me:
Any reason why there's two instances for the classes? Seems like a pain to have to duplicate code.

Also wanted to look into how the current spec for C++ implements name mangling for their module proposal. If we are going to add a new syntax we should make sure it plays nice with what might end up into C++ at some point. I think clang has an experimental implementation of it already.

thewilsonator:
In addition to failing tests it would be great if @TurkeyMan could build this locally and test it against representative uses (e.g. definition of the same namespace in multiple files etc.) so that there will be less (hopefully no) bug reports associated with this.

thewilsonator:
> Any reason why there's two instances for the classes? 

I _think_ this is for 
```
extern(C++, foo)
{
    // A scope
    void someSymols();
}
```

and 

`extern(C++, bar) void noScopeOneSymbol()`, but I'm not sure.

look-at-me:
I mean all the classes in astbase.d which are duplicated. They are never built for DMD, at least for me locally. But they are in the tests. Not sure why two instances of the classes exist when they are similar for almost all the types.

```D
extern(C++, "std", "chrono")
{
    extern(C++, bob)
    {
        void func(); //mangles as std::chrono::bob::func()
    }
    // __gshared int bob; // error, symbol conflict between namespace and var as expected
}

int bob; // ok
```
Nothing unexpected but I know too much about how they work. Would be less confusing to only use one of the two options. That'd have to be the users choice though.

thewilsonator:
I think thats for front end as a library. @RazvanN7 ?

thewilsonator:
This will also need a spec PR updating the grammar before it can be merged, no hurry though.

TurkeyMan:
@thewilsonator Happy to give it a whirl... but I've been super busy. Might be able to have a go tomorrow.

thewilsonator:
Good. I don't need to tell you that it in your interests to make sure it works for what you need it for ;)

RazvanN7:
>I think thats for front end as a library. @RazvanN7 ?

Indeed, astbase defines the minimum entities that are required to build the parser separately from dmd. The end objective is to refactor the dmd codebase so that astbase is reduced to just import all ast nodes. Progress on that has been made by refactoring out the semantic methods out of the ast nodes, but there is still a lot to be done and I haven't been working on this continuously.

ibuclaw:
> The whole front end is usable as a library today.

As it has been 8 years ago too, to varying level of degrees - otherwise gdc and ldc wouldn't be able to use it.

ibuclaw:
> @ibuclaw Well, this won't merge, unless there's a coup, and people just merge it... So why are we reviewing this?

If it's going in, it might as well be implemented properly.

I don't really agree with adding new syntax, but I don't have any strong feelings against either, so I'll abstain from any decision about the feature itself.

TurkeyMan:
We've been trying to get this feature for about 6 years... It's really, really important >_<

look-at-me:
@TurkeyMan 
If you want to look over the DIP and make any suggestions, or anyone else. The draft is here: https://github.com/look-at-me/DIPs/blob/please-for-the-love-of-all-things-good-and-holy-fix-cpp-mangling-please-ok-sorry/DIPs/DIPxCPP.md

TurkeyMan:
> @thewilsonator: I don't need to tell you that it in your interests to make sure it works for what you need it for ;)

Just FYI, I have tested this fairly extensively. I am overwhelmed with a sense of relaxation and relief as I realise everything is as it should be!

Sadly it all comes crashing down when I switch branch ;)

TurkeyMan:
Fixes issue 15512 - (Just FYI)

src/dmd/astbase.d
```diff
@@ -4537,6 +4543,44 @@ struct ASTBase
             this.sz = 1;                    // work around LDC bug #1286
         }
 
+        /**********************************************
+        * Write the contents of the string to dest.
+        * Use numberOfCodeUnits() to determine size of result.
```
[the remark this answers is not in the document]
look-at-me: That function  and doc is from the expression side https://github.com/dlang/dmd/blob/v2.082.0/src/dmd/expression.d#L2293 and it looks like it does use numberOfCodeUnits, but it reimplements it and only actually converts for most basic case.

src/dmd/parse.d
```diff
@@ -2152,6 +2154,42 @@ final class Parser(AST) : Lexer
...
+                                const(char)[] name = stringExp.toStringz();
+                                if(name.length == 0)
+                                {
+                                    error("invalid zero length C++ namespace");
+                                    idents = null;
+                                    break;
+                                }
+                                else if(!Identifier.isValidIdentifier(name))
+                                {
+                                    error("C++ namespace `%s` is invalid", name.ptr);
```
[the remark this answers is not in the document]
look-at-me: I think the rules for what is a valid identifier is well known. I changed the wording a bit to reflect that.

WalterBright:
Need a spec pull, too.

Geod24:
Build it and they will ~come~ merge!
This was missing [a little something](https://github.com/dlang/dmd/pull/8714).
@look-at-me : Let us know what your plans for the spec PR

src/dmd/parse.d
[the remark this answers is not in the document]
jacob-carlborg: There might be words that are keywords in C++ but not in D. How are those handled?

jacob-carlborg:
Well, look at that 😃.

src/dmd/astbase.d
```diff
@@ -4537,6 +4543,44 @@ struct ASTBase
...
+        * Params:
+        *  dest = destination
+        *  tyto = encoding type of the result
+        *  zero = add terminating 0
+        */
+        void writeTo(void* dest, bool zero, int tyto = 0) const
```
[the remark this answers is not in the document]
look-at-me: It was already present in the expression side of the symbol. I kept the same signature as it seems some parts of the code run for both classes.

https://github.com/dlang/dmd/blob/v2.082.0/src/dmd/expression.d#L2293

look-at-me:
@Geod24 
Yah I can create that PR for the spec, wasn't expecting this to be merged :P.

src/dmd/astbase.d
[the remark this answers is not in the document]
BBasile: Sorry, but anyway i've written my review and didn't see it was already merged. Just ignore it.

src/dmd/parse.d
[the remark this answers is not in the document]
look-at-me: Well it could depend on the C++ version, `constexpr` for example wasn't a keyword until C++11 same goes for `declytype`. You'd get a link error if you did try to use a C++ keyword, from the C++ side you would just receive an error. I don't know if it is worth maintaining that sort of list in D.

TurkeyMan:
@look-at-me I've been further experimenting with this patch (now it's merged, I ported everything).
I noticed that error messages showing symbol names are wrong.
For s symbol `extern(C++, "ns") int x;`, any error message involving `x` will print the name as `module.ns.x`, which is wrong, since it's not in an `ns` scope.
I guess the logic that produces full symbol names isn't aware of your flag.

WalterBright:
@TurkeyMan file a bug report!

look-at-me:
@TurkeyMan 
I'll look into it tomorrow, I ran into a few bugs after I updated DMD and used the master branch.

TurkeyMan:
@look-at-me There's another one; it doesn't attempt to evaluate StringExp's... infact, if it's not a literal, it's rejected.

PR #8670, opened 2018-09-06
Make Identifier hold a slice internally

Geod24:


src/dmd/identifier.d
```diff
@@ -30,17 +30,24 @@ extern (C++) final class Identifier : RootObject
...
 public:
+    /**
+       Construct an identifier from a D slice
+
+       Note: Since `name` needs to be `\0` terminated for `toChars`,
+       no slice overload is provided yet.
 
+       Params:
+         name = the identifier name
+                There must be `\0` a termination at name[$].
```
jacob-carlborg: Can an assert be added to verify this?

Geod24: Nope, since it's out of bounds

src/dmd/identifier.d
```diff
@@ -56,28 +63,28 @@ public:
...
     override bool equals(RootObject o) const
     {
-        return this == o || strncmp(name, o.toChars(), len + 1) == 0;
+        return this == o || name == o.toString();
     }
 
     override int compare(RootObject o) const
     {
-        return strncmp(name, o.toChars(), len + 1);
+        return strncmp(name.ptr, o.toChars(), name.length + 1);
```
jacob-carlborg: It would be nice if we could get rid of these functions that operate on C strings.

Geod24: What should we replace `strncmp` with ?

src/dmd/identifier.d
jacob-carlborg: So how can the C string returned by `toChars` be used?

src/dmd/identifier.d
jacob-carlborg: Using the regular comparison operators?

Geod24: Uh, didn't even think it'd work.

However doesn't it include two comparison ? I can write `"a" < "b"` but I cannot get the integer value that I need to return here. So I'd need to test for comparison as well. Am I missing something ?

jacob-carlborg: I guess.

src/dmd/identifier.d
Geod24: Oh right, this is a pointer 🤦‍♂️ 
So we could add it, indeed.

However I'm a bit wary of doing so though, as this might quickly proliferate all over the place. In fact there is no precedent for it.
This is just adding documentation, not changing this part of the code. If anything, it reduces the dependency on `\0`.

jacob-carlborg: ok

src/dmd/identifier.d
jacob-carlborg: Or it can be extracted to a utility function, I'm sure there are other places that need this as well.

RazvanN7 (approved):

thewilsonator (changes requested):

src/dmd/identifier.d
```diff
@@ -30,17 +30,24 @@ extern (C++) final class Identifier : RootObject
 {
 private:
     const int value;
-    const char* name;
-    const size_t len;
+    const char[] name;
```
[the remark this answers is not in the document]
Geod24: I did, didn't I?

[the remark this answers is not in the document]
jacob-carlborg: Yes, you did.

jacob-carlborg (approved):

thewilsonator:
```
Error downloading remote file: One or more errors occurred.
Inner Exception: Unable to connect to the remote server
```

Geod24:
@thewilsonator : See Slack's `#ci`. `nightlies.dlang.org` is broken :(

thewilsonator:
Ah.

src/dmd/identifier.d
```diff
@@ -57,28 +64,28 @@ public:
...
     {
-        return strncmp(name, o.toChars(), len + 1);
+        return strncmp(name.ptr, o.toChars(), name.length + 1);
     }
 
 nothrow:
     override void print() const
     {
-        fprintf(stderr, "%s", name);
+        fprintf(stderr, "%.*s", name.length, name.ptr);
```
WalterBright: Must cast `name.length` to `int`.

WalterBright: see documentation on `printf` format strings.

Geod24: Are those print method actually useful ? Shouldn't we just use `toChars` / `toString` everywhere ?

Geod24: Asking because if not, I might submit a separate PR to get rid of them

WalterBright (changes requested):
`%.*s` must have `int` argument

src/dmd/identifier.d
WalterBright: Probably not. Mixing I/O into a class is usually a failure of encapsulation.

Geod24:
Updated:
```diff
diff --git a/src/dmd/identifier.d b/src/dmd/identifier.d
index 794deb81c..d06a32f12 100644
--- a/src/dmd/identifier.d
+++ b/src/dmd/identifier.d
@@ -75,7 +75,7 @@ public:
 nothrow:
     override void print() const
     {
-        fprintf(stderr, "%.*s", name.length, name.ptr);
+        fprintf(stderr, "%.*s", cast(int)name.length, name.ptr);
     }

     override const(char)* toChars() const pure
```

WalterBright (approved):

src/dmd/identifier.d
```diff
@@ -31,17 +31,24 @@ extern (C++) final class Identifier : RootObject
...
 
 public:
+    /**
+       Construct an identifier from a D slice
+
+       Note: Since `name` needs to be `\0` terminated for `toChars`,
+       no slice overload is provided yet.
 
+       Params:
+         name = the identifier name
```
ZombineDev: It seems that DAutoTest fails with:
```
dmd/identifier.d(47): Warning: Ddoc: parameter count mismatch
```

Geod24:
Auto-merge toggled on

PR #8673, opened 2018-09-06
Fix issue 18946:  assert message can throw hijacking the assert failure

thewilsonator:
I presume the protocol is to deprecate this first.

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -5369,6 +5369,11 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
                 return;
             }
         }
+
+        // https://issues.dlang.org/show_bug.cgi?id=18946
+        if (exp.msg && canThrow(exp.msg,sc.func,false))
+            exp.msg.deprecation("`%s` assert error messages must not throw",exp.msg.toChars());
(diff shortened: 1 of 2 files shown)


Geod24:
When submitting a PR, please provide a title which allow us to see what it's about from the notification.

Geod24:
I am not sure the linked issue is valid. In the end, your exception happens last, and that's what gets reported. If the function was to `assert`, that would also be reported, wouldn't it ?

thewilsonator:
Instead of throwingFunc replace that with a call to format or the like, which one might reasonably expect to catch from some higher call chain call...

Geod24:
@thewilsonator : Exactly. But forcing people to put `nothrow` on their string-formatting functions is quite excessive IMO.

thewilsonator:
Hmm, perhaps this is not the correct way to fix this. It would be better to wrap the message evaluation in a try catch.

Geod24:
Yeah, that would be much better

Geod24:
If you don't mind, I'm going to close this for the time being since this approach won't work.
This way it won't get scheduled into the CI queue.
Feel free to reopen when you implement the different approach.

PR #8675, opened 2018-09-06
Fix Issue 18771 - Identical overload sets in different modules have different identities

RazvanN7:
__traits(isSame) simply follows the alias chain and sees if the symbols resolve to the same AST node. In the case of overload sets in different files that does not work anymore and the components of the overload sets need to be compared for equality. The current patch compares the overload sets by comparing each element of an one to each of the elements of the other in O(n*n) complexity; that is because the elements of the overload set are stored in an array of Dsymbols. A smarter approach would be to use a hashtable for the components of the overload set, but that would lead to major refactorizations.

### DIFF
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -1646,7 +1646,36 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
         if (cmp(s1,s2) || cmp(s2,s1))
             return True();
 
-        return (s1 == s2) ? True() : False();
+        if (s1 == s2)
+            return True();
+
+        // https://issues.dlang.org/show_bug.cgi?id=18771
+        // OverloadSets are equal if they contain the same functions
+        auto overSet1 = s1.isOverloadSet();
+        if (!overSet1)
+            return False();
+
+        auto overSet2 = s2.isOverloadSet();
+        if (!overSet2)
+            return False();
+
+        if (overSet1.a.dim != overSet2.a.dim)
+            return False();
+
+        // OverloadSets contain array of Dsymbols => O(n*n)
+        // to compare for equality as the order of overloads
+        // might not be the same
+Lnext:
+        foreach(overload1; overSet1.a)
+        {
+            foreach(overload2; overSet2.a)
+            {
+                if (overload1 == overload2)
+                    continue Lnext;
+            }
+            return False();
+        }
+        return True();
     }
     if (e.ident == Id.getUnitTests)
     {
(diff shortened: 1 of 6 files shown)


src/dmd/traits.d
```diff
@@ -1645,7 +1645,38 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
-        return (s1 == s2) ? True() : False();
+        if (s1 == s2)
+            return True();
+
+        // https://issues.dlang.org/show_bug.cgi?id=18771
+        // OverloadSets are equal if they contain the same functions
+        auto overSet1 = s1.isOverloadSet();
+        auto overSet2 = s2.isOverloadSet();
+        if (overSet1 && overSet2)
+        {
```
thewilsonator: Can you invert the clauses so that the return are all early returns? (this reduces indentation
e.g.
```
if (!overSet1 || !overSet2)
    return False();
else if (overSet1.a.dim != overSet2.a.dim)
    return False();
```

src/dmd/traits.d
```diff
@@ -1645,7 +1645,38 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+            // to compare for equality as the order of overloads
+            // might not be the same
+            foreach(elem1; overSet1.a)
+            {
+                bool flag;
+                foreach(elem2; overSet2.a)
+                {
+                    if (elem2 == elem1)
+                    {
+                        flag = true;
```
thewilsonator: Any reason to not just `return False();` here?

src/dmd/traits.d
```diff
@@ -1645,7 +1645,38 @@ extern (C++) Expression semanticTraits(TraitsExp e, Scope* sc)
...
+        auto overSet2 = s2.isOverloadSet();
+        if (overSet1 && overSet2)
+        {
+            if (overSet1.a.dim != overSet2.a.dim)
+                return False();
+
+            // OverloadSets contain array of Dsymbols => O(n*n)
+            // to compare for equality as the order of overloads
+            // might not be the same
+            foreach(elem1; overSet1.a)
```
WalterBright: Calling an overload set symbol an "elem" is rather confusing, as elem is used as a type.

src/dmd/traits.d
WalterBright: Even better would be just label the outer loop, then use `continue label;` here. No `flag` is required.

RazvanN7:
@thewilsonator @WalterBright I addressed your comments.

jacob-carlborg (approved):

PetarKirov (approved):

ZombineDev:
Auto-merge toggled on

PR #8676, opened 2018-09-08
[DO NOT MERGE] Have __xtoHash delegate to core.internal.hash.hashOf so the two match

n8sh:
https://github.com/dlang/druntime/pull/2243 was intended to make `typeid(x).getHash(&x)` match `hashOf(x)` but did not account for structs with auto-generated `__xtoHash`.

### DIFF
--- a/src/dmd/clone.d
+++ b/src/dmd/clone.d
@@ -806,14 +806,26 @@ FuncDeclaration buildXtoHash(StructDeclaration sd, Scope* sc)
      * hash value will also contain the result of parent class's toHash().
      */
     const(char)* code =
-        "size_t h = 0;" ~
-        "foreach (i, T; typeof(p.tupleof))" ~
-        // workaround https://issues.dlang.org/show_bug.cgi?id=17968
-        "    static if(is(T* : const(.object.Object)*)) " ~
-        "        h = h * 33 + typeid(const(.object.Object)).getHash(cast(const void*)&p.tupleof[i]);" ~
-        "    else " ~
-        "        h = h * 33 + typeid(T).getHash(cast(const void*)&p.tupleof[i]);" ~
-        "return h;";
+    `static if (!__traits(isDeprecated, typeof(p))) ` ~
+    `{ ` ~
+        // Don't use this branch when S is deprecated because if compiling
+        // with deprecations as errors compilation will fail.
+        `scope(failure) assert(0); ` ~ // For `nothrow`.
+        `static import core.internal.hash; ` ~
+        `return core.internal.hash.hashOf(p); ` ~
+    `} ` ~
+    `else ` ~
+    `{ ` ~
+        // Fall back to behavior using TypeInfo.
+        `size_t h;` ~
+        `foreach (i, T; typeof(p.tupleof)) ` ~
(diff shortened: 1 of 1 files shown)


jacob-carlborg (approved):

wilzbach:
Seems like `hashOf` would need to be `nothrow`:

```
Error: function `core.internal.hash.hashOf!(const(Demangle!(PrependHooks))).hashOf` is not `nothrow`
Error: `nothrow` generated function `core.demangle.Demangle!(PrependHooks).Demangle.__xtoHash` may throw
src/core/demangle.d(2082): Error: template instance `core.demangle.Demangle!(PrependHooks)` error instantiating
src/core/demangle.d(2066): Error: template instance `object.RTInfo!(Replacement)` error instantiating
Error: function `core.internal.hash.hashOf!(const(PrependHooks)).hashOf` is not `nothrow`
Error: `nothrow` generated function `core.demangle.reencodeMangled.PrependHooks.__xtoHash` may throw
```

n8sh:
The problem is that `PrependHooks` has an AA and `_aaApply2` is not nothrow.

n8sh:
For some reason this patch makes DMD segfault when compiling `test/runnable/aliasthis.d`.

thewilsonator:
That was merged, does this PR need to be resuscitated?

n8sh:
This PR shouldn't be merged before https://github.com/dlang/druntime/pull/2289 or it will cause compilation failure for every program with a struct that needs `__xtoHash` and has a field that's a vector.

n8sh:
Compilation currently failing for structs that have a non-const `toHash` and don't have any non-bitwise-hashable members, e.g.
```d
struct NotOkay
{
    int canBitwiseHashInts;
    size_t toHash() /* not const */ { return 0; }
}

struct Okay
{
    float cantBitwiseHashFloats;
    size_t toHash() /* not const */ { return 0; }
}

void main()
{
    const okay = Okay.init;
    hashOf(okay); // Compiles but writes a message about the non-callable toHash.
    
    const notOkay = NotOkay.init;
    hashOf(notOkay); // Compilation fails!
}
```
I need to fix that in druntime before this PR can move forward. Maybe `toHash` should always require struct arguments to be const, although that causes some annoyance when the struct has members that are objects.

n8sh:
Now waiting on https://github.com/dlang/druntime/pull/2337.

thewilsonator:
Does this need any other druntime PRs? Also please rebase.

n8sh:
Currently failing if compiling with `-de` (which makes of deprecated features an error) if anywhere in the program is defined a `@deprecated` struct that needs `__xtoHash`.

n8sh:
Is there any way to suppress deprecation?

n8sh:
While examining why link11931 wasn't compiling I ran into this bug: https://issues.dlang.org/show_bug.cgi?id=19400. It seems to be a distinct bug from 11931 and 12099 since it occurs without "-g" and without mutually dependent modules.

RazvanN7:
@n8sh the mentioned issue is closed as invalid. is there anything else besides a rebase that is blocking this PR?

PR #8677, opened 2018-09-09
Fix issues 1870 and 12790: Write out string mixin code to a file for debugging purpose

thewilsonator:
There is no test case for this yet, I'm not sure how to check the contents of generated files with dmd's test suite. I also don't know how to test debug info, any help appreciated.

cc @TurkeyMan


BBasile:
doesn't the source file mars.d need to be updated to activate the new option ?

thewilsonator:
True I forgot to sync before opening the PR

src/dmd/cli.d
```diff
@@ -239,6 +239,9 @@ dmd -cov -unittest myprog.d
             compiling for symbolic debugging instead of $(B $(LIB)).
             If $(I libname) is not supplied, then no default library is linked in.`
         ),
+        Option("debugmixin",
+            "write out arguments to mixins so that the debuuger has source"
```
drug007: typo `debugger`

RazvanN7:
Does it really need to be a compiler switch? Why not do this by default whenever mixins are used? @WalterBright is usually against adding new switches. How hard is it to simply print the offending line with any line number? 

Also, if this gets merged it needs a changelog entry, updating the documentation and also the messages that dmd outputs if no parameter is provided.

src/dmd/globals.d
```diff
@@ -151,6 +151,7 @@ struct Param
                             // https://issues.dlang.org/show_bug.cgi?id=16997
     bool vsafe;             // use enhanced @safe checking
     bool ehnogc;            // use @nogc exception handling
+    bool debugMixins;       // write out argument to mixin so that the debuuger has source
```
jacob-carlborg: Ddoc comment please.

src/dmd/parse.d
```diff
@@ -247,9 +248,23 @@ final class Parser(AST) : Lexer
             /* Create a pseudo-filename for the mixin string, as it may not even exist
              * in the source file.
              */
-            char* filename = cast(char*)mem.xmalloc(strlen(loc.filename) + 7 + (loc.linnum).sizeof * 3 + 1);
-            sprintf(filename, "%s-mixin-%d", loc.filename, cast(int)loc.linnum);
-            scanloc.filename = filename;
+            auto len = strlen(loc.filename) + 7 + (loc.linnum).sizeof * 3 + 1;
```
jacob-carlborg: Looks like this can be const.

src/dmd/parse.d
```diff
@@ -247,9 +248,23 @@ final class Parser(AST) : Lexer
...
+            char[] filename = cast(char[])mem.xmalloc(len)[0 .. len];
+            sprintf(filename.ptr, "%s-mixin-%d", loc.filename, cast(int)loc.linnum);
+
+            // If we are debugging mixins, write `input` out to a file so that the debugger has some
+            // source to look at.
+
+            if (global.params.debugMixins)
+            {
+                auto tmp = FileName.combine(global.params.objdir,filename.ptr);
```
jacob-carlborg: Looks like this can be const.

src/dmd/parse.d
```diff
@@ -247,9 +248,23 @@ final class Parser(AST) : Lexer
...
+                filename = cast(char[])tmp[0 .. strlen(tmp)];
+                auto f = File.create(filename.ptr);
```
dhasenan: This is certainly better than the status quo.

However, consider the following code:

void foo(string s)() { mixin(s); }
void main()
{
  foo!(`writeln(a);`);
  foo!(`writeln("hello world");`);
}

This will write `writeln(a);` to file `scratch.d-mixin-5`, then write `writeln("hello world");` to the same file, overwriting what was there previously.

WalterBright (changes requested):
When creating PRs for bugzilla issues, please include a link in the issue to the PR. This prevents multiple people from unwittingly working on the same issue. And even if it is rejected, it provides valuable information to anyone trying again.

I'm tired of repeatedly pointing this out to everyone making PRs.

thewilsonator:
@RazvanN7 This will result in a lot of unnecessary I/O for projects with lots of mixins, not to mention disk space. This is also so that the debugger has source so it needs to be on disk for that.

> Also, if this gets merged it needs a changelog entry,

Will do.

> updating the documentation

isn't that what `cli.d` is for?

> the messages that dmd outputs if no parameter is provided.

This is a parameterless switch.

src/dmd/cli.d
WalterBright: Write out where? I suggest indicating that it is written to a file.

Geod24:
> When submitting a PR, please provide a title which allow us to see what it's about from the notification.

Also, is there no better way to provide this to GDB than writing it to a file ? @ibuclaw ?

thewilsonator:
As I understand it, the problem is that source locations are provided to the debugger (GDB/LLDB/whatever MS uses) but the source doesn't exist on disk, because its expanded at compile time, and so `b somewhereinamixin` will work, but `show me the source where I am` fails.

Geod24:
I understand the problem, and agree that it needs fixing, I'm just wondering if debuggers have another way to cope with this than writing a file to disk.
And if we do end up writing a file to disk, I think it should live alongside objects file.

thewilsonator:
Can you embed source on an object file? I can't think of any other way to do it.

> And if we do end up writing a file to disk, I think it should live alongside objects file.

Hence this is written to`objdir`.

src/dmd/parse.d
```diff
@@ -247,9 +248,23 @@ final class Parser(AST) : Lexer
             /* Create a pseudo-filename for the mixin string, as it may not even exist
              * in the source file.
              */
-            char* filename = cast(char*)mem.xmalloc(strlen(loc.filename) + 7 + (loc.linnum).sizeof * 3 + 1);
-            sprintf(filename, "%s-mixin-%d", loc.filename, cast(int)loc.linnum);
-            scanloc.filename = filename;
+            auto len = strlen(loc.filename) + 7 + (loc.linnum).sizeof * 3 + 1;
+            char[] filename = cast(char[])mem.xmalloc(len)[0 .. len];
```
Geod24: Cast to a pointer then slice. Casting arrays incur additional checks.

src/dmd/parse.d
```diff
@@ -247,9 +248,23 @@ final class Parser(AST) : Lexer
...
+            // If we are debugging mixins, write `input` out to a file so that the debugger has some
+            // source to look at.
+
+            if (global.params.debugMixins)
+            {
+                auto tmp = FileName.combine(global.params.objdir,filename.ptr);
+                filename = cast(char[])tmp[0 .. strlen(tmp)];
+                auto f = File.create(filename.ptr);
+                // We're going to write it out straight away so the cast below is fine
+                f.setbuffer(cast(void*)input.ptr,input.length);
```
Geod24: Nit: space after comma (other places in the diff as well)

src/dmd/parse.d
Geod24: Use `new File(filename.ptr)`. `create` is a C++ wrapper.

thewilsonator: Is that likely to ever occur in real code?

thewilsonator: Actually doesn't need to even be `new`'d

thewilsonator:
Addressed all comments except doc comment for `global.params.debugMixins` and collision between mixing in a template parameter. Added changeling and tests for 1879/12790. Haven't added a test for debug info generation

src/dmd/parse.d
dhasenan: Any mixin that occurs inside a template that's instantiated more than once. That might be a minority of cases, but it certainly does occur. For instance, std.typecons has at least a handful of these. I wrote some code to automatically extract a postgresql row into a struct, and it contains mixins inside templates that are instantiated multiple times.

Also the standard shortcut for implementing opBinary is with mixins. That's why we switched to opBinary!operator, instead of opAdd and friends.

TurkeyMan: > Is that likely to ever occur in real code?

I think it happens more often than not.
In my experience, most string mixins are wrapped in a mixin template for sanitary reasons, or in a template function. If the enclosing scope is not a template, then there's little motivation for a string mixin... you could just put the code directly. I reckon it's the common case for a mixin to appear in a parametric environment.

TurkeyMan:
I don't want to add friction here, but I think a great implementation would emit enclosing scope in addition to the mixin expansion. That way, when you step and the debugger jumps into the mixin file, you don't suddenly lose visibility of the surrounding context.

TurkeyMan:
It would also be quite trivial for @dlang-bot to create these links automatically. I think that's the right thing to do here.

thewilsonator:
Garrgh:
```
Test fail_compilation/debugmixins.d failed: 
expected:
----
test_results/fail_compilation/debugmixins.d-20.mixin(3): Error: undefined identifier `b`
----
actual:
----
generated/fail_compilation/fail_compilation/debugmixins.d-20.mixin(3): Error: undefined identifier `b`
----
```
I can't appease both semaphore and the autotester, how does one do partial matching in this test suite? Alternately can some show me where in the ci scripts to make everything go to `generated/`?

TurkeyMan:
@thewilsonator I mean, if the debugger jumps into a `.mixin` file with a single line or 2 of expanded mixin, and that's just a fragment appended to the file after a bunch of other fragments, it will feel very jankey. When you step into a line expanded from a mixin, it will be ideal to have the surrounding scope also emit to the mixin file for context. You'll be able to hover over nearby text and inspect values of local variables, even if they're not a part of the mixin expansion.

thewilsonator:
Ah, I get you. Not sure how to do that though.

ibuclaw:
Why can't we just fix the line numbers of mixins to be their location in the source file?

Sure, string mixins can look convoluted when stepping through them, but it's better than no location.

TurkeyMan:
@ibuclaw I'm not sure what you mean? There is no source typically. Any mixin of meaningful complexity was probably generated.

ibuclaw:
@TurkeyMan but the strings come from somewhere, even if joined from two different places.

Stepping through debug could point to those string locations, rather than a fake location.

thewilsonator:
Well what happens with macros in C/C++?

TurkeyMan:
@ibuclaw they're more likely formatted, and composed from a complex call tree. At least, they usually are when I use them.

@thewilsonator C++ macros aren't really anything like mixin.

TurkeyMan:
Sorry, to answer; C++ doesn't really do anything... If the macro represents multiple lines of code, it just points at the #define line pointlessly, and since macros require `\` to join multiple lines into one line, it can only ever point at the first line.

Accepting that we do a terrible job because C++ does a terrible job isn't very classy ;)
We need to demonstrate superior experience to C++ in general. I'm excited to try out what you've done here and see how it feels.

thewilsonator:
> C++ doesn't really do anything

 Ah, I was wondering if they did anything worthwhile copying, apparently not. Let me know how this goes for using a debugger, I haven't tested it. 

Note that this PR is effectively blocked until the CI can be made to be consistent in terms of object output directory (and also pass). I'm really out of my depth for that though.

TurkeyMan:
I've been trying out this PR. I think emitting separate files is not sustainable.

All mixin fragments need to emit to one file named for the object file that's being written rather than many files named for each mixin instantiation.
As-is leads to *thousands* of 1-line files being written to my object directory, many of which are emit by just importing modules from phobos.

TurkeyMan:
More issues:
- The debuginfo is populated such that it points to the `.mixin` files even if `-debugmixins` is not supplied, which causes the debugger to complain about missing source files. Debuginfo should not refer to files that weren't created.
- The `.mixin` file is not written to the same location as the object file (they should be). It should live right beside any `.o/obj` by default.
  - Possibly accept an additional command line option to specify an output location, or perhaps an output file path allowing the user to override the default output filename.
  - Alternatively just have a single option, which is all-in-one, eg; `-debugmixin=./obj/release_x64/objfilename.mixin`

Other than those, it works in the debugger well, and this will be VERY valuable!

thewilsonator:
> I think emitting separate files is not sustainable.

Hmm, one file per mixin is easy, one file total is easy, one `.mixin` per `.o` is not, one per "instantiating" module might be doable.

> The debuginfo is populated such that it points to the .mixin files even if -debugmixins is not supplied,

This is no change from the status quo, right? I could make it implied by `-g`. Also does `-g` produce `.pdb` an the like?

> The .mixin file is not written to the same location as the object file (they should be)

Damn right they should be, I was writing them to the object directory. I wonder what happened there?

TurkeyMan:
> one file per mixin is easy, one file total is easy, one .mixin per .o is not, one per "instantiating" module might be doable.

But... one invocation of the compiler will only emit one output. What makes it hard to emit one `.mixin` beside the `.o` file?
Ie, one compiler execution -> one `.o` file... so, one execution -> one `.mixin` file?

> > The debuginfo is populated such that it points to the .mixin files even if -debugmixins is not supplied,
>
> This is no change from the status quo, right? I could make it implied by -g. Also does -g produce .pdb an the like?

I'm not sure what you mean here... there is no status-quot, since there are no `.mixin` files, and no debugging.
Basically, I suggest, if `-debugmixin` is not supplied, do not emit such `#file`/`#line` information to inform the debuginfo that the source is in the external file?
I think the default behaviour is that code introduced by a mixin naturally points to the mixin statement(?)

> > The .mixin file is not written to the same location as the object file (they should be)
>
> Damn right they should be, I was writing them to the object directory. I wonder what happened there?

Dunno, but in my tests, they ended up in the cwd (next to the source), and not where the output was written.
I do think it would be useful to control the output filename though... so maybe make that a thing?

src/dmd/parse.d
```diff
@@ -247,9 +249,44 @@ final class Parser(AST) : Lexer
             /* Create a pseudo-filename for the mixin string, as it may not even exist
              * in the source file.
              */
-            char* filename = cast(char*)mem.xmalloc(strlen(loc.filename) + 7 + (loc.linnum).sizeof * 3 + 1);
-            sprintf(filename, "%s-mixin-%d", loc.filename, cast(int)loc.linnum);
-            scanloc.filename = filename;
```
TurkeyMan: Oh yeah, it did do this didn't it. This is why I've always been so annoyed ;)
Ideally, for cases where not `-debugmixin`, it would set the file/line to point right at the mixin statement, rather than hassling about an artificial filename that doesn't exist.

thewilsonator: Yes, thats why i was a little confused. That should be doable.

test/runnable/test18322.d
```diff
@@ -13,9 +13,9 @@ void main(){
     auto filename = "runnable" ~ sep ~ "test18322.d";
 
     fun(filename);
-    mixin(`fun(filename ~ "-mixin-16");`);
+    mixin(`fun(filename ~ "-16.mixin");`);
 
     #line 100 "poundlinefile.d"
     fun("poundlinefile.d");
-    mixin(`fun("poundlinefile.d-mixin-101");`);
+    mixin(`fun("poundlinefile.d-101.mixin");`);
```
TurkeyMan: "poundline" ... wow, that sounds weird! ;)

src/dmd/parse.d
TurkeyMan: I'd consider it a stretch goal ;) .. There's no actual regression if we don't get this in there.

TurkeyMan: Nobody will notice, because as soon as `-debugmixin` exists, there would be no reason not to use it whenever you use `-g`.

thewilsonator:
>But... one invocation of the compiler will only emit one output. What makes it hard to emit one .mixin beside the .o file?
Ie, one compiler execution -> one .o file... so, one execution -> one .mixin file?

Now I'm confused. Let me back up a bit.

With Rainers solution you get exactly one `.mixin` for one invocation of `dmd`.
With the current solution you get one file per mixin (statement/expression).

You are suggesting one `.mixin` foreach `.o`. 
The problem is, as much as it may seem weird, `.d -> .o` is not a straight forward mapping, nor is it invertible (templates make that really weird,they end up where you use them, not where they are defined; modules that don't have a correspond `.o` get shoved into the first module on the command line's `.o; probably other stuff I'm not aware of). 

Whereas `.d -> .mixin` is a lot simpler and solves the morass of files problem of the current solution.

src/dmd/parse.d
thewilsonator: I'll make `-g` imply `-debugmixins`. The reverse in not necessarily true. e.g. when trying to decipher why a doubly (or more) nested mixin does not compile one does not need debug info one needs source, so this will remain a separate flag.

TurkeyMan:
Right... but I think you're overthinking it.

I think it should be simply:
One dmd -> one `.o` + one `.mixin`.

Every module supplied on the command line for compilation, and all template instantiations from imported modules get shoved into the one `.o` file, and I expect the same for the `.mixin` too.
Every module, and every template instantiation all append to a single `.mixin` file beside the `.o` file.

If you run dmd twice, and produce two `.o` files, mixins for template instantions will appear in both `.mixin` files for each `.o` file. That's totally fine.
I feel like anything more than that is just unnecessary I/O.

But probably most importantly, if the compiler emits more than one `.mixin` file, then it's very hard for the build system to provide output file names, or expect what files (or how many) files were written.
This is a nightmare for dependency tracking within build environments. 1:1 rule->output file is good, just like for `.o` files.

TurkeyMan:
> they end up where you use them, not where they are defined

Right, and looking at is this way; the `.o` is 'where you use them'. Where they're defined is not interesting. What's relevant is the mixin source that matches the `.o` file that the debugger will go looking for.
Link a `.o` file, and you need that particular source file (and none others). The `.o` and `.mixin` are coupled.

src/dmd/parse.d
TurkeyMan: While I find it agreeable, I'm not sure that's a good idea. I think you'll receive strong resistance.
I have a sense that the compiler probably shouldn't write any more output than it was told to...

TurkeyMan:
I guess I imagined a process like:
On program entry, allocate a buffer to accumulate mixin fragments.
Any mixin instantiation will append text to this buffer, and record the line number (+filename) to the debuginfo. (add a couple of newlines between each append)
When compilation is successful and the compiler writes the `.o` file, also write the accumulated text buffer to a `.mixin` file.

thewilsonator:
The other problem is what to do with `__traits(compiles, mixin("blah"))`, that doesn't end up anywhere? I think just a `.d ->.o` is fine since thats only useful for compilation error debugging.

TurkeyMan:
I mean, I'm not sure 'what symbols go where' is part of the problem? We just don't care... just append text to the global mixin buffer :)

thewilsonator:
So you want Rainers' solution?

TurkeyMan:
> [...] __traits(compiles, mixin("blah")) [...]

Do nothing, just allow that instantiation to appear in the file for now. It does no harm.

We can power this up in the future, but MVP should be the smallest patch possible, otherwise it'll get bogged down reviewing/bike-shedding and testing ;)

TurkeyMan:
I'm not actually sure what "Rainers solution" is exactly... but if it's like I say, then that is probably what I'm asking for :)

It all comes back to: the `.o` and `.mixin` are strongly coupled. Any effort to decouple them is not particularly useful, and only creates future friction for tooling/build environments. I think it's most convenient that they be a reliable and associated pair.

thewilsonator:
Rainers solution is: one invocation of dmd gives one `.mixin`. There is therefore no relation between `.mixin.` and `.o` because there is only one `.mixin`.

TurkeyMan:
That's the same relationship that one invocation has with `.o` though... one invocation of the compiler produces one object file...? Their relationship is implicit?

TurkeyMan:
It sounds like you're saying that the compiler may emit multiple `.o` files? I'm not sure how that's possible..?

thewilsonator:
> one invocation of the compiler produces one object file

If only one module is on the command line. You can definitely produce one `.o` foreach module on the command line in a single invocation.

TurkeyMan:
If I supply 10 modules on the command line, the compiler emits one single `.o` file with all 10 modules in it (in this case, there would be one `.mixin`).
If I invoke the compiler 10 times for each module, I will naturally emit 10 `.mixin` files before each of the 10 `.o` files...?

thewilsonator:
LDC's `-singleobj` will produce one `.o` for any given number of modules in a single invocation, but that is not the default. E.g. LDC's `-oq` (Preserve source path for output files) exists to prevent collisions on `.o` names. No idea what dmd does.

TurkeyMan:
```
dmd m1.d m2.d m3.d -c -ofmodules.o -debugmixin=modules.mixin

or

dmd m1.d -c -ofm1.o -debugmixin=m1.mixin
dmd m2.d -c -ofm2.o -debugmixin=m2.mixin
dmd m3.d -c -ofm3.o -debugmixin=m3.mixin
```
I can't imagine another scenario that you might be thinking of?

TurkeyMan:
Oh, okay... so the issue here is that LDC has some unusual machinery I'm not aware of?
What's the value of emitting a bunch of object files from a single compiler invocation given many modules?

thewilsonator:
Thats because you're telling it what to call it. `-of` only produces one file. Without that you get `m1.o`, `m2.o`, `.m3.o`.

TurkeyMan:
I don't think DMD does that... (does it?)

thewilsonator:
`$ls .dub/obj`

```
dcompute.driver.backend.o		dcompute.driver.ocl120.kernel.o		dcompute.std.index.o			derelict.opencl.loader.o
dcompute.driver.cuda.buffer.o		dcompute.driver.ocl120.memory.o		dcompute.std.integer.o			derelict.opencl.types.o
dcompute.driver.cuda.context.o		dcompute.driver.ocl120.o		dcompute.std.memory.o			derelict.util.exception.o
dcompute.driver.cuda.device.o		dcompute.driver.ocl120.platform.o	dcompute.std.o				derelict.util.loader.o
...
```

Done with dub with a single invocation.

TurkeyMan:
I can see the problem now. As a simple resolution, I would still be perfectly happy with a single `.mixin` file for the entire invocation... although it might call for a future enhancement request.

Out of curiosity, how is `-H/-Hf` or `-X/-Xf` handled in that context? It sounds like the same problem..

TurkeyMan:
> Done with dub with a single invocation.

I still don't understand the advantage... why is writing/loading/parsing many object files better than 1?

TurkeyMan:
Can the compiler somehow perform dirty checking and not rebuild files that are already clean?
That sounds like a job for the build system, but make/ninja/whatever wouldn't be able to dynamically express this to the rule in the form of the list of and subset of 'dirty' files..

TurkeyMan:
This is the worst place ever to have a conversation... we need to take this offline.

jacob-carlborg:
I'm not sure if/what discussed in Slack but I'll give an answer here.

> Can the compiler somehow perform dirty checking and not rebuild files that are already clean?
That sounds like a job for the build system, but make/ninja/whatever wouldn't be able to dynamically express this to the rule in the form of the list of and subset of 'dirty' files..

The compiler won't do any dirty checking but it does support separate compilation, for example:

```
$ touch foo.d bar.d
$ dmd -c foo.d bar.d
$ ls *.o
bar.o foo.o
```

test/fail_compilation/extra-files/mixin.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+
+diff -pu --strip-trailing-cr ${EXTRA_FILES}/mixin.d.out test_results/fail_compilation/mixin.d.out
```
wilzbach: DO NOT hard-code the test-results folder here!
That's problematic and thus the CIs rightfully fail.
Reading the README often helps (https://github.com/dlang/dmd/tree/master/test#bash-tests), you probably want `${OUTPUT_BASE}` ;-)

test/fail_compilation/mixin.d
```diff
@@ -0,0 +1,24 @@
+// REQUIRED_ARGS:  -mixin=test_results/fail_compilation/mixin.d.out
```
wilzbach: Use a variable here too, e.g. `${RESULTS_DIR}`

thewilsonator:
Thanks!

thewilsonator:
@wilzbach I'm done testing this. I've no idea how to test debug info and don't really want to waste more time on this. This will work for what Manu wants.

jacob-carlborg:
> I've no idea how to test debug info

If there is no better way you can extract that information using an object dumper.

thewilsonator:
Is that a question or a statement?

Its mostly that I'm sick of trying to cooperate with DMD's horrible test suite architecture. I've spent more time trying to get the tests to pass the suite (note, not to be confused with getting the tests to work: they worked first time) than I have developing, debugging and ascertaining user requirements (thanks Manu ;) ).

thewilsonator:
Statements that start with "if" and are a proposition(?) break my parser.

wilzbach:
> Its mostly that I'm sick of trying to cooperate with DMD's horrible test suite architecture. I've spent more time trying to get the tests to pass the suite

Would be interesting to know what's so hard for you and whether sth. could be done to improve this.

thewilsonator:
Principally: partial matching of output,  dishonourable mention: dealing with generated files.

In LDC this is trivial: `FileCheck` has support for regex matches and takes its input on stdin. So the run command would look like 

`//RUN: not %ldc %s -mixin=%t.mixin | FileCheck --check-prefix ERRORMSG && FileCheck --check-prefix MIXIN < %t.mixin`
`not` because the command is expected to fail.
`%s` is the current source file.
 `%t` is a generated temporary which is removed after the test is run (whereas DMD errors out that you haven't removed it yourself, _another_ anti-feature)
`--check-prefix` because we want two modes for the test: that the error message originates from the generated file, and that the generated file contains what we think it does.

The checks would look like:
```
//RUN: not %ldc %s -mixin=%t.mixin | FileCheck --check-prefix ERRORMSG && FileCheck --check-prefix MIXIN < %t.mixin
// ERRORMSG: .mixin:(7)
// MIXIN: expansion at
// MIXIN: int z = x + b;

/* source code here*/
```
Four lines, Done. No extra files, no extra shell script. I suspect the above is why DMD has no asm tests or codegen verification tests. Tbh, I'm not even sure the current test is actually run, I'm a little suspicious that it passed first time.

thewilsonator:
@ZombineDev It would effectively be under whatever licence LLVM uses (which is about to change) which I'm not sure @WalterBright would be particularly happy with, let alone @ibuclaw 

I'll double check with Manu that this works for him, do one more pass through the PR (style etc.) remove the test because its not even being run.  Could someone then please merge it? If I get around to either writing a better test runner or someone can come up with the correct incantations we can add a test case for it. 

Otherwise this PR is dead: I cannot write a test for this.

rainers:
I've pushed a couple of changes to the test.

thewilsonator:
If that works, then I bow to your wizardly test suite powers. There are still a few style issues I need to fix.

Thanks!

test/fail_compilation/mixin.d
```diff
@@ -0,0 +1,26 @@
+// REQUIRED_ARGS: -mixin=test.mixin
+//
+/*
+TEST_OUTPUT:
+---
+test.mixin(7): Error: undefined identifier `b`
```
rainers: I think the original location of the mixin needs to be mentioned in the error, too. Otherwise it's too inconvenient to jump to the source code to fix the error, especially in an IDE where you expect to double click on the error message only.

src/dmd/parse.d
```diff
@@ -218,6 +218,41 @@ private StorageClass getStorageClass(AST)(PrefixAttributes!(AST)* pAttrs)
     return stc;
 }
 
+/**************************************
+ * dump mixin expansion to file for better debugging
+ */
+bool writeMixin(const(char)[] s, ref Loc loc)
+{
+    if (!global.params.mixinOut)
+    return false;
```
rainers: nit: bad indentation.

thewilsonator: I know thats what I meant by style.

test/fail_compilation/mixin.d
thewilsonator: if you jump to the source it will be 'headered' by `expansion from ...` I'm not sure how to inject an error supplemental, given the error call could come from anywhere.

src/dmd/mars.d
```diff
@@ -1566,6 +1582,21 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
...
+                error("-mixin=<file> can only be provided once!");
+                break;
+            }
+            auto tmp = p + 6 + 1;
+            if (!tmp[0])
+                goto Lnoarg;
+            //Don't rely on argv being valid after main has returned, this is used in atexit.
+            params.mixinFile = mem.xstrdup(tmp);
+            params.mixinOut = new OutBuffer;
+            atexit(&flushMixins);
```
rainers: Not too happy about the use of atexit, but I agree that the result might be of interest even in the case of errors. That makes it a bit difficult to place it somewhere in tryMain in mars.d. Maybe add it as a scope(exit) flushMixins() there?

thewilsonator: Does that catch segfautls (e.g. Windows/OSX)?

test/fail_compilation/mixin.d
rainers: It probably needs something like a mixin location stack, similar to template instantiations. Maybe this can be reused.

thewilsonator: Maybe, I'd have to look into it, I'm currently rather ... constrained ... at the moment, will be at least a week.

src/dmd/mars.d
jacob-carlborg: You mean if `flushMixins` when the compiler segfaults? Functions registered using `ataxia` are not called if the application terminates due to a segfault on macOS.

thewilsonator: Well, `scope(exit)` it is then.

test/runnable/mixin.sh
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+src_file=${EXTRA_FILES}/${TEST_NAME}.d
+expect_file=${EXTRA_FILES}/${TEST_NAME}.mixin
+tmp_file=${RESULTS_TEST_DIR}/${TEST_NAME}.mixin
```
wilzbach: You could use `${OUTPUT_BASE}.mixin` too

test/fail_compilation/extra-files/mixin-postscript.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+rm -f "test.mixin"
```
wilzbach: As you noticed this doesn't work.

test/fail_compilation/mixin.d
```diff
@@ -0,0 +1,26 @@
+// REQUIRED_ARGS: -mixin=test.mixin
```
wilzbach: You can use variables here too, e.g. 
https://github.com/dlang/dmd/blob/master/test/compilable/json.d

.circleci/run.sh
```diff
@@ -163,8 +163,9 @@ check_clean_git()
     # Remove temporary directory + install script
     rm -rf _generated
     rm -f install.sh
-    # auto-removal of this file doesn't work on CirleCi
+    # auto-removal of these files doesn't work on CirleCi
     rm -f test/compilable/vcg-ast.d.cg
+    rm -f test/test.mixin
```
wilzbach: This was meant as a safe guard to check that the testsuite didn't modify any checked-in files nor created new ones which aren't in test_results.

wilzbach (commented):
Would be nice to avoid creating temporary files that aren't in test_results

wilzbach (commented):
Would be nice to avoid creating temporary files that aren't in test_results

wilzbach (commented):
Would be nice to avoid creating temporary files that aren't in test_results

wilzbach (commented):
Would be nice to avoid creating temporary files that aren't in test_results

test/fail_compilation/mixin.d
rainers: The trouble is that the file name is also in the tested output. Can this be replaced aswell?

test/fail_compilation/extra-files/mixin-postscript.sh
rainers: It works locally, but not for circleci !?

ZombineDev:
> @ZombineDev It would effectively be under whatever licence LLVM uses

@thewilsonator I didn't mean to port the source code in D, but to implement the features you miss in the existing test suite code.

thewilsonator:
Rebased. Fixed style.

No "mixin stack" yet. While nice, I'm not (yet) convinced that it will be necessary, and that can be done later if it is.

thewilsonator:
All green except semaphore which is segfaulting. @rainers any idea why?

rainers:
> All green except semaphore which is segfaulting. @rainers any idea why?

Not sure, an earlier build reports "double free or corruption".

Maybe it has to with `flushMixin` doing something unexpected when executed by `atexit`.

rainers:
> No "mixin stack" yet. While nice, I'm not (yet) convinced that it will be necessary, and that can be done later if it is.

Ok, being opt-in should not block it for dmd to progress.

I'd love to enable this feature in Visual D by default, but not being able to easily jump to the actual source of an error is probably a deal breaker.

src/dmd/globals.d
```diff
@@ -198,6 +198,10 @@ struct Param
     const(char)* jsonfilename;          // write JSON file to jsonfilename
     JsonFieldFlags jsonFieldFlags;      // JSON field flags to include
 
+    OutBuffer* mixinOut;                // write expanded mixins for debugging
+    const(char)* mixinFile;             // .mixin file output name
+    int mixinLines;                     // Number of lines in writeMixins
+
```
rainers: mixinOut and mixinLines should better go into the Global struct rather than Param. I guess you followed the moduleDeps example, but I'd consider it a bad one.

src/dmd/mars.d
```diff
@@ -1508,6 +1525,21 @@ bool parseCommandLine(const ref Strings arguments, const size_t argc, ref Param
             params.map = true;
         else if (arg == "-multiobj")
             params.multiobj = true;
+        else if (startsWith(p + 1, "mixin="))
+        {
+            if (params.mixinFile)
```
rainers: This follows the -deps example, too, but no other option disallows overriding previous settings.

src/dmd/parse.d
```diff
@@ -242,7 +279,7 @@ final class Parser(AST) : Lexer
         //printf("Parser::Parser()\n");
         scanloc = loc;
 
-        if (loc.filename)
+        if (!writeMixin(input, scanloc) && loc.filename)
```
rainers: Not too happy about `writeMixin` being in the parse module. Maybe it would be better to do this (and the existing pseudo-filename creation) at the callsite with a small helper function.

thewilsonator: I tried that originally, but that constructor is called from `MixinStatement` and `MixinDeclaraion` and I didn't really want to duplicate the code.

thewilsonator:
>Not sure, an earlier build reports "double free or corruption".

Aha. I fix the old problem with `main`'s argv parameter no longer being valid in `atexit`, so I guess `params.mixinOut` being `new`'d means the GC might be getting to it?

src/dmd/mars.d
thewilsonator: Will fix.

thewilsonator:
@rainers any idea why testing of `--DRT-trapExceptions=0` has started failing?
```
#4  0x00000000004276f1 in D main ()
...
grep "in D main (args=...) at src/rt_trap_exceptions_drt.d:9" > /dev/null < ./generated/linux/release/64/rt_trap_exceptions_drt_gdb.output
make[2]: *** [generated/linux/release/64/rt_trap_exceptions_drt_gdb.done] Error 1
```
or why `(args=...)` has disappeared?

thewilsonator:
Unrelated adding `72h -> merge`

test/fail_compilation/extra-files/mixin-postscript.sh
wilzbach: Yes, because one shall not create temporary files in the git repository. CircleCi was checking this and the solution shouldn't have been to hack the check, but placing the file in `TEST_RESULTS`

.circleci/run.sh
wilzbach: https://github.com/dlang/dmd/pull/9236

PR #8685, opened 2018-09-10
[REG master] Partially revert #8623

ibuclaw:
In the following test, the front-end wrongly adds an extra field as of #8623. Which trips the tight asserts present during code gen.
```
// 1684

template Test1684( uint memberOffset ){}

class MyClass1684 {
    int flags2;
    mixin Test1684!(cast(uint)flags2.offsetof) t1; // compiles ok
    mixin Test1684!(cast(int)flags2.offsetof)  t2; // compiles ok
    mixin Test1684!(flags2.offsetof)           t3; // Error: no property 'offsetof' for type 'int'
}
```

### DIFF
--- a/src/dmd/aggregate.d
+++ b/src/dmd/aggregate.d
@@ -169,7 +169,7 @@ extern (C++) abstract class AggregateDeclaration : ScopeDsymbol
     {
         if (_scope)
             dsymbolSemantic(this, null);
-        if (fields.dim > 0)
+        if (sizeok != Sizeok.none)
             return true;
 
         //printf("determineFields() %s, fields.dim = %d\n", toChars(), fields.dim);


WalterBright (dismissed):
Please add the test to the test suite.

ibuclaw:
It's already in the testsuite (template9.d). GDC asserts when compiling it without this.

ibuclaw:
This isn't the first time problems have arisen from changes to this code. What gdc does differently to dmd, is that it walks all `ad.members`, then asserts that the fields added equals `ad.fields`.

The assert typically fails because `ad.fields` was reset to zero.

PetarKirov (approved):

jacob-carlborg (approved):

PR #8686, opened 2018-09-11
cod4.d: fix whitespace

WalterBright:
As always, turning off whitespace diffs when reviewing is useful.

### DIFF
--- a/src/dmd/backend/cod4.d
+++ b/src/dmd/backend/cod4.d
@@ -54,9 +54,9 @@ __gshared const uint[4] dblreg = [ BX,DX,cast(uint)-1,CX ];
 
 private int intree(Symbol *s,elem *e)
 {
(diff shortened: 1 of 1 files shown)


Geod24 (approved):

src/dmd/backend/cod4.d
```diff
@@ -876,40 +885,48 @@ else
     uint jop;
 
 
-  switch (op)                   // select instruction opcodes
-  {     case OPpostinc: op = OPaddass;                  // i++ => +=
+    switch (op)                   // select instruction opcodes
+    {
+        case OPpostinc: op = OPaddass;                  // i++ => +=
```
jacob-carlborg: I would move down the code to below the case statement. Now the code gets quite right adjusted.

PR #8687, opened 2018-09-11
[RFC]  Make the Outbuffer / message function tuple-variadic and slice-aware

Geod24:
This turn the ubiquitous first layer of `message`, `warning`,
`deprecation[Supplemental]`, `error` from C-style variadic into D-style variadic,
while moving the previous implementation as `cmessage`, `cwarning`, etc...

The intent is to make it easy to print a slice using those functions.
At the moment, when one which to print a slice, one can not rely on `\0` termination
of the slice, so the following code:
```
error(loc, "Identifier not found: %s", id.toChars());
```
has to be turned into:
```
error(loc, "Identifier not found: %.*s", cast(int)id.toString().length, id.toString().ptr);
```

This is obviously not convenient and a major burden on the call site.
Worse, some `toString` might perform some computation (e.g. strlen at the moment)
and calling it might not be trivial, which usually leads to storing the slice as an extra variable.

As the goal of our conversion from `char*` to `char[]` is to improve the code quality,
a better solution is obviously needed.
With this patchset, the previous call to `error` becomes:
```
error(loc, "Identifier not found: %.*s", id.toString());
```
As a result, only the format string has to be altered.

### DIFF
--- a/src/dmd/errors.d
+++ b/src/dmd/errors.d
@@ -2,6 +2,19 @@
  * Compiler implementation of the
  * $(LINK2 http://www.dlang.org, D programming language).
  *
+ * This module provides up to 3 kinds of overload to a given function:
+ * - A va_arg accepting function which will be the actual implementation
+ *   This overload is not always present
+ * - A C++ facing implementation, with the prefix which accept C-style varargs.
+ * - An `extern(D)` wrapper to the C++ implementation which accepts
+ *   variadic template arguments.
+ *   Those functions have a 'd' prefix, e.g. `derror`
+ *
+ *   The goal of the wrapper is to turn slices into the length/ptr pair that
+ *   is accepted by printf. This wrapper does not alter the format string,
+ *   and as such the user should call `dmessage("%.*s", my_string)`
+ *   and not `dmessage("%s", my_string)` as this could result in a SIGSEGV
+ *
  * Copyright:   Copyright (C) 1999-2018 by The D Language Foundation, All Rights Reserved
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
@@ -20,6 +33,7 @@ import dmd.globals;
 import dmd.root.outbuffer;
 import dmd.root.rmem;
 import dmd.console;
+import dmd.utils;
 
 /**
  * Color highlighting to classify messages
@@ -33,12 +47,20 @@ enum Classification
 }
 
 /**
- * Print an error message, increasing the global error count.
- * Params:
- *      loc    = location of error
- *      format = printf-style format specification
- *      ...    = printf-style variadic arguments
- */
+Print an error message, increasing the global error count.
+
+Params:
+loc    = location of error
+format = printf-style format specification
+args   = variadic list of arguments
+...    = printf-style variadic arguments
+*/
+void derror(Args...)(const ref Loc loc, const(char)* format, Args args)
+{
+    mixin("error(loc, format, " ~ mapSlices!Args() ~ ");");
+}
+
+/// ditto
 extern (C++) void error(const ref Loc loc, const(char)* format, ...)
 {
     va_list ap;
@@ -69,8 +91,15 @@ extern (D) void error(Loc loc, const(char)* format, ...)
  *      linnum   = line in the source file
  *      charnum  = column number on the line
  *      format   = printf-style format specification
+ *      args   = variadic list of arguments
  *      ...      = printf-style variadic arguments
  */
+void derror(Args...)(const(char)* filename, uint linnum, uint charnum, const(char)* format, Args args)
+{
+    mixin("error(filename, linnum, charnum, format, " ~ mapSlices!Args() ~ ");");
+}
+
+/// ditto
 extern (C++) void error(const(char)* filename, uint linnum, uint charnum, const(char)* format, ...)
 {
     const loc = Loc(filename, linnum, charnum);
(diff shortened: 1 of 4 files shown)


src/dmd/root/outbuffer.d
```diff
@@ -425,6 +430,45 @@ struct OutBuffer
...
+and not the more natural `printf(format, mixin(mapSlices!Args()));`
+
+Params:
+  Args = Type of the arguments to match
+
+Returns:
+  A string describing the parameters to pass to `printf`.
+  The parameters must be a tuple named `args`.
+*/
+string mapSlices(Args...)()
```
jacob-carlborg: Perhaps this is better located in `dmd.utils`.

src/dmd/errors.d
```diff
@@ -69,9 +91,16 @@ extern (D) void error(Loc loc, const(char)* format, ...)
  *      linnum   = line in the source file
  *      charnum  = column number on the line
  *      format   = printf-style format specification
+ *      args   = variadic list of arguments
  *      ...      = printf-style variadic arguments
  */
-extern (C++) void error(const(char)* filename, uint linnum, uint charnum, const(char)* format, ...)
+void error(Args...)(const(char)* filename, uint linnum, uint charnum, const(char)* format, Args args)
+{
+    cerror(filename, linnum, charnum, format, mixin(mapSlices!Args()));
```
Geod24: This is wrong, the mixin should be the whole expression.

src/dmd/errors.d
```diff
@@ -33,13 +47,21 @@ enum Classification
...
+
+Params:
+loc    = location of error
+format = printf-style format specification
+args   = variadic list of arguments
+...    = printf-style variadic arguments
+*/
+void error(Args...)(const ref Loc loc, const(char)* format, Args args)
+{
+    mixin("cerror(loc, format, " ~ mapSlices!Args() ~ ");");
```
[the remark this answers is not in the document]
Geod24: Good catch. Yeah it has to. My original approach was what you saw in the other overload, but it doesn't work (by design)

src/dmd/errors.d
```diff
@@ -2,6 +2,19 @@
...
+ * - A va_arg accepting function which will be the actual implementation
+ *   This overload is not always present
+ * - A C++ facing implementation, with the prefix 'c' (`cerror`, `cwarning`,
+ *   `cmessage`...) which accept C-style varargs.
+ * - An `extern(D)` wrapper to the 'c'-prefixed implementation which accepts
+ *   variadic template arguments.
+ *
+ *   The goal of the wrapper is to turn slices into the length/ptr pair that
+ *   is accepted by printf. This wrapper does not alter the format string,
+ *   and as such the user should call `cmessage("%.*s", my_string)`
```
jacob-carlborg: Should be `message`?

ibuclaw (changes requested):
I'm going to block this until I have time to see what's going on.

I initially see spurious breaking of API, which has alarm bells going off left right and centre.

src/dmd/root/outbuffer.d
Geod24: I was wondering that too, the issue I have is that `outbuffer` is supposed to be close to a leaf module AFAIK (it does import `rmem` and `rootobject`, but those are leaf).
Since utils is a grab-bag of dependency, I'm not so inclined to put it here.
Additionally, this function is very specific to `printf`.
But we could find a new home for it, as it doesn't feel like it belong to `outbuffer` so well either.

Geod24:
Updated following @jacob-carlborg 's comments.

@ibuclaw :
> I initially see spurious breaking of API, which has alarm bells going off left right and centre.

You mean renaming a bunch of symbol to have a `c` prefix ?

ibuclaw:
> You mean renaming a bunch of symbol to have a c prefix ?

I don't use the error module as-is, rather it only has all necessary declarations in order to compile. Then I provide implementations for on the C++ side that hook into the gcc diagnostic system (we have carets, but at some point underlining and fixits would be nice to support too).  This is also the reason why we have the format attributes in the headers.

Any sort of change that makes providing an alternative implementation of diagnostics harder is a worry for me.

Geod24:
> Any sort of change that makes providing an alternative implementation of diagnostics harder is a worry for me.

Then worry not :)
When you'll have time to review it, you'll see that this does nothing but forwarding arguments to the same functions as before (which are now named `c`-something). So my guess is that all you'll have to do is rename your implementation.

By the way this is literally the first line of my PR description:
> This turn the ubiquitous first layer of `message`, `warning`,
> `deprecation[Supplemental]`, `error` from C-style variadic into D-style variadic,
> **while moving the previous implementation as cmessage, cwarning, etc...**

thewilsonator:
@Geod24 It would be nice if the c++ versions could keep the name the same (by e.g. putting them in a struct)
```d
void error(Args...)(const ref Loc loc, const(char)* format, Args args)
 {
     mixin("AvailForCPP.error(loc, format, " ~ mapSlices!Args() ~ ");");
}
private struct AvailForCPP // obvs needs better name
{
    extern (C++) static void cerror(const(char)* filename, uint linnum, uint charnum, const(char)* format, ...)
    {
        //...
    }
   //...
}
```

This would then require no changes on the C++ side.

WalterBright:
I'm having a hard time seeing the value in renaming all the error printing functions. Also, since we are using printf, it's easier to just stick with C strings for that.

Frankly, I think the error message code has gotten out of hand. It used to be pretty simple.

Geod24:
@WalterBright : I went with renaming them in order to automatically switch to the slice-compatible code everywhere. If you prefer, I can just keep the old name and use `dmessage` or similar.

> since we are using printf, it's easier to just stick with C strings for that.

I tried to convert some code to use slices instead of pointer, but keep bumping into issues with the printing functions. There is currently no way to have good-looking printing code and slices at the same time.

> Frankly, I think the error message code has gotten out of hand. It used to be pretty simple.

I'm open to suggestions. I tried to go with the least invasive method, but if you have better ideas I'm all for it.

Geod24:
Updated without a rename to the error functions @ibuclaw @WalterBright

jacob-carlborg:
> Updated without a rename to the error functions

Now we’re going to end up with ugly function names like `derror`. Which I would vote for need to be changed when all C strings have been removed.

WalterBright:
I don't think the added layers of complexity this adds to calling `printf` is worth it. It isn't something users experience, the number of calls to `error()` is finite, the `.*s` notation is not that onerous (I use it frequently). It just makes the compiler harder to understand, and it's hard enough already.

PR #8688, opened 2018-09-11
Implement copy constructor

RazvanN7:
As per DIP: [1].

More tests will be added soon.

[1] https://github.com/dlang/DIPs/pull/129


jacob-carlborg:
Making the `@implicit` keyword a compiler recognizes UDA would avoid some breaking changes.

src/dmd/dsymbolsem.d
```diff
@@ -348,6 +360,115 @@ private extern (C++) FuncDeclaration buildPostBlit(StructDeclaration sd, Scope*
     return xpostblit;
 }
 
+private ModBits createModKey(MOD mod1, MOD mod2)
```
UplinkCoder: No docs.

src/dmd/dstruct.d
```diff
@@ -227,6 +229,9 @@ extern (C++) class StructDeclaration : AggregateDeclaration
     FuncDeclarations postblits; // Array of postblit functions
     FuncDeclaration postblit;   // aggregate postblit
 
+    CopyCtorDeclaration copyCtor;       // copy constructor
+    bool[ModBits] copyCtorTypes;        // source-destination qualifiers for the struct copy constructors
```
RazvanN7: Is there any way to use D's hashtable here and translate it to the C++ headers somehow? Or should I work around it?
Does making the variable private and declare public methods that operate on the hashtable work?

stefan-koch-sociomantic: The bultin-hashtables do not have a ABI gurantee. Therefore avoid them in datastructures part which are part of the AST-Interface.

thewilsonator: Is that piece of data required to be accessed from the glue layer? if not `void*` will do.

src/dmd/expression.d
```diff
@@ -418,13 +418,13 @@ extern (C++) Expression valueNoDtor(Expression e)
  * Input:
  *      sc      just used to specify the scope of created temporary variable
```
WalterBright: Ddoc for destinationType

src/dmd/expression.d
```diff
@@ -448,7 +450,7 @@ private Expression callCpCtor(Scope* sc, Expression e)
 /************************************************
  * Handle the postblit call on lvalue, or the move of rvalue.
```
WalterBright: Ddoc for `t`

src/dmd/expressionsem.d
```diff
@@ -1719,7 +1719,15 @@ private bool functionParameters(const ref Loc loc, Scope* sc,
...
-                if (!tprm.equals(arg.type))
+
+                bool hasCopyCtor;
+                if (arg.type.ty == Tstruct)
+                {
+                    TypeStruct targ = cast(TypeStruct)(arg.type);
+                    if (targ.sym.copyCtor)
+                        hasCopyCtor = true;
+                }
+                if (!hasCopyCtor && !tprm.equals(arg.type))
```
WalterBright: de Morgan this. https://en.wikipedia.org/wiki/De_Morgan%27s_laws

src/dmd/dstruct.d
ibuclaw: See https://github.com/dlang/dmd/blob/aa9411b9b59fcd9746ebf93467f33f50e26933e2/src/dmd/template.h#L65

We already have a precedence for having hash tables in AST.

src/dmd/semantic3.d
```diff
@@ -822,7 +822,17 @@ private extern(C++) final class Semantic3Visitor : Visitor
...
                         }
-                        exp = exp.implicitCastTo(sc2, tret);
+
+                        bool hasCopyCtor;
+                        if (exp.type.ty == Tstruct)
+                        {
+                            TypeStruct ts = cast(TypeStruct)(exp.type);
+                            if (ts.sym.copyCtor)
+                                hasCopyCtor = true;
+                        }
```
WalterBright: This can be written as a one-liner:
```
const bool hasCopyCtor = exp.type.ty == Tstruct &&
                                 (cast(TypeStruct)exp.type).sym.copyCtor !is null;
```
```

File: src/dmd/dsymbolsem.d [hunk 7]
```diff
@@ -86,6 +86,13 @@ private extern (C++) FuncDeclaration buildPostBlit(StructDeclaration sd, Scope*
     if (sd.isUnionDeclaration())
         return null;
 
+    bool hasCopyCtor = sd.copyCtor !is null;
```
WalterBright: `hasCopyCtor` seems pointless. Just `if (sc.copyCtor)` will work fine.

src/dmd/dsymbolsem.d
```diff
@@ -348,6 +360,115 @@ private extern (C++) FuncDeclaration buildPostBlit(StructDeclaration sd, Scope*
     return xpostblit;
 }
 
+private ModBits createModKey(MOD mod1, MOD mod2)
+{
+    return ((mod1 << 8) | mod2);
+}
+
+private CopyCtorDeclaration generateCopyCtorDeclaration(StructDeclaration sd, StorageClass paramStc, StorageClass funcStc)
```
WalterBright: Ddoc comments?

src/dmd/dsymbolsem.d
```diff
@@ -348,6 +360,115 @@ private extern (C++) FuncDeclaration buildPostBlit(StructDeclaration sd, Scope*
...
+
+    // if any field defines a copy constructor
+    if (copyCtorTable.length)
+    {
+        // generate the body that does memberwise initialization
+        auto copyCtorBody = generateCopyCtorBody(sd);
+        foreach (key; copyCtorTable.keys)
+        {
+            MOD paramMod = cast(MOD)(key >> 8);
+            MOD funcMod = cast(MOD)key;
```
WalterBright: These should be `const`.

src/dmd/dsymbolsem.d
```diff
@@ -3628,6 +3761,55 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+            sc = cctd._scope;
+            cctd._scope = null;
+        }
+
+        cctd.parent = sc.parent;
+        Dsymbol p = cctd.toParent2();
+        StructDeclaration sd = p.isStructDeclaration();
+        if (!sd)
+        {
+            error(cctd.loc, "copy constructor can only be a member of aggregate, not %s `%s`", p.kind(), p.toChars());
```
WalterBright: Aggregates include classes, which don't have copy constructors. Error message is erroneous.

src/dmd/dsymbolsem.d
```diff
@@ -3628,6 +3761,55 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
+        // this is needed when generating copy constructors for fields;
+        ModBits key = createModKey(param.type.mod, tf.mod);
+        sd.copyCtorTypes[key] = true;
+
+        // copy constructor parameter type needs to be the same as
+        // the struct containing it (not taking into account qualifiers)
+        Type unqualParamType = param.type.mutableOf().unSharedOf();
+        Type unqualStructType = sd.type.mutableOf().unSharedOf();
+        if (unqualParamType != unqualStructType)
+            error(cctd.loc, "the copy constructor parameter basic type needs to be `%s`, not `%s`", unqualStructType.toChars(), unqualParamType.toChars());
```
WalterBright: Base type, not basic type.

src/dmd/expression.d
```diff
@@ -418,13 +418,13 @@ extern (C++) Expression valueNoDtor(Expression e)
  * Input:
  *      sc      just used to specify the scope of created temporary variable
  */
-private Expression callCpCtor(Scope* sc, Expression e)
+private Expression callCpCtor(Scope* sc, Expression e, Type destinationType)
```
WalterBright: Ddoc for destinationType

src/dmd/expression.d
```diff
@@ -448,7 +450,7 @@ private Expression callCpCtor(Scope* sc, Expression e)
 /************************************************
  * Handle the postblit call on lvalue, or the move of rvalue.
  */
-extern (C++) Expression doCopyOrMove(Scope *sc, Expression e)
+extern (C++) Expression doCopyOrMove(Scope *sc, Expression e, Type t = null)
```
WalterBright: Try to avoid default arguments.

src/dmd/expressionsem.d
```diff
@@ -1719,7 +1719,15 @@ private bool functionParameters(const ref Loc loc, Scope* sc,
...
                     : p.type;
-                if (!tprm.equals(arg.type))
+
+                bool hasCopyCtor;
+                if (arg.type.ty == Tstruct)
+                {
+                    TypeStruct targ = cast(TypeStruct)(arg.type);
+                    if (targ.sym.copyCtor)
+                        hasCopyCtor = true;
+                }
```
WalterBright: Can be turned into a one liner like the other one.

src/dmd/expressionsem.d
```diff
@@ -7734,28 +7742,57 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
-                                exp.error("conversion error from `%s` to `%s`",
-                                    e2x.type.toChars(), e1x.type.toChars());
-                                return setError();
+                                /* Rewrite as:
+                                 * e1 = init, e1.copyCtor(e2);
+                                 */
+                                Expression einit = new BlitExp(exp.loc, exp.e1, getInitExp(sd, exp.loc, sc, t1));
+                                einit.type = e1x.type;
+
+                                if (!einit)
```
WalterBright: Wat? It would have seg faulted before this line is reached if einit was null.

src/dmd/expressionsem.d
```diff
@@ -7734,28 +7742,57 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+                                    return setError();
+
+                                Expression e;
```
WalterBright: Insert a comment showing what expression is being built.

src/dmd/expressionsem.d
```diff
@@ -7734,28 +7742,57 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
...
+                                 */
+                                result = e.expressionSemantic(sc);
+                                if (result.op != TOK.error)
+                                    return;
+                            }
+                            else
+                            {
+                                if (!e2x.type.implicitConvTo(e1x.type))
+                                {
+                                    exp.error("conversion error from `%s` to `%s`",
```
WalterBright: I imagine this error message would be completely baffling to a user.

src/dmd/func.d
```diff
@@ -2748,8 +2748,17 @@ extern (C++) FuncDeclaration resolveFuncCall(const ref Loc loc, Scope* sc, Dsymb
...
 
+            uint errors;
+            bool isCpCtor = fd.isCopyCtorDeclaration() !is null;
+            if (isCpCtor)
+                errors = global.startGagging();
+
             if (fd.checkDisabled(loc, sc))
+            {
+                if (isCpCtor)
+                    global.endGagging(errors);
```
WalterBright: This is awfully hacky. checkDisabled() should accept a parameter that says whether to print the error or not.

WalterBright: In fact, just replace the checkDisabled() call with `fd.storage_class & STC.disable`, as the rest of the function is either irrelevant or entirely concerned with printing the message, and you don't want to print the message. Then, you can get rid of all the gagging crud.

src/dmd/dsymbolsem.d
RazvanN7: It's not. It's used again at line 297

src/dmd/expressionsem.d
RazvanN7: Isn't the one on line 7742 enough?

src/dmd/expressionsem.d
RazvanN7: It was there before. It has nothing to do with this PR

src/dmd/func.d
RazvanN7: It's not just that. If there is no copy constructor suitable for the initialization, trivial copying is tried before issuing an error. This is done to avoid the following scenario : 

```d
struct A
{
    @implicit this(ref shared A rhs) immutable {}
} 

void main()
{
    A a;
    A b = a;     // error : no copy constructor for mutable-mutable copies
}
```

The rule of thumb is : see if there is a valid copy constructor to be called; if yes, call it; if not, try to employ trivial copying.

src/dmd/dstruct.d
RazvanN7: Thanks!

src/dmd/func.d
WalterBright: I don't think that generating a trivial cpctor is the right move if there is an existing one that doesn't match. It should produce an error instead - otherwise the user may erroneously assume it is calling the cpctor he wrote.

andralex: @WalterBright this point is well argued in the DIP. If defining one ctor disables all implicitly-defined ones, the user would need to define up to EIGHT copy constructors - a hecatomb of code. The proposed behavior is to let the user define the exceptional cases and let the default behavior take care of the simple ones.

src/dmd/dsymbolsem.d
iain-buclaw-sociomantic: This looks like a dead variable to me.

RazvanN7: Now it is, but it wasn't

src/dmd/identifier.d
```diff
@@ -95,7 +95,7 @@ nothrow:
         const(char)* p = null;
         if (this == Id.ctor)
             p = "this";
-        else if (this == Id.dtor)
+       else if (this == Id.dtor)
```
thewilsonator: This doesn't need to exist.

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +338,113 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+    foreach (v; sd.fields)
+    {
+        if (v.storage_class & STC.ref_)
+            continue;
+        if (v.overlapped)
+            continue;
+        Type tv = v.type.baseElemOf();
+        if (tv.ty != Tstruct)
+            continue;
+        StructDeclaration sdv = (cast(TypeStruct)tv).sym;
```
WalterBright: Lines 404..408:
```
auto ts = v.type.baseElemOf().isTypeStruct();
if (!ts)
    continue;
if (ts.sym.copyCtor)
```

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +338,113 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+   generated copy constructor is not possible (through implicit conversion
+   or copy construction), the generated copy constructor is marked `@disable`.
+ */
+private CtorDeclaration buildCopyCtor(StructDeclaration sd, Scope* sc)
+{
+    if (sd.postblit)
+        return null;
+
+    Dsymbol s = sd.search(sd.loc, Id.copyCtor);
+    if (s)
```
WalterBright:
```
if (auto s = sd.search(sd.loc, Id.copyCtor))
```

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +338,113 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+        if (tv.ty != Tstruct)
+            continue;
+        StructDeclaration sdv = (cast(TypeStruct)tv).sym;
+        if (auto copyCtor = sdv.copyCtor)
+        {
+            fieldCpCtor = true;
+            break;
+        }
+    }
+
```
WalterBright:
```
if (!fieldCpCtor)
    return null;
```

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +338,113 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+    // if any field defines a copy constructor
+    if (fieldCpCtor)
+    {
+        // generate the body that does memberwise initialization
+        const MOD paramMod = MODFlags.wild;
+        const MOD funcMod = MODFlags.wild;
+        auto ccd = generateCopyCtorDeclaration(sd, ModToStc(paramMod), ModToStc(funcMod));
+        //printf("generating for %s\n", ccd.type.toChars());
+        auto copyCtorBody = generateCopyCtorBody(sd);
+        ccd.fbody = copyCtorBody.syntaxCopy();
```
WalterBright: Why does it need to be copied?

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +338,113 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+        ccd.semantic2(sc2);
+        ccd.semantic3(sc2);
+        //printf("ccd semantic: %s\n", ccd.type.toChars());
+        sc2.pop();
+        if (global.endGagging(errors))
+        {
+            ccd.storage_class |= STC.disable;
+            ccd.fbody = null;
+        }
+        if (!s)
```
WalterBright: `s` is always null here

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +338,113 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+            s = ccd;
```
WalterBright: Just `return ccd;` here.

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +338,113 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+    }
+
+    return s ? s.isCopyCtorDeclaration : null;
```
WalterBright: dead code

src/dmd/dsymbolsem.d
```diff
@@ -3775,6 +3884,23 @@ private extern(C++) final class DsymbolSemanticVisitor : Visitor
...
                                     "but structs cannot have default constructors.");
                 }
+                else if ((dim == 1 || (dim > 1 && tf.parameterList[1].defaultArg)) &&
+                        ctd.ident != Id.copyCtor)
+                {
+                    //printf("tf: %s\n", tf.toChars());
+                    auto param = Parameter.getNth(tf.parameterList, 0);
+                    if (param.storageClass & STC.ref_ && param.type.mutableOf().unSharedOf() == sd.type.mutableOf().unSharedOf())
+                    {
+                        //printf("copy constructor\n");
```
WalterBright: I don't get what is going on here. Somehow, this is going from detecting a copy constructor to generating one? Why is it generating one here?

src/dmd/expressionsem.d
WalterBright: It's nearly 300 lines away, so no.

src/dmd/dsymbol.d
```diff
@@ -1180,6 +1180,11 @@ extern (C++) class Dsymbol : RootObject
         return null;
     }
 
+    inout(CtorDeclaration) isCopyCtorDeclaration() inout
```
WalterBright: This is unlike all the other `isXXXDeclaration()` is copy constructors are not a distinct object. It's a flag in the CtorDeclaration(). So it's confusing. Also, there's only one use of the virtual-ness of this, so this function can be discarded and replaced with isCtorDeclaration() && isCopyCtor.

src/dmd/dsymbolsem.d
RazvanN7: Whenever a constructor is identified as a copy constructor, a copy constructor definition is created. This is done for 2 reasons:

1. The constructor functions need to have a different name then the copy constructors. This makes it easier to resolve overloads and also avoids such situations:

```d
struct A
{
     int* p;
     this(A rhs) {}                          // 1
     this(ref A rhs) immutable {}   // 2
}

void main()
{
     A a;
     A b = a;  // rewritten to  b.__ctor(a) => will resolve to 1 bypassing the cpCtor
}
```

2. extern(C++) on a copy constructor affects the mangling of the constructor, but is ignored for the copy constructor.

src/dmd/dsymbolsem.d
RazvanN7: It doesn't work otherwise in the case of nested structs that define copy constructors.

src/dmd/dsymbolsem.d
WalterBright: 1. The easiest way to deal with this is to make `this(A rhs)` illegal. (It's illegal in C++.)
2. This should be straightforward to handle with the `isCopyCtor` field.

The neat thing is copy constructors won't have to be overload resolved separately from regular constructors.

RazvanN7: >The easiest way to deal with this is to make this(A rhs) illegal. (It's illegal in C++.)

In this case it would have to be all qualified versions:  `this($q1 A rhs) $q2`. This may be a little restrictive, although I can't think of a case where defining such a constructor would be useful. Should this code error even if a copy constructor is not defined?

WalterBright: > In this case it would have to be all qualified versions: this($q1 A rhs) $q2.

Yes, good point.

> This may be a little restrictive, although I can't think of a case where defining such a constructor would be useful.

Neither can I. Such code is most likely a newbie mistake.

> Should this code error even if a copy constructor is not defined?

Yes.

RazvanN7: This breaks phobos [1] and I suspect it will break more code out there.

[1] https://github.com/dlang/phobos/blob/master/std/typecons.d#L8957

ZombineDev: That constructor shouldn't be there. The only reason that I can think of for it's existence would be a workaround for a compiler bug. But regardless, we should try remove it.

WalterBright: Having such constructors results in infinite recursion. The parameter is copied to the stack, which requires the copy constructor, which requires copying the parameter to the stack, which requires the copy constructor, etc.

I understand the concern about breaking code. One possibility is to refuse to allow copy construction in code that has such constructors, similarly to refusal when a postblit is present.

WalterBright: Meanwhile, a PR to fix typecons should be submitted.

RazvanN7: >Having such constructors results in infinite recursion. The parameter is copied to the stack, which requires the copy constructor, which requires copying the parameter to the stack, which requires the copy constructor, etc.

That is not correct. Let's say we have:

```d
struct A
{
     this(A rhs) {}
     this(ref A rhs) {}
}
```

The first functions is a constructor, while the second is also a constructor and a copy constructor. You can imagine that there are 3 functions defined:

```d
struct A
{
     void __ctor(A rhs) {}                    // 1
     void __ctor(ref A rhs) {}               // 2
     void __copyCtor(ref A rhs) {}       // 3
}

ctor 1 will be called explicitly only on rvalues: A(fun()), where fun returns an object of type A. Since fun is in rvalue, the return value will be moved.
ctor 2 will be called excliptly only on lvalues: A(b) where b is a an object of type A. Note that the parameter is passed by reference, so no copy is employed.
ctor 3 will be called implicitly only on lvalues when a copy is performed (documented in the DIP).

So there is no possibility of infinite recursion.

Also, consider the following situation which has nothing to do with the current patch:

```d
import std.stdio : writeln;
struct A
{
    this(A a) { writeln("ctor"); }
}

void main()
{
     A a;
     A b = a;
}

For `A b = a` the compiler does not call the constructor, which means that when an initialization is done by copy the compiler either blits the data (if no postblit is present) or it calls the postblit (if present). Now consider this case:

```d
struct A
{
     int* p;
     this(immutable A rhs) {}
} 

immutable(A) fun() { immutable A a; return a; }

void main()
{
     A b = fun();
}
```

If A'a constructor becomes illegal, then there is no way to initialize an object from an rvalue with mutable indirections.

ZombineDev: When we fix `ref` to allow it to bind rvalues, the copy constructor will be to do its job properly ;)

RazvanN7: I think that the current model of using lvalues with ref is fine; we should let the user bind the rvalue to an lvalue and then call the function with it.

RazvanN7: >That's right, one would need to write a copy constructor to do it. That's the way it should be.
You cannot write a copy constructor for an rvalue. By the current rules, copy constructors work only on lvalues, so there is no copy constructor that you can define to make that work; you either have to explicitly bind the rvalue to an lvalue and then call the copy constructor on it, or you can do I mentioned and simply explicitly call the constructor for rvalues.

>if there is both a this(A) and this(ref A) then that is an error.

But things are not that simple. You may have this(immutable A) and this(ref A). These two constructors are meant for different things: the first is meant for constructing from an rvalue (explicitly), the second is meant for copy constructing (implicitly); it is only a matter of convenience that you can call the copy constructor explcitly as you would do with a normal constructor.  Banning this(immutable A) because you define this(ref A) just doesn't make sense  with the current language rules.

>This is just much more complicated than necessary. Nobody is going to understand the difference between 2 and 3

It really isn't that complicated. The difference between the 2 is that one is called explicilty while the other is implicitly. That's all. You can look at it this way: copy constructors can only be called on rvalues; for everything else you can just define a constructor as you would have done before the DIP. The only matter of confusion here is that we chose to define a copy constructor with the same syntax as a constructor, otherwise everything would have been more clearer.

>All the special cases, the need for naming them differently, the copying of the AST's, special cases with overload resolution, it all falls away.

There are no special cases. You rename them differently to make sure that constructors work exaclty the same as before this addition and you copy the AST to have them as different functions and that's all of it. In addition, the user does not know how it was implemented; for him, when he defines a copy constructor he defines both a constructor (which he can call explicitly) and a copy constructor (which is always called implicitly). I don'y see any reason of confusion here. If we want to make things clear, maybe we should just change the syntax and have copy constructors be a completely different thing from constructors.

WalterBright (approved):

wilzbach (dismissed):
Please add more tests as experience shows that every new feature comes with new uncovered edge cases.

src/dmd/astbase.d
```diff
@@ -824,7 +824,7 @@ struct ASTBase
 
     extern (C++) final class CtorDeclaration : FuncDeclaration
     {
-        extern (D) this(const ref Loc loc, Loc endloc, StorageClass stc, Type type)
+        extern (D) this(const ref Loc loc, Loc endloc, StorageClass stc, Type type, bool isCopyCtor = false)
```
jacob-carlborg: Parameters with default values are usually not allowed in DMD.

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +336,153 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
     return xpostblit;
 }
 
+/* Generates a copy constructor declaration with the specified storage
+   class for the parameter and the function : this(ref paramSTC S p) funcStc;
+*/
```
jacob-carlborg: All parameters and return value should be documented.

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +336,153 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+}
+
+/* Generates a trivial copy constructor body that simply does memberwise
+ * initialization:
+ *
+ *    this.field1 = rhs.field1;
+ *    this.field2 = rhs.field2;
+ *    ...
+ */
+private Statement generateCopyCtorBody(StructDeclaration sd)
```
jacob-carlborg: Same here.

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +336,153 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+        e = Expression.combine(e, ec);
+        //printf("e.toChars = %s\n", e.toChars());
+    }
+    Statement s1 = new ExpStatement(loc, e);
+    return new CompoundStatement(loc, s1);
+}
+
+/* Generates copy constructor for the fields that define copy constructors.
+ */
+private bool buildCopyCtor(StructDeclaration sd, Scope* sc)
```
jacob-carlborg: And here.

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +336,153 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+            {
+                rvalueCtor = ctorDecl;
+            }
+        }
+        return 0;
+    });
+
+    if (cpCtor && rvalueCtor)
+    {
+        .error(sd.loc, "`struct %s` may not define both a rvalue constructor (at line `%d`) and a copy constructor (at line `%d`)",
```
jacob-carlborg: I suggest splitting the error message in two, an error message and a supplemental message.

src/dmd/dsymbolsem.d
```diff
@@ -336,6 +336,153 @@ private FuncDeclaration buildPostBlit(StructDeclaration sd, Scope* sc)
...
+            continue;
+        if (ts.sym.hasCopyCtor)
+        {
+            fieldWithCpCtor = v;
+            break;
+        }
+    }
+
+    if (fieldWithCpCtor && rvalueCtor)
+    {
```
jacob-carlborg: Same here, split the error message.

src/dmd/expressionsem.d
```diff
@@ -1904,7 +1904,9 @@ private bool functionParameters(const ref Loc loc, Scope* sc,
                 Type tprm = p.type.hasWild()
                     ? p.type.substWildTo(wildmatch)
                     : p.type;
-                if (!tprm.equals(arg.type))
+
+                bool hasCopyCtor = (arg.type.ty == Tstruct) && (cast(TypeStruct)arg.type).sym.hasCopyCtor;
```
jacob-carlborg: `const`.

src/dmd/mtype.d
```diff
@@ -4669,7 +4670,51 @@ extern (C++) final class TypeFunction : TypeNext
                         m = targ.implicitConvTo(tprm);
                     }
                     else
-                        m = arg.implicitConvTo(tprm);
+                    {
+                        bool isRef = (p.storageClass & (STC.ref_ | STC.out_)) != 0;
```
jacob-carlborg: `const`.

jacob-carlborg (changes requested):
In addition to the inline comments. The implementation should not be accepted until the DIP has gone through the **whole** process and been accepted.

wilzbach:
> The implementation should not be accepted until the DIP has gone through the whole process and been accepted.

Though maybe we can use the new `-preview` flags for this and merge early, but still can adapt and improve it before exposing to the general public and being locked in for eternity.

WalterBright:
> Though maybe we can use the new `-preview` flags for this and merge early, but still can adapt and improve it before exposing to the general public and being locked in for eternity.

Works for me!

RazvanN7:
Spec change : https://github.com/dlang/dlang.org/pull/2599

jacob-carlborg (dismissed):
Please don’t just link to the DIP. I expect a bit more elaborate changelog.

RazvanN7:
@jacob-carlborg I fixed the changelog and the inline comments.

@wilzbach The only new thing added here is that the compiler inserts calls to the copy constructor where it used to insert for the postblit. This is tested. Moreover, there is a reference counted array implementation that uses the copy constructor and passes all the tests locally [1].

[1] https://github.com/dlang/druntime/pull/2348

jacob-carlborg:
Changelog looks good, almost a bit too long.

andralex:
Failures seem unrelated, I'll pull the plug on this.

src/dmd/dsymbolsem.d
atilaneves: I missed this discussion originally. The reason one one want to have:

```d
struct Foo {
    this(ref Foo);
    this(Foo);
}
```

is simple: C++ interop. I made it so that the dpp translations actually enabled D code to call a C++ move constructor by overloading on this.

MoonlightSentinel:
This introduced a regression: https://issues.dlang.org/show_bug.cgi?id=22997
`git bisect` blames commit db885735557357c6a2886d6384459cbe3ae66056

PR #8689, opened 2018-09-12
convert cgcod.c to D

WalterBright:


WalterBright:
buildkite claims it fails, but all the logs are green.

andralex (approved):

PR #8690, opened 2018-09-13
Remove more char* and use more slices

Geod24:
Best reviewed commit-by-commit.

### DIFF
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -71,30 +71,29 @@ version (CRuntime_Microsoft)
 /****************************************
  * Write filename to cmdbuf, quoting if necessary.
  */
-private void writeFilename(OutBuffer* buf, const(char)* filename, size_t len)
+private void writeFilename(OutBuffer* buf, const(char)[] filename)
 {
     /* Loop and see if we need to quote
      */
-    for (size_t i = 0; i < len; i++)
+    foreach (const char c; filename)
     {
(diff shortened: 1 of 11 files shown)


WalterBright (approved):

src/dmd/link.d
```diff
@@ -71,30 +71,29 @@ version (CRuntime_Microsoft)
...
+    foreach (const char c; filename)
     {
-        const char c = filename[i];
         if (isalnum(c) || c == '_')
             continue;
         /* Need to quote
          */
         buf.writeByte('"');
-        buf.write(filename, len);
+        buf.write(filename.ptr, filename.length);
```
jacob-carlborg: I suggest using `writestring` instead, which takes a `const(char)[]`.

src/dmd/link.d
```diff
@@ -71,30 +71,29 @@ version (CRuntime_Microsoft)
...
         buf.writeByte('"');
         return;
     }
     /* No quoting necessary
      */
-    buf.write(filename, len);
+    buf.write(filename.ptr, filename.length);
```
jacob-carlborg: Same here, `writestring`.

jacob-carlborg (approved):

Geod24:
Fixed @jacob-carlborg 's comment and the test failure (was missing `ex = tmp;` in link.d)

PR #8692, opened 2018-09-14
Semaphore: Use new URL for github/raw install.sh mirror

Geod24:
Running the comment on the previous URL leads to:
```
<html><body>You are being <a href="https://raw.githubusercontent.com/dlang/installer/master/script/install.sh">redirected</a>.</body></html>
```


andralex:
cc @MartinNowak

andralex:
cc @CyberShadow

CyberShadow:
Thanks. I believe @wilzbach would be the person in charge of this file.

PR #8696, opened 2018-09-14
Fix 19243 - [REG 2.081] Can no longer override pragma(lib) with -L switch

rtbo:
See https://issues.dlang.org/show_bug.cgi?id=19243
What I do here is to keep *.a files at the start of linker command line to keep #8172 on working, but all other switches are back to their previous locations.

### DIFF
--- a/src/dmd/link.d
+++ b/src/dmd/link.d
@@ -569,38 +569,21 @@ public int runLINK()
             argv.push("-Xlinker");
             argv.push("--gc-sections");
         }
-        /* Add each library, prefixing it with "-l".
-         * The order of libraries passed is:
-         *  1. any libraries passed with -L command line switch
-         *  2. libraries specified on the command line
+        /* Add libraries. The order of libraries passed is:
+         *  1. static libraries ending with *.a     (global.params.libfiles)
+         *  2. link switches passed with -L command line switch  (global.params.linkswitches)
          *  3. libraries specified by pragma(lib), which were appended
-         *     to global.params.libfiles.
-         *  4. link switches, that may also contain -l libraries
+         *     to global.params.libfiles. These are prefixed with "-l"
+         *  4. dynamic libraries passed to the command line (global.params.dllfiles)
          *  5. standard libraries.
          */
-        for (size_t i = 0; i < global.params.libfiles.dim; i++)
+        foreach (p; global.params.libfiles)
         {
-            const(char)* p = global.params.libfiles[i];
-            size_t plen = strlen(p);
-            if (plen > 2 && p[plen - 2] == '.' && p[plen - 1] == 'a')
+            if (FileName.equalsExt(p, "a"))
                 argv.push(p);
-            else
-            {
-                char* s = cast(char*)mem.xmalloc(plen + 3);
-                s[0] = '-';
-                s[1] = 'l';
-                memcpy(s + 2, p, plen + 1);
-                argv.push(s);
-            }
         }
-        for (size_t i = 0; i < global.params.dllfiles.dim; i++)
+        foreach (p; global.params.linkswitches)
         {
-            const(char)* p = global.params.dllfiles[i];
-            argv.push(p);
-        }
-        for (size_t i = 0; i < global.params.linkswitches.dim; i++)
-        {
-            const(char)* p = global.params.linkswitches[i];
             if (!p || !p[0] || !(p[0] == '-' && (p[1] == 'l' || p[1] == 'L')))
             {
                 // Don't need -Xlinker if switch starts with -l or -L.
@@ -608,6 +591,22 @@ public int runLINK()
                 // to take precedence over gcc defaults.
                 argv.push("-Xlinker");
             }
+            argv.push(p);
+        }
+        foreach (p; global.params.libfiles)
+        {
+            if (!FileName.equalsExt(p, "a"))
+            {
+                const plen = strlen(p);
+                char* s = cast(char*)mem.xmalloc(plen + 3);
+                s[0] = '-';
+                s[1] = 'l';
+                memcpy(s + 2, p, plen + 1);
+                argv.push(s);
+            }
+        }
+        foreach (p; global.params.dllfiles)
+        {
             argv.push(p);
         }
         /* D runtime libraries must go after user specified libraries
--- a/test/compilable/issue19243.sh
+++ b/test/compilable/issue19243.sh
@@ -1,0 +1,50 @@
+#! /usr/bin/env bash
+
+# bypassing this test:
+#  - on windows
+#  - on FreeBSD 32 bits
+#       test fails looling for liborig.so; don't know why but shouldn't block fixing all other platforms
+#  - on Circle CI with no_pic. (need PIC to run the test)
+if [[ $OS = *"win"* ]]; then exit 0; fi
+if [[ $OS = *"freebsd"* ]] && [[ $MODEL = *"32"* ]]; then exit 0; fi
+if [ ${PIC:-1} == "0" ]; then exit 0; fi
+
+TEST_DIR=${OUTPUT_BASE}
+ORIG_D=$TEST_DIR/orig.d
+ORIG_SO=$TEST_DIR/liborig${SOEXT}
+OVERRIDE_D=$TEST_DIR/override.d
+OVERRIDE_SO=$TEST_DIR/liboverride${SOEXT}
+APP_D=$TEST_DIR/app.d
+
+mkdir -p $TEST_DIR
+
+cat << EOF | $DMD -m$MODEL -fPIC -shared -of$ORIG_SO -
+import core.stdc.stdio;
+
+extern(C) int func()
+{
+    printf("liborig\n");
+    return 1;
+}
+EOF
+
+cat << EOF | $DMD -m$MODEL -fPIC -shared -of$OVERRIDE_SO -
+import core.stdc.stdio;
+
+extern(C) int func()
+{
+    printf("liboverride\n");
+    return 2;
+}
+EOF
+
+cat << EOF | LD_LIBRARY_PATH=$TEST_DIR $DMD -m$MODEL -L-L$TEST_DIR -L$OVERRIDE_SO -run -
+extern(C) int func();
+
+pragma(lib, "orig");
+
+void main()
+{
+    assert(func() == 2);
+}
+EOF
--- a/test/README.md
+++ b/test/README.md
@@ -215,3 +215,6 @@ Along with the environment variables provided by the Makefile (see above), an ad
     EXTRA_FILES        directory for extra files of this test type, e.g. runnable/extra-files
 
     LIBEXT             platform-specific extension for library files, e.g. .a or .lib
+
+    SOEXT              platform-specific extension for shared object files (aka. dynamic libraries),
+                       e.g. .so, .dll or .dylib
--- a/test/tools/exported_vars.sh
+++ b/test/tools/exported_vars.sh
@@ -13,6 +13,14 @@ else
     export LIBEXT=.a
 fi
 
+if [[ "$OS" == "win"* ]]; then
+    export SOEXT=.dll
+elif [[ "$OS" = "osx" ]]; then
+    export SOEXT=.dylib
+else
+    export SOEXT=.so
+fi
+
 # Default to DigitalMars C++ on Win32
 if [ "$OS" == "win32" ] && [ -z "${CC+set}" ] ; then
     CC="dmc"


rtbo:
Oups. should rebase to stable.

src/dmd/link.d
```diff
@@ -569,34 +569,21 @@ public int runLINK()
...
-         *  4. link switches, that may also contain -l libraries
-         *  5. standard libraries.
+         *  4. standard libraries.
          */
         for (size_t i = 0; i < global.params.libfiles.dim; i++)
         {
             const(char)* p = global.params.libfiles[i];
             size_t plen = strlen(p);
-            if (plen > 2 && p[plen - 2] == '.' && p[plen - 1] == 'a')
+            const bool is_archive = (plen > 2 && p[plen - 2] == '.' && p[plen - 1] == 'a');
```
jacob-carlborg: Please use the standard D naming conventions, i.e. `isArchive`.

src/dmd/link.d
```diff
@@ -610,6 +597,24 @@ public int runLINK()
             }
             argv.push(p);
         }
+        for (size_t i = 0; i < global.params.libfiles.dim; i++)
+        {
+            const(char)* p = global.params.libfiles[i];
+            size_t plen = strlen(p);
+            const bool is_archive = (plen > 2 && p[plen - 2] == '.' && p[plen - 1] == 'a');
```
jacob-carlborg: Same here, `isArchive`.

src/dmd/link.d
```diff
@@ -610,6 +597,24 @@ public int runLINK()
...
+            if (!is_archive) {
```
jacob-carlborg: Curly braces on their own lines.

src/dmd/link.d
```diff
@@ -610,6 +597,24 @@ public int runLINK()
             }
             argv.push(p);
         }
+        for (size_t i = 0; i < global.params.libfiles.dim; i++)
```
jacob-carlborg: Perhaps this could be a `foreach` loop.

jacob-carlborg (changes requested):

src/dmd/link.d
```diff
@@ -569,36 +569,23 @@ public int runLINK()
...
+         *  1. static libraries ending with *.a
+         *  2. link switches passed with -L command line switch
          *  3. libraries specified by pragma(lib), which were appended
          *     to global.params.libfiles.
-         *  4. link switches, that may also contain -l libraries
-         *  5. standard libraries.
+         *  4. standard libraries.
          */
-        for (size_t i = 0; i < global.params.libfiles.dim; i++)
+        foreach (i; 0 .. global.params.libfiles.dim)
```
Geod24: If you are going the foreach way;
```D
foreach (p; global.params.libfiles)
```

src/dmd/link.d
```diff
@@ -569,36 +569,23 @@ public int runLINK()
...
         {
             const(char)* p = global.params.libfiles[i];
-            size_t plen = strlen(p);
-            if (plen > 2 && p[plen - 2] == '.' && p[plen - 1] == 'a')
+            const plen = strlen(p);
+            const isArchive = (plen > 2 && p[plen - 2] == '.' && p[plen - 1] == 'a');
```
Geod24: `dmd.root.filename` has utilities to deal with filenames.
From the top of my head, `const isArchive = FileName.ext(p) == "a";` might be way more readable.

src/dmd/link.d
```diff
@@ -569,36 +569,23 @@ public int runLINK()
...
-                argv.push(s);
-            }
-        }
-        for (size_t i = 0; i < global.params.dllfiles.dim; i++)
-        {
-            const(char)* p = global.params.dllfiles[i];
-            argv.push(p);
         }
-        for (size_t i = 0; i < global.params.linkswitches.dim; i++)
+        foreach (i; 0 .. global.params.linkswitches.dim)
```
Geod24: If you convert it to a `foreach`, make use the syntax previously mentioned

src/dmd/link.d
rtbo: Didn't know it was possible with Array.
Still don't see what makes it legal. Is it opSlice? This isn't documented.

rtbo:
I can't get the test to work. 
I need the linker to find a lib (any lib in fact) specified with `pragma(lib)`.
Original test was with `ssl`, which can't be found in the test environment.
I tried to have the linker find `libsub.so` that I created in the test dir, but no success. (although it works on my machine)
I can't find any lib that would be common in all testing environments. Any ideas?

CyberShadow:
The following simple test script works for me in 2.080 and fails in 2.081:

```bash
#!/bin/bash
set -eu

cat > liborig.d <<'EOF'
import core.stdc.stdio;

extern(C) int func()
{
    printf("liborig\n");
    return 1;
}
EOF
dmd -fPIC -shared liborig.d

cat > liboverride.d <<'EOF'
import core.stdc.stdio;

extern(C) int func()
{
    printf("liboverride\n");
    return 2;
}
EOF
dmd -fPIC -shared liboverride.d

cat > prog.d <<'EOF'
extern(C) int func();

pragma(lib, "orig");

void main()
{
    assert(func() == 2);
}
EOF

LD_LIBRARY_PATH=. dmd -L-L. -L./liboverride.so -run prog.d
```

Hope this helps.

jacob-carlborg:

src/dmd/link.d
```diff
@@ -569,38 +569,21 @@ public int runLINK()
...
          * The order of libraries passed is:
-         *  1. any libraries passed with -L command line switch
-         *  2. libraries specified on the command line
+         *  1. static libraries ending with *.a
+         *  2. link switches passed with -L command line switch
          *  3. libraries specified by pragma(lib), which were appended
          *     to global.params.libfiles.
-         *  4. link switches, that may also contain -l libraries
-         *  5. standard libraries.
+         *  4. standard libraries.
```
jacob-carlborg: How does this PR affect dynamic libraries passed directly on the command line without a flag?

src/dmd/link.d
```diff
@@ -569,38 +569,21 @@ public int runLINK()
...
+        /* Add each library, prefixing it with "-l" if necessary.
          * The order of libraries passed is:
-         *  1. any libraries passed with -L command line switch
-         *  2. libraries specified on the command line
+         *  1. static libraries ending with *.a
+         *  2. link switches passed with -L command line switch
          *  3. libraries specified by pragma(lib), which were appended
          *     to global.params.libfiles.
-         *  4. link switches, that may also contain -l libraries
-         *  5. standard libraries.
```
jacob-carlborg: Assuming this, now changed, documentation was correct (actually matching the implementation) won't this PR introduce some other breaking changes?

src/dmd/link.d
rtbo: Not affecting at all.
Arguments not starting with `-L` and ending with `.so` are in `global.params.dllfiles`.
They are passed to the linker right before std lib link flags as before this PR and before the PR that introduced this regression.

src/dmd/link.d
rtbo: Not sure this doc was correct (will make this clearer). I am not a `ld` guru, but I don't think other breaking change be introduced.
Actual order has evolved in the following way
before #8172:
```d
        /*  1. link switches passed with -L command line switch (global.params.linkswitches)
         *  2. static libraries (ending with *.a)  (global.params.libfiles)
         *  3. libraries specified by pragma(lib), prefixed with "-l" (global.params.libfiles)
         *  4. dynamic libraries passed to the command line (global.params.dllfiles)
         *  5. standard libraries.
         */
```
after #8172:
```d
        /*  1. static libraries (ending with *.a)  (global.params.libfiles)
         *  2. libraries specified by pragma(lib), prefixed with "-l" (global.params.libfiles)
         *  3. link switches passed with -L command line switch (global.params.linkswitches)
         *  4. dynamic libraries passed to the command line (global.params.dllfiles)
         *  5. standard libraries.
         */
```
With this PR:
```d
        /*  1. static libraries (ending with *.a)  (global.params.libfiles)
         *  2. link switches passed with -L command line switch (global.params.linkswitches)
         *  3. libraries specified by pragma(lib), prefixed with "-l" (global.params.libfiles)
         *  4. dynamic libraries passed to the command line (global.params.dllfiles)
         *  5. standard libraries.
         */
```

src/dmd/link.d
rtbo: Sorry that's not right. I double checked. see next comment with argument order evolution

test/compilable/issue19243.sh
```diff
@@ -0,0 +1,56 @@
...
+#  - on Circle CI with no_pic. (need PIC to run the test)
+if [[ $OS = *"win"* ]]; then exit 0; fi
+if [[ $OS = *"freebsd"* ]] && [[ $MODEL = *"32"* ]]; then exit 0; fi
+if [ ! -z ${CIRCLE_STAGE-} ]; then
+    if [[ ${CIRCLE_STAGE} = *"no_pic"* ]]; then exit 0; fi
+fi
+
+TEST_DIR=${OUTPUT_BASE}
+ORIG_D=$TEST_DIR/orig.d
+ORIG_SO=$TEST_DIR/liborig.so
```
jacob-carlborg: Dynamic libraries on macOS has the `.dylib` extension. Probably doesn't really matter.

jacob-carlborg: BTW, dynamic libraries are not officially supported on macOS.

rtbo: Should I bypass on OSX too then?
Test might break later

CyberShadow:
Perhaps the existing tests in the test suite dealing with shared libraries can be used as a base for this test? Issues with `-fPIC` etc. were probably encountered and solved before.

test/compilable/issue19243.sh
jacob-carlborg: Well, if it works it works. That’s why it’s not officially supported because some things work and some won’t.

rtbo:
`*.d` tests can specify 
```d
/// REQUIRED_ARGS: -fPIC
```
That's all I can see about `-fPIC` in the test dir.

rtbo:
OK, unless further remark, I'm all set.

Wrap-up of the change of linker arguments order:
before #8172:
```d
        /*  1. link switches passed with -L command line switch (global.params.linkswitches)
         *  2. static libraries (ending with *.a)  (global.params.libfiles)
         *  3. libraries specified by pragma(lib), prefixed with "-l" (global.params.libfiles)
         *  4. dynamic libraries passed to the command line (global.params.dllfiles)
         *  5. standard libraries.
         */
```
after #8172:
```d
        /*  1. static libraries (ending with *.a)  (global.params.libfiles)
         *  2. libraries specified by pragma(lib), prefixed with "-l" (global.params.libfiles)
         *  3. dynamic libraries passed to the command line (global.params.dllfiles)
         *  4. link switches passed with -L command line switch (global.params.linkswitches)
         *  5. standard libraries.
         */
```
With this PR:
```d
        /*  1. static libraries (ending with *.a)  (global.params.libfiles)
         *  2. link switches passed with -L command line switch (global.params.linkswitches)
         *  3. libraries specified by pragma(lib), prefixed with "-l" (global.params.libfiles)
         *  4. dynamic libraries passed to the command line (global.params.dllfiles)
         *  5. standard libraries.
         */
```

jacob-carlborg (approved):

test/compilable/issue19243.sh
```diff
@@ -0,0 +1,63 @@
...
+
+# bypassing this test:
+#  - on windows
+#  - on FreeBSD 32 bits
+#       test fails looling for liborig.so; don't know why but shouldn't block fixing all other platforms
+#  - on Circle CI with no_pic. (need PIC to run the test)
+if [[ $OS = *"win"* ]]; then exit 0; fi
+if [[ $OS = *"freebsd"* ]] && [[ $MODEL = *"32"* ]]; then exit 0; fi
+if [ ! -z ${PIC-} ]; then
+    if [ ${PIC} -eq "0" ]; then
```
wilzbach: Simplify to one if condition: `"${PIC:-1}" == "0"`

test/compilable/issue19243.sh
```diff
@@ -0,0 +1,63 @@
...
+        exit 0
+    fi
+fi
+
+SOEXT=.so
+if [[ $OS = "osx" ]]; then
+    SOEXT=.dylib
+fi
```
wilzbach: Might make sense to add this to the globally available environment variables (https://github.com/dlang/dmd/blob/master/test/tools/exported_vars.sh).
Documentation is at the end of test/README.md

test/compilable/issue19243.sh
```diff
@@ -0,0 +1,63 @@
...
+
+cat > $ORIG_D <<'EOF'
+import core.stdc.stdio;
+
+extern(C) int func()
+{
+    printf("liborig\n");
+    return 1;
+}
+EOF
```
wilzbach: No need to write it to a file.

You can use " | dmd -"

RazvanN7 (approved):

PR #4972, opened 2015-08-28
[WIP] Add text serialization of IR

yebblies:
Generate a nice(ish) text version of the backend's IR.
The idea is that we will be able to write test cases in this format, then check the post-optimizer trees to ensure that the optimizer re-writes certain expressions, preserves others, etc.

Can also be used to test D -> IR conversions, IR -> machine code, etc

This will allow us to fork off from dmc without losing its test suite, by converting the C/C++ tests to IR tests that will continue to work with dmd.

Example output:

``` d
int g;

void main()
{
    if (g)
    {
        asm { nop; }
    }
    else
    {
        assert(0);
    }
}
```

```
DMD v2.069 DEBUG
function: _Dmain
block: 0
    exp: (ind int (add nptr (relconst nptr _D5testx1gi) (ind uint (var nptr _tls_array))))
    bc: iftrue
block: 1
    bc: goto
block: 2
    bc: asm
block: 3
    bc: goto
block: 4
    exp: (oror void (const int 0) (call void (var jfunc _D5testx8__assertFiZv) (const int 12)))
    bc: goto
block: 5
    exp: (const int 0)
    bc: retexp
block: 6
    bc: ret
function: main
block: 0
    exp: (call int (var nfunc _d_run_main) (param void (add nptr (relconst nptr _Dmain) (const uint 0)) (param void (var nptr argv) (var int argc))))
    bc: retexp
block: 1
    bc: ret
```

PR #8699, opened 2018-09-15
optimization: recognize empty trailing statements

WalterBright:
A small one, but anything we can do to eliminate unnecessary try-catches is a win.

### DIFF
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -250,7 +250,24 @@ private extern (C++) final class StatementSemanticVisitor : Visitor
                         sexception = sexception.statementSemantic(sc);
                     if (sexception)
                     {
-                        if (i + 1 == cs.statements.dim && !sfinally)
+                        /* Returns: true if statements[] are empty statements
+                         */
+                        static bool isEmpty(const Statement[] statements)
+                        {
+                            foreach (s; statements)
+                            {
+                                if (const cs = s.isCompoundStatement())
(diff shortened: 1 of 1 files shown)


Geod24:
Can you give some code sample in which this would kick in ? Also, any chance you could provide a test for it ?

WalterBright:
The trigger for it is https://github.com/dlang/dmd/pull/8697/files#diff-53d6b4d266fad20ec997b3209808d37dR1283 where fbody is empty. I don't know of any other triggers for it at the moment. As with most optimizations, it's a little hard to write a test case for it.

WalterBright:
Testing optimization transformations is an unresolved problem in the compiler.

wilzbach:
> Testing optimization transformations is an unresolved problem in the compiler.

While it would be amazing to have an IR, there are test for the output assembly already, e.g.

https://github.com/dlang/dmd/blob/master/test/runnable/test_cdvecfill.d

wilzbach:
> While it would be amazing to have an IR,

see also: https://github.com/dlang/dmd/pull/4972

wilzbach:
Though the existing `-vcg-ast` switch might help too, but it currently will always create a separate file and there's no way to force it to output this directly on stdout (which would be nice as then the `TEST_OUTPUT` section could be used directly and no custom `POST_SCRIPT` would be required.)

RazvanN7 (approved):

PR #8704, opened 2018-09-16
Link: Remove custom removeExt, use FileName instead

Geod24:


jacob-carlborg (approved):

WalterBright (approved):

adamdruppe:
This introduced a regression, it uses `ex` before initialization (the old code with memcpy actually initialized it to `n`, but zero terminated at the dot).

This fix is simple though: https://github.com/opendlang/opend/commit/3d498fa82a32cb4dff88691aef09d74c204bb2ce

PR #8707, opened 2018-09-17
Enable putting structs with void field initializers into BSS segment

WalterBright:
Initializers can go into BSS if they are all zero, where the initialization happens essentially for free. For this purpose, treat void field initializations as zero. Also works for using a memset to initialize a struct instance to all zeros rather than doing a memcpy from the .init data.

### DIFF
--- a/src/dmd/dstruct.d
+++ b/src/dmd/dstruct.d
@@ -401,6 +401,12 @@ extern (C++) class StructDeclaration : AggregateDeclaration
         {
             if (vd._init)
             {
+                if (vd._init.isVoidInitializer())
+                    /* Treat as 0 for the purposes of putting the initializer
+                     * in the BSS segment, or doing a mass set to 0
+                     */
+                    continue;
+
                 // Zero size fields are zero initialized
                 if (vd.type.size(vd.loc) == 0)
                     continue;


Geod24:
Red because this affects `union` as well

WalterBright:
> Red because this affects union as well

I fixed that.

jacob-carlborg:
Can we test this? A shell script compile some code and using an object dumper on the object file?

wilzbach:
Added such a simple shell test (Linux only), but it was working with master already :/

ZombineDev:
@wilbach you want to test `__gshared` global variables.

test/compilable/test8707.sh
```diff
@@ -0,0 +1,15 @@
+if [ "${OS}" != "linux" ]; then
+    echo "Skipping void initialization test on $OS."
+fi
```
rainers: missing `exit 0` here?

wilzbach: No need - that's the default.

rainers: You are not skipping the rest of the code here.

n8sh:
>Can we test this? 

A straightforward test that works on platforms can be added to [test/compilable/isZeroInit.d](https://github.com/dlang/dmd/blob/11573d214097d3fcb7eeb388eba0f6ddf8335694/test/compilable/isZeroInit.d):
```d
struct VoidFieldInitStruct
{
    int i = void;
}
static assert(__traits(isZeroInit, VoidFieldInitStruct));
```

WalterBright:
Adding the test causes the PR to fail.

WalterBright:
I removed test8707.sh, as in looking at the object file the .init was placed in BSS, so something was wrong with it.

test/compilable/extra-files/test8707.d
```diff
@@ -0,0 +1,9 @@
+struct Foo
```
rainers: This test doesn't do a lot now, could be
```
module test;

struct Foo
{
    int i = void;
}
static assert(__traits(isZeroInit, Foo));

void main()
{
    Foo f;
    version(Win32) 
        assert(typeid(Foo).initializer.ptr >= cast(void*)&_edata && 
               typeid(Foo).initializer.ptr < cast(void*)&_edata);
}

version(Win32)
{
    extern(C) extern __gshared
    {
        int _edata;     // &_edata is start of BSS segment
        int _end;       // &_end is past end of BSS
    }
}
```

rainers: Scratch that, a zero-initializer is not even placed into BSS! This would be a runtime check:
```
module test;

struct Foo
{
    int i = void;
}
static assert(__traits(isZeroInit, Foo));

void main()
{
    Foo f = Foo.init;
    assert(typeid(Foo).initializer.ptr == null);
}
```

rainers: > a zero-initializer is not even placed into BSS!

Wrong again ;-/ `Foo f;` uses "memset(0)", while `Foo f = Foo.init;` copies the actual init from DATA/BSS if the struct is larger than 32 bytes (even for optimized builds).

It is not allowed to get the address of `Foo.init`. Can't a zero-initialized `init` be removed completely from DATA/BSS?

n8sh (approved):
This change in this PR is obviously correct. The talk about the BSS segment and memset applies to any type with `isZeroInit` and isn't specific to this PR.

jacob-carlborg:
> I suggested a test in #8707 (comment).

The PR doesn't contain that test.

WalterBright:
> isZeroInit

It's not so simple. Consider:
```
struct S { int[100] buf = void; int x; }
```
We clearly do not want to expend resources actively setting `buf` to zero. If it is allocated on the stack, it should remain uninitialized, which is quite distinct from zero-initialized. However, if it is a static, some sort of initialization will happen, might as well make it zero. And putting it in BSS makes that "for free". Note that the subject is "Enable putting..." not "Putting...". The implementation should be able to do what is most effective for each situation.

> This PR introduces a change without any tests, it is as simple as that. That is not okay in my opinion.

It's not practical to test code for 100% coverage, and this is generally accepted. It's not really practical to test that a section of memory was *not* initialized. The various test proposals above are both brittle and unreliable.

thewilsonator:
In the interest of moving forward, and the difficulty of creating a useful non-brittle test with the current test suite, can we merge this?

rainers:
I'm ok with merging without a test, but the current test does nothing. It should better be removed.

thewilsonator:
Yeah, sounds good.

rainers:
> Are you able to push a commit to revert it? Would probably be a whole lot faster.

Force pushed without that commit.

WalterBright:
It'd be nice if someone on the core team approved this so it can be merged.

rainers (approved):

PR #8709, opened 2018-09-17
refactor _isZeroInit()

WalterBright:
The function is seriously confused and overly complex. Replaced with simpler and straightforward code.


WalterBright:
This function was added by https://github.com/dlang/dmd/pull/8583

ibuclaw (approved):
Seems reasonable.

ibuclaw:
Although it looks like this subtly changes codegen to cause one of the unittests to fail.

WalterBright:
Yah, missed a case.

PR #8712, opened 2018-09-18
Fix Issue 6777 - alias this disables casting for classes

RazvanN7:
The alias this was performed eagerly, before any checks. Now, alias this is performed only if the casting failed for the aggregate.

Besides this, it is unclear what the precedence of alias this is in relation with some language features (for example: https://issues.dlang.org/show_bug.cgi?id=5363)

### DIFF
--- a/src/dmd/dcast.d
+++ b/src/dmd/dcast.d
@@ -81,6 +81,46 @@ Expression implicitCastTo(Expression e, Scope* sc, Type t)
                     result.type = t;
                     return;
                 }
+
+                auto ad = isAggregate(e.type);
+                if (ad && ad.aliasthis)
+                {
+                    MATCH adMatch;
+                    if (ad.type.ty == Tstruct)
+                        adMatch = (cast(TypeStruct)(ad.type)).implicitConvToWithoutAliasThis(t);
+                    else
+                        adMatch = (cast(TypeClass)(ad.type)).implicitConvToWithoutAliasThis(t);
+
+                    if (!adMatch)
+                    {
+                        Type tob = t.toBasetype();
+                        Type t1b = e.type.toBasetype();
+                        AggregateDeclaration toad = isAggregate(tob);
+                        if (ad != toad)
+                        {
+                            if (t1b.ty == Tclass && tob.ty == Tclass)
+                            {
+                                ClassDeclaration t1cd = t1b.isClassHandle();
+                                ClassDeclaration tocd = tob.isClassHandle();
+                                int offset;
+                                if (tocd.isBaseOf(t1cd, &offset))
+                                {
+                                    result = new CastExp(e.loc, e, t);
+                                    result.type = t;
+                                    return;
+                                }
+                            }
+
+                            /* Forward the cast to our alias this member, rewrite to:
+                             *   cast(to)e1.aliasthis
+                             */
+                            result = resolveAliasThis(sc, e);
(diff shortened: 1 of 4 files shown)


Geod24:
Doesn't this change behavior though ?
Because if what is `alias this`ed was castable to the type, it would succeed, just with an unexpected result. So I think this does need a changelog entry at the very least.

Test case:
```d
class Foo {}
class Bar { Foo f; alias f this; }
void main ()
{
   scope b = new Bar;
   b.f = new Foo;
   scope ptr = cast(void*) b; // Will cast b.f, not `b` ?
}
```

test/compilable/test6777.d
```diff
@@ -0,0 +1,11 @@
+struct S {}
+
+class C {
+    S s;
+    alias s this;
+}
+
+void main() {
```
jacob-carlborg: Curly braces on their own lines.

wilzbach: FWIW we have never really enforced the style guide for the test cases as it isn't code that gets looked at a lot (though of course it's a fair point).

src/dmd/dcast.d
```diff
@@ -72,6 +72,23 @@ Expression implicitCastTo(Expression e, Scope* sc, Type t)
...
             {
+                // if the matching was done through alias this, do not cast
+                // the base type
+                auto ad = isAggregate(e.type);
+                if (ad && ad.aliasthis)
+                {
+                    auto aliasType = ad.type.aliasthisOf();
+                    MATCH aliasMatch = aliasType.implicitConvTo(t);
+                    // if alias has the same matching level as e
+                    // but their types are different, then alias this was matched
```
WalterBright: This just seems awfully brittle (second-guessing what implicitConvTo() may have done). Secondly, it doesn't take into consideration structs. I think any fix for this should go into implicitConvTo().

RazvanN7: I fixed it now.

thewilsonator:
D language server is still red, re-adding needs work.

test/fail_compilation/fail_casting.d
```diff
@@ -147,10 +147,10 @@ void test14154()
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/fail_casting.d(179): Error: cannot cast expression `__tup$n$.__expand_field_0` of type `int` to `object.Object`
-fail_compilation/fail_casting.d(179): Error: cannot cast expression `__tup$n$.__expand_field_1` of type `int` to `object.Object`
+fail_compilation/fail_casting.d(179): Error: cannot cast expression `point` of type `Tuple14093!(int, "x", int, "y")` to `object.Object`
```
RazvanN7: Looks like my fix also fixes : https://issues.dlang.org/show_bug.cgi?id=14632

RazvanN7:
@thewilsonator I think this is good to go now

RazvanN7:
@thewilsonator Done.

aG0aep6G:
[Issue 13392](https://issues.dlang.org/show_bug.cgi?id=13392) seems to be a corner case of 6777 that's not been fixed by this PR.

UplinkCoder:
this whole thing seems like a kludge.

MoonlightSentinel:
This introduced a regression: https://issues.dlang.org/show_bug.cgi?id=21039

PR #8716, opened 2018-09-19
Fix grammar of changelog a bit for cpp mangle update

schveiguy:


Geod24:
SemaphoreCI seems stuck here. Removing auto-merge until it comes back online, feel-free to re-apply yourself.

PetarKirov (approved):

schveiguy:
semaphoreci seems to have skipped this PR. I don't know why. Trying to close and reopen.

schveiguy:
Didn't come back. ping @wilzbach can you just merge this? It's a single word change in documentation.

Geod24:
@schveiguy I think it's because the branch is in the blessed repo. Can you recreate from your fork?

Geod24:
We cannot force merge because Semaphore is set as required.

jacob-carlborg:
> We cannot force merge because Semaphore is set as required.

Would merging using the command line work?

wilzbach:
No, but I just disabled the requirement check for SemaphoreCI.

schveiguy:
> Can you recreate from your fork?

Ugh, this is what github does by default when you are a member and edit via their web interface. I do that for simple doc fixes usually because I don't want to go through the hassle of doing it locally. If I was not a member, it would create it in my own fork.

I suppose it's possible to do it in my own fork (after it's updated from upstream), and then change the PR target? I have no idea, never thought of that process before.

schveiguy:
BTW, thanks @wilzbach.

PR #8717, opened 2018-09-19
Fix styling of if/while statements.

ghost:
https://github.com/dlang/dmd/pull/8667

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -2159,27 +2159,27 @@ final class Parser(AST) : Lexer
                             cppMangleOnly = true;
                             idents = new AST.Identifiers();
 
-                            while(1)
+                            while (1)
                             {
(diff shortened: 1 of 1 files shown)


Geod24 (approved):

ibuclaw:
Weren't we introducing enforcement of style through some linter. @wilzbach?

wilzbach:
Only for Phobos and only partially there because people objeced to large-scale reformatting.

ZombineDev:
@wilzbach I thought the same, but the logs are somewhat contradicting: https://buildkite.com/dlang/dmd/builds/1370#645c7c73-470f-474f-8036-649110656dad

wilzbach:
Well we do run dscanner here since a month or so, but only trivial points have been enabled so far and especially no code style checking :/

https://github.com/dlang/dmd/blob/master/src/.dscanner.ini

PR #8718, opened 2018-09-20
add -checkaction switch

WalterBright:
This adds a `-checkaction=D|C|halt` compiler switch. It covers action taken when an assert fails, a bounds check fails, or a final switch error happens. `D` means the usual D behavior of throwing an `Error`, `C` means call the C runtime library assert failure function, and `halt` means halt the program execution.

The `halt` is the main addition here, it enables very lightweight assert's.

### DIFF
--- a/src/dmd/mars.d
+++ b/src/dmd/mars.d
@@ -1437,6 +1437,20 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
             params.useDeprecated = 2;
         else if (arg == "-c")                // https://dlang.org/dmd.html#switch-c
             params.link = false;
+        else if (startsWith(p + 1, "checkaction=")) // https://dlang.org/dmd.html#switch-checkaction
+        {
+            /* Parse:
+             *    -checkaction=D|C|halt
+             */
+            if (strcmp(p + 13, "D") == 0)
+                params.checkAction = CHECKACTION.D;
+            else if (strcmp(p + 13, "C") == 0)
+                params.checkAction = CHECKACTION.C;
+            else if (strcmp(p + 13, "halt") == 0)
+                params.checkAction = CHECKACTION.halt;
+            else
+                goto Lerror;
+        }
         else if (startsWith(p + 1, "color")) // https://dlang.org/dmd.html#switch-color
         {
             params.color = true;
--- a/changelog/checkaction.dd
+++ b/changelog/checkaction.dd
@@ -1,0 +1,8 @@
+Add `-checkaction=D|C|halt` compiler switch.
+
+It covers action taken when an assert fails, a bounds check fails,
+or a final switch error happens. D means the usual D behavior of
+throwing an `Error`, C means call the C runtime library assert failure
+function, and halt means halt the program execution.
+
+The halt is the main addition here, it enables very lightweight assert's.
--- a/src/dmd/cli.d
+++ b/src/dmd/cli.d
@@ -169,6 +169,16 @@ struct Usage
         Option("c",
             "compile only, do not link"
         ),
+        Option("checkaction=D|C|halt",
+            "behavior on assert/boundscheck/finalswitch failure",
+            `Sets behavior when an assert fails, and array boundscheck fails,
+             or a final switch errors.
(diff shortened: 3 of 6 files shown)


jacob-carlborg:
Is this flag only affecting when asserts are enabled, including when `halt` is specified?

src/dmd/mars.d
```diff
@@ -1508,6 +1508,20 @@ private bool parseCommandLine(const ref Strings arguments, const size_t argc, re
             params.useDeprecated = 2;
         else if (arg == "-c")                // https://dlang.org/dmd.html#switch-c
             params.link = false;
+        else if (startsWith(p + 1, "checkaction=")) // https://dlang.org/dmd.html#switch-checkaction
+        {
+            /* Parse:
+             *    -checkaction=D|C|halt
+             */
+            if (strcmp(p + 13, "D") == 0)
```
jacob-carlborg: Perhaps a case insensitive comparison.

WalterBright:
> Is this flag only affecting when asserts are enabled, including when halt is specified?

Yes, there are two orthogonal concepts. 1) enabling the various checks and 2) what to do when the checks are enabled. That's why there's no `off` setting for `-checkaction`

src/dmd/mars.d
WalterBright: We don't do it for any other switches.

jacob-carlborg: We don't have any other flag values that are capital letters.

wilzbach:
There's more betterC stuff which we want to be configurable (and is already configurable at GDC/LDC), e.g. useModuleInfo, useTypeInfo, useExceptions ... so it might make sense to use a more generic feature flag?

ibuclaw:
Does ophalt expand to hlt or ud2?

ZombineDev:
ud2 as of https://github.com/dlang/dmd/pull/7391.

src/dmd/mars.d
WalterBright: Nobody refers to D or C as d or c.

jacob-carlborg: Nobody uses capital letters in command line flags 😉.

WalterBright:
What's preventing this from moving forward? Having cheap asserts available is very important.

Geod24 (approved):
I agree with @jacob-carlborg 's that case-insensitive would be a bit more user friendly, but I wouldn't block the PR on that.

src/dmd/mars.d
ibuclaw: I would err on the side of @jacob-carlborg here, all cross language tools I can think off the top of my head refer to languages in lower case on the command-line.

E.g
```
gcc -x c++ foo.cc -x c bar.c
c++filt -s java
highlight -S c
```

But I don't think I should dictate what dmd should do, as I can do whatever I feel is appropriate for gdc command-line options anyway.

ibuclaw (approved):
This is fine, so long as the bikeshed topic is not blocking.

src/dmd/mars.d
jacob-carlborg: Same with Clang.

jacob-carlborg:
Is there a PR to update the documentation?

WalterBright:
I'll add that after this is pulled.

src/dmd/mars.d
klickverbot: +1 – same with other LDC switches.

wilzbach:
> I'll add that after this is pulled.

The CLI documentation gets automatically generated since a few months. Check DAutoTest, e.g.

http://dtest.dlang.io/artifact/website-64ebaad30e370d311a91075149cd3637248e257e-44e8174fb810cb968527f865e3fd8f63/web/dmd-linux.html#switch-checkaction

We even can preview the generated man page: http://dtest.dlang.io/diff/website-64ebaad30e370d311a91075149cd3637248e257e-d1d2be798b67f5905a2b1eba935c1cfe/website-64ebaad30e370d311a91075149cd3637248e257e-44e8174fb810cb968527f865e3fd8f63/web/docs/man/man1/dmd.1

PR #8719, opened 2018-09-20
Fix some of Issue 19249

thewilsonator:


wilzbach:
Isn't building with LDC already tested on appveyor?
https://github.com/dlang/dmd/blob/master/appveyor.yml

Maybe that setup could be improved to avoid further regressions...

src/dmd/root/longdouble.d
```diff
@@ -143,16 +143,21 @@ nothrow @nogc pure:
...
         {
             static      if (is(T == bool))   return mantissa != 0 || (exp_sign & 0x7fff) != 0;
-            else static if (is(T == byte))   return cast(T)ld_read(&this);
-            else static if (is(T == ubyte))  return cast(T)ld_read(&this);
-            else static if (is(T == short))  return cast(T)ld_read(&this);
-            else static if (is(T == ushort)) return cast(T)ld_read(&this);
-            else static if (is(T == int))    return cast(T)ld_read(&this);
-            else static if (is(T == uint))   return cast(T)ld_read(&this);
+            // Account for __c_[u]long
+            else static if (__traits(isIntegral,T))
```
kinke: The problem is that on Win32, the host druntime's `alias d_int32 = core.stdc.stdint.int32_t = core.stdc.config.cpp_long = __c_long` (which is a magic enum). I've had a quick look into the VS 2017 `stdint.h`, and there `int32_t` is defined as `int`, not `long`, which would fix this as well.

kinke: => https://github.com/dlang/druntime/pull/2302

kinke:
> Isn't building with LDC already tested on appveyor?

This is a `core.stdc.stdint` regression in druntime 2.081. The used LDC for AppVeyor is 1.8 = 2.078 (and performs a 32-bit-only build AFAICT).

src/dmd/root/longdouble.d
thewilsonator: Closing in favour of dlang/druntime#2302

PR #8721, opened 2018-09-21
Fix new C++ mangle syntax producing parent as Nspace.

ghost:
@TurkeyMan 
Let me know if this works for you. Do you have a sample code for any other situation? I think maybe an error message?

Don't merge yet, need to look at some other stuff -- just testing.


test/compilable/cppmangle3.d
```diff
@@ -1,4 +1,6 @@
+module cppmangle3;
 
+import std.traits : fullyQualifiedName;
```
thewilsonator: Using Phobos is not allowed in the test suite.  The test below is sufficient.

Geod24: In order to test the name, you could also add a test to `fail_compilation` which checks the error message.

TurkeyMan:
This addressed the cases I know how to reproduce.
I caught a few other error messages that printed the wrong string along the way, but I forget what I did to produce them now. trying to remember... but I reckon this will have addressed them too.

test/compilable/cppmangle3.d
TurkeyMan: That's a good idea.

look-at-me: Yah I couldn't think of one either, the only one that I could think of was removed. The symbol clashing lol.

TurkeyMan: There are heaps of error messages that do emit the symbol name. I ran into a bunch in short order... but now i've got nothing! ;)

src/dmd/dsymbol.d
```diff
@@ -382,16 +382,30 @@ extern (C++) class Dsymbol : RootObject
         return parent.pastMixin();
     }
 
+    final inout(Dsymbol) pastMixinAndNspace() inout
```
WalterBright: No Ddoc comment

TurkeyMan:
I mean, you always hassle me to squash my patches before merging them, and I never PR a line if 8 wip's ;)

PR #8722, opened 2018-09-21
[refactor] scopeForTemplateParameters

thewilsonator:
This is to reduce the diff against my template constraints DIP, which uses this.

Not sure what to do about the "// should use hasnestedArgs and enclosing?" comment, but it possibly applies to the other section.

### DIFF
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -794,6 +794,31 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
         return result;
     }
 
+    /******************************
+     * Create a scope for the parameters of the TemplateInstance
+     * ti in the parent scope sc from the ScopeDsymbol paramsym.
+     *
+     * If paramsym is null a new ScopeDsymbol is used in place of
+     * paramsym.
+     * Params:
+     *      ti         = the TemplateInstance whose parameters to generate the scope for.
+     *      sc         = the parent scope of ti
+     *      paramsym   = the ScopeDsymbol to create the scope under
+     * Returns:
+     *      a scope for the parameters of ti
+     */
+    Scope* scopeForTemplateParameters(TemplateInstance ti, Scope* sc, ScopeDsymbol paramsym)
+    {
+        if (!paramsym) paramsym = new ScopeDsymbol();
+        paramsym.parent = _scope.parent;
+        Scope* paramscope = _scope.push(paramsym);
+        paramscope.tinst = ti;
+        paramscope.minst = sc.minst;
+        paramscope.callsc = sc;
+        paramscope.stc = 0;
+        return paramscope;
+    }
+
     /***************************************
      * Given that ti is an instance of this TemplateDeclaration,
      * deduce the types of the parameters to this, and store
@@ -846,12 +871,7 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
 
         // Set up scope for template parameters
         auto paramsym = new ScopeDsymbol();
-        paramsym.parent = _scope.parent;
-        Scope* paramscope = _scope.push(paramsym);
-        paramscope.tinst = ti;
-        paramscope.minst = sc.minst;
-        paramscope.callsc = sc;
-        paramscope.stc = 0;
+        Scope* paramscope = scopeForTemplateParameters(ti,sc,paramsym);
 
         // Attempt type deduction
         m = MATCH.exact;
@@ -1124,12 +1144,7 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
 
         // Set up scope for parameters
         auto paramsym = new ScopeDsymbol();
-        paramsym.parent = _scope.parent; // should use hasnestedArgs and enclosing?
-        Scope* paramscope = _scope.push(paramsym);
-        paramscope.tinst = ti;
-        paramscope.minst = sc.minst;
-        paramscope.callsc = sc;
-        paramscope.stc = 0;
+        Scope* paramscope = scopeForTemplateParameters(ti,sc,paramsym);
 
         TemplateTupleParameter tp = isVariadic();
         Tuple declaredTuple = null;
--- a/src/dmd/dtemplate.d
+++ b/src/dmd/dtemplate.d
@@ -796,20 +796,19 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
 
     /******************************
      * Create a scope for the parameters of the TemplateInstance
-     * ti in the parent scope sc from the ScopeDsymbol paramsym.
+     * `ti` in the parent scope sc from the ScopeDsymbol paramsym.
      *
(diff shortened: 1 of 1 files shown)


src/dmd/dtemplate.d
```diff
@@ -794,6 +794,30 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
         return result;
     }
 
+    /******************************
+     * Create a scope for the parameters of the TemplateInstance
```
jacob-carlborg: Please use a short summary, an empty new line and then the description.

thewilsonator: Ah I was wondering why the doc tester was failing.

thewilsonator: ... or not. Seems like html generation is broken?

wilzbach: No you broke it with this PR:

```
dmd/dtemplate.d(810): Warning: Ddoc: parameter count mismatch
```

(From the bottom of the log)

src/dmd/dtemplate.d
```diff
@@ -794,6 +794,31 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
...
+     * ti in the parent scope sc from the ScopeDsymbol paramsym.
+     *
+     * If paramsym is null a new ScopeDsymbol is used in place of
+     * paramsym.
+     * Params:
+     *      ti:         the TemplateInstance whose parameters to generate the scope for.
+     *      sc:         the parent scope of ti
+     *      paramsym:   the ScopeDsymbol to create the scope under
```
wilzbach: Ddoc uses = for the parameters

src/dmd/dtemplate.d
thewilsonator: Eh? I might have been a musician in a previous life, but even _I_ can count to three. Also if thats compiling with "`-Werror`" it would be useful to show it as such. I searched for "Error:" and all I found was a bunch of stuff related to html output, none of which told me where in the source I stuffed up.

src/dmd/dtemplate.d
thewilsonator: like `ti       = the TemplateInstance ...`? (I've never actually used ddoc).

wilzbach: Yep and that's why DAutoTest (the doc tester) complained FWIW ;-)

See also: https://dlang.org/spec/ddoc.html#special_sections

src/dmd/dtemplate.d
wilzbach: Well because you used `:` instead of `=` for the Params it couldn't find any and thus it was 0 vs. 3.

src/dmd/dtemplate.d
thewilsonator: Yeah, but `Warning: Ddoc: parameter count mismatch` doesn't tell me that.

src/dmd/dtemplate.d
wilzbach: It's working now: http://dtest.dlang.io/artifact/website-826ea94809d870328bd98d9aeac2472b657d0c49-bd107e0b4d1624be86beb6c8553e79cf/web/phobos-prerelease/dmd_dtemplate.html#.TemplateDeclaration.scopeForTemplateParameters (the URL will be outdated with the next git push / build though)

wilzbach: > I searched for "Error:" and all I found was a bunch of stuff related to html output, none of which told me where in the source I stuffed up.

For future reference: scroll down to the end of the log and read the error stack trace from there. That's what I did. Only auto-tester is special in that way that it doesn't show you the errors at the end :/

thewilsonator: All I'm saying it that the error message should a) be an _error_ message, and b) should tell me that I need to use `=` (or even just `no = found, use = for parameters`.

src/dmd/dtemplate.d
```diff
@@ -794,6 +794,31 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
...
+     * If paramsym is null a new ScopeDsymbol is used in place of
+     * paramsym.
+     * Params:
+     *      ti         = the TemplateInstance whose parameters to generate the scope for.
+     *      sc         = the parent scope of ti
+     *      paramsym   = the ScopeDsymbol to create the scope under
+     * Returns:
+     *      a scope for the parameters of ti
+     */
+    Scope* scopeForTemplateParameters(TemplateInstance ti, Scope* sc, ScopeDsymbol paramsym)
```
wilzbach: Wouldn't it better to define `paramsym` as an `out` variable then?

src/dmd/dtemplate.d
```diff
@@ -794,6 +794,31 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
         return result;
     }
 
+    /******************************
+     * Create a scope for the parameters of the TemplateInstance
+     * ti in the parent scope sc from the ScopeDsymbol paramsym.
```
wilzbach: Doc nit: typically backticks are used to highlight variables / symbols (though DMD's codebase isn't a good role model :/

src/dmd/dtemplate.d
thewilsonator: I can't remember why I did that. I'll take another look tomorrow morning.

src/dmd/dtemplate.d
wilzbach: > a) be an error message

Well you can use `-w` to turn warnings into errors (which is what the CI uses) and warnings are an historic relict of DMD anyhow :/

> and b) should tell me that I need to use = (or even just no = found, use = for parameters.

Typically it's a mismatch, but yeah even these parameters aren't shown.
The error message could definitely be improved:

https://github.com/dlang/dmd/blob/a6dad1d2e8ae99afcce5235297407567f96c2c26/src/dmd/doc.d#L297-L300

thewilsonator: > Well you can use -w to turn warnings into errors

Then it should say `Error:` not `Warning:`. 
I see that CI has failed -> must be an error -> ^F "Error:" -> not helpful -> don't know why has failed.

src/dmd/dtemplate.d
```diff
@@ -794,6 +794,31 @@ extern (C++) final class TemplateDeclaration : ScopeDsymbol
...
+     * Params:
+     *      ti         = the TemplateInstance whose parameters to generate the scope for.
+     *      sc         = the parent scope of ti
+     *      paramsym   = the ScopeDsymbol to create the scope under
+     * Returns:
+     *      a scope for the parameters of ti
+     */
+    Scope* scopeForTemplateParameters(TemplateInstance ti, Scope* sc, ScopeDsymbol paramsym)
+    {
+        if (!paramsym) paramsym = new ScopeDsymbol();
```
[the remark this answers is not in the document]
WalterBright: In fact, `paramsym` is only used once, and since it can be derived from `paramscope`, there is no reason to return it.

PR #8723, opened 2018-09-21
convert cv8.c to D, add varstats.d

WalterBright:
A reboot of https://github.com/dlang/dmd/pull/8476

### DIFF
--- a/src/dmd/backend/cv8.d
+++ b/src/dmd/backend/cv8.d
@@ -5,65 +5,81 @@
  * Copyright:    Copyright (C) 2012-2018 by The D Language Foundation, All Rights Reserved
  * Authors:     $(LINK2 http://www.digitalmars.com, Walter Bright)
  * License:     $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost License 1.0)
- * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/code.c, backend/code.c)
+ * Source:      $(LINK2 https://github.com/dlang/dmd/blob/master/src/dmd/backend/cv8.d, backend/cv8.d)
  */
 
 // This module generates the .debug$S and .debug$T sections for Win64,
 // which are the MS-Coff symbolic debug info and type debug info sections.
 
-#if !SPP
+module dmd.backend.cv8;
 
-#include        <stdio.h>
-#include        <string.h>
-#include        <stdlib.h>
-#include        <time.h>
+version (MARS)
+{
 
-#include        "cc.h"
-#include        "el.h"
-#include        "code.h"
-#include        "oper.h"
-#include        "global.h"
-#include        "type.h"
-#include        "dt.h"
-#include        "exh.h"
-#include        "cgcv.h"
-#include        "cv4.h"
-#include        "obj.h"
-#include        "outbuf.h"
-#include        "varstats.h"
+import core.stdc.stdio;
+import core.stdc.stdlib;
+import core.stdc.string;
+extern (C) char* getcwd(char*, size_t);
 
-static char __file__[] = __FILE__;      /* for tassert.h                */
-#include        "tassert.h"
+import dmd.backend.cc;
+import dmd.backend.cdef;
+import dmd.backend.cgcv;
+import dmd.backend.code;
+import dmd.backend.code_x86;
+import dmd.backend.cv4;
+import dmd.backend.memh;
+import dmd.backend.el;
+import dmd.backend.exh;
+import dmd.backend.global;
+import dmd.backend.obj;
+import dmd.backend.oper;
+import dmd.backend.outbuf;
+import dmd.backend.rtlsym;
+import dmd.backend.ty;
+import dmd.backend.type;
+import dmd.backend.varstats;
+import dmd.backend.xmm;
 
-#if _MSC_VER || __sun
-#include        <alloca.h>
-#endif
+extern (C++):
 
-#define null NULL
-typedef unsigned char ubyte;
-typedef unsigned short ushort;
-#define mask(m) (1 << (m))
+static if (TARGET_WINDOS)
+{
 
-#if MARS
-#if TARGET_WINDOS
+int REGSIZE();
+
+// Determine if this Symbol is stored in a COMDAT
+bool symbol_iscomdat(Symbol* s)
+{
+    version (MARS)
+    {
+        return s.Sclass == SCcomdat ||
+            config.flags2 & CFG2comdat && s.Sclass == SCinline ||
+            config.flags4 & CFG4allcomdat && s.Sclass == SCglobal;
+    }
+    else
+    {
+        return s.Sclass == SCcomdat ||
+            config.flags2 & CFG2comdat && s.Sclass == SCinline ||
+            config.flags4 & CFG4allcomdat && (s.Sclass == SCglobal || s.Sclass == SCstatic);
+    }
+}
+
 
 // if symbols get longer than 65500 bytes, the linker reports corrupt debug info or exits with
 // 'fatal error LNK1318: Unexpected PDB error; RPC (23) '(0x000006BA)'
-#define CV8_MAX_SYMBOL_LENGTH 0xffd8
-
-#include        <direct.h>
+enum CV8_MAX_SYMBOL_LENGTH = 0xffd8;
 
 // The "F1" section, which is the symbols
-static Outbuffer *F1_buf;
+private __gshared Outbuffer *F1_buf;
 
 // The "F2" section, which is the line numbers
-static Outbuffer *F2_buf;
+private __gshared Outbuffer *F2_buf;
 
 // The "F3" section, which is global and a string table of source file names.
-static Outbuffer *F3_buf;
+private __gshared Outbuffer *F3_buf;
 
 // The "F4" section, which is global and a lists info about source files.
-static Outbuffer *F4_buf;
+private __gshared Outbuffer *F4_buf;
 
 /* Fixups that go into F1 section
  */
@@ -72,9 +88,9 @@ struct F1_Fixups
     Symbol *s;
     uint offset;
     uint value;
-};
+}
 
-static Outbuffer *F1fixup;      // array of F1_Fixups
+private __gshared Outbuffer *F1fixup;      // array of F1_Fixups
 
 /* Struct in which to collect per-function data, for later emission
  * into .debug$S.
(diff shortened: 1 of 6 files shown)


src/dmd/backend/varstats.d
```diff
@@ -0,0 +1,76 @@
...
+ * Authors:     Rainer Schuetze
+ * License:     Distributed under the Boost Software License, Version 1.0.
+ *              http://www.boost.org/LICENSE_1_0.txt
+ * Source:      https://github.com/dlang/dmd/blob/master/src/dmd/backend/varstats.d
+ */
+
+module dmd.backend.varstats;
+
+/******************************************
+ * support for lexical scope of local variables
```
Geod24: Make it part of the module DDOC

src/dmd/backend/varstats.d
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+import dmd.backend.cc;
+import dmd.backend.cdef;
+
+extern (C++):
+
+// estimate of variable life time
```
Geod24: DDOC

src/dmd/backend/varstats.d
```diff
@@ -0,0 +1,76 @@
...
+struct LifeTime
+{
+    Symbol* sym;
+    int offCreate;  // variable created before this code offset
+    int offDestroy; // variable destroyed after this code offset
```
Geod24: DDOC

src/dmd/backend/varstats.d
```diff
@@ -0,0 +1,76 @@
...
+struct LineOffset
+{
+    targ_size_t offset;
+    uint linnum;
+    uint diffNextOffset;
+}
+
+struct VarStatistics
+{
+    //this();
```
Geod24: Remove ?

src/dmd/backend/cv8.d
```diff
@@ -5,65 +5,82 @@
...
-typedef unsigned char ubyte;
-typedef unsigned short ushort;
-#define mask(m) (1 << (m))
-
-#if MARS
-#if TARGET_WINDOS
+module dmd.backend.cv8;
+
+version (MARS)
+{
```
Geod24: This covers the whole file, doesn't it ?

src/dmd/backend/cv8.d
```diff
@@ -5,65 +5,82 @@
...
+
+import core.stdc.stdio;
+import core.stdc.stdlib;
+import core.stdc.string;
+extern (C) char* getcwd(char*, size_t);
```
Geod24: `import core.stdc.unistd` ?

src/dmd/backend/cv8.d
```diff
@@ -5,65 +5,82 @@
...
+static if (TARGET_WINDOS)
+{
+
+int REGSIZE();
+extern (D) uint mask(uint m); //{ return 1 << m; }
+
+// Determine if this Symbol is stored in a COMDAT
+bool symbol_iscomdat(Symbol* s)
+{
+    version (MARS)
```
Geod24: Isn't this redundant given this is already in `version (MARS)` ?

src/dmd/backend/cv8.d
```diff
@@ -83,48 +100,45 @@ struct FuncData
...
 
-static Outbuffer *funcdata;     // array of FuncData's
+__gshared FuncData currentfuncdata;
 
-static Outbuffer *linepair;     // array of offset/line pairs
+private __gshared Outbuffer *funcdata;     // array of FuncData's
 
-uint cv8_addfile(const char *filename);
-void cv8_writesection(int seg, uint type, Outbuffer *buf);
+private __gshared Outbuffer *linepair;     // array of offset/line pairs
```
Geod24: Nit: `Outbuffer* linepair`

src/dmd/backend/cv8.d
```diff
@@ -133,121 +147,150 @@ void cv8_writename(Outbuffer *buf, const char* name, size_t len)
...
 {
     //printf("cv8_initfile()\n");
 
     // Recycle buffers; much faster than delete/renew
 
     if (!F1_buf)
-        F1_buf = new Outbuffer(1024);
-    F1_buf->setsize(0);
+    {
+        __gshared Outbuffer f1buf;
```
Geod24: Why did you move away from `new` ? There's a lot of redundant code here.

src/dmd/backend/cv8.d
```diff
@@ -639,73 +685,78 @@ void cv8_outsym(Symbol *s)
...
+            buf.writeWordn(0x1111);
+            buf.write32(cast(uint)(s.Soffset + base + BPoff));
+            buf.write32(typidx);
+            buf.writeWordn(I64 ? 334 : 22);       // relative to RBP/EBP
             cv8_writename(buf, id, len);
-            buf->writeByte(0);
-#else
+            buf.writeByte(0);
+}
+else
```
Geod24: Should we keep this dead code ?

src/dmd/backend/varstats.d
WalterBright: First translate, then fix/refactor.

src/dmd/backend/varstats.d
WalterBright: First translate, then fix/refactor

src/dmd/backend/cv8.d
WalterBright: yes

src/dmd/backend/varstats.d
WalterBright: first translate, then fix/refactor

src/dmd/backend/cv8.d
WalterBright: no such file

src/dmd/backend/cv8.d
WalterBright: Yes, but there are problems with where it should go. Needs a refactoring, but first translate, then fix/refactor.

src/dmd/backend/cv8.d
WalterBright: first translate, then fix/refactor

src/dmd/backend/cv8.d
WalterBright: because `new` doesn't work in betterC, and it's never free'd anyway.

src/dmd/backend/cv8.d
WalterBright: First translate, then fix/refactor.

src/dmd/backend/cv8.d
Geod24: Then `version (MARS):` would be a better transaltion

WalterBright:
I've said this many times, and will say it again. Translating code is fraught with difficulties and bugs that get introduced. It can be very hard to track down what went wrong. I've found through decades of experience that the way to minimize the problems is minimize the diffs, with the idea being the fewer diffs, the fewer places to look for the bug.

The procedure is:

1. prep the file for conversion, i.e. try to minimize the use of idioms that won't easily translate
2. pass the test suite
3. translate the code with as few edits as practical. Do not reformat the code. Do not refactor it. Do not fix anything, no matter how tempting. Reproduce the behavior of the original as much as possible.
4. pass the test suite
5. now reformat, refactor, fix (as separate PRs, of course, and passing the test suite in between each).
6. pass the test suite

I learned this the hard way, many times over, because I'm a slow learner.

src/dmd/backend/cv8.d
Geod24: Right, that is [`core.sys.posix.unistd`](https://github.com/dlang/druntime/search?q=getcwd&unscoped_q=getcwd)

WalterBright: Windows isn't Posix, and the whole file is version'd for `Posix`. Dealing with these issues is beyond the scope of this PR, and won't work anyway because DMD is compiled with an older version of DMD that would retain these issues.

Geod24:
@WalterBright : I agree with most of that. Didn't see the changes requested as things that would break with that though.

Geod24 (approved):

src/dmd/backend/cv8.d
```diff
@@ -514,23 +559,23 @@ uint cv8_addfile(const char *filename)
...
-    ubyte *p = F3_buf->buf;
+    uint length = cast(uint)F3_buf.size();
+    ubyte *p = F3_buf.buf;
     size_t len = strlen(filename);
 
     // ensure the filename is absolute to help the debugger to find the source
     // without having to know the working directory during compilation
-    static char cwd[260];
-    static uint cwdlen;
+    __gshared char[256] cwd;
```
rainers: Why change the size here? MAX_PATH is defined as 260.

src/dmd/backend/cv8.d
```diff
@@ -514,23 +559,23 @@ uint cv8_addfile(const char *filename)
...
+    __gshared uint cwdlen;
     bool abs = (*filename == '\\') ||
                (*filename == '/')  ||
                (*filename && filename[1] == ':');
 
     if (!abs && cwd[0] == 0)
```
rainers: This won't work anymore as `char` is initialized to 0xff, not 0.

src/dmd/backend/cv8.d
```diff
@@ -639,73 +684,78 @@ void cv8_outsym(Symbol *s)
...
+            {   base = cast(uint)Fast.size;
                 goto L1;
             }
             goto L2;
 
         case SCregister:
-            if (s->Sfl != FLreg)
+            if (s.Sfl != FLreg)
                 goto case_auto;
+            goto L2;
```
rainers: nit: `goto case;` would be a closer translation.

rainers (dismissed):

src/dmd/backend/cv8.d
WalterBright: A stupid mistake of mine.

src/dmd/backend/cv8.d
WalterBright: good catch

src/dmd/backend/cv8.d
WalterBright: ok

src/dmd/backend/cv8.d
```diff
@@ -5,65 +5,81 @@
...
+
+extern (C++):
+
+static if (TARGET_WINDOS)
+{
+
+int REGSIZE();
+
+// Determine if this Symbol is stored in a COMDAT
+bool symbol_iscomdat(Symbol* s)
```
rainers: Why is symbol_iscomdat defined here, if it already exists (as comment from C++ conversion of cc.h) in cc.d? It is needed elsewhere, too.

WalterBright: Because I'm doing as little as possible to translate the code to D. Refactoring comes later.

rainers: I'd consider copying the function from elsewhere more refactoring than leaving it in place but YMMV.

jacob-carlborg:
> The multiple commits can be squashed when merging (@wilzbach can this be done automatically with auto-merge?).

We should only do this for fixup commits.

rainers:
> Is the semaphoreci failure real or spurious? I couldn't find anything comprehensible in the log.

It seems to have timed out after an hour. As cv8 is unlikely to affect a linux build, I'd consider it spurious. semaphoreci isn't set to be required, so why isn't this merging with the auto-merge label being set?

rainers:
> > The multiple commits can be squashed when merging (@wilzbach can this be done automatically with auto-merge?).
> 
> We should only do this for fixup commits.

I was thinking of an additional/alternative label `auto-squash` that would have to be set.

Geod24:
Auto-merge toggled on

PR #8724, opened 2018-09-21
Add __traits(getTargetInfo) in favour of ___CXXLIB__

TurkeyMan:
Yet another alternative solution to #8701 

Spec: dlang/dlang.org#2471


test/compilable/traits.d
```diff
@@ -11,3 +11,6 @@ class C19152
         return 0;
     }
 }
+
+static assert(is(typeof(__traits(getTargetInfo, "cxxlib")) == string));
+static assert(!__traits(compiles, __traits(getTargetInfo, "not_a_target_info")));
```
ZombineDev: Don't forget about:

```D
version(CRuntime_Microsoft)
{
    static assert(__traits(getTargetInfo, "cxxlib") == "libcmt");
}
```

;)

TurkeyMan:
Obviously not finished. I just want feedback on the idea...

src/dmd/target.d
```diff
@@ -774,6 +775,38 @@ struct Target
...
+    extern (C++) static Expression getTargetInfo(const(char)* name, const ref Loc loc)
+    {
+        switch (name.toDString)
+        {
+            case "cxxlib":
+                return new StringExp(loc, global.params.mscrtlib);
+            default:
+                break;
+        }
+        return new ErrorExp();
```
ibuclaw:
```
default:
  return new ErrorExp();
```

src/dmd/target.d
```diff
@@ -774,6 +775,38 @@ struct Target
...
+    }
+
+    /**
+     * Get all supported target-info keys
+     * Returns:
+     *  Array of targetInfo keys
+     */
+    extern (C++) static const(char)** allTargetInfos(size_t* numTargetInfos)
```
jacob-carlborg: Are we allowed to use `out` parameters?

jacob-carlborg: Would be nice if D strings and arrays can be used.

src/dmd/target.d
TurkeyMan: Is this the worst you've got? :)

src/dmd/target.d
TurkeyMan: These have to go in the .h file... i figured i need to stick with in C++ for these functions?

TurkeyMan: Of course, I'd love to use dstrings and arrays!

klickverbot: >  i figured i need to stick with in C++ for these functions?

Yes – DArray with mangling support would work, but IIRC @WalterBright was against that?

src/dmd/target.d
```diff
@@ -774,6 +775,38 @@ struct Target
...
+    {
+        static const(char)*[] keys = [ "cxxlib" ];
```
klickverbot: Should probably be `immutable` (or `__gshared`) – it's going to be in TLS otherwise.

TurkeyMan: I was just trying to show the footprint of the patch. It's all done now.

TurkeyMan:
I think this is done.

TurkeyMan:
Spec: https://github.com/dlang/dlang.org/pull/2471

src/dmd/target.d
```diff
@@ -774,6 +776,60 @@ struct Target
...
+     *  name = name of targetInfo to get
+     *  loc = location to use for error messages
+     * Returns:
+     *  Expression for the requested targetInfo
+     */
+    extern (C++) static Expression getTargetInfo(const(char)* name, const ref Loc loc)
+    {
+        switch (name.toDString)
+        {
+            static foreach (e; platformInfo)
```
ibuclaw: Sorry, `static foreach` won't be allowed in the compiler for at least another year.

If someone wants to backport the implementation of static foreach to C++ so it lands in GCC-9, then you'll reduce the time you have to wait.

TurkeyMan: I thought you said that GDC and LDC have local forks of `target.d`? I put that magic in `target.d` with the specifically intent that it shouldn't affect GDC/LDC... is that wrong?

TurkeyMan:
I named it this such that all targets could populate it.

> would we map, say, libcxx vs. libstdcxx to this?

Yes, I think so.

> We use `cpp` not `cxx` in other places

Hmm, yeah I suppose. `cpp` is easy to confuse with the preprocessor. I guess that's not at risk of occurring in the context of D though...

> `cppRuntimeLibraryName`

I didn't feel `cxxlib` was cryptic, but I'll use your suggestion.

TurkeyMan:
While I'm at it, what other details should I pre-populate this PR with?
Ideally, this would include everything knowable about the target.

Does DMD have a 'target triple'? Is that a thing?

changelog/targetinfo.dd
```diff
@@ -0,0 +1,12 @@
+Add `__traits(getTargetInfo, "key")` and `__traits(allTargetInfos)` to query details about the compilation target.
+
+`getTargetInfo` accepts a string key as argument to select the target detail of interest.
+`allTargetInfos` evaluates to a tuple of strings containing all keys accepted by the implementation.
+
+---
+static assert([ __traits(allTargetInfos) ] == [ "cxxlib" ])
```
klickverbot: Do we actually need this? In which application would one want to query the available keys?

TurkeyMan: I can imagine it may be useful to 1. check if a thing you want to query is supported for the target/toolchain, 2. dump as much diagnostic information as available for debug builds and such.

There's a discoverability problem for values that might be supported only on specific backends or toolchains. I think this helps with that.

I added it because someone else asked for it. It was trivial to add, and I can imagine how it might be useful.

src/dmd/target.d
TurkeyMan: Oh. I see.
Okay, I guess I'll expand the whole thing...

TurkeyMan:
Is this okay now?

ibuclaw:
I have no problems, will let the testsuite decide the rest.

TurkeyMan:
The test suite's loving it!

src/dmd/target.d
jacob-carlborg: Now that I think about the ABI is not compatible on Windows for returning D arrays.

src/dmd/target.d
```diff
@@ -774,6 +776,37 @@ struct Target
...
+    }
+
+    /**
+     * Get all supported target-info keys
+     * Returns:
+     *  Array of targetInfo keys
+     */
+    extern (C++) static const(char*)* allTargetInfos(size_t* numTargetInfos)
+    {
+        __gshared immutable(char*[]) keys = [ "cppRuntimeLibrary" ];
```
jacob-carlborg: As I mentioned in the other PR, I think this should be an enum which can be used by the above function.

test/compilable/traits.d
```diff
@@ -11,3 +11,17 @@ class C19152
         return 0;
     }
 }
+
+static assert(is(typeof(__traits(allTargetInfos)[0]) == string));
+static assert(is(typeof(__traits(getTargetInfo, "cppRuntimeLibrary")) == string));
+static assert(!__traits(compiles, __traits(getTargetInfo, "not_a_target_info")));
```
jacob-carlborg: Should be moved to `fail_compilation` to be able to test the error message.

jacob-carlborg (changes requested):
As mentioned in the other PR several of the errors are missing tests.

test/fail_compilation/traits.d
```diff
@@ -0,0 +1,24 @@
...
+/************************************************************/
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/traits.d(100): Error: getTargetInfo key "not_a_target_info" not supported by this implementation
+fail_compilation/traits.d(101): Error: string expected as argument of __traits getTargetInfo instead of 100
+fail_compilation/traits.d(102): Error: expected 1 arguments for getTargetInfo but had 2
+fail_compilation/traits.d(103): Error: expected 1 arguments for getTargetInfo but had 0
+fail_compilation/traits.d(104): Error: expected 0 arguments for allTargetInfos but had 1
```
ibuclaw: actual:
```
fail_compilation/traits.d(100): Error: `getTargetInfo` key `"not_a_target_info"` not supported by this implementation
fail_compilation/traits.d(101): Error: string expected as argument of __traits `getTargetInfo` instead of `100`
fail_compilation/traits.d(102): Error: expected 1 arguments for `getTargetInfo` but had 2
fail_compilation/traits.d(103): Error: expected 1 arguments for `getTargetInfo` but had 0
fail_compilation/traits.d(104): Error: expected 0 arguments for `allTargetInfos` but had 1
```

TurkeyMan: That's not what I get when I run locally...
I literally cut the pasted the error messages from the local execution :/

TurkeyMan: Why is everything so bloody hard and time-wasting! >_<

TurkeyMan: Ready for the windows tests to fail now...?

src/dmd/target.d
```diff
@@ -774,6 +776,37 @@ struct Target
...
+
+    /**
+     * Get targetInfo by key
+     * Params:
+     *  name = name of targetInfo to get
+     *  loc = location to use for error messages
+     * Returns:
+     *  Expression for the requested targetInfo
+     */
+    extern (C++) static Expression getTargetInfo(const(char)* name, const ref Loc loc)
```
WalterBright: Should return a StringExp.

WalterBright: No `static` for function declarations.

WalterBright: Please, use D strings, not `char*`

WalterBright:
I don't understand the use case for `allTargetInfos`

src/dmd/target.d
ibuclaw: > No static for function declarations.

All functions are static in this interface.

TurkeyMan: > Should return a StringExp.

It doesn't just return strings.

> No `static` for function declarations.

Every function in `Target` is static.

> Please, use D strings, not `char*`

You specifically didn't want to mangle D strings for C++. I know other people wanted to, but you rejected it. As such, the API's must still use C strings.

TurkeyMan:
I was responded to review feedback, it was specifically requested.
I can imagine it may be useful if you want to dump a large set of info for debug logging, or for discoverability. I suspect it's moderately useful.

jacob-carlborg (approved):

test/fail_compilation/traits.d
```diff
@@ -0,0 +1,24 @@
...
+---
+fail_compilation/traits.d(100): Error: `getTargetInfo` key `"not_a_target_info"` not supported by this implementation
+fail_compilation/traits.d(101): Error: string expected as argument of __traits `getTargetInfo` instead of `100`
+fail_compilation/traits.d(102): Error: expected 1 arguments for `getTargetInfo` but had 2
+fail_compilation/traits.d(103): Error: expected 1 arguments for `getTargetInfo` but had 0
+fail_compilation/traits.d(104): Error: expected 0 arguments for `allTargetInfos` but had 1
+---
+*/
+
+#line 100
```
jacob-carlborg: Minor detail, I would avoid using `#line`, instead the `AUTO_UPDATE=1` [1] flag can be used to update the `TEST_OUTPUT` section.

[1] https://github.com/dlang/dmd/tree/5dba233b45f639594bafa0138ee936c33210b956/test#automatically-update-the-test_output-segments

ibuclaw:
> I was responded to review feedback, it was specifically requested.

More specifically.  The initial request was to replace `__CXXLIB__` with `__traits(getCxxLib)`.  on the basis that I don't think that the former should have been allowed in, and the latter is the way that this compile time information should have been exposed as in the first place.

This eventually became `__traits(getTargetInfo, "cxxlib")`, and then naturally leading to #8728.

src/dmd/target.h
```diff
@@ -86,4 +86,6 @@ struct Target
     static TypeTuple *toArgTypes(Type *t);
     static bool isReturnOnStack(TypeFunction *tf, bool needsThis);
     static d_uns64 parameterSize(const Loc& loc, Type *t);
+    static Expression getTargetInfo(const char* name, const Loc& loc);
+    static const char *const * allTargetInfos(size_t* numTargetInfos);
```
ibuclaw: Are you sure you checked the signature is correct here?

TurkeyMan: Yes: `extern (C++) static const(char*)* allTargetInfos(size_t* numTargetInfos)`
Is there a problem?

TurkeyMan: This use of const mangles correctly. It's not a head-const.
Do you think this PR will survive?

ibuclaw (approved):
I have no problem with this. Don't really have any strong preference over what you name the trait, just so long as the cxxlib information is exposed from a trait.

jacob-carlborg:
Feel free to merge before 72 hours if anyone would like to.

src/dmd/target.d
WalterBright: It returns a StringExp or null, meaning the return type can be StringExp.

WalterBright (changes requested):
Fix the return type, and I'll approve.

TurkeyMan:
But this particular patch only returns one thing, but the API is intended to return lots of things.

src/dmd/target.d
jacob-carlborg: There’s a follow up PR that returns `IntegerExp` as well.

jacob-carlborg:
> But this particular patch only returns one thing, but the API is intended to return lots of things.

Change the return type here and change it back in the other PR the get things rolling.

TurkeyMan:
There's no guarantee the other one will get merged, and this API is absolutely intended to return `Expression`. This PR is correct as-is, and it's holding up GDC.
This API isn't define by what DMD's `target.d` returns; DMD's `target.d` will be mostly uneventful because the backend has no interesting information to share, GDC and LDC have a whole lot more information to offer, like target CPU details, cache sizes, etc. If this is all that gets merged in DMD, it needs to enable GDC/LDC to achieve the intended goal, otherwise GDC/LDC will just have to push through another DMD patch to change this return type to `Expression` like it already is.

The cache sizes are super-useful data. The code I'm writing at work right now would benefit so much from knowing that detail to tune some buffer sizes and tweak iteration strides.

TurkeyMan:
So... I don't intend to change this to `StringExp`... should I close the PR?

jacob-carlborg:
> So... I don't intend to change the return type to StringExp... should I close the PR?

I would really hate to see that.

jacob-carlborg:
Would it be such a pain to change it to `StringExp` and change it back. I know it's stupid.

TurkeyMan:
I think it's reasonable to assume that if this is rejected on that account, then changing it in a follow-up PR will also be rejected. I undermine a good part of my motive for creating this API by changing it.

12345swordy:
@TurkeyMan Please don't throw the baby out with the bathwater. Just make the changes as @WalterBright requested, and follow up with a separate pull-request with the changes you wanted. I hate to see good work wasted, because of a small dispute.

TurkeyMan:
The baby is the bathwater. I can already move forward with `__CXXLIB__`, this PR is about GDC/LDC having a well defined hook to make backend data available. That data is not necessarily string.
I don't want to see `to!int(__traits(targetInfo, "whatever"))`, which is effectively the substance of the request here.

I think this is much better design than `__CXXLIB__` and it should merge, but if it's rejected for not-strings, then its value is arbitrary limited. If we know up-front that it's rejected for int, then the follow up you both speak of is implicitly rejected... why would you expect it'll be merged in a follow up when it's rejected now?

TurkeyMan:
You can amend my PR if you want to. But I can't do it.

src/dmd/target.d
thewilsonator: > Should return a StringExp.

No it should _not_, yes _this_ implementation _currently_ returns only strings, but this is not the only implementation. This will create an unnecessary upstream diff for LDC and GDC.

thewilsonator:
@WalterBright 

>this PR is about GDC/LDC having a well defined hook to make backend data available

strings are not integers, string are not tuples, string are strings and do not cover all uses of this API.

As I said in #8728 the _entire_ purpose of this is to be _extensible_ and discoverable. Wether or not this is extended further in DMD is irrelevant, it _will_ be extended in LDC, and I presume GDC, and `Expression` is the correct return type.

(We _really_ ought to have some procedure for dealing with objections to `72h no objection -> merge` where the intent on all sides, except for the person raising the objection, is to proceed when that person does not respond further.)

src/dmd/target.d
WalterBright: Ok, returning Expression is ok.

WalterBright:
> I was responded to review feedback, it was specifically requested.

For what purpose?

> I can imagine it may be useful if you want to dump a large set of info for debug logging, or for discoverability. There's no other mechanism to query what an implementation supports. Alternative would be implementation maintainers keep a list on a webpage somewhere; it will ALWAYS be stale, and almost certainly inaccurate/incomplete.
I suspect it's moderately useful.

Let's wait until a compelling rationale is found before just adding things. I suggest removing that feature until then. D is a very feature-laden language already.

thewilsonator:
> For what purpose?

As I have said before this is about extensibility and discoverability. 

There will end up being quite a lot of information available through this trait, and as more information is added, much of which will be compiler/compiler middle and backend (version!)/target/system/command line flag/etc specific, and older compiler versions lag behind it will become difficult to know what information is available for a given build configuration, let alone collate it together for diagnostics.

With this, collating information for diagnostics will be as simple as:
```d
static foreach( ti; __traits(allTargetInfo))
      log(ti, ":",  __traits(getTargetInfo,ti));
```

This discoverability also manifests when upgrading compilers, what new target information is available (likely scattered across multiple sets of release notes)? What ones are applicable to the systems I target? Which ones are actually useful for what I want? It is hard to answer the latter questions without knowing which ones are available.

> D is a very feature-laden language already.

Yes but it is important to consider the _size vs impact_ of said features: Adding a trait is very easy ( getting it past review is another question :) ), it is very easy to use, is not significantly language changing (thats the whole point of traits) and the value it adds is real. 

` __traits(getTargetInfo,...)` significantly aids discoverability.

TurkeyMan:
It has trivial surface area, and virtually no impact on the language. But it makes this unknown quantity of data discoverable. There are lots of compilers with lots of different versions, knowing what's in this box is impossible without a query.

Also, why wait over a week to give this criticism? Why not in the last round when you said change `Expression` and you'll merge. This process is agonisingly slow enough already. Everyone else here reviewed favorably within a day but we're still held up. Do you not trust your community of reviewers?

WalterBright:
> It has trivial surface area, and virtually no impact on the language.

That's not a reason to put it in the language. To go in, it must have a compelling positive reason.

> But it makes this unknown quantity of data discoverable. There are lots of compilers with lots of different versions, knowing what's in this box is impossible without a query.

Knowing that the compiler supports a "blart" data is useless without consulting the documentation.

For example, C compilers have a switch to spit out print out all the macros that are predefined. But it tells you nothing about what those macros expand to or why they are predefined. In fact, often there is no documentation for those macros. (It's still useful because it guides the preprocessor output, but that reason doesn't transfer to this D feature.)

In fact, one compiler may define "blart" to mean one thing and another may defined it differently. I don't believe such a feature is useful without having the documentation correct and up to date anyway.

> Also, why wait over a week to give this criticism? Why not in the last round when you said change Expression and you'll merge. This process is agonisingly slow enough already. Everyone elses reviews were resolved within a day but we're still held up. Do you not trust your community of reviewers?

Several issues here.

1. Core language changes need my approval. 72  hrs default approval is not appropriate for this - it's only appropriate for minor and routine things, such as changing scripts or moving code between branches.
2. This PR has two separate language changes in it. One is held up because of the other. Trying to push through approval on one because you need the other is exactly why such combination PRs are not the right way to do things.
3. "Why not" is never an acceptable rationale for a new feature. Features must have a compelling positive rationale for it. Otherwise, D will capsize like the Vasa.
4. We shouldn't be so hasty about adding language enhancements, certainly one day turnaround is awfully hasty. I approved the last one, __CXXLIB__, only to have it dumped a few days later with this one.
5. This is not blocking your work, although I acknowledge there is some inconvenience for the first part, for the `all` enhancement there is no use case in the code you're working on.

TurkeyMan:
> > It has trivial surface area, and virtually no impact on the language.

> That's not a reason to put it in the language. To go in, it must have a compelling positive reason.

Sure, and I feel like we made that case.

> > But it makes this unknown quantity of data discoverable. There are lots of compilers with lots of different versions, knowing what's in this box is impossible without a query.

> Knowing that the compiler supports a "blart" data is useless without consulting the documentation.

Umm. I disagree. But okay.
If you're looking for something, you'll know when you see it. And you can easily dump all data with:
```d
static foreach( ti; __traits(allTargetInfo))
      log(ti, ":",  __traits(getTargetInfo,ti));
```
And that should be extremely revealing.

> consulting the documentation

I expect no such documentation will exist. No way all compilers will keep all data up to date on a webpage somewhere for all targets. And then there's experimental or niche targets that get very little use like CUDA/SPIRV, or microcontrollers, etc. Nobody will maintain that. We can't even maintain the most high-traffic documentation effectively. Niche users of niche features of niche toolchains are doomed!

I think it's useful. We can't make that case any better than we have.
It's also trivial to emit the list of things that is supported by `getTargetInfo`, the cost for this is very low.

> C compilers have a switch to spit out print out all the macros that are predefined

That feels like a bit of a straw-man. C++ macros are not sources of hard data, and that dump you speak of is also laden with countless aliases where 40 years of compiler baggage has been accumulated into a big list of macros. Most of the bump is legacy from a time where there was absolutely no attempt to standardise the information delivery between compiler vendors. We do not suffer that here; LDC and GDC appear to strongly push towards standardisation in every example I've ever seen, and somewhat strangely, it's often DMD that resists their efforts.

> I approved the last one, `__CXXLIB__`, only to have it dumped a few days later with this one.

That's somewhat fair, but it's not like `__CXXLIB__` just appeared either. There was a lot of discussion, iteration and repeated amendments prior to that too.
Reviewers were fine with it... but @ibuclaw had a better proposal.
So while it was superseded within a couple of days, it's still the case that I've been peddling this trivia for.... probably well over a month now, just kinda waiting to get on with stuff.

I'm not sure how I could have changed the process to arrive here in the first place. I think it was just when Iain looked at the patch with intent to merge it into GDC that he raised the issue, so it was a product of _merging_ that other patch that motivated this change.

LDC and GDC are much closer to the front-line now... what work can be done in the future to bring them right to the front-line?

WalterBright:
We disagree on several points, and that's fair. But what we do agree on is `getTargetInfo`. So I recommend separating `allTargetInfo` into a separate PR, and we can merge `getTargetInfo`.

thewilsonator:
> To go in, it must have a compelling positive reason.

It must have reason compelling _relative to the complexity of the feature_. This is a `pressure = Force / Area` thing: the compulsion of the Force is reasonable, the area is tiny...

>Knowing that the compiler supports a "blart" data is useless without consulting the documentation.

Knowing that the compiler supports a "blart" is the first step to consulting the documentation (or asking for information about what it is if the docs are not up to scratch). -- Budda (probably)

>In fact, one compiler may define "blart" to mean one thing and another may defined it differently. I don't believe such a feature is useful without having the documentation correct and up to date anyway.

Compilers do not deal in "blarts", they deal in CPU names/features, cache sizes and so on: are you seriously suggesting that things in those categories are going to coincide with different meaning between different compilers? We are looking to unify the differences between the extensions of what GDC and LDC provide, not make them more different. 
As for documentation, much this information will be derived from the backend which changes with every release, it is not feasible to document everything, but the name alone should provide sufficient insight to understand the values one gets back from it or seek further information based on it, e.g upstream documentation (see also: the above).

> 3. "Why not" is never an acceptable rationale for a new feature

Correct, and that is not the reason we have given.

>Features must have a compelling positive rationale for it. Otherwise, D will capsize like the Vasa.

Yes, see the top of this post.

> there is no use case in the code you're working on.

False: diagnostics was mentioned several times.

> certainly one day turnaround is awfully hasty

_This_ PR was opened a week ago, there were other PRs with releated discussion before that. If your timeframe is in reference to acceptance of `Expression` as the return type, then fair enough. But please don't hesitate to give all your criticism all at once, it makes things much faster :)

jacob-carlborg:
> Core language changes need my approval.

I disagree that this is a core language change.

TurkeyMan:
Split to #8763... It should still exist though.

WalterBright (approved):

PR #8726, opened 2018-09-21
test/runnable: Remove references to Project Gutenberg

ibuclaw:
As per https://github.com/CyberShadow/DustMite/issues/50

Patch taken from https://github.com/CyberShadow/DustMite/commit/e95dff8987c01f2436370a41f96f4d71c9a096ef

This was picked up by gcc review as well.


dnadlinger (approved):
Shouldn't be an issue as far as I can tell, but if it makes people sleep more easily…

CyberShadow:
The text says:

> if you either delete this "Small Print!" **and all other references to Project Gutenberg**

so I don't believe that would be enough.

PR #6656, opened 2017-03-26
refactor: replace 'goto Lerror;' with 'return errorReturn();'

WalterBright:
Better than a rat's nest of goto's.

PR #8728, opened 2018-09-22
More target info

TurkeyMan:
Further populate `__traits(targetInfo)`

### DIFF
--- a/src/dmd/target.d
+++ b/src/dmd/target.d
@@ -717,6 +717,14 @@ struct Target
         return global.params.is64bit ? (sz + 7) & ~7 : (sz + 3) & ~3;
     }
 
+    // this guarantees `getTargetInfo` and `allTargetInfos` remain in sync
+    private enum TargetInfoKeys
+    {
+        cppRuntimeLibrary,
+        floatAbi,
+        objectFormat,
+    }
+
     /**
      * Get targetInfo by key
      * Params:
@@ -727,10 +735,30 @@ struct Target
      */
     extern (C++) static Expression getTargetInfo(const(char)* name, const ref Loc loc)
     {
-        switch (name.toDString)
+        StringExp stringExp(const(char)[] sval)
         {
-            case "cppRuntimeLibrary":
-                return new StringExp(loc, cast(void*)global.params.mscrtlib, global.params.mscrtlib ? strlen(global.params.mscrtlib) : 0);
+            return new StringExp(loc, cast(void*)sval.ptr, sval.length);
+        }
+
+        switch (name.toDString) with (TargetInfoKeys)
+        {
+            case objectFormat.stringof:
+                if (global.params.isWindows)
+                    return stringExp(global.params.mscoff ? "coff" : "omf");
+                else if (global.params.isOSX)
+                    return stringExp("macho");
+                else
+                    return stringExp("elf");
+            case floatAbi.stringof:
+                return stringExp("hard");
+            case cppRuntimeLibrary.stringof:
+                if (global.params.isWindows)
+                {
+                    if (global.params.mscoff)
+                        return stringExp(global.params.mscrtlib ? global.params.mscrtlib.toDString : "");
+                    return stringExp("snn");
+                }
+                return stringExp("");
             default:
                 return null;
         }
--- a/test/compilable/traits.d
+++ b/test/compilable/traits.d
@@ -13,8 +13,17 @@ class C19152
 }
 
 static assert(is(typeof(__traits(getTargetInfo, "cppRuntimeLibrary")) == string));
-
 version (CppRuntime_Microsoft)
 {
     static assert(__traits(getTargetInfo, "cppRuntimeLibrary") == "libcmt");
 }
+
+version (D_HardFloat)
+    static assert(__traits(getTargetInfo, "floatAbi") == "hard");
+
+version (Win64)
+    static assert(__traits(getTargetInfo, "objectFormat") == "coff");
+version (OSX)
+    static assert(__traits(getTargetInfo, "objectFormat") == "macho");
+version (linux)
+    static assert(__traits(getTargetInfo, "objectFormat") == "elf");


src/dmd/target.d
```diff
@@ -774,6 +776,90 @@ struct Target
...
+    }
+
+    /**
+     * Get all supported target-info keys
+     * Returns:
+     *  Array of targetInfo keys
+     */
+    extern (C++) static const(char*)* allTargetInfos(size_t* numTargetInfos)
+    {
+        __gshared immutable(char*[]) keys = [ "arch", "bits", "cppRuntimeLibrary", "floatAbi", "objectFormat", "pointerBits", "system" ];
```
jacob-carlborg: This could be an enum which can be used in the above function to avoid having to repeat this list.

test/compilable/traits.d
```diff
@@ -11,3 +11,13 @@ class C19152
         return 0;
     }
 }
+
+static assert(is(typeof(__traits(allTargetInfos)[0]) == string));
+static assert(is(typeof(__traits(getTargetInfo, "bits")) == int));
+static assert(is(typeof(__traits(getTargetInfo, "cppRuntimeLibrary")) == string));
+static assert(!__traits(compiles, __traits(getTargetInfo, "not_a_target_info")));
```
jacob-carlborg: Please move this to `fail_compilation` to be able to test the error message.

src/dmd/traits.d
```diff
@@ -1776,6 +1778,44 @@ Lnext:
...
+    if (e.ident == Id.getTargetInfo)
+    {
+        if (dim != 1)
+            return dimError(1);
+
+        auto ex = isExpression((*e.args)[0]);
+        StringExp se = ex ? ex.ctfeInterpret().toStringExp() : null;
+        if (!ex || !se || se.len == 0)
+        {
+            e.error("string expected as argument of __traits `%s` instead of `%s`", e.ident.toChars(), ex.toChars());
```
jacob-carlborg: This is not covered by the tests.

src/dmd/traits.d
```diff
@@ -1776,6 +1778,44 @@ Lnext:
...
+            return new ErrorExp();
+        }
+
+        se = se.toUTF8(sc);
+        if (se.sz != 1)
+        {
+            e.error("string must be chars");
```
jacob-carlborg: This is not covered by the tests.

src/dmd/traits.d
```diff
@@ -1776,6 +1778,44 @@ Lnext:
...
+            return new ErrorExp();
+        }
+
+        Expression r = Target.getTargetInfo(se.toPtr(), e.loc);
+        return r.expressionSemantic(sc);
+    }
+    if (e.ident == Id.allTargetInfos)
+    {
+        if (dim != 0)
+            return dimError(0);
```
jacob-carlborg: This is not covered by the tests.

src/dmd/traits.d
```diff
@@ -1776,6 +1778,44 @@ Lnext:
         Type tb = t.baseElemOf();
         return tb.isZeroInit(e.loc) ? True() : False();
     }
+    if (e.ident == Id.getTargetInfo)
+    {
+        if (dim != 1)
+            return dimError(1);
```
jacob-carlborg: This is not covered by the tests.

jacob-carlborg (changes requested):

src/dmd/target.d
TurkeyMan: It's not clear what you have in mind... how do drive a switch by a string array?
Switched on indexed lookups and hope the cases stay in sync?

TurkeyMan:
Spec for this PR: https://github.com/dlang/dlang.org/pull/2473

WalterBright:
We already have predefined versions for these. What's the point?

TurkeyMan:
I think data about the target should be available in a consolidated place, and in the canonical form.
Versions are for grey-ing out code, they're not for providing information or data. If you wanted to get any of this data in canonical form, you'd need to do:
```d
version(Windows)
  enum system = "windows";
version(OSX)
  enum system = "darwin";
version(iOS)
  enum system = "darwin";
version(Linux)
  enum system = "linux";
version(FreeBSD)
  enum system = "freebas";
version(OpenBSD)
  enum system = "openbsd";
version(DragonflyBSD)
  enum system = "dragonfly";
version(Solaris)
  enum system = "solaris";
version(Playstation4)
  enum system = "ps4";
version(Playstation3)
  enum system = "ps3";
version(XBoxOne)
  enum system = "xbone";
....etc (this could go forever!)
```
And you'd have to have blocks like that for every single piece of data you want to make available.
Versions are a terrible data-source.
There are a *lot* of targets. LDC and GDC target a huge number of machines. Some of them present different architectural details that are target-specific, those details can be made available in a consistent and canonical way.

I have made so many requests for compile-time knowledge for so many years. It's usually added, and just wasted unnecessary turn-around time to the end-user.
If we have this, we can stop making those requests to DMD for a large amount of that data.

This patch allows toolchain's forks of `target.d` to do the work, without further impacting DMD in any way.

ibuclaw:
> but Iain hated my patches as usual, 

Woah now, hate is a strong word.  But yes, I am the catalyst here.

TurkeyMan:
;)

src/dmd/target.d
thewilsonator:
```d
private extern(D)  enum TargetInfoKeys
{
     arch = "arch",
    //...
}
//
switch (cast(TargetInfoKeys)name.toDString) with(TargetInfoKeys)
         {
             case arch:
                //...
//
extern (C++) static (DArray!(const char)* allTargetInfos(size_t* numTargetInfos)
{
      __gshared DArray!(const char)[] keys = cast(DArray!(const char)[]) [ __traits(allMembers, TargetInfoKeys) ];
      *numTargetInfos = keys.length;
      return keys.ptr;

}
```

thewilsonator:
@WalterBright We also have a bunch of code that does basically the same thing in LDC. It would be nice to have common infrastructure for such things.

https://github.com/ldc-developers/ldc/blob/master/tests/semantic/target_traits.d#L12

test/fail_compilation/traits.d
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+/************************************************************/
+
+/*
+TEST_OUTPUT:
+---
+fail_compilation/traits.d(100): Error: `getTargetInfo` key `"not_a_target_info"` not supported by this implementation
+fail_compilation/traits.d(101): Error: string expected as argument of __traits `getTargetInfo` instead of `100`
+fail_compilation/traits.d(102): Error: expected 1 arguments for `getTargetInfo` but had 2
```
thewilsonator: Very minor nit. Most error messages use found, so:
"string expected as argument of __traits `getTargetInfo`, found `100`"
"expected 1 arguments for `getTargetInfo`, found 2"

src/dmd/target.d
thewilsonator: Good, improvement is always welcome.

thewilsonator: `__gshared auto` would probably be a start.

test/fail_compilation/traits.d
TurkeyMan: As usual, I copy&pasted the error message from the ones nearby to maintain the same language as the surrounding precedent.
The error message about the argument count is emit by a utility function that produces the same error for EVERY `__traits`.

I don't know how many more times I can be told to change code that I copied from a few lines away ;)

thewilsonator: Haha.

>I don't know how many more times I can be told to change code that I copied from a few lines away ;)

Lots ;) Don't worry about it.

TurkeyMan:
Also, what's with semaphore? Why doesn't it ever work? Is it my fault?

thewilsonator:
I'm pretty sure it timing out.

WalterBright:
> you'd need to do:

You'd only need to do it once, and then import it. Secondly, there's no evidence that the data from __traits will be in a suitable form for the application, so the application will have to do the equivalent anyway. Thirdly, this can always be done via a library module.

thewilsonator:
> You'd only need to do it once, and then import it. 

You would have to do it for _every_ combination of `version`able information. Which is the whole point of this: extensibility and discoverability. 
To give an example:  for dcompute there is a ton of information available (for a given GPU: available memory, cache sizes, feature availability etc.) that I would love to make available but I can't make a trait for each of them because there are too many of them.

> Thirdly, this can always be done via a library module.

It cannot.
* A library solution, in order to provide accurate information, would have to be specific to every compiler vendor/release combination.
*  Compiler would have to provide those versions or a special keyword (á la `__CXXLIB__` which Iain objected to) anyway. 
* Versions do not cover all uses cases of this: e.g. using `-march=native` what is the size of the L2 cache of this CPU?, GDC and LDC know that, because they can ask their backends.

> Secondly, there's no evidence that the data from __traits will be in a suitable form for the application,

There is: precedence in the form of LDC's `targetCPU`. It is very easy to use, and "does one thing well". 

Generally the kind of data one is going to get from `__traits` is either going to be:
* a string that one can "`static switch`" on, as in the case of `targetCPU`. (Yes this could be done with versions, but it suffers the combinatorial compiler vendor/release problem)
* integer data that forms constants to behaviour for optimisation e.g. control loop unrolling so it fits in cache.

WalterBright:
You're proposing replacing:
```
version (Windows)
```
with:
```
static if (__traits(getTargetInfo, "system") == "windows")
```
This doesn't look good. In fact, it looks awful.

thewilsonator:
This does not replace, it is a complement.

WalterBright:
Having n ways to do the same thing is not an improvement.

thewilsonator:
No, but it is not making things strictly worse (also n=2).

Being able to do what you couldn't before is an improvement. I'd like to reiterate that the purpose of this is extensibility and discoverability. Yes, we have versions for the things that are in _this PR_ (except for CppRuntimeLibrary, which was the straw that made Iain say no), but it is a lot easier to extend this than to add versions (which don't cover everything), and there will be a lot more in LDC and given that Iain wanted this a lot more in GDC.

TurkeyMan:
This isn't a replacement for versions. Versions are for grey-ing out code. This is not for that. These provide diagnostic information. Versions don't provide diagnostic information, they grey out code.

src/dmd/target.d
TurkeyMan: Huh? It is currently `__gshared immutable`...?

thewilsonator: Sorry I meant improvement over what I wrote, which was just `__gshared`.

jacob-carlborg:
> We already have predefined versions for these. What's the point?

Having this information in as part of `__traits` allows to use it where runtime values are otherwise expected. For example, the below code could be part of a script that downloads DMD for the current running platforms.

```d
void main()
{
    enum urls = [
        "darwin": "http://downloads.dlang.org/releases/2.x/2.082.0/dmd.2.082.0.osx.tar.xz",
        "linux": "http://downloads.dlang.org/releases/2.x/2.082.0/dmd.2.082.0.linux.tar.xz",
        "windows": "http://downloads.dlang.org/releases/2.x/2.082.0/dmd.2.082.0.windows.7z"
    ];
    
    enum url = urls[__traits(getTargetInfo, "system")];
}
```

I'm using a similar pattern in one of my scripts to download LLVM. But I need to explicitly defined which is the current system with the help of `version` blocks.

With the help of this new `__traits` we can create a nice API that works both at compile time and runtime (we discussed this in Slack):

```d
import core.target;

enum system = Target.system; // works at compile time

void main()
{
    auto system = Target.system; // works at runtime
    auto vers = Target.systemVersion; // only available at runtime
}
```

This will also provide one interface to get target information instead of as it is now split between `version` blocks and magic constants. This also avoids adding more reserved symbols, even though they start with `__` the compiler doesn't prevent users from using identifiers starting with `__`.

src/dmd/target.d
```diff
@@ -774,6 +776,90 @@ struct Target
...
+                else if (global.params.isFreeBSD)
+                    sval = "freebsd";
+                else if (global.params.isOpenBSD)
+                    sval = "openbsd";
+                else if (global.params.isDragonFlyBSD)
+                    sval = "dragonfly";
+                else if (global.params.isSolaris)
+                    sval = "solaris";
+                else
+                    sval = "unknown";
```
jacob-carlborg: Is it better to return `null`?

src/dmd/target.d
jacob-carlborg: I don't like that the keys are repeated twice. I also don't like that there's no easy way to visually see all the available keys in the source code.

src/dmd/target.d
TurkeyMan: perhaps this should just be an assert() since this case should be impossible unless DMD is broken somehow.

thewilsonator: Baremetal?

TurkeyMan: I don't think DMD can do that. If all these are false, I expect DMD fails? (I doubt it's possible to invoke?)

src/dmd/target.d
TurkeyMan: I fixed it.

jacob-carlborg (approved):

WalterBright:
I don't agree with adding __traits that are redundant with versions. It's just more and more stuff heaped into the core language. Add it as a library routine in druntime if you really want it. Things that are doable as library routines really should not be in the core language.

WalterBright (changes requested):
Remove stuff that is redundant with versions.

TurkeyMan:
I'll do what you say, but I think it's the wrong call.. it's not redundant, it's complementary.

thewilsonator:
> I'll do what you say, but I think it's the wrong call.. it's not redundant, it's complementary.

It'll be easy to add back if Walter changes his mind.

> Things that are doable as library routines really should not be in the core language.

@WalterBright This is already not the case with `__traits`:

* isArithmetic
* isFloating
* isScalar
* isUnsigned
* isStaticArray
* isAssociativeArray
* isZeroInit
* getPointerBitmap

are all _doable_ in library, yet they serve a purpose as traits. This PR is no different, I refer you to 
https://github.com/dlang/dmd/pull/8728#issuecomment-423788778 The compiler is the authoritative source on the name of the system and can include values that are otherwise unchecked by users version switching.

Anyway lets get this merged (Grumble, grumble, semaphore).

TurkeyMan:
So, I think this is a minimal set that @WalterBright can agree on right?

WalterBright (approved):

ibuclaw:
Should there be tests for these for coverage sake?

src/dmd/target.d
```diff
@@ -787,13 +795,39 @@ struct Target
...
+                else
+                    sval = "";
+                goto LstringExpr;
             default:
                 return null;
         }
+    LintExpr:
+        return new IntegerExp(loc, ival, Type.tint32);
+    LstringExpr:
+        return new StringExp(loc, cast(void*)sval.ptr, sval.length);
```
ZombineDev: Replace gotos with inner functions. See for example: https://github.com/dlang/dmd/pull/6656/files

PetarKirov (approved):
Looks good as a starting point. It remains to be seen how easy would be for LDC and GDC to extend the behavior. I suppose that if `Target` was a class, a natural approach would be to make `getTargetInfo` a virtual function. @kinke do you have any thoughts on the matter from LDC's point of view?

ibuclaw:
@ZombineDev I would hope that ldc implements their own version of dmd.target, otherwise they are doing it wrong.

https://github.com/D-Programming-GDC/GDC/blob/master/gcc/d/d-target.cc

thewilsonator:
We do: https://github.com/ldc-developers/ldc/blob/master/gen/target.cpp

src/dmd/target.d
TurkeyMan: I was inspired by code literally all over DMD to use that style...

ZombineDev: I guessed so, but we want to remove bad practices from the code base, see also: https://forum.dlang.org/thread/p6oibo$1lmi$1@digitalmars.com

TurkeyMan: Incidentally, I think the local functions in your example are possibly uglier than the code they replace (some goto labels)... but they're certainly more 'principled' ;)

ibuclaw:
@TurkeyMan, there's also a number of fields inside this Target structure that may be of interest to library maintainers. To pick a few random ones:
- classinfosize
- maxStaticDataSize
- c_longsize
- c_long_doublesize
- cppExceptions

I don't think any of this knowledge is exposed by any other means, except for mirroring (keeping in sync) version logic between compiler implementation and the library.

Food for thought at least, doesn't have to be done immediately unless there's consensus on it.

ZombineDev:
@ibuclaw 

> * classinfosize

We already have https://dlang.org/spec/traits.html#classInstanceSize. The number one missing feature related to classes IMO is the analog of structs' `T.init` (could be exposed as `__traits(classInstanceInit, T)`). This will allow us to replace the current TypeInfo emitted by the compiler by a purely template-driven implementation.

> * c_longsize
> * c_long_doublesize

Why would you use this, when you can already do `c_long.sizeof` and `c_long_double.sizeof`?

> * maxStaticDataSize
> * cppExceptions

Yes.

ibuclaw:
> Why would you use this, when you can already do c_long.sizeof and c_long_double.sizeof?

Have you seen the implementation? The library logic is just mirroring the compiler. It's a frail implementation.  (note, this is not true for gdc __builtin_clong).

ibuclaw:
> We already have

DMD will error if typeinfo size is wrong. People who write their own druntime are at mercy of the compiler. ;-)

ZombineDev:
Ah, you mean the size of `ClassInfo` instance? Yeah that would be useful, though it would be better if we remove the compiler's hard dependency on TypeInfo all together in future.

ibuclaw:
> The library logic is just mirroring the compiler. It's a frail implementation

What I mean is that `c_long.sizeof` is (for dmd) a value that comes from the library, not the compiler. It's not that difficult to get the two out of sync with each other.

With a trait, the library side will look something like:
```
static if (__traits(getTargetInfo, cLongSize) == 4) {
  alias c_long = int;
  alias c_ulong = uint;
} else static if (... = 8) {
  ... = long;
 ... = ulong;
} else
  static assert (0, "c long size neither 4 nor 8");
```

jacob-carlborg:
> What I mean is that c_long.sizeof is (for dmd) a value that comes from the library, not the compiler. It's not that difficult to get the two out of sync with each other.
With a trait, the library side will look something like:

Isn’t all the current complexities due to the need to get the correct mangling for C++ interoperability? Not sure if a trait like `cLongSize` would help with that.

ibuclaw:
> Isn’t all the current complexities due to the need to get the correct mangling for C++ interoperability?

No, that is a solved problem with cpp_long.

jacob-carlborg:
Right, forgot it’s a separate type for that.

ibuclaw (changes requested):
Coverage and I'll merge!

TurkeyMan:
spec: https://github.com/dlang/dlang.org/pull/2473

test/compilable/traits.d
```diff
@@ -13,8 +13,16 @@ class C19152
 }
 
 static assert(is(typeof(__traits(getTargetInfo, "cppRuntimeLibrary")) == string));
-
 version (CppRuntime_Microsoft)
 {
     static assert(__traits(getTargetInfo, "cppRuntimeLibrary") == "libcmt");
 }
+
+static assert(__traits(getTargetInfo, "floatAbi") == "hard");
```
ibuclaw: version (D_HardFloat)

TurkeyMan: There is no DMD build where this is not true.

ibuclaw: How do you think gdc and ldc test their compilers?

TurkeyMan: Really? I'd be astonished if the DMD test suite ran on all platforms/targets unmodified O_o

ZombineDev: LDC and GDC each use different build systems, but apart from that they all use dmd, druntime and phobos's test suites, in addition to their own test suites. Very often you can see PRs from GDC or LDC to upstream fixing bugs in some tests, exactly because they've run the tests on platforms that dmd doesn't support, where the assumptions doesn't hold, or because the test was relying on undefined behavior which got stressed by their backend optimizers in a different way.

ibuclaw: The less modifications I have to maintain, the better. Ideally the D2 testsuite would just be copied verbatim.

After all, it should be representative of what a *conforming* D compiler should be. :-)

ibuclaw (approved):

test/compilable/traits.d
```diff
@@ -13,8 +13,17 @@ class C19152
...
 
 static assert(is(typeof(__traits(getTargetInfo, "cppRuntimeLibrary")) == string));
-
 version (CppRuntime_Microsoft)
 {
     static assert(__traits(getTargetInfo, "cppRuntimeLibrary") == "libcmt");
 }
+
+version (D_HardFloat)
+    static assert(__traits(getTargetInfo, "floatAbi") == "hard");
```
klickverbot: This seems wrong; the ARM `softfp` targets have floating point instruction support (`D_HardFloat`), but the ABI doesn't use the FP registers.

(Also, Phobos style generally seems to be to capitalize abbreviations, FWIW, even though I personally prefer strict camel case.)

ibuclaw: You seem to be getting a habit here of commenting on things after merge.

Try commenting before.

TurkeyMan: Why would `D_HardFloat` be set for `softfp`... should this test just limit to windows maybe?

ibuclaw: `D_HardFloat` is set if the compiler will use the hardware for float operations.

For the distinction, I'll just quote the manual.

https://gcc.gnu.org/onlinedocs/gcc/ARM-Options.html

> Specifying ‘soft’ causes GCC to generate output containing library calls for floating-point operations. ‘softfp’ allows the generation of code using hardware floating-point instructions, but still uses the soft-float calling conventions. ‘hard’ allows generation of floating-point instructions and uses FPU-specific calling conventions.

TurkeyMan: I understand the options... I'm just surprised by the version.

PR #8729, opened 2018-09-22
[stable] Fix tests for proper core.stdc.stdint aliases

kinke:
Hardcoded mangles as ground truth are a bad idea - cumbersome due to required `version()`ing and prone to errors. Instead, test by actually linking C++ and D objects; this would have caught the wrong 32-bit aliases for 32-bit Windows/MSVC instead of blocking the druntime fix (https://github.com/dlang/druntime/pull/2302), besides requiring only 22 instead of 262 test lines.


kinke:
Turns out some `fast` aliases are also wrong for some platforms:

* Win64: `[u]int_fast8_t` (byte, not char)
* FreeBSD32: `[u]int_fast8_t` should be 32 bits
* FreeBSD64: `[u]int_fast{8,16,32}_t` should be 32 bits

jacob-carlborg (commented):
Do we need to check that the correct function is called?

kinke:
They aren't overloaded, so a proper mangling check by linking should suffice IMO.

kinke:
@klickverbot: Thanks for merging the first druntime PR. I've temporarily disabled some tests here, so that this can be merged, followed by https://github.com/dlang/druntime/pull/2307 and finally a revert of the last commit here.

WalterBright:
> Hardcoded mangles as ground truth are a bad idea

I did them for a reason:
1. C++ mangles rarely, if ever, change.
2. If it doesn't match the hardcoded string, debugging is straightforward
3. If it doesn't match the C++ code, it'll be a while before you figure out what the string is supposed to be
4. It can't be that hard to maintain it - it's just adding things
5. It makes for a handy reference as to what mangles to what

If you want to add the C++ code tests as a separate test file, ok, but I object to deleting the hardcoded ones.

kinke:
LOL!

test/runnable/extra-files/externmangle.cpp
```diff
@@ -215,6 +215,13 @@ size_t testCppSizeTMangle(ptrdiff_t a, size_t b)
     return a + b;
 }
 
+void testCppI8Mangle (int8_t,  uint8_t,  int_least8_t,  uint_least8_t,  int_fast8_t,  uint_fast8_t)  {}
+void testCppI16Mangle(int16_t, uint16_t, int_least16_t, uint_least16_t, int_fast16_t, uint_fast16_t) {}
+void testCppI32Mangle(int32_t, uint32_t, int_least32_t, uint_least32_t, int_fast32_t, uint_fast32_t) {}
+void testCppI64Mangle(int64_t, uint64_t, int_least64_t, uint_least64_t, int_fast64_t, uint_fast64_t) {}
+void testCppIntPtrMangle(intptr_t, uintptr_t) {}
+void testCppIntMaxMangle(intmax_t, uintmax_t) {}
```
Geod24: It would be nice to make them return a unique value, so that we can be sure the right function is actually called.

Geod24:
> If you want to add the C++ code tests as a separate test file, ok, but I object to deleting the hardcoded ones.

@kinke that looks like the path of least resistance, doesn't it ?

test/runnable/extra-files/externmangle.cpp
kinke: I can live with that.

kinke:
I refuse to keep/fixup the old problematic tests for the aforementioned reasons, and yes, I do have a strong opinion about it [and quite frankly thought that the later encountered problems on Win64 and FreeBSD + the need for 3 PRs in total would more than clearly show just how bad of an idea hardcoded mangles are].

Wrt. extracting the new tests into a separate pair of C++/D files, that's doable, but I don't see why it has to be now; `externmangle.{d,cpp}` is the obvious choice, and the new tests fit right into the other C++ `long`, `long long`, `size_t` and `ptrdiff_t` mangling tests in that file.

kinke:
Test functions now returning a unique value & tests extracted into separate pair of .cpp/.d files.

PetarKirov (approved):
LGTM, nice work!

PR #8738, opened 2018-09-24
Fix Issue 19246 - Binary literal `0b_` allowed

wolframw:
... not anymore!
`0b_` and `0B_` will now trigger a deprecation warning.

### DIFF
--- a/changelog/deprecated_binary_literals.dd
+++ b/changelog/deprecated_binary_literals.dd
@@ -1,0 +1,9 @@
+Deprecate invalid binary literals
+
+Prior to this release, binary literals without any digits after the prefix `0b`
+were considered valid. This has now been deprecated.
+---
+auto foo = 0b;   // deprecated
+auto bar = 0b_;  // deprecated
+auto baz = 0b0;  // conforming equivalent
+---
--- a/test/fail_compilation/fix19246.d
+++ b/test/fail_compilation/fix19246.d
@@ -2,9 +2,10 @@
  * PERMUTE_ARGS:
  * TEST_OUTPUT:
 ---
-fail_compilation/fix19246.d(15): Deprecation: `0b_` isn't a valid integer literal, use `0b0` instead
-fail_compilation/fix19246.d(16): Deprecation: `0B_` isn't a valid integer literal, use `0B0` instead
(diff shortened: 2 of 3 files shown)


jacob-carlborg (changes requested):
Since this is a deprecation it should have a changelog describing what and why has been deprecated and what to do to keep the existing behavior without getting the deprecation message.

wolframw:
Added the changelog now, even though in the other PR which deprecated invalid hex literals (see https://github.com/dlang/dmd/pull/8396) no changelog was submitted.

I also opened a companion PR to update the spec accordingly, see https://github.com/dlang/dlang.org/pull/2475.

changelog/deprecated_binary_literals.dd
```diff
@@ -0,0 +1,9 @@
+Deprecate invalid binary literals
+
+Prior to this release, binary literals without any digits after the prefix `0b`
+were considered valid. This has now been deprecated.
+---
+auto foo = 0b;   // deprecated
```
jacob-carlborg: I don’t see any tests for this case.

jacob-carlborg (dismissed):

thewilsonator:
@jacob-carlborg your requested changes were made, semaphore OOM'ed. Merge this?

test/fail_compilation/fix19246.d
```diff
@@ -0,0 +1,18 @@
...
+---
+ */
+
+// https://issues.dlang.org/show_bug.cgi?id=19246
+
+void foo()
+{
+    auto a = 0b_;
+    auto b = 0B_;
+    auto c = 0b;
```
jacob-carlborg: Since both `0b_` and `0B_` are tested, should we test `0B` as well?

ZombineDev: Ping @wolframw we this fixed, this PR should be good to go.

thewilsonator: Added.

thewilsonator (approved):

jacob-carlborg (approved):

schveiguy:
This needs a grammar change, is there one in the works? The original reason why this case was missed is because the grammar allowed it.

thewilsonator:
@schveiguy not at the moment, if you write one I'll merge it.

PR #8739, opened 2018-09-24
dmd.mtype: Move modifiersApply and attributesApply to free functions

ibuclaw:
Remove `extern(C++)` linkage, and replace `const(char)*` with `string`

### DIFF
--- a/src/dmd/mtype.d
+++ b/src/dmd/mtype.d
@@ -159,60 +159,54 @@ MOD MODmerge(MOD mod1, MOD mod2) pure nothrow @nogc @safe
  */
 void MODtoBuffer(OutBuffer* buf, MOD mod) nothrow
 {
-    switch (mod)
-    {
-    case 0:
-        break;
-
-    case MODFlags.immutable_:
-        buf.writestring(Token.toString(TOK.immutable_));
-        break;
-
-    case MODFlags.shared_:
-        buf.writestring(Token.toString(TOK.shared_));
-        break;
-
-    case MODFlags.shared_ | MODFlags.const_:
-        buf.writestring(Token.toString(TOK.shared_));
-        buf.writeByte(' ');
-        goto case; /+ fall through +/
-    case MODFlags.const_:
-        buf.writestring(Token.toString(TOK.const_));
-        break;
-
-    case MODFlags.shared_ | MODFlags.wild:
-        buf.writestring(Token.toString(TOK.shared_));
-        buf.writeByte(' ');
-        goto case; /+ fall through +/
-    case MODFlags.wild:
-        buf.writestring(Token.toString(TOK.inout_));
-        break;
-
-    case MODFlags.shared_ | MODFlags.wildconst:
-        buf.writestring(Token.toString(TOK.shared_));
-        buf.writeByte(' ');
-        goto case; /+ fall through +/
-    case MODFlags.wildconst:
-        buf.writestring(Token.toString(TOK.inout_));
-        buf.writeByte(' ');
-        buf.writestring(Token.toString(TOK.const_));
-        break;
-
-    default:
-        assert(0);
-    }
+    buf.writestring(MODtoString(mod));
 }
 
 /*********************************
- * Return modifier name.
+ * Returns:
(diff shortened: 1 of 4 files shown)


src/dmd/mtype.d
```diff
@@ -6521,3 +6459,80 @@ const(char*)[2] toAutoQualChars(Type t1, Type t2)
     }
     return [s1, s2];
 }
+
+
+/**
+ * For each active modifier (MODFlags.const_, MODFlags.immutable_, etc) call `fp` with a
+ * void* for the work param and a string representation of the attribute.
+ */
+final int modifiersApply(TypeFunction tf, void* param, int function(void*, const(char)[]) fp)
```
jacob-carlborg: * Look at this on my phone, if this is a free function it should not be `final`
* Missing standard Ddoc sections

src/dmd/mtype.d
```diff
@@ -6521,3 +6459,80 @@ const(char*)[2] toAutoQualChars(Type t1, Type t2)
...
+        }
+    }
+
+    return 0;
+}
+
+/**
+ * For each active attribute (ref/const/nogc/etc) call `fp` with a void* for the
+ * work param and a string representation of the attribute.
+ */
```
jacob-carlborg: Same here, missing standard Ddoc sections.

RazvanN7 (approved):

src/dmd/mtype.d
```diff
@@ -6521,3 +6459,80 @@ const(char*)[2] toAutoQualChars(Type t1, Type t2)
...
+int modifiersApply(TypeFunction tf, void* param, int function(void*, const(char)[]) fp)
+{
+    immutable ubyte[4] modsArr = [MODFlags.const_, MODFlags.immutable_, MODFlags.wild, MODFlags.shared_];
+
+    foreach (modsarr; modsArr)
+    {
+        if (tf.mod & modsarr)
+        {
+            if (int res = fp(param, MODtoString(modsarr)))
```
WalterBright: This is a poor solution (even though what you're replacing did the same thing) in that it is pointlessly allocating memory to make a copy of a string literal, and leaking that memory.

src/dmd/mtype.d
```diff
@@ -214,6 +214,16 @@ char* MODtoChars(MOD mod) nothrow
...
 }
 
+/// Ditto
+const(char)[] MODtoString(MOD mod) nothrow
+{
+    OutBuffer buf;
+    buf.reserve(16);
+    MODtoBuffer(&buf, mod);
+    size_t len = buf.offset;
+    return buf.extractData()[0 .. len];
```
WalterBright: This implementation should be swapped with the implementation of ModtoBuffer, then there will be no need to make the copy.

WalterBright (changes requested):
Might as well fix the underlying issue it has with leaking memory.

src/dmd/mtype.d
ibuclaw: Call ModToBuffer from here? OK, seems reasonable.

src/dmd/mtype.d
```diff
@@ -6521,3 +6449,83 @@ const(char*)[2] toAutoQualChars(Type t1, Type t2)
...
+    immutable ubyte[4] modsArr = [MODFlags.const_, MODFlags.immutable_, MODFlags.wild, MODFlags.shared_];
+
+    foreach (modsarr; modsArr)
+    {
+        if (tf.mod & modsarr)
+        {
+            OutBuffer buf;
+            MODtoBuffer(&buf, modsarr);
+            const len = buf.offset;
+            if (int res = fp(param, buf.extractData()[0 .. len]))
```
WalterBright: It still leaks memory, once each time through the loop. The strings produced by MODtoString are all literals and don't need to be duplicated.

WalterBright: I mean MODtoBuffer.

ibuclaw:
MODtoString returns a constant.

ibuclaw:
Ping?

RazvanN7:
ping @WalterBright

WalterBright (approved):

PR #8740, opened 2018-09-25
Fix issue 3507: with -deps, emit depsUnused line on unused imports

FeepingCreature:
The purpose of this PR is to enable tools to warn on unused imports in user code.

To this end, a new line is added to the `-deps` output: `depsUnused`, of the format `depsUnused ::= ModuleFullyQualifiedName " (" FilePath "(" LineNumber ")) : " ModuleFullyQualifiedName`, listing first the module in which the import is located, and second the import that was not used.

Important note: all that this line indicates is that the import can be removed _given the current compiler invocation_. Imports that are used in code that is statically disabled, or that are used in templates that are never instantiated, or that are used in modules not listed on the commandline, will be listed as unused. This is unavoidable. Due to the Turing-complete nature of D metaprogramming, it is impossible to prove for an arbitrary program that an import is never used. All that can be shown is that an import was not used during compilation. However, imports that are themselves hidden in `template`s, `version`s or `static if`s will not be output.

### DIFF
--- a/src/dmd/arraytypes.h
+++ b/src/dmd/arraytypes.h
@@ -64,3 +64,5 @@ typedef Array<class GotoStatement *> GotoStatements;
 typedef Array<class TemplateInstance *> TemplateInstances;
 
 typedef Array<struct Ensure> Ensures;
+
+typedef Array<struct ImportInfo> Imports;
--- a/src/dmd/dimport.d
+++ b/src/dmd/dimport.d
@@ -223,7 +223,7 @@ extern (C++) final class Import : Dsymbol
                     protection = sc.protection;
                 if (!isstatic && !aliasId && !names.dim)
                 {
-                    sc.scopesym.importScope(mod, protection);
+                    sc.scopesym.importScope(loc, mod, protection);
                 }
             }
         }
--- a/src/dmd/dsymbol.d
+++ b/src/dmd/dsymbol.d
@@ -1241,11 +1241,19 @@ extern (C++) class ScopeDsymbol : Dsymbol
     uint endlinnum;             // the linnumber of the statement after the scope (0 if unknown)
 
 private:
-    /// symbols whose members have been imported, i.e. imported modules and template mixins
-    Dsymbols* importedScopes;
-    Prot.Kind* prots;            // array of Prot.Kind, one for each import
+    import dmd.root.array : Array, BitArray;
 
-    import dmd.root.array : BitArray;
+    /// symbols whose members have been imported, i.e. imported modules and template mixins
+    struct ImportInfo
+    {
+        Dsymbol symbol;
+        Prot.Kind prot;
+        Loc loc;
+        bool used;
+    }
+
+    Array!ImportInfo imports;
+
     BitArray accessiblePackages, privateAccessiblePackages;// whitelists of accessible (imported) packages
 
 public:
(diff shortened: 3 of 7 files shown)


wilzbach:
I thought most tools use the Json output? Or DMD directly as a library?

FeepingCreature:
The Json output has imports? I hadn't checked that. That'd possibly be a better place.

A problem might be that it's kind of important that this info is output after all modules have finished processing recursively, so that templates instantiated in other modules that use global template imports properly mark those imports as used. If the json info is file-by-file, it might go wrong. I'll try this tomorrow.

nordlow:
Nice work. I've been thinking about adding this for a while myself. How difficult would it be to this to include other unused symbols

wilzbach:
The Json output is one file only (at the moment).

FeepingCreature:
It shouldn't be unreliable in the case that there's a main program that is passed to dmd. Any import that is reported as not used should be one that actually isn't used in that program. For code hidden behind `static-if`s and `versions`, reporting that the import is not used in that build is _correct_ behavior, since the import arguably belongs inside the version case.

jacob-carlborg:
I recommend looking into using DMD as a library.

FeepingCreature:
I don't see how that would be possible as it stands, since DMD does not seem to expose a way to track identifier lookups. If there's a change to DMD needed anyway, might as well export the information from DMD directly, no?

jacob-carlborg:
> If there's a change to DMD needed anyway, might as well export the information from DMD directly, no?

Not necessarily.

FeepingCreature:
> Not necessarily. I've made a PR to add additional information that the compiler currently doesn't use but is useful when using DMD as a library and could be used by the compiler in the future.

I mean ... I don't want to use this information at some indeterminate point in the future, I want to use this information as soon as possible to integrate it into our tooling.

If 2.083 includes "used-or-not" information on imports, will I be able to then create a tool linked against DMD as a library, that exports this information?

jacob-carlborg:
> I mean ... I don't want to use this information at some indeterminate point in the future, I want to use this information as soon as possible to integrate it into our tooling.

What I meant is that data would be available immediately in the library but the compiler wouldn't use it (because it has no need yet). But a tool that are using the library can use the data.

> If 2.083.0 includes "used-or-not" information on imports, will I be able to then create a tool linked against DMD 2.083.0 as a library, that exports this information?

Yes.

> Is this mode of use documented anywhere? make -f posix.mak does not seem to create a library.

The DMD library is available today as a Dub package [1]. You can see how I'm using it in my tool [2].

[1] http://code.dlang.org/packages/dmd
[2] https://github.com/jacob-carlborg/dlp/blob/master/source/dlp/visitors/leaf_functions.d

FeepingCreature:
> Yes.

Awesome.

> [1] http://code.dlang.org/packages/dmd

Is it me or does that package completely lack tagged releases?

test.d
```diff
@@ -0,0 +1,7 @@
+module test;
```
RazvanN7: why is this test located at top-level?

jacob-carlborg:
> Pull request stands as is, due to dub dmd being unviable.

I've asked if tags can be created for each release for the Dub package. As long as the first digit of the version is 0, the normal semantic versioning rules don't apply: "Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.".

WalterBright:
> For code hidden behind static-ifs and versions,

And uninstantiated templates.

> reporting that the import is not used in that build is correct behavior,

That depends on what the point of these reports is. An obvious use is to remove the import declarations for unused imports, except then a template gets instantiated and it won't compile anymore. Given the heavy use of templates, is this feature useful?

FeepingCreature:
> That depends on what the point of these reports is. An obvious use is to remove the import declarations for unused imports, except then a template gets instantiated and it won't compile anymore. Given the heavy use of templates, is this feature useful?

Given the point of this feature is to allow the removal of dead code (imports), a template in a program that is never instantiated is itself a problem, since it's just as dead. In that case, the proper recourse would be to move the imports into the template, because that's where they semantically belong anyways.

_:shrugs:_ We'd certainly find it useful, that's why I added it. I can't say to what extent other people would find it useful, though there's been several people on the ML saying they'd welcome this functionality. In any case, the cost it imposes seems small. It doesn't take up much code, and if you ever decide to remove the functionality, all that happens is that some diagnostics stop displaying - it won't even break any builds.

WalterBright:
> since it's just as dead

If you do separate compilation, it's not necessarily dead.

> the cost it imposes seems small

There is a credibility cost in advertising a feature that does not work reliably and has no path to making it reliable.

> unittest [...] problematic lack of coverage

We already have a very nice (and far more comprehensive) feature for auditing coverage of unittests.

FeepingCreature:
> We already have a very nice (and far more comprehensive) feature for auditing coverage of unittests.

Yes, which we use, which is why I don't think it's a problem.

> If you do separate compilation, it's not necessarily dead.

So ... we can just not do that.

Like, you seem intent to construct a scenario where this feature is not useful. Of course there's scenarios where it's not useful. I think there's scenarios where it's useful, that's why I made it.

FeepingCreature:
Brief Update: This PR is **not working correctly** at the moment. In cases where a module imports another module that imports modules that are used in a template instantiated from the first module, the output is sensitive to the order in which modules are specified on the command line. Amusingly, `depsImport` seems like it gives up entirely and lists the third module as imported from the location where the template is _used_, not where it's _declared_. So there'll probably be no way but to wait until all the listed object files have been processed, then walking them and checking for unused imports.

FeepingCreature:
Closed due to (my own) lack of activity.

PR #8741, opened 2018-09-26
fix issue 19266 - Some source files names are no longer accepted

rainers:
only add the \\?\ prefix if the filename is actually long and does not start with \\ already.

### DIFF
--- a/src/dmd/root/filename.d
+++ b/src/dmd/root/filename.d
@@ -836,20 +836,24 @@ version(Windows)
             // path.
             static immutable prefix = `\\?\`w;
 
-            // +1 for the null terminator
-            const bufferLength = pathLength + prefix.length + 1;
+            // prefix only needed for long names and non-UNC names
+            const needsPrefix = pathLength >= MAX_PATH && (wpath[0] != '\\' || wpath[1] != '\\');
+            const prefixLength = needsPrefix ? prefix.length : 0;
 
-            wchar[1024] absBuf;
+            // +1 for the null terminator
+            const bufferLength = pathLength + prefixLength + 1;
+
+            wchar[1024] absBuf = void;
             auto absPath = bufferLength > absBuf.length ? new wchar[bufferLength] : absBuf[];
 
-            absPath[0 .. prefix.length] = prefix[];
+            absPath[0 .. prefixLength] = prefix[0 .. prefixLength];
 
             const absPathRet = GetFullPathNameW(wpath,
-                                                cast(uint)(absPath.length - prefix.length),
-                                                &absPath[prefix.length],
+                                                cast(uint)(absPath.length - prefixLength),
+                                                &absPath[prefixLength],
                                                 null /*filePartBuffer*/);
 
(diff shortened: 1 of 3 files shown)


Geod24:
Do you have the PR that introduced this ?

src/dmd/root/filename.d
```diff
@@ -836,20 +836,24 @@ version(Windows)
             // path.
             static immutable prefix = `\\?\`w;
 
+            // prefix only needed for long names and non-UNC names
+            bool needsPrefix = pathLength >= MAX_PATH && (wpath[0] != '\\' || wpath[1] != '\\');
```
jacob-carlborg: can be `const`.

src/dmd/root/filename.d
```diff
@@ -836,20 +836,24 @@ version(Windows)
...
+            size_t prefixlen = needsPrefix ? prefix.length : 0;
```
jacob-carlborg: * `const`
* I prefer to not use abbreviations and to use camel case names

jacob-carlborg (changes requested):
Should also have a test if possible.

rainers:
> Do you have the PR that introduced this ?

https://github.com/dlang/dmd/pull/7299

rainers:
> Should also have a test if possible.

Not so easy as it requires either a network share or an absolute path to be passed with `\\?\`, but I'll try to add a shell script that does the latter.

rainers:
Added test and changed code according to comments.

test/compilable/extra-files/test19266.d
```diff
@@ -0,0 +1,3 @@
+module test19266;
+
+pragma(msg, __FILE__);
```
jacob-carlborg: Is this doing anything for the test?

jacob-carlborg (approved):

test/compilable/extra-files/test19266.d
rainers: No, but you can look it up in the test_result folder. If `\\?\` is given on the command line it is also found in `__FILE__`.

src/dmd/root/filename.d
```diff
@@ -836,20 +836,24 @@ version(Windows)
...
 
+            // prefix only needed for long names and non-UNC names
+            const needsPrefix = pathLength >= MAX_PATH && (wpath[0] != '\\' || wpath[1] != '\\');
+            const prefixLength = needsPrefix ? prefix.length : 0;
+
             // +1 for the null terminator
-            const bufferLength = pathLength + prefix.length + 1;
+            const bufferLength = pathLength + prefixLength + 1;
 
             wchar[1024] absBuf;
```
WalterBright: Do an `= void` initialization.

rainers: Done.

Geod24:
The AppVeyor failure is due to stdint and unrelated to this. #8729 will fix it.
I checked that the added tests pass, so going to merge this manually.

PR #8743, opened 2018-09-26
 Fix issue 19269: Cannot throw C++ exceptions from D

Geod24:
This allow D functions to throw C++ exceptions.

However, there is a big caveat to this: Since the `std::exception` definition is in druntime (`core.stdcpp.exception`), and throwing an exception pulls in the classinfo, it ties the druntime build to the C++ lib it was build against, which is something we should avoid (CC @TurkeyMan ).

I'm not so familiar with TypeInfo and what pulls it in, so I don't know for which reason it's pulled in here, if we can do away with it, or if we'll end up needing it anyway and we're just lucky that it worked without it so far.

As a result, this currently fails to link. Applying the following diff to druntime makes it link:
```diff
diff --git a/mak/SRCS b/mak/SRCS
index f998eb29..757d9d24 100644
--- a/mak/SRCS
+++ b/mak/SRCS
@@ -46,6 +46,8 @@ SRCS=\
        src\core\stdc\time.d \
        src\core\stdc\wchar_.d \
        \
+       src\core\stdcpp\exception.d \
+       \
        src\core\sync\barrier.d \
        src\core\sync\condition.d \
        src\core\sync\config.d \
```

Note that this wouldn't be a problem if those bindings were not in Druntime.

### DIFF
--- a/src/dmd/dmodule.d
+++ b/src/dmd/dmodule.d
@@ -1279,11 +1279,28 @@ extern (C++) final class Module : Package
         return this.importedFrom == this;
     }
 
-    // true if the module source file is directly
-    // listed in command line.
-    bool isCoreModule(Identifier ident)
+    /**
+     * Compare a provided list of identifier with this module
+     *
+     * Params:
+     *   idents = Identifiers to compare to, should be provided
+     *            in the "natural" order.
+     *            E.g. if this module is `core.attribute`,
+     *            use `attrmodule.isModuleName(Id.core, Id.attribute)`
+     *
+     * Returns:
+     *   `true` if this module correspond to the provided identifiers
+     */
+    extern(D) final bool isModuleName(scope Identifier[] idents...)
     {
-        return this.ident == ident && parent && parent.ident == Id.core && !parent.parent;
+        Dsymbol sym = this;
+        foreach_reverse (id; idents)
+        {
+            if (!sym || id != sym.ident)
+                return false;
+            sym = sym.parent;
+        }
+        return sym is null;
     }
 
     // Back end
--- a/src/dmd/id.d
+++ b/src/dmd/id.d
@@ -97,6 +97,7 @@ immutable Msgtable[] msgtable =
     { "typeinfo" },
     { "outer" },
     { "Exception" },
+    { "cppException", "exception" },
     { "RTInfo" },
     { "Throwable" },
     { "Error" },
@@ -333,11 +334,12 @@ immutable Msgtable[] msgtable =
     // varargs implementation
     { "va_start" },
 
-    // Builtin functions
+    // Builtin functions / modules
     { "std" },
     { "core" },
     { "etc" },
     { "attribute" },
+    { "stdcpp" },
     { "math" },
     { "sin" },
     { "cos" },
--- a/src/dmd/module.h
+++ b/src/dmd/module.h
@@ -136,9 +136,6 @@ public:
     int imports(Module *m);
 
     bool isRoot() { return this->importedFrom == this; }
-    // true if the module source file is directly
-    // listed in command line.
-    bool isCoreModule(Identifier *ident);
 
     // Back end
 
--- a/src/dmd/objc.d
+++ b/src/dmd/objc.d
@@ -468,7 +468,7 @@ extern(C++) private final class Supported : Objc
         if (sd.ident != Id.udaSelector || !sd.parent)
             return false;
         Module _module = sd.parent.isModule();
-        return _module && _module.isCoreModule(Id.attribute);
+        return _module && _module.isModuleName(Id.core, Id.attribute);
     }
 }
 
--- a/src/dmd/statementsem.d
+++ b/src/dmd/statementsem.d
@@ -3936,12 +3936,12 @@ else
         checkThrowEscape(sc, ts.exp, false);
 
         ClassDeclaration cd = ts.exp.type.toBasetype().isClassHandle();
-        if (!cd || ((cd != ClassDeclaration.throwable) && !ClassDeclaration.throwable.isBaseOf(cd, null)))
+
+        if (!cd.canBeThrown)
         {
-            ts.error("can only throw class objects derived from `Throwable`, not type `%s`", ts.exp.type.toChars());
+            ts.error("can only throw class objects derived from `Throwable` or `core.stdcpp.exception.exception`, not type `%s`", ts.exp.type.toChars());
             return setError();
         }
-
         result = ts;
     }
 
@@ -4272,4 +4272,38 @@ TupleForeachRet!(isStatic, isDecl) makeTupleForeach(bool isStatic, bool isDecl)(
     {
         return v.makeTupleForeach!(isStatic, isDecl)(fs, args);
     }
+}
+
+/**
+   Check if this `ClassDeclaration` derives from either
+   `core.stdcpp.exception : exception` (aka `std::exception`)
+   or `object : Exception`
+
+   Note:
+     For `std::exception`, the check is actually "is the base class
+     defined in `core.stdcpp.exception`.
+     `std::exception`'s ClassDeclaration cannot just be a global property
+     like `object : Exception` is because the mechanism which
+     is used for the later works at parsing time, and so relies on those
+     definitions not being versioned (or within `static if`).
+
+   Returns:
+     `true` if this type can be `throw`n and caught
+*/
+private bool canBeThrown (ClassDeclaration cd)
+{
+    if (!cd)
+        return false;
+    if (cd == ClassDeclaration.throwable)
+        return true;
+    if (ClassDeclaration.throwable.isBaseOf(cd, null))
+        return true;
+    if (!cd.isCPPclass)
+        return false;
+    // Get the root class (C++ class don't derive from `Object`)
+    while (cd.baseClass !is null)
+        cd = cd.baseClass;
+    // The module defining the root class should be `core.stdcpp.exception`
+    auto mod = cd.getModule();
+    return mod.isModuleName(Id.core, Id.stdcpp, Id.cppException);
 }
--- a/test/fail_compilation/ice10651.d
+++ b/test/fail_compilation/ice10651.d
@@ -1,7 +1,7 @@
 /*
 TEST_OUTPUT:
 ---
-fail_compilation/ice10651.d(11): Error: can only throw class objects derived from `Throwable`, not type `int*`
+fail_compilation/ice10651.d(11): Error: can only throw class objects derived from `Throwable` or `core.stdcpp.exception.exception`, not type `int*`
 ---
 */
 
--- a/test/runnable/cppa.d
+++ b/test/runnable/cppa.d
@@ -6,6 +6,8 @@ import core.stdc.stdarg;
 import core.stdc.config;
 import core.stdc.stdint;
 
+import core.stdcpp.exception;
+
 extern (C++)
         int foob(int i, int j, int k);
 
@@ -500,8 +502,6 @@ extern (C++, std)
         }
     }
 
-    class exception { }
-
     // 14956
     extern(C++, N14956)
     {
@@ -982,7 +982,7 @@ void testeh()
             {
                 throwit();
             }
-            catch (std.exception e)
+            catch (exception e)
             {
                 caught = true;
             }
@@ -1019,7 +1019,7 @@ version (linux)
             {
                 dFunction();
             }
-            catch(std.exception e)
+            catch(exception e)
             {
                 assert(raii_works);
             }
@@ -1047,7 +1047,7 @@ void testeh3()
             {
                throwle();
             }
-            catch (std.exception e)  //polymorphism test.
+            catch (exception e)  //polymorphism test.
             {
                 caught = true;
             }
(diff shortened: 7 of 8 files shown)


TurkeyMan:
> it ties the druntime build to the C++ lib it was build against

We should certainly avoid this. Can you explain how we tie druntime to the C++ lib it was built against?
I'm not sure what you mean about typeinfo; C++ typeinfo or D typeinfo?

I've avoided c++ runtime affinity so far with the other C++ types, they're header-only material, and no hard-functions are C++ runtime dependent. I'm trying to make sure it's the client code that imports the cpp material that becomes locked to the C++ runtime that the client builds against. druntime should not assert the affinity.

Geod24:
> I've avoided c++ runtime affinity so far [...]

I completely agree with your statement and your goal here. I don't think having this affinity burned into druntime would be a good design.

> I'm not sure what you mean about typeinfo; C++ typeinfo or D typeinfo?

D typeinfo

> Can you explain how we tie druntime to the C++ lib it was built against?

Sure thing. So I originally added the frontend code, and got linker errors.
After a bit of debugging I ended up with the following call stack (don't mind the broken file/line, just look at the functions):
```
dmd/access.d:623 toSymbol::ToSymbol::visit(ClassDeclaration*) [0x4aafd09]
dmd/access.d:623 ClassDeclaration::accept(Visitor*) [0x496a365]
dmd/access.d:623 toSymbol(Dsymbol*) [0x4aaf1f1]
dmd/access.d:623 toElem::ToElemVisitor::visit(NewExp*) [0x4abdcdf]
dmd/access.d:623 NewExp::accept(Visitor*) [0x49cd73d]
dmd/access.d:623 toElem(Expression*, IRState*) [0x4abc390]
dmd/access.d:623 toElemDtor(Expression*, IRState*) [0x4acb030]
dmd/access.d:623 S2irVisitor::visit(ThrowStatement*) [0x4ab46b6]
dmd/access.d:623 ThrowStatement::accept(Visitor*) [0x4a70b35]
dmd/access.d:623 Statement_toIR(Statement*, IRState*) [0x4ab629c]
dmd/access.d:623 S2irVisitor::visit(CompoundStatement*) [0x4ab42a0]
dmd/access.d:623 CompoundStatement::accept(Visitor*) [0x4a6e095]
dmd/access.d:623 Statement_toIR(Statement*, IRState*) [0x4ab629c]
dmd/access.d:623 FuncDeclaration_toObjFile(FuncDeclaration*, bool) [0x4aaa02d]
dmd/access.d:623 toObjFile::ToObjFile::visit(FuncDeclaration*) [0x4ab722c]
dmd/access.d:623 FuncDeclaration::accept(Visitor*) [0x49fd005]
dmd/access.d:623 toObjFile::ToObjFile::visit(AttribDeclaration*) [0x4ab8091]
dmd/access.d:623 ParseTimeVisitor<ASTCodegen>::visit(LinkDeclaration*) [0x4a874c2]
dmd/access.d:623 LinkDeclaration::accept(Visitor*) [0x493dccd]
dmd/access.d:623 toObjFile::ToObjFile::visit(AttribDeclaration*) [0x4ab8091]
dmd/access.d:623 ParseTimeVisitor<ASTCodegen>::visit(LinkDeclaration*) [0x4a874c2]
dmd/access.d:623 LinkDeclaration::accept(Visitor*) [0x493dccd]
dmd/access.d:623 toObjFile(Dsymbol*, bool) [0x4ab718d]
dmd/access.d:623 genObjFile(Module*, bool) [0x4aa86c5]
dmd/access.d:623 int dmd.mars.tryMain(ulong, const(char)**) [0x4a37600]
```

`toSymbol::ToSymbol::visit(ClassDeclaration*)` generates [an extern ref to `__Class`](https://github.com/dlang/dmd/blob/31ed54c208c1b08ee81ca7c592cfc1ccda988c65/src/dmd/tocsym.d#L408-L417) (the `ClassInfo`).

Now the funny part: Does the following code compiles for you ?
```D
import core.stdcpp.exception;
void main () { auto e = new exception; }
```

For me it yields (on Linux, because on OSX the bindings just don't exist):
```
$ dmd -run test.d -Lstdc++
/usr/bin/ld: test.o: in function `_Dmain':
test.d:(.text._Dmain[_Dmain]+0x7): undefined reference to `_D4core6stdcpp9exception3stdQo7__ClassZ'
/usr/bin/ld: test.d:(.text._Dmain[_Dmain]+0x14): undefined reference to `std::exception::exception()'
Error: linker exited with status
```

So the `ClassInfo` does not exists, and neither does the constructor. 
The [constructor is defined as extern](https://github.com/dlang/druntime/blob/5b700c40b427d2a4783ad94b7bff5a5108be7d70/src/core/stdcpp/exception.d#L66) in our binding.
The `ClassInfo` is not present because the sources are not compiled into `libdruntime.a`.
Since it's a non-templated `class`, the compiler assumes the `ClassInfo` is emitted when compiling `core.stdcpp.exception`.

And if we want the `ClassInfo` included in druntime, we need to compile in the sources into druntime, but then the `ClassInfo` will be the one druntime was compiled in.
Maybe we could do some hack with `#pragma(mangle)` and aliases (compiling all classes, but only aliasing one to `exception`).

TurkeyMan:
We should probably have C++ `new` in druntime, so we can alloc memory that C++ is able to free (and vice versa)

thewilsonator:
>Now the funny part: Does the following code compiles for you ?
```d
import core.stdcpp.exception;
void main () { auto e = new exception; }
```

Interestingly it is not just the `____ClassZ` symbol thats missing:

```d
import core.stdcpp.exception;
void main () { scope e = new exception; }
```

fails with missing `__initZ`.

ZombineDev:
@thewilsonator that's also expected as the compiler essentially emits the class initialization on the stack itself.

Geod24:
So all things considered, I'm not sure we should be able to throw C++ exceptions from D.
It's surely more convenient but creating a C++ function that does it is trivial. It can even be done in a lib.

Independent of that, this uncovered some issue which will happen with any classes, not only exceptions.

1) Not compiling those modules leads to missing `ClassInfo` / `init`
2) We don't want to have Druntime and the C++ lib tighly coupled
3) The relationship between druntime and stdc++ is different that the one between druntime and (C) stdlib because C libs only contains global variables and functions, while a C++ lib contains C++'s typeinfo. And while we can trivially call C functions / use global C variables, we need a ClassInfo for classes to fully work in D.

There are possible workarounds for that, but I can not think of a proper fix ATM.

We cannot get rid of the need for `ClassInfo`.
Things like using templates + `pragma(mangle)` are hackish at best and completely unreliable.
Splitting the bindings into another lib are also problematic because the compiler relies on `std::type_info` / `std::exception` to do exception handling.

Any ideas / thoughts / suggestions are welcome.

jacob-carlborg:
> We cannot get rid of the need for ClassInfo.

Why not? I don't think non-D classes should have ClassInfo. It's mostly broken anyway.

TurkeyMan:
What's the use of the ClassInfo for C++ classes? C++ classes need C++ RTTI struct generated...

jacob-carlborg:
> What's the use of the ClassInfo for C++ classes? 

I’m guessing it has never been removed from C++ classes.

TurkeyMan:
How does GC call C++ destructor?

Geod24:
> Why not? I don't think non-D classes should have ClassInfo. It's mostly broken anyway.

Can you elaborate on "mostly broken" ?
Currently I think there are a few things preventing getting rid of it. The worst offender, AFAIK, is `_d_newclass`.

Take the following code:
```D
extern(C++, foo)
class Bar { void func() { assert(0); } }

class Foo { void func() { assert(0); } }

void main ()
{
    auto c = new Bar;
    auto f = new Foo;
}
```

Gives the following (OSX):
```
0000000000000060 s EH_frame0
0000000000000118 s _TMP0
                 U __D14TypeInfo_Class6__vtblZ
0000000000000040 S __D3cpp12__ModuleInfoZ
0000000000000138 S __D3cpp3Foo4funcMFZv
00000000000000a0 S __D3cpp3Foo4funcMFZv.eh
0000000000000220 S __D3cpp3Foo6__initZ
00000000000002d0 S __D3cpp3Foo6__vtblZ
0000000000000230 S __D3cpp3Foo7__ClassZ
0000000000000150 S __D3cpp3foo3Bar6__initZ           // Oops
0000000000000210 S __D3cpp3foo3Bar6__vtblZ
0000000000000160 S __D3cpp3foo3Bar7__ClassZ        // Also oops
                 U __D6Object7__ClassZ
                 U __D6object6Object5opCmpMFCQqZi
                 U __D6object6Object6toHashMFNbNeZm
                 U __D6object6Object8opEqualsMFCQtZb
                 U __D6object6Object8toStringMFZAya
0000000000000000 T __Dmain
                 U __Dmain
00000000000000c8 S __Dmain.eh
0000000000000120 S __ZN3foo3Bar4funcEv
0000000000000078 S __ZN3foo3Bar4funcEv.eh
                 U __d_assertp
                 U __d_newclass
                 U __d_run_main
                 U _main
0000000000000020 T _main
00000000000000f0 S _main.eh
```

I **think** in order to make it work, `_d_newclass` needs to be templated (which @ibuclaw have been arguing for for a long time now). It would also allows it to be inlined, potentially leading to some speedup.

jacob-carlborg:
> Can you elaborate on "mostly broken" ?

For example, I recently tried to use a C++ class as a key in an associate array (by accident). The resulted in some ... interesting behavior. What happened was the the associate array implementation calls `TypeInfo_Class.getHash`, which receives the object as `void*`. It then casts the pointer to `Object` and calls `toHash`. What happened to me, since I had a C++ class, it called a random method at the same position in the vtable as `toHash`. I have a suspicion that other parts of druntime have a similar implementation that doesn't check what kind of class it receives.

I now had a look at the other methods in `TypeInfo_Class`, the following methods have the same or similar problem as `toHash`: `equals`, `compare`, `tsize` and `create`. I bet there are several other places in druntime with the same problem.

jacob-carlborg:
> I think in order to make it work, _d_newclass needs to be templated

What would be nice. Then I might not need to add any special case for `new` to work with Objective-C classes (as I planned to do in the future).

ZombineDev:
The main thing blocking template-izing `_d_newclass` is `__traits(classInstanceInit, T)` (the non-`TypeInfo` based analog of `typeid(T).initializer`).  If we had this, we could build TypeInfo in druntime, instead of emitting it specially in the compiler.

thewilsonator:
> `__traits(classInstanceInit, T)`

we have `__traits(classInstanceSize, T)` already, it is perhaps worth rolling them up into one trait like `__traits(getTargetInfo, "key")`

Are there other fields worth exposing from classInfo statically so that we can eventually move to a "pay for what you use" model?
e.g. 
`__traits(classInstanceInit, T)` -> `__traits(classInstance, T , "init")`
`__traits(classInstanceSize, T)` -> `__traits(classInstance, T , "size")`

Geod24:
@ZombineDev : I have to give it a shot, but the compiler could simply pass the `init` symbol as an `alias` template parameter. Since one type == one `init` symbol it wouldn't result in more template instantiations.

Geod24:
> I don't like the idea of special passing of template parameters as it introduces an unnecessary coupling between dmd and druntime.

I would argue that introducing a new traits increase the coupling at this point, since it's a whole new feature for a single usage.
I tend to follow the [rule of three](https://blog.codinghorror.com/rule-of-three/) whenever I can, and so far the only (non-hypothetic) usage we have is for `_d_newclass`.

So I'm not opposed to the `trait` by itself, I just think if there's a solution which would already work without introducing a new feature we should go for that first.

> preemptively emit all init symbols instead doing this on demand, but I may be wrong about that.

Isn't it already the case ?

ZombineDev:
Actually I didn't even consider using the trait for `_d_newclass`. The main use case for me is using classes in `@nogc` + druntime, `D_BetterC` + libc and pure `FreeStanding` environments. Even for `@nogc` + druntime, in addition to `std.conv.emplace` and `std.experimental.allocator.make` there are probably at least 5-10 dub packages that need this functionality in one form or another.

ZombineDev:
> I would argue that introducing a new traits increase the coupling at this point

How is this introducing more coupling between dmd and druntime? All traits can be used without druntime.

> [..] I just think if there's a solution which would already work without introducing a new feature we should go for that first.

It's going to be a new feature no matter how it ends up implemented, I see no reason to go with a bad one just because it may look smaller (but actually more convoluted IMO).

> Isn't it already the case ?

AFAIK, the compiler emits all init symbols that may or may not be needed for modules that are compiled (not just impprted). I want to address this deficiency with on demand use of the trait.

thewilsonator:
One other idea would be to have `__traits(classInstanceType, T)` which would dereference the class reference so that

```d
class Foo { ... }
alias FooInst = __triats(classInstanceType, T);
enum size = FooInst.sizeof; // == __traits(classInstanceSize, Foo)
enum initial = FooInst.init; // == __traits(classInstanceInit, T)

// even maybe this? equiv to C++: class Foo {}; void (Foo&  f){}
exten(C++) void foo(ref FooInst f) {} 
```

thewilsonator:
Heh, while this will be _really_ simple from a users perspective, I suspect its going to be really _not_ simple from the compilers perspective.

TurkeyMan:
I have what suspicion also... although it shouldn't be *tooo* bad, since obviously the struct does exists, it has a layout, codegens, etc. It must have a hidden definition somewhere, or nothing would work.

thewilsonator:
> almost

Indeed, `const T* const` is still going to be a PiTA

TurkeyMan:
> `const T* const` is still going to be a PiTA

Yes, that's precisely the exception I intended by 'almost' ;)

thewilsonator:
Its not just that, its a new base type, with all the hassle involved with that.

thewilsonator:
> Sorry I don’t understand, what would this trait do?

Given 
```d
class Foo {}
alias FooInst = __triats(classInstanceType, Foo);
```
`Foo f;` declares a _reference_ to a class instance of `Foo`, whereas `FooInst fi;` would declare an instance of `Foo` (rather than a reference to one). Hence `FooInst.sizeof` gives the the size of a class instance of Foo (currently accessible by `__trait(classInstanceSize,Foo)`), `FooInst.init` would give the initialiser of the instance and so on.

Geod24:
I think we should seriously weight any feature we plan to add against its usefulness and how much it breaks from what we have.
Currently `class` are reference types by design, and while it does sometimes feel very limiting (e.g. not being able to have a mutable reference to a `const class`), making the non-ref type available to the user will yield its own set of problem, especially if not limited to C++ classes.

Anyway, this will have to wait:
- Other stuff (e.g. template mangling) are much more broken and blocking
- This isn't so high prio and easy to work around (just write a C++ function to do the job)
- It's blocked by the `TypeInfo` stuff anyway, and I'd like to look into it before committing to a solution.

TurkeyMan:
So, Is this still held up, since my other PR was merged? I think this could theoretically be concluded... and it would be really awesome to have!

TurkeyMan:
This patch will let us catch C++ exceptions too right? That seems more useful to me, and it doesn't seem to be possible right now.

PR #8746, opened 2018-09-29
[Refactoring] Move the printing of overload candidates to a separate function

wolframw:
In `func.resolveFuncCall`, there were two nearly identical pieces of code for printing candidates in case more than one template overload matches or none of the function overloads are callable with the given argument types.
I mainly did this change in preparation for [issue 12694](https://issues.dlang.org/show_bug.cgi?id=12694), where I'd like to reuse this function.

### DIFF
--- a/src/dmd/func.d
+++ b/src/dmd/func.d
@@ -2719,25 +2719,7 @@ extern (C++) FuncDeclaration resolveFuncCall(const ref Loc loc, Scope* sc, Dsymb
                 td.kind(), td.parent.toPrettyChars(), td.ident.toChars(),
                 tiargsBuf.peekString(), fargsBuf.peekString());
 
-            // Display candidate templates (even if there are no multiple overloads)
-            int numToDisplay = numOverloadsDisplay;
-            overloadApply(td, (Dsymbol s)
-            {
-                auto td = s.isTemplateDeclaration();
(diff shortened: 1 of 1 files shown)


src/dmd/func.d
```diff
@@ -2848,6 +2802,52 @@ extern (C++) FuncDeclaration resolveFuncCall(const ref Loc loc, Scope* sc, Dsymb
...
+
+        // Too many overloads to sensibly display.
+        // Just show count of remaining overloads.
+        int num = 0;
+        overloadApply(nextOverload, (s) { ++num; return 0; });
+
+        if (num > 0)
+            .errorSupplemental(loc, "... (%d more, -v to show) ...", num);
+        return 1;   // stop iterating
+    });
```
wolframw: I didn't pass the scope here (as it was previously done [here](https://github.com/dlang/dmd/pull/8746/files#diff-ca782fb8b1ce12422a29fed42ffe0cadL2829)) because `overloadApply` doesn't use it anyway when the passed symbol is a `TemplateDeclaration` or a `FuncDeclaration`.

thewilsonator: I certainly need it for my template constraints DIP. but I can add it back later.

RazvanN7 (approved):

PR #8749, opened 2018-09-29
dmd/hdrgen.h: Move declarations of functions defined in hdrgen here

ibuclaw:
For consistency, extern(C++) functions in foo.d should have their equivalent in foo.h


jacob-carlborg (approved):

PetarKirov (approved):

ZombineDev:
Auto-merge toggled on

PR #8757, opened 2018-09-29
dmd.imphint: Remove C++ linkage from importHint

ibuclaw:
Update signature to return a const(char)[] instead.

### DIFF
--- a/src/dmd/expressionsem.d
+++ b/src/dmd/expressionsem.d
@@ -2417,8 +2417,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
 
         /* Look for what user might have meant
          */
-        if (const n = importHint(exp.ident.toChars()))
-            exp.error("`%s` is not defined, perhaps `import %s;` is needed?", exp.ident.toChars(), n);
+        if (const n = importHint(exp.ident.toString()))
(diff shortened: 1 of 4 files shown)


src/dmd/expressionsem.d
```diff
@@ -2326,8 +2326,8 @@ private extern (C++) final class ExpressionSemanticVisitor : Visitor
 
         /* Look for what user might have meant
          */
-        if (const n = importHint(exp.ident.toChars()))
-            exp.error("`%s` is not defined, perhaps `import %s;` is needed?", exp.ident.toChars(), n);
+        if (const n = importHint(exp.ident.toString()))
+            exp.error("`%s` is not defined, perhaps `import %s;` is needed?", exp.ident.toChars(), n.ptr);
```
jacob-carlborg: Isn't it better to use `%.*` and pass the length and pointer separately?

ibuclaw: Was taking advantage of that strings from importHint are null terminated.  Can do though.

thewilsonator:
The autotester OOM'd on FreeBSD.

src/dmd/expressionsem.d
jacob-carlborg: I’m guessing it won’t be null terminated after most code uses D strings.

jacob-carlborg (approved):

PetarKirov (approved):

PR #8763, opened 2018-09-29
__traits(allTargetInfos)

TurkeyMan:
Complements `getTargetInfo` in #8724 

Spec: https://github.com/dlang/dlang.org/pull/2477

### DIFF
--- a/changelog/alltargetinfos.dd
+++ b/changelog/alltargetinfos.dd
@@ -1,0 +1,9 @@
+Add `__traits(allTargetInfos)` to query the set of targetInfo data available from the backend.
+
+`allTargetInfos` evaluates to a tuple of strings containing all keys accepted by the implementation.
+
+---
+static assert([ __traits(allTargetInfos) ] == [ "cppRuntimeLibrary", "targetTriple", ... ]) // not a true list
+---
+
+Keys are implementation defined, allowing relevant data for exotic targets. A reliable subset which are always available shall be mentioned in the spec.
--- a/src/dmd/id.d
+++ b/src/dmd/id.d
@@ -406,6 +406,7 @@ immutable Msgtable[] msgtable =
     { "isReturnOnStack" },
     { "isZeroInit" },
     { "getTargetInfo" },
+    { "allTargetInfos" },
 
     // For C++ mangling
     { "allocator" },
--- a/src/dmd/target.d
+++ b/src/dmd/target.d
@@ -795,6 +795,18 @@ struct Target
                 return null;
         }
     }
+
+    /**
+     * Get all supported target-info keys
+     * Returns:
+     *  Array of targetInfo keys
+     */
+    extern (C++) static const(char*)* allTargetInfos(size_t* numTargetInfos)
+    {
+        __gshared immutable(char*[]) keys = [ "cppRuntimeLibrary" ];
+        *numTargetInfos = keys.length;
+        return keys.ptr;
+    }
 }
 
 /******************************
--- a/src/dmd/target.h
+++ b/src/dmd/target.h
@@ -87,4 +87,5 @@ struct Target
     static bool isReturnOnStack(TypeFunction *tf, bool needsThis);
     static d_uns64 parameterSize(const Loc& loc, Type *t);
     static Expression *getTargetInfo(const char* name, const Loc& loc);
+    static const char *const * allTargetInfos(size_t* numTargetInfos);
 };
--- a/src/dmd/traits.d
+++ b/src/dmd/traits.d
@@ -148,7 +148,8 @@ shared static this()
         "getVirtualIndex",
         "getPointerBitmap",
         "isZeroInit",
-        "getTargetInfo"
+        "getTargetInfo",
+        "allTargetInfos"
     ];
 
     traitsStringTable._init(48);
@@ -1798,6 +1799,21 @@ Lnext:
             return new ErrorExp();
         }
         return r.expressionSemantic(sc);
+    }
+    if (e.ident == Id.allTargetInfos)
+    {
+        if (dim != 0)
+            return dimError(0);
+
+        size_t numTargetInfos;
+        const(char*)[] targetInfoKeys = Target.allTargetInfos(&numTargetInfos)[0 .. numTargetInfos];
+
+        Expressions* exps = new Expressions();
+        foreach (k; targetInfoKeys)
+            exps.push(new StringExp(e.loc, cast(char*)k));
+
+        Expression ex = new TupleExp(e.loc, exps);
+        return ex.expressionSemantic(sc);
     }
 
     extern (D) void* trait_search_fp(const(char)* seed, ref int cost)
--- a/test/compilable/traits.d
+++ b/test/compilable/traits.d
@@ -12,7 +12,13 @@ class C19152
     }
 }
 
+static assert(is(typeof(__traits(allTargetInfos)[0]) == string));
 static assert(is(typeof(__traits(getTargetInfo, "cppRuntimeLibrary")) == string));
+
+version (DigitalMars)
+{
+    static assert([__traits(allTargetInfos)] == ["cppRuntimeLibrary"]);
+}
 
 version (CppRuntime_Microsoft)
 {
--- a/test/fail_compilation/traits.d
+++ b/test/fail_compilation/traits.d
@@ -12,6 +12,7 @@ fail_compilation/traits.d(100): Error: `getTargetInfo` key `"not_a_target_info"`
 fail_compilation/traits.d(101): Error: string expected as argument of __traits `getTargetInfo` instead of `100`
 fail_compilation/traits.d(102): Error: expected 1 arguments for `getTargetInfo` but had 2
 fail_compilation/traits.d(103): Error: expected 1 arguments for `getTargetInfo` but had 0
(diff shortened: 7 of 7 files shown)


WalterBright:
Significant discussion of pros and cons of this feature are at https://github.com/dlang/dmd/pull/8724 but should be continued here.

TurkeyMan:
I'm not sure what further case can be made. @thewilsonator @jacob-carlborg ?
I'm not sure how you want us to convince you at this point. Should we consider this patch rejected?

thewilsonator:
@TurkeyMan please rebase on master and squash so only one relevant commit is present.

thewilsonator:
> Significant discussion of pros and cons 

What cons? There was no technical disadvantages listed (aside for perhaps docs, in which case I assume you missed my Bhudda quote: that this is _much_ better than no documentation and won't get out of sync). Process grumbles, sure, but no technical dismerit. 

I really don't know what else to say, other than to repeat myself again.

jacob-carlborg:
I don’t have anything to add that hasn’t already been said. I’m not the best person at convincing others.

WalterBright:
Seems like an ideal candidate for the Phantom Zone.

TurkeyMan:
@WalterBright I feel like you could make your case stronger...

thewilsonator:
> Phantom Zone
> Has value/information for future work but closed for now

That is usually applied to enhancement requests that have lingered around for a while, have some objection or technical reason why they are perhaps not the best approach to the problem of that the problem is OK left as is unsolved.

This has been open for two(!) days. You have given _no reasons_ as to why you think this shouldn't go through and we believe we have presented a compelling case but apparently not (not that you have bothered to tell us _why_ you think it isn't compelling enough), and have only grumbled about process. 

As I have said before this is about discoverability.

There will end up being quite a lot of information available through this trait, and as more information is added, much of which will be compiler/compiler middle and backend (version!)/target/system/command line flag/etc specific, and older compiler versions lag behind it will become difficult to know what information is available for a given build configuration, let alone collate it together for diagnostics.

The quip that this is useless with proper documentation misses the forest for the trees: You can't know to consult the documentation if you don't know that  the thing you need or want exists!

To suggest:

>there is no use case in the code you're working on.

Is laughable (But in case you missed it: discoverability and diagnostics). As is the suggestion that different might define the same keys with the differing meanings and that it is useless without documentation (How many ways are there to interpret what "cacheline.size" is?  Do you _really_ need to consult documentation for a name like that?).

Sound familiar? 
Thats because _I have said it all before_! 

If you're going to stonewall this, we'll just add it to LDC/GDC and there can be another reason not to use DMD for anything serious.

/rant

TurkeyMan:
> My advise is to present the motivation behind the PR and show why other solutions doesn't work or have its own set of problems.

We have. At risk of repeating;
1. Discoverability; the set of things are implementation defined, will be niche, and will NOT be maintained on a webpage somewhere. Find me the webpage that even lists LDC and GDC's command line args (I'm sure it exists... but *find* it) ;)
2. Dumping this state can be very useful for logging and crash-dumps. There's a ton of diagnostic info here.

Walter hasn't made any strong arguments who those 2 key usages are worthless. In terms of implementation baggage, this is trivial to maintain (the way I implemented it is self-maintaining), and the footprint is very small.

WalterBright:
> This has been open for two(!) days.

My intent is for you to not waste your time pouring effort into this. If I wait, then I will be criticized for letting you waste your time. Since I was going to be criticized anyway, I will take it for not wasting your time :-) Your time is valuable.

> I feel like you could make your case stronger...

This is backwards. To get a core feature into D, the proponents need to make a strong case. It is not up to me to make a strong case against it. You have made a case, and I have responded to it. It is not a compelling case. I did not say "worthless", and I did give reasons.

>  Find me the webpage that even lists LDC and GDC's command line args (I'm sure it exists... but find it) ;)

You should speak to the LDC and GDC crew about that. You can find all the dmd switches by typing one of the following:
```
dman dmd
dmd -man
```
and an abbreviated list with:
```
dmd
```
If you'd like to add a `dmd -traits` switch to list them, that may be a good idea, and is pretty lightweight (being just text and not adding to the core language).

The purpose of the Phantom Zone is for good ideas that aren't quite good enough, and their time may come or they can serve as inspiration for a better idea. Some of my PRs wound up there, too.

thewilsonator:
>This is backwards. To get a core feature into D, the proponents need to make a strong case.

Which we believe we did.

> It is not up to me to make a strong case against it. 

True but you stated that 

>> We disagree on several points,

and never said what you disagreed about.

> You have made a case, and I have responded to it. It is not a compelling case. 

We'd like to know _why_ you think the case we have made is not compelling.

> I did not say "worthless",

No, you said:

>>there is no use case in the code you're working on.

which is close enough.

> and I did give reasons.

Humour me and list them again, please, so we don't waste time in the future. If you are referring to 

> Several issues here

1 & 2 are about process. 
3, we believe we have provided compelling evidence. You have stated that it is not but have not given reasons for. 
4 is about process and this was Iain's request because `__CXXLIB__` was merged before he got the time to review it, such is the nature of merging things.
5 is process and contains the false statement "there is no use case in the code you're working on.".

ibuclaw:
> 4 is about process and this was Iain's request because __CXXLIB__ was merged before he got the time to review it, such is the nature of merging things.

Frankly I'm surprised `__CXXLIB__` got approved. It should have been have been pushed to be a trait from the start (eg: `__traits(getCxxLib)`), as all other compiler internals are exposed in this way.

What came about after discussing this was getTargetInfo, and naturally allTargetInfo followed when it became obvious that getTargetInfo itself may potentially embody a lot of information.

WalterBright:
> 1. Discoverability

I'd like to see a use case for how this is superior to, say, having a command line switch to print it out. Because I'm not seeing it.

> 2. Logging and crash dumps

https://dlang.org/phobos/core_cpuid.html

thewilsonator:
Thank you! That would have been useful to know earlier ;)

1. Will think.

2.
>Currently only works on x86 and Itanium CPUs.

LDC supports ARM, AArch64, NVPTX, OpenCL, WebAssembly, probably a whole lot more. GDC  supports ARM and probably a bunch of others. Also that is only available dynamically, so it is not the same, e.g. if something has crashed, crosschecking the known true dynamic info against the static info could prevent some serious headaches.

TurkeyMan:
> > 1. Discoverability
> 
> I'd like to see a use case for how this is superior to, say, having a command line switch to print it out. Because I'm not seeing it.

I've mentioned iterating/dumping the data multiple times.
Also, I think the patch to implement a command line version would have a much larger surface area than this, and also be less useful.

> > 1. Logging and crash dumps
> 
> https://dlang.org/phobos/core_cpuid.html

That's a completely different thing. runtime detection vs compile-target. They're complementary, but they're not the same thing.

WalterBright:
>> Discoverability
> I've mentioned iterating/dumping the data multiple times.

Computer programs cannot programmatically "discover" things. This is why I ask.

> Also, I think the patch to implement a command line version would have a much larger surface area than this, and also be less useful.

It's just a text string, and more importantly, it does not change the core language. I have no idea why reading the output text from a flag is not as useful as reading the output text of a program you wrote to do the same thing. And, you don't have to write a program to do it.

> That's a completely different thing. runtime detection vs compile-target. They're complementary, but they're not the same thing.

I don't understand what your use case is. If you care what your compiler CPU target was, surely you must have specified it at some point, you don't need to discover it with a fishing expedition for undocumented traits.

TurkeyMan:
> Computer programs cannot programmatically "discover" things. This is why I ask.

I'm not sure what you mean?
Are you asking how to use it in __traits form?
When I'm working in D, I do a crazy amount of my diagnostic work with pragma(msg)... like, SO MUCH with it.
Going to C++, I think it's objectively the thing I miss the most.
pragma(msg) is how I would 'discover' data in that pool. It's convenient, and doubly useful in that form for writing actual programs as it is for diagnostics.

> And, you don't have to write a program to do it.

I like and prefer writing programs. That's what I do.
I have no idea how to operate DMD (or LDC, or GDC) from the command line, but I know how to write D code.
I've spent ages trying to penetrate `-dumpspecs` and various command line tooling trying to understand compilers capabilities. It's a mess. Providing it within the compiler is uniform, reliable, easier to access, and standardised.

> I don't understand what your use case is. If you care what your compiler CPU target was, surely you must have specified it at some point

This is a very case-sensitive question. In many cases I want to respond to the user selection; if I'm trying to populate a stack buffer for instance relative to a cache line or something, then you want to sample that from the backend. If you want to emit diagnistic data, user will usually only specify some very high level detail, and the rest is a big surprise waiting to be revealed.

TurkeyMan:
@rainers That's true, that would also work... but I don't think it's the name of the trait that's up for debate. It's the idea in principle.

jacob-carlborg:
Not adding a separate trait would reduce the surface area slightly. 

> It's the idea in principle.

Didn’t Walter suggested a flag?

thewilsonator:
As in the principle of a `__trait`, he did suggest a flag. However a flag only solves the discoverability problem (unless you are Manu, then it doesn't :) ), id doesn't solve runtime data dumping problem.

thewilsonator:
https://github.com/dlang/dmd/pull/10002

PR #8764, opened 2018-09-29
dmd.expression: Remove C++ linkage from static member functions

ibuclaw:
The `extractLast` function can use an `out` parameter instead of accepting a pointer.  I also noticed that the documentation seems to be wrong (it returns the rightmost, not the leftmost part).

Also noticed there are a few places where `combine(combine(...), ...)` was being used instead of one of the helper functions.

### DIFF
--- a/src/dmd/constfold.d
+++ b/src/dmd/constfold.d
@@ -1455,6 +1455,49 @@ int sliceCmpStringWithArray(const StringExp se1, ArrayLiteralExp ae2, size_t lo1
     return 0;
 }
 
+/** Copy element `Expressions` in the parameters when they're `ArrayLiteralExp`s.
+ * Params:
+ *      e1  = If it's ArrayLiteralExp, its `elements` will be copied.
+ *            Otherwise, `e1` itself will be pushed into the new `Expressions`.
+ *      e2  = If it's not `null`, it will be pushed/appended to the new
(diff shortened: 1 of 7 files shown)


WalterBright:
I don't get it. How does this improve the code?

ibuclaw:

src/dmd/constfold.d
```diff
@@ -1455,6 +1455,49 @@ int sliceCmpStringWithArray(const StringExp se1, ArrayLiteralExp ae2, size_t lo1
...
+ * Params:
+ *      e1  = If it's ArrayLiteralExp, its `elements` will be copied.
+ *            Otherwise, `e1` itself will be pushed into the new `Expressions`.
+ *      e2  = If it's not `null`, it will be pushed/appended to the new
+ *            `Expressions` by the same way with `e1`.
+ * Returns:
+ *      Newly allocated `Expressions`. Note that it points to the original
+ *      `Expression` values in e1 and e2.
+ */
+private Expressions* copyElements(Expression e1, Expression e2 = null)
```
RazvanN7: Isn't this change identical with : https://github.com/dlang/dmd/pull/8778/files ?

ibuclaw: Yes, I'm going to close this PR, and push individual changes into new ones.

PR #8765, opened 2018-09-29
dmd.identifier: Remove C++ linkage from static members

ibuclaw:
Make static variables private to module.

### DIFF
--- a/src/dmd/identifier.d
+++ b/src/dmd/identifier.d
@@ -128,7 +128,7 @@ nothrow:
         return DYNCAST.identifier;
     }
 
-    extern (C++) __gshared StringTable stringtable;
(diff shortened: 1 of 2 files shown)


WalterBright:
What is gained by de-encapsulating a private static member?

ibuclaw:
> What is gained by de-encapsulating a private static member?

I was about to say they are all public, though now I see that `fullPathStringTable` isn't.

Just tested, and Identifier.stringtable is private to the class as well, so I'll move them back in.

ibuclaw:

jacob-carlborg (approved):

ibuclaw:
Cherry-picked into #8773.

PR #8766, opened 2018-09-29
dmd.init: Remove Initializer.arraySyntaxCopy

ibuclaw:
This is dead code.


WalterBright:
I suggest removing it rather than moving it, and update the title.

ibuclaw:
> I suggest removing it rather than moving it, and update the title.

OK... just making sure that it won't be used for some future purpose.

ibuclaw:
Done.

jacob-carlborg (approved):

PR #8767, opened 2018-09-29
dmd.dmodule: Remove C++ linkage from static members

ibuclaw:
(Module is peculiar as the driver interacts with it, so is left alone for now...)


WalterBright (changes requested):
Please fix the title, as the current title has nothing to do with the actual changes.

Please rename `resolve` to something that is both a little more illuminating as to its purpose, and something that is greppable. There are too many diverse functions already named `resolve`.

ibuclaw:

jacob-carlborg (approved):

ibuclaw:
Cherry-picked into #8773.

PR #8770, opened 2018-09-29
dmd.dscope: Remove C++ linkage from static members

ibuclaw:


WalterBright:
Please start using more descriptive titles.

WalterBright:
I don't understand what is improved by removing all the static member functions. It just looks like a giant reshuffling of the code, making it very difficult to track changes to code across time, and for what purpose?

ibuclaw:
> I don't understand what is improved by removing all the static member functions. It just looks like a giant reshuffling of the code, making it very difficult to track changes to code across time, and for what purpose?

I have two more patches pending, but I'll make a pause.

This is removing from the C++ interface functions that have no purpose being exposed.

Alternatively can just mark them as `extern(D)` and leave them where they are, I'm not really fussed one way or the other, I made them free functions as many can be marked private to the module they are defined in.

WalterBright:
They can be marked `extern (D)`, and `private` as needed, and then it would be clear they are not part of the C++ interface. I understand trying to minimize the C++ interface. I'd prefer that, as large code reshuffles make diffing different versions very hard, and will cause much rebasing of existing PRs.

WalterBright:
thanks!

ibuclaw:
As these PRs are become trivial /  couple line changes, it might be better squashing them into one PR with them all combined.

ibuclaw:
Fixed up as discussed.

jacob-carlborg (approved):

PR #8775, opened 2018-09-29
Fix C++ headers following merge of #8724

ibuclaw:
I had my suspicions that there would be something wrong with #8724.


TurkeyMan:
Who so suspicious?

jacob-carlborg (approved):

PetarKirov (approved):

PR #8776, opened 2018-09-30
Add aliases to ASTCodegen for dmd#8591

thewilsonator:
So that #8591's dtoh can be built against an unmodified dmd (e.g as a dub package).

cc @ibuclaw

### DIFF
--- a/src/dmd/astcodegen.d
+++ b/src/dmd/astcodegen.d
@@ -41,12 +41,31 @@ struct ASTCodegen
 
     alias MODFlags                  = dmd.mtype.MODFlags;
     alias Type                      = dmd.mtype.Type;
-    alias Tident                    = dmd.mtype.Tident;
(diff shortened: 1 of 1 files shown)


jacob-carlborg (approved):

src/dmd/astcodegen.d
```diff
@@ -41,12 +41,31 @@ struct ASTCodegen
 
     alias MODFlags                  = dmd.mtype.MODFlags;
     alias Type                      = dmd.mtype.Type;
-    alias Tident                    = dmd.mtype.Tident;
```
ZombineDev: @ibuclaw why did you remove those aliases in #8591?

thewilsonator: They weren't, they were reordered alphabetically.

ZombineDev: Right, nevermind.

PR #8779, opened 2018-09-30
dmd.dmodule: Remove workaround for compilers older than 2.069

ibuclaw:
Minimum version should be circa 2.071 by now.


jacob-carlborg (approved):

kinke:
LDC on Linux can still be bootstrapped with our C++-only LTS version based on 2.068; I guess it's the same for macOS, but unfortunately that's not CI-tested.

ibuclaw:
@kinke - from what I recall, this was added due to a dmd-specific backend bug.  LDC 2.068 LTS is safe.

PetarKirov (approved):

PR #8780, opened 2018-09-30
dmd.globals: Move vendor string back to Global struct

ibuclaw:
This is blocking #7534 and #7559, due to the circular dependencies that happen between dmd.compiler and dmd.globals.

### DIFF
--- a/src/dmd/compiler.d
+++ b/src/dmd/compiler.d
@@ -12,7 +12,10 @@
 
 module dmd.compiler;
 
+/**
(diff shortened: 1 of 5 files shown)


WalterBright:
Circular dependencies? dmd.compiler doesn't import anything. It has no dependencies, circular or otherwise.

RazvanN7 (approved):

ibuclaw:
@RazvanN7 it shouldn't. All functionality being added is part of the semantic.

PR #8781, opened 2018-10-01
Support `extern(C++, "A::B")`

TurkeyMan:
Such that it's able to compose namespaces more easily.

Blocks: https://github.com/dlang/druntime/pull/2286

### DIFF
--- a/src/dmd/parse.d
+++ b/src/dmd/parse.d
@@ -2162,20 +2162,40 @@ final class Parser(AST) : Lexer
                             while (1)
                             {
                                 AST.StringExp stringExp = cast(AST.StringExp)parsePrimaryExp();
-                                const(char)[] name = stringExp.toStringz();
-                                if (name.length == 0)
+                                const(char)[] ns = stringExp.toStringz();
+                                bool again = true;
+                                while (again)
                                 {
-                                    error("invalid zero length C++ namespace");
-                                    idents = null;
-                                    break;
+                                    again = false;
+                                    const(char)[] name = ns;
+                                    if (name.length >= 2)
+                                    {
+                                        // handle extern(C++, "namespace::namespaces")
+                                        foreach (i; 0 .. name.length - 1)
+                                        {
+                                            if (name[i .. i + 2] == "::")
+                                            {
+                                                ns = name[i + 2 .. $];
+                                                name = name[0 .. i];
+                                                again = true;
+                                                break;
+                                            }
+                                        }
+                                    }
+                                    if (name.length == 0)
+                                    {
+                                        error("invalid zero length C++ namespace");
+                                        idents = null;
+                                        break;
+                                    }
+                                    else if (!Identifier.isValidIdentifier(name))
+                                    {
+                                        error("expected valid identifer for C++ namespace but got `%s`", name.ptr);
+                                        idents = null;
+                                        break;
+                                    }
+                                    idents.push(Identifier.idPool(name));
                                 }
-                                else if (!Identifier.isValidIdentifier(name))
-                                {
-                                    error("expected valid identifer for C++ namespace but got `%s`", name.ptr);
-                                    idents = null;
-                                    break;
-                                }
-                                idents.push(Identifier.idPool(name));
                                 if (token.value == TOK.comma)
                                 {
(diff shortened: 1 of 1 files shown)


TurkeyMan:
@look-at-me I put this back because it's much easier to compose namespaces when they're just one string.
Problem is, it doesn't accept a `StringExp`...

TurkeyMan:
Code like this doesn't work:

```d
version (OSX)
    enum UseCxx11Namespace = true;
else version (FreeBSD)
    enum UseCxx11Namespace = true;
else
    enum UseCxx11Namespace = false;

extern (C++, "std" ~ (UseCxx11Namespace ? "::__1" : "")):
```

TurkeyMan:
Of course, this would also be fine:
```d
version (OSX)
    alias NS = AliasSeq!("std", "__1");
else version (FreeBSD)
    alias NS = AliasSeq!("std", "__1");
else
    alias NS = AliasSeq!("std");

extern (C++, NS):
```

TurkeyMan:
Well this patch isn't very big, so supporting '::' isn't exactly un-simple. It seemed simpler for user-code to compose a single string with '::' than to manufacture a tuple, and then `expand`.

But yes. A tuple of StringExp's... that would be good.

rainers:
I suspect anything that evaluates the identifiers in the parser won't help https://github.com/dlang/druntime/pull/2286, it needs to be delayed into the semantic phase to allow symbol lookup and CTFE.

thewilsonator:
```d
version (OSX)
    alias NS = AliasSeq!("std", "__1");
else version (FreeBSD)
    alias NS = AliasSeq!("std", "__1");
else
    alias NS = AliasSeq!("std");

extern (C++, NS):
```
As much as I like it, surely it is ambiguous, because `extern (C++, NS):` _generates_ an identifier `NS` which would clash.

I think @rainers is right.

or you could go with `__trait(getTargetInfo, "c++.std.namespace")` (or the like) ;)

TurkeyMan:
Yeah, I couldn't work out how to move this block into the semantic... which is why I stopped hacking on it >_<

thewilsonator:
Maybe store an identifier and check later to see if it is already in scope? 

OTOH, allowing `A::B` in the name and using `__trait(getTargetInfo, "c++.std.namespace")` _would_ be an effective solution since it won't be mistaken for a identifier.

TurkeyMan:
I thought about that... but it's not really a general solution. It's a hack ;)
This is gonna come up again.

thewilsonator:
Whatever you decide on, it needs a test.

TurkeyMan:
Okay, so I thought of something overnight...
We can take a leaf out of C++'s book and wrap the ambiguity in parens to make it not-ambiguous ;)

I think this is as good as it gets here:
```d
extern(C++, NS): // scoped namespace
extern(C++, "literal"): // accept literal, since is 99% case
extern(C++, (StringExp)): // add parens for disambiguation, any string expression is accepted
```

TurkeyMan:
That may be tidied in the future too; there were people who called that extern(C++) and declaring a scoped namespace should be separate parallel features (that may be useful in conjunction).
It may be desirable to move the declaration for a named scope out of extern(C++) and into its own syntax. After such a transition, the parens can be lifted.

jacob-carlborg:
Can the scoped namespace (`extern(C++, NS)`) be emulated with the string literal syntax (`extern(C++, "NS"`) and a struct (or something)? If that's the case perhaps we should look into deprecate the scoped namespace syntax.

WalterBright:
Marking as WIP because it clearly needs much more discussion and design work.

look-at-me:
The ideal solution would be just allow CTFE as you would normally expect, but that would be a breaking change, which is why it wasn't implemented in the first place. Deprecating the old syntax and then eventually allowing CTFE might be possible. With some sort of workaround to allow "mixin" to be used in the mean time.

```D
enum NS = `"std", "__1"`;
extern(C++, mixin(NS)) void foo();
```

Not pretty but would allow for CTFE in the meantime until the deprecation is over should that route be taken.

TurkeyMan:
Closing in favour of #8787

